Issue #1955, opened 2026-06-24
Notification component — new core-ui severity notification (accent bar + icon + title/description + actions + suggestion + close)


jeremy-babylonlabs:
## Summary
The liquidation/notification cards (#1947–#1954) all share the same anatomy:
**left severity accent bar, icon, title, description, optional action buttons,
optional suggestion box, severity-tinted border/background, and an optional
close/dismiss control.** Build this as a new core-ui primitive — `Notification`
— and have every notification card consume it. This is a prerequisite for the
notification tickets.

## Why a new component (not extend `Callout`)
The existing core-ui `Callout`
(`packages/babylon-core-ui/src/components/Callout/Callout.tsx`) is a different
pattern and stays untouched:

| | `Callout` (existing) | `Notification` (new) |
|---|---|---|
| Layout | icon chip + text, vertical | left accent bar + icon + text, actions right |
| Border / bg | neutral | severity-tinted + left accent stripe |
| Actions | none | inline, right-aligned buttons |
| Dismiss / suggestion box | none | optional |
| Role | inline message in flows/dialogs | dashboard-level position notification |

## Current state (what this replaces)
The vault position banner re-implements this presentation inline:
- `services/vault/src/components/simple/PositionNotificationBanner/PositionNotificationBanner.tsx`
  (title/detail/suggestion stack, secondary warnings)
- `…/PositionNotificationBanner/constants.ts` → `SEVERITY_STYLES`
  (colored-border boxes: `red`/`yellow`/`green`/`soft`)
- `…/PositionNotificationBanner/BannerActions.tsx` (action buttons)

## Proposed API
New `packages/babylon-core-ui/src/components/Notification/Notification.tsx`:
- `variant` / `severity` — drives accent bar + tinted border/bg + default icon
  (map from the vault `BannerSeverity`: red / yellow / green / soft).
- `title?`, `children` (description), `icon?` (override).
- `actions?: ReactNode` — buttons slot.
- `actionsPlacement?: "inline" | "below"` — same row as content, or stacked
  below the suggestion box. Default `"inline"` (matches Figma).
- `suggestion?: ReactNode` — nested sub-box below the body.
- `onClose?: () => void` — renders a dismiss (X) control when provided; omitted
  = no close affordance.
- Export from `components/index.ts`; add `Notification.stories.tsx`.

## Scope
1. Build `Notification` in core-ui per the API above (left accent bar, severity
[...]

Issue #1947, opened 2026-06-24
Notification — Liquidation callout box


jeremy-babylonlabs:
## Summary
The "Liquidation is X% away" callout is implemented but its styling does not
match Figma — needs a red callout box with a danger icon and inline action
buttons.

## Current state
- Implemented: `services/vault/src/components/simple/PositionNotificationBanner/`
  (`PositionNotificationBanner.tsx`, `BannerActions.tsx`).
- Copy exists: `COPY.liquidationWarnings.urgent.approachingTitle` /
  `approachingDetail` / `approachingSuggestion`, buttons
  `COPY.banner.addCollateral` / `COPY.banner.repayDebt`.
- Styling differs from Figma (layout/icon/box treatment).

## Design (Figma)
- Red callout box with danger icon
- Title: **Liquidation is 4.3% away**
- Body: *"A drop of BTC to $80,621 (just 4.3% lower than now) triggers your first
  liquidation event."*
- Inline buttons: **Add Collateral** / **Repay Debt**

https://www.figma.com/design/7Nf1LhRG3ad3RYhMhgHiQE/TBV-x-Master-File--Updated-Flow-?node-id=6502-106278&t=U8J29QsulIdS2d7e-4

## Scope
1. Restyle the urgent banner to the Figma callout (red box, danger icon, inline
   button row).
2. No new copy (reuse existing strings); confirm icon asset.
3. Visual regression / snapshot update for `PositionNotificationBanner`.

## Acceptance criteria
- [ ] Urgent callout matches Figma (red box + danger icon + inline buttons)
- [ ] Existing copy reused; lint + tests pass

PR #1960, opened 2026-06-24
feat(vault): critical near-liquidation top banner

### LINKED ISSUE #1946: Notification — Top critical banner
Relates to #1947 

## Summary
Render a full-width red top banner ("⚠️ Critical — liquidation in X%") above the
header when the position is at critical liquidation risk. Not implemented today.

## Current state
- `core-ui` `TopBanner` component exists:
  `packages/babylon-core-ui/src/components/TopBanner/TopBanner.tsx`.
- `usePositionNotifications` (`services/vault/src/applications/aave/hooks/usePositionNotifications.ts`)
  + `deriveBannerState` (`applications/aave/positionNotifications/bannerSeverity.ts`)
  already compute a `red` severity for urgent warnings.
- No critical top banner is rendered anywhere.

## Design (Figma)
- Full-width red banner above the header
- Text: **⚠️ Critical — liquidation in 4.3%** (distance % from the urgent warning)

https://www.figma.com/design/7Nf1LhRG3ad3RYhMhgHiQE/TBV-x-Master-File--Updated-Flow-?node-id=6502-106261&t=U8J29QsulIdS2d7e-4

## Scope
1. Render `TopBanner` (red variant) above the app header when
   `deriveBannerState(...).severity === "red"`.
2. Drive the % from the urgent warning's `distancePct` (already computed).
3. Add copy to `copy.ts` (e.g. `COPY.topBanner.critical(distancePct)`).
4. Decide placement (above header layout) and dismiss behavior (likely
   non-dismissible while critical — confirm with design).

## Acceptance criteria
- [ ] Red top banner shows above header when severity is `red`, hidden otherwise
- [ ] Distance % matches the urgent warning

gbarkhatov:
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/1946

Screenshot_2026-06-24_16-38-53

Screenshot_2026-06-24_16-38-46

## What this does

Adds a full-width red banner above the app header that warns the user when their position is close to liquidation — for example "⚠ Critical — liquidation in 4.3%". It only appears when the position is at critical (red) risk, and it cannot be dismissed while critical.

## Why

When a position is near liquidation the user can lose BTC, so this warning needs to be impossible to miss. The existing position notification lives in the middle of the dashboard, below the fold; this banner puts the same critical signal at the very top of the screen where the user sees it immediately. It is the first item in the notification design set.

## Scope (important for the reviewer)

The Figma "URGENT" screen shows two things: the top red strip (this PR) and a detailed callout box ("Liquidation is 4.3% away" with Add Collateral / Repay Debt buttons). Those are two separate issues. This PR only builds the top red strip. The detailed callout already exists in code today and its Figma restyle is tracked separately in https://github.com/babylonlabs-io/babylon-toolkit/issues/1947 — it is not touched here.

## How it works

The banner reuses the data the dashboard already computes (`usePositionNotifications` → `deriveBannerState`). When the severity is "red" it reads the distance-to-liquidation and renders the strip; otherwise it renders nothing. The text and percentage stay in sync with the existing mid-page banner because both read the exact same severity source.

## The main design decision (and why)

The banner has to appear above the header, but the data it needs (the Aave position providers) only exists inside the dashboard route, lower in the tree. So a component placed at the header level cannot read that data directly. We considered two approaches:

- Approach A (chosen): keep the banner owned by the dashboard (where the data and the debug override already live) and "portal" its output up into a small slot rendered above the header. No provider changes, no second data fetch, and it automatically works with the existing debug panel because it shares the same override the mid-page banner uses.
- Approach B (rejected): move the Aave providers and a new shared notifications context above the header so a header-level component could fetch the data itself. This is cleaner in theory (single source of truth) but it changes how those providers mount for the whole app and touches the router — more risk and a bigger change than this banner warrants.

We chose A because it is the smallest, lowest-risk change and it keeps the debug panel working out of the box, which was an explicit requirement.

## Notable detail: non-dismissible

Per design, the user must not be able to hide an imminent-liquidation warning, so the banner has no close button. The shared core-ui `TopBanner` previously always rendered a close button, so it was changed to make the close (and click) optional. This is backward compatible — the only other user of `TopBanner` still passes those callbacks and keeps its close button.

## Changes

- `packages/babylon-core-ui` — `TopBanner`: `onDismiss` and `onClick` are now optional; the close button renders only when `onDismiss` is provided. Added a story for the non-dismissible variant.
- `services/vault` — new `CriticalLiquidationTopBanner` component that portals the red strip above the header and shows only at red severity.
- `services/vault` — `RootLayout`: added the portal slot at the top of the banner stack.
- `services/vault` — `DashboardPage`: feeds the banner the same debug-aware result as the existing banner, plus a click handler that scrolls down to the detailed notification.
- `services/vault` — `copy.ts`: added the banner strings.
- Tests for the new component.

## How to test it

Run the vault app with `NEXT_PUBLIC_FF_ENABLE_LIQUIDATION_NOTIFICATIONS=true` and `NEXT_PUBLIC_FF_POSITION_DEBUG_PANEL=true`. Open the debug panel, switch to manual mode, and push the position into the red (near-liquidation) range. You should see the red strip appear above the header (with no close button) at the same time as the existing mid-page banner, both showing the same percentage. Move the position back out of the red range and both disappear.

## Related

- Depends conceptually on the shared notification primitive in https://github.com/babylonlabs-io/babylon-toolkit/issues/1955
- Pairs with the detailed callout restyle in https://github.com/babylonlabs-io/babylon-toolkit/issues/1947

### DIFF
--- a/packages/babylon-core-ui/src/components/TopBanner/TopBanner.tsx
+++ b/packages/babylon-core-ui/src/components/TopBanner/TopBanner.tsx
@@ -30,6 +30,13 @@ export interface TopBannerProps {
    * Optional custom icon
    */
   icon?: React.ReactNode;
+  /**
+   * Optional ARIA role for the banner wrapper. Use for a non-interactive banner
+   * that should still be announced — e.g. `"alert"` for a critical warning. When
+   * omitted, a clickable banner is exposed as a `button` and a non-clickable one
+   * carries no role.
+   */
+  role?: React.AriaRole;
 }
 
 export const TopBanner = ({
@@ -39,6 +46,7 @@ export const TopBanner = ({
   onDismiss,
   className,
   icon,
+  role,
 }: TopBannerProps) => {
   if (!visible) {
     return null;
@@ -59,7 +67,7 @@ export const TopBanner = ({
         className,
       )}
       onClick={onClick}
-      role={isClickable ? "button" : undefined}
+      role={role ?? (isClickable ? "button" : undefined)}
       tabIndex={isClickable ? 0 : undefined}
     >
       <div className="bbn-top-banner-content">
--- a/packages/babylon-core-ui/src/components/TopBanner/TopBanner.tsx
+++ b/packages/babylon-core-ui/src/components/TopBanner/TopBanner.tsx
@@ -13,13 +13,15 @@ export interface TopBannerProps {
    */
   message: string;
   /**
-   * Callback when banner is clicked
+   * Callback when banner is clicked. Omit for a non-interactive banner — the
+   * banner then drops its `button` role and pointer affordance.
    */
-  onClick: () => void;
+  onClick?: () => void;
   /**
-   * Callback when banner is dismissed
+   * Callback when banner is dismissed. Omit to render a non-dismissible banner
+   * (no close button) — e.g. a critical alert the user must not be able to hide.
    */
-  onDismiss: () => void;
+  onDismiss?: () => void;
   /**
    * Optional custom className
    */
@@ -44,15 +46,21 @@ export const TopBanner = ({
 
   const handleDismiss = (e: React.MouseEvent) => {
     e.stopPropagation();
-    onDismiss();
+    onDismiss?.();
   };
+
+  const isClickable = Boolean(onClick);
 
   return (
     <div
-      className={twJoin("bbn-top-banner", className)}
+      className={twJoin(
+        "bbn-top-banner",
+        !isClickable && "!cursor-default",
+        className,
+      )}
       onClick={onClick}
-      role="button"
-      tabIndex={0}
+      role={isClickable ? "button" : undefined}
+      tabIndex={isClickable ? 0 : undefined}
     >
       <div className="bbn-top-banner-content">
         {icon}
@@ -60,14 +68,16 @@ export const TopBanner = ({
           {message}
         </Text>
       </div>
-      <button
-        onClick={handleDismiss}
-        className="bbn-top-banner-dismiss-btn"
-        aria-label="Dismiss banner"
-        type="button"
-      >
-        <CloseIcon size={16} className="bbn-top-banner-dismiss-icon" />
-      </button>
+      {onDismiss && (
+        <button
+          onClick={handleDismiss}
+          className="bbn-top-banner-dismiss-btn"
+          aria-label="Dismiss banner"
+          type="button"
+        >
+          <CloseIcon size={16} className="bbn-top-banner-dismiss-icon" />
+        </button>
+      )}
     </div>
   );
 };
--- a/services/vault/src/components/simple/CriticalLiquidationTopBanner/CriticalLiquidationTopBanner.tsx
+++ b/services/vault/src/components/simple/CriticalLiquidationTopBanner/CriticalLiquidationTopBanner.tsx
@@ -23,18 +23,17 @@ interface CriticalLiquidationTopBannerProps {
    * shows only when this resolves to a `red` (urgent) banner severity.
    */
   result: CalculatorResult | null;
-  /** Scroll the detailed position banner into view when the strip is clicked. */
-  onShowDetails: () => void;
 }
 
 /**
  * Full-width critical (near-liquidation) banner shown above the header when the
- * position is at `red` severity. Non-dismissible by design — an imminent
- * liquidation warning the user must not be able to hide.
+ * position is at `red` severity. A non-interactive `role="alert"` so assistive
+ * tech announces it; non-dismissible by design — an imminent liquidation warning
+ * the user must not be able to hide. The actionable Add Collateral / Repay Debt
+ * controls live on the detailed position banner below.
  */
 export function CriticalLiquidationTopBanner({
   result,
-  onShowDetails,
 }: CriticalLiquidationTopBannerProps) {
   // The portal target lives in RootLayout (mounted before this component);
   // resolve it on mount so we can portal into it once available.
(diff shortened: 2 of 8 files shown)


packages/babylon-core-ui/src/components/TopBanner/TopBanner.tsx
```diff
@@ -44,30 +46,38 @@ export const TopBanner = ({
...
+        "bbn-top-banner",
+        !isClickable && "!cursor-default",
+        className,
+      )}
       onClick={onClick}
-      role="button"
-      tabIndex={0}
+      role={isClickable ? "button" : undefined}
+      tabIndex={isClickable ? 0 : undefined}
     >
```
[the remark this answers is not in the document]
gbarkhatov: No need to have keyboard, we did it in a better way in latest commit. It's an info (alert) non dismissable component

jonybur (approved):

services/vault/src/components/simple/CriticalLiquidationTopBanner/CriticalLiquidationTopBanner.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  // distancePct is negative while approaching liquidation and >= 0 once the
+  // position is already liquidatable (same sign convention the dashboard gauge
+  // uses), so negate it before formatting the remaining buffer.
+  const message =
+    firstGroup.distancePct >= 0
+      ? COPY.topBanner.liquidatable
+      : COPY.topBanner.critical(
+          formatLiquidationDistancePercent(-firstGroup.distancePct),
+        );
```
jonybur, line 60: **Banner can display a reassuring "liquidation in 30%" while the position is actually critical, diverging from the mid-page banner.**

`usePositionNotifications` forces an `urgent` warning via its live-HF guardrail when the on-chain health factor ≤ 1.05 *even though the calculator's own distance check did not fire* — its documented purpose is exactly the case where stale/inflated indexed `totalBtc` made the cascade look safe (`usePositionNotifications.ts:128-148`). In that case `groups[0].distancePct` reflects the stale data (e.g. -30%), so this branch renders `Critical — liquidation in 30.0%` — self-contradictory and not "critical-looking."

The mid-page banner instead renders the warning's own `primaryWarning.title` (`Critical — health factor 1.03`, `PositionNotificationBanner.tsx:143`). So the two banners diverge precisely in the scenario the guardrail exists to catch: same severity source, but different *displayed value* source.

Suggest driving the displayed text from the warning that actually caused `red` (i.e. `bannerState.primaryWarning`) rather than from `groups[0].distancePct`.

services/vault/src/components/simple/CriticalLiquidationTopBanner/CriticalLiquidationTopBanner.tsx
```diff
@@ -0,0 +1,75 @@
...
+  // resolve it on mount so we can portal into it once available.
+  const [slot, setSlot] = useState<HTMLElement | null>(null);
+  useEffect(() => {
+    setSlot(document.getElementById(CRITICAL_BANNER_SLOT_ID));
+  }, []);
+
+  const bannerState = result ? deriveBannerState(result) : null;
+  const firstGroup = result?.groups[0] ?? null;
+
+  if (!slot || bannerState?.severity !== "red" || !firstGroup) {
```
jonybur, line 48: **Banner silently renders nothing when red severity is forced with empty groups.**

With invalid governance params the calculator suppresses the cascade and returns `groups: []` while emitting only a soft `weird-params` warning (`calculate.ts:131-138`). If the on-chain HF is also ≤ 1.05, the live-HF guardrail then forces an `urgent` warning onto that empty-groups result, so `deriveBannerState` returns `"red"` but `firstGroup` is `null`.

This guard (`!firstGroup`) therefore drops the banner entirely, while the mid-page banner still surfaces the warning via `primaryWarning`. For a warning the PR explicitly wants to be "impossible to miss," silently suppressing it is the wrong failure mode — consider falling back to the `liquidatable` copy when `firstGroup` is absent but severity is red.

(Same root fix as the `message` comment: when severity is red, derive text from the causing warning so the strip can never under-state risk or vanish.)

services/vault/src/components/simple/CriticalLiquidationTopBanner/CriticalLiquidationTopBanner.tsx
```diff
@@ -0,0 +1,75 @@
...
+    firstGroup.distancePct >= 0
+      ? COPY.topBanner.liquidatable
+      : COPY.topBanner.critical(
+          formatLiquidationDistancePercent(-firstGroup.distancePct),
+        );
+
+  return createPortal(
+    <TopBanner
+      visible
+      role="alert"
```
jonybur, line 65: **(minor / a11y)** `role="alert"` is an assertive, one-shot live region. Because the message text changes as the BTC price moves (`…in 4.3%` → `…in 4.2%`), screen readers re-interrupt with the whole string on every update. Since the banner is persistent and non-dismissible, `role="status"` / `aria-live="polite"` would announce the critical state without spamming on every refresh.

jonybur (commented):
Review of the critical near-liquidation top banner. The core-ui change, portal wiring, copy, and tests look clean. Two findings concern *what value the banner displays*, both rooted in the same design choice: the banner decides **whether** to show from `deriveBannerState(...).severity`, but derives **what to show** from `groups[0].distancePct` — and those two have different provenance. A third is a minor a11y note. See inline comments.

PR #1962, opened 2026-06-24
feat(packages): improves display of data on split vault form

jonybur:
Screenshot 2026-06-24 at 17 57 31

### DIFF
--- a/services/vault/src/components/simple/DepositForm.tsx
+++ b/services/vault/src/components/simple/DepositForm.tsx
@@ -11,6 +11,7 @@ import type { VaultProviderListItem } from "@/types/vaultProvider";
 import { CollateralFactorRow } from "./CollateralFactorRow";
 import { DepositFeesBreakdown } from "./DepositFeesBreakdown";
 import { FeesSection, type FeeRow } from "./FeesSection";
+import { SplitTooLowHint } from "./SplitTooLowHint";
 import { UtxoSplitSelector } from "./UtxoSplitSelector";
 import { VaultProviderSelector } from "./VaultProviderSelector";
 
@@ -28,6 +29,8 @@ interface PartialLiquidationProps {
   canSplit: boolean;
   isLoading: boolean;
   splitRatioLabel: string | null;
+  minDepositForSplit: bigint;
+  isSplitAmountTooLow: boolean;
 }
 
 export interface DepositAmountState {
@@ -393,6 +396,15 @@ export function DepositForm({
           btcPrice={btcPrice}
           hasPriceFetchError={hasPriceFetchError}
         />
+        {/* When the amount is too low to split, an inline hint sits inside the
+            amount card (below "Max to Borrow") telling the user the minimum
+            they need. The split selector below stays visible with its
+            two-vault option disabled. */}
+        {partialLiquidation?.isSplitAmountTooLow && (
+          <SplitTooLowHint
+            minDepositForSplit={partialLiquidation.minDepositForSplit}
+          />
+        )}
       </Card>
 
       {partialLiquidation && (
--- a/services/vault/src/components/simple/SplitTooLowHint.tsx
+++ b/services/vault/src/components/simple/SplitTooLowHint.tsx
@@ -19,7 +19,7 @@ export function SplitTooLowHint({ minDepositForSplit }: SplitTooLowHintProps) {
     <div
       role="status"
       aria-live="polite"
-      className="flex w-full items-center justify-center gap-2 rounded-lg border border-secondary-strokeLight px-[46px] py-2 text-center"
+      className="flex w-full items-center justify-center gap-2 rounded-lg border border-secondary-strokeLight px-3 py-2 text-center"
     >
       <IoInformationCircle
         size={18}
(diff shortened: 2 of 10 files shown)


services/vault/src/constants.ts
```diff
@@ -25,10 +25,10 @@ export const REPLAYS_ON_ERROR_RATE = Number.parseFloat(
...
 export const SUPPORT_URL = "https://discord.com/invite/babylonglobal";
 
-// Partial-liquidation (2-UTXO split) docs link, surfaced from the split-option
-// description in UtxoSplitSelector. TODO(product): swap in the exact
-// partial-liquidation doc page once confirmed — the docs root is a placeholder.
-export const PARTIAL_LIQUIDATION_DOCS_URL = "https://docs.babylonlabs.io";
+// Two-vault split docs link, surfaced from the split-option description in
+// UtxoSplitSelector. Points at the "decide how to split your BTC" step.
+export const PARTIAL_LIQUIDATION_DOCS_URL =
+  "https://docs.babylonlabs.io/trustless-bitcoin-vault/use-for-lending/create-a-vault/#step-1-decide-how-to-split-your-btc";
```
gbarkhatov, line 31: Two non-blocking notes on this constant:

1. **Identifier naming drift** — `PARTIAL_LIQUIDATION_DOCS_URL` (and the `partialLiquidation` prop threaded through the form) still uses the old 'partial liquidation' vocabulary while all user-facing copy is now 'two-vault split'. Fine if intentional, since partial liquidation is the underlying mechanism — just flagging the half-done rename at the identifier level.
2. **Brittle deep-link anchor** — `#step-1-decide-how-to-split-your-btc` will break if the docs page is restructured. Worth a quick manual click-through to confirm the anchor resolves, since it replaces the placeholder root.

services/vault/src/components/simple/DepositForm.tsx
```diff
@@ -28,6 +29,8 @@ interface PartialLiquidationProps {
   canSplit: boolean;
   isLoading: boolean;
   splitRatioLabel: string | null;
+  minDepositForSplit: bigint;
+  isSplitAmountTooLow: boolean;
```
gbarkhatov, line 33: `PartialLiquidationProps` here now carries `minDepositForSplit` / `isSplitAmountTooLow`, but the separate copy of this shape in `UtxoSplitSelector.tsx` (lines 7-13) was not updated. It compiles today (structural superset assigned to a subset), but the two definitions have now diverged. Consider exporting one shared type so they cannot drift apart.

services/vault/src/components/simple/SplitTooLowHint.tsx
```diff
@@ -0,0 +1,36 @@
...
+    >
+      <IoInformationCircle
+        size={18}
+        className="mt-px shrink-0 text-accent-primary"
+      />
+      <span className="min-w-0 text-sm text-accent-secondary">
+        {hint.prefix}{" "}
+        <span className="text-accent-primary">{hint.splitName}</span>
+        {hint.middle}{" "}
+        <span className="text-accent-primary">{hint.minimum}</span>
```
gbarkhatov, line 32: The hook logic behind this hint is well covered, but the visible message is reassembled here from `{prefix, splitName, middle, minimum}` fragments — a dropped space separator or a reordered span would be a visual bug no test catches. Low risk given how small it is; just noting the coverage gap.

services/vault/src/copy.ts
```diff
@@ -404,6 +407,27 @@ export const COPY = {
...
+      // below stays visible with its two-vault option disabled. `minBtc` already
+      // carries the network coin symbol (e.g. "0.4 BTC"). The split name and
+      // minimum are
+      // emphasized (primary text) by the component; the rest stays secondary.
+      // The component joins these fragments with explicit `{" "}` separators.
+      splitTooLowHint: (minBtc: string) => ({
+        prefix: "To use",
+        splitName: TWO_VAULT_SPLIT_NAME,
+        middle: ", increase your deposit to",
+        minimum: `at least ${minBtc}`,
```
gbarkhatov, line 429: 'To use Two-vault split, increase…' capitalizes the option name mid-sentence. It's deliberate (proper-noun treatment, matching the emphasized selector title), so it's internally consistent — just flagging in case product prefers lowercase mid-sentence.

gbarkhatov (approved):
Reviewed the split-vault display changes — clean, well-scoped, and conventions-compliant. All user-facing strings are routed through `COPY`, the shared `TWO_VAULT_SPLIT_NAME` constant keeps the split title and the too-low hint from drifting, the removed helpers (`getSplitOptionMainLabel`, `RECOMMENDED_SUFFIX`) leave no orphaned references, and `isSplitAmountTooLow` is complementary with `canSplit` and well covered by the new boundary tests. No critical-path code touched — display/copy only. A few non-blocking notes left inline. Approving.

jrwbabylonlab (approved):

services/vault/src/copy.ts
```diff
@@ -404,6 +407,27 @@ export const COPY = {
       providerMetricPlaceholder: "—",
       // Accessible label / tooltip for the per-provider explorer link.
       providerExplorerLinkLabel: "View vault provider on explorer",
+      // Split-option title. "Two-vault split" (not "UTXO split") because the
+      // UTXO concept is never introduced to the depositor; the ratio (e.g.
+      // "26/74") shows how the deposit is divided across the two BTC Vaults.
+      splitOptionLabel: (splitRatioLabel: string | null) =>
```
jrwbabylonlab, line 413: This title now says "Two-vault split" but the description right next to it still says "multiple vaults" — worth aligning the wording. (Also the "no amount" test in `useAllocationPlanning.test.ts` sets both amount and min to 0n, so it doesn't actually isolate the zero-amount guard it's named for.)

PR #1963, opened 2026-06-24
feat(packages): changes copy of hardware wallets

jonybur:
Screenshot 2026-06-24 at 16 45 12

### DIFF
--- a/packages/babylon-wallet-connector/src/components/WalletButton/index.tsx
+++ b/packages/babylon-wallet-connector/src/components/WalletButton/index.tsx
@@ -1,6 +1,13 @@
 import { Avatar, Chip, Text } from "@babylonlabs-io/core-ui";
 import { twMerge } from "tailwind-merge";
 
+// Green = software wallet detected/installed; yellow = hardware wallet that is
+// available but still needs connecting; secondary/gray = not installed.
+const INSTALLED_ACCENT = "text-[#15B768] dark:text-[#00E676]";
+const INSTALLED_DOT = "bg-[#15B768] dark:bg-[#00E676]";
+const HARDWARE_ACCENT = "text-[#FFB300]";
+const HARDWARE_DOT = "bg-[#FFB300]";
+
 interface WalletButtonProps {
   className?: string;
   logo: string;
@@ -8,6 +15,7 @@ interface WalletButtonProps {
   name: string;
   fallbackLink?: string;
   installed?: boolean;
+  label?: string;
   onClick?: () => void;
 }
 
@@ -18,8 +26,13 @@ export function WalletButton({
   logo,
   fallbackLink,
   installed = true,
+  label,
   onClick,
 }: WalletButtonProps) {
+  // Hardware wallets (and other connect-modal entries) carry a static
+  // descriptor label and are always reachable, so they show a yellow
+  // "available" badge instead of the detection-based Installed/Uninstalled one.
+  const hasDescriptor = Boolean(label);
   const btnProps = installed
     ? { as: "button", disabled, onClick }
     : { as: "a", href: fallbackLink, target: "_blank", rel: "noopener noreferrer" };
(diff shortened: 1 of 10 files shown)


packages/babylon-wallet-connector/src/components/WalletButton/index.tsx
```diff
@@ -18,8 +26,13 @@ export function WalletButton({
   logo,
   fallbackLink,
   installed = true,
+  label,
   onClick,
 }: WalletButtonProps) {
+  // Hardware wallets (and other connect-modal entries) carry a static
+  // descriptor label and are always reachable, so they show a yellow
+  // "available" badge instead of the detection-based Installed/Uninstalled one.
+  const hasDescriptor = Boolean(label);
```
gbarkhatov: Main concern: `hasDescriptor = Boolean(label)` drives the yellow 'available' badge (and, via `walletDisplayRank`, the sort tier) off any non-empty label — but `label` is not a hardware-only signal. `injectable` ('Injectable'), `appkit-btc` ('Connect BTC Wallet') and `appkit-eth` ('Connect ETH Wallet') all carry labels that were never displayed before this PR.

In the vault this is mostly masked because `injectable`/`appkit-btc`/`ledger*` are disabled on BTC (VaultWalletConnectionProvider `ALWAYS_DISABLED_WALLETS`), so only Keystone gets the badge — the intended effect. But `appkit-eth` is NOT disabled and is the sole entry in the ETH chain, so the 'Select Ethereum Wallet' button flips from green 'Installed' to yellow 'Connect ETH Wallet' — a visible change outside this PR's stated scope. Any other consumer of this shared package that does not disable `injectable`/`appkit-btc` will likewise see 'Injectable' / 'Connect BTC Wallet' leak as yellow badges.

Recommend gating on an explicit signal (e.g. `kind: 'hardware'` / `hardware: true` on the metadata) instead of 'has any label', or clearing the non-hardware labels. Worth confirming the ETH screen against the design.

packages/babylon-wallet-connector/src/components/WalletButton/index.tsx
```diff
@@ -1,13 +1,21 @@
 import { Avatar, Chip, Text } from "@babylonlabs-io/core-ui";
 import { twMerge } from "tailwind-merge";
 
+// Green = software wallet detected/installed; yellow = hardware wallet that is
+// available but still needs connecting; secondary/gray = not installed.
+const INSTALLED_ACCENT = "text-[#15B768] dark:text-[#00E676]";
+const INSTALLED_DOT = "bg-[#15B768] dark:bg-[#00E676]";
+const HARDWARE_ACCENT = "text-[#FFB300]";
```
gbarkhatov, line 8: Minor: `HARDWARE_ACCENT`/`HARDWARE_DOT` (`#FFB300`) have no `dark:` variant, unlike the `INSTALLED_*` pair just below. Confirm `#FFB300` has acceptable contrast on the dark chip background, or add a `dark:` override for parity. (Also: these remain raw hex rather than core-ui semantic tokens — I found no matching token, so this just extends the existing pattern; low priority.)

packages/babylon-wallet-connector/src/components/Wallets/index.tsx
```diff
@@ -18,13 +18,22 @@ export interface WalletsProps {
   onSelectWallet?: (chain: IChain, wallet: IWallet) => void;
 }
 
+// Connect-list ordering: installed software wallets (green) first, then
+// hardware wallets (yellow — available but need connecting), then uninstalled
+// download-only options (gray) at the bottom. A lower rank sorts higher, so
+// when no software wallet is installed the hardware options surface to the top.
+const walletDisplayRank = (wallet: IWallet): number => {
```
gbarkhatov, line 25: Minor: `walletDisplayRank` is a pure 3-way function with a non-obvious branch ('no software wallet installed -> hardware surfaces to the top') but no unit test pinning the ordering — a small test would lock in intent. It also shares the `Boolean(label)`-as-hardware assumption flagged on WalletButton: an installed wallet carrying a label (e.g. `injectable`) is reranked out of tier 0 into the hardware tier.

gbarkhatov (dismissed):
Reviewed the hardware-wallet copy/badge changes. The core change is sound and, in the vault, correctly scoped to Keystone on BTC; the constant extraction and the clearer `walletDisplayRank` are improvements. Approving — findings are non-blocking. One worth a look before merge: the badge/sort heuristic keys off `Boolean(label)`, which is broader than 'hardware wallet', most visibly flipping the ETH appkit entry to a yellow 'Connect ETH Wallet' badge (see inline). Drive-by nit, pre-existing and outside this diff: the Keystone story fixture has a typo `name: 'Keysone'` in packages/babylon-wallet-connector/src/components/Wallets/index.stories.tsx.

packages/babylon-wallet-connector/src/components/WalletButton/index.tsx
jrwbabylonlab: `Boolean(label)` is too broad — injectable/AppKit entries carry labels (`Injectable`, `Connect ETH/BTC Wallet`) that aren't hardware, so an installed injected wallet shows a yellow descriptor badge instead of green Installed and sorts into the hardware tier. Same hits BBN injectable in simple-staking. Can we gate on something hardware-specific rather than any non-empty label?

jrwbabylonlab (approved):

PR #1968, opened 2026-06-25
feat(vault): add operator-configurable notice banner

kirugan:


services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -127,6 +128,10 @@ export default function RootLayout() {
             here so it renders above the header, atop the operational banners. */}
         <div id={CRITICAL_BANNER_SLOT_ID} />
         <TestingBanner visible={shouldDisplayTestingMsg()} />
+        <NoticeBanner
+          visible={!isGeoBlocked && Boolean(FeatureFlags.noticeBannerMessage)}
+          message={FeatureFlags.noticeBannerMessage}
+        />
         <AddressScreeningBanner
           visible={!isGeoBlocked && isWalletConnected && isAddressBlocked}
         />
```
jrwbabylonlab, line 141: haha. we are stacking the banners now.
i think u should re-use the `DepositDisabledBanner`? or merge it together somehow?

services/vault/.env.example
```diff
@@ -55,6 +55,10 @@ NEXT_PUBLIC_REOWN_PROJECT_ID=your-reown-project-id-here
 # Authenticate the artifact stream with a gRPC-subject token (auth_createDepositorTokenGrpc).
 # Must match the VP proxy's ENABLE_GRPC_ARTIFACTS; leave off to use the JSON-RPC bearer.
 # NEXT_PUBLIC_FF_ENABLE_GRPC_ARTIFACTS=true
+# Freeform notice banner shown at the top of the app (e.g. for intermittent
+# peg-in errors or degraded service). Leave blank to hide. Independent of
+# NEXT_PUBLIC_FF_DISABLE_DEPOSIT — a notice can be shown while deposits stay on.
+# NEXT_PUBLIC_NOTICE_BANNER_MESSAGE=Peg-ins may be intermittently delayed. Retrying usually resolves it.
```
gbarkhatov, line 61: 🔴 This var also needs adding to `.github/workflows/service-release-vault.yml`, in the env block that maps GitHub `vars` next to the other `NEXT_PUBLIC_*` entries. Vite only inlines `NEXT_PUBLIC_*` vars the release job explicitly passes, so without it the banner can never appear in staging/testnet/prod — even after a redeploy. One line:

```yaml
NEXT_PUBLIC_NOTICE_BANNER_MESSAGE: ${{ vars.NEXT_PUBLIC_NOTICE_BANNER_MESSAGE }}
```

services/vault/src/components/shared/NoticeBanner.tsx
```diff
@@ -0,0 +1,26 @@
...
+  message: string;
+}
+
+export function NoticeBanner({ visible, message }: NoticeBannerProps) {
+  if (!visible || !message) {
+    return null;
+  }
+
+  return (
+    <div className="flex flex-row items-center justify-center gap-2 bg-amber-100 px-4 py-3 text-center text-amber-800 dark:bg-amber-900/30 dark:text-amber-300">
```
gbarkhatov, line 19: Nit (non-blocking): consider `role="status"` + `aria-live="polite"` on this container so screen readers announce the notice when it appears. The sibling banners lack it too, so not a regression — just a small a11y win for an operator message.

gbarkhatov (dismissed):
Reviewed the full diff plus the existing Greptile and @jrwbabylonlab threads. Clean, well-tested addition that follows the existing thin-banner pattern (`AddressTypeBanner`), no new deps, the message is escaped (no XSS), and `whitespace-pre-line` preserves operator formatting. Approving — with one must-fix for it to actually work once deployed, and a few endorsements of the existing notes.

**🔴 Must-fix before relying on it in deployed envs:** `NEXT_PUBLIC_NOTICE_BANNER_MESSAGE` isn't passed through `.github/workflows/service-release-vault.yml`, so the release build never receives it from GitHub `vars` and the banner can't appear in staging/testnet/prod regardless of config (details inline). This sits *under* Greptile's P1: even a redeploy won't surface it until the workflow line is added.

**Endorsing the existing threads:**
- Greptile P2 (geo-block) — agree. A global operator notice should show even to geo-blocked sessions; the top banner stack renders above the geo-block screen, so dropping `!isGeoBlocked` from *this* banner is safe (keep it on the wallet/user-specific ones).
- @jrwbabylonlab (banner stacking) — agree, and it grows with #1969 (a ProtocolPauseBanner). I'd not literally fold this into `DepositDisabledBanner` (different semantics — freeform notice vs deposit kill-switch), but a shared presentational top-banner (core-ui already has a `TopBanner` primitive) or one operational-banner slot would stop the proliferation. Worth settling alongside #1969 so we don't ship two overlapping operator-banner systems.
- Greptile P1 (build-time baking) — true, but consistent with every `NEXT_PUBLIC_*` flag incl. `DISABLE_DEPOSIT`; operators redeploy. Fine as-is; a runtime-config follow-up only matters if rapid mid-incident messaging is a hard requirement.

**Nit:** also document the var in `README.md`'s env section (currently only in `.env.example`).

Nice, focused PR — approving on the understanding the workflow wiring lands before anyone relies on it.

services/vault/src/components/shared/NoticeBanner.tsx
gbarkhatov: @kirugan might ignore this

kirugan: yeah I won't implement it

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -127,6 +128,10 @@ export default function RootLayout() {
             here so it renders above the header, atop the operational banners. */}
         <div id={CRITICAL_BANNER_SLOT_ID} />
         <TestingBanner visible={shouldDisplayTestingMsg()} />
+        <NoticeBanner
+          visible={!isGeoBlocked && Boolean(FeatureFlags.noticeBannerMessage)}
```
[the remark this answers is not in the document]
kirugan: good point, fixed

services/vault/.env.example
kirugan: thanks, added

services/vault/src/config/featureFlags.ts
```diff
@@ -108,6 +108,22 @@ export default {
...
+   * for situations like intermittent peg-in errors or service degradation.
+   * Why needed: Lets DevOps surface an ad-hoc notice via env var without a
+   * code change. Intentionally decoupled from DISABLE_DEPOSIT so a notice can
+   * be shown while deposits remain enabled.
+   * Default: empty (banner hidden unless a non-empty message is set).
+   *
+   * Non-boolean gating config, so it uses the NEXT_PUBLIC_ prefix without _FF_.
+   */
+  get noticeBannerMessage() {
+    return (process.env.NEXT_PUBLIC_NOTICE_BANNER_MESSAGE ?? "").trim();
```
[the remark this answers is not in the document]
kirugan: Acknowledged — this is intentional: NEXT_PUBLIC_NOTICE_BANNER_MESSAGE is build-time inlined like every other
  NEXT_PUBLIC_* flag here, and operators surface it by setting the var and redeploying (the release workflow now
  passes it through). A no-redeploy runtime config would be a separate architecture and should land as a follow-up
  covering all operational banners uniformly, not a one-off here.

gbarkhatov (approved):

services/vault/src/components/pages/RootLayout.tsx
kirugan: I'm leaving it as is, since there is no consensus now and Govard needs my PR for his to work

PR #1969, opened 2026-06-25
feat(vault): protocol soft/hard pause banners

### LINKED ISSUE #1953: Notification — Protocol soft-paused banner
## Summary
Show a banner when the protocol is soft-paused: deposits/borrows/withdrawals
disabled, repay still allowed, liquidations active. Not implemented.

## Current state
- Only a per-reserve `paused` flag exists in
  `applications/aave/services/fetchConfig.ts` (used to still surface debt).
- No protocol-wide soft-pause banner or copy.

## Design (Figma)
- Title: **Protocol is soft-paused**
- Body: *"New deposits, borrows, and withdrawals are disabled. You can still
  repay debt — liquidations remain active. Learn more"*

https://www.figma.com/design/7Nf1LhRG3ad3RYhMhgHiQE/TBV-x-Master-File--Updated-Flow-?node-id=6508-114956&t=U8J29QsulIdS2d7e-4

## Scope
1. Determine the protocol-wide soft-pause signal (extend `fetchConfig` if the
   contract exposes a global/soft-pause distinct from per-reserve `paused`).
2. Build the banner (style per Figma) with a "Learn more" link (confirm URL).
3. Add copy to `copy.ts`.
4. Gate affected actions (deposit/borrow/withdraw disabled; repay enabled) — or
   note if action-gating is out of scope and tracked separately.

## Acceptance criteria
- [ ] Banner shows when protocol is soft-paused
- [ ] "Learn more" links correctly; strings in `copy.ts`; lint + tests pass

Related: fully-paused banner (share the pause-state detection in `fetchConfig`)

### LINKED ISSUE #1954: Notification — Protocol fully-paused banner
## Summary
Show a banner when the protocol is fully paused: all operations disabled
(including liquidations), debt keeps accruing. Not implemented.

## Current state
- No protocol-wide pause signal beyond per-reserve `paused`
  (`applications/aave/services/fetchConfig.ts`); no banner or copy.

## Design (Figma)
- Title: **Protocol is fully paused**
- Body: *"All operations are disabled, including liquidations. Debt continues
  accruing interest. Monitor official announcements. Learn more"*

https://www.figma.com/design/7Nf1LhRG3ad3RYhMhgHiQE/TBV-x-Master-File--Updated-Flow-?node-id=6508-115016&t=U8J29QsulIdS2d7e-4

## Scope
1. Determine the full-pause signal (extend `fetchConfig`; distinguish from
   soft-pause).
2. Build the banner with a "Learn more" link (confirm URL).
3. Add copy to `copy.ts`.
4. Gate all actions off (or note action-gating scope).

## Acceptance criteria
- [ ] Banner shows when protocol is fully paused
- [ ] "Learn more" links correctly; strings in `copy.ts`; lint + tests pass

Related: soft-paused banner (share the pause-state detection in `fetchConfig`)

gbarkhatov:
# Protocol pause banners (soft + hard)

Screenshot_2026-06-25_15-01-57

Screenshot_2026-06-25_15-02-24

Screenshot_2026-06-25_15-02-37

Screenshot_2026-06-25_15-02-51

Screenshot_2026-06-25_15-03-02

Screenshot_2026-06-25_15-03-31

Screenshot_2026-06-25_15-03-43

Screenshot_2026-06-25_15-03-57

Screenshot_2026-06-25_15-04-09

## What this does

Adds two status banners that tell users when the protocol is paused, matching the Figma designs.

- Soft pause (teal): "Protocol is soft-paused — new deposits, borrows, and withdrawals are disabled. You can still repay debt — liquidations remain active."
- Full pause (red): "Protocol is fully paused — all operations are disabled, including liquidations. Debt continues accruing interest. Monitor official announcements."

Each banner shows a pause icon, a title, the message, and a "Learn more" link. DevOps turns them on with a feature flag, and can optionally override the message text with an env variable.

## Why

During the recent testnet incident the team paused deposits and wanted a clearer "we are paused" message for users (the old behavior was just a thin "deposits unavailable" strip). These two banners give users an accurate, readable picture of what is and isn't working during a pause. The key difference between the two is liquidations: in a soft pause liquidations still happen (so users can and should still act), while a full pause says everything is stopped.

## How it's controlled (DevOps)

- `NEXT_PUBLIC_FF_PROTOCOL_SOFT_PAUSED=true` → shows the teal soft-pause banner.
- `NEXT_PUBLIC_FF_PROTOCOL_FULLY_PAUSED=true` → shows the red full-pause banner. If both are set, full pause wins.
- `NEXT_PUBLIC_PAUSE_BANNER_MESSAGE="..."` → optional; replaces the banner body text. If empty, the default copy is used.

The banner only reflects the flags — it does not try to detect the on-chain pause state. This is on purpose: we cannot reliably read "the protocol is paused" from the contracts or the indexer today, and a banner that wrongly claims "liquidations are off" would be dangerous (a user could stop watching their position and get liquidated). So pausing is an explicit operator decision, the same way `DISABLE_DEPOSIT` already works.

## Approaches we considered

1. Detect the pause state automatically (from the Aave reserve flags or a contract read). Rejected: there is no protocol-wide pause signal available today, and guessing it from per-reserve flags could show the wrong thing — especially the "liquidations disabled" claim, which we cannot verify.
2. Reuse the existing `DISABLE_DEPOSIT` kill-switch as the trigger. Rejected: "deposits disabled" is not the same as "protocol soft/hard paused," and the two banners need different titles, colors, and copy that a single deposit flag can't express.
3. Dedicated feature flags per pause level, rendered through the shared core-ui Notification component, plus an optional text override. Chosen.

## Why we picked this approach

It is honest (the UI never claims something it can't verify), it matches how DevOps already operates during incidents (flip a flag), and it reuses the shared `Notification` severity component instead of building one-off markup. Soft vs hard is a small explicit switch, so the title and color are always correct for the message being shown.

## What changed

- core-ui: new `PauseIcon`, and a new `halted` Notification variant (red `error-light`) to sit next to the existing `paused` variant (teal `info-light`). The `paused` variant now defaults to the pause icon. Added Storybook stories for both.
- vault: new `ProtocolPauseBanner` component that renders the correct card based on the flags, with the message override and a "Learn more" link. It is mounted site-wide in `RootLayout` and uses the same content `Container` as the page, so it lines up with the dashboard column.
- vault: replaced the old thin `DepositDisabledBanner` (the `DISABLE_DEPOSIT` kill-switch still disables the deposit action and its in-page empty state — only its banner was replaced).
- copy: added the two default messages to `copy.ts`.
- config + CI: added the two flags and the message variable to `featureFlags.ts`, the release workflow (`service-release-vault.yml`), `.env.example`, and the README.

## DevOps follow-up (not in this repo)

The release workflow now passes `NEXT_PUBLIC_FF_PROTOCOL_SOFT_PAUSED`, `NEXT_PUBLIC_FF_PROTOCOL_FULLY_PAUSED`, and `NEXT_PUBLIC_PAUSE_BANNER_MESSAGE` from GitHub `vars`. The values of those `vars` still need to be created per environment (repo/environment settings or the deployment repo). Until they exist they default to empty, so the flags read `false` and the banners stay hidden — merging this changes nothing in production until DevOps flips a flag.

## Heads-up: overlap with https://github.com/babylonlabs-io/babylon-toolkit/pull/1968

PR https://github.com/babylonlabs-io/babylon-toolkit/pull/1968 adds a separate, generic operator notice strip (`NoticeBanner`, a freeform message shown at the top of the app, independent of any pause). It serves a different purpose than these pause banners, but the two PRs touch the same three files: `RootLayout.tsx`, `featureFlags.ts`, and `.env.example`.

If https://github.com/babylonlabs-io/babylon-toolkit/pull/1968 merges first, rebase this branch on `main` and resolve the conflicts by keeping both: keep the `NoticeBanner` in the top strip stack and our `ProtocolPauseBanner` in the content column, keep both feature-flag getters, and keep both `.env.example` entries (our removal of `DepositDisabledBanner` should stay). The banners can coexist — one is a generic notice, the other is the structured pause state.

Two small notes for whoever reconciles them: there is some redundancy between our `NEXT_PUBLIC_PAUSE_BANNER_MESSAGE` and #1968's `NEXT_PUBLIC_NOTICE_BANNER_MESSAGE` (both override banner text), so the team may want to keep only one later; and #1968 does not add its env var to the release workflow, so its banner won't be controllable in deployed environments until that line is added too.

## Testing

- New unit test for `ProtocolPauseBanner` (hidden when no flag, soft card, full card, full wins over soft, message override).
- core-ui and vault build pass; full vault test suite passes; lint is clean.
- Manually verified both banners in the dev server (soft, full, precedence, message override) and that the card aligns with the dashboard content column when connected and disconnected.

## Closes

Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/1953
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/1954


services/vault/src/config/featureFlags.ts
```diff
@@ -34,6 +34,52 @@ export default {
...
+   * Purpose: Surfaces the red "Protocol is fully paused" status banner (all
+   * operations disabled including liquidations; debt keeps accruing). Takes
+   * precedence over the soft-paused banner when both are set.
+   * Why needed: Same operator-controlled model as PROTOCOL_SOFT_PAUSED; DevOps
+   * escalates to this when the whole market is halted. The banner only reflects
+   * the flag — it does not detect the on-chain pause state.
+   * Default: false (banner hidden unless explicitly set to "true")
+   */
+  get isProtocolFullyPaused() {
+    return process.env.NEXT_PUBLIC_FF_PROTOCOL_FULLY_PAUSED === "true";
```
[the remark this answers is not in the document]
gbarkhatov: by design

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -187,6 +182,7 @@ export default function RootLayout() {
           <GeoBlockState />
         ) : (
           <ActivatingVaultsProvider>
```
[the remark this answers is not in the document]
gbarkhatov: done

services/vault/src/components/shared/ProtocolPauseBanner.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {
+  type ProtocolPauseLevel,
+  resolveProtocolPauseLevel,
+} from "@/components/shared/protocolPauseLevel";
+import featureFlags from "@/config/featureFlags";
+import { COPY } from "@/copy";
+
+// TODO(#1953): swap for the confirmed protocol-pause docs URL once product
+// provides it.
+const PROTOCOL_PAUSE_LEARN_MORE_URL = "https://docs.babylonlabs.io";
```
jonybur: `PROTOCOL_PAUSE_LEARN_MORE_URL` is a placeholder (`docs.babylonlabs.io`) with a TODO. Fine to merge since the banners are default-off, but during a real pause it'd be a poor moment for 'Learn more' to land on a generic docs root - worth making the real URL a release gate on enabling these flags (tracked on #1953) rather than an open-ended TODO.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -196,6 +203,7 @@ export default function RootLayout() {
...
+            <ProtocolPauseBanner />
```
jonybur: `ProtocolPauseBanner` renders inside the non-geo-blocked branch, so a geo-blocked session never sees a full protocol pause. That's asymmetric with `NoticeBanner`, which is intentionally placed in the top stack so geo-blocked users still get operator notices. If a full pause is a service-wide condition a geo-blocked user should also see, this arguably belongs in the same top stack; if the current placement is intentional, a one-line comment on why pause differs from notice would keep it from drifting.

services/vault/src/copy.ts
```diff
@@ -1037,6 +1037,21 @@ export const COPY = {
...
     },
   },
+  // Operator-controlled protocol-pause status banners. The body may be
+  // overridden per incident via NEXT_PUBLIC_PAUSE_BANNER_MESSAGE; these are the
+  // defaults. Each renders a trailing "Learn more" link.
+  protocolPause: {
+    soft: {
+      title: "Protocol is soft-paused",
+      body: "New deposits, borrows, and withdrawals are disabled. You can still repay debt — liquidations remain active.",
+      learnMore: "Learn more",
```
jonybur: `learnMore: "Learn more"` is duplicated verbatim across `soft` and `hard`. Minor, but a single shared constant referenced by both levels would match how the other shared strings are hoisted at the top of this file.

services/vault/src/components/shared/__tests__/ProtocolPauseBanner.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const link = screen.getByRole("link", { name: "Learn more" });
+    expect(link.getAttribute("href")).toMatch(/^https?:\/\//);
+  });
+
+  it("shows the fully-paused card when fully paused", () => {
+    featureFlagsMock.isProtocolFullyPaused = true;
+
+    render(<ProtocolPauseBanner />);
+
+    expect(screen.getByText("Protocol is fully paused")).toBeInTheDocument();
```
jonybur: Tests cover off/soft/hard/precedence/override well. The one behavioral distinction not asserted is the a11y role: `Notification` resolves `halted` to `role="alert"` (assertive) and soft `paused` to `role="status"`, which is the actual difference the new `halted` variant introduces. A `getByRole('alert')` assertion in this fully-paused test would lock that in.

jonybur (commented):
Clean and well-modeled - the pause level is a real discriminated union (`'soft' | 'hard' | null` with hard winning), copy lives in copy.ts, the operator override is rendered as React-escaped text (no XSS / no dangerouslySetInnerHTML), and the flags are `NEXT_PUBLIC_FF_*`, strictly parsed, default-off. A couple of things to confirm before these are enabled in an environment; rest is minor.

services/vault/src/components/shared/ProtocolPauseBanner.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+import { PAGE_CONTENT_CLASS } from "@/components/shared/layoutClasses";
+import {
+  type ProtocolPauseLevel,
+  resolveProtocolPauseLevel,
+} from "@/components/shared/protocolPauseLevel";
+import featureFlags from "@/config/featureFlags";
+import { COPY } from "@/copy";
+
+// TODO(#1953): swap for the confirmed protocol-pause docs URL once product
```
jrwbabylonlab: Let's drop the issue number from this TODO — we keep tracker refs out of code comments. `// TODO: swap for the confirmed protocol-pause docs URL once product provides it.` is fine.

services/vault/src/copy.ts
```diff
@@ -1037,6 +1037,21 @@ export const COPY = {
       tooltip: "Bonus percentage awarded to liquidators on seized collateral.",
     },
   },
+  // Operator-controlled protocol-pause status banners. The body may be
+  // overridden per incident via NEXT_PUBLIC_PAUSE_BANNER_MESSAGE; these are the
+  // defaults. Each renders a trailing "Learn more" link.
+  protocolPause: {
+    soft: {
+      title: "Protocol is soft-paused",
+      body: "New deposits, borrows, and withdrawals are disabled. You can still repay debt — liquidations remain active.",
```
jrwbabylonlab: We don't have a withdraw kill-switch in the FE, so "withdrawals are disabled" isn't actually enforced — a user can still withdraw while this banner says they can't. Can we reword to operator intent (e.g. "being paused — see announcements") or confirm the ops behavior backs this?

services/vault/src/config/featureFlags.ts
```diff
@@ -34,6 +34,52 @@ export default {
     return process.env.NEXT_PUBLIC_FF_DISABLE_BORROW === "true";
   },
 
+  /**
+   * PROTOCOL_SOFT_PAUSED feature flag
+   *
+   * Purpose: Surfaces the teal "Protocol is soft-paused" status banner (new
+   * deposits/borrows/withdrawals disabled, repay still allowed, liquidations
+   * active).
+   * Why needed: Pause state is operator-controlled, not exposed on-chain or by
```
jrwbabylonlab: Minor: pause state is exposed on-chain — `TBVPausableUpg.pauseState` is public. Suggest rewording to something like "we intentionally drive the banner from an operator flag rather than reading `pauseState`" instead of "not exposed on-chain".

services/vault/src/components/shared/ProtocolPauseBanner.tsx
gbarkhatov: Kept the placeholder — the real docs URL is a product dependency we don't have yet — but reframed the TODO so enabling the pause flags in an environment is gated on the real URL being set. Tracked as a release-gate item on the Freeze/Pause follow-ups (#1974 / #1975). Also dropped the issue ref from the comment per the convention.

services/vault/src/components/pages/RootLayout.tsx
gbarkhatov: Intentional — added a one-line comment explaining it: a geo-blocked session is already fully blocked from transacting (it sees the geo-block screen), so it doesn't need the pause card the way it still needs operator notices. The deeper "consolidate the operator banners" question is tracked separately.

services/vault/src/copy.ts
gbarkhatov: Done — hoisted to a shared `PROTOCOL_PAUSE_LEARN_MORE` constant at the top of copy.ts (matching the other hoisted shared strings), referenced by both `soft` and `hard`.

services/vault/src/components/shared/__tests__/ProtocolPauseBanner.test.tsx
gbarkhatov: Done — added `expect(screen.getByRole("alert")).toBeInTheDocument()` to the fully-paused test to lock in the `halted` → `role="alert"` distinction (vs the soft `paused` variant's `role="status"`).

services/vault/src/components/shared/ProtocolPauseBanner.tsx
gbarkhatov: Done — `// TODO:` now, no tracker ref.

services/vault/src/copy.ts
gbarkhatov: Fixed. This PR now actually disables new deposits and borrows when paused (the pause state is wired into the deposit/borrow kill-switches via `isDepositBlocked()` / `isBorrowBlocked()`), and the copy was trimmed to only claim what's enforced ("New deposits and borrows are disabled. You can still repay debt — liquidations remain active."). There's no withdraw kill-switch in the FE yet, so withdrawal enforcement is tracked as the remaining Freeze work in #1974.

services/vault/src/config/featureFlags.ts
gbarkhatov: Good catch — reworded. The JSDoc no longer says "not exposed on-chain"; it now states we intentionally drive the banner from an operator flag rather than reading `TBVPausableUpg.pauseState`. Whether to actually read `pauseState` (auto-detect vs operator flag) is the open detection decision we're raising with DeGen/Aleksa and noted on #1974 / #1975.

gbarkhatov:
@greptile review

services/vault/src/config/featureFlags.ts
```diff
@@ -34,6 +34,55 @@ export default {
...
+   * deposit + borrow enforcement with the freeze flag (gating the remaining Aave
+   * actions is the Pause follow-up). Takes precedence over the frozen banner
+   * when both are set. Pause is the full stop (a last-resort emergency).
+   * Why needed: Same operator-controlled model as PROTOCOL_FROZEN; DevOps
+   * escalates to this when the whole market is halted.
+   * Default: false (banner hidden unless explicitly set to "true")
+   */
+  get isProtocolPaused() {
+    return process.env.NEXT_PUBLIC_FF_PROTOCOL_PAUSED === "true";
+  },
```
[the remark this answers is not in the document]
gbarkhatov: Correct, and intentional for this PR. #1969's scope is the status banner plus *entry* gating — deposit and borrow — which is the full enforcement set shared by both Frozen and Paused (both block new entry). The additional actions Pause stops (withdraw, repay, activation, redemption, reorder, vault-swap, fee claims) are tracked in #1975, deliberately split out because Pause gating blocks *exits* and time-critical activation while having to keep recovery paths open (expired-pegin cleanup, fee refunds, BTC-side refund/WOTS). Wiring that without trapping funds is the substance of #1975, not a one-line guard. The PAUSED flag is operator-controlled and won't be enabled in any environment until that enforcement lands.

jeremy-babylonlabs (approved):

Issue #1958, opened 2026-06-24
Notification — Too many vaults for this position (max vaults reached)


gbarkhatov:
## Summary
Warn the user when a position has reached the maximum number of BTC Vaults and new
vaults can't be added until one is closed/removed. Not implemented. Distinct from
#1952 — that is the optimizer's ordering-cap fallback (position still works); this is a
hard cap that blocks creating another vault.

## Current state
- No `"max-vaults"` warning type:
  `WarningType = "urgent" | "dust" | "weird-params"`
  (`applications/aave/positionNotifications/types.ts`).
- No "Maximum vaults reached" copy in `copy.ts`.
- No guard in the deposit / split flow blocking an additional vault at the cap
  (`hooks/deposit/useOptimalSplit.ts`; ts-sdk `integrations/aave/utils/vaultSplit.ts`).
- A `MAX_DP_N = 10` constant exists in ts-sdk
  (`integrations/aave/utils/optimalOrder.ts`), but it is the DP-optimizer size limit,
  not a user-facing per-position cap — confirm the actual cap source/value.
- Current vault count is available via `collateralVaults.length`
  (`hooks/useDashboardState.ts`).

## Design (Figma)
- Severity: **warning** (orange accent)
- Title: **Maximum vaults reached**
- Body: *"This position already has the maximum number of BTC Vaults (10)."*
- Annotation: *"This position has reached the maximum number of BTC Vaults (10). New
  vaults cannot be created until an existing vault is closed or removed."*
- Also shown inline in the deposit modal as the user approaches the cap (e.g. at 9 vaults).

https://www.figma.com/design/7Nf1LhRG3ad3RYhMhgHiQE/TBV-x-Master-File--Updated-Flow-?node-id=8162-41260

## Scope
1. Confirm the cap source + value (Figma shows 10; reconcile with `MAX_DP_N` / any
   protocol limit) and extract to a named constant.
2. Add a `"max-vaults"` warning/notification variant; surface via `bannerSeverity` /
   the notification list as a warning Notification.
3. Add copy (title + detail, interpolating the cap) to `copy.ts`.
4. Gate the deposit / add-vault flow at the cap (block creating a new vault; surface
   the inline deposit-modal message).

## Acceptance criteria
- [ ] "Maximum vaults reached" notification shows when the position is at the vault cap
- [ ] Deposit / add-vault blocked at the cap; copy interpolates the cap and matches Figma
[...]

Issue #1950, opened 2026-06-24
Notification — Reorder to lose less (title + suggested-order chips)


jeremy-babylonlabs:
## Summary
The reorder suggestion + "Apply Suggested Order" action are implemented, but the
title doesn't match Figma and the SUGGESTED ORDER chip visualization is missing.

## Current state (partial)
- Logic + button done: `PositionNotificationBanner` / `BannerActions`,
  `applyReorderedOrder`, copy `COPY.banner.applySuggestedOrder`.
- Order data available: `result.suggestedVaultOrder`.
- Current title: `COPY.liquidationWarnings.reorder.title` =
  "BTC Vaults aren't in the safest liquidation order".

## Design (Figma)
- Title: **Reorder vaults to lose less**
- **SUGGESTED ORDER** chip row: `Vault 3 · 0.60 BTC → Vault 1 · 0.10 BTC →
  Vault 2 · 0.30 BTC`
- Button: **Apply Suggested Order**

https://www.figma.com/design/7Nf1LhRG3ad3RYhMhgHiQE/TBV-x-Master-File--Updated-Flow-?node-id=6502-111184&t=U8J29QsulIdS2d7e-4

## Scope
1. Update `COPY.liquidationWarnings.reorder.title` to "Reorder vaults to lose
   less" (confirm detail copy too).
2. Add the suggested-order chip row, rendered from `result.suggestedVaultOrder`
   (vault name + BTC amount, arrow separators).
3. Reuse existing apply action.

## Acceptance criteria
- [ ] Title matches Figma
- [ ] Chip row renders the suggested order with amounts and arrows
- [ ] Apply action unchanged; strings in `copy.ts`; lint + tests pass

Issue #1949, opened 2026-06-24
Notification — First liquidation suggestion (withdraw / re-deposit)


jeremy-babylonlabs:
## Summary
For a single large vault, suggest enabling partial liquidation by withdrawing and
re-depositing as two smaller vaults (sacrificial + protected) with computed
amounts. Not a defined warning type today.

## Current state
- `WarningType = "urgent" | "dust" | "weird-params"`
  (`applications/aave/positionNotifications/types.ts`) — no variant for this.
- Generic copy exists (`splitOptionDescription`: "Split your Bitcoin into
  multiple vaults…") but not the computed-amount suggestion.
- Split sizing available from `useOptimalSplit`.

## Design (Figma)
- Title: **First liquidation takes everything**
- SUGGESTION box: *"To enable partial liquidation, withdraw your 0.70 BTC and
  re-deposit as two smaller vaults: 0.40 BTC sacrificial + 0.30 BTC protected.
  Alternatively: add collateral or repay debt…"*

https://www.figma.com/design/7Nf1LhRG3ad3RYhMhgHiQE/TBV-x-Master-File--Updated-Flow-?node-id=7064-77201&t=U8J29QsulIdS2d7e-4

## Scope
1. Add a new warning/notification variant for the single-big-vault case.
2. Compute the withdraw amount + split sizes from `useOptimalSplit`.
3. Add copy (interpolating the three amounts) to `copy.ts`.
4. Wire `bannerSeverity` / notification surfacing + any CTAs.

## Acceptance criteria
- [ ] Suggestion renders with computed withdraw + sacrificial + protected amounts
- [ ] New variant integrated into severity/notification logic
- [ ] Strings in `copy.ts`; lint + tests pass

Related: sacrificial-vault card (shared `useOptimalSplit` data)

gbarkhatov:
**Status update (post #1970): calculator logic shipped, Figma alignment still pending**

The liquidation-calculator port (#1970) added the single-vault `cliff` warning behind `NEXT_PUBLIC_FF_ENABLE_LIQUIDATION_NOTIFICATIONS`. When a sacrificial vault is affordable it offers an add-vault CTA (the #1948 path); when it is **not** affordable it falls back to a generic re-create / repay suggestion — conceptually this CLIFF B case. What it does **not** match is the Figma (node `7064-77201`):

| Aspect | Shipped (#1970) | Figma (CLIFF B) |
|---|---|---|
[...]

Issue #1948, opened 2026-06-24
Notification — Sacrificial vault card


jeremy-babylonlabs:
Relates to #1949 

## Summary
When a single liquidation event would seize all of a user's BTC, surface a card
recommending a small "sacrificial" vault as a buffer. Card UI does not exist;
the sizing logic does.

## Current state
- Split/sizing logic exists: `applications/aave/hooks/useOptimalSplit.ts`
  (returns sacrificial + protected sizes); margins in `aave/constants.ts`;
  related concept in `components/simple/ReorderVaults/constants.ts`.
- No notification card UI; copy not in `copy.ts`.

## Design (Figma)
- Title: **First liquidation takes everything**
- Body: *"With your current vaults, a single liquidation event seizes all your
  BTC — nothing remains protected behind it."*
- Info box: *"Adding a 0.14 BTC sacrificial vault creates a buffer — it gets
  liquidated first, your existing BTC survives."*
- CTA: **Add sacrificial vault**

## Scope
1. Build the card UI (matches Figma callout/info-box styling).
2. Source the 0.14 BTC sizing from `useOptimalSplit` (sacrificial size).
3. Wire **Add sacrificial vault** CTA to the split/deposit flow.
4. Add all copy to `copy.ts`.
5. Define the trigger condition (single-vault → first event seizes everything)
   and where it surfaces in `bannerSeverity` / notification list.

## Acceptance criteria
- [ ] Card renders with computed sacrificial size when a single event seizes all BTC
- [ ] CTA opens the split flow pre-sized to the sacrificial amount
- [ ] Strings in `copy.ts`; lint + tests pass

Related: first-liquidation suggestion (withdraw/re-deposit) variant

gbarkhatov:
**Status update (post #1970): calculator logic shipped, Figma alignment still pending**

The liquidation-calculator port (#1970) added a working `cliff` notification behind `NEXT_PUBLIC_FF_ENABLE_LIQUIDATION_NOTIFICATIONS`. For this CLIFF A "add a sacrificial vault" case, the calculator now emits a `cliff` warning with an actionable **"Add a {X} BTC vault"** CTA that pre-fills the deposit. So the build exists — what remains is matching the Figma (node `6502-110902`):

| Aspect | Shipped (#1970) | Figma (CLIFF A) |
|---|---|---|
[...]

PR #1970, opened 2026-06-25
feat(vault): port reference liquidation calculator

### LINKED ISSUE #1952: Notification — Too many vaults to optimize (17-cap fallback)
## Summary
Beyond the optimizer's vault cap (17), warn the user that the best liquidation
order can't be guaranteed and suggest consolidating. Not implemented.

## Current state
- Optimizer logic in `applications/aave/positionNotifications/calculate.ts`.
- No warning type or copy for the 17-vault cap; no card.

## Design (Figma)
- Title: **Too many vaults to optimize**
- Body: *"You have 18 vaults. Beyond 17, the optimizer can't guarantee the best
  liquidation order…"*
- Suggestion: *"Consider consolidating smaller vaults…"*

https://www.figma.com/design/7Nf1LhRG3ad3RYhMhgHiQE/TBV-x-Master-File--Updated-Flow-?node-id=7048-61969&t=U8J29QsulIdS2d7e-4

## Scope
1. Detect the cap breach in `calculate.ts` (confirm the cap constant — extract to
   a named constant if not already).
2. Add a new warning/notification variant for the cap fallback.
3. Add copy (interpolating the vault count + cap) to `copy.ts`.
4. Surface via `bannerSeverity` / notification list.

## Acceptance criteria
- [ ] Warning appears when vault count exceeds the optimizer cap
- [ ] Copy interpolates actual count + cap; matches Figma

gbarkhatov:
# Port the liquidation calculator to match the source-of-truth repo

Screenshot_2026-06-25_18-48-13

## What this does

It brings our in-app liquidation calculator in line with the canonical calculator in the `tbv-liquidations` repo (the agreed source of truth, recently rewritten by @ngfam), so the dApp produces the full set of partial-liquidation notifications instead of just a few.

Before this PR our calculator only produced three warnings: `urgent` (close to liquidation), `dust` (position too small), and `weird-params` (invalid protocol params), plus a basic reorder suggestion. The reference calculator produces a much richer, actionable set. This PR adds the missing pieces.

New notifications:
- `cliff` — all your vaults would be seized in a single liquidation event (no partial-liquidation protection). For a single-vault position it tells you the exact size of a sacrificial vault to add so your current vault becomes protected.
- `rebalance` — your first liquidation group over-seizes (vault sizes aren't ideal); suggests adding a correctly-sized vault.
- `reorder` — a safer vault order exists (now a first-class warning with the reference's swap / reduce-first-event / deepen-cascade variants, plus a guard that stops a reorder ↔ rebalance flip-flop).
- `too-many-vaults` — past the optimizer's cap the ordering is no longer guaranteed optimal; suggests consolidating.

New actions on the banner:
- `cliff` and `rebalance` now show an "Add a X BTC vault" button that opens the deposit flow with the suggested amount pre-filled (a single, non-split deposit). This reuses plumbing that already existed.
- The reorder "Apply Optimal Order" button is unchanged.

## Why

We originally split a user's BTC into a sacrificial + protected vault at deposit time so that one vault can be liquidated first while the other survives. That setup can break later (protocol params change, the user adds/removes a vault), and then the user can lose everything in one event. The fix the team agreed on is not to silently reorder for the user, but to *show actionable notifications* and let the user decide what to do. The reference calculator is where that logic now lives, so our job is to mirror it faithfully.

## Approaches considered

1. Incremental alignment — keep our current structure and bolt on the missing warnings one at a time.
2. Wholesale port — replace our advisory layer with a faithful port of the reference calculator and adopt its scenario suite as golden-vector tests.

We went with the wholesale port. The reference is the agreed source of truth and is what the test scenarios were validated against, so mirroring it 1:1 keeps us from drifting and makes future re-syncs trivial. We adapted it to our codebase only where we had to: we keep our strings in `copy.ts`, we keep our existing protective gate that hides the cascade under invalid protocol params, and we reuse the shared optimizer in the ts-sdk rather than duplicating it.

The core liquidation math (seizure fraction, liquidation price, over-seizure, fairness payments, dust) was already identical to the reference, so it is unchanged — this PR only adds the advisory/notification layer on top.

## Key decisions a reviewer should sanity-check

- Optimizer cap raised from 10 to 17 with a largest-first fallback past the cap, matching the reference (we dropped the old "pre-joint smaller vaults" approach). We kept our existing first-event tiebreaker. The audited-path characterization by vault count: ≤10 vaults — unchanged from what ships today; 11–17 vaults — now an **exact** DP (the old code compressed n>10 into 10 pre-joint groups and ran an approximate DP), so for these sizes the suggested order genuinely *can* differ; 18+ vaults — now largest-first (was the pre-joint heuristic). This is safe regardless of cap: the on-chain reorder check (`assertOptimalOrderMatchesOnChain`) re-runs the same `calculate()` that produced the CTA, so the CTA and the guard stay self-consistent. That guard's tests still pass.
- `expectedHF` stays 0.95. The project spec doc says 0.97, but the reference *code* defaults to 0.95 (its own UI even says "Hardcoded 0.95 in production"); the doc (confluence) is stale.
- `urgent` copy stays as our shipped wording ("Liquidation is X% away" / "Liquidation can trigger now") rather than the reference's "Critical…" strings. Tests assert behavior, not the reference's urgent text.
- `cliff` / `rebalance` / `too-many-vaults` render as soft (info) banners, not red. Red is reserved for imminent-liquidation (`urgent`). This means a healthy single-vault position shows a soft "No backup vault" advisory. Worth confirming against the cliff Figma severity.
- The whole surface stays behind the existing `NEXT_PUBLIC_FF_ENABLE_LIQUIDATION_NOTIFICATIONS` flag.

## What is NOT in this PR

This PR is scoped to the liquidation calculator. It does not implement the separate hard per-position vault cap from https://github.com/babylonlabs-io/babylon-toolkit/issues/1958 ("Too many vaults for this position") — that one is a contract-read cap (`maxVaultsPerPosition`) that blocks creating a new vault in the deposit flow, which is a different, deposit-flow feature.

## Testing

- ts-sdk: optimizer + cascade tests updated for the new cap/fallback; full aave suite passes (148 tests).
- vault: typecheck clean, production build clean, lint clean (0 errors). Added 14 golden-vector tests ported from the reference scenario suite (single-vault cliff, two-vault cliff, reorder/swap, rebalance, optimal 3-vault, tolerance boundary, too-many-vaults) — the suggested vault amounts matched the reference exactly. Added banner tests for the new "Add a X BTC vault" action.
- The on-chain reorder verification path (`assertOptimalOrderMatchesOnChain`) was re-checked — all 22 of its tests pass after the optimizer change.

## Related issues

- Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/1952 (too many vaults to optimize — the 17-cap optimizer ordering warning).
- Relates to https://github.com/babylonlabs-io/babylon-toolkit/issues/1948 and https://github.com/babylonlabs-io/babylon-toolkit/issues/1949 (the cliff "add a sacrificial vault" notifications).
- Builds on https://github.com/babylonlabs-io/babylon-toolkit/issues/1950 (reorder to lose less — already shipped, now upgraded).
- Out of scope: https://github.com/babylonlabs-io/babylon-toolkit/issues/1958 (hard per-position vault cap).

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/optimalOrder.ts
+++ b/packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/optimalOrder.ts
@@ -168,8 +168,15 @@ export function computeOptimalOrder<T extends CascadeVault>(
   const order = groupsReversed.flat();
 
   // Reconstruction safety: if the chain is incomplete for any reason, fall back
-  // to largest-first rather than throwing.
+  // to largest-first rather than throwing (calculate() runs in the notification
+  // render and must not crash). This path is a DP-invariant violation that
+  // should never occur for valid input — log it so a regression is observable
+  // even though the fallback keeps the UI alive and the result well-shaped.
   if (order.length !== n) {
+    console.error(
+      `computeOptimalOrder: DP reconstruction produced ${order.length}/${n} vaults; ` +
+        `falling back to largest-first. This indicates a DP-invariant regression.`,
+    );
     const fallback = [...vaults].sort((a, b) => b.btc - a.btc);
     const sim = simulateCascade(
       fallback,
--- a/packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/optimalOrder.ts
+++ b/packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/optimalOrder.ts
@@ -4,74 +4,29 @@
  * Finds the vault ordering that maximizes collateral surviving a multi-group
  * liquidation cascade.
  *
- * The optimizer is a bitmask DP over seized subsets. To keep 2^n memory and
- * 3^n work bounded, vault sets larger than MAX_DP_N are first compressed into
- * pre-joint groups (the smallest vaults merged together), so the DP always runs
- * on ≤ MAX_DP_N states and never throws regardless of vault count.
+ * The optimizer is a bitmask DP over seized subsets. 2^n memory and 3^n work
+ * blow up past `MAX_DP_N` vaults, so for larger sets the optimizer falls back to
+ * a largest-first heuristic — the suggested order is then no longer guaranteed
+ * optimal, and `calculate()` surfaces a `too-many-vaults` warning to say so.
  */
 
 import {
   simulateCascade,
   type CascadeVault,
-  type PreJointVaults,
 } from "./cascadeSimulation.js";
 
 /**
- * Hard cap on the number of states for the bitmask DP optimizer.
- * 2^n memory + 3^n work blow up past this, so larger vault sets are first
- * compressed into pre-joint groups by merging the smallest vaults together.
+ * Hard cap on vault count for the bitmask DP optimizer. 2^n memory + 3^n work
+ * blow up past this. For n > MAX_DP_N the optimizer falls back to a
+ * largest-first heuristic. Benchmark: n=18 ≈ 720ms, n=20 ≈ 5.8s — anything past
+ * n=17 is too slow for interactive UI, so we cap here.
  */
-export const MAX_DP_N = 10;
-
-/**
- * Reduce a vault list to at most `maxSize` pre-joint groups for DP optimization.
- *
- * Starts with one group per vault, then repeatedly merges the two smallest
- * groups until the group count fits the DP cap. Groups are kept sorted by
- * descending BTC sum, and vaults inside each group are sorted by descending BTC
- * before returning.
- */
-function groupByPreJoints<T extends CascadeVault>(
-  vaults: T[],
-  maxSize: number,
-): PreJointVaults<T>[] {
-  const preJoints: PreJointVaults<T>[] = vaults.map((v) => ({
-    vaults: [v],
-    sum: v.btc,
-  }));
-
-  preJoints.sort((a, b) => b.sum - a.sum);
-
-  while (preJoints.length > maxSize) {
-    const first = preJoints.pop()!;
-    const second = preJoints.pop()!;
-    const combined: PreJointVaults<T> = {
-      vaults: [...first.vaults, ...second.vaults],
-      sum: first.sum + second.sum,
-    };
-
-    let inserted = false;
-    for (let i = 0; i < preJoints.length; i++) {
-      if (combined.sum > preJoints[i].sum) {
-        preJoints.splice(i, 0, combined);
-        inserted = true;
-        break;
-      }
-    }
-    if (!inserted) preJoints.push(combined);
-  }
-
-  for (const pj of preJoints) {
-    pj.vaults.sort((a, b) => b.btc - a.btc);
-  }
-
-  return preJoints;
-}
+export const MAX_DP_N = 17;
 
 /**
  * Main optimizer: bitmask DP over seized subsets.
  *
- * State: T = bitmask of pre-joint groups that have already been seized.
+ * State: T = bitmask of vaults that have already been seized.
  * Transition: for each valid "last group" G ⊆ T, dp[T] = dp[T\G] + btcAfter
  *   where btcAfter = totalBtc − btcOf(T)   (BTC remaining after T is seized).
  * Validation: btcOf(G) must cover target seizure at the moment G fires, i.e.
@@ -94,9 +49,9 @@ export function computeOptimalOrder<T extends CascadeVault>(
   maxLB: number,
   expectedHF: number,
 ): { order: T[]; sumBtcAfterEvents: number; btcAfterG1: number } {
-  if (vaults.length === 0)
-    return { order: [], sumBtcAfterEvents: 0, btcAfterG1: 0 };
-  if (vaults.length === 1) {
+  const n = vaults.length;
+  if (n === 0) return { order: [], sumBtcAfterEvents: 0, btcAfterG1: 0 };
+  if (n === 1) {
     const sim = simulateCascade(
       vaults,
       totalDebt,
@@ -110,11 +65,28 @@ export function computeOptimalOrder<T extends CascadeVault>(
     return { order: [...vaults], ...sim };
   }
 
-  const preJoints = groupByPreJoints(vaults, MAX_DP_N);
-  const n = preJoints.length;
+  // Safety cap: 2^n memory and 3^n work blow up past MAX_DP_N vaults. Real
+  // users have far fewer — each vault is a separate peg-in with a fixed fee.
+  // For the unlikely n > MAX_DP_N, fall back to a largest-first heuristic;
+  // calculate() surfaces a 'too-many-vaults' warning to flag the loss of the
+  // optimality guarantee.
+  if (n > MAX_DP_N) {
+    const order = [...vaults].sort((a, b) => b.btc - a.btc);
+    const sim = simulateCascade(
+      order,
+      totalDebt,
+      seizedFraction,
+      seizureTol,
+      CF,
+      THF,
+      maxLB,
+      expectedHF,
+    );
+    return { order, ...sim };
+  }
 
   const N = 1 << n;
-  const totalBtc = preJoints.reduce((s, v) => s + v.sum, 0);
+  const totalBtc = vaults.reduce((s, v) => s + v.btc, 0);
   const coverFraction = seizedFraction * (1 - seizureTol);
 
   // Precompute btcOf[T] — standard bitmask sum trick in O(2^n).
@@ -122,7 +94,7 @@ export function computeOptimalOrder<T extends CascadeVault>(
   for (let T = 1; T < N; T++) {
     const lsb = T & -T;
     const bit = 31 - Math.clz32(lsb);
-    btcOf[T] = btcOf[T ^ lsb] + preJoints[bit].sum;
+    btcOf[T] = btcOf[T ^ lsb] + vaults[bit].btc;
   }
 
   // dpSum[T] = best sumBtcAfterEvents to reach state T (T = seized subset).
@@ -183,10 +155,10 @@ export function computeOptimalOrder<T extends CascadeVault>(
   const groupsReversed: T[][] = [];
   for (let T = fullMask; T > 0; ) {
     const G = prev[T];
-    if (G === -1) break; // unreachable chain; length assertion below will fail
+    if (G === -1) break; // unreachable chain — fall back below
     const gVaults: T[] = [];
     for (let bit = 0; bit < n; bit++) {
-      if (G & (1 << bit)) gVaults.push(...preJoints[bit].vaults);
+      if (G & (1 << bit)) gVaults.push(vaults[bit]);
     }
     gVaults.sort((a, b) => b.btc - a.btc); // canonical within-group order
     groupsReversed.push(gVaults);
@@ -195,10 +167,21 @@ export function computeOptimalOrder<T extends CascadeVault>(
   groupsReversed.reverse();
   const order = groupsReversed.flat();
 
-  if (order.length !== vaults.length) {
-    throw new Error(
-      `DP reconstruction failed — expected ${vaults.length} vaults in order, got ${order.length}.`,
+  // Reconstruction safety: if the chain is incomplete for any reason, fall back
+  // to largest-first rather than throwing.
+  if (order.length !== n) {
+    const fallback = [...vaults].sort((a, b) => b.btc - a.btc);
+    const sim = simulateCascade(
+      fallback,
+      totalDebt,
+      seizedFraction,
+      seizureTol,
+      CF,
+      THF,
+      maxLB,
+      expectedHF,
     );
+    return { order: fallback, ...sim };
   }
 
   // Compute real metrics via cascade simulation (debt-aware, matches what
--- a/services/vault/src/applications/aave/positionNotifications/__tests__/calculate.test.ts
+++ b/services/vault/src/applications/aave/positionNotifications/__tests__/calculate.test.ts
@@ -369,4 +369,24 @@ describe("golden vectors (reference scenario suite)", () => {
     );
     expect(result.optimalVaultOrder).toBeNull();
   });
+
+  it("17-vault rebalance: uses the deterministic manual order past the optimizer cap", () => {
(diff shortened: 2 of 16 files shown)


services/vault/src/applications/aave/positionNotifications/calculate.ts
```diff
@@ -289,14 +313,120 @@ export function calculate(params: CalculatorParams): CalculatorResult {
...
+
+    const optTarget = totalBtc * seizedFraction;
+    const optOver = Math.max(0, optimalG1Btc - optTarget);
+    const optProtected = totalBtc - optimalG1Btc;
+    const optIsCliff = optimalG1Vaults.length === nVaults;
+    const optWouldTriggerRebalance = optOver > optProtected && !optIsCliff;
+
+    if (!currentHasRebalanceCond && optWouldTriggerRebalance) {
+      reorderWouldHelp = false;
+    }
```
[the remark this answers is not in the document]
gbarkhatov: This is intentional and ported verbatim from the reference

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/optimalOrder.ts
```diff
@@ -195,10 +167,21 @@ export function computeOptimalOrder<T extends CascadeVault>(
...
+      totalDebt,
+      seizedFraction,
+      seizureTol,
+      CF,
+      THF,
+      maxLB,
+      expectedHF,
     );
+    return { order: fallback, ...sim };
   }
```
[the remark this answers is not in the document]
gbarkhatov: done

services/vault/src/applications/aave/positionNotifications/calculate.ts
```diff
@@ -319,14 +449,265 @@ export function calculate(params: CalculatorParams): CalculatorResult {
...
+      ).reduce((s, v) => s + v.btc, 0);
+      const optProtected = allTotal - optG1Btc;
+      const optOver = optG1Btc - allTotal * seizedFraction;
+      if (optOver > optProtected) {
+        const sortedSmall = [...smallVaults].sort((a, b) => b.btc - a.btc);
+        suggestedRebalanceOrder = [placeholder, ...sortedSmall, largest];
+      } else {
+        suggestedRebalanceOrder = optOrder;
+      }
+    }
```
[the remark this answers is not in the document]
gbarkhatov: done

services/vault/src/components/simple/PositionNotificationBanner/BannerActions.tsx
```diff
@@ -55,14 +58,32 @@ export function buildBannerActions({
...
+  const suggestedVaultBtc =
+    result.suggestedNewVaultBtc ?? result.suggestedRebalanceVaultBtc;
+  if (isCliffOrRebalancePrimary && suggestedVaultBtc !== null) {
+    const amountBtc = suggestedVaultBtc.toFixed(2);
+    actions.push({
+      label: COPY.banner.addVault(amountBtc),
+      onClick: () => onDeposit(amountBtc),
+      emphasis: "primary",
+    });
+  }
```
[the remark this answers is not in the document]
gbarkhatov: done

services/vault/src/applications/aave/positionNotifications/calculate.ts
```diff
@@ -319,14 +423,242 @@ export function calculate(params: CalculatorParams): CalculatorResult {
...
+    : totalBtc;
+  const rebalanceImprovementBtc = rebalanceNeeded
+    ? Math.max(0, idealProtectedBtc - currentProtectedBtc)
+    : 0;
+
+  // Suggested new sacrificial vault: s >= (T × liqFactor − smallVaultsSum) /
+  // (1 − liqFactor). Cheaper than a standalone vault because the existing small
+  // vaults contribute to covering the target seizure.
+  let suggestedRebalanceVaultBtc: number | null = null;
+  let suggestedRebalanceOrder: Vault[] | null = null;
```
jonybur: `suggestedRebalanceOrder` is computed here (including the extra full `computeOptimalOrder` pass below) and returned on `CalculatorResult`, but I can't find a production consumer - it's only asserted in tests. Same for `recommendedSacrificialBtc`, the returned `isFullLiquidation`, and the returned `rebalanceImprovementBtc` (its local is used for the detail string, but the field itself isn't read). The actual add-vault amount in BannerActions comes from `suggestedNewVaultBtc`/`suggestedRebalanceVaultBtc`, so these four look redundant. Per the zero-dead-code policy, either wire them into the UI or drop them - the `suggestedRebalanceOrder` one is the most wasteful since it runs a second optimizer pass purely to populate an unread field.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/optimalOrder.ts
```diff
@@ -4,74 +4,29 @@
...
-    if (!inserted) preJoints.push(combined);
-  }
-
-  for (const pj of preJoints) {
-    pj.vaults.sort((a, b) => b.btc - a.btc);
-  }
-
-  return preJoints;
-}
+export const MAX_DP_N = 17;
```
jonybur, line 24: At n=17 the bitmask DP is ~O(3^17) and runs roughly 100-200ms (the new 17-vault test carries a 20s timeout). `calculate()` is memoized at the call site and a real 17-separate-peg-in position is extraordinarily rare, so this is acceptable. Worth noting it's paid a second time synchronously on the signing path - `assertOptimalOrderMatchesOnChain` re-runs the full `calculate()` before the wallet prompt - so either confirm that path can absorb a worst-case ~100-200ms or drop a comment that it's off the blocking path.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/optimalOrder.ts
```diff
@@ -195,10 +167,28 @@ export function computeOptimalOrder<T extends CascadeVault>(
...
-  if (order.length !== vaults.length) {
-    throw new Error(
-      `DP reconstruction failed — expected ${vaults.length} vaults in order, got ${order.length}.`,
+  // Reconstruction safety: if the chain is incomplete for any reason, fall back
+  // to largest-first rather than throwing (calculate() runs in the notification
+  // render and must not crash). This path is a DP-invariant violation that
+  // should never occur for valid input — log it so a regression is observable
+  // even though the fallback keeps the UI alive and the result well-shaped.
+  if (order.length !== n) {
+    console.error(
```
jonybur, line 176: Falling back to largest-first rather than throwing here is reasonable (calculate() runs in the notification render and must not crash). One thought: this `console.error` is the only signal of a DP-invariant regression, and the fallback result is shape-identical to a normal too-many-vaults result, so a regression would be effectively silent in production. Routing this through the app logger instead of `console.error` would make it observable in telemetry.

services/vault/src/applications/aave/positionNotifications/__tests__/calculate.test.ts
```diff
@@ -194,14 +194,22 @@ describe("calculate", () => {
     expect(result.optimalVaultOrder ?? vaults).toHaveLength(14);
```
jonybur, line 194: This test is named for handling 'more vaults than the optimizer DP cap', but it uses 14 vaults and the cap is now 17 - so 14 runs the exact DP and no longer exercises the past-cap fallback (the new 18-vault test is what covers that). Worth bumping this above 17 or relabeling it so the name matches what it actually verifies.

services/vault/src/copy.ts
```diff
@@ -1063,13 +1066,81 @@ export const COPY = {
...
       detail:
         "A different vault order makes the first liquidation event smaller — less BTC seized when it triggers.",
       suggestedOrderLabel: "Suggested order",
       vaultChip: (name: string, amount: string) => `${name} · ${amount}`,
     },
+    // Cliff: all vaults consolidate into one liquidation group, so partial
+    // liquidation is no longer possible. Variant by vault count.
+    cliff: {
+      single: {
+        title: "No backup vault",
```
jonybur, line 1083: These liquidation strings use bare 'vault' throughout ('No backup vault', later 'Too many vaults to optimize', 'Add a X BTC vault'), whereas copy.ts's own style note calls for 'BTC Vault'. It's consistent with the existing `liquidationWarnings` copy and is a faithful port of the reference, so this is probably an intentional exception for the liquidation surface - just flagging so the copy owner can confirm it's intentional rather than drift.

jonybur (commented):
Solid, well-tested port. I reproduced the A1/B2 seizure vectors (1.0 -> 0.72, [0.8,0.2] -> 0.38) and confirmed `cascadeSimulation.ts` is byte-identical except the removed `PreJointVaults` type, so the core liquidation math is genuinely untouched - the suggested-vault sizing also rounds in the safe direction (`Math.ceil`).

One correction on the description, though: the claim that the suggested order is "identical to what ships today" for positions up to 17 vaults only holds for n <= 10. The old code compressed any n > 10 into 10 pre-joint groups and ran an *approximate* DP; the new code runs an *exact* DP for all n in [2, 17], so for 11-17 vaults the order genuinely can differ from what ships today. This is safe - `assertOptimalOrderMatchesOnChain` re-runs the same `calculate()` that produced the CTA, so the CTA and the on-chain guard stay self-consistent regardless of cap - but the audited-path characterization should read: <=10 unchanged, 11-17 now exact (was a pre-joint heuristic), 18+ now largest-first. Inline notes below.

jrwbabylonlab (dismissed):

services/vault/src/applications/aave/positionNotifications/calculate.ts
```diff
@@ -319,14 +423,242 @@ export function calculate(params: CalculatorParams): CalculatorResult {
...
+      const manualOrder = [placeholder, ...sortedSmall, largest];
+
+      if (allVaults.length > MAX_DP_N) {
+        // Adding the placeholder crosses the optimizer cap, so computeOptimalOrder
+        // would only return a largest-first heuristic — not a guaranteed optimum.
+        // Use the manual protective order directly rather than presenting the
+        // heuristic as if it were optimizer-backed.
+        suggestedRebalanceOrder = manualOrder;
+      } else {
+        const { order: optOrder } = computeOptimalOrder(
```
jrwbabylonlab: This second `computeOptimalOrder` feeds `suggestedRebalanceOrder`, which nothing consumes (only its null-ness is read for the suggestion text) — same for `recommendedSacrificialBtc` and the result-level `isFullLiquidation`. Can we drop the three unused fields and this extra call, or note them as intentional parity-with-reference fields?

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/optimalOrder.ts
gbarkhatov: Added a comment at the `calculate()` re-run in `assertReorderMatchesOnChain.ts` noting the worst-case ~100-200ms exact DP at the 17-vault cap is acceptable there: it is a one-shot guard fired immediately before a wallet prompt (seconds-scale), on an extraordinarily rare 17-separate-peg-in position, and off any hot render path.

services/vault/src/applications/aave/positionNotifications/calculate.ts
gbarkhatov: Done. Dropped `recommendedSacrificialBtc`, the result-level `isFullLiquidation`, `suggestedRebalanceOrder`, and the returned `rebalanceImprovementBtc` from `CalculatorResult` (the `rebalanceImprovementBtc` *local* stays — it feeds the rebalance detail string). That let me delete the second `computeOptimalOrder` pass entirely; the rebalance block now just computes the suggested vault amount, and the redundant `suggestedRebalanceVaultBtc !== null && suggestedRebalanceOrder !== null` guard collapses to the first half. `optimalVaultOrder` (the only output the signing path consumes) is untouched — it comes from the first optimizer pass. Typecheck + suites green.

services/vault/src/applications/aave/positionNotifications/__tests__/calculate.test.ts
gbarkhatov: Renamed to "handles a large in-cap position (14 vaults, exact DP) without throwing" — it was misnamed (14 < 17 runs the exact DP, not the fallback). The 18-vault test is the one that actually exercises the past-cap fallback.

services/vault/src/applications/aave/positionNotifications/calculate.ts
gbarkhatov: Done — dropped all three unused fields (`suggestedRebalanceOrder`, `recommendedSacrificialBtc`, result-level `isFullLiquidation`) plus the returned `rebalanceImprovementBtc`, and removed the extra `computeOptimalOrder` call. The rebalance block now only computes the suggested vault amount; the suggestion text keys off `suggestedRebalanceVaultBtc` alone. No parity-with-reference fields kept around unread.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/optimalOrder.ts
gbarkhatov: Keeping `console.error` here. `optimalOrder.ts` lives in the SDK package, which intentionally has no dependency on the vault app logger. `calculate()` already refuses to trust a past-cap result via `optimizerReliable = nVaults <= MAX_DP_N`, and this branch is a should-never-happen invariant check; since `assertOptimalOrderMatchesOnChain` re-runs the same `calculate()`, the CTA and the on-chain guard stay self-consistent even if it ever fired. Can revisit with a returned `isExact` flag if we want it in telemetry later.

services/vault/src/copy.ts
gbarkhatov: Confirmed intentional. The whole `liquidationWarnings` surface uses bare "vault" — it is a faithful port of the reference and reads more naturally in these sentences (e.g. "All vaults seized in one liquidation event"). Leaving it as the documented exception for this surface rather than introducing drift between the new and existing liquidation copy.

jeremy-babylonlabs (approved):

PR #1973, opened 2026-06-25
feat(vault): detect locked BTC wallet and prompt to unlock

jonybur:
Screenshot 2026-06-26 at 17 41 33
Screenshot 2026-06-26 at 17 02 43
Screenshot 2026-06-26 at 17 40 20
Screenshot 2026-06-26 at 17 40 38


packages/babylon-wallet-connector/src/core/wallets/btc/okx/provider.ts
```diff
@@ -165,6 +165,30 @@ export class OKXProvider implements IBTCProvider {
     return this.walletInfo.publicKeyHex;
   };
 
+  getAccounts = async (): Promise<string[]> => {
```
jrwbabylonlab: OKX's BTC `getAccounts()` returns the cached selected address here — the extension only clears it on disconnect, not on lock (`wallet_unlockStateChanged` just flips an internal `isUnlocked` flag). So a locked OKX still returns `[address]`, `areBtcAccountsLocked` stays false, and the banner never fires. With `MIN_OKX_VERSION` 4.5.0 that's every OKX user. Can we gate OKX out of the lock poll until we wire a real signal (`isUnlock()` / the LOCKED event) instead of relying on empty-array?

packages/babylon-wallet-connector/src/core/wallets/btc/onekey/provider.ts
```diff
@@ -210,6 +210,30 @@ export class OneKeyProvider implements IBTCProvider {
     return this.walletInfo.publicKeyHex;
   };
 
+  getAccounts = async (): Promise<string[]> => {
```
jrwbabylonlab: Same concern as OKX here: the wrapped `btcwallet.getAccounts()` returns dApp-connected accounts (its `getProviderState` hardcodes `isUnlocked: true`), so it reflects authorization, not keyring lock — a locked-but-authorized OneKey likely returns its cached address and never trips the banner. Can we confirm against a real locked OneKey before relying on empty-array, and gate it out otherwise?

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -295,6 +346,79 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
     enabled: Boolean(btcWalletProvider && address),
   });
 
+  // Detect a *silent* wallet auto-lock. Injected extensions (notably UniSat)
+  // re-lock on an idle timer and emit no event, while the cached `getAddress()`
+  // keeps returning a stale address — so `connected` stays true and the UI
+  // looks fine until the user hits a signing call. `getAccounts()` is the
+  // non-interactive probe (it never prompts) and returns [] when locked.
+  const checkLock = useCallback(async () => {
```
jrwbabylonlab, line 412: `checkLock` captures provider/address, awaits `getAccounts()`, then unconditionally `setLocked()`. If the user disconnects or switches wallets while the request is in flight, a stale response can flip `locked = true` after disconnect already cleared it — showing the unlock banner for a wallet that's no longer current. Worth a generation/cancellation guard before `setLocked`.

javi0guerrero4-cloud (approved):

services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx
```diff
@@ -242,6 +251,12 @@ export function DepositProgressView(props: DepositProgressViewProps) {
         // Callouts live here (not in the scrollable body) so error/success
         // banners stay pinned above the CTA, always visible.
         <div className="flex flex-col gap-4">
+          {walletLocked && (
```
gbarkhatov: **Minor (UX) — worth addressing before merge.** This lock `Callout` renders unconditionally, including when `isComplete` (success `Callout` at line 266) or `isTerminalSuccess` (line 268). If UniSat auto-locks right after a finished deposit, the user sees a red "unlock to continue" banner stacked next to the green success message. Suggest gating it:

```tsx
{walletLocked && !isComplete && !isTerminalSuccess && (
  <Callout variant="error" title={COPY.wallet.locked.title}>...
)}
```

services/vault/src/components/Wallet/Connect.tsx
```diff
@@ -82,6 +106,20 @@ export const Connect: React.FC<ConnectProps> = ({ loading = false, text }) => {
     [selectedWallets, btcConnected, ethConnected, btcConnector, ethConnector],
   );
 
+  // A silently locked BTC wallet keeps `connected` true (cached session), so it
+  // would otherwise render the connected wallet menu. Surface an unlock button
+  // in the navbar instead so the user can re-authorize in one click.
+  if (btcLocked && !isGeoBlocked && !isGeoLoading) {
```
gbarkhatov, line 96: **UX / product call.** When `btcLocked`, this returns early and replaces `BtcEthWalletMenu` entirely with the unlock button. A user with both BTC + ETH connected loses access to disconnect/switch and the ETH affordances until they unlock — and someone who intentionally de-authorized the dApp (which also reads as locked, see the `checkBTCConnection` probe in `BTCWalletProvider`) has no navbar path to fully disconnect. Probably acceptable since a locked BTC wallet blocks the core flow anyway, but confirm it's the intended tradeoff.

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -254,11 +284,61 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
...
+      if (accounts !== undefined) {
+        // The read is async — drop it if the wallet was disconnected or
+        // switched while it was in flight, so a stale response can't write
+        // `locked` for a session that is no longer current.
+        const probeContext = lockProbeContextRef.current;
+        if (probeContext.provider !== btcWalletProvider || probeContext.address !== address) {
+          return;
+        }
+        const probe = classifyBtcAccountsProbe(accounts, address);
+        if (probe === "locked") {
```
gbarkhatov, line 353: **Low — product confirm.** Behavior change: previously an empty/failed accounts read fell through to the `disconnect()` path below. Now, for UniSat, `getAccounts() === []` sets `locked` and returns — but `[]` covers both a silent auto-lock and a genuine de-authorization (user revoked the site's permission). The de-auth case is now surfaced as "locked" and the session is never torn down. Recovery via reconnect still works, but the two states are indistinguishable from this one signal. Acknowledged in the comments — flagging for an explicit product decision.

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -295,6 +375,95 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
...
+    if (probeContext.provider !== probedProvider || probeContext.address !== probedAddress) {
+      return;
+    }
+
+    const probe = classifyBtcAccountsProbe(accounts, probedAddress);
+    // A different active account is an account switch, not a lock: leave the
+    // cached address/pubkey refresh to the accountsChanged event and the
+    // visibility check. Clearing `locked` here would present an unlocked
+    // session against a stale address.
+    if (probe === "changed") return;
```
gbarkhatov, line 447: **Low (self-healing).** If the wallet is `locked` and the user unlocks into a *different* account than the cached one, this returns `"changed"` early without clearing `locked`, and `onAccountsChanged` (≈ lines 243-255) refreshes the address but also never calls `setLocked(false)`. So `locked` stays true until the next poll tick re-reads the now-current address as `"current"` (up to one interval). Adding `setLocked(false)` in the `newAddress !== address` branch of `onAccountsChanged` would clear it immediately and make intent explicit — an `accountsChanged` event is itself proof the wallet answered.

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -295,6 +375,95 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
...
+    if (!btcWalletProvider || !address) return;
+    if (typeof window === "undefined" || typeof document === "undefined") return;
+    if (typeof btcWalletProvider.getAccounts !== "function") return;
+
+    let intervalId: ReturnType<typeof setInterval> | undefined;
+
+    const startPolling = () => {
+      if (intervalId !== undefined) return;
+      void checkLock();
+      intervalId = setInterval(() => void checkLock(), BTC_WALLET_LOCK_POLL_INTERVAL_MS);
```
gbarkhatov, line 465: **Nit (polling).** The interval (`BTC_WALLET_LOCK_POLL_INTERVAL_MS = 10s`) equals UniSat's `getAccounts` timeout (`UNISAT_RPC_TIMEOUT_MS = 10s`), and `setInterval` doesn't await — so a probe that stalls near the timeout can overlap the next tick. The `lockProbeContextRef` guard makes this harmless, but an in-flight flag (or `interval > timeout`) would avoid concurrent probes. Separately, on tab return up to three probes fire near-simultaneously: this immediate `checkLock()`, the `focus` handler's `checkLock()`, and `checkBTCConnection`'s pre-check. All non-network, but consider guarding `handleFocus` on `document.visibilityState === "visible"` (or dropping it, since `visibilitychange` already covers the common case).

services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx
```diff
@@ -166,6 +170,11 @@ export function DepositProgressView(props: DepositProgressViewProps) {
     onSign,
   } = props;
 
+  // Every flow that renders this view requires the BTC wallet, so surface a
+  // silent lock here (as an error-style Callout) regardless of which flow —
+  // deposit, resume/broadcast, payout signing, or continuation — mounted it.
+  const { locked: walletLocked } = useBTCWallet();
```
gbarkhatov, line 177: **Verify (not blocking).** This comment claims every BTC-signing flow renders this view. Please confirm vault activation — the HTLC-secret submission in `vaultActivationService` (a critical signing path) — actually mounts inside `DepositProgressView` or otherwise reaches a lock gate. The click-time liveness probe is a backstop, so this is a coverage thoroughness check rather than a correctness gap.

services/vault/src/components/simple/DepositSignContent.tsx
```diff
@@ -130,6 +130,7 @@ export function DepositSignContent({
         </button>
       </div>
     );
+
```
gbarkhatov: **Nit.** Stray blank line added with no functional change — drop it.

gbarkhatov (dismissed):
Approving — a solid, defensive, well-documented change that adds silent-lock detection and gates the BTC-signing CTAs without touching any critical value-movement path. The async-race guards (`lockProbeContextRef`), the feature-detection discipline, and the OKX/OneKey omission rationale are all correct, and the helper unit tests cover the malformed-shape edge cases.

Left a handful of inline notes, all optional / non-blocking. The one most worth addressing before merge is the lock `Callout` rendering on completed/terminal deposit screens (you'd get a red "unlock to continue" banner stacked next to the green success message). The navbar-hides-ETH-menu and disconnect-reclassified-as-locked items each deserve a quick product call on the locked-vs-disconnected UX; the rest are nits.

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -254,11 +284,61 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
     };
   }, [btcWalletProvider, address, callbacks, disconnect]);
 
+  // Keep the lock-probe context in sync so an in-flight `getAccounts` resolution
+  // can detect that the session changed underneath it (see checkLock).
+  useEffect(() => {
+    lockProbeContextRef.current = { provider: btcWalletProvider, address };
```
jrwbabylonlab, line 313: This ref is synced in an effect (async), but `disconnect()` clears provider/address synchronously — so an in-flight `getAccounts()` can resolve in the gap, pass the stale check, and `setLocked(true)` on a dead session. `Connect.tsx:112` then shows "Unlock" for a disconnected wallet. Update the ref synchronously in `disconnect()`/connect, or capture a session id in the probe.

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
jrwbabylonlab, line 447: On `"changed"` this returns without touching `locked`. If it was locked and the user unlocks + switches account, steady polling leaves the banner stuck (no visibility event to clear it). Clear/refresh here instead of early-returning.

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -295,6 +375,95 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
     enabled: Boolean(btcWalletProvider && address),
   });
 
+  // Detect a *silent* wallet auto-lock. Injected extensions (notably UniSat)
+  // re-lock on an idle timer and emit no event, while the cached `getAddress()`
```
jrwbabylonlab, line 402: This "emit no event on lock" premise only holds for UniSat ≥ 1.7.14 (fixed 2025-12-13). Older builds emit `accountsChanged([])` on lock, which the unchanged `onAccountsChanged` handler (~L221) turns into a disconnect — so lock detection silently depends on `MIN_UNISAT_VERSION`. Worth a one-line note so lowering the gate later doesn't break it.

packages/babylon-wallet-connector/src/core/wallets/btc/okx/provider.ts
```diff
@@ -165,6 +165,13 @@ export class OKXProvider implements IBTCProvider {
     return this.walletInfo.publicKeyHex;
   };
 
+  // `getAccounts` is intentionally omitted: OKX's BTC provider returns the
```
jrwbabylonlab: OKX is closed-source and its docs don't specify locked behavior, so "returns the cached address even when locked" isn't verifiable. The omission is the right fail-safe regardless — reword to not assert closed-source internals.

jrwbabylonlab (commented):
Verified against UniSat source that the lock premise holds for the supported range: UniSat ≥ 1.7.14 (the connect floor) no longer emits `accountsChanged([])` on lock (fixed 2025-12-13, shipped in the 1.7.14 tag), so polling `getAccounts` is the correct signal and the banner is reachable. Solid overall, and no irreversible-value risk — the click-time liveness probe still gates signing. A few inline notes; the main one is a stale-probe write after disconnect. Separately, the provider-side lock engine (stale-guard, poll lifecycle, reconnect-clears-lock) is essentially untested — worth adding given there's no feature flag.

jrwbabylonlab (approved):

PR #1977, opened 2026-06-26
feat(vault): align cliff liquidation notification to Figma and smooth the stale price banner

### LINKED ISSUE #1948: Notification — Sacrificial vault card

### LINKED ISSUE #1949: Notification — First liquidation suggestion (withdraw / re-deposit)

gbarkhatov:
# Cliff liquidation notification (Figma alignment) + stale-price banner UX

Screenshot_2026-06-26_18-03-41

Screenshot_2026-06-26_18-03-54

Screenshot_2026-06-26_18-05-23

Screenshot_2026-06-26_18-05-39

## Summary

This PR does two related things on the dashboard's position-notifications surface. First, it makes the "first liquidation takes everything" (cliff) notification match the Figma design — the wording, the orange severity, the two variants, and the card layout. Second, it stops the "Position notifications temporarily unavailable" (stale BTC price) banner from flashing for brief hiccups, and makes it recover faster. The notification math itself was already shipped earlier in https://github.com/babylonlabs-io/babylon-toolkit/pull/1970 using developer placeholder copy; this PR is the design alignment plus a UX fix.

---

### What it looked like before

The cliff notification was working but used the reference repo's developer wording ("No backup vault" / "Both vaults seized together…"), rendered grey/soft, split its message by how many vaults you had (1 / 2 / 3+), and laid the action button out to the right of the text.

### What it looks like now

One title and body for every cliff case — "First liquidation takes everything" — rendered in the orange warning style. The message is now split by what you can actually do about it, which is what the two Figma frames show: if adding a small "sacrificial" vault is affordable, you get that suggestion plus an "Add sacrificial vault" button; if it isn't (your one vault is too big), you instead get guidance to withdraw and re-deposit as two smaller vaults, with the three amounts spelled out (withdraw / sacrificial / protected). The card is also laid out top-to-bottom per Figma: no icon chip next to the title, the button stacked below the box, an info "i" icon inside the affordable suggestion box, and a "SUGGESTION" label inside the unaffordable one.

### Approaches considered, and why we chose this one

- **Split the message by vault count (what existed) vs. by what the user can do (chosen).** Figma only draws two cliff frames, and they differ by the available action, not by a vault count. We re-mapped the existing logic onto those two action-based variants so the UI matches the design and reads as advice, not as an internal classification.
- **Ship [#1949](https://github.com/babylonlabs-io/babylon-toolkit/issues/1949) alone vs. ship [#1948](https://github.com/babylonlabs-io/babylon-toolkit/issues/1948) and [#1949](https://github.com/babylonlabs-io/babylon-toolkit/issues/1949) together (chosen).** They are two render variants of the same warning — same title, same body, same severity, emitted from the same place in the code. Doing them separately would mean editing the exact same lines twice, so they ship as one change.
- **Put the amount on the button vs. in the suggestion text (chosen).** Figma uses a plain "Add sacrificial vault" button with the amount in the sentence above it. We matched Figma; the button still pre-fills the deposit with the computed amount when clicked.
- **Change the shared component vs. keep it app-side (chosen).** The new layout (button below, no title icon, an icon/label inside the suggestion box) is all achievable with options the shared Notification component already exposes, plus passing our own content into the suggestion box. So we changed nothing in the shared component — which means zero risk to every other notification that uses it.
- **Multi-vault cliffs (2 and 3+ vaults).** Figma only specifies the single-vault story. For genuine multi-vault cliffs we apply the shared shell (same orange title and body) but keep their existing, accurate guidance text, because "withdraw and re-deposit as two smaller vaults" is wrong when you already hold several vaults.

Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/1948
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/1949

---

### The problem

Sometimes the dashboard shows "Position notifications temporarily unavailable — BTC price data is stale or unavailable," and refreshing the page doesn't help. The reason is that "stale" is decided by the on-chain Chainlink oracle's own last-updated timestamp (older than 1 hour), not by our cache being old. We already auto-refetch the price every 60 seconds, but re-reading a stale feed — or reloading the page — returns the same old timestamp, so the banner can't be refreshed away. It's mostly a devnet/testnet artifact, because that price feed updates infrequently; on mainnet BTC updates every few minutes.

### What we changed

- **Don't flash the banner for brief blips.** The live banner now only appears after the stale/failed condition has persisted for a grace window (60 seconds). A one-off failed read, or the short gap before a fresh poll lands, clears within that window and never shows the banner. During the grace window we show nothing — and importantly, we never feed a stale price into the liquidation math.
- **Recover faster.** While the price is stale or failing, we poll every 15 seconds instead of 60, and we also re-check when you switch back to the tab or your network reconnects. So once the feed (or the connection) recovers, the banner clears in seconds instead of up to a minute.

### Approaches considered, and why we chose this one

- **"Just refetch harder" — rejected as a fix on its own.** We already poll automatically; the staleness is on the oracle's side, so no amount of refetching makes a genuinely stale feed fresh. We kept faster polling only as a way to recover sooner once the feed actually updates.
- **Loosen the 1-hour staleness threshold — deliberately not done.** Using an older price in the liquidation calculation is exactly what that guard prevents, so we left it untouched. This change only affects when the UI banner appears, never whether stale data is used.
- **Show the banner only after N retries (chosen, as a time-based grace window).** This is what the grace window is: ride out transient problems, and only inform the user once the condition is real and sustained.

This part is a UX fix with no tracking issue; it was added on this branch because it touches the same banner.

---

## Testing

- New and updated unit tests: the cliff copy/severity and the two variants, the affordable vs. unaffordable suggestion box (info icon vs. "SUGGESTION" label), the grace-window hook (timer and reset behavior), the price-health helper, and the banner deferring the live stale-price banner until the window elapses.
- Full vault test suite passes (868 tests), typecheck is clean, and lint reports 0 errors.
- Manually verified each cliff variant and the stale-price behavior via the dashboard debug panel.

### DIFF
--- a/services/vault/src/applications/aave/positionNotifications/calculate.ts
+++ b/services/vault/src/applications/aave/positionNotifications/calculate.ts
@@ -418,9 +418,12 @@ export function calculate(params: CalculatorParams): CalculatorResult {
     detail: reorder.detail,
   };
 
-  // CASE 1: Single vault — always fully seized. To protect it, add a NEW
-  // sacrificial vault at position 1. Adding a vault increases total BTC, hence
-  // target seizure: s >= existingBtc × liqFactor / (1 − liqFactor).
+  // CASE 1: Single vault — always fully seized. Two Figma variants share the
+  // title/body and differ only by which fix is feasible:
+  //  • Affordable add (CLIFF A, #1948): a sacrificial vault smaller than the
+  //    position buffers it. s >= existingBtc × liqFactor / (1 − liqFactor).
+  //  • Oversized (CLIFF B, #1949): that add would exceed the position, so the
+  //    fix is to withdraw and re-deposit the same BTC as two smaller vaults.
   let suggestedNewVaultBtc: number | null = null;
   if (nVaults === 1) {
     const canSplit = liqFactor < 1;
@@ -435,25 +438,37 @@ export function calculate(params: CalculatorParams): CalculatorResult {
 
     let suggestion: string;
     if (suggestedNewVaultBtc !== null) {
-      suggestion = cliff.single.actionableSuggestion(
-        btc2(suggestedNewVaultBtc),
-        btc2(vaults[0].btc),
-      );
-    } else if (!canSplit) {
-      suggestion = cliff.single.noSplitSuggestion;
+      // Variant A — affordable sacrificial add; the CTA carries the action.
+      suggestion = cliff.addSacrificialSuggestion(btc2(suggestedNewVaultBtc));
     } else {
-      suggestion = cliff.single.oversizedSuggestion(btc2(raw));
+      // Variant B — re-split the existing vault. seizedFraction depends only on
+      // CF/maxLB/THF, so re-splitting the same total is valid; size the
+      // sacrificial to cover the seizure first and protect the remainder.
+      const withdrawBtc = vaults[0].btc;
+      const sacrificialBtc = Math.ceil(withdrawBtc * liqFactor * 100) / 100;
+      const protectedBtc =
+        Math.round((withdrawBtc - sacrificialBtc) * 100) / 100;
+      if (canSplit && protectedBtc > 0) {
+        suggestion = cliff.withdrawResplitSuggestion(
+          btc2(withdrawBtc),
+          btc2(sacrificialBtc),
+          btc2(protectedBtc),
+        );
+      } else {
+        // Splitting disallowed or the re-split degenerates — fall back.
+        suggestion = cliff.noSplitSuggestion;
+      }
     }
 
     warnings.push({
       type: "cliff",
-      title: cliff.single.title,
-      detail: cliff.single.detail,
+      title: cliff.title,
+      detail: cliff.body,
       suggestion,
     });
   }
 
-  // CASE 2: Exactly two vaults.
+  // CASE 2: Exactly two vaults — shared shell, keep the structural suggestion.
   else if (nVaults === 2) {
     if (isCliff && group1ReorderWouldHelp && optimalVaultOrder) {
       // Cliff that swapping the order fixes — surfaced as the reorder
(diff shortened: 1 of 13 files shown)


javi0guerrero4-cloud (approved):

services/vault/src/applications/aave/positionNotifications/calculate.ts
```diff
@@ -435,25 +438,37 @@ export function calculate(params: CalculatorParams): CalculatorResult {
...
+      suggestion = cliff.addSacrificialSuggestion(btc2(suggestedNewVaultBtc));
     } else {
-      suggestion = cliff.single.oversizedSuggestion(btc2(raw));
+      // Variant B — re-split the existing vault. seizedFraction depends only on
+      // CF/maxLB/THF, so re-splitting the same total is valid; size the
+      // sacrificial to cover the seizure first and protect the remainder.
+      const withdrawBtc = vaults[0].btc;
+      const sacrificialBtc = Math.ceil(withdrawBtc * liqFactor * 100) / 100;
+      const protectedBtc =
+        Math.round((withdrawBtc - sacrificialBtc) * 100) / 100;
```
jrwbabylonlab: These three amounts can fail to reconcile: `withdrawBtc` is full-precision while `sacrificialBtc` is ceil'd to cents and `protectedBtc` is round'd, so a ~0.045 BTC vault under aggressive seizure params renders "withdraw 0.04 ... 0.03 sacrificial + 0.02 protected" (0.05 > 0.04, an infeasible re-deposit larger than what was withdrawn). Snapping `withdrawBtc` to cents before deriving the parts lines all three up. Non-blocking.

jrwbabylonlab (dismissed):

services/vault/src/applications/aave/positionNotifications/calculate.ts
gbarkhatov: done

jrwbabylonlab (approved):

javi0guerrero4-cloud (commented):
Check out

PR #1978, opened 2026-06-27
feat/animation

jeremy-babylonlabs:


services/vault/src/globals.css
```diff
@@ -21,8 +21,38 @@ html {
    * Used by the deposit progress bar, the steps-completed pill, and completed
    * step checks. */
   --success-bright: 21 183 104;
+  /* Motion spec — Modals (Figma 7831:25100 §1) */
+  --motion-duration-modal: 220ms;
+  --motion-duration-modal-out: 180ms;
+  --motion-duration-backdrop: 220ms;
```
jrwbabylonlab, line 27: Backdrop-out is 220ms but `modal-out` is 180ms, and the dialog unmounts on the modal panel's `onAnimationEnd` — so on close the backdrop fade gets cut ~40ms early (small pop), and it plays ease-out while motion-system.md lists the backdrop exit as 180ms ease-in. A dedicated `backdrop-out` token matching modal-out would reconcile both. Non-blocking.

docs/motion-system.md
jrwbabylonlab, line 1: what's this file for? are we plan to add more motion related stuff?

jrwbabylonlab (approved):

PR #1980, opened 2026-06-29
fix(wallet-connector): loop signPsbt when Utila omits batch signing

jrwbabylonlab:
## Summary

`UtilaProvider.signPsbts` called `this.provider.signPsbts(...)` on the injected
`window.utila.bitcoin` object without checking the method exists. Utila's MPC
wallet does not implement batch signing yet (their Vault integration is still in
progress on their side), so that call throws `TypeError: signPsbts is not a
function`. The Utila team asked us to fall back to singular `signPsbt`.

This adds a presence guard that loops `signPsbt` when `signPsbts` is absent,
mirroring the existing `deriveContextHash` guard in the same adapter. Batch
signing is "STRONGLY RECOMMENDED, not required" per the integration guide, and
the sequential fallback is already the documented degradation path.

## Why the guard has to live in the adapter

The SDK-layer `signPsbtsWithFallback` already checks `typeof wallet.signPsbts
=== "function"` — but `wallet` there is the `UtilaProvider` instance, which
always defines `signPsbts` as a class method, so that check is always true for
Utila. The only place that can see whether the *underlying* injected provider
supports batch signing is the adapter itself.

## Changes
- Guard `typeof this.provider.signPsbts !== "function"` in
  `UtilaProvider.signPsbts`; loop `this.signPsbt` (keeps per-call connection
  guards + error mapping) with `options?.[i]` aligned to `psbtsHexes[i]`.
- Update the adapter's header comment, which incorrectly claimed Utila forwards
  every method through / fully implements the interface.a

### DIFF
--- a/packages/babylon-wallet-connector/src/core/wallets/btc/utila/provider.ts
+++ b/packages/babylon-wallet-connector/src/core/wallets/btc/utila/provider.ts
@@ -18,11 +18,11 @@ const UTILA_PROMPT_TIMEOUT_MS = 60_000;
 
 /**
  * Utila is an MPC wallet that injects an `IBTCProvider`-compatible object at
- * `window.utila.bitcoin` (see docs/vault-integration-guide.md — it implements
- * the documented Bitcoin Wallet Interface directly). Unlike the UniSat/OneKey
- * adapters there is no request-shape translation; this adapter forwards each
- * method through and adds the connection guards + deriveContextHash error
- * mapping the dApp expects.
+ * `window.utila.bitcoin`, targeting Babylon's Bitcoin Wallet Interface
+ * (docs/vault-integration-guide.md). Most methods forward straight through with
+ * connection guards + deriveContextHash error mapping. `signPsbts` is the
+ * exception: Utila does not implement batch signing, so the adapter falls back
+ * to looping `signPsbt` (confirmed with the Utila team).
  *
  * Its `deriveContextHash` is MPC-based, not HD/HKDF — cross-wallet portability
  * is not provided, which the spec permits for non-HD wallets (the dApp only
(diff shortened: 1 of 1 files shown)


jonybur (approved):

packages/babylon-wallet-connector/src/core/wallets/btc/utila/provider.ts
```diff
@@ -168,6 +168,16 @@ export class UtilaProvider implements IBTCProvider {
       });
     }
 
+    // Batch signing is STRONGLY RECOMMENDED, not required. If Utila's injected
+    // provider omits signPsbts, sign each PSBT sequentially via signPsbt.
+    if (typeof this.provider.signPsbts !== "function") {
```
gbarkhatov, line 173: Non-blocking: there are no `UtilaProvider` tests anywhere in the repo, so this new fallback branch (and the `deriveContextHash` guard below it) is uncovered. A focused unit test with a stub `provider` would lock in the behavior this PR depends on: (a) loops `signPsbt` per PSBT in order with `options[i]` aligned when `provider.signPsbts` is absent; (b) takes the native `signPsbts` path when it exists; (c) a mid-loop rejection propagates as a mapped `WalletError`. Cheap to add and worth it given this is signing code, but not a blocker since the adapter has zero existing tests.

gbarkhatov (approved):
Clean, focused fix. The guard sits after the `walletInfo` + non-empty-array checks, so input validation still runs before any signing path. `options?.[i]` aligns each option with `psbtsHexes[i]`, matching the SDK-layer `signPsbtsWithFallback`. Delegating to `this.signPsbt` preserves the per-call connection guard and `mapPromptRejection` mapping, and a mid-loop failure throws (no partial result). The fallback pushes exactly one signed PSBT per input, so count integrity holds; the native path's length is asserted at the SDK layer as the PR body describes. The header-comment update keeps the docs accurate.

Note (UX, not correctness): the sequential fallback means N separate MPC approval round-trips for a multi-vault deposit instead of one batch prompt — this is the documented "STRONGLY RECOMMENDED, not required" degradation path, so acceptable, just flagging the behavioral change for Utila users.

One non-blocking note on test coverage inline. Approving.

PR #1984, opened 2026-06-29
feat(packages): makes deposit card not clickable when finished

jonybur:


jrwbabylonlab (approved):

services/vault/src/components/simple/PendingDepositCard.tsx
```diff
@@ -99,6 +100,13 @@ export function PendingDepositCard({
   const isDanger = displayVariant === "danger";
   const dotColor = isDanger ? undefined : STATUS_DOT_COLORS[displayVariant];
 
+  // A refunded deposit is terminal: the card's click opens the refund modal,
+  // which has nothing left to do once the refund has settled on-chain. Drop the
+  // handler so the row isn't presented as actionable (no hover, no pointer).
+  const isRefunded = peginState.displayLabel === PEGIN_DISPLAY_LABELS.REFUNDED;
```
jrwbabylonlab, line 106: the REFUNDED label here comes from the indexer htlcVout, which we don't validate on-chain (fetchVaults.ts:366). if that vout's wrong and lands on a spent output, a still-refundable vault reads REFUNDED and the card goes dead — killing the only path to the modal that rechecks on-chain. narrow + pre-existing, but maybe leave it clickable and let the modal decide?

services/vault/src/components/simple/__tests__/PendingDepositCard.test.tsx
```diff
@@ -32,18 +36,21 @@ vi.mock("../VaultDetailCard", () => ({
...
     disabled?: boolean;
     disabledTooltip?: string;
+    onClick?: () => void;
     txHashRow?: ReactNode;
   }) => (
     <div
       data-testid="vault-detail-card"
       data-disabled={disabled ? "true" : "false"}
       data-disabled-tooltip={disabledTooltip ?? ""}
+      data-clickable={onClick ? "true" : "false"}
```
jrwbabylonlab, line 53: the mock reads onClick for data-clickable but never wires it on the div, so this only checks the prop was passed, not that the click is actually dead. could wire onClick + fireEvent.click here (not called for REFUNDED, called w/ depositId for EXPIRED).

services/vault/src/components/simple/__tests__/PendingDepositCard.test.tsx
```diff
@@ -146,6 +153,50 @@ describe("PendingDepositCard — disabled (ownership mismatch) surface", () => {
...
 });
 
+describe("PendingDepositCard — refunded cards are not clickable", () => {
+  function renderWithClick(displayLabel: string) {
+    mockUseDepositPollingResult.mockReturnValue({
+      loading: false,
+      peginState: {
+        contractStatus: ContractStatus.EXPIRED,
+        displayLabel,
+        displayVariant: "inactive",
```
jrwbabylonlab, line 163: nit: a non-refunded EXPIRED card is displayVariant 'warning', not 'inactive'.

jrwbabylonlab (commented):
looks good overall — small stuff inline, main one's the refund-entry gating.

jrwbabylonlab:
Scope nit: both changed files are under `services/vault` and nothing in `packages/` changes — suggest retitling the squash to `feat(vault)` so the changelog scope is correct.

services/vault/src/components/simple/PendingDepositCard.tsx
jrwbabylonlab: following up — i dug into the indexer + contracts and the stranding worry i raised here is basically moot for our flow. the indexer's htlcVout isn't computed, it's copied straight from the PegInSubmitted event (pegin.ts:121) = the same value the contract stores (PeginLogic.sol:130/358) that getVaultFromChain reads, so indexer == on-chain 1:1. and only the depositor can spend an expired vault's refund leaf, so a "spent HTLC" really means the refund landed. so this REFUNDED gate is fine as-is — no rework needed.

PR #1987, opened 2026-06-30
feat(vault): pause per-scope on-chain detection and gate all exits

### LINKED ISSUE #1975: Pause: gate all state-changing actions; keep recovery paths open
## Context (per the governance spec)

Follow-up to #1954 and PR #1969. Per the governance spec ([Final Proposal] Governance, Pausing & Access Control) and Aleksa's confirmation ("forget soft/full — freeze and pause"), **Pause is the full stop** across both scopes — a last-resort emergency reserved for bugs in the exit / activation paths.

## What Pause blocks vs. preserves (two independent scopes)

- **Protocol scope** (`BTCVaultsManager`, `ProtocolParams`, `ApplicationRegistry`, `FeeEscrow`) — Paused additionally blocks: activation (`activateVaultWithSecret`), redemption (`redeemVaultForDepositor` / `...ForAVK`), fee claims. Preserves ONLY: expired-pegin cleanup (`reportExpiredPegins`), depositor fee refunds.
- **Aave scope** (`AaveIntegrationController`, `VaultBTC`) — Paused additionally blocks: activation, repay, withdraw, liquidation, vault-swap. Preserves ONLY: delegation revoke/renounce.

Plus **BTC-side recovery** (Pre-PegIn refund, WOTS self-claim) is always available — off-chain, pre-signed txs with the depositor's own keys.

**Activation nuance:** preserved under Freeze (time-critical — a depositor with BTC already locked must activate before expiry), **blocked under Pause.**

## Cross-scope viability (spec §4 — the FE gating contract)

Pause viability differs by scope; several exits depend on BOTH scopes (✓ = viable):

| Flow | Protocol Paused | Aave Paused |
|---|---|---|
| Activate | **blocked** | **blocked** |
| Withdraw (+redeem) | **blocked** | **blocked** |
| Repay | **✓** | **blocked** |
| Borrow / reorder / delegation-approve | ✓ | **blocked** |
| Peg-in fee claims | **blocked** | ✓ |
| Revoke / renounce delegation | ✓ | ✓ |
| Expired pegin cleanup | ✓ | ✓ |

Withdraw and activation are blocked if **either** scope pauses; **repay** is blocked only by Aave-pause; **fee claims** only by protocol-pause.

## Exit-gating MUST be scope-accurate (do NOT reuse the single Freeze/Pause flag)

Entry-gating (#1969) can safely over-block with one flag — blocking a deposit you didn't strictly need to is harmless. Exit-gating cannot. From the matrix:

- **Repay is viable under Protocol-Pause** (it needs only the Aave scope).
- **Fee claims are viable under Aave-Pause** (they need only the protocol scope).

If we gate exits on a single generic "Paused" flag, we would **wrongly disable repay during a protocol-only pause** — blocking a user from de-risking exactly when they need to. That is worse than an optimistic banner. **Therefore per-scope on-chain detection is a prerequisite for this issue, not a deferred design choice.** (Withdraw / activation are safe under a conservative union since they are blocked if either scope pauses.)

## Shipped in #1969

Deposit + borrow disabled (shared with Freeze); `NEXT_PUBLIC_PROTOCOL_STATUS_MESSAGE` banner-body override.

## Remaining Pause work

- **Gate the depositor exit CTAs** the FE exposes — withdraw (+redeem), repay, **activation**, reorder, redemption — each on its correct scope (see warning above).
- **Verify the preserved recovery paths stay open** under Pause: expired-pegin cleanup, depositor fee refunds, BTC refund, WOTS self-claim, artifact download.
- **Per-scope on-chain detection** — prerequisite, not optional (see above).
- Out of FE scope: liquidation (`liquidateCorePosition`), `FeeEscrow.claim`, and vault-swap redemption (`redeemVaultFromSwap`) are keeper / VP / arbitrageur actions — contract-enforced, no depositor CTA to gate.

## Detection

Each scope exposes Frozen/Paused state + `Frozen` / `Paused` / `Unfrozen` / `Unpaused` events (`freeze()` → Frozen, `pause()` → Paused). Read Aave-scope state from `AaveIntegrationController`, protocol-scope from `BTCVaultsManager`. Listening to all contracts is "not optimal"; the operator flag is a valid interim for the banner, but exit-gating needs the real per-scope read (above).

Optional awareness (contract-side, surface only if useful): Pause auto-expires after 7 days (§2.6); a 1-hour post-unpause liquidation grace period (§2.5) lets borrowers restore health before liquidations resume.

## Links

- Spec: https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/398655537 (sections 2.1, 2.3, 3.1, 3.3, 4)
- Original: #1954 · Built on: #1969 · Sibling (Freeze): #1974

## Acceptance criteria

- [ ] Depositor exit CTAs (withdraw / repay / activation / reorder / redemption) disabled under Pause, each gated on its correct scope (repay = Aave-pause only; withdraw / activation = either scope)
- [ ] recovery paths (expired cleanup, fee refunds, BTC refund / WOTS, artifacts) confirmed available under Pause
- [ ] per-scope on-chain detection in place (prerequisite for exit-gating); lint + tests pass

gbarkhatov:
# Pause: per-scope detection and exit gating

Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/1975

## What this does

When the protocol is **Paused**, this disables every depositor action that changes on-chain state — deposit, borrow, reorder, withdraw, repay, and activation — while leaving the recovery paths (BTC refund, recovery-artifact download) working. Freeze keeps doing what it did before: it blocks new entry (deposit/borrow/reorder) but leaves exits open. To do this correctly the app now reads the real pause/freeze state from the contracts, per scope, instead of relying only on a single operator flag.

## Why

The protocol has two independent "scopes" — the core protocol (`BTCVaultRegistry`) and the Aave integration (`AaveIntegrationAdapter`) — and each can be Frozen or Paused on its own. The rules differ by action: withdrawing and activating are blocked if **either** scope is paused; repaying is blocked **only** by an Aave pause; depositing depends **only** on the protocol scope. A single global flag can't express this — and getting it wrong is harmful: if we blocked repay during a protocol-only pause, we'd stop a user from reducing their debt exactly when they need to. So scope-accurate state is required, not optional.

## Approaches considered, and what we chose

**How to know the pause state.** Option A: keep using the single operator flag. Option B: read the real per-scope state from the contracts. We chose B because exit-gating must be scope-accurate (Option A would over-block). Both contracts expose a `pauseState()` view; the Aave one was already in our bundled ABI, and the protocol one needed a tiny ABI fragment added. We kept the operator flag as a **manual override on top** of the on-chain reads (a DevOps "big red button"), which also serves as the safe fallback while the read is loading or if it fails.

**Sync helpers vs a reactive hook.** The old gating helpers were plain no-argument functions. Since the state now comes from an async contract read, we made the helpers **pure functions of a passed-in state object** (easy to unit-test against the full rules matrix) and added a small React-Query hook that supplies that state. Components call `useProtocolGateState()` and pass the snapshot to the matching `is*Blocked(gate)` predicate to disable buttons; transaction hooks read the same state and guard their execution chokepoint — the same disable-the-button-plus-guard-the-handler pattern we already use for reorder.

**Fail-open for exits.** If the on-chain read fails (e.g. a brief RPC hiccup), exits are **not** blocked — only the operator flag can block them. Over-blocking an exit traps users, so exits fail open; entries can still be operator-blocked.

**Enum mapping safety.** The contract returns a numeric `PauseState` enum. The number→meaning mapping (`0 = normal, 1 = frozen, 2 = paused`) is **confirmed against the contract source** (`ITBVPausable.sol`: `enum PauseState { None, Frozen, Paused }`). To guard a future enum change, an unrecognized value **throws** instead of defaulting to "not paused".

## Changes

- New `clients/eth-contract/pause-state/query.ts` — reads `pauseState()` from both contracts in one multicall and maps it to a per-scope status (throws on an unknown value).
- New `hooks/useProtocolGate.ts` — `useProtocolPauseStatus` (polls the read) and `useProtocolGateState` (composes the on-chain reads with the operator flag).
- `components/shared/protocolStatus.ts` — refactored to per-scope: status types, `composeGateState`, and pure `is*Blocked(gate)` predicates for deposit/borrow/reorder/withdraw/repay/activation, plus `resolveBannerStatus`.
- Wired the gating: re-scoped deposit (protocol) and borrow/reorder (aave); added withdraw, repay (aave-pause only), and activation gating at both the button and the transaction chokepoint. Every state-changing action — deposit, borrow, reorder, withdraw, repay, activation — guards its execution path, not just its button. The position-notification banner CTAs (Add Collateral / Add vault / Repay Debt) are gated too. The status banner reflects the real per-scope state.
- Repay shows an explanatory label + callout when blocked by a pause (mirroring Borrow), so a user near liquidation sees that it's a governance pause, not a silently greyed-out button.
- Tests — the full rules matrix as pure-function tests, the enum mapping (including throw-on-unknown), the activation guard (never reveals the secret while paused), the borrow guard, and the banner CTA gating. A default unblocked gate is mocked in the test setup so unrelated tests don't need a query client.

## What this intentionally does NOT change

- **Recovery stays available** under Pause: BTC refund, WOTS self-claim, artifact download, and expired-pegin cleanup are not gated.
- No change to the actual transaction-building or value logic; this only decides when a CTA is allowed to run.

## How to verify

- tsc, lint (0 errors), and the full vault test suite pass (2056 tests).
- Manually, with no chain needed: set `NEXT_PUBLIC_FF_PROTOCOL_PAUSED=true` and restart — deposit, borrow, reorder, withdraw, repay, and activation are all disabled, and refund/artifact recovery still works. Set `NEXT_PUBLIC_FF_PROTOCOL_FROZEN=true` instead — only deposit/borrow/reorder are disabled; withdraw, repay, and activation stay enabled. On a testnet whose contracts report a real `pauseState()`, the on-chain read drives the gating per scope.

### DIFF
--- a/services/vault/src/applications/aave/hooks/useRepayTransaction.ts
+++ b/services/vault/src/applications/aave/hooks/useRepayTransaction.ts
@@ -12,7 +12,9 @@ import { parseUnits } from "viem";
 import { useAccount, useWalletClient } from "wagmi";
 
 import { ERC20 } from "@/clients/eth-contract";
+import { isRepayBlocked } from "@/components/shared/protocolStatus";
 import { getETHChain } from "@/config/network";
+import { useProtocolGateState } from "@/hooks/useProtocolGate";
 import { logger } from "@/infrastructure";
 import {
   ErrorCode,
@@ -111,6 +113,7 @@ export function useRepayTransaction({
   const { address } = useAccount();
   const queryClient = useQueryClient();
   const chain = getETHChain();
+  const gate = useProtocolGateState();
 
   const clearError = useCallback(() => setError(null), []);
 
@@ -123,6 +126,10 @@ export function useRepayTransaction({
     const { preSignValidation, repayAmountRaw } = options;
 
     if (repayAmount <= 0) return false;
+
+    // Repay is an aave-scope EXIT: blocked only by an aave Pause (not a protocol
+    // pause, and never by Freeze). Guard the chokepoint behind the disabled button.
+    if (isRepayBlocked(gate)) return false;
 
     setError(null);
     setIsProcessing(true);
--- a/services/vault/src/components/simple/SimpleDeposit.tsx
+++ b/services/vault/src/components/simple/SimpleDeposit.tsx
@@ -14,6 +14,7 @@ import { useDepositPeginFee } from "@/hooks/deposit/useDepositPeginFee";
 import { useDialogStep } from "@/hooks/deposit/useDialogStep";
 import { usePendingVaultOverlapCheck } from "@/hooks/deposit/usePendingVaultOverlapCheck";
 import { useProtocolFeeRows } from "@/hooks/useProtocolFeeRows";
+import { useProtocolGateState } from "@/hooks/useProtocolGate";
 import { useVaultCountCap } from "@/hooks/useVaultCountCap";
 import { resolveVaultCapState } from "@/services/deposit/vaultCap";
 import type { VaultActivity } from "@/types/activity";
@@ -74,6 +75,7 @@ function SimpleDepositContent({
   onClose,
   initialAmountBtc,
 }: SimpleDepositBaseProps) {
+  const gate = useProtocolGateState();
   const { isGeoBlocked, isLoading: isGeoLoading } = useGeoFencing();
   const { isBlocked: isAddressBlocked, isLoading: isScreeningLoading } =
     useAddressScreening();
@@ -313,7 +315,7 @@ function SimpleDepositContent({
     // if a deposit entry point that bypasses the disabled buttons (e.g. the
     // Activity empty-state CTA or the urgent Add Collateral banner) opens this
     // dialog.
-    if (isDepositBlocked()) return;
+    if (isDepositBlocked(gate)) return;
 
     // Per-position BTC Vault cap: never start a deposit that would push the
     // position past the on-chain cap, or when the cap couldn't be read (fail
@@ -455,7 +457,7 @@ function SimpleDepositContent({
                   isReconnectingWallet,
                 }}
                 gatingState={{
-                  isDepositDisabled: isDepositBlocked(),
+                  isDepositDisabled: isDepositBlocked(gate),
                   isGeoBlocked: isGeoBlocked || isGeoLoading,
                   isAddressBlocked: isAddressBlocked || isScreeningLoading,
                   ordinalsCheckPending,
--- a/services/vault/src/copy.ts
+++ b/services/vault/src/copy.ts
@@ -856,6 +856,11 @@ export const COPY = {
       "Borrowing is temporarily unavailable. Please check back later.",
     priceUnavailable:
       "Price data unavailable. Borrowing is temporarily disabled.",
+    // Shown on the Repay tab when repay is blocked by a protocol pause (not a
+    // technical/user error), so a user near liquidation knows it's governance,
+    // not a bug. Repay is gated only by an aave-scope pause.
+    repayingUnavailable:
+      "Repaying is temporarily unavailable while the protocol is paused. It will resume once the pause is lifted.",
     // Borrow tab — action-button labels (also used as the status-callout title).
     borrow: {
       action: "Borrow",
@@ -911,6 +916,8 @@ export const COPY = {
     repay: {
       action: "Repay",
       processing: "Processing...",
+      // Action-button label when repay is blocked by a protocol pause.
+      unavailable: "Repaying Unavailable",
       enterAmount: "Enter an amount",
       amountTooSmall: "Amount too small",
       amountExceedsDebt: "Amount exceeds debt",
--- a/services/vault/src/copy.ts
+++ b/services/vault/src/copy.ts
@@ -84,6 +84,11 @@ export const COPY = {
     },
     // Row label for the vault creation time (rendered as relative time).
     createdLabel: "Created",
+    // Shown if activation is attempted while the protocol is paused — surfaced
+    // as the activation error so the spinner clears and the user understands
+    // it's a governance pause (not a failed secret). Activation resumes on unpause.
+    activationPaused:
+      "Activation is paused by a protocol governance action. Your BTC Vault stays safe — activation will resume once the pause is lifted.",
     messages: {
       payoutSignaturesSubmitted:
         "Payout signatures submitted. Vault provider is verifying and collecting acknowledgments...",
@@ -483,6 +488,11 @@ export const COPY = {
     errors: {
       invalidSecret:
         "Invalid secret: SHA256(secret) does not match the BTC Vault's hashlock. Please check your secret and try again.",
+      // Surfaced if deposit execution is reached while the protocol is frozen or
+      // paused — aborted up front, before the on-chain registration and the BTC
+      // broadcast, so no funds are locked and the user can retry once it resumes.
+      protocolPaused:
+        "New deposits are temporarily disabled while the protocol is frozen or paused. No Bitcoin was sent — please try again once it resumes.",
       chainSwitchRequired: (network: string) =>
         `Please switch to ${network} in your wallet`,
       ethereumMainnet: "Ethereum Mainnet",
@@ -1079,12 +1089,19 @@ export const COPY = {
   protocolStatus: {
     frozen: {
       title: "Protocol is frozen",
-      body: "New deposits and borrows are disabled. You can still repay debt — liquidations remain active.",
+      // Non-specific about which actions: gating is per-scope, so naming
+      // "deposits and borrows" can be inaccurate (e.g. an aave-only freeze
+      // leaves deposits working). Exits are always preserved under a freeze, so
+      // that reassurance is safe to state. The per-action buttons are the
+      // precise source of truth.
+      body: "Some new actions are temporarily restricted while the protocol is frozen. Any unavailable action is disabled and explains why. Your exits — repay, withdraw, and activation — stay available.",
       learnMore: PROTOCOL_STATUS_LEARN_MORE,
     },
     paused: {
       title: "Protocol is paused",
-      body: "New deposits and borrows are disabled. Debt continues accruing interest. Monitor official announcements.",
+      // Non-specific for the same reason — under a per-scope pause the exact set
+      // of blocked actions varies. Each affected button explains itself.
+      body: "Some actions are temporarily unavailable while the protocol is paused. Any unavailable action is disabled and explains why. Debt continues accruing interest — monitor official announcements.",
       learnMore: PROTOCOL_STATUS_LEARN_MORE,
     },
   },
--- a/services/vault/src/hooks/deposit/useVaultActions.ts
+++ b/services/vault/src/hooks/deposit/useVaultActions.ts
@@ -25,12 +25,16 @@ import { useEffect, useRef, useState } from "react";
 import type { Hex } from "viem";
 import { getWalletClient, switchChain } from "wagmi/actions";
 
-import { isActivationBlocked } from "@/components/shared/protocolStatus";
+import {
+  composeGateState,
+  isActivationBlocked,
+} from "@/components/shared/protocolStatus";
 import { getETHChain } from "@/config/network";
 import { COPY } from "@/copy";
 import { useProtocolGateState } from "@/hooks/useProtocolGate";
 
 import { getVaultFromChain } from "../../clients/eth-contract/btc-vault-registry/query";
+import { getOnChainPauseState } from "../../clients/eth-contract/pause-state/query";
 import { getVaultRegistryReader } from "../../clients/eth-contract/sdk-readers";
 import {
   ContractStatus,
@@ -349,8 +353,13 @@ export function useVaultActions(): UseVaultActionsReturn {
     // Activation is an EXIT blocked under Pause (either scope); preserved under
     // Freeze (time-critical — a depositor with BTC locked must still activate).
     // Guard the chokepoint behind the disabled Activate button; never reveal the
-    // secret on-chain while paused.
-    if (isActivationBlocked(gate)) return;
+    // secret on-chain while paused. Surface a paused error (rather than a silent
+    // return) so the caller's spinner clears via `!activationError` and the user
+    // gets feedback. A fresh on-chain re-check below closes the stale-gate window.
+    if (isActivationBlocked(gate)) {
+      setActivationError(COPY.pegin.activationPaused);
+      return;
+    }
 
     const {
       vaultId,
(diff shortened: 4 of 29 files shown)


services/vault/src/hooks/useProtocolGate.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * The effective per-scope gate state: on-chain reads composed with the operator
+ * flags (more severe wins). While loading or on a read failure, on-chain is
+ * treated as `null`, so the state falls back to the operator-flag value — exits
+ * stay available unless an operator explicitly paused.
+ */
+export function useProtocolGateState(): ProtocolGateState {
+  const { data } = useProtocolPauseStatus();
+  return useMemo(() => composeGateState(data ?? null), [data]);
+}
```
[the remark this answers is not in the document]
gbarkhatov: done

services/vault/src/hooks/useProtocolGate.ts
```diff
@@ -0,0 +1,79 @@
...
+      depositBlocked: isDepositBlocked(gate),
+      borrowBlocked: isBorrowBlocked(gate),
+      reorderBlocked: isReorderBlocked(gate),
+      withdrawBlocked: isWithdrawBlocked(gate),
+      repayBlocked: isRepayBlocked(gate),
+      activationBlocked: isActivationBlocked(gate),
+    }),
+    [gate],
+  );
+}
```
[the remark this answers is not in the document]
gbarkhatov: done

gbarkhatov:
All greptile comments resolved (so as `services/vault/src/applications/aave/components/LoanCard/Repay/index.tsx, line 381-399 (link) P1 Repay button silently d...`)

services/vault/src/hooks/useProtocolGate.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+export function useProtocolGateState(): ProtocolGateState {
+  const { data } = useProtocolPauseStatus();
+  // Depend on the primitive scope values (not the `data` object) so the gate
+  // keeps a stable reference across polls when the status is unchanged — this
+  // avoids churning the `useCallback`s in the tx hooks that consume it. Operator
+  // flags are env-static, so these two values are the only dynamic inputs to
+  // `composeGateState` (passing `{ protocol: null, aave: null }` is equivalent
+  // to passing `null` — it collapses to the operator-flag value).
+  const protocol = data?.protocol ?? null;
```
jrwbabylonlab, line 55: `data?.protocol ?? null` collapses an unknown-enum throw (from `mapPauseState`) into the same fail-open path as a transient read error — silently defeating the throw-on-unknown safety property `query.ts` deliberately adds. A future `PauseState` enum change would leave actions open instead of failing closed. Suggest distinguishing unknown-enum from read-failure and failing closed on the former.

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -342,6 +346,12 @@ export function useVaultActions(): UseVaultActionsReturn {
    * Handle vault activation — reveal HTLC secret on Ethereum
    */
   const handleActivation = async (params: ActivateVaultParams) => {
+    // Activation is an EXIT blocked under Pause (either scope); preserved under
+    // Freeze (time-critical — a depositor with BTC locked must still activate).
+    // Guard the chokepoint behind the disabled Activate button; never reveal the
+    // secret on-chain while paused.
+    if (isActivationBlocked(gate)) return;
```
jrwbabylonlab: Two things on this activation guard: (1) it reads a cached gate (20s refetch / 10s stale) with no fresh `pauseState()` preflight before the secret is submitted — a paused activation in the stale window still reverts on-chain (`activateVaultWithSecret` is `whenNotPaused`), but the HTLC secret is already in the broadcast calldata (public). A fresh preflight for activation only closes that. (2) The silent `return` leaves the consumer stuck: `useActivationState.handleActivation` sets `localActivating=true` and only clears it in the success callback or `catch`, so a blocked call spins forever — return a boolean / throw a typed blocked-error and clear in `finally`.

services/vault/src/components/simple/SimpleDeposit.tsx
```diff
@@ -313,7 +315,7 @@ function SimpleDepositContent({
     // if a deposit entry point that bypasses the disabled buttons (e.g. the
     // Activity empty-state CTA or the urgent Add Collateral banner) opens this
     // dialog.
-    if (isDepositBlocked()) return;
+    if (isDepositBlocked(gate)) return;
```
jrwbabylonlab, line 318: This gates the dialog-open, but the actual BTC broadcast happens later in `DepositSignContent.startFlow → useDepositFlow.executeDeposit`, which has no gate. If pause lands between opening the sign step and clicking Sign, the pre-pegin BTC tx still broadcasts (the ETH register reverts via `whenNotFrozen`, but the user has locked BTC needing refund). Unlike the ETH actions, the BTC broadcast is not contract-enforceable — gate `executeDeposit` too.

services/vault/src/applications/aave/hooks/useRepayTransaction.ts
```diff
@@ -124,6 +127,10 @@ export function useRepayTransaction({
 
     if (repayAmount <= 0) return false;
 
+    // Repay is an aave-scope EXIT: blocked only by an aave Pause (not a protocol
+    // pause, and never by Freeze). Guard the chokepoint behind the disabled button.
+    if (isRepayBlocked(gate)) return false;
```
jrwbabylonlab, line 132: Guard is correct (aave-pause only), but this is the subtlest rule — repay must stay available under a protocol-only pause so a near-liquidation user can de-risk — and it has no hook test, while borrow/activation do. A future mis-wire to `isWithdrawBlocked` (blocks on either scope) would trap users into liquidation with every test still green. Add a hook test: `{protocol:"paused",aave:null}` → proceeds; `{aave:"paused"}` → returns false.

services/vault/src/copy.ts
```diff
@@ -856,6 +856,11 @@ export const COPY = {
       "Borrowing is temporarily unavailable. Please check back later.",
     priceUnavailable:
       "Price data unavailable. Borrowing is temporarily disabled.",
+    // Shown on the Repay tab when repay is blocked by a protocol pause (not a
+    // technical/user error), so a user near liquidation knows it's governance,
+    // not a bug. Repay is gated only by an aave-scope pause.
+    repayingUnavailable:
```
jrwbabylonlab, line 872: This accurate per-tab copy now contradicts the global status-banner body (`paused.body`, ~line 1082, unchanged: "You can still repay debt"), written for the old global-flag model. Under a real pause the banner tells users repay works while this string says it does not. Reconcile the banner body with the per-scope gating, or make it non-specific.

jrwbabylonlab (dismissed):

services/vault/src/hooks/useProtocolGate.ts
gbarkhatov: Good catch — fixed in `ad7d3212`. `mapPauseState` now returns `"paused"` (most-restrictive) on an unrecognized value instead of throwing, so an unknown enum **fails closed**. A genuinely reverted multicall still throws → the React-Query hook fails **open** (so an RPC blip never traps a user mid-exit). The two failure modes are now distinct and both tested in `pause-state/__tests__/query.test.ts` (`[3,0]` → `{protocol:"paused"}`; reverted read → throws).

services/vault/src/hooks/deposit/useVaultActions.ts
gbarkhatov: Both addressed in `ad7d3212`.

**(1) Stuck spinner:** the blocked guard now `setActivationError(COPY.pegin.activationPaused)` and returns. `useActivationState`'s `isActivating = (vaultActivating || localActivating) && !activationError`, so setting the error both clears the spinner and surfaces feedback instead of spinning forever.

**(2) Stale-gate secret reveal:** added a fresh `getOnChainPauseState()` preflight, folded into the existing parallel on-chain read (`Promise.all` with `getVaultData`) and re-checked via `composeGateState` **before** the secret reaches `activateVaultWithSecret` calldata. A failed fresh read falls back to the cached gate (activation is time-critical — an RPC blip mustn't trap a depositor near their deadline). New `useVaultActions.test` cases cover both the cached-gate block and a fresh-preflight catch in the stale window.

services/vault/src/components/simple/SimpleDeposit.tsx
gbarkhatov: I dug into this one and ended up **not** adding a pre-broadcast preflight — I don't think the scenario can occur, because the flow is **register-first**:

- `useDepositFlow.ts:6` (header): *"the Pre-PegIn is broadcast only after ETH registration."*
- `useDepositFlow.ts:558`: `await registerPeginBatchAndWait(...)` — contract-enforced (`whenNotFrozen`) and **awaited**.
- `useDepositFlow.ts:707`: `broadcastPrePeginTransaction(...)` runs only after that resolves.

So a pause **before** register → the register reverts → the flow throws → **BTC never broadcasts**. A pause **after** register → broadcasting completes a *valid, registered* pegin (and a PENDING resume record exists either way). A pre-broadcast preflight would actually be wrong here — it'd strand a registered pegin (the register-first design exists precisely so "a failed batch never strands BTC in unregistered HTLCs").

What I did add: a clean **entry guard** at `useDepositFlow.ts:324` — `if (isDepositBlocked(gate)) throw …` — so a deposit reached while frozen/paused aborts up front (before the wallet popup and the doomed on-chain register) instead of reverting, plus a `useDepositFlow.test` case asserting paused → returns `null`, sets the pause error, and calls none of prepare/register/broadcast. Let me know if you see a path I'm missing.

services/vault/src/applications/aave/hooks/useRepayTransaction.ts
gbarkhatov: Added in `ad7d3212` — `useRepayTransaction.test.tsx` with exactly your two cases: `{aave:"paused"}` → returns `false` before any on-chain read; `{protocol:"paused", aave:null}` → proceeds (reaches the reserve check). That locks the aave-pause-only rule against a future mis-wire to `isWithdrawBlocked`.

services/vault/src/copy.ts
gbarkhatov: Reconciled in `ad7d3212`. Both `protocolStatus.frozen.body` and `paused.body` are now non-specific about which actions are affected — they defer to the per-action buttons (the precise source of truth) — so the global banner can't contradict per-scope gating. The frozen body keeps the (always-true) "exits stay available" reassurance, since the frozen banner only renders when neither scope is paused.

Minor note: the quoted "You can still repay debt" was actually the **frozen** body (and is correct — repay is preserved under a freeze); the paused body didn't claim repay works. But the underlying imprecision you're pointing at was real — "New deposits and borrows are disabled" is wrong under a single-scope pause (e.g. an aave-only pause leaves protocol deposits working) — and that's what's fixed.

jonybur (approved):

jeremy-babylonlabs (approved):

PR #1988, opened 2026-06-30
fix(vault): keep the deposit card inert while a refund is in flight

jrwbabylonlab:
Extends #1984's refund-click gate from `Refunded` to also cover the in-flight
`Refunding` state — an expired vault whose refund was already broadcast no longer
shows a clickable card that reopens a do-nothing refund modal.

`Refunding` fires from our own broadcast OR the mempool HTLC-spend probe, so this
also catches refunds sent from another device. Extracted the check as
`isRefundInFlightOrSettled` in `peginStateMachine.ts`.

No new RPC (reads an already-computed label); both labels are EXPIRED-only so
pending/batched cards are untouched.

### DIFF
--- a/services/vault/src/components/simple/PendingDepositCard.tsx
+++ b/services/vault/src/components/simple/PendingDepositCard.tsx
@@ -22,7 +22,7 @@ import { DepositFlowStep } from "@/hooks/deposit/depositFlowSteps/types";
 import {
   canPerformAction,
   getPeginDisplayStep,
-  PEGIN_DISPLAY_LABELS,
+  isRefundInFlightOrSettled,
 } from "@/models/peginStateMachine";
 import { getTokenBrandColor } from "@/services/token/tokenService";
 import type { VaultProvider } from "@/types/vaultProvider";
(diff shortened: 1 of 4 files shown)


services/vault/src/components/simple/PendingDepositCard.tsx
```diff
@@ -100,12 +100,13 @@ export function PendingDepositCard({
   const isDanger = displayVariant === "danger";
   const dotColor = isDanger ? undefined : STATUS_DOT_COLORS[displayVariant];
 
-  // A refunded deposit is terminal: the card's click opens the refund modal,
-  // which has nothing left to do once the refund has settled on-chain. Drop the
-  // handler so the row isn't presented as actionable (no hover, no pointer).
-  const isRefunded = peginState.displayLabel === PEGIN_DISPLAY_LABELS.REFUNDED;
+  // The card's click opens the refund modal. Keep it inert once a refund is in
```
gbarkhatov, line 103: Wording nit (non-blocking): this comment calls it "the refund modal," but the file header (lines 7-11) and the `onCardClick` prop doc describe the click as opening the deposit multistepper modal for the whole batch. Since this PR rewrites this exact comment, it's a low-cost chance to align the wording. No behavioral impact.

services/vault/src/models/peginStateMachine.ts
```diff
@@ -224,6 +224,19 @@ export function canPerformAction(
...
 
+/**
+ * True when an EXPIRED vault's refund is already in flight (Refunding — our own
+ * broadcast, or an HTLC spend the mempool probe sees) or settled (Refunded).
+ * Both labels are produced only in the EXPIRED branch, so this is the
+ * display-layer signal that the refund modal has nothing left to do.
+ */
+export function isRefundInFlightOrSettled(state: PeginState): boolean {
+  return (
+    state.displayLabel === PEGIN_DISPLAY_LABELS.REFUNDING ||
```
gbarkhatov, line 235: Noting for the record (non-blocking): this gate is coupled to the `Refunding`/`Refunded` label *strings* only ever being produced in the EXPIRED branch. The JSDoc above already calls this out, which is the right mitigation — flagging it only because reusing either label elsewhere would silently extend this card-inert gate. No action needed now.

gbarkhatov (approved):
Approving — clean, well-scoped fix. The change correctly extends #1984's gate from `Refunded` to also cover the in-flight `Refunding` state, the new helper reuses the already-computed display label (right altitude — no re-deriving the refundSettlement/localStorage precedence), and I verified both labels are produced only in the EXPIRED branch (peginStateMachine.ts:575/582/592), so there's no false-positive risk from another contract state. Tests cover both the suppressed and still-clickable paths.

Two minor, non-blocking inline notes below. I also considered the test-overlap observation (the new component test for `Refunding` inertness overlaps the existing `Refunded` block and the new unit tests) and deliberately did not file it — that's acceptable layered coverage (unit + component), not worth trimming, and matches the repo's "three similar tests over one abstraction" guidance.

PR #1991, opened 2026-06-30
feat(vault): hide overview panel when position is empty

jonybur:
Screenshot 2026-06-30 at 13 45 03


services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -74,6 +74,7 @@ export function DashboardPage() {
     hasLoans,
     hasCollateral,
     hasDisplayCollateral,
+    hasDebt,
```
jrwbabylonlab: `hasDebt` is value-identical to the already-destructured `hasLoans` (both are `debtValueUsd > 0` off the same source) — reuse `hasLoans` for `hasOverviewData` instead of adding a duplicate flag.

jrwbabylonlab (dismissed):

gbarkhatov (dismissed):

jeremy-babylonlabs (approved):

PR #1996, opened 2026-07-01
fix(vault): align user-facing copy with terminology style guide

jrwbabylonlab:
- Replaces all bare "vault"/"vaults" with "BTC Vault"/"BTC Vaults" in user-facing
strings across copy.ts, AddressScreeningBanner, validations.ts, notification
banners, and liquidation warning copy (~30 instances)
- Replaces bare "provider" with "vault provider" in copy.ts, vaultProviderStatus.ts,
and formatting.ts (~12 instances)
- Eliminates "reclaim" as a third synonym for the expired-deposit refund flow,
standardizing on "refund" (copy.ts, errorMessages.ts)
- Fixes "pegin" → "peg-in" and "pre-pegin" → "Pre-Pegin" in thrown error strings
(ResumeDepositContent.tsx)
- Aligns "Bitcoin" vs "BTC" in adjacent split-option descriptions

### DIFF
--- a/services/vault/src/applications/aave/positionNotifications/__tests__/calculate.test.ts
+++ b/services/vault/src/applications/aave/positionNotifications/__tests__/calculate.test.ts
@@ -291,7 +291,7 @@ describe("golden vectors (reference scenario suite)", () => {
     // existing vault becomes protected.
     expect(result.suggestedNewVaultBtc).toBe(0.72);
     expect(getWarning(result.warnings, "cliff")?.suggestion).toContain(
-      "0.72 BTC sacrificial vault creates a buffer",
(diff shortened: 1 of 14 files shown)


jonybur:
Minor (non-blocking): in `services/vault/src/components/simple/__tests__/PostDepositContinuationView.test.tsx:168`, the `@/copy` mock still hardcodes the old title `"Provider not found"`. It is harmless (the mock is self-contained and the test asserts against its own mock), but it now diverges from the real copy, which this PR updates to `"Vault provider not found"`. Worth aligning so the mock does not read as stale.

jonybur (approved):

PR #1997, opened 2026-07-01
feat(vault): drop the rebalance

### LINKED ISSUE #1971: Rebalance ("Undersized sacrificial vault") notification has no Figma design — spec it or remove it
## Summary

The liquidation-calculator port (#1970) ships a `rebalance` notification ("Undersized sacrificial vault") that exists **only** in the [`tbv-liquidations`](https://github.com/babylonlabs-io/tbv-liquidations) reference calculator. Searching the full Figma notifications canvas (`6097-3197`) returns **zero matches for "rebalance" / "undersized"** — no frame, no copy, no severity, no action design. It is currently live behind `NEXT_PUBLIC_FF_ENABLE_LIQUIDATION_NOTIFICATIONS` using the reference repo's developer copy. Unlike cliff / too-many-vaults / reorder (all of which have Figma frames), this one is undesigned — so the first question is whether it should exist at all.

## What we have (shipped in #1970, behind the FF)

- **Trigger:** the first liquidation group over-seizes — `overSeizure > protectedBtc`, multi-vault, not a cliff, and not already covered by a cliff/reorder warning.
- **Copy (reference wording, not Figma):** title "Undersized sacrificial vault"; body "Group 1 seizes {X} BTC but target is only {Y} BTC — over-seizure of {Z} BTC. With optimal vault sizes, {W} more BTC would be protected."; suggestion proposes adding a correctly-sized vault (computed amount + order) or re-creating the position.
- **Action:** an "Add a {X} BTC vault" CTA that pre-fills the deposit.
- **Severity:** soft / info. Reachable in practice (e.g. an `[0.80, 0.20]` position triggers it).

## What we miss in Figma

- **No frame exists** — no design for copy, severity, action, or whether to surface it to depositors at all.

## Decision needed (product / design)

1. Should "Undersized sacrificial vault" be a user-facing notification at all? (Narrow case, overlaps cliff/reorder, currently dev copy.)
2. If **yes** → design a Figma frame (copy / severity / action), then align the code.
3. If **no** → remove the `rebalance` warning from the calculator output (drop/suppress the `suggestedRebalance*` payloads).

## References

- Calculator: `services/vault/src/applications/aave/positionNotifications/calculate.ts` (rebalance section), `copy.ts` (`liquidationWarnings.rebalance`)
- Related: #1948, #1949 (cliff — also pending Figma alignment), #1970 (the port)

gbarkhatov:
# Drop the "Undersized sacrificial vault" (rebalance) notification

Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/1971

## What this does

Removes the `rebalance` liquidation notification (shown as "Undersized sacrificial vault") from the position-notification calculator and UI. It was live behind the existing liquidation-notifications feature flag, using placeholder developer copy.

## Why

This notification only ever existed in the `tbv-liquidations` reference calculator we ported from. It has **no Figma design**, it's **not in the simplified product spec**, and it conceptually overlaps the cliff/reorder notifications. Product made the call to drop it. Leaving it in means shipping an undesigned notification with dev copy, so it needs to be removed rather than left behind the flag.

## Approaches considered, and what we chose

**Spec it vs. remove it.** The two options were: (a) design a Figma frame for it and keep it, or (b) remove it. Product chose remove, so this PR deletes it. The reference calculator stays the source of truth for the liquidation *math*; Figma/product decides what we actually *show* — and this one isn't shown.

**The subtle part — two different things are named "rebalance".** There's the user-facing rebalance *warning* (the notification), and there's a small internal *reorder guard* that uses an "over-seizure" condition to avoid suggesting a vault reorder that would make the position worse. Only the first is being dropped. I removed the warning and everything that feeds it, but kept the reorder guard (it's about reorder quality, not this notification) and just renamed its wording away from "rebalance" so the code isn't misleading.

**Delete the dead helpers too.** Removing the warning left a copy string and a result field with no remaining users. Per the repo's no-dead-code rule I deleted those as well (the calculator result no longer carries `suggestedRebalanceVaultBtc`, and the banner's "Add a X BTC vault" label is gone — the cliff path uses "Add sacrificial vault").

## Changes

- `positionNotifications/types.ts` — removed `"rebalance"` from the warning types and the `suggestedRebalanceVaultBtc` result field.
- `positionNotifications/calculate.ts` — deleted the whole rebalance-warning section; kept and reworded the internal over-seizure reorder guard.
- `copy.ts` — removed the rebalance copy block and the now-unused "Add a X BTC vault" label.
- `bannerSeverity.ts`, `BannerActions.tsx`, the debug panel, and a few comments — dropped the rebalance references; the add-a-vault CTA is now cliff-only.
- Tests — removed the rebalance-specific tests/fixtures and updated the calculator tests to the new behavior (see below).

## Behavior change to be aware of

A position whose first liquidation group over-seizes because of sub-optimal vault sizes — e.g. `[0.80, 0.20]` — now shows **no** notification, because rebalance was the only warning that covered that shape. That's the intended tradeoff of dropping it. Everything else (urgent, cliff, reorder, dust, too-many-vaults) is unchanged.

## How to verify

- tsc, lint (0 errors — same warning baseline as `main`), and the full vault test suite pass (2113 tests).
- Manually (debug panel with the liquidation-notifications flag on, Manual Mode): an `[0.80, 0.20]` position no longer shows "Undersized sacrificial vault"; cliff / reorder / dust / urgent / too-many-vaults still behave, and the cliff "Add sacrificial vault" CTA still works.

### DIFF
--- a/services/vault/src/applications/aave/components/PositionNotificationsDebugPanel.tsx
+++ b/services/vault/src/applications/aave/components/PositionNotificationsDebugPanel.tsx
@@ -34,7 +34,6 @@ const SEVERITY_COLORS: Record<BannerSeverity, string> = {
 const WARNING_TYPE_COLORS: Record<WarningType, string> = {
   urgent: "bg-red-600 text-white",
   cliff: "bg-orange-600 text-white",
-  rebalance: "bg-amber-500 text-white",
   reorder: "bg-yellow-500 text-black",
   dust: "bg-gray-500 text-white",
   "weird-params": "bg-blue-500 text-white",
--- a/services/vault/src/applications/aave/positionNotifications/__tests__/calculate.test.ts
+++ b/services/vault/src/applications/aave/positionNotifications/__tests__/calculate.test.ts
@@ -204,7 +204,6 @@ describe("calculate", () => {
     const allowed = new Set([
       "urgent",
       "cliff",
-      "rebalance",
       "reorder",
       "dust",
       "weird-params",
@@ -345,12 +344,13 @@ describe("golden vectors (reference scenario suite)", () => {
     expect(result.groups[1].isFullLiquidation).toBe(true);
   });
 
-  it("B2 — [0.80, 0.20]: over-seizure triggers a rebalance with a suggested vault", () => {
+  it("B2 — [0.80, 0.20]: over-seizure is no longer surfaced (rebalance dropped)", () => {
+    // The first group over-seizes, but with the "Undersized sacrificial vault"
+    // (rebalance) notification removed there is no actionable cliff/reorder
+    // warning for this shape — the position is left un-notified by design.
     const result = calculate(makeParams([v(0.8), v(0.2)]));
     expect(hasWarning(result.warnings, "cliff")).toBe(false);
     expect(hasWarning(result.warnings, "reorder")).toBe(false);
-    expect(hasWarning(result.warnings, "rebalance")).toBe(true);
-    expect(result.suggestedRebalanceVaultBtc).toBe(0.38);
   });
 
   it("C1 — [0.35, 0.65] wrong order: reorder notification, single group", () => {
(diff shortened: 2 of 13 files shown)


gbarkhatov:
@greptile review

jonybur (dismissed):

jeremy-babylonlabs (approved):

PR #1998, opened 2026-07-01
chore: cleanup wallet connect modal

jeremy-babylonlabs:
Screenshot 2026-07-02 at 02 57 27
Screenshot 2026-07-02 at 02 57 19


gbarkhatov:
@jeremy-babylonlabs on you 2nd screenshot 
image
 I see that metamask is overlapping with Unisat, can this be fixed?

packages/babylon-wallet-connector/src/components/Chains/index.tsx
```diff
@@ -54,7 +54,7 @@ export const Chains = memo(
             disabled={disabled}
             fluid
             onClick={onConfirm}
-            className="text-sm"
+            className="text-sm disabled:!bg-[#CCCCCC] disabled:!opacity-100 dark:disabled:!bg-secondary-strokeDark"
```
gbarkhatov: Magic hex color: `disabled:!bg-[#CCCCCC]` hardcodes the light-mode disabled background as a raw hex, while dark mode uses the `secondary-strokeDark` token right beside it — inconsistent. CLAUDE.md's *No Magic Numbers* rule asks for hardcoded colors to be extracted to a named token/constant, and there's precedent in this package (`HARDWARE_DOT = "bg-[#FFB300]"` in `WalletButton`). If no palette token matches this light disabled state, that's a design-system gap worth a token rather than an inline hex. Non-blocking.

packages/babylon-wallet-connector/src/core/wallets/btc/onekey/index.ts
```diff
@@ -7,6 +7,7 @@ const metadata: WalletMetadata<IBTCProvider, BTCConfig> = {
   id: "onekey",
   name: WALLET_PROVIDER_NAME,
   icon: logo,
+  iconBackground: "#1FE121",
```
gbarkhatov: Nit (optional): `#1FE121` is a magic hex — unlike `#000000` (okx/unisat) its meaning isn't self-evident. Since these are config values it's borderline, but a named brand-color constant would read better and stay consistent with the token-extraction pattern used elsewhere (e.g. `HARDWARE_DOT`). Non-blocking.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -41,6 +42,15 @@ const DISABLED_WALLETS: string[] = [
 
 const context = typeof window !== "undefined" ? window : {};
 
+// The wallet dialog is a full-viewport overlay, so its close/settings buttons
+// position with `fixed left`/`right`, not inside the page's 1080px content
+// box. These match that box's edge (per Figma: both inset 180px on the 1512px
```
gbarkhatov: Comment math doesn't match the code: the formula `(100vw-1080px)/2+20px` yields ~236px at a 1512px viewport (content-box edge is (1512-1080)/2 = 216px, +20px), not the "180px" stated here. The 180px figure isn't derivable from either the 1080px box or the formula — worth fixing the comment so it doesn't mislead the next reader (the rendered result is presumably fine).

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -228,6 +238,9 @@ export const WalletConnectionProvider = ({ children }: PropsWithChildren) => {
       disabledWallets={DISABLED_WALLETS}
       requiredChains={["BTC", "ETH"]}
       disableTomo
+      dialogActions={<StandardSettingsMenu theme={theme} setTheme={setTheme} />}
```
gbarkhatov, line 242: Worth a visual check (not blocking): a second `StandardSettingsMenu` is already mounted in `RootLayout.tsx:199`; this adds one inside the full-screen wallet dialog. When the dialog (a `fixed z-10` overlay) is open, please confirm the settings popover opened from this trigger stacks above the dialog content and there's no z-index / double-menu weirdness on desktop.

gbarkhatov (dismissed):
Reviewed the wallet-connect modal cleanup. Verified the load-bearing mechanisms — `Avatar` forwards `style` (so `logoBackground` applies), `AvatarGroup` propagates `variant="circular"` to children via `cloneElement`, `size="large"` is a valid Avatar size, the reworked logo SVGs render via `` so the per-wallet clip-path ids can't collide, and the new color tokens (`primary-contrast`, `accent-secondary`, `secondary-strokeDark`) all resolve through the core-ui preset. No correctness issues found.

Approving. Left a few non-blocking maintainability nits (magic hex colors, plus a comment whose stated inset doesn't match the formula) and one visual check regarding the second settings menu now mounted inside the dialog.

gbarkhatov (approved):

PR #2002, opened 2026-07-02
fix(vault): fix small icon bug

jonybur:
Screenshot 2026-07-02 at 02 23 01

### DIFF
--- a/services/vault/src/components/Activity/LiquidationGroupCard.tsx
+++ b/services/vault/src/components/Activity/LiquidationGroupCard.tsx
@@ -14,6 +14,9 @@ interface LiquidationGroupCardProps {
 export function LiquidationGroupCard({ row }: LiquidationGroupCardProps) {
   const [expanded, setExpanded] = useState(false);
   const [collateralIcon, debtIcon] = row.tokenIcons;
+  // The overlap margin only makes sense when both icons render; on a
+  // collateral-only liquidation it would eat into the gap before the text.
+  const showDebtIcon = Boolean(debtIcon && row.summary.debt);
 
   const summary = row.summary.debt
     ? `${row.summary.collateral.value} ${row.summary.collateral.symbol} / ${row.summary.debt.value} ${row.summary.debt.symbol}`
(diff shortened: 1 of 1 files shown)


services/vault/src/components/Activity/LiquidationGroupCard.tsx
```diff
@@ -22,27 +25,29 @@ export function LiquidationGroupCard({ row }: LiquidationGroupCardProps) {
...
+          <div className="flex shrink-0 items-center">
             <Avatar
               url={collateralIcon}
               alt={row.summary.collateral.symbol}
-              size="small"
-              className="-mr-2"
+              size="large"
+              className={showDebtIcon ? "-mr-3" : undefined}
             />
             {debtIcon && row.summary.debt && (
```
gbarkhatov, line 36: Minor cleanup: `showDebtIcon` (line 19) is defined as `Boolean(debtIcon && row.summary.debt)` — the exact condition inlined here. Using `{showDebtIcon && (` single-sources the "should the debt icon render" rule and keeps the render condition and the overlap margin (line 34) in sync if the predicate ever changes. Non-blocking.

services/vault/src/components/Activity/LiquidationGroupCard.tsx
```diff
@@ -22,27 +25,29 @@ export function LiquidationGroupCard({ row }: LiquidationGroupCardProps) {
...
                 url={debtIcon}
                 alt={row.summary.debt.symbol}
-                size="small"
+                size="large"
               />
             )}
           </div>
-          <div className="flex flex-col gap-1">
+          <div className="flex min-w-0 flex-col gap-1">
             <span className="text-xl leading-none text-accent-primary">
```
gbarkhatov, line 45: Nit: only `summary` gets `truncate` — this classification label (`row.type`) has no overflow guard, so the two spans in this column handle overflow asymmetrically. `row.type` values are short/controlled so it's unlikely to bite; just flagging the asymmetry. Non-blocking.

gbarkhatov (approved):
Reviewed the diff — cosmetic layout fix for `LiquidationGroupCard`: avatars go small→large, the overlap margin is now scoped to the two-icon case, and the summary can truncate instead of pushing the layout. Verified `size="large"` is a valid `Avatar` prop and the `showDebtIcon` guard correctly limits the overlap margin to when both icons render. Approving.

One PR-level nit: the title "fix small icon bug" undersells the change — the avatars are deliberately enlarged (small→large), not just fixed. Worth a line in the description so anyone reading the log later understands the intent. The screenshot presumably confirms the intended look.

Inline notes below are minor and non-blocking.

PR #2003, opened 2026-07-02
fix: borrow repay figma

jeremy-babylonlabs:
Screenshot 2026-07-02 at 16 39 46

### DIFF
--- a/services/vault/src/applications/aave/hooks/useWithdrawCollateralTransaction.ts
+++ b/services/vault/src/applications/aave/hooks/useWithdrawCollateralTransaction.ts
@@ -12,7 +12,6 @@ import { useAccount, useWalletClient } from "wagmi";
 
 import { isWithdrawBlocked } from "@/components/shared/protocolStatus";
 import { getETHChain } from "@/config/network";
-import { useError } from "@/context/error";
 import { useProtocolGateState } from "@/hooks/useProtocolGate";
 import { logger } from "@/infrastructure";
 import {
@@ -33,6 +32,10 @@ export interface UseWithdrawCollateralTransactionResult {
   executeWithdraw: (vaultIds: string[]) => Promise<boolean>;
   /** Whether transaction is currently processing */
   isProcessing: boolean;
+  /** Last failure message, shown inline under the action (null when none). */
+  error: string | null;
+  /** Clear the last failure message (e.g. when the dialog reopens). */
+  clearError: () => void;
 }
 
 /**
@@ -46,12 +49,14 @@ export interface UseWithdrawCollateralTransactionResult {
  */
 export function useWithdrawCollateralTransaction(): UseWithdrawCollateralTransactionResult {
   const [isProcessing, setIsProcessing] = useState(false);
+  const [error, setError] = useState<string | null>(null);
   const { data: walletClient } = useWalletClient();
   const { address } = useAccount();
   const queryClient = useQueryClient();
-  const { handleError } = useError();
   const { markVaultsAsPending } = usePendingVaults();
   const gate = useProtocolGateState();
+
+  const clearError = useCallback(() => setError(null), []);
 
   const executeWithdraw = useCallback(
     async (vaultIds: string[]) => {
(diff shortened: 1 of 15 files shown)


services/vault/src/applications/aave/hooks/useWithdrawCollateralTransaction.ts
```diff
@@ -33,6 +32,10 @@ export interface UseWithdrawCollateralTransactionResult {
   executeWithdraw: (vaultIds: string[]) => Promise<boolean>;
   /** Whether transaction is currently processing */
   isProcessing: boolean;
+  /** Last failure message, shown inline under the action (null when none). */
+  error: string | null;
+  /** Clear the last failure message (e.g. when the dialog reopens). */
+  clearError: () => void;
```
gbarkhatov, line 38: `clearError` is added to the interface + return here, but no consumer calls it. `WithdrawFlow` unmounts on close (`if (!props.open) return null`), so the error state already self-resets on reopen — as written this is dead code (repo ZERO DEAD CODE POLICY). Suggest dropping `clearError` (its interface entry + the `useCallback`) from this hook.

Note the reorder hook's `clearError` is the opposite case — worth *keeping* and actually wiring; see the comment on `useReorderModal.ts`.

services/vault/src/components/simple/PositionNotificationBanner/PositionNotificationBanner.tsx
```diff
@@ -238,7 +243,16 @@ export function PositionNotificationBanner({
   // secondary warnings (e.g. urgent + cliff).
   let suggestion: ReactNode;
   if (isStandaloneReorder && result.optimalVaultOrder) {
-    suggestion = <OptimalOrderChips vaults={result.optimalVaultOrder} />;
+    suggestion = (
+      <div className="flex flex-col gap-2">
+        <OptimalOrderChips vaults={result.optimalVaultOrder} />
+        {reorderError && (
```
gbarkhatov, line 249: This callout only renders inside the `isStandaloneReorder` branch, but `buildBannerActions` also pushes the "Apply Optimal Order" CTA as a *secondary* action on urgent/cliff banners (`showApplyOrder = suggestReorder && result.optimalVaultOrder !== null`). That CTA runs the same `handleApplyOrder → executeReorder`, which sets `reorderError` on failure — yet in the non-standalone branch nothing renders it.

Since this PR removes the global error modal, a failed reorder triggered from an urgent/cliff banner drops the user back into the same UI with no feedback. Consider surfacing `reorderError` in the `else` branch too (wherever the reorder CTA can appear), not just the standalone one.

services/vault/src/components/simple/ReorderVaults/useReorderModal.ts
```diff
@@ -32,7 +34,7 @@ export function useReorderModal({
   const [baselineVaultIds, setBaselineVaultIds] = useState<readonly Hex[]>(() =>
     vaults.map((v) => v.vaultId as Hex),
   );
-  const { executeReorder, isProcessing } = useReorderVaults();
+  const { executeReorder, isProcessing, error } = useReorderVaults();
```
gbarkhatov, line 37: `error` is surfaced from `useReorderVaults` but never cleared on reopen. `CollateralSection` keeps `ReorderVaultsModal` mounted and only toggles `isOpen`, and the `isOpen` effect just below re-snapshots the vault order without touching `error`. After a wallet rejection/failure, closing and reopening the modal shows the stale "Transaction failed" callout before the user has retried.

Suggest also pulling `clearError` from `useReorderVaults` and calling it inside the `if (isOpen)` effect — this is exactly what that hook's `clearError` was intended for (and resolves its otherwise-unused status).

gbarkhatov (approved):
Reviewed the diff (no branch switch). The borrow/repay Figma fixes and the error-modal → inline-callout migration look solid and correctly mirror the existing `useBorrowTransaction`/`useRepayTransaction` pattern. APR wiring (keys align on `reserveId.toString()`, graceful `{}`-on-error), the copy-key removals, and the test-mock cleanup all check out, and leaving `GeoFencingProvider` on the modal is the right call. Approving. Three non-blocking notes on error surfacing / dead code inline below.

PR #2007, opened 2026-07-06
feat(vault): expand god mode to walk the deposit flow and simulate activation

jonybur:
Lets a dev/QA mock a deposit and click through the real deposit multistepper with all external effects simulated:

- Mock "ready to activate" vaults open the activation walk (gate -> simulated submit -> VaultActivatedView) with no wallet/registry/contract calls; every other flow step opens a read-only progress preview driven by the panel slider.
- Panel restructured into Normal / Expired / Different-wallet modes, with the slider scrubbing within the selected mode.
- Adds a Light / Dark / System theme toggle to the panel.

All demo branches are import.meta.env.DEV-gated and tree-shaken from production; the real activation machinery gains no branches.


gbarkhatov (dismissed):

kirugan (dismissed):

services/vault/src/hooks/deposit/useArtifactDownload.ts
```diff
@@ -111,6 +115,15 @@ export function useArtifactDownload(options?: {
 
       const normalizedPeginTxid = stripHexPrefix(peginTxid);
 
+      // Dev/QA-only simulation of the artifact fetch (never reaches a VP),
+      // so the dialogs' progress states can be exercised. Opted into via the
+      // god-mode panel's "Mock artifact download" toggle; off in production
+      // builds, where the god-mode gate is compile-time false.
+      const demoDownload = isArtifactDownloadDemoEnabled();
```
jrwbabylonlab, line 122: The mock still falls through to `markArtifactsDownloaded(vaultId)` on the shared success path, and CollateralSection resolves clicks against all `collateralVaults` when the toggle is on — so mocking a download on a real vault permanently satisfies the risk-ack gate without a file ever saved.

services/vault/src/components/simple/PostDepositContinuationContent.tsx
```diff
@@ -31,13 +32,29 @@ export function PostDepositContinuationContent({
...
   const scoped = useMemo(() => {
     const ids = new Set<string>(vaultIds);
+    const demoActivities = demo
+      ? demo.pendingActivities.filter((a) => ids.has(a.id))
+      : [];
     return {
-      activities: activities.filter((a) => ids.has(a.id)),
+      // Demo activity ids can never collide with real ones (distinct id space),
+      // so the concat introduces no duplicates.
+      activities: [
```
jrwbabylonlab: Merging demo activities into the polling provider's list feeds them to `useSigningRequiredNotifications` — a demo ready-to-activate state can fire a real desktop notification.

services/vault/src/components/simple/PostDepositContinuationView.tsx
```diff
@@ -261,6 +275,56 @@ export function PostDepositContinuationView({
...
+              activity={activity}
+              siblingVaultIds={siblingVaultIds}
+              onClose={onClose}
+            />
+          </Suspense>
+        </ActivationGate>
+      );
+    }
+    const demoStep = pollingResult?.displayStepOverride ?? waitStep;
+    const { perVaultSteps: demoPerVaultSteps } = deriveSplitVaultProgress(
```
jrwbabylonlab, line 309: Sibling columns here derive steps from `peginState` only, so in a batched demo they ignore `displayStepOverride` (unlike `demoStep` one line up) and fall back to the base state.

jrwbabylonlab (dismissed):

services/vault/src/hooks/deposit/useArtifactDownload.ts
gbarkhatov: Confirmed on the current branch — still unresolved. The demo path only swaps the fetch fn (`fetchArtifacts = demoFetchAndDownloadArtifacts`) but then falls through to the shared success block at `useArtifactDownload.ts:238-241`, which runs `if (vaultId) markArtifactsDownloaded(vaultId)` — a write to `artifactDownloadStorage` (persistent). Because `CollateralSection.handleArtifactDownload` resolves clicks against all `collateralVaults` when the toggle is on, and `isArtifactDownloadDemoEnabled()` is global (not per-vault), a QA click on a *real* collateral vault runs the simulated fetch and permanently flips that real vault's risk-ack gate to `downloaded` with no file ever saved — and it survives god-mode being switched off. This is the one finding of the three that mutates state a real user would later see. Suggested fix: skip `markArtifactsDownloaded(vaultId)` when `demoDownload` is true, since the mock never produced a file to gate on.

services/vault/src/components/simple/PostDepositContinuationContent.tsx
gbarkhatov: Confirmed — still unresolved. `PostDepositContinuationContent` passes `scoped.activities` (demo ∪ real) into the nested `PeginPollingProvider`, which mounts `useSigningRequiredNotifications(activities, ...)` at `PeginPollingContext.tsx:452`. That hook iterates the activities and pushes a notification for any owned candidate vault whose `availableActions` include an actionable one — and `ACTIVATE_VAULT` is present in `ACTION_NOTIFICATION_COPY` (`useSigningRequiredNotifications.ts:35`, consumed at lines 64-68) — so a demo ready-to-activate vault qualifies. It only surfaces when the signing-notification flag is on and the tab is hidden, but that means opening a mock deposit and switching tabs can fire a real desktop notification for a vault that does not exist. Suggested fix: exclude `demoVaultIds` from the activities handed to the notification observer (or have the observer skip demo ids).

services/vault/src/components/simple/PostDepositContinuationView.tsx
gbarkhatov: Confirmed — still unresolved, cosmetic / dev-only. The driven column uses `demoStep = pollingResult?.displayStepOverride ?? waitStep`, but `deriveSplitVaultProgress` resolves every non-active sibling purely from `getPollingResult(id)?.peginState` via `getPeginDisplayStep` (`useSplitVaultProgress.ts:77`, then 85-88) and never consults that sibling's own `displayStepOverride`. So in a batched demo the driven column scrubs with the slider while the sibling columns stay pinned to their base `peginState` step. Impact is confined to the god-mode split preview. Suggested fix: fold `displayStepOverride` into the value `getPeginDisplayStep` sees for a demo polling result, so both the active and sibling columns derive their step from one source.

gbarkhatov (approved):

PR #2010, opened 2026-07-07
feat(vault): set correct download modal

jonybur:
## What

Aligns the artifact-download and vault-activation dialogs with the latest Figma designs, and removes a redundant step from the activation flow.

### Activate your BTC Vault dialog (activation flow)
- Body copy updated per design, with the "download the recovery artifacts" instruction emphasized in the primary text color.
- The card's action chip shows **Sign Transaction** while the cold-cache auth prime waits on the BTC wallet's signature prompt, then flips to the fetch/progress states.
- New downloaded state per design: document-with-checkmark icon, **"Artifacts downloaded"** heading, "Your files are stored locally and never uploaded." body, risk checkbox hidden, **Activate vault** enabled.
- **Activate vault** is disabled while a download streams — confirming would unmount the dialog and abandon the in-flight transfer.
- Button casing normalized to sentence case ("Activate vault").

### Download BTC Vault artifacts dialog (collateral entry)
- Downloading state shows the document-with-download-arrow icon per design.

### Both dialogs
- No X close button — these are inner-flow dialogs; dismissal goes through the footer actions (Escape/backdrop still cancel any in-flight download and close).
- Footer buttons use the design's 14px label / 16px side padding (`size="medium"` + 40px height).
- The 90x90 state icon is extracted into a shared `ArtifactModalIcon` (pending / downloading / downloaded glyphs).

### Flow: in-step artifact callout removed
The `vault-artifacts.json (Recommended) / Skip / Download Artifacts` interstitial after confirming activation is gone — it duplicated the download and risk acknowledgement the confirmation dialog already owns. `ActivationGate` now proceeds straight from confirmation to the activation step.

### Download progress formatting
- Byte counts use decimal (SI) units with rounded rollover so ~1 GB renders as "1.00 GB" (never "954 MB" or "1000 MB"), matching the design and the "~1 GB" card copy.
- The " / " separator renders in the primary text color per design; received bytes and percent stay clamped to the total.

### Cancel-restart race fix (`useArtifactDownload`)
Staleness is now derived from each download invocation's own `AbortController` instead of a shared flag. Previously, cancelling a flow parked in a non-abortable await (wallet prime, the 10s VP retry sleep) and restarting could resurrect the old flow and clobber the new download's state; a stale flow now stays permanently dead, and the same check feeds the service's `isCancelled` poll so its stream stops too.

## Testing

- New test suites for the download dialog states and the card's byte-progress panel (formatting, clamps, indeterminate fallback).
- Hook tests for the Sign Transaction status and the cancel-then-restart race; activate/gate suites updated for the new states.
- Full vault suite, lint, and typecheck pass.


services/vault/src/copy.ts
```diff
@@ -465,6 +483,9 @@ export const COPY = {
         "Cannot authenticate with the vault provider. Please refresh and try again.",
       // Progress/status lines surfaced in the card while the download hook
       // works through its fetch / re-auth / wait-for-signatures states.
+      // The signature status shows while the cold-cache auth prime waits on
+      // the BTC wallet's signature prompt.
+      signTransaction: SIGN_TRANSACTION_LABEL,
```
gbarkhatov, line 488: Non-blocking nit: this reuses the button label `SIGN_TRANSACTION_LABEL` ("Sign Transaction", title case) as a **status** line for the recovery-artifacts card. The copy conventions at the top of this file call for sentence-case status labels, so strictly as a status it would read "Sign transaction". Sharing one constant keeps the stepper button and this status in lockstep, so it's a reasonable trade-off -- just flagging the casing in case you'd prefer a dedicated sentence-case status string here.

gbarkhatov (approved):
Approve. I re-read the full diff plus the enclosing functions and callers: the `useArtifactDownload` cancel-staleness rewrite (per-invocation `AbortController` instead of a shared flag), the SI-unit `formatBytes` rollover, the new `!isDownloading` activation guard, the `body` copy shape change (string -> `EmphasisBodySegment[]`), and the `ActivationGate` flow simplification are all correct and well-covered by the new/updated tests. No orphaned references to the deleted `InStepArtifactCallout` or the removed copy keys.

I also evaluated a set of follow-up findings. Disposition:

- **Signal doc-comment (`artifactDownloadService.ts`):** no action. That file isn't part of this PR, and its current comment already documents the exact behavior the finding recommends -- that `signal` does not cover the body stream and that the `isCancelled` poll between chunks is the streaming-phase cancel mechanism.
- **`PROGRESS_BAR_FILL_COLOR` inline hex:** no action. Unchanged pre-existing line, documented, and required because `ProgressBar` takes a raw CSS color rather than a class.
- **`RecoveryArtifactsCard` mounted in the downloaded state:** no action. Intended -- it renders the green "downloaded" card per the design.
- **"Sign Transaction" status casing:** left a single non-blocking inline note on `copy.ts`.

Nothing blocks merge.

PR #1943, opened 2026-06-23
fix(vault): validate large artifact responses by envelope prefix

kirugan:

PR #2020, opened 2026-07-08
Address comments from another PR

kirugan:
Comments from this PR https://github.com/babylonlabs-io/babylon-toolkit/pull/1943 have been addressed

### DIFF
--- a/services/vault/src/services/artifacts/__tests__/artifactDownloadService.test.ts
+++ b/services/vault/src/services/artifacts/__tests__/artifactDownloadService.test.ts
@@ -40,16 +40,24 @@ function responseFor(body: unknown): Response {
  * Build a Response backed by a real ReadableStream so the streaming path
  * (readBodyWithProgress) is exercised, with Content-Length present only when
  * asked. A stream body never auto-populates Content-Length, which lets us
- * test the header-absent fallback deterministically.
+ * test the header-absent fallback deterministically. `chunkSizeBytes` splits
+ * the body across multiple stream chunks (default: one chunk) so multi-chunk
+ * assembly paths can be exercised.
  */
 function streamingResponse(
   body: string,
-  { withContentLength = false }: { withContentLength?: boolean } = {},
+  {
+    withContentLength = false,
+    chunkSizeBytes,
+  }: { withContentLength?: boolean; chunkSizeBytes?: number } = {},
 ): Response {
   const bytes = new TextEncoder().encode(body);
   const stream = new ReadableStream<Uint8Array>({
     start(controller) {
-      controller.enqueue(bytes);
+      const size = chunkSizeBytes ?? bytes.byteLength;
+      for (let offset = 0; offset < bytes.byteLength; offset += size) {
+        controller.enqueue(bytes.subarray(offset, offset + size));
+      }
       controller.close();
     },
   });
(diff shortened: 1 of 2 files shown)


services/vault/src/services/artifacts/artifactDownloadService.ts
```diff
@@ -341,9 +341,14 @@ function validateArtifactPayload(
...
- * `error` rejects (even alongside a `result` key), and a top-level `result`
- * that is absent from the prefix rejects fail-closed.
+ * Within the prefix the verdict matches the parsed small-response path: any
+ * non-null top-level `error` rejects (even alongside a `result` key), and a
+ * missing top-level `result` rejects. The match is not exact under
+ * truncation, though: an envelope that declares `result` early and pads a
+ * top-level `error` past the prefix window passes here where the parsed
+ * path would reject. That divergence adds nothing exploitable — the body is
+ * not validated on this path, so a malicious VP already passes with a plain
+ * `{"result": <garbage>}` envelope (see below).
```
kirugan, line 351: Addressing [this comment](https://github.com/babylonlabs-io/babylon-toolkit/pull/1943#discussion_r3543075232) from #1943:

> Minor: the "mirrors the parsed small-response path / fails closed" claim is not exact. For `{"result":null, "padding":"<70 KiB…>", "error":{…}}` the scan sees `result` early and never reaches the `error` pushed past the 64 KiB window, so it passes — while the parsed small-path rejects the same envelope on `error != null`. So truncation fails *closed* only when `result` is absent; in this shape it fails *open* relative to the parsed path. Non-exploitable […], but worth softening the wording so a future reader does not over-trust the invariant.

Softened in both places: this JSDoc now says the parsed-path match holds only within the prefix window and spells out exactly that early-`result` / padded-`error` divergence (non-exploitable since the body is not validated on this path), and the `scanEnvelopePrefix` doc now says truncation fails closed only for `result`.

services/vault/src/services/artifacts/__tests__/artifactDownloadService.test.ts
```diff
@@ -131,6 +139,55 @@ describe("fetchAndDownloadArtifacts", () => {
     expect(triggerDownloadSpy).toHaveBeenCalledTimes(1);
   });
 
+  it("triggers download for a body larger than the validation prefix when result precedes the cut", async () => {
```
kirugan, line 142: Addressing [this comment](https://github.com/babylonlabs-io/babylon-toolkit/pull/1943#discussion_r3543075237) from #1943:

> The large-payload fixtures top out at ~8 KB (`"ab".repeat(4096)` = 8192 B), well below `PREFIX_VALIDATION_BYTES` (65536). So the truncation branches this PR adds — the `chunk.subarray(0, remaining)` clamp in `readBodyWithProgress` and the `subarray(0, PREFIX_VALIDATION_BYTES)` slice in `validateLargePayloadEnvelopePrefix` — never execute under test. […] Worth adding one fixture with a body > 64 KiB: `result` key before the cut → downloads; a padded error whose top-level `error` sits past the cut → rejects. That same fixture also pins down the both-fields-past-prefix edge noted on the source file.

Added both: a >64 KiB success envelope streamed in 10 000-byte chunks (`result` before the cut → downloads), which drives the prefix-copy loop and the `subarray` clamp across chunk boundaries; and an error envelope with both top-level keys padded past the prefix → rejects fail-closed on missing `result`, pinning the both-fields-past-prefix edge.

gbarkhatov (approved):

PR #2021, opened 2026-07-08
feat(vault): gate the v3 UI behind an ENABLE_V3_UI feature flag

### LINKED ISSUE #2014: [v3] ENABLE_V3_UI feature flag
**Area:** `services/vault` (`src/config/featureFlags.ts`, `src/router.tsx`, `RootLayout`)

## Context

TBV v3 is a full UI restructure landing across roughly forty child issues over many PRs. It must be gated so `main` keeps shipping today's v2 dashboard while v3 lands incrementally, and so v3 can be switched on per environment (devnet → testnet → mainnet) without a code change.

Follows the existing convention in `services/vault/src/config/featureFlags.ts` — `NEXT_PUBLIC_FF_` prefix, opt-in `=== "true"` semantics, defaults to `false`. Same shape as `isLiquidationNotificationsEnabled`.

Unlike the `DISABLE_DEPOSIT` / `PROTOCOL_PAUSED` kill-switches, this flag is temporary scaffolding: it exists only until v3 becomes the default, and it must then be deleted along with the v2 shell.

## Scope

- [ ] Add an `isV3UiEnabled` getter (`NEXT_PUBLIC_FF_ENABLE_V3_UI`) to `services/vault/src/config/featureFlags.ts`, with the standard Purpose / Why needed / Default doc block
- [ ] Single switch at the shell seam: flag **off** → current `Header` + top-nav + single-page dashboard; flag **on** → sidebar shell + per-page routes
- [ ] Guard the v3-only routes (`/vaults`, `/loans`, `/liquidations`) at the router, not merely hide the nav items — a direct load must redirect when the flag is off
- [ ] Add the variable to `services/vault/.env.example` and the per-environment `.env.*` files; hand off to DevOps for mainnet
- [ ] File the flag-retirement issue (remove v2 shell + flag) as the last item of the v3 milestone

## Acceptance criteria

- [ ] Flag unset or `false` → app behavior is identical to today's v2: no sidebar, no v3 routes reachable, no v3 code paths executed
- [ ] Flag `true` → the v3 shell and pages render in light and dark
- [ ] Flipping the flag requires only an env change, no rebuild of anything else; no other feature flag changes behavior

## Depends on

Nothing — this is the first v3 issue to land.

jeremy-babylonlabs:
Add the isV3UiEnabled getter (NEXT_PUBLIC_FF_ENABLE_V3_UI) following the existing opt-in convention, and guard the v3-only sections at the router so they are unreachable while the flag is off.

The v3 sections are guarded as whole subtrees (`vaults/*` rather than `vaults`), so a direct load or a stale deep link redirects to the v2 dashboard instead of falling through to the 404 route once the nested v3 routes land.

Unlike the DISABLE_DEPOSIT / PROTOCOL_PAUSED kill-switches this flag is temporary scaffolding: it keeps main shipping the v2 dashboard while v3 lands incrementally, and is removed together with the v2 shell once v3 becomes the default.

The v3 shell and its pages are out of scope here and land behind this flag in the follow-up issues.

Closes #2014

### DIFF
--- a/services/vault/src/__tests__/router.test.tsx
+++ b/services/vault/src/__tests__/router.test.tsx
@@ -7,6 +7,8 @@
  * 2. The reserve detail (/app/aave/reserve/:reserveId) is an overlay on top of
  *    the dashboard, not a sibling route that replaces it. The dashboard must
  *    stay mounted underneath so opening the overlay never blanks the page.
+ * 3. The v3-only sections are reachable only when ENABLE_V3_UI is on. With the
+ *    flag off a direct load of one of them redirects to the v2 dashboard.
  *
  * These tests lock in that wiring so a future router refactor can't silently
  * regress it.
@@ -16,7 +18,21 @@ import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { render, screen, waitFor } from "@testing-library/react";
 import type { ReactNode } from "react";
 import { MemoryRouter, Outlet } from "react-router";
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const featureFlagsState = vi.hoisted(() => ({ isV3UiEnabled: false }));
+
+vi.mock("@/config/featureFlags", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@/config/featureFlags")>();
+  return {
+    default: Object.create(actual.default, {
+      isV3UiEnabled: {
+        get: () => featureFlagsState.isV3UiEnabled,
+        enumerable: true,
+      },
+    }),
+  };
+});
 
 vi.mock("../components/pages/RootLayout", () => ({
   default: () => <Outlet context={{ openDeposit: () => {} }} />,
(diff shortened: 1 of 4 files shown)


services/vault/src/__tests__/router.test.tsx
```diff
@@ -16,7 +18,21 @@ import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { render, screen, waitFor } from "@testing-library/react";
 import type { ReactNode } from "react";
 import { MemoryRouter, Outlet } from "react-router";
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const featureFlagsState = vi.hoisted(() => ({ isV3UiEnabled: false }));
+
+vi.mock("@/config/featureFlags", async (importOriginal) => {
```
jrwbabylonlab, line 25: The `importOriginal` + prototype-delegation setup is heavier than needed — the plain `vi.hoisted` object pattern from protocolStatus.test.ts / useApplicationCap.test.tsx behaves the same here, since everything else reading the flags is already mocked.

jrwbabylonlab (approved):

services/vault/src/router.tsx
```diff
@@ -29,6 +31,9 @@ const AaveReserveDetail = lazyWithRetry(() =>
   importAaveReserveDetail().then((m) => ({ default: m.AaveReserveDetail })),
 );
 
+// Guarded as whole subtrees, so a v3 deep link redirects rather than 404s.
+const V3_ROUTE_PATHS = ["vaults", "loans", "liquidations"] as const;
```
gbarkhatov, line 35: Non-blocking: these three names are a hardcoded guess of the v3 top-level structure. If a v3 section lands under a different name (e.g. `portfolio`), its stale deep links would 404 rather than redirect while the flag is off. Acceptable as scaffolding, but worth noting in the follow-up issue so this list is updated alongside the real routes — it's the single sync point for the guard.

services/vault/src/__tests__/router.test.tsx
```diff
@@ -183,3 +199,54 @@ describe("Router — reserve detail is an overlay over the persistent dashboard"
     );
   });
 });
+
+describe("Router — v3-only routes are gated on the ENABLE_V3_UI flag", () => {
+  const V3_ROUTE_PATHS = ["vaults", "loans", "liquidations"];
```
gbarkhatov, line 204: Non-blocking: this literal duplicates `V3_ROUTE_PATHS` in `router.tsx`. Since the router's constant isn't exported, adding a fourth v3 section there would guard it in production but silently leave it uncovered here — the parametrized tests would still pass while the new path goes untested. Exporting the constant from `router.tsx` and importing it here closes the drift gap.

gbarkhatov (approved):
Approving — clean, well-tested flag gate. Route ranking makes `vaults/*` win over `*` regardless of declaration order, `replace` keeps stale deep links out of history, and the subtree guard (`vaults/*` vs `vaults`) matches the stated intent. Two non-blocking maintainability notes inline.

Non-blocking (out of scope, pre-existing): `components/pages/not-found.tsx` inlines the user-facing strings "Page not found" and "Go back home", which the CLAUDE.md copy policy would place in `copy.ts`. Not introduced by this PR — just flagging for a future cleanup.

jonybur (approved):
**Adversarial review — ✅ Approve**

Clean and well-scoped. The `isV3UiEnabled` getter follows the existing opt-in convention in `featureFlags.ts`, the v3 sections are guarded as whole subtrees (`path/*` → redirect, not fall-through 404), and the router tests cover flag-on/flag-off plus deep-link cases. Nothing to change.

One forward-looking note for the follow-ups (not this PR): #2029 later narrows these subtree guards to exact-path routes, which drops the "deep link redirects rather than 404s" guarantee this PR deliberately adds. Flagged on that PR.

PR #2022, opened 2026-07-08
fix(vault): stop silent zero-balance on failed UTXO fetch / wrong-network session

jonybur:


packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -253,6 +293,10 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
         }
 
         if (newAddress !== address) {
+          // A mid-session account change can also be a network switch; refuse a
+          // foreign-network address (the catch below disconnects) rather than
+          // committing it to the network-scoped balance queries.
+          assertConfiguredNetwork(newAddress);
```
gbarkhatov: Mid-session teardown here is incomplete. When this guard throws, the `catch` at :319 calls the **local** `disconnect()` (defined at :154) — which only clears React state and fires `onDisconnect`; it never calls `btcConnector.disconnect()`. So unlike the connect path (`await btcConnector?.disconnect()` at :204, which fires the connector `disconnect` event that clears `accountStorage` + wallet-widget state), a mid-session network switch leaves `btcConnector.connectedWallet` and the persisted `accountStorage` entry intact and the wallet widget still showing connected. The wrong-network session then auto-reconnects on the next reload (only caught properly there). This contradicts the comments at :192 and :296 ("the catch below disconnects … clears the persisted entry and widget state via the disconnect event"), which hold only for the connect path. The visibility path at :422 has the same gap.

Root cause: the guard + address-adoption logic is duplicated across four call sites with divergent error handling. A single "adopt this address" helper that owns both the check and the full teardown would prevent this drift.

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -392,7 +439,7 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
       console.error("BTC wallet connection check failed:", error instanceof Error ? error.message : "Unknown error");
       disconnect();
```
gbarkhatov, line 456: This `catch` disconnects but never calls `callbacks?.onError?.` — unlike the connect (`:203`), account-change (`:322`), and reconnect (`:592`) paths. A wrong-network switch detected on tab refocus therefore tears the session down with zero user-facing signal (and since `address` is cleared to `""`, the new `balanceLoadError` alert can't render either). The new network test asserts "onError is the only user-facing signal for a refused address" — that guarantee is violated on exactly this path. Consider reporting the error before `disconnect()` here.

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -132,6 +132,25 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
...
+  // reconnecting after the dApp was closed, or a mid-session switch), a
+  // foreign-network address would otherwise be adopted and read a silent zero
+  // balance. Prefix-level check only (mainnet `bc1` vs signet/testnet `tb1`);
+  // signet and testnet share an encoding and cannot be told apart this way.
+  // Throws on mismatch so each caller routes the failure through its own error
+  // handling (connect refuses + disconnects; the mid-session paths disconnect
+  // via their catch; reconnect re-throws to its caller).
+  const assertConfiguredNetwork = useCallback(
+    (candidateAddress: string) => {
+      if (!btcConnector) return;
```
gbarkhatov: Fails open: when `btcConnector` is null this returns without validating, so a wrong-network address is adopted unchecked. In `connectBTC` the connector is guaranteed present, but the reconnect/mid-session paths read it via `useChainConnector`, where it can momentarily be null. Per the repo's "No Silent Fallbacks on Critical Paths" rule, a value/security check that can't run should throw rather than silently pass. Low likelihood, but cheap to make fail-closed.

packages/babylon-wallet-connector/src/providers/BTCWalletProvider.tsx
```diff
@@ -132,6 +132,25 @@ export const BTCWalletProvider = ({ children, callbacks }: BTCWalletProviderProp
...
+      validateAddress(btcConnector.config.network, candidateAddress);
```
gbarkhatov: Scope note (the comment above already acknowledges this): `validateAddress` only compares the common prefix, and TESTNET and SIGNET both use `tb1`. So a signet-configured app + testnet wallet (or vice-versa) passes this guard, the address is adopted, and the mempool query succeeds against the wrong network — a wrong/zero balance with no `utxoError`. The fix covers mainnet↔test-prefix but not signet↔testnet; worth stating explicitly for QA. It also rejects legacy/P2SH `1…`/`3…`/`m…`/`2…` addresses on the *correct* network — fine given the app is taproot/native-segwit-only, just confirm no supported provider defaults to those.

services/vault/src/components/simple/DepositForm.tsx
```diff
@@ -512,7 +527,13 @@ export function DepositForm({
...
           isVerifyingWallet ||
           isVaultCapReached ||
-          vaultCountCapUnavailable
+          vaultCountCapUnavailable ||
+          // A failed balance fetch can leave stale UTXOs in place (React Query
+          // keeps prior data on a refetch error), so the CTA could otherwise
+          // stay enabled while the balance alert says deposits are unavailable.
+          // Block submission until the balance reloads — depositing against a
+          // stale/unknown UTXO set risks an underfunded transaction.
+          !!utxoError
```
gbarkhatov, line 536: `utxoError` is any error from `useUTXOs`, which has `staleTime: 30_000` and no `placeholderData` — so on a failed **background** refetch React Query keeps the last-good `data` while setting `error`. In that case the form shows the real balance *and* the alert *and* disables the CTA, even though a valid ~30s-old UTXO set is present. And the copy (`copy.ts:514`) says "Refresh the page to try again — deposits are unavailable until your balance loads," but the query self-heals on the next focus/interval refetch with no refresh. Consider gating on "error with no usable data" (e.g. `error && !data`) and softening the copy so a transient blip doesn't block an otherwise-fundable deposit.

gbarkhatov (dismissed):
Approving — the two guards are wired correctly (interfaces, prop chain, `config.network` used as the static configured network, and the single `DepositForm` construction site all check out) and the mechanisms are sound. The findings are about the wrong-network guard's teardown being inconsistent across its four call sites and the UTXO-error surface being over-eager; none block merge. Inline comments below.

A few notes that don't map cleanly onto a changed line:

- **Scope of the title.** "stop silent zero-balance" covers the thrown-error and mainnet↔test-prefix cases, but an empty-but-successful fetch (`[]` from the mempool API against a wrong-but-same-prefix network, e.g. signet↔testnet) is still indistinguishable from an empty wallet — arguably the more common config-drift failure. Worth noting the limitation in the PR description.
- **`onError` context inconsistency (nit).** The account-change catch (`BTCWalletProvider.tsx:322`) calls `callbacks?.onError?.(error)` with no context object, while connect (`:203`) and reconnect (`:592`) pass `{ address }`. A consumer keying off the second arg won't see which address triggered a mid-session network switch.
- **Pre-existing, outside this diff.** `validateAddress` (`core/utils/wallet.ts:98-105`) destructures `NETWORKS[network]` *before* the `!(network in NETWORKS)` check, so an out-of-enum network throws a raw `TypeError` instead of the intended `UNSUPPORTED_NETWORK` `WalletError`, leaving that branch unreachable. This guard now routes network values through it — worth a follow-up reorder.

Issue #2017, opened 2026-07-08
[v3] Route restructure


jeremytsngtsng:
**Area:** `services/vault` (`src/router.tsx`)

**Design:** [TBV v.3 – Premium Design](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-) (structural — no visual component)

## Context

Today the app has `/` (the full dashboard, with the Aave reserve detail rendered as an overlay) and `/activity`. v3 promotes each dashboard section to its own page.

## Scope

- [ ] Routes: `/` Overview, `/vaults`, `/loans`, `/liquidations`, `/activity` (exists), plus the Explore destination
- [ ] Preserve the Aave reserve-detail overlay behavior
- [ ] All new routes gated by the v3 feature flag

## Open question

⚠️ **Explore has no defined destination** in code or in the design. Is it an external link, the borrowing-markets page, or a new in-app page? Product decision needed before the nav item is wired.

## Acceptance criteria

- [ ] Each sidebar item deep-links to its page; refresh and direct load work
- [ ] With the v3 flag off, the new routes are not reachable

## Depends on

- #2015

Issue #2016, opened 2026-07-08
[v3] Page-title header


jeremytsngtsng:
**Area:** `services/vault` (uses the `core-ui` `Header`)

**Design:** [Overview (dark)](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10084-22951)

## Context

The header currently carries the logo plus the horizontal nav. v3 moves navigation into the sidebar; the header shows the current page title on the left and keeps the network badge, wallet connect, and settings menu on the right.

## Scope

- [ ] Per-page title ("Overview", "Vaults", …) driven by the active route
- [ ] Retain the network badge, `Connect`, `StandardSettingsMenu`, and the `Deposit` action

## Acceptance criteria

- [ ] Title updates per route; right-side actions unchanged in behavior
- [ ] Matches Figma in light and dark

## Depends on

- #2015

Issue #2013, opened 2026-07-08
[v3] App Shell, Navigation & Theming


jeremy-babylonlabs:
The frame every v3 screen shares. Replaces today's horizontal top-nav with a persistent left sidebar + page-title header, and restructures routes so each section becomes its own page.

**Every v3 surface ships behind the feature flag in #2014** — `main` stays deployable on the current v2 dashboard throughout the build.

**Design:** [TBV v.3 – Premium Design](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-) — see [Overview (dark)](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10084-22951) and [Overview (light)](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10084-22592).

**Area:** `services/vault`, plus `packages/core-ui` for the shared `Sidebar` primitive.

## Children

- [ ] #2014 — `ENABLE_V3_UI` feature flag
- [ ] #2015 — Build left sidebar navigation rail
- [ ] #2016 — Page-title header
- [ ] #2017 — Route restructure
- [ ] #2018 — Conditional "Liquidations" nav entry
- [ ] #2019 — Light/dark theming audit

## Scope

- Persistent left sidebar (Overview · Vaults · Loans · Activity · Liquidations · Explore) replacing the top-nav
- Header reduced to page title + network badge, wallet, settings
- Each dashboard section promoted to its own route
- Both themes first-class on every new surface

## Acceptance criteria

- [ ] With the v3 flag off, the app is indistinguishable from today's v2 dashboard
- [ ] With the flag on, the sidebar shell renders on every page and each nav item deep-links to a working route
- [ ] Every v3 surface verified against Figma in light and dark

PR #2023, opened 2026-07-09
feat(vault): 2015 left sidebar navigation rail

### LINKED ISSUE #2015: [v3] Build left sidebar navigation rail
**Area:** `packages/core-ui` + `services/vault`

**Design:** [Overview (dark)](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10084-22951) · [Overview (light)](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10084-22592)

## Context

No sidebar exists today; navigation lives in the horizontal `Nav` inside `core-ui`'s `Header`. v3 needs a vertical rail (~240px, full height) with a brand lockup at the top, icon + label nav items, and an active state.

The Figma `Sidebar` component is used on every v3 screen, so it belongs in `core-ui` rather than the vault app.

## Scope

- [ ] New shared `Sidebar` component in `core-ui` — icon, label, active/hover states, brand lockup slot
- [ ] Nav items: Overview · Vaults · Loans · Activity · Liquidations · Explore
- [ ] Wire into the `services/vault` layout; retain the existing footer/social block

## Acceptance criteria

- [ ] Sidebar renders on every page in light and dark; the active item reflects the current route
- [ ] Spacing and iconography match Figma

## Depends on

- #2014

jeremy-babylonlabs:
Closes #2015. Part of #2013.

## Screenshot
Screenshot 2026-07-09 at 11 53 12
Screenshot 2026-07-09 at 11 53 03

## What this PR does

Adds the v3 left sidebar navigation rail, replacing the horizontal top-nav for the v3 app shell:

   - **New `Sidebar` / `SidebarItem` / `SidebarBrandLockup` components in `core-ui`** — a router-agnostic 242px rail (brand lockup, six nav items in two groups, a bottom block with social links + Terms of Use / Privacy Policy) composed by the app with `react-router`'s `NavLink`, matching `Header`'s existing slot-prop pattern.
   - **Six nav icons transcribed exactly from Figma** — Overview, Vaults, Loans, Activity, Liquidations, Explore, using the literal `<path>` geometry from the Figma Sidebar node rather than a similar-looking icon library substitute. Each icon accounts for its own inset within the 24px slot (only Activity fills edge-to-edge; the rest are padded or off-center per Figma) instead of stretching a tight-bounds viewBox, which would have rendered them 15–33% too heavy.
   - **Brand lockup matches Figma's literal colors** — the Babylon/Aave logo paths are plain black (light) / white (dark), not the brand-orange tint `Header`'s own lockup uses; the divider is a separate literal gray, independent of the logo's color.
   - **Wired into `services/vault`'s `RootLayout`** behind `FeatureFlags.isV3UiEnabled` — desktop shows the sidebar and drops the top-nav;
mobile keeps the existing hamburger overlay, now populated with the same six items. Flag off renders identically to today's v2 dashboard.
   - **Sidebar is `sticky top-0 h-svh`**, pinned to the viewport instead of stretching to match the main content column's height — otherwise, on any page taller than one viewport, the nav and footer would scroll out of reach. Banners render inside the content column (instead of above the row) specifically when the desktop sidebar is showing, so the sidebar's position isn't pushed down by banner height — which was clipping the footer at initial scroll position whenever an operator banner is visible (the testing banner defaults to visible).
   - **New `nav` key in `copy.ts`** for the six labels, kept separate from existing page headings so the two can diverge later.

## Why

The v3 app shell needs persistent left navigation instead of the horizontal `Header` nav. The Figma `Sidebar` component is used on every v3 screen, so it lives in `core-ui` rather than the vault app.

## What's verified

   - `tsc --noEmit` and `vite build` clean on both `@babylonlabs-io/core-ui` and `@services/vault`.
   - New `AppSidebar` unit tests (all 6 items render; active-route highlighting, including the exact-root-path case for Overview) plus the existing router test suite, all green.
   - Lint clean on both packages (0 errors).
   - Icon offsets and brand colors verified against the actual served dev bundle, not just source.

   ## Known limitations / follow-ups

   - `/vaults`, `/loans`, `/liquidations`, `/explore` have no routes yet — clicking those items hits the router's `NotFound` catch-all. Expected: route restructure is #2017.
   - `Header` still renders its own logo when v3 is on, redundant with the sidebar's brand lockup. Deferred to #2016 (page-title header rebuild), which is the natural place to drop it.
   - Sidebar is not literally confirmed against a scrolling/sticky Figma spec (static frames can't show that) — the sticky-viewport behavior is an engineering decision to keep nav and footer reachable regardless of page height, not a guessed Figma value.

## Where to look

   - `packages/babylon-core-ui/src/components/Sidebar/` — `Sidebar.tsx`, `SidebarItem.tsx`, `SidebarBrandLockup.tsx`.
   - `packages/babylon-core-ui/src/components/Icons/common/{Overview,Vaults,Loans,Activity,Liquidations,Explore}Icon.tsx` and the six social icons.
   - `services/vault/src/components/shared/AppSidebar.tsx` — nav groups, footer, brand composition.
   - `services/vault/src/components/pages/RootLayout.tsx` — shell wiring, banner placement, sticky layout.

## How to run

`NEXT_PUBLIC_FF_ENABLE_V3_UI=true` in `services/vault/.env.local`, then `pnpm dev` from `services/vault`.


services/vault/src/router.tsx
```diff
@@ -29,6 +31,9 @@ const AaveReserveDetail = lazyWithRetry(() =>
   importAaveReserveDetail().then((m) => ({ default: m.AaveReserveDetail })),
 );
 
+// Guarded as whole subtrees, so a v3 deep link redirects rather than 404s.
+const V3_ROUTE_PATHS = ["vaults", "loans", "liquidations"] as const;
```
jrwbabylonlab: `/explore` is in the sidebar but not in this list, so with the flag off a direct `/explore` load hits the 404 catch-all instead of redirecting like its siblings.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -167,161 +174,180 @@ export default function RootLayout() {
...
-          // Tint the logo brand-orange in light mode; keep the default
-          // light-on-dark contrast in dark mode. The wrapper's `[&_svg]` selector
-          // overrides the SVG's hardcoded `text-accent-primary` color, which the
-          // paths inherit through `fill-current`.
-          logo={
-            <div className="flex items-center gap-3">
-              <div className="[&_svg]:!h-8 [&_svg]:!w-auto [&_svg]:!text-secondary-main dark:[&_svg]:!text-accent-primary">
-                <SmallLogo />
+    <div className="relative flex min-h-svh w-full flex-col bg-surface">
+      {!showV3Sidebar && banners}
```
jrwbabylonlab: `banners` renders in two different tree positions (here and line 228), so crossing the 768px breakpoint remounts the critical-banner slot div — `CriticalLiquidationTopBanner` resolves the slot once on mount, so the banner ends up portaled into a detached node. Rendering it once as the content column's first child avoids the remount.

services/vault/src/components/shared/AppSidebar.tsx
```diff
@@ -0,0 +1,184 @@
...
+// Matches Figma's Social Links block exactly (icon order + Terms/Privacy
+// text). URLs are the same handles already used by the page footer, plus a
+// verified Telegram handle (t.me/babylonlabs_io) that block didn't carry.
+const SIDEBAR_SOCIAL_LINKS: SidebarSocialLink[] = [
+  {
+    name: "GitHub",
+    url: "https://github.com/babylonlabs-io",
+    Icon: GithubIcon,
+  },
+  { name: "Telegram", url: "https://t.me/babylonlabs_io", Icon: TelegramIcon },
```
jrwbabylonlab: Is `t.me/babylonlabs_io` actually ours? Everywhere else in the org (countdown, tbv-faucet, coming-soon) links `t.me/babyloncommunity`, and this one resolves to a ~200-subscriber channel.

packages/babylon-core-ui/src/components/Icons/index.ts
```diff
@@ -43,4 +43,16 @@ export { ThreeDotsMenuIcon } from "./common/ThreeDotsMenuIcon";
...
+export { OverviewIcon } from "./common/OverviewIcon";
+export { VaultsIcon } from "./common/VaultsIcon";
+export { LoansIcon } from "./common/LoansIcon";
+export { ActivityIcon } from "./common/ActivityIcon";
+export { LiquidationsIcon } from "./common/LiquidationsIcon";
+export { ExploreIcon } from "./common/ExploreIcon";
+export { GithubIcon } from "./common/GithubIcon";
+export { TelegramIcon } from "./common/TelegramIcon";
+export { LinkedinIcon } from "./common/LinkedinIcon";
+export { MailIcon as SidebarMailIcon } from "./common/MailIcon";
```
jrwbabylonlab: Is the alias needed? There's no other `MailIcon` export in core-ui, so a plain `export { MailIcon }` would match the other 11 icons.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -167,161 +174,180 @@ export default function RootLayout() {
...
+  // v3 desktop keeps the sidebar's natural position at y=0 so it's fully
+  // in the viewport immediately, with zero clipping, regardless of which
+  // banners are visible.
+  const banners = (
+    <>
+      {/* Portal target for the critical near-liquidation banner. Owned by
+          the dashboard (where the Aave data + debug override live) but
+          portaled here so it renders above the header, atop the
+          operational banners. */}
+      <div id={CRITICAL_BANNER_SLOT_ID} />
```
gbarkhatov: The critical-liquidation banner can silently disappear on a viewport resize. `banners` (including this portal slot) is now rendered in two swappable positions — `{!showV3Sidebar && banners}` (L224) and `{showV3Sidebar && banners}` (L228). Crossing the 768px breakpoint flips `showV3Sidebar`, so React unmounts one subtree and mounts the other, giving `#critical-liquidation-banner-slot` a brand-new DOM node.

But `CriticalLiquidationTopBanner` resolves the slot once on mount and caches it (`useEffect([], () => setSlot(document.getElementById(...)))`) and never re-resolves — after the remount it keeps `createPortal`-ing into the old, now-detached node, so this non-dismissible near-liquidation alert vanishes until its source component happens to remount. Only the portaled banner is affected (the inline banners just re-render in the new position).

Suggest rendering a single stable slot independent of the banner stack's position (e.g. as the content column's first child), so it never remounts on the breakpoint flip.

services/vault/src/router.tsx
gbarkhatov: `explore` is missing from this list. The sidebar links six sections but only three are guarded, so with the flag off a direct load / stale deep link to `/explore` falls through to the `NotFound` catch-all instead of redirecting to `/` like its siblings — which contradicts the comment right above ("a v3 deep link redirects rather than 404s"). (`/activity` is correctly excluded — it has a real route at L114.) Add `explore`.

Separately, this list duplicates the paths already declared on `V3_NAV_GROUPS` in `AppSidebar.tsx` (and again in `router.test.tsx`). Worth deriving from a single shared const so the guard can't silently desync from the rail — same two-sources-of-truth shape flagged on #2025.

services/vault/src/components/shared/AppSidebar.tsx
```diff
@@ -0,0 +1,184 @@
...
+interface SidebarSocialLink {
+  name: string;
+  url: string;
+  Icon: ComponentType<{ size?: number; className?: string; title?: string }>;
+}
+
+// Matches Figma's Social Links block exactly (icon order + Terms/Privacy
+// text). URLs are the same handles already used by the page footer, plus a
+// verified Telegram handle (t.me/babylonlabs_io) that block didn't carry.
+const SIDEBAR_SOCIAL_LINKS: SidebarSocialLink[] = [
```
gbarkhatov: This set duplicates the URLs already in `FOOTER_SOCIAL_LINKS` (`RootLayout.tsx:85`) — GitHub / LinkedIn / Email / Discord / X are identical. The icons legitimately differ (Figma-exact vs react-icons), but the name+URL set could be a single shared const both import, so a handle change doesn't have to be made in two places.

Also please double-check the new Telegram handle on L106 (`t.me/babylonlabs_io`) — per the comment above it's the one URL not carried over from the existing footer set, and I couldn't find it referenced anywhere else in the repo to cross-check. A wrong social link would ship to every v3 user.

packages/babylon-core-ui/src/components/Icons/index.ts
gbarkhatov: Minor: `MailIcon` is the only mail icon in core-ui, so the `as SidebarMailIcon` alias isn't needed — a plain `export { MailIcon }` matches the other icons exported here. The `Sidebar` prefix reads like it's disambiguating a name collision that doesn't exist.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -167,161 +174,180 @@ export default function RootLayout() {
...
-              <SimpleDeposit
-                open={isDepositOpen}
-                onClose={closeDeposit}
-                initialAmountBtc={initialDepositAmountBtc}
+                  initialAmountBtc={initialDepositAmountBtc}
+                />
+              </AaveConfigProvider>
+            </ActivatingVaultsProvider>
+          )}
+          {!FeatureFlags.isV3UiEnabled && (
```
gbarkhatov, line 356: With the v3 flag on and on mobile (`showV3Sidebar` false, so no rail), the page footer is gated off here while the Terms of Use / Privacy Policy + social block lives only inside `V3MobileNavigation` (the hamburger overlay). A mobile v3 user who never opens the menu has no path to the legal links anywhere on the page. Worth confirming that's intended — it's adjacent to the mobile-v3 coverage gaps raised on #2025.

gbarkhatov (dismissed):
Reviewed the sidebar rail end-to-end — clean split into router-agnostic core-ui `Sidebar`/`SidebarItem`/`SidebarBrandLockup`, sensible flag gating, copy correctly routed through `copy.ts`, and good test coverage (six items render, active-route incl. exact-root, on/off-flag redirects). Approving; nothing below blocks the flag-gated desktop path.

One item is worth a look before this sees real traffic: the critical-liquidation banner can detach on a viewport resize now that its portal slot moves between two render positions (inline comment on `RootLayout.tsx`). The rest are a minor correctness gap (`/explore` gating) and cleanups.

A couple of things I looked at and deliberately did *not* flag inline:
- Greptile's "mobile first render flashes the desktop shell" — not applicable here: `useIsMobile` reads `window.innerWidth` in a lazy initializer and vault is a client-rendered Vite SPA (no SSR/hydration), so `isMobile` is correct on the first paint.
- The sidebar shows no active item while an Aave reserve-detail overlay is open (`/app/aave/reserve/:id`), since Overview is an exact (`end`) match — real, but arguably fine for an overlay; leaving it as a judgment call.
- `XIcon`'s `0 0 14 12` viewBox vs the `16 16` siblings, and the trailing blank line in `SidebarBrandLockup`, are cosmetic.

jonybur (commented):
**Adversarial review — 💬 Comment (mergeable behind the flag, but one real bug must be fixed before v3 is enabled)**

Nice work on the rail — icons transcribed cleanly, flag-off renders v2, and the AppSidebar tests are reasonable. I read every file at full context and ran an adversarial verification pass. One genuine bug, plus minor polish.

---

### 🔴 Critical near-liquidation banner can silently vanish in v3 on a viewport resize

`RootLayout.tsx` — the `banners` fragment contains the **fixed-id portal target** `<div id={CRITICAL_BANNER_SLOT_ID}/>` and is now rendered in two positions under **different parents**:

```
{!showV3Sidebar && banners}        // outer root div
...
  {showV3Sidebar && banners}       // inner content-column div
```

The consumer `CriticalLiquidationTopBanner` resolves that target **once on mount** (`useEffect(() => setSlot(document.getElementById(CRITICAL_BANNER_SLOT_ID)), [])`) and stays mounted inside `<Outlet>`. When `showV3Sidebar` (`isV3UiEnabled && !isMobileView`) flips on a **resize across the 768px breakpoint** — phone rotation, window resize — React can't move a DOM node between different parents: it destroys the slot `<div>` at one position and creates a fresh one at the other. The consumer keeps a **stale, detached** reference, so `createPortal(..., slot)` renders into a removed node and the red liquidation warning disappears until a full page reload.

- Verified against the real code. **Introduced by this PR**: on `main` the slot lives at a single stable position and never relocates, so the cached reference stays valid.
- **Dormant while the flag is off**, so it doesn't block merging behind the default-off flag — but it's a safety-critical regression in the exact shell this PR builds, so it must be fixed before v3 is turned on anywhere (devnet included). Please don't let "it's flag-gated" carry it past review.
- Fix options: keep the `<div id=…>` mounted at one stable position that never unmounts and move only the *visible* banners; or have `CriticalLiquidationTopBanner` re-resolve the target on each render / via a ref-callback / MutationObserver instead of a one-shot mount effect.

---

### Minor / nits

- **Flag-off DOM is no longer byte-identical to today's v2** (`RootLayout.tsx` root wrapper class changed + two nested flex wrappers always inserted). Equivalent by analysis, but given the explicit "flag off === today's v2" rule, worth a quick visual QA of the v2 path.
- **Social-link names bypass `copy.ts`** — `AppSidebar.tsx` inlines `"GitHub"/"Telegram"/…` and passes them as the icons' accessible `title`/tooltip. Terms/Privacy already live in `copy.ts`; these should too (URLs stay as config).
- `MailIcon` re-exported as `SidebarMailIcon` with no actual name collision to disambiguate; `Icons/index.ts` also has no trailing newline at EOF.
- No test exercises `V3MobileNavigation` or the flag-on desktop sidebar mount. Active-state assertion uses `getByText(...).closest("div")`, brittle to `SidebarItem` markup changes.

gbarkhatov (dismissed):

services/vault/src/config/socialLinks.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Shared social link URLs, so the page footer (v2 + mobile-v3) and the v3
+ * sidebar's own social block can't drift on a handle/URL change. Icons and
```
jrwbabylonlab, line 3: "cannot drift" is not accurate yet — the page footer still renders its own `FOOTER_SOCIAL_LINKS` copies rather than this config. Migrate the footer or fix the comment.

gbarkhatov:
@jeremy-babylonlabs conflict

jonybur (dismissed):
Approving. The critical near-liquidation banner slot bug I flagged earlier is fixed: `<div id={CRITICAL_BANNER_SLOT_ID} />` is now unconditionally the first child of the same content column regardless of `showV3Sidebar`, so a viewport-resize across 768px no longer unmounts/remounts the portal target. A dedicated test (`keeps the critical-banner portal slot mounted across the mobile breakpoint`) asserts the DOM node identity is preserved.

Sidebar rail, icons, and `AppSidebar`/`V3MobileNavigation` are behind `FeatureFlags.isV3UiEnabled` and `useIsMobile`; flag-off leaves the v2 shell (`BrandLockup`, `DesktopNavigation`/`MobileNavigation`, page `Footer`) intact. New RootLayout tests lock in the v2 vs v3 header wiring and the v3-mobile SidebarFooter path. CI green, no substantive findings.

jrwbabylonlab (approved):

jonybur (approved):
Re-approving after force-push. The follow-up commit "resolve outstanding #2021/#2023/#2025 review feedback" addresses my earlier flags cleanly:

- Critical near-liquidation banner portal slot is now unconditionally the content column's first child regardless of `showV3Sidebar`, so a viewport resize across 768px no longer unmounts/remounts the target — `RootLayout.test.tsx` locks that in.
- v3 nav sections (path + label + which subtrees the flag-off block redirects) live in one `config/v3Navigation.ts` module, consumed by `router.tsx`, `usePageTitle`, and `AppSidebar`, so a renamed section can't desync. `/explore` now has its redirect guard too.
- Social link URLs are shared via `config/socialLinks.ts`; sidebar Telegram now points at the org's actual `t.me/babyloncommunity` channel.
- Mobile v3 keeps the in-page Overview/Activity headings and gains a persistent `SidebarFooter` with social + legal links, so a mobile user who never opens the hamburger still has a path.
- `NetworkBadge` drops the unreachable TESTNET branch and its dead test; `usePageTitle` matches case-insensitively; the `SidebarMailIcon` alias is gone.

CI green, no substantive findings.

PR #2025, opened 2026-07-09
feat(vault): show page title in v3 header instead of nav

jeremy-babylonlabs:
image

Add a per-page title to the v3 header, driven by the active route, replacing the header's logo/nav slot. usePageTitle() maps the current pathname to the matching sidebar section's title (Overview, Vaults, Loans, Activity, Liquidations, Explore), falling back to Overview for the root dashboard and its AAVE reserve-detail overlay routes. Matching is segment-boundary aware (mirrors react-router's NavLink), so a route like /vaultsfoo doesn't false-match /vaults.

Add a NetworkBadge chip to the header's right actions (Figma's "Testnet" pill), shown only on non-mainnet networks — mirrors simple-staking's existing NetworkBadge convention. Connect, StandardSettingsMenu, and the Deposit action are unchanged.

Drop the v3 header's own brand logo (BrandLockup), which duplicated the sidebar's brand lockup — a known limitation called out in PR #2023 (left sidebar navigation rail). v2 is untouched: same BrandLockup, same nav, no NetworkBadge.

Also gate OverviewSection's and ActivityList's own in-page headings behind !FeatureFlags.isV3UiEnabled: both duplicated the new header title in v3 (confirmed live — Activity showed "Activity" twice in the disconnected state). ActivityList's row also hosts the Aave avatar + FilterDropdown when connected, so its fix keeps that row rendering (right-aligned, no heading) whenever connected, and drops the row entirely when disconnected to avoid an empty-row gap.

Zero out the v3 header's bottom margin (mb-20 -> mb-0): it left a large gap between the title and page content, confirmed live. v2 keeps the default margin.

Refs #2016

### DIFF
--- a/services/vault/src/components/pages/RootLayout.tsx
+++ b/services/vault/src/components/pages/RootLayout.tsx
@@ -2,6 +2,7 @@ import {
   Footer,
   FullScreenDialog,
   Header,
+  Heading,
   Loader,
   MobileLogo,
   Nav,
@@ -35,6 +36,7 @@ import { useAddressScreening } from "@/context/addressScreening";
 import { useAddressType } from "@/context/addressType";
 import { useGeoFencing } from "@/context/geofencing";
 import { COPY } from "@/copy";
+import { usePageTitle } from "@/hooks/usePageTitle";
 import { useProtocolGateState } from "@/hooks/useProtocolGate";
 
 import {
@@ -46,6 +48,7 @@ import { AddressScreeningBanner } from "../shared/AddressScreeningBanner";
 import { AddressTypeBanner } from "../shared/AddressTypeBanner";
 import { DepositDisabledBanner } from "../shared/DepositDisabledBanner";
 import { GeoBlockState } from "../shared/GeoBlockState";
+import { NetworkBadge } from "../shared/NetworkBadge";
 import { NoticeBanner } from "../shared/NoticeBanner";
 import {
   isDepositBlocked,
@@ -155,6 +158,7 @@ export default function RootLayout() {
   const { isBlocked: isAddressBlocked } = useAddressScreening();
   const { isSupportedAddress } = useAddressType();
   const isMobileView = useIsMobile();
+  const pageTitle = usePageTitle();
   const showV3Sidebar = FeatureFlags.isV3UiEnabled && !isMobileView;
 
   const isWalletConnected = btcConnected && ethConnected;
@@ -228,11 +232,27 @@ export default function RootLayout() {
           {showV3Sidebar && banners}
           <Header
             size="md"
+            // v3 owns its own vertical rhythm via each page's content
+            // padding (matches Figma: no gap between the header's bottom
+            // border and the content start). v2 keeps the default mb-20.
+            className={FeatureFlags.isV3UiEnabled ? "mb-0" : undefined}
             // `PAGE_CONTENT_CLASS` carries `!max-w-[1080px]`, overriding the
             // `container` width core-ui's Header applies by default so the navbar
             // shares the same 1080px content box as the page body and footer.
             containerClassName={PAGE_CONTENT_CLASS}
-            logo={<BrandLockup />}
+            logo={
+              FeatureFlags.isV3UiEnabled ? (
+                <Heading
+                  variant="h5"
+                  as="h1"
+                  className="font-normal text-accent-primary"
+                >
+                  {pageTitle}
+                </Heading>
+              ) : (
+                <BrandLockup />
+              )
+            }
             mobileLogo={
               <div className="[&_svg]:!text-secondary-main dark:[&_svg]:!text-accent-primary">
                 <MobileLogo />
@@ -250,6 +270,7 @@ export default function RootLayout() {
             }
             rightActions={
               <div className="flex items-center gap-4">
+                {FeatureFlags.isV3UiEnabled && <NetworkBadge />}
                 {isWalletConnected &&
                   !isDepositOpen &&
                   !isGeoBlocked &&
--- a/services/vault/src/components/shared/NetworkBadge.tsx
+++ b/services/vault/src/components/shared/NetworkBadge.tsx
@@ -1,0 +1,26 @@
+import { Text } from "@babylonlabs-io/core-ui";
+import { Network } from "@babylonlabs-io/wallet-connector";
+
+import { getBTCNetwork } from "@/config";
+import { COPY } from "@/copy";
+
+// Mirrors simple-staking's NetworkBadge convention (services/simple-staking/
+// src/ui/common/components/NetworkBadge/NetworkBadge.tsx): shown only on
+// non-production networks, hidden entirely on mainnet.
+const NON_MAINNET_NETWORKS: readonly Network[] = [
+  Network.SIGNET,
+  Network.TESTNET,
+];
+
+/** v3 header network indicator — the Figma "Testnet" chip (node 10084:22952). */
+export function NetworkBadge() {
+  if (!NON_MAINNET_NETWORKS.includes(getBTCNetwork())) return null;
+
+  return (
+    <span className="flex items-center rounded-full bg-secondary-highlight px-2.5 py-1">
+      <Text as="span" variant="caption" className="text-warning-main">
+        {COPY.header.networkBadge}
+      </Text>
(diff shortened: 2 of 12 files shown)


services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -228,11 +232,27 @@ export default function RootLayout() {
...
+            // v3 owns its own vertical rhythm via each page's content
+            // padding (matches Figma: no gap between the header's bottom
+            // border and the content start). v2 keeps the default mb-20.
+            className={FeatureFlags.isV3UiEnabled ? "mb-0" : undefined}
             // `PAGE_CONTENT_CLASS` carries `!max-w-[1080px]`, overriding the
             // `container` width core-ui's Header applies by default so the navbar
             // shares the same 1080px content box as the page body and footer.
             containerClassName={PAGE_CONTENT_CLASS}
-            logo={<BrandLockup />}
+            logo={
```
jrwbabylonlab, line 243: Header renders `mobileLogo` below 768px, so this title never shows on mobile — and since the in-page Overview/Activity headings are now hidden on the flag alone, v3 mobile ends up with no page title at all. Gating the heading-hiding on `!isMobile` too (like `showV3Sidebar` above) would cover it.

services/vault/src/hooks/usePageTitle.ts
```diff
@@ -0,0 +1,28 @@
...
+  ["/explore", COPY.nav.explore],
+];
+
+/** Current page title for the v3 header, driven by the active route. */
+export function usePageTitle(): string {
+  const { pathname } = useLocation();
+  // Segment-boundary match — mirrors react-router's `NavLink` (used by
+  // AppSidebar's nav links), so `/vaults` matches but `/vaultsfoo` doesn't.
+  const match = ROUTE_TITLES.find(
+    ([path]) => pathname === path || pathname.startsWith(`${path}/`),
```
jrwbabylonlab, line 25: react-router matches routes and NavLink active-state case-insensitively by default, so `/Activity` renders the page with its nav item active but falls back to the Overview title here.

services/vault/src/components/shared/NetworkBadge.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { getBTCNetwork } from "@/config";
+import { COPY } from "@/copy";
+
+// Mirrors simple-staking's NetworkBadge convention (services/simple-staking/
+// src/ui/common/components/NetworkBadge/NetworkBadge.tsx): shown only on
+// non-production networks, hidden entirely on mainnet.
+const NON_MAINNET_NETWORKS: readonly Network[] = [
+  Network.SIGNET,
+  Network.TESTNET,
```
jrwbabylonlab, line 12: `getBTCNetwork()` only returns MAINNET or SIGNET in vault, so this entry (and the testnet test case) is unreachable — a plain `!== Network.MAINNET` check covers it.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -228,11 +232,27 @@ export default function RootLayout() {
...
+              FeatureFlags.isV3UiEnabled ? (
```
gbarkhatov, line 244: **Mobile v3: the page title disappears entirely (no `h1`, no `h2`).**

The page-title `h1` is passed only to the Header's `logo` slot, which core-ui `Header` renders exclusively on desktop — on mobile it renders `mobileLogo` (the `MobileLogo` brand) instead. Meanwhile `OverviewSection.tsx:91` and `ActivityList.tsx:47` suppress their in-page heading on the raw `FeatureFlags.isV3UiEnabled` with no mobile carve-out.

Net: with the flag on, a mobile user on Overview or Activity sees no page title anywhere and the page has no heading element at all — a UX and a11y regression vs. v2, which always rendered the `h2`. Mobile v3 is a real path (`V3MobileNavigation` is wired just below at line 266).

Suggest keying the heading suppression off the same condition that actually shows the header title (desktop-scoped, like `showV3Sidebar`) rather than raw `isV3UiEnabled` — or surfacing the title in the mobile header too.

services/vault/src/components/shared/NetworkBadge.tsx
gbarkhatov, line 12: `Network.TESTNET` is unreachable here. `getBTCNetwork()` resolves via `NETWORK_BY_NAME`, whose keys are only `mainnet | signet` (`config/btc.ts`), so it can only ever return `MAINNET` or `SIGNET` — never `TESTNET`. The entry can't match, and `NetworkBadge.test.tsx`'s "renders the badge on testnet" case only passes by mocking a value the real config can't emit — so that assertion doesn't verify production behavior.

This trips the ZERO DEAD CODE POLICY and the "every assertion must verify behavior of code in `src/`" test rule. Either drop `TESTNET` (and its test), or wire testnet into config so it's actually reachable.

services/vault/src/hooks/usePageTitle.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { COPY } from "@/copy";
+
+// Longest-prefix match against the v3 sidebar's routes (see
+// components/shared/AppSidebar.tsx's `V3_NAV_GROUPS`), so a path the sidebar
+// treats as "Vaults" (etc.) shows that same title in the header. Anything
+// that doesn't match — the root dashboard and its AAVE reserve-detail
+// overlay routes (`/app/aave/reserve/:id/...`) — is the Overview page, so
+// it's the fallback rather than a listed prefix.
+const ROUTE_TITLES: readonly (readonly [path: string, title: string])[] = [
```
gbarkhatov, line 11: **Cleanup:** `ROUTE_TITLES` re-declares the same path→`COPY.nav.*` set that `AppSidebar.tsx:53-68` already owns as `to`+`label` on `V3_NAV_GROUPS`. Two sources of truth: renaming or re-pathing a nav section means editing both in lockstep, and a miss makes the header title silently disagree with the sidebar's active item. Consider deriving this map from the sidebar's array (or a shared const both import).

Minor: the doc comment above says "Longest-prefix match", but the implementation is first-match (`.find(...)`). It's harmless since the routes don't overlap, but "first segment-boundary match" would describe the code more accurately.

gbarkhatov (approved):
Approving — the v3 header rebuild is clean, well-scoped, and nicely tested (hook, badge, and RootLayout header wiring). A few items to consider, one of which is a real mobile regression worth a follow-up; none block the desktop path this ships behind the flag.

**Non-blocking notes (left as body notes to keep the diff comments focused):**

- **`NetworkBadge` reads "Testnet" on Signet.** `getBTCNetwork()` returns `Network.SIGNET` on the signet deployment, so the chip renders but is labeled "Testnet" rather than "Signet". This deliberately mirrors simple-staking's convention and the Figma pill, so it's likely intended — flagging it as a conscious label choice rather than a fix.
- **Out of scope / pre-existing:** with the v3 flag on, the sidebar links `/vaults`, `/loans`, `/liquidations`, `/explore` have no routes registered under `RootLayout` (the redirect fallbacks in `router.tsx` are gated to `!isV3UiEnabled`), so they fall through to `NotFound`. `usePageTitle`'s mappings for those paths are effectively unreachable until the pages exist. `router.tsx` isn't in this diff — noting only because the new title map assumes those routes.

jonybur (approved):
**Adversarial review — ✅ Approve**

Solid. `usePageTitle`'s segment-boundary matching is correct (`/vaultsfoo` doesn't false-match `/vaults`), the `NetworkBadge` and the flag-gated `ActivityList`/`OverviewSection` heading changes are clean, and flag-off parity holds. Good test coverage on the header wiring and the connected/disconnected activity-row cases.

Nits only (non-blocking):

- `usePageTitle` comment says **"Longest-prefix match"** but the implementation is first-match (`Array.find`); correct only because no configured prefix is a prefix of another. Reword to "first-match" or make it order-independent so a future `/loans` vs `/loans-x` addition can't surprise.
- `NetworkBadge` uses an allowlist (`SIGNET`, `TESTNET`) rather than `!== MAINNET`, so any other non-production network would show no badge. Matches simple-staking's convention, so fine — just noting the latent edge.
- `ActivityList`'s flag-off class string is reordered vs `main` (byte-parity nit only; visually identical).

PR #2027, opened 2026-07-10
feat(vault): derive the deposit time estimate from on-chain depth

jrwbabylonlab:
The Deposit Progress header hardcoded "~60 min" while the in-flow
counter computed minPrepeginDepth × 10 min from chain params — on
testnet (depth 12) the same card showed 60 and 120 min. Compute the
header from the deposit's pinned depth plus a fixed pipeline allowance.

### DIFF
--- a/services/vault/src/components/simple/DepositProgressView/DepositCardShell.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/DepositCardShell.tsx
@@ -21,7 +21,10 @@ import { Heading, Text } from "@babylonlabs-io/core-ui";
 import type { ReactNode } from "react";
 
 import { COPY } from "@/copy";
+import { useRequiredPrePeginDepth } from "@/hooks/deposit/useRequiredPrePeginDepth";
+import { formatDurationShort } from "@/utils/formatting";
 
+import { computeTotalEstimateMinutes } from "./btcConfirmationProgress";
 import { DEPOSIT_VIEW_MAX_WIDTH_CLASS } from "./layout";
 
 interface DepositCardShellProps {
@@ -45,6 +48,12 @@ interface DepositCardShellProps {
    * do-not-spend warning shown while the deposit is in flight.
    */
   footnote?: ReactNode;
+  /**
+   * The deposit's registered offchain-params version, so the header estimate
+   * uses the same pinned confirmation depth the flow actually gates on.
+   * Omitted pre-sign, where no version is registered yet → latest params.
+   */
+  offchainParamsVersion?: number;
 }
 
 export function DepositCardShell({
@@ -52,7 +61,11 @@ export function DepositCardShell({
   footer,
   progressBar,
   footnote,
+  offchainParamsVersion,
 }: DepositCardShellProps) {
+  const requiredDepth = useRequiredPrePeginDepth(offchainParamsVersion);
+  const estimateMinutes = computeTotalEstimateMinutes(requiredDepth);
+
   return (
     <div
       // Cap the card height to the viewport and lay it out as a column so the
@@ -64,7 +77,11 @@ export function DepositCardShell({
         <Heading variant="h5" className="text-accent-primary">
           {COPY.deposit.progress.heading}{" "}
           <Text as="span" variant="body1" className="text-accent-secondary">
-            ({COPY.deposit.progress.summary.estimate})
+            (
+            {COPY.deposit.progress.summary.estimate(
+              formatDurationShort(estimateMinutes),
+            )}
+            )
           </Text>
         </Heading>
 
--- a/services/vault/src/components/simple/DepositProgressView/__tests__/DepositProgressView.test.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/__tests__/DepositProgressView.test.tsx
@@ -19,6 +19,7 @@ vi.mock("../BtcConfirmationDetailContainer", () => ({
 vi.mock("@/context/ProtocolParamsContext", () => ({
   useProtocolParamsContext: () => ({
     config: { offchainParams: { minPrepeginDepth: 6 } },
+    getOffchainParamsByVersion: () => undefined,
   }),
 }));
 
--- a/services/vault/src/components/simple/DepositProgressView/__tests__/DepositProgressView.test.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/__tests__/DepositProgressView.test.tsx
@@ -14,6 +14,14 @@ vi.mock("../BtcConfirmationDetailContainer", () => ({
   ),
 }));
 
+// DepositCardShell derives the header estimate from the on-chain confirmation
+// depth; the estimate itself is covered in DepositCardShell.test.tsx.
+vi.mock("@/context/ProtocolParamsContext", () => ({
+  useProtocolParamsContext: () => ({
+    config: { offchainParams: { minPrepeginDepth: 6 } },
+  }),
+}));
+
 // DepositProgressView self-sources the BTC wallet-lock state to render its
 // unlock notice and pre-sign unlock CTA. Drive it through a mutable mock;
 // default unlocked.
--- a/services/vault/src/components/simple/PostDepositContinuationView.tsx
+++ b/services/vault/src/components/simple/PostDepositContinuationView.tsx
@@ -3,9 +3,9 @@ import { useNavigate } from "react-router";
 import type { Address, Hex } from "viem";
 
 import { usePeginPolling } from "@/context/deposit/PeginPollingContext";
-import { useProtocolParamsContext } from "@/context/ProtocolParamsContext";
 import { COPY } from "@/copy";
 import { DepositFlowStep } from "@/hooks/deposit/depositFlowSteps";
+import { useRequiredPrePeginDepth } from "@/hooks/deposit/useRequiredPrePeginDepth";
 import { deriveSplitVaultProgress } from "@/hooks/deposit/useSplitVaultProgress";
 import {
   getPeginDisplayStep,
@@ -52,6 +52,7 @@ function StatusView({
   vaultCount = 1,
   currentVaultIndex = null,
   perVaultSteps,
+  offchainParamsVersion,
 }: {
   currentStep: DepositFlowStep;
   onClose: () => void;
@@ -64,6 +65,7 @@ function StatusView({
   vaultCount?: number;
   currentVaultIndex?: number | null;
   perVaultSteps?: DepositFlowStep[];
+  offchainParamsVersion?: number;
 }) {
   return (
     <DepositProgressView
@@ -81,6 +83,7 @@ function StatusView({
       onClose={onClose}
       successMessage={successMessage}
       btcConfirmationDetail={btcConfirmationDetail}
+      offchainParamsVersion={offchainParamsVersion}
     />
   );
 }
@@ -93,7 +96,6 @@ export function PostDepositContinuationView({
   onClose,
 }: PostDepositContinuationViewProps) {
   const { refetch, getPollingResult } = usePeginPolling();
-  const { config, getOffchainParamsByVersion } = useProtocolParamsContext();
   const navigate = useNavigate();
 
   const handleGoToDashboard = useCallback(() => {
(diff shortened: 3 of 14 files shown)


services/vault/src/hooks/deposit/useRequiredPrePeginDepth.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export function useRequiredPrePeginDepth(
+  offchainParamsVersion?: number,
+): number {
+  const { config, getOffchainParamsByVersion } = useProtocolParamsContext();
+  return (
+    (offchainParamsVersion !== undefined
+      ? getOffchainParamsByVersion(offchainParamsVersion)?.minPrepeginDepth
+      : undefined) ?? config.offchainParams.minPrepeginDepth
+  );
```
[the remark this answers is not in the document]
jrwbabylonlab: Keeping the fallback — a registered version can be legitimately absent (versions that fail validation are skipped from byVersion, see fetchAllOffchainParams's onSkipped), and this hook only feeds display estimates while the real depth gate is enforced VP/contract-side. Same convention as getRequiredPrePeginDepth and resolveTimelockAssertBlocks; throwing here would crash the resume modal over display copy.

services/vault/src/components/simple/PostDepositContinuationView.tsx
```diff
@@ -368,6 +368,7 @@ export function PostDepositContinuationView({
       canContinueInBackground
       onClose={onClose}
       btcConfirmationDetail={btcConfirmationDetail}
+      offchainParamsVersion={activity?.offchainParamsVersion}
```
jonybur, line 376: Minor / optional consistency: `offchainParamsVersion` is threaded here (and into the `Resume*` views) so the header estimate pins to the deposit's registered params version. But the three `StatusView` call sites in the `!currentVaultId` branch don't pass it and fall back to latest params:

- `COMPLETED` (~L189) and the warning/danger view (~L231) - estimate is moot in a terminal/failed state, so no real impact.
- the multi-vault `AWAIT_BTC_CONFIRMATION` aggregate (~L253) - this one is actively in flight. For a batched deposit that straddles a `minPrepeginDepth` change, the aggregate view would show a header estimate derived from *latest* params while the per-vault view shows the *pinned* one - the exact drift this PR is closing.

`activities` is in scope in that branch, so a representative sibling's `offchainParamsVersion` could be passed through. Since batch siblings share a version in practice this is an edge case, but flagging in case it's worth closing fully.

jonybur (dismissed):
LGTM. Clean, well-tested refactor that pins the header estimate to the deposit's registered offchain-params depth and dedupes the inline version-pinning logic into `useRequiredPrePeginDepth`. Verified: no dangling `useProtocolParamsContext` references after removal, every `DepositCardShell` render path is under the (blocking) `ProtocolParamsProvider` so there's no missing-provider/NaN risk, and the `formatDurationShort`/`computeTotalEstimateMinutes` test assertions all check out.

Left one minor, optional inline note: the `StatusView` call sites in the `!currentVaultId` branch (notably the multi-vault AWAIT_BTC_CONFIRMATION aggregate) don't pass `offchainParamsVersion`, so they fall back to latest params. Edge case only - not blocking.

jeremy-babylonlabs (dismissed):

services/vault/src/hooks/deposit/useRequiredPrePeginDepth.ts
```diff
@@ -0,0 +1,18 @@
...
+ * `PeginPollingContext.getRequiredPrePeginDepth` — or the latest version
+ * otherwise (pre-sign, where the deposit has no registered version yet).
+ */
+export function useRequiredPrePeginDepth(
+  offchainParamsVersion?: number,
+): number {
+  const { config, getOffchainParamsByVersion } = useProtocolParamsContext();
+  return (
+    (offchainParamsVersion !== undefined
+      ? getOffchainParamsByVersion(offchainParamsVersion)?.minPrepeginDepth
```
gbarkhatov: Non-blocking (reuse): this usefully collapses the two inline copies that were in `PostDepositContinuationView` and `ResumeWotsContent`, but the same `getOffchainParamsByVersion(v)?.minPrepeginDepth ?? config.offchainParams.minPrepeginDepth` rule still lives in `PeginPollingContext.getRequiredPrePeginDepth` (which this hook's own JSDoc cites as the reference). That leaves two sources of truth for one protocol rule — change one and the other drifts. Consider having the context expose a version-taking selector that this hook reuses.

services/vault/src/components/simple/DepositProgressView/DepositCardShell.tsx
```diff
@@ -45,14 +48,24 @@ interface DepositCardShellProps {
...
 }
 
 export function DepositCardShell({
   children,
   footer,
   progressBar,
   footnote,
+  offchainParamsVersion,
 }: DepositCardShellProps) {
+  const requiredDepth = useRequiredPrePeginDepth(offchainParamsVersion);
```
gbarkhatov, line 66: Non-blocking (altitude): `DepositCardShell` was a pure layout wrapper (children/footer/progressBar/footnote slots); it now reaches into `ProtocolParamsContext` and computes the estimate. Since the PR already threads `offchainParamsVersion` through `DepositProgressView` and every call site, this pays for both prop-drilling and a data dependency inside a presentational component. Computing the estimate string once at the container layer and passing it down as a plain string would keep the shell dumb and drop the context coupling.

services/vault/src/utils/formatting.ts
```diff
@@ -332,6 +332,25 @@ export function formatDuration(totalMinutes: number): string {
...
+/** Compact approximate duration for headline estimates: "70 min", "1.5 h",
+ *  "2 h". Rounds to 5-minute steps under 90 minutes, half hours above. */
+export function formatDurationShort(totalMinutes: number): string {
+  const roundedMinutes =
+    Math.round(totalMinutes / SHORT_DURATION_MINUTE_STEP) *
+    SHORT_DURATION_MINUTE_STEP;
+  if (roundedMinutes < SHORT_DURATION_HOURS_FROM_MINS) {
+    return `${roundedMinutes} min`;
+  }
+  const halfHours = Math.round(totalMinutes / SHORT_DURATION_HALF_HOUR_MINS);
```
gbarkhatov, line 353: Non-blocking (clarity/nit): the minutes-vs-hours decision tests `roundedMinutes` (snapped to 5-min steps) but this hours branch re-rounds from the raw `totalMinutes / 30`, so the branch base and the format base differ — e.g. 89 snaps to 90, falls here, `round(89/30)=3` → "1.5 h", and nothing ever renders as "90 min". Output stays monotonic for the real input domain (max minute output 85 min < min hours output 90 min) so it's harmless, but rounding once and branching+formatting off that single value would be easier to reason about. Minor: the `/ 2` on the next line is the one quantity left inline while everything else was extracted to a `SHORT_DURATION_*` constant.

services/vault/src/components/simple/DepositProgressView/__tests__/DepositProgressView.test.tsx
```diff
@@ -14,6 +14,14 @@ vi.mock("../BtcConfirmationDetailContainer", () => ({
   ),
 }));
 
+// DepositCardShell derives the header estimate from the on-chain confirmation
+// depth; the estimate itself is covered in DepositCardShell.test.tsx.
+vi.mock("@/context/ProtocolParamsContext", () => ({
+  useProtocolParamsContext: () => ({
```
gbarkhatov, line 20: Non-blocking (test-only): this mock returns `config` but omits `getOffchainParamsByVersion`, which `useRequiredPrePeginDepth` (now reached via `DepositCardShell`) calls whenever a version is supplied. The current tests pass only because none set `offchainParamsVersion`, so the `version !== undefined` branch short-circuits. The first test that renders `DepositProgressView` with a defined `offchainParamsVersion` (e.g. a resume / await-confirmation case) will hit `TypeError: getOffchainParamsByVersion is not a function`. Worth adding the function to the mock now to defuse the trap.

gbarkhatov (dismissed):
Reviewed the diff — deriving the header estimate from the deposit's pinned on-chain depth is a clean, well-tested change. No correctness issues: the `useRequiredPrePeginDepth` refactor is behavior-preserving, the hook is correctly hoisted above early returns (Rules of Hooks), `BTC_BLOCK_TIME_MINS` (10) equals the old `AVG_BTC_BLOCK_MINUTES` so the rename is value-preserving, `formatDurationShort` is monotonic across the real input domain, and every `DepositCardShell` render path sits inside `ProtocolParamsProvider` so the new context dependency can't throw.

Approving. A few non-blocking quality notes inline: a duplicated depth rule, a presentational-shell altitude point, a latent test-mock trap, and a formatter clarity nit.

services/vault/src/components/simple/PostDepositContinuationView.tsx
jrwbabylonlab: Good catch — the aggregate view now passes a representative sibling's version (siblings share it by construction: one registration tx per batch), with a test pinning it. Left the terminal/warning sites as-is per your note.

services/vault/src/hooks/deposit/useRequiredPrePeginDepth.ts
jrwbabylonlab: Consolidated, but in the reverse direction — the context can't be the source because the fresh-deposit modal renders outside PeginPollingProvider. The hook module now exposes useRequiredPrePeginDepthResolver and getRequiredPrePeginDepth delegates to it, so the rule lives in one place.

services/vault/src/components/simple/DepositProgressView/DepositCardShell.tsx
jrwbabylonlab: There isn't really one container layer — DepositProgressView has 7 render sites across 3 files, so hoisting the string would duplicate the hook per site, and computing it in DepositProgressView just moves the context read into a component that already reads wallet context. Since the shell's only consumer is DepositProgressView, keeping the estimate next to the header it renders feels like the right altitude.

services/vault/src/utils/formatting.ts
jrwbabylonlab: The two bases are deliberate — hours round from raw so the half-hour snap doesn't compound the 5-min snap (103 min would flip from 1.5 h to 2 h), while branching on the snapped value is what keeps '90 min' from rendering. Added a comment saying exactly that; kept the /2 inline since it's a unit conversion rather than a tunable.

services/vault/src/components/simple/DepositProgressView/__tests__/DepositProgressView.test.tsx
jrwbabylonlab: Added the stub — matches the mock shape the sibling test files already use.

services/vault/src/components/simple/PostDepositContinuationView.tsx
jrwbabylonlab: resolved

services/vault/src/hooks/deposit/useRequiredPrePeginDepth.ts
jrwbabylonlab: resolved

services/vault/src/components/simple/DepositProgressView/__tests__/DepositProgressView.test.tsx
jrwbabylonlab: resolved

services/vault/src/components/simple/DepositProgressView/DepositCardShell.tsx
jrwbabylonlab: Keeping as is — 7 render sites, hoisting would duplicate the computation per site.

services/vault/src/utils/formatting.ts
jrwbabylonlab: Keeping as is — a single rounding base would flip 103 min from 1.5 h to 2 h; documented in a code comment instead.

jonybur (approved):

PR #2029, opened 2026-07-11
feat(vault): 2017 restructure to flat v3 page routes behind ENABLE_V3_UI

jeremy-babylonlabs:
Implements babylonlabs-io/babylon-toolkit#2017 as a structural-only route cutover behind `NEXT_PUBLIC_FF_ENABLE_V3_UI`.

Stacked on #2025 (per-page titles), which stacks on #2023 (left sidebar). This branch supplies the real `/vaults`, `/loans`, `/liquidations` routes those two depend on — resolves the "section title on routes with no routed content" gap flagged in #2025's review.

## Route contract

- Public paths are flat: `/`, `/vaults`, `/loans`, `/activity`, `/liquidations`.
- The former `/app/<app>/...` route extension is removed. No nested legacy redirect is retained; old `/app/aave/reserve/...` URLs fall through to Not Found.
- Reserve-detail overlays stay deep-linkable without adding path depth: v2 uses `/?reserve=<id>&tab=<borrow|repay>`, v3 uses `/loans?reserve=<id>&tab=<borrow|repay>`.
- With the flag off, `/vaults`, `/loans`, and `/liquidations` redirect to `/`; the v2 dashboard and `/activity` remain available.
- With the flag on, the routes now render through the inherited sidebar + per-page title shell from #2023/#2025 — live-verified.

## Fix included

This PR's route restructure moves `DashboardPage` to render through `AaveOverlayLayout`'s inner `<Outlet/>`. React Router always wraps an `<Outlet/>` in a context provider, so the unforwarded one shadows `RootLayout`'s outlet context (`openDeposit`) with `undefined`, crashing `DashboardPage`'s `useOutletContext()` destructure. `AaveOverlayLayout` now forwards its inherited context explicitly. Added `router.test.tsx` coverage that exercises the real outlet-context path (the prior mock never called `useOutletContext()`, so it never caught this).

## Verification

Router tests: 27/28 passing (11 new tests for flag-aware routing, wrong-base gating, subtree redirects; 1 pre-existing Activity test failure).

Before merging, verify locally:
- `pnpm exec tsc --noEmit -p tsconfig.lib.json` (TypeScript)
- `pnpm exec eslint src/router.tsx src/routes.ts src/__tests__/router.test.tsx src/applications/aave/components/Detail/index.tsx src/applications/aave/components/AssetPill/AssetPill.tsx src/components/simple/DashboardPage.tsx --max-warnings 0` (lint)
- `pnpm build` (production build)


services/vault/src/router.tsx
```diff
@@ -86,62 +92,31 @@ const ActivityWithProviders = () => (
...
-            key={path}
-            path={`${path}/*`}
-            element={<Navigate to="/" replace />}
-          />
-        ))}
-      <Route path="*" element={<NotFound />} />
-    </Routes>
-  );
-};
+      <Route path={ROUTES.VAULTS} element={<V3Placeholder />} />
```
gbarkhatov, line 120: Behavioral change worth confirming is intentional: the old router registered `path="${section}/*"` → `<Navigate to="/" replace/>`, so with the flag **off** a deep link like `/vaults/anything` redirected to the dashboard. These exact-path routes only match `/vaults`, `/loans`, `/liquidations`, so `/vaults/foo` now falls through to `<Route path="*" element={<NotFound/>}>` and shows Not Found. No v3 URLs are nested today so it's effectively a no-op, and the old deep-link redirect test was intentionally dropped — just flagging to confirm losing the graceful redirect is intended.

gbarkhatov (commented):
Comment-only review (PR is in draft, so not approving). The routing rewrite looks correct — route matching, overlay gating, reserve-id case handling, tab fallback, and the `getAllApplications`/`Routes` removal are all clean, and the outlet-context fix is a real catch. Inline notes are all minor. A few observations that don't map onto changed lines:

- **Stale comment + redundant test for the removed overlay route** (files not in this diff): `usePageTitle.ts:8-9` still cites `/app/aave/reserve/:id/...` as an example fallback path, and `usePageTitle.test.ts:57` still renders `/app/aave/reserve/abc/borrow` as its "unmatched path" case. That route no longer exists — the comment describes removed behavior (repo convention is to update such comments) and the test's URL is now fictional (assertion still valid, already covered by the generic case). Trivial cleanup.
- **PR description vs. code:** the description says "v3 uses `/loans?reserve=<id>&tab=...`", but the reserve detail only mounts when `pathname === '/'` (and the test asserts `/loans?reserve=usdc&tab=repay` shows the placeholder, not the detail). `getReserveDetailRoute` always returns `/?reserve=...`. Not a code bug — `/loans` is a contentless placeholder today — just a slightly misleading contract line.
- **`/explore` (pre-existing, out of scope):** the v3 sidebar NavLink to `/explore` and `usePageTitle`'s `/explore` title have no matching route, so clicking Explore hits `path="*"` → NotFound. Not introduced here; worth tracking separately.

services/vault/src/router.tsx
```diff
@@ -1,19 +1,28 @@
...
   AaveConfigProvider,
   PendingVaultsProvider,
   ReorderOverrideProvider,
 } from "./applications/aave/context";
-import RootLayout from "./components/pages/RootLayout";
+import RootLayout, {
+  type RootLayoutContext,
+} from "./components/pages/RootLayout";
 import NotFound from "./components/pages/not-found";
+import featureFlags from "./config/featureFlags";
```
gbarkhatov: Nit (consistency): this switches the flag import from the `@/config/featureFlags` alias to a relative `./config/featureFlags`. Elsewhere the app reads the same flags via `FeatureFlags` from `@/config` (e.g. `RootLayout`). Same underlying object, so harmless — just off from the codebase norm; consider `@/config`.

services/vault/src/router.tsx
```diff
@@ -86,62 +92,31 @@ const ActivityWithProviders = () => (
...
 );
 
-export const Router = () => {
-  // Narrow to apps that actually expose Routes so the element below can render
-  // <AppRoutes /> unconditionally — no dead fallback branch.
-  const apps = getAllApplications().filter(
-    (app): app is typeof app & { Routes: ComponentType } => Boolean(app.Routes),
+const V3Placeholder = () =>
+  featureFlags.isV3UiEnabled ? (
+    <div data-testid="v3-placeholder" />
```
gbarkhatov, line 108: Nit: `V3Placeholder` ships a bare `<div data-testid="v3-placeholder" />` to production markup — fine as temporary scaffolding while these pages are built, but a test-id-only element in prod is a small smell. Fine to leave; noting for when the real pages land.

jonybur (commented):
**Adversarial review — 💬 Comment (draft; core approach is good, three substantive items before it's approve-ready)**

The route restructure is sound and the rewritten router tests are genuinely better — they exercise the real outlet context instead of mocking it away. I read every file at full context and adversarially verified each point below.

---

### 1. The PR description's central "crash" claim is inaccurate

The body states main crashes `DashboardPage`'s `useOutletContext()` "on every load in both v2 and v3 — not just under the flag." **This is not correct.** On `main`, `<DashboardPage/>` is rendered as a *sibling* of `AaveOverlayLayout`'s inner `<Outlet/>` (the index element is `null`), so it already reads `RootLayout`'s provider and gets `openDeposit` — and `main` ships to production, which a crash-on-every-load would obviously break.

The context shadowing is **introduced by this PR's own restructure** (moving `DashboardPage` to render *through* `AaveOverlayLayout`'s inner `<Outlet/>`) and then correctly neutralized by forwarding `context={outletContext}`. The fix and its regression test are good — just please correct the narrative so reviewers don't believe `main` is currently broken.

### 2. Removing `/app/aave/reserve/:id/:tab` breaks old deep links in v2 (flag OFF), with no redirect

`router.tsx` deletes those routes entirely (replaced by `/?reserve=<id>&tab=<tab>`) with no shim, so old bookmarked/shared reserve-detail URLs now fall through to a shell-less `NotFound`. This is **flag-independent** — it changes shipping v2 behavior, not just v3. Blast radius is small (transient modal overlay, rarely bookmarked, internal nav updated to the new format), but a one-line back-compat `<Navigate>` route is cheap insurance. If the break is intentional, please state it explicitly and confirm nothing external emits the old format.

### 3. Flag-off redirect guarantee silently narrowed

#2021 deliberately guarded `vaults/*` as a **subtree** so a flag-off deep link *redirects* rather than 404s. This PR switches to exact-path routes, so `/vaults/foo` (flag off) now 404s instead of redirecting — quietly dropping the guarantee #2021's own description emphasized. Low impact, but should be a decision rather than a silent change.

---

### Cleanups

- `canShowReserveDetail = pathname === ROUTES.OVERVIEW` is always true inside `AaveOverlayLayout` (its only child route is `/`), so the guard can never be false — dead defensive code.
- The `reserve` / `tab` query-param keys are bare string literals duplicated across `routes.ts` (producer) and `router.tsx` (reader); extract to shared constants so they can't desync.
- Cross-PR staleness introduced by this route change: `usePageTitle.ts`'s comment and its `usePageTitle.test.ts` case both still reference the deleted `/app/aave/reserve/:id/...` path (the test name is now false and it asserts a fallback for a route that 404s).
- `Explore` sidebar item still has no route/placeholder → flag-on lands on `NotFound`, inconsistent with the `vaults`/`loans`/`liquidations` placeholders. A placeholder would keep parity.

services/vault/src/router.tsx
```diff
@@ -71,11 +82,15 @@ const AaveOverlayLayout = () => {
...
         <ReorderOverrideProvider>
           <Suspense fallback={<RouteFallback />}>
-            <DashboardPage />
-          </Suspense>
-          <Suspense fallback={null}>
-            <Outlet />
+            <Outlet context={outletContext} />
           </Suspense>
+          {reserveId &&
+            pathname ===
```
jrwbabylonlab, line 88: React Router matches `/loans/` and case variants tolerantly, but this exact `pathname ===` check does not — the page renders and the reserve overlay silently drops. Normalize the pathname before comparing.

jrwbabylonlab:
Base branch is green now (ActivityList fix landed, 214/214 on e4aede74) — rebase and the failing Activity test here passes without any changes to this diff.

jonybur (dismissed):
Approving. Re-read the router at the current head. The reserve-detail overlay is now gated on `pathname === getReserveDetailBaseRoute(isV3Enabled)`, so flag-off + `/loans?reserve=...` and flag-on + `/?reserve=...` both fall through to the dashboard, and the new tests cover both wrong-base cases. The subtree redirect guarantee I flagged earlier is preserved (`${path}/*` in the flag-off block). The old `/app/aave/reserve/:id/:tab` deep-link path 404s with no back-compat shim, but that break is intentional and low-blast-radius (transient modal overlay, internal nav already updated to the new query-param URL). CI green, no substantive findings.

jrwbabylonlab (approved):

PR #2031, opened 2026-07-13
fix(vault): center the disconnected landing screen with margin auto

jonybur:
Screenshot 2026-07-13 at 15 09 14

Adds `my-auto` to the disconnected-state Dashboard Container, completing its built-in `mx-auto` to a full `margin: auto` so the landing/connect screen is vertically centered in the remaining viewport height. Only the disconnected state is affected; the connected dashboard is unchanged.

### DIFF
--- a/services/vault/src/components/pages/RootLayout.tsx
+++ b/services/vault/src/components/pages/RootLayout.tsx
@@ -157,9 +157,13 @@ export default function RootLayout() {
   // On that screen only, drop the footer's top margins (the wrapper's `mt-auto`
   // and the Footer's own `mt-24`) so the footer sits directly below the centered
   // content instead of competing for the free vertical space. Scoped to the
-  // dashboard route so an unlogged /activity keeps its sticky footer.
+  // dashboard route so an unlogged /activity keeps its sticky footer, and to
+  // the non-geo branch: geo-loading/geo-blocked render short content instead of
+  // the centered Container, so the footer must keep `mt-auto` to stay at the
+  // viewport bottom.
   const { pathname } = useLocation();
-  const isDisconnectedLanding = !isWalletConnected && pathname === "/";
+  const isDisconnectedLanding =
+    !isWalletConnected && pathname === "/" && !isGeoBlocked && !isGeoLoading;
   const [isDepositOpen, setIsDepositOpen] = useState(false);
   const [initialDepositAmountBtc, setInitialDepositAmountBtc] = useState<
     string | undefined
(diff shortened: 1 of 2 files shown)


jonybur:
@greptile-apps

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -152,6 +152,14 @@ export default function RootLayout() {
...
   const isWalletConnected = btcConnected && ethConnected;
   const showAddressTypeBanner = isWalletConnected && !isSupportedAddress;
+  // The disconnected dashboard landing (DisconnectedOverview / "Native Bitcoin
+  // backed borrowing") is vertically centered via `my-auto` on its Container.
+  // On that screen only, drop the footer's top margins (the wrapper's `mt-auto`
+  // and the Footer's own `mt-24`) so the footer sits directly below the centered
+  // content instead of competing for the free vertical space. Scoped to the
+  // dashboard route so an unlogged /activity keeps its sticky footer.
+  const { pathname } = useLocation();
+  const isDisconnectedLanding = !isWalletConnected && pathname === "/";
```
jrwbabylonlab: This is also true when disconnected at `/` but geo-blocked or still geo-loading — those branches don't render the `my-auto` container, yet the footer's `mt-auto`/`mt-24` are still stripped, leaving the footer floating mid-page on tall viewports. Gate on `!isGeoBlocked && !isGeoLoading` too.

gbarkhatov: `isDisconnectedLanding` is derived from route + connection only, but the `my-auto` `Container` is only rendered in the non-geo branch. In the content ternary below (lines 261–310):

- `isGeoLoading` → a `min-h-[50vh]` `Loader` (Outlet not rendered)
- `isGeoBlocked` → `GeoBlockState`, a short `mx-auto … py-20` card with no vertical fill (Outlet not rendered)
- else → Outlet → `DashboardPage` → the `my-auto` `Container`

Geo-blocking (indexer HTTP 451) is region-based and independent of wallet connection, so a disconnected + geo-blocked user on `/` is a common path. There, `isDisconnectedLanding` is still `true`, so the footer gets `mt-0` / `!mt-0` — but nothing on screen absorbs the free vertical space, so the footer floats up directly beneath the geo-block card instead of sticking to the viewport bottom (a regression from the current `mt-auto` behavior). The `isGeoLoading` case has the same issue but is transient.

Suggest tying the flag to the state that actually renders the centered `Container`:

```ts
const isDisconnectedLanding =
  !isWalletConnected && pathname === "/" && !isGeoBlocked && !isGeoLoading;
```

gbarkhatov (dismissed):
LGTM — the centering approach is sound: `my-auto` on the disconnected `Container` sandwiches the footer (set to `mt-0`) between two auto margins, which vertically centers the landing while keeping the footer pinned to the bottom, matching the screenshot.

One edge case worth a look before merge (left as an inline comment): the `isDisconnectedLanding` flag keys off route + connection only, so it also strips the footer's `mt-auto` in the geo-loading and geo-blocked states, where the centered `Container` isn't the thing rendered.

jrwbabylonlab (approved):

PR #2033, opened 2026-07-13
test(vault): borrow E2E CLI

### LINKED ISSUE #2032: test(vault): add a borrow action to the real-wallet E2E CLI (step 5)
## Step 5 — Borrow against BTC-Vault collateral (real-wallet E2E CLI)

Extends the real-wallet E2E CLI (`services/vault/e2e/real/`) to the next step of the depositor lifecycle: **borrow**. Builds on connect (#2005), single-vault pegin (#2012), and the two-vault split pegin (#2024). This is `5 - borrow.md` in `markdown/ai/`.

### Goal

Add a `borrow` action that draws a borrowable token (USDC / USDT / WBTC) against an activated BTC-Vault position, end-to-end on real signet + Sepolia, and finishes at the "Borrow successful" screen with the debt confirmed on-chain.

### Scope — the borrow action

Two shapes, selected by the CLI:

- **Reuse existing collateral** (default): borrow against the depositor's already-activated vaults. A pre-flight reads the position from real data (contract + indexer) and refuses before the browser if there is no collateral.
- **Pegin-first** (`--pegin-first`): peg in fresh collateral first (reusing the existing pegin flow), then borrow — all in one browser session. Works for a fresh position and for adding a vault to an existing one.

The token comes from the live borrowable-reserve list; the amount defaults to a conservative fraction of the real-data max (kept well clear of the liquidation edge), and is overridable via `--borrow-token` / `--borrow-amount` (`--borrow-amount=max` uses the form's Max). The live form stays the authoritative gate; the CLI's computed values are best-effort defaults.

- **Max-BTC-Vaults-per-position guard** — a pegin is refused (pre-flight + fast form-gate detection) when the position is already at its on-chain per-position vault cap, instead of launching a doomed deposit.
- **WOTS-readiness fast-fail** — when the vault provider isn't ready for WOTS-key submission in time (e.g. a slow Bitcoin block), a doomed pegin now aborts in ≤20 min with a clear, VP-attributed message instead of hanging the full 2.5-hr budget.
- **Post-activation collateral-settle wait** — after a pegin-first activation, the borrow waits for the new collateral to register on-chain (baseline-relative, so it correctly waits for a *second* vault added to an existing position) before entering the amount — the collateral is not immediately borrowable because the UI shows the vault active optimistically while the on-chain position read lags.
- **`networkContracts.ts` shared module + params-layer DRY cleanup** — network/env resolution, the Sepolia + GraphQL clients, and the vBTC-reserve-id read are shared between the pegin and borrow pre-flights; a small set of `data-testid`s were added to the borrow controls in `services/vault/src` for robust, copy-drift-proof selectors.

### Acceptance

- A headed, real borrow on devnet completes and the borrowed debt is confirmed on-chain, for: reuse against existing collateral; pegin-first with a fresh vault; and pegin-first adding a second vault where the borrow amount only clears against the *combined* collateral (proving the new vault is counted).
- Existing pegin (single + split) still passes — no regression.

- Repay and withdraw actions.
- Resuming an in-flight pegin (still open from earlier steps).

gbarkhatov:
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2032

## What this PR does

Adds a `borrow` action to the real-wallet E2E CLI (`services/vault/e2e/real/`) — the next step of the depositor lifecycle after connect and pegin. It drives the vault dApp with real wallet extensions (UniSat + MetaMask) on real testnets (signet BTC + Sepolia ETH) to borrow a token (USDC / USDT / WBTC) against an activated BTC-Vault position, end-to-end, and finishes at the "Borrow successful" screen.

The borrow runs in one of two shapes, chosen by the CLI:

- **Reuse existing collateral** (default): borrow against vaults the depositor already has. Before opening the browser, a pre-flight reads the position from real on-chain + indexer data and refuses the run early if there is no collateral to borrow against.
- **Pegin-first** (`--pegin-first`): peg in a fresh vault first (reusing the existing pegin flow), then borrow — all in one browser session. This works both for a brand-new position and for adding another vault to an existing one.

You pick the token with `--borrow-token` (defaults to the first borrowable one) and the amount with `--borrow-amount` (`--borrow-amount=max` uses the form's Max button). When you don't pass an amount, the CLI reads the real max from chain data and borrows a conservative fraction of it so the position stays well clear of liquidation.

Alongside borrow, this PR adds a few things that the borrow flow needed or that harden the existing pegin (all in the same E2E CLI):

- **"Maximum BTC Vaults reached" guard** — a pegin is refused early when the position is already at its on-chain per-position vault cap, instead of launching a deposit that the app would block anyway.
- **Vault-provider readiness fast-fail** — if the vault provider isn't ready for the WOTS key step in time (usually because a Bitcoin block is slow), the pegin now stops in at most ~20 minutes with a clear "this is a vault-provider issue, not the CLI" message, instead of hanging for the full 2.5-hour budget.
- **Wait for fresh collateral before borrowing** — after a pegin-first activation, the app shows the new vault as active immediately (optimistically), but the amount you can borrow is read from the on-chain position, which lags by a variable 15–30+ seconds. So the CLI now waits for the new collateral to actually register on chain before it enters the borrow amount. It waits for the collateral to rise *above what was there before*, so adding a second vault to an existing position is handled correctly (not just "collateral > 0").
- **Shared plumbing + cleanup** — network/env resolution, the Sepolia and GraphQL clients, and the vBTC-reserve-id read are now shared between the pegin and borrow pre-flights (`networkContracts.ts`) instead of duplicated, and a handful of stable `data-testid`s were added to the borrow controls in `services/vault/src` so the test can target them without depending on button text.

## Why

The E2E CLI is being grown one depositor step at a time (connect → pegin → borrow → repay → withdraw) so we can drive and verify each real flow unattended. Borrow is the next step. The extra guards exist because both of them showed up as real failure modes while testing borrow on devnet: a position at its vault cap, and a pegin that couldn't finish because the vault provider wasn't ready — both of which used to fail slowly and confusingly.

## Approaches, and why we chose them

- **The CLI re-derives a little app logic against the SDK instead of importing the app's React code.** The CLI is a separate Node runtime — it can't cleanly import the app's hooks, `@/` aliases, or single-environment singletons. So for anything that must match the app exactly (protocol minimums, the borrowable-token list, the max-borrow math) it calls the same pure SDK functions the app calls, and only re-implements the small glue (a couple of GraphQL queries and the max-borrow formula, both commented as "keep in sync"). Crucially, all of these are treated as *best-effort defaults*: the live form (its Max button and validation) is always the authoritative gate, so any small drift in the CLI's estimate can never cause a wrong borrow — the form catches it.

- **Conservative default borrow amount, not Max.** Borrowing the maximum drives the health factor right to the liquidation edge, which is risky and not repeatable for a test. So the default is a safe fraction of the max; Max is still reachable explicitly.

- **Support both reuse and pegin-first rather than only one.** Reuse is fast and cheap to run repeatedly (no new pegin); pegin-first proves the whole composite works from nothing. Keeping both means we can iterate quickly on the borrow UI (reuse) and still validate the full path (pegin-first).

- **Wait on the on-chain collateral amount, in sats, not the USD value.** The USD collateral value moves with the oracle price between reads, so a "did it go up?" check on USD would need a fuzzy threshold. The BTC amount in sats only changes when a vault is actually registered as collateral, so waiting for a strict increase there is exact and price-independent.

- **Testid-first selectors with a tolerant fallback.** The borrow UI had no stable test hooks, and button text drifts between the source and the deployed build (this already bit us once). So the test prefers a `data-testid` (added in this PR) and falls back to role/text/class for a build that predates the testid — the same pattern the pegin flow already uses.

## How it was validated

Run end-to-end on devnet with real funds, confirmed on-chain each time:

- Reuse borrow against existing collateral — debt appeared on-chain.
- Pegin-first with a fresh vault — full pegin → activation → borrow, debt confirmed.
- Pegin-first adding a *second* vault, borrowing an amount that only clears against the *combined* collateral — proving the new vault's collateral is actually counted (the borrow was briefly rejected as "exceeds maximum" until the second vault registered, then went through).
- The vault-provider fast-fail was also exercised (a slow Bitcoin block) and aborted cleanly in ~20 minutes.

Existing single-vault and split pegins still pass — no regression. `pnpm exec eslint e2e/real` and `pnpm exec tsc --noEmit -p tsconfig.e2e.json` are clean.

## Note for reviewers

This branch bundles the borrow feature with the two pegin guards and the shared-plumbing cleanup described above — they landed together because borrow depended on them. Everything is test tooling under `services/vault/e2e/real/` plus four one-line `data-testid` additions in `services/vault/src`; no product or protocol logic is changed.

### DIFF
--- a/services/vault/e2e/real/actions/borrow.ts
+++ b/services/vault/e2e/real/actions/borrow.ts
@@ -1,0 +1,433 @@
+/**
+ * The "borrow" action: draw a borrowable token (Aave-style) against an activated BTC-Vault position,
+ * end-to-end on real Sepolia. Two shapes, selected by the CLI:
+ *   - REUSE (default): borrow against the depositor's existing collateral. run.ts already refused the
+ *     run before the browser if there's no active collateral (fetchBorrowContext), so we land on the
+ *     dashboard with the Borrow button enabled.
+ *   - PEGIN-FIRST (`--pegin-first`): peg in fresh collateral first (the shared `runPeginFlow`, which
+ *     ends on the dashboard with an active vault), then borrow — all in one browser session.
+ *
+ * The borrow flow itself is short and has NO multi-minute on-chain gates (unlike pegin): open Loans →
+ * Borrow, pick the token in the "Select asset" modal, enter the amount (a conservative fraction of the
+ * real-data max, or the form's Max), submit, and approve the single MetaMask transaction. Then the
+ * "Borrow successful" screen confirms it.
+ *
+ * Selectors are testid-first (added to the src borrow controls, mirroring `activate-vault-button`) with
+ * tolerant text/role/class fallbacks so a deployed build that predates the testids still works. No SDK
+ * / product logic is reimplemented — the amount is a best-effort default and the live form is the
+ * authoritative gate (its Max button + validation).
+ *
+ * NEVER run without an explicit go-ahead: it moves real value (a borrow draw, plus a full pegin when
+ * `--pegin-first`).
+ */
+import type { BrowserContext, Locator, Page } from "@playwright/test";
+
+import {
+  CONSERVATIVE_BORROW_FRACTION,
+  fetchCollateralSats,
+  fetchMaxBorrow,
+  formatBorrowAmount,
+} from "../borrowParams";
+import { formatBtc } from "../preflight";
+import {
+  BORROW_BUTTON_ENABLE_TIMEOUT_MS,
+  BORROW_CTA_ENABLE_TIMEOUT_MS,
+  BORROW_TX_TIMEOUT_MS,
+  FORM_SETTLE_MS,
+  FRESH_COLLATERAL_POLL_MS,
+  FRESH_COLLATERAL_TIMEOUT_MS,
+  MS_PER_SECOND,
+  STEP_TIMEOUT_MS,
+} from "../timing";
+
+import { installPopupApprover, sweepApprovals } from "./approver";
+import { runPeginFlow } from "./pegin";
+import { startRecording } from "./recording";
+import { FLUID_CTA_SELECTOR, firstByTestid } from "./selectors";
+import { type Action, type ActionContext } from "./types";
+import { connectWallets } from "./walletConnect";
+
+// Dashboard "Loans" section → Borrow (testid-first; the fallback matches the button only while it reads
+// exactly "Borrow", which is fine on the dashboard where the CTA isn't relabeled).
+const LOANS_BORROW_TESTID = '[data-testid="loans-borrow-button"]';
+const BORROW_BUTTON_RX = /^borrow$/i; // COPY.loans.borrowButton
+const ASSET_SELECT_TITLE = "Select asset"; // COPY.loans.assetSelection.title
+// Borrow form: the AmountSlider's numeric input (`inputmode="decimal"`, placeholder "0") + its Max
+// button (a <button> whose visible text is "Max"), and the fluid submit button.
+const AMOUNT_INPUT = 'input[inputmode="decimal"]';
+const MAX_BUTTON_RX = /^max$/i;
+const BORROW_SUBMIT_TESTID = '[data-testid="borrow-submit-button"]';
+const BORROW_SUBMIT_ENABLED_LABEL = "Borrow"; // COPY.loans.borrow.action (enabled state)
+// Submit labels that won't resolve by waiting — fail fast with the callout (COPY.loans.borrow.*).
+// "Borrowing Unavailable" is protocol-gated; "Amount too small" / "…exceeds available liquidity" are
+// fixed properties of the entered amount + reserve, not of how much collateral has propagated.
+const BORROW_INSTANT_FAIL_LABELS = new Set([
+  "Borrowing Unavailable",
+  "Amount too small",
+  "Amount exceeds available liquidity",
+]);
+// Labels that depend on the position's collateral and so can be TRANSIENT right after a pegin-first
+// activation — the borrow form's max/health-factor grow as the just-activated vault propagates into
+// its on-chain position read. We poll through these (not instant-fail) and only surface them if they
+// persist to the enable deadline (COPY.loans.borrow.*).
+const BORROW_COLLATERAL_DEPENDENT_LABELS = new Set([
+  "Amount exceeds maximum",
+  "Health factor too low",
+]);
+// Best-effort: the validation callout body phrases (COPY.loans.validation.*) + the availability ones,
+// surfaced in the fail-fast message so a blocked run says why.
+const CALLOUT_BODY_RX =
+  /(minimum borrowable|maximum borrowable|available to borrow|health factor|temporarily unavailable|Price data unavailable)[^.]*\./i;
+// Success screen (testid-first; tolerant fallbacks for a pre-testid deployed build).
+const SUCCESS_DONE_TESTID = '[data-testid="loan-success-done-button"]';
+const DONE_BUTTON_RX = /^done$/i; // COPY.loans.borrowSuccess.doneButton
+const BORROW_SUCCESS_RX = /borrow successful/i; // COPY.loans.borrowSuccess.title
+const TX_FAILED_RX = /transaction failed/i; // COPY.common.transactionFailedTitle
+const MAX_AMOUNT_KEYWORD = "max";
+
+/** The resolved borrow amount: click the form's Max button, or fill a specific token amount. */
+type BorrowAmount = { mode: "max" } | { mode: "amount"; value: string };
+
+/**
+ * Resolve the amount to borrow. An explicit `--borrow-amount` wins (a number, or `max`). Otherwise it
+ * computes a conservative fraction of the real-data max — this runs after any pegin-first pegin has
+ * activated, so fresh collateral is already readable — falling back to the form's Max if that read is
+ * zero or fails. The live form stays the authoritative gate either way.
+ */
+async function resolveBorrowAmount(ctx: ActionContext): Promise<BorrowAmount> {
+  const raw = ctx.config.borrowAmount?.trim();
+  if (raw && raw.toLowerCase() === MAX_AMOUNT_KEYWORD) return { mode: "max" };
+  if (raw) return { mode: "amount", value: raw };
+
+  // No explicit amount → compute a conservative fraction of the real-data max. This runs AFTER any
+  // pegin-first pegin has activated, so the fresh collateral is already on-chain and readable here
+  // (getPosition/getUserAccountData are on-chain, not indexer). Falls back to the form's Max only if
+  // the read is zero (indexer/state lag) or fails — the form stays the authoritative gate either way.
+  const token = ctx.config.borrowToken?.trim();
+  if (!token) return { mode: "max" };
+  try {
+    const max = await fetchMaxBorrow(
+      ctx.config.network,
+      ctx.eth.address,
+      token,
+    );
+    if (max.maxTokens > 0) {
+      const value = formatBorrowAmount(
+        max.maxTokens * CONSERVATIVE_BORROW_FRACTION,
+        max.decimals,
+      );
+      ctx.log(
+        `Borrow amount: ${value} ${max.symbol} (~${Math.round(CONSERVATIVE_BORROW_FRACTION * 100)}% of max ${formatBorrowAmount(max.maxTokens, max.decimals)} ${max.symbol}).`,
+      );
+      return { mode: "amount", value };
+    }
+    ctx.log(
+      `⚠️ Computed max borrow for ${token} is 0 — using the form's Max (it will gate the amount).`,
+    );
+  } catch (error) {
+    ctx.log(
+      `⚠️ Could not compute a conservative default (${error instanceof Error ? error.message : error}); using the form's Max.`,
+    );
+  }
+  return { mode: "max" };
+}
+
+/** Open the borrow flow from the dashboard: click Loans → Borrow, wait for the "Select asset" modal. */
+async function openBorrow(page: Page, log: (m: string) => void): Promise<void> {
+  const borrow = firstByTestid(
+    page,
+    LOANS_BORROW_TESTID,
+    page.getByRole("button", { name: BORROW_BUTTON_RX }),
+  );
+  await borrow.waitFor({ state: "visible", timeout: STEP_TIMEOUT_MS });
+  // The button is gated on `hasCollateral` (collateral > 0). A JUST-activated vault (pegin-first) takes
+  // a moment to propagate into the app's position read, so the button can be briefly disabled right
+  // after "Go to Dashboard". Poll for it to enable rather than failing on the first check; only treat a
+  // persistently-disabled button as "no collateral". (A reuse run already passed run.ts's collateral
+  // gate, so its button is enabled almost immediately.)
+  const deadline = Date.now() + BORROW_BUTTON_ENABLE_TIMEOUT_MS;
+  let enabled = await borrow.isEnabled().catch(() => false);
+  while (!enabled && Date.now() < deadline) {
+    await page.waitForTimeout(FORM_SETTLE_MS);
+    enabled = await borrow.isEnabled().catch(() => false);
+  }
+  if (!enabled)
+    throw new Error(
+      `The dashboard Borrow button stayed disabled for ${Math.round(BORROW_BUTTON_ENABLE_TIMEOUT_MS / MS_PER_SECOND)}s — this position has no active BTC Vault collateral to borrow against. Peg in first (or re-run with --pegin-first).`,
+    );
+  log("Opening the borrow flow (Loans → Borrow)");
+  await borrow.click();
+  await page
+    .getByText(ASSET_SELECT_TITLE, { exact: true })
+    .first()
+    .waitFor({ state: "visible", timeout: STEP_TIMEOUT_MS });
+}
+
+/**
+ * Pick the borrow token in the "Select asset" modal. Prefer the per-symbol testid; fall back to the row
+ * whose text contains the symbol (case-insensitive — symbols are alphanumeric so no regex escaping is
+ * needed). With no token specified, take the first asset row (testid-based; requires the testid build).
+ * Returns the symbol used, for the success log.
+ */
+async function selectAsset(
+  page: Page,
+  log: (m: string) => void,
+  token: string | undefined,
+): Promise<string | undefined> {
+  const row = token
+    ? firstByTestid(
+        page,
+        `[data-testid="asset-select-row-${token.toLowerCase()}"]`,
+        page.getByRole("button").filter({ hasText: new RegExp(token, "i") }),
+      )
+    : page.locator('[data-testid^="asset-select-row-"]').first();
+  if (!(await row.isVisible().catch(() => false)))
+    throw new Error(
+      token
+        ? `Borrow token "${token}" was not found in the asset picker.`
+        : "No borrow token specified and no asset rows were found — re-run with --borrow-token=<symbol>.",
+    );
+  const label = (await row.innerText().catch(() => ""))
+    .replace(/\s+/g, " ")
+    .trim();
+  await row.click();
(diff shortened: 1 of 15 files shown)


services/vault/e2e/real/actions/pegin.ts
```diff
@@ -421,6 +469,21 @@ async function walkStepMachine(
     if ((await activatedViewReached(page)) && activationCount >= expectedVaults)
       return prePeginTxid;
 
+    // Fast-fail on a vault-provider readiness dead-end. If the VP never became ready for WOTS-key
+    // submission, the app skips that vault and it can never activate. Once EVERY expected vault is
+    // skipped there is no route to the activated view — abort now (with a clear, VP-attributed message)
+    // instead of polling out the multi-hour budget. A split where only one sibling is skipped keeps
+    // going (skipped < expectedVaults), matching the app, which continues the ready sibling.
+    const skippedVaults = await countWotsSkippedVaults(page);
+    if (skippedVaults >= expectedVaults)
```
jrwbabylonlab: With one of two split vaults WOTS-skipped this never fires (1 < 2), but the finish gate `activationCount >= expectedVaults` is also unreachable — a skipped vault never activates — so the run burns the full 2.5h budget, the exact slow failure this check targets. Fail once `skippedVaults > 0` and the remaining siblings have activated (i.e. `activationCount >= expectedVaults - skippedVaults`).

services/vault/e2e/real/actions/borrow.ts
```diff
@@ -0,0 +1,433 @@
...
+  const token = ctx.config.borrowToken?.trim();
+  if (!token) return { mode: "max" };
+  try {
+    const max = await fetchMaxBorrow(
+      ctx.config.network,
+      ctx.eth.address,
+      token,
+    );
+    if (max.maxTokens > 0) {
+      const value = formatBorrowAmount(
```
jrwbabylonlab: The 25% default ignores the reserve-liquidity cap the form applies (`effectiveMaxBorrowAmount`) — a pegin-first run can complete a real peg-in and then fail the borrow with exceeds-available-liquidity.

services/vault/e2e/real/actions/borrow.ts
```diff
@@ -0,0 +1,433 @@
...
+  // the read is zero (indexer/state lag) or fails — the form stays the authoritative gate either way.
+  const token = ctx.config.borrowToken?.trim();
+  if (!token) return { mode: "max" };
+  try {
+    const max = await fetchMaxBorrow(
+      ctx.config.network,
+      ctx.eth.address,
+      token,
+    );
+    if (max.maxTokens > 0) {
```
jrwbabylonlab: `maxTokens > 0` can still format to "0" (25% of a tiny max floors to 0 at ≤6 decimals), which fills 0 and stalls at "Enter an amount" until the CTA deadline. Guard on the formatted value instead.

services/vault/e2e/real/actions/borrow.ts
```diff
@@ -0,0 +1,433 @@
...
+    }
+    ctx.log(
+      `⚠️ Computed max borrow for ${token} is 0 — using the form's Max (it will gate the amount).`,
+    );
+  } catch (error) {
+    ctx.log(
+      `⚠️ Could not compute a conservative default (${error instanceof Error ? error.message : error}); using the form's Max.`,
+    );
+  }
+  return { mode: "max" };
```
jrwbabylonlab: If the conservative-max read fails or returns 0 (indexer outage, pegin-first position lag), this silently falls back to borrowing the form's full Max — pinning HF at the 1.05 edge, the exact risk the conservative default exists to avoid. Fail the run instead and reserve full-max for an explicit `--borrow-amount=max`.

services/vault/e2e/real/actions/borrow.ts
```diff
@@ -0,0 +1,433 @@
...
+  token: string | undefined,
+): Promise<string | undefined> {
+  const row = token
+    ? firstByTestid(
+        page,
+        `[data-testid="asset-select-row-${token.toLowerCase()}"]`,
+        page.getByRole("button").filter({ hasText: new RegExp(token, "i") }),
+      )
+    : page.locator('[data-testid^="asset-select-row-"]').first();
+  if (!(await row.isVisible().catch(() => false)))
```
jrwbabylonlab: One-shot probe: the modal shows "Loading assets…" until the oracle-prices query resolves (`openBorrow` only waits for the title), so this can run before any row exists and fail a healthy run with "token not found". Use `row.waitFor({ state: "visible", timeout })` and throw only on timeout.

services/vault/e2e/real/actions/borrow.ts
```diff
@@ -0,0 +1,433 @@
...
+  );
+  const success = page.getByText(BORROW_SUCCESS_RX).first();
+  const txFailed = page.getByText(TX_FAILED_RX).first();
+  const deadline = Date.now() + BORROW_TX_TIMEOUT_MS;
+  while (Date.now() < deadline) {
+    await sweepApprovals(context, page, log);
+
+    if (
+      (await success.isVisible().catch(() => false)) ||
+      (await done.isVisible().catch(() => false))
```
jrwbabylonlab: Any visible "Done" button counts as success even if "Borrow successful" never appeared — and deposit/withdraw/repay modals have Done buttons too. Gate success on the title (or loan-success testid) and use `done` only as the click target.

services/vault/e2e/real/actions/borrow.ts
```diff
@@ -0,0 +1,433 @@
...
+  await fillBorrowAmount(page, log, amount);
+  const cta = await waitForBorrowCta(page, log);
+
+  onStep("borrow-submit");
+  log(
+    "Borrow CTA enabled — submitting (the approver will confirm the MetaMask tx)",
+  );
+  await cta.click();
+
+  await confirmBorrowSuccess(page, context, log, symbol);
```
jrwbabylonlab, line 414: Success is only asserted via the UI. `fetchBorrowContext` already reads `currentDebtUsd` — snapshot it before submit and assert it rose, same baseline pattern as `waitForFreshCollateral`.

services/vault/e2e/real/borrowParams.ts
```diff
@@ -0,0 +1,388 @@
...
+
+/** The depositor's aggregate Aave position (collateral + debt). */
+export interface BorrowContext {
+  /** True when the position currently holds borrowable collateral (`totalCollateralBTC > 0`) — the same
+   *  condition that enables the dashboard's Borrow button (`hasCollateral = collateralBtc > 0`). */
+  hasCollateral: boolean;
+  collateralUsd: number;
+  currentDebtUsd: number;
+  /** The Aave proxy that keys all spoke reads (zero address string when no position exists yet). */
+  proxy: string;
```
jrwbabylonlab: `BorrowContext.proxy` (and the `zeroAddress` import feeding it) and `MaxBorrow.tokenPriceUsd`/`collateralUsd`/`currentDebtUsd`/`hasCollateral` are never read by any caller — drop the dead fields.

services/vault/e2e/real/actions/selectors.ts
```diff
@@ -0,0 +1,24 @@
...
+ * src controls), else a pre-built role/text/css `Locator` for a deployed build that predates the testid.
+ * `.first()` guards against duplicates. The fallback is a `Locator` (not a name) because it differs per
+ * site — `getByRole({name})`, `getByRole().filter({hasText})`, or a raw `page.locator(...)`.
+ */
+export function firstByTestid(
+  page: Page,
+  testid: string,
+  fallback: Locator,
+): Locator {
+  return page.locator(testid).or(fallback).first();
```
jrwbabylonlab, line 27: `or().first()` resolves by DOM order, so a broad fallback matching an earlier element wins even when the testid exists — this doesn't "prefer the stable data-testid". Tighten the fallbacks or fix the comment.

services/vault/e2e/real/networkContracts.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+`;
+
+interface VbtcReserveIdResponse {
+  aaveConfig: { vaultBtcReserveId: string } | null;
+}
+
+/**
+ * Read the vBTC (collateral) reserve id from the indexer's singleton `aaveConfig`. Shared by the pegin
+ * split-minimum and the borrow reserve list so the `aaveConfig(id:1)` key + null contract live in one
```
jrwbabylonlab: "Shared by … the borrow reserve list" isn't accurate — `fetchAaveReserveConfig` in borrowParams.ts duplicates the `aaveConfig(id: 1)` query rather than using this.

services/vault/e2e/real/cli.ts
```diff
@@ -337,6 +362,46 @@ async function resolveConfig(
...
+        ? flags["borrow-amount"]
+        : undefined;
+    if (borrowAmount !== undefined && borrowAmount.toLowerCase() !== "max") {
+      const parsed = Number(borrowAmount);
+      if (!Number.isFinite(parsed) || parsed <= 0)
+        throw new Error(
+          `--borrow-amount must be a positive number of tokens or "max" (got "${borrowAmount}")`,
+        );
+    }
+    if (action === "borrow" && borrowToken === undefined) {
```
jrwbabylonlab: Discovery failures are non-fatal here and validation is skipped when `--borrow-token` is set, so a pegin-first run can move real funds before discovering the token isn't selectable. Hard-fail the preflight before the peg-in.

services/vault/e2e/real/peginParams.ts
```diff
@@ -270,3 +183,120 @@ export async function fetchProviders(
...
+        CAP_SLOT_STATUSES.has(item.status) &&
+        item.applicationEntryPoint.toLowerCase() === adapter
+      )
+        currentCount++;
+    if (!page.vaults.pageInfo.hasNextPage) break;
+    after = page.vaults.pageInfo.endCursor;
+    if (!after) break;
+  }
+
+  return { maxVaults, currentCount };
```
jrwbabylonlab, line 317: This returns the partial count when `MAX_VAULT_PAGES` is exceeded, unlike `fetchVaultsByDepositor` which fails closed — the preflight can undercount occupied slots and let a doomed peg-in proceed.

services/vault/e2e/real/actions/pegin.ts
gbarkhatov: Fixed. The fast-fail now fires as soon as there's no route left to a full completion — `skippedVaults > 0 && activationCount >= expectedVaults - skippedVaults` — instead of only when *every* vault is skipped. So a partial split (one sibling WOTS-skipped, the other activated) aborts immediately with a clear VP-attributed message rather than burning the multi-hour budget; the all-skipped case (single vault, or all split vaults) is covered by the same condition.

services/vault/e2e/real/actions/borrow.ts
gbarkhatov: Fixed. `resolveBorrowAmount` now guards on the *formatted* value (`Number(value) <= 0`), not raw `maxTokens` — so a max that floors to "0" at the token's precision no longer fills 0 and stalls at "Enter an amount"; it fails loudly instead.

services/vault/e2e/real/actions/borrow.ts
gbarkhatov: Fixed. A failed or zero max read now **throws** instead of silently falling back to the form's Max. Full-max is only ever used when explicitly requested via `--borrow-amount=max`, so a read failure can't pin the health factor at the 1.05 edge.

services/vault/e2e/real/actions/borrow.ts
gbarkhatov: Fixed. `selectAsset` now waits for the row (`row.waitFor({ state: "visible", timeout: STEP_TIMEOUT_MS })`) and only throws on timeout, so the "Loading assets…" → rows-render gap can no longer race into a false "token not found" on a healthy run.

services/vault/e2e/real/actions/borrow.ts
gbarkhatov: Fixed. Success is now gated only on borrow-success-specific markers — the "Borrow successful" title or the `loan-success-done-button` testid — never the generic Done role (which deposit/withdraw/repay modals also have). The generic-role Done is used only as the click target, after success is confirmed.

services/vault/e2e/real/actions/borrow.ts
gbarkhatov: Fixed. The borrow now snapshots `currentDebtUsd` before submit and, after the success screen, polls `fetchBorrowContext` until the on-chain debt rises above that baseline (same pattern as `waitForFreshCollateral`), throwing if it doesn't. Verified live on devnet in the re-run: debt `$0.00 → $111.81`.

services/vault/e2e/real/borrowParams.ts
gbarkhatov: Fixed. Dropped the dead fields: `BorrowContext.proxy` (and the now-unused `zeroAddress` import) and `MaxBorrow.{tokenPriceUsd,collateralUsd,currentDebtUsd,hasCollateral}`. `MaxBorrow` is now `{ symbol, decimals, maxTokens }` and `BorrowContext` is `{ hasCollateral, collateralUsd, currentDebtUsd }` — only what callers read.

services/vault/e2e/real/actions/selectors.ts
gbarkhatov: Fixed (comment). Corrected the doc to describe the real semantics: `.or().first()` is a match-either union resolved in DOM order, not a testid priority. It behaves correctly here because each testid and its fallback target the *same* control; the comment now says so and warns to keep each fallback tightly scoped so a broad one can't match an earlier element and win the `.first()`.

services/vault/e2e/real/networkContracts.ts
gbarkhatov: Fixed (comment). Corrected: `fetchVbtcReserveId` is used by the pegin split-minimum (`peginParams.fetchMinDepositForSplitBtc`). `borrowParams.fetchAaveReserveConfig` intentionally reads the id as part of one combined `aaveConfig` + `aaveReserves` query (single round-trip), so it doesn't call this — the doc no longer claims it's shared by the borrow path.

services/vault/e2e/real/cli.ts
gbarkhatov: Fixed. The borrow pre-flight now fetches the live reserve list up front and, when `--borrow-token` is set, validates it (hard-fail listing the available tokens) before the run — so an unselectable token can't waste a funded pegin-first peg-in. It also canonicalizes the symbol casing. On a discovery outage it warns and lets the form gate rather than hard-failing on a transient.

services/vault/e2e/real/peginParams.ts
gbarkhatov: Fixed. `fetchVaultCountCap` now fails **closed** — it throws if the depositor's vaults don't fit within `MAX_VAULT_PAGES` instead of returning a partial (undercounted) slot count, matching the app's `fetchVaultsByDepositor`. The run.ts pre-flight treats the throw as non-fatal (warn → let the form gate), so it never blocks a legit run; it just refuses to under-report.

services/vault/e2e/real/actions/borrow.ts
gbarkhatov: Acknowledged — valid, replying without a fix in this PR. The 25% default is collateral-based and doesn't apply the form's reserve-liquidity cap (`effectiveMaxBorrowAmount`), so on a thin reserve a pegin-first run could complete the peg-in and then hit "exceeds available liquidity". We're deferring the fix: the Hub liquidity read (`getAssetLiquidity`) isn't exposed by the SDK — the app hand-rolls it in `aaveHub.ts` — so a faithful cap means replicating a Hub ABI read in the CLI, and on devnet/testnet the reserves are ~10M so the 25% default never approaches the cap. The live form stays the authoritative gate (and `waitForBorrowCta` fast-fails on the exceeds-liquidity label), so the worst case is a wasted peg-in on a genuinely thin reserve. Tracking as a follow-up.

jrwbabylonlab (approved):

PR #2040, opened 2026-07-14
fix(vault): surface deposit fee/wallet failures instead of spinning

jrwbabylonlab:
The deposit form's CTA showed "Calculating fees..." indefinitely whenever an
upstream dependency failed, giving the user no indication anything was wrong.
This fixes three independent silent-forever paths so each failure surfaces an
actionable message or recovery CTA.

- **Network fee-rate failure** (`validations.ts`): `getDepositCtaState` returned
the null-fee "Calculating fees..." label before the `isFeeError` branch could
ever run, making that branch dead code — every fee error also has `fee === null`.
Moved the fee-error check ahead of the null-fee label so a failed estimate shows
its message; removed the now-duplicate later block. Guarded on `btcBalance > 0n`
so "No available balance" still wins for an empty wallet.
- **Silent wallet public-key failure** (`useBtcPublicKey.ts`): the hook swallowed
`getPublicKeyHex()` errors and left the key `undefined`, permanently disabling
the downstream claim-value query. It now returns `{ publicKey, error, refetch }`;
the error promotes the CTA to the existing "Reconnect Wallet" recovery action,
and `refetch()` (called after a successful reconnect) clears it deterministically.
- **Keeper-less registry** (`useDepositPageForm.ts`): a settled registry with
providers but zero vault keepers could never enable the `minPeginFee` query.
Now surfaced as a terminal `minPeginFeeError` instead of an infinite spinner.

Note: the trigger that exposed this (a WAF 403 on the signet mempool fees
endpoint) is an infra/region issue — this change makes the app *report* the
failure rather than fixing the endpoint.

image

### DIFF
--- a/services/vault/src/components/simple/PostDepositContinuationContent.tsx
+++ b/services/vault/src/components/simple/PostDepositContinuationContent.tsx
@@ -28,7 +28,7 @@ export function PostDepositContinuationContent({
   onClose,
 }: PostDepositContinuationContentProps) {
   const { connected: btcConnected } = useBTCWallet();
-  const btcPublicKey = useBtcPublicKey(btcConnected);
+  const { publicKey: btcPublicKey } = useBtcPublicKey(btcConnected);
   const { activities, pendingPegins } = useVaultDeposits(depositorEthAddress);
 
   const scoped = useMemo(() => {
--- a/services/vault/src/components/simple/SimpleDeposit.tsx
+++ b/services/vault/src/components/simple/SimpleDeposit.tsx
@@ -9,6 +9,7 @@ import { useAddressScreening } from "@/context/addressScreening";
 import { useGeoFencing } from "@/context/geofencing";
 import { ProtocolParamsProvider } from "@/context/ProtocolParamsContext";
 import { useBTCWallet, useETHWallet } from "@/context/wallet";
+import { COPY } from "@/copy";
 import { useBtcWalletState } from "@/hooks/deposit/useBtcWalletState";
 import { useDepositPeginFee } from "@/hooks/deposit/useDepositPeginFee";
 import { useDialogStep } from "@/hooks/deposit/useDialogStep";
@@ -130,6 +131,8 @@ function SimpleDepositContent({
     isSplitAmountTooLow,
     depositorClaimValue,
     depositorClaimValueError,
+    btcPublicKeyError,
+    refetchBtcPublicKey,
     ordinalsCheckPending,
     validateForm,
     resetForm,
@@ -301,6 +304,10 @@ function SimpleDepositContent({
     try {
       await reconnectBtcWallet();
       setWalletConnectionError(null);
+      // reconnect() re-auths the raw provider without emitting a connector
+      // event, so re-read the public key explicitly to clear a stuck
+      // btcPublicKeyError once the wallet is unlocked again.
+      refetchBtcPublicKey();
     } catch {
       // The underlying provider throws dev-facing strings (e.g. "BTC wallet
       // provider returned an empty address"). Surface a single polished
@@ -331,7 +338,7 @@ function SimpleDepositContent({
     // (which triggers the wallet's unlock/re-authorization prompt) instead of
     // attempting another deposit. The deposit attempt itself is only retried
     // once the user successfully reconnects and the error/lock state clears.
-    if (walletConnectionError || isBtcWalletLocked) {
+    if (walletConnectionError || isBtcWalletLocked || btcPublicKeyError) {
       await handleReconnectWallet();
       return;
     }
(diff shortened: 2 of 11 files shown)


services/vault/src/components/simple/SimpleDeposit.tsx
```diff
@@ -301,6 +304,10 @@ function SimpleDepositContent({
     try {
       await reconnectBtcWallet();
       setWalletConnectionError(null);
+      // reconnect() re-auths the raw provider without emitting a connector
+      // event, so re-read the public key explicitly to clear a stuck
+      // btcPublicKeyError once the wallet is unlocked again.
+      refetchBtcPublicKey();
```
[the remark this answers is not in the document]
jrwbabylonlab: Good catch — now awaiting the pubkey re-read before clearing the reconnecting state, so the CTA can't look actionable while the error is still stale. Added a generation guard so a slow initial read can't clobber the fresh key either.

services/vault/src/services/deposit/validations.ts
```diff
@@ -391,6 +391,18 @@ export function getDepositCtaState(params: DepositCtaParams): DepositCtaState {
     return { disabled: true, label: "Calculating fees..." };
   }
 
+  // Surface a terminal network-fee failure before `getDepositButtonLabel`:
+  // a failed estimate also has `estimatedFeeSats` absent, so the null-fee
+  // "Calculating fees..." label would otherwise shadow the error forever.
+  // Skip when there's no balance — "No available balance" is more actionable
+  // than a fee error the user can't act on anyway.
+  if (params.btcBalance > 0n && params.isFeeError) {
```
gbarkhatov, line 399: Non-blocking, ordering nit. This branch now sits **after** the `minPeginFee === null` gate (line 390) and **before** `commissionUnavailable` (415) / `ordinalsCheckPending` (419):

- **After 390:** when the fee-rate estimate has failed but the independent `minPeginFee` query is still loading, the CTA transiently shows "Calculating fees…" and shadows the fee error until `minPeginFee` settles. Self-resolving (a keeper-less config trips line 377 first, and with keepers present the query settles), but moving this branch above line 390 would surface the error immediately.
- **Before 415/419:** the old block ran after both, so a terminal fee-rate failure now shadows the transient "Loading commission…" / "Checking for inscriptions…" labels. Defensible (terminal error > transient spinner) — just flagging it's a behavior change beyond un-deadcoding the branch.

services/vault/src/hooks/deposit/useDepositPageForm.ts
```diff
@@ -291,6 +306,18 @@ export function useDepositPageForm(): UseDepositPageFormResult {
     () => vaultKeepers.map((vk) => vk.btcPubKey),
     [vaultKeepers],
   );
+  // A settled registry with selectable providers but zero keepers can never
```
gbarkhatov, line 309: Two non-blocking nits on this block:

1. The comment says "selectable providers" but the guard uses `rawProviders.length`, which includes metadata-rejected / unhealthy providers that can't actually be selected. No wrong output (the CTA only reads `minPeginFeeError` past the `!hasProvider` gate), but the comment overstates what the condition checks.
2. `new Error("No vault keepers registered for the application")` (line 317) is fine as-is — it's internal and the CTA renders the generic "Fee estimate unavailable" — just note it would bypass the copy surface if anything ever displayed `minPeginFeeError.message` directly.

services/vault/src/components/simple/PostDepositContinuationContent.tsx
```diff
@@ -28,7 +28,7 @@ export function PostDepositContinuationContent({
   onClose,
 }: PostDepositContinuationContentProps) {
   const { connected: btcConnected } = useBTCWallet();
-  const btcPublicKey = useBtcPublicKey(btcConnected);
+  const { publicKey: btcPublicKey } = useBtcPublicKey(btcConnected);
```
gbarkhatov, line 31: Non-blocking, confirm intentional. This drops the newly-available `error` field (same pattern in `usePendingDeposits.ts:30`), so a locked-wallet pubkey read yields `undefined` here with no user signal. Unchanged from prior behavior (the hook returned `undefined` before too), and these are passive list/continuation contexts rather than the deposit CTA — so silence is defensible. Flagging only because the error is now available if you'd want to surface it here too.

services/vault/src/copy.ts
```diff
@@ -811,6 +811,8 @@ export const COPY = {
       unlockButton: "Unlock wallet",
       unlocking: "Unlocking wallet...",
     },
+    publicKeyUnavailable:
```
gbarkhatov, line 814: Organizational nit (non-blocking). `publicKeyUnavailable` is a sibling of `wallet.liveness` / `wallet.locked`, but its wording mirrors `wallet.liveness.emptyAddress` almost exactly ("Your BTC wallet did not return an address/public key. Please reconnect your wallet and try again."). Nesting it under `wallet.liveness` would group like with like.

gbarkhatov (approved):
Approving. This cleanly fixes three silent-forever CTA paths (dead `isFeeError` branch, swallowed pubkey read, keeper-less registry), and the tests cover the real failure shapes — including the stale-mount-read vs refetch race in the generation guard. No blocking issues. Traced the `useBtcPublicKey` return-shape change across all three call sites (all migrated) and confirmed the deleted `isFeeError` block was genuinely unreachable in production, so no behavior is lost. A few non-blocking, author's-discretion nits inline.

jrwbabylonlab:
same here. i will check comments tmr. they all nits from quick glance. @gbarkhatov

PR #2045, opened 2026-07-14
chore(vault): remove dead code from services/vault

jrwbabylonlab:
Removes dead code from `services/vault` — 794 lines across 38 files. No behavior change.

Everything deleted was verified to have zero consumers across the whole monorepo
(`services/*`, `packages/*`, `e2e/`, `.github/`), not just within the vault app.

**Whole files (4)**
- `components/shared/BackButton.tsx`, `components/shared/SubmitModal.tsx` — no importers
(simple-staking has its own separate `SubmitModal`)
- `hooks/useProtocolParams.ts` — the hook it was named for is long gone; the file was a
stale interface
- `utils/tokenConversion.ts` — orphaned once its only caller went

**Superseded flow code**
`startDeposit` / `confirmReview` / `DepositStep.REVIEW` implemented a three-stage
FORM → REVIEW → SIGN deposit. There is no REVIEW screen: nothing renders or compares that
step, so calling `startDeposit` would have navigated to a blank state. `SimpleDeposit`
consolidated this into one handler that goes straight to SIGN.

**Test-only production code** (code + the tests that were its only callers)
`formatLLTV`, `getFormattedRepayAmount`, `transformVaultsToActivities`,
`isTransientPollingError`, `runHealthChecks`, `transformFormToTransactionData`.
Their live siblings are untouched — `transformVaultToActivity` (singular) and
`isTerminalPollingError` both stay.

**Dead context/state members**
`clearOptimisticStatus`, `DepositStateData`, `processing`/`setProcessing`, four
`PendingVaultsContext` flags + `clearPendingCollateralVaultIds`, and never-read
`AppError` fields.

Two of the `PendingVaultsContext` flags could never fire: `hasPendingAdd` reads
`operations.includes("add")`, but the only writer is `markVaultsAsPending(ids, "withdraw")`.
The surviving consumer (`useAaveVaults`) already recomputes `pendingVaults.size > 0` locally.

**`ProtocolParamsContext.getUniversalChallengersByVersion`**
Removed. Its JSDoc said "use for payout signing existing vaults", but no caller ever used it,
and its body returned `[]` on a cache miss. Payout/refund signing correctly read the
version-bonded challenger set on-chain via the SDK readers keyed on
`vault.universalChallengersVersion`, and throw when empty — so this was an unused accessor
that invited a silent-empty-set bug on a critical path.

**`utils/queryKeys.ts`**
Two `invalidateQueries` calls were silent no-ops — no query is registered under the
`borrowableVaults` or `peginRequests` namespaces. Invalidations for the real namespaces
(`vaults`, `aaveUserPosition`) are preserved.

**Copy**
Two keys had been orphaned by components inlining the literal instead — a copy-policy
violation. Rather than delete, the inlines are migrated back to `copy.ts`
(`COPY.common.loading`, new `COPY.pegin.dateLabel`). No user-visible string changes.
Also drops `NEXT_PUBLIC_FF_SIMPLIFIED_TERMS` from `.env.example` / README — it is read
nowhere in any code.


services/vault/src/utils/queryKeys.ts
```diff
@@ -54,16 +24,6 @@ export async function invalidateVaultQueries(
...
-    }),
-
-    // Invalidate pegin requests to update vault usage status (isInUse)
-    queryClient.invalidateQueries({
-      queryKey: queryKeys.peginRequests(address),
-    }),
-
     // Invalidate vaults query to refresh vault list
     queryClient.invalidateQueries({
       queryKey: [VAULTS_QUERY_KEY, address],
```
gbarkhatov, line 29: ✅ Verified safe: the removed `borrowableVaults` and `peginRequests` invalidations were genuine no-ops — neither key is registered by any query anywhere in the monorepo, so nothing could go stale under them. The real invalidations (`VAULTS_QUERY_KEY` here and `aaveUserPosition` just below) are preserved, so post-transaction refresh behavior is unchanged.

services/vault/src/storage/notificationPromptStorage.ts
```diff
@@ -17,13 +15,9 @@ export function loadNotificationPromptDismissed(): boolean {
   }
 }
 
-export function setNotificationPromptDismissed(dismissed: boolean): void {
+export function dismissNotificationPrompt(): void {
```
gbarkhatov, line 18: Minor behavior narrowing (non-blocking): the old `setNotificationPromptDismissed(false)` re-offer/revert path is gone — dismissal is now one-way. No caller ever passed `false`, so this is dead-code removal rather than a regression. Worth a mental note only if a future "re-enable notifications prompt" feature is ever planned.

services/vault/src/context/ProtocolParamsContext.tsx
```diff
@@ -112,14 +110,6 @@ export function ProtocolParamsProvider({
     return ucData.byVersion.get(ucData.latestVersion) ?? [];
   }, [ucData]);
 
-  const getUniversalChallengersByVersion = useCallback(
```
gbarkhatov, line 115: Nice cleanup on a critical path: this accessor returned `[]` on a cache miss and had zero callers. Payout/refund signing reads the version-bonded challenger set on-chain via the SDK readers (which throw on an empty set), so removing this eliminates a latent silent-empty-set footgun rather than dropping coverage.

services/vault/src/services/vault/vaultPayoutSignatureService.ts
```diff
@@ -50,14 +50,6 @@ const VP_COMMISSION_BPS_EXCLUSIVE_MAX = 10_000;
...
 
-export interface PayoutVaultKeeper {
-  btcPubKey: string;
-}
-
-export interface PayoutUniversalChallenger {
-  btcPubKey: string;
-}
-
 export interface PrepareSigningContextParams {
```
gbarkhatov, line 53: Heads-up for reviewers: this file is on the CLAUDE.md "critical paths" list (#3, payout/refund signing), which mandates two reviewers. The change here is only the removal of two unused local interfaces (`PayoutVaultKeeper`, `PayoutUniversalChallenger`) — no signing logic is touched — so the surface is trivial, but the two-reviewer rule still applies.

services/vault/src/context/error/ErrorProvider.tsx
```diff
@@ -61,22 +54,20 @@ export const ErrorProvider: FC<ErrorProviderProps> = ({ children }) => {
...
-            error: errorData,
-            modalOptions: {
-              retryAction: displayOptions?.retryAction,
-              noCancel: displayOptions?.noCancel,
-              blocking: displayOptions?.blocking,
-            },
-          };
-        });
-      }
+      setState((prev) => {
```
gbarkhatov, line 57: Looks like a behavior change but isn't: the removed `showModal` gate (`displayOptions?.showModal ?? true`) defaulted to `true`, and no caller in the vault app ever passed `showModal`. The modal was always shown before and still is — `setState` is just unconditional now. Confirmed no `showModal` usage remains in `services/vault`.

gbarkhatov (approved):
## Review: dead-code cleanup ✅

Traced every removed symbol across `services/*`, `packages/*`, and `e2e/` — all confirmed unreferenced, and the behavior-sensitive deletions (error-modal gating, query invalidations, `AppError` fields, notification dismissal) verified dead within the vault app. Apparent collisions (`getUniversalChallengersByVersion` SDK reader, `hasPendingOperations` local in `useAaveVaults`, simple-staking's own `SubmitModal`/`showModal`/`setProcessing`, `WithdrawStep.REVIEW`) are all distinct from the removed symbols.

No behavior change; the copy migration (`COPY.common.loading`, renamed `COPY.pegin.dateLabel`) is consistent and the rendered label was already "Date".

Left a few inline notes — all non-blocking reassurances, plus a heads-up that two touched files are on the CLAUDE.md critical-paths list (only unused-interface removals there).

Recommend confirming `pnpm --filter vault run test` is green in CI before merge — the deleted test files were the only callers of the removed production code, so the suite should pass unchanged.

Approving.

PR #2049, opened 2026-07-14
test(vault): repay E2E CLI

### LINKED ISSUE #2048: test(vault): add a repay action to the real-wallet E2E CLI (step 6)
## Step 6 — Repay a loan drawn against BTC-Vault collateral (real-wallet E2E CLI)

Extends the real-wallet E2E CLI (`services/vault/e2e/real/`) to the next step of the depositor lifecycle: **repay**. Builds on connect (https://github.com/babylonlabs-io/babylon-toolkit/pull/2005), single-vault pegin (https://github.com/babylonlabs-io/babylon-toolkit/pull/2012), two-vault split pegin (https://github.com/babylonlabs-io/babylon-toolkit/pull/2024), and borrow (https://github.com/babylonlabs-io/babylon-toolkit/pull/2033).

### Goal

Add a `repay` action that pays down (or fully clears) a loan drawn against a BTC-Vault position, end-to-end on real signet + Sepolia, and finishes at the "Repay successful" screen with the debt confirmed to have fallen on-chain. With this, the CLI covers the whole depositor lifecycle: **pegin → borrow → repay**, in one session.

### Scope — the repay action

Three shapes, selected by the CLI:

- **Repay an existing loan** (default): a pre-flight reads the position from real data and refuses before the browser if there is no debt to repay.
- **Borrow + repay** (`--borrow-first`): borrow against existing collateral first, then repay the new loan — one browser session.
- **Pegin + borrow + repay** (`--borrow-first --pegin-first`): peg in fresh collateral, borrow against it, then repay — the full lifecycle in one run.

The repay amount defaults to a conservative fraction of the outstanding debt (partial, always affordable, leaves the loan open so the test is repeatable), overridable via `--repay-token` / `--repay-amount` (`--repay-amount=max` clicks the form's Max for a full clear). The live form stays the authoritative gate; the CLI's computed values are best-effort defaults.

### Two behaviours the repay flow must handle (that borrow does not)

- A depositor with exactly **one** loan is routed straight to the repay form; only **multiple** loans open the "Select asset" picker. The action waits for either landing.
- A repay is **one or two** MetaMask transactions: an ERC-20 approve of the debt token to the adapter (only when the current allowance is insufficient), then the repay. The pop-up approver confirms whichever appear.

- **Shared "borrow, optionally pegin-first" seam** — the borrow flow's "peg in (optional) then borrow" sequence is extracted into one exported function that both the borrow action and the repay action call, so `repay --borrow-first --pegin-first` reuses the identical, already-proven pegin + borrow path instead of duplicating it.
- **Borrow-fail guard** — if the pegin/borrow leg of a `--borrow-first` run fails, the run stops before the repay (there is no new loan to repay) with a clear message, rather than falling through.
- **Connect approval sweep** — the "waiting for connected state" step now sweeps approval pop-ups while it polls, so MetaMask's intermittent "Review permissions / Use your enabled networks" prompt gets confirmed instead of hanging the connect. This hardens connect for every action (connect / pegin / borrow / repay).
- **Params-layer + selector DRY cleanup** — repay reuses the borrow pre-flight's on-chain plumbing (reserve list, position/proxy, Core Spoke) and the SDK's debt read; a small `tokenAmount` helper and the loan-form/success-modal selectors shared by borrow and repay were de-duplicated into shared modules; two `data-testid`s were added to the repay controls in `services/vault/src`.

### Acceptance

- A headed, real run on devnet completes and the on-chain debt change is confirmed, for: repay an existing loan (partial and full Max clear); borrow + repay in one session; and the full pegin → borrow → repay lifecycle in one session.
- Existing pegin (single + split) and borrow still pass — no regression.

- Withdraw action.
- Resuming an in-flight pegin (still open from earlier steps).

gbarkhatov:
# Repay action for the real-wallet E2E CLI (step 6)

Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2048

## What this does

Adds a `repay` action to the real-wallet end-to-end CLI (`services/vault/e2e/real/`) — the tool that drives the vault dApp with real Chrome wallet extensions (signet BTC + Sepolia ETH) through the depositor lifecycle. With repay landed, the CLI now covers the whole journey a real depositor takes: **pegin → borrow → repay**.

You run it with one command. Three shapes, picked by flags:

- `--action=repay` — repay a loan you already have.
- `--action=repay --borrow-first` — borrow against collateral you already have, then repay it, in one browser session.
- `--action=repay --borrow-first --pegin-first` — peg in fresh BTC collateral, borrow against it, then repay — the full lifecycle in one run.

By default it repays a conservative 25% of the debt (a partial repay that always fits your wallet balance and leaves the loan open, so the test can be re-run). `--repay-amount=max` clears the loan in full by clicking the form's own "Max" button, and `--repay-amount=<n>` repays an exact amount.

## Why

The CLI already did connect, pegin, split-pegin, and borrow (https://github.com/babylonlabs-io/babylon-toolkit/pull/2033). Repay was the missing piece to exercise the full depositor lifecycle against real wallets and real chains — so we can prove, end to end, that a user can borrow against their BTC Vault and pay it back, with the debt change confirmed on-chain (not just from the UI saying "success").

## How it works (and the two things repay does that borrow doesn't)

The action drives the live UI only — it never rebuilds transactions or reimplements protocol logic. It reads the depositor's on-chain debt (via the SDK) to size a safe default amount, fills the form, and lets the live form's own validation be the authority. After the success screen it re-reads the on-chain debt and asserts it actually fell.

Two behaviours are specific to repay and are handled explicitly:

- **A single loan skips the asset picker.** With one loan the dApp routes straight to the repay form; only with multiple loans does the "Select asset" screen appear. The action waits for whichever shows up.
- **Repay can be one or two wallet transactions.** Repaying pulls the debt token from your wallet, so it first needs an ERC-20 approval — but only when the current allowance is too low. So sometimes it's just the repay (1 pop-up), sometimes approve-then-repay (2 pop-ups). The pop-up approver confirms whichever appear; we never hard-assume a fixed number.

## Approach — and why this one

**Reuse over rebuild.** Rather than write a second copy of the borrow flow for the `--borrow-first` case, we extracted the borrow flow's "optionally peg in, then borrow" sequence into a single shared function that both the borrow action and the repay action call. So `repay --borrow-first --pegin-first` runs the exact same pegin + borrow path we already proved for borrow — no duplication, and the pegin logic has one home. In the same spirit, repay reuses the borrow pre-flight's on-chain reads (reserve list, position, Core Spoke) and the SDK's debt read, and the UI selectors shared by both forms (amount input, Max button, "Select asset" picker, success modal) were moved into one shared module instead of being copied into each action.

**Conservative default, explicit full-clear.** We considered defaulting to a full repay, but a full clear depends on the wallet holding at least the accrued debt and can leave interest "dust" otherwise. A 25% partial repay is always affordable and repeatable, so it's the default; a full clear is available on demand via `--repay-amount=max`, which clicks the form's real Max button (the app then repays debt + a tiny buffer and refunds the excess, so there's no dust when the balance covers it).

**Fail loud, never guess.** If the debt read fails, or the amount would round to zero, the action throws instead of silently doing something surprising. And if the borrow leg of a `--borrow-first` run fails, the run stops before the repay (there's no new loan to repay) with a clear message.

## Two robustness fixes included (help every action, not just repay)

- **Connect approval sweep.** MetaMask sometimes shows an extra "Review permissions / Use your enabled networks" prompt after connecting. The connect step now sweeps and confirms approval pop-ups while it waits for the connected state, so that prompt no longer hangs the run. This benefits connect, pegin, borrow, and repay alike.
- **Borrow-fail guard.** A `--borrow-first` run that hits a failed borrow (e.g. a transient wallet/RPC error) now stops cleanly before the repay, instead of falling through.

## How to test

Run the CLI via its package script (from the repo root). With no flags it drops into an interactive menu; flags make it non-interactive. Localhost against devnet (real signet + Sepolia, real funds — needs a funded wallet):

```
pnpm --filter vault run e2e:cli --target=localhost --network=devnet \
  --btc=unisat --eth=metamask --action=repay --data=real --borrow-first --yes
```

Drop `--borrow-first` to repay an existing loan; add `--pegin-first` for the full pegin → borrow → repay; add `--repay-amount=max` to clear a loan in full. `pnpm --filter vault run e2e:cli` with no flags launches the interactive menu.

Verified live on devnet, each confirmed on-chain: repay-only ($111.81 → $83.86 and a full Max clear $132.67 → $0.00); borrow + repay in one session ($83.86 → $176.89 → $132.67, plus $0 → $114 → $0); and the full pegin → borrow → repay (peg in 0.01 sBTC, borrow $0 → $227.72, repay → $170.79). Existing pegin (single + split) and borrow still pass; `pnpm exec eslint` and `pnpm exec tsc` are clean.

## Files

- New: `e2e/real/repayParams.ts` (on-chain debt/balance reads), `e2e/real/actions/repay.ts` (the action), `e2e/real/tokenAmount.ts` (shared amount formatter).
- Changed: `e2e/real/actions/borrow.ts` (extracted the shared pegin-first-then-borrow function), `actions/index.ts`, `actions/selectors.ts` (shared loan-form selectors), `actions/walletConnect.ts` (connect approval sweep), `borrowParams.ts`, `cli.ts`, `config.ts`, `run.ts`, `timing.ts`.
- Two `data-testid`s added to the repay controls in `services/vault/src` (`loans-repay-button`, `repay-submit-button`).

### DIFF
--- a/services/vault/e2e/real/actions/repay.ts
+++ b/services/vault/e2e/real/actions/repay.ts
@@ -46,6 +46,7 @@ import { runBorrowWithOptionalPegin } from "./borrow";
 import { startRecording } from "./recording";
 import {
   AMOUNT_INPUT,
+  ASSET_ROW_TESTID_PREFIX,
   ASSET_SELECT_TITLE,
   DONE_BUTTON_RX,
   firstByTestid,
@@ -154,9 +155,11 @@ async function resolveRepayAmount(
  * Open the repay flow from the dashboard: click Loans → Repay, then wait for EITHER the "Select asset"
  * picker (multiple loans) OR the repay form itself (a single loan skips the picker — see the app's
  * DashboardPage.handleRepay). Returns whether the picker appeared, so the caller knows to select an
- * asset. The Repay button only renders when the position has a loan and is enabled once connected; a
- * loan created earlier in the same session (borrow-first) takes a moment to propagate, so we poll for it
- * to enable rather than failing on the first check.
+ * asset. The Repay button is only RENDERED when the position has a loan (`{hasLoans && …}`) and is
+ * enabled once connected — so after a borrow-first the button's VISIBILITY, not just its enabled state,
+ * lags while the new loan propagates. We therefore poll for it to appear AND enable within a single
+ * window (not a short visibility wait before a long enable wait, which would abort early on a
+ * slow-to-render button).
  */
 async function openRepay(
   page: Page,
@@ -167,12 +170,12 @@ async function openRepay(
     LOANS_REPAY_TESTID,
     page.getByRole("button", { name: REPAY_BUTTON_RX }),
   );
-  await repay.waitFor({ state: "visible", timeout: STEP_TIMEOUT_MS });
   const deadline = Date.now() + REPAY_BUTTON_ENABLE_TIMEOUT_MS;
-  let enabled = await repay.isEnabled().catch(() => false);
+  let enabled = false;
   while (!enabled && Date.now() < deadline) {
-    await page.waitForTimeout(FORM_SETTLE_MS);
-    enabled = await repay.isEnabled().catch(() => false);
+    if (await repay.isVisible().catch(() => false))
+      enabled = await repay.isEnabled().catch(() => false);
+    if (!enabled) await page.waitForTimeout(FORM_SETTLE_MS);
   }
   if (!enabled)
     throw new Error(
@@ -208,7 +211,9 @@ async function openRepay(
 /**
  * Pick the debt token in the "Select asset" picker (repay mode). Prefer the per-symbol testid; fall back
  * to the row whose text contains the symbol. With no token specified, take the first row. Waits for the
- * row (the picker's rows are the user's loans, available immediately). Returns the symbol used.
+ * row (the picker's rows are the user's loans, available immediately). Returns the token SYMBOL — read
+ * from the chosen row's `data-testid` (`asset-select-row-<symbol>`), NOT its free-text label — so the
+ * caller's debt lookup (`resolveRepayAmount`) can match it against `debt.symbol`.
  */
 async function selectAsset(
   page: Page,
@@ -218,10 +223,10 @@ async function selectAsset(
   const row = token
     ? firstByTestid(
         page,
-        `[data-testid="asset-select-row-${token.toLowerCase()}"]`,
+        `[data-testid="${ASSET_ROW_TESTID_PREFIX}${token.toLowerCase()}"]`,
         page.getByRole("button").filter({ hasText: new RegExp(token, "i") }),
       )
-    : page.locator('[data-testid^="asset-select-row-"]').first();
+    : page.locator(`[data-testid^="${ASSET_ROW_TESTID_PREFIX}"]`).first();
   const appeared = await row
     .waitFor({ state: "visible", timeout: STEP_TIMEOUT_MS })
     .then(() => true)
@@ -232,12 +237,17 @@ async function selectAsset(
         ? `Repay token "${token}" was not found in the asset picker within ${Math.round(STEP_TIMEOUT_MS / MS_PER_SECOND)}s.`
         : "No repay token specified and no loan rows were found in the picker.",
     );
-  const label = (await row.innerText().catch(() => ""))
-    .replace(/\s+/g, " ")
-    .trim();
+  // An explicit token wins; otherwise read the symbol from the row's testid (the no-token branch selects
+  // rows BY that prefix, so the attribute is always present on the chosen row).
+  let symbol = token;
+  if (!symbol) {
+    const testid = await row.getAttribute("data-testid").catch(() => null);
+    if (testid?.startsWith(ASSET_ROW_TESTID_PREFIX))
+      symbol = testid.slice(ASSET_ROW_TESTID_PREFIX.length);
+  }
   await row.click();
-  log(`Selected repay token: ${token ?? label}`);
-  return token ?? label;
+  log(`Selected repay token: ${symbol ?? "(unknown)"}`);
+  return symbol;
 }
 
 /** Enter the repay amount: click the form's Max button, or fill the numeric input. */
--- a/services/vault/e2e/real/actions/repay.ts
+++ b/services/vault/e2e/real/actions/repay.ts
@@ -1,0 +1,505 @@
+/**
+ * The "repay" action: pay down (or clear) a loan drawn against a BTC-Vault position, end-to-end on real
+ * Sepolia. Two shapes, selected by the CLI:
+ *   - REUSE (default): repay a loan the depositor already holds. run.ts refuses the run before the
+ *     browser if there's no debt, so we land on the dashboard with the Repay button ready.
+ *   - BORROW-FIRST (`--borrow-first`): borrow first (the shared `runBorrowWithOptionalPegin`), then
+ *     repay that loan — all in one browser session. Adding `--pegin-first` pegs in fresh collateral
+ *     ahead of the borrow, giving the full pegin → borrow → repay lifecycle in one run.
+ *
+ * The repay flow is short but has two shapes the CLI must handle that borrow doesn't:
+ *   - A depositor with exactly ONE loan is routed straight to the repay form; only MULTIPLE loans open
+ *     the "Select asset" picker. So after clicking Repay we wait for EITHER the picker OR the form.
+ *   - Repaying can take ONE or TWO MetaMask transactions: an ERC-20 approve of the debt token to the
+ *     adapter (only when the current allowance is insufficient), then the repay. The CTA reads
+ *     "Processing…" across both; the pop-up approver confirms whatever appears.
+ *
+ * Selectors are testid-first (added to the src repay controls, mirroring borrow) with tolerant
+ * text/role/class fallbacks. No SDK / product logic is reimplemented — the amount is a best-effort
+ * default and the live form is the authoritative gate (its Max button + validation).
+ *
+ * NEVER run without an explicit go-ahead: it moves real value (a debt repayment, plus real ERC-20
+ * approval + repay gas).
+ */
+import type { BrowserContext, Locator, Page } from "@playwright/test";
+
+import { fetchBorrowContext } from "../borrowParams";
+import { type NetworkName } from "../config";
+import {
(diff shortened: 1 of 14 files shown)


services/vault/e2e/real/actions/repay.ts
```diff
@@ -0,0 +1,505 @@
...
+    LOANS_REPAY_TESTID,
+    page.getByRole("button", { name: REPAY_BUTTON_RX }),
+  );
+  await repay.waitFor({ state: "visible", timeout: STEP_TIMEOUT_MS });
+  const deadline = Date.now() + REPAY_BUTTON_ENABLE_TIMEOUT_MS;
+  let enabled = await repay.isEnabled().catch(() => false);
+  while (!enabled && Date.now() < deadline) {
+    await page.waitForTimeout(FORM_SETTLE_MS);
+    enabled = await repay.isEnabled().catch(() => false);
+  }
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in the latest push. Good catch — the Repay button is conditionally rendered (`{hasLoans && <Button>}` in `LoansSection.tsx`), unlike the always-rendered Borrow button, so after a `--borrow-first` run its *visibility* (not just its enabled state) lags while `hasLoans` propagates. The 30s upfront `waitFor` was gating that visibility *before* the 180s enable-poll even started, so the intended propagation window was effectively capped at 30s.

`openRepay` now folds visibility into a single `REPAY_BUTTON_ENABLE_TIMEOUT_MS` (180s) poll — each tick checks `isVisible()` then `isEnabled()`, and only throws after the full window. Verified live: a borrow-first run reached the repay form cleanly right after the borrow.

services/vault/e2e/real/actions/repay.ts
```diff
@@ -0,0 +1,505 @@
...
+  if (!appeared)
+    throw new Error(
+      token
+        ? `Repay token "${token}" was not found in the asset picker within ${Math.round(STEP_TIMEOUT_MS / MS_PER_SECOND)}s.`
+        : "No repay token specified and no loan rows were found in the picker.",
+    );
+  const label = (await row.innerText().catch(() => ""))
+    .replace(/\s+/g, " ")
+    .trim();
+  await row.click();
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in the latest push. `selectAsset` no longer returns the row's free-text label as the symbol — in the no-token path it now reads the symbol from the chosen row's `data-testid` (`asset-select-row-<symbol>`) and strips the prefix, so the downstream `resolveRepayAmount` lookup (`debts.find(d => d.symbol === …)`) matches. The no-token branch already selects rows *by* that testid prefix, so the attribute is guaranteed present on the chosen row.

Applied the same change to `borrow.ts`'s `selectAsset` for consistency (there the returned symbol only feeds a success-log line, so it was cosmetic — but the two functions should stay in sync), and extracted the shared `ASSET_ROW_TESTID_PREFIX` into `selectors.ts`.

jrwbabylonlab (dismissed):

services/vault/e2e/real/repayParams.ts
```diff
@@ -0,0 +1,97 @@
...
+ * balance. Iterates the borrowable reserves (the only ones a position can owe on) and reads on-chain
+ * `getUserTotalDebt(spoke, reserveId, proxy)` + `balanceOf(wallet)` per reserve. Returns an empty list
+ * when there's no position (nothing borrowed yet) — the run.ts pre-flight turns that into a clear "no
+ * debt to repay" error.
+ */
+export async function fetchRepayableDebts(
+  network: NetworkName,
+  ethAddress: string,
+): Promise<RepayableDebt[]> {
+  const reserves = await fetchBorrowableReserves(network);
```
jrwbabylonlab: `fetchBorrowableReserves` filters out paused/frozen reserves, but the app's Repay form deliberately uses `allBorrowReserves` so existing debt on frozen reserves can still be cleared — the pre-flight cannot see those positions.

services/vault/e2e/real/borrowParams.ts
```diff
@@ -49,18 +50,11 @@ export const CONSERVATIVE_BORROW_FRACTION = 0.25;
...
- * default can't drift above the max) to min(decimals, MAX_INPUT_DECIMALS) places, then trimmed.
+ * Format a token amount for the borrow form's numeric input — the shared floored formatter (see
+ * tokenAmount.ts). Kept as a named re-export so the borrow action's import is unchanged.
  */
-export function formatBorrowAmount(amount: number, decimals: number): string {
-  const places = Math.min(decimals, MAX_INPUT_DECIMALS);
-  const scale = 10 ** places;
-  return (Math.floor(amount * scale) / scale).toString();
-}
+export const formatBorrowAmount = formatTokenAmount;
```
jrwbabylonlab: Backward-compat alias — update borrow.ts to import `formatTokenAmount` directly and drop this.

gbarkhatov: done

services/vault/e2e/real/repayParams.ts
gbarkhatov: done

jrwbabylonlab (approved):

services/vault/e2e/real/repayParams.ts
gbarkhatov: Good catch — fixed. `fetchRepayableDebts` now iterates the full non-vBTC reserve set via a new `fetchAllReserves` (mirroring the app's `allBorrowReserves` in `fetchConfig.ts`), instead of the borrowable-filtered `fetchBorrowableReserves`. So debt on a reserve that was borrowable at borrow-time but has since been paused/frozen is now discovered by the pre-flight. `fetchAaveReserveConfig` returns both `all` (no borrowable/paused/frozen filter) and the `borrowable` subset the borrow path still uses. Behavior-neutral on devnet (all reserves are borrowable there) and verified live (borrow + repay, the repay default was sized from the debt and fell on-chain).

services/vault/e2e/real/borrowParams.ts
gbarkhatov: Done — dropped the `formatBorrowAmount = formatTokenAmount` alias (and its now-unused `formatTokenAmount` import) from `borrowParams.ts`. `borrow.ts` now imports `formatTokenAmount` from `../tokenAmount` directly, and the three call sites in `resolveBorrowAmount` were updated. Verified live.

PR #2050, opened 2026-07-15
feat(vault): add v3 position summary overview (#2035)

### LINKED ISSUE #2035: [v3] Position summary row
**Area:** `services/vault`

**Design:** [Position summary](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10163-47147) — `Position Container` → `Stats` → three `Stat Cards`.

## Context

Reshape `OverviewSection` into three position-level stat blocks with progress meters and inline actions. `useDashboardState` currently has no aggregate Available to Borrow value; the existing maximum-borrow calculation lives per reserve inside the Borrow form.

## Scope

- [ ] Total Collateral Value with a Deposit action
- [ ] Available to Borrow with a Borrow action
- [ ] Total Borrowed with a Repay action
- [ ] Add a hook or selector for the position-level Available to Borrow aggregate
- [ ] Render progress meters and position-aware enabled/disabled action states
- [ ] Keep the implementation behind `ENABLE_V3_UI`

## Open questions

- ⚠️ Confirm whether Available to Borrow is `collateralValueUsd × collateralFactor − debtValueUsd` or an existing SDK-derived value.
- ⚠️ Define each meter denominator: protocol cap utilization, remaining position headroom, or another percentage from the design.

These decisions are required before implementation; do not duplicate the per-reserve Borrow form calculation without confirming the position-level contract.

## Acceptance criteria

- [ ] Displayed collateral and borrowed values match `useDashboardState`
- [ ] Available to Borrow matches the confirmed aggregate calculation across all reserves
- [ ] Deposit, Borrow, and Repay navigate to the corresponding flow
- [ ] Meter values and action states cover no-position, available, and unavailable-action cases
- [ ] The row matches Figma in light and dark

## Depends on

- #2017

jeremy-babylonlabs:
Replace OverviewSection's v2 gauge+rows with a single "Position" summary card.

The v3 Overview now shows only this summary plus safety notifications; the legacy supply-cap, pending deposit/withdrawal, collateral, and loans sections are hidden.

Available-to-borrow capacity is derived via a shared `calculateBorrowCapacityUsd` helper so the dashboard and the Borrow form can't drift.

Closes #2035

## Screenshot
image

### DIFF
--- a/services/vault/src/components/simple/OverviewSection.tsx
+++ b/services/vault/src/components/simple/OverviewSection.tsx
@@ -36,8 +36,9 @@ interface OverviewSectionProps {
   availableToBorrow: string;
   collateralBtc: string;
   availableMeterPercent: number;
-  availableLoading: boolean;
   borrowedMeterPercent: number;
+  borrowCapacityLoading: boolean;
+  borrowCapacityError: Error | null;
   onDeposit: () => void;
   onBorrow: () => void;
   onRepay: () => void;
@@ -84,8 +85,9 @@ export function OverviewSection({
   availableToBorrow,
   collateralBtc,
   availableMeterPercent,
-  availableLoading,
   borrowedMeterPercent,
+  borrowCapacityLoading,
+  borrowCapacityError,
   onDeposit,
   onBorrow,
   onRepay,
@@ -116,6 +118,8 @@ export function OverviewSection({
   const statCards: PositionStatCard[] = useMemo(() => {
     const clampPct = (ratio: number) =>
       Math.round(Math.min(1, Math.max(0, ratio)) * 100);
+    const capacityUnavailable =
+      borrowCapacityLoading || borrowCapacityError != null;
     return [
       {
         label: COPY.overview.totalCollateralValueLabel,
@@ -128,8 +132,12 @@ export function OverviewSection({
       },
       {
         label: COPY.overview.availableToBorrowLabel,
-        value: availableLoading ? COPY.common.loading : availableToBorrow,
-        meter: availableLoading
+        value: borrowCapacityLoading
+          ? COPY.common.loading
+          : borrowCapacityError
+            ? COPY.common.emptyValue
+            : availableToBorrow,
+        meter: capacityUnavailable
           ? undefined
           : {
               percent: availableMeterPercent,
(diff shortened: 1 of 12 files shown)


jrwbabylonlab (dismissed):

services/vault/src/hooks/useDashboardState.ts
```diff
@@ -43,6 +46,17 @@ export function useDashboardState(connectedAddress: string | undefined) {
     debtValueUsd,
   });
 
+  const { params: splitParams, isLoading: isBorrowCapacityLoading } =
+    useVaultSplitParams(connectedAddress);
```
jrwbabylonlab: `error` is dropped here — if the split-params reads fail, the card shows $0 available with Borrow disabled forever, indistinguishable from a maxed-out position. Forward it like `useAaveReserveDetail` does with `ancillaryError`.

services/vault/src/components/simple/OverviewSection.tsx
```diff
@@ -91,6 +113,63 @@ export function OverviewSection({
...
+            },
+        actionLabel: COPY.overview.borrowAction,
+        onAction: onBorrow,
+        actionDisabled: !canBorrow,
+      },
+      {
+        label: COPY.overview.totalBorrowedLabel,
+        value: totalBorrowed,
+        meter: {
+          percent: borrowedMeterPercent,
```
jrwbabylonlab: Not gated on `availableLoading` like the available card, so a user with real debt sees "0% borrowed" while split params load (and forever if they error).

services/vault/src/components/simple/PositionStatCards.tsx
```diff
@@ -0,0 +1,111 @@
...
+      >
+        {card.actionLabel}
+      </button>
+    </div>
+  );
+}
+
+export function PositionStatCards({ cards }: { cards: PositionStatCard[] }) {
+  return (
+    <div className="rounded-lg bg-secondary-highlight p-6 dark:bg-[#202020]">
```
jrwbabylonlab: `dark:bg-[#202020]` — use a semantic token; this would be the only raw-hex dark background in the app.

jonybur (dismissed):
Approving. The v3 Position summary card is cleanly gated behind `isV3UiEnabled` and the extracted `calculateBorrowCapacityUsd` helper matches the prior inline formula exactly - refactor is behavior-preserving. The v2 rendering path is left untouched, and the disconnected screen still short-circuits before the v3 summary shows. CI green, no substantive findings.

jrwbabylonlab (approved):

PR #2053, opened 2026-07-15
feat(vault): add depositor funnel failure tracking (Phase 2) to Sentry

jonybur:
## Context

Phase 2 of the Sentry funnel tracking. Phase 1 (#2052) added the success spine (`deposit.started → registered → broadcast.succeeded → activation.activated`). This PR completes the **failure side** of the funnel.

> Stacked on `feat/vault-sentry-funnel-tracking` (#2052) - it reuses the `telemetryEvents.ts` helpers. Rebase onto `main` once #2052 merges.

## The gap

Several deposit failure paths were **completely invisible to Sentry** - they only wrote UI state (`setError` / `setLocalError`), with no capture. The worst are on the resume path, including a **critical-path #3** payout-signing failure and a WOTS derivation-drift mismatch.

## What it adds

New `TELEMETRY_STAGE` tags (`activation.wots` / `payouts` / `secret` / `reveal`). A failure is a `captureException`, not a named message, so it's tagged `funnelStage` and carries a **scrubbed `vaultId`** - the same per-vault key the Phase 1 milestones use, so failures join to the funnel per vault.

Closed the three Sentry-invisible resume-path catches:

| Capture | File | Was |
| --- | --- | --- |
| `activation.payouts` (critical path #3) | `usePayoutSigningState.ts` | `setError` only |
| `activation.wots` (+ `wotsMismatch` flag) | `ResumeDepositContent.tsx` (`ResumeWotsContent`) | `setError` only |
| `activation.secret` | `ResumeDepositContent.tsx` (`ResumeActivationContent`) | `setLocalError` only |

Plus enriched the existing reveal-failure capture (`useActivationState.ts`) with `vaultId` + the `activation.reveal` stage instead of a vague context string.

**Leak-safe:** these error messages carry only tx hashes (matched and redacted by the hex regex) and derivation errors - never secret bytes. The payout `AbortError` (user cancel / unmount) is excluded so cancellations aren't logged as failures.

## Deliberately deferred to Phase 3

The intra-activation **success** milestones (`activation.wots.submitted`, `activation.payouts.signed`) are **not** here: `submitWotsPublicKey` / `signAndSubmitPayouts` are idempotent and return `void` ("If the VP has already moved past that status, submission is skipped"), so emitting on resolution would double-count on resume/retry. They need transition-based emission at the localStorage status advance, and land in Phase 3 alongside the polling terminals (`deposit.completed`, `activation.verified`, daemon/pegout terminals), which share the same transition-diff machinery.

## Testing

- New tests: a payout signing failure is captured with the `activation.payouts` stage + scrubbed `vaultId`; a user-cancelled (`AbortError`) attempt is not captured.
- Full vault suite green: **211 files, 2228 tests**. eslint clean; `tsc -p tsconfig.lib.json` clean.

### DIFF
--- a/services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
+++ b/services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
@@ -74,6 +74,11 @@ vi.mock("../../../../utils/errors/formatting", () => ({
   }),
 }));
 
+const mockLoggerError = vi.hoisted(() => vi.fn());
+vi.mock("@/infrastructure", () => ({
+  logger: { error: mockLoggerError },
+}));
+
 const ACTIVITY = {
   id: "0xvault" as Hex,
   peginTxHash: "0xpegin" as Hex,
@@ -183,6 +188,42 @@ describe("usePayoutSigningState", () => {
         completed: 9,
         total: 9,
       });
+    });
+  });
+
+  describe("failure telemetry", () => {
+    it("captures a signing failure to Sentry with the activation.payouts stage and a scrubbed vaultId", async () => {
+      mockSignAndSubmitPayouts.mockRejectedValueOnce(
+        new Error("VP rejected the depositor graph"),
+      );
+
+      const { result } = renderHookWithProps();
+
+      await act(async () => {
+        await result.current.handleSign();
+      });
+
+      expect(mockLoggerError).toHaveBeenCalledTimes(1);
+      const [err, ctx] = mockLoggerError.mock.calls[0];
+      expect(err).toBeInstanceOf(Error);
+      expect(ctx.tags.funnelStage).toBe("activation.payouts");
+      expect(ctx.data.vaultId).toBe("0xvault");
+      expect(result.current.error?.title).toBe("Sign Error");
+    });
+
+    it("does not capture a user-cancelled (AbortError) signing attempt", async () => {
+      const abort = new Error("aborted");
+      abort.name = "AbortError";
+      mockSignAndSubmitPayouts.mockRejectedValueOnce(abort);
+
+      const { result } = renderHookWithProps();
+
+      await act(async () => {
+        await result.current.handleSign();
+      });
+
+      expect(mockLoggerError).not.toHaveBeenCalled();
+      expect(result.current.error).toBeNull();
     });
   });
 
--- a/services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
+++ b/services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
@@ -13,6 +13,8 @@ import { useCallback, useEffect, useRef, useState } from "react";
 import type { Hex } from "viem";
 
 import { COPY } from "@/copy";
+import { logger } from "@/infrastructure";
+import { shortId, TELEMETRY_STAGE } from "@/infrastructure/telemetryEvents";
 import type { PayoutSigningProgress } from "@/services/vault/vaultPayoutSignatureService";
 
 import { usePeginPolling } from "../../../context/deposit/PeginPollingContext";
@@ -22,8 +24,8 @@ import { LocalStorageStatus } from "../../../models/peginStateMachine";
 import { fetchVaultPayoutScriptPubKey } from "../../../services/vault/fetchVaults";
 import type { VaultActivity } from "../../../types/activity";
 import {
-  BtcWalletLivenessError,
   btcAddressToScriptPubKeyHex,
+  BtcWalletLivenessError,
   shouldProbeWalletLiveness,
   verifyBtcWalletLiveness,
 } from "../../../utils/btc";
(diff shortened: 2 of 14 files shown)


jeremy-babylonlabs (approved):

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -304,6 +306,15 @@ export function usePayoutSigningState({
           setSigning(false);
           return;
         }
+        // Critical-path #3 presign failure on the resume path — previously
+        // only surfaced to UI state, invisible to Sentry.
+        logger.error(err instanceof Error ? err : new Error(String(err)), {
```
gbarkhatov, line 311: Nit (non-blocking): user-cancellations get logged as funnel *failures*. This catch (and the wots/secret/reveal ones) logs every caught error. On the resume paths `deriveVaultRoot` / payout signing / the reveal tx open a wallet popup; a user clicking **Reject** throws a user-rejection error that lands here as an `activation.*` funnel failure, inflating the exact metric this PR produces. The site already early-returns on `AbortError`, but user-rejection is the more common drop-off and isn't filtered at any of the four sites. There's a ready-made helper: `classifyError(err) === "user-rejection"` in `utils/errors/formatting.ts` (covers EIP-1193 4001 / viem `UserRejectedRequestError` / wallet `CONNECTION_REJECTED`) — guard each capture with it. Happy to defer to Phase 3.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -421,6 +422,17 @@ export function ResumeWotsContent({
...
     } catch (err) {
+      // Capture regardless of mount — these resume flows have no abort signal,
+      // so a real WOTS-submission / derivation-drift failure is worth knowing
+      // even if the user has already closed the modal. Only the UI update below
+      // is mount-gated. A mismatch is flagged for faceting.
+      logger.error(err instanceof Error ? err : new Error(String(err)), {
+        tags: { funnelStage: TELEMETRY_STAGE.ACTIVATION_WOTS },
+        data: {
+          vaultId: shortId(activity.id),
+          wotsMismatch: isWotsMismatchError(err),
```
gbarkhatov, line 433: Nit (non-blocking): the Pre-PegIn tx-hash trust-boundary mismatch (a potential-attack signal, thrown a few lines up as "Aborting to prevent potential attack") lands here under `activation.wots` with `wotsMismatch: false`, so it's bucketed with ordinary WOTS submission failures. Fine if you only ever facet on `funnelStage`; if you'd want the attack signal isolated later it won't be distinguishable from derivation drift.

services/vault/src/infrastructure/telemetryEvents.ts
```diff
@@ -24,6 +24,23 @@ export const TELEMETRY_EVENT = {
...
+ * the success milestones.
+ */
+export const TELEMETRY_STAGE = {
+  ACTIVATION_WOTS: "activation.wots",
+  ACTIVATION_PAYOUTS: "activation.payouts",
+  ACTIVATION_SECRET: "activation.secret",
+  ACTIVATION_REVEAL: "activation.reveal",
+} as const;
+
+export type TelemetryStage =
```
gbarkhatov, line 41: Nit (non-blocking): `TelemetryStage` is exported but has zero consumers — `logger.error` types `tags` as `Record<string, string>` and every call site passes the `TELEMETRY_STAGE.*` value directly. The repo ZERO DEAD CODE POLICY flags unused types. In fairness it mirrors the pre-existing, also-unused `TelemetryEvent` type, so this is a consistency nit rather than a regression — either wire it into the tags type or drop it.

services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
```diff
@@ -186,6 +191,42 @@ describe("usePayoutSigningState", () => {
...
+      const [err, ctx] = mockLoggerError.mock.calls[0];
+      expect(err).toBeInstanceOf(Error);
+      expect(ctx.tags.funnelStage).toBe("activation.payouts");
+      expect(ctx.data.vaultId).toBe("0xvault");
+      expect(result.current.error?.title).toBe("Sign Error");
+    });
+
+    it("does not capture a user-cancelled (AbortError) signing attempt", async () => {
+      const abort = new Error("aborted");
+      abort.name = "AbortError";
```
gbarkhatov, line 216: Nit (non-blocking): this test can't catch a regression in the abort guard. The real cancellation in prod is a `DOMException("...", "AbortError")` (what `AbortController.signal.reason` / `abortableSleep` produce), but the guard is `err instanceof Error && err.name === "AbortError"` and `DOMException instanceof Error` is false under jsdom — while this test throws a hand-rolled plain `Error` with `.name` set, so it never exercises the actual abort type. Modern browsers have `DOMException extends Error` so prod likely works, and this matches the existing `useRefundState.ts` pattern — but either construct a real `new DOMException(...)` here, or drop the `instanceof` gate in favor of `err?.name === "AbortError"` (as `graphql/client.ts` already does).

gbarkhatov (approved):
Approving — Phase 2 cleanly closes the four Sentry-invisible failure catches, scrubbing is centralized so there's no leak risk, `vaultProviderAddress` is in scope, and there's no double-capture. All comments below are **minor nits**: none block merge. Fine to ship as-is, or fold them into a follow-up PR.

One extra nit not worth an inline thread: only the payout capture attaches `providerId`; the wots/secret/reveal captures don't. Scope-limited and already acknowledged in the PR body — just flagging the asymmetry.

PR #2060, opened 2026-07-16
fix(vault): derive the mempool API path from the BTC network

jrwbabylonlab:
Two small, unrelated vault fixes (kept as separate commits).

### 1. Mempool API URL derived from the BTC network

`getNetworkConfigBTC` appended `/signet` for signet on top of
`NEXT_PUBLIC_MEMPOOL_API`, which only worked when that value was exactly
`https://mempool.space`. A base already containing `/signet` double-prefixed to
`.../signet/signet/api`; a dedicated signet host got a bogus `/signet` inserted —
both 404 / CORS-fail.

`NEXT_PUBLIC_MEMPOOL_API` is now the full mempool base for the network, WITHOUT a
trailing `/api` (the reader appends it). `resolveMempoolApiUrl`:
- **mempool.space** — path derived from `NEXT_PUBLIC_BTC_NETWORK` (signet →
  `/signet`, mainnet → root), so a signet app can't read mainnet data through it.
- **custom / self-hosted host** — used verbatim.
- **unset** — network-correct mempool.space default.

| NEXT_PUBLIC_MEMPOOL_API | network | result |
|---|---|---|
| `https://mempool.space` | signet | `https://mempool.space/signet/api` |
| `https://mempool.space` | mainnet | `https://mempool.space/api` |
| custom signet host | signet | `<host>/api` |
| unset | signet | `https://mempool.space/signet/api` |

Regression origin: #1904 introduced the hardcoded `/signet` append.

### 2. Remove the placeholder docs link from the protocol-status banner

`ProtocolStatusBanner` (frozen/paused, incl. freeform DevOps messages) always
appended a "Learn more" link to a placeholder `docs.babylonlabs.io` URL that
product never replaced. Removed the link and the now-dead copy
(`protocolStatus.*.learnMore`, `PROTOCOL_STATUS_LEARN_MORE`) per zero-dead-code.

### DIFF
--- a/services/vault/src/config/network/runtime.ts
+++ b/services/vault/src/config/network/runtime.ts
@@ -39,7 +39,13 @@ export interface BabylonConfigOptions {
   /** Bitcoin network. Must be "mainnet" or "signet". */
   btcNetwork: BtcNetworkName;
 
-  /** Optional mempool API base URL. Defaults to `https://mempool.space`. */
+  /**
+   * Optional mempool API base URL, WITHOUT a trailing `/api` (the reader
+   * appends `/api`). For `mempool.space` the network path is derived from
+   * `btcNetwork`; a custom / self-hosted host is used verbatim. When omitted,
+   * defaults to the network-correct mempool.space URL.
+   * See {@link resolveMempoolApiUrl}.
+   */
   mempoolApiUrl?: string;
 }
 
@@ -50,7 +56,39 @@ export interface BabylonConfigState {
   mempoolApiUrl: string;
 }
 
-const DEFAULT_MEMPOOL_API_URL = "https://mempool.space";
+// Public mempool.space origin used as the default and as the one host whose
+// network path we derive from `btcNetwork` (see resolveMempoolApiUrl).
+const MEMPOOL_SPACE_ORIGIN = "https://mempool.space";
+
+/**
+ * Resolve the mempool API base URL (WITHOUT a trailing `/api`) for the declared
+ * BTC network.
+ *
+ * mempool.space encodes the network in the URL path (mainnet at the root, signet
+ * under `/signet`), so for that known public host we derive the path from
+ * `btcNetwork` — a signet app can never accidentally read mainnet data through
+ * it. Any other (custom / self-hosted) mempool base is network-specific by
+ * deployment (e.g. a dedicated signet host serving `/api` at its root) and is
+ * used verbatim.
+ */
+export function resolveMempoolApiUrl(
+  base: string | undefined,
+  network: BtcNetworkName,
+): string {
+  const trimmed = (base ?? MEMPOOL_SPACE_ORIGIN).replace(/\/+$/, "");
+  let host: string;
+  try {
+    host = new URL(trimmed).host;
+  } catch {
+    throw new Error(`Invalid NEXT_PUBLIC_MEMPOOL_API URL: "${base}"`);
+  }
+  if (host === "mempool.space") {
+    return network === BTC_SIGNET
+      ? `${MEMPOOL_SPACE_ORIGIN}/signet`
+      : MEMPOOL_SPACE_ORIGIN;
+  }
+  return trimmed;
+}
 
 let state: BabylonConfigState | null = null;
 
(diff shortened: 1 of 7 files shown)


jeremy-babylonlabs (approved):

jonybur (approved):
Both fixes look correct.

**Mempool URL derivation:** `resolveMempoolApiUrl` is called at `configureBabylonConfig` time and stores the fully-resolved base (`https://mempool.space/signet` for signet) into `BabylonConfigState.mempoolApiUrl`; `getNetworkConfigBTC` now passes that through instead of appending `/signet` again - no more double-prefix on `.../signet`-suffixed bases or bogus insertion on dedicated signet hosts. The `endsWith("/signet")` idempotency guard closes the double-prefix path, trailing-slash trimming is in place, and `getMempoolApiUrl` still tacks on `/api` cleanly so nothing else has to change. Test coverage (signet append, mainnet no-op, custom host, idempotency, undefined default, trailing slash) exercises the meaningful shapes.

**ProtocolStatusBanner cleanup:** the placeholder link, its constant, and the two `learnMore` copy fields are removed together with no dangling references - the `learnMore` hits in `UtxoSplitSelector` and `PendingWithdrawCard` are unrelated (deposit form / withdrawal card), so this is a clean zero-dead-code deletion. Tests correctly flip to `queryByRole("link")` being absent for both the default frozen body and the freeform-override branch.

services/vault/src/config/network/runtime.ts
```diff
@@ -50,7 +55,27 @@ export interface BabylonConfigState {
...
+ * that already ends in `/signet` is left as-is, so it can never double to
+ * `.../signet/signet`.
+ */
+export function resolveMempoolApiUrl(
+  base: string | undefined,
+  network: BtcNetworkName,
+): string {
+  const trimmed = (base ?? DEFAULT_MEMPOOL_HOST).replace(/\/+$/, "");
+  if (network !== BTC_SIGNET) return trimmed;
+  return trimmed.endsWith("/signet") ? trimmed : `${trimmed}/signet`;
```
gbarkhatov, line 77: Shipped behavior is correct, but the PR description contradicts it.

Verified against the real proxy: `https://mempool.signet.babylonlabs.io/signet/api/...` → 200, bare root `/api` → 403. So appending `/signet` to every signet host is right — no regression here.

The PR description, though, still describes the earlier commit-1 model — it says a "custom / self-hosted host — used verbatim" and the table row maps `custom signet host → <host>/api`, whereas the merged code produces `<host>/signet/api`. Worth updating the body (and dropping the "Regression origin #1904 → verbatim" framing) so the description matches what shipped.

Residual caveat on the assumption baked in here: every signet mempool host must serve signet under `/signet`. That holds for mempool.space and the babylon proxy; a hypothetical signet-only mirror serving its API at the root would get a spurious `/signet`. `.env.example` documents the expectation, so this is acceptable — just flagging the assumption.

services/vault/src/config/network/runtime.ts
```diff
@@ -50,7 +55,27 @@ export interface BabylonConfigState {
...
+ * Resolve the mempool API base URL (WITHOUT a trailing `/api`) for the declared
+ * BTC network.
+ *
+ * Every mempool host we use — public mempool.space and the babylon proxy alike —
+ * serves signet under a `/signet` path (mainnet is at the root), so `/signet` is
+ * appended for signet regardless of host. It is appended idempotently: a base
+ * that already ends in `/signet` is left as-is, so it can never double to
+ * `.../signet/signet`.
+ */
+export function resolveMempoolApiUrl(
```
gbarkhatov, line 71: Production semantics were rewritten in a commit labeled `test`.

This function was rewritten in `6b70e12d` ("test(vault): cover the mempool.space https normalization") — despite the `test(...)` prefix it changed production behavior: removed the `new URL()` validation + `throw` on a malformed `NEXT_PUBLIC_MEMPOOL_API`, dropped the `host === "mempool.space"` special-case, and switched to the unconditional idempotent `/signet` append. A behavior reversal under a test-labeled commit is easy to miss in review — worth calling out for next time.

The removed `new URL()` validation is fine to drop in practice: `parseOptionalUrl(process.env.NEXT_PUBLIC_MEMPOOL_API)` in `services/vault/src/config/env.ts:215` already validates and nulls a malformed value before the only production caller (`configureBabylonConfig`) reaches this function. Noting it only because the validation existed one commit earlier.

gbarkhatov (commented):
Post-merge review notes on the mempool-URL change.

I verified the shipped behavior against the real babylon signet proxy: `https://mempool.signet.babylonlabs.io/signet/api/blocks/tip/height` returns 200, while the bare root `/api` returns 403 (the root serves the frontend HTML). So the merged code — append `/signet` to every signet host — is **correct** for the actual deployment; the earlier worry that it breaks a dedicated signet host does not apply to this proxy.

Two smaller, non-blocking notes inline: the PR description no longer matches what merged, and the production rewrite landed under a `test(...)` commit.

PR #1, opened 2025-02-02
Init monorepo with nx and include proto-ts

supertong:
I think it's easier to talk about it with some code, so here it is.

A monorepo isn't simply a repo for hosting all the codes together, it usually comes with a set of toolings for
* release management
* module resolution
* dependency management
* task ochestration
* ...

In our cases, I think it makes sense to start with an open source solution instead of in-housing the tooling. After some researching, I ended up with nx. Main reason being
* it's mature
* good ecosystem
* good extensibility
* good enough for our use case

Of course, the down side being some learning curve to understand the tooling but we should be able to hide the details with `npm run xxx` command so that most developers doesn't need to worry about it.

In this PR, I have initialized the repo and included the proto-ts lib but here is a few things to note:
* **External Dependency Management**: Nx is pretty flexible on the dependency management. It can be all hoisted to the root package.json to make sure consistency of all dependencies or it's also fine for each package to specify its own. I didn't hoist the dependency for proto-ts in this PR since I don't want to make the call here. My recommendation: consistent dependency version.
* **Release Management**: Nx has built-in support for [Conventional Commits](https://nx.dev/recipes/nx-release/automatically-version-%20ith-conventional-commits) which I know @totraev likes. We don't have to build tooling around it to make it work for monorepo which is a plus. That's why I didn't include changeset in this PR. Most things should already be ready to use if we opt in for Convertional Commits except we might need something to make sure people using the right commit types in the message.
* **Task execution**: Nx support specifying task input to take advantage of caching. I didn't do that in this PR and prefer to leave that as a optimization later on.
* More comments in the code.

PR #2069, opened 2026-07-16
feat(vault): add on-chain funnel terminals (Phase 3) to Sentry

jonybur:
## Context

Phase 3 of the Sentry funnel tracking. Phase 1 (#2052) added the success spine, Phase 2 (#2053) the failure side. This PR adds the two **on-chain terminal milestones** that close the deposit funnel.

> **Rebased onto `main` now that Phase 2 has merged.** The diff was previously showing Phase 2's files as if they were new — the branch was stacked and its merge-base was still Phase 1. It is now Phase-3-only: 4 files.

## What it adds

Emitted from `PeginPollingContext` as each vault's `contractStatus` transitions, per-vault with a scalar `vaultId`:

| Event | When |
| --- | --- |
| `activation.verified` | `contractStatus` VERIFIED - VP verified the presigned payouts |
| `deposit.completed` | `contractStatus` ACTIVE - **the true conversion terminal** (promised since Phase 1) |

This completes the primary funnel (`started → registered → broadcast.succeeded → activated → completed`) and lets `activation.verified` split the value-at-risk stall alert into VP-side (verified absent) vs depositor-side.

## The transition-diff problem, and how it's handled

The polling loop re-runs every cycle, so a naive emit-on-observe would (a) fire repeatedly for a vault sitting in a terminal state and (b) emit a **burst** on any dashboard load full of already-completed deposits. Two rules make emission per-vault and once-per-transition:

- **Seed on first observation.** The first time a vault is seen, if it's already at/past a milestone (a prior-session completion or a resumed deposit), it's marked emitted *without* emitting. So a page load never bursts.
- **Emit once thereafter**, the first cycle a *seen* vault is observed to have reached that state, then never again.

The detection is a **pure function** (`collectTerminalMilestones` in `terminalMilestones.ts`) - so the transition logic is unit-tested directly, not through the React tree.

## Review fixes since the last round

Thanks @gbarkhatov and @jrwbabylonlab — all three findings against Phase 3's own code were real. Each fix has a test that was verified to fail without it.

**1. Double-emit through co-mounted providers** (@gbarkhatov). Confirmed, and worse than reported. Tracking was a per-provider `useRef`, but several providers observe the same vault at once, so each emitted its own copy — the headline conversion metric counted twice on the normal completion path. Reproduced by mounting two real providers and counting `logger.event` calls: `["activation.verified", "activation.verified", "deposit.completed", "deposit.completed"]`.

There are three co-mount paths, and the third rules out the cheaper fix:

- `PendingDepositSection` → `PostDepositContinuationContent` mounts a second provider (nested).
- `PendingDepositSection` → `PendingDepositModals` → `SimpleDeposit` → `DepositSignContent` (nested).
- `RootLayout` renders `<Outlet/>` (→ dashboard) and `<SimpleDeposit>` as **siblings** — neither is an ancestor of the other, so "only emit in the outermost provider" would leave both self-identifying as outermost and both still emitting.

So tracking is now an **app-scoped module store keyed by vaultId**, which holds however many providers are mounted. It stays in-memory deliberately: a reload must reset it, so the seeding rule re-suppresses prior-session completions rather than replaying them. Regression test covers the nested path, the sibling path, and seeding.

**2. Seeding from localStorage-only rows** (@jrwbabylonlab). Confirmed — this was the sharpest catch, and it defeated the very burst-suppression the seeding rule exists for. `contractStatus` is never actually `undefined` in the real pipeline: the `usePeginStorage` merge fabricates a synthetic `PENDING` with `isPending: true` for rows the indexer hasn't returned. On a cold load with a stale pending row, the tracker seeded from that fabricated PENDING, then emitted both terminals one poll later for a transition that happened entirely while the user was away.

Milestones are now decided from an indexer-sourced status only (`isPending` rows are skipped). The `contractStatus === undefined` half of the guard stays as type-level narrowing for the optional field — with its comment corrected, since it was advertising itself as a load-in-progress guard it never was. Both halves have their own test; the `isPending` half has two.

**3. LIQUIDATED / DEPOSITOR_WITHDRAWN** (@gbarkhatov). Included. The right model is "terminal statuses that imply the vault reached ACTIVE" — a liquidated vault *did* convert, it activated first — which is exactly the contract-status set of the existing `isVaultPastActivation`, so `hasReachedActive` mirrors it rather than inventing a set. EXPIRED and INVALID are pre-activation aborts and still count for nothing; there's a test pinning that too.

**Not taken:**

- *Lazy `useRef` init* (@gbarkhatov, optional): moot — fix #1 removes the ref entirely.
- *Documenting slow registration as a seeding trigger* (@gbarkhatov, non-blocking): the header comment already names the two real triggers ("a prior-session completion, or a resumed deposit"). A vault can't realistically first appear already VERIFIED — the depositor must broadcast BTC and wait for confirmations first, which is what makes registration→activities latency a non-issue here. Adding it would make an accurate comment wrong.
- *Phase 2's failure-capture issues* (@gbarkhatov #2–#5): real, but they're against code that merged in #2053 and is no longer in this diff. Split into a follow-up as you suggested — see the linked PR.

**Known limitation** (unchanged): tracking is session-scoped, so a transition that happened entirely while the user was away is treated as pre-existing and not emitted. Acceptable for a first cut — it avoids localStorage persistence and the historical-burst problem.

## Deliberately deferred to Phase 4

The remaining transition-sensitive events: `activation.wots.submitted` / `payouts.signed` (localStorage status advance), `activation.daemon.terminal` (VP daemon terminal statuses), and the exit/redemption pegout terminals.

## Testing

- `terminalMilestones.test.ts` (11): the verified→completed sequence, no-burst seeding, fast PENDING→ACTIVE and →REDEEMED jumps, →LIQUIDATED, EXPIRED emitting nothing, the missing-status guard, the two localStorage-only seeding cases, in-session emission surviving the guard, and per-vault independence in a batch.
- `terminalMilestoneEmission.test.tsx` (3): once-per-vault emission across nested and sibling providers, and no emission for an already-ACTIVE vault.
- Each fix mutation-checked: reverting it fails exactly its own tests and nothing else.
- Full vault suite green: **228 files, 2406 tests**. eslint clean; `tsc -p tsconfig.lib.json` clean.

### DIFF
--- a/services/vault/src/context/deposit/PeginPollingContext.tsx
+++ b/services/vault/src/context/deposit/PeginPollingContext.tsx
@@ -21,6 +21,8 @@ import {
 } from "react";
 
 import { useDemoDeposit } from "@/dev/demoDeposit";
+import { logger } from "@/infrastructure";
+import { shortId } from "@/infrastructure/telemetryEvents";
 
 import { usePeginPollingQuery } from "../../hooks/deposit/usePeginPollingQuery";
 import { useRequiredPrePeginDepthResolver } from "../../hooks/deposit/useRequiredPrePeginDepth";
@@ -55,6 +57,10 @@ import { isVaultOwnedByWallet } from "../../utils/vaultWarnings";
 import { useProtocolParamsContext } from "../ProtocolParamsContext";
 
 import { computeDepositPollingResult } from "./computeDepositPollingResult";
+import {
+  collectTerminalMilestones,
+  getSharedTerminalMilestoneTracking,
+} from "./terminalMilestones";
 
 /** React Query namespace for the Pre-PegIn confirmation poller. */
 const PREPEGIN_CONFIRMATIONS_QUERY_KEY = "prePeginMempoolConfirmations";
@@ -338,6 +344,28 @@ export function PeginPollingProvider({
       return next;
     });
   }, [htlcRefundByDepositId, refundedHtlcVaultIds]);
+
+  // Emit the on-chain funnel terminals — activation.verified and
+  // deposit.completed — once per vault as its contractStatus transitions. The
+  // detector seeds already-terminal vaults on first observation, so a dashboard
+  // load never emits a burst for prior-session completions. Tracking is
+  // app-scoped rather than per-provider: the continuation modal mounts a second
+  // provider over the same vault, and per-instance tracking would count every
+  // terminal twice. It is a plain module store, not state — emitting telemetry
+  // must not trigger a re-render.
+  useEffect(() => {
+    const milestones = collectTerminalMilestones(
+      activities,
+      getSharedTerminalMilestoneTracking(),
+    );
+    for (const milestone of milestones) {
+      logger.event(milestone.event, {
+        level: "info",
+        category: milestone.category,
+        vaultId: shortId(milestone.vaultId),
+      });
+    }
+  }, [activities]);
 
   // Optimistic status handlers
   const setOptimisticStatus = useCallback(
--- a/services/vault/src/context/deposit/terminalMilestones.ts
+++ b/services/vault/src/context/deposit/terminalMilestones.ts
@@ -1,0 +1,153 @@
+/**
+ * Pure transition detector for the on-chain deposit-funnel terminals emitted
+ * from the polling context: `activation.verified` (VP verified the presigned
+ * payouts) and `deposit.completed` (vault ACTIVE — funnel terminal).
+ *
+ * Emission must be per-vault and once-per-transition. Two rules make that hold
+ * against a poll loop that re-runs every cycle:
+ *  - A vault is SEEDED on its first indexer-sourced observation: if it is
+ *    already at/past a milestone then (a prior-session completion, or a resumed
+ *    deposit), it is marked emitted WITHOUT emitting, so a page load never emits
+ *    a burst. localStorage-only rows are skipped outright — their status is
+ *    fabricated, not observed.
+ *  - Thereafter a milestone emits once, the first cycle a seen vault is observed
+ *    to have reached that state, and never again.
+ */
+
+import {
+  TELEMETRY_EVENT,
+  type TelemetryEvent,
+} from "../../infrastructure/telemetryEvents";
+import { ContractStatus } from "../../models/peginStateMachine";
+
+export interface TerminalMilestoneTracking {
+  /** Vaults observed at least once (drives first-observation seeding). */
+  seen: Set<string>;
+  emittedVerified: Set<string>;
+  emittedCompleted: Set<string>;
+}
+
+export interface TerminalMilestone {
+  event: TelemetryEvent;
+  category: "activation" | "deposit";
+  /** Raw vaultId; the caller shortens it before it enters event context. */
+  vaultId: string;
+}
+
+export function createTerminalMilestoneTracking(): TerminalMilestoneTracking {
+  return {
+    seen: new Set(),
+    emittedVerified: new Set(),
+    emittedCompleted: new Set(),
+  };
+}
+
+/**
+ * App-scoped tracking shared by every mounted `PeginPollingProvider`.
+ *
+ * Several providers observe the same vault at once: the dashboard section polls
+ * every activity, while the continuation modal mounts its own provider scoped to
+ * the viewed batch — nested inside that section, or as a sibling under the root
+ * layout. Per-provider tracking emits each terminal once per provider, so the
+ * once-per-vault guarantee is keyed to the vault id here rather than to a
+ * provider instance, and holds however many providers are mounted.
+ *
+ * In-memory by design: a reload must reset it, so that the seeding rule above
+ * re-suppresses prior-session completions rather than replaying them.
+ */
+const sharedTracking = createTerminalMilestoneTracking();
+
+export function getSharedTerminalMilestoneTracking(): TerminalMilestoneTracking {
+  return sharedTracking;
+}
+
+/**
+ * Test-only reset: the shared store outlives any single provider, so cases
+ * asserting emissions must start from a clean slate.
+ */
+export function resetSharedTerminalMilestoneTracking(): void {
+  sharedTracking.seen.clear();
+  sharedTracking.emittedVerified.clear();
+  sharedTracking.emittedCompleted.clear();
+}
+
+interface StatusBearingActivity {
+  id: string;
+  contractStatus?: number;
+  /** True on localStorage-only rows the indexer has not returned yet. */
+  isPending?: boolean;
+}
+
+/**
+ * Terminal statuses only reachable from ACTIVE: a vault must have activated to
+ * be redeemed, liquidated (collateral seized against a debt) or withdrawn. A
+ * liquidated vault still converted — it activated first — so it counts. EXPIRED
+ * and INVALID are pre-activation aborts and never do. Mirrors the contract
+ * statuses of `isVaultPastActivation`, minus its optimistic localStorage branch:
+ * an on-chain milestone must be decided from chain truth alone.
+ */
+function hasReachedActive(status: ContractStatus): boolean {
+  return (
+    status === ContractStatus.ACTIVE ||
+    status === ContractStatus.REDEEMED ||
+    status === ContractStatus.LIQUIDATED ||
+    status === ContractStatus.DEPOSITOR_WITHDRAWN
+  );
+}
+
+/**
+ * Return the milestones to emit for this poll, mutating `tracking` so each fires
+ * at most once per vault. Any status implying the vault reached ACTIVE counts as
+ * both "reached verified" and "reached completed", and VERIFIED counts as
+ * reached-verified — so a vault that jumps straight past VERIFIED between polls
+ * still yields both milestones once.
+ */
+export function collectTerminalMilestones(
+  activities: readonly StatusBearingActivity[],
+  tracking: TerminalMilestoneTracking,
+): TerminalMilestone[] {
+  const out: TerminalMilestone[] = [];
+
+  for (const activity of activities) {
+    // Classify a vault only from an indexer-sourced status. A localStorage-only
+    // row carries a fabricated PENDING (`isPending: true` — see the merge in
+    // usePeginStorage) and exists only until the indexer returns the vault.
+    // Seeding from it would mark an already-terminal vault as first seen at
+    // PENDING, then emit a spurious milestone one poll later for a transition
+    // that happened in a previous session. The `contractStatus` half is
+    // defensive: no caller reaches it today, but the field is optional and the
+    // cast below would otherwise read `undefined` as PENDING (0).
+    if (activity.isPending || activity.contractStatus === undefined) continue;
+    const status = activity.contractStatus as ContractStatus;
+    const reachedCompleted = hasReachedActive(status);
+    const reachedVerified =
+      status === ContractStatus.VERIFIED || reachedCompleted;
+
+    if (!tracking.seen.has(activity.id)) {
+      tracking.seen.add(activity.id);
+      // First observation: seed pre-existing terminal state without emitting.
+      if (reachedVerified) tracking.emittedVerified.add(activity.id);
+      if (reachedCompleted) tracking.emittedCompleted.add(activity.id);
+      continue;
+    }
+
+    if (reachedVerified && !tracking.emittedVerified.has(activity.id)) {
+      tracking.emittedVerified.add(activity.id);
+      out.push({
+        event: TELEMETRY_EVENT.ACTIVATION_VERIFIED,
+        category: "activation",
+        vaultId: activity.id,
+      });
+    }
+    if (reachedCompleted && !tracking.emittedCompleted.has(activity.id)) {
+      tracking.emittedCompleted.add(activity.id);
+      out.push({
+        event: TELEMETRY_EVENT.DEPOSIT_COMPLETED,
+        category: "deposit",
+        vaultId: activity.id,
+      });
+    }
+  }
+
+  return out;
+}
--- a/services/vault/src/infrastructure/telemetryEvents.ts
+++ b/services/vault/src/infrastructure/telemetryEvents.ts
@@ -19,6 +19,10 @@ export const TELEMETRY_EVENT = {
   DEPOSIT_BROADCAST_SUCCEEDED: "deposit.broadcast.succeeded",
   /** HTLC secret revealed on Ethereum; activation tx submitted (optimistic). */
   ACTIVATION_ACTIVATED: "activation.activated",
+  /** VP verified the depositor's presigned payouts on-chain (contractStatus VERIFIED). */
+  ACTIVATION_VERIFIED: "activation.verified",
+  /** Vault confirmed ACTIVE on-chain — funnel terminal / primary conversion. */
+  DEPOSIT_COMPLETED: "deposit.completed",
 } as const;
 
 export type TelemetryEvent =
--- a/services/vault/src/context/deposit/__tests__/terminalMilestoneEmission.test.tsx
+++ b/services/vault/src/context/deposit/__tests__/terminalMilestoneEmission.test.tsx
@@ -1,0 +1,154 @@
+import { render } from "@testing-library/react";
+import type { Hex } from "viem";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import {
+  ContractStatus,
+  PEGIN_DISPLAY_LABELS,
+} from "../../../models/peginStateMachine";
+import type { VaultActivity } from "../../../types/activity";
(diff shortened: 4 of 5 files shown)


jonybur:
@greptile-apps

services/vault/src/context/deposit/PeginPollingContext.tsx
```diff
@@ -332,6 +339,28 @@ export function PeginPollingProvider({
...
 
+  // Emit the on-chain funnel terminals — activation.verified and
+  // deposit.completed — once per vault as its contractStatus transitions. The
+  // detector seeds already-terminal vaults on first observation, so a dashboard
+  // load never emits a burst for prior-session completions. Tracking lives in a
+  // ref: emitting telemetry must not trigger a re-render.
+  const terminalMilestoneTrackingRef = useRef(
+    createTerminalMilestoneTracking(),
+  );
+  useEffect(() => {
```
gbarkhatov, line 356: **Double-emit of the funnel terminals through the continuation modal.** `PendingDepositSection` mounts this provider over `allActivities`, and inside it renders `PostDepositContinuationContent`, which mounts a *second, nested* `PeginPollingProvider` scoped to the viewed vault (`PostDepositContinuationContent.tsx:66`). The viewed vault is in both providers' `activities`, and each owns its own `terminalMilestoneTrackingRef` + this effect. Activation completes inside that modal, so on VERIFIED→ACTIVE both trackers (each seeded pre-terminal) emit — `deposit.completed`/`activation.verified` counted twice on the normal completion path. Seeding doesn't save it: both providers observed the vault before it went terminal. Consider hoisting the tracker to a single app-scoped store keyed by vaultId, or only running this effect in the outermost provider.

services/vault/src/context/deposit/PeginPollingContext.tsx
```diff
@@ -332,6 +339,28 @@ export function PeginPollingProvider({
     });
   }, [htlcRefundByDepositId, refundedHtlcVaultIds]);
 
+  // Emit the on-chain funnel terminals — activation.verified and
+  // deposit.completed — once per vault as its contractStatus transitions. The
+  // detector seeds already-terminal vaults on first observation, so a dashboard
+  // load never emits a burst for prior-session completions. Tracking lives in a
+  // ref: emitting telemetry must not trigger a re-render.
+  const terminalMilestoneTrackingRef = useRef(
```
gbarkhatov: Nit (optional): `useRef(createTerminalMilestoneTracking())` runs the factory on every render, allocating three `Set`s that are discarded after the first. Idiomatic lazy init:

```ts
const ref = useRef<TerminalMilestoneTracking | null>(null);
ref.current ??= createTerminalMilestoneTracking();
```

services/vault/src/hooks/deposit/useActivationState.ts
```diff
@@ -108,7 +109,8 @@ export function useActivationState({
         });
       } catch (err) {
         logger.error(err instanceof Error ? err : new Error(String(err)), {
-          data: { context: "Vault activation failed" },
+          tags: { funnelStage: TELEMETRY_STAGE.ACTIVATION_REVEAL },
```
gbarkhatov, line 115: **This capture is unreachable — reveal/activation failures reach Sentry nowhere.** The catch wraps only `await vaultHandleActivation(...)`, but `useVaultActions.handleActivation` wraps its whole body in try/catch and never rethrows (only `setActivationError`/`setActivating`, `useVaultActions.ts:489-501`). So the awaited promise resolves even when `activateVaultWithSecret` reverts, this catch never runs, and `TELEMETRY_STAGE.ACTIVATION_REVEAL` is dead code. The same failure also isn't seen by the `activation.secret` catch (it's swallowed before that outer catch). To actually capture on-chain reveal failures, emit from inside `useVaultActions.handleActivation`'s catch (or have it rethrow).

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -304,6 +306,15 @@ export function usePayoutSigningState({
           setSigning(false);
           return;
         }
+        // Critical-path #3 presign failure on the resume path — previously
+        // only surfaced to UI state, invisible to Sentry.
+        logger.error(err instanceof Error ? err : new Error(String(err)), {
```
gbarkhatov, line 311: **User wallet-rejections are logged as `activation.payouts` failures.** The guard above filters only `err.name === 'AbortError'` (modal close); a wallet decline is EIP-1193 `4001` / `UserRejectedRequestError`, not `AbortError`, so it falls through to this capture — and `formatPayoutSignatureError(err)` on the next line classifies that same error as a benign `user-rejection` for the UI. The codebase already has `classifyError` (`utils/errors/formatting.ts:127-133`); gate the capture on `classifyError(err) !== 'user-rejection'` so routine cancellations don't inflate the failure rate this tag alerts on.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -421,6 +422,17 @@ export function ResumeWotsContent({
         onSuccess();
       }
     } catch (err) {
+      // Capture regardless of mount — these resume flows have no abort signal,
+      // so a real WOTS-submission / derivation-drift failure is worth knowing
+      // even if the user has already closed the modal. Only the UI update below
+      // is mount-gated. A mismatch is flagged for faceting.
+      logger.error(err instanceof Error ? err : new Error(String(err)), {
```
gbarkhatov, line 429: Same user-rejection concern as the payout capture: this try calls `deriveVaultRoot` / `submitWotsPublicKey`, both wallet-popup operations. A user declining the popup throws a `user-rejection` (per `classifyError`, `formatting.ts:127-133`) and is captured here as an `activation.wots` failure with no filter. Real derivation-drift / hash-mismatch failures *should* be captured; gate on `classifyError(err) !== 'user-rejection'` to exclude cancellations.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -659,6 +671,13 @@ export function ResumeActivationContent({
       // error there, not a silent submission.
       await handleActivation(secretHex);
     } catch (err) {
+      // Capture regardless of mount (no abort signal on this flow). The error
+      // message carries only tx hashes (regex-scrubbed) and derivation errors,
+      // never secret bytes. Only the UI update below is mount-gated.
+      logger.error(err instanceof Error ? err : new Error(String(err)), {
```
gbarkhatov, line 678: Same user-rejection concern: this try fires the `deriveVaultRoot` wallet popup (line 652), so a decline is captured here as an `activation.secret` failure with no filter — gate on `classifyError(err) !== 'user-rejection'`. Separately note that genuine on-chain activation failures still aren't captured here either: `handleActivation` swallows them (see the `activation.reveal` comment in `useActivationState.ts`).

services/vault/src/infrastructure/telemetryEvents.ts
```diff
@@ -19,11 +19,32 @@ export const TELEMETRY_EVENT = {
...
+ * the success milestones.
+ */
+export const TELEMETRY_STAGE = {
+  ACTIVATION_WOTS: "activation.wots",
+  ACTIVATION_PAYOUTS: "activation.payouts",
+  ACTIVATION_SECRET: "activation.secret",
+  ACTIVATION_REVEAL: "activation.reveal",
+} as const;
+
+export type TelemetryStage =
```
gbarkhatov, line 45: `TelemetryStage` is exported but referenced nowhere in source or tests (grep returns only this definition; contrast `TelemetryEvent`, which is consumed in `terminalMilestones.ts`). Per the repo ZERO DEAD CODE POLICY, either drop it or put it to work typing the `funnelStage` tag (e.g. `tags: { funnelStage: TelemetryStage }` at the capture sites).

services/vault/src/infrastructure/telemetryEvents.ts
```diff
@@ -19,11 +19,32 @@ export const TELEMETRY_EVENT = {
...
   (typeof TELEMETRY_EVENT)[keyof typeof TELEMETRY_EVENT];
 
+/**
+ * Funnel-stage tags for failure captures. A failure is a `captureException`
+ * (logger.error), not a named message, so tag it with the stage it failed in —
+ * this groups failures by stage in Sentry and lets an alert facet on them. Pair
+ * with a scrubbed `vaultId` in the error's data for the per-vault join back to
+ * the success milestones.
+ */
+export const TELEMETRY_STAGE = {
```
gbarkhatov, line 38: Optional (altitude): the four new failure captures (`usePayoutSigningState.ts:311`, `ResumeDepositContent.tsx:429` & `:677`, `useActivationState.ts:111`) repeat the same `logger.error(err instanceof Error ? err : new Error(String(err)), { tags: { funnelStage }, data: { vaultId: shortId(...) } })` shape. A `captureFunnelFailure(stage, err, data)` helper here would centralize the tag schema, the `shortId` convention, and the user-rejection filter from the comments above — so the fix lands in one place instead of four.

services/vault/src/context/deposit/terminalMilestones.ts
```diff
@@ -0,0 +1,101 @@
...
+    // it now (as PENDING) would emit a spurious milestone one poll later for a
+    // vault that was already terminal.
+    if (activity.contractStatus === undefined) continue;
+    const status = activity.contractStatus as ContractStatus;
+    const reachedVerified =
+      status === ContractStatus.VERIFIED ||
+      status === ContractStatus.ACTIVE ||
+      status === ContractStatus.REDEEMED;
+    const reachedCompleted =
+      status === ContractStatus.ACTIVE || status === ContractStatus.REDEEMED;
```
gbarkhatov: Minor: `reachedCompleted` counts `REDEEMED` but not `LIQUIDATED` (4) or `DEPOSITOR_WITHDRAWN` (6), both of which necessarily passed through `ACTIVE` on-chain. Normal in-session paths are fine (emitted at ACTIVE). But a *seen* vault that jumps straight PENDING→LIQUIDATED between polls never emits `deposit.completed`, whereas the equivalent PENDING→REDEEMED does (there's a test for that). Either include LIQUIDATED/DEPOSITOR_WITHDRAWN, or add a one-line comment that liquidation is deliberately not counted as a conversion. Not blocking.

services/vault/src/context/deposit/terminalMilestones.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    if (!tracking.seen.has(activity.id)) {
```
gbarkhatov, line 126: Minor: seeding can't distinguish 'prior-session already-terminal (suppress)' from 'brand-new deposit whose first poll is already past a milestone (should emit)'. The happy inline path is fine (a fresh vault appears at PENDING and transitions live). The lost case: if registration→activities latency surfaces a vault already at VERIFIED/ACTIVE on first observation, `activation.verified` is silently dropped. This is a sub-case of the session-scoped known limitation the PR body already owns — worth being explicit that slow registration triggers it.

gbarkhatov (dismissed):
Reviewed the funnel-terminal + failure-capture changes. The pure `collectTerminalMilestones` detector is clean and well-tested. A few things worth addressing (details inline):

- **#1 (correctness):** `deposit.completed`/`activation.verified` double-emit when a deposit is completed through the continuation modal — nested `PeginPollingProvider`s each run the emit effect over the same vault, so the headline conversion metric is counted twice on the normal completion path.
- **#2 (correctness):** the `activation.reveal` capture in `useActivationState` is unreachable (the awaited `handleActivation` swallows all errors and never rethrows), so on-chain activation/reveal failures reach Sentry nowhere.
- **#3 (noise):** the payout / WOTS / secret captures log user wallet-rejections as `activation.*` failures. Gate on `classifyError(err) !== 'user-rejection'`.
- **#4 / #5 (cleanup):** unused `TelemetryStage` type; the four capture blocks could share a `captureFunnelFailure` helper.
- Two minor detector notes: LIQUIDATED/DEPOSITOR_WITHDRAWN handling, and seeding vs. slow registration.

Confirmed `activities` is wallet-scoped upstream (`useVaultDeposits(depositorEthAddress)`), so unfiltered emission over the array is fine. Approving — the issues are telemetry-only (no value-movement risk); #1–#3 are worth a follow-up before relying on the funnel counts/alerts.

services/vault/src/context/deposit/terminalMilestones.ts
```diff
@@ -0,0 +1,101 @@
...
+  tracking: TerminalMilestoneTracking,
+): TerminalMilestone[] {
+  const out: TerminalMilestone[] = [];
+
+  for (const activity of activities) {
+    // A missing status means the activity hasn't finished loading; defer both
+    // seeding and emission so it is classified only from a real status. Seeding
+    // it now (as PENDING) would emit a spurious milestone one poll later for a
+    // vault that was already terminal.
+    if (activity.contractStatus === undefined) continue;
```
jrwbabylonlab: This guard never fires in the real pipeline: indexer rows always carry a status, and localStorage-only rows are fabricated with synthetic `PENDING` + `isPending: true` in the `usePeginStorage` merge. So on a cold load with a stale pending row, the tracker seeds from the synthetic PENDING before the first GraphQL resolve, then emits `activation.verified`/`deposit.completed` for a transition that happened entirely while the user was away — the page-load burst seeding exists to prevent, plus a cross-session double-count of `activation.verified`. Skip `isPending` rows here too, so seeding only happens from an indexer-sourced status.

jonybur:
@greptile-apps

jeremy-babylonlabs (approved):

services/vault/src/context/deposit/PeginPollingContext.tsx
jonybur: Removed the ref entirely as part of the double-emit fix above - tracking is now an app-scoped module store (`getSharedTerminalMilestoneTracking`), so there's no per-render `useRef` allocation left to lazy-init. Thanks for the catch.

services/vault/src/context/deposit/terminalMilestones.ts
jonybur: Left as-is after checking it out. A vault can't realistically be first observed already at VERIFIED/ACTIVE: the depositor has to broadcast the Pre-PegIn and wait for confirmations before the graph can be signed, so registration -> activities latency surfaces it at PENDING well before it advances. The header comment already names the two real seeding triggers (a prior-session completion, a resumed deposit); adding slow registration as a third would describe a path that doesn't occur. Happy to revisit if we ever see it in the funnel counts.

services/vault/src/hooks/deposit/useActivationState.ts
jonybur: Confirmed - fixed in #2091. This capture lives in code that merged in #2053, so it's outside this PR's (now rebased) diff. The fix moves the capture into `useVaultActions.handleActivation`'s catch, the only place the error is actually caught since that function never rethrows, and removes the dead one here.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
jonybur: Fixed in #2091 (Phase 2 code, outside this PR's rebased diff). The capture is now gated on `classifyError(err) !== 'user-rejection'`, centralized in a new `captureFunnelFailure` helper so the filter lands in one place across all four sites.

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: Same fix in #2091 - the WOTS capture now goes through `captureFunnelFailure`, which drops user-rejections before they reach Sentry while the UI still surfaces them.

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: Same user-rejection fix in #2091 for the secret capture. And you're right that genuine on-chain activation failures weren't captured here either - #2091 also moves that capture into `handleActivation`'s catch where it's reachable (your `activation.reveal` point).

services/vault/src/infrastructure/telemetryEvents.ts
jonybur: Fixed in #2091 - `TelemetryStage` now types the `stage` parameter of the new `captureFunnelFailure` helper, so it's put to work rather than dropped.

services/vault/src/infrastructure/telemetryEvents.ts
jonybur: Done in #2091 - added `captureFunnelFailure(stage, err, vaultId, extra)` in `telemetryEvents.ts`. It owns the tag schema, the `shortId` on the join key, and the user-rejection filter, and all four capture sites now call it.

gbarkhatov (approved):

PR #2075, opened 2026-07-16
feat(vault): source tx-graph versions from chain at every build site

jrwbabylonlab:
- Replace the TX_GRAPH_VERSION_V1 pins with a required vaultCoreVersion threaded to every WASM build site
- Fresh = activeVaultCoreVersion() (same multicall as offchain params); resume = the vault's stamped on-chain version; version 0 fails closed
- Broadcast gate: stamped === built asserted via verifyRegisteredVaultVersions (+ persisted buildVaultCoreVersion)
- Refund sibling discovery pre-filters by prePeginTxHash so an unrelated invalid vault can't block a refund

Verified live on testnet/devnet contracts (both members present, active = 1, no 0-stamped vaults) — safe to deploy now; no FE redeploy needed when governance flips to v2.


jrwbabylonlab:
Re the two peginStorage points: records carrying the other three build fields could only have been built with graph v1 (every construction site was pinned before this PR), so they're now backfilled to 1 on read instead of filtered. The validator also rejects a stored 0 now. Both in 4fef9842.

gbarkhatov (approved):

PR #2078, opened 2026-07-16
test(vault): e2e resume, self-provisioning setup, 24 words support

### LINKED ISSUE #2070: test(vault): add a resume action to the real-wallet E2E CLI (step 7)
## Step 7 — Resume an in-flight peg-in from the dashboard (real-wallet E2E CLI)

Extends the real-wallet E2E CLI (`services/vault/e2e/real/`) to recover an interrupted peg-in. Builds on connect, single-vault pegin, two-vault split pegin, borrow (https://github.com/babylonlabs-io/babylon-toolkit/issues/2032), repay (https://github.com/babylonlabs-io/babylon-toolkit/issues/2048), and withdraw (https://github.com/babylonlabs-io/babylon-toolkit/issues/2054).

### Goal

Add a `resume` action that continues a peg-in which was interrupted — the deposit modal was closed/refreshed, or the frontend's 20-min WOTS-readiness gate skipped the vault — and drives it from the **dashboard's Pending Deposits UI** through to an activated vault, end-to-end on real signet + Sepolia. This closes the last gap in the depositor lifecycle: today the CLI only drives the happy single-session path, so a WOTS-skip (or any interruption) dead-ends instead of being resumed.

### Why it's needed

A real peg-in spans 30 min–2 hr, and the app is deliberately built to survive interruption: the Pre-PegIn is broadcast on-chain, and if the session is interrupted the dashboard resurfaces the deposit under "Pending Deposits" with a resume action driven from live vault-provider state (no re-peg, no lost funds). The most common trigger is the recoverable `wotsReadinessTimeout` — the Pre-PegIn takes longer than 20 min to confirm on signet, the client gives up, and the deposit must be resumed once the VP catches up.

### Scope — the resume action

Resume enters the **same** signing/activation step machine as the initial pegin, but through a **different front door** and skipping the deposit form:

- **Entry:** dashboard `PendingDepositSection` ("Pending Deposits (N)") → a `PendingDepositCard` (or `BatchedDepositGroup` for a split's two siblings) → a `FullScreenDialog` hosting `PostDepositContinuationView`, which branches on live pegin state and mounts one of four `Resume*Content` components (Broadcast / SubmitWOTS / SignPayouts / Activate). Each renders the **identical `DepositProgressView` stepper** the existing `pegin` step-machine walker already understands.
- **Auto-firing steps:** the resume steps fire on mount — there is no per-step "Sign"/"Continue" click. The only interactions are the dashboard entry click, the Activate confirmation gate (`activate-vault-button`), and the terminal "Go to Dashboard". There is **no artifact-Skip** step in the resume path.
- **Targeting:** default resumes the first actionable pending deposit; `--txid=<prePeginTxid>` selects a specific one when several are pending (a split's two vaults share one Pre-PegIn).
- **Shapes:** single-vault resume and split resume (both siblings via the batched group).

- **Extract the shared step-machine core** — lift `walkStepMachine` (and its Activate / Go-to-Dashboard / progressbar-reader / WOTS-skip-detector selectors) out of `pegin.ts` into a shared module both the `pegin` and `resume` actions import (behavior-preserving; the artifact-Skip gate stays pegin-only). Resume reuses the connect + pop-up approver + recorder scaffolding and the `assertActivatedAndOnDashboard` finish line.
- **`data-testid`s for the resume path** — the dashboard Pending Deposits UI is almost entirely testid-free (only `activate-vault-button` exists). Add a few load-bearing hooks (pending section/card, resume CTA, batched-group controls, the shared progress footer, "Go to Dashboard"), each with an inline comment pointing at the consuming CLI, mirroring the withdraw/borrow/repay testid work.
- **Pre-flight** — refuse the run before opening the browser when the depositor has no resumable in-flight deposit (mirrors the borrow "no collateral" / repay "no debt" gates); validate `--txid` against the live pending list when given.

### Acceptance

- A headed real run on devnet resumes an interrupted deposit from the dashboard to an activated vault, confirmed on-chain (the vault shows as collateral). Verified for a single-vault resume and a split resume.
- Ideally demonstrated by chaining off a real WOTS-skip: let a `pegin` skip at the 20-min gate, wait for the VP to catch up, then `--action=resume` completes it with no re-peg.
- Existing pegin (single + split), borrow, repay, and withdraw still pass — the `walkStepMachine` extraction is behavior-preserving.

- Mock↔real step matrix (scenario composition / start mid-flow).
- The AI runner loop (unattended run → diagnose → fix → rerun).

gbarkhatov:
# Real-wallet E2E CLI: resume an interrupted peg-in, one-command setup, and 24-word wallets

## What this PR does

This PR adds three things to the real-wallet E2E CLI (`services/vault/e2e/real/`), which drives the depositor lifecycle with real browser wallets: (1) a new `resume` action that recovers a peg-in that was interrupted (page refreshed, or the peg-in modal closed), (2) automatic setup so a fresh checkout can run with a single command, and (3) support for 24-word seed phrases when importing wallets.

Closes [#2070](https://github.com/babylonlabs-io/babylon-toolkit/issues/2070).

## Why

A real peg-in takes 30 minutes to a couple of hours and the app is built to survive interruption: once the Pre-PegIn is broadcast the deposit is on-chain, and if you refresh or close the tab the dashboard resurfaces it with a "resume" flow (Submit WOTS → Sign Payouts → Activate). Until now the CLI could only drive a peg-in start-to-finish inside one modal in one session, so it could never test that resume path — the last gap in the lifecycle coverage. On top of that, other engineers who tried to run the CLI hit two setup papercuts (missing wallet extensions, and Playwright's browser not installed), and one engineer's wallet uses a 24-word phrase, which the importer silently got wrong. This PR closes all three.

## 1. Resume an interrupted peg-in

The new `resume` action opens the interrupted deposit from the dashboard's Pending Deposits list and drives it the rest of the way to an activated vault. It has two modes: by default it resumes a deposit that is already pending (optionally targeted with `--txid=<prePeginTxid>`); with `--interrupt-fresh` it is fully self-contained — it pegs in a new deposit, interrupts it right after the Pre-PegIn broadcast (reloads the page to a truly cold state), then resumes it.

Approach and why: the dashboard's resume screens reuse the exact same progress stepper as a normal peg-in, so instead of writing a second driver we extracted the shared step-machine core (`walkStepMachine`, `assertActivatedAndOnDashboard`, and their helpers) out of `pegin.ts` into a new `stepMachine.ts`, and both `pegin` and `resume` now import it. This keeps a single source of truth for "walk the stepper" and means the peg-in path is a pure move with no behaviour change. The one genuinely new thing on a cold resume is that the app re-derives the vault secret (`deriveContextHash` → WOTS) after the reload; the existing pop-up approver already handles that dialog (it is the same one peg-in step 1 shows), so no approver changes were needed. We also added a few load-bearing `data-testid`s to the pending-deposit resume UI so the CLI can find and click the resume controls.

While verifying `--interrupt-fresh` we found and fixed a real bug: right after the reload the app re-hydrates and the "Connect" button briefly renders visible-but-disabled while it auto-reconnects; the reconnect helper was clicking that disabled button and timing out. It now waits for the app to auto-reconnect or for the button to actually be enabled before clicking.

## 2. One-command setup (auto-provisioning)

The CLI now provisions its own prerequisites before launching the browser: **_it installs Playwright's Chromium if it is missing_**, and **_downloads (and updates) the wallet extensions the run needs_**. So a new engineer can clone, `pnpm i && pnpm run build`, and then just run `pnpm --filter vault run e2e:cli` — no separate `playwright install` and no `extensions:download` step.

Approach and why: this lives in a small new `provision.ts` called at the top of the run. It reuses the wallet-connector's existing download primitives rather than reimplementing them (it only adds the "check what's missing and fetch it" orchestration). Both checks are fast no-ops when everything is already present, so provisioned machines pay nothing. We considered making the update check opt-in behind a flag, but the team chose to always check-and-update every run so wallets stay fresh; the extension check degrades gracefully to the on-disk copy if it can't reach the network, and only fails hard when an extension is genuinely missing.

## 3. 24-word mnemonic support

Wallet import now works with 24-word seed phrases, not just 12, across UniSat, OKX, OneKey and MetaMask. Previously the importers assumed a 12-input seed grid, so a 24-word phrase either lost its extra words or got dumped whole into the first box — importing the wrong wallet.

Approach and why: MetaMask and OneKey already handle any length (MetaMask uses a single text box, OneKey fans a paste across the grid and auto-expands it), so they needed no change. UniSat and OKX default to a 12-box grid with a word-count selector that must be switched first. The important constraint here was raised in review: selectors must not depend on the visible wording, because colleagues run their OS in different languages (Russian, Chinese, English) and OKX in particular renders in the OS language, not English. So for OKX we drive the selector by its stable structural data attributes, and for UniSat — whose options are bare spans with no attributes — we match on the digits in the label ("24"), which are identical in every language. This is why the fix does not simply match the text "24 words".

## How it was verified

Auto-provisioning was tested by actually removing each prerequisite and re-running: with Chromium deleted the run re-installed it and continued; with a wallet extension deleted the run re-downloaded and unpacked it. 24-word support was verified by importing a throwaway 24-word phrase into each of the four wallets and confirming the derived address matches the expected one; 12-word imports were re-checked for all four to make sure nothing regressed. The resume flow was driven live on devnet through connect → detect the pending deposit → cold WOTS re-derivation → the progress stepper, and the `--interrupt-fresh` path was driven through peg-in → Pre-PegIn broadcast → interrupt/reload → reconnect. The peg-in path itself was exercised (form → sign → broadcast a real Pre-PegIn) through the refactored code, so this PR does not change peg-in behaviour, and none of the value-critical files were touched.

## Known limitation

Count not observe one fully green resume run all the way to on-chain activation, because the devnet vault provider is currently stalling every deposit at "Step 8 — prepare claim and payout transactions" (multiple deposits have been stuck there for days, and a fresh Pre-PegIn that is already confirmed on signet still isn't advanced). That is a provider/infra issue, not this code — a normal uninterrupted peg-in stalls at the same step right now — and it has been reported to DevOps. Every individual step of the resume path has been exercised; the moment the provider is healthy, a single `resume` run should close that last gap.

### DIFF
--- a/services/vault/e2e/real/actions/resume.ts
+++ b/services/vault/e2e/real/actions/resume.ts
@@ -1,0 +1,314 @@
+/**
+ * The "resume" action: recover an interrupted peg-in from the dashboard's Pending Deposits UI and drive
+ * it through to an activated vault.
+ *
+ * A real peg-in spans 30 min–2 hr and the app is built to survive interruption: once the Pre-PegIn is
+ * broadcast the deposit is on-chain, and the dashboard resurfaces it under "Pending Deposits" with a
+ * resume action (Submit WOTS Key → Sign Payouts → Activate) driven from live vault-provider state. This
+ * action opens that pending card and hands off to the SAME `walkStepMachine` the `pegin` action uses —
+ * the resume flow renders the IDENTICAL `DepositProgressView` stepper — so the walk, the Activate-Vault
+ * gate, and the activated-view finish line are shared, not reimplemented.
+ *
+ * Two shapes:
+ *   - default: resume an ALREADY-pending deposit (by `--txid`, else the first actionable card).
+ *   - `--interrupt-fresh`: peg in a fresh deposit, interrupt it right after Pre-PegIn broadcast (reload
+ *     the page to a cold state), then resume it from the dashboard — a fully self-contained run.
+ *
+ * Cold-resume specifics: unlike the happy path (which holds the vault root in memory), the dashboard
+ * resume RE-DERIVES the vault root from the BTC wallet — a `deriveContextHash` approval dialog — to
+ * recompute the WOTS keys and verify them against the on-chain hash. That dialog is the same one normal
+ * peg-in step 1 (DERIVE_VAULT_SECRET) fires, so the shared pop-up approver already handles it; the
+ * resume steps otherwise auto-fire on mount, so the only dapp-page interactions are the pending-card
+ * click, the Activate gate, and "Go to Dashboard" — all owned by `walkStepMachine`.
+ *
+ * NEVER run `--interrupt-fresh` without an explicit go-ahead: it spends real signet BTC + Sepolia ETH.
+ */
+import type { Page } from "@playwright/test";
+
+import {
+  CONNECT_STATE_POLL_MS,
+  CONNECT_STATE_TIMEOUT_MS,
+  RESUME_ACTIONABLE_TIMEOUT_MS,
+  RESUME_CARD_APPEAR_TIMEOUT_MS,
+  RESUME_POLL_INTERVAL_MS,
+  STEP_TIMEOUT_MS,
+} from "../timing";
+
+import { installPopupApprover, sweepApprovals } from "./approver";
+import { fillDepositForm, startSigning } from "./pegin";
+import { startRecording } from "./recording";
+import {
+  assertActivatedAndOnDashboard,
+  walkStepMachine,
+  walkUntilPrePeginBroadcast,
+} from "./stepMachine";
+import { type Action, type ActionContext } from "./types";
+import { connectWallets } from "./walletConnect";
+
+const DEPOSIT_BUTTON_TESTID = '[data-testid="deposit-button"]'; // connected-state signal (dashboard)
+const CONNECT_BUTTON_TESTID = '[data-testid="connect-wallet-button"]'; // shown when disconnected
+// PendingDepositSection's ExpandMenuButton — the pending list (and each card's resume CTA) lives behind
+// it (aria-label="Pending deposit details"). The button is only rendered when a pending deposit exists,
+// so its presence doubles as "there is something to resume".
+const PENDING_EXPAND_RX = /pending deposit details/i;
+// Added to services/vault/src for this action (see PendingDepositCard.tsx): the card root carries the
+// deposit's Pre-PegIn txid (its tx-hash row links it), and the resume CTA renders ONLY once the deposit
+// is actionable — so waiting for the CTA is waiting for vault-provider readiness.
+const PENDING_CARD_TESTID = '[data-testid="pending-deposit-card"]';
+const PENDING_RESUME_CTA_TESTID = '[data-testid="pending-deposit-resume-cta"]';
+
+/** Normalize a Pre-PegIn txid flag to the bare lowercase hex the card's explorer href contains. */
+function normalizeTxid(txid: string | undefined): string | undefined {
+  const hex = txid?.trim().replace(/^0x/i, "").toLowerCase();
+  return hex && /^[0-9a-f]{64}$/.test(hex) ? hex : undefined;
+}
+
+/**
+ * After the interrupt reload, make sure the app is connected before we look for the pending card (the
+ * pending list reads the connected ETH address). wagmi/AppKit usually auto-reconnects on reload (the
+ * deposit button reappears with no pop-up); if instead the Connect button is shown, re-run the connect
+ * flow. Neither appearing is left to the pending-card wait to surface a clearer error.
+ */
+async function ensureConnected(ctx: ActionContext): Promise<void> {
+  const { page, context, log } = ctx;
+  const deposit = page.locator(DEPOSIT_BUTTON_TESTID).first();
+  const connect = page.locator(CONNECT_BUTTON_TESTID).first();
+  const deadline = Date.now() + CONNECT_STATE_TIMEOUT_MS;
+  while (Date.now() < deadline) {
+    if (await deposit.isVisible().catch(() => false)) return; // auto-reconnected
+    // Only click Connect once it's actually ENABLED. Right after the reload the app re-hydrates and the
+    // Connect button renders visible-but-DISABLED for a beat while wagmi/AppKit auto-reconnects; clicking
+    // that transient disabled button just times out (it detaches the instant the connected state swaps
+    // in). Waiting instead lets the auto-reconnect win and the deposit button appear above.
+    if (
+      (await connect.isVisible().catch(() => false)) &&
+      (await connect.isEnabled().catch(() => false))
+    ) {
+      log("Reconnecting wallets after the interrupt reload");
+      await connectWallets(ctx);
+      return;
+    }
+    await sweepApprovals(context, page, log);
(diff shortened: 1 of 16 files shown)


gbarkhatov:
All greptile concerns solved (comments outside of the diff)

jonybur (approved):
Reviewed the resume action, self-provisioning, and 24-word support. The `pegin` -> `stepMachine.ts` extraction is a faithful move (walk loop, Activate/Skip/Retry gates, WOTS-skip fast-fail, and dashboard cross-check are unchanged; the only behavioral additions are the optional `amountBtc` guard and `walkUntilPrePeginBroadcast`). The seed-grid switching in UniSat/OKX fails loudly when the grid does not match the phrase length instead of silently dropping words, and the locale-independent selector strategy is sound. Product-code changes are limited to load-bearing `data-testid`s with the required consumer-pointing comments per the E2E hooks policy; no value-critical paths touched. One documented limitation to be aware of: a plain `resume` of a split deposit relies on the operator passing `--split` for the correct `expectedVaults` count - acceptable for an operator-driven CLI since it is called out in the code. CI is green.

jrwbabylonlab (approved):

services/vault/e2e/real/actions/resume.ts
```diff
@@ -0,0 +1,361 @@
...
+// Added to services/vault/src for this action (see PendingDepositCard.tsx): the card root carries the
+// deposit's Pre-PegIn txid (its tx-hash row links it), and the resume CTA renders ONLY once the deposit
+// is actionable — so waiting for the CTA is waiting for vault-provider readiness.
+const PENDING_CARD_TESTID = '[data-testid="pending-deposit-card"]';
+const PENDING_RESUME_CTA_TESTID = '[data-testid="pending-deposit-resume-cta"]';
+
+/** Normalize a Pre-PegIn txid flag to the bare lowercase hex the card's explorer href contains. */
+function normalizeTxid(txid: string | undefined): string | undefined {
+  const hex = txid?.trim().replace(/^0x/i, "").toLowerCase();
+  return hex && /^[0-9a-f]{64}$/.test(hex) ? hex : undefined;
```
jrwbabylonlab, line 63: A typo'd `--txid` comes out of here as `undefined` and `openPendingDeposit` silently falls back to the first actionable card — the run then auto-approves WOTS/payout/activation popups against a deposit the operator didn't name (the docstring above `peginUntilInterrupt` warns about exactly this hazard). Throw on a present-but-invalid txid instead of degrading.

services/vault/e2e/real/provision.ts
```diff
@@ -0,0 +1,100 @@
...
+  for (const wallet of wallets) {
+    const storeId = EXTENSION_CHROME_STORE_IDS[wallet];
+    const config = EXTENSIONS.find((extension) => extension.id === storeId);
+    if (!config)
+      throw new Error(
+        `No extension is configured for wallet "${wallet}" (store id ${storeId}).`,
+      );
+    const present = hasExtension(storeId);
+    try {
+      await downloadExtension(config);
```
jrwbabylonlab, line 73: `downloadExtension`'s skip branch resolves on bare `existsSync(crxPath)` without extracting, so a first download interrupted mid-stream leaves a partial .crx that every later run "skips" — provisioning wedges until the file is deleted by hand. Re-check `hasExtension` after this resolves, and delete + retry the stale .crx.

PR #2084, opened 2026-07-16
Fix v3 borrow meter endpoint labels

jeremy-babylonlabs:
## Summary
- prevent nonzero borrow and remaining-capacity meters from displaying misleading 0% or 100% labels
- preserve exact endpoint labels for genuinely empty or full capacity
- add v3 regression coverage for both endpoint-adjacent cases

### Before
Screenshot 2026-07-17 at 02 14 49

### After
Screenshot 2026-07-17 at 02 14 42

## Verification
- `pnpm exec vitest run src/components/simple/__tests__/OverviewSection.test.tsx --maxWorkers=2`
- `pnpm exec eslint src/components/simple/OverviewSection.tsx src/components/simple/__tests__/OverviewSection.test.tsx src/copy.ts`
- `pnpm exec tsc --noEmit -p tsconfig.lib.json`


jonybur (approved):
Reviewed the meter endpoint-label fix. The boundary logic is correct for all four cases: a strictly nonzero ratio that rounds to an endpoint gets the `<1%`/`>99%` label, while genuinely empty/full ratios keep their exact `0%`/`100%` labels, and the two special branches are mutually exclusive so ordering is safe. New strings are in `copy.ts` and both endpoint-adjacent cases have regression tests. Display-only change; CI is green.

PR #2086, opened 2026-07-17
feat(vault): make tx-graph v2 deposits correct and fail closed

jrwbabylonlab:
- Replace the loose HTLC plausibility bound with the exact decomposition htlc = pegin + claim + anchor(v) + minPeginFee(v); bind every byte of the WASM-built PegIn before signing (input outpoint, vault output = requested amount, depositor-claim value + JS-derived script, P2A anchor, txid)
- Reserve the v2 240-sat P2A anchor in the max-deposit math and fee display; new peginP2aAnchorOutput / validatePeginP2aAnchor wrappers
- Fail-closed preflight: fee queries and the deposit CTA wait for supportedTxGraphVersions() ("App update required" when unbuildable); resume actions (broadcast / payout signing / refund) guard before any wallet popup
- Fix the commission preview to charge on the deposit amount only (btc-vault: floor(peginAmount × bps / 10000)) — it previously used the full HTLC value; batch-sized display rows now share submit's effective-split condition
- v2 golden gates pinned alongside v1 (Pre-PegIn hex + derived nVersion-3 PegIn); v1 gates unchanged


jeremy-babylonlabs (dismissed):

services/vault/src/utils/errors/depositErrors.ts
```diff
@@ -118,6 +118,13 @@ export function mapDepositError(err: unknown): DepositErrorContent {
     return ERRORS.wrongWalletAccount;
   }
 
+  // 4c'. Version preflight slipped through (defense in depth): the facade's
+  // stable "unsupported tx graph version" prefix means this build's WASM
+  // cannot construct the required graph — same remedy as the preflight CTA.
+  if (msg.includes("unsupported tx graph version")) {
+    return ERRORS.appVersionUnsupported;
+  }
```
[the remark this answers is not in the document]
jrwbabylonlab: Fixed — mapDepositError now matches the guard's body too, since the resume broadcast surface stringifies the error before mapping.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
```diff
@@ -133,25 +149,21 @@ export async function assertWasmPeginSizing(
...
-          `${impliedFee}).`,
-      );
-    }
-    if (impliedFee > maxImpliedFee) {
+    // Formula: htlcValue = peginAmount + depositorClaimValue +
+    // p2aAnchorValue + minPeginFee. The reserve must match exactly — a
+    // shortfall starves the PegIn's fee/anchor, an excess locks sats
+    // irrecoverably in the HTLC.
+    const impliedReserve = htlcValue - peginAmount - result.depositorClaimValue;
+    if (impliedReserve !== expectedReserve) {
```
gbarkhatov, line 188: This exact-equality check makes the builder's internal `minPeginFee` and the standalone `computeMinPeginFee` entry point a load-bearing identity — if they ever diverge (rounding, a per-version vsize-model change), every v2 deposit hard-fails right here at build time. The good news is this is already CI-guarded: `assertWasmPeginSizing` runs inside `buildPrePeginPsbt` (pegin.ts:204), and the v2 golden test pins `htlcValue === pegin + claim + 240 + 311` (pegin.test.ts:256), so a future vsize drift trips CI rather than production. Not a change request — just worth a one-line note in the test making that guarantee explicit, since the whole v2 path now leans on it.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
```diff
@@ -286,10 +305,197 @@ export async function buildPeginTxFromFundedPrePegin(
...
+}
+
+/**
+ * Independently derive the PegIn depositor-claim output's scriptPubKey in
+ * JS: a Taproot output with the NUMS internal key and a single
+ * `<depositor> OP_CHECKSIG` leaf — btc-vault's `SingleKeyConnector`
+ * (`crates/vault/src/connectors/mod.rs`, identical across graph versions).
+ * Uses the same `tapInternalPubkey` NUMS constant the HTLC connector pins.
+ */
+function deriveDepositorClaimScriptPubKey(depositorPubkey: string): Buffer {
```
gbarkhatov, line 486: Nice independent defense — re-deriving the claim P2TR in JS rather than trusting the binary verbatim. One caveat: this hard-codes the `SingleKeyConnector` shape (NUMS internal key + `<depositor> OP_CHECKSIG` leaf, default leaf version) as "identical across graph versions". If btc-vault ever changes the claim leaf (CSV, leaf version) for a future version, this silently rejects all valid txs. The v2 golden vector guards today's shape; consider a golden vector per supported version so a btc-vault change trips CI here instead of in the field.

services/vault/src/utils/errors/depositErrors.ts
```diff
@@ -118,6 +118,18 @@ export function mapDepositError(err: unknown): DepositErrorContent {
...
   }
 
+  // 4c'. App build can't construct the required graph version: either the
+  // WASM facade threw its stable "unsupported tx graph version ..." error
+  // directly, or assertVaultCoreVersionSupported fired with the user-facing
+  // body (which survives paths that stringify the error, e.g. the resume
+  // broadcast surface). Both get the actionable "App update required" title.
+  if (
+    msg.includes("unsupported tx graph version") ||
+    msg.includes(ERRORS.appVersionUnsupported.body.toLowerCase())
```
gbarkhatov, line 128: This routes the "App update required" state by substring-matching the user-facing copy body. It works today (`msg` is lowercased upstream), but it couples error classification to the exact copy string: reword `appVersionUnsupported.body` and the resume-broadcast surface (which stringifies the error, losing the Error type) silently falls through to the generic error. Consider having `assertVaultCoreVersionSupported` throw a sentinel/typed error (named class or stable code) and matching that here, instead of matching against copy.ts.

services/vault/src/components/simple/SimpleDeposit.tsx
```diff
@@ -230,6 +203,45 @@ function SimpleDepositContent({
     !isSplitCapReached &&
     (!hasActiveVaults || FeatureFlags.isForcePartialLiquidationSplit);
 
+  // Effective split = the same condition handleDeposit uses at submit
+  // (`shouldSplit`). Every batch-sized display row (protocol fee, total
+  // claim reserve, commission) must match what will actually be submitted —
+  // a stale split intent (below split minimum, split no longer allowed)
+  // falls back to a single vault at submit.
+  const isEffectiveSplit = isTwoVaultSplit && allowSplit && !!vaultAmounts;
```
gbarkhatov, line 211: Good fix making `isEffectiveSplit` mirror submit's `shouldSplit` exactly. Note the max-deposit reserve still uses `vaultCount = isTwoVaultSplit && canSplit` (no `allowSplit`) in useDepositPageForm.ts:430 — so when `canSplit && !allowSplit`, Max reserves for 2 vaults while submit does 1. That's a conservative over-reservation (understated Max, never underfunding) and pre-existing, so not blocking — but a one-line comment confirming the divergence is deliberate would help the next reader.

services/vault/src/components/simple/SimpleDeposit.tsx
```diff
@@ -230,6 +203,45 @@ function SimpleDepositContent({
...
+      ? depositorClaimValue * BigInt(depositBatchSize)
+      : undefined;
+
+  // Per-vault deposit amounts the protocol charges commission on
+  // (btc-vault computes `floor(peginAmount × bps / 10000)` per payout —
+  // claim value, PegIn fee, and the P2A anchor are NOT part of the basis).
+  // Split deposits keep per-vault values distinct to preserve each floor.
+  // The undefined placeholder applies only while a feasible split's
+  // amounts still load.
+  const commissionBaseValues =
```
gbarkhatov, line 240: Nit (optional): the middle predicate `isTwoVaultSplit && allowSplit && canSplit` re-derives a subset of `isEffectiveSplit`, so a reader has to mentally subtract branch 1 to see the middle case is "split feasible but `vaultAmounts` not yet loaded". A named `isSplitPending` const or an if/else chain would read top-to-bottom. Not blocking.

services/vault/src/utils/vaultCoreVersionSupport.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+  return supportedVersionsPromise;
+}
+
+/**
+ * Throw a user-facing error when this build's WASM cannot construct the
+ * given vault core version. Call before any signing/broadcast work on a
+ * version sourced from chain (fresh: active; resume: stamped).
+ */
+export async function assertVaultCoreVersionSupported(
```
gbarkhatov, line 34: Altitude question (fine either way): this guard is invoked imperatively at three signing entry points (payoutSigning.ts:77, useVaultActions.ts:192, vaultRefundService.ts:359) plus the query/boolean form in useDepositPageForm.ts. A new signing entry point can silently forget it with no compile error. Each is a genuinely distinct pre-wallet-popup flow and the SDK layer fails closed independently, so this is shippable — but a shared "pre-sign preflight" choke point would make the invariant unforgettable. Worth a maintainer's eye, not a change request.

gbarkhatov (dismissed):
Reviewed the full diff end-to-end (WASM boundary, SDK sizing/binding, and the vault-app fail-closed preflight) with a set of focused correctness, removed-behavior, cross-file, and convention passes. This is a clean critical-path change: the exact `htlc = pegin + claim + anchor(v) + minPeginFee(v)` decomposition plus byte-level PegIn binding (input outpoint, vault value+script, claim value+script, anchor, txid) is the right posture, and the new v2 golden vectors pin every risky value byte-for-byte. Removals (`FEE_SAFETY_MARGIN`, `MAX_REASONABLE_PEGIN_VBYTES`) and the `commissionHtlcValues → commissionBaseValues` rename are fully consistent with no dangling references, and all `getDepositCtaState` / `DepositFeeState` call sites pass the new required fields.

Approving. The inline notes below are non-blocking — coupling/maintainability observations, not correctness defects. Per the critical-path rule this should still get a second human reviewer who can explain each changed line unaided.

jonybur (dismissed):
Reviewed the full diff at head 7ccab33 with focus on the CLAUDE.md critical paths this touches (WASM boundary, fee consistency, presigning preflight):

- The exact reserve identity in `assertWasmPeginSizing` (`htlcValue - pegin - claim === computeMinPeginFee(v) + p2aAnchorValue(v)`) is a correct tightening of the old plausibility cap, and it is honestly documented as a WASM-vs-WASM consistency check rather than an independent recompute.
- `assertPeginTxShape` binds every byte that matters before signing: input outpoint (txid + requested htlcVout), output count per version, vault value against the requested amount and the encoded output, depositor-claim value against `computeMinClaimValue` plus a fully JS-derived claim script, anchor rules via `validatePeginP2aAnchor`, and the reported txid against the encoded tx. Verified `initEccLib(ecc)` runs at the vault app entry (`services/vault/src/main.tsx`) and in the SDK test setup, so the new `payments.p2tr` derivation cannot hit the uninitialized-ECC failure.
- The fail-closed version preflight covers fresh deposits (CTA + fee queries gated on `supportedTxGraphVersions()`) and all three resume surfaces (broadcast, payout signing, refund) before any wallet popup, with the memoized promise correctly clearing on rejection so transient WASM-load failures can retry.
- The commission-basis fix (deposit amount only, floored per vault) matches the vault-output-equals-pegin-amount bind this same PR enforces, and the display rows now share submit's effective-split condition.
- Greptile's earlier findings against 1fb101a are resolved at head: `mapDepositError` now also matches the preflight's user-facing body (with a test), and the removed `FEE_SAFETY_MARGIN` / `MAX_REASONABLE_PEGIN_VBYTES` / `commissionHtlcValues` leave no dangling references (build and full CI green).

No substantive findings. Note this PR touches critical-path files, so the repo's two-code-owner / author-can-explain-every-line bar still applies before merge.

jeremy-babylonlabs (dismissed):

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
jrwbabylonlab: Fixed — added the note on the v2 golden pin.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
jrwbabylonlab: Not planning to fix — both supported versions already have derived-PegIn golden vectors exercising this bind, and an unknown version fails closed at the facade before this derivation runs.

services/vault/src/utils/errors/depositErrors.ts
jrwbabylonlab: Not planning to fix — the matcher reads the same copy constant the guard throws, so a reword updates both together; and the broadcast surface stringifies the error before mapping, so a typed error wouldn't reach the matcher.

services/vault/src/components/simple/SimpleDeposit.tsx
jrwbabylonlab: Fixed — added a comment noting the looser predicate is deliberate (over-reserves, never underfunds).

services/vault/src/components/simple/SimpleDeposit.tsx
jrwbabylonlab: Fixed — named it isSplitPending.

services/vault/src/utils/vaultCoreVersionSupport.ts
jrwbabylonlab: Not planning to fix for now — a forgotten call is backstopped by the facade's fail-closed error plus the mapping to the same copy. A shared pre-sign seam is a good follow-up if the entry points multiply.

gbarkhatov (approved):

jonybur (approved):
Re-reviewed the full diff at the rebased head f31be88, focusing on the two commits since my last pass:

- The reserve check in assertWasmPeginSizing now keeps both layers explicitly: the pure-JS positivity + MAX_REASONABLE_PEGIN_VBYTES plausibility cap (binary-independent, limits what a doctored binary can burn) alongside the exact `reserve === computeMinPeginFee(v) + p2aAnchorValue(v)` identity (WASM-vs-WASM). Keeping the independent bound beside the exact identity is the right posture, and the rewritten constant doc states the trust boundary honestly.
- The version-preflight error now maps on stringifying surfaces too: mapDepositError matches both the facade's "unsupported tx graph version" prefix and the copy body thrown by assertVaultCoreVersionSupported, with tests for each, so the resume-broadcast path can't degrade the fail-closed guard into a generic error.

Everything else matches what I verified at the previous head: the byte-level PegIn binds in assertPeginTxShape (input outpoint + requested htlcVout, vault value = requested amount, JS-derived depositor-claim script, version-shaped output count, anchor rules, txid), the fail-closed preflight on all three resume surfaces before any wallet popup, the commission basis fix to the deposit amount with per-vault flooring, and the effective-split condition shared between display and submit. CI is green.

No substantive findings. This touches critical-path files, so the two-reviewer / author-can-explain-every-line bar still applies before merge.

PR #2090, opened 2026-07-17
feat(vault): add base Banner

jeremy-babylonlabs:
Add a shared full-width Banner to core-ui with critical and notice variant

## Screenshot
Screenshot 2026-07-17 at 18 44 34
Screenshot 2026-07-17 at 18 44 27

### DIFF
--- a/packages/babylon-core-ui/src/components/Banner/Banner.css
+++ b/packages/babylon-core-ui/src/components/Banner/Banner.css
@@ -1,0 +1,19 @@
+.bbn-banner {
+  @apply flex w-full items-center justify-center px-4 text-center text-accent-contrast;
+
+  &-content {
+    @apply flex items-center justify-center gap-4;
+  }
+
+  &-icon {
+    @apply shrink-0;
+  }
+
+  &-critical {
+    @apply bg-error-dark py-2;
+  }
+
+  &-notice {
+    @apply bg-secondary-main py-3;
+  }
+}
--- a/packages/babylon-core-ui/src/components/Banner/Banner.stories.tsx
+++ b/packages/babylon-core-ui/src/components/Banner/Banner.stories.tsx
@@ -1,0 +1,28 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { WarningIcon } from "../Icons";
+import { Text } from "../Text";
+import { Banner } from "./Banner";
+
+const meta: Meta<typeof Banner> = {
+  component: Banner,
+  title: "Components/Data Display/Indicators/Banner",
+  tags: ["autodocs"],
+};
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Critical: Story = {
+  args: {
+    children: <span className="text-base font-bold leading-6 tracking-0.15">Critical — liquidation in 2.3%</span>,
+    icon: <WarningIcon color="text-accent-contrast" size={20} />,
+  },
+};
+
+export const Notice: Story = {
+  args: {
+    children: <Text variant="body2">Deposits are currently disabled while we address an issue.</Text>,
+    variant: "notice",
+  },
+};
--- a/packages/babylon-core-ui/src/components/Banner/Banner.stories.tsx
+++ b/packages/babylon-core-ui/src/components/Banner/Banner.stories.tsx
@@ -4,6 +4,9 @@ import { WarningIcon } from "../Icons";
 import { Text } from "../Text";
 import { Banner } from "./Banner";
 
+// Matches the 20px WarningFilled icon in the Figma banner (node 10088:75288).
+const BANNER_ICON_SIZE = 20;
+
 const meta: Meta<typeof Banner> = {
   component: Banner,
   title: "Components/Data Display/Indicators/Banner",
@@ -16,7 +19,7 @@ type Story = StoryObj<typeof meta>;
 export const Critical: Story = {
   args: {
     children: <span className="text-base font-bold leading-6 tracking-0.15">Critical — liquidation in 2.3%</span>,
-    icon: <WarningIcon color="text-accent-contrast" size={20} />,
+    icon: <WarningIcon color="text-accent-contrast" size={BANNER_ICON_SIZE} />,
   },
 };
 
--- a/packages/babylon-core-ui/src/components/Banner/Banner.tsx
+++ b/packages/babylon-core-ui/src/components/Banner/Banner.tsx
@@ -1,0 +1,27 @@
+import { type HTMLAttributes, type ReactNode } from "react";
+import { twJoin } from "tailwind-merge";
+
+import "./Banner.css";
+
+export type BannerVariant = "critical" | "notice";
+
+export interface BannerProps extends HTMLAttributes<HTMLDivElement> {
+  children: ReactNode;
+  icon?: ReactNode;
+  variant?: BannerVariant;
+}
+
+export function Banner({ children, className, icon, role, variant = "critical", ...props }: BannerProps) {
+  return (
+    <div
+      className={twJoin("bbn-banner", `bbn-banner-${variant}`, className)}
+      role={role ?? (variant === "critical" ? "alert" : "status")}
+      {...props}
+    >
+      <div className="bbn-banner-content">
+        {icon && <div className="bbn-banner-icon">{icon}</div>}
+        {children}
+      </div>
+    </div>
+  );
+}
--- a/services/vault/src/components/pages/RootLayout.tsx
+++ b/services/vault/src/components/pages/RootLayout.tsx
@@ -13,7 +13,13 @@ import {
   useIsMobile,
 } from "@babylonlabs-io/core-ui";
 import { useTheme } from "next-themes";
-import { useCallback, useState } from "react";
+import {
+  type CSSProperties,
+  useCallback,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from "react";
 import { BsDiscord, BsGithub, BsLinkedin } from "react-icons/bs";
 import { FaXTwitter } from "react-icons/fa6";
 import { NavLink, Outlet, useLocation } from "react-router";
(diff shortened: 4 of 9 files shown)


services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -164,6 +170,22 @@ export default function RootLayout() {
...
+    isWalletConnected &&
+    FeatureFlags.isDepositDisabled &&
+    resolveBannerStatus(gate) === null;
+  // The deposit-disabled banner renders full-width above the sidebar (per Figma
+  // node 10084:28515). Measure its height and expose it as a CSS variable so the
+  // sticky v3 sidebar can offset its top/height and avoid clipping its footer.
+  // Zero when the banner is hidden, so the common case is unaffected.
+  const topBannerRef = useRef<HTMLDivElement>(null);
+  const [topBannerHeight, setTopBannerHeight] = useState(0);
+  useLayoutEffect(() => {
```
gbarkhatov, line 186: The measured height never updates on resize — the effect keys only on `showDepositDisabledBanner`, so the one scenario measurement exists for (text wrapping to two lines between the 768px sidebar breakpoint and ~1000px, browser zoom, longer copy, a late webfont swap) is precisely the one it misses. A stale value reproduces the exact bug this PR fixes: sidebar footer clipped, or a gap above it.

Live impact is low today — the current copy is one line at sidebar widths — but the code pays measurement's cost without getting its robustness. There's an in-repo precedent that handles it: `useRailWidth` in `services/vault/src/components/simple/RiskSection/RiskPriceRail.tsx:44-57` (observe once on mount, no visibility dep needed since the wrapper collapses to 0 when the child returns `null`). core-ui also exports `useResizeObserver`, though it takes an element rather than a ref, so the RiskPriceRail shape fits better here.

Alternatively, if the banner is contractually single-line, drop the JS entirely for a fixed height token — that's less machinery than a half-correct measurement.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -214,22 +236,19 @@ export default function RootLayout() {
...
 
   return (
-    <div className="relative flex min-h-svh w-full flex-col bg-surface">
+    <div
+      className="relative flex min-h-svh w-full flex-col bg-surface"
+      style={
+        { "--tbv-top-banner-height": `${topBannerHeight}px` } as CSSProperties
+      }
+    >
+      <div ref={topBannerRef} className="sticky top-0 z-30">
```
gbarkhatov, line 249: Three things about this wrapper, all stemming from it being unconditional:

**Scope.** The clipping bug and the Figma node are v3-desktop-specific, but this applies everywhere. In v2 and on mobile there's no sidebar, nothing consumes `--tbv-top-banner-height`, and the only observable change is that a banner which previously scrolled away now pins to the top. Was that intended for those layouts, or should the `sticky` class be gated on `showV3Sidebar`?

**Split banner stack.** `NoticeBanner`, `AddressScreeningBanner`, `AddressTypeBanner`, and `TestingBanner` still render inside the content column and still scroll away, so there are now two banner placements with different scroll behavior and no note saying why this one is special. It also means a notice-severity banner now sits *above* the critical near-liquidation portal slot (`CRITICAL_BANNER_SLOT_ID`, line 264) — worth confirming that's the intended severity ordering.

**z-index tie.** `.bbn-table-header` is also `sticky top-0 z-30` (`packages/babylon-core-ui/src/components/Table/Table.css:32`). Equal z-index means DOM order decides, and the table header comes later — so when a scrolled table container passes under this banner, its sticky header paints over the banner. Narrow, but worth a look on a v3 page with a tall table.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -214,22 +236,19 @@ export default function RootLayout() {
...
-          state. */}
-      <DepositDisabledBanner
-        visible={
-          !isGeoBlocked &&
-          isWalletConnected &&
-          FeatureFlags.isDepositDisabled &&
-          resolveBannerStatus(gate) === null
-        }
-      />
     </>
```
gbarkhatov, line 239: nit: the comment block above this JSX (lines 216-221) is now partly stale. It describes the sidebar as `sticky top-0 h-svh` — which this PR just changed — and reasons that banners live in the content column *specifically so* their height never pushes the sidebar's sticky position down or clips its footer. That's exactly what the new top wrapper does, then patches back with the CSS var. Defensible trade, but as written the comment reads as contradicting the code 30 lines below it.

services/vault/src/components/shared/AppSidebar.tsx
```diff
@@ -81,6 +81,7 @@ function V3NavLinks() {
 export function AppSidebar() {
   return (
     <Sidebar
+      className="top-[var(--tbv-top-banner-height,0px)] h-[calc(100svh_-_var(--tbv-top-banner-height,0px))]"
```
gbarkhatov, line 84: nit: `--tbv-top-banner-height` is defined in `RootLayout` and consumed here, with no local note pointing at the producer. The `0px` fallback means a rename or a dropped `style` prop degrades silently to the old clipped-footer behavior rather than failing loudly. A one-line comment naming `RootLayout`'s `topBannerRef` block would make the contract greppable from both ends.

packages/babylon-core-ui/src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,27 @@
...
+export interface BannerProps extends HTMLAttributes<HTMLDivElement> {
+  children: ReactNode;
+  icon?: ReactNode;
+  variant?: BannerVariant;
+}
+
+export function Banner({ children, className, icon, role, variant = "critical", ...props }: BannerProps) {
+  return (
+    <div
+      className={twJoin("bbn-banner", `bbn-banner-${variant}`, className)}
```
gbarkhatov, line 17: `twJoin` should probably be `twMerge` here. `BannerProps` extends `HTMLAttributes` and accepts `className`, but with `twJoin` a consumer can't override the base `px-4` or a variant's `py-2`/`py-3` without reaching for `!important` — which is exactly the pattern `CriticalLiquidationTopBanner` had to adopt against `TopBanner`. `Sidebar` in this same package uses `twMerge` for this reason, and it's what makes the `className` override in `AppSidebar` work in this very PR.

packages/babylon-core-ui/src/components/Banner/Banner.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+const meta: Meta<typeof Banner> = {
+  component: Banner,
+  title: "Components/Data Display/Indicators/Banner",
+  tags: ["autodocs"],
+};
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Critical: Story = {
```
gbarkhatov, line 19: The `critical` variant has no production consumer — this story is its only exercise, and it restates a banner that already ships: `services/vault/src/components/simple/CriticalLiquidationTopBanner/CriticalLiquidationTopBanner.tsx` renders the same red strip via `TopBanner` plus `className="!h-10 bg-error-dark !py-2 [&_.bbn-top-banner-message]:!font-semibold [&_.bbn-top-banner-message]:!text-white"`. `.bbn-banner-critical` (`bg-error-dark py-2` + the base `text-accent-contrast`) is that override stack, minus the `!important`s. Shipping an unused variant cuts against the zero-dead-code policy in CLAUDE.md; migrating that component onto `<Banner variant="critical">` deletes the overrides and makes this PR genuinely "shared" instead of adding a fourth banner primitive next to `TopBanner` / `TestingBanner`.

Related, and the reason the migration isn't quite drop-in: the critical typography lives in neither place. This story hand-rolls `text-base font-bold leading-6 tracking-0.15` on a raw `<span>` while the Notice story below uses `<Text variant="body2">`, and the shipping banner separately forces `!font-semibold !text-white` — three consumers, three answers, and they disagree on the weight. If critical is always bold body text, put it in `.bbn-banner-critical` so nobody re-derives it.

If the critical look isn't design-final yet, landing only `notice` and adding the variant with its first real consumer would be the cleaner split.

packages/babylon-core-ui/src/components/Banner/Banner.css
```diff
@@ -0,0 +1,19 @@
+.bbn-banner {
+  @apply flex w-full items-center justify-center px-4 text-center text-accent-contrast;
+
+  &-content {
+    @apply flex items-center justify-center gap-4;
```
gbarkhatov, line 5: nit: the notice banner's gap goes `gap-2` -> `gap-4` in this move (the old markup in `DepositDisabledBanner` used `gap-2`). It's unobservable today since that banner renders no icon, but it'll surface the moment one is added — worth confirming 16px is what Figma specifies for the notice variant and not just inherited from the critical frame.

services/vault/src/components/shared/__tests__/DepositDisabledBanner.test.tsx
```diff
@@ -6,9 +6,10 @@ import { COPY } from "@/copy";
 import { DepositDisabledBanner } from "../DepositDisabledBanner";
 
 describe("DepositDisabledBanner", () => {
-  it("shows the maintenance message when visible", () => {
+  it("renders the maintenance message in the notice banner", () => {
     render(<DepositDisabledBanner visible />);
 
+    expect(screen.getByRole("status")).toHaveClass("bbn-banner-notice");
```
gbarkhatov, line 12: This asserts a core-ui-internal class name from the vault package, which is the kind of implementation coupling CLAUDE.md's test philosophy warns against — `bbn-banner-notice` is a styling detail that a core-ui refactor can rename with no compile error here, and the test would then fail for a reason unrelated to this component's behavior.

`getByRole("status")` already captures the meaningful contract (the notice variant announces non-assertively); the `toHaveClass` adds nothing behavioral. Suggest dropping just the class assertion and keeping the role lookup.

gbarkhatov (approved):
Approving — the core construction is right. Making the deposit-disabled strip `sticky top-0` and shrinking the sidebar's sticky box by exactly its height is the correct fix for the clipped-footer bug, and `Sidebar`'s `twMerge` genuinely lets `top-[var(...)]` / `h-[calc(...)]` override the baked-in `top-0 h-svh`. Verified `z-30` sits below Dialog's `z-40` backdrop, and `bg-error-dark` / `tracking-0.15` are real tokens. Nothing here touches the critical value paths in CLAUDE.md.

Two things I'd like your read on before follow-up work, both inline:

1. **The `critical` variant has no production consumer.** Its only exercise is a Storybook story that restates `CriticalLiquidationTopBanner`, which still goes through `TopBanner` with a stack of `!important` overrides. Migrating that banner is what makes this PR genuinely "shared" rather than adding a fourth banner primitive alongside `TopBanner` / `TestingBanner` / `CriticalLiquidationTopBanner`.
2. **The relocation is unconditional**, but the Figma node and the clipping bug are v3-desktop-specific. Side effects: the notice becomes sticky in v2/mobile where the offset var is unused, and the banner stack is now split, with a notice-severity banner pinned above the critical near-liquidation portal slot.

The rest are fold-in nits. Nothing blocking a merge.

jonybur (approved):
Reviewed the full diff at head b3a8126. The Banner primitive is a clean extraction: variant classes are token-based (bg-error-dark / bg-secondary-main), the role default (alert for critical, status for notice) is right, and the DepositDisabledBanner migration is behavior-preserving with the test updated to pin the notice variant. The sticky relocation plus the measured --tbv-top-banner-height offset on the v3 sidebar is a sound fix for the clipped footer, and the height var correctly resets to 0 when the banner is hidden.

The open points from the earlier review (the critical variant's only consumer being Storybook, and the relocation being unconditional across v2/mobile) are already on record and read as follow-up scope rather than blockers. No new substantive findings; CI is green.

PR #2091, opened 2026-07-17
fix(vault): make the funnel failure captures reachable, quiet and shared

jonybur:
## Context

Follow-up to @gbarkhatov's review on #2069, which flagged four issues in the **Phase 2** failure captures (#2053). Those merged before the review landed, so they're out of scope for Phase 3's diff — split out here as suggested.

Each is against code now in `main`. Every fix has a test verified to fail without it.

## 1. `activation.reveal` was unreachable — on-chain reveal failures reached Sentry nowhere

The capture sat in `useActivationState`'s catch, wrapped around `await vaultHandleActivation(...)`:

```ts
try {
  await vaultHandleActivation({ ... });
} catch (err) {
  logger.error(err, { tags: { funnelStage: TELEMETRY_STAGE.ACTIVATION_REVEAL }, ... });
}
```

But `useVaultActions.handleActivation` wraps its whole body in try/catch and **never rethrows** — its catch only calls `setActivationError` / `setActivationErrorTerminal` / `setActivating`. So the awaited promise resolves even when `activateVaultWithSecret` reverts, and that catch never runs. The try block contains exactly one statement, so there was no other path into it.

Net effect: a wrong secret, a passed activation deadline, or any on-chain revert during the reveal was invisible — not captured there, and not by the `activation.secret` catch either (swallowed before it).

**Fix:** emit from inside `handleActivation`'s catch, the only place the error is observable. Not by making it rethrow — two call sites depend on it resolving.

The caller's catch stays as a defensive state reset, but its dead capture is removed rather than left looking like coverage:

```ts
} catch {
  // Defensive. `vaultHandleActivation` reports its own failures (including
  // the activation.reveal capture) and resolves rather than rethrowing, so
  // this cannot fire today — it only resets local state if that contract
  // ever changes. Capturing here instead would be dead code.
  if (mountedRef.current) setLocalActivating(false);
}
```

## 2. Wallet declines were logged as `activation.*` failures

All four stages front a wallet popup, and a decline is EIP-1193 `4001` / `UserRejectedRequestError` — routine depositor drop-off, not a failure. The payouts guard only filtered `AbortError` (modal closed), which a decline is not, so it fell straight through to the capture — while `formatPayoutSignatureError` on the very next line classified that same error as a benign `user-rejection` for the UI. The two views disagreed about the same error.

**Fix:** gate every capture on `classifyError(err) !== "user-rejection"`. The UI still renders its rejection copy; only the Sentry capture is suppressed. `AbortError` keeps its separate early-return — it must also skip `setError` (the modal is gone), which the rejection path must not.

## 3. + 4. One helper instead of four copies, and `TelemetryStage` put to work

The four captures repeated the same shape, so the fixes above would have had to land four times. `captureFunnelFailure` centralizes the tag schema, the `shortId` convention on the per-vault join key, and the user-rejection filter:

```ts
export function captureFunnelFailure(
  stage: TelemetryStage,
  err: unknown,
  vaultId: string,
  extra: Record<string, string | number | boolean> = {},
): void {
  if (classifyError(err) === "user-rejection") return;
    tags: { funnelStage: stage },
    data: { vaultId: shortId(vaultId), ...extra },
  });
}
```

`vaultId` is shortened inside so the join key can't be forgotten at a call site. This also gives `TelemetryStage` a consumer — it was exported but referenced nowhere, so it's used rather than deleted (zero dead code).

It imports `logger` from the `@/infrastructure` barrel, not `./logger`, so the existing `vi.mock("@/infrastructure")` in tests still intercepts the capture. No cycle: the barrel only re-exports `logger`.

## Testing

- `useVaultActions.test.ts`: the reveal capture fires from where the error is caught, with the `activation.reveal` stage and a scrubbed vaultId; a wallet decline is suppressed but still surfaces to the user.
- `usePayoutSigningState.test.tsx`: a decline is not captured but still renders its error. The `formatting` mock now keeps the real `classifyError` via `importOriginal` and stubs only the message formatter, so the filter is exercised against genuine EIP-1193 classification rather than a stub that could agree with a wrong implementation.
- Each fix mutation-checked: removing the filter fails exactly the two decline tests; removing the reveal capture fails exactly the reveal test.
- Full vault suite green: **226 files, 2395 tests**. eslint clean; `tsc -p tsconfig.lib.json` clean.

### DIFF
--- a/services/vault/src/components/simple/ResumeDepositContent.tsx
+++ b/services/vault/src/components/simple/ResumeDepositContent.tsx
@@ -45,7 +45,10 @@ import { useRequiredPrePeginDepth } from "@/hooks/deposit/useRequiredPrePeginDep
 import { useSplitVaultProgress } from "@/hooks/deposit/useSplitVaultProgress";
 import { useRunOnce } from "@/hooks/useRunOnce";
 import { logger } from "@/infrastructure";
-import { shortId, TELEMETRY_STAGE } from "@/infrastructure/telemetryEvents";
+import {
+  captureFunnelFailure,
+  TELEMETRY_STAGE,
+} from "@/infrastructure/telemetryEvents";
 import {
   ContractStatus,
   getPeginDisplayStep,
@@ -426,12 +429,8 @@ export function ResumeWotsContent({
       // so a real WOTS-submission / derivation-drift failure is worth knowing
       // even if the user has already closed the modal. Only the UI update below
       // is mount-gated. A mismatch is flagged for faceting.
-      logger.error(err instanceof Error ? err : new Error(String(err)), {
-        tags: { funnelStage: TELEMETRY_STAGE.ACTIVATION_WOTS },
-        data: {
-          vaultId: shortId(activity.id),
-          wotsMismatch: isWotsMismatchError(err),
-        },
+      captureFunnelFailure(TELEMETRY_STAGE.ACTIVATION_WOTS, err, activity.id, {
+        wotsMismatch: isWotsMismatchError(err),
       });
       if (mountedRef.current) {
         const msg =
@@ -675,10 +674,7 @@ export function ResumeActivationContent({
       // Capture regardless of mount (no abort signal on this flow). The error
       // message carries only tx hashes (regex-scrubbed) and derivation errors,
       // never secret bytes. Only the UI update below is mount-gated.
-      logger.error(err instanceof Error ? err : new Error(String(err)), {
-        tags: { funnelStage: TELEMETRY_STAGE.ACTIVATION_SECRET },
-        data: { vaultId: shortId(activity.id) },
-      });
+      captureFunnelFailure(TELEMETRY_STAGE.ACTIVATION_SECRET, err, activity.id);
       if (mountedRef.current) {
         const msg =
           err instanceof Error ? err.message : "Failed to activate BTC Vault";
--- a/services/vault/src/hooks/deposit/__tests__/useVaultActions.test.ts
+++ b/services/vault/src/hooks/deposit/__tests__/useVaultActions.test.ts
@@ -35,6 +35,19 @@ vi.mock("@/config/network", () => ({
   getETHChain: vi.fn(() => ({ id: 11155111 })),
 }));
 
+// `captureFunnelFailure` reaches the logger through this barrel, so mocking it
+// here intercepts the capture. `event` must be present: handleActivation's
+// success path calls logger.event, and omitting it would fail the happy paths.
+const mockLoggerError = vi.hoisted(() => vi.fn());
+vi.mock("@/infrastructure", () => ({
+  logger: {
+    error: mockLoggerError,
+    event: vi.fn(),
+    warn: vi.fn(),
+    info: vi.fn(),
+  },
+}));
+
 vi.mock("@babylonlabs-io/ts-sdk/tbv/core", async (importOriginal) => ({
   ...(await importOriginal<typeof import("@babylonlabs-io/ts-sdk/tbv/core")>()),
   ensureHexPrefix: vi.fn((v: string) => (v.startsWith("0x") ? v : `0x${v}`)),
@@ -944,5 +957,54 @@ describe("useVaultActions — handleActivation hashlock source", () => {
     expect(mockActivateVaultWithSecret).toHaveBeenCalledWith(
       expect.objectContaining({ hashlock: ON_CHAIN_HASHLOCK }),
     );
+  });
+
+  // handleActivation catches its own failures and never rethrows, so this catch
+  // is the only place a reveal failure is observable. A capture in a caller's
+  // catch (useActivationState) would never run.
+  it("captures an on-chain reveal failure with the activation.reveal stage and a scrubbed vaultId", async () => {
+    const reader = readerReturning({
+      depositorSignedPeginTx: "0xdeadbeef",
+      hashlock: ON_CHAIN_HASHLOCK,
+    });
+    mockGetVaultRegistryReader.mockReturnValue(reader);
+    mockActivateVaultWithSecret.mockRejectedValue(
+      new Error("execution reverted: InvalidSecret"),
+    );
+
+    const { result } = renderHook(() => useVaultActions());
+
+    await act(async () => {
+      await result.current.handleActivation(baseActivationParams);
+    });
+
+    expect(mockLoggerError).toHaveBeenCalledTimes(1);
+    const [err, ctx] = mockLoggerError.mock.calls[0];
+    expect(err).toBeInstanceOf(Error);
+    expect(ctx.tags.funnelStage).toBe("activation.reveal");
+    expect(ctx.data.vaultId).toBe("0xva...ltId");
+    expect(result.current.activationError).toContain("execution reverted");
+  });
+
+  it("does not capture a wallet decline of the activation tx, but still surfaces it", async () => {
+    const reader = readerReturning({
+      depositorSignedPeginTx: "0xdeadbeef",
+      hashlock: ON_CHAIN_HASHLOCK,
+    });
+    mockGetVaultRegistryReader.mockReturnValue(reader);
+    // EIP-1193 4001 — the depositor hit Reject in their wallet. Routine
+    // drop-off, not a reveal failure; it must not reach Sentry.
+    mockActivateVaultWithSecret.mockRejectedValue(
+      Object.assign(new Error("User rejected the request"), { code: 4001 }),
+    );
+
+    const { result } = renderHook(() => useVaultActions());
+
+    await act(async () => {
+      await result.current.handleActivation(baseActivationParams);
+    });
+
+    expect(mockLoggerError).not.toHaveBeenCalled();
+    expect(result.current.activationError).not.toBeNull();
   });
 });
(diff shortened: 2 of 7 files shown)


jeremy-babylonlabs (dismissed):

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -503,6 +508,11 @@ export function useVaultActions(): UseVaultActionsReturn {
 
       if (mountedRef.current) setActivating(false);
     } catch (err) {
+      // Capture regardless of mount — activation has no abort signal, so a real
+      // reveal failure is worth knowing even if the modal closed mid-flight.
+      // This is the only place the error is caught: handleActivation never
+      // rethrows, so a capture in any caller's catch would be unreachable.
+      captureFunnelFailure(TELEMETRY_STAGE.ACTIVATION_REVEAL, err, vaultId);
```
gbarkhatov: The capture is now correctly placed, but its scope is wider than the `activation.reveal` stage it tags — the `try` spans more than the reveal on both ends.

**Before the reveal — expected states get captured as failures:**

- L424 `throw new Error(COPY.pegin.activationPaused)` — protocol paused. Operator action; when it fires it fires for every depositor at once, spiking the exact rate this tag alerts on.
- L446 `cannotActivateInState(label)` — status isn't `VERIFIED`. This is the concerning one: that branch exists precisely to absorb the indexer-lag race the surrounding comment describes ("a lagging indexer reporting VERIFIED while the contract is still PENDING"), and the code deliberately leaves it retryable. It's a normal, self-resolving transient — likely the single most common thing to land in this catch.

That's the same argument the PR makes for filtering wallet declines.

The pause case also lands inconsistently: the cached-gate check at L381 early-returns *without* capturing, while the fresh-gate check at L424 captures — identical user-visible outcome, one silent and one paged.

**After the reveal — successes get captured as failures:** L489 `calculateBtcTxHash(protocolInfo.depositorSignedPeginTx)` (cross-device resume path, parses tx hex), plus L506-507 `onShowSuccessModal()` / `onRefetchActivities()`, all run *after* `activateVaultWithSecret` resolved. In `useActivationState`, `onShowSuccessModal` does `parseFloat`, `setOptimisticStatus`, `addActivatingVault`. A throw in any of them is captured as `activation.reveal` **and** renders a failure to a depositor whose activation actually succeeded on-chain — inverting the metric's meaning.

Suggestion: suppress the paused case and the non-terminal non-`VERIFIED` case alongside the rejection filter (keep `ActivationNotPossibleError` / EXPIRED — a genuine dead-end worth knowing about), and set a `revealed = true` flag right after `activateVaultWithSecret` resolves, checked before capturing, to close the post-success half.

services/vault/src/infrastructure/telemetryEvents.ts
```diff
@@ -8,6 +8,8 @@
  * followed by the stage and, where it adds signal, the outcome.
  */
 
+import { logger } from "@/infrastructure";
```
gbarkhatov, line 11: Two things about this import, neither blocking.

**It's test-shaped.** Importing from `@/infrastructure` rather than `./logger` is motivated purely by keeping `vi.mock("@/infrastructure")` working, and the comment in the test is honest about that, which I appreciate. But the production import direction is being chosen by test-mock convenience, and the day someone adds `export * from "./telemetryEvents"` to the barrel it becomes a real cycle with no test failure to warn them. Either import `./logger` and mock that in the two test files, or drop a one-line note in `infrastructure/index.ts` about why `telemetryEvents` must not be re-exported.

**This module is no longer constants-only.** Anything importing `TELEMETRY_EVENT` — `useBroadcastState`, `useDepositFlow` — now transitively pulls in the logger, Sentry, and (via `classifyError`) `COPY`. Not a problem at the current call-site count, but if the constants gain more consumers, splitting `captureFunnelFailure` into its own module would keep them cheap.

services/vault/src/infrastructure/telemetryEvents.ts
```diff
@@ -41,6 +43,31 @@ export const TELEMETRY_STAGE = {
...
+ * A wallet decline is routine depositor drop-off, not a failure, so it is never
+ * captured — every one of these stages fronts a wallet popup, and counting
+ * cancellations would inflate the very rate these tags alert on. The caller
+ * still renders its own rejection copy; only the Sentry capture is suppressed.
+ *
+ * `vaultId` is shortened here so the per-vault join key back to the success
+ * milestones cannot be forgotten at a call site. Pass any further identifiers in
+ * `extra` already shortened via `shortId`.
+ */
+export function captureFunnelFailure(
```
gbarkhatov, line 58: This is now the single choke point for four capture sites, but it has no direct unit test — its contract (`vaultId` shortened, non-`Error` values wrapped, `extra` merged, rejection filtered) is only verified indirectly through two of the four callers.

A small `telemetryEvents.test.ts` would be cheap insurance, particularly for the non-`Error` wrapping path, which no current test exercises at all.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -426,12 +429,8 @@ export function ResumeWotsContent({
       // so a real WOTS-submission / derivation-drift failure is worth knowing
       // even if the user has already closed the modal. Only the UI update below
       // is mount-gated. A mismatch is flagged for faceting.
-      logger.error(err instanceof Error ? err : new Error(String(err)), {
-        tags: { funnelStage: TELEMETRY_STAGE.ACTIVATION_WOTS },
-        data: {
-          vaultId: shortId(activity.id),
-          wotsMismatch: isWotsMismatchError(err),
-        },
+      captureFunnelFailure(TELEMETRY_STAGE.ACTIVATION_WOTS, err, activity.id, {
```
gbarkhatov, line 432: This site and the `activation.secret` one below both gained user-rejection suppression as a side effect of adopting the helper, with no test at either.

Worth covering, because they exercise a different branch of `classifyError` than the new tests do. This path signs via `btcWallet: btcWalletProvider`, so a decline arrives from the BTC wallet-connector as `CONNECTION_REJECTED` — a string code matched by `isWalletRejectionError` — not the numeric `4001` that both new tests use. The two branches are independent, so the `4001` tests don't tell you this one works.

services/vault/src/hooks/deposit/__tests__/useVaultActions.test.ts
```diff
@@ -35,6 +35,19 @@ vi.mock("@/config/network", () => ({
   getETHChain: vi.fn(() => ({ id: 11155111 })),
 }));
 
+// `captureFunnelFailure` reaches the logger through this barrel, so mocking it
+// here intercepts the capture. `event` must be present: handleActivation's
+// success path calls logger.event, and omitting it would fail the happy paths.
+const mockLoggerError = vi.hoisted(() => vi.fn());
+vi.mock("@/infrastructure", () => ({
```
gbarkhatov, line 42: Nit: this replaces the whole barrel, so any export the barrel gains later silently becomes `undefined` here rather than failing loudly. Fine today — `infrastructure/index.ts` exports only `logger` — and the comment about `event` being required already hints at the same fragility; just noting it applies to future exports too.

services/vault/src/hooks/deposit/useActivationState.ts
```diff
@@ -110,11 +108,11 @@ export function useActivationState({
             setActivated(true);
           },
         });
-      } catch (err) {
-        logger.error(err instanceof Error ? err : new Error(String(err)), {
-          tags: { funnelStage: TELEMETRY_STAGE.ACTIVATION_REVEAL },
-          data: { vaultId: shortId(activity.id) },
-        });
+      } catch {
```
gbarkhatov, line 111: Flagging this pre-emptively since the repo has a ZERO DEAD CODE policy and a second reviewer may push back: by the PR's own reasoning this catch is unreachable, so it is arguably dead code.

I'd keep it. The comment justifies it well, and it's also the only thing that would ever reset `localActivating` on a failure path. I traced that: today `isActivating` is masked by `&& !activationError`, and `activationError` only clears inside `handleActivation`, which re-sets `localActivating(true)` — so the stale `true` is unobservable and there's no bug hiding here. Worth having the reasoning on the record either way.

gbarkhatov (dismissed):
Approving — the diagnosis is right and I verified it independently: `handleActivation`'s catch calls only `setActivationError` / `setActivationErrorTerminal` / `setActivating`, and the pre-`try` guards can't throw, so the capture in `useActivationState` really was unreachable. Removing it rather than leaving it looking like coverage is the right call — a capture that never fires is worse than none, because it reads as instrumented.

Test rigor is strong. Keeping `classifyError` real via `importOriginal` means the filter is exercised against genuine EIP-1193 classification rather than a stub that could agree with a wrong implementation, and mutation-checking each fix is exactly the right level of care for telemetry, which fails silently. Also confirmed there's no import cycle today: `infrastructure/index.ts` re-exports only `logger`, and nothing re-exports `telemetryEvents`.

One finding worth addressing before merge — the reveal capture's scope is wider than the `activation.reveal` stage, so expected states (notably the indexer-lag `cannotActivateInState` race) will inflate the same rate this tag alerts on, which is the PR's own argument for filtering declines. Details inline. The rest are follow-up material.

jrwbabylonlab (approved):

Issue #2073, opened 2026-07-16
[v3] Protocol paused states (soft / full)


jeremy-babylonlabs:
**Area:** `services/vault` (`protocolStatus.ts`, protocol pause-state query, `copy.ts` protocol-status copy)

**Design:** [SOFT-PAUSED](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10094-25217) · [FULLY PAUSED](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10094-25847)

## Context

On-chain pause detection, operator-flag override, and per-action gating already exist. The v3 work applies the final presentation and wording: soft-paused maps to the existing frozen state where exits remain available, while fully paused maps to the existing paused state.

## Scope

- [ ] Apply the final v3 design and wording to soft- and fully-paused states
- [ ] Integrate the pause presentation with the Overview notification composition
- [ ] Decide with design whether pause banners share the notification container or remain a separate banner surface
- [ ] Preserve all existing gate-state queries and per-action predicates
- [ ] Keep the implementation behind `ENABLE_V3_UI`

## Business-logic invariant

Do not change `pauseState()` reads, scope precedence, operator-flag behavior, or which actions are blocked by frozen versus paused states. This ticket changes presentation and final copy only.

## Acceptance criteria

- [ ] Soft- and fully-paused states match Figma in light and dark
- [ ] Frozen and paused inputs block exactly the same actions as v2
- [ ] Exit actions that remain permitted under frozen state continue to work
- [ ] Protocol-status copy accurately reflects the existing enforced gates

## Depends on

- #2072

Issue #2072, opened 2026-07-16
[v3] Notification states restyle (urgent / cliff A/B / reorder / dust / too-many)


jeremy-babylonlabs:
**Area:** `services/vault` (`PositionNotificationBanner/`, `applications/aave/positionNotifications/`)

**Design:** [CLIFF A](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10093-28840) · [CLIFF B](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10093-20144) · [REORDER](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10094-18130) · [TOO SMALL](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10094-20675) · [TOO MANY](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10094-26791) · [URGENT A](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10204-45310)

## Context

Existing warning types map to the designed states: `urgent`, `cliff`, `reorder`, `dust`, and `too-many-vaults`. Figma splits the current single `cliff` warning into A and B presentations based on its existing position context.

## Scope

- [ ] Restyle urgent, cliff A/B, reorder, dust, and too-many-vault states
- [ ] Select CLIFF A versus CLIFF B presentation from the existing warning payload
- [ ] Preserve existing actions, disabled states, advisory dismissal, and priority behavior
- [ ] Preserve a safe fallback for warning types without a dedicated Figma frame
- [ ] Integrate the notification surface into the v3 Overview composition
- [ ] Keep the implementation behind `ENABLE_V3_UI`

## Business-logic invariant

Do not change calculator warnings, warning thresholds, severity mapping, primary-warning ordering, liquidation math, or action-gating predicates. The CLIFF A/B split is presentation-only and must not introduce a new warning type or protocol state.

## Acceptance criteria

- [ ] Every mapped notification state matches Figma in light and dark
- [ ] The same v2 inputs select the same warning, severity, priority, and permitted actions
[...]

Issue #2071, opened 2026-07-16
[v3] Notifications & Protocol Status


jeremy-babylonlabs:
Integrate the v3 notification and protocol-status presentations into the Overview page while preserving the existing notification engine and protocol action gates.

**Every v3 surface ships behind the feature flag in #2014** so `main` remains deployable with the current v2 dashboard throughout the build.

**Design:** [CLIFF A](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10093-28840) · [URGENT A](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10204-45310) · [Soft paused](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10094-25217) · [Fully paused](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10094-25847)

**Area:** `services/vault` (`PositionNotificationBanner/`, `applications/aave/positionNotifications/`, `protocolStatus.ts`, protocol pause-state query)

## Children

- [ ] #2072 — Notification states restyle (urgent / cliff A/B / reorder / dust / too-many)
- [ ] #2073 — Protocol paused states (soft / full)

## Scope

- Restyle urgent, cliff A/B, reorder, dust, and too-many-vault notification states
- Apply final v3 presentation and wording to soft- and fully-paused states
- Integrate notification cards and banners into the v3 Overview composition
- Preserve existing CTAs, dismissibility, priority, and protocol gating

## Business-logic invariant

This epic is presentation and composition work. Do not change warning calculations, thresholds, warning priority, liquidation classification, pause-state detection, or per-action gating. CLIFF A/B are two presentations of the existing `cliff` warning payload, not new protocol states.

## Open question

The 1025px “Left Pane” frames appear to be design annotation panels rather than an in-app notification center. Confirm with design before adding any new list or notification-center surface.

## Acceptance criteria

- [ ] Every specified notification and pause state matches Figma in light and dark
- [ ] Existing notification selection, priority, CTAs, dismissibility, and action gating remain behaviorally identical to v2
- [ ] V2 behavior remains unchanged while `ENABLE_V3_UI` is off

## Depends on

- #2034
- #2037

PR #2092, opened 2026-07-17
feat(vault): v3 notifications & protocol status

jeremy-babylonlabs:
Implements #2072 and #2073, under epic #2071

## Screenshots
Screenshot 2026-07-21 at 22 22 08
Screenshot 2026-07-21 at 22 21 48


services/vault/src/components/shared/NotificationCardV3.tsx
```diff
@@ -0,0 +1,264 @@
...
+  title?: ReactNode;
+  children?: ReactNode;
+  icon?: ReactNode | null;
+  actions?: NotificationAction[];
+  actionsPlacement?: NotificationActionsPlacement;
+  suggestion?: ReactNode;
+  onClose?: () => void;
+}
+
+export function NotificationCardV3({
```
gbarkhatov: **This is a near-verbatim fork of core-ui's `Notification`.** I diffed the two render bodies with comments/blank lines/indentation stripped (`Notification.tsx:185-257` vs `:188-262` here) and the entire delta is:

```
> data-tone={tone}              # added
< accent.accentBg → accent.chipBg  # renamed
```

plus one prettier reflow of the `hasBelowActions` div. On top of that, `:161-186` re-implements `NotificationActionButton` inline with a byte-identical `ACTION_BASE`, and `isPresent`, `ICON_SIZE`, `CLOSE_ICON_SIZE`, `ON_DARK_ACCENT`, `ON_LIGHT_ACCENT` are copied verbatim. Of 264 new lines, only `TONE_ACCENT` and the `assertive` flag are genuinely new.

I checked the precedent before raising this, and it cuts against the fork. Every other v3 restyle so far keeps the v2 component and branches on the flag *inline* — `FeesSection`, `OverviewSection`, `DashboardPage`, `ActivityList`, `SimpleDeposit`, `UtxoSplitSelector` all do this; there is no other standalone `*V3` component file in `services/vault/src`. And the one time v3 needed a genuinely new primitive, it went into core-ui, not the app (06876b48, "Add a router-agnostic Sidebar/SidebarItem pair to babylon-core-ui").

The "keep v2 untouched" guidance in `markdown/v3-ui/loans.md:65` is about not disturbing the flag-off path — which an added core-ui variant satisfies just as well, since v2 keeps using its existing variants.

The alternative is small: add the seven tones plus `assertive: boolean` to core-ui's `VARIANT_ACCENT`, and export `NotificationActionButton` from `Notification/index.ts` (it isn't exported today, which is presumably why the button got re-implemented). This file then collapses to a ~40-line tone→variant map. Worth noting `assertive`-per-entry is strictly better than core-ui's hardcoded `variant === "error" || "halted"` — that improvement deserves to go upstream rather than be stranded in the app.

Why it matters beyond tidiness: v3 is a multi-PR epic, so this fork lives for months. Every future fix to core-ui `Notification` — a11y, the mandatory `prefers-reduced-motion` handling in `docs/motion-system.md`, layout — silently misses v3, and when v3 becomes the default, v2's accumulated fixes get deleted rather than inherited.

If the fork *is* deliberate (freeze v2 while v3 churns), that's a fine answer — but there's no docblock saying so. Please add one either way; it's the difference between "intentional" and "didn't know `Notification` existed" for the next reader.

services/vault/src/components/shared/NotificationCardV3.tsx
```diff
@@ -0,0 +1,264 @@
...
+export type NotificationCardV3Tone =
+  | "urgent"
+  | "cliff"
+  | "reorder"
+  | "dust"
+  | "too-many"
+  | "soft-paused"
+  | "fully-paused";
+
+const REORDER_BORDER = "border-[#ffb300]";
```
gbarkhatov: This re-declares core-ui's `SUGGESTION_BORDER` (`Notification.tsx:58`) with the same literal, minus the comment that explains it:

> Gold has no core-ui theme token, so the "suggestion" variant uses a one-off accent (Figma #ffb300). Kept as named constants because **Tailwind needs the arbitrary-value class strings to appear literally at build time.**

That second clause is load-bearing — it's what stops someone from "cleaning up" the constant into a template string and silently breaking the generated CSS. Per CLAUDE.md's "if a number appears in code, it must be obvious why that value was chosen," please carry the comment over (or drop the constant entirely if the tones move into core-ui).

services/vault/src/components/shared/NotificationCardV3.tsx
```diff
@@ -0,0 +1,264 @@
...
+    onChip: ON_DARK_ACCENT,
+    actionBg: "bg-secondary-main",
+    onAction: ON_DARK_ACCENT,
+    tint: "",
+    icon: InfoIcon,
+    assertive: false,
+  },
+  reorder: {
+    border: REORDER_BORDER,
+    chipBg: "bg-warning-light",
```
gbarkhatov: Question: `reorder` splits the gold accent. core-ui's `suggestion` variant uses `#ffb300` for both the border and the accent bg; here the border stays `border-[#ffb300]` but chip/action switch to `bg-warning-light`.

Since the reorder card passes `icon={null}` (`PositionNotificationBanner.tsx:338`), the chip never renders — so the only visible effect is that the **Apply Optimal Order** button's fill no longer matches its own card border. Intentional design change, or should `actionBg` also be `bg-[#ffb300]`?

services/vault/src/components/shared/NotificationCardV3.tsx
```diff
@@ -0,0 +1,264 @@
...
+  },
+};
+
+const ICON_SIZE = 24;
+const CLOSE_ICON_SIZE = 14;
+
+const ACTION_BASE =
+  "rounded-full px-4 py-2 text-sm transition-colors disabled:cursor-not-allowed disabled:opacity-50";
+
+const isPresent = (node: ReactNode): boolean => node != null && node !== false;
```
gbarkhatov: Three more explanatory comments dropped in the copy — the code came across, the reasoning didn't:

- **`isPresent`** loses "React drops `false`/`null`/`undefined`, but `0` and `""` are valid content — gate optional slots on absence, not truthiness, so they are not swallowed" (`Notification.tsx:139-140`). Without it this reads like an over-complicated null check and invites simplification to `Boolean(node)`.
- **`CLOSE_ICON_SIZE`** (:113) loses the 14-in-a-24px-frame explanation (`Notification.tsx:125-127`) — otherwise 14 is an unexplained magic number next to `ICON_SIZE = 24`.
- **`centerAlign`** (:158) loses the stacked-content rationale (`Notification.tsx:170-172`).

Same for the `urgent` tint at :46 — `bg-gradient-to-r from-error-dark/[0.06] to-error-dark/[0.06]` is a same-stop gradient, which reads like a mistake until you know it's a `twMerge` workaround (a flat `bg-*` would clobber `bg-secondary-highlight`). core-ui carries that note; this copy dropped it.

services/vault/src/components/shared/NotificationCardV3.tsx
```diff
@@ -0,0 +1,264 @@
...
+        <div className="flex shrink-0 flex-wrap items-center justify-end gap-2">
+          {actionButtons}
+        </div>
+      )}
+
+      {onClose && (
+        <button
+          type="button"
+          onClick={onClose}
+          aria-label="Dismiss notification"
```
gbarkhatov: Minor: `"Dismiss notification"` is an inlined user-facing string in `services/vault/src`. CLAUDE.md requires all vault user-visible copy to live in `copy.ts`, and screen-reader labels are user-visible. (core-ui isn't bound by that rule, so the original is fine where it is — this copy isn't.) Moot if the tones move into core-ui.

services/vault/src/components/simple/PositionNotificationBanner/PositionNotificationBanner.tsx
```diff
@@ -311,6 +316,50 @@ export function PositionNotificationBanner({
     }
   }
 
+  let v3Tone: NotificationCardV3Tone | null = null;
```
gbarkhatov: **This chain isn't exhaustive over `WarningType`.** It covers 4 of the 6 members directly (`urgent`, `cliff`, `too-many-vaults`, `dust`), `reorder` via the `isStandaloneReorder` composite above, and `weird-params` deliberately falls through to v2. Because it's an if/else chain rather than a map, adding a 7th warning type later silently renders the v2 `Notification` under the v3 flag — no compile error, no test failure, and only visible if someone happens to look at that state with the flag on.

This file already establishes the better pattern 260 lines up: `SEVERITY_VARIANT` (:58-66) is a `Record<Exclude<BannerSeverity, "hidden">, NotificationVariant>` with a comment noting it's deliberately total over its domain. A `Record<WarningType, NotificationCardV3Tone | null>` for the `primaryWarning` lookup gets the same compiler guarantee, with `"weird-params": null` *documenting* the intentional fallback instead of leaving it implicit. The `isStandaloneReorder` check stays ahead of the lookup, since that one is a composite condition rather than a plain type match.

services/vault/src/components/simple/PositionNotificationBanner/PositionNotificationBanner.tsx
```diff
@@ -311,6 +316,50 @@ export function PositionNotificationBanner({
...
+    v3Tone = "urgent";
+  } else if (isCliffPrimary) {
+    v3Tone = "cliff";
+  } else if (primaryWarning?.type === "too-many-vaults") {
+    v3Tone = "too-many";
+  } else if (primaryWarning?.type === "dust") {
+    v3Tone = "dust";
+  }
+
+  if (featureFlags.isV3UiEnabled && v3Tone) {
```
gbarkhatov: This return block and the v2 one at :363-390 pass an identical prop set — `title`, `icon`, `actions`, `actionsPlacement`, `suggestion`, `onClose`, `data-testid`, `data-severity`, `{detail}`, and the trailing `ReorderSuccessModal` are all the same. Only the component name and `tone` vs `variant` differ.

That's 28 lines set up to drift: a prop added to one branch and not the other is a silent, flag-dependent bug with no compile error and no test to catch it (the suites exercise the branches separately). Collapsing to a single return that picks the component and the accent prop removes the drift risk — and falls out for free if the fork resolves into core-ui (`variant={v3Variant ?? variant}`).

services/vault/src/components/shared/ProtocolStatusBanner.tsx
```diff
@@ -35,6 +36,22 @@ export function ProtocolStatusBanner() {
...
 
   const copy = COPY.protocolStatus[status];
+
+  if (featureFlags.isV3UiEnabled) {
+    const v3 = COPY.protocolStatus.v3[status];
+    const v3Body = featureFlags.protocolStatusMessage ?? v3.body;
+    return (
+      <Container className={`${PAGE_CONTENT_CLASS} py-6`}>
+        <NotificationCardV3
+          tone={status === "frozen" ? "soft-paused" : "fully-paused"}
```
gbarkhatov: Implicit else on a status mapping. `STATUS_VARIANT` eight lines above (:20-23) maps the same domain with an exhaustive `Record<ProtocolStatus, …>`; this ternary doesn't. If `ScopeStatus` gains a third state, the ternary silently renders it as "fully paused" — the Record fails the typecheck instead:

```ts
const STATUS_TONE: Record<ProtocolStatus, NotificationCardV3Tone> = {
  frozen: "soft-paused",
  paused: "fully-paused",
};
```

Same theme as the tone chain in `PositionNotificationBanner` — cheap now, silent later.

Nit while you're here: `const copy = COPY.protocolStatus[status]` at :38 is computed above this early return and unused on the v3 path. Move it below the branch.

services/vault/src/copy.ts
```diff
@@ -1302,6 +1302,16 @@ export const COPY = {
       // of blocked actions varies. Each affected button explains itself.
       body: "Some actions are temporarily unavailable while the protocol is paused. Any unavailable action is disabled and explains why. Debt continues accruing interest — monitor official announcements.",
     },
+    v3: {
```
gbarkhatov: The v3 bodies are word-for-word copies of v2 with only "frozen"→"soft-paused" and "paused"→"fully-paused" substituted — but none of the three comments came along. The v2 entries carry the INTERIM-copy caveat (:1284-1288) and, twice, the explanation that the wording is *deliberately* non-specific because gating is per-scope (an aave-only freeze leaves deposits working, so naming "deposits and borrows" would be inaccurate).

Those constraints apply identically to the v3 strings, so the next person editing them can't see the trap they're walking into. A short pointer comment on the `v3` block would do it.

Structural note: nesting `v3` as a sibling of `frozen`/`paused` makes `COPY.protocolStatus` heterogeneous — two `ProtocolStatus` keys plus a namespace. `COPY.protocolStatus[status]` typechecks today only because `ProtocolStatus` excludes `"v3"`; a future `Object.keys` or `Record<ProtocolStatus, …>` pass over that section would pick up the stray key. This PR introduces the first `v3:` copy key in the file, so it's setting the precedent for the rest of the epic — a flat `protocolStatusV3` sibling avoids the shape problem and keeps the v2/v3 pair deletable in one edit when the flag retires. Worth a moment's thought either way.

services/vault/src/components/simple/PositionNotificationBanner/__tests__/PositionNotificationBanner.test.tsx
```diff
@@ -40,6 +40,10 @@ vi.mock("@/clients/eth-contract/client", () => ({
 // and surface variant/severity as data-attributes.
 vi.mock("@babylonlabs-io/core-ui", () => ({
   InfoIcon: () => <span data-testid="suggestion-info-icon" />,
+  Text: ({ children }: { children?: ReactNode }) => <div>{children}</div>,
+  WarningIcon: () => <span data-testid="warning-icon" />,
+  PauseIcon: () => <span data-testid="pause-icon" />,
+  CloseIcon: () => <span data-testid="close-icon" />,
```
gbarkhatov, line 46: Worth knowing what these stubs mean for coverage: this suite mocks core-ui wholesale, so the five tones exercised here render `NotificationCardV3` against stub icons and a stub `Text` — what's actually covered is the tone-*selection* logic, not the card. `ProtocolStatusBanner.test.tsx` doesn't mock core-ui, so it genuinely renders the card and covers the role/assertive mapping, but only for 2 of the 7 tones.

Untested anywhere: `actionsPlacement="below"` vs `inline` placement, `emphasis="secondary"` styling, `icon={null}` hiding the chip, `centerAlign`. Low priority if the fork resolves upstream — core-ui's own tests and stories then cover it. If the fork stays, it needs its own test file, because nothing currently detects it drifting from the original.

gbarkhatov (dismissed):
Approving — the v2 path is genuinely untouched, the flag gating is disciplined (every v3 branch gated, both suites reset the flag in `beforeEach`), and the `data-testid` E2E hooks are carried onto both branches.

**Verified locally:** `pnpm exec vitest run` on both changed test files → 43 passed. `eslint` on all three changed source files → clean. The v3 role mapping (frozen→`status`, paused→`alert`) matches core-ui's `variant === "error" || "halted"` rule exactly, so screen-reader behavior doesn't regress under the flag.

Nothing here blocks merge. The one ask I'd like an answer on before the flag is enabled anywhere real is the first comment — `NotificationCardV3` is a ~170-line copy of a design-system primitive, and I diffed the render bodies to confirm it comes down to two substantive lines. Either extending core-ui or a docblock explaining the deliberate fork resolves it; both are cheap now and get expensive once a third v3 card needs the same treatment.

Two comments (the non-exhaustive tone chain, and the duplicated return blocks in `PositionNotificationBanner`) are the spots most likely to produce a real bug later — both are silent, flag-dependent, and compile fine.

One rollout question, not inline since it's scope rather than code: under `ENABLE_V3_UI=true`, `weird-params`, `green`, and `stale-price` still render the v2 `Notification` while the other tones render v3, so a dashboard shows both visual languages at once. The tests assert this deliberately and the PR is marked WIP, so I assume it's scoped — just confirm it's tracked before the flag is turned on beyond devnet.

packages/babylon-core-ui/src/components/Button/Button.css
```diff
@@ -1,5 +1,5 @@
 .bbn-btn {
-  @apply rounded-md tracking-0.4 transition-all duration-200;
+  @apply rounded-lg tracking-0.4 transition-all duration-200;
```
jrwbabylonlab: This restyles every button in every consumer, flag off included — outside the ENABLE_V3_UI constraint. Needs gating or explicit design sign-off that the radius ships globally now.

packages/babylon-core-ui/src/components/Notification/NotificationActionButton.tsx
```diff
@@ -14,28 +14,37 @@ export interface NotificationAction {
 }
 
 const ACTION_BASE =
-  "rounded-full px-4 py-2 text-sm transition-colors disabled:cursor-not-allowed disabled:opacity-50";
+  "rounded-lg px-4 py-2 text-sm transition-colors disabled:cursor-not-allowed disabled:opacity-50";
```
jrwbabylonlab: Same leak here — v2's live notification action buttons go from pills to rounded-lg with the flag off.

services/vault/src/components/simple/PositionNotificationBanner/PositionNotificationBanner.tsx
```diff
@@ -198,6 +219,29 @@ export function PositionNotificationBanner({
...
+    advisoryType !== null &&
+    (DISMISSIBLE_WARNINGS.has(advisoryType) ||
+      (featureFlags.isV3UiEnabled &&
+        V3_DISMISSIBLE_WARNINGS.has(advisoryType)));
+  const dismissibleAdvisoryType = isDismissible ? advisoryType : null;
+  if (
+    dismissibleAdvisoryType &&
+    dismissedAdvisories.has(dismissibleAdvisoryType)
+  )
+    return null;
```
jrwbabylonlab, line 260: When cliff is primary with a too-many-vaults secondary (18+ vaults all consolidating), dismissing the cliff card hides the whole banner including the warning the comment above says is dismissible in neither. Maybe only allow dismiss when there's no non-dismissible secondary — worth a test either way.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -246,22 +274,30 @@ export default function RootLayout() {
...
+
+            Rendered unconditionally, and outside every conditional branch: the
+            consumer (`CriticalLiquidationTopBanner`) resolves this node once on
+            mount via `getElementById` and holds the reference, so if the node
+            were ever unmounted and re-created — e.g. by crossing the 768px
+            `showV3Sidebar` breakpoint — the portal would silently keep writing
+            into a detached element. Owned by the dashboard (where the Aave data
+            + debug override live) but portaled here so it renders above the
+            header and above the deposit-disabled banner. */}
+        <div id={CRITICAL_BANNER_SLOT_ID} />
```
jrwbabylonlab: Heads-up: this relocation reaches v2 too — the critical banner is gated on the liquidation-notifications flag, not v3, so v2 gets it sticky and stacked above DepositDisabledBanner where it previously scrolled away.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -109,6 +109,17 @@ const FOOTER_SOCIAL_LINKS: SocialLink[] = [
...
+// Stacking order of the full-bleed top-banner wrapper.
+// core-ui's Dialog / FullScreenDialog render at `z-50` (backdrop `z-40`) from a
+// portal whose container (`providers.tsx` app root) establishes no stacking
+// context, so they resolve against the root one — the same one this wrapper
+// competes in. Figma §2 (node 10092-19911) requires the critical banner to stay
+// visible above an open modal (e.g. the repay modal), so v3 outranks them.
+// It has to live on the wrapper: `sticky` + a z-index makes the wrapper its own
+// stacking context, so a raised z-index on a banner inside it could not escape.
+// v2 keeps `z-30` so its banner stack still passes under modals exactly as before.
+const TOP_BANNER_Z_CLASS = FeatureFlags.isV3UiEnabled ? "z-[60]" : "z-30";
```
jrwbabylonlab: z-[60] on the shared wrapper also lifts DepositDisabledBanner above dialog backdrops — the design note only asks that of the critical banner.

services/vault/src/copy.ts
```diff
@@ -1470,7 +1487,8 @@ export const COPY = {
         "Consider consolidating smaller BTC Vaults into fewer larger ones — fewer BTC Vaults means lower fees and better optimization.",
     },
     maxVaults: {
-      title: "Maximum BTC Vaults reached",
+      // Figma v3 §9 verbatim: the title drops the "BTC" qualifier the body keeps.
+      title: "Maximum vaults reached",
```
jrwbabylonlab: Shared key, so the v2 title changes too — intended?

gbarkhatov (dismissed):

gbarkhatov (approved):

jonybur (commented):
One issue with the RootLayout banner restructure, in `services/vault/src/components/pages/RootLayout.tsx`:

**The two top banners can no longer actually stick.** Before this change, the wrapper itself was `sticky top-0 z-30`, and its containing block is the full-height root column, so the deposit-disabled banner stayed pinned while the page scrolled. The new markup makes the measured wrapper static and moves `sticky top-0` onto its two children (the critical-banner slot and the deposit-disabled div). A sticky element can only travel within its containing block - here the wrapper - and the wrapper is exactly as tall as those two children combined, so there is no slack:

- the deposit-disabled div (last child) has zero travel and never sticks - a regression that also applies with the v3 flag off, since this markup change is unconditional;
- the critical slot's travel equals the deposit banner's height (usually 0), so it effectively never sticks either.

Net effect: both banners scroll off-screen with the page, and the code comment's "the two therefore stick independently" doesn't hold. The z-index reasoning for splitting them is right, but the split needs the sticky elements to be direct children of the page-height root column (siblings of the content row) rather than nested in a tight wrapper - e.g. render them as two root-level siblings and have the measurement effect observe both nodes with the one ResizeObserver.

Easy to verify manually: enable the deposit-disabled banner on main vs this branch and scroll.

Issue #2043, opened 2026-07-14
[v3] Liquidation Dashboard (new page)


jeremy-babylonlabs:
A new Liquidation Dashboard for understanding and simulating the position's liquidation cascade. It reuses the existing `usePositionNotifications` → `groups` computation; this is new visualization and interaction work, not a second financial model.

**Every v3 surface ships behind the feature flag in #2014** so `main` remains deployable with the current v2 dashboard throughout the build.

**Design:** [Liquidation Dashboard](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10209-67763)

**Area:** `services/vault`, plus `packages/babylon-core-ui` for the reusable chart primitive

## Scope

- Position Overview strip with collateral, debt, Health Factor, and seized/collateral summaries
- BTC price simulator with slider, numeric input, and reset behavior
- Full and compact liquidation waterfall/staircase chart variants driven by cascade groups
- Per-event breakdown cards covering seized vaults, estimated liquidation, post-liquidation position, and outcome status
- Keep simulated-price evaluation isolated from the live notification path

## Data and design decisions

- Confirm whether the chart line is an analytical value-versus-price curve or historical BTC price; the app currently has spot price only
- Derive post-liquidation Health Factor from each group's remaining collateral and debt plus collateral factor
- Choose a lightweight reusable chart implementation that follows the existing no-heavy-library motion constraints

## Acceptance criteria

- [ ] Simulator changes update the chart and event cards in real time without changing live notification state
- [ ] Chart bands and event cards map directly to existing liquidation cascade groups
- [ ] Full and compact chart variants share data and rendering semantics
- [ ] Position and event values use existing protocol calculations, with only post-liquidation Health Factor derived
- [ ] The dashboard matches Figma in light and dark

## Depends on

- #2017

jonybur:
Closing — shipped by #2151 (`0574a455`) and refined by #2160 (`dc458742`, the visx rewrite).

The page lives at `services/vault/src/components/pages/Liquidations/`, and `index.tsx:39` cites this issue directly. `ROUTES.LIQUIDATIONS` is registered in `services/vault/src/routes.ts:11`, and each scope bullet has a file behind it — `PositionOverview.tsx` for the overview strip, `SimulationToolbar.tsx` for the BTC price simulator.

[...]

PR #2093, opened 2026-07-18
feat(vault): seizure map & timeline charts components

jeremy-babylonlabs:
## Summary

Adds a `LiquidationChart` core-ui component (Seizure Map + Timeline sharing one price frame, hand-rolled HTML/CSS so band text truncates instead of overflowing), plus the vault-side data projection, copy strings, and position-notification `params` plumbing that lets the price simulator re-run the pure `calculate()` with an overridden BTC price.

## Screenshots
Screenshot 2026-07-18 at 13 08 19
Screenshot 2026-07-18 at 13 07 50


packages/babylon-core-ui/src/components/LiquidationChart/LiquidationChart.css
```diff
@@ -0,0 +1,515 @@
...
+  --liq-band-text: #ffffff;
+  --liq-candle-up: #34d1a6;
+  --liq-candle-down: #ff5c5c;
+  --liq-surface: #202020;
+  --liq-surface-text: #ffffff;
+  --liq-surface-muted: #b0b0b0;
+  --liq-surface-stroke: #2f2f2f;
+}
+
+:root:not([data-mode="dark"]) {
```
gbarkhatov, line 29: **Dark mode never applies in the vault.** This gates the light palette on `:root:not([data-mode="dark"])`, but the vault signals dark via the `dark` class (`services/vault/tailwind.config.ts:29`, `darkMode: "class"`) — it never sets `data-mode`. So this selector always matches and the chart renders light tokens on a dark page.

The rest of core-ui triples the selector up:

```css
:root.dark .bbn-list-item,
:root[data-mode="dark"] .bbn-list-item,
.dark .bbn-list-item { ... }
```

(`List.css:33-35`, same in `Slider.css:187-193`, `SubSection.css:9-10`.) This needs to match.

While you're here: consider flipping the default so light lives at `:root` and dark is the override, matching the package convention and this file's own comment.

packages/babylon-core-ui/src/components/LiquidationChart/LiquidationChart.css
```diff
@@ -0,0 +1,515 @@
...
+.bbn-liq-chart {
+  --liq-gutter: clamp(3rem, 7cqi, 4.25rem);
+  --liq-band-radius: 2px;
+  --liq-font-axis: clamp(0.625rem, 1.12cqi, 0.75rem);
+  --liq-font-label: clamp(0.7rem, 1.3cqi, 0.875rem);
+  --liq-font-amount: clamp(0.75rem, 1.5cqi, 1rem);
+
+  container-type: inline-size;
+  width: 100%;
+  font-family: "Px Grotesk", sans-serif;
```
gbarkhatov, line 47: Hardcoded font stack. Everything else in the package inherits from the app's typography. Prefer `font-family: inherit` or an existing core-ui token so a font change doesn't have to be chased into this file.

packages/babylon-core-ui/src/components/LiquidationChart/Timeline.tsx
```diff
@@ -0,0 +1,336 @@
...
+  const crosshairEnabled = Boolean(interactions?.crosshair) && candles.length > 0;
+
+  // Derive the time axis from the visible candles so it stays coherent while
+  // panning/zooming; fall back to the caller's static labels without candles.
+  const xAxisLabels = useMemo(() => {
+    if (compact) return undefined;
+    if (!windowed.length) return timeAxisLabels;
+    const ticks = Math.min(7, windowed.length);
+    return Array.from({ length: ticks }, (_, i) => {
+      const idx = Math.round((i / (ticks - 1)) * (windowed.length - 1));
```
gbarkhatov, line 175: **Throws on a single-candle window.** `ticks = Math.min(7, windowed.length)`, so at `windowed.length === 1` this is `0 / 0` → `Math.round(NaN)` → `windowed[NaN]` is `undefined` → `.time` raises a TypeError.

The guard on line 172 only covers the empty case. Reachable with `candles.length === 1`, or `visibleCandles={1}`, or zooming down to one bar.

```ts
if (windowed.length === 1) return [formatTime(windowed[0].time)];
```

or guard `ticks < 2` before the `Array.from`.

packages/babylon-core-ui/src/components/LiquidationChart/Timeline.tsx
```diff
@@ -0,0 +1,336 @@
...
+
+      {/* Candle + interaction region, right of the band gutter. */}
+      <div
+        ref={regionRef}
+        className={interactive ? "bbn-liq-candles bbn-liq-candles--interactive" : "bbn-liq-candles"}
+        style={{ left: pct(gutterFrac) }}
+        onPointerMove={interactive ? onPointerMove : undefined}
+        onPointerLeave={crosshairEnabled ? () => setHoverIndex(null) : undefined}
+        onPointerDown={panEnabled ? onPointerDown : undefined}
+        onPointerUp={panEnabled ? endDrag : undefined}
```
gbarkhatov, line 264: Missing `onPointerCancel`. If the browser cancels the pointer — touch interruption, context menu, an OS gesture — `pointerup` never fires, so `drag.current` stays set and `dragging` stays `true`. The chart then keeps panning on every subsequent move with no button held, and the `grabbing` cursor sticks.

```tsx
onPointerCancel={panEnabled ? endDrag : undefined}
```

packages/babylon-core-ui/src/components/LiquidationChart/PriceFrame.tsx
```diff
@@ -0,0 +1,240 @@
...
+              {currentPriceLabel}
+            </span>
+          ) : null}
+        </div>
+      </div>
+
+      {xAxisLabels?.length ? (
+        <div
+          className="bbn-liq-chart__xaxis"
+          style={plotInsetLeft > 0 ? { paddingLeft: pct(plotInsetLeft) } : undefined}
```
gbarkhatov, line 228: The x-axis inset won't line up with the vertical gridlines, for two independent reasons:

1. **Percentage padding resolves against the wrong box.** `paddingLeft: pct(plotInsetLeft)` resolves against `.bbn-liq-chart`'s width (the containing block), while the candle region's `left: 22%` resolves against `.bbn-liq-chart__plot` — which is chart width *minus* `--liq-gutter` (up to `4.25rem`). Time labels end up roughly `0.22 × gutter` to the right of their gridlines.
2. **`space-between` insets the ends.** The axis distributes label *boxes*, so the first and last are inset by their own text width, whereas gridlines sit at exact fractions. The mismatch grows with label length.

If alignment matters here, position the labels absolutely at the same `tickFraction` values the gridlines use rather than relying on flex distribution.

packages/babylon-core-ui/src/components/LiquidationChart/PriceFrame.tsx
```diff
@@ -0,0 +1,240 @@
...
+          aria-hidden
+        >
+          {xAxisLabels.map((label) => (
+            <span key={label} className="bbn-liq-chart__xaxis-label">
```
gbarkhatov, line 232: `key={label}` collides whenever two ticks format identically. `Timeline` derives these labels from candle timestamps (`Timeline.tsx:172-178`), so any window spanning less than one formatted unit — intraday candles under the default `formatTime` — yields seven identical `"Jun 8"` keys. `shareAxisLabels` collides the same way when two bands round to the same percent.

Key by index.

packages/babylon-core-ui/src/components/LiquidationChart/SeizureMap.tsx
```diff
@@ -0,0 +1,74 @@
...
+      showPriceLineLabel={showPriceLineLabel}
+      priceLineColor={priceLineColor}
+      priceLineLabelColor={priceLineLabelColor}
+      variant={variant}
+      className={twJoin(hideBandLabels && "bbn-liq-chart--no-band-labels", className)}
+      topLegend={compact ? undefined : <ShareLegend bands={bands} />}
+    >
+      {/* Vertical gridlines at the band boundaries (Figma "Background Chart"). */}
+      {[0, ...bands.map((b) => b.shareEnd)].map((f) => (
+        <span key={f} className="bbn-liq-chart__vgridline" style={{ left: pct(f) }} aria-hidden />
```
gbarkhatov, line 62: Same duplicate-key class as `PriceFrame.tsx:232`. `key={f}` over `[0, ...bands.map(b => b.shareEnd)]`: any group with `combinedBtc === 0` gives a `shareEnd` equal to its predecessor, and when `totalBtc === 0` *every* fraction is `0` and all keys collide.

Not hypothetical — `liquidationChartData.ts:134-136` explicitly guards `totalBtc > 0 ? ... : 0`, so that case is already anticipated upstream. Key by index.

packages/babylon-core-ui/src/components/LiquidationChart/BandLayer.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const cancelClose = () => {
+    if (closeTimer.current) {
+      clearTimeout(closeTimer.current);
+      closeTimer.current = null;
+    }
+  };
+  const scheduleClose = () => {
+    cancelClose();
+    closeTimer.current = setTimeout(() => setHovered(null), HOVER_CLOSE_DELAY_MS);
```
gbarkhatov, line 46: `closeTimer` is cleared on hover transitions but never on unmount, so a pending `setTimeout(setHovered)` survives the component. Unmounting mid-hover — tab switch, or a price-simulator re-render — fires `setHovered` on a dead component.

```ts
useEffect(() => cancelClose, []);
```

packages/babylon-core-ui/src/components/LiquidationChart/BandLayer.tsx
```diff
@@ -0,0 +1,150 @@
...
+                  <dt>{m.label}</dt>
+                  <dd className={m.emphasis ? `bbn-liq-popover__value--tone-${hovered.band.tone}` : undefined}>
+                    {m.value}
+                  </dd>
+                </div>
+              ))}
+            </dl>
+            {hovered.band.cumulativeLabel ? (
+              <div className="bbn-liq-popover__row bbn-liq-popover__row--footer">
+                <dt>Cumulative</dt>
```
gbarkhatov, line 129: Two issues in this footer block:

1. **Hardcoded user-facing copy in core-ui.** `"Cumulative"` is the one label the app doesn't supply, which breaks the "the app does ALL formatting" contract that `types.ts` opens with — the vault can't route it through `copy.ts`. Same for the `↓` appended to `bandClickHint` on line 142. Make it a prop (`cumulativeRowLabel`) and let the caller bring its own affordance glyph.
2. **Invalid HTML.** The `<dl>` closes on line 126, so these `<dt>`/`<dd>` sit in a plain `<div>` outside any description list. Move the footer row inside the `<dl>`, or use `<span>`s.

packages/babylon-core-ui/src/components/LiquidationChart/axisDeclutter.ts
```diff
@@ -0,0 +1,51 @@
...
+  center: number;
+  /** Rendered extent along the axis, px. */
+  height: number;
+}
+
+/**
+ * Returns key → resolved centre (px), order preserved. `gap` is the minimum
+ * space between adjacent labels.
+ *
+ * ponytail: assumes the labels' total height fits `track`; denser than that and
```
gbarkhatov, line 21: **Stray `ponytail:` marker** — reads like an internal marker word that leaked into the diff. `grep` confirms it appears nowhere else in the repo (only here and `liquidationChartData.ts:26`), so it isn't an established convention. Rewrite as normal prose or `TODO(#2043)`.

Separately: this module and `scale.ts` are the two things in the PR most worth unit-testing and neither has a test. `declutterCenters` has a three-pass forward/backward/clamp structure plus the overflow failure mode this very comment documents; `segmentedFraction` has piecewise interpolation, clamping, zero-span segments, and a `ticks.length < 2` branch. Meanwhile the straightforward projection layer gets nine tests. Worth inverting that ratio.

packages/babylon-core-ui/src/components/LiquidationChart/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./SeizureGutter";
```
gbarkhatov, line 1: `SeizureGutter` is an internal composition detail of `Timeline` — its own docstring calls it "the pluggable seizure-map column inside the Timeline." Re-exporting it commits the package to its prop shape. Drop it unless standalone use is intended, keeping the public surface to the two charts plus their types.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -0,0 +1,206 @@
...
+ * derivation and formatting happens here.
+ *
+ * Three values are derived (no new financial model — see issue #2043):
+ *  - post-liquidation Health Factor: remaining collateral × price × CF ÷ debt.
+ *    Uses the collateral factor, matching `calculate.ts`, NOT the SDK's
+ *    liquidation-threshold HF.
+ *  - fairness payment in wBTC: `fairnessPaymentUsd / btcPrice`.
+ *  - Sacrificial vs Protected badge: no backing field exists. The first-seized
+ *    group (index 0) is the sacrificial one; the rest are protected.
+ *    // ponytail: index===0 heuristic — confirm the rule with design.
```
gbarkhatov, line 26: **Stray `ponytail:` marker** (also in `axisDeclutter.ts:21`) — rewrite as prose or `TODO(#2043)`.

More importantly, this comment is carrying an unresolved product question into shipped code. The Sacrificial/Protected badge is a pure `group.index === 0` guess with no backing field, and the comment itself says the rule is unconfirmed. A wrong badge tells a depositor the wrong vault is at risk — please get the rule from design before this renders, and encode the answer rather than the heuristic.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -0,0 +1,206 @@
...
+  /** Simulated (or live) BTC price driving which bands read as liquidated. */
+  btcPrice: number;
+  /** Collateral factor (0–1) from on-chain params, for post-liq HF. */
+  collateralFactor: number;
+}
+
+const toneFor = (index: number): LiquidationBandTone =>
+  TONES[index % TONES.length];
+
+function formatSignedPct(pct: number): string {
```
gbarkhatov, line 69: `formatSignedPct` never emits a `+`. Negative values read fine via their own minus sign, but a positive distance renders `10.0%` where the name promises `+10.0%`. Either add the sign for positives or rename to `formatPct`.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -0,0 +1,206 @@
...
+  return `${pct.toFixed(1)}%`;
+}
+
+function hfAfter(
+  group: LiquidationGroup,
+  btcPrice: number,
+  cf: number,
+): string {
+  if (group.debtRemainingAfter <= 0) return "∞";
```
gbarkhatov, line 78: No guard on `btcPrice`. This divides by it on line 80, as does the wBTC fairness conversion on line 94 (`group.fairnessPaymentUsd / btcPrice`). The hook guards `btcPrice <= 0` upstream, but `buildLiquidationChartData` is exported and independently callable — and driving it with an arbitrary price is the entire point of the new `params` plumbing. A `0` from the simulator slider renders `Infinity BTC` / `NaN`.

Per CLAUDE.md's no-silent-fallbacks rule, throw on a non-positive `btcPrice` at the function boundary rather than letting it propagate into labels.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -0,0 +1,206 @@
...
+  const totalBtc = groups.reduce((sum, g) => sum + g.combinedBtc, 0);
+
+  let cumulativeBtc = 0;
+  const bands: LiquidationBand[] = groups.map((group, i) => {
+    const shareStart = totalBtc > 0 ? cumulativeBtc / totalBtc : 0;
+    cumulativeBtc += group.combinedBtc;
+    const shareEnd = totalBtc > 0 ? cumulativeBtc / totalBtc : 0;
+    // Band spans from where this event triggers down to the next event's
+    // trigger; the last band bottoms out at the axis floor (0 = fully wiped).
+    const priceBottom =
```
gbarkhatov, line 139: This assumes `groups` is strictly descending in `liquidationPrice`. If a non-descending pair ever appears, `priceBottom > priceTop` → negative band height → the band silently disappears from the chart with no error.

Given the repo's no-silent-fallback rule, assert the ordering here instead of trusting it.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -0,0 +1,206 @@
...
+    };
+  });
+
+  // Segmented axis: current price, each trigger price, and the floor, sorted
+  // descending and deduped. Sorting (not prepending btcPrice) keeps the axis
+  // ordered when the simulator drops btcPrice below a trigger — otherwise
+  // segmentedFraction collapses the current-price rule and first event to the
+  // top. Dedup avoids duplicate tick values (React keys) when prices coincide.
+  const axisValues = Array.from(
+    new Set([btcPrice, ...groups.map((g) => g.liquidationPrice), 0]),
```
gbarkhatov, line 187: The unconditional `0` floor makes this axis wrong for the `Timeline`. The Seizure Map's *segmented* scale absorbs it fine — every tick gets equal weight regardless of price gap — but `Timeline` maps `linearFraction(priceAxis[0], priceAxis.at(-1))`, i.e. a domain of `[btcPrice, 0]`. With a realistic `btcPrice` around $88k and triggers at $77k/$40k, every candle and level marker gets crushed into the top ~12% of the plot with 60%+ dead space below.

`fixtures.ts` effectively concedes this by defining a separate `timelinePriceAxis` (40k–90k, no zero floor). Since the page isn't in this diff the mismatch is latent — but this function should emit a windowed timeline axis rather than leaving the page to rediscover it.

services/vault/src/applications/aave/hooks/usePositionNotifications.ts
```diff
@@ -31,6 +32,13 @@ export interface UsePositionNotificationsResult {
    * prompt fires. Non-null only when `status === "ready"`.
    */
   reorderVerificationContext: ReorderVerificationContext | null;
+  /**
+   * The exact params `result` was computed from. Non-null only when
+   * `status === "ready"`. Lets what-if consumers (the liquidation
+   * dashboard's price simulator) re-run the pure `calculate()` with an
+   * overridden `btcPrice` without touching this live path.
+   */
+  params: CalculatorParams | null;
```
gbarkhatov, line 41: No consumer reads `params` — `DashboardPage`, `PositionNotificationBanner`, and the dev debug panel all destructure only `result`/`status`. Same story for `buildLiquidationChartData`, the `LiquidationEventCard` projection, and ~25 of the ~30 new `COPY.liquidations` strings: the only references are tests and Storybook stories.

Against the ZERO DEAD CODE POLICY that's the largest single item in the PR. If the consuming page lands in a stacked follow-up, please note that explicitly in the description so the intent is on the record; otherwise trim to what actually ships.

gbarkhatov (approved):
## Overview

Adds a `LiquidationChart` family to core-ui — `SeizureMap` and `Timeline` sharing one `PriceFrame` + `BandLayer`, hand-rolled in HTML/CSS rather than SVG so band text truncates instead of overflowing — plus the vault-side projection, copy, and the `params` plumbing for the price simulator.

The architecture is the right call: core-ui never imports `LiquidationGroup`, and all derivation/formatting stays app-side. Type docs are unusually good, motion uses `--motion-*` tokens per `docs/motion-system.md`, the container-query text-dropout ladder is a nice touch, the CSS-import pattern matches the rest of the package, and the story fixtures are deterministic. The 9 new tests pass.

Approving so this isn't blocked on process, but **two items should land before merge** — flagged inline:

1. **Dark mode never applies in the vault.** `LiquidationChart.css:29` gates the light palette on `:root:not([data-mode="dark"])`. The vault signals dark via the `dark` class (`darkMode: "class"`), never `data-mode` — so that selector always matches and the chart renders light tokens on a dark page. The rest of core-ui triples up (`:root.dark`, `:root[data-mode="dark"]`, `.dark` — see `List.css:33-35`, `Slider.css:187-193`).
2. **`Timeline` throws on a single-candle window.** `Timeline.tsx:175` computes `i / (ticks - 1)`; at `windowed.length === 1` that's `0/0` → `windowed[NaN].time` → TypeError. Reachable via `candles.length === 1` or `visibleCandles={1}`.

## Zero-dead-code policy

The vault half of this PR has no production consumer. `services/vault/src/components/pages/Liquidations/` contains only `liquidationChartData.ts` and its test — there is no page component. Confirmed by grep: `buildLiquidationChartData` / `LiquidationEventCard` are referenced only by their own test; the new `params` field is ignored by `DashboardPage`, `PositionNotificationBanner`, and the debug panel; and roughly 25 of the ~30 new `COPY.liquidations` strings are unreferenced (`heading`, both tab labels, `simulationChip`, `simulateLabel`, `reset`, `bandClickHint`, all of `position.*`, all of `safeZone.*`, `emptyNoWallet`, `emptyNoVaults`, most of `events.*`).

The core-ui components are defensible — Storybook stories are a legitimate consumer. The vault projection, the copy block, and the hook plumbing are scaffolding for a page that isn't in this diff. If the page lands in a stacked follow-up, please say so explicitly in the description; otherwise trim to what ships.

## Test coverage

The two modules most worth testing carry none. `segmentedFraction`/`linearFraction` (`scale.ts`) and `declutterCenters` (`axisDeclutter.ts`) are tricky pure functions with documented edge cases, while the straightforward projection layer gets nine tests. Worth inverting that ratio.

## Security / performance

Nothing concerning — no key material, no external data trusted for security decisions, no critical-path files touched. Perf is sound: `layoutCandles`, `windowed`, `levelMarkers`, and `seriesPoints` are all memoized, the wheel listener is correctly non-passive with cleanup, and there's no per-row hook instantiation. One thing to watch when the page lands: `PriceFrame`'s `useLayoutEffect` depends on `priceAxis` identity, so the consuming page must memoize the `buildLiquidationChartData` result or the `ResizeObserver` gets torn down and rebuilt every render.

Issue #2041, opened 2026-07-14
[v3] Vaults Page


jeremy-babylonlabs:
A dedicated page for the complete vault lifecycle, extracted from the current dashboard sections.

**Every v3 surface ships behind the feature flag in #2014** so `main` remains deployable with the current v2 dashboard throughout the build.

**Design:** [Vaults empty](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10044-12635) · [Vaults full list](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10182-41471)

**Area:** `services/vault`

## Scope

- Vault summary with Total Collateral Value, Active Vaults ordering, Health Factor, Deposit, and Reorder actions
- Pending deposits with status, ETA, custody, pegin/pre-pegin hashes, details, and activation actions
- Ordered active-vault rows with custody, transaction, created-date, withdrawal, and reorder behavior
- Inactive and expired vault rows with withdrawal actions
- Empty state with a Deposit action
- Pending and completed withdrawal sections relocated to `/vaults`

## Data and design decisions

- Confirm whether pending-deposit ETA can be derived from polling state or should remain a static estimate
- Confirm whether the displayed created date is the collateral-added timestamp or the original BTC-deposit timestamp

## Acceptance criteria

- [ ] `/vaults` presents the complete vault lifecycle without duplicating the current dashboard sections
- [ ] Deposit, activation, details, reorder, withdrawal, and pegout-polling behavior remain unchanged
- [ ] Empty, pending, active, inactive, expired, and withdrawal states match Figma in light and dark
- [ ] Summary values and vault ordering match existing dashboard state

## Depends on

- #2017

PR #1671, opened 2026-05-15
chore(vault): remove dead code and prune barrels

jrwbabylonlab:
Cleanup of dead code in `services/vault` flagged by knip and
Pure code removal — no behavior change.

PR #2095, opened 2026-07-20
feat(vault): v3 Vaults page

jonybur:
Builds the v3 `/vaults` page (#2041) behind `NEXT_PUBLIC_FF_ENABLE_V3_UI`, from route registration through the full populated layout.

Screenshot 2026-07-21 at 17 48 15

## What's included

- **Route + shell**: `/vaults` registered inside the Aave provider layout (deep links redirect to `/` when the flag is off). Empty, loading, and load-error states; the emptiness predicate (`useVaultsPageEmptiness`) never claims an empty account on the back of a failed read.
- **Summary card**: total collateral value (BTC + USD), active-vault count with the liquidation-order sequence, and the health factor with the shipped color mapping, plus Deposit and Reorder actions.
- **Pending Deposit section**: one row per in-flight deposit with real step progress (step X of 15 from the polling tree), status dot + tooltip, provider, tx hash, and the state's primary action - broadcast opens the batch-aware broadcast modal, everything else resumes via the deposit multistepper.
- **Active Vaults section**: per-vault rows with liquidation ordinal, in-use status, provider, hash, and a per-row Withdraw that opens the real `WithdrawFlow` with that vault preselected.
- **Inactive Vaults section**: inactive is the v3 name for expired deposits; the row's Withdraw performs the HTLC refund via the existing refund modal. Refunds that are maturing, in flight, or settled render without an action.
- **Copy**: all new strings live in `copy.ts` (including assembled strings like the order sequence, ordinals, and percent labels).

## Review notes

- **Section order** is Pending -> Active -> Inactive. Pending and Inactive share one `ProtocolParamsProvider` + `PeginPollingProvider` tree (mirroring v2's `PendingDepositSection`), so `VaultsLifecycleSections` owns both lists and renders the Active section as `children` between them.
- **Action-flow safety**: `WithdrawFlow` and `ReorderVaultsModal` only ever receive raw indexer-backed collateral entries and demo-unaware financial values. God-mode demo rows are merged for display only (`useVaultsPageData` keeps the two lists separate) and demo ids route to the read-only stepper walk instead of the real modals. Per-row Withdraw passes `vaultId` (the withdraw selection key) and is enabled only for in-use vaults.
- **E2E contract**: the `deposit-button` testid moves between the empty state and the summary card; the two branches are mutually exclusive so the selector stays unique.
- **Dev tooling**: the god-mode panel mounts on the page (dev + flag gated, stripped from production builds), and with demo injection on, the populated layout renders even while disconnected so states can be exercised without a wallet. New eslint dev-boundary seams cover the three files that import `@/dev`.

## Known gaps (follow-ups within #2041)

- The Figma's "Activate within Xh Ym" countdown is not built - there is no wall-clock deadline mechanism today (only the block-based activation-deadline gate), so pending rows show the step position instead.
- Withdrawal sections join the page with the relocation step of #2041; pending withdrawal rows are not on this page yet.

## Testing

- Unit: router guards, `VaultsPage` branch rendering, `useVaultsPageEmptiness` (14 cases), `useVaultsPageData` demo-merge/totals/order. Full vault suite green (224 files / 2386 tests), lint 0 errors.
- Manual: driven with god-mode demo data across deposit, expired, and collateral states in light and dark themes; real-wallet data verified against a funded devnet account (pending deposit with broadcast CTA + active vault).


services/vault/src/components/pages/VaultsPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (hasError) {
+      return (
+        <div className="flex items-center justify-center py-12">
+          <p className="text-base text-accent-secondary">
+            {COPY.vaults.loadError}
+          </p>
+        </div>
+      );
+    }
+    if (!isEmpty) return null;
```
jrwbabylonlab: Non-empty connected positions render a blank container here until the follow-up sections land — a flag-on tester with real collateral sees nothing. Consider a neutral placeholder in the interim.

jonybur:
@greptile-apps

jonybur:
@greptile-apps

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -0,0 +1,522 @@
...
+      return;
+    }
+    if (action === PeginAction.REFUND_HTLC) {
+      onRefund(activity.id);
+      return;
+    }
+    onOpenDetails(activity.id);
+  };
+
+  const hash = activity.prePeginTxHash ?? activity.peginTxHash;
```
jrwbabylonlab, line 152: Pre-broadcast rows will 404 this explorer link — v2 keeps the Pre-PegIn hash copy-only until broadcast (PendingDepositCard's linkPrePegin gate). Same gating needed here.

services/vault/src/hooks/useVaultsPageData.ts
```diff
@@ -0,0 +1,85 @@
...
+      healthFactor,
+      healthFactorStatus,
+    },
+    /** Demo-merged entries for display sections only. */
+    displayVaults,
+    /** Raw indexer-backed entries — the only list action flows may receive. */
+    rawCollateralVaults: collateralVaults,
+    collateralBtc,
+    collateralValueUsd,
+    isLoading,
```
jrwbabylonlab: isLoading has no consumer — drop it.

services/vault/src/hooks/useVaultsPageData.ts
```diff
@@ -0,0 +1,85 @@
...
+      totalCollateralUsd: formatUsdValue(collateralValueUsd),
+      activeVaultsCount: displayVaults.length,
+      liquidationOrder,
+      healthFactor,
+      healthFactorStatus,
+    },
+    /** Demo-merged entries for display sections only. */
+    displayVaults,
+    /** Raw indexer-backed entries — the only list action flows may receive. */
+    rawCollateralVaults: collateralVaults,
```
jrwbabylonlab, line 89: This list also carries the optimistic isActivating entries, so "raw indexer-backed" overstates it — the action surfaces do filter them, but worth filtering here or adjusting the name/doc.

services/vault/src/hooks/useVaultsPageEmptiness.ts
```diff
@@ -0,0 +1,70 @@
...
+    hasDisplayCollateral,
+    isLoading: isPositionLoading,
+    positionError,
+  } = useDashboardState(isConnected ? address : undefined);
+  const {
+    pendingActivities,
+    expiredActivities,
+    isLoading: isDepositsLoading,
+    error: depositsError,
+  } = usePendingDeposits();
```
jrwbabylonlab: This second usePendingDeposits also instantiates an unused broadcast/refund modal state pair — consider passing the lifecycle data in instead.

services/vault/src/components/shared/EmptyState.tsx
```diff
@@ -4,77 +4,67 @@
...
+            />
+          ))}
 
         {/* Primary Text */}
         <p className="text-xl text-accent-primary">{title}</p>
 
         {/* Secondary Text */}
         {description && (
-          <p className="text-sm text-accent-secondary">{description}</p>
+          <p className="max-w-[600px] text-center text-base text-accent-secondary">
```
jrwbabylonlab: Heads-up: this restyle also applies to the v2 reserve-detail connect prompt, outside the flag.

services/vault/src/hooks/useVaultsPageData.ts
gbarkhatov: **Dead code (ZERO DEAD CODE policy).** `isLoading` is returned here, but the only consumer — `VaultsPage` — reads its loading state from `useVaultsPageEmptiness` and never destructures `isLoading` from this hook. Drop it from the return (and the matching `isLoading: false` line in the test mock).

services/vault/src/hooks/useVaultsPageData.ts
```diff
@@ -0,0 +1,85 @@
...
+      displayVaults.map((vault) => formatBtcValue(vault.amountBtc)),
+      btcConfig.coinSymbol,
+    );
+  }, [displayVaults]);
+
+  return {
+    summary: {
+      totalCollateralBtc: formatBtcAmount(shownCollateralBtc),
+      totalCollateralUsd: formatUsdValue(collateralValueUsd),
+      activeVaultsCount: displayVaults.length,
```
gbarkhatov, line 76: **Low / cosmetic.** `activeVaultsCount` and the `liquidationOrder` sequence both include `isActivating` (optimistic) rows, but `VaultsActiveSection` hides the ordinal for activating rows and `reorderableVaults` excludes them. During the indexer gap the summary can show N amounts in the order while the list renders N−1 ordinals. Probably fine — worth a conscious product call.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -0,0 +1,522 @@
...
+                variant="h6"
+                as="h2"
+                className="font-normal text-accent-primary"
+              >
+                {COPY.vaults.sections.pendingDepositsTitle}{" "}
+                <span className="text-accent-secondary">
+                  {COPY.vaults.sections.count(pendingActivities.length)}
+                </span>
+              </Heading>
+              <div className="h-[18px] w-[18px] animate-spin rounded-full border-2 border-accent-primary border-t-transparent" />
```
gbarkhatov: **Motion consistency.** This hand-rolled `animate-spin` div is caught by core-ui's global `prefers-reduced-motion` reset and — unlike the core-ui `Loader` (the one re-enabled functional-spinner exception) — won't spin under reduced motion. Every other spinner in this file uses `<Loader size={16} />`; prefer it here too.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -0,0 +1,522 @@
...
+      {isRefundAvailable && (
+        <button
+          type="button"
+          onClick={() => onRefund(activity.id)}
+          className={PRIMARY_ROW_BUTTON_CLASS}
+        >
+          {COPY.vaults.actions.withdraw}
+        </button>
+      )}
+      {actionStatus.type === "disabled" && isRefundBlocked && (
```
gbarkhatov: **Nit.** `isRefundBlocked` (defined at L261) already requires `actionStatus.type === "disabled"`, so the leading `actionStatus.type === "disabled" &&` here is redundant — `{isRefundBlocked && (…)}` suffices.

services/vault/src/components/vaults/VaultsActiveSection.tsx
```diff
@@ -0,0 +1,153 @@
...
+function ActiveVaultRow({
+  vault,
+  onWithdraw,
+  isWithdrawDisabled,
+}: {
+  vault: CollateralVaultEntry;
+  onWithdraw: (vaultId: string) => void;
+  isWithdrawDisabled: boolean;
+}) {
+  const hash = vault.peginTxHash ?? vault.prePeginTxHash;
```
gbarkhatov, line 40: **Low / confirm intent.** Active rows prefer `peginTxHash ?? prePeginTxHash`, but Pending/Inactive rows use the opposite precedence (`prePeginTxHash ?? peginTxHash`). Plausibly deliberate (pegin is canonical once confirmed; pre-pegin exists first while in flight), but the asymmetry is easy to misread — a one-line comment at each site would help.

services/vault/src/components/pages/VaultsPage.tsx
```diff
@@ -0,0 +1,246 @@
...
+  }, []);
+  const handleWithdrawClose = useCallback(() => setWithdrawVaultIds(null), []);
+
+  // Mirrors CollateralSection: dismissing the success modal hands display
+  // back to the indexer by refetching the order-dependent queries.
+  const handleReorderSuccessClose = useCallback(() => {
+    setIsReorderSuccess(false);
+    if (address) {
+      queryClient.invalidateQueries({
+        queryKey: ["vaultOrder", address.toLowerCase()],
```
gbarkhatov: **Nit.** The `["vaultOrder", address.toLowerCase()]` key is built inline right beside the `invalidateVaultQueries` helper call. It mirrors `CollateralSection`, so it's an established pattern — but centralizing this key with the other vault query keys would prevent drift.

services/vault/src/components/vaults/VaultsSummaryCard.tsx
```diff
@@ -0,0 +1,157 @@
...
+  healthFactorStatus,
+  onDeposit,
+  isDepositDisabled,
+  onReorder,
+  isReorderDisabled,
+}: VaultsSummaryCardProps) {
+  const healthFactorColor = getHealthFactorColor(healthFactorStatus);
+
+  return (
+    <section className="flex w-full flex-col gap-6 rounded-lg border border-secondary-strokeLight bg-secondary-highlight p-6 dark:bg-[#202020] xl:flex-row xl:items-stretch">
```
gbarkhatov, line 96: **Low / pre-existing.** `dark:bg-[#202020]` is a magic hex repeated across the three new components (also `VaultsActiveSection` L41 and `VaultsLifecycleSections` L269). It matches the existing `RiskSection` pattern, so it's consistent rather than novel — but ideally it becomes a token like its light-mode `bg-secondary-highlight` counterpart.

gbarkhatov (dismissed):
Reviewed the diff and verified each note against the checkout. Approving — no correctness or security blockers, and none of the CLAUDE.md critical value-movement paths are touched (this is presentation/routing that delegates to the existing withdraw/reorder/broadcast flows). Test coverage is strong, especially the 14-case emptiness matrix and the demo/real separation in `useVaultsPageData`.

Left a handful of non-blocking inline notes. Only the `isLoading` dead-code cleanup is a policy item (ZERO DEAD CODE) worth doing before merge; the rest are low-priority / nits and several match existing patterns, so treat them as optional.

(One item from my draft — the duplicated `useDashboardState` / `usePendingDeposits` calls across the emptiness and data hooks — I'm intentionally not raising: React Query dedupes the network layer and it matches app conventions, so it's redundant computation only.)

jonybur:
@greptile-apps

jeremy-babylonlabs (approved):

services/vault/src/components/pages/VaultsPage.tsx
```diff
@@ -0,0 +1,253 @@
...
+    if (hasError) {
+      return (
+        <div className="flex items-center justify-center py-12">
+          <p className="text-base text-accent-secondary">
+            {COPY.vaults.loadError}
+          </p>
+        </div>
+      );
+    }
+    if (!isEmpty) return populatedBody;
```
jrwbabylonlab, line 175: This page-level gate can defeat `VaultsLifecycleSections`' keep-mounted-while-modal-open guard: with no collateral and a single pending deposit, if that deposit leaves both lists (e.g. goes INVALID) while the broadcast modal is open mid-signing, `isEmpty` flips and the EmptyState unmounts the provider tree and the open modal. v2 is immune because DashboardPage mounts the section unconditionally. Suggest exposing `hasOpenModal` from the shared `usePendingDeposits` result and gating here with `if (!isEmpty || hasOpenModal) return populatedBody` — the `hasError` branch has the same hole.

services/vault/src/utils/queryKeys.ts
```diff
@@ -9,6 +9,15 @@ import type { Address } from "viem";
 
 import { VAULTS_QUERY_KEY } from "../hooks/useVaults";
 
+/**
+ * Query key for the depositor's vault liquidation order. The address is
+ * lowercased so callers with checksummed and lowercase addresses hit the
+ * same cache entry.
+ */
+export function vaultOrderQueryKey(address: string): [string, string] {
```
jrwbabylonlab, line 17: No query registers `["vaultOrder", ...]` — the only producer (`useVaultOrder`) was removed in the #1671 dead-code prune, so all three invalidation call sites are no-ops (the real refresh comes from `invalidateVaultQueries` next to them). This helper documents a cache entry that doesn't exist; suggest deleting the invalidations and the helper instead of centralizing the dead key.

services/vault/src/components/pages/VaultsPage.tsx
```diff
@@ -0,0 +1,253 @@
...
+  const { isConnected } = useConnection();
+  const { address } = useETHWallet();
+  const gate = useProtocolGateState();
+  const queryClient = useQueryClient();
+  // The page's single usePendingDeposits instance — shared by the emptiness
+  // hook and the lifecycle sections so the broadcast/refund modal state pair
+  // is instantiated once.
+  const deposits = usePendingDeposits();
+  const { isLoading, isEmpty, hasError, hasPartialError } =
+    useVaultsPageEmptiness(deposits);
```
jrwbabylonlab, line 65: `useDashboardState` ends up instantiated twice per render — once inside `useVaultsPageEmptiness`, once inside `useVaultsPageData` — duplicating its subscriptions, memos, and reconciliation effects. The page already threads a single `usePendingDeposits` result for exactly this reason; passing the dashboard state down the same way would match.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -0,0 +1,555 @@
...
+      {actionStatus.type === "available" ? (
+        <button
+          type="button"
+          onClick={() => routeAction(actionStatus.action.action)}
+          className={PRIMARY_ROW_BUTTON_CLASS}
+        >
+          {actionStatus.action.label}
+        </button>
+      ) : actionStatus.type === "disabled" ? (
+        <Hint tooltip={actionStatus.tooltip} attachToChildren>
```
jrwbabylonlab, line 243: Disabled button directly inside `<Hint attachToChildren>` (also line 372) diverges from the existing disabled-control pattern (`DepositButton`, `Connect`), which wraps in a span because disabled form controls suppress mouse events in some browsers — the tooltip may never fire, leaving a disabled action with no explanation.

services/vault/src/components/pages/VaultsPage.tsx
```diff
@@ -0,0 +1,253 @@
...
+          isWithdrawDisabled={isWithdrawBlocked(gate)}
+        />
+      </VaultsLifecycleSections>
+    </div>
+  );
+
+  const renderBody = () => {
+    // Dev-only: with demo injection on, always show the populated layout —
+    // even while disconnected — so the page can be exercised without a wallet.
+    if (isDevToolingEnabled && demo) return populatedBody;
```
jrwbabylonlab, line 158: This branch (the PR-stated "populated layout while disconnected" guarantee) is unreachable in the page tests — `useDemoDeposit` is pinned to `null` in every case — so a `renderBody()` reordering would kill it with no test signal. Also, on that disconnected walk, clicking a demo pending row sets `viewingBatch` but the stepper dialog is gated on `ethAddress`, so it silently no-ops.

services/vault/src/copy.ts
```diff
@@ -1272,6 +1272,47 @@ export const COPY = {
...
+      title: "Some of your BTC Vault data couldn't be loaded",
+      body: "Totals or deposits shown may be incomplete. Refresh the page to try again.",
+    },
+    summary: {
+      totalCollateralLabel: "Total Collateral Value",
+      activeVaultsLabel: "Active Vaults",
+      healthFactorLabel: "Health Factor",
+      healthFactorCaption:
+        "When the ratio falls below 1.0, liquidation may occur.",
+      vaultCount: (count: number) =>
```
jrwbabylonlab, line 1294: `"1 Vault"/"N Vaults"` runs against this file's own "BTC Vault, never bare vault" rule — the same page shows "BTC Vaults" in the empty state and "2 Vaults" here. Section headings have shipped bare-"Vaults" precedent; this rendered stat value is the one worth aligning (or getting a ruling on).

PR #2099, opened 2026-07-21
feat(vault): add daemon and pegout terminal events (Phase 4) to Sentry

jonybur:
## Context

Phase 4 of the Sentry funnel tracking (#2052 → #2053 → #2069 → #2091). This closes the remaining **silent-drop blind spots**: a deposit the VP daemon terminally rejects, and a redemption that completes, blocks, or stalls, all stopped polling and transmitted **nothing** - invisible to monitoring.

Based directly on `main` (all prior phases merged).

## What it adds

Four events, per-vault with a scalar `vaultId`, all transition-diffed with the discipline Phase 3 established (pure collector + shared in-memory tracking + thin effect; already-terminal vaults are **seeded** on first observation, so a dashboard load never emits a burst for prior-session terminals):

| Event | Level | When | Facet |
| --- | --- | --- | --- |
| `activation.daemon.terminal` | warning | VP daemon reports a terminal drop | `daemonStatus` (Expired / AmlRejected / IngestionRejected / InvalidSigInContract / ExpiredCleanedUp / ExpiredInClaim) |
| `exit.redeem.payout_broadcast` | info | Redemption success terminal - BTC payout broadcast | - |
| `exit.redeem.payout_blocked` | warning | Claim/assert on-chain but payout blocked | - |
| `exit.redeem.pegout_timeout` | warning | Pegout polling gave up | `timeoutReason` (consecutive_failures / unknown_status / provider_not_found) |

**Alerts these enable:** AML/ingestion rejection spike faceted by `daemonStatus` (VP or compliance pipeline regression), stuck-redemption page (`pegout_timeout` rate + `payout_blocked`), and the redemption success terminal for eventual initiated→broadcast conversion.

## Design notes

- **Daemon-terminal seeding keys off the *polled set*, not the errors map.** A vault only enters the errors map once it errors - so its first appearance there would always look pre-existing and nothing would ever emit. The collector takes the polled ids as the observation universe and the errors map as the classification source.
- **Dedup is per (vault, status/kind), not per vault.** A real `Expired → ExpiredCleanedUp` progression yields both signals, and a timed-out pegout that later recovers to `payout_broadcast` still emits the success terminal.
- `PegoutPollingResult` gains an optional `timeoutReason`, set at the three existing give-up sites - so the collector classifies from data instead of reference-comparing `TIMED_OUT_STATE`, and the reason becomes a Sentry facet for free. Additive; no display behavior changes.
- Non-terminal errors (provider connectivity) are ignored entirely - they're transient and already breadcrumbed.

## Testing

- `daemonTerminalEvents.test.ts` (4): healthy→terminal emits once; already-terminal-at-first-observation seeds silently; `Expired → ExpiredCleanedUp` emits both; connectivity errors never emit.
- `pegoutTerminalEvents.test.ts` (5): each terminal kind emits once with the right level/facet; no-burst seeding; timeout→recovery still emits the success terminal.
- Full vault suite green: **234 files, 2436 tests**. eslint clean; `tsc -p tsconfig.lib.json` clean (after the documented core-ui/ts-sdk `dist` rebuild).

## Review follow-ups (728ec508)

- **Atomic polled-id snapshot.** `PollingQueryData` now carries `polledIds`, captured inside the queryFn, and the daemon-terminal effect collects from `data.polledIds` + `data.errors` - always the same resolved poll, so a wallet switch or late-arriving activities under `keepPreviousData` can no longer seed new vaults against a stale errors map. The hook's returned `depositsToPoll` lost its last consumer and was removed.
- **Failed observations no longer seed.** Found while testing the above - a second route to the prior-session burst with zero snapshot tearing: a vault whose first observation is a failed poll (plain connectivity `Error` on the pegin side; a result with no `response` and no `timeoutReason` on the pegout side) was seeded as "seen healthy" even though the poll learned nothing, so its prior-session terminal emitted as fresh on the next successful poll. Both collectors now skip failed observations entirely; seeding waits for a genuinely observed status.
- **Emission-level integration tests.** `daemonTerminalEmission.test.tsx` (4) mounts the real provider (incl. co-mounted providers); `pegoutTerminalEmission.test.tsx` (3) drives the real `usePegoutPolling` through react-query with only the SDK batch poller mocked. Both consume the `resetShared*` exports in `beforeEach` and pin the effect wiring: the `polledIds` id source, resolved-poll guards, `shortId`, level/category, and `timeoutReason` absence on non-timeout payloads.
- **Timeouts dedup per give-up reason** (`timeout:<reason>`), mirroring the daemon side's per-status keying - a re-timeout down a different give-up path emits the new reason; a repeat of the same reason stays deduped. Collector unit tests grew to 6 (daemon) and 8 (pegout).
- Types: `daemonStatus` is `DaemonStatus`; `timeoutReason` keeps its three-literal union. Telemetry header now lists `exit` as a funnel phase.

## Remaining backlog (Phase 5)

`activation.wots.submitted` / `activation.payouts.signed` (need transition-based emission at the localStorage status advance - the SDK calls are idempotent, so emit-on-resolution would double-count on resume), `exit.redeem.initiated` (completes the redemption conversion pair), and the `depositStep` tag on the inline outer catch.

### DIFF
--- a/services/vault/src/context/deposit/PeginPollingContext.tsx
+++ b/services/vault/src/context/deposit/PeginPollingContext.tsx
@@ -22,7 +22,7 @@ import {
 
 import { useDemoDeposit } from "@/dev/demoDeposit";
 import { logger } from "@/infrastructure";
-import { shortId } from "@/infrastructure/telemetryEvents";
+import { shortId, TELEMETRY_EVENT } from "@/infrastructure/telemetryEvents";
 
 import { usePeginPollingQuery } from "../../hooks/deposit/usePeginPollingQuery";
 import { useRequiredPrePeginDepthResolver } from "../../hooks/deposit/useRequiredPrePeginDepth";
@@ -57,6 +57,10 @@ import { isVaultOwnedByWallet } from "../../utils/vaultWarnings";
 import { useProtocolParamsContext } from "../ProtocolParamsContext";
 
 import { computeDepositPollingResult } from "./computeDepositPollingResult";
+import {
+  collectDaemonTerminalEvents,
+  getSharedDaemonTerminalTracking,
+} from "./daemonTerminalEvents";
 import {
   collectTerminalMilestones,
   getSharedTerminalMilestoneTracking,
@@ -146,6 +150,7 @@ export function PeginPollingProvider({
     pendingDepositorSignatures,
     isLoading,
     refetch,
+    depositsToPoll,
   } = usePeginPollingQuery({
     activities,
     pendingPegins,
@@ -366,6 +371,28 @@ export function PeginPollingProvider({
       });
     }
   }, [activities]);
+
+  // Emit activation.daemon.terminal once per (vault, daemonStatus) as the VP
+  // daemon reports a terminal drop (Expired / AmlRejected / ...). These states
+  // stop polling and previously transmitted nothing — a rejected deposit was
+  // invisible. Same seeding rule and shared-store rationale as the milestone
+  // effect above; gated on `errors` so it only runs once a poll has resolved.
+  useEffect(() => {
+    if (!errors) return;
+    const terminalEvents = collectDaemonTerminalEvents(
+      depositsToPoll.map((deposit) => deposit.activity.id),
+      errors,
+      getSharedDaemonTerminalTracking(),
+    );
+    for (const terminal of terminalEvents) {
+      logger.event(TELEMETRY_EVENT.ACTIVATION_DAEMON_TERMINAL, {
+        level: "warning",
+        category: "activation",
+        vaultId: shortId(terminal.vaultId),
+        daemonStatus: terminal.daemonStatus,
+      });
+    }
+  }, [errors, depositsToPoll]);
 
   // Optimistic status handlers
   const setOptimisticStatus = useCallback(
--- a/services/vault/src/context/deposit/PeginPollingContext.tsx
+++ b/services/vault/src/context/deposit/PeginPollingContext.tsx
@@ -144,13 +144,13 @@ export function PeginPollingProvider({
 
   // Use the polling query hook
   const {
+    polledIds,
     errors,
     needsWotsKey,
     pendingIngestion,
     pendingDepositorSignatures,
     isLoading,
     refetch,
-    depositsToPoll,
   } = usePeginPollingQuery({
     activities,
     pendingPegins,
@@ -376,11 +376,16 @@ export function PeginPollingProvider({
   // daemon reports a terminal drop (Expired / AmlRejected / ...). These states
   // stop polling and previously transmitted nothing — a rejected deposit was
   // invisible. Same seeding rule and shared-store rationale as the milestone
-  // effect above; gated on `errors` so it only runs once a poll has resolved.
+  // effect above; runs only once a poll has resolved. `polledIds` is captured
+  // inside the queryFn, so ids and errors are always the same poll's snapshot —
+  // pairing the live `depositsToPoll` memo with `errors` instead would, under
+  // `keepPreviousData`, seed new vaults (wallet switch, late-arriving
+  // activities) against a stale map and later emit their pre-existing
+  // terminals as fresh transitions.
   useEffect(() => {
-    if (!errors) return;
+    if (!polledIds || !errors) return;
     const terminalEvents = collectDaemonTerminalEvents(
-      depositsToPoll.map((deposit) => deposit.activity.id),
+      polledIds,
       errors,
       getSharedDaemonTerminalTracking(),
     );
@@ -392,7 +397,7 @@ export function PeginPollingProvider({
         daemonStatus: terminal.daemonStatus,
       });
     }
-  }, [errors, depositsToPoll]);
+  }, [polledIds, errors]);
 
   // Optimistic status handlers
   const setOptimisticStatus = useCallback(
--- a/services/vault/src/context/deposit/daemonTerminalEvents.ts
+++ b/services/vault/src/context/deposit/daemonTerminalEvents.ts
@@ -1,0 +1,97 @@
+/**
+ * Pure transition detector for `activation.daemon.terminal` — a VP daemon
+ * terminal drop (Expired / AmlRejected / IngestionRejected / ...) observed by
+ * the pegin polling loop. These states stop polling and previously transmitted
+ * nothing, so a rejected deposit was invisible to monitoring.
+ *
+ * Same emission discipline as `terminalMilestones`:
+ *  - A vault is SEEDED on the first poll observation that includes it: if it is
+ *    already terminal then (a prior-session drop rediscovered on reload), it is
+ *    marked emitted WITHOUT emitting, so a dashboard load never emits a burst.
+ *  - Thereafter, each distinct terminal daemonStatus emits once per vault —
+ *    keyed per status, so a real Expired → ExpiredCleanedUp progression yields
+ *    both signals.
+ *
+ * Seeding is keyed to the POLLED set, not the errors map: a vault only enters
+ * the errors map once it errors, so its first appearance there would always
+ * look pre-existing and nothing would ever emit.
+ */
+
+import { TerminalPeginPollingError } from "../../utils/peginPolling";
+
+export interface DaemonTerminalTracking {
+  /** Vaults observed in at least one poll result (drives seeding). */
+  seen: Set<string>;
+  /** Emitted `${vaultId}:${daemonStatus}` pairs. */
+  emitted: Set<string>;
+}
+
+export interface DaemonTerminalEvent {
+  /** Raw vaultId; the caller shortens it before it enters event context. */
+  vaultId: string;
+  /** VP daemon status name (e.g. "Expired", "AmlRejected") — safe to emit. */
+  daemonStatus: string;
+}
+
+export function createDaemonTerminalTracking(): DaemonTerminalTracking {
+  return { seen: new Set(), emitted: new Set() };
+}
+
+/**
+ * App-scoped tracking shared by every mounted `PeginPollingProvider`, for the
+ * same reason `terminalMilestones` shares its store: several providers can
+ * observe the same vault at once, and the once-per-transition guarantee must
+ * key to the vault, not to a provider instance. In-memory by design — a reload
+ * resets it so the seeding rule re-suppresses prior-session terminals.
+ */
+const sharedTracking = createDaemonTerminalTracking();
+
+export function getSharedDaemonTerminalTracking(): DaemonTerminalTracking {
+  return sharedTracking;
+}
+
+/**
+ * Test-only reset: the shared store outlives any single provider, so cases
+ * asserting emissions must start from a clean slate.
+ */
+export function resetSharedDaemonTerminalTracking(): void {
+  sharedTracking.seen.clear();
+  sharedTracking.emitted.clear();
+}
+
+/**
+ * Return the daemon-terminal events to emit for this poll, mutating `tracking`
+ * so each (vault, status) pair fires at most once. `polledIds` is the set of
+ * deposits the poll observed; `errors` is the per-deposit error map from the
+ * same poll (non-terminal entries are ignored).
+ */
+export function collectDaemonTerminalEvents(
+  polledIds: readonly string[],
+  errors: ReadonlyMap<string, Error>,
+  tracking: DaemonTerminalTracking,
+): DaemonTerminalEvent[] {
+  const out: DaemonTerminalEvent[] = [];
+
+  for (const vaultId of polledIds) {
+    const error = errors.get(vaultId);
+    const daemonStatus =
+      error instanceof TerminalPeginPollingError ? error.daemonStatus : null;
+
+    if (!tracking.seen.has(vaultId)) {
+      tracking.seen.add(vaultId);
+      // First observation: seed a pre-existing terminal without emitting.
+      if (daemonStatus !== null) {
+        tracking.emitted.add(`${vaultId}:${daemonStatus}`);
+      }
+      continue;
+    }
+
+    if (daemonStatus === null) continue;
+    const key = `${vaultId}:${daemonStatus}`;
+    if (tracking.emitted.has(key)) continue;
+    tracking.emitted.add(key);
+    out.push({ vaultId, daemonStatus });
+  }
+
+  return out;
+}
--- a/services/vault/src/context/deposit/daemonTerminalEvents.ts
+++ b/services/vault/src/context/deposit/daemonTerminalEvents.ts
@@ -15,7 +15,16 @@
  * Seeding is keyed to the POLLED set, not the errors map: a vault only enters
  * the errors map once it errors, so its first appearance there would always
  * look pre-existing and nothing would ever emit.
+ *
+ * A polled id whose errors entry is a plain (non-terminal) Error is a FAILED
+ * observation — provider unreachable, entry omitted or duplicated — so the
+ * poll learned nothing about that vault. It neither seeds nor emits; seeding
+ * waits for a poll that genuinely observes the vault, otherwise a
+ * connectivity blip on the first poll after reload would mis-seed the vault
+ * as healthy and its prior-session terminal would emit as a fresh warning.
  */
+
+import type { DaemonStatus } from "@babylonlabs-io/ts-sdk/tbv/core/clients";
 
 import { TerminalPeginPollingError } from "../../utils/peginPolling";
 
@@ -30,7 +39,7 @@ export interface DaemonTerminalEvent {
   /** Raw vaultId; the caller shortens it before it enters event context. */
   vaultId: string;
   /** VP daemon status name (e.g. "Expired", "AmlRejected") — safe to emit. */
-  daemonStatus: string;
+  daemonStatus: DaemonStatus;
 }
 
 export function createDaemonTerminalTracking(): DaemonTerminalTracking {
@@ -63,7 +72,10 @@ export function resetSharedDaemonTerminalTracking(): void {
  * Return the daemon-terminal events to emit for this poll, mutating `tracking`
  * so each (vault, status) pair fires at most once. `polledIds` is the set of
  * deposits the poll observed; `errors` is the per-deposit error map from the
- * same poll (non-terminal entries are ignored).
+ * SAME resolved poll (non-terminal entries are ignored). Pairing ids and
+ * errors from different polls breaks seeding: a vault new to the id set would
+ * be seeded as healthy against a map that could not contain it, and its
+ * pre-existing terminal would then emit as a fresh transition.
  */
 export function collectDaemonTerminalEvents(
   polledIds: readonly string[],
@@ -76,6 +88,10 @@ export function collectDaemonTerminalEvents(
     const error = errors.get(vaultId);
     const daemonStatus =
       error instanceof TerminalPeginPollingError ? error.daemonStatus : null;
+
+    // Plain error = failed observation: the poll learned nothing about this
+    // vault, so skip it entirely — in particular it must not seed as healthy.
+    if (error !== undefined && daemonStatus === null) continue;
 
     if (!tracking.seen.has(vaultId)) {
       tracking.seen.add(vaultId);
--- a/services/vault/src/hooks/pegoutTerminalEvents.ts
+++ b/services/vault/src/hooks/pegoutTerminalEvents.ts
@@ -1,0 +1,139 @@
+/**
+ * Pure transition detector for the redemption (pegout) terminals observed by
+ * `usePegoutPolling`: `exit.redeem.payout_broadcast` (BTC returned to the
+ * depositor — the redemption success terminal), `exit.redeem.payout_blocked`
+ * (claim/assert on-chain but the payout is blocked), and
+ * `exit.redeem.pegout_timeout` (polling gave up). All three previously
+ * surfaced only as breadcrumbs, which transmit nothing on their own — a stuck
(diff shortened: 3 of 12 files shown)


services/vault/src/context/deposit/PeginPollingContext.tsx
```diff
@@ -367,6 +372,28 @@ export function PeginPollingProvider({
...
 
+  // Emit activation.daemon.terminal once per (vault, daemonStatus) as the VP
+  // daemon reports a terminal drop (Expired / AmlRejected / ...). These states
+  // stop polling and previously transmitted nothing — a rejected deposit was
+  // invisible. Same seeding rule and shared-store rationale as the milestone
+  // effect above; gated on `errors` so it only runs once a poll has resolved.
+  useEffect(() => {
+    if (!errors) return;
+    const terminalEvents = collectDaemonTerminalEvents(
+      depositsToPoll.map((deposit) => deposit.activity.id),
```
jrwbabylonlab: The seeding here can run on mismatched snapshots: depositsToPoll is current-render while errors keeps the previous query key's data under keepPreviousData. On a wallet switch (or any change to the polled id set) the new ids get seeded as healthy against the stale map, so a prior-session terminal then emits as a fresh transition — the exact burst the seeding is meant to suppress, straight into the AML-spike facet. Returning the polled ids inside the query data would keep ids and errors on the same resolved poll. The pegout collector is fine — its universe and classification come from the same map.

services/vault/src/context/deposit/daemonTerminalEvents.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export function getSharedDaemonTerminalTracking(): DaemonTerminalTracking {
+  return sharedTracking;
+}
+
+/**
+ * Test-only reset: the shared store outlives any single provider, so cases
+ * asserting emissions must start from a clean slate.
+ */
+export function resetSharedDaemonTerminalTracking(): void {
```
jrwbabylonlab, line 66: Nothing calls this (the tests use the factories) — same for resetSharedPegoutTerminalTracking. Either drop both or add the provider-mounting integration test à la terminalMilestoneEmission — that's also the test that would have caught the seeding issue in the effect.

services/vault/src/context/deposit/daemonTerminalEvents.ts
```diff
@@ -0,0 +1,97 @@
...
+  seen: Set<string>;
+  /** Emitted `${vaultId}:${daemonStatus}` pairs. */
+  emitted: Set<string>;
+}
+
+export interface DaemonTerminalEvent {
+  /** Raw vaultId; the caller shortens it before it enters event context. */
+  vaultId: string;
+  /** VP daemon status name (e.g. "Expired", "AmlRejected") — safe to emit. */
+  daemonStatus: string;
```
jrwbabylonlab: Could be DaemonStatus rather than string — same for timeoutReason in pegoutTerminalEvents, which is a three-literal union at the source. Keeps the facet values typechecked.

services/vault/src/infrastructure/telemetryEvents.ts
```diff
@@ -23,6 +23,14 @@ export const TELEMETRY_EVENT = {
   ACTIVATION_VERIFIED: "activation.verified",
   /** Vault confirmed ACTIVE on-chain — funnel terminal / primary conversion. */
   DEPOSIT_COMPLETED: "deposit.completed",
+  /** VP daemon reported a terminal drop (Expired / AmlRejected / ...) — the deposit leaves the funnel. */
+  ACTIVATION_DAEMON_TERMINAL: "activation.daemon.terminal",
+  /** Redemption terminal success — the BTC payout tx is broadcast to the depositor. */
+  EXIT_REDEEM_PAYOUT_BROADCAST: "exit.redeem.payout_broadcast",
```
jrwbabylonlab, line 29: File header still says names lead with deposit / activation — worth adding exit while here.

services/vault/src/context/deposit/daemonTerminalEvents.ts
gbarkhatov, line 66: **Dead code (ZERO DEAD CODE POLICY).** `resetSharedDaemonTerminalTracking` is exported but referenced nowhere — the new tests use `createDaemonTerminalTracking()` for isolated instances instead. The milestone precedent's reset earns its keep via `terminalMilestoneEmission.test.ts`, which drives the *shared* store through mounted providers; this PR copied the boilerplate without the equivalent effect-wiring test, so the reset has no consumer. ESLint won't flag an unused export, so CI stays green, but the policy is explicit.

Pick one:
- Delete this export (and its doc comment), or
- Add the integration test that exercises the daemon effect through the shared store and calls this reset in `beforeEach` — the higher-value option, since it also closes the effect-wiring coverage gap.

services/vault/src/hooks/pegoutTerminalEvents.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export function getSharedPegoutTerminalTracking(): PegoutTerminalTracking {
+  return sharedTracking;
+}
+
+/**
+ * Test-only reset: the shared store outlives the hook, so cases asserting
+ * emissions must start from a clean slate.
+ */
+export function resetSharedPegoutTerminalTracking(): void {
```
gbarkhatov, line 73: **Dead code (ZERO DEAD CODE POLICY).** Same as `resetSharedDaemonTerminalTracking`: `resetSharedPegoutTerminalTracking` has no consumer — the tests use `createPegoutTerminalTracking()`. Either delete it, or give it a purpose via an emission-level integration test for `usePegoutPolling` that drives the shared store and resets it in `beforeEach`.

services/vault/src/context/deposit/daemonTerminalEvents.ts
gbarkhatov: Minor: `daemonStatus` is typed `string`, but it's read from `error.daemonStatus`, which is a `DaemonStatus`. Typing the field `DaemonStatus` keeps the classification type-safe end-to-end and matches the tests, which assert against enum members. The downstream event context still gets a string facet, so this is harmless — just a lost type.

services/vault/src/context/deposit/PeginPollingContext.tsx
```diff
@@ -367,6 +372,28 @@ export function PeginPollingProvider({
...
+    );
+    for (const terminal of terminalEvents) {
+      logger.event(TELEMETRY_EVENT.ACTIVATION_DAEMON_TERMINAL, {
+        level: "warning",
+        category: "activation",
+        vaultId: shortId(terminal.vaultId),
+        daemonStatus: terminal.daemonStatus,
+      });
+    }
+  }, [errors, depositsToPoll]);
```
gbarkhatov: Minor: `depositsToPoll` (poll input) and `errors` (poll output) update on different ticks, so this effect can run with a fresh `depositsToPoll` and a stale `errors`. It's safe — an as-yet-unpolled vault just seeds as healthy and emits on a later cycle, and the collector is idempotent across re-runs. Worth a one-line comment noting the two are intentionally not required to be same-cycle, so a future reader doesn't 'fix' it into a bug.

gbarkhatov (dismissed):
Solid, low-risk, idiomatic Phase 4 work — the seed-on-first-observation / emit-on-transition discipline and the per-`(vault, kind)` dedup keying are faithfully mirrored from `terminalMilestones`, and none of it touches a CRITICAL PATH (observability only). `classifyTerminal` correctly checks `timeoutReason` before `claimer.status`, so the `unknown_status` timeout (which also carries a `response`) is classified as a timeout rather than misread. Approving.

Two things to resolve, plus minor nits inline:

**Dead-code + missing effect coverage are one linked issue.** The two `resetShared*` exports (flagged inline) have no consumer, and the reason is that the new events have only pure-collector unit tests — no emission-level integration test like `terminalMilestoneEmission.test.ts`, which mounts real providers and asserts `logger.event` fires through the shared store. Nothing currently pins the effect wiring: the `depositsToPoll.map(d => d.activity.id)` projection, the `if (!errors) return` / `if (!data) return` guards, `shortId`, or the conditional `timeoutReason` spread. The daemon effect's contract is especially worth pinning — it observes the *polled set* but classifies from the *errors* map, so passing the wrong id source would silently never emit. Writing that integration test (using the resets in `beforeEach`) closes both; otherwise delete the reset exports.

**Minor — module placement asymmetry.** The daemon detector lives in `context/deposit/` and the pegout detector in `hooks/`, though the two files are structural twins. It tracks each consumer's location (`PeginPollingContext` vs `usePegoutPolling`), so it's defensible — noting only for symmetry.

Not blocking the approval — the dead-code item is the only policy-backed must-fix.

jonybur (commented):
Two follow-ups on the daemon-terminal collector wiring:

1. **Seeding can pair a live polled set with a stale errors snapshot.** The new effect in `PeginPollingContext` feeds `collectDaemonTerminalEvents` the ids from `depositsToPoll` (a live memo over activities) and `errors` from the query data. `usePeginPollingQuery` keys the query on the joined deposit ids and uses `placeholderData: keepPreviousData`, so when the polled set changes, the effect runs with the new `depositsToPoll` but the previous key's `errors`. Any deposit new to the set gets seeded as seen-and-healthy against a map that could not contain it; when the next poll reports it terminal, the collector emits `activation.daemon.terminal` for a pre-existing terminal - exactly the prior-session burst the seeding is meant to suppress. Concrete triggers: a btcPublicKey/wallet switch (the whole new set seeded against the old wallet's map) or activities arriving after the first poll resolves. Fix: return the polled-id snapshot from the queryFn alongside `errors` so the two are always from the same poll, and feed that snapshot to the collector instead of the live memo.

2. **Unused test-only resets.** `resetSharedDaemonTerminalTracking` and `resetSharedPegoutTerminalTracking` are exported but nothing imports them - both new test files build fresh tracking via `create*Tracking()`. Per the zero-dead-code policy they should either be used or dropped.

services/vault/src/context/deposit/PeginPollingContext.tsx
jrwbabylonlab: `depositsToPoll` and `errors` aren't atomic: the query uses `placeholderData: keepPreviousData`, so when the poll set changes (account switch, mid-session growth) this effect runs the new ids against the previous set's errors map. Those vaults seed as healthy, and when the real poll resolves, prior-session terminals emit as fresh warnings — the burst seeding is meant to suppress. Suggest having the queryFn return the ids it actually polled in `PollingQueryData` and collecting from `data.polledIds` + `data.errors`, atomic like the pegout side.

services/vault/src/context/deposit/daemonTerminalEvents.ts
jrwbabylonlab, line 66: Unused — the tests build local tracking via `createDaemonTerminalTracking()` (same for `resetSharedPegoutTerminalTracking`). Either add an effect-level emission test that needs them (like Phase 3's `terminalMilestoneEmission.test.tsx` — it would also cover the effect wiring) or drop both.

services/vault/src/hooks/pegoutTerminalEvents.ts
```diff
@@ -0,0 +1,139 @@
...
+  emitted: Set<string>;
+}
+
+export interface PegoutTerminalEvent {
+  event: TelemetryEvent;
+  level: "info" | "warning";
+  /** Raw vaultId; the caller shortens it before it enters event context. */
+  vaultId: string;
+  /** Only set for pegout_timeout — which give-up path fired. */
+  timeoutReason?: string;
```
jrwbabylonlab: Nit: `PegoutPollingResult["timeoutReason"]` would keep the union; likewise `DaemonTerminalEvent.daemonStatus` could be `DaemonStatus`.

services/vault/src/hooks/pegoutTerminalEvents.ts
```diff
@@ -0,0 +1,139 @@
...
+  level: "info" | "warning";
+  timeoutReason?: string;
+}
+
+function classifyTerminal(
+  result: PegoutPollingResult,
+): ClassifiedTerminal | null {
+  if (result.timeoutReason !== undefined) {
+    return {
+      kind: "timeout",
```
jrwbabylonlab: Dedup keys on kind alone, so a vault that times out, recovers, then times out again for a different reason emits nothing the second time. Intentional?

services/vault/src/infrastructure/telemetryEvents.ts
jrwbabylonlab, line 29: Nit: the header docblock still says names lead with `deposit` / `activation` — worth adding `exit`.

services/vault/src/context/deposit/daemonTerminalEvents.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (!tracking.seen.has(vaultId)) {
+      tracking.seen.add(vaultId);
+      // First observation: seed a pre-existing terminal without emitting.
+      if (daemonStatus !== null) {
+        tracking.emitted.add(`${vaultId}:${daemonStatus}`);
+      }
+      continue;
+    }
+
```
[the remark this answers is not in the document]
jonybur: By design - at first observation, a same-session first-poll terminal is indistinguishable from a prior-session terminal rediscovered on reload, and suppressing the reload burst wins that tradeoff (documented in the module header and the PR description; a burst of stale warnings would poison the `daemonStatus` facet). In practice the scenario described here is the reload case: a deposit created in-session enters the polled set while the daemon is still ingesting, so it is observed healthy first and any later rejection is a transition, which emits. 728ec508 tightens seeding further: failed observations (provider unreachable / omitted / duplicate entry) no longer seed, so suppression now applies only to genuinely observed statuses.

services/vault/src/hooks/pegoutTerminalEvents.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    if (!tracking.seen.has(vaultId)) {
+      tracking.seen.add(vaultId);
+      // First observation: seed a pre-existing terminal without emitting.
+      if (terminal) tracking.emitted.add(`${vaultId}:${terminal.kind}`);
+      continue;
+    }
+
+    if (!terminal) continue;
+    const key = `${vaultId}:${terminal.kind}`;
```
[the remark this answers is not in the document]
jonybur: By design - same seeding tradeoff as the daemon side. A redemption initiated in-session is observed in progress before it can reach `PayoutBroadcast`/`PayoutBlocked`, so its terminal is a transition and emits; and a timed-out first observation cannot occur in-session because the give-up counters need many cycles while seeding happens on the first collected poll. First-observation-already-terminal is in practice the reload case, which seeding is designed to suppress. 728ec508 additionally excludes failure-shaped results (no `response`, no `timeoutReason`) from seeding, so a provider blip on the first poll after reload no longer mis-seeds the vault as non-terminal.

services/vault/src/context/deposit/PeginPollingContext.tsx
jonybur: Fixed in 728ec508 exactly as suggested: the queryFn snapshots the ids it actually polled into `PollingQueryData.polledIds`, and the effect collects from `data.polledIds` + `data.errors` - both always from the same resolved poll, atomic like the pegout side. The hook's returned `depositsToPoll` lost its last consumer in the process and was removed.

Writing the emission test for this surfaced a second route to the same burst, with zero snapshot tearing: a vault whose first observation is a failed poll (plain connectivity `Error` in the same snapshot - unreachable provider, omitted or duplicate entry) was seeded as "seen healthy" even though the poll learned nothing, so its prior-session terminal emitted as fresh on the next successful poll. Failed observations now don't seed at all; unit tests cover both routes, and `daemonTerminalEmission.test.tsx` pins the id source (its mocked hook result carries no id source other than `polledIds`).

services/vault/src/context/deposit/daemonTerminalEvents.ts
jonybur: Took the integration-test option in 728ec508: `daemonTerminalEmission.test.tsx` mounts the real provider (including the co-mounted-providers case) and `pegoutTerminalEmission.test.tsx` drives the real `usePegoutPolling` through react-query with only the SDK batch poller mocked. Both consume the resets in `beforeEach` and pin the effect wiring: the `polledIds` id source, the resolved-poll guards, `shortId`, level/category, and that `timeoutReason` stays absent on non-timeout payloads. The daemon one also would have caught the seeding issue in the effect.

services/vault/src/context/deposit/daemonTerminalEvents.ts
jonybur: Done in 728ec508 - `DaemonStatus` on the daemon event, `PegoutPollingResult["timeoutReason"]` on both the classified terminal and the emitted event.

services/vault/src/infrastructure/telemetryEvents.ts
jonybur: Added in 728ec508.

services/vault/src/context/deposit/daemonTerminalEvents.ts
jonybur: Took the integration-test option in 728ec508: `daemonTerminalEmission.test.tsx` mounts the real provider (including the co-mounted-providers case) and `pegoutTerminalEmission.test.tsx` drives the real `usePegoutPolling` through react-query with only the SDK batch poller mocked. Both consume the resets in `beforeEach` and pin the effect wiring: the `polledIds` id source, the resolved-poll guards, `shortId`, level/category, and that `timeoutReason` stays absent on non-timeout payloads. The daemon one also would have caught the seeding issue in the effect.

services/vault/src/hooks/pegoutTerminalEvents.ts
jonybur: Done in 728ec508 - `pegoutTerminalEmission.test.tsx` drives the real `usePegoutPolling` through react-query (only the SDK batch poller mocked), consumes this reset in `beforeEach`, and covers the broadcast terminal, no-burst seeding, and the real ten-failure timeout path with its reason facet.

services/vault/src/context/deposit/daemonTerminalEvents.ts
jonybur: Done in 728ec508 - `DaemonStatus` on the daemon event, `PegoutPollingResult["timeoutReason"]` on both the classified terminal and the emitted event.

services/vault/src/context/deposit/PeginPollingContext.tsx
jonybur: Went further than a comment - the sibling thread showed the non-atomic pairing wasn't benign: under `keepPreviousData` a poll-set change runs the new ids against the previous key's errors map, so a vault seeded as healthy there emits its prior-session terminal as a fresh warning once the real poll resolves. 728ec508 makes the pairing atomic (the queryFn returns `polledIds` inside `PollingQueryData`; the effect reads ids and errors from the same `data`), and the effect comment now documents why the live memo must not be used.

services/vault/src/context/deposit/PeginPollingContext.tsx
jonybur: Fixed in 728ec508 exactly as suggested: the queryFn snapshots the ids it actually polled into `PollingQueryData.polledIds`, and the effect collects from `data.polledIds` + `data.errors` - both always from the same resolved poll, atomic like the pegout side. The hook's returned `depositsToPoll` lost its last consumer in the process and was removed.

Writing the emission test for this surfaced a second route to the same burst, with zero snapshot tearing: a vault whose first observation is a failed poll (plain connectivity `Error` in the same snapshot - unreachable provider, omitted or duplicate entry) was seeded as "seen healthy" even though the poll learned nothing, so its prior-session terminal emitted as fresh on the next successful poll. Failed observations now don't seed at all; unit tests cover both routes, and `daemonTerminalEmission.test.tsx` pins the id source (its mocked hook result carries no id source other than `polledIds`).

services/vault/src/context/deposit/daemonTerminalEvents.ts
jonybur: Took the integration-test option in 728ec508: `daemonTerminalEmission.test.tsx` mounts the real provider (including the co-mounted-providers case) and `pegoutTerminalEmission.test.tsx` drives the real `usePegoutPolling` through react-query with only the SDK batch poller mocked. Both consume the resets in `beforeEach` and pin the effect wiring: the `polledIds` id source, the resolved-poll guards, `shortId`, level/category, and that `timeoutReason` stays absent on non-timeout payloads. The daemon one also would have caught the seeding issue in the effect.

services/vault/src/hooks/pegoutTerminalEvents.ts
jonybur: Done in 728ec508 - `DaemonStatus` on the daemon event, `PegoutPollingResult["timeoutReason"]` on both the classified terminal and the emitted event.

services/vault/src/hooks/pegoutTerminalEvents.ts
jonybur: It wasn't intentional - good catch. 728ec508 keys timeouts per give-up reason (`timeout:<reason>`), mirroring the daemon side's per-status keying: a vault that times out, recovers, then gives up down a different path emits the new reason, while a repeat of an already-emitted reason stays deduped (bounded at three timeout emissions per vault). Unit test added.

services/vault/src/infrastructure/telemetryEvents.ts
jonybur: Added in 728ec508.

gbarkhatov (approved):

Issue #2098, opened 2026-07-21
[v3] Loans: borrow-market asset-detail page (Borrow APY & Utilization charts)


gbarkhatov:
## Summary

The v3 Loans page (see #2042) implements the loans page itself: the borrow-capacity / health-factor summary, the Active Loans list, and borrow/repay through the existing reserve-detail overlay. It does NOT implement the full borrow-market / asset-detail experience shown in the design — a dedicated per-asset page with historical **Borrow APY** and **Utilization** line charts, a date-range selector, and asset stats (APY / available liquidity / price).

This was deliberately left out of the raw v3 rollout because it does not exist in the current v2 UI and is a substantial new build (new data source + charts), not a reuse of existing functionality. Filing it as a separate ticket so the v3 Loans page could ship behind the feature flag.

## Design

[TBV v.3 Premium — borrow market / asset detail](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=735-48726)

## Scope

- A per-asset borrow-market detail surface (route or overlay) reachable from the Loans page / asset rows.
- Historical **Borrow APY** chart and **Utilization** chart with an all-time / date-range selector.
- Asset stats section (APY, available liquidity, price).
- A time-series data source for APY/utilization history (the current hooks only expose the live point-in-time values via the Aave Hub — `useAaveBorrowAprs`, `useAaveReserveLiquidity`).

## Out of scope

- The borrow and repay actions themselves — those already work today via the existing reserve-detail overlay (`services/vault/src/applications/aave/components/Detail`) and are not blocked by this ticket.

## Notes

- Live (non-historical) borrow APR and liquidity/utilization already come from `services/vault/src/applications/aave/hooks/useAaveBorrowAprs.ts` and `useAaveReserveLiquidity.ts`; the charts additionally need historical series which we do not fetch yet.

Issue #2097, opened 2026-07-21
[vault] Loans: active-loan amounts show full token precision (e.g. 1.000002 USDC)


gbarkhatov:
## Summary

In the Loans "Active Loans" list, each borrowed amount is shown at the token's full precision, so accrued interest surfaces as trailing digits — e.g. `1.000002 USDC`, `5.000002 USDT`, `0.00705306 WBTC` — while the "Total borrowed" summary shows the rounded USD value (`$1.00`). The value is correct (it is the real on-chain debt), but the extra digits look noisy and the exact-vs-rounded mismatch between the row and the summary can confuse.

This is pre-existing behavior: it shows the same way in the current v2 Loans section and in the new v3 Loans page (both read the same data). This ticket is to decide whether we want to clean up the display, not to fix a correctness bug.

## Why it happens (implementation detail)

The amount comes from `useAaveBorrowedAssets` → `transformToBorrowedAsset` in `services/vault/src/applications/aave/hooks/useAaveBorrowedAssets.ts`. It formats the on-chain debt (`debtPosition.totalDebt`, which is principal + continuously-accrued variable-borrow interest) with the token's own decimals:

`amount = formatAmount(Number(formatUnits(debtPosition.totalDebt, reserve.token.decimals)), reserve.token.decimals)`

USDC/USDT have 6 decimals, so up to 6 fractional digits render, and Aave debt ticks up every block, so the last digits keep changing. The summary's "Total borrowed" instead uses `formatUsdValue(debtValueUsd)` (2 decimals), which rounds the sub-cent accrual away — hence `$1.00` next to `1.000002 USDC`.

Consumers of `BorrowedAsset.amount`:
- v3 row: `services/vault/src/components/simple/ActiveLoansList.tsx`
- v2 section: `services/vault/src/components/simple/LoansSection.tsx`

## Decision needed

Do we want to reduce the displayed precision for a cleaner look, or keep the exact debt?

- Keep exact (today): most transparent for repayment; noisy trailing digits.
- Round for display: cleaner, closer to the Figma mocks (which use round numbers), but slightly hides that the debt is fractionally higher than shown. A flat 2-decimal cap would under-round BTC-like tokens (WBTC), so it needs per-token precision (stablecoins ~2, WBTC more), and the row would need the raw numeric debt threaded through rather than the preformatted string.

## Where to change

[...]

PR #2100, opened 2026-07-21
feat(vault): v3 loans page

### LINKED ISSUE #2042: [v3] Loans Page
A dedicated Loans page extracted from the current dashboard while preserving the existing Aave borrow and repay flows.

**Every v3 surface ships behind the feature flag in #2014** so `main` remains deployable with the current v2 dashboard throughout the build.

**Design:** [Loans before deposit](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10044-13820) · [Loans after deposit](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10042-3403) · [Loans borrow page](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10088-48689)

**Area:** `services/vault` (`LoansSection`, Aave reserve-detail overlay, `AssetSelectionModal`)

## Scope

- Move the current `LoansSection` to `/loans`
- Preserve borrowed-asset presentation and borrow/repay entry points
- Reuse the existing reserve-detail overlay and asset-selection modal
- Retain the two-column reserve-card layout and live per-reserve borrow APR
- Align loan rows and lifecycle statuses with the v3 design

## Acceptance criteria

- [ ] `/loans` renders the before-deposit, after-deposit, and active-borrow states from Figma
- [ ] Borrow and repay continue through the existing reserve-detail flow
- [ ] Live APR and reserve data remain sourced from the current implementation
- [ ] Loan rows and statuses render correctly in light and dark
- [ ] The Loans section is no longer duplicated on the Overview page

## Depends on

- #2017

gbarkhatov:
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2042

Screenshot_2026-07-21_19-00-28

---

## What this does

Adds the new "Loans" page for the v3 (premium) UI at `/loans`, behind the existing `NEXT_PUBLIC_FF_ENABLE_V3_UI` feature flag. With the flag off nothing changes (the page redirects to the v2 dashboard); with it on, the sidebar shows a dedicated Loans page.

The page shows, for a connected depositor with collateral:
- A summary card with **Available to Borrow**, **Total Borrowed**, and **Health Factor**.
- An **Active Loans** list (one row per borrowed asset) with the live Borrow APR, available liquidity, and utilization, plus per-asset Borrow / Repay buttons.
- An empty state when there are no active loans yet.

Borrowing and repaying happen through the **exact same flow the v2 app already uses** — the asset-selection modal and the reserve-detail overlay — so no borrow/repay logic was rewritten.

## Why

v3 is a UI restructure that splits the old single-page dashboard into separate pages. The Loans section needed its own page. The guiding rule for this rollout (the flag is being removed shortly) was: **reuse what already works, ship it raw, and file follow-up tickets for anything that would be a big new build.**

## Approach and why we chose it

**Reuse the existing borrow/repay engine instead of rebuilding it.** The `/loans` route already sits under the layout that renders the Aave reserve-detail overlay (driven by `?reserve=&tab=` in the URL), so we mounted the real page there and pointed the Borrow/Repay buttons at the existing asset-selection modal and overlay. The alternative — rebuilding the borrow/repay UI for v3 — was rejected: it would duplicate value-moving code, add risk, and take far longer than the flag timeline allows.

**Show the summary and a Borrow entry whenever the user has collateral (matches v2).** In v2 the Loans section always offered a Borrow button once you had a pegin. Our first version gated the whole page on "already has debt", which meant a depositor with collateral but no loan saw only a "Deposit" prompt and could not borrow. We fixed this so the page keys off "has a position" (collateral or loans), restoring the v2 behavior.

**Disable Borrow when there is effectively no capacity left.** A fully-borrowed position leaves a sub-cent dust residual, so a naive `available > 0` check kept the Borrow buttons enabled at "$0.00 available". We added a small named threshold (`MIN_BORROWABLE_USD`) and compute a single `canBorrow` in one place (`useDashboardState`) that both the Loans page and the Overview page use, so the Borrow buttons disable when the position is maxed out — matching the design.

**Reuse and extend shared UI instead of one-off components.** The summary reuses the same stat-card component the v3 Overview uses (we extracted the two shared "borrow capacity" cards so both pages build them the same way and added a no-button health-factor variant). The empty state reuses the shared `EmptyState` component, which we improved (optional custom icon, a single card surface, and the brand-orange primary button) — this also cleans up the one other place that component is used.

**Small shared hooks for the wiring.** `useLoanActions` holds the borrow/repay button orchestration (shared with the dashboard so it is not duplicated) and `useActiveLoans` merges each borrowed asset with its live liquidity/utilization.

## Not included (follow-up tickets)

- **Borrow-market asset-detail page with historical APY / utilization charts** — the per-asset "borrow market" screens with graphs and a date selector. This does not exist in v2 and is a substantial new build, so it was left out of this raw rollout. Tracked in https://github.com/babylonlabs-io/babylon-toolkit/issues/2098 <- cc @jeremy-babylonlabs you need to think who should take this ticket
- **Loan amount precision** — active-loan amounts show the exact on-chain debt at full token precision (e.g. `1.000002 USDC`) while the summary shows rounded USD. This is pre-existing behavior in both v2 and v3; whether to round the display is tracked (with implementation notes) in https://github.com/babylonlabs-io/babylon-toolkit/issues/2097 <- @jeremy-babylonlabs this is present in v2

## Testing

- `pnpm --filter vault run lint` — clean.
- `pnpm --filter vault run test` — full suite passes (added tests for the empty state, the `canBorrow` threshold, and the `/loans` route).
- `tsc --noEmit` — clean.
- Manual, flag on: empty / connected-no-loans / active-loans states; borrow and repay through the reserve-detail overlay; Borrow disabled when fully borrowed; light and dark themes. Flag off: v2 dashboard unchanged, `/loans` redirects.

Note for local runs: rebuild `core-ui` first (`pnpm --filter @babylonlabs-io/core-ui build`) so the built package is current.

### DIFF
--- a/services/vault/src/components/pages/Loans.tsx
+++ b/services/vault/src/components/pages/Loans.tsx
@@ -1,0 +1,121 @@
+/**
+ * Loans page (v3)
+ * Dedicated `/loans` route: borrow-capacity + health-factor summary and the
+ * list of active loans, reusing the existing Aave data hooks and the
+ * reserve-detail borrow/repay overlay (route-driven via `?reserve=&tab=`).
+ */
+
+import { Container } from "@babylonlabs-io/core-ui";
+import { useOutletContext } from "react-router";
+
+import { AssetSelectionModal } from "@/applications/aave/components/AssetSelectionModal";
+import { LOAN_TAB } from "@/applications/aave/constants";
+import { useActiveLoans } from "@/applications/aave/hooks";
+import type { RootLayoutContext } from "@/components/pages/RootLayout";
+import { EmptyState, EmptyStateIcon } from "@/components/shared";
+import { PAGE_CONTENT_CLASS } from "@/components/shared/layoutClasses";
+import { useConnection, useETHWallet } from "@/context/wallet";
+import { COPY } from "@/copy";
+import { useDashboardState } from "@/hooks/useDashboardState";
+import { useLoanActions } from "@/hooks/useLoanActions";
+import { formatUsdValue } from "@/utils/formatting";
+
+import { ActiveLoansList } from "../simple/ActiveLoansList";
+import { LoansSummary } from "../simple/LoansSummary";
+
+export default function Loans() {
+  const { openDeposit } = useOutletContext<RootLayoutContext>();
+  const { address } = useETHWallet();
+  const { isConnected } = useConnection();
+
+  const {
+    debtValueUsd,
+    maxTotalDebtUsd,
+    availableToBorrowUsd,
+    canBorrow,
+    healthFactor,
+    healthFactorStatus,
+    borrowedAssets,
+    hasLoans,
+    hasCollateral,
+    selectableBorrowedAssets,
+    isBorrowCapacityLoading,
+    borrowCapacityError,
+  } = useDashboardState(isConnected ? address : undefined);
+
+  const { openBorrowPicker, openRepay, goToReserve, assetModalProps } =
+    useLoanActions({ borrowedAssets, selectableBorrowedAssets });
+
+  const activeLoans = useActiveLoans(borrowedAssets);
+
+  // A borrow position exists once there is collateral to borrow against (or an
+  // active loan). With collateral but no debt yet, the summary still renders so
+  // the depositor can borrow via its Borrow action — mirroring the v2 Loans
+  // section, whose Borrow button was enabled whenever collateral was present.
+  const hasPosition = hasCollateral || hasLoans;
+
+  // Nothing to borrow against yet: prompt to deposit collateral first (or to
+  // connect a wallet when disconnected).
+  if (!isConnected || !hasPosition) {
+    return (
+      <Container className={`${PAGE_CONTENT_CLASS} pb-6`}>
+        <EmptyState
+          icon={<EmptyStateIcon />}
+          title={COPY.loans.noActiveLoans.title}
+          description={COPY.loans.noActiveLoans.body}
+          isConnected={isConnected}
+          actionLabel={COPY.overview.depositAction}
+          onAction={() => openDeposit()}
+          withCard
+        />
+      </Container>
+    );
+  }
+
+  const availableMeterPercent =
+    maxTotalDebtUsd > 0 ? availableToBorrowUsd / maxTotalDebtUsd : 0;
+  const borrowedMeterPercent =
+    maxTotalDebtUsd > 0 ? debtValueUsd / maxTotalDebtUsd : 0;
+
+  return (
+    <Container className={`${PAGE_CONTENT_CLASS} pb-6`}>
+      <div className="space-y-6">
+        <LoansSummary
+          availableToBorrow={formatUsdValue(availableToBorrowUsd)}
+          availableMeterPercent={availableMeterPercent}
+          totalBorrowed={formatUsdValue(debtValueUsd)}
+          borrowedMeterPercent={borrowedMeterPercent}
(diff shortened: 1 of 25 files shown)


jrwbabylonlab (dismissed):

services/vault/src/components/pages/Loans.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  // A borrow position exists once there is collateral to borrow against (or an
+  // active loan). With collateral but no debt yet, the summary still renders so
+  // the depositor can borrow via its Borrow action — mirroring the v2 Loans
+  // section, whose Borrow button was enabled whenever collateral was present.
+  const hasPosition = hasCollateral || hasLoans;
+
+  // Nothing to borrow against yet: prompt to deposit collateral first (or to
+  // connect a wallet when disconnected).
+  if (!isConnected || !hasPosition) {
```
jrwbabylonlab, line 73: isLoading isn't consulted here, so a depositor with a position sees the full-page empty state + deposit CTA flash on hard refresh before the summary lands. v2 has the same flicker, but as a whole-page state it's much more visible.

services/vault/src/components/simple/ActiveLoansList.tsx
```diff
@@ -0,0 +1,133 @@
...
+                <Metric
+                  label={COPY.loans.utilizationLabel}
+                  value={utilization}
+                />
+              </div>
+
+              <div className="flex flex-shrink-0 gap-3">
+                <RowActionButton
+                  label={COPY.loans.borrowButton}
+                  onClick={() => onBorrow(row.symbol)}
```
jrwbabylonlab, line 119: borrowedAssets derives from allBorrowReserves, so debt in a frozen/paused reserve still gets a Borrow button — a path the v2 asset picker never offers. It fails closed in the detail form, so just a soft dead-end, but worth hiding or disabling for non-borrowable reserves.

services/vault/src/hooks/useLoanActions.ts
```diff
@@ -0,0 +1,96 @@
...
+import { getReserveDetailRoute } from "@/routes";
+
+interface UseLoanActionsProps {
+  /** Borrowed assets (used to resolve the single-asset repay shortcut). */
+  borrowedAssets: { symbol: string }[];
+  /** Borrowed assets shaped for the repay-mode asset picker. */
+  selectableBorrowedAssets: Asset[];
+}
+
+export interface AssetSelectionModalProps {
```
jrwbabylonlab: The export has no consumer and hand-duplicates the modal's own props shape — keep it local (or type against the modal's props) so they can't drift.

jeremy-babylonlabs (dismissed):

services/vault/src/components/pages/Loans.tsx
gbarkhatov: done

services/vault/src/hooks/useLoanActions.ts
gbarkhatov: done

services/vault/src/components/simple/ActiveLoansList.tsx
gbarkhatov: done

jrwbabylonlab (approved):

PR #2118, opened 2026-07-22
feat(vault): liquidation analysis card on the v3 overview

jeremy-babylonlabs:
Adds the Liquidation Analysis section to the v3 overview


services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -367,6 +399,16 @@ export function DashboardPage() {
           />
         )}
 
+        {featureFlags.isV3UiEnabled && (
+          <LiquidationAnalysisSection
```
jrwbabylonlab, line 407: This renders the card whenever `isV3UiEnabled`, and its `position` state draws a hardcoded fixture (`FIXTURE_CASCADE`, 0.6/0.4/0.1 BTC @ $88,400) as if it were the user's real liquidation analysis — a connected user with collateral + a loan sees fabricated liquidation prices with no "sample" labeling. Since there's no real data yet, the chart should sit behind its own feature flag (default off) so it can't display on devnet/testnet until the live-data follow-up lands — gating on `isV3UiEnabled` alone isn't enough. The empty states are fine (they read the live position); it's specifically the fixture-backed chart that needs to stay hidden.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -0,0 +1,167 @@
...
+  onBorrow,
+  cascadeOverride,
+}: LiquidationAnalysisSectionProps) {
+  const navigate = useNavigate();
+
+  const btcPrice = cascadeOverride?.btcPrice ?? FIXTURE_BTC_PRICE;
+
+  const { bands, priceAxis, shareAxisTicks } = useMemo(
+    () =>
+      buildLiquidationChartData(cascadeOverride?.result ?? FIXTURE_CASCADE, {
```
jrwbabylonlab: The fixture fallback (`?? FIXTURE_CASCADE`) is what surfaces fake data to real users once the new flag above is on but live data isn't wired. When you add the gate, prefer making live/override data required to render the chart rather than silently falling back to the fixture in production — that keeps the placeholder strictly a dev/QA path.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -0,0 +1,167 @@
...
+            <span className="text-base leading-[1.5] tracking-[0.15px] text-accent-primary">
+              {COPY.liquidations.simulateLabel}
+            </span>
+            <span className="text-sm leading-[1.43] tracking-[0.17px] text-accent-secondary">
+              {COPY.liquidations.simulateDescription}
+            </span>
+          </div>
+          <button
+            type="button"
+            onClick={() => navigate(ROUTES.LIQUIDATIONS)}
```
jrwbabylonlab: `ROUTES.LIQUIDATIONS` is still `<V3Placeholder/>` (`router.tsx:133`), so this "Explore" CTA is a dead-end today. Fine mid-build, but worth gating or disabling it until the destination page is real so users don't land on a placeholder.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -0,0 +1,167 @@
...
+ * panel's manual mode overrides it in the meantime.
+ */
+
+/** core-ui's `Button` is still on the pre-v3 spec. */
+const BUTTON_CLASS =
+  "flex h-10 w-[120px] shrink-0 items-center justify-center rounded-lg text-base leading-[1.5] tracking-[0.15px] transition-[filter] hover:brightness-110";
+const PRIMARY_BUTTON_CLASS = `${BUTTON_CLASS} bg-secondary-main text-accent-contrast`;
+const NEUTRAL_BUTTON_CLASS = `${BUTTON_CLASS} bg-secondary-strokeLight text-accent-primary`;
+
+function EmptyState({
```
jrwbabylonlab, line 40: Nit: this local `EmptyState` duplicates `@/components/shared/EmptyState`, which already serves the other v3 pages. Different shape (no icon/Connect), so defensible, but worth reusing the shared one or noting why not.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -0,0 +1,167 @@
...
+ * collateral without a loan, or the seizure map. There is no simulator here —
+ * Explore hands off to the full page for that.
+ *
+ * The cascade is a fixture; the follow-up PR swaps it for the live
+ * `usePositionNotifications` result DashboardPage already holds. The god-mode
+ * panel's manual mode overrides it in the meantime.
+ */
+
+/** core-ui's `Button` is still on the pre-v3 spec. */
+const BUTTON_CLASS =
```
jrwbabylonlab: Nit: these v3 button classes duplicate the pattern in `components/vaults/buttonClasses.ts` from the earlier v3 work — consider consolidating the shared v3 button styles into one module.

gbarkhatov: Nit (non-blocking): `BUTTON_CLASS` re-implements the button spec inline because core-ui's `Button` is still pre-v3 — justified as the comment says, but this pre-v3 button spec is now duplicated across sections. A shared v3 button primitive would be a good follow-up so the copy-paste doesn't drift.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -0,0 +1,167 @@
...
+  return (
+    <div className="flex min-h-[312px] flex-col items-center justify-center gap-1 text-center">
+      <p className="text-xl leading-[1.6] tracking-[0.15px] text-accent-primary">
+        {title}
+      </p>
+      <p className="max-w-[600px] text-sm leading-[1.43] tracking-[0.17px] text-accent-secondary">
+        {description}
+      </p>
+      {/* Wrapped: passing by reference hands `openDeposit` the click event. */}
+      <button
```
gbarkhatov, line 60: Heads-up for the `e2e/real` owner (non-blocking): this deposit/borrow button is a new wallet-action entry surface but carries no `data-testid`. Per CLAUDE.md's E2E-hook rule, if the real-wallet flow ever needs to drive deposit/borrow from this overview card it'll want a testid here. Fine to defer — just flagging it.

packages/babylon-core-ui/src/components/LiquidationChart/PriceFrame.tsx
```diff
@@ -222,7 +246,27 @@ export function PriceFrame({
         </div>
       </div>
 
-      {xAxisLabels?.length ? (
+      {xAxisTicks?.length ? (
+        <div
+          className="bbn-liq-chart__xaxis bbn-liq-chart__xaxis--positioned"
+          style={plotInsetLeft > 0 ? { paddingLeft: pct(plotInsetLeft) } : undefined}
```
gbarkhatov: Latent, not triggered today (SeizureMap passes `plotInsetLeft: 0`): the tick spans in this branch are `position: absolute; left: <fraction>%`, and an absolutely-positioned child resolves `left%` from its containing block's padding edge. So `paddingLeft` here doesn't translate the ticks inward — it only widens the padding box, slightly rescaling the percentage. If this branch is ever reused with a non-zero inset, the ticks won't align with the plot. The evenly-spaced `xAxisLabels` branch below is normal flow, so it does honor the padding — hence the difference.

gbarkhatov (dismissed):
Approving — solid, well-tested UI-only PR that stays clear of every CRITICAL PATH (WASM / fees / presigning / vault-secret derivation / activation). The compressed-share and axis-floor math checks out against the tests, and all referenced copy keys resolve.

A few non-blocking notes are inline, plus two general ones:

- `PAGE_CONTENT_CLASS` is now runtime-branched on `isV3UiEnabled`, and the v3 value stacks `mx-10` — worth confirming no other v3 page still applies `mx-auto` on top of it (last margin utility wins). This PR correctly drops it from Activity.tsx.
- `fixtures.ts` is a clearly-marked temporary UI seam (#2043); please ensure the follow-up data PR deletes it to honor the zero-dead-code policy.

On the card's `dark:bg-[#202020]`: I checked it against the sibling cards and it matches the existing RiskSection bordered-card pattern, so I'm not flagging it as an inconsistency here — a systemic token cleanup could revisit both.

jonybur (approved):

PR #2119, opened 2026-07-22
fix(vault): make the Aave repay flow tolerant of lagging RPC backends

jrwbabylonlab:
1. **Approvals are verified from the approve tx's own receipt** (its
 `Approval` event, last matching log) instead of a follow-up allowance
 read that can hit a lagging backend. A bounded re-read remains as
 fallback for receipts without a matching event (wallet cancel/speed-up
 replacements, non-emitting tokens). A wallet-edited spending cap below
 the required amount gets its own accurate error.
2. **Repay retries only provably pre-broadcast failures.** Errors raised at
 pre-flight simulation are tagged; a simulation-phase
 `ERC20InsufficientAllowance` gets up to three attempts with catch-up
 delays on every path. When the approve was short-circuited by the
 allowance pre-check, the second failure forces a fresh approve before
 the last attempt. Mined reverts never auto-retry (no second wallet
 prompt, no double gas), and each absorbed retry logs a warning so
 rescued flows stay visible in monitoring.
3. **`approveERC20` is attempt-first**: single prompt for standard tokens.
 The `approve(0)` reset (shown by wallets as "revoke approval") fires
 only when the direct approve reverts at pre-broadcast simulation AND
 the current allowance is non-zero — the USDT zero-first signature.
 User rejections, network errors, and mined reverts rethrow immediately,
 and if the gate's own allowance read fails, the original revert
 surfaces rather than the read error.
4. Approval-verification errors are humanized (token units, not raw wei)
 and live in `copy.ts`, worded to survive the error-message rewriter.

### DIFF
--- a/services/vault/src/applications/aave/services/positionTransactions.ts
+++ b/services/vault/src/applications/aave/services/positionTransactions.ts
@@ -16,6 +16,7 @@ import type {
 import { formatUnits, isAddressEqual, maxUint256, parseEventLogs } from "viem";
 
 import { COPY } from "@/copy";
+import { logger } from "@/infrastructure";
 
 import { ERC20 } from "../../../clients/eth-contract";
 import { ErrorCode, isSimulationPhaseError } from "../../../utils/errors";
@@ -253,11 +254,13 @@ function isStaleAllowanceSimulationError(err: unknown): boolean {
 }
 
 /**
- * Run the repay, absorbing stale-backend allowance reverts at simulation: one
- * plain delayed retry first; if the approve was short-circuited (a stale HIGH
- * allowance read can skip a genuinely needed approve), force the approve
- * before the final attempt. Mined reverts and all other errors propagate
- * untouched — auto-retrying a broadcast tx would re-prompt the wallet.
+ * Run the repay, absorbing stale-backend allowance reverts at simulation:
+ * three attempts with catch-up delays on every path. When the approve was
+ * short-circuited (a stale HIGH allowance read can skip a genuinely needed
+ * approve), the second failure forces the approve before the last attempt; a
+ * receipt-verified approve can't be improved on, so that path just
+ * re-simulates. Mined reverts and all other errors propagate untouched —
+ * auto-retrying a broadcast tx would re-prompt the wallet.
  * Note: repayMaxCapped can revert here legitimately (accrued interest pushed
  * debt past the approved balance cap); the retries just delay that error.
  */
@@ -272,18 +275,21 @@ async function repayWithStaleSimulationRetry(params: {
     return await execute();
   } catch (firstError) {
     if (!isStaleAllowanceSimulationError(firstError)) throw firstError;
+    logger.warn("Repay simulation saw a stale allowance; retrying", {
+      data: { attempt: 1, approveSent },
+    });
     await wait(REPAY_STALE_SIM_RETRY_DELAY_MS);
   }
 
   try {
     return await execute();
   } catch (secondError) {
-    if (!isStaleAllowanceSimulationError(secondError) || approveSent) {
-      throw secondError;
-    }
-    // Approve was skipped on a possibly stale-high read; force it once, then
-    // give the backends the same catch-up window before the final attempt.
-    await forceApprove();
+    if (!isStaleAllowanceSimulationError(secondError)) throw secondError;
+    logger.warn("Repay simulation saw a stale allowance; retrying", {
+      data: { attempt: 2, approveSent },
+    });
+    // Approve was skipped on a possibly stale-high read; force it once.
+    if (!approveSent) await forceApprove();
     await wait(REPAY_STALE_SIM_RETRY_DELAY_MS);
   }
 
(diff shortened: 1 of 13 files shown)


services/vault/src/applications/aave/services/positionTransactions.ts
```diff
@@ -95,47 +98,196 @@ export async function repay(
...
+}
+
+/**
+ * Run the repay, absorbing stale-backend allowance reverts at simulation: one
+ * plain delayed retry first; if the approve was short-circuited (a stale HIGH
+ * allowance read can skip a genuinely needed approve), force the approve
+ * before the final attempt. Mined reverts and all other errors propagate
+ * untouched — auto-retrying a broadcast tx would re-prompt the wallet.
+ * Note: repayMaxCapped can revert here legitimately (accrued interest pushed
+ * debt past the approved balance cap); the retries just delay that error.
```
gbarkhatov: Non-blocking (worth a conscious accept + one more sentence here): this Note covers the *delay*, but the `approveSent === false` branch below also fires a real `forceApprove()` — a wallet prompt + on-chain gas — in this same legitimate-revert case. When accrued interest genuinely pushes debt past the approved balance cap, the simulation revert is `ERC20InsufficientAllowance`, which is indistinguishable from a stale-backend read. So `forceApprove(balanceAmount)` runs, changes nothing (the cap is still below the true debt), and the final attempt still reverts. Rare, and hard to disambiguate cleanly, so fine to accept — but acknowledging the surprise re-approve in the comment would set expectations.

services/vault/src/copy.ts
```diff
@@ -1121,6 +1121,12 @@ export const COPY = {
       // Shown when the wallet balance query fails so the user isn't left with a
       // disabled repay button and no explanation.
       balanceLoadError: "Couldn't load your balance. Please try again.",
+      // Post-approve verification failures. Wording must avoid the substrings
+      // getEnhancedErrorMessage rewrites (see utils/errors/contract.ts).
+      approvalNotConfirmed: (required: string, observed: string) =>
```
gbarkhatov, line 1126: Cosmetic: these strings are thrown as plain `Error`, then re-wrapped by the repay hook's catch via `mapViemErrorToContractError(error, "Repay")`. With no revert data and no trigger substrings, `getEnhancedErrorMessage` prefixes them, so the user actually sees `Repay failed: Token approval could not be confirmed on-chain…`. Harmless, just slightly redundant given the copy already reads as a complete sentence. (The substring-avoidance note above is correct and still needed.)

gbarkhatov (dismissed):
Approving. Reviewed the full diff plus the surrounding error-mapping, allowance, and copy code.

The core design is sound and the safety-critical invariant holds: mined reverts are never auto-retried (post-broadcast errors are mapped without `tagSimulationPhase`, so `isSimulationPhaseError` is false), while only a simulation-phase `ERC20InsufficientAllowance` is retried. Receipt-event verification correctly filters by token/owner/spender with last-match-wins for Safe batch siblings, and `ERC20InsufficientAllowance` is decodable via `COMMON_ERROR_ABI`. The tests are genuinely good — they exercise the receipt-event path, the fallback-read exhaustion path, the retry -> force-approve ladder, and the crucial "mined revert never auto-retries" case.

Two non-blocking notes inline — both intentional trade-offs, not defects: a rare wasted approve prompt + gas on a legitimate `repayMaxCapped` revert, and a cosmetic double-prefix on the new approval-verification copy.

services/vault/src/clients/eth-contract/erc20/transaction.ts
```diff
@@ -38,38 +40,44 @@ export async function approveERC20(
...
+      !(error instanceof ContractError) ||
+      error.code !== ErrorCode.CONTRACT_REVERT
+    ) {
+      throw error;
+    }
   }
 
+  await executeWrite({
+    ...approveArgs,
+    args: [spenderAddress, 0n],
```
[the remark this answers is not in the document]
jrwbabylonlab: Fixed — the fallback now requires a pre-broadcast simulation revert AND a non-zero current allowance (read on the failure path only; if that read fails, the original revert is rethrown). A generic token restriction rethrows without ever prompting the reset. Each gate is test-pinned.

services/vault/src/applications/aave/services/positionTransactions.ts
jrwbabylonlab: Fixed — the note now calls out the no-op forceApprove (one prompt plus gas, two for USDT-style tokens) on the short-circuit path.

services/vault/src/copy.ts
jrwbabylonlab: Agreed — accepting the redundancy. Dropping the prefix would mean per-operation special-casing in getEnhancedErrorMessage, which isn't worth it for a cosmetic doubling.

jeremy-babylonlabs (approved):

jrwbabylonlab:
solution was not good enough. i will open a new one as i have done quite some refactoring

PR #2120, opened 2026-07-22
feat(vault): add artifact and provider-list telemetry (Phase 5) to Sentry

jonybur:
## Context

Phase 5 of the Sentry funnel tracking (#2052 → #2053 → #2069 → #2091 → #2099). A 46-path coverage audit of the surfaces the funnel traverses found the **artifact pipeline** (the `RETRIEVE_SECRET` gate) and the **VP picker** were complete telemetry blind spots - 21 confirmed gaps, each verified with a full propagation trace.

## The gaps this closes

**Artifact surface - no logger existed anywhere on it.** Every failure died in `setError` as red modal text: CWT/bearer rejection of the gRPC-subject token, wire `auth_expired`/missing-bearer, HTTP 401/403, envelope-prefix validation rejections (#1943/#2020), exhausted network retries, mid-stream failures - and the on-chain `prePeginTxHash` mismatch **attack guard** firing on the artifact path, a potential compromised-indexer signal that reached only the modal.

**VP picker - silent top-of-funnel outage.** A proxy marking every provider disabled, or an indexer schema regression (null `rpcUrl` / failed validation on every row), empties the deposit picker for everyone. One of those filters had no logging of any kind.

## What it adds

| Signal | Kind | When |
| --- | --- | --- |
| `activation.artifacts` | failure capture (via `captureFunnelFailure`) | any of the four artifact failure sites (`site`: `prime` / `reprime` / `download` / `cold_registry`) |
| `activation.artifacts.downloaded` | milestone (info) | **first** real download per vault (gated on `hasArtifactsDownloaded`) |
| `activation.artifacts.stalled` | warning event | VP still reports "processing" past ~5 min (30 × 10s retries), once per attempt |
| `onboarding.providers.empty` | warning event | providers exist upstream but every row dropped - `reason`: `all_rows_invalid` (fetch) / `all_disabled` (proxy), deduped per application per session |

Design notes:
- Failure captures go through #2091's `captureFunnelFailure`, so **wallet declines are not counted** and vaultId is scrubbed; cancellation (`ArtifactDownloadCancelledError`) is never captured. The collateral re-download path (no `vaultId`) falls back to the pegin txid as the join key.
- The stall event does **not** change the retry loop's behavior - the depositor experience is identical; the stall just stops being invisible. Bounding the loop is a separate product decision.
- The milestone completes stall localization in the activation funnel: `payouts.signed → verified → artifacts.downloaded → activated`.
- `onboarding.providers.empty` distinguishes a systemic outage from a legitimately provider-less application (upstream non-empty required), and is module-deduped so multiple mounts / poll refreshes emit once.

## Testing

11 new tests: capture with stage tag + scrubbed vaultId per site, capture-vs-cancel discrimination, first-download-only milestone, stall threshold + no-re-emit via fake timers, and systemic-vs-empty provider distinction on both drop paths. Full vault suite green: **243 files, 2600 tests**. eslint clean; `tsc -p tsconfig.lib.json` clean.

### DIFF
--- a/services/vault/src/services/providers/fetchProviders.ts
+++ b/services/vault/src/services/providers/fetchProviders.ts
@@ -1,6 +1,7 @@
 import { gql } from "graphql-request";
 
 import { logger } from "@/infrastructure";
+import { TELEMETRY_EVENT } from "@/infrastructure/telemetryEvents";
 
 import { graphqlClient } from "../../clients/graphql";
 import type {
@@ -175,11 +176,21 @@ export async function fetchAppProviders(
     { appController: applicationEntryPoint.toLowerCase() },
   );
 
-  const vaultProviders: VaultProvider[] = response.vaultProviders.items
-    .filter(
-      (provider): provider is typeof provider & { rpcUrl: string } =>
-        provider.rpcUrl !== null,
-    )
+  const rawProviders = response.vaultProviders.items;
+  const withRpcUrl = rawProviders.filter(
+    (
+      provider,
+    ): provider is (typeof rawProviders)[number] & { rpcUrl: string } =>
+      provider.rpcUrl !== null,
+  );
+  if (withRpcUrl.length < rawProviders.length) {
+    logger.warn("Dropped vault providers with null rpcUrl from indexer", {
+      dropped: rawProviders.length - withRpcUrl.length,
+      total: rawProviders.length,
+    });
+  }
+
+  const vaultProviders: VaultProvider[] = withRpcUrl
     .filter((provider) => validateAppProvider(provider) !== null)
     .map((provider) => ({
       id: provider.id,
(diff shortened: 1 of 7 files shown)


services/vault/src/services/providers/fetchProviders.ts
```diff
@@ -190,6 +201,18 @@ export async function fetchAppProviders(
...
     }));
 
+  // The indexer knows providers but every row was dropped (null rpcUrl or
+  // failed validation): a systemic schema/data regression that blocks every
+  // deposit at the picker. Distinct from a legitimately empty application.
+  if (rawProviders.length > 0 && vaultProviders.length === 0) {
+    logger.event(TELEMETRY_EVENT.ONBOARDING_PROVIDERS_EMPTY, {
+      level: "warning",
+      category: "onboarding",
+      reason: "all_rows_invalid",
```
jrwbabylonlab, line 228: Unlike its `all_disabled` sibling in `useVaultProviders.ts` (deduped once-per-app via the module Set), this emit has no dedup, so a persistently-bad indexer re-counts the same condition on each refetch. It also omits the `applicationId: shortId(applicationEntryPoint)` facet the sibling carries, so the signal can't be attributed to an app in Sentry. Bounded by React Query caching, not a burst — but worth making the two emits symmetric.

jrwbabylonlab (dismissed):

gbarkhatov (dismissed):

jrwbabylonlab (approved):

PR #2121, opened 2026-07-22
fix(vault): v3 deposit flow polish and shared modal shell

jeremy-babylonlabs:
- Split UtxoSplitSelector and VaultProviderSelector into dedicated v3 components (own files, gated once in DepositForm) instead of inline flag branches; both v3 pickers auto-collapse on select.
- VaultProviderSelector v3: match the design row (avatar status badge, metric columns, docs intro), drop the blue selected tint and checkmark.
- Provider/loan/vault lists share ListRow and the row button classes; VaultsSummaryCard now renders through PositionStatCards like the Overview and Loans summaries.
- FeesSection v3 disclosure animates via the shared Accordion motion and uses a boxed plus/minus toggle.
- Deposit form: app logo in the modal header, corrected amount-row fields, accent-primary CTA with stroke-primary disabled state.
- Deposit stepper width matches the design (612px).
- Fix the health-factor rail gradient so the current-price marker sits on the red-to-amber hand-off.
- Add a shared V3ModalShell (close left, network and settings right on the page column) and route deposit, borrow/repay, reorder, withdraw and the redesigned reorder-success modal through it.

### DIFF
--- a/services/vault/src/components/shared/V3ModalShell.tsx
+++ b/services/vault/src/components/shared/V3ModalShell.tsx
@@ -1,0 +1,117 @@
+/**
+ * Shared full-screen modal shell for the v3 flows.
+ *
+ * Every v3 overlay (deposit, borrow/repay, reorder, withdraw, and their
+ * success screens) renders through this so they share one header: a close
+ * button on the left and the network chip + settings menu on the right, both
+ * aligned to the same content column the page uses. The card itself is
+ * centered and capped at `contentClassName`.
+ *
+ * Pre-v3 it falls back to the plain centered `FullScreenDialog` those flows
+ * used before, so the v2 look is unchanged.
+ */
+
+import {
+  FullScreenDialog,
+  StandardSettingsMenu,
+} from "@babylonlabs-io/core-ui";
+import { useTheme } from "next-themes";
+import type { ReactNode } from "react";
+import { IoArrowBack, IoClose } from "react-icons/io5";
+import { twJoin } from "tailwind-merge";
+
+import { NetworkBadge } from "@/components/shared/NetworkBadge";
+import { PAGE_CONTENT_CLASS } from "@/components/shared/layoutClasses";
+import { FeatureFlags } from "@/config";
+import { COPY } from "@/copy";
+
+interface V3ModalShellProps {
+  open: boolean;
+  /** Omit to lock dismissal (e.g. while a tx is in flight). */
+  onClose?: () => void;
+  /** When set, the top-left control is a back arrow instead of a close X. */
+  onBack?: () => void;
+  disableEscapeClose?: boolean;
+  /** Max-width (and any extra layout) for the centered card. */
+  contentClassName?: string;
+  children: ReactNode;
+}
+
+export function V3ModalShell({
+  open,
+  onClose,
+  onBack,
+  disableEscapeClose,
+  contentClassName,
+  children,
+}: V3ModalShellProps) {
+  const { theme, setTheme } = useTheme();
+
+  if (!FeatureFlags.isV3UiEnabled) {
+    return (
+      <FullScreenDialog
+        open={open}
+        onClose={onClose}
+        onBack={onBack}
+        disableEscapeClose={disableEscapeClose}
+        className="items-center justify-center p-6"
+      >
+        <div className={twJoin("mx-auto w-full", contentClassName)}>
+          {children}
+        </div>
+      </FullScreenDialog>
+    );
+  }
+
+  return (
+    <FullScreenDialog
+      open={open}
+      onClose={onClose}
+      onBack={onBack}
+      disableEscapeClose={disableEscapeClose}
+      // The shell renders its own header aligned to the page column, so hide
+      // FullScreenDialog's fixed built-in close/back button (onClose is still
+      // wired for backdrop + Escape dismissal).
+      closeButtonClassName="!hidden"
+    >
+      {/* Header row: close/back on the left, network + settings on the right,
+          both on the page's content column so the modal chrome lines up with
+          the page underneath. */}
+      <div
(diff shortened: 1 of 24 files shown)


services/vault/src/components/simple/ReorderVaults/ReorderSuccessModal.tsx
```diff
@@ -19,25 +16,31 @@ export function ReorderSuccessModal({
...
-    <ResponsiveDialog open={isOpen} onClose={onClose}>
-      <DialogHeader title={REORDER_SUCCESS_TITLE} onClose={onClose} />
-      <DialogBody>
-        <ReorderSuccessIcon className="mx-auto my-3 text-accent-primary" />
-        <p className="py-3 text-center text-accent-secondary">
-          {REORDER_SUCCESS_TEXT}
-        </p>
-      </DialogBody>
-      <DialogFooter>
+    <V3ModalShell
```
jrwbabylonlab, line 52: This modal is reachable in v2 (rendered by `CollateralSection` only when the v3 flag is off), and its base was `ResponsiveDialog`. `V3ModalShell`'s pre-v3 fallback is `FullScreenDialog`, so in v2 the success screen flips from a mobile bottom-sheet to a full-screen overlay — a v2 regression, and it contradicts the shell's own docstring ("the v2 look is unchanged"). Every other migrated modal was already FullScreenDialog, so this one is the sole divergence. Functionally safe (`onClose` preserved).

services/vault/src/components/shared/V3ModalShell.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { PAGE_CONTENT_CLASS } from "@/components/shared/layoutClasses";
+import { FeatureFlags } from "@/config";
+import { COPY } from "@/copy";
+
+interface V3ModalShellProps {
+  open: boolean;
+  /** Omit to lock dismissal (e.g. while a tx is in flight). */
+  onClose?: () => void;
+  /** When set, the top-left control is a back arrow instead of a close X. */
+  onBack?: () => void;
```
jrwbabylonlab: `onBack` (plus the back-button branch, `IoArrowBack` import, and the `COPY.common.back` string it renders) has no consumer — every V3ModalShell user is in this diff and none passes it. Zero-dead-code: drop the prop and the copy key until something needs it.

services/vault/src/constants.ts
```diff
@@ -25,6 +25,10 @@ export const SUPPORT_URL = "https://discord.com/invite/babylonglobal";
 export const TWO_VAULT_SPLIT_DOCS_URL =
   "https://docs.babylonlabs.io/trustless-bitcoin-vault/use-for-lending/create-a-vault/#step-1-decide-how-to-split-your-btc";
 
+// Vault provider docs link, surfaced from the v3 provider picker intro.
+// TODO(product): swap in the exact vault-provider page once confirmed.
+export const VAULT_PROVIDER_DOCS_URL = "https://docs.babylonlabs.io";
```
jrwbabylonlab: Placeholder docs homepage, but the V3 provider picker ships a live link whose copy promises the vault-provider / create-your-own page. User-reachable when v3 is on — point it at the real VP docs before this flag flips, or gate the link.

services/vault/src/components/simple/ReorderVaults/ReorderSuccessModal.tsx
```diff
@@ -1,10 +1,7 @@
...
-  DialogFooter,
-  DialogHeader,
-  ResponsiveDialog,
-} from "@babylonlabs-io/core-ui";
+import { Button } from "@babylonlabs-io/core-ui";
+
+import { V3ModalShell } from "@/components/shared/V3ModalShell";
+import { COPY } from "@/copy";
 
 import { REORDER_SUCCESS_TEXT, REORDER_SUCCESS_TITLE } from "./constants";
```
jrwbabylonlab: Nit: the button moved to `COPY.reorder.doneButton`, but the title/body are still pulled from `./constants` — per the copy rule they belong in `copy.ts` too, since the file was already touched.

services/vault/src/components/simple/VaultProviderSelectorV3.tsx
```diff
@@ -0,0 +1,244 @@
...
+                            <IoCheckmarkCircle
+                              className="text-success-bright"
+                              size={12}
+                            />
+                          )}
+                        </span>
+                      </span>
+                      <span className="flex flex-col justify-center">
+                        <span className="text-sm text-accent-primary">
+                          {provider.name}
```
jrwbabylonlab: Feature regression vs v2: `VaultProviderListItem` carries `explorerUrl` (`vaultProvider.ts:113`) and the v2 selector renders it via `<ExplorerLink>` (`VaultProviderSelector.tsx:192-196`), but this V3 row has no `explorerUrl`/`ExplorerLink` — so when `NEXT_PUBLIC_TBV_VP_EXPLORER_URL` is set and a provider has one, v3 users lose the link v2 users have. Carry the explorer link into the V3 row (or confirm it's intentionally dropped).

jrwbabylonlab (approved):

jonybur (approved):
Reviewed the full diff with a focus on the modal shell migration. V3ModalShell forwards onClose/disableEscapeClose through both branches, so the deliberate three-path dismissal lock during in-flight transactions (close button, backdrop, ESC) is preserved for the Aave detail flow, and the v2 fallback stays byte-equivalent. The e2e-load-bearing selectors survive the selector rewrite - the deposit-button testid is carried through PositionStatCards, and the aria-pressed/role-button hooks the real-wallet CLI drives are intact in the new v3 pickers. Copy moved to copy.ts, the FeesSection disclosure now rides the shared Accordion motion, and the v2/v3 gating is collapsed to one seam in DepositForm with the flag branches removed from the v2 components. Looks good.

PR #2122, opened 2026-07-23
Fix aave repay stale rpc hardening

jrwbabylonlab:
## 1. Repays tolerate lagging RPC backends

The load-balanced RPC can serve one-block-stale state, which failed
every repay attempt before broadcast. Now: approvals are verified from
the approve tx's own receipt (`Approval` event) instead of a re-read;
only provably pre-broadcast `ERC20InsufficientAllowance` simulation
failures are retried (bounded, logged — mined reverts never re-prompt);
and the `approve(0)` reset ("revoke approval" popup) fires only for
genuine USDT-style reverts.

## 2. Full repays clear to exactly zero

Amount-based full repay quoted the fee-blind Spoke debt; the adapter's
fee skim plus share-floor rounding could leave unburnable dust that
redisplays forever. One `repayAll` replaces `repayFull`/`repayMaxCapped`:
fee-inclusive quote from the position proxy (new minimal ABI + reader),
approval capped at `min(quote + 0.5% buffer, balance)`, repay-all
sentinel (`maxUint256`) resolved exactly in-tx — dust impossible by
construction. `RepayMode` collapses to `"partial" | "max"`; allowance
machinery and retry engine extracted to their own modules.


services/vault/src/applications/aave/services/repayStaleSimulationRetry.ts
```diff
@@ -0,0 +1,79 @@
...
+ * only mean the simulating backend has not caught up to the approve block.
+ */
+function isStaleAllowanceSimulationError(err: unknown): boolean {
+  return (
+    isSimulationPhaseError(err) &&
+    err.code === ErrorCode.CONTRACT_REVERT &&
+    err.reason === ERC20_INSUFFICIENT_ALLOWANCE_REASON
+  );
+}
+
```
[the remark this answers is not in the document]
jrwbabylonlab: Conscious accept, documented in the JSDoc right below: at this layer a legit accrual-past-cap shortfall is indistinguishable from a stale backend, and the budget is hard-bounded — three simulations, at most one no-op approve, ~6s — before the real error surfaces. repayAll's pre-throw removes the common case (balance short at quote time); only mid-flight accrual remains.

services/vault/src/clients/eth-contract/erc20/transaction.ts
```diff
@@ -38,38 +50,60 @@ export async function approveERC20(
...
+      throw error;
+    }
+    // The quirk only exists for non-zero → non-zero approvals: with a zero
+    // allowance the revert has some other cause — surface the original error
+    // instead of firing a pointless reset prompt. Failure-path-only read; if
+    // the read itself fails the condition is unproven, so the original revert
+    // (not the read error) surfaces.
+    const ownerAddress = walletClient.account?.address;
+    if (!ownerAddress) throw error;
+    const currentAllowance = await getERC20Allowance(
```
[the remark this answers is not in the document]
jrwbabylonlab: Both legs of the fallback are simulation-gated, so a paused/blocklisted token reverts the reset pre-popup too — nothing broadcasts. The reachable residual needs a token where approve(0) passes but approve(N) fails for a non-zero-first reason (e.g. an amount cap): none of our listed tokens, and the allowance being reset is already insufficient for the repay by construction (ensureAllowance only approves when allowance < required). Accepting.

jonybur (dismissed):
Went through the full diff. The receipt-based approval verification (token/owner/spender-filtered Approval event with the bounded read fallback) closes the stale-read failure cleanly, and the retry engine is strictly scoped: bounded attempts, only pre-broadcast ERC20InsufficientAllowance simulation reverts, mined reverts never re-enter, absorbed retries logged. On the value path, the fee-inclusive proxy quote, ceil-rounded buffer, and min(quote + buffer, balance) cap check out - the sentinel never reaches an approve, and the pre-throw means the marginal-balance band now gets an actionable error instead of the opaque simulation revert it hit before. Dead code from the old repayFull/repayMaxCapped modes is fully purged, including tests and docs. Nice work on documenting the accepted tradeoffs inline.

services/vault/src/applications/aave/services/repayStaleSimulationRetry.ts
```diff
@@ -0,0 +1,79 @@
...
+  try {
+    return await execute();
+  } catch (secondError) {
+    if (!isStaleAllowanceSimulationError(secondError)) throw secondError;
+    logger.warn("Repay simulation saw a stale allowance; retrying", {
+      data: { attempt: 2, approveSent },
+    });
+    // Approve was skipped on a possibly stale-high read; force it once.
+    if (!approveSent) await forceApprove();
```
gbarkhatov, line 74: `abortableSleep` is called without a signal here (and at L63, plus `ensureAllowance.ts:94`) — none of the three retry/verify sleeps are cancellable. If the user abandons the flow, the loop still runs to ~6s and, on the `approveSent === false` branch, this `forceApprove` can pop a wallet prompt *after* they've left the screen. Consider threading an `AbortSignal` from the component's unmount/cancel into the retry engine. Low severity — the util already supports it.

services/vault/src/applications/aave/services/repayStaleSimulationRetry.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Run the repay, absorbing stale-backend allowance reverts at simulation:
+ * three attempts with catch-up delays on every path. When the approve was
+ * short-circuited (a stale HIGH allowance read can skip a genuinely needed
+ * approve), the second failure forces the approve before the last attempt; a
+ * receipt-verified approve can't be improved on, so that path just
+ * re-simulates. Mined reverts and all other errors propagate untouched —
+ * auto-retrying a broadcast tx would re-prompt the wallet.
+ * Note: a repay-all capped by the user's balance can revert here legitimately
```
gbarkhatov, line 43: Good that this edge is documented. One follow-up thought: on the balance-capped legit revert (interest accrues past the cap during the ~6–9s of retries), the user ends up seeing an `ERC20InsufficientAllowance`-style error rather than a “balance can't cover accrued interest” message, plus one no-op `forceApprove` prompt. Extreme edge, consciously accepted — worth revisiting only the surfaced copy, and only if it shows up in practice.

services/vault/src/applications/aave/services/positionTransactions.ts
```diff
@@ -172,123 +137,122 @@ export async function repayPartial(
...
-  if (currentDebt === 0n) {
+  if (quote === 0n) {
     throw new Error("No debt to repay");
   }
+  // Pre-throw instead of letting the sentinel hit ERC20InsufficientAllowance
+  // on-chain: that revert is the stale-simulation retry's trigger, so an
+  // underfunded repay would burn the retry budget before erroring.
+  if (balanceRaw < quote) {
+    throw new Error(
+      "insufficient balance to fully repay: not enough stablecoin to cover the debt plus interest",
```
gbarkhatov: This user-facing string is inlined in the service; per CLAUDE.md it belongs in `copy.ts` (like the new `approvalNotConfirmed` / `approvalBelowRequired`). It also becomes newly reachable via the `"max"` route: `pickRepayParams` decides on the fee-*exclusive* Spoke debt while `repayAll` quotes fee-*inclusive*, so a user with balance ≥ their visible debt but < the fee-inclusive quote lands here and sees “insufficient balance to fully repay” — confusing when the displayed debt looks covered. Migrating to `copy.ts` (and softening the wording) would help.

services/vault/src/applications/aave/services/positionTransactions.ts
```diff
@@ -172,123 +137,122 @@ export async function repayPartial(
...
+ * re-resolves the fee-inclusive debt inside the repay tx and pulls exactly
+ * that, hitting the contract's exact full-repay branch — every debt share is
+ * burned, so no rounding dust can survive. The proxy quote is a hard
+ * dependency: a failed read blocks with a retryable error rather than
+ * degrading to a fee-blind amount.
  *
- * Fetches exact debt from contract, handles approval, then repays.
- * Uses pinned adapter and spoke addresses from trusted environment config.
+ * Approval semantics (all consciously accepted):
+ * - Success leaves `cap − pulled` allowance standing (≈ the buffer for funded
```
gbarkhatov, line 183: Behavioral change worth an explicit second-reviewer sign-off (critical path): with the sentinel the adapter pulls only the actual debt, so `cap − pulled` (≈ the 0.5% buffer) remains a **standing ERC-20 allowance** to the adapter — whereas the old `repayFull` left residual allowance = 0. The reasoning here (env-pinned adapter derives the position from `borrower` on-chain, so an allowance can only ever clear the user's own debt) is sound; flagging the invariant change, not disputing it.

packages/babylon-ts-sdk/src/tbv/integrations/aave/clients/positionProxy.ts
```diff
@@ -0,0 +1,35 @@
...
+  reserveId: bigint,
+): Promise<bigint> {
+  const result = await publicClient.readContract({
+    address: proxyContract,
+    abi: AaveAdapterPositionProxyABI,
+    functionName: "getReserveTotalDebt",
+    args: [reserveId],
+  });
+
+  return result as bigint;
```
gbarkhatov: `result as bigint` returns the contract value with no bounds check, which sits slightly against the repo's “assert external/contract outputs on value-bearing paths” ethos. Impact is bounded in practice — `cap = min(quote + buffer, balance)` caps over-approval at the user's balance, and `quote === 0` pre-throws upstream — so an understated quote just reverts on-chain and burns the retry budget. A cheap sanity assert wouldn't hurt. Low priority.

services/vault/src/clients/eth-contract/erc20/transaction.ts
```diff
@@ -38,38 +54,60 @@ export async function approveERC20(
...
+    ) {
+      throw error;
+    }
+    // The quirk only exists for non-zero → non-zero approvals: with a zero
+    // allowance the revert has some other cause — surface the original error
+    // instead of firing a pointless reset prompt. Failure-path-only read; if
+    // the read itself fails the condition is unproven, so the original revert
+    // (not the read error) surfaces.
+    const ownerAddress = walletClient.account?.address;
+    if (!ownerAddress) throw error;
```
gbarkhatov, line 91: This failure-path guard replaces the old up-front `if (!ownerAddress) throw "Wallet account not available…"` — a no-account wallet now falls through to the first `executeWrite` instead of erroring early. All repay callers already check `walletClient.account?.address`, so this is only a robustness reduction for direct callers of `approveERC20`. Low priority; noting it since the explicit guard was dropped.

gbarkhatov (dismissed):
Approving. Careful, well-tested hardening of the Aave repay critical path — simulation-vs-mined error tagging, receipt-event approval verification, and the `maxUint256` sentinel (exact full-repay branch → no rounding dust) are all sound, and coverage is thorough (99 affected tests green locally). Left a few inline notes; all are low-severity or follow-up.

Per the critical-path rule this still needs a second human reviewer — the standing-allowance invariant change (inline on `repayAll`) is the one to sign off on explicitly.

services/vault/src/applications/aave/services/positionTransactions.ts
jrwbabylonlab: Fixed — moved to copy.ts as balanceBelowFullRepay with formatted amounts and fee-aware wording: "Repaying in full needs {quote} (including accrued interest and fees), but your balance is {balance}…" — names the real requirement instead of contradicting the displayed debt. The old mapper rewrite branch lost its only producer and is removed.

packages/babylon-ts-sdk/src/tbv/integrations/aave/clients/positionProxy.ts
jrwbabylonlab: Fixed — the cast is now a typeof assert that throws with the reserve id.

services/vault/src/applications/aave/services/repayStaleSimulationRetry.ts
jrwbabylonlab: Deferred as a follow-up: proper cancellation needs an AbortController plumbed from the hook's unmount through repayAll/ensureAllowance signatures, which deserves its own change. Exposure is bounded — ~6s of sleeps, at most one prompt, same class as other flows' post-navigation prompts.

services/vault/src/applications/aave/services/repayStaleSimulationRetry.ts
jrwbabylonlab: Agreed — will revisit the surfaced copy if the retry logs show it happening in practice.

services/vault/src/applications/aave/services/positionTransactions.ts
jrwbabylonlab: Acknowledged — intentional invariant change, documented in the JSDoc for exactly this sign-off. The residual is ≈ the 0.5% buffer on success and is strictly smaller than the standing balance-sized allowance the old max-capped path already left.

services/vault/src/clients/eth-contract/erc20/transaction.ts
jrwbabylonlab: Kept as-is: executeWrite's own account guard (transactionFactory.ts) fires before any popup for direct callers too, so the early throw still happens one layer down; the failure-path guard covers the allowance read.

jeremy-babylonlabs (approved):

PR #2123, opened 2026-07-23
fix: empty state

jeremy-babylonlabs:
unfied empty state UI for vault loan and activity tab


services/vault/src/components/shared/EmptyState.tsx
```diff
@@ -87,43 +79,44 @@ export function EmptyState({
...
       ) : (
-        avatarUrl && (
-          <Avatar
-            url={avatarUrl}
-            alt={avatarAlt ?? ""}
-            size="xlarge"
-            className="mb-2 h-[100px] w-[100px]"
-          />
-        )
+        <EmptyStateIcon />
```
gbarkhatov: **Illustration is now always-on for icon-less callers — a side effect outside this PR's scope.** The old branch was `icon ? … : avatarUrl && <Avatar/>`, so a caller passing neither rendered *no* illustration. This new `avatarUrl ? <Avatar/> : <EmptyStateIcon/>` always renders the 94×101 doc illustration when there's no avatar.

`LiquidationAnalysisSection` (the no-collateral / no-loans states) passes neither and is outside the stated scope (vaults / loans / activity). It will now gain the large illustration **and** a `size="large"` button where it previously had neither, with no test covering it. Please confirm that's wanted there — if not, the always-show default is too aggressive.

services/vault/src/components/shared/EmptyState.tsx
```diff
@@ -87,43 +79,44 @@ export function EmptyState({
...
+        )}
+      </div>
     </div>
   );
 
   if (withCard) {
-    return <Card>{content}</Card>;
+    // Figma: background/secondary fill, 16px radius, 24/40 padding, no border.
+    return (
+      <Card className="border-0 bg-background-secondary px-6 py-10">
```
gbarkhatov: **v3 card fill + text sizing now reach the one remaining v2 caller.** Deleting `descriptionVariant` means the reserve-detail connect prompt (`aave/components/Detail/index.tsx`, the `avatarUrl` path) loses its old `compact` look: the description goes `text-sm` → `text-base` centered/width-capped, and the card here gains `bg-background-secondary` + `border-0` + `py-10`.

The override itself is correct (source-order wins), so this isn't a bug — but that v2 surface is out of the stated scope and now looks identical to the v3 empty states. Worth a visual sanity check that it still reads as intentional.

services/vault/src/components/pages/__tests__/Loans.test.tsx
```diff
@@ -122,6 +129,11 @@ describe("Loans page — loading gate", () => {
       "data-connected",
       "false",
     );
+    // Disconnected prompts for a wallet — it must not claim the depositor has
+    // no loans when we haven't looked at an address yet.
+    expect(
+      screen.getByText("Connect your wallet to view your loans"),
```
gbarkhatov: Prefer `COPY.loans.emptyDisconnected` over this hardcoded literal — the sibling `VaultsPage.test.tsx` references `COPY.vaults.empty.disconnected`, and a hardcoded string silently rots if the copy ever changes without the assertion following.

services/vault/src/components/pages/Loans.tsx
```diff
@@ -74,9 +74,12 @@ export default function Loans() {
...
-          title={COPY.loans.noActiveLoans.title}
-          description={COPY.loans.noActiveLoans.body}
+          title={
+            isConnected
+              ? COPY.loans.noActiveLoans.title
+              : COPY.loans.emptyDisconnected
+          }
+          description={isConnected ? COPY.loans.noActiveLoans.body : undefined}
           isConnected={isConnected}
           actionLabel={COPY.overview.depositAction}
```
gbarkhatov, line 108: Minor: `actionLabel` / `onAction` are inert on this branch while disconnected — `EmptyState` renders `<Connect/>` (not the labeled action) when `isConnected={false}`. Harmless, but slightly misleading to a reader who'd expect the Deposit action to show here.

services/vault/src/components/shared/icons/EmptyStateIcon.tsx
```diff
@@ -12,49 +13,24 @@ interface EmptyStateIconProps {
 export function EmptyStateIcon({ className = "" }: EmptyStateIconProps) {
```
gbarkhatov: Two nits:
1. `className` is now dead — the sole caller renders `<EmptyStateIcon />` with no args. Per the zero-dead-code policy it's a removal candidate (defensible to keep on a reusable icon).
2. The new paths drop `strokeWidth` (was `2.5`), so strokes render at the default `1px`. The commit message says this is the Figma export, so presumably intended — just worth a glance against the design, since it's a visible weight change.

services/vault/src/components/pages/VaultsPage.tsx
```diff
@@ -202,7 +194,8 @@ export default function VaultsPage() {
             data-testid="deposit-button"
             variant="contained"
             color="secondary"
-            size="medium"
+            size="large"
+            className="min-w-[120px]"
```
gbarkhatov: `min-w-[120px]` re-inlines the `ACTION_WIDTH_CLASS` constant defined in `EmptyState.tsx`. Since the vaults CTA uses its own `DepositButton` (not the shared button) the duplication is unavoidable, but the two now have to stay in sync by hand — exporting the constant would prevent drift.

gbarkhatov (dismissed):
Approving — a clean, well-scoped consolidation of the v3 empty state onto a single surface. Dead code (the `icon` / `descriptionVariant` props, the standalone `vaults-empty.svg`, the barrel re-exports) is fully removed, and the disconnected copy is centralized via the `connectToView` builder. All comments below are non-blocking.

Verified as non-issues: the `.bbn-card` padding/border override works (core-ui's CSS is `@import`ed before the app's `@tailwind utilities`, so equal-specificity app utilities win by source order); the disconnected-vs-paused precedence on Vaults is correct and newly tested; and every shared-`EmptyState` caller passes `withCard`, so dropping the non-card `py-16` strands nothing.

The one thing worth confirming before merge is the first comment: the illustration is now always-on for icon-less callers, which pulls `LiquidationAnalysisSection` (out of scope, untested) into the new look.

gbarkhatov:
@jeremy-babylonlabs are comments resolved?

jrwbabylonlab (approved):

jonybur (approved):
Well-scoped consolidation. The `icon` prop and `descriptionVariant` are cleanly retired in favour of the `variant` split (v3 / compact), and every caller is updated. Dead assets go too: `vaults-empty.svg` deleted and `EmptyStateIcon` dropped from both the icons barrel and the shared re-export. `EmptyStateIcon` remains reachable only via direct import from `EmptyState.tsx`, its one consumer. The disconnected-vs-paused precedence on `VaultsEmptyState` is right — pause outranks the connect prompt so a depositor learns deposits are off without connecting first, pinned down by a dedicated test. Disconnected-title-only copy is centralised via `connectToView(...)` so the three tabs can't drift.

PR #2126, opened 2026-07-23
fix(vault): show the vaults empty state while a deposit is pending

jeremy-babylonlabs:
A pending deposit kept the /vaults page in its populated layout, and VaultsActiveSection rendered nothing with zero vaults — so the Vaults section disappeared entirely instead of showing the empty state.

The section now renders a "Vaults" heading with the empty-state card until a vault confirms and activates, at which point it switches back to the "Active Vaults (n)" rows. The prompt itself moves into VaultsEmptyState, shared by the page-level and section placements.

## Screenshot

Screenshot 2026-07-23 at 18 22 32

### DIFF
--- a/services/vault/src/components/pages/VaultsPage.tsx
+++ b/services/vault/src/components/pages/VaultsPage.tsx
@@ -15,7 +15,6 @@ import { useOutletContext } from "react-router";
 import type { Address } from "viem";
 
 import type { RootLayoutContext } from "@/components/pages/RootLayout";
-import { DepositButton, EmptyState } from "@/components/shared";
 import { PAGE_CONTENT_CLASS } from "@/components/shared/layoutClasses";
 import {
   isDepositBlocked,
@@ -28,6 +27,7 @@ import {
 } from "@/components/simple/ReorderVaults";
 import WithdrawFlow from "@/components/simple/WithdrawFlow";
 import { VaultsActiveSection } from "@/components/vaults/VaultsActiveSection";
+import { VaultsEmptyState } from "@/components/vaults/VaultsEmptyState";
 import { VaultsLifecycleSections } from "@/components/vaults/VaultsLifecycleSections";
 import { VaultsSummaryCard } from "@/components/vaults/VaultsSummaryCard";
 import { FeatureFlags } from "@/config";
@@ -39,8 +39,6 @@ import { useProtocolGateState } from "@/hooks/useProtocolGate";
 import { useVaultsPageData } from "@/hooks/useVaultsPageData";
 import { useVaultsPageEmptiness } from "@/hooks/useVaultsPageEmptiness";
 import { invalidateVaultQueries, vaultOrderQueryKey } from "@/utils/queryKeys";
-
-const EMPTY_ILLUSTRATION_SRC = "/images/vaults-empty.svg";
 
 // Dev-only god-mode panel, lazily imported behind `import.meta.env.DEV` so its
 // code is dropped from production builds entirely (same pattern as
@@ -86,6 +84,16 @@ export default function VaultsPage() {
   const [isReorderSuccess, setIsReorderSuccess] = useState(false);
 
   const isDepositsPaused = FeatureFlags.isDepositDisabled;
+
+  const renderEmptyState = (withCard: boolean) => (
+    <VaultsEmptyState
+      isConnected={isConnected}
+      isDepositsPaused={isDepositsPaused}
+      isDepositDisabled={isDepositBlocked(gate)}
+      onDeposit={() => openDeposit()}
+      withCard={withCard}
+    />
+  );
 
   const reorderableVaults = useMemo(
     () => rawCollateralVaults.filter((vault) => !vault.isActivating),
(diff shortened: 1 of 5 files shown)


services/vault/src/components/pages/VaultsPage.tsx
```diff
@@ -147,6 +155,10 @@ export default function VaultsPage() {
           vaults={displayVaults}
           onWithdraw={handleWithdrawRow}
           isWithdrawDisabled={isWithdrawBlocked(gate)}
+          // Pending deposits keep the page populated while the vault list is
+          // still empty — the section shows the empty state until the deposit
+          // confirms and activates.
+          emptyState={renderEmptyState(true)}
```
gbarkhatov, line 161: **Duplicate `data-testid="deposit-button"` in the pending state (medium).** `VaultsSummaryCard` always renders a `deposit-button` testid inside `populatedBody`, and passing `emptyState` here makes the section render a *second* one whenever `displayVaults` is empty (the pending-deposit state) — two identical testids on the page at once.

The current real-wallet E2E consumers use `.first()` (`pegin.ts`, `resume.ts`, `walletConnect.ts`) and drive from the dashboard route, so nothing breaks today, but it's a latent trap for any future test that doesn't. It also makes `VaultsSummaryCard`'s comment ("it takes over from the empty state's Deposit button once vaults exist") stale, since the two now coexist.

Consider dropping the testid from this section-placement CTA (the summary card already carries it in this state) and updating that comment either way.

services/vault/src/components/vaults/VaultsActiveSection.tsx
```diff
@@ -128,8 +132,23 @@ export function VaultsActiveSection({
   vaults,
   onWithdraw,
   isWithdrawDisabled,
+  emptyState,
 }: VaultsActiveSectionProps) {
-  if (vaults.length === 0) return null;
+  if (vaults.length === 0) {
```
gbarkhatov, line 137: **Empty state now renders for the inactive-only case too (product question).** Because `emptyState` is passed unconditionally from `VaultsPage`, this branch fires whenever `displayVaults.length === 0` — including when the account has *only* inactive/expired vaults (no pending, no active). `hasAnythingToShow` is true via `expiredActivities`, so `populatedBody` renders and the user sees the large "Your BTC Vaults will appear here" + Deposit prompt sitting *above* their Inactive Vaults list, where previously the section was simply absent.

The PR only motivates the pending case — is the inactive-only presentation intended? Related: `hasDisplayCollateral` (`useDashboardState`) and `displayVaults` (`useVaultsPageData`) come from different hooks; if they ever diverge you'd get this empty prompt beside a summary card showing real totals. Worth confirming that can't happen.

services/vault/src/components/vaults/VaultsEmptyState.tsx
```diff
@@ -0,0 +1,78 @@
...
+  onDeposit: () => void;
+  /** Render on the page's card surface — the section placement sits on its
+   *  own panel, matching the summary card and the lifecycle rows. */
+  withCard?: boolean;
+}
+
+/** Same surface as the summary card and the lifecycle rows, at the empty
+ *  state's larger 16px radius. */
+const CARD_CLASS =
+  "w-full rounded-2xl border border-secondary-strokeLight bg-secondary-highlight dark:bg-[#202020]";
```
gbarkhatov, line 31: **Card radius differs from the adjacent summary card (minor / design confirm).** This uses `rounded-2xl` (16px), but the summary card (`VaultsSummaryCard`) and the lifecycle/active rows (`VaultsActiveSection`) are all `rounded-lg` (8px). In the pending-deposit view this empty card sits directly below the summary card with a visibly different corner radius.

The comment above frames 16px as deliberate ("the empty state's larger 16px radius"), so this may be intentional — just worth a one-line confirm from design; if not, `rounded-lg` matches its neighbors. Not a correctness issue.

services/vault/src/components/vaults/VaultsEmptyState.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+interface VaultsEmptyStateProps {
+  isConnected: boolean;
+  /** Deposits kill-switch — swaps in the paused copy. */
+  isDepositsPaused: boolean;
+  isDepositDisabled: boolean;
+  onDeposit: () => void;
+  /** Render on the page's card surface — the section placement sits on its
+   *  own panel, matching the summary card and the lifecycle rows. */
+  withCard?: boolean;
```
gbarkhatov, line 25: **Terminology note (no change needed).** The shared `EmptyState` already has its own `withCard` prop that wraps content in a core-ui `<Card>`. This `withCard` deliberately does something different — wraps in `CARD_CLASS` to get the v3 summary-card surface — and does *not* forward to `EmptyState.withCard`. Reasonable choice, but `withCard` now means two different things across the two components; flagging for the next reader.

gbarkhatov (approved):
Approving — a clean, well-documented fix that correctly restores the Vaults empty state while a deposit is pending, with focused tests. A few non-blocking notes left inline:

- A duplicate `data-testid="deposit-button"` in the pending state (and a now-stale comment in `VaultsSummaryCard`).
- A product question: the empty state now also renders in the inactive-only case.
- A card-radius mismatch vs the adjacent summary card, worth a design confirm.
- A minor `withCard` terminology note.

None of these block merge.

PR #1859, opened 2026-06-09
feat(vault): add security headers and extract process polyfill

jonybur:
## What

Adds a Content-Security-Policy meta tag, Subresource Integrity on the built entry scripts, and dev/preview-server response headers to the Vite-based vault app, and eliminates the one inline script that was blocking a strict CSP.

**Changes:**
- `index.html` — adds a `Content-Security-Policy` meta tag (`wasm-unsafe-eval` for WASM, `worker-src 'self' blob:` for Sentry Replay's worker, `frame-src` for the WalletConnect/AppKit verify + secure iframes, `http:`/`ws:` in `connect-src` for local and e2e), plus `Referrer-Policy`
- `public/process-polyfill.js` — extracts the `window.process` polyfill from an inline `<script>` into a static file so no inline scripts remain
- `src/build/sriPlugin.ts` — build plugin that adds Subresource Integrity (`integrity` + `crossorigin`) to every local entry script / module-preload in the built `index.html`, hashing the bytes Vite writes to disk, and **fails the build** if any local JS is left unprotected
- `vite.config.ts` — adds dev/preview-server `SECURITY_HEADERS` (`X-Frame-Options: DENY`, `X-Content-Type-Options`, `Referrer-Policy`, `Permissions-Policy`) and wires the SRI plugin. The CSP is **not** duplicated here — the `index.html` meta tag is the single source of truth (a header CSP would also govern React Fast Refresh's inline preamble and white-screen `pnpm dev`)
- `eslint.config.js` — excludes `public/` from linting (static assets are not TypeScript)

## What this PR protects in production vs. not

The vault app is built to static files, archived, and uploaded to S3, then served by a CDN (see `service-release-vault.yml`). There is **no application server in production** — so only protections baked into the bundle reach real users:

| Protection | Reaches production? | How |
| --- | --- | --- |
| Content-Security-Policy | ✅ Yes | `<meta>` tag in `index.html` (ships in the bundle) |
| Referrer-Policy | ✅ Yes | `<meta name="referrer">` in `index.html` |
| `X-Frame-Options: DENY` | ❌ No | only sent by the Vite dev/preview server |
| `X-Content-Type-Options: nosniff` | ❌ No | header-only; `<meta http-equiv>` form is a spec no-op |
| `Permissions-Policy` | ❌ No | only sent by the Vite dev/preview server |
| CSP `frame-ancestors` | ❌ No | header-only; ignored when delivered via `<meta>` |

[...]

PR #2127, opened 2026-07-23
fix(vault): tighten index.html meta CSP

jonybur:
## Summary

Follow-up to #1859. An external audit finding ("No Content-Security-Policy delivered from the client") was filed against a pre-#1859 snapshot; the meta CSP already ships, but several directives were effectively unbounded. This PR tightens them. The meta tag is the only CSP on mainnet and localhost (the edge CSP currently covers testnet only), so the wildcards were load-bearing there.

## Changes

- **connect-src**: drops blanket `http:` / `ws:` (any host was previously reachable over plaintext). Keeps `https:` / `wss:` - vault-provider proxy, RPC, mempool, indexer, and Sentry hosts differ per environment and mainnet hosts are not visible from the repo - plus explicit `http://localhost:*` and `ws://localhost:*` carve-outs. The carve-outs are load-bearing: Vite HMR's websocket and the cross-port Playwright mock servers (see the new comment in `vite.config.ts`).
- **font-src**: pins the `https:` wildcard to the only two font hosts in the built bundle: `fonts.gstatic.com` (Poppins via the Tomo SDK import) and `fonts.reown.com` (AppKit KHTeka). Keeps `data:` for the BaseSans font the Coinbase/Base wallet SDK injects.
- **frame-src**: removes `secure.walletconnect.com` - both installed `@reown/appkit-wallet` trees (1.7.8, 1.8.12) hardcode only `https://secure.walletconnect.org/sdk`; the `.com` host appears nowhere in the built bundle.
- **form-action 'self'** (new): closes the one exfiltration channel the old policy left fully open (an injected `<form action=...>`); all forms in the bundle omit `action` and preventDefault, so `'self'` is a no-op for legitimate flows.
- **env.ts**: the intentional dead-endpoint sentinel `FALLBACK_RPC_URL` is now `https://invalid.local` so it stays within `connect-src https:` and fails as a clean network error instead of a CSP violation.

## Deliberately not tightened

- `img-src https:` stays: the sidecar `/logo` endpoint returns arbitrary https logo URLs with no host validation, so any pin would silently degrade provider logos. Needs an infra commitment to a fixed logo host list first.
- Per-environment `connect-src` host allowlists via the `%VAR%` templating `index.html` already uses are feasible but deferred: mainnet hosts exist only as GitHub environment vars, and three build paths (lint_test, PR verify, local `pnpm build`) run without endpoint env vars and would fail closed. Best done together with edge/meta CSP convergence.

## Verification

- Typecheck, eslint, prettier pass.
- Full vitest suite: 2598 passed.
- Playwright e2e ran the app in Chromium under the new policy: 15/19 passed; the 4 failures reproduce identically on clean `main` (local `.env` files leak into the "missing configuration" webServer) and are unrelated to this change.

## Maintenance note

The `frame-src` and `font-src` pins are version-coupled to `@reown/appkit*`: on any AppKit bump, re-grep the built assets for `secure.walletconnect` / `verify.walletconnect` and `https://fonts.` before shipping.

### DIFF
--- a/services/vault/index.html
+++ b/services/vault/index.html
@@ -26,7 +26,10 @@
     <meta name="twitter:image:width" content="2048" />
     <meta name="twitter:image:height" content="1170" />
     <meta name="version" content="%NEXT_PUBLIC_COMMIT_HASH%" />
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' 'wasm-unsafe-eval'; connect-src 'self' https: http: wss: ws:; img-src 'self' data: https: blob:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; font-src 'self' data: https:; object-src 'none'; base-uri 'self'; worker-src 'self' blob:; frame-src 'self' https://verify.walletconnect.com https://verify.walletconnect.org https://secure.walletconnect.com https://secure.walletconnect.org;" />
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self'; script-src 'self' 'wasm-unsafe-eval'; connect-src 'self' https: wss: http://localhost:* ws://localhost:*; img-src 'self' data: https: blob:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; font-src 'self' data: https://fonts.gstatic.com https://fonts.reown.com; object-src 'none'; base-uri 'self'; form-action 'self'; worker-src 'self' blob:; frame-src 'self' https://verify.walletconnect.com https://verify.walletconnect.org https://secure.walletconnect.org;"
+    />
     <meta name="referrer" content="strict-origin-when-cross-origin" />
     <title>Babylon - Vault</title>
 
--- a/services/vault/src/config/env.ts
+++ b/services/vault/src/config/env.ts
@@ -227,7 +227,7 @@ function validateEnvVars(): EnvValidationResult {
   // the blocking error modal.
   const FALLBACK_ETH_CHAIN_ID: EthChainId = ETH_SEPOLIA_CHAIN_ID;
   const FALLBACK_BTC_NETWORK: BtcNetworkName = "signet";
-  const FALLBACK_RPC_URL = "http://invalid.local";
+  const FALLBACK_RPC_URL = "https://invalid.local";
   try {
     configureBabylonConfig({
       ethChainId,
--- a/services/vault/vite.config.ts
+++ b/services/vault/vite.config.ts
@@ -17,6 +17,12 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 // injects, which has no 'unsafe-inline'/nonce and would white-screen `pnpm dev`. The meta CSP
 // does not govern that preamble because Vite injects it above the meta tag, and a meta policy
 // only applies to content that follows it.
+//
+// The meta CSP's connect-src carve-outs `http://localhost:*` and `ws://localhost:*` are
+// load-bearing for dev, not leftovers: ws://localhost is the Vite HMR websocket (a page's
+// 'self' does not reliably match the ws: scheme), and http://localhost:<port> covers the
+// cross-port Playwright mock servers (playwright.config.ts). Removing them breaks HMR and
+// the e2e suite with silent CSP blocks, not build errors.
 const SECURITY_HEADERS = {
   "X-Content-Type-Options": "nosniff",
   "X-Frame-Options": "DENY",


services/vault/index.html
```diff
@@ -26,7 +26,10 @@
     <meta name="twitter:image:width" content="2048" />
     <meta name="twitter:image:height" content="1170" />
     <meta name="version" content="%NEXT_PUBLIC_COMMIT_HASH%" />
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' 'wasm-unsafe-eval'; connect-src 'self' https: http: wss: ws:; img-src 'self' data: https: blob:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; font-src 'self' data: https:; object-src 'none'; base-uri 'self'; worker-src 'self' blob:; frame-src 'self' https://verify.walletconnect.com https://verify.walletconnect.org https://secure.walletconnect.com https://secure.walletconnect.org;" />
+    <meta
```
gbarkhatov, line 29: Suggestion (non-blocking): consider turning the maintenance note into a CI guard so the `font-src` / `frame-src` pins stay self-enforcing instead of tribal knowledge. A small test that greps the built `dist/` assets for `https://fonts.` and `(secure|verify)\.walletconnect\.` and asserts every host found is a subset of the pins here would fail the build when an AppKit/SDK bump introduces a new font or iframe host — which is the main silent-breakage risk. Caveat: it only catches *added* hosts and can't protect the un-greppable `fonts.gstatic.com` pin (see the `font-src` thread below), so it complements the reasoning rather than replacing it.

services/vault/index.html
```diff
@@ -26,7 +26,10 @@
...
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self'; script-src 'self' 'wasm-unsafe-eval'; connect-src 'self' https: wss: http://localhost:* ws://localhost:*; img-src 'self' data: https: blob:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; font-src 'self' data: https://fonts.gstatic.com https://fonts.reown.com; object-src 'none'; base-uri 'self'; form-action 'self'; worker-src 'self' blob:; frame-src 'self' https://verify.walletconnect.com https://verify.walletconnect.org https://secure.walletconnect.org;"
```
gbarkhatov, line 31: `font-src` note for future maintainers: the PR's maintenance note says to re-grep built assets for `https://fonts.` on an AppKit bump, but `fonts.gstatic.com` will never appear in that grep — it's pulled in at runtime by the Google Fonts `fonts.googleapis.com/css2?family=Poppins` stylesheet (governed by `style-src`), not referenced as a literal string in the bundle. Whoever tightens this next has to reason about it from the googleapis stylesheet, not expect to find it. Worth a one-line comment here so a future pass doesn't 'helpfully' delete the `fonts.gstatic.com` pin and silently break Poppins.

gbarkhatov, line 31: Minor / not blocking: because this meta CSP is static in `index.html` and shared across environments, the `http://localhost:*` / `ws://localhost:*` carve-outs also ship to mainnet. Harmless in practice (localhost is same-machine), but it's the kind of thing an auditor will flag again. The per-env `%VAR%` templating you already use here is the real fix — fine to defer to the edge/meta CSP convergence work as the description notes.

services/vault/src/config/env.ts
```diff
@@ -227,7 +227,7 @@ function validateEnvVars(): EnvValidationResult {
   // the blocking error modal.
   const FALLBACK_ETH_CHAIN_ID: EthChainId = ETH_SEPOLIA_CHAIN_ID;
   const FALLBACK_BTC_NETWORK: BtcNetworkName = "signet";
-  const FALLBACK_RPC_URL = "http://invalid.local";
+  const FALLBACK_RPC_URL = "https://invalid.local";
```
gbarkhatov, line 230: Confirmed correct: `.local` is a reserved TLD that won't resolve, and switching to `https://` keeps the sentinel inside `connect-src https:`, so a missing RPC URL now fails as a clean network error instead of a CSP violation — which is exactly what the blocking error-modal path above wants.

gbarkhatov (approved):
Approving. I independently verified the three load-bearing claims against the installed deps and test config:

- `frame-src` `.com` removal is safe — the only iframe host (`SECURE_SITE`) in both AppKit trees is `secure.walletconnect.org` (`appkit-common/ConstantsUtil.js`); the only `.com` occurrence is `secure-mobile.walletconnect.com` (mobile deep-link), which the old policy never frame-allowed either, so there's no regression.
- `font-src` pins are complete — font files resolve to `fonts.reown.com/KHTeka-*.woff2` and `fonts.gstatic.com` (via the googleapis Poppins/Inter stylesheet, which is `style-src`, not `font-src`); no other font host surfaced.
- The `http://localhost:*` / `ws://localhost:*` carve-outs match the Playwright mock ports (9996–9999) and the Vite HMR websocket.

Scope is honest but modest: `connect-src` still keeps full `https:`/`wss:` wildcards and `img-src https:` stays open, so this removes plaintext + arbitrary form/frame/font vectors, not arbitrary-host exfil over TLS. The PR is upfront about deferring per-env allowlists and the `/logo` sidecar constraint — reasonable given mainnet hosts live only in GitHub env vars.

One pre-existing gap worth tracking (not introduced here, out of scope for this PR): a meta CSP can't enforce `frame-ancestors`, so mainnet clickjacking protection depends entirely on the edge sending `X-Frame-Options` / `frame-ancestors` — the `X-Frame-Options: DENY` in `vite.config.ts` is dev-server only. Best handled alongside the edge/meta CSP convergence.

Inline notes are all non-blocking. Well-reasoned hardening.

jrwbabylonlab (approved):

PR #2128, opened 2026-07-23
feat(vault): god-mode  improvement for the v3 Loans and Activity

jeremy-babylonlabs:
Add `loan` and `activity` mock types so the panel drives the two v3 pages that had no injectable state. Loan mocks render as Active Loans rows (four scenarios: borrowable, repay-only, missing liquidity read, pending APR) with the summary totalling the rendered rows; activity mocks render in the feed (nine scenarios covering both flat log rows and the expandable liquidation group card), dated relative to now so the Today / Yesterday group headers apply.

Add Loans summary overrides for the states a real position cannot be talked into on demand: a health factor per production band, and the borrow-capacity cards in their loading and error states. The forced health-factor status is derived with the real banding function so it cannot drift from production.

Mount the panel once per route subtree (dev/GodModeMount) rather than a copy per page. Overview, Vaults, Loans and Activity now all get the debug sections as well, and the panel keeps its open/dragged position across navigation within a subtree. This replaces the per-page mounts added in #2124.

Restyle the liquidation-analysis and position-notification debug sections to the panel's zinc chrome; they were light-theme purple cards sitting inside a fixed dark box.

Mock rows are display-only: their ids and symbols resolve to no real reserve or event, both row actions stay disabled, nothing is polled or persisted, and the whole of src/dev is dropped from production builds.

### DIFF
--- a/services/vault/src/dev/GodModePanel.tsx
+++ b/services/vault/src/dev/GodModePanel.tsx
@@ -46,6 +46,7 @@ import {
 } from "./demoArtifactDownload";
 import {
   addDemoItem,
+  amountUnitFor,
   type DemoCta,
   type DemoItem,
   type DemoType,
@@ -79,12 +80,6 @@ const TYPE_LABELS: Record<DemoType, string> = {
   loan: "Loan",
   activity: "Activity",
 };
-
-/** Unit a mock item's amount is denominated in: BTC everywhere except loans,
- *  whose amount is the borrowed stablecoin. */
-function amountUnit(type: DemoType): string {
-  return type === "loan" ? "USDC" : "BTC";
-}
 
 const CTA_BADGE: Record<DemoCta, { label: string; className: string }> = {
   primary: { label: "Orange CTA", className: "bg-orange-500 text-white" },
@@ -257,7 +252,7 @@ function ItemRow({ item, index }: { item: DemoItem; index: number }) {
       </div>
 
       <label className="flex items-center justify-between gap-2 text-xs">
-        <span>Amount ({amountUnit(item.type)})</span>
+        <span>Amount ({amountUnitFor(item)})</span>
         <input
           type="number"
           min="0"
@@ -265,7 +260,7 @@ function ItemRow({ item, index }: { item: DemoItem; index: number }) {
           value={item.amount}
           onChange={(e) => setDemoItemAmount(item.key, e.target.value)}
           className="w-28 rounded border border-zinc-600 bg-zinc-800 px-2 py-1 text-xs"
-          aria-label={`Mock ${position} amount (${amountUnit(item.type)})`}
+          aria-label={`Mock ${position} amount (${amountUnitFor(item)})`}
         />
       </label>
 
--- a/services/vault/src/dev/GodModePanel.tsx
+++ b/services/vault/src/dev/GodModePanel.tsx
@@ -7,8 +7,9 @@
  * app — no cross-window plumbing).
  *
  * It is a controller only: it never renders the cards itself. It manages a list
- * of mock items (each a deposit / withdrawal / collateral at some state) that
- * render inside the REAL dashboard sections (see demoDeposit.ts).
+ * of mock items (each a deposit / withdrawal / collateral / loan at some state)
+ * that render inside the REAL dashboard, vaults and loans sections (see
+ * demoDeposit.ts). It is mounted once for those routes (see dev/GodModeMount).
  *
  * Chrome is intentionally theme-independent (fixed zinc colors) and inline, not
  * routed through copy.ts — none of it is shown to depositors.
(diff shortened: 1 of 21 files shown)


services/vault/src/router.tsx
```diff
@@ -101,7 +126,27 @@ const AaveOverlayLayout = () => {
...
+        The panel's full context dependency set is AaveConfig (mounted here),
+        ReorderOverride (mounted here) and ActivatingVaults (RootLayout).
+        Notably NOT PendingVaults: that context is reached only through
+        `useAaveVaults` and `useWithdrawCollateralTransaction`, neither of
+        which this subtree mounts — so the Aave layout's PendingVaultsProvider
+        is deliberately not duplicated here.
+
+        The page reads the indexer's activity list directly and does not
+        consume the demo store, so the panel drives theme, the protocol-status
+        / max-vaults overrides and the shared mock list here — no mock rows
```
gbarkhatov: This comment is now stale and contradicts the code. `Activity` → `useActivitiesWithPending` → `useDemoActivity()` merges demo rows into the returned feed, so mock activity rows **do** render on this page — including while disconnected. `Activity.tsx`'s own comment says exactly that ("demo rows are merged in by this hook … so the feed can be exercised without a wallet"), and the PR description lists nine activity scenarios rendering here.

The rationale above about which providers are / aren't mounted is correct and worth keeping — please just fix this last sentence so it states that Activity renders mock rows via the demo store, rather than the opposite.

services/vault/src/dev/GodModePanel.tsx
```diff
@@ -58,13 +65,27 @@ import {
...
   withdrawal: "Withdrawal",
   collateral: "Collateral",
+  loan: "Loan",
+  activity: "Activity",
 };
 
+/** Unit a mock item's amount is denominated in: BTC everywhere except loans,
+ *  whose amount is the borrowed stablecoin. */
+function amountUnit(type: DemoType): string {
+  return type === "loan" ? "USDC" : "BTC";
```
gbarkhatov: Dev-only nit: the amount field is labelled `Amount (BTC)` for `activity` mocks, but the `Borrow` and `Repay` activity scenarios ignore the amount arg and hard-code `DEMO_ACTIVITY_DEBT_AMOUNT` (USDC) — so for those two the control has no effect and the unit doesn't match the value the row shows. Deposit / withdraw / redeem / liquidation activity rows do use the amount as BTC, so the label is right for them. Low priority; if you want it tidy, derive the unit per activity scenario (and/or disable the field where it's inert). Not blocking.

gbarkhatov (dismissed):
Approving. This is a well-structured, thoroughly tested extension of the dev-only god-mode tooling, and it stays clear of every critical value-movement path in CLAUDE.md. Nice touches: the health-factor status is re-derived with the real banding function (can't drift from prod), the borrow-capacity snapshots keep a stable identity for `useSyncExternalStore`, forced summary states replace the live state wholesale, and mock rows are provably action-inert with non-colliding ids.

Two inline notes below — one worth fixing (a stale comment), one optional dev-only nit.

Two non-blocking observations, no change required:

1. **PR-description precision.** `useActivitiesWithPending.ts` and `Loans.tsx` statically import from `dev/demoDeposit`, so that module (and its `LOAN_SCENARIOS` / `ACTIVITY_SCENARIOS` data) sits in the production import graph and relies on Rollup DCE of the `import.meta.env.DEV` dead branches to be stripped. "The whole of `src/dev` is dropped from production builds" is strictly true only for the dynamically-imported `GodModeMount`; the statically-imported demo hooks are inert-but-present (same pre-existing pattern as `useDemoLoan` / `useDemoCollateral`). Functionally fine.

2. **`useActivitiesWithPending` `isLoading` flip (intended).** When connected with the live query still loading and activity mocks injected, `isLoading` flips to `false` immediately, suppressing the real feed's spinner so real rows appear afterward. This is deliberate and already documented in the code comment — noting only for visibility.

jrwbabylonlab (approved):

jonybur (approved):
Dev-tooling extension that stays properly gated. Every runtime read is behind `import.meta.env.DEV && featureFlags.isGodModePanelEnabled`, and the simulated-failure Error string in `DEBUG_BORROW_CAPACITY_SNAPSHOTS` is inside a `import.meta.env.DEV ? ... : null` ternary so it drops from production. The health-factor override runs through the real `getHealthFactorStatusFromValue` banding (with a dedicated test that the offered values land in three distinct bands), and the forced borrow-capacity state cleanly REPLACES the live loading/error rather than being merged field-by-field. Loan mocks are `displayOnly`, gating both borrow and repay buttons, and reserve ids use the `demo-reserve-` prefix so they can't collide with real decimal ids. Activity mocks use `demo-activity-` id prefix and dates relative to `Date.now()` with fake timers in the test. The router-level GodModeMount correctly avoids duplicating PendingVaultsProvider on /activity since that context isn't reached from that subtree.

PR #2129, opened 2026-07-23
fix(vault): verify Aave proxy contract on-chain to block indexer poisoning

gbarkhatov:
# Verify the Aave proxy contract on-chain (F8)

Closes https://github.com/babylonlabs-io/baby-auditor/issues/37

## What this fixes

When you click "Repay all" on a loan, the app needs to know exactly how much debt to clear. It reads that debt amount from your `proxyContract` — the address that holds your position on-chain — and that proxy address comes from an off-chain indexer (GraphQL). Until now we trusted the indexer's proxy as-is and sized the repayment straight from it.

That's a problem. If the indexer is compromised (or just wrong), it can return a **different** proxy — one with less debt. "Repay all" would then quote against the wrong debt, so you'd under-repay (your real debt stays, and you can't withdraw your collateral until it's cleared). Money can't be redirected — the transaction always settles against your own connected wallet — but the amount you sign for could be wrong.

## What we did

Before "Repay all" reads the debt, we now cross-check the indexer's proxy against the blockchain and refuse to proceed if they don't match ("fail closed"). The blockchain is the source of truth.

Concretely:

- Added a small shared guard, `assertProxyMatchesOnChain`, that reads the real proxy from the on-chain Aave adapter (`getPosition`) and throws a clear `ProxyMismatchError` if the indexer's value disagrees (case-insensitive compare). It caches the on-chain lookup per user so we don't repeat the call.
- Wired it into `repayAll` — the one path that sizes a signed transaction from the proxy's debt — right before the debt quote, and read the debt against the verified proxy.
- Added a user-facing message so a blocked repay reads as an integrity warning ("your position details don't match what's registered on-chain — refresh and try again"), not a generic failure.

This mirrors two guards already in the codebase that follow the same "check the indexer against the chain, fail closed" pattern (`assertReserveMatchesOnChain` for the token, and the reorder-flow membership check for vault IDs). We deliberately reuse the same style so reviewers and future maintainers see one consistent approach.

## Approaches we considered

- **Option A — verify once at the read chokepoint.** Resolve the real proxy on-chain in the live position query and use it for every read (health factor, borrow limits, notifications) as well as repay. Covers the whole surface in one place.
- **Option B — verify at the value-funding site.** Check the proxy on-chain right where "Repay all" sizes the signed transaction, and fail closed there.
- **Option C — a dedicated cached React hook for the verified proxy.** Rejected: the read path runs inside a data-fetching function and the repay path inside a click handler — neither consumes a hook cleanly, so this was the most plumbing for the worst fit.

## Why we chose Option B (repay-only)

The first revision of this PR did **A + B**. On review we scoped it down to **B only**, and removed the read/display-path check, because:

- The display figures don't sign anything. The two consequential actions are already anchored to on-chain truth independently of the indexer: **borrow** re-validates the health factor against a fresh on-chain read at sign time (`useBorrowTransaction`) and, more fundamentally, Aave enforces the real health factor on-chain — so a spoofed *displayed* number can't push a real borrow past the protocol's limit. **Repay all** is the one place that sizes a signed value from the proxy, and that's exactly what Option B guards.
- Verifying the display path added no real security bar but introduced a reliability regression: it turned a transient `getPosition` RPC blip — or normal indexer/chain lag on a freshly-deployed proxy — into a fail-closed **blank Loans page** (the query is polled every ~30s). That's a bad trade on a path that only shows numbers.

So the fix is deliberately minimal: one on-chain cross-check at the single site that funds a signed transaction.

## How it behaves

- Normal case (indexer agrees with the chain): "Repay all" works exactly as before; the only difference is one extra, cached on-chain read before the debt quote.
- Tampered/mismatched case: "Repay all" refuses with the integrity message — before the debt is read and before any transaction is built.
- The on-chain adapter address is the environment-pinned one, never an indexer-supplied address, so an attacker can't point us at a fake adapter that would "agree" with a fake proxy.

## Tests

- New `assertProxyMatchesOnChain` unit tests: match → returns the on-chain address; mismatch (attacker's healthier proxy) → throws; checksummed-vs-lowercase → passes; on-chain says "no position" while the indexer supplied one → throws (fail closed); RPC failure → propagates and doesn't poison the cache.
- `repayAll` tests: verifies the proxy on-chain before quoting debt; mismatch → throws before reading debt and builds no transaction; the debt quote (which sizes the approval) keys off the verified proxy, not the indexer's.
- Repay-hook test: a `ProxyMismatchError` surfaces as the integrity copy and the action returns without submitting.

Full Aave suite passes, typecheck and lint clean.

## Note

Rebased onto `main` to pick up the repay stale-RPC hardening (#2122), which renamed `repayFull` → `repayAll` and moved the debt read to `AaveProxy.getPositionReserveTotalDebt`; the F8 check is applied on top of that new structure.

### DIFF
--- a/services/vault/src/applications/aave/services/index.ts
+++ b/services/vault/src/applications/aave/services/index.ts
@@ -32,6 +32,11 @@ export {
 
 // On-chain integrity guards
 export {
+  ProxyMismatchError,
+  assertProxyMatchesOnChain,
+  resolveTrustedProxyContract,
+} from "./assertProxyMatchesOnChain";
+export {
   PositionChangedError,
   assertOptimalOrderMatchesOnChain,
   assertReorderBaseline,
--- a/services/vault/src/applications/aave/services/index.ts
+++ b/services/vault/src/applications/aave/services/index.ts
@@ -34,7 +34,6 @@ export {
 export {
   ProxyMismatchError,
   assertProxyMatchesOnChain,
-  resolveTrustedProxyContract,
 } from "./assertProxyMatchesOnChain";
 export {
   PositionChangedError,
--- a/services/vault/src/applications/aave/services/positionService.ts
+++ b/services/vault/src/applications/aave/services/positionService.ts
@@ -14,6 +14,7 @@ import {
 } from "../clients";
 import { hasDebtFromPosition } from "../utils";
 
+import { assertProxyMatchesOnChain } from "./assertProxyMatchesOnChain";
 import {
   fetchAaveActivePositionsWithCollaterals,
   type AavePosition,
@@ -85,6 +86,13 @@ export interface GetUserPositionsOptions {
    * Required for fetching collateral position data from Spoke.
    */
   vbtcReserveId: bigint;
+  /**
+   * Env-pinned Aave adapter address (from `getAaveAdapterAddress()`, never
+   * `config.coreSpokeAddress` or any indexer-derived address). Used to resolve
+   * the authoritative on-chain proxy and reject an indexer-spoofed one before
+   * any Spoke read runs against it.
+   */
+  trustedAdapterAddress: Address;
 }
 
 /**
@@ -119,7 +127,8 @@ export async function getUserPositionsWithLiveData(
   spokeAddress: Address,
   options: GetUserPositionsOptions,
 ): Promise<AavePositionWithLiveData[]> {
-  const { borrowableReserveIds, vbtcReserveId } = options;
+  const { borrowableReserveIds, vbtcReserveId, trustedAdapterAddress } =
+    options;
 
   // Fetch active positions with collaterals in a single GraphQL call
   const positions = await fetchAaveActivePositionsWithCollaterals(depositor);
@@ -130,7 +139,16 @@ export async function getUserPositionsWithLiveData(
 
   // User can only have one position in Babylon vault integration
   const position = positions[0];
-  const proxyAddress = position.proxyContract as Address;
+
+  // Verify the indexer-supplied proxy against the env-pinned adapter's
+  // on-chain position before any Spoke read keys on it. Fails closed on
+  // disagreement (a spoofed proxy would corrupt every risk figure below).
+  // Downstream consumers inherit the verified address via `proxyContract`.
+  const proxyAddress = await assertProxyMatchesOnChain(
+    trustedAdapterAddress,
+    depositor as Address,
+    position.proxyContract as Address,
+  );
 
   // One multicall for both live reads (vBTC collateral position + aggregate
   // account data) instead of two parallel `eth_call`s.
@@ -165,6 +183,9 @@ export async function getUserPositionsWithLiveData(
   return [
     {
       ...position,
+      // Overwrite the indexer-supplied proxy with the on-chain-verified one so
+      // every downstream consumer (e.g. the repay `proxyContract` prop) uses it.
+      proxyContract: proxyAddress,
       liveData: {
         drawnShares: spokePosition.drawnShares,
         premiumShares: spokePosition.premiumShares,
--- a/services/vault/src/applications/aave/services/positionService.ts
+++ b/services/vault/src/applications/aave/services/positionService.ts
@@ -14,7 +14,6 @@ import {
 } from "../clients";
 import { hasDebtFromPosition } from "../utils";
 
-import { assertProxyMatchesOnChain } from "./assertProxyMatchesOnChain";
 import {
   fetchAaveActivePositionsWithCollaterals,
   type AavePosition,
@@ -86,13 +85,6 @@ export interface GetUserPositionsOptions {
    * Required for fetching collateral position data from Spoke.
    */
   vbtcReserveId: bigint;
-  /**
-   * Env-pinned Aave adapter address (from `getAaveAdapterAddress()`, never
-   * `config.coreSpokeAddress` or any indexer-derived address). Used to resolve
-   * the authoritative on-chain proxy and reject an indexer-spoofed one before
-   * any Spoke read runs against it.
-   */
-  trustedAdapterAddress: Address;
 }
 
 /**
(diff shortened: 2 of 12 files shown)


gbarkhatov:
@greptile review

services/vault/src/applications/aave/services/positionService.ts
```diff
@@ -130,7 +139,16 @@ export async function getUserPositionsWithLiveData(
...
+
+  // Verify the indexer-supplied proxy against the env-pinned adapter's
+  // on-chain position before any Spoke read keys on it. Fails closed on
+  // disagreement (a spoofed proxy would corrupt every risk figure below).
+  // Downstream consumers inherit the verified address via `proxyContract`.
+  const proxyAddress = await assertProxyMatchesOnChain(
+    trustedAdapterAddress,
+    depositor as Address,
+    position.proxyContract as Address,
+  );
```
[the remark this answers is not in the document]
gbarkhatov: Thanks for flagging — this one is safe as-is, because `spokeAddress` here is **not** indexer-derived.

It comes from `config.coreSpokeAddress`, which `fetchAaveAppConfig` resolves **on-chain**, not from GraphQL:

- `services/vault/src/applications/aave/services/fetchConfig.ts` (L213–231): the indexer's `adapterAddress` is equality-checked against the env-pinned `getAaveAdapterAddress()` and the whole config is rejected on mismatch — then `coreSpokeAddress = await getCoreSpokeAddress(adapterAddress)`.
- `services/vault/src/applications/aave/clients/transaction.ts` (L38–48): `getCoreSpokeAddress` reads the **immutable** `BTC_VAULT_CORE_SPOKE` property directly off that env-pinned adapter.

So the GraphQL-supplied spoke is never used. These reads key off the exact same on-chain-derived spoke that `repayFull` and `assertReserveMatchesOnChain` independently compute — a poisoned indexer cannot substitute a malicious Spoke-compatible address here. This on-chain resolution has been in place since #1456.

The thing that makes this look exploitable is a stale comment in `assertReserveMatchesOnChain` that describes `config.coreSpokeAddress` as "indexer-supplied" (it stopped being so in #1456). Correcting that comment (and the copy of it in the new `assertProxyMatchesOnChain`) in this PR so it doesn't mislead again.

Resolving.

services/vault/src/applications/aave/services/positionService.ts
```diff
@@ -130,7 +139,16 @@ export async function getUserPositionsWithLiveData(
 
   // User can only have one position in Babylon vault integration
   const position = positions[0];
-  const proxyAddress = position.proxyContract as Address;
+
+  // Verify the indexer-supplied proxy against the env-pinned adapter's
+  // on-chain position before any Spoke read keys on it. Fails closed on
+  // disagreement (a spoofed proxy would corrupt every risk figure below).
+  // Downstream consumers inherit the verified address via `proxyContract`.
+  const proxyAddress = await assertProxyMatchesOnChain(
```
jrwbabylonlab: The `repayFull` rebind is the correct fix, but adding proxy verification here is over-scoped for F8. This runs on the live position query (`useAaveUserPosition`, polled every 30s) whose consumers — health factor, max-borrow, notifications — are display-only. F8's only consequential path is the `repayFull` debt quote; a spoofed proxy can't redirect settlement (adapter resolves from `borrower`) or over-repay (contract caps + refunds). Verifying the display path adds no security bar but turns a transient `getPosition` RPC failure into a fail-closed dependency that can blank the whole Loans page. Suggest cutting the display-path check (`trustedAdapterAddress` threading + this call + the overwrite) and keeping only the `repayFull` verification. (Independently confirmed by a second review.)

services/vault/src/applications/aave/services/index.ts
```diff
@@ -32,6 +32,11 @@ export {
 } from "./positionTransactions";
 
 // On-chain integrity guards
+export {
+  ProxyMismatchError,
+  assertProxyMatchesOnChain,
+  resolveTrustedProxyContract,
```
jrwbabylonlab: `resolveTrustedProxyContract` is exported but has no external consumer (only used internally by `assertProxyMatchesOnChain`). Drop the export (zero-dead-code).

jonybur (dismissed):
Solid F8 hardening. Verified the two things this needed to get right: the guard resolves the on-chain proxy via `getPosition` against the env-pinned adapter (never the indexer's), and it fails closed on both the missing-position case and RPC failures (the `.catch` also deletes the cache entry so a transient failure doesn't poison future reads). The position service applies the check at the chokepoint AND overwrites `position.proxyContract` with the verified value so downstream consumers can't accidentally re-read the indexer one; `repayFull` re-verifies at the value-site as an independent check. The Spoke read is safe too, though the PR doesn't advertise it: `config.coreSpokeAddress` is itself resolved on-chain in `fetchConfig` from the same pinned adapter. Case-insensitive compare is correct for checksummed-vs-lowercase inputs, and the per-user cache dedupe keeps the extra RPC at one call. Tests cover the match, mismatch, missing-position, RPC-failure, and repay-amount-byte-identity cases.

services/vault/src/applications/aave/services/positionService.ts
gbarkhatov: Agreed — done. Cut the display-path check entirely: removed the `trustedAdapterAddress` threading, the read-site `assertProxyMatchesOnChain` call, and the proxy overwrite in `getUserPositionsWithLiveData` / `useAaveUserPosition` (and reverted the read-path test). The verification now runs **only** at `repayAll` — the sole value-funding site, where a spoofed proxy would actually corrupt a signed amount. Borrow/display stay safe via their existing on-chain-anchored guards (borrow re-validates HF at sign time + Aave enforces on-chain), so no coverage is lost, and we no longer risk a transient `getPosition` failure blanking the Loans page.

Also rebased onto `main` so this sits on the #2122 repay refactor (`repayFull` → `repayAll`, debt via `AaveProxy.getPositionReserveTotalDebt`). Thanks for the catch — and for the second-review confirmation.

services/vault/src/applications/aave/services/index.ts
gbarkhatov: Done — `resolveTrustedProxyContract` is now a module-internal helper (dropped the `export` and removed the line from `services/index.ts`). Only `assertProxyMatchesOnChain` consumes it.

jrwbabylonlab (approved):

PR #2130, opened 2026-07-23
fix(vault): align v3 activity, loan and modal 

jeremy-babylonlabs:
## Screenshot
Screenshot 2026-07-23 at 22 01 42
Screenshot 2026-07-23 at 22 01 39

Activity page:
- Move the "Show all" filter to the left of the row and drop the Aave avatar beside it; the dropdown opens bottom-start so its 200px menu no longer runs off-screen from a left-aligned trigger.
- Give each row its own card, the same one the Vaults tab uses, instead of bundling a whole date group into one divider-separated card. The grouped treatment is now reserved for a liquidation's child events, which is the only place the design uses it.
- Rebuild the row on the Figma column spec: 180px cells, body-2 text, 12px status dot with a 4px gap, timestamp in the secondary colour.
- Replace copy-to-clipboard on the row hash with a single open-in-explorer link (ActivityHashLink). CopyableHash is unchanged and still used everywhere else; the new component is reached only from the v3 rows.
- Offer the HTLC refund on an expired deposit's row. The button reuses the Vaults inactive-row gating, so it appears only while the refund is actually available and renders disabled-with-tooltip while blocked. Rows matching a refundable deposit report "Expired" rather than "Pending".

Loans page:
- Rebuild the active-loan card on the shared row card. The metrics group now shrinks before the actions do, so Borrow / Repay stay on the row instead of wrapping underneath.

Full-screen modals:
- Add FullScreenModal, the base every full-screen surface now goes through. It supplies the top bar the success screens had lost entirely — close on the left, network badge and settings on the right — and insets it to the 120px the design's shared Header uses, correcting a bar that was flush at 16px on every modal.

Status dots resolve to the v3 foundation palette by hex, reusing the existing --risk-* channels; the liquidation orange had no token, so --risk-orange joins them. The row card class the Vaults sections had duplicated moves to layoutClasses so the three lists cannot drift.

### DIFF
--- a/services/vault/src/applications/aave/components/LoanCard/LoanSuccessModal.tsx
+++ b/services/vault/src/applications/aave/components/LoanCard/LoanSuccessModal.tsx
@@ -1,11 +1,6 @@
-import {
-  Avatar,
-  Button,
-  FullScreenDialog,
-  Heading,
-  Text,
-} from "@babylonlabs-io/core-ui";
+import { Avatar, Button, Heading, Text } from "@babylonlabs-io/core-ui";
 
+import { V3ModalShell } from "@/components/shared/V3ModalShell";
 import { COPY } from "@/copy";
 import { formatAmount } from "@/utils/formatting";
 
@@ -29,9 +24,9 @@ const COPY_BY_VARIANT = {
 /**
  * Full-screen success screen shown after a successful borrow or repay. The
  * layout is identical for both operations; only the copy differs, selected by
- * `variant`. Confirms the amount and dismisses via the "Done" CTA — the sole
- * control. `onClose` is intentionally withheld from the dialog so there is no
- * close (X), backdrop, or escape dismissal competing with the single button.
+ * `variant`. Confirms the amount and dismisses via the "Done" CTA, the close
+ * (X), the backdrop, or escape — all four land on the same `onDone`, since
+ * there is nothing left to cancel once the transaction has settled.
  */
 export function LoanSuccessModal({
   open,
@@ -46,7 +41,7 @@ export function LoanSuccessModal({
   const formattedAmount = formatAmount(amount, decimals);
 
   return (
-    <FullScreenDialog open={open} className="items-center justify-center p-6">
+    <V3ModalShell open={open} onClose={onDone}>
       <div className="mx-auto flex w-full max-w-[564px] flex-col gap-10 rounded-3xl border border-secondary-strokeLight px-6 pb-6 pt-[72px] text-center text-accent-primary">
         <div className="flex flex-col items-center gap-6">
           <Avatar
@@ -84,6 +79,6 @@ export function LoanSuccessModal({
           {copy.doneButton}
         </Button>
       </div>
-    </FullScreenDialog>
+    </V3ModalShell>
   );
 }
--- a/services/vault/src/components/Activity/ActivityHashLink.tsx
+++ b/services/vault/src/components/Activity/ActivityHashLink.tsx
@@ -1,0 +1,54 @@
+/**
+ * ActivityHashLink — the v3 activity row's transaction-hash cell.
+ *
+ * The v3 design drops copy-to-clipboard in favour of a single open-in-explorer
+ * affordance: the truncated hash, underlined, followed by an external-link
+ * arrow (Figma 10233:50689). The whole thing is one link. Rows outside the v3
+ * activity list keep using CopyableHash.
+ *
+ * BTC hashes are displayed without the 0x prefix (Bitcoin convention); EVM
+ * hashes keep it.
+ */
+
+import { stripHexPrefix } from "@babylonlabs-io/ts-sdk/tbv/core";
+import { RiArrowRightUpLine } from "react-icons/ri";
+
+import type { HashChain } from "@/components/shared/CopyableHash";
+import { COPY } from "@/copy";
+import { truncateHash } from "@/utils/addressUtils";
+
+const EXTERNAL_ICON_SIZE = 16;
+
+interface ActivityHashLinkProps {
+  /** Raw hash (may have 0x prefix) */
+  hash: string;
+  /** Source chain — determines whether the 0x prefix is stripped for display */
+  chain: HashChain;
+  /** Explorer URL the row opens in a new tab. */
+  explorerUrl: string;
+}
+
+export function ActivityHashLink({
+  hash,
+  chain,
+  explorerUrl,
+}: ActivityHashLinkProps) {
+  const truncated = truncateHash(chain === "BTC" ? stripHexPrefix(hash) : hash);
+
+  return (
+    <a
+      href={explorerUrl}
+      target="_blank"
+      rel="noopener noreferrer"
+      aria-label={COPY.activity.viewTransaction(chain, truncated)}
+      className="flex items-center gap-2 text-sm leading-[1.43] tracking-[0.17px] text-accent-primary underline transition-colors hover:text-accent-secondary"
+    >
+      {truncated}
+      <RiArrowRightUpLine
+        size={EXTERNAL_ICON_SIZE}
+        aria-hidden
+        className="shrink-0 no-underline"
+      />
+    </a>
+  );
+}
--- a/services/vault/src/components/Activity/ActivityHashLink.tsx
+++ b/services/vault/src/components/Activity/ActivityHashLink.tsx
@@ -41,13 +41,15 @@ export function ActivityHashLink({
       target="_blank"
       rel="noopener noreferrer"
       aria-label={COPY.activity.viewTransaction(chain, truncated)}
-      className="flex items-center gap-2 text-sm leading-[1.43] tracking-[0.17px] text-accent-primary underline transition-colors hover:text-accent-secondary"
+      className="flex items-center gap-2 text-sm leading-[1.43] tracking-[0.17px] text-accent-primary transition-colors hover:text-accent-secondary"
     >
-      {truncated}
+      {/* The underline sits on the text alone: on the anchor it propagates to
+          the arrow, and a descendant `no-underline` does not lift it. */}
+      <span className="underline">{truncated}</span>
       <RiArrowRightUpLine
         size={EXTERNAL_ICON_SIZE}
         aria-hidden
-        className="shrink-0 no-underline"
+        className="shrink-0"
       />
     </a>
   );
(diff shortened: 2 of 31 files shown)


services/vault/src/components/pages/Activity.tsx
```diff
@@ -36,6 +41,54 @@ export default function Activity() {
...
+  // keyed by vault id. Correlate these against a row's `vaultId` and never its
+  // `id`: an indexed row's id is its event, not its vault (see ActivityLog).
+  const refundableVaultIds = useMemo(
+    () => new Set<string>(expiredActivities.map((a) => a.id)),
+    [expiredActivities],
+  );
+
+  // Such a row arrives from localStorage still marked pending; the contract
+  // status says otherwise, so correct it before it reaches the status column.
+  const rows = useMemo<ActivityRow[]>(
```
jrwbabylonlab: This forces `isExpired: true` on refundable rows, and `usePendingDeposits` + the polling-provider scaffolding mount, in v2 too — not just v3 (no `isV3UiEnabled` gate). So a refundable-expired deposit that rendered "Pending" in v2 now renders "Expired", and v2 mounts the modals dead. Looks like an intended improvement, but the PR is framed as v3 alignment — confirm the v2 behavior change is intended, or gate it.

services/vault/src/components/Activity/ActivityList.tsx
```diff
@@ -112,15 +113,12 @@ export function ActivityList({ activities, isConnected }: ActivityListProps) {
             </Heading>
           )}
           {isConnected && (
```
jrwbabylonlab, line 119: The Aave logo next to the filter was removed for all modes, so the v2 Activity header loses it too (its two v2 tests were deleted, so this reads as deliberate). Same note: confirm the v2 change is intended given the PR is scoped as v3 alignment.

jonybur (dismissed):
The v3 activity/loan/modal consolidation reads cleanly. The refundable-vault correlation keys on `vaultId` (never row `id`) at both the status flip and the Withdraw slot, with a test that pins down the two are separate id spaces. `ExpiredWithdrawButton` correctly reuses the Vaults inactive-row gating, so the button only surfaces when the refund is actually available. The `Activity.tsx` guard that renders the bare feed when `refundableVaultIds.size === 0` avoids holding the disconnected empty state behind the ProtocolParamsProvider read. `V3ModalShell` migration and the shared `ListRowCard` / `CARD_SHELL_CLASS` refactors don't drift chrome across the three lists.

services/vault/src/components/Activity/ActivityList.tsx
jeremy-babylonlabs: good point, fixing now

jeremy-babylonlabs: good point, fixing now

services/vault/src/components/Activity/ExpiredWithdrawButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+  vaultId: string;
+  onWithdraw: (vaultId: string) => void;
+}
+
+export function ExpiredWithdrawButton({
+  vaultId,
+  onWithdraw,
+}: ExpiredWithdrawButtonProps) {
+  const result = useDepositPollingResult(vaultId);
+  const actionStatus: ReturnType<typeof getActionStatus> = result
```
gbarkhatov: This available/blocked computation is a verbatim copy of `InactiveRow` in `VaultsLifecycleSections.tsx` — the `type === "available"` + `REFUND_HTLC` check, and the `disabled` + `REFUND_HTLC` + `!isRefundInFlightOrSettled(peginState)` guard. Two copies of a refund-adjacent gating decision will drift. Consider extracting a shared hook, e.g. `useRefundRowAction(vaultId)` returning `{ available, blockedTooltip }`, and consuming it from both rows.

services/vault/src/components/shared/buttonClasses.ts
```diff
@@ -14,6 +14,10 @@ export const PRIMARY_BUTTON_CLASS =
 export const NEUTRAL_ROW_BUTTON_CLASS =
   "ml-auto flex h-9 min-w-[120px] shrink-0 items-center justify-center whitespace-nowrap rounded-lg bg-secondary-strokeLight px-4 text-sm leading-[1.43] tracking-[0.17px] text-accent-primary transition-[filter] enabled:hover:brightness-110 disabled:cursor-not-allowed disabled:text-accent-disabled";
 
+/** 36px-tall destructive (red) action button — the expired deposit's Withdraw. */
+export const ERROR_ROW_BUTTON_CLASS =
```
gbarkhatov, line 21: Two things on this new constant:

1. **Color consistency (question):** the Activity-page Withdraw uses this red class, but the Vaults-page Withdraw for the *same* HTLC refund uses the orange `PRIMARY_ROW_BUTTON_CLASS`. Is the red intentional per Figma? One protocol action in two colors across two pages reads as inconsistent unless the design calls for it.
2. **Dead styles (nit):** the button only ever renders enabled (the blocked state falls back to `NEUTRAL_ROW_BUTTON_CLASS`), so `disabled:cursor-not-allowed disabled:opacity-30` here are never exercised.

services/vault/src/components/Activity/__tests__/ActivityList.test.tsx
```diff
@@ -23,6 +23,23 @@ vi.mock("@/components/Wallet", () => ({
   Connect: () => <button type="button">Connect</button>,
 }));
 
+// The real button reads the deposit-polling context to decide whether the
+// refund is available — that gating is the Vaults page's, covered there. These
+// tests are about which row the list hands the action to, and with which id.
+vi.mock("../ExpiredWithdrawButton", () => ({
```
gbarkhatov, line 29: Mocking `ExpiredWithdrawButton` out here (and in `Activity.test.tsx`) leaves its available/blocked/null gating with no coverage of its own — the PR leans on "covered on the Vaults page," but that's a separate copy of the logic (see the duplication note on the component). A single direct test asserting the three branches would lock it down, or it can be folded in when the shared hook is extracted.

services/vault/src/components/Activity/ActivityHashLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+      target="_blank"
+      rel="noopener noreferrer"
+      aria-label={COPY.activity.viewTransaction(chain, truncated)}
+      className="flex items-center gap-2 text-sm leading-[1.43] tracking-[0.17px] text-accent-primary underline transition-colors hover:text-accent-secondary"
+    >
+      {truncated}
+      <RiArrowRightUpLine
+        size={EXTERNAL_ICON_SIZE}
+        aria-hidden
+        className="shrink-0 no-underline"
```
gbarkhatov: `no-underline` on the icon is effectively a no-op: the underline is `text-decoration` on the ancestor `<a>` (line 44), and in most browsers a descendant `text-decoration: none` doesn't lift the underline the ancestor propagates under it. If the intent is an un-underlined arrow, move `underline` off the anchor onto an inner `<span>` wrapping only `{truncated}`.

services/vault/src/components/Activity/FilterDropdown.tsx
```diff
@@ -52,7 +52,7 @@ export function FilterDropdown<V extends string>({
       <Popover
         open={open}
         anchorEl={anchorRef.current}
-        placement="bottom-end"
+        placement="bottom-start"
```
gbarkhatov: This `bottom-end` → `bottom-start` flip is unconditional, but `FilterDropdown` is shared and v2 still renders the trigger at the right edge (`justify-between`) — so this is a v2 behavior change in a PR whose commit 4 was specifically about not touching v2. In practice core-ui's `Popover` leaves popper's `flip`/`preventOverflow` defaults active (it only adds an offset modifier), so the 200px menu stays on-screen and it's cosmetic — but worth a conscious call, or gating on `isV3`.

services/vault/src/applications/aave/components/LoanCard/LoanSuccessModal.tsx
```diff
@@ -46,7 +41,7 @@ export function LoanSuccessModal({
   const formattedAmount = formatAmount(amount, decimals);
 
   return (
-    <FullScreenDialog open={open} className="items-center justify-center p-6">
+    <V3ModalShell open={open} onClose={onDone}>
```
gbarkhatov, line 44: Passing `onClose={onDone}` now makes this success screen dismissible via the X, backdrop, and Escape — reversing the previous "onClose intentionally withheld, Done is the sole control" behavior (the old comment was emphatic about it). The PR body frames this as intended with the top bar returning; just confirm design signed off on the extra dismissal paths all landing on `onDone`.

services/vault/src/components/shared/layoutClasses.ts
```diff
@@ -22,3 +35,12 @@ export const PAGE_CONTENT_CLASS = featureFlags.isV3UiEnabled
...
 /** Width + vertical padding for the dashboard section summary cards. */
 export const SUMMARY_CARD_CLASS = "w-full border-0 !py-[34px]";
+
+/**
+ * Chrome of a v3 list card — border, radius, surface — with no padding or
+ * layout of its own. For a single row prefer `ListRowCard`, which adds that
+ * row's padding and flex; this is for a card holding several
+ * divider-separated rows, where the padding belongs to the card.
+ */
+export const CARD_SHELL_CLASS =
```
gbarkhatov, line 45: `CARD_SHELL_CLASS` hardcodes the same `border-secondary-strokeLight bg-secondary-highlight … dark:bg-[#202020]` chrome that `ListRowCard` sets inline. Given this PR's stated aim ("the three lists cannot drift"), `ListRowCard` could compose `CARD_SHELL_CLASS` so the surface tokens live in one place. Optional.

gbarkhatov (dismissed):
Nice, well-isolated v3 UI-alignment pass. The `id`-vs-`vaultId` correlation is correct (verified `VaultActivity.id` is the derived vault id and `useRefundModal.handleRefundClick` looks up by that same key), the deposit lifecycle is mounted only on v3 and only when a refund is actually outstanding, and none of the CLAUDE.md critical value-movement paths are touched — the refund still flows through the existing `RefundModal`/`vaultRefundService`, so this only gates button visibility and does display-only status reconciliation. Approving.

Inline notes below are mostly questions/nits. The two I'd weigh before merge:
- Refund-gating logic is duplicated verbatim between `ExpiredWithdrawButton` and `InactiveRow` — extract a shared hook.
- Confirm the red-vs-orange color choice for the same Withdraw action across the two pages is intentional.

Everything else (test coverage for the new button, the v2 `FilterDropdown` placement leak, `LoanSuccessModal` becoming dismissible, `CARD_SHELL_CLASS`/`ListRowCard` duplication, cosmetic `no-underline`) is lower-stakes.

Nit on the PR description: it says "Add FullScreenModal" but the component is `V3ModalShell` — body text is slightly out of date.

gbarkhatov (approved):

PR #2131, opened 2026-07-23
fix(vault): stop two expected conditions from flooding Sentry

jonybur:
## Context

A Discover export of the `tbv` Sentry project showed **~60% of all events** were two expected, non-actionable conditions - drowning the depositor-funnel signal the Sentry epic just built across five phases. With alerts about to be configured, that noise is alert-fatigue fuel. This is the companion cleanup: build the signal, then clear the channel.

## The two sources (traced, not guessed)

**1. `OrdinalsClassifierUnavailableError` (~33% of events).** The root cause isn't the ordinals code - it's [queryClient.ts](services/vault/src/config/queryClient.ts): `QueryCache.onError` `logger.error`s **every** React Query failure, turning any blip into a captured issue. The wallet-connector ordinals hook throws this whenever no `ordinalsApiUrl` is configured (devnet/testnet), and it's retried up to 3×, so every deposit-form mount produces several captured errors. The vault already treats missing ordinals data as non-fatal (all UTXOs available, see `useUTXOs`).

**2. `Chainlink price data is stale …` (~30%).** A `logger.event` at [chainlink/query.ts](services/vault/src/clients/eth-contract/chainlink/query.ts#L259) fired on **every** price read while a feed was stale - dozens of identical events per stale episode (devnet feeds are permanently stale).

## The fix

- **Query firehose**: known-expected query conditions (matched by `error.name`, so the match survives bundling across the package boundary) route to a **breadcrumb** instead of a captured error. They still attach to any genuine error captured afterwards - not erased, just not standalone issues. Starts with `OrdinalsClassifierUnavailableError`; the list is trivially extensible.
- **Chainlink stale**: **deduped to one event per stale episode per feed**, cleared when the feed reports fresh again (so a later stale episode re-alerts). The `metadata.isStale` flag is untouched, so the UI's stale badge is unaffected.

## Why not just `ignoreErrors`

Both are **legitimate signals on mainnet** - a stale price feed means bad collateral valuation; a missing ordinals classifier means the app can't filter inscription UTXOs (risk of spending an ordinal). A blunt global drop would erase them everywhere. Downgrading / deduping keeps the mainnet signal while removing only the expected-condition flood from environments that produce it. That's the difference between reducing noise and going blind.

## Testing

- `queryClient.test.ts` (5): expected error → breadcrumb not capture, genuine error → capture with the right context tag (Query/Mutation), and the by-`name` match across a distinct class identity.
- `chainlink/query.test.ts` (+3): stale event emitted once per episode, re-emitted after recovery→stale, and `metadata.isStale` still true on a deduped read. Existing stale-logging tests updated to reset the module-scoped dedup store.
- Full vault suite green: **247 files, 2612 tests**. eslint clean; `tsc -p tsconfig.lib.json` clean.

Note: this is independent of the funnel epic and of the still-pending testnet release - it improves the signal in every environment that already transmits (devnet today, testnet once promoted).

### DIFF
--- a/services/vault/src/clients/eth-contract/chainlink/query.ts
+++ b/services/vault/src/clients/eth-contract/chainlink/query.ts
@@ -202,6 +202,21 @@ function emitForSymbol(
 }
 
 /**
+ * Feeds already reported stale this session. A stale price feed is a legitimate
+ * signal (bad collateral valuation), but every component that reads the price
+ * re-runs this, so an undeduped event fires dozens of times per stale episode.
+ * Emit once when a feed goes stale and clear the flag when it reports fresh
+ * again, so a later stale episode re-alerts. Module-scoped — the signal is about
+ * the feed, not any caller.
+ */
+const reportedStaleFeeds = new Set<string>();
+
+/** Test-only: the dedup store outlives any single call. */
+export function resetStaleFeedReporting(): void {
+  reportedStaleFeeds.clear();
+}
+
+/**
  * Translate one feed's raw multicall results into a price + metadata, or an
  * error metadata entry if either call failed or the price is invalid.
  */
@@ -256,17 +271,25 @@ function readoutForFeed(
   const ageSeconds = Math.floor(Date.now() / 1000) - Number(updatedAt);
   const isStale = !isPriceFresh(roundData);
 
+  const feedKey = feedAddress.toLowerCase();
   if (isStale) {
-    if (answeredInRound < roundId) {
-      logger.event(
-        `Chainlink price data is stale: incomplete round (answeredInRound=${answeredInRound} < roundId=${roundId}). Using last known price.`,
-      );
-    } else {
-      const ageHours = (ageSeconds / SECONDS_PER_HOUR).toFixed(1);
-      logger.event(
-        `Chainlink price data is stale (${ageHours} hours old). Using last known price.`,
-      );
+    // One event per stale episode; subsequent reads while still stale are silent.
+    if (!reportedStaleFeeds.has(feedKey)) {
+      reportedStaleFeeds.add(feedKey);
+      if (answeredInRound < roundId) {
+        logger.event(
+          `Chainlink price data is stale: incomplete round (answeredInRound=${answeredInRound} < roundId=${roundId}). Using last known price.`,
+        );
+      } else {
+        const ageHours = (ageSeconds / SECONDS_PER_HOUR).toFixed(1);
+        logger.event(
+          `Chainlink price data is stale (${ageHours} hours old). Using last known price.`,
+        );
+      }
     }
+  } else {
+    // Recovered — allow the next stale episode for this feed to re-alert.
+    reportedStaleFeeds.delete(feedKey);
   }
 
   return {
--- a/services/vault/src/config/queryClient.ts
+++ b/services/vault/src/config/queryClient.ts
@@ -22,24 +22,49 @@ const shouldRetry = (failureCount: number, error: Error): boolean => {
   return true;
 };
 
-const logError = (error: Error, context?: string): void => {
-  const errorContext = context ? `[${context}]` : "";
+/**
+ * Query errors that reflect an expected, non-actionable condition rather than a
+ * fault — e.g. an optional API not configured in this environment. Recorded as
+ * a breadcrumb (so they still attach to any genuine error captured afterwards)
+ * instead of a standalone Sentry issue, so they don't drown the signal. This
+ * onError handler fires for EVERY query failure, so an unfiltered logger.error
+ * here turns any transient blip into a captured issue. Matched by error name so
+ * the match survives bundling across the wallet-connector package boundary.
+ */
+const EXPECTED_QUERY_ERROR_NAMES = new Set<string>([
+  // The wallet-connector ordinals hook throws this when inscription filtering
+  // has no data source (no ordinalsApiUrl configured). The vault already treats
+  // missing ordinals data as non-fatal (every UTXO available, see useUTXOs), so
+  // it is an expected environment condition, not a captured error.
+  "OrdinalsClassifierUnavailableError",
+]);
+
+export function isExpectedQueryError(error: Error): boolean {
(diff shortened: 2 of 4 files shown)


jrwbabylonlab (approved):

services/vault/src/clients/eth-contract/chainlink/query.ts
```diff
@@ -256,17 +271,25 @@ function readoutForFeed(
...
-      logger.event(
-        `Chainlink price data is stale: incomplete round (answeredInRound=${answeredInRound} < roundId=${roundId}). Using last known price.`,
-      );
-    } else {
-      const ageHours = (ageSeconds / SECONDS_PER_HOUR).toFixed(1);
-      logger.event(
-        `Chainlink price data is stale (${ageHours} hours old). Using last known price.`,
-      );
+    // One event per stale episode; subsequent reads while still stale are silent.
+    if (!reportedStaleFeeds.has(feedKey)) {
```
gbarkhatov, line 277: Both stale message variants (incomplete-round and age-based) share this single dedup slot per feed, so a feed transitioning stale-incomplete → stale-age without going fresh in between emits only the first variant. Correct and intended — it's the same stale episode — just noting the collapse for future readers.

services/vault/src/clients/eth-contract/chainlink/query.ts
```diff
@@ -256,17 +271,25 @@ function readoutForFeed(
...
+      reportedStaleFeeds.add(feedKey);
+      if (answeredInRound < roundId) {
+        logger.event(
+          `Chainlink price data is stale: incomplete round (answeredInRound=${answeredInRound} < roundId=${roundId}). Using last known price.`,
+        );
+      } else {
+        const ageHours = (ageSeconds / SECONDS_PER_HOUR).toFixed(1);
```
gbarkhatov, line 284: Because the event fires once per episode, the reported age is the age at first detection, not the peak — the Sentry event no longer reflects how long a feed has stayed stale. Fine for noise reduction, and `metadata.ageSeconds` still carries the live age if on-call ever needs it.

services/vault/src/clients/eth-contract/chainlink/query.ts
```diff
@@ -256,17 +271,25 @@ function readoutForFeed(
...
+        logger.event(
+          `Chainlink price data is stale (${ageHours} hours old). Using last known price.`,
+        );
+      }
     }
+  } else {
+    // Recovered — allow the next stale episode for this feed to re-alert.
+    reportedStaleFeeds.delete(feedKey);
```
gbarkhatov, line 292: Worth a one-line note here that the flag only clears on a successful *fresh* read: `readoutForFeed` early-returns for a multicall failure and for `answer <= 0n` / out-of-range before reaching this branch, so a `stale → (RPC failure or invalid answer) → stale` sequence won't re-alert. Low impact (the feed is still genuinely degraded), but a future reader could otherwise assume any recovery re-arms the alert.

services/vault/src/clients/eth-contract/chainlink/query.ts
```diff
@@ -201,6 +201,21 @@ function emitForSymbol(
...
+ * signal (bad collateral valuation), but every component that reads the price
+ * re-runs this, so an undeduped event fires dozens of times per stale episode.
+ * Emit once when a feed goes stale and clear the flag when it reports fresh
+ * again, so a later stale episode re-alerts. Module-scoped — the signal is about
+ * the feed, not any caller.
+ */
+const reportedStaleFeeds = new Set<string>();
+
+/** Test-only: the dedup store outlives any single call. */
+export function resetStaleFeedReporting(): void {
```
gbarkhatov, line 215: Flagging per the zero-dead-code sensibility: this is public API surface existing purely for test teardown. It's documented as test-only and harmless, and module-scoped mutable state does need a reset hatch — acceptable given the small surface, no change required.

services/vault/src/config/queryClient.ts
```diff
@@ -22,24 +22,49 @@ const shouldRetry = (failureCount: number, error: Error): boolean => {
...
+/**
+ * Query errors that reflect an expected, non-actionable condition rather than a
+ * fault — e.g. an optional API not configured in this environment. Recorded as
+ * a breadcrumb (so they still attach to any genuine error captured afterwards)
+ * instead of a standalone Sentry issue, so they don't drown the signal. This
+ * onError handler fires for EVERY query failure, so an unfiltered logger.error
+ * here turns any transient blip into a captured issue. Matched by error name so
+ * the match survives bundling across the wallet-connector package boundary.
+ */
+const EXPECTED_QUERY_ERROR_NAMES = new Set<string>([
```
gbarkhatov, line 34: Good call matching on `error.name` via a `Set` rather than `instanceof` — it survives bundling across the wallet-connector package boundary where class identity wouldn't, and the test deliberately constructs a distinct class identity to prove it. Please keep it name-based if this is ever refactored.

gbarkhatov (approved):
Approve — well-scoped, correctly traced to root cause, and the "downgrade/dedup rather than a blunt `ignoreErrors`" call is the right one: it preserves the mainnet signal (stale feed = bad collateral valuation; missing ordinals classifier = can't filter inscription UTXOs) while clearing only the expected-condition flood from devnet/testnet. I traced the core claim through `logger` to confirm the routing actually changes Sentry behavior: `logger.warn` → `addBreadcrumb` (not a standalone issue), `logger.error` → `captureException`, `logger.event` → `captureMessage` — so both the breadcrumb downgrade and the stale-event dedup genuinely reduce captured events. Not on any CRITICAL PATH (no value computation, signing, or secret derivation touched). Test coverage is strong on both branches.

One factual nit in the description (cosmetic, doesn't affect the fix): it says the ordinals query is "retried up to 3×", but the hook sets `retry: false` (`useOrdinalsHook.ts:81`) — the repeated captures are driven by `refetchInterval`, not retries. Worth correcting the description so the mechanism is accurately documented.

Inline notes below are all non-blocking (flags / future-reader context), no changes required.

PR #2134, opened 2026-07-24
fix(vault): show friendly copy for insufficient-ETH-for-gas errors

jrwbabylonlab:
A borrow (and repay/reorder/withdraw) that fails at send time because the
wallet can't cover gas surfaced viem's raw node dump in the UI. The shared
`mapViemErrorToContractError` already tagged this case but passed the raw
message through; map it to the existing "not enough ETH for gas" copy,
guarded so a decoded contract revert keeps its own message.

### DIFF
--- a/services/vault/src/copy.ts
+++ b/services/vault/src/copy.ts
@@ -827,13 +827,15 @@ export const COPY = {
     },
     // Friendly copy for known viem / EIP-1193 / wallet-connector failure
     // categories. Consumed by `sanitizeErrorMessage` in
-    // `src/utils/errors/formatting.ts`. Cross-feature surface (deposit,
-    // refund, activation) so lives under `common` rather than `deposit`.
+    // `src/utils/errors/formatting.ts` and by `mapViemErrorToContractError`
+    // in `src/utils/errors/contract.ts`. Cross-feature surface (deposit,
+    // refund, activation, Aave borrow/repay/withdraw) so lives under
+    // `common` rather than `deposit`.
     classifiedErrors: {
       userRejection:
         "Transaction rejected in your wallet. No changes were made — try again when you're ready.",
       insufficientFunds:
-        "Not enough ETH to cover the deposit fee and gas. Add ETH to your wallet and try again.",
+        "Not enough ETH to cover the network gas fee. Add ETH to your wallet and try again.",
       walletDisconnected:
         "Your wallet was disconnected. Reconnect it and try again.",
       unauthorized:
--- a/services/vault/src/utils/errors/contract.ts
+++ b/services/vault/src/utils/errors/contract.ts
@@ -7,8 +7,11 @@
 
 import { type Abi, type Hash, decodeErrorResult } from "viem";
 
+import { COPY } from "@/copy";
+
 import { COMMON_ERROR_ABI } from "./commonErrorAbi";
 import { CONTRACT_ERROR_MESSAGES } from "./errorMessages";
+import { matchesInsufficientGasFundsMessage } from "./formatting";
 import { ActivationNotPossibleError, ContractError, ErrorCode } from "./types";
 
 /** Minimum length of a revert hex payload: `0x` + a 4-byte (8 hex char) selector. */
@@ -288,11 +291,18 @@ export function mapViemErrorToContractError(
       ) {
         code = ErrorCode.CONTRACT_REVERT;
         reason = reason || message;
-      } else if (
-        message.includes("gas") ||
-        message.includes("insufficient funds for gas") ||
-        message.includes("out of gas")
-      ) {
+      } else if (!decoded && matchesInsufficientGasFundsMessage(message)) {
+        // Wallet can't cover gas * price + value. This fails at send time
+        // (not simulation), so surface friendly copy instead of the raw node
+        // diagnostics dump. Shares the ETH-side matcher with `formatting.ts`
+        // (case-insensitive, one vocabulary). Guarded on `!decoded` so a
+        // genuine contract revert whose wrapper message happens to contain
+        // "insufficient funds" keeps its decoded reason and message rather
+        // than being relabeled as gas.
+        code = ErrorCode.CONTRACT_INSUFFICIENT_GAS;
+        reason = reason || message;
+        enhancedMessage = COPY.common.classifiedErrors.insufficientFunds;
+      } else if (message.includes("gas")) {
         code = ErrorCode.CONTRACT_INSUFFICIENT_GAS;
         reason = reason || message;
       } else if (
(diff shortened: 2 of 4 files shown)


services/vault/src/utils/errors/contract.ts
```diff
@@ -289,10 +291,20 @@ export function mapViemErrorToContractError(
         code = ErrorCode.CONTRACT_REVERT;
         reason = reason || message;
       } else if (
-        message.includes("gas") ||
-        message.includes("insufficient funds for gas") ||
-        message.includes("out of gas")
+        !decoded &&
+        (message.includes("insufficient funds") ||
+          message.includes("exceeds the balance of the account") ||
```
gbarkhatov: These three phrases duplicate `INSUFFICIENT_GAS_FUNDS_PATTERN` in `formatting.ts:45-46`, which maps to the same `classifiedErrors.insufficientFunds` copy. Maintaining the node-message vocabulary in two places invites drift when a new phrasing shows up. Consider exporting a shared matcher (e.g. `matchesInsufficientGasFundsMessage`) or the pattern and calling it here. Bonus: the regex is case-insensitive (`/i`), whereas these `String.includes` checks are case-sensitive, so a node returning `"Insufficient Funds"` capitalized would slip past this branch.

One difference to preserve: `formatting.ts` applies `BTC_FUNDS_GUARD_PATTERN` (sats/pegin) so it doesn't collapse the BTC selector's "need N sats" message. Omitting that guard here is fine because this path only runs for ETH-side Aave operations — but a one-line comment saying so would stop a future copy-paste from silently dropping it.

services/vault/src/utils/errors/contract.ts
```diff
@@ -289,10 +291,20 @@ export function mapViemErrorToContractError(
         code = ErrorCode.CONTRACT_REVERT;
         reason = reason || message;
       } else if (
-        message.includes("gas") ||
-        message.includes("insufficient funds for gas") ||
-        message.includes("out of gas")
+        !decoded &&
+        (message.includes("insufficient funds") ||
```
gbarkhatov: Awareness only, no change requested: this widens the match from `insufficient funds for gas` to any `insufficient funds`. A non-decoded error whose message merely contains "insufficient funds" for a non-gas reason, and which lacks "revert"/"execution reverted" text, would now be relabeled as an ETH-gas shortfall with the friendly copy. Low risk in practice — Aave reverts are either decodable (skipped by `!decoded`) or carry "execution reverted" (caught by branch 1) — and it's the same assumption `formatting.ts` already makes, so the system stays internally consistent. Flagging for the record.

services/vault/src/utils/errors/contract.ts
```diff
@@ -289,10 +291,20 @@ export function mapViemErrorToContractError(
...
       ) {
+        // Wallet can't cover gas * price + value. This fails at send time
+        // (not simulation), so surface friendly copy instead of the raw node
+        // diagnostics dump. Guarded on `!decoded` so a genuine contract revert
+        // whose wrapper message happens to contain "insufficient funds" keeps
+        // its decoded reason and message rather than being relabeled as gas.
+        code = ErrorCode.CONTRACT_INSUFFICIENT_GAS;
+        reason = reason || message;
+        enhancedMessage = COPY.common.classifiedErrors.insufficientFunds;
+      } else if (message.includes("gas") || message.includes("out of gas")) {
```
gbarkhatov: Nit: `"out of gas"` contains `"gas"`, so `message.includes("out of gas")` is fully subsumed by `message.includes("gas")` — the second disjunct is dead. This redundancy predates the PR, but the line was reformatted here, so it's a cheap cleanup to just `} else if (message.includes("gas")) {` per the zero-dead-code policy.

services/vault/src/copy.ts
```diff
@@ -833,7 +833,7 @@ export const COPY = {
       userRejection:
         "Transaction rejected in your wallet. No changes were made — try again when you're ready.",
       insufficientFunds:
-        "Not enough ETH to cover the deposit fee and gas. Add ETH to your wallet and try again.",
+        "Not enough ETH to cover the network gas fee. Add ETH to your wallet and try again.",
```
gbarkhatov, line 838: Stale doc comment: the block comment above this key (copy.ts:828-831) states `classifiedErrors` is "Consumed by `sanitizeErrorMessage` in `src/utils/errors/formatting.ts`." As of this PR, `mapViemErrorToContractError` in `utils/errors/contract.ts` also consumes `insufficientFunds`. Per the repo rule to update comments that reference old behavior, consider generalizing that line to mention the contract-error path too.

gbarkhatov (dismissed):
Approving — this correctly routes send-time "can't cover gas * price + value" failures to the friendly copy, and the `!decoded` guard (with its dedicated test) properly protects genuine contract reverts whose wrapper text happens to contain "insufficient funds". No blockers. A few minor, non-blocking notes inline: reuse the shared detection pattern from `formatting.ts` to avoid drift, a redundant `out of gas` disjunct, and a now-incomplete doc comment in `copy.ts`.

jonybur (dismissed):
Correctly routes send-time "insufficient funds for gas * price + value" failures (and the equivalent "exceeds the balance of the account" / "exceeds transaction sender account balance" wordings) to the friendly `insufficientFunds` copy instead of the raw node dump. The `!decoded` guard is the important bit — a genuine contract revert whose wrapper text happens to contain "insufficient funds" keeps its decoded reason, and the new negative test pins that exactly. The copy tweak from "deposit fee and gas" to "network gas fee" is more accurate for the ETH-only code path this maps to.

services/vault/src/utils/errors/contract.ts
jrwbabylonlab: Fixed — reusing `matchesInsufficientGasFundsMessage` from formatting.ts now, so it's one vocabulary and case-insensitive. The BTC guard comes along for free (harmless on the ETH path). Added a case-insensitive regression test.

services/vault/src/utils/errors/contract.ts
jrwbabylonlab: Fixed — collapsed to `message.includes("gas")`.

services/vault/src/copy.ts
jrwbabylonlab: Fixed — the comment now names the contract.ts consumer too.

services/vault/src/utils/errors/contract.ts
jrwbabylonlab: Ack — same assumption formatting.ts already makes; kept consistent by reusing its matcher.

services/vault/src/copy.ts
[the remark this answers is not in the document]
jrwbabylonlab: These calls attach zero ETH value, so gas is the only balance component here. Keeping the copy as-is.

jeremy-babylonlabs (approved):

PR #2136, opened 2026-07-24
feat(vault): promote funnel facets to Sentry tags

jonybur:
## What

Promote the depositor-funnel join keys and breakdown facets from Sentry event `extra` into queryable **tags**, and extend `logger.event` to forward a `tags` object.

## Why

Funnel milestones (`deposit.*`, `activation.*`, `exit.redeem.*`) emit through `logger.event`, which today only puts its payload into `captureMessage`'s `extra`. **Event `extra` is not filterable or facetable in Sentry** - so `vaultId`, `daemonStatus`, `timeoutReason`, and `providerId` were invisible to alerts and dashboards. The only queryable tags were `funnelStage` (on the `logger.error` failure captures) and `message`.

This blocks the alerting plan:

- **Per-vault stall alerts** ("broadcast.succeeded but no activation.activated within N hours") need `vaultId` as a tag to join the two events per entity.
- **Rejection breakdown** needs `daemonStatus` to facet `activation.daemon.terminal`.
- **Pegout breakdown** needs `timeoutReason` to facet `exit.redeem.pegout_timeout`.
- **Per-VP segmentation** needs `providerId`.

## Changes

- `logger.event` accepts a `tags?: Record<string, string>` and forwards it to `captureMessage` - passed raw and scrubbed at send by `scrubSentryEvent`, exactly as `logger.error` already handles tags.
- Each emit site moves its join/facet keys from `extra` into `tags`:
  | event | tags added |
  | --- | --- |
  | `deposit.started` | `providerId` |
  | `deposit.registered` | `vaultId`, `providerId` |
  | `deposit.broadcast.succeeded` (inline + resume) | `vaultId` |
  | `activation.activated` | `vaultId` |
  | `activation.daemon.terminal` | `vaultId`, `daemonStatus` |
  | `exit.redeem.*` | `vaultId`, `timeoutReason` (when set) |
  | `captureFunnelFailure` | `vaultId` (alongside `funnelStage`) |
- The deposit-flow outer catch now carries a `depositStep` tag (the `DepositFlowStep` enum member name), read from a `currentStepRef` because the flow callback closes over a stale `currentStep`.

## Safety

No new values are transmitted. `vaultId` is already `shortId`-shortened (`first4...last4`, which survives the address/hex scrubbers), and tags pass through the same send-time `scrubSentryEvent` pass as `extra`. Descriptive-only fields (`batchId`, `ethTxHash`, `prePeginTxid`, `vaultCount`, `amountBucket`) stay in `extra`.

## Testing

- `pnpm --filter vault run test` - 2627 passed, 6 skipped, 0 failed
- `tsc -p tsconfig.lib.json --noEmit` - clean
- `eslint` on all edited files - 0 errors
- Added a `logger.event` test asserting tags forward to `captureMessage` and never leak into `extra`; updated the daemon/pegout/failure emit-shape assertions to the tag schema.


services/vault/src/infrastructure/telemetryEvents.ts
```diff
@@ -75,8 +76,8 @@ export function captureFunnelFailure(
 ): void {
   if (classifyError(err) === "user-rejection") return;
   logger.error(err instanceof Error ? err : new Error(String(err)), {
-    tags: { funnelStage: stage },
-    data: { vaultId: shortId(vaultId), ...extra },
+    tags: { funnelStage: stage, vaultId: shortId(vaultId) },
```
gbarkhatov: `vaultId` is per-entity unique, so promoting it to a Sentry **tag** (here, and at the milestone emit sites) creates unbounded indexed-tag cardinality — the pattern Sentry flags for cost and tag-value indexing limits (values can be truncated/dropped on high-volume projects). It's deliberate and required for the per-vault join/alert plan, and everything else is bucketed, so the tradeoff is justified — just worth confirming the Sentry plan/quotas tolerate a unique-per-vault indexed tag before the alerting joins come to depend on it.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1223,6 +1232,7 @@ export function useDepositFlow(
         if (!signal.aborted) {
           setError(mapDepositError(err));
           logger.error(err instanceof Error ? err : new Error(String(err)), {
+            tags: { depositStep: DepositFlowStep[currentStepRef.current] },
```
gbarkhatov, line 1239: The `currentStepRef` mirror + this `depositStep` tag is the one behaviorally subtle change in the PR — it exists specifically because the flow callback closes over a `currentStep` that's stale by the time it throws. Worth a test that advances the flow past `DERIVE_VAULT_SECRET`, forces the outer catch, and asserts the tag reflects the *current* step rather than the initial one; otherwise a regression back to reading `currentStep` directly would pass silently. (Minor: `DepositFlowStep[...]` relies on the enum staying numeric — a future switch to a string enum would make this `undefined`.)

services/vault/src/infrastructure/__tests__/logger.test.ts
```diff
@@ -123,5 +123,24 @@ describe("logger", () => {
...
+      });
+
+      expect(captureMessage).toHaveBeenCalledWith(
+        "deposit.registered",
+        expect.objectContaining({
+          tags: { vaultId: "0x11...2222", providerId: "0xab...cdef" },
+          extra: expect.objectContaining({ batchId: "batch-1" }),
+        }),
+      );
+      expect(vi.mocked(captureMessage).mock.calls[0][1]).not.toHaveProperty(
```
gbarkhatov: Minor: `not.toHaveProperty("extra.tags")` treats the dot as a nested path (`extra` → `tags`), so it would also pass vacuously if `extra` were absent entirely. The `objectContaining` assertion above is doing the real work (it pins `extra` present with `batchId` and `tags` at top level), so coverage is fine — just flagging that this line is weaker than it looks.

gbarkhatov (dismissed):
Approving. Reviewed the full diff plus the send-time scrubbing path — `scrubSentryEvent` runs `redactData` over `event.tags` and is wired into `beforeSend`, so tags get the same safety net as `extra`, and the `shortId` `first4...last4` values survive the address/hex regexes. The `DepositFlowStep` enum is numeric, so `DepositFlowStep[currentStepRef.current]` reverse-maps to the member name correctly, and every migrated join/facet key now lives under `tags` with descriptive fields (`batchId`, `ethTxHash`, `prePeginTxid`, `vaultCount`, `amountBucket`) left in `extra`. Clean, well-scoped change that touches none of the value/signing critical paths.

Left three inline notes, all non-blocking: a design confirmation on `vaultId` tag cardinality, a test-coverage gap on the `currentStepRef`/`depositStep` staleness fix, and a minor weak-assertion note.

One consistency observation outside this diff: other `logger.error` sites (e.g. `useRefundState.ts`) still keep `vaultId` in `data` rather than `tags`. Those aren't funnel milestones, so they're correctly out of scope here — just flagging that `vaultId`-as-a-tag is now non-uniform across the codebase, in case a future alert expects it everywhere.

(I also considered a possible double-capture between `captureFunnelFailure` and the new outer-catch `depositStep` tag, but the `captureFunnelFailure` sites all live in sibling hooks outside the `useDepositFlow` try, so the same error doesn't reach both paths — no issue.)

gbarkhatov:
@jonybur verify error

services/vault/src/infrastructure/telemetryEvents.ts
jonybur: Checked the account before merging this. Short answer: safe, with room to spare.

**Billing** — Sentry bills per event, never per tag or per distinct tag value, so an unbounded `vaultId` tag adds exactly zero cost. Confirmed against docs.sentry.io/pricing.

**Headroom** — `tbv` took 220 accepted errors in the last 30 days against a 100,000/mo reserved quota on our Business plan, plus a $50 PAYG buffer and Spike Protection on. That's under 3% of quota. Current tag cardinality on the project tops out at 47 distinct values (`user`); everything else is ≤26.

**Sentry's own guidance runs the other way from the usual instinct** — [docs.sentry.io/guides/issues-errors](https://docs.sentry.io/guides/issues-errors/) says verbatim: *"Tags are searchable, so use them for high-cardinality data like IDs, regions, and feature flags"*, with a sample tagging `order_id: order.id`. There's no documented ingest cap on distinct values per key; the 1000 people remember is read-path pagination. High cardinality degrades tag-search UX (top-9 display, approximate `uniq()` counts) — it never affects ingest or quota.

**Precedent in our own org** — `simple-staking-phase-2-mainnet` already runs a per-entity `user` tag at 3,306 distinct values with no ill effect.

**Value length** — `shortId()` gives us an 11-char `first4...last4`, well under Sentry's 200-char tag-value limit, so no truncation.

Threshold to watch is total event volume approaching 100k/mo, not the tag's cardinality. At current volume we're years from either.

jonybur:
@greptile-apps

gbarkhatov (approved):

PR #2138, opened 2026-07-24
fix(vault): report a repay shortfall as a balance problem

jrwbabylonlab:
Three unrelated fixes surfaced while auditing the borrow/repay path.

**Repay error message.** `repayPartial`'s balance backstop threw a string
containing "not enough", which `mapViemErrorToContractError` rewrote to
"Repay failed: Insufficient liquidity. Please try a smaller amount." The
message now lives in `copy.ts`, names the actual numbers, and is worded
around the rewrite rules. The form already blocks this case, so it only
fires when the balance drops between render and submit.

**SDK `./tbv/core/vault-secrets` subpath.** `package.json` exported it but
`vite.config.ts` had no matching entry — only the `.d.ts` was emitted, so
it typechecked and then failed at runtime. It was the only one of the 13
export keys without an entry. Added the entry rather than dropping the
export, since the export is already published. No derivation logic changed.

**`NEXT_PUBLIC_FF_ENABLE_SIGNING_NOTIFICATIONS`.** Never passed by the
release workflow, so the per-environment toggle its docs describe couldn't
be used. (The env var still needs setting for it to take effect.)


jonybur (approved):
Three well-scoped fixes.

- The repay balance-below-amount error now flows through `COPY.loans.repay.balanceBelowRepayAmount` with the actual required/available amounts, and the added contract-mapping test pins that the wording survives the `Repay failed:` rewrite. The form still blocks this path; this only fires on a balance change between render and submit, matching the comment.
- The `./tbv/core/vault-secrets` subpath was exported from `package.json` but had no `vite.config.ts` entry, so only the `.d.ts` was emitted and any runtime import would fail. Adding the entry restores the published export without touching any derivation logic (no changes under `vault-secrets/` or its re-exports from the WASM package).
- `NEXT_PUBLIC_FF_ENABLE_SIGNING_NOTIFICATIONS` is now plumbed through the release workflow so the per-environment toggle is actually settable.

services/vault/src/copy.ts
```diff
@@ -1164,6 +1164,10 @@ export const COPY = {
       // it can exceed the displayed debt. Wording avoids the rewrite substrings.
       balanceBelowFullRepay: (required: string, balance: string) =>
         `Repaying in full needs ${required} (including accrued interest and fees), but your balance is ${balance}. Acquire a little more and try again.`,
+      // Submit-time backstop for a partial repay: the form blocks this, so it
+      // only fires when the balance drops between render and submit.
```
gbarkhatov, line 1168: This claim is narrower than what the code does. The form's guard is a float comparison (`repayAmount > maxRepayAmount` in `validateRepayAction.ts:117`), while the hook converts with `parseUnits(repayAmount.toFixed(Math.min(onChainDecimals, SAFE_TOFIXED_PRECISION)), onChainDecimals)` — and `SAFE_TOFIXED_PRECISION` is 15. For a debt token with more than 15 decimals that round-trip can round **up** past the raw balance, tripping this same throw with a perfectly stable balance.

It's reachable without anyone hand-typing the balance: `pickRepayParams` returns `mode: "partial"` with `amount: freshBalanceAmount` whenever `freshBalanceRaw < freshDebtRaw`, so clicking **Max** on an 18-decimal token with balance below debt sends the balance-as-float down the partial path. `Number(formatUnits(raw, 18)).toFixed(15)` then re-parses to a value that can sit up to ~1e3 wei above `raw`, and `repayPartial`'s check fires. The hook's own `repayAmountRaw` JSDoc describes exactly this hazard ("≥16-significant-digit raw values (any 18-decimal token with > ~10 tokens in the wallet)") for max mode, and the inline comment on the partial branch justifies itself with "the user typed it" — which is false on that path.

The message is still the right one for that case, so nothing to change in the copy. Either widen the comment (drop "only", add the round-up case), or — better, as a follow-up — let the partial path accept an exact bigint the way max mode already does, which removes the failure rather than describing it.

services/vault/src/utils/errors/__tests__/contract.test.ts
```diff
@@ -559,5 +559,17 @@ describe("Contract Error Mapping", () => {
...
       expect(mapped.message).toBe(`Repay failed: ${body}`);
     });
+
+    it("preserves the balance-below-full-repay copy under the Repay mapping", () => {
+      const body = COPY.loans.repay.balanceBelowFullRepay("3 USDC", "2 USDC");
+      const mapped = mapViemErrorToContractError(new Error(body), "Repay");
+      expect(mapped.message).toBe(`Repay failed: ${body}`);
+    });
+
+    it("preserves the balance-below-repay-amount copy under the Repay mapping", () => {
```
gbarkhatov, line 569: These two pin the strings that exist today, which is the right scope for this PR, but the pattern is accumulating: four `COPY.loans.repay.*` entries now carry a hand-maintained "wording avoids the rewrite substrings" constraint, and nothing stops a fifth from landing untested. Any new rule added to `getEnhancedErrorMessage` silently widens the trap for all of them at once.

The class-level fix is a marker error type (or a flag on `ContractError`) that `mapViemErrorToContractError` passes through unmodified, so messages we formatted ourselves are never substring-matched against chain-error heuristics — the heuristics only exist for opaque node/contract strings. Worth a ticket, not this PR.

Cheaper interim step if the ticket sits: make this block table-driven over every function-valued entry under `COPY.loans.repay`, so new copy is covered by construction instead of by remembering.

gbarkhatov (approved):
Approved — three independent, low-risk fixes. I verified each diagnosis against the branch and ran the two touched test files (74 tests pass).

**Verified:**

- The old `"not enough"` string really did hit the rewrite at `utils/errors/contract.ts:216-219`, and the new copy clears every substring rule in `getEnhancedErrorMessage` (`paused` / `frozen` / `not enough` / `insufficient liquidity` / the `cap` family / `denied`). It also no longer contains `insufficient`, so `formatErrorMessage` wouldn't flatten it either — free hardening, even though that isn't the live path.
- The SDK diagnosis is exact: `dist/tbv/core/vault-secrets/` currently holds only `.d.ts` / `.d.ts.map` — no `index.js`, no `index.cjs` — while `package.json` publishes all three conditions. The new entry key maps to exactly the published paths, and the WASM package is already `external`, so the new chunk can't duplicate WASM state.
- `NEXT_PUBLIC_FF_ENABLE_SIGNING_NOTIFICATIONS` was the only production flag missing from the workflow — `NEXT_PUBLIC_FF_GOD_MODE_PANEL` is hard-gated on `import.meta.env.DEV`, so its absence is deliberate. Unset repo var → empty string → `false`, so no environment changes behavior until someone sets it.

**Non-blocking, beyond the two inline notes:**

- Please confirm with a real `pnpm --filter @babylonlabs-io/babylon-ts-sdk build` that `dist/tbv/core/vault-secrets/index.js` and `.cjs` are actually emitted — I verified the pre-fix state, not the post-fix output. Worth a parity test asserting `package.json` `exports` keys ↔ `vite.config.ts` lib entries ↔ `knip.json` entries; three hand-synced lists is how this drifted in the first place.
- `positionTransactions.test.ts`: `beforeEach` calls `vi.clearAllMocks()`, which does not clear `mockResolvedValue`, and `mockGetERC20Balance` is the one mock with no default pinned there — so the new `999999n` leaks into every later test until the next explicit pin at line 556. Nothing breaks today, but the file's own comment at line 125 warns about exactly this pattern.
- The `describe` block the two new cases land in is still named "repay approval copy survives message rewriting", but now holds balance copy too — worth dropping "approval".
- Unrelated and pre-existing: `useRepayTransaction.ts:243` still inlines the "Asset integrity check failed…" user-facing string instead of sourcing it from `copy.ts`. Worth sweeping the next time that file is touched.

Issue #2109, opened 2026-07-22
Ledger: dedicated vault provider over the DMK (connection + intent ceremony)


jrwbabylonlab:
New provider — `LedgerProviderV2` stays untouched. Built directly on `@ledgerhq/device-management-kit` + web-hid transport with **our own raw APDUs**.

**Scope change (2026-08-12):** Ledger confirmed the published `@ledgerhq/device-signer-kit-bitcoin` will not drive the vault app and will not be extended — "the Vault needs its own signer kit built on DMK". This issue therefore covers the **connection + intent ceremony** only; signing is split into its own milestones: #2219 (SIGN_PSBT host protocol), #2220 (ts-sdk seam threading), #2221 (PoP), #2222 (Pre-PegIn policy path).

Delivered by PR #2214 (`feat/vault-intent-seam`):

- DMK session lifecycle (WebHID discovery → connect, session refresher disabled for the whole connection until Ledger confirms which commands nullify a loaded intent), `GET_APP_AND_VERSION` connect preflight, idempotent `connectWallet` (tab-visibility rechecks must not kill a live session).
- Key read as a raw `GET_EXTENDED_PUBKEY` APDU at the pinned BIP-86 leaf; taproot address derived locally from the pubkey (no device address read — address integrity is enforced at signing time by the firmware, which rebuilds every script from its own seed).
- `DERIVE_CONTEXT_HASH` (0x81) and the three-phase `APPROVE_VAULT_INTENT` (0x80) with the TLV encoder golden-vectored against Ledger's own Python client.
- Device-envelope gate (`envelope.ts` / `deviceCaps.ts`) rejecting out-of-range terms BEFORE any device I/O, with the seam's typed `DepositTermsRejectedError`.
- Host-side mirror of the firmware's vault state machine (IDLE → HASH_DERIVED → INTENT_LOADED; HASH_DERIVED is single-use) so state errors surface actionably instead of as opaque `SW_BAD_STATE`.
- Status-word classification mirrored from DMK's `CommandUtils` (user-refusal 0x5501/0x6985, locked set, base-app + vault codes) and an APDU trace hook (`globalThis.__LEDGER_VAULT_APDU_TRACE__`) for device-day debugging.
- Ships behind `NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET` (default off; registered in the release workflow). `signPsbt`/`signPsbts`/`signMessage` throw a typed capability error until #2219.

[...]

Issue #2110, opened 2026-07-22
Ledger: deposit-flow UX, device error mapping, enable + emulator E2E


jrwbabylonlab:
Scope:
- New `DepositFlowStep`s + `copy.ts` strings for the intent-wallet deposit flow.
- Two-approval flow: under v22 this is the DESIGNED model, not a fallback — one intent approval per device connection; the second approval is the normal cross-connection case. Same-Ledger ETH signing still forces the reconnection (the app switch wipes the vault session). Design unblocked.
- Nullification-resume UX: any signing error, disconnect, or cap breach nullifies the intent + root — offer reconnect → silent re-derive → re-approve → re-sign; never treat it as fatal.
- Resume-path DepositTerms rebuild: `runDepositorPresignFlow` throws for approval-capable wallets when `depositTerms` is absent (fresh flows thread `PreparePeginResult.depositTerms`; the SDK error says the resume rebuild is not wired). Rebuild the terms from on-chain/indexer data on resume and pass them through `signAndSubmitPayouts`.
- Device error mapping — wait for the real status-word taxonomy from Ledger (our previously assumed 0x6985/0x6A80 were inferred from an outdated guide, never in the spec).
- Claim destination fixed to the device-derived address for Ledger depositors (no arbitrary payout-address input; PayoutFinalize destination is spec-restricted to the device's BIP-86 key).
- E2E under emulator — done: Speculos stages in #2316, ELF `b0c0ac4d` in #2496. (The wallet is only ever added to the disabled list via the flag branch, so there is nothing to remove from `ALWAYS_DISABLED_WALLETS`.)

**Dependencies:** #2109 (connection — done), #2219 (signing), and #2220 (the resume-path DepositTerms rebuild this issue's resume UX needs).

---
**Update 2026-08-04:** deposit-terms envelope rejections now surface as the typed `DepositTermsRejectedError` (exported guard `isDepositTermsRejectedError` in ts-sdk `deposit-terms/depositTermsErrors.ts`). Add here: map it in `depositErrors.ts` to dedicated copy (+ tests). Until then it falls through to the generic raw-message bucket. The #2109 emulator E2E asserts the typed rejection; the mapped-copy assertion belongs to this issue.

---
**Update 2026-08-12:** Ledger will not extend their Bitcoin signer kit — we own the SIGN_PSBT host protocol (#2219). Consequences for this issue:
[...]

PR #2139, opened 2026-07-24
feat(ts-sdk): add DepositTerms seam for intent-based signing wallets

### LINKED ISSUE #2102: Ledger: DepositTerms seam — protocol type, thin builder, capability-gated call sites
Protocol-shaped `DepositTerms` type + thin `buildDepositTerms` builder + a `supportsDepositApproval()` capability (interface `DepositTermsApprover.approveDepositTerms`) with no-op gated call sites. The ts-sdk stays protocol-only — **no Ledger wire concepts** (no `coinType`/SLIP-44, no TLV framing, no device-sort, no `version` literal); those live in the Ledger provider (#2109). Extension wallets are unaffected.

Design C, agreed via review (3 agents + codex). Full record: `todo/ledger/2102-vault-intent-seam-design.md`.

Scope:
- `tbv/core/deposit-terms/`: `DepositTerms` (per-vault groups, timelocks, prepeginTxid, fees, keys — all protocol values the deposit flow already holds), `DepositTermsApprover`, `supportsDepositApproval()`, `buildDepositTerms()`.
- `buildDepositTerms` is a thin projection + commission cap: it validates `commissionBps` (required; the vault app is the SDK's only consumer and always quotes it) is bounded `[0, MAX_VP_COMMISSION_BPS_EXCLUSIVE)` (reusing the shared constant), computes the commission floor, and checks txid shape / non-empty amounts / positive timelocks. It does **not** re-sort keys, re-check role disjointness, or re-check non-negativity — all already enforced upstream (`validateOnChainParticipantKeys`, `signDepositorGraph`, WASM sizing asserts).
- Ordering contract (what survived #2103): the device's intent-load ordering — keeper/challenger key lists ascending lexicographic, vault groups ascending `htlc_vout` — is met by documented contract, not builder asserts. Keys arrive sorted from `validateOnChainParticipantKeys` (the dApp passes the `...Sorted` outputs) and `htlcVout` is index-derived; the contract is documented on the `DepositTerms` type.
- The three vault-app `{...wallet}` wrapper sites (deposit flow ×2, payout signing) forward `approveDepositTerms` explicitly — object spread drops prototype methods, and the capability gate must not fail open. Regression-tested with a prototype-method wallet at both wrappers.
- Approval runs before PegIn signing in `preparePeginCommit` (derive → approve → sign); re-approval at the top of `runDepositorPresignFlow`.
- Threaded dApp-side as an opaque pass-through: `PreparePeginResult.depositTerms` → useDepositFlow → signAndSubmitPayouts → runDepositorPresignFlow.

Fixes a real bug found during review: the first cut duplicated the commission formula with its own constant and **no upper-bound check** (which `payout.ts` enforces) — closed by reusing `MAX_VP_COMMISSION_BPS_EXCLUSIVE`.

Acceptance:
- `buildDepositTerms` unit-tested (field projection + commission bound, incl. an upper-bound reject).
- Call sites are no-ops for all current wallets; ts-sdk + vault build/lint/test green.
- Approved terms pinned field-for-field in the seam test, with the two WASM-sourced sat amounts recomputed independently so same-typed swaps can't pass.

jrwbabylonlab:
Closes #2102

## What

Foundation for Ledger's intent-based signing (device spec HLD v22): a typed
`VaultIntent` (the deposit terms the device displays for approval), a pure
`buildVaultIntent()` builder, and a `supportsVaultIntent()` capability with two
gated call sites — after Pre-PegIn construction in `PeginManager.preparePeginCommit`
(derive → approve → sign PegIns), and at the top of `runDepositorPresignFlow`
(re-approval on the payout connection). The intent threads from
`PreparePeginResult` through the vault service into `signAndSubmitPayouts`.

## Why now

The Ledger provider (#2109) arrives with their DMK package (~mid-Sept). Landing
the seam first freezes the flow shape so that integration is an adapter drop-in,
and the typed intent doubles as our field-by-field checklist against the device spec.

## Zero behavior change for current wallets

No existing wallet implements `approveVaultIntent`, so both call sites are no-ops —
proven by dedicated tests (flow completes with no extra wallet interactions). New
required params/fields (`commissionBps` in, `vaultIntent` out) are compile-time
threading only; no value computation changed. `PeginManager` changes are purely
additive between txid computation and PegIn signing.

The builder validates participant-set invariants the device also enforces
(sorted key lists, role disjointness, non-negative amounts) — these throw for
all wallets, but only on data that would indicate real protocol-invariant
corruption upstream.


packages/babylon-ts-sdk/src/tbv/core/intent/buildVaultIntent.ts
```diff
@@ -0,0 +1,96 @@
...
+    throw new Error("buildVaultIntent: depositor key must be disjoint from all participant keys");
+  }
+
+  const commissionBps = BigInt(inputs.commissionBps);
+  const vaults: VaultIntentVaultGroup[] = inputs.vaultAmounts.map((vaultAmount, index) => ({
+    htlcVout: index,
+    vaultProviderPk,
+    vaultAmount,
+    // floor(V * bps / 10_000) — matches btc-vault transactions/mod.rs commission math.
+    commissionFee: (vaultAmount * commissionBps) / BPS_DENOMINATOR,
```
gbarkhatov: **The commission the device approves is not the one the contract binds.**

`commissionFee` is derived from the *quoted* bps, and `buildVaultIntent` runs inside `preparePeginCommit` — before ETH registration. Registration then binds `maxAcceptableCommissionBps = min(quoted + COMMISSION_BPS_HEADROOM, 9999)` (`PeginManager.ts:100-115`, `:1292`), and that headroom exists precisely so the VP may raise its rate between quote and submit.

So the commission actually charged — the one baked into the VP-built payout outputs the device will later verify against the loaded intent — can legally exceed what the depositor approved on-device by up to 25 bps, or be lower if the VP dropped its rate. If tags `0x0109-0x010E` compare `commission_fee` for exact equality, an in-headroom drift refuses signing *after* BTC is committed.

Two things worth settling before #2109 lands:

1. Pin against the device spec whether `commission_fee` is exact or a ceiling. If it's a ceiling, build it from `maxAcceptableCommissionBps` and say so in the field comment.
2. Either way, critical path #3 says re-derive from an independent source and assert rather than accepting a value handed to us. `getVaultProviderCommission()` already exists (`clients/eth/vault-registry-reader.ts:234`); reading it here would also let `commissionBps` drop off the public `PreparePeginParams` entirely.

packages/babylon-ts-sdk/src/tbv/core/intent/vaultIntent.ts
```diff
@@ -0,0 +1,85 @@
...
+ * - MUST retain the intent for the connection lifetime; MUST autonomously silent
+ *   re-derive (P2=0x01) + re-approve after any nullification (disconnect, signing
+ *   error, cap breach) — including before the Pre-PegIn signPsbt that happens AFTER
+ *   ETH registration (same-device app switch wipes the session, v22 L2311-2313).
+ * - MUST be idempotent for a byte-equal intent on a live session (no device
+ *   round-trip if already loaded) — the per-vault payout loop re-calls it. [R]
+ * - A second call with a DIFFERENT intent replaces the loaded one (spec-normal).
+ * - MUST reject (throw) on user refusal — dApp treats as cancel, like signPsbt.
+ * - MUST be defined as a class FIELD, not prototype method — deposit-flow wallet
+ *   wrappers spread the object ({...wallet}) and would strip prototype methods. [R]
```
gbarkhatov: **Capability detection fails open across the app's three `{...wallet}` spreads.**

This line is currently the only thing standing between the seam and a silent no-op. The vault app hands the SDK three spread copies of the wallet, and each forwards a hand-picked allowlist of methods while dropping everything else on the prototype:

- `phaseTrackingBtcWallet` — `useDepositFlow.ts:471-481` (this is the wallet `PeginManager` receives)
- `postBroadcastBtcWallet` — `useDepositFlow.ts:872-873`
- `graphProgressWallet` — `usePayoutSigningState.ts:232`

If the #2109 provider exposes `approveVaultIntent` as a prototype method, all three drop it, `supportsVaultIntent` returns `false`, and both gates skip — signing proceeds with no device approval and nothing throws. That is the wrong failure direction for a security control, and it rests on an external package honouring a comment.

Forwarding it explicitly makes the contract enforceable instead of aspirational, exactly the way `signPsbts` already is at `useDepositFlow.ts:478-480`:

```ts
...(typeof confirmedBtcWallet.approveVaultIntent === "function"
  ? { approveVaultIntent: (i) => confirmedBtcWallet.approveVaultIntent!(i) }
  : {}),
```

packages/babylon-ts-sdk/src/tbv/core/services/deposit/runDepositorPresignFlow.ts
```diff
@@ -332,6 +339,19 @@ export async function runDepositorPresignFlow(
 
   signal?.throwIfAborted();
 
+  // Intent-signing wallets must approve the intent before any signing call
+  // it authorizes, including the Phase 3/4 payout and depositor-graph signing below.
+  if (supportsVaultIntent(btcWallet)) {
+    if (!vaultIntent) {
```
gbarkhatov: **Resume is an unrecoverable dead end for exactly the wallets this seam exists for.**

`usePayoutSigningState.ts:281` (the PayoutSignModal resume path) calls `signAndSubmitPayouts` without `vaultIntent`, so every intent wallet resuming a deposit lands here.

This message never reaches the user, which makes it worse rather than better: it's a plain `Error`, so `formatPayoutSignatureError` matches none of its substrings and falls through to the generic `"An unexpected error occurred while signing payouts. Please try again or contact support."` (`utils/errors/formatting.ts:387`). The user gets a Retry button that can never succeed and no indication why. #2110 is named in the PR, but the user-facing shape of the gap isn't handled.

Minimum before a Ledger provider ships: a dedicated branch in `formatPayoutSignatureError` with copy in `copy.ts`, or gate the modal's Sign action off for `supportsVaultIntent` wallets so they never enter a path that can only fail.

packages/babylon-ts-sdk/src/tbv/core/intent/buildVaultIntent.ts
```diff
@@ -0,0 +1,96 @@
...
+function normalizeXOnly(label: string, key: string): string {
+  const stripped = (key.startsWith("0x") ? key.slice(2) : key).toLowerCase();
+  if (!/^[0-9a-f]+$/.test(stripped) || stripped.length !== XONLY_HEX_LENGTH) {
+    throw new Error(`buildVaultIntent: ${label} is not a 32-byte x-only hex key: ${key}`);
+  }
+  return stripped;
+}
+
+function sortedUnique(label: string, keys: readonly string[]): string[] {
+  if (keys.length === 0) throw new Error(`buildVaultIntent: ${label} list is empty`);
```
gbarkhatov: **An empty universal-challenger set now aborts every deposit, for every wallet.**

Nothing upstream requires >= 1 universal challenger. `validateOnChainParticipantKeys` (`services/deposit/validateOnChainParticipantKeys.ts:85-97`) only checks that the indexer set equals the on-chain `ProtocolParams` set — it passes two empty lists through happily — and no dApp-side check adds a lower bound. So a zero-UC on-chain config works today and would start throwing here for all wallets, not just intent wallets.

The keeper case is fine: `deriveLocalChallengers` already throws on an empty VK set, so this matches existing behaviour. For challengers, either drop the emptiness check, or — if zero UCs really is invalid — put it in `validateOnChainParticipantKeys` where it fails early with an actionable message, rather than deep inside a builder most wallets never consume.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -899,6 +921,34 @@ export class PeginManager {
       calculateBtcTxHash(fundedPrePeginTxHex),
     );
 
+    // #2109: always build the intent so callers get it back regardless of
+    // wallet capability; only intent-capable wallets need the approval below.
+    const peginMaxFee = await computeMinPeginFee(
```
gbarkhatov: **Second, unasserted `computeMinPeginFee` call where an asserted one already exists.**

`assertWasmPeginSizing` already calls `computeMinPeginFee` with these same four arguments and asserts that the reserve inside each `htlcValue` equals it *exactly* (`primitives/psbt/assertWasmPeginSizing.ts:117`, reached via the `buildPrePeginPsbt` call a few lines above). This recomputes it from independently reassembled arguments.

Same answer today, but it's the duplicate-derivation shape critical paths #1/#2 warn about: the value the device will enforce against is no longer the value that was cross-checked against the funded HTLC, so a future change to either argument list desynchronizes them silently.

Prefer threading the asserted number out — add `minPeginFee` to `PrePeginPsbtResult` and read `prePeginResult.minPeginFee`. Failing that, assert `prePeginResult.htlcValues[i] - params.amounts[i] - depositorClaimValue - p2aAnchorValue === peginMaxFee` right here.

packages/babylon-ts-sdk/src/tbv/core/intent/buildVaultIntent.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  // Device rejects cross-role key overlap (v22 L1362-1366).
+  const keeperSet = new Set(keeperPks);
+  for (const pk of challengerPks) {
+    if (keeperSet.has(pk)) throw new Error(`buildVaultIntent: key ${pk} appears in both keeper and challenger lists`);
+  }
+  if (keeperSet.has(vaultProviderPk) || challengerPks.includes(vaultProviderPk)) {
+    throw new Error("buildVaultIntent: vault provider key must not appear in keeper/challenger lists");
+  }
+  if (depositorPk === vaultProviderPk || keeperSet.has(depositorPk) || challengerPks.includes(depositorPk)) {
```
gbarkhatov: **Throws where the neighbouring SDK path deliberately tolerates.**

`deriveLocalChallengers` (`services/deposit/signDepositorGraph.ts:105-123`) hits the same condition and *filters* the depositor out of the VK set, with a comment framing it as defense-in-depth over the Rust-side `TxGraphParams::validate`. This throws instead, earlier in the flow, for every wallet.

Both stances are defensible; having both is not. I'd make `deriveLocalChallengers` throw too and keep this one — silently filtering changes the challenger set the depositor signs for, which is the more dangerous of the two behaviours. Either way, a line here pointing at whichever becomes the single policy would help.

packages/babylon-ts-sdk/src/tbv/core/intent/vaultIntent.ts
```diff
@@ -0,0 +1,85 @@
...
+  // rate fed to WASM (PeginManager.ts:878), NOT mempoolFeeRate
+  peginCsvTimelock: number; // u32 TLV (tag 0x0101); P. Source == uint16(timelockAssert)
+  payoutTimelock: number; // u32 (tag 0x0102); t2 == timelockAssert. Derived from the same
+  // param as P — provably equal (protocol-params-reader.ts:104-111
+  // throws >65535), builder sets both from one input
+  htlcRefundTimelock: number; // u32 (tag 0x0103); T_refund, 72..4320 in v22
+  prepeginTxid: string; // 64-char DISPLAY-order hex; TLV wire is little-endian u256
+  // (v22 L1290) — the #2109 adapter MUST byte-reverse on encode
+  prepeginMaxFee: bigint; // u64 (tag 0x010F) = funded sizing.fee (exact)
+  keeperPks: string[]; // tag 0x0107; ascending byte-lex, builder-sorted + asserted
```
gbarkhatov: **Two things the #2109 adapter author will otherwise have to rediscover.**

- `builder-sorted + asserted` (this line) overstates it. `sortedUnique` sorts and rejects duplicates; it never asserts the *input* arrived sorted. Callers pass `vaultKeeperBtcPubkeysSorted` / `universalChallengerBtcPubkeysSorted`, so a caller that silently regressed to unsorted input gets quietly re-sorted rather than flagged. Either drop `+ asserted`, or assert input order and mean it.
- `challengerPks` (next line) carries universal challengers only, but the graph's challenger set is `local ∪ universal` where local = the VKs (`signDepositorGraph.ts:93`). Presumably the device recomputes local from `keeperPks` (0x0107) — nothing here says so, and it's the most surprising omission in the struct. One sentence saves the adapter author a round trip.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -202,6 +208,11 @@ export interface PreparePeginParams {
    */
   vaultProviderBtcPubkey: string;
 
+  /**
+   * VP commission in basis points; feeds the intent's per-vault commissionFee.
+   */
+  commissionBps: number;
```
gbarkhatov: **Required field on a published public interface, for a value only the intent consumes.**

`PreparePeginParams` is exported from `@babylonlabs-io/ts-sdk@0.1.2`, so a required addition is breaking for external consumers. Note that `RegisterPeginParams.quotedCommissionBps` (`:457`) models the same quantity as optional, so the two now disagree on the same manager.

If the commission is read on-chain instead (see the note on `buildVaultIntent.ts:78`) this param disappears entirely, which is the better outcome. If it stays, either make it optional and omit the intent's commission fields when absent, or call it out in the release notes.

packages/babylon-ts-sdk/src/tbv/core/intent/vaultIntent.ts
```diff
@@ -0,0 +1,85 @@
...
+  challengerPks: string[]; // tag 0x0108; ascending byte-lex, sorted independently
+  vaults: VaultIntentVaultGroup[]; // tags 0x0109-0x010E; ascending htlcVout
+  // structure_type (0x01): ABSENT — constant unassigned in v22 (TYPE_???). Add when pinned.
+  // depositor_derivation_path (0x69): ABSENT — provider-injected at TLV-encode time from its
+  //   own account state (encoding contract still an open Ledger pin). [R: field deleted]
+  // council fields: NOT included — no assigned tags exist; the Assert:0 fee fix is
+  //   Ledger-side and any v23 tags are speculative. Re-add = two-line change when a real
+  //   v23 draft assigns them. [R: dropped, reverses rev-1 Option A]
```
gbarkhatov: **Review-thread markers left in shipped source.**

`[R]`, `[R: field deleted]`, `[R: dropped, reverses rev-1 Option A]` (L31, L34, L45, L49, L57) and `signet pin: 06-v22-review-findings` (also `buildVaultIntent.ts:3`) point at a review conversation nobody reading this file in six months will have. Same class as the no-audit-references convention.

The substance behind each is worth keeping — *why* `depositor_derivation_path` is absent, *why* the council fields were dropped — just drop the provenance tags. `v22 L1362-1366` is fine; that's a real locator into a document that exists.

packages/babylon-ts-sdk/src/tbv/core/intent/vaultIntent.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface VaultIntentSigner {
+  approveVaultIntent(intent: VaultIntent): Promise<void>;
+}
+
+/**
+ * Seam invariant [R]: no `deriveContextHash` call between an `approveVaultIntent`
+ * and the last intent-bound signature of that connection (DERIVE-while-loaded
+ * nullifies, v22 L2150-2151). Current flow satisfies it; the sites that would
+ * violate it if reordered: VP auth (ensureAuthenticatedVpClient.ts:81),
+ * WOTS/activation re-derives (useDepositFlow.ts:862, :984 — both after all
```
gbarkhatov: **Hardcoded cross-package line numbers, one already stale.**

`useDepositFlow.ts:862` is already wrong — that line sits inside a `minPrepeginDepth` comment; the `deriveContextHash` override is at `:877`. `ensureAuthenticatedVpClient.ts:81` and `PeginManager.ts:878` (L17) will drift the same way, and the SDK has no way to be checked against the app.

Name the functions instead: `ensureAuthenticatedVpClient`, the `postBroadcastBtcWallet` wrapper, `preparePeginCommit`. The invariant itself checks out, for what it's worth — VP auth runs before `runDepositorPresignFlow` (`payoutSigning.ts:94`), and a cached token refresh doesn't re-touch the wallet.

packages/babylon-ts-sdk/src/tbv/core/intent/buildVaultIntent.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export function buildVaultIntent(inputs: BuildVaultIntentInputs): VaultIntent {
+  const txid = inputs.prepeginTxid.toLowerCase();
+  if (!/^[0-9a-f]+$/.test(txid) || txid.length !== TXID_HEX_LENGTH) {
+    throw new Error(`buildVaultIntent: prepeginTxid must be 64 hex chars, got "${inputs.prepeginTxid}"`);
+  }
+  if (inputs.vaultAmounts.length === 0) {
+    throw new Error("buildVaultIntent: at least one vault amount is required");
+  }
+  if (inputs.timelockPegin <= 0 || inputs.timelockRefund <= 0) {
```
gbarkhatov: **Validation is looser than the contract the types document.**

- `htlcRefundTimelock` is documented as `72..4320 in v22` (`vaultIntent.ts:22`) but only checked `> 0` here.
- Neither timelock is range-checked against its u32 wire width, and `peginCsvTimelock`'s source is a `uint16` (`protocol-params-reader.ts`), so `> 65535` is representable in the input type but not on the wire.
- `normalizeXOnly` strips `0x` but not `0X`, so an uppercase-prefixed key throws "not a 32-byte x-only hex key" rather than normalizing — confusing given the test at `buildVaultIntent.test.ts:76` advertises uppercase normalization.

None of these is reachable from the current call sites. They're worth closing while this is the only consumer, since the whole point of the builder is to be the field-by-field checklist against the device spec.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -23,6 +23,7 @@ import * as bitcoin from "bitcoinjs-lib";
 import { Psbt, Transaction } from "bitcoinjs-lib";
 import { Buffer } from "buffer";
 
+import { computeMinPeginFee } from "@babylonlabs-io/babylon-tbv-rust-wasm";
```
gbarkhatov: Every other consumer in the SDK reaches `computeMinPeginFee` through `../primitives`, which re-exports it (`primitives/index.ts:86`) — including `assertWasmPeginSizing`, which calls the same function. This one goes straight to the WASM package and lands between the `bitcoinjs-lib` and `viem` blocks instead of in the grouped import below. eslint doesn't flag it, but it's the odd one out. Moot if you thread the fee through from `buildPrePeginPsbt` instead.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/__tests__/runDepositorPresignFlow.test.ts
```diff
@@ -543,4 +579,95 @@ describe("runDepositorPresignFlow", () => {
...
+      expect(wallet.approveVaultIntent).not.toHaveBeenCalled();
+      expect(
+        presignClient.requestDepositorPresignTransactions,
+      ).not.toHaveBeenCalled();
+      expect(
+        presignClient.submitDepositorPresignatures,
+      ).not.toHaveBeenCalled();
+    });
+
+    it("ignores vaultIntent for non-capability wallets", async () => {
```
gbarkhatov: The test's own comment concedes the assertion doesn't prove the name, and `expect(wallet.approveVaultIntent).toBeUndefined()` asserts on the fixture rather than on anything in `src/`.

What it actually verifies is real and worth keeping: the flow completes for a capability-less wallet even when a `vaultIntent` is passed. Rename to that and drop the fixture assertion — `submitDepositorPresignatures` having been called once is the whole proof.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -738,9 +756,11 @@ export class PeginManager {
       authAnchorHash,
     );
 
+    const { vaultIntent, ...commitTransaction } = commit;
```
gbarkhatov: Minor ordering: the `htlcVout === i` invariant the intent depends on (`vaultIntent.ts:5`) is checked at `:737-745`, i.e. after `preparePeginCommit` has already built *and device-approved* the intent whose `vaults[]` assumes it.

Tautological today — both sides derive the index from the same loop counter — so this is only about keeping the check meaningful if that ever stops being true. Moving the loop above the intent build inside `preparePeginCommit` costs nothing.

gbarkhatov (dismissed):
Approving — the seam is well shaped, the no-op-for-current-wallets claim holds, and the ordering test (derive -> approve -> sign) pins the right thing. Comments below are for follow-up rather than blockers, but two of them should be settled before the #2109 provider actually lands.

**Verified locally** at `65ed758`: SDK tests 89/89 across the three affected files, vault tests 17/17, `tsc --noEmit` clean for both packages, eslint clean on the changed SDK files. I also traced the seam invariant by hand — VP auth's `deriveContextHash` runs before `runDepositorPresignFlow`, and cached token refresh doesn't re-touch the wallet, so no derive lands between approval and the intent-bound signatures.

**Worth settling before #2109:**

- The intent's `commissionFee` comes from the *quoted* bps and is built before ETH registration, while the contract binds `quoted + 25`. If the device compares `commission_fee` exactly, a legal in-headroom drift refuses signing after BTC is committed.
- `supportsVaultIntent` fails open across the app's three `{...wallet}` spreads. If the provider exposes `approveVaultIntent` on the prototype, all three drop it and the seam silently no-ops — currently defended only by a comment.

**Worth a look now:**

- `sortedUnique` throws on an empty universal-challenger list, and nothing upstream enforces >= 1 — that's a new all-wallet abort on a config that works today.
- `peginMaxFee` is recomputed rather than threaded from the value `assertWasmPeginSizing` already asserted against the funded HTLC.
- `buildVaultIntent` throws on depositor-in-keeper-set where `deriveLocalChallengers` deliberately filters; pick one policy.
- The resume path (`usePayoutSigningState.ts:281`) passes no intent, so intent wallets hit the new throw and get a generic "unexpected error" with a Retry that can never succeed. The `#2110` text in the error never reaches the user.

The rest are comment-hygiene and validation-tightening nits.

PR #2140, opened 2026-07-24
fix(vault): share optimistic step completions across nested polling providers

jonybur:
## The bug

Reported by QA on the v3 `/vaults` page, with two pending deposits:

1. **"Submit WOTS Key" stayed after submitting.** The row kept the button, and clicking it again fired a fresh wallet approval for a submission that had already landed.
2. **"Sign Payouts" stayed after signing, permanently.** Both rows sat on "Signing required" for the rest of the session. Only a page reload cleared it.

## Root cause

The dashboard mounts a `PeginPollingProvider` over every activity. The continuation modal mounts a **second one, nested**, scoped to the viewed batch. Every action hook — `usePayoutSigningState`, `ResumeWotsContent` — runs under the *inner* provider, because `usePeginPolling` is a plain `useContext`. So the optimistic status written on success landed in state the dashboard row could not read, and the row kept offering the button it had just satisfied.

For payout signing this compounded into something permanent:

- The VP poll halts once every polled deposit reports `PendingDepositorSignatures`, which is exactly the reported state.
- That freezes `transactionsReady` at `true`.
- `applyTrackingOverrides` then discards the correctly-persisted `PAYOUT_SIGNED` as "stale or tampered" — behaving exactly as designed, on a bad input.

WOTS was a shorter version: a successful submission recorded **nothing anywhere** — no localStorage status, no optimistic marker. `getNextLocalStatus` has no `SUBMIT_WOTS_KEY` case. The row could only self-correct on the next 30s poll.

## The fix

Hoist optimistic step completions into an app-scoped store, `context/deposit/optimisticDepositState.ts`. This mirrors `terminalMilestones.ts` and `daemonTerminalEvents.ts`, which were already moved to module scope for this exact nesting — see the rationale comment in `PeginPollingContext`. Adds a WOTS completion marker, recorded on both the resume and first-run paths.

**This changes the scope of the signal, not the trust model.** The store is in-memory only, so a reload drops it and the VP stays ground truth. `applyTrackingOverrides` is untouched and its anti-tamper reconciliation still applies to persisted localStorage exactly as before. No critical-path file is modified — in particular `depositFlowSteps/payoutSigning.ts` is untouched, since the defect was entirely in UI state scoping.

## Not a v3 regression

The defect is identical in v2 and predates the v3 page. `PendingDepositSection` mounts the same outer provider and the same nested `PostDepositContinuationContent` inside it, and `PendingDepositCard` derives its button from the same pair of helpers. v2 only hid it behind a collapsed `ExpandablePanel` that defaults shut, so you had to expand before a per-row button existed on screen. v3 renders every row and its button unconditionally.

So this fixes v2 too, with no v2 edits. Flagging it so nobody bisects to #2095.

## Deliberately not included

`usePeginPollingQuery`'s `refetchInterval` returns `false` once every deposit reports `PendingDepositorSignatures`, permanently stopping the poll. After this PR that can no longer produce a *wrong* button, but a pending deposit that goes bad provider-side (expires, gets rejected) while the dashboard is open will keep a live button until reload — the same symptom, via a route this PR does not close.

Left out to keep this a clean, reviewable state-scoping change. Worth a follow-up: the fix is to drop the `pendingDepositorSignatures` arm of the stop condition and keep the terminal-error arm, since "waiting for your signature" is the one resolved state the user's own action invalidates. Cost is one batched request per vault provider per 30s, not per deposit.

## Testing

- 2695 tests pass, 0 lint errors, no new type errors.
- Two new regression tests were confirmed to **fail without the fix** by stashing the two source files and re-running:
  - `hides Sign Payouts on the dashboard when the signing modal's own nested provider records the completion`
  - `hides Submit WOTS Key once the submission resolves, before the poll clears needsWotsKey`
- Plus a negative control (a sibling deposit keeps its button), store unit tests, and coverage that the WOTS marker is recorded on success and not on failure.

## Review notes

- The store is module-scoped and outlives individual providers, so **any test touching optimistic status needs `resetOptimisticDepositState()` in `beforeEach`**. Added to the existing suite — without it the pre-existing payout tests leak into each other.
- `getSnapshot` must keep a stable reference between writes or `useSyncExternalStore` loops. Covered by a test.
- God-mode demo rows are unaffected: `getPollingResult` short-circuits demo ids before reaching any of this. That also means the demo panel cannot be used to verify this change.

### DIFF
--- a/services/vault/src/components/simple/ResumeDepositContent.tsx
+++ b/services/vault/src/components/simple/ResumeDepositContent.tsx
@@ -32,6 +32,7 @@ import { getVaultRegistryReader } from "@/clients/eth-contract/sdk-readers";
 import { computeDepositDerivedState } from "@/components/deposit/DepositSignModal/depositStepHelpers";
 import { usePayoutSigningState } from "@/components/deposit/PayoutSignModal/usePayoutSigningState";
 import { useDepositPollingResult } from "@/context/deposit/PeginPollingContext";
+import { markWotsSubmitted } from "@/context/deposit/optimisticDepositState";
 import { COPY } from "@/copy";
 import {
   DepositFlowStep,
@@ -417,6 +418,11 @@ export function ResumeWotsContent({
         btcWallet: btcWalletProvider,
         unsignedPrePeginTxHex: activity.unsignedPrePeginTx,
       });
+
+      // Recorded regardless of mount: the submission landed, so the dashboard
+      // row must stop offering "Submit WOTS Key" even if the user already
+      // closed this modal. The store is app-scoped, not tied to this tree.
+      markWotsSubmitted(activity.id);
 
       if (mountedRef.current) {
         setLoading(false);
--- a/services/vault/src/context/deposit/computeDepositPollingResult.ts
+++ b/services/vault/src/context/deposit/computeDepositPollingResult.ts
@@ -21,7 +21,7 @@ import { isVaultOwnedByWallet } from "../../utils/vaultWarnings";
 /** Optimistic override wins over persisted `pendingPegins` status. */
 function resolveLocalStatus(
   depositId: string,
-  optimisticStatuses: Map<string, LocalStorageStatus>,
+  optimisticStatuses: ReadonlyMap<string, LocalStorageStatus>,
   pendingPegins: PendingPeginRequest[],
 ): LocalStorageStatus | undefined {
   const pendingPegin = pendingPegins.find((p) => p.id === depositId);
@@ -35,7 +35,7 @@ function resolveLocalStatus(
  */
 function resolveRefundBroadcastAt(
   depositId: string,
-  optimisticRefundBroadcastAt: Map<string, number>,
+  optimisticRefundBroadcastAt: ReadonlyMap<string, number>,
   pendingPegins: PendingPeginRequest[],
 ): number | undefined {
   return (
@@ -76,8 +76,15 @@ export interface DepositPollingInputs {
    */
   activationDeadlinePassed: boolean;
   isLoading: boolean;
-  optimisticStatuses: Map<string, LocalStorageStatus>;
-  optimisticRefundBroadcastAt: Map<string, number>;
+  optimisticStatuses: ReadonlyMap<string, LocalStorageStatus>;
+  optimisticRefundBroadcastAt: ReadonlyMap<string, number>;
+  /**
+   * Deposits whose WOTS submission resolved in this page session. The VP poll
+   * keeps reporting `needsWotsKey` until the daemon advances, so without this
+   * the row would re-offer "Submit WOTS Key" — and a second click re-runs the
+   * whole derivation, including a fresh wallet popup, for a no-op submission.
+   */
+  wotsSubmitted: ReadonlySet<string>;
   btcPublicKey: string | undefined;
 }
 
@@ -102,6 +109,7 @@ export function computeDepositPollingResult(
     isLoading,
     optimisticStatuses,
     optimisticRefundBroadcastAt,
+    wotsSubmitted,
     btcPublicKey,
   } = inputs;
   const depositId = activity.id;
@@ -129,6 +137,13 @@ export function computeDepositPollingResult(
   const transactionsReady = justSignedPayoutsThisSession
     ? false
     : (pendingDepositorSignatures?.has(depositId) ?? false);
+
+  // Same shape as the payout suppression above: a step this session watched
+  // resolve outranks the poll snapshot that has not caught up yet.
+  const justSubmittedWotsThisSession = wotsSubmitted.has(depositId);
+  const stillNeedsWotsKey = justSubmittedWotsThisSession
+    ? false
+    : needsWotsKey?.has(depositId);
 
   // Cache is OR'd with live count: on refresh, cached txids are
   // filtered out of polling, so the live map is empty for them.
@@ -204,7 +219,7 @@ export function computeDepositPollingResult(
     localStatus,
     transactionsReady,
     isInUse: activity.isInUse,
-    needsWotsKey: needsWotsKey?.has(depositId),
+    needsWotsKey: stillNeedsWotsKey,
     pendingIngestion: pendingIngestion?.has(depositId),
     prePeginBroadcastConfirmed,
     prePeginBroadcastSeen,
--- a/services/vault/src/context/deposit/optimisticDepositState.ts
+++ b/services/vault/src/context/deposit/optimisticDepositState.ts
@@ -1,0 +1,105 @@
+/**
+ * App-scoped store for "the user completed this step in this page session".
+ *
+ * Every mounted `PeginPollingProvider` reads the same snapshot. That scoping is
+ * the whole point: the dashboard section mounts a provider over every activity,
+ * while the continuation modal mounts its own provider scoped to the viewed
+ * batch — nested inside that section. The hooks that drive an action
+ * (`usePayoutSigningState`, `ResumeWotsContent`, …) run under the *modal's*
+ * provider, so per-provider state meant the dashboard row that offered the
+ * button never learned the action had succeeded and kept re-offering it. Same
+ * reasoning, and same fix, as the tracking stores in `terminalMilestones.ts`
+ * and `daemonTerminalEvents.ts`.
+ *
+ * This changes the SCOPE of the signal, not the trust model. It records only
+ * outcomes this session watched resolve — it is never persisted, so a reload
+ * drops it and the vault provider's polled status is ground truth again. The
+ * anti-tamper reconciliation in `applyTrackingOverrides`, which distrusts a
+ * localStorage status the VP contradicts, is deliberately left untouched.
+ */
+
+import type { LocalStorageStatus } from "../../models/peginStateMachine";
+
+export interface OptimisticDepositState {
+  /** Per-deposit status set the moment its action resolved. */
+  statuses: ReadonlyMap<string, LocalStorageStatus>;
+  /** Companion `Date.now()` for REFUND_BROADCAST, anchoring its suppression TTL. */
+  refundBroadcastAt: ReadonlyMap<string, number>;
+  /**
+   * Deposits whose WOTS public key submission resolved. WOTS has no
+   * `OffChainTrackingStatus` of its own — it is not a status the deposit rests
+   * in — so the completion is recorded as its own set rather than by widening
+   * the status enum.
+   */
+  wotsSubmitted: ReadonlySet<string>;
+}
+
+const EMPTY_STATE: OptimisticDepositState = {
+  statuses: new Map(),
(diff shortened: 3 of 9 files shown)


services/vault/src/context/deposit/computeDepositPollingResult.ts
```diff
@@ -130,6 +138,13 @@ export function computeDepositPollingResult(
     ? false
     : (pendingDepositorSignatures?.has(depositId) ?? false);
 
+  // Same shape as the payout suppression above: a step this session watched
+  // resolve outranks the poll snapshot that has not caught up yet.
+  const justSubmittedWotsThisSession = wotsSubmitted.has(depositId);
+  const stillNeedsWotsKey = justSubmittedWotsThisSession
```
gbarkhatov, line 144: **[Medium] The marker never expires and has no invalidation path.** This hard-forces `needsWotsKey = false` for the rest of the page session. If the VP ever legitimately re-reports `PENDING_DEPOSITOR_WOTS_PK` after our call resolved — daemon rollback, key rejected post-acceptance, a lost write behind a 200 — the row collapses to a non-actionable wait (`getDisplay` falls through to "waiting for payout prep", pending badge, no CTA) with no route back short of a reload. That's the mirror image of the bug being fixed, and unlike the payout case there's no localStorage counterpart to reconcile against.

The codebase already has the precedent for why a suppression marker must not be sticky: `isRefundBroadcastWithinTtl` at `peginStateMachine.ts:473`, with the rationale spelled out directly above it.

Better than a TTL here: clear the marker the first time a poll observes the deposit **absent** from `needsWotsKey`. It fires once, on exactly the transition the marker exists to bridge, and it gives eviction for free. One subtlety for whoever implements it — gate on `needsWotsKey !== undefined`, otherwise the pre-first-poll `undefined` clears a marker written during the deposit flow before any poll has resolved.

services/vault/src/context/deposit/computeDepositPollingResult.ts
```diff
@@ -204,7 +219,7 @@ export function computeDepositPollingResult(
     localStatus,
     transactionsReady,
     isInUse: activity.isInUse,
-    needsWotsKey: needsWotsKey?.has(depositId),
+    needsWotsKey: stillNeedsWotsKey,
```
gbarkhatov, line 222: [Low] `getPeginState` forwards this value straight into the `applyTrackingOverrides` reconciliation `vpState` (`peginStateMachine.ts:357`), so the suppressed value also masks one of the three anti-tamper inputs there.

Inert today, and I checked both directions: with `needsWotsKey` suppressed the protocol state has already collapsed to `[]`, so both reconciliation branches return `[]` — and passing the raw value in would return `[]` too. So I'd rather not add churn to the call site.

But the PR body's "`applyTrackingOverrides` is untouched and its anti-tamper reconciliation still applies to persisted localStorage exactly as before", and the same claim in the `optimisticDepositState` doc comment, assert a decoupling the code doesn't strictly have — the function is untouched, its VP-side input is not. Worth narrowing the wording so a later reader doesn't rely on the stronger version of the claim while changing something around it.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -418,6 +419,11 @@ export function ResumeWotsContent({
         unsignedPrePeginTxHex: activity.unsignedPrePeginTx,
       });
 
+      // Recorded regardless of mount: the submission landed, so the dashboard
+      // row must stop offering "Submit WOTS Key" even if the user already
+      // closed this modal. The store is app-scoped, not tied to this tree.
+      markWotsSubmitted(activity.id);
```
gbarkhatov, line 425: [Low] Timing change in the split-batch continuation modal — worth a QA pass rather than a code change.

`PostDepositContinuationView` picks its rendered branch by actionability (`isActionable` → `getPollingResult`, `PostDepositContinuationView.tsx:120-148`), and every `Resume*Content` auto-fires on mount via `useRunOnce`. With this marker, vault A leaves actionable the moment its submission resolves instead of at the next poll, so sibling B's `ResumeWotsContent` mounts — and pops a wallet approval — up to one poll interval earlier, immediately after the approval that just resolved. Same end state, and it brings WOTS in line with how the payout path already behaves in that view (its optimistic write always reached the modal's own provider), so I expect it's fine. But nothing covers it: please exercise a 2-vault split end to end.

While you're here — the placement of this call above the `mountedRef.current` gate is load-bearing for a second reason beyond the one in the comment. React batches the store notify, so `onSuccess()` still runs before the re-target unmounts this tree; publishing after the gate, or synchronously flushing, would drop the dashboard refetch. Worth a clause in the comment.

services/vault/src/context/deposit/optimisticDepositState.ts
```diff
@@ -0,0 +1,105 @@
...
+ * while the continuation modal mounts its own provider scoped to the viewed
+ * batch — nested inside that section. The hooks that drive an action
+ * (`usePayoutSigningState`, `ResumeWotsContent`, …) run under the *modal's*
+ * provider, so per-provider state meant the dashboard row that offered the
+ * button never learned the action had succeeded and kept re-offering it. Same
+ * reasoning, and same fix, as the tracking stores in `terminalMilestones.ts`
+ * and `daemonTerminalEvents.ts`.
+ *
+ * This changes the SCOPE of the signal, not the trust model. It records only
+ * outcomes this session watched resolve — it is never persisted, so a reload
```
gbarkhatov, line 15: [Low] Not persisting is the right call for the trust model, but it leaves a residual the ticket should record rather than close as fully fixed: a reload inside the daemon-lag window re-offers "Submit WOTS Key", and clicking it still costs a wallet approval — `ResumeWotsContent` derives the vault root before `submitWotsPublicKey` is reached, and that call then returns early on the idempotent path (documented at `useDepositFlow.ts:1013`).

Much narrower than what QA reported — one poll interval after a deliberate reload, versus the whole session unconditionally — but it's the same symptom, so the bug report shouldn't be closed as "button can no longer appear after submitting".

services/vault/src/context/deposit/optimisticDepositState.ts
```diff
@@ -0,0 +1,105 @@
...
+export function getOptimisticDepositState(): OptimisticDepositState {
+  return currentState;
+}
+
+export function setOptimisticDepositStatus(
+  depositId: string,
+  status: LocalStorageStatus,
+  refundBroadcastAt?: number,
+): void {
+  publish({
```
gbarkhatov, line 77: [Nit] Guard asymmetry: `markWotsSubmitted` below early-returns on an unchanged write, this one always publishes. Re-setting the same status therefore re-renders every mounted provider and invalidates the `getPollingResult` memo for no state change. Guard both or neither — here that's an early return when the stored status already equals `status` and `refundBroadcastAt` adds nothing new.

gbarkhatov (approved):
Approving — the diagnosis is right, the fix is the correct shape and minimal, and the write-up is unusually candid about what it leaves open. Nothing below blocks; finding 1 is the one I'd like an answer on.

**What I verified locally** (checkout at `91ed16a7`, so reads match the diff):

- Ran the four affected suites: 53/53 pass.
- Traced the new suppression through the SDK state machine: with `needsWotsKey` forced false and `pendingIngestion === false`, `getPeginProtocolState` returns `[]` → `NONE`, so the suppression cannot fall through to re-offering the broadcast. That holds because `needsWotsKey` and `pendingIngestion` are mutually exclusive per poll cycle — every branch in `applyPerDepositStatus` deletes from one while adding to the other. Worth knowing that the new test's assumption is load-bearing on that invariant: if a VP ever reported both, suppression would unmask `SIGN_AND_BROADCAST_TO_BITCOIN` and prompt a re-broadcast of an already-broadcast tx. A `computeDepositPollingResult` case pinning `wotsSubmitted` + `pendingIngestion: true` → no broadcast action would cost one test and close that off.
- Confirmed the store's write sites only fire after a resolved `submitWotsPublicKey`, on both paths, and that the resume path deliberately records before the `mountedRef` gate.

**Findings** (inline): (1) medium — the WOTS marker has no invalidation path; (2) low — the "`applyTrackingOverrides` untouched" claim is narrower than stated; (3) low — untested timing change in the split-batch continuation modal, worth a QA pass; (4) low — residual after reload, worth recording on the ticket rather than closing it as fully fixed; (5) nit — guard asymmetry between the two writers.

**Also good:** the module store mirrors `terminalMilestones.ts` / `daemonTerminalEvents.ts` down to the "Test-only reset" convention; the `ReadonlyMap`/`ReadonlySet` tightening on `DepositPollingInputs` is a real improvement; no critical-path file touched; the snapshot-identity test and the per-render `captured` capture in the nested-provider test (with its comment on why holding the pre-action context object would pass regardless of the fix) are the parts that usually get missed. Test coverage gap besides the one above: the first-run `markWotsSubmitted` in `useDepositFlow` is uncovered — only the resume path has a test.

Agree with deferring the `refetchInterval` stop-condition fix; it's genuinely separate and the residual symptom is named precisely.

services/vault/src/context/deposit/computeDepositPollingResult.ts
jonybur: Confirmed, and fixed in #2145.

You undersold it in one place: `PostDepositContinuationView` gates `ResumeWotsContent` on the same suppressed `availableActions`, and that is its only mount site, so the continuation modal loses its resume branch too. There is no second door - reload really is the only way back.

One correction the other way: the payout marker has the same stickiness. `getPeginState` forwards the already-suppressed `transactionsReady` into `applyTrackingOverrides`'s `vpState`, so the `:436` reconciliation cannot fire against a same-session `PAYOUT_SIGNED` either. WOTS is not being held to a lower bar than what already shipped.

On the proposed fix - I built it first, then dropped it. Two defects, both reproduced with throwaway tests:

1. **Absence is not an affirmative observation.** `polledIds` is populated unconditionally, but a deposit whose VP call errored (`onWholeBatchError`, `onMissing`, `onDuplicate`) is absent from `needsWotsKey` while still being in `polledIds`. A single transient blip inside the lag window destroys the marker, and the next good poll re-offers the button - the exact symptom #2140 exists to fix.
2. **It races across the two nested providers.** The store is app-scoped by design, so the first provider to observe the advance retires the marker for both. The publish notifies the sibling, which recomputes while still holding a pre-advance snapshot, and its row re-offers the button for up to a poll interval.

It also does not buy recovery, only eviction. Clearing on advance fires only when the daemon *advances*, and in the likeliest triggers - a write lost behind a 200, a post-acceptance rejection - the daemon never leaves `PENDING_DEPOSITOR_WOTS_PK`, so the deposit is never absent and the marker never clears.

So #2145 uses the TTL you pointed at with `isRefundBroadcastWithinTtl`, whose comment already documents why a suppression marker must never be sticky. A clock is read identically in every provider and cannot disagree with itself. Ten minutes is the one tunable - happy to move it.

PR #2141, opened 2026-07-24
refactor(vault): read protocol params non-blockingly in the pegin polling tree

jonybur:
Collapses four `PeginPollingProvider` mounts into one app-scoped instance, plus the decoupling that makes it possible.

## Why two commits

The single provider has to mount above the routes that own `ProtocolParamsProvider` — but that provider is **blocking** (renders a spinner in place of its children, an error panel on failure) and `useProtocolParamsContext` throws outside it. Hoisting directly would have gated every route and the wallet UI behind three contract reads.

1. **`read protocol params non-blockingly`** — the polling tree reads the two params queries directly, using the same query keys. Shared cache, no extra request. `ProtocolParamsProvider` doesn't move or split.
2. **`collapse the pegin polling tree to a single provider mount`** — `AppPeginPollingProvider` mounts once in `RootLayout`. A second mount is now a hard error in dev, a logged error in production.

Two of the four mounts nested, which is why an optimistic completion recorded in the continuation modal was invisible to the dashboard row behind it. #2140 patched that with an app-scoped store; this removes the reason that patch is load-bearing.

## What changed for the user

Params can now be unresolved when the polling tree renders, so the required Pre-PegIn depth is `number | undefined` end to end. **Unknown depth withholds a conclusion rather than defaulting to one** — it can never assert a Bitcoin depth the chain hasn't reached. Unresolved params present as loading; *failed* params present as failed.

A params failure suppresses CTAs rather than rendering a message. That's deliberate — depth and `tRefund` gate Broadcast and Refund, and offering an action derived from params we couldn't read is worse than offering none. The user-facing surface stays `ProtocolParamsProvider`'s error panel.

The VP poll, mempool reads and signing notifications now run on every route rather than only where a deposit surface is mounted. Not started for a geo-blocked session or a disconnected wallet.

## Testing

2867 pass, 0 lint errors, typecheck clean. Both behavioural review follow-ups in `de78838` are mutation-verified — reverting either makes its new test fail.

## Known

`PostDepositContinuationView.tsx:115-124` refetches on a `refetch` identity re-created per poll, so it may re-fire more than intended. Statically traced, not observed.


services/vault/src/context/deposit/computeDepositPollingResult.ts
```diff
@@ -240,12 +257,15 @@ export function computeDepositPollingResult(
...
+    confirmations ??
+    (cachedAtDepth && requiredDepth !== undefined ? requiredDepth : null);
 
   return {
     depositId,
     loading: isLoading,
-    error: errors?.get(depositId) ?? null,
+    // A per-deposit VP error is more specific, so it wins; the params failure
+    // is the fallback so an outage is never silently swallowed.
+    error: errors?.get(depositId) ?? protocolParamsError,
```
gbarkhatov, line 305: **[Blocking-ish] This field is an action-suppression gate, not a display channel.** I traced every consumer: `getActionStatus` (`actionStatus.ts:92` — `if (error || !actionButton) return { type: "noAction" }`) is the only reader, and `noAction` renders neither a button nor a message. `getSectionActionRequiredLabel` routes through the same helper, so the section's action-required badge goes quiet too, and `PendingDepositCard` doesn't destructure `error` at all.

So a params-query failure removes Sign Payouts, Submit WOTS Key, Broadcast, Activate and Refund from **every** deposit with nothing shown. The PR body describes this as the failure "surfacing… instead of leaving rows frozen with nothing shown to the user" — that's inverted. Worth noting that only the broadcast and refund actions have any dependency on protocol params (via depth and `tRefund` maturity); payout signing and activation would work fine, so a global query failure is being funnelled through a field whose consumers were built for per-deposit VP failures.

Latent today (the blocking provider blanks the subtree first), which is exactly why the semantics are cheaper to settle here than to discover after PR 2. Two coherent options: keep `error` per-deposit and carry the params failure on its own field with a visible banner, or accept the blackout deliberately and give it a rendered surface. Either way the decision should be explicit — right now the field is doing something its own doc comment ("so a params outage presents as a loud failure") doesn't describe.

services/vault/src/context/deposit/PeginPollingContext.tsx
```diff
@@ -471,7 +471,10 @@ export function PeginPollingProvider({
         activationDeadlinePassed: activationDeadlinePassedIds.has(
           activity.id.toLowerCase(),
         ),
-        isLoading,
+        // Params still resolving is a loading state, not a resolved "depth
+        // unknown" — otherwise a cold load reads as a stalled deposit.
+        isLoading: isLoading || !params.ready,
```
gbarkhatov: [Medium] On a params **failure** this latches: the three retries exhaust, `data` stays undefined, `ready` never flips, so every deposit reports `loading: true` permanently — alongside a non-null `error`. A result object asserting both is contradictory, and the consumers take the loading branch:

- `PendingDepositCard.tsx:144` — `step = loading ? null : …`, so no step, no progress bar, and `btcConfirmationSummary` stays null (no countdown).
- `PostDepositContinuationView.tsx:257` — `hasMissingOrLoadingVault` keys on `result.loading`, so the continuation modal sits on `AWAIT_BTC_CONFIRMATION` indefinitely.

Combined with the error being invisible, that's precisely the frozen row this PR set out to eliminate. The comment here is right that *resolving* params is a loading state; a *failed* params load is not, and the two shouldn't collapse into the same flag. Suggest `isLoading: isLoading || (!params.ready && !params.error)` so the failure can present as a failure.

One genuine upside of the current gate, worth preserving whatever you change: `useSigningRequiredNotifications` skips on `result.loading`, so a params outage can't fire a browser notification for an action the UI won't offer.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
```diff
@@ -0,0 +1,84 @@
...
+    (offchainParamsVersion?: number): number | undefined =>
+      offchainParamsVersion !== undefined
+        ? offchainParams?.byVersion.get(offchainParamsVersion)?.tRefund
+        : undefined,
+    [offchainParams],
+  );
+
+  return {
+    ready: config !== undefined && offchainParams !== undefined,
+    error: configError ?? offchainError,
```
gbarkhatov: [Medium] `error` isn't gated on `ready`, so the suppression in the other comment can fire while the data is present and perfectly usable. React Query v5 retains `data` when a **refetch** fails — confirmed in the installed types, `QueryObserverRefetchErrorResult` declares both `data: TData` and `error: TError` — so `ready === true` with `error !== null` is a real state, not a theoretical one. Trigger: a mount after the 5-minute `staleTime` refetches, the refetch fails its 3 retries, and now every row carries a params error while the depth it needs is cached and correct.

Reachability depends on PR 2's shape: today the blocking provider replaces the subtree on `allError`, so the polling tree is gone. It bites to the degree PR 2 stops wrapping the rows in the blocking provider.

Fix is one line and matches this hook's own philosophy of only withholding when a value is genuinely unknown: `error: ready ? null : (configError ?? offchainError)`.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
```diff
@@ -0,0 +1,84 @@
...
+  /** True once both queries resolved; every resolver below is live. */
+  ready: boolean;
+  /** First query error, or null. Surfaced per-deposit rather than swallowed. */
+  error: Error | null;
+  /** On-chain activation window. `undefined` until {@link ready}. */
+  pegInActivationTimeout: bigint | undefined;
+  /**
+   * Required Pre-PegIn confirmation depth (`minPrepeginDepth`), pinned to the
+   * deposit's registered offchain-params version, or the latest version when it
+   * has none (pre-sign). Same rule as `useRequiredPrePeginDepthResolver`, which
```
gbarkhatov: [Low] The `pinned ?? latest` rule now exists twice, and the older copy's docs no longer tell the truth. `useRequiredPrePeginDepth.ts:9` still reads "Single source of this rule — also consumed by `PeginPollingContext.getRequiredPrePeginDepth`", which this PR falsifies; the repo convention is to update comments that describe old behavior. Related: `useRequiredPrePeginDepthResolver` now has **zero importers outside its own module** (it's still called internally at `useRequiredPrePeginDepth.ts:29`), so it should lose its `export` — not be deleted.

The duplication has a concrete drift surface rather than being merely untidy: `BtcConfirmationDetailContainer` takes `requiredDepth` from the blocking resolver while its `confirmations` come from the polling tree computed against this copy, so any divergence renders a count and a threshold sourced from two different implementations of the same rule.

I'd extract a pure `resolveMinPrepeginDepth(config, allOffchainParams, version)` that both hooks call, rather than reimplementing `useRequiredPrePeginDepth` over this hook — the latter widens its return to `number | undefined` for three components (`DepositCardShell`, `ResumeDepositContent`, `PostDepositContinuationView`) that legitimately render under the blocking provider and feed `BtcConfirmationDetailData.requiredDepth: number`. Keep their guaranteed-value contract; share the rule, not the loading model.

services/vault/src/context/deposit/computeDepositPollingResult.ts
```diff
@@ -240,12 +257,15 @@ export function computeDepositPollingResult(
   // is past depth. Treat that as "at least requiredDepth" so consumers don't
   // see a regression.
   const reportedConfirmations =
```
gbarkhatov, line 295: [Nit] Worth one comment on the asymmetry, because it's a deliberate split that reads like an oversight: with an unknown depth a cached at-depth txid still yields `prePeginBroadcastConfirmed === true` above, while the count here withholds to `null`. Both are right — the cache entry was written when the threshold *was* known and depth never rewinds, so the boolean is sound, but the number isn't recoverable without the threshold.

It does make one PR-body bullet slightly overstated: "`prePeginBroadcastConfirmed` requires a defined threshold" — it doesn't when the cache already answered. And the new test covers only the uncached path (`confirmedTxids: new Set()`), so the cached-plus-unknown-depth combination — where a card can assert protocol depth while showing no count — is the one case in this area with no coverage.

gbarkhatov (dismissed):
Approving the mechanical refactor — sharing the query keys instead of splitting the blocking provider is the right call, and the withhold-rather-than-default discipline is applied consistently. One thing should be settled before PR 2 makes it reachable: what `DepositPollingResult.error` means (finding 1). It's a one-line semantics decision now versus a user-visible action blackout later.

**Verified independently** (checkout at `d647b6ed`):

- **Behavior-neutrality holds.** `ProtocolParamsProvider` gates children on `configLoading || ucLoading || offchainLoading` (`ProtocolParamsContext.tsx:130`) — including the same `offchainParamsQueryOptions()` this hook reads — and all three `PeginPollingProvider` mounts sit under it (`PendingDepositSection:131`, `VaultsLifecycleSections:467`, plus the modal's nested one, which couldn't render today without throwing). So `ready` is always `true` on current mounts, and the new branches really are unreachable.
- **Shared cache confirmed:** identical keys, no second fetch. Nothing calls `resetQueries`/`removeQueries` on these keys, so `ready` can't flip false under a live mount.
- 67 tests pass across the affected suites; `tsc --noEmit` on `services/vault` is clean.
- `getTxidsCrossingThreshold` already skips an `undefined` threshold, so an unresolved depth cannot poison the confirmed-at-depth localStorage cache — that path was already safe, which is why the new `requiredDepth: number | undefined` doesn't leak into the persisted cache.
- The params error is deliberately **not** routed through `vpTerminalError`, so it can't mislabel a row "Failed". That's the right seam and worth keeping in PR 2.

**Findings** (inline): (1) blocking-ish — `error` is an action-suppression gate, not a display channel, so a params failure silently strips every CTA; (2) medium — `loading` is permanently true on a params failure, contradicting the error set beside it; (3) medium — `error` isn't gated on `ready`, so the suppression can fire while the data is present and fine; (4) low — the depth rule now exists twice and the old resolver's export/JSDoc are stale; (5) nit — the cached-at-depth asymmetry deserves a comment.

On the test for the error path: it asserts the field value (`result.error` toBe the params error) rather than any outcome, so it pins the premise instead of checking it. Once the semantics are settled, the test worth having is the outcome one — e.g. that a params failure does *or* does not leave `getActionStatus` at `noAction`, whichever you decide is correct.

One correction to item 3 as filed: `useRequiredPrePeginDepthResolver` isn't dead — it's still called internally by `useRequiredPrePeginDepth` at `useRequiredPrePeginDepth.ts:29`. What died is its *export*: zero importers outside its own module now. So drop the `export` keyword rather than the function. Details inline.

The test table's honesty about the NaN-coercion case passing either way is the right instinct — that kind of note is what makes a fails-without-fix table trustworthy.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * Non-blocking access to the protocol params the peg-in polling tree needs.
+ *
+ * Mirrors the `pegInConfig` and `allOffchainParams` queries owned by
+ * ProtocolParamsContext and shares their React Query cache via the same keys,
+ * so this adds no extra request. Unlike `useProtocolParamsContext` it does NOT
+ * block: `PeginPollingProvider` is mounted above the routes that own the
```
jrwbabylonlab, line 7: Present tense, but at HEAD this provider is a direct child of the blocking one (`PendingDepositSection.tsx:131`, `VaultsLifecycleSections.tsx:467`), so the new unresolved/error branches are all unreachable. Worth phrasing as intent — `PendingDepositSection.tsx:124-129` also still credits `useProtocolParamsContext`.

services/vault/src/context/deposit/computeDepositPollingResult.ts
```diff
@@ -139,9 +166,14 @@ export function computeDepositPollingResult(
   const confirmations = prePeginCanonical
     ? prePeginConfirmationsByTxid.get(prePeginCanonical)
     : undefined;
+  // An unknown `requiredDepth` can only withhold "confirmed", never assert it:
+  // comparing against a defaulted threshold would claim protocol depth the
+  // chain may not have reached.
   const prePeginBroadcastConfirmed =
     cachedAtDepth ||
```
jrwbabylonlab, line 208: `cachedAtDepth ||` short-circuits ahead of the new guard, so the boolean still asserts at-depth on an unknown threshold while the count withholds.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
jonybur: Fixed — took your line verbatim.

```js
error: ready ? null : (configError ?? offchainError),
```

You were right that it is real rather than theoretical: React Query v5 retains `data` on a failed *refetch*, so `ready === true` with `error !== null` is a reachable state, and the hook was reporting an error over a depth it could already resolve. That contradicted the philosophy in its own header.

New test pins it (`usePeginPollingProtocolParams.test.ts`, "does not report a refetch failure while the cached params stay usable"): seed both query keys, fail the refetch, assert `ready` stays true, `error` stays null, and the pinned depth still resolves. Confirmed load-bearing — against the old `configError ?? offchainError` it fails.

services/vault/src/context/deposit/PeginPollingContext.tsx
jonybur: Fixed, with your suggested line:

```js
isLoading: isLoading || (!params.ready && !params.error),
```

Unresolved is still a loading state; failed is not. Test added asserting `loading === false` on the params-error path, plus its sibling asserting `loading === true` while merely resolving so the two halves are pinned separately. The first is mutation-verified: against `isLoading || !params.ready` it fails.

Kept the upside you flagged — `useSigningRequiredNotifications` skips on `result.loading`, which no longer covers the failure case. Worth noting it also skips on nothing else, so a params outage can now fire a notification for an action the row will refuse. That is the `error` suppression doing its job one layer down rather than the loading flag doing it by accident, which I think is the right place for it, but flagging it since it changed.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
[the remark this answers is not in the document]
jonybur: Fixed. `loading` and `error` can no longer both be asserted: `isLoading: isLoading || (!params.ready && !params.error)`, so a terminal params failure presents as a failure instead of latching the spinner. Test added asserting `loading === false` on that path — the combination you noted was untested is now pinned.

services/vault/src/context/deposit/computeDepositPollingResult.ts
jonybur: Settled deliberately, and the code now says so instead of the opposite.

You traced it correctly: `getActionStatus` is the only reader, `noAction` renders neither button nor message, and `PendingDepositCard` never destructures `error`. I am keeping the suppression rather than adding a display channel, for one reason — depth and `tRefund` maturity gate Broadcast and Refund, and offering an action derived from params we could not read is worse than offering none. You are right that payout signing and activation going quiet with them is over-broad; it is the safe direction of over-broad, and splitting per-action gating is not something I want to introduce in a refactor PR.

What I am not keeping is the doc comment that described this as the opposite. `protocolParamsError` now reads as a suppression gate, names `getActionStatus` as its only consumer, and points at `ProtocolParamsProvider` error panel as the actual user-facing surface — it gates on a superset of these same query keys, so in the current tree a row cannot be silently stripped without that panel already being up.

One honest gap while verifying that superset claim: `ProtocolParamsContext.tsx:141` gates on `allError || !configData` and never on `!offchainParamsData`, and `getOffchainParamsByVersion` returns `undefined` when that data is absent. A paused offline query (`isPending` true, `isFetching` false) therefore renders children with an unresolved version map. Unrelated to this PR and pre-existing, but it is the one seam where the superset does not strictly hold. Happy to fix it here if you would rather not leave it.

If you would rather have the banner after all, say so and I will do it as a follow-up rather than widen this one.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
jonybur: Both concrete parts fixed; the extraction named rather than done.

- `useRequiredPrePeginDepthResolver` loses its `export` — you were right that it is still called internally at `useRequiredPrePeginDepth.ts:29`, so dropping the keyword rather than the function. Confirmed zero importers outside the module.
- The JSDoc no longer claims to be the single source of a rule this tree also implements. It now says what it actually is: the blocking half, for components typed against a guaranteed `number`, with the non-blocking twin named and the reason they are separate stated — sharing the loading model would widen the return to `undefined` for three components that legitimately cannot receive it, which was your point.

On the extraction itself: I agree `resolveMinPrepeginDepth(config, allOffchainParams, version)` is the right end state, and the drift surface you named via `BtcConfirmationDetailContainer` is real. I have named it as the open cleanup in the comment rather than doing it here — it touches the blocking resolver three more components depend on, and I would rather that be its own reviewable change than a fourth thing in a refactor PR. Say the word if you want it in this one.

services/vault/src/context/deposit/computeDepositPollingResult.ts
jonybur: Comment added, and the coverage gap you named is closed.

The `why:` now spells out that the split is deliberate: the cache entry was only ever written while the threshold WAS known and depth never rewinds, so the boolean stays sound without re-reading it, while the number is not recoverable without inventing the very threshold we are missing.

New test covers the cached-plus-unknown-depth case from the boolean side — `confirmedTxids` non-empty, `requiredDepth` undefined, asserting the state machine reaches its ingesting branch (which fires solely on `prePeginBroadcastConfirmed`) while `prePeginConfirmations` stays null. So both halves of the asymmetry are pinned in one test rather than only the count.

You are right about the PR-body bullet. It is corrected in the rewritten body — `prePeginBroadcastConfirmed` requires a defined threshold *or* a cache entry that already answered.

services/vault/src/context/deposit/computeDepositPollingResult.ts
jonybur: Covered. New test in `computeDepositPollingResult.test.ts` takes the boolean side: `confirmedTxids: new Set([CANONICAL_PREPEGIN])` with `requiredDepth: undefined`, asserting the ingesting branch is reached (it keys solely on `prePeginBroadcastConfirmed`) while the count still withholds to null. Same case as `:177` from the other direction, so the short-circuit is now pinned rather than incidental.

The short-circuit itself stays — it is correct, and the `why:` above it now explains why the two sides legitimately disagree.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
jonybur: True at HEAD now, rather than reworded. The follow-on commit hoists the single provider to `RootLayout.tsx:419`, above `<Outlet>` and `<SimpleDeposit>`, so the present tense describes the tree as it actually is — no `ProtocolParamsProvider` above it, and `ready` genuinely false on first render.

`PendingDepositSection.tsx` no longer credits `useProtocolParamsContext` for polling either; it keeps its own `ProtocolParamsProvider` only for the children that need the guaranteed-value contract, and says so.

jonybur:
@greptile-apps

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  const resolveRequiredPrePeginDepth = useCallback(
+    (offchainParamsVersion?: number): number | undefined => {
+      if (!config || !offchainParams) return undefined;
+      const pinned =
+        offchainParamsVersion !== undefined
+          ? offchainParams.byVersion.get(offchainParamsVersion)
+              ?.minPrepeginDepth
+          : undefined;
+      return pinned ?? config.offchainParams.minPrepeginDepth;
```
jrwbabylonlab: Silent fallback: a skipped pinned version confirms at the wrong depth, and that gets persisted.

services/vault/src/context/deposit/computeDepositPollingResult.ts
```diff
@@ -240,12 +272,15 @@ export function computeDepositPollingResult(
...
+    confirmations ??
+    (cachedAtDepth && requiredDepth !== undefined ? requiredDepth : null);
 
   return {
     depositId,
     loading: isLoading,
-    error: errors?.get(depositId) ?? null,
+    // A per-deposit VP error is more specific, so it wins; the params failure
+    // is the fallback so an outage is never silently swallowed.
+    error: errors?.get(depositId) ?? protocolParamsError,
```
jrwbabylonlab, line 305: `useSigningRequiredNotifications` filters `loading` but not `error` — notifies while the CTA is suppressed.

services/vault/src/components/simple/PostDepositContinuationView.tsx
```diff
@@ -112,6 +112,17 @@ export function PostDepositContinuationView({
...
 
+  // Refresh the VP poll on open. This used to happen for free: the modal
+  // mounted its own provider, whose query key was scoped to the viewed batch,
+  // so opening it always produced a cache miss and a fresh fetch. Sharing the
+  // app-wide provider means the key no longer changes — and the poll may
+  // already have halted (`refetchInterval` stops once every deposit reports
+  // PendingDepositorSignatures), so without this the user can open the modal
+  // onto a stale snapshot and never see the action they came for.
+  useEffect(() => {
+    refetch();
```
gbarkhatov: **[High] This is a self-sustaining refetch loop, not an occasional extra fire.**

`refetch` here is `contextValue.refetch`, re-created as `() => refetch()` whenever the provider's `contextValue` memo recomputes — which happens whenever `getPollingResult` changes identity. Two of its dependencies make that *every render*: `useBtcHtlcRefundStatus.ts:104` and `useBtcMempoolConfirmations.ts:81` both return `query.data ?? new Map()`, allocating a fresh `Map` on each render whenever the query is disabled or unloaded. The HTLC one is disabled in any session without expired vaults, so that is the common case, not the edge case.

Even with both loaded, `replaceEqualDeep` (query-core `utils.js:113`) bails to `b` for anything that is not a plain object or array, so a `Map` can never be structurally shared — every resolved poll mints new identities anyway.

Net effect while the continuation modal is mounted: `refetch()` → new data → new context value → effect → `refetch()`, back to back at network latency, and it defeats the `refetchInterval: false` halt the polling design relies on. The new test cannot see it — it mocks `usePeginPolling` with a stable `mockRefetch` and asserts only `toHaveBeenCalled()`.

One-shot it:

```ts
const refetchRef = useRef(refetch);
refetchRef.current = refetch;
useEffect(() => {
  refetchRef.current();
}, []);
```

and assert `toHaveBeenCalledTimes(1)` across a re-render so the test is actually load-bearing.

services/vault/eslint.config.js
```diff
@@ -231,6 +256,17 @@ export default tseslint.config(
       ],
     },
   },
+  // Single-polling-provider boundary: the sanctioned mount and the tests may
+  // import `PeginPollingProvider`. Placed last so it wins over both blocks
+  // above, and re-declares the usehooks ban for the same wholesale-replacement
+  // reason. The src/dev pattern ban is not re-declared because every file here
+  // is already exempt from it.
```
gbarkhatov: **[Medium] This claim doesn't hold for the first entry, and the block silently drops that file's `src/dev` ban.**

`PEGIN_POLLING_PROVIDER_MOUNT_FILES` is `["src/context/deposit/AppPeginPollingProvider.tsx", "**/__tests__/**", "**/*.test.{ts,tsx}"]`. The two test globs are exempt via the dev block's own `ignores`, but `AppPeginPollingProvider.tsx` is not in that list — so the dev ban does cover it, and this block, being last, replaces its `no-restricted-imports` config wholesale with a `paths`-only object. The `patterns` entry disappears, and the sanctioned single mount becomes the one production file that may import `@/dev/*` unchecked.

Nothing is broken today (the file imports no dev module), but it is the guardrail layer this PR exists to add, weakened at exactly the file it is protecting — and by the same wholesale-replacement mechanic the comment two blocks up correctly warns about.

Either re-declare the `patterns` entry here, or narrow `files` to the two test globs and add `AppPeginPollingProvider.tsx` to the dev block's `ignores` instead, so both bans stay declared in one place.

services/vault/eslint.config.js
```diff
@@ -7,6 +7,40 @@ import reactRefresh from "eslint-plugin-react-refresh";
...
+// the only sanctioned mount; everything else consumes the hooks.
+//
+// `no-restricted-imports` matches the literal specifier, not the resolved path,
+// so every spelling in use has to be listed — same caveat as the src/dev ban.
+const PEGIN_POLLING_PROVIDER_BANS = [
+  "@/context/deposit/PeginPollingContext",
+  "./PeginPollingContext",
+  "../PeginPollingContext",
+  "../../context/deposit/PeginPollingContext",
+  "../../../context/deposit/PeginPollingContext",
```
gbarkhatov: [Low] Same literal-specifier caveat the comment above names, but the list stops at three levels: a file under `src/components/simple/DepositProgressView/` importing `"../../../../context/deposit/PeginPollingContext"` is unbanned, and the array has to grow every time the tree deepens.

The `src/dev` block two overrides down already solves this with `patterns`, and `patterns` entries accept `importNames` too:

```js
patterns: [
  {
    group: ["@/context/deposit/PeginPollingContext", "**/PeginPollingContext"],
    importNames: ["PeginPollingProvider"],
    message: "…",
  },
]
```

That covers every spelling at once instead of keeping the alias and relative forms in sync by hand.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Per-deposit `tRefund`. Strict by design: never falls back to the latest
+   * version, because a since-lowered `tRefund` would mark a vault mature early
+   * and Bitcoin would reject the refund with `non-BIP68-final`.
+   */
+  resolveRefundTimelock: (offchainParamsVersion?: number) => number | undefined;
+}
+
+export function usePeginPollingProtocolParams(): PeginPollingProtocolParams {
+  const { data: config, error: configError } = useQuery(
```
gbarkhatov: [Low] Neither query carries an `enabled` gate, and this hook now mounts in `RootLayout` for every session on every route. Both contract reads therefore fire on page load for a disconnected visitor, and on routes that never mount `ProtocolParamsProvider`.

"Shared cache, no extra request" holds on routes that already mount the blocking provider; elsewhere this is two new multicalls per session that nothing on the page consumes. Probably acceptable given `staleTime: 5m` and `refetchOnWindowFocus: false`, but worth being deliberate about rather than inheriting it from the hoist — the provider already knows `activities` and `btcPublicKey`, so an `enabled` following the same condition `usePeginPollingQuery` uses would keep the pre-PR footprint.

services/vault/src/context/deposit/PeginPollingContext.tsx
```diff
@@ -122,6 +121,54 @@ const PeginPollingContext = createContext<PeginPollingContextValue | null>(
...
+  useEffect(() => {
+    mountedProviderCount += 1;
+    if (mountedProviderCount > MAX_CONCURRENT_PROVIDERS) {
+      const message =
+        `${mountedProviderCount} PeginPollingProvider instances are mounted at once. ` +
+        "The app must mount exactly one, via AppPeginPollingProvider — a second " +
+        "instance forks polling and optimistic-completion state, so a completed " +
+        "action stops hiding its own button.";
+      if (import.meta.env.DEV) {
+        throw new Error(message);
```
gbarkhatov, line 162: [Low] The counter is already incremented when this throws, and an effect setup that throws never registers its cleanup — so `mountedProviderCount` stays elevated for the rest of the session. After the dev crash, a corrected tree keeps tripping the invariant on every HMR update until a full reload, which is the same leak `resetPeginPollingProviderCount` exists to paper over in tests.

Decrementing before the throw (or `try`/`finally` around the check) makes it recoverable without changing the guarantee.

services/vault/src/context/deposit/computeDepositPollingResult.ts
```diff
@@ -65,8 +65,27 @@ export interface DepositPollingInputs {
...
+   * Per-vault min depth, pre-resolved from `offchainParamsVersion`.
+   * `undefined` while the protocol params are still loading (or failed) — the
+   * confirmation conclusion is then withheld rather than defaulted, so a
+   * deposit can never read as at-depth on an unknown threshold.
+   */
+  requiredDepth: number | undefined;
+  /**
+   * Protocol-param load failure. Surfaced on every deposit's `error` when no
+   * more specific per-deposit error exists, so a params outage is never
+   * silently swallowed.
```
gbarkhatov: [Nit] "never silently swallowed" is the one bit of the old framing left: the `why:` block directly below explains that the only reader collapses to `noAction`, which renders neither a button nor a message — silent is exactly what it is from the user's side. Something like "so a params outage suppresses actions on every deposit rather than being dropped" would match the paragraph under it.

services/vault/src/components/simple/PendingDepositSection.tsx
```diff
@@ -117,19 +114,15 @@ export function PendingDepositSection() {
...
+  // pending (the section early-returns above), so an empty dashboard doesn't
+  // pay the params load.
   return (
     <ProtocolParamsProvider>
-      <PeginPollingProvider
-        activities={allActivities}
-        pendingPegins={pendingPegins}
-        btcPublicKey={btcPublicKey}
-      >
+      <>
```
gbarkhatov: [Nit] Leftover from the provider this replaced — `ProtocolParamsProvider` takes the children directly, so the fragment can go. Same pair at `VaultsLifecycleSections.tsx:447` / `:522`.

services/vault/src/context/deposit/PeginPollingContext.tsx
```diff
@@ -569,19 +617,20 @@ export function useDepositPollingResult(depositId: string) {
...
+ *
+ * A deposit is unindexed in the moments right after broadcast, before the
+ * indexer has it; callers fall back to a direct mempool read there.
  */
 export function useOptionalDepositPollingResult(
   depositIds: readonly string[],
 ): DepositPollingResult | undefined {
-  const polling = usePeginPollingOptional();
-  if (!polling) return undefined;
+  const { getPollingResult } = usePeginPolling();
```
gbarkhatov, line 642: [Nit] Switching from `usePeginPollingOptional` to the throwing `usePeginPolling` is right now that the provider is app-scoped — I checked every call site and they are all under the mount. But the name still promises a tolerance the function no longer has: "Optional" now refers only to the result, not to the provider. Renaming (`useFirstIndexedDepositPollingResult`, or similar) would stop the next caller reaching for this as the safe-outside-the-provider variant.

gbarkhatov (dismissed):
Re-reviewed at `de78838`. All five of my earlier findings are genuinely fixed in code — I re-derived each at HEAD rather than taking the thread at face value:

- `error: ready ? null : (configError ?? offchainError)` at `usePeginPollingProtocolParams.ts:90`, with a refetch-failure test that fails against the old line.
- `isLoading: isLoading || (!params.ready && !params.error)` at `PeginPollingContext.tsx:531`, with `loading === false` and `loading === true` pinned separately.
- The `error`-as-suppression semantics settled deliberately, and the doc comment now describes what the field does instead of the opposite. I'm fine with the call: depth and `tRefund` gate Broadcast and Refund, and over-broad in that direction is the safe direction.
- `useRequiredPrePeginDepthResolver` lost its `export` (function kept, still called internally), JSDoc corrected, the extraction named as the open cleanup rather than left implicit.
- The `cachedAtDepth` asymmetry has its `why:`, and the boolean side is now covered as well as the count side.

Approving the collapse. Three things below, one of which I would fix before merge.

**The "Known" item in the description is a self-sustaining loop, not an occasional extra fire.** Details inline at `PostDepositContinuationView.tsx:123` — two of `getPollingResult`'s dependencies allocate a fresh `Map` on every render, so the effect re-fires on every render, not just on every poll.

**@jrwbabylonlab's two open threads are both valid, and one is stronger than filed.**

- `usePeginPollingProtocolParams.ts:73` — the fallback is reachable, not theoretical. `fetchAllOffchainParams` *skips* versions that fail validation (`useOffchainParams.ts:37-42` logs `failed validation, skipping`), so `byVersion.get(pinnedVersion)` legitimately misses and the deposit is then measured against the latest depth. If latest is shallower, `getTxidsCrossingThreshold` → `addConfirmedPrePeginTxid` persists that conclusion to localStorage, after which `cachedAtDepth` short-circuits it permanently. Splitting the two cases keeps the pre-sign fallback and withholds on a miss — which is what `resolveRefundTimelock`'s "strict by design" doc three lines below already promises for the sibling value.
- `computeDepositPollingResult.ts:283` — the reply argues the `error` gate handles this one layer down, but it can't: `useSigningRequiredNotifications` builds its list from `peginState.availableActions`, and `error` never touches `peginState`. `useSigningRequiredNotifications.ts:59` filters `loading` and never reads `error`, so the `isLoading` fix in `de78838` is precisely what opened this.

Also verified and fine: the VP poll is `enabled` only when deposits need polling, so the app-wide mount costs nothing for a session with none; `useVaultDeposits(ethAddress).activities` is identical to the `allActivities` the old mounts passed, so no deposit set changed; the geo-block gate correctly sits above the provider; and no dangling references to the removed `usePeginPollingOptional` / resolver export.

One item outside the diff, from your own reply: `ProtocolParamsContext.tsx:141` gates on `allError || !configData` and never on `!offchainParamsData`. Worth a follow-up — it is the one seam where the superset argument behind the CTA blackout doesn't strictly hold.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
jonybur: Fixed - the fallback now applies only pre-sign, when the deposit has no registered version. A registered version missing from `byVersion` withholds (`undefined`) instead, the same discipline `resolveRefundTimelock` already had - which matters here exactly because the at-depth conclusion persists (`confirmedTxids`), so a wrong-threshold confirm would outlive the params catching up.

The blocking twin keeps its latest-version fallback deliberately: its consumers only display the depth and persist nothing from it. Both docs now record that divergence instead of claiming "same rule". Test added; against the old `pinned ?? latest` it fails by resolving the latest depth where it must withhold.

One consequence checked deliberately: a version can be missing permanently, not just transiently - `fetchAllOffchainParams` warn-and-skips a version that fails validation. Such a deposit now presents with the conclusion withheld (no countdown; the broadcast CTA stays suppressed by the depth-independent chain signals; advancement stays VP-driven, so it does not hard-stall) rather than confirming at the latest depth. That is the safe side of the trade, since the persisted cache would otherwise outlive the wrong threshold - and if a permanently-invalid version should get loud, the params outage surface is the place, not this resolver.

services/vault/src/context/deposit/computeDepositPollingResult.ts
jonybur: Fixed - the scan now skips a result with `error` set, exactly parallel to `loading`: `getActionStatus` collapses an errored result to `noAction`, so the notification would point at a CTA the continuation view refuses to render. A params outage sets `error` on every deposit, so that was the whole dashboard nudging into dead ends. Test added (suppressed when the polling result carries an error); it fails without the filter.

services/vault/src/components/simple/PostDepositContinuationView.tsx
jonybur: Confirmed and fixed at both ends. The one-shot is your shape - `refetchRef` plus a mount-only effect - and the test now hands the view a fresh `refetch` identity on every render (as the provider does whenever its context value recomputes) and asserts `toHaveBeenCalledTimes(1)` across a re-render. Against the old `[refetch]` effect it fails with a second call - verified by reverting, not assumed.

The two identity mints you traced are also fixed at the source: `useBtcHtlcRefundStatus` and `useBtcMempoolConfirmations` now return module-level `EMPTY_REFUNDS` / `EMPTY_CONFIRMATIONS` singletons instead of a per-render `new Map()`, so the disabled/unloaded case - the common one, as you said - stops minting fresh identities into `getPollingResult`'s deps at all. The loop needed both halves: the ref kills this effect's cycle, and the singletons stop the context value churning for every other consumer of those maps.

services/vault/eslint.config.js
jonybur: Fixed with a split rather than either single option: the trailing override is now two blocks, and both bans became single consts so neither can silently drop out of a scope again.

- `AppPeginPollingProvider.tsx` keeps the dev ban and the usehooks ban and is exempt only from the provider ban - the guardrail stays intact at exactly the file it protects.
- The test globs keep their existing exemptions unchanged.
- `DEV_IMPORT_PATTERN_BAN` is declared once and referenced from the dev block and the mount block, so the wholesale-replacement mechanic you flagged now has one source of truth to re-declare instead of a copy to forget.

services/vault/eslint.config.js
jonybur: Adopted verbatim - `PEGIN_POLLING_PROVIDER_PATTERN_BAN` with your group pair plus `importNames`, replacing the five-entry list. The comment above it now explains why patterns rather than paths, so the next new specifier spelling doesn't quietly grow a hand-maintained list.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
jonybur: Made deliberate - the hook now takes a required `enabled` and the provider passes `activities.length > 0`, so this hook fires nothing for a session with no deposits to conclude about. While disabled the resolvers sit in their withheld `undefined` state, which is already the correct answer with nothing to evaluate; the moment activities appear, React Query fires both fetches on the enabled flip. Starvation was checked at every consumer: pendingPegins-only entries merge into `activities` upstream, demo deposits bypass the params path, and the deadline gate and notifications iterate `activities`, so an empty list needs nothing. Test pins it: no contract read while disabled - fails when force-enabled.

One honest caveat on the footprint claim: in the current tree, `RootLayout` also mounts `SimpleDeposit`, whose blocking `ProtocolParamsProvider` fires the same query keys for every non-geo-blocked visitor regardless. That predates this PR and is unchanged by it - so the gate makes THIS hook cost nothing on its own, but the session-level "two multicalls on page load" only disappears once that mount is gated too. Flagging rather than fixing here; happy to take it as a follow-up.

services/vault/src/context/deposit/PeginPollingContext.tsx
jonybur: Fixed with the decrement before the throw. Only the offending mount's increment is undone - the surviving provider keeps its count, StrictMode's setup-cleanup-setup still nets to one, and a corrected tree stops re-tripping the invariant on HMR updates. The production path is unchanged: it logs, and its cleanup registers normally since nothing throws there.

services/vault/src/context/deposit/computeDepositPollingResult.ts
jonybur: Took your wording - the doc now says the outage suppresses actions on every deposit rather than being dropped, which is what the paragraph below it was already describing.

services/vault/src/components/simple/PendingDepositSection.tsx
jonybur: Both gone - this one and the `VaultsLifecycleSections` pair - with prettier re-indenting the bodies.

services/vault/src/context/deposit/PeginPollingContext.tsx
jonybur: Renamed to `useFirstIndexedDepositPollingResult` across the context, `BtcConfirmationDetailContainer`, and its test - the name now describes what it selects instead of promising a tolerance the function no longer has. No other callers existed to migrate, which your call-site sweep already established.

gbarkhatov (dismissed):

services/vault/src/components/Activity/ActivityListWithRefund.tsx
```diff
@@ -28,15 +31,8 @@ export function ActivityListWithRefund({
...
-    expiredActivities,
-    allActivities,
-    pendingPegins,
-    btcPublicKey,
-    ethAddress,
-    broadcastModal,
-    refundModal,
-  } = usePendingDeposits();
+  const { expiredActivities, ethAddress, broadcastModal, refundModal } =
+    usePendingDeposits();
```
jrwbabylonlab, line 35: Dropping the `pendingPegins` / `btcPublicKey` reads here (and the matching ones in `PendingDepositSection` and `VaultsLifecycleSections`) leaves both returns on `usePendingDeposits` with zero production consumers. Worth removing them from the hook in the same pass.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
```diff
@@ -0,0 +1,118 @@
...
+    ...offchainParamsQueryOptions(),
+    enabled,
+  });
+
+  const resolveRequiredPrePeginDepth = useCallback(
+    (offchainParamsVersion?: number): number | undefined => {
+      if (!config || !offchainParams) return undefined;
+      // Pre-sign: the deposit has no registered version yet → latest params.
+      if (offchainParamsVersion === undefined) {
+        return config.offchainParams.minPrepeginDepth;
```
jrwbabylonlab, line 91: The "pre-sign, no registered version yet" premise does not hold: `useDepositFlow` persists the pending row after ETH registration, so the vault already has a pinned version, and the record even carries it via `buildOffchainParamsVersion` — `usePeginStorage` just never maps it onto the activity. Those rows reach here with `undefined`, get the latest depth, and the at-depth conclusion is persisted, which contradicts the comment justifying the `cachedAtDepth ||` short-circuit downstream. Low impact given the 1h cache TTL, but the fix is plumbing the existing value through rather than a new branch.

gbarkhatov (dismissed):

services/vault/src/components/Activity/ActivityListWithRefund.tsx
jonybur: Done in `227b075f` — removed from the hook in the same pass, as suggested. `pendingPegins` and `btcPublicKey` are off the return, the `useBtcPublicKey` import is gone, and the `useVaultDeposits` destructure narrowed to what is still consumed. Test expectations for both dropped returns removed from `usePendingDeposits.test.tsx`, `router.test.tsx` and `Activity.test.tsx`.

services/vault/src/hooks/deposit/usePeginPollingProtocolParams.ts
jonybur: Fixed in `227b075f`, taking the plumbing route rather than a new branch.

`usePeginStorage` now maps the existing value across:

```ts
offchainParamsVersion: pending.buildOffchainParamsVersion,
```

You were right that the premise was simply wrong — the value was persisted at registration, re-asserted on chain, and just never mapped onto the activity. So registered rows no longer arrive as `undefined`, no longer pick up the latest depth, and no longer persist an at-depth conclusion drawn against the wrong version. That removes the contradiction with the `cachedAtDepth ||` short-circuit rather than papering over it.

The comment that stated the bad premise is corrected too: the `undefined` branch is now documented as "not registered on Ethereum yet", which is the only case that actually reaches it. 68 lines of coverage added in `usePeginStorage.test.tsx`.

Agreed on low impact given the 1h TTL — but the persisted-conclusion part is what made it worth doing properly.

gbarkhatov (approved):

PR #2145, opened 2026-07-27
fix(vault): bound the WOTS submission suppression with a TTL

jonybur:
## The bug

#2140 added a `wotsSubmitted` marker so a row stops re-offering "Submit WOTS Key" while the daemon catches up. The set is **add-only**.

So if a vault provider ever asks for a key *again* — rejected key, rotated key, or a write lost behind a 200 — the button stays hidden for the life of the tab. The row shows "Waiting for vault provider…" with no CTA, and the continuation modal loses its resume branch too (same `availableActions`, only mount site). Reload is the only way out.

## The fix

Record *when* each submission resolved, not just *that* it did, and expire the suppression after **20 minutes**.

`wotsSubmitted: ReadonlySet<string>` → `wotsSubmittedAt: ReadonlyMap<string, number>`, read through `isWotsSubmissionWithinTtl`.

Sized against the continuation modal rather than the dashboard row — it's the harsher consumer, since a premature lapse bounces a parked user back to the WOTS screen with nothing wrong. Reasoning is on the constant.

## Also here

- The suppression **re-arms**: a submission landing after a lapse records a fresh timestamp instead of keeping the expired one.
- A re-offer **waits for a click** instead of auto-firing a wallet prompt at an idle modal.
- `setOptimisticDepositStatus` is idempotent, matching `markWotsSubmitted`.

## Testing

2736 pass, 0 lint errors, typecheck clean. The TTL, re-arm, and no-auto-fire tests were each confirmed to fail without their fix.

Lands cleanly with #2141 — merge verified both directions, no conflicts, order doesn't matter.

### DIFF
--- a/services/vault/src/components/simple/ResumeDepositContent.tsx
+++ b/services/vault/src/components/simple/ResumeDepositContent.tsx
@@ -32,7 +32,10 @@ import { getVaultRegistryReader } from "@/clients/eth-contract/sdk-readers";
 import { computeDepositDerivedState } from "@/components/deposit/DepositSignModal/depositStepHelpers";
 import { usePayoutSigningState } from "@/components/deposit/PayoutSignModal/usePayoutSigningState";
 import { useDepositPollingResult } from "@/context/deposit/PeginPollingContext";
-import { markWotsSubmitted } from "@/context/deposit/optimisticDepositState";
+import {
+  hasWotsSubmissionRecord,
+  markWotsSubmitted,
+} from "@/context/deposit/optimisticDepositState";
 import { COPY } from "@/copy";
 import {
   DepositFlowStep,
@@ -271,10 +274,24 @@ export function ResumeWotsContent({
     null;
   const connectedBtcAddress = btcConnector?.connectedWallet?.account?.address;
 
-  // Starts true: useRunOnce auto-fires handleSubmit on mount, so the
-  // first render must show processing — not a false-success banner from
-  // `isComplete = !loading && !error`.
-  const [loading, setLoading] = useState(true);
+  // A submission already recorded for this deposit means the user has been
+  // through this step in this session, so this mount is the suppression TTL
+  // lapsing and re-offering the action — not a first visit. Auto-submitting
+  // there would fire a wallet prompt at an idle open modal with no user
+  // gesture behind it, so a re-offer waits for an explicit click instead.
+  // Read once at mount: `markWotsSubmitted` below flips it, and re-reading
+  // would swap the component into the wrong mode mid-flight.
+  const [isReoffer] = useState(() => hasWotsSubmissionRecord(activity.id));
+
+  // `started` false parks DepositProgressView on its pre-sign entry state,
+  // where the CTA calls `onSign` — the same seam DepositSignContent uses.
+  const [started, setStarted] = useState(!isReoffer);
+
+  // Starts true on the auto-submit path: useRunOnce fires handleSubmit on
+  // mount, so the first render must show processing — not a false-success
+  // banner from `isComplete = !loading && !error`. A re-offer has not
+  // submitted anything yet, so it starts idle.
+  const [loading, setLoading] = useState(!isReoffer);
   const [error, setError] = useState<string | null>(null);
 
   // Track mount for setState guards after the long async chain below.
@@ -470,7 +487,18 @@ export function ResumeWotsContent({
   // wallet before its account hydrates: in that case useRunOnce (one-shot)
   // would defer rather than fire into the "not connected" guard. When there is
   // genuinely no provider it fires, so the real "not connected" error surfaces.
-  useRunOnce(handleSubmit, !btcWalletProvider || Boolean(connectedBtcAddress));
+  //
+  // `!isReoffer` keeps the auto-run to a first visit; a re-offer submits only
+  // through `handleStart`, behind a click.
+  useRunOnce(
+    handleSubmit,
+    !isReoffer && (!btcWalletProvider || Boolean(connectedBtcAddress)),
+  );
+
+  const handleStart = useCallback(() => {
+    setStarted(true);
+    void handleSubmit();
+  }, [handleSubmit]);
 
   // Reconcile the displayed step with the polled VP status instead of trusting
   // local `loading`/`error` alone. Without this the modal computes its step
@@ -496,7 +524,13 @@ export function ResumeWotsContent({
   // status confirms acceptance. Then the modal sits on the next step as a
   // closeable background wait ("Close & continue later"), matching the other
   // resume waits — no separate success banner needed.
-  const advanced = (!loading && !error) || pastWots;
+  //
+  // `started` gates the local half: a re-offer sits idle (not loading, no
+  // error) until the user clicks, and without this that idle state would read
+  // as "submit resolved" and skip the step entirely. `pastWots` is unguarded
+  // — the VP confirming acceptance advances regardless of what this instance
+  // did.
+  const advanced = pastWots || (started && !loading && !error);
   const renderStep = advanced
     ? DepositFlowStep.AWAIT_PAYOUT_TRANSACTIONS
     : DepositFlowStep.SUBMIT_WOTS_KEYS;
@@ -540,6 +574,8 @@ export function ResumeWotsContent({
       perVaultSteps={perVaultSteps}
       onClose={onClose}
       onRetry={error ? handleSubmit : undefined}
+      started={started}
+      onSign={handleStart}
       btcConfirmationDetail={btcConfirmationDetail}
       wotsApprovalHint={COPY.deposit.resume.wotsWalletApprovalHint}
       offchainParamsVersion={activity.offchainParamsVersion}
--- a/services/vault/src/components/simple/ResumeDepositContent.tsx
+++ b/services/vault/src/components/simple/ResumeDepositContent.tsx
@@ -281,6 +281,15 @@ export function ResumeWotsContent({
   // gesture behind it, so a re-offer waits for an explicit click instead.
   // Read once at mount: `markWotsSubmitted` below flips it, and re-reading
   // would swap the component into the wrong mode mid-flight.
+  //
+  // Deliberately gated even when the user just clicked the re-offered row
+  // action — where that click was already a gesture and this costs a second
+  // one. A re-offer means the VP is still asking after a submission this
+  // session watched resolve, so something may genuinely be wrong; making the
+  // user confirm the fresh wallet popup on that abnormal path is worth more
+  // than the click it saves, and it spares the mount site from having to
+  // report whether this render is a fresh open or a branch swap under an
+  // already-open modal.
   const [isReoffer] = useState(() => hasWotsSubmissionRecord(activity.id));
 
   // `started` false parks DepositProgressView on its pre-sign entry state,
--- a/services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
+++ b/services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
@@ -380,7 +380,7 @@ describe("ResumeWotsContent — submission marker", () => {
 
     await waitFor(() => {
       expect(
-        getOptimisticDepositState().wotsSubmitted.has(baseActivity.id),
+        getOptimisticDepositState().wotsSubmittedAt.has(baseActivity.id),
       ).toBe(true);
     });
   });
@@ -399,9 +399,9 @@ describe("ResumeWotsContent — submission marker", () => {
     await waitFor(() => {
       expect(getByTestId("error").textContent).toContain("VP rejected the key");
     });
-    expect(getOptimisticDepositState().wotsSubmitted.has(baseActivity.id)).toBe(
-      false,
-    );
+    expect(
+      getOptimisticDepositState().wotsSubmittedAt.has(baseActivity.id),
+    ).toBe(false);
   });
 });
 
--- a/services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
+++ b/services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
@@ -6,13 +6,14 @@
  * indexer can ask the wallet to derive over attacker-chosen funding outpoints.
  */
 
-import { render, waitFor } from "@testing-library/react";
+import { fireEvent, render, waitFor } from "@testing-library/react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import { getVaultRegistryReader } from "@/clients/eth-contract/sdk-readers";
 import { usePayoutSigningState } from "@/components/deposit/PayoutSignModal/usePayoutSigningState";
 import {
   getOptimisticDepositState,
+  markWotsSubmitted,
   resetOptimisticDepositState,
 } from "@/context/deposit/optimisticDepositState";
 import { COPY } from "@/copy";
@@ -208,6 +209,8 @@ vi.mock("../DepositProgressView", () => ({
     canContinueInBackground,
     onRetry,
     wotsApprovalHint,
+    started,
+    onSign,
   }: {
     currentStep?: string;
     error?: { title: string; body: string } | null;
@@ -217,8 +220,16 @@ vi.mock("../DepositProgressView", () => ({
     canContinueInBackground?: boolean;
     onRetry?: () => void;
     wotsApprovalHint?: string | null;
+    started?: boolean;
+    onSign?: () => void;
   }) => (
     <div data-testid="progress-view">
+      {/* Mirrors the real prop default so views that never pass it read as
+          started, the same as they render today. */}
+      <span data-testid="started">{String(started !== false)}</span>
+      <button type="button" data-testid="sign" onClick={onSign}>
+        sign
+      </button>
       <span data-testid="wots-hint">{wotsApprovalHint ?? ""}</span>
       <span data-testid="step">{String(currentStep)}</span>
       <span data-testid="error">{error?.body ?? ""}</span>
@@ -266,6 +277,15 @@ function readerWith(prePeginTxHash: string) {
     getVaultProtocolInfo: vi.fn(),
   } as unknown as ReturnType<typeof getVaultRegistryReader>;
 }
+
+// The optimistic store is module-scoped, so it outlives every render here. A
+// leaked WOTS marker is not inert: `ResumeWotsContent` reads it at mount to
+// decide whether it may auto-submit, so one block's marker would silently
+// turn off the auto-submit every later block depends on. Reset for the whole
+// file rather than per-describe.
+beforeEach(() => {
+  resetOptimisticDepositState();
+});
 
 describe("ResumeWotsContent — Pre-PegIn tx hash trust boundary", () => {
   beforeEach(() => {
@@ -402,6 +422,49 @@ describe("ResumeWotsContent — submission marker", () => {
     expect(
       getOptimisticDepositState().wotsSubmittedAt.has(baseActivity.id),
     ).toBe(false);
+  });
+
+  it("submits automatically on a first visit", async () => {
+    render(
+      <ResumeWotsContent
+        activity={baseActivity}
+        onClose={vi.fn()}
+        onSuccess={vi.fn()}
+      />,
+    );
+
+    await waitFor(() => {
+      expect(mockSubmitWotsPublicKey).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  it("waits for a click instead of auto-submitting when the suppression lapsed", async () => {
+    // The TTL expiring re-offers SUBMIT_WOTS_KEY, which remounts this
+    // component. Auto-firing there would open a wallet prompt at a modal the
+    // user left sitting open, with no gesture behind it.
+    markWotsSubmitted(baseActivity.id);
+
+    const { getByTestId } = render(
+      <ResumeWotsContent
+        activity={baseActivity}
+        onClose={vi.fn()}
+        onSuccess={vi.fn()}
+      />,
+    );
+
+    // `getVaultRegistryReader` is the first thing handleSubmit touches and it
+    // runs synchronously, so it is a reliable "the submit path started" probe.
+    // Asserting on `submitWotsPublicKey` here would not be: it sits behind
+    // several awaits and reads as un-called whether or not the guard holds.
+    expect(mockGetVaultRegistryReader).not.toHaveBeenCalled();
+    expect(getByTestId("started").textContent).toBe("false");
+
+    fireEvent.click(getByTestId("sign"));
+
+    await waitFor(() => {
+      expect(mockSubmitWotsPublicKey).toHaveBeenCalledTimes(1);
+    });
+    expect(getByTestId("started").textContent).toBe("true");
   });
 });
 
--- a/services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
+++ b/services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
@@ -386,7 +386,6 @@ describe("ResumeWotsContent — submission marker", () => {
     mockCalculateBtcTxHash.mockReturnValue(ON_CHAIN_HASH);
     mockGetVaultRegistryReader.mockReturnValue(readerWith(ON_CHAIN_HASH));
     mockDeriveVaultRoot.mockResolvedValue(new Uint8Array(32));
-    resetOptimisticDepositState();
   });
 
   it("records the WOTS submission so the dashboard row stops offering the button", async () => {
@@ -442,7 +441,16 @@ describe("ResumeWotsContent — submission marker", () => {
     // The TTL expiring re-offers SUBMIT_WOTS_KEY, which remounts this
     // component. Auto-firing there would open a wallet prompt at a modal the
     // user left sitting open, with no gesture behind it.
+    //
+    // Record the marker 21 minutes in the past (fake timers only for the
+    // write, real timers restored for the async render below) so the fixture
+    // is the production scenario the title names: a marker that is present
+    // but past the 20-minute TTL — not merely present.
+    const lapsedStamp = Date.now() - 21 * 60 * 1000;
+    vi.useFakeTimers();
+    vi.setSystemTime(lapsedStamp);
     markWotsSubmitted(baseActivity.id);
+    vi.useRealTimers();
 
     const { getByTestId } = render(
       <ResumeWotsContent
--- a/services/vault/src/context/deposit/__tests__/PeginPollingContext.test.tsx
+++ b/services/vault/src/context/deposit/__tests__/PeginPollingContext.test.tsx
@@ -336,11 +336,17 @@ describe("PeginPollingContext", () => {
     ).toContain(PeginAction.SUBMIT_WOTS_KEY);
   });
 
-  it("re-offers Submit WOTS Key once the suppression window has elapsed and the vault provider is still asking", () => {
-    // The marker only bridges daemon lag. A VP still asking ten minutes on is
-    // asking for real — a rejected or rotated key, or a submission lost behind
-    // a 200 — and an unbounded marker would leave the row with no action at
-    // all until the user thought to reload.
+  it("recomputes Submit WOTS Key as available once the suppression window has elapsed and the vault provider is still asking", () => {
+    // The marker only bridges daemon lag. A VP still asking twenty minutes on
+    // is asking for real — a rejected or rotated key, or a submission lost
+    // behind a 200 — and an unbounded marker would leave the row with no
+    // action at all until the user thought to reload.
+    //
+    // Scope: this calls `getPollingResult` directly after advancing the clock,
+    // so it pins the COMPUTATION flipping, not a re-render. No dep of that
+    // `useCallback` changes when the clock crosses the boundary, so this would
+    // pass even if nothing re-rendered. What carries it in production is
+    // `refetchInterval` — see `WOTS_SUBMISSION_SUPPRESSION_MS`.
     vi.useFakeTimers();
     vi.setSystemTime(new Date("2026-07-27T12:00:00Z"));
     mockQueryResult.needsWotsKey = new Set([ACTIVITY_ID]);
@@ -355,7 +361,7 @@ describe("PeginPollingContext", () => {
       result.current.getPollingResult(ACTIVITY_ID)?.peginState.availableActions,
     ).toEqual([PeginAction.NONE]);
 
-    vi.advanceTimersByTime(11 * 60 * 1000);
+    vi.advanceTimersByTime(21 * 60 * 1000);
 
     expect(
       result.current.getPollingResult(ACTIVITY_ID)?.peginState.availableActions,
--- a/services/vault/src/context/deposit/__tests__/PeginPollingContext.test.tsx
+++ b/services/vault/src/context/deposit/__tests__/PeginPollingContext.test.tsx
@@ -1,7 +1,7 @@
 import { act, render, renderHook, waitFor } from "@testing-library/react";
 import type { PropsWithChildren } from "react";
 import type { Hex } from "viem";
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { COPY } from "../../../copy";
 import {
@@ -142,6 +142,12 @@ describe("PeginPollingContext", () => {
     // Optimistic completions are app-scoped, so they outlive any single
     // provider — and therefore any single test.
     resetOptimisticDepositState();
+  });
+
+  // Restored here, not at the end of each timer test: a failing assertion would
+  // otherwise leak fake timers into every later `waitFor` and cascade timeouts.
+  afterEach(() => {
+    vi.useRealTimers();
   });
 
   it("trusts an in-memory PAYOUT_SIGNED over a stale-cached transactionsReady so the Sign button hides immediately after signing", () => {
@@ -327,6 +333,32 @@ describe("PeginPollingContext", () => {
 
     expect(
       result.current.getPollingResult(OTHER_ID)?.peginState.availableActions,
+    ).toContain(PeginAction.SUBMIT_WOTS_KEY);
+  });
+
+  it("re-offers Submit WOTS Key once the suppression window has elapsed and the vault provider is still asking", () => {
+    // The marker only bridges daemon lag. A VP still asking ten minutes on is
+    // asking for real — a rejected or rotated key, or a submission lost behind
+    // a 200 — and an unbounded marker would leave the row with no action at
+    // all until the user thought to reload.
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date("2026-07-27T12:00:00Z"));
+    mockQueryResult.needsWotsKey = new Set([ACTIVITY_ID]);
+    mockQueryResult.pendingIngestion = new Set();
+
+    const { result } = renderProvider();
+
+    act(() => {
+      markWotsSubmitted(ACTIVITY_ID);
+    });
+    expect(
+      result.current.getPollingResult(ACTIVITY_ID)?.peginState.availableActions,
+    ).toEqual([PeginAction.NONE]);
+
+    vi.advanceTimersByTime(11 * 60 * 1000);
+
+    expect(
+      result.current.getPollingResult(ACTIVITY_ID)?.peginState.availableActions,
     ).toContain(PeginAction.SUBMIT_WOTS_KEY);
   });
 
--- a/services/vault/src/context/deposit/__tests__/optimisticDepositState.test.ts
+++ b/services/vault/src/context/deposit/__tests__/optimisticDepositState.test.ts
@@ -145,7 +145,7 @@ describe("optimisticDepositState", () => {
     vi.setSystemTime(new Date("2026-07-27T12:00:00Z"));
     markWotsSubmitted(DEPOSIT_ID);
 
-    vi.advanceTimersByTime(11 * 60 * 1000);
+    vi.advanceTimersByTime(21 * 60 * 1000);
 
     expect(
       isWotsSubmissionWithinTtl(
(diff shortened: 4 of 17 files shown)


services/vault/src/context/deposit/optimisticDepositState.ts
```diff
@@ -83,19 +113,44 @@ export function setOptimisticDepositStatus(
...
-    wotsSubmitted: currentState.wotsSubmitted,
+    wotsSubmittedAt: currentState.wotsSubmittedAt,
   });
 }
 
 export function markWotsSubmitted(depositId: string): void {
-  if (currentState.wotsSubmitted.has(depositId)) return;
+  // First write wins: a repeat call must not slide the TTL forward, or a retry
+  // loop could keep the action suppressed indefinitely.
+  if (currentState.wotsSubmittedAt.has(depositId)) return;
```
jrwbabylonlab: `.has()` means the window only ever opens once per deposit per session — after the TTL lapses a successful resubmission gets dropped and the button comes straight back on the next poll. `if (isWotsSubmissionWithinTtl(currentState.wotsSubmittedAt.get(depositId))) return;` keeps the don't-slide-forward property and fixes it.

services/vault/src/context/deposit/optimisticDepositState.ts
```diff
@@ -26,18 +26,36 @@ export interface OptimisticDepositState {
...
+ * rejected, rotated, or a write we lost behind a 200 — and an indefinite marker
+ * would leave the row with no action at all and no route back short of a
+ * reload. So the suppression lapses and the user can retry.
+ *
+ * Same reasoning as `REFUND_BROADCAST_SUPPRESSION_MS` in `peginStateMachine`,
+ * which documents why a suppression marker must never be sticky. Generous
+ * against real daemon lag, short enough that a stuck deposit recovers on its
+ * own within one sitting.
+ */
+const WOTS_SUBMISSION_SUPPRESSION_MS = 10 * 60 * 1000;
```
jrwbabylonlab: Worth thinking about what the lapse re-enables: `ResumeWotsContent` is picked off `availableActions`, so expiry remounts it and `useRunOnce` auto-fires `handleSubmit` — a wallet prompt on an idle open modal with no user gesture.

services/vault/src/context/deposit/optimisticDepositState.ts
gbarkhatov: **Blocking** — the marker can never re-arm, so the recovery path this PR creates is single-use.

This guard tests presence, not freshness, and nothing ever deletes an entry (`resetOptimisticDepositState` is test-only). Walk the PR's own headline scenario:

1. T0 — submit, marker = T0, suppressed. Correct.
2. T0+10m — TTL lapses, VP still asking, the action comes back. The fix working.
3. User clicks Submit WOTS Key, it succeeds → `markWotsSubmitted` early-returns here, timestamp stays T0.
4. `isWotsSubmissionWithinTtl(T0)` is already false, so the second — successful — submission is suppressed for zero seconds. The row re-offers the button through the entire daemon-lag window, and a click re-runs the whole derivation plus a fresh wallet popup for a no-op. That is the #2140 symptom, now permanent for that deposit for the life of the tab.

Both call sites inherit it: `useDepositFlow.ts:1012` and `ResumeDepositContent.tsx:425`.

The guard's stated rationale only argues against sliding the TTL *within* the window, and freshness satisfies that — retries land seconds apart, well inside the bound:

```suggestion
  if (isWotsSubmissionWithinTtl(currentState.wotsSubmittedAt.get(depositId))) return;
```

`isWotsSubmissionWithinTtl` is a function declaration, so the forward reference hoists — no reordering needed. The comment above wants a word about re-arming after expiry.

services/vault/src/context/deposit/optimisticDepositState.ts
```diff
@@ -83,19 +113,44 @@ export function setOptimisticDepositStatus(
...
+ * Deliberately time-based rather than driven by an observed poll transition.
+ * Retiring the marker when a poll stops reporting `needsWotsKey` looks tighter
+ * but is not sound: absence is not an affirmative observation — a deposit whose
+ * VP call errored, or that the provider has not polled, is absent too — and
+ * because the store is app-scoped, the first of the two nested providers to
+ * retire it would un-suppress the other's row while that one still holds a
+ * pre-advance snapshot. A clock is read identically everywhere and cannot
+ * disagree between providers.
+ */
+export function isWotsSubmissionWithinTtl(
```
gbarkhatov, line 169: Non-blocking, consistency. The doc above cites `REFUND_BROADCAST_SUPPRESSION_MS` as the same shape, but its predicate `isRefundBroadcastWithinTtl` (`peginStateMachine.ts:473`) takes an injectable `now`, already threaded through as `PeginStateInputs.now` — which is why those tests need no fake timers and these did, hence the `afterEach` reasoning you had to add.

Matching it would also restore something the module header of `computeDepositPollingResult.ts` still claims: 'Pure per-deposit compute … so the decision tree is testable without a React render'. That compute now reads the wall clock transitively through this helper. Either thread an optional `now` through `DepositPollingInputs`, or update that header so it stops overpromising. Co-locating the two TTL predicates would be a bonus.

services/vault/src/context/deposit/optimisticDepositState.ts
gbarkhatov: On the tunable you flagged: ten minutes may be tight, because the dashboard row isn't the only consumer. `PostDepositContinuationView.tsx:362` renders `ResumeWotsContent` off `actions.includes(SUBMIT_WOTS_KEY)`, so a VP that takes longer than the TTL to leave `PENDING_DEPOSITOR_WOTS_PK` bounces a user parked in the continuation modal back to the WOTS resume screen while nothing is actually wrong.

With the re-arm fix that self-corrects on the next submission; without it, it flaps. 20–30m still comfortably clears the 'recovers within one sitting' bar. No strong opinion on the number beyond this: it should be chosen against the modal, which is the harsher consumer, not against the row.

services/vault/src/context/deposit/__tests__/optimisticDepositState.test.ts
```diff
@@ -75,4 +82,95 @@ describe("optimisticDepositState", () => {
...
+        getOptimisticDepositState().wotsSubmittedAt.get(DEPOSIT_ID),
+      ),
+    ).toBe(false);
+  });
+
+  it("never suppresses a deposit with no recorded submission", () => {
+    expect(isWotsSubmissionWithinTtl(undefined)).toBe(false);
+  });
+
+  it("keeps the original timestamp when the same submission is recorded twice", () => {
```
gbarkhatov, line 162: This is the test gap that let the re-arm bug through: it advances 5 minutes, which is *inside* the window, where first-write-wins is the correct behaviour. The interesting case is a `markWotsSubmitted` after the TTL has lapsed.

Keep this one and add the sibling: mark → advance 11m → mark again → assert the stored timestamp moved and `isWotsSubmissionWithinTtl` reads true again. It fails against the current guard, which is what makes it worth having.

services/vault/src/context/deposit/__tests__/PeginPollingContext.test.tsx
```diff
@@ -330,6 +336,32 @@ describe("PeginPollingContext", () => {
     ).toContain(PeginAction.SUBMIT_WOTS_KEY);
   });
 
+  it("re-offers Submit WOTS Key once the suppression window has elapsed and the vault provider is still asking", () => {
```
gbarkhatov: Worth being precise about what this pins: it calls `getPollingResult` directly after advancing the clock, so it asserts the *computation* flips. It would pass even if nothing re-rendered — no dep of that `useCallback` changes when the clock crosses the boundary.

The production path does hold, for a reason worth one line in the constant's comment: `refetchInterval` keeps polling a deposit stuck at `needsWotsKey` (it only stops at `pendingDepositorSignatures` or a terminal error), and the queryFn hands back fresh `Set`s, which React Query's structural sharing doesn't descend into — so the deps change every tick and the row re-renders. Net effect: recovery is bounded by TTL + one poll interval, not TTL. Nothing to change in the test; just don't let the name imply a re-render assertion it doesn't make.

gbarkhatov (dismissed):
Approving — the diagnosis is right and a TTL is the correct mechanism. The case against 'clear it when a poll observes the advance' holds up against the code: absence really isn't an affirmative observation, and the store is app-scoped, so the first provider to retire the marker would un-suppress the sibling's row.

**One must-fix before merge:** `markWotsSubmitted` guards on presence rather than freshness, so the marker can never be re-armed once it expires. The recovery path this PR adds works exactly once per deposit per tab, and the second — successful — submission gets no suppression at all. One-line fix inline.

Also inline: the test that would have caught it, a consistency nit against `isRefundBroadcastWithinTtl`'s injectable `now`, a note on what the new context test actually pins, and a question about the ten-minute value given `PostDepositContinuationView` gates the resume modal on the same action.

Checked separately and fine:
- The UI does recover on lapse. `refetchInterval` keeps polling a deposit stuck at `needsWotsKey`, and the queryFn returns fresh `Set`s that React Query's structural sharing doesn't descend into, so `getPollingResult`'s deps change every tick.
- `setOptimisticDepositStatus` idempotence is correct in all four argument combinations, and no caller relies on the dropped publish.
- The new `useDepositFlow` assertions are load-bearing and deterministic — the WOTS loop is sequential with `MAX_WOTS_ATTEMPTS = 2`, so the mock ordering can't drift.
- Consumer sweep is clean: no stale set-shaped `wotsSubmitted` reads left anywhere.

services/vault/src/context/deposit/optimisticDepositState.ts
jonybur: Fixed — you walked it exactly right, including that step 4 is the #2140 symptom made permanent. The guard now tests freshness rather than presence:

```js
if (isWotsSubmissionWithinTtl(currentState.wotsSubmittedAt.get(depositId))) return;
```

That keeps the do-not-slide-forward property inside the window and lets a post-lapse submission re-arm.

Confirmed it is a real regression test rather than a passing one: against the old `.has()` guard the new case fails with `expected 1785153600000 not to be 1785153600000` — the timestamp pinned at T0, precisely your step 3.

jonybur: Fixed, and I took your line verbatim — it is the minimal change that keeps the do-not-slide-forward property while letting a post-TTL submission re-arm.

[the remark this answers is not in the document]
jonybur: Fixed. The guard now tests freshness rather than presence, so a submission landing after the window lapsed records a new timestamp instead of retaining the expired one.

services/vault/src/context/deposit/optimisticDepositState.ts
jonybur: Good catch, and it was real — `ResumeWotsContent` says so itself ("Starts true: useRunOnce auto-fires handleSubmit on mount"). Fixed rather than deferred.

The store already knew the difference between a first visit and a re-offer; it just was not being read. Added `hasWotsSubmissionRecord` (presence, deliberately distinct from `isWotsSubmissionWithinTtl`'s freshness) and captured it once at mount:

```js
const [isReoffer] = useState(() => hasWotsSubmissionRecord(activity.id));
```

Read once because `markWotsSubmitted` flips it and re-reading would swap the component into the wrong mode mid-flight. A re-offer parks on `DepositProgressView`'s existing `started={false}` / `onSign` entry state — the same seam `DepositSignContent` uses — so the wallet only opens on a click. No new UI.

One non-obvious consequence: `advanced` had to become `pastWots || (started && !loading && !error)`. Without the `started` guard an idle un-clicked re-offer is indistinguishable from "submit resolved" and skips the step entirely.

services/vault/src/context/deposit/optimisticDepositState.ts
jonybur: Threaded the `now` rather than weakening the header. `isWotsSubmissionWithinTtl` now takes an optional `now` matching `isRefundBroadcastWithinTtl`, fed from a new `DepositPollingInputs.now`, so the compute no longer reads the wall clock transitively and the module header stops overpromising.

services/vault/src/context/deposit/optimisticDepositState.ts
jonybur: Taken — 20 minutes, and I documented the reasoning on the constant so the next person sizing it knows the modal is the binding consumer, not the row. Also recorded the point from your test comment there: recovery is bounded by TTL plus one poll interval, since it is `refetchInterval` and the fresh `Set`s defeating structural sharing that actually carry the re-render, not the clock crossing the boundary.

Separately, the auto-fire jrwbabylonlab raised is now fixed rather than mitigated, so the lapse no longer bounces a parked user into a wallet prompt at all.

services/vault/src/context/deposit/__tests__/optimisticDepositState.test.ts
jonybur: Added the sibling, kept this one. mark → advance past the TTL → mark again → assert the timestamp moved and the predicate reads true again.

You were right that it is the case worth having: against the old guard it fails with `expected 1785153600000 not to be 1785153600000`. I checked that explicitly rather than assuming, because my first attempt at the auto-fire test in `ResumeDepositContent` passed against the broken code — it asserted on a mock sitting behind several awaits, so absence proved nothing. That one now probes `getVaultRegistryReader`, which `handleSubmit` hits synchronously.

services/vault/src/context/deposit/__tests__/PeginPollingContext.test.tsx
jonybur: Renamed to "recomputes Submit WOTS Key as available…" and spelled out the scope in the test: it pins the computation flipping, not a re-render, and would pass even if nothing re-rendered. The `refetchInterval` plus structural-sharing explanation now lives on `WOTS_SUBMISSION_SUPPRESSION_MS`, with the TTL-plus-one-poll-interval bound stated there.

Also bumped the advance to 21 minutes to stay outside the new TTL.

services/vault/src/context/deposit/optimisticDepositState.ts
jonybur: Following up on this one — my earlier reply overstated it.

The `now` parameter landed, but nothing passed it: the provider omitted it and both TTL tests still used `vi.setSystemTime`, so the compute kept falling through to `Date.now()` and the seam was real only in the type. Which means the header stopped overpromising, but the thing you actually asked for — the reason `isRefundBroadcastWithinTtl` needs no fake timers and these did — was not delivered.

Fixed in `fc02e6d`: both sides of the suppression window are now exercised at the compute level through an injected `now`, with no fake timers, matching how `peginStateMachine.test.ts` drives its twin.

Load-bearing rather than decorative — dropping the `now` argument at the call site makes the inside-the-window case fail, since the real clock sits well past the fixture's submission time.

Leaving the provider call site on the `Date.now()` default, same as `PeginStateInputs.now`. Passing `Date.now()` there would be a behavioural no-op that feeds a per-call-changing value into the memo path. Say the word if you would rather have it explicit at both ends.

jonybur:
@greptile-apps

services/vault/src/context/deposit/optimisticDepositState.ts
```diff
@@ -83,19 +127,68 @@ export function setOptimisticDepositStatus(
...
+  submittedAt: number | undefined,
+  now?: number,
+): boolean {
+  if (submittedAt === undefined) return false;
+  const currentTime = now ?? Date.now();
+  return currentTime - submittedAt < WOTS_SUBMISSION_SUPPRESSION_MS;
```
jrwbabylonlab: Backwards clock jump makes this negative, so suppression outlasts the TTL.

services/vault/src/context/deposit/computeDepositPollingResult.ts
```diff
@@ -79,13 +81,21 @@ export interface DepositPollingInputs {
...
-  wotsSubmitted: ReadonlySet<string>;
+  wotsSubmittedAt: ReadonlyMap<string, number>;
   btcPublicKey: string | undefined;
+  /**
+   * Override `Date.now()` for the WOTS suppression TTL (testing only). Mirrors
+   * `PeginStateInputs.now`, and keeps this module's "pure compute" promise
+   * honest — without it the decision tree reads the wall clock transitively
+   * through `isWotsSubmissionWithinTtl` and needs fake timers to pin.
+   */
+  now?: number;
```
gbarkhatov, line 101: The seam is real for the WOTS half but stops there, so this comment's "keeps this module's 'pure compute' promise honest" is still only half true.

`getPeginState` at line 234 has its own `now` (`peginStateMachine.ts:177`), and it drives `isRefundBroadcastWithinTtl` at `:460` and `:664`. This call site does not forward it:

```ts
const peginState = getPeginState(contractStatus, {
  localStatus,
  ...
  refundBroadcastAt,
  // now is missing
});
```

So a caller injecting `now` gets a WOTS decision at the injected time and a refund decision at the real wall clock — from one call, on one deposit. That is a worse failure mode than no seam at all: `computeDepositPollingResult.test.ts` can pin the WOTS suppression with no fake timers, but the moment someone writes a refund-TTL case the same way it will silently read `Date.now()` and they will have to work out why.

One line — add `now,` to the options object at line 234, alongside `refundBroadcastAt`. Same class of half-landed seam as the one you caught yourself in `fc02e6d`, which is why it seems worth closing rather than leaving for the next person to trip over.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -540,6 +574,8 @@ export function ResumeWotsContent({
       perVaultSteps={perVaultSteps}
       onClose={onClose}
       onRetry={error ? handleSubmit : undefined}
+      started={started}
```
gbarkhatov, line 586: The gesture gate is the right fix, but `started` carries more than "has the user clicked" — and the rest of it does not fit a mid-flow step.

`DepositProgressView` pins `visualStep` and `aggregateVisualStep` to `0` when `!started` (`:233`, `:247`). That is correct for `DepositSignContent`, whose pre-sign entry genuinely sits at visual step 1. `SUBMIT_WOTS_KEYS` is visual step 7 (`steps.ts:245`), so on a re-offer:

- `buildStepGroups(0)` marks all four `STEP_GROUPS` `upcoming` — none `completed`, none `expanded`. The user sees zero progress on a deposit whose Pre-PegIn is broadcast and confirmed on chain.
- `completedSteps` is 0, so `showOverallProgress` is false and the progress bar disappears.
- `GroupBlock` early-returns the collapsed header when `!group.expanded` (`:42`), so `activeStepDetail` never renders — meaning `wotsApprovalHint` is suppressed. That hint exists precisely because "resume's cold path fires a wallet approval the WOTS step copy doesn't mention, and some extensions queue it without surfacing a popup" (`copy.ts:612`). It goes missing on the one path where the approval is least expected.
- The CTA falls to the generic `signTransaction` label ("Sign Transaction") with no surrounding context, since nothing is expanded to explain it.

This lands on the continuation modal — the consumer you explicitly sized the 20-minute TTL against. A user parked there watches the deposit appear to reset to zero.

Minimal fix, preserving `DepositSignContent` exactly (its entry step is 1, so `1 - 1 = 0`):

```ts
const visualStep = !started
  ? Math.max(0, getVisualStep(currentStep) - 1)
  : isComplete
    ? TOTAL_VISUAL_STEPS + 1
    : getVisualStep(currentStep);
```

Same shape for `aggregateVisualStep` off `aggregateRawStep`. That leaves the "Register deposit" group completed and expands the WOTS group, which brings the approval hint back with it.

Worth a `DepositProgressView` test pinning it — `ResumeDepositContent.test.tsx` mocks the view, so nothing here can catch it.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -470,7 +487,18 @@ export function ResumeWotsContent({
   // wallet before its account hydrates: in that case useRunOnce (one-shot)
   // would defer rather than fire into the "not connected" guard. When there is
   // genuinely no provider it fires, so the real "not connected" error surfaces.
-  useRunOnce(handleSubmit, !btcWalletProvider || Boolean(connectedBtcAddress));
+  //
+  // `!isReoffer` keeps the auto-run to a first visit; a re-offer submits only
+  // through `handleStart`, behind a click.
+  useRunOnce(
+    handleSubmit,
+    !isReoffer && (!btcWalletProvider || Boolean(connectedBtcAddress)),
```
gbarkhatov, line 504: Not a defect — checking the tradeoff is deliberate, since it is invisible from this file.

`ResumeWotsContent` is only ever mounted by `PostDepositContinuationView:364`, so every route to it goes through that modal. `isReoffer` cannot tell "the modal was sitting open and the branch swapped under it" (jrwbabylonlab's case, where the gate is exactly right) from "the user just clicked Submit WOTS Key on the row and the modal opened" (where the click already was the gesture). The second costs two clicks where a first visit costs one.

A case for leaving it: a re-offer means the VP is asking again, so something may actually be wrong, and making the user confirm a fresh wallet popup rather than firing one at them is defensible on its own terms.

A case for closing it: the mount site knows which it is — `PostDepositContinuationView` could pass whether this render is a fresh open or a branch swap under an already-open modal.

Either is fine; just confirm it is a decision rather than a side effect, and if it stays, the reasoning belongs in the comment above `isReoffer` alongside the idle-modal rationale already there.

services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
```diff
@@ -267,6 +278,15 @@ function readerWith(prePeginTxHash: string) {
   } as unknown as ReturnType<typeof getVaultRegistryReader>;
 }
 
+// The optimistic store is module-scoped, so it outlives every render here. A
+// leaked WOTS marker is not inert: `ResumeWotsContent` reads it at mount to
+// decide whether it may auto-submit, so one block's marker would silently
+// turn off the auto-submit every later block depends on. Reset for the whole
+// file rather than per-describe.
+beforeEach(() => {
```
gbarkhatov, line 286: This is the right call, and the comment argues it well — but the per-describe reset it replaces is still there. `resetOptimisticDepositState()` at line 389 in the `ResumeWotsContent — submission marker` block now runs immediately after this one, so that describe resets twice per test.

Harmless at runtime, but it contradicts this comment's own "Reset for the whole file rather than per-describe", and the next reader has to work out which of the two is authoritative. Drop line 389.

services/vault/src/context/deposit/__tests__/optimisticDepositState.test.ts
```diff
@@ -75,4 +82,114 @@ describe("optimisticDepositState", () => {
...
+    markWotsSubmitted(DEPOSIT_ID);
+    const first = getOptimisticDepositState().wotsSubmittedAt.get(DEPOSIT_ID);
+
+    vi.advanceTimersByTime(21 * 60 * 1000);
+    markWotsSubmitted(DEPOSIT_ID);
+
+    const second = getOptimisticDepositState().wotsSubmittedAt.get(DEPOSIT_ID);
+    expect(second).not.toBe(first);
+    expect(isWotsSubmissionWithinTtl(second)).toBe(true);
+  });
```
gbarkhatov, line 195: Small gap while you are here: `hasWotsSubmissionRecord` is the one exported predicate this file does not test directly — it is covered only transitively, through the click assertion in `ResumeDepositContent.test.tsx`.

The property worth pinning is the one that makes it a separate function at all: presence survives expiry, so the two predicates disagree past the TTL. It sits naturally right after the re-arm test:

```ts
it("keeps the submission record after the suppression window lapsed", () => {
  vi.useFakeTimers();
  vi.setSystemTime(new Date("2026-07-27T12:00:00Z"));
  markWotsSubmitted(DEPOSIT_ID);

  vi.advanceTimersByTime(21 * 60 * 1000);

  // The action is back on offer, but the user has still been through this
  // step — which is what tells a re-offer apart from a first visit.
  expect(hasWotsSubmissionRecord(DEPOSIT_ID)).toBe(true);
  expect(
    isWotsSubmissionWithinTtl(
      getOptimisticDepositState().wotsSubmittedAt.get(DEPOSIT_ID),
    ),
  ).toBe(false);
});
```

gbarkhatov (dismissed):
Approving. Every thread from the last round came back with a real code change rather than a reply, and I checked the two I care most about are load-bearing: the `markWotsSubmitted` guard now tests freshness, and the compute-level clock tests genuinely fail when the `now` argument is dropped at the call site. The self-correction in `fc02e6d` — catching that the first `now` fix landed the parameter but nothing passed it — is the right instinct.

jrwbabylonlab's open comment on the backwards clock jump is valid and is the only unresolved thread; note the sibling `isRefundBroadcastWithinTtl` (`peginStateMachine.ts:479`) has the identical shape, so it is a pre-existing pattern rather than something this PR introduced. Worth deciding whether to fix both or scope to one and say why.

Four things inline, none blocking. The first is the one I would not merge without at least a decision on: `now` is still only half-threaded, so the seam this PR restored leaks for the refund half of the same compute. The second is the only user-visible one: a re-offer renders the stepper as if the deposit had not started.

Checked separately and clean: `setOptimisticDepositStatus` idempotence holds in all four argument combinations and the omitted-`refundBroadcastAt` branch preserves rather than clears; no stale set-shaped `wotsSubmitted` reads remain anywhere in the repo; `started` defaults to `true` in `DepositProgressView`, so no existing caller shifts; leaving the provider on the `Date.now()` default is right, since passing `Date.now()` there would feed a per-call-changing value into the memo path.

One thing I looked at and decided against raising as a defect: `pastWots` flipping true while `started` is false would give `advanced` behind an un-clicked Sign CTA, but it is unreachable rather than merely unlikely. `getPeginDisplayStep` returns `SUBMIT_WOTS_KEYS` exactly when `availableActions` contains `SUBMIT_WOTS_KEY` (`peginStateMachine.ts:816`), and `PostDepositContinuationView` branches off that same `availableActions` from the same `getPollingResult(currentVaultId)` — so the parent unmounts this component in the same pass that would set `pastWots`. A `started={started || advanced}` guard would be free, but it would guard a state the types and the mount site already rule out.

services/vault/src/context/deposit/optimisticDepositState.ts
jonybur: Fixed - the predicate now computes elapsed and requires `elapsedMs >= 0`, so a timestamp the clock has fallen behind reads as expired rather than suppressing until the clock catches back up past it. Expired is the safe direction on all three axes: the re-offer waits for a click (the auto-submit gate reads `hasWotsSubmissionRecord`, which is clock-independent), a redundant submission is a no-op the VP ignores, and because `markWotsSubmitted`'s re-arm guard reads this same predicate, the next submission re-records against the corrected clock - the marker self-heals instead of sticking in the future.

`isRefundBroadcastWithinTtl` had the identical shape, and its own doc already commits to fail-open ("the user can always retry"), so it got the same guard - the refund side is additionally backstopped by chain ground truth (`refundSettlement`), and a duplicate broadcast is rejected by the network.

The guard also reaches the refund predicate's third call site, `shouldRemoveFromLocalStorage`: a future-dated REFUND_BROADCAST marker on an EXPIRED contract now clears as stale rather than surviving until the clock catches back up - consistent with the same "expired is the safe reading" rule, and pinned with its own case.

All three new tests were verified to fail against the bare `< TTL` comparison by reverting the guard: the store test drives the predicate with an injected `now` behind the timestamp, the state-machine tests walk EXPIRED + REFUND_BROADCAST with `refundBroadcastAt` an hour ahead of `now` through both the action-suppression and localStorage-eviction paths.

services/vault/src/context/deposit/computeDepositPollingResult.ts
jonybur: Added - `now,` sits alongside `refundBroadcastAt` in the options object, and the `DepositPollingInputs.now` doc now states the contract you identified: one injected clock drives every suppression TTL a single call judges.

Pinned at this level so the person you predicted - writing a refund-TTL case the same way - gets a red test instead of a silent `Date.now()` read: a new "refund suppression clock" describe drives both sides of the refund window through the injected `now`, no fake timers. The inside-the-window case is the load-bearing one: dropping the `now,` forward makes it fail, because the real clock sits years past the fixture's broadcast time. Verified by reverting the forward, not assumed.

One residual your framing surfaced, left alone deliberately: `formatExpiredTimeAgo` still reads the raw wall clock and is reachable from this compute via `getDisplay` when `expiredAt` is set. It only formats the "Xh ago" display string - it never feeds a TTL or an action decision - so the doc comment scopes the promise to suppression TTLs rather than claiming total purity. Happy to thread `now` through it in a follow-up if you think the header should promise more.

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: The defect was exactly as you described, but the suggested line does not produce the outcome it describes, so I went a different way - flagging the arithmetic since it changes what the fix should be. With `visualStep = 7 - 1 = 6`, `buildStepGroups(6)` computes group 1's status from `6 > 6`, which is false - so "Register deposit" stays **active-expanded**, not completed: the pill stays hidden (`completedGroups` counts `aggregateVisualStep > group.endStep`), the WOTS group stays collapsed, the hint stays suppressed, and row 6 renders as `active` - a spinner on "Confirming deposit", the one step that is provably done.

What landed instead: pre-entry keeps the **real** visual step, and `!started` now gates group *expansion* rather than zeroing progress - a `started` param on `buildStepGroups`, threaded through `GroupedProgress` and `SplitGroupedProgress`. A parked user now sees the "1 of 4 steps completed" pill, the 6/15 progress bar, and "Submit WOTS key" as the next collapsed header directly above the CTA - while nothing expands and nothing spins on an idle flow. Two details that fell out of making that honest end to end:

- A pre-entry group with none of its own work done also reads *not-started* rather than *active* - otherwise the step-1 entry's first header would flip to "In progress" styling and aria on a flow that has not begun. With that, `DepositSignContent` renders identically to before down to the aria labels, preserved by having nothing to show rather than by decrementing.
- On the split path, the gate applies only to columns mirroring the flow's own un-started step. Sibling lanes on a different step are driven by their own polled state - their expansion (and live confirmation-depth panel) reflects a genuinely running remote process and stays visible, as it did before.

On the hint: it stays hidden pre-click and appears the moment the user clicks (`started` flips, row 7 goes active, the detail renders) - which is when the approval it warns about actually fires and can be silently queued. Pre-click there is no pending approval to explain.

Tests added in `DepositProgressView.test.tsx` as requested - four cases under "pre-entry state (started=false)": completed work stays visible at a WOTS entry, nothing expands or announces "In progress" pre-click, the step-1 entry keeps its fully collapsed all-not-started render, and a sibling vault's live progress stays expanded on a split re-offer. Each was verified to fail against the code it guards by reverting that piece (the zeroing, the status downgrade, the per-column gate respectively). `buildStepGroups`'s own unit suite also covers the new parameter directly, including the mid-group pre-entry branch (visual step 8, one sub-step done: the group reads active for its real progress but still must not expand).

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: A decision, and now a documented one. The gate stays even on the path where the opening click was itself the gesture: a re-offer means the VP is still asking after a submission this session watched resolve, so something may genuinely be wrong, and confirming the fresh wallet popup on that abnormal path is worth more than the click it saves. Keeping it uniform also spares `PostDepositContinuationView` from having to report fresh-open vs branch-swap - a distinction that would exist only to shave one click off the failure path.

Both halves - the two-click cost on the row-click path and the reasoning for paying it - are in the comment above `isReoffer`, alongside the idle-modal rationale.

services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
jonybur: Dropped. That describe's `beforeEach` now carries only its mock setup, and the file-level reset is the single authority its comment claims to be. No ordering was lost: vitest runs outer hooks before inner ones, so the deleted call only ever re-ran the reset that had just executed.

services/vault/src/context/deposit/__tests__/optimisticDepositState.test.ts
jonybur: Added, essentially verbatim, right after the re-arm test - presence surviving the lapse while the freshness predicate flips false is exactly the disagreement that justifies two predicates, and your comment inside the test says so, so I kept it.

It also picked up a sibling as the clock-jump fix landed: `isWotsSubmissionWithinTtl` now rejects a timestamp ahead of the clock (jrwbabylonlab's backwards-jump case), and `hasWotsSubmissionRecord` deliberately does not - presence is presence regardless of what the clock did - which makes the two-predicate split earn its keep twice.

services/vault/src/context/deposit/optimisticDepositState.ts
jonybur: Fixed - the predicate now computes elapsed and requires `elapsedMs >= 0`, so a timestamp the clock has fallen behind reads as expired rather than suppressing until the clock catches back up past it. Expired is the safe direction on all three axes: the re-offer waits for a click (the auto-submit gate reads `hasWotsSubmissionRecord`, which is clock-independent), a redundant submission is a no-op the VP ignores, and because `markWotsSubmitted`'s re-arm guard reads this same predicate, the next submission re-records against the corrected clock - the marker self-heals instead of sticking in the future.

`isRefundBroadcastWithinTtl` had the identical shape, and its own doc already commits to fail-open ("the user can always retry"), so it got the same guard - the refund side is additionally backstopped by chain ground truth (`refundSettlement`), and a duplicate broadcast is rejected by the network.

The guard also reaches the refund predicate's third call site, `shouldRemoveFromLocalStorage`: a future-dated REFUND_BROADCAST marker on an EXPIRED contract now clears as stale rather than surviving until the clock catches back up - consistent with the same "expired is the safe reading" rule, and pinned with its own case.

All three new tests were verified to fail against the bare `< TTL` comparison by reverting the guard: the store test drives the predicate with an injected `now` behind the timestamp, the state-machine tests walk EXPIRED + REFUND_BROADCAST with `refundBroadcastAt` an hour ahead of `now` through both the action-suppression and localStorage-eviction paths.

services/vault/src/context/deposit/computeDepositPollingResult.ts
jonybur: Added - `now,` sits alongside `refundBroadcastAt` in the options object, and the `DepositPollingInputs.now` doc now states the contract you identified: one injected clock drives every suppression TTL a single call judges.

Pinned at this level so the person you predicted - writing a refund-TTL case the same way - gets a red test instead of a silent `Date.now()` read: a new "refund suppression clock" describe drives both sides of the refund window through the injected `now`, no fake timers. The inside-the-window case is the load-bearing one: dropping the `now,` forward makes it fail, because the real clock sits years past the fixture's broadcast time. Verified by reverting the forward, not assumed.

One residual your framing surfaced, left alone deliberately: `formatExpiredTimeAgo` still reads the raw wall clock and is reachable from this compute via `getDisplay` when `expiredAt` is set. It only formats the "Xh ago" display string - it never feeds a TTL or an action decision - so the doc comment scopes the promise to suppression TTLs rather than claiming total purity. Threading `now` through it would be cosmetic; happy to do it in a follow-up if you think the header should promise more.

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: The defect was exactly as you described, but the suggested line does not produce the outcome it describes, so I went a different way - flagging the arithmetic since it changes what the fix should be. With `visualStep = 7 - 1 = 6`, `buildStepGroups(6)` computes group 1's status from `6 > 6`, which is false - so "Register deposit" stays **active-expanded**, not completed: the pill stays hidden (`completedGroups` counts `aggregateVisualStep > group.endStep`), the WOTS group stays collapsed, the hint stays suppressed, and row 6 renders as `active` - a spinner on "Confirming deposit", the one step that is provably done.

What landed instead: pre-entry keeps the **real** visual step, and `!started` now gates group *expansion* rather than zeroing progress - a `started` param on `buildStepGroups`, threaded through `GroupedProgress` and `SplitGroupedProgress`. A parked user now sees the "1 of 4 steps completed" pill, the 6/15 progress bar, and "Submit WOTS key" as the next collapsed header directly above the CTA - while nothing expands and nothing spins on an idle flow. Two details that fell out of making that honest end to end:

- A pre-entry group with none of its own work done also reads *not-started* rather than *active* - otherwise the step-1 entry's first header would flip to "In progress" styling and aria on a flow that has not begun. With that, `DepositSignContent` renders identically to before down to the aria labels, preserved by having nothing to show rather than by decrementing.
- On the split path, the gate applies only to columns mirroring the flow's own un-started step. Sibling lanes on a different step are driven by their own polled state - their expansion (and live confirmation-depth panel) reflects a genuinely running remote process and stays visible, as it did before.

On the hint: it stays hidden pre-click and appears the moment the user clicks (`started` flips, row 7 goes active, the detail renders) - which is when the approval it warns about actually fires and can be silently queued. Pre-click there is no pending approval to explain.

Tests added in `DepositProgressView.test.tsx` as requested - four cases under "pre-entry state (started=false)": completed work stays visible at a WOTS entry, nothing expands or announces "In progress" pre-click, the step-1 entry keeps its fully collapsed all-not-started render, and a sibling vault's live progress stays expanded on a split re-offer. Each was verified to fail against the code it guards by reverting that piece (the zeroing, the status downgrade, the per-column gate respectively). `buildStepGroups`'s own unit suite also covers the new parameter directly, including the mid-group pre-entry branch (visual step 8, one sub-step done: the group reads active for its real progress but still must not expand).

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: A decision, and now a documented one. The gate stays even on the path where the opening click was itself the gesture: a re-offer means the VP is still asking after a submission this session watched resolve, so something may genuinely be wrong, and confirming the fresh wallet popup on that abnormal path is worth more than the click it saves. Keeping it uniform also spares `PostDepositContinuationView` from having to report fresh-open vs branch-swap - a distinction that would exist only to shave one click off the failure path.

Both halves - the two-click cost on the row-click path and the reasoning for paying it - are in the comment above `isReoffer`, alongside the idle-modal rationale.

services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
jonybur: Dropped. That describe's `beforeEach` now carries only its mock setup, and the file-level reset is the single authority its comment claims to be. No ordering was lost: vitest runs outer hooks before inner ones, so the deleted call only ever re-ran the reset that had just executed.

services/vault/src/context/deposit/__tests__/optimisticDepositState.test.ts
jonybur: Added, essentially verbatim, right after the re-arm test - presence surviving the lapse while the freshness predicate flips false is exactly the disagreement that justifies two predicates, and your comment inside the test says so, so I kept it.

It also picked up a sibling while the clock-jump fix landed: `isWotsSubmissionWithinTtl` now rejects a timestamp ahead of the clock (jrwbabylonlab's backwards-jump case), and `hasWotsSubmissionRecord` deliberately does not - presence is presence regardless of what the clock did - which makes the two-predicate split earn its keep twice.

gbarkhatov (approved):

PR #2146, opened 2026-07-28
refactor(sdk): drop unused Aave exports

jrwbabylonlab:
Deletes every export in the SDK's Aave integration that nothing consumes:
9 functions, 9 types, and 4 constant/ABI re-exports. Verified per symbol,
not per file — the only remaining matches are unrelated local props of the
same name in vault components and the e2e CLI.

Gone: `hasCollateral`, `hasDebt` (the client wrappers — `hasDebtFromPosition`
stays and is live), `calculateBorrowRatio`, `isHealthFactorHealthy`,
`checkRebalanceNeeded`, all of `vaultSelection.ts` and `seizureSimulation.ts`,
`USDC_DECIMALS`, `BPS_TO_PERCENT_DIVISOR`, `DepositorStruct`, and the
barrel-only re-exports of `WAD_DECIMALS` and `AaveSpokeABI` — those two
constants/JSON stay, since internal code uses them.

Also removes the three dead SDK re-exports from the vault's
`applications/aave/utils` barrel, and fixes two references the deletions
orphaned: a `vaultSplit.ts` comment naming `checkRebalanceNeeded` as a
caller, and the now-unused `hasDebtFromPosition` import in `spoke.ts`.


jonybur (dismissed):
Cleanly scoped dead-code removal. Verified per symbol: no repository consumer imports the removed exports (the remaining hasCollateral/hasDebt matches are unrelated local props in the vault and e2e CLI), the barrels and tests for the deleted modules are updated in lockstep, and hasDebtFromPosition/WAD_DECIMALS/AaveSpokeABI are correctly retained as live users. The vaultSplit.ts touch is limited to removing the unused rebalance-check API and its stale caller comment; the live seizure-fraction and split calculations are untouched.

packages/babylon-ts-sdk/docs/integrations/aave/quickstart.md
```diff
@@ -198,14 +197,14 @@ if (!position) throw new Error("No position found");
...
-const userHasDebt = await hasDebt(
+const remainingDebt = await getUserTotalDebt(
   publicClient,
   SPOKE,
   USDC_RESERVE_ID,
   proxyAddress,
 );
 
-if (userHasDebt) {
+if (remainingDebt > 0n) {
```
gbarkhatov: `getUserTotalDebt()` isn't a drop-in for the deleted `hasDebt()`. This is the one spot in the PR where a swap changes meaning rather than just removing something.

- `hasDebt()` wrapped `hasDebtFromPosition`, which checks `drawnShares > 0n || premiumShares > 0n` — **share level**.
- `getUserTotalDebt()` returns **token units**, and its own doc calls it "Display/routing only", notes it excludes the adapter's uncollected interest fee, and warns that "rounding can leave residual debt shares (dust)" (`clients/spoke.ts:200-234`).

So a position carrying residual dust shares now passes this guard, and the withdraw reverts on-chain instead. Production code still uses the share-level check — `services/vault/src/applications/aave/services/positionService.ts:208` gates on `hasDebtFromPosition`.

Both `getUserPosition` and `hasDebtFromPosition` survive this PR and are already exported from the barrel (`index.ts:84` and `:111`), so the original semantics are still expressible:

```ts
const position = await getUserPosition(
  publicClient,
  SPOKE,
  USDC_RESERVE_ID,
  proxyAddress,
);

if (hasDebtFromPosition(position)) {
  throw new Error("Repay all debt before withdrawing");
}
```

That needs `getUserPosition` and `hasDebtFromPosition` added to the import block above — keep `getUserTotalDebt`, the repay section still uses it.

Same fix applies to `docs/integrations/aave/README.md:90`, which now reads `` `getUserTotalDebt()` → must be 0 ``.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/vaultSplit.ts
```diff
@@ -310,71 +278,3 @@ export function computeMinDepositForSplit(
 
   return BigInt(Math.max(minFromProtected, minFromSacrificial));
 }
```
gbarkhatov, line 280: Removing `checkRebalanceNeeded` here — and `seizureSimulation.ts`, which imported it — leaves `assertSafePrecision` at line 33 exported with no cross-file consumer. It's in neither `utils/index.ts` nor `index.ts`, and no test imports it.

The function is still live (called at `:190` and `:258`), so only the `export` keyword is dead:

```ts
function assertSafePrecision(value: bigint, name: string): void {
```

Minor, but it's the same category this PR is cleaning up. The `useOptimalSplit.test.tsx:76` comment naming it describes behavior rather than importing it, so that can stay.

packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts
```diff
@@ -26,25 +26,6 @@ export const AAVE_FUNCTION_NAMES = {
  */
 export const BTC_DECIMALS = 8;
```
gbarkhatov: `BTC_DECIMALS` looks like it was missed by the same sweep that removed `USDC_DECIMALS` just below it. It has no consumers: the only other occurrence in the repo is a separate local `const BTC_DECIMALS = 8` declared in `services/vault/e2e/real/peginParams.ts:42`, and the barrel re-export at `index.ts:53` has no importers.

Not blocking — but by this PR's own criterion it should either go too, or be kept deliberately as public API with a line saying so.

packages/babylon-ts-sdk/src/tbv/integrations/aave/index.ts
```diff
@@ -133,16 +120,7 @@ export type {
...
-  PrefixSeizureResult,
-  RebalanceCheckParams,
-  RebalanceCheckResult,
-  SelectableVault,
-  TargetSeizureParams,
-  VaultSelectionResult,
 } from "./utils/index.js";
 
 // Export ABIs for application registration
 export { default as AaveIntegrationAdapterABI } from "./clients/abis/AaveIntegrationAdapter.abi.json";
```
gbarkhatov, line 125: Worth a deliberate call before merge: this barrel is the published API surface of `@babylonlabs-io/ts-sdk` (`0.1.2`, published to npm with `--access public` by `.github/workflows/package-release.yml`), and this PR removes 22 exports from it.

Releases here are conventional-commit driven — `nx.json` sets `release.version.conventionalCommits: true` over `packages/*`, and merging to main runs `ci:release` → `releaseVersion()` with no explicit specifier (`tools/release/index.ts:7-13`). Under nx's default type map, `refactor` is `semverBump: 'none'` with `changelog.hidden: true`.

So as titled, this breaking removal produces no version bump and no changelog entry — it rides into whatever the next `feat:` / `fix:` release happens to be, invisibly.

Two ways out, depending on how you want to version it:

- `refactor(sdk)!:` or a `BREAKING CHANGE:` footer — nx sets `isBreaking` and forces a **major**, so `0.1.2` → `1.0.0`. Accurate, but probably more than you want pre-1.0.
- `feat(sdk):` — lands `0.2.0` with a visible changelog entry; note the removals in the commit body. The conventional way to signal a break while still on `0.x`.

Either works. The current title is the one option that hides it.

gbarkhatov (dismissed):
Approving — the deletion is clean, and I verified it per symbol rather than taking the PR body on trust.

**Verification** (local checkout at `1d8f145`, identical to the PR head):

| Check | Result |
| --- | --- |
| `tsc --noEmit` (SDK) | clean |
| `eslint src/tbv/integrations/aave` | clean |
| `vitest run src/tbv/integrations/aave` | 104/104 pass |
| `typedoc --emit none` | 0 errors (484 warnings all pre-existing) |
| Grep for all 22 removed symbols, monorepo-wide | zero remaining references |
| Generated `docs/api/integrations/aave.md` line numbers vs source | match exactly |

The per-symbol claim holds up — the surviving `hasCollateral` / `hasDebt` matches are unrelated local props in the e2e CLI plus the `getHealthFactorStatus(healthFactor, hasDebt)` parameter, exactly as described. Nothing is left orphaned: `WAD_DECIMALS` still feeds `aaveConversions.ts:46`, `AaveSpokeABI` is imported directly by both `spoke.ts` and `oracle.ts`, `Hex` stays live in `types.ts` after `DepositorStruct` leaves, and `getUserPosition` keeps three real callers. I also checked `SEIZURE_TOL`, `simulateCascade` and `CascadeVault` — correctly retained, since `applications/aave/positionNotifications/` imports all three from the barrel.

Test deletions are proportional: ~488 lines removed, all covering deleted production code. No test for surviving code was dropped — `vaultSplit.test.ts` keeps 25 cases, `healthFactor.test.ts` keeps 15.

**On the `critical-path` label:** the `vaultSplit.ts` change (CLAUDE.md critical path #6) is purely subtractive — `checkRebalanceNeeded`, its two interfaces, and a stale comment. `computeOptimalSplit`, `computeSeizedFraction`, `computeSeizedFractionDetailed` and `computeMinDepositForSplit` are byte-identical and their tests are untouched. Low risk, but per CLAUDE.md it still needs a second human reviewer.

Four inline notes. The quickstart one is the only one with behavioral content; the rest are a missed constant, an orphaned `export` keyword, and a release-versioning call to make before merge. None block.

packages/babylon-ts-sdk/docs/integrations/aave/quickstart.md
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/vaultSplit.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts
jrwbabylonlab: fixed

gbarkhatov (approved):

PR #2147, opened 2026-07-28
test(vault): raise waitFor ceiling for cold async-gated renders

jrwbabylonlab:
`router.test.tsx > AaveConfigProvider wiring > renders the Activity page
heading` went red on main, failing `Unable to find ... "Activity"`. It passes
locally every time.

The heading is gated behind `AaveConfigProvider`, which shows a spinner until
its react-query fetch resolves. The test waits for the heading with `waitFor`
(1000ms default). On CI the first cold render of that heavy graph took 1102ms —
just over the ceiling — while the identical assertion later in the file ran
warm at 56ms and passed. Nothing in the failing code changed; added tests
elsewhere shifted vitest's worker distribution and raised contention enough to
tip the cold render past 1s.

Fix: set `asyncUtilTimeout: 4000` in the test setup. `waitFor` returns the
instant an assertion passes, so green tests don't slow; only a failing
assertion's wait extends, kept under vitest's 5000ms testTimeout so the
descriptive error still surfaces first.

### DIFF
--- a/services/vault/src/test/setup.ts
+++ b/services/vault/src/test/setup.ts
@@ -3,7 +3,13 @@
  */
 
 import "@testing-library/jest-dom";
+import { configure } from "@testing-library/react";
 import { afterAll, beforeAll, vi } from "vitest";
+
+// Cold async-gated page renders (e.g. /activity behind AaveConfigProvider)
+// exceed the 1000ms waitFor default under CI load; a passing assertion still
+// returns immediately, so raising the ceiling only helps flakes.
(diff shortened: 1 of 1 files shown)


jonybur (approved):
Test-only bump: raising asyncUtilTimeout to 4000ms only extends the wait on a failing assertion, so green tests still return immediately, and it stays below the 5000ms testTimeout so the underlying assertion error still surfaces first. The inline comment captures why the flake showed up (cold render behind AaveConfigProvider under CI load) and why the fix is safe.

PR #2148, opened 2026-07-28
feat(vault): add Market Info action and borrowing markets page shell

jeremy-babylonlabs:
Screenshot 2026-07-28 at 15 01 38


services/vault/src/applications/aave/components/Detail/index.tsx
```diff
@@ -1,202 +1,147 @@
...
-          />
-        </>
-      )}
-    </>
+    <V3ModalShell
+      open
+      // Withholding `onClose` hides the X and no-ops the backdrop click;
+      // `disableEscapeClose` covers ESC. Together they lock every dismiss path.
+      onClose={isTxInFlight ? undefined : close}
+      disableEscapeClose={isTxInFlight}
```
gbarkhatov, line 153: **The success step loses its X, backdrop and ESC.**

`isTxInFlight` is now shared across all three steps of one shell, but nothing clears it when the flow advances to success:

- `useBorrowTransaction` does `setIsProcessing(false)` in its `finally` (`useBorrowTransaction.ts:159`), then the form calls `onBorrowSuccess` in the awaited continuation (`Borrow/index.tsx:244`). Same for repay (`useRepayTransaction.ts:258` / `Repay/index.tsx:252`).
- React batches both into one commit, `showSuccess` flips true, and `ReserveDetailPanel` unmounts in that same commit — so the `onProcessingChange(isProcessing)` effect (`Borrow/index.tsx:131-133`, `Repay/index.tsx:148-150`) never fires with `false`, and there is no unmount cleanup.

Net: `isTxInFlight` stays `true` on the success step, so `onClose` is withheld and `disableEscapeClose` is on. Only the Done CTA dismisses. That's a regression — previously the success screen was its own `V3ModalShell` with an unconditional `onClose`, and `LoanSuccessPanel`'s own JSDoc still promises "the Done CTA, the close (X), the backdrop, or escape — all four land on the same `onDone`".

Simplest fix is to clear it where success lands, e.g. `onSuccess={(s) => { setIsTxInFlight(false); setSuccess(s); }}` on line 113 (gating on `!showSuccess` here would work too).

`LoanFlowOverlay.test.tsx` mocks `ReserveDetailPanel` and never calls `onProcessingChange`, so nothing covers this — worth a case that reports processing, then settles, and asserts the shell still has a close control. Manual check: complete a repay, press ESC.

services/vault/src/applications/aave/components/Detail/index.tsx
```diff
@@ -1,202 +1,147 @@
...
+  const [success, setSuccess] = useState<LoanSuccessState | null>(null);
 
-  // Use `replace` so dismissing the overlay doesn't leave a history entry that
-  // browser Back would use to reopen the just-closed flow.
-  const detailBaseRoute = getReserveDetailBaseRoute(FeatureFlags.isV3UiEnabled);
+  // Same source the dashboard reads, so React Query serves both from one entry.
+  const { position, debtValueUsd } = useAaveUserPosition(
+    isConnected ? address : undefined,
+  );
+  const { borrowedAssets } = useAaveBorrowedAssets({ position, debtValueUsd });
```
gbarkhatov, line 65: **Cold `?picker=repay` renders a false empty state.**

`repayAssets` used to come from `useDashboardState`, which was already resolved by the time the Repay button was clickable. Sourcing them here means that on a direct load / refresh of `/loans?picker=repay` the position query is still in flight, `borrowedAssets` is `[]`, and `AssetSelectionPanel` falls straight through to `COPY.loans.assetSelection.emptyRepay` — "No assets available" for a user who does have debt.

The panel's loading branch is guarded by `!isRepay` (`AssetSelectionPanel.tsx:173-174`) and its comment "Repay assets arrive ready" is now stale — that was true for the old caller, not this one. Suggest threading the position query's `isLoading` through as a prop and widening that branch to cover repay mode.

services/vault/src/hooks/useLoanActions.ts
```diff
@@ -48,46 +33,18 @@ export function useLoanActions({
   );
 
   const openBorrowPicker = useCallback(() => {
-    setAssetModalMode(LOAN_TAB.BORROW);
-    setIsAssetModalOpen(true);
-  }, []);
+    navigate(getAssetPickerRoute(LOAN_TAB.BORROW, isV3));
```
gbarkhatov, line 36: **Opening the picker pushes a history entry, so Back after closing reopens the just-closed flow.**

The picker used to be local state and left no entry. Now: `/loans` (A) → push `?picker=borrow` (B) → push `?reserve=…` (C), and `close()` only `replace`s the top entry, leaving `[A, B, /loans]`. One Back lands on B and the picker reopens — which is exactly what the `replace` comment in `Detail/index.tsx:85-86` says it prevents.

`{ replace: true }` on both picker navigations (here and line 46) keeps the whole flow in one entry while preserving the picker → form push, so Back from the form still returns to the picker.

services/vault/src/applications/aave/components/AssetSelectionPanel/AssetSelectionPanel.tsx
```diff
@@ -191,64 +207,83 @@ export function AssetSelectionModal({
...
-                    <span className="flex-1">{row.aprLabel}</span>
-                  </>
-                )}
-              </div>
-            </button>
+              </button>
+              {!isRepay && (
+                <button
+                  type="button"
+                  onClick={() => handleMarketInfoClick(row.symbol)}
```
gbarkhatov, line 282: **Market Info dead-ends when the v3 flag is off.**

`isV3UiEnabled` reads `NEXT_PUBLIC_FF_ENABLE_V3_UI === "true"` and is commented out in `.env.example`, so v2 is the default. The picker is still reachable there — `getAssetPickerRoute(tab, false)` resolves to `/?picker=borrow`, which matches `AaveOverlayLayout`'s `pathname === getReserveDetailBaseRoute(false)` guard — and this button renders unconditionally. But `getMarketDataRoute` takes no flag, and `/markets/:symbol` renders `<Navigate to={ROUTES.OVERVIEW} replace />` when the flag is off.

So in v2 the user clicks Market Info, the overlay silently closes, and they land on the dashboard. Gate the button on `FeatureFlags.isV3UiEnabled` (or thread the flag through `getMarketDataRoute` the way the other route helpers do).

services/vault/src/applications/aave/components/AssetSelectionPanel/AssetSelectionPanel.tsx
```diff
@@ -191,64 +207,83 @@ export function AssetSelectionModal({
...
+                  aria-label={COPY.loans.assetSelection.marketInfoAriaLabel(
+                    row.symbol,
+                  )}
+                  className={`${NEUTRAL_ROW_BUTTON_CLASS} cursor-pointer`}
```
gbarkhatov: Nit: this is the only `NEUTRAL_ROW_BUTTON_CLASS` consumer that appends `cursor-pointer` — `ActiveLoansList.tsx:81,89`, `VaultsActiveSection.tsx:124` and `VaultsLifecycleSections.tsx` all use it bare. Either fold `cursor-pointer` into the shared class in `buttonClasses.ts` or drop it here, so the row buttons don't drift.

services/vault/src/applications/aave/components/AssetSelectionPanel/AssetSelectionPanel.tsx
```diff
@@ -62,19 +63,35 @@ interface AssetRow {
 /** Width of the leading Asset column; the stats share the remaining row. */
 const ASSET_COL_CLASS = "flex w-[220px] shrink-0 items-center gap-4";
 
-export function AssetSelectionModal({
-  isOpen,
-  onClose,
+/** Header spacer for the Market Info column: the button's 120px minimum plus
+ *  the row's 16px gap, so the stat columns line up with the rows below. */
+const MARKET_INFO_COL_CLASS = "w-[136px] shrink-0";
```
gbarkhatov: Nit: `w-[136px]` hardcodes `NEUTRAL_ROW_BUTTON_CLASS`'s `min-w-[120px]` plus the row's 16px gap. The comment says so, but nothing links the two — a change to the shared button's min-width silently misaligns this header against the rows. A derived value (or at least a pointer to `buttonClasses.ts` in the comment) would keep them honest.

services/vault/src/components/pages/BorrowingMarketsData.tsx
```diff
@@ -0,0 +1,144 @@
...
+      ),
+    };
+  }, [borrowableReserves, symbol]);
+
+  return (
+    <Container className={`${PAGE_CONTENT_CLASS} pb-6`}>
+      <div className="space-y-5">
+        <button
+          type="button"
+          onClick={() => navigate(-1)}
```
gbarkhatov: **`navigate(-1)` can exit the app while the label promises "Back to Assets".**

`/markets/wbtc` is a shareable URL — opened in a new tab or from a bookmark there is no in-app entry to go back to, and this walks out of the SPA. `getAssetPickerRoute` already exists for exactly this destination; fall back to it when `useLocation().key === "default"`.

services/vault/src/copy.ts
```diff
@@ -1178,6 +1185,24 @@ export const COPY = {
         "Position integrity check failed: your position details returned by the indexer don't match what's registered on-chain. Refresh and try again. If this persists, do not proceed.",
     },
   },
+  // Borrowing markets data page (Figma node 10088-60956).
+  marketData: {
+    pageTitle: "Borrowing markets data",
+    backToAssets: "Back to Assets",
```
gbarkhatov: Nit: `backToAssets` ("Back to Assets") and `borrowMarkets.title` ("Borrow Markets", line 1201) are Title Case while `interestRateModel.title` ("Interest rate model") on the same page is sentence case. `copy.ts`'s style header exempts button labels only, so these three read inconsistently side by side. Suggest "Back to assets" / "Borrow markets".

services/vault/src/components/pages/__tests__/Loans.test.tsx
```diff
@@ -123,7 +123,6 @@ const CONNECTED_LOADED = {
   borrowedAssets: [],
   hasLoans: false,
   hasCollateral: true,
-  selectableBorrowedAssets: [],
   isBorrowCapacityLoading: false,
```
gbarkhatov, line 122: While you're cleaning up this fixture: line 67 still has `vi.mock("@/applications/aave/components/AssetSelectionModal", …)` for a module this PR deletes. It passes only because nothing imports that path anymore, so vitest never resolves it — a dangling reference to removed code. Same block at `DashboardPage.test.tsx:84`.

services/vault/src/components/simple/__tests__/DashboardPage.test.tsx
```diff
@@ -42,7 +42,6 @@ vi.mock("@/hooks/useDashboardState", () => ({
     hasCollateral: true,
     hasDisplayCollateral: true,
     collateralVaults: [],
-    selectableBorrowedAssets: [],
     isLoading: false,
```
gbarkhatov, line 45: Same as the Loans fixture: line 84 still mocks the deleted `@/applications/aave/components/AssetSelectionModal` module. Worth removing alongside `selectableBorrowedAssets`.

gbarkhatov (dismissed):
Nice refactor — collapsing the three stacked dialogs into one route-driven `LoanFlowOverlay` is the right call, and the reasoning (an opaque `.bbn-dialog-fullscreen` can't cross-fade cleanly) is well documented at the top of `Detail/index.tsx`. The decomposition is clean: `ReserveDetailPanel` splits out cleanly, `useBorrowRepayModals` / `selectableBorrowedAssets` are properly deleted along with their consumers, and the hardcoded strings move into `copy.ts`. Pricing repay rows off `allBorrowReserves` so debt in a frozen reserve still gets a price is a real bug fix, and it's covered by a test.

No CRITICAL PATHS file is touched — no WASM boundary, fee math, presigning, secret derivation or split-transaction code — and the load-bearing E2E testids (`asset-select-row-*`, `loan-success-done-button`) are carried over correctly. Lint is clean and all affected suites pass locally (68 tests).

Approving, but there are two I'd want fixed before this goes out, both inline:

1. **The success step loses its X, backdrop and ESC** — `isTxInFlight` is never cleared when the flow advances to success, so all three dismiss paths stay locked and only the Done CTA works. Regression vs. the old separate success shell.
2. **Market Info dead-ends when the v3 flag is off** — the picker is reachable in v2, the button renders unconditionally, and `/markets/:symbol` redirects to the dashboard.

Then three smaller behavioral ones (Back-after-close reopening the picker, the cold `?picker=repay` false empty, `navigate(-1)` leaving the app), plus dead mocks for the deleted `AssetSelectionModal` module.

A few things I couldn't anchor to the diff:

- `Borrow/index.tsx:130` and `Repay/index.tsx:147` still say "see AaveReserveDetail" — the component this PR renamed. `Detail/__tests__/PositionGate.test.tsx:3` too.
- The three step widths now live in three different files (`FORM_WIDTH_CLASS` in `Detail/index.tsx`, `LOAN_SUCCESS_WIDTH_CLASS` in `LoanSuccessPanel.tsx`, `getAssetPickerWidthClass` in `AssetSelectionPanel.tsx`). Co-locating them would also clear the two new `react-refresh/only-export-components` warnings this PR adds.
- Test gaps: `__tests__/router.test.tsx` covers flag-on/flag-off for every other v3 route but not `/markets/:symbol`, and `hooks/__tests__/usePageTitle.test.ts` isn't extended for the new `OFF_NAV_ROUTE_TITLES` entry.
- The `waitFor` bump from 1s to 10s is reasonably justified in its comment, but it does mean a genuinely broken lazy chunk now takes 10s to fail.
- `e2e/real/actions/repay.ts` drives this picker, which is now route-driven rather than local state — worth a real-wallet run before merge even though the testids are unchanged.

services/vault/src/routes.ts
```diff
@@ -30,3 +40,7 @@ export function getReserveDetailRoute(
   });
   return `${baseRoute}?${params.toString()}`;
 }
+
+export function getMarketDataRoute(assetSymbol: string) {
```
jrwbabylonlab, line 44: This file conflicts with #2153, which changes `getReserveDetailRoute` to take a `bigint` reserve id (audit F7). Resolving toward this PR reinstates symbol routing and undoes the fix.

Also: `BorrowingMarketsData.tsx` resolves this route via `borrowableReserves.find(r => r.token.symbol...)` — the pattern #2153 removes. Display-only, so nothing at risk yet.

jonybur (approved):
LGTM. The two blocking items raised on the prior head are fixed here:

- `LoanFlowOverlay.onSuccess` now calls `setIsTxInFlight(false)` before flipping to the success step, so the X / backdrop / ESC controls return after a tx settles. The new "restores the close control on the success step after a tx settles" test locks that in.
- Market Info is gated on `!isRepay && featureFlags.isV3UiEnabled`, and `/markets/:symbol` redirects to Overview when v3 is off, so v2 depositors can no longer land on a dead-end button or page.

Also verified independently:

- The one-dialog step model derives the step from URL + local success: `showSuccess` requires `success.reserveId === reserveId`, and the effect drops the local success as soon as the route leaves that reserve. So Back off the form doesn't keep the success visible, and a settled success never bleeds into a different asset.
- Asset picker step lives on the same route as the form (`?picker=borrow|repay`); the reserveId is lowercased through `getReserveDetailRoute`, matching how the router parses it back.
- `AssetSelectionPanel` in repay mode prices from `allBorrowReserves` rather than `borrowableReserves`, so a frozen or paused reserve carrying debt still gets a price. The `assetsLoading` wiring holds the loading state on a cold `?picker=repay` load instead of falsely showing "No assets available".
- `BorrowingMarketsData` back button distinguishes `location.key === "default"` (direct URL, so navigate into the borrow picker) from an in-app entry (`navigate(-1)`), so the shareable URL doesn't leave the app.
- Load-bearing E2E test-ids (`asset-select-row-*`, `loan-success-done-button`) carry over intact.
- No CRITICAL PATHS file touched.

PR #2149, opened 2026-07-28
fix(vault): anchor risk rail gradient on the liquidation price

jeremy-babylonlabs:
The gradient stops were placed relative to the current-price marker, so the ramp slid with the marker: at a health factor of 9 the rail rendered red right up to a green marker under a green "Safe" label.

Place the stops by price instead — red at the liquidation price, green at the price clearing the safe threshold, with a minimum ramp width so the scale stays legible when liquidation is far below. The current-marker ring now also reads amber in the moderate state.

Wire the god-mode health-factor override into the risk card so the forced states chart: a forced health factor implies the liquidation price that produces it, and the card derives its price, distance and label from that.

## Screenshot
Screenshot 2026-07-28 at 17 22 39
Screenshot 2026-07-28 at 17 22 36
Screenshot 2026-07-28 at 17 22 32

### DIFF
--- a/services/vault/src/components/simple/DashboardPage.tsx
+++ b/services/vault/src/components/simple/DashboardPage.tsx
@@ -12,12 +12,14 @@ import { AssetSelectionModal } from "@/applications/aave/components/AssetSelecti
 import { useSyncPendingVaults } from "@/applications/aave/context";
 import { useAaveVaults } from "@/applications/aave/hooks";
 import { usePositionNotifications } from "@/applications/aave/hooks/usePositionNotifications";
+import { getHealthFactorStatusFromValue } from "@/applications/aave/utils";
 import type { RootLayoutContext } from "@/components/pages/RootLayout";
 import { PAGE_CONTENT_CLASS } from "@/components/shared/layoutClasses";
 import featureFlags from "@/config/featureFlags";
 import { useConnection, useETHWallet } from "@/context/wallet";
 import { COPY } from "@/copy";
 import {
+  useDebugHealthFactorOverride,
   useDebugManualMode,
   useDebugManualParams,
   useDebugPositionOverride,
@@ -252,9 +254,6 @@ export function DashboardPage() {
     btcMetadata !== undefined &&
     !btcMetadata.isStale &&
     !btcMetadata.fetchFailed;
-  const liquidationPrice = firstLiquidationGroup
-    ? formatUsdPrice(firstLiquidationGroup.liquidationPrice)
-    : COPY.common.emptyValue;
   const usableBtcPriceUsd =
     isBtcPriceUsable && btcPriceUsd !== undefined && btcPriceUsd > 0
       ? btcPriceUsd
@@ -263,10 +262,29 @@ export function DashboardPage() {
     usableBtcPriceUsd !== null
       ? formatUsdPrice(usableBtcPriceUsd)
       : COPY.common.emptyValue;
-  const liquidationPriceUsd = firstLiquidationGroup?.liquidationPrice ?? null;
-  const pctToLiquidation = firstLiquidationGroup
-    ? formatLiquidationDistancePercent(-firstLiquidationGroup.distancePct)
-    : COPY.common.emptyValue;
+  // God-mode override (dev only; null in production). Health factor is
+  // btcPrice / liquidationPrice, so a forced value implies the liquidation
+  // price that produces it — the rail is charted from that price, not the HF.
+  const healthFactorOverride = useDebugHealthFactorOverride();
+  const shownHealthFactor = healthFactorOverride ?? healthFactor;
+  const shownHealthFactorStatus =
+    healthFactorOverride !== null
+      ? getHealthFactorStatusFromValue(healthFactorOverride)
+      : healthFactorStatus;
+  const liquidationPriceUsd =
+    healthFactorOverride !== null && usableBtcPriceUsd !== null
+      ? usableBtcPriceUsd / healthFactorOverride
+      : (firstLiquidationGroup?.liquidationPrice ?? null);
+  const liquidationPrice =
+    liquidationPriceUsd !== null
+      ? formatUsdPrice(liquidationPriceUsd)
+      : COPY.common.emptyValue;
+  const pctToLiquidation =
+    healthFactorOverride !== null
+      ? formatLiquidationDistancePercent(100 * (1 - 1 / healthFactorOverride))
+      : firstLiquidationGroup
+        ? formatLiquidationDistancePercent(-firstLiquidationGroup.distancePct)
+        : COPY.common.emptyValue;
   const collateralFactorText =
     collateralFactorBps !== null
       ? formatBasisPointsAsPercent(collateralFactorBps)
@@ -332,8 +350,8 @@ export function DashboardPage() {
 
         {showOverview && (
           <OverviewSection
-            healthFactor={healthFactor}
-            healthFactorStatus={healthFactorStatus}
+            healthFactor={shownHealthFactor}
+            healthFactorStatus={shownHealthFactorStatus}
             totalCollateralValue={totalCollateralValue}
             totalBorrowed={totalBorrowed}
             liquidationPrice={liquidationPrice}
@@ -357,8 +375,8 @@ export function DashboardPage() {
 
         {featureFlags.isV3UiEnabled && (
           <RiskSection
-            healthFactor={healthFactor}
-            healthFactorStatus={healthFactorStatus}
+            healthFactor={shownHealthFactor}
+            healthFactorStatus={shownHealthFactorStatus}
             hasPosition={hasOverviewData}
             liquidationPriceText={liquidationPrice}
             btcPriceText={btcPrice}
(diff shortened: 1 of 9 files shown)


services/vault/src/components/simple/RiskSection/railLayout.ts
```diff
@@ -43,10 +44,10 @@ export interface RailLayout {
...
 const MARKER_EDGE_MARGIN_PCT = 1.5;
 const MAX_TICKS = 8;
-// Over the rail, red ends at the current-price marker, amber peaks ~21
-// points later, green runs to the end.
-const GRADIENT_AMBER_OFFSET = 21;
-const GRADIENT_GREEN_STOP = 100;
+// On this axis the health factor is price / liquidationPrice, so red sits at
+// the liquidation price and green at the price clearing the safe threshold.
+// Far from liquidation that ramp collapses to a hairline, hence the minimum.
+const GRADIENT_MIN_RAMP_PCT = 45;
```
gbarkhatov, line 54: Worth saying in this comment that the minimum has a truthfulness cost, not just a legibility benefit: once it engages the stops stop marking the HF bands. At HF 9 ($63,488 / $6,962) the honest safe-threshold stop is at 5.3% but green lands at 45.3%, so ~40pp of rail reads amber for prices that are comfortably safe. The marker itself stays green (81.7%), so the reported bug is fixed either way — but the next reader will otherwise assume the amber stop still means "HF 1.25 here".

services/vault/src/components/simple/RiskSection/railLayout.ts
```diff
@@ -103,25 +104,27 @@ export function computeRailLayout(
...
-  // midpoint instead left the marker sitting in the red band while it was
-  // rendered green.
   let gradient: string | null = null;
-  if (liquidationPct !== null) {
-    const amber = Math.min(currentPct + GRADIENT_AMBER_OFFSET, 100);
+  if (hasLiquidation) {
+    const stopPct = (price: number) => clamp(pct(price), 0, 100);
+    const red = stopPct(liquidationPriceUsd);
+    const green = Math.max(
+      stopPct(liquidationPriceUsd * HEALTH_FACTOR_WARNING_THRESHOLD),
```
gbarkhatov, line 133: This binds the rail's colours to the HF bands through the liquidation price, which is only equivalent if the HF labelling the card and this liquidation price come from the same inputs. They don't: the displayed HF is on-chain `accountData.healthFactor` (Aave oracle, `useAaveUserPosition.ts:183`), while `liquidationPriceUsd` comes from the cascade, computed off indexed vault rows plus the `usePrices` BTC feed. `usePositionNotifications` already carries a guardrail for exactly that divergence ("stale indexed rows inflated `totalBtc`" — it forces urgency off the on-chain HF when the cascade misses it).

When those sources disagree, the ring/label and the band under the marker drift apart again — the same class as the bug being fixed, just rarer and smaller. Two options: derive the green stop from the HF that labels the card (`green = pct(currentPriceUsd / (healthFactor / HEALTH_FACTOR_WARNING_THRESHOLD))`, passing HF into `computeRailLayout`), or keep the price-side derivation and note the shared-basis assumption here.

services/vault/src/components/simple/RiskSection/RiskPriceRail.tsx
```diff
@@ -107,7 +106,9 @@ export function RiskPriceRail({
   const currentRingCss =
     state === "liquidatable"
       ? "rgb(var(--risk-red))"
-      : "rgb(var(--risk-green))";
+      : state === "moderate"
```
gbarkhatov, line 109: The amber ring is the other user-visible change in this PR and nothing covers it. `RiskSection.test.tsx` already queries `risk-marker-current`, so it's a couple of lines: render with `healthFactorStatus: "warning"` and assert the marker's `style.border` resolves to `--risk-amber` (and stays green for `safe`).

services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -263,10 +262,29 @@ export function DashboardPage() {
...
       : COPY.common.emptyValue;
-  const liquidationPriceUsd = firstLiquidationGroup?.liquidationPrice ?? null;
-  const pctToLiquidation = firstLiquidationGroup
-    ? formatLiquidationDistancePercent(-firstLiquidationGroup.distancePct)
-    : COPY.common.emptyValue;
+  // God-mode override (dev only; null in production). Health factor is
+  // btcPrice / liquidationPrice, so a forced value implies the liquidation
+  // price that produces it — the rail is charted from that price, not the HF.
+  const healthFactorOverride = useDebugHealthFactorOverride();
+  const shownHealthFactor = healthFactorOverride ?? healthFactor;
```
gbarkhatov: These six lines are now verbatim identical to `Loans.tsx:76-82`. Worth pulling into a `useShownHealthFactor(healthFactor, healthFactorStatus)` next to `useDebugHealthFactorOverride` so the two pages can't drift on how a forced value maps to a status.

services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -263,10 +262,29 @@ export function DashboardPage() {
...
+  const shownHealthFactorStatus =
+    healthFactorOverride !== null
+      ? getHealthFactorStatusFromValue(healthFactorOverride)
+      : healthFactorStatus;
+  const liquidationPriceUsd =
+    healthFactorOverride !== null && usableBtcPriceUsd !== null
+      ? usableBtcPriceUsd / healthFactorOverride
+      : (firstLiquidationGroup?.liquidationPrice ?? null);
```
gbarkhatov, line 288: When `usableBtcPriceUsd` is null the override branch falls through to the real cascade price, so the card renders a forced health factor beside a live liquidation price and a live distance. Dev-only, but it's confusing precisely when you're using god mode to inspect the stale-price path. Forcing `null` here (placeholder) would keep the forced card internally consistent.

services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -263,10 +262,29 @@ export function DashboardPage() {
...
+  const liquidationPrice =
+    liquidationPriceUsd !== null
+      ? formatUsdPrice(liquidationPriceUsd)
+      : COPY.common.emptyValue;
+  const pctToLiquidation =
+    healthFactorOverride !== null
+      ? formatLiquidationDistancePercent(100 * (1 - 1 / healthFactorOverride))
```
gbarkhatov: No test covers this derivation. `Loans.test.tsx` already mocks `useDebugHealthFactorOverride` (see its 0.95 / 1.25 cases) — the same shape here would pin both the implied liquidation price and the distance, including the sub-1 clamp to `0.00%`.

services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -357,8 +375,8 @@ export function DashboardPage() {
 
         {featureFlags.isV3UiEnabled && (
           <RiskSection
-            healthFactor={healthFactor}
-            healthFactorStatus={healthFactorStatus}
+            healthFactor={shownHealthFactor}
+            healthFactorStatus={shownHealthFactorStatus}
             hasPosition={hasOverviewData}
```
gbarkhatov: A forced health factor doesn't reach `hasPosition`, so on a wallet with no position `getRiskDisplayState` returns `noPosition` and the override silently charts nothing — solid green rail, "No position" label. `Loans.tsx:94-103` already solves this with `godModeAffectsPage`; mirroring it here would make the forced states reachable without first setting up a real position. Dev-only.

gbarkhatov (dismissed):
Approving — the core fix is right and the math holds.

I checked the inverse the god-mode wiring relies on: the first cascade group's price is `remainingDebt / (remainingBtc * CF)` (`calculate.ts:172`), so `HF = price / liqPrice` is exact for the charted group, and `100 * (1 - 1/HF)` reproduces `-distancePct` identically. `formatLiquidationDistancePercent` clamps at 0, so a sub-1 forced HF renders `0.00%` — same as the live path. Ran the suite for the touched area (23 passing) and ESLint on the four files (clean).

Comments below are follow-up grade, nothing blocking. One item doesn't fit inline because the file isn't in the diff: `GodModePanel.tsx:449` still frames the health-factor override as "Loans-summary overrides (v3 /loans)" under a "Loans summary" heading. It now also drives the dashboard Overview and Risk cards, so both the comment and the section title understate its reach.

jonybur (approved):
LGTM. Verified independently:

- `resolveShownHealthFactor` correctly re-derives status from the override with the production `getHealthFactorStatusFromValue`, so a forced card can't drift on banding.
- `railLayout.ts` math checks out: red at `pct(liquidationPriceUsd)`, green at `pct(liquidationPriceUsd * HEALTH_FACTOR_WARNING_THRESHOLD)` or `red + GRADIENT_MIN_RAMP_PCT`, whichever is larger, so the red stop always lands on the liquidation marker. The 45pp minimum-ramp trade-off is documented and covered by the "keeps a readable ramp" test.
- The DashboardPage god-mode wiring correctly derives `liquidationPriceUsd = usableBtcPriceUsd / healthFactorOverride` and only when a usable BTC price backs it - so a stale price path renders placeholders rather than mixing forced HF with live liquidation numbers, and the test for that behavior is present.
- No hardcoded protocol parameters. No inline user-facing strings. `HEALTH_FACTOR_WARNING_THRESHOLD` is re-exported from the SDK, not duplicated. The removed constants (`GRADIENT_AMBER_OFFSET`, `GRADIENT_GREEN_STOP`) are gone with no dangling references. No critical-path file touched.

PR #2151, opened 2026-07-28
feat/liquidation dashboard page

jeremy-babylonlabs:


jonybur (dismissed):
LGTM. The page is UI scaffolding behind `isV3UiEnabled && isLiquidationAnalysisChartEnabled`, and the fixture cascade is explicitly labeled as the whole data seam that the follow-up PR replaces with `usePositionNotifications().result`. Independent checks hold:

- `axisTickFractions` math: sqrt-compressed gap sizes clamped up to `AXIS_SEGMENT_MIN_FRACTION`, with the remaining segments scaled down so the total is exactly 1. Cannot overflow because `minFraction = Math.min(0.12, 1/gaps.length)`, so `smallTotal <= 1` for every input.
- Static-axis switch to `axisTopPrice`: pinning the axis top to the live price (not the simulated one) lets `segmentedFraction` interpolate the price line through each band, instead of re-segmenting the axis around itself and pinning the line mid-band during a fall. The "keeps the axis static under simulation" test locks the tick set in.
- Header `LiquidationSummary` counts a group as liquidated on `btcPrice <= liquidationPrice`, the same rule the band state uses, so the chart and the header can't disagree at any simulated price. Verified against the fixture: falling to $40,283 seizes 0.6 + 0.4 = 1.0 of 1.1 BTC (~91%), which is what the simulator test asserts.
- `SimulationToolbar.PriceInput` commits on blur/Enter, clamps to `[floorPrice, livePrice]`, and Reset is disabled at the live price, tested via the 10M clamp case.
- No CRITICAL PATHS file touched; the fixture is a page-local module unreachable outside the two-flag gate.

`livePrice = FIXTURE_BTC_PRICE` is a static $88,400 rather than a wired price feed - the header comment already promises the follow-up will replace this, so noting it for the record, not as a blocker.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -143,6 +202,7 @@ function toCard(
     key: String(group.index),
     title: COPY.liquidations.eventTitle(group.index + 1),
     badge: sacrificial ? "sacrificial" : "protected",
+    tone: toneFor(group.index),
```
jrwbabylonlab: `calculate()` emits 1-based `group.index` (`groupIndex = 1`; the dust branch hardcodes `index: 1`) and `usePositionNotifications` passes it through unchanged, but this projection assumes 0-based. With live data: `sacrificial` (`index === 0`) never matches so the badge disappears, titles render "Event 2/3/4", and this `tone` diverges from the band's `toneFor(i)` — the mismatch it exists to prevent. Suggest keying title/badge/tone off array position instead of `group.index`.

services/vault/src/components/pages/Liquidations/fixtures.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const distanceFrom = (liquidationPrice: number) =>
+  (liquidationPrice / FIXTURE_BTC_PRICE - 1) * 100;
+
+const groups: LiquidationGroup[] = [
+  {
+    index: 0,
+    vaults: [{ id: "vault-1", name: "Vault 1", btc: 0.6 }],
+    combinedBtc: 0.6,
+    liquidationPrice: 77_682,
```
jrwbabylonlab: With this fixture's own inputs the first trigger is `debt/(btc·CF)` = 44287/(1.1×0.5) = **$80,522**, not $77,682 (liquidation fires at HF < 1; `AaveAdapterActionLiquidation.sol:276`), so the page understates risk (−8.9% real vs −12.1% shown). Running `calculate()` on this position also cannot produce this 3-event cascade — group 1's over-seizure fairness-repays the remaining debt in one event under any plausible (THF, LB) — and `debtRepaid` here excludes `fairnessDebtRepay`, contrary to the calculator's `debtRepaid = debtToRepay + fairnessDebtRepay`. CF 0.5 / 5% bonus also match no vault-contracts-aave-v4 config (deploy defaults: 75% / 10% bonus). Since this page exists for stakeholder review, consider deriving the fixture from a real `calculate()` run over a position that genuinely cascades in 3 events.

services/vault/src/components/pages/Liquidations/SimulationToolbar.tsx
```diff
@@ -0,0 +1,156 @@
...
+}) {
+  const [draft, setDraft] = useState<string | null>(null);
+
+  useEffect(() => {
+    setDraft(null);
+  }, [price]);
+
+  const commit = () => {
+    if (draft === null) return;
+    const parsed = Number(draft.replace(/[^0-9.]/g, ""));
```
gbarkhatov: **Blank entry jumps the simulation to the floor.** `Number("")` is `0`, which is finite, so select-all + delete + blur clamps to `min` (~$3,417) and paints the whole position as liquidated. Same for anything that strips to empty (`"abc"` -> `""` -> `0`), whereas `"1.2.3"` -> `NaN` correctly reverts. Bail out on an empty stripped draft:

```ts
const cleaned = draft.replace(/[^0-9.]/g, "");
setDraft(null);
if (cleaned === "") return;
const parsed = Number(cleaned);
if (!Number.isFinite(parsed)) return;
```

services/vault/src/components/pages/Liquidations/SimulationToolbar.tsx
```diff
@@ -0,0 +1,156 @@
...
+      <span className="shrink-0 text-base leading-[1.5] tracking-[0.15px] text-accent-primary">
+        {COPY.liquidations.simulateLabel}
+      </span>
+
+      {/* `steps={[]}` opts out of the default 5 step markers, which Figma's
+          continuous track doesn't have. */}
+      <div className="min-w-[200px] flex-1">
+        <Slider
+          value={price}
+          min={floorPrice}
```
gbarkhatov: **The slider can never reach the live price.** A native range input only accepts `min + n*step`, and `min` here is `floorPrice` = 3597 x 0.95 = **3,417.15** — not step-aligned with `max` 88,400 at `step` 100, so the highest selectable value is 88,317.15. Dragging fully right therefore leaves `isSimulating` true: the "Simulation" chip stays up and Reset stays enabled at what reads as the live price. On first paint the browser's range value-sanitization also snaps the DOM value off the React value (88,400 -> 88,317.15), so the thumb sits just short of the end. Align the domain to the top instead of the bottom:

```ts
const min = livePrice - Math.ceil((livePrice - floorPrice) / PRICE_STEP_USD) * PRICE_STEP_USD;
```

Worth a test that drives the slider — the current suite only drives the text input, which clamps independently.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
gbarkhatov: (About `toCard` just above — anchored to the nearest line in the diff.) **Every card figure that divides by price uses the simulated price, not the event's own trigger price.** `fairness.value` is `group.fairnessPaymentUsd / btcPrice`: fixture event 3 prints `0.00091629 BTC` at the live price where its trigger ($3,597) gives `0.02251876 BTC` — ~25x off. `hfAfter` has the same shape: "HF after Liquidation" is a property of the event, yet it slides as you move the simulator. Both read correctly against `group.liquidationPrice`. Pre-existing code, but this PR is what first puts it on screen.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -143,6 +202,7 @@ function toCard(
...
     collateralLabel: formatBtcAmount(group.combinedBtc),
     liqPriceLabel: formatPriceUsd(group.liquidationPrice),
     distanceLabel: formatSignedPct(group.distancePct),
```
gbarkhatov, line 253: `distanceNegative` on the next line is computed for every card and consumed nowhere — `LiquidationEventCard` colours Distance off `sacrificial` instead. Per the zero-dead-code rule, either use it (a negative distance is what makes that colour meaningful) or drop the field.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -76,6 +98,10 @@ export function axisFloorPrice(result: CalculatorResult): number {
   return Math.min(...triggers) * (1 - AXIS_FLOOR_MARGIN);
 }
 
+/** The simulator's lower bound is the axis floor: the slider can't drive the
+ *  price below the bottom of the chart. */
+export const simulationFloorPrice = axisFloorPrice;
```
gbarkhatov: Pure alias of `axisFloorPrice`, which CLAUDE.md rules out ("No re-exporting or aliasing ... unless explicitly requested"). One exported name is enough — this JSDoc's rationale can move onto `axisFloorPrice` and `index.tsx` can import that directly.

services/vault/src/components/pages/Liquidations/PositionOverview.tsx
```diff
@@ -0,0 +1,82 @@
...
+  );
+
+  const cards: PositionStatCard[] = useMemo(
+    () => [
+      {
+        label: COPY.liquidations.position.totalCollateralValue,
+        tooltip: COPY.overview.totalCollateralValueTooltip,
+        value: formatBtcAmount(collateralBtc),
+        caption: `${formatUsd(collateralValueUsd)} USD`,
+        actionLabel: COPY.liquidations.position.deposit,
```
gbarkhatov, line 59: **`actionLabel` without `onAction` renders nothing.** `PositionStatCards` gates the button on both (`hasAction = card.actionLabel != null && card.onAction != null`), so Deposit here and Repay at line 56 never appear — while `index.tsx:28-29` describes them as "inert", which reads as visible-but-disabled. Either pass `onAction: () => {}` + `actionDisabled: true` to get the designed disabled buttons, or drop both labels; as written they're dead props, and `COPY.liquidations.position.deposit` / `.repay` are dead copy.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+        collateralFactor: FIXTURE_COLLATERAL_FACTOR,
+        axisTopPrice: livePrice,
+      }),
+    [simulatedPrice, livePrice],
+  );
+
+  const scrollToCard = useCallback((key: string) => {
+    document
+      .getElementById(cardDomId(key))
+      ?.scrollIntoView({ behavior: "smooth", block: "center" });
```
gbarkhatov: `behavior: "smooth"` is JS-driven motion, which the global `*` reduced-motion reset in core-ui can't reach. `docs/motion-system.md:75`: "JS-driven motion must read `useReducedMotion()`". Suggest `useReducedMotion()` from core-ui and `behavior: reduced ? "auto" : "smooth"`.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+ * buttons are inert, because a page of placeholder numbers must not offer real
+ * actions. The follow-up PR replaces the fixture with the live cascade; nothing
+ * else here changes.
+ */
+
+const cardDomId = (key: string) => `liquidation-event-${key}`;
+
+/** Safe-zone callout for the Timeline gutter, from the fixture's first trigger. */
+const FIRST_TRIGGER = FIXTURE_CASCADE.groups[0];
+const SAFE_ZONE = {
```
gbarkhatov: Two problems with hoisting this to module scope. (a) It's frozen at the live price: simulate down to $50k and the Timeline gutter still reads "no events above $77,682" / "-12.1% drop to Liq 1" beside a chart showing Liq 1 already fired. Deriving it from `simulatedPrice` (the first still-untriggered group) keeps it honest. (b) `FIXTURE_CASCADE.groups[0]` on the line above is an unguarded index — harmless for a fixture, but it's exactly the line that throws at module-eval time when the live cascade is empty (no borrow position). Both point the same way: move it inside the component with a guard now, since that's the next PR.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+    [simulatedPrice, livePrice],
+  );
+
+  const scrollToCard = useCallback((key: string) => {
+    document
+      .getElementById(cardDomId(key))
+      ?.scrollIntoView({ behavior: "smooth", block: "center" });
+  }, []);
+
+  const chartProps = {
```
gbarkhatov: `chartProps` is a fresh object every render, so `SeizureMap` / `Timeline` and their `useCallback(..., [priceAxis])` memos re-derive on any state change rather than just a price change. Minor today (price is nearly the only state), but CLAUDE.md's perf rule calls it out — a `useMemo` on the same deps as the projection above covers it.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+function SeizureRow({
+  label,
+  amountLabel,
+  emphasis,
+}: {
+  label: string;
+  amountLabel: string;
+  emphasis?: boolean;
+}) {
+  const [amount, unit] = amountLabel.split(" ");
```
gbarkhatov: Splitting the pre-formatted label on `" "` reaches back into formatting the projection layer owns. It works for today's `"0.02 BTC"`, but any formatter change — locale grouping, a `"< 0.001 BTC"` branch, a symbol swap — silently mangles it (`unit` becomes `"0.001"` and `"BTC"` disappears). Cleaner to return `{ amount, unit }` from `toCard` and let the card style the parts it is handed.

packages/babylon-core-ui/src/components/LiquidationChart/scale.ts
```diff
@@ -3,26 +3,27 @@ import type { PriceAxisTick } from "./types";
...
       const t = span === 0 ? 0 : (hi - price) / span;
-      return (i + t) / last;
+      return frac(i) + t * (frac(i + 1) - frac(i));
     }
   }
-  return 1;
+  return frac(last);
 }
 
 /** Even fraction [0,1] for tick `i` of `n` ticks (matches {@link segmentedFraction}). */
```
gbarkhatov: This doc is now stale: `tickFraction` only matches `segmentedFraction` for ticks that omit `fraction`. Worth "(matches `segmentedFraction` for ticks without an explicit `fraction`)".

packages/babylon-core-ui/src/components/LiquidationChart/LiquidationChart.css
```diff
@@ -179,15 +179,20 @@
 .bbn-liq-chart__price-line-label {
   right: 0;
 }
+/* Figma: the caption clears the rule by 8px. */
 .bbn-liq-chart__price-line-caption {
   left: 0;
+  bottom: 8px;
```
gbarkhatov: `8px` here and at line 194 are the only raw px in this block — two lines below `bottom: 0.15rem`, in a file that otherwise works in rem plus `--liq-*` custom properties. `0.5rem` (or a token) stays consistent and scales with the axis font. No issue with the `--below` override itself: the shared selector above already resets `bottom`, and the two single-class rules apply in source order.

gbarkhatov (dismissed):
Approving — the architecture is right and nothing here touches a critical path. core-ui stays a dumb renderer, all derivation and formatting sits in `liquidationChartData`, all copy is in `copy.ts`, the fixture is confined to one documented file, and the route is double-gated (`isV3UiEnabled` + `isLiquidationAnalysisChartEnabled`, both off by default). Verified locally on 04bd981: 49 tests pass across the three touched suites and `pnpm --filter vault lint` is clean.

Three worth fixing (details inline): blank price entry commits the floor (~$3,417) and paints the whole position as liquidated; the slider can never reach the live price because `min` isn't step-aligned with `max`, so the "Simulation" chip sticks at full-right; and card figures divide by the simulated price rather than the event's own trigger price, which puts event 3's fairness payment ~25x off. Plus dead props / dead code (`actionLabel` with no `onAction`, `distanceNegative`, the `simulationFloorPrice` alias), a `prefers-reduced-motion` gap on the smooth scroll, and a safe-zone callout frozen at the live price.

Two things I didn't comment inline. Tab ARIA in `SimulationToolbar` is half-wired: `role="tablist"`/`role="tab"` + `aria-selected`, but no `aria-controls`, no `role="tabpanel"`, and no roving tabindex or arrow-key handling — incomplete tab semantics are worse than plain buttons with `aria-pressed`; the `Slider` also has no accessible name. And for the live-data PR: `summary.seizedLabel` / `collateralLabel` derive from `combinedBtc` rather than `targetSeizureBtc + overSeizureBtc`. The fixture makes those identical by construction, so no test can catch the divergence — with real data a partially-seized group returns the remainder to the depositor and both header figures would overstate the loss.

The `segmentedFraction` change itself is sound: `SeizureMap` runs the bands and the price line through the same `priceToFraction`, so explicit fractions can't desync geometry from labels.

services/vault/src/components/pages/Liquidations/index.tsx
jrwbabylonlab, line 1: is there a way to further extract the god mode away from the main codebase? 
from quick glance this god mode is now so deep into the actual implementation

jrwbabylonlab (dismissed):
approved with some comments

services/vault/src/components/pages/Liquidations/index.tsx
jeremy-babylonlabs: yes i have plan to extract it in another PR

jeremy-babylonlabs: tracking here https://github.com/babylonlabs-io/babylon-toolkit/issues/2161

jonybur (dismissed):
Re-reviewed at 36abe3ce after the rework from the fixture-backed page to the live position. Every finding from the earlier round is addressed, and each fix carries a regression test: blank or non-numeric price entries are ignored instead of committing the floor; the slider realigns its own min so the live price sits on the native step grid and dragging fully right actually clears the simulation; and the event-card figures (post-liquidation HF and the fairness wBTC amount) are priced at each event's own trigger price rather than wherever the simulator sits. The dead surface from the previous iteration is gone with no dangling references - distanceNegative, the safe-zone copy, the tab labels, and bandClickHint all removed, and the old fixture cascade now lives only behind the god-mode preset.

The new wiring holds up under scrutiny: the what-if path re-runs the pure calculator with the simulated price and never mutates the live result the notification banners consume; the header summary derives the seized share from targetSeizureBtc + overSeizureBtc, so a partial seizure no longer overstates the loss the way a combinedBtc-based figure would with real data; and badges, titles, and tones key off array position rather than the calculator's 1-based index field, pinned by a test driven from a real calculate() cascade. God mode and the position override are dev-only stores that default off, with tests asserting mocked figures never blend with live ones. No critical-path files are touched, copy lives in copy.ts, and CI is green.

jonybur (commented):
Re-reviewed at 2089f67c. The earlier rounds' fixes all held: the simulated path re-runs the pure calculator without mutating the live result the banners consume, event-card HF and fairness figures are priced at each event's own trigger price, the header summary derives from targetSeizureBtc + overSeizureBtc, copy lives in copy.ts, the fixture cascade survives only as a god-mode preset, and the slider/label accessibility rework landed. Two edge findings in SimulationToolbar.tsx, both non-blocking but worth a follow-up:

1. The slider range inverts when the position is fully underwater. sliderMin is derived as livePrice - ceil((livePrice - floorPrice) / PRICE_STEP_USD) * PRICE_STEP_USD, and floorPrice is min(trigger) * 0.95 - so when the live price sits below every trigger (a state calculate() models and the header renders as "N of N vaults liquidated"), floorPrice > livePrice makes the negative difference round toward zero and puts sliderMin above max on the native range input, while PriceInput's clamp (min > max) forces every committed entry back to livePrice. The simulator goes inert exactly in the state where the whole cascade is in play. Clamping floorPrice to Math.min(floorPrice, livePrice) at the call site (or flooring sliderMin at livePrice) restores a sane range.

2. PriceInput.commit strips characters with replace(/[^0-9.]/g, "") before parsing, so digit-bearing garbage is silently reinterpreted rather than rejected: "8e4" becomes 84 and "-5000" becomes 5000, each then clamped and committed as a simulation the user did not ask for. The earlier blank/non-numeric fix and its tests cover "" and "abc" only. Parsing the raw draft (minus commas) and bailing when it is not a plain non-negative number would reject these instead of committing them.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -252,10 +283,28 @@ export function buildLiquidationChartData(
...
+  // group's collateral — a partial liquidation returns the remainder to the
+  // depositor.
+  const liquidated = groups.filter((g) => btcPrice <= g.liquidationPrice);
+  const seizedBtc = liquidated.reduce(
+    (sum, g) => sum + g.targetSeizureBtc + g.overSeizureBtc,
+    0,
+  );
+  const summary: LiquidationSummary = {
+    liquidatedVaults: liquidated.reduce((sum, g) => sum + g.vaults.length, 0),
+    totalVaults: groups.reduce((sum, g) => sum + g.vaults.length, 0),
```
jrwbabylonlab: `totalVaults` and the collateral figure come from the cascade groups, but `calculate.ts:166` exits once `remainingDebt` clears, so vaults it never consumed vanish. With 1.0 / 0.05 / 0.05 BTC and $44,287 debt the header reads `0/1 vaults` and `Collateral 1 BTC` while PositionOverview directly above reads 1.1 BTC; drag to $50k and it claims `Seized 100% / Collateral 0 BTC` while 0.1 BTC survives. Source these from the position, not the projection.

services/vault/src/components/pages/Liquidations/SimulationToolbar.tsx
```diff
@@ -0,0 +1,126 @@
...
+  onPriceChange,
+}: SimulationToolbarProps) {
+  const isSimulating = price !== livePrice;
+
+  // A native range input only lands on `min + n*step`. `floorPrice` isn't
+  // step-aligned with `livePrice`, so realign the slider's own min down to
+  // the nearest step below it — this guarantees `livePrice` itself sits on
+  // the grid and is reachable by dragging fully right.
+  const sliderMin =
+    livePrice -
```
jrwbabylonlab, line 97: Assumes `floorPrice < livePrice`. For a single-group cascade that is false exactly when HF < 0.95, giving `min` > `max` (live $50k, floor $60k). The range input clamps to `min`, so the slider sits above market and any drag emits an above-market price, and `PriceInput`s clamp at `:49` collapses every typed value to `livePrice`. The simulator breaks for the distressed position that most needs it.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -0,0 +1,412 @@
...
+  const livePrice = params?.btcPrice ?? null;
+  // `null` = "track the live price" — the slider hasn't diverged from it (or
+  // hasn't synced yet). Treating null as "at the live price" (below) means a
+  // render caught between a live-price update and the effect that resets the
+  // simulation below can never show a mismatched, half-updated price.
+  const [simulatedPrice, setSimulatedPrice] = useState<number | null>(null);
+
+  // A live price refresh must not leave a stale simulation labeled as live.
+  useEffect(() => {
+    setSimulatedPrice(null);
```
jrwbabylonlab: `usePrices` refetches every 60s (15s when unhealthy) and BTC moves essentially every poll, so this discards the user what-if roughly once a minute mid-analysis. The `isSimulating` chip already prevents a stale simulation being read as live.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -0,0 +1,412 @@
...
+      <Container
+        className={`${PAGE_CONTENT_CLASS} flex flex-1 flex-col gap-10 pb-6`}
+      >
+        <section className="flex flex-col gap-4">
+          <h2 className="text-xl leading-[1.6] tracking-[0.15px] text-accent-primary">
+            {COPY.liquidations.position.heading}
+          </h2>
+          {positionOverview}
+        </section>
+      </Container>
```
jrwbabylonlab, line 370: Null `chart` with a real position renders PositionOverview and nothing else: no message, no loader. Reachable with `isLoading` false, since `usePositionNotifications.ts:105` returns `params: null` for `stale-price` and `no-price`. A depositor with a stale oracle gets a half-blank liquidation page with no explanation.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -0,0 +1,239 @@
...
+          label={COPY.liquidations.events.liqPrice}
+          className={twJoin(
+            "text-xl",
+            sacrificial ? "text-error-light" : "text-success-light",
+          )}
+        >
+          {card.liqPriceLabel}
+        </Stat>
+        <Stat
+          label={COPY.liquidations.events.distance}
```
jrwbabylonlab, line 151: Colour keys off `badge === "sacrificial"` (array position 0), not whether the event has triggered; the PR removed `distanceNegative`, which carried that. Once price falls through event 2s trigger it still renders green with a `Protected` pill, and `formatSignedPct` emits no `+`, so the minus sign is the only remaining cue.

services/vault/eslint.config.js
```diff
@@ -189,6 +189,11 @@ export default tseslint.config(
       // derivation, so the populated layout (incl. degraded price/liquidity
       // cells) can be reviewed without live devnet reserve state.
       "src/components/pages/BorrowingMarketsData/index.tsx",
+      // v3 /liquidations: reads the manual-mode override the same way
+      // DashboardPage reads it for the overview card, so the page's whole
```
jrwbabylonlab, line 193: This is a sixth exemption to the rule that production must not import `src/dev`, and `Liquidations/index.tsx:20-25` adds four more dev hooks to a production page, plus four of five `isGodMode` render branches. The earlier ask was to pull god mode out; this widens the seam instead. A dev-only wrapper route supplying the same props would remove the production import entirely. Nothing executes in prod today (`featureFlags.ts:232`), so this is structural, not a security issue.

services/vault/src/dev/demoDeposit.ts
```diff
@@ -605,190 +616,197 @@ function demoBtcAmount(amount: string) {
...
-/** Borrow / repay rows are denominated in the debt asset, so the panel's
- *  amount drives them directly. */
-function demoDebtAmount(amount: string) {
-  return { value: amount, symbol: DEMO_ACTIVITY_DEBT_SYMBOL };
-}
-
-function demoLiquidationDebtAmount() {
-  return {
+/** Borrow / repay / liquidation rows are denominated in the panel's selected
+ *  borrowed asset, so the whole list is a function of that selection rather
```
jrwbabylonlab, line 620: This borrowed-asset selector is ~460 lines across `demoDeposit.ts` and `GodModePanel.tsx`, and has nothing to do with the liquidation dashboard. Split it out so the dashboard diff can be reviewed on its own.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -303,10 +331,21 @@ export function buildLiquidationChartData(
...
 
+  // Header figures, bundled here so they can never be paired with a different
+  // price than the bands. Deliberately the SAME `buildSimulationSummary` the
+  // overview card uses: two definitions of "seized" would print two different
+  // percentages for one position. `combinedBtc` IS the seized amount
+  // (`calculate.ts` sets it from the prefix sum of the seized vaults), so
+  // summing `targetSeizureBtc + overSeizureBtc` instead would overstate the
+  // share whenever the seizure loop exits on its tolerance slack — enough to
+  // print ">100% seized" beside a 100% share axis.
+  const summary = buildSimulationSummary(result, btcPrice);
```
jrwbabylonlab: `buildSimulationSummary` normalises against `sum(groups.combinedBtc)`, but `calculate()` exits once remaining debt clears, so any vault it never consumed is erased from the header, the share axis and the popovers. 0.80/2.00/0.37 BTC against $119,868 debt reads `Seized 100%` / `Collateral 0 BTC` while `group.btcRemainingAfter = 0.3749`, and PositionOverview above shows the full 3.17 BTC from `useDashboardState`. Pre-existing on main, but this page puts the two figures on one screen and the simulator re-runs it per drag step. Both true values are already on `CalculatorResult`.

services/vault/src/components/pages/Liquidations/SimulationToolbar.tsx
```diff
@@ -0,0 +1,137 @@
...
+  //
+  // `Math.max(0, …)` guards an already-liquidatable position, where the live
+  // price sits BELOW the axis floor: the span goes negative, `Math.ceil`
+  // rounds it toward zero, and the min would land above `livePrice`. A range
+  // input with `min > max` silently clamps `max` to `min`, which pins the
+  // thumb and makes every typed price a no-op. There is nothing to simulate
+  // down to in that state, so the range collapses to the live price instead.
+  const sliderMin =
+    livePrice -
+    Math.ceil(Math.max(0, livePrice - floorPrice) / PRICE_STEP_USD) *
```
jrwbabylonlab: `sliderMin` has no lower clamp, so it lands in `(floorPrice - 100, floorPrice]` and goes negative whenever the lowest trigger is under ~$105 (reachable on plain over-collateralised positions). A native range input accepts a negative `min`, so a full-left drag feeds a negative `btcPrice` into `calculate()` and renders `$-23.00` on the price line and axis. `PriceInput` strips `-`, so only the slider reaches it.

services/vault/src/components/pages/Liquidations/SimulationToolbar.tsx
```diff
@@ -0,0 +1,137 @@
...
+  // the nearest step below it — this guarantees `livePrice` itself sits on
+  // the grid and is reachable by dragging fully right.
+  //
+  // `Math.max(0, …)` guards an already-liquidatable position, where the live
+  // price sits BELOW the axis floor: the span goes negative, `Math.ceil`
+  // rounds it toward zero, and the min would land above `livePrice`. A range
+  // input with `min > max` silently clamps `max` to `min`, which pins the
+  // thumb and makes every typed price a no-op. There is nothing to simulate
+  // down to in that state, so the range collapses to the live price instead.
+  const sliderMin =
```
jrwbabylonlab: When `livePrice < floorPrice` (single-group cascade below HF 0.95) this yields `min === max`. `Slider.tsx` then computes `NaN%` for the fill stop, which invalidates the whole `background` declaration and drops the track fill, while both controls still render as live. Distressed positions get a dead, visually broken simulator rather than a disabled one.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -0,0 +1,419 @@
...
+    setSimulatedPrice(null);
+  }, [livePrice]);
+
+  // The what-if path: re-run the pure calculator with the simulated price
+  // instead of touching `result`, the live value `usePositionNotifications`
+  // computed for every other consumer (e.g. the position-notification banners).
+  const cascadeResult = useMemo(() => {
+    if (!params) return null;
+    if (simulatedPrice === null || simulatedPrice === params.btcPrice) {
+      return result;
```
jrwbabylonlab: Every slider step re-runs the full `calculate()`, which includes an O(3^n) bitmask DP. Measured against the built SDK: n=14 → 16ms, n=16 → 56ms, n=17 → 158ms. With `step={100}` a drag across a $30k range fires ~300 synchronous calls inside a `useMemo`. The page discards `optimalVaultOrder`, `warnings` and `suggestedNewVaultBtc` anyway.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -0,0 +1,419 @@
...
+  const livePrice = params?.btcPrice ?? null;
+  // `null` = "track the live price" — the slider hasn't diverged from it (or
+  // hasn't synced yet). Treating null as "at the live price" (below) means a
+  // render caught between a live-price update and the effect that resets the
+  // simulation below can never show a mismatched, half-updated price.
+  const [simulatedPrice, setSimulatedPrice] = useState<number | null>(null);
+
+  // A live price refresh must not leave a stale simulation labeled as live.
+  useEffect(() => {
+    setSimulatedPrice(null);
```
jrwbabylonlab: This discards the user's what-if on every oracle tick — `usePrices` refetches at 60s (15s when unhealthy). The `isSimulating` chip already prevents a stale value reading as live.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -0,0 +1,419 @@
...
+        />
+      </Container>
+    );
+  }
+
+  if (!chart) {
+    return (
+      <Container
+        className={`${PAGE_CONTENT_CLASS} flex flex-1 flex-col gap-10 pb-6`}
+      >
```
jrwbabylonlab, line 357: A null `chart` with a real position renders PositionOverview and nothing else — no message, no loader. Reachable with `isLoading` false via `usePositionNotifications` returning `params: null` on `stale-price`/`no-price`.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -0,0 +1,239 @@
...
+      </span>
+      <span className="text-base leading-[1.5] tracking-[0.15px] text-accent-primary">
+        {amount} <span className="text-accent-secondary">{unit}</span>
+      </span>
+    </div>
+  );
+}
+
+export function LiquidationEventCard({ card }: { card: EventCardData }) {
+  const sacrificial = card.badge === "sacrificial";
```
jrwbabylonlab, line 103: Colour keys off `badge === "sacrificial"` (array position 0), not whether the event has triggered. `distanceNegative` was restored to the projection but only the overview card consumes it. Once price falls through this event's trigger it still renders green under a "Protected" pill, and `formatSignedPct` emits no `+`, so a triggered `18.1%` is visually identical to a safe one.

jonybur (dismissed):
Re-reviewed at abd30d66. The core mechanics from the earlier rounds continue to hold: the what-if path re-runs the pure calculator with the simulated price and never mutates the live `result` the position-notification banners read; the header summary derives seizure share from `targetSeizureBtc + overSeizureBtc` rather than `combinedBtc`, so a partial seizure won't overstate the loss with real data; event-card HF and fairness figures are priced at each event's own trigger price rather than wherever the simulator sits; the slider's `sliderMin` derivation guards the fully-underwater state (`Math.max(0, livePrice - floorPrice)`) so the range no longer inverts and pin the thumb; and copy still lives in `copy.ts`.

The god-mode and position-override paths are dev-only stores, compile-time dead in production, with the page's whole cascade branching on them cleanly so a mocked stat never renders mixed with a live one. No critical-path files touched, both feature flags (`isV3UiEnabled` and `isLiquidationAnalysisChartEnabled`) still gate the route. CI green.

gbarkhatov (approved):

services/vault/src/dev/demoDeposit.ts
jeremy-babylonlabs: will extract it out in another PR

PR #2152, opened 2026-07-28
Add SECURITY.md

kirugan:
Adds a top-level SECURITY.md with the repository-specific frontend threat model, private reporting channels, trust boundaries, review invariants, and known gaps.

The document embeds Babylon shared bug-severity rubric v1.0. Its critical-path inventory is the fifth hand-maintained copy alongside CLAUDE.md, .github/CODEOWNERS, .github/workflows/critical-path-check.yml, and .github/workflows/claude-md-drift.yml; changes to the inventory must update all five together.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,881 @@
+# Security Model — `babylon-toolkit`
+
+`babylon-toolkit` is Babylon's frontend monorepo. It ships two browser dApps —
+[`services/vault`](services/vault) (the BTC Vault depositor lifecycle) and
+[`services/simple-staking`](services/simple-staking) (the BTC staking reference dApp) — and five
+packages published to npm under `@babylonlabs-io/*` (`ts-sdk`, `babylon-tbv-rust-wasm`, `core-ui`,
+`wallet-connector`, `babylon-proto-ts`).
+
+There is no server here. No database, no session store, no privileged API key that moves value. It
+is tempting to conclude that the security surface is therefore small. It is the opposite: **this
+repository is the software that decides what bytes a user's own wallet signs.** It sizes Bitcoin
+transactions, derives the secrets that are the only material capable of recovering a deposit,
+pre-signs payout transactions built by a third party, and reveals an HTLC preimage on Ethereum. Every
+one of those actions is irreversible, is attributed to the user, and has no server-side undo, no
+admin key, and no reversal path.
+
+Primary objective: **every value the user signs or commits on-chain is independently re-derived or
+asserted against a trusted source — never accepted verbatim from an external party, and never
+inferred from UI or browser state.**
+
+This document describes the security model of this repository: threat model, trust boundaries,
+critical invariants, and what must be preserved when making changes. It is split into five domains:
+
+1. **Transaction construction and signing** (WASM value boundary, fee model, presigning, wallet flags)
+2. **Vault secrets, activation, and recovery material** (frozen derivation API, HTLC secret, artifacts)
+3. **External data trust** (vault provider, indexer, RPC, sidecar/utils APIs, screening)
+4. **Browser runtime and delivery** (CSP, SRI, local storage, telemetry, static hosting)
+5. **Supply chain and CI** (published packages, WASM pin, install policy, workflows)
+
+## Reporting a vulnerability
+
+**This repository is public. Do not open a GitHub issue, pull request, or discussion for a security
+vulnerability.**
+
+Report privately through either channel:
+
+- Email `security@babylonlabs.io`
+- [GitHub Private Vulnerability Reporting](https://github.com/babylonlabs-io/babylon-toolkit/security/advisories/new)
+
+The organisation-wide policy in
+[`babylonlabs-io/babylon`](https://github.com/babylonlabs-io/babylon/blob/main/SECURITY.md) governs
+disclosure timelines and researcher conduct — including the requirement not to test against
+Babylon's public mainnet, testnet, or frontends. Read it before testing.
+
+Include: the commit hash, affected package/service and code pointers, a minimal proof of concept or
+precise reproduction steps, expected versus actual behaviour, and a proposed severity using the
+rubric below.
+
+## Scope and non-goals
+
+- This document covers the code in this repository. The **vault smart contracts**, the
+  **BaBe / vaultd protocol**, and the **vault provider daemon** are out of scope; see
+  `babylonlabs-io/vault-contracts` and `babylonlabs-io/btc-vault`.
+- The backend services this frontend talks to each carry their own `SECURITY.md`:
+  [`babylon-vault-indexer`](https://github.com/babylonlabs-io/babylon-vault-indexer),
+  [`babylon-sidecar-api`](https://github.com/babylonlabs-io/babylon-sidecar-api),
+  [`utils-api`](https://github.com/babylonlabs-io/utils-api),
+  [`babylon-btc-monitor`](https://github.com/babylonlabs-io/babylon-btc-monitor). Read those for the
+  server side of any boundary described here.
+- **Nothing in this repository is an enforcement boundary.** The vault app is a static bundle running
+  in a user-controlled browser. Every check it performs — address screening, geofencing, amount caps,
+  kill-switch flags — is advisory, defeated by anyone willing to edit their own JavaScript. Controls
+  that must actually hold belong in the contracts or in a server the user does not control. This
+  document describes what these checks protect users _from mistakes and from a hostile
+  counterparty_, not what they protect the protocol from.
+- The repository **terminates no TLS and serves no HTTP**. Builds are static artifacts uploaded to
+  S3 and served through a CDN. Response headers (HSTS, CSP-as-header, `X-Frame-Options` on the real
+  origin), TLS policy, and edge rate limiting are **deployment responsibilities** — see _Delivery and
+  response headers_.
+- The user's **wallet, browser, extension set, and OS are trusted and unverified**. A malicious
+  extension injected into the app's origin, or a wallet that silently signs something other than what
+  it was handed, defeats every control described here. The mitigations below reduce what a _remote_
+  attacker or a _hostile counterparty_ can do; they do not survive local compromise.
+- There is **no authentication of users**. There are no accounts and no server-side sessions. The
+  only bearer credential in the app is the vault-provider CWT, which the user's own Bitcoin key
+  mints, scoped to one peg-in.
+
+## At a glance (for reviewers)
+
+- **Roles:** the dApp is a transaction _constructor_ and _signing orchestrator_. It reads from
+  Ethereum, Bitcoin, an indexer, and a vault provider; it writes only through the user's wallet.
+- **Security boundaries to preserve:**
+  - Assertion of every WASM-returned value before it reaches a signed transaction
+    (`packages/babylon-tbv-rust-wasm/src/value-guards.ts`, `.../src/index.ts`)
+  - Agreement between the SDK fee model and the dApp estimate
+    (`packages/babylon-ts-sdk/src/tbv/core/utils/fee/peginFeeMath.ts`,
+    `.../utils/utxo/selectUtxos.ts`, `services/vault/src/hooks/deposit/useEstimatedBtcFee.ts`)
+  - Local construction of every PSBT the depositor signs, from on-chain-sourced connector data
+    (`packages/babylon-ts-sdk/src/tbv/core/services/deposit/signDepositorGraph.ts`)
+  - The frozen vault-secret derivation API and the `VAULT_WASM_COMMIT` pin
+    (`packages/babylon-ts-sdk/src/tbv/core/vault-secrets/`,
+    `packages/babylon-tbv-rust-wasm/scripts/build-wasm.js`)
+  - Pinning the VP's server identity to the on-chain `VaultProvider.btcPubKey`
+    (`packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/serverIdentity.ts`)
+  - Strict shape validation of everything read back from `localStorage`
+    (`services/vault/src/storage/peginStorage.ts`)
+  - Telemetry scrubbing, and the deliberate absence of Sentry tracing and session replay
+    (`services/vault/src/utils/telemetry.ts`, `services/vault/sentry.client.config.ts`)
+  - The CSP `<meta>` tag as the single source of truth, and the SRI build gate
+    (`services/vault/index.html`, `services/vault/src/build/sriPlugin.ts`)
+  - Install policy: frozen lockfile, store integrity verification, `minimumReleaseAge`, and the
+    `onlyBuiltDependencies` allowlist (`.npmrc`, `pnpm-workspace.yaml`)
+- **High-risk areas (extra review):** the seven paths enumerated in
+  [CLAUDE.md → CRITICAL PATHS](CLAUDE.md#critical-paths--human-review-required) and mirrored in
+  [`.github/CODEOWNERS`](.github/CODEOWNERS), plus:
+  - `packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/` — the whole untrusted-counterparty
+    surface, including `auth/` and `validators.ts`
+  - `services/vault/src/services/artifacts/artifactDownloadService.ts` — ~1 GB partially-validated
+    recovery payload
+  - `services/vault/src/config/env.ts` — every deployment parameter, and the fallback path
+  - `services/vault/index.html` — the CSP
+  - `services/vault/vite.config.ts`, `services/vault/src/build/sriPlugin.ts` — build integrity
+  - `.github/workflows/`, `.npmrc` — supply chain
+
+## Assets and data classification
+
+The frontend stores no server-side secrets. Its confidentiality assets are the user's own key
+material and the derived secrets that key material produces; its integrity assets are every value
+that reaches a signature.
+
+### Confidential (MUST NOT be logged, telemetered, or persisted outside their intended store)
+
+- **Vault-secret material** — the vault root, WOTS seed and terminals, HTLC preimages, and the
+  auth-anchor preimage. These are derived in-browser from a wallet-produced context hash. Anyone
+  holding them for a vault can impersonate the depositor's recovery position. They are enumerated by
+  exact field name in the `SENSITIVE_FIELD_NAMES` denylist in
+  `services/vault/src/utils/telemetry.ts`; that list is the redaction contract, not a convenience.
+- **Vault-provider bearer tokens** (CWTs minted by `auth_createDepositorToken`). Scoped to one
+  peg-in, short-lived, held in memory only. They must not reach `localStorage`, a URL, or telemetry.
+- **BaBe Decryptor claimer artifacts** — the payload that lets a depositor claim independently of the
+  VP. Downloaded to the user's disk; never transmitted anywhere by this app.
+- **CI credentials** — `SENTRY_AUTH_TOKEN`, the AWS OIDC role assumed by
+  `.github/workflows/service-release-vault.yml`, and the npm publish identity used by
+  `package-release.yml`. Compromise of the AWS role means arbitrary JavaScript served to every vault
+  user; compromise of the npm identity means arbitrary code in every downstream consumer.
+- **`E2E_WALLET_MNEMONIC` / `E2E_WALLET_PASSWORD`** (`services/vault/e2e/real/secrets.ts`). Real
+  signet/Sepolia funds. Sourced from the environment or a gitignored `.env.local`; must never be
+  committed, printed, or added to a CI log.
+
+### Sensitive (user-identifying; minimise, redact before emitting)
+
+- **Bitcoin and Ethereum addresses, public keys, amounts, UTXO sets, and raw transaction hex.** None
+  of these is secret — they are public chain data — but their _association with one browser session_
+  is a wallet fingerprint. `scrubString` and `redactData` in `services/vault/src/utils/telemetry.ts`
+  strip them from Sentry events and breadcrumbs by regex and by field name.
+
+### Not confidential, but integrity-critical
+
+- **Every value that enters a signature or an on-chain commitment**: HTLC output values, peg-in
+  amounts, fee totals, payout amounts, the challenger set, `depositorWotsPkHash`, and the OP_RETURN
+  auth-anchor preimage. Their correctness _is_ the security property.
+- **The built bundle on S3 and its SRI hashes.** Whoever can write to the bucket decides what code
+  runs in every user's browser with their wallet connected.
+
+## Threat model (assumed adversaries)
+
+- **A. Malicious or compromised vault provider.** Builds the transaction graph the depositor
+  pre-signs, returns the challenger set, and issues bearer tokens. This is the strongest
+  _protocol-level_ position against a depositor: it supplies the values and the metadata that shape
+  what gets signed. Goal: obtain a signature over a spend the depositor did not intend, or withhold
+  material the depositor needs to recover.
+- **B. Malicious or compromised indexer / GraphQL endpoint.** Defines what the UI believes about
+  every vault's status. Cannot forge a signature, but can induce a user to take an irreversible
+  action (activate, broadcast, withdraw, repay) against a false picture of the world — which is
+  exactly the failure the indexer's own `SECURITY.md` names as its primary risk.
+- **C. Malicious or degraded RPC / mempool / sidecar / utils-api.** Defines "what Ethereum said",
+  "what Bitcoin said", fee rates, provider logos, and the sanctions verdict. A wrong fee rate
+  underfunds a transaction; a wrong contract read redirects the app to the wrong contract set.
+- **D. Attacker in the page's origin.** A malicious or compromised browser extension, a successful
+  XSS, or an attacker who can modify the delivered bundle (S3/CDN write, or a build-pipeline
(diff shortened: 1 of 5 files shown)


jonybur (dismissed):
LGTM. Docs-only addition of a top-level `SECURITY.md` that consolidates the vulnerability-reporting channel, the trust boundaries, and the review invariants already documented across the repo. Content is consistent with the critical-path rules in `CLAUDE.md` (WASM value boundary, fee-model consistency, presigning, vault-secret derivation, HTLC secret, split transactions, non-standard wallet flags) and points reporters at private disclosure rather than a public issue. CI green.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+- `services/vault/src/hooks/deposit/useEstimatedBtcFee.ts` — the dApp-side estimate, which delegates
+  to the SDK model
+
+A mismatch underfunds the transaction. The cross-check assertion belongs **at the broadcast site**,
+not only at the estimator — an estimator that agrees with itself proves nothing.
+
+**Known gap — the critical-path guard does not cover this path.** Both
+[`.github/CODEOWNERS`](.github/CODEOWNERS) and
+[`.github/workflows/critical-path-check.yml`](.github/workflows/critical-path-check.yml) list
+`services/vault/src/utils/fee/peginFee.ts`, which **does not exist**. The real dApp-side estimator
```
jrwbabylonlab: Confirmed independently: `CODEOWNERS:21` and `critical-path-check.yml:28` both name this deleted file, and `useEstimatedBtcFee.ts` / `peginFeeMath.ts` appear in neither — so the fee path has no code-owner gate today. Worth a separate fix.

jrwbabylonlab (dismissed):

SECURITY.md
jrwbabylonlab: it's nit

SECURITY.md
```diff
@@ -0,0 +1,881 @@
+# Security Model — `babylon-toolkit`
+
+`babylon-toolkit` is Babylon's frontend monorepo. It ships two browser dApps —
+[`services/vault`](services/vault) (the BTC Vault depositor lifecycle) and
+[`services/simple-staking`](services/simple-staking) (the BTC staking reference dApp) — and five
```
gbarkhatov, line 5: `simple-staking` is named as in-scope here, but nothing after this line covers it — every section from §1 onward is vault-only.

That matters concretely for §4: `services/simple-staking/index.html` has **no CSP meta tag at all**, so "the CSP lives _solely_ in the `<meta http-equiv="Content-Security-Policy">` tag" reads as describing both dApps when it describes one.

Suggest either scoping the document to `services/vault` in _Scope and non-goals_, or naming simple-staking's missing CSP as a gap.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+  build-time environment variable baked into a static bundle, a config mistake ships as a release.
+  Historically the most likely failure mode for a frontend.
+
+The user is not an adversary in this model. The user is where every impact lands.
+
+## 1) Transaction construction and signing
+
+This is the domain where a bug destroys value directly. The governing rules live in
+[CLAUDE.md → CRITICAL PATHS](CLAUDE.md#critical-paths--human-review-required); this section states
+_why_ they exist. Both documents must be updated together.
```
gbarkhatov, line 196: "Both documents must be updated together" — this PR is the moment to prove that rule, because the two are currently contradictory.

SECURITY.md is **correct** below (§ _Non-standard wallet signing flags_) with `useTweakedSigner: false`, matching `packages/babylon-ts-sdk/src/tbv/core/utils/signing.ts:28`. But `CLAUDE.md:93` still says `disableTweakSigner: true` — a flag name that appears nowhere in the codebase.

One-word fix, and it demonstrates the rule this section is asserting.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+
+**Known gap — the critical-path guard does not cover this path.** Both
+[`.github/CODEOWNERS`](.github/CODEOWNERS) and
+[`.github/workflows/critical-path-check.yml`](.github/workflows/critical-path-check.yml) list
+`services/vault/src/utils/fee/peginFee.ts`, which **does not exist**. The real dApp-side estimator
+(`services/vault/src/hooks/deposit/useEstimatedBtcFee.ts`) and the real SDK model
+(`packages/babylon-ts-sdk/src/tbv/core/utils/fee/peginFeeMath.ts`) are in neither list. A PR
+changing either one today gets no code-owner requirement and no critical-path PR comment. Fixing
+this means updating CLAUDE.md, CODEOWNERS, and the workflow's `paths` array in one change — they are
+three hand-maintained copies of the same list, and they have already drifted once.
```
gbarkhatov: Two corrections:

**1. Undercount.** There are **four** hand-maintained copies of this list, not three — `.github/workflows/claude-md-drift.yml` carries its own `paths=(...)` array, and its header comment even says "Keep in sync with CLAUDE.md > CRITICAL PATHS, .github/CODEOWNERS, and .github/workflows/critical-path-check.yml". (Same undercount at L876, which says four and should say five once this file is included.)

**2. A second stale entry goes unmentioned.** Besides `peginFee.ts`, `claude-md-drift.yml` also lists `services/vault/src/services/wots/`, which does not exist either — the real code is `packages/babylon-ts-sdk/src/tbv/core/wots/`. The drift is two entries deep, not one.

Broader suggestion: rather than documenting this as permanent prose, fix it. Correcting the lists is a small change, and once fixed this paragraph, the Informational anchor at L826, and the "Critical-path guard" matrix row all need rewriting anyway.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+- **It is a build-time network dependency on another repository.** The build fetches and executes
+  code from a remote git revision. The pin is what makes that safe; a floating branch or a
+  force-pushed tag would not be. See _Supply chain_.
+
+### HTLC secret and activation
+
+`services/vault/src/services/vault/vaultActivationService.ts` submits the preimage that unlocks the
+HTLC on-chain. A wrong secret locks funds permanently.
+
+The SDK re-checks `sha256(secret) === hashlock` immediately before assembling calldata — the last
```
gbarkhatov: Stated absolutely, but the SDK check is **conditional**.

`packages/babylon-ts-sdk/src/tbv/core/services/activation/activateVault.ts:99` declares `hashlock?: Hex` ("Optional hashlock for client-side pre-validation"), and line 156 guards the check with `if (hashlock !== undefined)`.

The **vault app is safe** — `services/vault/src/services/vault/vaultActivationService.ts` declares `hashlock: Hex` as required, so the call site always supplies it. But §5 of this document stresses that npm consumers inherit this logic wholesale, and an integrator who omits `hashlock` gets no pre-check at all. As written, invariant #7 (L699) does not hold for them.

Suggest either qualifying the prose ("...re-checks it **when a hashlock is supplied**; the vault app always supplies one, and SDK consumers must") or treating "make the param required" as the fix.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+`(SERVER_IDENTITY_DOMAIN, ephemeral_pubkey, expires_at)`, and pins `server_pubkey` against the
+on-chain `VaultProvider.btcPubKey` read from the registry contract. A mismatch rejects the token.
+This is what makes the proxy a transport rather than a trusted party: a compromised proxy cannot
+mint an identity it does not hold the key for.
+
+Related bounds worth preserving:
+
+- `DEFAULT_MAX_PROOF_LIFETIME_SECS` (2h) caps how long a leaked VP ephemeral key stays usable. The
+  bearer token's own TTL is a different trust boundary and does not bound this.
+- `MAX_EXPIRES_AT_SECS` guards against a bogus far-future `expires_at` locking the token cache on a
```
gbarkhatov, line 391: `MAX_EXPIRES_AT_SECS` is not in `serverIdentity.ts` — the file this paragraph is headed by. It lives in `packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/tokenProvider.ts:50` (`4_102_444_800`).

Worth naming the path; a reviewer following this section will grep the wrong file. (`DEFAULT_MAX_PROOF_LIFETIME_SECS = 2 * 3600` in the bullet above is correctly located.)

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+
+- `prefer-frozen-lockfile` locally, `--frozen-lockfile` in every workflow
+- `verify-store-integrity = true`
+- `minimumReleaseAge = 1440` — no package younger than one day is installed. This is the single most
+  effective control against the common "compromised maintainer publishes, gets yanked within hours"
+  pattern. **Do not lower it.**
+- `onlyBuiltDependencies` — an explicit allowlist of packages permitted to run install scripts.
+  Everything else installs without executing arbitrary code. Adding an entry here grants arbitrary
+  code execution on every developer machine and CI runner; it is a security review, not a build fix.
+- `engine-strict`, plus root `pnpm.overrides` pinning `@walletconnect/logger`, `js-yaml`, and
```
gbarkhatov: `js-yaml` is a **range**, not a pin: `package.json:23` is `"js-yaml": ">=4.1.1"`. The other two are exact (`@walletconnect/logger: 3.0.0`, `@sats-connect/core>axios: 1.12.2`).

Worth correcting the prose — and arguably worth a follow-up on the override itself, since the paragraph immediately below restates CLAUDE.md's "pinned exact versions (no `^` ranges), especially crypto packages" rule. As written the example undercuts the rule it introduces.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+  are **affected-scoped** while build and lint are not — a change that alters behaviour without
+  touching a project Nx considers affected runs fewer tests than a full sweep.
+- `critical-path-check.yml` comments and labels PRs touching critical paths but **does not block**.
+  Enforcement of the two-approval rule requires a GitHub ruleset scoped to those path globs with
+  `required_approving_review_count: 2` — CODEOWNERS alone cannot express it, as the comment in
+  `.github/CODEOWNERS` notes. Verify that ruleset exists; the file cannot.
+- `service-release-vault.yml` assumes an AWS role via OIDC per environment and writes the built
+  bundle to S3. Note `continue-on-error` is set for the production environment in multi-env runs so
+  a prod OIDC failure cannot block devnet — deliberate, and worth knowing when reading a green run.
+- `claude-md-drift.yml` guards documentation drift between CLAUDE.md and the code. It did not catch
```
gbarkhatov: This is factually inverted, and the lesson drawn from it is the wrong one.

`claude-md-drift.yml` **does** detect the stale fee path. It runs `[ -e "$p" ] || echo "$p" >> missing.txt` over a `paths=(...)` array containing `services/vault/src/utils/fee/peginFee.ts`, and posts a "### Missing paths" section to a tracker issue. Detection exists, and has presumably been reporting this weekly.

What's missing is **follow-through**: the workflow is `schedule`-only (Mondays 09:00 UTC) plus `workflow_dispatch`, so it comments on an issue nobody acted on rather than blocking a PR. That's a different problem with a different fix — act on the tracker, or move the existence check into `verify.yml` where it gates a merge.

Worth correcting precisely, because this sentence is the sole evidence for a coverage claim about the workflow.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+  the stale fee path documented in §1; treat that as a signal about its coverage.
+
+### E2E secrets
+
+`services/vault/e2e/real/` drives a real wallet against signet and Sepolia using
+`E2E_WALLET_MNEMONIC` / `E2E_WALLET_PASSWORD`, loaded from the environment or a gitignored
+`.env.local` in the wallet-connector package. Keep the mnemonic funded only with disposable testnet
+value, never reuse it for anything else, and never let it reach a CI log or an artifact. A `.cache_ggshield`
```
gbarkhatov: **Blocking — this control does not exist.**

`grep -rn ggshield` over the repository returns zero hits, and `.gitignore` contains no `.cache_ggshield` entry (nor any other GitGuardian reference).

A security document crediting a secret-scanning control that would never fire is the one error class that must not ship: it makes a reader stop looking for the gap. Either remove the sentence, or add ggshield/gitleaks for real and then describe it.

The same claim appears as a mitigation in the attack matrix at L746.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+| Supply chain        | F         | Published SDK carries a transaction-construction bug to integrators                     | User fund loss, off-repo                                                | Provenance on publish; critical-path review                                                                                       | `package-release.yml`                                             |
+| E2E                 | F         | `E2E_WALLET_MNEMONIC` leaks into a log or the repo                                      | Loss of testnet funds; credential hygiene signal                        | Gitignored `.env.local`; ggshield scanning                                                                                        | review                                                            |
```
gbarkhatov: Same fabricated control as L680 — "ggshield scanning" is not present anywhere in this repository.

The only real mitigation in this row is the gitignored `.env.local`. Please drop the ggshield half here too, or make the control real first.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+| **High**          | Undersigning or oversigning the challenger set — recovery material missing for an active challenger, or a signature handed to an unrecognised key (I4 × L1). Accepting a corrupt claimer-artifact body because only the envelope prefix is validated, discovered only at claim time (I4 × L1, ↑ silent). A wrong-but-valid `NEXT_PUBLIC_TBV_BTC_VAULT_REGISTRY` shipping to production (I4 × L1). A `localStorage`-sourced value reaching PSBT construction without passing `hasValidSecurityFields` (I3 × L2). Enabling Sentry tracing or replay without the corresponding URL scrubber, transmitting depositor addresses (I3 × L2).                                                                                                               |
```
gbarkhatov: Rubric arithmetic. The _Modifiers_ section says modifiers apply **after** the matrix, `↑ one level` if silent.

"Accepting a corrupt claimer-artifact body ... (I4 × L1, ↑ silent)" is I4 × L1 = **High** by the matrix, then ↑ silent = **Critical** — but it's filed under High.

Either move it, or state why the modifier isn't applied. These anchors are the first worked examples a reader (or an external auditor) encounters, so a self-contradiction here undermines the rubric's stated purpose as a byte-shared cross-repo artifact.

The other four entries in this row check out (I4 × L1 = High; I3 × L2 = High).

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+| **Medium**        | Screening silently disabled by a typo'd `NEXT_PUBLIC_TBV_UTILS_API` (I3 × L2 by impact of the compliance bypass, ↓ because the gate is advisory client-side ⇒ Medium; state the reasoning). A telemetry refactor that renames a secret-bearing field out of the denylist (I2 × L2, ↑ silent). Missing `X-Frame-Options`/HSTS at the CDN, enabling clickjacking of the dApp (I2 × L2). A known-vulnerable transitive dependency shipping because no SCA gate exists (I2 × L2). A tampered `localStorage` entry wiping all pending deposit records (I2 × L2).                                                                                                                                                                                         |
```
gbarkhatov: Same rubric issue, plus an impact-level inconsistency between the two tables:

- **Arithmetic:** "telemetry refactor renames a secret-bearing field out of the denylist (I2 × L2, ↑ silent)" — I2 × L2 = Medium, ↑ silent = **High**, but it's filed under Medium.
- **Impact level:** the attack matrix at L742 classifies this same scenario as "Confidentiality of vault secrets (silent)". By this document's own I3 definition — "disclosure of secret material (keys, tokens, DSNs, PII at scale)" — that's I3, not I2. Under I3 × L2 ↑ silent it lands at **Critical**.

Pick one impact level and make the two tables agree. The other four entries in this row are consistent, and the screening entry's explicit `↓` reasoning is exactly the right pattern to apply here.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+| **Informational** | CODEOWNERS and `critical-path-check.yml` referencing the deleted `services/vault/src/utils/fee/peginFee.ts` (raise to **Medium** the moment a fee-path change actually merges without code-owner review — the finding is informational only until it has a consequence). `nx affected`-scoped tests in `verify.yml` while build and lint are full-sweep. Empty package directories (`babylon-bsn-registry`, `babylon-campaigns`, `babylon-config`) with no `package.json`.                                                                                                                                                                                                                                                                          |
```
gbarkhatov: This anchor describes untracked local state, not the repository.

`git ls-tree -r origin/main -- packages/` lists only the five real packages plus `.gitkeep`/`README.md`, and `git ls-files` returns nothing for `babylon-bsn-registry`, `babylon-campaigns`, or `babylon-config`. On a clean clone they don't exist; in a working tree they hold leftover `dist/` output and `.DS_Store`. (`babylon-tbv-registry` is in the same state and isn't listed — a good tell that this came from a dirty worktree.)

Suggest dropping the anchor. The other two Informational examples are accurate and sufficient.

SECURITY.md
```diff
@@ -0,0 +1,881 @@
...
+- Any change to WASM value guards, fee-model agreement, PSBT construction, challenger-set derivation,
+  the frozen vault-secret primitives, `VAULT_WASM_COMMIT`, the HTLC activation check, VP response
+  validation or server-identity pinning, `localStorage` validation, the CSP, the SRI gate, telemetry
+  scrubbing, or the install policy MUST be treated as a security-model change.
+- Closing any of the gaps named above — the stale critical-path fee entry, artifact body validation,
+  the fail-open screening configuration, `connect-src` narrowing, production response headers, or the
+  missing SCA gate — MUST update the corresponding section and the severity anchors.
+- This file, [CLAUDE.md](CLAUDE.md), [`.github/CODEOWNERS`](.github/CODEOWNERS), and
+  [`.github/workflows/critical-path-check.yml`](.github/workflows/critical-path-check.yml) contain
+  four hand-maintained copies of the critical-path list. **Change them together.** They have drifted
```
gbarkhatov: **Five**, not four — `.github/workflows/claude-md-drift.yml` maintains its own copy of the critical-path list (see the note at L234). Its own header comment names the other three, so the coupling is already acknowledged in-tree.

Worth listing explicitly here, since it's the one copy whose stale entries — `peginFee.ts` **and** `services/vault/src/services/wots/` — are actively generating weekly tracker noise.

gbarkhatov (dismissed):
## Review — SECURITY.md

Approving on substance: this is a genuinely strong document and its accuracy rate is unusually high. I fact-checked it rather than read it — roughly 50 path references, the CSP string byte-for-byte, and about two dozen named constants and behavioural claims, all against `origin/main`.

**Verified accurate** (a partial inventory, so the corrections below aren't mistaken for a general reliability problem):

- The CSP block matches `services/vault/index.html:31` exactly.
- Constants: `DEFAULT_MAX_PROOF_LIFETIME_SECS = 2 * 3600`, `VP_ERROR_PREVIEW_MAX_LEN = 200`, `DEFAULT_MAX_RESPONSE_BYTES = 2 * 1024 * 1024`, `ERROR_RESPONSE_SIZE_THRESHOLD = 4096`, `PREFIX_VALIDATION_BYTES = 64 * 1024`, `minimumReleaseAge = 1440`, `GRAPHQL_REQUEST_TIMEOUT_MS = 30_000`, `retryCount: 1`, `https://invalid.local`, chain ids `1`/`11155111`.
- Behaviour: `defaultRetryableFor` really does retry read-only methods only; `callRaw` really is uncapped; SRI hashing really is in `closeBundle` with a hard `throw`; screening really is fail-open when `UTILS_API_URL` is unset; `AddressScreeningProvider` really hard-blocks without caching; **no** HTML sink exists in any non-test source; every `target="_blank"` anchor carries `rel="noopener noreferrer"`; all workflow `uses:` are 40-char SHA-pinned.
- The stale-CODEOWNERS finding is real and correctly reported.
- The challenger-set prose is correct: `deriveLocalChallengers` exists at `signDepositorGraph.ts:105` and returns a key set (`string[]`), which is what the document claims. (Flagging this because it superficially resembles `computeNumLocalChallengers` in `primitives/challengers.ts` — that is a *different* function, used for UI minimum-deposit math, and the document is not confusing the two.)

**Two items I'd fix before merge**, both inline:

1. **The ggshield claim is fabricated** (L680, and as a mitigation at L746). `grep -rn ggshield` returns zero hits repo-wide; `.gitignore` has no such entry. A security document crediting a scanning control that would never fire is the one error class that shouldn't ship — it stops a reader from looking for the gap.
2. **The `claude-md-drift.yml` claim is inverted** (L672). The workflow *does* detect the stale fee path. What's missing is follow-through, not coverage — a different problem with a different fix.

The rest are accuracy and consistency corrections: two severity anchors that misapply the rubric defined 60 lines above them, an anchor describing untracked local state, a range-vs-pin error, a wrong file attribution, a scope gap around `simple-staking`, and an undercount of the hand-maintained critical-path lists (there are five, and two of them carry stale entries).

### Not inline

- **Prefer fixing the stale fee path to documenting it.** Correcting CODEOWNERS + `critical-path-check.yml` + `claude-md-drift.yml` is a small change. Documenting a known process hole as permanent prose is strictly worse than closing it, and once closed this document needs three edits anyway (L234, L826, the "Critical-path guard" matrix row).
- **Confirm GitHub Private Vulnerability Reporting is actually enabled** before merging. The doc routes reporters to `/security/advisories/new`; if PVR is off that link is a dead end for exactly the people you most want to reach. The `security@babylonlabs.io` channel and the org policy in `babylonlabs-io/babylon` both check out. Root placement is correct — there is no pre-existing `SECURITY.md` anywhere in the repo.
- **Editorial call, not a defect.** This repository is public, and the document enumerates six unmitigated weaknesses with file-level precision (fail-open screening, artifact body validation, `connect-src`, prod response headers, no SCA gate, the critical-path guard). Each is individually defensible — this is the honest way to write a threat model — but together they read as a prioritised target list. Worth a deliberate decision rather than a default, and worth asking whether the fail-open screening and artifact-validation gaps should be closed *before* the document advertising them lands.
- **Spelling.** British forms throughout (organisation, behaviour, minimise, defence, recognise) plus `artefact` once against `artifact` everywhere else. Not a blocker, but this text is intended to be partly byte-identical across repos, so pick one convention.
- **PR body is empty.** For a file this load-bearing, a couple of lines noting the shared-rubric version (1.0) and that this becomes a fifth hand-maintained copy of the critical-path list would help future reviewers.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the untracked package-directory example was removed from the Informational anchor.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the activation prose, invariant, and matrix now state that the SDK pre-check is conditional on hashlock being supplied, while the vault app always supplies it and SDK consumers must do the same.

SECURITY.md
kirugan: Addressed in f8f4cfd7: CLAUDE.md now uses the actual useTweakedSigner: false option, matching signing.ts and SECURITY.md.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the secret-bearing telemetry scenario is now I3 x L2 to High, then silent modifier to Critical, and the anchor agrees with the attack matrix.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the corrupt claimer-artifact scenario moved to Critical with the explicit I4 x L1 to High, then silent modifier to Critical calculation.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the deleted fee entry was replaced in CODEOWNERS and both workflows with peginFeeMath.ts and useEstimatedBtcFee.ts, and the related critical-path inventories are now aligned.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the workflow description now states that claude-md-drift.yml detects missing paths weekly and reports them to a tracker, but does not gate merges and still requires follow-through.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the nonexistent ggshield claim was removed. The document now states that no repository-configured secret-scanning control is present.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the scope now says sections 1-4 primarily describe vault and its SDK support, and the missing simple-staking CSP is documented in Scope, the CSP section, and the attack matrix.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the install-policy wording now distinguishes the two exact overrides from the js-yaml >=4.1.1 minimum-version floor.

SECURITY.md
kirugan: Addressed in f8f4cfd7: both stale paths were fixed, all five critical-path inventories are named, and the known-gap prose, matrix row, and Informational anchor were rewritten to reflect the current non-blocking drift check.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the update guidance now explicitly lists all five inventories, including claude-md-drift.yml, and both stale paths in that workflow were corrected.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the MAX_EXPIRES_AT_SECS bullet now names packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/tokenProvider.ts.

SECURITY.md
kirugan: Addressed in f8f4cfd7: the attack-matrix mitigation now lists only the gitignored local file and explicitly notes the absence of a repository-configured secret-scanning control.

jonybur (approved):

PR #2153, opened 2026-07-28
fix(vault): resolve loan reserves by on-chain id, not indexer symbol

gbarkhatov:
Closes [babylonlabs-io/baby-auditor#40](https://github.com/babylonlabs-io/baby-auditor/issues/40) (audit finding F7).

> **Rebased onto `main` after [#2148](https://github.com/babylonlabs-io/babylon-toolkit/pull/2148).** That PR renamed and restructured most of the files this one touches, so the conflicts were resolved by taking `main` and re-applying F7 on top — never the reverse. It also added a new symbol-keyed `/markets/:symbol` route, which is now folded into this change. See "What changed since the first review" at the bottom.

## The problem

The borrow/repay screen figured out *which asset it was about* by asking the GraphQL indexer. Two things flowed from that.

**1. The URL identified the reserve by its token symbol.** Clicking USDC in the picker navigated to `?reserve=usdc`, and the screen then looked for the reserve whose indexer-reported symbol was `"usdc"`. The indexer decides what that string maps to. If it serves two reserves both claiming `USDC`, the code took whichever one the indexer happened to list first — so the indexer could steer you to a different, genuine reserve than the row you clicked.

**2. The label and every number on the screen came from the indexer too** — symbol, name, and decimals, copied straight out of GraphQL. So a compromised indexer could keep a real reserve but label it as a safer token, and the page would show that label right up to the moment you signed.

The second point turned out to matter more than the audit write-up said. The repay form computed the displayed debt, your wallet balance and the **Max** button using the indexer's decimals, while the actual transaction used decimals read from the token contract. If the indexer reported 18 for a token that really has 6, "Max" would render a trillion times too small. You click Max, sign, repay dust — and the success screen tells you the position is cleared while it is still fully liquidatable.

Separately, the repay form read your wallet balance at `token.address` while the transaction guard proved `reserve.underlying`. Those are two independent indexer fields with nothing forcing them to be equal, so the balance could be read from a token you don't actually owe.

## What this PR does

The screen now proves what the reserve is before it renders anything.

Only three things are trusted after this change: the numeric reserve id in the URL, the `underlying` token address read from the chain via the env-pinned adapter, and our own address-keyed token registry. The indexer's `token.symbol`, `token.name` and `token.decimals` are no longer read anywhere on these screens.

**Routing and resolution by reserve id.** The URL is now `?reserve=5` — the reserve's on-chain id, which the indexer cannot reinterpret. Every place that navigates already had the id in hand and was throwing it away; they now pass it. Resolution takes the single matching reserve or nothing at all, so a duplicate can never be silently picked.

**A new verification step.** `verifyReserveIdentity` asserts that the reserve id maps to the expected token on-chain (reusing the existing `assertReserveMatchesOnChain` from the F8/F9 fixes), reads decimals from the token contract, and gets the symbol and name from our address-keyed registry — falling back to the token contract's own `symbol()`/`name()` for testnet addresses we don't know. If nothing trustworthy can name the token, it refuses rather than falling back to the indexer's string.

Decimals always come from the chain, even when the registry has its own copy. The registry is a hand-maintained table; the chain is what the signed transaction parses against, and the two must not disagree.

**Proven values flow to the forms.** `LoanContext` now carries a `tokenIdentity` object, so Borrow, Repay and the success step read proven symbol, name, address and decimals instead of the indexer's. An ESLint rule blocks `selectedReserve.token.*` in those components so this can't quietly regress in a later feature.

**Failing closed, with the right message.** If verification *completes and fails*, the screen blocks with a deliberately alarming message and **no Retry button** — the result is deterministic, and a Retry button would just train people to click through a security warning. If verification *can't complete* (network or RPC problem), the message is neutral and Retry is offered. Those are genuinely different events and shouldn't look the same.

The order of the render branches matters and is commented as such: the identity block sits above the loading spinner, because `isLoading` combines four sources and a still-pending price query would otherwise hide a resolved integrity failure behind a spinner.

**Copy cleanup.** The strings this adds live in `COPY.loans.detail`; the pre-existing inline strings in `Detail/` were migrated as part of the change, which was overdue per the copy rule.

## Approaches considered

**On the URL format:**

- *Keep the symbol, add the id alongside* (`?reserve=usdc&rid=5`). Readable URLs, but it keeps two parameters in sync and leaves a symbol somewhere in the path.
- *Numeric id, but accept old symbol links when exactly one reserve matches.* Old bookmarks keep working. But it keeps a symbol-keyed lookup alive, which is the thing being removed.
- **Chosen: numeric id, old symbol links blocked** with a "this link is outdated, please select the asset again" message. It's the only option where no symbol lookup survives anywhere in the resolution path. The cost is small: these are deep links into a modal, not pages people bookmark, and all in-app navigation regenerates from the route builder.

**On how far the trusted-metadata part should reach:**

- *Label only.* Verify the reserve and fix the page title, but leave the forms computing debt, balance and Max from indexer decimals. This is the literal reading of the maintainer's comment on the issue, and it's the smallest diff — but it leaves the wrong-Max-on-repay problem open, on the same screen, one click before a signature.
- *Everything, including the picker and dashboard rows.* Closest to the auditor's recommendation, but gating a list behind one RPC per row would be a real performance regression, and unknown testnet tokens would drop out of the list entirely.
- **Chosen: every surface that names a specific reserve** — the borrow/repay overlay and the borrowing markets data page. Label and every number come from proven sources. This matches the maintainer's "one on-chain read at the consequence point, not storage-layer metadata validation", while treating the numbers as part of the consequence rather than just the title.

**On failing closed vs. failing open for repay.** Blocking a repay is uncomfortable — someone near liquidation needs that button. But the repay path *already* makes these same two on-chain calls before signing, so an RPC outage blocks repay today too, just thirty seconds later after you've read a number and typed an amount. And failing open is strictly worse here: wrong decimals produce a wrong Max, you repay dust, and the success screen tells you a still-liquidatable position is cleared. Failing closed with a fast Retry keeps you engaged; a silently underpaid repay does not.

## Deliberately not in this PR

- **The asset picker and dashboard loan rows still show indexer labels.** Under an active spoof the list could say "USDC" while the detail screen blocks or says "WETH". That contradiction is intended — the detail screen is the last surface before a signature and is where the truth has to win. The follow-up is to move those surfaces onto the address-keyed registry and show "Unknown asset" for addresses we don't know.
- **No prefetching of the identity from the Loans list.** The spoke address is already cached per adapter and the result is cached for the session, so the cost is roughly two RPC calls the first time you open a reserve. If that proves noticeable, prefetching against the same cache key is the fix.
- **`fetchConfig` still doesn't cross-check that `reserve.underlying` equals `underlyingToken.address`.** Not needed now that the screens prove `reserve.underlying` directly, but worth a follow-up issue.

## Files worth reviewing closely

- `services/vault/src/applications/aave/services/verifyReserveIdentity.ts` — the new proof; note that on-chain decimals win over the registry's copy even on a registry hit.
- `services/vault/src/clients/eth-contract/erc20/query.ts` — `getERC20Symbol`/`getERC20Name` return null only for a contract-level failure and rethrow transport errors (see the review fix below).
- `services/vault/src/applications/aave/hooks/useVerifiedReserveIdentity.ts` — caching and retry policy; note it deliberately does *not* use `keepPreviousData`, unlike the price hook, because carrying the previous reserve's decimals across an asset switch is the exact bug being fixed.
- `services/vault/src/applications/aave/components/Detail/ReserveDetailPanel.tsx` — branch order, and the single `verified` narrowing that the success step shares so it can't report a label the form refused.
- `services/vault/src/applications/aave/components/LoanCard/Repay/index.tsx` — the balance read moving to the proven address.

## What changed since the first review

**1. `jrwbabylonlab`'s P1 on the symbol/name fallback.** `.catch(() => null)` treated an RPC failure identically to a token with no `symbol()`. A null symbol throws `UnknownReserveTokenError`, which `isIntegrityFailure` classifies as a conclusion — so the query didn't retry and the screen rendered integrity copy with no retry affordance. One transient blip became a permanent spoof warning until reload.

The handling moved into the readers, where the ABI knowledge already is: `getERC20Symbol`/`getERC20Name` return null only when the failure is the contract's own answer (viem's `ContractFunctionZeroDataError` for an absent method, or a revert), and rethrow anything else. `symbol()` and `name()` are optional in ERC-20, so an absent method is a legitimate null; a dropped connection is not, and now propagates as an ordinary error that retries and offers Retry. `verifyReserveIdentity` drops both catches and the comment that justified them.

**2. `jrwbabylonlab`'s note on `routes.ts` conflicting with #2148.** Confirmed and handled: every conflict was resolved by taking `main` and re-applying F7, so none of the symbol routing came back. `useDashboardState.selectableBorrowedAssets` was deleted by #2148 and stays deleted; the picker's repay rows now carry their own reserve id, which also let the symbol-keyed `priceBySymbol` workaround go away.

**3. Market Info folded in.** #2148 added a per-row Market Info button routing to `/markets/:symbol`, resolved by `borrowableReserves.find(r => r.token.symbol === ...)` — the same pattern F7 removes. It now routes by reserve id, resolves to exactly one reserve or none, and renders its header only from the verified identity. `ReserveIdentityBlock` moved up to the shared components directory since two surfaces use it.

## Testing

- New: `verifyReserveIdentity.test.ts`, `useVerifiedReserveIdentity.test.tsx`, `routes.test.ts`, `AssetPill.test.tsx`, `ReserveDetailPanel.test.tsx`, `BorrowingMarketsData.test.tsx`.
- Extended: `useAaveReserveDetail.test.tsx` covers duplicate symbols, duplicate ids, a spoofed symbol, wrong indexer decimals, garbage and legacy URL params, and the integrity hard-block. `query.test.ts` covers the sanitizing of hostile `symbol()` values plus the missing-method vs transport-failure split.
- Updated for the id-based route: `router.test.tsx`, `AssetSelectionPanel.test.tsx`, `LoanFlowOverlay.test.tsx`.
- Full vault suite: 2834 passing across 263 files. Typecheck and lint clean.

Manual checks:

1. Loans → Borrow → pick an asset. The URL should read `?reserve=<number>`, and the label, debt and Max should match the real token. The picker, form and success screen stay in one dialog (#2148's fix must survive).
2. Market Info on a picker row → `/markets/<number>`, header labelled from the verified identity; Back to assets returns to the picker.
3. Change the URL to `?reserve=usdc` → outdated-link message, no numbers rendered. `?reserve=99999`, `?reserve=abc`, `?reserve=0x5`, `/markets/usdc` → blocked, no crash.
4. Switch assets with the pill — no flash of the previous asset's symbol or decimals, and the open asset's row is the one marked.
5. Repay with exactly one loan → still skips the picker and opens the repay form directly.
6. Go offline in devtools and open a reserve → neutral "Couldn't verify this asset" **with** a working Retry (not the alarming no-Retry copy); restore the network and it recovers on reconnect.
7. God-mode demo loan rows still render with both actions disabled and never navigate.

To see the compromised state (the alarming, retry-less block), temporarily throw from `verifyReserveIdentity` — there is no way to trigger it from the UI, which is the point.

### DIFF
--- a/services/vault/src/applications/aave/services/verifyReserveIdentity.ts
+++ b/services/vault/src/applications/aave/services/verifyReserveIdentity.ts
@@ -118,9 +118,17 @@ export async function verifyReserveIdentity(
   // Unknown address (typically a testnet deployment). The address is proven and
   // is what drives value, so a misleading `symbol()` here is a display concern
   // only — but it is still bound to the address, unlike the indexer's string.
+  //
+  // `symbol()` and `name()` are both optional in ERC-20, so a read that fails
+  // here is almost certainly an absent method rather than a dead RPC: the
+  // `getERC20Decimals` call above is uncaught and already completed a
+  // successful read against this same address, so the transport is known good.
+  // Treat a failure as "the token doesn't provide it" — a missing name falls
+  // back to the symbol below, and a missing symbol leaves nothing trustworthy
+  // to label with, which is a hard block rather than a retryable fault.
   const [symbol, name] = await Promise.all([
-    getERC20Symbol(underlying),
-    getERC20Name(underlying),
+    getERC20Symbol(underlying).catch(() => null),
+    getERC20Name(underlying).catch(() => null),
   ]);
 
   if (!symbol) {
--- a/services/vault/src/applications/aave/services/verifyReserveIdentity.ts
+++ b/services/vault/src/applications/aave/services/verifyReserveIdentity.ts
@@ -119,16 +119,13 @@ export async function verifyReserveIdentity(
   // is what drives value, so a misleading `symbol()` here is a display concern
   // only — but it is still bound to the address, unlike the indexer's string.
   //
-  // `symbol()` and `name()` are both optional in ERC-20, so a read that fails
-  // here is almost certainly an absent method rather than a dead RPC: the
-  // `getERC20Decimals` call above is uncaught and already completed a
-  // successful read against this same address, so the transport is known good.
-  // Treat a failure as "the token doesn't provide it" — a missing name falls
-  // back to the symbol below, and a missing symbol leaves nothing trustworthy
-  // to label with, which is a hard block rather than a retryable fault.
+  // Both reads resolve to null when the token simply doesn't implement the
+  // method, and throw when the RPC couldn't be reached. That split is what
+  // keeps a transient blip from being reported as a spoof: it propagates as an
+  // ordinary error, which `isIntegrityFailure` rejects, so the query retries.
   const [symbol, name] = await Promise.all([
-    getERC20Symbol(underlying).catch(() => null),
-    getERC20Name(underlying).catch(() => null),
+    getERC20Symbol(underlying),
+    getERC20Name(underlying),
   ]);
 
   if (!symbol) {
--- a/services/vault/src/applications/aave/services/verifyReserveIdentity.ts
+++ b/services/vault/src/applications/aave/services/verifyReserveIdentity.ts
@@ -1,0 +1,155 @@
+/**
+ * Prove what asset a reserve actually is, before its label or any of its
+ * amounts reach the screen.
+ *
+ * The loan detail screen used to take the reserve's symbol, name and decimals
+ * straight from the GraphQL indexer. A compromised indexer could keep a real
+ * `reserveId`/`underlying` pair but spoof the symbol, so the page labelled one
+ * asset while the transaction targeted another — and report decimals that made
+ * the displayed debt and Max differ from the amount actually signed.
+ *
+ * This resolves the identity from sources the indexer does not control:
+ *  1. `assertReserveMatchesOnChain` proves `reserveId -> underlying` against
+ *     the env-pinned adapter's spoke — the only authority on which token the
+ *     page is about.
+ *  2. `getERC20Decimals(underlying)` supplies decimals. This is the exact
+ *     helper the borrow/repay signing paths use, so display and signing can't
+ *     disagree.
+ *  3. The address-keyed token registry supplies the label, falling back to the
+ *     token contract's own `symbol()`/`name()` for addresses it doesn't know
+ *     (testnet/devnet deployments).
+ *
+ * The indexer's `token.symbol/name/decimals` are never consulted. If no
+ * trusted label can be produced, this throws rather than borrowing one.
+ */
+
+import type { Address } from "viem";
+
+import {
+  InvalidTokenDecimalsError,
+  getERC20Decimals,
+  getERC20Name,
+  getERC20Symbol,
+} from "@/clients/eth-contract/erc20";
+import { getRegisteredTokenByAddress } from "@/services/token/tokenService";
+
+import {
+  ReserveMismatchError,
+  assertReserveMatchesOnChain,
+} from "./assertReserveMatchesOnChain";
+
+/**
+ * The underlying is proven, but nothing trustworthy can name it: it is absent
+ * from the token registry and its contract returned no usable `symbol()`.
+ * Falling back to the indexer's string here is exactly the substitution this
+ * module exists to prevent, so callers must hard-block instead.
+ */
+export class UnknownReserveTokenError extends Error {
+  readonly code = "UNKNOWN_RESERVE_TOKEN";
+}
+
+export interface VerifiedReserveIdentity {
+  /**
+   * The proven underlying token address. Identical to the address passed in —
+   * the assert throws unless the on-chain `underlying` matches it — so callers
+   * may treat this as the on-chain value and should use it in place of the
+   * indexer-supplied `reserve.token.address`.
+   */
+  address: Address;
+  /** Registry label for `address`, or the token's own `symbol()` as last resort. */
+  symbol: string;
+  /** Registry name, the token's own `name()`, or the symbol. */
+  name: string;
+  /** On-chain `decimals()` — the same source the signing path parses against. */
+  decimals: number;
+  /** Registry icon; undefined for on-chain-derived identities. */
+  icon: string | undefined;
+  /**
+   * Where the label came from. `"onchain"` strings are attacker-influenceable
+   * display text (a token contract may return anything); `"registry"` labels
+   * are curated in this codebase.
+   */
+  source: "registry" | "onchain";
+}
+
+/**
+ * Prove a reserve's identity from on-chain and locally-curated sources.
+ *
+ * @param trustedAdapterAddress - Env-pinned Aave adapter, the same address the
+ *   borrow/repay tx is sent to. Never accept this from a caller-supplied config.
+ * @param reserveId - Reserve id being displayed.
+ * @param underlying - The reserve's underlying token address as claimed by the
+ *   indexer. Proven against the chain before anything is derived from it.
+ * @throws {ReserveMismatchError} the claimed underlying isn't what the spoke maps
+ *   `reserveId` to.
+ * @throws {InvalidTokenDecimalsError} the token reports implausible decimals.
+ * @throws {UnknownReserveTokenError} no trusted label exists for the underlying.
+ */
+export async function verifyReserveIdentity(
+  trustedAdapterAddress: Address,
+  reserveId: bigint,
+  underlying: Address,
+): Promise<VerifiedReserveIdentity> {
+  // Assert first, matching the borrow path's ordering, so a mismatch surfaces
+  // its own error instead of being masked by a parallel RPC rejection.
+  await assertReserveMatchesOnChain(
+    trustedAdapterAddress,
+    reserveId,
+    underlying,
+  );
+
+  // Read decimals before the registry branch, deliberately: even on a registry
+  // hit we never use the registry's copy. The registry is a hand-maintained
+  // table; on-chain decimals are what the signed transaction parses against.
+  const decimals = await getERC20Decimals(underlying);
+
+  const registered = getRegisteredTokenByAddress(underlying);
+  if (registered) {
+    return {
+      address: underlying,
+      symbol: registered.symbol,
+      name: registered.name,
+      decimals,
+      icon: registered.icon,
+      source: "registry",
+    };
+  }
+
+  // Unknown address (typically a testnet deployment). The address is proven and
+  // is what drives value, so a misleading `symbol()` here is a display concern
+  // only — but it is still bound to the address, unlike the indexer's string.
+  const [symbol, name] = await Promise.all([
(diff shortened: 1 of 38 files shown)


services/vault/src/applications/aave/services/verifyReserveIdentity.ts
```diff
@@ -0,0 +1,155 @@
...
+    };
+  }
+
+  // Unknown address (typically a testnet deployment). The address is proven and
+  // is what drives value, so a misleading `symbol()` here is a display concern
+  // only — but it is still bound to the address, unlike the indexer's string.
+  const [symbol, name] = await Promise.all([
+    getERC20Symbol(underlying),
+    getERC20Name(underlying),
+  ]);
```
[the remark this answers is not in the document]
gbarkhatov: done

jonybur (dismissed):
LGTM. The audit F7 fix is comprehensive and holds up on independent review:

- Route resolution moved to numeric reserve id (`parseReserveId` rejects `usdc`, `0x5`, `5.0`, `5e3`, `-1`, whitespace, empty, all covered by tests). No symbol-keyed lookup survives anywhere in the resolution path.
- The `verifyReserveIdentity` proof asserts the id maps to the expected underlying on-chain, then reads decimals from the token contract and takes symbol/name from the address-keyed registry (falling back to `symbol()`/`name()` for unknown testnet tokens, with `catch(() => null)` distinguishing missing-method from RPC-blip). Decimals always come from the chain even on a registry hit, so the Max computation can't be off by a factor of 10^N because the registry disagrees with the ERC20.
- Fail-closed with correct copy split: `compromised` (proven mismatch or no trusted name) has no Retry; the transient RPC branch does. The identity block sits above the loading spinner in the branch order, so a still-pending price query can't hide a resolved integrity failure - the `AaveReserveDetail` test locks that ordering in.
- The ESLint boundary rule blocks `selectedReserve.token.*` on the loan surfaces and forces the verified `tokenIdentity` through `LoanContext`, so this can't quietly regress in a later feature.
- All new user-visible strings live under `COPY.loans.detail`; the `Detail/` folder's pre-existing inline strings were migrated as part of the change, which was overdue per the copy rule.

Concur with greptile that the registry-unknown-token fallback is worth watching, but the code correctly treats `null` symbol as a hard integrity signal (no trustworthy label exists) rather than an RPC fault - that's the right call for the consequence screen.

services/vault/src/applications/aave/services/verifyReserveIdentity.ts
```diff
@@ -0,0 +1,163 @@
...
+  //
+  // `symbol()` and `name()` are both optional in ERC-20, so a read that fails
+  // here is almost certainly an absent method rather than a dead RPC: the
+  // `getERC20Decimals` call above is uncaught and already completed a
+  // successful read against this same address, so the transport is known good.
+  // Treat a failure as "the token doesn't provide it" — a missing name falls
+  // back to the symbol below, and a missing symbol leaves nothing trustworthy
+  // to label with, which is a hard block rather than a retryable fault.
+  const [symbol, name] = await Promise.all([
+    getERC20Symbol(underlying).catch(() => null),
```
jrwbabylonlab: `.catch(() => null)` treats an RPC failure identically to a token with no `symbol()`. A null symbol throws `UnknownReserveTokenError`, which `isIntegrityFailure` classifies as a conclusion — so the query does not retry and the screen renders integrity copy with no retry affordance.

Net effect: for a registry-miss token, one transient blip becomes a permanent spoof warning until the user reloads. The justifying comment ("the transport is known good" because `getERC20Decimals` succeeded) does not cover it — that is a separate round-trip.

Suggest catching only contract-level failures here and letting transport errors propagate as transient.

jrwbabylonlab (dismissed):

services/vault/src/routes.ts
```diff
@@ -18,14 +18,46 @@ export function getReserveDetailBaseRoute(isV3Enabled: boolean): string {
...
+}
+
+/**
+ * Build the reserve detail route.
+ *
+ * The `reserve` query value is the reserve's on-chain id, never its token
+ * symbol: the symbol comes from the indexer, so routing by it lets a
+ * compromised indexer decide which reserve a link opens.
+ */
 export function getReserveDetailRoute(
```
jrwbabylonlab, line 65: Conflicts with #2148, which keeps `reserveId: string` + `.toLowerCase()` here and deletes `selectableBorrowedAssets`. Resolving toward that PR reinstates symbol routing and undoes F7 — worth coordinating merge order.

gbarkhatov:
@greptile review

services/vault/src/applications/aave/services/verifyReserveIdentity.ts
gbarkhatov: Good catch — you're right that the justification didn't hold; `getERC20Decimals` is a separate round-trip.

Fixed in 61f7f49, but wider than catching only contract-level failures at this call site: the same hole existed for `symbol()`, and there it was worse. An absent `symbol()` threw a plain `Error`, which `isIntegrityFailure` rejects, so the screen showed the neutral "couldn't verify" copy with a Retry button that could never succeed.

The handling moved down into the readers, where the ABI knowledge already lives. `getERC20Symbol`/`getERC20Name` now return `null` only when the failure is the contract's own answer — viem's `ContractFunctionZeroDataError` (absent method) or `ContractFunctionRevertedError`, matched by walking the cause chain — and rethrow everything else. `verifyReserveIdentity` drops both catches and the comment.

Net effect:

- method absent → `null` → a missing name falls back to the symbol; a missing symbol → `UnknownReserveTokenError` → integrity, no Retry.
- RPC failure → propagates → ordinary error → not an integrity failure → retried twice, then the neutral block **with** Retry.

Covered in `query.test.ts` (zero-data → null, revert → null, transport → rejects) plus a propagation test in `verifyReserveIdentity.test.ts`.

services/vault/src/routes.ts
gbarkhatov: Confirmed, and thanks for flagging it before I hit it. #2148 has since merged, and the branch is now rebased onto it.

Every conflict was resolved by taking `main` and re-applying F7 on top, never the reverse — so `getReserveDetailRoute` keeps the `bigint` id and no symbol routing came back. `selectableBorrowedAssets` was deleted by #2148 and stays deleted; the picker's repay rows carry their own reserve id instead, which also let the symbol-keyed `priceBySymbol` workaround go away.

One thing #2148 added that needed folding in: the Market Info button routes to `/markets/:symbol`, resolved by `borrowableReserves.find(r => r.token.symbol.toUpperCase() === symbol)` — the same pattern this PR removes. That page now routes by reserve id, resolves to exactly one reserve or none, and renders its header only from the verified identity.

jrwbabylonlab (approved):

Issue #2106, opened 2026-07-22
Ledger: enforce device participant/index caps at deposit validation


jrwbabylonlab:
The device caps keepers ≤ 32, challengers ≤ 32 (u8), vaults ≤ 10, and BIP-86 indexes (account ≤ 100, address_index ≤ 10000). Nothing on the TS signing path enforces these — the existing `MAX_LOCAL_CHALLENGER_COUNT = 100` in `services/vault/src/utils/wasm.ts` is a display-path UX guard only. The product currently caps deposits at 2 vaults, so the vault cap is future-proofing, not a live bug.

Scope:
- Enforce keeper/challenger counts + BIP-86 index caps at deposit validation.
- Decision to make here: enforce for all wallets (protocol params should never exceed device caps anyway) or gate behind the Ledger capability.
- Timelock ranges (unblocked — v22 pinned them): `htlc_refund_timelock` 72..4320, `pegin_csv_timelock` (P) 72..1008, `payout_timelock` (t2) >90 and <4032. P and t2 are both fed from `timelockAssert`, so the deployed value must sit in [91,1008] — verify it. Enforce these now. No on-chain backstop exists for any device cap — this is FE-only validation.
- Commission value-sanity floors (found 2026-07-29 in the device code, `vault_tlv.c` on develop): the device rejects the intent when `commissionFee < 546` (dust) or `vaultAmount <= commissionFee + 2*546` — with an opaque SW_INCORRECT_DATA and full session nullification. Protocol-infeasible anyway (Rust enforces the same floor VP-side, bps pinned 1..=9999), but the FE bounds gate must catch it first with comprehensible copy. Also enumerate: `base_fee_rate` <= u32 max, challenger count >= 1.

---
[...]

PR #2154, opened 2026-07-29
deposit terms seam hardening

jrwbabylonlab:


packages/babylon-ts-sdk/src/tbv/core/services/deposit/runDepositorPresignFlow.ts
```diff
@@ -332,6 +342,19 @@ export async function runDepositorPresignFlow(
...
+  if (supportsDepositApproval(btcWallet)) {
+    if (!depositTerms) {
+      throw new Error(
+        "runDepositorPresignFlow: this wallet requires approved deposit terms but none were " +
+          "provided. Fresh deposits must pass PreparePeginResult.depositTerms; resume-path " +
+          "rebuild is not wired yet.",
+      );
+    }
+    await btcWallet.approveDepositTerms(depositTerms);
+  }
```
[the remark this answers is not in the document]
jrwbabylonlab: Deliberate fail-loud — no shipped wallet implements `approveDepositTerms` yet, and the resume-path terms rebuild is tracked in #2110.

jeremy-babylonlabs (dismissed):

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -935,6 +960,28 @@ export class PeginManager {
...
+      prepeginMaxFee: sizing.fee,
+      vaultProviderPk: vaultProviderBtcPubkey,
+      keeperPks: vaultKeeperBtcPubkeys,
+      challengerPks: universalChallengerBtcPubkeys,
+      commissionBps: params.commissionBps,
+      vaultAmounts: params.amounts,
+      depositorClaimValue: prePeginResult.depositorClaimValue,
+      peginMaxFee: prePeginResult.minPeginFee,
+    });
+    if (supportsDepositApproval(this.config.btcWallet)) {
```
gbarkhatov, line 996: **[High] This approval selects the device's Session 2, so the Pre-PegIn signature is left unapproved — and Session 1 becomes unreachable.**

I cross-read the Ledger firmware at `app-babylon-vault` @ `develop` (`5e0ef7d`), since the TODO in `buildDepositTerms.ts` already cites `vault_tlv.c` there as the reference. The device models two sessions, selected inside `APPROVE_VAULT_INTENT` (`handler/approve_vault_intent.c:168-181`):

```c
if (memcmp(G_vault_context.htlc_hashlock, zeros, VAULT_HASH256_LEN) != 0 &&
    memcmp(G_vault_intent.prepegin_txid, zeros, VAULT_HASH256_LEN) != 0) {
    ... transition to VAULT_STATE_SESSION2_PEGIN_EXPECTED
}
```

So Session 2 is entered only when a hashlock is already held **and** the intent carries a non-zero `prepegin_txid`. Both are always true here: `deriveVaultRoot` (`PeginManager.ts:697`) runs before this call and sets the hashlock, and `buildDepositTerms.ts:18` *requires* a 64-hex `prepeginTxid`. Every approval this seam can issue therefore selects Session 2 — Session 1 (`INTENT_LOADED`, the only state in which `sign_psbt_validate.c:162` permits Pre-PegIn validation) is unreachable by construction.

The consequence is on the other side: `signAndBroadcast` signs the funded Pre-PegIn with no `approveDepositTerms` call at all, by which point the device has left `INTENT_LOADED`. The dispatcher (`sign_psbt_validate.c:1357-1372`) is strictly state-gated, so that PSBT lands in a PegIn/Payout validator rather than the Pre-PegIn one.

Caveat I want to be explicit about: `sign_psbt_validate.c:941-945` defers the `SESSION2_PEGIN_EXPECTED` → `SESSION2_PAYOUT_EXPECTED` transition to `sign_custom_inputs`, which is a documented stub (`main.c:16`, NAPPS-1377). So I can't predict the exact status word from this checkout. The *shape* of the mismatch is stub-independent, though. Worth confirming the intended seam boundaries with the firmware / #2109 owners before this merges.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/runDepositorPresignFlow.ts
```diff
@@ -332,6 +342,19 @@ export async function runDepositorPresignFlow(
 
   signal?.throwIfAborted();
 
+  // Approval-capable wallets must approve the deposit terms before any signing
+  // call they authorize, including the Phase 3/4 payout and depositor-graph signing below.
+  if (supportsDepositApproval(btcWallet)) {
```
gbarkhatov, line 347: **[High] The re-approval can't repair the session: PegIn and Payout signing are separated by an intent-invalidating derive, and a fresh approval lands in `PEGIN_EXPECTED`, not `PAYOUT_EXPECTED`.**

The device expects PegIn and Payout signatures within one approved intent: `SESSION2_PEGIN_EXPECTED` → (PegIn signed) → `SESSION2_PAYOUT_EXPECTED` (`vault_context.h:11-34`). This flow splits them across a lot of intervening work — the PegIn PSBTs are signed back in `preparePeginCommit`, while the Payout PSBTs are signed below in Phase 3, after ETH registration, the BTC confirmation wait, and the VP auth-anchor derive in `ensureAuthenticatedVpClient`.

That derive is the problem. `derive_context_hash.c:16` calls `vault_context_invalidate` unconditionally at handler entry, zeroing the loaded intent back to `IDLE`. This is exactly the invariant `depositTerms.ts:64` documents — but re-approving here doesn't restore the state that was lost. `approve_vault_intent.c:168-181` has exactly one auto-advance, and it targets `SESSION2_PEGIN_EXPECTED`; there is no path from a fresh approval to `SESSION2_PAYOUT_EXPECTED`. The payout PSBTs signed a few lines below would then hit the PegIn validator.

Same NAPPS-1377 caveat as on `PeginManager.ts` applies — but this half doesn't depend on the stub: no matter how the deferred transition lands, a derive between the two signing groups means they can never share one device intent.

Related, and probably worth resolving in the same conversation: because `signAndSubmitPayouts` is called once per vault lane in `useDepositFlow`, an approval-capable wallet gets N+1 approvals per deposit (one in `preparePeginCommit` plus one per lane), each carrying the full multi-vault terms. Worth confirming that matches the device's session model too.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -935,6 +960,28 @@ export class PeginManager {
...
+    // Always build the deposit terms so callers get them back regardless of
+    // wallet capability; only approval-capable wallets need the call below.
+    // peginMaxFee reuses assertWasmPeginSizing's already-asserted minPeginFee
+    // (via prePeginResult) instead of recomputing it.
+    const depositTerms = buildDepositTerms({
+      protocolFeeRate: params.protocolFeeRate,
+      timelockPegin: params.timelockPegin,
+      timelockRefund: params.timelockRefund,
+      prepeginTxid: prePeginTxid,
+      prepeginMaxFee: sizing.fee,
```
gbarkhatov, line 987: **[Medium] `prepeginMaxFee` comes from the sizing pass, not from the tx that actually gets signed.**

`sizing.fee` was computed by `selectUtxosForPegin` against the *sizing*-pass output set. The transaction the device caps and the depositor signs is `fundedPrePeginTxHex`, built in this commit pass — and `fundPeginTransaction` validates `changeAmount` bounds but never asserts `inputs − outputs === sizing.fee`.

The two passes are deterministic and `assertWasmPeginSizing` pins the HTLC values, so they agree today. But nothing enforces it, and per the docstring on `DepositTerms.prepeginMaxFee` this number is a hardware signing bound — CLAUDE.md critical path #2 puts the cross-check at the broadcast site, not only at the estimator.

`prePeginResult.totalOutputValue` is already in scope, so this is cheap:

```ts
const fundedFee =
  sizing.selectedUTXOs.reduce((sum, u) => sum + BigInt(u.value), 0n) -
  prePeginResult.totalOutputValue -
  sizing.changeAmount;
if (fundedFee !== sizing.fee) {
  throw new Error(
    `Pre-PegIn funded fee ${fundedFee} does not match the sizing-pass fee ` +
      `${sizing.fee}; refusing to publish a deposit-terms fee bound that the ` +
      `funded tx does not satisfy.`,
  );
}
```

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -229,53 +233,62 @@ export function usePayoutSigningState({
...
+                    }
+                  }
+                },
+              }
+            : {}),
+          // Object spread drops prototype methods (class-instance wallets), so
+          // approveDepositTerms must be forwarded explicitly like signPsbts above.
+          ...(supportsDepositApproval(wallet)
+            ? {
+                approveDepositTerms: (terms) =>
```
gbarkhatov: **[Medium] This forwarding makes the resume path a guaranteed failure for approval-capable wallets.**

Forwarding `approveDepositTerms` makes `supportsDepositApproval` return true on the wrapper, but the `signAndSubmitPayouts` call at line 294 never passes `depositTerms` — this hook has no source for it. `runDepositorPresignFlow` then takes the `!depositTerms` branch and throws "resume-path rebuild is not wired yet", and it throws *after* the user has already approved the auth-anchor popup.

Fail-closed is the right instinct and I'd keep it. What concerns me is when this path runs: the Pre-PegIn is broadcast and the vaults are registered on-chain by then, so a depositor who reloads mid-flow can't complete presigning at all and waits out the HTLC refund timelock.

No live regression today since nothing implements the capability yet, but the seam ships with the dead end already in place. Three options, roughly in order of preference: wire the rebuild from on-chain params plus the stored pegin record; gate approval-capable wallets out of the deposit entry point until it lands; or don't forward the capability at a site that structurally cannot supply terms. Same call you made for OneKey — if the signing path is broken, disable the wallet rather than ship the trap.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/buildDepositTerms.ts
```diff
@@ -0,0 +1,65 @@
...
+export function buildDepositTerms(inputs: BuildDepositTermsInputs): DepositTerms {
+  const txid = inputs.prepeginTxid.toLowerCase();
+  if (!/^[0-9a-f]+$/.test(txid) || txid.length !== TXID_HEX_LENGTH) {
+    throw new Error(`buildDepositTerms: prepeginTxid must be 64 hex chars, got "${inputs.prepeginTxid}"`);
+  }
+  if (inputs.vaultAmounts.length === 0) {
+    throw new Error("buildDepositTerms: at least one vault amount is required");
+  }
+  // TODO(#2106): device-range validation — timelock ranges, participant caps,
+  // and the device's commission value-sanity floors (commissionFee >= 546 dust,
```
gbarkhatov, line 28: **[Medium] Both numbers in this TODO are wrong, and one of the cases it defers is reachable today.**

Checked against `app-babylon-vault` @ `develop`:

- `vault_tlv.c:73` rejects `commission_fee == 0` outright (`VAULT_TLV_ERR_VALIDATION`) — it is not a `>= 546` floor.
- `vault_tlv.c:179-183` enforces `vault_amount > commission_fee + 2 * VAULT_DUST_LIMIT`, and `vault_constants.h:41` defines `VAULT_DUST_LIMIT` as **330**, not 546.

So the comment should read `commissionFee != 0` and `vaultAmount > commissionFee + 2*330`.

The zero case isn't hypothetical: `commissionBps: 0` passes the check below, and even a non-zero bps floors to `0n` on a small enough `vaultAmount` (e.g. 1 bps on 9,999 sats). Both produce terms the device will reject at parse time. Since this is one cheap comparison and it's the same class of check already being done for `commissionBps`, I'd pull the `commissionFee != 0` assertion forward rather than defer it to #2106 with the rest.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/buildDepositTerms.ts
```diff
@@ -0,0 +1,65 @@
...
+  }));
+
+  return {
+    baseFeeRate: inputs.protocolFeeRate,
+    peginCsvTimelock: inputs.timelockPegin,
+    payoutTimelock: inputs.timelockPegin,
+    htlcRefundTimelock: inputs.timelockRefund,
+    prepeginTxid: txid,
+    prepeginMaxFee: inputs.prepeginMaxFee,
+    keeperPks: [...inputs.keeperPks],
```
gbarkhatov, line 64: **[Medium] The pubkey fields are passed through unvalidated, but the published type documents guarantees about them.**

`depositTerms.ts` documents `vaultProviderPk` as "x-only lowercase hex (64 chars)" and these two lists as "x-only hex, sorted ascending". Nothing enforces any of that: `PeginManager` applies only `stripHexPrefix` (no lowercasing, no length check, no sortedness check), unlike `depositorBtcPubkey`, which goes through `normalizeXOnlyPubkey`.

These are the strings a depositor reads on the device screen to decide whether to approve, so the guarantees aren't cosmetic. The "not a second validator" docstring is also already contradicted by the four checks above — txid, non-empty vaults, timelocks, and bps. Either add a 64-lowercase-hex assertion plus an ascending-order assertion for both lists, or drop the claims from the JSDoc so a provider author doesn't build a wire encoder on top of them.

Small related gap: `buildDepositTerms.test.ts` pins that the lists pass through *unchanged*, but not that they're **copies**. The `[...]` spread here is load-bearing — a caller mutating `inputs.keeperPks` after the fact must not mutate approved terms — and a regression to `keeperPks: inputs.keeperPks` survives the current suite.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/buildDepositTerms.ts
```diff
@@ -0,0 +1,65 @@
...
+    inputs.commissionBps < 0 ||
+    inputs.commissionBps >= MAX_VP_COMMISSION_BPS_EXCLUSIVE
+  ) {
+    throw new Error(
+      `buildDepositTerms: commissionBps must be an integer in ` +
+        `[0, ${MAX_VP_COMMISSION_BPS_EXCLUSIVE}), got ${inputs.commissionBps}`,
+    );
+  }
+
+  const bpsDenominator = BigInt(MAX_VP_COMMISSION_BPS_EXCLUSIVE);
```
gbarkhatov: Nit: a validation bound is doing double duty as an arithmetic denominator. They're numerically equal today, but they aren't the same concept — if the protocol ever tightened the commission cap to, say, 5000 bps, this line would silently change the commission math rather than just the accepted range.

`payout.ts:463` shares the conflation, so a shared `BPS_DENOMINATOR = 10_000` next to `MAX_VP_COMMISSION_BPS_EXCLUSIVE` would fix both sites and make the coupling explicit.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -478,6 +481,14 @@ export function useDepositFlow(
           ...(typeof confirmedBtcWallet.signPsbts === "function"
             ? { signPsbts: signPeginBatch }
             : {}),
+          // Object spread drops prototype methods (class-instance wallets), so
+          // approveDepositTerms must be forwarded explicitly like signPsbts above.
+          ...(supportsDepositApproval(confirmedBtcWallet)
+            ? {
+                approveDepositTerms: (terms) =>
```
gbarkhatov: Nit: this six-line block is now identical in three places — here, line 961 below, and `usePayoutSigningState.ts:283`. The comment explaining *why* it's needed is duplicated verbatim too.

Since the whole point is that it's easy to forget at a new wrapper site, a helper exported next to `supportsDepositApproval` would collapse all three and give the next wrapper an obvious thing to reach for:

```ts
export function forwardDepositApproval(
  wallet: BitcoinWallet,
): Partial<DepositTermsApprover> {
  return supportsDepositApproval(wallet)
    ? { approveDepositTerms: (terms) => wallet.approveDepositTerms(terms) }
    : {};
}
```

Then each site is `...forwardDepositApproval(confirmedBtcWallet),`.

packages/babylon-ts-sdk/src/tbv/core/managers/__tests__/PeginManager.test.ts
```diff
@@ -69,6 +75,24 @@ vi.mock("../../primitives/psbt/verifyScriptPathSchnorrSignature", () => ({
   assertScriptPathSchnorrSignature: vi.fn(),
 }));
 
+// Passthrough observer: records each per-vault PegIn build so the seam test can
+// assert the htlcVout bind-checks run BEFORE deposit-terms approval (F6 hoist).
```
gbarkhatov: Nit: "F6 hoist" doesn't mean anything to a reader who wasn't in the review thread, and it won't mean anything to you in six months either. Same pattern at line 751 ("F6 hoist"), line 745 ("design §6"), `pegin.test.ts:115` ("F4"), and `useDepositFlow.test.tsx:216`/`:1514` + `usePayoutSigningState.test.tsx:668` ("F2 regression").

The behaviours these describe are good and worth keeping — just state them directly, e.g. "both per-vault PegIn builds must run before the depositor approves on the device" instead of "F6 hoist". Issue references like `#2106`/`#2109` are fine; it's the opaque finding IDs that don't survive the review they came from.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * sat/vB; the tx-graph fee rate (protocolFeeRate), NOT the mempool funding
+   * rate. An approving wallet bounds each payout's fee at baseFeeRate x a
+   * conservative vsize estimate (v22 §4.9.7.1) — pass the exact graph rate,
+   * not an inflated ceiling.
+   */
+  baseFeeRate: bigint;
+  /** Vault-UTXO CSV timelock (blocks). */
+  peginCsvTimelock: number;
+  /** Assert:0 payout timelock; comes from the same protocol param as peginCsvTimelock. */
```
gbarkhatov: Typo: `Assert:0` looks like a mangled `Assert→Payout`.

Worth fixing precisely because the rest of the line is correct and load-bearing — I verified `timelockPegin` really is `Number(timelockAssert)` (`protocol-params-reader.ts:104`), so mapping both fields from the one param is right. The garbled arrow just makes a correct decision look like a copy-paste slip.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts
```diff
@@ -0,0 +1,89 @@
...
+  /** x-only hex, sorted ascending. */
+  keeperPks: string[];
+  /**
+   * x-only hex, sorted ascending independently of keeperPks. Universal
+   * challengers only — the full graph challenger set is keeperPks ∪
+   * challengerPks (vault keepers are the local challengers).
+   */
+  challengerPks: string[];
+  /** Per-vault groups, ordered by ascending htlcVout. */
+  vaults: DepositTermsVaultGroup[];
```
gbarkhatov: Nit: these three arrays are mutable while `BuildDepositTermsInputs` already uses `readonly string[]` / `readonly bigint[]`. For a value object that gets handed to a device for approval and then held across later signing calls, the output is the side that most wants to be frozen.

`readonly string[]` on `keeperPks`/`challengerPks` and `readonly DepositTermsVaultGroup[]` here would make the defensive copies in `buildDepositTerms` meaningful at the type level too.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Implemented only by depositor-approval wallets (e.g. a Ledger vault provider).
+ * Either a class field or a prototype method works — the deposit flow spreads
+ * the wallet object but forwards this method explicitly at every wrapper site.
+ */
+export interface DepositTermsApprover {
+  approveDepositTerms(terms: DepositTerms): Promise<void>;
+}
+
+/**
+ * Seam invariant: never call deriveContextHash between approveDepositTerms and
```
gbarkhatov: Doc placement: this block describes the seam invariant and the design alternatives that were rejected, none of which describe what `supportsDepositApproval` actually does — it's a one-line duck-type probe. A consumer hovering it in an editor gets architecture notes instead of "returns true if the wallet implements `approveDepositTerms`".

The invariant itself is the most important sentence in this file, so I'd move it up to `DepositTermsApprover` (or the module doc in `index.ts`, which is what a provider author reads first) and leave a one-liner here. See my comment on `runDepositorPresignFlow.ts:347` — the invariant as written is exactly what the current PegIn/Payout split appears to violate, which is another reason it deserves a spot people will actually find.

gbarkhatov (dismissed):
Reviewed the full diff plus the surrounding call graph, and cross-read the Ledger firmware at `app-babylon-vault` @ `develop` (`5e0ef7d`) for the device-side claims.

**Approving as requested — but I'd treat the two High findings as gating before any approval-capable wallet is enabled for deposits, not as post-merge cleanup.** They're both about the seam's placement relative to the device's session state machine rather than about this code's internal consistency, so they may well be resolved by a conversation with the firmware / #2109 owners rather than a code change here.

### What's good

- **The SDK-internal ordering is sound.** I traced every `deriveContextHash` site (`deriveVaultRoot` in `preparePegin`, `ensureAuthenticatedVpClient` in `payoutSigning`) and each one precedes its `approveDepositTerms`. The test pins it with `callOrder.lastIndexOf(...)` rather than `indexOf` — the right assertion, and an easy one to get subtly wrong.
- **The prototype-method forwarding fix.** `{...wallet}` silently dropping a class-prototype `approveDepositTerms` is a genuinely nasty failure class, and it's fixed at all three wrapper sites *and* regression-tested with a real `class` plus a `#private` field so a wrong-`this` forward throws instead of passing.
- **Hoisting the per-vault PegIn builds above the approval** so `htlcVout` bind-checks fail before the depositor commits on hardware.
- **Threading the already-asserted `minPeginFee`** out of `assertWasmPeginSizing` instead of recomputing it.
- **Test quality is well above average** — distinct amounts so a vault-order swap is visible, distinct `protocolFeeRate` vs `minPeginFeeRate` so two same-typed rates can't be confused, independent `computeMinClaimValue`/`computeMinPeginFee` recomputation with an explicit `not.toBe` between them, and per-call rather than any-call assertions over the per-vault loop.

### Findings

| Severity | Where | What |
|---|---|---|
| High | `PeginManager.ts:981` | Every approval this seam issues selects the device's Session 2, so the Pre-PegIn signature is never approved and Session 1 is unreachable |
| High | `runDepositorPresignFlow.ts:347` | PegIn and Payout signing are split by an intent-invalidating derive; re-approving lands in `PEGIN_EXPECTED`, not `PAYOUT_EXPECTED` |
| Medium | `PeginManager.ts:972` | `prepeginMaxFee` is taken from the sizing pass, not measured from the funded tx it bounds |
| Medium | `usePayoutSigningState.ts:287` | Resume path forwards the capability but structurally cannot supply terms → guaranteed throw, after the on-chain commitment |
| Medium | `buildDepositTerms.ts:25` | `commissionFee: 0` is producible and the device rejects it; both numbers in the TODO are wrong (546 → `!= 0`, and the dust limit is 330) |
| Medium | `buildDepositTerms.ts:61` | Pubkey fields pass through unvalidated despite documented guarantees; the defensive copy isn't pinned by a test |
| Low | 6 more | denominator/bound conflation, triplicated forwarding block, audit-style markers in test names, doc nits |

### Checked and *not* raising

- `payoutTimelock: timelockPegin` is correct — `timelockPegin` is literally `Number(timelockAssert)` (`protocol-params-reader.ts:104`), so both fields legitimately come from the one protocol param.
- The commission math matches `payout.ts:463` exactly — same constant, same floor, same base value.
- `PreparePeginParams.commissionBps` becoming required is a breaking change on a published type (`@babylonlabs-io/ts-sdk` 0.1.2, not private). There's no `.changeset` convention in this repo, so there's nothing mechanical to add — flagging only in case there are external consumers to notify.

I did not execute the test suites.

jonybur (dismissed):
LGTM. Read the SDK ordering and the wrapper-forwarding fixes against the CLAUDE.md critical paths — the SDK-internal sequence in `preparePegin` (sizing → derive → per-vault build → approve → sign) is right, the F6 hoist so `buildPeginTxFromFundedPrePegin`'s htlcVout bind-check runs before device approval is the right guard, and the three wrapper sites all forward `approveDepositTerms` explicitly rather than relying on spread. The `assertWasmPeginSizing` returning `minPeginFee` (rather than recomputing) is a nice tightening on a WASM-boundary value. gbarkhatov's inline High findings around device session placement and the resume-path guaranteed-throw are the real gating discussion — deferring to those. CI green.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/buildDepositTerms.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/buildDepositTerms.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useDepositFlow.ts

packages/babylon-ts-sdk/src/tbv/core/managers/__tests__/PeginManager.test.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
jrwbabylonlab: Accurate for current develop — but that's the mid-refactor state. Their own progress tracker commits to removing the two-session model and accepting any intent-bound PSBT in any order under a single INTENT_LOADED state (still TO DO on their side); we're confirming with Ledger it lands in their v1 (Aug 10).

packages/babylon-ts-sdk/src/tbv/core/services/deposit/runDepositorPresignFlow.ts
jrwbabylonlab: Same root cause — approve→PAYOUT_EXPECTED only exists after their any-order refactor, which their tracker commits to but hasn't built (their caps + tests are blocked behind it too), so we're chasing the timeline. The seam stays choreography-agnostic; #2109's provider owns session handling.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
jrwbabylonlab: Keeping the fail-loud — the terms rebuild is scoped in #2110, and the enable flip keeps approval-capable wallets out until it lands.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/buildDepositTerms.ts
jrwbabylonlab: develop HEAD (66323a0) has VAULT_DUST_LIMIT = 546 (vault_constants.h:67) and rejects val < VAULT_DUST_LIMIT at vault_tlv.c:221 — looks like 5e0ef7d predates it. The floor check lands with the rest of the device bounds in #2106.

gbarkhatov (approved):

PR #2155, opened 2026-07-29
fix(vault): don't label RPC/node errors

jrwbabylonlab:
## What
`classifyError` labeled every `RpcRequestError` without revert data as a
`network` error → *"Check your connection."* But viem wraps every JSON-RPC error
(rate limit, `-32603` internal, `-32002` already-pending, nonce/"already known")
in an inner `RpcRequestError`, so non-connection failures blamed the user's
connection.

## Fix
Split the classifier: `network` now covers only true transport failures
(`HttpRequestError`, `WebSocketRequestError`, `SocketClosedError`, viem
`TimeoutError`); a new `rpc-error` kind handles `RpcRequestError` → *"We couldn't
complete your request right now. Please wait a moment and try again."* Contract-revert
fall-through preserved.

### DIFF
--- a/services/vault/src/utils/errors/__tests__/formatting.test.ts
+++ b/services/vault/src/utils/errors/__tests__/formatting.test.ts
@@ -11,6 +11,7 @@ import { resolve } from "path";
 import {
   HttpRequestError,
   InsufficientFundsError,
+  RpcRequestError,
   UserRejectedRequestError,
 } from "viem";
 import { describe, expect, it } from "vitest";
@@ -278,7 +279,30 @@ describe("Error Formatting", () => {
         cause: revertWrap,
         walk: () => {},
       });
-      expect(sanitizeErrorMessage(outer)).not.toMatch(/Network error/i);
+      // Pin the fall-through: the `0x` guard sends the walker to
+      // `revertRpc.cause` (undefined) → classifyError returns null →
+      // sanitizeErrorMessage surfaces the outer message. Asserting the exact
+      // string keeps the guard protected (the `rpcError` copy also lacks
+      // "Network error", so a looser matcher would pass without it).
+      expect(sanitizeErrorMessage(outer)).toBe("execution reverted");
+    });
+
+    it("falls through a data-less revert (code 3, no 0x data) to the real reason", () => {
+      // Some providers return a revert with no data. `code === 3` still marks
+      // it a revert, so it must surface the reason — not "wait and retry".
+      const revertRpc = Object.assign(new Error("RPC Request failed."), {
+        name: "RpcRequestError",
+        code: 3,
+        walk: () => {},
+      });
+      const outer = Object.assign(new Error("execution reverted: SomeReason"), {
+        name: "ContractFunctionExecutionError",
+        cause: revertRpc,
+        walk: () => {},
+      });
+      const msg = sanitizeErrorMessage(outer);
+      expect(msg).not.toMatch(/wait a moment and try again/i);
+      expect(msg).toBe("execution reverted: SomeReason");
     });
 
     it("classifies RpcRequestError without revert data as an rpc-error, not the user's connection", () => {
@@ -295,12 +319,27 @@ describe("Error Formatting", () => {
       expect(msg).toMatch(/wait a moment and try again/i);
     });
 
-    it("collapses HttpRequestError (RPC transport failure)", () => {
+    it("collapses a status-less HttpRequestError (fetch threw) as network", () => {
+      // No `status` = the request never got a response = a real transport
+      // failure, so "check your connection" is honest.
       const err = Object.assign(
         new Error("HTTP request failed.\nURL: https://eth-rpc.example/key"),
         { name: "HttpRequestError" },
       );
       expect(sanitizeErrorMessage(err)).toMatch(/Network error/i);
+    });
+
+    it("routes an HttpRequestError WITH a status (429/5xx) to rpc-error, not the connection", () => {
+      // viem sets `status` when the server answered (rate limit / outage) —
+      // a provider problem, so it must NOT tell the user to check their
+      // connection. This is the most likely RPC failure over `http()`.
+      const err = Object.assign(new Error("HTTP request failed."), {
+        name: "HttpRequestError",
+        status: 429,
+      });
+      const msg = sanitizeErrorMessage(err);
+      expect(msg).not.toMatch(/check your connection/i);
+      expect(msg).toMatch(/wait a moment and try again/i);
     });
 
     it("collapses TimeoutError (request timed out) — viem shape", () => {
@@ -335,11 +374,12 @@ describe("Error Formatting", () => {
       expect(sanitizeErrorMessage(err)).toMatch(/Network error/i);
     });
 
-    it("does NOT tell the user to check their connection for a nonce error wrapped in RpcRequestError", () => {
-      // Regression: viem's chain for "already known" / nonce errors is
+    it("classifies a nonce-too-low / already-known chain as already-submitted, not a retry", () => {
+      // viem's chain for "already known" / nonce errors is
       // TransactionExecutionError -> NonceTooLowError -> RpcRequestError. The
-      // inner RpcRequestError carries no `0x` revert data, so before the
-      // transport/RPC split the walker labeled it "Check your connection".
+      // NonceTooLowError frame (above the RpcRequestError frame) means the tx
+      // is already in the mempool, so the copy must send the user to their
+      // wallet / an explorer rather than invite a retry.
       const inner = Object.assign(new Error("already known"), {
         name: "RpcRequestError",
         code: -32000,
@@ -357,11 +397,18 @@ describe("Error Formatting", () => {
       });
       const msg = sanitizeErrorMessage(outer);
       expect(msg).not.toMatch(/check your connection/i);
-      expect(msg).toMatch(/wait a moment and try again/i);
+      expect(msg).not.toMatch(/wait a moment and try again/i);
+      expect(msg).toMatch(/already submitted/i);
+    });
+
+    it("classifies a raw 'already known' provider message as already-submitted", () => {
+      // No viem class wrapping — just the geth mempool string.
+      const msg = sanitizeErrorMessage(new Error("already known"));
+      expect(msg).toMatch(/already submitted/i);
     });
 
     // The rest of the suite builds synthetic errors via Object.assign so we
-    // can construct arbitrary cause chains. The three smoke tests below
+    // can construct arbitrary cause chains. The four smoke tests below
     // instead use real viem class constructors — if viem renames a class
     // in a future version (e.g. `InsufficientFundsError` →
     // `InsufficientFundsRpcError`), CI fails here loudly rather than
@@ -384,6 +431,18 @@ describe("Error Formatting", () => {
         url: "https://rpc.example/",
       });
       expect(sanitizeErrorMessage(err)).toMatch(/Network error/);
+    });
+
+    it("matches a real viem RpcRequestError instance (node/provider error)", () => {
+      // A JSON-RPC error the node returned. `walk` comes from BaseError so
+      // `isViemShape` passes; `data` is undefined and code is not 3, so it
+      // lands in `rpc-error` — not "check your connection".
+      const err = new RpcRequestError({
+        body: { method: "eth_call" },
+        error: { code: -32603, message: "internal error" },
+        url: "https://rpc.example/",
+      });
+      expect(sanitizeErrorMessage(err)).toMatch(/wait a moment and try again/i);
     });
 
     it("collapses Vite chunk 404 (Chrome/Edge wording) wrapped in viem error", () => {
--- a/services/vault/src/utils/errors/__tests__/formatting.test.ts
+++ b/services/vault/src/utils/errors/__tests__/formatting.test.ts
@@ -281,15 +281,18 @@ describe("Error Formatting", () => {
       expect(sanitizeErrorMessage(outer)).not.toMatch(/Network error/i);
     });
 
-    it("still classifies RpcRequestError without revert data as network", () => {
-      // Pure transport failure — no `.data`, no revert. Should hit the
-      // network bucket.
+    it("classifies RpcRequestError without revert data as an rpc-error, not the user's connection", () => {
+      // A JSON-RPC error the node returned (here -32603 InternalRpcError).
+      // It's a provider/node failure, not transport, so it must NOT tell the
+      // user to check their connection.
       const err = Object.assign(new Error("RPC Request failed."), {
         name: "RpcRequestError",
         code: -32603,
         walk: () => {},
       });
-      expect(sanitizeErrorMessage(err)).toMatch(/Network error/i);
+      const msg = sanitizeErrorMessage(err);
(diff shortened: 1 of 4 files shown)


jeremy-babylonlabs (dismissed):

services/vault/src/utils/errors/formatting.ts
```diff
@@ -174,35 +176,42 @@ export function classifyError(err: unknown): ErrorKind | null {
...
-    // enough not to collide; `TimeoutError` is generic (AbortSignal.timeout,
-    // `ky`, DOM APIs) and must be gated on viem shape.
+    // Transport failures — the request never got a usable response. These
+    // are the only errors where "check your connection" is honest.
+    // `HttpRequestError` / `WebSocketRequestError` / `SocketClosedError`
+    // names are viem-specific enough not to collide; `TimeoutError` is
+    // generic (AbortSignal.timeout, `ky`, DOM APIs) and must be gated on
+    // viem shape.
     if (
       obj.name === "HttpRequestError" ||
```
gbarkhatov: **Medium — the bug this PR fixes survives on the most likely path.**

viem's http transport throws `HttpRequestError` for two different things (`viem/_esm/utils/rpc/http.js:61-79`):

- `!response.ok` → constructed **with** `status` (429 rate limit, 403, 5xx from Alchemy/Infura). The request *did* get a response.
- `fetch` threw → constructed **without** `status` (`cause` = TypeError). This is the real connection case.

`buildRequest`'s default arm rethrows it unchanged (`if (err_ instanceof BaseError) throw err_`), so both land here and both say *"Check your connection"* — contradicting this block's own new comment. Since the vault wires `http()` exclusively (`clients/eth-contract/client.ts:22` with `retryCount: 1`, `config/wagmi.ts:100`), a provider 429/503 is the most likely RPC failure a depositor actually hits, and it's the headline case in the PR description.

`status` is a clean discriminator:

```ts
if (obj.name === "HttpRequestError") {
  // A `status` means the server answered (429 rate limit, 5xx outage) — not
  // the user's connection. Only a status-less HttpRequestError (fetch threw)
  // is a real transport failure.
  return typeof (obj as { status?: unknown }).status === "number"
    ? "rpc-error"
    : "network";
}
if (obj.name === "WebSocketRequestError" || obj.name === "SocketClosedError") {
  return "network";
}
```

viem already retries 403/408/413/429/5xx internally, so anything reaching the user here is persistent — "wait a moment and try again" is the honest copy. Note this flips the existing `"collapses HttpRequestError (RPC transport failure)"` test only if it sets a `status` (it currently doesn't), so that test stays green as the transport case.

services/vault/src/utils/errors/__tests__/formatting.test.ts
```diff
@@ -281,15 +281,18 @@ describe("Error Formatting", () => {
       expect(sanitizeErrorMessage(outer)).not.toMatch(/Network error/i);
```
gbarkhatov: **Medium — this test no longer pins the `.data` guard it was written to protect.**

The assertion is `not.toMatch(/Network error/i)`, which was load-bearing when deleting the guard flipped the result to the `network` copy. After the split, deleting the guard yields the `rpcError` copy — which also doesn't contain "Network error", so the test passes either way and the guard is now unprotected.

With the guard intact, `classifyError` returns `null` (the `0x` data sends the walker to `revertRpc.cause`, which is `undefined`) and `sanitizeErrorMessage` falls through to the outer message. Pinning that restores the protection:

```ts
expect(sanitizeErrorMessage(outer)).toBe("execution reverted");
```

services/vault/src/utils/errors/formatting.ts
```diff
@@ -174,35 +176,42 @@ export function classifyError(err: unknown): ErrorKind | null {
...
+    // "already known") whose inner frame is an RpcRequestError. None of
+    // these are the user's connection, so they get their own copy — never
+    // "check your connection".
+    //
+    // Exception: viem also reuses `RpcRequestError` to carry contract-revert
+    // data (`{ code: 3, data: "0x..." }`). That's a real revert, not an RPC
+    // failure — fall through so the underlying reason bubbles up.
+    if (obj.name === "RpcRequestError" && isViemShape(obj)) {
+      const data = (obj as { data?: unknown }).data;
+      if (typeof data === "string" && data.startsWith("0x")) {
```
gbarkhatov: **Low (optional) — data-less reverts now get "wait a moment and try again".**

`contract.ts:96-98` already documents that some RPC providers don't return revert data. Such a chain — `ContractFunctionExecutionError` → `ExecutionRevertedError` → `RpcRequestError { code: 3, data: undefined }` — misses this gate and classifies as `rpc-error`, telling the user to retry a deterministic failure while the actual reason sits one frame up on `ExecutionRevertedError`.

viem's `ExecutionRevertedError.code === 3` (`viem/_esm/errors/node.js:13-19`) is a reliable second gate that extends this branch's stated "fall through so the underlying reason bubbles up" intent to the provider class it currently misses:

```ts
const EXECUTION_REVERTED_RPC_CODE = 3; // viem ExecutionRevertedError.code
...
const isRevert =
  (typeof data === "string" && data.startsWith("0x")) ||
  obj.code === EXECUTION_REVERTED_RPC_CODE;
if (isRevert) {
  cur = obj.cause;
  continue;
}
```

Not a regression — the old path said "Network error" here — so purely a follow-up.

services/vault/src/utils/errors/formatting.ts
```diff
@@ -174,35 +176,42 @@ export function classifyError(err: unknown): ErrorKind | null {
...
       return "network";
     }
     if (obj.name === "TimeoutError" && isViemShape(obj)) {
       return "network";
     }
 
+    // `RpcRequestError` wraps a JSON-RPC error the node/provider returned:
+    // rate limit (-32005), resource unavailable (-32002), tx rejected
+    // (-32003), internal (-32603), and node-execution errors (nonce /
+    // "already known") whose inner frame is an RpcRequestError. None of
```
gbarkhatov, line 235: **Open question — `already known` isn't really a "wait and retry".**

viem folds `already known` / `transaction already imported` into `NonceTooLowError`, and that error means the tx is already in the mempool. "Please wait a moment and try again" invites a retry that re-sends and reproduces the same error. Strictly better than "check your connection", so nothing to block on.

Worth noting the node condition now has two taxonomies across two consumers: `contract.ts:313` already routes `nonce` / `replacement transaction underpriced` / `already known` to a dedicated `CONTRACT_NONCE_ERROR` bucket, while this walker lands the same condition in the generic `rpc-error`. A `NonceTooLowError` / `NonceTooHighError` name check would win here (those frames sit *above* the `RpcRequestError` frame in viem's chain) and could carry copy like "This transaction was already submitted — check your wallet or a block explorer." Out of scope for this PR, but a good candidate for the follow-up that fixes the HTTP-status case.

services/vault/src/utils/errors/formatting.ts
```diff
@@ -174,35 +176,42 @@ export function classifyError(err: unknown): ErrorKind | null {
...
+    // are the only errors where "check your connection" is honest.
+    // `HttpRequestError` / `WebSocketRequestError` / `SocketClosedError`
+    // names are viem-specific enough not to collide; `TimeoutError` is
+    // generic (AbortSignal.timeout, `ky`, DOM APIs) and must be gated on
+    // viem shape.
     if (
       obj.name === "HttpRequestError" ||
-      obj.name === "WebSocketRequestError"
+      obj.name === "WebSocketRequestError" ||
+      obj.name === "SocketClosedError"
```
gbarkhatov, line 222: Nit — `SocketClosedError` is unreachable in this app today: viem only emits it from the WebSocket/IPC transports, and the vault wires `http()` exclusively (`clients/eth-contract/client.ts:22`, `config/wagmi.ts:100`); ETH wallets arrive over EIP-1193, not viem's socket transport.

Cheap insurance and I wouldn't remove it, but the repo has an explicit zero-dead-code policy, so it's worth a word in the comment above saying this one is forward-looking (in case a WS transport lands later) rather than a live path — otherwise the next reader assumes it's covering something observed in production.

services/vault/src/utils/errors/formatting.ts
```diff
@@ -84,7 +84,8 @@ type ErrorKind =
   | "unauthorized" // EIP-1193 4100 / UnauthorizedProviderError
   | "chain-switch-failed" // EIP-1193 4902 / viem SwitchChainError
   | "receipt-timeout" // viem WaitForTransactionReceiptTimeoutError
-  | "network" // HttpRequestError / WebSocketRequestError / TimeoutError / RpcRequestError
+  | "network" // transport failure: HttpRequestError / WebSocketRequestError / SocketClosedError / viem TimeoutError
+  | "rpc-error" // node/provider JSON-RPC failure via RpcRequestError (rate limit, resource unavailable, tx rejected, internal, nonce) — not the user's connection
```
gbarkhatov: Nit — this enumerates the same five sub-cases (rate limit, resource unavailable, tx rejected, internal, nonce) that the branch comment at lines 196-201 spells out with codes. Two copies of the same list will drift the first time someone adds a code to one of them.

Suggest shortening this line to the discriminator only — e.g. `// node/provider JSON-RPC failure via RpcRequestError — not the user's connection` — and keeping the enumerated detail at the branch, matching how the other union members here stay to one short clause.

services/vault/src/utils/errors/__tests__/formatting.test.ts
```diff
@@ -325,6 +328,38 @@ describe("Error Formatting", () => {
...
+        name: "TransactionExecutionError",
+        cause: nonce,
+        walk: () => {},
+      });
+      const msg = sanitizeErrorMessage(outer);
+      expect(msg).not.toMatch(/check your connection/i);
+      expect(msg).toMatch(/wait a moment and try again/i);
+    });
+
     // The rest of the suite builds synthetic errors via Object.assign so we
```
gbarkhatov, line 410: Nit — this block exists precisely so a viem class rename fails CI loudly, yet the linchpin of the new branch (`RpcRequestError`, and now `SocketClosedError`) is only covered by `Object.assign` synthetics. A real-constructor case would close that gap and keep the block's stated rationale true:

```ts
it("matches a real viem RpcRequestError instance", () => {
  const err = new RpcRequestError({
    body: { method: "eth_call" },
    error: { code: -32603, message: "internal error" },
    url: "https://rpc.example/",
  });
  expect(sanitizeErrorMessage(err)).toMatch(/wait a moment and try again/i);
});
```

(`walk` comes from `BaseError`, so `isViemShape` passes, and `data` is `undefined` since the raw error carries none — it lands in `rpc-error` as intended.) The prose above says "the three smoke tests below" and would need bumping to four.

gbarkhatov (dismissed):
Approving — the split is correct and the comments accurately describe viem's real behavior. I checked each claim against viem 2.38.2 in the lockfile rather than taking it on faith:

- `SocketClosedError` exists with `name: 'SocketClosedError'` (`viem/_esm/errors/request.js:83`).
- `NonceTooLowError.nodeMessage = /nonce too low|transaction already imported|already known/` — the new regression test's chain is faithful, not invented.
- **User-rejection precedence survives.** `RpcRequestError`'s constructor does `this.code = error.code`, so a 4001 frame still hits the user-rejection check first within the same frame. No silent shadowing from the reordering.
- Nothing else branches on `"network"` — `classifyError` results are switched on only for `user-rejection` (`telemetryEvents.ts:94`, `erc20/transaction.ts:73`), `stale-deploy` (`lazyWithRetry.ts:44`, `global-error.tsx:15`) and `insufficient-funds` (`depositErrors.ts:223`). `Record<ErrorKind, string>` forces exhaustiveness on the new kind.

Lint clean, 50/50 tests pass. Inline comments below — the two `Medium` ones are worth a follow-up; the rest are nits and can be ignored.

### Out-of-diff: the `COPY` mock in `PostDepositContinuationView.test.tsx` wasn't updated

`services/vault/src/components/simple/__tests__/PostDepositContinuationView.test.tsx:189` carries an explicit comment that `formatting.ts` builds `FRIENDLY_MESSAGES` from `classifiedErrors` at module load, then lists every key — `userRejection`, `insufficientFunds`, `walletDisconnected`, `unauthorized`, `chainSwitchFailed`, `receiptTimeout`, `network`, `staleDeploy` — except the new `rpcError`. So inside that module graph `FRIENDLY_MESSAGES["rpc-error"]` is `undefined`. It passes today only because the view never reaches the branch, which makes it a latent trap rather than a live bug. Adding the key keeps the mock honest to its own stated contract.

jonybur (dismissed):
LGTM. The transport-vs-provider split is clean and the classifier ordering (transport names first, then `RpcRequestError` with the revert-data fall-through preserved) can't be poisoned by 4001 user-rejection being wrapped in `RpcRequestError` since 4001 is caught earlier in the walker. `SocketClosedError` unguarded by `isViemShape` is fine — the name isn't generic. Copy addition follows the `copy.ts` rule. CI green.

services/vault/src/utils/errors/formatting.ts
jrwbabylonlab: Fixed — `HttpRequestError` now routes on `status`: a numeric status (429/5xx) → `rpc-error`, status-less (fetch threw) → `network`. Added tests for both.

services/vault/src/utils/errors/__tests__/formatting.test.ts
jrwbabylonlab: Fixed — pinned to `expect(...).toBe("execution reverted")`.

services/vault/src/utils/errors/formatting.ts
jrwbabylonlab: Fixed — added `code === 3` (`EXECUTION_REVERTED_RPC_CODE`) to the revert fall-through, so a data-less revert surfaces the reason instead of the retry copy. Added a test.

services/vault/src/utils/errors/formatting.ts
jrwbabylonlab: Fixed in this PR — added an `already-submitted` bucket for `NonceTooLowError` / "already known" with copy: "This transaction was already submitted. Check your wallet or a block explorer for its status." `NonceTooHigh` stays in `rpc-error` (a gap, not a duplicate). Tests added.

services/vault/src/utils/errors/formatting.ts
jrwbabylonlab: Fixed — the comment now flags `SocketClosedError` as forward-looking (WS transport) rather than a live path.

services/vault/src/utils/errors/formatting.ts
jrwbabylonlab: Fixed — shortened the union comment to the discriminator only; the enumerated codes stay at the branch.

services/vault/src/utils/errors/__tests__/formatting.test.ts
jrwbabylonlab: Fixed — added a real `new RpcRequestError(...)` smoke test and bumped the prose to four.

jrwbabylonlab:
Also addressed the out-of-diff note from the review: added the `rpcError` (and the new `alreadySubmitted`) keys to the `PostDepositContinuationView.test.tsx` COPY mock, so `FRIENDLY_MESSAGES` stays complete in that module graph.

jonybur (approved):

PR #2156, opened 2026-07-29
fix(vault): validate, bind and durably save the recovery artifact bundle

gbarkhatov:
# Validate, bind and durably save the recovery artifact bundle

Closes https://github.com/babylonlabs-io/baby-auditor/issues/39

## Background

When someone deposits BTC, the vault provider (VP) holds the material they would need to claim their funds **without** the provider's help. The app downloads that material — a ~1.4 GB bundle — saves it to disk, and then tells the user their vault is protected.

The problem: the app had almost no basis for that claim. It asked the VP for the bundle, waited for the HTTP response to finish, wrote `"true"` to localStorage, and showed the green "Artifacts downloaded" state. Nothing verified that the response was a real bundle, that it belonged to this vault, or that a file ever reached disk.

Net effect: you could activate a vault believing you held an independent escape hatch when you didn't. If the provider later stopped cooperating, that path was gone.

## Summary

| # | What we had | The problem | What we do now |
|---|---|---|---|
| 1 | Bodies ≥ 4096 bytes skipped schema validation | Real bundles are ~1.4 GB, so validation *never* ran on a real download | Stream-validate every byte in one pass |
| 2 | A dropped connection looked like a finished download | Half a bundle was recorded as success | Parser must reach the closing brace |
| 3 | Nothing linked the bundle to your deposit | A bundle for someone else's vault passed | Check the tx graph against the request |
| 4 | An `<a download>` click, then mark it done | A cancelled save was invisible; 1.4 GB sat in memory | Stream to a user-picked file, detect cancel |
| 5 | localStorage held the string `"true"` | Couldn't tell a correct bundle from a stale one | Store a receipt bound to the pegin txid |

### 1. Large responses skipped validation entirely

**What we had.** `artifactDownloadService.ts` had a rule: if the body is 4096 bytes or larger, don't parse it — just glance at the first 64 KB to check it looks like a JSON-RPC success envelope.

**Why that's a problem.** Real bundles are ~1.4 GB, so they *always* took that branch. The schema validator only ever ran on tiny error responses. The old code documented this about itself: a VP returning `{"result": <garbage>}` passed. Worth saying clearly — this was deliberate, and for a real reason. You cannot `JSON.parse` a 1.4 GB body: turning it into one JS string blows past V8's ~512 MB string limit and freezes the tab.

**What we did.** A byte-level JSON state machine (`streamingArtifactValidator.ts`) reads the whole body in one pass while holding only a few KB of state. The small fields (`tx_graph_json`, `verifying_key_hex`, the challenger keys) are buffered normally; the huge `decryptor_artifacts_hex` values are checked character-by-character as they fly past and never held. The reconstructed small object then goes through the existing SDK validator. No size branch remains anywhere.

It also never assumes field order, which matters: the proxy only guarantees ordering on its gRPC path, which is off by default — otherwise the body is a byte-for-byte passthrough of the Rust serializer.

### 2. A truncated download counted as success

**What we had.** Nothing checked that the response was complete.

**Why that's a problem.** When the backend fails partway through, the proxy signals it by destroying the connection. You get a syntactically incomplete bundle and no error — valid-looking bytes that simply stop. The app treated that as a completed download and recorded it as such.

**What we did.** This one is free once the parser exists: it must reach the closing brace, so an incomplete document fails. No extra check needed.

### 3. Nothing tied the bundle to your vault

**What we had.** The response echoes back neither the pegin txid nor the depositor key — the only correlation field is the JSON-RPC `id`.

**Why that's a problem.** "Valid JSON of the right shape" did not mean "yours". A provider serving another deposit's bundle, or a stale cache entry, was indistinguishable from a correct response.

**What we did.** `artifactBinding.ts` checks, before anything is recorded: `claim_tx` and `payout_tx` each spend an output of the **requested** pegin transaction; `depositor_pubkey` is the depositor we asked on behalf of; and the `babe_sessions` keys are exactly the graph's own `challenger_pubkeys.local ∪ .universal`. Every expected value is a parameter of the request we just made, so this needs no on-chain read.

### 4. "Downloaded" meant "the fetch resolved"

**What we had.** `useArtifactDownload` marked the download complete as soon as the fetch resolved. The download itself was an `<a download>` click, which hands the browser a blob and returns immediately. The whole 1.4 GB was buffered in memory as a chunk array first.

**Why that's a problem.** If you cancelled the save dialog, the app never found out and still recorded success. And holding 1.4 GB in memory is its own risk.

**What we did.** The body now streams straight to a user-chosen file via the File System Access API. That API writes to a temporary swap file and only commits on close, so a validation failure mid-download leaves nothing behind, and a cancelled save is detectable. The write is awaited, so disk I/O applies backpressure to the network and the body never accumulates in memory. Browsers without the API (Firefox, Safari) fall back to the old anchor download plus an explicit "I saved the file" confirmation, with a size ceiling instead of an out-of-memory tab kill.

Because `showSaveFilePicker` needs transient user activation, the picker has to open before the wallet signature prompt — there's a comment in the hook saying so, since it looks reorderable and isn't.

### 5. The stored flag couldn't tell bundles apart

**What we had.** `tbv:artifacts-downloaded:<vaultId>` held the string `"true"`.

**Why that's a problem.** That records "something happened once for this vault" — not which bundle, or whether it was the right one. A stale record, or one from a different deposit, counted just the same.

**What we did.** We store `{ peginTxid, sha256, byteLength, filename, savedAt, method }`, written only from a validated and saved bundle. `hasArtifactsDownloaded(vaultId, peginTxid)` returns true only for a matching record. The activation modal's existing `(downloaded || acknowledged)` gate then keeps the risk checkbox until a real bound receipt exists — that logic needed no change, just the extra argument.

### Bonus: god mode can now test all of this

**What we had.** The artifact mock ticked a progress bar and touched none of the real code.

**Why that's a problem.** All the interesting logic now lives in the validator and the file sink, so a mock that skips both tests nothing.

**What we did.** The mock feeds a synthetic response through the *real* pipeline, with a scenario selector: valid bundle / truncated stream / junk in a success envelope / padded JSON-RPC error / corrupt hex / bundle for a different deposit. Plus a "Clear artifact receipts" button, because once a vault's gate is satisfied the card hides its download button and there was no way to re-run the flow. A simulated download still writes no receipt, so it can never satisfy a real vault's gate.

## Approaches considered

**Validating a 1.4 GB body.** Three options: keep the size threshold (status quo — leaves the hole open); parse it in a Web Worker (keeps the UI responsive but does not beat V8's string limit, so it does not actually work); or a streaming parser. Only the third is possible, so that is what shipped. Measured throughput is **255 MB/s** — the full 1.4 GB validates in 5.5 s, about 2% of a 4.5-minute download, so the Web Worker turned out to be unnecessary anyway.

**Binding — and a correction worth flagging.** I initially concluded this was *impossible* and wrote it up as such. The reasoning: `tx_graph_json` is opaque — no schema, no golden file, no sample anywhere in babylon-toolkit, vault-provider-proxy or app-babylon-vault, and the proxy's proto explicitly calls it "serialized at the Rust boundary". Every example in every repo is an invented placeholder like `{"nodes":[],"edges":[]}`.

Then we did a real download and looked at the file. The graph has 14 top-level keys, and its claim and payout transactions spend the pegin outpoint in display order. **It was knowable all along — just not from the repos.** So binding is implemented rather than deferred, and it is cheaper than planned: because the expected values come from the request, none of the on-chain reader plumbing originally budgeted for is needed.

**Challenger-set binding against on-chain data** was the alternative, and is *not* implemented. The expected challenger set is public on-chain data, so a malicious VP reads the same contracts and puts the correct keys on a decoy bundle for the cost of one call — it stops no real attacker, while a wrong assumption about the set would hard-block honest users from their recovery bundle. Instead we check the set for internal consistency against the graph's own declaration, which catches the case that actually matters (a missing session leaving recovery material incomplete) at no cost and no false-reject risk.

**Detecting a cancelled save.** Either attestation everywhere (a checkbox — simple, but weak evidence and keeps the memory problem), or File System Access with attestation as the fallback. We took the second: it is the only way to get a real save/cancel signal, and it fixes the memory issue at the same time. Chromium-only is acceptable here — Chromium is the only browser this repo e2e-tests and every supported wallet is a Chrome extension — and the fallback warns the user up front.

**Old localStorage flags** are invalidated rather than migrated. The whole point is that the old signal was not evidence, so carrying it forward would defeat the change. Affected users see the artifact warning again and can re-download. Note `JSON.parse("true")` returns a boolean rather than throwing, so this needed an explicit shape check — without it every legacy flag would have silently survived.

## What this does not do

Every value checked is public, so a determined provider can satisfy all of it while returning garbage payloads. The `decryptor_artifacts_hex` bytes are charset- and length-validated, **never proven to decrypt correctly** — that needs a BaBe verifier, which does not exist client-side. What this catches is a broken or careless provider, a bundle for the wrong deposit, a stale cache, a dropped connection, and a save that never happened. Not a targeted forgery.

## Testing

- `pnpm --filter vault run test` — 2929 passing. New: 50 parser tests (key-order permutations, escapes and multi-byte characters, truncation, malformed numbers, duplicate keys, and a case proving 1-byte chunking gives an identical verdict), 13 binding tests, plus service, storage, hook and modal coverage.
- **Verified against the real 1.4 GB bundle**: it passes the validator and the binding, and the same data with a foreign pegin txid is rejected.
- **Manual, in Chrome against devnet**: save dialog appears before the wallet prompt, file lands at the chosen path, receipt written with `method: "file-system-access"`. Cancelling the picker persists nothing.
- **Hostile paths** via the god-mode scenario selector: each one rejected, nothing committed, no receipt, risk checkbox still required.

The real bundle also grounded three things that had been guesses: the parser's size caps (measured 1.79 MB tx graph, 528-char verifying key, 8 sessions), the CPU cost above, and a **bug it exposed** — the progress bar's fallback estimate was 1,395,864,371 bytes against a real 1,410,824,702-byte body, so it saturated before the download finished.

## Notes for reviewers

This is critical-path recovery-material code and the parser is the risky part in both directions: a bug that falsely rejects locks honest users out of their bundle, and one that falsely accepts waves a bad bundle through. Worth two reviewers.

Two behaviour changes to be aware of:

- Error envelopes now always surface as `JsonRpcError`, including padded ones that previously became `VpResponseValidationError`. This is also a fix: the auth-expired retry only triggers on `JsonRpcError`, so a padded `auth_expired` used to defeat it.
- In `ActivateConfirmationModal`, `peginTxid` is optional. A vault without one now always reads as not-downloaded and requires the acknowledgement. That is the correct fail-closed reading, and it matches `canRenderCard`, which is already false in exactly that case.

Two things still open:

- Binding reads `claim_tx`, `payout_tx`, `depositor_pubkey` and `challenger_pubkeys` by name. A tx-graph schema change would falsely reject, so these should be treated as a compatibility surface and confirmed against tx-graph v2.
- The claim that `FileSystemWritableFileStream.write()` applies real backpressure rather than buffering held up on a real download, but has not been deliberately memory-profiled.

### DIFF
--- a/services/vault/src/hooks/deposit/useArtifactDownload.ts
+++ b/services/vault/src/hooks/deposit/useArtifactDownload.ts
@@ -54,12 +54,6 @@ interface ArtifactDownloadState {
   progress: string;
   error: string | null;
   downloaded: boolean;
-  /**
-   * True after a fallback (anchor) download fired. That path gives no
-   * save/cancel signal from the browser, so the receipt is withheld until the
-   * user confirms via `confirmSaved`.
-   */
-  awaitingSaveConfirmation: boolean;
   /** Bytes received so far from the in-flight artifact stream. */
   receivedBytes: number;
   /**
@@ -75,16 +69,9 @@ const INITIAL_STATE: ArtifactDownloadState = {
   progress: "",
   error: null,
   downloaded: false,
-  awaitingSaveConfirmation: false,
   receivedBytes: 0,
   totalBytes: 0,
 };
-
-/** A validated, saved bundle waiting to be turned into a stored receipt. */
-interface PendingReceipt {
-  peginTxid: string;
-  outcome: ArtifactDownloadOutcome;
-}
 
 interface PrimeContext {
   vaultId: Hex;
@@ -134,17 +121,13 @@ export function useArtifactDownload(options?: {
   // instead of letting it run to completion in the background.
   const abortControllerRef = useRef<AbortController | null>(null);
 
-  // Held outside React state so `confirmSaved` can persist it without a
-  // side effect inside a state updater.
-  const pendingReceiptRef = useRef<PendingReceipt | null>(null);
-
   /**
    * Turn a completed download into the stored receipt that the activation
    * gate reads. Only ever called with an outcome from a validated, saved
    * bundle — never from a fetch that merely resolved.
    */
   const persistReceipt = useCallback(
-    ({ peginTxid, outcome }: PendingReceipt) => {
+    (peginTxid: string, outcome: ArtifactDownloadOutcome) => {
       if (!vaultId) return;
       // Milestone only on the first real download — re-downloads are a
       // routine user action, not funnel progress.
@@ -197,7 +180,6 @@ export function useArtifactDownload(options?: {
       const isStale = () =>
         abortController.signal.aborted ||
         abortControllerRef.current !== abortController;
-      pendingReceiptRef.current = null;
       setState({
         ...INITIAL_STATE,
         loading: true,
@@ -408,15 +390,18 @@ export function useArtifactDownload(options?: {
           }
 
           if (outcome.method === "browser-download") {
-            // The anchor path gives no save/cancel signal, so the file may or
-            // may not have reached disk. Hold the receipt until the user says
-            // it did — see `confirmSaved`.
-            pendingReceiptRef.current = { peginTxid, outcome };
-            setState({ ...INITIAL_STATE, awaitingSaveConfirmation: true });
+            // The anchor path only proves a link was clicked: the browser
+            // reports nothing about whether the file reached disk, and it may
+            // have been blocked or the save dialog dismissed. Asking the user
+            // to attest would let a click satisfy the activation gate, which
+            // is the exact substitution this flow exists to remove. Show the
+            // file as delivered, but withhold the receipt so the vault keeps
+            // warning until there is real evidence.
+            setState({ ...INITIAL_STATE, downloaded: true });
             return;
           }
 
-          persistReceipt({ peginTxid, outcome });
+          persistReceipt(peginTxid, outcome);
           setState({
             ...INITIAL_STATE,
             downloaded: true,
@@ -527,27 +512,12 @@ export function useArtifactDownload(options?: {
 
   const cancel = useCallback(() => {
     abortControllerRef.current?.abort();
-    pendingReceiptRef.current = null;
     setState(INITIAL_STATE);
   }, []);
-
-  /**
-   * Record that the user saved the file on a browser that cannot tell us
-   * itself. Attestation is weaker evidence than a confirmed write, but on the
-   * anchor-download path it is the only evidence available.
-   */
-  const confirmSaved = useCallback(() => {
-    const pending = pendingReceiptRef.current;
-    if (!pending) return;
-    pendingReceiptRef.current = null;
-    persistReceipt(pending);
-    setState({ ...INITIAL_STATE, downloaded: true });
-  }, [persistReceipt]);
 
   return {
     ...state,
     download,
     cancel,
-    confirmSaved,
   };
 }
--- a/services/vault/src/hooks/deposit/useArtifactDownload.ts
+++ b/services/vault/src/hooks/deposit/useArtifactDownload.ts
@@ -53,7 +53,18 @@ interface ArtifactDownloadState {
   loading: boolean;
   progress: string;
   error: string | null;
+  /**
+   * A validated bundle reached disk and a receipt was written. This is the
+   * only success signal the activation gate accepts, so it must never be set
+   * from a path that cannot prove the file was saved.
+   */
   downloaded: boolean;
+  /**
+   * The file was handed to the browser but we cannot prove it was saved (the
+   * anchor fallback, or a mocked download). Drives the card's "we can't
+   * confirm this" state and nothing else — it must never satisfy the gate.
+   */
+  delivered: boolean;
   /** Bytes received so far from the in-flight artifact stream. */
   receivedBytes: number;
   /**
@@ -69,6 +80,7 @@ const INITIAL_STATE: ArtifactDownloadState = {
   progress: "",
   error: null,
   downloaded: false,
+  delivered: false,
   receivedBytes: 0,
   totalBytes: 0,
 };
@@ -383,24 +395,27 @@ export function useArtifactDownload(options?: {
 
           // No outcome means a mocked download: nothing was validated and no
           // file was written, so it must not satisfy the real activation gate.
-          // The demo still shows the downloaded UI.
+          // `delivered` drives the demo UI without ever claiming proof.
           if (!outcome) {
-            setState({ ...INITIAL_STATE, downloaded: true });
+            setState({ ...INITIAL_STATE, delivered: true });
             return;
           }
 
           if (outcome.method === "browser-download") {
             // The anchor path only proves a link was clicked: the browser
             // reports nothing about whether the file reached disk, and it may
-            // have been blocked or the save dialog dismissed. Asking the user
-            // to attest would let a click satisfy the activation gate, which
-            // is the exact substitution this flow exists to remove. Show the
-            // file as delivered, but withhold the receipt so the vault keeps
-            // warning until there is real evidence.
-            setState({ ...INITIAL_STATE, downloaded: true });
+            // have been blocked or the save dialog dismissed. So this reports
+            // `delivered`, not `downloaded` — the card says the save could not
+            // be confirmed, the risk acknowledgement stays required, and the
+            // vault keeps warning until there is real evidence. Setting
+            // `downloaded` here would let a click satisfy the activation gate,
+            // which is the exact substitution this flow exists to remove.
+            setState({ ...INITIAL_STATE, delivered: true });
             return;
           }
 
+          // The one path with real evidence: validated, streamed to a file the
+          // user chose, and receipted.
           persistReceipt(peginTxid, outcome);
           setState({
             ...INITIAL_STATE,
--- a/services/vault/src/hooks/deposit/useArtifactDownload.ts
+++ b/services/vault/src/hooks/deposit/useArtifactDownload.ts
@@ -24,11 +24,19 @@ import {
 import { isPreDepositorSignaturesError } from "@/models/peginStateMachine";
 import {
   ArtifactDownloadCancelledError,
+  type ArtifactDownloadOutcome,
+  ArtifactDownloadTooLargeError,
+  ArtifactFileAccessError,
+  type ArtifactSaveTarget,
   fetchAndDownloadArtifacts,
+  type FetchArtifactsOptions,
+  openArtifactSaveTarget,
 } from "@/services/artifacts";
 import {
+  ARTIFACT_RECEIPT_VERSION,
   hasArtifactsDownloaded,
-  markArtifactsDownloaded,
+  normalizePeginTxid,
+  saveArtifactDownloadReceipt,
 } from "@/utils/artifactDownloadStorage";
 
 const ARTIFACT_RETRY_INTERVAL_MS = 10_000;
@@ -46,6 +54,12 @@ interface ArtifactDownloadState {
   progress: string;
   error: string | null;
   downloaded: boolean;
+  /**
+   * True after a fallback (anchor) download fired. That path gives no
+   * save/cancel signal from the browser, so the receipt is withheld until the
+   * user confirms via `confirmSaved`.
+   */
+  awaitingSaveConfirmation: boolean;
   /** Bytes received so far from the in-flight artifact stream. */
   receivedBytes: number;
   /**
@@ -61,9 +75,16 @@ const INITIAL_STATE: ArtifactDownloadState = {
   progress: "",
   error: null,
   downloaded: false,
+  awaitingSaveConfirmation: false,
   receivedBytes: 0,
   totalBytes: 0,
 };
+
+/** A validated, saved bundle waiting to be turned into a stored receipt. */
+interface PendingReceipt {
+  peginTxid: string;
+  outcome: ArtifactDownloadOutcome;
+}
 
 interface PrimeContext {
   vaultId: Hex;
@@ -113,8 +134,59 @@ export function useArtifactDownload(options?: {
   // instead of letting it run to completion in the background.
   const abortControllerRef = useRef<AbortController | null>(null);
 
+  // Held outside React state so `confirmSaved` can persist it without a
+  // side effect inside a state updater.
+  const pendingReceiptRef = useRef<PendingReceipt | null>(null);
+
+  /**
+   * Turn a completed download into the stored receipt that the activation
+   * gate reads. Only ever called with an outcome from a validated, saved
+   * bundle — never from a fetch that merely resolved.
+   */
+  const persistReceipt = useCallback(
+    ({ peginTxid, outcome }: PendingReceipt) => {
+      if (!vaultId) return;
+      // Milestone only on the first real download — re-downloads are a
+      // routine user action, not funnel progress.
+      const firstDownload = !hasArtifactsDownloaded(vaultId, peginTxid);
+      saveArtifactDownloadReceipt(vaultId, {
+        version: ARTIFACT_RECEIPT_VERSION,
+        peginTxid: normalizePeginTxid(peginTxid),
+        filename: outcome.filename,
+        byteLength: outcome.byteLength,
+        sha256: outcome.sha256,
+        savedAt: Date.now(),
+        method: outcome.method,
+      });
+      if (firstDownload) {
+        logger.event(TELEMETRY_EVENT.ACTIVATION_ARTIFACTS_DOWNLOADED, {
+          level: "info",
+          category: "activation",
+          tags: { vaultId: shortId(vaultId) },
+        });
+      }
+    },
+    [vaultId],
+  );
+
   const download = useCallback(
     async (providerAddress: string, peginTxid: string, depositorPk: string) => {
+      // Dev/QA-only simulation of the artifact fetch (never reaches a VP),
+      // so the dialogs' progress states can be exercised. Opted into via the
+      // god-mode panel's "Mock artifact download" toggle; off in production
+      // builds, where the god-mode gate is compile-time false.
+      const demoDownload = isArtifactDownloadDemoEnabled();
+
+      // DO NOT REORDER: showSaveFilePicker() requires transient user
+      // activation, which any preceding `await` destroys. Opening the save
+      // target has to be the first thing this handler does, while the click
+      // that triggered it is still the active gesture. The promise is awaited
+      // further down, once the synchronous setup is out of the way.
+      //
+      // The demo path opens a real save target too — it writes a real (if
+      // synthetic) file through the real pipeline, and only skips the network.
+      const saveTargetPromise = openArtifactSaveTarget(peginTxid);
+
       // Each invocation owns its controller, and staleness is derived from
       // it rather than a shared flag: a cancelled flow parked in a
       // non-abortable await (wallet prime, retry sleep) stays permanently
@@ -125,10 +197,13 @@ export function useArtifactDownload(options?: {
       const isStale = () =>
         abortController.signal.aborted ||
         abortControllerRef.current !== abortController;
+      pendingReceiptRef.current = null;
       setState({
         ...INITIAL_STATE,
         loading: true,
-        progress: COPY.deposit.recoveryArtifacts.fetchingArtifacts,
+        // The save dialog is already open at this point, so the first status
+        // the user sees names it rather than the fetch behind it.
+        progress: COPY.deposit.recoveryArtifacts.choosingSaveLocation,
       });
 
       const normalizedPeginTxid = stripHexPrefix(peginTxid);
@@ -138,15 +213,6 @@ export function useArtifactDownload(options?: {
       // deposit and is public on-chain data (shortened before emission anyway).
       const telemetryVaultId = vaultId ?? normalizedPeginTxid;
 
-      // Dev/QA-only simulation of the artifact fetch (never reaches a VP),
-      // so the dialogs' progress states can be exercised. Opted into via the
-      // god-mode panel's "Mock artifact download" toggle; off in production
-      // builds, where the god-mode gate is compile-time false.
-      const demoDownload = isArtifactDownloadDemoEnabled();
-      const fetchArtifacts = demoDownload
-        ? demoFetchAndDownloadArtifacts
-        : fetchAndDownloadArtifacts;
-
       // Stop the flow with an error message. Used by every fail path
       // below so the rendered modal state stays consistent.
       const setError = (message: string) =>
@@ -154,6 +220,56 @@ export function useArtifactDownload(options?: {
           ...INITIAL_STATE,
           error: message,
         });
+
+      // Resolve the save location before priming, so the user answers the
+      // file dialog and the wallet prompt in that order rather than both at
+      // once.
+      let saveTarget: ArtifactSaveTarget;
+      try {
+        saveTarget = await saveTargetPromise;
+      } catch (err) {
+        if (err instanceof ArtifactDownloadCancelledError) {
+          setState(INITIAL_STATE);
+          return;
+        }
+        if (isStale()) return;
+        captureFunnelFailure(
+          TELEMETRY_STAGE.ACTIVATION_ARTIFACTS,
+          err,
+          telemetryVaultId,
+          { tags: { site: "save_target" } },
+        );
+        setError(
+          err instanceof Error
+            ? err.message
+            : COPY.deposit.recoveryArtifacts.fileAccessDenied,
+        );
+        return;
+      }
+      if (isStale()) return;
+      setState((prev) => ({
+        ...prev,
+        progress: COPY.deposit.recoveryArtifacts.fetchingArtifacts,
+      }));
+
+      // The demo yields no outcome, which is what keeps a simulated download
+      // from ever writing the receipt that satisfies the activation gate.
+      const runDownload = (
+        fetchOptions: FetchArtifactsOptions,
+      ): Promise<ArtifactDownloadOutcome | null> =>
+        demoDownload
+          ? demoFetchAndDownloadArtifacts(
+              saveTarget,
+              { peginTxid: normalizedPeginTxid, depositorPk },
+              fetchOptions,
+            ).then(() => null)
+          : fetchAndDownloadArtifacts(
+              providerAddress,
+              peginTxid,
+              depositorPk,
+              saveTarget,
+              fetchOptions,
+            );
 
       // Ensure the bearer is in cache before any artifact request. The
       // RPC is auth-gated server-side (AUTH_GATED_METHODS), so a
@@ -268,7 +384,7 @@ export function useArtifactDownload(options?: {
         if (isStale()) return;
 
         try {
-          await fetchArtifacts(providerAddress, peginTxid, depositorPk, {
+          const outcome = await runDownload({
             onProgress: (receivedBytes, totalBytes) => {
               // Drop progress events that arrive after cancel — they would
               // otherwise re-show the bar after the UI has reset.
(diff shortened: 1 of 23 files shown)


services/vault/src/services/artifacts/artifactBinding.ts
```diff
@@ -0,0 +1,162 @@
...
+): void {
+  const peginTxid = normalizeHex(expected.peginTxid);
+
+  for (const field of PEGIN_SPENDING_TX_KEYS) {
+    const txids = inputTxids(txGraph[field], field);
+    if (!txids.includes(peginTxid)) {
+      throw new VpResponseValidationError(
+        `Artifact bundle is for a different deposit: "${field}" spends ${txids.join(", ")}, expected ${peginTxid}`,
+      );
+    }
```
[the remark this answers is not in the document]
gbarkhatov: This one does not hold — it conflates the Pre-PegIn with the pegin transaction.

`claim_tx` and `payout_tx` do not spend the Pre-PegIn. They spend outputs of the **pegin** transaction. The Pre-PegIn is the shared one that carries a per-sibling HTLC output, and it appears at `pegin_tx.input[0]`, which binding never looks at.

Confirmed against a real vault-devnet bundle:

```
pegin_tx.tx.input[0].previous_output   a18c1d8f…:0   <- Pre-PegIn (shared across siblings)
claim_tx.tx.input[0].previous_output   f545b4a3…:1   <- pegin tx (per vault)
payout_tx.tx.input[0].previous_output  f545b4a3…:0   <- pegin tx (per vault)
```

`f545b4a3…` is exactly the `pegin_txid` the request was made with.

Siblings cannot share that txid. From `services/vault/src/clients/eth-contract/btc-vault-registry/query.ts:59`:

```
vaultId = keccak256(abi.encode(peginTxHash, depositor))
```

Two distinct vaults for the same depositor sharing a pegin txid would hash to the same `vaultId` — they would be the same vault. So a sibling has a different pegin txid and fails the check on the txid alone; the vout is not load-bearing.

Put another way: the RPC parameters (`pegin_txid` + `depositor_pk`) *are* the vault’s unique key, which is why binding to both pins the vault.

I also considered asserting the vout anyway as cheap hardening and decided against it: there is no independently known expected value to compare against (`htlcVout` indexes the Pre-PegIn, not the pegin), so it would assert the graph against itself while adding false-reject risk if output ordering ever changes.

No code change.

services/vault/src/services/artifacts/artifactDownloadService.ts
```diff
@@ -233,291 +201,80 @@ async function readBodyWithProgress(
...
+    await discardQuietly(stream);
+    throw err;
+  } finally {
+    reader.releaseLock();
   }
-}
 
-/** JSON whitespace per RFC 8259: space, tab, line feed, carriage return. */
-const JSON_WHITESPACE = new Set([" ", "\t", "\n", "\r"]);
+  await stream.commit();
```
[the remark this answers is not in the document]
gbarkhatov: Good catch — real, and fixed in db8f1990.

`commit()` is now wrapped, and a rejected `close()` abandons the writable explicitly rather than leaving the transaction open with its temporary file intact:

```ts
try {
  await stream.commit();
} catch (err) {
  // A rejected close leaves the writable un-aborted, so its temporary file
  // can linger until the browser reclaims it. Abandon the transaction
  // explicitly rather than relying on that.
  await discardQuietly(stream);
  throw err;
}
```

`discardQuietly` swallows and logs a failing abort, so cleanup can never mask the original error.

Added a regression test — "discards the writable when the final commit fails" — asserting `discard()` fires when `commit()` rejects.

services/vault/src/services/artifacts/streamingArtifactValidator.ts
```diff
@@ -0,0 +1,993 @@
...
+      babe_sessions: this.babeSessionsSeen ? babeSessions : undefined,
+    };
+
+    validateRequestDepositorClaimerArtifactsResponse(skeleton);
+
+    const txGraph = parseTxGraphJson(skeleton.tx_graph_json);
+
+    const sessionHexLengths: Record<string, number> = {};
+    for (const [pubkey, hex] of this.sessionHex) {
+      sessionHexLengths[pubkey] = hex.length;
```
jrwbabylonlab: Hex lengths recorded but never checked — 2-char artifacts pass and get the downloaded receipt.

services/vault/src/hooks/deposit/useArtifactDownload.ts
```diff
@@ -282,23 +398,25 @@ export function useArtifactDownload(options?: {
...
+          if (!outcome) {
+            setState({ ...INITIAL_STATE, downloaded: true });
+            return;
+          }
+
+          if (outcome.method === "browser-download") {
+            // The anchor path gives no save/cancel signal, so the file may or
+            // may not have reached disk. Hold the receipt until the user says
+            // it did — see `confirmSaved`.
+            pendingReceiptRef.current = { peginTxid, outcome };
```
jrwbabylonlab: Checkbox click persists the receipt even if the download was blocked or cancelled.

services/vault/src/services/artifacts/artifactSaveTarget.ts
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Ceiling for the in-memory fallback path.
+ *
+ * A full-size bundle (~1.3 GB) cannot be held as a Blob without killing the
+ * tab, so the fallback refuses anything above this and the caller surfaces
+ * copy pointing at a Chromium browser. Failing loudly beats an out-of-memory
+ * crash or a silently truncated recovery bundle.
+ */
+const FALLBACK_MAX_BODY_BYTES = 512 * 1024 * 1024;
```
jrwbabylonlab: 512 MiB cap vs a ~1.4 GB bundle — always fails on Safari/Firefox. Block up front?

services/vault/src/services/artifacts/artifactDownloadService.ts
```diff
@@ -1,52 +1,69 @@
...
+ *   is skipped on the basis of size.
+ * - **Success means the bytes are durably saved**, not that a fetch resolved.
+ *   The body is piped into an {@link ArtifactSaveTarget} and only committed
+ *   after the validator has seen the closing brace.
+ *
+ * Bytes are validated, hashed, and written chunk-by-chunk in a single pass,
+ * so the full body is never held in memory on the File System Access path.
+ *
+ * What this does NOT establish: that the artifact bytes decrypt correctly, or
+ * that the bundle belongs to this vault. The response carries no pegin txid or
```
jrwbabylonlab: Stale: `tx_graph_json` does carry pegin txid + depositor key, and this PR binds against them.

jonybur (commented):
The parser held up under adversarial reading in both directions - I traced chunk-boundary handling, duplicate keys, truncation, and the error-envelope path and found no false-accept or false-reject route - and the receipt can only be constructed after finish() plus binding plus commit(). Two things in `artifactDownloadService.ts` should be fixed before merge:

1. The rewritten module header still denies the binding the module now performs. The "What this does NOT establish" paragraph says the response "carries no pegin txid or depositor key to bind against" and that "a provider returning a well-formed bundle for the wrong vault is not detected here" - but `downloadArtifactsFromResponse` calls `assertBundleBoundToVault`, and the binding tests reject a foreign pegin txid. This reads like a leftover from the earlier draft where binding was judged impossible. On recovery-material code, a header claiming the wrong-vault check does not exist is an invitation for a future edit to drop or bypass the call; the paragraph should claim only what is actually not established (decryption correctness, targeted forgery by a provider that reads the same public data).

2. Early rejection never cancels the response body. In the read loop, `reader.cancel()` runs only on the explicit user-cancel branch; when the validator throws mid-stream (error envelope, junk prefix, `ArtifactDownloadTooLargeError`) or binding fails after `finish()`, the catch discards the save stream and rethrows while the `finally` only does `releaseLock()`. The fetch then keeps pulling the remainder of a ~1.4 GB body in the background after the verdict is already in, holding bandwidth and a connection while the user sees the failure and possibly retries on top of it. The failure path deserves the same `await reader.cancel().catch(() => undefined)` treatment as the cancel branch.

services/vault/src/services/artifacts/streamingArtifactValidator.ts
gbarkhatov: Right on both counts — the value was unused *and* the check it enabled was missing.

Added a floor enforced at session close, next to the existing odd-length check:

```ts
export const MIN_DECRYPTOR_HEX_CHARS = 64 * 1024;
```

Charset, non-emptiness and even length are all satisfied by `"ab"`, so without this a provider could return a bundle that is structurally perfect and materially empty — and earn a receipt for it. A measured devnet session carries ~176,129,340 characters, so the floor sits ~2700x below that: room for a smaller circuit, no room for a token payload.

`sessionHexLengths` is deleted rather than wired up — nothing in production read it, and the floor is now enforced inline where the count already lives, so returning it would just be dead surface. The per-session `{length, preview}` record collapsed to a plain prefix string too.

Tests: rejects `"ab"`, rejects one character below the floor, and rejects when only one of several sessions is undersized. Fixtures build their hex from the exported constant so moving the floor does not need a sweep. Re-ran the real 1.4 GB bundle end-to-end — still passes.

services/vault/src/hooks/deposit/useArtifactDownload.ts
gbarkhatov: Agreed, and I went further than hardening it — **the attestation is removed entirely, and the fallback path no longer writes a receipt at all.**

There is no way to harden it. On that path `commit()` only means "we clicked a link"; the browser reports nothing about whether the file reached disk, so a blocked download or a dismissed save dialog is indistinguishable from success. Keeping the checkbox meant a click satisfying the activation gate — which is the exact substitution this whole PR set out to remove. Every other path here earns its receipt from evidence.

Now: the file still downloads and the card shows it as delivered, but no receipt is written, so the vault keeps showing the artifact warning and the user proceeds via the existing risk acknowledgement. That is the honest state given what we actually know.

Deleted `confirmSaved`, `pendingReceiptRef`, `awaitingSaveConfirmation`, the checkbox in `RecoveryArtifactsCard`, and `COPY…saveConfirmation`. The two attestation tests collapse into one asserting the fallback never writes a receipt.

Trade-off worth naming: a Firefox/Safari user who really did save the file still sees the warning. Fail-closed is the right direction here, and the `fallbackSaveHint` copy now says so up front so it is not a surprise.

services/vault/src/services/artifacts/artifactSaveTarget.ts
gbarkhatov: Real, and worse than "always fails" — it fails only *after* streaming 512 MB of the user's bandwidth. Fixed, but not by blocking, and the reason is worth recording.

**I checked whether blocking would strand anyone, and it would.** I had assumed non-Chromium users cannot reach this screen because every BTC wallet is a Chrome extension. That is wrong: `packages/babylon-wallet-connector/src/core/wallets/btc/index.ts:22` registers `keystone`, which connects by animated QR through `@keystonehq/sdk` — no injected `window` global, no WebUSB/WebHID, only `getUserMedia`, which Firefox and Safari support. It is not in the vault's disabled list (`VaultWalletConnectionProvider.tsx:31-41` disables AppKit, injectable and both Ledger variants — not Keystone), and it implements `deriveContextHash`, so deposits work through it. Blocking up front would cut those users off from their own recovery material.

So instead:

- `FALLBACK_MAX_BODY_BYTES` raised from 512 MiB to **2 GiB**, above the measured 1,410,824,702-byte body, so the path is no longer guaranteed to fail.
- `fallbackSaveHint` now states plainly, before the download starts, that this browser must hold the entire file in memory, that it may run out on a smaller device, that the save cannot be confirmed so the warning will persist, and that a Chromium browser is recommended.

Honest limitation: this is best-effort, not reliable — the chunk array and the resulting Blob can both be resident, so a full-size bundle may still exhaust memory. The durable fix is a streaming sink that works outside Chromium: OPFS via a worker `createSyncAccessHandle`, supported in Firefox 111+ and Safari 15.2+. That is its own piece of work and I would rather file it than bolt it onto this PR — happy to open the issue if you agree.

services/vault/src/services/artifacts/artifactDownloadService.ts
gbarkhatov: Correct — that paragraph predates the binding work and was never updated, so the file documented the opposite of what it does. Rewritten:

```
 * The bundle is also bound to the deposit it was requested for: the envelope
 * names no vault, but the transaction graph inside it does, so its claim and
 * payout transactions are checked against the requested pegin txid and its
 * depositor key against the requested one. See `artifactBinding.ts`.
 *
 * What this does NOT establish: that the artifact bytes decrypt correctly. No
 * client-side BaBe verifier exists, so the payload is proven well-formed and
 * addressed to this deposit, never proven usable.
```

gbarkhatov:
All four review comments addressed in `8e39d0cf` (details in each resolved thread).

| Comment | Resolution |
|---|---|
| Hex lengths recorded but never checked | `MIN_DECRYPTOR_HEX_CHARS = 64 KiB` enforced at session close; `sessionHexLengths` deleted — nothing read it |
| Checkbox persists the receipt regardless | Attestation removed entirely; the fallback path now writes **no** receipt |
| 512 MiB cap vs a ~1.4 GB bundle | Raised to 2 GiB + explicit up-front warning. **Not** blocked — see below |
| Stale `tx_graph_json` comment | Rewritten to describe the binding the PR added |

2932 tests passing, lint and typecheck clean. The real 1.4 GB devnet bundle was re-run through the validator and binding to confirm the new floor does not reject it.

**One correction worth surfacing.** The "block up front" suggestion was not taken because it would strand real users, and my earlier reasoning about that was wrong. I had assumed non-Chromium users cannot reach this screen since every BTC wallet is a Chrome extension. `keystone` disproves it — animated-QR transport, no injected global, no WebUSB/WebHID, only `getUserMedia`, which Firefox and Safari support. It is not in the vault's disabled list and implements `deriveContextHash`, so deposits work through it. Blocking would cut those users off from their own recovery material.

**Two things still open, both stated rather than papered over:**

- The 2 GiB fallback is best-effort, not reliable — the chunk array and the resulting Blob can both be resident, so a full-size bundle may still exhaust memory on a constrained device. The durable fix is an OPFS streaming sink via a worker `createSyncAccessHandle` (Firefox 111+, Safari 15.2+). Happy to file that as a follow-up.
- The hex floor is a new false-reject surface. 64 KiB sits ~2700x below the measured ~176 M characters per session, so the margin is wide, but it would hard-fail a download if a future circuit produced dramatically smaller artifacts. Worth a sanity check with whoever owns the BaBe side.

Note on the Greptile review above: its 2/5 score is driven by a binding finding it later **retracted** in-thread ("My original comment was wrong. No change needed."). Its other finding was real and is fixed in `db8f1990`.

services/vault/src/hooks/deposit/useArtifactDownload.ts
```diff
@@ -282,23 +380,28 @@ export function useArtifactDownload(options?: {
...
+
+          if (outcome.method === "browser-download") {
+            // The anchor path only proves a link was clicked: the browser
+            // reports nothing about whether the file reached disk, and it may
+            // have been blocked or the save dialog dismissed. Asking the user
+            // to attest would let a click satisfy the activation gate, which
+            // is the exact substitution this flow exists to remove. Show the
+            // file as delivered, but withhold the receipt so the vault keeps
+            // warning until there is real evidence.
+            setState({ ...INITIAL_STATE, downloaded: true });
```
jrwbabylonlab: Withholding the receipt isn't enough — `downloaded: true` still satisfies the gate.

`RecoveryArtifactsCard.tsx:160` fires `onDownloaded` on this path → `ActivateConfirmationModal.tsx:162` sets `downloaded` → checkbox removed at `:167`, Activate enabled at `:91`.

So on Firefox/Safari a blocked or cancelled save unlocks activation for the session, and `copy.ts:522` promises the opposite. Same for the mock at `:388`.

`ActivateConfirmationModal.test.tsx:215` pins the current behaviour, so it won't catch this.

services/vault/src/services/artifacts/streamingArtifactValidator.ts
```diff
@@ -0,0 +1,995 @@
...
+ * universal); 1024 is far above any plausible signer set and exists only to
+ * stop an unbounded session map being used as a memory sink.
+ */
+const MAX_BABE_SESSIONS = 1024;
+
+/**
+ * A JSON-RPC error object is small by construction. Anything larger is not a
+ * plausible honest error response, so we fail closed rather than buffer it.
+ */
+const MAX_ERROR_VALUE_BYTES = 64 * 1024;
```
jrwbabylonlab: An error object over this cap throws `VpResponseValidationError` from `CaptureBuffer.push` before `throwWireError` can raise `JsonRpcError` — and the re-prime at `useArtifactDownload.ts:92` requires `JsonRpcError`. A padded `auth_expired` above 64 KiB still defeats the retry.

`demoArtifactDownload.ts:63` pads 500,000 chars, 7.6x this cap, so the "padded error" scenario hits the cap instead of the path it names.

services/vault/src/services/artifacts/streamingArtifactValidator.ts
```diff
@@ -0,0 +1,995 @@
...
+            `Artifact response decryptor_artifacts_hex is ${this.hexCount} characters, below the ${MIN_DECRYPTOR_HEX_CHARS}-character minimum for a real artifact`,
+          );
+        }
+        const pubkey = this.frames[this.frames.length - 2]?.currentKey;
+        if (pubkey) {
+          this.sessionHex.set(pubkey, this.hexPreview);
+        }
+        break;
+      }
+      default:
```
jrwbabylonlab, line 741: This branch never decodes the capture, so invalid UTF-8 in `jsonrpc`, `id` or any unknown field is accepted — `consumeStringByte` only rejects bytes below 0x20. Tracked fields and keys go through the fatal `TextDecoder`; these don't.

A raw `0x80` reaches `finish()`, commits, and earns a receipt for a file that isn't valid JSON. Nothing covers this in the tests.

services/vault/src/hooks/deposit/useArtifactDownload.ts
gbarkhatov: Good catch, and you traced it exactly right. Fixed.

You're correct that the receipt was never the load-bearing check on this path: `onDownloaded` fires off the hook flag alone, and `canActivate = (downloaded || acknowledged)` never re-reads the receipt after mount. So a blocked or dismissed save on Firefox/Safari removed the acknowledgement checkbox and enabled Activate with neither proof nor attestation — while `fallbackSaveHint` promised the user the warning would stay. The mock at `:388` had the same hole, which also broke the invariant `demoArtifactDownload.ts`'s own header claims ("no simulated download can ever unlock a real vault's activation").

The fix separates delivery from proof rather than trying to make the receipt do more work:

- The hook now has a distinct `delivered` flag. Both unverifiable paths (anchor fallback, mock) set `delivered: true`; the validated-and-receipted path remains the only producer of `downloaded: true`.
- `RecoveryArtifactsCard` fires `onDownloaded` only on `downloaded`, so the activation gate is untouched by an unverifiable save. It renders a new state for `delivered`: says plainly that the save can't be confirmed and the warning will remain, keeps the download available as "Download Again", and does not go green.
- `ActivateConfirmationModal` needed no change — the checkbox now simply stays, so Activate requires the acknowledgement. That's strictly stronger than before (which required nothing) and doesn't add a new escape hatch; it just stops suppressing the existing one.
- `ArtifactDownloadModal` does accept `delivered` for its footer "Done", since it gates nothing and the user would otherwise be stuck.

On the test you flagged: `ActivateConfirmationModal.test.tsx:215` stays, since a mocked card calling `onDownloaded` is still the correct Chromium case. Added alongside it: the checkbox survives a delivered-only download, Activate then unlocks only after the acknowledgement, and at the card level `onDownloaded` is not called on the fallback path. The two hook tests that asserted `downloaded === true` for these paths now assert `delivered === true, downloaded === false`.

services/vault/src/services/artifacts/streamingArtifactValidator.ts
gbarkhatov: Confirmed on the mechanism — `CaptureBuffer.push` runs on every byte before `processByte`, so it throws at byte 65,537 and `popFrame` → `throwWireError` is never reached; `VpResponseValidationError` and `JsonRpcError` are unrelated siblings, so `isAuthFailure`'s `instanceof` check is false and the re-prime is skipped.

I'd rather leave it, though. It's a deliberate trade already pinned by a test: `artifactDownloadService.test.ts:518` feeds a 200,000-char error and asserts `VpResponseValidationError`, with the comment "The download fails either way; only the error type differs." An honest `auth_expired` is a few hundred bytes, so 64 KiB is ~300x a plausible one; reaching it needs a hostile or badly broken VP, and a VP that wants to stop the download can simply stop it. The cost is one lost round-trip and a generic message, with `commit` asserted not called so nothing partial lands on disk. Preserving the retry would mean synthesising a `JsonRpcError` when the cap trips inside `error`, which adds a branch to a validated path to serve an input we don't think is real. Happy to revisit if you disagree on that last point.

I did fix the comment you implicitly caught: "Size no longer changes the verdict" at `artifactDownloadService.test.ts:492` was inaccurate above the cap, and now says so.

The demo half is a real bug and is fixed. You were exact on 7.6x: 500,060 bytes against a 65,536-byte cap, so `error-envelope` tripped the cap ~13% in and never produced the `JsonRpcError` its label advertises. It now derives its padding from an exported `MAX_ERROR_VALUE_BYTES` so the two can't drift. I also changed its message — `Invalid state: PendingBabeSetup` matches `isPreDepositorSignaturesError`, so once the padding was corrected the scenario would have exercised the still-processing retry instead of the padded-error path. One scenario can't demonstrate both; if a still-processing scenario is worth having in the panel, happy to add it separately.

services/vault/src/services/artifacts/streamingArtifactValidator.ts
gbarkhatov: Right on all counts, and fixed. `consumeStringByte` only rejects bytes below 0x20, nothing anywhere tracks continuation bytes, and keys / `tx_graph_json` / `verifying_key_hex` / the whole `error` object get the fatal `TextDecoder` while `jsonrpc`, `id` and unknown fields did not. A raw 0x80 reached `finish()`, committed, and earned a receipt for a bundle no strict parser will read — against a file whose own header claims it proves the whole document is well-formed, and RFC 8259 §8.1 makes UTF-8 part of that.

One refinement to your description that makes the fix cheaper than it looks: the bytes *are* captured on that path (`beginStringValue`'s default allocates a `MAX_UNKNOWN_STRING_BYTES` buffer), they were just discarded undecoded. So it's a one-liner reusing the existing helper:

```ts
default:
  decodeJsonString(this.stringCapture, "string value");
  break;
```

`decryptor_artifacts_hex` is unaffected — its capture is `null` and the hex table already rejects anything >= 0x80 — so the ~1.3 GB payload doesn't go through this and the cost stays bounded at <=1 MiB per untracked string.

On coverage, you were right that nothing could catch it: the suite's `validate()` helper runs every fixture through `TextEncoder`, which cannot emit invalid UTF-8. Added a raw-`Uint8Array` helper plus three tests — a lone 0x80 in `jsonrpc`, the same in an unknown forward-compat field, and valid multibyte (`café ✓`) still accepted so the check can't over-reject. Verified all three fail against the old `default: break;`.

jeremy-babylonlabs (approved):

jrwbabylonlab (approved):

PR #2157, opened 2026-07-29
fix(wallet-connector): update wallet mark

jeremy-babylonlabs:
## Problem

Wallet logos were clipped to a circle regardless of their artwork, which cropped self-coloured logos like the MetaMask fox, and the source assets had circle-clipped rasters baked in. The design puts a single-colour mark in an avatar filled with its brand colour and leaves self-coloured logos bare.

## Changes

- Add core-ui `WalletIcon`: brand-filled circle with an inset glyph when the wallet declares an icon background, otherwise a bare, unclipped logo. Used by the wallet list, the connected badge, the wallet menu card and the vault header trigger so every surface agrees.
- Restore the OKX, UniSat, OneKey and Keystone logos to vectors and recolour the single-colour glyphs to white so they read on their disc; black-on-transparent marks were invisible in dark mode.
- Prefer our own asset when AppKit reports an ETH wallet we also ship, so OKX no longer shows a square tile on Ethereum and a circular mark on Bitcoin.

## Testing

- Wallet list, connected badge, wallet menu card and vault header trigger in both light and dark mode.
- OKX on both Bitcoin and Ethereum renders the same circular mark.


packages/babylon-core-ui/src/components/Avatar/WalletIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+    );
+  }
+
+  return (
+    <Avatar
+      url={url}
+      alt={alt}
+      size={size}
+      variant="rounded"
+      className={twMerge("!overflow-visible object-contain", className)}
```
jrwbabylonlab, line 44: `object-contain` is on the wrapper; the img keeps `object-cover`, so bare logos can still crop.

jonybur (approved):
Wallet mark handling checks out. WalletIcon cleanly separates the two cases the design calls for - a brand-filled disc with an inset glyph when the wallet declares an icon background, and a bare, unclipped logo otherwise - and the iconBackground metadata is wired consistently through the wallet list, connected badge, wallet menu card and the vault header trigger, so every surface agrees. The colour values are named shared constants, removing UniSat's black fill matches its self-coloured mark, and moving ONEKEY_BRAND_GREEN into constants.ts leaves no dangling references. Dropping variant="circular" from the AvatarGroup in Connect.tsx is safe now that WalletIcon owns its own shape (the cloned variant prop is simply ignored). The first-party OKX override in resolveEthDisplayWallet only fires on a name match and falls through to AppKit's own artwork otherwise, and no ETH-side wallet metadata carries an iconBackground that the override could drop. CI is green.

PR #2159, opened 2026-07-29
feat(vault): add activate-and-redeem escape hatch for stuck verified deposits

jonybur:
## Why

A reverted activation (e.g. Aave paused) leaks the HTLC secret via calldata. The VP can then sweep the peg-in on Bitcoin while the vault is still Verified on Ethereum - the depositor gets no collateral and can never refund (the HTLC outpoint is spent). The UI currently shows "Ready to activate" with a button that cannot succeed.

## What

Support the contract escape hatch `activateVaultWithSecretAndRedeem`: reveal the secret and immediately redeem the vault to the depositor, skipping application activation.

- Stuck state detected (Verified on ETH + HTLC spent on BTC) -> the deposit card stays in Pending, shows an "Activation incomplete" warning with an always-visible reassuring line ("Your BTC is not lost - withdraw to receive it back."), and its CTA becomes Withdraw.
- Withdraw opens a dedicated self-contained modal (confirm with risk acknowledgment -> in-place progress -> success), same pattern as the refund modal - no multistepper, no nested dialogs, deliberately no red/danger styling.
- Also always reachable on any Verified vault via an "advanced" link in the activation dialog, warning that waiting for expiry + refund is the safe default.
- Refund still appears only after expiry, so the two never compete.

## How

- SDK: `activateVaultWithSecretAndRedeem` ABI entry (signature verified against `vault-contracts-aave-v4`), `activateVaultAndRedeem()` sharing validation with `activateVault()`, new `PeginAction.ACTIVATE_AND_REDEEM` driven by an `htlcSpent` protocol-state flag.
- Detection: the existing HTLC outspend probe widened to owned Verified vaults (display only).
- Tx path: `redeemImmediately` mode in `useVaultActions`, reusing every pre-reveal gate - fresh on-chain VERIFIED read, double hashlock check, zero-wiped secret buffers (derivation extracted to a shared `htlcSecretDerivation` service). The pause gate checks the protocol scope only; an Aave pause is what the hatch escapes.
- UI: `EmergencyWithdrawModal` + `useEmergencyWithdrawModal` bundle mounted next to the refund modal; rows/cards route `ACTIVATE_AND_REDEEM` straight to it; all strings in `copy.ts`.

Tests: new SDK / state-machine / polling / pause-gate coverage; full suites, `tsc`, and lint pass.

## UI

| Stuck state detected | Advanced entry |
|---|---|
| Withdraw modal, stuck state | Withdraw modal, advanced path |

| Link in the activation dialog | Success |
|---|---|
| Activation dialog with advanced withdraw link | Withdrawal submitted |


services/vault/src/components/shared/protocolStatus.ts
```diff
@@ -99,6 +99,17 @@ export function isActivationBlocked(gate: ProtocolGateState): boolean {
...
 }
 
+/**
+ * The activate-and-redeem escape hatch is an EXIT blocked ONLY by a
+ * protocol-scope pause: `activateVaultWithSecretAndRedeem` never touches the
+ * application adapter, and an aave-scope pause is precisely the situation the
+ * hatch exists to escape — blocking on it would disable the recovery when it
+ * is most needed.
+ */
+export function isActivateAndRedeemBlocked(gate: ProtocolGateState): boolean {
```
jrwbabylonlab, line 121: Also needs `getApplicationStatus == Active` — contract reverts `ApplicationNotActive` after the secret is out.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/peginState.ts
```diff
@@ -151,6 +168,16 @@ export function getPeginProtocolState(
   }
 
   if (contractStatus === ContractStatus.VERIFIED) {
+    // A spent HTLC while still Verified means the peg-in was swept without
+    // activation: activating normally would hand the collateral to an
+    // application flow that already failed once, and refunding is impossible
+    // (the outpoint is gone). Surface only the escape hatch.
+    if (htlcSpent) {
```
jrwbabylonlab: A spent HTLC may be the refund, not the PegIn. Compare `spendingTxid` against the PegIn txid.

services/vault/src/components/shared/protocolStatus.ts
jonybur: Implemented, the full version.

My first instinct was to defer this: the getter is absent from both bundled ABI slices, and guessing a selector on a funds path is not something I would do. But that was a false dichotomy - this repo already has the pattern for exactly this situation in `pause-state/query.ts`, which inlines a self-contained single-function fragment for a getter deliberately kept out of the signing-subset ABI. So the gate is built the same way: `getApplicationStatus(address) returns (uint8)` with the enum `None=0, Active=1, Paused=2`, both confirmed against upstream `IApplicationRegistry.sol`, and `RedeemLogic.sol:93-98` as the authority that the redeem path requires `== Active` on `vBasic.applicationEntryPoint`.

How it reads: one multicall for the vault's `applicationEntryPoint` (`getBtcVaultBasicInfo`) plus the registry's `applicationRegistry()` address, then `getApplicationStatus` against that registry. The confirm button is withheld with an explanatory notice **only on a confirmed non-Active** status. A failed or pending read does **not** block - this is an exit whose user has already had their peg-in swept, so over-blocking strands exactly the person the hatch exists for, and the pre-broadcast simulation still refuses to sign into a genuinely inactive application. Four tests cover active / inactive / unknown / protocol-paused; the inactive case was verified to fail without the gate.

Being precise about what each layer buys, since they are not interchangeable:

- The CTA gate stops offering an action that cannot succeed - truthful UI at render time.
- `executeWrite`'s mandatory pre-broadcast simulation refuses to sign into a known-failing state, so a precondition already failing at submission never publishes the secret.
- Neither closes the simulate-to-mine window. A registration deactivated after a passing check still mines a reverting tx with the secret in public calldata. Both checks are point-in-time; that residual is inherent to the contract enforcing the condition at all, and the docs now say so rather than claiming a guarantee they cannot make.

One related correction worth flagging, because it cuts the wrong way on a funds surface: the `ApplicationNotActive` entry in `errorMessages.ts` originally claimed "this application is currently paused" (wrong condition), and my first fix replaced it with wording that included "your secret has not been revealed". That map is phase-agnostic - the same entry is served for a simulation failure *and* for a mined revert - so that sentence would have been a false assurance in precisely the race above. The shared entry now states only what is phase-independent, and the "your secret has not been revealed" assurance lives on the confirm-screen notice, where no transaction has been attempted and it is unconditionally true.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/peginState.ts
jonybur: Fixed, and renamed so the contract is visible in the type: `htlcSpent` is now `htlcSpentByPeginTx` across the SDK input, the vault model, and the deriver, whose docs now require the caller to prove the spender rather than infer it.

The deriver compares canonicalized txids - `spendingTxid === activity.peginTxHash` - and fails safe in both ambiguous directions: a spend by any other txid (your refund case) keeps the normal `ACTIVATE_VAULT` flow, and a missing `spendingTxid` also keeps it, because the hatch reveals the secret and must not be offered on ambiguous evidence.

Three compute-level tests pin it: PegIn-spend flags the stuck state, refund-spend does not, unknown spender does not. Both "does not" cases were verified to fail against the old bare `spent === true` derivation by reverting the comparison - they are regression tests, not decorations.

One deliberate fixture detail: the spend arrives from esplora as bare lowercase hex while the activity's PegIn txid is 0x-prefixed indexer hex, so the passing test deliberately mismatches those forms. Same-form fixtures would stay green even if canonicalization broke - which would silently disable the hatch for genuinely swept vaults, the exact failure this comment is protecting against.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/peginState.ts
```diff
@@ -151,6 +176,16 @@ export function getPeginProtocolState(
   }
 
   if (contractStatus === ContractStatus.VERIFIED) {
+    // A spent HTLC while still Verified means the peg-in was swept without
+    // activation: activating normally would hand the collateral to an
+    // application flow that already failed once, and refunding is impossible
+    // (the outpoint is gone). Surface only the escape hatch.
+    if (htlcSpentByPeginTx) {
```
jrwbabylonlab, line 183: The stuck state offers only ACTIVATE_AND_REDEEM and strips ACTIVATE_VAULT, but it is derived purely from BTC-side evidence and carries no signal about whether the breaking condition still holds. `activateVault` is `whenNotPaused` on the adapter — the pause this feature exists to escape. If the adapter unpauses later, normal activation would succeed and give the depositor the collateral position, but the UI permanently hides it and offers only an irreversible redeem. Suggest offering both, hatch as fallback.

services/vault/src/context/deposit/PeginPollingContext.tsx
```diff
@@ -253,17 +253,32 @@ export function PeginPollingProvider({
...
+            status !== ContractStatus.EXPIRED &&
+            status !== ContractStatus.VERIFIED
+          )
+            return false;
+          // Once this device has submitted the reveal (CONFIRMED), a spend is
+          // the expected VP sweep, not the stuck state — and the display
+          // ignores the probe anyway, so skip the request.
+          if (
+            status === ContractStatus.VERIFIED &&
+            localStatusById.get(a.id) === LocalStorageStatus.CONFIRMED
```
jrwbabylonlab, line 347: A per-device localStorage marker is the only guard against the activation-succeeded-but-indexer-lagging race. On a second device or a cleared profile a healthy deposit renders "Activation incomplete", and `useSigningRequiredNotifications` bypasses both gates for this action so a push fires saying the vault could not be activated — clicking it opens a BTC signing popup, since the secret is derived before any on-chain status read. Fresh VERIFIED check stops real harm, but it is a false alarm plus a spurious signature. `activationDeadlinePassed` already uses an authoritative chain read; this should too.

services/vault/src/components/deposit/EmergencyWithdrawModal/EmergencyWithdrawConfirmContent.tsx
```diff
@@ -0,0 +1,153 @@
...
+  // the redeem in that state, so offering it would spend the user's
+  // acknowledgement on a transaction that cannot succeed. `undefined` (loading
+  // or a failed read) does NOT block — over-blocking strands a depositor whose
+  // peg-in is already swept, and the pre-broadcast simulation still refuses to
+  // sign into a genuinely inactive application.
+  const applicationActive = useVaultApplicationActive(vaultId);
+  const applicationInactive = applicationActive === false;
+  const canWithdraw =
+    acknowledged &&
+    !withdrawing &&
```
jrwbabylonlab, line 64: `canWithdraw` includes the pause gate but only `applicationInactive` and `error` render a callout, so under a protocol pause the confirm button is silently disabled with no explanation — on a modal whose user believes their BTC is stuck. The copy for it exists but is only set by the handler behind the disabled button.

services/vault/src/context/deposit/PeginPollingContext.tsx
jonybur: Fixed in `a93d99dfb` — the stuck state now requires a positive on-chain confirmation, in the same tiered shape as `useActivationDeadlineGate` you pointed at.

New `useStuckVaultChainConfirm`:
- **Tier 1** — suspects are indexer-VERIFIED vaults with the HTLC proven swept by the pegin tx. Free: reuses the existing BTC probe, no extra request.
- **Tier 2** — each suspect is checked against `getVaultBasicInfo`; only a vault the *chain* still reports `VERIFIED` is promoted to stuck. `ACTIVE` means activation landed and the indexer is behind.

`computeDepositPollingResult` now requires `stuckStateConfirmedOnChain` for `htlcSpentByPeginTx`, so the false "Activation incomplete" and the spurious signing prompt cannot fire on a second device or a cleared profile.

It **fails open** in every uncertain case (RPC failure, unregistered vault, already-ACTIVE). Deliberate: failing the other way reproduces exactly the defect, while failing open costs a genuinely stuck depositor one poll interval before the hatch appears — recoverable, where the false prompt is not.

The `localStorage` check stays, demoted to a request-saving shortcut with a comment saying so. You were right that it was never a correctness guard; it just avoids a probe when this device already revealed.

The test that matters is the negative one: identical BTC evidence to the positive case — because the VP's sweep is what activation looks like on the BTC side — with the only difference being chain confirmation.

services/vault/src/components/deposit/EmergencyWithdrawModal/EmergencyWithdrawConfirmContent.tsx
jonybur: Fixed in `a93d99dfb`. You had it exactly: the copy existed but `COPY.pegin.activateAndRedeemPaused` was only set by the click handler *behind the button the pause had already disabled*, so it was unreachable by construction.

It now renders as a `Callout`, ordered ahead of the application-inactive one since a protocol pause is protocol-scope and outranks an application-scope check. Three tests: it shows under a protocol pause, it does not when nothing is paused, and an aave-scope pause neither withholds the exit nor renders it (`isActivateAndRedeemBlocked` is protocol-only).

Agreed this was the worst place to leave an unexplained dead control.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/peginState.ts
jonybur: Not changed, and I want to put the reason in front of you rather than pick a shape — your argument is right but the UI cannot currently express it.

I implemented the suggestion (`[ACTIVATE_VAULT, ACTIVATE_AND_REDEEM]`, hatch as fallback) and it made things **worse**, so I backed it out.

The card renders exactly **one** action. `actionStatus.ts` returns a single `getPrimaryActionButton` result and `VaultsLifecycleSections` dispatches on that one action; there is no secondary-CTA affordance anywhere. `getPrimaryActionButton` checks `ACTIVATE_VAULT` **before** `ACTIVATE_AND_REDEEM`, so adding both with activation first does not offer both — it makes the CTA "Activate" and removes the escape hatch from the UI entirely. A genuinely stuck depositor would click Activate, watch it fail against the very pause the hatch exists to escape, and have no route to the hatch at all.

Two existing tests caught it — `returns Withdraw for the stuck state` started returning Activate.

So the options are:

1. **Add a secondary action to the card.** Delivers the intent properly. Touches `actionStatus.ts`, the lifecycle rendering and the badge priority — real UI work, and a product call about how a two-action deposit row should look.
2. **Carry `ACTIVATE_VAULT` in `availableActions` but keep the hatch as the CTA.** No regression, but the extra action is inert until something renders it — which felt like claiming the fix without making it.
3. **Leave as-is** until 1 is scoped.

Your underlying point stands and I am not arguing it: the stuck state is derived purely from BTC-side evidence and carries no signal about whether the pause still holds, so a depositor whose adapter later unpauses is steered into an irreversible redeem when normal activation would have given them the collateral position. That is a real gap. It just needs a second CTA to fix, which is bigger than this thread.

Leaving this one open for you to weigh in. The other two are addressed in `a93d99dfb`.

jonybur (commented):
Reviewed the full diff at head 3cbf9b5 (main merge). The two-tier stuck-state detection (BTC outspend proven to be the PegIn tx + positive on-chain VERIFIED reconfirmation), the fail-open direction on every uncertain read, and the shared `deriveHtlcSecretHex` extraction all hold up against the CLAUDE.md critical-path rules - the hashlock is re-fetched from the registry and re-checked in the SDK before calldata is assembled on both reveal paths, and `executeWrite`'s pre-broadcast simulation (transactionFactory.ts) backs the fail-open claims.

One finding, plus a doc nit:

**Case-normalization miss in the Tier-1 suspect scan** - `services/vault/src/context/deposit/PeginPollingContext.tsx`, in the new `stuckSuspectIds` memo:

```ts
const spend = htlcRefundByDepositId.get(a.id);
```

`refundByDepositId` is keyed by **lowercased** vault id (`useBtcHtlcRefundStatus.ts:49` - `o.depositId.toLowerCase()`), but this lookup passes `a.id` unnormalized. Every sibling lookup in this file lowercases - including two added by this same diff (`stuckConfirmedIds.has(activity.id.toLowerCase())`, the `expiredIds` set in the refunded-cache effect) and the pre-existing `refundedHtlcVaultIds.has(a.id.toLowerCase())` three lines above. If an activity id ever arrives with uppercase hex, the suspect never forms, Tier 2 never runs, and the "Activation incomplete" card + Withdraw CTA silently never appear for a genuinely stuck deposit - the exact defect this PR exists to fix, failing invisibly. (The advanced link keeps the hatch reachable manually, and the failure direction is safe - no wrongful reveal - so this is a detection-robustness bug, not a fund-safety one.) Fix is one call: `htlcRefundByDepositId.get(a.id.toLowerCase())`. The neighboring `localStatusById.get(a.id)` skip in the outpoint filter is worth the same one-line treatment if `localStatusById` is also keyed lowercased.

**Doc nit** (no action needed for merge): in `EmergencyWithdrawConfirmContent`, the `vaultId` prop's JSDoc says "`undefined` while unknown (loading or a failed read)" but the prop is typed non-optional `Hex` - the sentence describes `useVaultApplicationActive`'s return, not the prop.

services/vault/src/context/deposit/PeginPollingContext.tsx
```diff
@@ -342,13 +362,41 @@ export function PeginPollingProvider({
...
+  // is equally consistent with a healthy deposit whose activation the indexer
+  // has not caught up to yet.
+  const stuckSuspectIds = useMemo(() => {
+    const ids: Hex[] = [];
+    for (const a of activities) {
+      if (
+        ((a.contractStatus ?? 0) as ContractStatus) !== ContractStatus.VERIFIED
+      )
+        continue;
+      const spend = htlcRefundByDepositId.get(a.id);
```
gbarkhatov, line 384: `htlcRefundByDepositId` is keyed **lowercased** (`useBtcHtlcRefundStatus.ts:58` — `map.set(o.depositId.toLowerCase(), ...)`), but this lookup passes `a.id` unnormalized.

Every sibling normalizes: `refundedHtlcVaultIds.has(a.id.toLowerCase())` 34 lines up, the `expiredIds` set in the refunded-cache effect, both the `stuckConfirmedIds` and `activationDeadlinePassedIds` reads in the provider value, and `computeDepositPollingResult:160` before hitting this same map.

If an activity id ever carries uppercase hex, the Tier-1 suspect never forms, Tier 2 never runs, and the "Activation incomplete" card plus Withdraw CTA silently never appear for a genuinely stuck deposit — the exact defect this PR exists to fix, failing invisibly. The direction is safe (no wrongful reveal) and the advanced link keeps the hatch reachable manually, so this is detection robustness rather than fund safety.

Fix is one call: `htlcRefundByDepositId.get(a.id.toLowerCase())`.

The neighbouring `localStatusById.get(a.id)` in the outpoint filter does *not* need the same treatment — that map is built from raw `p.id` and `optimisticStatuses` keys, so both sides already use the same convention.

packages/babylon-ts-sdk/src/tbv/core/contracts/abis/BTCVaultRegistry.abi.ts
```diff
@@ -227,6 +227,24 @@ export const BTCVaultRegistryABI = [
     outputs: [],
     stateMutability: "nonpayable",
   },
+  {
+    type: "function",
+    name: "activateVaultWithSecretAndRedeem",
```
gbarkhatov, line 232: Flagging for the second code-owner rather than for you: this selector and the `ApplicationStatus { None, Active, Paused }` mapping in `services/vault/src/clients/eth-contract/application-status/query.ts` are the only two facts in this PR that cannot be checked from inside this repo. Both are asserted against upstream Solidity in comments, and both fail silently if wrong — the two `bytes32` arguments make a swapped order type-invisible, and a shifted enum would have the CTA gate read Paused as Active and offer an exit that reverts.

Per the critical-path rule the sign-off should include eyes on `vault-contracts-aave-v4` at the deployed rev, not just on the comment asserting it.

services/vault/src/components/deposit/actionStatus.ts
```diff
@@ -97,6 +97,9 @@ export function getActionStatus(
 }
 
 const ACTION_REQUIRED_BADGE_PRIORITY: PeginAction[] = [
+  // The stuck-state recovery outranks everything: the deposit cannot progress
+  // any other way once the peg-in was swept without activation.
+  PeginAction.ACTIVATE_AND_REDEEM,
```
gbarkhatov, line 102: Worth pinning while the `peginState.ts` thread is open: this list ranks `ACTIVATE_AND_REDEEM` **above** `ACTIVATE_VAULT`, which is the inverse of the order `getPrimaryActionButton` checks them in (`peginStateMachine.ts` tests `ACTIVATE_VAULT` first).

Harmless today because the two never coexist. But if that thread is resolved by offering both actions, these two orderings disagree: the badge would read "Withdraw required" while the CTA reads "Activate". If a secondary CTA lands, the intended precedence should be decided once and expressed in one place rather than encoded twice in opposite directions.

services/vault/src/hooks/useStuckVaultChainConfirm.ts
```diff
@@ -0,0 +1,114 @@
...
+ * Fails OPEN in every uncertain case: an absent id means "not confirmed stuck",
+ * so an RPC failure, an unregistered vault, or a vault the chain already reports
+ * ACTIVE all leave the deposit rendering normally. That direction is deliberate.
+ * Failing the other way produces exactly the defect this exists to remove — a
+ * false alarm plus a signing prompt on a healthy deposit. The cost of failing
+ * open is that a genuinely stuck depositor waits one poll interval for the
+ * escape hatch to appear, which is recoverable; the cost of failing closed is
+ * not.
+ */
+export function useStuckVaultChainConfirm(
```
gbarkhatov, line 53: No direct test for this hook, and its fail-open branches are precisely what the review thread it answers was about: RPC throw, `depositor === zeroAddress`, chain already reporting `ACTIVE`, and the intersect-with-live-suspects step that stops a confirmation outliving the suspicion that produced it.

Today those are only exercised indirectly through `computeDepositPollingResult` fixtures that pass `stuckStateConfirmedOnChain` as a plain boolean — which tests the consumer, not the gate. A small hook test asserting each uncertain read yields an empty set would pin the direction itself, so a later refactor cannot quietly flip it to fail closed.

services/vault/src/components/deposit/EmergencyWithdrawModal/EmergencyWithdrawConfirmContent.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { useVaultApplicationActive } from "@/hooks/useVaultApplicationActive";
+
+interface EmergencyWithdrawConfirmContentProps {
+  /** True when the stuck state was detected on-chain — drives the body copy. */
+  stuckStateDetected: boolean;
+  /**
+   * Vault whose application registration gates this exit. `undefined` while
+   * unknown (loading or a failed read) — see {@link useVaultApplicationActive}.
+   */
+  vaultId: Hex;
```
gbarkhatov, line 24: Doc nit: the prop is a non-optional `Hex`, so it is never `undefined`. That sentence describes `useVaultApplicationActive`'s return value — which is already documented on the `applicationActive` local below. Worth reducing to just naming the vault whose application registration gates this exit.

gbarkhatov (approved):
Reviewed the full diff at head `3cbf9b58`. Approving.

The reveal path holds up against the critical-path rules: the hashlock is re-fetched from the registry and re-checked in the SDK before calldata is assembled on both entry points, `deriveHtlcSecretHex` reads its signing-critical inputs (depositor pubkey, `htlcVout`) from chain and verifies the indexer-supplied Pre-PegIn against the on-chain hash before the wallet popup, buffers are wiped in `finally`, and every uncertain read fails open in the direction that can only cost a poll interval instead of raising a false alarm.

Confirmed the four addressed threads land as described, in the code rather than in the replies:

- application-status gate — `isVaultApplicationActive` + `useVaultApplicationActive`, callout and withheld confirm only on a *confirmed* non-Active
- spender proof — `htlcSpentByPeginTx` with canonicalized txid equality in both `PeginPollingContext` and `computeDepositPollingResult`, ambiguous evidence keeping the normal flow
- indexer-lag false positive — `stuckStateConfirmedOnChain` is now required, `localStorage` demoted to a request-saving shortcut with a comment saying so
- unexplained disabled confirm under a protocol pause — `Callout` rendered, ordered ahead of the application-scope one

Ran the affected suites at head: 308 vault tests and 189 SDK tests green. Note for anyone else checking locally — three of these fail spuriously against a stale `packages/babylon-ts-sdk/dist`; rebuild the SDK first.

Still open: the `peginState.ts` thread about stripping `ACTIVATE_VAULT` from the stuck state. Your reasoning for backing the change out is sound and the regression you hit is real, so this needs @jrwbabylonlab's call on whether a stuck row gets a secondary CTA rather than another round from you. One trap if it is ever implemented is noted inline on `actionStatus.ts`.

Five non-blocking notes inline. Only the first is a behavior bug, and it is a one-liner.

PR #2160, opened 2026-07-29
refactor(vault): render the liquidation charts with visx

jonybur:
## What

Rewrites the core-ui liquidation charts (`SeizureMap` + `Timeline`) on visx SVG and ships the overview's liquidation simulator surface. Behaviour changes (product decisions from the chart design doc + tooltip spec, made during the PR):

- **Event hover tooltip** on bands (outlined chip, At price / Distance / Vaults / Seizes, Cumulative row) — hover/focus only, click interactions removed
- **Liquidated events** swap their vault list + amount for a "Liquidated" label on band and legend, and dim
- **Timeline Y scale is non-uniform**: regions weighted by √(price span) with a 44px label floor, safe zone guaranteed ≥ the largest event; gutter blocks tile trigger-to-trigger
- **Price simulator** on the overview section: $0→live slider (live-updating), Seized/Collateral header figures, Simulation chip, and Liquidation Events cards with a "Liquidated in Simulation" state

## Why

The old charts were hand-positioned HTML/CSS needing DOM measurement passes; visx gives real scales with pixel-known geometry. The simulator surface and interaction model implement the answers in the chart design Q&A doc.

## How

One measured-width SVG per chart; polylinear d3 scales (segmented, anchored span-weighted, linear); analytic label declutter and text dropout; canvas-measured truncation with clipPath backstops; the vault projection derives all figures (event aftermath priced at each event's own trigger) and copy stays in `copy.ts`. visx is pinned at 4.0.0, bundled, and lives in devDependencies (the public type graph never references it).

## Breaking changes (core-ui package surface)

- `SeizureGutter` / `SeizureGutterProps` are no longer exported (Timeline-internal)
- `onBandClick` / `bandClickHint` removed from both chart prop types (hover-only interaction)

The squash-merge commit must carry a `BREAKING CHANGE:` footer listing the above.

## Try it

Live Storybook of this branch (share freely): https://babylon-charts-storybook.vercel.app
- [SeizureMap Simulator](https://babylon-charts-storybook.vercel.app/?path=/story/components-data-display-charts-seizuremap--simulator) - drag the BTC price and watch events liquidate
- [Timeline Simulator](https://babylon-charts-storybook.vercel.app/?path=/story/components-data-display-charts-timeline--simulator) - candles, span-weighted regions, safe zone


jonybur:
@greptile-apps

packages/babylon-core-ui/src/components/LiquidationChart/priceScale.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+  const domain = ticks.map((t) => t.value);
+  assertStrictlyDescending(domain);
+  const range = domain.map((_, i) => (i / last) * plotHeight);
+  return scaleLinear<number>({ domain, range, clamp: true });
+}
+
+/** Linear price → pixel offset from the plot top over [max, min] (Timeline). */
+export function createLinearPriceScale(priceMax: number, priceMin: number, plotHeight: number): PriceScale {
+  if (priceMax === priceMin) {
```
jrwbabylonlab: Only exact equality guarded — reversed/NaN bounds invert the axis or give NaN coords.

jonybur (commented):
Two findings, both in the new core-ui chart code:

1. `textMeasure.ts` lands as a binary file - the diff shows `Binary files /dev/null and b/.../textMeasure.ts differ`, meaning the source contains non-text bytes (the measurement cache key appears to use literal NUL characters as delimiters, and those raw 0x00 bytes made it into the file itself). As committed, the file cannot be reviewed on GitHub and diff/blame/grep tooling degrades from here on. Writing the delimiters as the escape sequence backslash-u0000 keeps the runtime key identical while restoring a plain text file.

2. `__tests__/Timeline.test.tsx` asserts `getByText("Jun 15")` for a candle at epoch `1_750_000_000_000`, which is 2025-06-15T15:06:40Z. `defaultFormatTime` renders through `toLocaleDateString("en-US", ...)` in the machine's local timezone and the vitest config does not pin `TZ`, so on any machine at UTC+9 or later this renders "Jun 16" and the suite fails. Either pass an explicit `formatTime` in the test, use a mid-day UTC timestamp, or pin `TZ` for the suite.

Everything else held up under scrutiny: the removed contract surface (popoverMetrics, cumulativeLabel, onBandClick, bandClickHint and the related copy.ts keys) is gone with no dangling references, the visx packages and new test deps are pinned exact, the sibling package.json edits are the same pins applied consistently, the scale/declutter math guards its degenerate domains, and the transitions use the motion tokens.

packages/babylon-core-ui/src/components/LiquidationChart/SeizureMap.tsx
```diff
@@ -38,17 +70,39 @@ export function SeizureMap({
...
-  const priceToFraction = useCallback((price: number) => segmentedFraction(priceAxis, price), [priceAxis]);
+  const hasTopLegend = !compact && showShareLegend && bands.length > 0;
+  const hasXAxis = !compact && Boolean(shareAxisTicks?.length || shareAxisLabels?.length);
+  const { parentRef, layout } = useChartLayout({ axisSide: "left", hasTopLegend, hasXAxis });
+
+  const priceScale = useMemo(
+    () => createSegmentedPriceScale(priceAxis, layout.plotHeight),
+    [priceAxis, layout.plotHeight],
+  );
+  const shareScale = useMemo(
```
jrwbabylonlab, line 99: `shareScale` lacks `clamp: true` (both price scales have it) — reversed bands give negative width.

packages/babylon-core-ui/src/components/LiquidationChart/SeizureMap.tsx
```diff
@@ -38,17 +76,40 @@ export function SeizureMap({
...
+  );
+
+  const bandRect = (b: LiquidationBand): BandRect => {
+    const x = shareScale(b.shareStart);
+    const y = priceScale(b.priceTop);
+    return {
+      x,
+      y,
+      width: shareScale(b.shareEnd) - x,
+      height: priceScale(b.priceBottom) - y,
```
jrwbabylonlab: `shareScale` (`:92`) still has no `clamp: true` while both price scales do. `shareStart 0.6 / shareEnd 0.4` gives `width="-189.6"` and the rect is not rendered; `shareEnd 1.4` paints at 1.4x plotWidth, and `LiquidationChart.css:48` sets `overflow: visible`, so it lands on adjacent page content instead of being clipped. `ShareLegend` at `:33` guards with `Math.max(0, ...)`; `bandRect` does not.

packages/babylon-core-ui/src/components/LiquidationChart/priceScale.ts
jrwbabylonlab: Still only exact equality. `assertStrictlyDescending` is wired into `createSegmentedPriceScale` only, so the Timeline path (`Timeline.tsx:104`) is unguarded: reversed bounds map cheap prices to the top and invert the chart silently, and a NaN bound passes `!==` then makes d3 `normalize` return `constant(NaN)`, so every coordinate is NaN and the chart renders blank. Same `PriceAxisTick[]` type, opposite contracts in the two components.

packages/babylon-core-ui/src/components/LiquidationChart/priceScale.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+  return scaleLinear<number>({ domain: [priceMax, priceMin], range: [0, plotHeight], clamp: true });
+}
+
+/** A duplicate or misordered tick would silently shift every band by up to
+ * half a segment, so a broken axis is an error, not a best-effort render. */
+function assertStrictlyDescending(domain: number[]): void {
+  for (let i = 1; i < domain.length; i++) {
+    if (!(domain[i] < domain[i - 1])) {
+      throw new Error(
```
jrwbabylonlab, line 149: Throwing during render escalates a degraded chart into the app-level error boundary. `liquidationChartData.ts:209` builds the axis from a `Set` that keeps a NaN `btcPrice`, and the sort leaves it in place, so one price-feed miss takes the whole vault app to the global error page. Old `scale.ts:22` handled the zero-span case and still rendered; an axis is exactly the optional display concern CLAUDE.md allows a fallback for.

packages/babylon-core-ui/src/components/LiquidationChart/types.ts
```diff
@@ -100,14 +86,16 @@ interface LiquidationChartBase {
   /** Background grid. Omit for the default dashed full grid. */
   grid?: ChartGridConfig;
   /**
-   * Hide all in-band text (tiny bands already drop text automatically); the
-   * hover popover still names the event. For dense/preview surfaces.
+   * Hide all in-band text (tiny bands already drop text automatically).
+   * For dense/preview surfaces.
    */
   hideBandLabels?: boolean;
```
jrwbabylonlab, line 112: Dropping `popoverMetrics` / `cumulativeLabel` / `bandClickHint` removes a feature that was live: at merge-base `liquidationChartData.ts:203` supplied At price / Distance / Vaults / Seizes and `copy.ts:1291` held the hint, so the production SeizureMap loses per-event hover detail with nothing replacing it. The old `role="button"` / `tabIndex` / Enter-Space handling goes with it, which was the only keyboard affordance. Too much for a PR titled `refactor`.

packages/babylon-core-ui/src/components/LiquidationChart/SeizureMap.tsx
```diff
@@ -1,24 +1,62 @@
...
+              y={0}
+              width={width}
+              height={layout.legendHeight}
+              rx={BAND_RADIUS_PX}
+            />
+            <Text
+              className="bbn-liq-legend__text"
+              x={x + width / 2}
+              y={layout.legendHeight / 2}
+              textAnchor="middle"
```
jrwbabylonlab: Legend text has no `clipPath`, unlike `BandLayer.tsx:113` and `SeizureGutter.tsx:88`. Combined with `textMeasure` returning 0 when canvas is unavailable (which short-circuits `truncateToWidth` into returning the full string), an SSR or canvas-less render paints labels across neighbouring segments with nothing clipping them.

packages/babylon-core-ui/src/components/LiquidationChart/LiquidationChart.css
```diff
@@ -39,326 +35,170 @@
...
-.bbn-liq-chart__yaxis-label {
-  position: absolute;
-  right: 0.5rem;
-  transform: translateY(-50%);
-  white-space: nowrap;
-  font-size: var(--liq-font-axis);
+
+.bbn-liq-axis-text {
+  fill: var(--liq-axis-text);
   letter-spacing: 0.4px;
```
jrwbabylonlab: `letter-spacing: 0.4px` here and at `:105` / `:117` duplicates `AXIS_LETTER_SPACING_PX` in `textMeasure.ts:17`, which is what sizes the axis pills (`ChartFrame.tsx:171`). Editing one silently mis-sizes every pill. Emit the TS constant as a CSS var.

packages/babylon-core-ui/vitest.config.ts
```diff
@@ -0,0 +1,20 @@
...
+import react from "@vitejs/plugin-react";
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  plugins: [react()],
+  test: {
+    globals: true,
+    environment: "jsdom",
+    setupFiles: ["./src/test/setup.ts"],
+    include: ["src/**/*.test.{ts,tsx}"],
```
jrwbabylonlab, line 12: `dts()` in `vite.config.ts:9` excludes `*.stories` but not tests, and `tsconfig.lib.json` includes all of `src`, so `dist/**/__tests__/*.d.ts` and `dist/test/setup.d.ts` ship in the published `@babylonlabs-io/core-ui` tarball.

packages/babylon-core-ui/src/components/LiquidationChart/index.ts
```diff
@@ -1,4 +1,3 @@
-export * from "./SeizureGutter";
 export * from "./SeizureMap";
 export * from "./Timeline";
```
jrwbabylonlab, line 2: Dropping the `SeizureGutter` re-export removes `SeizureGutter`, `SeizureGutterProps`, `PriceFrame`, `PriceFrameProps`, `AxisTick`, `LevelMarker`, `BandPopoverMetric` and four `LiquidationBand` fields from the published package surface. Nothing in the monorepo still references them, but this is a published semantic-release package and needs a `BREAKING CHANGE:` footer.

packages/babylon-core-ui/package.json
```diff
@@ -96,6 +102,13 @@
   "dependencies": {
     "@hookform/resolvers": "3.9.1",
     "@popperjs/core": "2.11.8",
+    "@visx/event": "4.0.0",
```
jrwbabylonlab: These seven are runtime `dependencies` but are absent from `rollupOptions.external` in `vite.config.ts:22`, so they are bundled into `dist` and installed by every consumer. `@visx/scale` pulls `@visx/vendor`, which brings `d3-geo`, `d3-delaunay`, `delaunator` and `robust-predicates` that this chart never uses. Either externalize them or drop them from `dependencies`.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -114,6 +114,7 @@ function LiquidationChartPanel({
         priceAxis={priceAxis}
         shareAxisTicks={shareAxisTicks}
         showShareLegend={false}
+        liquidatedLabel={COPY.liquidations.liquidatedBandLabel}
```
gbarkhatov, line 175: This is a product change, not a refactor — it should be in the PR description. A liquidated event now reads `Liq Event N` / `Liquidated` instead of `(contain vault N)` / `0.6 BTC`. Stacked on the popover removal, the production Seizure Map loses *every* per-event detail: the vault list, the seized amount, and the hover metrics. Two behaviour changes ride in a PR titled `refactor`, and neither is mentioned.

packages/babylon-core-ui/src/components/LiquidationChart/SeizureGutter.tsx
```diff
@@ -1,77 +1,140 @@
...
-      // Anchor at the true price; `cursor` only prevents overlap when a
-      // previous band's minimum height ran past this band's trigger.
-      const top = Math.max(priceToFraction(b.priceTop), cursor);
-      const bottom = Math.max(priceToFraction(b.priceBottom), top + MIN_BAND_FRAC);
-      geoms.set(b.key, { top, height: bottom - top });
-      cursor = bottom;
-    }
+    if (!bands.length) return geoms;
+    const firstTop = priceScale(bands[0].priceTop);
+    const blockHeight = Math.max(0, (plotHeight - firstTop) / bands.length);
```
gbarkhatov: Three things follow from the equal-block sizing, all worth resolving before merge:

1. **`bands` ordering is assumed, never asserted.** `bands[0].priceTop` is the stack top and the array index *is* vertical position. `types.ts` documents no ordering for `bands`. Odd asymmetry: this PR adds a hard throw for a misordered `priceAxis` while the `bands` invariant — now more load-bearing than the axis — fails silently into an upside-down gutter.

2. **`blockHeight` can collapse to 0 with no signal.** `priceScale` clamps, so a `bands[0].priceTop` below the axis floor gives `firstTop === plotHeight` → every block is zero-height *and* the safe-zone rect (`height={firstBandTop}`) fills the whole gutter: a fully-cascaded position renders as "all safe". The JSDoc promise that an event below the floor "keeps its block but goes unmarked" only holds while the *first* band is on-scale. The mirror case (`priceTop` above the axis top) leaves a zero-height safe-zone rect still painting its 1px stroke.

3. **Block edges are not price positions, and nothing says so.** The safe-zone box and the level lines/pills are price-anchored; these edges are index-anchored. Same vertical axis, two mappings. With this PR's own Timeline fixture (`bands` + `timelinePriceAxis`, 1016px) band 2's block starts at ~162px — which reads as ~$65k against the axis — while its `$40,283` level line sits at ~325px. I take it that's the intent of e5220db8, but it wants a design sign-off and an explicit JSDoc line that a block edge carries no price.

packages/babylon-core-ui/src/components/LiquidationChart/BandLayer.tsx
```diff
@@ -1,150 +1,145 @@
...
+          hideBandLabels,
+          fontLabel,
+          fontAmount,
+          liquidatedLabel,
+        );
+        const lineHeights = lines.map((line) => Math.round(line.fontSize * TEXT_LINE_HEIGHT));
+        const stackHeight =
+          lineHeights.reduce((sum, h) => sum + h, 0) + Math.max(0, lines.length - 1) * BAND_LINE_GAP_PX;
+        const maxTextWidth = rect.width - 2 * BAND_PAD_X_PX;
+        const clipId = `${clipBaseId}-${band.key}`;
```
gbarkhatov: `clipId` interpolates a caller-supplied `band.key` into a DOM id that's dereferenced as `url(#…)` at :125. A key containing whitespace or a quote produces an invalid reference, and browsers drop the entire clipped `<g>` rather than ignoring the clip — the band's text vanishes with no error and no console warning. `types.ts` types `key` as an unconstrained `string` on a published package; the vault happens to pass `String(group.index)`. Using the map index (or sanitizing) removes the whole class.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
```diff
@@ -0,0 +1,370 @@
...
+    return { pills: [], hiddenTickValues: new Set<number>() };
+  }
+  const items: DeclutterItem[] = markers.map((m) => ({ key: m.key, center: m.center, height: pillHeight }));
+  const tops = declutterCenters(items, layout.plotHeight);
+  const hidden = new Set<number>();
+  for (const tick of priceAxis) {
+    const tickY = priceScale(tick.value);
+    for (const m of markers) {
+      const top = tops.get(m.key);
+      if (top != null && Math.abs(top - tickY) < pillHeight) {
```
gbarkhatov: The suppression radius is a full `pillHeight`, but the collision here is pill (`pillHeight`) vs tick label (`fontAxis * TEXT_LINE_HEIGHT`, i.e. `pillHeight - 2 * PILL_PAD_Y_PX`). The geometric threshold is half the sum — ~17px at the reference layout, not the 19px used — so ticks with ~2px of real clearance get dropped.

Minor, same block: `tops` holds centres, not tops, which reads as the opposite of what :172 does with it.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
```diff
@@ -0,0 +1,370 @@
...
+        break;
+      }
+    }
+  }
+  const pills: PillGeom[] = markers.map((m) => ({
+    key: m.key,
+    label: m.label,
+    tone: m.tone,
+    current: m.current,
+    x: axisSide === "right" ? layout.plotWidth + PILL_AXIS_GAP_PX : 0,
```
gbarkhatov: The `left` branch puts the pill at `x = 0` — inside the plot, on top of the bands — rather than in the gutter; the mirror of the right-hand case is `-(PILL_AXIS_GAP_PX + width)`. Unreachable today (only the Timeline renders pills, always `axisSide="right"`), so per the zero-dead-code policy this is either a bug to fix or a branch to delete.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
```diff
@@ -0,0 +1,370 @@
...
+    width: measureText(m.label, font, AXIS_LETTER_SPACING_PX) + 2 * PILL_PAD_X_PX,
```
gbarkhatov: Pill labels — plus the axis ticks at :334 and the price-line labels at :292 — are the only text in the chart with no width budget. Everything else goes through `truncateToWidth`, and `LiquidationChart.css:48` is deliberately `overflow: visible`, so there is no backstop either. The gutter ceilings at 68px and floors at 48px, while a pill is `text + 12px` starting at `plotWidth + 4`: `$88,700` at the 10px axis-font floor is already ~53px against 44px of available gutter, and it paints onto whatever sits beside the chart.

This is parity with the old CSS (`white-space: nowrap` in a fixed-width gutter), so not a regression — but the width is computed right here now, so clamping it and running the label through `truncateToWidth` is a couple of lines.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
```diff
@@ -0,0 +1,370 @@
...
+      ? xAxisLabels.map((label, i) => ({
+          fraction: xAxisLabels.length < 2 ? 0 : i / (xAxisLabels.length - 1),
+          label,
+        }))
+      : undefined;
+
+  return (
+    <div ref={parentRef} className={twJoin("bbn-liq-chart", `bbn-liq-chart--gridstyle-${gridStyle}`, className)}>
+      <div className="bbn-liq-chart__canvas">
+        <svg className="bbn-liq-chart__svg" width={layout.chartWidth} height={layout.svgHeight}>
```
gbarkhatov: The chart is now a single `<svg>` with no `role="img"`, `aria-label`, or `<title>`, and everything informative inside it is `aria-hidden` — axis ticks (:344), pills (:307), gridlines, level lines, crosshair. What remains for a screen reader is the raw band text, `Liq Event 1 (contain vault 1) 0.6 BTC`: no prices, no indication it is a chart. The old HTML render at least had DOM structure and `role="button"` bands. `role="img"` plus an `aria-label` summarising the current price and the first trigger would close it — this is the naming half of the a11y regression already raised on the keyboard side.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
```diff
@@ -0,0 +1,370 @@
...
+                  {pill.label}
+                </Text>
+              </g>
+            ))}
+          </Group>
+
+          {priceAxis.map((tick) =>
+            hiddenTickValues.has(tick.value) ? null : (
+              <Text
+                key={tick.value}
```
gbarkhatov: `key={tick.value}` is unique only because `createSegmentedPriceScale` asserts strict descent — and that assert isn't on the Timeline path (`createLinearPriceScale`), which consumes the same `PriceAxisTick[]`. A duplicated tick value there produces duplicate React keys. `hiddenTickValues` as a `Set<number>` carries the same assumption.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
```diff
@@ -0,0 +1,370 @@
...
+  layout: ChartLayout;
+} {
+  const { parentRef, width } = useParentSize({
+    debounceTime: 16,
+    initialSize: { width: FALLBACK_CHART_WIDTH_PX },
+  });
+  // Re-render when a webfont finishes loading so text measured against the
+  // fallback font is redone with the real metrics (see textMeasure.ts).
+  useSyncExternalStore(subscribeFontEpoch, getFontEpoch, getFontEpoch);
+  const chartWidth = width > 0 ? width : FALLBACK_CHART_WIDTH_PX;
```
gbarkhatov: A container that genuinely measures 0 falls back to 1016px instead of collapsing. With `overflow: visible` on the SVG, a chart in a zero-width flex child (a missing `min-width: 0`, a hidden tab) then paints a 1016px chart across its siblings rather than rendering nothing. Since `useParentSize` is seeded with `initialSize`, `width` is already 1016 before the first measurement — so this branch only fires on a real zero, and could return a null/zero-height frame instead.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
```diff
@@ -0,0 +1,370 @@
...
+  const { axisSide, hasTopLegend, hasXAxis } = input;
+  const layout = useMemo(
+    () => computeChartLayout({ chartWidth, axisSide, hasTopLegend, hasXAxis }),
+    [chartWidth, axisSide, hasTopLegend, hasXAxis],
+  );
+  return { parentRef, layout };
+}
+
+/** An x-axis tick placed at an explicit fraction [0,1] of the plot. */
+export interface AxisTick {
```
gbarkhatov: Duplicates the inline shape at `types.ts:110` (`shareAxisTicks?: { fraction: number; label: string }[]`). Since `index.ts` no longer re-exports `ChartFrame`, this copy isn't reachable from the package either — worth putting the type in `types.ts` and having both sites use it.

packages/babylon-core-ui/src/components/LiquidationChart/Timeline.tsx
```diff
@@ -172,50 +190,45 @@ export function Timeline({
     if (!windowed.length) return timeAxisLabels;
     const ticks = Math.min(7, windowed.length);
     return Array.from({ length: ticks }, (_, i) => {
-      const idx = Math.round((i / (ticks - 1)) * (windowed.length - 1));
+      // A single candle gets a single tick; the even spread needs ticks >= 2.
+      const idx = ticks < 2 ? 0 : Math.round((i / (ticks - 1)) * (windowed.length - 1));
```
gbarkhatov: The label names candle `idx`, but `ChartFrame` draws it at the even fraction `i / (ticks - 1)` (:355) while `layoutCandles` centres candle `idx` at `(idx + 0.5) / n`. Those never coincide: at `visibleCandles={44}` on the 1016px reference the offset is a constant ~8.4px — exactly half a slot — so every tick sits on the boundary between the candle it names and its neighbour. `GridColumns` at :319 inherits the same fractions.

The math is carried over from main, but `ChartFrame` already accepts `xAxisTicks: {fraction, label}[]` (the Seizure Map uses it), so emitting `fraction: (idx + 0.5) / windowed.length` fixes the labels and the gridlines together.

packages/babylon-core-ui/src/components/LiquidationChart/Timeline.tsx
```diff
@@ -172,50 +190,45 @@ export function Timeline({
...
+      const deltaFrac = regionWidth === 0 ? 0 : (e.clientX - drag.current.x) / regionWidth;
       const deltaIndex = Math.round(deltaFrac * windowSize);
       const next = Math.min(maxStart, Math.max(0, drag.current.start - deltaIndex));
       setStartIndex(next === maxStart ? null : next);
       return;
     }
     if (crosshairEnabled) {
-      const f = regionFraction(e.clientX);
+      const f = regionFraction(e);
       setHoverIndex(Math.min(windowed.length - 1, Math.max(0, Math.floor(f * windowed.length))));
```
gbarkhatov, line 276: `hoverIndex` indexes `windowed`, and nothing invalidates it when the window changes — a wheel-zoom or pan while hovering leaves the crosshair and the OHLC readout describing a *different* candle until the next pointermove. Carried over from main, but cheap to fix here: reset on `clampedStart`/`windowSize` change, or track the hovered candle's `time` rather than its index.

Related: `hovered` at :239 is annotated `CandleGeom | null` while `candleGeom[hoverIndex]` can be `undefined` once the window shrinks.

packages/babylon-core-ui/src/components/LiquidationChart/Timeline.tsx
```diff
@@ -138,7 +156,7 @@ export function Timeline({
   // Wheel zoom needs a native non-passive listener (React's root wheel
   // listener is passive, so preventDefault would be ignored).
   useEffect(() => {
-    const el = regionRef.current;
+    const el = hitRef.current;
     if (!el || !zoomEnabled) return;
     const onWheel = (e: WheelEvent) => {
       e.preventDefault();
```
gbarkhatov, line 207: `preventDefault()` is unconditional, so once the window bottoms out at `MIN_ZOOM_CANDLES` (or covers every candle) the chart still swallows page scroll with nothing left to zoom. Guarding on whether `zoomBy` would actually change `windowSize` hands the page back at the ends.

Same effect: `zoomBy` closes over `clampedStart`, so this native listener is torn down and re-added on every pan tick — a ref keeps it stable. Both carried over from main.

packages/babylon-core-ui/src/components/LiquidationChart/chartGeometry.ts
```diff
@@ -0,0 +1,122 @@
...
+const FONT_AXIS: FluidSize = { minRem: 0.625, cqiPct: 1.12, maxRem: 0.75 };
+const FONT_LABEL: FluidSize = { minRem: 0.7, cqiPct: 1.3, maxRem: 0.875 };
+const FONT_AMOUNT: FluidSize = { minRem: 0.75, cqiPct: 1.5, maxRem: 1 };
+
+const DEFAULT_ROOT_FONT_PX = 16;
+
+/** Root font-size in px (the CSS clamps were in `rem`). */
+function rootFontPx(): number {
+  if (typeof document === "undefined") return DEFAULT_ROOT_FONT_PX;
+  const parsed = Number.parseFloat(getComputedStyle(document.documentElement).fontSize);
```
gbarkhatov, line 35: `getComputedStyle(document.documentElement)` during render puts a live style read back into the layout path this file exists to make measurement-free, and it makes `computeChartLayout` environment-dependent — which is why `chartGeometry.test.ts` has to assume a 16px root without saying so. The root font-size effectively never changes at runtime; a module-scope memo would make the header comment's "no measurement passes" literally true.

packages/babylon-core-ui/src/test/setup.ts
```diff
@@ -0,0 +1,13 @@
...
+// @visx/responsive. The stub never fires, so components render from their
+// deterministic fallback width (see FALLBACK_CHART_WIDTH_PX in ChartFrame),
+// which is exactly what layout assertions in the chart tests rely on.
+class ResizeObserverStub {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
+
+global.ResizeObserver = ResizeObserverStub as unknown as typeof ResizeObserver;
```
gbarkhatov, line 13: Worth stubbing `getContext("2d")` here as well. jsdom has no canvas backend, so `measureText` returns 0 in every component test → `truncateToWidth` is the identity function and every pill collapses to `width = 12`. The declutter / truncation / pill geometry — the central claim of the rewrite — is therefore covered only by the hand-stubbed `textMeasure.test.ts`, and the overflow noted at `ChartFrame.tsx:171` ships green. A fixed per-character width stub would let the component tests assert real pill boxes.

packages/babylon-core-ui/src/components/LiquidationChart/__tests__/Timeline.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  it("sizes the gutter blocks equally regardless of their price spans", () => {
+    // Event 1 spans $37k of price, event 2 spans $183 — the gutter still
+    // gives them the same block height; only the axis pills carry the prices.
+    renderTimeline();
+    const h1 = Number.parseFloat(screen.getByTestId("liq-band-1").getAttribute("height") ?? "0");
+    const h2 = Number.parseFloat(screen.getByTestId("liq-band-2").getAttribute("height") ?? "0");
+    expect(h1).toBeGreaterThan(0);
+    expect(h1).toBeCloseTo(h2, 6);
+  });
```
gbarkhatov, line 174: Good coverage of the render paths. The gap is the interaction state machine — the most stateful code in the rewrite, with no tests at all: `zoomBy`'s right-edge anchoring, the `end >= candles.length ? null` pin that keeps the view on the latest candle, pan clamping to `maxStart`, `resetView`, and crosshair indexing. `axisDeclutter`'s own documented failure mode (a stack taller than the track re-overlapping) is untested too.

gbarkhatov (dismissed):
Approving — the architecture change is right and unusually well argued. Moving the CSS `@container`/`cqi` semantics into `chartGeometry.ts` as plain numbers is the correct call, `createSegmentedPriceScale` is a clean replacement for the hand-rolled math, and the JSDoc explaining *why* each constant exists is genuinely better than what it replaces. Nothing below blocks merge; the ones I'd want resolved first are the `SeizureGutter` geometry comment and the `LiquidationAnalysisSection` behaviour change.

### `textMeasure.ts` contains NUL bytes and is unreviewable

Not inlineable, because there is no diff to attach it to — that is the finding. GitHub renders the file as `Binary files /dev/null and b/.../textMeasure.ts differ`, and `file(1)` reports `data`. The cause is line 69:

```
const key = `${font}\0${letterSpacingPx}\0${text}`;
```

where the two `\0` are **literal NUL bytes (0x00) written into the source**, not escape sequences (bytes 2454 and 2473). Git classifies the file as binary, so it has no reviewable diff, no line-level blame, no textual merge, and no inline review comments — permanently, until the bytes are removed. The delimiter choice itself is good (it's what makes the cache key unambiguous); it just needs to be the two-character escape `\0` in the source. Same runtime value, plain-text file.

### Verified as not-issues — no need to chase these

- No dangling references to `popoverMetrics` / `cumulativeLabel` / `bandClickHint` / `onBandClick` / `.bbn-liq-popover` anywhere in tracked sources, and `formatSignedPct` / `totalBtc` / `combinedBtc` all remain used in `liquidationChartData.ts`. The removal is clean.
- `dts()`'s `exclude: ["src/**/*.stories.(ts|tsx)"]` does work — the current `dist` contains zero `*.stories.d.ts`.
- `jsdom@20.0.3` and `@testing-library/dom@10.4.0` match `babylon-wallet-connector`; the syncpack alignment is deliberate, not staleness.
- CI does pick the new suite up: `verify.yml` runs `nx affected --target=test` and nx infers `test` from the new package.json script.
- Vault is a Vite SPA, not Next.js (CLAUDE.md is drifted on this), so there is no SSR render path — the canvas-less `measureText` behaviour only affects jsdom and downstream library consumers.

### Carry-overs, flagged so they aren't read as regressions

The half-slot time-label offset, the stale `hoverIndex`, the unconditional wheel `preventDefault()`, and the candle `key={c.time}` all exist identically on `main`. I've raised them because this PR rewrites those files wholesale and the fixes are now one-liners — not because they're new.

Out of scope, but while you're in the area: `axisDeclutter.ts:22` (untouched by this PR) has a doc comment that begins `ponytail: assumes the labels' total height fits`, which looks like an editing artifact.

I did not execute the suite — the new dependencies aren't installed in my checkout, so the test observations above are from reading, not from a run.

jonybur (commented):
Three findings in the newest push (the overview simulator surface), none previously raised on this PR:

1. **Event-card figures are priced at the simulated price, not each event's trigger price.** `LiquidationChartPanel` rebuilds cards via `buildLiquidationChartData(cascade.result, { btcPrice: effectivePrice })`, and `toCard` computes `hfAfterLabel` (`hfAfter(group, btcPrice, cf)`) and the full-group fairness wBTC conversion (`fairnessPaymentUsd / btcPrice`) from that price. Dragging the simulator to $50k re-values the aftermath of an event that triggers at $77k using $50k. #2151 hit this exact defect and fixed it by pricing each event's figures at its own `liquidationPrice`, with regression tests - as is, the overview cards will disagree with the dashboard's cards for the same position once both land.

2. **`toCard` and the band labels key off the calculator's 1-based `index`.** `calculate()` emits `group.index` starting at 1 (main loop `groupIndex = 1`; the dust path also emits `index: 1`), but `toCard` sets `sacrificial = group.index === 0` - never true on real output, so the sacrificial badge can never render - and titles with `eventTitle(group.index + 1)`, so the first event renders as event 2 (the bands' `label` at `liquidationChartData.ts:219` has the same shift). The new section test masks this with a hand-built `index: 0` fixture. The bug is latent in `toCard` on main, but this PR ships its first consumer (`LiquidationEventCards`), which makes it live. #2151 fixed the identical issue by keying badges/titles/tones off array position, pinned by a test driven from a real `calculate()` cascade - the same fix applies here.

3. **The simulator slider cannot return to the live price.** The section passes `min={0} step={SIM_PRICE_STEP_USD} max={cascade.btcPrice}`, and with `steps={[]}` the core-ui `Slider` falls through to the native range input, whose emitted values align to `min + k*step`. The live BTC price is a float, so dragging fully right lands on the highest $100 multiple below it, `effectivePrice < livePrice` keeps `simulating` true, and the Simulation chip sticks until Reset is clicked. Same class of bug as #2151's toolbar, fixed there by aligning the step grid so the live price is reachable.

Minor: the PR body still says "The vault call site is unchanged; its band projection just stops producing popover data" - no longer accurate now that the section ships a simulator, a header summary, and event cards; worth updating so reviewers know the surface grew after the earlier rounds.

packages/babylon-core-ui/package.json
```diff
@@ -57,6 +58,9 @@
     "@storybook/react": "^8.6.15",
     "@storybook/react-vite": "^8.6.15",
     "@storybook/test": "^8.6.15",
```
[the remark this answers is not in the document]
jonybur: Fixed in a12123e67: the three test deps are pinned exact (vitest 4.0.9, @testing-library/react 16.3.0, @testing-library/jest-dom 6.5.0), and the same pins were synced workspace-wide in fd63f5d2d to satisfy syncpack.

packages/babylon-core-ui/src/components/LiquidationChart/priceScale.ts
jonybur: Fixed in the next push: createLinearPriceScale now degrades reversed or non-finite bounds to the flat degenerate scale (no inversion, no NaN coordinates), with a unit test covering reversed, NaN, and -Infinity bounds. The Timeline's main path now goes through createAnchoredPriceScale, which asserts strictly-descending stops.

packages/babylon-core-ui/src/components/LiquidationChart/SeizureMap.tsx
jonybur: Fixed in the next push: shareScale gets clamp: true and bandRect floors width/height at 0.

packages/babylon-core-ui/src/components/LiquidationChart/SeizureMap.tsx
jonybur: Fixed in the next push: clamp: true on shareScale, plus Math.max(0, ...) on both band width and height in bandRect - reversed bands render empty instead of painting outside the plot.

packages/babylon-core-ui/src/components/LiquidationChart/priceScale.ts
jonybur: Addressed in the next push at the layer you suggested: the vault projection filters non-finite values out of the axis (a price-feed NaN degrades the axis instead of feeding the assertion), and createLinearPriceScale degrades bad bounds to a flat render. The strictly-descending throw stays for genuinely malformed programmer-supplied axes, which the projection can no longer produce from feed data - regression-tested on both sides.

packages/babylon-core-ui/src/components/LiquidationChart/types.ts
jonybur: Resolved across later pushes: the hover tooltip is back per the design spec (39a3003bb - outlined chip, At price/Distance/Vaults/Seizes, Cumulative row) with popoverMetrics/cumulativeLabel restored to the contract and the vault projection. Product explicitly dropped the click interaction (hover-only), so onBandClick/bandClickHint stay removed - but keyboard users now get the tooltip on focus (tabIndex + focus/blur in the next push), plus an aria-label on each hoverable band.

packages/babylon-core-ui/src/components/LiquidationChart/SeizureMap.tsx
jonybur: Fixed in the next push: each legend segment's text is wrapped in a per-segment clipPath, same as the band and gutter text, so a canvas-less render clips instead of overflowing.

packages/babylon-core-ui/src/components/LiquidationChart/LiquidationChart.css
jonybur: Fixed in the next push: ChartFrame emits --liq-axis-letter-spacing from AXIS_LETTER_SPACING_PX on the chart root, and the three CSS rules read the var (the literal remains only as the var fallback).

packages/babylon-core-ui/vitest.config.ts
jonybur: Fixed in the next push: the dts plugin now excludes src/**/__tests__/** and src/test/**; verified the rebuilt dist contains no test or setup .d.ts.

packages/babylon-core-ui/src/components/LiquidationChart/index.ts
jonybur: Agreed. The PR description now carries a Breaking changes section, and the squash commit needs a BREAKING CHANGE: footer - final surface delta: SeizureGutter/SeizureGutterProps unexported, onBandClick/bandClickHint removed (hover-only per product), popoverMetrics/cumulativeLabel retained.

packages/babylon-core-ui/package.json
jonybur: Fixed in the next push: all seven @visx/* packages move to devDependencies. They are bundled into dist, and the published type graph (index -> components -> types) never references a visx type - priceScale.d.ts does, but nothing reachable from the entry imports it - so consumers get neither a double install nor a type hole. Verified against the rebuilt dist.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
jonybur: Right on both counts - this and the popover round-trips are product decisions that arrived mid-PR via the design tooltip spec and the chart design doc. The PR description is now rewritten to name every behaviour change (liquidated label swap, hover tooltip, span-weighted timeline scale, simulator surface) rather than reading as a pure refactor.

packages/babylon-core-ui/src/components/LiquidationChart/SeizureGutter.tsx
jonybur: Superseded by the span-weighted scale (39a3003bb + next push): gutter blocks now tile trigger-to-trigger off priceScale, and the anchored scale asserts strictly-descending stops - a misordered bands array now throws the same loud error as a misordered axis instead of rendering upside-down. Zero-height collapse below the axis floor is now documented as the contract (extend the axis to reveal the event), and the vault's floor (min trigger x 0.95) guarantees it can't happen in production.

packages/babylon-core-ui/src/components/LiquidationChart/BandLayer.tsx
jonybur: Fixed in the next push: clip ids interpolate the array index, never the caller-supplied key.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
jonybur: Fixed in the next push: the suppression radius is now (pillHeight + tickTextHeight) / 2, and the map is named centers with a comment noting declutterCenters returns centres.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
jonybur: Fixed in the next push: the left branch mirrors the right one - x = -(PILL_AXIS_GAP_PX + width) - rather than deleting it, since it's a live branch of the axisSide prop.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
jonybur: Keeping this as parity, deliberately: truncating a price misreads as a different number, which is worse than overflow at widths the design doesn't target. If narrow layouts become a real surface, the fix should be abbreviated formatting ($88.7k) at the app layer, which owns formatting. Happy to open a follow-up issue if you want it tracked.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
jonybur: Fixed in the next push: the SVG gets role=img with an aria-label assembled from the price caption + current price label, and hoverable bands are focusable with per-band aria-labels, opening the tooltip on focus - covering both halves of the a11y regression.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
jonybur: Fixed in the next push: axis tick keys are value:index, so duplicated values on the linear/anchored path can't collide.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
jonybur: Fixed in the next push: useChartLayout reports collapsed when the ResizeObserver measures a real 0 (the initial seed is the fallback, so this only fires on genuine collapse), and both charts render nothing in that state instead of a 1016px chart across siblings.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
jonybur: Fixed in the next push: the shape lives in types.ts as ShareAxisTick; ChartFrame and SeizureMapProps both use it.

packages/babylon-core-ui/src/components/LiquidationChart/Timeline.tsx
jonybur: Fixed in the next push: the Timeline emits positioned xAxisTicks at (idx + 0.5) / n - each label sits on its candle's centre - and the vertical gridlines consume the same fractions.

packages/babylon-core-ui/src/components/LiquidationChart/Timeline.tsx
jonybur: Fixed in the next push: a clampedStart/windowSize change drops the hover index, and the lookup is candleGeom[hoverIndex] ?? null so a shrunken window can't render a stale readout.

packages/babylon-core-ui/src/components/LiquidationChart/Timeline.tsx
jonybur: Fixed in the next push: the wheel handler reads window/candle state through a ref (listener attaches once per zoomEnabled), and only calls preventDefault when the zoom can actually act - at the floor/ceiling the page gets its scroll back.

packages/babylon-core-ui/src/components/LiquidationChart/chartGeometry.ts
jonybur: Fixed in the next push: the root font-size is read once and memoised at module scope; the layout path is measurement-free after the first call.

packages/babylon-core-ui/src/test/setup.ts
jonybur: Fixed in the next push: the shared test setup stubs getContext(2d) with a fixed 7px-per-character advance, so component tests exercise real truncation and pill geometry; textMeasure.test.ts keeps its own stubs (including the no-canvas path) via module reset.

packages/babylon-core-ui/src/components/LiquidationChart/__tests__/Timeline.test.tsx
jonybur: Partially addressed in the next push: added interaction tests for the zoom controls (zoom in -> 35 candles, zoom out, reset restores the default window) and a declutter test documenting the taller-than-track degradation (order preserved, bottom clamped). Pan needs setPointerCapture, which jsdom doesn't implement - that slice stays for a browser-driven test, noted as a follow-up.

packages/babylon-core-ui/src/components/LiquidationChart/priceScale.ts
jonybur: Fixed in the next push, together with the thread above: createLinearPriceScale degrades reversed or non-finite bounds to the flat degenerate scale (no silent inversion, no NaN blank render), unit-tested for reversed/NaN/-Infinity. The Timeline's primary path is now createAnchoredPriceScale, which asserts strictly-descending stops, and the vault projection filters non-finite feed values before they reach either scale - so the two components no longer carry opposite contracts for the same PriceAxisTick[] type.

jonybur:
Addressed the full review backlog; the fixes land in the next push (every inline thread has a specific reply).

**Daily-review findings** (both reviews):
- `textMeasure.ts` is a plain text file again — the cache-key NUL delimiters are now written as escape sequences, identical at runtime.
- The timezone-dependent `"Jun 15"` assertion now uses an explicit `formatTime`, so the suite passes in any TZ.
- Event-card aftermath figures (post-liquidation HF, fairness wBTC) are priced at each event's **own trigger**, matching the #2151 semantics; regression-tested.
- Titles/badges/keys follow **array position**, not the calculator's 1-based `group.index`; the section test fixture now uses a 1-based index like real `calculate()` output, and a projection test pins it.
- The simulator slider steps by $1, so a full drag right lands exactly on the float live price and the Simulation chip clears.

**Not changed, by design** (see individual replies): price pills/ticks are never ellipsized (a truncated price misreads); the band click interaction stays removed per product (hover-only), with keyboard focus now opening the tooltip instead.

The PR description is rewritten to name every behaviour change and the breaking package surface. The squash commit should carry a `BREAKING CHANGE:` footer (`SeizureGutter`/`SeizureGutterProps` unexported; `onBandClick`/`bandClickHint` removed).

jonybur:
@greptile-apps

jonybur (commented):
`services/vault/src/components/simple/LiquidationAnalysisSection.tsx` — the slider passes `min={0}, max={cascade.btcPrice}, step={SIM_PRICE_STEP_USD}` with `SIM_PRICE_STEP_USD = 1`. The comment on the constant claims $1 granularity lets a full drag right land on the live price, but the underlying `cascade.btcPrice` comes from `prices["BTC"]` (a Chainlink-scaled float, not a rounded integer), and `LiquidationCascade.btcPrice: number` places no integer constraint on it. Native `<input type="range">` snaps its reachable values to `min + k*step`, so for a price like 88_412.37 the highest reachable value is 88_412. Dragging fully right emits 88_412, `effectivePrice = min(88_412, 88_412.37) = 88_412`, and `simulating = 88_412 < 88_412.37 = true` — the Simulation chip sticks and Reset stays enabled even though the user believes they returned to live. #2151 solved this in `SimulationToolbar` by aligning the slider's own min so the live price sits on a reachable step; the equivalent fix here is either to align min/max to the step grid or to make the `simulating` predicate step-tolerant (e.g. `livePrice - effectivePrice >= SIM_PRICE_STEP_USD`).

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -104,16 +134,62 @@ function LiquidationChartPanel({
...
+          <Slider
+            value={effectivePrice}
+            min={0}
+            max={cascade.btcPrice}
+            step={SIM_PRICE_STEP_USD}
+            steps={[]}
+            onChange={onSimulate}
+            activeColor={PRICE_SLIDER_COLOR}
+            aria-label={COPY.liquidations.simulateLabel}
+          />
```
gbarkhatov, line 150: The Aug 3 daily review flagged this, and it's the one thing on the PR still open — nothing pushed since `38f2b37a`, and because it landed as a top-level review body rather than an inline thread it can't be resolved and now sits below 28 resolved threads, so it's easy to scroll straight past. The finding is correct. Two corrections to it.

**It's reproducible today, not once prices are wired.** The bot says `cascade.btcPrice` comes from `prices["BTC"]`; it doesn't — `DashboardPage.tsx:80-92` builds the cascade only from the god-mode debug store, and the **Dust** preset is `btcPrice: 61722.5` (`dev/debugPositionStore.ts:136`). Load it and drag the slider fully right: the native range's reachable values are `min + k * step`, i.e. integers, so it emits `61722`, `effectivePrice = min(61722, 61722.5) = 61722`, and `simulating` stays true at `:213`. Simulation chip sticks, Reset stays enabled.

**The suggested predicate only fixes the chip.** `livePrice - effectivePrice >= SIM_PRICE_STEP_USD` clears the badge but leaves `effectivePrice` at `61722`, so the price line, the header figures and the event cards all keep rendering a simulated price while the UI claims live. Five consumers read `effectivePrice`; snapping it once is what makes them agree:

```ts
const rawPrice = simulatedPrice === null ? livePrice : Math.min(simulatedPrice, livePrice);
// A native range input snaps to `min + k * step`, so a float live price is never
// exactly reachable; anything within one step of it reads as live.
const effectivePrice = livePrice - rawPrice < SIM_PRICE_STEP_USD ? livePrice : rawPrice;
const simulating = showChart && effectivePrice < livePrice;
```

Two things in the same block while you're in here.

The JSDoc on `SIM_PRICE_STEP_USD` (`:39-43`) doesn't just miss this — it asserts the opposite: *"Must be 1: … a coarser grid would make the top of the range (the live price) unreachable."* Step 1 doesn't make a float reachable either; only the step base does. Whatever the fix, that comment has to be rewritten to say what the constant actually buys (granularity), or the next reader will trust it and walk straight back into this.

`steps={[]}` (`:146`) reads as a no-op but is load-bearing: an empty array takes `Slider`'s `Array.isArray(steps)` branch in `computedSteps` while failing the `length > 0` guard in `handleChange`, which is exactly what suppresses the five default dots *and* the snapping. Drop the prop and the dots come back. Worth one line of comment saying so.

For the test: `__tests__/LiquidationAnalysisSection.test.tsx` only exercises integer prices right now. A case with `btcPrice: 61722.5` that fires the slider at `61722` and asserts the Simulation chip is absent and Reset is disabled would pin all of it.

packages/babylon-core-ui/src/components/LiquidationChart/ChartFrame.tsx
```diff
@@ -0,0 +1,400 @@
...
+    for (const m of markers) {
+      const center = centers.get(m.key);
+      if (center != null && Math.abs(center - tickY) < collisionRadius) {
+        hidden.add(tick.value);
+        break;
+      }
+    }
+  }
+  const pills: PillGeom[] = markers.map((m) => {
+    const width = measureText(m.label, font, AXIS_LETTER_SPACING_PX) + 2 * PILL_PAD_X_PX;
```
gbarkhatov, line 175: Following up on the resolved thread rather than reopening it: I agree with the call. A truncated price misreads as a different number, and this is parity with `main`, so nothing regressed here.

But "happy to open a follow-up issue if you want it tracked" never became an issue, and this PR strengthened the case for tracking it. The width is computed right here in `computePills` now, so the fix finally has a home it didn't have before; and `LiquidationChart.css:53` is deliberately `overflow: visible`, so nothing clips a pill that outgrows the gutter — which is `clamp(3rem, 7cqi, 4.25rem)`, i.e. 48–68px, while `$88,700` at the 10px axis-font floor already measures ~53px and paints onto whatever sits beside the chart.

Could you open it with the conclusion you reached in the thread — abbreviated formatting (`$88.7k`) at the app layer, which owns formatting — so the reasoning survives outside a PR thread?

packages/babylon-core-ui/src/components/LiquidationChart/__tests__/Timeline.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+    expect(safeHeight).toBeCloseTo(h1, 2);
+    expect(safeHeight / plotHeight).toBeGreaterThan(0.4);
+  });
+
+  it("dims no gutter block while the price line sits above the safe zone floor", () => {
+    const { container } = renderTimeline();
+    expect(container.querySelectorAll(".bbn-liq-band--liquidated")).toHaveLength(0);
+  });
+
+  it("zooms the candle window in and out and resets to the default view", () => {
```
gbarkhatov, line 176: The zoom/reset coverage here, plus the taller-than-track case in `axisDeclutter.test.ts`, closes most of what I raised. The remaining slice is pan, which you noted as a follow-up — no issue exists for that one either.

Worth filing rather than dropping: pan is now the only untested branch of the interaction state machine, and it's the one with clamp math that's easy to get subtly wrong — `Math.min(maxStart, Math.max(0, drag.current.start - deltaIndex))`, plus the `next === maxStart ? null` re-pin that hands the window back to "follow the latest candle". A `setPointerCapture` stub in `src/test/setup.ts` would unblock it in jsdom; a browser-driven test would cover it properly.

gbarkhatov (approved):
Re-approving. I re-checked all 28 inline threads against the code at `38f2b37a` rather than trusting the reply text, and every one of them landed — including the ones I'd have expected to slip. The span-weighted anchored scale genuinely supersedes the equal-gutter geometry comment rather than papering over it: blocks now tile trigger-to-trigger off `priceScale`, so a block edge *is* a price position, and a misordered `bands` array throws the same loud error as a misordered axis. `clipId` off the array index, the `(pillHeight + tickTextHeight) / 2` collision radius, the mirrored left-pill branch, `(idx + 0.5) / n` tick fractions shared with the gridlines, the hover-index reset, the ref-stable wheel handler with its `canAct` guard, the `collapsed` path, and the canvas stub in `src/test/setup.ts` are all in the tree and do what the replies say they do.

The two #2151-class defects the daily review found in the simulator surface are fixed at the right layer too: aftermath figures are priced at `group.liquidationPrice` rather than the simulated price, and titles/badges/keys follow array position rather than the calculator's 1-based `group.index`.

One finding is still open, and two follow-ups you offered were never filed — details inline. None of it blocks: the section is behind `NEXT_PUBLIC_FF_LIQUIDATION_ANALYSIS_CHART`, only reachable through god mode, and the open bug is a sticky chip rather than a wrong number.

One caveat on my verification: I read the tests, I didn't run them — `@visx/*` isn't installed in my checkout. Worth confirming `pnpm --filter @babylonlabs-io/core-ui run test` is green from a clean install before merge.

And don't lose the `BREAKING CHANGE:` footer on the squash commit (`SeizureGutter`/`SeizureGutterProps` unexported, `onBandClick`/`bandClickHint` removed).

PR #2166, opened 2026-07-30
chore(vault): gate the market detail page behind FF

jeremy-babylonlabs:


jrwbabylonlab (changes requested):
u forgot to add to the release pipeline

jonybur (approved):
The release-pipeline gap flagged earlier is closed: NEXT_PUBLIC_FF_ENABLE_MARKET_DETAIL_PAGE is now threaded through service-release-vault.yml, which is the only workflow that passes FF vars to the vault build. The gate itself is consistent across all three surfaces (route redirect, Market Info button, picker width), matches the Explore/liquidations double-flag pattern, defaults off, and both new tests pin the off state - deep link redirect and button omission. Docs and .env.example updated.

jrwbabylonlab (approved):

services/vault/src/applications/aave/components/AssetSelectionPanel/AssetSelectionPanel.tsx
```diff
@@ -92,7 +92,9 @@ const MARKET_INFO_COL_STYLE = {
 /** Borrow is wider than repay to fit the Market Info column without squeezing
  *  the stat columns (Figma 6058-44070 measures it at ~700px). */
 export function getAssetPickerWidthClass(mode: LoanTab) {
```
jrwbabylonlab, line 94: This re-implements the `showMarketInfo` predicate at `:118` in inverted form, with no test covering it (the only test that reaches it mocks the return value). When the two drift, the picker renders 700px with no Market Info column or 612px crushing the stat columns, and CI stays green. Export one predicate and use it at both sites.

services/vault/src/__tests__/router.test.tsx
```diff
@@ -349,6 +352,16 @@ describe("Router — new v3 placeholder routes", () => {
...
 
+  it("redirects /markets/:reserveId to / when the market detail flag is off, even with v3 on", async () => {
+    setV3Flag("true");
+    featureFlagsState.isMarketDetailPageEnabled = false;
+    renderAt("/markets/1");
+
+    await waitFor(() => {
+      expect(screen.getByTestId(DASHBOARD_TESTID)).toBeInTheDocument();
+    });
+  });
```
jrwbabylonlab, line 363: Only the flag-off state is covered. `/liquidations` (`:325`) and `/explore` (`:390`) both have on+off tests; nothing asserts `/markets/1` renders `BorrowingMarketsDataPage` with both flags on, which is the state this ships in.

PR #2167, opened 2026-07-30
fix(vault): read an EVM out-of-funds halt as insufficient ETH

jrwbabylonlab:
image
image

### DIFF
--- a/services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx
@@ -13,7 +13,14 @@
  */
 
 import { Button, Callout, Loader, Text } from "@babylonlabs-io/core-ui";
-import { type ReactNode, useCallback, useMemo } from "react";
+import {
+  type ReactNode,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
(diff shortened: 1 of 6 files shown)


jonybur (approved):
Reviewed the classifier change and the diagnostics affordance end to end; approving.

- The `EVM_OUT_OF_FUNDS_PATTERN` check sits below the user-rejection block in `classifyError`, preserving the documented precedence, and the pattern has no collision surface: the phrase appears nowhere else in the repo, and the BTC-side selector errors carry the `sats`/`pegin` guard vocabulary this check deliberately skips.
- The regression test rebuilds the captured failure with real viem classes (`EstimateGasExecutionError` -> `TransactionRejectedRpcError` -> `RpcRequestError`), so the shape can't silently drift with a viem upgrade.
- `readViemMessage` requires all three BaseError markers before trimming, so non-viem errors keep their richer `message` - the test pins that too.
- The copy affordance checks out: `Callout` really does support `actions` with `emphasis`, the clipboard write settles honestly (no optimistic "Copied", sync throw outside secure contexts caught), diagnostics only attach to the unclassified fallback bucket where the raw error adds information, the clamp keeps summary fields ahead of the truncatable raw message, and all new user-facing strings live in copy.ts.

CI is green across all checks.

PR #2168, opened 2026-07-30
ci: publish frontend checksum manifest as immutable release asset (#253)

toroviej0x0:
Companion to babylonlabs-io/babylon-dapp-deployment#321 (the frontend integrity checker) for devops-backlog#253.

## Why
The checker verifies that the JS/CSS a browser downloads from the live staking + vault dApps matches the sha256 manifest this CI produced for the deployed commit. For that it needs a **trusted, public** source of truth for the manifest — one an attacker who poisons the CDN/edge can't also rewrite.

CI already generates `<service>-<env>-<sha>-files-checksum.txt` and uploads it to a **private** S3 bucket. Private S3 forces credentials onto every checker (an AWS key inside the Cloudflare Worker). This PR also publishes it to a **public GitHub release**, so:
- Both checker runtimes (AWS Lambda + Cloudflare Worker) read it with **zero credentials**.
- It lives in a **different trust domain** (GitHub) than the AWS serving path — attacker must break two systems.
- It's literally "a commit from babylon-toolkit", matching the issue.

## What
- New step in both `service-release-*.yml` (staking + vault): after the S3 upload, `gh release upload` the files-checksum to a pinned release **`frontend-manifests`**.
- Deploys fire on push-to-main (per-sha), not tags — so one long-lived release acts as an **append-only, per-sha** asset store. Each `<service>-<env>-<sha>` file is written once ⇒ immutable in practice. `create ... || true` keeps it idempotent across the parallel env matrix.
- `contents: write` scoped to the `s3_publish` job only (lint_test stays read).

No change to the existing S3 upload — this is additive.

## Checker reads
`https://github.com/babylonlabs-io/babylon-toolkit/releases/download/frontend-manifests/<service>-<env>-<sha>-files-checksum.txt`

## Note
Asset store grows one file per deploy per env. Harmless (GitHub handles it); prune old ones later if desired.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/service-release-simple-staking.yml
+++ b/.github/workflows/service-release-simple-staking.yml
@@ -187,3 +187,13 @@ jobs:
               curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
             fi
           done
+          # Keep the release well under GitHub's 1000-asset/release cap. Only the
+          # currently-deployed sha per env is ever looked up, so older manifests
+          # are dead weight — keep the most recent 500, drop the rest. Best-effort
+          # (|| true) so a prune race across the env matrix can't fail the publish.
+          KEEP=500
+          gh release view "$RELEASE_TAG" --json assets \
+            --jq ".assets | sort_by(.createdAt) | reverse | .[${KEEP}:] | .[].name" \
+          | while IFS= read -r OLD; do
+              [ -n "$OLD" ] && gh release delete-asset "$RELEASE_TAG" "$OLD" --yes 2>/dev/null || true
+            done || true
--- a/.github/workflows/service-release-simple-staking.yml
+++ b/.github/workflows/service-release-simple-staking.yml
@@ -40,6 +40,9 @@ jobs:
 
   s3_publish:
     needs: [lint_test]
+    permissions:
+      id-token: write # OIDC assume-role for AWS
+      contents: write # publish checksum manifest as a release asset (#253)
     runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
@@ -138,6 +141,22 @@ jobs:
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}.tar.gz s3://${{ vars.S3_PUBLIC_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.sha }}.tar.gz
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.sha }}-files-checksum.txt
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.sha }}-global-checksum.txt
+
+      - name: Publish checksum manifest as immutable release asset
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Append-only per-deploy sha256 manifest for the frontend integrity
+          # checker (devops-backlog#253). Public + immutable per-sha => the
+          # checker reads it with no credentials, from a different trust domain
+          # than the AWS serving path. One pinned release acts as the store.
+          RELEASE_TAG=frontend-manifests
+          ASSET="/tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt"
+          gh release create "$RELEASE_TAG" \
+            --title "Frontend integrity manifests" \
+            --notes "Append-only sha256 manifests consumed by the frontend integrity checker (devops-backlog#253). Do not delete." \
+            --latest=false 2>/dev/null || true
+          gh release upload "$RELEASE_TAG" "$ASSET" --clobber
 
       - name: Upload to S3 with Git Tag
         if: startsWith(github.ref, 'refs/tags/')
--- a/.github/workflows/service-release-simple-staking.yml
+++ b/.github/workflows/service-release-simple-staking.yml
@@ -190,15 +190,23 @@ jobs:
               echo "slot ${TAG} already published — trying next"
               continue
             fi
-            if gh release create "$TAG" "${FILES[@]}" \
+            if ERR=$(gh release create "$TAG" "${FILES[@]}" \
                  --title "Frontend manifest ${PREFIX}" \
                  --notes "sha256 manifest for the frontend integrity checker (devops-backlog#253). Do not delete." \
-                 --latest=false; then
+                 --latest=false 2>&1); then
               curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
               echo "published ${TAG}"
               exit 0
             fi
-            echo "create ${TAG} lost a race — trying next slot"
+            # Advance to the next slot ONLY on a create race ("already exists").
+            # Any other failure must fail the step: skipping it would leave a
+            # numbering gap, and the checker stops at the first missing slot —
+            # the next run's higher slot would never be found (silent gap).
+            if ! grep -qi "already exists" <<<"$ERR"; then
+              echo "::error::gh release create ${TAG} failed: ${ERR}"
+              exit 1
+            fi
+            echo "slot ${TAG} already exists (lost create race) — trying next slot"
           done
           echo "::error::could not publish manifest for ${PREFIX} after 10 slots"
           exit 1
--- a/.github/workflows/service-release-simple-staking.yml
+++ b/.github/workflows/service-release-simple-staking.yml
@@ -154,7 +154,7 @@ jobs:
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.ref_name }}-global-checksum.txt
 
       # Last step: publishing the manifest must not block the S3 uploads above.
-      - name: Publish checksum manifests as write-once release assets
+      - name: Publish checksum manifest as an immutable per-sha release
         # Only real deploy refs — workflow_dispatch can run from any branch, and
         # the public store must hold manifests only for builds that shipped.
         if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/release/v1.x' || startsWith(github.ref, 'refs/tags/')
@@ -162,38 +162,36 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           set -euo pipefail
-          # Append-only per-deploy sha256 manifests for the frontend integrity
-          # checker (devops-backlog#253). Public, so the checker reads them with
-          # no credentials, from a different trust domain than the AWS serving path.
-          RELEASE_TAG=frontend-manifests
-          gh release view "$RELEASE_TAG" >/dev/null 2>&1 || \
-            gh release create "$RELEASE_TAG" \
-              --title "Frontend integrity manifests" \
-              --notes "Append-only sha256 manifests for the frontend integrity checker (devops-backlog#253). Do not delete." \
-              --latest=false || true
-          for ASSET in \
-            "/tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt" \
-            "/tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt"; do
-            NAME="$(basename "$ASSET")"
-            URL="https://github.com/${GITHUB_REPOSITORY}/releases/download/${RELEASE_TAG}/${NAME}"
-            # Write-once, checked against the URL the checker actually reads (not
-            # the producer's asset list): survives the create race, asset-list
-            # truncation and a draft release. Builds aren't byte-reproducible
-            # (run_id embedded), so a rerun must never swap a published manifest.
+          # Per-deploy sha256 manifests for the frontend integrity checker
+          # (devops-backlog#253). One release per (service, env, sha) with both
+          # files attached at create time: this repo has Immutable Releases
+          # enabled, so assets can't be added after publish — attaching at create
+          # is the immutable-safe path, and it makes write-once a platform
+          # guarantee (published releases can't be modified) and the two-file
+          # publish atomic. No asset-cap pruning needed (2 assets per release).
+          #
+          # Same-sha rebuilds are sanctioned (tag cut on an already-built commit,
+          # redeploy dispatch, job re-run) and are NOT byte-reproducible (run_id
+          # in the bundle), so each rebuild takes the next free -rN slot. The
+          # checker probes slots upward and verifies against the highest one.
+          PREFIX="simple-staking-${{ matrix.environment }}-${{ github.sha }}"
+          FILES=("/tmp/${PREFIX}-files-checksum.txt" "/tmp/${PREFIX}-global-checksum.txt")
+          for i in 1 2 3 4 5 6 7 8 9 10; do
+            TAG="fm-${PREFIX}"; [ "$i" -gt 1 ] && TAG="${TAG}-r${i}"
+            URL="https://github.com/${GITHUB_REPOSITORY}/releases/download/${TAG}/${PREFIX}-files-checksum.txt"
             if curl -sfIL -o /dev/null "$URL"; then
-              echo "manifest $NAME already published — keeping the original"
-            else
-              gh release upload "$RELEASE_TAG" "$ASSET"
-              curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
+              echo "slot ${TAG} already published — trying next"
+              continue
             fi
+            if gh release create "$TAG" "${FILES[@]}" \
+                 --title "Frontend manifest ${PREFIX}" \
+                 --notes "sha256 manifest for the frontend integrity checker (devops-backlog#253). Do not delete." \
+                 --latest=false; then
+              curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
+              echo "published ${TAG}"
+              exit 0
+            fi
+            echo "create ${TAG} lost a race — trying next slot"
           done
-          # Keep the release well under GitHub's 1000-asset/release cap. Only the
-          # currently-deployed sha per env is ever looked up, so older manifests
-          # are dead weight — keep the most recent 500, drop the rest. Best-effort
-          # (|| true) so a prune race across the env matrix can't fail the publish.
-          KEEP=500
-          gh release view "$RELEASE_TAG" --json assets \
-            --jq ".assets | sort_by(.createdAt) | reverse | .[${KEEP}:] | .[].name" \
-          | while IFS= read -r OLD; do
-              [ -n "$OLD" ] && gh release delete-asset "$RELEASE_TAG" "$OLD" --yes 2>/dev/null || true
-            done || true
+          echo "::error::could not publish manifest for ${PREFIX} after 10 slots"
+          exit 1
--- a/.github/workflows/service-release-simple-staking.yml
+++ b/.github/workflows/service-release-simple-staking.yml
@@ -59,8 +59,11 @@ jobs:
             aws_account: 490721144737
             prod: true
 
-    # Prod environments will fail OIDC on feature branches; allow failure so devnet jobs aren't blocked
-    continue-on-error: ${{ matrix.prod == true }}
+    # Prod environments will fail OIDC on feature branches; allow failure there
+    # so devnet jobs aren't blocked. On real deploy refs (main/release/tags) a
+    # prod failure must fail the run — a green run with a failed prod deploy or
+    # missing integrity manifest is a silent gap (#253 review).
+    continue-on-error: ${{ matrix.prod == true && github.ref != 'refs/heads/main' && github.ref != 'refs/heads/release/v1.x' && !startsWith(github.ref, 'refs/tags/') }}
     environment: ${{ matrix.environment }}
     steps:
       - name: Checkout repository
@@ -148,10 +151,14 @@ jobs:
 
       - name: Upload to S3 with Git Tag
         if: startsWith(github.ref, 'refs/tags/')
+        env:
+          # Bound via env, not interpolated into run: — git refnames accept
+          # shell metacharacters and this job holds AWS OIDC + contents:write.
+          GIT_TAG: ${{ github.ref_name }}
         run: |
-          aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}.tar.gz s3://${{ vars.S3_PUBLIC_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.ref_name }}.tar.gz
-          aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.ref_name }}-files-checksum.txt
-          aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.ref_name }}-global-checksum.txt
+          aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}.tar.gz "s3://${{ vars.S3_PUBLIC_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${GIT_TAG}.tar.gz"
+          aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt "s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${GIT_TAG}-files-checksum.txt"
+          aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt "s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${GIT_TAG}-global-checksum.txt"
 
       # Last step: publishing the manifest must not block the S3 uploads above.
       - name: Publish checksum manifest as an immutable per-sha release
--- a/.github/workflows/service-release-simple-staking.yml
+++ b/.github/workflows/service-release-simple-staking.yml
@@ -65,6 +65,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2 https://github.com/actions/checkout/releases/tag/v6.0.2
+        with:
+          # Job holds contents:write; don't leave the token in .git/config while
+          # pnpm install / build run untrusted deps.
+          persist-credentials: false
 
       - name: Set up Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0 https://github.com/actions/setup-node/releases/tag/v4.4.0
@@ -142,34 +146,44 @@ jobs:
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.sha }}-files-checksum.txt
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.sha }}-global-checksum.txt
 
-      - name: Publish checksum manifest as immutable release asset
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: |
-          # Append-only per-deploy sha256 manifest for the frontend integrity
-          # checker (devops-backlog#253). Public + immutable per-sha => the
-          # checker reads it with no credentials, from a different trust domain
-          # than the AWS serving path. One pinned release acts as the store.
-          RELEASE_TAG=frontend-manifests
-          ASSET="/tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt"
-          gh release create "$RELEASE_TAG" \
-            --title "Frontend integrity manifests" \
-            --notes "Append-only sha256 manifests consumed by the frontend integrity checker (devops-backlog#253). Do not delete." \
-            --latest=false 2>/dev/null || true
-          # Write-once: never overwrite an already-published manifest for this
-          # sha. Builds aren't byte-reproducible (run_id embedded), so a rerun
-          # with --clobber could swap the checker's source of truth. Keeping the
-          # original is fail-safe — a stale manifest alerts, it can't hide.
-          ASSET_NAME="$(basename "$ASSET")"
-          if gh release view "$RELEASE_TAG" --json assets --jq '.assets[].name' | grep -qxF "$ASSET_NAME"; then
-            echo "manifest $ASSET_NAME already published — keeping the immutable original"
-          else
-            gh release upload "$RELEASE_TAG" "$ASSET"
-          fi
-
       - name: Upload to S3 with Git Tag
         if: startsWith(github.ref, 'refs/tags/')
         run: |
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}.tar.gz s3://${{ vars.S3_PUBLIC_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.ref_name }}.tar.gz
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.ref_name }}-files-checksum.txt
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.ref_name }}-global-checksum.txt
+
+      # Last step: publishing the manifest must not block the S3 uploads above.
+      - name: Publish checksum manifests as write-once release assets
+        # Only real deploy refs — workflow_dispatch can run from any branch, and
+        # the public store must hold manifests only for builds that shipped.
+        if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/release/v1.x' || startsWith(github.ref, 'refs/tags/')
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          # Append-only per-deploy sha256 manifests for the frontend integrity
+          # checker (devops-backlog#253). Public, so the checker reads them with
+          # no credentials, from a different trust domain than the AWS serving path.
+          RELEASE_TAG=frontend-manifests
+          gh release view "$RELEASE_TAG" >/dev/null 2>&1 || \
+            gh release create "$RELEASE_TAG" \
+              --title "Frontend integrity manifests" \
+              --notes "Append-only sha256 manifests for the frontend integrity checker (devops-backlog#253). Do not delete." \
+              --latest=false || true
+          for ASSET in \
+            "/tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt" \
+            "/tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt"; do
+            NAME="$(basename "$ASSET")"
+            URL="https://github.com/${GITHUB_REPOSITORY}/releases/download/${RELEASE_TAG}/${NAME}"
+            # Write-once, checked against the URL the checker actually reads (not
+            # the producer's asset list): survives the create race, asset-list
+            # truncation and a draft release. Builds aren't byte-reproducible
+            # (run_id embedded), so a rerun must never swap a published manifest.
+            if curl -sfIL -o /dev/null "$URL"; then
+              echo "manifest $NAME already published — keeping the original"
+            else
+              gh release upload "$RELEASE_TAG" "$ASSET"
+              curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
+            fi
+          done
--- a/.github/workflows/service-release-simple-staking.yml
+++ b/.github/workflows/service-release-simple-staking.yml
@@ -156,7 +156,16 @@ jobs:
             --title "Frontend integrity manifests" \
             --notes "Append-only sha256 manifests consumed by the frontend integrity checker (devops-backlog#253). Do not delete." \
             --latest=false 2>/dev/null || true
-          gh release upload "$RELEASE_TAG" "$ASSET" --clobber
+          # Write-once: never overwrite an already-published manifest for this
+          # sha. Builds aren't byte-reproducible (run_id embedded), so a rerun
+          # with --clobber could swap the checker's source of truth. Keeping the
+          # original is fail-safe — a stale manifest alerts, it can't hide.
+          ASSET_NAME="$(basename "$ASSET")"
+          if gh release view "$RELEASE_TAG" --json assets --jq '.assets[].name' | grep -qxF "$ASSET_NAME"; then
+            echo "manifest $ASSET_NAME already published — keeping the immutable original"
+          else
+            gh release upload "$RELEASE_TAG" "$ASSET"
+          fi
 
       - name: Upload to S3 with Git Tag
         if: startsWith(github.ref, 'refs/tags/')
--- a/.github/workflows/service-release-simple-staking.yml
+++ b/.github/workflows/service-release-simple-staking.yml
@@ -64,6 +64,16 @@ jobs:
     # prod failure must fail the run — a green run with a failed prod deploy or
     # missing integrity manifest is a silent gap (#253 review).
     continue-on-error: ${{ matrix.prod == true && github.ref != 'refs/heads/main' && github.ref != 'refs/heads/release/v1.x' && !startsWith(github.ref, 'refs/tags/') }}
+    # Serialize same-(env, sha) runs. Without this, git push --follow-tags or a
+    # redeploy dispatch can overlap two runs on one sha: S3 is last-write-wins
+    # while the manifest slots are first-create-wins, so the highest slot can
+    # belong to a DIFFERENT build than the bytes in S3 — a false tamper page on
+    # a clean deploy. cancel-in-progress stays false so the second run publishes
+    # its S3 write and its slot strictly after the first (a cancelled queued run
+    # writes neither store, which keeps them consistent).
+    concurrency:
+      group: ${{ github.workflow }}-${{ matrix.environment }}-${{ github.sha }}
+      cancel-in-progress: false
     environment: ${{ matrix.environment }}
     steps:
       - name: Checkout repository
@@ -162,51 +172,13 @@ jobs:
 
       # Last step: publishing the manifest must not block the S3 uploads above.
       - name: Publish checksum manifest as an immutable per-sha release
-        # Only real deploy refs — workflow_dispatch can run from any branch, and
-        # the public store must hold manifests only for builds that shipped.
-        if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/release/v1.x' || startsWith(github.ref, 'refs/tags/')
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: |
-          set -euo pipefail
-          # Per-deploy sha256 manifests for the frontend integrity checker
-          # (devops-backlog#253). One release per (service, env, sha) with both
-          # files attached at create time: this repo has Immutable Releases
-          # enabled, so assets can't be added after publish — attaching at create
-          # is the immutable-safe path, and it makes write-once a platform
-          # guarantee (published releases can't be modified) and the two-file
-          # publish atomic. No asset-cap pruning needed (2 assets per release).
-          #
-          # Same-sha rebuilds are sanctioned (tag cut on an already-built commit,
-          # redeploy dispatch, job re-run) and are NOT byte-reproducible (run_id
-          # in the bundle), so each rebuild takes the next free -rN slot. The
-          # checker probes slots upward and verifies against the highest one.
-          PREFIX="simple-staking-${{ matrix.environment }}-${{ github.sha }}"
-          FILES=("/tmp/${PREFIX}-files-checksum.txt" "/tmp/${PREFIX}-global-checksum.txt")
-          for i in 1 2 3 4 5 6 7 8 9 10; do
-            TAG="fm-${PREFIX}"; [ "$i" -gt 1 ] && TAG="${TAG}-r${i}"
-            URL="https://github.com/${GITHUB_REPOSITORY}/releases/download/${TAG}/${PREFIX}-files-checksum.txt"
-            if curl -sfIL -o /dev/null "$URL"; then
-              echo "slot ${TAG} already published — trying next"
-              continue
-            fi
-            if ERR=$(gh release create "$TAG" "${FILES[@]}" \
-                 --title "Frontend manifest ${PREFIX}" \
-                 --notes "sha256 manifest for the frontend integrity checker (devops-backlog#253). Do not delete." \
-                 --latest=false 2>&1); then
-              curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
-              echo "published ${TAG}"
-              exit 0
-            fi
-            # Advance to the next slot ONLY on a create race ("already exists").
-            # Any other failure must fail the step: skipping it would leave a
-            # numbering gap, and the checker stops at the first missing slot —
-            # the next run's higher slot would never be found (silent gap).
-            if ! grep -qi "already exists" <<<"$ERR"; then
-              echo "::error::gh release create ${TAG} failed: ${ERR}"
-              exit 1
-            fi
-            echo "slot ${TAG} already exists (lost create race) — trying next slot"
-          done
-          echo "::error::could not publish manifest for ${PREFIX} after 10 slots"
-          exit 1
+        # Only real deploy refs (a dispatch from an arbitrary branch must not
+        # notarize builds), and only envs the integrity checker monitors — the
+        # other envs would mint releases+tags nobody ever reads (~7/push
+        # otherwise). Extend the env list together with the checker's
+        # targets.json.
+        if: (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/release/v1.x' || startsWith(github.ref, 'refs/tags/')) && contains(fromJSON('["mainnet","testnet"]'), matrix.environment)
+        uses: ./.github/actions/publish-integrity-manifest
+        with:
+          service: simple-staking
+          environment: ${{ matrix.environment }}
--- a/.github/workflows/service-release-vault.yml
+++ b/.github/workflows/service-release-vault.yml
@@ -107,10 +107,11 @@ jobs:
       matrix: ${{ fromJSON(needs.setup.outputs.matrix) }}
 
     # Prod OIDC fails on feature branches; tolerate it ONLY in the multi-env run
-    # so it can't block devnet. When prod (vault-testnet) is the single requested
-    # environment, fail loud — otherwise a scoped prod dispatch reports success
-    # while publishing nothing.
-    continue-on-error: ${{ matrix.prod == true && needs.setup.outputs.multi == 'true' }}
+    # so it can't block devnet — and only OFF real deploy refs. On main/release/
+    # tags a prod failure must fail the run: a green run with a failed prod
+    # deploy or missing integrity manifest is a silent gap (#253 review). A
+    # scoped single-prod dispatch also fails loud.
+    continue-on-error: ${{ matrix.prod == true && needs.setup.outputs.multi == 'true' && github.ref != 'refs/heads/main' && github.ref != 'refs/heads/release/v1.x' && !startsWith(github.ref, 'refs/tags/') }}
     environment: ${{ matrix.environment }}
     steps:
       - name: Checkout repository
@@ -228,10 +229,14 @@ jobs:
 
       - name: Upload to S3 with Git Tag
         if: startsWith(github.ref, 'refs/tags/')
+        env:
+          # Bound via env, not interpolated into run: — git refnames accept
+          # shell metacharacters and this job holds AWS OIDC + contents:write.
+          GIT_TAG: ${{ github.ref_name }}
         run: |
-          aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}.tar.gz s3://${{ vars.S3_PUBLIC_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.ref_name }}.tar.gz
-          aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.ref_name }}-files-checksum.txt
-          aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.ref_name }}-global-checksum.txt
+          aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}.tar.gz "s3://${{ vars.S3_PUBLIC_BUCKET_NEW }}/vault/${{ matrix.environment }}/${GIT_TAG}.tar.gz"
+          aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt "s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${GIT_TAG}-files-checksum.txt"
+          aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt "s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${GIT_TAG}-global-checksum.txt"
 
       # Last step: publishing the manifest must not block the S3 uploads above.
       - name: Publish checksum manifest as an immutable per-sha release
--- a/.github/workflows/service-release-vault.yml
+++ b/.github/workflows/service-release-vault.yml
@@ -98,6 +98,9 @@ jobs:
 
   s3_publish:
     needs: [lint_test, setup]
+    permissions:
+      id-token: write # OIDC assume-role for AWS
+      contents: write # publish checksum manifest as a release asset (#253)
     runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
@@ -218,6 +221,22 @@ jobs:
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}.tar.gz s3://${{ vars.S3_PUBLIC_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.sha }}.tar.gz
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.sha }}-files-checksum.txt
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.sha }}-global-checksum.txt
+
+      - name: Publish checksum manifest as immutable release asset
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Append-only per-deploy sha256 manifest for the frontend integrity
+          # checker (devops-backlog#253). Public + immutable per-sha => the
+          # checker reads it with no credentials, from a different trust domain
+          # than the AWS serving path. One pinned release acts as the store.
+          RELEASE_TAG=frontend-manifests
+          ASSET="/tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt"
+          gh release create "$RELEASE_TAG" \
+            --title "Frontend integrity manifests" \
+            --notes "Append-only sha256 manifests consumed by the frontend integrity checker (devops-backlog#253). Do not delete." \
+            --latest=false 2>/dev/null || true
+          gh release upload "$RELEASE_TAG" "$ASSET" --clobber
 
       - name: Upload to S3 with Git Tag
         if: startsWith(github.ref, 'refs/tags/')
--- a/.github/workflows/service-release-vault.yml
+++ b/.github/workflows/service-release-vault.yml
@@ -115,6 +115,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2 https://github.com/actions/checkout/releases/tag/v6.0.2
+        with:
+          # Job holds contents:write; don't leave the token in .git/config while
+          # pnpm install / build run untrusted deps.
+          persist-credentials: false
 
       - name: Set up Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0 https://github.com/actions/setup-node/releases/tag/v4.4.0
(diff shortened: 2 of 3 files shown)


.github/workflows/service-release-simple-staking.yml
```diff
@@ -139,6 +142,22 @@ jobs:
...
+          # checker reads it with no credentials, from a different trust domain
+          # than the AWS serving path. One pinned release acts as the store.
+          RELEASE_TAG=frontend-manifests
+          ASSET="/tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt"
+          gh release create "$RELEASE_TAG" \
+            --title "Frontend integrity manifests" \
+            --notes "Append-only sha256 manifests consumed by the frontend integrity checker (devops-backlog#253). Do not delete." \
+            --latest=false 2>/dev/null || true
+          gh release upload "$RELEASE_TAG" "$ASSET" --clobber
+
```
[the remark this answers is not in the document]
toroviej0x0: Fixed in c92a6eb8. Dropped `--clobber` — now write-once: the step checks whether the per-sha manifest asset already exists and skips the upload if so, keeping the original. A rerun/dispatch can no longer swap the manifest. Applied to both service-release workflows. (A stale manifest from a same-sha rebuild would fail safe — mismatch alert — rather than silently replacing the source of truth.)

jonybur (dismissed):
The head addresses the overwrite concern from the earlier round: the write-once guard skips the upload when the per-sha asset already exists instead of clobbering it, so a rerun cannot swap the checker's source of truth. Job-level permissions are correct - the block redeclares id-token: write alongside the new contents: write, so the OIDC assume-role path keeps working and lint_test stays read-only. The release create is idempotent across the parallel env matrix, asset names are unique per service/env/sha so the matrix jobs cannot collide, and the existing S3 path is untouched. One thing to keep an eye on operationally rather than in this diff: the store grows one asset per deploy per env, and the existence check reads the release's asset list, so pruning (as the description already anticipates) is worth scheduling before the list gets huge.

.github/workflows/service-release-vault.yml
```diff
@@ -219,6 +222,31 @@ jobs:
...
+            --latest=false 2>/dev/null || true
+          # Write-once: never overwrite an already-published manifest for this
+          # sha. Builds aren't byte-reproducible (run_id embedded), so a rerun
+          # with --clobber could swap the checker's source of truth. Keeping the
+          # original is fail-safe — a stale manifest alerts, it can't hide.
+          ASSET_NAME="$(basename "$ASSET")"
+          if gh release view "$RELEASE_TAG" --json assets --jq '.assets[].name' | grep -qxF "$ASSET_NAME"; then
+            echo "manifest $ASSET_NAME already published — keeping the immutable original"
+          else
+            gh release upload "$RELEASE_TAG" "$ASSET"
```
jrwbabylonlab: One release holds both services x all envs forever. Measured from `gh run list`: ~100 runs/month each, 3 + 4 envs = ~700 assets/month. GitHub caps a release at 1000 assets (422 `file_count limited to 1000 assets per release`), so every upload starts failing in ~6 weeks. Needs sharding plus an asset-count alarm.

.github/workflows/service-release-vault.yml
```diff
@@ -98,6 +98,9 @@ jobs:
 
   s3_publish:
     needs: [lint_test, setup]
+    permissions:
```
jrwbabylonlab, line 101: `contents: write` now covers the job that runs `pnpm install` and `pnpm run build`, and `actions/checkout@v6.0.2` defaults to `persist-credentials: true` so the token sits in `.git/config` while build plugins execute. Every transitive build dep goes from read-repo to push-to-main and rewrite-any-release. Publish from a separate job with no install and `persist-credentials: false`, passing the manifest via upload-artifact. Same on `service-release-simple-staking.yml:43`.

.github/workflows/service-release-vault.yml
```diff
@@ -219,6 +222,31 @@ jobs:
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.sha }}-files-checksum.txt
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.sha }}-global-checksum.txt
```
jrwbabylonlab, line 229: This job is `continue-on-error` on prod (`:113`, staking `:63`). If the upload fails, the run stays green, S3 has already published, and the live bundle has no manifest. A detection control should not fail open.

.github/workflows/service-release-vault.yml
```diff
@@ -219,6 +222,31 @@ jobs:
...
 
+      - name: Publish checksum manifest as immutable release asset
+        env:
```
jrwbabylonlab: Placed between the sha upload and the tag upload. Default `bash -e {0}` means a failed `gh release upload` ends the job, so a tag push during a Releases API incident ships `<sha>.tar.gz` but never `<tag>.tar.gz`. Move it to the last step.

.github/workflows/service-release-vault.yml
```diff
@@ -219,6 +222,31 @@ jobs:
...
+          # checker (devops-backlog#253). Public + immutable per-sha => the
+          # checker reads it with no credentials, from a different trust domain
+          # than the AWS serving path. One pinned release acts as the store.
+          RELEASE_TAG=frontend-manifests
+          ASSET="/tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt"
+          gh release create "$RELEASE_TAG" \
+            --title "Frontend integrity manifests" \
+            --notes "Append-only sha256 manifests consumed by the frontend integrity checker (devops-backlog#253). Do not delete." \
+            --latest=false 2>/dev/null || true
+          # Write-once: never overwrite an already-published manifest for this
```
jrwbabylonlab: Release side is first-write-wins, S3 side is last-write-wins (fixed `<sha>.tar.gz` key, no `If-None-Match`). A `workflow_dispatch` redeploy gets a fresh `run_id`, which reaches the bundle via `NEXT_PUBLIC_DIST_ID` -> `sentry.client.config.ts:75`: different bytes, pinned manifest, permanent false alarm, and the only fix is the deletion the notes forbid. Separately the comment names the wrong case: `GITHUB_RUN_ID` does not change on re-run. Key the manifest by `run_id`.

.github/workflows/service-release-vault.yml
```diff
@@ -219,6 +222,31 @@ jobs:
...
+          # sha. Builds aren't byte-reproducible (run_id embedded), so a rerun
+          # with --clobber could swap the checker's source of truth. Keeping the
+          # original is fail-safe — a stale manifest alerts, it can't hide.
+          ASSET_NAME="$(basename "$ASSET")"
+          if gh release view "$RELEASE_TAG" --json assets --jq '.assets[].name' | grep -qxF "$ASSET_NAME"; then
```
jrwbabylonlab: First run after merge: 7 matrix jobs across both workflows hit this concurrently. `|| true` swallows the losers 422, but nothing waits before `gh release view`, so a job reading before the create commits gets not-found and then fails the upload. Up to 6 of 7 fail.

.github/workflows/service-release-vault.yml
```diff
@@ -219,6 +222,31 @@ jobs:
...
+          # Append-only per-deploy sha256 manifest for the frontend integrity
+          # checker (devops-backlog#253). Public + immutable per-sha => the
+          # checker reads it with no credentials, from a different trust domain
+          # than the AWS serving path. One pinned release acts as the store.
+          RELEASE_TAG=frontend-manifests
+          ASSET="/tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt"
+          gh release create "$RELEASE_TAG" \
+            --title "Frontend integrity manifests" \
+            --notes "Append-only sha256 manifests consumed by the frontend integrity checker (devops-backlog#253). Do not delete." \
+            --latest=false 2>/dev/null || true
```
jrwbabylonlab: `2>/dev/null` plus `|| true` discards the only diagnostic if `contents: write` is ever revoked by an org policy or ruleset; the job then dies several lines later with a worse message.

mpastecki (dismissed):

.github/workflows/service-release-vault.yml
```diff
@@ -219,6 +222,31 @@ jobs:
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.sha }}-files-checksum.txt
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.sha }}-global-checksum.txt
 
+      - name: Publish checksum manifest as immutable release asset
```
gbarkhatov: **`workflow_dispatch` from any branch notarizes arbitrary builds.** This step publishes unconditionally, but `workflow_dispatch` runs from any branch. Anyone with write access can push a branch, dispatch, and land a manifest in the trusted public store for a build that never went through main. The store's implied contract is "manifests for things that were actually deployed" — one line restores it:

```yaml
if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/release/v1.x' || startsWith(github.ref, 'refs/tags/')
```

Same on `service-release-simple-staking.yml:145`.

Nit while here: the step name says "immutable", but the behaviour it implements is write-once — see the summary comment on why those differ.

.github/workflows/service-release-vault.yml
gbarkhatov: **The guard fails open, and it verifies the producer's view rather than the consumer's.** Two compounding issues:

1. No `set -euo pipefail` here (the `setup` job at `:71` does use it), so a failing `gh release view` yields empty stdout, `grep` returns false, and control falls through to the upload.
2. This relies on `--json assets` returning the complete list. As the asset count grows toward the cap raised above, any truncation has exactly the same fail-open effect.

Checking the URL the checker actually reads is simpler and self-validating — it also covers the draft case below:

```bash
URL="https://github.com/${GITHUB_REPOSITORY}/releases/download/$RELEASE_TAG/$ASSET_NAME"
if curl -sfIL -o /dev/null "$URL"; then
  echo "already published — keeping the original"
else
  gh release upload "$RELEASE_TAG" "$ASSET"
  curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
fi
```

Related: nothing asserts the release isn't a draft. `--title`/`--notes` apply only at create time, so if `frontend-manifests` already exists or is ever flipped to draft, uploads succeed while the public download URL 404s — step goes green, checker gets nothing.

.github/workflows/service-release-vault.yml
```diff
@@ -219,6 +222,31 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Append-only per-deploy sha256 manifest for the frontend integrity
+          # checker (devops-backlog#253). Public + immutable per-sha => the
+          # checker reads it with no credentials, from a different trust domain
+          # than the AWS serving path. One pinned release acts as the store.
+          RELEASE_TAG=frontend-manifests
+          ASSET="/tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt"
```
gbarkhatov: **Only `files-checksum.txt` is published — `global-checksum.txt` isn't.** The S3 path uploads both (`:222-223`). The global checksum is the single-line root hash over the manifest: the cheapest thing for a checker to pin and cache, and the natural place to hang a signature if attestations get added. As it stands every check has to pull and trust the whole blob. One extra `gh release upload`.

.github/workflows/service-release-simple-staking.yml
```diff
@@ -139,6 +142,31 @@ jobs:
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.sha }}-files-checksum.txt
           aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${{ github.sha }}-global-checksum.txt
 
+      - name: Publish checksum manifest as immutable release asset
```
gbarkhatov: **25 identical lines duplicated across the two workflows**, and there's no `.github/actions/` directory in the repo yet. This is security-relevant logic — the write-once guard specifically — that must not drift between services, and the next service will copy it a third time. A composite action taking `service` / `environment` / `sha` inputs would fix that, and would also let the separate-publish-job restructure suggested above happen once instead of twice.

gbarkhatov (commented):
A few things beyond what's already been raised (the `--clobber` mutability, asset cap, `contents: write` on the build job, `continue-on-error`, step placement, run_id keying, create race, and swallowed diagnostics are all covered elsewhere). Inline comments carry the line-anchored ones; these three are architectural.

**"Immutable" overstates what a release asset gives you.** Release assets are deletable and replaceable by any token holding `contents: write` — including, per the comment above, the one this job now holds while untrusted build deps execute. The write-once behaviour is a convention this step follows, not something GitHub enforces; a later workflow or an admin can replace the file with nothing to detect it. If the goal is genuinely "an attacker must break two systems", the primitive that delivers it is `actions/attest-build-provenance` — Sigstore-signed, recorded in a public transparency log, verifiable offline with `gh attestation verify`. This repo already runs provenance for npm publishing (`package-release.yml:31`), so it's a known path. Worth deciding now rather than retrofitting, since adding attestations later means changing the consumer too.

**The `frontend-manifests` tag is unprotected.** Current repo rulesets are `OnlyAdminsDeleteRepos` (repository target), `enforce-committers-whitelist`, `enforce-pinned-actions-hashes`, and `global-branch-policy` — all branch or repository targets, none targets tags. So `git push --delete origin frontend-manifests` from any write-access account 404s every `releases/download/frontend-manifests/*` URL at once. "Do not delete." in the release notes is a comment, not a control; a tag ruleset restricting deletion and update on that tag is.

**Cross-environment substitution — the checker contract needs writing down.** `mainnet`, `testnet` and the devnets all share one public store, and nothing in this repo records that dapp-deployment#321 depends on this exact tag and filename scheme. Two consequences: (a) if the checker derives the manifest name from anything the page reports (`NEXT_PUBLIC_COMMIT_HASH` is baked into the bundle at `service-release-vault.yml:162`), a CDN attacker can serve a staging bundle to mainnet users and point at the staging manifest — every hash matches, no alert. The `<service>-<env>-` prefix has to come from the checker's own config, never from the response. (b) Renaming the checksum file, changing the tag, or adopting the `run_id` keying suggested above silently breaks the checker with no failing test here — those need to land in lockstep.

On the positive side: the job-scoped permissions correctly redeclare `id-token: write` so the OIDC path keeps working and `lint_test` stays read-only, asset names are unique per service/env/sha so matrix jobs can't collide, `--latest=false` keeps this out of the "Latest" badge among the many package tags, and the S3 path is untouched so the change is genuinely additive.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed in 54c0bd58 — the manifest publish is now the last step, after both S3 uploads, so a Releases API hiccup can't block `<tag>.tar.gz`.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed in 54c0bd58. No longer depends on `gh release view` timing: each asset is checked with `curl -sfIL` against the public download URL and uploaded only if missing. The create is idempotent (`|| true`); losers of the create race still upload fine because the URL check, not the view, gates it.

.github/workflows/service-release-vault.yml
toroviej0x0: Improved in 54c0bd58. `set -euo pipefail` + the upload is verified retrievable afterward (`curl ... || exit 1`), so a real failure surfaces loudly. Only the `gh release create` keeps `|| true` (expected 'already exists' across the matrix); the load-bearing upload/verify no longer swallow errors.

.github/workflows/service-release-vault.yml
toroviej0x0: Addressed in 54c0bd58 with `persist-credentials: false` on the publish job's checkout, so the token isn't left in `.git/config` while pnpm install/build run. Full separate-publish-job (no install, artifact hand-off) is a reasonable next step — noted as a follow-up; the concrete .git/config vector you named is closed now.

.github/workflows/service-release-vault.yml
toroviej0x0: Partly. On vault the job already fails loud for a single prod env (`continue-on-error: matrix.prod && needs.setup.outputs.multi == 'true'`), and the step itself is now `set -euo pipefail` + verify-after-upload so failures show in the run. The job-level `continue-on-error` stays because prod OIDC legitimately fails on feature branches. A dedicated non-CoE publish job would fully close the fail-open on the multi-env path — tracking as follow-up.

.github/workflows/service-release-vault.yml
toroviej0x0: Can't key by run_id without breaking the consumer: the checker only learns the commit sha (from the page's `<meta version>`), never the run_id, so a run_id-keyed filename isn't derivable at check time. Write-once + fail-safe is the deliberate trade: a same-sha rebuild leaves the original manifest, so a divergent rebuild alerts (mismatch) rather than silently passing. Also fair point that GITHUB_RUN_ID is stable on re-run — comment reworded. Making the store immutable at the platform level (attestation/tag ruleset) is the real fix; see the summary reply.

.github/workflows/service-release-vault.yml
toroviej0x0: Real issue — acknowledged. ~700 assets/month vs the 1000/release cap means this needs sharding (e.g. per-service or per-month release tags) plus an asset-count alarm before ~6 weeks. That changes the checker's URL derivation too, so it must land in lockstep. Tracking as a fast-follow, not silently — thanks for the measured numbers.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed in 54c0bd58 — added `if: github.ref == 'refs/heads/main' || 'refs/heads/release/v1.x' || startsWith(github.ref, 'refs/tags/')` on both workflows, so a workflow_dispatch from an arbitrary branch no longer lands a manifest in the public store. Renamed the step to 'write-once' too.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed in 54c0bd58. Now `set -euo pipefail`, and the guard checks the public download URL the checker reads (`curl -sfIL`) instead of `--json assets` — so it's fail-closed and covers truncation and the draft-release case. After an upload it re-checks the URL and `exit 1` if not retrievable.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed in 54c0bd58 — the step now loops over both `files-checksum.txt` and `global-checksum.txt`, so the single root hash is published too (cheap to pin, and the natural anchor if we add signatures).

.github/workflows/service-release-simple-staking.yml
toroviej0x0: Agreed the duplicated write-once logic shouldn't drift. Left inline for this PR to keep the diff reviewable, but a composite action (`service`/`environment`/`sha` inputs) is the right home — tracking as a follow-up, to land together with the separate-publish-job restructure so it's done once. No `.github/actions/` exists yet; that PR creates it.

toroviej0x0:
@gbarkhatov thanks — on the three architectural points:

1. **Attestations vs "immutable".** Agreed the release-asset write-once is a convention, not a platform guarantee. `actions/attest-build-provenance` (already used at `package-release.yml:31`) is the real primitive and I want it — but it changes the consumer too (the checker must `gh attestation verify` / verify the Sigstore bundle). Capturing as a follow-up decision so we do producer + consumer together, rather than blocking this PR. Renamed the step from "immutable" to "write-once" to stop overstating.

2. **Unprotected `frontend-manifests` tag.** Correct — "Do not delete." is a comment, not a control, and none of the current rulesets target tags. A tag ruleset restricting delete/update on that tag is the fix. It's a repo-settings change (outside this PR's files); I'll raise it with the follow-up and flag it to the repo admins.

3. **Cross-env substitution.** Point (a) is already satisfied by the consumer design: the checker derives the `<service>-<env>-` prefix from its **own config** (`targets.json`), and takes only the sha from the page. So a mainnet check always looks up `simple-staking-mainnet-<sha>`; a staging bundle served to mainnet yields a staging sha whose `-mainnet-` manifest doesn't exist → 404 → check error (alert), not a silent pass. Point (b) stands: the tag + `<service>-<env>-<sha>-files-checksum.txt` scheme is a contract with dapp-deployment#321 and nothing here fails if it's renamed — I'll add a note to both PRs so any change lands in lockstep.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed in f3a58614. After each publish the step prunes the release to the newest 500 manifests (`sort_by(.createdAt) | reverse | .[500:]` deleted, best-effort so a matrix prune race can't fail publish). Only the currently-deployed sha per env is ever looked up, so older manifests are dead weight — this keeps the release far under the 1000 cap indefinitely, no sharding + no checker change needed. A count alarm is still a nice backstop; noting as optional follow-up.

jonybur (changes requested):
The trust-domain idea is sound and worth landing, and a lot of the execution is careful (scoped `contents: write`, `persist-credentials: false`, the write-once probe, race-tolerant create). But checking the mechanism against this repo's live settings and the companion checker branch turned up one fatal incompatibility and four design defects, all in the same theme: the write-once model doesn't match how this repo actually builds and deploys. Everything marked **verified** below was reproduced against the live GitHub API / the checker branch, not reasoned from memory.

## Blocking

**1. This repo has Immutable Releases enabled - the mechanism cannot work at all (verified).**
Every release published here since 2026-01-29 has `immutable: true` (294 of them, including yesterday's). Under immutable releases, assets cannot be added after publish - so the very first `gh release upload` to the just-created, non-draft `frontend-manifests` release 422s, `set -euo pipefail` fails the step, and every deploy job goes red while the checker has nothing to read. I suspect this was tested on a fork without the setting.

Fix that also improves the design: publish **one release per (service, env, sha)** with the file(s) attached at create time (`gh release create <tag> <files> --latest=false` - gh's internal draft-upload-publish sequence is the immutable-safe path). Write-once then becomes a platform guarantee instead of a curl-probe convention, and findings 3-5 mostly dissolve.

**2. The global newest-500 prune evicts manifests of still-live prod deployments in ~10 days (verified).**
Both services share the release; every main push fires both workflows and adds 14 assets (staking 4 envs + vault 3 envs, x2 files). At the measured cadence (~3.5 main pushes/day), 500 assets is ~36 pushes, or about 10 days of history. The monitored prod targets serve tag-pinned shas routinely older than that - so the checker's most important targets would page "missing manifest" precisely because the prune deleted a still-live build's manifest. The comment's justification ("only the currently-deployed sha per env is ever looked up") assumes per-env recency the global sort doesn't provide. If the single-release design survives: prune per `<service>-<env>` prefix (newest ~30 pairs per env = 420 assets, still under the cap), or exempt currently-deployed shas.

**3. Vault asset names don't match what the checker fetches - every vault check 404s from day one (verified cross-repo).**
Vault's matrix env values are already service-prefixed (`vault-testnet`), so the uploaded asset is `vault-vault-testnet-<sha>-files-checksum.txt`. The checker (both the Go Lambda and the Cloudflare Worker on the companion branch) builds `<service>-<env>-<sha>` from `targets.json` `{service: "vault", env: "testnet"}` and fetches `vault-testnet-<sha>-...` - a name this PR never publishes. Staking lines up; vault pages permanently. Normalize on one side and pin the exact per-service asset-name contract in both PRs (ideally asserted by the checker's self-test).

**4. Write-once collides with non-reproducible same-sha rebuilds - legitimate redeploys become permanent tamper alerts (verified).**
Builds embed `github.run_id` (`NEXT_PUBLIC_DIST_ID`), so rebuilding a sha produces different bytes. The sha-keyed S3 uploads are unconditional overwrites, while the manifest keeps the first run's hashes. Three sanctioned flows rebuild an already-manifested sha: cutting a tag on a commit already built from main (the standard prod release flow), the unified-redeploy `workflow_dispatch`, and a plain job re-run. After any of them, served content permanently disagrees with the frozen manifest - a false "UI poisoning" page that write-once forbids repairing. Bonus race: `git push --follow-tags` runs the main and tag builds concurrently; both pass the absent-check and the loser's duplicate upload fails the deploy job (no `--clobber`). Per-sha assets need a run-identity component (with the checker accepting any matching manifest for the deployed sha), or one-build-per-sha enforced on the S3 side too.

**5. The two-asset pair is published non-atomically but is cryptographically linked (verified).**
`global-checksum.txt` is sha256 of `files-checksum.txt`, yet the loop uploads them sequentially with real failure windows between (the post-upload verification itself can `exit 1`). A partial publish means the rerun pairs its global with the first run's files - inconsistent forever under write-once. Simplest fix: publish a single asset (the global hash is derivable from it), which also halves the prune pressure in (2).

## Worth addressing

- **Token capture surface:** `contents: write` shares the job with `pnpm install`/build. `persist-credentials: false` is good hygiene but doesn't stop a malicious postinstall from poisoning `$GITHUB_ENV`/`PATH` to capture `GH_TOKEN` when the last step runs. A separate publish job consuming the manifest as an artifact closes it - though a build-time compromise can forge manifest *content* regardless, so it's worth stating that the scheme protects against post-build CDN/edge tampering, not build compromise.
- **`continue-on-error` on prod legs** swallows manifest-publish failures on exactly the monitored environments - a silent S3-vs-release gap on prod deploys.
- **Ref-gate mismatch:** `workflow_dispatch` from a non-deploy ref still uploads to S3 (devnet OIDC succeeds on feature branches) but skips the manifest - deploys the checker can't verify.
- **HEAD probe conflates transient failure with 404:** a network blip on an already-published asset routes into the upload branch and fails the deploy job.
- **No liveness binding:** the store is content-addressed only, so a rollback to any previously-manifested build verifies clean. Fine - but worth documenting as an accepted limitation.

## Docs/nits

- PR body drift: "prune old ones later if desired" (the diff prunes now, KEEP=500); "the checker reads files-checksum" (both files are published); release notes say "Do not delete" while the same step deletes assets.
- The duplicated 45-line block follows repo convention, but unlike the duplicated S3 steps, both copies mutate *shared* state (one release tag, one prune policy) - if the single-release design survives, extract it; divergent edits here corrupt a shared store.
- The tag gate `startsWith(github.ref, 'refs/tags/')` is broader than each workflow's own `on.push.tags` namespace; harmless today, but mirroring the per-file tag pattern would be tighter.

## What holds up (verified sound)

- The HEAD probe works through the release-asset redirect chain (200/exit 0 existing; 404/non-zero missing).
- The job-level `permissions` block correctly re-grants `id-token: write` (OIDC intact) and starves no other job; `persist-credentials: false` breaks nothing downstream.
- The shell is correct under `set -euo pipefail` (create race covered, pipeline `|| true` placement right, jq slice semantics fine); asset names are collision-free across the matrix and both services; no script-injection risk in the interpolations.

## Bottom line

As written, the first deploy fails outright on this repo's Immutable Releases setting; with that fixed, vault monitoring 404s from day one, prod manifests get pruned while live, and every routine same-sha redeploy becomes a permanent false page. Restructuring to one immutable release per (service, env, sha) with a single asset resolves (1), (4) (with a run-id component), (5), and most of (3) in one move, and turns write-once into a platform guarantee. I'd ask for that rework plus the pinned name contract before merge.

.github/workflows/service-release-vault.yml
```diff
@@ -225,3 +232,48 @@ jobs:
...
+            if curl -sfIL -o /dev/null "$URL"; then
+              echo "manifest $NAME already published — keeping the original"
+            else
+              gh release upload "$RELEASE_TAG" "$ASSET"
+              curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
+            fi
+          done
+          # Keep the release well under GitHub's 1000-asset/release cap. Only the
+          # currently-deployed sha per env is ever looked up, so older manifests
+          # are dead weight — keep the most recent 500, drop the rest. Best-effort
```
jrwbabylonlab: Neither workflow has a `paths:` filter, so every push to main fires both: 3 envs x 2 files here plus 4 x 2 in staking = 14 assets per push. At the observed run rate `KEEP=500` is roughly ten days of history, while deployed builds live far longer. Any env serving a build older than that has its manifest deleted and the checker 404s on exactly the builds it exists to verify. Retention should be bounded by max deploy age, not count.

.github/workflows/service-release-vault.yml
```diff
@@ -225,3 +232,48 @@ jobs:
...
+              --latest=false || true
+          for ASSET in \
+            "/tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt" \
+            "/tmp/vault-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt"; do
+            NAME="$(basename "$ASSET")"
+            URL="https://github.com/${GITHUB_REPOSITORY}/releases/download/${RELEASE_TAG}/${NAME}"
+            # Write-once, checked against the URL the checker actually reads (not
+            # the producer's asset list): survives the create race, asset-list
+            # truncation and a draft release. Builds aren't byte-reproducible
+            # (run_id embedded), so a rerun must never swap a published manifest.
```
jrwbabylonlab: The existence guard keys on the asset URL 404ing, so once the prune evicts a name a later run for that sha re-uploads under it. Combined with `run_id` reaching the bundle via `NEXT_PUBLIC_DIST_ID`, a rerun of an evicted sha can republish different bytes as the trusted record. The release body still says "Append-only. Do not delete."

.github/workflows/service-release-vault.yml
```diff
@@ -225,3 +232,48 @@ jobs:
...
+            fi
+          done
+          # Keep the release well under GitHub's 1000-asset/release cap. Only the
+          # currently-deployed sha per env is ever looked up, so older manifests
+          # are dead weight — keep the most recent 500, drop the rest. Best-effort
+          # (|| true) so a prune race across the env matrix can't fail the publish.
+          KEEP=500
+          gh release view "$RELEASE_TAG" --json assets \
+            --jq ".assets | sort_by(.createdAt) | reverse | .[${KEEP}:] | .[].name" \
+          | while IFS= read -r OLD; do
```
jrwbabylonlab: The prune reads `gh release view --json assets` — the mechanism the guard above was deliberately moved off for being truncation-prone. If that list ever returns under 500 entries the prune silently no-ops forever, with `2>/dev/null` and two `|| true` suppressing every signal. No asset-count alarm.

.github/workflows/service-release-vault.yml
jrwbabylonlab, line 101: `contents: write` is repo-wide, so this token can delete assets and releases across the whole repo including the published core-ui / ts-sdk packages — and the step now exercises deletion, not just append. `persist-credentials: false` closes the `.git/config` path, but the token is still minted at write scope for a job that runs `pnpm install` and `pnpm run build`. A separate publish job with no install is the safer shape.

.github/workflows/service-release-vault.yml
```diff
@@ -225,3 +232,48 @@ jobs:
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}.tar.gz s3://${{ vars.S3_PUBLIC_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.ref_name }}.tar.gz
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.ref_name }}-files-checksum.txt
           aws s3 cp /tmp/vault-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/vault/${{ matrix.environment }}/${{ github.ref_name }}-global-checksum.txt
+
```
jrwbabylonlab, line 240: `${{ github.ref_name }}` interpolated into `run:`. Git accepts shell metacharacters in refnames — `git check-ref-format 'refs/tags/vault/v1$(id)'` exits 0, likewise backticks, `;` and `&`. A tag matching `vault/*` executes in a job that now holds `contents: write` alongside prod AWS OIDC. Pass it via `env:` and quote it.

.github/workflows/service-release-simple-staking.yml
```diff
@@ -62,6 +65,10 @@ jobs:
     steps:
       - name: Checkout repository
```
jrwbabylonlab, line 69: `continue-on-error` covers mainnet and testnet unconditionally here, so a failed manifest publish leaves the run green with S3 already deployed and the live bundle unattested. A detection control should not fail open on prod.

gbarkhatov (commented):
Short version on the standing question: rounds 1–2 were addressed, and well — the write-once probe, the step reordering, the ref gate and `set -euo pipefail` are all real fixes. The third round hasn't been picked up: head is still `f3a5861` (Jul 31 13:45), which predates @jonybur's `CHANGES_REQUESTED` by two hours and @jrwbabylonlab's six comments by three days.

I re-verified the two blocking claims that had no inline anchor (they were in a review body, which is easy to lose) against the live API and the `feat/frontend-integrity-checker` branch. Both hold, and the naming one is narrower and more fixable than it first reads. Details inline.

Not re-raising the six open threads — the prune, token scope and `github.ref_name` findings stand as written and I have nothing to add to them.

On direction: the one-release-per-`(service, env, sha)` restructure is the only shape that survives Immutable Releases, and it happens to delete the prune along with all four findings the prune introduced. That seems worth doing before more rounds of patching this one.

.github/workflows/service-release-vault.yml
```diff
@@ -225,3 +232,48 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          # Append-only per-deploy sha256 manifests for the frontend integrity
+          # checker (devops-backlog#253). Public, so the checker reads them with
+          # no credentials, from a different trust domain than the AWS serving path.
+          RELEASE_TAG=frontend-manifests
+          gh release view "$RELEASE_TAG" >/dev/null 2>&1 || \
+            gh release create "$RELEASE_TAG" \
```
gbarkhatov: **Immutable Releases is on for this repo — the first `gh release upload` 422s (verified).**

Anchoring @jonybur's blocking finding to the line, since it landed in a review body and hasn't been picked up. The evidence is a little stronger than stated:

- 738 releases total. The flag flips between `simple-staking/v1.5.20` (2026-01-20, `immutable: false`) and `@babylonlabs-io/core-ui/1.77.0` (2026-01-29, `immutable: true`). **All 309 releases published since are `immutable: true`**, including this week's.
- **0 of those 738 releases carry a single asset.** No release in this repo has ever had one attached, so this publish path has never run here — consistent with it having been exercised on a fork.

`gh release create` without `--draft` publishes immediately, which seals the release. The `gh release upload` sixteen lines down then fails, and under `set -euo pipefail` that reds the deploy job for every env on every push to main — while the checker still has nothing to read. The failure isn't gradual; it's the first run after merge.

The immutable-safe path is attaching the file at create time (`gh release create <tag> <files>` does draft → upload → publish internally). That only works with a tag unique per publish, i.e. the one-release-per-`(service, env, sha)` restructure — which also turns write-once into a platform guarantee rather than a curl-probe convention.

Same on `service-release-simple-staking.yml:170`.

.github/workflows/service-release-vault.yml
```diff
@@ -225,3 +232,48 @@ jobs:
...
+              --title "Frontend integrity manifests" \
+              --notes "Append-only sha256 manifests for the frontend integrity checker (devops-backlog#253). Do not delete." \
+              --latest=false || true
+          for ASSET in \
+            "/tmp/vault-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt" \
```
gbarkhatov: **Vault asset names don't match what the checker fetches — every vault check 404s from day one (verified cross-repo).**

@jonybur raised this; confirming it byte-exactly against `feat/frontend-integrity-checker`, because it's a small fix that otherwise silently kills half the feature.

Vault's matrix values are already service-prefixed (`vault-devnet`, `vault-staging-devnet`, `vault-testnet`), so this line publishes `vault-vault-testnet-<sha>-files-checksum.txt`.

Both checker runtimes build the name from `targets.json` `{"service":"vault","env":"testnet"}`:

```go
// frontend-integrity-checker/main.go:346
u := fmt.Sprintf("%s/%s-%s-%s-files-checksum.txt", g.base, service, env, sha)
```

```js
// frontend-integrity-checker/cloudflare-worker/src/index.mjs:10
const res = await fetch(`${base}/${service}-${envName}-${sha}-files-checksum.txt`);
```

Both fetch `vault-testnet-<sha>-files-checksum.txt` — a name this PR never publishes.

Staking lines up (`simple-staking` + `mainnet` → `simple-staking-mainnet-<sha>-…`), so the mismatch is vault-only and won't surface in a staking smoke test. `targets.json` monitors exactly two targets, `staking-mainnet` and `vault-testnet`, so this is half the monitored coverage erroring permanently.

Worth deriving the asset name from an explicit `service`/`env` pair rather than reusing `matrix.environment`, and pinning the exact name in the checker's selftest so the two repos can't drift again.

.github/workflows/service-release-vault.yml
```diff
@@ -225,3 +232,48 @@ jobs:
...
+            "/tmp/vault-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt"; do
```
gbarkhatov: **`global-checksum.txt` has no consumer — it doubles the asset burn for nothing.**

This one's on me: I asked for it last round on the theory that the root hash is the cheap thing for a checker to pin. Having now read the consumer, the theory was wrong — `global-checksum` appears nowhere in either runtime. Both fetch only `<service>-<env>-<sha>-files-checksum.txt` and hash live assets against the parsed manifest.

So every deploy publishes a second asset nobody reads, which doubles the rate the release fills and halves the effective retention of the prune below. Suggest dropping it and adding it back only alongside a consumer — the global hash is derivable from the files manifest anyway.

It also removes the non-atomic-pair problem @jonybur raised: two cryptographically linked assets uploaded sequentially, where a failure between them freezes a mismatched pair permanently under write-once.

.github/workflows/service-release-vault.yml
gbarkhatov: Confirming this against the repo's actual cadence, because I think it's worse than a tuning problem.

Measured: 105 commits to main in the last 30 days, and 111 `vault-publish` + 105 `simple-staking-publish` runs on main over the same window. Neither workflow has a `paths:` filter, so essentially every push fires both — 4 staking envs + 3 vault envs, ×2 files = **14 assets/push ≈ 49/day**. `KEEP=500` is therefore **~10 days** of history, which matches your estimate.

The part that makes it a correctness bug rather than a threshold to tune: the checker's `targets.json` is exactly `[staking-mainnet, vault-testnet]`, and staking mainnet deploys from `simple-staking/*` tags — not from the main-push firehose generating those 49/day. Its live sha is routinely far older than 10 days while main keeps evicting it. The prune deletes the manifest for the single most important monitored build, and the checker then reports `could-not-verify` on it forever.

Agreed on your framing — retention bounded by max deploy age, not count. Though if the release goes per-`(service, env, sha)` to survive Immutable Releases, the prune disappears and takes this with it.

.github/workflows/service-release-simple-staking.yml
gbarkhatov: Worth widening this one: it isn't only staking. `continue-on-error` is active on **both** monitored targets, so the fail-open covers 100% of the checker's coverage rather than part of it.

- staking `:63` — `continue-on-error: ${{ matrix.prod == true }}` → unconditional on `mainnet`.
- vault `:113` — `continue-on-error: ${{ matrix.prod == true && needs.setup.outputs.multi == 'true' }}`. `vault-testnet` is the `prod: true` entry, and a push to main leaves `inputs.ENVIRONMENT` empty so the setup job selects all three envs and sets `multi=true`. CoE is therefore active on `vault-testnet` for every ordinary deploy.

Since `targets.json` is `[staking-mainnet, vault-testnet]`, a manifest-publish failure on either monitored env leaves the run green with S3 already deployed and the live bundle unattested. The narrowing added for vault only helps a scoped single-env dispatch, which isn't the path that actually deploys.

toroviej0x0:
@jonybur excellent catch on Immutable Releases — you're right, the shared-release design could never work here. Reworked in 05453382 along the lines you suggested:

**One immutable release per (service, env, sha)** — tag `fm-<service>-<env>-<sha>`, both checksum files attached at `gh release create`. That resolves the blocking set:

1. **Immutable Releases**: attach-at-create is the immutable-safe path; write-once is now a platform guarantee, not a curl convention.
2. **Prune eviction**: prune deleted entirely — 2 assets per release, no cap pressure, no still-live-prod manifest ever evicted.
3. **Vault name mismatch**: contract pinned as "prefix = `<service>-<matrix.environment>-<sha>`", i.e. vault publishes `vault-vault-testnet-<sha>-…`. The checker's vault target env is now the matrix value `vault-testnet` (babylon-dapp-deployment#321, cda1511), so both sides derive the same name. Noted in a comment at the publish step.
4. **Same-sha rebuilds**: sanctioned rebuilds take the next free `-rN` slot (publisher probes upward, creates at first free; a lost create race just advances). The checker probes slots upward and verifies against the **highest** — the build whose bytes are actually in S3. A tag cut on an already-built commit no longer becomes a permanent false tamper page. Trade-off documented: an actor with repo-write could publish a higher slot — consistent with the stated threat model (post-build CDN/edge tampering, not repo/build compromise).
5. **Atomicity**: both files in one create — no partial-pair window.

On the non-blocking items: separate publish job + prod `continue-on-error` remain follow-ups (the step is still last + `set -euo pipefail` + verify-after-create); ref-gate/S3 mismatch on feature-branch dispatch affects only devnet envs the checker doesn't monitor — documented; HEAD-probe transient on the checker side now fails toward "check error" (monitoring-gap alert), not a false tamper page; rollback liveness documented as an accepted limitation in the checker README.

jonybur (commented):
Reviewed the rework at head `05453382`. The five blocking findings from my last review at `f3a58614` all land in the code:

1. **Immutable Releases** — one release per (service, env, sha) with both files attached at `gh release create`. Attach-at-create is the immutable-safe path; write-once is now a platform guarantee rather than a curl-probe convention.
2. **Prune eviction** — removed entirely; 2 assets per release, no cap pressure, no still-live-prod manifest gets deleted.
3. **Vault naming mismatch** — pinned on both sides as `<service>-<matrix.environment>-<sha>` (checker's vault env moved to `vault-testnet`, publisher yields `vault-vault-testnet-<sha>`), and the intentional double prefix is documented at the step.
4. **Same-sha rebuilds** — the `-rN` slot design lets sanctioned rebuilds (tag on an already-built commit, redeploy dispatch, job re-run) publish new bytes instead of becoming permanent false tampers.
5. **Two-asset atomicity** — both files in the single `gh release create`; no partial-pair window.

One new concern in the `-rN` design worth flagging before it ships:

**Concurrent same-sha runs can invert the slot order relative to S3, producing a permanent false tamper.** Neither workflow has a `concurrency:` block, so overlapping runs on the same sha are possible (unified-redeploy `workflow_dispatch` while a push run is still in flight, two dispatches back-to-back, or a rerun triggered near a dispatch). A bad interleaving:

- Run A: `aws s3 cp <sha>.tar.gz` (bytes A) — S3 = A
- Run B: `aws s3 cp <sha>.tar.gz` overwrites — S3 = B
- Run B: wins the `fm-<...>-<sha>` slot-1 create (bytes B)
- Run A: loses slot-1, advances, creates `fm-<...>-<sha>-r2` (bytes A)

Highest slot is now `-r2` (A) while S3 serves B. Under the stated \"checker verifies against the highest slot\" semantics, the checker permanently reports mismatch on a sha the CDN is serving cleanly — and write-once forbids repair without a fresh run to bump another slot. The failure direction is the alarming one (false tamper on a sha in prod), not the safe one.

A `concurrency:` group on `s3_publish` scoped to `(workflow, matrix.environment, github.sha)` serialises same-sha same-env runs and closes it:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ matrix.environment }}-${{ github.sha }}
  cancel-in-progress: false
```

`cancel-in-progress: false` matters here — the second run needs to publish its own slot after the first is fully done (S3 + manifest), so a rerun's higher slot always corresponds to the S3 bytes that are live.

Same suggestion on both files. Not tagging as blocking on top of the open threads jrwbabylonlab and gbarkhatov already have on `contents: write` scope, `github.ref_name` injection, and the prod `continue-on-error` — but this is a design-level correctness issue in the new slot scheme rather than a residual from the old one.

CI: `visual` still pending at this head.

.github/workflows/service-release-vault.yml
toroviej0x0: Resolved by the per-sha redesign (05453382): the prune is gone entirely. One immutable release per (service, env, sha) — nothing is ever evicted, so a tag-pinned prod sha keeps its manifest indefinitely. Retention is unbounded by design now.

.github/workflows/service-release-vault.yml
toroviej0x0: Resolved by the redesign: no prune, and Immutable Releases means a published per-sha release can't be re-uploaded under the same name. A rebuild takes the next -rN slot instead of silently replacing the trusted record.

.github/workflows/service-release-vault.yml
toroviej0x0: Resolved by the redesign: the prune (and its --json assets read, and the swallowed diagnostics) is deleted. Nothing left to no-op.

.github/workflows/service-release-vault.yml
toroviej0x0: Partly resolved, remainder tracked. The deletion path is gone (no prune — the token only ever creates releases now), and Immutable Releases blocks editing/deleting any published release including the core-ui/ts-sdk packages, so the write token's blast radius on existing releases is zero at the platform level. What remains is create-scope exposure while pnpm install/build run — the separate no-install publish job is the fix and stays a tracked follow-up.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed in 126e4158 — `GIT_TAG` is bound via `env:` and quoted in both workflows' tag-upload steps; nothing from the refname reaches the shell as code.

.github/workflows/service-release-simple-staking.yml
toroviej0x0: Fixed in 126e4158. `continue-on-error` on the prod legs now applies only OFF deploy refs (feature-branch OIDC failures — its actual purpose). On main/release/tags a failed prod deploy or manifest publish reds the run, in both workflows — covers both monitored envs incl. the ordinary push-to-main path gbarkhatov traced. Backstop unchanged: a missing manifest for a deployed sha also fires FrontendIntegrityCheckError.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed in 05453382 exactly as described: one release per (service, env, sha) with the files attached at `gh release create` (draft→upload→publish internally = immutable-safe). Write-once is now the platform guarantee. Thanks for the 738-release verification — that made the fix direction unambiguous.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed across both PRs: contract pinned as prefix = `<service>-<matrix.environment>-<sha>` (comment at the publish step), and the checker's vault target env is now the matrix value `vault-testnet` (dapp#321 cda1511), so both sides derive `vault-vault-testnet-<sha>-…` identically. Go and Worker both updated.

.github/workflows/service-release-vault.yml
toroviej0x0: Kept, because the per-sha redesign removes both stated costs: the pair is attached atomically in one `gh release create` (no partial-pair window) and there's no shared release to fill (2 assets per release, no prune, no retention pressure). The root hash rides along at zero marginal cost and is the natural anchor if we add attestation later. If you still prefer dropping it, happy to — one-line change.

toroviej0x0:
## Ready for re-review — summary of the final design

All review threads answered and resolved. The publish mechanism was redesigned after the Immutable Releases finding; current state (`126e4158`):

### Design
**One immutable release per (service, env, sha)** — tag `fm-<service>-<env>-<sha>`, both checksum files attached at `gh release create` (draft→upload→publish internally, the only path that works with this repo's Immutable Releases). Same-sha rebuilds (tag cut on a built commit, redeploy dispatch, re-run) take the next free `-rN` slot; the checker probes slots upward and verifies against the highest — the build whose bytes are actually in S3.

### What each blocking finding became
| Finding | Resolution |
|---|---|
| Immutable Releases → first upload 422s | attach-at-create; write-once is now a **platform guarantee** |
| Prune evicts still-live prod manifests | prune deleted — nothing is ever evicted, retention unbounded |
| Vault name mismatch (`vault-vault-testnet`) | contract pinned: prefix = `<service>-<matrix.environment>-<sha>`; checker env set to `vault-testnet` (dapp#321) — both sides derive identical names |
| Same-sha rebuild = permanent false alarm | `-rN` slots + checker takes highest |
| Non-atomic 2-file pair | both files in one create |

### Hardening in this round
- `github.ref_name` no longer interpolated into `run:` — bound via `env:` and quoted (refnames accept shell metacharacters; job holds OIDC + `contents:write`).
- `continue-on-error` on prod legs applies **only off deploy refs** — on main/release/tags a failed prod deploy or manifest publish reds the run (both workflows, both monitored envs).
- Publish step is last (can't block S3), gated to deploy refs (dispatch from arbitrary branches can't notarize builds), `set -euo pipefail`, verify-retrievable after create.
- `persist-credentials: false` on the publish job's checkout.

### Accepted limitations / follow-ups (documented, not silent)
- Scheme detects **post-build CDN/edge tampering**, not build/CI compromise — an actor with repo-write could publish a higher `-rN` slot (attestation via `attest-build-provenance` is the upgrade path, needs a consumer change, tracked).
- Separate no-install publish job (removes create-scope token exposure during `pnpm install`) — follow-up.
- Rollback to a previously-manifested build verifies clean (content-addressed store) — accepted.
- `global-checksum.txt` kept (atomic attach, zero marginal cost, future attestation anchor) — say the word if you'd rather drop it.

Companion changes are already on the checker branch (dapp#321 `cda1511`: slot probing, vault env name, MANIFEST_BASE root).

.github/workflows/service-release-vault.yml
```diff
@@ -221,7 +229,57 @@ jobs:
...
+            fi
+            if gh release create "$TAG" "${FILES[@]}" \
+                 --title "Frontend manifest ${PREFIX}" \
+                 --notes "sha256 manifest for the frontend integrity checker (devops-backlog#253). Do not delete." \
+                 --latest=false; then
+              curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
+              echo "published ${TAG}"
+              exit 0
+            fi
+            echo "create ${TAG} lost a race — trying next slot"
```
jrwbabylonlab, line 282: A transient `gh release create` failure leaves a gap in the numbering, then the loop publishes the next number and exits green. The checker stops at the first missing number (`main.go:356-370`), so it never finds the manifest — integrity verification is silently off for that build.

Suggest advancing only when the release already exists, and failing on any other error.

toroviej0x0: Fixed in fc89eec6. The loop now captures stderr and only advances to the next slot when create failed with "already exists" (real race); any other error fails the step with the message. No gap can form, so the checker's first-miss-stop probe stays correct.

.github/workflows/service-release-vault.yml
```diff
@@ -98,20 +98,28 @@ jobs:
 
   s3_publish:
     needs: [lint_test, setup]
+    permissions:
```
gbarkhatov, line 101: **No `concurrency:` on this job — overlapping same-sha runs can invert slot order against S3.**

Anchoring @jonybur's 2026-08-10 finding to a line, because it landed in a review body with no anchor and hasn't been picked up — the same failure mode that lost the Immutable Releases finding for ten days. Verified at head: `grep -c concurrency` is 0 in both workflows, and the word appears in none of the replies since.

The two sides of the `-rN` scheme disagree about which slot is authoritative. The publisher takes the **first free** slot (the loop at `:267`); the checker verifies against the **highest occupied** one. Those are the same slot only if same-sha runs are serialised, and nothing serialises them: `git push --follow-tags` starts the main-push and tag-push runs concurrently on one sha, and the unified redeploy dispatch can overlap an in-flight push run.

- run A: `aws s3 cp <sha>.tar.gz` → S3 = A
- run B: same fixed key, last-write-wins → S3 = B
- run B: wins the slot-1 create → manifest slot 1 = B
- run A: loses the create, advances → `-r2` = A

Highest slot is now A while S3 serves B, so the checker reports tamper on a sha the CDN is serving cleanly. It isn't permanent — any later run for that sha rewrites S3 and takes a higher slot together — but until someone works that out, a monitored prod target is paging falsely, which is the alarming direction rather than the safe one.

Four lines on this job close it:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ matrix.environment }}-${{ github.sha }}
  cancel-in-progress: false
```

`cancel-in-progress: false` is the load-bearing part — the second run has to publish its slot *after* the first has finished both its S3 write and its manifest. Gating at job level rather than step level is what keeps the two stores in the same order. The known trade-off is that a third queued run for the same `(env, sha)` cancels the second pending one; that's fine here, because a cancelled job writes neither S3 nor a manifest, so the stores stay consistent.

Same on `service-release-simple-staking.yml:43`.

.github/workflows/service-release-vault.yml
```diff
@@ -98,20 +98,28 @@ jobs:
...
+      id-token: write # OIDC assume-role for AWS
+      contents: write # publish checksum manifest as a release asset (#253)
```
gbarkhatov, line 103: **Re-anchoring the `contents: write` scope so it survives the merge.**

Crediting what actually changed since @jrwbabylonlab raised this: the deletion path is gone (the prune went with the redesign, so this token only ever creates), `persist-credentials: false` closes the `.git/config` vector, and Immutable Releases means the published core-ui / ts-sdk artifacts can't be rewritten. The blast radius really did shrink from "rewrite any release" to "create new releases and tags".

What's left is unchanged: the token is still minted on the job that runs `pnpm install` and `pnpm run build`, so a malicious postinstall can poison `$GITHUB_ENV`/`PATH` and capture `GH_TOKEN` when the publish step runs several minutes later. Create-scope is enough to publish a higher `-rN` slot for any sha, which under the "checker takes the highest slot" rule is the whole store.

Not blocking, and I agree it's out of scope here — the scheme's stated threat model is post-build CDN/edge tampering, not build compromise, and a build-time compromise could forge manifest content regardless. The ask is only that "tracked follow-up" resolves to an actual issue linked from this PR rather than to a resolved comment thread. Same for the composite-action thread.

.github/workflows/service-release-vault.yml
```diff
@@ -221,7 +229,65 @@ jobs:
...
+          # redeploy dispatch, job re-run) and are NOT byte-reproducible (run_id
+          # in the bundle), so each rebuild takes the next free -rN slot. The
+          # checker probes slots upward and verifies against the highest one.
+          # NOTE: vault matrix env values are already service-prefixed
+          # (vault-testnet), so PREFIX is vault-vault-testnet-<sha>; the checker
+          # config uses env "vault-testnet" to derive the same name.
+          PREFIX="vault-${{ matrix.environment }}-${{ github.sha }}"
+          FILES=("/tmp/${PREFIX}-files-checksum.txt" "/tmp/${PREFIX}-global-checksum.txt")
+          for i in 1 2 3 4 5 6 7 8 9 10; do
+            TAG="fm-${PREFIX}"; [ "$i" -gt 1 ] && TAG="${TAG}-r${i}"
```
gbarkhatov: **Nobody has quantified what this mints: roughly 7 releases and 7 git tags per push to main.**

3 vault envs + 4 staking envs, and neither workflow has a `paths:` filter, so effectively every push to main fires both. At the cadence measured earlier in this review (~105 pushes to main in 30 days) that's **~735 releases and ~735 tag refs a month**. The repo currently holds 740 releases and 760 tags in its *entire history* — so this adds a full repo-history's worth of both every month, ~8.8k/year.

That's not an argument against the per-sha design, which is the right shape. It's that the cost should be a decision rather than something discovered in six months:

- The Releases page and `gh release list` become `fm-*` noise, and those are what people and the npm publish flow use to find `@babylonlabs-io/core-ui`, `ts-sdk`, `simple-staking/*` and `vault/*`.
- Every full clone and `git fetch --tags` carries the ref growth. CI is unaffected — `actions/checkout` defaults to `fetch-tags: false`.
- No retrigger risk, for the record: `fm-*` matches neither workflow's tag filter, and tags created with `GITHUB_TOKEN` don't start workflow runs.

Cheapest option if you want to cut it: publish only for the environments the checker actually monitors. `targets.json` is exactly `staking-mainnet` and `vault-testnet`, so gating on that takes 7 per main push down to 1 (vault-testnet), plus 1 per staking release tag. Everything else is currently manifesting builds nothing reads. Alternatively keep all seven and say so in the PR body — either is fine, silence isn't.

.github/workflows/service-release-vault.yml
```diff
@@ -221,7 +229,65 @@ jobs:
...
+                 --latest=false 2>&1); then
+              curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
+              echo "published ${TAG}"
+              exit 0
+            fi
+            # Advance to the next slot ONLY on a create race ("already exists").
+            # Any other failure must fail the step: skipping it would leave a
+            # numbering gap, and the checker stops at the first missing slot —
+            # the next run's higher slot would never be found (silent gap).
+            if ! grep -qi "already exists" <<<"$ERR"; then
```
gbarkhatov: **Non-blocking hardening: the occupancy signal and the race signal disagree about what "occupied" means.**

First, to be fair to this fix — I checked before writing, and the pattern does match today. `gh` doesn't surface the raw API code (`Release.tag_name already_exists`); it emits its own message, `a release with the same tag name already exists: <tag>` (the literal is in the 2.97 binary). So the create-race path works as intended and this isn't a live bug.

The residual is structural. Occupancy is judged by the **public asset URL** (`:270`), while the race is judged by **gh's human-readable error text** (this line). A release that exists but isn't publicly downloadable satisfies neither: `gh release create` with assets creates a draft, uploads, then publishes, so a step killed mid-upload (runner eviction, cancelled run, `--latest=false` patch failing) leaves a draft holding that tag name. What happens on the next run then depends on whether GitHub rejects a duplicate `tag_name` among drafts, which isn't documented either way — best case an orphan draft accumulates, worst case the loop reads the 422 as a lost race and advances, leaving the numbering gap that the checker's stop-at-first-miss probe can't cross.

One re-probe makes the loop correct under either behaviour, and takes a CLI string out of a decision that now gates a prod deploy:

```bash
# after a failed create, re-probe: only a *published* competitor is a real race
if curl -sfIL -o /dev/null "$URL"; then
  echo "slot ${TAG} published by a concurrent run — trying next slot"
  continue
fi
echo "::error::gh release create ${TAG} failed and ${URL} is not retrievable: ${ERR}"
exit 1
```

Same on `service-release-simple-staking.yml:205`.

.github/workflows/service-release-simple-staking.yml
```diff
@@ -141,7 +151,62 @@ jobs:
...
+          # Bound via env, not interpolated into run: — git refnames accept
+          # shell metacharacters and this job holds AWS OIDC + contents:write.
+          GIT_TAG: ${{ github.ref_name }}
+        run: |
+          aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}.tar.gz "s3://${{ vars.S3_PUBLIC_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${GIT_TAG}.tar.gz"
+          aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-files-checksum.txt "s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${GIT_TAG}-files-checksum.txt"
+          aws s3 cp /tmp/simple-staking-${{ matrix.environment }}-${{ github.sha }}-global-checksum.txt "s3://${{ vars.S3_CHECKSUM_BUCKET_NEW }}/simple-staking/${{ matrix.environment }}/${GIT_TAG}-global-checksum.txt"
+
+      # Last step: publishing the manifest must not block the S3 uploads above.
+      - name: Publish checksum manifest as an immutable per-sha release
```
gbarkhatov, line 174: **The duplicated block is now ~50 lines, up from the 25 I flagged last round.**

Diffed it: this step and `service-release-vault.yml:241-293` are byte-identical apart from the service prefix and vault's 3-line naming NOTE. Deferring the composite action was reasonable while the diff was still churning — it's a weaker argument for the permanent shape, because this is the code that decides what lands in the trusted store, and the next service will copy it a third time.

Same ask as the token-scope thread: an issue linked from here rather than a resolved comment thread. The `service` / `environment` / `sha` inputs are obvious, and it would let the separate no-install publish job be built once instead of twice.

gbarkhatov (dismissed):
**Approving the rework.** I re-checked all five blocking findings from my last review against head `fc89eec6` and they're genuinely in the code, not just in replies: attach-at-create per `(service, env, sha)`, prune gone, naming contract pinned on both sides, `-rN` slots for sanctioned rebuilds, single atomic create. The round-3 hardening landed too — `GIT_TAG` bound via `env:`, `continue-on-error` narrowed to non-deploy refs so both monitored envs now fail loud, publish step last, ref-gated, `set -euo pipefail`, verify-after-create.

Two things aren't done, and one of them is a correctness bug in the new design rather than a residual of the old one.

**1. @jonybur's concurrency finding (2026-08-10) has no code and no reply.** Details inline on `service-release-vault.yml`. It's the same failure mode that lost the Immutable Releases finding for ten days: it landed in a review body without a line anchor, and the "Ready for re-review" summary postdates it without mentioning it. Four lines fix it.

**2. The PR description still documents the design this PR abandoned.** It describes `gh release upload` into a pinned `frontend-manifests` release, `create ... || true` for idempotency, an append-only per-sha asset store, and "prune old ones later if desired" — and it gives the checker's read URL as `releases/download/frontend-manifests/<service>-<env>-<sha>-files-checksum.txt`, which is not a URL this PR produces. Body drift was flagged two rounds ago and the rework widened it. This body is the de-facto contract doc for babylonlabs-io/babylon-dapp-deployment#321 and it's what lands in the merge commit — please rewrite it to the `fm-<service>-<env>-<sha>` / `-rN` scheme, with the intentional `vault-vault-testnet` double prefix spelled out.

**On `global-checksum.txt` — you asked, so: keep it.** My objection was asset burn and the non-atomic pair, and the per-sha redesign removed both; it now rides along at genuinely zero marginal cost. Worth one line in the checker README noting it's currently unconsumed, so nobody later assumes a green check verified it.

**On the two threads closed as "tracked follow-up"** — separate no-install publish job, and the composite action. Both really are follow-ups and deferring them is the right call. But "tracked" currently resolves to a resolved comment thread, which is exactly where the last unanchored finding went to die. Issues with links from here, please.

I've also flagged inline something nobody has quantified: at the measured cadence this mints ~735 releases and ~735 git tags a month, against a repo that holds 740 releases and 760 tags in its entire history. Not an objection to the design — just a number that should be decided on rather than discovered.

**Nit:** the release notes still say "Do not delete." as an instruction. With Immutable Releases that's enforced now, so it can state the fact instead of asking.

jonybur (dismissed):
The per-sha immutable-release design lands the invariants this needs: one release per (service, env, sha), both files attached atomically at create time, no prune, and same-sha rebuilds take the next -rN slot. The fix at fc89eec6 closes the numbering-gap concern by advancing only on 'already exists' and failing loudly on any other create error, which keeps the checker's first-miss-stop probe correct. Other safety controls are in place: GIT_TAG bound via env before shell interpolation, persist-credentials: false on the checkout, continue-on-error narrowed to non-deploy refs so a prod failure or missing manifest reds the run on main / release / tags, and the publish step is now last so a Releases API hiccup can't block the S3 uploads. The known follow-ups (composite action to dedupe the ~50-line block across the two workflows, separate no-install publish job to shrink the contents:write blast radius during pnpm install/build, a paths: filter and/or an asset-count alarm at the current ~735 releases/month cadence, dropping global-checksum once the consumer confirms it's unused, and job-level concurrency) are all worth landing as fast-follows but do not block this.

.github/workflows/service-release-vault.yml
toroviej0x0: Fixed in fb5fb3b4 — added your exact concurrency block (group per workflow+env+sha, cancel-in-progress: false) on the s3_publish job in both workflows. Same-sha runs are now serialized, so S3 order and slot order can't invert.

.github/workflows/service-release-vault.yml
toroviej0x0: Took your suggestion in fb5fb3b4 — the race check is now the URL re-probe: after a failed create, only a slot actually retrievable at the public download URL advances the loop; anything else (transient, draft, permission) fails with the error. gh's error text is out of the decision entirely.

.github/workflows/service-release-simple-staking.yml
toroviej0x0: Done in fb5fb3b4 — publish logic now lives in .github/actions/publish-integrity-manifest (composite action, service/environment inputs), both workflows call it. Local path uses are commit-pinned by definition so the pinned-actions rule holds. No more duplicated block to drift.

.github/workflows/service-release-vault.yml
toroviej0x0: Cut it in fb5fb3b4: manifests are now published only for the envs the checker monitors — staking mainnet/testnet and vault-testnet — via an env allowlist on the step's if. That's ~3 releases+tags per push instead of ~7 (~315/month vs ~735). The comment ties the list to the checker's targets.json so they extend together. Devnet builds mint nothing. If volume still bites later, retention/deletability of immutable fm-* releases is the follow-up to investigate.

.github/workflows/service-release-vault.yml
toroviej0x0: Agreed with your accounting — remaining exposure is create-scope while pnpm install/build run untrusted code, and the fix is the separate no-install publish job consuming the checksums as artifacts. Tracked as the standing follow-up (with the attestation work, since both touch the same job split). Not silently dropped — it's in the PR summary's follow-ups list.

jonybur (approved):
Approving. The per-service-env-sha redesign is the right shape for Immutable Releases: both checksum files are attached at gh release create time so the two-file publish is atomic, and the -rN slot loop only advances when the previous slot is actually retrievable at the public download URL - that closes the numbering-gap failure mode the checker relies on.

Verified the hardening from earlier rounds is present at head:
- concurrency block on s3_publish in both workflows, group keyed by workflow+env+sha, cancel-in-progress: false, so same-sha runs serialize and S3 order cannot invert against release order.
- github.ref_name bound via env GH_TAG before interpolation, no shell metacharacters reach the run block.
- contents: write scoped to s3_publish only; persist-credentials: false on the publish job's checkout.
- Env allowlist limits publishes to staking mainnet/testnet and vault-testnet; the composite action lives at .github/actions/publish-integrity-manifest so the duplicated block is gone.
- continue-on-error on prod legs now only applies to non-deploy refs.

Follow-ups such as a separate no-install publish job consuming the checksums as artifacts, and an asset-count alarm, are legitimate but non-blocking.

jonybur (approved):
Approving the current head. The per-sha redesign (fb5fb3b4) plus the composite-action extraction resolve every blocking finding from the prior rounds, and the added hardening holds up.

Verified against the head:

- **Immutable-safe publish** — one release per (service, env, sha), files attached at `gh release create` (draft→upload→publish internally), no post-create `gh release upload` that would 422 under Immutable Releases. Write-once is now a platform guarantee, not a curl convention.
- **-rN slot ordering vs S3** — the concurrency block (`group: workflow-env-sha`, `cancel-in-progress: false`) serialises same-sha runs, so slot creation order matches S3 write order; a lost create race advances only when `curl -sfIL` proves the slot is publicly retrievable (the same signal the checker uses), which closes the numbering-gap failure mode. Non-race `gh release create` failures capture stderr into `ERR` and exit 1 with that message rather than silently advancing.
- **Vault name contract** — the vault workflow's env allowlist is `["vault-testnet"]`, the composite action forms `PREFIX=${SERVICE}-${ENVIRONMENT}-${GITHUB_SHA}`, and the header comment pins the resulting `vault-vault-testnet-<sha>` name against the checker's target config.
- **Ref-gated notarisation** — the publish step's `if` restricts to `main`/`release/v1.x`/tags, so a `workflow_dispatch` from an arbitrary branch cannot mint a manifest for a build that only exists on that branch.
- **Fail-loud on deploy refs** — `continue-on-error` on prod legs now applies only OFF deploy refs, so a failed prod deploy or manifest publish on main/release/tags reds the run instead of silently gapping the integrity signal.
- **Injection surface** — `GIT_TAG` bound via `env:` and quoted in both `aws s3 cp` invocations; the composite action's `SERVICE`/`ENVIRONMENT` come from static workflow inputs and `GITHUB_SHA`/`GITHUB_REPOSITORY` are trusted context. `persist-credentials: false` on the publish job's checkout keeps the token out of `.git/config` while pnpm install / build execute untrusted deps.
- **Volume** — env allowlist collapses this to ~3 per push (2 staking + 1 vault), which is a sustainable per-sha release surface. `--latest=false` keeps the tags out of the "Latest" badge.
- **Composite action** — write-once + retry logic now lives in one place (`.github/actions/publish-integrity-manifest`) with typed inputs, so the two workflows cannot drift.

The known follow-ups (separate no-install publish job to shed create-scope during pnpm install, provenance attestations, tag ruleset restricting delete/update on the `fm-*` tags) are correctly deferred as consumer-coupled changes.

jonybur (approved):

PR #2169, opened 2026-07-31
feat(ts-sdk): band the payout implicit fee between floor and ceiling

jrwbabylonlab:


jonybur (commented):
The band math itself held up under scrutiny: floor and ceiling are inclusive and pinned by at-bound and off-by-one-sat tests on both ends, all sat arithmetic stays in bigint, the floor's inputs come from the version-locked vaultCoreVersion, the on-chain participant lists and the stamped offchainParams.feeRate, both WASM wrappers funnel the return through assertWasmBigint, and the removed MAX_PAYOUT_FEE_FRACTION_* surface is fully gone with docs updated. Three findings:

1. `packages/babylon-tbv-rust-wasm/scripts/build-wasm.js` - the VAULT_WASM_COMMIT bump to `ae00321b` keeps the old pin comment ("main incl. PR #2 (P2A anchor exports); bundles btc-vault v1 @ a0ad5503, v2 @ d7e33b26"). That comment is the human-readable record of what the vendored binary contains, and CLAUDE.md treats this pin as part of the frozen vault-secret surface with a golden-vector gate to re-run on every bump. Please update it to describe the new rev (the payout-fee-floor export and the VP output-sizing model set) and the btc-vault revs it bundles, and state explicitly in the PR that the three expander outputs are unchanged at this pin - the un-skipped acceptance-gate test covers it in CI, but the claim belongs in the record.

2. `packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/depositorPayout.ts` - buildDepositorPayoutPsbt constructs the same depositor payout PSBT but gets no fee band; only buildPayoutPsbt is guarded. It is exported from primitives/index.ts and documented in docs/quickstart/primitives.md, so an external SDK consumer following the documented primitives path pre-signs with no implicit-fee check at all. Pre-existing and unused by the monorepo flows, but it undercuts the guarantee this PR establishes - either thread the band through it or mark it loudly as unguarded.

3. `assertPayoutFeeBand.ts` - assertPayoutFeeBandDomain validates protocolFeeRate and both participant counts but forwards councilSize to the WASM floor with no integer or range check. Low impact today (it is sourced from securityCouncilKeys.length and the floor is council-invariant at the current pins), but it is the one band input that crosses the WASM boundary unvalidated - worth adding to the same domain check so the accepted domain stays equal to the proven domain.

packages/babylon-ts-sdk/docs/api/primitives.md
```diff
@@ -2087,8 +2113,27 @@ If sum of output values exceeds sum of input values (invalid tx)
...
+
+If implicit fee exceeds the ceiling `protocolFeeRate *
+  (MAX_PAYOUT_VSIZE_BASE + MAX_PAYOUT_VSIZE_PER_PARTICIPANT *
+  (keepers + challengers) + script excess over
+  {@link PAYOUT_BOUND_ASSUMED_SCRIPT_LEN})`
+
+#### Throws
+
+If the commission scriptPubKey length is outside `[1,
+  {@link MAX_PAYOUT_SCRIPT_LEN}]`, or participant counts exceed
```
jrwbabylonlab: Generated from an intermediate revision. Source has three fee-band `@throws` (`payout.ts:185-195`), worded differently, and the five symbols referenced here (`MAX_PAYOUT_VSIZE_BASE`, `MAX_PAYOUT_VSIZE_PER_PARTICIPANT`, `MAX_PAYOUT_PARTICIPANTS_PER_ROLE`, `MAX_BASE_FEE_RATE_SAT_PER_VB`, `PAYOUT_BOUND_ASSUMED_SCRIPT_LEN`) are module-private in `assertPayoutFeeBand.ts`, so the links resolve to nothing in the published reference. The anchor at `:2066` also says `payout.ts:203` while the function is at `197`. CI cannot catch this: `docs:validate` is `typedoc --emit none` and never diffs the committed md. Run `pnpm docs:clean` and commit.

jeremy-babylonlabs (dismissed):

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts
```diff
@@ -179,112 +197,50 @@ export interface PayoutPsbtResult {
...
-    commissionBps: params.commissionBps,
-  });
+    peginPrevOut.value,
+    params,
+  );
 
-  // Bound the implicit fee — blocks a VP deflating output values and burning
-  // the difference as miner fee.
-  const inputValueSats = peginPrevOut.value + input1PrevOut.value;
+  const inputValueSats = peginPrevOut.value + assertPrevOut.value;
```
gbarkhatov, line 247: **`assertPrevOut.value` is never pinned.** It comes from the VP-supplied `assertTxHex`, and nothing bounds it before it enters the fee basis.

The device does pin it — `_payout_check_witness_utxo(dc, &input_map, expected_spk, VAULT_DUST_LIMIT)` on input 1 — and the comment directly above its `max_fee` computation leans on exactly that: *"Assert:0_amount was already validated as VAULT_DUST_LIMIT by `_payout_check_witness_utxo` above, so the two are equivalent"* (`sign_psbt_validate.c` @ `c8db53e`; `VAULT_DUST_LIMIT = 546`).

Our ceiling constants are the device's, so they are calibrated to a `vault_amount + 546` basis, but we compute `pegin + assert` and assert the equivalence only in a comment (lines 31-33). An off-spec `Assert:0` silently shifts our band relative to the device's — below 546 we accept fees the device would reject, by up to 546 sats.

An explicit check that `assertPrevOut.value` equals the 546-sat connector value makes the stated invariant actually true, and catches a substituted or malformed Assert tx before signing. Two notes on doing it: it is not quite a one-liner, since `createTestAssertTransaction()` funds `Assert:0` with `TEST_CLAIM_VALUE` (50_000) and the fixtures would need updating — which is itself evidence the invariant is untested today — and I would add a named constant for the connector rather than reusing `PAYOUT_ANCHOR_DUST_SATS` (same value, different role).

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertPayoutFeeBand.ts
```diff
@@ -0,0 +1,185 @@
...
+): Promise<void> {
+  const { implicitFeeSats, out0Len, out1Len } = measured;
+  const numParticipants =
+    params.numVaultKeepers + params.numUniversalChallengers;
+  const implicitFee = BigInt(implicitFeeSats);
+
+  // Ceiling first: synchronous arithmetic, no WASM round-trip.
+  const scriptExcess =
+    Math.max(0, out0Len - PAYOUT_BOUND_ASSUMED_SCRIPT_LEN) +
+    Math.max(0, (out1Len ?? 0) - PAYOUT_BOUND_ASSUMED_SCRIPT_LEN);
```
gbarkhatov: **`out1Len` is VP-controlled and only widens the ceiling.**

The module doc at lines 14-19 says both lengths are "layout-trusted … never re-measured from the raw tx", and the WASM wrapper doc says they "must be TRUSTED script byte-lengths (registered/pinned scripts)". True for `out0Len`, false for `out1Len`: `payout.ts:506-507` measures it off the VP-supplied tx, and `assertPayoutOutputLayout`'s own docstring says `outs[1].script` is deliberately unpinned — it is only range-checked to `[1, 128]`.

Measured with the real WASM (v1, N=M=1, council 3, vsize units at 1 sat/vB):

| out0 | out1 | floor | SDK ceiling | device `max_fee` |
|---|---|---|---|---|
| 34 | 34 | 432 | 610 | 610 |
| 34 | 128 | 432 | 704 | 610 |

The floor saturates at 34 bytes (399 at `out1=1`, 420 at 22, then 432 flat for 34 / 60 / 128), so across `[34, 128]` the excess term moves the ceiling and nothing else. Padding the commission script to 128 bytes buys the VP up to `94 × feeRate` sats of extra burnable headroom for free — 1,410 sats at 15 sat/vB.

Suggest dropping `out1Len` from `scriptExcess` (the pinned `out0` excess is the part the ceiling genuinely has to accommodate), or pinning `outs[1].script` so the doc becomes true.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertPayoutFeeBand.ts
```diff
@@ -0,0 +1,185 @@
...
+import { computePayoutFeeFloor } from "@babylonlabs-io/babylon-tbv-rust-wasm";
+
+/**
+ * Payout fee-bound vsize model: `maxVsize = BASE + PER_PARTICIPANT * (N + M)`
+ * where N = vault keepers, M = universal challengers. Constants and comparison
+ * match the Ledger device (`app-babylon-vault` `sign_psbt_validate.c`
+ * MAX_PAYOUT_VSIZE_BASE/PER_PARTICIPANT; HLD v22 §4.9.7.1) — a conservative
+ * upper estimate with ~13%+ headroom over the exact vsize the VP pays. Our fee
+ * basis (true prevout fee) is strictly stricter than the device's
+ * `vault_amount + 546` proxy, so our accept-set is a subset of the device's.
```
gbarkhatov: **Qualify this claim** — and it needs qualifying even after the `out1Len` fix above, so it is a separate point.

The device's bound is flat `base_fee_rate × (500 + 55 × (N + M))` with no script-length component (verified at `app-babylon-vault` `c8db53e`, `sign_psbt_validate.c`). So any `scriptExcess > 0` — including one contributed by the *pinned* `out0` — puts our ceiling strictly above the device's, and the accept-set stops being a subset.

It is unreachable on-device today: the firmware pins out0 to BIP-86 P2TR(depositor) (34 bytes) for the VP and depositor-claimer roles, and the SDK itself derives a 34-byte BIP-86 script for the VK-claimer role. So this is a comment fix, not a code change — the claim holds for ≤34-byte scripts, which is the only case that reaches a device, and saying so is more useful to the next reader than the unconditional form.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/runDepositorPresignFlow.ts
```diff
@@ -344,6 +352,22 @@ export async function runDepositorPresignFlow(
 
   // Approval-capable wallets must approve the deposit terms before any signing
   // call they authorize, including the Phase 3/4 payout and depositor-graph signing below.
+  // The terms the depositor approved must carry the same graph-build rate the
+  // payout bound will enforce — a mismatch means a params-version drift bug.
+  // Fresh-flow only: the resume path passes no depositTerms, where the rate is
+  // correct-by-construction (read from the vault's stamped params version).
+  if (
+    depositTerms !== undefined &&
+    depositTerms.baseFeeRate !== signingContext.protocolFeeRate
```
gbarkhatov, line 362: **This covers one of the two inputs to the bound it is protecting.**

The device computes `base_fee_rate × (500 + 55 × (keeper_count + challenger_count))` with *both* counts taken from the approved intent, while the SDK derives `N + M` from the signing context's array lengths. `DepositTerms` already carries `keeperPks` / `challengerPks` — same drift class, equally checkable, unchecked. The timelocks are in the same position, since the device binds those too.

Either extend the comparison to the keeper/challenger sets, or compare the params-version stamp once and drop the per-field check. If the rate genuinely is the only field that can drift independently of the others, that reasoning belongs in the comment above — right now the check reads as if it were exhaustive.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/__tests__/assertPayoutFeeBand.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    expect(() =>
+      assertPayoutFeeBandDomain({ ...BASE_PARAMS, councilSize: 1 }),
+    ).not.toThrow();
+    expect(() =>
+      assertPayoutFeeBandDomain({ ...BASE_PARAMS, councilSize: 100 }),
+    ).not.toThrow();
+  });
+});
+
+describe("assertPayoutFeeInBand", () => {
```
gbarkhatov, line 125: **Nothing here — or anywhere — asserts the band is non-empty (`floor <= ceiling`).**

I swept it against the real WASM (v1+v2 × N,M ∈ {1,2,4,8,16,24,32} × out0 ∈ {1,34,128} × out1 ∈ {34,128,undefined}): 0 violations, tightest margin 88.2% at v1 / N=M=32 / 34-34 scripts (floor 3544 vs ceiling 4020 per sat/vB). It holds today.

But it is unguarded and asymmetric: the floor moves with `VAULT_WASM_COMMIT` (a pinned Rust model set), while the ceiling is two hardcoded JS constants that never see `vaultCoreVersion`. A pin bump — or a future v3 graph with a heavier payout — that lifts the floor above the flat ceiling would reject *every* payout of that shape, with CI green, because the cases here are N=M=1 and N=2/M=1, both far from the tight corner. A corner sweep asserting `floor <= ceiling` runs in ~10 ms and closes the whole class.

gbarkhatov (dismissed):
Approving — the band is correct as implemented, and the WASM/device reasoning holds up where I could check it independently rather than take the comments at their word. Five inline notes, none blocking; the two I'd most like resolved before merge are the unpinned `Assert:0` value and the missing `floor <= ceiling` guard.

### What I verified

- `assertPayoutFeeBand.test.ts` + `payout.test.ts` + `runDepositorPresignFlow.test.ts`: **78 pass**. ESLint clean on the changed files.
- The golden-vector gate CLAUDE.md §4 mandates on a `VAULT_WASM_COMMIT` bump — `vault-secrets/__tests__/expand.test.ts`, **16 pass** against the committed binary.
- Firmware cross-check at `app-babylon-vault` `c8db53e` (current `origin/develop`): `MAX_PAYOUT_VSIZE_BASE 500u` / `MAX_PAYOUT_VSIZE_PER_PARTICIPANT 55u`, the `[1, 32]` keeper/challenger caps, the `base_fee_rate > UINT32_MAX` reject, and `VAULT_DUST_LIMIT = 546` all match what this PR encodes. The device's `max_fee` has **no** script-length term (see the two comments below).
- Band non-emptiness, swept against the real WASM (v1+v2 × N,M ∈ {1,2,4,8,16,24,32} × out0 ∈ {1,34,128} × out1 ∈ {34,128,undefined}): 0 violations, tightest margin 88.2%.
- Dead-code removal is clean — no dangling references to `buildDepositorPayoutPsbt` / `DepositorPayoutParams`, and the deleted module had no test file.
- `numLocalChallengers = N` is right for all three claimer roles; an unsupported `vaultCoreVersion` fails closed at the WASM boundary; the tests drive the real WASM floor rather than a mock.

### Not inline

- **Operational.** The floor is a new hard-block: previously only over-payment could stop signing, now under-payment does too, and it fails mid-flow with the deposit parked at `PENDING_DEPOSITOR_SIGNATURES` and no override. Worth confirming that every VP build currently serving in-flight vaults — on every network, including older ones — is inside the pinned model set, and that the vault app maps this throw to actionable copy rather than a raw SDK error string.
- **Process, since this carries `critical-path`.** The PR body is empty while the PR bumps `VAULT_WASM_COMMIT` and ships a regenerated `vault_wasm_bg.wasm` that no reviewer can rebuild (private repo). The golden-vector gate passes, but the description should say so and record the blob's digest next to the `ae00321b` pin so the artifact is auditable: `sha256:b5257739a4c64100ef9f212bcab162478453d30e632f2c3ed63390204b010e50`.
- **Nit.** Floor cost measured here: ~10 ms at N=M=32, ~0.8 ms at N=M=1. The batch path builds one payout per claimer, so a 32/32 vault is ~0.7 s of main-thread WASM spent on identical-shape calls — the existing `TODO: memoize` is proportionate, and the key is just `(version, N, M, council, out0, out1, rate)`.
- **Nit.** The `6_100` / `7_040` ceilings are hand-derived in two test files and will need to move together if the constants change. Consistent with the repo's "explicit values over imported constants" rule, so fine as-is.

The stale generated docs are already covered in the existing review comment — not repeating it here.

packages/babylon-ts-sdk/docs/api/primitives.md
jrwbabylonlab: Resolved — regenerated.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts
jrwbabylonlab: Went with documenting instead of a check here: the sighash already commits to the value (a misstated Assert:0 just yields an unusable signature), and the real graph builds it as 546 + the council fee, so an exact pin would reject real payouts.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertPayoutFeeBand.ts
jrwbabylonlab: Resolved — dropped out1Len from the ceiling excess (kept out0). Pinning outs[1].script needs registry reads the ABI doesn't expose yet, left as a follow-up.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertPayoutFeeBand.ts
jrwbabylonlab: Resolved.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/runDepositorPresignFlow.ts
jrwbabylonlab: Keeping just the rate check — the shared fields all come from append-only versioned reads and verifyRegisteredVaultVersions already pins them, so they can't drift under current flows. Noted in the comment.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/__tests__/assertPayoutFeeBand.test.ts
jrwbabylonlab: Resolved.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertPayoutFeeBand.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  // Local challengers are exactly the keeper count for every claimer role
+  // (btc-vault graph.rs derive_challengers: VKs, or VP + N-1 VKs).
+  // TODO: memoize if keeper counts grow enough for batch flows to feel it.
+  const minFeeSats = await computePayoutFeeFloor(
+    params.vaultCoreVersion,
+    params.numVaultKeepers,
+    params.numUniversalChallengers,
+    params.numVaultKeepers,
```
jrwbabylonlab, line 185: The comment above says local challengers are the keeper count for every claimer role, but btc-vault `graph.rs` gives `dedup(VP ∪ VKs) - {claimer}` for VP/VK claimers, which is N-1 when the VP key is also a registered keeper. The depositor presign flow is safe — `runDepositorPresignFlow` rejects that case on the duplicate check — but `PayoutManager.signPayoutTransaction` does not go through it, so a direct SDK consumer on such a vault gets a floor 25 vB x feeRate too high and every payout rejected. Worth stating why the guard makes this safe rather than asserting it always is.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertPayoutFeeBand.ts
```diff
@@ -0,0 +1,185 @@
+/**
```
jrwbabylonlab, line 1: The payout fee bound used to live in `payout.ts`, which is named in CLAUDE.md §3, CODEOWNERS, `critical-path-check.yml` and `claude-md-drift.yml`. This file is in none of them. Today the call site still trips the check, but a future PR editing only the vsize constants or the floor call bypasses two-reviewer enforcement on value-movement logic.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertPayoutFeeBand.ts
```diff
@@ -0,0 +1,185 @@
...
+    ) {
+      throw new Error(
+        `Participant count for ${role} (${count}) is outside the device ` +
+          `range [1, ${MAX_PAYOUT_PARTICIPANTS_PER_ROLE}].`,
+      );
+    }
+  }
+  // councilSize crosses the WASM boundary with no glue-side validation; a
+  // non-integer would silently truncate at the u32 ABI and a 0 would be
+  // silently promoted to a 1-member council by the pinned estimator.
```
jrwbabylonlab, line 129: `councilSize` does not affect the floor — measured across 11 values against the shipped WASM, the result is identical every time, because the upstream weight prediction reads only `challengers.len()`, script size and tree depth, and the council occupies one leaf regardless of size. It is a new required public field threaded through five files for a value that cannot change the output. Also bounded below but not above, while citing u32 truncation as the motivation.

jonybur (approved):
Re-reviewed at b45816dd. All three findings from the earlier round are addressed:

- `build-wasm.js` pin comment updated to reference PR #3 (`computePayoutFeeFloor` export) with the bundled btc-vault revs still called out, so the pin's contents are auditable at a glance again.
- The unguarded `buildDepositorPayoutPsbt` / `DepositorPayoutParams` surface is gone from primitives with no dangling references, so the documented depositor-payout path now runs through the same banded `buildPayoutPsbt`.
- `assertPayoutFeeBandDomain` now rejects a non-integer or sub-1 `councilSize` before it crosses the WASM boundary, so the accepted domain equals the proven domain for every band input.

Spot-checks that hold up: the band asserts `floor <= implicitFee <= ceiling` with the ceiling extended only by `out0Len` (the pinned script) and the floor allowed to see `out1Len` (the VP-controlled commission script - safe because padding can't raise the floor and shortening only lowers it); domain validation runs before any expensive tx parsing; every WASM return is funnelled through `assertWasmBigint`; and the terms-vs-signing rate check in `runDepositorPresignFlow` catches params-version drift before the payout bound applies. CI green.

PR #2171, opened 2026-07-31
feat(vault): address the market detail page by token symbol

jeremy-babylonlabs:


services/vault/src/routes.ts
```diff
@@ -94,7 +97,27 @@ export function getReserveDetailSearch(reserveId: bigint, tab: LoanTab) {
...
+ * registry's symbol for the reserve's underlying address, lowercased, falling
+ * back to the on-chain id for addresses the registry does not know (testnet
+ * mocks).
+ *
+ * Deliberately not the indexer's `token.symbol`. Link building and the page's
+ * lookup both key off the registry, so a compromised indexer that rewrites a
+ * reserve's underlying can only make the link resolve to nothing, never to a
+ * different market — and the id/underlying pair it does resolve to is still
+ * proven against the chain before anything renders
+ * (`useVerifiedReserveIdentity`, audit F7).
```
jrwbabylonlab, line 111: what's audit F7?

jrwbabylonlab (approved):

services/vault/src/routes.ts
```diff
@@ -1,4 +1,7 @@
+import type { Address } from "viem";
+
 import type { LoanTab } from "@/applications/aave/constants";
+import { getRegisteredTokenByAddress } from "@/services/token/tokenService";
```
gbarkhatov, line 4: `routes.ts` stops being a leaf here. `tokenService` calls `getNetworkConfigBTC()` at **module scope** (`tokenService.ts:17`), which calls `getBabylonConfigState()` — and that *throws* if `configureBabylonConfig()` hasn't run (`config/network/runtime.ts:148-155`). All 13 `@/routes` importers now inherit that import-time requirement, `router.tsx` included. The app only survives because `main.tsx:4` imports `@/config/env` before anything else.

That coupling is why this PR had to patch an otherwise unrelated `LoanFlowOverlay.test.tsx`; it won't be the last file to need it.

Suggest inverting: put the slug helper next to the registry it reads (`services/token/`) and let *it* import `ROUTES`. The dependency stays one-directional and `routes.ts` stays string-only.

services/vault/src/routes.ts
```diff
@@ -10,8 +13,8 @@ export const ROUTES = {
   MARKETS: "/markets",
 } as const;
 
-/** Path segment of `/markets/:reserveId` — the reserve's on-chain id. */
-export const MARKET_RESERVE_PARAM = "reserveId";
+/** Path segment of `/markets/:market` — see {@link getMarketSlug}. */
```
gbarkhatov, line 16: The param rename leaves `parseReserveId`'s JSDoc below (lines 57-66) saying the opposite of the code. It still documents "the `/markets/:reserveId` segment" as a source it parses, and calls `/markets/usdc` a legacy link that "must hard-block" — that is the canonical URL as of this PR.

Per CLAUDE.md ("check for comments/docs that reference old behavior"), that block should now reference `?reserve=` only.

services/vault/src/routes.ts
```diff
@@ -94,7 +97,27 @@ export function getReserveDetailSearch(reserveId: bigint, tab: LoanTab) {
...
-  return `${ROUTES.MARKETS}/${reserveId.toString()}`;
+/**
+ * Slug naming a reserve in `/markets/:market`: the compile-time token
+ * registry's symbol for the reserve's underlying address, lowercased, falling
+ * back to the on-chain id for addresses the registry does not know (testnet
+ * mocks).
+ *
+ * Deliberately not the indexer's `token.symbol`. Link building and the page's
+ * lookup both key off the registry, so a compromised indexer that rewrites a
+ * reserve's underlying can only make the link resolve to nothing, never to a
```
gbarkhatov, line 108: This claims a stronger property than the code enforces.

Before this PR the reserve id came from the URL, so the indexer had *zero* input into which reserve opened — it could only supply an `underlying`, which `verifyReserveIdentity` then rejected on mismatch. Now the page selects whichever reserve's indexer-supplied `underlying` slugs to the URL segment, so selection is indexer-influenced. The chain check still prevents a *mislabeled* market from rendering, but by rewriting one reserve's underlying and omitting another, the indexer can steer `/markets/usdc` from one genuine reserve to a different genuine reserve whose chain-proven underlying also carries a registry `USDC` symbol.

Narrow in practice (one deployment sees one network's addresses), but the wording asserts an absolute. Suggest softening to something like "can only steer among reserves whose chain-proven underlying carries the same registry symbol" — or making resolution collision-proof (see the comment on `index.tsx`).

services/vault/src/routes.ts
```diff
@@ -94,7 +97,27 @@ export function getReserveDetailSearch(reserveId: bigint, tab: LoanTab) {
...
+ * different market — and the id/underlying pair it does resolve to is still
+ * proven against the chain before anything renders
+ * (`useVerifiedReserveIdentity`, audit F7).
+ */
+export function getMarketSlug(reserveId: bigint, underlying?: Address): string {
```
gbarkhatov, line 113: `underlying` shouldn't be optional — here or on `getMarketDataRoute` (line 121).

Every real caller has it: `showMarketInfo` is borrow-only (`!isRepay`, `AssetSelectionPanel.tsx:118`) and `assets` is only ever passed in repay mode (`Detail/index.tsx:136`), so the panel's `.find` always hits. The optional param can therefore only be exercised by a future caller that forgot to thread it — and its failure mode is a silently broken link (an id-shaped URL for a reserve the page resolves by symbol), not a type error. Making it required deletes the class outright; the registry-miss fallback still lives inside the function where it belongs.

Nit while here: the symbol is interpolated into the path unencoded (line 122). Unreachable with today's alphanumeric registry symbols, but `encodeURIComponent` is free.

services/vault/src/applications/aave/components/AssetSelectionPanel/AssetSelectionPanel.tsx
```diff
@@ -143,7 +143,9 @@ export function AssetSelectionPanel({
   });
 
   const handleMarketInfoClick = (reserveId: bigint) => {
-    navigate(getMarketDataRoute(reserveId));
+    const underlying = pricedReserves.find((r) => r.reserveId === reserveId)
+      ?.reserve.underlying;
+    navigate(getMarketDataRoute(reserveId, underlying));
```
gbarkhatov, line 148: The panel's only test for this navigation now passes vacuously. The `borrowableReserves` fixtures in `AssetSelectionPanel.test.tsx:29-53` carry `reserve: { hub, assetId }` with no `underlying`, so `getMarketSlug` short-circuits and the asserted `/markets/2` exercises only the id fallback — the symbol path this PR exists for is untested on the link-builder side.

Worse, that file's `tokenService` mock (line 84) omits `getRegisteredTokenByAddress`, so the first fixture that gains an `underlying` dies with `is not a function` rather than reporting a slug mismatch.

Adding `underlying` to a fixture plus the missing mock export would cover the real path and keep the two sides of the round-trip honest.

services/vault/src/components/pages/BorrowingMarketsData/index.tsx
```diff
@@ -146,17 +142,17 @@ export default function BorrowingMarketsData() {
...
     demoMarketData?.collateralFactor ?? splitParams?.CF ?? null;
 
-  // Resolve by the reserve's on-chain id, never by a token symbol from the URL:
-  // the symbol is indexer-supplied, so labelling this page from it lets a
-  // compromised indexer decide which market the user is reading (audit F7).
-  // Exactly one match or none — `.find` would take whichever duplicate the
-  // indexer ordered first.
+  // Resolve by the registry slug the link was built from, never by the
+  // indexer's own symbol — see `getMarketSlug` for why that distinction is the
+  // audit-F7 boundary. Exactly one match or none: `.find` would take whichever
```
gbarkhatov, line 147: Minor / house style: "the audit-F7 boundary" (and `(audit F7)` in the new `getMarketSlug` JSDoc) names the finding rather than the property. Stating the invariant directly — indexer-supplied symbols never steer which reserve resolves — reads the same to a reviewer and keeps its meaning once the audit doc is out of reach.

services/vault/src/components/pages/BorrowingMarketsData/index.tsx
```diff
@@ -146,17 +142,17 @@ export default function BorrowingMarketsData() {
...
-      target === null
-        ? []
-        : effectiveReserves.filter((r) => r.reserveId === target);
+    const slug = params[MARKET_PARAM]?.toLowerCase();
+    const matches = !slug
+      ? []
+      : effectiveReserves.filter(
+          (r) => getMarketSlug(r.reserveId, r.reserve.underlying) === slug,
+        );
     if (matches.length === 1) return matches[0];
```
gbarkhatov, line 156: Registry symbols aren't unique: `TOKEN_REGISTRY` maps three addresses to `USDC` (`tokenService.ts:73, 81, 89`) and two to `btcConfig.coinSymbol`. Two registered reserves sharing a symbol therefore give `matches.length === 2` → a permanent "market not found" for that URL, with nothing in the logs to distinguish an ambiguous registry from a bad link.

Unreachable today (one deployment sees one network's addresses) but unenforced anywhere, and the registry is hand-maintained. A `logger.warn` on `matches.length > 1` plus a test (two reserves both slugging to `usdc` → not found) would make the strict-match rule self-documenting — the rationale for `filter` over `.find` is in the comment above, but the collision path it guards has no coverage.

Also on line 154: `getMarketSlug` runs per reserve per memo pass, and `getRegisteredTokenByAddress` `logger.warn`s on any address failing `isAddress`. A single malformed underlying from the indexer becomes repeated log noise rather than one report.

services/vault/src/__tests__/routes.test.ts
```diff
@@ -8,7 +8,14 @@
 
 import { describe, expect, it } from "vitest";
 
-import { getReserveDetailRoute, parseReserveId } from "../routes";
+import {
+  getMarketDataRoute,
```
gbarkhatov, line 12: The file header (lines 1-7) now contradicts the file. It scopes this to "Reserve-detail route building and `?reserve=` parsing" and states the reserve detail is addressed "never by its token symbol" — in a file that, as of this hunk, tests a deliberately symbol-keyed builder.

Worth widening the header to cover both routes and to spell out why one stays id-keyed while the other is registry-symbol-keyed; that distinction is the whole point of the change and this is where a reader will look for it.

gbarkhatov (approved):
## Overview

Replaces `/markets/:reserveId` with `/markets/:market`, where the slug is the compile-time token registry's symbol for the reserve's underlying address (lowercased), falling back to the on-chain reserve id for addresses the registry doesn't know. Link building and the page's lookup both key off the same `getMarketSlug`, so the two can't drift — that's the right shape for this change, and leaving the `useVerifiedReserveIdentity` gate untouched means nothing renders on an unproven id/underlying pair.

Approving: everything below is a doc, test-coverage, or module-layering concern, not a correctness bug on the rendered path.

## Verified locally

- `pnpm --filter vault run lint` → 0 errors (103 pre-existing warnings).
- All 7 test files touching this change pass (53 tests).
- The full suite shows 50 failures in `useDepositFlow.test.tsx` / `usePayoutSigningState.test.tsx`. **Unrelated to this PR** — `forwardDepositApproval` exists in `packages/babylon-ts-sdk/src` (added in 7fbc82e9) but not in the built `dist/`. Stale SDK build.

## Should fix before merge

1. **The link builder's only test now passes vacuously**, and the same file's mock will throw the moment anyone fixes it — see the `AssetSelectionPanel.tsx` comment.
2. **Two docs now say the opposite of the code** — `parseReserveId`'s JSDoc and `routes.test.ts`'s file header.

## Worth fixing

3. `underlying?` optional builds silently broken links; every real caller has it.
4. `routes.ts` is no longer a leaf module — inverting the dependency keeps it one-directional.
5. Registry symbols aren't unique, so a slug collision is a permanent unlogged 404. Unreachable today, unenforced anywhere.

## Minor

- Existing `/markets/<id>` links break with no redirect. Low impact since the page is still behind `isMarketDetailPageEnabled` (396a85f8), but **the PR body is empty and should say so** — this is a URL-format change, and the next person reading the log won't infer it from the title.
- `getRegisteredTokenByAddress` warn-logs once per reserve per memo pass on a malformed underlying.
- The slug isn't URL-encoded.

jonybur (approved):
The registry-slug approach holds up: link building and the page's lookup both key off the same getMarketSlug, so the URL and the resolver cannot drift, and the useVerifiedReserveIdentity gate stays in front of anything that renders, so an id/underlying pair the chain doesn't prove still shows nothing. Case handling is symmetric - the slug generator lowercases the registry symbol, and the resolver lowercases the URL param before comparing - so the match is case-insensitive without a hand-maintained normalisation step. Ambiguous matches (matches.length !== 1) fall through to reserveNotFound rather than picking whichever reserve the indexer ordered first, which is the right failure mode. CI is green and the previously-raised doc drift and slug-collision notes are already tracked on gbarkhatov's approval.

PR #2173, opened 2026-08-01
feat(ts-sdk, vault): resolve participant BTC keys by epoch (RFC-006)

### LINKED ISSUE #2150: RFC-006: resolve participant operation keys by epoch in the vault FE
## Context

[RFC-006](https://github.com/babylonlabs-io/btc-vault/blob/main/docs/rfc/rfc-006-operator-admin-operation-key-separation.md) lets an operator (VP / AVK / UC) **rotate its BTC operation key** without re-registering, via an append-only key history bound per-vault by a frozen **epoch**. The daemon (`vaultd`) already resolves participant keys by that epoch (#2304); `getVaultProviderBTCKey` (the registration key) has **zero non-test call sites** in `vaultd`.

The vault FE is the lagging consumer: it bakes participant BTC keys into the pre-pegin Bitcoin lock, the challenger set, and its VP-auth handshake, and it currently reads the **fixed registration / roster keys**. On devnet, a VP rotated with `setOperationBtcKey` and new pegins were rejected — the FE-built scripts used the old key while the protocol expected the rotated one.

**Two protocol breaking changes drive this work:**
- **Key resolution** — participant keys move from the fixed registration key to epoch-resolved operation keys (the devnet pegin rejection above).
- **Payout validation** — [btc-vault#2440](https://github.com/babylonlabs-io/btc-vault/pull/2440), **scheduled for v0.7**: RFC-006 lets a VP/VK set a custom payout script, so the FE's payout tamper-check must resolve the payout script (and commission) at the vault's frozen epoch instead of deriving it locally via BIP-86. See Stage 6.

**Backwards compatible:** before the first rotation on a network, epoch-resolved keys are byte-identical to what the FE reads today, so the key-resolution change is a no-op until an operator rotates — but it must be in place **before** any rotation. The payout change is the exception: BIP-86-derived validation breaks the moment a VP/VK sets a custom payout script, independent of key rotation.

## What we're changing

The whole change is one idea: **stop reading the fixed registration/roster keys; resolve each participant's operation key by epoch, the way the contract + daemon do.**

| # | Change | Why | Priority |
|---|---|---|---|
| 1 | **Plumbing** — sync contract ABIs + one epoch-aware key resolver | Nothing works without the new getters/struct | Foundational |
| 2 | **VP-auth pins → current key** (`getCurrentOperationBtcKey`) | Server identity is per-operator; today's v0 pin breaks auth for **all** of a VP's vaults on its first rotation | **Highest** |
| 3 | **New-pegin build → current keys, then verify-after-mine** | Build the lock with current keys; after the vault exists, read back its frozen epoch and fail closed if drifted | High |
| 4 | **Existing-vault build → frozen-epoch keys** (payout / refund / challenger set) | An old vault must resolve the keys it was created under, not the newest | High |

Buckets 2–4 are all "resolve at an epoch" — the difference is only *which* epoch: current (new/auth) vs the vault's frozen epoch (existing). Wrapped by a feature flag + fail-closed capability check, plus two-epoch rotation tests.

## Work checklist

**Stage 0 — ABI sync** (regenerate from aave-v4 forge `out/*.json`, byte-exact)
- [ ] `BTCVaultRegistry.abi.ts`: extend `getBtcVaultProtocolInfo` (+`vpKeyEpoch`, `appKeeperKeyEpoch`, `ucKeyEpoch`, and ensure **frozen commission BPS** + **`appVaultKeepersVersion`** decode); add `getCurrentOperationBtcKey`, `getOperationBtcKeyAtEpoch`, `vpKeyEpochCurrent`, `getPayoutScriptAtEpoch`
- [ ] `ApplicationRegistry.abi.ts`: add `getOperationBtcKeyAtEpochOrGenesis`, `appKeeperKeyEpochCurrent`, **`getPayoutScriptAtEpoch(app, keeperAdmin, epoch)`** (per-keeper); confirm `getVaultKeepersByVersion` present
- [ ] `ProtocolParams.abi.ts`: add `getOperationBtcKeyAtEpochOrGenesis`, `ucKeyEpochCurrent`

**Stage 1 — One epoch-aware resolver** (call on-chain getters; do **not** re-implement the epoch search off-chain; treat a zero-hash result as a hard error)
- [ ] `getVaultFrozenEpochs(vaultId)` + frozen membership versions
- [ ] Per-role resolution: **VP `getOperationBtcKeyAtEpoch` (no genesis)**, **VK/UC `getOperationBtcKeyAtEpochOrGenesis`** with the roster genesis from the vault's frozen membership version

**Stage 2 — New-pegin build** (`validateOnChainParticipantKeys.ts:48/54/58`, `PeginManager`)
- [ ] Resolve VP + VK + UC at the **current** epoch/version
- [ ] After the pegin-request mines: read back frozen epochs/versions, re-resolve **all** participant keys, assert match, **fail closed before BTC broadcast**

**Stage 3 — Existing-vault build**
- [ ] Payout `vaultPayoutSignatureService.ts:113/182/194` — VP + VK + UC at the vault's **frozen** epoch + membership version
- [ ] Refund `vaultRefundService.ts:400/403/407` — same

**Stage 4 — VP-auth pins → current key (highest priority)**
- [ ] `ensureAuthenticatedVpClient.ts:93`, `useDepositFlow.ts:825`, `ResumeDepositContent.tsx:333`: swap pin source from `getVaultProviderBTCKey` → **`getCurrentOperationBtcKey(vp)`**. The server identity is per-operator; not doing this breaks auth for **all** of a VP's vaults on its first rotation.

**Stage 5 — Challenger-set derivation** (`signDepositorGraph.ts`)
- [ ] `deriveLocalChallengers` / `assertChallengerSetMatchesExpected` compare against **epoch-resolved** roster keys per role, matching `PeginSignatureLogic._buildExpectedSignatureKeys` and `vaultd` `challenger.rs` byte-for-byte

**Stage 6 — Payout tamper-check ([btc-vault#2440](https://github.com/babylonlabs-io/btc-vault/pull/2440), v0.7)**

RFC-006 lets a VP/VK set a custom payout script, so **stop deriving payout addresses locally via BIP-86** — resolve them from chain at the vault's frozen epoch. Flow per the protocol team (btc-vault#2440):

Read the vault snapshot — `getBtcVaultBasicInfo(vaultId)` → `vaultProvider`, `applicationEntryPoint`; `getBtcVaultProtocolInfo(vaultId)` → `vpKeyEpoch`, `appKeeperKeyEpoch`, `appVaultKeepersVersion`, frozen commission BPS. Then:
- [ ] VP bonded key: `getOperationBtcKeyAtEpoch(vaultProvider, vpKeyEpoch)` — use it to identify the VP entry in the presign response
- [ ] VP payout script: `BTCVaultRegistry.getPayoutScriptAtEpoch(vaultProvider, vpKeyEpoch)`
- [ ] AVK roster: `getVaultKeepersByVersion(applicationEntryPoint, appVaultKeepersVersion)` → `(keeperAdmin, rosterBtcKey)`; per keeper resolve `getOperationBtcKeyAtEpochOrGenesis(applicationEntryPoint, keeperAdmin, appKeeperKeyEpoch, rosterBtcKey)` (keep the bonded-key → keeperAdmin mapping) and `ApplicationRegistry.getPayoutScriptAtEpoch(applicationEntryPoint, keeperAdmin, appKeeperKeyEpoch)`

Validate the presigned `PayoutTx`s from `vaultProvider_requestDepositorPresignTransactions`, routing by `claimer_pubkey` (VP / depositor / AVK):
- [ ] **VP tx:** Output 0 == depositor payout script; Output 1 == VP payout script @ `vpKeyEpoch`; Output 1 **value** == commission derived from the frozen commission BPS
- [ ] **Each AVK tx:** `claimer_pubkey` == bonded AVK key; Output 0 == that AVK's payout script @ `appKeeperKeyEpoch`
- [ ] **Remove** the local BIP-86 payout-address derivation (today's pin: `payout.ts` VK branch derives BIP-86 from `claimerBtcPubkey`; the keeper query reads only `btcPubKey` — the registered script is dropped)
- [ ] **Legacy graphs:** vaults built by pre-v0.7 vaultd pay BIP-86 destinations regardless of the registry — decide dual-accept (registered-script OR legacy BIP-86, keyed by graph age) vs environment reset before v0.7; a hard switch strands their resume
- [ ] **#2105 knock-on:** the payout fee band takes its trusted `out0Len`/`out1Len` from whichever script pins are active — this stage's PR switches those length sources (one line each) when the pins move to registered scripts

**Stage 7 — Rollout guard + tests**
- [ ] `NEXT_PUBLIC_FF_OP_KEY_ROTATION` in `featureFlags.ts`; **OFF = today's path (zero new calls)**, **ON = epoch-aware + fail-closed startup capability check** (never silently fall back to the legacy key)
- [ ] Two-epoch rotation fixture: old vault keeps old keys on resume/payout/refund; new pegin uses new keys; challenger set matches the daemon; zero-hash rejected

## Rollout — order matters

**Build order:** land Stage 0–1 (additive, no behavior change), then Stages 2–6 **behind the flag OFF** (a no-op, safe to merge anytime).

**Per network** (flag lets each flip independently):
1. Upgrade the contract → FE still legacy, keys still correct (no one's rotated). ✅
2. Flip the flag ON → FE goes epoch-aware; capability check hard-fails if the contract lacks the methods. ✅
3. Only now may operators rotate.

**Hard rule:** flag ON must precede the **first** rotation on a network. Sequence: **devnet** (already RFC-006, only place rotation is testable) → **testnet** → **mainnet**, each gated on its own contract upgrade. This matches the deployment process agreed in-thread: web accommodates both old + new (old by default) → contract upgrade → web drops old.

## Confirm before coding (cross-team)

- [ ] **Rotation race / read-after-write ordering** (RFC open-q1) — confirm "build-current → read-back-assert-all → fail-closed-before-broadcast", and whether a mismatch reverts the request or surfaces downstream.
- [ ] **Indexer-hint epoch context** — the FE cross-checks on-chain keys against indexer hints; ensure those checks still pass once on-chain reads return rotated keys.

*(Resolved: payout-script scope — spec'd by the protocol team in btc-vault#2440, now Stage 6.)*

## Out of scope

ETH operation-key rotation (RFC open-q8, treat as fixed for the FE); the admin key + setters (shipped as `vaultd` CLIs); key-retention guard + status RPC (remaining in #2308). Membership-version handling is unchanged — only the per-member **key** becomes epoch-resolved.

gbarkhatov:
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2150

**Follow-up:** #2182 — remove the dual-accept BIP-86 fallback once every network is past btc-vault#2440.
**Related:** #2181 — a VP token rejection is unrecoverable in-app (Retry does nothing, only a reload works). Pre-existing, but key rotation makes it far likelier to be hit; found during this testing.

## The problem in one sentence

Operators can now change their Bitcoin key without re-registering, and our frontend still reads the **old, fixed** key — so the Bitcoin locks we build no longer match what the protocol expects.

## Background

Every participant in a vault — the Vault Provider (VP), the Application Vault Keepers (VK), the Universal Challengers (UC) — has a Bitcoin public key. That key is baked into the taproot scripts of the Bitcoin lock we build at deposit time, and into the set of challengers we pre-sign payout / no-payout transactions for.

Until RFC-006 each operator had exactly **one** BTC key, fixed at registration forever. RFC-006 splits that into a cold **ETH admin key** (identity and authority) and a hot, rotatable **BTC operation key** (the thing that actually goes into scripts and signs). Rotation is append-only: `setOperationBtcKey` adds a new version and old versions stay readable forever. So that existing vaults don't break when someone rotates, each vault **freezes an epoch number** when it is created, and everyone answers "which key did this vault bond?" by asking the contract for the key at that frozen epoch.

That is exactly what happened on devnet: a VP's operation key was rotated, our frontend kept building locks with the registration key, and every new peg-in was rejected. The rotation was reverted and is waiting on this PR.

**RFC-006 contains two independent breaking changes**, and separating them is the single most important thing to understand about this PR:

| | **A — key rotation** | **B — registered payout scripts** |
|---|---|---|
| What breaks | scripts built with the pre-rotation key | payouts pinned to a locally-derived BIP-86 address |
| Triggered by | `setOperationBtcKey` | `setPayoutScript` |
| Contract side | shipped | shipped |
| Daemon side | [btc-vault#2304](https://github.com/babylonlabs-io/btc-vault/issues/2304), **in v0.7.0** | [btc-vault#2440](https://github.com/babylonlabs-io/btc-vault/pull/2440), **in no release — `main` only** |

They have different triggers and different deployment timelines. An earlier revision shipped them behind two feature flags; devnet testing showed neither was needed — see *No feature flags* below.

## What changed

| Area | Change |
|---|---|
| New peg-ins | Build with each participant's **current** operation key, then after registration mines, re-read the vault's frozen epochs, re-resolve, and fail closed **before the BTC broadcast** if anything drifted |
| Existing vaults (resume / payout / refund) | Resolve keys at the epochs **that vault** froze, so an old vault stays signable after a later rotation |
| VP authentication | Pin the VP's **current** operation key — that is what its server signs its auth tokens with |
| Payout transactions | Take keeper payout destinations and the VP commission destination from the registry instead of deriving BIP-86 locally, accepting the BIP-86 default alongside it so graphs built before btc-vault#2440 stay signable |
| Rollout | No flags and no ordering constraint — the client is correct against a vaultd fleet on either side of #2440 |

## No feature flags

An earlier revision of this PR shipped two: `NEXT_PUBLIC_FF_ENABLE_OPERATION_KEY_ROTATION` and `NEXT_PUBLIC_FF_ENABLE_REGISTERED_PAYOUT_SCRIPTS`. Devnet testing removed the case for both.

**The rotation flag guarded nothing.** Its purpose was to protect a registry that predates RFC-006 — and no network we ship to is in that state. devnet and testnet both have all four getters; mainnet will be a fresh RFC-006 deploy. On devnet, flag-*off* was already the broken setting. Its fail-closed capability probe went with it (see *Known issues*, item 1 in the previous revision): the classifier could never return `false`, so the entire negative path was unreachable.

**The payout flag's off state was the dangerous one.** With dual-accept in place, flag-on is never worse than flag-off and strictly better in one case: once a keeper registers a custom payout script, flag-off fails while flag-on passes. A flag whose innocuous-looking default is the unsafe setting is a trap.

Both are gone, along with the legacy BIP-86-only branch in `assertPayoutOutputLayout` and the `if (vpCommissionScriptPubKey !== undefined)` skip — which was the last configuration in which the VP commission destination went unvalidated. `vkClaimerPayoutScriptPubKeys` and `vpCommissionScriptPubKey` are now required on `PayoutParams`.

### 1. The ABI: extend the existing entry, or isolate it?

The obvious approach is to add the three new epoch fields to our existing `getBtcVaultProtocolInfo` ABI entry. **This is unsafe.** That function returns a *positional* tuple. If we declare 16 fields but point at a network whose registry has not been upgraded yet, the decode goes wrong. Worse, `getProtocolInfoBatch` and `getVaultData` read it through `multicall({ allowFailure: false })`, so one bad decode fails the entire batch — which would break vault status, resume, payout and refund **app-wide**. A feature flag would **not** have saved us here either, because the ABI shape is not flag-gated.

**Chosen:** put the extended 16-field shape in its own ABI file (`BTCVaultRegistryKeyEpochs.abi.ts`), read only on the epoch-aware path. The shared entry keeps the 13-field shape that decodes against every deployed registry. A test pins that the 13 fields are an exact prefix of the 16, so the two hand-maintained copies cannot drift apart.

**One finding that changed the design.** A not-yet-upgraded registry does **not** fail loudly under the extended ABI. Because the struct contains `bytes`, a real vault's response carries enough trailing data to satisfy the three extra fields — so viem returns garbage as epochs with no error. It only throws for a **nonexistent** vault, whose byte fields are empty. That asymmetry is why the earlier revision probed with a *nonexistent* vault id. The probe is gone now — it guarded a registry state no network we ship to is in, and its classifier could never actually match — but `assertEpochInRange` still bounds a garbage decode, and the ABI-decode tests pin both directions.

### 2. Where does the switch live?

The SDK has no feature flags and we kept it that way. Epoch resolution is driven by **passing the reader in**, and the payout destinations are ordinary required params. With the app-side flags now removed too, there is exactly one path through this code on every network — which is the main reason the test matrix shrank rather than grew.

### 3. Indexer hints: throw, warn, or accept either?

We cross-check the keys the indexer gives us against the chain. Once a VP rotates, the indexer will still be serving the old registration key for a while, so a strict equality check would throw and **block every deposit for that provider**.

**Chosen:** accept **either** the registration keys or the operation keys. The hint never influences which key we actually use — that is always read from chain. The check still catches a wrong VP address, a wrong application entry point, or a stale roster version, which is its real value. Each role is compared as a whole set, and there is an extra check that rejects a *mixed* answer, which means the indexer is serving a half-applied view.

### 4. Which epoch — current or frozen?

Both, depending on the situation. A peg-in being built right now uses **current** keys, because it has not frozen an epoch yet. Everything about an existing vault uses **that vault's frozen** epoch.

VP authentication deliberately uses **current**, because the VP's server signs its auth tokens with its live key — it is a per-operator server identity, not a per-vault binding. Confirmed against `vaultd`: its `AuthService` is bound to `wallet_expected_pubkey`, which the config docs define as the current operation key, and `getVaultProviderBTCKey` has zero live call sites in the daemon.

### 5. Payout commission: keep the cap, don't switch to exact equality

The issue suggested asserting the commission output value "matches" the frozen commission rate. We already assert it is **at most** the entitled amount. An upper bound is the right posture for a depositor — a VP taking *less* than it is owed harms nobody, and strict equality would reject that. Kept the cap and added the destination check on top.

## Things rotation newly makes possible

None of these were reachable before, and they shaped several decisions:

- **Sort order is no longer stable.** Keys are sorted before going into scripts, and a rotated key sorts somewhere else. So the code carries `(admin address → key)` **pairs** as the source of truth and derives the sorted arrays from them, never the other way round. The test fixture deliberately includes a keeper whose rotated key sorts *before* its original one.
- **Two operators can end up with the same key.** Nothing on-chain stops two keepers rotating onto the same key, or a keeper rotating onto the VP's key. That would silently collapse the participant set, so it is rejected at resolve time, before any script is built.
- **The registry now returns operator-chosen bytes.** Pinning a payout output to it without checking would let a bad script get the depositor to pre-sign an unspendable payout, so there is a new check that it is a standard, spendable output type. The accepted set (P2TR/P2WPKH/P2WSH/P2SH/P2PKH) matches `vaultd`'s registration-time gate exactly.
- **VP auth can break mid-session.** A provider that rotates while a session is open invalidates its outstanding tokens. Devnet testing showed this is *not* recoverable in-app: the client's re-auth path matches on a `data.kind` the daemon never sends, so **Retry does nothing and only a page reload recovers**. Tracked separately in #2181 — it is pre-existing, not introduced here, but rotation makes it far likelier to be hit.

## Devnet verification

Two rounds, on either side of the devnet fleet upgrade. Full logs and per-test baselines are in `markdown/rfc006/`.

### Round 1 — before the rotation (daemon `02d7c353`, pre-#2440)

Four full deposits through to activation, with temporary instrumentation capturing the resolved key material at each decision point (since removed).

| Run | Rotation | Payout | Result |
|---|---|---|---|
| 1 | off | off | Full deposit to active. **Zero** new contract-call selectors. Baseline. |
| 2 | on | on | Build, read-after-mine verify and BTC broadcast all passed. **Failed at payout signing** — see below. |
| 3 | on | off | Resume of run 2's vault. Payout signing proceeded. |
| 4 | on | off | Full deposit to active. |

Across runs 1 and 4 — one on the legacy path, one resolving through frozen epochs `12 / 16 / 12` — the resolved key material was **byte-identical**: VP `1db08a11…0008e`, the same 4 keepers and 4 challengers in the same sort order, frozen versions `1 / 1 / 2 / 2`, commission `10` bps. That is the "no-op until someone rotates" claim demonstrated end to end against a live chain rather than argued from source.

**Why run 2 failed.** The devnet VP has called `setPayoutScript` at least three times:

```
getPayoutScriptAtEpoch(VP, 0)  → 0x5120f436aa9c…   P2TR (BIP-86 backfill)
getPayoutScriptAtEpoch(VP, 1)  → 0x00142bbccc13…   P2WPKH
getPayoutScriptAtEpoch(VP, 8)  → 0x001438481735…   P2WPKH
getPayoutScriptAtEpoch(VP, 12) → 0x00142bbccc13…   P2WPKH
```

That daemon predates #2440, so it built output 1 as P2TR of the VP's operation key and ignored the registry. This **falsified a claim in an earlier revision of this description** — that the payout change is a no-op because the contract backfills BIP-86. That holds only for an operator who never called `setPayoutScript`.

### Round 2 — after the rotation and the fleet upgrade

devops rotated the devnet VP operation key at 05:31 UTC on 2026-08-04, then upgraded the fleet to `bf3e6719` (btc-vault#2488, which contains #2440) at 13:46 UTC. The VP's key history:

```
epoch  0–9   1db08a1171aa…0008e   original / registration key
epoch 10     8d93ed5c7b7c…2421f   first rotation (this is what broke pegins and prompted #2150)
epoch 11–12  1db08a1171aa…0008e   the "revert" — an append, not a rollback
epoch 13+    4a9c87a094b9…9835b   current
```

An epoch is **not** a rotation count: one per-operator counter also advances on `setPayoutScript`, which is why a vault created before any rotation froze `vpKeyEpoch 12` rather than 0.

| Test | Vault | Frozen epochs | Outcome |
|---|---|---|---|
| **A** — pre-rotation vault, post-rotation signing | `0xbcf3061d…85b3` | 12 / 16 / 12 | **Passed frontend-side.** Resolved to the superseded key `1db08a11…` and the VP accepted the signatures. Then blocked at the VP→keeper ACK by a daemon-side gap (below) — not a frontend failure |
| **B** — post-rotation vault | `0xc740c567…85f5` | **13** / 16 / 12 | **Passed.** Built with `4a9c87a0…` and was ingested. A control deposit from a client without this PR (`6e29c7bc…`) was `IngestionRejected` under the same conditions. **This is the bug being fixed, reproduced and then fixed on a live chain** |
| **C** — auth across the rotation | both cohorts | — | **Passed.** The pin follows `getCurrentOperationBtcKey`, so both pre- and post-rotation vaults authenticate against the rotated key |
| **D** — resume drift guard | `0x25991e0f…deaf` | 12 / 16 / 12 | **Premise was wrong.** Epochs freeze at registration, so a later rotation cannot create drift for an already-registered vault. The guard correctly resolved at the frozen epoch. The abort branch remains unexercised on-chain |

### Round 2b — registered payout scripts, after #2440 landed

| Subject | Payout resolution | `outs[1]` the daemon built | Outcome |
|---|---|---|---|
| Vault B — graph built ~3h **before** the upgrade | registered script expected | `51201d5e2825b587…` = `bip86(4a9c87a0…)` | **Rejected** — this is what proved dual-accept is mandatory |
| Vault B — same graph, legacy derivation | BIP-86 derived locally | same | Accepted; vault completed |
| Fresh deposit `0xa5b3e51b…190d` — graph built **after** the upgrade | registered script expected | `00142bbccc132d605bd10a4b0e14275b260c46f00d2f` | **Accepted and activated** |

The third row is the confirmation that **#2440 is live** on devnet — nothing else observable from outside distinguishes the two daemon builds, because `payout.ts` on `main` only length-checks `outs[1]` for the VP claimer, so a 22-byte P2WPKH and a 34-byte P2TR both pass there.

The first two rows together are why `acceptedPayoutScriptHexes` exists. Both runs returned **identical payout txids**, which independently confirms the protocol team's statement that a graph is never rebuilt once set.

### Round 3 — flags removed, review fixes applied (`bc5b2679`)

A full deposit on the final build: no feature flags, single unconditional epoch-aware path, and the review fixes below already in place. **Completed successfully.**

This closes the last open item in the plan. It matters for two reasons:

- It is the first end-to-end run where nothing is conditional. Rounds 1 and 2 exercised the flag-on path; this confirms the unconditional path behaves identically now that the flag-off branch no longer exists to fall back to.
- It exercises the review fixes on the happy path. All three are unreachable without drift, a rotation mid-deposit, or a split indexer, so the expectation was **no observable change** — and that is what happened. A regression here would have meant one of the fixes altered the normal path, which none of them should.

### Deployment facts established while testing

These correct #2150 and earlier revisions of this description:

- **#2440 was in no release** at the time of writing. It merged to `main` on 2026-07-29; the `v0.7.0` tag is 2026-08-03 and does not contain it. Any plan gated on "v0.7 vaultd" for the payout half was wrong.
- `tbv-networks/devnet/network_info.json` pins `btcVaultRef: 666112c8`, which is stale — that is where `release/v0.7.x` forked from `main`. The daemon's live `health_check` `commit` field is authoritative.
- **`vaultd` does not validate peg-in ingestion via `getCurrentOperationBtcKey`**, contrary to #2150. Ingestion resolves at the vault's **frozen** epoch. The observed rejection is still fully explained — the contract freezes `vpKeyEpoch` at `submitPeginRequest`, so a frontend building with the stale registration key produced scripts that did not match — but the stated mechanism was wrong, and it is the justification for the build-current-then-verify design.
- **testnet** is on the RFC-006 contracts, is un-rotated, and `getPayoutScriptAtEpoch` returns exactly the BIP-86 backfill (verified: `bip86(90e4d34b…7133) == 0x512029ce1605…7941`). The RFC-006 path is provably a no-op there today.

### Still not exercised on-chain

| Gap | Why it is acceptable |
|---|---|
| A **keeper** rotating, and the sort reordering that follows | Only the VP has ever rotated on devnet. The pairs-are-source-of-truth invariant only bites here; fixture-tested, with a keeper whose rotated key deliberately sorts before its original |
| The read-after-mine **abort** branch | It ran and passed on every run; the abort itself never fired. Unit-tested |
| The resume drift-guard **abort** branch | Same — test D's premise turned out to be unreachable by construction |
| Key-collision and off-curve rejection | Unit only |
| Refund | `tRefund` on devnet is 2016 blocks (~14 days on signet). It calls the same `resolveParticipantKeysAtEpochs` with the same frozen epochs as payout |

## What still needs to be done

- Nothing. The three known issues below are tracked as #2185, #2186 and #2187; none blocks merge.

One standing operational ask: **a keeper must not call `setPayoutScript` on a network whose vaultd fleet predates #2440 until this PR has shipped there.** The VP's own custom script is survivable, because `payout.ts` on `main` only length-checks `outs[1]`; a keeper's is not, because keeper destinations are compared exactly.

## Addressed in review (`bc5b2679`)

Review by @jrwbabylonlab found one live defect that the description had recorded only as a nested-condition smell, plus five documentation claims that were factually wrong. All are fixed; neither critical-path file changed.

**The resume path could broadcast a Pre-PegIn the key guard had just refused.** `verifyResumeParticipantKeys` threw `RegisteredVaultVersionMismatchError`, so the catch dropped the pending record. On retry that record — and with it the build-time key stamp — was gone, so `localUnsignedTxHex` was undefined, the transaction fell back to the indexer's copy, the whole guarded block was skipped, and the `prePeginTxHash` check passed because it *is* the registered transaction. `useDepositFlow.ts` had the identical bypass.

The hash proves the transaction was not substituted. It does not prove the vault's frozen epochs still resolve to the keys inside its scripts — and on drift they do not, because the registered hash commits to a transaction built with pre-rotation keys while the vault froze the post-rotation epoch. So the retry converted a clean no-loss abort into BTC locked until the refund timelock, while the error message still claimed "The Pre-PegIn was not broadcast".

Key drift now throws `ParticipantKeyDriftError`, a **sibling** of `RegisteredVaultVersionMismatchError` and deliberately not a subclass — subclassing would keep satisfying the type guard and preserve the bug. Cleanup now differs by design: a version mismatch still drops the record, key drift keeps it. `mapDepositError` gained its own copy for it, since rendering key drift as "Protocol parameters changed" invited exactly the retry that cannot help.

Also fixed:

- **Known issue 1** — `resolveVaultProviderBtcPubkey` now accepts the indexer hint against the registration key *or* `getCurrentOperationBtcKey`, matching the deposit path. The extra read is a fallback, not an unconditional second RPC.
- **Known issue 2** — the resume key guard moved out of the version block; its only precondition is the stamp.
- **Half-applied indexer telemetry** — `onIndexerHintsInconsistent` fires before the throw, wired to `logger.error`. The block clears only when the indexer converges, so "Refresh and try again" cannot help and it needs to be observable.
- **Five wrong doc claims**, checks unchanged: `getPayoutScriptAtEpoch` is *not* a no-op (a payout script is mandatory at registration and stored as v1, so it returns the operator's own script from epoch one — which is why dual-accept is required, not merely convenient); `standardPayoutScript` is *not* parity with `vaultd`, which uses registry scripts verbatim, so it is now justified on its own terms as a pre-signing bound that is also load-bearing for the fee band; the operation-key collision claim was overstated, since intra-role uniqueness is enforced on-chain.
- **Stale prose from the draft** — seven references to the removed capability probe and to "when the flag is off", across five files. Flagged by Greptile at one site; the sweep found the rest.

## Known issues, not yet fixed

These were found reviewing the change. All three are now tracked separately, so nothing here depends on this PR staying open — the list is kept for reviewers who want the reasoning inline. **None blocks merge:** one is test hardening, one harms only the VP, and one is a diagnosis defect on a path that still fails closed.

Two issues from the previous revision are gone by deletion rather than by fix: the capability probe's unreachable classifier, and the never-rendered `operationKeyRotationUnsupported.title` — both removed with the flags.

1. **Test coverage is thin app-side** — #2185. `vpAuthPinnedPubkey.ts` and `verifyResumeParticipantKeys.ts` still have no tests of their own. Separately, the challenger-set derivation is correct only *transitively* — `signDepositorGraph.ts` has no contract imports and receives keys through the context object — so a future refactor reintroducing a registration-key read in `prepareSigningContext` would break it with a green suite.

2. **No absolute dust floor on the commission value** — #2186. #2169 added a *rate* floor at the trust boundary — `max(offchainParams.minVpCommissionBps, MIN_REALIZABLE_VP_COMMISSION_BPS)`, the latter being `1` — so the bps is now bounded on both sides. But `vaultd` rejects a commission below the **546-sat dust limit**, and 1 bps of a small peg-in is well under that, so the sat-level gap remains. Harms only the VP, but it is an asymmetry with the protocol.

3. **On resume, a spent-input error preempts the key-drift diagnosis** — #2187. `useVaultActions.ts` calls `assertUtxosAvailable` (line 263) unconditionally, well before `verifyResumeParticipantKeys` (line 345). If the pending Pre-PegIn's inputs were spent by an unrelated transaction, the resume throws "UTXOs unavailable" and the drift check never runs — so in exactly the scenario the guard exists for, the user is told the wrong thing and the key drift is never surfaced or recorded. It still fails closed and never broadcasts, so this is a diagnosis defect rather than a safety one. It also makes the drift guard awkward to test: any deposit created between staging a pending peg-in and resuming it can consume that peg-in's inputs and mask the result.

## Rollout — no ordering constraint

Both feature flags were removed after devnet testing. There is no network where the flag-off path was correct and the flag-on path was wrong:

- **devnet** — RFC-006 contracts, VP rotated to epoch 13. Flag-off was already *broken* here: a deposit built with the registration key is rejected at ingestion because the vault freezes epoch 13.
- **testnet** — registry `0xE976a4f2…` has all four RFC-006 getters, is un-rotated, and `getPayoutScriptAtEpoch` returns exactly the BIP-86 backfill (verified: `bip86(90e4d34b…7133) == 0x512029ce1605…7941`). The RFC-006 path is provably a no-op there.
- **mainnet** — not deployed; it will be a fresh RFC-006 deploy.

With dual-accept for legacy payout graphs (below), the client is correct against a vaultd fleet on **either** side of #2440. So per network: land this, then upgrade the fleet and call `setOperationBtcKey` / `setPayoutScript` in any order.

The one operational caution left: a **keeper** calling `setPayoutScript` breaks every client that predates this PR, because the pre-RFC-006 code derives keeper destinations with BIP-86 and compares them exactly. The VP's own custom script does not, since `payout.ts` on `main` only length-checks `outs[1]`.

## Legacy payout graphs — deliberate dual-accept

A payout graph is built once at BaBe Setup and is **never rebuilt**, so a vault whose graph predates its network's #2440 upgrade pays BIP-86 for the rest of its life. Accepting only the registered script makes every such vault permanently unsignable, with the BTC already committed — observed on devnet, where vault `0xc740c567…` failed with `outs[1]` = `51201d5e2825b587…`, exactly `bip86(4a9c87a0…)`.

`acceptedPayoutScriptHexes` therefore accepts the registered script **or** the BIP-86 default of the same bonded key. Both candidates are derived from on-chain state, never from the VP's response, so the only substitution permitted is between two destinations the same operator already controls — and the commission value cap still bounds depositor exposure independently. Removal conditions are documented on the function and tracked in #2182.

## Rebase onto #2169 — what was removed from it, and why

[#2169](https://github.com/babylonlabs-io/babylon-toolkit/pull/2169) (payout implicit-fee band) landed on `main` while this was in review and changed the same function, `assertPayoutOutputLayout`. The rebase keeps both changes, but **one of #2169's checks became unreachable** and was removed. Flagging it here because it is someone else's just-landed invariant.

#2169 states its own premise in the `assertPayoutFeeBand` module doc:

> `out1Len` is measured from the **deliberately-unpinned** VP commission output and is **UNTRUSTED**: it feeds only the floor […] It must never widen the ceiling — a VP-controlled length there would buy burnable headroom.

This PR pins `outs[1]` to the operator's registered scriptPubKey, so that premise no longer holds. Only part of what it justified becomes dead:

| Element of #2169 | Kept? | Reason |
|---|---|---|
| `out1Len`'s `[1, 128]` cap | **removed** | `assertStandardPayoutScript` bounds the registered script to P2TR/P2WPKH/P2WSH/P2PKH/P2SH — 34 bytes at most. A 128-byte cap on `outs[1]` can no longer fire under any input |
| `out1Len` feeding the fee **floor** | kept | Still correct, and now read from a pinned value rather than a VP-controlled one — strictly more trustworthy than when #2169 shipped |
| `out1Len` excluded from the **ceiling** | kept | No longer strictly required once the output is pinned, but the separation is kept deliberately so the bound does not silently depend on that pin |
| `out0Len`'s cap and ceiling-excess term | kept — **still live** | vp-claimer and depositor-as-claimer `outs[0]` is `registeredPayoutScriptPubKey`, which we only hex-check, so it genuinely reaches 128 bytes. Only the vk-claimer's `outs[0]` is standard-gated |

Two of #2169's integration tests in `payout.test.ts` were deleted, because their premise — a paddable `outs[1]` — no longer exists:

- `gives a padded commission script no extra ceiling headroom`
- `rejects a commission scriptPubKey outside the contract registration cap`

**No coverage is lost.** The first one's property is already pinned at unit level by `does not widen the ceiling for the commission script length` in `assertPayoutFeeBand.test.ts`, which calls `assertPayoutFeeInBand` directly and so is unaffected by the pin. That test is kept; only its name changed, since it called the output "unpinned". The second test asserted a bound that is now enforced earlier and more strictly by the standard-script gate.

`assertPayoutFeeBand.ts`'s module doc and the inline comment above `scriptExcess` were also corrected — both described `outs[1]` as unpinned and VP-controlled.

One knock-on inside `assertPayoutOutputLayout`: #2169 computed `out0Len` from `expectedOut0Script.length`, which was sound when exactly one script was expected. Dual-accept means the accepted set can hold two candidates, so it now reads `payoutTx.outs[0].script.length` — the output actually present, which the equality check above has already pinned to one of the accepted values. Same value, and it mirrors how `out1Len` was already computed.

## Testing

- ts-sdk: 1448 tests across 83 files, all passing.
- vault: 3105 tests across 274 files, all passing.
- Typecheck, lint, and `pnpm run build` (all 8 nx targets) clean.
- All contract ABI entries verified byte-for-byte against the compiled forge artifacts in `vault-contracts-aave-v4`, including full tuple component types and 4-byte selectors, cross-checked against that repo's `snapshots/selectors.md`.

New tests worth looking at:

- A **two-epoch rotation fixture** with real secp256k1 keys: a VP that rotates, a keeper that rotates to a key sorting *before* its original, a keeper with a custom payout script, and controls that never rotate. It pins that resolving at epoch 0 reproduces today's keys exactly, that the sorted order reorders after rotation, and that collisions and off-curve keys are rejected.
- ABI decode tests pinning **both** directions: the shared 13-field entry keeps working against an upgraded registry, and the extended entry mis-decodes a populated legacy response (documenting the hazard) while throwing on the empty response the capability probe relies on.
- **Payout-destination resolution** — that destinations are always resolved, and that the keeper map is keyed from the roster-ordered pairs rather than index-joined against the sorted array, using a fixture whose roster order is the reverse of its sorted order.
- **Dual-accept** — that a BIP-86 destination is accepted when a different script is registered (the legacy-graph case), and that a third script is still rejected.

## If we split this into several PRs

Both halves are now verified on devnet, so urgency no longer separates them — but file count and code ownership still argue for a split. The rebase onto #2169 strengthens the case: `payout.ts` now carries two independent changes to the same validator, which is exactly the kind of file that benefits from a reviewer looking at it alone. `.github/workflows/critical-path-check.yml` matches a literal path list, and **two** changed files are on it: `packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts` and `packages/babylon-ts-sdk/src/tbv/core/services/deposit/signDepositorGraph.ts`. Both carry the same small change — the two RFC-006 payout-destination params and their forwarding — so isolating them together means the two required code owners spend their attention on payout logic rather than on 900 lines of ABI JSON. `signDepositorGraph.ts` also picks up a one-line import reorder: `prettier-plugin-organize-imports` requires it, and the file was drifted on `main` (it only checks clean outside the TS project, where the plugin cannot resolve module paths).

The natural seam is the payout-script work:

| # | Title | Behavior change | Reviewers |
|---|---|---|---|
| 1 | `feat(ts-sdk): add RFC-006 operation-key ABIs and chain readers` | none — nothing calls it | 1 |
| 2 | `feat(ts-sdk): resolve participant operation keys and validate them for new peg-ins` | none until wired | 1 |
| 3 | `feat(ts-sdk): pin payout destinations to registered scripts, accepting BIP-86 for legacy graphs` | none until wired | **2** (auto-labelled) |
|   | ↳ carries **both** critical-path files: `payout.ts` and `signDepositorGraph.ts` | | |
| 4 | `feat(vault): resolve participant keys by epoch across build, resume, payout and refund` | the behaviour change | 1 |

PR 3 has no dependencies and can open immediately in parallel, which helps because it needs two approvals and is the likely scheduling bottleneck.

**Do not go below three.** PR 3 should never merge into anything else — the moment `payout.ts` or `signDepositorGraph.ts` shares a PR with unrelated work, the two required code owners cannot realistically read the whole diff. Keeping both in PR 3 is what leaves PR 4 at a single reviewer.

## Note for reviewers

This touches the presigning and payout paths, which are on the critical-path list in `CLAUDE.md` and need two reviewers.

Beyond the deployment-fact corrections above, a few claims in https://github.com/babylonlabs-io/babylon-toolkit/issues/2150 did not survive checking against the contracts and this codebase:

- "The ABI sync is additive with no behaviour change" is **not** true for `getBtcVaultProtocolInfo` — see the ABI section.
- The issue references a payout fee band keyed on output script lengths (`out0Len`/`out1Len`). **No such code exists in this repo**; payout has only a 10%-of-inputs implicit fee cap.
- The issue's Stage 0 asks for `vpKeyEpochCurrent` / `appKeeperKeyEpochCurrent` / `ucKeyEpochCurrent`. **`vpKeyEpochCurrent` does not exist on the deployed contract** — it reverts. The "current" mode calls `getCurrentOperationBtcKey` directly and never needs an epoch number.
- The payout change is smaller than described and pointed at the wrong output: output 0 was **already** correct for VP and depositor claimers. The only wrong derivation was the vault-keeper branch. Pinning the VP commission destination is new hardening.

One naming note that cost some confusion: **an epoch is not a rotation count.** Devnet's VP had rotated only twice when its vaults were already freezing `vpKeyEpoch: 12`, because a single per-operator counter advances on `setPayoutScript` too. Key resolution is unaffected — the contract returns the latest version at or before the epoch — but note that `getOperationBtcKeyAtEpoch` never reverts on an out-of-range epoch (verified up to `2^64-1`), so a garbage epoch resolves silently rather than failing. `assertEpochInRange` is what bounds that.

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/contracts/__tests__/BTCVaultRegistryKeyEpochs.abi.test.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/contracts/__tests__/BTCVaultRegistryKeyEpochs.abi.test.ts
@@ -1,0 +1,173 @@
+import { decodeFunctionResult, encodeFunctionResult } from "viem";
+import { describe, expect, it } from "vitest";
+
+import { BTCVaultRegistryABI } from "../abis/BTCVaultRegistry.abi";
+import { BTCVaultRegistryKeyEpochsABI } from "../abis/BTCVaultRegistryKeyEpochs.abi";
+
+/**
+ * `getBtcVaultProtocolInfo` is declared twice on purpose: the shared ABI keeps
+ * the 13-field shape that decodes against every deployed registry, and the
+ * key-epochs ABI carries the three RFC-006 epoch fields. These tests pin the
+ * two properties that make that split safe.
+ */
+
+type AbiComponent = { readonly name: string; readonly type: string };
+
+function protocolInfoComponents(
+  abi: typeof BTCVaultRegistryABI | typeof BTCVaultRegistryKeyEpochsABI,
+): AbiComponent[] {
+  const entry = abi.find(
+    (item) =>
+      item.type === "function" && item.name === "getBtcVaultProtocolInfo",
+  );
+  if (!entry || !("outputs" in entry)) {
+    throw new Error("getBtcVaultProtocolInfo not found in ABI");
+  }
+  const output = entry.outputs[0] as { components?: readonly AbiComponent[] };
+  if (!output.components) {
+    throw new Error("getBtcVaultProtocolInfo output has no components");
+  }
+  return [...output.components];
+}
+
+/** A legacy (pre-RFC-006) protocol-info tuple, as a 13-field registry returns it. */
+const LEGACY_PROTOCOL_INFO = {
+  depositorSignedPeginTx: "0xdeadbeef",
+  universalChallengersVersion: 1,
+  appVaultKeepersVersion: 2,
+  offchainParamsVersion: 3,
+  verifiedAt: 4n,
+  depositorWotsPkHash: `0x${"11".repeat(32)}`,
+  hashlock: `0x${"22".repeat(32)}`,
+  htlcVout: 0,
+  depositorPopSignature: "0xc0ffee",
+  prePeginTxHash: `0x${"33".repeat(32)}`,
+  vaultProviderCommissionBps: 100,
+  claimExpiredUntil: 0n,
+  vaultCoreVersion: 1,
+} as const;
+
+const RFC006_PROTOCOL_INFO = {
+  ...LEGACY_PROTOCOL_INFO,
+  vpKeyEpoch: 7n,
+  appKeeperKeyEpoch: 8n,
+  ucKeyEpoch: 9n,
+} as const;
+
+describe("BTCVaultRegistryKeyEpochsABI", () => {
+  it("declares the 13 shared components as an exact prefix of its 16", () => {
+    const shared = protocolInfoComponents(BTCVaultRegistryABI);
+    const extended = protocolInfoComponents(BTCVaultRegistryKeyEpochsABI);
+
+    expect(shared).toHaveLength(13);
+    expect(extended).toHaveLength(16);
+
+    // Name AND type, in order. The tuple is positional, so a rename or a
+    // widened type in one copy and not the other silently misaligns every
+    // field after it.
+    expect(extended.slice(0, shared.length)).toEqual(shared);
+  });
+
+  it("appends exactly the three RFC-006 uint64 key epochs", () => {
+    const extended = protocolInfoComponents(BTCVaultRegistryKeyEpochsABI);
+
+    expect(extended.slice(13)).toEqual([
+      { name: "vpKeyEpoch", type: "uint64", internalType: "uint64" },
+      { name: "appKeeperKeyEpoch", type: "uint64", internalType: "uint64" },
+      { name: "ucKeyEpoch", type: "uint64", internalType: "uint64" },
+    ]);
+  });
+
+  it("silently mis-decodes a populated pre-RFC-006 registry response", () => {
+    // The hazard, stated exactly. `BTCVaultProtocolInfo` is a *dynamic* tuple
+    // (it contains `bytes`), so its encoding is a 13-word head followed by the
+    // tail holding the dynamic payloads. Asking for 16 head words does not run
+    // off the end of the returndata — it reads three words of *tail* and hands
+    // them back as epochs. No throw, no warning, plausible-looking small
+    // integers.
+    //
+    // So the extended ABI must never be read on a registry that has not been
+    // upgraded. Being careful with the values is not enough; the capability
+    // probe below is the only thing standing between this and resolving
+    // participant keys at a fabricated epoch.
+    const legacyReturndata = encodeFunctionResult({
+      abi: BTCVaultRegistryABI,
+      functionName: "getBtcVaultProtocolInfo",
+      result: LEGACY_PROTOCOL_INFO,
+    });
+
+    const decoded = decodeFunctionResult({
+      abi: BTCVaultRegistryKeyEpochsABI,
+      functionName: "getBtcVaultProtocolInfo",
+      data: legacyReturndata,
+    }) as unknown as Record<string, bigint>;
+
+    // Garbage, not an error — and not the zeros a real pre-rotation vault has.
+    expect(decoded.vpKeyEpoch).not.toBe(0n);
+  });
+
+  it("throws on an empty response from a legacy registry", () => {
+    // A *nonexistent* vault has empty `bytes` fields, so its tail is only two
+    // length words — too short to satisfy a 16-word head, and viem throws.
+    // Contrast with the populated-vault case above, where a legacy registry
+    // decodes silently into garbage epochs instead of failing.
+    //
+    // On an upgraded registry the same call decodes cleanly to all-zero.
+    const emptyLegacyReturndata = encodeFunctionResult({
+      abi: BTCVaultRegistryABI,
+      functionName: "getBtcVaultProtocolInfo",
+      result: {
+        ...LEGACY_PROTOCOL_INFO,
+        depositorSignedPeginTx: "0x",
+        depositorPopSignature: "0x",
+      },
+    });
+
+    expect(() =>
+      decodeFunctionResult({
+        abi: BTCVaultRegistryKeyEpochsABI,
+        functionName: "getBtcVaultProtocolInfo",
+        data: emptyLegacyReturndata,
+      }),
+    ).toThrow();
+  });
+
+  it("lets the shared ABI keep decoding an upgraded registry response", () => {
+    // The other direction must NOT throw: the 13-field shared entry is read on
+    // every legacy code path and has to keep working after a registry is
+    // upgraded, ignoring the trailing epoch fields.
+    const rfc006Returndata = encodeFunctionResult({
+      abi: BTCVaultRegistryKeyEpochsABI,
+      functionName: "getBtcVaultProtocolInfo",
+      result: RFC006_PROTOCOL_INFO,
+    });
+
+    const decoded = decodeFunctionResult({
+      abi: BTCVaultRegistryABI,
+      functionName: "getBtcVaultProtocolInfo",
+      data: rfc006Returndata,
+    });
+
+    expect(decoded).toMatchObject(LEGACY_PROTOCOL_INFO);
+  });
+
+  it("decodes the epochs from an upgraded registry response", () => {
+    const rfc006Returndata = encodeFunctionResult({
+      abi: BTCVaultRegistryKeyEpochsABI,
+      functionName: "getBtcVaultProtocolInfo",
+      result: RFC006_PROTOCOL_INFO,
+    });
+
+    const decoded = decodeFunctionResult({
+      abi: BTCVaultRegistryKeyEpochsABI,
+      functionName: "getBtcVaultProtocolInfo",
+      data: rfc006Returndata,
+    });
+
+    expect(decoded).toMatchObject({
+      vpKeyEpoch: 7n,
+      appKeeperKeyEpoch: 8n,
+      ucKeyEpoch: 9n,
+    });
+  });
+});
--- a/packages/babylon-ts-sdk/src/tbv/core/contracts/__tests__/BTCVaultRegistryKeyEpochs.abi.test.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/contracts/__tests__/BTCVaultRegistryKeyEpochs.abi.test.ts
@@ -87,9 +87,13 @@ describe("BTCVaultRegistryKeyEpochsABI", () => {
     // integers.
     //
     // So the extended ABI must never be read on a registry that has not been
-    // upgraded. Being careful with the values is not enough; the capability
-    // probe below is the only thing standing between this and resolving
-    // participant keys at a fabricated epoch.
+    // upgraded, and being careful with the decoded values is not enough — the
+    // `uint64` range check in `vault-registry-reader.ts` rejects the wild
+    // values but not the plausible ones. What stands between this and resolving
+    // participant keys at a fabricated epoch is a deployment invariant: every
+    // network this ships to already has the RFC-006 getters, and mainnet is a
+    // fresh RFC-006 deploy. This test pins the hazard so that invariant stays
+    // a conscious one.
     const legacyReturndata = encodeFunctionResult({
       abi: BTCVaultRegistryABI,
       functionName: "getBtcVaultProtocolInfo",
--- a/packages/babylon-ts-sdk/src/tbv/core/contracts/abis/BTCVaultRegistry.abi.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/contracts/abis/BTCVaultRegistry.abi.ts
@@ -171,15 +171,31 @@ export const BTCVaultRegistryABI = [
         type: "tuple[]",
         internalType: "struct BTCVaultRegistryTypes.BatchPeginRequest[]",
         components: [
-          { name: "depositorBtcPubKey", type: "bytes32", internalType: "bytes32" },
+          {
+            name: "depositorBtcPubKey",
+            type: "bytes32",
+            internalType: "bytes32",
+          },
           { name: "btcPopSignature", type: "bytes", internalType: "bytes" },
           { name: "unsignedPrePeginTx", type: "bytes", internalType: "bytes" },
-          { name: "depositorSignedPeginTx", type: "bytes", internalType: "bytes" },
+          {
+            name: "depositorSignedPeginTx",
+            type: "bytes",
+            internalType: "bytes",
+          },
           { name: "hashlock", type: "bytes32", internalType: "bytes32" },
           { name: "htlcVout", type: "uint8", internalType: "uint8" },
           { name: "referralCode", type: "uint32", internalType: "uint32" },
-          { name: "depositorPayoutBtcAddress", type: "bytes", internalType: "bytes" },
-          { name: "depositorWotsPkHash", type: "bytes32", internalType: "bytes32" },
+          {
+            name: "depositorPayoutBtcAddress",
+            type: "bytes",
+            internalType: "bytes",
+          },
+          {
+            name: "depositorWotsPkHash",
+            type: "bytes32",
+            internalType: "bytes32",
+          },
         ],
       },
     ],
@@ -233,23 +249,54 @@ export const BTCVaultRegistryABI = [
   {
     type: "function",
     name: "getVaultProviderBTCKey",
+    inputs: [{ name: "vpAddr", type: "address", internalType: "address" }],
+    outputs: [{ name: "", type: "bytes32", internalType: "bytes32" }],
+    stateMutability: "view",
+  },
+  // --- RFC-006 operation-key resolution ---------------------------------
+  // A vault provider's BTC key is no longer the fixed registration key
+  // returned by `getVaultProviderBTCKey` above: it is an append-only history
+  // of *operation* keys, rotated by the provider's cold ETH admin key. Each
+  // vault freezes a `vpKeyEpoch` at `submitPeginRequest` and resolves the key
+  // bonded at that epoch, so a later rotation never invalidates a live vault.
+  //
+  // Both getters fall back on-chain when the provider never rotated:
+  // `getCurrentOperationBtcKey` returns the registration key at version 0, and
+  // `getPayoutScriptAtEpoch` returns the BIP-86 P2TR of the epoch's operation
+  // key when no explicit payout script was registered. That is what makes
+  // adopting them a no-op until the first rotation on a network.
+  {
+    type: "function",
+    name: "getCurrentOperationBtcKey",
+    inputs: [{ name: "provider", type: "address", internalType: "address" }],
+    outputs: [{ name: "", type: "bytes32", internalType: "bytes32" }],
+    stateMutability: "view",
+  },
+  {
+    type: "function",
+    name: "getOperationBtcKeyAtEpoch",
     inputs: [
-      { name: "vpAddr", type: "address", internalType: "address" },
+      { name: "provider", type: "address", internalType: "address" },
+      { name: "epoch", type: "uint64", internalType: "uint64" },
     ],
-    outputs: [
-      { name: "", type: "bytes32", internalType: "bytes32" },
+    outputs: [{ name: "", type: "bytes32", internalType: "bytes32" }],
+    stateMutability: "view",
+  },
+  {
+    type: "function",
+    name: "getPayoutScriptAtEpoch",
+    inputs: [
+      { name: "provider", type: "address", internalType: "address" },
+      { name: "epoch", type: "uint64", internalType: "uint64" },
     ],
+    outputs: [{ name: "", type: "bytes", internalType: "bytes" }],
     stateMutability: "view",
   },
   {
     type: "function",
     name: "getVaultProviderCommission",
-    inputs: [
-      { name: "vpAddr", type: "address", internalType: "address" },
-    ],
-    outputs: [
-      { name: "", type: "uint16", internalType: "uint16" },
-    ],
+    inputs: [{ name: "vpAddr", type: "address", internalType: "address" }],
+    outputs: [{ name: "", type: "uint16", internalType: "uint16" }],
     stateMutability: "view",
   },
   {
@@ -361,8 +408,7 @@ export const BTCVaultRegistryABI = [
       {
         name: "vProtocol",
         type: "tuple",
-        internalType:
-          "struct BTCVaultRegistryTypes.BTCVaultProtocolInfo",
+        internalType: "struct BTCVaultRegistryTypes.BTCVaultProtocolInfo",
         components: [
           {
             name: "depositorSignedPeginTx",
--- a/packages/babylon-ts-sdk/src/tbv/core/contracts/abis/BTCVaultRegistry.abi.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/contracts/abis/BTCVaultRegistry.abi.ts
@@ -260,11 +260,23 @@ export const BTCVaultRegistryABI = [
   // vault freezes a `vpKeyEpoch` at `submitPeginRequest` and resolves the key
   // bonded at that epoch, so a later rotation never invalidates a live vault.
   //
-  // Both getters fall back on-chain when the provider never rotated:
-  // `getCurrentOperationBtcKey` returns the registration key at version 0, and
-  // `getPayoutScriptAtEpoch` returns the BIP-86 P2TR of the epoch's operation
-  // key when no explicit payout script was registered. That is what makes
-  // adopting them a no-op until the first rotation on a network.
+  // `getCurrentOperationBtcKey` falls back on-chain to the registration key at
+  // version 0, so adopting it is a no-op for a provider that never rotated.
+  //
+  // `getPayoutScriptAtEpoch` does NOT behave that way, and assuming it did was
+  // wrong in an earlier revision of this comment. A payout script is mandatory
+  // at registration and stored as version 1, so on an RFC-006 registry this
+  // returns the operator's own registered script from the very first epoch —
+  // not a BIP-86 derivation of its key. The contract's BIP-86 branch is a
+  // migration default for pre-RFC-006 rows and is not reachable from
+  // registration. An operator whose registered script happens to equal
+  // bip86(key) is a coincidence of what it chose to register, not a fallback.
+  //
+  // This is why depositor payout validation must accept the legacy BIP-86 form
+  // alongside the registered one: a payout graph is built once at BaBe Setup
+  // and never rebuilt, so a vault whose graph predates its network's
+  // btc-vault#2440 upgrade pays BIP-86 for the rest of its life. See
+  // `acceptedPayoutScriptHexes` in `primitives/psbt/payout.ts`.
   {
     type: "function",
     name: "getCurrentOperationBtcKey",
--- a/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts
@@ -40,6 +40,7 @@ import {
   PEGIN_VAULT_OUTPUT_INDEX,
   VP_CLAIMER_PAYOUT_OUTPUT_COUNT,
 } from "./constants";
+import { assertStandardPayoutScript } from "./standardPayoutScript";
 
 /**
  * Number of items in a Taproot script-path spend witness stack for a
@@ -149,6 +150,31 @@ export interface PayoutParams {
    * pinned models.
    */
   councilSize: number;
+
+  /**
+   * RFC-006. Expected `outs[0].script` per vault-keeper claimer, keyed by
+   * lowercased x-only **operation** pubkey (no `0x`), resolved from
+   * `ApplicationRegistry.getPayoutScriptAtEpoch` at the vault's frozen
+   * `appKeeperKeyEpoch`.
+   *
+   * Every VK claimer must be present: a claimer missing from the map is an
+   * error rather than a cue to derive BIP-86, because a gap means resolution
+   * was incomplete and we do not know what that keeper registered.
+   *
+   * Each entry accepts either that registered script or the BIP-86 default of
+   * the same bonded key, so graphs built before btc-vault#2440 remain signable
+   * — see {@link acceptedPayoutScriptHexes} for why that is required and when
+   * it can be dropped.
+   */
+  vkClaimerPayoutScriptPubKeys: Readonly<Record<string, string>>;
+
+  /**
+   * RFC-006. Expected `outs[1].script` for the VP-claimer commission output,
+   * from `BTCVaultRegistry.getPayoutScriptAtEpoch` at the vault's frozen
+   * `vpKeyEpoch`. The BIP-86 default of the bonded VP key is accepted alongside
+   * it — see {@link acceptedPayoutScriptHexes}.
+   */
+  vpCommissionScriptPubKey: string;
 }
 
 /**
@@ -383,15 +409,94 @@ function assemblePayoutPsbt(
 }
 
 /**
+ * The scriptPubKeys a payout output may legitimately pay for `bondedPubkey`.
+ *
+ * ## What this is
+ *
+ * A deliberate dual-accept: both the operator's RFC-006 **registered** payout
+ * script and the **BIP-86 default** of its bonded operation key are treated as
+ * valid destinations for the same output.
+ *
+ * ## Why both forms exist
+ *
+ * A vault provider running a daemon that predates btc-vault#2440 computes
+ * operator payout destinations itself, via BIP-86 over the bonded operation
+ * key. From #2440 onward it reads them from the registry instead
+ * (`getPayoutScriptAtEpoch`), so an operator can point payouts at cold custody.
+ *
+ * The two forms are not a transition state that resolves on its own. A payout
+ * graph is built once, at BaBe Setup, and is **never rebuilt** — so a vault
+ * whose graph was built before its network's #2440 upgrade pays BIP-86 for the
+ * rest of its life. Accepting only the registered form makes every such vault
+ * permanently unsignable: the depositor can never complete the deposit, and
+ * the BTC is already committed by then. Confirmed on devnet 2026-08-04, where
+ * a pre-upgrade vault failed exactly this way.
+ *
+ * ## Why this is not a weakening
+ *
+ * Both candidates are derived here from on-chain state — the registry read and
+ * a local BIP-86 derivation over the bonded key. Neither is taken from the VP's
+ * response, so a substituted or attacker-chosen script still fails.
+ *
+ * The only substitution this permits is between two destinations the *same
+ * operator* already controls, which moves that operator's own funds between its
+ * own addresses. It cannot redirect value to a third party, and it cannot touch
+ * the depositor's output. For the VP commission the value cap
+ * (`floor(peginValue × commissionBps / 10_000)`) still bounds depositor
+ * exposure independently of where the commission goes. The BIP-86 branch is
+ * also precisely what every pre-RFC-006 client pinned, so this is the older
+ * rule retained alongside the newer one, not a laxer rule invented for it.
+ *
+ * ## When this can be removed
+ *
+ * This validation runs at one point in the lifecycle: depositor payout signing
+ * (`PENDING` → `VERIFIED`). It is not re-run on refund or withdrawal. So the
+ * BIP-86 branch stops being reachable for a network once every vault
+ * registered before that network's #2440 upgrade has left
+ * `PendingDepositorSignatures` — signed and activated, or expired past its
+ * activation deadline. That is bounded by the vault lifetime, not indefinite.
+ *
+ * Removal is therefore gated by the **last** network to upgrade, since this
+ * code is shared. Concretely: drop the BIP-86 branch only once devnet, testnet
+ * and mainnet have all been on #2440 for longer than the activation deadline,
+ * and no vault registered before those upgrades is still awaiting signatures.
+ * Deleting it earlier strands deposits with BTC already locked; there is no
+ * recovery path short of the refund timelock.
+ *
+ * @internal Helper invoked by {@link assertPayoutOutputLayout}.
+ */
+function acceptedPayoutScriptHexes(
+  registeredScriptPubKey: string,
+  bondedPubkey: string,
+): string[] {
+  const registered = stripHexPrefix(registeredScriptPubKey).toLowerCase();
+  const legacyBip86 = stripHexPrefix(
+    deriveBip86ScriptPubKeyHex(bondedPubkey),
+  ).toLowerCase();
+  return registered === legacyBip86 ? [registered] : [registered, legacyBip86];
+}
+
+/** Whether `script` matches any of `acceptedHexes`. */
+function matchesAnyScript(script: Buffer, acceptedHexes: string[]): boolean {
+  return acceptedHexes.some((hex) => script.equals(Buffer.from(hex, "hex")));
+}
+
+/**
  * Validate a payout transaction's output structure for the claimer's role,
  * keyed on `claimerBtcPubkey`. Pins per role: `outs.length`, `outs[0].script`,
  * `outs[last].value` (anchor dust), and (VP-claimer) `outs[1].value` capped at
  * `floor(peginValue × commissionBps / 10_000)`. Canonical layouts: VP-claimer
  * = [payout, commission, anchor]; depositor/VK-claimer = [payout, anchor].
  *
- * `outs[1].script` and `outs[last].script` are intentionally not pinned: the
- * value pins above bound depositor exposure regardless of where those outputs
- * are sent, so the value pins — not script pins — are load-bearing.
+ * `outs[last].script` (the CPFP anchor) is intentionally not pinned: the value
+ * pin above bounds depositor exposure regardless of where that dust goes.
+ *
+ * `outs[1].script` (the VP commission) was unpinned for the same reason.
+ * RFC-006 supersedes that rationale: the commission destination is now an
+ * operator-registered scriptPubKey we can resolve independently at the vault's
+ * frozen `vpKeyEpoch`, so when `vpCommissionScriptPubKey` is supplied it is
+ * pinned too. The value cap stays — it is still what bounds exposure — and the
+ * script pin is added precision, not a replacement for it.
  *
  * Returns the layout-trusted non-anchor script lengths for the fee band:
  * `out0Len` from the pinned outs[0] script (registered payout script, or
@@ -414,6 +519,8 @@ function assertPayoutOutputLayout(
     vaultKeeperBtcPubkeys,
     registeredPayoutScriptPubKey,
     commissionBps,
+    vkClaimerPayoutScriptPubKeys,
+    vpCommissionScriptPubKey,
   } = params;
 
   if (!isValidHex(registeredPayoutScriptPubKey)) {
@@ -430,20 +537,39 @@ function assertPayoutOutputLayout(
   type Role = "vp-claimer" | "depositor-as-claimer" | "vk-claimer";
   let role: Role;
   let expectedOutCount: number;
-  let expectedOut0ScriptHex: string;
+  let acceptedOut0ScriptHexes: string[];
 
   if (claimer === vp) {
     role = "vp-claimer";
     expectedOutCount = VP_CLAIMER_PAYOUT_OUTPUT_COUNT;
-    expectedOut0ScriptHex = stripHexPrefix(registeredPayoutScriptPubKey);
+    acceptedOut0ScriptHexes = [stripHexPrefix(registeredPayoutScriptPubKey)];
   } else if (claimer === dep) {
     role = "depositor-as-claimer";
     expectedOutCount = NON_VP_CLAIMER_PAYOUT_OUTPUT_COUNT;
-    expectedOut0ScriptHex = stripHexPrefix(registeredPayoutScriptPubKey);
+    acceptedOut0ScriptHexes = [stripHexPrefix(registeredPayoutScriptPubKey)];
   } else if (keepers.includes(claimer)) {
     role = "vk-claimer";
     expectedOutCount = NON_VP_CLAIMER_PAYOUT_OUTPUT_COUNT;
-    expectedOut0ScriptHex = stripHexPrefix(deriveBip86ScriptPubKeyHex(claimer));
+    // RFC-006: a keeper's payout goes to the scriptPubKey it registered
+    // on-chain, resolved at the vault's frozen keeper epoch. Deriving BIP-86
+    // locally as the sole expectation would reject a valid payout the moment a
+    // keeper points its payouts at cold custody.
+    //
+    // A missing entry is an error, never a BIP-86 fallback: the registry
+    // backfills BIP-86 itself for keepers that never registered a script, so a
+    // gap here means the resolution was incomplete, not that this keeper is on
+    // the default.
+    const registered = vkClaimerPayoutScriptPubKeys[claimer];
+    if (registered === undefined) {
+      throw new Error(
+        `No registered payout script resolved for vault keeper claimer ${claimer}`,
+      );
+    }
+    assertStandardPayoutScript(
+      registered,
+      `Vault keeper payout script for claimer ${claimer}`,
+    );
+    acceptedOut0ScriptHexes = acceptedPayoutScriptHexes(registered, claimer);
   } else {
     throw new Error(
       `Unknown claimer pubkey ${claimer}: not VP, depositor, or a registered vault keeper`,
@@ -457,10 +583,11 @@ function assertPayoutOutputLayout(
     );
   }
 
-  const expectedOut0Script = Buffer.from(expectedOut0ScriptHex, "hex");
-  if (!payoutTx.outs[0].script.equals(expectedOut0Script)) {
+  if (!matchesAnyScript(payoutTx.outs[0].script, acceptedOut0ScriptHexes)) {
     throw new Error(
-      `Payout transaction output 0 does not pay the expected scriptPubKey for role ${role}`,
+      `Payout transaction output 0 does not pay the expected scriptPubKey for role ${role}. ` +
+        `Accepted: ${acceptedOut0ScriptHexes.join(" or ")}; ` +
+        `got ${payoutTx.outs[0].script.toString("hex")}`,
     );
   }
 
@@ -473,6 +600,27 @@ function assertPayoutOutputLayout(
   }
 
   if (role === "vp-claimer") {
+    // RFC-006 commission destination. Checked before the value cap so a
+    // substituted destination reports as such rather than surfacing later as a
+    // confusing amount error.
+    assertStandardPayoutScript(
+      vpCommissionScriptPubKey,
+      "Vault provider commission payout script",
+    );
+    const acceptedCommissionScriptHexes = acceptedPayoutScriptHexes(
+      vpCommissionScriptPubKey,
+      vp,
+    );
+    if (
+      !matchesAnyScript(payoutTx.outs[1].script, acceptedCommissionScriptHexes)
+    ) {
+      throw new Error(
+        `Payout transaction output 1 does not pay the vault provider's commission scriptPubKey. ` +
+          `Accepted: ${acceptedCommissionScriptHexes.join(" or ")}; ` +
+          `got ${payoutTx.outs[1].script.toString("hex")}`,
+      );
+    }
+
     // Structural guard only — a non-negative integer below the bps
     // denominator, so the cap math `floor(peginValue * bps / 10_000)` is
     // meaningful. The protocol minimum is enforced at the trust boundary
@@ -499,7 +647,10 @@ function assertPayoutOutputLayout(
     }
   }
 
-  const out0Len = expectedOut0Script.length;
+  // The accepted-script set can hold more than one candidate (dual-accept), so
+  // the length the fee band consumes is that of the output actually present —
+  // which the check above has already pinned to one of the accepted values.
+  const out0Len = payoutTx.outs[0].script.length;
   if (out0Len === 0 || out0Len > MAX_PAYOUT_SCRIPT_LEN) {
     throw new Error(
       `Payout receiver scriptPubKey length ${out0Len} is outside the ` +
@@ -507,18 +658,14 @@ function assertPayoutOutputLayout(
         `refusing to sign payout.`,
     );
   }
+  // No length cap on outs[1]: under RFC-006 the commission output is pinned to
+  // the operator's registered script, and `assertStandardPayoutScript` above
+  // already bounds that to a standard type (34 bytes at most). A separate
+  // 128-byte cap here would be unreachable. Still measured, because the fee
+  // floor consumes it — and now from a pinned source rather than a
+  // VP-controlled one.
   const out1Len =
     role === "vp-claimer" ? payoutTx.outs[1].script.length : undefined;
-  if (
-    out1Len !== undefined &&
-    (out1Len === 0 || out1Len > MAX_PAYOUT_SCRIPT_LEN)
-  ) {
-    throw new Error(
-      `Payout commission scriptPubKey length ${out1Len} is outside the ` +
-        `contract's registration cap [1, ${MAX_PAYOUT_SCRIPT_LEN}]; ` +
-        `refusing to sign payout.`,
-    );
-  }
 
   return { out0Len, out1Len };
 }
--- a/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/standardPayoutScript.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/standardPayoutScript.ts
@@ -7,9 +7,25 @@
  * shape would let a keeper that registered a garbage or unspendable script get
  * the depositor to pre-sign a payout nobody can ever claim.
  *
- * The registry accepts anything; the protocol only makes sense for outputs
- * that are spendable and representable as an address, which is the same gate
- * `vaultd` applies before it will build a graph.
+ * This bound is ours alone — it is deliberately NOT parity with `vaultd`.
+ * `vaultd` uses registry scripts verbatim at graph-build time and documents
+ * why: applying a predicate the contract does not apply would block every
+ * peg-in for an operator the instant the two disagreed. The registry itself
+ * only bounds length to 1..128 bytes, and the registration CLI has an
+ * `--allow-non-standard` escape hatch.
+ *
+ * We hold the stricter line because we occupy a different position: we are the
+ * party being asked to *pre-sign* against these bytes, and a depositor
+ * signature pinned to a provably unspendable output is not recoverable. The
+ * bound is also load-bearing for the payout fee band — capping a standard
+ * output at 34 bytes is what lets `assertPayoutFeeBand` reason about `outs[1]`
+ * without a separate length cap.
+ *
+ * The deliberate consequence: an operator that registers a non-standard script
+ * via `--allow-non-standard` can have its graphs built by `vaultd` and still be
+ * refused here, blocking deposits against that operator in this dApp. That is
+ * fail-closed by choice — the alternative is pre-signing into a destination we
+ * cannot verify is spendable.
  *
  * @module primitives/psbt/standardPayoutScript
  */
--- a/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/standardPayoutScript.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/standardPayoutScript.ts
@@ -1,0 +1,86 @@
+/**
+ * Standard-scriptPubKey validation for registry-supplied payout destinations.
+ *
+ * RFC-006 lets an operator register an arbitrary payout scriptPubKey, so the
+ * bytes we pin a `PayoutTx` output against are now chosen by that operator
+ * rather than derived by us. Pinning an output to them without checking the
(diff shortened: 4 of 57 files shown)


services/vault/src/utils/operationKeyRotationSupport.ts
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * A decode failure means the registry returned the shorter, pre-RFC-006 shape.
+ * Anything else — an RPC blip, a timeout, a reverted call — is transient and
+ * must be rethrown, so a flaky endpoint can never be mistaken for "this
+ * network doesn't support rotation".
+ */
+function isShortReturndataError(error: unknown): boolean {
+  const name = (error as { name?: string })?.name ?? "";
+  return (
+    name === "AbiDecodingDataSizeTooSmallError" ||
```
jrwbabylonlab: `isShortReturndataError` reads top-level `error.name`, but viem 2.38.2 `readContract` wraps decode failures via `getContractError` into a `ContractFunctionExecutionError` with the real error as `cause`. So this is always false, `probe()` always rethrows, and `supported: false` can never be produced — the negative cache, the throw below, and both `operationKeyRotationUnsupported` copy strings are unreachable. Fails closed, so it is broken diagnostics rather than unsafe. Walk `error.cause`.

services/vault/src/services/vault/vaultPayoutSignatureService.ts
```diff
@@ -200,17 +218,78 @@ export async function prepareSigningContext(
     );
   }
 
-  const vaultProviderBtcPubkey = await resolveVaultProviderBtcPubkey(
+  const registrationVpBtcPubkey = await resolveVaultProviderBtcPubkey(
```
jrwbabylonlab, line 244: `resolveVaultProviderBtcPubkey` runs unconditionally here, before the operation-key branch below, and throws on a VP-hint mismatch rather than warning. The day the indexer serves the rotated operation key while the registry getter still returns the registration key, payout signing throws for every depositor of that VP with no user workaround — triggered by an indexer change with no deploy on our side.

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -316,6 +317,16 @@ export function useVaultActions(): UseVaultActionsReturn {
               buildUniversalChallengersVersion,
             expectedVaultCoreVersion: buildVaultCoreVersion,
           });
+
```
jrwbabylonlab: The key guard is nested behind four unrelated version preconditions, though its only real precondition is `pendingPegin.buildParticipantOperationKeys`, which is checked separately. Any missing version silently disables the check.

services/vault/src/services/vault/verifyResumeParticipantKeys.ts
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Throw when the vault's frozen epochs no longer resolve to the keys the
+ * Pre-PegIn was built with. A no-op when the flag is off or the record predates
+ * the stamp.
+ */
+export async function verifyResumeParticipantKeys(params: {
+  vaultId: Hex;
+  expected: ExpectedParticipantOperationKeys;
+}): Promise<void> {
+  if (!featureFlags.isOperationKeyRotationEnabled) return;
```
jrwbabylonlab: Turning the flag off — the natural response to trouble — removes this pre-broadcast check for exactly the in-flight deposits that were built with it on, while the signing and refund paths simultaneously revert to registration keys for vaults whose scripts hold rotated keys. Worth documenting that rollback is unsafe while stamped records exist.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/validateOnChainParticipantKeys.ts
```diff
@@ -60,47 +106,132 @@ export async function validateOnChainParticipantKeys(
...
+  // in it rotated) matches both and constrains nothing. But if one role can
+  // *only* be explained by the registration keys while another can *only* be
+  // explained by the operation keys, the indexer is serving a half-applied
+  // view — accepting that would mean building with a key set no single
+  // snapshot of the indexer ever held.
+  const roles = [vpMatch, keeperMatch, challengerMatch];
+  const pinsRegistration = roles.some((m) => m.registration && !m.operation);
+  const pinsOperation = roles.some((m) => m.operation && !m.registration);
+
+  if (pinsRegistration && pinsOperation) {
```
jrwbabylonlab, line 211: A partially caught-up indexer hard-blocks every deposit for this provider. The indexer is eventually consistent over three contracts, so a window where the VP role has advanced but the keeper role has not is expected during a rotation — and in that window "Refresh and try again" is an action that cannot help. Needs an explicit decision plus telemetry.

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -316,6 +317,16 @@ export function useVaultActions(): UseVaultActionsReturn {
...
+          // RFC-006: the same guard the inline deposit path applies before
+          // broadcast. Gated on the stamp being present — records written with
+          // the flag off, or before it shipped, simply skip it.
+          if (pendingPegin?.buildParticipantOperationKeys) {
+            await verifyResumeParticipantKeys({
+              vaultId,
+              expected: pendingPegin.buildParticipantOperationKeys,
+            });
+          }
         } catch (err) {
```
jrwbabylonlab, line 320: Question rather than a finding: the drift catch removes the pending record, after which resume falls back to the GraphQL unsigned tx and this whole block is skipped. The fallback is hash-pinned to the on-chain `prePeginTxHash`, so it cannot broadcast arbitrary bytes. But does that hash stay authoritative if keys rotated after registration — i.e. is a registered tx still the right tx once the vault bonds new keys? If yes this is fine; if not, the retry silently bypasses the guard that just fired.

packages/babylon-ts-sdk/src/tbv/core/contracts/abis/BTCVaultRegistry.abi.ts
```diff
@@ -241,6 +241,45 @@ export const BTCVaultRegistryABI = [
...
+  // --- RFC-006 operation-key resolution ---------------------------------
+  // A vault provider's BTC key is no longer the fixed registration key
+  // returned by `getVaultProviderBTCKey` above: it is an append-only history
+  // of *operation* keys, rotated by the provider's cold ETH admin key. Each
+  // vault freezes a `vpKeyEpoch` at `submitPeginRequest` and resolves the key
+  // bonded at that epoch, so a later rotation never invalidates a live vault.
+  //
+  // Both getters fall back on-chain when the provider never rotated:
+  // `getCurrentOperationBtcKey` returns the registration key at version 0, and
+  // `getPayoutScriptAtEpoch` returns the BIP-86 P2TR of the epoch's operation
```
jrwbabylonlab: Not a no-op: a payout script is mandatory at registration and stored as v1 (`VPKeyRegistryLogic.sol:86` + `:140`, `ApplicationRegistry.sol:743` + `:777`), so on an RFC-006 registry this returns the operator's own script, not BIP-86. The contract calls the BIP-86 branch a migration default "not reachable from registration" (`VPKeyRegistryLogic.sol:423-425`). Also means run 2's failure was the normal case, not devnet having an unusual custom script.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/standardPayoutScript.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Standard-scriptPubKey validation for registry-supplied payout destinations.
+ *
+ * RFC-006 lets an operator register an arbitrary payout scriptPubKey, so the
+ * bytes we pin a `PayoutTx` output against are now chosen by that operator
+ * rather than derived by us. Pinning an output to them without checking the
+ * shape would let a keeper that registered a garbage or unspendable script get
+ * the depositor to pre-sign a payout nobody can ever claim.
+ *
+ * The registry accepts anything; the protocol only makes sense for outputs
+ * that are spendable and representable as an address, which is the same gate
```
jrwbabylonlab: vaultd does the opposite at graph-build time and documents why (`payout_scripts.rs:178-190`): registry scripts are used verbatim, because a predicate the contract does not apply "would block every pegin for that operator the instant the two disagreed". The registry only bounds length 1..128, and the CLI has `--allow-non-standard`. So a legitimately-registered non-standard script makes us refuse a payout vaultd built. The check may still be right — the parity claim is not.

packages/babylon-ts-sdk/src/tbv/core/services/participants/resolveParticipantKeys.ts
```diff
@@ -0,0 +1,176 @@
...
+    operationBtcPubkey,
+    rotated: operationBtcPubkey !== genesisBtcPubkey,
+  };
+}
+
+/**
+ * Assert no two participants resolved to the same operation key.
+ *
+ * Registration-time checks guarantee distinctness across the *registered*
+ * keys, but nothing on-chain prevents two operators from rotating onto the
```
jrwbabylonlab: Overstated: intra-role uniqueness is enforced in all three registries (`VPKeyRegistryLogic.sol:191-205`, `ApplicationRegistry.sol:482-485`, `ProtocolParams.sol:209-210`), and VP rotation also rejects current VK/UC roster keys. Genuinely unguarded: keeper→VP key (your example), keeper↔UC, UC→VP, and keepers across different apps.

gbarkhatov:
@greptile review

services/vault/src/utils/operationKeyRotationSupport.ts
gbarkhatov: Correct diagnosis, but this file no longer exists — it was deleted along with both feature flags after this draft was reviewed.

The flags guarded a registry that predates RFC-006, and no network we ship to is in that state: devnet and testnet both have all four getters, and mainnet will be a fresh RFC-006 deploy. On devnet, flag-*off* was already the broken setting. So the probe was guarding a case that cannot occur, and you'd found that its negative path was unreachable anyway.

What replaces it is a deployment invariant plus the `uint64` range backstop in `vault-registry-reader.ts`. That backstop is deliberately documented as weak — it rejects wild values but not plausible small ones — so the real guarantee is stated as a deployment one rather than implied to be a runtime check. Your comment is also what prompted a sweep for stale prose referring to the removed probe: seven sites across four files, all corrected in bc5b2679.

services/vault/src/services/vault/verifyResumeParticipantKeys.ts
gbarkhatov: Outdated in its specifics — there is no longer a flag to turn off, so the rollback hazard you describe has no trigger. Both flags are gone as of the post-draft revision.

The underlying point survived into the code, though, and turned out to matter more than the flag framing suggested. The stamp on the pending record is now understood as load-bearing state, not an optimisation: it is the only thing that lets a resume re-detect key drift. That is why key drift no longer clears the pending record — see the thread on `useVaultActions.ts:330`, where the same reasoning surfaced a live bypass.

services/vault/src/hooks/deposit/useVaultActions.ts
gbarkhatov: This was a real bypass, not just a question. Thank you — it is the most valuable finding in the review, and I'd have shipped it.

**Confirming your trace.** `verifyResumeParticipantKeys` threw `RegisteredVaultVersionMismatchError`, which satisfies `isRegisteredVaultVersionMismatchError`, so the catch called `removePendingPegin`. On the next attempt `pendingPegin` is gone → `localUnsignedTxHex` is `undefined` → `unsignedTxHex` falls back to the indexer copy → and the entire guarded block is skipped, because its first condition is `localUnsignedTxHex`. The hash check passes, and the broadcast proceeds.

**Answering the question you actually asked** — is the registered hash still authoritative once keys rotate? **No, and that is the crux.** The hash is authoritative against *substitution*: it proves the transaction is byte-for-byte the one the contract registered. It says nothing about whether the vault's frozen epochs still resolve to the keys embedded in that transaction's scripts. Those are stamped by the same `submitPeginRequest` call, so drift means the registered hash commits to a transaction built with the *pre-rotation* keys while the vault froze the *post-rotation* epoch. Every counterparty resolves at the frozen epochs, computes a different funding output, and never recognises the broadcast one. The deposit cannot activate; the only exit is the Pre-PegIn refund timelock.

So the guard's own message — "The Pre-PegIn was not broadcast; the registered ETH vault will time out per protocol rules" — was made false by retrying. Not broadcasting costs the user a timed-out ETH vault; the retry cost them locked BTC.

**Fix in bc5b2679.** Key drift now throws `ParticipantKeyDriftError`, a *sibling* of `RegisteredVaultVersionMismatchError` and deliberately never a subclass — subclassing would keep satisfying the type guard and preserve the bug. Cleanup semantics now differ by design: a version mismatch still drops the record (the registered transaction stays the authoritative copy), key drift keeps it (the stamp is the only thing that makes the guard re-fire).

`useDepositFlow.ts` had the identical bypass — the inline read-after-mine check shared that catch — so it moved out of the cleanup block too.

Two follow-ons: `mapDepositError` was rendering key drift as "Protocol parameters changed", which invites exactly the retry that cannot help, so it now has its own copy explaining no BTC was spent and a new deposit is needed. And there's a regression test asserting the record survives drift.

services/vault/src/hooks/deposit/useVaultActions.ts
gbarkhatov: Agreed, and fixed in bc5b2679. The guard now sits outside the version block, gated only on `pendingPegin?.buildParticipantOperationKeys` — a missing build version says nothing about whether an operator rotated.

Your phrasing — "any missing version silently disables the check" — is what made this concrete: the fail-open was invisible because the conditions read as a single 'do we have local state' test when they are really five independent ones.

Test added: the guard runs when the stamp is present but build versions are absent.

The second half of the same block turned out to be worse than fail-open — see the `:330` thread.

services/vault/src/services/vault/vaultPayoutSignatureService.ts
gbarkhatov: Valid, and fixed in bc5b2679. This was listed as known issue 1 in the description; your framing — an indexer change with no deploy on our side and no user workaround — is what moved it from 'known' to 'fix now'.

`resolveVaultProviderBtcPubkey` now applies the same accept-either policy `validateOnChainParticipantKeys` uses on the deposit path: the hint is accepted against the registration key **or** `getCurrentOperationBtcKey`. The second read is a fallback, not an unconditional extra RPC — it only runs when the hint fails the registration comparison, so the un-rotated path is unchanged.

It still returns the registration key, since that is only the genesis fallback for epoch resolution; the keys actually signed with come from `resolveParticipantKeysAtEpochs`. The hint never influences the result in either path, which is what makes accept-either safe rather than a weakening — it can only cause us to *not* reject.

Three tests: accepts a hint matching the rotated operation key, still rejects one matching neither, and does not read the operation key when the hint matches registration.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/validateOnChainParticipantKeys.ts
gbarkhatov: Partly addressed by code that landed after this review, partly by bc5b2679.

**The window is much narrower than when you reviewed this.** Each role is now compared against *both* candidate sets, so a role that never rotated matches both and constrains nothing. The scenario you describe — VP role advanced, keeper role not — no longer throws unless **both** roles have actually rotated *and* the indexer is split between them. A single-role rotation mid-propagation passes cleanly.

**The decision is explicit, and it is to fail closed.** When two roles can each *only* be explained by different key sets, the indexer is serving a view no single snapshot ever held, and building from it would mean constructing a lock from a key set that never existed on chain. Comparing whole sets per role rather than testing union membership is the same choice one level down — a keeper set holding one registration key and one operation key is an inconsistency, and union membership would wave it through.

**Telemetry was the real gap, and you were right about it.** `onIndexerServingOperationKeys` only fired on the healthy path, so the blocking case was invisible to us and surfaced only as user reports. bc5b2679 adds `onIndexerHintsInconsistent`, called immediately before the throw and wired to `logger.error` in `useDepositFlow`. It is reported as an error rather than info deliberately: it blocks every deposit for the provider and clears only when the indexer converges, so "Refresh and try again" genuinely cannot help — exactly your point.

packages/babylon-ts-sdk/src/tbv/core/contracts/abis/BTCVaultRegistry.abi.ts
gbarkhatov: You're right and the comment was wrong. Corrected in bc5b2679.

I had generalised `getCurrentOperationBtcKey`'s genesis fallback to `getPayoutScriptAtEpoch`, which does not behave that way. A payout script is mandatory at registration and stored as v1, so on an RFC-006 registry the getter returns the operator's own registered script from the very first epoch. The contract's BIP-86 branch is a migration default for pre-RFC-006 rows, not reachable from registration — your `VPKeyRegistryLogic.sol:423-425` citation says so explicitly.

The comment now scopes the no-op claim to `getCurrentOperationBtcKey` only, states the payout getter's real behaviour, and notes that an operator whose registered script happens to equal `bip86(key)` is a coincidence of what it chose to register rather than a fallback.

**Your last sentence is the part that changed my model of the feature.** Run 2's failure was the normal case, not devnet having an unusual custom script — which means the BIP-86 payout destination is not a transition state that resolves on its own. A payout graph is built once at BaBe Setup and never rebuilt, so any vault whose graph predates its network's #2440 upgrade pays BIP-86 permanently. That is why `acceptedPayoutScriptHexes` accepts both forms; accepting only the registered one would make every such vault permanently unsignable with BTC already committed. Removal conditions are documented on the function and in #2182. The ABI comment now points at that connection so the next reader doesn't re-derive it.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/standardPayoutScript.ts
gbarkhatov: The parity claim was false and is removed in bc5b2679. Thank you for the `payout_scripts.rs:178-190` pointer — `vaultd` uses registry scripts verbatim and documents precisely why, which is the opposite of what I'd written.

Keeping the check, as you suggested it might be right, but the docs now justify it on its own terms rather than by borrowing `vaultd`'s authority:

- We are the party being asked to **pre-sign** against these bytes, and a depositor signature pinned to a provably unspendable output is not recoverable. `vaultd` occupies a different position and its reasoning — that a predicate the contract doesn't apply would block every peg-in for that operator — is a liveness argument that doesn't transfer to a signer.
- It is **load-bearing for the fee band**. Capping a standard output at 34 bytes is what let #2169's now-unreachable `out1Len [1,128]` cap come out during the rebase. Relaxing this check would require revisiting that.

The deliberate consequence is now stated rather than glossed: an operator registering a non-standard script via `--allow-non-standard` can have its graphs built by `vaultd` and still be refused here, blocking deposits against it in this dApp. Fail-closed by choice — the alternative is pre-signing into a destination we cannot verify is spendable.

packages/babylon-ts-sdk/src/tbv/core/services/participants/resolveParticipantKeys.ts
gbarkhatov: Overstated is right — "nothing on-chain prevents" was simply wrong. Corrected in bc5b2679 with the contract guarantees named accurately: intra-role uniqueness enforced in all three registries on rotation as well as registration, plus VP rotation rejecting keys held in the current VK/UC rosters.

The comment now names what is genuinely unguarded, which is also what the check earns its place on: a keeper or challenger rotating **onto** the VP's key — the asymmetry you identified, since the VP-side check runs at VP rotation time and not again when a keeper moves — plus keeper↔challenger collisions and keepers colliding across applications.

Check unchanged; only the justification was wrong. It is cheap and runs before any script is constructed, and a duplicate would collapse two entries in the sorted key set and silently build a lock with the wrong participant count.

packages/babylon-ts-sdk/src/tbv/core/contracts/__tests__/BTCVaultRegistryKeyEpochs.abi.test.ts
```diff
@@ -0,0 +1,173 @@
...
+    // The hazard, stated exactly. `BTCVaultProtocolInfo` is a *dynamic* tuple
+    // (it contains `bytes`), so its encoding is a 13-word head followed by the
+    // tail holding the dynamic payloads. Asking for 16 head words does not run
+    // off the end of the returndata — it reads three words of *tail* and hands
+    // them back as epochs. No throw, no warning, plausible-looking small
+    // integers.
+    //
+    // So the extended ABI must never be read on a registry that has not been
+    // upgraded. Being careful with the values is not enough; the capability
+    // probe below is the only thing standing between this and resolving
```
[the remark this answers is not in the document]
gbarkhatov: Confirmed and fixed in bc5b2679. The probe was removed together with both feature flags, and this documentation was left pointing at it.

The finding was broader than the one site flagged — a sweep found **seven** stale references across five files:

- `clients/eth/types.ts` (x2)
- `clients/eth/vault-registry-reader.ts` (x2)
- `contracts/__tests__/BTCVaultRegistryKeyEpochs.abi.test.ts` (this one)
- `services/vault/src/clients/eth-contract/btc-vault-registry/query.ts`
- `services/vault/src/clients/eth-contract/sdk-readers.ts`

Your framing was the useful part: the docs implied a runtime guard where compatibility now rests on a deployment constraint. Each site now says that directly — every network this ships to already has the RFC-006 getters and mainnet is a fresh RFC-006 deploy — rather than referring to a probe that no longer exists.

The hazard itself is real and the test still pins it, so the test stays; only its explanation changed. The `uint64` range check in `vault-registry-reader.ts` is now explicitly described as a weak backstop that rejects wild values but not plausible small ones, so it isn't mistaken for the guarantee.

jonybur (commented):
Reviewed the full diff. The RFC-006 modelling is sound and several of the hard parts are handled better than I expected — flagging one blocking issue and three worth fixing before merge.

## What's right

- **The ABI quarantine.** Keeping the 16-field `getBtcVaultProtocolInfo` in its own file rather than extending the shared entry is the correct call, and the reasoning in the module doc is exactly right: `getProtocolInfoBatch` goes through `multicall({allowFailure: false})`, so a single bad decode breaks status/resume/payout/refund app-wide. The test that pins the **silent** mis-decode — not just the throwing one — is the part that makes this safe to maintain.
- **Pairs as source of truth**, sorted arrays derived from them, with a fixture whose rotated keeper key deliberately sorts before its original. That is precisely the failure mode rotation introduces and it would have been easy to miss.
- **`ParticipantKeyDriftError` as a sibling of `RegisteredVaultVersionMismatchError` rather than a subclass**, with deliberately divergent cleanup. This is the subtlest decision in the PR — subclassing would have kept satisfying the type guard and turned a clean no-loss abort into BTC locked until the refund timelock.
- **Dual-accept is correctly scoped**: both candidates derived from chain state, never from the VP's response, with the commission value cap retained independently.

## 1. Blocking — the refund path still compares the indexer key against the registration key only

`services/vault/src/services/vault/vaultRefundService.ts:436-443`

```ts
const indexerXOnly = processPublicKeyToXOnly(vaultProvider.btcPubKey).toLowerCase();
if (indexerXOnly !== vaultProviderOnChainPubkey.toLowerCase()) {
  throw new Error(`Indexer vault provider key for ${vault.vaultProvider} does not match on-chain registry. Aborting refund.`);
}
```

This PR relaxes exactly this comparison in both sibling paths — `validateOnChainParticipantKeys` (deposit) and `resolveVaultProviderBtcPubkey` (payout signing) — and the description states the reason plainly:

> Comparing only against the registration key would hard-fail payout signing for every depositor of a rotated provider the day the indexer starts serving operation keys — an indexer-side change, with no deploy on our side and no user workaround.

That argument applies verbatim here, and refund is the recovery path of last resort. devnet's VP is already rotated to epoch 13, so this fires the moment the indexer catches up: `Aborting refund` for every vault of that provider, with no other way out of the dApp.

The hint supplies no key material in any of the three sites — the resolved key comes from chain either way — so accepting the current operation key alongside the registration key costs nothing. Suggest extracting the dual-accept into one shared helper and calling it from all three, so they cannot drift apart again.

## 2. Stale JSDoc on `assertPayoutOutputLayout` — in a critical-path file

`packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts:484-506`

The doc block now contradicts the code in three places, all of them changed by this PR:

- *"`out0Len` from the pinned outs[0] script (registered payout script, or derived BIP-86 for the VK-claimer role — see #2150 Stage 6 for the pending registered-script switch)"* — the VK-claimer branch no longer derives BIP-86 as its expectation, and that "pending" switch is this PR.
- *"`out1Len` measured from the deliberately unpinned VP-claimer commission output"* — it is pinned now.
- *"Both are rejected outside the contract's registration cap `[1, 128]`"* — the `out1Len` cap was deleted here (correctly — `assertStandardPayoutScript` makes it unreachable).

The paragraph above also reads *"when `vpCommissionScriptPubKey` is supplied it is pinned too"*, but it is a required param with no absent case.

Same class of leftover at `validateOnChainParticipantKeys.ts:118`: *"Resolve the current operation keys, **if enabled**"* — there is no enable/disable any more.

Raising these as more than cosmetics because this is the file the two required code owners are asked to read, and the doc is the first thing they will read. The description says a stale-prose sweep covered five files; these were missed.

## 3. `verifyResumeParticipantKeys.ts` has no tests

Acknowledged as #2185, but I'd weight it higher than "test hardening": this function is the only thing standing between a rotated key set and a Pre-PegIn broadcast on the resume path. Its abort branch is unexercised on-chain (test D's premise turned out unreachable by construction), and the one test touching the surrounding behaviour — `useVaultActions.test.ts` — mocks the module wholesale, so the resolve-and-compare body has no coverage at all.

Worth noting one thing inside that uncovered body: the VP key is normalised through `processPublicKeyToXOnly().toLowerCase()` before comparison, while the keeper and challenger arrays are compared raw. Harmless today because `peginStorage` validates every stamped key against `/^[0-9a-f]{64}$/`, but the two comparisons have different tolerances, and the coupling is invisible from either file. Worth making uniform while adding the test.

## 4. `getPayoutScriptsAtEpochs` doesn't length-check its results

`packages/babylon-ts-sdk/src/tbv/core/clients/eth/operation-key-reader.ts:258-290`

`finalize` asserts the raw key arrays match the roster length; the payout-script read has no equivalent. A short array yields `undefined` map values, which does fail closed downstream via `payout.ts`'s `registered === undefined` throw — but the error then blames the claimer rather than the read that came up short. A length assertion mirroring `finalize`'s would name the real fault.

## Minor — partial-upgrade window

Worth recording since the module docs reason carefully about the fully-legacy case: a *fully* legacy registry is safe end to end. The epoch read returns garbage silently, but `resolveParticipantKeysAtEpochs` then calls `getOperationBtcKeyAtEpochOrGenesis`, which does not exist there, so the multicall reverts and the whole path fails closed. The residual gap is narrower than the docs imply — a registry carrying the operation-key getters but an un-extended `BTCVaultProtocolInfo` struct, where garbage epochs would resolve without error. No network is in that state, but it is the one case `assertEpochInRange` is genuinely load-bearing for, and it might be worth saying so where the docs currently describe the broader hazard.

## On dropping the feature flags

Agreed on both, and the reasoning holds up. The rotation flag's off-state was already the broken setting on devnet and its capability probe's classifier was unreachable, so it guarded nothing real; and with dual-accept in place, payout flag-on is a strict superset of flag-off, which makes flag-off the trap.

The consequence to own is that rollback is now "revert the PR", and the deploy changes payout signing, refund, resume and VP auth simultaneously. Since the registry-version dependency is a deployment invariant rather than a runtime check, I'd verify the `getBtcVaultProtocolInfo` struct shape against each target registry before promoting to it, per network.

## On splitting

The description's own split proposal makes the case better than I can, and I'd second it. `payout.ts` and `signDepositorGraph.ts` carry ~200 lines of genuinely load-bearing change and would get a materially better review standing alone than sitting under 900 lines of ABI JSON — particularly since they need two approvals and everything else needs one. The ABI + reader half is a no-op and could land immediately in parallel.

---

Not approving on account of item 1 — refund breaking for rotated providers is a recovery-path failure with BTC already committed. Items 2 and 3 I'd want in before merge; item 4 and the minor note are your call.

jrwbabylonlab (approved):
LGTM. let's ensure code is removed once protocol side rolled out after a week

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts
```diff
@@ -382,16 +408,95 @@ function assemblePayoutPsbt(
...
+ * Removal is therefore gated by the **last** network to upgrade, since this
+ * code is shared. Concretely: drop the BIP-86 branch only once devnet, testnet
+ * and mainnet have all been on #2440 for longer than the activation deadline,
+ * and no vault registered before those upgrades is still awaiting signatures.
+ * Deleting it earlier strands deposits with BTC already locked; there is no
+ * recovery path short of the refund timelock.
+ *
+ * @internal Helper invoked by {@link assertPayoutOutputLayout}.
+ */
+function acceptedPayoutScriptHexes(
```
jrwbabylonlab, line 468: Noting exactly what comes out later for #2182, so it's easy to find. Line numbers at `bc5b2679`, all in this file.

**Delete**
- `acceptedPayoutScriptHexes()` — L468-477
- `matchesAnyScript()` — L480-482 (only exists to check against a list)
- the `deriveBip86ScriptPubKeyHex` import — L22 (no other use in this file)

**Collapse back to one expected script**
- L572 — VK claimer: `acceptedPayoutScriptHexes(registered, claimer)` → `[stripHexPrefix(registered)]`
- L610-613 — VP commission: same, single script
- L586 and L615 — `matchesAnyScript(...)` → plain `.equals()`
- L540 — `acceptedOut0ScriptHexes: string[]` back to a single `string`

**Why:** these accept two destinations instead of one — the operator's registered script, plus the old BIP-86 default. The second is only needed for vaults whose payout graph was built before btc-vault#2440, and those graphs are never rebuilt. Once every network is past #2440 and no pre-#2440 vault is still in `PendingDepositorSignatures`, the BIP-86 branch is unreachable and this collapses back to what it was.

**Not this:** the accept-either in `vaultPayoutSignatureService` (registration key or current operation key) isn't transitional — that's permanent rotation support and stays.

Two notes: the gate is btc-vault#2440, not RFC-006 — easy to conflate since they landed together. And #2182 has no assignee, so this list currently has no owner.

PR #2176, opened 2026-08-04
fix(vault): stop reporting wallet cancellations and extension errors to Sentry

jonybur:
## What

Four changes that stop routine and third-party events being captured as errors:

- a real user-cancellation predicate, replacing an `includes("rejected")` substring test
- `ignoreErrors` for wallet/translation extension errors
- `notranslate` meta
- permanent HTTP statuses no longer retried or captured

## Why

Wallets express cancellation as rejected **and** cancelled / canceled / denied, plus WalletConnect's "Proposal expired". The old check caught one of them, so most cancellations — a depositor closing a popup — became error events, the single largest category in the project.

A 451 geo-block was retried on every poll cycle indefinitely, with two 30s-interval queries alternating for over an hour and each settled failure billing an issue.

The `removeChild` crashes come from translation extensions mutating React-owned text nodes; `notranslate` is the standard mitigation.

## How

`isUserCancellation` checks typed signals first (EIP-1193 4001, viem's `UserRejectedRequestError`, wallet-connector `CONNECTION_REJECTED`), then narrow wording, walking the `cause` chain. Applied at the wallet-connection boundary and again in `beforeSend`.

Deliberately narrow — these still report:

```
Failed to connect wallet: Connection timeout   # not a user action
The contract rejected this transaction         # deterministic failure
```

It lives in a dependency-free leaf module because `sentry.client.config.ts` imports it at init time; pulling in COPY or the SDK would put module-init surface ahead of `Sentry.init`.

22 new tests pinning both directions. Vault suite 3088 passes; lint clean; build succeeds.

## Not fixed here

A 451 still doesn't flip the app into its geo-blocked UI — the boot probe hits `/health`, which stays 200 while the data plane 451s. Downgrading the report doesn't worsen that, but doesn't fix it. Follow-up.

### DIFF
--- a/services/vault/index.html
+++ b/services/vault/index.html
@@ -5,6 +5,13 @@
     <link rel="icon" href="/favicon.ico" type="image/x-icon" sizes="32x32" />
     <link rel="canonical" href="%NEXT_PUBLIC_CANONICAL%" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <!--
+      Opt out of page translation. Translation extensions replace React-owned
+      text nodes with their own wrappers; React's next commit then calls
+      removeChild on a node that is no longer its child, which throws
+      NotFoundError and unmounts the whole app via the error boundary.
+    -->
+    <meta name="google" content="notranslate" />
     <meta property="og:title" content="Babylon - Vault" />
     <meta name="description" content="Vault Dashboard" key="desc" />
     <meta property="og:description" content="Vault Dashboard" />
--- a/services/vault/index.html
+++ b/services/vault/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" translate="no">
   <head>
     <meta charset="UTF-8" />
     <link rel="icon" href="/favicon.ico" type="image/x-icon" sizes="32x32" />
--- a/services/vault/sentry.client.config.ts
+++ b/services/vault/sentry.client.config.ts
@@ -13,6 +13,10 @@ import * as Sentry from "@sentry/react";
 import { v4 as uuidv4 } from "uuid";
 
 import { getCommitHash } from "@/config";
+import {
+  hasTranslationExtensionMarkers,
+  isReactDomMutationError,
+} from "@/utils/errors/domMutationErrors";
 import { isUserCancellation } from "@/utils/errors/userCancellation";
 import { redactData, scrubSentryEvent, scrubString } from "@/utils/telemetry";
 
@@ -23,10 +27,14 @@ const SENTRY_DEVICE_ID_KEY = "sentry_device_id";
  *
  * Every entry was confirmed absent from our source: they come from competing
  * wallet extensions racing over `window.ethereum` (we never write to it - the
- * connector uses AppKit/EIP-6963 discovery), from a wallet extension's own
- * message channel after it reloads mid-session, or from page-translation
- * extensions swapping React-owned text nodes (the `removeChild`/`insertBefore`
- * family). None is actionable and together they crowd out real reports.
+ * connector uses AppKit/EIP-6963 discovery), or from a wallet extension's own
+ * message channel after it reloads mid-session. None is actionable and together
+ * they crowd out real reports.
+ *
+ * `ignoreErrors` runs ahead of `beforeSend`, so anything listed here is dropped
+ * before a tag could rescue it. Only put errors here that our own code can
+ * never throw. The React DOM-mutation family is deliberately NOT here - it is
+ * classified in `beforeSend` instead, see `utils/errors/domMutationErrors.ts`.
  */
 const THIRD_PARTY_EXTENSION_ERRORS = [
   // Injected-provider collisions between wallet extensions
@@ -36,10 +44,15 @@ const THIRD_PARTY_EXTENSION_ERRORS = [
   // Extension messaging after the extension reloaded or was disabled
   "Could not establish connection. Receiving end does not exist.",
   "Extension context invalidated.",
-  // Page-translation extensions mutating nodes React owns
-  "The node to be removed is not a child of this node",
-  /Failed to execute '(removeChild|insertBefore)' on 'Node'/,
 ];
+
+/** The error text of an event, from either an exception or a message event. */
+function eventText(event: Sentry.ErrorEvent): string {
+  return (
+    event.exception?.values?.[0]?.value ??
+    (typeof event.message === "string" ? event.message : "")
+  );
+}
 
 const sentryDsn = process.env.NEXT_PUBLIC_SENTRY_DSN;
 
@@ -134,8 +147,29 @@ Sentry.init({
     // not a fault. Dropped here as well as at the call sites so a cancellation
     // reaching Sentry through a global handler (rather than logger.error) is
     // filtered too.
-    if (isUserCancellation(hint?.originalException)) {
+    //
+    // Exempt events the flow explicitly tagged as a partial failure. The
+    // multi-vault deposit path logs and then CONTINUES with the remaining
+    // vaults, so a depositor rejecting one prompt mid-batch leaves that vault
+    // short of recovery material while the deposit proceeds - CLAUDE.md
+    // critical-path #3 ("undersigning leaves recovery material missing for an
+    // active challenger"). The outer message does not read as a cancellation
+    // but the wrapped cause does, so an untagged drop swallows exactly the
+    // event worth keeping.
+    if (
+      isUserCancellation(hint?.originalException) &&
+      !event.tags?.partialFailure
+    ) {
       return null;
+    }
+
+    // React DOM-mutation failures: drop only with a translation extension
+    // actually present, otherwise keep at warning so our own reconciliation
+    // bugs stay visible and the opt-out's effect stays measurable.
+    if (isReactDomMutationError(eventText(event))) {
+      if (hasTranslationExtensionMarkers()) return null;
+      event.level = "warning";
+      event.tags = { ...event.tags, domMutation: "unattributed" };
     }
 
     event.extra = {
--- a/services/vault/sentry.client.config.ts
+++ b/services/vault/sentry.client.config.ts
@@ -13,9 +13,33 @@ import * as Sentry from "@sentry/react";
 import { v4 as uuidv4 } from "uuid";
 
 import { getCommitHash } from "@/config";
+import { isUserCancellation } from "@/utils/errors/userCancellation";
 import { redactData, scrubSentryEvent, scrubString } from "@/utils/telemetry";
 
 const SENTRY_DEVICE_ID_KEY = "sentry_device_id";
+
+/**
+ * Errors thrown by browser extensions in the page, not by this app.
+ *
+ * Every entry was confirmed absent from our source: they come from competing
+ * wallet extensions racing over `window.ethereum` (we never write to it - the
+ * connector uses AppKit/EIP-6963 discovery), from a wallet extension's own
+ * message channel after it reloads mid-session, or from page-translation
+ * extensions swapping React-owned text nodes (the `removeChild`/`insertBefore`
+ * family). None is actionable and together they crowd out real reports.
+ */
+const THIRD_PARTY_EXTENSION_ERRORS = [
+  // Injected-provider collisions between wallet extensions
+  "Cannot redefine property: ethereum",
+  /Cannot set property ethereum of .* which has only a getter/,
+  "trap returned falsish for property 'tronlinkParams'",
+  // Extension messaging after the extension reloaded or was disabled
+  "Could not establish connection. Receiving end does not exist.",
+  "Extension context invalidated.",
+  // Page-translation extensions mutating nodes React owns
+  "The node to be removed is not a child of this node",
+  /Failed to execute '(removeChild|insertBefore)' on 'Node'/,
+];
 
 const sentryDsn = process.env.NEXT_PUBLIC_SENTRY_DSN;
 
@@ -85,6 +109,8 @@ Sentry.init({
   // Setting this option to true will print useful information to the console while you're setting up Sentry.
   debug: false,
 
+  ignoreErrors: THIRD_PARTY_EXTENSION_ERRORS,
+
   beforeBreadcrumb(breadcrumb) {
     if (breadcrumb.message) {
       breadcrumb.message = scrubString(breadcrumb.message);
@@ -104,6 +130,14 @@ Sentry.init({
   // redaction hook (URL + href masking).
 
   beforeSend(event, hint) {
+    // A depositor declining or dismissing a wallet prompt is routine drop-off,
+    // not a fault. Dropped here as well as at the call sites so a cancellation
+    // reaching Sentry through a global handler (rather than logger.error) is
+    // filtered too.
+    if (isUserCancellation(hint?.originalException)) {
+      return null;
+    }
+
     event.extra = {
       ...(event.extra || {}),
       version: getCommitHash(),
--- a/services/vault/src/config/queryClient.ts
+++ b/services/vault/src/config/queryClient.ts
@@ -1,9 +1,41 @@
 import { MutationCache, QueryCache, QueryClient } from "@tanstack/react-query";
 
 import { logger } from "@/infrastructure";
+import { isUserCancellation } from "@/utils/errors/userCancellation";
 
 const calculateRetryDelay = (attemptIndex: number): number => {
   return Math.min(1000 * 2 ** attemptIndex, 30000);
+};
+
+/**
+ * HTTP statuses that will not change by trying again with the same request.
+ * 451 is the one that hurt in production: a geo-block made every poll cycle
+ * issue four doomed requests, and two independent 30s-interval queries then
+ * alternated for over an hour, each settled failure billing a Sentry issue.
+ * 408 and 429 stay retryable, as does everything 5xx.
+ */
+const PERMANENT_HTTP_STATUSES = new Set([400, 401, 403, 404, 410, 451, 501]);
+
+/**
+ * Pull an HTTP status off the error shapes this app actually throws:
+ * graphql-request's `ClientError` (`response.status`) and the app's own
+ * `ApiError` (`status`).
+ */
+const httpStatusOf = (error: unknown): number | undefined => {
+  if (!error || typeof error !== "object") return undefined;
+
+  const { status, response } = error as {
+    status?: unknown;
+    response?: { status?: unknown };
+  };
+  if (typeof status === "number") return status;
+  if (typeof response?.status === "number") return response.status;
+  return undefined;
+};
+
+const isPermanentHttpError = (error: unknown): boolean => {
+  const status = httpStatusOf(error);
+  return status !== undefined && PERMANENT_HTTP_STATUSES.has(status);
 };
 
 const shouldRetry = (failureCount: number, error: Error): boolean => {
@@ -11,11 +43,18 @@ const shouldRetry = (failureCount: number, error: Error): boolean => {
     return false;
   }
 
-  if (error.message?.includes("rejected")) {
+  if (isPermanentHttpError(error)) {
     return false;
   }
 
-  if (error.message?.includes("User rejected")) {
+  if (isUserCancellation(error)) {
+    return false;
+  }
+
+  // Kept deliberately broad for the retry decision only: any rejection wording
+  // (including a deterministic contract rejection, which `isUserCancellation`
+  // intentionally does not match) reproduces on retry, so re-sending is waste.
+  if (error.message?.includes("rejected")) {
     return false;
   }
 
@@ -53,6 +92,21 @@ export function reportQueryCacheError(
     });
     return;
   }
+
+  // A permanent status is a property of where the user is, not a fault we can
+  // act on, and it repeats on every poll for the rest of the session. Recorded
+  // as a breadcrumb so it still attaches to any genuine error captured later.
+  const permanentStatus = httpStatusOf(error);
+  if (permanentStatus !== undefined && isPermanentHttpError(error)) {
+    logger.warn(
+      `Permanent ${kind.toLowerCase()} failure: HTTP ${permanentStatus}`,
+      {
+        detail: error.message,
+      },
+    );
+    return;
+  }
+
   logger.error(error, {
     data: { context: `React Query Error [${kind}]` },
   });
--- a/services/vault/src/config/queryClient.ts
+++ b/services/vault/src/config/queryClient.ts
@@ -1,6 +1,11 @@
 import { MutationCache, QueryCache, QueryClient } from "@tanstack/react-query";
 
 import { logger } from "@/infrastructure";
+import {
+  HTTP_GEO_BLOCKED,
+  httpStatusOf,
+  isError451,
+} from "@/utils/errors/types";
 import { isUserCancellation } from "@/utils/errors/userCancellation";
 
 const calculateRetryDelay = (attemptIndex: number): number => {
@@ -16,42 +21,16 @@ const calculateRetryDelay = (attemptIndex: number): number => {
  *
  * 501 is here deliberately. It is 5xx by number but it means "this server does
  * not implement this operation", which no amount of retrying changes.
- */
-const NON_RETRYABLE_HTTP_STATUSES = new Set([
-  400, 401, 403, 404, 410, 451, 501,
-]);
-
-/**
- * Statuses that are also not worth a standalone Sentry issue.
  *
- * Deliberately narrower than {@link NON_RETRYABLE_HTTP_STATUSES}: not retrying
- * and not reporting are different decisions. A geo-block (451) is a property of
- * where the user is and repeats every poll for the rest of the session, so it
- * is breadcrumbed. Everything else - a 400 from a malformed query, a 501 from a
- * frontend/backend version skew - is a real defect someone should see, so it
- * still reaches `logger.error` even though it is never retried.
+ * Deliberately narrow. 401/403/404 are NOT here, because on this app's
+ * endpoints they are routinely transient: a just-broadcast transaction is not
+ * indexed yet, a subgraph is mid-redeploy, an RPC provider returns 401/403 for
+ * a rate limit or an edge condition, and a rotated key clears its own 401.
+ * The in-fetch retries are what bridge that propagation lag. This predicate
+ * also governs mutations, so a status added here stops retrying ETH
+ * submissions too.
  */
-const UNREPORTABLE_HTTP_STATUSES = new Set([451]);
-
-/**
- * Pull an HTTP status off the error shapes that actually reach this handler:
- * graphql-request's `ClientError` (`response.status`) and viem's
- * `HttpRequestError` (`status`).
- *
- * Note the app's own `ApiError` does NOT reach here - every throw site is
- * caught inside `healthCheckService`.
- */
-const httpStatusOf = (error: unknown): number | undefined => {
-  if (!error || typeof error !== "object") return undefined;
-
-  const { status, response } = error as {
-    status?: unknown;
-    response?: { status?: unknown };
-  };
-  if (typeof status === "number") return status;
-  if (typeof response?.status === "number") return response.status;
-  return undefined;
-};
+const NON_RETRYABLE_HTTP_STATUSES = new Set([400, 410, 451, 501]);
 
 const isNonRetryableHttpError = (error: unknown): boolean => {
   const status = httpStatusOf(error);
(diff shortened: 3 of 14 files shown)


services/vault/src/config/queryClient.ts
```diff
@@ -1,21 +1,60 @@
...
 };
 
+/**
+ * HTTP statuses that will not change by trying again with the same request.
+ * 451 is the one that hurt in production: a geo-block made every poll cycle
+ * issue four doomed requests, and two independent 30s-interval queries then
+ * alternated for over an hour, each settled failure billing a Sentry issue.
+ * 408 and 429 stay retryable, as does everything 5xx.
+ */
+const PERMANENT_HTTP_STATUSES = new Set([400, 401, 403, 404, 410, 451, 501]);
```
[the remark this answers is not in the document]
jonybur: Checked this against the live indexer and the 400 case does not apply here: it returns HTTP 200 for GraphQL-level failures, including an unknown field, a syntax error, and invalid JSON. Those surface as a `ClientError` with `response.status === 200`, so they were never in the suppressed set and already reach `logger.error` - the deployment-skew case is the one that was not being hidden.

The 501 point was right though. Retrying and reporting were sharing one list, which is what made the comment contradict the code. Split them: `NON_RETRYABLE_HTTP_STATUSES` keeps its exact membership so retry behaviour is byte-identical, and a separate `UNREPORTABLE_HTTP_STATUSES` (451 only) now drives the Sentry downgrade. A 400 or 501 is still never retried but is captured as a real error again. Added tests pinning 400-captured, 501-captured, 501-not-retried, and a viem-shaped top-level `status`.

jrwbabylonlab (dismissed):

services/vault/src/utils/errors/userCancellation.ts
```diff
@@ -0,0 +1,80 @@
...
+ * bare `includes("rejected")` caught only one of them, which is why
+ * cancellations were the single largest category of captured errors.
+ *
+ * Deliberately narrow: each alternative anchors on wallet-interaction phrasing
+ * rather than the bare word, so a genuine failure like "transaction was
+ * rejected by the node" is still reported. Connection *timeouts* are excluded
+ * on purpose - those are not user actions and may be real defects.
+ */
+const USER_CANCELLATION_PATTERN =
+  /user (?:rejected|denied|cancell?ed)|rejected by the wallet|rejected the \w+ (?:approval|request)|connection (?:cancell?ed|rejected)|connection to .+ was (?:cancell?ed|rejected)|denied request signature|proposal expired/i;
```
gbarkhatov: **Over-matches genuine failures.** I ran this exact pattern against candidate strings:

| message | result |
| --- | --- |
| `The contract rejected the borrow request` | **suppressed** |
| `The node rejected the signing request` | **suppressed** |
| `RPC error: server rejected the transaction request` | **suppressed** |
| `Connection to indexer failed after the tx was rejected by the network` | **suppressed** |
| `user denied access to this resource` | **suppressed** |
| `Ledger: transaction rejected by the wallet policy` | **suppressed** |

Two arms carry no wallet subject:

- `rejected the \w+ (?:approval|request)` matches *any* actor rejecting *any* noun-request — contract, node, VP daemon.
- `connection to .+ was (?:cancell?ed|rejected)` lets `.+` span unrelated clauses, so any message containing "Connection to …" plus a later "was rejected" matches.

The PR body's claim that `The contract rejected this transaction` still reports holds only for that exact phrasing — swap in "the borrow request" and it's gone. A hardware-wallet *policy* rejection (Ledger with blind signing off, app/firmware skew) also reads as a user cancellation, and that's a real UX defect you'd want to see.

Also `proposal expired` is unanchored, and WalletConnect emits it for relay/pairing timeouts the user never saw.

Suggest requiring a wallet subject: anchor the `rejected the …` arm on `wallet|user|you`, bound the connection arm with `[^.;]+` instead of `.+`, and scope the bare `denied` to `user denied|denied by (?:the )?user`.

services/vault/src/utils/errors/userCancellation.ts
```diff
@@ -0,0 +1,80 @@
...
+ * a signing prompt. Mirrors `ERROR_CODES.CONNECTION_REJECTED` from
+ * `@babylonlabs-io/wallet-connector`. Inlined to avoid pulling the full
+ * wallet-connector bundle into this file (and its test transform); the
+ * constant is the public contract - if it ever changes upstream, this string
+ * must change too.
+ */
+export const WALLET_CONNECTION_REJECTED_CODE = "CONNECTION_REJECTED";
+
+/**
+ * Wallet cancellation wording.
```
gbarkhatov, line 32: **Missing the reversed word order.** Coverage is the entire point of this module, and these all currently return `false`:

- `Transaction rejected by user`
- `Action cancelled by user` (Trezor)
- `denied by the user` (Ledger APDU `0x6985`)
- `Request rejected`
- `User closed modal` (WalletConnect)

`(?:rejected|denied|cancell?ed) by (?:the )?user` picks up the first three. Worth grepping the Ledger/Trezor adapters in `packages/babylon-wallet-connector` for the exact wording before pinning them.

Nit on the doc: "at least four vocabularies" undercounts once the reversed forms are in.

services/vault/src/utils/errors/userCancellation.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+export function isUserCancellation(error: unknown): boolean {
+  // Some wallet adapters reject with a bare string rather than an Error.
+  if (typeof error === "string") return USER_CANCELLATION_PATTERN.test(error);
+
+  let cur: unknown = error;
+
+  for (
+    let depth = 0;
+    depth <= MAX_CAUSE_DEPTH && cur && typeof cur === "object";
```
gbarkhatov: **The `cause` walk drops string causes.** Lines 49-50 handle a bare string at the top level — per their own comment, "some wallet adapters reject with a bare string" — but this loop guard is `typeof cur === "object"`, so the walk stops the moment a `cause` *is* a string:

```ts
isUserCancellation(
  new Error("Failed to broadcast", { cause: "User rejected the request" }),
); // false
```

Given the module's stated premise, the string check belongs inside the loop, not only ahead of it.

services/vault/sentry.client.config.ts
```diff
@@ -104,6 +130,14 @@ Sentry.init({
   // redaction hook (URL + href masking).
 
   beforeSend(event, hint) {
+    // A depositor declining or dismissing a wallet prompt is routine drop-off,
+    // not a fault. Dropped here as well as at the call sites so a cancellation
+    // reaching Sentry through a global handler (rather than logger.error) is
+    // filtered too.
+    if (isUserCancellation(hint?.originalException)) {
```
gbarkhatov: **This global drop silences partial-failure telemetry on the multi-vault deposit path.**

`hooks/deposit/useDepositFlow.ts` logs and then *continues with the remaining vaults* in two places:

- WOTS submission — `logger.error` with context `[Multi-Vault] Failed to submit WOTS key for vault`, then the vault is added to `wotsFailedVaultIds` and the loop moves on.
- Payout sign/submit — `logger.error` with context `[Multi-Vault] Failed to sign or submit payouts for vault`, followed by a literal `// Continue with other vaults`.

A depositor rejecting one prompt mid-batch leaves that vault short of recovery material while the flow proceeds. And because `isUserCancellation` walks the `cause` chain, `Failed to sign or submit payouts for vault 2` wrapping a rejection is dropped too — the outer message doesn't match, the inner one does.

CLAUDE.md's critical-path #3 names this case explicitly: "undersigning leaves recovery material missing for an active challenger". That's precisely the event to keep.

Suggest exempting tagged events rather than filtering globally — add `tags: { partialFailure: "multi-vault" }` at those two `logger.error` sites and gate here:

```ts
if (isUserCancellation(hint?.originalException) && !event.tags?.partialFailure) {
  return null;
}
```

services/vault/sentry.client.config.ts
```diff
@@ -13,10 +13,34 @@ import * as Sentry from "@sentry/react";
...
+  // Injected-provider collisions between wallet extensions
+  "Cannot redefine property: ethereum",
+  /Cannot set property ethereum of .* which has only a getter/,
+  "trap returned falsish for property 'tronlinkParams'",
+  // Extension messaging after the extension reloaded or was disabled
+  "Could not establish connection. Receiving end does not exist.",
+  "Extension context invalidated.",
+  // Page-translation extensions mutating nodes React owns
+  "The node to be removed is not a child of this node",
+  /Failed to execute '(removeChild|insertBefore)' on 'Node'/,
```
gbarkhatov: **This blinds the app-crash path.** `components/pages/global-error.tsx` reports every boundary crash via `logger.error(error, { tags: { source: "GlobalErrorBoundary" } })`, and inbound filters run ahead of `beforeSend`, so these are dropped before any tag can rescue them. Per the comment above, these are exactly the errors that unmount the whole app.

Two consequences:

1. React throws the same `NotFoundError` when *our* tree has a reconciliation bug — portal misuse, a `createPortal` container removed out from under it. Those go invisible along with the extension noise.
2. There's no longer any way to confirm the `notranslate` change worked. The metric that would prove it is the one being filtered.

Prefer keeping capture at `level: "warning"` with a `thirdParty: "translation-extension"` tag, or gate the drop on actually detecting a translation extension — `document.documentElement` picks up `translated-ltr` / `_msttexthash` when one is active. Then you can watch the rate fall and remove the filter on evidence.

Nit: Safari's wording for the same DOM failure is `The object can not be found here.`, which neither entry covers.

services/vault/index.html
```diff
@@ -5,6 +5,13 @@
     <link rel="icon" href="/favicon.ico" type="image/x-icon" sizes="32x32" />
     <link rel="canonical" href="%NEXT_PUBLIC_CANONICAL%" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <!--
+      Opt out of page translation. Translation extensions replace React-owned
+      text nodes with their own wrappers; React's next commit then calls
+      removeChild on a node that is no longer its child, which throws
+      NotFoundError and unmounts the whole app via the error boundary.
+    -->
+    <meta name="google" content="notranslate" />
```
gbarkhatov, line 14: **This only covers Google Translate** (Chrome's built-in and translate.google.com). It does nothing for Edge/Bing Translate, Firefox's translator, Safari, DeepL, or Immersive Translate — and those are the extensions the comment above is describing.

`<html lang="en">` is right there on line 2; `translate="no"` is one attribute and it's the HTML-standard signal that Edge and Firefox honour:

```html
<html lang="en" translate="no">
```

Worth adding both. As it stands the root-cause fix is partial while the `ignoreErrors` entry hides the evidence that it's partial.

services/vault/src/config/queryClient.ts
```diff
@@ -1,21 +1,80 @@
...
+ * HTTP statuses that will not change by trying again with the same request.
+ * 451 is the one that hurt in production: a geo-block made every poll cycle
+ * issue four doomed requests, and two independent 30s-interval queries then
+ * alternated for over an hour, each settled failure billing a Sentry issue.
+ * 408 and 429 stay retryable.
+ *
+ * 501 is here deliberately. It is 5xx by number but it means "this server does
+ * not implement this operation", which no amount of retrying changes.
+ */
+const NON_RETRYABLE_HTTP_STATUSES = new Set([
```
gbarkhatov: **401/403/404 aren't obviously permanent here.**

- **404** is routinely transient on the endpoints this app polls — a just-broadcast tx isn't indexed yet, a subgraph is mid-redeploy. `refetchInterval` still recovers, but you lose the in-fetch retries that were bridging propagation lag.
- **401/403** — some RPC providers return these for transient rate-limit or edge conditions, and behind key rotation a 401 clears on its own.
- `shouldRetry` governs **mutations** too (`defaultOptions.mutations.retry` below), so this also stops retrying ETH submissions on those statuses.

The set actually justified by the 451 story in the comment above is `{400, 410, 451, 501}`. If you want the rest, consider making the set overridable per query rather than one global list.

services/vault/src/config/queryClient.ts
```diff
@@ -1,21 +1,80 @@
...
+
+/**
+ * Pull an HTTP status off the error shapes that actually reach this handler:
+ * graphql-request's `ClientError` (`response.status`) and viem's
+ * `HttpRequestError` (`status`).
+ *
+ * Note the app's own `ApiError` does NOT reach here - every throw site is
+ * caught inside `healthCheckService`.
+ */
+const httpStatusOf = (error: unknown): number | undefined => {
```
gbarkhatov: **No range check, and top-level `status` wins over `response.status`.** Any error object carrying an unrelated numeric `status` — a contract receipt status, a socket state, a wrapped daemon status — reads as HTTP. Cheap hardening:

```ts
const isHttpStatus = (v: unknown): v is number =>
  Number.isInteger(v) && (v as number) >= 100 && (v as number) <= 599;
```

and prefer `response.status` when both are present, since that's the more specific shape.

Separately: `isError451` already exists at `utils/errors/types.ts:124` and is live in `healthCheckService.ts:51`. That's now a second 451 predicate, reading only the top-level `status`. Extending that one to also read `response.status` and reusing it here would avoid the drift.

services/vault/src/config/queryClient.ts
```diff
@@ -1,21 +1,80 @@
...
+
+  if (isUserCancellation(error)) {
     return false;
   }
 
-  if (error.message?.includes("User rejected")) {
+  // Kept deliberately broad for the retry decision only: any rejection wording
+  // (including a deterministic contract rejection, which `isUserCancellation`
+  // intentionally does not match) reproduces on retry, so re-sending is waste.
+  if (error.message?.includes("rejected")) {
```
gbarkhatov: Nit: subsumed by `isUserCancellation` above for the whole "rejected" vocabulary — only the cancelled/denied/expired arms of the predicate add anything these two rules don't share, so it reads as two rules doing one job. Simpler:

```ts
if (isUserCancellation(error) || error.message?.includes("rejected")) {
  return false;
}
```

with the existing comment kept as-is.

services/vault/src/config/queryClient.ts
```diff
@@ -53,6 +112,21 @@ export function reportQueryCacheError(
...
+
+  // An unreportable status is a property of where the user is, not a fault we
+  // can act on, and it repeats on every poll for the rest of the session.
+  // Recorded as a breadcrumb so it still attaches to any genuine error captured
+  // later. Note this set is NOT the non-retryable set: a 400 or 501 is equally
+  // pointless to retry but is a real defect, so it falls through to
+  // `logger.error` below.
+  const status = httpStatusOf(error);
+  if (status !== undefined && UNREPORTABLE_HTTP_STATUSES.has(status)) {
+    logger.warn(`Unreportable ${kind.toLowerCase()} failure: HTTP ${status}`, {
```
gbarkhatov: **Breadcrumb-only means a geo-blocked session produces no signal at all.** `logger.warn` is `addBreadcrumb`, so it only ships if some *other* error is captured in the same session. Combined with the "Not fixed here" note — a 451 doesn't flip the app into the geo-blocked UI because `/health` stays 200 — a user in a blocked region gets a broken app and you never learn it happened. The old behaviour was noisy but observable.

It's also self-defeating at volume: two 30s-interval queries at four requests per cycle blow through Sentry's default `maxBreadcrumbs: 100` in roughly 25 minutes, evicting exactly the context the "still attaches to any genuine error captured later" rationale depends on.

Suggest one deduped `logger.event("Geo-blocked data plane (HTTP 451)", { level: "warning" })` per session (or per query key) with a stable fingerprint. One issue per session is not one per settled query, and it keeps the signal alive until the geo-block UI lands.

Minor, same lines: for a graphql-request `ClientError`, `error.message` stringifies the whole response *and* request, so `detail` can be large and can carry query variables (BTC/ETH addresses). `scrubString` mitigates it, but the status plus the query key would be smaller and safer.

services/vault/src/utils/errors/formatting.ts
```diff
@@ -31,7 +26,7 @@ export function isWalletRejectionError(error: unknown): boolean {
 
 /** EIP-1193 provider error codes used by the classifier below. */
 const EIP1193 = {
-  USER_REJECTED: 4001,
+  USER_REJECTED: EIP1193_USER_REJECTED,
```
gbarkhatov: **There are now two definitions of "user rejected".** `classifyError` below (~L141-158) already walks the `cause` chain to depth 10 checking the same three typed signals — `EIP1193.USER_REJECTED`, `name === "UserRejectedRequestError"`, `isWalletRejectionError` — and returns `"user-rejection"`. One drives UI copy, one drives telemetry, and they can drift.

They already have. `Connection to Keystone was canceled` and `Proposal expired` are now dropped from Sentry, but `classifyError` still returns `null` for both, so the depositor gets generic error copy instead of `COPY.common.classifiedErrors.userRejection`. That's the same bug this PR is fixing, on the UX side. `depositErrors.ts:186-190` is a third wording list (`user rejected` / `user denied` / `user cancelled` / `user canceled`).

The dependency-free-leaf argument is sound — the fix is to move the typed-signal walk into `userCancellation.ts` and have `classifyError` and `depositErrors` call it, which also deletes two wording lists.

Nit while you're here: `USER_REJECTED: EIP1193_USER_REJECTED` is an alias of an alias — use the import directly at the two call sites.

gbarkhatov (dismissed):
Reviewed the full diff at `4ec4a15`. The diagnosis is right, the comments explain *why* rather than *what*, and the negative test cases (`Connection timeout`, `The contract rejected this transaction`) are the right ones to pin — they're exactly what a lazier regex would eat. Approving; none of the below leaves the app worse than `main`, but three are worth resolving before you trust the resulting Sentry volume.

**Verified correct as claimed:** `logger.warn` really is a breadcrumb; `httpStatusOf` covers both shapes that reach the handler (graphql-request `response.status`, viem top-level `status`); 408/429/5xx stay retryable; the 501 exclusion is justified and pinned; the self-referencing `cause` chain terminates; and the module-init-order argument for a dependency-free leaf is real, not theoretical.

**Three I'd fix before relying on this:**

1. The cancellation regex suppresses genuine failures — I ran the literal pattern against candidate strings, results inline on `userCancellation.ts`. Four realistic non-user messages are dropped silently.
2. The global `beforeSend` drop silences the multi-vault partial-failure path in `useDepositFlow.ts`, which is CLAUDE.md critical-path #3 territory.
3. A 451 now emits *zero* telemetry (breadcrumb-only) while the geo-block UI still doesn't fire — that trades noise for blindness on a condition the PR itself says is unresolved.

**Not anchorable to a diff line:**

- The "What" section still says permanent statuses are "no longer retried or captured". Only 451 is uncaptured — worth correcting so the next reader doesn't assume 400/501 went silent too.
- `sentry.client.config.ts` sits outside `src/`, so `beforeSend` and `ignoreErrors` — the highest-leverage new behaviour in the PR — have no test at all. The predicate is well covered; the wiring isn't. Extracting the options object (`export const sentryOptions = {...}; Sentry.init(sentryOptions)`) would make both testable without disturbing init order.
- `formatting.ts` inlines user-facing strings in `formatErrorMessage` ("Transaction was rejected", "Insufficient balance for this transaction") and `formatPayoutSignatureError` ("Signing rejected"). Pre-existing, but CLAUDE.md asks for migration to `copy.ts` when the file is being edited anyway.

services/vault/src/utils/errors/userCancellation.ts
jonybur: Fixed in `58f7e05dc`. All four reversed forms added — `(?:rejected|denied|cancell?ed) by (?:the )?user` covers `Transaction rejected by user`, Trezor's `Action cancelled by user` and Ledger's `denied by the user`, plus a separate `user closed (?:the )?modal` arm for WalletConnect. Pinned in `userCancellation.test.ts`.

Left out `Request rejected` — with no subject it is the same class of over-match as the arm above, and I could not find it in the connector. `packages/babylon-wallet-connector/src/error/index.ts:34-37` carries the same four-wording list (`user rejected` / `user denied` / `user cancelled` / `user canceled`), and `error/codes.ts` has `CONNECTION_REJECTED` / `CONNECTION_CANCELED` — both coded paths are already covered by the typed signals.

Doc updated: it no longer claims a count.

services/vault/src/utils/errors/userCancellation.ts
jonybur: Fixed in `58f7e05dc`, and your table is now the test case — the three `<actor> rejected the <noun> request` strings are pinned as *reporting* in `userCancellation.test.ts`.

- The `rejected the …` arm is anchored on `wallet|user|you`. It is kept rather than deleted because it backs a captured production wording, `Unisat Wallet rejected the deriveContextHash approval`.
- The connection arm bounds the name to at most three words (`connection to \w+(?: \w+){0,2} was …`) instead of `.+`. `[^.;]+` was not enough — your `Connection to indexer failed after the tx was rejected by the network` has no `.` or `;` in it and still matched. Also pinned.

Two I did **not** change, both deliberate:

**`rejected by the wallet` stays.** You are right that it catches a Ledger policy rejection. But it also backs `ContractError: borrow from Aave Core position was rejected by the wallet.` — a captured wording where the user *did* decline, already in the test file. The two are not separable by wording, so this is a straight trade between missing a blind-signing defect and re-reporting a real cancellation. Happy to drop it if you would rather have the defect; say so and I will, in this PR.

**`user denied` stays**, so `user denied access to this resource` is still suppressed. It is MetaMask's canonical wording (`User denied transaction signature`) and scoping it further loses the single most common real cancellation.

`proposal expired` I also left. Your read is right that WalletConnect emits it for relay timeouts, but I could not find a distinguisher in the string, and it is not obvious to me which way that should fail — that is a product call rather than a code one, so I would rather you make it.

services/vault/src/utils/errors/userCancellation.ts
jonybur: Fixed in `58f7e05dc`. The string check moved inside the loop and the guard is now `cur != null`, with a `typeof cur !== "object"` bail *after* the frame check rather than before it. Your exact repro is the test:

```ts
isUserCancellation(new Error("Failed to broadcast", { cause: "User rejected the request" })); // now true
```

The self-referencing-cause test still passes, so the depth bound still terminates.

services/vault/sentry.client.config.ts
jonybur: Fixed in `58f7e05dc`, taking your suggested shape. Both sites now pass `tags: { partialFailure: "multi-vault" }` and `beforeSend` gates on it:

```ts
if (isUserCancellation(hint?.originalException) && !event.tags?.partialFailure) {
  return null;
}
```

Your point about the wrapped cause is the part that made this worse than it looked — `Failed to sign or submit payouts for vault 2` does not match, the inner rejection does, so the walk was what swallowed it. The comment at the filter now names critical-path #3 explicitly so the next person to widen this filter sees why the exemption is there.

services/vault/sentry.client.config.ts
jonybur: Fixed in `58f7e05dc`, and this was the most consequential of the eleven — `ignoreErrors` running ahead of `beforeSend` is exactly why no tag could rescue the boundary crashes.

The DOM-mutation family is out of `ignoreErrors` entirely and into a new `src/utils/errors/domMutationErrors.ts`, classified in `beforeSend`:

- Translation extension detected → dropped.
- Not detected → kept at `level: "warning"` with `domMutation: "unattributed"`.

I took both of your suggestions rather than one: gating on detection is what removes the noise, and the warning-level fallback is what keeps our own reconciliation bugs and the opt-out metric visible. Detection is `translated-ltr` / `translated-rtl` on `<html>` (Google, incl. Chrome built-in) and `font[_msttexthash]` (Edge/Bing). Firefox and Safari expose no comparable marker, so those fall to the keep branch — biased that way on purpose, since a miss costs a warning rather than a hidden crash.

Safari's `The object can not be found here.` added.

Split into its own module rather than left inline because `sentry.client.config.ts` runs `Sentry.init` on import and is therefore untestable; the predicates now have 12 tests in `domMutationErrors.test.ts`. Same reason `userCancellation.ts` is already factored out.

services/vault/index.html
jonybur: Fixed in `58f7e05dc` — `<html lang="en" translate="no">`. You are right that the meta tag alone was Google-only while the `ignoreErrors` entry hid the evidence of that; with the filter change on the other thread, the residual rate is now observable at warning level, so we can actually see whether this covers Edge and Firefox in practice. Kept the meta tag alongside it.

services/vault/src/config/queryClient.ts
jonybur: Fixed in `58f7e05dc` — narrowed to exactly the set you justified, `{400, 410, 451, 501}`.

The mutation point is the one that decided it: this predicate is wired into `defaultOptions.mutations.retry` too, so a 401 from a rate-limited RPC was killing ETH submission retries, which is a much worse failure than an extra doomed poll. `401/403/404` retryable again, pinned by test, with the reasoning (unindexed tx, subgraph mid-redeploy, provider rate limit, key rotation) in the constant's doc so it does not get re-widened by someone reading only the 451 story.

Did not add the per-query override — no call site needs one yet, and it would be speculative. Easy to add when one does.

services/vault/src/config/queryClient.ts
jonybur: Fixed in `58f7e05dc`, both parts, and I took the drift fix rather than duplicating.

`httpStatusOf` and `isError451` now live together in `utils/errors/types.ts`; `queryClient.ts` imports them and its local copy is gone. `isError451` reads `response.status` as well as the top-level `status`, so the `healthCheckService` call site gets the graphql-request shape it was previously missing — that widening looks correct to me but it is a behaviour change on a file outside this PR, so worth a second look.

Range check is your snippet. Precedence flipped to `response.status` first, as you suggested. Two tests pin it: a receipt-shaped `status: 1` is not read as HTTP, and `{ status: 200, response: { status: 451 } }` resolves to 451.

services/vault/src/config/queryClient.ts
jonybur: Done in `58f7e05dc` — folded into one condition exactly as written, comment kept.

```ts
if (isUserCancellation(error) || error.message?.includes("rejected")) {
  return false;
}
```

services/vault/src/config/queryClient.ts
jonybur: Fixed in `58f7e05dc`. One captured event per session, as suggested:

```ts
if (isError451(error)) {
  if (!geoBlockReported) {
    geoBlockReported = true;
    logger.event("Geo-blocked data plane (HTTP 451)", { level: "warning", tags: { httpStatus: "451", kind }, source });
  }
  return;
}
```

The breadcrumb-eviction arithmetic is what settled it — a signal that destroys its own supporting context within 25 minutes is worse than either alternative. No explicit fingerprint: `logger.event` wraps `captureMessage` with no scope callback, and the message string is constant so Sentry groups on it anyway. Adding fingerprint plumbing to the logger felt like more surface than this needs; say the word if you would rather have it.

Your "minor" is fixed too, and I think it was the more important half. `error.message` is out of the payload entirely — a `ClientError` stringifies request *and* response, so the query variables (BTC/ETH addresses) were riding into telemetry with `scrubString` as the only thing in the way. It now sends the status plus the query key (`query.queryHash`, or the mutation key), threaded through a new `source` param. Test asserts an address in the message does not reach the event.

Note this does not fix the underlying gap you flagged — a 451 still does not flip the app into geo-blocked UI, because `/health` stays 200. This just makes it visible until that lands.

services/vault/src/utils/errors/formatting.ts
jonybur: Fixed in `58f7e05dc` — full consolidation, both wording lists deleted.

`userCancellation.ts` is now the single definition and everything calls into it:
- `isWalletRejectionError` moved there (it only needed the code constant, and the module is dependency-free by design).
- `classifyError`'s three-signal block is now one `isUserCancellationFrame(cur)` call.
- `depositErrors.ts` step 6's four-string list is now `isUserCancellation(err)`.

One thing worth flagging in the design. I could not have `classifyError` call `isUserCancellation` directly: that walks the cause chain itself, so an inner cancellation at depth 5 would have beaten an outer `insufficient-funds` at depth 0 and quietly broken the documented precedence (b). So there are two exports — `isUserCancellationFrame` (single frame, no walk) for the classifier, and `isUserCancellation` (walks) for telemetry and the deposit mapper. Both tested, including a case asserting the frame predicate does *not* reach into causes.

Your drift examples are now tests: `Connection to Keystone was canceled` and `Proposal expired` both classify as `user-rejection` and render `COPY.common.classifiedErrors.userRejection` instead of generic copy.

Alias-of-an-alias gone — `EIP1193.USER_REJECTED` removed, and with the frame predicate in place both call sites disappeared rather than needing the direct import.

gbarkhatov (approved):

Issue #1592, opened 2026-05-07
e2e: Single-vault deposit happy path


jonybur:
## Context
The single-vault deposit happy path is the primary user flow. It traverses every state transition in \`DEPOSIT_FLOW_STEPS\`: DERIVE_VAULT_SECRET → SIGN_PEGIN_BTC → SIGN_POP → SUBMIT_PEGIN → BROADCAST_PRE_PEGIN → SIGN_PAYOUTS → ARTIFACT_DOWNLOAD → ACTIVATE_VAULT → COMPLETED. A regression anywhere here is user-visible and irreversible (real BTC).

## Scope
**In:**
- Full sequence with the BTC + ETH mock wallets producing valid signatures at each prompt
- Stepper highlights advance in lockstep with \`currentStep\`
- Modal copy matches \`DEPOSIT_FLOW_STEPS[step].description.active\` at each phase
- Artifact download happens (file produced; content presence asserted, full schema validation in a follow-up if needed)
- Vault is active on-chain at the end

**Out:**
- Multi-vault batch (next ticket)
- Resume / interruption (separate ticket)
- Error paths (separate ticket)

## Code references
- [services/vault/src/hooks/deposit/depositFlowSteps/config.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/services/vault/src/hooks/deposit/depositFlowSteps/config.ts)
- [services/vault/src/hooks/deposit/useDepositFlow.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/services/vault/src/hooks/deposit/useDepositFlow.ts)
- [services/vault/src/components/simple/DepositSignContent.tsx](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/services/vault/src/components/simple/DepositSignContent.tsx)
- [services/vault/src/components/simple/DepositProgressView.tsx](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/services/vault/src/components/simple/DepositProgressView.tsx)
- [services/vault/src/services/vault/vaultActivationService.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/services/vault/src/services/vault/vaultActivationService.ts)

## Acceptance criteria
- [ ] Test asserts each of the 9 step transitions is observed (DOM data-attr or test-only event bus)
- [ ] Final on-chain state: vault is \`ACTIVE\` in the registry contract
[...]

PR #2180, opened 2026-08-04
test(ci): flag when a PR changes how the UI looks

jonybur:
> **Lockfile change here is safe** - audited against today's keyv/cacheable npm supply-chain attack. The `pnpm-lock.yaml` diff comes entirely from the 02:23 UTC commit, about 6.5 hours before the attack began at 09:00 UTC; the later 10:16 UTC commit touches CI scripts only. The three additions - `pixelmatch` 7.2.0, `pngjs` 7.0.0 and `@playwright/test` 1.56.1 - are not on the compromised list, were all published months before the attack, and declare no install hooks. Nothing was re-resolved; the diff is purely additive.

## What

CI now flags when a pull request changes how the UI looks.

It photographs 315 Storybook components and 12 vault screens, and posts a comment listing anything that changed, with before / after / diff images attached to the run.

Worth being precise about the vault half: those 12 screens are captured without a connected wallet, so most of them are the app shell plus an error card rather than a populated page. They still catch shell, nav and layout regressions, but they are not yet coverage of the connected states - #1592 is the unblock for that. The Storybook half is the real component coverage today.

## Why

Nothing catches visual regressions today. A stray colour, spacing or layout change ships unnoticed unless a reviewer happens to spot it by eye.

## How

Rather than storing "expected" screenshots in the repo, CI takes the pictures twice on every run: once for the branch this PR started from, once for the PR itself. Then it compares them.

That means there are no baseline images to keep up to date, and nothing to approve when a change is intentional.

The check is advisory for now - it comments, it doesn't block merging.

### DIFF
--- a/.github/workflows/visual-regression.yml
+++ b/.github/workflows/visual-regression.yml
@@ -82,6 +82,10 @@ jobs:
           cp -R packages/babylon-core-ui/visual "$H/core-ui-visual"
           cp packages/babylon-core-ui/playwright.visual.config.ts "$H/core-ui-playwright.visual.config.ts"
           cp scripts/visual-diff.mjs "$H/"
+          # The runner script itself. Easy to forget, and forgetting it fails
+          # the baseline side with a bare "No such file or directory": the
+          # script does not exist at the merge-base either.
+          cp .github/scripts/visual-capture.sh "$H/"
 
       - name: Capture candidate (PR head)
         env:
@@ -96,17 +100,25 @@ jobs:
           set -euo pipefail
           H="${{ runner.temp }}/harness"
           rm -rf services/vault/e2e/visual packages/babylon-core-ui/visual
-          mkdir -p services/vault/e2e packages/babylon-core-ui
+          mkdir -p services/vault/e2e packages/babylon-core-ui .github/scripts
           cp -R "$H/vault-visual" services/vault/e2e/visual
           cp "$H/playwright.visual.config.ts" services/vault/
           cp -R "$H/core-ui-visual" packages/babylon-core-ui/visual
           cp "$H/core-ui-playwright.visual.config.ts" packages/babylon-core-ui/playwright.visual.config.ts
+          cp "$H/visual-capture.sh" .github/scripts/visual-capture.sh
+          chmod +x .github/scripts/visual-capture.sh
 
-      # The merge-base may carry a different lockfile, so reinstall: the
-      # baseline must reflect the dependencies that commit actually shipped.
-      - name: Install dependencies at merge-base
-        run: pnpm install --frozen-lockfile --prefer-offline
-
+      # Deliberately NOT reinstalling here. The harness needs tooling this PR
+      # adds (core-ui gains @playwright/test), which the merge-base lockfile
+      # would prune - the baseline could then never run. Keeping HEAD's
+      # node_modules also means both sides are photographed by an identical
+      # instrument, which is the property this whole design is built on.
+      #
+      # The trade-off, stated plainly: the baseline renders merge-base SOURCE
+      # against HEAD's external npm packages. Workspace changes (core-ui,
+      # ts-sdk, wallet-connector) are still compared correctly because those
+      # are rebuilt from the checked-out source. A PR that only bumps an
+      # external UI dependency will not show a diff.
       - name: Capture baseline (merge-base)
         env:
           CAPTURE_ROOT: ${{ runner.temp }}/visual/baseline
@@ -115,14 +127,6 @@ jobs:
       - name: Restore PR head
         if: always()
         run: git checkout --force "${{ steps.refs.outputs.head_sha }}"
-
-      # MUST reinstall. The merge-base install above prunes any dependency
-      # this PR adds - including pixelmatch/pngjs, which the diff script
-      # imports. Restoring the files does not restore node_modules, so
-      # without this the diff dies with ERR_MODULE_NOT_FOUND every run.
-      - name: Reinstall dependencies for the diff step
-        if: always()
-        run: pnpm install --frozen-lockfile --prefer-offline
 
       - name: Compare baseline against candidate
         id: diff
--- a/.github/workflows/visual-regression.yml
+++ b/.github/workflows/visual-regression.yml
@@ -1,0 +1,222 @@
+name: Visual regression
+
+on:
+  pull_request:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+# Advisory to begin with: this reports and comments, it does not fail the
+# build. Promote it to blocking (add --fail-on-change to the diff step)
+# once its noise floor is proven on real PRs.
+jobs:
+  visual:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2 https://github.com/actions/checkout/releases/tag/v6.0.2
+        with:
+          # Full history so the merge-base with main is resolvable. The
+          # baseline is computed from that commit, not from a stored file.
+          fetch-depth: 0
+
+      - name: Resolve merge-base
+        id: refs
+        shell: bash
+        run: |
+          set -euo pipefail
+          git fetch --no-tags origin "${{ github.base_ref }}"
+          # Compare against the merge-base, NOT origin/main's head. Using
+          # main's head would attribute every UI change merged by someone
+          # else while this PR was open to this PR.
+          MERGE_BASE="$(git merge-base "origin/${{ github.base_ref }}" HEAD)"
+          echo "merge_base=${MERGE_BASE}" >> "$GITHUB_OUTPUT"
+          echo "head_sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+          echo "Baseline will be computed from ${MERGE_BASE}"
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0 https://github.com/actions/setup-node/releases/tag/v4.4.0
+        with:
+          node-version: '24.2.0'
+
+      - name: Install Corepack
+        run: npm install -g corepack && corepack enable
+
+      - name: Resolve pnpm store path
+        shell: bash
+        run: echo "STORE_PATH=$(pnpm store path --silent)" >> "$GITHUB_ENV"
+
+      - name: Cache pnpm store
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0 https://github.com/actions/cache/releases/tag/v4.3.0
+        with:
+          path: ${{ env.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile --prefer-offline
+
+      - name: Install Playwright browser
+        run: pnpm --filter @services/vault exec playwright install --with-deps chromium
+
+      # Stash the capture harness BEFORE any checkout juggling. The two
+      # sides must be photographed by the SAME harness, otherwise a change
+      # to the harness itself reads as an app change. It also solves the
+      # bootstrap problem: the merge-base does not have these files at all
+      # until this PR merges.
+      - name: Stash capture harness
+        run: |
+          set -euo pipefail
+          H="${{ runner.temp }}/harness"
+          mkdir -p "$H"
+          cp -R services/vault/e2e/visual "$H/vault-visual"
+          cp services/vault/playwright.visual.config.ts "$H/"
+          cp -R packages/babylon-core-ui/visual "$H/core-ui-visual"
+          cp packages/babylon-core-ui/playwright.visual.config.ts "$H/core-ui-playwright.visual.config.ts"
+          cp scripts/visual-diff.mjs "$H/"
+
+      - name: Capture candidate (PR head)
+        env:
+          CAPTURE_ROOT: ${{ runner.temp }}/visual/candidate
+        run: ./.github/scripts/visual-capture.sh
+
+      - name: Check out merge-base source
+        run: git checkout --force "${{ steps.refs.outputs.merge_base }}"
+
+      - name: Restore capture harness onto merge-base
+        run: |
+          set -euo pipefail
+          H="${{ runner.temp }}/harness"
+          rm -rf services/vault/e2e/visual packages/babylon-core-ui/visual
+          mkdir -p services/vault/e2e packages/babylon-core-ui
+          cp -R "$H/vault-visual" services/vault/e2e/visual
+          cp "$H/playwright.visual.config.ts" services/vault/
+          cp -R "$H/core-ui-visual" packages/babylon-core-ui/visual
+          cp "$H/core-ui-playwright.visual.config.ts" packages/babylon-core-ui/playwright.visual.config.ts
+
+      # The merge-base may carry a different lockfile, so reinstall: the
+      # baseline must reflect the dependencies that commit actually shipped.
+      - name: Install dependencies at merge-base
+        run: pnpm install --frozen-lockfile --prefer-offline
+
+      - name: Capture baseline (merge-base)
+        env:
+          CAPTURE_ROOT: ${{ runner.temp }}/visual/baseline
+        run: ./.github/scripts/visual-capture.sh
+
+      - name: Restore PR head
+        if: always()
+        run: git checkout --force "${{ steps.refs.outputs.head_sha }}"
+
+      # MUST reinstall. The merge-base install above prunes any dependency
+      # this PR adds - including pixelmatch/pngjs, which the diff script
+      # imports. Restoring the files does not restore node_modules, so
+      # without this the diff dies with ERR_MODULE_NOT_FOUND every run.
+      - name: Reinstall dependencies for the diff step
+        if: always()
+        run: pnpm install --frozen-lockfile --prefer-offline
+
+      - name: Compare baseline against candidate
+        id: diff
+        shell: bash
+        run: |
+          set -euo pipefail
+          TOTAL_CHANGED=0
+          for SURFACE in vault storybook; do
+            BASE="${{ runner.temp }}/visual/baseline/${SURFACE}"
+            CAND="${{ runner.temp }}/visual/candidate/${SURFACE}"
+            OUT="${{ runner.temp }}/visual/report/${SURFACE}"
+            [ -d "$CAND" ] || { echo "No candidate captures for ${SURFACE}, skipping."; continue; }
+            echo "### ${SURFACE}" | tee -a "${{ runner.temp }}/visual/diff-output.txt"
+            node scripts/visual-diff.mjs \
+              --baseline "$BASE" --candidate "$CAND" --out "$OUT" \
+              --baseline-ref "${{ steps.refs.outputs.merge_base }}" \
+              --candidate-ref "${{ steps.refs.outputs.head_sha }}" \
+              | tee -a "${{ runner.temp }}/visual/diff-output.txt"
+            N="$(node -e "console.log(require('${OUT}/summary.json').changedCount)")"
+            TOTAL_CHANGED=$((TOTAL_CHANGED + N))
+          done
+          echo "changed_count=${TOTAL_CHANGED}" >> "$GITHUB_OUTPUT"
+
+      - name: Upload visual report
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2 https://github.com/actions/upload-artifact/releases/tag/v4.6.2
+        if: always()
+        with:
+          name: visual-report
+          path: ${{ runner.temp }}/visual/report
+          retention-days: 14
+
+      - name: Summarise
+        if: always()
+        shell: bash
+        run: |
+          {
+            echo "## Visual regression"
+            echo ""
+            echo "Baseline \`${{ steps.refs.outputs.merge_base }}\` (merge-base) vs candidate \`${{ steps.refs.outputs.head_sha }}\`."
+            echo ""
+            if [ "${{ steps.diff.outputs.changed_count }}" = "0" ]; then
+              echo ":white_check_mark: No visual changes detected."
+            else
+              echo ":framed_picture: **${{ steps.diff.outputs.changed_count }} screen(s) changed.**"
+              echo ""
+              echo '```'
+              cat "${{ runner.temp }}/visual/diff-output.txt" || true
+              echo '```'
+            fi
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      # Sticky comment: edits the previous one instead of stacking one per
+      # push. Uses the preinstalled gh CLI rather than adding a third-party
+      # action, per the repo's supply-chain policy.
+      - name: Comment on the PR
+        if: always()
+        env:
+          GH_TOKEN: ${{ github.token }}
+          CHANGED: ${{ steps.diff.outputs.changed_count }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          MARKER="<!-- visual-regression-report -->"
+          BODY_FILE="${{ runner.temp }}/visual/comment.md"
+          mkdir -p "$(dirname "$BODY_FILE")"
+          {
+            echo "$MARKER"
+            echo "### Visual regression"
+            echo ""
+            if [ "${CHANGED:-0}" = "0" ]; then
+              echo ":white_check_mark: No visual changes against the merge-base."
+            else
+              echo ":framed_picture: **${CHANGED} screen(s) changed** against the merge-base."
+              echo ""
+              echo '```'
+              cat "${{ runner.temp }}/visual/diff-output.txt" || true
+              echo '```'
+              echo ""
+              echo "[Open the run](${RUN_URL}#artifacts), download **visual-report**, and open \`index.html\` for the before / after / diff view."
+              echo ""
+              echo "If these changes are intentional there is nothing to update - this repo stores no baseline images. The baseline is recomputed from the merge-base on every run."
+            fi
+          } > "$BODY_FILE"
+
+          EXISTING="$(gh api "repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/comments" \
+            --jq "map(select(.body | startswith(\"$MARKER\"))) | .[0].id // empty")"
+
+          if [ -n "$EXISTING" ]; then
+            gh api --method PATCH "repos/${{ github.repository }}/issues/comments/${EXISTING}" \
+              -F body=@"$BODY_FILE" > /dev/null
+            echo "Updated existing comment ${EXISTING}."
+          else
+            gh api --method POST "repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/comments" \
+              -F body=@"$BODY_FILE" > /dev/null
+            echo "Posted a new comment."
+          fi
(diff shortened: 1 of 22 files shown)


.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,226 @@
...
+        run: |
+          set -euo pipefail
+          MARKER="<!-- visual-regression-report -->"
+          BODY_FILE="${{ runner.temp }}/visual/comment.md"
+          mkdir -p "$(dirname "$BODY_FILE")"
+          {
+            echo "$MARKER"
+            echo "### Visual regression"
+            echo ""
+            if [ "${CHANGED:-0}" = "0" ]; then
```
gbarkhatov: **Blocker: a crashed run posts a green comment.**

`${CHANGED:-0}` substitutes on *null as well as unset*, so when the diff step is skipped or fails, `changed_count` is the empty string, collapses to `0`, and this takes the ":white_check_mark: No visual changes against the merge-base" branch.

This already happened on this branch — in run [30871725170](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/30871725170), `Capture baseline (merge-base)` **failed**, `Compare baseline against candidate` was **skipped**, and `Comment on the PR` still logged `Posted a new comment.` For an advisory check, affirmatively claiming "no changes" when nothing was compared is worse than staying silent.

`Summarise` (L170) has the mirror bug: it compares the raw expression, so `[ "" = "0" ]` is false and it writes `**  screen(s) changed.**` with a blank count.

Both need a branch on whether the comparison actually ran:

```yaml
if [ "${{ steps.diff.outcome }}" != "success" ]; then
  echo ":warning: The capture or comparison did not complete — no visual comparison was made."
elif [ "${CHANGED:-0}" = "0" ]; then
  ...
```

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,226 @@
...
+
+          EXISTING="$(gh api "repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/comments" \
+            --jq "map(select(.body | startswith(\"$MARKER\"))) | .[0].id // empty")"
+
+          if [ -n "$EXISTING" ]; then
+            gh api --method PATCH "repos/${{ github.repository }}/issues/comments/${EXISTING}" \
+              -F body=@"$BODY_FILE" > /dev/null
+            echo "Updated existing comment ${EXISTING}."
+          else
+            gh api --method POST "repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/comments" \
```
gbarkhatov, line 327: `pull_request` hands fork PRs a **read-only** token regardless of the `permissions:` block, so this `POST` returns 403 and, under `set -euo pipefail`, fails the step. This repo is public, so every external-contributor PR would collect a red X from a check that is meant to be advisory.

Guarding the step and letting the job summary carry the same content is enough:

```yaml
if: always() && github.event.pull_request.head.repo.full_name == github.repository
```

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,226 @@
...
+      # would prune - the baseline could then never run. Keeping HEAD's
+      # node_modules also means both sides are photographed by an identical
+      # instrument, which is the property this whole design is built on.
+      #
+      # The trade-off, stated plainly: the baseline renders merge-base SOURCE
+      # against HEAD's external npm packages. Workspace changes (core-ui,
+      # ts-sdk, wallet-connector) are still compared correctly because those
+      # are rebuilt from the checked-out source. A PR that only bumps an
+      # external UI dependency will not show a diff.
+      - name: Capture baseline (merge-base)
```
gbarkhatov, line 148: `retries: 0` in both capture configs plus `set -euo pipefail` means one story that never settles takes down the whole job — and on this side the failure comes from code already on `main`, not from the PR.

Run [30871725170](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/30871725170) is exactly this shape: baseline capture failed, everything downstream skipped, no comparison at all.

For an advisory check, `continue-on-error: true` on both capture steps — with the diff step reporting the missing side rather than the job aborting — keeps one bad screen from costing you the entire report.

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,226 @@
...
+      - name: Resolve merge-base
+        id: refs
+        shell: bash
+        run: |
+          set -euo pipefail
+          git fetch --no-tags origin "${{ github.base_ref }}"
+          # Compare against the merge-base, NOT origin/main's head. Using
+          # main's head would attribute every UI change merged by someone
+          # else while this PR was open to this PR.
+          MERGE_BASE="$(git merge-base "origin/${{ github.base_ref }}" HEAD)"
```
gbarkhatov, line 45: The comment above describes something that does not happen. For `pull_request`, `actions/checkout` checks out `refs/pull/N/merge` — the *merge commit* — so `git merge-base origin/main HEAD` returns **main's tip**, precisely what L36-38 says it is avoiding.

Verified on run [30899977924](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/30899977924): this resolved to `0574a455` (`feat(vault): liquidation dashboard page (#2151)`), which was main's tip at run time. The actual merge-base of this PR is `dc458742`, ten commits earlier.

The *outcome* is still correct, and arguably better than what the comment intends: because the candidate is the merge commit, comparing it against main's tip isolates this PR's contribution, and anything merged by someone else appears on both sides and cancels. So this is a documentation bug rather than a behavioural one — but worth fixing before someone "corrects" it to a true merge-base and silently changes the semantics.

Related, same step: `head_sha` is that merge commit, so the report cites a SHA that appears in neither of this PR's two commits (run 30899977924 reported candidate `bbc37d12`). `${{ github.event.pull_request.head.sha }}` would be more useful to a reader.

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,226 @@
...
+      - name: Stash capture harness
+        run: |
+          set -euo pipefail
+          H="${{ runner.temp }}/harness"
+          mkdir -p "$H"
+          cp -R services/vault/e2e/visual "$H/vault-visual"
+          cp services/vault/playwright.visual.config.ts "$H/"
+          cp -R packages/babylon-core-ui/visual "$H/core-ui-visual"
+          cp packages/babylon-core-ui/playwright.visual.config.ts "$H/core-ui-playwright.visual.config.ts"
+          cp scripts/visual-diff.mjs "$H/"
```
gbarkhatov: Nothing restores this one. The diff step runs after `Restore PR head`, so it uses the tracked `scripts/visual-diff.mjs` from the working tree and this copy is never read.

packages/babylon-core-ui/playwright.visual.config.ts
```diff
@@ -0,0 +1,71 @@
...
+  process.env.STORYBOOK_STATIC_DIR ?? path.join(__dirname, "storybook-static");
+
+export default defineConfig({
+  testDir: path.join(__dirname, "visual"),
+  testMatch: "**/*.visual.spec.ts",
+  fullyParallel: false,
+  // Stories are independent, so parallelism is safe here - unlike the
+  // vault capture, which shares one dev server. Kept modest so the
+  // machine's load does not affect render timing.
+  workers: 2,
```
gbarkhatov, line 43: `workers: 2` is inert. With `fullyParallel: false` Playwright parallelises per *file*, and all 315 tests come from this one spec — so it runs single-worker.

Confirmed in run [30899977924](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/30899977924): `Running 315 tests using 1 worker` / `315 passed (4.5m)`, twice, inside a 16m13s job (the two captures alone accounted for 14m42s).

`fullyParallel: true` should give back ~4-5 minutes, and the comment just above already makes the argument that parallelism is safe here — stories are independent and there is no shared dev server.

packages/babylon-core-ui/playwright.visual.config.ts
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Captures against a **pre-built static Storybook**, not `storybook dev`.
+ * The build is ~11s and the static server has no HMR client injecting
+ * itself into the frame, which removes a whole class of jitter.
+ */
+
+const VISUAL_PORT = 6007;
+
+/** Where the static Storybook is built to, and served from. */
+export const STORYBOOK_STATIC_DIR =
```
gbarkhatov: The `export` is unused: `visual/stories.visual.spec.ts:25-26` recomputes the identical default rather than importing it, so the same path is defined twice.

More consequential: `.github/scripts/visual-capture.sh:32` hardcodes `-o storybook-static`. Setting `STORYBOOK_STATIC_DIR` therefore moves where the spec and the preview server *look* without moving where the build *writes* — the override breaks the capture instead of redirecting it. Either honour the env var in the capture script or drop the override and make the path a plain constant.

services/vault/playwright.visual.config.ts
```diff
@@ -0,0 +1,102 @@
...
+ * compares dev-render against dev-render, so any dev-only artifact
+ * appears identically on both sides and cancels out. Skipping
+ * `tsc --noEmit && vite build` on both sides is what makes it viable to
+ * run this on every PR.
+ */
+
+const VISUAL_PORT = 5177;
+
+/** Where captured PNGs land. CI sets this per side (baseline vs candidate). */
+export const VISUAL_OUTPUT_DIR =
```
gbarkhatov, line 49: `VISUAL_OUTPUT_DIR` is imported nowhere — grepping the repo returns only this declaration. `e2e/visual/routes.visual.spec.ts:29-30` recomputes the same default independently.

Dead export plus a duplicated path, which is two hits on the zero-dead-code rule in CLAUDE.md. Either export it and have the spec import it, or drop the `export`.

services/vault/playwright.visual.config.ts
```diff
@@ -0,0 +1,102 @@
...
+  process.env.VISUAL_OUT_DIR ?? path.join(__dirname, "e2e/visual/__captures__");
+
+/**
+ * Same mock backend the e2e config pins, so no screen reaches a live
+ * host. Feature flags are forced ON: the v3 shell and its sections are
+ * the surface worth protecting, and a flag that flips between the two
+ * sides would otherwise read as a visual regression.
+ */
+const VISUAL_ENV_VARS = {
```
gbarkhatov, line 62: Seven of these nine keys are verbatim copies of `MOCK_ENV_VARS` in `services/vault/playwright.config.ts`. Worth exporting that object and spreading it:

```ts
const VISUAL_ENV_VARS = {
  ...MOCK_ENV_VARS,
  NEXT_PUBLIC_SENTRY_DSN: "",
  NEXT_PUBLIC_FF_ENABLE_V3_UI: "true",
  // ...visual-only overrides
};
```

Drift between the two copies would silently change what the baseline renders, and there is no signal when it happens — the capture would still pass, just against a different app.

scripts/visual-diff.mjs
```diff
@@ -0,0 +1,316 @@
...
+ * absorb GPU-level antialiasing jitter, not real colour differences.
+ */
+const PIXEL_MATCH_THRESHOLD = 0.1;
+
+/**
+ * Changed pixels below this fraction of the image are reported but do
+ * not mark the run as changed. Guards against a stray subpixel without
+ * hiding anything a reviewer would notice.
+ */
+const CHANGED_RATIO_TOLERANCE = 0.0001;
```
gbarkhatov, line 43: Because this is a *ratio*, sensitivity degrades with page height: a 1280x4000 full-page capture silently absorbs ~512 changed pixels, roughly a 24x24 icon.

That is the one place a scaling fudge factor is not needed — the whole argument for the computed baseline is that both sides render in the same container, same browser, same fonts. An absolute floor holds sensitivity constant regardless of how tall a screen is:

```js
const changed = changedPixels > Math.max(MIN_CHANGED_PIXELS, CHANGED_RATIO_TOLERANCE * width * height);
```

For calibration, the real diffs on this PR's own run landed between 0.090% and 0.569% — roughly 9x to 57x the current threshold, so there is plenty of headroom to tighten.

scripts/visual-diff.mjs
```diff
@@ -0,0 +1,316 @@
+/**
```
gbarkhatov, line 1: This file now decides whether a PR gets flagged, and it has real branching worth pinning: `added` / `removed` / `size-changed`, the tolerance cutoff, and `parseArgs`' bare-`--` and flag-without-value handling. It has no tests, and the root `eslint.config.mjs` only lints `**/package.json`, so these 316 lines of load-bearing CI logic are neither linted nor covered.

A few vitest cases over `parseArgs` and `compareOne` (two tiny in-memory PNGs) would be cheap, and would pin the tolerance semantics against exactly the kind of accidental loosening that makes a visual check quietly stop catching things.

Minor, same file: `main().catch` at L314 writes `error.message` only — a `PNG.sync.read` failure on a truncated capture will be painful to diagnose without the stack.

packages/babylon-core-ui/visual/stories.visual.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+    await page.waitForFunction(
+      () => document.body.classList.contains("sb-show-main"),
+      undefined,
+      { timeout: STORY_RENDER_TIMEOUT_MS },
+    );
+
+    // A story that threw renders Storybook's error panel instead. That is
+    // stable, so it would happily become a baseline and the broken story
+    // would never be noticed again. Fail loudly instead.
+    const failureClass = await page.evaluate(() => {
```
gbarkhatov, line 183: This check is unreachable, and the story it is guarding against fails with an opaque timeout instead of the message written for it.

`WebView.showErrorDisplay` calls `showMode("ERROR")`, and `showMode` *removes* every other mode class — from `@storybook/core@8.6.15`, `dist/preview-api/index.js`:

```js
showMode(e) {
  clearTimeout(this.preparingTimeout), Object.keys(qn).forEach((t) => {
    t === e ? j.body.classList.add(wt[t]) : j.body.classList.remove(wt[t]);
  });
}
```

So a story that throws never gets `sb-show-main`, the `waitForFunction` at L164 times out after 20s, and this block never runs. `sb-show-nopreview` has the same problem.

Gating on any terminal state makes it reachable, and keeps the actionable message:

```ts
await page.waitForFunction(() => {
  const c = document.body.classList;
  return c.contains("sb-show-main") || c.contains("sb-show-errordisplay") || c.contains("sb-show-nopreview");
}, undefined, { timeout: STORY_RENDER_TIMEOUT_MS });
```

services/vault/e2e/visual/routes.visual.spec.ts
```diff
@@ -0,0 +1,98 @@
+/**
```
gbarkhatov, line 1: **Blocker: this file is also collected by the regular e2e suite.**

`services/vault/playwright.config.ts` is `testDir: e2e` + `testMatch: "**/*.spec.ts"` with no `testIgnore`, and `e2e/visual/routes.visual.spec.ts` matches. Verified locally:

```
$ pnpm exec playwright test --list
...
[chromium] > visual/routes.visual.spec.ts:79:5 > capture overview at desktop
[chromium] > visual/routes.visual.spec.ts:79:5 > capture overview at mobile
...
Total: 31 tests in 6 files        # 19 tests in 5 files before this PR
```

Under that config they run on port 5175 without the v3/explore feature flags, without `reducedMotion: "reduce"` — so `waitForVisualStability` chases live animations for its full 15s before throwing — and with `retries: 2`, so each of the 12 does it three times.

Nothing in CI runs `test:e2e`, so this lands on developers rather than the pipeline, but it is a regression to an existing suite. Cheapest fix is `testIgnore: "**/visual/**"` in `playwright.config.ts`; renaming the spec to `*.visual.ts` would also work.

services/vault/e2e/visual/routes.visual.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+async function blockOffsiteRequests(page: Page): Promise<void> {
+  await page.route("**/*", (route) => {
+    const { hostname } = new URL(route.request().url());
+    const isLocal = hostname === "localhost" || hostname === "127.0.0.1";
+    return isLocal ? route.continue() : route.abort();
+  });
+}
+
+/**
+ * Deliberately NOT `mockVpProxy`. It answers `/vp-health` with an empty
```
gbarkhatov, line 49: Worth being explicit about what this surface buys, because this comment says it plainly: VP calls are blocked, contract reads fall through to a JSON-RPC error, and the app renders its "Something went wrong" card.

So the 12 vault captures are largely the shell plus an error card — and they are the expensive half of the run, since reaching them needs two `nx run-many` builds and two Vite dev servers. It also degrades silently: `MIN_RENDERED_TEXT_LENGTH = 20` passes happily on an error card, so if a route later breaks *harder*, nothing notices.

Not asking to block on it — #1592 is the real unblock. Either of these would work for me:

- land Storybook-only now and add the vault surface once connected states are reachable; or
- assert a per-route marker (a `data-testid` unique to each target) so a screen collapsing into the shared error shell fails the capture instead of quietly becoming the baseline.

Either way, the PR description's "12 vault screens" oversells what is covered today — worth a sentence there so the next reader does not over-trust it.

services/vault/tsconfig.lib.json
```diff
@@ -24,5 +24,11 @@
       "@/*": ["./src/*"]
     }
   },
-  "include": ["src", "e2e/fixtures", "e2e/pages", "sentry.client.config.ts"]
+  "include": [
+    "src",
+    "e2e/fixtures",
+    "e2e/pages",
+    "e2e/visual",
```
gbarkhatov, line 31: Good — but the mirror change for core-ui is missing. `packages/babylon-core-ui/tsconfig.lib.json` is `include: ["src"]` and `tsconfig.node.json` is `include: ["vite.config.ts"]`, so `visual/stories.visual.spec.ts` (207 lines) and `playwright.visual.config.ts` sit outside every project and are typechecked by nothing.

core-ui's `build` runs `tsc --noEmit -p tsconfig.lib.json`, so a type error in that spec surfaces only when CI actually runs the capture — seven minutes in, on the far side of a Storybook build.

(I confirmed this addition on the vault side introduces no type errors.)

packages/babylon-core-ui/src/widgets/sections/ValidatorSelector/ValidatorSelector.stories.tsx
```diff
@@ -18,7 +18,10 @@ const sampleValidators: ValidatorRow[] = Array.from({ length: 15 }, (_, i) => {
         ),
         name: `Validator ${index}`,
         apr: `${25 + (index % 6)}.${(index * 7) % 100}`.padEnd(6, "0") + "%", // simple varied APR
-        votingPower: `${(Math.random() * 1).toFixed(6)} sBTC`,
+        // Derived from the index, not Math.random(): the story is snapshotted
+        // by the visual-regression capture, and a random value would differ on
+        // every run and report a change that isn't one.
+        votingPower: `0.${String((index * 137) % 1_000_000).padStart(6, "0")} sBTC`,
```
gbarkhatov, line 24: Good catch, and worth flagging for whoever reads the check's first real output: the 6 changed screens the bot reported on this PR are all `widgets-modals-validatorselector--*`, i.e. this very fix. They are self-inflicted and will clear once this merges — nobody should go chasing them.

It is also the only determinism source that had to be fixed in a story rather than in the harness, which is a good sign for the approach.

gbarkhatov (dismissed):
Reviewed the harness end to end and cross-checked the claims against the two CI runs on this branch ([30871725170](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/30871725170), [30899977924](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/30899977924)), the Storybook 8.6.15 `preview-api` source, and a local `playwright test --list`.

The design is right, and I want to say so before the list of problems: computing the baseline from a second capture in the same job removes the entire "stale committed baselines" maintenance class, and the reasoning is documented where the next person will actually hit it — the rejected alternatives (`networkidle`, `sb-show-main` alone, scroll-dimension polling, `clock.install()`, `mockVpProxy`) are all recorded in place. Stashing the harness before the checkout juggling is the right call. Advisory-first with an explicit promotion path is the right rollout.

Approving, since the check is advisory and none of this can affect production. Two things I would want fixed before anyone relies on it:

1. **A crashed run posts a green comment.** `${CHANGED:-0}` collapses an empty `changed_count` to `0`, so a skipped diff step takes the ":white_check_mark: No visual changes" branch. This already happened in run 30871725170. A check that affirmatively says "no changes" when nothing was compared is worse than one that says nothing.
2. **The vault e2e suite now collects the visual specs.** `playwright.config.ts` has no `testIgnore`, so `pnpm test:e2e` goes from 19 tests in 5 files to 31 in 6 — running the captures without the visual env, without reduced motion, and with `retries: 2`.

Then a handful worth folding in while you are here: fork PRs will red-X on the comment step, `workers: 2` is inert (~4-5 min of the 16m13s job), the Storybook error-panel check is unreachable, the merge-base comment describes something that does not happen, and core-ui's `visual/` is typechecked by nothing.

Two judgement calls left to you, both inline: how much the vault surface is worth today given it largely captures the error shell, and whether the pixel tolerance should be absolute rather than a ratio.

Checked and found fine, so nobody redoes the work: the lockfile diff is purely additive and Playwright dedupes to a single 1.56.1 (the shared browser install is safe), the `tsconfig.lib.json` addition introduces no type errors, and the report artifact is only 6.78 MiB despite carrying all ~650 PNGs — not worth optimising.

scripts/visual-diff.mjs
jonybur: Partly taken. `parseArgs` and `compareOne` are now exported and the CLI entry is guarded with an `import.meta.url` check, so the module is importable without running a diff, and `main().catch` logs `error.stack` rather than just the message.

I stopped short of the tests themselves. There is no root vitest project to put them in: `vitest` is not a root dependency, nx has no vitest plugin (only `@nx/js/typescript` and `@nx/eslint/plugin`), and the repo root is not a pnpm workspace member, so `nx run-many -t test --all` would not pick them up. Adding one means a new root devDependency, a vitest config and an nx target - real infrastructure, and this PR is already large. Left as a follow-up rather than grown into here; the exports and the CLI guard are the part that makes it cheap to add.

The tolerance semantics you were most concerned about pinning are at least now named and documented: `MIN_CHANGED_PIXELS` is an absolute floor alongside `CHANGED_RATIO_TOLERANCE`, and `changedPixelCutoff()` is the single exported place that combines them.

packages/babylon-core-ui/src/widgets/sections/ValidatorSelector/ValidatorSelector.stories.tsx
jonybur: Agreed, no change needed - noting it here so the record is clear that the 6 flagged screens are this PR fixing itself and will clear on merge.

jonybur:
All 18 threads addressed in 5c8ae7f — 16 fixed, 2 answered (unit tests for `visual-diff.mjs`, and the ValidatorSelector note which needed no change).

Both blockers are fixed and verified:

- **Green comment on a crashed run.** `Summarise` and the comment step now branch on `steps.diff.outcome` and on a new `compared_surfaces` output, so an empty `CHANGED` can no longer collapse to `0` and take the success branch. Captures got `continue-on-error: true` and the diff step names the missing side rather than dying on ENOENT, so one unsettled story costs that surface instead of the whole report.
- **Visual specs collected by the e2e suite.** `testIgnore: "**/visual/**"` in `playwright.config.ts`. `playwright test --list` is back to `Total: 19 tests in 5 files`, from 31 in 6.

**One thing your review did not catch, which wiring the harness into a typecheck surfaced.** `reducedMotion` is not a top-level Playwright `use` option in 1.56 — it belongs under `use.contextOptions`. Both visual configs set it at the top level, where Playwright silently ignores it, so every capture on both surfaces has been running with animations live. That is almost certainly the motion `waitForVisualStability` was chasing for its full timeout. Moved in both configs. It only became visible because of your `tsconfig` thread, so that one paid for itself immediately.

Everything else: fork-PR guard on the comment step, the merge-base comment corrected (and `head.sha` reported instead of the merge commit), the never-read stashed `visual-diff.mjs` dropped, `MIN_CHANGED_PIXELS` absolute floor alongside the ratio, the Storybook render wait gated on any terminal mode so the error-panel message is reachable, `MOCK_ENV_VARS` spread instead of duplicated, the `STORYBOOK_STATIC_DIR` override dropped for a plain constant both specs import, `VISUAL_OUTPUT_DIR` imported rather than recomputed, `@playwright/test` pinned exactly, core-ui `tsconfig.node.json` picking up the harness with `build` extended to typecheck it, `error.stack` on diff failure, and a sentence in the PR description saying the 12 vault screens are captured unconnected.

`fullyParallel: true` confirmed working — a capture slice now reports `Running 4 tests using 2 workers` where it was silently single-worker.

Checks: core-ui typecheck clean, lint 0 errors, live capture slice green. The three `services/vault/src` type errors visible locally on this branch are pre-existing stale-`dist` artifacts, not from this change — confirmed by stashing.

jonybur (commented):
Two findings at 20f0bf7, both in the baseline-side capture path.

**1. The stashed visual config imports `MOCK_ENV_VARS` from a file that is not stashed - on the baseline side the import silently resolves to `undefined`, and the vault baseline runs with no mock env.**

`services/vault/playwright.visual.config.ts` does `import { MOCK_ENV_VARS } from "./playwright.config"`. The "Stash capture harness" step copies the visual config onto the merge-base, but `playwright.config.ts` stays at the merge-base version - where `MOCK_ENV_VARS` is a plain `const`, not exported (main's `playwright.config.ts:10` today). Playwright transpiles configs through Babel's CJS interop, so a missing named export is not an error: `MOCK_ENV_VARS` reads as `undefined`, `{ ...undefined }` spreads to `{}`, and `VISUAL_ENV_VARS` silently loses every mock contract address. The baseline dev server then boots without the `NEXT_PUBLIC_TBV_*` addresses, so the baseline side either renders the env-validation failure surface (all 12 vault screens report as changed on a PR that changed nothing visual) or never mounts (empty baseline dir, vault silently uncompared). Either way the vault report is wrong for every run of this PR.

The deeper issue is that the "both sides photographed by the SAME harness" guarantee has dangling dependencies: the stashed harness imports unstashed, commit-varying files - `playwright.config.ts` here, and `e2e/fixtures/*` from `routes.visual.spec.ts` (those happen to exist at today's merge-base, so they work by luck, not construction). Cheapest fix: stash `playwright.config.ts` alongside the harness and restore it too. Alternatively have the visual config throw loudly when the import comes back `undefined`, so a broken baseline fails the capture instead of photographing the wrong app.

**2. A surface that captured on neither side still lets the run read green.** The `COMPARED` gate fixed the fully-crashed case, but it is per-run, not per-surface: if the vault captures are empty on both sides (exactly what finding 1 can produce) while storybook compares clean, `COMPARED=1`, `CHANGED=0`, and the comment posts "No visual changes against the merge-base" with no hint that half the surface list was never compared. The green branch should also require `COMPARED` to equal the surface count (2), or carry the missing-surface lines from `diff-output.txt` into the green comment.

Minor, same workflow: the sticky-comment lookup (`gh api .../issues/N/comments`) is unpaginated, so once a PR accumulates more than 30 issue comments the marker match can miss and the step stacks a new comment per push.

scripts/visual-diff.mjs
```diff
@@ -0,0 +1,341 @@
...
+ * matter how large the image. Sized well under a glyph or icon: real diffs
+ * measured on this harness's own runs landed at 0.090%-0.569% of the frame,
+ * hundreds to thousands of pixels, so there is ample headroom above this.
+ */
+const MIN_CHANGED_PIXELS = 24;
+
+/** The cutoff a capture must exceed to count as changed. */
+export function changedPixelCutoff(width, height) {
+  return Math.max(MIN_CHANGED_PIXELS, CHANGED_RATIO_TOLERANCE * width * height);
+}
```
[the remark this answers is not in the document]
jonybur: The floor being inert at the current three viewports is intentional: `MIN_CHANGED_PIXELS` is a lower bound so the cutoff cannot collapse toward zero if a smaller capture viewport is ever added (the ratio term scales with area) - it is not an extra tolerance stacked on the current viewports, where the area-scaled term is meant to govern. The docblock above the constant records the measured basis: real diffs on this harness's own runs landed at 0.090%-0.569% of the frame (hundreds to thousands of pixels), far above either term. Sub-cutoff border/glyph diffs are exactly the antialiasing drift the tolerance exists to absorb, and the check is advisory, so the failure mode is a missed hint, not a false green gate.

jonybur:
@greptile-apps

jrwbabylonlab (approved):
interesting PR

PR #2184, opened 2026-08-04
chore(vault): retire the ENABLE_V3_UI feature flag and delete the v2 shell

jeremy-babylonlabs:
## Summary

Retires the `NEXT_PUBLIC_FF_ENABLE_V3_UI` / `isV3UiEnabled` feature flag and deletes the legacy v2 shell it gated.

The flag is set to `true` in **every deployed environment** — `vault-devnet`, `vault-staging-devnet`, and `vault-testnet` (the `prod` env in the release matrix). The devnet → staging → testnet rollout already happened, so v3 is what production serves today. This change codifies the deployed reality rather than shipping anything new.

## What changed

- Collapsed every `featureFlags.isV3UiEnabled && X` / ternary down to its v3 branch across the app (router, `RootLayout`, dashboard, deposit flow, activity, banners, modals, route helpers).
- Deleted the v2 shell and everything only it reached: the old `Header` + horizontal top-nav, the single-page dashboard path (supply-cap / pending / collateral / loans sections and the withdraw flow that lived on it), the page-level footer, the v2 branches of `ProtocolStatusBanner`, `MaxVaultsNotification`, `FeesSection`, `V3ModalShell`, `ReorderSuccessModal`, `OverviewSection`, `ActivityList`, and the v2 reserve-detail base route.
- Removed the `isV3UiEnabled` getter from `config/featureFlags.ts`, both `.env.example` entries, the workflow env var, and the now-unreferenced v2-only copy keys.
- Updated tests: dropped the flag default and per-test overrides, deleted tests that only asserted flag-off / v2 behaviour, and kept the v3 tests (which now need no flag setup).

## Deliberately out of scope

The other v3 section flags are **untouched** — each is still disabled in every environment and continues to gate its surface:

- `LIQUIDATION_ANALYSIS_CHART` (`isLiquidationAnalysisChartEnabled`) — still gates the liquidation-analysis chart, the sidebar Liquidations section, and `/liquidations`.
- `ENABLE_MARKET_DETAIL_PAGE` (`isMarketDetailPageEnabled`) — still gates `/markets/:reserveId` and the Market Info button.
- `ENABLE_EXPLORE` (`isExploreEnabled`) — still gates `/explore` and its sidebar entry.

`config/v3Navigation.ts` (the section-flag gates) is unchanged.

## Dead code removed with it

Retiring the flag left the v2-only component tree unreachable. Reachability analysis from the app entry points (`main`/`router`/`providers`) found 26 files reachable before the change and unreachable after; those are deleted here along with their orphaned tests and the second-order helpers only they used:

- v2 dashboard sections (supply cap, collateral, loans, pending deposit/withdraw) and their cards
- v2 activity rows (`ActivityCard`, `LiquidationGroupCard`) and the `components/Activity` barrel
- v2 `VaultProviderSelector` / `UtxoSplitSelector` (the v3 file already declares its own `TwoVaultSplitProps`)
- v2 `WithdrawVaults` modal set and the `ArtifactDownloadModal`
- `HealthFactorGauge` (+ its helper), the header `BrandLockup`, and `shouldDisplayTestingMsg`

**No capability is lost:** v3 keeps artifact download via `RecoveryArtifactsCard`, withdrawal via `WithdrawFlow`, and the vault lifecycle via the `VaultsPage` sections — all still reachable. Post-change reachability matches the pre-change baseline exactly (only the vite plugin, ws stub and ambient types are unreferenced, as before).

## Known follow-up

48 copy keys in `copy.ts` became unreferenced when those components went (e.g. the `artifactDownload` / `batchedDeposit` / `artifactCallout` blocks and the v2 pegout-section labels). They are left in place here deliberately: `copy.ts` is being edited concurrently on another branch, and widening this PR's footprint in that file would create avoidable conflicts. They carry no behaviour — worth a small follow-up PR once that lands.

## Real-wallet E2E hooks — needs an owner's eye

The deleted v2 components carried `data-testid`s that `services/vault/e2e/real/actions/` still selects on. Two classes:

- **Restored here:** `pending-deposit-resume-cta` had a clean 1:1 equivalent (the per-row primary action in `VaultsLifecycleSections`), so the hook and its "this is an E2E hook" comment are carried over per the repo rule.
- **Needs an E2E-side update (not done here):** `withdraw.ts` still selects `collateral-actions-button`, `collateral-withdraw-button`, `withdraw-modal-confirm-button`, `withdraw-disabled-reason` and a `withdraw-vault-row-*` prefix; `resume.ts` selects `pending-deposit-card`. These belonged to the v2 actions-menu → modal flow, which v3 does not have — v3 goes straight from a per-row Withdraw button into `WithdrawFlow`, which is separately instrumented (`withdraw-confirm-button`, `withdraw-done-button`, `withdraw-hf-block-warning`). Bolting the old ids onto v3 controls would not make the scripts pass; the two actions need rewriting against the v3 flow by someone who can run the real-wallet CLI.

Note on causation: the flag was already `true` in every deployed environment, so these selectors were already not matching there (and `pending-deposit-card` / `connect-wallet-button` never existed in `main` at all). Locally, with the flag defaulting off, the v2 markup still satisfied them. This PR removes that last fallback, so the drift is now unambiguous rather than latent.

## Verification

- `tsc --noEmit`, `eslint`, the full `vitest` suite, and the production build all pass.
- `grep -rn "isV3UiEnabled\|NEXT_PUBLIC_FF_ENABLE_V3_UI"` over the repo returns zero hits.
- The three surviving section flags are still present and still gating their surfaces.

## Follow-up (after merge)

The now-dead `NEXT_PUBLIC_FF_ENABLE_V3_UI` environment variable should be deleted from the three GitHub environments (`vault-devnet`, `vault-staging-devnet`, `vault-testnet`).


jonybur (dismissed):
Mechanical cleanup with the flag on everywhere in production - collapses each `isV3UiEnabled` ternary to its v3 branch and deletes the v2 shell it gated. The three remaining section-level flags (`ENABLE_EXPLORE`, `ENABLE_MARKET_DETAIL_PAGE`, `LIQUIDATION_ANALYSIS_CHART`) are correctly left alone, and their per-section guards still enforce them. Copy migration to `copy.ts` handled and dead imports removed cleanly across the affected components.

services/vault/src/copy.ts
```diff
@@ -538,11 +538,8 @@ export const COPY = {
       transactionReserveLabel: "Transaction Reserve",
       transactionReserveTooltip:
         "A small portion of your deposit is reserved in a dedicated output to fund a future protocol claim transaction. It remains locked until claim conditions are met and is returned to you if unused.",
-      // Amount-input left-field label; the slider renders its Max button when
-      // this reads "max" (case-insensitive), so keep the value as "Max".
-      maxLabel: "Max",
-      // v3 amount row: the depositable maximum is labelled as the balance,
-      // with `maxTooltip` explaining the fee buffer / cap adjustments.
+      // The depositable maximum is labelled as the balance, with `maxTooltip`
```
jrwbabylonlab, line 516: The components are gone but their strings are not: **46 keys in `copy.ts` are now referenced nowhere**. Measured as a differential against `main` (2 orphaned before this PR, 48 after), so these are newly dead rather than pre-existing noise.

Spot-checked consumers, all deleted by this PR:

| key | was used by |
|---|---|
| `healthFactorAtRisk` / `Healthy` / `Liquidatable` | `shared/utils/healthFactorGauge.ts` |
| `challengePeriodEndsIn`, `withdrawalTxLabel`, `withdrawalTxPending` | `simple/PendingWithdrawCard.tsx` |
| `borrowButton` | `simple/LoansSection.tsx` |

CLAUDE.md treats `copy.ts` as a reviewed surface under the zero-dead-code rule, so these should come out in the same pass.

Two look dead but are not — leave them: `artifactDownload` and `providerSelectDescription`, where a longer `…V3` key is still live.

<details><summary>All 46</summary>

`artifactCallout` `batchedDeposit` `bodyDownloading` `borrowButton` `broadcastHelper` `btcPriceLabel` `calloutLinkText` `cardSummaryProgressing` `challengeNote` `challengePeriodEndsIn` `challengePeriodEndsLabel` `challengePeriodEndsSoon` `challengePeriodTypicalDuration` `completedTitle` `confirmButtonWithAmount` `confirmationsLabel` `confirmationsValue` `contactSupport` `dateLabel` `detailsAria` `detailsAriaLabel` `downloadNow` `expiredTooltip` `groupLabel` `healthFactorAtRisk` `healthFactorHealthy` `healthFactorLiquidatable` `initiatedLabel` `learnMoreLink` `learnMorePrefix` `liquidationRiskLabel` `openDetailsAria` `pendingTitle` `providerActiveLabel` `providerGroupUnavailableLabel` `providerLinkLabel` `providerUnhealthyReason` `releaseDisabledTooltip` `releaseHfBreachTooltip` `singleLabel` `titleDownloading` `totalLabel` `triggerLabel` `vaultLinkLabel` `withdrawalTxLabel` `withdrawalTxPending`

</details>

The code deletion itself looks thorough — I swept all 511 modules under `services/vault/src` for reachability and found no dead files, components, hooks or routes. Only the strings were missed.

jrwbabylonlab (dismissed):

jonybur (approved):
The flag was already `true` in every deployed environment, so this is codifying deployed reality rather than shipping new behaviour. Cross-checked the mechanical parts that would silently break: `isV3UiEnabled` / `NEXT_PUBLIC_FF_ENABLE_V3_UI` are gone from code (only deletions remain), `V3_GUARDED_ROUTE_PATHS` and its test are removed together (no dead export), and `getReserveDetailRoute`'s reduced signature is applied at every callsite. The one carried-over E2E hook (`pending-deposit-resume-cta`) moves cleanly with its comment; the four other v2-only ids are correctly out of scope and flagged in the description for a separate real-wallet CLI update. Surviving router tests still pin the section flags and the overlay-over-any-Aave-page contract without relying on the removed flag setup. The 48 unreferenced copy keys deferred to a follow-up are a known deviation from zero-dead-code called out in the description; leaving them here keeps this from colliding with the concurrent copy.ts branch, which is the right trade.

PR #2189, opened 2026-08-05
fix(vault): align the light theme tokens to the v3 Figma palette

jeremy-babylonlabs:
## Summary

The v3 Figma light palette is neutral where the current light palette is teal-tinted. `accent-primary` is the app's primary text token — 76 files under `services/vault/src` use it — so this is an app-wide mismatch, not a bug in any one screen. It happened to surface on the entry screen, whose headline reads greenish in light mode.

Note v3 did not drop the navy: it moved it off body text onto `accent/secondary` (`#042f40`, already `primary-main`). That move is why the headline currently reads the way it does.

## Deltas

Verified against the entry-screen frames in Figma (light `10208-58290`, dark `10206-56371`).

**Light** — three wrong:

| Figma token | Figma | code token | before | after |
| --- | --- | --- | --- | --- |
| `text/primary` | `#000000` | `accent-primary` | `#12495E` | `#000000` |
| `stroke/secondary` | `#dddddd` | `secondary-strokeLight` | `#d7e1e7` | `#dddddd` |
| `stroke/primary` | `#cccccc` | `secondary-strokeDark` | `#387085` | `#cccccc` |

`accent-secondary` `#666666`, `secondary-main` `#CE6533`, `primary-main` `#042F40` and `background-secondary` `#F9F9F9` already matched.

**Dark** — one:

| Figma token | Figma | code token | before | after |
| --- | --- | --- | --- | --- |
| `text/primary` | `#ffffff` | `accent-primary` | `#F0F0F0` | `#ffffff` |

Everything else in dark already matched, including `stroke/secondary` `#2F2F2F`, `stroke/primary` `#5A5A5A` and `background/secondary` `#202020`.

Worth flagging for whoever touches these next: the stroke names invert between the two systems. Figma's `stroke/primary` is the code's `secondary-strokeDark`, and `stroke/secondary` is `secondary-strokeLight`. Figma's `accent/primary` is the orange `#ce6533`, which is the code's `secondary-main` — not `accent-primary`, which is text.

## Why the override is vault-scoped

The tokens come from `createThemes` in `packages/babylon-core-ui/tailwind.config.js`, and **both** apps consume it as a Tailwind preset:

- `services/vault/tailwind.config.ts`
- `services/simple-staking/tailwind.config.ts`

Editing core-ui's light palette would repaint the staking dApp too. So no core-ui file is touched here. This constraint is independent of the v3 flag retirement — it holds regardless.

tw-colors emits the palette as `--twc-*` HSL triplets on `.light` / `.dark`, which is why utilities resolve as `hsl(var(--twc-accent-primary) / <alpha>)`. Re-pointing those four custom properties in the vault's own `globals.css` overrides the variable the utilities already read, so there is no specificity race against the ~76 files' worth of `text-accent-primary`, and it stays theme-aware. `globals.css` already defines vault-only vars (`--success-bright`, `--risk-*`) with a `.dark` override block, so this follows the existing shape.

## Contrast fix folded in

`secondary-strokeDark` is not only a stroke in this codebase — core-ui also used it as a muted foreground. Repointing it to `#cccccc` would have taken those icons to **1.61:1** on `surface`. Checking that turned up a pre-existing failure too: in dark, `#5A5A5A` on `#121212` is **2.72:1**, already under the 3:1 floor for UI components in both apps.

So `6c1958ca` points the four foreground sites at `accent-secondary` (Figma `text/secondary`), which strictly raises contrast everywhere — light 1.61→5.74, dark 2.72→8.64, and simple-staking 5.49→5.74 in light while fixing its dark failure:

- `Hint.tsx:32` `ICON_COLOR.default` — covers 10 vault call sites that pass no `icon`
- `Accordion.css:25` chevron — `ExpandablePanel`, `FeesSection`, `UtxoSplitSelectorV3`, `VaultProviderSelectorV3`
- `WalletMenuSettingItem.tsx:51` — via `BtcEthWalletMenu`
- `DepositForm.tsx:331` (vault-owned)

The palette in `createThemes` is still untouched; this corrects three components that used a stroke token as a foreground. `Select`, `Toggle`, `BTCFeeRate` and `ValidatorSelector` also do this but are not imported by the vault, so they are left alone.

The disabled deposit CTA now resolves to `#cccccc`, which matches wallet-connector's existing `disabled:!bg-[#CCCCCC]` — that hardcode exists precisely because the light token was teal, so this is convergence on the design system.

## Deferred

`primary-dark` `#12495E`, `primary-light` `#387085` and `accent-disabled` `#9ab7c2` stay teal, so some pairings are mixed — `StepRow.tsx:67` keeps a teal step marker beside now-black text. The v3 entry-screen frames do not specify `primary/*` or `accent/disabled`, so neutralizing them needs its own design input. Also worth knowing: light `strokeLight` `#dddddd` and `strokeDark` `#cccccc` now sit 1.19:1 apart, so the two rule weights are near-indistinguishable in light. That is what the Figma palette specifies.

## Verification

Rebased onto `main` at `bc3d122c`, after the `ENABLE_V3_UI` retirement (#2184). All three tokens are still consumed post-deletion — `accent-primary` 76 files, `secondary-strokeLight` 42, `secondary-strokeDark` 13 — so none of the overrides is dead.

Computed from the production bundle (`services/vault/dist/assets/*.css`), taking the cascade-winning declaration per theme:

```
LIGHT  accent-primary #000000  secondary-strokeLight #dddddd  secondary-strokeDark #cccccc
DARK   accent-primary #ffffff  secondary-strokeLight #2f2f2f  secondary-strokeDark #5a5a5a
```

Same computation over a `simple-staking` Tailwind build still yields `#12495E` / `#d7e1e7` / `#387085` — unchanged.

- `git diff --stat` touches one file; nothing under `packages/babylon-core-ui/`
- `tsc --noEmit` clean
- `nx run vault:lint` — 0 errors
- `nx run vault:build` — passes
- `nx run vault:test` — 261 files, 2976 tests passing

### DIFF
--- a/services/vault/src/globals.css
+++ b/services/vault/src/globals.css
@@ -61,15 +61,18 @@ html {
   --motion-duration-skeleton: 1.5s;
 }
 
-/* Palette overrides (Figma light: text/primary #000000, stroke/secondary
- * #dddddd, stroke/primary #cccccc). core-ui's tw-colors themes are shared with
- * simple-staking, so the vault re-points the --twc-* HSL triplets here rather
- * than editing that palette. */
+/* Palette overrides. core-ui's tw-colors themes are shared with simple-staking,
+ * so the vault re-points the --twc-* HSL triplets here rather than editing that
+ * palette. The --twc-<token> names are tw-colors' output format, coupled to the
+ * exact "tw-colors": "3.3.2" pin and the key names in core-ui's createThemes.
+ * These blocks must stay below the @tailwind directives: tw-colors emits
+ * `.light,[data-theme=light]` at the same specificity, so source order is what
+ * makes the override win. */
 .light,
 [data-theme="light"] {
-  --twc-accent-primary: 0 0% 0%;
-  --twc-secondary-strokeLight: 0 0% 86.7%;
-  --twc-secondary-strokeDark: 0 0% 80%;
+  --twc-accent-primary: 0 0% 0%; /* Figma text/primary     #000000 */
+  --twc-secondary-strokeLight: 0 0% 86.7%; /* Figma stroke/secondary #dddddd */
+  --twc-secondary-strokeDark: 0 0% 80%; /* Figma stroke/primary   #cccccc */
 }
 
 /* Dark overrides — Figma dark text/primary #ffffff, plus the risk-card dark
@@ -77,7 +80,7 @@ html {
  * [data-theme="dark"] hook. */
 .dark,
 [data-theme="dark"] {
-  --twc-accent-primary: 0 0% 100%;
+  --twc-accent-primary: 0 0% 100%; /* Figma text/primary #ffffff */
   --risk-amber: 255 179 0;
   --risk-green: 0 230 118;
   --risk-muted: 124 124 124;
--- a/services/vault/src/globals.css
+++ b/services/vault/src/globals.css
@@ -61,10 +61,23 @@ html {
   --motion-duration-skeleton: 1.5s;
 }
 
-/* v3 risk-card dark overrides (Figma dark variants). Covers both the
- * `dark` class (tailwind darkMode:"class") and the [data-theme="dark"] hook. */
+/* Palette overrides (Figma light: text/primary #000000, stroke/secondary
+ * #dddddd, stroke/primary #cccccc). core-ui's tw-colors themes are shared with
+ * simple-staking, so the vault re-points the --twc-* HSL triplets here rather
+ * than editing that palette. */
+.light,
+[data-theme="light"] {
+  --twc-accent-primary: 0 0% 0%;
+  --twc-secondary-strokeLight: 0 0% 86.7%;
+  --twc-secondary-strokeDark: 0 0% 80%;
+}
+
+/* Dark overrides — Figma dark text/primary #ffffff, plus the risk-card dark
+ * variants. Covers both the `dark` class (tailwind darkMode:"class") and the
+ * [data-theme="dark"] hook. */
 .dark,
 [data-theme="dark"] {
+  --twc-accent-primary: 0 0% 100%;
   --risk-amber: 255 179 0;
   --risk-green: 0 230 118;
   --risk-muted: 124 124 124;
--- a/packages/babylon-core-ui/src/components/Accordion/Accordion.css
+++ b/packages/babylon-core-ui/src/components/Accordion/Accordion.css
@@ -22,7 +22,7 @@
   }
 
   &-icon {
-    @apply text-secondary-strokeDark absolute right-0 top-1/2 -translate-y-1/2;
+    @apply text-accent-secondary absolute right-0 top-1/2 -translate-y-1/2;
     transition-property: transform;
     transition-duration: var(--motion-duration-icon, 200ms);
     transition-timing-function: var(--motion-ease-icon, cubic-bezier(0.4, 0, 0.2, 1));
--- a/packages/babylon-core-ui/src/components/Hint/Hint.tsx
+++ b/packages/babylon-core-ui/src/components/Hint/Hint.tsx
@@ -29,7 +29,7 @@ const STATUS_COLORS = {
 } as const;
 
 const ICON_COLOR = {
-  default: "text-secondary-strokeDark",
+  default: "text-accent-secondary",
   warning: "text-warning-main",
   error: "text-error-main",
 } as const;
--- a/packages/babylon-core-ui/src/widgets/sections/WalletMenu/components/WalletMenuSettingItem.tsx
+++ b/packages/babylon-core-ui/src/widgets/sections/WalletMenu/components/WalletMenuSettingItem.tsx
@@ -48,7 +48,7 @@ export const WalletMenuSettingItem: React.FC<WalletMenuSettingItemProps> = ({
                 attachToChildren={true}
                 offset={[20, 8]}
               >
-                <span className="inline-block align-middle text-secondary-strokeDark cursor-pointer ml-1">
+                <span className="inline-block align-middle text-accent-secondary cursor-pointer ml-1">
                   {infoIcon}
                 </span>
               </Hint>
--- a/services/vault/src/components/simple/DepositForm.tsx
+++ b/services/vault/src/components/simple/DepositForm.tsx
@@ -328,7 +328,7 @@ export function DepositForm({
         tooltip={COPY.deposit.form.pendingConfirmationTooltip}
         attachToChildren
       >
-        <InfoIcon size={16} className="text-secondary-strokeDark" />
+        <InfoIcon size={16} className="text-accent-secondary" />
       </Hint>
     </span>
   ) : null;


jonybur (dismissed):
The vault-scoped override at the tw-colors CSS-variable layer is the right seam - it keeps simple-staking's palette untouched while the utilities already reading `hsl(var(--twc-*) / <alpha>)` pick up the new triplets without a specificity race. Selector shape mirrors the existing risk-token dark block (`.dark, [data-theme="dark"]`) so both the class-based and data-attribute theme paths are covered. HSL conversions check out (#dddddd = 86.7%, #cccccc = 80%), and the description's own bundle-cascade check confirms the new declarations win. Noting the description's own merge-order call: this needs to land after #2184 so v2 isn't repainted.

services/vault/src/globals.css
```diff
@@ -61,10 +61,23 @@ html {
...
- * `dark` class (tailwind darkMode:"class") and the [data-theme="dark"] hook. */
+/* Palette overrides (Figma light: text/primary #000000, stroke/secondary
+ * #dddddd, stroke/primary #cccccc). core-ui's tw-colors themes are shared with
+ * simple-staking, so the vault re-points the --twc-* HSL triplets here rather
+ * than editing that palette. */
+.light,
+[data-theme="light"] {
+  --twc-accent-primary: 0 0% 0%;
+  --twc-secondary-strokeLight: 0 0% 86.7%;
+  --twc-secondary-strokeDark: 0 0% 80%;
```
gbarkhatov: Worth fixing before merge, or splitting into a follow-up and saying so in the body. `secondary-strokeDark` is not only a stroke in this codebase — it is also the muted foreground color. In light mode this takes it from `#387085` to `#cccccc`: contrast on white drops **5.49:1 → 1.61:1**, and **1.53:1** on `background-secondary` `#F9F9F9`.

Vault-rendered sites that go near-invisible in light:

- `Hint.tsx:32` (`ICON_COLOR.default`) — 11 vault `<Hint>` call sites pass neither `icon` nor `attachToChildren` and so take this default: `VaultDetailCard:268`, `FeesSection:31`, `BorrowMarketsTable:64`, `MarketStatsBar:6,22`, `CollateralInfoCard:42`, `BorrowMetricsCard:90,108,118`, `RepayDetailsCard:51,71`.
- `Accordion.css:25` (`&-icon` chevron) — rendered by `ExpandablePanel`, `FeesSection`, `UtxoSplitSelectorV3`, `VaultProviderSelectorV3`.
- `WalletMenuSettingItem.tsx:51`, reached via `BtcEthWalletMenu` in `Connect.tsx:112`.
- Vault-owned: `DepositForm.tsx:349`, `FeesSection.tsx:93`.

The fix is already established here: 5 vault `<Hint>` sites pass `icon={<InfoIcon size={16} className="text-accent-secondary" />}`. `accent-secondary` (`#666666`) is Figma `text/secondary` and is already correct in both themes — promoting it to the `Hint` default beats repeating the per-site workaround.

Worth noting the vault-scoping is doing real work here: core-ui components the vault does *not* render (`Select`, `Toggle`, `BTCFeeRate`, `ValidatorSelector`) all use this token as a foreground too, and are untouched precisely because the override stops at the vault.

Separately, `DepositForm.tsx:35` sets `disabled:!bg-secondary-strokeDark disabled:!opacity-100` on the deposit CTA. The `#cccccc` fill itself looks intended — wallet-connector already hardcodes `disabled:!bg-[#CCCCCC]` for light — but the label comes from core-ui's contained variant as `text-accent-contrast` `#ffffff`, and `!opacity-100` removes the usual `disabled:opacity-30`. Worth eyeballing in light.

Verification in the PR body stops at "the four variables resolve to the right hex," which is why none of this surfaced.

services/vault/src/globals.css
```diff
@@ -61,10 +61,23 @@ html {
...
-/* v3 risk-card dark overrides (Figma dark variants). Covers both the
- * `dark` class (tailwind darkMode:"class") and the [data-theme="dark"] hook. */
+/* Palette overrides (Figma light: text/primary #000000, stroke/secondary
+ * #dddddd, stroke/primary #cccccc). core-ui's tw-colors themes are shared with
+ * simple-staking, so the vault re-points the --twc-* HSL triplets here rather
+ * than editing that palette. */
+.light,
+[data-theme="light"] {
+  --twc-accent-primary: 0 0% 0%;
+  --twc-secondary-strokeLight: 0 0% 86.7%;
```
gbarkhatov: Informational, no change requested. After this, light-mode `strokeLight` `#dddddd` and `strokeDark` `#cccccc` sit one step apart, where before they were `#d7e1e7` and `#387085`. Anywhere light mode uses both to express two weights of rule, the distinction effectively disappears — core-ui's outlined button (`border-secondary-strokeDark`) inside a card (`border-secondary-strokeLight`) is the clearest case. That is what the Figma palette says, so this is a "know you are spending it" note rather than a request.

services/vault/src/globals.css
```diff
@@ -61,10 +61,23 @@ html {
...
 
-/* v3 risk-card dark overrides (Figma dark variants). Covers both the
- * `dark` class (tailwind darkMode:"class") and the [data-theme="dark"] hook. */
+/* Palette overrides (Figma light: text/primary #000000, stroke/secondary
+ * #dddddd, stroke/primary #cccccc). core-ui's tw-colors themes are shared with
+ * simple-staking, so the vault re-points the --twc-* HSL triplets here rather
+ * than editing that palette. */
+.light,
+[data-theme="light"] {
+  --twc-accent-primary: 0 0% 0%;
```
gbarkhatov: Half-migrated light palette — worth one line in the PR body. `primary-dark` `#12495E`, `primary-light` `#387085` and `accent-disabled` `#9ab7c2` stay teal-tinted, so pure-black `accent-primary` produces some odd pairings: `BatchedDepositGroup.tsx:133-134` becomes a black border around a `bg-primary-light/10` teal fill, and `StepRow.tsx:67` keeps a teal step marker beside now-black text. Fine to defer — just say whether v3 neutralizes those too.

services/vault/src/globals.css
```diff
@@ -61,10 +61,23 @@ html {
   --motion-duration-skeleton: 1.5s;
 }
 
-/* v3 risk-card dark overrides (Figma dark variants). Covers both the
- * `dark` class (tailwind darkMode:"class") and the [data-theme="dark"] hook. */
+/* Palette overrides (Figma light: text/primary #000000, stroke/secondary
+ * #dddddd, stroke/primary #cccccc). core-ui's tw-colors themes are shared with
+ * simple-staking, so the vault re-points the --twc-* HSL triplets here rather
```
gbarkhatov: `--twc-*` is tw-colors' private output format, not a core-ui knob. If the prefix changes, or someone renames `secondary.strokeLight` in `createThemes`, this silently reverts the vault to teal with no build, lint, or test failure.

Low risk in practice — core-ui pins `"tw-colors": "3.3.2"` exactly — but worth naming that pin right here so the coupling is visible to whoever bumps it. The public-API alternative if this grows: a second `createThemes({ light: {...}, dark: {...} })` in `services/vault/tailwind.config.ts`, where plugin order still puts the preset first and the vault's block still wins, with hex values instead of hand-computed HSL.

services/vault/src/globals.css
```diff
@@ -61,10 +61,23 @@ html {
   --motion-duration-skeleton: 1.5s;
 }
 
-/* v3 risk-card dark overrides (Figma dark variants). Covers both the
- * `dark` class (tailwind darkMode:"class") and the [data-theme="dark"] hook. */
+/* Palette overrides (Figma light: text/primary #000000, stroke/secondary
+ * #dddddd, stroke/primary #cccccc). core-ui's tw-colors themes are shared with
```
gbarkhatov: The comment lists Figma names and the code lists code tokens, but never maps them — and the mapping inverts (`stroke/secondary` → `strokeLight`, `stroke/primary` → `strokeDark`). Per-declaration comments would save the next reader the lookup:

```css
--twc-secondary-strokeLight: 0 0% 86.7%; /* Figma stroke/secondary #dddddd */
--twc-secondary-strokeDark:  0 0% 80%;   /* Figma stroke/primary   #cccccc */
```

Also worth stating that this block has to stay below the `@tailwind` directives: tw-colors emits `.light,[data-theme=light]` at identical specificity, so it is source order, not specificity, that makes the override win.

gbarkhatov (dismissed):
Approving — the mechanism is sound, and I verified it rather than taking the description at face value. One thing worth fixing before or right after merge (inline on the `strokeDark` line), plus a stale section in the description.

## What I verified

- **Variable naming is right, and the case matters.** tw-colors 3.3.2 builds names as `--twc-` + the flattened key joined by `-`, with no case folding (`flattenColors` → `defaultProduceCssVariable`), so `--twc-secondary-strokeLight` is correct camelCase. Confirmed against the built bundle: `.light,[data-theme=light]{…--twc-secondary-strokeLight:202.5 25% 87.5%…}`.
- **Selectors match exactly**, so this is pure source order rather than specificity — and the block sits after `@tailwind utilities`, while both palette copies (core-ui's `style.css` @import on line 3 and the vault's own Tailwind output) land earlier. No specificity race across the ~418 `accent-primary` call sites.
- **No opacity var to chase.** tw-colors only emits `--twc-<name>-opacity` for colors carrying alpha; all four here are opaque, so overriding the triplet alone is complete and `hsl(var(--twc-accent-primary) / var(--twc-accent-primary-opacity, var(--tw-text-opacity)))` keeps its fallback path intact.
- **HSL round-trips exactly**: `0 0% 86.7%` → 221.09 → `#dd`, `0 0% 80%` → 204 → `#cc`, matching tw-colors' own 1-decimal rounding.
- No core-ui file touched, so simple-staking is genuinely unaffected.

## The "Ordering" section is stale

#2184 (retire the `ENABLE_V3_UI` flag, delete the v2 shell) is merged; this branch is 2 commits behind `main` and still carries `isV3UiEnabled` in `featureFlags.ts`. Rebase and drop that section — the hunk applies cleanly. Post-rebase, re-check `V3_CTA_CLASSES` in `DepositForm.tsx`, which was flag-gated and interacts with the inline note on `strokeDark`.

## Other dimensions

- **Tests**: nothing meaningful applies to a CSS token change. The 3094 passing tests confirm no regression, not the change itself.
- **Performance**: four extra custom-property declarations. Negligible.
- **Security**: n/a — nothing near the critical paths in CLAUDE.md.
- **Conventions**: follows the existing `globals.css` shape (`--success-bright` / `--risk-*` with a `.dark` block) and correctly keeps app-specific values out of core-ui.

services/vault/src/globals.css
jeremy-babylonlabs: Fixed — you were right, and it is worse than the light-mode numbers suggest.

Confirmed `Hint.tsx:32` sets `ICON_COLOR.default = "text-secondary-strokeDark"`, and I reproduced your contrast figures exactly: **5.49:1 → 1.61:1** on `surface`, **1.53:1** on `background-secondary`. Below the 3:1 floor for UI components, and caused by this PR.

The finding that decided the mechanism: in **dark**, `strokeDark` `#5A5A5A` on `#121212` is **2.72:1** — so the `Hint` default was *already* failing 3:1 in both apps, before any palette change. That makes promoting `accent-secondary` a strict contrast improvement everywhere rather than a style change to simple-staking:

| | before | after |
|---|---|---|
| light (vault, post-override) | 1.61:1 | **5.74:1** |
| light (simple-staking) | 5.49:1 | **5.74:1** |
| dark (both apps) | 2.72:1 | **8.64:1** |

A vault-only fix was never available — `Accordion` and `WalletMenuSettingItem` are core-ui internals no amount of vault CSS reaches. So `6c1958ca` points all four foreground sites at `accent-secondary`:

- `Hint.tsx:32` `ICON_COLOR.default` (covers the 10 default-icon call sites)
- `Accordion.css:25` `&-icon` chevron (`ExpandablePanel`, `FeesSection`, `UtxoSplitSelectorV3`, `VaultProviderSelectorV3`)
- `WalletMenuSettingItem.tsx:51` (reached via `BtcEthWalletMenu`, `Connect.tsx:112`)
- vault-owned `DepositForm.tsx:331`

The palette in `createThemes` is still untouched — what changed is three components using a stroke token as a foreground.

Two corrections to the list. It is **10** default-icon sites, not 11: `VaultDetailCard:268` was deleted by #2184, so that line predates the rebase. And the vault also renders `Input`, which you did not list — it uses `strokeDark` only as a border, so it is fine. Your scoping claim otherwise holds exactly: `Select`, `Toggle`, `BTCFeeRate` and `ValidatorSelector` are not imported by the vault and are deliberately left alone.

**On the disabled CTA — no change, and I think the `#cccccc` fill is now more correct, not less.** `wallet-connector` already ships the identical pairing:

```ts
// Chains/index.tsx:9
const DISABLED_CONNECT_BG = "disabled:!bg-[#CCCCCC] dark:disabled:!bg-secondary-strokeDark";
// :63  className={twMerge("text-sm disabled:!opacity-100", DISABLED_CONNECT_BG)}
```

Same `!opacity-100`, same white `text-accent-contrast` label from the contained variant. That light-mode hardcode exists *because* the token was teal — so `#cccccc` is the intended disabled fill, and this PR makes the vault converge on the design system instead of being the teal outlier. WCAG 1.4.3 also exempts inactive controls from contrast minimums. If white-on-grey disabled labels are wrong, that is a design-system question affecting wallet-connector equally, and worth its own change.

Fair hit on the verification: "the four variables resolve to the right hex" was the wrong stopping point. I have added computed contrast to the checks.

services/vault/src/globals.css
jeremy-babylonlabs: Confirmed, and taking it as noted rather than changing anything.

Verified the collapse: `#d7e1e7` / `#387085` (a 4.0:1 ratio between the two) becomes `#dddddd` / `#cccccc` — 1.19:1 between them, so as a *pair* they carry almost no weight distinction in light. Your outlined-button-inside-a-card example is the real one; `DepositCardShell.tsx:74` (`border-secondary-strokeDark`) next to `RefundSuccessContent.tsx:23` (`border-secondary-strokeLight`) is the same shape.

Keeping it, because it is what `stroke/primary` and `stroke/secondary` are in the Figma file — I re-pulled both frames to be sure rather than trusting the table. If two rule weights need to stay distinguishable in light, that is a Figma-side conversation, not something to fix by keeping the teal.

services/vault/src/globals.css
jeremy-babylonlabs: Confirmed and added to the PR body — deferring, with the reasoning stated.

Verified the three stay teal in `createThemes` light: `primary.dark` `#12495E` (:228), `primary.light` `#387085` (:229), `accent.disabled` `#9ab7c2` (:214). And `StepRow.tsx:67`'s `bg-primary-light` marker is real — a teal step dot beside now-black text.

One correction: `BatchedDepositGroup.tsx` no longer exists, deleted by #2184, so that example predates the rebase. `StepRow` stands.

To answer directly: the v3 entry-screen frames do not define `primary/*` or `accent/disabled`, so I have no Figma authority to neutralize them from this ticket's frames alone. They are also not aliases of the tokens here — `primary-main` `#042F40` is already correct as Figma `accent/secondary`, so this is genuinely a separate mapping question rather than the same delta spread across more tokens. Scoping this PR to the four tokens the frames *do* specify, and flagging the rest for whoever owns the next palette pass.

services/vault/src/globals.css
jeremy-babylonlabs: Agreed on the coupling, documented it in `b650a0c0` rather than restructuring.

Verified the pin is exact — `"tw-colors": "3.3.2"` in `packages/babylon-core-ui/package.json:116`, no caret — so the failure mode you describe is real but gated behind a deliberate version bump. The comment now names it at the override site:

```
 * The --twc-<token> names are tw-colors' output format, coupled to the
 * exact "tw-colors": "3.3.2" pin and the key names in core-ui's createThemes.
```

On the `createThemes`-in-the-vault-config alternative: I agree it is the better shape if this grows, and it would take hex instead of hand-computed HSL. Not taking it now for two reasons. It depends on plugin ordering being stable across the preset merge, which is its own undocumented coupling rather than a strictly weaker one. And with four values it trades a visible dependency for a subtler one. If a fifth token lands here, that is the point to switch — worth doing then with the hex values, so the HSL conversion step disappears too.

services/vault/src/globals.css
jeremy-babylonlabs: Both taken, in `b650a0c0` — adopted your snippet essentially verbatim.

```css
.light,
[data-theme="light"] {
  --twc-accent-primary: 0 0% 0%; /* Figma text/primary     #000000 */
  --twc-secondary-strokeLight: 0 0% 86.7%; /* Figma stroke/secondary #dddddd */
  --twc-secondary-strokeDark: 0 0% 80%; /* Figma stroke/primary   #cccccc */
}
```

Plus the dark one (`/* Figma text/primary #ffffff */`). The inversion tripped me up too — I had flagged it in the PR body but leaving it out of the file was the wrong call, since the file is where someone edits these.

Source-order point also added to the block comment:

```
 * These blocks must stay below the @tailwind directives: tw-colors emits
 * `.light,[data-theme=light]` at the same specificity, so source order is what
 * makes the override win.
```

Worth recording that I verified this rather than reasoned about it: in the built bundle the tw-colors declarations land at `.light,[data-theme="light"]` and the override block follows them, and the cascade-winning values resolve to `#000000` / `#dddddd` / `#cccccc`. Equal specificity, later wins — exactly as you describe, which is why the ordering constraint deserved to be written down.

jonybur (approved):
Re-reviewed at `b650a0c0`. The two commits since my earlier pass fold in the strokeDark-as-foreground migration (Hint default icon, Accordion chevron, WalletMenuSettingItem info icon, DepositForm's pending-confirmation icon) so lifting the light strokeDark triplet to `#cccccc` doesn't drop any icon below the 3:1 UI floor. Selector shape (`.light,[data-theme=light]`), source-order justification, and the tw-colors 3.3.2 variable naming (camelCase preserved) all check out against the built CSS. The palette override stays scoped to the vault via `services/vault/src/globals.css`, so `simple-staking`'s teal palette is genuinely untouched.

The remaining vault call sites for `secondary-strokeDark` outside this migration are borders or dark-mode background hovers, not light-mode foreground, so the palette change doesn't regress them.

PR #2193, opened 2026-08-05
feat(vault): replace the entry screen with the v3 design

jeremy-babylonlabs:
## Summary

- Replaces the disconnected landing with the v3 design. Rebased onto the retired-flag `main`: rather than adding a parallel `…V3` component, the new screen takes over `DisconnectedOverview` / `FeatureCard` and the `overview.disconnected` copy block, and the v2 versions are deleted. Nothing is flag-gated — the flag is gone.
- Chips move above the headline and read live data: TVL from the BTC locked in the application priced by the oracle (suppressed rather than approximated when the round is stale or failed), deposit cap, and max borrow.
- The borrow-rate row moves into the left column under its own heading with token avatars and dividers; the feature rows become one bordered panel with hairline dividers, and the two expandable rows sit at positions 2 and 5.
- The sidebar and page-title header are dropped while disconnected — the entry frame has neither — and the navbar, page body and footer share one capped content box (`ENTRY_CONTENT_CLASS`): a 1280px cap plus the 40px gutter puts content at the frame's 1200px, 120px from each edge. The header divider sits outside that box and stays full-bleed.
- Adds `EntryFooter`, since the sidebar normally carries the social and legal links and this screen hides it. The existing mobile-only block is suppressed here so only one footer renders.
- The headline accents only the dot of the first "i" by painting the font's dotless glyph (U+0131) over a dotted one — the two share an advance width, so this needs no offsets and scales with the `clamp()` size. The heading carries the unsplit sentence as its accessible name, because splitting the word makes the computed name read "B i tcoin".
- The hero is the page's `h1` and the rates heading its `h2`: the page-title `h1` only renders once connected, so the entry screen would otherwise start its outline at `h3`.

## Testing

- `pnpm exec nx run-many --target=build --all`
- `pnpm exec nx run-many --target=lint --all` (0 errors)
- `pnpm exec nx run-many --target=test --all`
- `pnpm --filter vault exec tsc --noEmit`
- Checked both the light and dark frames against the running app in each theme.

## Follow-up

- The v3 light palette is neutral (`text/primary` `#000000`) where the app's light theme is teal-tinted (`accent-primary` `#12495E`), so light-mode body text still renders navy. That is an app-wide token change affecting every screen, tracked separately.


services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -183,9 +187,12 @@ export default function RootLayout() {
   const { isSupportedAddress } = useAddressType();
   const isMobileView = useIsMobile();
   const pageTitle = usePageTitle();
-  const showV3Sidebar = FeatureFlags.isV3UiEnabled && !isMobileView;
 
   const isWalletConnected = btcConnected && ethConnected;
+  // Disconnected v3 is the entry screen: no sidebar, brand lockup back in the
+  // header. Same connection signal DashboardPage swaps the landing on.
+  const isV3AppShell = FeatureFlags.isV3UiEnabled && isWalletConnected;
+  const showV3Sidebar = isV3AppShell && !isMobileView;
```
gbarkhatov: `showV3Sidebar` now requires a connected wallet, but the chrome that replaces the sidebar is gated on `isV3EntryLayout` (:260), which *additionally* requires `pathname === "/"`. Nothing reconciles the two on the other v3 routes.

No route has a connection guard, and `/vaults` and `/activity` render disconnected states on purpose (`VaultsPage.tsx:65` — "so the sections can be exercised without a wallet"). So a disconnected desktop visitor on `/vaults`, `/loans`, `/activity`, `/explore` or `/liquidations` now gets: no sidebar, no desktop nav (`navigation={undefined}` at :386), no page-title `h1`, no footer — and `PAGE_CONTENT_CLASS`, which is `!max-w-none … mx-10`, sized to fill the column beside a sidebar that isn't rendered. Before this PR the sidebar was there and carried the nav plus Terms/Privacy.

Mobile still recovers through the hamburger (`V3MobileNavigation` is unconditional for v3), so desktop is the dead end — and the two ends of the breakpoint now disagree about whether these screens have navigation at all.

Both consumers here seem to mean "is this the entry frame", not "is a wallet connected":

```ts
const isV3EntryLayout = FeatureFlags.isV3UiEnabled && !isWalletConnected && pathname === "/";
const isV3AppShell = FeatureFlags.isV3UiEnabled && !isV3EntryLayout;
```

That needs `useLocation()` hoisted above this line, but it collapses the two signals into one and the non-root routes keep their shell.

services/vault/src/components/simple/DisconnectedOverviewV3.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+// TVL is the BTC locked across the application priced in USD. Suppressed rather
+// than approximated when the oracle round is stale or failed, matching the
+// dashboard's guard — a headline figure from a bad feed is worse than none.
+function tvlStatValue(
+  capSnapshot: CapSnapshot | null,
+  btcPriceUsd: number | undefined,
+  isBtcPriceUsable: boolean,
+): string {
+  if (!capSnapshot || !isBtcPriceUsable || btcPriceUsd === undefined) {
```
gbarkhatov: Two ways this renders a real-looking `TVL: $0` instead of the placeholder.

**1. The dashboard's guard has a fourth term this one drops.** `DashboardPage.tsx:260-262` pairs the same `isBtcPriceUsable` with `btcPriceUsd !== undefined && btcPriceUsd > 0`. Nothing upstream rejects a zero answer — `chainlink/query.ts:296` is `Number(answer) / 10 ** decimals`, and `isPriceFresh` only checks round completeness and age — so a fresh round answering `0` arrives here with `isStale: false, fetchFailed: false`, and `formatCompactUsd(0, true)` returns `"$0"` (`formatting.ts:202`). The comment above says this matches the dashboard's guard; adding `btcPriceUsd > 0` would make that true.

**2. `capSnapshot.totalBTC` is itself a fallback.** Per `capMath.ts:21-29` it "falls back to `0n` only if the usage RPC errored", and `useApplicationCap` deliberately shields that error so the deposit form stays unblocked. So an errored usage read prices the protocol at `$0` here — and shows `Deposit Cap: 0/100 BTC` in the next chip. Worth threading the query's error state through so both fall back to `COPY.common.emptyValue`.

This is the most prominent number on the entry screen, and `$0 TVL` reads as a fact rather than a failure.

services/vault/src/components/simple/DisconnectedOverviewV3.tsx
```diff
@@ -0,0 +1,307 @@
...
+  return (
+    <div className="grid grid-cols-1 items-start gap-10 md:grid-cols-2 md:gap-12">
+      <div className="flex flex-col">
+        <div className="flex flex-wrap gap-4">
+          {statChips.map((chip) => (
+            <StatChip key={chip.label} label={chip.label} value={chip.value} />
+          ))}
+        </div>
+
+        <h3
```
gbarkhatov: The entry screen ends up with no `h1` at all: the header's page-title `h1` is gated on `isV3AppShell` (`RootLayout.tsx:194`), which is false while disconnected, so the document's first heading is this `h3`, followed by the `h4` at :271. `RootLayout.test.tsx:222` now asserts that absence, which locks it in.

Since the accessible name is already handled here via `aria-label` and the styling is all utility classes, promoting this to `h1` is a one-element change that gives the page a proper heading root.

services/vault/src/components/shared/EntryFooter.tsx
```diff
@@ -0,0 +1,89 @@
...
+                rel="noopener noreferrer"
+                className={`text-accent-primary ${LINK_CLASS}`}
+              >
+                <Icon size={24} title={name} />
+              </a>
+            ))}
+          </div>
+          <p className="text-sm tracking-[0.17px] text-accent-secondary">
+            <a
+              href="https://babylonlabs.io/terms-of-use"
```
gbarkhatov: These two URLs are now hardcoded in a third and fourth place — `SidebarFooter.tsx:58,67` has the same pair, and core-ui hardcodes them again in `Footer.tsx:112,121` and `StandardSettingsMenu.tsx:57,65`. `config/socialLinks.ts` exists precisely so this class of link "can't drift on a handle/URL change"; a `LEGAL_LINK_URLS` next to `SOCIAL_LINK_URLS`, consumed by both vault footers, would close it.

The `{" - "}` separator on :75 is also user-visible text inlined in a component rather than in `copy.ts`. Same in `SidebarFooter`, so that'd be a migration rather than a new violation.

services/vault/src/components/shared/EntryFooter.tsx
```diff
@@ -0,0 +1,89 @@
...
+ * full-bleed while the content sits in the same box as the navbar and body.
+ */
+export function EntryFooter() {
+  return (
+    <footer className="border-t border-secondary-strokeLight py-8">
+      <div
+        className={`${V3_ENTRY_CONTENT_CLASS} flex flex-col gap-8 sm:flex-row sm:items-start sm:justify-between sm:gap-4`}
+      >
+        <div className="flex flex-col gap-3">
+          <Logo className="h-6 w-auto text-secondary-main dark:text-accent-primary" />
```
gbarkhatov: core-ui's `Logo` wraps its SVG in a bare `<a href="/">` (`Logo.tsx:8`) with no `<title>`, `aria-label` or text content — so this is an unlabeled link in the accessibility tree, and a plain anchor rather than a `NavLink`, so clicking it full-reloads the SPA to the route the user is already on. Rendering the mark without the anchor, or wrapping a labeled `NavLink`, avoids both.

services/vault/src/components/simple/DisconnectedFeatureCards/FeatureCardV3.tsx
```diff
@@ -0,0 +1,58 @@
...
+          className={`text-xs leading-[1.66] tracking-[0.4px] text-accent-secondary ${showFull ? "block" : "line-clamp-1 max-w-sm"}`}
+        >
+          {body}
+        </span>
+      </div>
+      {expandable && (
+        <ChevronRightIcon
+          size={18}
+          variant="secondary"
+          className={`mt-1 shrink-0 transition-transform duration-[var(--motion-duration-icon)] ease-[var(--motion-ease-icon)] ${expanded ? "-rotate-90" : ""}`}
```
gbarkhatov: This file is a near-verbatim fork of `FeatureCard` — same props minus `extra`, byte-identical content JSX, differing only in the outer wrapper (`rounded bg-secondary-highlight` vs none) and this line's motion tokens.

Flag-scoped duplication that deletes wholesale is defensible, but note the fork is the only copy that got `--motion-*`: `FeatureCard.tsx:43` still has a bare `transition-transform` on the same chevron, so the two rotate at different speeds today. A `wrapperClassName`/`variant` prop on `FeatureCard` would fix v2's motion and drop this file in the same change.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -484,7 +498,8 @@ export default function RootLayout() {
               />
             </div>
           )}
-          {FeatureFlags.isV3UiEnabled && isMobileView && (
+          {isV3EntryLayout && <EntryFooter />}
```
gbarkhatov: `EntryFooter` has no `mt-auto`, unlike both sibling footers (:470 and :509). The comment at :245-253 spells out why that matters: "geo-loading/geo-blocked render short content instead of the centered Container, so the footer must keep `mt-auto` to stay at the viewport bottom."

`isV3EntryLayout` is deliberately geo-free — correct, it stops the navbar resizing mid-load — so this also renders on both geo branches, where nothing absorbs the free space. `GeoBlockState` is a fixed-height card, so a geo-blocked user at `/` gets the footer floating mid-page with whitespace beneath it, and unlike the loader that state persists. `mt-auto` on the footer (or a wrapper here) covers all three cases: the entry screen's own `my-auto` Container already consumes the slack.

services/vault/src/components/pages/__tests__/RootLayout.test.tsx
```diff
@@ -176,6 +182,46 @@ describe("RootLayout — header wiring", () => {
...
+  it("v3 flag on, disconnected on mobile: renders one footer, not the sidebar block too", () => {
+    featureFlagsMock.isV3UiEnabled = true;
+    mobileMock.value = true;
+
+    const { container } = renderRootLayout();
+
+    expect(container.querySelectorAll("footer")).toHaveLength(1);
+  });
+
+  it("v3 flag on, disconnected: hides the sidebar and restores the BrandLockup entry header", () => {
```
gbarkhatov: `renderRootLayout()` pins `initialEntries={["/"]}` with no way to override, so every case in this file is the root route — which is exactly why the non-root regression flagged on `RootLayout.tsx:195` isn't caught by the suite. A `renderRootLayout(path = "/")` parameter plus one case asserting the sidebar survives on a disconnected `/activity` would pin the intended boundary.

Minor and separate: :191 and its sibling at :163 assert on Tailwind class strings (`.\!max-w-\[1280px\]`). Understandable when width *is* the behaviour and there's no better handle, but it means a class rename that preserves the layout still fails the test.

gbarkhatov (dismissed):
## Review

Read the full diff plus the surrounding shell (`RootLayout`, `router.tsx`, `SidebarFooter`, `usePrices`, `useApplicationCap`, `useVaultSplitParams`, the Chainlink query client) at 7b480a2.

Approving — flag-off behaviour is genuinely untouched, and the two load-bearing claims in the description both hold up under checking:

- **"the two cannot disagree"** — `useConnection().isConnected` is literally `btcConnected && ethConnected` (`useConnection.ts:28`), so `RootLayout` and `DashboardPage` really do switch on the same value.
- **"the two share an advance width"** — checked the shipped font rather than taking it on trust: Px Grotesk's cmap contains U+0131 in both weights at an advance identical to U+0069 (Regular 256/256, Light 243/243 at 1000 upem). The glyph trick is safe, and handling the accessible name around it is the right call.

Also good: placeholder handling is more consistent than v2's (which mixes a literal `—` with `COPY.common.emptyValue`), the stale/failed oracle suppression is tested, and `usePrices` was already mounted by `DashboardPage`, so the TVL chip costs no extra network.

Two things I'd fix before this goes on by default:

1. **Disconnected v3 on any non-`/` route loses all navigation and the legal links** — the sidebar is gated on connection, its replacement on the route, and nothing reconciles the two. Detail inline on `RootLayout.tsx:195`.
2. **TVL can render a real-looking `$0` from bad data** — the guard is one term short of the dashboard's, and `totalBTC` is itself a `0n` fallback on an errored usage read. Detail inline on `DisconnectedOverviewV3.tsx:76`.

The rest are non-blocking. One concern I had about the webfont subset I dropped after the cmap check above came back clean.

services/vault/src/components/simple/DisconnectedOverviewV3.tsx
```diff
@@ -0,0 +1,307 @@
...
+                aria-hidden="true"
+                className="absolute inset-0 select-none text-accent-primary"
+              >
+                {HERO_TITLE.accentWord.dotless}
+              </span>
+            </span>
+            {HERO_TITLE.accentWord.after}
+          </span>
+          {HERO_TITLE.rest}
+        </h3>
```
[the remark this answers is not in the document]
jeremy-babylonlabs: Agreed, and fixed. Verified the premise first: `RootLayout` only renders the page-title `h1` once both wallets are connected, so on this screen there is genuinely no preceding heading and the outline started at `h3`.

Promoting only the hero would have left `h1` -> `h4` for the rates heading, so both moved: the hero is now `h1` and "Current Borrowing Rates" is `h2`. No visual change, since Tailwind preflight resets heading sizes and both carry explicit size/weight classes.

`DisconnectedOverview.test.tsx` pins the outline as `[1, 2]` so a future heading can't reintroduce a skip.

services/vault/src/components/pages/RootLayout.tsx
jeremy-babylonlabs: Confirmed and fixed — this was a real regression, and your diagnosis of the cause was exact.

Collapsed the two signals into one, as suggested: `isEntryLayout = !isWalletConnected && pathname === "/"`, with `showV3Sidebar = !isMobileView && !isEntryLayout`. The sidebar is now dropped only on the entry frame, so a disconnected desktop visitor to `/vaults`, `/loans`, `/activity`, `/explore` or `/liquidations` keeps the sidebar, the nav and `PAGE_CONTENT_CLASS` sized for the column that is actually there.

This narrowed the original instruction (hide the sidebar whenever disconnected) to "hide it on the entry frame", which is what the design actually shows and the only reading that doesn't strand those routes.

`RootLayout.test.tsx` now takes a path (`renderRootLayout(path = "/")`) and asserts the shell survives on a disconnected `/vaults` — the boundary was previously untestable because the helper pinned `initialEntries` to `/`.

services/vault/src/components/simple/DisconnectedOverviewV3.tsx
jeremy-babylonlabs: Both confirmed and fixed.

**1.** `btcPriceUsd > 0` added. You're right that the comment claimed parity with the dashboard's guard while dropping a term — a fresh round answering `0` passes `isPriceFresh`, and `formatCompactUsd(0, true)` short-circuits to `"$0"` before the compact branch.

**2.** `useApplicationCap` already returns `error`, so it's threaded to the screen as a `capError` prop and now suppresses both cap-derived chips — TVL and Deposit Cap — rather than trusting a snapshot whose `totalBTC` fell back to `0n`. `maxCf` is unaffected as it comes from `useVaultSplitParams`.

The comment now names all three untrustworthy inputs instead of claiming parity with the dashboard. Two tests cover it: an errored usage read, and a fresh round answering zero.

services/vault/src/components/simple/DisconnectedOverviewV3.tsx
jeremy-babylonlabs: Agreed and fixed. Verified the premise: the page-title `h1` is gated on the connected shell, so this screen genuinely had no heading root.

Promoting only the hero would have left `h1` -> `h4` for the rates heading, so both moved — hero to `h1`, "Current Borrowing Rates" to `h2`. No visual change, since Tailwind preflight resets heading sizes and both carry explicit size/weight classes. `DisconnectedOverview.test.tsx` pins the outline as `[1, 2]`.

The `RootLayout.test.tsx` assertion you flagged as locking in the absence is still there, but it now scopes to the header slot only — that test renders no route content.

services/vault/src/components/shared/EntryFooter.tsx
jeremy-babylonlabs: Agreed on both. Added `LEGAL_LINK_URLS` next to `SOCIAL_LINK_URLS` in `config/socialLinks.ts` and pointed both vault footers at it, so the pair is defined once rather than in four places. Left core-ui's own copies alone — separate package, separate change.

The separator moved to `COPY.footer.legalSeparator`, and I migrated `SidebarFooter` in the same commit rather than leaving a matched pair half-fixed.

services/vault/src/components/shared/EntryFooter.tsx
jeremy-babylonlabs: Confirmed — `Logo` hardcodes `<a href="/">` with no accessible name, and `SmallLogo` does the same, so neither can be labelled from the call site.

Swapped to `SidebarBrandLockup`, which is a plain `svg` with `role="img"` and `aria-label="Babylon Aave"` — no anchor, so both the unlabeled link and the SPA full-reload are gone, and it's the mark the entry header already uses.

Trade-off worth flagging: the frame's footer shows the Babylon wordmark alone, whereas this lockup includes the divider and Aave mark. Rendering the frame-accurate mark without an anchor would need a core-ui change (exporting the bare wordmark, or letting `Logo` take a wrapper), which felt out of scope here. Happy to go that way if you'd rather keep the footer mark exact.

services/vault/src/components/simple/DisconnectedFeatureCards/FeatureCardV3.tsx
jeremy-babylonlabs: Resolved by restructuring rather than by patching the fork.

Rebasing onto the retired-flag `main` removed the reason for a parallel file: the new screen now takes over `DisconnectedOverview` / `FeatureCard` and the `overview.disconnected` copy block, and the v2 versions are deleted. There is one `FeatureCard` again, and it's the copy carrying the `--motion-*` tokens — so the two-speeds-on-the-same-chevron problem is gone rather than inherited.

services/vault/src/components/pages/RootLayout.tsx
jeremy-babylonlabs: Confirmed and fixed — `mt-auto` added to `EntryFooter`.

Your reasoning about the geo branches is what makes it necessary: `isEntryLayout` is deliberately geo-free, so the footer also renders over `GeoBlockState`, which is a fixed-height card with nothing to absorb the slack — and unlike the loader that state persists. The entry screen's own `my-auto` Container already consumes the free space there, so this only changes the two geo cases.

services/vault/src/components/pages/__tests__/RootLayout.test.tsx
jeremy-babylonlabs: Fixed the first point: `renderRootLayout(path = "/")` now takes a path, and there's a case asserting the sidebar survives on a disconnected `/vaults` — which is exactly the regression flagged in the other thread, previously invisible to this suite.

On the class-string assertions: keeping them, with the reasoning you anticipated — the capped width *is* the behaviour, and there's no accessible handle for "the navbar is in the entry box". Noted that a class rename preserving the layout would fail them; the alternative I considered (a data attribute purely for tests) seemed worse than a brittle-but-honest assertion.

services/vault/src/components/shared/EntryFooter.tsx
jeremy-babylonlabs: Fixed properly rather than by the trade-off above — Jeremy confirmed the footer should carry the Babylon wordmark alone, so the frame stays exact and the anchor problem is still gone.

Root cause was broader than the anchor: `SidebarBrandLockup` paints both marks from one `currentColor`, so theming it made the navbar all-black and the footer all-orange. core-ui's `Logo` now exposes an anchor-free `LogoMark` (`role="img"`, `aria-label="Babylon"`), extracted from the existing SVG so `Logo` keeps wrapping it — additive, no change for other consumers.

- Navbar: a vault-side `BrandLockup` — themed wordmark + divider + Aave's own image, so Aave keeps its brand colours (Figma #113 / #121). Wrapped in `NavLink`, so no full page reload either.
- Footer: `LogoMark` alone, no anchor, no Aave mark.

`RootLayout.test.tsx` asserts the Aave image renders in the entry header, which pins the lockup since it appears nowhere else on that screen.

jonybur (approved):
Reviewed at `9a444be0`. Verified both open threads from @gbarkhatov's 7b480a2 pass are genuinely resolved by the follow-up commits:

- **Non-`/` disconnected navigation** — `RootLayout.tsx` now gates the sidebar hide on `!isWalletConnected && pathname === "/"` via a single `isEntryLayout` boolean, so a disconnected user on `/vaults` (or anywhere else) keeps the shell and its legal links. `EntryFooter` only replaces the sidebar's legal/social block on the root path.
- **TVL `$0` from bad data** — `tvlStatValue` in `DisconnectedOverview.tsx` now also gates on the deposit-cap read (`capError`, threaded from `DashboardPage`), on top of the existing `isBtcPriceStale` / `btcPriceFetchFailed` / `price > 0` guards. `totalBTC` no longer feeds a rendered TVL when the cap read errored.

Copy discipline holds: `DisconnectedOverview`, `EntryFooter`, and `BrandLockup` route user-visible strings through `COPY` / `socialLinks`, no inlined literals. Presentation stays in `FeatureCard`; the live-data logic and error gates stay in the container.

PR #2194, opened 2026-08-05
feat(core-ui): line-chart primitive on a shared chart foundation

jeremy-babylonlabs:
Screenshot 2026-08-05 at 23 04 48
Screenshot 2026-08-05 at 23 04 57

## Summary

Adds the shared line-chart primitive to `@babylonlabs-io/core-ui`, so the interest-rate-model curve and the borrow-APR history chart can both be built on one renderer.

Most of the scaffolding already existed under `components/LiquidationChart/` but was misfiled there. This extracts the genuinely generic parts into `components/charts/` and adds the piece that was actually missing: a plotted line series. `SeizureMap` and `Timeline` draw bands and candles; neither draws a line.

## What moved

`components/charts/` now holds the chart-agnostic scaffolding:

| Module | Notes |
| --- | --- |
| `chartLayout.ts` | Pixel geometry + `useChartLayout` (lifted out of `ChartFrame`), now with an optional plot aspect ratio |
| `textMeasure.ts` | Canvas text measurement, moved unchanged |
| `axisDeclutter.ts` | 1-D label displacement, moved unchanged |
| `types.ts` | `ChartAxisTick`, `ChartGridConfig` |
| `chartTokens.css` | Neutral chrome tokens (`--bbn-chart-*`) |

`ChartFrame` stays with `LiquidationChart` — despite a low grep count for "liquidation" it is built around a price rule, tone pills and `--liq-*` styling, so it is not reusable as-is. Same for `priceScale.ts`, whose segmented and span-weighted scales exist for the seizure map specifically.

`LiquidationChart.css` keeps every `--liq-*` token name (the vault references several of them) and now derives the neutral ones from the shared tokens, so there is one source of truth and no consumer-visible change.

## The primitive

```tsx
<LineChart
  data={points}
  interpolation="linear" | "step"
  xDomain={[min, max]}  yDomain={[min, max]}   // omit either to fit the data
  yDomainPadding={0.05}                        // headroom on a fitted axis
  xTicks={ticks} yTicks={ticks}                // labels pre-formatted by the app
  markers={markers}                            // rule + dot + collision-aware callout
  hoverMode="interpolate" | "nearest"
  renderTooltip={fn} onHoverChange={fn}
  aspectRatio grid color animate ariaLabel
/>
```

It carries no domain semantics — no utilisation, APR, price or BTC. Every label, tick and callout string is supplied pre-formatted by the app.

Callout collisions resolve in **array order**: a callout hangs off its marker's rule on a preferred side and only flips when it would overlap an already-placed callout or leave the plot. The marker listed first keeps its preferred side.

## Test plan

Storybook: **Components → Data Display → Charts → LineChart** (14 stories).

- `AnnotatedMarkersColliding` / `AnnotatedMarkersClear` — the two callout-collision cases from the design reference
- `AnnotatedMarkersCrowded` — callouts stack rather than escape the plot
- `HoverAnywhere` — reads an interpolated value between data points
- `RateHistory` — laid out as the design has it: stepped line, fitted (non-zero-based) axis, horizontal-only gridlines, date + value tooltip, and a host-owned heading, range figure and 1D/1W/1M/6M/1Y/All selector defaulting to 1W
- `Narrow` — responsive layout and collision at a small width

Checked in both light and dark.

Automated: 41 new unit tests over domain fitting, step interpolation, nearest-point lookup, callout displacement and the touch readout. `pnpm test` in core-ui 98/98 and in vault 3094/3094; `tsc --noEmit`, `eslint` and the core-ui build all clean.

`SeizureMap` and `Timeline` are behaviourally untouched and their tests pass unmodified.

## Notes for reviewers

- `src/test/setup.ts` now stubs `PointerEvent`. jsdom does not implement it, so `@testing-library` silently falls back to a plain `Event` and drops `clientX` — pointer-driven interactions read `NaN` coordinates without this. Any future pointer-driven component test depends on it.
- The chart reads pointer position from the hit rect's client rect rather than `@visx/event`'s `localPoint`, whose non-CTM fallback path returns `NaN` on SVG elements. The SVG carries no `viewBox`, so one user unit is one CSS pixel and the arithmetic is exact.
- Touch: a tapped readout persists until the next tap. Touch pointers are destroyed on lift, so the browser fires `pointerleave` straight after `pointerup` — clearing there made a tap flash the readout and lose it. Only a mouse leaving clears now, which is the tap-is-hover behaviour product asked for. The hit area uses `touch-action: pan-y` rather than `none`, so a swipe over the plot still scrolls the page while horizontal movement drags the readout.
- Entry animation is a single opt-in reveal driven by `--motion-*` tokens with per-property fallbacks, per `docs/motion-system.md` — knobs here, values in the consuming app. Reduced motion is handled by the existing global reset.

## Out of scope

The two charts themselves and any `services/vault` page wiring. This is the primitive plus its Storybook proof only.


jonybur (approved):
Reviewed at `15f79c0f`. Presentation-only, no critical-path exposure.

Spot-checked the load-bearing pieces:

- `nearestIndex` — binary search bounds hold: `hi` starts at `points.length - 1` and never grows, `lo` only advances to `mid + 1` where `mid < hi`, so both `before` and `after` stay in range for any non-empty series. The empty case short-circuits to `-1`. The `-100` / `100` extrapolation tests exercise the endpoints without an OOB.
- `valueAtX` — bracket search sits inside explicit `x <= points[0]` / `x >= last` guards, so `hi - lo > 1` never underflows; step vs. linear only differ inside the bracket, which is what the tests pin.
- `fitDomain` filters non-finite points before selecting extrema, so a series with a NaN doesn't collapse the axis.
- Pointer handling — the mouse-vs-touch split on `pointerleave` (mouse clears, touch keeps the readout) is the behaviour the description promises and matches the test in `LineChart.test.tsx`. The `PointerEvent` shim in `src/test/setup.ts` extends `MouseEvent` so `clientX`/`clientY` survive jsdom; without it interaction tests would silently read `NaN`, which is what the shim is documented to prevent.
- `SeizureMap` / `Timeline` rewiring — both now import geometry/text-measure/types from `../charts/`; `ChartFrame` keeps its liquidation-specific renderer, and `--liq-*` token names are preserved with values still derived from the shared `--bbn-chart-*` tokens, so external consumers see the same computed values.
- Entry animation reads `var(--motion-...)` with per-property fallbacks and is gated by the opt-in `animate` prop; reduced-motion is handled by the existing global reset.

Storybook coverage matches the annotated-marker collision cases the description calls out, and the 41 new unit tests cover the domain/interpolation/callout math directly rather than through the component.

PR #2198, opened 2026-08-05
Add security model file

kirugan:
Please note that we already have `SECURITY.md` but it was before security model and I decided to put `SECURITY_MODEL.md` not for entire project, but for vaults specifically. Maybe we should remove `SECURITY.md` ? wdyt?


jonybur (dismissed):
Documentation-only. Reads clean and internally consistent:

- The invariants (1-11) are meaningful and testable, not aspirational.
- The imported-guarantees table correctly attributes byte-stable derivation and PSBT construction to the SDK/WASM layer rather than claiming them here.
- The four known deviations each name the specific invariant they violate and their current mitigation, which is the honest framing.
- All referenced paths resolve: `SECURITY.md`, `docs/specs/derive-vault-secrets.md`, `docs/specs/derive-context-hash.md`, and the `CLAUDE.md` critical-paths anchor all exist.

The `*TBD*` tracking-issue placeholders in the deviations table are self-flagged by the doc itself ("Each needs a tracking issue before this document is merged"), so I'm treating that as an in-flight follow-up rather than a review block.

jrwbabylonlab:
> Please note that we already have `SECURITY.md` but it was before security model and I decided to put `SECURITY_MODEL.md` not for entire project, but for vaults specifically. Maybe we should remove `SECURITY.md` ? wdyt?

How would it work with the ts-sdk part given the new file is under the vault? couldn't we place it at the root but make it only point to the ts-sdk and the FE vault only?

services/vault/SECURITY_MODEL.md
```diff
@@ -0,0 +1,213 @@
+# Security model — Babylon Vault dApp (`services/vault`)
+
+> Status: first iteration, pending component-owner review.
```
gbarkhatov, line 3: **Status vs. normative weight.** This line says "pending component-owner review", but the CLAUDE.md bullet landing in the same PR makes the file binding on every vault change ("treat its invariants as constraints on the change"). An unreviewed document becomes a constraint the moment it merges.

Either get owner sign-off before landing the CLAUDE.md rule, or split the PR: doc now, rule after review.

services/vault/SECURITY_MODEL.md
```diff
@@ -0,0 +1,213 @@
...
+pre-signs a transaction graph proposed by a third party, and reveals an HTLC preimage on Ethereum.
+Each of those actions is irreversible, is attributed to the user, and has no administrative undo.
+
+Normative references:
+
+| Concern | Document |
+| --- | --- |
+| TBV protocol, contracts, enforcement | [`babylonlabs-io/vault-contracts`](https://github.com/babylonlabs-io/vault-contracts), [`babylonlabs-io/btc-vault`](https://github.com/babylonlabs-io/btc-vault) |
+| Vault-secret and context-hash derivation | [`docs/specs/derive-vault-secrets.md`](../../docs/specs/derive-vault-secrets.md), [`docs/specs/derive-context-hash.md`](../../docs/specs/derive-context-hash.md) |
+| Repository threat analysis, attack matrix, severity rubric, vulnerability disclosure | [`SECURITY.md`](../../SECURITY.md) |
```
gbarkhatov, line 22: Two things about this row.

**Content.** Root `SECURITY.md` is 895 lines titled `# Security Model — babylon-toolkit`, and its sections 1–4 are per-file invariants for *this* component — not "repository threat analysis" in the sense this row implies. A reader deciding which of the two documents to open will pick wrong. If both files survive, retitle root's H1 so the names stop colliding.

**Mechanics.** The branch is 27 commits behind main, and its merge-base (`b4474f4d`) predates `a5e3bfaa` (#2152), which added root `SECURITY.md`. Every `../../SECURITY.md` link — this one and the one at line 213 — is dead while browsing the PR branch. Rebase so reviewers can click them.

services/vault/SECURITY_MODEL.md
```diff
@@ -0,0 +1,213 @@
...
+- **The derivation seam matches its Rust reference byte for byte**, as fixed by the specs and the
+  pinned WASM source revision.
+- **The contracts enforce the protocol correctly**, and the vault-provider proxy enforces its SSRF
+  policy on provider-supplied URLs.
+
+## Non-claims, accepted risks, and known deviations
+
+### Non-claims
+
+- **This component is not an enforcement boundary.** It is a static bundle running in a
```
gbarkhatov, line 179: This Non-claims block restates root `SECURITY.md`'s "Scope and non-goals" (lines 64–81) nearly item-for-item: "not an enforcement boundary" ↔ 64–70, "no protection against local compromise" ↔ 75–78, "no user authentication" ↔ 79–81, and the deployment-layer accepted risk below ↔ 71–74.

Not an argument against stating non-claims here — a component model should. It is an argument that one of the two copies has to become a pointer, otherwise they drift and a reader has no way to tell which one is current.

services/vault/SECURITY_MODEL.md
```diff
@@ -0,0 +1,213 @@
...
+### Accepted risks
+
+- A user bypassing advisory client-side checks on their own machine.
+- Operator configuration errors shipping as a release. Mitigated by review, not by any runtime
+  control, because configuration is baked in at build time.
+- Deployment-layer controls (HSTS, framing policy, TLS) that a static bundle cannot establish.
+
+### Known deviations
+
+Invariants that are currently violated or incompletely enforced. Each needs a tracking issue before
```
gbarkhatov, line 202: The document sets a merge condition it does not meet: all four Issue cells below are `*TBD*`.

File the four issues and fill the cells, or drop the sentence. As written, the file states that it is not mergeable.

services/vault/SECURITY_MODEL.md
```diff
@@ -0,0 +1,213 @@
...
+this document is merged.
+
+| # | Deviation | Invariant | Impact | Current mitigation | Issue |
+| --- | --- | --- | --- | --- | --- |
+| 1 | Recovery-artifact bodies are validated only structurally, from the envelope prefix. A hostile provider can return a well-formed envelope wrapping a corrupt body; the client then records the artifacts as obtained and stops warning. | 8 | Loss of independent claim capability, discovered only at claim time. Treat as fund recovery, not robustness. | None client-side. Full validation is deferred until the backend supports streaming delivery. | *TBD* |
+| 2 | Address screening fails open when its endpoint is unset or malformed. | 10 | Compliance bypass. Bounded by the non-claim above: screening is advisory in any case. | Startup warning only. | *TBD* |
```
gbarkhatov, line 208: **The invariant mapping is off.** Invariant 10 narrows itself to "network, chain, or contract set"; a screening endpoint is none of those. Either broaden invariant 10 to cover any security-relevant operating parameter the app cannot establish, or add an invariant this row can point at.

**Separately**, "Bounded by the non-claim above: screening is advisory in any case" undersells the row rather than qualifying it. The advisory non-claim covers a user editing their own bundle. It does not cover a typo'd `NEXT_PUBLIC_TBV_UTILS_API` shipping a build with screening disabled for *everyone* and no user-visible signal — an operator-error failure with a real fix (a production startup gate), which root `SECURITY.md` already recommends. Worth keeping as a deviation rather than folding into non-claims.

services/vault/SECURITY_MODEL.md
```diff
@@ -0,0 +1,213 @@
...
+| 3 | Egress is not constrained to known hosts, so exfiltration of derived secrets is prevented by discipline — field denylists, absence of HTML sinks — rather than by a runtime boundary. | 7 | Confidentiality of vault-secret material under an in-origin adversary. | Redaction contract in telemetry; no HTML sinks in source. | *TBD* |
+| 4 | The entry document itself is not integrity-protected by subresource hashing; only its assets are. | 11 | Origin write access remains the residual trust for the whole application. | Scoped deployment credentials; per-release hash manifests. | *TBD* |
+
+Deviations 1–4 are stated at model level. Their implementation detail, adversary mapping, and test
```
gbarkhatov, line 212: This says the detail is "not repeated here", but the table directly above repeats Impact and Current mitigation for all four — at lower fidelity than root `SECURITY.md`'s attack matrix, which already marks the same four as **Known gap**: artifact body validation, screening fail-open, `connect-src https:` egress, and `index.html` not being SRI-protected.

Two documents holding the same gap register, with no shared identifiers — this file numbers deviations against its own invariants 1–11, root's matrix keys rows by adversary letters A–G — is a silent-drift generator. Reduce this table to pointers plus the invariant→gap mapping and let root hold the single authoritative register (or the inverse), but not both in their current form.

CLAUDE.md
```diff
@@ -219,6 +219,7 @@ These paths handle irreversible value movement. An AI-generated mistake here is
 
 ## SECURITY
 
+- **Security models live next to the code they describe.** A security-critical component carries a `SECURITY_MODEL.md` at its root, stating what that component must protect, against whom, and under which assumptions. Before changing a component, read the nearest `SECURITY_MODEL.md` at or above the files you are touching, and treat its invariants as constraints on the change rather than as a checklist to satisfy afterwards. Coverage is being added component by component — the absence of a model is not evidence that a component is uncritical.
```
gbarkhatov, line 222: **Coverage hole in the rule as written.** "Read the nearest `SECURITY_MODEL.md` at or above the files you are touching" resolves to nothing for `packages/babylon-ts-sdk/**` and `packages/babylon-tbv-rust-wasm/**` — which is exactly where SECURITY_MODEL.md:155 says properties 1–6 and 9 are actually enforced, and where critical paths 1–4 and 7 from the section above live. The rule is silent precisely on the highest-risk code in the repo.

The trailing caveat ("absence of a model is not evidence that a component is uncritical") is honest but points nowhere. Add a fallback clause: when no component model exists at or above the files, root `SECURITY.md` applies — it already covers both packages in depth. Better still, land the SDK/WASM model in the same series so the rule has teeth on day one.

gbarkhatov (dismissed):
Approving — docs-only, the writing is strong, and I checked the substance rather than taking it on trust. The boundary/seams framing, the imported-vs-exported guarantees table, and the non-claims section are the parts most security docs skip.

**Verified:** all four deviations hold against both the code and root `SECURITY.md` (screening fail-open via `parseOptionalUrl`, envelope-prefix-only artifact validation, `connect-src https:`, SRI covering assets but not `index.html`). Every relative link resolves against `main`. Anchors are correct, including the non-obvious double hyphen in `#critical-paths--human-review-required`.

Two things worth settling before merge, plus an answer to your question.

## On "maybe we should remove `SECURITY.md`?" — no, don't

Three things in it are repo-scope and have no component to sit next to:

- **Disclosure.** GitHub surfaces a repository's security policy from a file named exactly `SECURITY.md` in root, `.github/`, or `docs/`. Deleting it removes that entry and the pointer contributors see when opening an issue.
- **The severity rubric**, which is an org standard versioned across repositories.
- **Supply chain / CI, published packages, and `simple-staking`** — none of which live inside `services/vault`.

## The real problem is that both files are now the vault's security model

Root `SECURITY.md` is 895 lines, titled `# Security Model — babylon-toolkit`, and its sections 1–4 plus most of its attack matrix are about `services/vault` specifically. The overlap with this file is structural, not incidental — see the inline notes on lines 179, 208, and 212.

A split that works:

- **Root** = disclosure, severity rubric, supply chain / CI, published packages, cross-service map, `simple-staking`, and the single authoritative gap register. Retitle its H1 so it stops colliding with this file.
- **Component** = boundary, actors, invariants, non-claims, with the deviations table reduced to pointers plus the invariant→gap mapping.

That is also the version that makes the new CLAUDE.md rule mean something, since the code pointers and test evidence would then sit next to the code they describe.

## One housekeeping item this PR should carry

Root `SECURITY.md`'s "When to update this file" section names five hand-maintained inventories that must change together, and notes "They have drifted before." This PR makes it six without touching that bullet.

## Unrelated, but cheap while you're already in the file

CLAUDE.md:93 (critical path 7) says `disableTweakSigner: true`; the actual flag is `useTweakedSigner: false` (`packages/babylon-ts-sdk/src/tbv/core/utils/signing.ts:28`). Root `SECURITY.md` invariant 5 has it right, so CLAUDE.md is the stale copy.

services/vault/SECURITY_MODEL.md
```diff
@@ -0,0 +1,213 @@
...
+A user bypassing an advisory client-side check on their own machine is **not** a security failure of
+this component. See [Non-claims](#non-claims-accepted-risks-and-known-deviations).
+
+## Critical security properties / invariants
+
+1. **No signed value is taken on trust.** Every value entering a signature or an on-chain commitment
+   is independently re-derived, or asserted against a source other than the party that proposed it.
+   Values originating from a counterparty, the indexer, browser storage, or interface state are
+   never signed verbatim.
+2. **Every transaction the depositor signs is constructed locally**, from data sourced on-chain —
```
jrwbabylonlab, line 129: `signDepositorGraph` signs over VP-supplied `payout_tx`, `assert_tx` and `nopayout_tx` hex, so this is not accurate as stated.

cyphersnake:
Please add me as a reviewer — I'm reviewing all `SECURITY_MODEL.md` docs across the repos and want to keep the cross-repo context consistent. [Thread](https://babylonlabsworkspace.slack.com/archives/C07ES6DG71P/p1785921721850349?thread_ts=1785173072.502849&cid=C07ES6DG71P).

cyphersnake (approved):
Approved for the next Baby Auditor run. I will carry out the detailed review asynchronously and log any findings as separate GitHub issues. I need a little more time to get fully onboard with the FE 😅

services/vault/SECURITY_MODEL.md
kirugan: fixed

services/vault/SECURITY_MODEL.md
kirugan: Kept both in this PR, but narrowed what the rule does with an unreviewed model. The CLAUDE.md bullet now distinguishes the two cases: a model reviewed by its component owner is binding; one still marked *pending component-owner review* is informative — read it, and raise a conflict with it on the PR rather than treating it as settled.

The status line and the rule now agree, and this file gains teeth when you sign it off rather than when it merges.

services/vault/SECURITY_MODEL.md
kirugan: **Content.** Agreed, the row misdescribed root. Root isn't "repository threat analysis" — its domain sections are the deepest per-file description of this component that exists. The row now reads "adversaries A–G, per-file controls and their code pointers, the authoritative gap register, severity rubric, vulnerability disclosure", and I added a short *which of the two security documents to read* note: root is authoritative on control detail and where the code is, this file on what must remain true. Neither is a summary of the other.

I left root's H1 alone for now — say the word and I'll retitle it in a follow-up.

**Mechanics.** Rebasing onto main, after which the links resolve on the branch.

services/vault/SECURITY_MODEL.md
kirugan: Agreed that one copy has to become a pointer. The four that duplicate root's *Scope and non-goals* (enforcement boundary, local compromise, user auth, deployment layer) are now one sentence pointing at that section, named only so a reader of this file knows they exist, with "if this file and root ever disagree on one of them, root is correct."

Only the three root doesn't cover stay stated here: durability of resume state, protocol correctness and economic outcomes, chain privacy.

services/vault/SECURITY_MODEL.md
kirugan: Dropped the sentence and the four `*TBD*` cells. Root's matrix already carries all four as **Known gap** with a remediation column, so a parallel issue register here was the wrong place to track them.

services/vault/SECURITY_MODEL.md
kirugan: Both taken.

**Mapping.** Invariant 10 now covers any security-relevant operating parameter, explicitly including one whose absence silently disables a control — "a missing endpoint that turns a check into a no-op is an unestablished parameter, not a configured absence" — so row 2 points at something real.

**Framing.** You're right that the advisory non-claim was doing work it can't do. The row now says the fail-open is operator error with a real fix (a production startup gate), not a user editing their own bundle. *Accepted risks* is narrowed to the valid-but-wrong config case and explicitly excludes a malformed or absent parameter that disables a control.

services/vault/SECURITY_MODEL.md
kirugan: Reduced to pointers. The table no longer restates Impact or Current mitigation — it's now invariant broken → gap → the root matrix row that owns it (`Artifacts`/A, `Screening`/G, `CSP`/D, `Delivery`/D), under an explicit statement that root holds the single register and this table exists only to answer what root can't: which invariant of *this component* each gap breaks.

Added the sync rule: gap closed in root, row deleted here; new gap, root first and mapped here second.

CLAUDE.md
kirugan: Fixed. The rule now carries the fallback: when no component model exists at or above the files you're touching, the repository-level `SECURITY.md` applies — naming `babylon-ts-sdk` and `babylon-tbv-rust-wasm`, since that's where most of the critical paths are actually enforced. The closing caveat now ends "…and never leaves a change unconstrained" rather than pointing nowhere.

Agreed the SDK/WASM model is the right next step; not in this PR.

jonybur (approved):
Approving. The revised structure resolves the prior round: the CLAUDE.md rule now has an explicit fallback to root SECURITY.md for packages/babylon-ts-sdk and packages/babylon-tbv-rust-wasm, so the coverage hole is closed, and the deviations table is pointer-only with the invariants-to-root-matrix mapping preserved rather than duplicated.

Cross-checked the referenced paths and anchors - SECURITY.md, the two docs/specs files, and the #critical-paths--human-review-required double-hyphen anchor all resolve at head. The self-imposed TBD blocker has been dropped along with the four TBD cells, which is the right call given root's attack matrix already tracks all four as Known gaps.

Root vs component split remains a broader structural question worth its own PR, but nothing this file introduces is factually wrong about the code, and the CLAUDE.md rule usefully binds the model to future vault changes.

PR #2200, opened 2026-08-06
fix(vault): line up the vault list rows and let their text fit

jonybur:
Screenshot 2026-08-06 at 03 34 08

## What

The rows on the Vaults page — Pending Deposit, Active Vaults and Inactive Vaults — now read as one table. Before this, each row sized itself independently, so:

- Rows sat at slightly different heights.
- Columns started at different places from one row to the next, so "Pending", "Available" and "Expired" did not line up down the page.
- The refund notice on an inactive vault was cut off mid-sentence with an ellipsis.
- A large gap opened up between the last column and the button on the right.
- The round provider logo was squeezed into an oval whenever the text next to it ran long.

The Activity page rows had the same spacing problem and are included.

## Why

All five symptoms came from how the row shared out its width. Four of the five cells were pinned to a fixed size and only one was allowed to stretch, so every spare pixel piled into that single cell — which is where the empty gap came from. The text next to the logo had nowhere to go, so the browser shrank the logo instead. And because a row without a button did not reserve the button's space, its columns slid left and stopped matching its neighbours.

## How

Every cell now stretches, so spare width is shared out along the row rather than collecting in one place.

The first cell, which holds the amount and the line beneath it, is given enough room for the longest message these rows can show — measured in the app at its real font size rather than guessed. That sentence now reads in full.

Rows share a minimum height so they stand level, and a row whose button only appears in certain states reserves that space regardless, keeping its columns aligned with the rest. The logo is no longer allowed to shrink.

Rows stay on a single line down to 1366px wide. Below that they wrap onto two lines, as they already did on narrow screens, rather than cramming the content.

## Testing

Checked in the running app at 1366, 1440 and 1512 wide, measuring the live page rather than eyeballing it: row heights identical, column positions identical across all three sections, no cut-off text, and every logo perfectly round. Lint and typecheck clean; the full vault suite passes (3042 tests).

### DIFF
--- a/services/vault/src/components/Activity/ActivityRowV3.tsx
+++ b/services/vault/src/components/Activity/ActivityRowV3.tsx
@@ -12,6 +12,7 @@
 import { Avatar } from "@babylonlabs-io/core-ui";
 import type { ReactNode } from "react";
 
+import { LIST_ROW_COLUMN_CLASS } from "@/components/shared/ListRow";
 import { COPY } from "@/copy";
 import { type ActivityLog, PENDING_DEPOSIT_TYPE } from "@/types/activityLog";
 import { getExplorerTxUrl } from "@/utils/explorer";
@@ -20,9 +21,9 @@ import { formatActivityTime } from "@/utils/formatting";
 import { ActivityHashLink } from "./ActivityHashLink";
 import { STATUS_DOT } from "./statusDot";
 
-/** Fixed column width shared by every cell, so rows line up across cards
- *  regardless of amount length (Figma: 180px). */
-const COLUMN_CLASS = "flex w-[180px] shrink-0 items-center";
+/** Column shared by every cell, so rows line up across cards regardless of
+ *  amount length. */
+const COLUMN_CLASS = `flex items-center ${LIST_ROW_COLUMN_CLASS}`;
 
 /** Figma body 2 — every cell's text except the deferred USD sub-line. */
 const CELL_TEXT_CLASS = "text-sm leading-[1.43] tracking-[0.17px]";
--- a/services/vault/src/components/Activity/ActivityRowV3.tsx
+++ b/services/vault/src/components/Activity/ActivityRowV3.tsx
@@ -12,7 +12,10 @@
 import { Avatar } from "@babylonlabs-io/core-ui";
 import type { ReactNode } from "react";
 
-import { LIST_ROW_COLUMN_CLASS } from "@/components/shared/ListRow";
+import {
+  LIST_ROW_ACTION_SLOT_CLASS,
+  LIST_ROW_COLUMN_CLASS,
+} from "@/components/shared/ListRow";
 import { COPY } from "@/copy";
 import { type ActivityLog, PENDING_DEPOSIT_TYPE } from "@/types/activityLog";
 import { getExplorerTxUrl } from "@/utils/explorer";
@@ -117,7 +120,10 @@ export function ActivityRowLayout({
   return (
     <div className="flex w-full flex-wrap items-center gap-4">
       <div className={`${COLUMN_CLASS} gap-2`}>
-        <Avatar url={icon} alt={iconAlt} size="medium" />
+        {/* `.bbn-avatar` is inline-flex with no shrink-0 of its own, and this
+            cell can now shrink to its 120px floor — without this a long amount
+            squashes the circle into an ellipse. */}
+        <Avatar url={icon} alt={iconAlt} size="medium" className="shrink-0" />
         <span
           className={`min-w-0 truncate ${CELL_TEXT_CLASS} text-accent-primary`}
         >
@@ -148,9 +154,14 @@ export function ActivityRowLayout({
         {time}
       </span>
 
-      {action && (
-        <div className="ml-auto flex shrink-0 items-center">{action}</div>
-      )}
+      {/* Reserved even when there is no action. Only refundable expired
+          deposits get a button and liquidation-group children never do, so
+          without the empty slot those rows would hand the button's width back
+          to the five cells and start their columns tens of pixels off the rows
+          around them. */}
+      <div className={`${LIST_ROW_ACTION_SLOT_CLASS} items-center`}>
+        {action}
+      </div>
     </div>
   );
 }
--- a/services/vault/src/components/ApplicationLogo.tsx
+++ b/services/vault/src/components/ApplicationLogo.tsx
@@ -8,14 +8,17 @@ interface ApplicationLogoProps {
   shape?: "circle" | "rounded";
 }
 
+// `shrink-0` on every size: the logo sits beside truncating text in flex rows,
+// where without it flexbox squashes the circle into an ellipse instead of
+// letting the text shrink.
 const SIZE_CLASSES: Record<
   NonNullable<ApplicationLogoProps["size"]>,
   string
 > = {
-  xs: "h-4 w-4",
-  sm: "h-6 w-6",
-  small: "h-8 w-8",
-  large: "h-10 w-10",
+  xs: "h-4 w-4 shrink-0",
+  sm: "h-6 w-6 shrink-0",
+  small: "h-8 w-8 shrink-0",
+  large: "h-10 w-10 shrink-0",
 };
 
 const FALLBACK_TEXT_CLASSES: Record<
--- a/services/vault/src/components/ApplicationLogo.tsx
+++ b/services/vault/src/components/ApplicationLogo.tsx
@@ -8,18 +8,20 @@ interface ApplicationLogoProps {
   shape?: "circle" | "rounded";
 }
 
-// `shrink-0` on every size: the logo sits beside truncating text in flex rows,
-// where without it flexbox squashes the circle into an ellipse instead of
-// letting the text shrink.
 const SIZE_CLASSES: Record<
   NonNullable<ApplicationLogoProps["size"]>,
   string
 > = {
-  xs: "h-4 w-4 shrink-0",
-  sm: "h-6 w-6 shrink-0",
-  small: "h-8 w-8 shrink-0",
-  large: "h-10 w-10 shrink-0",
+  xs: "h-4 w-4",
+  sm: "h-6 w-6",
+  small: "h-8 w-8",
+  large: "h-10 w-10",
 };
+
+// A layout invariant rather than a size, so it sits outside SIZE_CLASSES: the
+// logo sits beside truncating text in flex rows, where without it flexbox
+// squashes the circle into an ellipse instead of letting the text shrink.
+const LOGO_LAYOUT_CLASS = "shrink-0";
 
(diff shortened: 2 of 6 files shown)


services/vault/src/components/shared/ListRow.tsx
```diff
@@ -12,6 +12,50 @@ import type { ReactNode } from "react";
...
+  "min-w-[120px] shrink grow-[2] basis-[384px]";
+
+/**
+ * Trailing action cell. Rows whose action is conditional reserve it anyway, so
+ * a row without a button still lines its cells up with the rows that have one.
+ * The width matches `ROW_BUTTON_MIN_WIDTH_PX`, spelled literally because
+ * Tailwind only emits classes it can find as whole strings in the source.
+ */
+export const LIST_ROW_ACTION_SLOT_CLASS =
+  "flex min-w-[120px] shrink-0 justify-end";
```
gbarkhatov: **Column starts now vary with the CTA label.**

The slot is `min-w-[120px]` with auto width and `shrink-0`, while every other cell has `grow`. So the action's rendered width comes straight out of the slack the four data cells share: free space = W - 64 (gaps) - 816 (bases) - buttonWidth, split 2:1:1:1.

`Withdraw` / `View Details` sit at the 120px floor, but PendingRow's label comes from `COPY.pegin.primaryAction` - `Broadcast Pre-Pegin` is ~170px and `Submit WOTS Key` ~145px at `text-sm` + `px-4`. A 50px wider button drags that row's status column ~20px left, provider ~30px, hash ~40px, relative to the Active and Inactive rows.

Worth flagging that this is a step back rather than a case the PR doesn't cover yet: with `w-[180px] shrink-0` on cells 1-3 and `flex-1` only on the hash, the first four column starts were identical in every row no matter how wide the button was. Which also means the symptom in the description ("columns started at different places from one row to the next") can't have come from these three rows - the pooled gap, the heights, the ellipsis and the squashed logo all check out, that one doesn't.

Suggestion: give the slot a fixed basis sized to the widest row action (`shrink-0 grow-0 basis-[176px]` or similar) and route PendingRow's button (`VaultsLifecycleSections.tsx:241-267`) and ActiveVaultRow's (`VaultsActiveSection.tsx:123-135`) through it too. With the action column constant, every row's remaining slack is identical and the alignment guarantee comes back - this time without pinning the data cells.

services/vault/src/components/Activity/ActivityRowV3.tsx
```diff
@@ -20,9 +21,9 @@ import { formatActivityTime } from "@/utils/formatting";
 import { ActivityHashLink } from "./ActivityHashLink";
 import { STATUS_DOT } from "./statusDot";
 
-/** Fixed column width shared by every cell, so rows line up across cards
- *  regardless of amount length (Figma: 180px). */
-const COLUMN_CLASS = "flex w-[180px] shrink-0 items-center";
+/** Column shared by every cell, so rows line up across cards regardless of
+ *  amount length. */
+const COLUMN_CLASS = `flex items-center ${LIST_ROW_COLUMN_CLASS}`;
```
gbarkhatov, line 29: **Activity rows with an action no longer line up with rows without one.**

The five cells now grow, but the action at lines 151-153 is still rendered conditionally, `ml-auto ... shrink-0`, with no space reserved. Only refundable expired deposits get one, and liquidation-group children (`LiquidationGroupCardV3.tsx:40`) never do - so inside the same list a row with a Withdraw button hands ~136px (button + gap) to nothing, and its five cells each land ~30px narrower: status shifts ~30px left, type ~60px, hash ~90px, time ~110px, against the rows directly above and below it.

Under the old fixed 180px columns this was harmless - `ml-auto` floated the button right and the columns stayed pinned. It's also why the comment two lines above ("so rows line up across cards") no longer holds. Incidentally `ml-auto` is now inert: with every cell growing there is no free space left for an auto margin to absorb.

The reserved-slot fix applied to InactiveRow (`VaultsLifecycleSections.tsx:365-384`) is exactly right here - render the wrapper unconditionally and put `action` inside it, so actionless rows reserve the same width.

services/vault/src/components/shared/ListRow.tsx
```diff
@@ -23,7 +67,7 @@ export function ListRowCard({
   return (
     <div
       data-testid={testId}
-      className={`${CARD_SHELL_CLASS} flex flex-wrap items-center gap-x-4 gap-y-3 p-4`}
+      className={`${CARD_SHELL_CLASS} ${LIST_ROW_MIN_HEIGHT_CLASS} flex flex-wrap items-center gap-x-4 gap-y-3 p-4`}
```
gbarkhatov: **This min-height reaches the Activity and Loans lists too.**

`ListRowCard` is also the card for the Activity list (`ActivityList.tsx:148`) and Active Loans (`ActiveLoansList.tsx:52`). Both are single-line rows - a 32px avatar or a 36px button plus 34px of padding and border, ~70px - so every row on those two pages gets ~8px taller.

That may well be intended (three surfaces, one row height), but it isn't in the description or the testing notes, and the constant's own doc scopes it to "every lifecycle row", which reads as the vault sections only. Either widen the doc to say it deliberately sets the height for all three lists, or move the class to the vault call sites and leave Activity/Loans as they were.

services/vault/src/components/shared/ListRow.tsx
```diff
@@ -12,6 +12,50 @@ import type { ReactNode } from "react";
...
+/**
+ * The leading cell — a logo beside a two-line amount / sub-line block. Its
+ * basis fits the longest sub-line the rows produce (the refund-maturity
+ * notice, measured at 343px in the app's 12px face) plus the 32px logo and its
+ * 8px gap, so that copy reads in full rather than ellipsing. It takes a double
+ * share of any remaining slack, and gives width back proportionally when the
+ * viewport is too narrow for every cell to sit at its basis.
+ */
+export const LIST_ROW_LEADING_COLUMN_CLASS =
+  "min-w-[120px] shrink grow-[2] basis-[384px]";
```
gbarkhatov: Two notes on this docblock, which is otherwise doing real work.

1. "gives width back proportionally when the viewport is too narrow for every cell to sit at its basis" isn't what happens under `flex-wrap`. A wrap container breaks the line *before* an item that doesn't fit and never shrinks items to avoid the break, so shrink only engages on a line that still overflows with a single item on it - below ~384px here. The docblock right above (lines 20-25) states this correctly, so the two comments in this file contradict each other. `shrink` is also flex's default and never engages in practice, on either constant.

2. 343px is one measurement of a string that varies: `Your refund will be claimable in ~${blocks} Bitcoin ${block|blocks} (~${hours}h).` (`copy.ts:155-156`) - block count and hours are both interpolated and the noun switches on plural. The basis leaves the text 344px, so a longer digit combination ellipses again at the narrow end of the single-line range, which is the state this was measured to prevent. A pointer from that copy entry back to this constant would at least make the coupling visible.

Also worth being aware that the wrap point is emergent from the base sum rather than pinned - it rose from 904px to 1000px of content with this change, so the 1366 in the description drifts whenever any basis moves.

services/vault/src/components/ApplicationLogo.tsx
```diff
@@ -8,14 +8,17 @@ interface ApplicationLogoProps {
   shape?: "circle" | "rounded";
 }
 
+// `shrink-0` on every size: the logo sits beside truncating text in flex rows,
```
gbarkhatov: Good fix and the reasoning holds. It doesn't reach the Activity row though: that leading cell uses core-ui `Avatar` (`ActivityRowV3.tsx:120`), and `.bbn-avatar` is `inline-flex` with `size-8` and no `shrink-0` (`packages/babylon-core-ui/src/components/Avatar/Avatar.css`). The cell around it was `w-[180px] shrink-0` before and is now shrinkable down to `min-w-[120px]`, so the ellipse is newly reachable there for a long amount + symbol. `Avatar` takes `className`, so `className="shrink-0"` closes it.

Nit: `shrink-0` is a layout invariant rather than a size, so it repeats four times in a map keyed by size. `sizeClasses` is interpolated into both branches - appending it once at those two sites keeps the map about sizes.

gbarkhatov (dismissed):
Approving - the diagnosis is right, and the pooled-gap, leading-cell and logo fixes are real wins carried by unusually good rationale comments. Two things worth picking up before or right after merge:

- **The action column is still `min-width` + auto**, so a row's column starts move with its button label. Pending's `Broadcast Pre-Pegin` shifts that row's columns ~20-40px against Active/Inactive. Pinning the slot to a fixed width and using it for all four row types restores the guarantee the old fixed columns gave for free.
- **Activity kept its conditional action**, so under `grow` its rows now misalign against each other depending on whether a Withdraw button is present - the same reserved-slot fix applies there.

Rest is low severity: the shared min-height also retunes Activity and Loans row heights, the leading-column docblock describes a shrink behaviour `flex-wrap` doesn't perform, the 343px measurement has ~1px of headroom against a string whose digits vary, and Activity's core-ui `Avatar` still lacks `shrink-0`.

Nothing in CI locks any of this - no test asserts a layout class and there are no snapshots, so the 3042 passing tests are true but unrelated to what changed. That's consistent with the repo's test philosophy; it just means the measurements in the description are the only guard, which is worth keeping in mind for the two items above.

jonybur:
All six threads addressed in 3a6adae.

The core finding was right, and worse than estimated. I measured the three lifecycle rows in headless Chromium at 1280px with the real Px Grotesk face:

```
BEFORE   status 496,515,515   misaligned by 19px
         provider 697,725,725             28px
         hash   897,934,934               37px
         action 1097,1144,1144            47px
AFTER    all five column starts identical across all three rows
```

- **Fixed action slot.** `LIST_ROW_ACTION_SLOT_CLASS` is now `basis-[168px] grow-0 shrink-0`, and PendingRow, ActiveVaultRow and ActivityRowV3 all route through it with the wrapper rendered unconditionally. 168px is measured, not estimated: every row-action label rendered at `text-sm` + `tracking-[0.17px]` + `px-4`, and `Broadcast Pre-Pegin` is the widest at 167px. Everything else sits at the 120px floor.
- **min-height.** Moved off `ListRowCard` to the vault call sites via a `className` prop, so Activity and Active Loans stay at their ~70px rather than being silently pinned to 78px.
- **Avatar.** `className="shrink-0"` added. `shrink-0` also lifted out of `ApplicationLogo`s size map into a named layout constant.
- **Docblock.** The "gives width back proportionally" sentence is gone, the inert `shrink` dropped from both column constants, and `copy.ts`s `refundMaturing` now points back at the basis it sizes.

Two consequences worth flagging. The reserved slot raises the wrap point: lifecycle rows go from ~1000px to ~1048px of content, and Activity rows from ~784px to ~968px since they previously reserved nothing. And I left `ml-auto` on the shared row-button classes even though it is now inert inside the slot, because `ActiveLoansList` and `AssetSelectionPanel` still use those classes outside one.

jrwbabylonlab (approved):

PR #2204, opened 2026-08-06
feat(vault): borrow rate history and interest rate model charts

jeremy-babylonlabs:
Screenshot 2026-08-06 at 05 45 15
Screenshot 2026-08-06 at 05 45 08

## Summary

Adds the two remaining charts to the borrowing markets data page, built on the `LineChart` primitive from #2194.

- **Borrow APR history**: stepped line from the indexer's `/api/aave/reserves/:reserveId/history` endpoint. Six range buttons (opens on 1W). The header shows the min–max over the selected range. Hover reads date + rate. The line is stepped because the indexer buckets by last sample: a borrow rate is a level that holds until the next state change.
- **Interest rate model**: the live on-chain curve, sampled through the asset's own `irStrategy.calculateInterestRate` — 21 even utilization points plus the exact kink (from `getInterestRateData`) and the live totals. The utilization denominator stays invariant across the sweep. The y-axis runs to the strategy's real max rate. `Current` and `Optimal (Kink)` markers use collision-aware callouts; hover interpolates anywhere on the curve.

## Implementation notes

- No new env var. The history client derives its REST base from `NEXT_PUBLIC_TBV_GRAPHQL_ENDPOINT` — the same Hono app serves both APIs.
- `LineChart` now measures its y-label column and renders labels flush left, matching the frames. Other chart families keep the fluid gutter.
- A shared request-timeout helper in `utils/async.ts` now backs both the GraphQL and history clients. The timeout bounds the full request, including body streaming.
- The devnet token registry gains the post-redeploy USDC/USDT/WBTC addresses, so `/markets/usdc` resolves again.

## Verification

- Vault suite: 3161 tests green. core-ui suite green. `tsc`, lint, and build clean in both packages.
- Live devnet check: the sampled curve's current APR matches the Hub's `getAssetDrawnRate` exactly; all six history ranges serve; the indexer's latest bucket equals the live Hub rate.
- Rendering verified in the browser against the Figma frames (node 10088-60956), including hover, tap persistence, range switching, and the callout collision cases.

### DIFF
--- a/packages/babylon-core-ui/src/components/charts/chartLayout.ts
+++ b/packages/babylon-core-ui/src/components/charts/chartLayout.ts
@@ -7,7 +7,7 @@
 
 import { useMemo, useSyncExternalStore } from "react";
 import { useParentSize } from "@visx/responsive";
-import { getFontEpoch, subscribeFontEpoch } from "./textMeasure";
+import { AXIS_LETTER_SPACING_PX, chartFont, getFontEpoch, measureText, subscribeFontEpoch } from "./textMeasure";
 
 /** The plot's default aspect ratio (was `aspect-ratio: 1016 / 350`). */
 export const DEFAULT_PLOT_ASPECT_RATIO = 1016 / 350;
@@ -50,6 +50,27 @@ export const X_AXIS_MARGIN_TOP_PX = 6; // xaxis margin-top 0.375rem
 const LEGEND_PAD_Y_PX = 6; // legend-seg padding 0.375rem
 const LEGEND_MARGIN_BOTTOM_PX = 12; // top-legend margin-bottom 0.75rem
 
+/** Figma frames' `gap-[24px]` between the flush-left y-axis label column and
+ * the plot (see {@link measureYAxisGutter}). */
+export const Y_AXIS_LABEL_PLOT_GAP_PX = 24;
+
+/** The y-axis font size a given chart width implies. Shared by the fluid
+ * layout and the measured-gutter path so both agree on what size the tick
+ * labels actually render at. */
+function resolveAxisFontPx(chartWidth: number): number {
+  return resolveFluid(FONT_AXIS, chartWidth, rootFontPx());
+}
+
+/** Measures the flush-left y-axis label column: the widest tick label at this
+ * chart's axis font size, plus {@link Y_AXIS_LABEL_PLOT_GAP_PX}. No labels
+ * means no column — the plot spans the full chart width. */
+function measureYAxisGutter(chartWidth: number, labels: readonly string[]): number {
+  if (labels.length === 0) return 0;
+  const font = chartFont(resolveAxisFontPx(chartWidth));
+  const widest = Math.max(...labels.map((label) => measureText(label, font, AXIS_LETTER_SPACING_PX)));
+  return widest + Y_AXIS_LABEL_PLOT_GAP_PX;
+}
+
 /** Approximates a `line-height: normal` line box for Px Grotesk. */
 export const TEXT_LINE_HEIGHT = 1.2;
 
@@ -74,11 +95,14 @@ export function computeChartLayout(input: {
   hasXAxis: boolean;
   /** Plot width / plot height. Defaults to {@link DEFAULT_PLOT_ASPECT_RATIO}. */
   aspectRatio?: number;
+  /** Replaces the fluid gutter clamp when provided (see
+   * {@link measureYAxisGutter}); absent keeps the fluid clamp. */
+  gutterPx?: number;
 }): ChartLayout {
-  const { chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio = DEFAULT_PLOT_ASPECT_RATIO } = input;
+  const { chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio = DEFAULT_PLOT_ASPECT_RATIO, gutterPx } = input;
   const remPx = rootFontPx();
-  const gutter = resolveFluid(GUTTER, chartWidth, remPx);
-  const fontAxis = resolveFluid(FONT_AXIS, chartWidth, remPx);
+  const gutter = gutterPx ?? resolveFluid(GUTTER, chartWidth, remPx);
+  const fontAxis = resolveAxisFontPx(chartWidth);
   const fontLabel = resolveFluid(FONT_LABEL, chartWidth, remPx);
   const fontAmount = resolveFluid(FONT_AMOUNT, chartWidth, remPx);
 
@@ -115,6 +139,11 @@ export function useChartLayout(input: {
   hasTopLegend: boolean;
   hasXAxis: boolean;
   aspectRatio?: number;
+  /** y-axis tick label text. When provided (even as an empty array), the
+   * gutter is measured from these labels instead of the fluid clamp — pass
+   * `undefined` to keep the fluid gutter (the default for charts that don't
+   * render a flush-left label column). */
+  yAxisLabels?: readonly string[];
 }): {
   parentRef: (node: HTMLDivElement | null) => void;
   layout: ChartLayout;
@@ -126,18 +155,21 @@ export function useChartLayout(input: {
     initialSize: { width: FALLBACK_CHART_WIDTH_PX },
   });
   // Re-render when a webfont finishes loading so text measured against the
-  // fallback font is redone with the real metrics (see textMeasure.ts).
-  useSyncExternalStore(subscribeFontEpoch, getFontEpoch, getFontEpoch);
+  // fallback font is redone with the real metrics (see textMeasure.ts). The
+  // epoch is also a dependency of the layout memo below: measureYAxisGutter
+  // reads measureText, whose cache the font load invalidates outside React's
+  // dependency tracking.
+  const fontEpoch = useSyncExternalStore(subscribeFontEpoch, getFontEpoch, getFontEpoch);
   // `width` starts at the fallback (initialSize) and only becomes 0 when the
   // ResizeObserver reports a genuinely collapsed container — hidden tab,
   // zero-width flex child. Rendering the fallback there would paint a 1016px
   // chart across the siblings (the SVG overflows visibly), so collapse instead.
   const collapsed = width <= 0;
   const chartWidth = collapsed ? FALLBACK_CHART_WIDTH_PX : width;
-  const { axisSide, hasTopLegend, hasXAxis, aspectRatio } = input;
-  const layout = useMemo(
-    () => computeChartLayout({ chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio }),
-    [chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio],
-  );
+  const { axisSide, hasTopLegend, hasXAxis, aspectRatio, yAxisLabels } = input;
+  const layout = useMemo(() => {
+    const gutterPx = yAxisLabels === undefined ? undefined : measureYAxisGutter(chartWidth, yAxisLabels);
+    return computeChartLayout({ chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio, gutterPx });
+  }, [chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio, yAxisLabels, fontEpoch]);
   return { parentRef, layout, collapsed };
 }
--- a/services/vault/src/applications/aave/clients/aaveIrm.ts
+++ b/services/vault/src/applications/aave/clients/aaveIrm.ts
@@ -1,0 +1,262 @@
+/**
+ * Interest-rate-model (IRM) curve for one Hub asset's on-chain kinked
+ * strategy — the data source for the C2 borrow-rate chart.
+ *
+ * The curve is piecewise-linear in utilization with a single breakpoint (the
+ * kink), so sampling only needs: 21 evenly-spaced points (0%, 5%, …, 100%)
+ * plus the exact kink and exact live-utilization points. With the breakpoint
+ * itself an exact sample, linear interpolation between samples reproduces the
+ * true on-chain curve exactly — no off-chain reimplementation of the rate
+ * math, same stance as `getProjectedBorrowAprPercentsSafe`.
+ *
+ * Sampling exploits that `T = liquidity + drawn + swept` is invariant: for a
+ * synthetic sample at utilization `u`, `drawn' = T·u`, `liquidity' = T − drawn'`,
+ * `swept' = 0` folds the sweep into liquidity without changing `T`. The one
+ * exception is the live-utilization sample, which passes the untouched
+ * `(liquidity, drawn, swept)` so it is bit-identical to the Hub's own
+ * `getAssetDrawnRate`.
+ */
+
+import type { Abi, Address } from "viem";
+
+import { ethClient } from "../../../clients/eth-contract/client";
+
+import {
+  IR_STRATEGY_ABI,
+  rateRayToPercent,
+  readHubAssetTotalsSafe,
+} from "./aaveHub";
+
+/** Utilization sampling step in BPS: 21 even points from 0% to 100%. */
+const IRM_SAMPLE_STEP_BPS = 500;
+/** 100% expressed in basis points. */
+const BPS_SCALE = 10_000n;
+/** BPS per whole percentage point (1% = 100 BPS). */
+const BPS_PER_PERCENT = 100;
+
+const IRM_CURVE_DATA_ABI = [
+  {
+    type: "function",
+    name: "getInterestRateData",
+    stateMutability: "view",
+    inputs: [{ name: "assetId", type: "uint256" }],
+    outputs: [
+      {
+        name: "",
+        type: "tuple",
+        components: [
+          { name: "optimalUsageRatio", type: "uint256" },
+          { name: "baseDrawnRate", type: "uint256" },
+          { name: "rateGrowthBeforeOptimal", type: "uint256" },
+          { name: "rateGrowthAfterOptimal", type: "uint256" },
+        ],
+      },
+    ],
+  },
+] as const;
+
+/** Identifies one Hub asset to read the interest-rate-model curve for. */
+export interface InterestRateModelCurveRequest {
+  /** Hub contract address (from the reserve's `hub` field). */
+  hub: Address;
+  /** Asset identifier on that Hub (from the reserve's `assetId` field). */
+  assetId: number;
+}
+
+export interface IrmCurvePoint {
+  /** Utilization in percent, 0–100. */
+  utilizationPercent: number;
+  /** Borrow APR percent at that utilization, from the on-chain strategy. */
+  aprPercent: number;
+}
+
+export interface InterestRateModelCurveResult {
+  /** Sorted by utilization; null when any leg failed (no half-curve). */
+  curve: IrmCurvePoint[] | null;
+  /** Kink utilization percent from getInterestRateData, or null. */
+  kinkUtilizationPercent: number | null;
+  /** Live utilization percent (drawn / (liquidity+drawn+swept)), or null. */
+  currentUtilizationPercent: number | null;
+  /** APR at the live totals — bit-identical to the Hub's getAssetDrawnRate. */
+  currentAprPercent: number | null;
+  /**
+   * The y-axis ceiling, percent: base+growthBefore+growthAfter, raised to the
+   * highest sampled APR when the live deficit pushes the curve above it.
+   */
+  maxAprPercent: number | null;
+  error: Error | null;
+}
+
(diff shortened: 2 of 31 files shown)


packages/babylon-core-ui/src/components/charts/chartLayout.ts
```diff
@@ -134,10 +163,10 @@ export function useChartLayout(input: {
...
-  const { axisSide, hasTopLegend, hasXAxis, aspectRatio } = input;
-  const layout = useMemo(
-    () => computeChartLayout({ chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio }),
-    [chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio],
-  );
+  const { axisSide, hasTopLegend, hasXAxis, aspectRatio, yAxisLabels } = input;
+  const layout = useMemo(() => {
+    const gutterPx = yAxisLabels === undefined ? undefined : measureYAxisGutter(chartWidth, yAxisLabels);
+    return computeChartLayout({ chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio, gutterPx });
+  }, [chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio, yAxisLabels]);
```
gbarkhatov: **The measured gutter never re-measures after the webfont loads.**

`useChartLayout` subscribes to the font epoch at line 159 purely to force a re-render when Px Grotesk resolves, and `LineChart` deliberately keeps its marker measurement *outside* `useMemo` for exactly this reason — LineChart.tsx:104-105: *"Not memoised: the widths come from measureText, whose cache is invalidated by font loads outside React's dependency tracking."*

`measureYAxisGutter` is now inside this memo and the epoch isn't in the deps. On `loadingdone` the cache clears and the component re-renders, but none of `[chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio, yAxisLabels]` changed — so the gutter keeps its fallback-font width for the life of the component while the callouts and axis text re-measure against the real font. First paint normally precedes the webfont resolving, so this is the common path rather than an edge case.

Capture the epoch at line 159 (`const fontEpoch = useSyncExternalStore(...)`) and add it here:

```suggestion
  }, [chartWidth, axisSide, hasTopLegend, hasXAxis, aspectRatio, yAxisLabels, fontEpoch]);
```

services/vault/src/applications/aave/hooks/useBorrowRateHistory.ts
```diff
@@ -0,0 +1,65 @@
...
+}: {
+  reserveId: bigint | null;
+  range: HistoryRange;
+}): UseBorrowRateHistoryResult {
+  const reserveKey = reserveId?.toString() ?? null;
+
+  const { data, isLoading, error } = useQuery({
+    queryKey: [QUERY_KEY, reserveKey, range],
+    queryFn: () =>
+      fetchBorrowRateHistory({ reserveId: reserveId as bigint, range }),
```
gbarkhatov: React Query's `signal` is never forwarded, so `fetchBorrowRateHistory`'s `signal` parameter has no production caller — it exists for types and tests only. Two consequences beyond the unused parameter: `withRequestTimeout`'s caller-abort branch is unreachable for this client (its doc comment cites "React Query unmount" as the motivating case), and clicking 1D → 1W → 1M in quick succession leaves every superseded request running to completion, up to the 30s timeout.

```suggestion
    queryFn: ({ signal }) =>
      fetchBorrowRateHistory({ reserveId: reserveId as bigint, range, signal }),
```

Narrowing `reserveId` above the `useQuery` call would let the `as bigint` cast go too.

services/vault/src/components/pages/BorrowingMarketsData/borrowChartData.ts
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Y-axis ticks: `count` even whole-percent-rounded steps from 0 to `max`
+ * (D4 — the axis tops at the on-chain max, never a hardcoded figure).
+ */
+export function percentTicks(
+  maxAprPercent: number,
+  count: number,
+): ChartAxisTick[] {
+  return Array.from({ length: count }, (_, i) => {
+    const value = Math.round((maxAprPercent * i) / (count - 1));
```
gbarkhatov: Each step is rounded independently while the domain stays `[0, maxAprPercent]`, so the ticks and the scale can disagree. Two concrete cases:

- `maxAprPercent = 64.5` → `[0, 16, 32, 48, 65]`. The top tick sits above the domain, and since `.bbn-line-chart__svg` is `overflow: visible`, that gridline and its label render above the plot box rather than clipping.
- `maxAprPercent = 2` → `[0, 1, 1, 2, 2]`. Duplicate values mean duplicate React keys (`LineChart` keys ticks as ``key={`y-${tick.value}`}``, so `y-1` and `y-2` each render twice) plus two pairs of exactly overlapping labels and gridlines.

Rounding the ceiling first and deriving *both* the domain and the ticks from that value fixes both at once.

Separately: this divides by `count - 1` with no guard, while its sibling `rateTicks` explicitly handles `count <= 1`. Only called with `5` today, but the asymmetry is a trap for the next caller.

services/vault/src/applications/aave/clients/aaveIrm.ts
```diff
@@ -0,0 +1,303 @@
...
+    optimalUsageRatio: bigint;
+    baseDrawnRate: bigint;
+    rateGrowthBeforeOptimal: bigint;
+    rateGrowthAfterOptimal: bigint;
+  };
+
+  const kinkBps = optimalUsageRatio;
+  const maxAprPercent =
+    Number(baseDrawnRate + rateGrowthBeforeOptimal + rateGrowthAfterOptimal) /
+    BPS_PER_PERCENT;
```
gbarkhatov: `maxAprPercent` is derived purely from the three shape params, but every sampled leg is called with the reserve's live `deficit`. If the strategy contributes anything at all for deficit, the sampled curve can exceed this ceiling — and the consumer feeds it straight into `yDomain` on a chart whose SVG is `overflow: visible`, so the line spills above the plot box instead of clipping.

Cheap to make structurally impossible — clamp once `curve` is built:

```ts
maxAprPercent: Math.max(shapeMaxPercent, ...curve.map((p) => p.aprPercent)),
```

services/vault/src/applications/aave/clients/aaveIrm.ts
```diff
@@ -0,0 +1,303 @@
...
+ * partially-sampled strategy.
+ */
+export async function getInterestRateModelCurveSafe({
+  hub,
+  assetId,
+}: InterestRateModelCurveRequest): Promise<InterestRateModelCurveResult> {
+  const publicClient = ethClient.getPublicClient();
+  const assetIdArg = BigInt(assetId);
+
+  let totals;
```
gbarkhatov: Lines 112-188 are a near-verbatim copy of the totals block in `getProjectedBorrowAprPercentsSafe` (aaveHub.ts:323-388): same five-call multicall, same status check, same destructuring — comments included, down to *"getAssetOwed returns [drawn, premium]…"* and *"The strategy takes the deficit in asset units…"*. That's ~75 duplicated lines.

Worth extracting a `readHubAssetTotalsSafe` in aaveHub.ts that both call. They have to stay in lockstep anyway — the module doc's "same stance as `getProjectedBorrowAprPercentsSafe`" is precisely the claim that they read the same denominator, and right now nothing enforces it.

Unrelated, same file: the module doc at line 15 points at `context-sections-0-2.md`, which isn't in the repo. Either inline the D2 derivation or drop the pointer.

services/vault/src/applications/aave/clients/aaveIrm.ts
```diff
@@ -0,0 +1,303 @@
...
+  const maxAprPercent =
+    Number(baseDrawnRate + rateGrowthBeforeOptimal + rateGrowthAfterOptimal) /
+    BPS_PER_PERCENT;
+
+  // Sample set: 21 even points ∪ { kink, current }, deduplicated and sorted.
+  const sampleBpsSet = new Set<bigint>();
+  for (let bps = 0n; bps <= BPS_SCALE; bps += BigInt(IRM_SAMPLE_STEP_BPS)) {
+    sampleBpsSet.add(bps);
+  }
+  sampleBpsSet.add(kinkBps);
```
gbarkhatov, line 193: `kinkBps` enters the sample set unvalidated. If a strategy ever returned `optimalUsageRatio` on a different scale — RAY rather than BPS; the declared ABI is four bare `uint256`s, so nothing rules it out at decode time — `sampleDrawn` exceeds `totalLiquidity` and `sampleLiquidity` goes negative at line 256, surfacing as an opaque viem encoding throw from inside `multicall`.

A one-line `kinkBps > BPS_SCALE` guard returning `NULL_CURVE_RESULT` with a named error turns that into something triageable, in the same shape as the `totalLiquidity === 0n` check above.

services/vault/src/components/pages/BorrowingMarketsData/BorrowRateHistoryCard.tsx
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/** Range over the loaded series; a single figure (no dash) when it's flat. */
+function headerFigure(points: BorrowRateHistoryPoint[]): string {
+  const rates = points.map((p) => p.ratePercent);
+  const min = Math.min(...rates);
+  const max = Math.max(...rates);
+  return min === max
+    ? formatAprPercent(min)
+    : `${formatAprPercent(min)} - ${formatAprPercent(max)}`;
```
gbarkhatov: User-visible string composed in the component. Per CLAUDE.md, interpolating strings belong in `copy.ts` as a function:

```ts
rateRange: (min: string, max: string) => `${min} – ${max}`,
```

That also settles the dash — the PR description says "min–max", but this emits a spaced ASCII hyphen.

services/vault/src/components/pages/BorrowingMarketsData/BorrowRateHistoryCard.tsx
```diff
@@ -0,0 +1,208 @@
...
+            <span className="text-xs leading-[1.66] tracking-[0.4px] text-accent-secondary">
+              {COPY.marketData.charts.borrowAprLabel}
+            </span>
+            <Hint tooltip={COPY.loans.borrowAprTooltip} />
+          </div>
+          {value}
+        </div>
+        <div className="flex items-center gap-1">
+          {RANGE_OPTIONS.map((option) => (
+            <button
```
gbarkhatov, line 96: The active range is conveyed by background color alone. `aria-pressed={option === range}` on each button (or a `role="radiogroup"` wrapper) makes the selection available to screen readers for free.

services/vault/src/components/pages/BorrowingMarketsData/InterestRateModelCard.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+import { percentTicks } from "./borrowChartData";
+
+/** Design's accent orange for both IRM markers and the legend's utilization
+ *  dot (dark frame; a visual-polish pass may re-map this to a token). */
+const MARKER_ORANGE = "#ce6533";
+/** core-ui's LineChart series default / the design's success-green token
+ *  (both themes) — shared explicitly by the curve and the legend's green dot
+ *  so they can never drift apart. */
+const SERIES_GREEN = "#518665";
```
gbarkhatov: These are literal hex applied via inline `style` on the legend dots and passed as `color` / `marker.color`, so they're frozen across themes — and the comment above concedes the orange was picked off the dark frame. The vault ships both themes (`dark:` variants throughout the app), so on a light surface the marker rules and legend dots keep their dark-frame values.

Design tokens / CSS custom properties are the fix. If the visual-polish pass is a follow-up, worth making sure it lands before this page is reachable in light mode.

services/vault/src/components/pages/BorrowingMarketsData/InterestRateModelCard.tsx
```diff
@@ -0,0 +1,251 @@
...
+    () =>
+      curve?.map((point) => ({
+        x: point.utilizationPercent,
+        y: point.aprPercent,
+      })) ?? [],
+    [curve],
+  );
+
+  const yDomain = useMemo<[number, number]>(
+    () => [0, maxAprPercent ?? 0],
```
gbarkhatov: The null-guard early return below doesn't cover `0`: a strategy whose max rate is genuinely 0 yields a degenerate `[0, 0]` scale plus five identical "0%" ticks from `percentTicks`. Worth folding into the same guard that already treats a null max as "the read failed".

Also nearby — the `curve.find(...)!` at lines 139-141 is safe by construction today (`kinkBps` is always in the sample set, and both sides divide by the same constant), but the assertion is load-bearing and degrades to `undefined` rather than an error if that invariant ever moves. A thrown error would match this file's own no-silent-fallback stance.

services/vault/src/clients/indexer/aaveHistoryClient.ts
```diff
@@ -0,0 +1,126 @@
...
+export async function fetchBorrowRateHistory({
+  reserveId,
+  range,
+  signal,
+}: {
+  reserveId: bigint;
+  range: HistoryRange;
+  signal?: AbortSignal;
+}): Promise<BorrowRateHistoryPoint[]> {
+  const endpoint = `${indexerRestBaseUrl()}/api/aave/reserves/${reserveId.toString()}/history?range=${range}&resolution=auto`;
```
gbarkhatov: Two small things on this request:

- `resolution=auto` is an inline magic string in the URL. The repo's rule is to name it — `const HISTORY_RESOLUTION = "auto"`, alongside `HISTORY_REQUEST_TIMEOUT_MS` right above.
- The comment below explains that the non-ok throw was deliberately lifted *outside* `withRequestTimeout` so it wouldn't be reclassified — but `await response.json()` stayed inside, so a malformed body surfaces as `Borrow rate history request to indexer.test failed: Unexpected token…`, a parse error dressed as a network failure. Returning `await response.text()` and parsing outside closes that last gap.

services/vault/src/components/pages/BorrowingMarketsData/borrowChartData.ts
```diff
@@ -0,0 +1,104 @@
...
+    return { value, label: `${value}%` };
+  });
+}
+
+function pad2(value: number): string {
+  return String(value).padStart(2, "0");
+}
+
+/** Tooltip date: 1D range → "Jul 4, 14:00"; longer → "Jul 4, 2026". */
```
gbarkhatov: The time component only appears for `1d`, but the card opens on `1w`. If the indexer serves sub-daily buckets there — which `resolution=auto` suggests it might — adjacent points render an identical "Jul 4, 2026" and the tooltip stops distinguishing them. Intended, or should the time show for `1w` too?

services/vault/src/components/pages/BorrowingMarketsData/__tests__/BorrowRateHistoryCard.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    });
+
+    const { container } = render(
+      <BorrowRateHistoryCard reserveId={5n} symbol="vBTC" />,
+    );
+
+    // jsdom reports an all-zero client rect for the hit area, so `clientX`
+    // reads straight through as the plot-local x (see LineChart's own
+    // tests). The fallback chart width is 1016px with a 68px gutter, so the
+    // plot is 948px wide — hovering its right edge selects the SECOND
```
gbarkhatov: Stale after this PR's own gutter change: with the measured gutter and no canvas backend in the vault test setup (`measureText` returns 0), the gutter is 24px and the plot is 992px — not 68px / 948px.

The assertion still passes (948 is still past the midpoint), but this comment is what justifies the magic `948`, so it's what a future reader will trust. Same staleness the sibling core-ui test was updated for in this PR.

services/vault/src/copy.ts
```diff
@@ -1216,6 +1216,26 @@ export const COPY = {
       },
       empty: "No borrow markets available.",
     },
+    // Borrow APR history + interest rate model charts (Figma node
+    // 10088-61056). Title Case header labels per the Figma; "Utilisation" is
+    // respelled to the American form this file mandates.
+    charts: {
+      borrowAprLabel: "Borrow APR",
+      utilizationRateLabel: "Utilization rate",
+      legendUtilizationRate: "Utilization Rate",
```
gbarkhatov: `utilizationRateLabel` ("Utilization rate") and `legendUtilizationRate` ("Utilization Rate") both render on the interest-rate-model card — header caption and legend entry, a few hundred pixels apart. copy.ts's own style rules call for sentence case. If the Figma genuinely specifies Title Case for the legend, worth saying so in the comment above; otherwise it reads as a typo.

services/vault/src/services/token/tokenService.ts
```diff
@@ -93,6 +93,30 @@ const TOKEN_REGISTRY: Record<string, TokenMetadata> = {
     decimals: 6,
     icon: TOKEN_ICONS.USDC,
   },
+  // USDC - Vault Devnet (2026-08 redeploy)
```
gbarkhatov, line 88: Is the pre-redeploy devnet USDC entry above (lines 88-95, `0xc137E7…`) still live, or stale as of this redeploy? Happy either way — just want it to be a decision rather than something that quietly outlives the deployment it points at.

gbarkhatov (dismissed):
Nice piece of work. The T-invariant sampling with the kink as an exact sample is the right call — linear interpolation between samples reproduces the on-chain curve exactly, with no off-chain reimplementation of the rate math. The all-or-nothing fault isolation (never a half-curve), the field-by-field validation in `parseHistoryPayload`, and the reserve-scoped `placeholderData` + withhold-on-error in `useBorrowRateHistory` all match the repo's no-silent-fallback stance, and each has a test named for the behavior it locks in. Extracting `withRequestTimeout`, with the stalled-body regression test, is a real improvement on the inline version it replaces.

Approving — nothing below blocks a merge, though the first two are worth taking before this ships.

**Verified locally** at `2187c1c`: core-ui 26/26 green, vault 99/99 green across the 14 touched suites.

### The two I'd act on

- **The measured y-axis gutter never re-measures after the webfont loads.** `measureYAxisGutter` sits inside a `useMemo` with no font-epoch dep — exactly the trap `LineChart.tsx:104-105` documents for its own measurement. Since first paint normally precedes the font resolving, the gutter that ships is the fallback-font one.
- **`useBorrowRateHistory` never forwards React Query's `signal`**, so nothing cancels on unmount or a rapid range switch, and the client's `signal` parameter has no production caller.

The rest are inline: a tick/domain rounding mismatch that can push a gridline outside the plot or emit duplicate React keys, ~75 lines duplicated from `getProjectedBorrowAprPercentsSafe`, hardcoded hex that won't survive the light theme, a string built outside `copy.ts`, and a few defensive/copy nits.

### Performance / security

- Three sequential multicall rounds per reserve on a 60s interval, the third carrying ~22 calls. The waterfall is inherent (round 2 needs `irStrategy`, round 3 needs the kink) and the cost is fine for a single reserve page.
- Both charts poll independently at 60s; React Query's default `refetchIntervalInBackground: false` keeps background tabs quiet.
- The indexer payload is validated per field and throws rather than rendering an empty chart. Rates aren't range-checked (a negative rate would pass), but this is display-only and `fittedRateDomain` clamps the floor at 0.
- Deriving the REST base from the existing GraphQL endpoint instead of adding a second env var is the right trade — one fewer place for the two to drift.

No CRITICAL PATHS files touched.

jrwbabylonlab (dismissed):

jonybur (approved):
Reviewed the full diff at head 94999ee. No substantive findings, and this head resolves the actionable items from the earlier (now-dismissed) review round:

- The y-axis gutter memo in `useChartLayout` now depends on the font epoch, so the measured label column re-measures when the webfont resolves instead of shipping the fallback-font width.
- `useBorrowRateHistory` forwards React Query's `signal` through `skipToken`-narrowed `queryFn`, so superseded range switches and unmounts abort instead of running to the 30s timeout - and the client's `signal` parameter now has a production caller.
- `percentAxis` derives domain and ticks from one rounded-up ceiling, closing the gridline-outside-plot / duplicate-key rounding mismatch.
- The duplicated Hub totals multicall is extracted to `readHubAssetTotalsSafe`, so the projection and IRM curve provably start from the same denominator.
- Chart colors moved to theme tokens (`--chart-irm-*` in globals.css) and the min-max header string moved into `copy.ts`.

Also checked: `AXIS_LABEL_GAP_PX` keeps its LiquidationChart consumer after LineChart drops it (no dead code); the kink-marker equality lookup is safe because both sides derive from the identical `Number(bps) / BPS_PER_PERCENT` computation; the no-silent-fallback stance holds through the new indexer client (field-by-field validation, throw on malformed payloads, empty array only for a genuinely young reserve); the frozen critical-path files are untouched. CI is fully green including the visual suite.

PR #2212, opened 2026-08-10
fix(vault): stop blaming the vault provider for ack-timeout expiries

jonybur:
## Why

The expired-vault tooltip told every `ack_timeout` vault that "the vault provider did not acknowledge in time". That reason only says the acknowledgment window lapsed - it happens just as often because the depositor never finished their side (never signed the payouts, for example), so the sentence is frequently wrong. Raised by Filippos.

## What

`ack_timeout` vaults now show the heading and expiry time alone: "This BTC Vault has expired. Expired 58d ago." `proof_timeout` and `activation_timeout` keep their explanations, since those name an unambiguous missed step.

## How

Dropped the `ack_timeout` entry from `COPY.pegin.expiration.reasons` and widened the label map to `Partial<Record<ExpirationReason, string>>`. The message builder already skipped a missing reason, so no logic changed. Tests updated for the new tooltip text.

### DIFF
--- a/services/vault/src/copy.ts
+++ b/services/vault/src/copy.ts
@@ -206,12 +206,12 @@ export const COPY = {
       REFUND_HTLC: "Refund available",
     },
     expiration: {
-      // `ack_timeout` deliberately has no sentence here: the acknowledgment
-      // window can lapse because the depositor never finished their side (for
-      // example, never signed the payouts), not only because the vault
-      // provider was late. Naming a cause would often be wrong, so those
-      // vaults show the heading alone.
       reasons: {
+        // The acknowledgment window can lapse because the depositor never
+        // finished their side (for example, never signed the payouts), not
+        // only because the vault provider was late. Naming a cause would
+        // often be wrong, so these vaults show the heading alone.
+        ack_timeout: null,
         proof_timeout: "The inclusion proof was not submitted in time",
         activation_timeout: "The BTC Vault was not activated in time",
       },
--- a/services/vault/src/copy.ts
+++ b/services/vault/src/copy.ts
@@ -206,8 +206,12 @@ export const COPY = {
       REFUND_HTLC: "Refund available",
     },
     expiration: {
+      // `ack_timeout` deliberately has no sentence here: the acknowledgment
+      // window can lapse because the depositor never finished their side (for
+      // example, never signed the payouts), not only because the vault
+      // provider was late. Naming a cause would often be wrong, so those
+      // vaults show the heading alone.
       reasons: {
-        ack_timeout: "The vault provider did not acknowledge in time",
         proof_timeout: "The inclusion proof was not submitted in time",
         activation_timeout: "The BTC Vault was not activated in time",
       },
--- a/services/vault/src/models/__tests__/peginStateMachine.test.ts
+++ b/services/vault/src/models/__tests__/peginStateMachine.test.ts
@@ -566,7 +566,7 @@ describe("peginStateMachine", () => {
 
     it("surfaces a CSV-maturing countdown when refund timelock has not elapsed", () => {
       const state = getPeginState(ContractStatus.EXPIRED, {
-        expirationReason: "ack_timeout",
+        expirationReason: "proof_timeout",
         canRefund: false,
         refundMaturityState: "maturing",
         refundMaturesInBlocks: 24,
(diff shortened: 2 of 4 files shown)


jrwbabylonlab (dismissed):

services/vault/src/models/peginStateMachine.ts
```diff
@@ -206,7 +206,7 @@ const REFUND_BROADCAST_SUPPRESSION_MS = 6 * 60 * 60 * 1000;
 // Expiration helpers
 // ============================================================================
 
-const EXPIRATION_REASON_LABELS: Record<ExpirationReason, string> =
+const EXPIRATION_REASON_LABELS: Partial<Record<ExpirationReason, string>> =
   COPY.pegin.expiration.reasons;
```
gbarkhatov, line 210: `Partial<Record<ExpirationReason, string>>` drops the compile-time contract that every reason has a copy decision. `ExpirationReason` is owned upstream in `packages/babylon-ts-sdk/src/tbv/core/services/deposit/peginState.ts:36`; before this PR, adding a fourth variant there broke the build until someone wrote copy for it. Now it silently degrades to the bare heading — no type error, no failing test. The same union is mirrored by hand in `services/vault/src/services/vault/fetchVaults.ts:316-320`, so there is already one place that can drift.

Keeping the guard costs one line, and it puts the "we deliberately say nothing here" decision in front of whoever next edits `copy.ts` instead of leaving it only in a comment above the object:

```ts
// copy.ts
reasons: {
  // The acknowledgment window can lapse because the depositor never finished
  // their side (never signed the payouts, say), not only because the vault
  // provider was late — naming a cause would often be wrong.
  ack_timeout: null,
  proof_timeout: "The inclusion proof was not submitted in time",
  activation_timeout: "The BTC Vault was not activated in time",
},
```

```suggestion
const EXPIRATION_REASON_LABELS: Record<ExpirationReason, string | null> =
  COPY.pegin.expiration.reasons;
```

`buildExpiredMessage` already truthy-checks the label, so no logic changes and the tests stay as written.

services/vault/src/copy.ts
```diff
@@ -206,8 +206,12 @@ export const COPY = {
       REFUND_HTLC: "Refund available",
     },
     expiration: {
+      // `ack_timeout` deliberately has no sentence here: the acknowledgment
+      // window can lapse because the depositor never finished their side (for
+      // example, never signed the payouts), not only because the vault
+      // provider was late. Naming a cause would often be wrong, so those
+      // vaults show the heading alone.
       reasons: {
```
gbarkhatov, line 214: Non-blocking, but worth confirming with whoever raised this: `ack_timeout` is plausibly the most common expiry, and those depositors now get no explanation at all. For a vault with no `expiredAt` the tooltip is exactly "This BTC Vault has expired." — the same words as the `Expired` badge sitting next to it.

A blame-neutral sentence that is true either way would keep the information without the wrong attribution, e.g. `ack_timeout: "The acknowledgment window closed before the vault was ready"`. If "show nothing" really is the intended outcome, then please use the explicit `ack_timeout: null` form from the comment on `peginStateMachine.ts` rather than omitting the key.

services/vault/src/models/__tests__/peginStateMachine.test.ts
```diff
@@ -568,7 +576,7 @@ describe("peginStateMachine", () => {
       expect(state.refundMaturesInBlocks).toBe(24);
       // 24 blocks × 10 min = 240 min → ceil(240/60) = 4h.
       // The countdown lives only in `inlineSubtext`; `message` (tooltip)
-      // stays focused on the expired reason so the user doesn't see the
+      // stays focused on the expiry itself so the user doesn't see the
```
gbarkhatov, line 579: Small follow-on from dropping the sentence: this test and the sibling at line 601 still pass `expirationReason: "ack_timeout"`, which no longer affects anything they assert. `message` is now just the heading, so `expect(state.message).not.toContain("claimable in")` passes trivially rather than proving the countdown stayed out of a composed message.

Switching both to `proof_timeout` restores the original intent — the tooltip carries the reason, the countdown lives only in `inlineSubtext`. (Or drop the field entirely if you would rather these tests say nothing about expiry reasons.)

gbarkhatov (dismissed):
Approving — the change is correct and well scoped, and `buildExpiredMessage` already skipped a missing reason, so there genuinely is no logic change. Ran the affected suite locally (`vitest run src/models/__tests__/peginStateMachine.test.ts`): 106 pass. No critical path is touched; this is display copy only.

One thing I would like changed before merge — the `Partial<>` widening, inline on `peginStateMachine.ts`. Everything else is a nit.

Two stale-text items fall outside the diff, so I could not anchor them inline:

- `services/vault/src/models/peginStateMachine.ts:738` still reads "Tooltip stays focused on the expired reason + when". For `ack_timeout` there is no longer a reason in the tooltip. The PR updated the two mirroring comments in the test file but not this one.
- `services/vault/src/utils/__tests__/vaultTransformers.test.ts:107` is still named `"shows expired with ack_timeout reason"` while asserting only `displayLabel === "Expired"` — the same rename this PR already did in the sibling suite. (The two `as any` casts on `expirationReason` there are pre-existing, but they are exactly what would hide a future union change.)

gbarkhatov (approved):

jrwbabylonlab (approved):

PR #2213, opened 2026-08-11
feat(ts-sdk, vault): gate Activate on the peg-in activation delay

jrwbabylonlab:
## What

Holds the **Activate** button closed until a vault's on-chain activation window has actually opened, and shows the depositor how long is left instead of letting the click fail.

Ships behind `NEXT_PUBLIC_FF_ENABLE_ACTIVATION_DELAY`, off by default.

## Why

The Testnet v0.8 contracts introduce a delay between a vault being verified (its final ACK) and the moment it may be activated. The dApp knew nothing about it, so:

- Activate looks ready the instant a vault verifies, and the click fails with a raw contract revert the depositor cannot act on.
- That click also puts the HTLC secret into the transaction simulation — and so in front of the RPC provider — for a call the contract was always going to refuse.

The flag exists because the parameter is not deployed everywhere yet: devnet has it, testnet and staging do not. The gate fails closed, so it is turned on per environment only once `peginActivationDelay()` answers there.

## How

- The SDK gains a standalone `peginActivationDelay()` read, kept out of the shared protocol-params multicall so deployments without the parameter keep working.
- The dashboard reads the delay against the vault's verification block and, while the window is still shut, replaces the Activate button with a disabled one and a countdown ("Opens in 12 blocks (~3 min)").
- The same check runs again on fresh reads immediately before the secret is used, so a governance change between render and click cannot slip through.
- Anything unreadable — a missing getter, a failed block read — keeps the vault gated rather than letting the secret through.
- "Activate and redeem" is deliberately exempt, matching the contract: it mints nothing, so the delay does not apply, and gating it would block the documented way to recover BTC.

Test coverage: 41 new tests across the floor maths, the gate hook, the state machine, the action row, and the pre-submit re-check, plus god-mode scenarios for the waiting state.

> [!IMPORTANT]
> Touches the activation path that reveals the HTLC secret — two reviewers per `CLAUDE.md`.


jonybur (dismissed):

jonybur (approved):
Reviewed at head 008e68d8. The gate is where it needs to be and the secret-reveal path is protected: `handleActivation` re-reads pause / basic+protocol / block / `peginActivationDelay` in parallel, exact-matches the vault against the fresh VERIFIED set, re-runs the floor with `getActivationFloorReleaseInfo`, and only then calls `validateSecretAgainstHashlock(secretHex, protocolInfo.hashlock)` before `revealSecretOnChain`. The SDK re-checks the same hash before calldata assembly, so `hash(secret) === expectedHash` holds at two independent seams.

Fail-closed properties hold: `peginActivationDelay()` throwing or resolving `undefined`, `getBlockNumber` failing, or `verifiedAt === 0n` while the delay is nonzero all propagate through `onFloorReadFailure` and abort before the secret is used. The redeem exemption is correct - `_requireActivationDelayElapsed` on-chain guards `activateVaultWithSecret` only, and gating "activate and redeem" would block the documented recovery.

Feature-flag off is a true no-op: `isActivationDelayEnabled` returns false, `useActivationFloorGate` returns `EMPTY_GATE`, no contract reads issue, and `useVaultActions` skips the delay/block reads entirely - the state machine's `isActivationFloorGating(undefined)` is false so nothing downstream sees a gate.

Copy hygiene checks out: the new user-visible strings (`AWAITING_ACTIVATION_WINDOW`, `activationWindowOpening`/`Tooltip`/`Subtext`/`SubtextUnknown`/`NotOpen`/`Unavailable`) all live in `services/vault/src/copy.ts`, and `ActivationDelayNotElapsed` is keyed in `errorMessages.ts` next to the other ABI errors. The internal "activation floor inputs missing after a settled read" is passed only as `Error.cause` and never surfaced. No dead code, no silent fallbacks on the reveal path.

PR #2214, opened 2026-08-11
feat(wallet-connector): add the Ledger vault provider over the DMK

jrwbabylonlab:
New ledger_btc_vault wallet over the DMK (WebHID), shipped dark behind NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET.

- Connect + GET_APP_AND_VERSION preflight; BIP-86 key read as a raw APDU; taproot address derived locally
- DERIVE_CONTEXT_HASH + 3-phase APPROVE_VAULT_INTENT, TLV golden-vectored against Ledger's reference client
- Device-envelope gate rejects bad terms before any device I/O; provider mirrors the firmware state machine (IDLE → HASH_DERIVED → INTENT_LOADED) so state errors are actionable, not opaque status words
- No signing yet — signPsbt/signPsbts/signMessage throw typed capability errors; the SIGN_PSBT host protocol is #2219 (the published Bitcoin signer kit can't express the vault's no-policy flows)


packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -0,0 +1,360 @@
...
+  // a sequential loop. The input is named in the error so logs identify the caller.
+  signPsbt = async (psbtHex: string): Promise<string> => this.notWired(`signPsbt (${psbtHex.length / 2} bytes)`);
+
+  signPsbts = async (psbtsHexes: string[]): Promise<string[]> =>
+    this.notWired(`signPsbts (${psbtsHexes.length} psbt(s))`);
+
+  // TODO(#2221): BIP-322 PoP — a SIGN_PSBT with tx_version 0, not the base
+  // app's SIGN_MESSAGE.
+  signMessage = async (message: string, type: "bip322-simple" | "ecdsa"): Promise<string> =>
+    this.notWired(`signMessage (${type}, ${message.length} chars)`);
```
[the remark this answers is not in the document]
jrwbabylonlab: on-purpose. 
see the comment

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/envelope.ts
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Assert the terms lie inside the device's accepted intent domain.
+ *
+ * @throws {DepositTermsRejectedError} on the first violation
+ */
+export function assertDepositTermsDeviceCompatible(terms: DepositTerms): void {
+  // The firmware hardcodes the v2 PegIn shape, but a v1 intent LOADS fine and
+  // only fails at PSBT time — i.e. after the user has physically approved.
+  if (terms.vaultCoreVersion < DEVICE_MIN_VAULT_CORE_VERSION) {
```
gbarkhatov, line 42: This is a floor check, but `deviceCaps.ts:73-76` says the device **hardcodes** the v2 PegIn shape. Those are different statements: a future `vaultCoreVersion: 3` passes `< 2` and then reproduces exactly the failure this gate exists to prevent — the intent loads, and it only fails at PSBT time, after the depositor has physically approved on the device.

If the device hardcodes one shape, the gate should pin that shape (`!== DEVICE_MIN_VAULT_CORE_VERSION`, renamed accordingly) rather than bound it from below. A v3 rollout would then fail loudly at the envelope with an actionable message, which is the whole point of running this before any I/O.

Low priority — v3 does not exist — but the asymmetry between "hardcodes v2" in the constant's doc and `<` in the check is the kind of thing that reads as intentional later.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/deviceCaps.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Vault groups per intent (`vault_constants.h` VAULT_MAX_VAULTS). Multi-vault
+ * PegIn signing works in one session — `_validate_pegin` auto-detects the
+ * group from Input 0's PSBT_IN_OUTPUT_INDEX.
+ */
+export const DEVICE_MAX_VAULTS_PER_INTENT = 10;
+
+/**
+ * Fee-rate cap (sat/vB): the intent parser rejects > UINT32_MAX
+ * (`vault_tlv.c`). The `>= 1` floor next to it is the contract invariant,
+ * not a device check — the firmware parses rate 0 fine.
```
gbarkhatov, line 30: This claim is wrong at the pinned commit. `vault_tlv.c:73` @ `8f99b8b` is:

```c
if (rate == 0 || rate > UINT32_MAX) return VAULT_TLV_ERR_VALIDATION;
```

The firmware does reject rate 0. The `>= 1n` check in `envelope.ts:51` is correct either way, so nothing is broken — but the comment tells a future reader the floor is a contract invariant rather than a device check, which is an open invitation to move or delete it during a refactor. It is in fact both.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -0,0 +1,360 @@
...
+ * Ships behind `NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET` (default off).
+ * Covers connect, the key read, and the intent ceremony; signing needs the
+ * host-side SIGN_PSBT client (#2109) — the published Bitcoin signer kit
+ * cannot express the vault's no-policy flows.
+ */
+export class LedgerVaultProvider implements IBTCProvider {
+  private session: DmkSessionHandle | undefined;
+  private send: ApduSender | undefined;
+  /** See {@link DeviceIntentState}. Updated pessimistically around device I/O. */
+  private deviceState: DeviceIntentState = { phase: "idle" };
```
gbarkhatov, line 54: The memo's lifetime rests on one assumption that is not derived from firmware: that PegIn signing does not consume `INTENT_LOADED`. That is explicitly flagged as unconfirmed in two files (`dmkSession.ts:11-16`, `dmkApduSender.ts:18-23`, both #2110), and it is the single load-bearing input to the idempotence design.

Concrete path once #2219 lands: a same-session deposit hits the token cache at `ensureAuthenticatedVpClient.ts:54-57` and returns early **without** calling `deriveContextHash`, so `runDepositorPresignFlow`'s `approveDepositTerms` matches the memo and no-ops — against a device that may already have dropped to IDLE after the PegIn signature. The next terms-bound signature then dies with SW_BAD_STATE and the host has no idea why.

Relatedly, the `DepositTermsApprover` contract this PR rewrites (`depositTerms.ts:100-102`) now requires the memo be cleared on "a signing failure". `deriveContextHash` clears it; the signing half is vacuous today because signing throws, and there is nothing here recording the obligation.

Not blocking — no signing exists. Please leave a `TODO(#2219)` on this field so it is not rediscovered the hard way, and settle #2110 before that PR rather than after.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -0,0 +1,360 @@
...
+  groups: IntentVaultGroup[];
+  keeperPubkeys: Uint8Array[];
+  challengerPubkeys: Uint8Array[];
+}): string {
+  const hex = (b: Uint8Array) => Buffer.from(b).toString("hex");
+  return JSON.stringify(
+    {
+      s: { ...intent.scalars, prepeginTxidInternal: hex(intent.scalars.prepeginTxidInternal) },
+      g: intent.groups.map((g) => ({ ...g, vaultProviderPubkey: hex(g.vaultProviderPubkey) })),
+      k: intent.keeperPubkeys.map(hex),
```
gbarkhatov, line 334: The fingerprint is order-sensitive but the wire is not, so the memo can reject a re-approval that would have been a byte-for-byte no-op.

`encodeKeyBatches` sorts both rosters ascending before sending (`intentTlv.ts:195-203`). This fingerprints them in **caller** order. Same roster arriving in a different order — an SDK map/filter reordering between `preparePegin` and `runDepositorPresignFlow`, a roster rebuilt from a different query — produces identical APDUs but a different key, so the second `approveDepositTerms` takes the `intent-loaded` branch at line 267 and throws *"already holds a different approved intent … Restart the flow from derivation"*. The depositor is sent back through a ceremony that did not need re-running.

Cleanest fix is to fingerprint what actually goes on the wire rather than the input object: hash the concatenation of `encodeIntentScalars`, each `encodeIntentGroup`, and `encodeKeyBatches`. That makes the memo mean literally "byte-equal re-approval", which is what `DepositTermsApprover` documents, and it stops the fingerprint from drifting away from the encoder the next time one of them changes.

Worth a test: same terms, rosters reversed, second approval sends no APDUs. It fails today.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/envelope.ts
```diff
@@ -0,0 +1,159 @@
...
+      );
+    }
+    if (vault.depositorClaimValue < DEVICE_MIN_DEPOSITOR_CLAIM_VALUE_SATS) {
+      throw new DepositTermsRejectedError(
+        `${RANGE_MSG}: vault ${vault.htlcVout} depositorClaimValue ` +
+          `${vault.depositorClaimValue} below the ` +
+          `${DEVICE_MIN_DEPOSITOR_CLAIM_VALUE_SATS}-sat device floor`,
+      );
+    }
+    if (vault.peginAmount <= vault.commissionFee + crossFieldFloor) {
```
gbarkhatov, line 151: `peginMaxFee` never reaches this gate. It goes straight to `tlvU64` → `uintBE`, which throws a plain `Error("peginMaxFee … does not fit in 8 bytes")` for a negative or oversized value — not the `{ name: "DepositTermsRejectedError", reason: "device-envelope" }` shape the seam contract requires, so the SDK's friendly-error mapping silently drops it.

That is precisely the failure mode you already guard against for `htlcVout` at lines 92-98, with a comment saying so. Same reasoning applies here; the field just got missed. Same for the u64 ceiling on `peginAmount`/`commissionFee`/`depositorClaimValue` — unreachable with real sats, but the assertion is one line and the guard is about error shape, not arithmetic.

Note the firmware has no `pegin_max_fee` floor (`vault_tlv.c:252-256` parses it unchecked), so this is purely about not leaking a raw encoder error through the seam.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/dmkSession.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Build (or reuse) the DMK instance.
+ *
+ * Every `@ledgerhq/device-*` import is dynamic so a provider that ships
+ * disabled stays off the bundle's critical path — DMK pulls in xstate,
+ * inversify, reflect-metadata and purify-ts.
+ */
+async function getDmk(): Promise<DeviceManagementKit> {
+  if (dmkInstance) return dmkInstance;
```
gbarkhatov, line 45: Check-then-act across two awaits: concurrent callers both pass this guard, both `await` the dynamic imports, and both construct a `DeviceManagementKitBuilder` — registering the duplicate WebHID listener pair the comment above warns about, and leaking the loser.

Memoize the promise rather than the instance, which is the pattern `provider.ts:159-170` already uses for the pubkey read:

```ts
let dmkPromise: Promise<DeviceManagementKit> | undefined;
function getDmk() {
  dmkPromise ??= buildDmk();
  return dmkPromise;
}
```

`closeDmk()` then clears `dmkPromise` instead of `dmkInstance`.

`connectWallet` has the same shape — `isSessionAlive` is awaited before `this.session` is assigned, so two concurrent calls can both reach `connectDmkSession()`. Less likely to fire since it is user-gesture driven, but the visibility-check path that motivated the idempotence guard is exactly the one that could race it.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/derivation.ts
```diff
@@ -0,0 +1,69 @@
...
+import type { ApduSender } from "./vaultCommands";
+
+/** `GET_EXTENDED_PUBKEY` header (`commands.h`; P2 is the protocol version). */
+const CLA_BITCOIN_BASE = 0xe1;
+const INS_GET_EXTENDED_PUBKEY = 0x00;
+const P1_NONE = 0x00;
+const P2_PROTOCOL_VERSION = 0x01;
+
+/** Silent export — no confirmation screen for allowlisted paths. */
+const DISPLAY_OFF = 0x00;
```
gbarkhatov, line 22: Worth confirming this with Ledger explicitly rather than settling it in review.

The module JSDoc justifies host-side address derivation on the grounds that the firmware rebuilds every script from its own seed at signing time, so a lied-about address cannot receive a valid vault signature. That is true and it covers signing. It does not cover the other thing this address is used for: it becomes the depositor's registered payout destination, and a compromised host can substitute its own key at that point with nothing on the device screen to contradict it. On-device address confirmation is the mitigation that exists for exactly this, and `DISPLAY_OFF` opts out of it.

May well be the intended trade-off — the depositor does approve the full intent on-screen a moment later, and a silent read keeps `connect()` from requiring a button press. But the reasoning recorded here answers a narrower question than the one the flag raises, and this is the sort of thing that should be an explicit decision with Ledger's sign-off rather than an inherited default.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/derivation.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (!node.publicKey || node.publicKey.length !== COMPRESSED_PUBKEY_BYTES) {
+    throw new Error(`Ledger returned an extended key without a ${COMPRESSED_PUBKEY_BYTES}-byte public key`);
+  }
+  return Buffer.from(node.publicKey.subarray(1)).toString("hex");
+}
+
+/** `display(1) ‖ n(1) ‖ n×u32BE` — the GET_EXTENDED_PUBKEY payload. */
+function encodeGetExtendedPubkeyData(path: readonly number[]): Uint8Array {
+  const data = new Uint8Array(2 + path.length * 4);
+  data[0] = DISPLAY_OFF;
```
gbarkhatov, line 59: Nit: `path.length` is written into a single byte with no bound, so a >255-level path truncates mod 256 and sends a well-formed APDU describing the wrong path. Unreachable through `depositorPath` (always 5 levels), but every other encoder in this PR validates its inputs — `deriveContextHash` bounds the path at 1..10, `encodeIntentScalars` demands exactly 5 — and this one is the outlier.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -0,0 +1,360 @@
...
+    return this.send;
+  }
+
+  private notWired(method: string): never {
+    throw new WalletError({
+      code: ERROR_CODES.WALLET_METHOD_NOT_SUPPORTED,
+      message:
+        `${WALLET_PROVIDER_NAME} does not implement ${method} yet. This build ` +
+        `covers connect and the intent ceremony only; SIGN_PSBT for the ` +
+        `vault's no-policy flows is not implemented (#2109).`,
```
gbarkhatov, line 102: Nit: this points the user at #2109, but the `TODO` two lines below (line 296) says #2219, and the PR description confirms #2219 is the SIGN_PSBT host protocol. #2109 is the epic. Same swap at line 47 and in `dmkApduSender.ts:4`. Since this string is what a depositor actually sees when they hit the wall, it should name the issue that tracks the fix.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/dmkApduSender.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+    if (sw !== SW_OK) {
+      const known = STATUS_WORDS[sw];
+      // Name the app seen at connect ("BOLOS" = dashboard); the user may have
+      // switched apps since, hence the phrasing.
+      const appHint =
+        sw === SW_CLA_NOT_SUPPORTED && handle.appName
+          ? ` (app at connect time: "${handle.appName}"${handle.appVersion ? ` v${handle.appVersion}` : ""})`
+          : "";
+      throw new Error(
```
gbarkhatov, line 143: Refused and locked throw `WalletError`; every other non-OK status word throws a plain `Error`. So the two paths a user is most likely to hit outside the happy case — a device error, a wrong app — arrive at the app layer untyped, and anything downstream switching on `code` cannot classify them.

The import-cycle constraint noted at line 122 does not block this: the two branches above already construct `WalletError` without the `wallet` field, so the same trick works here. There is no obviously-right existing code — `UNKNOWN_ERROR` is the catch-all, or `codes.ts` gains one for device-reported failures. The status-word message text is good; it just deserves to survive the trip.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -0,0 +1,360 @@
...
+  /** Hardware wallets hold no inscription index. */
+  getInscriptions = async (): Promise<InscriptionIdentifier[]> => [];
+
+  /** A USB device has no account-switch event to subscribe to. */
+  on = (): void => {};
+  off = (): void => {};
+
+  getWalletProviderName = async (): Promise<string> => WALLET_PROVIDER_NAME;
+
+  getWalletProviderIcon = async (): Promise<string> => "";
```
gbarkhatov, line 319: Nit: returns `""` where every other BTC provider returns its logo import (`ledger-v2/provider.ts:306`, `unisat`, `okx`, `keystone`, …). Harmless today — the BTC connect UI reads `icon` off the metadata in `index.ts`, and only the ETH path re-resolves through this method — but `logo.svg` is already imported one file over, so returning it costs nothing and removes the question of whether the empty string was deliberate.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/envelope.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  it("rejects one vault group past the firmware maximum", () => {
+    const vault = makeTerms().vaults[0];
+    const eleven = Array.from({ length: 11 }, (_, i) => ({ ...vault, htlcVout: i }));
+    expect(() => assertDepositTermsDeviceCompatible(makeTerms({ vaults: eleven }))).toThrow(
+      /vault count 11 not in \[1, 10\]/,
+    );
+  });
+
+  it("accepts the payout timelock at both inclusive bounds", () => {
```
gbarkhatov, line 92: The name says "payout timelock at both inclusive bounds", but the values are `DEVICE_PAYOUT_TIMELOCK_MIN_BLOCKS` (91) and `DEVICE_PEGIN_CSV_TIMELOCK_MAX_BLOCKS` (1008) — the payout bounds are `[91, 4031]`. What this actually pins is the coupled band `[91, 1008]`, which is the intersection the next test names correctly.

The assertion is the right one — coupling is what production does — so just rename to match, e.g. "accepts the coupled timelock band at both ends". As written the file appears to test the payout range at both ends and does not, which is the sort of gap that survives review precisely because the name says otherwise.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/vaultCommands.ts
```diff
@@ -0,0 +1,210 @@
...
+import {
+  MAX_APDU_DATA_BYTES,
+  encodeIntentGroup,
+  encodeIntentScalars,
+  encodeKeyBatches,
+  type IntentScalars,
+  type IntentVaultGroup,
+} from "./intentTlv";
+
+const CLA_VAULT = 0xe1;
```
gbarkhatov, line 19: Nit: this and `derivation.ts:16` (`CLA_BITCOIN_BASE`) are both `0xe1`, and necessarily so — the vault app reuses the Bitcoin app's CLA, which is why `GET_EXTENDED_PUBKEY` works over the same session. Two independently-named constants with no cross-reference invite someone to "fix" one of them. A one-line comment on each pointing at the other is enough.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/runDepositorPresignFlow.ts
```diff
@@ -200,8 +200,18 @@ function assertDepositTermsMatchSigningContext(
...
   };
 
-  if (terms.protocolFeeRate !== context.protocolFeeRate) {
-    refuse("protocolFeeRate", terms.protocolFeeRate, context.protocolFeeRate);
+  // Every scalar the two types share: each one shapes the graph, and the
+  // timelocks are what payout.ts pins the input sequences to.
+  const scalars = [
+    ["protocolFeeRate", terms.protocolFeeRate, context.protocolFeeRate],
+    ["vaultCoreVersion", terms.vaultCoreVersion, context.vaultCoreVersion],
+    ["timelockPegin", terms.timelockPegin, context.timelockPegin],
```
gbarkhatov, line 208: Widening this from `protocolFeeRate` alone to all four shared scalars is the right direction — each one shapes the graph, and the timelocks are what `payout.ts` pins input sequences to, so a silent disagreement here is exactly the class of bug that ships a valid-looking signature over the wrong transaction.

One downstream consequence worth tracking rather than changing: `timelockPegin` is derived from the current params version while the context is version-locked to the vault, so a governance bump landing between broadcast and presign now hard-fails a PegIn that is already on-chain. Fail-closed is correct and this is the same class as the pre-existing `protocolFeeRate` check, so no change requested — but there is no recovery UX for a depositor who lands in that window, and this change widens the target. Worth an issue if one does not exist.

gbarkhatov (approved):
Approving — the adapter is dark-shipped, signing is absent, and the wire format holds up.

I cross-checked this against the firmware rather than reviewing it on its own terms, and it survives that: TLV tags, field order, u8/u32/u64 widths and the group record shape match `build_intent_tlv`/`build_group_tlv` in `tests/vault_client.py` @ `d74bf278` exactly, and `encodeKeyBatches` reproduces the reference's batching — keeper records then challenger records, concatenated and sliced at 7 **across** the role boundary, so a mixed batch is correct and the test that pins it is right. Every device cap traces to `8f99b8b`: 32/role (`vault_intent.h:8-9`), 10 vaults, dust 546, `[72,1008]`/`[72,4320]`, payout exclusive `(90,4032)` → `[91,4031]`, `prepegin_max_fee != 0` (`vault_tlv.c:163`), `vault_amount > commission_fee + 2*dust` (`vault_tlv.c:278-284`), strictly-ascending `htlc_vout` (`approve_vault_intent.c:142-144`).

Two decisions I want to call out as correct, because they are the ones that would have hurt: refusing tx-graph v1 **before** any device I/O (a v1 intent loads fine and only fails at PSBT time — after the depositor has physically approved), and pinning `P2_SHOW` so a silently-derived root can never load an intent. `statusWordOf` rejecting a short status word, so `[0x90]` cannot read as `0x9000`, is the right paranoia for this layer.

Inline comments below. Only two are worth acting on before merge (the fingerprint ordering and the `getDmk` race); the rest are for the SIGN_PSBT PR or are comment/naming drift.

**One process note, not a code finding:** roughly 1,400 of the 3,505 added lines are `docs/api/{clients,deposit-terms,services}.md` regenerating the `getVaultProviderBtcPubKey` → `getVaultProviderGenesisBtcPubKey` rename that already landed on `main` — the new name is in `main`'s source but absent from `main`'s docs, so this PR is absorbing someone else's drift. Roughly 40% of the diff has nothing to do with Ledger, on a change that touches critical path #3 and therefore needs two reviewers reading closely. Worth splitting into its own commit or PR so the adapter reviews clean.

Issue #2196, opened 2026-08-05
Generated ts-sdk API docs drift because nothing regenerates them


gbarkhatov:
## Summary

The generated TypeDoc API docs under `packages/babylon-ts-sdk/docs/api/` are tracked in git but are not regenerated as part of the PRs that change the source they document. They have drifted, and they currently document at least one symbol that no longer exists.

**This should be done after https://github.com/babylonlabs-io/babylon-toolkit/pull/2195 merges** — see [Why after #2195](#why-after-2195) for the sequencing reason.

## Current state

Docs were last regenerated in https://github.com/babylonlabs-io/babylon-toolkit/pull/2169. Everything merged since then has drifted.

Regenerating produces a diff across **5 files, roughly +1,554 / -237 lines**:

| File | Lines changed |
|---|---|
| `docs/api/clients.md` | 981 |
| `docs/api/services.md` | 609 |
| `docs/api/primitives.md` | 83 |
| `docs/api/utils.md` | 60 |
| `docs/api/managers.md` | 58 |

(Measured on the #2195 branch. Regenerating from main after that merges will produce a similar diff plus whatever else lands in the meantime.)

The overwhelming majority is mechanical churn — **310 of the changed lines in `clients.md` alone are just `Defined in:` line-number updates**, e.g. `PayoutManager.ts:178` → `PayoutManager.ts:187`, from source edits that moved code without changing the public API.

## The part that is actually wrong

Line-number drift is cosmetic. This is not:

`docs/api/clients.md` documents `AUTH_EXPIRED_DATA_KIND` as an exported const, at three places:

- `clients.md:1177` — "If the request fails with a wire-origin `auth_expired` error and a…"
- `clients.md:3645` — "additionally retries once on wire `auth_expired` (invalidate +…"
- `clients.md:5855` — `const AUTH_EXPIRED_DATA_KIND: "auth_expired" = "auth_expired";`

https://github.com/babylonlabs-io/babylon-toolkit/pull/2195 **deletes that export**. It was dead code — it gated the SDK's reactive re-auth path on an `error.data.kind` field that no server has ever emitted, which is the bug that PR fixes. Once it merges, the published docs describe a public symbol that does not exist, and describe a recovery mechanism that works in a way it no longer works.

That is worse than stale line numbers: someone reading the API docs would reasonably write code against `AUTH_EXPIRED_DATA_KIND`, or assume the client still keys off `data.kind`.

## Why after #2195

[...]

PR #2215, opened 2026-08-11
chore(vault): name the vault-provider key mocks for the getter they stub

gbarkhatov:
Follow-up to https://github.com/babylonlabs-io/babylon-toolkit/pull/2206. Test-local identifiers only — no production code, no assertions, no behaviour change.

## Why

@jonybur flagged one leftover while approving #2206: `vaultRefundService.test.ts` still named its mock `mockGetVaultProviderBtcPubKey` after the property it stubs became `getVaultProviderGenesisBtcPubKey`.

The rename sweep in #2206 matched `getVaultProviderBtcPubKey` with a lowercase `g`. These identifiers carry a capital `G` after the `mock` prefix, so a case-sensitive substring rename skipped them — and would skip any `mockXxx`-prefixed identifier the same way.

Chasing it turned up a second site that is **not** cosmetic and **predates #2206**.

## The two sites

**`vaultRefundService.test.ts`** — the variable was stale relative to the property it stubs. Cosmetic, exactly as reported. It now reads `mockGetVaultProviderGenesisBtcPubKey`, which also distinguishes it from `mockGetCurrentVaultProviderOperationBtcKey` in the same file — two different getters, two different keys, and the file now says which is which.

**`ensureAuthenticatedVpClient.test.ts`** — actively misleading. It read:

```ts
getCurrentVaultProviderOperationBtcKey: mockGetVaultProviderBtcPubKey,
```

A variable named for the *genesis* key stubbing the *current operation* key getter — a different value for any provider that has rotated. This is precisely the naming ambiguity that caused the bug #2173 fixed and that #2206 renamed the reader method to prevent, sitting in a test for the VP auth pin. A reader would reasonably conclude the pin resolves the registration key. It does not, deliberately.

Note the two renames are **not** the same: each takes the name of the getter it actually stubs. Renaming both toward "Genesis" would have preserved the defect.

Also added four lines above that mock recording why the auth pin is operation-key-only — accepting either key would hollow out the pin — with a pointer to `vpAuthPinnedPubkey.ts`. The rename makes the pairing correct; the comment stops someone "fixing" it toward the genesis key later.

## Verification

Both files pass with unchanged case counts — 33 + 3 = 36, rename-only. `tsc --noEmit -p tsconfig.lib.json` clean, `pnpm run lint` 0 errors.

```
git grep -in "getVaultProviderBtcPubKey" -- '*.ts' '*.tsx' | grep -vi GenesisBtcPubKey
```

No hits. The only remaining references to the old name are in the generated `docs/api/clients.md`, deliberately left alone — that is #2196, which needs the source-link churn fix first or the regeneration diff is four figures of noise.

Full suite not run: nothing outside two test files changed, and no exported or production surface is touched.

## Note for whoever hits this next

The vault typecheck fails on fresh `main` until `ts-sdk` is rebuilt — `activateVaultAndRedeem` and `PeginAction.ACTIVATE_AND_REDEEM` from #2159 exist in SDK source but not in a stale `dist/`. Not related to this change; `pnpm --filter @babylonlabs-io/ts-sdk run build` first.

### DIFF
--- a/services/vault/src/hooks/deposit/depositFlowSteps/__tests__/ensureAuthenticatedVpClient.test.ts
+++ b/services/vault/src/hooks/deposit/depositFlowSteps/__tests__/ensureAuthenticatedVpClient.test.ts
@@ -15,11 +15,16 @@ import { ensureAuthenticatedVpClient } from "../ensureAuthenticatedVpClient";
 const ON_CHAIN_PRE_PEGIN_HASH = "0xmatching_pre_pegin_hash";
 const ATTACKER_HASH = "0xattacker_chosen_hash";
 
-const mockGetVaultProviderBtcPubKey = vi.fn();
+// The auth pin resolves the VP's *current operation* key, deliberately — not
+// its genesis key, and not either-of-the-two. Accepting either would hollow out
+// the pin, which exists so a substituted server key cannot be used. See
+// `vpAuthPinnedPubkey.ts` for the reasoning before "correcting" this.
+const mockGetCurrentVaultProviderOperationBtcKey = vi.fn();
 const mockGetVaultProtocolInfo = vi.fn();
 vi.mock("@/clients/eth-contract/sdk-readers", () => ({
   getVaultRegistryReader: () => ({
-    getCurrentVaultProviderOperationBtcKey: mockGetVaultProviderBtcPubKey,
+    getCurrentVaultProviderOperationBtcKey:
+      mockGetCurrentVaultProviderOperationBtcKey,
     getVaultProtocolInfo: mockGetVaultProtocolInfo,
   }),
 }));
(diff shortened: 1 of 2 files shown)


jrwbabylonlab (approved):

jonybur (approved):
Test-only rename that aligns the mock identifiers with the getters they stub. The `ensureAuthenticatedVpClient.test.ts` change is the meaningful one — the previous `mockGetVaultProviderBtcPubKey` name wired into `getCurrentVaultProviderOperationBtcKey` was actively misleading given the genesis-vs-operation-key distinction the surrounding code cares about; the new name plus the pin-rationale comment closes that gap. `vaultRefundService.test.ts` also correctly picks the genesis-getter name to distinguish it from `mockGetCurrentVaultProviderOperationBtcKey` in the same file. No production or exported surface touched; assertion counts unchanged.

PR #2216, opened 2026-08-11
fix(vault): disable borrow and repay while the protocol scope is paused

jeremy-babylonlabs:


jrwbabylonlab (approved):

services/vault/src/components/shared/protocolStatus.ts
```diff
@@ -122,7 +128,7 @@ export function isActivateAndRedeemBlocked(gate: ProtocolGateState): boolean {
   return gate.protocol === "paused";
 }
 
-/** Repay is an aave-scope EXIT blocked ONLY by an aave pause (not protocol). */
+/** Repay is an EXIT blocked if EITHER scope is paused (preserved under freeze). */
 export function isRepayBlocked(gate: ProtocolGateState): boolean {
-  return gate.aave === "paused";
+  return gate.aave === "paused" || gate.protocol === "paused";
```
gbarkhatov, line 133: Worth spelling out what this does to the user it most affects. Under `{protocol: "paused", aave: null}`, the imminent-liquidation banner (`BannerActions.tsx:62-79`) now renders **both** of its CTAs disabled — "Add Collateral" via `depositBlocked` and "Repay Debt" via `repayBlocked` — while the pause banner tells that same user *"Debt continues accruing interest — monitor official announcements"* (`copy.ts:1674`). So a position at HF 1.02 is told its debt is growing and simultaneously denied every action that shrinks it, in a state where (per the comment above) the chain would have accepted the repay.

If that's the deliberate call, could the rationale live here rather than just "the UI is deliberately stricter than the chain there"? That line says *that* it's stricter, not *why* stricter is right for an exit — and the reasoning it replaced was explicit enough that someone will otherwise revert this in six months on the old argument. See the review body for the #1987 history and the two questions I'd want answered.

services/vault/src/components/shared/protocolStatus.ts
```diff
@@ -71,17 +71,23 @@ export function resolveBannerStatus(gate: ProtocolGateState): ScopeStatus {
 // Pure functions of the gate state (plus the standalone kill-switches). Each
 // implements one row of the governance matrix. ENTRY actions block on any
 // non-null status for their scope; EXIT actions block only on "paused" (Freeze
-// preserves exits). Scope accuracy matters: e.g. repay must stay available
-// under a protocol-only pause, so it checks the aave scope only.
+// preserves exits). Additionally, a protocol-scope PAUSE also blocks borrow
+// and repay, even though the adapter contract itself would still accept them —
+// the UI is deliberately stricter than the chain there. A protocol-scope
+// FREEZE keeps its narrow scope (deposit only).
```
gbarkhatov, line 77: This new paragraph contradicts one that's still in the file. `composeGateState`'s docblock (lines 48-53) argues the opposite policy in as many words:

> This gives the safe default for EXITS: a failed RPC read never blocks an exit on its own … **Over-blocking an exit traps users, so exits fail open**; entries can still be operator-blocked.

That's now only half true — repay is an exit that this PR deliberately over-blocks. The read-failure behavior it describes is still accurate, so it isn't wrong, but the *principle* it states is no longer the file's principle. A sentence there noting the protocol-pause exception would stop the file reading as self-contradictory to the next person deriving the rule from it.

services/vault/src/components/shared/protocolStatus.ts
```diff
@@ -71,17 +71,23 @@ export function resolveBannerStatus(gate: ProtocolGateState): ScopeStatus {
...
 export function isBorrowBlocked(gate: ProtocolGateState): boolean {
-  return featureFlags.isBorrowDisabled || gate.aave !== null;
+  return (
+    featureFlags.isBorrowDisabled ||
+    gate.aave !== null ||
+    gate.protocol === "paused"
+  );
 }
 
 /** Reorder is an aave-scope ENTRY action. */
```
gbarkhatov, line 93: The new rule isn't applied uniformly: reorder didn't get the protocol-pause clause. It's an aave-scope entry write — the same category as borrow, which did get it — so under `{protocol: "paused", aave: null}` a user can still submit `reorderVaults` while borrow and repay are both blocked.

If a protocol pause is a "UI full stop" (per the comment above), reorder is the one action that escapes it. Either add the clause here too, or note explicitly why reorder is exempt — right now the asymmetry looks unintentional rather than reasoned.

services/vault/src/components/shared/protocolStatus.ts
```diff
@@ -122,7 +128,7 @@ export function isActivateAndRedeemBlocked(gate: ProtocolGateState): boolean {
   return gate.protocol === "paused";
 }
 
-/** Repay is an aave-scope EXIT blocked ONLY by an aave pause (not protocol). */
+/** Repay is an EXIT blocked if EITHER scope is paused (preserved under freeze). */
```
gbarkhatov, line 131: Minor, and only worth doing if you're touching this anyway: `isWithdrawBlocked`, `isActivationBlocked`, and `isRepayBlocked` now have byte-identical bodies, and `|| gate.protocol === "paused"` is hand-rolled twice in this PR with no shared name.

The side effect is that the mis-wire class #1987 was worried about is now undetectable — a repay path that accidentally called `isWithdrawBlocked` would be behaviorally indistinguishable, and no test could catch it. A shared `isEitherScopePaused(gate)` that all three delegate to would keep the three named matrix rows (they document distinct governance decisions that could diverge again) while making the shared rule a single edit site.

services/vault/src/applications/aave/hooks/__tests__/useRepayTransaction.test.tsx
```diff
@@ -76,19 +76,18 @@ describe("useRepayTransaction — pause gating (aave-scope only)", () => {
...
+    let resolved: boolean | undefined;
     await act(async () => {
-      await result.current.executeRepay(100, RESERVE);
+      resolved = await result.current.executeRepay(100, RESERVE);
     });
 
-    expect(mockAssertReserve).toHaveBeenCalledTimes(1);
+    expect(resolved).toBe(false);
+    expect(mockAssertReserve).not.toHaveBeenCalled();
+    expect(mockRepayAll).not.toHaveBeenCalled();
```
gbarkhatov, line 90: Good that the repay chokepoint got a test for the new branch — but borrow gained the same branch and didn't. `useBorrowTransaction.test.tsx` only sets `{ protocol: null, aave: "paused" }` (line 52), so the new `gate.protocol === "paused"` short-circuit in `useBorrowTransaction.ts:74` has no hook-level coverage. Worth adding the mirror of this test — same shape, `executeBorrow` returning `false` before any on-chain read.

services/vault/src/copy.ts
```diff
@@ -1082,7 +1082,7 @@ export const COPY = {
       "Price data unavailable. Borrowing is temporarily disabled.",
     // Shown on the Repay tab when repay is blocked by a protocol pause (not a
     // technical/user error), so a user near liquidation knows it's governance,
-    // not a bug. Repay is gated only by an aave-scope pause.
+    // not a bug. Repay is gated by a pause on either scope.
```
gbarkhatov, line 1085: Borrow gained a blocking condition in this PR but no explanatory copy. The stated reason this repay string exists — *"so a user near liquidation knows it's governance, not a bug"* — now applies to borrow too, but a blocked borrow still falls back to `borrowingUnavailable` (line 1079): *"Borrowing is temporarily unavailable. Please check back later."* That reads like a transient technical fault, not a governance pause.

A pause-specific borrow string, selected the same way the repay tab selects this one, would keep the two tabs consistent about *why* the action is gone.

gbarkhatov (approved):
Approving on implementation — the predicates stay pure, both call sites already route through them (so the button `disabled` state and the hook chokepoint can't drift apart), and I ran the two touched test files locally: 21 pass. What follows is one question I'd like answered before this lands, plus comment/test debt this change creates in files outside the diff.

## The question: this deletes a guard that was added on review request in #1987

That reviewer wrote, on `useRepayTransaction.ts`:

> repay must stay available under a protocol-only pause so a near-liquidation user can de-risk … A future mis-wire to `isWithdrawBlocked` (blocks on either scope) would trap users into liquidation with every test still green.

A hook test was added in `ad7d3212` to lock exactly that, and this PR inverts it. The PR body is empty — no rationale, no link to a governance-matrix revision. See the inline note on `isRepayBlocked` for the user-facing shape of it.

Two things to confirm:

**(a) Does a `BTCVaultRegistry` pause actually halt Aave-side liquidations?** Nothing the frontend reads suggests it does — liquidation is an aave-scope action, and the aave scope is healthy in this state. If liquidations keep running, a protocol-scope governance pause becomes a liquidation event for every position near threshold, because the UI removes the only debt-reducing action while interest keeps accruing.

**(b) Was the Confluence matrix the file's docblock cites as source of truth updated to match?** The docblock still points at it as authoritative for the per-operation rows, and two rows just changed.

In fairness on scope: this only bites on a genuine on-chain registry-only pause. The operator kill-switch applies `operatorScopeStatus()` to *both* scopes in `composeGateState`, so it can never produce `{protocol: "paused", aave: null}`. If the intent is "ops wants one switch that stops everything," that's a legitimate product call — it just needs to be written down in the comment so the next reader doesn't flip it back on the old reasoning.

## Stale comments outside the diff (four; two are now flatly wrong)

CLAUDE.md's "after changes, check for comments that reference old behavior and update them" applies — these all describe the pre-PR rule:

- `services/vault/src/applications/aave/hooks/useRepayTransaction.ts:131` — *"Repay is an aave-scope EXIT: blocked only by an aave Pause (not a protocol pause, and never by Freeze)"* sits directly above the guard that now does the opposite.
- `services/vault/src/components/simple/PositionNotificationBanner/BannerActions.tsx:76` — *"Repay is blocked only by an aave Pause (preserved under Freeze)"*, on the near-liquidation CTA. Same inversion, and it's the comment that explains why that CTA is safe to disable.
- `services/vault/src/components/simple/PositionNotificationBanner/BannerActions.tsx:26` — the `repayBlocked` prop doc, *"An aave Pause blocks repay"*.
- `services/vault/src/applications/aave/hooks/useBorrowTransaction.ts:71` — *"Borrow is an aave-scope ENTRY action: Freeze or Pause blocks it"* — now incomplete, it omits the protocol pause.

## Testing caveat

This change can't be exercised in a dev build, which is worth knowing before it's called verified. The god-mode status override feeds the banner only — `ProtocolStatusBanner` does `status = statusOverride ?? resolveBannerStatus(gate)` and the override never reaches `useProtocolGateState` — so toggling "Fully paused" in the dev panel disables no CTA. And `NEXT_PUBLIC_FF_*` protocol-pause flag sets both scopes, so it can't isolate the protocol-only case either. Reproducing the state this PR changes needs a real registry pause on testnet.

jonybur (approved):
Reviewed at head `3870365e`. The predicate change is coherent (`isBorrowBlocked` and `isRepayBlocked` now check protocol-scope PAUSE), the docblock at `protocolStatus.ts:71` calls out the deliberate stricter-than-chain policy for these two rows, and both the unit-level (`protocolStatus.test.ts`) and hook-level (`useRepayTransaction.test.tsx`) tests were updated to lock in the new behavior. The `useRepayTransaction.ts` chokepoint delegates to `isRepayBlocked` so the disabled button and the broadcast guard can't drift.

gbarkhatov's inline threads on the banner-CTA design under `{protocol: paused, aave: null}`, the reorder-not-included inconsistency, the `composeGateState` docblock contradiction, the shared-name helper for the three now-identical exit predicates, and the four stale comments outside the diff all stand as substantive follow-ups to address before or after merge; nothing new to add there.

CI green.

PR #2218, opened 2026-08-11
feat(ci): show before and after screenshots in the pull request comment

jonybur:
## What
The visual regression check now puts the actual pictures in its comment. When a
pull request changes how something looks, the before and after sit side by side
right there in the thread.

## Why
The check already took the pictures and compared them, but the comment was a
list of file names and percentages. Seeing anything meant opening the run,
downloading a zip, unzipping it and opening a file in a browser. Nobody did, so
the check was reporting into the void.

## How
Each changed screen is cropped to the part that actually moved, and the before
and after are stitched into one image. A comment can only show images that live
at a public web address, so they go to a side branch that holds nothing else and
is cleaned up as pull requests close.

One repainted button can change dozens of screens, so the comment shows a capped
selection rather than all of them, with the component library and the app each
guaranteed a slot. The full list stays underneath, so nothing is hidden.

Publishing needs write access, so the job that runs this pull request's own code
and the job that holds the write token are kept separate - the first never sees
the token.

The rules deciding what gets pictured, and how it is cropped, now have tests.
They run first, before the check does anything slow.

Still advisory - it comments, it doesn't block merging.

### DIFF
--- a/.github/workflows/visual-regression.yml
+++ b/.github/workflows/visual-regression.yml
@@ -26,22 +26,43 @@ env:
   VISUAL_DIFF_BRANCH: visual-diffs
 
   # The surfaces that get captured, compared and published, as one list.
-  # Adding a third means editing this and .github/scripts/visual-capture.sh,
-  # and nothing else - the diff loop, the embed step and the publish
-  # allowlist all read it from here. In the publish job it doubles as a
-  # security allowlist, which is why that job takes it from the workflow
-  # file and never from the payload it downloads.
+  # The diff loop, the embed step and the publish allowlist all read it from
+  # here. In the publish job it doubles as a security allowlist, which is
+  # why that job takes it from the workflow file and never from the payload
+  # it downloads.
+  #
+  # Adding a third means editing this, .github/scripts/visual-capture.sh,
+  # and MAX_EMBEDDED_GROUPS in scripts/visual-embed.mjs. That third edit is
+  # the one that looks optional and is not: the per-surface floor is 2 and
+  # the cap on groups is 5, so three surfaces want 6 reserved slots out of
+  # 5. selectEmbeddedGroups resolves that by letting the last surface in
+  # rank order lose its reservation - silently, and that surface is the one
+  # just added. Raise the cap in the same change, or the guarantee this list
+  # is supposed to carry quietly stops holding.
   VISUAL_SURFACES: vault storybook
 
   # How much of the head sha names the published image directory. The sha is
-  # for humans reading the branch; uniqueness comes from the run id appended
-  # to it. Both are needed: GitHub proxies comment images through
-  # camo.githubusercontent.com and caches by URL, so a path that repeats
-  # serves the first render forever - and a sha alone repeats on a re-run,
-  # or on a reopen after the images were reaped, where what camo has cached
-  # may be a 404. Shared because the job that builds the URLs and the job
-  # that writes the files are different jobs, and a mismatch would 404 every
-  # image while both steps still exit 0.
+  # for humans reading the branch; uniqueness comes from the run id AND the
+  # run attempt appended to it. All three are needed: GitHub proxies comment
+  # images through camo.githubusercontent.com and caches by URL, so a path
+  # that repeats serves the first render forever - and a sha alone repeats
+  # on a re-run, or on a reopen after the images were reaped, where what
+  # camo has cached may be a 404.
+  #
+  # The attempt is not decoration. `github.run_id` is documented as not
+  # changing when a run is re-run - only `run_attempt` increments - so the
+  # id alone leaves the path byte-identical across attempts. That collides
+  # exactly where it hurts: the baseline is re-resolved per run rather than
+  # pinned, so people re-run this precisely BECAUSE main moved, attempt 2
+  # legitimately composes different pictures, and without the attempt it
+  # writes them to the URL camo already cached from attempt 1. The reviewer
+  # then reads attempt 2's percentages beside attempt 1's images, with every
+  # step still exiting 0.
+  #
+  # Shared because the job that builds the URLs and the job that writes the
+  # files are different jobs, and a mismatch would 404 every image while
+  # both steps still exit 0. The sha, the run id and the attempt all have to
+  # stay in lockstep across those two sites.
   VISUAL_SHA_PATH_LENGTH: 12
 
 # Advisory to begin with: this reports and comments, it does not fail the
@@ -115,6 +136,15 @@ jobs:
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile --prefer-offline
+
+      # Before the browser, because it costs seconds and gates the rules the
+      # rest of this job depends on. The selection and cropping rules are
+      # invisible when they regress - a wrong crop still produces a picture,
+      # just not one of the change - so a green report is not evidence they
+      # held. Deliberately not `continue-on-error`: everything downstream is
+      # advisory, and this is the one place the job should refuse to go on.
+      - name: Test the report scripts
+        run: pnpm run visual:test
 
       - name: Install Playwright browser
         run: pnpm --filter @services/vault exec playwright install --with-deps chromium
@@ -276,7 +306,7 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
-          RUN_DIR="${REPORT_SHA:0:${VISUAL_SHA_PATH_LENGTH}}-${GITHUB_RUN_ID}"
+          RUN_DIR="${REPORT_SHA:0:${VISUAL_SHA_PATH_LENGTH}}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
           node scripts/visual-embed.mjs \
             --report "${VISUAL_DIR}/report" \
             --surfaces "${VISUAL_SURFACES// /,}" \
@@ -353,9 +383,24 @@ jobs:
 
   # The only job holding a token that can write anything, and deliberately
   # the only one with no `actions/checkout`: the pull request's source and
-  # its dependency tree never land here, and this job runs none of it. What
-  # does cross is inert - the PNG bytes and the Markdown body out of the
-  # artifact - and both are treated as untrusted below.
+  # its dependency tree never land here, and this job runs none of it.
+  #
+  # What crosses the boundary is an artifact the capture job wrote, and the
+  # honest description of how it is handled is narrower than "untrusted".
+  # Enforced: PNG *filenames* are matched against SAFE_NAME and dropped if
+  # they miss, the total published size is capped, and the comment body is
+  # length-checked. Not enforced: the PNG *contents* are never parsed, and
+  # the Markdown body is `cat` into the comment as written. So code running
+  # in the capture job cannot post as itself or push anywhere, but it can
+  # hand this job arbitrary Markdown to post under github-actions[bot] and
+  # arbitrary bytes named *.png to commit to a public branch. Smaller than a
+  # write token, which is what the split is for - not an airlock.
+  #
+  # Closing that gap means making the crossing payload data rather than
+  # presentation: visual-embed.mjs emitting a manifest of typed fields
+  # (screen, status, ratio, coverage) and renderBody moving to this side of
+  # the boundary. Worth doing before this check becomes blocking; out of
+  # scope for the change that introduced the images.
   #
   # Be precise about what that buys. It is NOT protection against a
   # malicious PR author - on `pull_request` the workflow file itself comes
@@ -428,7 +473,14 @@ jobs:
           # contribution, not the branch: what bounds the branch is the reap
           # of closed pull requests below, so the live tree stays
           # proportional to how many are open at once.
-          MAX_PUBLISHED_BYTES=$(( 4 * 1024 * 1024 ))
+          #
+          # Sized against what this actually produces, not against what the
+          # repo could survive: a run publishes ~10 composites, and the eight
+          # on the branch today total 77KB. A ceiling in the megabytes only
+          # trips once the branch already outweighs the repo everyone clones,
+          # which is far too late to be a runaway detector. 512KB is ~6x
+          # observed and catches a misconfigured cap on the first run.
+          MAX_PUBLISHED_BYTES=$(( 512 * 1024 ))
           # The publish path is not serialised across pull requests -
           # `concurrency` above is keyed on github.ref, which only serialises
           # a PR against itself - so two PRs finishing together will collide.
@@ -446,7 +498,7 @@ jobs:
           # as soon as any one of its lines matched.
           SAFE_NAME='^[A-Za-z0-9][A-Za-z0-9._-]*\.png$'
 
-          RUN_DIR="${REPORT_SHA:0:${VISUAL_SHA_PATH_LENGTH}}-${GITHUB_RUN_ID}"
+          RUN_DIR="${REPORT_SHA:0:${VISUAL_SHA_PATH_LENGTH}}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
           PR_DIR="pr${PR_NUMBER}"
           REMOTE="https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
 
@@ -542,7 +594,8 @@ jobs:
               '- Shares no history with `main`. Never merge it anywhere.' \
               '- Rewritten as one parentless commit per publish, so the branch has no' \
               '  history to grow and a fresh clone transfers only the live tree.' \
-              '- `pr<number>/<head-sha>/` is deleted once that pull request closes, so' \
+              '- Laid out as `pr<number>/<short-sha>-<run-id>-<run-attempt>/`. The whole' \
+              '  `pr<number>/` directory is deleted once that pull request closes, so' \
               '  images in comments on closed pull requests will 404.' \
               '- Safe to delete outright. The next run recreates it.' \
               > "${WORK_DIR}/README.md"
@@ -598,6 +651,10 @@ jobs:
           PR_NUMBER: ${{ github.event.pull_request.number }}
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
           PAYLOAD_DIR: ${{ runner.temp }}/payload
+          # The commit each branch of the body below names. Same output the
+          # published path is keyed on, so the sha in the text and the sha in
+          # the image URLs cannot disagree.
+          REPORT_SHA: ${{ needs.visual.outputs.report_sha }}
         shell: bash
         run: |
           set -euo pipefail
@@ -607,8 +664,27 @@ jobs:
           # it ever comes back oversized this must fall back to the text
           # listing, not fail and leave the pull request with no report.
           MAX_COMMENT_BYTES=60000
+          # The text listing is the fallback, so it cannot be the thing that
+          # busts the limit. It is one line per changed screen and unbounded
+          # by anything this job controls - the cap upstream is on pictures,
+          # not on rows - and it is embedded in BOTH paths: inside the
+          # <details> of comment-images.md and directly below. Once the text
+          # alone clears the ceiling, the check above rejects the image body
+          # and the fallback exceeds it too, gh returns 422, `set -e` fails
+          # the step and the pull request gets no report at all. Not
+          # reachable at today's screen count; bounded so the sentence above
+          # is true unconditionally rather than by arithmetic that holds
+          # until someone adds screens.
+          MAX_DIFF_TEXT_BYTES=40000
           MARKER="<!-- visual-regression-report -->"
           BODY_FILE="${RUNNER_TEMP}/comment.md"
+          # Names the commit the report describes, on every branch below.
+          # A comment is rewritten in place, so without the sha a reader has
+          # no way to tell a current report from one left behind by a run
+          # that failed to update it - and these branches, the ones that
+          # fire when something went wrong, are the most likely to be that
+          # stale one.
+          AT_SHA="for \`${REPORT_SHA:0:${VISUAL_SHA_PATH_LENGTH}}\`"
           {
             echo "$MARKER"
             # `${CHANGED:-0}` substitutes on empty as well as unset, so a
@@ -619,7 +695,7 @@ jobs:
             if [ "${DIFF_OUTCOME}" != "success" ] || [ -z "${COMPARED}" ] || [ "${COMPARED}" = "0" ]; then
               echo "### Visual regression"
               echo ""
-              echo ":warning: The capture or comparison did not complete - no visual comparison was made."
+              echo ":warning: The capture or comparison did not complete ${AT_SHA} - no visual comparison was made."
               echo ""
               echo "[Open the run](${RUN_URL}) to see which step failed."
             # Per-surface gate: a surface that captured empty on both sides
@@ -627,17 +703,17 @@ jobs:
             elif [ -n "${MISSING}" ]; then
               echo "### Visual regression"
               echo ""
-              echo ":warning: No comparisons were produced for: ${MISSING}. The remaining surface(s) alone do not make this a clean run."
+              echo ":warning: No comparisons were produced ${AT_SHA} for: ${MISSING}. The remaining surface(s) alone do not make this a clean run."
               echo ""
               echo '```'
-              cat "${PAYLOAD_DIR}/diff-output.txt" || true
+              head -c "${MAX_DIFF_TEXT_BYTES}" "${PAYLOAD_DIR}/diff-output.txt" || true
               echo '```'
               echo ""
               echo "[Open the run](${RUN_URL}) to see why that capture came up empty."
             elif [ "${CHANGED}" = "0" ]; then
               echo "### Visual regression"
               echo ""
-              echo ":white_check_mark: No visual changes against the merge-base."
+              echo ":white_check_mark: No visual changes ${AT_SHA} against the merge-base."
             # The body with the pictures in it, used only once the images it
             # points at are actually being served. Composing or publishing
             # them is allowed to fail; posting URLs that 404 is not.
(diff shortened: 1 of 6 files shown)


.github/workflows/visual-regression.yml
```diff
@@ -316,15 +660,39 @@ jobs:
...
+          # back empty and the job posts a SECOND report instead of editing
+          # the first.
+          #
+          # The first line is taken with a parameter expansion rather than
+          # `| head -n1`, which would be a trap here: --jq runs per page, so
+          # head exits after page one, gh takes SIGPIPE writing page two,
+          # and `set -o pipefail` turns that into a failed step - losing the
+          # comment on precisely the busy pull requests --paginate is for.
+          MATCHES="$(gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" --paginate \
             --jq "map(select(.body | startswith(\"$MARKER\"))) | .[0].id // empty")"
```
gbarkhatov: **Two things in this one `--jq`.**

**The reaping loop at :693 can never fire for the case it was written for.** `--paginate` without `--slurp` applies the filter *per page* — the comment three lines up says exactly that — and the filter ends in `.[0].id`, so this emits **at most one id per page**. Duplicates come from consecutive runs of the old un-paginated lookup, so they sit adjacent, on the same page. That collapses to a single id, `MATCHES == EXISTING`, and the guard at :693 is false. Duplicates only get reaped when they happen to straddle a page boundary.

**The DELETE at :695 matches on the marker alone.** No author filter, so any comment whose body *starts with* the public marker is a target: the first match gets PATCHed with this run's report, the rest get deleted. Narrow — you'd have to paste the raw source including the HTML comment, not the rendered text — but the outcome is an irreversible DELETE of someone else's comment by CI, and the fix rides along in the same expression.

Both close with:

```
--jq ".[] | select((.body | startswith(\"$MARKER\")) and .user.login == \"github-actions[bot]\") | .id"
```

`.[]` emits every match per page; `EXISTING="${MATCHES%%$'\n'*}"` and `${MATCHES#*$'\n'}` below already handle multi-line and need no change. Note the inner parens — `select(.body | startswith(...) and ...)` would pipe `.body` into the whole conjunction and evaluate `.user` against a string.

.github/workflows/visual-regression.yml
```diff
@@ -8,9 +8,41 @@ concurrency:
...
+  # allowlist all read it from here. In the publish job it doubles as a
+  # security allowlist, which is why that job takes it from the workflow
+  # file and never from the payload it downloads.
+  VISUAL_SURFACES: vault storybook
+
+  # How much of the head sha names the published image directory. The sha is
+  # for humans reading the branch; uniqueness comes from the run id appended
+  # to it. Both are needed: GitHub proxies comment images through
+  # camo.githubusercontent.com and caches by URL, so a path that repeats
+  # serves the first render forever - and a sha alone repeats on a re-run,
```
gbarkhatov: **`run_id` does not change on a re-run, so this doesn't defend against the case it names.**

The comment justifies appending the run id because "a sha alone repeats on a re-run". But [`github.run_id`](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context) is documented as *"This number does not change if you re-run the workflow run"* — only `run_attempt` increments. So `RUN_DIR` is byte-identical across attempts of the same run.

That matters here specifically because the baseline is not pinned: it is `git merge-base origin/${base_ref} HEAD` resolved at capture time. People re-run this job precisely *because* main moved, so attempt 2 legitimately produces different composites — and writes them to the same path attempt 1 already published and camo already cached. By the model asserted in this comment, the reviewer then reads updated percentages next to attempt 1's pictures, with both steps exiting 0.

One token in each of the two places `RUN_DIR` is built (:279 and :449):

```
RUN_DIR="${REPORT_SHA:0:${VISUAL_SHA_PATH_LENGTH}}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
```

The two sites have to stay in lockstep, which is what the last sentence of this comment is already about — worth extending it to say the attempt is part of the key.

.github/workflows/visual-regression.yml
```diff
@@ -242,72 +338,320 @@ jobs:
...
-      # action, per the repo's supply-chain policy.
-      # Skipped on fork PRs: `pull_request` hands them a read-only token
-      # regardless of the `permissions:` block above, so the POST would 403
-      # and fail an advisory check for every external contributor. The job
-      # summary carries the same content for those runs.
+  # The only job holding a token that can write anything, and deliberately
+  # the only one with no `actions/checkout`: the pull request's source and
+  # its dependency tree never land here, and this job runs none of it. What
+  # does cross is inert - the PNG bytes and the Markdown body out of the
+  # artifact - and both are treated as untrusted below.
```
gbarkhatov: **"both are treated as untrusted below" is stronger than what the code does.**

Only the PNG *filenames* are constrained (`SAFE_NAME` at :447). The Markdown body is length-checked and otherwise `cat`'d verbatim into the comment (:647), and the PNG *contents* are never looked at.

So under the threat this split is explicitly built for — a compromised transitive dep inside the capture job — that code cannot post a comment itself, but it can hand this job arbitrary Markdown to post as `github-actions[bot]` on every PR, and arbitrary ≤4MB blobs named `*.png` to commit to a public org branch. That is a smaller capability than a write token, so the split still earns its keep; it just isn't the airlock the sentence claims.

The structural fix is to make the crossing payload data rather than presentation: have `visual-embed.mjs` emit a JSON manifest (screen name, status, ratio, coverage) and move `renderBody` into this job, so the body is assembled from typed fields this side of the boundary. That also picks up the escaping that `visual-diff.mjs` already does with `escapeHtml` at :157 and `visual-embed.mjs` does not — `group.group` and `groupHeadline` are interpolated raw into `<code>…</code>`, which is inert today only because Storybook ids are `[a-z0-9-]`.

If that is more than this PR wants to take on, the cheap version is to soften the sentence to say what is actually enforced.

.github/workflows/visual-regression.yml
```diff
@@ -242,72 +338,320 @@ jobs:
...
+          PAYLOAD_DIR: ${{ runner.temp }}/payload
         shell: bash
         run: |
           set -euo pipefail
+          # GitHub rejects a comment body over 65536 characters with a 422.
+          # The body is built from a capped set of screens so it lands near
+          # 8KB, but it is assembled by a job that ran the PR's own code - if
+          # it ever comes back oversized this must fall back to the text
+          # listing, not fail and leave the pull request with no report.
+          MAX_COMMENT_BYTES=60000
```
gbarkhatov, line 666: Worth noting the fallback this protects isn't actually a safe harbour: the same unbounded `diff-output.txt` is embedded in *both* paths — inside the `<details>` in `comment-images.md`, and `cat`'d directly at :654. Once that text alone clears 65536, the cap rejects the image body and the fallback busts the limit too; `gh api` returns 422, `set -e` fails the step, and the PR gets no report at all — the outcome the comment says must not happen.

Not reachable today (341 screens × ~80 chars ≈ 27KB, and the ceiling is the screen count, not the diff size), so this is a latent edge rather than a bug. A `head -c` on the text in the fallback branch would make the claim unconditionally true.

.github/workflows/visual-regression.yml
```diff
@@ -242,72 +338,320 @@ jobs:
...
+            # points at are actually being served. Composing or publishing
+            # them is allowed to fail; posting URLs that 404 is not.
+            elif [ "${PUBLISH_OUTCOME}" = "success" ] \
+              && [ -s "${PAYLOAD_DIR}/comment-images.md" ] \
+              && [ "$(wc -c < "${PAYLOAD_DIR}/comment-images.md")" -le "${MAX_COMMENT_BYTES}" ]; then
+              cat "${PAYLOAD_DIR}/comment-images.md"
             else
+              echo "### Visual regression"
+              echo ""
               echo ":framed_picture: **${CHANGED} screen(s) changed** against the merge-base."
```
gbarkhatov: Minor consistency: [`visual-embed.mjs:1245-1247`](https://github.com/babylonlabs-io/babylon-toolkit/blob/587abf1ee4b8f186f4e404bc8a9d2178b3cccaa9/scripts/visual-embed.mjs) argues — correctly — that a comment rewritten in place must name the commit it describes, "so without the sha a reader has no way to tell a current report from one left behind by a run that failed to update it". The three branches this job owns (:622, :630, :640, and this one) don't carry it, and they are the branches most likely to be the stale one. `${{ needs.visual.outputs.report_sha }}` is already available in this job.

.github/workflows/visual-regression.yml
```diff
@@ -242,72 +338,320 @@ jobs:
...
+          # of this repository fetches every branch, so this tree lands in
+          # each developer's checkout - against a ~31MB repo, an unbounded
+          # publish would make the diff images the largest thing in it.
+          # Normal runs land far below this; tripping it means the caps in
+          # visual-embed.mjs are misconfigured, and refusing loudly beats
+          # quietly bloating everyone's clone. It bounds one PR's
+          # contribution, not the branch: what bounds the branch is the reap
+          # of closed pull requests below, so the live tree stays
+          # proportional to how many are open at once.
+          MAX_PUBLISHED_BYTES=$(( 4 * 1024 * 1024 ))
```
gbarkhatov: 4MB is ~50× what this actually produces. The eight images currently on `origin/visual-diffs` for #2211 total 77KB, and the caps upstream bound it at 10 composites. As a runaway detector, a ceiling this high only trips once the branch is already heavier than the 31MB repo everyone clones. 512KB would still leave 6× headroom over observed and would catch a misconfigured cap on the first run rather than the fiftieth.

.github/workflows/visual-regression.yml
```diff
@@ -242,72 +338,320 @@ jobs:
...
+              '# visual-diffs' \
+              '' \
+              'Machine-written. Holds the before/after PNGs that the Visual regression' \
+              'check embeds in pull request comments, and nothing else.' \
+              '' \
+              '- Written only by `.github/workflows/visual-regression.yml`.' \
+              '- Shares no history with `main`. Never merge it anywhere.' \
+              '- Rewritten as one parentless commit per publish, so the branch has no' \
+              '  history to grow and a fresh clone transfers only the live tree.' \
+              '- `pr<number>/<head-sha>/` is deleted once that pull request closes, so' \
```
gbarkhatov: The layout is `pr<number>/<sha12>-<run-id>/`, not `pr<number>/<head-sha>/` — e.g. `pr2211/083e7a5a2df9-31475186640/` on the branch today. Since this README exists to answer "what is this and can I delete it" for someone who has never read the workflow, the path shape is the one line worth getting exact.

.github/workflows/visual-regression.yml
```diff
@@ -8,9 +8,41 @@ concurrency:
...
+  # raw.githubusercontent.com can serve them into a comment; an artifact
+  # cannot be embedded in Markdown. Orphan: it shares no history with main
+  # and must never be merged anywhere. Named here because the URL baked into
+  # the comment and the branch the images are pushed to have to agree
+  # exactly, and they are built in two different jobs.
+  VISUAL_DIFF_BRANCH: visual-diffs
+
+  # The surfaces that get captured, compared and published, as one list.
+  # Adding a third means editing this and .github/scripts/visual-capture.sh,
+  # and nothing else - the diff loop, the embed step and the publish
```
gbarkhatov: "and nothing else" doesn't hold for a third surface: `MIN_EMBEDDED_GROUPS_PER_SURFACE * 3 = 6 > MAX_EMBEDDED_GROUPS = 5`, and `selectEmbeddedGroups` already documents that the last surface in rank order silently loses its reserved slots. Since this comment is a maintenance instruction, it should name `MAX_EMBEDDED_GROUPS` as the third edit — otherwise whoever adds the surface follows the instruction and gets the guarantee quietly broken.

scripts/visual-embed.mjs
```diff
@@ -0,0 +1,781 @@
...
+    );
+  }
+
+  const ranked = rankGroups(surfaces.flatMap((surface) => surface.groups));
+  const embedded = selectEmbeddedScreens(ranked);
+
+  await fs.mkdir(outDir, { recursive: true });
+  const coverage = new Map();
+  for (const entry of embedded) {
+    coverage.set(
```
gbarkhatov: No per-entry guard, so one truncated capture takes down all ten pictures: `PNG.sync.read` throws out of `main`, the step exits 1, `embed_outcome != 'success'`, publish is skipped, and the comment silently degrades to the text listing. Both captures are `continue-on-error`, so a half-written PNG is a realistic input rather than a hypothetical one.

Everything else in this pipeline degrades per-surface or per-file — this is the one place a single bad byte costs the whole feature. A `try/catch` around the `writeComposite` call that drops that entry (and filters it out of `embedded` before `renderBody`, so no group renders an image URL that was never written) keeps the other nine.

scripts/visual-embed.mjs
```diff
@@ -0,0 +1,781 @@
...
+    diffText: await fs.readFile(diffTextFile, "utf8"),
+  });
+  await fs.writeFile(bodyFile, `${body}\n`);
+
+  process.stdout.write(
+    `Composed ${embedded.length} before/after image(s) into ${outDir}.\n`,
+  );
+}
+
+main().catch((error) => {
```
gbarkhatov: `main()` is called unconditionally and the module exports nothing, so none of this is reachable from a test — while `visual-diff.mjs:340` added the `import.meta.url` guard deliberately so that it would be.

The functions that guard is worth having for are exactly the ones whose comments record a rule that was already got wrong once: `cropWindow`'s rule 2 ("looked harmless and was not: … two byte-identical panels under a caption reading 78% of pixels changed"), `changedRowBand`, and the per-surface floor in `selectEmbeddedGroups`. All are pure and take plain objects. Same guard here plus a few named exports costs three lines.

Related: no test file in the repo imports `visual-diff.mjs`, so its "Importing this module (the unit tests do)" at :338 is already false, and `changedPixelCutoff`/`compareOne` are exported for tests that don't exist. This PR makes that guard load-bearing for the first time — worth correcting the comment to say so.

scripts/visual-embed.mjs
```diff
@@ -0,0 +1,781 @@
...
+ * Groups each surface is guaranteed, before the rest of the budget is
+ * handed out on rank alone.
+ *
+ * Rank alone is not enough. Repainting one shared component moves a
+ * Storybook story by 17% of its small frame and a vault route by 1% of a
+ * tall page, so a purely global ranking spends every slot on stories and
+ * shows the reviewer nothing of the app that ships. This floor is what
+ * keeps a vault screen in the comment.
+ */
+const MIN_EMBEDDED_GROUPS_PER_SURFACE = 2;
```
gbarkhatov, line 67: Optional, but there's evidence it bites: on #2211 both of vault's guaranteed slots went to `vault/activity--mobile.png` and `vault/liquidations--mobile.png`, which are **byte-identical** (both blob `d5d6aa8e` on `origin/visual-diffs`) — presumably the same empty/unconnected state on two routes. Grouping can't catch it, since they are different groups by construction.

Hashing the composite bytes and skipping a screen whose hash is already embedded would spend that slot on a distinct finding instead. Cheap to add in the `writeComposite` loop, and it composes with the floor rather than replacing it.

gbarkhatov (dismissed):
## Review

Turns the visual-regression comment from a filename listing into embedded before/after images, and splits the workflow so the write-capable token lives in a job that never checks out the PR.

The design holds up. The privilege split, `--force-with-lease` with an explicit lease (and an empty lease asserting branch-nonexistence), repeating the whole read-modify-write inside the retry loop rather than just the push, `!cancelled()` over `always()`, the dependabot carve-out, and `cropWindow` anchoring the height window on the changed band instead of row 0 are each correct and each non-obvious. The reasoning is written down at the point it applies, which is what made this reviewable at all. I checked `origin/visual-diffs` — the mechanism works end to end, 8 images at 77KB for #2211.

Approving; nothing below blocks. Three are small and I'd take them before this becomes a blocking check:

- **`visual-regression.yml:674`** — `--paginate` applies `--jq` per page, so `.[0].id` yields one id per page and the duplicate-reaping loop at :693 never fires for same-page duplicates, which is the case it was written for. Same expression is also missing an author filter on a path that ends in `DELETE`.
- **`visual-regression.yml:40`** — `run_id` is stable across re-runs (only `run_attempt` increments), so `RUN_DIR` repeats exactly where the comment says it must not. Since the baseline is re-resolved per run, attempt 2 writes different composites to the URL camo cached for attempt 1.
- **`visual-embed.mjs:753`** — one unreadable PNG throws out of `main` and loses all ten pictures.

The rest are follow-ups: the `visual-embed.mjs` module has no test seam (and the pure functions it hides are the ones whose comments record a rule that was already wrong once), the publish job's "both are treated as untrusted" claim is broader than what it enforces, and a few comments that promise slightly more than the code delivers.

jrwbabylonlab (dismissed):

.github/workflows/visual-regression.yml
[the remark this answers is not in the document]
jonybur: Fixed in eb919eb5b - and you and @gbarkhatov independently found the same thing.

Took the version that closes both this and the author filter in one expression, since they live in the same `--jq`:

```
--jq ".[] | select((.body | startswith(\"$MARKER\")) and .user.login == \"github-actions[bot]\") | .id"
```

`.[]` rather than `map(...) | .[0]` so every match on a page is emitted, which is what the reap loop below needed. The comment above it now says why `.[0]` was wrong rather than just why `head -n1` would be.

.github/workflows/visual-regression.yml
```diff
@@ -316,15 +660,39 @@ jobs:
...
+          # There must be exactly one of these on a pull request, and only
+          # the one patched above is current. A second can exist from a run
+          # whose lookup missed - the un-paginated lookup this replaced left
+          # some behind - and it would sit there forever showing a result
+          # from a commit that has since been rewritten. Reap rather than
+          # leave two contradictory reports.
+          if [ "${MATCHES}" != "${EXISTING}" ]; then
+            for STALE in ${MATCHES#*$'\n'}; do
+              gh api --method DELETE "repos/${GITHUB_REPOSITORY}/issues/comments/${STALE}" > /dev/null
+              echo "Deleted duplicate report comment ${STALE}."
```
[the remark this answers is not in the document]
jonybur: Fixed in eb919eb5b. The lookup now filters on `.user.login == "github-actions[bot]"` as well as the marker, so a participant quoting the raw marker is no longer a delete target.

Worth stating the reachability, since it is narrower than it first reads: the marker is an HTML comment, so it would take pasting the raw source rather than the rendered text. Fixed regardless - the path ends in an irreversible DELETE of someone else's writing, and the fix rode along in the same expression as the pagination one.

.github/workflows/visual-regression.yml
```diff
@@ -242,72 +338,320 @@ jobs:
...
+  # `main` and `release/*` are out of reach either way - the org ruleset
+  # requires signed commits and a pull request, and GITHUB_TOKEN cannot
+  # bypass it.
+  publish:
+    needs: visual
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+      pull-requests: write
```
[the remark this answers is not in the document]
jonybur: Accurate, and it stands as written - this is a documented limitation rather than an oversight. On `pull_request` the workflow file comes from the PR branch, so a same-repo author can move the permission wherever they like; no arrangement of `permissions:` inside this file can prevent that.

What the split is actually for is the realistic threat: a compromised transitive dependency executing inside the capture job, which installs and runs the PR's whole tree. That code reaches a read-only token.

`main` and `release/*` are out of reach either way - the org ruleset requires a pull request and hardware-signed commits, and GITHUB_TOKEN cannot bypass it.

eb919eb5b did rewrite the comment on this job to be precise about what is and is not enforced, because @gbarkhatov was right that "both are treated as untrusted" overclaimed. See that thread.

.github/workflows/visual-regression.yml
jonybur: Both fixed in eb919eb5b, with your expression verbatim - the inner parens included.

You are right that the reap could never fire for the case it was written for: duplicates come from consecutive runs, so they sit adjacent and share a page, and `.[0].id` collapsed them to one id per page. Only a duplicate straddling a page boundary ever got reaped.

The comment above the lookup now records why `.[0]` was wrong, not only why `head -n1` would be - the old note explained the per-page behaviour and then the code depended on it being otherwise, which is how this survived being written down.

.github/workflows/visual-regression.yml
jonybur: Fixed in eb919eb5b. `${GITHUB_RUN_ATTEMPT}` appended at both sites, and the env comment rewritten to say the attempt is part of the key.

Your framing of why it matters is the part I had wrong: I was treating the path as "identifies the commit", when the baseline is re-resolved per run and re-running is exactly what people do when main has moved. So attempt 2 is not a retry of the same comparison, it is a different one - writing to a URL camo had already cached, with both steps exiting 0. That is a worse failure than the 404 the comment was worried about, because it looks correct.

.github/workflows/visual-regression.yml
jonybur: Fixed in eb919eb5b. Now `pr<number>/<short-sha>-<run-id>-<run-attempt>/` - the attempt is new in this push, from your `run_id` thread.

Also split the sentence so the deletion claim attaches to the `pr<number>/` directory, which is what actually gets reaped, rather than to the leaf.

.github/workflows/visual-regression.yml
jonybur: Fixed in eb919eb5b - 512KB, with the reasoning you gave written into the comment.

The 4MB figure came from asking "what would be bad for the repo" instead of "what would mean the caps upstream are broken". A detector that only trips once the branch outweighs the 31MB everyone clones is not detecting anything useful.

.github/workflows/visual-regression.yml
jonybur: Fixed in eb919eb5b. `head -c` on the text in both branches that embed it, via a named `MAX_DIFF_TEXT_BYTES`.

Latent as you say - the ceiling is the screen count, not the diff size - but the comment above `MAX_COMMENT_BYTES` claims the fallback is unconditionally safe, and it was only safe by arithmetic that holds until someone adds screens. Cheaper to make the sentence true than to leave the reader to re-derive it.

.github/workflows/visual-regression.yml
jonybur: Fixed in eb919eb5b. All four branches now carry it, via an `AT_SHA` built once from `needs.visual.outputs.report_sha` - same output the published path is keyed on, so the sha in the text cannot disagree with the sha in the image URLs.

One correction to the line reference: the argument you quoted is in `renderBody`, not at :1245 - the file is 781 lines. Does not change the point.

Left the job summary alone: it already names both refs on its first line.

scripts/visual-embed.mjs
jonybur: Fixed in eb919eb5b. `writeComposite` is now `composeComposite` and returns the bytes instead of writing them, so the caller owns the loop and a throw costs one screen.

Two things fell out of doing it that way. The body is rendered from what was actually written rather than from the selection, so a group cannot emit an image URL for a file nobody wrote - that would have swapped one failure mode for a subtler one. And the drop is not silent: `renderGroup` folds the screen into its "N more changed screens in this group" line and the full text listing still names it.

Verified against a synthetic report carrying a deliberately truncated PNG: the bad screen is skipped, the other three are written, exit 0.

scripts/visual-embed.mjs
jonybur: Taken, in eb919eb5b - the composite bytes are hashed and a screen whose hash is already embedded is skipped.

One honest caveat, written into the comment: this frees the slot rather than reassigning it. Selection has already run by that point, so the comment shows fewer pictures, not a different one. Reassigning means restructuring selection into a fill-until-full loop, which I did not want to fold into a review-fix commit - happy to take it as a follow-up if you think the half-measure is worse than nothing here.

Verified on a fixture with two byte-identical vault routes: second one skipped, and no URL for it in the body.

scripts/visual-embed.mjs
jonybur: Fixed in eb919eb5b, and the related note with it.

Added the same `import.meta.url` guard and exported `changedRowBand`, `cropWindow` and `selectEmbeddedGroups` - the three you named. 16 tests across `scripts/__tests__/`, including the two rules whose comments record a mistake already made: that `cropWindow` anchors an over-tall window on the band rather than row 0, and that the per-surface floor keeps a vault screen in the comment when rank alone would not.

Also pinned the third-surface degradation as a test rather than leaving it as prose - see your `VISUAL_SURFACES` thread.

On the related part: you were right that `visual-diff.mjs`'s guard comment was already false and that `changedPixelCutoff`/`compareOne` were exported for tests that did not exist. Rather than correct the comment to say so, I wrote the tests - both are now genuinely imported, along with `escapeHtml` and `parseArgs`. The corrected comment names `visual-embed.mjs` as the real importer.

These run in the workflow before the Playwright install, deliberately not `continue-on-error`: everything downstream is advisory, and a broken crop rule still produces a picture, just not of the change.

.github/workflows/visual-regression.yml
jonybur: Fixed in eb919eb5b, though not the way you suggested - tell me if you would rather have the cap raised.

The comment now names `MAX_EMBEDDED_GROUPS` as the third edit and spells out the arithmetic (floor 2 x 3 surfaces = 6 reserved against a cap of 5) and the consequence: the surface that loses its reservation is the last in rank order, which is the one just added. I left the cap where it is because raising it pre-emptively for a surface that does not exist would be guessing at the right comment length.

Also pinned it as a test - `selectEmbeddedGroups under-serves the last surface once a third is added` - so the behaviour is documented somewhere that fails when it changes, not only in a comment.

.github/workflows/visual-regression.yml
jonybur: Took the cheap version in eb919eb5b, plus one piece of the structural one. Flagging that this is the only thread I did not close out fully.

The comment on the publish job now separates enforced from not: filenames matched against `SAFE_NAME`, total published size capped, comment body length-checked - versus PNG contents never parsed and the Markdown `cat` in as written. It says plainly that capture-job code can hand this job arbitrary Markdown to post as `github-actions[bot]` and arbitrary `*.png` bytes to commit to a public branch, that this is smaller than a write token, and that it is not an airlock.

The piece I did take: `group.group` and `groupHeadline` now go through `escapeHtml`, which I exported from `visual-diff.mjs` rather than reimplementing, so a fix in one renderer cannot leave the other behind. You were right that it is inert only because Storybook ids happen to match `[a-z0-9-]`, which is a property of the current capture rather than a guarantee.

What I did not do is move `renderBody` across the boundary and make the payload a typed manifest. That is the actual fix and I think it should land before this check goes blocking, but it restructures the script/job seam and does not belong in a commit that is otherwise review fixes. The comment names it as the intended direction so it is not lost. Happy to open it as a follow-up issue - or to do it here if you would rather not merge the seam as it stands.

jonybur:
@greptile-apps

gbarkhatov (approved):

PR #2224, opened 2026-08-12
fix(ts-sdk): assert operation-key multicall results are roster-aligned

### LINKED ISSUE #2191: sdk: getPayoutScriptsAtEpochs does not length-check its multicall result, so a short read blames the wrong claimer
Raised by @jonybur reviewing [#2173](https://github.com/babylonlabs-io/babylon-toolkit/pull/2173). Verified present at `bc5b2679`.

`ViemOperationKeyReader.getPayoutScriptsAtEpochs` returns registry-supplied payout scripts index-aligned with the caller's keeper roster, but never checks that the returned array is as long as that roster. Its sibling read of operation keys does check.

## The asymmetry

`packages/babylon-ts-sdk/src/tbv/core/clients/eth/operation-key-reader.ts:155-187`

```ts
    })) as readonly Hex[];

    return {
      vaultProvider: results[0],
      vaultKeepers: [...results.slice(1)],
    };
```

No length assertion. Compare `finalize` in `packages/babylon-ts-sdk/src/tbv/core/services/participants/resolveParticipantKeys.ts`, which guards the equivalent case for keys:

```ts
  if (
    raw.vaultKeepers.length !== query.vaultKeepers.length ||
    raw.universalChallengers.length !== query.universalChallengers.length
  ) {
    throw new Error(
      `Operation-key resolution returned ${raw.vaultKeepers.length} keeper and ` +
      ...
```

## Why it matters — error attribution, not safety

A short array yields `undefined` for some keeper in the map that `vaultPayoutSignatureService` builds from `payoutScripts.vaultKeepers[i]`. That **does** fail closed: `payout.ts` throws on `registered === undefined` before anything is signed. So there is no path to signing against a wrong or missing destination.

What is wrong is the blame. The error surfaces as *"no registered payout script for claimer `<pubkey>`"*, pointing at a claimer that is perfectly fine, when the actual fault is a multicall that came back short. On a value-movement path where the operator response is already the least trusted input, an error that names the wrong party costs real debugging time.

A length assertion mirroring `finalize`'s would name the real fault at the point it happens.

## Suggested fix

Add the check in `getPayoutScriptsAtEpochs` before constructing the return value: assert `results.length === 1 + query.vaultKeepers.length`, and throw naming both the expected and actual counts, in the style of `finalize`'s message.

Note `universalChallengers` has no counterpart here by design — a UC is never a claimer, so it has no payout script and no call in this multicall. The expected length is `1 + keepers`, not `1 + keepers + challengers`.

Add a unit test driving a short `multicall` result and asserting the new error, alongside the existing reader tests.

## Priority

Low. @jonybur explicitly left this one to our discretion; it fails closed today and only the diagnostic is wrong. Cheap enough to fold into any other work in this file.

## Context

- Review: https://github.com/babylonlabs-io/babylon-toolkit/pull/2173

gbarkhatov:
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2191

`ViemOperationKeyReader` reads RFC-006 operation keys and payout scripts with one `multicall` per method and then splits the flat result positionally — vault provider first, then keepers in roster order, then challengers. That positional alignment is the whole contract of the class: it is what lets the resolver pair a resolved key back to its admin address. None of the three methods checked that the result was actually as long as the roster it is being aligned against.

A short result does not throw. `results.slice(1)` just returns fewer elements, and the missing tail entries become `undefined`.

## With / without

| | Without this change | With this change |
|---|---|---|
| Short payout-script read | `vaultPayoutSignatureService` builds a claimer→script map with an `undefined` value, and `payout.ts` later throws **"No registered payout script resolved for vault keeper claimer `<pubkey>`"** | `getPayoutScriptsAtEpochs: multicall returned 2 results for 3 calls, so the results are not roster-aligned; refusing to use them` |
| Who gets blamed | A keeper that is perfectly fine — the one that happened to land in the missing slot | The read that came back incomplete |
| Short operation-key read via `resolveParticipantKeys` | Caught downstream by `finalize` | Caught at the reader, before partitioning |
| Short operation-key read by a direct `OperationKeyReader` consumer | Not caught at all | Caught at the reader |

Both cases already fail closed — nothing gets signed against a wrong or missing destination. What was wrong is the attribution, and on a value-movement path where the operator response is the least trusted input, an error naming the wrong party costs real debugging time.

## What changed

`packages/babylon-ts-sdk/src/tbv/core/clients/eth/operation-key-reader.ts`

A single `assertMulticallLength(actual, expected, label)` helper, called from the two places an untrusted array arrives:

- `readAll`, which backs `getCurrentOperationKeys` and `getOperationKeysAtEpochs`. Expected length is `1 + keepers + challengers`, so `readAll` now takes the challenger count and a label alongside the keeper count it already had.
- `getPayoutScriptsAtEpochs`, which does not go through `readAll`. Expected length is `1 + keepers`, with **no** challenger term — a universal challenger is never a claimer, so it has no payout script and no call in that multicall. That asymmetry is now commented, because it looks like an omission and isn't.

The error message follows the style of `finalize` in `resolveParticipantKeys.ts`, which already guards the equivalent case for keys: name both the expected and the actual count, and name the operation.

## Beyond the issue as filed

The issue scoped this to `getPayoutScriptsAtEpochs`. The `partition` helper behind the other two methods has the identical hole. Today that is covered downstream by `finalize`, but `OperationKeyReader` is public SDK surface — its own docs say "callers can use their own implementation" — so a consumer reading keys without going through `resolveParticipantKeys` gets no guard at all. Same defect class, same file, same helper, so all three are fixed here rather than leaving two-thirds of it for later.

The assertion sits in `readAll` rather than in `partition` itself: `readAll` is where the multicall result arrives, and keeping `partition` a pure slice means the guard cannot be bypassed by a future method that calls `partition` directly. Coverage is identical — `partition` has exactly one caller.

## Tests

New `packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/operation-key-reader.test.ts`, following the hand-rolled-stub style of the three sibling reader tests in that directory. `multicall` is stubbed directly, so no ABI encoding is involved.

- `getPayoutScriptsAtEpochs` returns the VP script and keeper scripts in roster order on a full-length result
- `getPayoutScriptsAtEpochs` throws naming both counts on a short result
- `getPayoutScriptsAtEpochs` issues `1 + keepers` calls and none for challengers — pins the asymmetry above
- `getOperationKeysAtEpochs` throws on a short result
- `getCurrentOperationKeys` throws on a short result

## Verification

```
pnpm --filter @babylonlabs-io/ts-sdk run build     # clean
pnpm run build                                     # clean
pnpm --filter @babylonlabs-io/ts-sdk exec vitest run   # 1497 passed (86 files), +5 from this PR
pnpm --filter vault run test                       # 3187 passed, 6 skipped
cd services/vault && npx tsc --noEmit -p tsconfig.lib.json   # clean
pnpm run lint                                      # 0 errors
```

No behaviour change for a well-formed multicall response, which is every response the reader has ever seen in practice — this only adds an error where there was previously a silent `undefined`.

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/operation-key-reader.test.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/operation-key-reader.test.ts
@@ -1,0 +1,139 @@
+import type { Address, Hex } from "viem";
+import { describe, expect, it, vi } from "vitest";
+
+import { ViemOperationKeyReader } from "../operation-key-reader";
+import type { KeyEpochs, OperationKeyQuery } from "../types";
+
+const MOCK_CONTRACTS = {
+  btcVaultRegistry: "0x1111111111111111111111111111111111111111" as Address,
+  applicationRegistry: "0x2222222222222222222222222222222222222222" as Address,
+  protocolParams: "0x3333333333333333333333333333333333333333" as Address,
+};
+
+const VP_GENESIS_KEY =
+  "0xaaaa000000000000000000000000000000000000000000000000000000000000" as Hex;
+
+/** Two keepers and one challenger, so a short read is unambiguous. */
+const QUERY: OperationKeyQuery = {
+  vaultProviderEthAddress:
+    "0x4444444444444444444444444444444444444444" as Address,
+  vaultProviderGenesisBtcPubkey: VP_GENESIS_KEY,
+  applicationEntryPoint:
+    "0x5555555555555555555555555555555555555555" as Address,
+  vaultKeepers: [
+    {
+      ethAddress: "0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" as Address,
+      btcPubKey:
+        "0xbbbb000000000000000000000000000000000000000000000000000000000000" as Hex,
+    },
+    {
+      ethAddress: "0xcccccccccccccccccccccccccccccccccccccccc" as Address,
+      btcPubKey:
+        "0xdddd000000000000000000000000000000000000000000000000000000000000" as Hex,
+    },
+  ],
+  universalChallengers: [
+    {
+      ethAddress: "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee" as Address,
+      btcPubKey:
+        "0xffff000000000000000000000000000000000000000000000000000000000000" as Hex,
+    },
+  ],
+};
+
+const EPOCHS: KeyEpochs = {
+  vpKeyEpoch: 3n,
+  appKeeperKeyEpoch: 4n,
+  ucKeyEpoch: 5n,
+};
+
+const VP_PAYOUT_SCRIPT = "0x0014aaaa" as Hex;
+const KEEPER_0_PAYOUT_SCRIPT = "0x0014bbbb" as Hex;
+const KEEPER_1_PAYOUT_SCRIPT = "0x0014cccc" as Hex;
+
+/** The subset of viem's `multicall` argument these tests assert on. */
+type MulticallArgs = { contracts: { functionName: string }[] };
+
+/** Stubs `multicall` with a fixed result, bypassing ABI encoding entirely. */
+function createReader(multicallResult: readonly Hex[]) {
+  const multicall = vi.fn(async (_args: MulticallArgs) => multicallResult);
+  const reader = new ViemOperationKeyReader(
+    { multicall } as never,
+    MOCK_CONTRACTS,
+  );
+  return { reader, multicall };
+}
+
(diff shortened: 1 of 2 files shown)


jrwbabylonlab (approved):

jonybur (approved):
Correct fix. Asserting the multicall result length at the reader turns the silent-undefined into a named error that identifies the read that came back short, rather than blaming whichever operator happened to land in the missing slot. Placing the guard in readAll rather than partition keeps the pure-slice contract and prevents future callers from bypassing it. The asymmetry on getPayoutScriptsAtEpochs (1 + keepers, no challenger term) is correct and documented. Tests cover full-length success, short-result failure with named counts, and pin the challenger-omission asymmetry.

Issue #2192, opened 2026-08-05
ops: verify RFC-006 registry shape before promoting the vault app to a network (no runtime check exists)


gbarkhatov:
RFC-006 shipped in [#2173](https://github.com/babylonlabs-io/babylon-toolkit/pull/2173) with **no feature flags** — deliberately, and the reasoning holds. The consequence, raised by @jonybur in review, is that the app's dependence on an RFC-006 registry is now a **deployment invariant rather than a runtime check**, and nothing enforces it before we point the app at a network.

There is no owner for that check today. This issue is to add one to the release/promotion process.

## Why a runtime check can't cover this

The epoch read goes through a deliberately isolated ABI (`BTCVaultRegistryKeyEpochs.abi.ts`) because `getBtcVaultProtocolInfo` returns a **dynamic tuple**. Asking for 16 head words against a 13-field registry does not run off the end of the returndata — it reads three words of *tail* and returns them as epochs. No revert, no warning, plausible-looking small integers.

`assertEpochInRange` in `vault-registry-reader.ts` rejects wild values, but it is explicitly documented as a weak backstop: a mis-decode can land on a small integer it accepts. The test at `packages/babylon-ts-sdk/src/tbv/core/contracts/__tests__/BTCVaultRegistryKeyEpochs.abi.test.ts` pins exactly this hazard.

## What is actually reachable

Narrower than the module docs currently imply (see #2190 item 3):

- Against a **fully pre-RFC-006 registry** the path fails closed anyway. `resolveParticipantKeysAtEpochs` goes on to call `getOperationBtcKeyAtEpochOrGenesis` on ApplicationRegistry / ProtocolParams; those do not exist there, so the multicall reverts loudly.
- The reachable bad state is a registry that **has the operation-key getters but an un-extended `BTCVaultProtocolInfo` struct**. Garbage epochs resolve with no error, and participant keys are resolved at a fabricated epoch.

That is the case this check exists to rule out.

## What blast radius looks like

Because both flags were dropped, there is no runtime switch to fall back to: rollback is "revert the deploy". And one deploy moves payout signing, refund, resume and VP auth simultaneously — they all resolve keys through the same epoch path.

## The check to add

Before promoting the vault app to **any** network (new network, redeployed registry, or a registry upgrade), verify against that network's `BTCVaultRegistry`:

1. All four RFC-006 getters exist and do not revert:
   - `getCurrentOperationBtcKey(address)`
[...]

PR #2225, opened 2026-08-12
docs(ts-sdk): correct the assertPayoutOutputLayout JSDoc

### LINKED ISSUE #2190: RFC-006: assertPayoutOutputLayout JSDoc contradicts its code (critical-path file), plus two narrower doc drifts
Raised by @jonybur reviewing [#2173](https://github.com/babylonlabs-io/babylon-toolkit/pull/2173). Three doc blocks describe behaviour that #2173 changed. All verified still present at `bc5b2679`.

This is comment-only work — no logic changes — but the first item is in a file that requires two code-owner approvals, and **@jonybur asked for it before #2173 merges**. If #2173 has already merged by the time anyone picks this up, it is straightforward standalone cleanup.

## 1. `assertPayoutOutputLayout` JSDoc contradicts its own code — critical-path file

`packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts:501-507`

The block describes the pre-#2173 behaviour in three places, each changed by that PR:

```
 * Returns the layout-trusted non-anchor script lengths for the fee band:
 * `out0Len` from the pinned outs[0] script (registered payout script, or
 * derived BIP-86 for the VK-claimer role — see #2150 Stage 6 for the pending
 * registered-script switch); `out1Len` measured from the deliberately
 * unpinned VP-claimer commission output, `undefined` for the other roles.
 * Both are rejected outside the contract's registration cap `[1, 128]`.
```

- **"derived BIP-86 for the VK-claimer role"** — the VK-claimer branch no longer derives BIP-86 as its expectation; it pins the registered script and accepts BIP-86 only as the transitional second value (see #2182).
- **"see #2150 Stage 6 for the pending registered-script switch"** — that pending switch *is* #2173. The forward reference is now backwards.
- **"`out1Len` measured from the deliberately unpinned VP-claimer commission output"** — `outs[1]` is pinned now.
- **"Both are rejected outside the contract's registration cap `[1, 128]`"** — the `out1Len` cap was deliberately deleted in #2173, because `assertStandardPayoutScript` bounds that output to 34 bytes and makes the cap unreachable. Only `out0Len` still has it.

Four lines up, the same block says the commission script is pinned *"when `vpCommissionScriptPubKey` is supplied"* — but #2173 made that parameter required on `PayoutParams`, so there is no absent case.

**Why this one matters more than a normal stale comment:** `payout.ts` is on the critical-path list in `.github/CODEOWNERS` and `CLAUDE.md`. Two code owners are required to read it and to be able to explain every line. The JSDoc is the first thing they read, and right now it describes a validator that no longer exists.

## 2. Flag leftover in `validateOnChainParticipantKeys`

`packages/babylon-ts-sdk/src/tbv/core/services/deposit/validateOnChainParticipantKeys.ts:118`

```ts
// Resolve the current operation keys, if enabled. This is what we build the
// Bitcoin lock with; the registration keys above stay only as the primary
// comparison target for indexer hints.
```

There is no enable/disable any more — both RFC-006 feature flags were removed before #2173 merged. This survived the stale-prose sweep in `bc5b2679` because that sweep grepped for `flag`/`capability`, and this phrases it as "if enabled".

Worth grepping for other phrasings of the same idea while fixing: `if enabled`, `when enabled`, `opt-in`, `gated`.

## 3. The partial-upgrade hazard is described more broadly than it is

`packages/babylon-ts-sdk/src/tbv/core/clients/eth/vault-registry-reader.ts` (the `UINT64_EXCLUSIVE_UPPER_BOUND` block and `getVaultKeyEpochs`), and the matching text in `clients/eth/types.ts`.

Those docs say reading the extended `getBtcVaultProtocolInfo` ABI against a pre-RFC-006 registry silently yields garbage epochs, and that correctness rests on a deployment invariant. True as far as it goes, but @jonybur pointed out it overstates the reachable risk:

- Against a **fully legacy** registry the path fails closed anyway. `resolveParticipantKeysAtEpochs` goes on to call `getOperationBtcKeyAtEpochOrGenesis` on ApplicationRegistry/ProtocolParams, which does not exist there, so the multicall reverts and the garbage epochs are never used.
- The genuinely reachable gap is narrower: a registry that **has** the operation-key getters but whose `BTCVaultProtocolInfo` struct is **not** extended. Garbage epochs would then resolve without error.

That narrower case is the one `assertEpochInRange` is actually load-bearing for, and the docs should say so — right now they justify the check with a scenario that cannot occur.

No network is in that state, so this is documentation accuracy rather than a live risk.

## Verification

Comment-only, so the suites should be untouched:

```bash
nvm use 24
pnpm run build
pnpm --filter @babylonlabs-io/ts-sdk exec vitest run
pnpm --filter vault run test
pnpm run lint
```

Format only the files you touch, running prettier from inside the package directory — `prettier-plugin-organize-imports` only resolves inside the TS project.

## Context

- Review: https://github.com/babylonlabs-io/babylon-toolkit/pull/2173
- Related cleanup that will touch the same JSDoc: #2182

gbarkhatov:
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2190

Comment-only. No logic changes, no test changes, no exported surface touched.

Two doc blocks describe behaviour that https://github.com/babylonlabs-io/babylon-toolkit/pull/2173 changed. One of them is the JSDoc on the payout-output validator, which is on the critical-path list in `CLAUDE.md` and `.github/CODEOWNERS` — the list whose whole premise is that two reviewers can explain every line. The JSDoc is the first thing they read, and it currently describes a validator that no longer exists.

## With / without

| | Without this change | With this change |
|---|---|---|
| A code owner reviewing `payout.ts` | Reads that `outs[1]` is "deliberately unpinned", that the VK-claimer branch derives BIP-86 as its expectation, and that both script lengths are capped at `[1, 128]` — none of which the code does | Reads what the code does |
| A reader of `getVaultKeyEpochs` | Told the hazard is "a registry that predates RFC-006", which is the one case that actually fails closed | Told the hazard is a registry with the getters but an un-extended struct, which is the case that resolves silently |
| `assertEpochInRange` | Justified by a scenario it is not reachable for | Justified by the scenario it is the only guard for |

Nothing here is a live risk. No network is in the state item 2 describes, and the payout validator behaves correctly — it is only documented wrongly.

## Item 1 — `assertPayoutOutputLayout` JSDoc

`packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts`

What the code actually does, which is what the JSDoc now says:

| Role | `outs.length` | `outs[0].script` accepted | `outs[1]` |
|---|---|---|---|
| vp-claimer | 3 | registered payout script only | script pinned (registered **or** BIP-86 over the VP key); value capped at `floor(peginValue × bps / 10_000)` |
| depositor-as-claimer | 2 | registered payout script only | n/a |
| vk-claimer | 2 | registered script **or** BIP-86 over the claimer key | n/a |

Corrections:

1. **"derived BIP-86 for the VK-claimer role"** — the VK branch pins the *registered* script and accepts BIP-86 only as a transitional second candidate, via `acceptedPayoutScriptHexes`. The old wording had it as the expectation rather than the fallback.
2. **"see #2150 Stage 6 for the pending registered-script switch"** — deleted. That switch *was* #2173, so the forward reference pointed backwards.
3. **"`out1Len` measured from the deliberately unpinned VP-claimer commission output"** — `outs[1].script` is pinned now.
4. **"Both are rejected outside the contract's registration cap `[1, 128]`"** — only `out0Len` is. The `out1Len` cap was deliberately deleted in #2173 because `assertStandardPayoutScript` bounds that output to a standard type and makes the cap unreachable; the code already carries that reasoning inline, and the JSDoc now agrees with it.
5. **"when `vpCommissionScriptPubKey` is supplied"** — it is a required field on `PayoutParams`, so there is no absent case.

Two additions beyond correcting what was wrong:

- The summary paragraph listed what is pinned per role and omitted `outs[1].script`. A code owner reading only the summary would not learn that the commission destination is checked at all, so it is listed now.
- A short paragraph on the dual-accept asymmetry: `acceptedPayoutScriptHexes` backs the VK-claimer's `outs[0]` and the VP commission `outs[1]`, while the VP-claimer and depositor-as-claimer `outs[0]` is pinned to the registered script alone. That difference is load-bearing and was not written down anywhere.

The cap is now referred to as `MAX_PAYOUT_SCRIPT_LEN` rather than restating `128`, so a future bump cannot re-stale the prose.

## Item 2 — already fixed, no change here

The issue's second item was a `validateOnChainParticipantKeys.ts` comment saying "Resolve the current operation keys, **if enabled**" after both RFC-006 feature flags were removed. That was fixed incidentally by https://github.com/babylonlabs-io/babylon-toolkit/pull/2197 (`637fe865`); the comment now reads "Resolve the current operation keys. This is what we build the Bitcoin lock with".

I still ran the sweep the issue asked for. A repo-wide grep for `if enabled` / `when enabled` / `opt-in` / `gated` returns a lot, but all of it describes live mechanisms — auth-gated VP RPC methods, current feature flags, conditional UI. Narrowed to the surface where a removed RFC-006 flag could plausibly still be described — `services/participants/`, `services/deposit/`, `clients/eth/` and `services/vault/src/services/vault/` — there are no hits for "enabled" at all, and no residual RFC-006 flag identifier anywhere in either package. So item 2 is closed with nothing left over.

## Item 3 — narrow the partial-upgrade hazard

`packages/babylon-ts-sdk/src/tbv/core/clients/eth/vault-registry-reader.ts` (the `UINT64_EXCLUSIVE_UPPER_BOUND` block and `getVaultKeyEpochs`) and the matching interface doc in `clients/eth/types.ts`.

All three justified the epoch range check with "a registry that predates RFC-006". True but overstated, as @jonybur pointed out: against a *fully* legacy registry the path fails closed on its own, because `resolveParticipantKeysAtEpochs` goes on to call `getOperationBtcKeyAtEpochOrGenesis` on ApplicationRegistry and ProtocolParams, which do not exist there — the multicall reverts and the garbage epochs are never used.

The reachable gap is narrower: a registry that **has** the operation-key getters but whose `BTCVaultProtocolInfo` struct is **not** extended. There the tail-data epochs resolve with no error, and the range check is the only thing looking at them. All three blocks now say that, and the reader block points at https://github.com/babylonlabs-io/babylon-toolkit/issues/2192 for where the deployment-side check is owned.

The existing "only a backstop, and a weak one" caveat is kept — a mis-decode can still land on a small integer the range accepts. That part was accurate, and it is the reason #2192 exists.

## Verification

Comment-only, so the check is that nothing moved:

```
pnpm --filter @babylonlabs-io/ts-sdk run build     # clean
pnpm run build                                     # clean
pnpm --filter @babylonlabs-io/ts-sdk exec vitest run   # 1492 passed (85 files) — unchanged
pnpm --filter vault run test                       # 3187 passed, 6 skipped — unchanged
cd services/vault && npx tsc --noEmit -p tsconfig.lib.json   # clean
pnpm run lint                                      # 0 errors
```

All three files were already prettier-clean, so no `--write` ran and no unrelated region could be reformatted. `git diff` contains only comment lines — every changed line starts with `*`, `/**` or `*/`.

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/eth/types.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/eth/types.ts
@@ -144,11 +144,13 @@ export interface VaultRegistryReader {
   /**
    * Read a vault's frozen RFC-006 key epochs.
    *
-   * Uses the extended `getBtcVaultProtocolInfo` ABI. Against a registry that
-   * predates RFC-006 this returns silent garbage for a populated vault rather
-   * than throwing, so it must only be called against an RFC-006 registry —
-   * a deployment invariant, not something this call can detect. See
-   * `BTCVaultRegistryKeyEpochs.abi.ts`.
+   * Uses the extended `getBtcVaultProtocolInfo` ABI. Against a registry whose
+   * `BTCVaultProtocolInfo` struct is not extended this returns silent garbage
+   * for a populated vault rather than throwing, so it must only be called
+   * against an RFC-006 registry — a deployment invariant, not something this
+   * call can detect. A registry missing the operation-key getters altogether is
+   * the safer case: key resolution reverts downstream and these epochs are never
+   * used. See `BTCVaultRegistryKeyEpochs.abi.ts`.
    */
   getVaultKeyEpochs(vaultId: Hex): Promise<KeyEpochs>;
   /** {@link getVaultKeyEpochs} for many vaults in one multicall. */
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/eth/vault-registry-reader.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/eth/vault-registry-reader.ts
@@ -91,11 +91,24 @@ function mapVaultProtocolInfo(result: RawVaultProtocolInfo): VaultProtocolInfo {
  *
  * viem does not range-check a decoded `uint64`, so a misaligned decode can
  * hand back a value far larger than the field can hold. Rejecting those is
- * cheap defence in depth against reading the extended ABI on a registry that
- * predates RFC-006 — but it is only a backstop, and a weak one: a misaligned
- * decode can also land on a small, plausible-looking integer that this range
- * accepts. Correctness rests on only ever pointing at an RFC-006 registry.
- * See `BTCVaultRegistryKeyEpochs.abi.ts`.
+ * cheap defence in depth against a mis-decode of the extended ABI.
+ *
+ * The state this is load-bearing for is narrower than "a registry that predates
+ * RFC-006". Against a *fully* pre-RFC-006 registry the path fails closed on its
+ * own: `resolveParticipantKeysAtEpochs` goes on to call
+ * `getOperationBtcKeyAtEpochOrGenesis` on ApplicationRegistry and
+ * ProtocolParams, which do not exist there, so the multicall reverts and these
+ * epochs are never used. The reachable gap is a registry that *has* the
+ * operation-key getters but whose `BTCVaultProtocolInfo` struct is not
+ * extended — there the tail-data epochs resolve with no error at all, and this
+ * range check is the only thing looking at them.
+ *
+ * Even for that case it is only a backstop, and a weak one: a misaligned decode
+ * can land on a small, plausible-looking integer this range accepts. Correctness
+ * rests on only ever pointing at an RFC-006 registry, which is a deployment
+ * precondition rather than something this call can establish — see
+ * https://github.com/babylonlabs-io/babylon-toolkit/issues/2192 for where that
+ * check is owned, and `BTCVaultRegistryKeyEpochs.abi.ts` for the decode hazard.
  */
 const UINT64_EXCLUSIVE_UPPER_BOUND = 1n << 64n;
 
@@ -218,11 +231,17 @@ export class ViemVaultRegistryReader implements VaultRegistryReader {
    * Read a vault's frozen RFC-006 operation-key epochs.
    *
    * Reads `getBtcVaultProtocolInfo` through the **extended** ABI, which is only
-   * valid against an RFC-006 registry: against one that predates RFC-006 this
-   * call does not fail for a populated vault, it silently returns three words
-   * of tail data as epochs. Nothing here can detect that, so the guarantee is a
-   * deployment one — every network this ships to has the RFC-006 getters, and
-   * mainnet is a fresh RFC-006 deploy. See `BTCVaultRegistryKeyEpochs.abi.ts`.
+   * valid against an RFC-006 registry: against one whose `BTCVaultProtocolInfo`
+   * struct is not extended this call does not fail for a populated vault, it
+   * silently returns three words of tail data as epochs. Nothing here can detect
+   * that, so the guarantee is a deployment one — every network this ships to has
+   * the RFC-006 getters, and mainnet is a fresh RFC-006 deploy.
+   *
+   * A registry missing the operation-key getters entirely is the *safer* of the
+   * two cases: `resolveParticipantKeysAtEpochs` reverts downstream and these
+   * epochs never reach key resolution. See {@link UINT64_EXCLUSIVE_UPPER_BOUND}
+   * for which state the range check actually guards, and
+   * `BTCVaultRegistryKeyEpochs.abi.ts` for the decode hazard.
    */
   async getVaultKeyEpochs(vaultId: Hex): Promise<KeyEpochs> {
     const result = (await this.publicClient.readContract({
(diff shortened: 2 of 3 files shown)


jrwbabylonlab (approved):

jonybur (approved):
Comment-only doc corrections. The new JSDoc on assertPayoutOutputLayout matches what the code does: outs[1].script is pinned when vpCommissionScriptPubKey is a required parameter, the VK-claimer accepts the registered script with BIP-86 as a transitional fallback (via acceptedPayoutScriptHexes), the deleted #2150 forward reference is gone, and MAX_PAYOUT_SCRIPT_LEN replaces the hardcoded 128. The narrowed partial-upgrade note on UINT64_EXCLUSIVE_UPPER_BOUND / getVaultKeyEpochs is accurate: a fully legacy registry fails closed downstream via the missing ApplicationRegistry/ProtocolParams getters; the reachable gap is a registry that exposes the operation-key getters but whose BTCVaultProtocolInfo struct is not extended, and that is where the range check is the only guard.

Issue #2228, opened 2026-08-12
[eth-first] Use the Vault with just an Ethereum wallet


jonybur:
## Outcome

Users can enter and use the Vault with Ethereum alone. Bitcoin is requested only for actions that need its wallet.

Status: the feature is complete and merged behind a disabled control. Only testing and validation remain. They are not part of the feature estimate.

## Done (feature complete)

- Bitcoin wallet setup is self-contained. #2229
- Apps can require one wallet and offer the other. #2445
- Each wallet disconnects without the other. #2439
- Vault and staking apply the selected consent rule. #2491
- Real wallets restore their sessions after reload. #2483
- The Vault asks for Bitcoin before Bitcoin actions. Cancel starts nothing. #2482
- The Vault keeps Ethereum when Bitcoin disconnects. #2499
- Payout addresses stay correct with Ethereum alone. #2500
- Ethereum-only access works behind the disabled control. #2501
- CI runs the wallet and Vault browser checks. #2463

## Validation (not part of the feature estimate)

- [ ] #2232: the dashboard, loan, activity, and liquidation pages, borrow, and the refund guard are covered. Add control-on tests without Bitcoin for repay and withdraw. Repair the reorder test, which a mock clash on main stops from running. Then close #2232.
- [ ] #2355: the real OKX and MetaMask Vault run passed, and the runner drives the Vault with Ethereum alone. Run what is left: withdraw, multi-hub, and Keplr staking against a working RPC host. Record the result and close #2355.
- [ ] #2354: #2491 merged the consent rule, but its decision record and release note did not. Decide where they live, close #2354, and close #2230 after #2355.

## After this epic (required before release)

- Release #2233: final browser journey, release owner, approved release, and tested rollback.
- Security #2231. A security defect or missing required review that applies to the release blocks the release.

## Optional (blocks nothing)

- Deferred fixes #2506 (Retry after consent loss) and #2507 (cached download authentication).
- Raw engine rollback #2361 released as engine and SDK 1.0.0. It needs a second qualified review.
- Download-size reduction, a new address encoder, rollout stages, a rehearsal, and control removal.
- Aave repairs, product redesigns, and repository governance (#2359/#2370) are outside scope.

## Decisions

- 2026-09-16, jonybur: the feature is complete. Testing, validation, bug fixes, security work, and release actions are not part of the feature estimate. Bug fixes, security work, and release can ship after this epic closes.
[...]

Issue #2231, opened 2026-08-12
Complete Bitcoin engine safeguards and release checks


jonybur:
## Outcome

Complete the remaining public Bitcoin engine safeguards, qualified reviews, and coordinated release.

Security work is mandatory. A security defect or missing required security review that applies to the release blocks that release, regardless of which ticket tracks it. Separate tracking is not a waiver. Unrelated Vault implementation may proceed in parallel. Only performance improvements are optional; existing lazy loading and package contracts remain in place.

The engine, retry fix, browser tests, and package contract are delivered. The raw engine path is rolled back in #2361 and released as engine and SDK 1.0.0. The review conditions below remain.

## Problem

Some merged critical changes lack a second distinct qualified review record.

## Scope

- Complete #2361. #2516 rolled the raw path back to the engine's earlier exports. Engine and SDK 1.0.0 released it. A second qualified review remains.
- Complete the review conditions below. #2357 and #2386 retain the detailed findings and delivered code evidence.

## Complete when

- [x] Failed downloads can retry. The browser entry and production facade have direct tests. See #2356 and #2333.
- [x] Calculations match the previous implementations for supported inputs, with documented validation differences. [Comparison and package CI passed](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/33628092288).
- [x] The Ethereum-only package contract and release checks pass. #2346 records SDK 0.73.0 and optional engine 0.18.0. [Release passed](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/33760142986).
- [x] The raw path is rolled back. #2361 records the release: engine and SDK 1.0.0.
- [ ] Two distinct qualified reviewers approve the final identifier, key, and registration changes in #2332 and #2336. The [#2332 approval](https://github.com/babylonlabs-io/babylon-toolkit/pull/2332#pullrequestreview-5092102185) and [#2336 approval](https://github.com/babylonlabs-io/babylon-toolkit/pull/2336#pullrequestreview-5065768088) both name gbarkhatov.
- [ ] Two distinct qualified reviewers approve the final Pre-PegIn output checks in #2398. Preserve [gbarkhatov's final approval](https://github.com/babylonlabs-io/babylon-toolkit/pull/2398#pullrequestreview-5139029000) and [passing CI](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34167533604): 34 Chromium tests and 1,966 SDK Node tests.
[...]

PR #2210, opened 2026-08-10
feat(packages): make Bitcoin wallets optional in the vault

jonybur:
## Status

Closed without merge. Do not use this pull request as the current plan.

It was the first full prototype for #2228. It is superseded by #2234, #2267, #2307, #2332, and the replacement work planned from #2336.

## Why it was closed

- 206 files changed.
- The diff mixed wallet behavior, package changes, tests, and value-moving code.
- Review and release risk were too high for one pull request.
- Later review found defects that its green checks did not detect.

## Useful result

The prototype proved that an Ethereum-first Vault is possible. Its work was split into the five steps in #2228.

For current scope, status, and blockers, read #2228.

PR #2234, opened 2026-08-12
fix(wallet-connector): set up Bitcoin curve support inside the connector

### LINKED ISSUE #2229: [eth-first] Make Bitcoin wallet setup self-contained
**Epic:** #2228
**Step:** 1 of 5
**Status:** Complete

## Outcome

The wallet connector now loads the Bitcoin curve support that it needs. Host applications no longer need to load it first.

## Delivered

- #2234 added cold-start taproot and connection tests.
- Taproot address and signing paths initialize the curve before use.
- The new critical paths were added to the repository safety lists.

## Evidence

- Current wallet connector suite: 173 tests passed.
- The connector source has one direct `initEccLib` call.
- The wider merge-rule gap is tracked by #2359.

jonybur:
## Result

Merged and accepted. This completes #2229, step 1 of #2228.

## Change

- The wallet connector initializes Bitcoin curve support before taproot work.
- Host applications no longer need to initialize it first.
- Cold-start taproot and wallet connection tests protect the behavior.
- New sensitive paths were added to the repository safety lists.

## Verification at merge

- Wallet connector tests passed.
- Ledger signer tests passed.
- Build, type check, lint, and critical-path checks passed.
- Review confirmed that the new tests fail when curve setup is removed.

Closes #2229.


.github/workflows/claude-md-drift.yml
```diff
@@ -40,6 +40,14 @@ jobs:
             "services/vault/src/services/vault/vaultActivationService.ts"
             "packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/vaultSplit.ts"
             "packages/babylon-ts-sdk/src/tbv/core/utils/signing.ts"
+            "packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts"
```
gbarkhatov, line 57: **Should fix before merge — this job will cry wolf every Monday.**

The analyze step does `[ -e "$p" ] || echo "$p" >> missing.txt` and sets `has_signal=true` on any non-empty result. I confirmed all eight added paths are absent on this branch, so from next Monday until #2228 lands the tracker issue gets a weekly section headed *"Paths referenced in CLAUDE.md / CODEOWNERS / CI no longer exist. Update or restore"* — inverted (they don't exist *yet*), and eight false entries a week is how people learn to skim past the section that reports real drift.

The other two guards are genuinely fine with absent paths: `critical-path-check.yml` only pattern-matches changed files, and CODEOWNERS entries for non-existent paths are inert. So this is a one-file fix — hold these eight back from the drift job until the code lands, or give the job a second `pending=(...)` array that is excluded from `missing.txt` and reported under its own heading.

CLAUDE.md
```diff
@@ -93,6 +93,25 @@ These paths handle irreversible value movement. An AI-generated mistake here is
 - Uses `useTweakedSigner: false` and `autoFinalized: false` for taproot script-path spends. Wallet support is inconsistent; silent failures produce invalid signatures.
 - **Rule:** Validate every signature produced with these flags against the expected sighash before treating the PSBT as signed. Do not rely on the wallet returning success.
 
+### 8. Dependency-free reimplementations of Bitcoin primitives
```
gbarkhatov: **Should fix before merge — SECURITY.md is the fifth inventory, and its own rule says update it.**

The PR body's justification ("references these paths in prose rather than a list") doesn't survive reading the file. Three concrete spots go stale the moment this merges:

- `SECURITY.md:107` — "the **seven** critical-path groups enumerated in CLAUDE.md". This makes it eight.
- `SECURITY.md:232-235` — "The critical-path inventory is hand-maintained in five places: this file, CLAUDE.md, CODEOWNERS, `critical-path-check.yml`, and `claude-md-drift.yml`. **Update all five together** when a path moves or is added." This PR updates four.
- `SECURITY.md:737` — the risk register row for *"A critical path moves but one hand-maintained inventory keeps the stale path"* asserts "All current paths are aligned", which becomes false for eight paths. A control that documents itself as green while drifting is worse than no control.

The count and the register row are one-line edits; the group itself can stay prose.

CLAUDE.md
```diff
@@ -93,6 +93,25 @@ These paths handle irreversible value movement. An AI-generated mistake here is
...
+
+These paths are registered ahead of the code arriving (see #2228 / #2229). Separating the Ethereum-only
```
gbarkhatov: Nit — formatting. This section is hard-wrapped at ~105 columns while §1-§7 run unwrapped (172-757 chars per line). Reflowing to one line per paragraph and per bullet matches the rest of the file and keeps future diffs here line-scoped instead of reflowing the whole block.

packages/babylon-wallet-connector/src/core/utils/__tests__/wallet.taproot.test.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Re-import the module under test with no prior curve registration in this
+ * module registry, mirroring a cold page load.
+ */
+async function loadWalletUtils() {
+  vi.resetModules();
+  return import("@/core/utils/wallet");
+}
+
+describe("taproot derivation without host-side curve setup", () => {
+  beforeEach(() => {
```
gbarkhatov, line 59: **"Fresh module registry" holds per file, not per test — and the two resets are redundant.**

`vi.resetModules()` only walks Vitest's own module map. `bitcoinjs-lib` is externalized (no `deps.inline`), so `_ECCLIB_CACHE` survives; what actually resets is this package's `initialised` flag. Once test 1 registers the curve, every later test in the file inherits it.

I checked this rather than assuming: removing `initBTCCurve()` from `generateP2TRAddressFromXpub` **only** (leaving the one in `getTaprootAddress`) still passes all 5. So the xpub entry point is currently unguarded by this file.

Clearing bitcoinjs's cache explicitly makes each test genuinely cold — `initEccLib` supports being passed a falsy value for exactly this. Verified both directions locally: passes with the fix intact, and fails the xpub test with the xpub-side init removed.

```ts
import { initEccLib } from "bitcoinjs-lib";

beforeEach(() => {
  initEccLib(undefined as never); // clears bitcoinjs's _ECCLIB_CACHE
  vi.resetModules();
});
```

With that in place, `loadWalletUtils()`'s own `vi.resetModules()` is dead — drop it and keep the reset in one place.

packages/babylon-wallet-connector/src/core/utils/__tests__/wallet.taproot.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    // BIP-86 test-vector account xpub, first receive address.
+    const xpub =
+      "xpub6BgBgsespWvERF3LHQu6CnqdvfEvtMcQjYrcRzx53QJjSxarj2afYWcLteoGVky7D3UKDP9QyrLprQ3VCECoY49yfdDEHGCtMMj92pReUsQ";
+    const { address, publicKeyHex, scriptPubKeyHex } = generateP2TRAddressFromXpub(
+      xpub,
+      "m/0/0",
+      toNetwork(Network.MAINNET),
+    );
+
+    expect(address).toMatch(/^bc1p/);
```
gbarkhatov: **These assert shape when the published vector is right there.** A wrong derivation path, a flipped parity byte, or a mis-sliced x-only key all produce a well-formed `bc1p…` / `5120…` string and pass.

I ran the derivation: `HDKey.fromExtendedKey(xpub).derive("m/0/0")` yields `03cc8a4b…c115` — the `03` prefix plus this file's own `XONLY` constant, and the address is exactly `MAINNET_ADDRESS`. So two of the three expectations are already declared 60 lines above:

```ts
expect(address).toBe(MAINNET_ADDRESS);
expect(publicKeyHex).toBe(`03${XONLY}`);
expect(scriptPubKeyHex).toBe("5120a60869f0dbcf1dc659c9cecbaf8050135ea9e8cdc487053f1dc6880949dc684c");
```

Note the scriptPubKey is **not** `"5120" + XONLY` — it carries the BIP-341 *tweaked output* key, which is the whole point of the curve dependency this test exists to pin, so it's worth asserting literally rather than deriving it in the test.

Same objection this PR's own new CLAUDE.md §8 rule makes: a shape check pins the format, not the function.

packages/babylon-wallet-connector/src/core/utils/initBTCCurve.ts
```diff
@@ -1,7 +1,24 @@
...
+
+/**
+ * Register bitcoinjs-lib's elliptic-curve implementation.
+ *
+ * Taproot needs it: BIP-341 key tweaking is what `payments.p2tr` performs, and
+ * decoding a taproot scriptPubKey back to an address needs the same curve. The
+ * connector must never rely on the host application having done this — a host
+ * that loads its Bitcoin dependencies lazily (so an Ethereum-only session stays
+ * light) would otherwise break every taproot wallet connection. Every entry
+ * point in this package that builds or decodes a taproot address calls this
```
gbarkhatov: Nit — "every entry point in this package that builds or decodes a taproot address calls this first" isn't quite true yet. `keystone/provider.ts:372` calls `signedPsbt.finalizeAllInputs()`, which for a key-spend input reaches bitcoinjs `_finalizeTaprootInput` → `payments.p2tr({ output, signature })`, and p2tr's validation block (`opts.validate` defaults true) calls `getEccLib().isXOnlyPoint(pubkey)`. No `initBTCCurve()` on that path.

It works today only because Keystone's `connect()` runs `generateP2TRAddressFromXpub` first — the exact implicit ordering dependency this PR is removing everywhere else. Either add the call at the top of the Keystone sign path, or soften the docstring to "builds or decodes a taproot address or finalizes a taproot input" so the claim stays honest.

packages/babylon-wallet-connector/src/core/utils/initBTCCurve.ts
```diff
@@ -1,7 +1,24 @@
...
+ * first, and this is the only place that calls `initEccLib`.
+ *
+ * Idempotent, so callers can invoke it unconditionally.
+ */
 export const initBTCCurve = () => {
+  if (initialised) return;
```
gbarkhatov: Nit, low priority — the flag is redundant and its only behavioural effect points the wrong way.

bitcoinjs already no-ops on repeat: `else if (eccLib !== _ECCLIB_CACHE.eccLib)` in `ecc_lib.js`, so `verifyEcc` runs once regardless of how often this is called. What the local flag adds is a stale-state failure mode: if anything ever clears or swaps the global (`initEccLib(undefined)` — supported, and it's what I'd suggest the new test use), this returns early forever and derivation starts throwing again.

Dropping `initialised` keeps the docstring's idempotence promise intact, since bitcoinjs provides it.

gbarkhatov (dismissed):
Approving — the fix is at the right layer and the entry-point coverage claim holds for every path that matters today. A few things to pick up, none of which block the mechanism itself.

## What I verified

- **Coverage.** Grepped the package for everything that needs a live curve: `payments.p2tr` in `wallet.ts:48` and `:149`, `btcAddress.fromOutputScript` in `unisat/provider.ts:448`. All three now init up front. `p2wpkh` and `toXOnly` need no curve; OKX/OneKey/UniSat finalize inside the extension.
- **The test is real.** Stripping both `initBTCCurve()` calls from `wallet.ts` fails all five with `No ECC Library provided. You must call initEccLib()` — the exact error the connect handler was converting into "Connection Failed".
- **Placement rationale checks out.** `"test": "vitest run"`, config includes `src/**/*.test.{ts,tsx}` and excludes `tests/**`, so `nx affected --target=test` picks this up and a file under `tests/unit` would not have run.
- Lint on the touched files: 0 errors (7 pre-existing `any` warnings in unisat). Full `vitest run` locally: 102 passed, plus `dmkSession.test.ts` failing to resolve `rxjs` — untouched by this PR, local environment.

## What's right

Fixed at the layer that owns the dependency. Deleting both catch-and-retry blocks is the real win — they matched on the substrings `"ECC"` and `"has no matching address"`, and the unisat one let every other decode failure escape unwrapped; now every failure comes back as a `WalletError`. `generateP2TRAddressFromXpub` also loses its duplicated derivation and four `!` assertions. And dropping `initEccLib` from the ledger-vault test quietly turns its existing BIP-86 case into a second independent cold guard.

## Two things not inline-able

**SECURITY.md is the fifth inventory and it is now stale.** Details on the CLAUDE.md thread — the "prose rather than a list" justification does not survive reading the file, which names a count, an explicit five-file sync rule, and an "all aligned" assertion in the risk register.

**Static import of the curve library works against the initiative's goal.** `initBTCCurve.ts` imports `@bitcoin-js/tiny-secp256k1-asmjs` (2.7 MB unpacked) at module scope, and `BTCWalletProvider.tsx:15` pulls `toXOnlyPublicKeyHex` from `@/core/utils/wallet` even though it lives in `@/core/utils/publicKey` and is only re-exported there. So an Ethereum-only session that mounts the connector still drags bitcoinjs + the asmjs curve into the graph — the thing #2210 is trying to avoid. Not a regression, the edge predates this PR, but this is the PR that decides "the package owns curve setup", so it is the moment to decide. The cheap half is a one-line import change in `BTCWalletProvider`; the other half — a lazy `await import()` — turns `getTaprootAddress` and friends async and ripples into the connect flow, so it belongs in #2228 if anywhere. Worth a line in the PR body either way.

## Follow-up, out of scope

`packages/babylon-ts-sdk/src/tbv/core/primitives/utils/bitcoin.ts:301-316` still enforces "the consuming application must call `initEccLib(ecc)`". The SDK needs the same treatment before #2210 can drop the call from `main.tsx`, or the vault trades a connector failure for an SDK one.

.github/workflows/claude-md-drift.yml
```diff
@@ -40,6 +40,14 @@ jobs:
...
+            "packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts"
+            "packages/babylon-ts-sdk/src/tbv/core/wasm/"
+            "packages/babylon-tbv-rust-wasm/src/wasm-loader.ts"
+            "packages/babylon-tbv-rust-wasm/src/wasm-loader-node.ts"
+            "packages/babylon-tbv-rust-wasm/src/raw.ts"
+            "packages/babylon-tbv-rust-wasm/src/raw-node.ts"
+            "services/vault/src/utils/btc/scriptPubKeyAddress.ts"
```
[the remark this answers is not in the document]
jonybur: Fixed. The eight pre-registered paths now live in a separate `pending=()` array that the existence check never reads, so they can no longer land in `missing.txt` or be reported as "no longer exist. Update or restore".

They are not simply dropped either: a `pending` path that *does* exist is written to a new `promote.txt` and reported under its own heading, "Pending paths that have landed", telling the maintainer to move it up into `paths`. They also stay in the 7-day commit scan, so a merged PR touching one is still surfaced.

Verified by extracting the `Analyze drift` step and running it under `bash -e` against the real repo: before, `missing.txt` had all eight entries; now it is empty, `promote.txt` is empty, and the step exits 0. Creating one of the eight makes it appear under the promote heading, still exit 0.

.github/workflows/claude-md-drift.yml
jonybur: Done, using your second option - a separate `pending=(...)` array excluded from the existence check.

Rather than folding the promote signal back into `missing.txt` (which would put a path that *now exists* under a heading saying it does not), landed entries go to their own `promote.txt` and render under "Pending paths that have landed", with instructions to move the entry up into `paths`. `pending` stays in the 7-day commit scan, so a merged PR touching one of these files is still reported.

Also confirmed your read of the other two guards, which is why neither needed touching: `critical-path-check.yml` only prefix-matches changed files and does no existence check, and no workflow validates CODEOWNERS globs.

I ran the extracted step under `bash -e` (GitHub's `bash -e {0}`) in four states - nothing pending, one pending path present, all eight present, and a real `paths` entry deleted. Exit 0 in all four, with `missing.txt` empty in the first three.

Two things I picked up while testing that file, both now included:

- `git log --pretty=format:` emits no trailing newline, so the `while read` loop was silently discarding the **oldest** commit for every path. A path touched by exactly one commit in the window was reported zero times - which is precisely the shape of a pending path landing. Switched to `--pretty=tformat:`. Before the fix the last run missed #2201 entirely, which touches three section-1/3/4 paths; it now shows up.
- The "Keep in sync with..." comment sat above `paths` alone and was misleading once the array was split - somebody following it literally would paste the eight back and reinstate the false alarm. It now says `paths` and `pending` are one inventory and it is their union that must stay in sync.

CLAUDE.md
jonybur: You were right, and it was worse than the thread describes, because main moved underneath this branch while it sat in review.

#2236 inserted "7. Ledger vault signer package" and pushed wallet signing to 8, and main's SECURITY.md already says "eight". So the seven -> eight edit would have been a no-op against main, and the merge would have landed **cleanly** with two `### 8.` headings in CLAUDE.md, two `# 8.` blocks in CODEOWNERS, and a count that was wrong in the other direction. Only `critical-path-check.yml` conflicts, so nothing would have forced anyone to look.

I merged main in and renumbered this group to 9. The three spots you named:

- `SECURITY.md:107` - now "the **nine** critical-path groups".
- The five-places paragraph - the rule itself was accurate, this PR was just violating it, so I left the sentence alone and appended the missing nuance: a group may be registered before its files exist, section 9 is registered ahead of #2228, and those paths stay in the drift workflow's `pending` list until they land.
- The risk-register row - now "Sections 1-8 are aligned; section 9 is pre-registered ahead of #2228; the existence check does not gate merges". The replacement fits inside the existing 111-char column budget, so the table padding is unchanged and the diff stays a one-line hunk.

Two more spots you did not cite but which the same change made stale: the row's evidence column omitted SECURITY.md itself, the very inventory this thread is about, and the CI section still described the drift job without the `pending` carve-out. Both updated.

Checked afterwards that all five inventories agree: 24 entries each, `paths` (16) + `pending` (8) equals the `critical-path-check.yml` list equals CODEOWNERS, and exactly the eight `pending` entries are the ones absent on disk.

Leaving the deeper structural gaps for #2228 - a SECURITY.md domain subsection for group 9, and a "Critical invariants" entry. The invariants list is by construction about things that currently hold, and group 9's rule governs code that does not exist yet.

CLAUDE.md
jonybur: Reflowed - the paragraph and the `**Rule:**` bullet are each one line now, matching sections 1-8.

Confirmed the reflow is prose-only: collapsing whitespace runs on both versions gives byte-identical strings, and the em dashes are intact.

jonybur: Reflowed - the paragraph and the `**Rule:**` bullet are each one line now, matching sections 1-8.

Confirmed the reflow is prose-only: collapsing whitespace runs on both versions gives byte-identical strings, and the em dashes are intact.

packages/babylon-wallet-connector/src/core/utils/__tests__/wallet.taproot.test.ts
jonybur: Confirmed and fixed - your mutation reproduced exactly: deleting `initBTCCurve()` from `generateP2TRAddressFromXpub` only still passed 5/5, and fails now.

Two deltas from your snippet, both deliberate:

**Dropped the `as never`.** bitcoinjs-lib declares `initEccLib(eccLib: TinySecp256k1Interface | undefined, ...)`, with an explicit `if (!eccLib)` clearing branch, so plain `initEccLib(undefined)` typechecks - `tsc -p tsconfig.lib.json --noEmit` exits 0, and substituting a number does produce TS2345, so the file really is in that program. The cast would only hide a future signature change.

**Dropped `vi.resetModules()` entirely**, not just the copy inside `loadWalletUtils`. Keeping it cancels out the clear for the one production regression the other thread's change exists to prevent: `resetModules` re-evaluates `initBTCCurve.ts`, which resets any module-level "already initialised" memo. I checked by restoring the memoised `initBTCCurve.ts` from this branch's first commit - with `resetModules` in `beforeEach`, all 5 still pass; without it, 4 fail. Clearing the curve cache is the cold-start simulation that matters, and the first test is a genuine cold module load anyway.

The mutation matrix now, all four of which passed before this round:

| mutation | result |
|---|---|
| no `initBTCCurve()` in `getTaprootAddress` | 4 fail |
| no `initBTCCurve()` in `generateP2TRAddressFromXpub` | 1 fail |
| reintroduce the `initialised` memo flag | 4 fail |
| derivation path `m/0/0` -> `m/0/1` | 1 fail |

The file header still claimed the tests "never call `initEccLib`", which stopped being true the moment this landed; it now says they clear the registration and never register a working one.

packages/babylon-wallet-connector/src/core/utils/__tests__/wallet.taproot.test.ts
jonybur: Fixed - all three are literal assertions now.

I recomputed the vector independently of bitcoinjs (via `@scure/bip32` + `@noble/curves` + `@scure/base`) rather than reading it back out of the function under test, and your three values are exact:

```
address:         bc1p5cyxnuxmeuwuvkwfem96lqzszd02n6xdcjrs20cac6yqjjwudpxqkedrcr
publicKeyHex:    03cc8a4bc64d897bddc5fbc2f670f7a8ba0b386779106cf1223c6fc5d7cd6fc115
scriptPubKeyHex: 5120a60869f0dbcf1dc659c9cecbaf8050135ea9e8cdc487053f1dc6880949dc684c
```

So the xpub at `m/0/0` is `03` + this file's own `XONLY`, and the address is the same `MAINNET_ADDRESS` the first test pins - the two tests reach the same BIP-86 key by two different code paths, which is worth having.

Your point about the scriptPubKey carrying the tweaked output key rather than `XONLY` is the reason it is a named constant with a comment saying so, instead of being built as `"5120" + XONLY` in the assertion.

Teeth check: changing the path to `m/0/1` passed under the old shape assertions and fails now.

packages/babylon-wallet-connector/src/core/utils/initBTCCurve.ts
jonybur: Good catch - confirmed at runtime, not just by reading. I built a P2TR key-spend PSBT, cleared the curve cache, and replayed the Keystone path: `Psbt.fromHex` is fine, then `finalizeAllInputs()` throws `No ECC Library provided`, via `_finalizeTaprootInput` -> `payments.p2tr({output, signature})` -> `getEccLib().isXOnlyPoint`.

Did both halves rather than either:

- `initBTCCurve()` now runs immediately before `signedPsbt.finalizeAllInputs()`. It sits at the finalize site rather than the top of `sign()`, matching the UniSat precedent, so `autoFinalized: false` callers - who never touch the curve - do not pay for it. Zero bundle cost: the provider already imports `@/core/utils/wallet`, which pulls `initBTCCurve` in.
- The docstring now says "every entry point in this package that needs the curve - building or decoding a taproot address, and finalizing a taproot key-spend input".

I swept for anything else before softening the wording, so the claim is exhaustive rather than merely vaguer: `payments.p2tr` at `wallet.ts:48` and `:149`, `fromOutputScript` at `unisat/provider.ts:448`, and this one are the only sites in `src/` reaching `getEccLib()`. `ledger` and `ledger-v2` finalize `@scure/btc-signer` transactions with their own curve; appkit only serialises; `tapleafHash`, `toXOnly`, `pubkeyInScript` and `p2wpkh` need no curve; and the script-path finaliser never goes through `p2tr`, which is why "key-spend" is the right qualifier.

Worth noting it covers two public entries, not one: `signMessageBIP322` also calls the private `sign()` with no options, so it finalises too, and `BIP322.encodeWitness` depends on that succeeding.

One correction to the framing: it is latent rather than live. Beyond the connect ordering you identified, the vault still calls `initEccLib` eagerly in `main.tsx`, so no user can reach this today - it bites when #2210/#2232 makes that lazy. Hardening ahead of that change rather than a shipped bug.

Also changed "the only place that calls `initEccLib`" to "the only place in `src/` that registers a curve implementation" - the new test calls `initEccLib(undefined)` to clear it, so the absolute wording stopped being true in the same commit. That does technically break a verbatim acceptance criterion on #2229 (`grep -rn "initEccLib" src` returns only `initBTCCurve.ts`); flagging it rather than quietly leaving it red.

packages/babylon-wallet-connector/src/core/utils/initBTCCurve.ts
jonybur: Agreed, flag removed.

Confirmed your reading of `ecc_lib.js` by instrumenting `isXOnlyPoint`: three `initEccLib(ecc)` calls with the same object run `verifyEcc` exactly once, gated on the `eccLib !== _ECCLIB_CACHE.eccLib` reference check. And I reproduced the failure mode - with the flag, `initEccLib(undefined)` followed by `initBTCCurve()` leaves the curve unregistered forever; without it, it recovers.

Cost of dropping it is nil in the normal case: the warm short-circuit measures ~5.5 ns/call, versus ~10.8 ms for a genuine first registration, which happens once either way.

One correction for the record: the parenthetical about the new test making this a live concern does not hold. I ran it both ways - the test passes with the flag present, because the same `beforeEach` used to call `vi.resetModules()`, which re-evaluates the module and resets `initialised`. So the two threads were independent. That is also why I ended up removing `resetModules` from the test (see the other thread) - keeping it meant this exact regression could be reintroduced with the suite still green. With the flag gone and `resetModules` gone, reinstating the memo now fails 4 of the 5 tests.

The docstring keeps the idempotence promise but now says where it comes from: `initEccLib` verifies and stores only when the library differs, so repeat calls cost nothing and a cleared or swapped global is re-registered rather than skipped.

Only theoretical cost I found: `vite.config.ts` externalises `bitcoinjs-lib` but not `@bitcoin-js/tiny-secp256k1-asmjs`, so the ESM and CJS bundles each embed their own `ecc` namespace. A realm loading both would alternate two distinct objects and re-verify (~9 ms) each time, which the flag used to cap. The vault loads only the ESM build, so it is unreachable today; adding the asmjs package to `rollupOptions.external` would close it properly if it ever matters.

jonybur:
Thanks for the review - all eight threads are addressed, with replies inline on each.

**Merged main in first.** #2236 landed a "7. Ledger vault signer package" critical-path group and renumbered wallet signing to 8, and main's SECURITY.md already reads "eight". Left alone, this branch would have merged *cleanly* into two `### 8.` headings, two `# 8.` CODEOWNERS blocks, and a group count that was wrong in the other direction - only `critical-path-check.yml` conflicts, so nothing would have forced a human to look. This group is now 9, and all five inventories agree: 24 entries each, with exactly the eight pre-registered ones absent on disk.

Summary of the code changes:

| Thread | Change |
|---|---|
| drift job cries wolf (x2) | Pre-registered paths moved to a `pending` array, exempt from the existence check, reported under their own heading once they land |
| SECURITY.md stale | Count, pre-registration note and risk-register row updated, plus two spots the thread did not name |
| §8 hard-wrapped | Reflowed; prose verified byte-identical |
| test not cold per-case | Clears bitcoinjs's curve cache per case; `vi.resetModules()` removed entirely, see below |
| shape assertions | Literal BIP-86 address, pubkey and scriptPubKey |
| Keystone finalize | `initBTCCurve()` added there, and the docstring widened - both, not either |
| `initialised` memo flag | Removed |

**Two places I did not follow the suggestion exactly**, both explained in the threads: `initEccLib(undefined)` needs no `as never` (the published type admits `undefined`), and I removed `vi.resetModules()` rather than just the duplicate, because keeping it let the memo-flag regression slip through green. The mutation matrix now kills all four mutations that previously passed.

**On the static-import point.** Took the one-line `BTCWalletProvider` import fix, but the bundle claim does not hold and the PR body now says so: cutting that edge leaves the curve fully reachable through `WalletDialog -> useWalletConnectors -> wallet.ts` and through the BTC wallet registry, where `core/wallets/btc/index.ts` eagerly imports every provider and `unisat/provider.ts` imports `initBTCCurve` directly. The connector also builds as one chunk, so `ETHWalletProvider` and `BTCWalletProvider` load identical bytes. An Ethereum-only session gets lighter only if the registry is code-split, which is larger than the lazy `await import()` you scoped to #2228. Taking the import change as layering hygiene; the now-orphaned `toXOnlyPublicKeyHex` re-export in `wallet.ts` is deleted with it.

**On the ts-sdk follow-up.** Confirmed and out of scope here. `bitcoin.ts:300-322` still puts `initEccLib(ecc)` on the consumer, nothing in `ts-sdk/src` registers a curve (only the vitest setup does - the same masking harness this PR removed from Storybook), and the guard is incomplete: `pegin.ts:498` and `noPayout.ts:148` are unguarded, and `bip322Verify.ts:122` swallows a missing curve into a `false` that surfaces as "BIP-322 signature verification failed". One nuance on sequencing: because bitcoinjs-lib is external and shared, this PR's `initBTCCurve()` on the connect path registers the curve globally, so a taproot-only vault would not break the moment #2210/#2232 drops `main.tsx` - it becomes an undeclared ordering dependency that breaks when ts-sdk is used without the connector, i.e. the Ethereum-only entry points in #2231. Suggest it becomes a scope item there rather than a new issue.

**One drive-by**, flagged because it is in a loop this PR widened: `git log --pretty=format:` emits no trailing newline, so the drift scan was silently dropping the oldest commit per path - and a path touched by exactly one commit was reported zero times, which is exactly how a pending path lands. Switched to `tformat:`; the last run had been missing #2201 entirely.

Verification: connector 56 tests / 4 files, ledger-vault-signer 139 / 9, lint 0 errors (87 pre-existing warnings), `tsc --noEmit` clean, build clean, and the drift analyze step run end to end under `bash -e` in four states.

packages/babylon-wallet-connector/src/core/utils/__tests__/wallet.taproot.test.ts
jonybur: Confirmed and fixed - your mutation reproduced exactly: deleting `initBTCCurve()` from `generateP2TRAddressFromXpub` only still passed 5/5, and fails now.

Two deltas from your snippet, both deliberate:

**Dropped the `as never`.** bitcoinjs-lib declares `initEccLib(eccLib: TinySecp256k1Interface | undefined, ...)`, with an explicit `if (!eccLib)` clearing branch, so plain `initEccLib(undefined)` typechecks - `tsc -p tsconfig.lib.json --noEmit` exits 0, and substituting a number does produce TS2345, so the file really is in that program. The cast would only hide a future signature change.

**Dropped `vi.resetModules()` entirely**, not just the copy inside `loadWalletUtils`. Keeping it cancels out the clear for the one production regression the other thread's change exists to prevent: `resetModules` re-evaluates `initBTCCurve.ts`, which resets any module-level "already initialised" memo. I checked by restoring the memoised `initBTCCurve.ts` from this branch's first commit - with `resetModules` in `beforeEach`, all 5 still pass; without it, 4 fail. Clearing the curve cache is the cold-start simulation that matters, and the first test is a genuine cold module load anyway.

The mutation matrix now, all four of which passed before this round:

| mutation | result |
|---|---|
| no `initBTCCurve()` in `getTaprootAddress` | 4 fail |
| no `initBTCCurve()` in `generateP2TRAddressFromXpub` | 1 fail |
| reintroduce the `initialised` memo flag | 4 fail |
| derivation path `m/0/0` -> `m/0/1` | 1 fail |

The file header still claimed the tests "never call `initEccLib`", which stopped being true the moment this landed; it now says they clear the registration and never register a working one.

jonybur:
Pushed the connect-path test — the last acceptance criterion of #2229 that was still open, and the one CodeRabbit's linked-issue check flagged.

**Why it was missing, and what changed.** Driving the BTC connect handler needs a rendered hook, so the file has to run under jsdom — and jsdom installs its own `Uint8Array`, which leaves a Node `Buffer` failing `instanceof` and makes bitcoinjs-lib's `initEccLib` self-check reject the curve library outright. That is why `wallet.taproot.test.ts` is pinned to the node environment, and it is why a connect test looked impossible. It is a realm artifact, not a product problem: a browser has one realm. The file restores Node's constructor so the environment matches production, rather than conceding to it. Without that line no jsdom test in this package can register a curve at all.

Three cases, all cold (`initEccLib(undefined)` before each):

- a taproot wallet stays selected — `removeWallet` and `disconnect` are never called
- no error screen: `displayError` is not called and the flow advances to `displayChains`
- a mismatched public key surfaces as "Public Key Mismatch", not "Connection Failed" — which is what distinguishes validation *running and returning false* from validation *throwing*

Mutation-checked: stripping either `initBTCCurve()` call from `wallet.ts` fails all three with the production symptom, `"Connection Failed"` / `"No ECC Library provided. You must call initEccLib()"`.

**Also in this commit**, from the CodeRabbit nits: the combined testnet/signet case is split in two, and both now assert the published BIP-86 address rather than a `tb1p` prefix (a wrong derivation path produces a well-formed `tb1p…` too). The xpub, path and parity-byte vector inputs are named.

**Two nits not taken**, both for consistency rather than disagreement with the rule: extracting the fallback error message in `unisat/provider.ts` and the message plus `"hex"` literal in `wallet.ts`. That unisat file has 23 inline `message:` strings and no named message constants; `wallet.ts` has four sibling `WalletError` throws with inline messages and uses inline `"hex"` in two other places. Naming one of each makes the file less uniform, not more readable.

Verification: connector suite 60 passed / 5 files (was 56 / 4), lint clean on the touched files, `tsc -p tsconfig.lib.json --noEmit` clean.

gbarkhatov (approved):

jrwbabylonlab (dismissed):

.github/workflows/claude-md-drift.yml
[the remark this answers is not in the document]
mrjorticio-hash: 33f2fee9a6023f6b88f84d612a7a7928d184a22e

gbarkhatov (dismissed):

gbarkhatov (approved):

Issue #2223, opened 2026-08-12
Ledger: extract @babylonlabs-io/ledger-vault-signer package (audit boundary)


jrwbabylonlab:
Extract the Ledger vault protocol code into a dedicated `packages/babylon-ledger-vault-signer` (`@babylonlabs-io/ledger-vault-signer`) so the signer has an isolated blast radius and a clean audit scope. Decided 2026-08-12 (4-agent analysis + Codex LGTM); sequenced as the FIRST commit of the signer milestone — after #2214 merges, before any #2219 protocol code is written, so every line of the signer is born, reviewed, and audited in its final home.

What moves (verbatim, `git mv`): `intentTlv`, `vaultCommands`, `derivation`, `envelope`, `deviceCaps`, `types`, `dmkSession`, `dmkApduSender` + their tests. What stays in wallet-connector: `provider.ts` + `index.ts` as the thin `IBTCProvider` adapter (the only modules that import connector internals — verified).

Scope:
- Scaffold from the ts-sdk template (subpath exports); headless — importable from node, which unlocks Speculos-driven tests from the E2E CLI and ts-sdk integration tests.
- `"private": true` initially — `nx.json` releases everything under `packages/*`, so run the release dry-run before flipping public.
- Package-owned typed errors (e.g. `LedgerUserRefusedError`) replace the `WalletError` throws in `dmkApduSender` — the adapter maps them to `ERROR_CODES`; this also deletes the import-cycle workaround.
- LICENSE/NOTICE home for the ~600 vendored Apache-2.0 `ledger-bitcoin` lines (#2219).
- Add the package to the critical-path review gate.

~0.5 day. Blocks #2219 (its code lands in this package; only provider wiring lands in wallet-connector).

---
**Scope addition (verified by 2-agent analysis):** harden `deviceCaps.ts` against firmware drift.

[...]

Issue #2221, opened 2026-08-12
Ledger: BIP-322 proof of possession via SIGN_PSBT (tx_version 0)


jrwbabylonlab:
`signMessage` for the vault provider is the BIP-322 proof of possession, and it does NOT use the base app's SIGN_MESSAGE: the vault app implements PoP as a SIGN_PSBT with `tx_version == 0` as the discriminator — 1-in/1-out, BIP-86 keypath input with a witness-utxo value of 0, and the message carried in the global proprietary key `FC 06 "bvault" 00`.

Message grammar (btc-vault `pop.rs` is authoritative): `<eth_address>:<chain_id>:pegin:<registry_address>` — lowercase hex with literal `0x` prefixes.

**Dependencies:** depends on #2219 (the signer core — this is one more SIGN_PSBT). Sequenced before #2222 (Pre-PegIn): the deposit flow runs PoP at step 2, Pre-PegIn signing at step 4.

---
**Update 2026-08-18 (#2219 B3):** the signing seam this issue builds on is in place — the staged
API (`prepareSignPsbt` + `signPreparedVaultPsbt`) is exported, `SignVaultPsbtResult.yields`
carries the raw collected yields (the PoP detached-witness seam), and the provider's key-path
pre-rejection currently names this issue. `signMessage` stays `notWired` until this wires it.

---
**Update 2026-08-19 — scope grows: the wallet-policy machinery lands HERE, not in #2222.**
Verified on firmware PR #7 `fix/audit-findings @ 4decf822` (base app `e400d8d8`): the base app signs
key-path inputs only under a wallet policy (`sign_psbt.c:142-148`), and a tx_version-0 PSBT with no policy
now returns `SW_OK` with **zero yields** (`sign_custom_inputs.c:107` returns without signing) — a silent
failure. Every firmware PoP test signs under `tr(@0/**)`. So this issue also owns:
- `GET_MASTER_FINGERPRINT` (INS **0x05** — `docs/apdu.md` says 0x10; code wins) and the account xpub at
  `m/86'/coin'/0'` (pass the device's base58 string through verbatim);
- the default policy `DefaultWalletPolicy("tr(@0/**)", …)`, policy-mode `prepareSignPsbt`
  (wallet_id = policy id, hmac stays 32×00) and `addKnownWalletPolicy` seeding;
- `signMessage('bip322-simple')` must assert exactly one 64-byte yield; package `0x01 0x40 ‖ sig`
[...]

Issue #2219, opened 2026-08-12
Ledger: own the SIGN_PSBT host protocol (merkleized PSBT + client-command loop)


jrwbabylonlab:
Ledger confirmed (Slack, 2026-08-12) that `@ledgerhq/device-signer-kit-bitcoin` will not drive the vault app and will not be extended — "the Vault needs its own signer kit built on DMK". The host side of SIGN_PSBT is ours to build. This issue is the signer core: everything `signPsbt` needs to produce valid signatures for the no-policy vault flows (PegIn, Payout, NoPayout, Refund, Claim, Assert, WronglyChallenged).

Scope (all protocol code lands in `@babylonlabs-io/ledger-vault-signer` — #2223 extracts the package first; only the provider wiring lands in wallet-connector):

- **Transport seam**: non-throwing `sendRaw(apdu) → { sw, data }` with an injectable transport, so the same loop runs over DMK/WebHID in the browser and a Speculos TCP client in node tests.
- **Protocol primitives**: Bitcoin varints (little-endian — the vault TLV scalars are big-endian, do not reuse `u32BE`), RFC-6962-style Merkle (`k` = largest power of 2 strictly smaller than `n`), MerkleMap.
- **PSBT map model + v0→v2 normalization** at the key/value level; unknown fields pass through byte-identical (they carry the taproot data the vault validator reads).
- **SIGN_PSBT commitment header** (195 bytes). `wallet_id` and `wallet_hmac` are 32 zero bytes for every vault-graph tx — the all-zero `wallet_id` IS the routing switch into the vault validators. P2 = 1.
- **Client-command interpreter**: YIELD / GET_PREIMAGE / GET_MERKLE_LEAF_PROOF / GET_MERKLE_LEAF_INDEX / GET_MORE_ELEMENTS, 255-byte chunking; "not found" is normal control flow, never an error.
- **The 0xE000 / CLA F8 CONTINUE loop.** Hard constraint from firmware: the device kills the app after ~5s of host silence per round trip (`INTERRUPTION_TIMEOUT_TICKS`) — no network fetches or lazy WASM loads inside the loop.
- **Signature collection + merge** into the original v0 PSBT as `tapScriptSig`, asserted against a host-side expected-leaf table (tapleaf hashes computed from each input's `tapLeafScript`); never finalize. CRITICAL PATH (CLAUDE.md §3) — two reviewers.
[...]

PR #2236, opened 2026-08-12
Extract ledger vault signer

jrwbabylonlab:
Extracts the Ledger vault device protocol into a dedicated package and folds in the #2214 review follow-ups. Two commits: the review fixes, then the extraction.

**New package `@babylonlabs-io/ledger-vault-signer` (#2223)** — Ledger won't extend their Bitcoin signer kit, so the vault's SIGN_PSBT host protocol (#2219) is ours to build. Moving the device-protocol modules into their own package gives that value-bearing work an isolated, auditable home before it lands. The 8 modules + tests move verbatim (git renames, blame preserved); `provider.ts` and `index.ts` stay in wallet-connector as the thin adapter, mapping the package's typed device errors (`LedgerUserRefusedError` / `LedgerDeviceLockedError` / `LedgerDeviceError`) onto `WalletError` at one seam. DMK + rxjs deps move to the signer; the private package bundles into wallet-connector's published dist. Registered in all five critical-path inventory files.

**Review fixes (#2214):**
- `vaultCoreVersion` gate is now exact-match — the intent TLV carries no version field, so the device can't reject a mismatched graph at load; it approves on-screen and fails at PSBT time
- Intent idempotence fingerprint hashes the encoded wire bytes, so a roster reorder can't fake a "different intent"
- u64 fields fail with the seam's typed `DepositTermsRejectedError` instead of a raw encoder `Error` mid-ceremony
- `getDmk()` memoizes its build promise (no duplicate DMK / leaked WebHID listeners under concurrent connects)
- status-word typing, BIP-32 path bound, `#2109`→`#2219` refs, plus comment/test corrections

**Adversarial hardening** (fresh review pass):
- non-Error DMK objects now map to `WalletError` instead of surfacing as `[object Object]`
- roster/VP keys equal to the depositor's own key are pre-empted as a shaped error (firmware rejects them only after the whole ceremony)
- `connectWallet` shares one in-flight session across concurrent calls
- a per-connection generation guard prevents a disconnect/reconnect from committing stale ceremony state onto a new connection

Signing stays unimplemented — `signPsbt`/`signPsbts`/`signMessage` throw typed capability errors (`TODO(#2219)`/`TODO(#2221)`); the wallet remains behind `NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET`, off by default.


jonybur (approved):
Approving. The extraction is a clean move — 8 modules and their tests rename verbatim into the new package, and `provider.ts` stays the one seam that maps the signer's typed errors onto the connector's WalletError taxonomy. The critical-path rule in CLAUDE.md for the new package covers exactly what needs covering: firmware-cited encodings, envelope-gate-before-I/O, no payload logging.

Verified against the head:

- **Concurrency lifecycle** — `connectPromise` shares one attempt across overlapping `connectWallet` calls, `disconnectToken` is captured before any await in `doConnect` and re-checked after the liveness probe and the DMK connect, and a session that resolves after a racing disconnect is torn down rather than installed behind a disconnected wallet. `teardownSession` invalidates host state synchronously before awaiting the transport close, so a late APDU on the old sender cannot commit state onto the new connection — the `connectionGeneration` guard is the belt to that suspender in `approveDepositTerms`/`deriveContextHash`.
- **`getDmk()` promise memo** — the catch handler is identity-guarded (`dmkPromise === build`) so a failed build clears the memo without clobbering a successor. Fresh-build test after `closeDmk()` still passes.
- **Envelope gate** — `vaultCoreVersion` is now exact-match (was floor), matching the reality that the intent TLV carries no version field. u64 range checks (`requireU64`) run before the semantic floors on all four per-vault fields plus `prepeginMaxFee`, so out-of-range values fail as `DepositTermsRejectedError` before the encoder ever sees them. Depositor-key clash pre-empted with `canonicalPubkey` comparison against the cached pubkey.
- **Idempotence fingerprint** — hashes the encoded wire bytes (`encodeIntentScalars` + groups + key batches), so a roster reorder produces a byte-identical fingerprint (encoder sorts before wire) and cannot fake a "different intent." Value-sensitive on `peginAmount`/keeper swap, confirmed by the new tests.
- **Error mapping** — `withWalletErrorMapping` covers the three typed errors from the signer, and the DMK-shaped `{_tag, originalError}` object gets mapped rather than surfacing as `[object Object]`.
- **Package boundary** — dependencies moved cleanly (DMK + rxjs move to the signer, `@babylonlabs-io/ledger-vault-signer` added to wallet-connector), no Buffer usage in the standalone package (`toHex` replaces it in `derivation.ts`), tsconfig references updated, CODEOWNERS + drift/critical-path workflows registered.

Signing paths remain typed capability errors gated by `NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET`, so the surface exposed today is exactly what the tests cover.

PR #2238, opened 2026-08-13
feat(ledger-vault-signer): vendor ledger-bitcoin byte primitives for …

jrwbabylonlab:
First slice of the SIGN_PSBT host protocol (#2219): vendor the byte-level primitives from Ledger's `ledger-bitcoin` client, and make the signer a self-contained publishable package.

- **Vendored** `varint` / `Merkle` / `MerkleMap` into `ledger-vault-signer` — the deprecated `ledger-bitcoin` ships these, and its successor needs `bitcoinjs-lib` 7 against our `6.1.7` pin, so we vendor the v6-compatible source (Apache-2.0) with provenance headers and a re-diff procedure.
- **Golden-vector gated** against Ledger's own Python client (the version the vault firmware's device tests use) — varint, Merkle `n=0..9`, and MerkleMap commitments from real PSBT fixtures.
- **Publish the signer** (the `core-ui` pattern): it now ships its own LICENSE + THIRD-PARTY-NOTICES; wallet-connector externalizes it and the duplicate notice is removed.


jeremy-babylonlabs (approved):

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/merkleMap.golden.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  it.each(maps)("commits $label to the oracle bytes", ({ entries, commitment }) => {
+    const byKey = new Map(entries.map(([k, val]) => [k, val]));
+    // Build in the device's byte-lexicographic key order.
+    const keys = commitment.sorted_key_order.map((k) => Buffer.from(k, "hex"));
+    const values = commitment.sorted_key_order.map((k) => Buffer.from(byKey.get(k)!, "hex"));
+
+    const map = new MerkleMap(keys, values);
+    expect(map.keysTree.getRoot().toString("hex")).toBe(commitment.keys_root);
+    expect(map.valuesTree.getRoot().toString("hex")).toBe(commitment.values_root);
+    expect(map.commitment().toString("hex")).toBe(commitment.commitment_hex);
```
gbarkhatov, line 60: **The fixtures ship `inputs_maps_root` / `outputs_maps_root` and nothing asserts them — and they are derivable today, no `psbtv2` required.**

The vectors carry a lot more than this test consumes: `psbt_hex`, `psbt_v2_hex`, `inputs_maps_root`, `outputs_maps_root`, `sign_psbt_cdata_hex`, `apdu`. The two maps-roots are reachable with just the vendored `Merkle` + `MerkleMap`. I confirmed for all three fixtures that

```
Merkle(input_commitments.map(c => hashLeaf(Buffer.from(c.commitment_hex, "hex")))).getRoot() === inputs_maps_root
```

and the same for outputs.

This is worth adding because it is a **distinct failure mode** from what the per-map assertions cover. The outer tree hashes `hashLeaf(commitment)`, while the inner trees hash `hashLeaf(key)` / `hashLeaf(value)`. Wrong leaf treatment at the outer level — or maps committed in the wrong order — passes every assertion in this file today. It is also the composition the device actually Merkle-verifies in the SIGN_PSBT header, so pinning it here makes `merkleMap`'s gate real now rather than "once `psbtv2` lands".

While you are in there: the full 195-byte `sign_psbt_cdata_hex` also reconstructs today from `global_commitment ‖ varint(n_inputs) ‖ inputs_maps_root ‖ varint(n_outputs) ‖ outputs_maps_root` plus a 64-byte zero tail — verified on all three fixtures. That would exercise `createVarint` in its real production context instead of in isolation.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/merkleMap.golden.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  ];
+}
+
+const maps = [peginVector, payoutVector, graphVector].flatMap((v) => mapsOf(v as unknown as SignPsbtVector));
+
+describe("MerkleMap commitment golden vectors", () => {
+  it.each(maps)("commits $label to the oracle bytes", ({ entries, commitment }) => {
+    const byKey = new Map(entries.map(([k, val]) => [k, val]));
+    // Build in the device's byte-lexicographic key order.
+    const keys = commitment.sorted_key_order.map((k) => Buffer.from(k, "hex"));
```
gbarkhatov, line 54: Keys are built from the oracle's `sorted_key_order`, so these vectors prove the commitment **layout** but never that our key ordering agrees with the device's. Deriving that order is genuinely `psbtv2`'s job later — but it is cheap to pin now, and right now the file reads as if ordering were covered:

```ts
// The device commits keys in byte-lexicographic order; pin that our sort agrees
// with the oracle's, not just that the commitment matches its ordering.
expect(keys).toEqual([...keys].sort(Buffer.compare));
expect(commitment.sorted_key_order.length).toBe(entries.length);
```

(`MerkleMap`'s own check uses hex-string comparison, which is order-equivalent to `Buffer.compare` for lowercase hex — so the two agree, but nothing says so.)

packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
```diff
@@ -0,0 +1,222 @@
...
+## Ledger Bitcoin JS client
+
+- Project: LedgerHQ/app-bitcoin (formerly app-bitcoin-new)
+- Source: https://github.com/LedgerHQ/app-bitcoin
+- Version: ledger-bitcoin@0.3.0 (commit 0a9e9e141f3340d29e7c6181177d4e5e9483a9f7)
+- Files: bundled from packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/
+- License: Apache License 2.0
+- Modifications: see the per-file provenance headers in the vendored source
+  (explicit Buffer import, defensive strict-null guards, formatting; the
+  object-level PSBT conversion was removed).
```
gbarkhatov, line 18: This describes a modification that is not in the artifact — no PSBT file is vendored yet, so nothing here dropped an object-level PSBT conversion. This file carries redistribution weight under Apache-2.0 §4(b), so it should describe what actually ships. Suggest trimming to the three files present and re-adding the `psbtv2` note in the increment that vendors it.

packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
```diff
@@ -0,0 +1,222 @@
+# Third-party notices
+
+This file covers third-party source vendored into this repository's package tree
+and bundled into published artifacts. It is not an exhaustive inventory of every
```
gbarkhatov, line 4: Minor accuracy: "and bundled into published artifacts" is not true yet. Nothing under `src/vendor/` is reachable from `src/index.ts`, so rollup drops it and this PR's `dist/` is unchanged. Fine for an incremental slice — the wording just runs ahead of the code by one PR.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
```diff
@@ -0,0 +1,56 @@
...
+(`buffertools.ts`, `psbtv2.ts`, `merkelizedPsbt.ts`) and the client-command
+interpreter (`clientCommands.ts`, `policy.ts`) follow in later increments.
+
+## Golden-vector gate
+
+`__tests__/*.golden.test.ts` check the vendored modules against vectors generated
+by **Ledger's own Python client** (`ledger-bitcoin==0.4.0`, the version the vault
+firmware's device tests drive SIGN_PSBT through). `varint` and `merkle` have
+direct golden vectors (`varint.json`, `merkle_mth.json`); `merkleMap`'s
+commitment is exercised through the PSBT map vectors once `psbtv2` lands. Any
```
gbarkhatov, line 44: This contradicts the tree as merged: `merkleMap.golden.test.ts` already does exactly this today, against committed per-map commitments from the three PSBT fixtures. It is not waiting on `psbtv2`.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
```diff
@@ -0,0 +1,56 @@
...
+  exact modifications made (§4(b)).
+
+## Modifications (applied to every file)
+
+- Explicit `import { Buffer } from "buffer"` — no implicit Node global, since this
+  package ships to the browser and the connector's vite polyfill cannot resolve a
+  bare `Buffer` from the standalone dist.
+- Strict-null / strict-index fixes for this repo's strict `tsconfig`.
+- Prettier/quote formatting to the package style.
+- `psbtv2.ts` (lands in a later increment): the object-level `fromBitcoinJS` is
```
gbarkhatov, line 28: This bullet sits under a heading that reads "Modifications (applied to **every file**)" while describing a file that does not exist yet. Worth splitting into a "planned, not yet vendored" subsection so the section keeps meaning exactly what it says — this is the same list `THIRD-PARTY-NOTICES.md` points readers at for the §4(b) modification record.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
```diff
@@ -0,0 +1,56 @@
...
+  dropped (it throws on taproot inputs — exactly our case; the map-level
+  `normalizeToV2` is the path we use).
+
+Everything else — protocol logic, byte layout, function surface — is verbatim.
+
+Vendored so far: `varint.ts`, `merkle.ts`, `merkleMap.ts`. The PSBT layer
```
gbarkhatov, line 34: Worth stating outright here that these three modules are currently unreferenced by any production module — reachable only from tests, so rollup tree-shakes them and `dist/` is unchanged.

The repo's ZERO DEAD CODE policy is absolute enough that a reviewer arriving cold will stop on this, and it is also what makes the new `bitcoinjs-lib` / `buffer` runtime deps look unused. One sentence saying "unreferenced until the next slice wires them into the SIGN_PSBT client" saves that round trip.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
```diff
@@ -0,0 +1,56 @@
...
+behavioural drift from upstream fails these. The Python client is the primary
+oracle; the deprecated npm package is a secondary check.
+
+## Re-diff procedure (on an upstream update)
+
+1. `git -C <clone of LedgerHQ/app-bitcoin> fetch && git diff \
+0a9e9e141f3340d29e7c6181177d4e5e9483a9f7..<new> -- bitcoin_client_js/src/lib/`
+   — review the per-file deltas.
+2. Apply the deltas onto these copies, re-run the style pass, and update each
+   header's `Version` / `Modifications` / upstream sha256.
```
gbarkhatov, line 54: This step is the right procedure, but it is hand-run and nothing keeps the headers honest between updates. I verified them by hand today — fetched `LedgerHQ/app-bitcoin@0a9e9e141f33` and all three `Upstream sha256` values match exactly, with every source delta accounted for by the documented modifications — but that check evaporates the moment someone edits a vendored file without touching its header.

A CI step that curls the pinned upstream and asserts each header's sha256 would make this self-enforcing, mirroring how `VAULT_WASM_COMMIT` pins the wasm facade. Given this package is on the critical-path list, that seems proportionate.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/vectors/merkle_mth.json
```diff
@@ -0,0 +1,402 @@
+{
+ "description": "ledger_bitcoin.merkle.MerkleTree \u2014 MTH({})=32x00; MTH({d0})=SHA256(0x00||d0); MTH(D[n])=SHA256(0x01||MTH(D[0:k])||MTH(D[k:n])), k = largest power of 2 strictly smaller than n. proofs[i] = sibling hashes bottom-up (prove_leaf).",
```
gbarkhatov, line 2: The oracle is named here and in the README (`ledger-bitcoin==0.4.0`, Python), but no generator script ships with it — so these vectors cannot be regenerated, extended to new `n`, or re-derived if the oracle version moves. That is out of step with the `golden_vectors_pinned` convention in `btc-vault`, and it matters more here because the next increments will want vectors over the same fixtures.

Committing the generator (even a 30-line script with the pinned `ledger-bitcoin` version in a comment) would make the gate reproducible.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/merkle.ts
```diff
@@ -0,0 +1,141 @@
...
+ * Upstream:        https://github.com/LedgerHQ/app-bitcoin (formerly app-bitcoin-new)
+ * File:            bitcoin_client_js/src/lib/merkle.ts
+ * Version:         ledger-bitcoin@0.3.0 (npm gitHead 0a9e9e141f3340d29e7c6181177d4e5e9483a9f7)
+ * Upstream sha256: 733ecf485c0b98ad3732234945506bb6dda0d9814e27f8802f1def78cdad52d0
+ * Vendored:        2026-08-13
+ * License:         Apache-2.0 — see ./LICENSE (verbatim upstream copy)
+ * Modifications:   explicit `import { Buffer } from "buffer"`; defensive
+ *                  strict-null guards on getLeafHash/getProof and a
+ *                  `leaves[0] as Buffer` narrowing at the n===1 branch
+ *                  (behaviour-preserving; valid indices unchanged); formatting.
```
gbarkhatov, line 13: Small precision point on the modification record: "behaviour-preserving" holds for valid indices but understates what changed for invalid ones.

Upstream `getProof` guarded with `if (index >= this.leaves.length)`, so `getProof(-1)` fell through and threw a raw `TypeError` from `proveNode(undefined)`; upstream `getLeafHash` had no guard at all. Both now throw a clean `Error("Index out of bounds")`. That is strictly an improvement — but since this header is the §4(b) record of what was changed, it is worth being exact: "invalid indices now throw a typed Error instead of a TypeError; valid indices unchanged".

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/varint.ts
```diff
@@ -0,0 +1,127 @@
...
+  const first = data[offset];
+  if (first === undefined) {
+    throw Error("Buffer too small");
+  }
+
+  if (first < 0xfd) {
+    return [BigInt(first), 1];
+  } else {
+    let size: number;
+    if (first === 0xfd) size = 2;
```
gbarkhatov, line 99: Worth recording as known upstream behaviour rather than changing: `parseVarint` accepts non-canonical encodings — `fd0100` decodes to `1`, which `createVarint` would emit as `01`. It matches upstream and the Python oracle, so it is correct for this package's purpose and should stay.

Flagging it because it becomes a live hazard one increment from now: once the PSBT layer parses externally-supplied bytes and re-serializes them into a commitment, a non-canonical input silently changes the bytes the device Merkle-verifies. Better to note it in the README's provenance section now than to rediscover it when a commitment mismatches.

packages/babylon-ledger-vault-signer/package.json
```diff
@@ -26,18 +26,20 @@
     "test:watch": "vitest"
   },
   "files": [
-    "dist"
+    "dist",
+    "THIRD-PARTY-NOTICES.md"
```
gbarkhatov, line 30: Worth noting what this line is now load-bearing for: `vite.config.ts:37` sets `esbuild.legalComments: "none"`, which strips the vendored files' provenance headers from `dist/`. So once the vendor modules become reachable from `src/index.ts`, the Apache-2.0 §4(b) notices you carefully wrote exist in the published package **only** via this `THIRD-PARTY-NOTICES.md` entry.

That is defensible, but it is a silent coupling. Either flip `legalComments` to `"eof"` for this package now while it is a one-word change, or add a comment in `vite.config.ts` explaining that attribution deliberately lives in the notices file. Doing it now avoids finding out at the increment where the vendored code actually ships.

(`LICENSE` itself is fine — npm always includes it regardless of `files`.)

packages/babylon-ledger-vault-signer/package.json
```diff
@@ -26,18 +26,20 @@
...
     "type": "git",
     "url": "https://github.com/babylonlabs-io/babylon-toolkit.git"
   },
-  "license": "ISC",
   "description": "Host-side client for the Ledger Babylon Vault app: DMK session lifecycle, raw APDU framing, key reads, and the vault intent ceremony",
   "dependencies": {
     "@ledgerhq/device-management-kit": "1.7.1",
     "@ledgerhq/device-transport-kit-web-hid": "1.2.4",
     "@scure/bip32": "1.4.0",
+    "bitcoinjs-lib": "6.1.7",
```
gbarkhatov, line 41: Nit, and probably "keep it" — but worth being a decision rather than a default. `bitcoinjs-lib` becomes a hard runtime dependency of this package for exactly one symbol: `crypto.sha256`, as the default `hasher` in `merkle.ts`. That hasher is already an injectable constructor argument, and `@scure/bip32` (existing dep) pulls in `@noble/hashes`, so `sha256` is reachable without it.

Keeping `bitcoinjs-lib` keeps the vendored source closer to upstream, which has real value for the re-diff procedure — that is a good reason. Just noting that the later PSBT increments will need `bitcoinjs-lib` for more than a hash, so it is worth confirming the v6 pin is the long-term answer before more code depends on it.

packages/babylon-ledger-vault-signer/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@babylonlabs-io/ledger-vault-signer",
-  "version": "0.1.0",
-  "private": true,
+  "version": "0.0.0-semantic-release",
```
gbarkhatov, line 3: This is the package's first publish, which makes two things worth checking before merge to `main`:

1. **Version resolution.** There is no `ledger-vault-signer/*` git tag yet, so nx's `fallbackCurrentVersionResolver: "disk"` resolves the current version from this sentinel. `core-ui` carries the same on-disk sentinel but resolves from its existing tags, so it is not evidence that a fresh package works — worth a `nx release --dry-run` to see what version actually comes out.
2. **Publish ordering and access.** `babylon-wallet-connector/vite.config.ts` now externalizes this package, so published wallet-connector no longer inlines it and hard-requires it from npm. If the signer's first publish fails (scope access) or lands after wallet-connector, the published wallet-connector is broken for external consumers even though CI is green.

Neither is a code problem — just the kind of thing that is much cheaper to check before the release run than after.

packages/babylon-ledger-vault-signer/LICENSE
```diff
@@ -0,0 +1,169 @@
...
+8.  in violation of applicable laws, rules or regulations in your relevant jurisdiction;
+9.  that circumvents any sanctions or export controls targeting you or the country or territory where
+    you are located; or
+10. in any activity that transmits, exchanges, or is otherwise supported by the direct or indirect
+    proceeds of criminal or fraudulent activity.
+
+Any production use of the Licensed Work by you confirms your agreement to the foregoing limitations,
+and your understanding and agreement that they are limitations and not restrictions.
+
+Change Date:          2027-03-15 (March 15th, 2027)
```
gbarkhatov, line 88: Legal's call, not a code review point — but flagging since it looks inherited rather than chosen. This file is the root/`core-ui` BUSL text with only the `Licensed Work:` line changed, so it carries `(c) 2024` and `Change Date: 2027-03-15` onto a package first published in 2026.

BUSL's own default is "the fourth anniversary of first publication", so a 2026 package on a fixed 2027-03-15 date gets roughly seven months of BUSL rather than four years. That may well be deliberate — aligning the whole toolkit on one change date is a coherent choice — but it is worth being explicit rather than a side effect of copying the file.

gbarkhatov (approved):
## Approve — provenance and vectors both verified independently

Reviewed the vendored primitives closely given the `critical-path` label. The two things that actually matter here both check out, and I verified them rather than taking the headers on trust:

**Provenance is genuine.** I fetched `LedgerHQ/app-bitcoin@0a9e9e141f3340d29e7c6181177d4e5e9483a9f7` and all three claimed `Upstream sha256` values match exactly:

| file | sha256 | |
|---|---|---|
| `merkle.ts` | `733ecf485c0b98ad…` | matches |
| `merkleMap.ts` | `39c5058d58c0dcbb…` | matches |
| `varint.ts` | `c5636971b4a96743…` | matches |

Diffing upstream against the vendored copies, every single delta is prettier/quote reflow, the added `import { Buffer } from "buffer"`, or one of the documented strict-null narrowings. No logic changed, nothing smuggled in. The pinned-commit + per-file-sha256 + re-diff-procedure approach is the right shape for vendored crypto and I would like to see it reused.

**Golden vectors are correct.** I reimplemented the spec from scratch — RFC-6962-variant MTH with `k` = largest power of two strictly `< n`, Bitcoin varint, `varint(n) ‖ keysRoot ‖ valuesRoot` — and checked every committed vector against it:

- 10 Merkle trees, `n = 0..9`, roots **and** all bottom-up proofs — 0 mismatches
- 15 varints through 2^64−1, with boundary coverage at `0xfd` / `0xffff` / `0xffffffff` — 0 mismatches
- 15 MerkleMap commitments across the 3 PSBT fixtures — 0 mismatches; `sorted_key_order` is genuinely byte-sorted and is a permutation of each map's key set in every case

Test design is good too: the independent `refDecodeVarint` rather than round-tripping through `parseVarint`, and the `crypto.sha256("")` assertion in `beforeAll` to catch a mis-wired hasher before the vectors do, are both the right instincts.

### Comments below

All follow-ups, none blocking. The one I would most like to see in this PR is the **unasserted `inputs_maps_root` / `outputs_maps_root`** — they are already in the fixtures and derivable today with no `psbtv2`, and they cover a failure mode the current per-map assertions cannot reach. The rest is documentation accuracy on files that carry redistribution weight, plus two pre-merge checks on the first publish.

### Two things that do not anchor to a line

- **The PR description is stale.** It says wallet-connector's "duplicate notice is removed", but there is no such change in the diff and no `THIRD-PARTY-NOTICES.md` outside this package — the signer's is the only one in the repo.
- **Nothing vendored reaches `dist/` yet.** None of the three modules is exported from `src/index.ts` or imported by production code, so rollup tree-shakes them all and this PR's published bundle is unchanged. That is fine for increment 1, but it is the single fact behind three separate observations below (the notices file's "bundled into published artifacts", the new `bitcoinjs-lib`/`buffer` deps looking unused, and the ZERO DEAD CODE tension). One line in the PR body would head all three off.

Not run: `pnpm lint` / `pnpm test` on this branch — CI covers both.

jrwbabylonlab:
fix all comments in next PR

packages/babylon-ledger-vault-signer/package.json
```diff
@@ -1,7 +1,6 @@
...
   "type": "module",
```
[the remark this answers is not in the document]
jrwbabylonlab: Already BUSL-1.1 — a valid SPDX id, better for tooling than SEE LICENSE IN LICENSE.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/merkleMap.golden.test.ts
jrwbabylonlab: Fixed in follow-up — added the outer maps-roots + cdata golden vectors.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/merkleMap.golden.test.ts
jrwbabylonlab: Skipped — redundant; the MerkleMap constructor already enforces strictly-increasing keys.

packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
jrwbabylonlab: Fixed in follow-up.

packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
jrwbabylonlab: Fixed in follow-up — and excluded the vendored dir from dist, so nothing vendored ships now.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
jrwbabylonlab: Fixed in follow-up.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
jrwbabylonlab: Fixed in follow-up — moved to a Planned subsection.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
jrwbabylonlab: Fixed in follow-up.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
jrwbabylonlab: Deferring to #2219 with the rest of the PSBT layer.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/vectors/merkle_mth.json
jrwbabylonlab: Deferring to #2219 — one generator for the full vector set.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/merkle.ts
jrwbabylonlab: Fixed in follow-up.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/varint.ts
jrwbabylonlab: Fixed in follow-up — noted in the README provenance.

packages/babylon-ledger-vault-signer/package.json
jrwbabylonlab: Fixed in follow-up — documented, and the vendored dir is now excluded from the published dist too.

packages/babylon-ledger-vault-signer/package.json
jrwbabylonlab: Keeping it — stays closer to upstream, and the PSBT layer needs it for more than a hash.

packages/babylon-ledger-vault-signer/package.json
jrwbabylonlab: Access is sorted (0.0.1 published); 0.2.0 still needs a manual publish to unbreak wallet-connector@1.66.1.

packages/babylon-ledger-vault-signer/LICENSE
jrwbabylonlab: Left as-is — toolkit-wide change date; flagging to whoever owns licensing.

Issue #2220, opened 2026-08-12
ts-sdk: thread the intent ceremony through signAndBroadcast and resume-broadcast


jrwbabylonlab:
## Status

- **Part 1 — DONE (#2240, merged):** the derive → approve ceremony threaded into `PeginManager.signAndBroadcast` (new optional `depositTerms`) and the vault app's `broadcastPrePeginTransaction`, plus the fresh-deposit call site. Shared `ensurePrePeginTermsApproval` helper; no-op for non-intent wallets; txid-bound.
- **Part 2 — IMPLEMENTED (branch `feat/vault-resume-terms-rebuild`, PR pending):** the resume-broadcast path — `DepositTerms` rebuilt from chain + WASM only (never browser storage), then the same ceremony runs before signing.

## Part 2 — as shipped

- ts-sdk pure core `deposit-terms/rebuildDepositTermsCore.ts`: Gate 0 self-verify (funded tx hashes to the on-chain `prePeginTxHash`), auth-anchor OP_RETURN completeness (`vout === siblingCount`, single-hit), WASM sizing recompute with positivity + reserve-plausibility bounds, Gate 1 per-sibling HTLC value + scriptPubKey byte-match, projection via `buildDepositTerms`.
- Vault orchestrator `services/vault/src/services/vault/rebuildDepositTerms.ts`: stamped-version chain reads (mirrors `prepareSigningContext`), sibling discovery by on-chain `prePeginTxHash` with contiguity + **sibling-batch homogeneity** asserts (stamps/provider/application/commission must equal the target's — Gate 1 cannot see `timelockPegin`/`timelockAssert`/commission), required connected-depositor ownership check.
- `resolveFundedTxFee.ts`: mempool-only prevout resolution (`prepeginMaxFee` = Σin−Σout; the same record feeds the broadcast PSBT, so terms-fee ≡ signed-fee); `useVaultActions.ts` `handleBroadcast` branches on `supportsDepositApproval` — software wallets byte-identical to before.
- Commission ceiling: interim proxy via `resolveMaxAcceptableCommissionBps.ts` → #2252.
- Cross-path golden `PeginManager.rebuildDepositTerms.test.ts`: rebuilt terms deep-equal `preparePegin`'s production-built terms (v1 + v2).

## Resolution of the #2240 deferred review items

[...]

PR #2240, opened 2026-08-13
feat(ts-sdk): thread the intent ceremony through the Pre-PegIn broadcast

jrwbabylonlab:
Part 1 of #2220: give the Pre-PegIn broadcast an intent-approval seam so Ledger (intent) wallets can sign it.

- **New `ensurePrePeginTermsApproval` helper** (ts-sdk) — the shared derive→approve ceremony both broadcast paths run immediately before signing. No-op for software wallets; for intent wallets it verifies the terms match the tx, derives the vault root over the funding outpoints, then approves.
- **`PeginManager.signAndBroadcast`** — new optional `depositTerms`; ceremony runs before `signPsbt`.
- **Vault app** — `broadcastPrePeginTransaction` gains the same ceremony (device-rejection errors pass through unwrapped so the UI shows intent-rejection copy); the fresh-deposit call site forwards the wallet's approval capability + terms.
- Exported `capMaxAcceptableCommissionBps` (needed by the resume rebuild).

**Follow-up (Part 2, tracked in #2220):** the *resume* path — reconstruct terms from chain/indexer (`rebuildBroadcastDepositTerms`, with byte-verified sibling completeness) and rewire `useVaultActions` handleBroadcast.

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/__tests__/prePeginApproval.test.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/__tests__/prePeginApproval.test.ts
@@ -1,0 +1,211 @@
+/**
+ * Tests for the Pre-PegIn approval ceremony helper. The device layer is
+ * abstracted behind the structural PrePeginApprovalWallet, so these pin the
+ * orchestration: the capability gate, the txid match, and the strict
+ * derive-then-approve order for approval wallets.
+ */
+
+import { Transaction } from "bitcoinjs-lib";
+import { Buffer } from "buffer";
+import { describe, expect, it, vi } from "vitest";
+
+import {
+  hexToUint8Array,
+  uint8ArrayToHex,
+} from "../../primitives/utils/bitcoin";
+import { buildVaultContext, VAULT_APP_NAME } from "../../vault-secrets";
+import type { DepositTerms } from "../depositTerms";
+import {
+  ensurePrePeginTermsApproval,
+  type PrePeginApprovalWallet,
+} from "../prePeginApproval";
+
+/** A funded Pre-PegIn tx with two inputs, so the funding-outpoint derivation is exercised. */
+function makeFundedTx(): { hex: string; txid: string; inputTxids: string[] } {
+  const tx = new Transaction();
+  tx.version = 2;
+  // Non-palindromic on purpose: reverse(txid) !== txid, so a dropped .reverse()
+  // anywhere on the derive path changes the funding outpoints — and the context
+  // assertion below would then fail instead of passing on a byte-order bug.
+  const inputTxids = ["0011".repeat(16), "2233".repeat(16)];
+  for (const txid of inputTxids) {
+    // input.hash is internal byte order — the reverse of the display txid.
+    tx.addInput(Buffer.from(txid, "hex").reverse(), 0);
+  }
+  tx.addOutput(
+    Buffer.from(
+      "512000000000000000000000000000000000000000000000000000000000000000",
+      "hex",
+    ),
+    1000,
+  );
+  return { hex: tx.toHex(), txid: tx.getId(), inputTxids };
+}
+
+const DEPOSITOR_PUBKEY = "ab".repeat(32);
+
+function makeTerms(txid: string): DepositTerms {
+  return {
+    vaultCoreVersion: 2,
+    protocolFeeRate: 2n,
+    timelockPegin: 684,
+    timelockAssert: 684,
+    timelockRefund: 2016,
+    prepeginTxid: txid,
+    prepeginMaxFee: 1500n,
+    vaultKeeperBtcPubkeys: ["cc".repeat(32)],
+    universalChallengerBtcPubkeys: ["dd".repeat(32)],
+    vaults: [
+      {
+        htlcVout: 0,
+        vaultProviderBtcPubkey: "ff".repeat(32),
+        peginAmount: 1_000_000n,
+        commissionFee: 10_000n,
+        depositorClaimValue: 20_000n,
+        peginMaxFee: 800n,
+      },
+    ],
+  };
+}
+
+describe("ensurePrePeginTermsApproval", () => {
+  it("is a no-op for a wallet without approveDepositTerms", async () => {
+    const { hex } = makeFundedTx();
+    const wallet: PrePeginApprovalWallet = { deriveContextHash: vi.fn() };
+
+    await expect(
+      ensurePrePeginTermsApproval({
+        wallet,
+        depositTerms: undefined,
+        fundedPrePeginTxHex: hex,
+        depositorBtcPubkey: DEPOSITOR_PUBKEY,
+      }),
+    ).resolves.toBeUndefined();
+    expect(wallet.deriveContextHash).not.toHaveBeenCalled();
+  });
+
+  it("validates supplied terms against the txid even for a non-approval wallet", async () => {
+    const { hex } = makeFundedTx();
+    const wallet: PrePeginApprovalWallet = {};
+
+    await expect(
+      ensurePrePeginTermsApproval({
+        wallet,
+        depositTerms: makeTerms("00".repeat(32)),
+        fundedPrePeginTxHex: hex,
+        depositorBtcPubkey: DEPOSITOR_PUBKEY,
+      }),
+    ).rejects.toThrow(/do not match the transaction/);
+  });
+
+  it("throws when an approval wallet is given no terms", async () => {
+    const { hex } = makeFundedTx();
+    const wallet: PrePeginApprovalWallet = {
+      deriveContextHash: vi.fn(async () => "ab".repeat(32)),
+      approveDepositTerms: vi.fn(),
+    };
+
+    await expect(
+      ensurePrePeginTermsApproval({
+        wallet,
+        depositTerms: undefined,
+        fundedPrePeginTxHex: hex,
+        depositorBtcPubkey: DEPOSITOR_PUBKEY,
+      }),
+    ).rejects.toThrow(/requires approved deposit terms/);
+    expect(wallet.approveDepositTerms).not.toHaveBeenCalled();
+  });
+
+  it("throws for an approval wallet that cannot derive", async () => {
+    const { hex, txid } = makeFundedTx();
+    const wallet: PrePeginApprovalWallet = { approveDepositTerms: vi.fn() };
+
+    await expect(
+      ensurePrePeginTermsApproval({
+        wallet,
+        depositTerms: makeTerms(txid),
+        fundedPrePeginTxHex: hex,
+        depositorBtcPubkey: DEPOSITOR_PUBKEY,
+      }),
+    ).rejects.toThrow(/must also implement deriveContextHash/);
+  });
+
+  it("rejects terms for a different transaction before any device call", async () => {
+    const { hex } = makeFundedTx();
+    const wallet: PrePeginApprovalWallet = {
+      deriveContextHash: vi.fn(async () => "ab".repeat(32)),
+      approveDepositTerms: vi.fn(),
+    };
+
+    await expect(
+      ensurePrePeginTermsApproval({
+        wallet,
+        depositTerms: makeTerms("99".repeat(32)),
+        fundedPrePeginTxHex: hex,
+        depositorBtcPubkey: DEPOSITOR_PUBKEY,
+      }),
+    ).rejects.toThrow(/do not match the transaction/);
+    expect(wallet.deriveContextHash).not.toHaveBeenCalled();
+    expect(wallet.approveDepositTerms).not.toHaveBeenCalled();
+  });
+
+  it("derives then approves, in that order, for a matching approval wallet", async () => {
+    const { hex, txid, inputTxids } = makeFundedTx();
+    const order: string[] = [];
+    const wallet: PrePeginApprovalWallet = {
+      deriveContextHash: vi.fn(async () => {
+        order.push("derive");
+        return "ab".repeat(32);
+      }),
+      approveDepositTerms: vi.fn(async () => {
+        order.push("approve");
+      }),
+    };
+    const terms = makeTerms(txid);
+
+    await ensurePrePeginTermsApproval({
+      wallet,
+      depositTerms: terms,
+      fundedPrePeginTxHex: hex,
+      depositorBtcPubkey: DEPOSITOR_PUBKEY,
+    });
+
+    expect(order).toEqual(["derive", "approve"]);
+    expect(wallet.approveDepositTerms).toHaveBeenCalledWith(terms);
+
+    // Pin the exact derive input: the funding outpoints must be display-order
+    // txids (proving the input.hash reversal), keyed to the depositor pubkey.
+    // A dropped reversal or wrong pubkey would change this context hex.
+    const expectedContext = uint8ArrayToHex(
+      buildVaultContext({
+        depositorBtcPubkey: hexToUint8Array(DEPOSITOR_PUBKEY),
+        fundingOutpoints: inputTxids.map((t) => ({
+          txid: hexToUint8Array(t),
+          vout: 0,
+        })),
+      }),
+    );
+    expect(wallet.deriveContextHash).toHaveBeenCalledWith(
+      VAULT_APP_NAME,
+      expectedContext,
+    );
+  });
+
+  it("accepts a 0x-prefixed prepeginTxid in the terms", async () => {
+    const { hex, txid } = makeFundedTx();
+    const wallet: PrePeginApprovalWallet = {
+      deriveContextHash: vi.fn(async () => "ab".repeat(32)),
+      approveDepositTerms: vi.fn(),
+    };
+
+    await expect(
+      ensurePrePeginTermsApproval({
+        wallet,
+        depositTerms: makeTerms(`0x${txid}`),
+        fundedPrePeginTxHex: hex,
+        depositorBtcPubkey: DEPOSITOR_PUBKEY,
+      }),
+    ).resolves.toBeUndefined();
+    expect(wallet.approveDepositTerms).toHaveBeenCalledTimes(1);
+  });
+});
--- a/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/index.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/index.ts
@@ -10,3 +10,4 @@
 export * from "./buildDepositTerms";
 export * from "./depositTerms";
 export * from "./depositTermsErrors";
+export * from "./prePeginApproval";
--- a/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
@@ -1,0 +1,134 @@
+/**
+ * The Pre-PegIn approval ceremony for intent-based signing wallets.
+ *
+ * A shared helper because there are two Pre-PegIn broadcast paths — the SDK's
+ * {@link PeginManager.signAndBroadcast} and the vault app's own
+ * `broadcastPrePeginTransaction` (a near-duplicate). Both must run the exact
+ * same derive → approve sequence immediately before `signPsbt`, or the two
+ * copies drift on a signing-critical path.
+ *
+ * The device signs a Pre-PegIn only from `INTENT_LOADED`, and only if the
+ * approved intent's `prepegin_txid` matches this tx and its fee is within the
+ * approved `prepegin_max_fee`. So an approval-capable wallet must be shown the
+ * terms (and derive the context root that gates the intent) before it signs.
+ *
+ * @module deposit-terms/prePeginApproval
+ */
+
+import { Transaction } from "bitcoinjs-lib";
+
+import { normalizeXOnlyPubkey } from "../managers/pegin/normalizeWalletInputs";
+import { hexToUint8Array } from "../primitives/utils/bitcoin";
+import { deriveVaultRoot, parseFundingOutpointsFromTx } from "../vault-secrets";
+import type { DepositTerms } from "./depositTerms";
+
+/**
+ * Minimal structural wallet for the Pre-PegIn ceremony. Mirrors
+ * `DeriveContextHashCapableWallet` so an app-side wrapper object qualifies
+ * without implementing all of `BitcoinWallet`. Both methods are optional so
+ * the capability probe below can run on any wallet.
+ */
+export interface PrePeginApprovalWallet {
+  deriveContextHash?(appName: string, context: string): Promise<string>;
+  approveDepositTerms?(terms: DepositTerms): Promise<void>;
+}
+
+export interface EnsurePrePeginTermsApprovalParams {
+  wallet: PrePeginApprovalWallet;
+  /** The approved terms — required for approval-capable wallets, ignored (but still txid-checked) otherwise. */
+  depositTerms: DepositTerms | undefined;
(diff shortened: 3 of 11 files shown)


packages/babylon-ts-sdk/src/tbv/core/deposit-terms/index.ts
```diff
@@ -10,3 +10,4 @@
 export * from "./buildDepositTerms";
 export * from "./depositTerms";
 export * from "./depositTermsErrors";
+export * from "./prePeginApproval";
```
gbarkhatov, line 13: This barrel export changes the public API surface, and `docs/api/` is a committed, CI-gated artifact.

I ran the gate locally on this branch:

```
pnpm --filter @babylonlabs-io/ts-sdk run docs:clean
git status --porcelain -- packages/babylon-ts-sdk/docs/api
 M packages/babylon-ts-sdk/docs/api/deposit-terms.md   # +138: PrePeginApprovalWallet, EnsurePrePeginTermsApprovalParams, ensurePrePeginTermsApproval
 M packages/babylon-ts-sdk/docs/api/managers.md        # +14:  SignAndBroadcastParams.depositTerms
```

The "Check generated API docs are current" step in `.github/workflows/verify.yml` fails on exactly this. Regenerate and commit the two files.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  if (typeof wallet.deriveContextHash !== "function") {
+    throw new Error(
+      "A deposit-approval wallet must also implement deriveContextHash, but this one does not.",
+    );
+  }
+
+  // Derive over the funded tx's inputs — the same funding outpoints preparePegin
+  // used, in display order (input.hash is internal byte order).
+  const fundingOutpoints: FundingOutpoint[] = tx.ins.map((input) => ({
```
gbarkhatov: This block — together with the hex-strip + `Transaction.fromHex` on L76-79 — re-implements `parseFundingOutpointsFromTx` byte for byte.

That helper lives in `vault-secrets/parseFundingOutpoints.ts`, is exported from the very barrel this file already imports on L23, is golden-tested (`vault-secrets/__tests__/parseFundingOutpoints.test.ts`), and is already used by `htlcSecretDerivation.ts:73`, `ensureAuthenticatedVpClient.ts:84` and `ResumeDepositContent.tsx:382`. It additionally rejects `tx.ins.length === 0`, which this copy does not — an input-less tx reaches `buildFundingOutpointsCommitment` and fails with a less specific error.

Two copies of the display-order reversal on a derivation path is the drift that module's docstring says it exists to prevent:

```ts
import { deriveVaultRoot, parseFundingOutpointsFromTx } from "../vault-secrets";
...
const fundingOutpoints = parseFundingOutpointsFromTx(fundedPrePeginTxHex);
```

That re-parses the tx a second time (this function still needs `Transaction.fromHex` for `getId()`); if the extra parse bothers you, the alternative is widening the shared helper to take a `Transaction` and keeping one implementation.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
```diff
@@ -0,0 +1,131 @@
...
+    txid: Uint8Array.from(Buffer.from(input.hash).reverse()),
+    vout: input.index,
+  }));
+
+  // The root gates the intent on-device; the broadcast path needs no secrets,
+  // so it is discarded immediately.
+  await deriveVaultRoot(
```
gbarkhatov: `deriveVaultRoot` returns a live 32-byte secret, and dropping the reference is not the same as clearing it — the bytes sit in the heap until GC. Every other consumer in this SDK wipes it explicitly: `PeginManager.ts:771`, `expandPerVaultSecrets.ts:77`, `blockDerivation.ts:270`.

```ts
const root = await deriveVaultRoot(
  { deriveContextHash: wallet.deriveContextHash.bind(wallet) },
  { ... },
);
// the broadcast path needs no secrets — the derive only gates the intent
root.fill(0);
```

Nothing between the call and the wipe can throw, so no `try/finally` needed.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const cleanHex = fundedPrePeginTxHex.startsWith("0x")
+    ? fundedPrePeginTxHex.slice(2)
+    : fundedPrePeginTxHex;
+  const tx = Transaction.fromHex(cleanHex);
+  const txid = tx.getId();
+
+  // Whatever the wallet, if terms were supplied they must be for THIS tx —
+  // the broadcast-path analogue of assertDepositTermsMatchSigningContext.
+  if (
```
gbarkhatov, line 89: Two notes on this check.

**(a) The comment's analogy is off.** `assertDepositTermsMatchSigningContext` (`services/deposit/runDepositorPresignFlow.ts:191`) compares graph scalars (`protocolFeeRate`, `vaultCoreVersion`, the two timelocks) and the keeper/challenger rosters as sets, and never looks at a txid. The two are complements, not analogues — worth rewording so a later reader doesn't assume the scalar checks are covered here.

**(b) Nothing binds the terms' per-vault *values* to the transaction.** The txid equality proves the terms name this tx; it does not prove the amounts the depositor is about to see on the device screen describe it. Not reachable today — terms and tx come out of the same `preparePegin` result — but Part 2 rebuilds terms from chain/indexer, and a mis-rebuilt `peginAmount`/`depositorClaimValue` passes this check unchanged.

The identity is computable from the terms alone, no WASM round-trip:

```
tx.outs[htlcVout].value === peginAmount + depositorClaimValue + peginMaxFee + (vaultCoreVersion === 2 ? 240 : 0)
```

(CLAUDE.md Critical Path #1; `assertEncodedHtlcOutputsMatch` in `primitives/psbt/assertWasmPeginSizing.ts:230` is the existing precedent.) This helper is the natural home for it, and landing it now means Part 2 inherits the check rather than having to remember it.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
```diff
@@ -0,0 +1,131 @@
...
+ * approved `prepegin_max_fee`. So an approval-capable wallet must be shown the
+ * terms (and derive the context root that gates the intent) before it signs.
+ *
+ * @module deposit-terms/prePeginApproval
+ */
+
+import { Transaction } from "bitcoinjs-lib";
+import { Buffer } from "buffer";
+
+import { normalizeXOnlyPubkey } from "../managers/pegin/normalizeWalletInputs";
```
gbarkhatov, line 20: Layering nit: `deposit-terms/` is a typedoc entry point and sits below `managers/`, and this is the only edge from it into the manager tree. No runtime cycle (the target is a leaf module), but `normalizeXOnlyPubkey` is a generic pubkey normalizer — it would sit more naturally next to `processPublicKeyToXOnly` in `primitives/utils/bitcoin`, which it already wraps.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  if (!depositTerms) {
+    throw new Error(
+      "This wallet requires approved deposit terms before signing the Pre-PegIn transaction, " +
+        "but none were provided. Pass PreparePeginResult.depositTerms (fresh flows) or a resume rebuild.",
+    );
+  }
+
+  if (typeof wallet.deriveContextHash !== "function") {
```
gbarkhatov, line 110: Two small robustness notes on the guards:

- This one can never fire for the vault-app caller. The wrapper at `useDepositFlow.ts:818-819` always defines `deriveContextHash` as an arrow function, so a wallet missing the method underneath yields a raw `TypeError` from inside the closure rather than this actionable message. Not wrong here — just worth knowing the good error only protects direct SDK callers.
- The capability probe on L68/L72 uses truthiness where the rest of the seam uses `supportsDepositApproval` (i.e. `typeof === "function"`). A wrapper that spreads a non-function `approveDepositTerms` passes this gate and throws `TypeError` at L130 instead. The structural param type makes the existing probe awkward to reuse directly, but `typeof wallet.approveDepositTerms === "function"` matches the convention at zero cost.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -122,7 +123,7 @@ const MAX_ACCEPTABLE_COMMISSION_BPS_CAP = 9999;
  * Single source for both consumers — feed it the SAME quoted bps at prepare
  * and register time so device-accept stays coextensive with contract-accept.
  */
-function capMaxAcceptableCommissionBps(bps: number): number {
+export function capMaxAcceptableCommissionBps(bps: number): number {
```
gbarkhatov: `managers/index.ts` uses explicit named exports and doesn't re-export this, so the `export` keyword adds nothing reachable — no public consumer, no docs entry, and no importer anywhere in the repo (the only call sites are in-file, L1045 and L1687).

Per the zero-dead-code / no-speculative-API rule, better landed with Part 2's resume rebuild, together with the barrel export it will actually need.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -1195,6 +1205,18 @@ export class PeginManager {
       });
     }
 
+    // Step 3.5: intent-wallet ceremony (derive → approve) immediately before
+    // signing. Placed after prevout resolution — a network failure there must
+    // not burn a two-screen device ceremony — and adjacent to signPsbt to keep
+    // the approve→sign gap minimal (the seam invariant). No-op for wallets that
+    // do not support deposit approval.
+    await ensurePrePeginTermsApproval({
```
gbarkhatov, line 1213: Suggestion, aimed at Part 2 rather than this PR: `impliedFee` is already computed at L1172, a few lines above, and `depositTerms.prepeginMaxFee` is the exact bound the device enforces on the signed fee. Asserting

```ts
if (params.depositTerms && impliedFee > params.depositTerms.prepeginMaxFee) throw ...
```

here fails loudly host-side instead of burning a two-screen ceremony for an opaque device-envelope rejection.

Fresh flows are safe by construction — `preparePegin` throws on `fundedFee !== sizing.fee` at L978 — but the Part 2 resume path reconstructs the terms from chain/indexer, which is precisely the "cross-check assertions belong at the broadcast site, not only at the estimator" case in CLAUDE.md Critical Path #2.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -815,9 +815,14 @@ export function useDepositFlow(
...
+              deriveContextHash: (appName: string, context: string) =>
+                confirmedBtcWallet.deriveContextHash(appName, context),
+              // Object spread drops prototype methods — see forwardDepositApproval.
+              ...forwardDepositApproval(confirmedBtcWallet),
             },
             depositorBtcPubkey: batchResult.depositorBtcPubkey,
             expectedUtxos: utxosToExpectedRecord(batchResult.selectedUTXOs),
+            depositTerms: batchResult.depositTerms,
           });
         } catch (error) {
```
gbarkhatov, line 828: This catch defeats the deliberate unwrapped rethrow added in `vaultPeginBroadcastService.ts:347-349`.

That site preserves `DepositTermsRejectedError` specifically so the UI can tell an intent rejection apart from a broadcast failure. This frame — the only production site the ceremony is wired into — re-wraps every error into a plain `Error`, so `name` and `reason` are gone. `mapDepositError` then string-matches `"broadcast"` (`utils/errors/depositErrors.ts:209`) and returns `ERRORS.broadcastFailed`: exactly the generic copy the rethrow set out to avoid. `depositErrors.ts:191-194` already documents this wrapper hazard for the coded-rejection path.

Mirror the service's guard:

```ts
} catch (error) {
  if (isDepositTermsRejectedError(error)) throw error;
  const errorMsg = error instanceof Error ? error.message : String(error);
  throw new Error(`Failed to broadcast batch Pre-PegIn transaction: ${errorMsg}`);
}
```

Related, and fine to leave: `mapDepositError` has no `isDepositTermsRejectedError` branch and `copy.ts` has no intent-rejection string, so there's no consumer at the far end yet — tracked as #2110 per `depositTermsErrors.ts`. But without the guard above, #2110 will still receive a generic `Error` when it lands.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -815,9 +815,14 @@ export function useDepositFlow(
             btcWalletProvider: {
               signPsbt: (psbtHex: string) =>
                 confirmedBtcWallet.signPsbt(psbtHex),
+              deriveContextHash: (appName: string, context: string) =>
+                confirmedBtcWallet.deriveContextHash(appName, context),
```
gbarkhatov, line 820: This forwards `deriveContextHash` raw, unlike the pegin-signing wrapper at L479-483 which wraps it in `advanceStep(DepositFlowStep.DERIVE_VAULT_SECRET)`.

For a Ledger depositor this step now fires two device prompts (derive, then approve terms) while the stepper still reads "Broadcast Pre-PegIn". No correctness impact, but the UI goes quiet during the part of the flow that needs the most explanation.

services/vault/src/services/vault/vaultPeginBroadcastService.ts
```diff
@@ -62,11 +68,22 @@ export interface BroadcastPrePeginParams {
...
    */
   btcWalletProvider: {
     signPsbt: (psbtHex: string) => Promise<string>;
-  };
+  } & PrePeginApprovalWallet;
+
+  /**
+   * Approved deposit terms — required when `btcWalletProvider` supports deposit
+   * approval. Fresh flows pass `PreparePeginResult.depositTerms`; the resume
+   * path rebuilds them (`rebuildBroadcastDepositTerms`). Ignored (but still
```
gbarkhatov: `rebuildBroadcastDepositTerms` doesn't exist anywhere in the repo yet — fine as a forward reference to Part 2 if that's deliberate, but today it reads as a pointer to code the reader can't find.

More substantively, the resume path this sentence describes is the one that stays silently broken: `useVaultActions.ts:363-370` passes `btcWalletProvider: { signPsbt }` and nothing else — neither `depositTerms` nor the approval capability. Both are optional, so `ensurePrePeginTermsApproval` short-circuits to a no-op at `prePeginApproval.ts:72`, and a Ledger resume fails later at the device with no host-side explanation instead of the actionable "requires approved deposit terms" this PR adds. Nothing forces Part 2 to notice.

Forwarding the capability there now (`...forwardDepositApproval(...)`, as the fresh path does) would turn that into the SDK's typed error today.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/__tests__/prePeginApproval.test.ts
```diff
@@ -0,0 +1,186 @@
...
+import { describe, expect, it, vi } from "vitest";
+
+import type { DepositTerms } from "../depositTerms";
+import {
+  ensurePrePeginTermsApproval,
+  type PrePeginApprovalWallet,
+} from "../prePeginApproval";
+
+/** A funded Pre-PegIn tx with two inputs, so the funding-outpoint derivation is exercised. */
+function makeFundedTx(): { hex: string; txid: string; inputTxids: string[] } {
```
gbarkhatov, line 24: Good orchestration coverage — capability gate, txid mismatch before any device I/O, and the strict `derive → approve` ordering are the right things to pin. Three gaps:

- **Nothing pins the display-order reversal.** `deriveContextHash` is a `vi.fn()` whose arguments are never asserted, so a helper that dropped `.reverse()` on L114 still passes all 7 tests — on a derivation path where that byte order is the entire point. `makeFundedTx` already returns `inputTxids`; asserting the `context` argument equals `uint8ArrayToHex(buildVaultContext({ depositorBtcPubkey, fundingOutpoints: inputTxids.map((t, i) => ({ txid: hexToUint8Array(t), vout: 0 })) }))` would close it.
- **No app-side test at all.** `services/vault/src/services/vault/__tests__/vaultPeginBroadcastService.test.ts` (17 tests) is untouched, so neither the new ceremony call nor the `isDepositTermsRejectedError` branch has coverage — which is how the re-wrap at `useDepositFlow.ts:827` went unnoticed.
- **No test that an `approveDepositTerms` rejection propagates unwrapped** through `signAndBroadcast`, i.e. that nothing in the manager swallows or re-wraps it.

Nit on L25: `Buffer.from(Buffer.from(txid, "hex"))` — the outer copy is redundant.

gbarkhatov (dismissed):
Reviewed the diff against the surrounding code and ran the affected suites locally on `de25ed5`: `prePeginApproval.test.ts` 7/7, `PeginManager.test.ts` 63/63, vault `vaultPeginBroadcastService.test.ts` + `useDepositFlow.test.tsx` 58/58 — all green.

The mechanics hold up where being wrong would be silent:

- `input.hash` reversal produces display order, matching `preparePegin`'s `hexToUint8Array(u.txid)`, and `buildFundingOutpointsCommitment` sorts canonically — so the broadcast-time derive reproduces the prepare-time root regardless of input ordering.
- `prepeginTxid` is `calculateBtcTxHash(fundedPrePeginTxHex)` over the same hex, so the new txid assertion can't false-positive on the fresh path, and it now runs for software wallets too — good defence in depth.
- Ceremony placement (after prevout resolution, adjacent to `signPsbt`) is well argued, and the `["derive", "approve", "sign"]` ordering test pins it without touching the network.

Approving. Two I'd fix before merge:

1. **`docs/api/` is stale — CI fails.** Verified locally: `docs:clean` dirties `deposit-terms.md` (+138) and `managers.md` (+14). Inline on `deposit-terms/index.ts`.
2. **The unwrapped `DepositTermsRejectedError` rethrow is destroyed one frame up**, in `useDepositFlow`'s catch, and `mapDepositError` then string-matches its way to the generic `broadcastFailed` copy. Inline on `useDepositFlow.ts:827`.

Of the rest, the two I'd still take are the duplicated `parseFundingOutpointsFromTx` and the unzeroed vault root; the value-binding and fee-bound suggestions are aimed at Part 2, where chain-rebuilt terms make them reachable.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -815,9 +815,14 @@ export function useDepositFlow(
...
+              // Object spread drops prototype methods — see forwardDepositApproval.
+              ...forwardDepositApproval(confirmedBtcWallet),
             },
             depositorBtcPubkey: batchResult.depositorBtcPubkey,
             expectedUtxos: utxosToExpectedRecord(batchResult.selectedUTXOs),
+            depositTerms: batchResult.depositTerms,
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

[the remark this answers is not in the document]
jrwbabylonlab: Fixed.

packages/babylon-ts-sdk/src/tbv/core/managers/__tests__/PeginManager.test.ts
```diff
@@ -1703,6 +1703,111 @@ describe("PeginManager", () => {
...
+      });
+      (
+        btcWallet as unknown as { approveDepositTerms: typeof approveSpy }
+      ).approveDepositTerms = approveSpy;
+
+      const ethWallet = new MockEthereumWallet();
+      const manager = new PeginManager({
+        btcNetwork: "signet",
+        btcWallet,
+        ethWallet: ethWallet as any,
```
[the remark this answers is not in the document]
jrwbabylonlab: Matches the existing pattern here — every ethWallet mock in this file uses the same cast.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/index.ts
jrwbabylonlab: Fixed — regenerated the two docs files.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
jrwbabylonlab: Fixed — now calls the shared parseFundingOutpointsFromTx.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
jrwbabylonlab: Fixed — wiping the root right after the derive now.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
jrwbabylonlab: Reworded the comment. The value-binding is Part 2 — it only bites once terms are rebuilt from chain.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
jrwbabylonlab: Leaving for now — no cycle, and moving it touches the other importers. Separate cleanup.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/prePeginApproval.ts
jrwbabylonlab: Switched the probe to typeof — fixed. Noted the first point.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
jrwbabylonlab: Fixed — un-exported. Comes back with Part 2's barrel export.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
jrwbabylonlab: Part 2 — reachable once the resume path rebuilds terms from chain.

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: Fixed.

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: No correctness impact — folding it into the Ledger stepper UX work.

services/vault/src/services/vault/vaultPeginBroadcastService.ts
jrwbabylonlab: Reworded to a Part 2 reference. Forwarding the capability in the resume path is Part 2.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/__tests__/prePeginApproval.test.ts
jrwbabylonlab: Fixed — added the context-arg assertion (non-palindromic vectors), app-side ceremony + unwrapped-rejection tests, and dropped the redundant Buffer copy.

gbarkhatov (dismissed):

gbarkhatov (approved):

PR #2241, opened 2026-08-13
fix(vault): move the split minimum notice into the split options panel

jonybur:
## What
When a deposit is too small to be split across two vaults, the app shows a
notice saying how much you would need. That notice has moved: it now sits
inside the split options panel, directly under the two-vault option it is
talking about, instead of up in the deposit amount box.

## Why
Matches the design. The notice is a note about one of the two split choices,
so it reads better sitting next to that choice than floating above the panel.

## How
The notice is rendered by the split selector now rather than the deposit form.
It sits just outside the two-vault option's tappable area, so it does not get
greyed out along with the disabled option and cannot be clicked by mistake.

Worth knowing: the split panel starts closed, so the minimum is only visible
once you open it. It used to be on screen at all times.


jonybur:
Pushed the review fixes.

**The actionable one.** The negative case built its query from a hardcoded `"0.4 BTC"` that had nothing to do with the props under test, so a copy or format change would have made it match nothing and pass for the wrong reason. It now asserts the hint's own `role="status"` is absent.

**Added a collapsed-panel case.** The PR body calls out that the notice is now only on screen once the split panel is open — that visibility change had no test. It does now: amount too low to split, panel collapsed, no hint.

**Named the minimum-deposit fixture.**

**One nit not taken:** extracting `"px-4 pb-4"` into a constant. That exact class string already appears inline a few lines below on the loading row, and the file is inline Tailwind throughout, so naming this one occurrence would make it look meaningful in a way the sibling isn't.

8 tests pass, lint clean.

services/vault/src/components/simple/UtxoSplitSelectorV3.tsx
```diff
@@ -123,6 +125,18 @@ export function UtxoSplitSelectorV3({
             )}
           </div>
 
+          {/* When the amount is too low to split, the minimum is spelled out
+              under the (disabled) two-vault option it applies to. It sits
+              outside that option's clickable box so it is neither dimmed by the
+              disabled styling nor a click target. */}
+          {twoVaultSplit.isSplitAmountTooLow && (
```
gbarkhatov: **The minimum is now unreachable in the default state.**

`DepositForm` mounts this panel collapsed — `openPanel` starts as `null` (`DepositForm.tsx:256`) — and below the minimum `canSplit` is false, which makes `isEnabled` (`isTwoVaultSplit && canSplit`, `SimpleDeposit.tsx:266`) false too. So the collapsed header reads just "Do not split": no signal that a split option exists, that it is currently unavailable, or what the minimum is. The user has to open a panel they have no reason to open in order to find out why the thing they never saw is off.

It is also self-erasing. `handleSelectNoSplit` → `select(false)` → `onExpandedChange(false)`, so acknowledging the panel collapses it and hides the notice again while the sub-minimum amount is still entered.

The PR body notes the visibility change, but this feels like a product decision rather than a footnote — previously the minimum was unconditionally on screen next to the amount that violated it, which is where the user is looking. If the placement is settled, two cheap ways to keep discoverability:

- auto-expand once when `isSplitAmountTooLow` first flips true, mirroring the existing one-shot `hasAutoChecked` pattern at `SimpleDeposit.tsx:249-255`;
- or surface the minimum in the collapsed header alongside "Do not split", so the reason travels with the state it explains.

Not blocking the approval — just the one thing I would want an answer on before this ships.

services/vault/src/components/simple/UtxoSplitSelectorV3.tsx
```diff
@@ -123,6 +125,18 @@ export function UtxoSplitSelectorV3({
...
+            <div className="px-4 pb-4">
+              <SplitTooLowHint
```
gbarkhatov: **In this position `SplitTooLowHint`'s live region is inert** — same root cause as the comment above, and either fix there resolves this too.

`AccordionDetails` defaults to `unmountOnExit = false` and computes `mounted = visible || !unmountOnExit`, so children are *always* rendered; collapsing is done purely with `height: 0; visibility: hidden` on the wrapper (`packages/babylon-core-ui/src/components/Accordion/components/AccordionDetails.tsx`, lines 15, 23 and 49-53). Two consequences for the `role="status" aria-live="polite"` region inside `SplitTooLowHint`:

- when the amount crosses below the minimum while the panel is collapsed — the common case, since it starts collapsed — the region mounts inside a `visibility: hidden` subtree, which is outside the accessibility tree. Nothing is announced at the moment it matters.
- when the panel is expanded later, the region is already present and its content unchanged, so there is no mutation for a live region to announce. Screen-reader behaviour on a visibility flip is inconsistent at best, so this cannot be relied on as the announcement path.

In its previous position in the amount card it mounted visible and did announce on amount change. So the regression is a bit sharper than "you have to open the panel to see it" — for screen-reader users the minimum is effectively no longer communicated. Worth weighing when deciding comment 1.

services/vault/src/components/simple/__tests__/UtxoSplitSelectorV3.test.tsx
```diff
@@ -95,6 +99,65 @@ describe("UtxoSplitSelectorV3", () => {
...
+          canSplit: false,
+          isSplitAmountTooLow: true,
+          minDepositForSplit: MIN_DEPOSIT_FOR_SPLIT_SATS,
+        })}
+        expanded={false}
+        onExpandedChange={vi.fn()}
+      />,
+    );
+
+    expect(screen.queryByRole("status")).toBeNull();
```
gbarkhatov: This assertion does not verify what the test name claims. `AccordionDetails` keeps its children mounted when collapsed (`mounted = visible || !unmountOnExit`, with `unmountOnExit` defaulting to false), so the hint node *is* in the DOM here. `queryByRole("status")` returns `null` only because `dom-accessibility-api` sees `visibility: hidden` inherited from the collapsed wrapper.

Two problems with that. The test passes for a reason a reader cannot see anywhere in the test body, and it is load-bearing on an accordion implementation detail: switch the accordion to `display: none` or `unmountOnExit`, or query at the DOM level (an E2E `data-testid` lookup, say), and the behaviour is unchanged while this test's meaning quietly changes.

Since `@testing-library/jest-dom` is already set up (`src/test/setup.ts`), making the intent explicit is a one-liner — assert the node exists and is not visible rather than absent:

```ts
expect(screen.getByText(/* hint minimum */)).not.toBeVisible();
```

and rename to something like "keeps the minimum-deposit hint out of the accessibility tree while the panel is collapsed". That states the real behaviour, survives an accordion change, and doubles as documentation for the live-region point above.

services/vault/src/components/simple/DepositForm.tsx
```diff
@@ -450,15 +449,6 @@ export function DepositForm({
...
-            they need. The split selector below stays visible with its
-            two-vault option disabled. */}
-        {twoVaultSplit?.isSplitAmountTooLow && (
-          <SplitTooLowHint
-            minDepositForSplit={twoVaultSplit.minDepositForSplit}
-          />
-        )}
         {/* Near the per-position vault cap: a split would overflow, so the
             deposit proceeds as a single vault. Surface usage + why split is off. */}
         {vaultCapSplitUnavailable && vaultCapUsage && (
```
gbarkhatov, line 454: Nit, pre-existing, explicitly not for this PR: this banner's markup is identical to `SplitTooLowHint` — same `role="status" aria-live="polite"`, same `flex w-full items-center justify-center gap-2 rounded-lg border border-secondary-strokeLight px-3 py-2 text-center`, same `IoInformationCircle size={18} className="mt-px shrink-0 text-accent-primary"`, same inner span classes. Only the message differs.

The duplication predates this change, but until now the two rendered a few lines apart in this file, which kept it visible; after the move this inline copy is the only one left here. A shared `InfoNotice` taking children (with `SplitTooLowHint` and this one as its two callers) would be a decent follow-up.

services/vault/src/copy.ts
```diff
@@ -648,9 +648,9 @@ export const COPY = {
...
           : TWO_VAULT_SPLIT_NAME,
       splitOptionRecommended: "(Recommended)",
-      // Shown inside the amount card (below "Max to Borrow") when the deposit is
-      // below the minimum needed to split across two vaults; the split selector
-      // below stays visible with its two-vault option disabled. `minBtc` already
+      // Shown inside the expanded split selector, under the two-vault option,
+      // when the deposit is below the minimum needed to split across two
+      // vaults; that option stays visible but disabled. `minBtc` already
       // carries the network coin symbol (e.g. "0.4 BTC"). The split name and
       // minimum are
```
gbarkhatov: Comment-reflow artifact: the rewrite leaves `// minimum are` as a one-word line before `// emphasized (primary text) by the component;`. Worth reflowing the tail so the paragraph reads continuously.

gbarkhatov (dismissed):
Reviewed the placement move end to end: the notice has exactly one renderer now, no dangling references were left behind, the `copy.ts` comment was updated alongside it, gating is consistent (`isSplitAmountTooLow` always implies `canSplit === false`, and `allowSplit` is amount-independent so the panel still renders), and the affected tests pass. No correctness or value-path issues — nothing here touches amounts, fees, or signing.

What is worth a decision before merge is a consequence of the new position rather than a bug in it: the panel is mounted collapsed, so below the minimum the user sees only a "Do not split" header with no reason attached, and the hint's live region is hidden from assistive tech while collapsed. Details inline on `UtxoSplitSelectorV3.tsx` — both share one root cause and one fix. The remaining three comments are a test-strength note and two nits.

Approving, since none of this blocks: the change does what it says and is a net improvement in grouping. Comment 1 is the one I would like a product answer on.

services/vault/src/components/simple/UtxoSplitSelectorV3.tsx
jonybur: Confirmed and fixed in 82ca1349c. I checked it directly: with the panel collapsed the hint node IS in the DOM, the wrapper carries `height: 0px; visibility: hidden`, and `queryByRole("status")` returns null — so the region was outside the accessibility tree at exactly the moment worth announcing.

The fix keeps the visual design untouched. `UtxoSplitSelectorV3` now owns an `sr-only` live region rendered outside the accordion, and `SplitTooLowHint` is no longer a live region itself, so the sentence is announced exactly once whether the panel is open or closed. The region is rendered unconditionally and emptied rather than unmounted, since a region has to be in the DOM before its content changes to be announced at all; `sr-only` is absolutely positioned and the parent uses `gap` rather than `space-y`, so it takes no part in the layout.

`splitTooLowHint` now also returns the sentence unsplit, assembled from the same fragments rather than written twice — of two copies of one sentence, the invisible one is the one that drifts unnoticed.

services/vault/src/components/simple/UtxoSplitSelectorV3.tsx
jonybur: Taken as a product decision, and the answer is to keep the placement and fix what it broke — see the live-region thread for the implementation.

The reasoning: the placement matches the design and the notice genuinely does belong next to the choice it qualifies, so neither auto-expanding nor moving the minimum into the collapsed header was the right trade. What was not intended is that screen-reader users lost the sentence entirely, and that is now fixed by an off-screen live region outside the accordion, announced when the amount crosses below the minimum.

So for sighted users the behaviour you describe stands as designed — the minimum lives inside the panel, and the header still reads just "Do not split". The self-erasing collapse is likewise unchanged. Flagging that plainly rather than resolving it as a non-issue: if discoverability turns out to matter in practice, the one-shot auto-expand mirroring `hasAutoChecked` is the cheapest follow-up.

services/vault/src/components/simple/__tests__/UtxoSplitSelectorV3.test.tsx
jonybur: Right, and fixed in 82ca1349c. Confirmed the mechanism first — the node is in the DOM while collapsed and `queryByRole` returned null only because of the inherited `visibility: hidden`.

The test now asserts the node exists and is not visible, and is renamed to "keeps the minimum-deposit hint out of the accessibility tree while the panel is collapsed". A second test covers the new off-screen region still carrying the minimum in that state, so the two halves of the behaviour are pinned separately.

services/vault/src/copy.ts
jonybur: Fixed in 82ca1349c — reflowed.

services/vault/src/components/simple/DepositForm.tsx
jonybur: Agreed on both the duplication and that it is out of scope here, so I have left it. Noting for whoever picks it up: after this move the two are no longer a few lines apart, so a shared `InfoNotice` taking children is a slightly easier change now than it was — `SplitTooLowHint` also lost its `role="status"`/`aria-live` in this PR, which removes the one behavioural difference between them.

gbarkhatov (dismissed):
Re-approving after the review follow-ups. Checked each thread against the code at 8da60076e rather than the replies alone.

The live-region fix is the substantive one and it holds up: the selector now owns an unconditional `sr-only` `role="status"` region rendered outside the accordion, emptied rather than unmounted, and `SplitTooLowHint` no longer carries a region of its own — so the sentence is announced once whether the panel is open or closed. `splitTooLowHint` returning the sentence assembled from the same fragments is the right call; the invisible copy is the one that would have drifted.

The collapsed-panel test now asserts the node exists and is not visible, under a name that says so, so it no longer passes for a reason the test body doesn't show. The copy.ts comment is reflowed, and the test fixture is named. Nine tests pass locally.

Declining the `px-4 pb-4` constant is fine — the identical string sits inline on the loading row a few lines below, and naming one occurrence would imply a meaning the sibling doesn't have.

On my first comment: taking the placement as settled and fixing what it broke is a reasonable answer, and the screen-reader half was the part that was actually a regression. The collapsed header still reads just "Do not split" below the minimum, and selecting an option re-collapses the panel — worth watching, and the one-shot auto-expand mirroring `hasAutoChecked` remains the cheap fix if it turns out to matter. Not blocking. The DepositForm/InfoNotice duplication stays a follow-up as scoped.

gbarkhatov (approved):

PR #2242, opened 2026-08-13
perf(vault): cache the IRM curve read and stop per-minute polling

jeremy-babylonlabs:
## What

Two commits:

1. `perf(vault)`: the interest-rate-model chart's curve read is cached for an hour client-side and the live "Current" marker moves onto the page's existing 60-second reserve reads.
2. `feat(vault)`: the curve is then fetched from the vault indexer's new `GET /api/aave/reserves/:id/irm` route (server-side sampled, Redis-cached) instead of each browser issuing its own `eth_call` batches. Total RPC load for the chart becomes O(1) in audience size.

## ⚠ Deploy dependency

**The indexer must deploy the `/irm` route before this merges** (babylonlabs-io/babylon-vault-indexer#164). Without it the chart renders its unavailable state.

## Why this is safe

The curve is a pure function of the strategy's governance-set parameters. `AssetInterestRateStrategy.calculateInterestRate` ignores its `deficit` argument and depends only on the utilization ratio, so the shape cannot change between governance actions. A stale comment claiming the live deficit shifts the curve is corrected.

## Design points

- Endpoint contract: 200 only with a complete curve; every degraded state (unknown reserve, upstream RPC failure, zero liquidity) is a non-200 — required because the client caches any 200 for an hour.
- Hook: `staleTime` = `gcTime` = 1h; hourly refetch while mounted (`staleTime` alone never re-invokes the queryFn and focus refetch is disabled app-wide); an errored query retries every 60s, single attempt per cycle; on error the hook withholds retained data, same guard as the sibling hooks.
- Card: the Current marker derives entirely from live utilization — its callout APR is interpolated off the cached curve at that x (new pure `aprAtUtilization` helper), so the dot and label can never disagree. Missing live figures degrade to a kink-only chart.
- The browser-side sampling client is deleted; the hub helpers it used become internal to `aaveHub.ts`.

## Verification

- Vault suite green (582 tests across the touched areas in the final scoped run; full suite green on commit 1 with 3168). Lint 0 errors, `tsc --noEmit` clean.
- Live-verified on devnet (commit 1): the curve read fires once per session instead of every 60s; range/marker updates ride the existing indexer + batch reads.


services/vault/src/applications/aave/hooks/useInterestRateModelCurve.ts
```diff
@@ -43,23 +63,41 @@ export function useInterestRateModelCurve({
...
+            const result = await getInterestRateModelCurveSafe({
+              hub: reserve.reserve.hub,
+              assetId: reserve.reserve.assetId,
+            });
+            // A Safe-result failure must surface as a query ERROR, not be
+            // cached as hour-fresh data: stored as success, one transient RPC
+            // failure would pin "chart unavailable" for the full cache
+            // window. As an error it is refetched every 60s (below) and
+            // refetched on the next mount regardless of `staleTime`.
+            if (result.error !== null) throw result.error;
```
gbarkhatov: **Rethrowing turns a silent failure into a per-minute Sentry issue.**

A thrown queryFn error reaches `queryCache.onError` → `reportQueryCacheError` → `logger.error` → `captureException` (`queryClient.ts:139-141`, `infrastructure/logger.ts:46`). `Error("Interest-rate strategy curve read reverted")` is neither a 451 nor in `EXPECTED_QUERY_ERROR_NAMES`, so it takes the full capture path. Combined with the 60s error `refetchInterval`, a market whose strategy read reverts bills ~60 captured issues per hour per viewer for as long as the tab stays open. Before this PR: zero.

That is the failure mode the comment at `queryClient.ts:15-33` was written about — "a geo-block made every poll cycle issue four doomed requests… each settled failure billing a Sentry issue".

The goal here (never cache a failure as hour-fresh) doesn't require minting an error event. react-query 5.90.20 is installed and supports functional `staleTime`, so keeping the Safe-result-as-data shape gets you the same guarantee:

```ts
staleTime: (query) => (query.state.data?.error ? 0 : CURVE_CACHE_MS),
refetchInterval: (query) =>
  query.state.data?.error ? ERROR_REFETCH_INTERVAL_MS : CURVE_CACHE_MS,
```

Note the `refetchInterval` predicate has to move to `data?.error` as well — with the failure carried as data the status stays `"success"`, so the current `status === "error"` check would never fire.

If you'd rather keep the throw, the minimal fix is to back the error interval off toward the hour after N consecutive failures, so a permanently broken market stops reporting forever.

services/vault/src/applications/aave/clients/aaveIrm.ts
```diff
@@ -4,17 +4,23 @@
...
  * Sampling exploits that `T = liquidity + drawn + swept` is invariant: for a
  * synthetic sample at utilization `u`, `drawn' = T·u`, `liquidity' = T − drawn'`,
- * `swept' = 0` folds the sweep into liquidity without changing `T`. The one
- * exception is the live-utilization sample, which passes the untouched
- * `(liquidity, drawn, swept)` so it is bit-identical to the Hub's own
- * `getAssetDrawnRate`.
+ * `swept' = 0` folds the sweep into liquidity without changing `T`.
+ *
+ * The result is a pure function of the strategy's governance-set parameters
+ * and the sampled utilization ratios — `calculateInterestRate` ignores its
```
gbarkhatov: **The cache's safety argument is asserted in prose but not committed to in code.**

This says the curve is "a pure function of the strategy's governance-set parameters and the sampled utilization ratios", and that purity is the entire justification for the hour cache downstream. But line 212 still passes the **live** `deficit` into every sampled leg, and `AssetInterestRateStrategy` isn't in this repo — so nothing here actually holds the assumption in place.

Both statements can't be load-bearing. If a strategy upgrade ever starts consuming that argument, the cached curve is silently wrong for up to an hour and no test catches it. Either pass `0n` so the read is independent of live state by construction (my preference — it makes the doc true rather than aspirational), or soften the wording to "unused as of `AssetInterestRateStrategy` @ \<commit/version\>" so the next reader knows the cache rests on an off-repo fact.

Smaller version of the same thing one level down: `sampleDrawn = (totalLiquidity * bps) / BPS_SCALE` is integer division, so the sampled ratios do depend on live `T` at the rounding level. Immaterial for a display chart, but "pure function" overstates it.

services/vault/src/components/pages/BorrowingMarketsData/InterestRateModelCard.tsx
```diff
@@ -167,19 +182,16 @@ export function InterestRateModelCard({
         ),
         lines: [
           COPY.marketData.charts.calloutApr(
-            formatAprPercent(currentAprPercent),
+            formatAprPercent(
+              aprAtUtilization(curve, currentUtilizationPercent),
```
gbarkhatov, line 189: **This figure and the stats bar's Borrow APR used to be equal by construction; now they can drift.**

The old current-leg passed live totals to `calculateInterestRate` — the same value `getAssetDrawnRate` returns, which is exactly what the stats bar above renders via `useAaveBorrowAprs` → `getAssetDrawnRatesSafe`. Two numbers on one page that could not disagree.

Interpolation itself is exact (piecewise-linear curve, kink is an exact sample), so in steady state the difference sits below `formatAprPercent`'s two decimals. The real window is a governance parameter update: the stats bar moves within 60s while this callout — and the kink marker beside it — stay stale for up to an hour, side by side on the same screen.

Cheapest resolution is to accept it and say so in the hook's module doc; it's the one user-visible cost of the cache and currently isn't written down anywhere. If you want it self-healing, the page already holds the live APR, so comparing it against `aprAtUtilization(curve, currentUtilizationPercent)` and invalidating the curve query on divergence would turn the purity claim into an enforced invariant. That's real added coupling for a display chart though — only worth it if the governance-change window actually bothers you.

services/vault/src/applications/aave/hooks/useInterestRateModelCurve.ts
```diff
@@ -43,23 +63,41 @@ export function useInterestRateModelCurve({
...
+    retry: false,
   });
 
   return {
-    curve: data?.curve ?? null,
-    kinkUtilizationPercent: data?.kinkUtilizationPercent ?? null,
-    currentUtilizationPercent: data?.currentUtilizationPercent ?? null,
-    currentAprPercent: data?.currentAprPercent ?? null,
-    maxAprPercent: data?.maxAprPercent ?? null,
+    curve: error ? null : (data?.curve ?? null),
```
gbarkhatov: **One failed refetch blanks a curve this PR argues cannot have changed.**

A single blip on the hourly refetch drops the card to "Chart unavailable" for at least 60s, discarding a shape that by this PR's own reasoning is hour-stable. The `useAaveReserveLiquidity` precedent cited in the module doc doesn't transfer cleanly: there, retained data is a stale *live* figure and genuinely misleading; here it's a constant.

Worth considering retaining the last-good curve and letting the 60s retry heal it silently. This interacts with the Sentry comment above — if you move to Safe-result-as-data plus functional `staleTime`, `error` never populates and this guard falls away on its own, so the two are best decided together.

services/vault/src/components/pages/BorrowingMarketsData/borrowChartData.ts
```diff
@@ -90,6 +91,34 @@ export function percentAxis(
...
+  for (let i = 1; i < curve.length; i++) {
+    const b = curve[i];
+    if (utilizationPercent > b.utilizationPercent) continue;
+    const a = curve[i - 1];
+    const span = b.utilizationPercent - a.utilizationPercent;
+    if (span === 0) return b.aprPercent;
+    const t = (utilizationPercent - a.utilizationPercent) / span;
+    return a.aprPercent + t * (b.aprPercent - a.aprPercent);
+  }
+  return last.aprPercent;
```
gbarkhatov: Unreachable given the two clamps above — for any `utilizationPercent` strictly between first and last, the loop always returns. The one way to land here is an unsorted `curve`, where this silently returns a wrong APR. `throw new Error("unreachable: curve must be sorted ascending")` fits the no-silent-fallbacks rule better and documents the precondition in the same stroke. The `span === 0` branch on line 115 is likewise unreachable from the only producer (`sampleBpsSet` is a `Set`, so no two samples share a utilization).

Either way the JSDoc should state the ascending-sort precondition — the signature reads as generic, but the function is only correct on sorted input.

services/vault/src/applications/aave/hooks/__tests__/useInterestRateModelCurve.test.tsx
```diff
@@ -112,4 +106,46 @@ describe("useInterestRateModelCurve", () => {
     expect(result.current.isLoading).toBe(false);
     expect(getInterestRateModelCurveSafe).not.toHaveBeenCalled();
   });
+
+  it("withholds the retained curve when a refetch after a successful load fails", async () => {
```
gbarkhatov: Good addition. The gap sitting next to it: nothing locks the cadence this PR is actually about. `staleTime`, `gcTime` and the `refetchInterval` callback could all be edited back to a flat 60s and the entire suite still passes — the change would silently undo itself.

`vi.useFakeTimers()` plus an advance past 60s in the error state and past an hour in the success state would pin both branches of the callback.

services/vault/src/components/pages/BorrowingMarketsData/__tests__/InterestRateModelCard.test.tsx
```diff
@@ -250,20 +255,33 @@ describe("InterestRateModelCard", () => {
...
-    renderCard({ paused: true });
+    renderCard({ reserveOverrides: { paused: true } });
 
     expect(screen.getByText("Optimal (Kink) 80%")).toBeInTheDocument();
     expect(
       screen.queryByText(COPY.marketData.charts.chartUnavailable),
     ).not.toBeInTheDocument();
   });
 
+  it("renders a kink-only chart when the live figures are missing", () => {
```
gbarkhatov, line 284: Same shape of gap for the interpolation. `MOCK_CURVE` has an exact sample at `utilizationPercent: 68` and the card renders at `utilizationBps: 6800`, so every card-level test here passes identically whether `aprAtUtilization` interpolates or does a plain exact lookup — the behavior this PR introduces is only covered in the `borrowChartData` unit tests, never through the component.

A case at `utilizationBps: 5900` (between the 50 and 68 samples) asserting the interpolated callout would pin the wiring.

gbarkhatov (dismissed):
Approve — the decoupling is the right shape. Tying the marker's x and its callout APR to one source removes a class of disagreement that two independently-polled queries always had, and the RPC saving is real.

Verified locally at `b444fab`: affected suites 89/89 pass, `tsc --noEmit` clean, `eslint` 0 errors on the touched trees. Also checked the two things the change silently depends on — `useAaveReserveLiquidity`'s `utilizationBps` is `drawn * 10_000 / (available + drawn + swept)`, byte-for-byte the derivation the deleted `currentBps` used (the marker doesn't move, it only loses sub-BPS precision to the floor), and interpolation between grid points is exact for a piecewise-linear strategy whose kink is an exact sample. No dead code left behind; comments were updated alongside behavior, including the stale deficit claim.

Two I'd want resolved before merge, both inline: the new per-minute Sentry capture on a persistently failing read, and the fact that the hour cache's safety argument is asserted in prose while the code still passes live state. The rest are non-blocking.

**Nits, not worth threads:**

- `BPS_PER_PERCENT = 100` is now a third private copy (`aaveIrm.ts:41`, `InterestRateModelCard.tsx:49`, `utils/formatting.ts:82`). A shared `bpsToPercent` would retire all three.
- `isLoading: reserve !== null && isLoading` (`useInterestRateModelCurve.ts:100`) is redundant under `skipToken` — a skipped query is never `isLoading`. Leftover from the `enabled` idiom it replaced.
- `aaveIrm.ts:242-245` swapped a concrete mechanism (the deficit rode every leg) for "the on-chain math rounds rates up", unsourced, guarding a clamp that should be unreachable for a conforming strategy. What the test actually simulates is a non-conforming custom strategy — say that instead.

services/vault/src/clients/indexer/__tests__/aaveIrmClient.test.ts
```diff
@@ -0,0 +1,384 @@
...
+    const controller = new AbortController();
+
+    const promise = fetchIrmCurve({ reserveId: 1n, signal: controller.signal });
+    expect(capturedSignal?.aborted).toBe(false);
+
+    controller.abort();
+
+    await expect(promise).rejects.toThrow(/Aborted/);
+    expect(capturedSignal?.aborted).toBe(true);
+  });
```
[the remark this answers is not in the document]
gbarkhatov: @jeremy-babylonlabs I think this is the last one

gbarkhatov (dismissed):

jonybur (dismissed):
Reviewed the diff end to end. The two blockers gbarkhatov flagged on the earlier head are addressed - the persistent-failure retry cadence is a single 60s tick with no compounding, and the live-utilization marker is passed separately rather than baked into the hour cache. Dead-code removal is clean: the old browser-side sampling client and its test are gone with no dangling references, and the schema validation in the new indexer client is properly gated (rejects non-200s, invalid semantic curves surface as errors instead of stale data). No fee/value paths touched.

services/vault/src/applications/aave/hooks/useInterestRateModelCurve.ts
```diff
@@ -1,65 +1,146 @@
...
+      reserve === null
+        ? skipToken
+        : async ({ signal }): Promise<CurveQueryData> => {
+            try {
+              const result = await fetchIrmCurve({
+                reserveId: reserve.reserveId,
+                signal,
+              });
+              return { ...result, error: null };
+            } catch (err) {
```
jrwbabylonlab, line 144: Resolving failures as data removes this query from `QueryCache.onError` → Sentry, and the returned `error` has no consumer (the card destructures 4 fields, not `error`).

A broken IRM endpoint becomes a silent outage: no telemetry, no UI signal, and the retained curve chains forward with no age cap — not the "up to an hour" the module doc states.

services/vault/src/clients/indexer/aaveIrmClient.ts
```diff
@@ -0,0 +1,228 @@
...
+      throw new Error(
+        `IRM curve point ${index} from ${endpoint} is not an object`,
+      );
+    }
+    const { utilizationPercent, aprRay, aprPercent } = point as {
+      utilizationPercent?: unknown;
+      aprRay?: unknown;
+      aprPercent?: unknown;
+    };
+    if (typeof aprRay !== "string") {
```
jrwbabylonlab: `aprRay` is required here, then dropped at L111.

The indexer trimming a field no client reads would blank every IRM chart — and per the hook error-as-data path, with no Sentry event to explain it.

services/vault/src/components/pages/BorrowingMarketsData/InterestRateModelCard.tsx
```diff
@@ -124,14 +135,12 @@ export function InterestRateModelCard({
   // The curve always includes an exact sample at the kink utilization (see
   // useInterestRateModelCurve module doc) — looked up by equality, never
   // interpolated. That invariant is load-bearing, so its failure throws
-  // rather than degrading to a chart with no kink marker.
+  // rather than degrading to a chart with no kink marker. The "Current"
+  // marker derives entirely from the live utilization — its callout APR is
+  // read off the cached curve at that x, so the dot and the label can never
+  // disagree (previously they came from two independently-polled queries).
   const markers = useMemo<LineChartMarker[]>(() => {
```
jrwbabylonlab, line 147: This memo throws during render, and the curve is now untrusted HTTP data rather than an in-process construction.

The only boundary is the app-wide one (`main.tsx:37`), so a bad kink replaces the entire app with the global error screen instead of degrading to a kink-less chart.

jrwbabylonlab (approved):

PR #2246, opened 2026-08-13
fix(vault): left-align the entry footer logo

jeremy-babylonlabs:


jonybur (approved):
One-line align fix on the entry footer logo; contained to a single wrapper class. Nothing else touched.

PR #2247, opened 2026-08-13
chore(ledger-vault-signer): address #2238 review

jrwbabylonlab:
Follow-up to #2238 — the vendored `ledger-bitcoin` §4(b) record ran ahead of the
code, and the reviewer's top ask (outer maps-roots coverage) was derivable today.

- **§4(b) accuracy** (NOTICES / README / `merkle.ts` header): dropped the
not-yet-vendored `psbtv2` modification + "bundled" claim, fixed the
invalid-index note (raw `TypeError` → typed `Error`), recorded non-canonical
`parseVarint`, moved `psbtv2` to a Planned subsection.
- **`dist/` now ships no vendored code** — excluded the vendor dir from `dts`
emission (was shipping `dist/vendor/**/*.d.ts`).
- **+6 golden assertions** — outer maps-roots + `sign_psbt_cdata_hex`, all 3
fixtures.

### DIFF
--- a/packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
+++ b/packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
@@ -1,8 +1,8 @@
 # Third-party notices
 
-This file covers third-party source vendored into this repository's package tree
-and bundled into published artifacts. It is not an exhaustive inventory of every
-bundled npm dependency — those retain their own license files. Vendored source:
+This file covers third-party source vendored into this repository's package tree.
+It is not an exhaustive inventory of every bundled npm dependency — those retain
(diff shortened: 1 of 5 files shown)


jonybur (approved):
The NOTICES/README updates reflect what actually ships today, and dropping the vendor dir from the dts build closes the gap where declarations were leaking out even though the JS was tree-shaken. The header on merkle.ts is now accurate about the typed Error("Index out of bounds") behaviour.

The six new golden assertions cover the two things that were actually derivable now - the outer maps-roots (a distinct byte treatment from the inner keys/values trees, so it can fail while every per-map assertion passes) and the SIGN_PSBT client data reconstruction - and they exercise createVarint in its real context. Solid follow-up.

PR #2250, opened 2026-08-13
test(vault): e2e v3 ui

gbarkhatov:
# Restore the real-wallet E2E CLI against the v3 UI

## What was broken

The real-wallet E2E CLI (`services/vault/e2e/real/`) drives the whole depositor lifecycle — connect, peg-in, borrow, repay, withdraw, resume — against real signet BTC and Sepolia ETH. It stopped working and nobody noticed, because it is run by hand rather than in CI.

The cause was [#2184](https://github.com/babylonlabs-io/babylon-toolkit/pull/2184), which retired the `ENABLE_V3_UI` flag and deleted the v2 shell. Two things broke:

1. **One page became six routes.** The v2 dashboard had deposit, collateral, pending deposits and the loan buttons all on `/`. In v3 they live on `/vaults` and `/loans`. The harness still opened `/` and looked for everything there.
2. **Nine selectors were deleted along with the v2 components.** Five of them were the entire first half of the withdraw flow.

The worst part: **connect itself failed**. It waited for the deposit button as its "we're connected now" signal, and that button only exists on `/vaults` in v3. So the wallets connected fine and then the run sat for 60 seconds and died.

Last green run was 22 July; last harness change 24 July. The v3 rewrite landed after both.

## What this PR does

Two sides, both small, because **the flows themselves did not change — only where they live and what identifies them**.

### App side: five test hooks added back

Each one carries the pointer comment that the "E2E TEST HOOKS" rule in `CLAUDE.md` requires. That comment is exactly what was missing before: five withdraw hooks had no comment, so [#2184](https://github.com/babylonlabs-io/babylon-toolkit/pull/2184) deleted them and nothing flagged it.

| testid | file | used by |
|---|---|---|
| `nav-<id>` | `AppSidebar.tsx` | changing section |
| `wallet-menu-trigger` | `Wallet/Connect.tsx` | "connected" signal |
| `vault-row-<vaultId>` | `VaultsActiveSection.tsx` | finding a vault |
| `vault-withdraw-button` | `VaultsActiveSection.tsx` | withdraw entry |
| `pending-deposit-row` | `VaultsLifecycleSections.tsx` | resume entry |

`vault-row` uses `ListRowCard`'s existing `testId` prop, which was already there but no caller passed.

No behaviour changes, no copy changes, no critical-path files touched.

### Test side

- **New `goToSection()` helper.** Every action now navigates to its page first.
- **Connect** keys on the wallet menu instead of a page button, so it no longer depends on which route we're on.
- **Peg-in, borrow, repay** each needed one navigation call and nothing else.
- **Withdraw** was rewritten. v3 deleted the "⋯" menu, the vault-selection modal and its confirm step — you now click Withdraw on a vault row and go straight to Review. Four functions and five dead selectors removed.
- **Resume** now waits for a pending row instead of the deleted expander.

### How to navigate between pages

| option | why not / why yes |
|---|---|
| `page.goto('/vaults')` | Simple, but it is a full page reload. That re-initialises the wallet connection and throws away in-memory deposit state. Chained runs (`withdraw --borrow-first --pegin-first`) change page mid-lifecycle, so this could break a run halfway through a real deposit. |
| **Click the sidebar link** ✅ | Normal client-side navigation, nothing is lost, and it tests the real routing the way a user would. |

### How to find things on the page

| option | why not / why yes |
|---|---|
| Text and structure only (no app changes) | Zero diff in `src`, but it means matching on button labels and DOM shape. That breaks again the next time someone edits copy or moves a `div`. It is also how we got here. |
| **Add test IDs back to the app** ✅ | Five one-line additions, each with a comment saying which test file depends on it. Stable, and it makes the dependency visible to the next person who refactors the component. |

### `--withdraw-all` with no multi-select

v3 opens the withdraw flow with exactly one vault preselected, so the old "tick several checkboxes" approach cannot be reproduced. We could have faked it, but the honest equivalent is to repeat the whole row → Review → Done cycle once per vault. That is one on-chain transaction per vault, which is what actually happens.

## Bugs found while verifying

Three real bugs surfaced during the test runs. Two were pre-existing; the third was introduced and fixed within this PR.

**1. The Pre-PegIn transaction ID was read from the wrong place.** The helper scanned the whole page for the first transaction link. The deposit dialog is a full-screen overlay with `/vaults` still mounted underneath, so as soon as the depositor had any vault, it picked up an unrelated vault's hash. A split run made this obvious: the final check "matched" a vault from a previous run and then blamed indexer lag. Now scoped to the dialog. Pre-existing — `resume.ts` already documented the same trap and worked around it.

**2. The split check compared two different transactions.** For a two-vault split it matched rows by Pre-PegIn transaction ID, but an active row shows the **peg-in** transaction (`peginTxHash ?? prePeginTxHash`), which is a different transaction. It could only ever match in the brief window before the indexer fills that field in — so it was guaranteed to fail later. Replaced with a simpler check: count the vault rows before and after, and assert the count went up by the expected number.

**3. That new check was wrong on the first attempt.** It counted rows immediately after navigating, before the list had loaded, so the "before" number was always 0 — which made the check pass no matter what. A real run caught it (`up 8 from 0` where 6 was correct). Fixed by counting only once the page has visibly loaded. This one was ours, found and fixed inside this PR.

**4. Repeated withdrawals could start before the previous success screen closed.** Flagged in review. The withdraw flow is an overlay that never changes the URL, so the `goToSection` call between passes synchronised nothing, and the next pass could scan rows still covered by the overlay — where a covered button still reports as enabled, so the run would fail later with a confusing timeout. The next pass now waits for the overlay to close first. The Done click stays tolerant of failure on purpose: success is already proven by the initiated screen, so a missed click must not fail a real withdrawal.

Also removed: the dead `NEXT_PUBLIC_FF_ENABLE_V3_UI` env var in the visual config, and the orphaned `COPY.withdraw.modal` / `COPY.collateral.menu` blocks — all left behind by [#2184](https://github.com/babylonlabs-io/babylon-toolkit/pull/2184).

## Testing

Every leg was run for real on signet + Sepolia. Each one checks on-chain state, not just that a screen appeared.

| leg | proof |
|---|---|
| connect | both wallet addresses matched |
| peg-in | vault activated and shown on `/vaults` |
| borrow | debt $0.00 → $113.22 |
| repay | debt $113.22 → $84.92 |
| withdraw | collateral 1,000,000 → 0 sats |
| resume | interrupted deposit → activated vault |
| split peg-in | 8 → 10 vaults (`up 2 from 8`) |
| split resume | one row's button drove both vaults |
| withdraw all | 10 vaults released in one run, 17,353,536 → 0 sats |

Also: `tsc` clean, lint clean (0 errors), **3207 unit tests pass**, and every test ID the harness uses was confirmed to exist in the app.

## Reviewer notes

- The `src` diff is five one-line test IDs plus their comments, one deleted env var and two deleted copy blocks. Everything else is test code.
- If you move or rename any element carrying one of those test IDs, carry the ID over — there is no compile error if you drop one, the E2E run just fails later.

### DIFF
--- a/services/vault/e2e/pages/AppShell.ts
+++ b/services/vault/e2e/pages/AppShell.ts
@@ -10,14 +10,7 @@
 
 import type { Locator, Page } from "@playwright/test";
 
-/** Sidebar section ids, mirroring `src/config/v3Navigation.ts`. */
-export type NavSection =
-  | "overview"
-  | "vaults"
-  | "loans"
-  | "activity"
-  | "liquidations"
-  | "explore";
+import type { V3NavItemId } from "@/config/v3Navigation";
 
 export class AppShell {
   constructor(public readonly page: Page) {}
@@ -36,12 +29,17 @@ export class AppShell {
    * A sidebar nav link, by section id (`AppSidebar`'s `nav-<id>` testid).
    * The sidebar is hidden on the disconnected entry layout, so this only
    * resolves once the app is connected or on a non-root route.
+   *
+   * The id type comes from the production nav config rather than a copy here,
+   * so adding a section can't leave this file silently stale — a new id is
+   * accepted automatically, and a renamed one fails to compile. Type-only, so
+   * nothing from `src` is pulled in at runtime.
    */
-  navLink(section: NavSection): Locator {
+  navLink(section: V3NavItemId): Locator {
     return this.page.getByTestId(`nav-${section}`);
   }
 
-  async openSection(section: NavSection): Promise<void> {
+  async openSection(section: V3NavItemId): Promise<void> {
     await this.navLink(section).click();
   }
 }
--- a/services/vault/e2e/pages/AppShell.ts
+++ b/services/vault/e2e/pages/AppShell.ts
@@ -1,7 +1,6 @@
 /**
- * Page object for the vault dApp's persistent shell: top nav, connect
- * button, theme toggle, and the route-level entry points (Applications
- * tab, Activity tab).
+ * Page object for the vault dApp's persistent shell: sidebar nav, connect
+ * button, and the route-level entry points.
  *
  * Selectors prefer role+name with COPY constants so that copy edits
  * keep tests passing. Where a stable data-testid exists in source, it
@@ -10,6 +9,15 @@
  */
 
 import type { Locator, Page } from "@playwright/test";
+
+/** Sidebar section ids, mirroring `src/config/v3Navigation.ts`. */
+export type NavSection =
+  | "overview"
+  | "vaults"
+  | "loans"
+  | "activity"
+  | "liquidations"
+  | "explore";
 
 export class AppShell {
   constructor(public readonly page: Page) {}
@@ -24,19 +32,16 @@ export class AppShell {
     return this.page.getByRole("button", { name: /^Connect/i });
   }
 
-  get applicationsTab(): Locator {
-    return this.page.getByRole("link", { name: /Applications/i });
+  /**
+   * A sidebar nav link, by section id (`AppSidebar`'s `nav-<id>` testid).
+   * The sidebar is hidden on the disconnected entry layout, so this only
+   * resolves once the app is connected or on a non-root route.
+   */
+  navLink(section: NavSection): Locator {
+    return this.page.getByTestId(`nav-${section}`);
   }
 
-  get activityTab(): Locator {
-    return this.page.getByRole("link", { name: /Activity/i });
-  }
-
-  async openActivity(): Promise<void> {
-    await this.activityTab.click();
-  }
-
-  async openApplications(): Promise<void> {
-    await this.applicationsTab.click();
+  async openSection(section: NavSection): Promise<void> {
+    await this.navLink(section).click();
   }
 }
--- a/services/vault/e2e/real/actions/resume.ts
+++ b/services/vault/e2e/real/actions/resume.ts
@@ -1,25 +1,28 @@
 /**
- * The "resume" action: recover an interrupted peg-in from the dashboard's Pending Deposits UI and drive
+ * The "resume" action: recover an interrupted peg-in from the /vaults pending-deposits list and drive
  * it through to an activated vault.
  *
  * A real peg-in spans 30 min–2 hr and the app is built to survive interruption: once the Pre-PegIn is
- * broadcast the deposit is on-chain, and the dashboard resurfaces it under "Pending Deposits" with a
- * resume action (Submit WOTS Key → Sign Payouts → Activate) driven from live vault-provider state. This
- * action opens that pending card and hands off to the SAME `walkStepMachine` the `pegin` action uses —
- * the resume flow renders the IDENTICAL `DepositProgressView` stepper — so the walk, the Activate-Vault
- * gate, and the activated-view finish line are shared, not reimplemented.
+ * broadcast the deposit is on-chain, and /vaults resurfaces it as a pending row with a resume action
+ * (Submit WOTS Key → Sign Payouts → Activate) driven from live vault-provider state. This action opens
+ * that pending row and hands off to the SAME `walkStepMachine` the `pegin` action uses — the resume
+ * flow renders the IDENTICAL `DepositProgressView` stepper — so the walk, the Activate-Vault gate, and
+ * the activated-view finish line are shared, not reimplemented.
  *
  * Two shapes:
- *   - default: resume an ALREADY-pending deposit (by `--txid`, else the first actionable card).
+ *   - default: resume an ALREADY-pending deposit (by `--txid`, else the first actionable row).
  *   - `--interrupt-fresh`: peg in a fresh deposit, interrupt it right after Pre-PegIn broadcast (reload
- *     the page to a cold state), then resume it from the dashboard — a fully self-contained run.
+ *     the page to a cold state), then resume it from /vaults — a fully self-contained run.
  *
- * Cold-resume specifics: unlike the happy path (which holds the vault root in memory), the dashboard
- * resume RE-DERIVES the vault root from the BTC wallet — a `deriveContextHash` approval dialog — to
- * recompute the WOTS keys and verify them against the on-chain hash. That dialog is the same one normal
- * peg-in step 1 (DERIVE_VAULT_SECRET) fires, so the shared pop-up approver already handles it; the
- * resume steps otherwise auto-fire on mount, so the only dapp-page interactions are the pending-card
- * click, the Activate gate, and "Go to Dashboard" — all owned by `walkStepMachine`.
+ * Cold-resume specifics: unlike the happy path (which holds the vault root in memory), the resume
+ * RE-DERIVES the vault root from the BTC wallet — a `deriveContextHash` approval dialog — to recompute
+ * the WOTS keys and verify them against the on-chain hash. That dialog is the same one normal peg-in
+ * step 1 (DERIVE_VAULT_SECRET) fires, so the shared pop-up approver already handles it; the resume steps
+ * otherwise auto-fire on mount, so the only dapp-page interactions are the pending-row CTA click, the
+ * Activate gate, and "Go to Dashboard" — all owned by `walkStepMachine`.
+ *
+ * v3 note (markdown/e2e-v3/06-resume.md): pending deposits are flat, always-visible rows — the v2
+ * expander they used to hide behind is gone, so there is nothing to expand.
  *
  * NEVER run `--interrupt-fresh` without an explicit go-ahead: it spends real signet BTC + Sepolia ETH.
  */
@@ -29,57 +32,55 @@ import {
   CONNECT_STATE_POLL_MS,
   CONNECT_STATE_TIMEOUT_MS,
   RESUME_ACTIONABLE_TIMEOUT_MS,
-  RESUME_CARD_APPEAR_TIMEOUT_MS,
   RESUME_POLL_INTERVAL_MS,
+  RESUME_ROW_APPEAR_TIMEOUT_MS,
   STEP_TIMEOUT_MS,
 } from "../timing";
 
 import { installPopupApprover, sweepApprovals } from "./approver";
+import { goToSection } from "./navigation";
 import { fillDepositForm, startSigning } from "./pegin";
 import { startRecording } from "./recording";
 import {
   assertActivatedAndOnDashboard,
+  countActiveVaultRows,
   walkStepMachine,
   walkUntilPrePeginBroadcast,
 } from "./stepMachine";
 import { type Action, type ActionContext } from "./types";
-import { connectWallets } from "./walletConnect";
+import { connectWallets, WALLET_MENU_TRIGGER_TESTID } from "./walletConnect";
 
-const DEPOSIT_BUTTON_TESTID = '[data-testid="deposit-button"]'; // connected-state signal (dashboard)
 const CONNECT_BUTTON_TESTID = '[data-testid="connect-wallet-button"]'; // shown when disconnected
-// PendingDepositSection's ExpandMenuButton — the pending list (and each card's resume CTA) lives behind
-// it (aria-label="Pending deposit details"). The button is only rendered when a pending deposit exists,
-// so its presence doubles as "there is something to resume".
-const PENDING_EXPAND_RX = /pending deposit details/i;
-// Added to services/vault/src for this action (see PendingDepositCard.tsx): the card root carries the
-// deposit's Pre-PegIn txid (its tx-hash row links it), and the resume CTA renders ONLY once the deposit
-// is actionable — so waiting for the CTA is waiting for vault-provider readiness.
-const PENDING_CARD_TESTID = '[data-testid="pending-deposit-card"]';
+// VaultsLifecycleSections' PendingRow: the row carries the deposit's Pre-PegIn txid (its hash cell
+// links it), and it is rendered only while a deposit is pending — so its presence doubles as "there is
+// something to resume". The resume CTA inside renders ONLY once the deposit is actionable, so waiting
+// for the CTA is waiting for vault-provider readiness.
+const PENDING_ROW_TESTID = '[data-testid="pending-deposit-row"]';
 const PENDING_RESUME_CTA_TESTID = '[data-testid="pending-deposit-resume-cta"]';
 
-/** Normalize a Pre-PegIn txid flag to the bare lowercase hex the card's explorer href contains. */
+/** Normalize a Pre-PegIn txid flag to the bare lowercase hex the row's explorer href contains. */
 function normalizeTxid(txid: string | undefined): string | undefined {
   const hex = txid?.trim().replace(/^0x/i, "").toLowerCase();
   return hex && /^[0-9a-f]{64}$/.test(hex) ? hex : undefined;
 }
 
 /**
- * After the interrupt reload, make sure the app is connected before we look for the pending card (the
+ * After the interrupt reload, make sure the app is connected before we look for the pending row (the
  * pending list reads the connected ETH address). wagmi/AppKit usually auto-reconnects on reload (the
- * deposit button reappears with no pop-up); if instead the Connect button is shown, re-run the connect
- * flow. Neither appearing is left to the pending-card wait to surface a clearer error.
+ * header wallet menu reappears with no pop-up); if instead the Connect button is shown, re-run the
+ * connect flow. Neither appearing is left to the pending-row wait to surface a clearer error.
  */
 async function ensureConnected(ctx: ActionContext): Promise<void> {
   const { page, context, log } = ctx;
-  const deposit = page.locator(DEPOSIT_BUTTON_TESTID).first();
+  const walletMenu = page.locator(WALLET_MENU_TRIGGER_TESTID).first();
   const connect = page.locator(CONNECT_BUTTON_TESTID).first();
   const deadline = Date.now() + CONNECT_STATE_TIMEOUT_MS;
   while (Date.now() < deadline) {
-    if (await deposit.isVisible().catch(() => false)) return; // auto-reconnected
+    if (await walletMenu.isVisible().catch(() => false)) return; // auto-reconnected
     // Only click Connect once it's actually ENABLED. Right after the reload the app re-hydrates and the
     // Connect button renders visible-but-DISABLED for a beat while wagmi/AppKit auto-reconnects; clicking
     // that transient disabled button just times out (it detaches the instant the connected state swaps
-    // in). Waiting instead lets the auto-reconnect win and the deposit button appear above.
+    // in). Waiting instead lets the auto-reconnect win and the wallet menu appear above.
     if (
       (await connect.isVisible().catch(() => false)) &&
       (await connect.isEnabled().catch(() => false))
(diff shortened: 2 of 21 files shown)


services/vault/e2e/real/actions/withdraw.ts
```diff
@@ -391,13 +298,28 @@ export async function runWithdrawFlow(
...
+    log(
+      `Released vault ${shortenVaultId(vaultId)} (${released.size} this run).`,
+    );
+    if (!all) break;
+
+    // Another pass only if a withdrawable row remains. The just-released vault leaves the active list,
+    // so this settles at "nothing left to release" rather than needing a count decided up front.
+    await goToSection(page, "vaults", log);
+    const next = await findWithdrawableVaultId(page, released);
+    if (!("vaultId" in next)) break;
```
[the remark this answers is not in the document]
gbarkhatov: done

jonybur (dismissed):
The src-side diff is exactly the shape called for by the E2E TEST HOOKS rule in CLAUDE.md: five testids added back with the pointer comments the harness needs, plus the deletion of the ENABLE_V3_UI env var and the orphan copy blocks left behind by #2184. No behaviour changes.

Real value here is the three latent bugs found and fixed during verification:

- readPrePeginTxid scoped to the deposit dialog stops a returning depositor's first vault-row hash from being mis-captured as this run's Pre-PegIn - the failure mode was silent (single-vault fell back to amount match; split reported 0/N and blamed indexer).
- The split cross-check switching from txid-match to row-count delta is the right call: an active row renders peginTxHash ?? prePeginTxHash, so a match against the mid-flow Pre-PegIn was guaranteed to fail once the indexer landed.
- Counting rows AFTER the deposit-CTA renders (not right after navigate) fixes the baseline-of-0 that made the delta assertion pass no matter what.

Withdraw-all's overlay-closed wait between passes closes the last one - a covered button reports isEnabled, so scanning early would have handed back a vaultId that then hung on pointer intercept.

Nothing here touches a critical path.

jrwbabylonlab (dismissed):

services/vault/e2e/real/actions/withdraw.ts
```diff
@@ -391,13 +298,47 @@ export async function runWithdrawFlow(
...
+      .then(() => true)
+      .catch(() => false);
+    if (!overlayClosed)
+      throw new Error(
+        `The withdraw success screen stayed open for ${Math.round(STEP_TIMEOUT_MS / MS_PER_SECOND)}s after releasing vault ${shortenVaultId(vaultId)} — its Done button did not dismiss it, so the remaining vaults can't be reached. ${released.size} vault(s) were released; re-run to continue.`,
+      );
+
+    // Another pass only if a withdrawable row remains. The just-released vault leaves the active list,
+    // so this settles at "nothing left to release" rather than needing a count decided up front.
+    await goToSection(page, "vaults", log);
```
jrwbabylonlab, line 381: `--withdraw-all` can release a subset and still exit green.

This scan runs once. The entry path polls the same predicate for ~90s (`openWithdrawForRow`, :135-148) because the row list lags the chain by a poll cycle — and `findWithdrawableVaultId` turns any locator error into "nothing left" (`.catch(() => 0 / null / false)`, :100-111). A sibling still rendering `isActivating` has Withdraw disabled (`VaultsActiveSection.tsx:136-145`), so it reads as absent.

Nothing downstream catches it: `assertCollateralDecreased` only needs `sats < beforeSats` (:264-269), and `released.size` is logged but never asserted. If the pre-withdraw collateral read failed, :343-347 skips even that. Result is "✅ Withdraw complete." with collateral still held.

Suggest polling this scan with the same budget as the entry path, and asserting `released.size` against the pre-flight vault count.

services/vault/e2e/real/actions/stepMachine.ts
```diff
@@ -387,106 +411,112 @@ export async function walkUntilPrePeginBroadcast(
...
-            .catch(() => 0)
-        : Promise.resolve(0);
-    let matched = await matchingCards();
+    if (baselineRowCount === undefined) {
+      log(
+        `⚠️ No pre-deposit row count was captured, so the ${expectedVaults}-vault split cross-check was skipped — activation still succeeded.`,
+      );
+      return;
+    }
+    const target = baselineRowCount + expectedVaults;
```
jrwbabylonlab: This counts rows the harness itself created, so it can't observe whether the split landed.

The app renders one `vault-row-*` per just-activated vault from in-memory state — `useDashboardState.ts:84-129` appends the optimistic entries, `useActivationState.ts:99-124` adds them in the activation-success callback, and `ActivatingVaultsContext.tsx:99-117` shows they're indexer-independent and expire after 90s. So the delta measured here is produced by the activations we just drove.

If neither vault is ingested, the optimistic rows already satisfy `>= baseline + expectedVaults` on the first `rows.count()`, it logs "the N-vault split landed", and 90s later they disappear. The upstream gate at :261-267 doesn't help — it proves we clicked N activation controls, not that anything was ingested.

Suggest excluding activating rows, or comparing against the node-side vault count already fetched pre-flight.

services/vault/e2e/real/actions/withdraw.ts
gbarkhatov: Confirmed and fixed — thank you, this was a real hole.

The asymmetry was exactly as you described: the entry path polled for 90s while the continuation scan ran once, and `findWithdrawableVaultId` reports every locator failure as "nothing withdrawable". Since /vaults re-renders right after each withdrawal settles, a single transient there ended the batch silently.

What changed:

- Extracted `waitForWithdrawableVaultId`, so the loop decides "is there another one?" on the **same polled terms** as the entry path. One predicate, one standard.
- Replaced `assertCollateralDecreased` with `assertVaultsReleased`: it now asserts `vaultCount === beforeCount - released` **exactly**, read from `getPosition`. You were right that `sats < beforeSats` could not distinguish 1-of-3 from 3-of-3 — the snapshot is taken once before the first pass and asserted once after the last, so the first release satisfied it permanently.
- `--withdraw-all` now throws if vaults remain and there is no debt to health-factor-gate them; with debt it warns and names the amount, since leftovers are legitimate there.
- The failed pre-read no longer skips the check. It fails, and says the run would be unverifiable. A verification that skips itself was the anti-pattern.
- Once `released.size` reaches the on-chain count the loop exits without scanning, so the added polling costs nothing in the common case.

Verified on signet + Sepolia: a two-vault split peg-in followed by `--withdraw-all` released both and logged `✅ On-chain vaults fell: 2 → 0 (released 2)`.

services/vault/e2e/real/actions/stepMachine.ts
gbarkhatov: Confirmed and fixed — this was the sharper of the two, and you were right that the check could not observe what it claimed.

Worth recording: a verification run reproduced it exactly. At the moment the row check ran, this is what was on the page:

```
row: "V 0.01000001 sBTC Activating collateral... V vault-provider-0 Withdraw"
```

`Activating collateral…` is `COPY.collateral.activating` — the optimistic in-memory row from `ActivatingVaultsContext`, under the same `vault-row-<vaultId>` testid as a real one. So the delta was counting rows this run's own activations created, and would have reported "the 2-vault split landed" regardless of ingestion.

What changed:

- Peg-in and resume now assert **on-chain**: new `assertVaultCountRose` polls `getPosition` until `vaultIds.length` reaches `before + expectedVaults`, then throws. This closes a gap where peg-in was the only action without an on-chain post-condition — borrow asserts debt rose, repay asserts it fell, withdraw asserts collateral fell, peg-in asserted nothing. That gap is what let a UI check stand in for verification.
- The row-count delta, its `baselineRowCount` plumbing and `countActiveVaultRows` are **deleted** rather than patched — they existed only to feed a check that could not fail.
- `assertActivatedAndOnDashboard` survives as a display check with a comment stating outright that it cannot tell you the peg-in landed, so it is not mistaken for verification again. Your point about the single-vault branch applies too: optimistic entries carry `amountBtc`, so the amount match was equally satisfiable.

Verified on signet + Sepolia: a two-vault split logged `✅ On-chain vaults rose: 0 → 2 (+2, expected +2)` — while the UI check, correctly, only warned.

services/vault/e2e/real/actions/withdraw.ts
```diff
@@ -335,78 +256,145 @@ async function confirmWithdrawSuccess(
...
     }
     await ctx.page.waitForTimeout(WITHDRAW_VERIFY_POLL_MS);
   }
   throw new Error(
-    `Withdraw reached the success screen but on-chain collateral did not fall within ${Math.round(WITHDRAW_TX_TIMEOUT_MS / MS_PER_SECOND)}s (before ${formatBtc(beforeSats)}, last ${formatBtc(lastSats)}) — the position doesn't reflect the withdrawal.`,
+    lastCount == null
+      ? `Withdraw reached the success screen but the on-chain position could not be read within ${Math.round(WITHDRAW_TX_TIMEOUT_MS / MS_PER_SECOND)}s, so the ${releasedCount} release(s) are unverified. Check the Sepolia RPC and the position manually.`
+      : `Withdraw released ${releasedCount} vault(s) in the UI but the on-chain position holds ${lastCount}, not the expected ${target} (${beforeCount} before this run − ${releasedCount}) within ${Math.round(WITHDRAW_TX_TIMEOUT_MS / MS_PER_SECOND)}s — at least one withdrawal did not take effect.`,
   );
 }
```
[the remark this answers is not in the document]
gbarkhatov: Taken — applied a fail-fast guard right after the baseline read.

The path is narrow (`run.ts` already refuses a non-`--pegin-first` run with no collateral, and `--pegin-first` creates the vaults before this point), but you are right about the consequence: a `0` baseline plus any release makes `target` negative, and `assertVaultsReleased` would then poll to timeout and report an impossible expectation — which reads as a broken assertion rather than a bad starting state.

```ts
if (before.vaultCount === 0)
  throw new Error(
    "withdraw: the on-chain position reports 0 vaults before withdrawing — there is nothing to release, and no baseline to verify a release against.",
  );
```

Also covers the case you flagged in `withdrawParams.ts:44-51`, where a position that exists but holds no collateral reports `vaultCount: 0`.

services/vault/e2e/real/actions/resume.ts
```diff
@@ -300,16 +277,18 @@ async function runResumeFlow(
...
+  // amount. Either way it only confirms the row rendered — the on-chain assert below is the real one.
   await assertActivatedAndOnDashboard(
     page,
     log,
     ctx.config.peginAmountBtc,
     knownTxid ?? prePeginTxid,
-    expectedVaults,
   );
+
+  await assertVaultCountRose(ctx, vaultsBefore, expectedVaults);
```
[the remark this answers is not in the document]
gbarkhatov: Valid, and pre-existing rather than introduced here — `expectedVaults = ctx.config.split ? 2 : 1` predates this PR. Leaving this thread open rather than resolving it, because it is a real footgun that is not fixed yet.

The two directions differ in severity:

- **`--split` on a single-vault batch** — the step machine waits for a second activation that never arrives, so the run burns to its 2.5h budget. Loud, if slow.
- **A split batch resumed without `--split`** — the quieter one. `assertVaultCountRose` polls for `>= before + 1`, so two activations satisfy a target of one and the run passes having verified less than it should.

On the suggested fix: deriving cardinality from the rendered rows has a failure mode of its own. If a sibling row has not rendered when we count, we would derive `1` for a split batch and silently weaken the assertion — trading a visible problem for an invisible one, which is the exact trade this PR has been undoing. Rejecting a mismatch before the click is the safer half of the suggestion, but it still depends on that same count being reliable.

So I would rather fix it deliberately than quickly: it wants a real resume run to validate, and this PR's runs are already spent. Happy to take it as a follow-up issue, or in this PR if you would prefer it not ship as-is — your call.

gbarkhatov:
Heads-up for @jonybur, since this changes something you explicitly approved.

Your review called the split cross-check's switch to a row-count delta "the right call". @jrwbabylonlab then showed that delta cannot fail: `/vaults` renders an optimistic row per just-activated vault from in-memory state (`ActivatingVaultsContext`, 90s TTL, no indexer) under the same `vault-row-<vaultId>` testid as a real row — so counting rows after driving N activations just measures the N rows those activations created.

A verification run caught it in the act. At the moment the row check ran:

```
row: "V 0.01000001 sBTC Activating collateral... V vault-provider-0 Withdraw"
```

So the delta is gone. Peg-in and resume now assert on-chain via `getPosition` (`assertVaultCountRose`), closing the real gap — peg-in was the only action without an on-chain post-condition, which is what let a UI check stand in for verification in the first place. `--withdraw-all` likewise now asserts `vaultCount === before - released` exactly, instead of "collateral fell at all".

Re-verified end to end on signet + Sepolia:

- split peg-in → `✅ On-chain vaults rose: 0 → 2 (+2, expected +2)`
- `--withdraw-all` → `✅ On-chain vaults fell: 2 → 0 (released 2)`

One of @jrwbabylonlab's threads is deliberately left open: resume derives its expected vault count from the `--split` flag rather than the selected batch. Pre-existing, and I would rather fix it with its own verification run than rush it into this PR.

jonybur (dismissed):
The src-side changes are the shape the E2E TEST HOOKS rule calls for: five one-line testids restored with pointer comments, plus the ENABLE_V3_UI env var and orphan copy blocks left behind by #2184. The three verification-time bug fixes in the writeup (dialog-scoped readPrePeginTxid, row-count delta instead of the guaranteed-to-fail txid match, counting rows only after list load) are genuine catches. The open jrwbabylonlab thread about deriving expectedVaults from ctx.config.split is a real pre-existing footgun as gbarkhatov acknowledged; worth a separate fix but not a block.

services/vault/e2e/pages/AppShell.ts
```diff
@@ -11,6 +10,15 @@
...
 import type { Locator, Page } from "@playwright/test";
 
+/** Sidebar section ids, mirroring `src/config/v3Navigation.ts`. */
+export type NavSection =
+  | "overview"
+  | "vaults"
+  | "loans"
+  | "activity"
+  | "liquidations"
+  | "explore";
```
jrwbabylonlab: would be good if we can make it a variable in the production code and re-use it here.
this is so that when we create a new nav tab, we don;t need to remember we had to adjust the test here as well

services/vault/e2e/pages/AppShell.ts
```diff
@@ -11,6 +10,15 @@
 
 import type { Locator, Page } from "@playwright/test";
 
+/** Sidebar section ids, mirroring `src/config/v3Navigation.ts`. */
```
jrwbabylonlab: why there is v3 keywords?
we already removed v1, v2. no?

jrwbabylonlab (dismissed):
small nits

services/vault/e2e/pages/AppShell.ts
gbarkhatov: Good call — done. The hand-copied union is gone; it now imports the production type:

```ts
import type { V3NavItemId } from "@/config/v3Navigation";
```

So adding a nav section can't leave this file silently stale: a new id is accepted automatically, and a renamed one fails to compile. It's a **type-only** import, so nothing from `src` is pulled in at runtime — `e2e/pages` already sits in `tsconfig.lib.json` with the `@/*` alias, and the import is erased before Playwright ever resolves it.

One place I deliberately did **not** apply this: `e2e/real/actions/navigation.ts` keeps its own `SECTION_PATHS`. `e2e/real` builds under `tsconfig.e2e.json`, which has no `@/` alias and includes only `e2e/real` — and unlike the type above, the paths are needed as runtime *values*, so importing them would drag `featureFlags` and `COPY` (and their Vite `import.meta.env` assumptions) into a plain `tsx` process. Happy to revisit if you'd rather share those too, but it needs more than an import.

services/vault/e2e/pages/AppShell.ts
gbarkhatov: Fair question, and the answer is that the "v3" is production naming, not something this PR introduced — the config file is literally `src/config/v3Navigation.ts`, exporting `V3NavItemId`, `V3_NAV_ITEMS`, `isV3SectionEnabled`, `getVisibleV3NavGroups`. My comment was just pointing at it.

You're right that it's vestigial though. #2184 deleted the v2 shell, so there is no v2 left to distinguish from. The footprint is wider than this file:

```
22 files, 107 references
v3Navigation.ts, V3NavItemId, V3_NAV_ITEMS, V3_NAV_GROUPS, isV3SectionEnabled,
getVisibleV3NavGroups, getFlagDisabledV3SectionPaths, V3NavLinks,
V3MobileNavigation, V3ModalShell
```

That's a mechanical but repo-wide rename touching router, `usePageTitle`, sidebar, modals and their tests — none of it E2E. Folding it into a test-harness PR would bury the actual change and force a re-review of files that have nothing to do with it, so I'd rather not do it here.

The stale wording I *did* own is gone: the `NavSection` copy and its "mirroring v3Navigation.ts" comment were removed by the type import in the sibling thread. Happy to open a follow-up issue for the production rename if you want it tracked.

jrwbabylonlab (approved):

services/vault/e2e/pages/AppShell.ts
gbarkhatov: Filed as https://github.com/babylonlabs-io/babylon-toolkit/issues/2257 with the full scope, so this doesn't get lost.

Mapped the footprint first. It's 9 files with `v3` in the name and ~25 identifiers, the biggest being `V3ModalShell` (45 refs / 13 files), `ActivityRowV3` (17), `isV3SectionEnabled` (15), `UtxoSplitSelectorV3` (12). Two need an actual naming decision rather than a mechanical rename — `V3ModalShell` (`ModalShell` is ambiguous next to core-ui's `Dialog`/`FullScreenDialog`) and `EmptyState`'s `variant?: "v3" | "compact"`, where both branches are live so `"v3"` needs a shape-describing name.

The issue also carries a **do-not-touch list**, which turned out to be the important half. A naive `s/V3//` would hit several unrelated version axes:

- **`vaultCoreVersion` / `txGraphVersion`** — 276 refs. The on-chain axis selecting the Bitcoin tx graph (v1 = 2 outputs; v2 = 3 outputs + 240-sat P2A anchor, TRUC). Includes ABI name strings where a rename breaks the `eth_call` at runtime, not compile time.
- **`offchainParamsVersion`** — pinned per vault for timelocks and council keys. `fetchPositions.ts:152` selects it **by name in a GraphQL selection set**, and `protocol-params-reader.test.ts:315` has literal `const v1/v2/v3 = {...}` locals, where the replace produces a syntax error.
- **`CHAINLINK_AGGREGATOR_V3_ABI`** → `CHAINLINK_AGGREGATOR__ABI`. **`ledger_btc_v2`** is a persisted wallet id string.
- **SVG path data** — `V` is vertical-lineto, so `V35`/`V34` appear in inline icons and would silently change rendered shapes. Bech32 literals contain `v3` runs, so a case-insensitive replace can mutate a Bitcoin address.

Also worth flagging separately: several comments are now actively wrong rather than just noisy — `AppSidebar.tsx:38` says the sidebar "only renders under v3", `NotificationCard.tsx:28` says "when v3 becomes the default", and `src/dev/demoDeposit.ts:836-850` has six **user-visible** god-mode labels naming the deleted v2 UI.

One correction while I was in there: there is no Aave v3 in this repo — the integration targets `aave-v4` (0 hits for `aave.*v3`), so that naming shouldn't be "cleaned up" either.

Issue #2252, opened 2026-08-13
Ledger resume: source maxAcceptableCommissionBps from the PegInSubmittedV2 event (replace interim proxy)


jrwbabylonlab:
## Context
#2220 Part 2 rebuilds `DepositTerms` on a resume-broadcast so a Ledger wallet can re-approve cross-device. Every field is recoverable from chain/indexer/WASM **except** the depositor's per-deposit commission ceiling `maxAcceptableCommissionBps`: it's a `submitPeginRequest` argument that is bound-checked once and discarded (`btc-vault PeginLogic.sol:207-208`) — never stored, never emitted. (It is NOT `minVpCommissionBps` (protocol floor) nor the stored `vaultProviderCommissionBps` (VP's actual) — those are different values that would rebuild a different `commissionFee`.)

## Interim (this issue tracks removing it)
The resume rebuild sources the bps through a single helper — `services/vault/src/services/vault/resolveMaxAcceptableCommissionBps.ts` — which currently returns the stored `vaultProviderCommissionBps` as a proxy, marked with a TODO pointing here.

**Why the proxy is safe for Part 2:** `commissionFee` is not load-bearing for the Pre-PegIn signature, does not enter DERIVE_CONTEXT_HASH, and is in no byte-verification gate — it only feeds the device's parse-time dust gate (`commission_fee >= DUST`) + on-screen display. The flow is feature-flagged and pre-mainnet, so no user hits the interim. Caveat: the stored actual can fall below the dust floor for a very small pegin — test the interim with a normal-sized deposit.

[...]

PR #2253, opened 2026-08-14
Feat/vault resume terms rebuild

jrwbabylonlab:
- Part 2 of #2220. On resume-broadcast an intent (Ledger) wallet must re-approve the Pre-PegIn, but the in-memory DepositTerms from preparePegin are gone after a browser switch / private mode. This rebuilds them from chain + WASM only — never browser storage.
- rebuildDepositTermsCore (ts-sdk, pure): fails closed at four choke points — Gate 0 (funded tx hashes to the on-chain prePeginTxHash), auth-anchor OP_RETURN completeness (vout === siblingCount), independent bounds on the WASM sizing outputs, Gate 1 (per-sibling HTLC value + scriptPubKey byte-match) — then projects via buildDepositTerms.
- rebuildDepositTerms (vault): stamped-version chain reads + sibling discovery by on-chain prePeginTxHash, with contiguity, batch-wide PENDING, and sibling-homogeneity asserts (stamps/provider/application/commission must match the target — the fields Gate 1 can't see), plus a connected-wallet ownership check.
- resolveFundedTxFeeAndUtxos: mempool-only prevout resolution; the same record feeds the broadcast PSBT, so the fee the device bounds is the fee it signs.
- handleBroadcast branches on supportsDepositApproval; the software-wallet path is byte-identical to before.
- Commission ceiling uses an interim proxy (vaultProviderCommissionBps) until the contract emits maxAcceptableCommissionBps — #2252.

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
@@ -42,7 +42,7 @@ export interface RebuildDepositTermsCoreInput {
   vaultCoreVersion: number;
   /** Sibling HTLCs ordered by htlcVout; index === htlcVout (asserted by the app). */
   siblings: readonly RebuildSibling[];
-  /** Funded Pre-PegIn tx hex — already hash-verified vs on-chain prePeginTxHash by the app (Gate 0). */
+  /** Funded Pre-PegIn tx hex. Gate 0 (hash vs prepeginTxid) is SELF-verified below — callers need not pre-verify. */
   fundedPrePeginTxHex: string;
 
   // Stamped-version participant data (already resolved + sorted by the app).
--- a/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
@@ -1,0 +1,188 @@
+/**
+ * Pure core of the resume-broadcast DepositTerms rebuild (#2220 Part 2).
+ *
+ * Takes plain chain-derived data (the app orchestrator does the chain reads +
+ * sibling discovery) and: (1) asserts sibling completeness against the funded
+ * tx's auth-anchor OP_RETURN, (2) recomputes the amount-independent sizing
+ * (depositorClaimValue, peginMaxFee, anchor) via WASM, (3) byte-matches each
+ * HTLC output's value + scriptPubKey against the funded tx (Gate 1), then (4)
+ * projects into DepositTerms. No chain access, no browser state — unit-testable.
+ *
+ * btc-vault is the protocol source of truth (`compute_min_htlc_value`,
+ * `derive_challengers_for`). Design: `todo/ledger/2220-part2-resume-rebuild-design.md`.
+ */
+
+import {
+  computeMinClaimValue,
+  computeMinPeginFee,
+  getPrePeginHtlcConnectorInfo,
+  peginP2aAnchorOutput,
+} from "@babylonlabs-io/babylon-tbv-rust-wasm";
+import { Transaction } from "bitcoinjs-lib";
+
+import { findAuthAnchorOpReturn } from "../managers/pegin/assertAuthAnchorOpReturn";
+import { assertEncodedHtlcOutputsMatch } from "../primitives/psbt/assertWasmPeginSizing";
+import { stripHexPrefix } from "../primitives/utils/bitcoin";
+import { calculateBtcTxHash } from "../utils/transaction/btcTxHash";
+
+import { buildDepositTerms } from "./buildDepositTerms";
+import type { DepositTerms } from "./depositTerms";
+
+/** One HTLC in the shared Pre-PegIn tx, ordered by (and contiguous in) htlcVout. */
+export interface RebuildSibling {
+  /** x-only or 0x-prefixed hex hashlock, per-vault (feeds the HTLC scriptPubKey). */
+  hashlock: string;
+  /** btc-vault `pegin_amount` for this sibling (satoshis). */
+  amount: bigint;
+}
+
+export interface RebuildDepositTermsCoreInput {
+  /** Stamped tx-graph version (NOT chain-active). */
+  vaultCoreVersion: number;
+  /** Sibling HTLCs ordered by htlcVout; index === htlcVout (asserted by the app). */
+  siblings: readonly RebuildSibling[];
+  /** Funded Pre-PegIn tx hex — already hash-verified vs on-chain prePeginTxHash by the app (Gate 0). */
+  fundedPrePeginTxHex: string;
+
+  // Stamped-version participant data (already resolved + sorted by the app).
+  depositorBtcPubkey: string;
+  vaultProviderBtcPubkey: string;
+  vaultKeeperBtcPubkeys: readonly string[];
+  universalChallengerBtcPubkeys: readonly string[];
+
+  // Version-locked scalars.
+  protocolFeeRate: bigint;
+  minPeginFeeRate: bigint;
+  councilQuorum: number;
+  councilSize: number;
+  timelockPegin: number;
+  timelockAssert: number;
+  timelockRefund: number;
+
+  // Pass-through into DepositTerms.
+  prepeginTxid: string;
+  /** Funded-tx fee (Σin − Σout), computed by the app; the device's `prepegin_max_fee` bound. */
+  prepeginMaxFee: bigint;
+  maxAcceptableCommissionBps: number;
+
+  /** WASM network descriptor for scriptPubKey derivation. */
+  network: Parameters<typeof getPrePeginHtlcConnectorInfo>[0]["network"];
+}
+
+export async function rebuildDepositTermsCore(
+  input: RebuildDepositTermsCoreInput,
+): Promise<DepositTerms> {
+  const siblingCount = input.siblings.length;
+  if (siblingCount === 0) {
+    throw new Error(
+      "rebuildDepositTermsCore: at least one sibling is required",
+    );
+  }
+  if (input.prepeginMaxFee <= 0n) {
+    throw new Error(
+      `rebuildDepositTermsCore: prepeginMaxFee must be > 0, got ${input.prepeginMaxFee}`,
+    );
+  }
+
+  // Gate 0 (self-verified, not merely trusted from the caller): the funded tx
+  // bytes must hash to the on-chain-pinned prepeginTxid. Everything below trusts
+  // these bytes — without this a substituted tx that replicates the first N HTLC
+  // outputs passes Gate 1 while carrying a different fee/txid the device would
+  // then be asked to approve. Mirrors the refund flow's Gate 0.
+  const expectedTxid = stripHexPrefix(input.prepeginTxid).toLowerCase();
+  const actualTxid = stripHexPrefix(
+    calculateBtcTxHash(input.fundedPrePeginTxHex),
+  ).toLowerCase();
+  if (actualTxid !== expectedTxid) {
+    throw new Error(
+      `Funded Pre-PegIn tx hashes to ${actualTxid}, expected ${expectedTxid} ` +
+        `(on-chain prepeginTxid). Resume refused.`,
+    );
+  }
+
+  // Completeness anchor: the funded tx's auth-anchor OP_RETURN sits at
+  // vout === HTLC count. Production pegins always commit exactly one anchor, so
+  // its absence (or an ambiguous set) MUST reject — this is the only guard
+  // against an indexer-lagged partial sibling set (Gate 1 only inspects 0..N-1).
+  const found = findAuthAnchorOpReturn(
+    stripHexPrefix(input.fundedPrePeginTxHex),
+  );
+  if (found === undefined || found.vout !== siblingCount) {
+    throw new Error(
+      `Auth-anchor OP_RETURN ${
+        found === undefined ? "absent or ambiguous" : `at vout ${found.vout}`
+      } does not match sibling count ${siblingCount}; sibling set is incomplete. ` +
+        `Resume refused.`,
+    );
+  }
+
+  // Amount-independent sizing (btc-vault: DCV + minPeginFee depend on the signer
+  // set + rates, not the pegin amount). Depositor-as-claimer:
+  // numLocalChallengers === numVks (VP excluded) — graph.rs derive_challengers_for.
+  const numVks = input.vaultKeeperBtcPubkeys.length;
+  const numUcs = input.universalChallengerBtcPubkeys.length;
+  const depositorClaimValue = await computeMinClaimValue(
+    input.vaultCoreVersion,
+    numVks,
+    numUcs,
+    input.councilQuorum,
+    input.councilSize,
+    input.protocolFeeRate,
+  );
+  const peginMaxFee = await computeMinPeginFee(
+    input.vaultCoreVersion,
+    numVks,
+    numUcs,
+    input.minPeginFeeRate,
+  );
+  const anchor = await peginP2aAnchorOutput(input.vaultCoreVersion);
+  const anchorValue = anchor?.value ?? 0n;
+
+  // Gate 1: per sibling, expected HTLC value = amount + DCV + peginMaxFee + anchor
+  // (btc-vault compute_min_htlc_value), and expected scriptPubKey from the
+  // sibling's on-chain hashlock. assertEncodedHtlcOutputsMatch byte-matches both
+  // against the funded tx's outputs 0..N-1.
+  const expectedHtlcValues = input.siblings.map(
+    (s) => s.amount + depositorClaimValue + peginMaxFee + anchorValue,
+  );
+  const expectedHtlcScriptPubKeys = await Promise.all(
+    input.siblings.map(async (s) => {
+      const connector = await getPrePeginHtlcConnectorInfo({
+        txGraphVersion: input.vaultCoreVersion,
+        depositorPubkey: input.depositorBtcPubkey,
+        vaultProviderPubkey: input.vaultProviderBtcPubkey,
+        vaultKeeperPubkeys: [...input.vaultKeeperBtcPubkeys],
+        universalChallengerPubkeys: [...input.universalChallengerBtcPubkeys],
+        hashlock: stripHexPrefix(s.hashlock),
+        timelockRefund: input.timelockRefund,
+        network: input.network,
+      });
+      return connector.scriptPubKey;
+    }),
+  );
+  const fundedOutputs = Transaction.fromHex(
+    stripHexPrefix(input.fundedPrePeginTxHex),
+  ).outs.map((o) => ({ value: o.value, script: o.script }));
+  assertEncodedHtlcOutputsMatch(
+    fundedOutputs,
+    expectedHtlcValues,
+    expectedHtlcScriptPubKeys,
+  );
+
+  return buildDepositTerms({
+    vaultCoreVersion: input.vaultCoreVersion,
+    protocolFeeRate: input.protocolFeeRate,
+    timelockPegin: input.timelockPegin,
+    timelockAssert: input.timelockAssert,
+    timelockRefund: input.timelockRefund,
+    prepeginTxid: input.prepeginTxid,
+    prepeginMaxFee: input.prepeginMaxFee,
+    vaultProviderBtcPubkey: input.vaultProviderBtcPubkey,
+    vaultKeeperBtcPubkeys: input.vaultKeeperBtcPubkeys,
+    universalChallengerBtcPubkeys: input.universalChallengerBtcPubkeys,
+    maxAcceptableCommissionBps: input.maxAcceptableCommissionBps,
+    peginAmounts: input.siblings.map((s) => s.amount),
+    depositorClaimValue,
+    peginMaxFee,
+  });
+}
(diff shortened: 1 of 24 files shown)


jeremy-babylonlabs (dismissed):

services/vault/src/services/vault/resolveMaxAcceptableCommissionBps.ts
```diff
@@ -0,0 +1,25 @@
...
+ * of the Pre-PegIn signature, DERIVE_CONTEXT_HASH, or any byte-verification gate — only
+ * the device's parse-time dust gate + on-screen display use it. The flow is feature-
+ * flagged and pre-mainnet, so no user hits the interim. Caveat: the stored actual can dip
+ * below the device dust floor for a very small pegin; the emitted ceiling removes that.
+ */
+export function resolveMaxAcceptableCommissionBps(vault: {
+  vaultProviderCommissionBps: number;
+}): number {
+  // TODO(#2252): return vault.maxAcceptableCommissionBps once emitted on-chain.
+  return vault.vaultProviderCommissionBps;
```
gbarkhatov: **Missing trust-boundary validation on this RPC read, and the proxy is strictly tighter than the fresh path.** Two related things, both cheap to fix here.

**1. No `minVpCommissionBps` floor check.** This returns the chain-read `vaultProviderCommissionBps` verbatim; the only downstream guard is `buildDepositTerms`' `[0, 10000)` range check. The sibling path for the *identical field* — `prepareSigningContext` in `vaultPayoutSignatureService.ts:190-206` — validates it against `[max(minVpCommissionBps, MIN_REALIZABLE_VP_COMMISSION_BPS), VP_COMMISSION_BPS_EXCLUSIVE_MAX)` with the comment "Trust-boundary check on the VP commission read from chain", precisely because it is an RPC read feeding a signing decision. A mis-decoded `0` here yields `commissionFee: 0` on every vault in the terms the device is asked to approve. `rebuildDepositTerms` already holds `offchainParams` at the point it calls this, so threading `minVpCommissionBps` in costs nothing.

**2. The ceiling is tighter than the one the depositor already approved.** The fresh path uses `capMaxAcceptableCommissionBps` = quoted bps + 25 headroom (`PeginManager.ts:141-144`); this returns the raw stored bps. The device rejects `commissionFee < 546` (`envelope.ts:152`). At 250 bps that means fresh needs `peginAmount >= 19,855` while resume needs `>= 21,840` — a deposit in that window is approved at submit time and then permanently unbroadcastable on the Ledger. The JSDoc caveat above acknowledges this but frames it as a display concern; it is actually a hard device gate.

Exporting and applying `capMaxAcceptableCommissionBps` on this path too would make both paths compute an identical ceiling from the same stamped bps, remove the asymmetry entirely, and not have to wait on #2252. It would also make the cross-path golden meaningful for production values — right now that test feeds `quoted + 25` into the rebuild, so nothing covers the value this function actually returns.

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -360,14 +371,42 @@ export function useVaultActions(): UseVaultActionsReturn {
...
-          signPsbt: (psbtHex: string) => btcWalletProvider.signPsbt(psbtHex),
-        },
-        depositorBtcPubkey,
-        expectedUtxos,
-      });
+      // Intent (Ledger) resume: rebuild the DepositTerms from chain + WASM and
+      // run the derive→approve ceremony before signing. Prevouts are resolved
+      // once, mempool-only (never the local cache), so the fee the device
+      // approves is the fee the broadcast signs. Software wallets unchanged.
+      if (supportsDepositApproval(btcWalletProvider)) {
```
gbarkhatov, line 366: `expectedUtxos` is still computed at lines 296-299 and then discarded on this branch. `utxosToExpectedRecord` throws on a malformed stored UTXO (bad txid, non-integer value, bad scriptPubKey), so a corrupt localStorage record now aborts a resume that deliberately never reads local UTXOs — the one path where the corruption is irrelevant. Moving that computation into the `else` branch fixes it and makes the intent path's "mempool-only" claim literally true.

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -360,14 +371,42 @@ export function useVaultActions(): UseVaultActionsReturn {
...
+        const { expectedUtxos: resolvedUtxos, fundedTxFee } =
+          await resolveFundedTxFeeAndUtxos(unsignedTxHex);
+        const depositTerms = await rebuildDepositTerms({
```
gbarkhatov, line 369: `rebuildDepositTerms` re-reads the same vault record this function already fetched: `getVaultFromChain(vaultId)` at line 220 and again at `rebuildDepositTerms.ts:242`, plus `assertVaultCoreVersionSupported` at line 224 and again at `rebuildDepositTerms.ts:246`.

That's two extra RPC round-trips on the pre-approval path and a window where the two reads can disagree. Both sites fail closed so it isn't a safety bug, but passing `onChainVault` (and the already-verified version) into `rebuildDepositTerms` removes both the latency and the divergence.

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -360,14 +371,42 @@ export function useVaultActions(): UseVaultActionsReturn {
...
+          fundedPrePeginTxHex: unsignedTxHex,
+          connectedDepositorAddress: depositorEthAddress as Hex,
+          depositorBtcPubkey,
+          fundedTxFee,
+        });
+        await broadcastPrePeginTransaction({
+          unsignedTxHex,
+          btcWalletProvider: {
+            signPsbt: (psbtHex: string) => btcWalletProvider.signPsbt(psbtHex),
+            deriveContextHash: (appName: string, context: string) =>
```
gbarkhatov: This shim defeats an actionable guard the seam deliberately added. `ensurePrePeginTermsApproval` checks `typeof wallet.deriveContextHash !== "function"` and throws "A deposit-approval wallet must also implement deriveContextHash, but this one does not" (`prePeginApproval.ts:110-114`). Because this always defines the property — while `supportsDepositApproval` only probes `approveDepositTerms` — that check can never fire, and such a wallet gets a raw `TypeError` from inside `deriveVaultRoot` mid-ceremony instead.

`forwardDepositApproval` on the next line exists for exactly this reason; a matching conditional spread for `deriveContextHash` would preserve the typed error. `useDepositFlow.ts:818-819` has the same shape, so this is a pre-existing pattern rather than something the PR introduced — but the fix belongs next to the new branch.

services/vault/src/services/vault/resolveFundedTxFee.ts
```diff
@@ -0,0 +1,48 @@
...
+  fundedPrePeginTxHex: string,
+): Promise<{ expectedUtxos: UtxoRecord; fundedTxFee: bigint }> {
+  const cleanHex = fundedPrePeginTxHex.startsWith("0x")
+    ? fundedPrePeginTxHex.slice(2)
+    : fundedPrePeginTxHex;
+  const tx = Transaction.fromHex(cleanHex);
+
+  const expectedUtxos: UtxoRecord = {};
+  let totalInputValue = 0n;
+  for (const input of tx.ins) {
```
gbarkhatov: **Serialized prevout resolution, and the fee is only lower-bounded here.**

The `await` inside the loop makes this N sequential mempool round-trips before the device prompt appears — a 10-input funded tx is 10 serial RTTs of dead time. Results are keyed by outpoint so order is irrelevant; `Promise.all` over `tx.ins` is a drop-in.

Separately, `fundedTxFee` is checked only for `> 0`. `MAX_REASONABLE_FEE_SATS` does get enforced before the device sees anything — `broadcastPrePeginTransaction` calls `createPsbtFromTransaction` (line 320) before `ensurePrePeginTermsApproval` (line 328), and `validateInputOutputBalance` runs unconditionally inside `addInputsToPsbt` — so there is no live exposure. But this module computes the value that becomes the device's `prepegin_max_fee` bound, and the invariant currently depends on call ordering in a different file. An upper bound here would make it self-contained.

While you're in the test file: there's no case covering a `fetchUTXOFromMempool` rejection propagating out.

services/vault/src/services/vault/rebuildDepositTerms.ts
```diff
@@ -0,0 +1,279 @@
...
+          `(${siblings.map((s) => s.htlcVout).join(", ")}); resume refused.`,
+      );
+    }
+  }
+}
+
+/**
+ * Discover the complete sibling set sharing this Pre-PegIn tx, ordered by
+ * htlcVout. Membership is on-chain `prePeginTxHash` equality (a stale indexer
+ * can only enumerate ids); completeness is backstopped by the core's OP_RETURN
```
gbarkhatov, line 137: This claim — "completeness is backstopped by the core's OP_RETURN anchor check" — is the load-bearing assumption of the whole discovery step, and it's the one thing the test suite doesn't exercise end-to-end. `rebuildDepositTerms.test.ts` mocks `rebuildDepositTermsCore` away, and `rebuildDepositTermsCore.test.ts` builds its sibling vector by hand; neither ever puts a truncated `fetchVaultIdsByDepositor` result in front of the real core.

A single test with a lagging indexer (2 siblings on chain, 1 returned) reaching the unmocked core would prove the sentence, and would fail loudly if the anchor check is ever loosened.

services/vault/src/services/vault/rebuildDepositTerms.ts
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/**
+ * Read everything version-locked to the vault's stamps (mirrors
+ * `prepareSigningContext`): offchain params + timelocks at
+ * `offchainParamsVersion`, rosters at their stamped versions, operation keys
+ * at the vault's frozen epochs.
+ */
+async function readStampedVaultContext(vaultId: Hex, target: OnChainVaultData) {
+  const protocolParamsReader = await getProtocolParamsReader();
```
gbarkhatov: ~8 sequential awaits here, most of them mutually independent: offchain params, timelock, keeper roster, UC roster, the VP genesis pubkey, and the key epochs don't depend on each other. Combined with the duplicate reads noted on the caller, that's a lot of serial RTT before the Ledger prompt appears. `prepareSigningContext` already `Promise.all`s its comparable stage — same treatment would apply cleanly.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
```diff
@@ -0,0 +1,215 @@
...
+  network: Parameters<typeof getPrePeginHtlcConnectorInfo>[0]["network"];
+}
+
+export async function rebuildDepositTermsCore(
+  input: RebuildDepositTermsCoreInput,
+): Promise<DepositTerms> {
+  const siblingCount = input.siblings.length;
+  if (siblingCount === 0) {
+    throw new Error(
+      "rebuildDepositTermsCore: at least one sibling is required",
```
gbarkhatov, line 79: Minor: no guard on non-positive `siblings[i].amount`, where the fresh path has one (`assertWasmPeginSizing` asserts `peginAmount <= 0` before the same value reaches `buildDepositTerms`).

To be clear about severity — Gate 1 does catch it. A mis-decoded `amount` shifts `amount + DCV + fee + anchor` away from the funded output's value, and Gate 0 pins that output to the on-chain hash, so the mismatch is unavoidable. So this is error attribution rather than a hole: today the failure surfaces as "HTLC output[i] value does not match the cross-checked htlcValue", which points at the funded tx when the actual bad input was the chain-read amount. Two lines next to the existing sibling-count guard would name the real culprit and keep the two paths symmetric.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
```diff
@@ -0,0 +1,215 @@
...
+    calculateBtcTxHash(input.fundedPrePeginTxHex),
+  ).toLowerCase();
+  if (actualTxid !== expectedTxid) {
+    throw new Error(
+      `Funded Pre-PegIn tx hashes to ${actualTxid}, expected ${expectedTxid} ` +
+        `(on-chain prepeginTxid). Resume refused.`,
+    );
+  }
+
+  // Completeness anchor: the single auth-anchor OP_RETURN sits at vout === HTLC
```
gbarkhatov, line 110: This hard-requires exactly one auth-anchor OP_RETURN, but the anchor is modelled as optional elsewhere: `peginOutputCount(vaultCount, hasAuthAnchor)` takes it as a flag, and `findAuthAnchorOpReturn`'s own JSDoc calls out "legacy non-auth-anchored Pre-PegIn" as a supported `undefined` case.

If any PENDING vault predates the anchor, its Ledger resume is permanently refused — with the message "sibling set is incomplete", which misdiagnoses the cause and would send someone hunting the indexer. Two things worth doing: confirm no such vaults exist (pre-mainnet plus the feature flag suggests not), and split the error so `found === undefined` reads as "this Pre-PegIn carries no auth-anchor commitment" rather than as an incomplete sibling set. The `found.vout !== siblingCount` branch is the one that genuinely means incomplete.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * Takes plain chain-derived data (the app orchestrator does the chain reads +
+ * sibling discovery) and: (1) asserts sibling completeness against the funded
+ * tx's auth-anchor OP_RETURN, (2) recomputes the amount-independent sizing
+ * (depositorClaimValue, peginMaxFee, anchor) via WASM, (3) byte-matches each
+ * HTLC output's value + scriptPubKey against the funded tx (Gate 1), then (4)
+ * projects into DepositTerms. No chain access, no browser state — unit-testable.
+ *
+ * btc-vault is the protocol source of truth (`compute_min_htlc_value`,
+ * `derive_challengers_for`). Design: `todo/ledger/2220-part2-resume-rebuild-design.md`.
```
gbarkhatov: Two doc nits in this file.

**Dangling design-doc pointer.** `todo/ledger/2220-part2-resume-rebuild-design.md` isn't in the repo — `/todo` is gitignored (`.gitignore:57`), and no other source file in the tree cites a path under it. `rebuildDepositTerms.ts:12` carries the same reference. Either commit the doc or drop the pointer; a citation no reviewer can open doesn't earn its line.

**`RebuildSibling.hashlock` is documented as "x-only or 0x-prefixed hex"** (line 34). "x-only" is pubkey terminology — a hashlock is a 32-byte SHA-256 digest, and there's no even/odd-Y notion to be x-only about. The intended meaning is presumably "32-byte hex, 0x prefix optional". This has already shipped into `docs/api/deposit-terms.md` in this PR, so fixing it at the source regenerates cleanly.

services/vault/src/hooks/deposit/__tests__/useVaultActions.test.ts
```diff
@@ -1205,3 +1217,128 @@ describe("useVaultActions — handleActivation hashlock source", () => {
     expect(mockLoggerError).not.toHaveBeenCalled();
   });
 });
+
+// Placed last: these tests override the useChainConnector mock's return value,
```
gbarkhatov: This comment documents a landmine rather than removing one — the next person appending a describe to this file gets a leaked `useChainConnector` mock and a confusing failure, with nothing in the compiler or test runner to warn them.

Restoring the default return value in this file's top-level `beforeEach` (or an `afterEach` here) makes the position irrelevant and lets the comment go away.

gbarkhatov (dismissed):
Approving — the cryptographic core is careful and the test story is strong. Comments are follow-ups, none blocking except possibly the first.

**What I checked:** all 76 tests across the six touched suites pass, and `tsc --noEmit` on `services/vault` is clean at `1685583`.

**What's genuinely good here**

- The layered fail-closed design reuses already-hardened primitives (`findAuthAnchorOpReturn`, `assertEncodedHtlcOutputsMatch`, `MAX_REASONABLE_PEGIN_VBYTES`) rather than reimplementing them. The OP_RETURN-at-`vout === siblingCount` rule is doing double duty: it's the completeness anchor against a lagging indexer *and* the upper bound on Gate 1, which by itself only checks outputs `0..N-1`.
- `PeginManager.rebuildDepositTerms.test.ts` is the strongest JS-side proof available for this: production `createPrePeginTransaction` vs. the standalone WASM entry points, full deep-equal with zero exclusions, both graph versions, asymmetric VK/UC counts and deliberately distinct scalars so a swapped argument can't hide in a symmetry.
- On cross-device resume the intent path is arguably *safer* than the software path: `verifyResumeParticipantKeys` is skipped without a local record, but Gate 1 byte-matches the chain-epoch-derived participant keys against the funded tx's HTLC scripts — an independent check the software path never gets.
- `assertSiblingBatchHomogeneous` targets exactly the fields Gate 1 structurally cannot see, and the reasoning is written down rather than assumed.

**Design points worth a reply rather than a code change**

- *Mempool as the trust anchor on the intent path.* On same-device resume `resolveFundedTxFeeAndUtxos` replaces `pendingPegin.selectedUTXOs` even when that record is available and provably tied to the local tx. The rationale is documented and the failure mode is a bad signature rather than loss (taproot sighash commits to all prevout amounts), but it is a real reduction in defense-in-depth versus the software path. An alternative that keeps both properties: resolve from mempool as now, and when a local record exists cross-check and refuse on mismatch — the device's data stays chain-derived, and the local anchor survives.
- *Error copy.* CLAUDE.md puts user-visible strings in `copy.ts`. These throws land in `broadcastError` verbatim: "Sibling vaults of this Pre-PegIn disagree on offchainParamsVersion (9 vs 3); the batch cannot be described by one set of deposit terms. Resume refused." Consistent with the existing inline throws in this hook so not a regression, but it is depositor-facing text that reads like a stack trace. Worth deciding whether to map to `copy.ts` and log the detail.
- *#2110 scope.* The comment at `useVaultActions.ts:433` correctly notes `broadcastError` is a string by contract, so when #2110 adds a typed `isDepositTermsRejectedError` branch to `mapDepositError` this path will bypass it. Make sure #2110's scope explicitly includes the resume surface, not just the inline deposit flow.

**Minor, no comment thread needed:** `rebuildDepositTermsCore.test.ts:88` uses `network: "signet" as never` while the golden describe in the same file correctly uses `as Network` — worth making consistent.

services/vault/src/services/vault/resolveMaxAcceptableCommissionBps.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useVaultActions.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useVaultActions.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useVaultActions.ts
jrwbabylonlab: fixed

services/vault/src/services/vault/resolveFundedTxFee.ts
jrwbabylonlab: fixed

services/vault/src/services/vault/rebuildDepositTerms.ts
jrwbabylonlab: fixed

services/vault/src/services/vault/rebuildDepositTerms.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/__tests__/useVaultActions.test.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/rebuildDepositTermsCore.ts
jrwbabylonlab: Fixed the message split. Kept the hard anchor requirement — it's the only sibling-completeness backstop, and a pre-anchor vault can't reach this path (the builder has emitted the anchor unconditionally since April, and this branch only runs for the Ledger wallet, which shipped after).

services/vault/src/services/vault/vaultPayoutSignatureService.ts
```diff
@@ -54,6 +54,31 @@ const VP_COMMISSION_BPS_EXCLUSIVE_MAX = 10_000;
...
+  minVpCommissionBps: number,
+): void {
+  const minCommissionBps = Math.max(
+    minVpCommissionBps,
+    MIN_REALIZABLE_VP_COMMISSION_BPS,
+  );
+  if (
+    !Number.isInteger(bps) ||
+    bps < minCommissionBps ||
+    bps >= VP_COMMISSION_BPS_EXCLUSIVE_MAX
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

jeremy-babylonlabs (approved):

PR #2254, opened 2026-08-14
ci(release): retry manifest verify probe to absorb release CDN propagation (#253)

mpastecki:
## Problem

The first monitored-env deploy after #2168 landed went red: [run 31775552723](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/31775552723/job/94691275337), `s3_publish (vault-testnet)`, step *Publish checksum manifest as an immutable per-sha release*.

The publish itself **succeeded** — the failure was the action's own verification probe:

- `06:24:48Z` — `gh release create fm-vault-vault-testnet-23d78882…` succeeds; API shows the release published with both assets `state: uploaded`
- `06:24:49Z` — the single `curl -sfIL` HEAD probe of the public download URL fails → `::error::manifest not retrievable` → exit 1
- moments later — the same URL serves 200 (and the checker has since downloaded the manifest)

The public `releases/download/…` URL is eventually consistent, and this repo has Immutable Releases enabled which adds a finalization pass after create returns — so the asset can 404 on the CDN for a few seconds even though the upload succeeded. The fail-loud branch (built for drafts / permission errors / numbering gaps) can't distinguish *not yet visible* from *not there* with a single probe.

## Fix

Retry the post-create probe for up to 30s (6×5s) before failing. Still fail-closed: an asset that genuinely never appears reds the job exactly as before.

Deliberately **not** changed: the pre-create probe and the create-race re-probe stay single-shot. A transient false 404 there is self-correcting (create fails on the existing tag → falls through to the loud branch), and same-sha runs are already serialized by the per-(workflow, env, sha) concurrency group, so the seconds-after-create window doesn't apply to them in practice.

## Impact of the original failure

None beyond the red run: the S3 upload steps completed before the manifest step, and the slot-1 manifest is published and retrievable. Attempt 2 of the run re-published into the sanctioned `-r2` slot.

### DIFF
--- a/.github/actions/publish-integrity-manifest/action.yml
+++ b/.github/actions/publish-integrity-manifest/action.yml
@@ -39,9 +39,19 @@ runs:
                --title "Frontend manifest ${PREFIX}" \
                --notes "sha256 manifest for the frontend integrity checker (devops-backlog#253). Do not delete." \
                --latest=false 2>&1); then
-            curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
-            echo "published ${TAG}"
-            exit 0
+            # The public download URL is eventually consistent — Immutable
+            # Releases finalizes after create returns, so the asset can 404
+            # for a few seconds even though the upload succeeded. Give the
+            # CDN a bounded window before declaring the publish broken.
+            for _ in 1 2 3 4 5 6; do
+              if curl -sfIL -o /dev/null "$URL"; then
+                echo "published ${TAG}"
+                exit 0
+              fi
+              sleep 5
+            done
+            echo "::error::manifest not retrievable at ${URL} 30s after create"
+            exit 1
           fi
           # Failed create: re-probe the public URL the checker reads. Only a
           # slot PUBLISHED by a concurrent run is a race worth advancing past —


mpastecki:
Second data point: the rerun of run 31775552723 (attempt 2) reproduced the exact same race on the next slot — `fm-…-r2` published at `07:21:37Z`, single probe failed at `07:21:38Z` ("manifest not retrievable"), and the same URL served 200 shortly after. Both assets `state: uploaded`.

So this is 2-for-2, not an occasional blip: the public download URL reliably lags release creation by at least a couple of seconds. Every monitored-env deploy will keep failing its own verification until this retry lands.

Note for whoever reruns: GitHub reruns execute the workflow at the run's original sha, so re-running 31775552723 again would use the old single-probe action, fail a third time, and mint another slot. The way to green main is to merge this PR — the merge commit's own push run exercises the fixed action.

.github/actions/publish-integrity-manifest/action.yml
```diff
@@ -39,9 +39,19 @@ runs:
...
                --latest=false 2>&1); then
-            curl -sfIL -o /dev/null "$URL" || { echo "::error::manifest not retrievable at $URL"; exit 1; }
-            echo "published ${TAG}"
-            exit 0
+            # The public download URL is eventually consistent — Immutable
+            # Releases finalizes after create returns, so the asset can 404
+            # for a few seconds even though the upload succeeded. Give the
+            # CDN a bounded window before declaring the publish broken.
+            for _ in 1 2 3 4 5 6; do
+              if curl -sfIL -o /dev/null "$URL"; then
```
gbarkhatov, line 47: **The bounded window isn't actually bounded — no `--max-time`.**

`curl` applies no overall timeout by default (`--connect-timeout` alone defaults to 300s), and `-L` means the stall can happen on the redirect to `objects.githubusercontent.com` rather than on github.com. A hanging CDN turns the intended 30s budget into minutes — six times over — which undercuts the premise of the change.

```bash
if curl -sfIL --max-time 10 -o /dev/null "$URL"; then
```

All three probes in this script use the same flags against the same `$URL`, so a helper defined once above the slot loop fixes it everywhere and dedupes the call sites:

```bash
probe() { curl -sfIL --max-time 10 -o /dev/null "$1"; }
```

.github/actions/publish-integrity-manifest/action.yml
```diff
@@ -39,9 +39,19 @@ runs:
...
+                echo "published ${TAG}"
+                exit 0
+              fi
+              sleep 5
+            done
+            echo "::error::manifest not retrievable at ${URL} 30s after create"
```
gbarkhatov, line 53: **The retry budget is stated in three places, and one of them is already wrong.**

Walking the loop: probes land at t≈0, 5, 10, 15, 20, 25, and the 6th `sleep 5` buys nothing — it only delays the error to t≈30. So `30s after create` describes elapsed wall clock, not the last thing actually observed (25s).

The count (`1 2 3 4 5 6`, L46), the interval, and the `30s` string can each drift independently. Hoisting the constants and computing the number keeps them in sync — and since you need an index anyway, an echo per failed attempt is worth adding: right now a transient 404 is completely silent unless all six fail, so there's no evidence in the logs to tell whether 30s is generous or marginal.

```bash
VERIFY_ATTEMPTS=6
VERIFY_INTERVAL_SECS=5
...
for attempt in $(seq 1 "$VERIFY_ATTEMPTS"); do
  if curl -sfIL --max-time 10 -o /dev/null "$URL"; then
    echo "published ${TAG}"
    exit 0
  fi
  echo "manifest not yet visible at ${URL} (attempt ${attempt}/${VERIFY_ATTEMPTS})"
  [ "$attempt" -lt "$VERIFY_ATTEMPTS" ] && sleep "$VERIFY_INTERVAL_SECS"
done
echo "::error::manifest not retrievable at ${URL} after ${VERIFY_ATTEMPTS} probes over $(( (VERIFY_ATTEMPTS - 1) * VERIFY_INTERVAL_SECS ))s"
```

The trailing `[ … ] && sleep` is safe under `set -euo pipefail` here (same shape as L32) — I checked. Note that six probes then span 25s, not 30s, so if you want the window to genuinely be 30s, make it `VERIFY_ATTEMPTS=7`.

gbarkhatov (approved):
Approving — the diagnosis is well-evidenced and the fix keeps the fail-closed contract: the only new success path still requires a real 200 from the URL the checker actually reads. Leaving the pre-create probe and the create-race re-probe single-shot is the right call, for the reasons in the description.

Two non-blocking notes inline (a missing timeout, and the retry budget being restated three times).

**One follow-up worth filing separately — not for this PR.** The description is right that a single probe "can't distinguish *not yet visible* from *not there*", and the retry widens the window without ever making that distinction: a genuinely broken publish still burns 30s and produces the same ambiguous error. `gh release view "$TAG" --json isDraft,assets` would supply it, and the evidence from run 31775552723 is that the signal was already available — the API reported the release published with both assets `state: uploaded` at the moment the probe 404'd.

Worth being precise about what it buys on *this* branch, though: we only reach it when `gh release create` exited 0, so drafts and permission failures land in the other branch, not here. What it actually adds is (a) fail-fast on the narrow "release exists but an asset never uploaded" case, and (b) the confidence to *wait longer* when the API says the asset is there — turning a fixed budget into "keep polling while the API insists it exists", which is a better contract than guessing a number.

jonybur (approved):
Fix is narrowly scoped and preserves the fail-closed contract. The retry loop only widens the success path for a probe that already caught a create-succeeded case, and the pre-create/create-race branches stay single-shot as the description reasons through. gbarkhatov's approval covers the substantive observations; nothing to add.

jrwbabylonlab (approved):

PR #2255, opened 2026-08-14
feat(ledger-vault-signer): vendor the PSBTv2 map layer with golden gates

jrwbabylonlab:
First building block for Ledger signing (#2219). To talk to the Ledger device, we need code that packages a Bitcoin transaction into the exact byte format the device expects. Rather than writing that from scratch, this copies two battle-tested files from Ledger's own official client library into our signer package (with our small fixes, each documented in the file headers).


packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/buffertools.ts
```diff
@@ -0,0 +1,165 @@
...
+
+// 2^53 − 1 (Number.MAX_SAFE_INTEGER) = 0x001f_ffff_ffff_ffff in 8-byte LE: any
+// value with byte 7 non-zero or byte 6 above 0x1f exceeds the safe range.
+const MAX_SAFE_INTEGER_TOP_BYTE = 0x1f;
+
+export function unsafeTo64bitLE(n: number): Buffer {
+  // we want to represent the input as a 8-bytes array
+  if (n > Number.MAX_SAFE_INTEGER) {
+    throw new Error("Can't convert numbers > MAX_SAFE_INT");
+  }
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/scripts/gen_vectors.py
```diff
@@ -0,0 +1,262 @@
...
+    text = path.read_text().strip()
+    hexes = json.loads(text) if path.suffix == ".json" else [text]
+    out = []
+    for h in hexes:
+        h = h.strip()
+        if not h.lower().startswith("70736274ff"):
+            print(f"  SKIP (not a PSBT, raw tx reference): {path.name}")
+            return []
+        out.append(h)
+    return out
```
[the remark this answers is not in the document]
jrwbabylonlab: Leaving as-is: every current fixture is homogeneous, the whole-file skip is printed and recorded in the committed index, and the vector diff is reviewed on every regeneration — per-entry skipping would silently include partial files instead.

jeremy-babylonlabs (dismissed):

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/buffertools.test.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Boundary tests for the vendored 8-byte-LE amount codec.
+ *
+ * PSBT amounts ride `unsafeTo64bitLE`/`unsafeFrom64bitLE` (PSBT_OUT_AMOUNT and
+ * witness-utxo values), and the pair silently caps at 2^53 − 1 — pin the bound
```
gbarkhatov: Stale after cc1552b8. This says the pair "silently caps at 2^53 − 1", but `unsafeTo64bitLE` now throws on anything that isn't a non-negative safe integer and `unsafeFrom64bitLE` throws above the bound — which is exactly what the four tests below assert. Suggest "rejects out-of-range input from both directions".

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/psbtv2.golden.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  input_maps: MapEntry[][];
+  output_maps: MapEntry[][];
+  global_commitment: Commitment;
+  input_commitments: Commitment[];
+  output_commitments: Commitment[];
+};
+
+// deposit-flow__pegin__0 is a stale pre-v22 signet capture (1-in/2-out; the
+// current validator requires 1-in/3-out) — fine for these byte-level goldens.
+const vectors = [
```
gbarkhatov, line 68: This array is the only place all 22 vectors are enumerated, and it's worth reusing: `merkleMap.golden.test.ts` still imports only the three vectors that pre-existed on `main` (`claimer_payout__2`, `depositor_graph__0`, `pegin__0`). The 19 added in this PR each carry `input_commitments`, `output_commitments`, `inputs_maps_root`, `outputs_maps_root` and `sign_psbt_cdata_hex` — the fields that actually pin the SIGN_PSBT wire format — and nothing asserts them.

Concretely, the merkleization gate currently sees only 2-in/3-out, 2-in/2-out and 1-in/2-out. This PR adds 3-in/1-out (eight of them), 2-in/4-out and 1-in/3-out. Pointing that test's imports at this same list is a near-free coverage win and the highest-value follow-up I found.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/psbtv2.ts
```diff
@@ -0,0 +1,641 @@
...
+  }
+  private readKeyPair(map: Map<string, Buffer>, buf: BufferReader): boolean {
+    const keyLen = sanitizeBigintToNumber(buf.readVarInt());
+    if (keyLen === 0) {
+      return false;
+    }
+    const keyType = buf.readUInt8();
+    const keyData = buf.readSlice(keyLen - 1);
+    const value = buf.readVarSlice();
+    set(map, keyType, keyData, value);
```
gbarkhatov: Duplicate keypairs are silently last-wins here — BIP-174 forbids them, and `set()` just overwrites.

Worth flagging because the monorepo already parses PSBTs with `bitcoinjs-lib` 6.1.7, whose `bip174` rejects them outright (`Format Error: Keys must be unique for each input: …`, `bip174/src/lib/parser/fromBuffer.js:85`). So the same bytes are a hard parse error on the validation side and a silently-normalized map on the path that merkleizes for the device.

Nothing consumes this yet and it's upstream-faithful, so deferring is fine — but it shouldn't survive into the `merkelizedPsbt` increment without either a duplicate check or an explicit entry in the vendor README's "Known upstream behaviour (kept)" list.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/psbtv2.ts
```diff
@@ -0,0 +1,641 @@
...
+      this.inputMaps[i] = inputMap;
+      while (this.readKeyPair(inputMap, buf));
+    }
+    for (let i = 0; i < nOutputs; i++) {
+      const outputMap: Map<string, Buffer> = new Map();
+      this.outputMaps[i] = outputMap;
+      while (this.readKeyPair(outputMap, buf));
+    }
+
+    this.normalizeToV2();
```
gbarkhatov, line 400: `deserialize` never asserts the buffer was fully consumed, so trailing bytes after the last output map are silently dropped and the PSBT re-serializes to something different from what came in.

The asymmetry is visible inside this PR: the test helper `splitMaps` in `psbtv2.golden.test.ts` *does* enforce `reader.available() !== 0`, but the production path doesn't.

Upstream behaviour, so keeping it is defensible — but given the README's hostile-PSBT framing, it belongs either in the "Known upstream behaviour (kept)" list next to the overlong-varint note, or as a ninth documented modification.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/psbtv2.ts
```diff
@@ -0,0 +1,641 @@
...
+  REDEEM_SCRIPT = 0x04,
+  WITNESS_SCRIPT = 0x05,
+  BIP32_DERIVATION = 0x06,
+  FINAL_SCRIPTSIG = 0x07,
+  FINAL_SCRIPTWITNESS = 0x08,
+  PREVIOUS_TXID = 0x0e,
+  OUTPUT_INDEX = 0x0f,
+  SEQUENCE = 0x10,
+  TAP_KEY_SIG = 0x13,
+  TAP_BIP32_DERIVATION = 0x16,
```
gbarkhatov, line 62: `psbtIn` stops at `TAP_BIP32_DERIVATION`, but the fixtures this PR commits carry `0x15` TAP_LEAF_SCRIPT and `0x17` TAP_INTERNAL_KEY — see the input-0 keys `15c150929b…` and `17` in `deposit-flow__claimer_payout__0.json`. `0x14` TAP_SCRIPT_SIG and `0x18` TAP_MERKLE_ROOT are absent too.

Nothing is broken: the map layer passes them through byte-identically and the golden tests prove it. But `getInputEntriesOfType` (planned in the README) will need these constants, and CLAUDE.md forbids inlining them as magic numbers — so they're worth adding alongside that consumer.

Related, on the class doc at L88-90: it says taproot script-path spending is "not implemented", which reads alarmingly given script-path is exactly our case. That's accurate for the typed accessor surface only, not for the map-level pass-through we actually rely on. Worth a one-line clarification, since this file is on a critical path and the sentence is inherited verbatim.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/psbtv2.golden.test.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  it.each(cases)("normalizes the v0 file and the oracle's v2 bytes of $id to identical canonical bytes", ({ v }) => {
+    const fromV0 = new PsbtV2();
+    fromV0.deserialize(Buffer.from(v.psbt_hex, "hex"));
+    const fromV2 = new PsbtV2();
+    fromV2.deserialize(Buffer.from(v.psbt_v2_hex, "hex"));
+    expect(fromV0.serialize().toString("hex")).toBe(fromV2.serialize().toString("hex"));
+  });
+
+  // Cross-check entry 7: the v0 pegin fixtures reach deserialize with ZERO-entry
```
gbarkhatov: Two small things on this comment:

1. It overstates what the test below does. The test only asserts the *pre-normalization* output maps are empty; the "normalizeToV2 must synthesize AMOUNT/SCRIPT" half is actually proven by the first golden test, which checks the normalized `03`/`04` entries against the oracle for these same two fixtures. Suggest "…pins the pre-normalization shape; the synthesis itself is asserted by the first test above." (This also means CodeRabbit's nitpick asking for synthesis coverage is already satisfied — the comment is what's off, not the coverage.)
2. "Cross-check entry 7" points at an external checklist a reader of this file can't resolve. Let's keep those references out of code and test descriptions.

packages/babylon-ledger-vault-signer/scripts/gen_vectors.py
```diff
@@ -0,0 +1,262 @@
...
+    """Fixture format (tests/vectors/README.txt): .json = JSON array of PSBT hexes,
+    .txt = a single hex. 70736274ff = PSBT; 02000000 = finalized raw tx (skip)."""
+    text = path.read_text().strip()
+    hexes = json.loads(text) if path.suffix == ".json" else [text]
+    out = []
+    for h in hexes:
+        h = h.strip()
+        if not h.lower().startswith("70736274ff"):
+            print(f"  SKIP (not a PSBT, raw tx reference): {path.name}")
+            return []
```
gbarkhatov: `return []` discards everything collected so far, so a fixture whose first entries are PSBTs and whose Nth entry is a raw tx would be dropped wholesale and recorded under `skipped_raw_tx_fixtures` — the gate silently shrinks behind a SKIP print.

Every current fixture is homogeneous, so nothing is lost today. `continue` would be more robust, with the print naming the batch index rather than just `path.name`.

gbarkhatov (dismissed):
Approving. The vendored layer is well-gated and the provenance work is the part I'd hold up as the model: pinned oracle (`ledger-bitcoin==0.4.0`), firmware fixture commit `8f99b8b` recorded, generator committed with self-validating asserts (`cdata == expected`, `p2 == CURRENT_PROTOCOL_VERSION`). Verified locally at cc1552b8: 223 tests pass, eslint / `tsc -b --noEmit` / prettier all clean, and nothing under `src/vendor/` is reachable from `src/index.ts`.

The four behavioural modifications all reduce risk, and each is recorded in the file header *and* THIRD-PARTY-NOTICES: `unsafeTo64bitLE` rejecting negative/NaN/fractional (upstream let negatives two's-complement wrap and NaN encode as zero — in an amount codec), `readInt32LE`→`readUInt32LE`, the version-probe `catch` narrowed to `NoSuchEntry`, and the `getRequired`/`getOptional` split removing non-null assertions. The `localeCompare` → byte-lexicographic comparator change is correct and necessary — locale-sensitive ordering of keys that get merkleized is a real hazard.

All inline comments are non-blocking. The one I'd most like to see land in this PR is widening `merkleMap.golden.test.ts` to the full 22-vector list: the 19 vectors added here ship commitment roots and `sign_psbt_cdata_hex` that nothing currently asserts.

Smaller notes not worth inlining:

- All 22 vectors are `original_psbt_version: 0`, so the native-v2 `deserialize` branch rests on the third golden test (which re-parses `psbt_v2_hex`) plus the ported upstream v2 round-trip. Adequate, but worth stating.
- `index.json` is listed as a generator output but isn't committed. That's probably the right call — it records `str(fixtures_root)`, an absolute local path — but the docstring could say so, so nobody commits it later.
- `as unknown as SignPsbtVector[]` gives up the free `resolveJsonModule` shape check on the vector files. There's no clean fix, though: JSON modules infer `string[][]`, and relaxing `MapEntry` to `string[]` would break the `new Map(...)` calls at L141/L157. Fine as-is.
- The typed accessors that will eventually carry value into the intent ceremony (`getOutputAmount`, `getInputWitnessUtxo().amount`) are unpinned. That's consistent with the README's land-with-consumer discipline for `getInputEntriesOfType`, so no action now — but they should get fixture-backed assertions in the same PR as their first consumer.
- Process note: this package still has no `SECURITY_MODEL.md` despite being critical path #7 in CLAUDE.md, so repo-level `SECURITY.md` applies by fallback. Pre-existing, but the package is growing into the signing path.

jonybur (dismissed):
Approving. The provenance discipline in this vendored layer is the strongest part: pinned oracle (`ledger-bitcoin==0.4.0`), firmware fixture commit `8f99b8b` recorded, generator committed with self-validating asserts (`cdata == expected`, `p2 == CURRENT_PROTOCOL_VERSION`). The two documented `psbtv2.ts` modifications - dropping the taproot-broken `fromBitcoinJS`, and unifying `serializeMap`'s comparator to byte-lexicographic order matching the device's Merkle expectations - are both correct and each is recorded in the file header and THIRD-PARTY-NOTICES.

Spot-checked: `unsafeTo64bitLE`/`unsafeFrom64bitLE` correctly reject negatives, NaN, fractionals, out-of-bounds, and short buffers. Length varints go through `sanitizeBigintToNumber(readVarInt())` with bounded downstream `readSlice`, so the parse-accept-overlong / emit-canonical stance is safe. For lowercase-hex Map keys, JS string code-unit compare is provably equivalent to byte-lexicographic order, so the comparator behaves as claimed. No log statements on payload bytes.

Coverage: the third `it.each` ("normalizes the v0 file and the oracle's v2 bytes to identical canonical bytes") does close the pegin gate on `normalizeToV2` synthesizing AMOUNT/SCRIPT into empty output maps - since `pegin0` and `generatedPegin0` are both in `vectors`, the assertion runs on them and would fail if synthesis regressed. The second-map round-trip is byte-identity against `psbt_v2_hex` in the oracle's recorded order, which is the right layering: the vendored serializer's byte-lex order is closed separately by `varint.json` + the ported upstream round-trips.

The vendor dir is unreachable from `src/index.ts` today, tree-shaken from the published bundle, and excluded from declaration emission, as the README states. Nothing here ships to consumers until the next increment lands its consumer.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/buffertools.test.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/psbtv2.ts
jrwbabylonlab: fixed — deserialize now rejects repeated keypairs, matching the bip174 side.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/psbtv2.ts
jrwbabylonlab: fixed — full-consumption assert added as a listed modification.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/psbtv2.golden.test.ts
jrwbabylonlab: fixed — both points.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/psbtv2.golden.test.ts
jrwbabylonlab: This lands in the stacked follow-up PR (base = this branch): a MerkelizedPsbt golden asserts the per-map commitments, maps roots and cdata over all 22 vectors — extending this file's test would be immediately superseded by it.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/psbtv2.ts
jrwbabylonlab: Agreed — they land with their consumer (getInputEntriesOfType) in the follow-up increment.

packages/babylon-ledger-vault-signer/scripts/gen_vectors.py
jrwbabylonlab: Took your version — per-entry continue with the batch index printed, and partially-skipped files now recorded per-entry in the accounting. Regeneration verified byte-identical.

gbarkhatov (approved):

PR #2259, opened 2026-08-14
Feat/ledger sign psbt b1b

jrwbabylonlab:


packages/babylon-ledger-vault-signer/scripts/gen_command_traces.py
```diff
@@ -0,0 +1,507 @@
...
+    root = tree.root
+    assert root in interp.known_trees
+
+    runner = TraceRun(interp)
+    resp = runner.run(
+        "GET_MERKLE_LEAF_PROOF",
+        leaf_proof_request(root, SYNTHETIC_N, 0),
+        f"SYNTHETIC 129-leaf tree, leaf 0: proof depth 8 -> 6 hashes in-message "
+        f"(floor((255-32-1-1)/32)), 2 spill to the queue as 32-byte elements",
+    )
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/vectors/command-traces/deposit-flow__claimer_payout__2.json
```diff
@@ -0,0 +1,380 @@
...
+   "root": "1facb95d5e66ea706686b2ed37acc0c47e01746a1b6b5fe996cbf5b45f7a40d4",
+   "n_leaves": 3
+  }
+ ],
+ "traces": [
+  {
+   "command_name": "GET_MERKLE_LEAF_PROOF",
+   "request_hex": "41519b38dae74447b72151f354cb138ca3591a5ff8ac813289b18a004e313216200500",
+   "response_hex": "fcf0a6c700dd13e274b6fba8deea8dd9b26e4eedde3495717cac8408c9c5177f0303583c7dfb7b3055d99465544032a571e10a134b1b6f769422bbb71fd7fa167a5d4b8c129ed14cce2c08cfc6766db7f8cdb133b5f698b8de3d5890ea7ff7f0a8d195811f41d3d5c58240be155bb7d1dcb8f47add7e3417c24e1d52d41653013926",
+   "notes": "global_keys: leaf 0 of 5; response = leaf_hash(32) | proof_len(1) | n_in_msg(1) | hashes (empty proof for a single-leaf tree)"
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed — the suffix is now emitted only for single-leaf trees; traces regenerated, all request/response bytes verified identical (notes-only diff).

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
```diff
@@ -31,23 +32,40 @@ v6-compatible source and gate it with golden vectors.
...
-### Planned (not yet vendored)
+Vendored so far: `varint.ts`, `merkle.ts`, `merkleMap.ts`, `buffertools.ts`,
+`psbtv2.ts`, `merkelizedPsbt.ts`, `clientCommands.ts`, `policy.ts` — the full
+#2219 vendoring closure. These modules are referenced only by their
+golden-vector and unit tests, not by any production entrypoint (nothing
+reachable from `src/index.ts` imports them), so the bundler tree-shakes them
+out of the JS bundle and the vendor dir is excluded from declaration emission
+(see `vite.config.ts`) — the published `dist/` ships none of the vendored code
+(no JS, no `.d.ts`), and the `bitcoinjs-lib` / `buffer` deps they pull in stay
+out of the bundle, until the SIGN_PSBT task (#2219 B1-d) wires them in.
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/scripts/gen_vectors.py
```diff
@@ -0,0 +1,266 @@
...
+    out = []
+    n_skipped = 0
+    for i, h in enumerate(hexes):
+        h = h.strip()
+        if not h.lower().startswith("70736274ff"):
+            print(f"  SKIP (not a PSBT, raw tx reference): {path.name}[{i}]")
+            n_skipped += 1
+            continue
+        out.append((i, h))
+    return out, n_skipped
```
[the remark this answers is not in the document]
jrwbabylonlab: Keeping per-entry skip: the whole-file behavior was reversed by the human review on #2255 (per-entry with printed batch index), and regeneration was verified byte-identical there. Your recorded learning predates that reversal.

jonybur (dismissed):
S4 of the SIGN_PSBT stack. Refactor splits the DMK sender into a non-throwing raw seam plus a shared status-word classifier; the throwing ceremony sender rebases on that, preserving behavior byte-for-byte. Bit-identical status-word mapping is verified by the extracted `classifyStatusWord` and its table tests; the pass-through of 0xE000 and 0x6A80 as data on the raw seam is pinned in `dmkRawApduSender.test.ts`, so a future SIGN_PSBT loop can consume them.

Vendored `clientCommands.ts` change is scoped to an optional `onYield` validator hook, documented in the provenance header and in THIRD-PARTY-NOTICES. Golden traces (`gen_command_traces.py`) use `ledger-bitcoin==0.4.0` as the oracle - every `response_hex` is literal `ClientCommandInterpreter.execute()` output, seeded to mirror `client.py sign_psbt` (lines 293-320) with citations. Wire-spec §4/§5 cites base-app encoder files for each APDU layout. Synthetic 129-leaf tree covers the >6-hash proof-spill case that no real PSBT tree exercises.

Zero dead code: removed `MerkelizedPsbt` fields (`n_inputs`, `n_outputs`, `inputs_maps_root`, `outputs_maps_root`) and the block using them cleanly delete their now-unused imports. Debug tracing at `dmkApduSender.ts:73,79` logs only header + status word - compliant with the critical-path #7 rule against logging payload bytes. The `onYield` seam is exercised by a dedicated abort-without-recording test in `clientCommands.test.ts`.

packages/babylon-ledger-vault-signer/src/rawApdu.ts
```diff
@@ -0,0 +1,128 @@
+/**
+ * Transport seam for the SIGN_PSBT interrupt/continue loop (#2219).
+ *
+ * The ceremony's `ApduSender` throws on every status word except 0x9000;
+ * SIGN_PSBT cannot ride it, because `0xE000` (SW_INTERRUPTED_EXECUTION — a
+ * client-command request follows in the response data) and the once-eaten
+ * `0x6A80` retry are loop inputs, not failures. A {@link RawApduSender}
```
gbarkhatov: ⚠️ **This claim contradicts the pinned firmware and the reference client.**

The doc says SIGN_PSBT can't ride the throwing sender because `0xE000` *and the once-eaten `0x6A80` retry* "are loop inputs, not failures." I can't find a source for the `0x6A80` half, and three independent sources contradict it:

- `bitcoin_app_base/src/boilerplate/sw.h:30` defines `SW_INCORRECT_DATA 0x6A80` — a terminal error. `SW_INTERRUPTED_EXECUTION 0xE000` is defined separately at `sw.h:90`.
- The reference client loops on `while sw == 0xE000:` **only** (`bitcoin_client/ledger_bitcoin/client.py:182`) and raises `DeviceException` for anything non-`0x9000` after the loop (`client.py:335`).
- The vault app's own tests assert `0x6A80` as a **terminal** `sign_psbt` failure — ~20 sites across `tests/test_screen5_assert.py` and `tests/test_screen8_payout_finalize.py`, all `assert exc.value.status == SW_INCORRECT_DATA`.

I grepped the firmware and its test suite for any retry/re-send semantics and found none.

Nothing acts on this today, so it isn't a live bug — but CLAUDE.md critical path #7 requires APDU claims to cite firmware/reference-client source, and as written this will steer B1-d into retrying a signing APDU the device rejected. Please either cite the source or drop the clause. The same claim is repeated in `src/__tests__/dmkRawApduSender.test.ts:3`.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/clientCommands.ts
```diff
@@ -0,0 +1,379 @@
...
+    let leafIndex = 0;
+    let found = 0;
+    for (let i = 0; i < mt.size(); i++) {
+      if (mt.getLeafHash(i).toString("hex") === leafHashHex) {
+        found = 1;
+        leafIndex = i;
+        break;
+      }
+    }
+    return Buffer.concat([Buffer.from([found]), createVarint(leafIndex)]);
```
gbarkhatov, line 257: **Test gap where the missed mutation produces a *wrong response*, not a throw.**

Replacing `createVarint(leafIndex)` with `Buffer.from([leafIndex])` passes all 319 tests. I checked every map in all 22 fixtures: the largest has **12 entries**, so the corpus can never reach the 253 varint-width boundary, and the dup test at `:266` asserts index 0.

For a map with ≥253 keys, `Buffer.from([300])[0] === 44` — the device is told the key lives at leaf 44, then Merkle-verifies leaf 44's value against a proof for a different key. Input maps are VP-supplied and `psbtv2.deserialize` only rejects *duplicate* keypairs, so 253 distinct `TAP_BIP32_DERIVATION` / `PARTIAL_SIG` entries in one input map is trivially constructible.

The vendored code is correct (it matches upstream) — this is purely coverage. One unit test over a synthetic 300-leaf known-list asserting `01 ‖ fd2c01` closes it.

packages/babylon-ledger-vault-signer/src/__tests__/dmkRawApduSender.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    // success on the raw seam too.
+    const { handle } = handleWith({ statusCode: new Uint8Array([0x90]), data: new Uint8Array() });
+
+    await expect(createDmkRawApduSender(handle)(apdu)).rejects.toThrow(/status word/);
+  });
+
+  it("still refuses a payload that cannot fit in the single-byte Lc", async () => {
+    const { handle } = handleWith({ statusCode: new Uint8Array([0x90, 0x00]), data: new Uint8Array() });
+
+    await expect(createDmkRawApduSender(handle)({ ...apdu, data: new Uint8Array(256) })).rejects.toThrow(
```
gbarkhatov, line 76: **The boundary that matters is 255, and it's untested.**

Both suites test 256 (must throw); nothing tests 255 (must *not* throw). Mutating `encodeApdu`'s `data.length > 0xff` to `>= 0xff` (`dmkApduSender.ts:34`) passes all 319 tests.

That off-by-one is live: `deriveContextHash` sends **exactly**-255-byte continuation chunks — `vaultCommands.ts:140` does `context.subarray(sent, sent + MAX_APDU_DATA_BYTES)` with `MAX_APDU_DATA_BYTES = 255` (`intentTlv.ts:44`), and `CONTEXT_MAX_BYTES` is 1024. Any context long enough to need a full continuation chunk would fail at runtime with "Lc is a single byte".

Suggest adding the positive case next to this one: 255 bytes of data yields a 260-byte APDU with `out[4] === 0xff`.

packages/babylon-ledger-vault-signer/src/dmkApduSender.ts
```diff
@@ -93,10 +57,11 @@ function hex4(value: number): string {
 }
 
 /**
- * Build a sender bound to one device session. A `sessionId` must never outlive
+ * Build a non-throwing sender bound to one device session: every status word
+ * — 0xE000 included — comes back as data. A `sessionId` must never outlive
  * its connection — recreate the sender whenever the session changes.
  */
-export function createDmkApduSender(handle: DmkSessionHandle): ApduSender {
+export function createDmkRawApduSender(handle: DmkSessionHandle): RawApduSender {
```
gbarkhatov, line 70: **The trace path has zero test coverage, and it's the one place CLAUDE.md #7's "never log payload bytes" can be violated.**

No test in the package sets `__LEDGER_VAULT_APDU_TRACE__`. Concrete mutation the suite doesn't catch: append `` ` data=${Array.from(apdu.data).map(hex2).join("")}` `` to `header` on the next line — 319/319 still pass. That would dump the DERIVE_CONTEXT_HASH payload (the vault context preimage, which commits to funding outpoints) and the APPROVE_VAULT_INTENT key batches into the browser console.

Worth closing now specifically because this refactor moved both `console.debug` calls into the **raw** sender, so B1-d's much larger merkleized-PSBT payloads will ride this same untested line. One test that enables the flag, spies `console.debug`, and asserts the logged string contains no hex rendering of `data`.

Related, worth a code comment rather than a change: DMK 1.7.1 itself logs `{sessionId, apdu, error}` on this path and the WebHID sender logs `formatApduSentLog(apdu)` — both inert here only because `dmkSession.ts:56` never calls `.addLogger()`. Anyone adding a `ConsoleLogger` there dumps full payloads. (The `console.debug(..., error)` on the transport path is clean — I traced every reachable DMK error class and none carries APDU bytes.)

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/clientCommands.ts
```diff
@@ -0,0 +1,379 @@
...
+    this.results = results;
+  }
+
+  execute(request: Buffer): Buffer {
+    const payload = Buffer.from(request.subarray(1));
+    if (this.onYield) {
+      // Local addition (vault seam): a throw here propagates out of
+      // interpreter.execute() before the payload is recorded. The validator
+      // gets its own copy so a mutating callback cannot corrupt the record.
+      this.onYield(Buffer.from(payload));
```
gbarkhatov, line 86: **This copy — the line that makes the local addition safe — is asserted nowhere.**

It's documented in three places (file header L18-19, the comment just above, README L52 "The validator receives a copy"), but the only assertion on the argument is `clientCommands.test.ts:65`'s `toHaveBeenCalledWith(Buffer.from([0x07, 0x08]))` — deep equality, which passes just as well for the *same* object.

Dropping `Buffer.from(payload)` here keeps 319/319 green, and the documented guarantee is gone: a validator doing `payload.fill(0)`, `payload.reverse()`, or any in-place normalization now corrupts the recorded yield, and `getYielded()` returns bytes that were never on the wire.

One test whose `onYield` mutates its argument and asserts `results[0]` is unchanged.

The rest of the seam is well covered and correctly ordered — the throw lands before both `results.push` and `progressCallback`, and both the unit-level (`:68-80`) and interpreter-level (`:389-396`) abort tests pin that. Also worth noting the positional `(progressCallback?, onYield?)` signature *can't* be misused as `new ClientCommandInterpreter(myValidator)` — `(payload: Buffer) => void` isn't assignable to `() => void`, so TS catches it.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/clientCommands.ts
```diff
@@ -0,0 +1,379 @@
...
+
+    for (const cmd of commands) {
+      if (this.commands.has(cmd.code)) {
+        throw new Error(`Multiple commands with code ${cmd.code}`);
+      }
+      this.commands.set(cmd.code, cmd);
+    }
+  }
+
+  getYielded(): readonly Buffer[] {
```
gbarkhatov, line 338: **`getYielded()` hands out the live internal array, and the "discard on abort" contract is purely conventional.**

This returns `this.yielded` itself — the same array `YieldCommand` mutates. `readonly Buffer[]` is compile-time only: the array keeps growing under a caller that snapshotted it, and `(interp.getYielded() as Buffer[])` defeats it outright.

Combined with the abort rule living only in the vendor README, this failure shape is available to B1-d: the exchange loop wraps `interpreter.execute()` in a `try/catch` that maps device/transport errors to a user-facing message (a very natural shape), the per-yield assertion throws for input *k*, the catch swallows it, and the caller then reads `getYielded()` and finalizes with *k−1* signatures — **an under-signed PSBT treated as complete**, which is exactly the asymmetric-failure shape CLAUDE.md #3 warns about for depositor-graph presigning.

~4 lines makes it structural instead of conventional: set an `aborted` flag when `onYield` throws, have `getYielded()` throw while it's set, and return a frozen copy otherwise. Happy to defer to B1-d if you'd rather keep the vendored file minimal — but then the enforcement should land in the SIGN_PSBT wrapper (one interpreter per attempt), not stay in prose.

packages/babylon-ledger-vault-signer/src/rawApdu.ts
```diff
@@ -0,0 +1,128 @@
...
+  readonly sw: number;
+  readonly data: Uint8Array;
+}
+
+/**
+ * Non-throwing transport: every status word comes back as data. Throws only
+ * when there is no status word to return — transport failure, a malformed
+ * (non-2-byte) status word, or a payload that cannot fit the single-byte Lc.
+ */
+export type RawApduSender = (apdu: Apdu) => Promise<RawApduResponse>;
```
gbarkhatov, line 40: **Design note for B1-d:** this seam makes a non-`0x9000` outcome ignorable *by construction*. `const { data } = await sendRaw(apdu)` type-checks and drops `sw` entirely. Before this PR, no caller could reach a non-`0x9000` outcome without an exception; that safety property is now gone by design.

Concrete scenario: a JS port of `_make_request` writes `while (sw === 0xE000) { … }` and omits the reference client's final `if (sw != 0x9000) raise` (`client.py:335`). A user declining on the last confirm screen returns `0x6985` with empty data, the loop exits, and the caller sees "zero yielded results" rather than "user refused" — presigning then reports missing signatures instead of a decline.

Inherent to the seam being requested, and nothing here is wrong today. But the PR ships no guard — no branded result, no `assertOk` helper. Related smell: `classifyStatusWord` returning `undefined` conflates "no error" with "is `0x9000`"; a discriminated result (`{ok: true, data} | {ok: false, error}`) would make the loop's obligation explicit at the type level.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/merkelizedPsbt.ts
```diff
@@ -0,0 +1,89 @@
...
+  public inputMerkleMaps: MerkleMap[] = [];
+  public outputMerkleMaps: MerkleMap[] = [];
+  public inputMapCommitments: Buffer[];
+  public outputMapCommitments: Buffer[];
+  constructor(psbt: PsbtV2) {
+    super();
+    psbt.copy(this);
+    this.globalMerkleMap = MerkelizedPsbt.createMerkleMap(this.globalMap);
+
+    for (let i = 0; i < this.getGlobalInputCount(); i++) {
```
gbarkhatov: **The count guard is one-directional.** Too *few* maps throws `Missing input map ${i}`; too *many* are silently ignored — surplus maps aren't merkelized, aren't committed, and never reach the device, yet `psbt.copy(this)` already copied them into `this.inputMaps`, so the inherited `serialize()` would emit a PSBT that doesn't match the commitment the depositor approved.

I traced whether a hostile PSBT can reach this and **it can't today**: the v2 path creates exactly `getGlobalInputCount()` maps (`psbtv2.ts:380-393`), and the v0 path has `normalizeToV2` *overwrite* any planted `PSBT_GLOBAL_INPUT_COUNT` with `tx.ins.length` (`psbtv2.ts:419`). So it's reachable only from a programmatically assembled `PsbtV2` — i.e. the not-yet-written B1-d consumer.

Given this class is the commitment producer on a critical path, making it symmetric is cheap:

```ts
if (this.inputMaps.length !== inputCount) {
  throw new Error(`Input map count ${this.inputMaps.length} != declared ${inputCount}`);
}
```

Worth knowing the current tests can't distinguish the two policies either: changing the loop bounds to `this.inputMaps.length` keeps 44/44 green, as does deleting all three new throw sites.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/policy.ts
```diff
@@ -0,0 +1,92 @@
...
+    const m = new Merkle(keyBuffers.map((k) => hashLeaf(k)));
+
+    const buf = new BufferWriter();
+    buf.writeUInt8(WALLET_POLICY_V2); // wallet version
+
+    // length of wallet name, and wallet name
+    buf.writeVarSlice(Buffer.from(this.name, "ascii"));
+
+    // length of descriptor template
+    buf.writeVarInt(this.descriptorTemplate.length);
```
gbarkhatov, line 78: **Length and hash disagree on encoding.** `writeVarInt(this.descriptorTemplate.length)` counts **UTF-16 code units**, while the next line hashes `Buffer.from(this.descriptorTemplate)` as **UTF-8 bytes**. (Same function is also inconsistent about `Buffer.from(name, "ascii")` vs `Buffer.from(template)`.)

For any non-ASCII template the serialized length and the hashed preimage describe different byte counts, so `getId()` diverges from the device's computation.

Faithful to upstream and **unreachable via `DefaultWalletPolicy`**, whose `DefaultDescriptorTemplate` union is ASCII-only — but `WalletPolicy`'s constructor takes an arbitrary `string`, so it's reachable if B1-d ever constructs one directly. I wouldn't patch the vendored file over it; a one-line note here (or in the header's known-upstream-behaviour list, alongside the existing `parseVarint` overlong-encoding note) would be enough. It's also the reason the single-vector golden shouldn't be read as covering `serialize()` generally.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
```diff
@@ -73,9 +80,21 @@ Python client emits over the firmware's PSBT fixtures (`__tests__/vectors/signps
 each map's keys/values, the serialized key order (== `sorted_key_order`), and
 full byte-identity with `psbt_v2_hex` — the oracle serializes maps in insertion
 order, so the identity is asserted by reassembling the normalized model in the
-oracle's recorded order. Any behavioural drift from upstream fails these. The
-Python client is the primary oracle; the deprecated npm package is a secondary
-check.
+oracle's recorded order. `merkelizedPsbt`'s gate re-derives, from each
+fixture's raw `psbt_hex` through the class, every per-map commitment, the
+outer inputs/outputs maps-roots, and the reassembled SIGN_PSBT client data
+(`sign_psbt_cdata_hex`) over all 22 fixtures — superseding the
```
gbarkhatov, line 94: This sentence correctly says `merkelizedPsbt`'s gate supersedes the hand-concatenated maps-root/cdata assertions — but **L75-77 above still credits `merkleMap` with them**:

> `merkleMap`'s commitment (keys root, values root, the `varint(n) ‖ keysRoot ‖ valuesRoot` layout, **and the outer Merkle-of-maps roots**) is checked today against per-map commitment vectors…

The outer Merkle-of-maps roots were removed from `merkleMap.golden.test.ts` in this PR (the deleted `describe("SIGN_PSBT header: outer maps-roots + client-data golden vectors")` block). Dropping that parenthetical clause at L75-77 resolves the self-contradiction.

(For what it's worth, I diffed the move assertion-by-assertion — nothing was lost, and it's strictly stronger: 3 fixtures → 22, oracle-hex inputs → class-derived inputs.)

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/policy.golden.test.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Golden test for the vendored WalletPolicy serialization.
+ *
+ * `serialize()` = `0x02 ‖ varint(|name|) ‖ name ‖ varint(|template|) ‖
+ * sha256(template) ‖ varint(n_keys) ‖ keysRoot`, and `getId()` (its sha256) is
+ * the `wallet_id` field of the SIGN_PSBT header — the Pre-PegIn / #2221/#2222
```
gbarkhatov: This overstates `policy`'s role in *this* protocol, in a way that could mislead #2221/#2222.

`getId()` is **not** the `wallet_id` the vault SIGN_PSBT header carries: `scripts/gen_vectors.py:56-62` overrides `NoWalletPolicy.id` to return `b"\x00" * 32`, and all 22 `sign_psbt_cdata_hex` goldens end in 64 zero bytes — which the vendor README (L104-107) itself explains is "the routing switch into the vault validators." In no-policy mode `WalletPolicy.serialize()` is load-bearing only as preimage material for `addKnownWalletPolicy`, and per `clientCommands.golden.test.ts:13-15` no trace even queries it.

The golden is worth keeping as a vendor-fidelity check — just retitle the rationale so a future reader doesn't wire `getId()` into the header.

Minor, while you're here: with `n_keys === 1` the Merkle root is the single-leaf pass-through, so `serialize()`'s tree-combining path isn't exercised by this vector. A second 2-key vector would close that cheaply.

packages/babylon-ledger-vault-signer/scripts/gen_command_traces.py
```diff
@@ -0,0 +1,508 @@
...
+Usage:
+  <venv>/bin/python gen_command_traces.py [<vectors_dir> [<out_dir>]]
+  defaults: vectors_dir = this script's directory, out_dir = <vectors_dir>/command-traces
+"""
+
+import json
+import sys
+from pathlib import Path
+
+from ledger_bitcoin.client_command import ClientCommandCode, ClientCommandInterpreter
```
gbarkhatov, line 42: Worth hardening: this imports whatever `ledger_bitcoin` happens to be installed, but the whole provenance argument rests on the oracle being **exactly** `0.4.0` (asserted in the vendor README L73-74 and L131-132, and pinned in `scripts/requirements.txt`).

If someone regenerates in a venv with a different version, the goldens silently drift to a different oracle and the replay test still passes — it only checks self-consistency with whatever was written. A fail-fast guard costs three lines:

```python
from importlib.metadata import version
_ORACLE_VERSION = "0.4.0"
assert version("ledger-bitcoin") == _ORACLE_VERSION, (
    f"oracle must be ledger-bitcoin=={_ORACLE_VERSION}, got {version('ledger-bitcoin')}"
)
```

Same argument applies to `gen_vectors.py` if it doesn't already have one.

packages/babylon-ledger-vault-signer/scripts/gen_command_traces.py
```diff
@@ -0,0 +1,508 @@
...
+        assert entry["root"] in interp.known_trees, f"{entry['name']} not seeded"
+
+    runner = TraceRun(interp)
+    for tree in trees:
+        trace_leaf_proofs(runner, tree)
+    # index lookups on the trees the device actually queries by hash: keys trees
+    # (get_merkleized_map_value pattern: H(0x00|key) against the map's keys root)
+    named = {name: (root, elements) for name, root, elements in reg}
+    for tree_name, source_map in [("global_keys", g), ("input0_keys", ins[0]), ("output0_keys", outs[0])]:
+        root, elements = named[tree_name]
```
gbarkhatov: Nit (also flagged by CodeRabbit): `elements` is unpacked but never used — the next line only needs `root`, and `canonical` supplies the elements. `root, _ = named[tree_name]` or `root = named[tree_name][0]`.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/clientCommands.golden.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+// A regenerated-but-empty fixture set must fail loudly, not create zero tests.
+if (TRACE_INDEX.length === 0) {
+  throw new Error("command-traces index.json lists no files");
+}
+
+describe("client-command interpreter replays the Python-oracle traces byte-for-byte", () => {
+  it.each(TRACE_INDEX)("$file", ({ file: fileName, n_traces }) => {
+    const file = loadJson<TraceFile>(join(TRACES_DIR, fileName));
+    expect(file.traces.length).toBe(n_traces);
```
gbarkhatov, line 88: `index.json` records a `per_command` breakdown per file that nothing asserts — only `n_traces` is checked here. A regeneration that changed the command mix while keeping the total identical would pass silently.

Cheap to close inside the existing loop:

```ts
const counts: Record<string, number> = {};
for (const t of file.traces) counts[t.command_name] = (counts[t.command_name] ?? 0) + 1;
expect(counts).toEqual(per_command);
```

Related gap: nothing detects a trace file present on disk but missing from `index.json` — a `readdirSync(TRACES_DIR)` cross-check would make the "regenerated-but-empty must fail loudly" guard at L79-81 complete.

Two things I checked that are genuinely fine: the index is consistent with the files today (counts and file list both verified), and omitting the `_NoWalletPolicy` seeding is safe — no trace queries it, and `addKnownWalletPolicy` is separately covered by `clientCommands.test.ts:351-378`.

packages/babylon-ledger-vault-signer/src/rawApdu.ts
```diff
@@ -0,0 +1,128 @@
...
+ * the caller via {@link classifyStatusWord}, so raw-seam consumers and the
+ * throwing sender raise identical typed errors.
+ *
+ * @module ledger-vault-signer/rawApdu
+ */
+
+import { LedgerDeviceError, LedgerDeviceLockedError, LedgerUserRefusedError } from "./errors";
+
+/** One APDU in the wire form the senders serialise: CLA ‖ INS ‖ P1 ‖ P2 ‖ Lc ‖ DATA. */
+export interface Apdu {
```
gbarkhatov, line 21: Nit — the split left two copies of things that were single before:

1. This is now the canonical APDU shape, but the same five fields are still declared inline at `vaultCommands.ts:51-57` (`ApduSender`'s parameter) and `dmkApduSender.ts:33` (`encodeApdu`'s parameter). Making it `export type ApduSender = (apdu: Apdu) => Promise<Uint8Array>` stops three structurally-identical declarations drifting once SIGN_PSBT adds a field.
2. `hex2`/`hex4` are duplicated verbatim between `rawApdu.ts:77-83` and `dmkApduSender.ts:51-57`.

Also worth considering while you're here: `encodeApdu` validates `data.length` but not `cla/ins/p1/p2`. `Uint8Array` assignment is mod-256, so `0x100 → 0x00` and `-1 → 0xFF` truncate silently — and `hex2()` prints the *pre*-truncation value, so the trace log and the wire diverge exactly when you'd need the log. No current caller can produce it (every call site uses module constants), but SIGN_PSBT is the first place P1/P2 could plausibly become computed. One `assertByte()` beside the existing Lc check.

packages/babylon-ledger-vault-signer/src/dmkApduSender.ts
```diff
@@ -115,29 +80,25 @@ export function createDmkApduSender(handle: DmkSessionHandle): ApduSender {
...
+/**
+ * Build the ceremony's throwing sender: the raw sender plus the shared
+ * status-word classification (typed error on anything but 0x9000).
+ */
+export function createDmkApduSender(handle: DmkSessionHandle): ApduSender {
+  const sendRaw = createDmkRawApduSender(handle);
+  return async (apdu) => {
+    const { sw, data } = await sendRaw(apdu);
+    const error = classifyStatusWord(sw, {
+      ins: apdu.ins,
```
gbarkhatov, line 102: Nit: this context wiring is untested — swapping `ins: apdu.ins` ↔ `p1: apdu.p1`, or dropping `appVersion`, survives the full suite. Every error-path test in `dmkApduSender.test.ts` uses `p1: 0, p2: 0`, and the only test asserting `/ins 0x80 p1 0x02/` (`rawApdu.test.ts:40`) hands the context in as a literal, so it never exercises this mapping.

Diagnostic-only — a mid-ceremony `0xB007` would report `ins 0x00 p1 0x80` and point a support engineer at the wrong APDU. Two lines to close: send `{ins: 0x80, p1: 0x02}`, return `0x6f42`, assert the message; and give the `0x6E00` test's handle an `appVersion`.

gbarkhatov (dismissed):
## Approve — with one item worth acting on before B1-d

I verified the vendoring end to end rather than reading it: I fetched `clientCommands.ts`, `merkelizedPsbt.ts` and `policy.ts` from `LedgerHQ/app-bitcoin@0a9e9e1` and hashed them — **all three sha256 values match the per-file headers exactly**, and every diff hunk vs upstream is accounted for in the declared `Modifications`. No undeclared drift. I also had the command-trace goldens independently recomputed from the Ledger spec in a from-scratch `node:crypto` implementation with zero imports from this package: **181/181 responses reproduced byte-for-byte, 0 mismatches**, plus 77/77 tree roots. `pnpm test` 319/319, lint and build clean, and I confirmed the "nothing vendored ships" claim (`dist/index.js` = 16.24 kB, 11 modules, zero vendored symbols, no vendor `.d.ts`). Licensing is in good shape too — full Apache text in the shipped `THIRD-PARTY-NOTICES.md`, modifications enumerated, per-file §4(b) headers.

This is careful work on a critical path and I'm happy to approve it. Comments below are one real correction, several test gaps that let a *wrong-response* mutation through, and some doc/nit cleanup.

### The one thing I'd fix before B1-d

**`rawApdu.ts:7` states `0x6A80` is a SIGN_PSBT loop input. The pinned firmware says otherwise** — details inline. It's a comment, not code, so nothing is broken today; but it's the kind of claim B1-d will implement against.

### Please add a PR description

An empty body on a 4k-LOC critical-path PR is the biggest review-ergonomics cost here — it took a full audit to establish what is actually in scope. Even four lines would do it: vendored Ledger PSBT/client-command layer (last three modules of the #2219 closure), the raw-APDU seam split out of `createDmkApduSender`, Python-oracle command-trace goldens, and **no production SIGN_PSBT entrypoint wired yet** (nothing reachable from `src/index.ts`, nothing in `dist`). That last point is the one a reviewer most needs up front, and it's currently only discoverable from the vendor `README.md`.

### Nits I couldn't anchor inline (files/lines outside the diff)

- `package.json` — `bitcoinjs-lib` and `buffer` are `dependencies`, but no non-vendor `src/**` file imports either and neither appears in `dist/index.js`. Every consumer installs them today for test-only code. `devDependencies` until B1-d wires the vendored layer in, or soften the README claim at L61-62 which implies more isolation than exists.
- `it.each([[0x6985], [0x5501]])("classifies 0x%s …")` renders **decimal** — the test names read "classifies 0x27013". Affects `rawApdu.test.ts:20,26,36` and `dmkApduSender.test.ts:65`. `dmkRawApduSender.test.ts` got this right by passing string labels.
- `clientCommands.ts:130,199` — you named three of the five chunking constants; the `- 1` (partial-length byte) and `- 1 - 1` (`n_proof_elements ‖ n_response_elements`) headers are still inline literals. And `:167`'s `"expected at least 34 bytes"` hardcodes 34 right next to a computed `SHA256_LEN + 1 + 1` check. Low value against vendor fidelity — your call.

### Explicitly checked and **not** a problem — please don't let a later review re-raise these

- **Class-field ordering in `MerkelizedPsbt`.** `inputMerkleMaps = []` / `outputMerkleMaps = []` always run before `psbt.copy(this)` — subclass field initializers run immediately after `super()` and before the constructor body, regardless of `useDefineForClassFields` (which resolves to `true` here). No field collides with `PsbtV2`'s.
- **`merkelizedPsbt.golden.test.ts` is not tautological.** It builds `inputsRoot`/`cdata` with `Merkle`/`hashLeaf`/`createVarint` from the code under test, but every construction is compared to an **oracle constant** (`v.inputs_maps_root`, `v.sign_psbt_cdata_hex`), and its self-referential inputs (`inputMapCommitments`, `getGlobalKeysValuesRoot()`, `getGlobalInputCount()`) are each independently pinned to the oracle in the preceding `it.each`. The gate holds.
- **The `merkleMap.golden.test.ts` deletion loses nothing.** I diffed every removed assertion against its replacement: 3 fixtures → 22, and oracle-hex inputs → class-derived inputs. Strictly stronger.
- **The goldens are genuinely independent of the TS implementation.** Beyond the byte-level recomputation above, the seeding paths genuinely cross: the test seeds from `psbt_hex` through TS `PsbtV2`+`MerkelizedPsbt`, while the goldens' roots come from Python's own PSBT parse. A TS map-ordering bug throws "unknown tree" rather than silently passing.

packages/babylon-ledger-vault-signer/src/rawApdu.ts
jrwbabylonlab: The source is the base dispatcher, not the vault handlers — while a command is interrupted, any non-CONTINUE APDU gets SW_INCORRECT_DATA and the interrupted command is dropped (bitcoin_app_base dispatcher.c:107-111 at the pinned e400d8d8), so the first resend after a host-abandoned loop eats one 0x6A80. Outside that single gated resend it's terminal — agreed the wording overclaimed; tightened the doc with the citation.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/clientCommands.ts
jrwbabylonlab: added — 301-leaf synthetic tree pinning 01‖fd2c01 at the 0x42 call site.

packages/babylon-ledger-vault-signer/src/__tests__/dmkRawApduSender.test.ts
jrwbabylonlab: added in both suites

packages/babylon-ledger-vault-signer/src/dmkApduSender.ts
jrwbabylonlab: fixed — exact-string log assertions through both senders, plus a no-addLogger note beside the DMK builder.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/clientCommands.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/clientCommands.ts
jrwbabylonlab: getYielded() has no production caller — the B1-d follow-up collects through the onYield seam with a fresh interpreter per attempt, the loop rethrows the typed mismatch errors, and assertComplete gates the merge in both directions, so a swallowed throw can't finalize k−1 signatures; the public result is spread-detached. Keeping the vendored getter verbatim.

packages/babylon-ledger-vault-signer/src/rawApdu.ts
jrwbabylonlab: Agreed on the seam property. The sole consumer (the follow-up's signing loop) switches on sw exhaustively — 0xE000 continues, 0x9000 exits through the completeness assert, everything else classifies to a typed throw — and the raw sender's doc scopes it to that loop; createDmkApduSender stays the default throwing path everywhere else.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/merkelizedPsbt.ts
jrwbabylonlab: fixed — symmetric guards both directions with tests; header note updated (upstream has no guard either way).

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/policy.ts
jrwbabylonlab: Kept verbatim — upstream counts UTF-16 code units for the length while hashing UTF-8, and the Python oracle (wallet.py:70) counts UTF-8 for both, but every template any flow constructs is ASCII where they coincide. Recorded in the file header and the vendor README instead of deviating.

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/policy.golden.test.ts
jrwbabylonlab: fixed — header rewritten (zero wallet_id is the routing switch; getId() becomes the header wallet_id only on the policy path), plus a second oracle-generated 2-key vector.

packages/babylon-ledger-vault-signer/scripts/gen_command_traces.py
jrwbabylonlab: fixed — explicit version check in both generators (raise, not assert — survives python -O); regeneration verified byte-identical.

packages/babylon-ledger-vault-signer/scripts/gen_command_traces.py
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/__tests__/clientCommands.golden.test.ts
jrwbabylonlab: fixed — per-command counts asserted in the replay loop, plus a disk↔index cross-check.

packages/babylon-ledger-vault-signer/src/rawApdu.ts
jrwbabylonlab: fixed — Apdu is canonical in rawApdu, hex helpers deduped, and encodeApdu now rejects non-byte cla/ins/p1/p2.

packages/babylon-ledger-vault-signer/src/dmkApduSender.ts
jrwbabylonlab: fixed — weave test with swap-killing values; the 0x6E00 test now pins the app version too.

gbarkhatov (approved):

jonybur (approved):
Approving. CLAUDE.md critical path #7 discipline holds throughout: every TLV/APDU layout cites its firmware or reference-client source, envelope terms are validated synchronously with a typed rejection before any device I/O, and the DMK sender logs headers and status words only (no payload bytes). Golden vectors cover both merkleized PSBT bytes and the client-command interpreter's traces, including a synthetic deep tree; oracle versions are pinned and enforced by the generators. The vendored ledger-bitcoin surface stays under package-internal use with no production entrypoint exposing it, matching the README's zero-dead-code claim until B1-d wires SIGN_PSBT in. Prior threads from gbarkhatov and coderabbit on TLV citations, count-guard symmetry, and log-content assertions were addressed, and the extra 255/256 boundary and leaf-index encoding tests close the mutation gaps that were raised.

PR #2262, opened 2026-08-15
refactor(vault): move god mode behind a production-owned override seam

jeremy-babylonlabs:
Production code imported `@/dev/*` directly to read demo state. That pointed the dependency the wrong way and spread the god-mode flag check across every consumer.

`src/overrides/` now owns the seam. Each domain exposes a store built from `createOverrideStore`, which checks
`featureFlags.isGodModePanelEnabled` in one place and returns `null` everywhere else. The dev panels publish resolved whole objects into those stores; production reads them and keeps its existing merge logic unchanged. No static `@/dev/*` import remains outside `src/dev/`; the two lazy DEV-gated mounts stay as they were.

The Liquidations precedence ladder moves verbatim into `useLiquidationsPageState`, and the dead demo-withdrawal seam is deleted, so the panel loses its withdrawal item type.

### DIFF
--- a/services/vault/src/components/pages/Liquidations/useLiquidationsPageState.ts
+++ b/services/vault/src/components/pages/Liquidations/useLiquidationsPageState.ts
@@ -1,0 +1,161 @@
+import { useMemo } from "react";
+
+import {
+  BPS_SCALE,
+  MIN_HEALTH_FACTOR_FOR_BORROW,
+} from "@/applications/aave/constants";
+import { usePositionNotifications } from "@/applications/aave/hooks/usePositionNotifications";
+import type {
+  CalculatorParams,
+  CalculatorResult,
+} from "@/applications/aave/positionNotifications";
+import {
+  calculateBorrowCapacityUsd,
+  getHealthFactorStatusFromValue,
+  type HealthFactorStatus,
+} from "@/applications/aave/utils";
+import { useDashboardState } from "@/hooks/useDashboardState";
+import { useLiquidationPositionOverride } from "@/overrides/liquidations";
+import { usePositionCascadeOverride } from "@/overrides/position";
+
+export interface LiquidationsPagePosition {
+  collateralBtc: number;
+  collateralValueUsd: number | null;
+  debtValueUsd: number;
+  maxTotalDebtUsd: number;
+  healthFactor: number | null;
+  healthFactorStatus: HealthFactorStatus;
+}
+
+export interface LiquidationsPageState {
+  position: LiquidationsPagePosition;
+  params: CalculatorParams | null;
+  result: CalculatorResult | null;
+  isGodMode: boolean;
+  /** The cascade override alone, or null. Distinct from `isGodMode`, which
+   *  also flips on for a position-override-only state that never touches the
+   *  chart — the chart's collateral factor must track the cascade. */
+  cascade: { result: CalculatorResult; params: CalculatorParams } | null;
+}
+
+/**
+ * The Liquidations page's precedence ladder for its position stats — the
+ * position override, then the god-mode cascade's own numbers, then the live
+ * `useDashboardState` figures — relocated out of the page component.
+ */
+export function useLiquidationsPageState(
+  account: string | undefined,
+): LiquidationsPageState {
+  const { result: liveResult, params: liveParams } =
+    usePositionNotifications(account);
+  const {
+    collateralBtc,
+    collateralValueUsd,
+    debtValueUsd,
+    maxTotalDebtUsd,
+    healthFactor,
+    healthFactorStatus,
+  } = useDashboardState(account);
+
+  // God-mode override (dev only): manual mode drives this page's whole
+  // cascade, bypassing the connection/empty-state ladder below, the way
+  // `DashboardPage.tsx` drives its overview card. Compile-time dead in
+  // production, like the other override reads (overrides/store.ts) — nothing
+  // ever sets this outside the debug panel, which is only imported behind
+  // `import.meta.env.DEV`.
+  // A status-only override (stale price) carries no cascade to chart, so it
+  // must not put this page into god mode — it falls through to live, exactly
+  // as a live stale price would.
+  const cascadeOverride = usePositionCascadeOverride();
+  const godMode = useMemo(
+    () =>
+      cascadeOverride?.result
+        ? { result: cascadeOverride.result, params: cascadeOverride.params }
+        : null,
+    [cascadeOverride],
+  );
+
+  // Position override (dev only): a lighter-weight god-mode control that sets
+  // this page's stat cards directly, without needing a real position or a
+  // Manual Mode cascade. It wins for the stat cards even over an active
+  // cascade above — see the `position` derivation below — while the cascade
+  // (godMode/`result`) keeps driving the event cards untouched.
+  const positionOverride = useLiquidationPositionOverride();
+  const isGodMode = godMode !== null || positionOverride !== null;
+
+  const result = godMode?.result ?? liveResult;
+  const params = godMode?.params ?? liveParams;
+
+  // Position stats shown above the chart. Precedence: the position override,
+  // then the god-mode cascade's own numbers, then the live `useDashboardState`
+  // figures — a mocked stat must never render mixed with a real one, so each
+  // branch is self-contained rather than layering overrides on a live base.
+  const position = useMemo((): LiquidationsPagePosition => {
+    if (positionOverride) {
+      const {
+        collateralBtc: overrideCollateralBtc,
+        debtUsd,
+        healthFactor: overrideHealthFactor,
+      } = positionOverride;
+      // Price: the active cascade's if Manual Mode is on, else the live oracle
+      // price — `params` already resolves that precedence. No price available
+      // = an absent caption, not a fabricated "$0.00 USD".
+      // `maxTotalDebtUsd = debtUsd * HF / MIN_HEALTH_FACTOR_FOR_BORROW` mirrors
+      // `calculateBorrowCapacityUsd`: live HF = collateralValueUsd * LT /
+      // (BPS_SCALE * debtUsd), and maxTotalDebtUsd = collateralValueUsd * LT /
+      // BPS_SCALE / MIN_HEALTH_FACTOR_FOR_BORROW, so substituting collapses to
+      // this — guarded so a non-positive HF can't divide by zero or produce a
+      // negative denominator.
+      const btcPrice = params?.btcPrice ?? null;
+      return {
+        collateralBtc: overrideCollateralBtc,
+        collateralValueUsd:
+          btcPrice !== null ? overrideCollateralBtc * btcPrice : null,
+        debtValueUsd: debtUsd,
+        maxTotalDebtUsd:
+          overrideHealthFactor > 0
+            ? (debtUsd * overrideHealthFactor) / MIN_HEALTH_FACTOR_FOR_BORROW
+            : 0,
+        healthFactor: overrideHealthFactor,
+        healthFactorStatus:
+          getHealthFactorStatusFromValue(overrideHealthFactor),
+      };
+    }
+    if (!godMode) {
+      return {
+        collateralBtc,
+        collateralValueUsd,
+        debtValueUsd,
+        maxTotalDebtUsd,
+        healthFactor,
+        healthFactorStatus,
+      };
+    }
+    const { result: gmResult, params: gmParams } = godMode;
+    const { maxTotalDebtUsd: gmMaxTotalDebtUsd } = calculateBorrowCapacityUsd({
+      collateralValueUsd: gmResult.collateralValue,
+      currentDebtUsd: gmParams.totalDebtUsd,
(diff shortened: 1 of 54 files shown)


services/vault/src/overrides/deposits.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Opens for any OWNED flow-state demo deposit (the 15 flow steps + the
+ * activated terminal), so the whole Deposit Progress view can be walked with
+ * god mode. Stays inert for a different-wallet (unowned) preview and for
+ * expired deposits (those live in `expiredActivities`, not the pending list).
+ *
+ * There is no "would this auto-run real signing?" guard: for every demo id
+ * PostDepositContinuationView renders a SAFE view (read-only progress, or the
+ * simulated activation walk) — the real signing/registry branches never mount.
+ */
+export function getDemoStepperBatch(
```
gbarkhatov, line 36: `getDemoStepperBatch` now exists twice, verbatim. This copy and the original at `dev/demoDeposit.ts:1132-1144` are identical apart from the parameter rename `demo` → `override` — and the two parameter types (`ActiveDemo` / `DepositOverride`) are structurally identical too, so it really is the same function twice. Both carry their own test suites: `overrides/__tests__/deposits.test.ts:68` and `dev/__tests__/GodModePanel.test.tsx:322-389`.

Production imports only this one (`components/vaults/VaultsLifecycleSections.tsx:56`), so the dev original has zero non-test consumers. Per the zero-dead-code policy, please delete `demoDeposit.ts:1119-1144` (the function and its JSDoc) together with the `describe("getDemoStepperBatch")` block at `GodModePanel.test.tsx:322-389` and the now-unused import at `GodModePanel.test.tsx:26`.

services/vault/src/dev/demoDeposit.ts
```diff
@@ -1375,21 +1219,6 @@ export function useDemoDeposit(): ActiveDemo | null {
   );
```
gbarkhatov, line 1219: Five exports are left with zero consumers after the move:

- `useDemoDeposit` (:1206), `useDemoActivity` (:1223), `useDemoLoan` (:1239), `useDemoCollateral` (:1255)
- `isArtifactDownloadDemoEnabled` (`dev/demoArtifactDownload.ts:163`), dead since `hooks/deposit/useArtifactDownload.ts:143` switched to `getArtifactDownloadOverride()`

ESLint doesn't flag unused *exports*, which is why this passes lint.

Their docs are stale as a result: the JSDoc immediately above (:1201-1203) still says "Consumed by usePendingDeposits (render lists) and PeginPollingContext (controlled results)", but both read `useDepositOverride()` now (`usePendingDeposits.ts:47`, `PeginPollingContext.tsx:196`). The file header at :18 also `{@link}`s `getDemoStepperBatch`, which lives in `@/overrides/deposits` after this PR.

Beyond the policy, keeping these alive leaves a second read path into demo state behind a *different* gate — `import.meta.env.DEV && flagOn && enabled` here, versus the store's `featureFlags.isGodModePanelEnabled` in `overrides/store.ts:38`. Two independently-gated ways to read the same state is exactly the divergence this seam is meant to remove.

services/vault/src/components/simple/__tests__/DashboardPage.test.tsx
```diff
@@ -66,15 +66,8 @@ vi.mock("@/hooks/usePrices", () => ({
   usePrices: () => pricesMock,
 }));
 
-vi.mock("@/dev/debugPositionStore", async (importOriginal) => ({
-  ...(await importOriginal<typeof import("@/dev/debugPositionStore")>()),
-  useDebugPositionOverride: () => ({ result: null, status: null }),
-  useDebugManualMode: () => false,
-}));
-
 vi.mock("@/dev/demoDeposit", () => ({
```
gbarkhatov, line 69: Dead mock — `DashboardPage` no longer imports from `@/dev/` at all (it reads `useLiquidationCardOverride` from `@/overrides/liquidations` now), so nothing in this test resolves `@/dev/demoDeposit` and this factory never runs. Safe to drop :69-71.

services/vault/src/dev/GodModePanel.tsx
```diff
@@ -302,6 +320,38 @@ function DemoControls() {
   const [clearedReceipts, setClearedReceipts] = useState<number | null>(null);
   const borrowSymbol = useDemoBorrowSymbol();
 
+  // Publish the resolved deposit-family aggregates for the real dashboard
+  // sections to read. "Inject demo" gates all four: merely adding a mock item
+  // must not inject it until the toggle opts in.
+  useEffect(() => {
```
gbarkhatov, line 326: **This is the one I'd want fixed before merge.**

These five effects are the only thing that publishes into the deposit / activity / collateral / loan / artifact-download override stores, and they live in `DemoControls` — which renders only inside `PanelBody` (:662), which mounts only when the panel is expanded (:770) or popped out (:764). `collapsed` defaults to `true` (:672). The three child panels are in the same position: `GodModeMount` passes them as `children` into `PanelBody` (:664), so `LiquidationAnalysisDebugPanel:49,:123` and `MarketDataDebugPanel:27` are equally mount-coupled.

Concrete break: expand the panel, enable "Inject demo", add a deposit item, then **hide the panel** and click the demo card to open the deposit multistepper. `DemoActivationContent` (`dev/DemoActivationContent.tsx:49-55`) fires `submitDemoVaultActivation`, which advances the demo store to `ACTIVATION_CONFIRMING` and schedules the next advance — but with `DemoControls` unmounted nothing re-runs `setDepositOverride`, so the card and the stepper stay frozen on the pre-activation snapshot. Before this PR `usePendingDeposits` called `useDemoDeposit()` directly, which is a `useSyncExternalStore` subscription, so it advanced. Hiding the panel is the *likely* path here rather than an exotic one, since it's a `fixed` overlay competing with the modal it's being used to drive.

Teardown is asymmetric for the same reason: only `PositionNotificationsDebugPanel:495` clears its override on unmount, so collapsing the panel leaves the other overrides published but frozen, while the cascade override clears.

The narrow fix is to keep `PanelBody` mounted when collapsed — render it hidden rather than unmounting it.

The fix I'd prefer is to make `overrides/*` the **owner** rather than a mirror: have the panel's change handlers call `setXOverride(...)` directly and delete the mirrored dev-store slices. As it stands every `useDebug*Override` / `useLiquidation*Override*` / `useDemoMarketData` hook has exactly one non-test consumer — the panel component that copies it into `overrides/*` — so the dev stores are still the source of truth and `overrides/*` is a mount-coupled shadow of them. Owning the state dissolves this bug rather than patching it, and it removes the duplicate state and type surface flagged on `overrides/liquidations.ts`.

services/vault/src/overrides/liquidations.ts
```diff
@@ -0,0 +1,50 @@
...
+ *  `null` (the default, and the only value production can ever see) leaves
+ *  it derived from the live position. */
+export type LiquidationCardOverride =
+  | "no-deposit"
+  | "no-loan"
+  | "position"
+  | null;
+
+/** The `/liquidations` page's stat-card figures, set directly from the panel. */
+export interface LiquidationPositionOverride {
```
gbarkhatov, line 13: This interface is declared twice, verbatim down to its JSDoc line — here and at `dev/liquidationDebugStore.ts:65`. `LiquidationAnalysisDebugPanel.tsx:32` imports the dev one; `Liquidations/__tests__/Liquidations.test.tsx:9` imports this one.

Same pattern for the four deposit-family aggregates: `ActiveDemo`/`DepositOverride` (`demoDeposit.ts:835` vs `overrides/deposits.ts:10`), `ActiveDemoActivity`/`ActivityOverride`, `ActiveDemoLoan`/`LoanOverride` (identical down to the `debtUsd` JSDoc), `ActiveDemoCollateral`/`CollateralOverride`.

Structural typing means a field added on the dev side type-checks everywhere and is then silently dropped at the seam — no compile error, no test failure, the value just never reaches consumers. If the mirror stays, have `src/dev/` import these types from `src/overrides/` (dev → production is the allowed direction) so each shape is declared once.

services/vault/src/overrides/liquidations.ts
```diff
@@ -0,0 +1,50 @@
+import { createOverrideStore } from "./store";
+
+/** Which of the `/liquidations` card's mutually exclusive states to force.
+ *  `null` (the default, and the only value production can ever see) leaves
+ *  it derived from the live position. */
+export type LiquidationCardOverride =
+  | "no-deposit"
+  | "no-loan"
+  | "position"
+  | null;
```
gbarkhatov, line 10: Minor: `null` is baked into `T` here, and `createOverrideStore<T>` already returns `T | null` (`store.ts:16-18`), so `useLiquidationCardOverride()` ends up typed `"no-deposit" | "no-loan" | "position" | null` with the null arriving twice, and `set(null)` means two indistinguishable things. It's harmless in practice because both nulls mean "no override" — but dropping `| null` from the union would keep the store's contract as the single source of nullability. `resolveLiquidationCardState`'s `case null` still works either way.

services/vault/src/components/pages/Liquidations/useLiquidationsPageState.ts
```diff
@@ -0,0 +1,161 @@
...
+  const { result: liveResult, params: liveParams } =
+    usePositionNotifications(account);
+  const {
+    collateralBtc,
+    collateralValueUsd,
+    debtValueUsd,
+    maxTotalDebtUsd,
+    healthFactor,
+    healthFactorStatus,
+  } = useDashboardState(account);
```
gbarkhatov, line 58: `useDashboardState(account)` now runs twice on `/liquidations` — here, and still at `Liquidations/index.tsx:76` for `hasCollateral` / `hasLoans` / `collateralFactorBps` / `isLoading` / `borrowedAssets`. It was a single call before this extraction.

The underlying queries dedupe, but the effects don't: `useDashboardState` runs two that mutate shared context — `clearReorderedOrder` (`hooks/useDashboardState.ts:134-139`) and `clearActivatingVault` (`:143-152`) — plus all of its derivation work, now doubled on every render of the page.

Returning the five extra fields from this hook and deleting the call at `index.tsx:76` gets it back to one call, and keeps the coupling in one place rather than threading the result in as a parameter.

services/vault/src/components/pages/Liquidations/useLiquidationsPageState.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  // God-mode override (dev only): manual mode drives this page's whole
+  // cascade, bypassing the connection/empty-state ladder below, the way
+  // `DashboardPage.tsx` drives its overview card. Compile-time dead in
+  // production, like the other override reads (overrides/store.ts) — nothing
```
gbarkhatov, line 63: Nit, but worth fixing because this comment is load-bearing for the safety argument and the same phrasing is now repeated across the new override reads: this read is **not** compile-time dead.

`isGodModePanelEnabled` is an object *getter* (`config/featureFlags.ts:248`), so the store's `getSnapshot` (`overrides/store.ts:38`) performs a runtime property read. A production build can constant-fold the getter's body to `false`, but it can't eliminate the call sites — this code ships and simply always returns `null`.

The conclusion holds (production can never observe an override); it's only the mechanism that's misstated. Something like "the store's gate forces this to null in production" would be accurate. Note the same wording appears in `featureFlags.ts:244-245`, which claims the bundler drops the god-mode code — that's true for the `src/dev/` subtree behind the lazy `import.meta.env.DEV` import in `router.tsx:51`, but not for these override reads.

gbarkhatov (approved):
Approving — the refactor does what it says on the tin.

I traced every production consumer and the precedence ladders come out identical to `main`: `useLiquidationsPageState` is a faithful extraction, `resolveShownHealthFactor` / `resolveLiquidationCardState` are unchanged, the flag gate is now checked in exactly one place (`overrides/store.ts:38`), and no static `@/dev/*` import survives outside `src/dev/`. The `import.meta.env.DEV` removals in `VaultsPage` and `VaultsLifecycleSections` are safe because the store's own gate subsumes them. `tsc --noEmit` and `eslint` are clean, and the affected suites pass (the 3 `useVaultActions.test.ts` failures are a pre-existing stale `@babylonlabs-io/ts-sdk` mock, unrelated to this PR).

The comments below are one regression I'd want fixed before merge, plus a cleanup sweep.

The regression: the seam moved from **pull** to **push**, and the publisher is a component that unmounts more often than the design assumes. Every `set*Override` call now lives in a `useEffect` inside `DemoControls` or one of the three child panels, all of which render only inside `PanelBody` — which is unmounted whenever the panel is collapsed, and `collapsed` defaults to `true`. Demo state mutated from *outside* the panel (the simulated activation walk being the obvious one) therefore never republishes. Details and a repro are on `GodModePanel.tsx`.

Everything else is dead code and duplicated declarations the move left behind: `getDemoStepperBatch` now exists twice verbatim, five exports have no consumers left, four aggregate shapes plus `LiquidationPositionOverride` are declared on both sides of the seam, `useDashboardState` runs twice on `/liquidations`, and one comment describes a compile-time mechanism that isn't what actually happens.

PR #2263, opened 2026-08-15
feat(vault): align tooltip copy

jeremy-babylonlabs:
Aligns every vault tooltip with the new Figma **💬 Tooltips** page. 18 tooltips were audited: 1 already matched, 9 are reworded, 8 are new.

## Shared copy keys

Collateral factor and health factor had drifted into 3 and 2 separate phrasings across the app. Both now have a single key in `copy.ts` — `tooltips.collateralFactor` and `tooltips.healthFactor` — referenced from every surface, so the screens cannot drift apart again. The per-screen duplicates are deleted.

## New tooltips

- **Liquidation event cards** — target seizure, over seizure, fairness payment.
- **Markets stat bar** — available liquidity, total borrowed, market utilization.
- **Deposit form** — small-amount warning (`splitTooLowTooltip`).

## Reworded to match Figma

Supplied, borrow-flow APR and utilization, and the protocol parameters (min deposit, effective minimum for split, liquidation threshold, liquidation bonus).

## Label rename

"Transaction Reserve" becomes "Reserve Claimer UTXO", per the Figma fee panel.

## Deliberate deltas from Figma

Three places where the implementation does not copy Figma verbatim, so reviewers diffing against the design know they are intentional:

| Figma | Implemented | Why |
| --- | --- | --- |
| "liquidty" | "liquidity" | Typo in the design. |
| "the liquidator is  seize" | "the liquidator may seize" | Typo in the design. |
| APY (borrow-flow tooltip) | APR | The app uses APR everywhere else; the underlying figure is a rate, not a yield. |
| "The penalty range applied during the liquidation." (LB) | "Bonus percentage awarded to liquidators on seized collateral." | The row renders one number, `(LB - 1) x 100`, not a range — and it is the liquidator's premium, which the label "Liquidation Bonus (LB)" already says. The Figma wording would tell the reader something untrue. |
| "…if the margin ratio falls below 1.0…" | "…if it falls below 1.0…" | "Margin ratio" appears nowhere else in the app and has no antecedent in its own sentence. Since this key is the single wording across six surfaces, the orphan term would propagate to all of them. |
| "BTCVaults" | "BTC Vaults" | `copy.ts` mandates the two-word form, used 75 times elsewhere in the file. |

The two typos are raised with design separately — the app should not ship them.

## Accessibility and markup

Two fixes that came out of wiring the tooltips in:

- `LiquidationEventCard` label wrappers are now `<div>`, matching every other label-plus-`Hint` site. `Hint` renders a block-level wrapper, so the previous `<span>` produced invalid nesting.
- `SplitTooLowHint`'s live region now wraps only the message rather than the whole row. `Hint` renders its tooltip inline with no portal, so a tooltip inside a `role="status"` region made a screen reader re-announce the entire hint on hover.

## Testing

- `tsc --noEmit -p tsconfig.lib.json` — clean.
- `eslint` on all touched files — clean.
- `vitest` on the touched suites — 47 tests passing, including new coverage for the liquidation event card tooltips and the fairness-payment tooltip variant.

### DIFF
--- a/services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
+++ b/services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
@@ -1,3 +1,4 @@
+import { Hint } from "@babylonlabs-io/core-ui";
 import type { ReactNode } from "react";
 import { twJoin } from "tailwind-merge";
 
@@ -23,21 +24,33 @@ const SECTION_TITLE_CLASS =
   "text-base uppercase leading-[1.5] tracking-[0.15px] text-accent-secondary";
 const ROW_CLASS = "flex w-full items-center justify-between text-sm";
 const DIVIDER_CLASS = "h-px w-full bg-secondary-strokeLight";
+// `Hint` renders a block-level wrapper, so a label that carries one must be a
+// block element itself.
+const HINTED_LABEL_CLASS = "flex items-center gap-1";
 
 function DetailRow({
   label,
   value,
   labelClassName,
+  tooltip,
 }: {
   label: string;
   value: string;
   labelClassName?: string;
+  tooltip?: string;
 }) {
   return (
     <div className={ROW_CLASS}>
-      <span className={twJoin("text-accent-primary", labelClassName)}>
+      <div
+        className={twJoin(
+          HINTED_LABEL_CLASS,
+          "text-accent-primary",
+          labelClassName,
+        )}
+      >
         {label}
-      </span>
+        {tooltip ? <Hint tooltip={tooltip} /> : null}
+      </div>
       <span className="text-accent-primary">{value}</span>
     </div>
   );
@@ -69,11 +82,13 @@ function SeizureRow({
   amount,
   unit,
   emphasis,
+  tooltip,
 }: {
   label: string;
   amount: string;
   unit: string;
   emphasis?: boolean;
+  tooltip?: string;
 }) {
   return (
     <div
@@ -83,15 +98,17 @@ function SeizureRow({
           "border-t border-secondary-strokeLight bg-background-contrast",
       )}
     >
-      <span
+      <div
         className={twJoin(
+          HINTED_LABEL_CLASS,
           emphasis
             ? "text-base leading-[1.5] tracking-[0.15px] text-accent-primary"
             : "text-sm leading-[1.43] tracking-[0.17px] text-accent-secondary",
         )}
       >
         {label}
-      </span>
+        {tooltip ? <Hint tooltip={tooltip} /> : null}
+      </div>
       <span className="text-base leading-[1.5] tracking-[0.15px] text-accent-primary">
         {amount} <span className="text-accent-secondary">{unit}</span>
       </span>
@@ -174,11 +191,13 @@ export function LiquidationEventCard({ card }: { card: EventCardData }) {
           ))}
           <SeizureRow
             label={COPY.liquidations.events.targetSeizure}
+            tooltip={COPY.liquidations.events.targetSeizureTooltip}
             amount={card.targetSeizure.amount}
             unit={card.targetSeizure.unit}
           />
           <SeizureRow
             label={COPY.liquidations.events.overSeizure}
+            tooltip={COPY.liquidations.events.overSeizureTooltip}
             amount={card.overSeizure.amount}
             unit={card.overSeizure.unit}
           />
@@ -208,6 +227,7 @@ export function LiquidationEventCard({ card }: { card: EventCardData }) {
           <DetailRow
             label={card.fairness.label}
             value={card.fairness.value}
+            tooltip={card.fairness.tooltip}
             labelClassName="text-info-light"
           />
         </div>
--- a/services/vault/src/components/pages/Liquidations/PositionOverview.tsx
+++ b/services/vault/src/components/pages/Liquidations/PositionOverview.tsx
@@ -73,7 +73,7 @@ export function PositionOverview({
       },
       {
         label: COPY.liquidations.position.healthFactor,
-        tooltip: COPY.overview.healthFactorTooltip,
+        tooltip: COPY.tooltips.healthFactor,
         value: healthFactorText,
         valueNode: (
           <span className="flex items-center gap-2">
@@ -81,7 +81,6 @@ export function PositionOverview({
             <HeartIcon color={healthFactorColor} />
           </span>
         ),
-        caption: COPY.risk.healthFactorDescription,
       },
     ],
     [
--- a/services/vault/src/components/pages/Liquidations/liquidationChartData.ts
+++ b/services/vault/src/components/pages/Liquidations/liquidationChartData.ts
@@ -72,7 +72,7 @@ export interface LiquidationEventCard {
   debtRepaidLabel: string;
   liquidatorProfitLabel: string;
   /** Full group shows a wBTC fairness payment; safe groups show fairness debt repaid. */
-  fairness: { label: string; value: string };
+  fairness: { label: string; value: string; tooltip?: string };
   btcRemainingLabel: string;
   debtRemainingLabel: string;
   hfAfterLabel: string;
@@ -234,6 +234,7 @@ function toCard(
     ? {
         label: COPY.liquidations.events.fairnessPaymentWbtc,
         value: `${formatUsd(group.fairnessPaymentUsd)} (${formatBtcAmount(group.fairnessPaymentUsd / group.liquidationPrice)})`,
+        tooltip: COPY.liquidations.events.fairnessPaymentTooltip,
       }
     : {
         label: COPY.liquidations.events.fairnessDebtRepaid,
--- a/services/vault/src/components/simple/SplitTooLowHint.tsx
+++ b/services/vault/src/components/simple/SplitTooLowHint.tsx
@@ -1,3 +1,4 @@
+import { Hint } from "@babylonlabs-io/core-ui";
 import { IoInformationCircle } from "react-icons/io5";
 
 import { COPY } from "@/copy";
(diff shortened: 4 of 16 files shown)


services/vault/src/components/pages/Liquidations/PositionOverview.tsx
```diff
@@ -73,15 +73,15 @@ export function PositionOverview({
...
+        tooltip: COPY.tooltips.healthFactor,
         value: healthFactorText,
         valueNode: (
           <span className="flex items-center gap-2">
             {healthFactorText}
             <HeartIcon color={healthFactorColor} />
           </span>
         ),
-        caption: COPY.risk.healthFactorDescription,
+        caption: COPY.tooltips.healthFactor,
```
gbarkhatov: `tooltip` (line 76) and `caption` here are now the same string, so the sentence renders as visible body text under the value **and** again on hover over that same row.

They were distinct before this change: `COPY.overview.healthFactorTooltip` ("Your position's safety margin…") and `COPY.risk.healthFactorDescription` ("Indicates the health of your position…").

Pointing the tooltip at `COPY.tooltips.healthFactor` is right. The caption probably wants either its own key or to be dropped — a tooltip that repeats the caption printed directly beneath it adds nothing.

services/vault/src/copy.ts
```diff
@@ -1390,13 +1402,21 @@ export const COPY = {
       distance: "Distance",
       seizedVaultsSection: "Seized Vaults",
       targetSeizure: "Target seizure",
+      targetSeizureTooltip:
+        "The collateral value that the liquidator may receive during the liquidation process.",
       overSeizure: "Over seizure",
+      overSeizureTooltip:
+        "An additional portion of the collateral value that the liquidator may seize due to the nature of indivisible BTCVaults.",
```
gbarkhatov: `BTCVaults` contradicts this file's own style rule at lines 22-24: *"BTC Vault" (capitalized, two words) when naming the product or a depositor's vault*. That spelling is used 75 times elsewhere in this file, so these are the outliers.

Same at line 1656 (`…splitting deposit into two BTCVaults`), which additionally reads "splitting deposit" — it wants an article: "splitting a deposit into two BTC Vaults".

services/vault/src/copy.ts
```diff
@@ -1634,26 +1648,23 @@ export const COPY = {
...
     },
     liquidationThreshold: {
       label: "Liquidation threshold (THF)",
-      tooltip: "Target health factor at which liquidation becomes profitable.",
+      tooltip: "The ideal health factor to restore during liquidation",
     },
     liquidationBonus: {
       label: "Liquidation Bonus (LB)",
-      tooltip: "Bonus percentage awarded to liquidators on seized collateral.",
+      tooltip: "The penalty range applied during the liquidation.",
```
gbarkhatov: This does not match what the row renders. `useProtocolFeeRows.ts:74` computes `const bonusPercent = (LB - 1) * PERCENT_SCALE` and lines 76-78 render it as a single percentage against this tooltip — so "range" is wrong; there is exactly one number.

"Penalty" also inverts the framing its own label sets: the row is labelled "Liquidation Bonus (LB)" and the value is the liquidator's premium. The copy this replaces — "Bonus percentage awarded to liquidators on seized collateral" — described it correctly.

Of the five things I flagged on this PR, this is the one I would fix first: it is the only one that tells the reader something untrue.

services/vault/src/copy.ts
```diff
@@ -91,6 +91,16 @@ export interface EmphasisBodySegment {
...
 export const COPY = {
+  // Tooltips whose concept appears on more than one screen. One key per
+  // concept, referenced from every surface, so the screens can't drift apart.
+  tooltips: {
+    // Risk card, markets collateral card, protocol parameters (LTV).
+    collateralFactor:
+      "The maximum percentage of borrowing power from the total collateral value.",
+    // Overview position cards, risk card body, borrow / repay detail cards.
+    healthFactor:
+      "Indicates the health of your position. If the margin ratio falls below 1.0, your position may be liquidated.",
```
gbarkhatov: "margin ratio" appears nowhere else in `services/vault/src` — this string is its only occurrence — and it is introduced with a definite article and no antecedent in its own sentence: *"Indicates the health of your position. If **the margin ratio** falls below 1.0…"*. A reader has no way to know it refers to the health factor the sentence just described.

That matters more than usual here, because this key is deliberately the single wording across six surfaces, so an orphan term propagates to all of them at once. The strings it replaces said "Your position's safety margin", which at least tied the term to its sentence; "If it falls below 1.0" would do the same.

services/vault/src/components/simple/SplitTooLowHint.tsx
```diff
@@ -21,9 +22,15 @@ export function SplitTooLowHint({ minDepositForSplit }: SplitTooLowHintProps) {
       aria-live="polite"
       className="flex w-full items-center justify-center gap-2 rounded-lg border border-secondary-strokeLight px-3 py-2 text-center"
     >
-      <IoInformationCircle
-        size={18}
-        className="mt-px shrink-0 text-accent-primary"
+      <Hint
```
gbarkhatov, line 21: This places a tooltip inside a live region.

The wrapper on line 20 is `role="status"` + `aria-live="polite"`, and core-ui's `Hint` renders react-tooltip's `<Tooltip>` element inline in the tree (`Hint.tsx:108-124` — no portal container), so the tooltip element lands inside that region.

`role="status"` carries an implicit `aria-atomic="true"`, so when the tooltip opens on hover, react-tooltip's content mutation inside the region makes a screen reader re-announce the **entire** hint rather than just the new text. Rendering the `Hint` as a sibling of the live region — or giving the tooltip a portal container — avoids it.

gbarkhatov (dismissed):
Verified all five against the head tree at 1b0f9da rather than from the diff alone.

Cleanup checks out: `healthFactorTooltip`, `healthFactorDescription`, `collateralFactorTooltip` and `factorTooltip` have zero remaining references anywhere under `services/vault` (tests and E2E included), and `transactionReserveLabel` is a value rename whose single consumer (`DepositFeesBreakdown.tsx:144`) reads it by key, so nothing text-based depends on the old string. Consolidating the repeated tooltips onto one `COPY.tooltips` key is the right move — that drift is exactly what this file exists to prevent.

Five things inline. Only one is a correctness problem: **`copy.ts:1667`**, where the new Liquidation Bonus tooltip describes a "penalty range" while the row renders a single `(LB - 1) * 100` percentage. That one tells the reader something untrue, so it is worth fixing regardless of the rest. The other four are a duplicated string, two style-rule slips, an orphan term, and one accessibility regression.

Approving — none of it blocks, and the consolidation is a clear improvement.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -234,6 +234,7 @@ function toCard(
     ? {
         label: COPY.liquidations.events.fairnessPaymentWbtc,
         value: `${formatUsd(group.fairnessPaymentUsd)} (${formatBtcAmount(group.fairnessPaymentUsd / group.liquidationPrice)})`,
+        tooltip: COPY.liquidations.events.fairnessPaymentTooltip,
```
gbarkhatov, line 237: Confirming greptile's P1, because it is real — but it isn't yours to fix in a copy PR.

`fairnessPaymentUsd` is not full-group-only, despite `types.ts:25` documenting it as *"Full group only: fairness payment to borrower in USD"*. `calculate.ts:219` also computes it in the **non**-full branch, whenever the group over-seizes by more than the debt it can still repay (`leftoverOverSeizure > 0`) — precisely the lumpy indivisible-vault case that this PR's own over-seizure tooltip describes.

Reproduced against the head tree:

```
calculate({ btcPrice: 100_000, totalDebtUsd: 50_000, CF: 0.7, THF: 1.0, maxLB: 1.1,
            vaults: [{ name: "A", btc: 0.2 }, { name: "B", btc: 0.9 }, { name: "C", btc: 0.01 }] })

group 2 -> isFullLiquidation: false, fairnessPaymentUsd: 11672.78, debtRemainingAfter: 0
```

Group 2 clears the debt entirely and owes the borrower $11,672.78, but takes the `:` arm — so the card shows "Fairness Debt Repaid $30,421.62" and the payment disappears, value and tooltip both.

The `isFullLiquidation ? … : …` split predates this PR; you only hung a tooltip on the arm that was already there, so this is not a regression and I wouldn't block on it. Worth a follow-up issue against the projector, which will also need a call on whether `types.ts:25`'s doc or `calculate.ts`'s behaviour is the intended contract.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -83,15 +98,17 @@ function SeizureRow({
           "border-t border-secondary-strokeLight bg-background-contrast",
       )}
     >
-      <span
+      <div
         className={twJoin(
+          HINTED_LABEL_CLASS,
           emphasis
             ? "text-base leading-[1.5] tracking-[0.15px] text-accent-primary"
             : "text-sm leading-[1.43] tracking-[0.17px] text-accent-secondary",
```
gbarkhatov, line 106: Minor, and raised only because this PR is what introduced the constant pattern a few lines above.

Two typography literals now each appear twice in this file: `text-sm leading-[1.43] tracking-[0.17px] text-accent-secondary` here and at line 70, and `text-base leading-[1.5] tracking-[0.15px] text-accent-primary` at line 105 and again on the value span at line 112. With `ROW_CLASS` / `HINTED_LABEL_CLASS` / `SECTION_TITLE_CLASS` / `DIVIDER_CLASS` sitting right above, folding these in would finish that convention rather than leave it half-applied.

Not the full extent of what coderabbit asked for, though — its other site, `PositionOverview.tsx:79`, is a single-use `flex items-center gap-2`, and a named constant for three utilities used once would cost a jump and buy nothing. That one is fine inline.

gbarkhatov (approved):
Approved — all five of my earlier comments are resolved in `f904262`, checked against the head tree rather than the diff alone.

- **`copy.ts:1667`** (the one that told the reader something untrue) — back to "Bonus percentage awarded to liquidators on seized collateral.", which matches the single `(LB - 1) * 100` the row actually renders.
- **`copy.ts:102`** — "margin ratio" is gone; "If it falls below 1.0" now has an antecedent, and the term appears nowhere else under `services/vault/src`.
- **`copy.ts:1409` / `:1656`** — canonical "BTC Vaults" at both sites, and the missing article added.
- **`PositionOverview.tsx`** — the caption is dropped, so the tooltip no longer repeats a sentence printed directly beneath it. `overview.healthFactorTooltip` and `risk.healthFactorDescription` are removed from `copy.ts` with no dangling references left behind.
- **`SplitTooLowHint.tsx`** — `role="status"` now wraps only the message with the `Hint` as a sibling, so an open tooltip can't re-announce the whole row. The comment explaining why is worth keeping.

Recording the deliberate Figma deltas in the PR description is the right call — the LB wording and "falls below 1.0" read as decisions now rather than drift, and the two design typos are visibly someone's problem rather than silently shipped.

Verified locally on the head tree: `vitest` over the touched Liquidations and BorrowingMarketsData suites, 122 passing. (`tsc -p tsconfig.lib.json` flags only `rebuildDepositTerms` / `useVaultActions` against a stale `ts-sdk` dist — nothing in this PR's files.)

Two non-blocking notes inline; neither should hold the merge.

PR #2264, opened 2026-08-15
Feat/ledger sign psbt e2e

jrwbabylonlab:
The offline signing engine — merkleized-PSBT prepare, the 0xE000 interrupt loop, per-YIELD assertions with set-equal completeness, and bip174 merge — behind a public signVaultPsbt.

Gated by 22 firmware-fixture goldens and a byte-replay of 175 real-flow traces (+6 synthetic deep-tree) against Ledger's Python oracle. A Speculos E2E runs against real v0.9.5 firmware — derive → approve intent → SIGN_PSBT through our production encoders, with the yielded Schnorr signature verified against an independently computed BIP-341 sighash — but only when SPECULOS_URL is set; default runs skip it.

### DIFF
--- a/packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
+++ b/packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
@@ -18,11 +18,20 @@ their own license files. Vendored source:
   section of that directory's README)
 - License: Apache License 2.0
 - Modifications: see the per-file provenance headers in the vendored source
-  (explicit Buffer import, defensive strict-null guards, formatting; psbtv2.ts
-  additionally drops `fromBitcoinJS`, unifies the serialization key comparator
-  to byte-lexicographic order, and adds two BIP-371 `psbtIn` enum members plus
-  a public `getInputEntriesOfType` reader; clientCommands.ts adds an optional
-  `onYield` validator hook to `YieldCommand` / `ClientCommandInterpreter`).
+  (explicit Buffer import, defensive strict-null guards, formatting; varint.ts
+  additionally makes `parseVarint` reject the non-minimal "overlong" CompactSize
+  encodings upstream accepts; psbtv2.ts additionally drops `fromBitcoinJS`,
+  unifies the serialization key comparator to byte-lexicographic order, rejects
+  repeated keypairs and trailing bytes on deserialize, and adds two BIP-371
+  `psbtIn` enum members plus a public `getInputEntriesOfType` reader;
+  clientCommands.ts adds an optional `onYield` validator hook to `YieldCommand`
+  / `ClientCommandInterpreter`; buffertools.ts additionally makes
+  `unsafeTo64bitLE` reject input that is not a non-negative safe integer, which
+  upstream wrapped or encoded as zero; merkelizedPsbt.ts additionally requires
+  the input/output map counts to equal the declared `PSBT_GLOBAL_*_COUNT`s in
+  both directions, where upstream ignored surplus maps; merkle.ts additionally
+  throws a typed `Error` rather than a raw `TypeError` on an out-of-bounds leaf
+  index).
 
 ### Apache License 2.0
 
--- a/packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
+++ b/packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
@@ -12,12 +12,16 @@ their own license files. Vendored source:
 - Source: https://github.com/LedgerHQ/app-bitcoin
 - Version: ledger-bitcoin@0.3.0 (commit 0a9e9e141f3340d29e7c6181177d4e5e9483a9f7)
 - Files: vendored under packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/
-  (source-in-tree; not yet reachable from the published bundle)
+  — the source is in-tree and, from #2219 onward, the compiled vendored code is
+  also bundled into the published dist/ (build strips the per-file provenance
+  headers, so this notice is the redistribution record; see the audit-boundary
+  section of that directory's README)
 - License: Apache License 2.0
 - Modifications: see the per-file provenance headers in the vendored source
   (explicit Buffer import, defensive strict-null guards, formatting; psbtv2.ts
-  additionally drops `fromBitcoinJS` and unifies the serialization key
-  comparator to byte-lexicographic order; clientCommands.ts adds an optional
+  additionally drops `fromBitcoinJS`, unifies the serialization key comparator
+  to byte-lexicographic order, and adds two BIP-371 `psbtIn` enum members plus
+  a public `getInputEntriesOfType` reader; clientCommands.ts adds an optional
   `onYield` validator hook to `YieldCommand` / `ClientCommandInterpreter`).
 
 ### Apache License 2.0
--- a/packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
@@ -100,8 +100,10 @@ describe.skipIf(SPECULOS_URL === "")("Speculos end-to-end vault signing", () => 
       const deriveScreens = await approveOnScreen(SPECULOS_URL, DERIVE_APPROVAL_TEXT);
       contextRoot = await rootPending;
       expect(contextRoot).toHaveLength(32);
+      // NB: never log the context root, sighash or signature — CLAUDE.md #7
+      // ("never log payload bytes") is unconditional here, and this suite can
+      // be pointed at a real device. Assertions below pin the values instead.
       console.log(`[speculos-e2e] derive screens: ${deriveScreens.join(" || ")}`);
-      console.log(`[speculos-e2e] context root: ${Buffer.from(contextRoot).toString("hex")}`);
 
       // Scalars and group APDUs answer immediately; the final key batch blocks
       // on the intent review screen — drive it concurrently too.
@@ -139,7 +141,6 @@ describe.skipIf(SPECULOS_URL === "")("Speculos end-to-end vault signing", () => 
       expect(yielded.signerXOnlyHex).toBe(DEPOSITOR_XONLY_HEX);
       expect(yielded.leafHashHex).toBe(fixture.leaf0Hash.toString("hex"));
       expect(yielded.signature).toHaveLength(SCHNORR_SIG_BYTES);
-      console.log(`[speculos-e2e] schnorr signature: ${Buffer.from(yielded.signature).toString("hex")}`);
     },
     SIGNING_TIMEOUT_MS,
   );
@@ -155,7 +156,6 @@ describe.skipIf(SPECULOS_URL === "")("Speculos end-to-end vault signing", () => 
 
       const sighash = computePeginSighash(psbtFixture.psbtHex, psbtFixture);
       const valid = verifySchnorrSignature(sighash, DEPOSITOR_XONLY_HEX, yielded.signature);
-      console.log(`[speculos-e2e] sighash: ${sighash.toString("hex")}`);
       console.log(`[speculos-e2e] verifySchnorr(depositor key): ${valid}`);
       expect(valid).toBe(true);
 
--- a/packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
@@ -1,0 +1,214 @@
+/**
+ * Speculos end-to-end signing test (#2219 B2): the full production path —
+ * envelope gate → DERIVE_CONTEXT_HASH → APPROVE_VAULT_INTENT → signVaultPsbt —
+ * against the real vault app in a Speculos container, ending in cryptographic
+ * verification of the device's Schnorr signature.
+ *
+ * Requires a running container with the vault app (nanosp, testnet build,
+ * firmware-test mnemonic). Skipped unless SPECULOS_URL is set:
+ *
+ *   SPECULOS_URL=http://127.0.0.1:5055 pnpm exec vitest run src/__tests__/e2e/
+ *
+ * The suite is stateful and ordered: the ceremony arms the device state the
+ * signing stage consumes. Each run re-runs the full ceremony, so the container
+ * does not need restarting between runs.
+ */
+
+import { Psbt } from "bitcoinjs-lib";
+import { Buffer } from "buffer";
+import { describe, expect, it } from "vitest";
+
+import { assertDepositTermsDeviceCompatible } from "../../envelope";
+import { signVaultPsbt, type SignVaultPsbtResult } from "../../signPsbt";
+import { approveVaultIntent, deriveContextHash } from "../../vaultCommands";
+import {
+  buildDepositTerms,
+  buildPeginPsbt,
+  computePeginSighash,
+  DEPOSITOR_PATH,
+  DEPOSITOR_XONLY_HEX,
+  DERIVE_CONTEXT,
+  HTLC_VOUT,
+  PREPEGIN_TXID_INTERNAL,
+  termsToIntent,
+  VAULT_APP_NAME,
+  vaultHashlock,
+  verifySchnorrSignature,
+  type PeginPsbtFixture,
+} from "./peginFixture";
+import {
+  approveOnScreen,
+  createSpeculosApduSender,
+  createSpeculosRawApduSender,
+  getAppAndVersion,
+  readScreenText,
+  sleep,
+} from "./speculosClient";
+
+const SPECULOS_URL = process.env.SPECULOS_URL ?? "";
+
+/** Live-verified nanosp review-screen texts (reference driver `lsk_ceremony2.py`). */
+const DERIVE_APPROVAL_TEXT = "Allow derivation";
+const INTENT_APPROVAL_TEXT = "Approve intent";
+
+/** Ceremony = APDUs + human-speed screen navigation; signing = ~40 APDU rounds. */
+const CEREMONY_TIMEOUT_MS = 120_000;
+const SIGNING_TIMEOUT_MS = 60_000;
+const SANITY_TIMEOUT_MS = 30_000;
+
+const SCHNORR_SIG_BYTES = 64;
+
+describe.skipIf(SPECULOS_URL === "")("Speculos end-to-end vault signing", () => {
+  const sendRaw = createSpeculosRawApduSender(SPECULOS_URL);
+  const send = createSpeculosApduSender(SPECULOS_URL);
+
+  // Ordered stages share state; a later stage failing on `undefined` means an
+  // earlier stage failed first — read its failure, not this one.
+  let contextRoot: Uint8Array | undefined;
+  let fixture: PeginPsbtFixture | undefined;
+  let signResult: SignVaultPsbtResult | undefined;
+
+  it(
+    "(1) reports the vault app name and version",
+    async () => {
+      const app = await getAppAndVersion(sendRaw);
+      expect(app.name).toContain("Babylon Vault");
+      expect(app.version).toMatch(/^\d+\.\d+\.\d+$/);
+      console.log(`[speculos-e2e] app: ${app.name} v${app.version}`);
+    },
+    SANITY_TIMEOUT_MS,
+  );
+
+  it(
+    "(2) ceremony: envelope gate, DERIVE_CONTEXT_HASH, APPROVE_VAULT_INTENT via the production encoders",
+    async () => {
+      const terms = buildDepositTerms();
+      // The envelope gate must accept these terms with zero device I/O.
+      assertDepositTermsDeviceCompatible(terms);
+      const intent = termsToIntent(terms);
+      // Byte-order seam: display-order terms txid maps to the internal-order
+      // prevout the PSBT carries and the firmware compares against.
+      expect(Buffer.from(intent.scalars.prepeginTxidInternal).equals(PREPEGIN_TXID_INTERNAL)).toBe(true);
+
+      // The final DERIVE APDU blocks on the review screen — drive it concurrently.
+      const rootPending = deriveContextHash(send, {
+        appName: VAULT_APP_NAME,
+        derivationPath: DEPOSITOR_PATH,
+        context: DERIVE_CONTEXT,
+      });
+      rootPending.catch(() => undefined); // handled at the await below
+      const deriveScreens = await approveOnScreen(SPECULOS_URL, DERIVE_APPROVAL_TEXT);
+      contextRoot = await rootPending;
+      expect(contextRoot).toHaveLength(32);
+      console.log(`[speculos-e2e] derive screens: ${deriveScreens.join(" || ")}`);
+      console.log(`[speculos-e2e] context root: ${Buffer.from(contextRoot).toString("hex")}`);
+
+      // Scalars and group APDUs answer immediately; the final key batch blocks
+      // on the intent review screen — drive it concurrently too.
+      await waitForIdleScreen();
+      const approvePending = approveVaultIntent(send, intent);
+      approvePending.catch(() => undefined); // handled at the await below
+      const intentScreens = await approveOnScreen(SPECULOS_URL, INTENT_APPROVAL_TEXT);
+      await approvePending;
+      console.log(`[speculos-e2e] intent screens: ${intentScreens.join(" || ")}`);
+    },
+    CEREMONY_TIMEOUT_MS,
+  );
+
+  it(
+    "(3) signVaultPsbt signs the PegIn PSBT on the device",
+    async () => {
+      expect(contextRoot, "ceremony stage must have produced a root").toBeDefined();
+      const hashlock = vaultHashlock(contextRoot as Uint8Array, HTLC_VOUT);
+      fixture = buildPeginPsbt(hashlock);
+
+      // PegIn signing is silent (fw test_sign_psbt_validate.py header) — the
+      // approved intent authorizes it, so no screen driving here.
+      // Shorter than the test timeout so the typed abort surfaces first.
+      signResult = await signVaultPsbt(sendRaw, {
+        psbtHex: fixture.psbtHex,
+        depositorXOnlyHex: DEPOSITOR_XONLY_HEX,
+        signal: AbortSignal.timeout(SIGNING_TIMEOUT_MS - 10_000),
+      });
+
+      expect(signResult.yields).toHaveLength(1);
+      const yielded = signResult.yields[0];
+      expect(yielded.kind).toBe("tapscript");
+      if (yielded.kind !== "tapscript") throw new Error("unreachable");
+      expect(yielded.inputIndex).toBe(0);
+      expect(yielded.signerXOnlyHex).toBe(DEPOSITOR_XONLY_HEX);
+      expect(yielded.leafHashHex).toBe(fixture.leaf0Hash.toString("hex"));
+      expect(yielded.signature).toHaveLength(SCHNORR_SIG_BYTES);
+      console.log(`[speculos-e2e] schnorr signature: ${Buffer.from(yielded.signature).toString("hex")}`);
+    },
+    SIGNING_TIMEOUT_MS,
+  );
+
+  it(
+    "(4) the device signature verifies against the BIP-341 script-path sighash",
+    () => {
+      expect(fixture, "signing stage must have built the fixture").toBeDefined();
+      expect(signResult, "signing stage must have produced yields").toBeDefined();
+      const psbtFixture = fixture as PeginPsbtFixture;
+      const yielded = (signResult as SignVaultPsbtResult).yields[0];
+      if (yielded.kind !== "tapscript") throw new Error("stage (3) asserted tapscript");
+
+      const sighash = computePeginSighash(psbtFixture.psbtHex, psbtFixture);
+      const valid = verifySchnorrSignature(sighash, DEPOSITOR_XONLY_HEX, yielded.signature);
+      console.log(`[speculos-e2e] sighash: ${sighash.toString("hex")}`);
+      console.log(`[speculos-e2e] verifySchnorr(depositor key): ${valid}`);
+      expect(valid).toBe(true);
+
+      // Negative control: a different message must not verify.
+      const tampered = Buffer.from(sighash);
+      tampered[0] ^= 0x01;
+      expect(verifySchnorrSignature(tampered, DEPOSITOR_XONLY_HEX, yielded.signature)).toBe(false);
+    },
+    SANITY_TIMEOUT_MS,
+  );
+
+  it(
+    "(5) the merged PSBT carries the tapScriptSig and finalizes",
+    () => {
+      expect(fixture, "signing stage must have built the fixture").toBeDefined();
+      expect(signResult, "signing stage must have produced a merged PSBT").toBeDefined();
+      const psbtFixture = fixture as PeginPsbtFixture;
+      const result = signResult as SignVaultPsbtResult;
+
+      const original = Psbt.fromHex(psbtFixture.psbtHex);
+      const merged = Psbt.fromHex(result.signedPsbtHex);
+      // Merge writes signature fields only — the unsigned tx is byte-identical.
+      expect(merged.data.globalMap.unsignedTx.toBuffer().equals(original.data.globalMap.unsignedTx.toBuffer())).toBe(
+        true,
+      );
+
+      const tapScriptSig = merged.data.inputs[0].tapScriptSig;
+      expect(tapScriptSig).toBeDefined();
+      expect(tapScriptSig).toHaveLength(1);
+      const entry = (tapScriptSig ?? [])[0];
+      expect(entry.pubkey.toString("hex")).toBe(DEPOSITOR_XONLY_HEX);
+      expect(entry.leafHash.toString("hex")).toBe(psbtFixture.leaf0Hash.toString("hex"));
+      expect(entry.signature).toHaveLength(SCHNORR_SIG_BYTES);
+
+      // Structural finalizability: bitcoinjs accepts the script-path input and
+      // assembles witness = [sig, leaf script, control block]. (Consensus
+      // validity additionally needs the co-signers + preimage — out of scope.)
+      merged.finalizeInput(0);
+      const witness = merged.data.inputs[0].finalScriptWitness;
+      expect(witness).toBeDefined();
+      expect((witness ?? Buffer.alloc(0)).includes(psbtFixture.leaf0Script)).toBe(true);
+      expect((witness ?? Buffer.alloc(0)).includes(psbtFixture.controlBlock)).toBe(true);
+    },
+    SANITY_TIMEOUT_MS,
+  );
+
+  /** Best-effort settle back to the dashboard between the two approval flows. */
+  async function waitForIdleScreen(): Promise<void> {
+    const deadline = Date.now() + 5_000;
+    while (Date.now() < deadline) {
+      const screen = await readScreenText(SPECULOS_URL);
+      if (screen.includes("app is ready")) return;
+      await sleep(150);
+    }
+  }
+});
--- a/packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculosClient.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculosClient.test.ts
@@ -3,20 +3,47 @@
  *
  * `getAppAndVersion` parses a length-prefixed device response, so every length
  * byte is attacker/emulator-controlled — the truncation guards are the only
- * thing between a malformed frame and a silently wrong result.
+ * thing between a malformed frame and a silently wrong result. Every rejection
+ * case matches the message of the one guard it pins, so collapsing two guards
+ * onto a shared message fails here rather than passing on a sibling's throw.
+ *
+ * The transport tests pin the posted wire bytes for a NON-EMPTY payload: this
+ * sender carries the SIGN_PSBT cdata (built in `signPsbtPrepare.ts` — a
+ * 65-byte global commitment `varint(n) ‖ keysRoot(32) ‖ valuesRoot(32)`, two
+ * count varints, and four 32-byte fields: inputsRoot, outputsRoot, walletId,
+ * walletHmac; 195 bytes across every committed signpsbt vector) and every
+ * CONTINUE payload, and a zero-length APDU cannot distinguish a correct frame
+ * from an all-zero buffer.
  */
 
-import { describe, expect, it } from "vitest";
+import { Buffer } from "buffer";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
-import type { RawApduSender } from "../../rawApdu";
-import { getAppAndVersion } from "./speculosClient";
+import { LEDGER_USER_REFUSED_ERROR_NAME } from "../../errors";
+import type { Apdu, RawApduSender } from "../../rawApdu";
+import { createSpeculosApduSender, createSpeculosRawApduSender, getAppAndVersion, SW_OK } from "./speculosClient";
 
-/** A sender that answers 0x9000 with the given response body. */
+/** A sender that answers success with the given response body. */
 function senderReturning(dataHex: string): RawApduSender {
-  return async () => ({ sw: 0x9000, data: Buffer.from(dataHex, "hex") });
+  return async () => ({ sw: SW_OK, data: Buffer.from(dataHex, "hex") });
 }
 
 describe("getAppAndVersion", () => {
+  it("asks for the BOLOS GET_APP_AND_VERSION with no data", async () => {
+    const sent: Apdu[] = [];
+    const send: RawApduSender = async (apdu) => {
+      sent.push(apdu);
+      return { sw: SW_OK, data: Buffer.from("010361707003312e30", "hex") };
+    };
+
+    await getAppAndVersion(send);
+
+    // B0 01 00 00 (Lc 0) per the reference client LedgerHQ/ledger-live@2ec1cda
+    // `libs/ledger-live-common/src/hw/getAppAndVersion.ts` — `transport.send(0xb0, 0x01, 0x00, 0x00)`.
+    expect(sent).toHaveLength(1);
+    expect(sent[0]).toEqual({ cla: 0xb0, ins: 0x01, p1: 0x00, p2: 0x00, data: new Uint8Array(0) });
+  });
+
   it("parses a well-formed response", async () => {
     // format(01) ‖ len(03) ‖ "app" ‖ len(03) ‖ "1.0"
     const app = await getAppAndVersion(senderReturning("010361707003312e30"));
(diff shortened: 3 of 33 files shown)


packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      expect(Buffer.from(intent.scalars.prepeginTxidInternal).equals(PREPEGIN_TXID_INTERNAL)).toBe(true);
+
+      // The final DERIVE APDU blocks on the review screen — drive it concurrently.
+      const rootPending = deriveContextHash(send, {
+        appName: VAULT_APP_NAME,
+        derivationPath: DEPOSITOR_PATH,
+        context: DERIVE_CONTEXT,
+      });
+      rootPending.catch(() => undefined); // handled at the await below
+      const deriveScreens = await approveOnScreen(SPECULOS_URL, DERIVE_APPROVAL_TEXT);
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      expect(yielded.kind).toBe("tapscript");
+      if (yielded.kind !== "tapscript") throw new Error("unreachable");
+      expect(yielded.inputIndex).toBe(0);
+      expect(yielded.signerXOnlyHex).toBe(DEPOSITOR_XONLY_HEX);
+      expect(yielded.leafHashHex).toBe(fixture.leaf0Hash.toString("hex"));
+      expect(yielded.signature).toHaveLength(SCHNORR_SIG_BYTES);
+      console.log(`[speculos-e2e] schnorr signature: ${Buffer.from(yielded.signature).toString("hex")}`);
+    },
+    SIGNING_TIMEOUT_MS,
+  );
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculosClient.ts
```diff
@@ -0,0 +1,186 @@
...
+  if (data.length < 3 || data[0] !== APP_AND_VERSION_FORMAT) {
+    throw new Error(`GET_APP_AND_VERSION answered an unknown format (${toHex(data)})`);
+  }
+  const nameLen = data[1];
+  const nameEnd = 2 + nameLen;
+  const versionLen = data[nameEnd];
+  const versionEnd = nameEnd + 1 + versionLen;
+  if (versionEnd > data.length) {
+    throw new Error(`GET_APP_AND_VERSION response truncated (${toHex(data)})`);
+  }
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md
```diff
@@ -58,22 +58,51 @@ with the exceptions recorded per file header; the substantive ones:
...
+  `0a9e9e14`. Provenance headers plus the golden-vector gates below are what
+  makes that surface auditable — not its absence from the bundle.
+
+## Known upstream behaviour (kept): v0 round-trip is not bitcoinjs-canonical
+
+`PsbtV2.serialize()` on a deserialized-but-NOT-normalized v0 model re-emits
+bytes that `bitcoinjs-lib` rejects ("Only one UNSIGNED_TX allowed"). No
+production path serializes an un-normalized model (prepare always normalizes),
+and `prepareSignPsbt`'s merge-target parse gate rejects such input pre-I/O.
+Kept upstream-faithful.
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

jonybur (dismissed):
Approve. The SIGN_PSBT client is correctly gated: prepareSignPsbt runs all validation (hex parse, PSBT round-trip, ownership scan) synchronously with typed errors before any device I/O, expectedSignatures.assertAndRecord enforces per-YIELD input/spend-type/signer/leaf-hash/64-byte-sig checks with set-equal completeness in both directions plus size, and signPsbtMerge only writes tapScriptSig/tapKeySig and never finalizes. TLV/APDU encodings cite firmware sources (base:constants.h, sw.h, init_global_state.c, sign_input.c, psbt.h) and the two new BIP-371 psbtIn enum members cite psbt.h:37-42. No payload byte logging; the speculos e2e log statements only carry booleans, screen strings, and app version and the comment at speculos.e2e.test.ts:551-553 documents that constraint. No silent fallbacks - the sole swallowing catch wraps onProgress display callbacks. All prior reviewer findings (payload logging, versionLen guard, psbtv2 normalization framing) are resolved at head.

packages/babylon-ledger-vault-signer/src/signPsbtPrepare.ts
```diff
@@ -0,0 +1,209 @@
...
+      `PSBT rejected at parse (merge target): ${error instanceof Error ? error.message : String(error)}`,
+    );
+  }
+
+  const psbt = new PsbtV2();
+  try {
+    // Parses the v0 unsigned tx for counts, then normalizeToV2 synthesizes the
+    // v2 fields and deletes UNSIGNED_TX; unknown keys (all taproot fields)
+    // pass through byte-identical (`psbtv2.ts` deserialize/normalizeToV2).
+    psbt.deserialize(Buffer.from(psbtHex, "hex"));
```
gbarkhatov, line 234: **Two parsers, no cross-check: a crafted PSBT can make the device sign one document while the host merges into another.**

`psbtHex` is parsed twice and the results are never compared — `BitcoinjsPsbt.fromHex` at :154 (the merge target, re-parsed in `signPsbtMerge.ts:29`) and `PsbtV2.deserialize` here (the model that gets merkleized and committed to the device).

The two disagree on non-minimal CompactSize. bip174's `fromBuffer.js` does `offset += varuint.encodingLength(keyLen)` — the *canonical* width — while the true wire width is `varuint.decode.bytes`. For `fd 01 00` that is 1 vs 3. The vendored `parseVarint` (`varint.ts:86-105`) correctly returns 3. bip174 also has no trailing-byte check, so the resulting desync is absorbed silently, whereas `psbtv2.ts:421-423` rejects trailing bytes.

I reproduced this. A PSBT whose first input-0 keypair is an unknown key `0xfc` with its value length written non-minimally as `fd 08 00` and a value ending `00 00`:

- `PsbtV2` semantics read input 0 as `[fc…, 15c050…, 01…]` — a well-formed tapscript input — with 0 trailing bytes.
- `Psbt.fromHex` **accepts**, and reads input 0 as `unknownKeyVals` only: `tapLeafScript: undefined`, `witnessUtxo: undefined`.
- Both see byte-identical `unsignedTx`.
- `updateInput(0, { tapScriptSig: [...] })` — exactly what `mergeYields` does — **succeeds**, writing a device signature into an input that has no leaf script and no witnessUtxo.

Net: the table classifies input 0 as tapscript, the ceremony runs, the device signs, and the merge silently produces a malformed signed PSBT. Fail-closed for fund loss (SIGHASH_DEFAULT commits to everything, and `tapScriptFinalizer` matches by leaf hash), but it burns an approved intent, and it voids the implicit "bip174 already validated these leaf bytes" mitigation that the control-block point on :150 otherwise leans on.

Cheapest fix at the root: reject non-minimal encodings in `parseVarint`. Belt-and-braces: after `normalizeToV2()`, reconstruct the implied unsigned tx from the v2 model and assert it equals `BitcoinjsPsbt.fromHex(psbtHex).data.globalMap.unsignedTx.toBuffer()`, plus compare the per-input fields the merge depends on.

packages/babylon-ledger-vault-signer/src/signPsbtPrepare.ts
```diff
@@ -0,0 +1,209 @@
...
+  // Buffer.from(hex) silently truncates at the first invalid character.
+  if (!PSBT_HEX_RE.test(psbtHex)) {
+    throw new LedgerSignPsbtProtocolError("psbtHex is not even-length hex");
+  }
+
+  // The merge stage folds signatures back via bitcoinjs Psbt — assert
+  // parseability BEFORE any device I/O so a PsbtV2-acceptable-but-
+  // bitcoinjs-rejected PSBT cannot burn an approved intent and then fail.
+  try {
+    BitcoinjsPsbt.fromHex(psbtHex);
```
gbarkhatov: **The zero-I/O gate misses already-signed and finalized inputs.**

This parse exists specifically so that "a PsbtV2-acceptable-but-bitcoinjs-rejected PSBT cannot burn an approved intent and then fail" — but the result is discarded, and the other member of that same failure class isn't checked: an input that already carries `tapKeySig`, a matching `(pubkey, leafHash)` `tapScriptSig`, or a `finalScriptWitness`.

Those are precisely the fields bip174 refuses to overwrite, and this PR's own `signPsbtMerge.test.ts:109-124` proves both throws. So the sequence is: prepare passes → full ceremony runs → device signs → `mergeYields` throws → `signVaultPsbt` wraps it as `merge failed after signing`. The intent and a device signature slot are already spent, which `errors.ts:74-79` and the `0xB00A` mapping in `rawApdu.ts` explain is not recoverable.

Fix is free: keep the `Psbt` parsed on this line and, after the table is built, assert for every `inputIndex` in `table.byInput` that there is no `finalScriptWitness`/`finalScriptSig`, no `tapKeySig` on a keypath input, and no matching `tapScriptSig` on a tapscript input.

packages/babylon-ledger-vault-signer/src/signPsbt.ts
```diff
@@ -0,0 +1,92 @@
...
+  readonly onProgress?: (progress: SignPsbtProgress) => void;
+  /**
+   * Host liveness control, checked before the initial send and before every
+   * CONTINUE. The loop has no round cap and no internal timeout (the abort
+   * signal is by design the only way to bound it) — production callers MUST
+   * always pass a signal; omit it only in tests with a scripted transport.
+   * The check runs between exchanges: an in-flight transport call that hangs
+   * is not itself cancelled, so a bounded transport remains the caller's job.
+   */
+  readonly signal?: AbortSignal;
```
gbarkhatov: **The public API doesn't enforce the liveness contract it documents.**

The JSDoc directly above says production callers MUST pass `signal` because the loop has no round cap and no internal timeout — but the exported type allows omission, and `signVaultPsbt` forwards it unvalidated.

A device that answers every CONTINUE with `0xE000` plus a valid `GET_PREIMAGE` for an already-known hash spins forever: the promise never settles, and the DMK session and UI ceremony stay pinned. Nothing in the loop breaks that.

No production consumer is wired yet (#2221), so making this `readonly signal: AbortSignal` is cheap right now; `RunSignPsbtLoopOptions.signal` can stay optional for the scripted-transport tests. A round cap is the alternative — after `prepareSignPsbt` the maximum honest round count is fully known (Σ map sizes + `expectedYieldCount`). Either way, a test with a device that answers `0xE000` forever would pin the behaviour.

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  // Precompute the depositor-owned scriptPubKeys once for the ownership scan.
+  const depositorP2trScript = bip86OutputScript(depositorXOnlyHex);
+  const depositorP2wpkhScripts = [
+    p2wpkhOutputScript(depositorXOnlyHex, COMPRESSED_KEY_EVEN_PREFIX),
+    p2wpkhOutputScript(depositorXOnlyHex, COMPRESSED_KEY_ODD_PREFIX),
+  ];
+
+  for (let inputIndex = 0; inputIndex < inputCount; inputIndex++) {
+    const leafEntries = psbt.getInputEntriesOfType(inputIndex, psbtIn.TAP_LEAF_SCRIPT);
```
gbarkhatov, line 238: **Classifying on `TAP_LEAF_SCRIPT` first and `continue`-ing skips every ownership check on that input.**

A tapscript input is only checked for: exactly one leaf entry, non-empty value, trailing leaf version `0xc0`, and witnessUtxo present. It never inspects the witnessUtxo `scriptPubKey` or amount, never uses the control block returned in `keyData`, and never checks that the depositor key appears in the leaf script — `expectedSignerXOnlyHex` is pure passthrough at :181. Compare the keypath branch just below, which independently re-derives the BIP-86 program and requires the witness program to equal it (:199-208).

Two consequences:

1. Adding a single `TAP_LEAF_SCRIPT` entry to an input carrying `TAP_INTERNAL_KEY = depositorKey` flips classification to tapscript and silently disables the strongest host pin in this file, `script.equals(depositorP2trScript)` at :204.
2. A PSBT with an extra input spending a depositor-owned script-path UTXO — e.g. the Depositor-Claim `P2TR(NUMS, {<D> OP_CHECKSIG})` shape that `__tests__/e2e/peginFixture.ts:186-188` builds — passes the table build, never reaches the ownership scan at :219-228 because of the `continue` at :183, and gets a depositor signature recorded and merged. The only thing between that PSBT and the coin is the on-device intent validator.

Nothing binds the pinned leaf to the output being spent, either: no check that `(controlBlock[0] & 0xfe) === leafVersion`, that `(controlBlock.length - 33) % 32 === 0`, or that `tweak(internalKey, root_from_path(leafHash, path))` equals `witnessUtxo.scriptPubKey.subarray(2)`. bip174's `tapLeafScript.decode` does enforce the first two as a side effect of the `Psbt.fromHex` gate — but that mitigation is unsound given the parser differential noted on :166.

Suggested: reject inputs carrying both `TAP_LEAF_SCRIPT` and `TAP_INTERNAL_KEY`; run the ownership scan on every input regardless of classification; assert the depositor x-only key is present in the leaf script; and verify the control block commits to the witness program. The `ecc` backend is already imported at :21, so the last one is ~15 lines.

packages/babylon-ledger-vault-signer/src/signPsbt.ts
```diff
@@ -0,0 +1,92 @@
...
+    onProgress: params.onProgress,
+    signal: params.signal,
+    resendOnceOnIncorrectData: params.resendOnceOnIncorrectData,
+  });
+  let signedPsbtHex: string;
+  try {
+    signedPsbtHex = mergeYields(prepared.originalPsbtHex, yields);
+  } catch (error) {
+    // Post-ceremony merge failures stay inside the typed contract.
+    throw new LedgerSignPsbtProtocolError(
```
gbarkhatov, line 92: **A merge failure throws away the signatures the device just produced.**

`yields` is in scope here and is dropped on the way out. Per `errors.ts:74-79` the ceremony is non-idempotent and cannot be replayed, so those signatures — and the #2221 PoP seam that `SignVaultPsbtResult.yields` exists to expose — are unrecoverable at exactly the moment they are most expensive.

This compounds with the gate suggested on :154: an already-signed PSBT is the most likely way to reach this throw, and this is what makes reaching it costly.

Suggested: attach the collected yields to the thrown error (or return them alongside the failure) so a caller can still package the witness data.

packages/babylon-ledger-vault-signer/src/errors.ts
```diff
@@ -58,6 +62,96 @@ export class LedgerDeviceError extends Error {
...
+ *   uses `resendOnceOnIncorrectData`.
+ * - A validation-phase abandonment leaves the intent intact; a Payout/NoPayout
+ *   SIGNING-phase abandonment invalidates it (`fw:sign_custom_inputs.c:207-344`)
+ *   — post-abort "intent-loaded" is a HINT only, and the retry path must also
+ *   absorb 0xB007.
+ * - After >5 s of host silence the app exits to the dashboard
+ *   (`base:io_ext.c:103-111` → SDK `app_exit()`); the host then sees 0x6E00 or
+ *   a transport failure — full re-ceremony required.
+ */
+export class LedgerSignPsbtAbortedError extends Error {
```
gbarkhatov, line 188: **The abort error can't tell the caller whether signatures were already burned.**

The abort check in `signPsbtLoop.ts:124-127` runs *after* `interpreter.execute()` at :97, so a round that delivered a YIELD records the signature and only then throws this no-payload error.

The doc comment right here says a validation-phase abandonment leaves the intent intact while a signing-phase one invalidates it — but the thrown error gives the retry path no way to distinguish, and the collected yields are unreachable. Concretely: the user cancels mid-ceremony on a Payout after yield 1 of 2, and the retry can't know the `payout_claimer_mask` bit is already set.

Suggested: add `readonly yieldedCount: number` (and optionally a `phase` discriminator), populated from `reportedYields` at the throw site.

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
```diff
@@ -0,0 +1,554 @@
...
+        tapscriptYieldRequestHex(0, TEST_DEPOSITOR_KEY_HEX, tableLeafHex(p.table, 0), TEST_SIG_HEX + "01"),
+    },
+    {
+      label: "33-byte augm (ECDSA compressed key)",
+      kind: "unexpected-encoding",
+      requestHex: () => "10" + "00" + "21" + "02" + TEST_DEPOSITOR_KEY_HEX + TEST_SIG_HEX,
+    },
+  ];
+
+  it.each(MUTATIONS)("$label -> $kind, no further sends", async ({ kind, requestHex }) => {
```
gbarkhatov, line 556: **Every mutation case runs against a tapscript fixture, so the keypath branch has zero coverage.**

All six `MUTATIONS` entries and this `it.each` — plus the duplicate-yield case at :459 — use `generated__deposit-flow__pegin__0`. That leaves `expectedSignatures.ts:293/332/337/340/349`, the entire `taproot-keypath` mismatch path, unreached by any test.

Concretely: delete `if (outputKeyHex !== expectation.expectedOutputKeyHex) throw` at `expectedSignatures.ts:336-338` and the whole suite still passes — while the device could then yield a keypath signature under an arbitrary output key on a Pre-PegIn PSBT and have it recorded and merged. Pre-PegIn is a production path, so this isn't hypothetical.

Suggested: parameterise `MUTATIONS` over both fixtures — `deposit-flow__pre_pegin__0` is already loaded in this file at :159.

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtPrepare.test.ts
```diff
@@ -0,0 +1,266 @@
...
+    // 64 valid hex chars, but not a point on secp256k1 — bitcoinjs p2tr throws
+    // a raw TypeError; the prepare contract must retype it.
+    expectPrepareRejects(VALID_PSBT_HEX, "ff".repeat(32), /PSBT rejected at preflight/);
+  });
+
+  it("wraps the vendored parse throw for an unsupported PSBT version", () => {
+    // magic ‖ global map { VERSION(0xfb) → 1 } ‖ end-of-map — version 1 is
+    // rejected inside PsbtV2.deserialize and surfaces as a typed error.
+    const versionOnePsbtHex = "70736274ff" + "01fb" + "0401000000" + "00";
+    expectPrepareRejects(versionOnePsbtHex, TEST_DEPOSITOR_KEY_HEX, /PSBT rejected at parse/);
```
gbarkhatov: **This test passes for the wrong reason — the `PsbtV2` version guard is untested.**

The fixture `70736274ff 01fb 0401000000 00` has no `PSBT_GLOBAL_UNSIGNED_TX`, so `BitcoinjsPsbt.fromHex` at `signPsbtPrepare.ts:154` throws first. I ran it: `Format Error: Only one UNSIGNED_TX allowed`. Execution never reaches `PsbtV2.deserialize`, so the `psbtv2.ts:393` version rejection that the test name and comment describe is not exercised.

The matcher is loose enough to hide it: `/PSBT rejected at parse/` also matches the merge-target wrapper `"PSBT rejected at parse (merge target): …"`.

Suggested: use a fixture with a valid `UNSIGNED_TX` plus `PSBT_GLOBAL_VERSION = 1`, and tighten the matcher to `/PSBT rejected at parse: /` (with the colon) so it can't match the merge-target variant.

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
```diff
@@ -0,0 +1,554 @@
...
+    ["YIELD truncated inside the varint", "10" + "fd01"],
+    ["YIELD truncated inside the augm block", "10" + "00" + "40" + "aa".repeat(8)],
+  ])("%s -> LedgerSignPsbtProtocolError, no further sends", async (_label, requestHex) => {
+    const prepared = prepareFromVector("generated__deposit-flow__pegin__0");
+    const script: ScriptedExchange[] = [
+      { expectApduHex: initialApduHexOf(prepared), respondSw: 0xe000, respondDataHex: requestHex },
+    ];
+    const { send, sent } = createScriptedSender(script);
+
+    await expect(runSignPsbtLoop(send, prepared, {})).rejects.toBeInstanceOf(LedgerSignPsbtProtocolError);
```
gbarkhatov: **One matcher-less assertion covers four cases across three different production sites.**

The four `it.each` rows exercise `clientCommands.ts:373` (unknown command code), `clientCommands.ts:147` (unknown preimage), `expectedSignatures.ts:278` (unparseable varint) and `expectedSignatures.ts:297` (truncated augm) — but `toBeInstanceOf` can't tell them apart. A regression that collapsed all four into the generic `signPsbtLoop.ts:104` wrapper would leave this green.

`signPsbtPrepare.test.ts:219` already uses the better pattern. Suggested: add a per-case message regex to the `it.each` tuple.

packages/babylon-ledger-vault-signer/src/signPsbtPrepare.ts
```diff
@@ -0,0 +1,209 @@
...
+const DEPOSITOR_X_ONLY_HEX_RE = /^[0-9a-f]{64}$/;
+const PSBT_HEX_RE = /^(?:[0-9a-fA-F]{2})+$/;
+
+/**
+ * #2221/#2222 seam. Both fields default to 32 zero bytes: all-zero `wallet_id`
+ * routes into the vault validators (`has_no_wallet_policy`,
+ * `base:init_global_state.c:192-198`), and a non-zero `wallet_hmac` is
+ * `SW_NOT_SUPPORTED` — B1 hard-zeroes both.
+ */
+export interface SignPsbtHeaderOptions {
```
gbarkhatov: **No production caller — this is the seam the vendor README refused one PR ago.**

`prepareSignPsbt` always calls `buildSignPsbtCdata(merkelized)` bare at :203, so `walletId`/`walletHmac`, this interface, and the 32-byte length guard exist only for `__tests__/signPsbtPrepare.test.ts:86-120`.

`src/vendor/ledger-bitcoin/README.md` drew exactly this line for `getInputEntriesOfType`: "lands with its consumer, not before — zero dead code". Same rule in CLAUDE.md's ZERO DEAD CODE POLICY and TEST PHILOSOPHY ("never write tests for functions that don't exist in production code").

Suggested: drop it until #2221 needs it — `buildSignPsbtCdata` hard-zeroing both fields is B1's actual contract — or note in the PR description why this seam is the exception.

packages/babylon-ledger-vault-signer/vite.config.ts
```diff
@@ -8,10 +8,13 @@ export default defineConfig({
...
-      // until #2219) — keep their declarations out of the published dist too,
-      // so nothing vendored ships until the code is actually wired in.
-      exclude: ["src/**/__tests__/**", "src/vendor/**"],
+      // Vendored files stay IN the declaration program (index.ts reaches them
+      // through signPsbt.ts; excluding them makes tsc report TS6307 on every
+      // vendor import), but their declarations are dropped before write.
+      exclude: ["src/**/__tests__/**"],
+      // No vendored .d.ts ships: the first-party modules type the vendored
+      // surfaces structurally, so no emitted declaration references src/vendor.
+      beforeWriteFile: (filePath) => (/[\\/]vendor[\\/]/.test(filePath) ? false : undefined),
```
gbarkhatov: **Two issues with this predicate.**

1. `vite-plugin-dts@4.5.4` calls `beforeWriteFile(path, content)` from `writeOutput`, and every call site resolves against `outDir` — so `filePath` is **absolute**. A checkout or CI workspace under any directory containing a `vendor` path segment would make this drop *every* declaration and silently publish a package with no types. Suggested: match on the path relative to `outDir` and drop only `vendor/**`.

2. Returning `false` suppresses the write without checking whether any surviving `.d.ts` still references the suppressed file, and `tsc -b --noEmit` passes because it happily type-checks `src/vendor`. It is clean today — the emitted declarations reference only first-party modules plus `@ledgerhq/device-management-kit` and `buffer`, and the three leak vectors (`let merkelized: MerkelizedPsbt`, `new ClientCommandInterpreter`, the `psbtIn` enum) are all closed by the explicit return annotations and the structural interfaces. But drop one annotation and `pnpm build` ships broken types green, with only the prose checklist in the vendor README as defence. Suggested: a build gate asserting no emitted `.d.ts` references `/vendor/` in an import or export position.

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
```diff
@@ -0,0 +1,389 @@
...
+      const leafHashHex = payload.subarray(offset + 1 + X_ONLY_KEY_BYTES, augmEnd).toString("hex");
+      const signature = payload.subarray(augmEnd);
+      if (signerXOnlyHex !== expectation.expectedSignerXOnlyHex) {
+        throw new LedgerYieldMismatchError("wrong-signer-key", inputIndex);
+      }
+      if (!expectation.expectedLeafHashHexes.has(leafHashHex)) {
+        throw new LedgerYieldMismatchError("unknown-leaf-hash", inputIndex);
+      }
+      if (signature.length !== SCHNORR_SIG_BYTES) {
+        throw new LedgerYieldMismatchError("non-default-sighash", inputIndex);
```
gbarkhatov: `non-default-sighash` is the wrong kind for a short signature. This branch (and the keypath twin at :340) reports every non-64-byte signature this way. That's accurate for 65 bytes — a Schnorr signature with an appended non-default sighash byte — but misleading for anything shorter, which is a truncated or corrupt YIELD payload, not a sighash choice.

Suggested: `< 64` → a protocol/truncation error; reserve `non-default-sighash` for `> 64`.

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
```diff
@@ -0,0 +1,389 @@
...
+    if (leafEntries.length > 0) {
+      // Exactly one leaf per input for now — >1 is an ambiguous leaf set,
+      // mirroring fw `sign_psbt_validate.c:2981,3004` and the ts-sdk rule.
+      if (leafEntries.length > 1) {
+        throw new LedgerSignPsbtProtocolError(
+          `input ${inputIndex} carries an ambiguous leaf set (${leafEntries.length} TAP_LEAF_SCRIPT entries)`,
+        );
+      }
+      const value = leafEntries[0].value;
+      if (value.length < 1) {
```
gbarkhatov: A `value.length === 1` (value = `[0xc0]`, empty script) passes this guard, and :175-176 then hashes `taggedHash("TapLeaf", c0 00)` — a well-defined hash for a leaf that can never be satisfied, since an empty tapscript leaves an empty stack. bip174 accepts it too (it only inspects the last byte), so nothing else catches it.

Suggested: require `value.length >= 2`.

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
```diff
@@ -0,0 +1,389 @@
...
+  readonly depositorXOnlyHex: string;
+}
+
+// bitcoinjs `initEccLib` re-verifies only when handed a DIFFERENT instance —
+// init once so a second tiny-secp256k1 copy in the bundle can't thrash the cache.
+let eccInitialized = false;
+
+/** BIP-86/BIP-341 keypath-only P2TR scriptPubKey of an x-only internal key. */
+function bip86OutputScript(xOnlyKeyHex: string): Buffer {
+  if (!eccInitialized) {
```
gbarkhatov: The comment says this guards against a second tiny-secp256k1 copy thrashing the cache, but the flag can't do that — if another copy calls `initEccLib(other)` after us, this guarantees we never restore ours. bitcoinjs's `initEccLib` already no-ops on instance identity (it only runs `verifyEcc` when the instance differs), so calling it unconditionally is both simpler and strictly safer.

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
```diff
@@ -0,0 +1,389 @@
...
+        kind: "taproot-keypath",
+        expectedOutputKeyHex: script.subarray(2).toString("hex"),
+      });
+      continue;
+    }
+
+    // Not signed by the device (Payout input 1, NoPayout inputs 1-2 today).
+    // Ownership scan: a depositor-owned UTXO here means our builder dropped
+    // the device-required metadata — fail before burning a ceremony.
+    const witnessUtxo = psbt.getInputWitnessUtxo(inputIndex);
```
gbarkhatov, line 318: The "depositor-owned UTXO with no signing metadata" canary only recognises P2TR and P2WPKH, so a P2SH-P2WPKH or P2PKH depositor UTXO slips through silently. Either extend the scan or say in the comment that those script types are deliberately out of scope — an incomplete canary reads as a complete one.

packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
```diff
@@ -0,0 +1,143 @@
...
+    if (opts.signal?.aborted) {
+      // Stop sending — the CONTINUE for this round is never sent.
+      throw new LedgerSignPsbtAbortedError();
+    }
+    lastSentApdu = { cla: CLA_FRAMEWORK, ins: INS_CONTINUE, p1: P1_CONTINUE, p2: P2_CONTINUE, data: continueData };
+    response = await send(lastSentApdu);
+  }
+
+  if (response.sw !== SW_OK) {
+    const terminal = classifyStatusWord(response.sw, { ins: lastSentApdu.ins, p1: lastSentApdu.p1 });
```
gbarkhatov: Only `{ ins, p1 }` is passed, so `classifyStatusWord`'s `0x6E00` app hint (`rawApdu.ts:120-124`) is never populated on this path — even though `errors.ts:143-146` documents the ">5 s of host silence → app exits to the dashboard → host sees 0x6E00" case as the reason that hint exists. `dmkApduSender.ts` supplies `appName`/`appVersion` on the ceremony seam; suggested: thread them through `RunSignPsbtLoopOptions`, or bind them into the `RawApduSender`.

services/simple-staking/vite.config.ts
```diff
@@ -59,7 +59,11 @@ export default defineConfig({
     tsconfigPaths({
       projects: [resolve(__dirname, "./tsconfig.lib.json")],
     }),
-    nodePolyfills({ include: ["buffer", "crypto"] }),
+    // "buffer" is deliberately not included (vault does the same): including it
```
gbarkhatov, line 62: The change is right, and I checked the mechanism rather than assuming it: in `vite-plugin-node-polyfills@0.23.0`, `include` feeds only `resolve.alias`, while the Buffer-global machinery (the dev esbuild banner, `optimizeDeps.esbuildOptions.inject`, and the build-time `@rollup/plugin-inject` map) keys solely off `globals.Buffer`. So the injected global survives in both dev and build — no `Buffer is not defined` risk. And this PR is the first to put `import { Buffer } from "buffer"` into the signer's `dist`, which the alias would have rewritten to a bare shim specifier unresolvable from that package. The fix is necessary and correct.

One loose end: with the alias gone, `bitcoinjs-lib@6.1.7` — a direct declared dependency here — does `require('buffer')` without declaring it, so it now resolves only through pnpm's hoisted `.pnpm/node_modules/buffer`. `services/vault` avoids that by declaring `"buffer": "6.0.3"` directly (and by pinning `globals: { Buffer: true, global: false }` explicitly rather than relying on defaults). Suggested: add the same pinned dep here so the two apps match and the phantom-dependency coupling goes away.

packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
```diff
@@ -12,12 +12,16 @@ their own license files. Vendored source:
 - Source: https://github.com/LedgerHQ/app-bitcoin
 - Version: ledger-bitcoin@0.3.0 (commit 0a9e9e141f3340d29e7c6181177d4e5e9483a9f7)
 - Files: vendored under packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/
-  (source-in-tree; not yet reachable from the published bundle)
+  — the source is in-tree and, from #2219 onward, the compiled vendored code is
+  also bundled into the published dist/ (build strips the per-file provenance
+  headers, so this notice is the redistribution record; see the audit-boundary
+  section of that directory's README)
 - License: Apache License 2.0
```
gbarkhatov, line 19: One thing to confirm before the first publish that ships the vendored bundle: Apache-2.0 §4(d) requires propagating upstream's `NOTICE` file if one exists. `src/vendor/ledger-bitcoin/README.md` asserts upstream carries none, but that isn't verifiable from the lockfile (the package isn't in the pnpm store) — worth one manual check against `LedgerHQ/app-bitcoin` at `0a9e9e14`.

§4(a) and §4(b) look satisfied: the full licence text is here and shipped via `files`, and because `sourcemapExcludeSources: false`, the emitted `.map` files carry every per-file provenance header verbatim. Worth being aware that this makes the sourcemaps compliance-load-bearing while also being the bulk of the tarball; a short `output.banner` pointing at this file would decouple the two.

gbarkhatov (dismissed):
Deep review of the signing core, the vendored diff, the test gate, and the packaging blast radius. Approving — the security architecture is sound and the standard here is high — but a few of these are worth resolving before this path carries value.

**What holds up.** The expected-signature table is built from the *same* `MerkelizedPsbt` instance that produced the commitment, so there is no parse/serialize gap on that side. Every YIELD is asserted before it is recorded (the `onYield` seam genuinely runs before `results.push`). Completion is set equality in both directions plus a size check. The merge writes signature fields only and never finalizes. The 22 goldens come from a genuinely independent oracle — `ledger-bitcoin==0.4.0` driven by `gen_vectors.py` with a hard version guard and a self-cross-check — and `generated__*` are firmware fixture names, not self-produced values. The Speculos stage verifies a real device signature against an independently computed BIP-341 sighash *and* carries a negative control, and its ordered stages guard with `toBeDefined()` before every cast so they cannot pass vacuously. `approveOnScreen` refusing to blind-press the idle dashboard is a nice touch.

**Worth resolving.** The four I would prioritise: the parser differential on `signPsbtPrepare.ts:166` (reproduced locally — a crafted PSBT makes the two parsers see different key sets and the merge silently succeeds), the missing already-signed/finalized gate on `:154`, the optional `signal` on an intentionally unbounded loop, and the `TAP_LEAF_SCRIPT`-first `continue` in `expectedSignatures.ts` that skips every ownership pin on that input. The rest are hardening, test-gate, and packaging notes.

**Three notes with no diff line to attach to:**

- `psbtv2.ts:430-434` rejects a spec-valid v0 PSBT carrying an explicit `PSBT_GLOBAL_VERSION = 0` (BIP-174 permits it; bitcoinjs accepts it), and throws a raw `RangeError` if the VERSION value is under 4 bytes. Separately, `psbtv2.ts:442` → `uint32LE` throws an untyped `RangeError` for `nVersion >= 0x80000000`, because bitcoinjs reads nVersion signed — `version >>> 0` or an explicit typed rejection would read better. Both are fail-closed and pre-I/O. TRUC/v3 is unaffected and correct.
- PR description: the "181-trace byte-replay" is 175 real-flow traces plus 6 synthetic deep-tree, and 4 of the 181 are YIELD traces whose `response_hex` is `""` (the fixture notes they are synthetic placeholders), so those assert `"" === ""`. The in-repo comment at `signPsbtLoop.test.ts:153` already says 175 and is accurate — worth aligning the body.
- The Speculos E2E is `skipIf(SPECULOS_URL === "")` and nothing sets that in CI, so it is collected-and-skipped — and it is the only place a real Schnorr signature meets a real sighash (every offline test uses `"cd".repeat(64)`). The last line of the PR body is honest about this; the earlier use of "gated" reads as if CI enforces it.

Worth filing separately: the package still has no component `SECURITY_MODEL.md` and is covered only by the repo-level `SECURITY.md`. Pre-existing, but this PR is the point where it acquires a signing engine.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculosClient.test.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import { describe, expect, it } from "vitest";
+
+import type { RawApduSender } from "../../rawApdu";
+import { getAppAndVersion } from "./speculosClient";
+
+/** A sender that answers 0x9000 with the given response body. */
+function senderReturning(dataHex: string): RawApduSender {
+  return async () => ({ sw: 0x9000, data: Buffer.from(dataHex, "hex") });
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/signPsbtPrepare.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/signPsbtPrepare.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/signPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
jrwbabylonlab: fixed the control-block commitment check; not planning the other two — PegIn always carries TAP_INTERNAL_KEY next to its leaf, and the ownership-scan scope is deliberate.

packages/babylon-ledger-vault-signer/src/signPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/errors.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
jrwbabylonlab: Added a test per keypath throw site; five of them had no test that failed when you deleted the line. Byte-flipped the key rather than swapping inputs, since both pre_pegin inputs share an output key.

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtPrepare.test.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts

packages/babylon-ledger-vault-signer/src/signPsbtPrepare.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/vite.config.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
jrwbabylonlab: not planning — P2PKH isn't structurally distinguishable there; the scope is documented in the comment.

packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
jrwbabylonlab: fixed

services/simple-staking/vite.config.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/THIRD-PARTY-NOTICES.md
jrwbabylonlab: resolved — no NOTICE file at the pinned rev, nothing to propagate.

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtPrepare.test.ts
```diff
@@ -0,0 +1,328 @@
...
```
gbarkhatov: This one didn't land. The reply said the fixture was fixed to go red with the guard and the matcher tightened, but neither follow-up commit touched this hunk — the fixture is still `70736274ff 01fb 0401000000 00` and the matcher is still the loose `/PSBT rejected at parse/`.

Ran it against `3bc051b`; the thrown message is:

```
SIGN_PSBT protocol failure: PSBT rejected at parse (merge target): Format Error: Only one UNSIGNED_TX allowed
```

The fixture has no `PSBT_GLOBAL_UNSIGNED_TX`, so the merge-target `BitcoinjsPsbt.fromHex` gate throws first and `PsbtV2.deserialize`'s version rejection is never reached — delete that guard in `psbtv2.ts` and this test stays green. The loose matcher is what hides it, since it also matches the `(merge target)` wrapper.

A fixture that does reach the guard is any bitcoinjs-acceptable v0 PSBT with a global VERSION keypair (`01fb 0401000000`) spliced in after the magic. Verified locally — bitcoinjs accepts it and prepare throws `PSBT rejected at parse: Only PSBTs of version 0 or 2 are supported`:

```
70736274ff01fb040100000001005e020000000101010101010101010101010101010101010101010101010101010101010101010000000000ffffffff0184030000000000002251200303030303030303030303030303030303030303030303030303030303030303000000000001012be80300000000000022512002020202020202020202020202020202020202020202020202020202020202020000
```

With `/PSBT rejected at parse: Only PSBTs of version 0 or 2 are supported/` as the matcher — the colon-space is what excludes the merge-target wrapper.

Test-only coverage, so not blocking the approval.

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
```diff
@@ -0,0 +1,758 @@
...
+    const prepared = prepareFromVector("generated__deposit-flow__pegin__0");
+    const script: ScriptedExchange[] = [
+      { expectApduHex: initialApduHexOf(prepared), respondSw: 0xe000, respondDataHex: requestHex },
+    ];
+    const { send, sent } = createScriptedSender(script);
+
+    await expect(runSignPsbtLoop(send, prepared, { signal: NEVER_ABORTED })).rejects.toBeInstanceOf(
+      LedgerSignPsbtProtocolError,
+    );
+    expect(sent()).toBe(1);
```
gbarkhatov, line 758: Smaller version of the same thing: the reply said this was split into per-case regexes, but the `it.each` still asserts nothing beyond `toBeInstanceOf(LedgerSignPsbtProtocolError)` for all four rows.

The gap is genuinely narrower than my original finding, though. The test you added below at ~L668 pins `outcome.detail` to `/^unparseable YIELD payload:/` and asserts the message does not contain "client command failed", which does kill the collapse-into-the-generic-wrapper mutation I described. What is left is that these four rows still cannot tell the `clientCommands.ts` unknown-command-code path from the unknown-preimage one, and the truncated-augm row (`expectedSignatures.ts`) carries no message assertion at all.

Non-blocking — a message regex per tuple row next time you touch the file.

gbarkhatov (dismissed):
Approving. I re-verified all 18 of my threads against `3bc051b` rather than reading the replies, and the substantive ones landed: the non-minimal CompactSize rejection in `parseVarint`, the pre-I/O `assertNoConflictingSignatures` gate, `signal` required on both `signVaultPsbt` and the loop, the control-block refold/parity check, identity-only `collectedYields` plus `yieldedCount` on abort, `<64` bytes as a truncation error rather than `non-default-sighash`, `MIN_TAP_LEAF_SCRIPT_VALUE_BYTES`, unconditional `initEccLib`, `appIdentity` threaded into `classifyStatusWord`, `walletId`/`walletHmac` dropped, the `beforeWriteFile` predicate anchored to `outDir` with the post-build `dist/index.d.ts` check, the keypath throw-site tests, and the payload logs gone from the Speculos suite. On my checkout `pnpm test` is 509 green (e2e skipped) and `pnpm lint` is clean.

I also re-checked the two you scoped down rather than fixed, and both hold up. On `finalScriptWitness`: I confirmed empirically that `psbt.updateInput` succeeds on a finalized input under bitcoinjs 6.1.7, so a finalized PSBT genuinely is not the burn-then-throw class that gate exists for — your reasoning is right. On the Apache-2.0 §4(d) question I looked at `LedgerHQ/app-bitcoin` at `0a9e9e14` myself: no NOTICE at the repo root or in `bitcoin_client_js`, so there is nothing to propagate.

Two test-only claims did not make it into the code — flagged inline, neither blocking.

Still unanswered, and worth one line each either way: CodeRabbit's `finalScriptWitness`/`finalScriptSig` preflight thread (the residual there is narrow — a hand-crafted PSBT carrying both a final witness and a `tapLeafScript` runs the full ceremony for a signature the final witness then overrides), and its varint-constants thread (the CompactSize widths `2`/`4`/`8` are still inline; only the minimum-value bounds got named).

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtPrepare.test.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
jrwbabylonlab: Also mine — the rows were untouched. Each row carries its own detail regex now, so the unknown-command-code, unknown-preimage and two truncation paths can't satisfy each other.

packages/babylon-ledger-vault-signer/src/signPsbtPrepare.ts
```diff
@@ -0,0 +1,247 @@
...
+      }
+      continue;
+    }
+    if (input.tapKeySig) {
+      throw new LedgerSignPsbtProtocolError(
+        `input ${inputIndex} already carries a tapKeySig and this round signs its keypath — pass an unsigned PSBT`,
+      );
+    }
+  }
+}
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/varint.ts
```diff
@@ -96,11 +108,34 @@ export function parseVarint(data: Buffer, offset: number): readonly [bigint, num
...
+    // src/serialize.h:333-358, blob a1395c47 — "non-canonical ReadCompactSize()").
+    if (value < minimum) {
+      throw RangeError(
+        `Non-minimal varint at offset ${offset}: a ${size + 1}-byte CompactSize must encode a ` +
+          `value of at least ${minimum}; re-encode it in its canonical (shorter) form`,
+      );
+    }
 
-    return [smallEndianToBigint(data, offset + 1, size), size + 1];
+    return [value, size + 1];
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — those lines are verbatim upstream; renaming them creates divergence we re-apply on every re-diff (rationale in the vendor README).

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
jrwbabylonlab: not planning — leaf-script internals are validated on-device against the approved intent; the host pins the leaf commitment only.

packages/babylon-ledger-vault-signer/vite.config.ts
jrwbabylonlab: The point-2 gate is in now too — the build scans every emitted d.ts for vendor references in import position (scripts/check-dist-types.cjs); the earlier reply had only landed the existence check.

jeremy-babylonlabs (approved):

PR #2265, opened 2026-08-15
test(vault): make the visual check look at the real app, and at the deposit form

### LINKED ISSUE #1592: e2e: Single-vault deposit happy path

jonybur:
## What
The visual check was not looking at the app. Ten of the twelve pictures it
took were photographs of the same "Something went wrong" card, so it reported
"no visual changes" whatever you changed. This gives it a working app to
photograph, and adds the deposit form to what it looks at.

## Why
It passed #2241 clean - a change that visibly moves a notice around inside the
deposit form. It was not wrong about what it compared; it had simply never
been able to see that screen, or most others. The pictures failed the same way
on both sides of every comparison, matched perfectly, and reported success.

Two separate reasons. The app needs a wallet and a server to show anything,
and the check gave it neither, so most pages fell back to an error card. And
the deposit form is a pop-up you reach by connecting a wallet and clicking
Deposit, which no test could do.

## How
The repo already contained a recording of a real deposit made on devnet a
while back - every question the app asked and every answer it got. The check
now replays that recording, so the pages fill in with real numbers instead of
failing. It replays the run up to the moment the depositor was looking at the
deposit form, so what is on screen is one consistent moment rather than a mix
of before and after.

Wallets connect the way a browser extension connects, so nothing in the app
changed to let a test in. Four new pictures cover the deposit flow: the
dashboard with a wallet connected, the form, the form with an amount typed in,
and the split options panel open.

It also adds four ways for the check to fail loudly rather than quietly
accept a bad picture - a page that would not start, a page showing an error,
a question the recording cannot answer, and a server it stopped being able to
reach. Each one was tested by deliberately breaking the thing it watches.

Worth knowing: every one of the twelve existing pictures will show as changed
here, because they were error cards and are now the real app. That is the
point of the change rather than a regression. The check goes from 12 pictures
to 20 and takes about 40 seconds longer.

Also worth knowing: the recording under `services/vault/e2e/artifacts/` is now
what every picture is drawn from, so it is no longer just an old test
artifact. If it goes missing the check fails and says so.

Does not close #1592, but unblocks it: connecting a wallet from a test was the
piece that ticket was waiting on, and it now works. The wallets here still do
not sign anything - they deliberately refuse - so the deposit happy path that
ticket describes is still to be built on top.


services/vault/e2e/fixtures/replay/recording.ts
```diff
@@ -0,0 +1,211 @@
...
+ * The peg-in recording committed to this repository.
+ *
+ * A single named constant rather than a glob: which run is replayed decides
+ * what every captured screen contains, so it is a deliberate choice, not
+ * whichever directory happens to sort first.
+ */
+export const PEGIN_RECORDING_PATH = path.join(
+  currentDir,
+  "../../artifacts/2026-07-20-173643-pegin/recording/http.jsonl",
+);
```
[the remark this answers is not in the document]
jonybur: Stale - this is what the head commit (28fe993) fixed. `PEGIN_RECORDING_PATH` now resolves to `recorded-run.jsonl` beside the loader, and that file is tracked (`git ls-files services/vault/e2e/fixtures/replay/` lists it; `git check-ignore` does not match it). Nothing reads `e2e/artifacts` at test time any more - only `scripts/build-replay-fixture.mjs`, which is a human-run regeneration tool. Verified by running the capture from a clean checkout: 14/14 pass, 20 screenshots.

services/vault/e2e/fixtures/__tests__/replay.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    // recording - it exists only as one leg of an aggregate3 batch. Answering
+    // it on its own is the whole point of taking batches apart, and is what
+    // lets the app re-compose its reads differently from the recording.
+    const pauseState = encodeFunctionData({
+      abi: parseAbi(["function pauseState() view returns (uint8)"]),
+    });
+
+    expect(
+      chain.answerCall(
+        "0xb331467c4db13dccc77fa66c2d185b74ed57ab80",
```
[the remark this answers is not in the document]
jonybur: Declining. CLAUDE.md's test philosophy asks for the opposite here: "Prefer explicit values over magic constants imported from elsewhere." These addresses are the test's own inputs, and at the `expect(chain.unanswered).toEqual([{ target: "0xb331..." }])` assertion they are the expected value - substituting the constant would assert the fixture against the same source the fixture code reads, which is tautological. On a re-record, `contracts.ts` and `recorded-run.jsonl` move together and these tests go red immediately, which is the signal you want.

services/vault/e2e/fixtures/pageWallets.ts
```diff
@@ -0,0 +1,265 @@
...
+            detail: Object.freeze({ info, provider: ethProvider }),
+          }),
+        );
+      };
+      // Announce on request AND immediately: wagmi listens before asking, but
+      // the order of the two is not guaranteed across a navigation, and a
+      // provider announced only once - before wagmi is listening - is a
+      // provider nobody sees.
+      window.addEventListener("eip6963:requestProvider", announce);
+      announce();
```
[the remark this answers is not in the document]
jonybur: Declining. These are protocol identifiers used exactly once each - `eth_accounts`, `eth_chainId`, `net_version`, `wallet_switchEthereumChain` in one `switch`, and the two `eip6963:*` names in one `dispatchEvent`. A constant would restate the string without adding information. The code also runs inside `page.addInitScript`, a serialization boundary where module-scope constants are out of scope, so this means widening the argument tuple purely to carry string aliases into the browser. The values that genuinely were non-obvious (`UNISAT_VERSION`, `UNISAT_SIGNET_CHAIN`, `BLANK_ICON`) are already hoisted and passed that way.

services/vault/e2e/visual/capture.ts
```diff
@@ -0,0 +1,165 @@
...
+ * this catches is everything the recording does not cover: it fails closed
+ * instead of reaching a live host, which would make a capture vary run to run
+ * and, on a fork PR, leak the request.
+ */
+async function blockOffsiteRequests(page: Page): Promise<void> {
+  await page.route("**/*", (route) => {
+    const { hostname } = new URL(route.request().url());
+    const isLocal = hostname === "localhost" || hostname === "127.0.0.1";
+    return isLocal ? route.continue() : route.abort();
+  });
```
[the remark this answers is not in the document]
jonybur: Declining - nit. `blockOffsiteRequests` is six lines; `page.route("**/*")` and `hostname === "localhost" || hostname === "127.0.0.1"` are self-describing at the point of use, and the doc comment above already states the policy. Naming them puts the reader one indirection further from the two values that define the boundary.

services/vault/e2e/visual/capture.ts
```diff
@@ -0,0 +1,165 @@
...
+      `baseline.`,
+  ).toHaveCount(0);
+
+  await expect(
+    page.getByTestId("app-error-state"),
+    `${label} captured the app's error state instead of the page. The app ` +
+      `booted and then failed - usually a read the recorded backend cannot ` +
+      `answer. Photographing it would bake "Something went wrong" in as the ` +
+      `expected look, and it diffs clean against itself forever.`,
+  ).toHaveCount(0);
```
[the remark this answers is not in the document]
jonybur: Declining. No such shared constant exists today, and creating one means importing app source into the Playwright runner - the exact hazard documented at `e2e/visual/targets.ts:12-19` (importing `src` pulls the network-config singleton in and throws before collection). The string appears in three places, and both `src` sites already carry the convention CLAUDE.md prescribes for load-bearing testids - `// \`app-error-state\` is read by the visual capture (...)` at `global-error.tsx:39` and `AaveConfigContext.tsx:62`. Grep-ability is what keeps these in sync here.

services/vault/e2e/visual/routes.visual.spec.ts
```diff
@@ -78,32 +41,10 @@ for (const target of VISUAL_TARGETS) {
...
-      const fileName = screenshotFileName(target, viewport);
-      const buffer = await page.screenshot({ fullPage: true });
-      await fs.writeFile(path.join(OUTPUT_DIR, fileName), buffer);
-
-      // A zero-byte or absurdly small PNG means the screen never
-      // painted; that must fail the capture rather than silently
-      // become an "expected" baseline the next run diffs against.
-      expect(buffer.byteLength).toBeGreaterThan(1000);
+      await capture(page, screenshotFileName(target, viewport));
+      await assertAppRendered(page, backend, target.name);
```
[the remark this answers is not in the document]
jonybur: Declining, having traced it against the real render path. For these six targets the gate above already catches what `requiredBoundaries` would. Five of the six (`/`, `/vaults`, `/loans`, `/liquidations` via `AaveOverlayLayout`; `/activity` via `ActivityWithProviders`) mount `AaveConfigProvider`, whose fallback renders `data-testid="app-error-state"` when `fetchAaveAppConfig` rejects - and that call needs both boundaries these disconnected screens use: the indexer GraphQL request plus two fail-closed on-chain reads (`getCoreSpokeAddress` and the `VAULT_BTC_RESERVE_ID` cross-check, `fetchConfig.ts:204-240`). Lose `eth-rpc` or `graphql` by any mechanism and the error state renders, so the check fires. The deposit-flow spec declares all four because it is the one walk that genuinely renders-but-emptier when a boundary goes missing.

services/vault/scripts/build-replay-fixture.mjs
```diff
@@ -0,0 +1,124 @@
...
+ * today; this is what stops a future cut at a later step from quietly
+ * committing a bearer token.
+ */
+const SECRET_PATTERNS = [
+  /bearer\s+[\w.-]+/i,
+  /"authorization"/i,
+  /createDepositorToken/,
+  /mnemonic/i,
+  /private[_-]?key/i,
+];
```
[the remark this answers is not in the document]
jonybur: Declining, after reading both the fixture and the recorder. Line 34 is the indexer's `GetVaultsByDepositorFirstPage` query; `secret` is one of ~30 requested fields and the recorded response is `{"data":{"vaults":{"items":[]}}}` - empty. More to the point, that field is the HTLC preimage, which the indexer only holds *after* the depositor has revealed it on-chain to activate; by the time it could appear in a recording it is public chain data, from a throwaway devnet wallet on signet. It is not a credential. The credential class the scrub exists for is already covered structurally: `e2e/real/actions/recording.ts:265-268` writes only `{t, step, method, url, status, reqBody, resBody}` and no headers, so the VP session token cannot reach the file. I also scanned the committed fixture independently - the only >=64-char hex values are a block hash, a taproot scriptPubKey and a txid.

services/vault/scripts/build-replay-fixture.mjs
```diff
@@ -0,0 +1,124 @@
...
+  const kept = lines.slice(0, cutoff + 1).filter((entry) => {
+    // Matched against hostname LABELS, never as a bare substring. A plain
+    // `hostname.includes("sentry.io")` also matches `sentry.io.example.com`,
+    // which drops a host nobody meant to drop - the same flaw CodeQL caught
+    // in the loader's classifier.
+    const labels = new URL(entry.url).hostname.split(".");
+    return !DROPPED_HOSTS.some((dropped) =>
+      labels.some((label) => label === dropped || label.includes(dropped)),
+    );
+  });
```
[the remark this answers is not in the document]
jonybur: Declining. The state described is real but needs every exchange tagged with the requested step to belong to a dropped host - for `deposit-form`, a run where the form made no chain, indexer or mempool call at all. If it ever happened the failure is loud and self-describing on the very next run: `loadRecordedRun` throws naming the step and listing the steps the file does contain, while the developer is still at the terminal that just regenerated the fixture. The suggested fix also is not available as written - `build-replay-fixture.mjs` is a plain `.mjs` script and cannot import the TypeScript loader's classification.

services/vault/e2e/fixtures/replay/chain.ts
```diff
@@ -0,0 +1,271 @@
...
+export function buildRecordedChain(run: RecordedRun): RecordedChain {
+  let tables = tableCache.get(run);
+  if (!tables) {
+    tables = buildTables(run);
+    tableCache.set(run, tables);
+  }
+  const { exact, bySelector, methods } = tables;
+  const unanswered: UnansweredCall[] = [];
+
+  const answerCall = (to: string, data: Hex): Hex | null =>
```
gbarkhatov, line 263: **The selector fallback answers an unrecorded argument with a wrong-but-plausible value, and no gate can see it.**

The second tier drops the arguments and returns the *last* recorded return for `(target, selector)`. That produces a **wrong** answer rather than a **missing** one — so `unanswered` stays empty, `misses` stays empty, and every gate in `capture.ts` passes. It is the one path in this fixture that can put an invented number on a screenshot.

Decoded from the committed fixture, `getReservesPrices(uint256[])` on `0xcdcb913a…` is recorded three times, each with a one-element array:

```
ids=[0] -> [100000000]
ids=[1] -> [100000000]
ids=[2] -> [6469260700000]
```

Reserve **3** is `aaveConfig.vaultBtcReserveId` — vBTC, the asset the deposit form is about, and the one the recorded `getReserve(3)` is for. It is never priced. `getReservesPrices([3])` therefore falls back to reserve 2's (WBTC) price and renders a plausible BTC figure the recording never justified.

`getReservesPrices([0,1,2,3])` is worse: the fallback returns the recorded **one-element** array for a four-reserve request, so `prices[1..3]` come back `undefined`.

Same shape on `0xb54f1d4a…|0x9f9b1990` — 3 distinct calldatas, returns `10000000000000` vs `2500000000000000`, a 250× spread depending on which argument you happen to ask with.

The header at lines 20-24 says the exact tier "keeps two calls to the same getter with different arguments from collapsing into one answer". It does — but only for arguments the recording contains; for anything else the two tiers collapse exactly as described there. Exposing an `answeredBySelectorFallback` counter that `assertAppRendered` asserts empty, the way it already asserts `unanswered`, would turn the one silent-wrong path back into a loud one.

services/vault/e2e/fixtures/replay/chain.ts
```diff
@@ -0,0 +1,271 @@
+/**
+ * A read-only fake chain assembled from a recorded run.
+ *
+ * The dApp reads the chain almost entirely through Multicall3: of the 227
```
gbarkhatov: These figures describe the gitignored raw artifact, not the fixture this now loads.

Decoded from `recorded-run.jsonl` as committed: **19** `eth_call`s (all 19 `aggregate3`), **39** inner calls, **28** distinct `(target, selector)` pairs — against the 227 / 226 / 609 / 35 claimed here. Stale since 28fe993 swapped `e2e/artifacts` for the committed subset.

The reasoning the numbers support is still correct; it is only that this file's style invites a reader to trust a cited figure.

services/vault/e2e/fixtures/replay/index.ts
```diff
@@ -0,0 +1,354 @@
...
+    .replace(
+      /\/v1\/validate-address\/[^/]+$/,
+      `/v1/validate-address/${ANY_ADDRESS}`,
+    )
+    .replace(/\/tx\/[0-9a-f]{64}$/i, `/tx/${ANY_TXID}`)
+    .replace(/^.*\/api\//, "/");
+}
+
+/** Name a GraphQL request by its operation, falling back to the raw query. */
+function graphqlKey(body: string | undefined): string {
```
gbarkhatov: **The key discards the path and the method, so the indexer `/health` GET owns the empty key.**

The handler matches the entire GraphQL origin (`on(GRAPHQL_ORIGIN, () => true)`), but the key comes only from the POST body. The fixture's `GET /health` line has neither `reqBody` nor `resBody`, so it indexes under `""` — which is also the key for *any* request to that origin whose body is absent or unparseable.

`fetchHealthCheck` therefore works by coincidence rather than by design, and anything else bodyless that ever reaches that origin is answered `200` with `"{}"` (via `recorded.resBody ?? "{}"`) — a body carrying neither `data` nor `errors`, which the client renders as an empty state **with no miss recorded**. Keying on `method + pathname + operationName` would close it.

Second, latent: `indexBy` keys on `operationName` alone, so `variables` are dropped. `GetVaultsByProvider` is recorded once, for the single provider in the recorded `GetAppProviders`. The day devnet has two, the app issues one query per provider, all of them key to the same string, and every provider row on the captured screen shows provider #1's vault list and TVL — silently. Unlike `chain.ts` there is not even an exact tier in front of it.

services/vault/e2e/fixtures/replay/index.ts
```diff
@@ -0,0 +1,354 @@
...
+      }
+      return route.fulfill({
+        status: recorded.status,
+        contentType: "application/json",
+        body: recorded.resBody ?? "{}",
+      });
+    }),
+  );
+
+  // A getter, not a snapshot: the count keeps rising while the page runs, and
```
gbarkhatov: These two comments say the same thing, and the first one is inaccurate — `served` is a plain object handed out by reference; there is no getter. Lines 347-349 can go.

services/vault/e2e/fixtures/replay/recording.ts
```diff
@@ -0,0 +1,242 @@
...
+ * wallet picker's remote images - the injected wallet needs none of it) and
+ * the dApp's own origin (documents, bundles, the wasm blob - served by the
+ * dev server under test).
+ *
+ * The dApp's own hosts are matched on a label rather than a full domain -
+ * `mempool`, `indexer-api` - because those names are deployment-specific and
+ * a re-recording against another environment must still classify. They are
+ * matched against hostname LABELS, never as a bare substring.
+ */
+function hasLabel(hostname: string, label: string): boolean {
```
gbarkhatov, line 112: Two small things about the classifier.

`classify()` anchors on `sentry.io` / `web3modal.org` / `walletconnect.org`, but `build-replay-fixture.mjs`'s `DROPPED_HOSTS` drops by label substring — so `walletconnect.com`, `web3modal.com` and a self-hosted `sentry.*` are dropped by the builder and fall into the `eth-rpc` catch-all in the loader. Reachable in practice, since `ReplayOptions.recordingPath` explicitly allows pointing this at a raw `e2e/artifacts/**/http.jsonl` that never went through the builder. Two hand-maintained copies of one policy is the drift this PR is careful to avoid elsewhere (`MOCK_ENV_VARS` is spread, not copied, for exactly that reason).

The docstring above also says `classify` returns null for "the dApp's own origin (documents, bundles, the wasm blob)". There is no such check — the `demo` label happens to cover the devnet host, and a recording made against `localhost` would fall through to `eth-rpc`.

services/vault/e2e/visual/capture.ts
```diff
@@ -0,0 +1,199 @@
...
+  ).toBeGreaterThan(MIN_CAPTURE_BYTES);
+  return { fileName, buffer };
+}
+
+/**
+ * Write staged photographs to disk. Call only after {@link assertAppRendered}
+ * has passed - reaching here is the test's statement that what it photographed
+ * is worth diffing against.
+ */
+export async function writeCaptures(
```
gbarkhatov, line 225: **Nothing counts the files, so a symmetric capture failure still prints "No visual changes".**

Staging in memory and flushing only after the gates pass is the right design, and it is what makes the CI step's `continue-on-error` sound. But it converts "photographed an error card" into "photographed nothing" — and the diff step only checks that each surface's directory is *non-empty*. `visual-diff.mjs` builds its name set from the union of the two sides, so a screen absent from **both** never enters the result set at all.

Because both sides run the same stashed harness and the same committed fixture, every fixture- or harness-caused failure is symmetric by construction. Concretely: the recording goes stale, the miss gate fires in `depositFlow.visual.spec.ts` at both viewports on both sides, all 8 deposit PNGs are correctly withheld, the 12 route shots still land — `COMPARED=1`, `MISSING=""`, `changed_count=0` → ✅ "No visual changes", for a run in which 8 of 20 screens were never taken.

`targets.ts` already holds the manifest. Asserting a per-surface expected count in the diff step, and treating a shortfall the way `MISSING` is treated, closes the last silent-green path this PR leaves open.

services/vault/e2e/visual/routes.visual.spec.ts
```diff
@@ -78,32 +42,11 @@ for (const target of VISUAL_TARGETS) {
...
-      const fileName = screenshotFileName(target, viewport);
-      const buffer = await page.screenshot({ fullPage: true });
-      await fs.writeFile(path.join(OUTPUT_DIR, fileName), buffer);
-
-      // A zero-byte or absurdly small PNG means the screen never
-      // painted; that must fail the capture rather than silently
-      // become an "expected" baseline the next run diffs against.
-      expect(buffer.byteLength).toBeGreaterThan(1000);
+      const shot = await capture(page, screenshotFileName(target, viewport));
+      await assertAppRendered(page, backend, target.name);
```
gbarkhatov: `requiredBoundaries` is omitted here, so the gate that `capture.ts:123-143` calls "the one that catches what the two above cannot" — and documents verifying by pointing `NEXT_PUBLIC_ETH_RPC_URL` at a dead port — covers only the 8 deposit-flow shots, not these 12. These are the screens that were error cards in the first place.

Partly mitigated for eth-rpc: a dead RPC makes `fetchAaveAppConfig` throw and `app-error-state` render, so gate 2 catches it. Not mitigated for a screen whose only missing boundary is mempool or graphql — that one renders emptier and diffs clean against itself forever, which is precisely the failure the gate was written for.

`VISUAL_TARGETS` is already data. A `requires: readonly RecordedBackend[]` field per target, passed through here, closes it and needs no Playwright knowledge at the call site.

services/vault/e2e/visual/depositFlow.visual.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+      height: viewport.height,
+    });
+
+    const backend = await preparePage(page);
+    await injectPageWallets(page, {
+      btcAddress: RECORDED_DEPOSITOR.BTC_ADDRESS,
+      // Compressed form of the address's x-only key. The wallet-connector
+      // adapters expect a 33-byte key; the recording holds no internal key,
+      // and nothing in a capture signs, so the output key stands in for it.
+      btcPublicKeyHex: `02${RECORDED_DEPOSITOR.BTC_X_ONLY_PUBLIC_KEY}`,
```
gbarkhatov: **A question rather than a claim: this reports a public key that does not match the address beside it, and the connector has a guard for exactly that.**

`BTC_X_ONLY_PUBLIC_KEY` is the tweaked taproot **output** key — the co-located test derives it from the address — and `getTaprootAddress` tweaks whatever it is handed:

```
p2tr(internal = 2503070eba5f…) = tb1plzqtvucpdmetwegp8twumz8p8z4ncn0ul7295g7xglnh8tpj70rscgf75f
RECORDED_DEPOSITOR.BTC_ADDRESS = tb1py5pswr46tl0y0qapun6gdm760a7rkkxcxclchu6072c3zykl7cnqjckc8a
```

`validateAddressWithPK` (`core/utils/wallet.ts:85-88`) is exactly `address === getTaprootAddress(publicKey, network)` for a `tb1p` address, and it sits on the connect path at `useWalletConnectors.tsx:112-132`, behind `if (!visible) return`. On failure it displays a "Public Key Mismatch" screen whose only buttons are Cancel / Continue Anyway — which would time out the `Connect` wait in `connectInjectedWallets`.

CI is green in 22s, so the guard is not firing, and I could not work out why from the source alone — that is the question. Either way `contracts.ts:53-58`'s claim that the difference "does not reach a captured screen" is load-bearing and currently unproven. Recording the wallet's internal key, or deriving the fixture address from a key the fixture owns, would make the pair self-consistent and retire the question.

services/vault/e2e/visual/depositFlow.visual.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+      await capture(
+        page,
+        flowScreenshotFileName(DEPOSIT_FLOW_STOPS.amountEntered, viewport),
+      ),
+    );
+
+    // The collapsed selector's header carries the current choice as its label.
+    // `.first()` is safe HERE and only here: the panel is still collapsed, so
+    // the option row with the same words has not been rendered yet.
+    await dialog.getByRole("button", { name: /do not split/i }).first().click();
```
gbarkhatov: The justification above does not hold: core-ui's `Accordion` always renders its children and collapses with CSS (`AccordionDetails` does not unmount), so the option row carrying the same words **is** in the DOM before this click. Two elements match `/do not split/i`.

`.first()` resolves to the header today only because of DOM order, plus `getByRole` excluding accessibility-hidden elements. If the split ever auto-enables (`SimpleDeposit.tsx:246-253`) and the header label flips, `.first()` starts resolving to the option row, whose click *selects* "no split" instead of expanding the panel — and `deposit-split-options` becomes a photograph of a collapsed panel, identical on both sides. A testid on the accordion header would make this unambiguous.

services/vault/e2e/visual/depositFlow.visual.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    // names them together. Nothing has reached disk yet, so a failure here
+    // withholds all four stops - which is what makes the capture step's
+    // `continue-on-error` safe.
+    //
+    // All four boundaries are named because this walk genuinely needs all
+    // four - the chain for borrow power and the split minimum, the indexer
+    // for the vault list, the VP proxy for the provider, mempool for the
+    // balance. A screen missing any of them still renders, just emptier, and
+    // that is precisely the failure a screenshot cannot be trusted to show.
+    await assertAppRendered(page, backend, `deposit flow at ${viewport.name}`, [
```
gbarkhatov: The comment is right about `misses` and `served` — those accumulate across the walk. It is not right about the other two: `error-dialog` and `app-error-state` are point-in-time DOM queries, so stops 1-3 are only ever verified against the DOM as it stands at stop 4.

If the `connected` stop is photographed inside a React Query retry window while the error fallback is up, that frame is static, `waitForVisualStability` settles on it, and `deposit-connected--*.png` is a photograph of "Something went wrong"; the retry then succeeds, this check passes, and `writeCaptures` flushes all four. The two DOM gates are cheap — run them per stop, and keep only the miss and boundary gates deferred to the end where they belong.

services/vault/e2e/fixtures/pageWallets.ts
```diff
@@ -0,0 +1,265 @@
...
+        requestAccounts: async () => [walletConfig.btcAddress],
+        getAccounts: async () => [walletConfig.btcAddress],
+        getPublicKey: async () => walletConfig.btcPublicKeyHex,
+        getVersion: async () => unisatVersion,
+        getChain: async () => signetChain,
+        switchChain: async () => signetChain,
+        // `{ list, total }` is the paging shape the adapter walks; an empty
+        // first page ends the walk immediately.
+        getInscriptions: async () => ({ list: [], total: 0 }),
+        signPsbt: async (psbtHex: string) => psbtHex,
```
gbarkhatov: This does not fail — it returns a well-formed **unsigned** PSBT, so a caller gets what looks like a successful signature and carries on. Same for `signPsbts`; `signMessage` returns a plausible 64-byte value. The header at lines 28-33 says a test walking into a signing step "fails there rather than appearing to succeed", which is the behaviour worth having, particularly for the seam #1592 extends.

Nothing on the captured path calls these — connect goes through `requestAccounts` / `getPublicKey` / `getVersion` — so making them throw is free today:

```ts
signPsbt: async () => {
  throw new Error("capture wallets never sign - extend the recording instead");
},
```

services/vault/e2e/fixtures/pageWallets.ts
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * The ETH side needs no interaction - wagmi has already authorised the
+ * announced provider by the time the dialog opens, so the dialog shows the
+ * account and only Bitcoin is left to choose.
+ */
+export async function connectInjectedWallets(page: Page): Promise<void> {
+  // By testid, not by its label: the header control is a labelled button on
+  // desktop and an icon-only button on mobile, so a text match connects at one
+  // width and times out at the other.
+  await page.getByTestId("connect-wallet-button").first().click();
```
gbarkhatov, line 276: These three controls already carry stable testids — `select-bitcoin-wallet-button` (`ChainButton/index.tsx:18`), `wallet-option-unisat` (`WalletButton/index.tsx:51`) and `chains-connect-button` (`Chains/index.tsx:64`) — and `e2e/real/actions/walletConnect.ts:59,79` drives two of them by testid.

Matching on "Select Bitcoin Wallet" / "Unisat" / "Connect" instead means a label change breaks this capture with a bare timeout while the real-wallet CLI keeps passing — the opposite of the guarantee `depositFlow.visual.spec.ts:84-87` claims when it deliberately reuses `deposit-button`. The `connect-wallet-button` line here already does the right thing; the three below could follow it.

Related: `/do not split/i` at `depositFlow.visual.spec.ts:110` is `copy.ts:603`, so a copy edit silently breaks the spec.

.github/workflows/visual-regression.yml
```diff
@@ -194,9 +209,9 @@ jobs:
         run: |
           set -euo pipefail
           H="${{ runner.temp }}/harness"
-          rm -rf services/vault/e2e/visual packages/babylon-core-ui/visual
-          mkdir -p services/vault/e2e packages/babylon-core-ui .github/scripts
-          cp -R "$H/vault-visual" services/vault/e2e/visual
+          rm -rf services/vault/e2e packages/babylon-core-ui/visual
```
gbarkhatov: **The Vite dep-optimizer cache survives this swap, so the baseline renders HEAD's `core-ui` and `wallet-connector`.**

`services/vault/vite.config.ts:81-88` puts `@babylonlabs-io/core-ui` and `@babylonlabs-io/wallet-connector` in `optimizeDeps.include`, so both are pre-bundled into `services/vault/node_modules/.vite/deps`. Vite's dep hash covers the lockfile and the vite config — never the pre-bundled package's *content* — and both are byte-identical across the two captures. This step deletes `services/vault/e2e` but not `node_modules/.vite`, and the job deliberately does not reinstall, so the candidate run's prebundle is reused by the baseline run. It is the documented "restart with `--force` after changing a linked dep" footgun.

`visual-capture.sh` does rebuild core-ui / ts-sdk / wallet-connector from the checked-out source on both sides, so `dist/` genuinely changes — and is then ignored for those two. A PR that only changes core-ui padding gets HEAD's core-ui served to the merge-base app and all 20 vault screens report clean.

That falsifies this job's own note at lines 227-232 ("Workspace changes (core-ui, ts-sdk, wallet-connector) are still compared correctly because those are rebuilt from the checked-out source"). `ts-sdk` is not in `include`, so it does compare correctly; the two that are named in `include` do not.

One line fixes it:

```bash
rm -rf services/vault/e2e packages/babylon-core-ui/visual services/vault/node_modules/.vite
```

.github/workflows/visual-regression.yml
```diff
@@ -159,7 +159,15 @@ jobs:
           set -euo pipefail
           H="${{ runner.temp }}/harness"
           mkdir -p "$H"
-          cp -R services/vault/e2e/visual "$H/vault-visual"
+          # The WHOLE e2e tree, not just e2e/visual. The capture now depends on
+          # `e2e/fixtures` (the recorded-backend replay and the injected
+          # wallets) and on `e2e/artifacts` (the recording itself), and those
```
gbarkhatov: No longer true after 28fe993 — the recording is committed at `e2e/fixtures/replay/recorded-run.jsonl`, and `e2e/artifacts` is gitignored and absent in CI. Copying the whole `e2e` tree is still right, just for one reason instead of two.

services/vault/src/applications/aave/context/AaveConfigContext.tsx
```diff
@@ -59,7 +59,15 @@ export function AaveConfigProvider({
...
     if (errorFallback !== undefined) return <>{errorFallback}</>;
     return (
-      <div className="flex min-h-[400px] flex-col items-center justify-center gap-3 px-4 text-center">
+      // `app-error-state` is read by the visual capture
+      // (services/vault/e2e/visual/capture.ts), which refuses to photograph
+      // an error surface as a baseline. This fallback covers every page that
+      // needs Aave config, so it is what a capture with no backend behind it
+      // renders - and it is stable, so it diffs clean against itself forever.
+      <div
+        data-testid="app-error-state"
```
gbarkhatov, line 68: **This gate is bypassed at the one call site that wraps the deposit dialog.**

Line 60 above returns `errorFallback` *before* reaching this div, and exactly one caller in the app passes it: `RootLayout.tsx:288-303`, the provider wrapping `SimpleDeposit`. That fallback is a `FullScreenDialog` reading "Something went wrong" carrying neither `app-error-state` nor `error-dialog`. `router.tsx:114,148` wrap the route pages without `errorFallback`, so the 12 route shots are covered and the 8 deposit-flow shots are not.

Combined with the missing screen-count assertion, an Aave-config failure times out the dialog wait in `depositFlow.visual.spec.ts`, withholds all 8 shots symmetrically, and the run still reports "No visual changes" — the gate that was meant to name this failure cannot see it. Putting the same testid on the `RootLayout` fallback closes it.

gbarkhatov (dismissed):
Verified this against the fixture and the CI run rather than by reading: the mechanism works. Run 31918074054 captured 20/20 vault screens on **both** sides, 14/14 tests passed on each, `MISSING` empty, `COMPARED=2` — so two independent runs of the deposit flow produced byte-identical PNGs, which is real evidence the determinism work holds. The 10 replay unit tests pass locally. Approving.

Three things I'd genuinely like fixed before this becomes the instrument everyone trusts — the selector fallback (`chain.ts`), the Vite dep cache (`visual-regression.yml`), and the missing screen-count assertion (`capture.ts`). The rest are inline and mostly comment accuracy.

### One thing about the description

It says *"every one of the twelve existing pictures will show as changed here"*, and the bot comment says **"✅ No visual changes"**. Both are correct, but only if you already know why: the harness is stashed from HEAD and restored onto the merge-base, so the baseline replays the same recording through the same wallets and renders the same populated app. The instrument change cancels out on both sides — that is the property the stashing design exists to provide, and zero changed screens is the right answer.

For this PR of all PRs, though, a reviewer told to expect 20 changed images and shown a green check will conclude the tool is still broken. The evidence it works is the counts (12 → 20 screens on both sides) and the gates passing, not a diff. Worth rewriting that paragraph.

### Items with no line in the diff to hang them on

- **`targets.ts:8-11`** still says "Every target here must render without a connected wallet […] page-side provider construction is #1592, so connected-state screens are not reachable yet and are deliberately absent" — 15 lines above the `DEPOSIT_FLOW_STOPS` this PR adds. Scope that paragraph to `VISUAL_TARGETS` and drop the "not reachable yet" clause.
- **`chain.ts:265` `answerMethod`** keys on the method name alone, so the recorded `eth_getBalance` answers 0.05 ETH for any address and any block tag. Harmless while the injected wallet is the recorded depositor; same class as the selector fallback, without even an exact tier.
- **Unused API** (`ZERO DEAD CODE`): `RecordedChain.size` / `get size()` (`chain.ts:104`, `:267`) is never read anywhere; `Supplement.because` (`supplements.ts:47`, `:92`) is populated and never read — a plain comment above the entry keeps the intent without an unread field.
- **`build-replay-fixture.mjs:76`** — `--step` with no value leaves `rest[stepFlag + 1]` undefined, and the error then reads `no exchange recorded during step "undefined"`.
- **`SECRET_PATTERNS`** only runs in the generator; nothing re-checks the committed fixture. A vitest case asserting `recorded-run.jsonl` is free of those patterns keeps the property enforced when someone regenerates by hand or extends the cut past `deposit-form` — which that comment says is the case it exists for.
- **Now-dead pre-existing code**, flagged only because this PR is what settles it: `walletInjection.ts` (`injectWallets` / `getInjectedWallets` / `E2E_WALLETS_GLOBAL`) and `NEXT_PUBLIC_E2E_MODE` have no consumer in `src/` — as this PR's own header correctly states — and their only remaining callers are the barrel re-export in `fixtures/index.ts:33-39` and their own unit test. Not this PR's job to delete, but it will not gain a consumer now.

### What's right and I'd keep exactly as-is

Staging the screenshot in memory and flushing only after the gates pass — that is the load-bearing insight, and documenting it at both ends (here and in the workflow) is the right call. Decomposing `aggregate3` instead of byte-matching batches, with the fallback's failure mode stated rather than hidden. `answerMulticall` returning `null` and recording the miss instead of throwing past the route handler, with a test for it. Cutting the run at a step so a screenshot shows one moment rather than a mix. Deriving the supplement from the run so a re-recording updates it and the app's own mismatch guard stays meaningful. Using the extension seams instead of adding a test-only path in `src/`, with a specific reason for rejecting `window.btcwallet`. And the two `data-testid` additions — two lines, no behaviour change, and they are what makes the second gate possible at all.

services/vault/e2e/fixtures/replay/chain.ts
jonybur: Confirmed and fixed in 6fcca5285. Decoding the committed fixture reproduces your numbers exactly: `getReservesPrices` on `0xcdcb913a…` is recorded three times with one-element arrays (ids 0 and 1 -> 100000000, id 2 -> 6469260700000), and reserve 3 — `aaveConfig.vaultBtcReserveId`, asserted as 3 by the co-located supplement test — is never priced. Same shape on `0xb54f1d4a…|0x9f9b1990` (10000000000000 vs 2500000000000000).

The selector table now refuses rather than guesses: a `(target, selector)` pair recorded under a second distinct calldata is marked ambiguous and stops answering by selector, so an unrecorded argument comes back `null` -> `unanswered` -> the gate fires. Four pairs in the recording are affected. The single-argument fallback you describe as the reason it exists is untouched.

Two tests pin it: `getReservesPrices([2])` still answers, `([3])` is now refused, and `pauseState()` still falls back.

services/vault/e2e/fixtures/replay/chain.ts
jonybur: Fixed in 6fcca5285. Decoding `recorded-run.jsonl` as committed gives 19 `eth_call`s (all 19 `aggregate3`), 39 inner calls, 28 distinct `(target, selector)` pairs — your figures. The header, the `tableCache` comment and the `~226 batches` note all now cite those.

services/vault/e2e/fixtures/replay/index.ts
jonybur: Both confirmed and fixed in 6fcca5285. `GET /health` is the only bodyless GraphQL-classified entry, so it did own the empty key, and `recorded.resBody ?? "{}"` answered anything else bodyless with a 200 carrying neither `data` nor `errors` and no miss.

One correction to the proposed fix, worth recording: `method + pathname + operationName` breaks the capture. The recorded indexer serves GraphQL at `/`, while `MOCK_ENV_VARS` pins `/graphql`, so keying a POST on its path matches nothing and every screen renders empty. The key is now method+path only when there is no operation (the bodyless case, where recorded and live paths do agree), and operation+`variables` when there is — which also closes your second point: `fetchVaultProviderStats` issues one `GetVaultsByProvider` per provider, and every call site lowercases its variables, so adding them introduces no spurious misses on the current fixture.

services/vault/e2e/fixtures/replay/index.ts
jonybur: Right on both counts — `served` is handed out by reference, there is no getter. The first comment is gone in 6fcca5285.

services/vault/e2e/fixtures/replay/recording.ts
jonybur: Confirmed and fixed in 6fcca5285. Running both policies side by side, the builder's drop set was a strict superset: `sentry.*` self-hosted, `web3modal.com` and `walletconnect.com` were dropped by the builder and replayed as `eth-rpc` by the loader — and `verify.walletconnect.com` is named in the app's own CSP in `index.html`, so it is not hypothetical.

There is now one exported `isDroppedHost` in `recording.ts`, imported by `build-replay-fixture.mjs`. Plain `node` resolves the `.ts` import because `recording.ts` is a leaf with no relative imports and uses only erasable syntax — both now stated in its docstring, since either would break the build script silently. The `.com` aliases are added, and the dot-anchoring is kept so `sentry.io.example.com` is still not dropped. Covered by a regression test.

The own-origin docstring claim is gone too; it now says plainly that the `demo` label is what covers the devnet host and that a localhost recording classifies its own requests as `eth-rpc`.

services/vault/e2e/visual/capture.ts
jonybur: Confirmed and fixed in 6fcca5285. I reproduced it: 12 route PNGs on both sides with the 8 deposit shots withheld gives `changedCount: 0`, `MISSING=""`, and a green "No visual changes".

Rather than read the count from `targets.ts` at diff time, each capture now writes an `expected-screens.txt` manifest at collection time, and `visual-diff.mjs` takes `--expected-baseline`/`--expected-candidate` and reports `absentScreens`. Two reasons for doing it per side rather than from the manifest: Storybook's expected set is derived from `storybook-static/index.json`, which both sides overwrite at the same path, so it is unrecoverable by diff time; and it makes the declaration a statement of intent made before anything can fail, so a spec that never ran still has its screens accounted for.

A shortfall goes into `MISSING`, which is already wired to the embed gate, the job summary, the PR comment and the job outputs. A missing manifest counts as a shortfall too, so it fails closed. The emptiness check is now PNG-based so a manifest-only directory cannot pass, and `visual-capture.sh` counts PNGs. Two unit tests cover it.

services/vault/e2e/visual/depositFlow.visual.spec.ts
jonybur: Good question, and the answer is worse than the doubt. The guard does run and does fail — the capture only survives because a second effect overwrites the error screen a tick later.

Reproduced with a MutationObserver over the real connect flow: the markers rendered are `Connecting Unisat`, `Public Key Mismatch`, `do not match`, `Continue Anyway`. `displayError` only swaps the dialog screen; `selectWallet` and `accountStorage.set` have already run, so with ETH already connected the auto-confirm effect fires `confirm()` + `displayChains()` and replaces the error with the chains screen in the same tick. With the ETH provider removed, the mismatch screen stays up — that is the guard's verdict with the race removed.

Fixed in 6fcca5285 by your second suggestion. The internal key is not in the recording (I scanned all 439 hex candidates; none tweaks to the recorded address), so the address is now derived from the key the fixture holds rather than the reverse. The recording still funds it because `mempoolKey` normalises the address out of every mempool path, and the recorded `validate-address` response is reusable since the app only checks the scriptPubKey is hex with a known prefix. A test re-derives the pair, so it cannot drift again.

services/vault/e2e/visual/depositFlow.visual.spec.ts
jonybur: You are right that the justification was wrong — the option row is in the DOM on every render; what excluded it was Playwright's accessibility-hidden filter over the inherited `visibility: hidden`, not non-rendering.

Fixed in 6fcca5285 with your suggestion: the accordion header carries `data-testid="split-selector-toggle"` and the spec clicks that, so `.first()` and the regex are gone. That also closes the sharper failure you note — the header label IS the current choice, so a pre-selected split would have left `/do not split/i` matching only the hidden row and timing out.

services/vault/e2e/visual/depositFlow.visual.spec.ts
jonybur: Correct, and fixed in 6fcca5285. The two DOM gates are now split into `assertNoErrorSurface` and run inside `capture()` per photograph, so a stop settled on a static error frame fails at that stop. The accumulating checks — misses and boundaries — stay deferred to the end as `assertRecordingCovered`, which no longer takes a `page` since it never needed one.

services/vault/e2e/fixtures/pageWallets.ts
jonybur: Fixed in 6fcca5285 — `signPsbt`, `signPsbts` and `signMessage` all throw now. As you say it costs nothing today, since connect goes through `requestAccounts`/`getPublicKey`/`getVersion`, and it makes the header's claim true rather than aspirational. The message is a named constant passed through `addInitScript`'s argument array, since closures do not survive the structured clone.

.github/workflows/visual-regression.yml
jonybur: Confirmed and fixed in 6fcca5285. `optimizeDeps.include` does hold `@babylonlabs-io/core-ui` and `@babylonlabs-io/wallet-connector` and not `ts-sdk`, and the capture is served by Vite, so the mechanism holds as described. The restore step now also removes `services/vault/node_modules/.vite`, which makes the note on the baseline step true rather than aspirational.

.github/workflows/visual-regression.yml
jonybur: Fixed in 6fcca5285 — the comment now gives the one remaining reason and no longer mentions `e2e/artifacts`.

services/vault/e2e/visual/depositFlow.visual.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+      // and nothing in a capture signs, so the output key stands in for it.
+      btcPublicKeyHex: `02${RECORDED_DEPOSITOR.BTC_X_ONLY_PUBLIC_KEY}`,
+      ethAddress: RECORDED_DEPOSITOR.ETH_ADDRESS,
+      ethChainIdHex: SEPOLIA_CHAIN_ID_HEX,
+      ethRpcUrl: MOCK_ENV_VARS.NEXT_PUBLIC_ETH_RPC_URL,
+    });
+
+    const shots: StagedShot[] = [];
+
+    await page.goto("/vaults", { waitUntil: "domcontentloaded" });
```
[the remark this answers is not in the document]
jonybur: Partly addressed in 6fcca5285, and the rest is deliberate. The `/do not split/i` matcher and the `"button"` role are gone entirely — that control is now clicked by `data-testid="split-selector-toggle"`, which is the more useful fix, since matching on the header's label was broken in a second way (the label IS the current choice, so a pre-selected split matched nothing).

`"/vaults"` and `"domcontentloaded"` are left inline. They are a route path and a Playwright load-state used once each, exactly as `routes.visual.spec.ts` uses them, and naming them would add indirection without making either safer to change.

services/vault/e2e/visual/routes.visual.spec.ts
jonybur: Fixed in 6fcca5285. `VisualTarget` now carries `requires: readonly RecordedBackend[]`, passed through at the call site.

The values are measured, not assumed: I instrumented a capture run to print `backend.served` per screen. All four boundaries are non-zero on all six routes at both viewports, and `vp-rpc` is zero everywhere — so requiring it would have failed every screen. Written per target rather than as one shared list, for the reason you give: a route that stops needing one should say so at its own entry.

services/vault/e2e/fixtures/pageWallets.ts
jonybur: Fixed in 6fcca5285. All three are driven by their existing testids now, and the trailing "Connect Wallets" wait is replaced by waiting for `chains-connect-button` to leave the DOM, so no step in the connect flow matches on copy any more.

The related `/do not split/i` point is fixed on its own thread — that control gained a `split-selector-toggle` testid.

services/vault/src/applications/aave/context/AaveConfigContext.tsx
jonybur: Confirmed and fixed in 6fcca5285. `RootLayout.tsx` is the only caller passing `errorFallback`, and it wraps `SimpleDeposit` — so the gate covered the 12 route shots and none of the 8 deposit-flow ones. The fallback dialog now carries `app-error-state`, with a comment pointing at the capture so it does not get dropped again.

Together with the screen-count manifest on the other thread, the failure you describe — Aave config fails, the dialog wait times out, all 8 shots are withheld symmetrically — is now caught twice over.

jrwbabylonlab (approved):

gbarkhatov (approved):
All fifteen review comments addressed in 6fcca5285, verified against the code rather than the replies: the multi-arg selector fallback now refuses instead of inventing a price, the GraphQL key no longer collapses bodyless requests onto `GET /health`, the expected-screens manifest closes the symmetric-withhold silent green, route targets declare their boundaries, the Vite prebundle is cleared with the harness swap, the injected wallet derives its address from the key it holds and throws on every signing call, and the deposit dialog's Aave fallback carries `app-error-state`.

The two mechanism corrections are fair — the accordion row is CSS-hidden rather than unmounted, and the pubkey guard did fire and was raced by auto-confirm — and both fixes hold regardless. The declined items are reasonable as argued.

Approving.

jonybur:
@greptile-apps

Issue #2230, opened 2026-08-12
[eth-first] Allow apps to require one wallet and offer the other


jonybur:
## Outcome

An app can require Ethereum while offering Bitcoin. Wallets can connect, confirm, restore, and disconnect safely.

The base API in #2267 is merged. Identity fixes #2350 and #2351, AppKit setup #2353, package checks #2360, and independent disconnects #2352 are complete. #2445 delivered the package change. #2354 and #2355 remain.

## Problem

The base API does not complete consent adoption, independent sessions, or both-product acceptance checks.

## Scope

| Remaining work | Status |
|---|---|
| #2352: disconnect either wallet independently | Complete. #2439 merged into main at `751069c6` on 2026-09-11. #2352 is closed. |
| #2354: apply the selected consent rule | #2491 merged into main at `36d72d60` on 2026-09-15. #2354 stays open for its decision record. |
| #2355: verify the agreed wallet session flows | #2483 merged in `05f60bf9` with passing CI. Real Vault checks pass. Withdraw, multi-hub, and the staking journey remain. |

Use small PRs with named completion conditions and passing checks. Apply the required-work rule in #2228. Do not add performance targets, a separate ticket per PR, or duplicate acceptance lists.

## Complete when

- [x] #2352 and #2354 deliver the approved disconnect and consent behavior with required review and CI.
- [ ] #2355 proves the agreed session checks in both consumer apps. Link its evidence.
- [x] Ethereum-only and Bitcoin consumers pass clean install checks. #2360.

## Dependencies

Epic: #2228. #2352 owns both approved disconnect directions. #2354 owns consent adoption. #2355 owns the finite connector session evidence list.

Vault action prompts in #2482 merged into main at `45d50cd9`. This work remains under #2232.

The final changed-Vault journey belongs to #2233 after #2232. It does not block this wallet milestone.

Issue #2232, opened 2026-08-12
[eth-first] Let Ethereum-only users enter and use the Vault


jonybur:
## Outcome

Ethereum alone can use the Vault behind a control that starts disabled. Bitcoin actions request Bitcoin first.

Status: every step is merged, and the control starts disabled. Main serves Ethereum-only sessions when the control is on. Repay, withdraw, reload, and restore still need control-on tests, and the reorder test does not collect on main.

| Step | Current state |
|---|---|
| Wallet audit | #2447 merged in #2465. |
| Bitcoin prompt | #2482 merged into main at `45d50cd9`, with the delayed-readiness payout fix #2505. |
| Session handling | #2439 merged at `751069c6`, #2499 at `3f92d2b0`, and consent #2491 at `36d72d60`. #2355 owns the remaining consumer checks. |
| Payout addresses | #2500 merged at `dd1b2e48`, covering five address forms on mainnet and signet, including invalid inputs. |
| Optional Bitcoin | #2501 merged at `e4bff664` and keeps the control disabled. Eligible reclaim offers Connect Bitcoin wallet before ownership checks. |

The deferral recorded in #2228 applies to #2506 and #2507. These two follow-ups do not block this issue. They are accepted exceptions to the related completion conditions below.

Decision approved by jonybur: download-size improvements are optional. Keep the current address converter and validation. No new encoder, Bitcoin library removal, curve-code deferral, or no-download proof is required to complete this ticket. Existing lazy loading and package contracts remain in place. The earlier loading-scope question no longer blocks implementation.

## Problem

Current connection checks block Ethereum actions without Bitcoin. Main also clears the whole session after a sustained Bitcoin disconnect.

## Scope

These steps share this ticket. Each PR names its behavior and verification. #2355 owns the agreed connector session checks. #2233 owns the final browser journey and release. Reuse their evidence.

1. Complete: the audit records the wallet requirement and current behavior for each action. See #2447 and #2465.
2. Add the Bitcoin prompt. Cancel leaves the pending action unstarted. Connection never runs it; require a separate explicit retry.
3. Apply the independent-session rule from #2352 and consent rule from #2354. Preserve valid restore and pending-action state.
4. Keep registered payout addresses correct with Ethereum alone. Preserve supported forms, networks, and invalid-input handling. Use the existing converter and checks in #2500.
5. Use the audit to make Bitcoin optional behind the disabled control. Require Bitcoin only for actions that need it.

[...]

PR #2267, opened 2026-08-17
feat(wallet-connector): let an app offer a wallet without demanding it

jonybur:
## Result

Merged as the base for #2230. #2230 stays open because the post-merge audit found blockers.

## Change

- Apps can show a wallet without requiring it.
- The dialog records the approved wallet identities.
- The public API supports per-chain connect and disconnect.
- The package has an Ethereum-only entry point.
- Closing an unfinished dialog keeps a successful connection.

## Post-merge audit

- #2350: confirmation can approve a changed identity.
- #2351: live identity changes can keep an old approval.
- #2352: a Bitcoin disconnect can also disconnect Ethereum.
- #2353: AppKit entry points can reuse incompatible state.
- #2354: consent timing and old-session migration need approval.
- #2355: the consumer and session test matrix is missing.
- #2360: the optional dependency check is missing.

## Verification at merge

Automated checks passed. They did not cover the real AppKit boundary or the full consumer matrix.


jonybur (commented):
One correctness finding on the per-chain disconnect wiring - the headline capability of this PR.

**BTC per-chain disconnect tears down ETH when both use the shared AppKit modal.**

- `packages/babylon-wallet-connector/src/core/wallets/btc/appkit/provider.ts:218` calls `modal.disconnect()` with no argument.
- Reown AppKit's `modal.disconnect(chainNamespace?)` disconnects **all** namespaces when called without one.
- In the unified (`WalletProvider` root) setup in `packages/babylon-wallet-connector/src/core/wallets/appkit/appKitModal.ts`, ETH's WagmiAdapter and BTC's BitcoinAdapter register on the **same** `AppKit` instance. `hasSharedBtcAppKitConfig()` is true whenever BTC and ETH both live on that modal.
- So the new `useWalletConnect().disconnect("BTC")` (`hooks/useWalletConnect.ts:51`) - which the WalletMenu wires to the per-wallet Disconnect button - ends up calling BTC provider `disconnect()` -> `modal.disconnect()` -> Reown disconnects both bip122 AND eip155. The ETH `disconnect` event then fires `removeWallet("ETH")`. In an ETH-required config this drops `confirmed`, deletes the `WALLET_CONFIRMATION_RECEIPT_KEY`, and boots the user through the connect dialog.
- The WalletMenu.test.tsx cases pass because they assert only the outer callback shape (`onDisconnect("BTC")`), not the underlying provider effect.

Fix direction: pass the namespace at the provider seam - `await modal.disconnect("bip122")` in the BTC provider (and mirror at ETH provider if/when it goes through modal for consistency). The greptile-flagged 'shared AppKit modal singleton with two independent initializers' concern on `modal.ts:86` is a separate issue with the same shape - both point at the shared-modal boundary being under-tested.

The rest of the change (session-persistence on close, receipt-only-tracks-required-chains, the case-insensitive `sameIdentity` on `confirmationReceipt.ts:49` - fine for bech32/EVM in practice, worth noting for base58 P2PKH edge cases only), and the `__fixtures__/ethOnlyConsumer.tsx` dts-exclusion all check out.

packages/babylon-wallet-connector/src/context/State.context.tsx
```diff
@@ -71,18 +81,47 @@ function filterWalletsByValidChains(
...
+    chains: chains.reduce((acc, chain) => ({ ...acc, [chain.id]: chain }), {}),
+    requiredChainIds: [...requiredChainIds],
+  }));
+  const stateRef = useRef(state);
+  stateRef.current = state;
+  const previousRequiredChainIdsRef = useRef([...requiredChainIds]);
 
   useEffect(() => {
+    if (hasNewRequiredChain(previousRequiredChainIdsRef.current, requiredChainIds)) {
+      storage?.delete(WALLET_CONFIRMATION_RECEIPT_KEY);
```
jrwbabylonlab: Compares against the *previous route* set. simple-staking derives requiredChains per route (`/baby` → `["BBN"]`), so navigating /baby → main makes BTC look new: receipt deleted here, `confirmed` dropped at L118.

Routine navigation logs the user out mid-session.

packages/babylon-wallet-connector/src/core/confirmationReceipt.ts
```diff
@@ -0,0 +1,154 @@
...
+  const receipt = parsed as Partial<ConfirmationReceipt>;
+  const requiredChains = normalizedRequiredChains(requiredChainIds);
+
+  if (
+    receipt.version !== 1 ||
+    !Array.isArray(receipt.requiredChains) ||
+    !Array.isArray(receipt.entries) ||
+    !receipt.requiredChains.every((chain) => typeof chain === "string") ||
+    !receipt.entries.every(isReceiptEntry) ||
+    receipt.requiredChains.length !== requiredChains.length ||
```
jrwbabylonlab: Exact set equality rejects a superset receipt. A {BTC,BBN} receipt fails on `/baby` (2 ≠ 1), and the `["BBN"]` receipt minted there then fails on main.

Reload signs the user out in both directions.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -34,41 +50,75 @@ export function WalletDialog({
...
   const walletWidgets = useWalletWidgets(connectors, config, onError);
   const { connect } = useWalletConnectors({ persistent, accountStorage: storage, onError });
-  const { disconnect: disconnectAll } = useWalletConnect();
 
-  const disconnectTimerRef = useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
+  // Closing without confirming leaves the connectors connected, so the user can
+  // reopen and finish where they left off. The receipt written on confirm is
+  // the single gate that lets a later reload auto-confirm — no receipt, no
+  // silent restore — so leaving the connection up grants nothing on its own.
+  const handleClose = useCallback(() => {
```
jrwbabylonlab, line 77: "Leaving the connection up grants nothing" does not hold for vault — it gates on provider connection, not `confirmed` (`Connect.tsx:58`).

Connect both wallets, close with X, and the full deposit/borrow UI is usable with no confirm and no ToS acceptance, surviving reload.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -34,41 +50,75 @@ export function WalletDialog({
...
+            connections,
+          });
+        }
+        await onConfirm?.(connections);
 
-  useEffect(() => clearDisconnectTimer, [clearDisconnectTimer]);
+        if (persistent) {
+          const receipt = createConfirmationReceipt(requiredChainIds, connections, connectors);
+          if (receipt) {
+            storage.set(WALLET_CONFIRMATION_RECEIPT_KEY, receipt);
```
jrwbabylonlab: The receipt inherits the 1h per-entry TTL and is never re-set, while per-chain entries slide on every reconnect.

Any reload more than an hour after the single confirm is a hard signout, regardless of continuous activity.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
gbarkhatov, line 77: Closing used to schedule `disconnectAll()` after 1s; now it leaves both connectors up. simple-staking absorbs that by gating its menu on `connected` (`selected && confirmed`) in `ui/common/components/Wallet/Connect.tsx` — the vault did not get the same gate.

`services/vault/src/components/Wallet/Connect.tsx:58` derives `isWalletConnected` from `useBTCWallet().connected && useETHWallet().connected`, and `useWalletConnect().connected` is not read anywhere under `services/vault/src`. So in the vault a user can pick a BTC wallet and an ETH wallet, dismiss the dialog with X, and land in the fully-connected UI without ever pressing **Connect** — the control the "By clicking Connect you agree with the Terms of Use and Privacy Policy" text hangs off (`components/Chains/index.tsx:69-101`).

Either give the vault the same `confirmed` gate simple-staking got here, or state why the vault is intentionally connector-gated.

packages/babylon-wallet-connector/src/context/State.context.tsx
```diff
@@ -71,18 +81,47 @@ function filterWalletsByValidChains(
...
+      const validChainIds = new Set(chains.map((chain) => chain.id));
       const filteredWallets = filterWalletsByValidChains(state.selectedWallets, validChainIds);
 
-      return { ...state, chains: newChains, selectedWallets: filteredWallets };
+      return {
+        ...state,
+        // A confirmation only ever covered the requirement set the user was
+        // shown. Asking for a chain that was not in that set must send them
+        // back through an explicit confirmation.
+        confirmed: hasNewRequiredChain(state.requiredChainIds, requiredChainIds) ? false : state.confirmed,
```
gbarkhatov: `hasNewRequiredChain` is evaluated against the immediately preceding requirement set, so a narrow-then-widen round trip reads as a brand-new requirement.

simple-staking derives `requiredChains` from the route (`ui/common/context/wallet/WalletConnectionProvider.tsx:56-59`): `["BTC","BBN"]` → `/baby` → `["BBN"]` → back → `["BTC","BBN"]`. The narrowing leg is correctly a no-op, but the return leg clears `confirmed` here and deletes the receipt on line 104 — for a set the user already approved.

Recovery is not automatic. `coldStartRestoreEligibleRef` (`hooks/useWalletConnectors.tsx:78`) was latched to `false` the first time `confirmed` went true, and `initialRequiredChainIdsRef` (line 77) is captured at mount and never re-baselined, so the cold-start effect can't re-confirm; a reload can't either, because the receipt is gone. Net effect: visiting `/baby` and coming back signs the user out with both wallets still connected. Pre-PR the auto-confirm effect had no one-shot guard and re-ran on every connectors bump, so this transition self-healed.

The receipt already persists its own `requiredChains` — comparing against that (the largest set actually approved) rather than the previous prop value would make the round trip a no-op while still catching a genuinely new requirement.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
gbarkhatov: The receipt is written once, inside `if (!confirmed)`, and never rewritten afterwards.

It goes through `createAccountStorage`, which stamps a timestamp per key and expires the entry at `ttl` — defaulted to `ONE_HOUR` in `WalletProvider/index.tsx:69`, and neither the vault nor simple-staking overrides it. Per-chain entries *are* re-stamped on every connect including auto-reconnect (`hooks/useWalletConnectors.tsx:114`, `:182`, `:193`), which is why a persisted session used to survive reloads indefinitely. Now `hasConfirmationReceipt` goes false roughly an hour after the last explicit confirm, and a reload past that point reconnects both wallets but silently drops back to the connect screen.

If a one-hour consent lifetime is intended, worth stating it in the PR description; otherwise re-stamp the receipt alongside the chain entries.

packages/babylon-wallet-connector/src/core/confirmationReceipt.ts
```diff
@@ -0,0 +1,154 @@
...
+  requiredChainIds: readonly string[],
+  connections: readonly ConfirmationConnection[],
+  connectors: Connectors,
+): string | undefined {
+  const requiredChains = normalizedRequiredChains(requiredChainIds);
+  const entries: ConfirmationReceiptEntry[] = [];
+
+  for (const chain of requiredChains) {
+    const connection = connections.find((candidate) => candidate.chain === chain);
+    if (!connection?.account || !connectorFor(connectors, chain)) return undefined;
```
gbarkhatov: When this returns `undefined` the caller skips the write (`WalletProvider/components/WalletDialog.tsx:88`) but still falls through to `confirm?.()` / `close?.()`. The session is confirmed with nothing persisted, and the next reload silently needs a fresh confirmation with no signal that anything went wrong.

It is reachable: the Connect button is gated on `selected`, which is `Boolean(selectedWallets[chainId])` (`hooks/useWalletConnect.ts:67`) and does not require an `account`, while `connections` in `handleConfirm` keeps only entries with `wallet?.account`. A required chain whose wallet is selected but account-less passes the gate and drops out of `connections`, landing exactly here.

CLAUDE.md's no-silent-fallbacks rule would have this surface through `onError` rather than pass quietly.

packages/babylon-wallet-connector/src/__tests__/eth.entry.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    const manifestPath = path.join(PACKAGE_ROOT, "../..", relative, "package.json");
+    const manifest = JSON.parse(readFileSync(manifestPath, "utf8")) as {
+      dependencies?: Record<string, string>;
+      devDependencies?: Record<string, string>;
+    };
+
+    return { relative, declared: { ...manifest.dependencies, ...manifest.devDependencies } };
+  });
+
+  it.each(optionalPeers)("%s is either externalised and declared by every consumer, or bundled", (peer) => {
```
gbarkhatov: This registers zero tests. `packages/babylon-wallet-connector/package.json` has no `peerDependenciesMeta`, so `optionalPeers` is `[]`, `it.each([])` registers nothing, and `declares every optional peer in peerDependencies` below then asserts `[]` equals `[]`.

```
$ npx vitest run src/__tests__/eth.entry.test.ts --reporter=verbose
✓ the ./eth entry point > ... (7 tests)
✓ optional peer dependencies > declares every optional peer in peerDependencies
Tests  8 passed (8)
```

The docstring presents this as the rule that keeps `./eth` installable, but nothing is enforced today. Either add the `peerDependenciesMeta` entries it is meant to police, or drop the block — CLAUDE.md's zero-dead-code / no-test-bloat rules.

packages/babylon-wallet-connector/src/hooks/useWalletConnect.ts
```diff
@@ -1,34 +1,72 @@
...
+   * Opens the wallet dialog, on one chain's wallet list when given a chain.
+   *
+   * Deliberately does not `reset()`: attaching an optional chain to a session
+   * that is already confirmed must not tear that session down. Requirement-set
+   * changes are handled separately — `State.context` drops the confirmation
+   * when a new required chain appears.
+   */
+  const open = useCallback(
+    (chain?: ChainId) => {
+      if (chain && chainMap[chain]) {
```
gbarkhatov: Nit: `disconnect` below guards the `onClick={disconnect}` case explicitly with `hasOwnProperty`, but `open` relies on `chainMap[chain]` simply missing for a `MouseEvent` key — it works, by accident.

The four simple-staking call sites were wrapped in this PR; `services/vault/src/components/Wallet/Connect.tsx:159` still passes `open` bare. Applying the same membership check here would make it correct regardless of how a consumer wires it, and remove the need for the wrappers.

gbarkhatov (dismissed):
Approving — the `./eth` entry-point split is the substantial half of this PR and it holds up. The module-graph guard test, `ChainMetadataMap` injection, the AppKit modal singleton extraction and the multi-entry Vite config (with `cssFileName` pinned so `./style.css` stays stable) all check out, and the two OKX SVGs deduped into `wallets/icons/okx.svg` are byte-identical. Package tests pass (89 wallet-connector, 7 core-ui), eslint is clean on the changed files, and `tsc -b --noEmit tsconfig.lib.json` is clean.

The findings below are all on the second half — the confirmed-session / receipt machinery. Two of them are behavioural regressions I'd want resolved before merge rather than after; the rest are smaller.

**Blocking-ish**

1. Closing the dialog no longer disconnects, and the vault has no `confirmed` gate — a vault user reaches the connected UI without ever pressing the button that carries the Terms of Use agreement.
2. Requirement-set expansion is measured against the previous prop value, so simple-staking's `/baby` round trip signs the user out of a session they already approved, with no automatic recovery.

**Smaller**

3. The confirmation receipt is written once and inherits the 1h storage TTL, capping a restorable session at an hour.
4. A receipt that can't be minted is skipped silently while the session still confirms.
5. The `optional peer dependencies` describe block registers zero tests.
6. Nit: `open` lacks the MouseEvent guard `disconnect` has.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -34,41 +50,75 @@ export function WalletDialog({
...
+          }
+        }
+      }
 
-  const handleClose = useCallback(() => {
-    close?.();
-    if (!confirmed) {
-      clearDisconnectTimer();
-      disconnectTimerRef.current = setTimeout(disconnectAll, ANIMATION_DELAY);
+      confirm?.();
```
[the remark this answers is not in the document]
jonybur: Fixed in 191da76 — this was a real race and it's now closed.

`handleConfirm` re-reads the live selection after awaiting `acceptTermsOfService` / `onConfirm`, via a ref rather than the snapshot taken before the await. If a required chain has disconnected in the meantime it throws instead of recreating the receipt, so the session can't be confirmed without a required wallet attached.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/modal.ts
```diff
@@ -0,0 +1,86 @@
...
+  const modal = createAppKit({
+    adapters: [wagmiAdapter],
+    networks: [chain],
+    projectId: config.projectId,
+    metadata: config.metadata,
+  });
+  setAppKitModal(modal);
+
+  return { modal, wagmiConfig: wagmiAdapter.wagmiConfig };
+}
```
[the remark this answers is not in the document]
jonybur: Fixed in 191da76.

You're right that the first caller wins and the second gets a modal whose adapters it never created. AppKit allows one modal per page and its adapters are fixed at creation, so the unified initializer can't retrofit Bitcoin onto a modal the Ethereum-only entry built.

It now detects exactly that case — an existing modal plus `config.btc` requested plus no `bitcoinAdapter` — and throws a typed `WalletError` naming the cause, rather than returning a modal that silently can never connect a Bitcoin wallet.

packages/babylon-wallet-connector/src/hooks/__tests__/useWalletConnect.optionalChains.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+vi.mock("@/hooks/useWidgetState", () => ({
+  useWidgetState: () => harness.widgetState,
+}));
+vi.mock("@/context/Chain.context", () => ({
+  useChainProviders: () => harness.connectors,
+}));
+
+const btcWallet = { id: "unisat", account: { address: "bc1p", publicKeyHex: "02ab" } } as IWallet;
+const ethWallet = { id: "metamask", account: { address: "0xabc", publicKeyHex: "04cd" } } as IWallet;
```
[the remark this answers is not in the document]
jonybur: Leaving this one as written, and flagging why rather than silently skipping it.

This repo's CLAUDE.md carries two rules that pull opposite ways here. The general TypeScript rule says extract hardcoded strings to named constants; the test philosophy section says the opposite for tests specifically — "prefer explicit values over magic constants imported from elsewhere", and "a new contributor should understand a test by reading it top-to-bottom".

For fixtures like `"unisat"` and `"bc1p…"` the explicit value at the assertion is what makes the test readable, so the test-specific rule wins. Happy to revisit if the maintainers read it the other way.

packages/babylon-wallet-connector/src/context/State.context.tsx
jonybur: Fixed in 191da76.

You and @gbarkhatov both landed on this and you were right that it made routine navigation a sign-out. The comparison against the previous requirement set is gone from `State.context` entirely — a change to the required set is no longer treated as a consent change there.

The receipt now records the chains the user actually approved (everything connected when they pressed Connect) rather than the required set, and is validated by coverage rather than equality. `/baby` → back is a no-op in both directions. Regression tests in `useWalletConnectors.requirementChurn.test.tsx` cover the narrow and widen legs, cold-start restore under each, and that a chain the user never approved still withdraws the confirmation.

packages/babylon-wallet-connector/src/core/confirmationReceipt.ts
jonybur: Fixed in 191da76.

Exact set equality is gone. `isValidConfirmationReceipt` now checks coverage: every currently-required chain must be named by the receipt and still match the live wallet, account and network. Extra chains in the receipt are fine, which is what makes the `/baby` ↔ main round trip work in both directions you describe.

The other half of your point mattered too — a receipt minted on `/baby` covering only `["BBN"]` would still fail back on main. So the receipt is minted over every *connected* chain at confirm time, not the required ones, and it survives both legs. `accepts a receipt covering more chains than are currently required` in `confirmationReceipt.test.ts` pins it.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in 191da76 — you're right, and the claim was wrong as written.

`services/vault/src/components/Wallet/Connect.tsx` now gates on `useWalletConnect().connected` alongside the provider connections, the same gate staking has. Connecting both wallets and closing with X leaves you on the connect button, not in the deposit UI.

I've stopped making the broader claim in the description too: "leaving the connection up grants nothing" only holds for a host that gates on `confirmed`, which is now true of both apps but isn't something the connector can assume.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in 191da76.

The receipt is now re-stamped while the session stays confirmed, so it slides exactly like the per-chain entries instead of being the one entry that expires. A reload after continuous activity no longer drops to the connect screen.

A one-hour consent lifetime was not intended — it was an oversight, not a decision.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in 191da76 — thank you, this was the sharpest catch in the review.

The vault now gates on the dialog's confirmation: `Connect.tsx` reads `useWalletConnect().connected` and requires it alongside `btcConnected && ethConnected`. So the X-dismiss path lands on the connect button rather than the fully-connected UI, and the Terms of Use acceptance the Connect control carries can't be skipped.

The vault was not intentionally different — I'd read the vault screens as out of scope for this deliverable (#2232 owns them), which was the wrong call when the change itself created the hole.

packages/babylon-wallet-connector/src/context/State.context.tsx
jonybur: Fixed in 191da76. The recovery analysis was exactly right and was the part I'd missed — the latch is what made this unrecoverable rather than merely annoying.

Both refs are gone. `initialRequiredChainIdsRef` and `coldStartRestoreEligibleRef` have been replaced by a single effect that keeps `confirmed` in step with the stored approval in *both* directions: it grants the confirmation when the approval covers the current requirements, and withdraws it when it stops doing so. A one-way latch could never re-open, which is precisely the failure you traced.

With the receipt now validated by coverage, the narrow-then-widen round trip never invalidates anything in the first place.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in 191da76 — re-stamped rather than documented, since the one-hour lifetime wasn't intended.

The receipt is now re-written while the session stays confirmed, so it expires with the session rather than an hour after the single confirm. Your read of the mechanism was exactly right: chain entries slide on every connect including auto-reconnect, and the receipt was the only entry that didn't.

packages/babylon-wallet-connector/src/core/confirmationReceipt.ts
jonybur: Fixed in 191da76.

`createConfirmationReceipt` no longer has an undefined return. `handleConfirm` checks up front that every required chain has a connected account and throws when one doesn't, which routes through the existing catch to `onError` plus the in-dialog error screen — so it surfaces instead of confirming with nothing persisted.

Your reachability analysis also pointed at a second hole next to it: the host's hooks are awaited, so a wallet can disconnect *during* the await and the old code would then confirm from the pre-await snapshot. The confirmation now re-reads the live connections after the await and refuses if a required chain has gone. Both paths are covered in `WalletDialog.test.tsx`.

packages/babylon-wallet-connector/src/__tests__/eth.entry.test.ts
jonybur: Removed in 191da76 — you're right, and thank you for running it to show the count.

I'd noticed it was vacuous when I wrote it and kept it anyway, which was the wrong call. Rather than add `peerDependenciesMeta` entries that aren't true yet, the block is gone; it belongs with the change that actually makes those peers optional (#2231), where it can assert something.

The rest of the entry-point test still has teeth — the module-graph walk is self-checked by asserting it *does* find Bitcoin in the root entry, so it can't silently pass by resolving nothing.

packages/babylon-wallet-connector/src/hooks/useWalletConnect.ts
jonybur: Fixed in 191da76.

`open` now does the same `hasOwnProperty` membership test as `disconnect`, so a bare `onClick={open}` is correct by construction rather than by a lookup miss. The vault's call site is unchanged and now correct.

I've kept the simple-staking wrappers: several of those call sites pass through props typed as zero-arg handlers, so the wrapper is what satisfies the type rather than the runtime behaviour. Happy to unwrap them in a follow-up if you'd rather they go.

jonybur (commented):
Re-review after 191da76. The earlier round covered the substance (unified AppKit singleton, coverage-based receipt, requirement-churn no longer signs out, receipt slides with the session, vault Connect gated on `walletSessionConfirmed`, per-chain disconnect, `hasOwnProperty` membership checks, and the eth-entry test is no longer vacuous). Traced each of the fix-in-191da76 replies against the code and they hold up.

One finding no earlier thread caught:

**`confirmedChains` is dead code.** In `packages/babylon-wallet-connector/src/core/confirmationReceipt.ts:144`, the exported `confirmedChains` helper is never imported anywhere in the monorepo — `grep -R confirmedChains packages services` returns only its definition. It ships as public API of the connector with no caller. Under the repo's zero-dead-code policy that is a delete, or a caller. If a specific consumer in a follow-up branch needs it, worth a comment naming that so a future sweep does not tear it out; otherwise remove it now — the receipt round-trip is already covered by `isValidConfirmationReceipt`, so nothing testable depends on this helper existing.

gbarkhatov (approved):

jonybur:
@greptile-apps

Issue #2222, opened 2026-08-12
Ledger: Pre-PegIn wallet-policy path (default BIP-86 policy + trusted address display)


jrwbabylonlab:
Pre-PegIn is the ONLY flow the device signs under a real wallet policy — `_validate_prepegin` is reached only when a policy is present, and it asserts the tx's SHA256d equals the approved `prepegin_txid`. Everything else in the vault uses the no-policy path.

Scope:

- Default BIP-86 policy: `name = ""`, template `tr(@0/**)`, `keys_info = ["[fpr/86'/<coin>'/0']xpub"]`; `wallet_id = sha256(serialized policy)`; no `registerWallet` needed for a default policy.
- Serve the policy preimage, raw template bytes and keys tree through the client-command loop (the one place the policy-side Merkle machinery is needed).
- `GET_MASTER_FINGERPRINT` — INS `0x05` per `commands.h`/`main.c` (their `docs/apdu.md` says `0x10`; the code wins — flagged to Ledger).
- This is also the only path that can put the depositor's funding address on the trusted display (`GET_WALLET_ADDRESS` with `display=1`) — closing the one security gap vs extension wallets identified in the 2026-08-12 architecture review.

**Dependencies:** depends on #2219 (signer core), sequenced after #2221 (PoP). ~1.5–2d.

---
**Scope clarification (verified during PR #2214 review):** on-device address confirmation must cover REGISTRATION, not just display. `depositorPayoutBtcAddress` is host-supplied at ETH registration (contract validates length only; the derives-from-PoP check is host-side and bypassable by host malware), and no device screen ever shows it. Bounded severity: for the Ledger flow the registered script must byte-equal P2TR(device depositor key) or the deposit wedges at depositor payout-presign and falls back to HTLC refund — a wedge/DoS vector, not theft. The trusted-display work here is what closes it.

---
**Update 2026-08-18 — two prerequisites recorded during #2219:**
- **Binding (b1c-spec OQ3): SDK far-side `tapKeySig` verification must land before any key-path
  flow ships.** Script-path already verifies far-side (`assertScriptPathSchnorrSignature`, four
  call sites); key-path has NO far-side check today — `PeginManager` finalizes Pre-PegIn without
  one, software wallets included.
- The generated vector corpus has no realistic key-path coverage (zero master fingerprints, empty
  paths in `TAP_BIP32_DERIVATION`) — extend btc-vault's `ledger-vector-gen` when this starts.

[...]

PR #2269, opened 2026-08-18
feat(ledger-vault-signer): wire provider signPsbt/signPsbts onto sign…

jrwbabylonlab:
Closes the B3 remainder of #2219: the Ledger vault provider's signPsbt/signPsbts
(previously notWired) now drive the merged SIGN_PSBT core.

The provider holds a mirror of the device's intent state around the signing loop:
before any device I/O it requires a loaded intent, rejects replays (sha256 over
decoded PSBT bytes — the device dedup mask would nullify the intent, and NoPayout
has no mask at all), rejects key-path tables (Pre-PegIn/PoP need the wallet-policy
path, #2222/#2221) and autoFinalized:true. On success the intent stays loaded for
further PSBTs; on any device error the mirror drops pessimistically to idle
(firmware error-path invalidation is mixed — never assume survival). One ceremony
runs at a time (a concurrent APDU would be eaten with 0x6A80 and desync the
interrupt loop), with a generation-scoped lock that teardown releases synchronously
so a stale settling call can never touch a new connection's state. signPsbts is
strictly sequential, array-order, fail-fast, naming the failing index.

Signer-side, the entry point is split into exported stages (prepareSignPsbt +
signPreparedVaultPsbt; signVaultPsbt is their composition — an equivalence test
pins zero drift) so the provider inspects the expectation table before I/O, and
LedgerSignPsbtAbortedError now reports whether the initial APDU went out (breaking
constructor change; required so a pre-send abort doesn't arm the 0x6A80 resend
recovery). The abort→retry recovery path is intentionally unreachable until #2110
adds a UI cancel — teardown bumps the generation before aborting (documented
in-code and in the plan).

### DIFF
--- a/packages/babylon-ledger-vault-signer/src/__tests__/signPsbt.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/signPsbt.test.ts
@@ -163,7 +163,7 @@ describe("signVaultPsbt", () => {
     expect(isLedgerSignPsbtAbortedError(outcome)).toBe(true);
     if (!isLedgerSignPsbtAbortedError(outcome)) throw new Error("expected an aborted error");
     // No dispatcher was interrupted — the caller must not arm resend-once.
-    expect(outcome.sentInitialApdu).toBe(false);
+    expect(outcome.dispatcherInterrupted).toBe(false);
     expect(sent()).toBe(0);
   });
 
@@ -190,7 +190,8 @@ describe("signVaultPsbt", () => {
 
     expect(isLedgerSignPsbtAbortedError(outcome)).toBe(true);
     if (!isLedgerSignPsbtAbortedError(outcome)) throw new Error("expected an aborted error");
-    expect(outcome.sentInitialApdu).toBe(true);
+    // The 0xE000 arrived, so a dispatcher is mid-interruption awaiting CONTINUE.
+    expect(outcome.dispatcherInterrupted).toBe(true);
     expect(sent()).toBe(1);
   });
 
@@ -213,6 +214,29 @@ describe("signVaultPsbt", () => {
     expect(stagedSender.sent()).toBe(script.length);
     expect(staged.signedPsbtHex).toBe(composed.signedPsbtHex);
     expect(staged.yields).toEqual(composed.yields);
+  });
+
+  it("threads resendOnceOnIncorrectData through the composed entry point", async () => {
+    // Pins the options forwarding: a first-exchange 0x6A80 with the recovery
+    // armed resends the identical APDU once, then the ceremony completes.
+    const leafHashHex = fixtureLeafHashHex();
+    const script = peginScript(leafHashHex);
+    const initialHex = script[0].expectApduHex;
+    const withEatenFirst: ScriptedExchange[] = [
+      { expectApduHex: initialHex, respondSw: 0x6a80, respondDataHex: "" },
+      ...script,
+    ];
+    const { send, sent } = createScriptedSender(withEatenFirst);
+
+    const result = await signVaultPsbt(send, {
+      psbtHex: vector.psbt_hex,
+      depositorXOnlyHex: TEST_DEPOSITOR_KEY_HEX,
+      signal: new AbortController().signal,
+      resendOnceOnIncorrectData: true,
+    });
+
+    expect(sent()).toBe(withEatenFirst.length);
+    expect(result.yields).toHaveLength(1);
   });
 
   it("rejects reuse of a prepared signing state after success, with zero device I/O", async () => {
--- a/packages/babylon-ledger-vault-signer/src/__tests__/signPsbt.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/signPsbt.test.ts
@@ -17,7 +17,8 @@ import { describe, expect, it } from "vitest";
 
 import { isLedgerSignPsbtAbortedError } from "../errors";
 import type { RawApduSender } from "../rawApdu";
-import { signVaultPsbt, type SignVaultPsbtParams } from "../signPsbt";
+import { signPreparedVaultPsbt, signVaultPsbt, type SignVaultPsbtParams } from "../signPsbt";
+import { prepareSignPsbt } from "../signPsbtPrepare";
 import { tapLeafHash } from "../tapLeafHash";
 
 const VECTORS_DIR = join(__dirname, "..", "vendor", "ledger-bitcoin", "__tests__", "vectors", "signpsbt");
@@ -160,6 +161,9 @@ describe("signVaultPsbt", () => {
     );
 
     expect(isLedgerSignPsbtAbortedError(outcome)).toBe(true);
+    if (!isLedgerSignPsbtAbortedError(outcome)) throw new Error("expected an aborted error");
+    // No dispatcher was interrupted — the caller must not arm resend-once.
+    expect(outcome.sentInitialApdu).toBe(false);
     expect(sent()).toBe(0);
   });
 
@@ -185,7 +189,30 @@ describe("signVaultPsbt", () => {
     );
 
     expect(isLedgerSignPsbtAbortedError(outcome)).toBe(true);
+    if (!isLedgerSignPsbtAbortedError(outcome)) throw new Error("expected an aborted error");
+    expect(outcome.sentInitialApdu).toBe(true);
     expect(sent()).toBe(1);
+  });
+
+  it("staged API signs identically: prepareSignPsbt + signPreparedVaultPsbt === signVaultPsbt", async () => {
+    const script = peginScript(fixtureLeafHashHex());
+    const composedSender = createScriptedSender(script);
+    const composed = await signVaultPsbt(composedSender.send, {
+      psbtHex: vector.psbt_hex,
+      depositorXOnlyHex: TEST_DEPOSITOR_KEY_HEX,
+      signal: new AbortController().signal,
+    });
+
+    const prepared = prepareSignPsbt({ psbtHex: vector.psbt_hex, depositorXOnlyHex: TEST_DEPOSITOR_KEY_HEX });
+    const stagedSender = createScriptedSender(script);
+    const staged = await signPreparedVaultPsbt(stagedSender.send, prepared, {
+      signal: new AbortController().signal,
+    });
+
+    expect(composedSender.sent()).toBe(script.length);
+    expect(stagedSender.sent()).toBe(script.length);
+    expect(staged.signedPsbtHex).toBe(composed.signedPsbtHex);
+    expect(staged.yields).toEqual(composed.yields);
   });
 
   it("requires a signal — params without one must not typecheck", () => {
--- a/packages/babylon-ledger-vault-signer/src/__tests__/signPsbt.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/signPsbt.test.ts
@@ -215,6 +215,37 @@ describe("signVaultPsbt", () => {
     expect(staged.yields).toEqual(composed.yields);
   });
 
+  it("rejects reuse of a prepared signing state after success, with zero device I/O", async () => {
+    // The prepared object's collector/interpreter are mutable ceremony state —
+    // a second run would repeat a non-idempotent ceremony with stale yields.
+    const script = peginScript(fixtureLeafHashHex());
+    const prepared = prepareSignPsbt({ psbtHex: vector.psbt_hex, depositorXOnlyHex: TEST_DEPOSITOR_KEY_HEX });
+    await signPreparedVaultPsbt(createScriptedSender(script).send, prepared, {
+      signal: new AbortController().signal,
+    });
+
+    const { send, sent } = createScriptedSender(script);
+    await expect(signPreparedVaultPsbt(send, prepared, { signal: new AbortController().signal })).rejects.toThrow(
+      /already used/,
+    );
+    expect(sent()).toBe(0);
+  });
+
+  it("rejects reuse of a prepared signing state after an abort", async () => {
+    const prepared = prepareSignPsbt({ psbtHex: vector.psbt_hex, depositorXOnlyHex: TEST_DEPOSITOR_KEY_HEX });
+    const aborted = new AbortController();
+    aborted.abort();
+    await expect(
+      signPreparedVaultPsbt(createScriptedSender([]).send, prepared, { signal: aborted.signal }),
+    ).rejects.toThrow(/abandoned/);
+
+    const { send, sent } = createScriptedSender([]);
+    await expect(signPreparedVaultPsbt(send, prepared, { signal: new AbortController().signal })).rejects.toThrow(
+      /already used/,
+    );
+    expect(sent()).toBe(0);
+  });
+
   it("requires a signal — params without one must not typecheck", () => {
     // @ts-expect-error `signal` is required; the loop has no round cap and no
     // internal timeout, so a caller without a signal cannot bound the ceremony.
--- a/packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
@@ -475,6 +475,35 @@ describe("abort signal (T7 + entry check)", () => {
     expect(withoutSignal.resendOnceOnIncorrectData).toBe(true);
   });
 
+  it("an abort racing the armed resend skips the recovery resend, sentInitialApdu true", async () => {
+    // The one 0x6A80 site with resend-once armed: the initial APDU went out,
+    // so the abort must report an interrupted dispatcher — but never resend.
+    const prepared = prepareFromVector("generated__deposit-flow__pegin__0");
+    const controller = new AbortController();
+    const script: ScriptedExchange[] = [
+      {
+        expectApduHex: initialApduHexOf(prepared),
+        respondSw: 0x6a80,
+        respondDataHex: "",
+        onRespond: () => controller.abort(),
+      },
+    ];
+    const { send, sent } = createScriptedSender(script);
+
+    const outcome = await runSignPsbtLoop(send, prepared, {
+      signal: controller.signal,
+      resendOnceOnIncorrectData: true,
+    }).then(
+      () => undefined,
+      (error: unknown) => error,
+    );
+
+    expect(isLedgerSignPsbtAbortedError(outcome)).toBe(true);
+    if (!isLedgerSignPsbtAbortedError(outcome)) throw new Error("expected an aborted error");
+    expect(outcome.sentInitialApdu).toBe(true);
+    expect(sent()).toBe(1);
+  });
+
   it("reports how many yields had already arrived when the host aborted", async () => {
     const prepared = prepareFromVector("generated__deposit-flow__pegin__0");
     const controller = new AbortController();
--- a/packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
@@ -32,7 +32,12 @@ import {
 import { createYieldCollector, type ExpectedSignatureTable } from "../expectedSignatures";
 import type { Apdu, RawApduSender } from "../rawApdu";
 import { runSignPsbtLoop, type RunSignPsbtLoopOptions, type SignPsbtProgress } from "../signPsbtLoop";
-import { prepareSignPsbt, type PreparedSignPsbt } from "../signPsbtPrepare";
+import {
+  getPreparedSignPsbtState,
+  makePreparedSignPsbt,
+  prepareSignPsbt,
+  type PreparedSignPsbt,
+} from "../signPsbtPrepare";
 import { ClientCommandInterpreter } from "../vendor/ledger-bitcoin/clientCommands";
 import { createVarint } from "../vendor/ledger-bitcoin/varint";
 
@@ -91,7 +96,7 @@ function createScriptedSender(script: readonly ScriptedExchange[]): { send: RawA
 }
 
 function initialApduHexOf(prepared: PreparedSignPsbt): string {
-  return SIGN_PSBT_HEADER_HEX + Buffer.from(prepared.cdata).toString("hex");
+  return SIGN_PSBT_HEADER_HEX + Buffer.from(getPreparedSignPsbtState(prepared).cdata).toString("hex");
 }
 
 /** `0x10 ‖ varint(i) ‖ 0x40 ‖ signer(32) ‖ leaf(32) ‖ sig(64)` — wire-spec §5 tapscript YIELD. */
@@ -219,13 +224,10 @@ describe("happy-path trace replay through the loop (T1, T11, T12)", () => {
     const collector = createYieldCollector(table);
     const interpreter = new ClientCommandInterpreter(undefined, (payload) => collector.assertAndRecord(payload));
     interpreter.addKnownList(elements);
-    const prepared: PreparedSignPsbt = {
-      cdata: Uint8Array.from(Buffer.alloc(4, 0x11)),
-      interpreter,
-      collector,
-      table,
-      originalPsbtHex: "",
-    };
+    const prepared = makePreparedSignPsbt(
+      { table, unsignedTxid: "00".repeat(32) },
+      { cdata: Uint8Array.from(Buffer.alloc(4, 0x11)), interpreter, collector, originalPsbtHex: "" },
+    );
 
     const traces = [
       ...traceFile.traces,
@@ -287,7 +289,7 @@ describe("resend-once on 0x6A80 (T2-T5)", () => {
         resendOnceOnIncorrectData: true,
         signal: controller.signal,
       }),
-    ).rejects.toMatchObject({ name: LedgerSignPsbtAbortedError.name });
+    ).rejects.toMatchObject({ name: LedgerSignPsbtAbortedError.name, dispatcherInterrupted: false });
     expect(sent()).toBe(1);
   });
 
@@ -473,35 +475,6 @@ describe("abort signal (T7 + entry check)", () => {
     const withoutSignal: RunSignPsbtLoopOptions = { resendOnceOnIncorrectData: true };
 
     expect(withoutSignal.resendOnceOnIncorrectData).toBe(true);
-  });
-
-  it("an abort racing the armed resend skips the recovery resend, sentInitialApdu true", async () => {
-    // The one 0x6A80 site with resend-once armed: the initial APDU went out,
-    // so the abort must report an interrupted dispatcher — but never resend.
-    const prepared = prepareFromVector("generated__deposit-flow__pegin__0");
-    const controller = new AbortController();
-    const script: ScriptedExchange[] = [
-      {
-        expectApduHex: initialApduHexOf(prepared),
-        respondSw: 0x6a80,
-        respondDataHex: "",
-        onRespond: () => controller.abort(),
-      },
-    ];
-    const { send, sent } = createScriptedSender(script);
-
-    const outcome = await runSignPsbtLoop(send, prepared, {
-      signal: controller.signal,
-      resendOnceOnIncorrectData: true,
-    }).then(
-      () => undefined,
-      (error: unknown) => error,
-    );
-
-    expect(isLedgerSignPsbtAbortedError(outcome)).toBe(true);
-    if (!isLedgerSignPsbtAbortedError(outcome)) throw new Error("expected an aborted error");
-    expect(outcome.sentInitialApdu).toBe(true);
-    expect(sent()).toBe(1);
   });
 
   it("reports how many yields had already arrived when the host aborted", async () => {
--- a/packages/babylon-ledger-vault-signer/src/errors.ts
+++ b/packages/babylon-ledger-vault-signer/src/errors.ts
@@ -189,9 +189,8 @@ export class LedgerSignPsbtAbortedError extends Error {
   /** Yields the device had already delivered when the host stopped sending. */
   readonly yieldedCount: number;
   /**
-   * False only when the abort fired BEFORE the initial SIGN_PSBT APDU went
-   * out: no dispatcher was interrupted, so the caller must NOT arm the
-   * resend-once recovery (it would mask one genuine 0x6A80 later).
+   * False only when the abort fired before the initial APDU went out — no
+   * dispatcher was interrupted, so callers must not arm the resend-once recovery.
    */
   readonly sentInitialApdu: boolean;
 
--- a/packages/babylon-ledger-vault-signer/src/errors.ts
+++ b/packages/babylon-ledger-vault-signer/src/errors.ts
@@ -188,11 +188,18 @@ export class LedgerSignPsbtProtocolError extends Error {
 export class LedgerSignPsbtAbortedError extends Error {
   /** Yields the device had already delivered when the host stopped sending. */
   readonly yieldedCount: number;
+  /**
+   * False only when the abort fired BEFORE the initial SIGN_PSBT APDU went
+   * out: no dispatcher was interrupted, so the caller must NOT arm the
+   * resend-once recovery (it would mask one genuine 0x6A80 later).
+   */
+  readonly sentInitialApdu: boolean;
 
-  constructor(yieldedCount: number) {
+  constructor(yieldedCount: number, sentInitialApdu: boolean) {
     super("SIGN_PSBT was abandoned host-side before completion");
     this.name = LEDGER_SIGN_PSBT_ABORTED_ERROR_NAME;
     this.yieldedCount = yieldedCount;
+    this.sentInitialApdu = sentInitialApdu;
   }
 }
 
--- a/packages/babylon-ledger-vault-signer/src/errors.ts
+++ b/packages/babylon-ledger-vault-signer/src/errors.ts
@@ -189,16 +189,16 @@ export class LedgerSignPsbtAbortedError extends Error {
   /** Yields the device had already delivered when the host stopped sending. */
   readonly yieldedCount: number;
   /**
-   * False only when the abort fired before the initial APDU went out — no
-   * dispatcher was interrupted, so callers must not arm the resend-once recovery.
+   * True only when the abort left a dispatcher mid-interruption (device
+   * awaiting CONTINUE) — the one state the resend-once recovery exists for.
    */
-  readonly sentInitialApdu: boolean;
+  readonly dispatcherInterrupted: boolean;
 
-  constructor(yieldedCount: number, sentInitialApdu: boolean) {
+  constructor(yieldedCount: number, dispatcherInterrupted: boolean) {
     super("SIGN_PSBT was abandoned host-side before completion");
     this.name = LEDGER_SIGN_PSBT_ABORTED_ERROR_NAME;
     this.yieldedCount = yieldedCount;
-    this.sentInitialApdu = sentInitialApdu;
+    this.dispatcherInterrupted = dispatcherInterrupted;
   }
 }
 
--- a/packages/babylon-ledger-vault-signer/src/index.ts
+++ b/packages/babylon-ledger-vault-signer/src/index.ts
@@ -39,6 +39,7 @@ export {
   type CollectedYieldRef,
   type LedgerYieldMismatchKind,
 } from "./errors";
+export type { ExpectedSignatureTable, InputSigExpectation } from "./expectedSignatures";
 export {
   encodeIntentGroup,
   encodeIntentScalars,
@@ -46,7 +47,14 @@ export {
   type IntentScalars,
   type IntentVaultGroup,
 } from "./intentTlv";
-export { type Apdu, type AppIdentity, type RawApduResponse, type RawApduSender } from "./rawApdu";
+export {
+  SW_BAD_STATE,
+  SW_CAP_EXCEEDED,
+  type Apdu,
+  type AppIdentity,
+  type RawApduResponse,
+  type RawApduSender,
+} from "./rawApdu";
 export {
   signPreparedVaultPsbt,
   signVaultPsbt,
--- a/packages/babylon-ledger-vault-signer/src/index.ts
+++ b/packages/babylon-ledger-vault-signer/src/index.ts
@@ -11,7 +11,7 @@
  */
 
 export { getXOnlyPublicKeyHex } from "./derivation";
-export { createDmkApduSender } from "./dmkApduSender";
+export { createDmkApduSender, createDmkRawApduSender } from "./dmkApduSender";
 export { closeDmk, connectDmkSession, disconnectDmkSession, isSessionAlive, type DmkSessionHandle } from "./dmkSession";
 export { assertDepositTermsDeviceCompatible } from "./envelope";
 export {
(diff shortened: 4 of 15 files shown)


packages/babylon-ledger-vault-signer/src/__tests__/signPsbt.test.ts
```diff
@@ -160,6 +161,9 @@ describe("signVaultPsbt", () => {
     );
 
     expect(isLedgerSignPsbtAbortedError(outcome)).toBe(true);
+    if (!isLedgerSignPsbtAbortedError(outcome)) throw new Error("expected an aborted error");
+    // No dispatcher was interrupted — the caller must not arm resend-once.
+    expect(outcome.sentInitialApdu).toBe(false);
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — the repo's test philosophy prefers explicit inline values over imported constants for test readability.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -414,12 +483,255 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+      this.withSignAbort((controller) => this.signOnePsbt(psbtHex, options, controller)),
+    );
+
+  /**
+   * Strictly sequential, array order, fail-fast — one interrupt loop at a
+   * time; a concurrent APDU would be eaten with 0x6A80 and desync the loop.
+   * A mid-batch failure rejects the whole batch (no partial results); the
+   * error's mirror rule decides whether a full re-ceremony is needed.
+   */
+  signPsbts = async (psbtsHexes: string[], options?: SignPsbtOptions[]): Promise<string[]> =>
```
gbarkhatov, line 541: **Batch gates run per element, not per batch.**

`signPsbts` runs every zero-device-I/O gate (hex, `autoFinalized`, `prepareSignPsbt`, keypath, fingerprint) lazily inside `signOnePsbt`, so the "zero device I/O on any rejection" property holds per element but not for the batch — which is the unit the caller submits and the unit that fails.

`signPsbts([A, B])` where B trips any host-detectable gate: A is signed on-device, the batch rejects, A's signature is discarded, and A's fingerprint now blocks the retry. Recovery is a full DERIVE + APPROVE, every value screen again. The test at `provider.test.ts:430` documents exactly this outcome.

The staged API this PR adds is precisely what a whole-batch pre-pass needs: validate options and hex, `prepareSignPsbt` each element (independent single-use objects), run the keypath gate, reject intra-batch duplicate fingerprints — then sign from the prepared array. `signPsbts([P, P])` is the fully decidable case: today element 0 burns a real device signature before element 1 hits the replay guard, and nothing but `psbtsHexes` is needed to catch it up front.

Minor, same site: `signPsbts([])` resolves `[]` having run no gates at all, on a disconnected or un-approved provider. `onekey`, `unisat`, `utila` and `keystone` all throw `PSBTS_HEXES_REQUIRED` for that input.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -414,12 +483,255 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+        throw new WalletError(
+          {
+            code: ERROR_CODES.WALLET_NOT_CONNECTED,
+            message: `${WALLET_PROVIDER_NAME} connection changed during signing; restart from derivation.`,
+            wallet: WALLET_PROVIDER_NAME,
+          },
+          { cause: error instanceof Error ? error : undefined },
+        );
+      }
+      if (isLedgerSignPsbtAbortedError(error)) {
```
gbarkhatov, line 734: **The abandonment branch ignores `yieldedCount`.**

This reads `sentInitialApdu` but not `yieldedCount`, which is the discriminator the error already carries: `yieldedCount > 0` means the device emitted signatures for this PSBT.

`errors.ts:174-177` states that a Payout/NoPayout **signing-phase** abandonment invalidates the intent. Keeping the mirror at `intent-loaded` and leaving `signedFingerprints` untouched means the aborted PSBT stays retryable — and since the fingerprint is only recorded on full success (`:654`), a retry re-runs SIGN_PSBT on a PSBT the device has already signed. NoPayout has no device dedup mask, so by this file's own comment at `:596-598` the host set is "the only defense for that class", and here it never fires.

Suggest: `yieldedCount > 0` → treat the intent as burned (drop the mirror, clear the bookkeeping); keep the optimistic mirror only for `yieldedCount === 0`. Unreachable today because teardown bumps the generation first — but so is the `sentInitialApdu` logic this PR adds, and both go live together with #2110.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -54,15 +61,19 @@ const HARDENED = 0x80000000;
  */
 type DeviceIntentState = { phase: "idle" } | { phase: "derived" } | { phase: "intent-loaded"; termsKey: string };
 
+/** Device answers when the loaded intent is gone (SW_BAD_STATE) or a cap/dedup mask tripped (SW_CAP_EXCEEDED). */
+const SW_BAD_STATE = 0xb007;
```
gbarkhatov: **Status words re-declared without a firmware citation.**

`0xb007`/`0xb00a` are already tabulated with their provenance in the signer package — `rawApdu.ts` `STATUS_WORDS`, cited to `app-babylon-vault sw.h`. Two copies can drift, and drift here is silent: the wrong constant simply stops matching, with no compile error and no failing test, which breaks the "intent gone" routing #2110's UX depends on.

CLAUDE.md §7 asks APDU-layer constants to cite their firmware source. Either export the two codes from the signer package and import them here, or carry the same `sw.h` citation.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -414,12 +483,255 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+            code: ERROR_CODES.WALLET_NOT_CONNECTED,
+            message: `${WALLET_PROVIDER_NAME} stopped signing (disconnected); reconnect and retry.`,
+            wallet: WALLET_PROVIDER_NAME,
+          },
+          { cause: error },
+        );
+      }
+      // Everything else: pessimistically assume the device dropped the intent
+      // (error-path invalidation is mixed in firmware — never assume survival).
+      this.deviceState = { phase: "idle" };
```
gbarkhatov, line 751: **After any non-abort signing failure, the presign flow's Retry button is dead.**

Dropping the mirror to `idle` is correct for the provider. The problem is on the caller side: `runDepositorPresignFlow.ts:462` calls `approveDepositTerms` but never `deriveContextHash`, and approving from `idle` throws "no freshly derived context root … Restart the flow from derivation."

Walked it end to end: depositor declines the SIGN_PSBT screen on NoPayout #2 → mirror `idle` → they press Retry → `ensureAuthenticatedVpClient.ts:54` hits `vpTokenRegistry.peek()` and returns early on the documented hot path, so no derive happens → `approveDepositTerms` throws. `useDepositFlow.ts:1310` deliberately keeps registry entries on non-abort errors ("the user may retry, in which case the cache hit avoids a second wallet popup"), so every in-modal retry takes the same cached path. `useReleaseVpTokenOnUnmount` does release on unmount, so closing and reopening the modal recovers — but Retry itself is a dead button.

Contrast the Pre-PegIn path, which derives then approves on every attempt. B3 turns this from unreachable into the default post-failure state, so it is worth closing with this PR: either re-derive before approving when the wallet is approval-capable, or release the registry entry on a mirror-invalidating error.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -414,12 +483,255 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+    // Canonical fingerprint over DECODED bytes — a case-variant hex of the
+    // same PSBT must not bypass the replay guard.
+    if (!/^(?:[0-9a-fA-F]{2})+$/.test(psbtHex)) {
+      throw new WalletError({
+        code: ERROR_CODES.INVALID_PARAMS,
+        message: `${label} needs even-length hex; got ${psbtHex.length} chars.`,
+        wallet: WALLET_PROVIDER_NAME,
+      });
+    }
+    const fingerprint = bcrypto.sha256(Buffer.from(psbtHex, "hex")).toString("hex");
```
gbarkhatov: **The replay guard is keyed on PSBT serialization, not on the signing request.**

The fingerprint is `sha256` over the PSBT wire bytes, but PSBT maps are unordered and carry fields that never enter a BIP-341 sighash — so two byte-different hexes can be the same signing request. A round-trip through `Psbt.fromHex(hex).toHex()`, or one extra unknown key-value, changes the fingerprint while `prepared.table.byInput` stays identical: same input, same leaf hash, same control block, same signature requested.

Payout is caught by the firmware dedup mask (0xB00A). NoPayout has no mask, and the comment two lines below calls this set the only defense for that class — so the guard protecting the unprotected case is keyed on a proxy for the thing it protects. Today's callers hand builder output over verbatim, so exposure is low; the concern is that the property is incidental rather than structural.

`prepared` already computes the real identity: the merge target's unsigned-tx txid plus the `(inputIndex, leafHash)` set from `table.byInput`. Keying on that closes it. It means moving the fingerprint check below `prepareSignPsbt`, which is safe — prepare is pure and host-only, so the zero-device-I/O property is unaffected.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -414,12 +483,255 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+  ): Promise<string> {
+    const label = batchIndex === undefined ? "signPsbt" : `signPsbts[${batchIndex}]`;
+    const session = this.requireSession();
+    const rawSend = this.requireRawSender();
+    const generation = this.connectionGeneration;
+    // Fail actionably now rather than with an opaque status word mid-loop. A
+    // dead session means the device state is gone (unplug wipes it) — the
+    // mirror and replay bookkeeping must not survive it. Generation-guarded:
+    // a reconnect may already have re-established fresh state.
+    if (!(await this.probeSessionAlive(session))) {
```
gbarkhatov, line 599: **The dead-session branch is a partial teardown.**

It concludes the session is dead and resets `deviceState`, `signedFingerprints` and `loopAbandoned` — but unlike `teardownSession` (`:271-289`) it leaves `session`, `send`, `rawSend` and, critically, `pubkeyHexPromise` installed, and does not bump `connectionGeneration`.

Consequence: user unplugs mid-flow; `signPsbt` correctly reports "was disconnected"; user re-plugs without reconnecting and opens the payout modal → `verifyBtcWalletLiveness(provider, addr, { probeConnection: false })` — hardware wallets are deliberately excluded from the round-trip probe — only calls `getAddress()`, which resolves from the stale cached pubkey and passes. The guard that exists to stop "modal opens, no signing popup appears" green-lights a device the provider has already proven gone; the flow then dies one step later inside the ceremony.

`teardownSession()`'s clears are a strict superset, so calling it here closes both. The `generation === this.connectionGeneration` check on `:551` only exists because this branch declines to bump.

Same shape at `:374`: `doApproveDepositTerms` detects the identical condition and throws with no reset at all, leaving `deviceState: "derived"`. If the session recovers without a teardown, a retried approve skips the `phase === "idle"` gate and dies at the first APDU with SW_BAD_STATE.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -414,12 +483,255 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+          message: `${label} rejected before device I/O: ${error instanceof Error ? error.message : String(error)}`,
+          wallet: WALLET_PROVIDER_NAME,
+        },
+        { cause: error instanceof Error ? error : undefined },
+      );
+    }
+    // B1 signs the no-policy tapscript flows only: a keypath expectation means
+    // Pre-PegIn/PoP, which need the wallet-policy path. Reject actionably here
+    // instead of letting the device answer an opaque 0x6A80 mid-loop.
+    for (const expectation of prepared.table.byInput.values()) {
```
gbarkhatov, line 669: **An empty expectation table round-trips an unsigned PSBT as "signed".**

This loop scans for keypath expectations but never asserts the map is non-empty. Given a PSBT whose inputs the table recognises none of, `expectedYieldCount === 0`, `collector.assertComplete()` passes vacuously, `mergeYields` returns the original bytes — and the provider hands them back as `signedPsbtHex` **and** records a fingerprint for them, so the PSBT is now unsignable under this intent while never having been signed.

Unreachable from today's SDK builders (they always attach `tapLeafScript`/`tapInternalKey` on the signed input) and it would fail later at signature extraction. But a `prepared.table.byInput.size > 0` check belongs right next to the keypath gate: it is free, and it fails loudly instead of silently.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -443,46 +755,61 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+    return new WalletError(
+      {
+        code: ERROR_CODES.CONNECTION_FAILED,
+        message: dmk.originalError?.message ?? dmk._tag,
+        wallet: WALLET_PROVIDER_NAME,
+      },
+      { cause: error as Error },
+    );
+  }
+  return undefined;
```
gbarkhatov, line 894: **The signer's own typed SIGN_PSBT errors fall through unclassified.**

`LedgerYieldMismatchError`, `LedgerSignPsbtIncompleteError` and `LedgerSignPsbtProtocolError` all reach this `undefined` and land in the `UNKNOWN_ERROR` bucket at `:702-709`, alongside genuinely unclassified junk. The package exports `isLedgerYieldMismatchError`, `isLedgerSignPsbtIncompleteError` and `isLedgerSignPsbtProtocolError`, and this file already imports the sibling guards.

A yield mismatch means the device returned a signature the expected-signature table rejected — the one outcome that must never be retried — and it currently arrives indistinguishable by `code` from a stray JS error, while `intentGone` gets dedicated routing. These are the two strongest signals the loop produces; they deserve at least as much classification as a status word.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -443,46 +755,61 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+  }
+  if (isLedgerDeviceError(error)) {
+    // A stable, distinguishable message for the two "intent gone" words —
+    // #2110's UX routes these to a restart-from-derivation screen.
+    const intentGone = error.statusWord === SW_BAD_STATE || error.statusWord === SW_CAP_EXCEEDED;
+    return new WalletError(
+      {
+        code: ERROR_CODES.UNKNOWN_ERROR,
+        message: intentGone
+          ? `The device no longer holds the approved intent (${error.message}) — restart the flow from derivation.`
```
gbarkhatov: **This wording is now phase-wrong for derive and approve.**

Behaviour change worth a second look: on `main`, `withWalletErrorMapping` passed `LedgerDeviceError.message` through verbatim for every status word. Now that `toSignerWalletError` is shared with the ceremony sender, a 0xB007 raised during `deriveContextHash` or `approveDepositTerms` renders as "The device no longer holds the approved intent … restart the flow from derivation" — which is wrong at derive time, since nothing has been approved yet.

Either word it phase-neutrally ("the device is not in the expected state for this step — restart the flow from derivation") or pass the phase in and use the intent wording only at the sign seam.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -414,12 +483,255 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+    try {
+      return await isSessionAlive(session);
+    } catch (error) {
+      throw toSignerWalletError(error) ?? error;
+    }
+  }
 
-  signPsbts = async (psbtsHexes: string[]): Promise<string[]> =>
-    this.notWired(`signPsbts (${psbtsHexes.length} psbt(s))`);
+  private requireRawSender(): RawApduSender {
```
gbarkhatov: **This guard cannot execute.**

`session`, `send` and `rawSend` are assigned together (`:233-235`) and cleared together (`:272-274`), and `signOnePsbt` calls `requireSession()` one line before `requireRawSender()`. So `rawSend` is never undefined here — a non-executing branch carrying its own error string, which the zero-dead-code policy asks us not to keep. Either fold the raw sender into the session accessor or drop the guard.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -414,12 +483,255 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+        { cause: error instanceof Error ? error : undefined },
+      );
+    }
+  }
+
+  /**
+   * Liveness probe for the ceremony gates: `isSessionAlive` rethrows DMK's
+   * plain `{_tag}` objects, which would otherwise escape unmapped.
+   */
+  private async probeSessionAlive(session: DmkSessionHandle): Promise<boolean> {
```
gbarkhatov, line 762: **Good helper — but one call site still uses the raw function.**

`doConnect:215` still calls `isSessionAlive` directly, and that call sits outside the method's `try`/`catch`. A DMK rejection other than a missing session escapes `connectWallet()` as a bare `{ _tag, originalError }` object: `instanceof Error` misses it, `String(err)` is `"[object Object]"`, and there is no `.code` for the UI to branch on. That is exactly the failure this helper was written to prevent, and it is applied at `:374` and `:550` but not `:215`.

packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
```diff
@@ -101,7 +101,7 @@ export async function runSignPsbtLoop(
   if (response.sw === SW_INCORRECT_DATA && opts.resendOnceOnIncorrectData === true) {
     if (opts.signal.aborted) {
       // Abort raced the first exchange — do not issue the recovery resend.
-      throw new LedgerSignPsbtAbortedError(collector.yields.length);
+      throw new LedgerSignPsbtAbortedError(collector.yields.length, true);
```
gbarkhatov: **I think this should be `false`.**

The field's contract, per `errors.ts:191-195`, is not "did the APDU go out" but "was a dispatcher interrupted, so the caller must arm the resend-once recovery".

At this point the device has already answered 0x6A80 to the initial SIGN_PSBT. Under either reading of that word the recovery is spent: if it was the dispatcher eating the pending APDU, `process_interruption` does `SEND_SW(dc, SW_INCORRECT_DATA); return -1;` and the interrupted handler unwinds (`base:dispatcher.c:107-111`, this module's own citation); if it was a genuine validation rejection, no dispatcher was interrupted in the first place. Either way nothing is left to recover from.

Returning `true` keeps the provider's `loopAbandoned` armed (`provider.ts:676`), so the *next* attempt resends once against a clean device and masks one genuine validation 0x6A80 — precisely the hazard the docstring says the field exists to prevent. `:95` (`false`) and `:150` (`true`) are both correct as they stand.

packages/babylon-ledger-vault-signer/src/signPsbt.ts
```diff
@@ -63,25 +63,42 @@ export interface SignVaultPsbtResult {
...
   // Completion (collector.assertComplete) runs inside the loop's completing state.
   const yields = await runSignPsbtLoop(send, prepared, {
-    onProgress: params.onProgress,
-    signal: params.signal,
-    appIdentity: params.appIdentity,
-    resendOnceOnIncorrectData: params.resendOnceOnIncorrectData,
+    onProgress: options.onProgress,
+    signal: options.signal,
+    appIdentity: options.appIdentity,
+    resendOnceOnIncorrectData: options.resendOnceOnIncorrectData,
```
gbarkhatov: **Forward the options object instead of re-spreading it.**

Two things at this hand-written re-spread.

`resendOnceOnIncorrectData` is the one option with no coverage through the composed entry point — every test that exercises it calls `runSignPsbtLoop` directly, and `provider.test.ts` asserts it against a `vi.fn()` stub of `signPreparedVaultPsbt`, i.e. against the mock. Delete this line and all 55 signer tests plus the 59 connector ledger-vault tests stay green, while the `loopAbandoned` → resend recovery becomes a permanent silent no-op.

`SignPreparedVaultPsbtOptions` is `Omit<SignVaultPsbtParams, "psbtHex" | "depositorXOnlyHex">`, which is structurally identical to `RunSignPsbtLoopOptions` — so `runSignPsbtLoop(send, prepared, options)` passes straight through and removes this regression class for every future option too. Worth a composed-level test either way: arm the flag, script 0x6A80 followed by the normal trace, assert two initial APDUs.

packages/babylon-ledger-vault-signer/src/errors.ts
```diff
@@ -188,11 +188,18 @@ export class LedgerSignPsbtProtocolError extends Error {
...
   readonly yieldedCount: number;
+  /**
+   * False only when the abort fired BEFORE the initial SIGN_PSBT APDU went
+   * out: no dispatcher was interrupted, so the caller must NOT arm the
+   * resend-once recovery (it would mask one genuine 0x6A80 later).
+   */
+  readonly sentInitialApdu: boolean;
 
-  constructor(yieldedCount: number) {
+  constructor(yieldedCount: number, sentInitialApdu: boolean) {
```
gbarkhatov: **A required second parameter on an exported error class ships as a minor bump.**

This package is released independently (`nx.json` → `release.version.conventionalCommits: true`, registry npmjs) and `LedgerSignPsbtAbortedError` is exported from `index.ts`. Both commits here are `feat:`/`fix:` with no `!` and no `BREAKING CHANGE:` footer, so the constructor signature change publishes as a minor — a JS consumer on the old signature gets `sentInitialApdu === undefined`, and the recovery then never arms, silently.

Only in-repo consumers exist today, so either fix works: default the parameter, or mark the commit breaking.

packages/babylon-ledger-vault-signer/src/index.ts
```diff
@@ -48,12 +48,15 @@ export {
...
 export {
+  signPreparedVaultPsbt,
   signVaultPsbt,
   type CollectedYield,
+  type SignPreparedVaultPsbtOptions,
   type SignPsbtProgress,
   type SignVaultPsbtParams,
   type SignVaultPsbtResult,
 } from "./signPsbt";
+export { prepareSignPsbt, type PreparedSignPsbt } from "./signPsbtPrepare";
```
gbarkhatov, line 67: **`PreparedSignPsbt` becomes public here carrying more than the documented consumer need.**

`signPsbt.ts:76-79` says a caller wants `prepared.table`, and `provider.ts:630` is the only consumer — but the exported type also hands out the live `collector` and `interpreter`.

Tampering: `collector.assertAndRecord(...)` mutates the seen-set, so a caller could seed a `(input, leaf)` the device never signed, `assertComplete()`'s set-equality would pass on a ceremony that skipped an input, and `mergeYields` would ship a PSBT missing a required tapscript signature with no error anywhere.

Exposure: `collector.yields` is the internal array of `CollectedYield`, each holding the 64-byte Schnorr signature, and the provider now holds `prepared` across its whole `try`/`catch` — a Sentry breadcrumb or a `JSON.stringify` serializes raw signature bytes. The package works hard to keep that material off loggable surfaces (`CollectedYieldRef.signature?: never`, `toCollectedYieldRefs`, the detach at `signPsbt.ts:115`, which is shallow).

Narrowing the exported shape to `table` plus an opaque brand keeps every documented use working. The `PreparedSignPsbt` field types (`ExpectedSignatureTable`, `InputSigExpectation`) are also unexported, so a consumer factoring the keypath scan into a helper cannot name its parameter type.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
```diff
@@ -90,6 +102,11 @@ const INS_APPROVE_VAULT_INTENT = 0x80;
 /** APDUs belonging to the approval ceremony, in send order. */
 const approveApdus = () => h.sent.filter((a) => a.ins === INS_APPROVE_VAULT_INTENT);
 
+/** Minimal prepared shape the provider consumes: the table plus the merge source. */
+function fakePrepared(psbtHex: string, kind: "tapscript" | "taproot-keypath" = "tapscript") {
```
gbarkhatov: **The mock is not type-anchored, so the keypath gate can go dead with a green suite.**

`prepareSignPsbt: vi.fn()` is untyped and `fakePrepared` hand-rolls two of `PreparedSignPsbt`'s five fields, so nothing ties the string literal `"taproot-keypath"` here to `InputSigExpectation["kind"]` in `expectedSignatures.ts`.

Rename that discriminant in the signer package and `provider.ts:631` stops matching: Pre-PegIn/PoP PSBTs are no longer pre-rejected and instead drive a real ceremony that dies on an opaque status word **after** the depositor has approved on the device, burning the intent. This file stays green, because the stub hardcodes the same stale literal on both sides of the assertion.

It is the one place in the suite where an assertion verifies the mock rather than the code, and it guards the gate that motivates the whole staged API. `vi.fn<typeof prepareSignPsbt>()` / `vi.mocked` would anchor it; better still, build `fakePrepared` by calling the real `prepareSignPsbt` on a fixture and stub only `signPreparedVaultPsbt`.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
```diff
@@ -247,22 +271,267 @@ describe("LedgerVaultProvider", () => {
...
+
+    it("throws on autoFinalized: true instead of silently not finalizing", async () => {
+      const provider = await approved();
+
+      await expect(provider.signPsbt(PSBT_A, { autoFinalized: true })).rejects.toThrow(/never finalizes/);
+      expect(signMock.prepareSignPsbt).not.toHaveBeenCalled();
+      expect(signMock.signPreparedVaultPsbt).not.toHaveBeenCalled();
+    });
+
+    it("rejects malformed hex before hashing or device I/O", async () => {
```
gbarkhatov, line 351: **This test does not test the ordering its name claims.**

The name says the hex gate runs before the fingerprint, but the assertions only check that `prepareSignPsbt` was not called. Move the regex at `provider.ts:588` to *after* `bcrypto.sha256` and this test still passes.

That reordering is a live replay-guard bypass: `Buffer.from("aabbzz", "hex")` truncates to `aabb`, so `signPsbt("aabb")` followed by `signPsbt("aabbzz")` would collide on one fingerprint. Pin the ordering directly — sign `"aabb"`, then assert `signPsbt("aabbzz")` rejects with the *hex* message rather than the replay message.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
```diff
@@ -247,22 +271,267 @@ describe("LedgerVaultProvider", () => {
...
+      const provider = await approved();
+      dmkSessionMock.isSessionAlive.mockRejectedValueOnce({
+        _tag: "TransportError",
+        originalError: { message: "hid gone" },
+      });
+
+      await expect(provider.signPsbt(PSBT_A)).rejects.toMatchObject({ code: ERROR_CODES.CONNECTION_FAILED });
+    });
+
+    it("admits one device ceremony at a time", async () => {
```
gbarkhatov: **One test, four behaviours — and two real-timer waits.**

This asserts that the lock rejects a concurrent `signPsbt`; that it also rejects `deriveContextHash` and `approveDepositTerms`; that teardown releases the lock synchronously; and that teardown aborts the in-flight loop. Against "each test verifies ONE specific behavior", the two teardown properties want their own test — as written, a regression in teardown's abort and a regression in the lock produce the same single failure.

Also: the `vi.waitFor` here and at `:489` poll on real timers, and neither test is genuinely time-dependent — both just need "the mock has been entered". Resolving a deferred from inside the mock implementation removes the timer dependency and the flake surface.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
```diff
@@ -247,22 +271,267 @@ describe("LedgerVaultProvider", () => {
...
+      // believing an intent is loaded, nor block re-signing after recovery.
+      const provider = await approved();
+      await provider.signPsbt(PSBT_A);
+      dmkSessionMock.isSessionAlive.mockResolvedValueOnce(false);
+
+      await expect(provider.signPsbt(PSBT_B)).rejects.toThrow(/was disconnected/);
+      await expect(provider.signPsbt(PSBT_B)).rejects.toThrow(/no approved intent/);
+    });
+
+    it("a fresh approval resets the replay guard — the device counters were reset too", async () => {
```
gbarkhatov, line 625: **Coverage gaps around this test, all on guards that protect committed state.**

This pins the fingerprint reset on a fresh approval but not the `loopAbandoned = false` reset sitting next to it at `provider.ts:468` (nor at `:344` for derive). Delete line 468 and the whole suite stays green: an abort arms the flag, the user re-derives and re-approves, and the first SIGN_PSBT under the fresh intent silently resends once on a **genuine** 0x6A80 — a real prepare/merkleization bug masked instead of surfaced. A test is cheap here: mock an abort with `sentInitialApdu: true`, re-derive, re-approve, sign, assert the flag reached the loop as `false`.

Two more worth adding. `signOnePsbt`'s generation guards at `:551` and `:613` are untested (only the success-path stale settlement is covered, at `:478`): drop `:551`'s guard and a stale probe wipes a freshly reconnected, freshly approved intent; drop `:613`'s and a reconnect racing the cached pubkey read builds the expected-signature table against the old device's key. And the between-elements abort branch at `:509-515` — unreachable today, since teardown clears `session` and `requireSession()` throws first, but load-bearing the moment #2110's UI cancel aborts without a teardown.

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
```diff
@@ -475,6 +475,35 @@ describe("abort signal (T7 + entry check)", () => {
     expect(withoutSignal.resendOnceOnIncorrectData).toBe(true);
   });
 
+  it("an abort racing the armed resend skips the recovery resend, sentInitialApdu true", async () => {
```
gbarkhatov: **This is a strict superset of T3b at `:270`.**

Same PSBT vector, same one-exchange script, same 0x6A80, same `sent()` assertion, same armed flag — the two differ only in whether the abort fires in a sender wrapper or in `onRespond`. Per the no-test-bloat rule, the new `sentInitialApdu` assertion belongs on T3b rather than in a second copy of it.

Separately: this test's comment ("the initial APDU went out, so the abort must report an interrupted dispatcher") states the inference I am disputing on `signPsbtLoop.ts:104`. If that value flips to `false`, this expectation and its comment flip with it.

gbarkhatov (dismissed):
Reviewed the full `0dd433acd..HEAD` diff (8 files, ~815 insertions) plus the surrounding provider and signer surfaces the new code drives, not just the hunks.

**Verification.** Rebuilt the stale `babylon-ledger-vault-signer` `dist/` first — it predated #2264, which is what makes the connector tests fail locally on a fresh checkout. After that: signer package 526 passed / 5 skipped, connector ledger-vault suite 59/59, `tsc --noEmit` on the connector clean.

**Assessment.** The state machine holds up under scrutiny. Generation guards, the synchronous lock claim, identity-guarded `finally` releases, teardown ordering (generation bump before abort), and the fingerprint reset points all line up with the device's documented lifecycle; the five `signedFingerprints` clear sites each pair the clear with either a mirror drop to `idle` or a fresh `approveVaultIntent`, so "fingerprints cleared ⇒ signing blocked until a new ceremony" holds throughout. The staged `prepareSignPsbt` + `signPreparedVaultPsbt` split is behaviour-preserving, and the WeakSet claim happens synchronously before the first await, so success, abort and concurrent reuse are all rejected at zero device cost. Approving on that basis — none of the findings below blocks a flag-off wallet, but the `yieldedCount` handling, the presign retry path, and the `sentInitialApdu` value at the resend site all want fixing before the flag goes on, since all three go live together with #2110.

**Four things that did not fit an inline anchor:**

*Release readiness.* With the flag on, a depositor completes derive + intent approval + real on-device signatures, then hits `signMessage` for PoP (`PeginManager.ts`, still `notWired`) and past that the key-path Pre-PegIn broadcast, refused by the new keypath gate. `VaultWalletConnectionProvider` gates on the boolean flag only — nothing asks whether a wallet can sign BIP-322 or key-path before offering it. Worth either keeping the flag off until #2222/#2221 or adding a capability gate.

*Depositor-facing copy.* `depositErrors.ts:250` returns `sanitizeErrorMessage(err)` unedited under `ERRORS.defaultTitle`, and the new provider messages match none of the earlier buckets. The keypath rejection is reachable in production today, so a depositor can be shown "cannot sign key-path inputs (signPsbt) yet — Pre-PegIn and proof-of-possession need the wallet-policy path (#2222/#2221)". Internal issue numbers and SDK-integrator instructions in a wallet UI; either add `mapDepositError` buckets backed by `COPY.deposit.errors` or make the messages unreachable from the callout.

*Abort-handle asymmetry.* `signPsbt` now has an `AbortController` that teardown fires; `deriveContextHash` and `approveDepositTerms` still have none, and `approveVaultIntent` walks a multi-APDU chunk sequence. The host side is safe (after teardown the stale sender targets a disconnected session, and a newly-acquired lock hits `requireSender()` first), so this is not an interleaving bug — but a disconnect mid-approve can leave the *device* parked mid-ceremony with no host-side way to have stopped it. Pre-existing; noting it because this PR establishes the pattern that would fix it.

*Found in passing, pre-existing.* `connectWallet()` can resolve while leaving the provider disconnected — `connectPromise` dedupes on identity, is never invalidated by `disconnect()`, and both `disconnectToken` bail-outs return rather than throw. Relatedly, `getAddress`/`getPublicKeyHex` are the only await-then-return sites without `assertSameConnection`, so a slow read can hand back device A's address after a reconnect to device B.

**Checked and cleared.** `signInputs` and the tweak flags are ignored, but I verified this against every real call site rather than the comment: `createTaprootScriptPathSignOptions` is the only producer, and the SDK builders attach `tapLeafScript`/`tapInternalKey` only to the depositor-signed input, so `buildExpectedSignatureTable` never arms the others — the device's signed set equals what `signInputs` asks for. (If a future builder ever annotates a non-depositor input, this provider would over-sign where every other wallet single-signs; `table.byInput.keys() ⊆ signInputs` would close that.) No caller passes `autoFinalized: true`. Feature gating and capability detection are unchanged and correct — `signPsbtsWithFallback` probes `typeof wallet.signPsbts === "function"`, already true before this PR. No changeset needed: `nx release` with conventional commits covers `packages/*`. No E2E `data-testid` is touched; the real-wallet CLI drives injected providers only. `getAddress`/`getPublicKeyHex` sit outside the operation lock, but `signOnePsbt` warms the pubkey cache before entering the loop, so no APDU can interleave. `isSessionAlive` reads DMK session state rather than sending an APDU, so the per-element liveness probe cannot consume the dispatcher's one eaten APDU. The batch-index rewrap preserves `code`, message matching and the `cause` chain. No key material is logged on any new path.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: not planning — the device never counts an aborted yield (mask updates are post-CONTINUE); a retry hits 0xB007 which already drops to idle.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: not planning — prepareSignPsbt already rejects an empty table, pinned by existing tests.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
jrwbabylonlab: fixed — renamed to dispatcherInterrupted; this site now reports false.

packages/babylon-ledger-vault-signer/src/signPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/errors.ts
jrwbabylonlab: checked — 0.4.0 predates this class, so minor is the correct bump; leaving the commit unflagged.

packages/babylon-ledger-vault-signer/src/index.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
jrwbabylonlab: fixed — pinned via the derive-path reset; the :468 clear is unreachable with the flag armed.

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts

jeremy-babylonlabs (approved):

PR #2272, opened 2026-08-18
fix(ci): show the before and after pictures even when the set is large

jonybur:
## What

The Visual regression comment now always shows the before / after pictures inline, including on a pull request that changes several screens at once. On a busy run the set is published a little smaller so that all of it fits; full-resolution copies stay in the run's `visual-report` artifact, which the comment already links to.

## Why

On #2241 the comment showed a list of filenames and a "download the artifact" line instead of pictures, even though every step of the run was green.

The publish step caps what one pull request may add to the images branch at 512KB and refuses the whole set past it. Nothing on the composing side knew that ceiling existed - the caps there bound how many pictures are made and how large they may be, but not what they weigh - so the five vault composites came to 678KB, the publish was refused, and the comment fell back to the text listing the pictures exist to replace. This was not specific to that pull request: at today's picture sizes anything moving more than three vault screens hits it.

## How

- `scripts/visual-embed.mjs` now fits the set it composes to a byte budget that sits below the workflow's ceiling. It publishes at the first of 1280 / 960 / 800 / 640px that the whole set fits under, so a pull request moving one screen still gets it at the capture's own size, and one moving five gets all five, slightly reduced.
- A picture is given up only when the smallest width still busts the budget, and then from the bottom of the ranking. The comment already reports how many screens it is not showing.
- Pictures are published as truecolour instead of RGBA. Every source is an opaque screenshot, so the alpha channel was around a tenth of the published bytes.
- Reduction now targets any width rather than a whole factor, so the step below life size is 960px instead of 640px.
- A test pins that the workflow's ceiling stays above the script's budget. That is the check that was missing when the two drifted apart, and it is what stops this recurring the next time either number moves.

## Checked against the real run

Re-ran the composer over the report artifact from #2241's run:

| published at | set total | outcome |
| --- | --- | --- |
| 1280px (life size) | 608KB | over budget |
| 960px | 446KB | published - all five screens |

A single-screen change still publishes at 1280px and the comment says nothing about reduction.

### DIFF
--- a/scripts/__tests__/visual-embed.test.mjs
+++ b/scripts/__tests__/visual-embed.test.mjs
@@ -13,6 +13,8 @@ import {
   cropWindow,
   fitPublishedImages,
   panelLayout,
+  PUBLISHED_BYTE_MARGIN,
+  PUBLISHED_COLOUR_TYPE,
   PUBLISHED_TOTAL_BYTE_BUDGET,
   reduceToWidth,
   screenCaption,
@@ -328,8 +330,9 @@ function textured(width, height) {
 
 /** What one picture costs at a published width. */
 function cost(composite, width) {
-  return PNG.sync.write(reduceToWidth(composite, width), { colorType: 2 })
-    .length;
+  return PNG.sync.write(reduceToWidth(composite, width), {
+    colorType: PUBLISHED_COLOUR_TYPE,
+  }).length;
 }
 
 test("reduceToWidth leaves a picture already inside the width alone", () => {
@@ -346,9 +349,9 @@ test("reduceToWidth carries a one-column change into a fractional reduction", ()
   const composite = new PNG({ width: 10, height: 1 });
   composite.data.fill(0);
   for (let i = 3; i < composite.data.length; i += 4) composite.data[i] = 255;
-  composite.data[4 * 4] = 255;
-  composite.data[4 * 4 + 1] = 255;
-  composite.data[4 * 4 + 2] = 255;
+  composite.data[3 * 4] = 255;
+  composite.data[3 * 4 + 1] = 255;
+  composite.data[3 * 4 + 2] = 255;
 
   const reduced = reduceToWidth(composite, 7);
 
@@ -357,7 +360,12 @@ test("reduceToWidth carries a one-column change into a fractional reduction", ()
   for (let x = 0; x < reduced.width; x += 1) {
     if (reduced.data[x * 4] > 0) lit.push(x);
   }
-  assert.deepEqual(lit, [3]);
+  // Column 2 averages the lit source column with one dark neighbour, so it
+  // comes back at half. A nearest-neighbour reduction samples
+  // Math.floor(2 * 10/7) = 2 -> source column 2, which is dark, and reports
+  // nothing lit at all.
+  assert.deepEqual(lit, [2]);
+  assert.equal(reduced.data[2 * 4], 128);
 });
 
 test("fitPublishedImages publishes at the composite's own width when the set fits", () => {
@@ -425,6 +433,11 @@ test("fitPublishedImages gives up the lowest-ranked picture when the floor still
     fit.dropped.map((picture) => picture.key),
     ["vault/b.png"],
   );
+  // The ladder restarts from the top once a picture is gone: what is left
+  // fits at full width, and holding it at the floor would be a reduction
+  // nothing asked for. An implementation resuming at the rung it failed on
+  // would publish this at 640.
+  assert.equal(fit.width, 1280);
 });
 
 // The invariant the whole budget rests on. Publishing over it hands the
@@ -442,6 +455,156 @@ test("fitPublishedImages publishes nothing rather than exceeding the budget", ()
   );
 });
 
+// The block that runs to the source edge. `width * columnScale` is exactly
+// `source.width` in real arithmetic but lands a hair under it in IEEE-754,
+// so flooring the last block's end used to leave the final source column
+// unread - the right edge of the rightmost panel, and the hairline this
+// averaging exists to keep.
+test("reduceToWidth carries the last source column into the reduction", () => {
+  const composite = new PNG({ width: 803, height: 4 });
+  for (let y = 0; y < 4; y += 1) {
+    for (let x = 0; x < 803; x += 1) {
+      const i = (y * 803 + x) * 4;
+      composite.data[i] = x === 802 ? 255 : 0;
+      composite.data[i + 3] = 255;
+    }
+  }
+
+  const reduced = reduceToWidth(composite, 800);
+
+  assert.ok(
+    reduced.data[(reduced.width - 1) * 4] > 0,
+    "the last source column was dropped",
+  );
+});
+
+test("reduceToWidth carries the last source row into the reduction", () => {
+  // 642x200 -> 640 is one of the sizes where the last block's computed end
+  // lands short of the source; 642x400 is not, and would pass either way.
+  const composite = new PNG({ width: 642, height: 200 });
+  for (let y = 0; y < 200; y += 1) {
+    for (let x = 0; x < 642; x += 1) {
+      const i = (y * 642 + x) * 4;
+      composite.data[i] = y === 199 ? 255 : 0;
+      composite.data[i + 3] = 255;
+    }
+  }
+
+  const reduced = reduceToWidth(composite, 640);
+
+  assert.ok(
+    reduced.data[(reduced.height - 1) * reduced.width * 4] > 0,
+    "the last source row was dropped",
+  );
+});
+
+// Height is derived from the column scale, and nothing else keeps a reduced
+// composite from being squashed. A regression that scaled width alone still
+// writes a valid PNG and still fits the budget.
+test("reduceToWidth keeps the composite's shape", () => {
+  const reduced = reduceToWidth(textured(1280, 800), 640);
+
+  assert.equal(reduced.width, 640);
+  assert.equal(reduced.height, 400);
+});
+
+// The floor exists so a reviewer sees both surfaces. Taking the tail would
+// give up exactly the screens it protects: groups are ranked globally, and
+// a shipped route moving 1% of a tall page ranks below a story moving 17%
+// of a small frame.
+test("fitPublishedImages keeps every surface pictured when it has to give one up", () => {
+  const pictures = [
+    { key: "storybook/s1.png", composite: textured(1280, 300) },
+    { key: "storybook/s2.png", composite: textured(1280, 300) },
+    { key: "storybook/s3.png", composite: textured(1280, 300) },
+    { key: "vault/v1.png", composite: textured(1280, 300) },
+    { key: "vault/v2.png", composite: textured(1280, 300) },
+  ];
+  const budget = cost(pictures[0].composite, 640) * 3 + 100;
+
+  const fit = fitPublishedImages(pictures, budget);
+
+  assert.ok(
+    fit.published.some((picture) => picture.key.startsWith("vault/")),
+    `vault lost every picture: ${fit.published.map((p) => p.key).join(", ")}`,
+  );
+  assert.deepEqual(
+    fit.dropped.map((picture) => picture.key),
+    ["storybook/s3.png", "vault/v2.png"],
+  );
+});
+
+// One picture too big for the budget used to evict every picture that would
+// have fit, one at a time, before finally going itself - and the comment
+// fell back to the filename listing it exists to replace.
+test("fitPublishedImages gives up an oversized picture rather than the set around it", () => {
+  const pictures = [
+    { key: "vault/heavy.png", composite: textured(1280, 2400) },
+    { key: "vault/a.png", composite: textured(40, 20) },
+    { key: "vault/b.png", composite: textured(40, 20) },
+  ];
+
+  const fit = fitPublishedImages(pictures, 20 * 1024);
+
+  assert.deepEqual(
+    fit.published.map((picture) => picture.key),
+    ["vault/a.png", "vault/b.png"],
+  );
+  assert.deepEqual(
+    fit.dropped.map((picture) => picture.key),
+    ["vault/heavy.png"],
+  );
+});
+
+// Encoding moved out of the per-screen try/catch when it moved in here, so
+// one unencodable composite took every picture in the run with it.
+test("fitPublishedImages gives up a picture it cannot encode", () => {
+  const pictures = [
+    { key: "vault/broken.png", composite: { width: 8, height: 8, data: null } },
+    { key: "vault/a.png", composite: textured(40, 20) },
+  ];
+
+  const fit = fitPublishedImages(pictures, PUBLISHED_TOTAL_BYTE_BUDGET);
+
+  assert.deepEqual(
+    fit.published.map((picture) => picture.key),
+    ["vault/a.png"],
+  );
+  assert.equal(fit.dropped.length, 1);
+  assert.match(fit.dropped[0].reason, /could not be encoded/);
+});
+
+// The rung is a cap, not a scale factor. A mobile composite already inside
+// it is published untouched, and saying the whole set is at the rung tells
+// a reviewer measuring a control the wrong thing.
+test("fitPublishedImages reports the width each picture is actually published at", () => {
+  const pictures = [
+    { key: "vault/wide.png", composite: textured(1280, 300) },
+    { key: "vault/narrow.png", composite: textured(788, 300) },
+  ];
+  const budget =
+    cost(pictures[0].composite, 1280) + cost(pictures[1].composite, 1280) - 1;
+
+  const fit = fitPublishedImages(pictures, budget);
+
+  assert.ok(fit.width < 1280, `published at ${fit.width}px`);
+  assert.equal(fit.reduced, true);
+  assert.deepEqual(
+    fit.published.map((picture) => picture.publishedWidth),
+    [fit.width, 788],
+  );
+});
+
+test("fitPublishedImages reports nothing reduced when every composite is inside the rung", () => {
+  const fit = fitPublishedImages(
+    [{ key: "vault/narrow.png", composite: textured(788, 300) }],
+    PUBLISHED_TOTAL_BYTE_BUDGET,
+  );
+
+  assert.equal(fit.reduced, false);
+  assert.equal(fit.published[0].publishedWidth, 788);
+});
+
 // Two caps in two files that have to stay ordered. They were not, and the
 // gap is what this budget exists to close: the workflow refused a 678KB set
 // against its 512KB ceiling, and nothing on this side knew the ceiling was
@@ -455,13 +618,30 @@ test("the workflow's publish ceiling stays above the budget this script fits to"
     "utf8",
   );
 
-  const ceiling = workflow.match(
-    /MAX_PUBLISHED_BYTES=\$\(\(\s*(\d+)\s*\*\s*1024\s*\)\)/,
+  // Matched as an arithmetic expression rather than as one spelling of it:
+  // `512 * 1024` and `1024 * 512` are the same ceiling, and a pattern that
+  // only knows the first fails the whole visual job on a harmless edit.
+  const assignment = workflow.match(/MAX_PUBLISHED_BYTES=\$\(\((.+?)\)\)/);
+
+  assert.ok(assignment, "MAX_PUBLISHED_BYTES not found in the workflow");
+  assert.match(
+    assignment[1],
+    /^[\d\s*+]+$/,
+    `MAX_PUBLISHED_BYTES is not a plain arithmetic expression: ${assignment[1]}`,
+  );
+  const ceiling = Number(
+    new Function(`return (${assignment[1]});`)(),
   );
 
-  assert.ok(ceiling, "MAX_PUBLISHED_BYTES not found in the workflow");
   assert.ok(
-    Number(ceiling[1]) * 1024 > PUBLISHED_TOTAL_BYTE_BUDGET,
-    `ceiling ${ceiling[1]}KB must exceed the ${PUBLISHED_TOTAL_BYTE_BUDGET}-byte budget`,
+    Number.isFinite(ceiling) && ceiling > 0,
+    `MAX_PUBLISHED_BYTES did not evaluate to a size: ${assignment[1]}`,
+  );
+  // The margin, not just the order. The run this budget was measured on
+  // cleared it by 2KB, so a ceiling one byte above the budget would pass a
+  // check on ordering alone while leaving no room at all.
+  assert.ok(
+    ceiling >= PUBLISHED_TOTAL_BYTE_BUDGET + PUBLISHED_BYTE_MARGIN,
+    `ceiling ${ceiling} must clear the ${PUBLISHED_TOTAL_BYTE_BUDGET}-byte budget by ${PUBLISHED_BYTE_MARGIN} bytes`,
   );
 });
--- a/scripts/__tests__/visual-embed.test.mjs
+++ b/scripts/__tests__/visual-embed.test.mjs
@@ -1,12 +1,20 @@
 import assert from "node:assert/strict";
+import fs from "node:fs";
+import path from "node:path";
 import test from "node:test";
+import { fileURLToPath } from "node:url";
+
+import { PNG } from "pngjs";
 
 import {
   changedRegions,
   changedRowBand,
   composeBeforeAfter,
   cropWindow,
+  fitPublishedImages,
   panelLayout,
+  PUBLISHED_TOTAL_BYTE_BUDGET,
+  reduceToWidth,
   screenCaption,
   selectEmbeddedGroups,
 } from "../visual-embed.mjs";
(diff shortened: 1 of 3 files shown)


scripts/visual-embed.mjs
```diff
@@ -747,10 +810,72 @@ async function composeComposite(reportDir, surface, name) {
...
+      encoded.set(key, encodeAtWidth(picture.composite, width));
+    }
+    return encoded.get(key);
   };
+
+  const kept = pictures.map((picture, index) => ({ ...picture, index }));
+  const dropped = [];
+
+  while (kept.length > 0) {
+    for (const width of PUBLISHED_WIDTH_LADDER) {
```
jrwbabylonlab, line 988: narrower is not always fewer bytes. with your own `textured` fixture, 1186x1200 is 25,983 B native but 44,920 B at the 640 rung. so on a set that busts the budget every rung makes it worse, and we drop pictures while the comment says we reduced them.

scripts/visual-embed.mjs
```diff
@@ -747,10 +810,72 @@ async function composeComposite(reportDir, surface, name) {
...
+        published: kept.map((picture, index) => ({
+          ...picture,
+          bytes: bytes[index],
+        })),
+        dropped,
+      };
+    }
+    // Lowest-ranked first: `pictures` arrives in the order the comment
+    // will read in, so the tail is the end of it.
+    dropped.unshift(kept.pop());
```
jrwbabylonlab: `pop()` takes the lowest globally ranked picture, and vault groups rank last, so this drops the vault screens `MIN_EMBEDDED_GROUPS_PER_SURFACE` exists to protect. also `renderBody` iterates `surfaces`, so vault renders first, not last.

scripts/visual-embed.mjs
```diff
@@ -1103,49 +1236,83 @@ async function main() {
...
       process.stderr.write(`Skipping ${key}: pixel-identical to ${twin}.\n`);
       continue;
     }
     digests.set(digest, key);
 
-    const target = path.join(outDir, entry.surface, entry.screen.name);
+    composed.push({ key, entry, ...picture });
+  }
+
+  const { width, published, dropped } = fitPublishedImages(composed);
```
jrwbabylonlab: this moved outside the try/catch that says "this is the one place a single bad byte could take the lot". now one encode failure exits 1 and we fall back to the filename listing.

scripts/visual-embed.mjs
```diff
@@ -747,10 +810,72 @@ async function composeComposite(reportDir, surface, name) {
...
+    if (!encoded.has(key)) {
+      encoded.set(key, encodeAtWidth(picture.composite, width));
+    }
+    return encoded.get(key);
   };
+
+  const kept = pictures.map((picture, index) => ({ ...picture, index }));
+  const dropped = [];
+
+  while (kept.length > 0) {
```
jrwbabylonlab, line 987: only drops from the tail. one oversized top picture pops all the others one by one, then itself, and we publish nothing and exit 1.

scripts/visual-embed.mjs
```diff
@@ -687,33 +732,50 @@ function composeBeforeAfter(baseline, candidate) {
...
+    const top = Math.floor(y * rowScale);
+    const bottom = Math.min(
+      source.height,
+      Math.max(top + 1, Math.floor((y + 1) * rowScale)),
+    );
     for (let x = 0; x < width; x += 1) {
+      const left = Math.floor(x * columnScale);
+      const right = Math.min(
+        source.width,
+        Math.max(left + 1, Math.floor((x + 1) * columnScale)),
```
gbarkhatov: **The last source column and the last source row are silently dropped.**

`width * columnScale` is mathematically exactly `source.width`, but in IEEE-754 it usually lands just below it, so `Math.floor` returns `source.width - 1`. The `Math.min(source.width, ...)` clamp only guards the upper side — nothing pulls the final block out to the edge. Same for `height * rowScale`.

Reproduced against the exported function:

- `803x40 -> 800`: last source column painted red, last output column comes back `R = 0`.
- `642x2398 -> 640`: last source row painted red, last output row comes back `R = 0`.
- Fires on 99 of 4000 `(sourceWidth, rung)` pairs over widths 641-1920. Row drops are commoner, since composite heights are unconstrained while widths are not.

That column is the right edge of the rightmost panel and that row is the bottom of the bottom panel — real capture content, and precisely the "hairline that moved" case this function's docblock says averaging exists to catch. The last block's average is also taken over a short block, so its colour is slightly off even when nothing changed.

```js
const bottom = y === height - 1 ? source.height : Math.min(...);
const right  = x === width  - 1 ? source.width  : Math.min(...);
```

Everything else in the block math checks out: an exhaustive sweep of `sw = 2..200` against `tw` in `{1,2,3,7,13,64,100,199}` found no gaps, no overlaps, no out-of-bounds reads, `sampled` never 0, and aspect ratio preserved to within one source row.

scripts/visual-embed.mjs
gbarkhatov: **Tail-dropping gives up exactly the vault screens `MIN_EMBEDDED_GROUPS_PER_SURFACE` exists to protect.**

The comment above says `pictures` arrives "in the order the comment will read in, so the tail is the end of it". It doesn't. `selectEmbeddedScreens` returns `rankGroups([...chosen])` — **global** rank across surfaces — while `renderBody` iterates surface-major (`for (const surface of surfaces)`), vault above Storybook regardless of rank. The tail is the globally lowest-ranked group, not the end of the comment.

And by this file's own reasoning at `MIN_EMBEDDED_GROUPS_PER_SURFACE`, the globally lowest-ranked groups are the vault ones: "a Storybook story by 17% of its small frame and a vault route by 1% of a tall page". So `kept.pop()` removes the reserved vault pictures first — the ones the floor was written to keep — and they are also the tall, expensive composites, so each drop buys the most bytes and does the most damage. A token repaint publishes a Storybook-only comment.

Drop from the non-reserved pool first, or re-apply the per-surface floor after trimming.

scripts/visual-embed.mjs
```diff
@@ -747,10 +810,72 @@ async function composeComposite(reportDir, surface, name) {
...
+    }
+    // Lowest-ranked first: `pictures` arrives in the order the comment
+    // will read in, so the tail is the end of it.
+    dropped.unshift(kept.pop());
+  }
+
+  // One picture that cannot fit at the floor. Publishing it anyway would
+  // trip the workflow's ceiling and cost the comment every picture, so it
+  // goes the same way as the rest and the text listing carries the run.
+  return { width: PUBLISHED_WIDTH_LADDER.at(-1), published: [], dropped };
```
gbarkhatov: **One oversized top-ranked picture takes every other picture down with it.**

`dropped.unshift(kept.pop())` only ever peels the tail. If the *head* is what can't fit at the floor, the loop discards every lower-ranked picture first — all of which would have fit — then the head, then returns `published: []`.

Reproduced: one heavy 1280x2400 composite plus two costing 2.3KB together, budget 20KB → `published: []`, all three dropped. The comment then falls back to the list of filenames, which is the outcome this budget exists to prevent.

At 448KB this needs a single composite over 448KB at 640px (~14x the observed size), so it isn't routine — but the failure is total and the fix is a line: once `kept.length === 1` and the floor still busts, drop that picture rather than letting it eat the tail on the way there.

scripts/visual-embed.mjs
```diff
@@ -172,6 +174,49 @@ const EMBED_MAX_WIDTH = 1280;
...
+ *
+ * The ladder is walked top down and the first rung the whole set fits under
+ * wins, so the ordinary pull request that moves one screen still publishes
+ * it at the capture's own width - the reduction is what a busy run pays,
+ * not what every run pays. The rungs are sizes worth having rather than a
+ * subdivision: 1280 is a desktop capture at life size, 960 and 800 bracket
+ * what a comment column actually renders, and 640 is the floor, past which
+ * a phone panel in a three-panel composite is 210px wide and the change it
+ * is meant to show stops being legible. A set that will not fit even at the
+ * floor drops its lowest-ranked picture instead of shrinking further, and
```
gbarkhatov: **"and says so in the comment" — it doesn't.**

A byte-budget drop is reported on `process.stderr` and in the stdout summary, both job-log only. `renderBody` has no `dropped` parameter and emits nothing about it. Worse, a surface whose pictures are all dropped disappears entirely via `if (pictured.length === 0) continue;`, taking the "Showing X of Y changed groups" line with it — so the reviewer sees no trace that a surface was omitted. Given the finding above (vault screens go first), that is the likely case.

The PR description makes the same claim ("The comment already reports how many screens it is not showing"). The reduction got an explicit line for exactly this reason; the drop deserves one too.

scripts/visual-embed.mjs
```diff
@@ -172,6 +174,49 @@ const EMBED_MAX_WIDTH = 1280;
...
+ * about it: the caps above bound the count and the dimensions but not the
+ * bytes, so five tall vault composites came to 678KB against a 512KB
+ * ceiling and the step refused the lot. Refusing is the worst outcome
+ * available - the comment falls back to a list of filenames, which is the
+ * state this whole script exists to replace - so the size is settled here,
+ * where a picture can be made smaller, rather than there, where it can only
+ * be dropped.
+ *
+ * Below the workflow's ceiling rather than equal to it, because the two
+ * count slightly different things: this counts the PNGs, that counts what
```
gbarkhatov: **The stated reason for the 448K/512K gap isn't real.**

The two count identically. The workflow sums `find "${EMBED_DIR}" -name '*.png' -printf '%s\n'` — `%s` is apparent byte size, not `du` block-rounding — over exactly the buffers this script writes. The comment body and `diff-output.txt` go to sibling directories, so no non-image bytes land in the measured tree. The workflow's own new comment doesn't make this claim, so the two files now disagree about why the constants differ.

The 64KB of headroom is worth keeping — just state the real reason. Worth pairing with two things: the PR's own evidence clears the budget by 2KB (446 against 448, 0.45%), and the new test only asserts `ceiling > budget`, which a 1-byte margin satisfies. If the headroom is deliberate, pin it (`ceiling >= budget + MARGIN`).

scripts/visual-embed.mjs
```diff
@@ -172,6 +174,49 @@ const EMBED_MAX_WIDTH = 1280;
...
+ * it at the capture's own width - the reduction is what a busy run pays,
+ * not what every run pays. The rungs are sizes worth having rather than a
+ * subdivision: 1280 is a desktop capture at life size, 960 and 800 bracket
+ * what a comment column actually renders, and 640 is the floor, past which
+ * a phone panel in a three-panel composite is 210px wide and the change it
+ * is meant to show stops being legible. A set that will not fit even at the
+ * floor drops its lowest-ranked picture instead of shrinking further, and
+ * says so in the comment.
+ */
+const PUBLISHED_TOTAL_BYTE_BUDGET = 448 * 1024;
```
gbarkhatov, line 231: **The ladder has less room above the legibility floor than it looks.**

Byte savings run well below the pixel cut, because box-averaging at fractional scales turns flat UI regions into gradients that deflate worse. Your own numbers: 1280 → 960 cut 44% of the pixels and 27% of the bytes. Measured on a UI-like fixture here (flat background, hairlines, text runs) it's weaker still, and not even monotonic — 1280px 32,565B, 960px 26,441B, **800px 38,746B**, 640px 25,634B. The rung choice is still correct (it picks on measured bytes), but 1280 → 640 buys only ~28% for a 75% pixel cut.

Scaling your run to the cap of `MAX_EMBEDDED_GROUPS * MAX_EMBEDDED_SCREENS_PER_GROUP` = 10 pictures: ~1216KB at 1280, ~890KB at 960, and extrapolating your observed bytes-vs-area relationship, ~570-710KB at the 800 and 640 rungs. A vault-heavy busy run — the "repaint of a shared component routinely changes 40+ screens" case named above — plausibly lands on 640, the rung this comment describes as the point where "the change it is meant to show stops being legible", or falls through to the drop path entirely.

Not a blocker, and the whole-set-one-rung argument is well made. But the policy is currently silent on trading count against width: ten pictures at the legibility floor may be worth less than seven at 800px. Worth saying which way it should go, since the drop path sacrifices vault screens first.

scripts/visual-embed.mjs
```diff
@@ -172,6 +174,49 @@ const EMBED_MAX_WIDTH = 1280;
...
+const PUBLISHED_WIDTH_LADDER = [EMBED_MAX_WIDTH, 960, 800, 640];
+
+/**
+ * Published as truecolour with no alpha channel.
+ *
+ * Every source here is an opaque screenshot and the reduction below writes
```
gbarkhatov: The justification doesn't cover the path every ordinary run takes. `reduceToWidth` returns `source` untouched when the composite is already within the rung, so on the top rung nothing "writes 255" — alpha there comes from `fill`/`paintRect` and from the captured PNGs via `blit`.

The invariant does hold (`fill` writes 255, and `capture.ts` doesn't set `omitBackground`, so every screenshot is opaque), so this is a wording fix, not a bug. One latent consequence worth a line though: if a capture ever did carry real transparency, the passthrough path would blend against pngjs's default white background while the reduced path ignores alpha and keeps raw RGB — the same picture changing colour depending on which rung won.

scripts/visual-embed.mjs
```diff
@@ -747,10 +810,72 @@ async function composeComposite(reportDir, surface, name) {
...
+ * Encodings are cached per picture and rung. The walk revisits both, and
+ * deflating a 1280x2400 composite is ~100ms - enough for a rung-by-rung
+ * search on a full set to become the slowest thing in the job.
+ */
+function fitPublishedImages(pictures, budget = PUBLISHED_TOTAL_BYTE_BUDGET) {
+  const encoded = new Map();
+  const bytesAt = (picture, width) => {
+    const key = `${picture.index}:${width}`;
+    if (!encoded.has(key)) {
+      encoded.set(key, encodeAtWidth(picture.composite, width));
```
gbarkhatov: Two cheap wins in this cache, both from the same cause — `reduceToWidth` returns `source` unchanged when it's already narrow enough:

1. **Duplicate encodes.** A 788px mobile composite (390*2 + gutter) is byte-identical at rungs 1280, 960 and 800, and gets three separate ~1MB cache entries and three deflates. `Math.min(width, picture.composite.width)` in the key collapses them. It also means the top three rungs buy a mobile-heavy set literally zero bytes, so it reaches the drop path much sooner than a four-rung ladder suggests.
2. **No early exit.** `kept.map(...)` encodes the whole set at a rung before summing, so a rung that is already over budget by picture 2 still pays for pictures 3..N. Accumulate and bail once the running total exceeds the budget.

Also worth knowing: `main()` now holds every raw composite alive through the ladder walk where the pre-PR code encoded and wrote one at a time — ~123MB of RGBA at the 10-composite cap, plus up to 40 cached buffers. Bounded and fine on a runner, but the docblock's "~100ms per deflate" undersells the change. Dropping `composite` from `published` entries once `bytes` exist would cover it.

scripts/visual-embed.mjs
```diff
@@ -1103,49 +1236,83 @@ async function main() {
...
-    coverage.set(key, picture.coverage);
-    composed.push(entry);
+    coverage.set(picture.key, picture.coverage);
   }
 
+  // Only when a picture actually lost pixels. The top rung is the width the
+  // composites are already built at, so on the ordinary run nothing is
+  // reduced and the comment should not claim otherwise.
+  const wasReduced = published.some(
+    (picture) => picture.composite.width > width,
```
gbarkhatov: `wasReduced` is set-wide, but the rung is a width **cap**, not a scale factor — `reduceToWidth` returns the source untouched when `source.width <= width`. Captures are 1280 desktop and 390 mobile, so a mobile three-panel composite is 1186 wide and a two-panel one is 788.

Published at rung 800, a 1280 desktop composite lands at 800 (0.63x) while the 788 mobile one is untouched at 788 (1x) — and the comment tells the reviewer both are at 800. That's the mis-measurement the sentence was added to prevent, pointed the other way. Phrasing it as a cap ("published at most Npx wide") would be true in both cases.

scripts/__tests__/visual-embed.test.mjs
```diff
@@ -296,3 +304,164 @@ test("selectEmbeddedGroups under-serves the last surface once a third is added",
...
+
+// The rule the fractional block exists for. At an integer factor every
+// output column samples the same number of source columns and a lone
+// changed column always lands in one of them; at 1.6 it does not, and a
+// nearest-neighbour reduction would skip it on some columns.
+test("reduceToWidth carries a one-column change into a fractional reduction", () => {
+  const composite = new PNG({ width: 10, height: 1 });
+  composite.data.fill(0);
+  for (let i = 3; i < composite.data.length; i += 4) composite.data[i] = 255;
+  composite.data[4 * 4] = 255;
```
gbarkhatov: **This test passes under the regression it exists to catch.**

Column 4 doesn't discriminate. At scale 10/7 nearest-neighbour samples `Math.floor(3 * 1.4286) = 4`, so a pixel-dropping implementation produces the identical `lit === [3]`.

I ran a nearest-neighbour reference against `reduceToWidth` for all ten source columns — only **3, 6, and 9** discriminate (averaged → `[2]`, `[4]`, `[6]`; nearest-neighbour → `[]`). Column 4 is one of the seven that don't.

One-character fix: light column 3 and expect `[2]` (the averaged value there is 128).

scripts/__tests__/visual-embed.test.mjs
```diff
@@ -296,3 +304,164 @@ test("selectEmbeddedGroups under-serves the last surface once a third is added",
...
+  const pictures = [
+    { key: "vault/a.png", composite: textured(1280, 160) },
+    { key: "vault/b.png", composite: textured(1280, 160) },
+  ];
+  const budget =
+    cost(pictures[0].composite, 640) + cost(pictures[1].composite, 640) - 1;
+
+  const fit = fitPublishedImages(pictures, budget);
+
+  assert.deepEqual(
```
gbarkhatov, line 429: Three gaps worth closing while this file is open:

1. **`fit.width` is never asserted.** The re-walk-from-the-top behaviour is called out as deliberate in the docblock, and this test does exercise it — with this budget the survivor publishes at 1280x160, not at the 640 floor — but nothing observes it. An implementation that resumed the ladder at the rung it failed on passes every test here. `assert.equal(fit.width, 1280)` closes it, with plenty of margin.
2. **Aspect ratio is never asserted.** `height` derived from `columnScale` is the only thing keeping a reduced composite from being squashed, and no test reads an output height — the fractional test uses a 1-row source, and the fit tests read `.width` only. A regression that scaled width but not height still writes a valid PNG and still fits the budget.
3. **`cost()` hardcodes `{ colorType: 2 }`** instead of importing `PUBLISHED_COLOUR_TYPE`. Every budget-derived test then brackets an encoding that only coincidentally matches production — change the constant to 6 and these budgets stop bracketing what `fitPublishedImages` actually emits. (The literal `2` in the colour-type assertion is fine as-is; that one is the expectation.)

scripts/__tests__/visual-embed.test.mjs
```diff
@@ -296,3 +304,164 @@ test("selectEmbeddedGroups under-serves the last surface once a third is added",
...
+  const workflow = fs.readFileSync(
+    path.resolve(
+      fileURLToPath(import.meta.url),
+      "../../../.github/workflows/visual-regression.yml",
+    ),
+    "utf8",
+  );
+
+  const ceiling = workflow.match(
+    /MAX_PUBLISHED_BYTES=\$\(\(\s*(\d+)\s*\*\s*1024\s*\)\)/,
```
gbarkhatov: This is brittle in a way that costs more than it should. `MAX_PUBLISHED_BYTES=524288` — a reasonable simplification — produces no match, `assert.ok(ceiling, ...)` fails, and `Test the report scripts` is deliberately not `continue-on-error`, so the whole `visual` job dies and every PR loses its visual comment until someone notices the grep.

It fails loud rather than silent, which is the right side to err on, and I checked that operand-swapped forms like `$(( 1024 * 512 ))` don't match either. But matching `MAX_PUBLISHED_BYTES=(.+)` and evaluating, or having both sides read one shared source, would pin the value rather than one spelling of it.

gbarkhatov (dismissed):
Reviewed at 43bd7664f (3 files). The diagnosis is right and the fix belongs where this PR puts it — the size decision should sit where a picture can be made smaller, not where it can only be refused.

I verified the parts that could have been silently wrong. The budget-vs-ceiling accounting genuinely holds: the workflow sums `%s` over exactly the files this script writes, no non-image bytes land in the measured tree. `coverage` and `embeddedNames` are both built from the published set, so no path can render a URL for a file nobody wrote. The zero-published case degrades safely — publish exits 1 under `continue-on-error` and the comment gates on `PUBLISH_OUTCOME == 'success'`, so no 404 image URLs are ever posted. The encode cache key stays unique across drops, the loop terminates, `dropped` keeps rank order, and truecolour is safe on both paths (`capture.ts` doesn't set `omitBackground`). Nothing in the workflow parses this script's output. No test assertion depends on zlib output being stable across Node versions, byte 25 is PNG-spec-fixed, and `visual:test` really is gated in CI. 40/40 pass. No `downscale` leftovers.

Two behavioural findings are worth a fix before this lands, both inline:

- `reduceToWidth` silently drops the last source column and the last source row to a floating-point edge in the block bounds. Reproduced on `803x40 -> 800` and `642x2398 -> 640`; fires on 99 of 4000 (width, rung) pairs. That is the outermost edge of the rightmost panel — exactly the hairline case the averaging exists to catch.
- Tail-dropping gives up the reserved vault screens first. `composed` is in global rank order, and by this file's own reasoning vault routes rank *below* Storybook stories (1% of a tall page vs 17% of a small frame), so `kept.pop()` removes what `MIN_EMBEDDED_GROUPS_PER_SURFACE` was written to protect — and does it without saying so in the comment.

Both want a test alongside the fix. Approving so this isn't blocked on me; the rest is comment accuracy and robustness.

### Also, outside the diff

- `panelLayout` (~L649-650) still argues from the integer reduction this PR deleted: "the integer reduction that fits 1280 renders each screen 428px wide - a third of life size". Under `reduceToWidth` it's ~639px, half life size. The conclusion survives, the arithmetic doesn't — and this paragraph is the whole stated rationale for the stacking threshold. Same stale figure in the test comment at ~L107-108 of `visual-embed.test.mjs`.
- `EMBED_MAX_WIDTH`'s docblock (~L138-154) still calls it "Width the composite is reduced to before it is published"; it's now rung 0 of the ladder and a no-op on every ordinary run. Its argument against publishing at "800-1000" is also now exactly what a busy run does, with no pointer forward to `PUBLISHED_WIDTH_LADDER`.
- `MAX_EMBEDDED_GROUPS` (~L52-54) still says there is deliberately no ceiling on the total, "a third cap could only bite by quietly giving the lowest-ranked groups fewer pictures than the promise above". The byte budget is that third cap, and that is exactly how it bites.
- `screenCaption` (~L908-911) reports `(picture trimmed to Xpx of Ypx)` from source rows. Those numbers used to match the published image exactly, because `downscale(composite, EMBED_MAX_WIDTH)` was a no-op for every real capture — `panelLayout` stacks anything wider than 1280. On a reduced run the caption now reads "trimmed to 1800px of 3000px" over a picture 900px tall. Same hazard the new "reduced to Npx" note was added to close, and it needs the same treatment.

scripts/visual-embed.mjs
```diff
@@ -1103,49 +1236,83 @@ async function main() {
     // slots on one fact is the thing the caps exist to prevent. Note this
     // frees the slot rather than reassigning it: selection has already run,
     // so the comment shows fewer pictures, not a different one.
-    const digest = createHash("sha256").update(picture.bytes).digest("hex");
+    // Hashed on the pixels rather than on an encoding, so a twin is caught
+    // before the width the set is published at has been decided - and stays
+    // caught whatever that width turns out to be.
+    const digest = createHash("sha256")
+      .update(picture.composite.data)
+      .digest("hex");
```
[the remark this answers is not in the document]
jonybur: Fixed. The digest now hashes `${width}x${height}:` before the pixel buffer, so two composites with the same pixel count in a different shape are no longer treated as twins.

scripts/visual-embed.mjs
jonybur: You are right that the premise in the docblock was wrong, and the docblock now says so: every rung is measured rather than assumed, and it cites your 1186px case. The walk itself was already choosing on measured bytes and takes the widest rung that fits, so a rung costing more than a wider one can never be selected - what was broken was the claim, and the message. The comment now says "published at most Npx wide" and names what was given up, so a run that reduces and still drops no longer reports only the reduction.

scripts/visual-embed.mjs
jonybur: Fixed. Dropping now takes from whichever surface is holding the most pictures, and the lowest-ranked within it, so both surfaces stay pictured until one has nothing left. New test: five pictures, three Storybook and two vault, budget forcing two drops - storybook/s3 and vault/v2 go, and vault keeps a picture. Reverting the fix fails it.

scripts/visual-embed.mjs
jonybur: Fixed. A picture that throws while encoding is now given up like a compose failure, with its reason on stderr, and the rest of the set still publishes. Test uses a composite with a null data buffer; making the throw fatal again fails it.

scripts/visual-embed.mjs
jonybur: Fixed. A picture that cannot fit the budget on its own is given up before the ladder walk starts, so it can no longer evict the pictures that would have fit on its way to being dropped itself. Your case - one heavy composite plus two cheap ones - now publishes the two.

scripts/visual-embed.mjs
jonybur: Fixed, and thank you for the reproduction. The last block now runs to the source edge on both axes. Both your cases come back lit (128 rather than 0), and a sweep of 1280 widths across the four rungs finds no remaining drops. Two tests cover it, at 803x40 -> 800 and 642x200 -> 640 - I checked that those sizes actually discriminate, since 642x400 passes either way.

scripts/visual-embed.mjs
jonybur: Fixed. The drop is chosen from the surface holding the most pictures rather than from the tail, so the per-surface floor survives trimming. The docblock now records why, and a test asserts vault keeps a picture on a Storybook-heavy set.

scripts/visual-embed.mjs
jonybur: Fixed as you suggested, one step earlier: a picture whose floor-rung encoding is larger than the whole budget is dropped before the walk begins. Your 1280x2400-plus-two-cheap case at a 20KB budget now publishes both cheap pictures instead of nothing.

scripts/visual-embed.mjs
jonybur: Fixed. `renderBody` takes a `droppedCount` and the footer now names how many pictures were left out and where to find them. A surface whose pictures were all dropped no longer disappears either: only a surface with no changed groups at all is skipped, so the heading and its "Showing 0 of N changed groups" line survive.

scripts/visual-embed.mjs
jonybur: Fixed. The docblock no longer claims the two count different things - they count the same bytes - and states the gap as a deliberate margin, citing the 2KB the measured run cleared it by. The margin is now a named constant and the test asserts `ceiling >= budget + margin` rather than just the ordering.

scripts/visual-embed.mjs
jonybur: Taken. The docblock now gives the measured figure rather than the pixel cut (1280 to 640 is a 75% pixel cut for about 28% of the bytes) and states which way the trade goes when the floor still busts: keep both surfaces pictured rather than keep the count up, on the same reasoning as MIN_EMBEDDED_GROUPS_PER_SURFACE. The drop policy now implements exactly that.

scripts/visual-embed.mjs
jonybur: Fixed as a wording change. It now says what actually keeps the channel constant - `fill` and `paintRect` write 255 and the captures carry no transparency - and notes that the reduction is not what makes it safe on the ordinary run, since a composite already inside the rung is published untouched.

scripts/visual-embed.mjs
jonybur: All three taken. The cache key is now `Math.min(width, composite.width)`, so the top rungs share one entry for a narrow composite. The total accumulates and bails as soon as it passes the budget, so a rung already over by picture 2 does not encode the rest. And `published` entries no longer carry `composite` - they carry the width the picture was actually published at, which is what the caller needed it for.

scripts/visual-embed.mjs
jonybur: Fixed. Each published picture now reports its own `publishedWidth`, the rung is described as a cap, and the comment reads "published at most Npx wide". `reduced` is computed from whether any composite was actually wider than the rung, so a set of narrow composites reports no reduction at all. Two tests cover both directions.

scripts/__tests__/visual-embed.test.mjs
jonybur: Fixed - column 3, expecting `[2]`, and asserting the averaged value is 128. I checked it discriminates: replacing the box average with nearest-neighbour now fails three tests, where before it passed.

scripts/__tests__/visual-embed.test.mjs
jonybur: All three closed. `fit.width` is asserted at 1280 on the drop test, so an implementation resuming the ladder at the rung it failed on fails. A new test asserts the reduced shape (1280x800 -> 640x400), which nothing observed before. And `cost()` now imports `PUBLISHED_COLOUR_TYPE` instead of hardcoding 2.

scripts/__tests__/visual-embed.test.mjs
jonybur: Fixed. It now matches `MAX_PUBLISHED_BYTES=$((...))`, checks the contents are plain arithmetic and evaluates them, so `1024 * 512` reads the same as `512 * 1024`. It also asserts the margin rather than just the ordering, so a ceiling one byte above the budget no longer passes.

gbarkhatov (dismissed):

scripts/__tests__/visual-embed.test.mjs
```diff
@@ -296,3 +306,342 @@ test("selectEmbeddedGroups under-serves the last surface once a third is added",
...
+
+  assert.ok(assignment, "MAX_PUBLISHED_BYTES not found in the workflow");
+  assert.match(
+    assignment[1],
+    /^[\d\s*+]+$/,
+    `MAX_PUBLISHED_BYTES is not a plain arithmetic expression: ${assignment[1]}`,
+  );
+  const ceiling = Number(
+    new Function(`return (${assignment[1]});`)(),
+  );
```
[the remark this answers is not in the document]
jonybur: Taken, in a slightly wider form. You're right that the "not found" message is wrong when the constant is plainly there, and that's the part worth fixing — it sends the next person hunting for a deleted constant instead of at a spelling they can see.

The suggested pattern closes one of the spellings that hard-fail, though. Measured against the current regex:

| spelling | today |
|---|---|
| `$(( 512 * 1024 ))`, `$((512*1024))`, `$(( 1024 * 512 ))` | pass |
| `=524288` | "not found" |
| `="524288"`, `="$(( 512 * 1024 ))"` | "not found" |
| `$(( 512 << 10 ))`, `$(( KB * 512 ))` | clear failure, names the expression |

Landed an anchored version that takes the bare literal and tolerates a trailing `#` comment:

```js
const assignment = workflow.match(
  /^[ \t]*MAX_PUBLISHED_BYTES=(?:\$\(\((.+?)\)\)|(\d+))[ \t]*(?:#.*)?$/m,
);
const expression = assignment[1] ?? assignment[2];
```

Anchoring matters with the `(\d+)` branch added: unanchored, a prose line in the 15-line comment block above the constant (`# raise MAX_PUBLISHED_BYTES=1048576 if ...`) would be read as the assignment. Verified it still picks the real line when that prose is present.

Both later `assignment[1]` references moved to `expression` — leaving either would throw a TypeError on the literal branch instead of asserting. `512 << 10` and variable references still hard-fail with the clear message, which is right for a guard that cannot read its input.

jonybur:
@gbarkhatov — the four "outside the diff" items from your review are all fixed in `68c8df764`.

**`panelLayout`.** You were right that the arithmetic did not survive. I re-derived it rather than taking either number: two 1280px panels plus the 8px gutter come to 2568px, `reduceToWidth` fits that to the 1280 rung at a factor of 2.00625, and 639 published columns carry each panel — half life size, not a third. The docblock now says that, and notes the stacked case publishes untouched at the top rung so a busy run reduces from life size rather than on top of a halving already taken. Same figure corrected in the test comment.

**`EMBED_MAX_WIDTH`.** Rewritten as what it now is — rung 0 of the ladder, a no-op on an ordinary run because `panelLayout` already keeps every composite inside it. The argument against publishing at 800-1000 now points at 960 and 800 as rungs the ladder walks down to, with the 856px figure dropped since the ladder can no longer produce it.

**`MAX_EMBEDDED_GROUPS`.** The "deliberately no third cap" paragraph is gone. It now says the byte budget is that cap and that it bites in the right order: width first, so a busy run keeps every picture the count caps promised, and only once the floor rung is still over budget does it give pictures up — from whichever surface holds the most, not from the bottom of the ranking.

**`screenCaption`.** This one was the user-facing half, as you flagged. It now reads `(shows Xpx of a Ypx screen)` rather than `(picture trimmed to Xpx of Ypx)`, so a reduced run no longer prints "trimmed to 1800px" over a 900px-tall picture. The comment says the count is in capture rows and why.

Also took a CodeRabbit note on the cross-file ceiling guard: it now accepts a plain byte count as well as the `$(( ))` form, and is anchored to an assignment line — with the literal branch added, an unanchored pattern would read `# raise MAX_PUBLISHED_BYTES=1048576 if ...` out of the comment block above the constant.

Your two inline findings I confirmed as genuinely fixed by `4398cedf2` before touching anything — `reduceToWidth carries the last source column/row into the reduction` and `fitPublishedImages keeps every surface pictured when it has to give one up` both fail if the fix is reverted. 48 tests green.

gbarkhatov (approved):

Issue #2209, opened 2026-08-10
Release pipeline publishes stale intra-workspace dependency versions (ts-sdk@0.58.1 → babylon-tbv-rust-wasm@0.1.0)


gbarkhatov:
## Summary

Every package we publish that depends on a sibling workspace package via `workspace:*` can ship a **stale, wrong, or non-existent** version of that sibling. The published manifest is only correct when the sibling happened to be released in the *same* CI run.

Concretely, `@babylonlabs-io/ts-sdk@0.58.1` (current `latest`) declares:

```json
"dependencies": {
  "@babylonlabs-io/babylon-tbv-rust-wasm": "0.1.0"
}
```

`0.1.0` is the first release of that package ever made. It exports 4 functions. `ts-sdk@0.58.1` imports **20 symbols** from it, including `computePayoutFeeFloor`, `expandAuthAnchor`, `expandHashlockSecret`, `expandWotsSeed`, `tapInternalPubkey`, `WasmPrePeginTx`, `createPayoutConnector`. A clean install of `ts-sdk@0.58.1` therefore resolves an exact-pinned `0.1.0` and fails on import.

This is not a one-off. It has been happening since the packages were split out, and it affects two package pairs.

## Impact

- **External npm consumers only.** In-repo consumers (`services/vault`) use `workspace:*` and resolve to the local build, which is why nothing is broken in this repo or in CI. Every downstream repo installing from npm is exposed.
- **Failure mode for `ts-sdk` → `rust-wasm` is loud**, not silent: `0.1.0` lacks the symbols entirely, so it is an import/resolution error rather than wrong arithmetic. That is the good news given the critical-path rules in `CLAUDE.md` — the vault-secret expanders (`expandAuthAnchor` / `expandHashlockSecret` / `expandWotsSeed`) simply do not exist in `0.1.0`, so there is no path where a consumer silently derives secrets from a stale WASM build. **But that is luck, not design:** the same mechanism would happily pin a *stale-but-existing* version (e.g. `0.14.0`), and that failure would be silent and would sit directly on a frozen on-chain-binding API.
- **Failure mode for `wallet-connector` → `core-ui` is a hard install failure**: `0.0.0-semantic-release` is not a published version, so `pnpm install` / `npm install` cannot resolve it at all.
[...]

PR #2273, opened 2026-08-18
fix(release): publish the real version of each sibling package

jonybur:
## What

Packages published from this repo now carry the correct version of every other package in this repo they depend on, and a release stops before publishing if any of those versions is wrong.

## Why

`@babylonlabs-io/wallet-connector` cannot be installed from npm right now — its latest release points at a version of the shared UI package that was never published. `@babylonlabs-io/ts-sdk` installs, but points at a WASM build missing 14 of the 20 functions it calls. 101 of 120 published `ts-sdk` versions and 65 of 237 `wallet-connector` versions are affected, and both of our internal consumers already carry a manual `overrides` workaround.

In the repo these dependencies are written as `workspace:*`, meaning "whatever version is here". Something has to turn that into a real number at publish time. nx does it correctly only when both packages happen to be released in the same run; otherwise pnpm fills it in from the sibling's `package.json` on disk, and those numbers have been frozen fiction ever since the release stopped committing version bumps back to git.

There is a second cause. A package marked `private` still gets a version, a git tag, and a pin from its public dependent, but nx never attaches a publish target to it and drops it from the run while still reporting success. That is how `wallet-connector@1.66.1` came to point at a signer version that exists only as a git tag, and which nx can never publish because it only rolls forward.

## How

- The release writes each sibling's real version into the manifest before publishing, so what sits on disk is exactly what ships.
- It refuses to publish if any sibling version is stale, a placeholder, still a `workspace:` protocol, a wildcard, or absent from npm. The check is equality against the version this run resolved, not just "some version exists".
- It publishes first and tags afterwards, so a failed publish no longer leaves a tag that permanently blocks retrying that version.
- It checks that every package that was supposed to publish actually did, instead of only checking exit codes — a package dropped before the executor ran previously passed.
- It rejects a public package that depends on a private workspace package.
- Release candidates go through the same code. They used `npm publish`, which cannot resolve `workspace:` at all, so every RC ever published shipped a literal `workspace:*`.
- `releaseTagPatternStrictPreid` is enabled. Without it, the first RC tag above the stable line would make every later stable release publish a prerelease to `latest`.
- `ts-sdk` and `wallet-connector` get their real version written into `package.json` — both a correction and the change that makes nx cut the corrected releases on merge.
- `tools/release` gains tests, lint and typecheck. nx picks the targets up from the package scripts, so no workflow change was needed for them.

Merging this publishes corrected `ts-sdk` and `wallet-connector` releases. Deprecating the versions already on npm is a manual follow-up: `npm deprecate` cannot use this repo's OIDC trusted publishing, so it needs someone with an interactive npm login on the `@babylonlabs-io` scope.

Refs #2209


tools/release/src/pinManifests.ts
```diff
@@ -0,0 +1,273 @@
...
+      const pinned = pinSiblingDependencies(manifest, siblings);
+      const violations = findSiblingPinViolations(pinned.manifest, siblings);
+      if (violations.length > 0) {
+        throw new Error(
+          formatViolations(packageName, releasePackage.manifestPath, violations)
+        );
+      }
+
+      for (const sibling of siblings.values()) {
+        assertSiblingIsInstallable(
```
jrwbabylonlab, line 121: core-ui 1.109.0 has a merged tag but was never published to npm (latest is 1.108.1). wallet-connector pins it via `workspace:*`, so this throws and the run aborts before publish. blocks this release and every wallet-connector one after it.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  assertEveryProjectPublished(projectsToPublish, publishResults, DRY_RUN);
+
+  /**
+   * Tagging runs only after a fully successful publish. The other order leaves a
+   * tag for a version that was never published, and because nx only ever rolls
+   * forward from tags that version can never be published afterwards - which is
+   * how `wallet-connector@1.66.1` came to pin a permanent 404.
+   */
+  await releaseChangelog({
```
jrwbabylonlab: `releasePublish` ships everything nx bumped, but `releaseChangelog` is filtered to `RELEASE_PROJECTS`. so a dependent pulled in by `updateDependents` gets published to npm with no git tag.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  try {
+    await execCommand('git', commandArgs);
+  } catch (error) {
+    /**
+     * Tags for a GitHub release already reach the remote through the releases
+     * API, so this push is usually a no-op. Everything of value has been
+     * published by this point, and failing here would report a successful
+     * release as broken.
+     */
+    console.warn(`Could not push to the remote: ${error}`);
```
jrwbabylonlab, line 180: on the RC path `createRelease` is false, so nothing else creates the tag. if this push fails the RC is on npm untagged, and the next run recomputes the same version and silently no-ops. this used to fail CI.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  const summary = `These projects were versioned but did not publish successfully: ${failed.join(', ')}.`;
+
+  /**
+   * A dry run never writes the new version to disk, so `pnpm publish
+   * --dry-run` reads the previous one and reports it as already published.
+   * That says nothing about whether a real release would work.
+   */
+  if (dryRun) {
+    console.warn(`${summary} Expected during a dry run, which does not write the new versions to disk.`);
```
jrwbabylonlab: a private package never shows up in `publishResults` at all, which is the bug this PR is fixing. dry run warns past that too, so it reports clean for a release that will drop a package.

tools/release/src/manifest.ts
```diff
@@ -0,0 +1,222 @@
...
+): void => {
+  if (sibling.publishedByThisRun) return;
+
+  if (publishedVersions === null) {
+    throw new Error(
+      `${sibling.packageName} is pinned at ${sibling.version} but has never been published. Publish it before publishing anything that depends on it.`
+    );
+  }
+
+  if (!publishedVersions.includes(sibling.version)) {
```
gbarkhatov, line 295: **Blocking — this gate hard-fails the first release after merge, and takes every unrelated package down with it.**

`@babylonlabs-io/core-ui/1.109.0` is a git tag on `origin/main` (`49110ecf3`) that was never published: `npm view @babylonlabs-io/core-ui@1.109.0` is a 404 and `latest` is `1.108.1`. Nothing has touched `packages/babylon-core-ui/` since.

On any run where `wallet-connector` is bumped and `core-ui` is not — which includes this PR's own merge, since it edits `packages/babylon-wallet-connector/package.json` under a `fix:` type — `resolveSibling` returns `core-ui@1.109.0` with `publishedByThisRun: false`, the registry has no such version, and this throws. That happens inside `materializeAndPinManifests`, *before* `releasePublish` is ever called, so `ts-sdk`'s corrected release does not happen either. The PR description's "Merging this publishes corrected ts-sdk and wallet-connector releases" is not true against the current registry state.

The guard is behaving exactly as designed — it is catching pre-existing damage this PR doesn't repair. The error's own remedy ("Re-run the release for `@babylonlabs-io/core-ui`") is the one thing that cannot work, because nx only rolls forward from tags and will never re-attempt `1.109.0`.

Cheapest fix, and symmetric with what was already done for the other two: write `1.109.0` into `packages/babylon-core-ui/package.json`. The commit then touches `core-ui`, nx bumps it to `1.109.1`, it co-releases, `publishedByThisRun` is true, and the gate passes. Alternatively hand-publish `1.109.0`, or delete the tag.

Same orphan-tag state exists at `babylon-proto-ts/1.22.2` and `/1.23.0`, `wallet-connector/1.68.0`, and `ledger-vault-signer/0.2.0` and `/0.3.0` — none of those block today only because the *newest* tag of each is either published or has no publishable dependent. Worth a one-off reconciliation pass, since the gate is workspace-global: one orphan tag anywhere blocks every unrelated package's release.

Also note `__tests__/manifest.test.ts:228` uses this exact version but models it as `publishedByThisRun: true`, which is not what a `wallet-connector`-only release does — so the test reads as if this case were covered when it isn't.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  /**
+   * Tagging runs only after a fully successful publish. The other order leaves a
+   * tag for a version that was never published, and because nx only ever rolls
+   * forward from tags that version can never be published afterwards - which is
+   * how `wallet-connector@1.66.1` came to pin a permanent 404.
+   */
+  await releaseChangelog({
+    // Must match the filter releaseVersion ran with: releaseChangelog reads a
+    // version for every project it is asked about and throws on a gap.
+    projects: RELEASE_PROJECTS.length > 0 ? RELEASE_PROJECTS : undefined,
```
gbarkhatov: **Blocking — `releasePublish` is unfiltered but `releaseChangelog` is filtered, so an RC dispatch publishes a dependent with no git tag.**

`version.updateDependents` is left at nx's default `auto`, and `findDependentsToProcess` in `release-group-processor.js` adds out-of-filter dependents to `allProjectsToProcess` and bumps them `patch` (`DEPENDENCY_WAS_BUMPED`). Reproduced:

```
$ nx release version --projects=@babylonlabs-io/babylon-tbv-rust-wasm --specifier=prerelease --preid=rc --dry-run
@babylonlabs-io/babylon-tbv-rust-wasm  ... prerelease -> 0.15.1-rc.0
@babylonlabs-io/ts-sdk                 ... patch, because a dependency was bumped -> 0.62.2
```

`projectsToPublish` (line 48-50) picks up both, so `releasePublish` ships `ts-sdk` too — to the `next` dist-tag, with a **stable-looking** version number, pinning a prerelease WASM. But this line filters `releaseChangelog` to `RELEASE_PROJECTS`, and `createGitTagValues` only iterates the filtered set, so `ts-sdk` lands on npm with no git tag and no changelog.

The consequence is the bad part: the next stable run on `main` resolves `ts-sdk` from its tags, recomputes that same version, and the publish executor sees it already on the registry and runs `npm dist-tag add … latest` returning `success: true` (`release-publish.impl.js:128-146`). So `latest` silently moves onto an RC-branch build that pins a prerelease WASM, and the release reports green while shipping nothing. That is a fresh instance of the exact bug #2209 is about, created by this PR's own RC path.

The old RC path published a single package directory, so this is a regression rather than a pre-existing hole.

Two ways out: pass `projects: projectsToPublish` here — it's a superset of `RELEASE_PROJECTS` and every entry has version data, so the `createGitTagValues` gap the comment above guards against still can't fire — or force `updateDependents: 'never'` on the RC path so the dependent is never bumped in the first place. The first is probably the safer default for both paths.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  await materializeAndPinManifests({
+    projectsToPublish,
+    projectsVersionData,
+    releasePackages,
+    releaseConfig,
+    registry: createRegistryClient(releaseConfig.registryUrl),
+    dryRun: DRY_RUN,
+  });
+
+  const publishResults = await releasePublish({
```
gbarkhatov, line 107: A partial publish can still ship the exact broken pin this PR exists to prevent, immutably.

`assertSiblingIsInstallable` returns early for any sibling with `publishedByThisRun: true` — it has to, the version can't be on the registry yet. But `releasePublish` runs `nx-release-publish` as independent tasks with no `dependsOn`, so there's no guarantee a dependency publishes before its dependent, and no rollback if it doesn't.

Concretely: a run bumps both `babylon-tbv-rust-wasm` and `ts-sdk`. `ts-sdk` publishes first, pinning `babylon-tbv-rust-wasm@0.16.0`. The WASM publish then fails on a registry 5xx. `assertEveryProjectPublished` correctly exits 1 — but `ts-sdk@x` is already immutable on npm pointing at a version that does not exist, and nx will never retry `0.16.0`.

Worth either publishing in topological order (dependencies first, bail on the first failure) or re-verifying every `publishedByThisRun` sibling actually landed before publishing its dependents.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+ * nx does that silently for any package without a publish target, which is every
+ * package marked private. Checking only the exit codes passes vacuously then.
+ */
+const assertEveryProjectPublished = (
+  projectsToPublish: readonly string[],
+  publishResults: PublishResults,
+  dryRun: boolean
+): void => {
+  const failed = projectsToPublish.filter(
+    (projectName) => publishResults[projectName]?.code !== 0
```
gbarkhatov: `code !== 0` can't distinguish "published this run" from "was already there".

When the version is already on the registry the publish executor short-circuits — `distTags[tag] === currentVersion` returns `success: true` outright, and otherwise it runs `npm dist-tag add` and returns `success: true` (`release-publish.impl.js:118-148`). So a run that publishes nothing new passes this check and reports green.

That's benign on a retry, which is the case it was presumably written for, but it means this assertion can't tell a successful release from a no-op — which matters for the RC path, where a lost tag makes every subsequent dispatch a silent no-op (see the `pushReleaseTags` comment below).

Separately: `assertEveryProjectPublished` can't be tested at all. `release()` is invoked at module scope on line 172, the function isn't exported, `process.exit(1)` is inline, and `DRY_RUN` / `RELEASE_*` are frozen into module constants at import time. It's the guard for one of the four headline fixes — the package nx silently drops for having no publish target — and nothing can exercise it. Extracting it into its own module that throws, and guarding the `release()` call, would make it testable without changing behaviour.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (failed.length === 0) return;
+
+  const summary = `These projects were versioned but did not publish successfully: ${failed.join(', ')}.`;
+
+  /**
+   * A dry run never writes the new version to disk, so `pnpm publish
+   * --dry-run` reads the previous one and reports it as already published.
+   * That says nothing about whether a real release would work.
+   */
+  if (dryRun) {
```
gbarkhatov: This branch means a dry run cannot fail, and the comment above it describes something nx doesn't do.

The executor skips the `npm view` check *entirely* under dry run — `release-publish.impl.js:110` is `if (!isDryRun && !options.firstRelease)` — and goes straight to `pnpm publish --dry-run`. So it never "reads the previous one and reports it as already published". A non-zero exit here is a genuine pack/publish failure (bad `files`, missing build output, invalid manifest), and it gets printed as "Expected during a dry run" with exit code 0.

Compounding it: `materializeAndPinManifests` writes nothing under `dryRun`, so the dry run packs the **un-pinned** manifest. The one pre-flight this driver offers therefore neither exercises the pinning nor can report a failure.

Suggest splitting the two cases: a project *absent* from `publishResults` is always fatal — that's the private-package drop this check was added to catch, and it is just as real in a dry run — while a non-zero exit code is the only thing worth tolerating.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+    console.warn(`${summary} Expected during a dry run, which does not write the new versions to disk.`);
+    return;
+  }
+
+  // When a publish target fails, we want to fail the CI
+  console.error(
+    `${summary} No git tags were created, so re-running the release will retry the same versions.`
```
gbarkhatov: This message overstates what a re-run does. If three of four packages published and the fourth failed, the executor will *skip* the three on the retry (already on the registry) rather than retry them — so those three end up on npm with no git tag, no changelog entry and no GitHub release, and if the retry's conventional commits bump past their version, they never get one.

Consider running `releaseChangelog` for the subset that did publish before failing, so a partial success at least leaves tags behind.

Minor, same block: `process.exit()` immediately after `console.error` can truncate the message when stderr is a pipe, which is how Actions runs the step — losing exactly the diagnostic that explains the failure. `release().catch()` on line 172 already produces exit code 1, so throwing here would be both safer and consistent with the rest of the module. Same applies to the `process.exit(0)` on line 54.

tools/release/src/index.ts
gbarkhatov, line 180: The justification for swallowing this doesn't hold on the path that needs it most.

"Tags for a GitHub release already reach the remote through the releases API" is true for the stable path, where nx sends `target_commitish` and GitHub creates the tag. But line 98 sets `createRelease: false` whenever `RELEASE_PREID` is set, and nx honours that — so on the RC path `git push --follow-tags` is the *only* route to the remote, with no backstop.

Failure modes that reach it: `--atomic` rejecting the whole push because the branch moved, a protected ref, or a `workflow_dispatch` against a tag ref leaving a detached HEAD ("You are not currently on a branch").

What follows is silent: the RC is on npm untagged, so the next dispatch finds no `-rc.` tag, falls back to the latest stable tag, re-derives the identical `x.y.z-rc.0`, the executor prints "Skipped … already exists" and returns 0, and `assertEveryProjectPublished` passes. The RC workflow becomes a permanent no-op that reports success. The previous driver threw here.

At minimum, rethrow when `RELEASE_PREID` is set.

tools/release/src/pinManifests.ts
```diff
@@ -0,0 +1,273 @@
...
+  manifest: PackageManifest;
+  releasePackages: ReleasePackages;
+  resolvedVersions: ResolvedVersions;
+  releaseTagPattern: string;
+}): Promise<ReadonlyMap<string, SiblingRelease>> => {
+  const siblings = new Map<string, SiblingRelease>();
+
+  for (const section of PUBLISHED_DEPENDENCY_SECTIONS) {
+    for (const dependencyName of Object.keys(manifest[section] ?? {})) {
+      if (!releasePackages.has(dependencyName) || siblings.has(dependencyName)) {
```
gbarkhatov, line 208: The gate only sees dependencies that resolve to a package under the `release.projects` globs (`packages/*`), which leaves a hole in the same bug class.

`resolveSiblings` skips anything not in `releasePackages`, and `findSiblingPinViolations` does `if (!sibling) continue`. So a `workspace:` / `file:` / `link:` spec in `dependencies`, `peerDependencies` or `optionalDependencies` that points at a `tools/*` or `services/*` workspace package is neither rewritten nor flagged — it ships to npm literally, unresolvable by any consumer. Nothing does that today, but the whole point of this PR is that "nothing does that today" is how the current 100 broken releases happened.

Cheap fix that closes it by construction: after the sibling pass, assert that *no* spec in the published sections still matches `isLocalDependencySpec`, sibling or not. That check is independent of package discovery, so it also covers the case where `readReleasePackages` fails to discover a package it should have.

tools/release/src/pinManifests.ts
```diff
@@ -0,0 +1,273 @@
...
+      if (violations.length > 0) {
+        throw new Error(
+          formatViolations(packageName, releasePackage.manifestPath, violations)
+        );
+      }
+
+      for (const sibling of siblings.values()) {
+        assertSiblingIsInstallable(
+          sibling,
+          await registry.fetchPublishedVersions(sibling.packageName)
```
gbarkhatov, line 123: The registry is hit before `assertSiblingIsInstallable` gets the chance to return early.

`assertSiblingIsInstallable` short-circuits at `manifest.ts:195` for `publishedByThisRun`, but this `await` has already completed by then, and `requestPublishedVersions` throws on any network error.

So: a `main` run bumps both `ts-sdk` and `babylon-tbv-rust-wasm`, npm returns a transient 502, and the release aborts with "Could not reach … to check which versions of babylon-tbv-rust-wasm are published" — over an answer that would have been discarded one line later. Hoisting the `publishedByThisRun` test above the fetch removes a whole class of avoidable release failures.

tools/release/src/pinManifests.ts
```diff
@@ -0,0 +1,273 @@
...
+
+  return tags
+    .split('\n')
+    .map((tag) => tag.trim())
+    .filter((tag) => tag.startsWith(prefix))
+    .map((tag) => tag.slice(prefix.length))
+    .find(
+      // git's version sort ranks 1.0.0-rc.1 above 1.0.0, and a stable dependent
+      // must never pin a prerelease.
+      (version) => isExactVersion(version) && !version.includes('-')
```
gbarkhatov: This fallback is the least self-evidently correct code in the package, it has zero test coverage, and the RC path now routes through it on every dispatch.

Three mutations survive the suite untouched: making the function always return `undefined`; deleting the `releaseTagPattern` must-end-in-`{version}` guard; and deleting this `!version.includes('-')` filter, which is the only thing stopping a stable dependent pinning a prerelease.

It also diverges from how nx resolves the same thing in `utils/git.js`:

- nx passes `--merged` (unless `checkAllBranchesWhen` matches) so it only considers tags reachable from the current branch. This lists every tag in the repo, so on the RC path — dispatched from a feature branch — a tag created on some unrelated branch can be selected as a sibling's version, pinning a release to code that isn't in the branch being released.
- nx passes `-c versionsort.suffix=-` and uses a real `isPrerelease()`. Raw `includes('-')` happens to be neutralised here by the stable-only filter, but it would also skip a legitimate stable `1.0.0+build-1`.

Exporting it, or injecting an `execCommand` seam, would make all of the above testable against a canned tag list.

tools/release/src/manifest.ts
```diff
@@ -0,0 +1,222 @@
...
+  'portal:',
+] as const;
+
+/**
+ * The version nx writes when `fallbackCurrentVersionResolver: "disk"` fires
+ * because no git tag matched, plus anything else in the 0.0.0 placeholder
+ * family. Publishing a pin to one of these is how
+ * `wallet-connector@1.67.2` came to reference a version that does not exist.
+ */
+const PLACEHOLDER_VERSION_PATTERN = /^0\.0\.0(?:[-+]|$)/;
```
gbarkhatov, line 54: This pattern misses the one version its own comment names, three lines further down the file.

`/^0\.0\.0(?:[-+]|$)/` matches only the `0.0.0` family, but `findSpecViolationReason`'s comment cites `babylon-tbv-rust-wasm@0.1.0` as "the exact value that shipped broken for months" — and `packages/babylon-tbv-rust-wasm/package.json` is *still* `0.1.0` on disk, while `fallbackCurrentVersionResolver: "disk"` is still configured.

So if the disk fallback ever fires for that package — tag deleted, tag pattern changed, project renamed — `currentVersion` becomes `0.1.0`, `isPlaceholderVersion('0.1.0')` is false, `0.1.0` genuinely *is* on the registry so `assertSiblingIsInstallable` passes, and `ts-sdk` ships pinning a build 14 minor versions stale. Silently, through both gates, reproducing the original bug byte-for-byte.

The regex is a heuristic for "nx used the disk fallback" rather than a check of it. Since the driver already has `latestReleasedVersionFromGitTags`, cross-checking nx's `currentVersion` against the tag-derived version would detect the fallback directly instead of guessing at it from the version string.

tools/release/src/__tests__/pinManifests.test.ts
```diff
@@ -0,0 +1,126 @@
...
+import type { RegistryClient } from '../registry.js';
+import type { ReleaseConfig, ReleasePackages } from '../workspace.js';
+
+const releaseConfig: ReleaseConfig = {
+  registryUrl: 'https://registry.npmjs.org',
+  releaseTagPattern: '{projectName}/{version}',
+  projectGlobs: ['packages/*'],
+};
+
+describe('materializeAndPinManifests', () => {
```
gbarkhatov, line 29: Both tests here assert a rejection, so the behaviour this PR exists to produce is never inspected — and that's #2209's first acceptance criterion.

Mutation-verified, both leave the suite green at 13/13:

- commenting out `manifest = pinned.manifest` (`pinManifests.ts:111`), so the pinning is computed and thrown away;
- disabling the version-materialization block (`pinManifests.ts:77-80`), so the manifest keeps its stale on-disk version.

`pinSiblingDependencies` is unit-tested in isolation, but the *wiring* — that the version fed to it is `newVersion ?? currentVersion` and not `manifest.version` — is exactly the mechanism behind `ts-sdk@0.62.1 → rust-wasm@0.1.0`, and nothing asserts it.

One test covering "only ts-sdk is bumped, wasm resolves from its tag, the resulting manifest pins `0.15.0`" would close it. Related: the `dryRun` "nothing is written" guarantee documented at `pinManifests.ts:52-55` is also unreachable — both tests throw before the write — so changing `if (!dryRun)` to `if (true)` keeps the suite green. Injecting a writer into `PinManifestsOptions` would cover both at once and let the test assert the exact bytes handed to it.

tools/release/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "composite": false,
+    "declaration": false,
+    "declarationMap": false,
+    "module": "esnext",
+    // tsx resolves like a bundler: `nx` is CommonJS with no "exports" map, so
+    // `nx/release` only resolves through directory-index lookup.
+    "moduleResolution": "bundler",
+    "noEmit": true,
```
gbarkhatov, line 11: The PR description says `tools/release` "gains tests, lint and typecheck" and that nx picks the targets up from the package scripts. `lint` and `test` are correctly inferred, but `typecheck` isn't — and `noEmit: true` here is why.

`nx show project @internal/release` reports the inferred target as disabled:

```
"typecheck": { "options": { "command": "echo \"The 'typecheck' target is disabled because one or more project references set 'noEmit: true' ...\"" } }
```

There's also no `typecheck` script in `package.json`, no workflow runs `--target=typecheck` (`verify.yml:58` deliberately omits the step, and `package-release.yml` runs only build/lint/test), this package has no `build` target, and `tsx` strips types without checking them. vitest does the same via esbuild.

Net: the release driver's TypeScript is never type-checked anywhere in CI, so a signature drift between production code and tests surfaces for the first time mid-release on `main`. `"typecheck": "tsc -p tsconfig.json"` in `package.json` — the tsconfig is already `noEmit` — plus a step in `package-release.yml` closes it.

tools/release/src/manifest.ts
```diff
@@ -0,0 +1,222 @@
...
+ */
+
+/**
+ * The manifest sections a consumer resolves when it installs the package.
+ * `devDependencies` is deliberately absent: consumers never install them, and
+ * every package here carries `@internal/eslint-config` there as `workspace:*`.
+ */
+export const PUBLISHED_DEPENDENCY_SECTIONS = [
+  'dependencies',
+  'peerDependencies',
```
gbarkhatov, line 14: Including `peerDependencies` here means a sibling peer gets forced to a single exact version, and a hand-written peer range is rejected outright.

`pinSiblingDependencies` rewrites `workspace:^` in this section to a bare version, and `findSpecViolationReason` rejects `^1.108.0` with "not an exact version - releases pin workspace siblings exactly". No sibling sits in `peerDependencies` today, but `"@babylonlabs-io/core-ui": "workspace:^"` in `wallet-connector`'s peers is the ordinary way to express a peer on a UI library, and an exact peer is unsatisfiable for any consumer that also depends on `core-ui` directly.

The same range-flattening applies to `workspace:^` / `workspace:~` in `dependencies` — it discards the range intent rather than translating it. Moot while syncpack forces `workspace:*`, but worth either honouring the prefix (nx's own `versionPrefix: 'auto'` does exactly that) or restricting the exact-pin rule to `dependencies` and `optionalDependencies`.

tools/release/src/pinManifests.ts
```diff
@@ -0,0 +1,273 @@
...
+    const changes: string[] = [];
+    let manifest = releasePackage.manifest;
+
+    const resolved = resolvedVersions.get(packageName);
+    if (resolved && manifest.version !== resolved.version) {
+      changes.push(`version ${manifest.version} -> ${resolved.version}`);
+      manifest = { ...manifest, version: resolved.version };
+    }
+
+    if (publishing.has(packageName)) {
```
gbarkhatov, line 93: `publishing` is a set of nx *project* names (it comes from `Object.entries(projectsVersionData)` in `index.ts:48-50`), but `packageName` here is the `releasePackages` map key, i.e. `manifest.name`.

They coincide today — every `packages/*/project.json` sets `name` to the npm package name — but ten lines down `resolveReleasedVersions` does the same lookup *through* `releasePackageForProject`, which `workspace.ts:128-131` documents as existing specifically to keep that assumption "explicit and loud". This line bypasses it, and fails in the worst direction if the two ever diverge: the version gets materialized, but sibling pinning, the violation check and the registry gate are all skipped silently and `workspace:*` ships.

Routing it through `releasePackageForProject` costs nothing and makes the divergence loud instead of silent.

tools/release/src/__tests__/manifest.test.ts
```diff
@@ -0,0 +1,248 @@
...
+  it('leaves devDependencies alone', () => {
+    // Every package depends on the private @internal/eslint-config there.
+    // Consumers never install devDependencies, and the package is not
+    // releasable, so rewriting it would fail the release for no benefit.
+    const manifest: PackageManifest = {
+      name: '@babylonlabs-io/core-ui',
+      devDependencies: { '@internal/eslint-config': 'workspace:*' },
+    };
+
+    const pinned = pinSiblingDependencies(manifest, new Map());
```
gbarkhatov: This test cannot fail. Passing `new Map()` means `siblings.get('@internal/eslint-config')` misses regardless of which sections are iterated — I added `'devDependencies'` to `PUBLISHED_DEPENDENCY_SECTIONS` and the suite stayed green at 13/13.

That matters because it's the only test guarding the decision that keeps `@internal/eslint-config` — private and unreleasable — from being rewritten and failing every release. Putting it in the siblings map with a concrete version makes the test fail if `devDependencies` ever enters the section list.

Two more gaps in this file while it's open:

- The "not an exact version" branch (`manifest.ts:173-175`) has no test — deleting it leaves the suite green. All four existing violation cases use `0.0.0-semantic-release`, which *is* valid exact semver, so none of them reaches that branch. `^0.15.0` is the most plausible real-world spec after `workspace:*`, and it would ship a floating range.
- `peerDependencies` and `optionalDependencies` are in `PUBLISHED_DEPENDENCY_SECTIONS` but every test uses `dependencies` only, and `formatViolations` — the entire operator-facing failure message — has no test at all.

tools/release/src/registry.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (cached) return cached;
+
+      const pending = requestPublishedVersions(baseUrl, packageName);
+      inFlight.set(packageName, pending);
+      return pending;
+    },
+  };
+};
+
+const requestPublishedVersions = async (
```
gbarkhatov, line 45: No test covers this file. Replacing the whole client with `Promise.resolve(null)` leaves the suite green at 13/13.

Untested: `404 → null` versus non-ok `→ throw` (that split is load-bearing — it's the difference between "never published" and "registry hiccup", and both must block, but for different reasons the operator needs to tell apart), the fetch-rejection wrapping, a non-JSON body, a packument with no `versions` map, the trailing-slash normalisation of `registryUrl`, the scoped-name encoding, and the `inFlight` memo.

`fetch` is trivially stubbable with `vi.stubGlobal`, so this is cheap to close.

On the memo: `inFlight` is named as if it were transient but is a permanent cache that also caches rejections, so one transient failure poisons every subsequent lookup for that package for the life of the process. Given the module's fail-closed docstring that's arguably intended — but the name says otherwise and nothing pins either reading.

tools/release/src/workspace.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (!registryUrl) {
+    throw new Error(
+      `${nxConfigPath} has no "release.publish.registry", so published versions cannot be verified before publishing.`
+    );
+  }
+
+  return { registryUrl, releaseTagPattern, projectGlobs };
+};
+
+export const readReleasePackages = (
```
gbarkhatov, line 88: No test covers this file either, and one of the mutations that survives re-ships this PR's own bug.

`findSiblingPinViolations` only inspects deps present in the siblings map, and that map is derived from whatever `readReleasePackages` discovers. So a package this function silently fails to discover has its `workspace:*` neither rewritten nor flagged — it publishes verbatim. There is no independent "no local protocol survives into a published manifest" assertion to catch it (see the separate note on `pinManifests.ts:169`).

Mutations that survive untouched: removing the `!manifest.name` throw on line 118-120, removing the `!registryUrl` and `!releaseTagPattern` guards in `readReleaseConfig`, and making `releasePackageForProject` return `undefined` instead of throwing.

All of these are pure functions of a directory tree, so a single `mkdtemp` fixture covers the lot.

packages/babylon-ts-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@babylonlabs-io/ts-sdk",
-  "version": "0.1.2",
+  "version": "0.62.1",
```
gbarkhatov: These corrections are already stale, and only two of six packages got one.

| manifest | this PR writes | tag + npm `latest` |
| --- | --- | --- |
| `babylon-ts-sdk` | `0.62.1` | `0.63.0` |
| `babylon-wallet-connector` | `1.67.2` | `1.69.0` |
| `babylon-core-ui` | untouched — `0.0.0-semantic-release` | `1.109.0` (tag only) |
| `babylon-proto-ts` | untouched — `0.0.0-semantic-release` | `1.23.0` (tag only) |
| `babylon-tbv-rust-wasm` | untouched — `0.1.0` | `0.15.0` |
| `babylon-ledger-vault-signer` | untouched — `0.0.0-semantic-release` | `0.5.0` |

The branch is four commits behind `main`, which is where `0.63.0` and `1.69.0` came from, so these will need re-deriving before merge and will rot again on the next one.

Harmless at runtime — `currentVersionResolver` is `git-tag` and the driver overwrites `manifest.version` from the resolved version before publishing — but the field is still the `fallbackCurrentVersionResolver: "disk"` input, which is exactly the path that reproduces the original bug for `babylon-tbv-rust-wasm@0.1.0` (see the note on `manifest.ts:38`).

Since the value is hand-maintained and nothing asserts it, fixing two of six invites the reader to trust all six. Either correct all of them — `core-ui` needs it anyway to unblock the release, per the note on `manifest.ts:203` — or add a CI assertion that each on-disk version matches its latest release tag.

tools/release/src/manifest.ts
```diff
@@ -0,0 +1,222 @@
...
+  readonly dependencyName: string;
+  readonly from: string;
+  readonly to: string;
+}
+
+export interface ManifestViolation {
+  readonly section: PublishedDependencySection;
+  readonly dependencyName: string;
+  readonly spec: string;
+  readonly reason: string;
```
gbarkhatov, line 109: `ManifestViolation` carries only free prose, so four tests in `manifest.test.ts` identify which rule fired by substring-matching the error message — `toContain('floats to whatever version is newest')` and friends. Re-wording an error for clarity then breaks a test with no behavioural change, and a test cannot actually assert "the wildcard rule fired" any other way.

CLAUDE.md's "prefer discriminated unions over optional fields when states are mutually exclusive" applies directly: adding `readonly code: 'local-protocol' | 'wildcard' | 'inexact' | 'placeholder' | 'mismatch'` lets the tests assert on the discriminant and leaves `reason` free to be as human as it likes.

.github/workflows/package-release.yml
```diff
@@ -70,6 +70,9 @@ jobs:
       - name: Test all
         run: pnpm exec nx run-many --target=test --all
 
+      # Must stay last for its event: the release resolves `workspace:*` deps to
+      # concrete versions on disk, which makes the specifiers recorded in
+      # pnpm-lock.yaml stale. Any `pnpm install` after this step would fail.
       - name: Release changed packages
```
gbarkhatov, line 79: Nothing in CI ever runs the driver in dry-run mode, so the gate this PR is built around first executes in production.

`tools/release` has a `dry-run` script, but no workflow invokes it, and the driver itself only runs on `push` to `main` and on `workflow_dispatch`. #2209's acceptance criteria explicitly call for verifying with `DRY_RUN=true pnpm run ci:release` and inspecting the printed manifest diff — a step doing exactly that on PRs would have surfaced the unpublished `core-ui@1.109.0` tag before merge instead of at release time.

Worth adding to `verify.yml` (or here, gated on `pull_request`). It needs the dry-run path to actually be able to fail first — see the note on `index.ts:133`.

gbarkhatov (dismissed):
Reviewed the release driver against nx 21.3.7's actual source, the live npm registry, and the repo's git tags. The mechanism is right — pinning from nx's resolved version data is truthful whether or not a sibling was co-released, which is a better answer than either option #2209 proposed, and the private-package and dropped-from-publish-run checks go beyond what the issue asked for. Approving the design.

Two things will bite on merge, though, and I'd want them resolved before this lands.

**The first release after merge aborts and nothing publishes.** `@babylonlabs-io/core-ui/1.109.0` is a git tag on `origin/main` (`49110ecf3`) that returns 404 on npm — latest there is `1.108.1`. Any run that bumps `wallet-connector` without bumping `core-ui` resolves the pin to `1.109.0`, and the new registry gate throws inside `materializeAndPinManifests`, before `releasePublish` runs at all — so `ts-sdk` never publishes either. Details inline on `manifest.ts`.

**The RC path publishes an untagged dependent.** `updateDependents` is `auto`, so an RC dispatch for `babylon-tbv-rust-wasm` also bumps and publishes `ts-sdk` — with a stable-looking version, pinned to a prerelease WASM, and no git tag, because `releaseChangelog` is filtered to `RELEASE_PROJECTS` while `releasePublish` is not. A later stable run then recomputes that same number, finds it on the registry, and moves `latest` onto it. Details inline on `index.ts:90`.

### Does it close #2209?

Mechanically yes; on merge, not yet. Four of the six acceptance criteria pass. AC 2 (a `wallet-connector`-only release against `core-ui`) hard-fails for the reason above, and ACs 5 and 6 are blocked behind it. AC 3 is met, though only for the reject paths — there is no success-path test (inline on `pinManifests.test.ts`). AC 1's stated verification (`DRY_RUN=true pnpm run ci:release`, inspect the printed manifest diff) is not shown anywhere and no CI step runs it.

For the record the bug is still fully live on `latest`: `ts-sdk@0.63.0` ships `babylon-tbv-rust-wasm: "0.1.0"`, and `wallet-connector@1.69.0` ships `core-ui: "0.0.0-semantic-release"`.

Worth splitting the npm remediation (deprecating the ~100 broken versions, re-pointing `ts-sdk`'s `next` dist-tag, which still resolves to `0.15.4-rc-0.16.0` — one of the raw-`workspace:*` prereleases) into its own tracked issue rather than letting #2209 close over it.

### Smaller things with no natural inline home

- The PR description says "its latest release points at a version of the ledger signer that was never published" — that has since drifted. `wallet-connector@1.69.0` pins `ledger-vault-signer@0.5.0`, which *is* published; what makes it uninstallable now is `core-ui@0.0.0-semantic-release`. The test comments citing `1.66.1` and `1.67.2` are accurate against those packuments.
- `DEVELOPMENT.md` is the release runbook and isn't updated. "Release tag for new package" still tells you to hand-create a tag — which under the new gate blocks every dependent's release with no forward path. The new failure messages and their recovery steps belong there.
- `pinManifests.ts:176` `manifest.name ?? '(unnamed package)'` is unreachable — `readReleasePackages` throws on a nameless manifest — and would put a placeholder into a release-blocking error message.
- `readJsonFile`, `isPlaceholderVersion`, `isLocalDependencySpec`, `LOCAL_DEPENDENCY_PROTOCOLS`, `PublishedDependencySection`, `ProjectVersions` are exported but referenced only inside their own file.

### Checked and clean, so nobody re-litigates

`--access public` / `--provenance` still arrive via each package's `publishConfig` plus `NPM_CONFIG_PROVENANCE`; `releaseTagPatternStrictPreid` is real in nx 21.3.7 and falls back to the latest stable tag when no preid tag exists; `RELEASE_PROJECTS` name resolution is correct (every `packages/*/project.json` sets `name` to the npm name); `publishResults` really is keyed by nx project name, and nx only throws on a missing publish target when *no* project has one — so `assertEveryProjectPublished` is the right guard for the mixed case; publish-before-tag does not permanently wedge a retry, because the publish executor is idempotent on an existing version; `workspace:^` / `~` / `1.2.3` are all handled; no empty `dependencies: {}` keys are invented; `encodeURIComponent` on a scoped name is accepted by registry.npmjs.org; the lockfile delta is exactly the new devDeps and syncpack passes; nothing runs after the release step and the deliberately dirty tree is safe because nx passes `--no-git-checks`.

tools/release/src/index.ts
jonybur: Fixed. `releaseChangelog` now gets `projects: projectsToPublish`, the same superset `releasePublish` uses, so a dependent nx pulls in via `updateDependents` is tagged as well as published.

tools/release/src/manifest.ts
jonybur: Fixed. All six on-disk versions are now set to their latest release tag, core-ui included, so core-ui is touched by this commit, nx bumps it to 1.109.1, it co-releases and `publishedByThisRun` is true. Verified: with core-ui in the run the gate passes and wallet-connector pins 1.109.1; with core-ui out of the run it still throws on 1.109.0, which is the correct reading of the registry.

tools/release/src/index.ts
jonybur: The premise does not hold: the implicit `nx-release-publish` target does carry `dependsOn: ['^nx-release-publish']` (`nx/src/utils/package-json.js:106-114`), and `nx show project @babylonlabs-io/ts-sdk --json` confirms it on this repo. Edges between members of the `projects` filter survive `create-task-graph.js:50-63`, and a failed dependency marks its dependents skipped with code 1. So ts-sdk cannot land before the WASM package, and cannot land at all if the WASM publish fails. Leaving this open in case you read the task-graph filtering differently.

tools/release/src/index.ts
jonybur: Testability fixed: the check moved to `publishResults.ts`, is exported, throws instead of calling `process.exit`, and has five tests including the dropped-private-package case. You are right that a zero exit cannot distinguish a fresh publish from the executor's already-published short-circuit; that is nx's executor contract and I have not tried to work around it. The consequence you name for the RC path is closed from the other end, by rethrowing a failed tag push when a preid is set.

tools/release/src/index.ts
jonybur: Fixed, and the comment was wrong. A project missing from `publishResults` is now always fatal, dry run included, since the drop is just as real there. Only a non-zero exit code is tolerated under dry run, and the comment now says why: nothing writes the new version to disk, so the packed manifest is the previous release.

tools/release/src/index.ts
jonybur: Fixed. The failure path now throws rather than calling `process.exit` after `console.error`, and the message no longer claims a re-run retries everything - it says which packages are already immutable and warns that the executor skips a version it finds published. `process.exit(0)` on the no-release path is now a plain return.

tools/release/src/index.ts
jonybur: Fixed. `pushReleaseTags` rethrows when `RELEASE_PREID` is set, since that path sets `createRelease: false` and the push is the only route to the remote. The swallow stays for the stable path, where the releases API has already created the tag.

tools/release/src/pinManifests.ts
jonybur: Fixed. Added `findSurvivingLocalProtocols`, which scans the published sections for any local protocol regardless of the sibling map, and runs alongside the sibling check. Covered by tests for both a non-sibling `workspace:*` and the devDependencies exemption.

tools/release/src/pinManifests.ts
jonybur: Fixed. The `publishedByThisRun` test is now above the fetch, so a version this run publishes never triggers a lookup and a transient registry error cannot abort the release over an answer that would be discarded.

tools/release/src/pinManifests.ts
jonybur: Fixed. Moved to `releaseTags.ts` behind an injectable git runner, with nine tests covering the prefix, parsing, prerelease detection and ordering. It now matches nx's own invocation: `-c versionsort.suffix=-`, `--sort -v:refname`, `--merged` last, plus nx's fallback to all branches when nothing is reachable. Prerelease detection strips build metadata first, so `1.0.0+build-1` is no longer treated as a prerelease.

tools/release/src/manifest.ts
jonybur: Fixed, and you are right that the regex was guessing. There is now a direct check: any version nx reports as `currentVersion` must appear in that package's release tag list, otherwise the disk fallback fired and the release stops. `0.1.0` has no tag, so it is rejected even though it is valid semver and really is on the registry. The placeholder check stays for the friendlier message.

tools/release/src/__tests__/pinManifests.test.ts
jonybur: Fixed. `writeManifest` is injected, and there is now a test asserting the exact object handed to it: ts-sdk bumped to 0.62.2 with the WASM dep rewritten from `workspace:*` to 0.15.0 while its on-disk version says 0.1.0. A second test asserts the writer is never called under `dryRun`, and a third covers the RC shape where the sibling is resolved from its latest stable tag.

tools/release/tsconfig.json
jonybur: Fixed. Added a `typecheck` script and wired it into both workflows, so the driver's TypeScript is now checked in the PR gate and again before a release.

tools/release/src/manifest.ts
jonybur: Fixed. `workspace:^` and `workspace:~` now resolve to `^x.y.z` and `~x.y.z` instead of a bare version, and the violation check accepts a `^`/`~` prefix as long as it names the version being released. A `workspace:` range that cannot be translated is left alone and reported rather than flattened.

tools/release/src/pinManifests.ts
jonybur: Fixed. The publishing set is now built by mapping each project name through `releasePackageForProject`, so a divergence between nx project name and package name fails loudly instead of silently skipping the pinning and the gate.

tools/release/src/__tests__/manifest.test.ts
jonybur: Fixed. The devDependencies test now puts `@internal/eslint-config` in the siblings map with a version, so adding `devDependencies` to `PUBLISHED_DEPENDENCY_SECTIONS` fails it. Added tests for the inexact branch (`>=0.15.0`), for `optionalDependencies` and `peerDependencies`, and for `formatViolations`.

tools/release/src/registry.ts
jonybur: Fixed. Nine tests via `vi.stubGlobal('fetch', ...)`: the 404-to-null versus non-ok-to-throw split, fetch rejection, non-JSON body, missing versions map, scoped-name encoding, trailing-slash normalisation and the memo. The cache is renamed `answers` and documented as deliberately caching rejections, since the gate fails closed. Also added a 30s request timeout, per CodeRabbit on the same file.

tools/release/src/workspace.ts
jonybur: Fixed. Ten tests over a `mkdtemp` fixture covering discovery, the manifest-less directory case, the missing-name throw, the unsupported glob shape, all three `readReleaseConfig` guards, the `releasePackageForProject` throw and the exact bytes `writeManifest` produces. The independent no-local-protocol-survives assertion you suggested on `pinManifests.ts:169` closes the discovery-failure hole from the other side.

packages/babylon-ts-sdk/package.json
jonybur: Fixed. All six are now set, re-derived from the release tags: ts-sdk 0.63.0, wallet-connector 1.69.0, core-ui 1.109.0, babylon-proto-ts 1.23.0, babylon-tbv-rust-wasm 0.15.0, ledger-vault-signer 0.5.0. Touching all six is also what makes each one co-release and clear the gate. The rot is real; the direct fix for the hazard you link to is the tag cross-check, which now rejects any version nx read off disk.

tools/release/src/manifest.ts
jonybur: Fixed. `ManifestViolation` now carries `code: 'local-protocol' | 'wildcard' | 'inexact' | 'placeholder' | 'mismatch'`, and every violation test asserts on the code instead of substring-matching the prose.

.github/workflows/package-release.yml
jonybur: Fixed. `verify.yml` gains a 'Rehearse the release' step running the driver with `DRY_RUN=true` on every PR, after the dry-run path was made able to fail. Checkout also sets `fetch-tags: true`, since the rehearsal resolves versions from release tags.

tools/release/package.json
```diff
@@ -4,10 +4,17 @@
...
+    "lint": "eslint .",
+    "run": "tsx src/index.ts",
+    "test": "vitest run"
   },
   "dependencies": {
     "nx": "21.3.7"
+  },
+  "devDependencies": {
+    "@internal/eslint-config": "workspace:*",
+    "typescript": "^5.8.3",
```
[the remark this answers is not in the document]
jonybur: Not taking this one. syncpack's default version group requires a single version for a dependency across the workspace, and every other package declares `typescript: ^5.8.3`. Pinning 5.9.2 here would fail the Single Version Policy step in verify.yml. Happy to pin it repo-wide in a separate change if that is the direction.

packages/babylon-wallet-connector/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@babylonlabs-io/wallet-connector",
-  "version": "0.0.0-semantic-release",
+  "version": "1.67.2",
```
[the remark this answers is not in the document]
jonybur: Fixed. All six manifests now carry their latest release tag version instead of a placeholder: ts-sdk 0.63.0, wallet-connector 1.69.0, core-ui 1.109.0, babylon-proto-ts 1.23.0, babylon-tbv-rust-wasm 0.15.0, ledger-vault-signer 0.5.0. The `workspace:*` specs stay as they are - syncpack requires them, and the driver resolves them at publish time.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+    verbose: true,
+  });
+
+  /**
+   * We intentionally not commit all the version changes but only push the tags
+   */
+  if (!DRY_RUN) {
+    await pushReleaseTags();
+  }
+};
```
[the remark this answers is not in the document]
jonybur: Fixed. The tag step is wrapped, and on failure it names every package that is already on the registry and says to create the tags by hand rather than re-run, before rethrowing.

tools/release/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  } catch (error) {
+    /**
+     * Tags for a GitHub release already reach the remote through the releases
+     * API, so this push is usually a no-op. Everything of value has been
+     * published by this point, and failing here would report a successful
+     * release as broken.
+     */
+    console.warn(`Could not push to the remote: ${error}`);
+  }
+};
```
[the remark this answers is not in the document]
jonybur: Fixed. `pushReleaseTags` rethrows when `RELEASE_PREID` is set, because that path sets `createRelease: false` and the push is the only route to the remote.

tools/release/src/registry.ts
```diff
@@ -0,0 +1,77 @@
...
+  try {
+    response = await fetch(url, {
+      headers: { accept: ABBREVIATED_PACKUMENT_ACCEPT },
+    });
+  } catch (cause) {
+    throw new Error(
+      `Could not reach ${url} to check which versions of ${packageName} are published. The release cannot verify its dependency pins, so it is stopping here.`,
+      { cause }
+    );
+  }
```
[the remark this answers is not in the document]
jonybur: Fixed. Added a 30s `AbortSignal.timeout` on the request, reported through the existing catch, plus tests for the fetch-rejection path.

tools/release/src/pinManifests.ts
jonybur: Fixed. All six on-disk versions now carry their latest release tag, core-ui included, so this commit touches core-ui, nx bumps it to 1.109.1 and it co-releases - `publishedByThisRun` is true and the registry check is skipped. Confirmed by a dry run with core-ui in the release set.

tools/release/src/index.ts
jonybur: Fixed. `releaseChangelog` now gets `projects: projectsToPublish`, the same list `releasePublish` uses, so a dependent nx drags in gets tagged too.

tools/release/src/index.ts
jonybur: Fixed. It rethrows when a preid is set. The swallow only stays for the stable path, where nx has already created the tag through the GitHub releases API.

tools/release/src/index.ts
jonybur: Fixed. A project missing from `publishResults` is now always fatal, dry run included. Only a non-zero exit code is tolerated under dry run, and the check moved to its own module so it is actually testable - there is a test for the dropped private package.

.github/workflows/package-release.yml
```diff
@@ -64,12 +64,18 @@ jobs:
       - name: Build all
         run: pnpm exec nx run-many --target=build --all
 
+      - name: Typecheck the release driver
+        run: pnpm --filter @internal/release run typecheck
```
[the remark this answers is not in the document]
jonybur: Taken for the typecheck: both workflows now run `nx run @internal/release:typecheck`. The rehearsal already goes through nx - `pnpm run ci:release` is `nx run @internal/release:run`, which is how the stable release path has always been invoked.

tools/release/src/pinManifests.ts
```diff
@@ -0,0 +1,276 @@
...
+      projectName
+    );
+
+    if (versions.newVersion === null) {
+      assertVersionWasReleased(
+        packageName,
+        versions.currentVersion,
+        await releaseTags.readReleasedVersions(projectName)
+      );
+    }
```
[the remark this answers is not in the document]
jonybur: Good catch, and taken. The check moved to the point of use: a version is only required to have a release tag when something published is about to pin it. A package under the release globs that has never been released no longer blocks anything, and the disk-fallback case is still rejected because it is rejected where it does damage. Test added for the new package; removing the relocated check still fails the disk-fallback test.

tools/release/src/publishResults.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  if (dryRun) {
+    console.warn(
+      `${summary} A dry run does not write the new versions to disk, so the packed manifest is the previous release and the registry already has it.`
+    );
+    return;
+  }
+
+  throw new Error(
+    `${summary} Any package that did publish is already immutable on the registry; check which ones landed before re-running, because the publish executor skips a version it finds already published.`
```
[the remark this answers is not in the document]
jonybur: Not taking this one. CLAUDE.md's test philosophy asks for the opposite in test files - "Prefer explicit values over magic constants imported from elsewhere" and "Each test makes its setup, action, and assertions clear in the function body" - so lifting fixture package names and status codes out of these tests would make them harder to read top to bottom, which is what that section optimises for. On the production side, `code !== 0` for a process exit code and the message templates are each used once at their point of use; naming them would add indirection without making the value's choice any clearer.

tools/release/src/publishResults.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  if (failed.length === 0) return;
+
+  const summary = `These projects were versioned but did not publish: ${failed.join(', ')}.`;
+
+  if (dryRun) {
+    console.warn(
+      `${summary} Expected during a dry run, which packs the previous release because the new versions are never written to disk.`
+    );
+    return;
```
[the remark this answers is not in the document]
jonybur: Not taking this one - it is the bug the dry-run rehearsal caught on its first CI run. A result can be absent for two reasons, and only one of them is nx omitting a package it has no publish target for. The other is nx *skipping* a project because a dependency's publish failed, which is what every dry run produces: run 32247765903 failed exactly this way, reporting ts-sdk and wallet-connector as unpublishable when both are public and fine. Both cases are still covered. The package-with-no-publish-target case is now caught before publishing by `assertProjectsArePublishable`, where the private package can be named directly, and in a real run an absent result is already fatal here because `publishResults[name]?.code !== 0` is true for `undefined`.

gbarkhatov (dismissed):

tools/release/src/index.ts
jonybur: Reopening this one: the fix I pushed earlier only closed half of it, and the half it left is worse than what you described.

Routing `releaseChangelog` through `projectsToPublish` fixed the *untagged* dependent. But the dependent is still published, and it now gets a **stable** git tag. Reproduced:

```
$ pnpm exec nx release version --projects=@babylonlabs-io/babylon-tbv-rust-wasm \
    --specifier=prerelease --preid=rc --dry-run

@babylonlabs-io/babylon-tbv-rust-wasm  -> 0.15.1-rc.0
@babylonlabs-io/ts-sdk                 -> 0.62.2   ("because a dependency was bumped")
   "@babylonlabs-io/babylon-tbv-rust-wasm": "0.15.1-rc.0"
```

`bumpVersionForProject(dependent, 'patch', 'DEPENDENCY_WAS_BUMPED')` in `release-group-processor.js` is hardcoded — the preid never reaches it. So an RC dispatch for the WASM package publishes a stable-numbered ts-sdk pinned to a release-candidate engine, and tags it on the stable line so no later run republishes that number. `next` does not contain it either: dist-tags do not constrain semver range resolution, so anything on `^0.63.0` picks it up.

WASM is one of the two dropdown options and has no prereleases on npm, so this path has never run — it would have failed on first use.

Fixed by gating the publish set: a run with a preid publishes prereleases or nothing (`selectProjectsToPublish`). Rehearsed both dropdown options — the WASM RC now cuts only `@babylonlabs-io/babylon-tbv-rust-wasm/0.15.1-rc.0`, and the ts-sdk RC only its own tag. Four tests, mutation-checked.

Flagging for a second reviewer since it lands on the ts-sdk/WASM critical path.

tools/release/src/index.ts
jonybur: Confirmed the ordering empirically rather than leaving it on a reading of the source — ran nx's own `createTaskGraph` against this workspace with the exact filter the driver passes:

```
filter: ['@babylonlabs-io/ts-sdk','@babylonlabs-io/babylon-tbv-rust-wasm']
target: nx-release-publish, excludeTaskDependencies: true

roots        = ['@babylonlabs-io/babylon-tbv-rust-wasm:nx-release-publish']
dependencies = { 'ts-sdk:nx-release-publish': ['babylon-tbv-rust-wasm:nx-release-publish'] }
```

ts-sdk is not a root, so it cannot start first. `nx show project --json` returns `dependsOn: ['^nx-release-publish']` for all six release packages, and `TasksSchedule.canBeScheduled` requires every dependency in `completedTasks`. On failure `TaskOrchestrator.complete` marks reverse-dependencies `skipped` recursively, and a skipped project has no `{code: 0}`, so `assertEveryProjectPublished` throws before any tag is cut.

So the scenario isn't reachable. Three residual fragilities though, none of which I can turn into a live break today — worth knowing they're load-bearing on nx internals that nothing in `tools/release` asserts:

- `publish.js` hardcodes `nxIgnoreCycles: true`, and `_makeAcyclic` **splices** edges out of the dependency lists. A future cycle among release packages would silently destroy the ordering.
- `publish.js` loops per release group with a separate task graph each, so a group-1 failure would not skip a group-2 dependent. Inert only because nx.json declares a single implicit group.
- The `excludeTaskDependencies` filter runs before `filterDummyTasks`, so an edge routed through a project without a publish target is dropped rather than resolved. Harmless for pins, which need a direct dependency and so a direct edge.

Happy to add a comment at the early return citing the mechanism, if you think that's worth pinning.

jonybur:
@gbarkhatov — picking up the items from your review that had no inline home, plus one correction to something I told you was fixed.

**The RC path was not fixed.** Details on the thread, but in short: routing `releaseChangelog` through `projectsToPublish` closed the *untagged* half and left the publish half worse. An RC dispatch for the WASM package published a **stable-numbered** ts-sdk pinned to the release candidate, and stable-tagged it. `next` would not have held it back — dist-tags do not constrain semver range resolution. Now gated: a run with a preid publishes prereleases or nothing.

**The first-release-after-merge abort.** I could not reproduce it on any reachable path. `RELEASE_PROJECTS` is only ever set by the RC dispatch, whose dropdown is limited to ts-sdk and the WASM package; `verify.yml` and the main-push release both run the whole set. Commit `46bc6b199` touches both `packages/babylon-core-ui/package.json` and `packages/babylon-proto-ts/package.json`, so nx bumps core-ui to 1.109.1 and proto-ts to 1.23.1, `publishedByThisRun` is true, and the registry lookup is short-circuited. Merging retires the unpublished-tag debt rather than tripping over it. I only got your failure by forcing `RELEASE_PROJECTS=@babylonlabs-io/wallet-connector`, which nothing sets. Say if you read that differently.

**The runbook.** Added. The new-package section now says the tag version is never published and that nothing may depend on the package until it has published once, plus a table mapping each message the driver can stop on to its recovery step — all ten checked against the source strings.

**`'(unnamed package)'`.** Unreachable, agreed — `readReleasePackages` throws on a nameless manifest and keys the map by that name. Dropped; the call site already holds the authoritative name, so it is threaded in.

**The file-local exports.** Took four of the six. `PublishedDependencySection` and `ProjectVersions` are field types on exported interfaces (`DependencyRewrite` / `ManifestViolation` / `PinManifestsOptions`), so un-exporting them leaves an exported interface whose field type a caller cannot name — it only compiles because `declaration: false`. Worth noting thirteen more exports in that directory are equally file-local, so the six were a subset rather than an outlier.

**Description drift.** Corrected — `wallet-connector@1.69.0` pins `ledger-vault-signer@0.5.0`, which is published; the uninstallable cause is the `core-ui` placeholder. Counts refreshed against the registry to 101/120 and 65/237.

**The npm remediation** is tracked separately, as you suggested — it needs an interactive npm login, which OIDC cannot do.

Not taken: the success-path test, which `46bc6b199` already added (`writes the version this run resolved and pins the sibling at its released version`, full-object equality).

DEVELOPMENT.md
```diff
@@ -97,6 +97,35 @@ git tag @babylonlabs-io/my-package/1.2.3
...
+
+### When a release fails
+
+The release stops rather than publish a manifest a consumer cannot install. Nothing is tagged unless every package published.
+
+| Message | What happened | Fix |
+|---------|---------------|-----|
+| `is pinned at X but has never been published` | A dependency has a release tag and nothing on NPM | Publish it first - see above |
+| `is pinned by this release but does not exist on the registry` | Its tag exists without a matching publish, so an earlier release failed or was skipped | Bump it in this release, or publish that version by hand |
+| `resolved to version X, which has no release tag` | nx fell back to the version in `packages/*/package.json`, which is not a released version | Tag the version that is actually on NPM |
```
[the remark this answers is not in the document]
jonybur: Good catch, taken in `6a3ef6737` — this was a real hole in the instruction rather than a wording nit.

nx resolves a project's current version from the newest release tag reachable from `HEAD`, and derives the next specifier from the conventional commits *since that tag*. So tagging `HEAD` with an already-published version puts every unreleased commit on the wrong side of it: those changes stop counting toward the next bump and drop out of the changelog, silently.

Reworded to name the commit and say what going wrong costs:

> Tag the version that is on NPM, on the commit that produced it. Tagging `HEAD` instead swallows every unreleased change before it

DEVELOPMENT.md
```diff
@@ -97,6 +97,35 @@ git tag @babylonlabs-io/my-package/1.2.3
...
 ```
 
+The tag version itself is never published - nx only rolls forward from it, so the `0.0.1` above stays a version that exists in git and not on NPM. Push the tag before the package's own code lands, so the commit that adds the package is a conventional commit nx can bump from.
+
+Nothing may depend on the new package until it has published at least once. The release refuses to pin a version that is not on the registry:
+
+```
+@babylonlabs-io/my-package is pinned at 0.0.1 but has never been published.
+Publish it before publishing anything that depends on it.
+```
```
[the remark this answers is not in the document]
jonybur: Taken in `6a3ef6737`. `text` is right here — the block is release output rather than a command.

Worth noting markdownlint is not wired into CI, and the two `git tag` fences just above this one in the same section are also unlabelled (they would want `bash`). Left those alone rather than widening this PR; they are pre-existing and unrelated to the release gate.

gbarkhatov (approved):

PR #2274, opened 2026-08-19
feat(vault): rebuild DepositTerms on presign resume; map typed deposi…

jrwbabylonlab:
Approval wallets resuming at payout signing now rebuild the terms from
chain state through the shared rebuildDepositTerms (new required
lifecycle param: presign mode tolerates advanced siblings and refuses
ack-expired targets before the device ceremony). Deposit-terms
rejections, unsupported wallet methods, and lifecycle refusals map to
dedicated copy on both error surfaces.

Review follow-ups: the approval-wallet presign path now runs the target
status/ack-window gates before the mempool fee read (a stalled deposit's
refund copy wins over a transient read error; the rebuild still re-checks);
terminal refusals hide Retry and stay out of funnel telemetry; typed
wrong-depositor error; resume-specific unsupported-method copy. The
unsupported-method mapping is forward-plumbing for #2219 — this branch
predates #2269, so on main the Ledger provider's payout `signPsbt` is
already wired and the copy only fires for wallets that genuinely lack a
method.

### DIFF
--- a/services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
+++ b/services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
@@ -39,6 +39,24 @@ vi.mock("../../../../services/vault/fetchVaults", () => ({
     mockFetchVaultPayoutScriptPubKey(...args),
 }));
 
+// Presign terms-rebuild collaborators (approval wallets only). Mocked as
+// modules so the software-wallet path can assert they are never even called.
+const mockGetVaultFromChain = vi.fn();
+vi.mock("../../../../clients/eth-contract/btc-vault-registry/query", () => ({
+  getVaultFromChain: (...args: unknown[]) => mockGetVaultFromChain(...args),
+}));
+
+const mockResolveFundedTxFeeAndUtxos = vi.fn();
+vi.mock("../../../../services/vault/resolveFundedTxFee", () => ({
+  resolveFundedTxFeeAndUtxos: (...args: unknown[]) =>
+    mockResolveFundedTxFeeAndUtxos(...args),
+}));
+
+const mockRebuildDepositTerms = vi.fn();
+vi.mock("../../../../services/vault/rebuildDepositTerms", () => ({
+  rebuildDepositTerms: (...args: unknown[]) => mockRebuildDepositTerms(...args),
+}));
+
 let mockBtcConnector: {
   connectedWallet?: {
     account?: { address: string };
@@ -97,6 +115,10 @@ const PROVIDER = { btcPubKey: "0xvpkey" };
 const BTC_WALLET = { signPsbt: vi.fn() };
 const onSuccess = vi.fn();
 
+const ON_CHAIN_VAULT = { marker: "on-chain-vault" };
+const REBUILT_TERMS = { marker: "rebuilt-terms" };
+const FUNDED_TX_FEE = 1234n;
+
 function setupHappyPath() {
   mockFindProvider.mockReturnValue(PROVIDER);
   mockBtcConnector = {
@@ -133,6 +155,12 @@ describe("usePayoutSigningState", () => {
     mockSignAndSubmitPayouts.mockResolvedValue(undefined);
     mockVerifyBtcWalletLiveness.mockResolvedValue(undefined);
     mockFetchVaultPayoutScriptPubKey.mockResolvedValue(null);
+    mockGetVaultFromChain.mockResolvedValue(ON_CHAIN_VAULT);
+    mockResolveFundedTxFeeAndUtxos.mockResolvedValue({
+      expectedUtxos: {},
+      fundedTxFee: FUNDED_TX_FEE,
+    });
+    mockRebuildDepositTerms.mockResolvedValue(REBUILT_TERMS);
   });
 
   describe("happy path", () => {
@@ -717,6 +745,131 @@ describe("usePayoutSigningState", () => {
       const terms = { marker: "payout-wrapper-terms" };
       await call.btcWallet.approveDepositTerms(terms);
       expect(underlyingWallet.approvedWith).toEqual([terms]);
+    });
+  });
+
+  describe("presign deposit-terms rebuild (approval wallets)", () => {
+    // Minimal approval-capable wallet: supportsDepositApproval only probes
+    // for an approveDepositTerms function.
+    function connectApprovalWallet() {
+      mockBtcConnector = {
+        connectedWallet: {
+          account: { address: "tb1test" },
+          provider: { signPsbt: vi.fn(), approveDepositTerms: vi.fn() },
+        },
+      };
+    }
+
+    it("refuses to sign when an approval wallet has no funded Pre-PegIn hex to rebuild terms from", async () => {
+      connectApprovalWallet();
+
+      const { result } = renderHookWithProps({
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        activity: { ...ACTIVITY, unsignedPrePeginTx: undefined } as any,
+      });
+
+      await act(async () => {
+        await result.current.handleSign();
+      });
+
+      expect(result.current.error?.title).toBe("Missing Pre-Pegin transaction");
+      expect(mockGetVaultFromChain).not.toHaveBeenCalled();
+      expect(mockResolveFundedTxFeeAndUtxos).not.toHaveBeenCalled();
+      expect(mockRebuildDepositTerms).not.toHaveBeenCalled();
+      expect(mockSignAndSubmitPayouts).not.toHaveBeenCalled();
+    });
+
+    it("rebuilds presign terms chain-fresh and threads them into signAndSubmitPayouts", async () => {
+      connectApprovalWallet();
+
+      const { result } = renderHookWithProps();
+
+      await act(async () => {
+        await result.current.handleSign();
+      });
+
+      expect(mockGetVaultFromChain).toHaveBeenCalledWith(ACTIVITY.id);
+      expect(mockResolveFundedTxFeeAndUtxos).toHaveBeenCalledWith(
+        ACTIVITY.unsignedPrePeginTx,
+      );
+      expect(mockRebuildDepositTerms).toHaveBeenCalledWith({
+        vaultId: ACTIVITY.id,
+        target: ON_CHAIN_VAULT,
+        fundedPrePeginTxHex: ACTIVITY.unsignedPrePeginTx,
+        connectedDepositorAddress: "0xeth",
+        depositorBtcPubkey: "0x" + "ab".repeat(32),
+        fundedTxFee: FUNDED_TX_FEE,
+        lifecycle: "presign",
+      });
+      expect(mockSignAndSubmitPayouts).toHaveBeenCalledOnce();
+      expect(mockSignAndSubmitPayouts.mock.calls[0][0].depositTerms).toBe(
+        REBUILT_TERMS,
+      );
+      expect(result.current.isComplete).toBe(true);
+    });
+
+    it("adds no rebuild calls and no depositTerms key for software wallets", async () => {
+      // setupHappyPath (beforeEach) connects the plain signPsbt-only wallet.
+      const { result } = renderHookWithProps();
+
+      await act(async () => {
+        await result.current.handleSign();
+      });
+
+      expect(mockGetVaultFromChain).not.toHaveBeenCalled();
+      expect(mockResolveFundedTxFeeAndUtxos).not.toHaveBeenCalled();
+      expect(mockRebuildDepositTerms).not.toHaveBeenCalled();
+
+      // Pre-change param shape, key-identical: no depositTerms key at all
+      // (not even an explicit undefined).
+      const call = mockSignAndSubmitPayouts.mock.calls[0][0];
+      expect("depositTerms" in call).toBe(false);
+      expect(Object.keys(call).sort()).toEqual(
+        [
+          "btcWallet",
+          "depositorBtcPubkey",
+          "depositorEthAddress",
+          "onProgress",
+          "peginTxHash",
+          "providerBtcPubKey",
+          "registeredPayoutScriptPubKey",
+          "signal",
+          "unsignedPrePeginTxHex",
+          "vaultId",
+        ].sort(),
+      );
+      expect(call.vaultId).toBe(ACTIVITY.id);
+      expect(call.peginTxHash).toBe(ACTIVITY.peginTxHash);
+      expect(call.depositorBtcPubkey).toBe("0x" + "ab".repeat(32));
+      expect(call.providerBtcPubKey).toBe(PROVIDER.btcPubKey);
+      expect(call.registeredPayoutScriptPubKey).toBe(
+        ACTIVITY.depositorPayoutBtcAddress,
+      );
+      expect(call.depositorEthAddress).toBe("0xeth");
+      expect(call.unsignedPrePeginTxHex).toBe(ACTIVITY.unsignedPrePeginTx);
+      expect(result.current.isComplete).toBe(true);
+    });
+
+    it("surfaces a rebuild failure through the modal's mapped error state", async () => {
+      connectApprovalWallet();
+      mockRebuildDepositTerms.mockRejectedValueOnce(
+        new Error("resume refused"),
+      );
+
+      const { result } = renderHookWithProps();
+
+      await act(async () => {
+        await result.current.handleSign();
+      });
+
+      // Mapped by formatPayoutSignatureError (stubbed above), not a guard title.
+      expect(result.current.error).toEqual({
+        title: "Sign Error",
+        message: "resume refused",
+      });
+      expect(mockSignAndSubmitPayouts).not.toHaveBeenCalled();
+      expect(result.current.signing).toBe(false);
+      expect(result.current.isComplete).toBe(false);
     });
   });
 });
--- a/services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
+++ b/services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
@@ -1,9 +1,11 @@
+import { OnChainBtcVaultStatus } from "@babylonlabs-io/ts-sdk/tbv/core/clients";
 import { act, renderHook, waitFor } from "@testing-library/react";
 import { StrictMode } from "react";
 import type { Hex } from "viem";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import { LocalStorageStatus } from "../../../../models/peginStateMachine";
+import { VaultLifecycleStateError } from "../../../../utils/errors/vaultLifecycleStateError";
 import { usePayoutSigningState } from "../usePayoutSigningState";
 
 // Mock the SDK adapter the hook delegates to. We don't care here whether the
@@ -53,7 +55,10 @@ vi.mock("../../../../services/vault/resolveFundedTxFee", () => ({
 }));
 
 const mockRebuildDepositTerms = vi.fn();
+const mockAssertPresignTargetSignable = vi.fn();
 vi.mock("../../../../services/vault/rebuildDepositTerms", () => ({
+  assertPresignTargetSignable: (...args: unknown[]) =>
+    mockAssertPresignTargetSignable(...args),
   rebuildDepositTerms: (...args: unknown[]) => mockRebuildDepositTerms(...args),
 }));
 
@@ -156,11 +161,18 @@ describe("usePayoutSigningState", () => {
     mockVerifyBtcWalletLiveness.mockResolvedValue(undefined);
     mockFetchVaultPayoutScriptPubKey.mockResolvedValue(null);
     mockGetVaultFromChain.mockResolvedValue(ON_CHAIN_VAULT);
+    // `vi.clearAllMocks` does not drain queued `*Once` implementations; reset
+    // the mocks that tests arm with one-shot failures before re-installing
+    // their defaults so an unconsumed queue cannot leak across tests.
+    mockResolveFundedTxFeeAndUtxos.mockReset();
+    mockRebuildDepositTerms.mockReset();
+    mockAssertPresignTargetSignable.mockReset();
     mockResolveFundedTxFeeAndUtxos.mockResolvedValue({
       expectedUtxos: {},
       fundedTxFee: FUNDED_TX_FEE,
     });
     mockRebuildDepositTerms.mockResolvedValue(REBUILT_TERMS);
+    mockAssertPresignTargetSignable.mockResolvedValue(undefined);
   });
 
   describe("happy path", () => {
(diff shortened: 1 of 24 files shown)


services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -657,14 +664,14 @@ export function ResumeActivationContent({
...
       setLoading(false);
       return;
     }
     if (!activity.unsignedPrePeginTx) {
-      setLocalError(
-        "Missing Pre-Pegin transaction; cannot recover HTLC secret",
-      );
+      setLocalError({
+        raw: "Missing Pre-Pegin transaction; cannot recover HTLC secret",
+      });
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

services/vault/src/services/vault/rebuildDepositTerms.ts
```diff
@@ -78,22 +96,59 @@ const SIBLING_HOMOGENEOUS_FIELDS = [
...
+    throw new VaultLifecycleStateError(
+      // The broadcast message is load-bearing: mapDepositError buckets on the
+      // word "broadcast" — keep it byte-identical.
+      lifecycle === "broadcast"
+        ? `A vault in this Pre-PegIn batch is no longer awaiting broadcast ` +
+          `(on-chain status ${member.vault.status}); the batch cannot be broadcast ` +
+          `as one transaction. Resume refused.`
+        : `Vault ${member.vaultId} is no longer awaiting payout signatures ` +
+          `(on-chain status ${member.vault.status}); deposit terms cannot ` +
+          `be rebuilt for signing. Resume refused.`,
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — this message is diagnostic-only; both mappers replace it with COPY before anything is rendered to the user.

services/vault/src/utils/errors/__tests__/formatting.test.ts
```diff
@@ -747,6 +750,150 @@ describe("Error Formatting", () => {
...
+    it("keeps the VP mapping when a JsonRpcError carries an unrelated unsupported-method cause", () => {
+      // Precedence: typed classifications run before the cause walk.
+      const err = Object.assign(
+        new JsonRpcError(RpcErrorCode.PEGIN_NOT_FOUND, "PegIn not found"),
+        { cause: { code: "WALLET_METHOD_NOT_SUPPORTED" } },
+      );
+      expect(formatPayoutSignatureError(err).title).toBe(
+        "Vault provider syncing",
+      );
+    });
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — asserting against COPY is the existing convention throughout this file; changing only the new tests would be locally inconsistent.

services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
```diff
@@ -719,4 +747,129 @@ describe("usePayoutSigningState", () => {
...
+        },
+      };
+    }
+
+    it("refuses to sign when an approval wallet has no funded Pre-PegIn hex to rebuild terms from", async () => {
+      connectApprovalWallet();
+
+      const { result } = renderHookWithProps({
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        activity: { ...ACTIVITY, unsignedPrePeginTx: undefined } as any,
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — matches the file's existing `ACTIVITY as any` fixture pattern (lines 141, 291).

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -330,7 +339,7 @@ export function ResumeWotsContent({
 
   const handleSubmit = useCallback(async () => {
     if (!btcWalletProvider || !connectedBtcAddress) {
-      setError("BTC wallet is not connected");
+      setError({ raw: "BTC wallet is not connected" });
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

jeremy-babylonlabs (dismissed):
Reviewed at fe5ade01ad9fae314e03bc8e96971b2ec0b5dd03.

Traced the fund safety critical path for the approval wallet presign terms rebuild.

rebuildDepositTermsCore self verifies the funded tx hash against the on chain prepeginTxid before trusting those bytes, then byte matches each HTLC output value and scriptPubKey against the funded tx before projecting DepositTerms.

assertDepositTermsMatchSigningContext refuses to sign when the approved terms disagree with the version locked signing context on graph shaping scalars, canonical keeper and challenger key sets, and vault provider membership.

The backfilled payout scriptPubKey flows through the same wallet match guard as the activity supplied value, closing the connectedBtcAddress bypass.

Reentrancy and in flight lock handling are covered, including the synchronous guard throw path that previously leaked the lock.

Validation run: depositErrors (39), formatting (81), ResumeDepositContent (24), usePayoutSigningState (28), useVaultActions (52), vaultPeginBroadcastService (20), and services/vault tsc --noEmit clean.

No material findings. Existing CodeRabbit and Greptile threads are resolved or declined with matching file convention rationale, nothing net new to add.

services/vault/src/utils/errors/formatting.ts
```diff
@@ -518,18 +546,10 @@ export function formatPayoutSignatureError(error: unknown): {
...
-      };
+      return PSE.contractCallFailed;
     }
 
-    return {
-      title: "Payout signing error",
-      message:
-        "An unexpected error occurred while signing payouts. Please try again or contact support.",
-    };
+    return PSE.unexpected;
```
gbarkhatov: This bucket swallows the actionable half of what the presign rebuild throws.

`rebuildDepositTerms` raises plain `Error`s for: wrong depositor wallet (`rebuildDepositTerms.ts:202`), sibling-batch disagreement (`:155`, `:164`), non-contiguous HTLC vector (`:181`), missing vault keepers / universal challengers (`:324`, `:329`), plus both fee-bound refusals in `resolveFundedTxFee.ts:52` and `:58` and every `fetchUTXOFromMempool` failure. None match the four substrings above, so all of them render as "An unexpected error occurred while signing payouts. Please try again or contact support."

The wrong-wallet case is the one that hurts: it is the most user-fixable error the rebuild can produce, and its message already says exactly what to do — "Connect with the depositor wallet to resume." That text never reaches the screen. Most of the rest can never succeed on retry, yet the copy invites one.

Note the inversion against the non-`Error` branch below (`:569-572`): a thrown *string* surfaces its own message, while an `Error` does not. `mapDepositError` bucket 11 (`depositErrors.ts:303-308`) already solves this — sanitized raw message under the generic title, plus `diagnostics` for a bug report. The same fallback here would close both.

Pre-existing shape, but this is the PR that makes it load-bearing.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -285,6 +297,31 @@ export function usePayoutSigningState({
         };
 
       try {
+        // Approval (intent) wallets have nothing in memory to approve on
+        // resume — rebuild the terms from chain + WASM, never browser storage.
+        let depositTerms: DepositTerms | undefined;
+        if (supportsDepositApproval(wallet)) {
+          const onChainVault = await getVaultFromChain(activity.id);
+          const { fundedTxFee } = await resolveFundedTxFeeAndUtxos(
```
gbarkhatov, line 327: Ordering: the cheap, decisive gates run after the expensive ones.

`resolveFundedTxFeeAndUtxos` does one mempool round-trip per Pre-PegIn input before `rebuildDepositTerms` is even called, and inside the rebuild the target status gate and `assertAckWindowOpen` run after `assertVaultCoreVersionSupported` plus full sibling discovery (indexer query + N `getVaultFromChain`).

Both target-only gates are decidable from the `onChainVault` already in hand on line 304, plus one block read and one params read. For the exact population this PR targets — stalled and ack-expired deposits — a mempool hiccup on any prevout surfaces the generic "unexpected error" copy instead of the "Deposit timed out" refund copy this PR exists to deliver.

Suggest hoisting the target status check and the ack-window check ahead of the fee resolution. Their relative order to each other should stay as-is: the comment at `rebuildDepositTerms.ts:364` is correct that running the ack check after the status gate is what lets the refusal report a genuinely still-PENDING status. Only the coupling to sibling discovery needs to go — lifting the target's status check out of `discoverSiblings` is enough.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -201,6 +206,14 @@ export function usePayoutSigningState({
         return;
       }
 
+      // Approval wallets rebuild deposit terms from the funded Pre-PegIn hex
+      // below; a localStorage-only merged activity shape can lack it.
+      const wallet = btcWalletProvider as BitcoinWallet;
+      if (supportsDepositApproval(wallet) && !activity.unsignedPrePeginTx) {
```
gbarkhatov: This guard is scoped to approval wallets, but the field it guards is needed by every wallet on this path.

`ensureAuthenticatedVpClient`'s cold path — which is exactly the cross-device resume this modal serves — calls `calculateBtcTxHash(params.unsignedPrePeginTxHex)` at `ensureAuthenticatedVpClient.ts:69` and `parseFundingOutpointsFromTx(...)` at `:84`, both unconditionally. With `unsignedPrePeginTx` undefined a software-wallet depositor throws inside those and lands on `PSE.unexpected` for the identical missing field, while an approval wallet gets the actionable "Missing Pre-Pegin transaction".

Dropping `supportsDepositApproval(wallet) &&` makes it one guard for one precondition. The copy needs a small edit alongside — `payoutSigningGuards.missingPrePeginTransaction` currently says "your wallet requires it to approve payout signing", which is not the general reason.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -285,6 +297,31 @@ export function usePayoutSigningState({
...
+            activity.unsignedPrePeginTx,
+          );
+          depositTerms = await rebuildDepositTerms({
+            vaultId: activity.id,
+            target: onChainVault,
+            fundedPrePeginTxHex: activity.unsignedPrePeginTx,
+            connectedDepositorAddress: depositorEthAddress,
+            depositorBtcPubkey: btcPublicKey,
+            fundedTxFee,
+            lifecycle: "presign",
```
gbarkhatov, line 337: Everything this rebuild throws lands in the `catch` below, which calls `captureFunnelFailure(TELEMETRY_STAGE.ACTIVATION_PAYOUTS, ...)`. That helper only suppresses `classifyError(err) === "user-rejection"` (`telemetryEvents.ts:94`).

An ack-window-elapsed refusal, or an already-VERIFIED target, is the routine outcome for a stalled deposit — not a payout-signing failure. And `ResumeSignContent` auto-fires `handleSign` on mount via `useRunOnce` (`ResumeDepositContent.tsx:116`), so simply opening the modal on such a deposit emits an error at the exact funnel stage those alerts facet on, once per visit.

This is the same argument `captureFunnelFailure`'s own JSDoc makes for dropping cancellations — "routine depositor drop-off, not a failure ... counting cancellations would inflate the very rate these tags alert on". Suggest skipping the capture when `isVaultLifecycleStateError(err)`, or extending the helper's suppression list.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -710,7 +717,8 @@ export function ResumeActivationContent({
   // "not connected" error surfaces.
   useRunOnce(handleSubmit, !btcWalletProvider || Boolean(connectedBtcAddress));
 
-  const error = localError ?? activationError;
+  const error: CaughtError | null =
+    localError ?? (activationError != null ? { raw: activationError } : null);
   // Terminal only applies to the activation failure (deadline passed), never a
   // local pre-flight error — which localError would override via `??` above.
   const isTerminal = localError == null && errorTerminal;
```
gbarkhatov, line 735: `ResumeSignContent` needs this same `isTerminal` seam and does not have one.

All three new presign outcomes are terminal by their own copy: `ackWindowElapsed` points at the refund path, `signaturesNoLongerNeeded` says signing is over, and the `depositTermsRejected` comment at `copy.ts:878-881` literally says "so no 'try again'". But `ResumeSignContent` still passes `onRetry={error ? handleSign : undefined}` (line 178), which `DepositProgressView` renders as a Retry CTA.

Each click re-runs the whole chain — `getVaultFromChain`, N mempool fetches, indexer query, N sibling reads, block and params reads — then fails identically and fires another `captureFunnelFailure`.

Suggest `usePayoutSigningState` return a terminal flag, set at the catch site from the typed error (`isVaultLifecycleStateError(err) && err.stage === "presign"`, or `isDepositTermsRejectedError(err)`) before `formatPayoutSignatureError` flattens it, and gate `onRetry` on it exactly as this branch does.

services/vault/src/services/vault/rebuildDepositTerms.ts
```diff
@@ -180,6 +247,41 @@ async function discoverSiblings(
...
+  target: OnChainVaultData,
+): Promise<void> {
+  const paramsReader = await getProtocolParamsReader();
+  const [currentBlock, tbvParams] = await Promise.all([
+    ethClient.getPublicClient().getBlockNumber(),
+    paramsReader.getTBVProtocolParams(),
+  ]);
+  const ackDeadlineBlock = target.createdAt + tbvParams.pegInAckTimeout;
+  // Exact contract mirror, all block numbers: expired iff
+  // block.number > createdAt + pegInAckTimeout (PeginLogic.sol:481, :807).
```
gbarkhatov: Two things about this citation, then a reuse note.

**The line numbers are wrong.** Verified against `babylonlabs-io/vault-contracts-aave-v4` @ `2e87a85a` (current `main`): `PeginLogic.sol:481` is inside a natspec `@param` block and `:807` is participant-role-overlap dedup. The real sites are **`:544`** (`submitACK`) and **`:870`** (`reportExpired`). The same citation appears in `__tests__/rebuildDepositTerms.test.ts:81`.

**The substance is right at both sites.** `submitACK:544` is `block.number > vBasic.createdAt + protocolParams.getTBVProtocolParams().pegInAckTimeout` — strict `>`, base `createdAt`, timeout read live — and `reportExpired:870` is the contrapositive `<=`, so boundary-equal is not expired at either site and the `+1n` boundary test is correct.

A bare line number into another repo drifts on every merge to that file, and this is a path where a second reviewer is expected to re-derive the cited claim. Suggest citing by symbol plus a revision pin: `(PeginLogic.submitACK / PeginLogic.reportExpired, vault-contracts-aave-v4 @ 2e87a85a)`.

**While there:** `PeginLogic.sol:539`, immediately above the ack check, is `if (vBasic.status != BTCVaultTypes.BTCVaultStatus.Pending) revert Errors.InvalidBTCVaultStatus();`. So the two presign gates here are exactly `submitACK`'s two preconditions, in the same order. That is a stronger justification than the comment at line 364 gives, and worth stating outright.

**Reuse:** this is the same arithmetic as `isActivationDeadlinePassedOnChain` (`packages/babylon-ts-sdk/src/tbv/core/services/deposit/peginState.ts:245`), with the same strict-`>` contract semantics already documented and boundary-tested there. A sibling `isAckDeadlinePassedOnChain({ currentBlock, createdAtBlock, pegInAckTimeout })` in that module would keep both contract mirrors in one place and leave only the throw here.

services/vault/src/copy.ts
```diff
@@ -866,6 +875,18 @@ export const COPY = {
...
+      // this deposit (terms outside the signing device's acceptable range can
+      // never be approved), so no "try again" — support and, once BTC has
+      // moved, the refund path.
+      depositTermsRejected: {
+        title: "Deposit terms not approved",
+        body: "Your signing device cannot approve this deposit's terms, so the deposit cannot continue with the connected wallet. Please contact support. If your Bitcoin has already been sent, it stays recoverable through the refund flow once its timelock opens.",
+      },
+      walletMethodNotSupported: {
+        title: "Wallet action not supported",
+        body: "Your connected wallet can't perform an action this deposit requires. Please reconnect with a supported wallet and try again.",
```
gbarkhatov, line 894: "Please reconnect with a supported wallet and try again" is not a recovery path for an in-flight deposit.

The deposit is bound to the wallet that derived its vault secrets — the WOTS public-key hash is committed on chain, and connecting a different wallet fails the WOTS-hash guard (`COPY.deposit.resume.wotsMismatchError` / `WRONG_WALLET_BODY`) rather than resuming.

Concretely, on the presign surface the most likely source of this copy today is the Ledger vault provider, whose `signPsbt`/`signPsbts` still throw `WALLET_METHOD_NOT_SUPPORTED` (`packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts:419-422`, `TODO(#2219)`) while `approveDepositTerms` at `:301` is implemented and `runDepositorPresignFlow` approves at `:463` before signing in Phase 3. So the sequence a Ledger depositor gets from this PR is: chain reads, full intent ceremony on the device, `signPsbt` throws, then this copy telling them to swap wallets.

For the fresh-deposit path through `mapDepositError` the wording is fine — nothing is committed yet. A resume-specific variant saying this build cannot complete signing on that device would be more honest; failing that, dropping "reconnect with a supported wallet" would at least not send people down an impossible path. Either way, worth saying in the PR description whether this is forward-plumbing for #2219.

services/vault/src/utils/errors/formatting.ts
```diff
@@ -473,37 +477,61 @@ export function formatErrorMessage(error: unknown): string {
 }
 
 /**
- * Format payout signature errors with user-friendly messages
+ * Format payout signature errors with user-friendly messages. Typed
+ * classifications run first (VP RPC, wallet rejection, deposit-terms
+ * rejection, lifecycle refusal), then the cause-walking method-not-supported
+ * check, then message-level matching — mirroring `mapDepositError`'s
+ * precedence so the two mappers cannot disagree on the same error.
```
gbarkhatov: "mirroring `mapDepositError`'s precedence so the two mappers cannot disagree on the same error" claims more than the code delivers.

`mapDepositError` has the cause-walking `isUserCancellation` at step 6 (`depositErrors.ts:252`); this function has only the top-level `isWalletRejectionError` at line 496, which matches the wallet-connector `CONNECTION_REJECTED` code alone. A cancellation one `cause` frame deep — or one carrying EIP-1193 `4001` / viem's `UserRejectedRequestError` — maps to `signingRejected` there and to `PSE.unexpected` here.

Not reachable on today's presign path (nothing in `runDepositorPresignFlow` re-wraps), but the sentence asserts an invariant the code does not provide. Either add the walk here or narrow the claim to describe the ordering only.

services/vault/src/utils/errors/depositErrors.ts
```diff
@@ -140,6 +149,26 @@ export function mapDepositError(err: unknown): DepositErrorContent {
...
+  // stage keeps the copy its generic-message predecessor landed on (the old
+  // message contained "broadcast", so it hit the broadcast bucket below).
+  if (isVaultLifecycleStateError(err) && err.stage === "broadcast") {
+    return ERRORS.broadcastFailed;
+  }
+
+  // 3f. Wallet lacks a required method. Cause-walking, so it must run AFTER
+  // every typed bucket above — an inner unsupported-method code must never
+  // override a meaningful outer wallet/VP/contract error.
+  if (isWalletMethodNotSupported(err)) {
```
gbarkhatov, line 187: The placement does not deliver this invariant for user rejections.

Step 1 above (`isWalletRejectionError`) matches only the wallet-connector `CONNECTION_REJECTED` code (`userCancellation.ts:76-82`). The other two typed cancellation signals — EIP-1193 `4001` and viem's `UserRejectedRequestError` — live in `isUserCancellationFrame` (`:105-107`) and are only reachable via `isUserCancellation` at step 6, *below* this check. So an error whose outer frame is a typed user rejection and whose chain carries an inner `WALLET_METHOD_NOT_SUPPORTED` maps to "Wallet action not supported" instead of "Signing rejected" — exactly the override this comment forbids, and the rejection is the more accurate of the two.

Low reachability, but it is a documented invariant the code does not hold. Hoisting a non-walking `isUserCancellationFrame(err)` above this block would close it without disturbing the cause-walk order.

gbarkhatov (dismissed):
Approving. Rebuilding `DepositTerms` from chain rather than storage, and typing the lifecycle refusals so the mappers branch on fields instead of message substrings, is the right shape — and the two presign gates match the contract exactly. Everything below is a follow-up rather than a blocker; the two I would act on before merge are the error-mapping collapse in `formatting.ts` and the gate ordering in `usePayoutSigningState`.

**Independently verified (critical-path #3 second-reviewer statements)**

- The rebuilt terms are cross-checked against the separately-built `PayoutSigningContext` by `assertDepositTermsMatchSigningContext` before `approveDepositTerms`, and `rebuildDepositTermsCore` re-verifies Gate 0/1. Nothing from browser storage reaches the terms.
- Ack arithmetic is unit-consistent and mirrors the contract: `OnChainVaultData.createdAt` is a block number, `TBVProtocolParams.pegInAckTimeout` a block count, strict `>`, fails closed on read error. Confirmed against `PeginLogic.submitACK` and `PeginLogic.reportExpired` at `vault-contracts-aave-v4 @ 2e87a85a` — see the inline note on the cited line numbers.
- The presign/broadcast status-gate split holds. Sibling enumeration is not status-filtered by the indexer, so relaxing the presign gate to target-only cannot break the contiguous-`htlcVout` assertion, and no presigned byte depends on a sibling's status.
- The broadcast refusal message stays byte-identical, so `mapDepositError`'s `"broadcast"` substring bucket still catches it, and the new typed `3e` branch preserves the same user-visible outcome.
- Recomputing the fee for an already-broadcast Pre-PegIn is safe: `getUtxoInfo` reads the transaction via `getTxInfo`, not the UTXO set, so spent prevouts still resolve.
- Rebuilt `DepositTerms` field types line up with `prepareSigningContext` (`timelockAssert: number`, `protocolFeeRate: bigint` on both sides), so the assert cannot spuriously refuse.
- `supportsDepositApproval` cannot disagree between the hook and `runDepositorPresignFlow` — `forwardDepositApproval` is gated on the same predicate.
- `isWalletMethodNotSupported`'s cause walk terminates on cycles and its depth bound matches its tests; the new `{ cause }` wrappers do not shadow the typed re-throws above them.

**Two things worth an explicit line in the PR description**

1. Both new lifecycle gates sit behind `supportsDepositApproval(wallet)`, so a software-wallet depositor with the same ack-expired deposit gets no status gate, no ack-window gate, and never sees the "Deposit timed out" copy — they run the full ceremony and fail at the VP with generic copy. The broadcast path, which is what actually commits BTC, has no ack-window gate at all. Reasonable to defer, but better as a stated follow-up than an implicit gap.
2. The two `{ cause: error }` additions are the right fix for the cause-walking classifiers, but the full inner chain now reaches `captureFunnelFailure` (Sentry serialises `cause`) and `formatErrorDiagnostics`. Wallet errors do not carry key material today; given the repo's "never log sensitive key material" rule this deserves a conscious sign-off rather than silence.

jonybur (dismissed):
The presign rebuild wires through the same chain-verified path the broadcast resume already uses, and gates it on `supportsDepositApproval` so the software-wallet path stays byte-identical (the test asserting no `depositTerms` key even as `undefined` is a nice guard against a shape-drift regression down the line). Splitting the status gate into `assertBatchLifecycleStatus` with the target/siblings/mode distinction reads clearly and the byte-identical broadcast message keeps `mapDepositError`'s existing bucket working. The ack-window mirror against `createdAt + pegInAckTimeout` with strict `>` matches the contract, and threading the timeout through `getTBVProtocolParams` rather than a constant is the right call. The typed `VaultLifecycleStateError` + reason/stage/role/status/vaultId shape lets the mappers branch on fields instead of substring-matching, and the tests pin the presign message not containing "broadcast" so a future edit can't collapse the two buckets. The follow-ups gbarkhatov raised (bucket coverage in `formatPayoutSignatureError`, gate ordering, guard scope, telemetry for expected refusals, the `ResumeSignContent` terminal seam, the copy phrasing for wallet-method-not-supported, and the `PeginLogic.sol` line-number citation) are all worth doing before or shortly after merge, but none of them block the shape of what's landing.

services/vault/src/utils/errors/formatting.ts
jrwbabylonlab: fixed — typed wrong-wallet + app-version buckets, diagnostics on the generic fallback; not planning the raw-message fallback — #1290 keeps raw Error.message out of the UI

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
jrwbabylonlab: fixed

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
jrwbabylonlab: fixed

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
jrwbabylonlab: fixed

services/vault/src/components/simple/ResumeDepositContent.tsx
jrwbabylonlab: fixed

services/vault/src/services/vault/rebuildDepositTerms.ts
jrwbabylonlab: fixed — symbol + pin; not planning the SDK helper — single caller for now

services/vault/src/copy.ts
jrwbabylonlab: fixed — resume-specific copy + PR note; on main #2269 already wires Ledger signPsbt

services/vault/src/utils/errors/formatting.ts
jrwbabylonlab: fixed

services/vault/src/utils/errors/depositErrors.ts
jrwbabylonlab: fixed

jonybur (approved):
The rebuild-and-cross-check shape on the presign resume path holds up on a second walk: rebuildDepositTermsCore self-verifies the funded tx hash against the on-chain prepeginTxid before trusting any bytes, byte-matches each HTLC output value and scriptPubKey against the funded tx, and assertDepositTermsMatchSigningContext refuses to sign when the approved terms disagree with the version-locked signing context on graph-shaping scalars, canonical keeper/challenger sets, or vault provider membership. Nothing from browser storage reaches the terms.

The typed VaultLifecycleStateError with stage/reason/role/status/vaultId lets the mappers branch on fields instead of substring-matching, which was the drift risk in the previous shape. Gate ordering is right: assertPresignTargetSignable and the ack-window mirror against createdAt + pegInAckTimeout (strict >) run before the mempool fee read, so a stalled deposit's refund copy wins over a transient prevout error. Terminal refusals correctly hide Retry and stay out of funnel telemetry via the isDepositTermsRejectedError / isVaultLifecycleStateError(stage==='presign') check.

The DepositorWalletMismatchError typing closes the substring-matching hazard on the wrong-depositor path, and the { cause } wrappers preserve the cause chain for the cause-walking classifiers without introducing key material into the serialized error. The presign/broadcast status-gate split holds - sibling status never affects HTLC vout ordering, so relaxing the presign gate to target-only cannot break the contiguous-htlcVout assertion.

The CodeRabbit and gbarkhatov follow-ups on inline copy strings and formatter tests are worth folding in as a nit-cleanup, but none of them block the shape landing here.

gbarkhatov (approved):

Issue #2203, opened 2026-08-06
ETH reorg recovery for an orphaned Pre-PegIn registration


jrwbabylonlab:
Split out of #1463, which stays as the decision record and reasoning.

## The scenario

Two things must both happen, which is why this is a fail-safe rather than a feature:

1. An ETH reorg deeper than 8 blocks — never observed in Ethereum's history, and
2. The `submitPeginRequestBatch` transaction is **not** re-included afterwards — unusual, most reorged transactions return to the mempool and land within a block or two.

If both occur after the Pre-PegIn BTC has been broadcast, the on-chain vault record is gone, the indexer drops it too, and the depositor's BTC sits in an HTLC the app can no longer see.

## Agreed approach

Deliberately minimal — a documented manual path, not a built-out product surface.

1. A script scans the rollback window and identifies affected ETH addresses and their Pre-PegIn tx hashes.
2. A **hidden page behind a feature flag**, off by default, enabled only if this ever fires. The user connects their wallet, pastes their Pre-PegIn tx hash, and performs the refund.
3. The FE is given the affected addresses and prompts those users toward the page when their wallet connects.

Feasible because the depositor's secrets are re-derivable from their wallet plus the Pre-PegIn transaction.

## Actions

- [ ] **Protocol team** — provide the script that identifies affected ETH addresses and their Pre-PegIn tx hashes
- [ ] **Web team** — document the recovery mechanism in-repo, share with DG and Jenks for FAQ
- [ ] **Web team** — build the hidden feature-flagged page
- [ ] **Runbook + a practice run** (raised by @vitsalis) — a path nobody has rehearsed is not a path

## Priority

Low, pending bandwidth. Acceptable to defer **as long as the mechanism is documented** — the documentation and runbook matter more than the page, since a manual rescue is workable and an undocumented one is not.

## Context worth keeping

The indexer is Ponder-based and treats a block as final at **65 confirmations** (~13 min). An 8-confirmation gate (~1.6 min) leaves roughly 57 blocks where the UI has moved on but the indexer would still roll the record back. That gap is why this recovery path needs to exist.

gbarkhatov:
## Revised action list after verifying this against the code

Two things changed: one ask in the current list cannot be fulfilled as written, and the dependency question is now settled — this is buildable by the web team alone.

### Remove: the post-hoc protocol-team script

> **Protocol team** — provide the script that identifies affected ETH addresses and their Pre-PegIn tx hashes

[...]

PR #2278, opened 2026-08-19
feat(ts-sdk): reconstruct peg-in parameters from a stranded pre-pegin

gbarkhatov:
# Reconstruct peg-in parameters from a stranded Pre-PegIn

Part 1 of 2 for [#2203](https://github.com/babylonlabs-io/babylon-toolkit/issues/2203). SDK only — no UI, no frontend wiring, no devnet dependency. PR 2 is the wallet-backed signet verification; there is no PR 3, since the runbook it was going to carry ships here.

## Scope

The SDK half of the recovery process that was agreed alongside the 8-confirmation gate — not an attempt to pre-empt or detect a reorg. The gate ([#2266](https://github.com/babylonlabs-io/babylon-toolkit/pull/2266)) stays and is not being second-guessed. The division of labour: the protocol side identifies affected vaults by scanning forked blocks, the frontend builds the refund, and the UI waits until it is actually needed.

## The problem

If an Ethereum reorg orphans a `submitPeginRequestBatch` registration and it is never re-included, the on-chain vault row and the indexer's copy both disappear while the depositor's BTC stays locked in the HTLC. Funds are not at risk — both taproot leaves hardcode the depositor's key — so this is liveness, and the escape hatch is the refund leaf. Taking it needs the exact HTLC parameters, and the refund builder reads them from the row that no longer exists.

## The shape of the answer: verify, do not guess

None of the destroyed fields are depositor input. The contract writes all of them from live state at registration, so they are recoverable — most practically by reading the same getters at the reorg height on the canonical chain, since a reorg spans a handful of blocks and these values only move on a governance action. That assumption is checkable rather than load-bearing: read either side of the reorged range and confirm the values agree.

So the frontend's job is **verification**. Parameters recovered from any off-chain source are untrusted input and must not feed a signed Bitcoin transaction unverified: every HTLC scriptPubKey is rebuilt from the supplied parameters and byte-matched against the funded Pre-PegIn, along with the output values, and the per-vault amounts are inverted from what the transaction actually pays. That check already exists in `rebuildDepositTermsCore`, which is reused verbatim so the judgement has exactly one implementation. Verifying a single supplied parameter set is the same code path with a one-element candidate list; multi-candidate search is the fallback for a responder who has the transaction but not the stamps.

Independently useful, and the reason the wallet half exists at all: re-deriving the vault root proves a pasted Pre-PegIn belongs to the connected wallet, via `sha256(expandAuthAnchor(root))` against the transaction's auth-anchor OP_RETURN. Zero protocol parameters, and it separates "wrong wallet, account or network" — the depositor's to fix — from "right wallet, parameters not found". Ordering is load-bearing: it runs before `expandPerVaultSecrets`, which takes ownership of the root and zeroes it.

## What this adds

`packages/babylon-ts-sdk/src/tbv/core/recovery/`, exported from `@babylonlabs-io/ts-sdk/tbv/core`:

- `deriveHashlocksFromPrePegin` — wallet root, auth-anchor validation, per-vault hashlocks.
- `reconstructPeginParams` — verify supplied parameters against the funded transaction, project `DepositTerms`, invert the amounts.
- `buildPeginParamsCandidates` — assemble the candidate(s).
- `recoveryErrors` — a typed taxonomy that distinguishes a depositor problem from ours, and a partial answer from a trustworthy one.
- `RECOVERY.md` — the responder runbook, moved up from the former part 3.

## Findings that shaped it

- **`vaultCoreVersion` cannot be discriminated by the transaction.** v1 and v2 produce byte-identical HTLC scriptPubKeys, and the amount is inverted using the supplied version's reserve, so both gates pass either way and a wrong value is accepted **in silence** — the refund is unaffected, the reported amount split is wrong. It is supplied, never searched, and `activeVaultCoreVersion()` is the *current* version and wrong for any deposit predating a bump. A test pins this hazard deliberately, asserting the exact shifted amount rather than mere inequality.
- **`timelockRefund` is the only offchain-params scalar reaching the scriptPubKey.** Two params versions sharing it cannot be told apart by a search — another reason to read the stamps rather than enumerate them.
- **Legacy pre-anchor Pre-PegIns are out of scope.** The verifier needs the anchor to prove the sibling set is *complete*, not merely to count outputs, so no other source rescues them. None appear in the sampled signet set.
- **There is no key-rotation ceiling.** An earlier draft of this work claimed a vault whose participants rotated more than once could never be reconstructed. That was wrong: the epochs are stamped from live state like everything else.

## Review rounds

All threads resolved. Four rounds, and the substantive changes came from human review rather than the bots:

- **Removed** `enumerateRosterVersions` — it returned registration keys where scripts are built from sorted RFC-006 operation keys, so every candidate carried a wrong script while reporting a complete space.
- **Removed** the `keyEpochPolicy` axis — the three epochs are stamped independently by three registries, so an all-genesis or all-current policy describes a state the protocol never holds, and a mixed rotation made a recoverable deposit report not-found.
- **Compare matches, do not count them.** Candidates differing only in labels the transaction cannot express both survive; refusing on a count alone rejected a deposit whose refund is fully determined — the common case, not an edge case. Refusal is now reserved for survivors whose projected terms actually differ.
- **Classify failures.** Only a Gate-1 byte mismatch counts as a rejection, via a typed `HtlcOutputMismatchError`; anything else means the candidate was never evaluated and is recorded as a gap, so an incidental failure on the true candidate can no longer hand back a look-alike as a trusted unique answer. An over-reserved candidate is a mismatch, not a gap — that distinction was got wrong on the first attempt and fixed in the round after.
- **Normalise `depositorBtcPubkey`** before it reaches the verifier; a `0x`-prefixed wallet string previously passed derivation and failed every candidate, presenting as "no parameters found".
- **Assert WASM sizing at the boundary** (CLAUDE.md critical path 1), escaping the search rather than counting as a rejection, since a malformed sizing output indicts the binary rather than the candidate.
- **`unresolvedVersions` is a required input** — `[]` is a positive claim that the enumeration was complete, so a caller cannot reach a trusted answer by neglecting to mention its own gaps.

## Tests

36 in the recovery module, all against real WASM — a mocked oracle would prove nothing, since the job is reproducing genuine connector bytes from the wrong end.

Differential recovery of a **two-vault** deposit (mandatory per CLAUDE.md for chain logic) and a single-vault v1 case; wrong-roster and wrong-provider rejection; identical-terms matches returning rather than refusing, and differing-terms matches still refusing; an unevaluable candidate raising the incomplete-space error instead of vanishing; an over-reserved candidate rejected with no gaps reported; a `0x`-prefixed pubkey working through both halves; legacy pre-anchor and ambiguous-anchor transactions refused; the wallet never called when the transaction has no usable anchor. One separate file mocks the oracle to exercise the sizing-integrity guards, kept apart so the main suite stays honest.

`lint`, `tsc --noEmit` and the full suites pass in both packages — **1582** SDK tests, **3387** vault tests. The vault run matters because `assertWasmPeginSizing.ts` is on the deposit path; the change there is `Error` → a subclass, so anything catching `Error` or matching messages is unaffected.

## What PR 2 adds

Historical contract reads (nothing in `clients/eth` accepts a block number today), candidate assembly resolving participant **operation** keys at the stamped epochs, an adapter from `DepositTerms` into `buildRefundPsbt`'s `PrePeginParams`, and the signet differential — reconstruct a known vault blind and diff every field against the row that still exists there, then build the refund PSBT twice and assert byte-identical before any broadcast.

## Review requirements

Touches CLAUDE.md critical paths 1, 3 and 8: WASM outputs feeding a signed transaction, and value reconstruction from a source untrusted by absence. Two reviewers, and the author walks every line unaided. No frozen derivation primitive is reimplemented — `deriveVaultRoot`, `expandAuthAnchor` and `expandPerVaultSecrets` are called, never reproduced.


packages/babylon-ts-sdk/src/tbv/core/recovery/deriveHashlocksFromPrePegin.ts
```diff
@@ -0,0 +1,131 @@
...
+ * needs zero protocol parameters, and is what separates "wrong wallet,
+ * account or network" from "right wallet, no candidate matched" — otherwise
+ * both surface as a fruitless parameter search.
+ *
+ * @throws {UnanchoredPrePeginError} If the transaction carries no single,
+ *   unambiguous auth-anchor OP_RETURN.
+ * @throws {VaultRootMismatchError} If the derived root does not commit to the
+ *   transaction's anchor.
+ */
+export async function deriveHashlocksFromPrePegin(
```
jrwbabylonlab, line 75: all the pegin information can be fetched from eth. why bother to do the derive ourself?

packages/babylon-ts-sdk/src/tbv/core/recovery/deriveHashlocksFromPrePegin.ts
```diff
@@ -0,0 +1,131 @@
...
+): Promise<DeriveHashlocksFromPrePeginResult> {
+  const { wallet, depositorBtcPubkey, fundedPrePeginTxHex } = input;
+
+  const cleanTxHex = stripHexPrefix(fundedPrePeginTxHex);
+  const anchor = findAuthAnchorOpReturn(cleanTxHex);
+  if (anchor === undefined) {
+    throw new UnanchoredPrePeginError(
+      `Pre-PegIn carries no single, unambiguous auth-anchor OP_RETURN, so the ` +
+        `number of HTLC outputs it funds cannot be determined. Legacy ` +
+        `pre-anchor deposits are not recoverable by this path.`,
```
jrwbabylonlab, line 86: i don't think we care about the legacy pre-anchor deposits? can u check if there are any vaults on testnet will be impacted if we do not care about legacy ones?

packages/babylon-ts-sdk/src/tbv/core/recovery/enumerateRosterVersions.ts
```diff
@@ -0,0 +1,113 @@
+/**
+ * Enumerate every historical vault-keeper and universal-challenger roster
+ * version (#2203).
+ *
+ * A reorg destroys the vault row's roster version stamps, but both rosters are
```
jrwbabylonlab: i'm a bit confused on the purpose of this PR. 
we not trying to deal with vaults that's being re-org. we simply try to wait for 8 blocks then broadcast the pre-pegins.
so there will never be a case in which we need to look backwards and try recover the already broadcasted prepegins

packages/babylon-ts-sdk/src/tbv/core/recovery/enumerateRosterVersions.ts
```diff
@@ -0,0 +1,146 @@
...
+ * costs one sequential RPC round trip. A corrupt or unexpectedly large value
+ * would otherwise drive an open-ended read loop against the node, so the
+ * enumeration refuses rather than starts one. Real rosters sit in the single
+ * digits; this is a backstop, not a working limit.
+ */
+const MAX_ENUMERABLE_ROSTER_VERSIONS = 256;
+
+export interface RosterVersionSnapshot {
+  version: number;
+  /** x-only BTC pubkeys, `0x` stripped, in the contract's stored order. */
```
jrwbabylonlab: `btcPubKey` is the operator genesis key, but scripts are built from operation keys, and the deposit path sorts them (`resolveParticipantKeys.ts:137`).

so wiring `snapshot.btcPubkeys` straight into `vaultKeeperBtcPubkeys` gives a wrong script on every candidate, and we return NotFound with `unresolvedLabels: []`, i.e. affirmatively claiming the space was complete. the two exported halves do not compose yet.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
```diff
@@ -0,0 +1,362 @@
...
+      });
+
+      survivors.push({
+        candidate,
+        terms,
+        peginAmounts,
+        authAnchorHash: anchor.hash,
+        candidatesTried: candidates.length,
+      });
+    } catch (err) {
```
jrwbabylonlab, line 392: allow-by-default with one exception, so a non-mismatch throw on the *true* candidate (wasm init, a bad roster key reaching the connector) removes it from the space at trial time.

that removal never reaches `unresolvedVersions`, so `IncompleteSpaceError` cannot fire and a look-alike is returned as a trusted unique survivor. fail-open, against the invariant the module is built on.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
```diff
@@ -0,0 +1,362 @@
...
+  const unresolvedLabels = unresolvedVersions.map(describeUnresolvedVersion);
+
+  if (survivors.length === 0) {
+    throw new PeginParamsNotFoundError(
+      candidates.length,
+      rejections,
+      unresolvedLabels,
+    );
+  }
+  if (survivors.length > 1) {
```
jrwbabylonlab: survivors are counted, never compared. `DepositTerms` carries no version labels, so when nobody rotated, genesis and current are script-identical, both survive, and we refuse a deposit whose refund is fully determined.

that is the common case rather than the edge case. comparing the projected terms before refusing would let it through.

packages/babylon-ts-sdk/src/tbv/core/recovery/peginParamsCandidates.ts
```diff
@@ -0,0 +1,182 @@
...
+ * is why {@link KeyEpochPolicy} has exactly two values and why they apply to
+ * the whole participant set at once: a vault whose participants rotated more
+ * than once since creation cannot be reconstructed.
+ *
+ * @module recovery/peginParamsCandidates
+ */
+
+import type { Address } from "viem";
+
+/**
```
jrwbabylonlab: `vpKeyEpoch`, `appKeeperKeyEpoch` and `ucKeyEpoch` are stamped separately by three registries, so mixed states are normal rather than impossible.

vp rotated before the deposit and a uc after it: genesis gets the vp wrong, current gets the uc wrong, neither matches, and a recoverable deposit reports NotFound.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
```diff
@@ -0,0 +1,362 @@
...
+
+      const { offchainParams: params, participants } = candidate;
+      const terms = await rebuildDepositTermsCore({
+        vaultCoreVersion: candidate.vaultCoreVersion,
+        siblings: hashlocks.map((hashlock, i) => ({
+          hashlock,
+          amount: peginAmounts[i],
+        })),
+        fundedPrePeginTxHex: cleanTxHex,
+        depositorBtcPubkey,
```
jrwbabylonlab: `depositorBtcPubkey` reaches the verifier un-stripped, but `deriveHashlocksFromPrePegin` documents `0x` as optional and strips it.

pass the same wallet string to both and step 1 succeeds while step 2 fails every candidate. the existing verifier caller normalises with `processPublicKeyToXOnly` first. note the tx hex two lines up is stripped, so this reads as an oversight.

packages/babylon-ts-sdk/src/tbv/core/recovery/index.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
jrwbabylonlab, line 1: scope: sdk-only is right, we do not need this wired into the frontend yet.

but can the runbook come in this PR instead of PR 3? a short RECOVERY.md next to these methods: call order, what a responder gathers, how to tell NotFound / Ambiguous / wrong-wallet apart, and the rotation limit. the emergency can land between PR 1 and PR 3, and docs were the firm half of the original decision.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
```diff
@@ -0,0 +1,362 @@
...
+
+  for (const candidate of candidates) {
+    try {
+      const cacheKey = sizingCacheKey(candidate);
+      let sizing = sizingCache.get(cacheKey);
+      if (sizing === undefined) {
+        sizing = await computePeginSizing(candidate);
+        sizingCache.set(cacheKey, sizing);
+      }
+      const reserve =
```
jrwbabylonlab, line 343: a wrong supplied version is silently accepted. v1 and v2 give byte-identical connector scriptPubKeys but different reserves (your own tests at 471-492 and 494-515), and `peginAmount` is inverted using the supplied version reserve, so the value check passes by construction and the script check cannot see it.

a responder reaching for `activeVaultCoreVersion()` on an older deposit gets a sole match, no error, and `peginAmounts` / `depositorClaimValue` / `peginMaxFee` all off by the reserve delta. the refund tx is unaffected, but the reported terms are wrong, and there is no guard and no negative test.

gbarkhatov:
## Review response, and a finding that changes the design

Thanks for the review — several of these are correct and two of them break the common case rather than an edge case. Before the per-thread replies, one discovery that reframes most of them.

### The destroyed stamps are in the orphaned block's log

`PegInSubmitted` carries every field a reorg destroys:

```
vaultId, peginTxHash, depositor, vaultProvider, amount,
vaultCoreVersion, universalChallengersVersion, appVaultKeepersVersion,
offchainParamsVersion, proverCircuitVersion,
hashlock, htlcVout, unsignedPrePeginTx, depositorWotsPkHash,
vpKeyEpoch, appKeeperKeyEpoch, ucKeyEpoch
```

That includes `vaultCoreVersion`, which the Bitcoin transaction provably cannot discriminate, and the three key epochs, which this PR documented as permanently unrecoverable. Both claims were wrong.

Since the recovery script scans forked blocks, and forked blocks are retrievable from explorer infrastructure even though standard JSON-RPC will not serve them, the parameters should be **read from the log and verified**, not enumerated and guessed.

One caveat that survives: a parameter set arriving from an off-chain script is untrusted input, and it must not feed a signed Bitcoin transaction unverified. Rebuilding each HTLC scriptPubKey and byte-matching it against the funded Pre-PegIn is what makes it safe, and that check is already in `rebuildDepositTermsCore`. Verifying one supplied parameter set is the existing search with a single-element candidate list, so the verification path stays and the enumeration around it mostly goes.

### What changes

**Removed** — `enumerateRosterVersions` (returns registration keys where scripts need sorted RFC-006 operation keys, and no longer needed once the epochs come from the log), and the `keyEpochPolicy` axis (the three epochs are stamped independently by three registries, so a globally-consistent policy describes a state the protocol never holds).

**Fixed** — non-mismatch throws classified as unresolved candidates instead of silently dropped; uniqueness decided by comparing projected terms rather than counting survivors; `depositorBtcPubkey` normalised before it reaches the verifier.

**Added** — `RECOVERY.md` alongside the methods, in this PR rather than PR 3.

**Corrected in the docs** — the rotation limit, and the claim that `vaultCoreVersion` must be supplied by a responder.

### Open with the protocol side

Whether logs/receipts are retrievable for forked blocks, or only the blocks and their transactions. Calldata alone carries `hashlock`, `htlcVout`, `depositorBtcPubKey`, `unsignedPrePeginTx` and the vault provider, but **not** the version stamps, because the contract stamps those from live state at registration. That answer decides how much of the enumeration has to stay as a fallback.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
```diff
@@ -0,0 +1,284 @@
...
+        sizing.depositorClaimValue + sizing.peginMaxFee + sizing.p2aAnchorValue;
+      const peginAmounts = observedHtlcValues.map((value) => value - reserve);
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in daedb42c, though only partly — worth being explicit about the residue.

`unresolvedVersions` is now a required input, and a sole match with a non-empty list raises `PeginParamsIncompleteSpaceError` rather than being returned as the answer. Required rather than optional so a caller cannot reach a trusted result by neglecting to mention its own gaps.

What that did not cover, raised in review since: a candidate that throws for a non-mismatch reason is removed from the space at trial time and never recorded as unresolved, so the guard cannot fire. Being addressed by classifying non-mismatch throws as unresolved rather than as rejections.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
```diff
@@ -0,0 +1,284 @@
...
+      numUcs,
+      params.minPeginFeeRate,
+    ),
+    peginP2aAnchorOutput(candidate.vaultCoreVersion),
+  ]);
+  return {
+    depositorClaimValue,
+    peginMaxFee,
+    p2aAnchorValue: anchorOutput?.value ?? 0n,
+  };
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in daedb42c.

`computePeginSizing` now asserts the WASM outputs before consuming them — claim value and fee strictly positive, anchor non-negative — and raises `PeginSizingIntegrityError`, which is deliberately re-thrown out of the search loop instead of being counted as a candidate rejection: a malformed sizing output indicts the binary rather than the candidate, and swallowing it would bury a broken engine under a not-found result. Covered by a test file that mocks the oracle, kept separate so the main suite still runs entirely against the real engine.

The plausibility band on the implied reserve is deliberately not duplicated here — it depends on the candidate's own `minPeginFeeRate`, so a candidate can legitimately fail it, and the verifier applies it where it belongs.

On the `?? 0n`: that is not a silent fallback but the facade's contract — `peginP2aAnchorOutput` returns `null` for graph versions whose PegIn carries no anchor, never a zero-valued placeholder, and it is the same reading `rebuildDepositTermsCore` takes.

packages/babylon-ts-sdk/src/tbv/core/recovery/enumerateRosterVersions.ts
```diff
@@ -0,0 +1,113 @@
...
+      const roster = await read(version);
+      if (roster.length === 0) {
+        throw new Error(
+          `${label} version ${version} resolved to an empty roster`,
+        );
+      }
+      snapshots.push({
+        version,
+        btcPubkeys: roster.map((entry) => stripHexPrefix(entry.btcPubKey)),
+      });
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in daedb42c by routing every contract-returned key through `assertOnChainBtcPubkey`, the package's single x-only validator — length, hex form and secp256k1 curve membership — so a malformed roster is reported to the skip observer with an actionable message instead of surfacing later as an opaque unmatched candidate.

Flagging that review has since found a deeper problem in the same function: it returns registration keys, where the scripts are built from sorted RFC-006 operation keys. `enumerateRosterVersions` is being removed rather than repaired, so this fix will go with it.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
```diff
@@ -0,0 +1,362 @@
...
+  }
+  // Reported after the ambiguity check: two survivors are the more specific
+  // diagnosis, and the caller learns of the gaps from that error's labels.
+  if (unresolvedLabels.length > 0) {
+    throw new PeginParamsIncompleteSpaceError(
+      describePeginParamsCandidate(survivors[0].candidate),
+      unresolvedLabels,
+    );
+  }
+  return survivors[0];
```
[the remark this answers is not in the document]
gbarkhatov: Agreed, and it is the same gap as the allow-by-default catch raised elsewhere in review: the completeness claim is only as good as the caller's account of its own failures, and a candidate that throws for a non-mismatch reason is dropped without ever being recorded.

Fix is to classify rather than swallow — a Gate-1 byte mismatch is a genuine rejection, anything else is an unresolved candidate feeding the same fail-closed path as an unreadable version.

The broader answer is that the destroyed stamps are recoverable directly from the orphaned block's `PegInSubmitted` log, which removes most of the enumeration this guard exists to protect.

packages/babylon-ts-sdk/src/tbv/core/recovery/deriveHashlocksFromPrePegin.ts
gbarkhatov: Fair, and more so than when this was written.

The premise was that a reorg takes the row with it, so nothing is reachable by `vaultId`. What I had missed is that `PegInSubmitted` carries every destroyed field — amount, all four version stamps, hashlock, htlcVout, `unsignedPrePeginTx`, and the three key epochs. If the recovery script reads the orphaned block's logs, the hashlock is simply given and this derivation is not needed for the refund.

Two narrower things it still buys:

1. It binds the transaction to the connected wallet. `sha256(expandAuthAnchor(root))` must equal the OP_RETURN payload, which proves the pasted Pre-PegIn belongs to whoever is connected — an authorisation check for a page where the user supplies the hash, and a way to tell "wrong wallet, account or network" from "right wallet, parameters not found".
2. It is the fallback if logs turn out not to be retrievable for forked blocks while the raw transactions are. Calldata carries the hashlock and htlcVout but not the version stamps, since the contract stamps those from live state.

Which of those matters depends on the open question with the protocol side about receipt availability for forked blocks.

packages/babylon-ts-sdk/src/tbv/core/recovery/deriveHashlocksFromPrePegin.ts
gbarkhatov: Checked: of 56 signet vaults mined from local e2e run recordings, all 56 carry the auth-anchor OP_RETURN and none are legacy pre-anchor. That is a sample rather than an exhaustive indexer query — strong, but say the word and I will run the real query to make it conclusive.

One clarification on what the anchor is doing here, since "do we care about legacy" is not quite the question it answers: it is the only structural signal for how many HTLC outputs the transaction funds, so without it the vault count would have to be guessed, and guessing it derives the wrong number of hashlocks. Refusing is not a policy choice about legacy deposits so much as declining to guess.

If the script supplies the decoded `PegInSubmitted` log, htlcVout comes from there and the dependency disappears anyway.

packages/babylon-ts-sdk/src/tbv/core/recovery/enumerateRosterVersions.ts
gbarkhatov: Settled in the thread since, so recording the outcome here: the 8-confirmation gate stays and is not being second-guessed, the FE owns the refund, and the protocol side provides a script that scans forked blocks for affected users.

So this is not an attempt to pre-empt or detect a reorg — it is the SDK half of the recovery process that was agreed alongside the gate, with the UI deliberately deferred until it is needed. The description and #2203 have been updated to say that plainly, since the original framing invited exactly this reading.

packages/babylon-ts-sdk/src/tbv/core/recovery/enumerateRosterVersions.ts
gbarkhatov: Confirmed, and it is worse than a wiring mismatch. `resolveParticipantKeys` produces `vaultKeeperOperationKeysSorted` — RFC-006 operation keys, sorted — whereas this returns registration keys in the contract's stored order. Both halves are wrong, and the consequence you name is the dangerous part: NotFound with `unresolvedLabels: []` affirmatively claims the space was complete while the true key set was never tried.

Resolution is to remove `enumerateRosterVersions` rather than fix it. The three key epochs are in the `PegInSubmitted` log, so a responder resolves operation keys at the exact epochs instead of enumerating roster versions and guessing a policy.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
gbarkhatov: Valid, and the more serious of the two completeness holes.

Allow-by-default means an incidental throw on the true candidate — WASM init, a bad roster key reaching the connector — removes it from the space at trial time and is never recorded, so the incomplete-space guard cannot fire and a look-alike is returned as a trusted unique survivor. Fail-open, against the invariant the module is built on, as you say.

Fix is to classify instead of swallow: a Gate-1 byte mismatch is a genuine rejection, anything else is an unresolved candidate and feeds the same fail-closed path as a version that could not be read. Same family as the bot finding on line 361.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
gbarkhatov: This is the one I am most glad you caught, because the failure lands on the common case rather than the edge case: with no rotation, genesis and current resolve to the same keys, both survive, and we refuse a deposit whose refund is fully determined.

Comparing the projected terms is the right rule. Every survivor byte-matched the same funded outputs, so they necessarily agree on each HTLC scriptPubKey and on each output's total value — if they also project identical terms there is nothing ambiguous and one should be returned. Refusal belongs only where survivors genuinely disagree, which is the split of the HTLC value and the version labels.

That becomes the uniqueness rule.

packages/babylon-ts-sdk/src/tbv/core/recovery/peginParamsCandidates.ts
gbarkhatov: Correct, and the assumption behind `keyEpochPolicy` is simply wrong: three registries stamp the three epochs independently, so a mixed state is ordinary rather than impossible, and your VP-before / UC-after example makes a recoverable deposit report NotFound.

Being removed rather than repaired. `PegInSubmitted` carries `vpKeyEpoch`, `appKeeperKeyEpoch` and `ucKeyEpoch` separately, so the epochs are per-participant inputs, not an axis with two globally-consistent settings.

This also retires the "a vault whose participants rotated more than once can never be reconstructed" limitation, which was wrong and is being removed from the docs and from #2203.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
gbarkhatov: Confirmed. `rebuildDepositTermsCore` receives `depositorBtcPubkey` verbatim — only the hashlock and the tx hex are stripped — so passing the same `0x`-prefixed wallet string to both halves lets the derivation succeed and then fails every candidate, which presents as "no parameters found" rather than as a normalisation bug.

Will normalise with `processPublicKeyToXOnly` the way the existing verifier caller does, and add a case that puts the same prefixed string through both halves. The current tests only ever use bare hex, which is exactly why this survived.

packages/babylon-ts-sdk/src/tbv/core/recovery/index.ts
gbarkhatov: Agreed, and for the reason you give: documentation was the firm half of the original decision, and the emergency can land between PR 1 and PR 3.

A `RECOVERY.md` next to these methods will land in this PR — call order, what a responder needs to gather, how to tell NotFound from Ambiguous from wrong-wallet, and the real limits. It will also carry the corrected input path, since the destroyed stamps come from the orphaned block's `PegInSubmitted` log rather than from enumeration.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
gbarkhatov: Valid, and there is no guard because the transaction carries no evidence either way: v1 and v2 produce byte-identical connector scriptPubKeys, and `peginAmount` is inverted using the supplied version's reserve, so the value check passes by construction and the script check cannot see it. Supplying it was the least-bad option once enumerating it could only ever yield ambiguity — but you are right that a responder reaching for `activeVaultCoreVersion()` on an older deposit gets a confident wrong answer with nothing flagged.

`PegInSubmitted` carries `vaultCoreVersion`, so the real fix is to take it from the log rather than have a responder guess it. Adding the negative test you are implicitly asking for — one that pins the silent-wrong-answer behaviour, so the hazard is visible in code and not only in prose.

packages/babylon-ts-sdk/src/tbv/core/recovery/enumerateRosterVersions.ts
```diff
@@ -0,0 +1,146 @@
...
+ * matched".
+ */
+export type OnSkippedRosterVersion = (version: number, error: Error) => void;
+
+async function enumerateVersions(
+  latestVersion: number,
+  read: (version: number) => Promise<{ btcPubKey: Hex }[]>,
+  label: string,
+  onSkippedVersion?: OnSkippedRosterVersion,
+): Promise<RosterVersionSnapshot[]> {
```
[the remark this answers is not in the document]
gbarkhatov: The principle is right and the diagnosis is exact: an *optional* observer means a caller can end up with no record of its own gaps, then pass `unresolvedVersions: []` and be handed a sole match from an incomplete space. The completeness claim was only ever as strong as the caller's diligence, which is not a guarantee.

Not fixing it here, though, because `enumerateRosterVersions` is being removed rather than repaired — it returns registration keys where the scripts are built from sorted RFC-006 operation keys, so it answers the wrong question regardless of how well its failures are reported, and the key epochs it exists to work around are available directly from the orphaned block's `PegInSubmitted` log.

The design point survives the deletion and is being carried into the replacement: an unresolved-version record has to be **part of the return value**, not an optional side channel. Same conclusion the review reached on the search loop, where a candidate that throws for a non-mismatch reason is currently dropped without being recorded.

packages/babylon-ts-sdk/src/tbv/core/recovery/enumerateRosterVersions.ts
gbarkhatov: Fixed in 8711a253 by removing the file. Both halves were wrong — registration keys where the connector needs operation keys, and contract order where the deposit path sorts — and once the epochs come from `PegInSubmitted` there is nothing left for it to do.

`ParticipantKeySetCandidate` now states the contract explicitly: sorted RFC-006 operation keys at each participant's own epoch, i.e. what `resolveParticipantKeysAtEpochs` returns, with a note that registration keys are not interchangeable because they only coincide while nobody has rotated.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
gbarkhatov: Fixed in 8711a253.

`assertEncodedHtlcOutputsMatch` now throws a typed `HtlcOutputMismatchError`, and the catch classifies on it: a byte mismatch is the transaction positively disagreeing with the candidate and stays a rejection, while anything else means the candidate was never evaluated and is recorded as a gap alongside the unreadable versions. So an incidental throw on the true candidate now forces `PeginParamsIncompleteSpaceError` instead of quietly promoting a look-alike.

Put the error in `assertWasmPeginSizing.ts` rather than `recoveryErrors.ts` — the suggestion in the bot thread had the primitive importing from the recovery module, which inverts the dependency.

Covered by a test that puts an unusable key in one candidate and asserts the gap is reported rather than counted as a rejection.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
gbarkhatov: Fixed in 8711a253. Matches are now compared, not counted.

Survivors all byte-matched the same funded outputs, so they already agree on every HTLC scriptPubKey and total value; the rule is now that identical projected terms are not ambiguity, and every equivalent candidate comes back in `matchedCandidates`. Refusal is reserved for survivors that genuinely disagree — same `timelockRefund` and participants, different fee rates, so the HTLC value splits differently.

Both directions are tested: identical-content versions return a match, and versions differing in `protocolFeeRate` still raise `PeginParamsAmbiguousError`.

packages/babylon-ts-sdk/src/tbv/core/recovery/peginParamsCandidates.ts
gbarkhatov: Fixed in 8711a253 by removing `keyEpochPolicy` entirely.

You are right that a mixed state is ordinary, and the axis encoded the opposite. Epochs are now per-participant inputs taken from `PegInSubmitted`, which carries `vpKeyEpoch`, `appKeeperKeyEpoch` and `ucKeyEpoch` separately, so the vp-rotated-before / uc-rotated-after case resolves correctly instead of failing both ways.

The "rotated more than once cannot be reconstructed" limitation is gone from the module docs, the PR description and #2203. It was wrong.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
gbarkhatov: Fixed in 8711a253. `depositorBtcPubkey` is normalised with `processPublicKeyToXOnly` before it reaches the verifier, matching what the existing caller does.

Added the test that was missing: the same `0x`-prefixed string through both halves. As you say, the old tests only ever used bare hex, which is exactly why this survived — the failure presented as "no parameters found" rather than as a normalisation bug, which is the worst way for it to show up.

packages/babylon-ts-sdk/src/tbv/core/recovery/index.ts
gbarkhatov: Done in 8711a253 — `RECOVERY.md` sits beside the methods.

Covers the call order, what a responder gathers and where each item comes from, a table mapping every typed error to what it means and what to do about it, and the limits. It leads with re-submitting the registration rather than refunding, since that keeps the deposit and the contract does not block it, with refund as the fallback once `t_stale` has passed.

It also ends by saying the path has never been rehearsed, and that a practice run on signet — reconstruct a known vault blind, diff against the row that still exists there — is worth more than the document.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
gbarkhatov: Addressed in 8711a253, though not by adding a guard, because there cannot be one: the transaction carries no evidence either way.

What landed is the negative test you were implicitly asking for. It asserts that a wrong supplied version produces a sole match with a shifted split and no error — deliberately pinning wrong behaviour, so the hazard lives in code rather than only in prose and a reviewer has to look at it.

The real fix is upstream: `PegInSubmitted` carries `vaultCoreVersion`, so it is read rather than guessed. Both the module docs and `RECOVERY.md` now say that plainly, including that `activeVaultCoreVersion()` is the current version and is wrong for any deposit predating a bump — which was the exact trap you described.

packages/babylon-ts-sdk/src/tbv/core/recovery/deriveHashlocksFromPrePegin.ts
gbarkhatov: Answered, and the scope moved in your direction in 8711a253.

With the parameters read from `PegInSubmitted`, the hashlock is given and this derivation is not on the critical path for a refund. What it still does is bind the transaction to the connected wallet — `sha256(expandAuthAnchor(root))` against the OP_RETURN — which is the authorisation check for a page where the user supplies the hash, and the thing that separates "wrong wallet, account or network" from "right wallet, parameters not found". `RECOVERY.md` documents it as step 1 in exactly that role, rather than as the source of the parameters.

It is also the fallback if forked-block logs turn out not to be retrievable while the raw transactions are — still open with the protocol side.

packages/babylon-ts-sdk/src/tbv/core/recovery/deriveHashlocksFromPrePegin.ts
gbarkhatov: Answered: of 56 signet vaults mined from local e2e recordings, all 56 carry the auth-anchor OP_RETURN and none are legacy pre-anchor. A sample rather than an exhaustive query — say the word if you want the real indexer query run.

Either way the dependency is now smaller than it was. The anchor was the only structural signal for the HTLC count, so refusing without it was declining to guess rather than a policy on legacy deposits; with `htlcVout` coming from `PegInSubmitted`, a responder does not need it. `RECOVERY.md` records that legacy pre-anchor deposits cannot use step 1 and should take `htlcVout` from the log instead.

packages/babylon-ts-sdk/src/tbv/core/recovery/reconstructPeginParams.ts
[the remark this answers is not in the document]
gbarkhatov: Fixed in 8711a253, along the lines suggested.

`assertEncodedHtlcOutputsMatch` now throws a typed `HtlcOutputMismatchError`, the search tracks unevaluated candidates separately from rejections, and the catch classifies on the type — sizing-integrity failures still escape, byte mismatches count as rejections, everything else is recorded as a gap and feeds the same fail-closed path as an unreadable version.

One deviation: the error type lives in `assertWasmPeginSizing.ts` rather than `recoveryErrors.ts`, since a primitive importing from the recovery module would invert the dependency direction.

packages/babylon-ts-sdk/src/tbv/core/recovery/enumerateRosterVersions.ts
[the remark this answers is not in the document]
gbarkhatov: Resolved by deletion in 8711a253 — `enumerateRosterVersions` is gone, so the optional-observer hole with it.

The principle was right and has been carried into the search: an unresolved record is now part of the return path rather than an optional side channel. A candidate that cannot be evaluated is tracked alongside the caller-declared unresolved versions and forces `PeginParamsIncompleteSpaceError` on a sole match, so a gap can no longer go unrecorded and be mistaken for a complete space.

packages/babylon-ts-sdk/src/tbv/core/recovery/RECOVERY.md
```diff
@@ -0,0 +1,65 @@
...
+**3. Build and broadcast the refund** — the normal `buildRefundPsbt` / `buildAndBroadcastRefund` path, from the returned terms.
+
+If the log is unavailable, step 2 becomes a search: enumerate the version-keyed reads that survive (`getOffchainParamsByVersion`, `getVaultKeepersByVersion`, `getUniversalChallengersByVersion`) and pass the product. Record every version you could not read in `unresolvedVersions` — see below.
+
+## Reading the failures
+
+| Error | Meaning | Do |
+| --- | --- | --- |
+| `VaultRootMismatchError` | Wrong wallet, account or network. The root is bound to all three, not just the seed. | Ask the depositor to reconnect on the exact account and network they deposited with. Not our bug. |
+| `UnanchoredPrePeginError` | No single auth-anchor OP_RETURN, so the HTLC count is unknowable. | Take `htlcVout` from the log instead. Legacy pre-anchor deposits cannot use step 1. |
```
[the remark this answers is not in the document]
gbarkhatov: Right, and the row was wrong in a way the suggested wording still understates — corrected, going further than proposed.

It is not only that no `htlcVout` override exists: `rebuildDepositTermsCore` requires the anchor to prove the sibling set is *complete*, not merely to count the outputs. So supplying the count from the log would not rescue an unanchored transaction either. Both steps refuse it, and legacy pre-anchor deposits are out of scope for this module today.

packages/babylon-ts-sdk/src/tbv/core/recovery/deriveHashlocksFromPrePegin.ts
gbarkhatov: Correction to my previous reply on this thread: I said `htlcVout` from `PegInSubmitted` removes the dependency on the anchor. That is wrong.

The anchor is not only the HTLC count — `rebuildDepositTermsCore` uses it to prove the sibling set is complete, so an unanchored transaction is refused by the verifier regardless of what the log supplies. Legacy pre-anchor deposits are out of scope for this module, full stop, and `RECOVERY.md` now says that rather than offering a fallback that does not exist.

The 56-of-56 finding is unchanged: no legacy vaults in the sampled signet set, so nothing is affected today.

gbarkhatov:
## Second review round — all four addressed

Three of these could not be posted inline, so answering them here.

### Non-positive inverted amounts were misclassified (major) — fixed

Correct, and it was a regression introduced by the classification change in the previous round rather than a pre-existing flaw. `rebuildDepositTermsCore` rejects `sibling.amount <= 0n` with a generic `Error`, which the new catch read as "never evaluated" and recorded as a gap — so in a fallback search every over-reserved candidate would have become a phantom gap and refused an otherwise sound match.

A candidate whose reserve exceeds the funded output is the transaction positively disagreeing with it, which is a mismatch. The amounts are now checked before the verifier is called and raise `HtlcOutputMismatchError`, so this is the one piece of discriminating the value side genuinely does. Pinned by a test asserting a not-found with **no** gaps reported.

### `unevaluated` was unbounded — fixed

Split as suggested: the count stays exact and drives the fail-closed decision, only the labels reaching the error message are capped at `MAX_REPORTED_REJECTIONS`. Worth noting the fix above also shrinks this a great deal, since most non-matches are now rejections rather than unevaluated candidates — but the cap is right regardless.

### `not.toBe(trueAmount)` was too weak — fixed

Agreed; it would have passed for a negative amount or any wrong magnitude, which is a poor way to pin a hazard. The assertion is now the exact expected value, `trueAmount + reserveV1 - reserveV2`, computed from `computeMinClaimValue` and `computeMinPeginFee` at both versions, so a change in the delta's direction or size fails the test.

### The `UnanchoredPrePeginError` row — fixed, and further than proposed

The row was wrong, and the suggested replacement still understated it. It is not only that no `htlcVout` override exists: `rebuildDepositTermsCore` needs the anchor to prove the sibling set is **complete**, not merely to count the outputs, so supplying the count from the log would not rescue an unanchored transaction either. Both steps refuse it, and legacy pre-anchor deposits are out of scope for this module.

That also corrects something stated in the previous round on the legacy-deposits thread, where I claimed the log removed the dependency. It does not.

---

`lint`, `tsc --noEmit` and the full suites pass in both packages — 1582 SDK tests, 3387 vault tests.

gbarkhatov:
All four second-round findings landed in `2b84b420`.

- Over-reserved candidates raise `HtlcOutputMismatchError` before the verifier is reached, so the value bound reads as a mismatch rather than a phantom gap — with a test asserting a not-found carrying **no** gaps.
- `unevaluated` keeps an exact count for the fail-closed decision and caps only the labels that reach the message.
- The `vaultCoreVersion` hazard test pins the exact delta, `trueAmount + reserveV1 - reserveV2`, instead of an inequality that a negative amount would have satisfied.
- The `UnanchoredPrePeginError` row in `RECOVERY.md` now says both steps refuse an unanchored transaction, since the verifier needs the anchor to prove sibling completeness rather than only to count outputs.

Every review thread is resolved. The one thing still open is not a code question: whether logs/receipts are retrievable for forked blocks, or only the blocks and their transactions. If logs are available the remaining enumeration can go entirely and this module reduces to verify-then-refund; if only raw transactions are, the search stays as the fallback for the version stamps, since calldata carries the hashlock and htlcVout but not the stamps.

jrwbabylonlab (approved):

PR #2280, opened 2026-08-19
feat(vault): chart the live position on the embedded liquidation preview

### LINKED ISSUE #2038: [v3] Embedded Liquidation Dashboard preview
**Area:** `services/vault`

**Design:** [Overview with liquidation preview](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10129-34736) — compact `Liquidation Chart` inside an Overview card with an Explore action.

## Context

Add a compact waterfall-chart preview to Overview for users with a position. The preview reuses the full Liquidation Dashboard chart implementation and deep-links to the dedicated page instead of creating a second chart model.

## Scope

- [ ] Embed the compact Liquidation Chart variant in the Overview page
- [ ] Reuse the full Liquidation Dashboard chart data, ordering, and semantics
- [ ] Show the preview only when a position exists
- [ ] Add an Explore action that navigates to `/liquidations`
- [ ] Keep the implementation behind `ENABLE_V3_UI`

## Acceptance criteria

- [ ] A position renders the compact preview with the same underlying values as the full dashboard
- [ ] No-position and disconnected states do not render the preview
- [ ] Explore navigates to `/liquidations`
- [ ] The preview matches Figma in light and dark without duplicating chart-domain logic

## Depends on

- #2043 — full Liquidation Dashboard chart implementation; link the dedicated chart child here if one is created later.

jeremy-babylonlabs:
Closes #2038.

Three commits: wire the embedded preview to the live position cascade, retire `NEXT_PUBLIC_FF_LIQUIDATION_ANALYSIS_CHART` now that nothing is left to gate, then fix three review findings on the charted cascade. Rebased onto `795b4dcf`.

## What changed

The Overview page's embedded Liquidation Analysis card now charts the user's real position. Until now it rendered only from the god-mode cascade override, which nothing populates outside the debug panel — so a depositor with a live position saw an empty card.

`DashboardPage.tsx` destructures `params` from the `usePositionNotifications` call it already makes, and feeds the section the same `{ result, params }` pair the full `/liquidations` dashboard charts. Same calculator, same params, so the compact preview and the full dashboard cannot disagree.

`LiquidationCascade` is now `{ result, params }` — the params the result was computed from, not values derived from them — so `LiquidationAnalysisSection` can re-run the calculator for the price simulator and derive `btcPrice`/`CF`/`vaultsTotal` itself. That removed the mapping in `DashboardPage` entirely, and the preview and `/liquidations` are now structurally identical. `components/pages/Liquidations/` and `src/overrides/` are untouched.

## Precedence ladder

1. **Override** — the god-mode cascade wins when the panel publishes one (`cascadeOverride?.result`, dev-only; the store is never written in production).
2. **Live** — otherwise the live position cascade, when `result` and `params` are both present.
3. **null** — neither has a result: no chart, and the section renders its own empty states rather than charting placeholder numbers.

Guarding on `cascadeOverride?.result` rather than `cascadeOverride` is what makes a status-only override (e.g. a simulated stale price, which carries no cascade) fall through to live instead of blanking the chart — matching `useLiquidationsPageState`.

`usePositionNotifications` returns `result: null, params: null` for every non-`ready` status (`loading`, `no-wallet`, `no-vaults`, `no-price`, `stale-price`), and non-null only together at `ready`. So the disconnected and no-position states chart nothing by construction, not by an extra guard.

## Flag retirement

The flag existed because the overview card's cascade was god-mode-only, and charting that beside a real position would present fabricated liquidation prices as the user's own. The first commit removes that reason, so the second commit removes the flag:

| File | Change |
| --- | --- |
| `config/featureFlags.ts` | delete the `isLiquidationAnalysisChartEnabled` getter and its doc block |
| `config/v3Navigation.ts` | drop `liquidations` from `V3_SECTION_FLAG_GATES` — the sidebar section and the `/liquidations` route are now unconditional |
| `components/simple/DashboardPage.tsx` | drop the early return from the cascade memo |
| `playwright.visual.config.ts`, `.env.example` | remove the entries |
| 4 test files | remove the flag mocks and the flag-off cases that are no longer reachable |

The section-gate mechanism itself stays — Explore still uses it. One consequence worth naming: with `liquidations` ungated, the second nav group always has at least one visible item, so the `filter(group => group.length > 0)` branch in `getVisibleV3NavGroups` is no longer reachable and its test is gone. The line is kept deliberately: it is part of the generic gate mechanism and would be needed again the moment another section in that group gets a flag.

**Deployment note:** this feature has run in no environment. Removing the flag means production is its first exposure, with no kill switch. Requested deliberately; calling it out for the reviewer rather than burying it.

## Review findings fixed

Three defects on the charted cascade, all confirmed against the code before changing anything:

**1. Optimistic activating vaults inflated the cascade.** `useDashboardState.collateralVaults` appends optimistic rows for vaults whose activation tx has landed but which the indexer has not seen, with a sentinel `liquidationIndex: Number.MAX_SAFE_INTEGER`. `usePositionNotifications` fed all of them to `calculate()`, so activating a vault while holding a loan inflated `totalBtc` with collateral the contract cannot seize — pushing every liquidation price DOWN, i.e. reading safer than reality — and labelled a band "Vault 9007199254740992".

Fixed in `usePositionNotifications`, not at the call site: the hook is shared, so `/liquidations`, the cascade banner and the critical top banner all had the same defect. The cascade now sees indexed vaults only, matching how `useVaultsPageData` already filters `isActivating`. A position whose every row is still activating reports `no-vaults` → null result → empty states, which is correct: debt cannot exist against collateral the contract has not registered.

Because that hook also feeds the reorder CTA, I traced the signing path. `ReorderVerificationContext` carries only scalars (CF, THF, maxLB, btcPrice, totalDebtUsd) — no vault list — and both guards re-derive the vault set from chain: `assertReorderMembership` refuses unless the submitted multiset equals `getPosition(user).vaultIds`, and `assertOptimalOrderMatchesOnChain` re-runs `calculate()` against `getBtcVaultBasicInfo` amounts. So this change cannot alter what gets signed; it can only make the guard refuse. The one behavioural consequence: while the indexer lags an already-confirmed activation, the "apply optimal order" CTA now fails closed on a membership mismatch instead of submitting an order derived from inflated collateral. Failing closed is the correct direction for a signing path, and the previous suggestion was computed from collateral the trusted recompute would have rejected anyway.

**2. An empty cascade charted a real position as zeroes.** `calculate()` returns `groups: []` with debt still present when the seizure fraction clamps out of range — reachable with plain governance values (CF 0.92, LB 1.05, THF 1.1 gives a raw fraction of ~1.14, and any THF ≤ LB·CF gives Infinity). The preview charted zero bands and a "Seized 0% / 0.00000000 BTC remaining" summary beside real collateral. Now guarded in the section, which is also where a mid-drag recompute can newly empty the groups; `/liquidations` guards the same condition at `index.tsx:129`. The section already renders nothing for a real position with no cascade, so no new empty state was needed.

**3. The price simulator never re-ran the calculator.** The preview re-projected the live-price result onto the dragged price, while `/liquidations` recomputes (`index.tsx:113-117`). Observable divergence: dragging a $88,400 position with a $73,333 trigger down to $60,000 reported the event as `-17.0%` away (the distance from the live price) while the band beside it already read as liquidated; recomputing reports `+22.2%`, i.e. past the trigger. The section now re-runs `calculate()` at the simulated price behind `useDeferredValue`, copying the `/liquidations` drag mechanics — `calculate()` is an O(3^n) bitmask DP, so the slider and its readout track the undeferred price while only the cascade defers.

## Tests

Extended `components/simple/__tests__/DashboardPage.test.tsx` in place, mocking only `usePositionNotifications` and the cascade override store:

- live result → cascade built from the live params (btcPrice, CF, vault count)
- override set → override wins over a live result
- status-only override + live result → live wins
- no result anywhere → null

The mock exposes a result-derived value alongside the params, so a regression pairing one side's params with the other side's result fails instead of passing.

`usePositionNotifications.test.tsx`: an activating row never reaches the calculator inputs and its sentinel name never escapes; an all-activating position reports `no-vaults`.

`LiquidationAnalysisSection.test.tsx`: its fixture is now real `calculate()` output rather than a hand-built result, so it cannot drift from the maths the component re-runs. Two new cases — the drag recomputes rather than re-projects, and a no-groups cascade charts nothing. Both were confirmed to fail against the pre-fix code before being kept.

Every live fixture is real `calculate()` output, so params and result cannot drift apart.

Verified locally on Node 24 after all three commits: full vault suite 3388 passed / 6 skipped (294 files), `tsc --noEmit -p tsconfig.lib.json` clean, eslint and prettier clean.

### DIFF
--- a/services/vault/src/__tests__/router.test.tsx
+++ b/services/vault/src/__tests__/router.test.tsx
@@ -31,7 +31,6 @@ import { Router } from "../router";
 import { getReserveDetailRoute } from "../routes";
 
 const featureFlagsState = vi.hoisted(() => ({
-  isLiquidationAnalysisChartEnabled: true,
   isExploreEnabled: true,
 }));
 
@@ -220,7 +219,6 @@ function renderAt(path: string): RenderResult {
 
 afterEach(() => {
   vi.restoreAllMocks();
-  featureFlagsState.isLiquidationAnalysisChartEnabled = true;
   featureFlagsState.isExploreEnabled = true;
 });
 
@@ -284,26 +282,6 @@ describe("Router — / and /activity keep their original components", () => {
 describe("Router — section routes", () => {
   beforeEach(() => {
     vi.clearAllMocks();
-  });
-
-  it("redirects /liquidations to / when the liquidation-analysis flag is off", async () => {
-    featureFlagsState.isLiquidationAnalysisChartEnabled = false;
-    renderAt("/liquidations");
-
-    await waitFor(() => {
-      expect(screen.getByTestId(DASHBOARD_TESTID)).toBeInTheDocument();
-    });
-    expect(screen.queryByTestId(LIQUIDATIONS_TESTID)).not.toBeInTheDocument();
-  });
-
-  it("redirects a deep link under /liquidations when its flag is off", async () => {
-    featureFlagsState.isLiquidationAnalysisChartEnabled = false;
-    renderAt("/liquidations/some-deep-link");
-
-    await waitFor(() => {
-      expect(screen.getByTestId(DASHBOARD_TESTID)).toBeInTheDocument();
-    });
-    expect(screen.queryByTestId("not-found")).not.toBeInTheDocument();
   });
 
   it("renders the market detail page at /markets/:reserveId", async () => {
--- a/services/vault/src/applications/aave/hooks/__tests__/usePositionNotifications.test.tsx
+++ b/services/vault/src/applications/aave/hooks/__tests__/usePositionNotifications.test.tsx
@@ -167,4 +167,40 @@ describe("usePositionNotifications — live-HF urgency guardrail", () => {
       result.current.result?.warnings.some((w) => w.type === "urgent"),
     ).toBe(false);
   });
+
+  // Optimistic activating rows are collateral the contract has not seen yet.
+  // Counting them inflates the collateral the cascade thinks is seizable, which
+  // pushes every liquidation price DOWN — the position reads safer than it is.
+  it("excludes optimistic activating vaults from the calculator inputs", () => {
+    const activating: CollateralVaultEntry = {
+      ...makeVault(VAULT_B, 5, Number.MAX_SAFE_INTEGER),
+      isActivating: true,
+    };
+    setDashboardState({
+      collateralVaults: [makeVault(VAULT_A, 0.5, 0), activating],
+      debtValueUsd: 30_000,
+    });
+
+    const { result } = renderHook(() => usePositionNotifications(USER));
+
+    expect(result.current.status).toBe("ready");
+    expect(result.current.params?.vaults.map((v) => v.id)).toEqual([VAULT_A]);
+    // The sentinel index would otherwise surface as "Vault 9007199254740992".
+    expect(result.current.params?.vaults.map((v) => v.name)).toEqual([
+      "Vault 1",
+    ]);
+  });
+
+  it("reports no vaults when every collateral row is still activating", () => {
+    setDashboardState({
+      collateralVaults: [{ ...makeVault(VAULT_A, 0.5, 0), isActivating: true }],
+      debtValueUsd: 30_000,
+    });
+
+    const { result } = renderHook(() => usePositionNotifications(USER));
+
+    expect(result.current.status).toBe("no-vaults");
+    expect(result.current.result).toBeNull();
+    expect(result.current.params).toBeNull();
+  });
 });
--- a/services/vault/src/applications/aave/hooks/usePositionNotifications.ts
+++ b/services/vault/src/applications/aave/hooks/usePositionNotifications.ts
@@ -116,7 +116,15 @@ export function usePositionNotifications(
         reorderVerificationContext: null,
         params: null,
       };
-    if (collateralVaults.length === 0)
+    // Optimistic activating rows carry collateral the contract has not seen
+    // yet, and a sentinel `liquidationIndex`. Including them would inflate
+    // `totalBtc`, pushing every liquidation price DOWN — understating the
+    // risk — and label a band "Vault 9007199254740992". The cascade models
+    // what the protocol would seize, so it sees indexed vaults only.
+    const indexedVaults = collateralVaults.filter(
+      (entry) => !entry.isActivating,
+    );
+    if (indexedVaults.length === 0)
       return {
         result: null,
         status: "no-vaults",
@@ -124,7 +132,7 @@ export function usePositionNotifications(
         params: null,
       };
 
-    const vaults: Vault[] = collateralVaults.map((entry) => ({
+    const vaults: Vault[] = indexedVaults.map((entry) => ({
       id: entry.vaultId,
       btc: entry.amountBtc,
       name: `Vault ${entry.liquidationIndex + 1}`,
--- a/services/vault/src/components/simple/DashboardPage.tsx
+++ b/services/vault/src/components/simple/DashboardPage.tsx
@@ -68,20 +68,15 @@ export function DashboardPage() {
   // live position cascade. A status-only override (stale price) carries no
   // cascade, so it falls through to live. Null when neither has a result: the
   // section shows its empty states rather than charting placeholder numbers.
-  const liquidationCascade = useMemo(() => {
-    const cascade = cascadeOverride?.result
-      ? { result: cascadeOverride.result, params: cascadeOverride.params }
-      : positionNotifications && positionParams
-        ? { result: positionNotifications, params: positionParams }
-        : null;
-    if (!cascade) return null;
-    return {
-      result: cascade.result,
-      btcPrice: cascade.params.btcPrice,
-      collateralFactor: cascade.params.CF,
-      vaultsTotal: cascade.params.vaults.length,
-    };
-  }, [cascadeOverride, positionNotifications, positionParams]);
+  const liquidationCascade = useMemo(
+    () =>
+      cascadeOverride?.result
+        ? { result: cascadeOverride.result, params: cascadeOverride.params }
+        : positionNotifications && positionParams
+          ? { result: positionNotifications, params: positionParams }
+          : null,
+    [cascadeOverride, positionNotifications, positionParams],
+  );
   const {
     collateralBtc,
     collateralValueUsd,
--- a/services/vault/src/components/simple/DashboardPage.tsx
+++ b/services/vault/src/components/simple/DashboardPage.tsx
@@ -62,21 +62,27 @@ export function DashboardPage() {
   // change.
   const cascadeOverride = usePositionCascadeOverride();
   const liquidationCardOverride = useLiquidationCardOverride();
-  // No live source yet: the chart renders only from the god-mode cascade,
-  // behind its own flag. Elsewhere it stays absent rather than charting a real
-  // position from placeholder numbers.
-  const liquidationCascade = useMemo(
-    () =>
-      featureFlags.isLiquidationAnalysisChartEnabled && cascadeOverride?.result
-        ? {
-            result: cascadeOverride.result,
-            btcPrice: cascadeOverride.params.btcPrice,
-            collateralFactor: cascadeOverride.params.CF,
-            vaultsTotal: cascadeOverride.params.vaults.length,
-          }
-        : null,
-    [cascadeOverride],
-  );
+  const { result: positionNotifications, params: positionParams } =
+    usePositionNotifications(isConnected ? address : undefined);
+  // The chart takes the god-mode cascade when the panel publishes one, else the
+  // live position cascade. A status-only override (stale price) carries no
+  // cascade, so it falls through to live. Null when neither has a result: the
+  // section shows its empty states rather than charting placeholder numbers.
+  const liquidationCascade = useMemo(() => {
+    if (!featureFlags.isLiquidationAnalysisChartEnabled) return null;
+    const cascade = cascadeOverride?.result
+      ? { result: cascadeOverride.result, params: cascadeOverride.params }
+      : positionNotifications && positionParams
+        ? { result: positionNotifications, params: positionParams }
+        : null;
+    if (!cascade) return null;
+    return {
+      result: cascade.result,
+      btcPrice: cascade.params.btcPrice,
+      collateralFactor: cascade.params.CF,
+      vaultsTotal: cascade.params.vaults.length,
+    };
+  }, [cascadeOverride, positionNotifications, positionParams]);
   const {
     collateralBtc,
     collateralValueUsd,
@@ -103,9 +109,6 @@ export function DashboardPage() {
     isConnected ? address : undefined,
   );
 
-  const { result: positionNotifications } = usePositionNotifications(
-    isConnected ? address : undefined,
-  );
   const { prices, metadata } = usePrices();
 
   const liquidationNotificationsEnabled =
--- a/services/vault/src/components/simple/DashboardPage.tsx
+++ b/services/vault/src/components/simple/DashboardPage.tsx
@@ -69,7 +69,6 @@ export function DashboardPage() {
   // cascade, so it falls through to live. Null when neither has a result: the
   // section shows its empty states rather than charting placeholder numbers.
   const liquidationCascade = useMemo(() => {
-    if (!featureFlags.isLiquidationAnalysisChartEnabled) return null;
     const cascade = cascadeOverride?.result
       ? { result: cascadeOverride.result, params: cascadeOverride.params }
       : positionNotifications && positionParams
--- a/services/vault/src/components/simple/LiquidationAnalysisSection.tsx
+++ b/services/vault/src/components/simple/LiquidationAnalysisSection.tsx
@@ -1,9 +1,13 @@
 import { Heading, SeizureMap, Slider } from "@babylonlabs-io/core-ui";
-import { useMemo, useState } from "react";
+import { useDeferredValue, useMemo, useState } from "react";
 import { useNavigate } from "react-router";
 import { twJoin } from "tailwind-merge";
 
-import type { CalculatorResult } from "@/applications/aave/positionNotifications/types";
+import { calculate } from "@/applications/aave/positionNotifications";
+import type {
+  CalculatorParams,
+  CalculatorResult,
+} from "@/applications/aave/positionNotifications/types";
 import { LiquidationEventCards } from "@/components/pages/Liquidations/LiquidationEventCards";
 import {
   buildLiquidationChartData,
@@ -17,17 +21,16 @@ import { COPY } from "@/copy";
 import { ROUTES } from "@/routes";
 import { formatPriceUsd } from "@/utils/formatting";
 
-/** The cascade the chart renders. Absent = nothing to chart yet. */
+/** The cascade the chart renders. Absent = nothing to chart yet.
+ *
+ *  Carries the params `result` was computed from, not values derived from
+ *  them: the price simulator has to re-run `calculate()` at the simulated
+ *  price, and `params.vaults` is the only source for the "x/y vaults" count
+ *  (`calculate()` stops emitting groups once the debt clears, so a vault it
+ *  never consumed has no group). */
 export interface LiquidationCascade {
   result: CalculatorResult;
-  btcPrice: number;
-  collateralFactor: number;
-  /**
-   * Vaults in the position. Not derivable from `result`: `calculate()` stops
-   * emitting groups once the debt clears, so a vault it never consumed has no
-   * group and would otherwise vanish from the "x/y vaults" count.
-   */
-  vaultsTotal: number;
+  params: CalculatorParams;
 }
 
 interface LiquidationAnalysisSectionProps {
@@ -88,29 +91,36 @@ function EmptyState({
 }
 
 function LiquidationChartPanel({
-  cascade,
-  effectivePrice,
+  params,
+  projectedResult,
+  projectedPrice,
+  sliderPrice,
   simulating,
   onSimulate,
   onReset,
   onExplore,
 }: {
-  cascade: LiquidationCascade;
-  effectivePrice: number;
+  params: CalculatorParams;
+  projectedResult: CalculatorResult;
+  projectedPrice: number;
+  sliderPrice: number;
   simulating: boolean;
   onSimulate: (price: number) => void;
   onReset: () => void;
   onExplore: () => void;
 }) {
+  // Built from `projectedPrice`, the price `projectedResult` was calculated
+  // at, so the bands and the price line can never disagree while the deferred
+  // cascade is catching up with the thumb.
   const { bands, priceAxis, shareAxisTicks, cards } = useMemo(
     () =>
-      buildLiquidationChartData(cascade.result, {
-        btcPrice: effectivePrice,
-        collateralFactor: cascade.collateralFactor,
-        livePrice: cascade.btcPrice,
-        vaultsTotal: cascade.vaultsTotal,
+      buildLiquidationChartData(projectedResult, {
+        btcPrice: projectedPrice,
+        collateralFactor: params.CF,
+        livePrice: params.btcPrice,
+        vaultsTotal: params.vaults.length,
       }),
-    [cascade, effectivePrice],
+    [projectedResult, projectedPrice, params],
   );
 
   // Cards and bands share keys, so the chart's liquidated state is the
@@ -146,9 +156,9 @@ function LiquidationChartPanel({
       <div className="flex items-center gap-4">
         <div className="flex-1">
           <Slider
-            value={effectivePrice}
+            value={sliderPrice}
             min={0}
-            max={cascade.btcPrice}
+            max={params.btcPrice}
             step={SIM_PRICE_STEP_USD}
             steps={[]}
             onChange={onSimulate}
@@ -157,7 +167,7 @@ function LiquidationChartPanel({
           />
         </div>
         <span className="min-w-[92px] rounded-md border border-secondary-strokeLight px-3 py-1.5 text-center text-sm font-medium tabular-nums text-accent-primary">
-          {formatPriceUsd(effectivePrice)}
+          {formatPriceUsd(sliderPrice)}
         </span>
         <button
           type="button"
@@ -173,8 +183,8 @@ function LiquidationChartPanel({
 
       <SeizureMap
         bands={bands}
-        currentPrice={effectivePrice}
-        currentPriceLabel={formatPriceUsd(effectivePrice)}
+        currentPrice={projectedPrice}
+        currentPriceLabel={formatPriceUsd(projectedPrice)}
         priceLineCaption={COPY.liquidations.bitcoinPriceCaption}
         priceAxis={priceAxis}
         shareAxisTicks={shareAxisTicks}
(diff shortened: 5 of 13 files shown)


services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -59,21 +59,26 @@ export function DashboardPage() {
...
+  // live position cascade. A status-only override (stale price) carries no
+  // cascade, so it falls through to live. Null when neither has a result: the
+  // section shows its empty states rather than charting placeholder numbers.
+  const liquidationCascade = useMemo(() => {
+    const cascade = cascadeOverride?.result
+      ? { result: cascadeOverride.result, params: cascadeOverride.params }
+      : positionNotifications && positionParams
+        ? { result: positionNotifications, params: positionParams }
+        : null;
+    if (!cascade) return null;
```
jrwbabylonlab: needs a `groups.length === 0` guard too. `calculate()` returns `groups: []` with debt present on the weird-params path (`calculate.ts:143`), reachable with plain governance values — CF 0.92, LB 1.05, THF 1.1 gives seizedFraction ~1.14, and any THF <= LB*CF gives Infinity.

we then chart zero bands and print "Seized 0% / Collateral remaining 0.00000000 BTC" next to a real position. `/liquidations` guards exactly this at `index.tsx:129`.

services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -59,21 +59,26 @@ export function DashboardPage() {
...
-            result: cascadeOverride.result,
-            btcPrice: cascadeOverride.params.btcPrice,
-            collateralFactor: cascadeOverride.params.CF,
-            vaultsTotal: cascadeOverride.params.vaults.length,
-          }
-        : null,
-    [cascadeOverride],
-  );
+  const { result: positionNotifications, params: positionParams } =
+    usePositionNotifications(isConnected ? address : undefined);
```
jrwbabylonlab, line 66: this cascade is built from `collateralVaults`, which includes the optimistic activating rows (`useDashboardState.ts:123-129`, `liquidationIndex: Number.MAX_SAFE_INTEGER`).

so activating a vault while holding a loan inflates `totalBtc` with collateral the contract has not seen, which pushes every displayed liquidation price DOWN, i.e. safer than reality. it also labels a band "Vault 9007199254740992". the `hasCollateral` gate deliberately excludes these rows, the charted cascade does not.

services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -59,21 +59,26 @@ export function DashboardPage() {
...
+  // section shows its empty states rather than charting placeholder numbers.
+  const liquidationCascade = useMemo(() => {
+    const cascade = cascadeOverride?.result
+      ? { result: cascadeOverride.result, params: cascadeOverride.params }
+      : positionNotifications && positionParams
+        ? { result: positionNotifications, params: positionParams }
+        : null;
+    if (!cascade) return null;
+    return {
+      result: cascade.result,
```
jrwbabylonlab: the preview passes this live-price result into the chart and never re-runs `calculate()` at the simulated price, while `/liquidations` recomputes (`index.tsx:113-117`).

so at any simulated price the two views disagree: `group.distancePct` stays anchored to the live price here but tracks the simulated one there, and drag low enough that `totalBtc * price < $1000` and the full page collapses to the dust group while this one keeps the multi-group cascade.

services/vault/src/config/featureFlags.ts
```diff
@@ -158,24 +158,6 @@ export default {
...
-   * Liquidations section and the /liquidations Liquidation Dashboard. The
-   * overview card's empty states read the live position and are not gated by
-   * this.
-   * Why needed: the /liquidations page charts the live
-   * `usePositionNotifications` cascade, but the overview card's cascade is
-   * still god-mode-only, and rendering that beside a real position would
-   * present fabricated liquidation prices as the user's own.
-   * Default: false (no chart unless explicitly set to "true")
-   */
-  get isLiquidationAnalysisChartEnabled() {
```
jrwbabylonlab, line 175: this deletes the only kill switch in the same PR that first exposes the feature to production, and the charted path runs the O(3^n) `calculate()` over live indexer vault sets on the highest-traffic page.

also `.github/workflows/service-release-vault.yml:201` still exports `NEXT_PUBLIC_FF_LIQUIDATION_ANALYSIS_CHART`, so the variable survives as a toggle that silently does nothing. worth keeping the flag for one release, or splitting into two deploys.

jonybur (dismissed):
The switch to a { result, params } cascade and re-running calculate() at the simulated price behind useDeferredValue puts the preview on the same numeric footing as the /liquidations page. The three defects called out in the description all hold up: filtering optimistic-activating rows in usePositionNotifications prevents the sentinel index from surfacing as 'Vault 9007199254740992' and removes the collateral inflation that pushed liquidation prices DOWN; the projectedResult.groups.length > 0 guard in LiquidationAnalysisSection covers the invalid-governance case where the seizure fraction clamps out of range; and re-running the calculator on the dragged price replaces the reprojection that reported a liquidated band as -17.0 percent away. The override then live then null precedence in DashboardPage reads correctly, and gating on cascadeOverride result rather than cascadeOverride is what lets a status-only override (stale price) fall through to live instead of blanking the chart. Nothing else consumes LiquidationCascade, so the interface change is safe. Flag retirement is production-first with no kill switch, but that was called out explicitly and is a business decision.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -146,9 +156,9 @@ function LiquidationChartPanel({
       <div className="flex items-center gap-4">
         <div className="flex-1">
           <Slider
-            value={effectivePrice}
+            value={sliderPrice}
             min={0}
-            max={cascade.btcPrice}
+            max={params.btcPrice}
```
gbarkhatov, line 161: **1 — The price simulator feeds `calculate()` prices the cascade cannot model (medium).**

`min={0}` was harmless while the dragged price only re-projected an existing result. Commit `e5fc5776` now hands it to `calculate()` at line 241, and that opens two degenerate regimes at the two ends of the drag.

**Full-left.** `calculate.ts:97-98` takes the dust branch when `collateralValue < DUST_THRESHOLD_USD` ($1,000) — below ~$1,667 for the test fixture's 0.6 BTC — collapsing a multi-band cascade into a single all-vaults band with a dust warning. At $0 exactly, `distancePct = ((liqPrice - btcPrice) / btcPrice) * 100` is `Infinity`, and `formatSignedPct` (`liquidationChartData.ts:203`) renders it verbatim: every event card and band popover reads `+Infinity%`.

**Full-right.** A native range input only lands on `min + n*step`, so with `min={0}` and `step={1}` the reachable values are the integers — and `params.btcPrice` is a Chainlink 8-decimal quotient, essentially never one. At a live $104,321.37 a full drag right yields `104321`, so `sliderPrice < livePrice` stays true and the simulator is stuck in "Simulation" with Reset the only escape. `formatPriceUsd` prints `$104,321` for both, so the cause is invisible. Note this is precisely the failure the `SIM_PRICE_STEP_USD` doc comment at lines 48-52 says `step = 1` prevents — `min={0}` defeats it, because step 1 only guarantees an integer grid, not a grid containing the live price.

`SimulationToolbar.tsx:79-100` already solves both, and its comment calls out the second by name — it realigns the slider's own min down to the nearest step below `axisFloorPrice`, which both keeps the drag inside the modellable range and puts `livePrice` on the grid. `axisFloorPrice` is exported from `liquidationChartData.ts:163`; reusing that computation here closes both ends.

Worth noting while you're in here: since `pLiq = remainingDebt / (remainingBtc * CF)` is price-independent, the recompute changes only `distancePct`, `currentHF` and which branch `calculate()` takes. The bands are identical either way — so the entire benefit of the change is the distance labels, and the entire risk is the branch selection above.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -251,11 +283,13 @@ export function LiquidationAnalysisSection({
         onAction={onBorrow}
       />
     );
-  } else if (cascade) {
+  } else if (cascade && projectedResult && chartable) {
```
gbarkhatov, line 286: **2 — The card vanishes on a stale price, where `/liquidations` shows an empty state (medium).**

`usePositionNotifications` returns `params: null` for both `stale-price` and `no-price` (lines 105-118), so `cascade` is null, this branch is skipped, and control reaches the `return null` at line 302 — the entire Liquidation Analysis card, heading included, disappears from the overview for a depositor who has collateral *and* debt.

`/liquidations` treats exactly this state as worth naming. `index.tsx:236-258` renders `COPY.liquidations.empty.unavailableTitle` / `unavailableDescription` for it, under a comment saying the page that exists *for* the analysis must not go silently half-blank. The copy already exists.

What changed is reachability. While the flag was off, this path was god-mode-only; with the flag retired it is the ordinary thing a stale oracle now does to the overview. It also contradicts the description's "the section renders its own empty states rather than charting placeholder numbers" — it does for `!hasCollateral` and `!hasLoans`, but not for this one.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -212,23 +222,45 @@ export function LiquidationAnalysisSection({
   const navigate = useNavigate();
   const [simulatedPrice, setSimulatedPrice] = useState<number | null>(null);
 
-  const showChart = hasCollateral && hasLoans && Boolean(cascade);
-  const livePrice = cascade?.btcPrice ?? 0;
+  const livePrice = cascade?.params.btcPrice ?? 0;
```
gbarkhatov, line 225: **3 — `?? 0` makes the first ready render calculate a cascade at $0 (medium-low).**

On the `null` → ready transition, `useDeferredValue` returns the *previous* value for the render in which `cascade` first becomes non-null — and the previous `sliderPrice` was `0`, from this fallback. So `projectedPrice (0) !== cascade.params.btcPrice`, the fast path at line 240 is missed, and line 241 runs `calculate({ ...params, btcPrice: 0 })`.

That is not a harmless wasted call. At `btcPrice: 0`, `collateralValue` is `0`, which is `< DUST_THRESHOLD_USD`, so `calculate()` returns the dust branch: one group, `distancePct: Infinity`, `currentHF: 0`. `groups.length > 0`, so `chartable` is true and the panel renders — a `+Infinity%` band under a "Simulation" chip (`0 < livePrice`) — before the deferred re-render corrects it.

`/liquidations` avoids this by construction rather than by luck: `livePrice = params?.btcPrice ?? null` (`index.tsx:86`), `currentPrice` stays nullable, and `cascadeResult` returns null while `projectedPrice === null` (`index.tsx:113-114`), so nothing is charted until the price is real.

Independent of whether the bad frame survives to paint, a `?? 0` on a price that feeds a financial calculation is the fallback CLAUDE.md's "No Silent Fallbacks on Critical Paths" rule is about — absence of a price is a real state here, and 0 is a value `calculate()` accepts and answers wrongly. Keeping it nullable, as the reference page does, removes the whole class.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -212,23 +222,45 @@ export function LiquidationAnalysisSection({
...
+
+  // Invalid governance params leave `groups` empty with debt still present
+  // (calculate.ts skips the prefix walk when `seizedFraction` clamps). Charting
+  // that renders zero bands and a "Seized 0% / 0.00000000 BTC remaining"
+  // summary beside a real position, so treat it as nothing to chart — the same
+  // guard the /liquidations page applies.
+  const chartable =
+    projectedResult !== null && projectedResult.groups.length > 0;
+  const showChart = hasCollateral && hasLoans && chartable;
+  const simulating = showChart && projectedPrice < livePrice;
```
gbarkhatov, line 252: **4 — `simulating` derives from the deferred price, which is backwards (low-medium).**

`/liquidations` deliberately uses the immediate price for this, with the reason written down at `index.tsx:262-264`: *"The immediate price, not the deferred one: the chip and Reset must respond on the first drag step rather than trailing the cascade."* This is the deferred one.

Two consequences:

- **On drag:** the chip and Reset lag a whole `calculate()` behind the thumb — the one thing deferring `projectedPrice` was supposed to avoid. The rest of the mechanic is right; only this derivation opts back into the latency.
- **On live price ticks:** `usePrices` repolls every 60s (15s unhealthy) and BTC moves on essentially every poll. `params.btcPrice` rises, `sliderPrice` follows immediately, `projectedPrice` trails one render — so `projectedPrice < livePrice` is briefly true and a "Simulation" chip plus an enabled Reset paint on an upward tick the user did nothing to cause. The same trailing render also misses the fast path at line 240 and runs a full O(3ⁿ) DP at the stale price.

`sliderPrice < livePrice` fixes both.

services/vault/src/applications/aave/hooks/usePositionNotifications.ts
```diff
@@ -116,15 +116,23 @@ export function usePositionNotifications(
         reorderVerificationContext: null,
         params: null,
       };
-    if (collateralVaults.length === 0)
+    // Optimistic activating rows carry collateral the contract has not seen
+    // yet, and a sentinel `liquidationIndex`. Including them would inflate
+    // `totalBtc`, pushing every liquidation price DOWN — understating the
+    // risk — and label a band "Vault 9007199254740992". The cascade models
+    // what the protocol would seize, so it sees indexed vaults only.
+    const indexedVaults = collateralVaults.filter(
```
gbarkhatov, line 124: **5 — The change is right; the rationale above it is inverted (low).**

The filter belongs here. But the comment — repeated in the description and in the new test — says these rows are "collateral the contract has not seen yet" that "the contract cannot seize", and that isn't what the code does. `addActivatingVault` fires from `onShowSuccessModal` *after* the activation transaction succeeds (`useActivationState.ts:99-122`), under a comment that says so: "while the Aave indexer catches up (~15s gap)". The collateral is on-chain and seizable; it's the indexer that lags.

The real justification is consistency, and it's stronger. `useDashboardState` already keeps every financial figure indexer-pure and hands the optimistic row to display only — `collateralBtc`, `collateralValueUsd`, `debtValueUsd` and `healthFactor` all come straight from `useAaveUserPosition`, and only `displayCollateralBtc` and `collateralVaults` carry the activating rows (lines 155-169, with "an optimistic activating row must NOT unlock financial actions" spelled out). Before this PR the cascade's `totalBtc` included the activating vault while the debt and health factor rendered beside it did not. That mismatch is what the filter removes.

Worth restating the direction in the description, because it's the opposite of what's written there now: the cascade doesn't stop overstating safety, it starts reading **conservatively** during the indexer gap. Two consequences follow that are worth a line:

- For up to ~15s after activating, the banners no longer reflect newly added collateral — they used to, optimistically — and the cascade's `currentHF` can sit well below the live on-chain one. Over-warning is the safe direction, and the live-HF guardrail below only ever *adds* urgency, so nothing corrects it back down. Still worth naming as a deliberate trade rather than leaving it to be rediscovered as a bug.
- The reorder CTA fails closed with `ReorderMembershipMismatchError`'s technical message during that window. The description covers the signing analysis well; it's the message the depositor sees that isn't named.

The sentinel-name half of the fix is unambiguously correct.

Nit, three lines down: `` name: `Vault ${entry.liquidationIndex + 1}` `` builds user-visible copy inline, which CLAUDE.md routes through `copy.ts`. A `COPY.liquidations.vaultName(n)` helper would also give the sentinel failure mode one place to be guarded instead of none.

services/vault/src/__tests__/router.test.tsx
```diff
@@ -283,26 +281,6 @@ describe("Router — section routes", () => {
     vi.clearAllMocks();
   });
 
-  it("redirects /liquidations to / when the liquidation-analysis flag is off", async () => {
```
gbarkhatov, line 289: **7 — Deleting these was right, but the branch they covered is still in `router.tsx` (low).**

With `liquidations` out of `V3_SECTION_FLAG_GATES`, `isV3SectionEnabled("liquidations")` returns the `?? true` default unconditionally, so the ternary at `router.tsx:207-217` can only ever take its first arm and the `<Navigate to={ROUTES.OVERVIEW} replace />` below it is unreachable. Removing the tests was the correct half; leaving the branch inverts the zero-dead-code policy, since the siblings `/vaults`, `/loans` and `/activity` are all plain `element={<Suspense>…}` with no gate.

Different case from `getVisibleV3NavGroups`' `.filter()`, which the description argues for keeping: that one is generic mechanism with no named section in it, whereas this is a specific call whose result is now a constant.

While you're there, `router.tsx:222` says Explore is "Gated on its own flag (like `/liquidations`)" — that comparison now points at an ungated route. Same for this file's header at lines 11-13, which still describes the deleted cases.

services/vault/src/applications/aave/hooks/__tests__/usePositionNotifications.test.tsx
```diff
@@ -167,4 +167,40 @@ describe("usePositionNotifications — live-HF urgency guardrail", () => {
...
+      collateralVaults: [makeVault(VAULT_A, 0.5, 0), activating],
+      debtValueUsd: 30_000,
+    });
+
+    const { result } = renderHook(() => usePositionNotifications(USER));
+
+    expect(result.current.status).toBe("ready");
+    expect(result.current.params?.vaults.map((v) => v.id)).toEqual([VAULT_A]);
+    // The sentinel index would otherwise surface as "Vault 9007199254740992".
+    expect(result.current.params?.vaults.map((v) => v.name)).toEqual([
```
gbarkhatov, line 189: **8 — The test asserts that the filter ran, not what it changes (low).**

Both assertions read `params.vaults`, which pins the input list and the sentinel-name fix. Neither touches `result` — and `result` is where this change is visible to a depositor.

With this exact fixture it's a sharp contrast: BTC $60,000, CF 0.7, debt $30,000, and after the filter a single 0.5 BTC vault, so `collateralValue` is $30,000 and `currentHF` comes out at **0.7** — a position the calculator reads as already liquidatable, with the urgent/cliff warnings that follow — while `healthFactor` from the same `setDashboardState` default is **2.0**. That divergence is exactly the conservative-during-the-indexer-gap trade-off from **5**, and the test passes without ever looking at it.

An assertion on `result.currentHF` (or on the warning types) would pin the trade-off deliberately, and would fail loudly if someone later "fixed" the divergence by feeding the optimistic rows back in.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -212,23 +222,45 @@ export function LiquidationAnalysisSection({
...
 
   const summary = useMemo(
```
gbarkhatov, line 254: **9 — `summary` is computed twice per render (nit).**

`buildLiquidationChartData` already returns a `summary` (`liquidationChartData.ts:383`), built by calling this same `buildSimulationSummary` with the same three arguments — `projectedResult`, `projectedPrice`, `params.vaults.length`. `LiquidationChartPanel` destructures `{ bands, priceAxis, shareAxisTicks, cards }` and drops it on the floor, and the parent recomputes it here.

`/liquidations` reads it off the single call (`chart.summary`). The comment at `liquidationChartData.ts:375-382` is explicit that the summary is bundled into that return *specifically* so it can't be paired with a different price than the bands, and that two definitions of "seized" would print two different percentages for one position — this reintroduces the second site the comment is arguing against.

Structural, since the header lives outside the panel: lifting `buildLiquidationChartData` into the parent and passing the pieces down, or having the panel report its summary up, would both work. Not worth blocking on, but the duplication is exactly what that comment warns about.

gbarkhatov (dismissed):
Reviewed all three commits (`217554e1` → `e5fc5776`) against the source: the live-cascade wiring, the flag retirement, and the follow-up fixes. Read `calculate.ts`, `liquidationChartData.ts`, `SimulationToolbar.tsx`, the `/liquidations` page, `useDashboardState.ts`, `useActivationState.ts`, `router.tsx` and the core-ui `Slider`. The precedence ladder is right — `cascadeOverride?.result` does fall through for a status-only override, and `params` is non-nullable on that type — and the reorder-path analysis in the description holds: both guards re-derive the vault set from chain, so the filter can only make the guard refuse.

Approving. Nothing here blocks the wiring, but **1** is worth fixing before this is production's first exposure to the feature, given the description's own note that there is no kill switch left.

Nine findings, five of them the same theme: this PR adopts `/liquidations`' recompute-at-the-simulated-price mechanic, and the reference page carries four guards around that mechanic that did not come with it.

| # | Where | |
| --- | --- | --- |
| 1 | `LiquidationAnalysisSection.tsx:161` | slider feeds `calculate()` prices it can't model |
| 2 | `LiquidationAnalysisSection.tsx:286` | card vanishes on a stale price |
| 3 | `LiquidationAnalysisSection.tsx:225` | `?? 0` calculates a cascade at $0 |
| 4 | `LiquidationAnalysisSection.tsx:252` | `simulating` reads the deferred price |
| 5 | `usePositionNotifications.ts:124` | right change, inverted rationale |
| 6 | `service-release-vault.yml:201` | retired flag still injected (below) |
| 7 | `router.tsx:207-217` | dead `<Navigate>` branch |
| 8 | `usePositionNotifications.test.tsx:189` | test asserts the filter ran, not what it changes |
| 9 | `LiquidationAnalysisSection.tsx:254` | `summary` computed twice |

---

### 6 — The retired flag is still live in CI config

`.github/workflows/service-release-vault.yml:201` still injects it into the vault release build:

```yaml
NEXT_PUBLIC_FF_LIQUIDATION_ANALYSIS_CHART: ${{ vars.NEXT_PUBLIC_FF_LIQUIDATION_ANALYSIS_CHART }}
```

A repo-wide grep for `LIQUIDATION_ANALYSIS_CHART` returns this line and nothing else, so it is the single site the retirement table missed. Harmless at runtime — nothing reads it now — but it keeps a dead repo variable looking load-bearing. The workflow isn't in the diff, so this couldn't be an inline comment.

Also outside the diff: `markdown/explore/explore.md:59-60` still describes `liquidations` as "still behind `isLiquidationAnalysisChartEnabled`". Historical planning spec, so this is a judgement call rather than a finding.

### On `getVisibleV3NavGroups`' `.filter(group => group.length > 0)`

Not raising it — the description already names it as unreachable and argues for keeping it as generic mechanism, which I agree with. Worth considering whether the deleted test should come back as a mixed-group case (`explore` off, `liquidations` on → group renders `liquidations` alone), so the surviving branch keeps its coverage.

jrwbabylonlab (dismissed):

jonybur (dismissed):
The wiring is right: DashboardPage destructures params from the same usePositionNotifications call the notification banner reads, and LiquidationCascade now carries the full CalculatorParams pair rather than a set of derived scalars, so LiquidationAnalysisSection can re-run calculate() at the simulated price and the embedded preview cannot disagree with /liquidations by construction. The three review findings the description names are addressed: chartable guards against calculate() returning empty groups with debt still present, indexedVaults filters optimistic activating rows so the cascade never inflates totalBtc from collateral the contract has not indexed yet, and the what-if path is now a real recompute rather than a projection. useDeferredValue on projectedPrice keeps the slider responsive without letting the read-out or the simulating chip lag the thumb. The composition tests assert on the cascade IDENTITY that reaches the section, not just the values re-derived from it, so a regression pairing an override's params with the live result cannot slip past. Flag retirement leaves nothing behind - the getter, the sidebar branch, the router branch and the section-flag gate all go, and tests are converted rather than deleted where the covered code still exists.

jonybur (approved):
Cascade wiring, override-then-live-then-null precedence, and the recompute-at-simulated-price mechanic are the right shape. The activating-vault filter in usePositionNotifications and the chartable groups.length guard both address concrete divergences from /liquidations. Substantive open items are already on the PR under gbarkhatov's review.

PR #2281, opened 2026-08-19
Feat/ledger pop wallet policy

jrwbabylonlab:
Ledger vault app support for the depositor flow: BIP-322 proof of possession, and Pre-PegIn signing under the device's default BIP-86 wallet policy (tr(@0/**)) — without a policy the device answers SW_OK with no signature.

### DIFF
--- a/packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
@@ -123,11 +123,13 @@ function encodeScriptNum(n: number): Buffer {
   return Buffer.from([bytes.length, ...bytes]);
 }
 
-/** N-of-N multisig fragment (test_sign_psbt_validate.py:142). */
+/**
+ * N-of-N multisig fragment. No single-key special case: `encode_multisig_group`
+ * (fw `src/vault_script.c:204-231` @ e2d0c45b) emits the counted form for N>=1,
+ * so a `<key> OP_CHECKSIG` shortcut would build a leaf 4 bytes short of the one
+ * the device rebuilds and compares.
+ */
 function multisigGroup(keys: readonly Buffer[], isFinal: boolean): Buffer {
-  if (keys.length === 1) {
-    return Buffer.concat([Buffer.from([PUSH_32]), keys[0], Buffer.from([isFinal ? OP_CHECKSIG : OP_CHECKSIGVERIFY])]);
-  }
   const parts: Buffer[] = [Buffer.concat([Buffer.from([PUSH_32]), keys[0], Buffer.from([OP_CHECKSIG])])];
   for (const key of keys.slice(1)) {
     parts.push(Buffer.concat([Buffer.from([PUSH_32]), key, Buffer.from([OP_CHECKSIGADD])]));
@@ -343,6 +345,13 @@ export function termsToIntent(terms: DepositTerms): IntentFixture {
 const PREPEGIN_INPUT_COUNT = 2;
 /** Distinct but otherwise arbitrary prevout hashes — the device never fetches them. */
 const PREPEGIN_PREVOUT_HASH_BASE = 0x40;
+/**
+ * `_validate_prepegin` (fw `sign_psbt_validate.c:334-545` @ e2d0c45b) does not
+ * constrain the sequence — it only hashes it into the txid it compares against
+ * the intent (`:283-292`). So this must match what the toolkit's funding path
+ * emits: bitcoinjs' DEFAULT_SEQUENCE (fundPeginTransaction.ts:148).
+ */
+const PREPEGIN_SEQUENCE_FINAL = 0xffffffff;
 
 export interface PrePeginPsbtFixture {
   /** v0, NOT augmented — the test augments it the way the provider does. */
@@ -378,10 +387,7 @@ export function buildPrePeginPsbt(hashlock: Buffer, changeXOnlyHex: string): Pre
     psbt.addInput({
       hash: Buffer.alloc(32, PREPEGIN_PREVOUT_HASH_BASE + i),
       index: i,
-      // SEQUENCE_FINAL: the fw (fix/audit-findings > e2d0c45b) requires Pre-PegIn
-      // inputs present AND final (no RBF/CSV) — and the toolkit's funding path
-      // emits bitcoinjs' default 0xffffffff (fundPeginTransaction.ts:148).
-      sequence: 0xffffffff,
+      sequence: PREPEGIN_SEQUENCE_FINAL,
       witnessUtxo: { script: depositorSpk, value: PREPEGIN_INPUT_VALUE_SATS },
       tapInternalKey: depositor,
     });
@@ -405,10 +411,13 @@ export interface PeginPsbtFixture {
   readonly controlBlock: Buffer;
 }
 
+/** RBF-signalling, per the fw builder. */
+const PEGIN_SEQUENCE = 0xfffffffe;
+
 /**
  * PegIn PSBTv0: input 0 spends the HTLC at (prepegin_txid, vout 0) via
  * Leaf 0; outputs are Vault, Depositor Claim, and the P2A anchor. Tx v3
- * (TRUC), locktime 0, sequence 0xFFFFFFFE — all per the fw builder.
+ * (TRUC), locktime 0, sequence {@link PEGIN_SEQUENCE} — all per the fw builder.
  */
 export function buildPeginPsbt(hashlock: Buffer, prepeginTxidInternal: Buffer): PeginPsbtFixture {
   const depositor = hexToBuffer(DEPOSITOR_XONLY_HEX);
@@ -435,7 +444,7 @@ export function buildPeginPsbt(hashlock: Buffer, prepeginTxidInternal: Buffer): 
   psbt.addInput({
     hash: prepeginTxidInternal, // Buffer form: used as-is (internal order)
     index: HTLC_VOUT,
-    sequence: 0xfffffffe,
+    sequence: PEGIN_SEQUENCE,
     witnessUtxo: { script: htlcScriptPubKey, value: HTLC_VALUE_SATS },
     tapInternalKey: NUMS_XONLY,
     tapMerkleRoot: merkleRoot,
--- a/packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
@@ -378,7 +378,10 @@ export function buildPrePeginPsbt(hashlock: Buffer, changeXOnlyHex: string): Pre
     psbt.addInput({
       hash: Buffer.alloc(32, PREPEGIN_PREVOUT_HASH_BASE + i),
       index: i,
-      sequence: 0xfffffffd,
+      // SEQUENCE_FINAL: the fw (fix/audit-findings > e2d0c45b) requires Pre-PegIn
+      // inputs present AND final (no RBF/CSV) — and the toolkit's funding path
+      // emits bitcoinjs' default 0xffffffff (fundPeginTransaction.ts:148).
+      sequence: 0xffffffff,
       witnessUtxo: { script: depositorSpk, value: PREPEGIN_INPUT_VALUE_SATS },
       tapInternalKey: depositor,
     });
--- a/packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
@@ -1,5 +1,10 @@
 /**
- * Signable PegIn fixture for the Speculos e2e suite.
+ * Signable Pre-PegIn and PegIn fixtures for the Speculos e2e suite.
+ *
+ * The Pre-PegIn is built FIRST and its txid feeds both the intent
+ * (`prepeginTxid`) and the PegIn input, because `_validate_prepegin` compares
+ * SHA256d(unsigned tx) against the intent's `prepegin_txid`
+ * (fw `sign_psbt_validate.c:529-537` @ 4decf822).
  *
  * The committed SIGN_PSBT vectors were generated from foreign seeds, so this
  * builder replicates the firmware's own signable test path byte-for-byte:
@@ -12,13 +17,16 @@
  */
 
 import * as ecc from "@bitcoin-js/tiny-secp256k1-asmjs";
-import { crypto as bcrypto, Psbt, Transaction } from "bitcoinjs-lib";
+import { crypto as bcrypto, initEccLib, payments, Psbt, Transaction } from "bitcoinjs-lib";
 import { Buffer } from "buffer";
 import { createHash, createHmac } from "node:crypto";
 
 import type { IntentScalars, IntentVaultGroup } from "../../intentTlv";
 import { tapLeafHash } from "../../tapLeafHash";
 import type { DepositTerms } from "../../types";
+
+// `payments.p2tr` tweaks the internal key, which needs the ecc backend.
+initEccLib(ecc);
 
 /** BIP-341 tapscript leaf version — the only one the vault graph uses. */
 const TAPSCRIPT_LEAF_VERSION = 0xc0;
@@ -59,8 +67,14 @@ export const PREPEGIN_MAX_FEE_SATS = 500_000;
  */
 export const HTLC_VALUE_SATS = VAULT_AMOUNT_SATS + DEPOSITOR_CLAIM_VALUE_SATS + PEGIN_ANCHOR_VALUE_SATS + 234_567;
 
-/** _PREPEGIN_TXID = bytes(range(32)) (test_sign_psbt_validate.py:135), INTERNAL order. */
-export const PREPEGIN_TXID_INTERNAL = Buffer.from(Array.from({ length: 32 }, (_, i) => i));
+/** Value of each Pre-PegIn funding input — two of them cover HTLC + anchor + change + fee. */
+export const PREPEGIN_INPUT_VALUE_SATS = 20_000_000;
+/** ≤ PREPEGIN_MAX_FEE_SATS, and far above 1 sat/vB for a ~254 vB transaction. */
+const PREPEGIN_FEE_SATS = 50_000;
+/** VAULT_DUST_LIMIT — the exact CPFP anchor value the device demands (fw vault_constants.h:64). */
+const CPFP_ANCHOR_VALUE_SATS = 546;
+/** Testnet BIP-32 version bytes — the signet build answers tpub/tprv. */
+export const TESTNET_VERSIONS = { public: 0x043587cf, private: 0x04358394 };
 
 /** _DERIVE_CONTEXT = bytes(range(72)) (test_sign_psbt_validate.py:591). */
 export const DERIVE_CONTEXT = Uint8Array.from(Array.from({ length: 72 }, (_, i) => i));
@@ -249,21 +263,22 @@ export function vaultHashlock(root: Uint8Array, htlcVout: number): Buffer {
 // ---------------------------------------------------------------------------
 
 /**
- * The e2e deposit terms, matching the fw test intent (`_setup_s2_state` with
- * `_PREPEGIN_TXID`). `prepeginTxid` is DISPLAY order — the seam converts
- * display → internal exactly like wallet-connector's provider, and the wire
- * carries the internal-order bytes the firmware compares against the PSBT.
- * `vaultCoreVersion` is 2: the device-supported tx-graph version gate — the
- * TLV's structure/version constants (both 1) are pinned inside the encoder.
+ * The e2e deposit terms, bound to the Pre-PegIn the suite actually signs —
+ * pass {@link PrePeginPsbtFixture.txidInternal}. `prepeginTxid` is DISPLAY
+ * order: the seam converts display → internal exactly like wallet-connector's
+ * provider, and the wire carries the internal-order bytes the firmware compares
+ * against the PSBT. `vaultCoreVersion` is 2: the device-supported tx-graph
+ * version gate — the TLV's structure/version constants (both 1) are pinned
+ * inside the encoder.
  */
-export function buildDepositTerms(): DepositTerms {
+export function buildDepositTerms(prepeginTxidInternal: Buffer): DepositTerms {
   return {
     vaultCoreVersion: 2,
     protocolFeeRate: BigInt(BASE_FEE_RATE),
     timelockPegin: PEGIN_CSV_TIMELOCK,
     timelockAssert: PAYOUT_TIMELOCK,
     timelockRefund: HTLC_REFUND_TIMELOCK,
-    prepeginTxid: Buffer.from(PREPEGIN_TXID_INTERNAL).reverse().toString("hex"),
+    prepeginTxid: Buffer.from(prepeginTxidInternal).reverse().toString("hex"),
     prepeginMaxFee: BigInt(PREPEGIN_MAX_FEE_SATS),
     vaultKeeperBtcPubkeys: [KEEPER_KEY_HEX],
     universalChallengerBtcPubkeys: [CHALLENGER_KEY_HEX],
@@ -321,6 +336,61 @@ export function termsToIntent(terms: DepositTerms): IntentFixture {
 }
 
 // ---------------------------------------------------------------------------
+// Pre-PegIn PSBT — the transaction the intent binds to
+// (`_validate_prepegin`, fw sign_psbt_validate.c:334-545 @ 4decf822)
+// ---------------------------------------------------------------------------
+
+const PREPEGIN_INPUT_COUNT = 2;
+/** Distinct but otherwise arbitrary prevout hashes — the device never fetches them. */
+const PREPEGIN_PREVOUT_HASH_BASE = 0x40;
+
+export interface PrePeginPsbtFixture {
+  /** v0, NOT augmented — the test augments it the way the provider does. */
+  readonly psbtHex: string;
+  /** Raw SHA256d of the unsigned tx (PSBT/internal order) — the intent's prepegin_txid and the PegIn prevout. */
+  readonly txidInternal: Buffer;
+  readonly htlcScriptPubKey: Buffer;
+}
+
+/**
+ * Toolkit-shaped Pre-PegIn: two depositor key-path inputs → [HTLC, CPFP anchor
+ * 546 P2TR(depositor), change P2TR(changeKey)]. No OP_RETURN (optional on-device).
+ * Layout matches btc-vault `[HTLC×n, OP_RETURN?, anchor, change*]`.
+ */
+export function buildPrePeginPsbt(hashlock: Buffer, changeXOnlyHex: string): PrePeginPsbtFixture {
+  const depositor = hexToBuffer(DEPOSITOR_XONLY_HEX);
+  const vp = hexToBuffer(VP_KEY_HEX);
+  const keepers = [hexToBuffer(KEEPER_KEY_HEX)];
+  const challengers = [hexToBuffer(CHALLENGER_KEY_HEX)];
+  const leaf0 = htlcLeaf0(depositor, vp, keepers, challengers, hashlock);
+  const leaf1 = htlcLeaf1(depositor, HTLC_REFUND_TIMELOCK);
+  const merkleRoot = tapBranch(tapLeafHash(TAPSCRIPT_LEAF_VERSION, leaf0), tapLeafHash(TAPSCRIPT_LEAF_VERSION, leaf1));
+  const htlcScriptPubKey = Buffer.concat([Buffer.from([0x51, PUSH_32]), tweakNums(merkleRoot).xOnly]);
+  const depositorSpk = payments.p2tr({ internalPubkey: depositor }).output!;
+  const changeSpk = payments.p2tr({ internalPubkey: hexToBuffer(changeXOnlyHex) }).output!;
+  const changeValue =
+    PREPEGIN_INPUT_COUNT * PREPEGIN_INPUT_VALUE_SATS - HTLC_VALUE_SATS - CPFP_ANCHOR_VALUE_SATS - PREPEGIN_FEE_SATS;
+
+  const psbt = new Psbt();
+  psbt.setVersion(2); // fw requires tx_version >= 2
+  psbt.setLocktime(0);
+  for (let i = 0; i < PREPEGIN_INPUT_COUNT; i++) {
+    psbt.addInput({
+      hash: Buffer.alloc(32, PREPEGIN_PREVOUT_HASH_BASE + i),
+      index: i,
+      sequence: 0xfffffffd,
+      witnessUtxo: { script: depositorSpk, value: PREPEGIN_INPUT_VALUE_SATS },
+      tapInternalKey: depositor,
+    });
+  }
+  psbt.addOutput({ script: htlcScriptPubKey, value: HTLC_VALUE_SATS });
+  psbt.addOutput({ script: depositorSpk, value: CPFP_ANCHOR_VALUE_SATS });
+  psbt.addOutput({ script: changeSpk, value: changeValue });
+  const tx = Transaction.fromBuffer(psbt.data.globalMap.unsignedTx.toBuffer());
+  return { psbtHex: psbt.toHex(), txidInternal: tx.getHash(), htlcScriptPubKey };
+}
+
+// ---------------------------------------------------------------------------
 // PegIn PSBT (test_sign_psbt_validate.py:309 `_build_pegin_psbt`)
 // ---------------------------------------------------------------------------
 
@@ -337,7 +407,7 @@ export interface PeginPsbtFixture {
  * Leaf 0; outputs are Vault, Depositor Claim, and the P2A anchor. Tx v3
  * (TRUC), locktime 0, sequence 0xFFFFFFFE — all per the fw builder.
  */
-export function buildPeginPsbt(hashlock: Buffer): PeginPsbtFixture {
+export function buildPeginPsbt(hashlock: Buffer, prepeginTxidInternal: Buffer): PeginPsbtFixture {
   const depositor = hexToBuffer(DEPOSITOR_XONLY_HEX);
   const vp = hexToBuffer(VP_KEY_HEX);
   const keepers = [hexToBuffer(KEEPER_KEY_HEX)];
@@ -360,7 +430,7 @@ export function buildPeginPsbt(hashlock: Buffer): PeginPsbtFixture {
   psbt.setVersion(3); // TRUC (BIP-431)
   psbt.setLocktime(0);
   psbt.addInput({
-    hash: PREPEGIN_TXID_INTERNAL, // Buffer form: used as-is (internal order)
+    hash: prepeginTxidInternal, // Buffer form: used as-is (internal order)
     index: HTLC_VOUT,
     sequence: 0xfffffffe,
     witnessUtxo: { script: htlcScriptPubKey, value: HTLC_VALUE_SATS },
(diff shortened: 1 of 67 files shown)


packages/babylon-ts-sdk/src/tbv/core/managers/pegin/verifyPopWitness.ts
```diff
@@ -0,0 +1,185 @@
...
+      throw new Error(
+        "proof of possession signature does not verify against the depositor key",
+      );
+    }
+    return { kind: "p2tr-verified" };
+  }
+
+  if (items.length === P2WPKH_WITNESS_ITEMS) {
+    return { kind: "p2wpkh-unverified" };
+  }
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — P2WPKH PoP verification is tracked in #2284 (Ledger emits P2TR only; two-item witnesses pass through with an explicit unverified verdict)

packages/babylon-ledger-vault-signer/src/policyPsbt.ts
```diff
@@ -0,0 +1,159 @@
...
+  const depositorKey = Buffer.from(depositorXOnlyHex, "hex");
+  psbt.data.inputs.forEach((input, i) => {
+    if (input.tapInternalKey && Buffer.from(input.tapInternalKey).equals(depositorKey)) {
+      psbt.updateInput(i, {
+        tapBip32Derivation: [
+          { masterFingerprint: fingerprint, pubkey: depositorKey, path: pathToString(depositorPath), leafHashes: [] },
+        ],
+      });
+    }
+  });
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — prepareSignPsbt already rejects inputs whose tapInternalKey is not the depositor key before any device I/O (same equality the augmenter uses), so this shape cannot reach the device

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -336,6 +364,35 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+   * Two silent reads, cached per connection; a failure clears the cache.
+   */
+  private getPolicyContext(): Promise<PolicyContext> {
+    if (!this.policyContextPromise) {
+      const send = this.requireSender();
+      const read = (async (): Promise<PolicyContext> => {
+        const [masterFingerprintHex, accountXpub] = await Promise.all([
+          getMasterFingerprintHex(send),
+          getExtendedPublicKey(send, this.accountPath, toNetwork(this.network).bip32),
+        ]);
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — DMK 1.7.1 serializes per session (DeviceSession.sendApdu routes through IntentQueueService.enqueue), so the two reads cannot overlap on the wire

packages/babylon-ledger-vault-signer/src/walletPolicy.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+  if (!Number.isInteger(coinType) || coinType < 0 || !Number.isInteger(accountIndex) || accountIndex < 0) {
+    throw new Error("coinType and accountIndex must be non-negative integers");
+  }
+  if (accountXpub.length === 0) {
+    throw new Error("accountXpub must be the device's extended key string");
+  }
+  const keyInfo = `[${masterFingerprintHex}/${BIP86_PURPOSE}'/${coinType}'/${accountIndex}']${accountXpub}`;
+  const walletIdHex = new DefaultWalletPolicy(DEFAULT_TAPROOT_DESCRIPTOR_TEMPLATE, keyInfo).getId().toString("hex");
+  return { descriptorTemplate: DEFAULT_TAPROOT_DESCRIPTOR_TEMPLATE, keyInfo, walletIdHex };
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/walletPolicy.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  it("rejects an index that would overflow hardened derivation", () => {
+    expect(() =>
+      buildDefaultTaprootPolicy({
+        masterFingerprintHex: FINGERPRINT,
+        coinType: 1,
+        accountIndex: 0x80000000,
+        accountXpub: TPUB,
+      }),
+    ).toThrow(/coinType and accountIndex/);
+  });
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ts-sdk/docs/api/deposit-terms.md
```diff
@@ -799,6 +819,8 @@ function supportsDepositApproval(wallet): wallet is BitcoinWallet & DepositTerms
 Defined in: [packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts)
 
 True when the wallet implements [DepositTermsApprover.approveDepositTerms](#approvedepositterms).
+Narrows to the whole interface: a provider with one method and not the other
+is a provider bug, not a shape this seam supports.
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -541,6 +542,14 @@ export function useDepositFlow(
         // commit pass drives the wallet popup(s).
         setPeginSigningProgress({ completed: 0, total: vaultAmounts.length });
 
+        // Approval (policy) wallets dictate the change address (BIP-86 change
+        // branch); software wallets keep change on the connected address.
+        const prePeginChangeAddress = supportsDepositApproval(
```
gbarkhatov, line 548: **Pre-PegIn change now lands on an address this app never scans.**

For approval wallets the change output routes to the wallet's BIP-86 change branch (`m/86'/coin'/0'/1/0`), but UTXO discovery here is single-address: `useUTXOs.ts:56` calls `getAddressUtxos(btcAddress)` for the connected *receive* address only, and that one set feeds the displayed balance (`useBtcWalletState.ts:13`), the deposit form (`useDepositPageForm.ts:351`), and the spendable set handed to UTXO selection. Nothing in the tree calls `getChangeAddress` for discovery — its only three consumers are this build path, the `preparePegin` gate, and the seam forward.

Concretely: deposit 0.5 of a 1 BTC UTXO on Ledger and the full 1 BTC leaves the scanned address, the 0.5 change is invisible in the UI, and a second deposit sees a 0 spendable balance. The funds are safe and recoverable in Ledger Live, but as written the Ledger path is effectively one deposit per funded receive address.

Either scan the change branch (index 0 alone would cover it, since `FIRST_CHANGE_INDEX` is pinned), or document this as a known limitation before the flag comes off default-off.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -1766,11 +1797,15 @@ export class PeginManager {
       "bip322-simple",
     );
 
-    return {
-      btcPopSignature: normalizePopSignature(raw),
-      depositorEthAddress,
+    const btcPopSignature = normalizePopSignature(raw);
+    // Fail before the Ethereum registration: vaultd rejects a bad PoP permanently.
+    verifyPopWitness(
```
gbarkhatov, line 1818: **This gate is strictly stricter than `main` for every wallet, not just Ledger — and its verdict is dropped.**

`signProofOfPossession` is the shared path for OKX / UniSat / OneKey / Keystone too. A taproot software wallet whose BIP-322 witness `verifyBip322Simple` doesn't reproduce (different sighash handling, a non-BIP-86 taproot address, a script-path signature) now cannot deposit at all, where before it proceeded. That is very likely the right call given vaultd rejects a bad PoP permanently — but it deserves a real-wallet matrix run before merge, especially since the module header says "No wallet test (hardware or software) proves a returned PoP validates — this is the first place anything checks it."

Separately: the `PopWitnessVerdict` return is discarded here, and the type has no other production consumer. So the P2WPKH pass-through is silent — `verifyPopWitness`'s module doc says "the verdict says so", but nothing reads it. Either branch/log on it or drop the type (ZERO DEAD CODE).

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -336,6 +364,35 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+        const policy = buildDefaultTaprootPolicy({
```
gbarkhatov, line 400: `getPolicyContext` takes two independent device reads — the depositor key at `.../0/0` via `getDevicePubkeyHex`, and the account xpub at `m/86'/coin'/0'` — and never asserts they agree.

A cheap `HDKey.fromExtendedKey(accountXpub, bip32).deriveChild(0).deriveChild(0)` compared against the cached depositor x-only key would pin it; `deriveChangeXOnlyHex` already imports `HDKey` for exactly this shape.

Without it, an inconsistency between the two reads binds the wallet policy — and therefore the `TAP_BIP32_DERIVATION` this provider writes — to a different key than the one the intent and the expected-signature table are pinned to. It first surfaces as an opaque `0x6A80` after the user has approved on-device, which is precisely the failure class the rest of this file goes to lengths to pre-empt host-side.

packages/babylon-ledger-vault-signer/src/policyPsbt.ts
```diff
@@ -0,0 +1,159 @@
...
+    throw new Error(`${name} must have exactly ${BIP86_PATH_LEVELS} levels, got ${levels.length}`);
+  }
+  for (const level of levels) {
+    if (!Number.isInteger(level) || level < 0 || level > U32_MAX) {
+      throw new Error(`${name} levels must be integers in 0..0xffffffff`);
+    }
+  }
+  // Any other hardening shape is unmatchable against the policy expression
+  // `@0/<0;1>/*`, so the device could never mark the input/output internal.
+  const accountHardened = levels.slice(0, PATH_ACCOUNT_LEVELS).every((level) => (level & HARDENED) !== 0);
```
gbarkhatov: `assertBip86Path` checks that the first three levels are hardened but never that `levels[0] === 86 | HARDENED` — even though its own error message on line 46 says "must harden purpose'/coin'/account'". And `assertPolicyPathPair` only compares `change.path` against `depositorPath`; nothing binds either path to the policy's key origin, which is built independently in `buildDefaultTaprootPolicy` from `coinType` / `accountIndex`.

So a caller that builds the policy with `coinType: 1, accountIndex: 0` and augments with `[84', 0', 0', 0, 0]` passes every host gate here; the device then can't match the derivation against `@0/<0;1>/*`, can't mark the input internal, and answers `0x6A80` after the intent screens have been approved.

The provider is self-consistent today, but this is the function whose job is to make that unrepresentable. Taking the policy (or its origin levels) as a parameter and comparing would close it.

packages/babylon-ledger-vault-signer/src/policyPsbt.ts
```diff
@@ -0,0 +1,159 @@
...
+    throw new Error("masterFingerprintHex must be 8 lowercase hex characters");
+  }
+  if (change && !X_ONLY_HEX_RE.test(change.xOnlyHex)) {
+    throw new Error("change.xOnlyHex must be 64 lowercase hex characters");
+  }
+  assertPolicyPathPair(depositorPath, change?.path);
+  const psbt = Psbt.fromHex(psbtHex);
+  const fingerprint = Buffer.from(masterFingerprintHex, "hex");
+  const depositorKey = Buffer.from(depositorXOnlyHex, "hex");
+  psbt.data.inputs.forEach((input, i) => {
```
gbarkhatov, line 126: Asymmetric with the change branch below: a change script matching no output throws (line 146, with a good comment explaining exactly why), but an input that doesn't match is silently left unmarked.

This module's own header states that `_validate_prepegin` requires **every** input internal. An input carrying neither `TAP_LEAF_SCRIPT` nor `TAP_INTERNAL_KEY` and a non-depositor prevout is also skipped by `buildExpectedSignatureTable`'s ownership scan, so it reaches the device unmarked and dies mid-ceremony after the approval screens have been burned.

Unreachable through today's SDK builders (a Pre-PegIn is funded entirely from depositor UTXOs), but the symmetric `marked === psbt.data.inputs.length` assertion is one line and matches the invariant the header already claims.

packages/babylon-ledger-vault-signer/src/popPsbt.ts
```diff
@@ -0,0 +1,124 @@
...
+/** Build the PoP to_sign PSBT (v0 hex) for `prepareSignPsbt` in wallet-policy mode. */
+export function buildPopPsbtHex(params: BuildPopPsbtParams): string {
+  const { message, depositorXOnlyHex, masterFingerprintHex, depositorPath } = params;
+  if (!X_ONLY_HEX_RE.test(depositorXOnlyHex)) {
+    throw new Error("depositorXOnlyHex must be 64 lowercase hex characters");
+  }
+  if (!FINGERPRINT_HEX_RE.test(masterFingerprintHex)) {
+    throw new Error("masterFingerprintHex must be 8 lowercase hex characters");
+  }
+  if (depositorPath.length !== BIP86_PATH_LEVELS) {
```
gbarkhatov: This validates `depositorPath.length` only, while `policyPsbt.assertBip86Path` — same 5-level BIP-86 path, same purpose — additionally enforces u32 range and the hardened/unhardened shape.

`pathToString` at line 82 is a verbatim copy of `policyPsbt.ts:72-74`, and `l & HARDENED` / `>>> 0` coerce silently: `[86+H, 0+H, 0+H, 0, 2**32]` renders as `m/86'/0'/0'/0/0` — a different address index than the caller asked for, with no throw. Unhardened or change-branch paths are likewise accepted here and only die on-device, after the user has reached the approval screen, which is exactly what the "any throw here means zero device I/O" contract exists to prevent.

The two copies have already drifted (one has a range gate in front of it, the other doesn't). Sharing one encoder plus one gate would keep them from drifting further. `popPsbt.test.ts:116-123` currently only exercises the length arm (`[1, 2]`).

packages/babylon-ledger-vault-signer/src/signPsbtPrepare.ts
```diff
@@ -303,9 +331,18 @@ export function prepareSignPsbt(params: PrepareSignPsbtParams): PreparedSignPsbt
   interpreter.addKnownList(merkelized.inputMapCommitments);
   interpreter.addKnownList(merkelized.outputMapCommitments);
 
+  // Policy mode: the device fetches the policy serialization, then walks the
+  // key-info tree and the template (`base:sign_psbt.c` wallet header checks)
+  // before signing internal inputs. Seed all three preimages (vendored
+  // `addKnownWalletPolicy`); no-policy mode seeds nothing and keeps wallet_id zero.
+  const walletId = walletPolicy ? Buffer.from(walletPolicy.walletIdHex, "hex") : NO_WALLET_POLICY_ID;
```
gbarkhatov, line 341: Nothing asserts that a table containing key-path expectations requires a `walletPolicy`. Per the `PrepareSignPsbtParams` JSDoc, without one the base app skips `sign_internal_inputs` and the device answers SW_OK with no yield — so an all-key-path PSBT prepared in no-policy mode is guaranteed to fail, but only at the collector's completion check, i.e. after the full device round-trip.

The table is built at line 307, so the check costs nothing and would fit the module's stated "any throw here means zero device I/O" contract.

Note that `popPsbt.test.ts:126-130` currently pins the permissive behaviour — it calls `prepareSignPsbt` on a PoP PSBT with no policy and asserts success — so that test would need to move to the policy form.

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts
```diff
@@ -105,13 +105,43 @@ export interface DepositTerms {
...
 }
 
-/** True when the wallet implements {@link DepositTermsApprover.approveDepositTerms}. */
+/**
+ * Probes {@link DepositTermsApprover.approveDepositTerms} alone but narrows to
+ * the whole interface: a conforming approval wallet implements all of
+ * {@link DepositTermsApprover}, `getChangeAddress` included, so a provider with
+ * one method and not the other is a provider bug, not a shape this seam supports.
+ */
 export function supportsDepositApproval(
```
gbarkhatov, line 142: This probes `approveDepositTerms` alone but narrows to `BitcoinWallet & DepositTermsApprover`, which now includes the non-optional `getChangeAddress` — and `PeginManager.ts:687` calls it straight off the narrowed value.

Not reachable today (the one implementor has both, and `forwardDepositApproval` was correctly updated to forward it), but the next provider that implements only `approveDepositTerms` gets `TypeError: wallet.getChangeAddress is not a function` in the middle of `preparePegin`, after the pubkey read.

The JSDoc argues this away as "a provider bug, not a shape this seam supports" — but one more `typeof` turns it into an actionable error at the seam instead of a runtime crash three frames down.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyKeyPathSchnorrSignature.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  if (returned.data.inputs.length !== requested.data.inputs.length) {
+    throw new Error(
+      `Returned PSBT input count ${returned.data.inputs.length} differs from the ` +
+        `requested ${requested.data.inputs.length}.`,
+    );
+  }
+
+  requested.data.inputs.forEach((input, inputIndex) => {
+    if (!isKeyPathEligible(input)) return;
```
gbarkhatov, line 227: `isKeyPathEligible` requires `tapInternalKey` **and** a P2TR `witnessUtxo`, so for a P2WPKH depositor every input is skipped, the function returns `void`, and the caller cannot distinguish "all signatures verified" from "nothing was verified".

P2WPKH depositors are first-class on this path (`isAddressFromPublicKey` accepts native segwit), and both call sites — `PeginManager.ts:1213` and `vaultPeginBroadcastService.ts:270` — read this as *the* CLAUDE.md §8 far-side check, with comments to that effect.

Not a regression (there was no check at all before), and a hard `verifiedCount === 0 → throw` would break P2WPKH depositors outright. But it's worth making legible: return or expose a verified count so "zero eligible inputs" is a visible outcome rather than a silent pass, and so the P2WPKH verifier follow-up has something concrete to hang off.

packages/babylon-ledger-vault-signer/src/policyPsbt.ts
```diff
@@ -0,0 +1,159 @@
...
+  return changeOutputIndices(Psbt.fromHex(psbtHex), changeXOnlyHex).length > 0;
+}
+
+export interface AugmentPsbtForWalletPolicyParams {
+  readonly psbtHex: string;
+  readonly depositorXOnlyHex: string;
+  readonly masterFingerprintHex: string;
+  readonly depositorPath: readonly number[];
+  /** The change output's key and path; omit when the PSBT carries no change. */
+  readonly change?: { readonly xOnlyHex: string; readonly path: readonly number[] };
```
gbarkhatov: `change.xOnlyHex` and `change.path` arrive as two unrelated values and nothing asserts they describe the same key. Passing a key derived at index 0 alongside `path = [..., 1, 1]` writes a `TAP_BIP32_DERIVATION` claiming one key next to a `tapInternalKey` that is another; the host accepts it and only the device catches it.

The provider derives the key with `FIRST_CHANGE_INDEX` and passes `this.changePath` from a separate constant (`provider.ts:806,817`), so the two are coupled by convention alone. Taking `accountXpub` + `addressIndex` and deriving both internally — `deriveChangeXOnlyHex` is right here in this module — would make the divergence unrepresentable.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -780,10 +947,73 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+          throw new WalletError({
+            code: ERROR_CODES.WALLET_METHOD_NOT_SUPPORTED,
+            message: `${WALLET_PROVIDER_NAME} signs BIP-322 (bip322-simple) messages only; ${type} is not supported.`,
+            wallet: WALLET_PROVIDER_NAME,
+          });
+        }
+        const ctx = await this.gateSignContext(false);
+        const { policy, masterFingerprintHex } = await this.getPolicyContext();
+        this.assertSameConnection(ctx.generation);
+        const psbtHex = buildPopPsbtHex({
```
gbarkhatov, line 986: `buildPopPsbtHex` is the one staging step in `signMessage` not wrapped in `toStagingWalletError` — the very next call is. Its `throw new Error(...)` escapes as an unclassified error with no `code` / `wallet`, so PoP staging failures skip the error routing every other pre-I/O rejection in this file goes through.

Same shape at `getChangeAddress` (line 418): `deriveChangeXOnlyHex` and `buildDefaultTaprootPolicy` throw plain `Error`s that leave this provider untyped. The APDU reads behind them are mapped by `withWalletErrorMapping`; the derivation and policy-construction throws are not.

packages/babylon-ledger-vault-signer/src/walletPolicy.ts
```diff
@@ -0,0 +1,66 @@
...
+    !Number.isInteger(coinType) ||
+    coinType < 0 ||
+    coinType > MAX_HARDENED_CHILD_INDEX ||
+    !Number.isInteger(accountIndex) ||
+    accountIndex < 0 ||
+    accountIndex > MAX_HARDENED_CHILD_INDEX
+  ) {
+    throw new Error("coinType and accountIndex must be non-negative integers in 0..0x7fffffff");
+  }
+  if (accountXpub.length === 0) {
```
gbarkhatov: `accountXpub` is checked for emptiness only. `WalletPolicy.serialize()` encodes the key-info string with `Buffer.from(k, "ascii")`, which masks anything above `0x7F`, and there's no base58 or version check — so a garbage string still yields a well-formed `walletIdHex` that the device can never be served a matching preimage for, failing mid-loop rather than here.

Related: the origin path is assembled here from `(coinType, accountIndex)` while the xpub comes from a separate `getExtendedPublicKey(this.accountPath)` read in the provider. Deriving the origin from the same source as the xpub would remove the chance of the two disagreeing.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -674,9 +676,23 @@ export class PeginManager {
...
-    // would drain the change after signing. Bind once at entry using the
-    // pubkey snapshot above (no second wallet read).
-    if (
+    // would drain the change after signing. Bind once at entry — against the
+    // wallet's own change branch when it has one, else the pubkey snapshot.
+    if (supportsDepositApproval(this.config.btcWallet)) {
+      // Approval (policy) wallets own their change branch: the device accepts a
+      // change output only on `.../1/i`, which is not derivable from the receive
+      // key. Any other change address fails mid-ceremony on the device, so this
+      // gate closes that window before the first device I/O.
```
gbarkhatov: "before the first device I/O" isn't accurate for the wallet this branch exists for. `getPublicKeyHex()` at line 672 is already a device read, and `getChangeAddress` on the next line is itself wrapped in `withDeviceOperation` and triggers the two `getPolicyContext` APDUs.

The gate is still worth having — it runs before the *ceremony*, which is the part that costs the user an approval — but the comment claims something stronger than it delivers.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
```diff
@@ -50,21 +74,60 @@ const SPECULOS_URL = process.env.SPECULOS_URL ?? "";
...
+/** The policy's key origin is DEPOSITOR_PATH's account prefix, m/86'/1'/0'. */
+const ACCOUNT_PATH_LEVELS = 3;
+const POLICY_COIN_TYPE = 1;
+const POLICY_ACCOUNT_INDEX = 0;
+/** Change lives at `account/1/0` — BIP-86 change branch, first address. */
+const BIP86_CHANGE_BRANCH = 1;
+const CHANGE_ADDRESS_INDEX = 0;
+
+/**
+ * BIP-322 to_sign shape, restated rather than imported from `popPsbt.ts` so the
```
gbarkhatov: This says the BIP-322 `to_sign` shape is "restated rather than imported from `popPsbt.ts` so the verification below is independent of the builder under test" — but line 431 imports `bip322ToSpendTxid` from exactly that module. Only the version / locktime / sequence constants are restated.

In practice the device *is* the independent oracle here: the firmware rebuilds `to_spend` from the message and its own BIP-86 tweaked key and compares against `PSBT_IN_PREVIOUS_TXID`, so a wrong byte order can't pass this stage. That's a stronger claim than the comment makes and worth stating instead.

It's also the answer to how `CANONICAL_TO_SPEND_TXID` in `popPsbt.test.ts:20` is pinned in raw rather than display order — a self-consistent unit test alone can't establish that, since it compares the code against itself.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyKeyPathSchnorrSignature.ts
```diff
@@ -0,0 +1,313 @@
...
+}
+
+/**
+ * Bitcoin CompactSize decode of a witness stack: item count, then (len, bytes)*.
+ * Strict: reads are bounds-checked and every byte must be accounted for.
+ */
+function witnessItems(witness: Uint8Array): Uint8Array[] {
+  let offset = 0;
+
+  const readCompactSize = (): number => {
```
gbarkhatov: Second CompactSize decoder in this PR. `verifyPopWitness.ts:60-93` does the same job and rejects non-minimal encodings — matching rust-bitcoin's `NonMinimalVarInt`, which is what vaultd applies — while this one accepts them. Different strictness, different constant style (`COMPACT_SIZE_UINT16_BYTES` here vs. bare `offset + 3` / `offset + 5` there), landing in one PR.

No security impact (the item still has to be 64 or 65 bytes and the signature still has to verify), but extracting one shared primitive would stop the two from drifting further apart.

Minor, same function: the `"finalScriptWitness item length out of range"` message at line 210 also fires on the item-*count* read at line 213.

packages/babylon-ledger-vault-signer/src/popPsbt.ts
```diff
@@ -0,0 +1,124 @@
...
+ */
+
+import { crypto as bcrypto, Psbt, Transaction } from "bitcoinjs-lib";
+import { Buffer } from "buffer";
+
+import { bip86OutputScript } from "./expectedSignatures";
+
+const BIP322_TAG = "BIP0322-signed-message";
+/** `0xFC ‖ 0x06 ‖ "bvault" ‖ 0x00` — `BIP322_PSBT_PROP_POP_MSG_KEY` (`bip322.c:10-11`; doc `bip322.h:22-29,37-38`). */
+export const POP_MESSAGE_PROPRIETARY_KEY = new Uint8Array([0xfc, 0x06, 0x62, 0x76, 0x61, 0x75, 0x6c, 0x74, 0x00]);
```
gbarkhatov: `POP_MESSAGE_PROPRIETARY_KEY` is re-exported from the package index (`index.ts:57`) but has no consumer outside this module and its own test. Per the ZERO DEAD CODE policy, either drop it from the public surface or keep it module-private.

`bip322ToSpendTxid` is in a similar position minus the index export — used only by `popPsbt.test.ts` and the Speculos e2e — which is more defensible, but worth a note if you're touching one.

gbarkhatov (dismissed):
Approving. The device-protocol work is careful and unusually well cited — the wallet-policy serialization, the `walletIdHex` mismatch gate, the BIP-322 `to_spend` byte order, the `hashType` plumbing and the key-path classification all hold up under a line walk.

Two things I'd want resolved before the feature flag comes off default-off, both left as inline comments:

1. **Pre-PegIn change lands on an address the dApp never scans.** UTXO discovery in the vault app is single-address, so after one partial deposit on Ledger the change is invisible in the balance and unusable for a second deposit. Funds are recoverable in Ledger Live, but as written the Ledger path is one deposit per funded receive address.
2. **The new PoP gate is strictly stricter than `main` for every wallet, not just Ledger.** `signProofOfPossession` is shared with OKX / UniSat / OneKey / Keystone, and a taproot software wallet whose witness our verifier doesn't reproduce now cannot deposit at all. Almost certainly the right call given vaultd rejects a bad PoP permanently — but the module header itself says no wallet test proves a returned PoP validates, so a real-wallet matrix run before merge would be worth it.

The rest of the inline comments are hardening and consistency: the policy/derivation binding gaps in `policyPsbt.ts` and `provider.ts`, the weaker path validation in `popPsbt.ts` versus its sibling, and a handful of low-priority notes.

**Three notes that couldn't be left inline** (not in a diff hunk):

- The PR description is empty. This touches CLAUDE.md §7 (Ledger vault signer package), so the intent-ceremony and firmware-citation context that lives in the commit messages should be in the PR body where reviewers will find it.
- `services/vault/src/config/featureFlags.ts:136` still says "signing is not implemented yet (signPsbt/signPsbts/signMessage throw)". Stale before this PR, but this is the change that finishes the signing story — good moment to fix it.
- The Speculos e2e is `describe.skipIf(SPECULOS_URL === "")` and `SPECULOS_URL` appears nowhere under `.github/`, so the new device paths have no CI-executed coverage. Reasonable as a deliberate opt-in given it needs a container and a firmware build — worth stating explicitly in the PR body so it isn't mistaken for automated coverage.

**Checked and found sound:** the policy serialization / wallet-id path and its zero HMAC; `prepareSignPsbt` policy-mode preimage seeding and the `walletIdHex` mismatch gate; `augmentPsbtForWalletPolicy`'s input matching against `buildExpectedSignatureTable`'s key-path classification (both agree on `tapInternalKey === depositorKey`, and a foreign internal key already throws at table build); the BIP-322 `to_spend` txid byte order and the new `hashType` parameter in `verifyBip322Simple`; `assertReturnedKeyPathSignatures` running against the *requested* (non-augmented) PSBT — `getPsbtInputFields` does set `tapInternalKey` on P2TR inputs at both call sites, so it is not a silent no-op for taproot wallets; `forwardDepositApproval` now forwarding `getChangeAddress` at all four `{...wallet}` wrapper sites; the `classifySignFailure` extraction preserving the previous mirror-reset semantics for stale / aborted / other failures; and the `esmExternals: ["bitcoinjs-lib"]` build fix being consistent with `rollupOptions.external`.

**Claims I checked and am explicitly not raising:** the missing `0x` guard in `decodeWitnessItems` (the param is viem `Hex` and the sole caller passes `normalizePopSignature`'s output, which always emits the prefix); `DEFAULT_TAPROOT_DESCRIPTOR_TEMPLATE` as dead code (it is used in the same module); the manager-level stubbing of the key-path verifier (real verification is covered by the new primitive test, and the stub mirrors the existing script-path one); and a missing change gate on the resume/broadcast path (a cross-wallet resume already fails host-side at `buildExpectedSignatureTable`).

packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
```diff
@@ -320,6 +335,64 @@ export function termsToIntent(terms: DepositTerms): IntentFixture {
...
+  psbt.setVersion(2); // fw requires tx_version >= 2
+  psbt.setLocktime(0);
+  for (let i = 0; i < PREPEGIN_INPUT_COUNT; i++) {
+    psbt.addInput({
+      hash: Buffer.alloc(32, PREPEGIN_PREVOUT_HASH_BASE + i),
+      index: i,
+      // SEQUENCE_FINAL: the fw (fix/audit-findings > e2d0c45b) requires Pre-PegIn
+      // inputs present AND final (no RBF/CSV) — and the toolkit's funding path
+      // emits bitcoinjs' default 0xffffffff (fundPeginTransaction.ts:148).
+      sequence: 0xffffffff,
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/policyPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/policyPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/popPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/signPsbtPrepare.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/deposit-terms/depositTerms.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyKeyPathSchnorrSignature.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/policyPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/walletPolicy.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyKeyPathSchnorrSignature.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/popPsbt.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: not planning — tracked in #2283

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
jrwbabylonlab: not planning — vaultd rejects a bad PoP permanently anyway

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: not planning — those guards can't fire

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

gbarkhatov (approved):

PR #2282, opened 2026-08-19
fix(vault): save the artifact payload, not its json-rpc envelope

kirugan:
How it is **now**: when user clicks on button to download artifacts they receive full json response including json-rpc i.e.
```
{
  "jsonrpc": "2.0",
  "result":  <real data for artifacts>,
  "id": 1
}
```

It doesn't hurt (since data is still there - and I was able to claim vault using artifacts). It just looks a bit strange.

How it should **be** (changes in this PR): user downloads file that contains data only from the result field (no json-rpc envelope)

**Notes:**
- in theory it could be done on vault-provider-proxy side, but it's wrong: the response should be always consistent (same json-rpc envelope for errors and good responses)

### DIFF
--- a/services/vault/src/services/artifacts/__tests__/artifactDownloadService.test.ts
+++ b/services/vault/src/services/artifacts/__tests__/artifactDownloadService.test.ts
@@ -70,6 +70,15 @@ function validEnvelope(): string {
 }
 
 /**
+ * What the saved file must contain: the envelope's `result` value alone.
+ * Bytes are sliced from the wire, so the file is the literal source span —
+ * which for a `JSON.stringify`d envelope is exactly the payload re-stringified.
+ */
+function expectedSavedPayload(result: unknown = VALID_ARTIFACT_RESULT): string {
+  return JSON.stringify(result);
+}
+
+/**
  * Build a Response backed by a real ReadableStream, with Content-Length only
  * when asked. A stream body never auto-populates the header, which is what
  * lets the header-absent fallback be tested deterministically.
@@ -144,9 +153,10 @@ describe("fetchAndDownloadArtifacts", () => {
   });
 
   describe("a valid response", () => {
-    it("writes the whole body and commits it", async () => {
-      const body = validEnvelope();
-      vi.mocked(fetch).mockResolvedValueOnce(streamingResponse(body));
+    it("writes the unwrapped payload and commits it", async () => {
+      vi.mocked(fetch).mockResolvedValueOnce(
+        streamingResponse(validEnvelope()),
+      );
       const { target, commit, discard, savedBytes } = fakeSaveTarget();
 
       await fetchAndDownloadArtifacts(
@@ -156,14 +166,16 @@ describe("fetchAndDownloadArtifacts", () => {
         target,
       );
 
-      expect(new TextDecoder().decode(savedBytes())).toBe(body);
+      expect(new TextDecoder().decode(savedBytes())).toBe(
+        expectedSavedPayload(),
+      );
       expect(commit).toHaveBeenCalledTimes(1);
       expect(discard).not.toHaveBeenCalled();
     });
 
-    it("returns a receipt describing what was saved", async () => {
+    it("returns a receipt describing the saved file, not the wire body", async () => {
       const body = validEnvelope();
-      const byteLength = new TextEncoder().encode(body).byteLength;
+      const saved = new TextEncoder().encode(expectedSavedPayload());
       vi.mocked(fetch).mockResolvedValueOnce(streamingResponse(body));
       const { target } = fakeSaveTarget();
 
@@ -174,12 +186,17 @@ describe("fetchAndDownloadArtifacts", () => {
         target,
       );
 
+      // The digest has to be reproducible by `sha256sum` on the file the user
+      // ends up with, so it covers the payload rather than the envelope.
       expect(outcome).toEqual({
         filename: "babylon-vault-artifacts-aaaaaaaa.json",
-        byteLength,
-        sha256: bytesToHex(sha256(new TextEncoder().encode(body))),
+        byteLength: saved.byteLength,
+        sha256: bytesToHex(sha256(saved)),
         method: "file-system-access",
       });
+      expect(outcome.byteLength).toBeLessThan(
+        new TextEncoder().encode(body).byteLength,
+      );
     });
 
     it("reassembles a body split across many chunks", async () => {
@@ -205,7 +222,9 @@ describe("fetchAndDownloadArtifacts", () => {
         target,
       );
 
-      expect(new TextDecoder().decode(savedBytes())).toBe(body);
+      expect(new TextDecoder().decode(savedBytes())).toBe(
+        expectedSavedPayload(),
+      );
     });
 
     it("accepts a result body that nests an error key", async () => {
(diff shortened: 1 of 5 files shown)


services/vault/src/services/artifacts/artifactDownloadService.ts
```diff
@@ -17,6 +17,15 @@
  * Bytes are validated, hashed, and written chunk-by-chunk in a single pass,
  * so the full body is never held in memory on the File System Access path.
  *
+ * What lands on disk is the envelope's `result` value alone, not the JSON-RPC
+ * response around it. `jsonrpc` and `id` are transport framing with no meaning
+ * in a saved file, and the unwrapped shape is what the rest of the toolchain
+ * already reads — btc-vault's `tools/artifact-downloader` writes the identical
```
gbarkhatov: The reasons given here don't hold up against btc-vault. There is no `tools/artifact-downloader`: `contents/tools` on `main` returns 404, and no branch, PR, or code search turns it up. The only CLI that calls `requestDepositorClaimerArtifacts` is `crates/depositor-cli/src/command/peg_in_and_borrow.rs`, which checks the artifacts in memory and writes no file.

The only file btc-vault reads for this flow is `WatchtowerArtifacts` (`crates/vaultd/src/cli/command/watchtower/artifacts.rs`). It uses `tx_graph` and `verifying_key` rather than `tx_graph_json` and `verifying_key_hex`. It requires `claim_tx`, `signatures`, `vault_id` and `prover_circuit_version`, and it is written pretty-printed. So the saved file isn't interchangeable with anything the toolchain reads.

`docs/delegated_claim.md` (lines 139-151) lists the three inputs but defines no file layout. The header of `streamingArtifactValidator.ts` (lines 11-18) repeats both claims.

The unwrapping is justified without any of this: `jsonrpc` and `id` mean nothing on disk, and the receipt hash now matches `sha256sum` of the saved file. I'd remove the tool, "identical layout" and "already reads" wording from both headers, or point to a source that exists.

services/vault/src/services/artifacts/__tests__/artifactDownloadService.test.ts
```diff
@@ -234,6 +253,104 @@ describe("fetchAndDownloadArtifacts", () => {
...
+        "babe_sessions",
+        "tx_graph_json",
+        "verifying_key_hex",
+      ]);
+      expect(saved).not.toHaveProperty("jsonrpc");
+      expect(saved).not.toHaveProperty("id");
+      expect(saved).not.toHaveProperty("result");
+    });
+
+    it("matches the layout btc-vault's artifact-downloader writes", async () => {
```
gbarkhatov: The name claims conformance with a btc-vault tool that doesn't exist (see the comment on `artifactDownloadService.ts:23`). The test's only assertion, `toEqual(VALID_ARTIFACT_RESULT)`, is already covered by the exact-string check in "writes the unwrapped payload and commits it" and the key check in the test above. I'd delete it.

services/vault/src/utils/artifactDownloadStorage.ts
```diff
@@ -38,7 +38,7 @@ export interface ArtifactDownloadReceipt {
   peginTxid: string;
   filename: string;
   byteLength: number;
-  /** SHA-256 of the response body, lowercase hex. */
+  /** SHA-256 of the saved file, lowercase hex. */
```
gbarkhatov, line 41: `sha256` and `byteLength` now describe the saved payload instead of the whole response, but `ARTIFACT_RECEIPT_VERSION` (line 28) stays `1`. Receipts written before this change still pass `isValidReceipt`, and their fields describe the envelope.

Nothing reads these fields yet (`useArtifactDownload` only stores them), so nothing breaks now. But users who already downloaded keep a file with the envelope around it, and `hasArtifactsDownloaded` won't ask them to download again. Any later check of the stored hash against the file would fail for them.

Your call: bump the version to `2` (the comment at line 24 already says re-downloading is safe), or say in the PR that you're keeping `1` on purpose.

services/vault/src/services/artifacts/artifactDownloadService.ts
```diff
@@ -238,6 +261,15 @@ export async function downloadArtifactsFromResponse(
       Object.keys(validated.result.babe_sessions),
       binding,
     );
+
+    // Unreachable while `finish()` requires a schema-valid result object, but
+    // a receipt is only meaningful if it describes real bytes on disk — so
+    // the invariant is asserted rather than assumed across future edits.
+    if (written === 0) {
```
gbarkhatov: Nit: the comment says this is unreachable, and no test covers it. `finish()` requires `resultSeen`, and the `{` that sets it is always inside the span, so `written` is at least 2 by this point. Under the zero-dead-code rule I'd remove it.

services/vault/src/services/artifacts/__tests__/streamingArtifactValidator.test.ts
```diff
@@ -610,3 +610,72 @@ describe("ArtifactStreamValidator — JSON-RPC error envelopes", () => {
...
+      jsonrpc: "2.0",
+      result: VALID_RESULT,
+      meta: { nested: { deeper: [1, 2, 3] } },
+      id: 7,
+    });
+
+    expect(spannedText(wire, 11)).toBe(JSON.stringify(VALID_RESULT));
+  });
+
+  it("reports no span for an envelope it rejects before result", () => {
```
gbarkhatov: Nit: the name says the validator rejects this input, but `{"jsonrpc":"2.0","id":7,` is a valid prefix and nothing is rejected. What the test checks is that a chunk ending before `result` opens returns `null`. Something like "reports no span for a chunk that ends before result opens" would fit.

gbarkhatov (dismissed):
I found no correctness defect in the span tracking, and nothing here blocks merge. The main problem is the reasoning in the header comments, which cites a btc-vault tool that doesn't exist (see the inline comment). Everything else is low severity or a nit.

This review is at c8920bc, which has the same diff as b646f6b after the rebase. It combines three independent reviews with a randomized test that split the input into chunks at random points 3,000 times. I checked the findings against btc-vault `main` and `vault-provider-proxy`.

Checked and fine, no need to revisit:
- `result` is only recognized at the envelope's top level. Nested `result` keys, duplicate keys, a `result` that isn't an object, and sibling values containing braces cannot open or reopen the span. The span closes only on the result object's own `}`.
- Chunk boundaries are handled: a brace on a chunk's first or last byte, one-byte chunks, and the long-hex fast path, which only runs inside `result`.
- A throw after partial writes discards the file. The commit happens only after `finish()` and the vault-binding check.
- `maxBytes` still limits bytes received off the wire, and progress still counts them. On both save paths, the hash and the file get the same slice.
- `result` comes after `id` on both proxy paths: the gRPC path writes it in that order, and the HTTP passthrough gets jsonrpsee's `jsonrpc, id, result`. Accepting any order is defensive, not a bug.
- The 65,536-byte case in the chunk-size test does split the body, because the fixture's hex alone is 65,536 characters.

CI: `verify` (build, lint, tests) was still pending when I reviewed, so the new tests haven't been shown to pass yet.

services/vault/src/services/artifacts/__tests__/artifactDownloadService.test.ts
kirugan: Fixed

services/vault/src/services/artifacts/artifactDownloadService.ts
kirugan: Fixed

services/vault/src/services/artifacts/__tests__/streamingArtifactValidator.test.ts
kirugan: Fixed

services/vault/src/services/artifacts/artifactDownloadService.ts
kirugan: Fixed

services/vault/src/utils/artifactDownloadStorage.ts
kirugan: Keeping `1` on purpose. Before this change the saved file was the whole response body, so an old receipt's `sha256` and `byteLength` already describe the file the user has on disk — a later hash check passes for both old and new receipts. A bump would only force every depositor to re-download ~1 GB.

gbarkhatov (dismissed):

services/vault/src/services/artifacts/artifactDownloadService.ts
```diff
@@ -220,9 +236,16 @@ export async function downloadArtifactsFromResponse(
         throw new ArtifactDownloadTooLargeError(received, target.maxBytes);
       }
 
-      validator.update(chunk.value);
-      hasher.update(chunk.value);
-      await stream.write(chunk.value);
+      // Only the bytes inside the envelope's `result` are saved; the
+      // JSON-RPC framing is dropped so the file matches what the CLI
```
jonybur: **Loop comment still says the saved file matches what the CLI toolchain reads; the header wording marked fixed on :23 is also still in head.**

The comment above the span write says the JSON-RPC framing is dropped "so the file matches what the CLI toolchain reads". That is the same unsupported claim discussed on line 23: no btc-vault tool reads this layout. The thread there is marked fixed, but at this head the header here (lines 22-25) and the header of streamingArtifactValidator.ts (lines 14-15) still cite `tools/artifact-downloader` and an "identical layout". This loop comment is a third copy that the thread didn't name. Please remove all three and keep only the reason that holds: `jsonrpc` and `id` mean nothing on disk, and the receipt hash now matches `sha256sum` of the saved file.

jonybur (commented):
Comments inline.

jonybur (approved):
Looks good to me.

services/vault/src/services/artifacts/artifactDownloadService.ts
kirugan: fixed

Issue #2059, opened 2026-07-16
[v3] Activity Page


jeremy-babylonlabs:
A dedicated v3 page for the existing activity feed. This epic covers the page layout, standard transaction rows, filters, and empty/loading states while preserving the current `/activity` data flow.

**Every v3 surface ships behind the feature flag in #2014** so `main` remains deployable with the current v2 dashboard throughout the build.

**Design:** [Activity before deposit](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10044-13346) · [Activity after deposit](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10051-26752)

**Area:** `services/vault` (`pages/Activity.tsx`, `components/Activity/ActivityList`, `ActivityCard`, `ActivityEmptyState`, `FilterDropdown`)

## Scope

- Restyle the existing `/activity` page for the v3 shell
- Match the v3 page title, spacing, filter control, and transaction-row system
- Preserve chronological ordering and the current merged pending/completed activity source
- Preserve transaction links, copy actions, amounts, dates, and pending/expired indicators
- Support connected, disconnected, filtered-empty, loading, and no-activity states
- Keep existing activity filtering behavior and wallet-gating semantics

## Out of scope

Liquidation event cards, partial/full liquidation treatment, shared liquidation status badges, and cross-page warning banners remain under the planned Liquidation States epic.

## Acceptance criteria

- [ ] `/activity` matches the before-deposit and after-deposit Figma screens in light and dark
- [ ] Standard activity rows retain their existing data, ordering, explorer links, and copy behavior
- [ ] Filters produce the correct transaction subset and a distinct filtered-empty state
- [ ] Connected, disconnected, loading, and no-activity states render the correct v3 treatment and actions
- [ ] V2 behavior remains unchanged while `ENABLE_V3_UI` is off

## Depends on

- #2017

PR #2286, opened 2026-08-20
feat(vault): per-row USD value and live deposit status on the Activity page

### LINKED ISSUE #2114: [v3] Activity page — per-row USD value
Follow-up to the v3 Activity page (https://github.com/babylonlabs-io/babylon-toolkit/issues/2059). Deferred from the initial restyle because current data doesn't back it.

Figma's "after deposit" frame shows a USD subline under each amount (e.g. `$5.234 USD`). `ActivityLog` carries only a preformatted token amount and symbol — no USD. Implementing it needs a price-source decision (price-at-transaction-time vs current price) and new data plumbing through the activity projection.

Files: `services/vault/src/types/activityLog.ts`, `services/vault/src/components/Activity/ActivityRowV3.tsx`, the activity projection under `services/vault/src/services/activity/`.

### LINKED ISSUE #2116: [v3] Activity page — live 'In use' vs settled status for deposits
Follow-up to the v3 Activity page (https://github.com/babylonlabs-io/babylon-toolkit/issues/2059). Deferred from the initial restyle.

The v3 status is derived from the row type: a settled Deposit shows "In use" (green). This reflects the deposit action, not a live check that the collateral is still locked — a since-withdrawn deposit's row would still read "In use". A precise status needs correlating each deposit with current vault/collateral state.

`services/vault/src/components/Activity/ActivityRowV3.tsx` (`getActivityStatus`).

jeremy-babylonlabs:
Closes #2114
Closes #2116

The last two follow-ups from the v3 Activity page (#2059). One commit each.

## Read this first — the USD value uses the CURRENT price

An activity row's USD sub-line is `amount × today's price`, not the price at the time of the transaction. The activity projection has no historical-price plumbing and no other surface in the app has any either, so a deposit made at $60k BTC shows its value at today's BTC price. Everything else in this PR is independent of that decision — if it is the wrong trade-off, the first commit is the one to hold back.

## Commit 1 — per-row USD value (#2114)

`ActivityAmount` now also carries the amount as a plain number alongside the preformatted display string, set wherever the source knows it unambiguously: the indexer projection (BTC collateral and, when the reserve is known, the borrow/repay asset), the localStorage pending peg-in, and the god-mode demo rows. A borrow/repay row with no resolved reserve has unknown decimals, so it stays unpriced.

`ActivityRowV3` turns that number into the sub-line with the existing `formatUsdValue` helper, priced from `usePrices()` — the same Chainlink read the dashboard, deposit form and vault list already use (`prices` covers BTC/sBTC, ETH/WETH, USDC, USDT, DAI, vBTC). Prices are fetched once for the whole list in `ActivityListWithRefund` and passed down, so no hook is instantiated per row.

A row that cannot be priced renders **no** sub-line — never `$0 USD` or `NaN`. That covers an unknown symbol, a missing numeric amount, and a non-positive product.

## Commit 2 — live "In use" vs settled for deposits (#2116)

The status column derived "In use" from the row *type*, so a withdrawn deposit still claimed its collateral was locked. The row now correlates the deposit against current vault state:

- **Join key: `ActivityLog.vaultId` → `Vault.id`.** `ActivityLog` documents `vaultId` as the only field safe to correlate with vault-keyed state — an indexed row's `id` is its event (`txHash-logIndex-type`), not its vault. It is the same key the expired-deposit refund action already matches on.
- **ACTIVE means in use.** `useVaults(ethAddress)` returns every vault of the depositor with its status (the query is cursor-paginated to completion and filters no status), so a vault that has left `ContractStatus.ACTIVE` — redeemed, withdrawn, liquidated, expired — is no longer holding collateral and the row reads "Done".
- **No new network.** That hook shares its query key with the deposit lifecycle `ActivityListWithRefund` already mounts, so it reads the same cache.
- **No wrong flash.** The active-id set is `undefined` until the vault read resolves; an undefined set — or a row with no vault id to correlate — keeps the existing type-derived label. Pending and expired still take precedence over both branches.

## Figma verification

Read live from *TBV v.3 Premium Design* (`NgDfn7fVnkQZP0XH9Uki63`), not from the local snapshot docs.

| Node | What it dictated |
| --- | --- |
| `13164:77269` + `13164:77644` (Activity, "After deposit" section, Dashboard page — dark and light) | The populated Activity frames. Every row carries a USD sub-line under its amount, on BTC rows **and** stable-side borrow/repay rows alike, so both get the treatment. |
| `13164:77311` (a row's amount + USD pair) | Sub-line format `$<value> USD` in *Typography/Desktop/caption* — Px Grotesk 12px, line-height 1.4, letter-spacing 0.4px, `text/secondary`. Implemented as `text-xs leading-[1.4] tracking-[0.4px] text-accent-secondary` with `formatUsdValue`. |
| `489:3270` (`Activity Currency`, BTC/USDC × Default/Liquidated) | Icon + amount only — the sub-line lives at row level, not in this component. Confirms the amount cell is where it belongs. |

### Conflicts resolved in Figma's favour, and one deliberately not

- **Three of the four node ids on the tracker no longer exist** (`10051-26752`, `10088-74585`, `10088-76983`). `10044-13346` still exists but is now the *empty-state* Activity frame. The populated frames are the two listed above — and #2114's own text points at the same "after deposit" frame.
- **Figma's current Activity row is a different layout from the shipped one:** transaction type + asset symbol at the left with a status *chip* (`Failed` / `Refund`) beside it, then the explorer link, then the time, then a right-aligned amount with the USD sub-line under it — and settled rows dimmed. There is no status-dot column at all.
  - For the USD sub-line I followed Figma's typography and format exactly but kept it under the amount **in the shipped 5-column row**. Re-cutting the row into Figma's layout is neither issue's scope; it wants its own ticket.
  - Because Figma no longer has a status column, it dictates no label for a settled deposit. Rather than invent one I reused the existing `COPY.activity.statusDone` — "Done", green dot, the same label every other settled action already shows. Worth a look: if the intent is distinct wording ("Withdrawn", "Settled"), it is a one-line copy change.
- **Figma also shows USD sub-lines on liquidation child rows.** Those render through `LiquidationGroupCardV3` / `LiquidationChildRow`, outside #2114's named files, so they are untouched. Follow-up if wanted.
- No feature flag anywhere — `ENABLE_V3_UI` was retired in #2184.

## Verification

- `tsc --noEmit -p tsconfig.lib.json` — clean.
- `eslint` on every touched path — clean.
- `nx build vault` — passes.
- `vitest run` — full vault suite green: 293 files, 3396 tests, 0 failures.
- Two page-level suites needed the new hooks stubbed: `Activity.test.tsx` and `router.test.tsx` render the Activity page without a `QueryClientProvider`, and `usePrices` reaches the built wallet-connector bundle that vitest cannot transform. Both now stub `usePrices` / `useVaults`, the same way that file already stubs the deposit-lifecycle graph.
- Activity row tests extended in place (`ActivityRowV3.test.tsx`): USD sub-line from amount × price; a stable-side row priced from its own symbol; no sub-line for an unpriced symbol, a missing numeric amount, or a zero amount; deposit with a live vault reads "In use"; deposit whose vault left the set reads "Done"; unresolved vault read and a row with no vault id both hold the type-derived label.

### DIFF
--- a/services/vault/src/__tests__/router.test.tsx
+++ b/services/vault/src/__tests__/router.test.tsx
@@ -187,6 +187,12 @@ vi.mock("@/hooks/usePendingDeposits", () => ({
   }),
 }));
 
+// Same reason: the Activity feed's price source imports the built
+// wallet-connector bundle for its network enum.
+vi.mock("@/hooks/usePrices", () => ({
+  usePrices: () => ({ prices: {} }),
+}));
+
 vi.mock("@/context/ProtocolParamsContext", () => ({
   ProtocolParamsProvider: ({ children }: { children: ReactNode }) => children,
 }));
--- a/services/vault/src/__tests__/router.test.tsx
+++ b/services/vault/src/__tests__/router.test.tsx
@@ -188,9 +188,14 @@ vi.mock("@/hooks/usePendingDeposits", () => ({
 }));
 
 // Same reason: the Activity feed's price source imports the built
-// wallet-connector bundle for its network enum.
+// wallet-connector bundle for its network enum, and its vault read wants a
+// depositor address this suite never connects.
 vi.mock("@/hooks/usePrices", () => ({
   usePrices: () => ({ prices: {} }),
+}));
+
+vi.mock("@/hooks/useVaults", () => ({
+  useVaults: () => ({ data: undefined }),
 }));
 
 vi.mock("@/context/ProtocolParamsContext", () => ({
--- a/services/vault/src/components/Activity/ActivityList.tsx
+++ b/services/vault/src/components/Activity/ActivityList.tsx
@@ -55,6 +55,8 @@ const FILTER_OPTIONS = (
 interface ActivityListProps {
   activities: ActivityRow[];
   isConnected: boolean;
+  /** Current USD price per token symbol, for each row's USD sub-line. */
+  prices?: Record<string, number>;
   /** Vault ids of expired deposits whose HTLC refund is still outstanding —
    *  those rows get the Withdraw action. Matched against a row's `vaultId`,
    *  never its `id` (see ActivityLog). */
@@ -65,6 +67,7 @@ interface ActivityListProps {
 export function ActivityList({
   activities,
   isConnected,
+  prices,
   refundableVaultIds,
   onWithdraw,
 }: ActivityListProps) {
@@ -148,6 +151,7 @@ export function ActivityList({
                       <ListRowCard>
                         <ActivityRowV3
                           row={r}
+                          prices={prices}
                           action={
                             r.vaultId &&
                             refundableVaultIds?.has(r.vaultId) &&
--- a/services/vault/src/components/Activity/ActivityList.tsx
+++ b/services/vault/src/components/Activity/ActivityList.tsx
@@ -57,6 +57,9 @@ interface ActivityListProps {
   isConnected: boolean;
   /** Current USD price per token symbol, for each row's USD sub-line. */
   prices?: Record<string, number>;
+  /** Vault ids ACTIVE on chain, for a deposit row's live status. `undefined`
+   *  while the vault read is in flight. */
+  activeVaultIds?: ReadonlySet<string>;
   /** Vault ids of expired deposits whose HTLC refund is still outstanding —
    *  those rows get the Withdraw action. Matched against a row's `vaultId`,
    *  never its `id` (see ActivityLog). */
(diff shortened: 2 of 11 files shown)


jonybur (approved):
Wiring is right: prices and activeVaultIds are read once at the container and threaded down; getUsdSubLine guards on missing numeric, missing price and non-positive product so the row silently drops rather than shows '$0 USD'. Deposit-status derivation joins on vaultId not id (matching how refundableVaultIds already does it), and undefined activeVaultIds correctly holds the type-derived label while the vaults query is in flight.

PR #2287, opened 2026-08-20
chore(vault): retire the market detail page flag

jeremy-babylonlabs:
Retires `NEXT_PUBLIC_FF_ENABLE_MARKET_DETAIL_PAGE`. The borrowing market detail page (`/markets/:reserveId`) and its entry point become permanent.

## Why now

The feature is complete — the page (#2148, #2158, #2166, #2171), both charts (#2204) and the chart perf pass (#2242) are all merged. The flag's stated reason to exist was that the market data "is not wired to a live source yet"; that no longer holds. The indexer endpoints the page reads are live: devnet probes of `/api/aave/reserves/:id/irm` and `/history` both return real data.

The flag is set in no GitHub environment, so there is nothing to coordinate with DevOps and nothing to clean up afterwards. Removing it changes deployed behaviour by exposing the finished page.

## What changed

- `config/featureFlags.ts` — the `isMarketDetailPageEnabled` getter and its doc block are gone.
- `router.tsx` — the `/markets/:reserveId` route renders the page unconditionally; the redirect-to-overview branch is deleted.
- `AssetSelectionPanel.tsx` — the Market Info button renders whenever the picker is not in repay mode, and `getAssetPickerWidthClass` drops the flag term (repay stays 612px, borrow 700px). The `featureFlags` import and the now-stale comment are removed.
- `.env.example`, `services/vault/README.md` — the documented variable is removed.
- `.github/workflows/service-release-vault.yml` — the release build no longer injects the variable. This reference is outside `services/` and beyond the original file list; it is included because leaving it would inject a build-time variable no code reads.

## Tests

Coverage is kept, the flag axis is dropped:

- `router.test.tsx` — the flag-off redirect test becomes `renders the market detail page at /markets/:reserveId`, asserting the route resolves to the page and not the dashboard. Deleting it outright would have left `/markets` with no route test at all, so it was converted rather than removed. The lazily-imported page is stubbed the same way `Explore` and `Liquidations` already are in that file.
- `AssetSelectionPanel.test.tsx` — the flag-off test is deleted. The `@/config/featureFlags` mock and its `beforeEach` reset go with it, since the component no longer imports the module. The behaviour tests (Market Info routes to the markets page, repay omits it) are untouched.

## Verification

- `grep -rn "isMarketDetailPageEnabled\|ENABLE_MARKET_DETAIL_PAGE"` over the repo: zero hits.
- `tsc --noEmit -p tsconfig.lib.json`: clean.
- ESLint on the touched paths: 0 errors (1 pre-existing `react-refresh` warning on `getAssetPickerWidthClass`, unchanged by this PR).
- `vitest run` on both touched suites: 30/30 pass.

## Scope

`isLiquidationAnalysisChartEnabled`, `isExploreEnabled` and `V3_SECTION_FLAG_GATES` are untouched — market detail was never part of that mechanism. No behaviour change to the page itself.

Closes nothing. #2098 stays open; the C4 discoverability question rides on it.


jonybur (approved):
Flag retirement is complete: the getter, the router branch, the AssetSelectionPanel branch, the workflow injection, the .env.example entry, and the README entry all go, and the flag mocks come out of the tests they were driving. Verified the removals by grep - no residual reference to `isMarketDetailPageEnabled` or `NEXT_PUBLIC_FF_ENABLE_MARKET_DETAIL_PAGE`. The router test is converted rather than deleted (keeps coverage on the `/markets/:reserveId` route), and the panel test now exercises Market Info unconditionally. `getAssetPickerWidthClass` still returns 700px for borrow and 612px for repay - the mode axis is preserved. Description names the production-first exposure explicitly, which is a call the reviewer already made. Nothing else in the diff hides.

gbarkhatov (approved):

PR #2288, opened 2026-08-20
feat(vault): chart the liquidation timeline with live BTC price candles

jeremy-babylonlabs:
## What

Replaces the seizure map with the redesigned **BTC price candlestick chart with liquidation bands** on both surfaces, and closes the one gap the chart component was built with: candles had no data source.

Figma:
- Liquidations page chart — [`10251:32415`](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10251-32415)
- Overview embedded card — [`10251:35805`](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10251-35805)

## Data source

BTC/USD candles come from the vault indexer's GraphQL API — the same `ENV.GRAPHQL_ENDPOINT` and `graphqlClient` the rest of the app uses. Candles are keyed by **price feed, not by reserve** (vaultBTC and WBTC share the BTC/USD feed), so the vaultBTC reserve resolves through `aavePriceFeedSource` first, then the candles are read for that feed.

Scaling is taken from the candle row's own `decimals` field, never a hardcoded exponent. The indexer stores raw feed integers and carries the feed's decimals on both `aavePriceFeedSource` and `aavePriceCandle` (`schemas/applications/aave/price.ts`: *"Feed decimals (8, matching the oracle's own scale)"*). Verified live against the devnet indexer:

```
aavePriceFeedSource(id: "2") -> { feed: 0xa5c0…8db, symbol: "BTC / USD", decimals: 8 }
aavePriceCandles(day_1)      -> { bucketStart: "1787184000", open: "6968071223262", decimals: 8 }
                                 6968071223262 / 10^8 = $69,680.71
```

`bucketStart` is unix seconds; the hook converts to the milliseconds `Candle.time` expects. Interval is fixed to `day_1` (the frames show no timeframe chooser), 180 daily candles fetched, 60 in view, the rest reachable by pan. `staleTime`/`refetchInterval` are 30 min — the feeds publish roughly hourly, so anything faster re-reads the same rows.

## What replaced what

| Surface | Before | After |
|---|---|---|
| Liquidations page | `SeizureMap` + simulator toolbar | `Timeline` (candles, bands, safe zone, crosshair + pan) + the same toolbar |
| Overview card | `SeizureMap` + slider + header figures + event cards | `Timeline` + title/description + Explore |

The bands, trigger prices, tones and safe-zone figures all derive from the live cascade that already fed the seizure map — no new financial model. The price axis is new: round-number ticks down the safe zone, then the axis floor. Ticks deliberately stop at the first trigger, because each trigger already carries its own dashed level line and price pill and the scale compresses the trigger-to-trigger regions.

## Notable decisions and Figma deltas

- **The overview card loses the simulator, the header figures (`x/y vaults liquidated`, seized %, collateral) and the event cards.** Frame `10251:35805` shows heading → title/description → Explore → chart and nothing else. That makes the card a read-only preview and Explore the way into the simulation. `LiquidationEventCards.tsx` had no other consumer and is deleted.
- **Band labels.** The page frame reads `Liq Event 1 (0.6 BTC)`, the overview frame reads `Liq Event 1`. Both are implemented as drawn — the dashboard folds the amount into the label, the preview does not.
- **Mock figures are internally inconsistent and were derived instead of copied.** The safe zone reads "no events above $59,050 / 4.3% drop to Liq 1" while the same frame's first trigger is $67,682; the overview frame labels the first trigger $77,682 where the page frame labels it $67,682. Both lines derive from the cascade: the first trigger price, and the drop from the current price to it.
- **Axis floor.** The page frame's bottom tick is a round `$1,000`; the implementation keeps the existing `axisFloorPrice` (lowest trigger − 5%), which is what the bands and the simulator's lower bound are already built on.
- **Band tones and price rule re-toned** to the frames (`#dc944d` / `#d1792c` / `#cc1818`, rule and price pill `#387085`). These are the shared `--liq-band-*` / `--liq-price-line` tokens, so the liquidation event cards' top rules follow the chart, as they did before.
- **Stale oracle** keeps the existing behaviour: `usePositionNotifications` returns no params on `stale-price`/`no-price`, and the page already renders the "liquidation analysis unavailable" state. No greyed-back chart was added — the frames show none.
- **Crosshair readout.** The abbreviated date axis is a tick format, not a timestamp format, so `Timeline` gained an optional `formatReadoutTime` (defaulting to `formatTime`); the dashboard's hover readout spells the candle's date out while the axis under it stays "5 / 12 / 19 / Jun".
- **Mobile is out of scope** for this PR.

## SeizureMap

`SeizureMap` now has **no application consumer** — only its own story and tests. Left in place deliberately; removing it (and the `priceAxis` / `shareAxisTicks` outputs of `buildLiquidationChartData` that only it consumed) belongs in a follow-up.

## Review

`/code-review high` raised two, both fixed in `83260a98`: a blank price string coerces to a finite `0` and would have plotted a $0 candle past the non-finite guard (now rejected, with tests for a blank price and a blank bucket start), and `COPY.liquidations.events.liquidatedInSimulation` was left dead by the deleted preview grid.

## Testing

- `services/vault`: `tsc --noEmit` clean; `nx build vault` clean; full suite **3408 passed / 6 skipped**; `nx lint vault` 0 errors.
- `@babylonlabs-io/core-ui`: **112 passed**; lint 0 errors.
- New coverage: candle fetch (feed resolution, decimal scaling incl. a non-8-decimal feed, seconds→ms, ordering, empty feed, malformed row throws), the Timeline price axis (round ticks, top clears the candles, no tick inside a compressed event region, already-liquidatable position, no events), the safe-zone derivation, the band-label fold, the date axis, and both surfaces rendering the frame with **no** candle series.


jonybur (dismissed):
Candle series is keyed by price feed rather than reserve, decimals are read from the row rather than assumed, and the seconds-to-ms conversion belongs at fetch time. Blank-price guard added in 83260a98 covers the numeric-coercion trap. Timeline anchor fractions and safe-zone derivation trace cleanly. The described Figma deltas and the deletion of the seizure-map preview are consistent with the frames.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -314,13 +343,18 @@ export default function Liquidations() {
             price={clampToLive(simulatedPrice, chart.livePrice)}
             onPriceChange={setSimulatedPrice}
           />
```
jrwbabylonlab, line 316: Removing this drops `COPY.liquidations.bitcoinPriceCaption`'s last consumer (the other call site goes too) but the key stays in copy.ts — dead copy, worth dropping the key in this PR.

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -132,16 +145,31 @@ export default function Liquidations() {
...
+    const data = buildLiquidationChartData(cascadeResult, {
+      btcPrice: projectedPrice,
+      collateralFactor,
+      livePrice,
+      vaultsTotal: params.vaults.length,
+    });
+
+    // The axis top has to clear the candles as well as the price rule —
+    // anything above the first tick is clipped out of the plot.
+    const topPrice = (candles ?? []).reduce(
```
gbarkhatov, line 160: **1 — The axis top is candle-aware, the axis floor is not: history below the floor flattens against the plot bottom.**

`topPrice` folds every candle high into the domain ceiling, with a comment explaining that anything above the first tick is clipped. Nothing does the mirror for the bottom: the domain floor is `axisFloorPrice(result)` = `min(trigger) * 0.95`, derived from the cascade alone and blind to the series. Both scales in `priceScale.ts` are built with `clamp: true`, and `layoutCandles` maps `high`/`low`/`open`/`close` straight through, so every candle that traded under the floor gets pinned to `plotHeight`.

Checked against the overview fixture in this PR — one trigger at $73,333, floor $69,666.35, 300px plot:

```
$75,000.00 ->   0.00 px
$73,333.00 -> 256.00 px
$70,000.00 -> 296.00 px
$69,666.35 -> 300.00 px
$68,000.00 -> 300.00 px
$60,000.00 -> 300.00 px
$45,000.00 -> 300.00 px
```

$68k, $60k and $45k are indistinguishable. Over a 180-day window on a position opened after a run-up, that is a run of identical flat-bottomed bars that reads as "price sat at the floor" when it was actually far below. Neither new fixture has a candle under the floor, so nothing catches it.

There is a real design question in the fix — extending the domain bottom to the lowest candle also has to move the last band's `priceBottom`, or the bottom band stops short of the axis — so worth deciding deliberately rather than patching the reduce. Either fold candle lows into the floor (and the band bottom with it), or drop out-of-domain candles instead of clamping them, so a flat bar always means a flat price.

Same expression at `LiquidationAnalysisSection.tsx:106`.

services/vault/src/applications/aave/hooks/useBtcPriceCandles.ts
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Days of history requested. Wider than the visible window so the Timeline's
+ * pan has somewhere to go.
+ */
+const CANDLE_LIMIT = 180;
+
+export interface UseBtcPriceCandlesResult {
+  candles: Candle[] | null;
+  isLoading: boolean;
+  error: Error | null;
```
gbarkhatov, line 36: **2 — `isLoading` and `error` are exposed and read by nobody, so a failed fetch renders as an empty chart.**

Both call sites destructure only `{ candles }` (`Liquidations/index.tsx:99`, `LiquidationAnalysisSection.tsx:96`). Under the zero-dead-code policy these two fields are dead API surface today.

The bigger effect is that it makes the guard on line 65 self-defeating: `error ? null : (data ?? null)` deliberately discards last-good candles "so a failed background refetch never leaves stale candles beside an error" — but no consumer renders an error, so the trade is good data for a blank price panel and no signal. A 500 from the indexer looks exactly like a feed with no history.

There is already a convention for this in the app on the same kind of series — `BorrowRateHistoryCard` renders `COPY.common.loading`, then `chartUnavailable` on `error || points === null`, then `historyEmpty` for a zero-length series. Either wire the same three states through the Timeline surfaces, or drop the two fields and change the module doc, which currently promises that a missing series "degrades to an empty price panel rather than an error state" — that is a decision worth stating on purpose, not by omission.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -91,45 +87,32 @@ function EmptyState({
...
-  // cards' liquidated state.
-  const liquidatedKeys = useMemo(
-    () =>
-      new Set(bands.filter((b) => b.state === "liquidated").map((b) => b.key)),
-    [bands],
-  );
+  const { candles } = useBtcPriceCandles();
+
+  const chart = useMemo(() => {
+    const { bands } = buildLiquidationChartData(cascade.result, {
```
gbarkhatov, line 99: **3 — Retiring `SeizureMap` from the app orphaned four values in `liquidationChartData.ts`.**

This call now takes only `{ bands }`, and the /liquidations page uses `chart.timelinePriceAxis`, not `data.priceAxis`. On `main` all four had real consumers; after this PR their only remaining references are the producers themselves and the tests:

- `LiquidationChartData.priceAxis` (`liquidationChartData.ts:84`) — was `SeizureMap`'s segmented axis; now built on every render for nobody.
- `LiquidationChartData.shareAxisTicks` (`:85`) — was `shareAxisTicks={chart.shareAxisTicks}` on `main:index.tsx:323`.
- `LiquidationChartOptions.livePrice` (`:101`) — still passed from `index.tsx:154`, but its only effect is `axisAnchor` at `:483`, which feeds the dead `priceAxis`. Live, but transitively dead.
- `LiquidationEventCard.distanceNegative` (`:68`) — its sole consumer was `LiquidationEventCards.tsx:89`, deleted in this PR. `LiquidationEventCard.tsx:174` prints `distanceLabel` and tints by `badge`.

`SeizureMap` itself is fine to keep — core-ui is a published library and it stays exported. These four are app code with no caller.

Related, same call: `buildLiquidationChartData` also runs `toCard()` for every group and `buildSimulationSummary` here, and the overview discards both.

services/vault/src/components/simple/LiquidationAnalysisSection.tsx
```diff
@@ -91,45 +87,32 @@ function EmptyState({
...
+      btcPrice: cascade.params.btcPrice,
+      collateralFactor: cascade.params.CF,
+      vaultsTotal: cascade.params.vaults.length,
+    });
+    // The axis top has to clear the candles as well as the price rule —
+    // anything above the first tick is clipped out of the plot.
+    const topPrice = (candles ?? []).reduce(
```
gbarkhatov, line 106: **4 — The preview's price domain is sized from 180 candles when only 60 are reachable.**

`useBtcPriceCandles` requests `CANDLE_LIMIT = 180` because "the Timeline's pan has somewhere to go", but this surface passes `visibleCandles={TIMELINE_VISIBLE_CANDLES}` (60) and no `interactions`, so pan and zoom are both off — the older 120 candles cannot be reached by any gesture. They still set the ceiling through this reduce, so a spike four months ago squashes the 60 candles the user can actually see.

The dashboard page is correct as written: `interactions={{ crosshair: true, pan: true }}` makes the whole 180 reachable, so its domain should cover them.

`candles.slice(-TIMELINE_VISIBLE_CANDLES)` for the reduce here, or share one derivation that takes the reachable window.

services/vault/src/applications/aave/services/fetchPriceCandles.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (!aavePriceFeedSource) return [];
+
+  const { aavePriceCandles } = await graphqlClient.request<CandlesResponse>({
+    document: GET_PRICE_CANDLES,
+    variables: { feed: aavePriceFeedSource.feed, interval, limit },
+    signal,
+  });
+
+  // The indexer serves newest first so `limit` takes the most recent window.
+  return aavePriceCandles.items.map(toCandle).reverse();
```
gbarkhatov, line 148: **5 — Ordering is a blind `.reverse()` of the server's response, and the test that claims otherwise only exercises one direction.**

The comment says "the indexer serves newest first", which is true because line 62 asks for `orderDirection: "desc"`. But the transform is order-blind: it reverses whatever arrives. Change that literal to `"asc"`, or have the indexer's default shift, and the chart plots time backwards — the candles are laid out by array index, not by `time`, so nothing downstream would notice.

`fetchPriceCandles.test.ts:91` is titled "returns the series oldest first, **whichever order the indexer served**" but feeds only descending rows, so it asserts the `.reverse()` rather than the claim. Either sort by `bucketStart` ascending here and make the title true, or rename the test to say it asserts the reversal of a descending response.

Cheap and total: `.sort((a, b) => a.time - b.time)` after the map makes the function correct for any server ordering and lets the test title stand.

services/vault/src/applications/aave/services/fetchPriceCandles.ts
```diff
@@ -0,0 +1,149 @@
...
+  const value = raw.trim() === "" ? Number.NaN : Number(raw) / scale;
+  if (!Number.isFinite(value)) {
+    throw new Error(
+      `Price candle has a non-finite "${field}": ${JSON.stringify(raw)}`,
+    );
+  }
+  return value;
+}
+
+function toCandle(item: CandleItem): Candle {
```
gbarkhatov, line 94: **6 — The parser trusts its declared TypeScript shape instead of narrowing from `unknown`, unlike the sibling indexer client.**

`CandleItem` is a compile-time assertion about an external payload, not a runtime one. If `open` comes back `null` or as a JSON number, `raw.trim()` throws `TypeError: raw.trim is not a function` — a throw, so nothing is silently plotted, but not the actionable error the module docstring at `:77-81` promises.

`clients/indexer/aaveHistoryClient.ts:30-83` handles the same class of input the other way and documents the same rationale: narrow from `unknown`, check each field's type, and name the endpoint and the row index in the message.

Two smaller gaps in the same function:

- No error message names the feed, the interval, or the bucket, so a bad row on a 180-row response gives nothing to look up.
- No `high >= low` (or `low <= open/close <= high`) coherence check. A transposed pair renders an inverted wick with every field individually finite.

Given this is external data feeding a chart a depositor reads prices off, matching the sibling client's pattern seems worth it.

packages/babylon-core-ui/src/components/LiquidationChart/priceScale.ts
```diff
@@ -48,79 +48,47 @@ export interface PriceAnchor {
...
+ * would make a $180 event an unreadable sliver next to a $37k one — pinned
+ * at `minFraction`. The safe zone gets whatever height remains, down to
+ * {@link MIN_SAFE_ZONE_FRACTION}: past that (a long cascade — the protocol
+ * allows up to 10 vaults), every event's row compresses evenly below the
+ * floor instead of being clipped individually or crushing the safe zone to
+ * nothing. A compressed row's in-band text degrades on its own via
+ * BandLayer's existing height-based dropout. Returns one fraction per span,
+ * summing to 1.
  */
 export function timelineRegionFractions(spans: number[], minFraction: number): number[] {
```
gbarkhatov, line 72: **7 — `timelineRegionFractions` never reads `spans`, only `spans.length`.**

Every branch uses `spans.length`; no element value is ever touched. That is the intended behaviour — the doc above explains that event rows are deliberately fixed-height and never scaled by their price span — but the signature still asks for the array, and `Timeline.tsx:134-137` builds it (`stops.slice(1).map((price, i) => stops[i] - price)`) for the sole purpose of having it discarded.

It also makes a test vacuous: `priceScale.test.ts:92` — "gives every event its floor fraction **regardless of price span**" — passes `[12_318, 37_399, 283]`, which the function cannot read. It would pass identically for `[0, 0, 0]`, so it pins nothing.

Take `regionCount: number` instead and drop the caller's map. The behaviour the test is reaching for then becomes true by construction rather than by assertion.

services/vault/src/applications/aave/services/fetchPriceCandles.ts
```diff
@@ -0,0 +1,149 @@
+/**
```
gbarkhatov, line 1: **8 — This module is the data source that four core-ui comments still say does not exist.**

All four are now false and none is touched by the diff, including one whose neighbouring JSDoc this PR does edit:

- `LiquidationChart/types.ts:8-10` — "`candles` is the only field with no data source today (issue #2043 price-history follow-up)"
- `LiquidationChart/types.ts:64` — "One OHLC bar. Timeline only; no data source exists yet."
- `LiquidationChart/types.ts:168` — "Empty until the price-history ticket lands." (on `candles?: Candle[]`)
- `LiquidationChart/Timeline.stories.tsx:185` — "No price-history feed yet: frame + band gutter render without candles."

The stories one wants a rewording rather than a deletion — the `NoCandles` story is still a valid state (a feed with no history, or a failed fetch), it just is not "no feed yet" any more.

Checked and *not* stale: `charts/types.ts:9` still correctly documents `"both"` as `ChartFrame`'s default; Timeline overriding it to `"vertical"` does not change the shared contract.

packages/babylon-core-ui/src/components/LiquidationChart/Timeline.tsx
```diff
@@ -114,10 +119,10 @@ export function Timeline({
...
-  // takes a vertical share weighted by its compressed price span — the split
-  // tracks the real values while small events stay readable, and the safe
-  // zone never drops below the largest event so the candles keep room.
+  // event region (trigger-to-trigger, then last trigger-to-floor) gets a
+  // fixed, compact row (MIN_REGION_PX) regardless of its real price span —
+  // matching the design, which draws every "Liq Event N" band the same
+  // size — so the safe zone (and its candles) keeps whatever height remains.
   // Without events (or with every trigger off-domain) it falls back to
   // linear.
   const priceScale = useMemo(() => {
```
gbarkhatov, line 128: **9 — Anchors are derived from band `priceTop` without the dedup the app applies to the other axis, and a duplicate throws rather than degrading.**

`bands.map((b) => b.priceTop).filter(...)` on line 129 feeds `stops`, which feeds `createAnchoredPriceScale` → `assertStrictlyDescending`, which throws. Two bands sharing a `priceTop` therefore crash the render, uncaught, instead of drawing one merged region.

The app already treats coinciding prices as reachable: `liquidationChartData.ts:486-492` builds the segmented axis through a `Set` with the comment "dedup avoids duplicate tick values (React keys) when prices coincide". If prices can coincide there, they can coincide here — the same `group.liquidationPrice` values are the input to both.

Low reachability (successive groups usually have strictly decreasing triggers), so this is a hardening ask, not a bug report. But `createAnchoredPriceScale` has no direct unit test of its own, and this is the one caller that constructs its input from unsanitised data — a `dedupe + sort` on `triggers` costs one line and turns a crash into a merged band.

services/vault/src/applications/aave/hooks/useBtcPriceCandles.ts
```diff
@@ -0,0 +1,69 @@
...
+      reserveId === null
+        ? skipToken
+        : ({ signal }) =>
+            fetchPriceCandles({
+              reserveId,
+              interval: CANDLE_INTERVAL,
+              limit: CANDLE_LIMIT,
+              signal,
+            }),
+    staleTime: CANDLE_STALE_TIME_MS,
```
gbarkhatov, line 57: **10 — `staleTime` of 30 minutes against the app's default `gcTime` of 5 minutes.**

`config/queryClient.ts:163` sets `gcTime: 5 * 60 * 1000` globally. With no `gcTime` here, an unmounted query is garbage-collected after 5 minutes, so navigating away from /liquidations and back six minutes later refetches all 180 rows — the 30-minute freshness window never gets used across a remount.

`useInterestRateModelCurve.ts:187` hits the identical situation and sets `gcTime: CURVE_CACHE_MS`, documenting exactly this: "`gcTime` matches `staleTime`: with the default 5-minute gc a route remount would refetch anyway."

`gcTime: CANDLE_STALE_TIME_MS` alongside line 57.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -166,6 +167,142 @@ export function axisFloorPrice(result: CalculatorResult): number {
...
+
+  if (!Number.isFinite(topPrice) || topPrice <= floorPrice) {
+    return [toTick(Math.max(topPrice, floorPrice)), toFloorTick(floorPrice)];
+  }
+
+  // Below the first trigger the scale is compressed per event, so the ticks
+  // stop there and the (unlabelled) floor tick closes the axis.
+  const tickFloor = topPrice > firstTrigger ? firstTrigger : floorPrice;
+  const span = topPrice - tickFloor;
+  if (span <= 0) return [toTick(topPrice), toFloorTick(floorPrice)];
```
gbarkhatov, line 228: **11 — Minor, grouped (no action needed on any of these individually).**

- **This line is unreachable.** Line 220 has already returned unless `topPrice > floorPrice`. If `topPrice > firstTrigger` then `tickFloor = firstTrigger` and `span = topPrice - firstTrigger > 0`; otherwise `tickFloor = floorPrice` and `span = topPrice - floorPrice > 0` by that same guard. `span <= 0` cannot hold. Dead branch under the zero-dead-code policy.
- **`TIMELINE_VISIBLE_CANDLES = 60` is defined twice** — `Liquidations/index.tsx:60` and `LiquidationAnalysisSection.tsx:49`, with different doc comments. Two surfaces sharing one design value; worth one exported constant.
- **Copy inconsistency:** `copy.ts:1631` renders "…% drop to **Liq 1**" while `eventTitle` at `:1646` names the band directly beneath it "**Liq Event 1**". Same event, two names, stacked vertically.
- **`fetchPriceCandles` is not exported from `applications/aave/services/index.ts`**, unlike every other service in that directory; the hook deep-imports the module.
- **`useBtcPriceCandles` is the only hook in its directory with no test.** Untested and cheap to pin: `skipToken` leaves `isLoading` false rather than a permanent spinner, `error` maps to `candles: null`, and the query key varies per reserve.

gbarkhatov (approved):
Approving — nothing here blocks the merge. Read every hunk, plus the surrounding files on both sides of the core-ui/vault seam, and re-derived the price scale against this PR's own fixtures.

Deliberately skipping what the bots already have: the degenerate `topPrice <= floorPrice` axis, `10 ** decimals` overflowing to `Infinity`, `bucketStart: "1e308"`, band-label contrast, `amountLabel` vs `ShareLegend`, triggers missing from the axis ceiling, the "Simulate BTC price" copy on the read-only preview, magic numbers, test fixture constants, and `bitcoinPriceCaption`.

Eleven findings inline. The one I would actually fix before merge is **1** — the axis is candle-aware at the top and blind at the bottom, so any history below `min(trigger) * 0.95` clamps to a flat run of identical bars. Verified numerically against the overview fixture: $68k, $60k and $45k all land on the same pixel. **2** and **3** are policy-level rather than behavioural (dead API surface, four values orphaned by the `SeizureMap` retirement), but **2** has a user-visible edge: a failed indexer fetch is currently indistinguishable from a feed with no history.

Two notes that have no line inside the diff to hang on:

**A pan drag can stick when candles vanish mid-gesture.** `Timeline.tsx:486-489` wires `onPointerUp`/`onPointerCancel` only when `panEnabled`, which now depends on async, fallible data (`candles.length > windowSize`). A failed refetch mid-drag drops the handler, so `endDrag` never runs: `drag.current` and the `dragging` state stay set and the pointer capture is never released. When candles return, the next pointer move pans with no button held. Newly reachable because this PR made `candles` a network result — worth an unconditional `endDrag`.

**The GraphQL contract has no test that can fail.** Both suites mock `graphqlClient` wholesale, so the enum rename in `9c0bb2d2` (`aave_price_interval!` → `aavePriceInterval!`) is asserted nowhere — a wrong type name surfaces at runtime as a validation error that the hook turns into an empty chart, silently (see finding 2). The live-verification block in the PR description predates that commit and still shows the old query. Since no codegen artifact exists in this repo, re-running that check against the indexer and pasting the result is the only verification available — worth doing before merge.

Checked and clean: the `"vertical"` grid-lines default (`LineChart`, `ChartFrame` and `SeizureMap` all still resolve to `"both"`, so no silent regression); `timelineRegionFractions` cannot emit a zero or negative fraction at any span count, so `createAnchoredPriceScale`'s ascending assertion cannot fire from it; `buildTimelinePriceAxis` is strictly descending and duplicate-free in every branch except the degenerate one the bots already flagged; UTC is used consistently across `formatCandleDate` and `formatCandleTimestamp`; no orphaned references to any symbol removed with `LiquidationEventCards.tsx`; and no `data-testid` consumed by the real-wallet E2E CLI was dropped.

Verification: `tsc --noEmit` clean on core-ui; the `LiquidationChart`, `Liquidations` and `fetchPriceCandles` suites pass locally (one unrelated vault test file fails to resolve the optional raw vault-WASM peer entry). I did not re-run the full suite, so the description's "3408 passed / 112 passed / 0 lint errors" is unverified here.

PR #2289, opened 2026-08-20
feat(vault): restyle the activity page to the updated design

jeremy-babylonlabs:
Screenshot 2026-08-21 at 20 18 36

Rebuilds the Activity page presentation to match the updated design frame: [Activity 13164:77644](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=13164-77644).

Stacked on `feat/activity-usd-and-live-status` (#2114 + #2116). It reuses that branch's `ActivityAmount.numeric` plumbing and its `usePrices` / `useVaults` wiring rather than adding its own.

## What changed

| Aspect | Before | After |
|---|---|---|
| Toolbar | "Show all" dropdown at the left | Search box at the left ("Search tx hash or type of transaction"), "Show All" dropdown pinned right |
| Date groups | Today / Yesterday / `YYYY-MM-DD` | Today / Yesterday / **Last week** (2–7 days back) / `YYYY-MM-DD` beyond that |
| Row anatomy | icon + amount, status dot + label, type, hash, time | icon, **type as the primary label with the asset symbol beneath it**, labelled explorer link, labelled timestamp, **right-aligned amount with its USD sub-line** |
| Status | Dot + "Done" / "In use" / "Expired" / "Pending" column | **Column removed.** Pending spins a 16px loader beside the type label; a refunded deposit gets a "Refund" chip and its card loses its fill while the row fades to 60% |
| Liquidation card | Child rows carried a repeated status dot | Same card, children now read "Liquidated" / "Debt repaid" and each carries its own USD sub-line |

Search filters the rows already loaded — transaction hash (with or without `0x`) and the visible type text, debounced at 200ms. No new indexer query. The type filter keeps its behaviour; only its styling and position changed.

## Decisions worth a look

**The explorer label is derived, not hardcoded.** The frame draws "Xangle Explorer :", but this cell links through `getExplorerTxUrl`, which points at mempool.space for BTC rows and the configured chain explorer for ETH rows — never the VP explorer that `NEXT_PUBLIC_TBV_VP_EXPLORER_URL` configures. The label is derived from the host actually linked ("Mempool", "Etherscan"), so it follows the environment instead of claiming an explorer the link never opens. Flagging for design: if the intent was to link the Babylon vault explorer instead, that is a different link, not a different label.

**"In use" has no surface in this design, and that is a question for design.** #2116 makes a deposit row's status reflect live vault state; this frame removes the status column entirely, so nothing renders that distinction any more. Per the brief the derivation stays wired — `useVaults` → `activeVaultIds` still flows from the container through the list into the row's props — but no component reads it, and the label rendering plus its tests are removed here (there is no longer a label for those tests to assert). **If depositors should still be able to tell a deposit whose collateral is in use from one that has been withdrawn or redeemed, the design needs a surface for it.** Re-rendering it is a small change once that surface exists.

**Row dimming is one opacity, not a recoloured palette.** The frame recolours every text run in a dimmed row to `text/tertiary`. This applies `opacity-60` to the row and drops the card fill instead — one class, and it holds in dark mode, where a hardcoded `#b0b0b0` would not.

## What the frame draws that the data cannot back yet

Listed as follow-ups, not implemented here:

- **"Failed" chip + dimmed row.** `ActivityLog` has no failure state — the indexer projection cannot emit a failed row today. The chip treatment is built (the refunded-deposit path uses it) and would only need the projection to mark a row as failed.
- **"Fairness Payment" rows.** No such event reaches the activity projection. The frame draws it as a third liquidation child.
- **"Debt repaid" as a standalone row.** It exists only as a liquidation child, which is how it renders here.

Also worth noting: the refunded ("Refund" chip) row still carries its Withdraw action when the HTLC refund is outstanding. The frame draws no action on that row, but dropping the button would remove the only way to reclaim those funds.

## Testing

- Extended the existing Activity suites in place: group labels including "Last week" and the future-date guard, the new row anatomy, search filtering (hash, type text, empty result, debounce), the refund chip and row dimming, the pending spinner, and the liquidation card's per-child pricing.
- New: `activitySearch.test.ts` for the matcher, `LiquidationGroupCardV3.test.tsx` for the group card.
- The base branch's In use / Done row tests are removed with the label they asserted; its USD tests are kept as-is.
- `tsc --noEmit`, ESLint on every touched path, and the full vault suite: **3412 passed, 6 skipped, 0 failures**.

Verified against the live frame. No dark twin was located next to it; the implementation uses the app's semantic colour tokens throughout, which theme automatically.


jeremy-babylonlabs:
@coderabbitai review

jonybur (dismissed):
Row anatomy and search wiring hold up: search is debounced at the list, memoized, and correctly hides on disconnect; usdSubLine is a single source shared by standalone and liquidation rows; ListRowCard now uses twMerge so caller overrides (bg-transparent on refund) actually win. Copy is properly routed through COPY. Deliberate design tradeoffs (In use surface removal, Refund-chip semantics on still-actionable expired rows) are called out in the description.

services/vault/src/components/Activity/ActivityRowV3.tsx
```diff
@@ -48,96 +50,36 @@ interface ActivityRowV3Props {
...
-    if (collateralInUse) {
-      return { dotClass: STATUS_DOT.settled, label: COPY.activity.statusInUse };
-    }
-  }
-  return { dotClass: STATUS_DOT.settled, label: COPY.activity.statusDone };
-}
+export function ActivityRowV3({ row, prices, action }: ActivityRowV3Props) {
+  // A deposit that expired before activation and was reclaimed: the design
+  // marks it with the Refund chip and dims the whole row.
+  const isRefunded = row.isExpired === true;
```
jrwbabylonlab: `isRefunded = row.isExpired` conflates a completed refund with an expired deposit still awaiting withdrawal — the row shows a Refund chip and dims the Withdraw button that's the only path to reclaim the funds. Needs a distinct refund-available vs refunded state.

services/vault/src/components/Activity/ActivityList.tsx
```diff
@@ -81,21 +90,37 @@ export function ActivityList({
...
   // headers stay correct on a page left open overnight (the feed doesn't poll).
   useMidnightTick();
 
-  // The filter control is hidden when the wallet disconnects, so leaving the
-  // selection in place would trap the user on an empty-but-filtered list with
-  // no visible way to clear it. Reset on disconnect.
+  // The toolbar is hidden when the wallet disconnects, so leaving a selection
+  // in place would trap the user on an empty-but-filtered list with no visible
+  // way to clear it. Reset both controls on disconnect.
   useEffect(() => {
```
jrwbabylonlab, line 100: A pending debounced applySearch survives the disconnect reset, so the list comes back silently filtered while the box shows empty. Route the reset through the debounced setter or cancel the timer.

services/vault/src/components/Activity/ActivityRowV3.tsx
```diff
@@ -48,96 +50,36 @@ interface ActivityRowV3Props {
...
   /**
-   * Vault ids that are ACTIVE on chain right now. `undefined` means the vault
-   * read has not resolved yet, and a deposit keeps its type-derived status
-   * rather than flashing the wrong one.
+   * Vault ids that are ACTIVE on chain right now, from the deposit-status
+   * derivation this branch is stacked on. The redesigned row has no surface
+   * for a collateral-in-use label, so nothing reads it yet; the derivation
+   * stays wired pending design's answer (see the PR body).
    */
   activeVaultIds?: ReadonlySet<string>;
```
jrwbabylonlab: `activeVaultIds` is dead plumbing — declared here but never read, and still threaded from ActivityListWithRefund's useMemo. Worth dropping the whole chain.

services/vault/src/components/Activity/ActivityRowV3.tsx
```diff
@@ -146,82 +81,129 @@ export function ActivityRowV3({
...
+  dimmed,
   action,
 }: ActivityRowLayoutProps) {
   return (
-    <div className="flex w-full flex-wrap items-center gap-4">
-      <div className={`${COLUMN_CLASS} gap-2`}>
+    <div
+      className={twJoin(
+        "flex w-full flex-wrap items-center gap-x-10 gap-y-3",
+        dimmed && "opacity-60",
```
gbarkhatov, line 120: **`opacity-60` on the row container drops every text run below WCAG AA, in both themes.**

The class sits on the outermost `div`, so it composites the whole subtree against the page background — and the card fill is simultaneously dropped to `bg-transparent`, so there is nothing behind the text but `surface`. Measured against the real tokens:

| element | light | dark |
|---|---|---|
| `accent-primary` type label / hash / time | `#12495E` @60% on `#fff` → **3.34:1** | — |
| `accent-secondary` sub-lines (12px) | `#666666` @60% on `#fff` → **2.52:1** | — |
| Refund chip (`error-dark` on `secondary-highlight`) | **2.77:1** | **1.68:1** |

Normal text needs 4.5:1. A refunded deposit renders "Deposit", "BTC", the explorer label, the hash, "Time:" and the amount at 2.5–3.3:1.

The PR description gives the rationale as *"one class, and it holds in dark mode, where a hardcoded `#b0b0b0` would not"* — but `#b0b0b0` **is** `accent-secondary`'s dark value. The frame's intent (recolour text runs to `text/tertiary`) is expressible in theme-aware tokens today: `accent-secondary`, `accent-disabled` and `--risk-muted` all already flip per theme, and unlike a blanket opacity they can be tuned per element — so the chip can stay legible while the body text recedes.

Related: because the dim is on the container, an **actionable** Withdraw button inside it renders at 60% too (see line 202) — and `ExpiredWithdrawButton` has a genuinely disabled variant, so the enabled and disabled states become hard to tell apart on exactly the row that has both.

services/vault/src/components/Activity/ActivityRowV3.tsx
```diff
@@ -146,82 +81,129 @@ export function ActivityRowV3({
...
+                role="status"
+                aria-label={COPY.activity.pendingLabel}
+                className="flex shrink-0 items-center"
+              >
+                <Loader size={SPINNER_SIZE} className="text-accent-secondary" />
+              </span>
+            )}
+            {chip && (
+              <span
+                className={`shrink-0 rounded-full bg-secondary-highlight px-3 py-0.5 ${CAPTION_TEXT_CLASS} text-error-dark`}
```
gbarkhatov, line 146: **The Refund chip fails AA in dark mode before any dimming is applied.**

`error-dark` is `#C62828` in *both* palettes (core-ui `tailwind.config.js` light `error.dark` and dark `error.dark` are the same hex), while `secondary-highlight` flips to `#252525`. Dark red on near-black is **2.73:1** undimmed, ~1.7:1 once the row's `opacity-60` lands.

Reading the chip is its entire job, so this is the worst-affected element on the row. Worth picking a chip foreground that is defined per theme (`error-light` `#EF5350` on `#252525` clears AA), independently of whatever happens to the dimming above.

services/vault/src/components/Activity/ActivityList.tsx
```diff
@@ -116,47 +137,55 @@ export function ActivityList({
             </Heading>
           )}
           {isConnected && (
-            <div className="flex items-center gap-4">
+            // Search at the left of the toolbar, the type filter pinned right.
+            <div className="flex flex-1 items-center justify-between gap-4">
```
gbarkhatov, line 141: **The mobile toolbar overflows horizontally.**

The outer row (line 124) is `flex items-center gap-4` with no `flex-wrap`, and on mobile `hideHeading` is `false`, so it carries heading + this `flex-1` group. Neither `ActivitySearchInput`'s `<label>` nor its `<input>` sets `min-w-0`, and an `<input>`'s `min-width: auto` resolves to its intrinsic size (~173px for the default `size=20`) — the classic reason inputs refuse to shrink inside flex containers.

Floors, at a 390px viewport:

- heading "Activity" (single unbreakable word) ≈ 90px
- search label: 2px border + 28px padding + 18px icon + 8px gap + ~173px input ≈ **229px**
- filter trigger ≈ 96px min-content
- 2 × `gap-4` = 32px

≈ **447px** of floor in ~358px of content width. `max-w-[360px]` caps the input's growth but never its floor.

Fix is `min-w-0` on both the label and the input, and/or letting the toolbar wrap below the heading on mobile.

services/vault/src/components/Activity/ActivitySearchInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+        aria-hidden
+        className="shrink-0 text-accent-secondary"
+      />
+      <input
+        type="search"
+        value={value}
+        aria-label={COPY.activity.searchLabel}
+        placeholder={COPY.activity.searchPlaceholder}
+        onChange={(event) => onChange(event.target.value)}
+        className="w-full bg-transparent text-sm leading-[1.43] tracking-[0.17px] text-accent-primary outline-none placeholder:text-accent-disabled"
```
gbarkhatov, line 36: **The placeholder is the search box's only visible description, at 1.96:1.**

`placeholder:text-accent-disabled` is `#9ab7c2` on the label's `bg-primary-contrast` `#F5F7F2` — **1.96:1**. There is no visible `<label>`; `searchLabel` is `aria-label`-only, so a sighted user has nothing else telling them what the box does, and "Search tx hash or type of transaction" is the longest string in the toolbar.

`text-accent-secondary` — which the adjacent `RiSearchLine` at line 28 already uses — measures **5.32:1** on the same fill and would match the icon.

services/vault/src/components/Activity/LiquidationGroupCardV3.tsx
```diff
@@ -40,19 +35,11 @@ export function LiquidationGroupCardV3({ row }: LiquidationGroupCardV3Props) {
           <ActivityRowLayout
             icon={child.tokenIcon}
             iconAlt={child.amount.symbol}
-            amount={`${child.amount.value} ${child.amount.symbol}`}
-            status={status}
             typeLabel={child.label}
```
gbarkhatov, line 38: **The group's Partially/Fully classification is now rendered nowhere, but the filter and the search still key on it.**

`row.type` was previously rendered via `COPY.activity.typeLabels[row.type]` in the status column; dropping that column removed its only consumer, and `__tests__/LiquidationGroupCardV3.test.tsx:68` codifies the absence. Meanwhile:

- `ActivityList.tsx:59` still builds `FILTER_OPTIONS` from `COPY.activity.filterTypes`, which includes "Partially Liquidated" and "Fully Liquidated"
- `activitySearch.ts:27` still matches `COPY.activity.typeLabels[row.type]`

So selecting the "Fully Liquidated" filter, or typing "partially", returns cards whose visible text is only "Liquidated" / "Debt repaid". There is no way to confirm the filter did what it says, and the search matches text that is not on screen. Either the card needs a surface for the classification or the two filter options should go.

Adjacent (pre-existing, not introduced here, but this PR edits both the type and `buildLiquidationGroup`): `LiquidationGroupRow.summary` and `tokenIcons` (`types/activityLog.ts:118-123`) are produced by `classification.ts:150-153` and read by nothing outside test fixtures. Natural point to delete them under the zero-dead-code policy.

services/vault/src/components/Activity/ActivityListWithRefund.tsx
```diff
@@ -75,13 +52,11 @@ export function ActivityListWithRefund({
...
   );
 
-  // Such a row arrives from localStorage still marked pending; the contract
-  // status says otherwise, so correct it before it reaches the status column.
   const rows = useMemo<ActivityRow[]>(
     () =>
       activities.map((row) =>
         row.kind === "row" && row.vaultId && refundableVaultIds.has(row.vaultId)
-          ? { ...row, isPending: false, isExpired: true }
+          ? { ...row, isPending: false }
```
gbarkhatov, line 59: **A deposit that expired before activation and has *not* been reclaimed yet now carries no status indication at all.**

This used to set `{ isPending: false, isExpired: true }`, which drove the "Expired" status label. It now only clears `isPending`, and `isRefunded` is set exclusively by `projectRefundedDeposit` — i.e. rows that are *already reclaimed*.

Result: a deposit still awaiting its refund renders identically to a settled one — no spinner, no chip, no dimming — and the only cue that anything is wrong is an unexplained Withdraw button appearing on it. The chip treatment already exists in this PR and would carry it directly.

services/vault/src/utils/formatting.ts
```diff
@@ -328,6 +333,7 @@ export function formatActivityDateGroup(
 
   if (dayDiff === 0) return labels.today;
   if (dayDiff === 1) return labels.yesterday;
+  if (dayDiff > 1 && dayDiff <= DAYS_IN_LAST_WEEK_GROUP) return labels.lastWeek;
```
gbarkhatov, line 336: **The "Last week" bucket erases the calendar day for everything 2–7 days old.**

`formatActivityTime` (line 303) renders time only, and its own docblock justifies that with *"the calendar day lives in the group header"*. That held while the header for those rows was `YYYY-MM-DD`. For `dayDiff` 2–7 the header is now the constant "Last week", so the day is rendered nowhere.

Reference Sep 8: deposits on Sep 2, Sep 4 and Sep 6 all land in one "Last week" group showing `12:00:00`, `12:00:00`, `12:00:00`. The user cannot tell which day any of them happened — information the pre-PR headers carried. A per-day header inside the group, or a date on the row for that bucket, would restore it.

The boundary arithmetic itself is correct and I verified it: `startOfDay` normalises to local midnight, `Math.round` absorbs 23h/25h DST days (7 days across spring-forward = 167h → 6.958 → 7 → "Last week"; 8 days = 191h → 7.958 → 8 → date), month/year boundaries compare absolute local-midnight timestamps, and the `dayDiff > 1` guard genuinely keeps future rows out. No off-by-one at any of the four boundaries.

services/vault/src/components/Activity/ActivityRowV3.tsx
```diff
@@ -146,82 +81,129 @@ export function ActivityRowV3({
...
+          <span className={`${CAPTION_TEXT_CLASS} text-accent-secondary`}>
+            {usdValue}
+          </span>
+        )}
       </div>
+
+      {/* Only a refundable expired deposit carries an action; the slot is
+          dropped otherwise so the amount stays flush with the card edge, as
+          the design draws it. */}
+      {action && (
```
gbarkhatov, line 202: **`{action && …}` tests the element, not its output.**

`action` is a `<ExpiredWithdrawButton>` React element, which is always truthy — but the component returns `null` when the refund is neither available nor blocked-with-a-tooltip (`ExpiredWithdrawButton.tsx:42`). In that case the fixed `basis-[168px]` slot still mounts, empty.

That row then keeps 168px of gutter while its neighbours (no `action` at all, so no slot) run their amount flush to the card edge — the exact column drift the old always-reserved slot existed to prevent, now inverted onto one row. Gating on rendered output rather than on the element would avoid it.

services/vault/src/components/Activity/ActivityList.tsx
```diff
@@ -116,47 +137,55 @@ export function ActivityList({
...
           )}
         </div>
       )}
 
       {visible.length === 0 ? (
-        <ActivityEmptyState
-          isConnected={isConnected}
-          isFiltered={filter !== null}
-        />
+        <ActivityEmptyState isConnected={isConnected} isFiltered={isNarrowed} />
```
gbarkhatov, line 162: **200ms "No activity" flash on disconnect.**

The disconnect effect (line 100) resets `filter` and `searchInput` synchronously but clears the query through `applySearch("")`, which is the 200ms debounce. So `searchQuery` — and therefore `isNarrowed` — stays truthy for that window.

`ActivityEmptyState` checks `isFiltered` **before** `!isConnected` (`ActivityEmptyState.tsx:21` vs `:29`), so a user who disconnects with text in the box gets the "No activity" card for 200ms before the connect prompt replaces it.

Calling `setSearchQuery("")` directly in that effect — alongside the other two synchronous resets — fixes it and makes the three resets consistent.

services/vault/src/types/activityLog.ts
```diff
@@ -77,7 +77,7 @@ export interface ActivityLog {
    * Whether the deposit expired before activation and was reclaimed via the
    * peg-in refund path. Shown as a red dot with the "Deposit expired" tooltip.
```
gbarkhatov, line 78: **Stale doc on the very field this PR renames.**

"Shown as a red dot with the 'Deposit expired' tooltip" describes UI deleted in this PR along with `statusDot.ts` — it is now the Refund chip plus a dimmed row.

Same stale claim at `services/activity/projection.ts:196-197` ("The UI renders this as a Deposit row with a red dot"). That line is outside the diff so I can't anchor a comment there, but it wants the same edit.

Worth flagging that the sibling doc two fields down *was* updated (`LiquidationChildRow.label`, line 100), so this is an inconsistent pass rather than a deliberate omission.

services/vault/src/copy.ts
```diff
@@ -1863,29 +1866,32 @@ export const COPY = {
       "Pending Deposit": "Deposit",
     } satisfies Record<ActivityType, string>,
     hashPending: "Pending…",
+    // Label above the row's hash link. The name is derived from the explorer
+    // the row actually links to (mempool for BTC, the chain explorer for ETH),
+    // never hardcoded.
+    explorerLabel: (explorerName: string) => `${explorerName} Explorer :`,
```
gbarkhatov, line 1872: **Space before the colon: "Mempool Explorer :".**

`timeLabel` on the next line is `"Time:"`. The two render side by side as sibling cell labels in the same row, with different colon spacing.

I see the frame draws "Xangle Explorer :", so this may be faithful to Figma — but a space before a colon is French typography, not American English, and `copy.ts`'s style block calls for American English. Suggest `` `${explorerName} Explorer:` ``.

services/vault/src/components/Activity/statusDot.ts
```diff
@@ -1,11 +0,0 @@
-/**
- * Status dot colors for the v3 activity rows, from the Figma v3 foundation
- * palette (foundation red/a400, amber/600, green/a400, warning/dark) — the same
- * `--risk-*` channels the v3 risk card uses, so the two surfaces can't drift.
- */
-export const STATUS_DOT = {
-  expired: "bg-risk-red",
-  pending: "bg-risk-amber",
-  settled: "bg-risk-green",
-  liquidated: "bg-risk-orange",
```
gbarkhatov, line 10: **`--risk-orange` is dead once this file goes.**

`STATUS_DOT.liquidated` was its only consumer — `grep -r risk-orange src/` now returns nothing but the two definition sites:

- `services/vault/src/globals.css:34`
- `services/vault/tailwind.config.ts:26`

The var even carries a comment naming its now-deleted consumer ("the activity feed's partially / fully liquidated dot"). ZERO DEAD CODE policy — both should go with it.

The other four `--risk-*` tokens stay live (`RiskPriceRail`, `RiskSection`, `railLayout`, `NotificationCard`), and I checked: the rest of the `statusDot.ts` deletion is clean — no dangling imports, and `statusPending` / `statusExpired` / `statusInUse` / `statusDone` came out of `copy.ts` with no remaining references.

services/vault/src/components/Activity/usdSubLine.ts
```diff
@@ -0,0 +1,24 @@
...
+ * "$0 USD" one.
+ */
+export function getUsdSubLine(
+  amount: ActivityAmount,
+  prices: Record<string, number> | undefined,
+): string | null {
+  const price = prices?.[amount.symbol];
+  if (amount.numeric === undefined || price === undefined) return null;
+  const usd = amount.numeric * price;
+  if (!Number.isFinite(usd) || usd <= 0) return null;
```
gbarkhatov, line 22: **Nit — the `usd <= 0` guard doesn't achieve what the docblock above claims.**

"Null renders no sub-line at all, never a '$0 USD' one" — but `formatUsdValue` rounds to two decimals, so any sub-cent product clears the guard and renders `"$0.00 USD"`. A dust row of 1 sat at $90,000/BTC gives `usd = 0.0009` → `$0.00 USD`, exactly the output the guard exists to prevent.

Suppressing a genuine zero is defensible; the sub-cent case is the gap. A `usd < 0.005` floor would match the docblock.

(`Number.isFinite` is the right guard for the NaN/Infinity path and I found no reachable NaN leak into it — `toNumericAmount` is finite for any valid bigint, `pendingActivities.ts:42` already guards the localStorage string, and a NaN price is caught here.)

services/vault/src/components/Activity/activitySearch.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import { COPY } from "@/copy";
+import type { ActivityRow } from "@/types/activityLog";
+
+/** Hashes are matched with and without the `0x` prefix, so pasting either form
+ *  from an explorer finds the row. */
+function matchesHash(hash: string, query: string): boolean {
+  const normalized = hash.toLowerCase().replace(/^0x/, "");
+  return normalized.includes(query.replace(/^0x/, ""));
```
gbarkhatov, line 14: **Nit — a bare `0x` query matches every row.**

`query.replace(/^0x/, "")` turns `"0x"` into `""`, and `normalized.includes("")` is always `true`. Typing `0x` returns the whole feed, BTC rows included — which is the opposite of what a user pasting a hash prefix expects.

Cheap guard: return `false` when the stripped query is empty (the caller already short-circuits the genuinely-empty query at line 26).

services/vault/src/components/Activity/FilterDropdown.tsx
```diff
@@ -48,7 +48,7 @@ export function FilterDropdown<V extends string>({
         onClick={() => setOpen((v) => !v)}
         aria-haspopup="listbox"
         aria-expanded={open}
-        className="flex items-center gap-2 rounded-lg border border-secondary-strokeDark px-4 py-2 text-sm text-accent-primary"
+        className="flex h-9 items-center gap-2 rounded-lg bg-secondary-strokeLight py-3 pl-4 pr-3 text-sm leading-[1.43] tracking-[0.17px] text-accent-primary"
```
gbarkhatov, line 51: **Nit — three small things on this line.**

1. **`py-3` is inert.** `h-9` fixes the box at 36px; with `border-box`, `py-3` (12px × 2) plus a 20px line-height needs 44px, so the padding is simply overridden. Drop it or drop `h-9`.
2. **2px height mismatch with the search box.** This trigger is 36px; `ActivitySearchInput` is `1px border + py-2 (8px) + 20px line-height + 8px + 1px` = 38px. They sit side by side under `items-center`, so it reads as two controls of unequal height.
3. **`bg-secondary-strokeLight` is a stroke token used as a fill.** It happens to land on the right value, but the semantic name will mislead the next person to touch either surface.

Also: the `align` docblock at lines 20-22 still says *"v3 puts it at the left of the row, v2 at the right"* — this PR moves the v3 trigger to the right (`align="end"`), so that comment is now backwards.

gbarkhatov (approved):
Reviewed the Activity restyle end to end — both commits, every non-test hunk, plus the surrounding files the diff leans on (`ActivityEmptyState`, `ExpiredWithdrawButton`, `ActivityHashLink`, `ListRow`, `explorer.ts`, `useDebounce`, `usePrices`, the core-ui palette).

The restructuring is sound and the rename is complete. Approving — nothing here blocks the merge. **The accessibility findings are the ones I'd act on before shipping**, since they affect every refunded row and the search box on every viewport; the rest are polish, dead code and stale docs.

**Ordered by what I'd fix first:**

1. `opacity-60` on the refunded row → all text 2.5–3.3:1 (ActivityRowV3.tsx:120)
2. Refund chip fails AA in dark mode *before* dimming (ActivityRowV3.tsx:146)
3. Mobile toolbar overflows horizontally (ActivityList.tsx:141)
4. Search placeholder at 1.96:1 is the box's only visible description (ActivitySearchInput.tsx:36)
5. Liquidation Partially/Fully classification renders nowhere, but filter + search still key on it (LiquidationGroupCardV3.tsx:38)
6. Expired-but-unreclaimed deposit lost its only status cue (ActivityListWithRefund.tsx:59)
7. "Last week" erases the calendar day for rows 2–7 days old (formatting.ts:336)
8. Empty 168px action slot when `ExpiredWithdrawButton` returns `null` (ActivityRowV3.tsx:202)
9. 200ms "No activity" flash on disconnect (ActivityList.tsx:162)
10. Stale JSDoc on `isRefunded`, and the same claim in `projection.ts` (activityLog.ts:78)
11. "Mempool Explorer :" — space before the colon (copy.ts:1872)
12. `--risk-orange` is dead now (statusDot.ts, deleted side)
13–15. Nits: sub-cent `$0.00 USD` (usdSubLine.ts:22), bare `0x` matches everything (activitySearch.ts:14), FilterDropdown sizing + stale docblock (FilterDropdown.tsx:51)

---

### One thing about the PR description

> Per the brief the derivation stays wired — `useVaults` → `activeVaultIds` still flows from the container through the list into the row's props — but no component reads it

The diff **deletes all of it**: `useVaults`, the `activeVaultIds` memo, the prop on `ActivityListProps`, and the `useVaults` mock in the page test. Deleting is the right call under the zero-dead-code policy — but design will read that paragraph and conclude the data is one component away from being renderable, when re-wiring it means restoring the hook, the memo and two prop hops. Worth correcting in the description so the "In use" design question is asked with the real cost attached.

### Things I checked and cleared

- **`isExpired` → `isRefunded` rename** is complete — no dangling references in source, tests, mocks or `demoDeposit`.
- **`getExplorerName`** is consistent with `getExplorerTxUrl` in every reachable configuration. BTC and ETH both go through the same host source; `explorerUrl` is a hardcoded literal per chain and `configureBabylonConfig` rejects any `ethChainId` other than 1/11155111, so `new URL()` cannot throw and there is no unknown-chain path (`ActivityChain` is a closed union). The only theoretical break is a multi-label public suffix (`mempool.emzy.de` → "Emzy"), which is not configured anywhere.
- **`activityRowMatchesSearch`** is safe across the `ActivityRow` union — both variants carry `transactionHash`, and `typeLabels` is exhaustive over `ActivityType` including both liquidation types.
- **`useDebounce`'s returned callback is stable** (`useCallback` over a stable `setState` and a module constant), so the disconnect effect can't loop on it.
- **`twMerge` in `ListRowCard`** does what the comment says: tailwind-merge classifies `bg-[#202020]` as a background-color arbitrary value and keys conflicts per-modifier, so `dark:bg-transparent` genuinely beats it while `CARD_SHELL_CLASS`'s border and `overflow-hidden` survive. The other two callers are undisturbed.
- **`transactionHash !== ""`** — `""` is the real sentinel, not a stand-in for `undefined`: the field is a required `string`, `resolveDisplayTx` returns `""` explicitly, and the pending path reads a required `Hex`.
- **Spinner a11y** — `role="status"` + `aria-label` on a wrapper around core-ui's `Loader` (a bare `<span><svg>` with no ARIA of its own) is correct; no duplicate role, no competing accessible name.
- **Magic numbers / motion** — `SPINNER_SIZE`, `EXTERNAL_ICON_SIZE`, `SEARCH_ICON_SIZE`, `SEARCH_DEBOUNCE_MS` are all named constants; no user-facing string is inlined outside `COPY`; no new animation is introduced, so `docs/motion-system.md` isn't engaged.
- **Pending/refunded precedence** (`isPending === true && !isRefunded`) is correct. The two are mutually exclusive in practice — `isRefunded` is only set by `projectRefundedDeposit`, which never sets `isPending` — so the guard is defensive rather than load-bearing. The one reachable overlap is indexer-vs-contract lag, and that case is covered by the dimmed-actionable-button note on line 120.

### Test run

I could not execute `ActivityList.test.tsx`, `ActivityRowV3.test.tsx` or `LiquidationGroupCardV3.test.tsx` locally — all three fail at **collection**, not assertion, on `Failed to resolve import "@babylonlabs-io/babylon-tbv-rust-wasm/raw"`. The installed wasm package's `exports` map has only `"."`, while the built `ts-sdk` dist imports `./raw`; nothing in this PR touches either package, so this is the stale-`dist` environment issue `CLAUDE.md` warns about, not a regression here. The rest of the suites I targeted passed: **113 tests, 4 files green** (`activitySearch`, `ExpiredWithdrawButton`, `FilterDropdown`, `formatting`), including the new "Last week" and future-date cases. The PR reports the full suite green in a correctly built tree.

jrwbabylonlab (approved):

PR #2290, opened 2026-08-20
fix(vault): stop labelling pre-broadcast failures as broadcast failures

jrwbabylonlab:
fix(vault): stop labelling pre-broadcast failures as broadcast failures

## What

`broadcastPrePeginTransaction` now wraps each stage under its own label with the original error as `cause`: "Failed to prepare Pre-Pegin transaction" (hex parse and prevout resolution), "Failed to sign Pre-Pegin transaction" (the intent ceremony, the wallet signature and the integrity checks on the returned PSBT) and "Failed to broadcast Pre-Pegin transaction" (the network push only). The labels live in `copy.ts` as `COPY.deposit.errors.prePeginStageFailed`, and `mapDepositError` matches them: two new callouts, "Signing failed" and "Preparation failed", and the broadcast callout is selected only by the explicit broadcast label. The old bare `"broadcast"` substring match is gone. The typed DepositTermsRejected and depositor-wallet mismatch errors still pass through the service unwrapped. The broadcast-stage lifecycle refusal (a batch member no longer PENDING on resume) maps to a terminal "Deposit can't be resumed" callout instead of the retryable broadcast copy.

The fresh deposit flow no longer re-wraps the service error, so the stage label and `cause` chain reach the mapper intact. It re-probes wallet liveness and UTXO availability after the Ethereum finality gate, since that wait is when a wallet locks or an input goes stale, and advances the stepper to the broadcast step first so those probes are attributed to it. Because the vaults are already registered on Ethereum at that point, a non-rejection signing failure, a liveness failure, or a UTXO re-check that could not reach the mempool sets `resumableVaultIds`, so the modal offers Retry the way it already does for a locked Ledger. An input found spent after registration gets its own terminal callout with no Retry.

Tests: the service pins each stage label and its `cause`; the mapper pins the new buckets, the sign-stage rejection paths, the terminal batch callout, the resumable set and that a bare "broadcast" mention no longer selects the broadcast callout; the flow pins the post-gate probe ordering and the Retry handoff for a wallet that locks during the gate, fails to sign after registration, or hits a mempool outage on the re-check, and the terminal callout for a spent input.

## Why

A wallet or prevout failure before the Pre-PegIn ever reached the network rendered as "We couldn't broadcast your Bitcoin transaction" (devnet report, 2026-08-19). The broadcast service wrapped every stage under one broadcast label, the flow wrapped that again, and the mapper matched the bare word "broadcast", which also appears in unrelated messages.

## Approaches

- Stage labels from `copy.ts`, matched by the mapper: chosen. Both sides read one constant, so the label and the match cannot drift, and the mapper stays a pure function of the error.
- A typed `PrePeginStageError` matched by `instanceof`: not taken here. It is a broader refactor of the mapper's message tier and does not change what the depositor sees.
- Keeping the flow-level re-wrap: rejected. It replaced the stage label with a broadcast one.
- Extending the resumable set to `preparationFailed` and `broadcastFailed`: rejected. Fresh-flow preparation failures are deterministic, and a broadcast failure may already have reached the network.
- A typed bucket for the PSBT integrity checks: rejected. They land in "Signing failed", whose copy offers the lock as a possibility, and every check runs before the network push, so a retry is safe.

## Not in this PR

- Pre-registration, a spent input still shows the SDK's own message under the generic title, as it did before this change.

## Pre-review

<details>
<summary>Pre-review: 19 findings · 6 fixed · 0 moot · 0 follow-up · 2 declined · 11 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | Typed lifecycle refusal (stage broadcast) still maps to the retryable broadcast-failed copy | normal | fixed |
| 2 | Post-gate liveness and UTXO probes run before the step advances to BROADCAST_PRE_PEGIN | normal | fixed |
| 3 | A transient prevout-fetch failure in the post-gate UTXO check is not resumable in the modal | normal | fixed |
| 4 | The 'Unknown error' sentinel is a literal in stageError and again in formatting.ts and the mapper fallback | normal | declined: pre-existing sentinel shared with formatting.ts; a separate cleanup |
| 5 | Stage routing is coupled to copy wording rather than a typed stage error | normal | declined: both sides read one COPY constant and tests pin it; a broader refactor than this fix |
| 6 | No test pins that the bare 'broadcast' substring bucket is gone | normal | fixed |
| 7 | Redundant mockClear in the liveness-ordering flow test | normal | fixed |
| 8 | Stale comment says the deposit flow's broadcast catches re-wrap wallet errors | normal | fixed |
| 9 | Mapper 3e comment says 'nothing was sent', the claim the copy was reworded to avoid | normal | open |
| 10 | The spent-input callout is flow-local; the resume path shows the SDK's raw sentence for the same typed error | normal | open |
| 11 | The spent-input callout asserts the deposit is over from a single mempool read while the dashboard still offers Broadcast | normal | open |
| 12 | Vacuous not.toEqual assertion in the lifecycle-refusal mapper test | normal | open |
| 13 | copy.ts guard comment overstates the collision surface | normal | open |
| 14 | No test pins that the step advances before the post-gate probes | normal | open |
| 15 | batchNoLongerPending says 'the batch' for a single-vault deposit | normal | open |
| 16 | preparationFailed copy sends the user to the dashboard, but its usual producer is the dashboard resume | normal | open |
| 17 | Seven-line stacked comment block in the flow catch | normal | open |
| 18 | vaultLifecycleStateError.ts says the broadcast-stage refusal keeps its existing mapping, now false | normal | open |
| 19 | cannotBroadcastInState says 'the vault', against the copy header rule | normal | open |

</details>

### DIFF
--- a/services/vault/src/copy.ts
+++ b/services/vault/src/copy.ts
@@ -782,11 +782,8 @@ export const COPY = {
         "New deposits are temporarily disabled while the protocol is frozen or paused. No Bitcoin was sent — please try again once it resumes.",
       cannotActivateInState: (state: string) =>
         `Cannot activate: BTC Vault is in ${state} state. Activation is only valid when VERIFIED.`,
-      // Deliberately worded without the token "broadcast". These are state
-      // preconditions, not broadcast failures, and `mapDepositError` matches
-      // "broadcast" on the message — which would replace this precise sentence
-      // with "Broadcast failed / please try again", wrong for a terminal state
-      // like EXPIRED where retrying can never succeed.
+      // Worded without "failed to broadcast": colliding with the mapper's
+      // stage label would show retryable broadcast copy for a terminal state.
       cannotBroadcastInState: (state: string) =>
         `Cannot continue: BTC Vault is in ${state} state. This step is only valid while the vault is PENDING.`,
       cannotBroadcastInOnChainState: (state: string) =>
@@ -853,6 +850,14 @@ export const COPY = {
       signingRejected: {
         title: "Signing rejected",
         body: "You rejected the request in your wallet. Click Retry to approve it and continue.",
+      },
+      signingFailed: {
+        title: "Signing failed",
+        body: "Your Bitcoin wallet couldn't sign the transaction. Check that your wallet is unlocked, then try again. No Bitcoin has been broadcast.",
+      },
+      preparationFailed: {
+        title: "Preparation failed",
+        body: "We couldn't prepare your Bitcoin transaction for signing. Please try again in a moment. No Bitcoin has been broadcast.",
       },
       walletNotConnected: {
         title: "Wallet not connected",
--- a/services/vault/src/hooks/deposit/useDepositFlow.ts
+++ b/services/vault/src/hooks/deposit/useDepositFlow.ts
@@ -16,7 +16,6 @@ import type { BitcoinWallet } from "@babylonlabs-io/ts-sdk/shared";
 import {
   ensureHexPrefix,
   forwardDepositApproval,
-  isDepositTermsRejectedError,
   isRegisteredVaultVersionMismatchError,
   stripHexPrefix,
   validateOnChainParticipantKeys,
@@ -833,6 +832,25 @@ export function useDepositFlow(
           expected: validatedKeys.participantKeys,
         });
 
+        // A flow abandoned during the minutes-long gate must not raise an
+        // unlock prompt or signing popup (same rule as the resume path).
+        signal.throwIfAborted();
+
+        // The wallet may have locked during the gate wait; probe before the
+        // signing popup so it fails with liveness copy, not a dead popup.
+        await verifyBtcWalletLiveness(confirmedBtcWallet, confirmedBtcAddress, {
+          probeConnection: shouldProbeWalletLiveness(
+            btcConnector?.connectedWallet?.id,
+          ),
+        });
+
+        // The inputs were validated before ETH registration; the gate
+        // stretched that window to minutes, so re-check before signing.
+        await assertUtxosAvailable(
+          batchResult.fundedPrePeginTxHex,
+          confirmedBtcAddress,
+        );
+
         // ========================================================================
         // Step 4b: Broadcast Pre-PegIn transaction to Bitcoin
         // Broadcast immediately after ETH registration so the VP can verify
@@ -845,35 +863,24 @@ export function useDepositFlow(
           vaultAmounts.map(() => DepositFlowStep.BROADCAST_PRE_PEGIN),
         );
 
-        let prePeginBroadcastTxid: string;
-        try {
-          prePeginBroadcastTxid = await broadcastPrePeginTransaction({
-            unsignedTxHex: batchResult.fundedPrePeginTxHex,
-            btcWalletProvider: {
-              signPsbt: (psbtHex: string) =>
-                confirmedBtcWallet.signPsbt(psbtHex),
-              deriveContextHash: (appName: string, context: string) =>
-                confirmedBtcWallet.deriveContextHash(appName, context),
-              // Object spread drops prototype methods — see forwardDepositApproval.
-              ...forwardDepositApproval(confirmedBtcWallet),
-            },
-            depositorBtcPubkey: batchResult.depositorBtcPubkey,
-            expectedUtxos: utxosToExpectedRecord(batchResult.selectedUTXOs),
-            depositTerms: batchResult.depositTerms,
-          });
-        } catch (error) {
-          // Preserve a typed intent rejection so the error mapper can show the
-          // intent-rejection copy instead of a generic broadcast failure — the
-          // broadcast service already keeps it typed at its boundary.
-          if (isDepositTermsRejectedError(error)) {
-            throw error;
-          }
-          const errorMsg =
-            error instanceof Error ? error.message : String(error);
-          throw new Error(
-            `Failed to broadcast batch Pre-PegIn transaction: ${errorMsg}`,
-          );
-        }
+        // An abort landing during the probes must not raise the signing popup.
+        signal.throwIfAborted();
+
+        // No re-wrap: the service labels each stage and preserves `cause`;
+        // a wrapper here would poison the label and sever that chain.
+        const prePeginBroadcastTxid = await broadcastPrePeginTransaction({
+          unsignedTxHex: batchResult.fundedPrePeginTxHex,
+          btcWalletProvider: {
+            signPsbt: (psbtHex: string) => confirmedBtcWallet.signPsbt(psbtHex),
+            deriveContextHash: (appName: string, context: string) =>
+              confirmedBtcWallet.deriveContextHash(appName, context),
+            // Object spread drops prototype methods — see forwardDepositApproval.
+            ...forwardDepositApproval(confirmedBtcWallet),
+          },
+          depositorBtcPubkey: batchResult.depositorBtcPubkey,
+          expectedUtxos: utxosToExpectedRecord(batchResult.selectedUTXOs),
+          depositTerms: batchResult.depositTerms,
+        });
 
         // Broadcast succeeded — update pending pegins from PENDING to CONFIRMING
         for (const peginResult of peginResults) {
--- a/services/vault/src/utils/errors/depositErrors.ts
+++ b/services/vault/src/utils/errors/depositErrors.ts
@@ -19,6 +19,10 @@
  *  - Wallet not connected / wallet client missing.
  *  - Wallet account changed mid-flow (the WOTS-vs-PoP key guard).
  *  - Wrong wallet connected on resume (WOTS hash mismatch).
+ *  - Preparation failure — the Pre-PegIn could not be prepared for signing
+ *    (e.g. prevout resolution against the mempool API failed).
+ *  - Signing failure — the wallet could not sign the Pre-PegIn and it was not
+ *    a rejection (locked wallet, stale extension, device transport drop).
  *  - Broadcast failure — Pre-PegIn could not be broadcast to Bitcoin.
  *  - Insufficient ETH — the Ethereum registration tx can't cover gas. Detected
  *    via the shared `classifyError` (viem typed error + node-message regex),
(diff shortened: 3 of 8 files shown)


services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1069,57 +1069,67 @@ export function useDepositFlow(
...
+        // signing popup so it fails with liveness copy, not a dead popup.
+        await verifyBtcWalletLiveness(confirmedBtcWallet, confirmedBtcAddress, {
+          probeConnection: shouldProbeWalletLiveness(
+            btcConnector?.connectedWallet?.id,
+          ),
+        });
+
+        // The inputs were validated before ETH registration; the gate
+        // stretched that window to minutes, so re-check before signing.
+        await assertUtxosAvailable(
```
gbarkhatov: This re-check runs after Ethereum registration, but neither of its failures maps to useful copy or offers Retry:

- A spent input throws `UtxoNotAvailableError`, whose message tells the user to "create a new peg-in request with a different UTXO" (`ts-sdk availability.ts:58-61`). That's wrong advice for a deposit that's already registered.
- A mempool API failure throws `Failed to get UTXOs for address …` (`mempoolApi.ts:416`).

Neither matches bucket 9's phrases (`spendable utxos` / `utxos available` / `failed to load utxos`), so both land in the generic callout with raw text. Neither is in `RESUMABLE_AFTER_REGISTRATION` or is a `BtcWalletLivenessError`, so `resumableVaultIds` isn't set (`:1694-1701`). The liveness probe two lines up does get Retry.

Net effect: a brief mempool API hiccup at this point leaves registered vaults with raw copy and no in-modal Retry. Suggest treating a fetch failure here as resumable like the liveness error, and mapping `UtxoNotAvailableError` by type to copy that says the deposit is registered and will expire.

services/vault/src/copy.ts
```diff
@@ -987,6 +994,12 @@ export const COPY = {
         title: "Broadcast failed",
         body: "We couldn't broadcast your Bitcoin transaction to the network. Please try again.",
       },
+      // Resume refused because a vault in the batch left PENDING: terminal, so
+      // no retry advice.
+      batchNoLongerPending: {
+        title: "Deposit can't be resumed",
+        body: "A vault in this deposit is no longer awaiting its Bitcoin transaction, so the batch can't be broadcast as one transaction. No Bitcoin was sent.",
```
gbarkhatov: "No Bitcoin was sent" isn't always true here. `assertBatchLifecycleStatus` throws with stage `broadcast` when the target or any sibling isn't PENDING (`rebuildDepositTerms.ts:139-157`), including VERIFIED. VERIFIED is set only once every ACK is in (`vault-contracts-aave-v4@ac22351c:src/protocol/lib/PeginLogic.sol:807-809`), and the VP only reaches `PendingACKs` after `PendingPrePegInConfirmations` (`btc-vault@a777d00c:crates/vaultd/src/workers/claimer/mod.rs:74`). Batch members share one Pre-PegIn, so a VERIFIED member means the shared transaction is already on Bitcoin. That can happen, for example, when resuming from a stale tab after broadcasting from another.

Suggest dropping that sentence, or branching on `err.status` so the "nothing was sent" wording is only used when no member got past PENDING.

Also a copy-style nit: "A vault in this deposit" breaks the header rule (`copy.ts:22-24`, "BTCVault … never bare 'vault'"). "A BTCVault in this deposit" fixes it.

services/vault/src/utils/errors/depositErrors.ts
```diff
@@ -126,8 +130,12 @@ const RESUMABLE_AFTER_REGISTRATION: ReadonlySet<DepositErrorContent> = new Set([
   ERRORS.deviceWrongApp,
   ERRORS.deviceCeremonyInvalid,
   ERRORS.signingRejected,
+  // Nothing was broadcast; the software-wallet twin of deviceLocked.
```
gbarkhatov, line 134: Author's call, since the copy comment shows the wording is deliberate. The sign stage also wraps the terms-mismatch check, `assertPsbtUnsignedTxMatches`, the returned-signature checks and finalization, so all of those now map to a resumable `signingFailed`. A wallet that hands back a wrong PSBT will fail every Retry, with "unlock it" advice. Funds are safe because it all throws before `pushTx`. If you want to separate them, a typed integrity error mapped to a non-resumable callout would do it.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1069,57 +1069,67 @@ export function useDepositFlow(
...
+          batchResult.fundedPrePeginTxHex,
+          confirmedBtcAddress,
+        );
+
+        // An abort landing during the probes must not raise the signing popup.
+        signal.throwIfAborted();
+
+        // No re-wrap: the service labels each stage and preserves `cause`;
```
gbarkhatov: Nit: the removed wrapper did pass `{ cause: error }` (old lines 400-402), so "sever that chain" isn't accurate. The real reason is the label. Relatedly, no flow test fails if a wrapper is reintroduced here: `Failed to broadcast batch Pre-Pegin transaction: <inner label>` still reaches 6b/6c/7 through `includes`. If the no-re-wrap rule matters, a flow test asserting the mapped callout for a sign-stage failure would pin it.

gbarkhatov (approved):
Nothing blocking. The signed and broadcast bytes are unchanged, and the stage labels are right: only the `pushTx` block carries the broadcast label, prevout and PSBT-build failures get prepare, and the terms ceremony and returned-PSBT checks get sign. `cause` survives with a single wrapping layer, and cancellation, method-not-supported and device-code detection still run ahead of the label buckets. Four inline points. The first two change what a depositor is told after registration.

**Not anchorable**
- `depositErrors.ts:34-35`: the header still says the lifecycle refusal's "broadcast stage keeps its historical broadcast-bucket copy". Bucket 3e now maps it to `batchNoLongerPending`.
- Nit: `preparationFailed` says "Please try again", but it's deliberately not in `RESUMABLE_AFTER_REGISTRATION`, so after registration there is no Retry button. Pointing at the dashboard resume would match what the user can actually do.

**Checked and clean**
- The post-gate re-check doesn't re-select or alter inputs. `verifyBtcWalletLiveness` compares addresses, and `assertUtxosAvailable` re-checks the same `fundedPrePeginTxHex` and address as the pre-registration check. Neither falls back, and `signal.throwIfAborted()` brackets both.
- Typed refusals (terms rejected, depositor wallet mismatch, BTC key mismatch) still pass through unwrapped.
- The late-cancel throw still carries `WALLET_CONNECTION_REJECTED_CODE`.
- A `BtcWalletLivenessError` becomes resumable only when `registeredVaultIds !== null`.
- `prePeginBroadcastFailed` and the flow's `isDepositTermsRejectedError` import have no references left, and no `data-testid` changed.
- `useVaultActions.ts` already ran the same liveness and UTXO checks on resume; only a comment changed there.

**Considered and dropped**
- "A real broadcast failure could now get prepare or sign copy": no, only `pushTx` is inside the broadcast-labelled try.
- "The re-check changes what gets signed": no, `expectedUtxos`, `depositTerms` and the pubkey are passed through unchanged.

services/vault/src/hooks/deposit/useDepositFlow.ts

services/vault/src/copy.ts

services/vault/src/utils/errors/depositErrors.ts
jrwbabylonlab: not planning — retry is safe; every check precedes pushTx

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: fixed; not planning — includes-matching makes that pin vacuous

gbarkhatov (dismissed):

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1069,57 +1069,67 @@ export function useDepositFlow(
...
-          prePeginBroadcastTxid = await broadcastPrePeginTransaction({
-            unsignedTxHex: batchResult.fundedPrePeginTxHex,
-            btcWalletProvider: {
-              signPsbt: async (psbtHex: string) => {
-                const signedPsbtHex = await runCancellableSign(
-                  confirmedBtcWallet,
-                  () => confirmedBtcWallet.signPsbt(psbtHex),
+        // A flow abandoned during the minutes-long gate must not raise an
+        // unlock prompt or signing popup (same rule as the resume path).
+        signal.throwIfAborted();
```
jonybur, line 1086: **The two new post-finality signal.throwIfAborted() guards have no test; removing either leaves the suite green.**

This change adds `signal.throwIfAborted()` before the post-gate liveness probe (line 1081) and again before `broadcastPrePeginTransaction` (line 1099). The comments promise that a deposit abandoned during the finality wait never raises an unlock prompt or a signing popup. The only abort test in `useDepositFlow.test.tsx` (around line 1849) aborts during batch creation, long before registration, so nothing exercises either guard. If one is dropped in a later refactor, an abandoned flow would pop the wallet's signing dialog and the tests would still pass. Please add a flow test that aborts while `waitForEthRegistrationDepth` is pending and asserts that `verifyBtcWalletLiveness` and `broadcastPrePeginTransaction` are never called, plus one that aborts during the probes and asserts no broadcast.

jonybur (commented):
Comments inline.

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: fixed

services/vault/src/copy.ts
```diff
@@ -991,10 +1001,25 @@ export const COPY = {
         title: "Bitcoin funds unavailable",
         body: "We couldn't confirm your Bitcoin funds are available. They may be in use by another deposit. Please try again in a moment.",
       },
+      // Post-registration only: a Pre-Pegin input was not found unspent. One
+      // mempool read decides this, so point at the dashboard, not at a
+      // new deposit; the registration expires on its own if it stays true.
+      inputSpentAfterRegistration: {
+        title: "Bitcoin funds no longer available",
+        body: "A Bitcoin input for this deposit is no longer available, so the Pre-Pegin can't be broadcast from here. The Ethereum registration will expire on its own; check the dashboard for the deposit's current status.",
```
jonybur, line 1009: **On resume, a Pre-Pegin input spent by this deposit's own earlier broadcast is reported as terminal, with copy saying the registration will expire.**

The resume path now maps every UtxoNotAvailableError from the address-UTXO re-check in useVaultActions.ts (line 439, mapped at line 570) to inputSpentAfterRegistration, which tells the user the Ethereum registration will expire on its own. That check only reads the address's unspent outputs from mempool, so it cannot tell a foreign spend from this deposit's own Pre-Pegin. If an earlier pushTx was accepted but its response timed out, the user sees "Broadcast failed ... Please try again" while the vault stays PENDING. The retry then finds the inputs spent by their own transaction and says the deposit is dead while it is actually progressing, which may push them to abandon it or start a duplicate. Before choosing this callout, check whether the registered Pre-Pegin txid (or the spender of the missing outpoints) is already in the mempool or on-chain and treat that as already broadcast, or drop the expiry claim from the copy.

jonybur (commented):
Comments inline.

gbarkhatov (approved):

PR #2293, opened 2026-08-20
fix(vault): move the max-vaults notice below the Position section

jeremy-babylonlabs:
## What

On the Overview page, the "Maximum vaults reached" notice rendered above the Position section while the other notification cards render below it. This moves `MaxVaultsNotification` into the shared below-Position slot, between the Position stats and the cascade banner.

## Why this placement

Figma frame `10094-26791` shows the Notifications instance (`10182:57535`) rendering the max-vaults content directly below the Position block, and frame `10204-45310` shows the identically-positioned instance rendering the cascade (critical-state) content — the two banners are one design slot with alternate content, not two stacked positions. The previous top placement cited frame `6508-114810`, which is superseded; the stale comment is replaced with the current frame references.

`CriticalLiquidationTopBanner` intentionally stays at page top: per frame `10204-45310` it is a full-bleed bar spanning above the sidebar and header (y=0, full page width), and the component portals into the root-layout slot above the content row (`10204-45613`). It is not part of the Position column.

## Changes

- `DashboardPage.tsx`: render order is now Position → max-vaults notice → cascade banner → Risk. Render conditions are unchanged (the notice stays always-on, independent of the liquidation-notifications flag).
- `DashboardPage.test.tsx`: a DOM-order assertion locks the notice below the Position section so the regression cannot return silently.

Layout move only — no styling, copy, or behaviour changes.

### DIFF
--- a/services/vault/src/components/simple/DashboardPage.tsx
+++ b/services/vault/src/components/simple/DashboardPage.tsx
@@ -209,8 +209,9 @@ export function DashboardPage() {
       ? formatBasisPointsAsPercent(collateralFactorBps)
       : COPY.common.emptyValue;
 
-  // The cascade banner sits between the Position and Risk sections (Figma
-  // 10204-45310).
+  // The cascade banner and the max-vaults notice share the same slot between
+  // the Position and Risk sections — same "Notifications" instance in both
+  // the default and critical states (Figma 10094-26791, 10204-45310).
   const cascadeBanner = liquidationNotificationsEnabled ? (
     <PositionNotificationBanner
       connectedAddress={address}
@@ -235,18 +236,12 @@ export function DashboardPage() {
   return (
     <Container className={`${PAGE_CONTENT_CLASS} pb-6`}>
       <div className="space-y-10">
-        {/* Notifications sit above Overview per Figma (frame 6508-114810). The
-            critical top banner, the max-vaults notice, and the cascade banner
-            share this slot. */}
+        {/* Full-bleed alert bar above the header/sidebar row, not part of this
+            column — it portals into RootLayout's top-banner slot (Figma frame
+            10204-45613; see CriticalLiquidationTopBanner). */}
         {liquidationNotificationsEnabled && (
           <CriticalLiquidationTopBanner result={criticalBannerResult} />
         )}
-
-        {/* "Maximum vaults reached" is a value-protection capacity fact shown
-            ALWAYS (independent of the liquidation-notifications flag and of BTC
-            price), and decoupled from the cascade banner so a stale-price or
-            all-pending position still surfaces it. */}
-        <MaxVaultsNotification connectedAddress={address} />
 
         <OverviewSection
           totalCollateralValue={totalCollateralValue}
@@ -263,6 +258,12 @@ export function DashboardPage() {
           canBorrow={canBorrow}
           canRepay={hasLoans}
         />
+
+        {/* "Maximum vaults reached" is a value-protection capacity fact shown
+            ALWAYS (independent of the liquidation-notifications flag and of BTC
+            price), and decoupled from the cascade banner so a stale-price or
+            all-pending position still surfaces it. */}
+        <MaxVaultsNotification connectedAddress={address} />
 
         {cascadeBanner}
 
--- a/services/vault/src/components/simple/__tests__/DashboardPage.test.tsx
+++ b/services/vault/src/components/simple/__tests__/DashboardPage.test.tsx
@@ -121,6 +121,15 @@ describe("DashboardPage composition", () => {
     expect(screen.getByTestId("critical-banner")).toBeInTheDocument();
     expect(screen.getByTestId("position-banner")).toBeInTheDocument();
     expect(screen.getByText(COPY.risk.title)).toBeInTheDocument();
+
+    // Figma (10094-26791, 10204-45310): the max-vaults notice and the cascade
+    // banner share the slot below Position, not above it.
+    const overview = screen.getByTestId("overview-section");
+    const maxVaults = screen.getByTestId("max-vaults");
+    expect(
+      overview.compareDocumentPosition(maxVaults) &
+        Node.DOCUMENT_POSITION_FOLLOWING,
+    ).toBeTruthy();
   });
 });
 


jonybur (approved):
Layout-only move: MaxVaultsNotification now renders below OverviewSection, matching the Figma frames cited in the description (10094-26791 and 10204-45310) where the notice and the cascade banner share one slot below Position. Render conditions are unchanged, the CriticalLiquidationTopBanner stays where it is because it portals into a different layout slot, and the new DOM-order assertion pins the placement so the regression cannot come back silently. Nothing else in the diff changes behaviour.

services/vault/src/components/simple/__tests__/DashboardPage.test.tsx
```diff
@@ -121,6 +121,15 @@ describe("DashboardPage composition", () => {
     expect(screen.getByTestId("critical-banner")).toBeInTheDocument();
     expect(screen.getByTestId("position-banner")).toBeInTheDocument();
     expect(screen.getByText(COPY.risk.title)).toBeInTheDocument();
+
+    // Figma (10094-26791, 10204-45310): the max-vaults notice and the cascade
+    // banner share the slot below Position, not above it.
+    const overview = screen.getByTestId("overview-section");
+    const maxVaults = screen.getByTestId("max-vaults");
+    expect(
```
gbarkhatov, line 129: **1 — the order assertion has no upper bound**

`overview.compareDocumentPosition(maxVaults) & Node.DOCUMENT_POSITION_FOLLOWING` only pins "somewhere after Overview". Moving the notice below `{cascadeBanner}`, below `RiskSection`, or to the very bottom after `LiquidationAnalysisSection` all still pass — so the invariant this PR actually introduces (the notice sits *between* Position and the cascade banner) is not locked. One more line closes it:

```ts
expect(maxVaults.compareDocumentPosition(positionBanner)).toBe(
  Node.DOCUMENT_POSITION_FOLLOWING,
);
```

Secondary: the `& FOLLOWING` mask is also truthy inside `CONTAINED_BY | FOLLOWING` (20 & 4 = 4), so it cannot distinguish "sibling after" from "nested inside". Not reachable today — the `OverviewSection` mock at line 86 is a childless `<div />` that drops children — but it goes green on a wrong DOM the moment that mock becomes children-transparent. The strict form at `UtxoSplitSelectorV3.test.tsx:123` rejects both cases and fails with `expected 20 to be 4` rather than `expected +0 to be truthy`. (`DisconnectedOverview.test.tsx:69` uses the loose form too, so treat this as a strictness suggestion, not a house-style violation.)

services/vault/src/components/simple/__tests__/DashboardPage.test.tsx
```diff
@@ -121,6 +121,15 @@ describe("DashboardPage composition", () => {
     expect(screen.getByTestId("critical-banner")).toBeInTheDocument();
     expect(screen.getByTestId("position-banner")).toBeInTheDocument();
     expect(screen.getByText(COPY.risk.title)).toBeInTheDocument();
+
+    // Figma (10094-26791, 10204-45310): the max-vaults notice and the cascade
```
gbarkhatov, line 125: **2 — the placement claim is appended to a presence-only test**

The enclosing test is `it("renders the overview summary, the risk card and the safety notifications")` — four `toBeInTheDocument` calls. DOM order is a different behavior, and CLAUDE.md is explicit: "Each test verifies ONE specific behavior" / "Test name describes the exact behavior being tested." This directory already follows that for ordering assertions — `DisconnectedOverview.test.tsx:56` ("...chips ahead of the headline") and `:135` ("orders the feature rows so..."). As written, a regression reports under "renders … safety notifications", so the first read is "an element went missing", not "the layout moved". A separate `it("renders the max-vaults notice below the Position section")` would name what actually broke.

Secondary: this assertion only runs with `isLiquidationNotificationsEnabled = true` (set at line 115), while `beforeEach` defaults the flag to `false` and `MaxVaultsNotification` is documented as always-on and flag-independent. The flag-off configuration — where the notice is the sole occupant of the slot — has no placement coverage.

services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -209,8 +209,9 @@ export function DashboardPage() {
       ? formatBasisPointsAsPercent(collateralFactorBps)
       : COPY.common.emptyValue;
 
-  // The cascade banner sits between the Position and Risk sections (Figma
-  // 10204-45310).
+  // The cascade banner and the max-vaults notice share the same slot between
```
gbarkhatov, line 212: **3 — "same 'Notifications' instance" describes the Figma, not the code**

There is no shared slot or wrapper here: `MaxVaultsNotification` and `{cascadeBanner}` are two independent siblings of `space-y-10`, separately gated (`isAtCap` vs `liquidationNotificationsEnabled` + severity), stacking 40px apart when both render. A reader may go looking for a `Notifications` component that does not exist, and "same instance" invites someone to "fix" the apparent duplication by merging them — which the comment at 262-265 explicitly argues against. Something like "renders in the same region as the cascade banner (Figma draws one Notifications instance with alternate content)" would match the code.

Related: this comment documents JSX *ordering*, but sits on `const cascadeBanner` — an element factory with no say in placement, which is decided ~50 lines down at 262-268. So the one comment that would stop someone moving the notice back up is invisible from the site where that move would happen, and a future reorder leaves 212-214 silently false. Worth moving the placement sentence down to the JSX block.

If the two genuinely are one design slot, a wrapper with its own tighter spacing would express that better than inheriting the 40px section rhythm.

services/vault/src/components/simple/DashboardPage.tsx
```diff
@@ -264,6 +259,12 @@ export function DashboardPage() {
           canRepay={hasLoans}
         />
 
+        {/* "Maximum vaults reached" is a value-protection capacity fact shown
+            ALWAYS (independent of the liquidation-notifications flag and of BTC
+            price), and decoupled from the cascade banner so a stale-price or
+            all-pending position still surfaces it. */}
+        <MaxVaultsNotification connectedAddress={address} />
```
gbarkhatov, line 266: **4 — worth confirming the stacking order with design now that the two are adjacent**

Relative order is unchanged by this PR, but the whole Overview block used to sit between them; now they are 40px apart in one visual stack, yellow advisory first. Two things that reads on:

- `MaxVaultsNotification` self-gates on `isAtCap` only, while `cascadeBanner` renders over every non-`hidden` severity including red `urgent`. So an at-cap, near-liquidation user reads: Position → yellow "Maximum vaults reached" → red urgent card carrying the actionable deposit/repay buttons. Should the slot sort by severity instead?
- Both can resolve to `tone="too-many"` — `MaxVaultsNotification.tsx:39`, and `PositionNotificationBanner.tsx:77` via the `WARNING_TONE` map. Same border, same `InfoIcon` chip, same surface; different facts ("Maximum vaults reached" from on-chain `maxVaultsPerPosition`, vs "Too many BTC Vaults to optimize" from `nVaults > MAX_DP_N`, where `MAX_DP_N = 17`). Two visually identical yellow cards stack whenever `too-many-vaults` is the primary warning and the position is at cap — reachable only if the deployed `maxVaultsPerPosition` exceeds 17. That is a live contract read, so I could not settle it statically; if the deployed cap is <= 17, non-issue.

gbarkhatov (approved):
Approving — the layout move itself is correct and I found no correctness bug in it. Four comments inline, none merge-blocking; 1 and 4 are the two I would want answered.

Two things that could not be anchored to the diff:

**Mobile:** below `xl` the position panel stacks (`PositionStatCards.tsx:112`, `flex flex-col gap-6 xl:flex-row`), putting it at roughly 350-400px, so the notice moves off the first screen. The collateral card's Deposit button is hardcoded `actionDisabled: false` (`OverviewSection.tsx:65`), so an at-cap user on mobile now sees an enabled Deposit CTA with the only explanation scrolled out of view. Above `xl` both are visible together. If frame 10094-26791 covers small breakpoints too this is an accepted cost — one mobile screenshot before merge would settle it.

**Figma provenance:** this deletes `6508-114810`, the sole recorded justification for the old placement, and replaces it with different node IDs. There is no Figma file key anywhere in the repo, so none of these citations are resolvable from the code alone. For a pure-layout change, a before/after screenshot in the description is the cheapest thing that makes it reviewable without Figma access — and worth confirming `6508-114810` is genuinely superseded rather than a second still-live frame that disagrees, since the whole PR rests on that.

Checked and clean, for the record: the `null` return plus `space-y-10` produces no phantom gap in either position; `NotificationCard` carries no position-tuned margins; the portal claim is accurate (`createPortal` into `CRITICAL_BANNER_SLOT_ID`, rendered at `RootLayout.tsx:205` above the sidebar/content row); zero surviving `6508-114810` references; one consumer of the component; production testid `max-vaults-notification` untouched and not E2E-consumed; render conditions genuinely unchanged; no dead code or inlined copy. `DashboardPage.test.tsx` passes 4/4, and the new assertion does fail if the move is reverted.

PR #2295, opened 2026-08-21
feat(packages): load the Bitcoin engine only when needed

### LINKED ISSUE #2231: Complete Bitcoin engine safeguards and release checks

jonybur:
## Status

Closed without merge. Do not review or release this branch.

This was the first large implementation of #2231. It was split because it changed 67 files and mixed several critical risks.

## Superseded by

- #2307: lazy Bitcoin engine foundation. Merged.
- #2332: dependency-free transaction and key calculations. Open.
- #2336: registration and package work. Draft and due to be split again.

## Why it still matters

Its tests and review findings informed the smaller pull requests. It is not current acceptance evidence.

Tracks #2231. #2231 stays open.


jonybur (commented):
A few substantive concerns from a fresh pass over the ETH-only extraction:

- packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts (calculatePeginTxHash / serializeWithoutWitness): The parser rejects a tx that has the SegWit marker + flag but all-empty witness stacks with "superfluous witness marker", where bitcoinjs-lib parses it. That is a real behaviour delta versus the WASM `deriveVaultId` / bitcoinjs implementation being replaced on the registration path. The seeded differential corpus in the parity test only writes witnesses when `fixture % 2 === 1` and uses `1 + (next() % 3)` items on the required witness input, so a segwit-marker-with-all-empty-witnesses shape is never generated and the divergence would not fail the guard. Section 9 of CLAUDE.md calls for byte-for-byte parity over golden vectors *plus* randomised inputs - please extend the corpus to cover this shape (empty-witness/segwit-marker, missing-witness-flag, unknown-flag) so the guarantee is real, or match bitcoinjs behaviour and drop the reject.

- packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts (registerPeginBatchOnChain): The batch path normalizes `depositorBtcPubkey` and `btcPopSignature` once at the top of the function and then re-normalizes them inside every `normalizeRequest(...)` call. `normalizeRequest` invokes `assertOnChainBtcPubkey`, whose curve-membership check is the expensive part. For a batch of N, that is N+1 curve verifications instead of 1 + N-1 for the per-vault fields. Either drop the outer normalization or skip the depositor-key / PoP re-normalization inside the per-item call.

- packages/babylon-ts-sdk/src/tbv/core/wasm/__tests__/value-guards.test.ts and wasm/value-guards.ts: The parity claim in the module header ("pinned to identical behaviour with the peer") is verified by feeding vectors to both the local guard and the peer's own `assertPositiveBigintArray`. If the peer's guard drifts, the oracle drifts with it and the test still passes. Please anchor at least one differential vector to a hard-coded expected outcome (or run the check against the Rust reference), so the "byte-for-byte identical" contract has an independent witness.

- packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/eth-import-boundary.test.ts (runtimeClosure): The recursive walk collects STATIC and SIDE_EFFECT specifiers but not DYNAMIC. A future ETH-only file reached solely via `await import('./local-thing')` (or transitively via one) would silently escape this source-level boundary test, even though check-eth-build.js would still catch it at build time. Add DYNAMIC to the recursion set so the source-side gate matches the emitted-side gate.

packages/babylon-ts-sdk/scripts/check-eth-build.js
```diff
@@ -0,0 +1,171 @@
...
+) {
+  throw new Error(
+    `Dependency-clean boundary fixture did not report its full import chain:\n${contaminatedFixture.violations.join("\n")}`,
+  );
+}
+
+// Pass 1: the ETH-only entry points must not reach the Bitcoin or WASM
+// dependencies at all, eagerly or lazily.
+const dependencyCleanEntries = [
+  "dist/tbv/core/clients/eth/index.js",
```
jrwbabylonlab, line 119: The README calls `tbv/core/contracts` dependency-clean for ETH consumers, but no gate covers that subpath. Add it here and to the boundary test, or drop the claim.

packages/babylon-ts-sdk/package.json
```diff
@@ -110,16 +135,28 @@
...
     "viem": "^2.38.2"
   },
+  "peerDependenciesMeta": {
+    "@babylonlabs-io/babylon-tbv-rust-wasm": {
+      "optional": true
+    },
+    "@bitcoin-js/tiny-secp256k1-asmjs": {
+      "optional": true
+    },
+    "bitcoinjs-lib": {
```
gbarkhatov, line 154: **Optional peers that are still eagerly imported.**

Only the engine actually received a lazy boundary. `bitcoinjs-lib` and `@bitcoin-js/tiny-secp256k1-asmjs` are marked optional here but are still statically imported by every root barrel. Walking the static import graph of the built `dist/`:

- `dist/index.js` → 34 modules, reaching `bitcoinjs-lib` through **14** chunks and `@bitcoin-js/tiny-secp256k1-asmjs` through **2**
- identical for `tbv/index.js` and `tbv/core/index.js`

Both were *required* peers before this PR, so package managers installed them and warned when they were absent. Now they do neither. A consumer following this PR's own install line (README.md:59-63, "only viem is required") and then running its Verify Installation snippet (README.md:104, which imports from `/tbv/core/primitives`) gets a bare `ERR_MODULE_NOT_FOUND` — none of the actionable messaging the engine peer got in `wasm/index.ts`. `dist/index.d.ts` breaks TS consumers without `skipLibCheck` for the same reason.

The engine marking is correct and well proven. Suggest keeping the other two required and marking only the engine optional.

One knock-on: `eth-import-boundary.test.ts:136-138` iterates `BANNED_RUNTIME_IMPORTS` — all three — and asserts `optional === true`, so it currently locks this configuration in as an expectation. That assertion needs to narrow to the engine along with the fix.

packages/babylon-ts-sdk/src/tbv/core/vault-secrets/deriveVaultRoot.ts
```diff
@@ -54,10 +54,14 @@ export interface DeriveContextHashCapableWallet {
...
-  return typeof wallet.deriveContextHash === "function"
+  // Captured before the closure so the narrowing survives it. The property
+  // read cannot be narrowed across a closure boundary, which is what forced
+  // the non-null assertion this replaces; forwarding is otherwise unchanged.
+  const deriveContextHash = wallet.deriveContextHash;
+  return typeof deriveContextHash === "function"
     ? {
         deriveContextHash: (appName, context) =>
-          wallet.deriveContextHash!(appName, context),
+          deriveContextHash.call(wallet, appName, context),
```
gbarkhatov, line 64: **Argument pass-through is unguarded on frozen critical path 4.**

The receiver fix is right and the new test pins it. The arguments are a different matter: I replaced `context` here with `undefined` and ran the full suite — **103 files / 1635 tests, all passing.**

Nothing catches it because neither test actually exercises this path:

- the only `forwardDeriveContextHash` test (`deriveVaultRoot.test.ts:42-62`) declares `async deriveContextHash(): Promise<string>` — zero parameters, returning `this.rootHex` no matter what it is called with;
- the wiring tests at :65-91 *do* assert `(VAULT_APP_NAME, expectedContextHex)`, but they hand an object literal straight to `deriveVaultRoot`, which never calls this forwarder.

This is the live production path for class-based providers — `services/vault/src/hooks/deposit/useVaultActions.ts:503` spreads `forwardDeriveContextHash(btcWalletProvider)`, which is exactly the Ledger case that motivated the `.call(wallet)` fix. A dropped `context` means the device derives a root that does not match the on-chain commitment: no activation, refund-only recovery, and it compiles and tests clean.

Give the fake a real two-parameter signature and assert both arguments arrive.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts
```diff
@@ -0,0 +1,440 @@
...
+    assertHtlcVout(params.htlcVout);
+    const unsignedPrePeginTx = ensureHex(
+      params.unsignedPrePeginTx,
+      "unsignedPrePeginTx",
+    );
+    const depositorSignedPeginTx = ensureHex(
+      params.depositorSignedPeginTx,
+      "depositorSignedPeginTx",
+    );
+    const payoutScript = ensureHex(
```
gbarkhatov, line 303: **`depositorPayoutScriptPubKey` is registered without being bound to the PoP key.**

It gets `ensureHex` plus the max-length check below and nothing else. `"0x0014" + <attacker hash160>` is 21 bytes, valid hex, far under `MAX_PAYOUT_SCRIPT_LEN` — accepted here and encoded verbatim into the registration call at :161.

The JSDoc at :82-86 states the requirement ("The BTC preparation boundary must validate that this script belongs to the PoP key"), but a comment is not a boundary. `resolvePayoutScriptPubKey` stays behind in `PeginManager.ts:1347`, so before this PR resolution and submission were the same call and this input did not exist as a seam. `ViemPeginRegistrationClient` is exported from the documented `./tbv/core/clients/eth` subpath, so this is now reachable public API — and what it registers is an irreversible payout sink.

Either brand the type so only the validated resolver can mint it, or re-derive and assert the script against the PoP key inside `normalizeRequest`.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts
```diff
@@ -0,0 +1,440 @@
...
+  publicClient: PublicClient;
+  btcVaultRegistry: Address;
+  receiptTimeoutMs?: number;
+  /**
+   * Refuse to register without a `quotedCommissionBps`. Set when the depositor
+   * approved deposit terms on a device: approval froze the commission ceiling
+   * from the quote, so the chain-current fallback could admit a commission the
+   * device would later refuse to pay out.
+   */
+  requireQuotedCommissionBps?: boolean;
```
gbarkhatov, line 71: **Optional flag whose default is the unsafe direction.**

This was previously derived unconditionally from `supportsDepositApproval` inside `PeginManager` — it still is, at `PeginManager.ts:1325` — and could not be bypassed. As an optional field on public config, omitting it is silently `false`.

The audience this PR creates is the one most likely to omit it: an ETH-only integrator has no BTC wallet to probe for the capability. Registering for a Ledger depositor without it falls through to `capMaxAcceptableCommissionBps(currentBps)` at :389, shipping a ceiling read from chain rather than the one frozen on-device at approval time — and the device then refuses the payout it approved.

Worth making required, or deriving it from something the client can verify rather than trusting the caller to opt in.

packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts
```diff
@@ -0,0 +1,250 @@
...
+export function loadRawTbvWasm(): Promise<RawTbvWasmModule> {
+  rawWasmModulePromise ??= import("@babylonlabs-io/babylon-tbv-rust-wasm/raw")
+    .then(async (wasm) => {
+      await wasm.initWasm();
+      return wasm;
+    })
+    .catch((error: unknown) => {
+      rawWasmModulePromise = undefined;
+      throw new Error(
+        "This operation requires the optional raw vault-WASM peer entry.",
```
gbarkhatov, line 66: **Instantiation failures are reported as a missing dependency.**

The `.catch` is chained after `.then(async (wasm) => { await wasm.initWasm(); ... })`, so it covers initialization as well as resolution. A CDN 404, a CSP block, or a failed `WebAssembly.compile` on the refund path all surface as "requires the optional raw vault-WASM peer entry" — pointing an operator at a package that is already installed. The real cause survives only in `cause`.

`loadTbvWasm` at :39-50 is correct by contrast: its `.catch` sits directly on the `import()`.

`loader.test.ts:96-114` currently encodes the wrong message as the expectation — it mocks `initWasm` throwing `"wasm instantiation failed"` and asserts the `"raw vault-WASM peer entry"` text, so the test locks in the mislabel.

Catching resolution and initialization separately, and claiming a missing dependency only for the former, fixes both.

packages/babylon-ts-sdk/scripts/check-eth-build.js
```diff
@@ -0,0 +1,171 @@
...
+  );
+}
+
+// Pass 2: the claim made by the module JSDoc of src/tbv/core/wasm/index.ts —
+// importing the SDK, including its legacy root barrels, does not resolve the
+// WASM package or generated binary. Only eager specifiers count: the lazy
+// facade's `import(...)` of the peer is the boundary being proven, not a
+// violation of it. Runtime entries only — the facade's .d.ts carries a type
+// import of the peer, and the patterns above match type imports deliberately.
+const lazyWasmEntries = [
```
gbarkhatov, line 149: **Pass 2 never inspects `./tbv/integrations/aave` or `./testing`.**

`src/index.ts` re-exports only `shared` + `tbv`, and `src/tbv/index.ts` only `./core/index.js`, so neither is reachable by transitivity from these four entries.

I reproduced the closure: **70 modules visited, 0 from `integrations/aave`, 0 from `dist/testing`** — matching the "verified across 70 emitted module(s)" this script prints. Both are declared package exports, so a top-level eager engine import in, say, `vaultSplit.ts` would print the success line and exit 0.

Adding `dist/tbv/integrations/aave/index.{js,cjs}` and `dist/testing/index.{js,cjs}` to this list closes it.

packages/babylon-ts-sdk/scripts/check-eth-build.js
```diff
@@ -0,0 +1,171 @@
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, extname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), "..");
+const eagerSpecifierPatterns = [
+  /\b(?:import|export)\s+(?:type\s+)?[^"';]*?\sfrom\s*["']([^"']+)["']/g,
```
gbarkhatov, line 7: **The gate is weaker than it reads, in two compounding ways.**

1. This pattern requires whitespace after `import` and before `from`, so a minified named import — `import{a as b}from"bitcoinjs-lib"` — is missed entirely. Vite output is unminified today, so the gate currently works by luck; flipping `build.minify` would collapse the closure to a handful of modules, find nothing, and report success while the boundary is actually broken.

2. The self-test fixtures (`scripts/fixtures/eth-import-boundary/{entry,intermediate}.js`) both use bare side-effect `import "…"`, which exercises only pattern [1]. The forms the emitted output actually produces — [0] for ESM, [2] for CJS, and the dynamic pattern for the lazy boundary itself — are never self-tested. And `visited.size` is printed at :166-171 but never asserted, so a collapsed closure still passes.

A fixture per pattern, plus a floor assertion on `visited.size`, would make the gate fail when it stops seeing the graph.

packages/babylon-tbv-rust-wasm/src/raw-node.ts
```diff
@@ -0,0 +1,30 @@
+/** Explicit eager Node.js entry for direct wasm-bindgen class access. */
+import { readFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+// prettier-ignore
+// @ts-expect-error - generated artifacts live in dist/generated
+import { initSync, WasmPeginTx, WasmPeginPayoutConnector, WasmPrePeginTx, WasmPrePeginHtlcConnector, WasmAssertPayoutNoPayoutConnector, WasmAssertChallengeAssertConnector } from './generated/vault_wasm.js';
+
+let initialized = false;
```
gbarkhatov, line 10: **Node keeps a private init flag while the browser side was unified.**

`raw.ts` was fixed to share one initializer with `wasm-loader.ts` — the race this PR closes. The Node side keeps this flag and its own `initSync`, independent of `wasm-loader-node.ts`'s cached promise.

Not a live bug: the pinned glue guards both `initSync` and `__wbg_init` with `if (wasm !== undefined) return wasm`, so the second call early-returns. Two things still worth tightening:

- `check-lazy-entries.js:135-211` has a browser arm for exactly this race and no Node equivalent — :213-255 uses `raw-node.js` only to bootstrap WASM for the browser-facade concurrency check. A `VAULT_WASM_COMMIT` bump making `initSync` non-idempotent would silently reopen on Node what this PR just closed on the browser.
- `readFileSync` at :19 is evaluated before `initSync` can early-return, so when both entries are used in one process — `primitives/psbt/refund.ts` reaches both — the 3.1 MB binary is read from disk twice.

Delegating to `wasm-loader-node.ts` removes the dependency on glue internals and both symptoms.

Separately, worth aligning: `raw.d.ts` exports `initWasm(): Promise<void>` while `raw-node.d.ts` exports `(): void`, so `initWasm().then(…)` compiles under the bundler condition and crashes under Node.

packages/babylon-tbv-rust-wasm/src/wasm-loader.ts
```diff
@@ -0,0 +1,27 @@
...
+
+async function importGeneratedBindings() {
+  // Generated artifacts are committed under dist/generated and deliberately
+  // absent from src. The build includes their declarations separately.
+  // @ts-expect-error - generated during the pinned Rust/WASM build
+  return import('./generated/vault_wasm.js');
+}
+
+/** Load and initialize browser WASM only when a facade operation is invoked. */
+export function getWasmBindings() {
```
gbarkhatov, line 11: **Returning `Promise<any>` gives up the link-time export check.**

`importGeneratedBindings` is `@ts-expect-error`'d, so its return type is `any` and `getWasmBindings()` is `Promise<any>`; every call site in this package destructures off that.

The *types* were already suppressed before this change, so that part is not new. What is new is the runtime failure mode. A static `import { WasmPeginTx } from './generated/vault_wasm.js'` fails at module instantiation with a SyntaxError if the export is renamed. Destructuring a namespace object yields `undefined` and defers the failure to `new undefined(...)` at first use.

Concretely: a `VAULT_WASM_COMMIT` bump that renames an export lets the app boot, connect, build and fund, and fail only at payout pre-signing — after commitment. That is a fail-fast regression rather than a wrong-value bug, but it lands on the worst possible step.

Declaring an explicit interface for the bindings surface and typing the loader's return as that restores the check without reintroducing eager glue loading. Worth noting the new `no-explicit-any` rule gives false assurance here, since the `any` is inferred rather than written.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/pegin-transaction.test.ts
```diff
@@ -0,0 +1,221 @@
...
+import { calculateBtcTxHash } from "../../../utils/transaction/btcTxHash";
+import { calculatePeginTxHash, derivePeginVaultId } from "../pegin-transaction";
+
+const GOLDEN_PEGIN_TX_HASH: Hex = `0x${"ab".repeat(32)}`;
+const GOLDEN_DEPOSITOR: Address = "0x1234567890abcdef1234567890abcdef12345678";
+const GOLDEN_VAULT_ID: Hex =
+  "0xf8d22e64c72a84a3dacdedb7d8b42e285bf06bd25850da911398c51d5a6c2dba";
+const TXID_DIFFERENTIAL_SEED = 0xa90f72ec;
+const TXID_DIFFERENTIAL_FIXTURE_COUNT = 64;
+const RANDOM_SCRIPT_LENGTHS = [0, 1, 20, 75, 252, 253, 300] as const;
```
gbarkhatov, line 17: **The differential corpus never reaches the multi-byte CompactSize *count* branches.**

`RANDOM_SCRIPT_LENGTHS` tops out at 300, and input/output counts are `1 + (next() % 5)` at :81-82. So `inputCount.encoded` / `outputCount.encoded` (`pegin-transaction.ts:151-158`) are only ever re-emitted as single bytes. The 0xfd/0xfe/0xff re-emission path — which is where the non-canonical guards live — appears only as rejection cases, never as a match against bitcoinjs.

I fuzzed those branches directly: 252/253/254/255/256/300/1000/65535/65536 inputs and outputs, with and without witnesses, plus scripts up to 70 KB, all differentially against `calculateBtcTxHash`. 28 cases, **zero divergence** — so this is hardening, not a fix.

Raising the count bound in the generator would close the last gap in §9's "goldens **plus** randomised inputs", on a file where that rule is the whole reason the reimplementation is allowed to land.

.github/workflows/claude-md-drift.yml
```diff
@@ -44,6 +44,13 @@ jobs:
             "packages/babylon-ledger-vault-signer/src/"
             "packages/babylon-ledger-vault-signer/scripts/"
             "packages/babylon-ts-sdk/src/tbv/core/clients/eth/onChainBtcPubkey.ts"
+            "packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts"
+            "packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts"
+            "packages/babylon-ts-sdk/src/tbv/core/wasm/"
+            "packages/babylon-tbv-rust-wasm/src/wasm-loader.ts"
+            "packages/babylon-tbv-rust-wasm/src/wasm-loader-node.ts"
+            "packages/babylon-tbv-rust-wasm/src/raw.ts"
+            "packages/babylon-tbv-rust-wasm/src/raw-node.ts"
```
gbarkhatov, line 53: **The prose half of this union did not move with the promotion.**

Promoting these seven from `pending` to `paths` is right. But the comment at :56 describes CLAUDE.md, CODEOWNERS and this workflow as one set that must stay in sync, and `CLAUDE.md` is not in this diff. §9 there still opens "These paths are registered ahead of the reimplementations landing (see #2228 / #2229)" and still says "Files (all but the last arrive with the optional-BTC work)" — they arrive here. Worth updating in the same change, since §9's differential rule is precisely what reviewers will be told to apply to this PR.

Related, and more substantive: this PR adds `@babylonlabs-io/babylon-tbv-rust-wasm/raw`, a documented and exported second crossing that hands out raw wasm-bindgen classes with no value guards — consumed by the SDK itself at `primitives/psbt/refund.ts:76`. §1 requires every WASM output consumed by JS to be asserted against expected bounds before use. Worth stating explicitly how `/raw` stands against that rule rather than leaving it to be inferred.

gbarkhatov (approved):
Reviewed the full diff (~3.6k added lines across the engine package, the SDK, and build tooling). Rather than re-running the PR's own gates and calling it verified, I mutated the code those gates guard and checked whether anything failed.

## Verified correct

- **The lazy boundary genuinely holds.** Walking the static import graph of the built `dist/`: `tbv/core/clients/eth/index.js` reaches 11 modules and `viem` alone; `tbv/core/wasm/index.js` reaches 1. The engine appears in no entry's eager closure. `check-lazy-entries.js` is a real adversarial gate — it copies `dist/`, deletes `generated/`, and proves at runtime that roots import clean and fail only on first facade call.
- **Txid parser matches bitcoinjs.** `serializeWithoutWitness` reproduces `Transaction.fromHex().getId()` including the segwit marker rule, the superfluous-witness rejection and the trailing-byte rejection, and is strictly stricter on non-canonical CompactSize — it fails closed. The differential test against both bitcoinjs and the WASM `deriveVaultId` oracle, with a seeded RNG, is what §9 asks for.
- **Curve check is sound.** Tonelli–Shanks for p ≡ 3 (mod 4), correct exponent; x=0 fails and x³+7 ≡ 0 has no solution (no 2-torsion on secp256k1), so the zero-hash guarantee holds.
- **`derivePeginVaultId`** — viem lowercases the address and `isAddress` accepts all-lowercase, so there is no checksum trap.
- **Derivation bytes did not change.** `context.ts` is untouched; `dist/generated/` and `VAULT_WASM_COMMIT` are byte-identical to the merge base. §4's hard-fork gate is not triggered and no existing deposit is invalidated.
- **Double-init is not a live bug.** Both `initSync` and `__wbg_init` in the pinned glue guard on `wasm !== undefined`, so the raw + facade paths cannot produce two instances today.

## Checked and dismissed

- **The removed connector cache is not a performance problem.** I measured it rather than reasoning about complexity: construction is 0.44 ms at 32 challengers and 0.63 ms at 50, so per-call ownership costs roughly 15 ms across a 32-challenger signing flow. Trading that for the removal of a shared mutable connector that could be freed out from under a concurrent reader is clearly the right call. No change wanted.
- **`loader.test.ts` "imports once and shares the module" does have teeth.** I replaced `??=` with an unconditional assignment; the test failed on `expect(first).toBe(second)`. (`expect(attempt).toBe(1)` is vacuous, since vitest caches the mock factory, but the identity assertion carries the test.)
- **Multi-byte CompactSize counts agree.** I fuzzed the branches the committed corpus never reaches — 252/253/254/255/256/300/1000/65535/65536 inputs and outputs, with and without witnesses, plus scripts up to 70 KB — against bitcoinjs. 28 cases, zero divergence. The related inline comment is about coverage, not a defect.

## Findings

Eleven inline. The two I would not merge without are:

1. **`peerDependenciesMeta`** — `bitcoinjs-lib` and `@bitcoin-js/tiny-secp256k1-asmjs` are now optional, but `dist/index.js` still reaches them through 14 and 2 chunks. A consumer following this PR's own README install line gets `ERR_MODULE_NOT_FOUND`. This is a regression introduced here; both were required peers before.
2. **`forwardDeriveContextHash` argument pass-through** — on frozen critical path 4. I replaced `context` with `undefined` and all 103 files / 1635 tests passed. The receiver fix in this PR is right and is tested; the arguments are not.

The rest are build-gate coverage gaps, error-message accuracy, and two public-API seams on the new ETH-only client that are weaker than the `PeginManager` path they extract from.

Approving: every inline item is fixable in place, none of them touch derivation bytes, and the headline claim of the PR holds up under adversarial checking.

**Process note.** This lands critical paths 4 and 9, so the two-reviewer requirement and §9's differential rule apply regardless of these comments.

jonybur:
Superseded. Closing in favour of three smaller PRs against `main`.

This branch was too large to review at 67 files, and the review it did get surfaced work that
changed the shape of the change. Rather than push another round on top, the reviewed content was
re-cut by content into three parts:

| | PR | files | diff |
|---|---|---|---|
| A | #2307 | 48 | +1526 / -399 |
| B | `split/b-pegin-identifiers` (opens after A merges) | 5 | +635 / -3 |
| C | `split/c-eth-client-optional-peer` (opens after B merges) | 31 | +2033 / -501 |

They go to `main` one at a time. `verify.yml` and `critical-path-check.yml` both declare
`on: pull_request: branches: [main]`, so a PR based on another split branch would get no build, no
tests and no critical-path reviewer prompt at all.

Applying all three produces a tree byte-identical to this branch plus the fixes below, so nothing
was dropped in the cut.

**Do not merge this branch.** The head here (`983b76cf5`) predates the review fixes and its CI is
red. The fixes live in A, B and C:

- `onChainBtcPubkey.test.ts` timed out in CI (6127 ms against a 5000 ms default) — the run on this
  branch was never green.
- `bitcoinjs-lib` and `@bitcoin-js/tiny-secp256k1-asmjs` were marked optional peers while still
  statically imported by every root barrel, so a consumer following the README got a bare
  `ERR_MODULE_NOT_FOUND`. Only the engine got a lazy boundary, so only the engine stays optional.
- `forwardDeriveContextHash` dropped its `context` argument with no test catching it — the full
  suite passed with the argument replaced by `undefined`. On frozen critical path 4.
- Plus the raw/facade Node initializer, the boundary-gate coverage gaps, the payout-script binding
  and the required commission flag.

Review history stays here. Thanks @gbarkhatov, @jrwbabylonlab and the bots — the two findings you
would not merge without are both fixed and pinned by tests that fail when the fix is removed.

PR #2296, opened 2026-08-21
fix(vault): stop the collateral stat column collapsing under its label

jeremy-babylonlabs:
## What

On the Overview Position panel, the "Total collateral value" label wrapped onto three lines with the info icon floating mid-wrap, because the first column got squeezed.

## Root cause

The columns used `flex-1` (shrink allowed) with unconstrained text. With wrapping allowed, a flex item's automatic minimum width collapses to its narrowest word, so the column with the longest label plus the info icon was the one that gave way. The design (Figma `10094-26791`, Stats instance) specifies `flex: 1 0 0` on all three columns and `whitespace-nowrap` on every label, value, and caption.

## Change

`PositionStatCards.tsx`: `flex-1` → `flex-[1_0_0]` on the column container, and `whitespace-nowrap` on the label row, value, and caption — matching the frame exactly. One file, no behaviour change.

### DIFF
--- a/services/vault/src/components/simple/PositionStatCards.tsx
+++ b/services/vault/src/components/simple/PositionStatCards.tsx
@@ -42,7 +42,7 @@ function StatMeter({
         aria-valuemin={0}
         aria-valuemax={100}
         aria-valuenow={Math.round(clamped * 100)}
-        className="h-1 w-[68px] overflow-hidden rounded-full bg-secondary-strokeLight"
+        className="h-1 w-[68px] overflow-hidden rounded-full bg-secondary-strokeLight xl:max-[1439px]:w-[48px]"
       >
         <div
           className="h-full rounded-full bg-secondary-main"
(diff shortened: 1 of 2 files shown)


jonybur (dismissed):
Correct fix. Switching flex-1 to flex-[1_0_0] prevents the automatic minimum from collapsing to the narrowest word, and whitespace-nowrap on the label/value/caption matches the Figma stats frame.

jonybur (approved):
Layout fix is correct and tightly scoped. Switching from flex-1 to flex-[1_0_0], guarding whitespace-nowrap behind xl:, and compressing gaps / action-button / meter widths in the 1280-1439px band avoids the earlier regression where shrink-0 with unconditional nowrap forced the row wider than its content and clipped the Repay button. Tests cover both the xl three-column single-line assertion and the below-xl wrap regression guard.

PR #2297, opened 2026-08-21
refactor(vault): reorganize the god-mode panel into surface tabs

jeremy-babylonlabs:
## What

The god-mode panel was one 848-line component holding every debug section inline, with three more panels bolted on as flat children — no mapping from a control to the app surface it drives, and two different UIs writing the same cascade override store.

This reorganizes `services/vault/src/dev/` around the app's own surfaces:

- **Thin shell** (278 lines): chrome (drag, pop-out), a tab rail, and a pinned override-summary chip. Tabs declared in a small `registry.ts` (id, label, component, flag gate).
- **Tabs**: Global (theme, protocol status, flag readout) · Position · Liquidations · Deposit & Vaults (mocks, receipts, artifact demo) · Loans · Markets — one file per tab under `dev/panels/`.
- **CascadeSimulator** replaces "manual mode": a Live | Simulated switch with one-click presets (incl. two new ones verified against the real `calculate()` — near-liquidation ~2% and fully-liquidated), stale-price as a toggle, and the freeform editor behind a Custom disclosure. Shared by the Position and Liquidations tabs — the duplicated cascade UIs collapse into one.
- **Summary chip**: lists every active override across all stores with per-item clear and Reset-all — overrides can no longer silently poison pages.

## Guarantees

- Zero production code changes: `src/overrides/*` APIs and `featureFlags.ts` untouched; the `import.meta.env.DEV` boundary is unchanged, so the whole subtree still drops from production builds.
- Behaviour parity: every existing capability relocated, none removed. All tabs stay mounted (hidden) so override-publish lifecycles are identical to the old flat body.
- Tests: 105/105, including new coverage for the registry, flag-gated tabs, preset writes, and reset-all.


jonybur (approved):
Dev-only refactor with clean surface tabs. Publisher effect is centralized (single mount) and cascade override is properly cleared on unmount. Guarantees of behaviour parity and full test coverage check out. LGTM.

PR #2298, opened 2026-08-23
feat(ledger-vault, vault): typed device errors, signing cancel, no ve…

jrwbabylonlab:
Ledger failures surfaced as raw text or generic fallbacks, a started
ceremony could not be backed out of, and the envelope's tx-graph pin would
have blocked v3 deposits the device signs fine. Errors now carry typed codes
mapped to real copy, cancelSigning() backs out safely (retry re-runs the
full ceremony), and version gating is left to the WASM capability gate.

### DIFF
--- a/packages/babylon-ledger-vault-signer/src/envelope.ts
+++ b/packages/babylon-ledger-vault-signer/src/envelope.ts
@@ -17,7 +17,6 @@ import {
   DEVICE_PEGIN_AMOUNT_DUST_MULTIPLE,
   DEVICE_PEGIN_CSV_TIMELOCK_MAX_BLOCKS,
   DEVICE_REFUND_TIMELOCK_MAX_BLOCKS,
-  DEVICE_SUPPORTED_VAULT_CORE_VERSION,
   DEVICE_TIMELOCK_MIN_BLOCKS,
   DEVICE_VAULT_DUST_LIMIT_SATS,
 } from "./deviceCaps";
@@ -37,16 +36,8 @@ function requireIntInRange(field: string, value: number, lo: number, hi: number)
  * @throws {DepositTermsRejectedError} on the first violation
  */
 export function assertDepositTermsDeviceCompatible(terms: DepositTerms): void {
-  // The firmware hardcodes one PSBT shape and the intent TLV carries no
-  // version field, so ANY mismatched graph (v1 or a future v3) loads fine and
-  // only fails at PSBT time — after the user has physically approved.
-  if (terms.vaultCoreVersion !== DEVICE_SUPPORTED_VAULT_CORE_VERSION) {
-    throw new DepositTermsRejectedError(
-      `${RANGE_MSG}: vaultCoreVersion ${terms.vaultCoreVersion} is not ` +
-        `${DEVICE_SUPPORTED_VAULT_CORE_VERSION}; this device only supports ` +
-        `tx-graph v${DEVICE_SUPPORTED_VAULT_CORE_VERSION} deposits.`,
-    );
-  }
+  // No tx-graph version check (2026-08-23 decision): the WASM capability gate
+  // stops unconstructable versions upstream; bad shapes fail closed on-device.
 
   // The >= 1 floor is enforced by both the contract and the device
   // (vault_tlv.c:73 rejects rate == 0).
--- a/packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
+++ b/packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
@@ -9,8 +9,8 @@ import {
   createDmkRawApduSender,
   DepositTermsRejectedError,
   deriveChangeXOnlyHex,
-  deriveReceiveXOnlyHex,
   deriveContextHash,
+  deriveReceiveXOnlyHex,
   disconnectDmkSession,
   encodeIntentGroup,
   encodeIntentScalars,
@@ -31,6 +31,7 @@ import {
   signPreparedVaultPsbt,
   SW_BAD_STATE,
   SW_CAP_EXCEEDED,
+  SW_CLA_NOT_SUPPORTED,
   type ApduSender,
   type DefaultTaprootWalletPolicy,
   type DepositTerms,
@@ -167,8 +168,6 @@ export class LedgerVaultProvider implements IBTCProvider {
   private rawSend: RawApduSender | undefined;
   /** Request-identity keys ({@link signingRequestKey}) signed under the CURRENT loaded intent. */
   private signedFingerprints = new Set<string>();
-  /** A host abandonment interrupted the dispatcher — the next sign resends once on 0x6A80. */
-  private loopAbandoned = false;
   /**
    * ONE in-flight device ceremony (derive/approve/sign) at a time — a
    * concurrent APDU would be eaten with 0x6A80 and desync the interrupt loop.
@@ -332,7 +331,6 @@ export class LedgerVaultProvider implements IBTCProvider {
     this.pubkeyHexPromise = undefined;
     this.policyContextPromise = undefined;
     this.signedFingerprints = new Set();
-    this.loopAbandoned = false;
     // Release the ceremony lock and stop an in-flight signing loop NOW — the
     // stale call's finally only releases its own token, and its rejection
     // commits nothing (the generation just changed).
@@ -383,11 +381,7 @@ export class LedgerVaultProvider implements IBTCProvider {
         // but only at SIGN_PSBT — by then approveDepositTerms has already spent
         // the intent ceremony. This guards a host-side desync (depositorPath vs
         // accountPath, coin type, a refactor of either getter), not a device fault.
-        const derivedXOnlyHex = deriveReceiveXOnlyHex(
-          accountXpub,
-          toNetwork(this.network).bip32,
-          ADDRESS_INDEX,
-        );
+        const derivedXOnlyHex = deriveReceiveXOnlyHex(accountXpub, toNetwork(this.network).bip32, ADDRESS_INDEX);
         if (derivedXOnlyHex !== depositorXOnlyHex) {
           throw new WalletError({
             code: ERROR_CODES.CONNECTION_FAILED,
@@ -472,7 +466,6 @@ export class LedgerVaultProvider implements IBTCProvider {
     // old intent's dedup state dies with it — clear the sign bookkeeping too.
     this.deviceState = { phase: "idle" };
     this.signedFingerprints = new Set();
-    this.loopAbandoned = false;
 
     const generation = this.connectionGeneration;
     const root = await deriveContextHash(this.requireSender(), {
@@ -599,7 +592,7 @@ export class LedgerVaultProvider implements IBTCProvider {
     if (this.deviceState.phase === "intent-loaded") {
       if (this.deviceState.termsKey === key) return;
       throw new WalletError({
-        code: ERROR_CODES.WALLET_NOT_CONNECTED,
+        code: ERROR_CODES.DEVICE_CEREMONY_INVALID,
         message:
           `${WALLET_PROVIDER_NAME} already holds a different approved intent ` +
           `on this connection — the device admits one ceremony per ` +
@@ -611,7 +604,7 @@ export class LedgerVaultProvider implements IBTCProvider {
     // the device) would die at the first APDU with SW_BAD_STATE.
     if (this.deviceState.phase === "idle") {
       throw new WalletError({
-        code: ERROR_CODES.WALLET_NOT_CONNECTED,
+        code: ERROR_CODES.DEVICE_CEREMONY_INVALID,
         message:
           `${WALLET_PROVIDER_NAME} has no freshly derived context root on ` +
           `this connection — the device requires DERIVE_CONTEXT_HASH before ` +
@@ -630,7 +623,6 @@ export class LedgerVaultProvider implements IBTCProvider {
     // fresh, so the same PSBT bytes are legitimately signable again. (The
     // byte-equal re-approval no-op above returns earlier and keeps both.)
     this.signedFingerprints = new Set();
-    this.loopAbandoned = false;
   };
 
   /**
@@ -701,12 +693,19 @@ export class LedgerVaultProvider implements IBTCProvider {
         }
         const signed: string[] = [];
         for (const one of staged) {
-          // A cancel landing BETWEEN elements must stop the batch — inside an
-          // element the loop's own signal checks cover it.
+          // Abort between elements: same generation = user's cancelSigning,
+          // changed = teardown/disconnect. In-element aborts settle in the loop.
           if (controller.signal.aborted) {
+            const cancelled = ctx.generation === this.connectionGeneration;
+            if (cancelled) {
+              this.deviceState = { phase: "idle" };
+              this.signedFingerprints = new Set();
+            }
             throw new WalletError({
-              code: ERROR_CODES.WALLET_NOT_CONNECTED,
-              message: `${WALLET_PROVIDER_NAME} stopped signing (disconnected) after ${signed.length} of ${psbtsHexes.length} PSBT(s).`,
+              code: cancelled ? ERROR_CODES.CONNECTION_REJECTED : ERROR_CODES.WALLET_NOT_CONNECTED,
+              message: cancelled
+                ? `Signing cancelled after ${signed.length} of ${psbtsHexes.length} PSBT(s) — the ceremony restarts from the device approval screens on retry.`
+                : `${WALLET_PROVIDER_NAME} stopped signing (disconnected) after ${signed.length} of ${psbtsHexes.length} PSBT(s).`,
               wallet: WALLET_PROVIDER_NAME,
             });
           }
@@ -716,6 +715,16 @@ export class LedgerVaultProvider implements IBTCProvider {
         return signed;
       }),
     );
+
+  /**
+   * User cancel of the in-flight ceremony: aborts WITHOUT teardown, settling
+   * as CONNECTION_REJECTED at the next exchange boundary — possibly only after
+   * the user acts on the device, so callers hold a "cancel requested" state
+   * until the sign promise settles. No-op when idle.
+   */
+  cancelSigning = (): void => {
+    this.signAbortController?.abort();
+  };
 
   /**
    * ONE AbortController per public sign call (plan D1) — it spans a whole
@@ -754,7 +763,7 @@ export class LedgerVaultProvider implements IBTCProvider {
     this.assertSameConnection(generation);
     if (requireIntent && this.deviceState.phase !== "intent-loaded") {
       throw new WalletError({
-        code: ERROR_CODES.WALLET_NOT_CONNECTED,
+        code: ERROR_CODES.DEVICE_CEREMONY_INVALID,
         message:
           `${WALLET_PROVIDER_NAME} holds no approved intent on this ` +
           `connection — signing requires DERIVE_CONTEXT_HASH and an approved ` +
@@ -877,25 +886,21 @@ export class LedgerVaultProvider implements IBTCProvider {
         signal: controller.signal,
         appIdentity:
           session.appName !== undefined ? { appName: session.appName, appVersion: session.appVersion } : undefined,
-        resendOnceOnIncorrectData: this.loopAbandoned,
       });
       this.assertSameConnection(generation);
       // Success never consumes the intent: further (different) PSBTs sign
       // under the same approval; this one never again.
       this.signedFingerprints.add(fingerprintKey);
-      this.loopAbandoned = false;
       return result.signedPsbtHex;
     } catch (error) {
       const walletError = this.classifySignFailure(error, generation, label);
-      // Mirror reset is signPsbt-only (a PoP failure never invalidates the
-      // device's vault context). Skip it for the two classifications that
-      // commit nothing: a stale rejection and an abandonment.
+      // Mirror reset is signPsbt-only (PoP failures don't invalidate the vault
+      // context); stale rejections and cancels were already handled in classify.
       if (generation === this.connectionGeneration && !isLedgerSignPsbtAbortedError(error)) {
         // Pessimistically assume the device dropped the intent (error-path
         // invalidation is mixed in firmware — never assume survival).
         this.deviceState = { phase: "idle" };
         this.signedFingerprints = new Set();
-        this.loopAbandoned = false;
       }
       throw walletError;
     }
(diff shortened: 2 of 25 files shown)


services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
```diff
@@ -705,6 +705,220 @@ describe("usePayoutSigningState", () => {
...
+      expect(pending.signal?.aborted).toBe(true);
+      expect(result.current.cancelRequested).toBe(true);
+      // The cancel is a request: the sign has not settled yet.
+      expect(result.current.signing).toBe(true);
+
+      await act(async () => {
+        pending.reject(signingCancelledError());
+        await signPromise;
+      });
+    });
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/__tests__/useDepositFlow.test.tsx
```diff
@@ -1511,6 +1511,329 @@ describe("useDepositFlow", () => {
...
+      );
+      const { wallet, settle } = pendingSignWallet(true);
+      vi.mocked(preparePeginTransaction).mockImplementation(async (w) => {
+        await w.signPsbt("psbt0", {});
+        return MOCK_BATCH_RESULT as any;
+      });
+
+      const { result } = renderHook(() =>
+        useDepositFlow({ ...MOCK_PARAMS, btcWalletProvider: wallet as any }),
+      );
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — matches this file's established mock pattern

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1419,6 +1468,24 @@ export function useDepositFlow(
...
+    // Device-only cancel (aborting the flow controller reads as a closed modal).
+    // Settles like an on-device reject: pre-broadcast = "Signing rejected"
+    // callout; post-broadcast payout stage = per-vault warning, loop continues.
+    provider.cancelSigning();
+  }, [btcWalletProvider]);
+
+  const canCancelDeviceSign =
+    deviceSignActive &&
+    typeof (btcWalletProvider as { cancelSigning?: unknown } | null)
+      ?.cancelSigning === "function";
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed the provider binding; not extracting the typeof check — language idiom, used as-is across the repo

services/vault/src/utils/errors/__tests__/formatting.test.ts
```diff
@@ -890,6 +890,88 @@ describe("Error Formatting", () => {
...
+      ).toEqual(COPY.deposit.payoutSignatureErrors.deviceLocked);
+      expect(
+        formatPayoutSignatureError(
+          new FakeWalletError(
+            "DEVICE_WRONG_APP",
+            "The running app does not handle vault instructions — open the Babylon Vault app",
+          ),
+        ),
+      ).toEqual(COPY.deposit.payoutSignatureErrors.deviceWrongApp);
+    });
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

services/vault/src/utils/errors/formatting.ts
```diff
@@ -14,6 +14,12 @@ import {
 import { COPY } from "@/copy";
 
 import { isDepositorWalletMismatchError } from "./depositorWalletMismatch";
+import {
+  deviceErrorCodeOfFrame,
+  isDeviceCeremonyInvalidError,
+  isDeviceLockedError,
+  isDeviceWrongAppError,
+} from "./deviceErrors";
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/envelope.ts
```diff
@@ -37,16 +36,8 @@ function requireIntInRange(field: string, value: number, lo: number, hi: number)
...
-  // only fails at PSBT time — after the user has physically approved.
-  if (terms.vaultCoreVersion !== DEVICE_SUPPORTED_VAULT_CORE_VERSION) {
-    throw new DepositTermsRejectedError(
-      `${RANGE_MSG}: vaultCoreVersion ${terms.vaultCoreVersion} is not ` +
-        `${DEVICE_SUPPORTED_VAULT_CORE_VERSION}; this device only supports ` +
-        `tx-graph v${DEVICE_SUPPORTED_VAULT_CORE_VERSION} deposits.`,
-    );
-  }
+  // No tx-graph version check (2026-08-23 decision): the WASM capability gate
+  // stops unconstructable versions upstream; bad shapes fail closed on-device.
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — deliberate removal; the WASM capability gate rejects unbuildable versions upstream

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1419,6 +1475,28 @@ export function useDepositFlow(
...
+    const provider = deviceSignProviderRef.current as {
+      cancelSigning?: () => void;
+    } | null;
+    if (!deviceSignActiveRef.current) return;
+    if (typeof provider?.cancelSigning !== "function") return;
+    setDeviceCancelRequested(true);
+    // Device-only cancel (aborting the flow controller reads as a closed modal).
+    // Settles like an on-device reject: pre-broadcast = "Signing rejected"
+    // callout; post-broadcast payout stage = per-vault warning, loop continues.
+    provider.cancelSigning();
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — cancelSigning is idempotent; a second dispatch is a no-op

packages/babylon-ledger-vault-signer/src/envelope.ts
```diff
@@ -37,16 +36,8 @@ function requireIntInRange(field: string, value: number, lo: number, hi: number)
...
-  // version field, so ANY mismatched graph (v1 or a future v3) loads fine and
-  // only fails at PSBT time — after the user has physically approved.
-  if (terms.vaultCoreVersion !== DEVICE_SUPPORTED_VAULT_CORE_VERSION) {
-    throw new DepositTermsRejectedError(
-      `${RANGE_MSG}: vaultCoreVersion ${terms.vaultCoreVersion} is not ` +
-        `${DEVICE_SUPPORTED_VAULT_CORE_VERSION}; this device only supports ` +
-        `tx-graph v${DEVICE_SUPPORTED_VAULT_CORE_VERSION} deposits.`,
-    );
-  }
+  // No tx-graph version check (2026-08-23 decision): the WASM capability gate
```
gbarkhatov, line 39: **The justification for removing the gate doesn't hold as written.**

This comment says "the WASM capability gate stops unconstructable versions upstream", but that gate admits `[1, 2, 3]` — `supportedTxGraphVersions()`, pinned in `vaultCoreVersion.test.ts:20` — and its only consumer is `assertVaultCoreVersionSupported` (`vaultCoreVersionSupport.ts:34-41`). Nothing on the Ledger path narrows it further, so the gate replacing the exact-match check is strictly wider on the one axis `DEVICE_SUPPORTED_VAULT_CORE_VERSION` was guarding.

The second half — "bad shapes fail closed on-device" — is exactly what the deleted comment denied. The intent TLV still carries no graph-version field: `intentTlv.ts:155-156` pins `TAG_STRUCTURE_TYPE`/`TAG_VERSION` to the protocol constants `0x01`, and `terms.vaultCoreVersion` is never an encoder input. A mismatched graph still loads, still renders the approval screens, and still fails at SIGN_PSBT after the depositor has physically approved.

The live risk isn't v1 (unlikely any Ledger depositor holds one) — it's the moment the contract's `activeVaultCoreVersion` moves to 3. Every fresh Ledger deposit then burns a ceremony, and resume is worse: `payoutSigning.ts:83` gates on the vault's stamped version and `runDepositorPresignFlow` approves terms rebuilt from chain state before any WASM construction.

If the firmware genuinely completes v3 under the same PSBT shape, cite it — this package's rules require TLV/APDU claims to carry a firmware source — and widen the gate to the supported set. As it stands `envelope.test.ts:45-51` pins the absence of protection without pinning anything in its place.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1419,6 +1475,28 @@ export function useDepositFlow(
...
+    // prop, which a mid-prompt wallet swap can replace.
+    const provider = deviceSignProviderRef.current as {
+      cancelSigning?: () => void;
+    } | null;
+    if (!deviceSignActiveRef.current) return;
+    if (typeof provider?.cancelSigning !== "function") return;
+    setDeviceCancelRequested(true);
+    // Device-only cancel (aborting the flow controller reads as a closed modal).
+    // Settles like an on-device reject: pre-broadcast = "Signing rejected"
+    // callout; post-broadcast payout stage = per-vault warning, loop continues.
```
gbarkhatov, line 1489: **"loop continues" is accurate, and that's the bug: a multi-vault deposit re-prompts the device right after the user cancels.**

`cancelDeviceSign` deliberately doesn't touch the flow controller, so `signal.aborted` is false and the `if (signal.aborted) throw error` at :1345 never fires. The failure falls through to `recordWarning({ terminal: false })` and "Continue with other vaults" at :1385. The next iteration calls `signAndSubmitPayouts`, which calls `ensureAuthenticatedVpClient` with `requireFreshDeviceCeremony: true` (`payoutSigning.ts:103`) — a full derive + approve, so the Ledger lights up with approval screens again seconds after the user asked to stop.

`deviceCancelRequested` was already cleared by `runCancellableSign`'s `finally`, so nothing carries the intent forward. A sticky "user cancelled" flag checked at the top of the vault loop would stop it.

Single-vault takes the same path and returns a successful result carrying only a warning.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -920,15 +925,14 @@ export class LedgerVaultProvider implements IBTCProvider {
       );
     }
     if (isLedgerSignPsbtAbortedError(error)) {
-      // The intent MAY survive an abandonment (hint only) — keep the mirror.
-      // Arm the one-shot 0x6A80 recovery only if a dispatcher is actually
-      // mid-interruption. NB: unreachable while teardown (the only abort
-      // source) bumps the generation first; goes live with #2110's cancel.
-      if (error.dispatcherInterrupted) this.loopAbandoned = true;
+      // Same-generation abort = user's cancelSigning; caps commit pre-yield, so
+      // idle + full re-ceremony (device aftermath: LedgerSignPsbtAbortedError doc).
```
gbarkhatov, line 929: **The 0x6A80 recovery is deleted in the very change that makes a host abort user-reachable — and the abort's documented aftermath is what it existed to absorb.**

`errors.ts:175-179`: within ~5s the device sits blocked awaiting CONTINUE, "the next non-CONTINUE APDU is eaten with 0x6A80 ... a direct retry uses `resendOnceOnIncorrectData`". After this cancel the mirror is reset to idle, so the retry starts with DERIVE_CONTEXT_HASH on a sender with no absorb and no resend — that first APDU is the one that gets eaten, surfacing as "The device rejected the data as invalid" (`rawApdu.ts:72`), which tells the user nothing. Past ~5s `errors.ts:184-186` says the app has exited to the dashboard, so the real instruction is "reopen the app" — which neither new copy string offers.

The provider tests mock the device, so they pass either way.

Either keep the recovery wired into the post-cancel path, or finish the removal: `resendOnceOnIncorrectData` (`signPsbtLoop.ts:78,102`, `signPsbt.ts:53`) and `dispatcherInterrupted` (`errors.ts:195`) now have no production caller at all, and the option's own doc comment still opens "Provider's `loopAbandoned` flag".

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -423,5 +461,43 @@ export function usePayoutSigningState({
...
-  return { signing, progress, error, errorTerminal, isComplete, handleSign };
+  // Capability probe: only the Ledger vault provider exposes cancelSigning.
+  // Reads the provider that started the sign, not the live connector — a
+  // wallet swapped in mid-prompt must not retarget the affordance. The ref is
+  // only ever set/cleared together with the `signing` state, so this render
+  // read stays in sync.
+  const signingProvider = signingProviderRef.current as {
+    cancelSigning?: unknown;
+  } | null;
+  const canCancel =
```
gbarkhatov, line 472: **`canCancel` isn't narrowed to a cancellable device window.**

It gates on bare `signing`, which spans the whole payout flow — unlike `useDepositFlow`, which narrows the same affordance to `deviceSignActive` around exactly the `signPsbt`/`signPsbts`/`signMessage` calls. `signing` is set at :280, before the terms rebuild, VP auth, every claimer round, and submission.

Only those three methods go through `withSignAbort` (`provider.ts:653,672,979`), and `signAbortController` is `undefined` outside them, so `cancelSigning()` is a no-op everywhere else — including during `deriveContextHash` and `approveDepositTerms` (`provider.ts:448,487`), which are real on-device approval screens. Meanwhile the modal shows "finish or reject the request on your signing device" during phases where no device prompt exists at all.

The local `abortRef.abort()` still stops the flow at the next `throwIfAborted` checkpoint, so nothing wedges — but it can land between claimer rounds after some payouts were already submitted, and the user gets a device-shaped notice for an HTTP-shaped wait.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1419,6 +1475,28 @@ export function useDepositFlow(
...
+    // Cancel the ceremony on the provider that started it — never the live
+    // prop, which a mid-prompt wallet swap can replace.
+    const provider = deviceSignProviderRef.current as {
+      cancelSigning?: () => void;
+    } | null;
+    if (!deviceSignActiveRef.current) return;
+    if (typeof provider?.cancelSigning !== "function") return;
+    setDeviceCancelRequested(true);
+    // Device-only cancel (aborting the flow controller reads as a closed modal).
+    // Settles like an on-device reject: pre-broadcast = "Signing rejected"
```
gbarkhatov, line 1488: **The pre-broadcast outcome misattributes the cancel, then names a button that isn't rendered.**

Because the provider now emits `CONNECTION_REJECTED`, `isTypedUserRejectionFrame` claims it at step 2 of `mapDepositError` (`depositErrors.ts:145-147`), so a user who clicked our own "Cancel signing" is told "Signing rejected — You rejected the request in your wallet. Click Retry to approve it and continue." (`copy.ts:880-883`). `useDepositFlow.test.tsx:1716` pins that.

And `DepositSignContent` never passes `onRetry`, so `DepositProgressView`'s button falls through to `onClose` and renders "Close" (`DepositProgressView.tsx:496-497,514-517`).

`deviceCancelRequested` is already tracked; the settle path can distinguish a self-requested cancel the same way `usePayoutSigningState` does at :404-411.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -904,8 +909,8 @@ export class LedgerVaultProvider implements IBTCProvider {
   /**
    * Classify one device-ceremony failure, for every SIGN_PSBT caller: a
    * disconnect must surface as WALLET_NOT_CONNECTED, never as a generic
-   * UNKNOWN_ERROR. Touches only {@link loopAbandoned} — the intent mirror is
-   * the caller's to reset, since PoP must never touch it.
+   * UNKNOWN_ERROR. Mirror resets are the caller's, except the user-cancel
+   * branch, which resets here so PoP cancels take the same re-ceremony path.
```
gbarkhatov, line 913: This new contract makes the `signMessage` JSDoc at :969-976 false. It says PoP signing "never touches the intent mirror or the signed-fingerprint set", and this branch — reached from `signMessage`'s catch at :1013 — now clears both.

The uniform policy is defensible (an interrupted dispatcher is tx-type-agnostic, and the test says so), but the cost is real: cancelling a PoP now forces a full derive + re-approve for a signature that never touched the vault context. Worth updating that comment either way.

services/vault/src/utils/errors/deviceErrors.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (!frame || typeof frame !== "object") return undefined;
+  const { code } = frame as { code?: unknown };
+  return DEVICE_CODES.find((c) => c === code);
+}
+
+/**
+ * True when the error or its cause chain carries the code. Cause-walking:
+ * callers must run it AFTER their typed top-frame buckets.
+ */
+function chainCarriesCode(error: unknown, code: string): boolean {
```
gbarkhatov, line 41: This is the third copy of the same cause walk — structurally identical to `isWalletMethodNotSupported` (`walletMethodNotSupported.ts:31-45`) and to `isUserCancellation`'s loop, just parameterised by code. `MAX_CAUSE_DEPTH = 10` now lives in three files plus a bare `10` at `formatting.ts:141`.

A dependency-free `causeChain.ts` exporting one `chainCarriesCode` satisfies this file's "must not import from formatting.ts or depositErrors.ts" constraint — `walletMethodNotSupported.ts` is already such a leaf — and collapses all four.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -423,5 +461,43 @@ export function usePayoutSigningState({
     onSuccess,
   ]);
 
-  return { signing, progress, error, errorTerminal, isComplete, handleSign };
+  // Capability probe: only the Ledger vault provider exposes cancelSigning.
+  // Reads the provider that started the sign, not the live connector — a
+  // wallet swapped in mid-prompt must not retarget the affordance. The ref is
+  // only ever set/cleared together with the `signing` state, so this render
+  // read stays in sync.
+  const signingProvider = signingProviderRef.current as {
```
gbarkhatov, line 469: `cancelSigning` isn't on `IBTCProvider`, so every call site hand-rolls a structural cast — four across two hooks (:469, :478, `useDepositFlow.ts:1481,1497`).

`IBTCProvider` already documents exactly this pattern for `getAccounts?()` (`core/types.ts:387-402`): optional method, callers MUST feature-detect. Declaring `cancelSigning?(): void` there plus one shared type guard would put the probe under the type system instead of four `as unknown` casts.

services/vault/src/copy.ts
```diff
@@ -58,6 +58,17 @@ const WRONG_WALLET_BODY =
...
+// two can't drift; wording warns that retry re-runs the device approval screens.
+const DEVICE_CEREMONY_INVALID_TITLE = "Device approval needed again";
+const DEVICE_CEREMONY_INVALID_BODY =
+  "Your signing device no longer holds this deposit's approval. Try again to restart from the device approval screens.";
+const DEVICE_LOCKED_TITLE = "Signing device locked";
+const DEVICE_LOCKED_BODY =
+  "Your signing device is locked. Unlock it with your PIN and try again.";
+const DEVICE_WRONG_APP_TITLE = "Wrong app on device";
+const DEVICE_WRONG_APP_BODY =
+  "A different app is open on your signing device. Open the Babylon Vault app on the device and try again.";
```
gbarkhatov, line 71: "Open the Babylon Vault app" is wrong on testnet — the app is named "Babylon Vault Testnet", which the signer itself calls out at `dmkSession.ts:96-99`. A testnet depositor is told to open an app that isn't the one they need.

The pre-existing `rawApdu.ts:79` string has the same problem, so one network-aware phrasing would fix both.

gbarkhatov (approved):
Reviewed the full non-test diff plus the surrounding signer / connector / vault-app context. The typed device-error work (`DEVICE_CEREMONY_INVALID` / `DEVICE_LOCKED` / `DEVICE_WRONG_APP`), the ordering of the new buckets ahead of the cause walks, and the provider-bound cancel are all sound.

Approving. The comments below are one blocker-shaped question about the version-gate removal, three behavioural gaps in the cancel path, and some cleanup.

Three claims that came up during review and do **not** hold, recorded so nobody re-raises them:

- `DepositTerms.vaultCoreVersion` is not dead — `provider.ts:514` still uses it as the first component of the `fingerprintTerms` idempotence key.
- `handleCancel`'s `abortRef.abort()` cannot flip the resume modal to the re-offer state while the device ceremony is still pending: the SDK presign flow uses `signal?.throwIfAborted()` checkpoints (`runDepositorPresignFlow.ts:445,476,507,535`), not a race against the awaited device call.
- The `!isLedgerSignPsbtAbortedError(error)` guard at `provider.ts:899` is redundant now, but only as a no-op re-assignment — harmless.

jonybur (approved):
Typed device-error work (DEVICE_CEREMONY_INVALID / DEVICE_LOCKED / DEVICE_WRONG_APP) and the provider-bound cancel path are sound. Cancel state (provider ref, in-flight signal, cancelRequested, pending-signing) is correctly threaded through useDepositFlow and usePayoutSigningState. Test coverage on the new buckets, the cause-walk ordering, and the connector-source drift assertions is solid.

packages/babylon-ledger-vault-signer/src/envelope.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

services/vault/src/utils/errors/deviceErrors.ts
jrwbabylonlab: fixed

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
jrwbabylonlab: fixed

services/vault/src/copy.ts
jrwbabylonlab: fixed

PR #2299, opened 2026-08-23
feat(ts-sdk): verify P2WPKH depositor signatures host-side

### LINKED ISSUE #2284: Verify P2WPKH depositor signatures host-side (PoP witness + Pre-PegIn inputs)
Two related host-side verification gaps for Native SegWit (P2WPKH) depositors, both explicit pass-throughs today:

1. **PoP:** `verifyPopWitness` (`packages/babylon-ts-sdk/src/tbv/core/managers/pegin/verifyPopWitness.ts`, #2221) verifies one-item P2TR witnesses but returns `{ kind: "p2wpkh-unverified" }` for two-item P2WPKH witnesses without verification. vaultd's reference: btc-vault `crates/btc-signer/src/message.rs:105-144` (cheap first check: the witness pubkey must match the depositor key — `WitnessPubkeyMismatch`; then full BIP-322 P2WPKH verify per the bip322 crate).
2. **Pre-PegIn signatures:** `assertReturnedKeyPathSignatures` (#2222) skips non-P2TR inputs by design, so P2WPKH depositor inputs get no host-side signature verification at either finalize site. Reference: btc-vault `crates/btc-wallet-remote/src/client.rs` `verify_finalized_p2wpkh_spend`.

A bad PoP is a PERMANENT vaultd ingestion failure, so (1) is the higher-value half. Neither is a regression — no host verification existed for these shapes before #2221/#2222 either.

jrwbabylonlab:
Closes #2284. Host-side verification for Native SegWit depositors: BIP-322
P2WPKH proof-of-possession witnesses (pubkey mismatch detected before
signature verify, mirroring vaultd's WitnessPubkeyMismatch) and finalized
P2WPKH Pre-PegIn inputs at both finalize sites. Every check cites its
oracle (btc-vault message.rs / client.rs, bip322 crate 0.0.10); goldens
are the official BIP-322 vectors pinned from bitcoin/bips@d77863fb.

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/__tests__/bip322Verify.test.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/__tests__/bip322Verify.test.ts
@@ -1,23 +1,35 @@
 /**
- * Golden-vector test for BIP-322 simple verify.
+ * Golden-vector tests for BIP-322 simple verify.
  *
- * The signature + message + pubkey are emitted by the Rust reference
- * in `btc-vault/crates/btc-auth/src/server_identity.rs`. The signer
- * seed is 7 and the payload is produced by
+ * P2TR: the signature + message + pubkey are emitted by the Rust
+ * reference in `btc-vault/crates/btc-auth/src/server_identity.rs`.
+ * The signer seed is 7 and the payload is produced by
  * `build_server_identity_payload` with a seed-42 ephemeral pubkey
  * and expires_at = 1_700_000_000. These were generated end-to-end
  * by the Rust production path, not constructed ad-hoc — so if this
  * test passes, the TypeScript implementation byte-exactly matches
  * what a real VP would sign.
  *
- * Canonical hex lives in `./goldenVectors.ts` so that all
- * server-identity tests share one source of truth and can't drift.
+ * P2WPKH: the vectors are BIP-322's official ones, pinned in
+ * `./bip322P2wpkhVectors.ts`.
+ *
+ * Canonical hex lives in `./goldenVectors.ts` / `./bip322P2wpkhVectors.ts`
+ * so the suites share one source of truth and can't drift.
  */
 
+import { Buffer } from "buffer";
+
+import * as ecc from "@bitcoin-js/tiny-secp256k1-asmjs";
+import { sha256 } from "@noble/hashes/sha2.js";
+import { script as bscript, payments, Transaction } from "bitcoinjs-lib";
 import { describe, expect, it } from "vitest";
 
-import { verifyBip322Simple } from "../bip322Verify";
+import { verifyBip322P2wpkhSimple, verifyBip322Simple } from "../bip322Verify";
 import { encodeServerIdentityPayload } from "../cbor";
+import {
+  BIP322_P2WPKH_PUBKEY_HEX,
+  BIP322_P2WPKH_SIGNATURES,
+} from "./bip322P2wpkhVectors";
 import {
   GOLDEN_EPHEMERAL_PUBKEY_COMPRESSED,
   GOLDEN_EXPIRES_AT,
@@ -131,5 +143,155 @@ describe("verifyBip322Simple — end-to-end via encodeServerIdentityPayload", ()
         false,
       );
     }
+  });
+});
+
+const VECTOR_PUBKEY = fromHex(BIP322_P2WPKH_PUBKEY_HEX);
+const utf8 = (s: string) => new TextEncoder().encode(s);
+
+// In-test BIP-322 P2WPKH signer for cases the official vectors can't cover
+// (wrong hashtype, short DER encoding). Reconstructs to_spend/to_sign per the
+// BIP-322 spec ("simple" format); the official-vector tests above anchor that
+// the production verifier follows the same construction.
+const TEST_PRIV = Buffer.alloc(32, 0);
+TEST_PRIV[31] = 1; // privkey 1 → pubkey G; public test material.
+const TEST_PUB = Buffer.from(ecc.pointFromScalar(TEST_PRIV, true)!);
+
+function bip322P2wpkhSighash(
+  messageBytes: Uint8Array,
+  pubkey: Buffer,
+  hashType: number,
+): Buffer {
+  const th = Buffer.from(sha256(utf8("BIP0322-signed-message")));
+  const messageHash = Buffer.from(
+    sha256(Buffer.concat([th, th, Buffer.from(messageBytes)])),
+  );
+  const p2wpkh = payments.p2wpkh({ pubkey });
+  const toSpend = new Transaction();
+  toSpend.version = 0;
+  toSpend.locktime = 0;
+  toSpend.addInput(
+    Buffer.alloc(32, 0),
+    0xffffffff,
+    0,
+    Buffer.concat([Buffer.from([0x00, 0x20]), messageHash]),
+  );
+  toSpend.addOutput(p2wpkh.output!, 0);
+  const toSign = new Transaction();
+  toSign.version = 0;
+  toSign.locktime = 0;
+  toSign.addInput(toSpend.getHash(), 0, 0);
+  toSign.addOutput(Buffer.from([0x6a]), 0);
+  const scriptCode = payments.p2pkh({ hash: p2wpkh.hash! }).output!;
+  return toSign.hashForWitnessV0(0, scriptCode, 0, hashType);
+}
+
+describe("verifyBip322P2wpkhSimple — BIP-322 official P2WPKH vectors", () => {
+  it("accepts every official vector (both messages, 71- and 72-byte encodings)", () => {
+    for (const { message, encodedSignatureHexes } of BIP322_P2WPKH_SIGNATURES) {
+      for (const sigHex of encodedSignatureHexes) {
+        expect(
+          verifyBip322P2wpkhSimple(
+            utf8(message),
+            VECTOR_PUBKEY,
+            fromHex(sigHex),
+          ),
+        ).toBe(true);
+      }
+    }
+  });
+
+  it("rejects a tampered signature byte", () => {
+    const sig = fromHex(BIP322_P2WPKH_SIGNATURES[1].encodedSignatureHexes[0]);
+    sig[10] ^= 0x01; // inside the DER r value
+    expect(
+      verifyBip322P2wpkhSimple(utf8("Hello World"), VECTOR_PUBKEY, sig),
+    ).toBe(false);
+  });
+
+  it("rejects the right signature under a different pubkey", () => {
+    expect(
+      verifyBip322P2wpkhSimple(
+        utf8("Hello World"),
+        TEST_PUB, // valid compressed point, not the vector signer
+        fromHex(BIP322_P2WPKH_SIGNATURES[1].encodedSignatureHexes[0]),
+      ),
+    ).toBe(false);
+  });
+
+  it("rejects the wrong message", () => {
+    expect(
+      verifyBip322P2wpkhSimple(
+        utf8("Hello World!"),
+        VECTOR_PUBKEY,
+        fromHex(BIP322_P2WPKH_SIGNATURES[1].encodedSignatureHexes[0]),
+      ),
+    ).toBe(false);
+  });
+
+  it("rejects a trailing hashtype byte that is not SIGHASH_ALL", () => {
+    for (const hashType of [0x00, 0x02, 0x03, 0x81]) {
+      const sig = fromHex(BIP322_P2WPKH_SIGNATURES[1].encodedSignatureHexes[0]);
+      sig[sig.length - 1] = hashType;
+      expect(
+        verifyBip322P2wpkhSimple(utf8("Hello World"), VECTOR_PUBKEY, sig),
+      ).toBe(false);
+    }
+  });
+
+  it("rejects a cryptographically valid SIGHASH_NONE signature (gate, not sig failure)", () => {
+    // Signed over the real NONE sighash: without the explicit SIGHASH_ALL
+    // gate (bip322 crate 0.0.10 verify.rs:156-161) this would verify.
+    const message = utf8("sighash gate");
+    const sighashNone = bip322P2wpkhSighash(message, TEST_PUB, 0x02);
+    const encoded = bscript.signature.encode(
+      Buffer.from(ecc.sign(sighashNone, TEST_PRIV)),
+      0x02,
+    );
+    expect(verifyBip322P2wpkhSimple(message, TEST_PUB, encoded)).toBe(false);
+  });
+
+  it("rejects an encoded signature shorter than 71 bytes even when cryptographically valid", () => {
+    // vaultd's verifier accepts ONLY 71/72-byte encodings (bip322 crate
+    // 0.0.10 verify.rs:141-153), so a short-DER signature would fail
+    // ingestion permanently — the host gate must reject it too. Grind the
+    // sign entropy until r or s DER-encodes one byte short.
+    const message = utf8("short der");
+    const sighash = bip322P2wpkhSighash(message, TEST_PUB, 0x01);
+    let shortEncoded: Buffer | undefined;
+    for (let i = 0; i < 5000 && !shortEncoded; i++) {
+      const entropy = Buffer.alloc(32, 0);
+      entropy.writeUInt32LE(i, 0);
+      const candidate = bscript.signature.encode(
+        Buffer.from(ecc.sign(sighash, TEST_PRIV, entropy)),
+        0x01,
+      );
+      if (candidate.length < 71) shortEncoded = candidate;
+    }
+    expect(shortEncoded).toBeDefined();
+    // Sanity: the short signature IS valid ECDSA for this sighash.
+    expect(
+      ecc.verify(
+        sighash,
+        TEST_PUB,
+        bscript.signature.decode(shortEncoded!).signature,
+        true,
+      ),
+    ).toBe(true);
+    expect(verifyBip322P2wpkhSimple(message, TEST_PUB, shortEncoded!)).toBe(
+      false,
+    );
+  });
+
+  it("rejects a pubkey that is not on the curve or not 33 bytes", () => {
+    const sig = fromHex(BIP322_P2WPKH_SIGNATURES[0].encodedSignatureHexes[0]);
+    const notAPoint = Buffer.concat([
+      Buffer.from([0x02]),
+      Buffer.alloc(32, 0xff), // x ≥ field prime — no point has this encoding
+    ]);
+    expect(verifyBip322P2wpkhSimple(utf8(""), notAPoint, sig)).toBe(false);
+    expect(
+      verifyBip322P2wpkhSimple(utf8(""), VECTOR_PUBKEY.subarray(1), sig),
+    ).toBe(false);
   });
 });
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/bip322Verify.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/bip322Verify.ts
@@ -69,6 +69,21 @@ const P2WPKH_ENCODED_SIG_MAX = 72;
 // plain `0` everywhere.
 const ZERO_SATS = 0;
 
+// Wire fields of the virtual to_spend/to_sign transactions (`bip322`
+// crate 0.0.10 `util.rs:18-85`).
+const BIP322_TX_VERSION = 0;
+const BIP322_TX_LOCKTIME = 0;
+const BIP322_INPUT_SEQUENCE = 0;
+/** to_spend prevout: all-zero 32-byte txid at index 0xFFFFFFFF (`util.rs:18-85`). */
+const TO_SPEND_PREVOUT_TXID_BYTES = 32;
+const TO_SPEND_PREVOUT_INDEX = 0xffffffff;
+/** to_sign spends to_spend's only output (`util.rs:18-85`). */
+const TO_SPEND_OUTPUT_INDEX = 0;
+const OP_0 = 0x00;
+/** Direct push of the 32-byte tagged message hash. */
+const OP_PUSHBYTES_32 = 0x20;
+const OP_RETURN = 0x6a;
+
 /**
  * BIP-340 tagged hash: `SHA256( SHA256(tag) || SHA256(tag) || data )`.
  * Used for both BIP-322 message hashing and BIP-341 tap-tweak.
@@ -112,27 +127,30 @@ function buildToSignTransaction(
   const messageHash = taggedHash(BIP322_TAG, messageBytes);
 
   const toSpend = new Transaction();
-  toSpend.version = 0;
-  toSpend.locktime = 0;
-  // scriptSig: OP_0 (0x00) + OP_PUSHBYTES_32 (0x20) + message_hash (32B)
+  toSpend.version = BIP322_TX_VERSION;
+  toSpend.locktime = BIP322_TX_LOCKTIME;
   const scriptSig = Buffer.concat([
-    Buffer.from([0x00, 0x20]),
+    Buffer.from([OP_0, OP_PUSHBYTES_32]),
     Buffer.from(messageHash),
   ]);
   toSpend.addInput(
-    Buffer.alloc(32, 0), // prev_txid = 0x0000...0000
-    0xffffffff, // prev_vout = 0xFFFFFFFF
-    0, // sequence = 0
+    Buffer.alloc(TO_SPEND_PREVOUT_TXID_BYTES, 0),
+    TO_SPEND_PREVOUT_INDEX,
+    BIP322_INPUT_SEQUENCE,
     scriptSig,
   );
   toSpend.addOutput(scriptPubKey, ZERO_SATS);
 
   const toSign = new Transaction();
-  toSign.version = 0;
-  toSign.locktime = 0;
+  toSign.version = BIP322_TX_VERSION;
+  toSign.locktime = BIP322_TX_LOCKTIME;
   // Bitcoin txid in natural-byte (little-endian) form.
-  toSign.addInput(toSpend.getHash(), 0, 0);
-  toSign.addOutput(Buffer.from([0x6a]), ZERO_SATS); // OP_RETURN
+  toSign.addInput(
+    toSpend.getHash(),
+    TO_SPEND_OUTPUT_INDEX,
+    BIP322_INPUT_SEQUENCE,
+  );
+  toSign.addOutput(Buffer.from([OP_RETURN]), ZERO_SATS);
 
   return toSign;
 }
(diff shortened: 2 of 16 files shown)


packages/babylon-ts-sdk/src/testing/MockBitcoinWallet.ts
```diff
@@ -137,35 +143,15 @@ export class MockBitcoinWallet implements BitcoinWallet {
...
-        `mock-signature-${type}-${message}-${this.config.publicKeyHex}`,
-      ),
+    // The SDK cryptographically verifies what a wallet returns
+    // (`verifyPopWitness`), so the mock signs a REAL BIP-322 P2WPKH witness
+    // with its private key. The type is not consulted: PoP flows only ever
+    // request "bip322-simple", and the interface tests assert shape only.
+    const witness = signBip322P2wpkhWitness(
+      new TextEncoder().encode(message),
+      Uint8Array.from(Buffer.from(this.config.privateKeyHex, "hex")),
     );
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/bip322Verify.ts
```diff
@@ -82,6 +100,43 @@ function tweakXOnlyKey(xOnly: Uint8Array): Uint8Array | null {
...
+    scriptSig,
+  );
+  toSpend.addOutput(scriptPubKey, ZERO_SATS);
+
+  const toSign = new Transaction();
+  toSign.version = 0;
+  toSign.locktime = 0;
+  // Bitcoin txid in natural-byte (little-endian) form.
+  toSign.addInput(toSpend.getHash(), 0, 0);
+  toSign.addOutput(Buffer.from([0x6a]), ZERO_SATS); // OP_RETURN
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed the wire literals; not extracting verdict/error strings — typed unions and message text, not magic values

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/__tests__/verifyKeyPathSchnorrSignature.test.ts
```diff
@@ -329,3 +331,375 @@ describe("assertReturnedKeyPathSignatures", () => {
...
+    badType.updateInput(1, {
+      partialSig: [{ pubkey: PUB, signature: signP2wpkh(req, 1) }],
+    });
+    expect(() =>
+      assertReturnedKeyPathSignatures({
+        requestedPsbtHex: req.toHex(),
+        returnedPsbtHex: badType.toHex(),
+      }),
+    ).toThrow(/input 0/);
+  });
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed; kept the differential loop paired — tamper rejection is only meaningful against its own accepted fixture

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/bip322Verify.ts
```diff
@@ -190,3 +228,69 @@ export function verifyBip322Simple(
...
+  }
+  // Any exception below (strict-DER decode, malformed pubkey) is a
+  // verification failure, not an error — a verifier returns a boolean.
+  try {
+    // Full curve-point parse, as `PublicKey::from_slice` (`verify.rs:82-83`).
+    if (!ecc.isPointCompressed(compressedPubkey)) return false;
+
+    // Strict DER decode, then SIGHASH_ALL only — deliberately NARROWER than
+    // `verify.rs:156-161` (its from_consensus maps more bytes to All); host-stricter is safe.
+    const { signature, hashType } = bscript.signature.decode(
```
gbarkhatov, line 271: **High — the host accepts a signature vaultd will reject.**

`bscript.signature.decode` does not bound DER integer width. Its `fromDER` copies an over-long R into a fixed 32-byte buffer and silently drops the overflow byte, and `bip66.decode` bounds the *total* length, not `lenR`. So appending one garbage byte to a 71-byte encoding's R integer yields a 72-byte encoding that passes the `P2WPKH_ENCODED_SIG_MIN/MAX` gate, decodes back to the original valid `(r, s)`, and verifies `true`.

Reproduced against the exported function on this branch:

```text
base  (71B)  3044 0220 1d43e2…a7a7      0220 7d1ba2…9091 01  -> verifyBip322P2wpkhSimple = true
forged(72B)  3045 0221 1d43e2…a7a7 ab   0220 7d1ba2…9091 01  -> verifyBip322P2wpkhSimple = true
```

Node/OpenSSL rejects that same forged DER (`crypto.verify` -> `false`), and libsecp256k1's `secp256k1_der_parse_integer` sets `overflow` once the integer exceeds 32 bytes and zeroes the scalar — so `Signature::from_der` + `verify_ecdsa` fails on the vaultd side. That is host-accepts / vaultd-rejects on the PoP path: the registration gets committed, then `InvalidDepositorPop` is permanent. Exactly the failure class this PR exists to prevent.

The comment on L269-270 ("deliberately NARROWER … host-stricter is safe") is therefore not accurate: on integer width the host is *looser* than `from_slice`, not stricter.

Suggested fix — re-encode and compare, which makes canonicality an explicit check rather than a property inherited from the decoder:

```ts
const raw = Buffer.from(encodedSignature);
const { signature, hashType } = bscript.signature.decode(raw);
if (!bscript.signature.encode(signature, hashType).equals(raw)) return false;
```

I checked this round-trips byte-identically for all four official vectors and 400 randomly drawn signatures, so it's safe to add.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyP2wpkhEcdsaSignature.ts
```diff
@@ -0,0 +1,247 @@
...
+        `prevout's witness program.`,
+    );
+  }
+
+  // Strict-DER decode with defined-hashtype gate — the accept set of
+  // ecdsa::Signature::from_slice / EcdsaSighashType::from_standard
+  // (client.rs:978-980) — then SIGHASH_ALL exactly (client.rs:981-987).
+  let decoded: { signature: Buffer; hashType: number };
+  try {
+    decoded = bscript.signature.decode(Buffer.from(encodedSignature));
```
gbarkhatov, line 199: **High — same root cause as the `bip322Verify.ts` comment, worse consequence.**

This site has no length gate at all, so any over-long DER integer `bip66` tolerates gets through. Reproduced end-to-end against `assertReturnedP2wpkhSignature` on this branch: a 72-byte non-canonical encoding (a valid 71-byte signature with one byte appended to R) is accepted with no throw.

Here the result isn't a rejected registration — the witness bytes we just blessed go on to be broadcast, and Bitcoin Core's `ecdsa_signature_parse_der_lax` sets overflow and zeroes the signature for `lenR > 32`, so the transaction is consensus-invalid. Under CLAUDE.md §8 this verifier is the only thing between a wallet's output and a broadcast.

Same fix: assert the encoding round-trips — `bscript.signature.encode(decoded.signature, decoded.hashType).equals(raw)` — before verifying.

packages/babylon-ts-sdk/src/tbv/core/managers/pegin/verifyPopWitness.ts
```diff
@@ -114,34 +124,48 @@ export function verifyPopWitness(
...
     if (witnessXOnlyHex !== depositorXOnlyHex.toLowerCase()) {
       throw new Error(
         `proof of possession witness pubkey does not match the depositor key: ` +
           `witness carries ${witnessXOnlyHex}, expected ${depositorXOnlyHex}`,
       );
     }
-    return { kind: "p2wpkh-unverified" };
+    // BIP-322 simple verification against the P2WPKH address of the witness
+    // pubkey (message.rs:125-133; network affects only bech32 encoding).
+    if (!verifyBip322P2wpkhSimple(messageBytes, pubkey, encodedSignature)) {
```
gbarkhatov, line 163: **Medium — this error blames the wrong thing, for ~1 in 260 depositors, permanently.**

`verifyBip322P2wpkhSimple` returns a bare `false` for five distinct causes: the 71/72-byte length gate, a non-compressed or off-curve pubkey, a DER decode failure, a non-SIGHASH_ALL hashtype, and a genuinely bad signature. All five land on "does not verify against the depositor key", which sends the depositor chasing a wrong-account problem that doesn't exist.

The length gate is the one that bites. Measured over 100 000 RFC6979 signatures, the encoded-length histogram is `{69: 5, 70: 380, 71: 49904, 72: 49711}` — **0.385%** encode to fewer than 71 bytes. And the PoP message is fully deterministic: `${eth}:${chainId}:pegin:${contract}` at `PeginManager.ts:1808`, no nonce, no timestamp. With RFC6979 an affected depositor gets the byte-identical short signature on every retry, forever. Low-R-grinding wallets are equally stuck, since grinding is deterministic too.

To be clear, the gate itself is right and should stay — it mirrors `verify.rs:141-153`, and without it the same depositor fails later at ingestion with a permanent `InvalidDepositorPop`. What's wrong is the diagnosis. Thirty lines up, each P2WPKH pubkey check gets its own named error; this one collapses five causes into a message that names the wrong one.

Suggest hoisting the length and hashtype checks into `verifyPopWitness` with their own messages — keeping the gates inside `verifyBip322P2wpkhSimple` so the exported verifier stays exactly as strict as vaultd — so an affected user gets "your wallet produced a 70-byte signature encoding, which the protocol does not accept" instead of "wrong key".

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyKeyPathSchnorrSignature.ts
```diff
@@ -224,10 +236,22 @@ export function assertReturnedKeyPathSignatures(
 
   let verified = 0;
   requested.data.inputs.forEach((input, inputIndex) => {
-    if (!isKeyPathEligible(input)) return;
+    const returnedInput = returned.data.inputs[inputIndex];
+
+    if (!isKeyPathEligible(input)) {
+      // Native SegWit funding inputs get the ECDSA check; anything else
+      // (script-path, P2WSH, ...) is another verifier's job.
```
gbarkhatov, line 243: **Medium — this comment, and the JSDoc at L211-212 that gets regenerated into `docs/api/primitives.md`, overstate coverage.**

"they have their own checks" / "another verifier's job" holds for script-path, but not for P2WSH. `getPsbtInputFields` (`utils/btc/psbtInputFields.ts:70-81`) *explicitly supports* P2WSH funding and emits a `witnessUtxo` + `witnessScript` for it. So a P2WSH-funded input reaches here, isn't key-path eligible, isn't `isP2wpkhScript`, and is silently skipped with no verification anywhere.

It also contributes nothing to `verified`, and both call sites' count gates are inert unless `supportsDepositApproval` / `expectAllInputsKeyPath` is set — so nothing catches the zero-coverage case for a software wallet.

On a §8 path, a comment asserting coverage that doesn't exist is what the next reviewer reads as settled. Either say "unverified" plainly, or better: throw on an input this function can classify neither way, so an unverifiable funding type can't reach broadcast unnoticed.

packages/babylon-ts-sdk/src/testing/signBip322P2wpkhWitness.ts
```diff
@@ -0,0 +1,142 @@
...
+import { script as bscript, payments, Transaction } from "bitcoinjs-lib";
+import { Buffer } from "buffer";
+
+/** BIP-322 message tag (BIP-340 tagged-hash style). */
+const BIP322_TAG = "BIP0322-signed-message";
+const SIGHASH_ALL = 0x01;
+const ZERO_SATS = 0;
+/**
+ * vaultd accepts only 71/72-byte encoded signatures (`bip322` crate 0.0.10
+ * `verify.rs:141-153`); a short-DER signature (~0.8% of draws) is ground
```
gbarkhatov, line 26: Nit — measured, `~0.8%` is about 2× high. Over 100 000 RFC6979 signatures the encoded-length histogram is `{69: 5, 70: 380, 71: 49904, 72: 49711}`, i.e. **0.385%** below 71 bytes. (The naive model double-counts: a 33-byte R with a 31-byte S still lands on 71.) The grind bound is fine either way — the error is in the safe direction — but the numbers should read `~0.4%` and `0.004^100`.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/bip322Verify.ts
```diff
@@ -190,3 +228,69 @@ export function verifyBip322Simple(
...
+    // `OP_DUP OP_HASH160 <20B> OP_EQUALVERIFY OP_CHECKSIG` and value 0
+    // (`verify.rs:163-173 p2wpkh_signature_hash`; scriptCode template per
+    // bitcoinjs-lib's own P2WPKH signer, `src/psbt.js:1245-1255`).
+    const scriptCode = payments.p2pkh({ hash: p2wpkh.hash }).output;
+    if (!scriptCode) return false;
+    const sighash = toSign.hashForWitnessV0(0, scriptCode, ZERO_SATS, hashType);
+
+    // strict=true rejects high-S, as libsecp256k1's verify does
+    // (`verify.rs:180-182`; secp256k1 crate `ecdsa/mod.rs:194`).
+    return ecc.verify(sighash, compressedPubkey, signature, true);
```
gbarkhatov, line 292: Nit — nothing pins `strict: true`, and it's the only thing enforcing low-S parity with vaultd. Dropping the argument (it defaults to `false`) or flipping it would pass CI silently while diverging from `verify_ecdsa`.

It's cheaply detectable: I confirmed on this branch that a negated-S signature gives `ecc.verify(…, true) === false` and `ecc.verify(…, false) === true`. One high-S vector asserting `false` here, and one for `verifyP2wpkhEcdsaSignature.ts:236` (same flag), closes it.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/__tests__/bip322Verify.test.ts
```diff
@@ -133,3 +145,153 @@ describe("verifyBip322Simple — end-to-end via encodeServerIdentityPayload", ()
...
+
+// In-test BIP-322 P2WPKH signer for cases the official vectors can't cover
+// (wrong hashtype, short DER encoding). Reconstructs to_spend/to_sign per the
+// BIP-322 spec ("simple" format); the official-vector tests above anchor that
+// the production verifier follows the same construction.
+const TEST_PRIV = Buffer.alloc(32, 0);
+TEST_PRIV[31] = 1; // privkey 1 → pubkey G; public test material.
+const TEST_PUB = Buffer.from(ecc.pointFromScalar(TEST_PRIV, true)!);
+
+function bip322P2wpkhSighash(
```
gbarkhatov, line 160: Nit — `bip322P2wpkhSighash` reimplements the production to_spend/to_sign construction and feeds exactly two tests (L242 SIGHASH_NONE, L254 short-DER), both asserting `false`. If the helper computed the wrong sighash, both would still pass — for the wrong reason — and stay green. The `ecc.verify` sanity check at L273 doesn't close this: it verifies against the same locally computed sighash, not production's.

One positive assertion using the helper (sign with `0x01`, keep an encoding ≥71 bytes, expect `verifyBip322P2wpkhSimple` -> `true`) retroactively makes both negatives meaningful.

Separately, the helper duplicates `src/testing/signBip322P2wpkhWitness.ts` almost exactly, and that one is already positively anchored end-to-end by `MockBitcoinWallet.test.ts`. It can't produce these two negative cases (it always grinds to ≥71 bytes and always SIGHASH_ALL), so a local helper is justified — but a shared to_sign builder would keep the two from drifting.

packages/babylon-ts-sdk/src/testing/MockBitcoinWallet.ts
```diff
@@ -127,7 +152,7 @@ export class MockBitcoinWallet implements BitcoinWallet {
 
   async signMessage(
     message: string,
-    type: "bip322-simple" | "ecdsa",
+    _type: "bip322-simple" | "ecdsa",
```
gbarkhatov, line 155: Nit — `_type` means `"ecdsa"` is silently accepted and served a BIP-322 P2WPKH witness. No production caller passes it, but `shared/wallets/interfaces/__tests__/BitcoinWallet.test.ts` — the suite whose job is to pin this interface contract — calls `signMessage(…, "ecdsa")` four times (L83, L91, L98, L106) and asserts on a BIP-322 witness. So the one place documenting "ecdsa" behaviour documents something the mock never does.

Renaming a parameter to `_x` is the classic shape of the silent fallback CLAUDE.md asks you to throw on instead. Suggest throwing for `"ecdsa"` and flipping those four call sites to `"bip322-simple"`.

packages/babylon-ts-sdk/src/testing/MockBitcoinWallet.ts
```diff
@@ -85,10 +104,16 @@ export class MockBitcoinWallet implements BitcoinWallet {
...
         : {}),
+      ...(config.privateKeyHex ? { privateKeyHex: config.privateKeyHex } : {}),
       ...(config.deriveContextHash
         ? { deriveContextHash: config.deriveContextHash }
         : {}),
     };
+    // A private-key override without a public key would otherwise sign
+    // against the DEFAULT pubkey — derive the matching one instead.
+    if (config.privateKeyHex && !config.publicKeyHex) {
+      this.config.publicKeyHex = xOnlyPubkeyHexOf(config.privateKeyHex);
```
gbarkhatov, line 115: Nit — this derivation is constructor-only, so `updateConfig` is asymmetric with it: `new MockBitcoinWallet({ privateKeyHex: X })` works, while `new MockBitcoinWallet().updateConfig({ privateKeyHex: X })` leaves `publicKeyHex` at the default and then trips the new divergence throw at sign time. It fails loudly rather than silently, so this is only a trap for the next test author — but extracting the derive into a small private helper both paths call would remove it.

packages/babylon-ts-sdk/src/testing/MockBitcoinWallet.ts
```diff
@@ -53,22 +67,27 @@ const defaultDeriveContextHash = async (
...
+    ),
+  );
+}
 
 const DEFAULT_CONFIG: Required<MockBitcoinWalletConfig> = {
+  // x-only pubkey of DEFAULT_PRIVATE_KEY_HEX (the secp256k1 generator's x
+  // coordinate), so the default wallet's PoP witnesses verify against it.
   publicKeyHex:
-    "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3f4a5b6c7d8e9f0a1b2",
+    "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
```
gbarkhatov, line 86: Nit — worth a comment recording that this value is load-bearing outside this file. It's `G.x` (privkey 1), which is *also* the value of `TEST_KEYS.DEPOSITOR` in `PeginManager.test.ts:159`, `PayoutManager.test.ts:50`, and `PeginManager.rebuildDepositTerms.test.ts:86`. That coincidence is the only reason the suites doing `new MockBitcoinWallet({ publicKeyHex: TEST_KEYS.DEPOSITOR })` and then signing still pass under the new consistency check.

Rotate `TEST_KEYS.DEPOSITOR` to any other valid x-only key and they all throw — with a remedy ("pass a matching publicKeyHex/privateKeyHex pair") that can't be satisfied, since it would need the discrete log. One line at each definition site saying the two are deliberately the same key would save someone an afternoon.

packages/babylon-ts-sdk/src/testing/signBip322P2wpkhWitness.ts
```diff
@@ -0,0 +1,142 @@
...
+ * same reference the production verifier
+ * (`tbv/core/clients/vault-provider/auth/bip322Verify.ts`) mirrors, and
+ * the BIP-322 official vectors pin.
+ *
+ * Test-only: never feed real key material to this module.
+ */
+
+import * as ecc from "@bitcoin-js/tiny-secp256k1-asmjs";
+import { sha256 } from "@noble/hashes/sha2.js";
+import { script as bscript, payments, Transaction } from "bitcoinjs-lib";
```
gbarkhatov, line 17: Nit — `src/testing/` had no `bitcoinjs-lib` or `tiny-secp256k1` import before this PR; this file and `MockBitcoinWallet.ts:1` add both. They're already package-level peer deps so nothing breaks today, but `./testing` is a published entry point and CLAUDE.md §9 registers the optional-BTC separation (#2228 / #2229) that would make those peers optional. Worth a conscious call now rather than a discovery during that split.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyP2wpkhEcdsaSignature.ts
```diff
@@ -0,0 +1,247 @@
...
+ */
+export function assertReturnedP2wpkhSignature(
+  params: AssertReturnedP2wpkhSignatureParams,
+): void {
+  const { requestedPsbtHex, returnedInput, inputIndex } = params;
+
+  const partials = returnedInput.partialSig;
+  // Consensus allows exactly one signature in a P2WPKH witness, so a second
+  // partialSig entry is ambiguous tampering, not a fallback to pick from.
+  if (partials !== undefined && partials.length > 1) {
```
gbarkhatov, line 100: Nit — this guard has no test. `verifyKeyPathSchnorrSignature.test.ts` covers the P2WPKH arm well (wrong prevout value, SIGHASH_NONE, corrupted DER, witness/partialSig disagreement), but every case builds a single-element `partialSig`, so the `> 1` branch never executes. It's a cheap one to add and it's the branch that rejects ambiguous tampering.

gbarkhatov (approved):
Approving — the crypto core is right, and this closes a real gap: P2WPKH depositor signatures were reaching both the Ethereum registration and the Pre-PegIn broadcast completely unverified.

**What I verified as correct**

- `assertP2wpkhEcdsaSignature` mirrors `verify_finalized_p2wpkh_spend` check for check: prevout script and value come from the *requested* PSBT (never the wallet's), 33-byte + `isPointCompressed` pubkey gate, `hash160(pubkey) === witness program`, an explicit `SIGHASH_ALL` gate before the hashType feeds the sighash, `ecc.verify(…, strict: true)` for low-S, and the out-of-range throw correctly treated as a failed verification.
- The `>1 partialSig` rejection and the byte-equality check between `partialSig` and the finalized witness are right, and verification runs *before* `finalizeAllInputs()` at both call sites.
- Dispatch is genuinely reachable — `getPsbtInputFields` emits a real `witnessUtxo` for P2WPKH at both build sites.
- `verified` keeps its key-path-only meaning, so the `supportsDepositApproval` / `expectAllInputsKeyPath` gates behave exactly as before for taproot wallets.
- `buildToSignTransaction` is byte-faithful to the previous P2TR path (the Rust golden vector still passes), and the P2WPKH arm is anchored on the official BIP-322 vectors plus in-test-signed negatives.
- Rename hygiene is clean: `p2wpkh-unverified` -> `p2wpkh-verified` fully propagated, no stale "not yet verified" text, docs regenerated for the two public symbols whose JSDoc changed.
- Full ts-sdk suite passes (101 files / 1694 tests); eslint clean on all six changed source files.

**Findings** — 2 high, 2 medium, 8 nits, inline.

The two high ones are a single root cause in two places: `bscript.signature.decode` doesn't bound DER integer width, so a non-canonical encoding (one garbage byte appended to R) is silently truncated back to the valid `(r, s)` and verifies `true`. I reproduced it end-to-end against both exported verifiers on this branch and confirmed OpenSSL rejects the same bytes. On the PoP path that's host-accepts / vaultd-rejects; on the PSBT path it blesses a witness that is consensus-invalid. The fix is a two-line re-encode-and-compare, which I checked round-trips byte-identically over the four official vectors and 400 random signatures.

Approving rather than requesting changes because none of this is wrong *crypto* — it's a strictness gap in a decoder, a misleading error string, and a comment that overstates coverage. Happy for the DER fix to land as a follow-up.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/bip322Verify.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyP2wpkhEcdsaSignature.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/managers/pegin/verifyPopWitness.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyKeyPathSchnorrSignature.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/testing/signBip322P2wpkhWitness.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/bip322Verify.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/auth/__tests__/bip322Verify.test.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/testing/MockBitcoinWallet.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/testing/MockBitcoinWallet.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/testing/MockBitcoinWallet.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/testing/signBip322P2wpkhWitness.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/verifyP2wpkhEcdsaSignature.ts
jrwbabylonlab: fixed

PR #2300, opened 2026-08-24
fix(vault, ledger-vault): stop the payout loop on cancel; honest canc…

jrwbabylonlab:
Addresses the nine open review comments on #2298 (all replied inline):
cancel stops the multi-vault payout loop (remaining vaults stay resumable
via the continuation view); self-cancel gets its own copy; the payout
modal's cancel button appears only during actual device-sign windows;
resendOnceOnIncorrectData/dispatcherInterrupted deleted end-to-end (no
production caller since the loopAbandoned removal — note for the signer's
next release notes: LedgerSignPsbtAbortedError lost a field and ctor arg);
envelope removal comment now cites the firmware-completes-v3 evidence;
shared causeChain leaf replaces three duplicated cause walks;
cancelSigning?() is a typed optional on IBTCProvider; wrong-app copy is
network-agnostic (testnet app is "Babylon Vault Testnet").

### DIFF
--- a/packages/babylon-ledger-vault-signer/src/envelope.ts
+++ b/packages/babylon-ledger-vault-signer/src/envelope.ts
@@ -36,8 +36,10 @@ function requireIntInRange(field: string, value: number, lo: number, hi: number)
  * @throws {DepositTermsRejectedError} on the first violation
  */
 export function assertDepositTermsDeviceCompatible(terms: DepositTerms): void {
-  // No tx-graph version check (2026-08-23 decision): the WASM capability gate
-  // stops unconstructable versions upstream; bad shapes fail closed on-device.
+  // No tx-graph version gate: v3 is Core 2's Bitcoin tx shape byte-for-byte (btc-vault
+  // e1e50f66; SDK parity vector pegin.test.ts "builds the v3 Pre-PegIn byte-identical to
+  // the v2 vector"); v1 is unreachable fresh (ProtocolParams setter rejects
+  // `newVersion <= prev`, and no Ledger v1 vault exists); v4+ needs a new WASM release.
 
   // The >= 1 floor is enforced by both the contract and the device
   // (vault_tlv.c:73 rejects rate == 0).
--- a/packages/babylon-ledger-vault-signer/src/errors.ts
+++ b/packages/babylon-ledger-vault-signer/src/errors.ts
@@ -169,36 +169,26 @@ export class LedgerSignPsbtProtocolError extends Error {
 
 /**
  * The host abandoned the SIGN_PSBT loop (abort signal) — the loop stops
- * sending, nothing else. What the device does next (verified at base app rev
- * `e400d8d8` + fw `90cf41f`; the `6fcad4fd`→`90cf41f` delta is payout-validation
- * only and touches neither path below):
+ * sending, nothing else. Retry after this is always a FULL re-ceremony (the
+ * provider resets its mirror to idle). Device aftermath (verified at base app
+ * rev `e400d8d8` + fw `90cf41f`):
  *
  * - Within ~5 s (50 ticks, `base:io_ext.h:28`) the device sits blocked
- *   awaiting CONTINUE; the next non-CONTINUE APDU is eaten with 0x6A80 and the
- *   interrupted handler unwinds (`base:dispatcher.c:107-111`) — a direct retry
- *   uses `resendOnceOnIncorrectData`.
- * - A validation-phase abandonment leaves the intent intact; a Payout/NoPayout
- *   SIGNING-phase abandonment invalidates it (`fw:sign_custom_inputs.c:207-344`)
- *   — post-abort "intent-loaded" is a HINT only, and the retry path must also
- *   absorb 0xB007.
+ *   awaiting CONTINUE and eats the next non-CONTINUE APDU with 0x6A80
+ *   (`base:dispatcher.c:107-111`) — a fast retry's first DERIVE APDU surfaces
+ *   as the generic invalid-data error; the second attempt is clean.
  * - After >5 s of host silence the app exits to the dashboard
- *   (`base:io_ext.c:103-111` → SDK `app_exit()`); the host then sees 0x6E00 or
- *   a transport failure — full re-ceremony required.
+ *   (`base:io_ext.c:103-111` → SDK `app_exit()`); the host then sees 0x6E00,
+ *   whose wrong-app copy guides reopening the app.
  */
 export class LedgerSignPsbtAbortedError extends Error {
   /** Yields the device had already delivered when the host stopped sending. */
   readonly yieldedCount: number;
-  /**
-   * True only when the abort left a dispatcher mid-interruption (device
-   * awaiting CONTINUE) — the one state the resend-once recovery exists for.
-   */
-  readonly dispatcherInterrupted: boolean;
 
-  constructor(yieldedCount: number, dispatcherInterrupted: boolean) {
+  constructor(yieldedCount: number) {
     super("SIGN_PSBT was abandoned host-side before completion");
     this.name = LEDGER_SIGN_PSBT_ABORTED_ERROR_NAME;
     this.yieldedCount = yieldedCount;
-    this.dispatcherInterrupted = dispatcherInterrupted;
   }
 }
 
--- a/packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
+++ b/packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
@@ -39,7 +39,6 @@ const P2_CONTINUE = 0x00;
 const SW_OK = 0x9000;
 /** A client-command request follows in the response data (`base:sw.h:90`). */
 const SW_INTERRUPTED_EXECUTION = 0xe000;
-const SW_INCORRECT_DATA = 0x6a80;
 
 export interface SignPsbtProgress {
   readonly inputIndex: number;
@@ -68,14 +67,6 @@ export interface RunSignPsbtLoopOptions {
    * diagnostics. Never gates control flow.
    */
   readonly appIdentity?: AppIdentity;
-  /**
-   * Provider's `loopAbandoned` flag: resend the initial SIGN_PSBT APDU once if
-   * it answers 0x6A80 — the dispatcher eats exactly one non-CONTINUE APDU
-   * after a host-abandoned interruption (`base:dispatcher.c:107-111`). A
-   * genuine validation 0x6A80 is indistinguishable on the wire, hence the
-   * explicit gate. Never applies to CONTINUEs; never more than one resend.
-   */
-  readonly resendOnceOnIncorrectData?: boolean;
 }
 
 /**
@@ -92,23 +83,13 @@ export async function runSignPsbtLoop(
   const { collector, interpreter, cdata } = getPreparedSignPsbtState(prepared);
   const { table } = prepared;
   if (opts.signal.aborted) {
-    // Abandoned before any I/O — no dispatcher interrupted.
-    throw new LedgerSignPsbtAbortedError(collector.yields.length, false);
+    // Abandoned before any I/O.
+    throw new LedgerSignPsbtAbortedError(collector.yields.length);
   }
 
   const signPsbtApdu = buildSignPsbtApdu(cdata);
   let lastSentApdu: Apdu = signPsbtApdu;
   let response = await send(signPsbtApdu);
-  if (response.sw === SW_INCORRECT_DATA && opts.resendOnceOnIncorrectData === true) {
-    if (opts.signal.aborted) {
-      // Abort raced the first exchange. The 0x6A80 either consumed the eaten
-      // slot or was a genuine rejection — either way the dispatcher is clean.
-      throw new LedgerSignPsbtAbortedError(collector.yields.length, false);
-    }
-    // Resend the SAME APDU once; this branch is structurally unreachable a
-    // second time — any later 0x6A80 (including on the resend) is terminal.
-    response = await send(signPsbtApdu);
-  }
 
   let reportedYields = 0;
   while (response.sw === SW_INTERRUPTED_EXECUTION) {
@@ -148,9 +129,9 @@ export async function runSignPsbtLoop(
       }
     }
     if (opts.signal.aborted) {
-      // Stop sending — the CONTINUE for this round is never sent, so the
-      // interrupted dispatcher will eat exactly one later non-CONTINUE APDU.
-      throw new LedgerSignPsbtAbortedError(collector.yields.length, true);
+      // Stop sending — the CONTINUE for this round is never sent. Recovery is
+      // a full re-ceremony (LedgerSignPsbtAbortedError's doc has the aftermath).
+      throw new LedgerSignPsbtAbortedError(collector.yields.length);
     }
     lastSentApdu = { cla: CLA_FRAMEWORK, ins: INS_CONTINUE, p1: P1_CONTINUE, p2: P2_CONTINUE, data: continueData };
     response = await send(lastSentApdu);
--- a/services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
+++ b/services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
@@ -709,18 +709,26 @@ describe("usePayoutSigningState", () => {
     // Ledger-shaped provider: the only BTC provider exposing cancelSigning.
     // No approveDepositTerms on purpose — the cancel seam is orthogonal to
     // the approval rebuild, so these tests stay on the software-wallet path.
+    // Its signPsbt never settles on its own: armPendingSdkCall's `pending`
+    // settles the SDK call while that device window is still held open.
     function connectCancellableWallet() {
       const cancelSigning = vi.fn();
       mockBtcConnector = {
         connectedWallet: {
           account: { address: "tb1test" },
-          provider: { signPsbt: vi.fn(), cancelSigning },
+          provider: {
+            signPsbt: vi.fn(() => new Promise<string>(() => {})),
+            cancelSigning,
+          },
         },
       };
       return { cancelSigning };
     }
 
-    // Holds the SDK call open so cancel/settle ordering is test-controlled.
+    // Holds the SDK call open so cancel/settle ordering is test-controlled,
+    // driving one wrapped signPsbt so the cancellable device window the
+    // affordance tracks is active while the call is held (an instant no-op
+    // window for the plain software wallet).
     function armPendingSdkCall() {
       const pending: {
         resolve: () => void;
@@ -728,11 +736,19 @@ describe("usePayoutSigningState", () => {
         signal: AbortSignal | undefined;
       } = { resolve: () => {}, reject: () => {}, signal: undefined };
       mockSignAndSubmitPayouts.mockImplementation(
-        ({ signal }: { signal: AbortSignal }) => {
+        ({
+          btcWallet,
+          signal,
+        }: {
+          btcWallet: { signPsbt: (hex: string) => Promise<string> };
+          signal: AbortSignal;
+        }) => {
           pending.signal = signal;
           return new Promise<void>((resolve, reject) => {
             pending.resolve = resolve;
             pending.reject = reject;
+            // Settled only via `pending` — swallow the held-open device sign.
+            void btcWallet.signPsbt("payout-psbt").catch(() => {});
           });
         },
       );
@@ -986,6 +1002,84 @@ describe("usePayoutSigningState", () => {
       // consumed so the modal cannot wedge on the disabled cancel button.
       expect(result.current.cancelRequested).toBe(false);
       expect(result.current.isComplete).toBe(true);
+    });
+
+    it("keeps canCancel false during a hung VP-auth (deriveContextHash) phase", async () => {
+      // deriveContextHash is a real device screen, but the provider's
+      // cancelSigning is a no-op outside signPsbt/signPsbts/signMessage —
+      // showing the affordance there would be a dead button.
+      const cancelSigning = vi.fn();
+      let settleAuth: (root: string) => void = () => {};
+      mockBtcConnector = {
+        connectedWallet: {
+          account: { address: "tb1test" },
+          provider: {
+            signPsbt: vi.fn(),
+            deriveContextHash: vi.fn(
+              () =>
+                new Promise<string>((resolve) => {
+                  settleAuth = resolve;
+                }),
+            ),
+            cancelSigning,
+          },
+        },
+      };
+      mockSignAndSubmitPayouts.mockImplementation(
+        async ({
+          btcWallet,
+        }: {
+          btcWallet: {
+            deriveContextHash: (app: string, ctx: string) => Promise<string>;
+          };
+        }) => {
+          await btcWallet.deriveContextHash("app", "aa");
+        },
+      );
+      const { result } = renderHookWithProps();
+
+      let signPromise!: Promise<void>;
+      act(() => {
+        signPromise = result.current.handleSign();
+      });
+      await waitFor(() => expect(result.current.signing).toBe(true));
+
+      expect(result.current.canCancel).toBe(false);
+
+      await act(async () => {
+        settleAuth("cc".repeat(32));
+        await signPromise;
+      });
+    });
+
+    it("reports canCancel true during a hung signPsbts device window", async () => {
+      const cancelSigning = vi.fn();
+      mockBtcConnector = {
+        connectedWallet: {
+          account: { address: "tb1test" },
+          provider: {
+            signPsbt: vi.fn(),
+            signPsbts: vi.fn(() => new Promise<string[]>(() => {})),
+            cancelSigning,
+          },
+        },
+      };
+      mockSignAndSubmitPayouts.mockImplementation(
+        async ({
+          btcWallet,
+        }: {
+          btcWallet: { signPsbts: (hexes: string[]) => Promise<string[]> };
+        }) => {
+          await btcWallet.signPsbts(["p0", "p1"]);
+        },
+      );
+      const { result } = renderHookWithProps();
+
+      act(() => {
+        void result.current.handleSign();
+      });
+
+      await waitFor(() => expect(result.current.canCancel).toBe(true));
     });
   });
 
--- a/services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
+++ b/services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
@@ -45,6 +45,7 @@ import {
   shouldProbeWalletLiveness,
   verifyBtcWalletLiveness,
 } from "../../../utils/btc";
+import { supportsCancelSigning } from "../../../utils/cancelSigning";
 import { formatPayoutSignatureError } from "../../../utils/errors/formatting";
 import { isUserCancellation } from "../../../utils/errors/userCancellation";
 import { isVaultLifecycleStateError } from "../../../utils/errors/vaultLifecycleStateError";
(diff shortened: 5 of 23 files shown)


services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
```diff
@@ -709,30 +709,46 @@ describe("usePayoutSigningState", () => {
...
+          btcWallet: { signPsbt: (hex: string) => Promise<string> };
+          signal: AbortSignal;
+        }) => {
           pending.signal = signal;
           return new Promise<void>((resolve, reject) => {
             pending.resolve = resolve;
             pending.reject = reject;
+            // Settled only via `pending` — swallow the held-open device sign.
+            void btcWallet.signPsbt("payout-psbt").catch(() => {});
           });
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — nothing is suppressed; the fixture exposes reject and every test settles it and asserts the outcome

services/vault/src/copy.ts
```diff
@@ -797,6 +799,10 @@ export const COPY = {
         `Vault ${vaultNumber}: WOTS key submission failed - ${error}`,
       payoutSigningFailed: (vaultNumber: number, error: string) =>
         `Vault ${vaultNumber}: Payout signing failed - ${error}`,
+      // Self-requested device cancel: the loop stops here, so later vaults
+      // are left unattempted (no warning) rather than marked failed.
+      payoutSigningCancelled: (vaultNumber: number) =>
+        `Vault ${vaultNumber}: Payout signing canceled - you can finish signing when you're ready`,
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — the COPY entry is the named constant; shared constants are for dual-surface strings only

services/vault/src/hooks/deposit/__tests__/useDepositFlow.test.tsx
```diff
@@ -1911,6 +1914,154 @@ describe("useDepositFlow", () => {
...
+      // cancelled warning; vault 1 is unattempted, not failed (no warning).
+      expect(result.current.lastWarnings).toEqual([
+        {
+          vaultId: "0xVault0Id",
+          stage: "payout",
+          terminal: false,
+          message: COPY.deposit.warnings.payoutSigningCancelled(1),
+        },
+      ]);
+    });
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/__tests__/useDepositFlow.test.tsx
```diff
@@ -1911,6 +1914,154 @@ describe("useDepositFlow", () => {
...
+      const wallet = { ...MOCK_BTC_WALLET, signPsbt, cancelSigning: vi.fn() };
+      vi.mocked(signAndSubmitPayouts).mockImplementation(
+        async ({ btcWallet }) => {
+          await btcWallet.signPsbt("graphPsbt");
+        },
+      );
+
+      const { result } = renderHook(() =>
+        useDepositFlow({ ...MOCK_PARAMS, btcWalletProvider: wallet as any }),
+      );
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — matches this file's established mock pattern

services/vault/src/hooks/deposit/__tests__/useDepositFlow.test.tsx
```diff
@@ -1911,6 +1914,154 @@ describe("useDepositFlow", () => {
...
+        .mockResolvedValue("mockSignedPsbtHex");
+      const wallet = { ...MOCK_BTC_WALLET, signPsbt, cancelSigning: vi.fn() };
+      vi.mocked(signAndSubmitPayouts).mockImplementation(
+        async ({ btcWallet }) => {
+          await btcWallet.signPsbt("graphPsbt");
+        },
+      );
+
+      const { result } = renderHook(() =>
+        useDepositFlow({ ...MOCK_PARAMS, btcWalletProvider: wallet as any }),
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — matches this file's established mock pattern

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1345,6 +1370,21 @@ export function useDepositFlow(
...
+            if (deviceCancelSettledRef.current) {
+              recordWarning({
+                vaultId: result.vaultId,
+                stage: "payout",
+                terminal: false,
+                message: COPY.deposit.warnings.payoutSigningCancelled(
+                  result.vaultIndex + 1,
+                ),
+              });
+              break;
```
gbarkhatov, line 1385: **1 — [High] The cancel is undone a moment later: the continuation view auto-re-prompts the vault the user just cancelled.**

The loop stop works at the hook boundary, but the flow then returns *normally*, and the continuation path re-arms the same ceremony:

- `executeDeposit` resolves non-null (the new test asserts exactly this: `expect(resolved).not.toBeNull()`).
- `DepositSignContent.tsx:98` seeds `setContinuationVaultIds(result.pegins.map(...))` — **all** vault ids, cancelled one included.
- `DepositSignContent.tsx:177-192` switches to `PostDepositContinuationContent`.
- The cancelled vault submitted no payout signature, so it still polls as `PeginAction.SIGN_PAYOUT_TRANSACTIONS` → `isActionable` (`PostDepositContinuationView.tsx:154-157, 182`), and the view refetches on mount (`:145-147`).
- `PostDepositContinuationView.tsx:409-425` renders `ResumeSignContent`, whose `useRunOnce(handleSign)` (`ResumeDepositContent.tsx:125`) fires the ceremony — and `ensureAuthenticatedVpClient` runs it with `requireFreshDeviceCeremony: true`, so it is a *full* derive + re-approve.

`ResumeSignContent` already has the right instinct at `:127-148` — a quiet cancel routes to a re-offer instead of auto-running — but that guard reads `cancelRequested` from **its own** `usePayoutSigningState` instance. This is a freshly mounted one, so `sawCancelRequestRef.current === false` and the auto-run is not suppressed.

Net effect for the depositor: press "Cancel signing", and the device lights up again for the same vault seconds later. That also makes the new warning copy (`copy.ts:805`, "you can finish signing when you're ready") a promise the UI immediately breaks.

The hook-level test at `useDepositFlow.test.tsx:1970` is correct about what it asserts and doesn't reach this seam. Worth either carrying the cancel into the continuation handoff (so the cancelled vault mounts in the re-offer state rather than auto-running), or excluding it from `continuationVaultIds`.

packages/babylon-ledger-vault-signer/src/errors.ts
```diff
@@ -169,36 +169,26 @@ export class LedgerSignPsbtProtocolError extends Error {
...
- *   awaiting CONTINUE; the next non-CONTINUE APDU is eaten with 0x6A80 and the
- *   interrupted handler unwinds (`base:dispatcher.c:107-111`) — a direct retry
- *   uses `resendOnceOnIncorrectData`.
- * - A validation-phase abandonment leaves the intent intact; a Payout/NoPayout
- *   SIGNING-phase abandonment invalidates it (`fw:sign_custom_inputs.c:207-344`)
- *   — post-abort "intent-loaded" is a HINT only, and the retry path must also
- *   absorb 0xB007.
+ *   awaiting CONTINUE and eats the next non-CONTINUE APDU with 0x6A80
+ *   (`base:dispatcher.c:107-111`) — a fast retry's first DERIVE APDU surfaces
+ *   as the generic invalid-data error; the second attempt is clean.
```
gbarkhatov, line 179: **2 — [High] The post-cancel 0x6A80 is documented but not handled, and it surfaces as "contact support".**

Deleting the unused recovery is right — `loopAbandoned` / `resendOnceOnIncorrectData` / `dispatcherInterrupted` are gone tree-wide and every construction site matches the new 1-arg ctor. But this PR is what makes host aborts *button-reachable*: after `cancelSigning` the resume modal returns to a clickable Sign, so the ~5 s window is now routine rather than theoretical.

This doc says the fast retry "surfaces as the generic invalid-data error". Traced end to end, it is worse than that:

- retry's first APDU is `DERIVE_CONTEXT_HASH` (`requireFreshDeviceCeremony: true` releases the token first)
- device answers 0x6A80 → `classifyStatusWord` (`rawApdu.ts:134-145`) has no special case → generic `LedgerDeviceError`
- `toSignerWalletError` (`provider.ts:1141-1146`) — `isLedgerDeviceError` true, `statusWord !== SW_CLA_NOT_SUPPORTED` → `ERROR_CODES.UNKNOWN_ERROR`
- vault side: `formatPayoutSignatureError` matches no bucket → **`PSE.unexpected`**: *"An unexpected error occurred while signing payouts. Please try again or contact support."*

That is a support-desk message for a state that self-heals in five seconds. `PSE.deviceCeremonyInvalid` is the accurate copy and already sits a few lines away in the same switch — mapping this one status word onto `DEVICE_CEREMONY_INVALID` costs nothing and finishes the honest-copy goal this PR sets out. No test covers the path now that T2/T3b/T4/T5 are gone.

services/vault/src/copy.ts
```diff
@@ -881,6 +887,14 @@ export const COPY = {
         title: "Signing rejected",
         body: "You rejected the request in your wallet. Click Retry to approve it and continue.",
       },
+      // Self-requested cancel (the in-app "Cancel signing" affordance), as
+      // opposed to a rejection on the wallet/device itself. Names no button:
+      // this surface renders only Close. True on every cancellable window —
+      // the Pre-PegIn broadcast happens only after its signature completes.
+      signingCancelled: {
+        title: "Signing canceled",
+        body: "You canceled the signature request, so the deposit did not continue. No Bitcoin was spent.",
```
gbarkhatov, line 896: **3 — [Medium] "the deposit did not continue. No Bitcoin was spent." is wrong once registration has landed.**

This copy is reached for a cancel at the Pre-PegIn signature, which runs *after* the ETH batch registration is mined, depth-confirmed and key-verified, and after `addPendingPegin` has written PENDING rows. Gas is spent and N vaults sit registered on chain awaiting a Pre-PegIn that never arrives — "the deposit did not continue" reads as "nothing happened", which is the one thing that isn't true.

The justifying comment above (`:892-893`) also states the wrong invariant. "True on every cancellable window — the Pre-PegIn broadcast happens only after its signature completes" is not what makes this safe; what makes it safe is that a *rejected* sign short-circuits the broadcast. The successful-sign case goes the other way (see 4).

Suggest a second variant for the post-registration cancel that says what was left behind and how to resume, and leaving this one for the pre-registration windows.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -320,12 +330,20 @@ export function useDepositFlow(
       setDeviceSignActive(true);
       try {
         return await sign();
+      } catch (error) {
+        // The requested cancel took effect (a late cancel that still signed
+        // successfully deliberately does NOT stick — the flow proceeds).
+        if (deviceCancelRequestedRef.current && isUserCancellation(error)) {
```
gbarkhatov, line 336: **4 — [Medium] A cancel that races a *successful* Pre-PegIn signature still broadcasts.**

The drop of a late cancel is deliberate and almost everywhere costs one wasted ceremony. The `BROADCAST_PRE_PEGIN` window is the exception: `broadcastPrePeginTransaction` (`:928-943`) signs *and* broadcasts inside one call, with this `runCancellableSign` wrapping only the `signPsbt`.

Nothing converts a completed-but-cancelled sign into a stop:

- `signPsbtLoop` exits its `while` on a terminal status word without re-checking `opts.signal.aborted`, so a cancel arriving during the final device round still returns the signature
- `withSignAbort` (`provider.ts:733-741`) has no post-call abort check — `signPsbts` guards *between elements* (`:698-711`), but the Pre-PegIn is a single `signPsbt`, so there is no between
- this catch only fires on rejection, and no check sits between the sign returning and the broadcast

So the abort window is the whole final round-trip, including the on-device approval screen — the user taps Cancel in the app, approves on the device out of habit, and the BTC goes out with no notice. An `if (deviceCancelRequestedRef.current) throw` between sign and broadcast would cost nothing (an unbroadcast signed tx is harmless), or disable the button once the signature returns.

services/vault/src/utils/errors/causeChain.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * The single `cause`-chain walk shared by the error classifiers
+ * (userCancellation.ts, deviceErrors.ts, walletMethodNotSupported.ts,
+ * formatting.ts) so their depth bound and walk semantics cannot drift apart.
```
gbarkhatov, line 4: **5 — [Medium] This header names four sharers, but `formatting.ts` kept its own walk — with the old guard.**

`formatting.ts` imports only `MAX_CAUSE_DEPTH`; `classifyError` still runs its own loop guarded `cur && typeof cur === "object"` (`formatting.ts:141-146`). That is precisely the guard this PR removed from `isUserCancellation` because it stopped at a bare-string `cause`.

So the two now disagree: `isUserCancellationFrame` tests a string frame (`userCancellation.ts:95`) and `chainMatchesFrame` hands it one, while `classifyError` never descends to it. A wrapper `Error` whose `cause` is a bare `"user rejected"` string is a cancellation to one and unclassified to the other.

That directly contradicts `classifyError`'s own load-bearing note — *"Shared with the telemetry-side drop so the two cannot drift"* — and the two Sentry drops diverge with it. Either route `classifyError` through `chainMatchesFrame` too, or drop `formatting.ts` from this list and say why it is exempt.

(Also: `causeChain.ts` and `cancelSigning.ts` both land without direct tests. The cause-walk semantics are the kind of thing worth pinning, especially the string-frame case that motivated the change.)

packages/babylon-ledger-vault-signer/src/envelope.ts
```diff
@@ -36,8 +36,10 @@ function requireIntInRange(field: string, value: number, lo: number, hi: number)
  * @throws {DepositTermsRejectedError} on the first violation
  */
 export function assertDepositTermsDeviceCompatible(terms: DepositTerms): void {
-  // No tx-graph version check (2026-08-23 decision): the WASM capability gate
-  // stops unconstructable versions upstream; bad shapes fail closed on-device.
+  // No tx-graph version gate: v3 is Core 2's Bitcoin tx shape byte-for-byte (btc-vault
+  // e1e50f66; SDK parity vector pegin.test.ts "builds the v3 Pre-PegIn byte-identical to
+  // the v2 vector"); v1 is unreachable fresh (ProtocolParams setter rejects
+  // `newVersion <= prev`, and no Ledger v1 vault exists); v4+ needs a new WASM release.
```
gbarkhatov, line 42: **6 — [Medium] `vaultCoreVersion` is still entirely ungated on a critical-path file.**

The new rationale is a real improvement on the one it retires, and the parity citation checks out. The residue is that this is a §7 critical-path seam where the field crosses to a hardware signer with no bound at all — `envelope.test.ts:74-75` pins that by asserting versions **1, 2, 3 and 4** all pass.

The two admitted edges rest on operational claims rather than invariants:

- **v1** — "ProtocolParams setter rejects `newVersion <= prev`, and no Ledger v1 vault exists". Not verifiable in this repo: the SDK exposes only `ProtocolParamsReader` (`clients/eth/types.ts:266-268`), read-only, no setter. And the resume path rebuilds terms from chain, so the version is externally sourced.
- **v4+** — "needs a new WASM release", which is a release-ordering assumption, not a check here.

`assertVaultCoreVersionSupported` (`payoutSigning.ts:82`) is upstream but gates on what the *WASM* can build, not on what the *device* accepts — a different question.

`requireIntInRange("vaultCoreVersion", terms.vaultCoreVersion, 2, 3)` turns both into an invariant and forces the question on the next WASM bump, which is exactly the review posture §7 asks for.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -335,6 +356,8 @@ export function usePayoutSigningState({
                 },
               }
             : {}),
+          signMessage: (message, type) =>
```
gbarkhatov, line 359: **7 — [Medium] This `signMessage` wrapper is unreachable, and it makes the two hooks disagree.**

`graphProgressWallet` feeds only `signAndSubmitPayouts`, whose entire path is `prepareSigningContext` → `assertVaultCoreVersionSupported` → `ensureAuthenticatedVpClient` (`deriveContextHash`) → `runDepositorPresignFlow` (`signPsbt` / `signPsbts`). No `signMessage`. The SDK's only `signMessage` caller is `PeginManager.ts:1809` (proof-of-possession), which is on the registration path and never on resume.

The sibling wrapper in the deposit flow — `postBroadcastBtcWallet` (`useDepositFlow.ts:1055-1135`) — feeds the *same* `signAndSubmitPayouts` and wraps `signPsbt`/`signPsbts` only. So the two now differ on a method neither can reach, which is the kind of asymmetry that reads as intentional later.

Nothing covers it either: the two new tests exercise `deriveContextHash` and `signPsbts`. Under the zero-dead-code policy, drop it — or leave a line naming the caller it is waiting for.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1345,6 +1370,21 @@ export function useDepositFlow(
             // If the user cancelled, stop immediately — don't continue with other vaults
             if (signal.aborted) throw error;
 
+            // A settled self-cancel is a stop, not a per-vault failure: mark
+            // THIS vault cancelled and end the loop. Routine drop-off — no
+            // partial-failure telemetry.
+            if (deviceCancelSettledRef.current) {
```
gbarkhatov, line 1376: **8 — [Low] The cancel branch doesn't reset `perVaultSteps[vi]`, unlike the timeout branch immediately below.**

`isPayoutReadinessTimeout` at `:1388-1397` walks the vault back to `AWAIT_PAYOUT_TRANSACTIONS` before continuing. This branch records the warning and breaks with `perVaultSteps[vi]` left wherever the last `onProgress` put it — typically `SIGN_DEPOSITOR_GRAPH`. The cancelled vault then renders as still mid-signing next to a warning saying signing was cancelled.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1276,6 +1295,12 @@ export function useDepositFlow(
 
           signal.throwIfAborted();
 
+          // A settled user cancel stops the loop: the next iteration would
+          // re-run the full device ceremony (requireFreshDeviceCeremony)
+          // seconds after the user asked to stop. Remaining vaults are left
+          // unattempted (no warning), not failed.
+          if (deviceCancelSettledRef.current) break;
```
gbarkhatov, line 1302: **9 — [Low] This guard is unreachable, and the test named for it exercises the catch instead.**

The flag can never survive into a later iteration:

- it is reset per run at `:400`
- the catch branch at `:1376` always `break`s, so it never sets the flag and continues
- no earlier stage can set it and reach here — the pre-broadcast sites (`:531`, `:548`, `:665`, `:932`) propagate straight out of `executeDeposit`, and the WOTS loop is the only swallowing catch but has no cancellable window (`submitWotsPublicKey` reaches the wallet only through `deriveContextHash`, which is deliberately *not* wrapped in `runCancellableSign`)

Harmless defence-in-depth, and the comment is a good description of the hazard. Worth knowing that "the next vault gets no ceremony" is actually proven by the catch path at `:1385`, not by this line — so if the catch were ever changed to `continue`, the test would still pass.

packages/babylon-ledger-vault-signer/src/errors.ts
```diff
@@ -169,36 +169,26 @@ export class LedgerSignPsbtProtocolError extends Error {
...
   /** Yields the device had already delivered when the host stopped sending. */
   readonly yieldedCount: number;
-  /**
-   * True only when the abort left a dispatcher mid-interruption (device
-   * awaiting CONTINUE) — the one state the resend-once recovery exists for.
-   */
-  readonly dispatcherInterrupted: boolean;
 
-  constructor(yieldedCount: number, dispatcherInterrupted: boolean) {
+  constructor(yieldedCount: number) {
```
gbarkhatov, line 188: **10 — [Low] Breaking public-API change shipping as a patch.**

`LedgerSignPsbtAbortedError` loses a public readonly field and drops from 2 ctor args to 1, in `@babylonlabs-io/ledger-vault-signer` at a published `0.5.0`, under a `fix(...)` subject with no `!` and no `BREAKING CHANGE:` footer — so nx conventional-commits will read it as a patch.

Every in-repo construction site is updated, so nothing breaks here. Flagging in case the release process expects the marker; if versioning is handled by a separate release PR, ignore.

services/vault/src/copy.ts
gbarkhatov, line 805: **11 — [Nit] "cancelled" keys against "canceled" strings, plus two stale echoes.**

`copy.ts` calls for American English and the new strings follow it, but the keys went the other way — `payoutSigningCancelled` here and `signingCancelled` at `:894` both hold "canceled" values. Harmless, just worth picking one.

Two spots still carry the old spelling as *content*, which is the part that can mislead later:

- `useDepositFlow.ts:1539` — the `cancelDeviceSign` comment describes the outcome as the `"Signing cancelled"` callout; the copy now reads "Signing canceled".
- `usePayoutSigningState.test.tsx:763` and `useDepositFlow.test.tsx:1545` still construct `"Signing cancelled after 0 of N PSBT(s) — ..."`, a message `provider.ts:707/935` no longer emits. The fixtures still work (they carry `CONNECTION_REJECTED`, which is what `isTypedUserRejectionFrame` matches), so nothing fails — which is exactly why the drift will persist.

packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
```diff
@@ -39,7 +39,6 @@ const P2_CONTINUE = 0x00;
 const SW_OK = 0x9000;
 /** A client-command request follows in the response data (`base:sw.h:90`). */
 const SW_INTERRUPTED_EXECUTION = 0xe000;
```
gbarkhatov, line 41: **12 — [Nit] Module header still pins `dispatcher.c` after its only citation left this file.**

Removing `SW_INCORRECT_DATA` and the resend branch took the last `dispatcher.c` reference out of `signPsbtLoop.ts`, but the header at `:12-15` still lists it among the load-bearing pinned paths (`io_ext.h`, `sw.h` and `dispatcher.c`). The equivalent pointer in `rawApdu.ts` was correctly retargeted to `sw.h` in this same PR; this one was missed. The behaviour it described now lives in `errors.ts:178`.

gbarkhatov (approved):
Reviewed the full diff (23 files, +583/-274) across `96eed7356` and `03a39d834`, and traced each claim to source rather than to the doc comments. Approving — the core change is right and the removals are clean — but **1** and **2** both undercut the PR's own stated goal and I'd land them before merge.

**1** is the significant one: the payout loop does stop, but the flow returns normally and the continuation view auto-remounts a full device ceremony for the vault the user just cancelled. The fix holds at the hook boundary and comes undone one component later.

Findings are numbered 1-12 as inline comments, roughly by severity:

| # | File | |
|---|---|---|
| 1 | `useDepositFlow.ts:1385` | High — cancel undone by the continuation view's auto-re-prompt |
| 2 | `errors.ts:179` | High — post-cancel retry surfaces "contact support" |
| 3 | `copy.ts:896` | Medium — "No Bitcoin was spent" wrong post-registration |
| 4 | `useDepositFlow.ts:336` | Medium — cancel racing a successful Pre-PegIn sign still broadcasts |
| 5 | `causeChain.ts:4` | Medium — `formatting.ts` never joined the shared walk |
| 6 | `envelope.ts:42` | Medium — `vaultCoreVersion` still ungated |
| 7 | `usePayoutSigningState.ts:359` | Medium — unreachable `signMessage` wrapper |
| 8 | `useDepositFlow.ts:1376` | Low — cancel branch skips the `perVaultSteps` reset |
| 9 | `useDepositFlow.ts:1302` | Low — unreachable guard; its test covers the catch |
| 10 | `errors.ts:188` | Low — breaking API change as a patch |
| 11 | `copy.ts:805` | Nit — cancelled/canceled drift + two stale echoes |
| 12 | `signPsbtLoop.ts:41` | Nit — header still pins `dispatcher.c` |

**Verified clean.** `resendOnceOnIncorrectData` / `dispatcherInterrupted` / `loopAbandoned` are gone repo-wide with every construction site on the new 1-arg ctor. The `chainMatchesFrame` / `chainCarriesCode` extraction is behaviour-identical to the three walks it replaces — frame-first vs guard-first ordering coincides because `chainCarriesCode`'s predicate re-checks `typeof frame === "object"` — and `causeChain.ts` is genuinely import-free, so the `sentry.client.config.ts` init-time constraint holds. The `deviceWindowActive` narrowing is correct: only `signPsbt`/`signPsbts`/`signMessage` route through `withSignAbort` (`provider.ts:653/672/982`), so `deriveContextHash` and `approveDepositTerms` really are dead-button windows. The multi-vault stop is fail-fast — no swallowing `catch` in `signDepositorGraph` or `runDepositorPresignFlow` — and `withDeviceWindow` / `runCancellableSign` being non-reentrant is safe because `signPsbtsWithFallback` is sequential. No test helpers were orphaned by the deleted T2/T3b/T4/T5 cases.

The CodeRabbit comments look declinable to me: the named-constant ones misread the boundary (`copy.ts` governs the vault app, not packages, and `provider.ts` carries ~35 inline messages against 1 hoisted constant), and the `armPendingSdkCall` one points near a real seam but misdiagnoses it — 7 is the accurate version of that concern.

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/errors.ts
jrwbabylonlab: not planning — self-heals in ~5s

services/vault/src/copy.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: fixed

services/vault/src/utils/errors/causeChain.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/envelope.ts
jrwbabylonlab: fixed

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: not planning — next release will be 0.7.0

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: not planning — kept as a backstop; comment now says so

packages/babylon-ledger-vault-signer/src/errors.ts
jrwbabylonlab: not planning — no code change; next release goes out as 0.7.0

services/vault/src/copy.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
jrwbabylonlab: fixed

PR #2302, opened 2026-08-24
fix(core-ui, vault): update Telegram footer links

jonybur:
## Summary

- Update the Vault footer Telegram URL.
- Restore the Telegram link in the shared footer used by simple-staking.
- Use `https://t.me/babylonofficialcommunity` in both applications.

### DIFF
--- a/services/vault/src/config/socialLinks.ts
+++ b/services/vault/src/config/socialLinks.ts
@@ -8,7 +8,7 @@ export const SOCIAL_LINK_URLS = {
   github: "https://github.com/babylonlabs-io",
   // Matches the org convention used elsewhere (countdown, tbv-faucet,
   // coming-soon) — babylonlabs_io is an unrelated ~200-subscriber channel.
-  telegram: "https://t.me/babyloncommunity",
+  telegram: "https://t.me/babylonofficialcommunity",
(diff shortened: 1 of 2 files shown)


services/vault/src/config/socialLinks.ts
```diff
@@ -8,7 +8,7 @@ export const SOCIAL_LINK_URLS = {
   github: "https://github.com/babylonlabs-io",
   // Matches the org convention used elsewhere (countdown, tbv-faucet,
   // coming-soon) — babylonlabs_io is an unrelated ~200-subscriber channel.
-  telegram: "https://t.me/babyloncommunity",
+  telegram: "https://t.me/babylonofficialcommunity",
```
jrwbabylonlab, line 11: t.me/babylonofficialcommunity has 254 members and a description copied from t.me/babyloncommunity (~97k members, "Official Babylon Public Community") — same pattern as the babylonlabs_io trap the comment above warns about. Can you confirm with marketing that the new group is ours before this merges? The justification comment above is stale now either way.

jrwbabylonlab (approved):

kkkk666:
Correct, thanks for the fix.

PR #2305, opened 2026-08-24
feat(vault): let depositors adjust pre-pegin BTC fee rate

### LINKED ISSUE #2304: Vault deposit: allow adjusting BTC network fee rate (pre-pegin)
## Summary

Depositors cannot override the BTC network fee rate (sat/vB) used when building/broadcasting the **pre-pegin funding transaction**. The vault UI uses mempool `fastestFee` only. During congestion (or when mempool estimates lag / under-estimate), pegins can sit unconfirmed with no UI/wallet path to raise the fee.

## Motivation

- Observed on signet: stuck pre-pegin ~1 hour; no fee bump control in the vault web UI.
- Fee is already mempool-derived (not hardcoded), but refresh can lag and estimates can be wrong.
- Especially important for phased mainnet launch when vault slots are limited and timely confirmation matters.
- In-repo **references** (not prescribed UX): simple-staking `FeeModal` (Fast/Medium/Slow/Custom) and vault refund `FeeRateField`.

## Proposed Solution

MVP — deposit form only:

1. Show current BTC network fee rate and allow override (exact UI TBD with design; see references below).
2. Thread the chosen rate through estimate → `setFeeRate` → `mempoolFeeRate` → prepare/sign/broadcast.
3. Recalculate estimated fee, max deposit, and CTA gating when the rate changes.
4. Default remains mempool `fastestFee`.
5. Reject non-positive rates; agree min/max bounds with design (staking’s `getFeeRateFromMempool` is one reference for bounds).

Do **not** conflate with protocol `minPeginFeeRate` / `protocolFeeRate` (offchain params). Target **mempool / funding** rate only.

Ledger: `prepeginMaxFee` is the funded tx fee at terms-build — compatible if terms are built after the user-selected rate.

## UI references (optional precedent)

Vault deposit does not have a dedicated fee-adjust design yet. These are useful precedents — **not** a requirement to copy either pattern:

| Precedent | Where |
|-----------|--------|
| Simple-staking Advanced Fee Settings | `services/simple-staking/.../Staking/FeeModal/` + `StakingFeesSection.tsx` |
| Vault refund inline sat/vB | refund `FeeRateField` |
| Staking fee bounds helper | `getFeeRateFromMempool` (min=`hourFee`, default=`fastestFee`, max=`max(128, nextPowerOfTwo(fastestFee))`) |

Final deposit UX is a product/design call (Alina / Figma).

## Out of scope

- RBF / bump of an already-broadcast stuck pre-pegin (separate CPFP / speed-up story)
- Changing on-chain / offchain protocol fee params
- ETH gas / protocol deposit fee
- Relying on external wallet fee UI alone

## Nice-to-have (follow-up)

- Withdraw flow fee-rate override (today display-only)
- Expose Fast/Medium/Slow tiers from full mempool payload in vault `useNetworkFees` if UI needs presets

## Acceptance Criteria

- [ ] User can change sat/vB on the deposit form before signing
- [ ] Custom rate is used in the built pre-pegin funding fee
- [ ] Fee estimate + max deposit recalculate on rate change
- [ ] Invalid rates blocked in UI
- [ ] Ledger deposit terms `prepeginMaxFee` matches funded fee after custom rate
- [ ] Refund fee edit still works (no regression)

## Implementation sketch

Primary: `babylon-toolkit/services/vault` — parameterize `useEstimatedBtcFee`, deposit form edit entry (`DepositState.setFeeRate` already exists), no silent overwrite of user rate on mempool refetch, tests for custom `mempoolFeeRate`.

## Notes

- Investigation scoped in Nexus worktree `feat/adjust-feerate` (`FEATURE.md` / `handoff.md`).
- No prior dedicated GH issue found (2026-08-24). Confluence “Web Priority ATM” mentioned in Slack — confirm/link Jira if one exists.
- Design: decide vault deposit fee UX; staking FeeModal and refund `FeeRateField` are references only.

jeremy-babylonlabs:
## Summary
- The deposit progress card now carries the designed inline **Network Fee Rate** selector in its pre-sign state — Slow / Avg / Fast cards fed by live mempool tiers, plus a Custom sat/vB input with an estimated-fee readout and a clear button. The rate the depositor picks is the rate the pre-pegin funding tx is built with.
- Slow/Avg/Fast track their mempool tier across the 60s refetch; a Custom value stays frozen until edited. Custom values are bounded by the minimum relay fee and simple-staking's `getFeeRateFromMempool`-style cap (`max(128, nextPowerOfTwo(fastestFee))`); an out-of-range value or a failing fee estimate disables Sign Transaction.
- `useEstimatedBtcFee` takes the chosen rate as an override feeding both the fee estimate and the max-deposit ceiling. The form's fee breakdown keeps read-only Network Fee Rate and BTC Network Fee rows.

Closes #2304

## Test plan
- [ ] Open deposit form; fee breakdown shows Network Fee Rate + BTC Network Fee (read-only)
- [ ] Continue to the signing card; the fee selector renders below the steps with live tier rates
- [ ] Pick Slow/Avg/Fast; header rate updates; leave it open past a mempool refetch (60s) — selection holds, tier values update
- [ ] Custom: type a rate; estimated BTC fee updates; out-of-range value disables Sign; ✕ returns to Fast
- [ ] Sign; confirm the built pre-pegin uses the selected rate
- [ ] `pnpm --filter @services/vault exec vitest run src/components/simple/ src/hooks/deposit/ src/utils/__tests__/feeRateBounds.test.ts`

### DIFF
--- a/services/vault/src/components/simple/DepositFeesBreakdown.tsx
+++ b/services/vault/src/components/simple/DepositFeesBreakdown.tsx
@@ -32,6 +32,10 @@ interface DepositFeesBreakdownProps {
    * `undefined` while a two-vault split's per-vault amounts are loading.
    */
   commissionBaseValues?: readonly bigint[];
+  /** BTC network fee rate (sat/vB) used for the pre-pegin funding tx. */
+  networkFeeRate?: number;
+  /** Estimated funding fee in satoshis at {@link networkFeeRate}. */
+  networkFeeSats?: bigint | null;
 }
 
 export function DepositFeesBreakdown({
@@ -44,6 +48,8 @@ export function DepositFeesBreakdown({
   amountSats,
   commissionBps,
   commissionBaseValues,
+  networkFeeRate,
+  networkFeeSats,
 }: DepositFeesBreakdownProps) {
   // Format a satoshi value as a BTC amount plus an optional "($X USD)" suffix,
   // matching the existing fee-line presentation. `null` sats render as the
@@ -67,6 +73,10 @@ export function DepositFeesBreakdown({
 
   const transactionReserve = formatSatsLine(
     depositorClaimValue === undefined ? null : depositorClaimValue,
+  );
+
+  const networkFee = formatSatsLine(
+    networkFeeSats === undefined ? null : networkFeeSats,
   );
 
   // The protocol charges commission on the vault deposit amount only —
@@ -98,8 +108,24 @@ export function DepositFeesBreakdown({
       ? FORM_COPY.vpCommissionLabel
       : `${FORM_COPY.vpCommissionLabel} (${formatBasisPointsAsPercent(commissionBps)})`;
 
+  const networkFeeRateValue =
+    networkFeeRate !== undefined && networkFeeRate > 0
+      ? `${networkFeeRate} sats/vB`
+      : METRIC_PLACEHOLDER;
+
   return (
     <div className="flex flex-col gap-2">
+      <FeeDetailRow
+        label={FORM_COPY.networkFeeRateLabel}
+        tooltip={FORM_COPY.networkFeeRateTooltip}
+        value={networkFeeRateValue}
+      />
+      <FeeDetailRow
+        label={FORM_COPY.btcNetworkFeeLabel}
+        tooltip={FORM_COPY.btcNetworkFeeTooltip}
+        value={networkFee.amount}
+        secondaryValue={networkFee.price}
+      />
       <FeeDetailRow
         label={FORM_COPY.transactionReserveLabel}
         tooltip={FORM_COPY.transactionReserveTooltip}
--- a/services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx
@@ -161,6 +161,10 @@ export interface DepositProgressViewProps {
   cancelSigningRequested?: boolean;
   /** Requests cancellation of the in-flight device signing ceremony. */
   onCancelSigning?: () => void;
+  /** Inline fee-rate panel rendered in the pre-sign entry state (`!started`). */
+  preSignFeeSelector?: ReactNode;
+  /** Disables the pre-sign entry CTA (e.g. an invalid custom fee rate). */
+  signDisabled?: boolean;
 }
 
 /**
@@ -265,6 +269,8 @@ export function DepositProgressView(props: DepositProgressViewProps) {
     canCancelSigning = false,
     cancelSigningRequested = false,
     onCancelSigning,
+    preSignFeeSelector,
+    signDisabled = false,
   } = props;
 
   // Every flow that renders this view requires the BTC wallet, so surface a
@@ -481,7 +487,7 @@ export function DepositProgressView(props: DepositProgressViewProps) {
                   ? cancelSigningRequested
                   : started
                     ? !canClose && !isTerminalSuccess
-                    : false
+                    : signDisabled
             }
             variant="contained"
             color="secondary"
@@ -570,6 +576,8 @@ export function DepositProgressView(props: DepositProgressViewProps) {
             started={started}
           />
         )}
+
+        {!started && preSignFeeSelector}
 
         {/* Persist through errors: a retry still needs signing, so the nudge
             stays useful. Only a finished deposit (complete / terminal success)
--- a/services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx
@@ -161,7 +161,11 @@ export interface DepositProgressViewProps {
   cancelSigningRequested?: boolean;
   /** Requests cancellation of the in-flight device signing ceremony. */
   onCancelSigning?: () => void;
-  /** Inline fee-rate panel rendered in the pre-sign entry state (`!started`). */
+  /**
+   * Inline fee-rate panel for the pre-sign entry state (`!started`). Rendered
+   * inside the "Register deposit" card, under the Pre-PegIn step whose
+   * transaction the rate pays for.
+   */
   preSignFeeSelector?: ReactNode;
   /** Disables the pre-sign entry CTA (e.g. an invalid custom fee rate). */
   signDisabled?: boolean;
@@ -351,8 +355,9 @@ export function DepositProgressView(props: DepositProgressViewProps) {
   // a current group with none of its own work done reads not-started, so
   // nothing spins or announces progress while the flow idles awaiting the
   // click. Flows entering at step 1 have nothing completed, so they render
-  // exactly as before: no bar, no pill, every group a collapsed not-started
-  // header.
+  // as no bar, no pill, and collapsed not-started headers — except the group
+  // owning the Pre-PegIn step, which opens to carry the fee selector on its
+  // one relevant row (see GroupBlock).
   const visualStep = isComplete
     ? TOTAL_VISUAL_STEPS + 1
     : getVisualStep(currentStep);
@@ -566,6 +571,7 @@ export function DepositProgressView(props: DepositProgressViewProps) {
             renderStepDetail={renderStepDetail}
             perVaultSteps={perVaultSteps}
             started={started}
+            preSignDetail={started ? undefined : preSignFeeSelector}
           />
         ) : (
           <GroupedProgress
@@ -574,10 +580,9 @@ export function DepositProgressView(props: DepositProgressViewProps) {
             activeStepDetail={activeStepDetail}
             hasError={Boolean(error)}
             started={started}
+            preSignDetail={started ? undefined : preSignFeeSelector}
           />
         )}
-
-        {!started && preSignFeeSelector}
 
         {/* Persist through errors: a retry still needs signing, so the nudge
             stays useful. Only a finished deposit (complete / terminal success)
--- a/services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
@@ -1,0 +1,290 @@
+/**
+ * FeeRateSelector
+ *
+ * Inline network fee-rate picker rendered in the Deposit Progress view's
+ * pre-sign entry state (replaces the old DepositFeeModal). Owns its own
+ * selection state, seeded once from the committed `feeRate` prop; a selected
+ * mempool tier live-follows its tier value on refetch, while a custom rate is
+ * frozen until the user edits it. See the parent spec for the full contract.
+ */
+
+import { Input, Text } from "@babylonlabs-io/core-ui";
+import { peginOutputCount } from "@babylonlabs-io/ts-sdk/tbv/core";
+import { useEffect, useMemo, useRef, useState } from "react";
+import { IoClose } from "react-icons/io5";
+
+import { getNetworkConfigBTC } from "@/config";
+import { MIN_RELAY_FEE_RATE_SATS_VB } from "@/constants";
+import { useBTCWallet } from "@/context/wallet";
+import { COPY } from "@/copy";
+import { useEstimatedBtcFee } from "@/hooks/deposit/useEstimatedBtcFee";
+import { useNetworkFees } from "@/hooks/useNetworkFees";
+import { useUTXOs } from "@/hooks/useUTXOs";
+import { satoshiToBtcNumber } from "@/utils/btcConversion";
+import { getFeeRateBounds } from "@/utils/feeRateBounds";
+import { formatBtcAmount } from "@/utils/formatting";
+
+const FEE_SELECTOR_COPY = COPY.deposit.feeSelector;
+const btcConfig = getNetworkConfigBTC();
+
+type TierKey = "slow" | "avg" | "fast" | "custom";
+
+interface FeeRateSelectorProps {
+  /** Per-vault deposit amounts (satoshis); sum is the total pre-pegin amount. */
+  vaultAmounts: bigint[];
+  /** Committed fee rate (sat/vB) — the DepositState value the sign flow uses. */
+  feeRate: number;
+  onFeeRateChange: (rate: number) => void;
+  /** False while the fee estimate errors, or custom is selected with no valid value. */
+  onValidityChange?: (canSign: boolean) => void;
+}
+
+function parseValidCustom(
+  value: string,
+  minFeeRate: number,
+  maxFeeRate: number,
+): number | null {
+  const parsed = Number.parseFloat(value);
+  if (!Number.isFinite(parsed)) return null;
+  if (parsed < minFeeRate || parsed > maxFeeRate) return null;
+  return parsed;
+}
+
+export function FeeRateSelector({
+  vaultAmounts,
+  feeRate,
+  onFeeRateChange,
+  onValidityChange,
+}: FeeRateSelectorProps) {
+  const {
+    defaultFeeRate: fastestFee,
+    halfHourFeeRate: avgFee,
+    hourFeeRate: slowFee,
+    isLoading,
+  } = useNetworkFees();
+  const { address } = useBTCWallet();
+  const { spendableMempoolUTXOs } = useUTXOs(address);
+
+  const totalAmountSats = useMemo(
+    () => vaultAmounts.reduce((sum, amount) => sum + amount, 0n),
+    [vaultAmounts],
+  );
+  const numOutputs = peginOutputCount(vaultAmounts.length, true);
+  const estimatedFee = useEstimatedBtcFee(
+    totalAmountSats,
+    spendableMempoolUTXOs,
+    numOutputs,
+    feeRate,
+  );
+
+  const [selectedKey, setSelectedKey] = useState<TierKey>("fast");
+  const [customValue, setCustomValue] = useState("");
+  const seededRef = useRef(false);
+
+  const { minFeeRate, maxFeeRate } = getFeeRateBounds({
+    defaultFeeRate: fastestFee,
+    hourFeeRate: slowFee,
+  });
+
+  // Seed the selection once, when the fee tiers first resolve, from the
+  // committed `feeRate` prop. After that this effect only live-follows: a
+  // selected tier's committed rate tracks the tier's current mempool value; a
+  // custom rate is frozen and only changes on user input. Never re-derive the
+  // selection itself from props or fee tiers again.
+  useEffect(() => {
+    if (isLoading) return;
+
+    let key = selectedKey;
+    let custom = customValue;
+
+    if (!seededRef.current) {
+      seededRef.current = true;
+      if (feeRate === slowFee) {
+        key = "slow";
+      } else if (feeRate === avgFee) {
+        key = "avg";
+      } else if (feeRate === fastestFee) {
+        key = "fast";
+      } else {
+        key = "custom";
+        custom = String(feeRate);
+      }
+      setSelectedKey(key);
+      setCustomValue(custom);
+    }
+
+    const tierRate =
+      key === "slow" ? slowFee : key === "avg" ? avgFee : fastestFee;
+    const derivedRate =
+      key === "custom"
+        ? parseValidCustom(custom, minFeeRate, maxFeeRate)
+        : tierRate;
+
+    if (derivedRate !== null && derivedRate !== feeRate) {
+      onFeeRateChange(derivedRate);
+    }
+  }, [
+    isLoading,
+    feeRate,
+    slowFee,
+    avgFee,
+    fastestFee,
+    selectedKey,
+    customValue,
+    minFeeRate,
+    maxFeeRate,
+    onFeeRateChange,
+  ]);
+
+  const lastValidityRef = useRef<boolean | null>(null);
+  useEffect(() => {
+    if (!onValidityChange) return;
+
+    const customInvalid =
+      selectedKey === "custom" &&
+      parseValidCustom(customValue, minFeeRate, maxFeeRate) === null;
+    const canSign = !customInvalid && !estimatedFee.error;
+
+    if (lastValidityRef.current === canSign) return;
+    lastValidityRef.current = canSign;
+    onValidityChange(canSign);
+  }, [
+    onValidityChange,
+    selectedKey,
+    customValue,
+    minFeeRate,
+    maxFeeRate,
+    estimatedFee.error,
+  ]);
+
+  const handleClearCustom = () => {
+    setCustomValue("");
+    setSelectedKey("fast");
+    onFeeRateChange(fastestFee);
+  };
+
+  const showLowFeeWarning = feeRate < slowFee;
+
+  const tiers: Array<{
+    key: TierKey;
+    label: string;
+    rate: number;
+    hint: string;
+  }> = [
+    {
+      key: "slow",
+      label: FEE_SELECTOR_COPY.slowLabel,
+      rate: slowFee,
+      hint: FEE_SELECTOR_COPY.slowHint,
+    },
+    {
+      key: "avg",
+      label: FEE_SELECTOR_COPY.avgLabel,
+      rate: avgFee,
+      hint: FEE_SELECTOR_COPY.avgHint,
+    },
+    {
+      key: "fast",
+      label: FEE_SELECTOR_COPY.fastLabel,
+      rate: fastestFee,
+      hint: FEE_SELECTOR_COPY.fastHint,
+    },
+  ];
+
+  return (
+    <div className="flex flex-col gap-2 rounded-lg border border-secondary-strokeLight bg-secondary-highlight p-3">
+      <div className="flex items-center justify-between">
+        <Text variant="body2" className="text-accent-primary">
+          {FEE_SELECTOR_COPY.title}
+        </Text>
+        <Text variant="body2" className="text-accent-primary">
+          {feeRate} {FEE_SELECTOR_COPY.headerUnit}
+        </Text>
+      </div>
+
+      <div className="grid grid-cols-4 gap-2.5">
+        {tiers.map((tier) => (
+          <button
+            key={tier.key}
+            type="button"
+            aria-pressed={selectedKey === tier.key}
+            onClick={() => setSelectedKey(tier.key)}
+            className={`flex flex-col items-center gap-1 rounded border p-2 text-center ${
+              selectedKey === tier.key
+                ? "border-accent-primary bg-accent-primary/10"
+                : "border-secondary-strokeLight"
+            }`}
+          >
+            <Text variant="body2">{tier.label}</Text>
+            <Text variant="caption">
+              <span className="text-accent-primary">
+                {tier.rate} {FEE_SELECTOR_COPY.cardUnit}
+              </span>{" "}
+              <span className="text-accent-secondary">{tier.hint}</span>
+            </Text>
+          </button>
+        ))}
+        <button
+          type="button"
+          aria-pressed={selectedKey === "custom"}
+          onClick={() => setSelectedKey("custom")}
+          className={`flex items-center justify-center rounded border p-2 ${
+            selectedKey === "custom"
+              ? "border-accent-primary bg-accent-primary/10"
+              : "border-secondary-strokeLight"
+          }`}
+        >
+          <Text variant="body2">{FEE_SELECTOR_COPY.customLabel}</Text>
+        </button>
+      </div>
+
+      {selectedKey === "custom" && (
+        <div className="flex flex-col gap-1">
+          <div className="flex items-center gap-2">
+            <Input
+              type="number"
+              min={MIN_RELAY_FEE_RATE_SATS_VB}
+              step={1}
+              value={customValue}
+              onChange={(e) => setCustomValue(e.currentTarget.value)}
+              suffix={
+                <Text as="span" variant="body2">
+                  {FEE_SELECTOR_COPY.customInputSuffix}
+                </Text>
+              }
+              wrapperClassName="flex-1"
+            />
+            <button
+              type="button"
+              aria-label={FEE_SELECTOR_COPY.clearCustomAria}
+              onClick={handleClearCustom}
+              className="flex size-10 shrink-0 items-center justify-center rounded border border-secondary-strokeLight"
+            >
+              <IoClose size={20} />
+            </button>
+          </div>
+          {!estimatedFee.error && estimatedFee.fee !== null && (
+            <Text variant="caption" className="text-accent-secondary">
+              {formatBtcAmount(satoshiToBtcNumber(estimatedFee.fee))}{" "}
+              {btcConfig.coinSymbol}
+            </Text>
+          )}
+        </div>
+      )}
+
+      {/* An estimate error blocks Sign for tier rates too, so it must be
+          visible regardless of which option is selected. */}
+      {estimatedFee.error && (
+        <Text variant="caption" className="text-error-main">
+          {estimatedFee.error}
+        </Text>
+      )}
+
+      {showLowFeeWarning && (
+        <Text variant="caption" className="text-warning-main">
+          {FEE_SELECTOR_COPY.lowFeeWarning}
+        </Text>
+      )}
+    </div>
+  );
+}
--- a/services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
@@ -13,7 +13,6 @@ import { peginOutputCount } from "@babylonlabs-io/ts-sdk/tbv/core";
 import { useEffect, useMemo, useRef, useState } from "react";
 import { IoClose } from "react-icons/io5";
 
-import { getNetworkConfigBTC } from "@/config";
 import { MIN_RELAY_FEE_RATE_SATS_VB } from "@/constants";
 import { useBTCWallet } from "@/context/wallet";
 import { COPY } from "@/copy";
@@ -25,7 +24,6 @@ import { getFeeRateBounds } from "@/utils/feeRateBounds";
 import { formatBtcAmount } from "@/utils/formatting";
 
 const FEE_SELECTOR_COPY = COPY.deposit.feeSelector;
-const btcConfig = getNetworkConfigBTC();
 
 type TierKey = "slow" | "avg" | "fast" | "custom";
 
@@ -202,21 +200,23 @@ export function FeeRateSelector({
         </Text>
       </div>
 
-      <div className="grid grid-cols-4 gap-2.5">
+      <div className="grid grid-cols-2 gap-2 sm:grid-cols-4">
         {tiers.map((tier) => (
           <button
             key={tier.key}
             type="button"
             aria-pressed={selectedKey === tier.key}
             onClick={() => setSelectedKey(tier.key)}
-            className={`flex flex-col items-center gap-1 rounded border p-2 text-center ${
+            className={`flex flex-col items-center gap-0.5 rounded border px-1.5 py-1.5 text-center ${
               selectedKey === tier.key
                 ? "border-accent-primary bg-accent-primary/10"
                 : "border-secondary-strokeLight"
             }`}
           >
-            <Text variant="body2">{tier.label}</Text>
-            <Text variant="caption">
+            <Text variant="caption" className="font-medium">
+              {tier.label}
+            </Text>
+            <Text variant="caption" className="whitespace-nowrap">
               <span className="text-accent-primary">
                 {tier.rate} {FEE_SELECTOR_COPY.cardUnit}
               </span>{" "}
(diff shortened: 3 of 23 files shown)


jonybur (dismissed):
The pre-sign fee-rate picker keeps the estimator/broadcast fee rate in agreement: the picker commits the user-selected sat/vB back into `DepositState.feeRate` via `onFeeRateChange`, and the sign flow forwards that same value to `useDepositFlow` as `mempoolFeeRate`. `useEstimatedBtcFee` prefers a positive override, otherwise the mempool default, so both the fee-line display and the broadcast site read the same rate. `peginOutputCount(vaultCount, true)` in `FeeRateSelector` matches `useDepositPageForm.ts:434`, so the vsize model doesn't drift. Custom-rate validation floors at `MIN_RELAY_FEE_RATE_SATS_VB` and caps at `max(128, nextPowerOfTwo(fastestFee))` — same shape as the simple-staking port cited in the header. Selected-tier live-follows the mempool refetch; custom stays frozen until edited; both invariants pinned by the new tests. `onValidityChange` gates the pre-sign Sign CTA through `signDisabled`, so an out-of-range custom entry or a fee-estimate error can't slip past.

services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
```diff
@@ -0,0 +1,290 @@
...
+  const { address } = useBTCWallet();
+  const { spendableMempoolUTXOs } = useUTXOs(address);
+
+  const totalAmountSats = useMemo(
+    () => vaultAmounts.reduce((sum, amount) => sum + amount, 0n),
+    [vaultAmounts],
+  );
+  const numOutputs = peginOutputCount(vaultAmounts.length, true);
+  const estimatedFee = useEstimatedBtcFee(
+    totalAmountSats,
```
gbarkhatov, line 72: **[blocking] The affordability gate measures the wrong amount.**

This estimates against `sum(vaultAmounts)` — the bare deposit. The transaction that actually gets funded is sized against `prePegin.totalOutputValue` in `PeginManager.ts:888-893`, i.e. `Σ (peginAmount + depositorClaimValue + p2aAnchorValue + minPeginFee) + CPFP anchor`. The gate is optimistic by the entire reserve stack.

Working the algebra through makes the gap exact. `Max` is set to `totalBalance − fee(fastestFee) − reserves − PRE_PEGIN_SAFETY_BUFFER_SATS` (`useDepositPageForm.ts:585-601`). Deposit at Max, then raise the rate:

- the **real build** needs `Max + reserves + fee(chosen) ≤ totalBalance`, which reduces to `fee(chosen) − fee(fastestFee) ≤ 3,000`;
- **this gate** only needs `Max + fee(chosen) ≤ totalBalance`, which reduces to `fee(chosen) − fee(fastestFee) ≤ reserves + 3,000`.

So the build starts failing at a 3,000-sat fee delta while Sign stays enabled until the delta also exceeds `reserves`. On a ~250 vB 2-vault Pre-PegIn, 5 → 128 sat/vB is roughly a 30k sat delta — comfortably past the buffer. The result is `Insufficient funds` thrown from `prepareSizing` mid-flow, after the gate said yes.

Related: `PRE_PEGIN_SAFETY_BUFFER_SATS` is documented (`useDepositPageForm.ts:50-58`) as sized to absorb "a ~10 sat/vB upward mempool spike in the click→broadcast window". This PR lets the depositor deliberately jump to ≥128 sat/vB in that same window, so that comment is now materially understated and should be revisited alongside the fix.

Either estimate against the same total the SDK will select against, or clamp `maxFeeRate` to what the balance actually affords at the committed amount.

services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx
```diff
@@ -571,6 +577,8 @@ export function DepositProgressView(props: DepositProgressViewProps) {
           />
         )}
 
+        {!started && preSignFeeSelector}
```
gbarkhatov: **[blocking] A fee-rate mistake is unrecoverable in place.**

The selector renders only while `!started`. `started` is set on the Sign click and never cleared (`DepositSignContent.tsx:83`), and `hasStartedRef` blocks re-entry. `DepositSignContent` never passes `onRetry`, so the error CTA falls back to its default of `onClose` → `resetAll()` → `feeRate = 0`.

Combined with the sizing gap in the `FeeRateSelector` comment, a depositor who picks a rate their balance can't fund cannot simply lower it — they have to close the modal and restart the whole deposit, re-entering the amount and re-selecting the provider. Since the fee rate is now user-controlled, the pre-sign state is worth making re-enterable on a pre-broadcast failure (or at minimum keeping the selector mounted through the error state).

services/vault/src/hooks/deposit/useEstimatedBtcFee.ts
```diff
@@ -20,45 +39,37 @@ export interface EstimatedBtcFeeResult {
...
+    feeRateOverride !== undefined &&
+    Number.isFinite(feeRateOverride) &&
+    feeRateOverride > 0
+      ? feeRateOverride
+      : defaultFeeRate;
+
   // Max deposit only depends on UTXOs + fee rate, not the user's amount
   const maxDeposit = useMemo(() => {
-    if (isLoading || defaultFeeRate === 0 || !utxos?.length) return null;
+    if (isLoading || feeRate === 0 || !utxos?.length) return null;
```
gbarkhatov, line 60: **[blocking] The hook now holds two disagreeing definitions of "ready".**

The result branch was relaxed to `isLoading && feeRate === 0` (line 72), but this `maxDeposit` memo still short-circuits on bare `isLoading`. With a usable override during a fee fetch the hook returns `isLoading: false` and a real computed `fee`, but `maxDeposit: null`.

That matters because `null` is not treated as "not ready" downstream — it's treated as "no ceiling". `amountExceedsMax` returns `false` when `maxDepositSats == null` (`validations.ts:199-203`), so the ceiling check is silently **skipped** rather than deferred, and `applyMaxAmount` no-ops.

Not reachable from today's only override caller (the form passes no override, and when it's loading `defaultFeeRate` is 0 anyway), so this is latent. But it's a trap for the next caller, in a file CLAUDE.md §2 lists as a critical path. The two guards should agree — presumably both on `feeRate === 0`.

services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
```diff
@@ -0,0 +1,290 @@
...
+  // selection itself from props or fee tiers again.
+  useEffect(() => {
+    if (isLoading) return;
+
+    let key = selectedKey;
+    let custom = customValue;
+
+    if (!seededRef.current) {
+      seededRef.current = true;
+      if (feeRate === slowFee) {
```
gbarkhatov, line 100: **Seeding by float equality against a live-refetching value has three distinct failure modes.**

The block binds a tier by comparing `feeRate` to values that change underneath it every 60s, which makes it fragile in ways the comment above doesn't anticipate:

1. **Tier-order collision.** `slowFee` is tested first. The SDK validates fee ordering as `hourFee ≤ halfHourFee ≤ fastestFee` (`mempoolApi.ts:493-499`) — equality is explicitly allowed, and on a quiet mempool all three are commonly equal. A committed *fastest* rate then binds to "Slow". Once the tiers diverge again, live-follow quietly walks the committed rate **down** to the hour fee, which is the opposite of what the depositor picked.
2. **Seed race → stray "Custom".** `SimpleDeposit.tsx:419` commits `setFeeRate(estimatedFeeRate)` at the form→sign transition; this seeds one render later. If the 60s refetch lands in between, `feeRate` matches no tier, and it seeds `key = "custom"` with a rate the depositor never typed — then freezes there.
3. **Runs with all-zero tiers on fetch error.** The only guard is `if (isLoading) return`, but `useNetworkFees` returns `isLoading: false` with all three rates `0` on error (`useNetworkFees.ts:47-53`). The cards render `0 sat/vB`, and `showLowFeeWarning` (line 166) is suppressed exactly when the rates are unknown.

There's also a fourth, smaller one from the same block: the tier buttons are clickable while `isLoading`, but a click made then is discarded — `setSelectedKey(key)` on line 112 overwrites it when the seed finally runs.

Seeding from an explicit tier identity rather than value equality (and gating the effect on the tiers being non-zero) would close all four.

services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
```diff
@@ -0,0 +1,290 @@
...
+          <Text variant="body2">{FEE_SELECTOR_COPY.customLabel}</Text>
+        </button>
+      </div>
+
+      {selectedKey === "custom" && (
+        <div className="flex flex-col gap-1">
+          <div className="flex items-center gap-2">
+            <Input
+              type="number"
+              min={MIN_RELAY_FEE_RATE_SATS_VB}
```
gbarkhatov, line 248: **An out-of-range custom rate disables Sign with no explanation.**

Validity flips false at line 143-150, but nothing renders for it — the only two conditional messages in the panel are the estimate error (line 277) and the low-fee warning (line 283). So typing `9999` leaves a greyed-out Sign button, an input showing `9999`, and the header on line 201 still reading the last *committed* rate — three surfaces disagreeing with no explanation of which one is authoritative.

The input also carries `min` but no `max`, so the bounds are never communicated even passively. And it's reachable without user error: clicking "Custom" from a seeded tier lands in the invalid state immediately (empty string → `Number.parseFloat` → `NaN`), as does a downward mempool refetch between submit and mount.

core-ui's `Input` already accepts `state="error"` (`packages/babylon-core-ui/src/components/Form/Input.tsx:12`), and `copy.ts:501` has a fee-rate-cap error string from the refund flow worth modelling the wording on.

services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
```diff
@@ -0,0 +1,290 @@
...
+  /** False while the fee estimate errors, or custom is selected with no valid value. */
+  onValidityChange?: (canSign: boolean) => void;
+}
+
+function parseValidCustom(
+  value: string,
+  minFeeRate: number,
+  maxFeeRate: number,
+): number | null {
+  const parsed = Number.parseFloat(value);
```
gbarkhatov, line 45: **Fractional rates are accepted, against the SDK's own contract.**

`Number.parseFloat` with no integer check means `1.5` commits straight into `mempoolFeeRate`. The SDK's own definition of a valid fee rate is `Number.isInteger(value) && value > 0 && value <= MAX_FEE_RATE` (`mempoolApi.ts:88-90`) — it rejects non-integers from the mempool API, then accepts one from us on the way back in.

Nothing underfunds (the vsize model over-estimates and `Math.ceil`s), so this isn't a value bug. But `step={1}` on line 247 is purely decorative — nothing calls `checkValidity()` — and the header then renders `2.7 sats/vB`, plus the value flows unrounded into the deposit terms and the Ledger envelope. Either round in `parseValidCustom` or drop the `step` pretense.

services/vault/src/utils/feeRateBounds.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ * Computes the min/default/max sat/vB range for the custom fee-rate input.
+ * The hard floor is the Bitcoin min relay fee (`MIN_RELAY_FEE_RATE_SATS_VB`);
+ * `hourFeeRate` is only a soft warn threshold, surfaced separately by callers.
+ */
+export function getFeeRateBounds(fees: {
+  defaultFeeRate: number;
+  hourFeeRate: number;
```
gbarkhatov, line 37: **Dead parameter and dead return field.**

`hourFeeRate` is a required parameter that the function body never reads — the doc comment on line 33 even says it "is only a soft warn threshold, surfaced separately by callers", which is an accurate description of a parameter that shouldn't be here.

Similarly, `defaultFeeRate` is returned on line 41 but the only caller destructures `{ minFeeRate, maxFeeRate }` (`FeeRateSelector.tsx:84`). `feeRateBounds.test.ts:19` asserts `bounds.defaultFeeRate` — a test for a field no production code reads, which the test-philosophy rule rules out on its own.

Both fall under the zero-dead-code policy; dropping the param and the field also simplifies the call site to `getFeeRateBounds(fastestFee)`.

services/vault/src/copy.ts
```diff
@@ -508,6 +508,21 @@ export const COPY = {
...
+      headerUnit: "sats/vB",
+      slowLabel: "Slow",
+      slowHint: "~1 hour",
+      avgLabel: "Avg",
+      avgHint: "~30 min",
+      fastLabel: "Fast",
+      fastHint: "~10 min",
+      customLabel: "Custom",
+      cardUnit: "sat/vB",
+      customInputSuffix: "sats vB",
```
gbarkhatov: **Three spellings of one unit inside a single panel, and one looks like a typo.**

`headerUnit: "sats/vB"` (line 513), `cardUnit: "sat/vB"` (line 521), and `customInputSuffix: "sats vB"` (this line — missing the slash entirely). All three render in the same panel: header, the tier cards directly beneath it, and the custom input below those.

The rest of the vault UI is uniformly `sats/vB` (`DepositFeesBreakdown.tsx:113`, `WithdrawReviewContent.tsx:107`, `RefundModal/FeeRateField.tsx:71,87`, core-ui's `BTCFeeRate.tsx:15`); the tier cards introduce `sat/vB` into a vault screen for the first time, and `FeeRateSelector.test.tsx:70-74` pins it there.

This is exactly the "capitalization / phrasing drift across screens" the header of this file says it exists to prevent. One constant, used three times.

services/vault/src/copy.ts
```diff
@@ -710,6 +725,12 @@ export const COPY = {
       // Fee-breakdown lines (DepositFeesBreakdown) shown before the user
       // submits. The commission label appends the percent, e.g. "VP commission
       // (2.50%)"; net payout is the deposit minus that commission.
+      networkFeeRateLabel: "Network Fee Rate",
```
gbarkhatov, line 736: **Two copy nits in the new fee-breakdown strings.**

1. `"Network Fee Rate"` and `"BTC Network Fee"` (line 731) are Title Case, while every sibling in this block is sentence case — `"VP commission"`, `"Net payout"`, `"Last deposit"`, `"Active BTC"`. The style header doesn't legislate fee-row labels specifically (its casing rules cover status labels and buttons), so this is consistency with neighbours rather than a rule violation — but they're rendered as a stacked list, so the drift is visible.
2. Both tooltips (lines 730 and 733) say "pre-pegin funding transaction". The style header on line 18 mandates **"Pre-Pegin"** (proper-noun form) for the broadcast phase / transaction, with lowercase "peg-in" reserved for regular prose. These are naming the transaction, so they should be "Pre-PegIn".

services/vault/src/components/simple/DepositFeesBreakdown.tsx
```diff
@@ -98,8 +108,24 @@ export function DepositFeesBreakdown({
       ? FORM_COPY.vpCommissionLabel
       : `${FORM_COPY.vpCommissionLabel} (${formatBasisPointsAsPercent(commissionBps)})`;
 
+  const networkFeeRateValue =
+    networkFeeRate !== undefined && networkFeeRate > 0
+      ? `${networkFeeRate} sats/vB`
```
gbarkhatov, line 113: **User-facing string inlined instead of living in `copy.ts`.**

This is the only hard-coded user-facing literal in the file — every other row gets its display string from a formatter (`formatSatsLine`, `formatBasisPointsAsPercent`) or from `FORM_COPY`. The copy rule asks for a function in `copy.ts` for interpolated strings, e.g. `feeRateValue: (rate: number) => ...`.

Worth doing here specifically because the selector's header renders the same unit from `feeSelector.headerUnit` — two independent literals for one unit across two screens is how they end up disagreeing, which they already do (see the `copy.ts` comment on `customInputSuffix`).

services/vault/src/constants.ts
```diff
@@ -37,3 +37,7 @@ export const BTC_BLOCK_TIME_MINS = 10;
 export const MINS_PER_HOUR = 60;
 export const MINS_PER_DAY = 1440;
 export const FALLBACK_FEE_RATE_SATS_VB = 1;
+
+// Bitcoin's default minimum relay fee. Anything below this is not forwarded by
+// most nodes, so a pre-pegin funded at a lower rate would never broadcast.
+export const MIN_RELAY_FEE_RATE_SATS_VB = 1;
```
gbarkhatov, line 43: **This bound is only ever enforced in a React component.**

CLAUDE.md §2 ("Fee calculation consistency") requires that "cross-check assertions belong at the broadcast site, not only at the estimator". This PR adds a brand-new *user-controlled* input to the fee model and validates it exactly once — in `parseValidCustom` — after which it travels as a bare `number` through `DepositState.feeRate` → `mempoolFeeRate` → `vaultTransactionService.ts:220` → `PeginManager.ts:891` with no re-check.

`vaultPeginBroadcastService.ts` has no fee-rate assertion at all. The only numeric bound near broadcast is `MAX_REASONABLE_FEE_SATS` (`PeginManager.ts:1231-1239`), which for a ~200 vB 1-vault Pre-PegIn corresponds to roughly 5,000 sat/vB — inert against a 128 sat/vB ceiling. `prepeginMaxFee` is only checked as `>= 1n` host-side (`envelope.ts:89-91`).

Since this constant and `maxFeeRate` are now the *only* thing standing between a typed number and a signed transaction, re-asserting them before signing seems worth the few lines.

(Minor, while here: this is the second constant equal to `1` in four lines — `FALLBACK_FEE_RATE_SATS_VB` on line 39 is also `1`. Different meanings, same value, adjacent; a word in the comment distinguishing them would help the next reader.)

services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
```diff
@@ -0,0 +1,290 @@
...
+            </Text>
+          )}
+        </div>
+      )}
+
+      {/* An estimate error blocks Sign for tier rates too, so it must be
+          visible regardless of which option is selected. */}
+      {estimatedFee.error && (
+        <Text variant="caption" className="text-error-main">
+          {estimatedFee.error}
```
gbarkhatov, line 280: **Raw SDK error text is rendered to the depositor.**

`estimatedFee.error` is whatever `err.message` `selectUtxosForPegin` threw — e.g. `Insufficient funds: need 12345 sats, have 10000` (`selectUtxos.ts:158-160`). That's an internal string reaching a user-facing surface, bypassing `copy.ts`.

Worth flagging as new rather than inherited: `DepositForm` deliberately does **not** render `feeError` as text — it uses it only as a boolean (`isFeeError` / `feeDisabled`, lines 351-357) and routes the user-facing wording through `getDepositCtaState`. This is the first place the raw message is displayed.

Given the sizing gap in the comment above, this is also the error a depositor is most likely to actually hit, so it's worth a written message that says what to do (lower the rate / reduce the amount).

services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
```diff
@@ -0,0 +1,290 @@
...
+            <button
+              type="button"
+              aria-label={FEE_SELECTOR_COPY.clearCustomAria}
+              onClick={handleClearCustom}
+              className="flex size-10 shrink-0 items-center justify-center rounded border border-secondary-strokeLight"
+            >
+              <IoClose size={20} />
+            </button>
+          </div>
+          {!estimatedFee.error && estimatedFee.fee !== null && (
```
gbarkhatov, line 268: **The absolute fee is shown only under "Custom" — is that deliberate?**

This readout sits inside the `selectedKey === "custom"` block, so picking Slow/Avg/Fast shows a rate but never a sats figure, at the exact moment the depositor is committing to it.

That gap got wider with this PR: the form's new `BTC Network Fee` row is computed at `fastestFee` (`useDepositPageForm.ts:442` passes no override), so once the depositor changes the rate here, the only absolute-fee number they've seen is stale. During congestion "Fast" is precisely the case where the sats figure matters most.

Hoisting the readout out of the custom block would cost nothing — `estimatedFee` is already computed for every selection.

services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx
```diff
@@ -0,0 +1,290 @@
+/**
+ * FeeRateSelector
+ *
+ * Inline network fee-rate picker rendered in the Deposit Progress view's
+ * pre-sign entry state (replaces the old DepositFeeModal). Owns its own
```
gbarkhatov, line 5: **`DepositFeeModal` has never existed in this repo.**

`git log --all -S"DepositFeeModal"` returns exactly one commit — this PR's own — and `grep` finds the string only in this comment. There was no modal to replace; the pre-sign state simply had no fee picker at all, which is why `DepositProgressView` gains `preSignFeeSelector` / `signDisabled` as brand-new props and the diff removes nothing.

(The nearest real thing is `services/simple-staking/.../Staking/FeeModal/`, a different service.) The PR description repeats the claim, so both are worth correcting — a future reader will go looking for the deleted component.

While here: this file re-implements the same four-option picker as `simple-staking`'s `FeeModal/index.tsx:64-76`, down to the same seeding algorithm, and `feeRateBounds.ts` copies `nextPowerOfTwo` byte-for-byte from `services/simple-staking/src/ui/common/utils/nextPowerOfTwo.ts`. The "don't import across services" rationale is reasonable — just worth stating in the description that this is a deliberate third copy, given `services/vault/src/components/deposit/RefundModal/FeeRateField.tsx` is a fourth fee-rate editor already inside this service.

services/vault/src/hooks/deposit/useEstimatedBtcFee.ts
```diff
@@ -1,3 +1,22 @@
+/**
```
gbarkhatov, line 1: **The moved JSDoc is now detached from the function it documents.**

Hoisting the block above the imports means its `@param` / `@returns` tags no longer attach to `useEstimatedBtcFee` — editors won't surface them on hover, and it now reads as a file-level comment that happens to document a function's signature.

The content change is the useful part (the new `feeRateOverride` line); it just wants to go back onto the function declaration.

services/vault/src/hooks/deposit/__tests__/useEstimatedBtcFee.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+import { useEstimatedBtcFee } from "../useEstimatedBtcFee";
+
+const networkFees = vi.hoisted(() => ({
+  defaultFeeRate: 10,
+  isLoading: false,
+  error: null as Error | null,
+}));
+
+vi.mock("@/hooks/useNetworkFees", () => ({
```
gbarkhatov, line 13: **The same module is mocked twice under two specifiers.**

`@/hooks/useNetworkFees` and `../../useNetworkFees` (line 17) resolve to the same file — vitest keys the mock registry by resolved path, so one of these is redundant. The hook under test imports via the relative specifier.

Separately, the shared `networkFees` object (lines 7-11) is the one mock this PR left without the new `halfHourFeeRate` / `hourFeeRate` fields, while the mock it *did* update (`WithdrawReviewContent.test.tsx:16-21`) never needed them and still omits `isLoading` / `error`. Neither is type-checked — `vi.mock`'s string-path factory is typed `Partial<unknown>` — so nothing is broken; it's just that the edit landed on the mock that didn't need it and skipped the one that did.

services/vault/src/components/simple/__tests__/DepositFeesBreakdown.test.tsx
```diff
@@ -89,14 +89,33 @@ describe("DepositFeesBreakdown commission disclosure", () => {
...
+        commissionBps={250}
+        networkFeeRate={12}
+        networkFeeSats={1500n}
+      />,
+    );
+
+    expect(screen.getByText("Network Fee Rate")).toBeInTheDocument();
+    expect(screen.getByText("BTC Network Fee")).toBeInTheDocument();
+    expect(screen.getByText("12 sats/vB")).toBeInTheDocument();
+    expect(screen.queryByRole("button")).not.toBeInTheDocument();
```
gbarkhatov, line 118: **This assertion passes vacuously.**

`DepositFeesBreakdown` never rendered a button in any state, so `queryByRole("button")` returning null proves nothing about the new rows being read-only — the test named "renders the network fee rate as a read-only row" would still pass if the rows were removed entirely. If read-only-ness is the behaviour worth pinning, assert the absence of the specific control (`queryByRole("spinbutton")`, or the selector's clear button by its aria-label).

Same test, line 105 area: bumping the tooltip count `4 → 6` pins a total that any future row breaks, for a property nobody is really asserting. Checking that the two new tooltips are present by content would survive unrelated additions.

gbarkhatov (dismissed):
Approving — the core threading is right and I verified it end to end rather than assuming it.

**Verified correct:**
- The selected rate genuinely reaches the built transaction. `mempoolFeeRate` is in `executeDeposit`'s dependency array (`useDepositFlow.ts:1572`) → `vaultTransactionService.ts:220` → `PeginManager.ts:891`. No ref/memo capture, no stale closure, and nothing keys or resets flow state when the rate changes.
- `peginOutputCount(vaultAmounts.length, true)` matches every other call site (`useDepositPageForm.ts:434`, `PeginManager.ts:892`).
- Fractional rates don't crash the fee math — `Math.ceil` precedes every `BigInt(...)` in `peginFeeMath.ts:61-73`.
- The live-follow effect does what the comment claims: a custom value derives from `parseValidCustom` and is genuinely immune to tier refetch.
- `spendableMempoolUTXOs` is confirmed-only despite the name, and is the same set (same query cache entry) the real build selects from.

**What I'd resolve before this goes out** — comments 1, 2 and 3 below:

1. The pre-sign affordability gate measures a smaller transaction than the one that gets funded. The build fails once `fee(chosenRate) − fee(rateAtMaxClick)` exceeds `PRE_PEGIN_SAFETY_BUFFER_SATS` (3,000), but the selector's gate doesn't trip until that delta also exceeds the whole reserve stack. A 5 → 128 sat/vB jump on a ~250 vB Pre-PegIn is a ~30k sat delta, so this is reachable, not theoretical.
2. That failure is unrecoverable in place — the selector unmounts on `started` and the error CTA closes the modal, resetting `feeRate` to 0.
3. `useEstimatedBtcFee` now holds two disagreeing definitions of "ready". Unreachable from today's only override caller, but it's a critical-path file (CLAUDE.md §2).

One thing worth deciding explicitly rather than inheriting: this PR adds a **user-controlled** input to the fee model and validates it exactly once, inside a React component. CLAUDE.md §2 asks for cross-checks at the broadcast site, not only at the estimator — see the comment on `constants.ts`. Nothing downstream re-asserts the rate, and the only numeric bound near broadcast (`MAX_REASONABLE_FEE_SATS`) is ~5,000 sat/vB effective for a 1-vault Pre-PegIn, so it's inert against a 128 sat/vB ceiling.

The rest (4–17) are cleanups — copy, dead code, and test assertions.

jonybur (dismissed):
The delta since the previous approval is a pure UX repositioning: the pre-sign fee selector moves from below the stepper into the 'Register deposit' card under the Pre-PegIn row, with three new tests around the placement. No fee math, broadcast wiring, or state-machine touch. The mempoolFeeRate threading through executeDeposit → vaultTransactionService → PeginManager is unchanged. The affordability-gate mismatch, unrecoverable-in-place failure, and useEstimatedBtcFee 'ready' split flagged earlier are still present and known to the author; the layout move does not touch them. CI green.

services/vault/src/components/simple/DepositProgressView/GroupBlock.tsx
```diff
@@ -38,8 +52,20 @@ export function GroupBlock({
...
+  // Pre-sign entry: this group owns the Pre-PegIn step, so it opens into the
+  // card and shows that one row carrying the fee selector. The other rows stay
+  // hidden — the entry screen is about the one decision left to make, not the
+  // whole group. The row reads active (that is what `StepRow` gates its detail
+  // on) even though nothing is running yet.
+  const showPreSignDetail =
+    !group.expanded &&
+    Boolean(preSignDetail) &&
+    group.startStep <= PRE_PEGIN_VISUAL_STEP &&
+    PRE_PEGIN_VISUAL_STEP <= group.endStep;
```
gbarkhatov, line 66: **The selector's mount is now implicitly coupled to group visibility.**

`showPreSignDetail` tests `!group.expanded` and step-range containment, but never `group.status`. Both parents drop `status === "completed"` groups before `GroupBlock` sees them (`GroupedProgress.tsx:48`, and the trunk filter in `SplitGroupedProgress.tsx`), so once `currentStep > 6` the Register-deposit group is gone and a supplied selector silently never mounts. The footer render this replaces (`{!started && preSignFeeSelector}`) had no such condition.

Not reachable today, and I'd argue the new behaviour is the better one where it does apply: for the un-started-mid-flow state covered by `DepositProgressView.test.tsx:155-181` the Pre-PegIn is already broadcast, so a fee selector there was always wrong. It's safe only because `ResumeDepositContent` passes no selector and `DepositSignContent` always enters at `DERIVE_VAULT_SECRET`.

Worth making that explicit rather than incidental, because the failure is silent in a direction that matters: `feeRateValid` defaults to `true` (`DepositSignContent.tsx:88`), so a future caller that supplies a selector no group claims gets an enabled Sign CTA and builds at a rate the depositor never saw. Either widen the predicate and fall back to rendering in the footer, or document the prop as "only honoured while the Pre-PegIn group is still visible" so the contract is stated where the next caller will read it.

services/vault/src/components/simple/DepositProgressView/GroupBlock.tsx
```diff
@@ -38,8 +52,20 @@ export function GroupBlock({
...
+
+  if (!group.expanded && !showPreSignDetail) {
```
gbarkhatov, line 68: **The group opens into its card while its own header still says "not started".**

Same root cause as the comment below on the `StepRow`, but a separate surface and a separate decision. At the entry state `buildStepGroups(1, false)` leaves group 1 `status: "upcoming"` (`steps.ts:212`). This branch now falls through to the card, which renders `<GroupHeader status={group.status} />` on line 88 — so `GroupHeader.tsx:26` emits `COPY.deposit.a11y.groupStatus.upcoming` while the row inside the same card announces "Step 5 active". One group, two contradictory states, both spoken.

`DepositProgressView.test.tsx:244-247` asserts all four groups read not-started, and its comment says "the pre-entry render must look exactly as it always has — four collapsed group headers". Neither is true of the production path any more; the test only holds because it passes no selector.

Whatever lands for the `StepRow` state should settle this too: if the group opens purely to carry the selector, decide deliberately whether its header reads `upcoming` (consistent with `buildStepGroups`, inconsistent with it being open) or `active` (the reverse), rather than inheriting the answer.

services/vault/src/components/simple/DepositProgressView/GroupBlock.tsx
```diff
@@ -68,34 +94,46 @@ export function GroupBlock({
...
+          {showPreSignDetail ? (
+            <StepRow
+              state="active"
+              number={PRE_PEGIN_VISUAL_STEP - group.startStep + 1}
+              ariaNumber={PRE_PEGIN_VISUAL_STEP}
+              label={steps[PRE_PEGIN_VISUAL_STEP - 1]?.label ?? ""}
+              detail={preSignDetail}
+              compact={compact}
+              inCard
+            />
```
gbarkhatov: **[blocking] The entry screen now spins a loader and announces "Step 5 active" while nothing is running.**

`StepRow` gates `detail` on `isActive` (`StepRow.tsx:148,205`), so rendering the selector here required declaring the row active. But `state="active"` with `inCard` is not a neutral container: `StepRow.tsx:43-53` renders an infinite `<Loader size={16}>` plus `aria-label={COPY.deposit.a11y.stepActive(5)}` → "Step 5 active". At this point nothing has started and `currentStep` is 1, so step 5 isn't even current.

This is precisely the invariant the file next door documents. `steps.ts:195-199` says the `started=false` gate exists so "no sub-step row can render as in-progress while nothing is actually running… visually or to a screen reader", and `DepositProgressView.test.tsx:183-219` was written to pin it. Both the rationale comment and that test were left untouched by this commit; the test still passes only because it supplies no `preSignFeeSelector`, while production (`DepositSignContent.tsx:228`) always supplies one. So the guard is intact in the tests and gone in the app.

The reduced-motion case is worse than the default one: per `docs/motion-system.md` the functional spinner is the single animation re-enabled under `prefers-reduced-motion`, so those users get a perpetual spinner on an idle screen with no way to stop it.

The comment on line 60-61 already names the mechanism ("The row reads active … even though nothing is running yet") without following it to either consequence.

Fix: decouple "shows a detail panel" from "is running". Adding a `forceDetail` prop to `StepRow` — the two `isActive` checks on lines 148 and 205 are the only gate — lets this pass `state="pending"` and keep the hollow ring and the "Step 5 not started" label that `StepCircle` already renders for that state.

gbarkhatov (dismissed):
Follow-up to my review of `2ba9ef15e` — dismissed by the push, findings intact. One commit since: `b23e8e1`, which moves the selector onto the Pre-PegIn step. It touches five files under `DepositProgressView/`; the other sixteen files in the PR are byte-identical to what I reviewed, so this is a narrow pass. I read the code at HEAD for every verdict below rather than inferring anything from thread state.

### Ledger — all 17 prior comments

| # | Claim | Verdict |
|---|---|---|
| 1 | **[blocking]** affordability gate sized on `sum(vaultAmounts)`, not `prePegin.totalOutputValue` | unaddressed — `FeeRateSelector.tsx:68-78` |
| 2 | **[blocking]** a fee-rate mistake is unrecoverable in place | unaddressed — `DepositProgressView.tsx:500-510`, `DepositSignContent.tsx:84,115-130,207-236` |
| 3 | **[blocking]** two disagreeing "ready" definitions; `maxDeposit: null` read downstream as "no ceiling" | unaddressed — `useEstimatedBtcFee.ts:60` vs `:72` |
| 4 | tier seeding by float equality against live-refetching values | unaddressed — `FeeRateSelector.tsx:94-137` |
| 5 | out-of-range custom rate disables Sign with no explanation; no `max` on the input | unaddressed — `FeeRateSelector.tsx:143-150,241-273` |
| 6 | fractional rates accepted against the SDK's `Number.isInteger` contract | unaddressed — `FeeRateSelector.tsx:47,247` |
| 7 | dead `hourFeeRate` param + dead `defaultFeeRate` return field | unaddressed — `feeRateBounds.ts:37,41`, `feeRateBounds.test.ts:12,33` |
| 8 | three spellings of one unit in one panel | unaddressed — `copy.ts:513,521,522` |
| 9 | Title Case labels + "pre-pegin" for the transaction | **half refuted** — part 1 withdrawn (below); part 2 stands at `copy.ts:730,733` |
| 10 | `sats/vB` inlined instead of living in `copy.ts` | unaddressed — `DepositFeesBreakdown.tsx:113` |
| 11 | fee-rate bound enforced only inside a React component | unaddressed — `constants.ts:39,43` |
| 12 | raw SDK error text rendered to the depositor | unaddressed — `FeeRateSelector.tsx:277-281` |
| 13 | absolute sats fee shown only under "Custom" | unaddressed — `FeeRateSelector.tsx:266-271` |
| 14 | `DepositFeeModal` never existed | unaddressed — `FeeRateSelector.tsx:5` |
| 15 | JSDoc hoisted above the imports, detached from the function | unaddressed — `useEstimatedBtcFee.ts:1-18` |
| 16 | same module mocked twice; shared mock missing the new fields | unaddressed — `useEstimatedBtcFee.test.tsx:7-19` |
| 17 | `queryByRole("button")` passes vacuously; brittle tooltip count | unaddressed — `DepositFeesBreakdown.test.tsx:97,118` |

1–8 and 10–17 stand exactly as written on the existing threads — I'm not restating them here, and the anchors above are at HEAD so they're easy to find. All three blockers are still blockers, and **2 is now worse**: the move gives the entry screen a spinning step row, so the state a depositor cannot back out of is now also the state that looks like it is already running.

### New in `b23e8e1`

Three inline comments below, on `GroupBlock.tsx`. The first is a merge-blocker; the second shares its root cause; the third is latent and cheap.

The short version: `StepRow` gates `detail` on `isActive`, so the only way to hang the selector off a step row was to declare that row active — which turns on an infinite spinner and a "Step 5 active" announcement on a screen where nothing has started. That is the exact thing the `started=false` gate was built to prevent, and the test written to pin it (`DepositProgressView.test.tsx:183-219`) still passes only because it omits `preSignFeeSelector`, which production always supplies.

Nothing else in the delta is wrong. No `data-testid` was dropped, and the split-trunk routing is correct — `VaultColumn` doesn't accept `preSignDetail`, so the per-vault columns can't duplicate the selector, which the new test at `:277-291` pins meaningfully.

### Refuted — mine, withdrawn

**9, part 1 — Title Case fee labels.** I called `"Network Fee Rate"` / `"BTC Network Fee"` casing drift this PR introduced. That was wrong: both strings already exist verbatim at the merge base `803cfb912` (`copy.ts:493-494`, `:1277-1278`, `:1346`). The new strings match the app's existing spelling of these exact labels rather than departing from it. Only the block-local mismatch against `vpCommissionLabel` and its neighbours is real, and it predates this PR — ignore that half of the comment. Part 2 is unaffected and still stands: `copy.ts:730` and `:733` say "pre-pegin funding transaction" where the file's own style header (`copy.ts:18`) mandates the proper-noun "Pre-Pegin" when naming the broadcast phase or transaction.

Approving — the placement is a genuine improvement over the footer, and what's left is yours to triage.

kirugan (approved):

jonybur (approved):
Read the delta since the previous approvals (dismissed by the last two pushes) and checked the outstanding threads at the current head. Nothing new to raise beyond the threads gbarkhatov and coderabbitai already opened — the fee-rate threading from FeeRateSelector through DepositState.feeRate into executeDeposit/vaultTransactionService/PeginManager is intact and the peginOutputCount call is consistent with the shared model. Approving on that basis; treat the still-open blocking threads (affordability gate sizing, unrecoverable-in-place, and useEstimatedBtcFee dual readiness) as author-owned triage rather than dismissed.

PR #2310, opened 2026-08-25
feat(vault, wallet-connector): gate the Ledger vault wallet on WebHID

jrwbabylonlab:
The Ledger vault wallet now appears only when NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET is true AND the browser has WebHID. With devnet's flag on, browsers without navigator.hid no longer show a dead hardware entry. Also corrects the flag's JSDoc and .env.example, which still claimed signing was unimplemented. Enablement stays per-environment: on in devnet, off in testnet.


jonybur (dismissed):
The wallet-metadata probe on `navigator.hid` and the parallel guard in `VaultWalletConnectionProvider` line up; both check the same thing, and the disabledWallets fallback now correctly hides the entry in Firefox/Safari. Tests cover both branches. .env.example and the feature-flag JSDoc are now accurate about signing being implemented.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/index.ts
```diff
@@ -6,13 +6,15 @@ import { LedgerVaultProvider, WALLET_PROVIDER_NAME } from "./provider";
...
- * "true" — gating lives in the consuming app, where env access exists.
+ * staking adapters. Nothing is injected into the page; the `wallet` probe
+ * reports installed only where WebHID exists (`navigator.hid` — Chromium,
+ * secure context; the DMK web-hid transport needs it). Hidden unless
+ * `NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET` is "true" — gating lives in the
+ * consuming app, where env access exists.
  */
 const metadata: WalletMetadata<IBTCProvider, BTCConfig> = {
   id: "ledger_btc_vault",
+  wallet: (context) => (context?.navigator?.hid ? context : null),
```
gbarkhatov: **The probe doesn't hide the entry, and where it does apply it makes the row worse than before.**

`installed: false` does not remove a wallet from the connect list:

- `components/Wallets/index.tsx:39` filters on `installed` only for `wallet.id === "injectable"`.
- `walletDisplayRank` (`Wallets/index.tsx:29-33`) returns `1` for any `hardware` wallet regardless of `installed`, so it doesn't even sink to the uninstalled tier.
- `components/WalletButton/index.tsx:41` — `const reachable = installed || hardware` — keeps it as `as: "button"` with the live `onClick`, and `:75-79` (`hardware ? label : ...`) keeps the yellow dot and the "Hardware wallet" chip.

The only visible delta between `installed` true and false here is two background shades and a greyed name. Clicking it goes `onSelectWallet` → `WalletConnector.connect` → `Wallet.ts:51-52` `throw Error("Provider not found")` — an untyped error with no `ERROR_CODES` code, because `core/index.ts:38` skips `createProvider` entirely when the probe returns null. So in a browser without `navigator.hid`, the probe alone produces a clickable, yellow-badged "Hardware wallet" row that errors on click: the dead hardware row this PR is trying to remove, with a worse failure than the DMK would have given.

There's direct precedent against this pattern in the same package — `core/wallets/eth/appkit/index.ts:25-31` sets `wallet: undefined` with the comment that gating a non-injected wallet on a browser global "made every user without an EVM browser extension (Safari, WalletConnect-QR, hardware-only) fail with 'Provider not found' before the AppKit modal could open." The three sibling hardware wallets (`keystone`, `ledger`, `ledger-v2`) have no `wallet` field either; `ledger_btc_vault` would become the only `hardware: true` entry with a probe — and `hardware: true` is exactly what neutralises it.

The probe does buy one real thing: it blocks the auto-reconnect at `core/index.ts:99-102` and `hooks/redetectReconnect.ts:43`, both of which key off `installed`. But that only matters in a consumer that hasn't disabled the wallet, which is none today (see the comment on the new test file).

Suggest dropping the `wallet` field and its test, keeping only the app-side `DISABLED_WALLETS` gate — the half that actually works. If the connector must self-gate, it needs to drop the entry from the list, not flip `installed`; e.g. have `Wallets` also filter `hardware && !provider`.

Minor, if the probe survives: every other metadata (`okx`, `onekey`, `utila`, `unisat`, `injectable`) returns the injected provider object, which `createProvider` genuinely consumes. Here `createProvider: (_wallet, config) => ...` discards it, so returning `context` is a truthiness sentinel that parks a whole `Window` reference on `Wallet.origin` (and `clone()` copies it). `context.navigator.hid` would be the truer origin.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/index.test.ts
```diff
@@ -0,0 +1,30 @@
+import { describe, expect, it } from "vitest";
+
+import { createWallet } from "@/core";
+import { Network, type BTCConfig } from "@/core/types";
+
+import metadata from "../index";
+
+const config = { network: Network.SIGNET } as BTCConfig;
+
+describe("ledger-vault wallet metadata — WebHID availability probe", () => {
```
gbarkhatov: **This suite pins a path no caller reaches.**

`core/index.ts:81-84` `continue`s on `disabledWallets` *before* calling `createWallet`, so the probe never runs for a gated-out wallet. Both consumers in this repo gate it out in exactly the case where the probe would return null:

- vault — `VaultWalletConnectionProvider.tsx:52`, the new `featureFlags.isLedgerVaultWalletEnabled && isWebHidAvailable` line;
- simple-staking — `WalletConnectionProvider.tsx:71`, unconditional `disabled.push("ledger_btc_vault")`.

And when WebHID *is* present, the probe returns truthy — identical to pre-PR. Net: the probe changes nothing for either consumer today. That's fine as defence-in-depth if the probe worked, but per the comment on `index.ts:17` it doesn't hide anything on its own, so this test is locking in behaviour that is both unreachable and not what the JSDoc claims.

services/vault/src/config/featureFlags.ts
```diff
@@ -132,9 +132,9 @@ export default {
...
    * `ledger_btc` / `ledger_btc_v2` staking adapters.
-   * Why needed: the provider is built ahead of Ledger's firmware being final,
-   * AND signing is not implemented yet (signPsbt/signPsbts/signMessage throw),
-   * so no deposit can complete with this wallet — it must stay invisible
+   * Why needed: the full depositor flow is implemented (intent ceremony, PoP,
+   * Pre-PegIn and payout signing), but Ledger's firmware is still in review,
+   * so the wallet ships per environment — it must stay invisible
    * everywhere by default while Ledger can switch it on in a test environment
    * without a code change or a release.
    * Default: false (hidden unless explicitly set to "true")
```
gbarkhatov, line 145: **This JSDoc now documents only half the gate.**

"Purpose: Reveals the Ledger vault wallet … Default: false (hidden unless explicitly set to `"true"`)" — since this PR the flag is necessary but no longer sufficient. `.env.example` got the new sentence ("the entry is hidden in browsers without navigator.hid regardless of this flag"); this JSDoc didn't.

Worth adding here too, because this is the file someone reads when the flag appears to do nothing. Two realistic cases: a Firefox/Safari QA pass, and a `http://192.168.x.x` LAN origin — `navigator.hid` is `[SecureContext]`-gated in the WebHID IDL, so it's absent over plain HTTP even on Chromium. Both currently produce a silently empty result with no explanation in the flag's own docs.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -42,9 +42,16 @@ const ALWAYS_DISABLED_WALLETS: string[] = [
 // environment that has not listed it.
 const LEDGER_VAULT_WALLET_ID = "ledger_btc_vault";
 
+// The DMK web-hid transport needs WebHID (`navigator.hid` — Chromium-only,
+// secure context); without it the entry would render as a dead hardware row.
+const isWebHidAvailable =
+  typeof navigator !== "undefined" && "hid" in navigator;
```
gbarkhatov: Nit: the same rule is now spelled two different ways — presence here (`"hid" in navigator`) and truthiness in the connector (`ledger-vault/index.ts:17`, `context?.navigator?.hid ? ...`).

They can disagree: if `hid` is present-but-falsy (a stub or polyfill assigning `hid = undefined`), this gate enables the wallet → it gets constructed → the probe returns null → `provider = null` with `hardware: true` → exactly the clickable "Provider not found" row described on `ledger-vault/index.ts:17`. That's the one configuration where both gates are live at once and they disagree. Contrived, but one form in both places reads better regardless — and moot if the connector-side probe is dropped.

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.disabledWallets.test.tsx
```diff
@@ -108,11 +109,23 @@ describe("WalletConnectionProvider — Ledger vault wallet opt-in", () => {
...
+    });
     await renderWithFlag("true");
     expect(h.captured.disabledWallets).not.toContain("ledger_btc_vault");
   });
 
+  it("hides it without WebHID even when the flag is on", async () => {
+    // jsdom default: no navigator.hid — the Firefox/Safari case, where the
+    // entry would otherwise render as a dead hardware row.
+    await renderWithFlag("true");
+    expect(h.captured.disabledWallets).toContain("ledger_btc_vault");
```
gbarkhatov, line 133: **The flag-off cases no longer isolate the flag.**

Every flag-off case in this block ("flag is unset", `"1"`) runs with `navigator.hid` absent, and the only `hid`-stubbed case has the flag on. So the quadrant *flag-off + WebHID-present* is untested — and deleting `featureFlags.isLedgerVaultWalletEnabled &&` from `VaultWalletConnectionProvider.tsx:52` leaves all five tests in this describe green (I walked each one). That line is the getter's only consumer and this file its only coverage, so nothing else catches it.

The other direction is fine: dropping `&& isWebHidAvailable` correctly fails this new test.

Worth adding: flag unset + `hid` stubbed → still disabled.

gbarkhatov (dismissed):
Approving — the app-side gate is correct and the doc cleanup is accurate. Findings below are about the connector-side half of the change, which doesn't do what its JSDoc says, plus two doc/test gaps.

**What I verified and cleared:**

- The removed "signing is not implemented yet" text was genuinely stale — `signPsbt`, `signPsbts`, `signMessage` and `deriveContextHash` are all real implementations on `LedgerVaultProvider`.
- WebHID is the right capability to gate on: `dmkSession.ts:52-57` registers `webHidTransportFactory` as the DMK's only transport, unlike `ledger` / `ledger-v2`, which have a WebUSB fallback.
- No SSR/hydration concern with the module-level `isWebHidAvailable` — `services/vault` is a Vite SPA, not Next.js, despite the `NEXT_PUBLIC_` naming convention. The `typeof navigator !== "undefined"` guard is harmless defensiveness.
- `navigator.hid` is `[SecureContext]`-gated in the WebHID IDL, so a presence check already covers the secure-context requirement — no separate `isSecureContext` test is needed in either gate.
- No new redetection/auto-reconnect loop: `useWalletRedetection`'s `stale` filter is already commonly non-empty (unisat/onekey/okx/utila all probe), and it is trigger-bounded, not a spin.
- `sync-env.mjs --check` parses `KEY=value` lines only, so the comment-only `.env.example` edit can't report drift. Prettier keeps the wrapped `isWebHidAvailable` assignment as written (81 chars unwrapped, printWidth 80) — no lint churn.

**One note that can't carry an inline anchor** (the file isn't in this diff): `core/wallets/btc/index.ts:23-24` says `ledgerVault` is "registered but hidden unless the consuming app enables it". At package level it's the reverse — it's listed for every consumer unless one disables it. Since this PR is partly a doc-accuracy pass, worth folding in.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/index.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/index.test.ts
jrwbabylonlab: fixed

services/vault/src/config/featureFlags.ts
jrwbabylonlab: fixed

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jrwbabylonlab: fixed

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.disabledWallets.test.tsx
jrwbabylonlab: fixed

jonybur (dismissed):
The follow-up commits addressed the earlier findings: metadata JSDoc now describes only what the connector controls (no wallet probe, no filtering), disabledWallets is the intended gate, and it defers availability gating to the consuming app. The provider probe test was rewritten to assert IProviderMetadata contract shape rather than the pre-createWallet path no caller reaches. featureFlags.ts now spells out both halves of the gate (flag AND WebHID) rather than describing only the flag. VaultWalletConnectionProvider.tsx switched to the presence check ('hid' in navigator), consistent with the wallet-metadata capability semantic. The disabledWallets test now stubs navigator.hid on every flag-off case so the flag is the real term being exercised, and adds the 'flag on but no WebHID' case. Doc cleanup on .env.example is accurate — signing is implemented; the flip is per-environment.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/index.ts
```diff
@@ -6,10 +6,13 @@ import { LedgerVaultProvider, WALLET_PROVIDER_NAME } from "./provider";
...
- * field: nothing is injected into the page, so the entry is always
- * "installed". Hidden unless `NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET` is
- * "true" — gating lives in the consuming app, where env access exists.
+ * staking adapters. Nothing is injected into the page, so like the other
+ * hardware entries there is no `wallet` probe: `installed: false` would not
+ * remove a `hardware` row from the connect list (`Wallets` filters on it only
+ * for `injectable`), it would just leave a clickable row that throws
+ * "Provider not found". Availability gating — the feature flag AND WebHID
+ * (`navigator.hid`, which the DMK web-hid transport needs) — lives in the
+ * consuming app's disabled-wallets list, where env access exists.
```
gbarkhatov, line 15: This rewrite is accurate — I checked each claim. `Wallets/index.tsx:39` does filter on `installed` only for the id `injectable`; all four `hardware: true` BTC entries (`ledger`, `ledger-v2`, `keystone`, `ledger-vault`) declare no `wallet` probe, so all four are permanently `installed === true` via `Wallet.ts:42-44`; `WalletButton/index.tsx:41` (`installed || hardware`) keeps such a row clickable; and `Wallet.ts:51` is the `"Provider not found"` throw. `eth/appkit/index.ts:25-32` is the standing precedent for exactly this reasoning.

The gap is its sibling one directory up. `core/wallets/btc/index.ts:23-24` still says:

```
// ledgerVault is registered but hidden unless the consuming app enables it
// (NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET) — see #2109.
```

That's wrong three ways now. At package level `ledgerVault` is registered and *visible* — `core/index.ts:82` only skips ids a consumer listed in `disabledWallets`; both consumers happen to list it, but that's their doing, not the package's. It names only the flag, when availability is flag AND WebHID. And it hardcodes a vault-app env var name inside the shared package, which is precisely the coupling the sentence I'm commenting on goes out of its way to avoid.

Since this PR is a doc-accuracy pass on this wallet, worth folding in — something like `// ledgerVault is registered here; consumers that don't support it disable it by id — #2109.`

services/vault/src/config/featureFlags.ts
```diff
@@ -132,11 +132,15 @@ export default {
    * connect UI. It drives Ledger's dedicated vault app over the DMK — a
    * different device app and a different transport stack from the legacy
    * `ledger_btc` / `ledger_btc_v2` staking adapters.
-   * Why needed: the provider is built ahead of Ledger's firmware being final,
-   * AND signing is not implemented yet (signPsbt/signPsbts/signMessage throw),
-   * so no deposit can complete with this wallet — it must stay invisible
+   * Why needed: the full depositor flow is implemented (intent ceremony, PoP,
```
gbarkhatov: "the full depositor flow is implemented" overstates it in the same direction the deleted line understated it.

Reclaim of the depositor-claim reserve is deliberately *not* supported on this wallet. `models/reclaimEligibility.ts:173-188` returns `{ type: "blocked", tooltip: COPY.reclaim.blocked.ledgerUnsupported }`, because the firmware routes any 34-byte `<D> OP_CHECKSIG` tapleaf to the claim validator, which hard-requires the protocol claim_tx shape of 1 input and 2 outputs and rejects a 1-in/1-out sweep on the output count before any device I/O. That block is what `LEDGER_VAULT_WALLET_ID` is exported for (`useReclaimRowAction.ts:87`). Refund *is* reachable on the device, so the exception is narrow and worth naming rather than leaving two files in the repo asserting opposite things.

Suggest "the full **deposit** flow is implemented (…); reclaim of the depositor-claim reserve is not — see `models/reclaimEligibility.ts`". `.env.example:100-102` carries the same claim ("The full depositor flow … runs against the device") and needs the same edit.

The rest of the paragraph checks out: `signPsbt` (`provider.ts:665`), `signPsbts` (`:684`) and `signMessage` (`:1003`) are all real, and the text this replaces was already stale on `origin/main` — good catch to fix it here.

services/vault/src/config/featureFlags.ts
```diff
@@ -132,11 +132,15 @@ export default {
...
+   * Pre-PegIn and payout signing), but Ledger's firmware is still in review,
+   * so the wallet ships per environment — it must stay invisible
    * everywhere by default while Ledger can switch it on in a test environment
    * without a code change or a release.
+   * Necessary but not sufficient: the entry also needs WebHID
+   * (`navigator.hid` — Chromium only, and only in a secure context, so it is
```
gbarkhatov: Precision nit, worth raising only because this text is new: "Chromium only" is *desktop* Chromium only. WebHID isn't implemented on Chrome for Android, so `navigator.hid` is absent there too. The gate does the right thing either way; the prose reads as though any Chromium qualifies, which is what a QA pass on a phone would trip over.

The secure-context half is exactly right, and I checked it in a browser rather than trusting the IDL: `"hid" in navigator` is `false` on a non-secure origin and `true` on `http://localhost`. So the presence check already covers it and no separate `isSecureContext` test is needed.

One case the presence check can't cover, if you want a clause for it: in a cross-origin iframe without `allow="hid"` the attribute is present and `requestDevice()` rejects instead. Not a concern for the vault, which is a top-level document — nothing to fix, just the one false-positive of this check.

services/vault/.env.example
```diff
@@ -97,11 +97,12 @@ NEXT_PUBLIC_REOWN_PROJECT_ID=your-reown-project-id-here
...
 # Ledger's dedicated vault app over the DMK, NOT the legacy ledger_btc staking
-# adapters. Hidden unless this is explicitly "true" — set it in a test
-# environment to exercise pairing, the address/pubkey read, DERIVE_CONTEXT_HASH
-# and APPROVE_VAULT_INTENT on a real device. Signing is not implemented yet, so
-# no deposit can be completed with this wallet. Requires Chromium and a secure
-# context (WebHID).
+# adapters. Hidden unless this is explicitly "true". The full depositor flow
+# (pairing, DERIVE_CONTEXT_HASH, APPROVE_VAULT_INTENT, Pre-PegIn and payout
+# signing) runs against the device, whose firmware is still in review — so this
+# is a per-environment opt-in: on in devnet, off elsewhere until the emulator
```
gbarkhatov: Two things about this sentence.

**It pins live deployment state in a checked-in file — the same decay mode this PR is fixing.** It's true today: the only wiring is `.github/workflows/service-release-vault.yml:199` passing `${{ vars.NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET }}`, and I read the environment variables — `vault-devnet` is `"true"`, while `vault-staging-devnet`, `vault-testnet`, `testnet` and `mainnet` are unset. But that state lives entirely in GitHub environment settings, with no test, lint or `.env` behind it, so if the variable is flipped back this file silently becomes wrong with nothing to catch it. `.env.dev-testnet:3` already says per-env flag state stays out of the repo. `.env.example` documents *the variable*, not *the deployment* — suggest keeping the contract here ("enabled per environment via the deployment's GitHub environment variable; also requires Chromium in a secure context") and leaving rollout status in the tracking issue, where it can actually be closed out.

**"until the emulator E2E gates the flip" names an E2E that doesn't exist yet** — `grep -ri ledger services/vault/e2e` returns nothing, and the Speculos seam is still package-internal by design (`dmkSession.ts:69-71`).

For the follow-up issue rather than this PR: `setDmkTransportOverride` (`dmkSession.ts:80-85`) bypasses WebHID entirely, so once that seam is exposed to the dApp, `isWebHidAvailable` becomes a second gate the emulator run has to satisfy while never touching WebHID. I checked the headless-shell concern and it doesn't bite here — Playwright resolves `chromium.launch()` to the full Chromium build (`chromium-1194`), and both that and `chromium_headless_shell-1194` expose `navigator.hid` on a `localhost` origin, so a Playwright run passes the gate today. Still worth planning for the gate to mean "the DMK has a usable transport" rather than "the browser has WebHID".

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -47,9 +47,16 @@ const ALWAYS_DISABLED_WALLETS: string[] = [
  */
 export const LEDGER_VAULT_WALLET_ID = "ledger_btc_vault";
 
+// The DMK web-hid transport needs WebHID (`navigator.hid` — Chromium-only,
+// secure context); without it the entry would render as a dead hardware row.
```
gbarkhatov: "would render as a dead hardware row" describes the state before `27a135c4e`, when the `wallet` probe was still in the metadata. With the probe dropped there's no `wallet` getter, so `Wallet.ts:42-44` (`get installed() { return Boolean(this.provider) }`) makes this entry permanently `installed === true` and the row renders as a completely normal hardware entry. Clicking it on Firefox/Safari reaches `dmkSession.ts:134-139` → `WebHidTransport.isSupported()` → `WebHidTransportNotSupportedError`, which `provider.ts:293-303` surfaces as a typed, coded `WalletError(CONNECTION_FAILED)`: *"Could not connect to …: WebHID not supported"*. The DMK imports are dynamic (`dmkSession.ts:128-140`), so there's no eager module-load crash either. It degrades cleanly, not deadly.

That matters beyond wording, because it's the whole trade this line is justifying. Before this PR an unsupported browser got a specific reason; after it the entry is simply absent, with no console output and no UI signal on the flag-on/no-WebHID branch. "I enabled the flag and the Ledger row still isn't there" then becomes indistinguishable from a mis-set flag, a non-Chromium browser, and a plain-http LAN origin — three very different support answers.

Hiding is still the right call; offering a wallet that can only fail is worse. Two suggestions: reword to what actually happens (e.g. "…without it the entry would still render and stay clickable — `WalletButton` treats every `hardware` row as reachable — and fail only on click"), and log the reason on that branch. The file already imports `logger` from `@/infrastructure`; a `console.info` is fine too if you'd rather this not reach Sentry.

The same "dead hardware row" phrasing is repeated at `__tests__/VaultWalletConnectionProvider.disabledWallets.test.tsx:130-131` and needs the same fix.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -47,9 +47,16 @@ const ALWAYS_DISABLED_WALLETS: string[] = [
...
+const isWebHidAvailable =
+  typeof navigator !== "undefined" && "hid" in navigator;
```
gbarkhatov: Two notes on this line, neither blocking.

`typeof navigator !== "undefined"` is unreachable-false. There's no SSR here — the vault is a Vite SPA (no `next.config.*`, no `"use client"` in `src`, no `hydrateRoot`, `index.html` ships an empty `#root`) — and Node ≥21 defines a global `navigator` anyway, which I confirmed on the repo's Node 24: `typeof navigator === "object"` and `"hid" in navigator === false`. So the guard is dead in every environment this module can load in, including a hypothetical server pass. It mirrors the existing `typeof window` guard on line 63, so consistency is a fair reason to keep it — flagging only against the zero-dead-code rule.

`"hid" in navigator` reads as a stylistic choice but is forced: TS's DOM lib doesn't declare `Navigator.hid` and there's no `@types/w3c-web-hid` in the lockfile, so `navigator.hid` wouldn't compile. Worth a clause in the comment above, otherwise the next person will try to "simplify" it and bounce off `tsc`.

On the CodeRabbit suggestion to extract `"hid"` into a named constant — I'd skip it. It's a spec-defined Web API property name, not a magic value, and the indirection makes the `in` narrowing harder to read rather than easier.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -47,9 +47,16 @@ const ALWAYS_DISABLED_WALLETS: string[] = [
...
+
 const DISABLED_WALLETS: string[] = [
```
gbarkhatov, line 56: While this list is being changed: two comments in this file now describe a gate that no longer exists, and they're split across the file rather than sitting with the code they explain.

Lines 27-31 — "which ones are hidden per environment … is controlled entirely by `NEXT_PUBLIC_TBV_DISABLED_BTC_WALLETS`, so no code change or redeploy is needed to toggle one." That was already half-true once the feature flag landed; the WebHID term makes it two hops off, since `ledger_btc_vault` is hidden per environment by neither of those things. One clause fixes it — e.g. "…apart from `ledger_btc_vault`, which has its own gate below."

Lines 39-42 then explain the flag term — and still say the provider is "built ahead of Ledger's firmware being final", which got updated in `featureFlags.ts:136` but not here — while lines 50-51 explain the WebHID term, with the exported `LEDGER_VAULT_WALLET_ID` JSDoc sitting between them. Worth folding both into a single block directly above this const, so the two terms of the gate are documented where they're composed.

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.disabledWallets.test.tsx
```diff
@@ -94,25 +94,45 @@ describe("WalletConnectionProvider — Ledger vault wallet opt-in", () => {
...
+  // jsdom's navigator has no `hid`, so every flag-off case stubs it: otherwise
+  // the WebHID term hides the entry on its own and the flag is never exercised.
+  const stubWebHid = () => {
+    Object.defineProperty(window.navigator, "hid", {
+      value: {},
+      configurable: true,
+    });
+  };
+
+  it("hides the Ledger vault wallet when the flag is unset, even with WebHID", async () => {
```
gbarkhatov, line 109: Optional, and I can see the argument against it. Adding `stubWebHid()` to the two flag-off cases is the right fix for what I raised last round — it makes the flag the term actually under test — but it *moved* those cases rather than adding one. On `origin/main` both ran with `hid` absent, i.e. they were the (flag off, no WebHID) quadrant. Coverage is now (off, present), (on, present) and (on, absent), with nothing in (off, absent) — which is the default for every non-Chromium browser.

It's trivially implied by both terms of the `&&` being false, so leaving it uncovered is defensible against the no-test-bloat rule. Just noting the quadrant was vacated rather than filled, in case you'd rather have the matrix complete.

Worth recording that the mutation gap I flagged last round is closed: deleting `featureFlags.isLedgerVaultWalletEnabled &&` from `VaultWalletConnectionProvider.tsx:57` now fails this test, and deleting `&& isWebHidAvailable` fails the one at :129. Both terms are load-bearing under the suite as it stands.

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.disabledWallets.test.tsx
```diff
@@ -94,25 +94,45 @@ describe("WalletConnectionProvider — Ledger vault wallet opt-in", () => {
...
   });
 
+  it("hides it without WebHID even when the flag is on", async () => {
+    // jsdom default: no navigator.hid — the Firefox/Safari case, where the
+    // entry would otherwise render as a dead hardware row.
+    await renderWithFlag("true");
+    expect(h.captured.disabledWallets).toContain("ledger_btc_vault");
+  });
+
   it("keeps the legacy staking Ledger adapters hidden either way", async () => {
```
gbarkhatov: "either way" no longer names anything. With a second axis it implies four combinations and exercises one — and because it renders `"true"` without `stubWebHid()`, it sits in the same quadrant as the test directly above it, making this a strict duplicate of that setup with a weaker assertion.

The assertion itself still holds, and the diff was right to leave it alone: both ids come unconditionally from `ALWAYS_DISABLED_WALLETS` (lines 32-37), and `toEqual(expect.arrayContaining([...]))` is a subset check, so it's deliberately insensitive to whether `ledger_btc_vault` is also present.

Adding `stubWebHid()` here would make the name true *and* cover the case that actually matters: in the one quadrant where `ledger_btc_vault` is revealed, nothing currently asserts that the legacy staking adapters stay hidden.

Minor, while you're in here: this test and :129 both depend on `hid` being absent, which is only true because :123's `afterEach` deleted it. It fails loudly rather than silently if that ever breaks, so it's acceptable — but an explicit line would let each test be read top-to-bottom on its own.

gbarkhatov (dismissed):
Approving. The five findings from the last round are genuinely fixed rather than papered over: the `wallet` probe and its unreachable test are gone, `featureFlags.ts` documents both terms of the gate, the presence-vs-truthiness split is moot now that only one gate exists, and the flag-off cases stub `hid` so the flag is the term actually under test. I mutation-checked both directions at `27a135c4e` — dropping either half of the `&&` on `VaultWalletConnectionProvider.tsx:57` fails a test — and the suite is 8/8 green.

**What I verified and cleared**, so it doesn't get re-litigated:

- **No SSR or hydration concern with the module-scope evaluation.** The vault is a Vite SPA — no `next.config.*`, no `"use client"` anywhere in `src`, no `hydrateRoot`, `index.html` ships an empty `#root`, and `vite.config.ts:117-131` is a plain static build. Module scope is also the better choice here: WebHID availability is fixed for the document's lifetime, and `DISABLED_WALLETS` having a stable identity matters because it is a dependency of `Chain.context.tsx:116` and `:262`. (Unrelated to this PR: `CLAUDE.md` describes this service as Next.js — it isn't.)
- **`disabledWallets` removes the row entirely**, it does not render it disabled — `core/index.ts:82` `continue`s before `createWallet`, so the `Wallet` is never constructed and never reaches `chain.wallets`.
- **The secure-context claim is right**, checked in a real browser rather than read off the IDL: `"hid" in navigator` is `false` on a non-secure origin and `true` on `http://localhost`. No separate `isSecureContext` test is needed.
- **The gate will not break a Playwright run.** `playwright.config.ts:81-89` is chromium-only, and both the full Chromium build and `chromium_headless_shell` expose `navigator.hid` on a secure origin, headless and headed.
- **"on in devnet" is accurate today** — the only wiring is `.github/workflows/service-release-vault.yml:199`, and `vault-devnet` has the variable set to `"true"` while staging-devnet, testnet and mainnet do not.
- **The deleted "signing is not implemented yet" text was genuinely stale**, and was already stale on `origin/main`: `signPsbt`, `signPsbts` and `signMessage` are all real implementations on `LedgerVaultProvider`.
- **Deleting the probe's test alongside the probe is correct** under the zero-dead-code rule, and no dangling references remain.

Nine comments below. The two worth acting on are the rationale on `VaultWalletConnectionProvider.tsx:51`, which describes the pre-`27a135c4e` behaviour rather than what the code now does, and the "full depositor flow" claim in `featureFlags.ts:135`, which overstates in the same direction the deleted line understated. There's also one stale sibling comment in the connector package that this doc pass didn't reach. The rest are doc precision and test naming.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/index.ts
jrwbabylonlab: fixed

services/vault/src/config/featureFlags.ts
jrwbabylonlab: fixed

services/vault/src/config/featureFlags.ts
jrwbabylonlab: fixed

services/vault/.env.example
jrwbabylonlab: fixed

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jrwbabylonlab: fixed; not planning — logger is breadcrumb-only, console is lint-banned

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jrwbabylonlab: fixed; fixed; not planning — spec property name, not a magic value

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jrwbabylonlab: fixed

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.disabledWallets.test.tsx
jrwbabylonlab: not planning — each false term is already covered alone

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.disabledWallets.test.tsx
jrwbabylonlab: fixed; not planning — :129 comment already states the jsdom default

jonybur (approved):
Both rounds of findings landed. The unreachable probe test is gone, the JSDoc no longer describes behaviour the connector does not control, and disabledWallets is the one gate on the dApp side. featureFlags.ts now names both terms of the gate (flag AND WebHID) rather than describing only the flag. VaultWalletConnectionProvider uses the presence check ('hid' in navigator), matching the wallet-metadata capability semantic. The disabledWallets test stubs navigator.hid on every flag-off case so the flag is the term actually under test, and adds the 'flag on but no WebHID' case. .env.example is accurate — signing is implemented and the flip is per-environment. The open replies on extracting the 'hid' string constant and filling the (flag-off, WebHID-absent) quadrant are defensible: 'hid' is a spec property name and each false term is already covered alone. CI green.

PR #2317, opened 2026-08-25
feat(ts-sdk, vault): reclaim the stranded depositor-claim reserve after settlement

### LINKED ISSUE #2276: Reclaim the depositor-claim reserve after a settled withdrawal
## Problem

Every peg-in reserves `depositorClaimValue` — roughly 33k sats on current devnet params — to fund the depositor's own `claim_tx`. That is the depositor's recourse if the vault provider fails during peg-out.

When the vault provider claims instead and the depositor withdraws normally, that reserve is never spent. Nothing in this repository can spend it, so it is stranded permanently.

`COPY.deposit.form.transactionReserveTooltip` in `services/vault/src/copy.ts` already tells the depositor the reserve "is returned to you if unused". Today that promise is kept only on the refund path.

## Where the reserve actually lives

It is **not** in the Pre-PegIn transaction. The Pre-PegIn HTLC output carries it inside its value:

```
htlcValue = peginAmount + depositorClaimValue + p2aAnchorValue + minPeginFee
```

It becomes a standalone UTXO only once the PegIn transaction is broadcast, landing at **PegIn vout 1**.

| Property | Value |
|---|---|
| Location | PegIn tx, vout 1 (`PEGIN_DEPOSITOR_CLAIM_VOUT`) |
| Value | `depositorClaimValue`, from `computeMinClaimValue` at runtime |
| Type | P2TR, NUMS internal key (no key-path spend) |
| Script tree | Exactly one tapleaf: `<depositorPubkey> OP_CHECKSIG` |
| Timelock | None |
| Counterparty paths | None |

Source: `deriveDepositorClaimScriptPubKey` in `packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts`, mirroring btc-vault's `SingleKeyConnector`.

## Why the refund path already returns it

The refund spends the Pre-PegIn HTLC via its timelock leaf, *before the PegIn transaction exists*. The reserve is still inside the HTLC value at that point, so it comes back as part of the refunded amount with no dedicated sweep.

Nothing in that path generalises to the settled case. Once the PegIn is broadcast the HTLC is gone and the refund code has nothing left to spend.

## Why this cannot go in the withdraw flow

The claim leaf carries no timelock, so the depositor could sweep the reserve at any moment with a single Schnorr signature. Nothing on chain prevents an early sweep — the only protection is application policy.

At the moment the depositor triggers a withdrawal the peg-out has not settled, and the reserve is precisely what funds the depositor's claim if the vault provider fails during the assert window.

> ⚠️ Offering the sweep before the peg-out settles gives the depositor a one-click way to destroy their own fallback against vault provider failure, at the one point where it matters. There is no on-chain protection against this and no way to undo it.

The withdraw flow also signs no Bitcoin transactions — it is a single Ethereum transaction — so the depositor's Bitcoin wallet may not even be connected there.

The action therefore has to be **post-settlement**.

## UI options — not yet decided, pending design

All four require the depositor's Bitcoin wallet to sign, and none can live inside the withdraw flow.

| # | Option | Strength | Weakness |
|---|---|---|---|
| 1 | Button on the vault row, in the Inactive section of `/vaults`, opening a review modal | Cheapest to build, identical to the existing refund pattern | Settled vaults sit in a section users stop visiting, so a small payout goes unnoticed |
| 2 | Aggregate banner — "0.00067 BTC reclaimable from 3 settled vaults → Reclaim all" | Best discovery, one signature for the whole batch | No per-vault control, and a persistent nag about a small amount irritates |
| 3 | Both — row button for control, banner for discovery | Banner drives discovery, button keeps it per-vault rather than a mystery sweep | Two surfaces to build and test |
| 4 | Line item inside an existing surface, no dedicated call to action | Lowest noise | Lowest discovery; most depositors would never reclaim |

Open sub-questions for design:

- The banner threshold, if a banner is used at all. One settled vault might warrant only a quiet row button.
- The **Ledger blocked state**, which must be visible and explained rather than hidden — see below.
- Whether to surface the action at all when current fee rates make it uneconomic.

`NotificationCard` in `services/vault/src/components/shared/NotificationCard.tsx` already has a `dust` tone that is designed and unit-tested but has no production consumer, if a banner is chosen.

A one-line note on the withdraw success screen setting the expectation is worth considering regardless of which option wins, so the later prompt does not read as an unexplained payout.

## Ledger: blocked by firmware

The Ledger vault app **cannot sign this transaction**, in either the single or batched shape. This is a firmware constraint, not a host-side gap.

In `app-babylon-vault`, `src/sign_psbt_validate.c`, the leaf-shape dispatcher routes any 34-byte `<D> OP_CHECKSIG` tapleaf to `_validate_display_claim`, which opens by requiring the protocol `claim_tx` shape:

```c
if (st->n_inputs != 1 || st->n_outputs != 2) {
    SEND_SW(dc, SW_INCORRECT_DATA);
    return false;
}
```

A 1-in/1-out sweep is rejected on `n_outputs != 2`; a batched sweep is rejected on `n_inputs != 1` as well. There is no alternative route, because the leaf shape is what selects the validator and that leaf is unique to this outpoint.

The action must be gated off for Ledger wallets, with an explanatory blocked state rather than a hidden button — otherwise those users never learn the reserve exists.

The gate must be reclaim-specific: refund *is* supported on the device (`_validate_display_refund` is reachable), so it must not be generalised to "no Bitcoin actions on Ledger".

## Eligibility

Offer the action only when all four hold. Any read that errors leaves the action hidden — this gate fails closed.

1. Indexer status is `ContractStatus.DEPOSITOR_WITHDRAWN`. Already polled.
2. On-chain corroboration: `getVaultFromChain(vaultId).status === OnChainBtcVaultStatus.REDEEMED`.
3. An esplora outspend check confirms `peginTxid:1` is unspent.
4. The vault is owned by the connected wallet (`isVaultOwnedByWallet`).

Conditions 1 and 2 are not redundant. The contract has no status 6 — the indexer refines the contract's `REDEEMED` into `DEPOSITOR_WITHDRAWN` once the vault UTXO spend is observed on Bitcoin. Requiring both means a compromised indexer cannot conjure eligibility for a vault that is still active.

Vault-provider-reported `PAYOUT_BROADCAST` is excluded deliberately: it is untrusted counterparty data, and it is also insufficient, since broadcast is not confirmed and that gap is exactly when the claim right still matters.

**Spend attribution.** Only `<D> OP_CHECKSIG` can spend this output, so any spend is by definition the depositor. The two depositor-authored spenders are the protocol `claim_tx` and this new sweep. Gate on the bare `spent` flag, but attribute for display by comparing the spending txid against a locally recorded reclaim txid, and label the unattributed case neutrally rather than asserting a reclaim happened.

## Sourcing the PegIn txid

`vaultId` is a one-way hash of the PegIn txid, so the txid cannot be recovered from it.

**Authoritative, and the only source fed into a signed transaction:** `VaultProtocolInfo.depositorSignedPeginTx` passed through `calculateBtcTxHash`. The SegWit txid excludes witness data, so this equals the broadcast PegIn's txid even though the vault provider adds its own witness. The pattern already ships in `services/vault/src/hooks/deposit/useVaultActions.ts`.

Parsing those same bytes also yields `outs[1].script` and `outs[1].value` directly, at no extra RPC cost.

The indexer's `activity.peginTxHash` is adequate for driving the outspend poll but must never reach a signed transaction. Local storage is cleared once the vault reaches a terminal status, so it is unavailable by reclaim time.

## Safety requirements

This path constructs and signs Bitcoin transactions moving real funds. It falls under `CLAUDE.md` critical paths #1, #8 and #9 — two reviewers required.

**Script binding, a three-way bind.** Require byte equality across the contract's parsed `depositorSignedPeginTx` `outs[1]`, the esplora UTXO lookup, and a JS re-derivation from the depositor pubkey. Any disagreement aborts.

The re-derivation input must be the **connected wallet's live pubkey**, not the indexer's `depositorBtcPubkey`. That is what makes the assertion meaningful — it proves the wallet about to sign is the wallet that can spend.

**Value binding.** Bind each input to the expected `computeMinClaimValue` result, and the output to `sum(inputs) - fee` exactly.

**Fee caps.** Port the refund's rate and fraction guardrails, rejecting rather than clamping.

**Signature verification.** `packages/babylon-ts-sdk/src/tbv/core/utils/signing.ts` uses `useTweakedSigner: false` and `autoFinalized: false`, so every signature must be validated against the expected sighash **before** finalize — for every input, not just input 0.

## Fee economics

Every witness element is fixed length — a 64-byte BIP-340 signature, a 34-byte leaf script, a 33-byte control block (single leaf, no merkle path) — so vsize is exact rather than conservative:

```
weight(N) = 214 + 299 × N
vsize(N)  = ceil(weight / 4)
```

| Inputs | vsize | Marginal |
|---|---|---|
| 1 | 129 vB | — |
| 2 | 203 vB | +74 |
| 3 | 278 vB | +75 |
| 4 | 353 vB | +75 |

At ~33 000 sats per vault, a single-input sweep costs 2.0% of the reserve at 5 sat/vB, 9.8% at 25, 39% at 100, and the entire reserve at 256 sat/vB.

Batching is the lever: three vaults at 50 sat/vB costs 13 900 sats to reclaim 99 000 (14%), against 19 350 sats (19.5%) done singly, plus one wallet approval instead of three.

> ⚠️ The fee-fraction basis here is the swept amount itself. In the refund the basis is the deposit and the swept amount is much larger, which is why the refund's 10% cap is right there and wrong here — it would block every sweep above roughly 25 sat/vB.

Blocking is safe and the copy should say so: the reserve is not at risk, it simply stays where it is until fees drop. That is a materially different posture from the refund, where blocking leaves BTC locked.

## Implementation sketch

The refund flow is the closest existing analogue; follow it file for file.

- **SDK** — extract the private `deriveDepositorClaimScriptPubKey` into a shared descriptor returning `{ scriptPubKey, leafScript, controlBlock }` so one definition serves both peg-in validation and the sweep. Add an N-in/1-out PSBT builder and a service mirroring `buildAndBroadcastRefund`.
- **State machine** — new `PeginAction`, display branch, primary-action entry, badge priority (lowest — nothing is at risk and nothing expires).
- **Eligibility hook** — sibling of `useRefundRowAction`, returning `{ available, blockedTooltip }`.
- **Modal** — sibling of `RefundModal/`, same four render states.
- **Copy** — all strings in `copy.ts`. The action is called "Reclaim", not "Withdraw", which already carries three meanings in this app.

> Settled vaults render **nowhere** on `/vaults` today. `VaultsActiveSection` is `ACTIVE`-only and `useAaveVaults` surfaces `REDEEMED` but not `DEPOSITOR_WITHDRAWN`. The rows this feature attaches to do not yet exist. Extend the inactive section rather than adding a new one.

## Open questions

- **`claimExpiredUntil` semantics.** Read on chain but never surfaced, carrying a `TODO(#1690)`. If it is the block deadline past which the depositor's claim right is void, it is arguably a second, independent sufficient condition for reclaim — the claim can no longer be exercised, so the reserve is dead weight regardless of settlement. Whether it is depositor- or claimer-scoped, when it is stamped, and whether it is populated on devnet could not be established from this repository. Needs contracts input.
- **`LIQUIDATED` and `INVALID` vaults.** Also terminal, and the reserve is equally the depositor's own key. Excluded pending an answer on whether the depositor's claim graph still has a role after liquidation.
- **Protocol-side alternative.** The cleaner long-term fix may be folding the reserve return into the transaction graph at deposit time rather than sweeping from the frontend. That is a btc-vault change and would remove this feature's need to exist. Worth deciding before building.
- **Ledger firmware roadmap.** Whether a Reclaim validator is planned, which determines whether the blocked-state copy should imply the limitation is temporary.
- **UI direction.** Pending design — see the options table above.

gbarkhatov:
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2276

https://github.com/user-attachments/assets/a92adc6a-d7bf-4fe0-b2d6-eec19a43120e

## What this fixes

Every peg-in quietly sets aside a small amount of the depositor's BTC — around 26,000–33,000 sats depending on the signer set and fee rate. That money funds the depositor's own `claim_tx`, which is their fallback if the vault provider fails during peg-out.

On the happy path the vault provider claims using its own wallet, the depositor withdraws normally, and that set-aside amount is never touched by anyone. Nothing in this repository could spend it, so it was stranded on Bitcoin permanently. Roughly $25 disappeared on every successful deposit.

The deposit screen has been promising the opposite the whole time. `COPY.deposit.form.transactionReserveTooltip` told depositors the reserve "is returned to you if unused" — a promise only the refund path could keep.

This PR adds a **Reclaim** action that sweeps that reserve back to the depositor once their vault has settled, and rewrites the deposit-time tooltip to describe what actually happens.

## Where the money actually is

This was the first thing worth getting right, because it is easy to assume the reserve sits somewhere it does not.

The reserve is **not** a separate output of the Pre-PegIn transaction. It is carried *inside* the Pre-PegIn HTLC output's value, and only becomes a standalone UTXO once the PegIn transaction is broadcast — landing at **PegIn output index 1**.

That output is a Taproot output with an unspendable internal key and exactly one script leaf: `<depositorPubkey> OP_CHECKSIG`. No timelock, no covenant, no counterparty. The depositor can spend it alone, at any time, with a single signature.

That also explains why the refund path already returns this money and why none of it could be reused: the refund spends the whole HTLC output *before the PegIn ever exists*, so the reserve comes back as part of the refunded amount. Once the PegIn is broadcast the HTLC is gone and the refund code has nothing left to spend. The reclaim transaction is genuinely new code.

## The important safety decision

Since the depositor can spend this output at any moment, the real question is not *can we* but *when should we offer it*.

Spending it early is worse than it sounds. The depositor's pre-signed `claim_tx` hard-codes this exact outpoint as its only input, and every downstream pre-signed transaction (Assert, Payout, NoPayout, the challenge transactions) commits to that transaction's id. There is no way to re-fund it. **Sweeping the reserve while the vault is still live permanently destroys the depositor's entire pre-signed recovery set.**

So the gate has to be certain the recovery set is already useless.

### Approach A — gate on the Ethereum vault status (what the issue proposed)

Offer the action once the contract reports `Redeemed` and the indexer reports `DEPOSITOR_WITHDRAWN`.

**Rejected, because it is not safe.** Reading the contracts showed that `RedeemLogic.redeemForDepositor` sets `Redeemed` *before* any Bitcoin claim has happened, and then emits `VaultClaimableBy` for **both** the vault provider's key and the depositor's key. That is permission granted to two parties, not a record of who acted. There is no `markClaimed`, no payout proof, and no claim-completion event anywhere in `BTCVaultRegistry.sol` — nothing on Ethereum distinguishes "the vault provider claimed" from "the depositor claimed".

A depositor could therefore withdraw, see the button appear immediately, sweep the reserve, and destroy their fallback during exactly the window where the vault provider might still fail them.

### Approach B — gate on Bitcoin (chosen)

Offer the action only once **PegIn output 0 (the vault UTXO) has been spent and is at least 6 confirmations deep**.

That spend *is* the Payout. Once the vault UTXO is consumed, the depositor's claim graph can no longer serve any purpose, so the reserve is provably dead weight. This is the only signal that is both trustworthy and verifiable, and it is checked directly against Bitcoin rather than taken from any indexer.

The full gate is: contract status `Redeemed`, **and** PegIn output 0 spent with ≥6 confirmations, **and** PegIn output 1 still unspent, **and** the vault belongs to the connected wallet. Any read that fails or has not loaded hides the action rather than showing it — it fails closed.

The indexer's `DEPOSITOR_WITHDRAWN` status is still used, but only as a cheap filter to decide *which* vaults are worth checking on Bitcoin. It never decides eligibility.

There is a comment at the gate explaining why, and a test named for the exact case Approach A would have wrongly allowed, so nobody "simplifies" this back into a status read later.

## What the depositor sees

A **Reclaim** button on the settled vault's row in the Inactive Vaults section, with the reclaimable amount beside it, then a review modal and a success screen.

The row keeps three states, mirroring how the refund path already shows Refunding → Refunded:

| Reserve state | Status | Amount shown | Button |
|---|---|---|---|
| Unspent, eligible | `Redeemed` | `26,228 sats / reclaimable` | Reclaim, enabled |
| Sweep sent, not yet confirmed | `Reclaiming` | `26,228 sats / reclaimable` | Reclaim, disabled |
| Sweep confirmed | `Redeemed` | — | — |

The row stays in the list permanently once reclaimed, which matches what an already-refunded expired vault does today.

## Deliberate differences from the design file

Three, all agreed with design beforehand and worth a look during review.

**The review modal gained fee rows.** As drawn it showed only the gross amount and a Confirm button. At ~26,000 sats the network fee is 2% of the reserve at 5 sat/vB and 39% at 100 sat/vB, so the number on screen was not what the depositor would receive — and the "fee is too high" states had nowhere to appear. The modal now shows the fee rate, the network fee, and the net amount received, reusing the refund modal's existing components. The design component already had these rows built and hidden, so this turns them on rather than inventing anything.

**The section keeps its name and the row says "Redeemed".** The design titled the section "Expired Deposits" and gave the reclaim row an "Expired" status. Both are wrong for a vault whose peg-out completed successfully — it did not expire. The section stays "Inactive Vaults" and the row uses the existing "Redeemed" label.

**Typo fix.** The design reads "Reclain Amount"; the copy says "Reclaim Amount".

Separately, and *not* changed here: the design shows the inactive row's Withdraw button in red, but the app currently renders it orange — only the Activity page uses red today. That is a pre-existing inconsistency on a surface this PR does not otherwise touch, so it is flagged rather than fixed.

## Fee handling

The transaction shape is completely fixed, so its size is exact rather than estimated: **129 vbytes** for a single-input sweep, plus 75 vbytes per additional input. This is shipped as a function with both terms named, not a magic number, and there is a test that builds a real transaction and measures it rather than trusting the arithmetic.

The existing peg-in fee helper is deliberately **not** reused — its P2TR input size assumes a key-path spend, and would under-fee this shape by about 25%.

Two safety caps mirror the refund path's, and reject rather than silently clamp: a maximum fee rate, and a maximum fee as a fraction of the amount being swept. One thing reviewers should not "tidy":

> The fee-fraction cap here is measured against the **swept reserve itself** (block above 25%, warn above 10%). The refund measures its cap against the much larger deposit, which is why its 10% is generous there and would be far too strict here — it would block every reclaim above roughly 25 sat/vB. There is a comment saying so at both the SDK and UI cap sites. Do not align the two.

Blocking is also safe here in a way it is not for the refund: nothing is at risk and nothing expires, the reserve simply stays where it is until fees drop. The copy says exactly that.

## Safety checks before signing

This path builds and signs a Bitcoin transaction that moves real money, so it falls under the critical-path rules in `CLAUDE.md` and needs two reviewers.

- **Three-way check on the output being spent.** The script and value must match across the contract's own copy of the signed PegIn transaction, an independent lookup of the UTXO on Bitcoin, and a fresh derivation in JavaScript. Any disagreement aborts before the wallet is asked to sign.
- **The derivation uses the connected wallet's live public key**, never the indexer's copy. That is what makes the check meaningful: it proves the wallet about to sign is the wallet that can actually spend, and turns a wrong-wallet situation into a clear error instead of an unspendable broadcast.
- **Value is pinned**: the input must equal the independently recomputed reserve value, and the output must equal inputs minus fee exactly.
- **Every signature is verified** against a recomputed sighash before the transaction is finalised — not just the first input. The wallet reporting success is not treated as evidence, because the signing flags used here are known to fail silently on some wallets.

## Two bugs found along the way

**A real validation gap in existing code.** `deriveDepositorClaimScriptPubKey` accepted a malformed public key — the script compiler emits a shorter push for a truncated key, producing a well-formed script that is not the claim leaf. That function was already live on the peg-in validation path. It now rejects anything that is not a 32-byte key.

**A duplicate-input case.** The builder now rejects a batch naming the same reserve twice with a clear message, instead of failing deep inside the PSBT library.

## Ledger is out of scope

cc @jrwbabylonlab 

The Ledger BTC Vault app **cannot sign this transaction**, and this is a firmware limitation rather than something fixable here. Its PSBT validator routes any 34-byte `<D> OP_CHECKSIG` leaf to the claim-transaction validator, which requires exactly 1 input and 2 outputs; a 1-in/1-out sweep is rejected on the output count before any device interaction.

A workaround exists and was **deliberately rejected**: output 0's script is not validated, so a 1-in/2-out transaction would pass — but the device would then display a "Claim transaction" approval for something that is not a claim, with no destination shown. That defeats the purpose of a hardware wallet.

So the action is gated off for Ledger wallets, showing a disabled button with an explanation rather than hiding it, so those users still learn the reserve exists and can sweep it from another wallet. The gate is deliberately specific to reclaim — refund *is* supported on the device and must keep working. Supporting this properly needs a new firmware validator with its own approval screen plus matching host-side work.

## Verified against real chain state

Tested end to end against a genuinely settled vault on testnet (signet), not just mocks:

- The PegIn transaction id derived from the contract's stored transaction matched the indexer's independently — confirming the id source is correct.
- The claim script derived in JavaScript came out **byte-identical** to the output actually sitting on signet, which validates the derivation against real Rust-produced output rather than only against our own test expectations.
- The gate correctly recognised the vault as eligible: status `Redeemed`, vault output spent 12,750 confirmations deep, reserve output unspent at 26,228 sats.
- The row rendered, the reclaim was performed, and the money arrived.

## Tests

- **SDK, 41 new.** The script derivation is checked against a second, independent BIP-341 implementation over a pinned vector **plus 64 randomised keys** — a single hardcoded vector pins one input, not the function. Plus one rejection test per safety check, value conservation, a two-input batch, and the measured transaction-size test.
- **App, 26 new.** Every eligibility state including the unsafe one Approach A would have allowed; the reclaiming and settled row states; and the modal's four screens.
- Full suites green: 1,735 SDK tests, 3,589 vault tests, lint clean, production build passes.

## Follow-ups, not in this PR

- Batching several reserves into one transaction. The transaction builder already takes a list of inputs, so this is a later change rather than a rewrite. Three vaults swept together cost about 14% of the total in fees versus 19.5% done one at a time. cc @jrwbabylonlab 
- Surfacing the action on the Activity page. That feed is an event log, so adding a call-to-action to a row that reads as finished is a product decision rather than a mechanical port.
- Mapping the design components to code via Code Connect.
- `LIQUIDATED` and `INVALID` vaults are excluded pending an answer on whether the depositor's claim graph still has a role after liquidation.
- `VaultProtocolInfo.claimExpiredUntil` is read on chain but never used, and may be a second independent reason the reserve is reclaimable. Its wording points at the HTLC grace window instead, so it needs input from the contracts side before anything relies on it.
- The refund path calls `signPsbt` without the device terms approval, so it would already fail on a Ledger vault wallet today. Unrelated to this change and worth its own issue.

### DIFF
--- a/packages/babylon-ts-sdk/docs/api/services.md
+++ b/packages/babylon-ts-sdk/docs/api/services.md
@@ -177,6 +177,83 @@ Error.constructor
 
 ***
 
+### ReclaimUneconomicalError
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/errors.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/errors.ts)
+
+Thrown when the fee would consume too much of the swept reserve — either the
+per-vbyte rate exceeds the safety ceiling, or the absolute fee exceeds the
+fraction cap.
+
+Distinct from a generic error because the caller's response differs: nothing
+is at risk and nothing expires. The reserve simply stays where it is until
+fee rates fall, and the UI should say so rather than presenting a failure.
+
+#### Extends
+
+- `Error`
+
+#### Constructors
+
+##### Constructor
+
+```ts
+new ReclaimUneconomicalError(
+   message, 
+   feeSats, 
+   sweptTotalSats): ReclaimUneconomicalError;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/errors.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/errors.ts)
+
+###### Parameters
+
+###### message
+
+`string`
+
+###### feeSats
+
+`bigint`
+
+###### sweptTotalSats
+
+`bigint`
+
+###### Returns
+
+[`ReclaimUneconomicalError`](#reclaimuneconomicalerror)
+
+###### Overrides
+
+```ts
+Error.constructor
+```
+
+#### Properties
+
+##### feeSats
+
+```ts
+readonly feeSats: bigint;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/errors.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/errors.ts)
+
+Fee the reclaim would have paid, in satoshis.
+
+##### sweptTotalSats
+
+```ts
+readonly sweptTotalSats: bigint;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/errors.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/errors.ts)
+
+Sum of the reserves the reclaim would have swept, in satoshis.
+
+***
+
 ### BIP68NotMatureError
 
 Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/refund/errors.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/refund/errors.ts)
@@ -2367,6 +2444,150 @@ drift.
 
 ***
 
+### ReclaimVaultData
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+One reserve to sweep, as the caller resolves it.
+
+#### Properties
+
+##### depositorSignedPeginTxHex
+
+```ts
+depositorSignedPeginTxHex: string;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+The contract's own copy of the depositor-signed PegIn transaction
+(`VaultProtocolInfo.depositorSignedPeginTx`). Must come from the chain —
+never the indexer. Its `outs[1]` is one leg of the three-way bind.
+
+##### observed
+
+```ts
+observed: object;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+Independent chain observation of `peginTxid:1`.
+
+###### scriptPubKey
+
+```ts
+scriptPubKey: string;
+```
+
+###### value
+
+```ts
+value: bigint;
+```
+
+##### expectedClaimValue
+
+```ts
+expectedClaimValue: bigint;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+This vault's reserve value, recomputed via `computeMinClaimValue`.
+
+***
+
+### ReclaimInput
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+#### Type Parameters
+
+##### R
+
+`R` *extends* [`BtcBroadcastResult`](#btcbroadcastresult) = [`BtcBroadcastResult`](#btcbroadcastresult)
+
+#### Properties
+
+##### vaultIds
+
+```ts
+vaultIds: `0x${string}`[];
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+##### depositorBtcPubkey
+
+```ts
+depositorBtcPubkey: string;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+The **connected wallet's live** BTC pubkey — compressed sec1 or x-only.
+Never the indexer's `depositorBtcPubkey`: re-deriving the claim script
+from the live key is what proves the wallet about to sign is the wallet
+that can spend.
+
+##### readVaults()
+
+```ts
+readVaults: () => Promise<ReclaimVaultData[]>;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+Resolve the reserves to sweep, in the same order as `vaultIds`. The SDK
+passes no arguments — the caller closes over whatever context it needs.
+
+###### Returns
+
+`Promise`\<[`ReclaimVaultData`](#reclaimvaultdata)[]\>
+
+##### feeRate
+
+```ts
+feeRate: number;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+Mempool-derived sat/vB fee rate. Caller fetches it before invoking.
+
+##### signPsbt
+
+```ts
+signPsbt: ReclaimPsbtSigner;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+BTC wallet signer; receives a PSBT hex + taproot script-path options.
+
+##### broadcastTx
+
+```ts
+broadcastTx: BtcBroadcaster<R>;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+Broadcast callback — returns whatever shape the caller needs.
+
+##### signal?
+
+```ts
+optional signal: AbortSignal;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+Checked at every async boundary.
+
+***
+
 ### VaultBatchEntry
 
 Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/refund/buildAndBroadcastRefund.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/refund/buildAndBroadcastRefund.ts)
@@ -2823,6 +3044,30 @@ type KeyResolutionMode =
 Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/participants/types.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/participants/types.ts)
 
 How a [ParticipantKeySet](#participantkeyset) was resolved. Carried for diagnostics.
+
+***
+
+### ReclaimPsbtSigner()
+
+```ts
+type ReclaimPsbtSigner = (psbtHex, opts) => Promise<string>;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+#### Parameters
+
+##### psbtHex
+
+`string`
+
+##### opts
+
+[`SignPsbtOptions`](managers.md#signpsbtoptions)
+
+#### Returns
+
+`Promise`\<`string`\>
 
 ***
 
@@ -3886,6 +4131,49 @@ thresholds) are a consumer-side concern.
 
 ***
 
+### buildAndBroadcastReclaim()
+
+```ts
+function buildAndBroadcastReclaim<R>(input): Promise<R>;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+Build, sign, and broadcast a reclaim transaction sweeping one or more
+depositor-claim reserves back to the depositor's BIP-86 address.
+
+#### Type Parameters
+
+##### R
+
+`R` *extends* [`BtcBroadcastResult`](#btcbroadcastresult) = [`BtcBroadcastResult`](#btcbroadcastresult)
+
+#### Parameters
+
+##### input
+
+[`ReclaimInput`](#reclaiminput)\<`R`\>
+
+#### Returns
+
+`Promise`\<`R`\>
+
+whatever the injected `broadcastTx` returns (generic pass-through)
+
+#### Throws
+
+[ReclaimUneconomicalError](#reclaimuneconomicalerror) if the fee breaches either cap
+
+#### Throws
+
+`Error` if any validation or script/value bind fails
+
+#### Throws
+
+anything `readVaults`, `signPsbt`, or `broadcastTx` throws
+
+***
+
 ### estimateRefundFeeSats()
 
 ```ts
@@ -4203,6 +4491,64 @@ by a client bug), and costs ~1.6 min at 12s slots. Deliberately not the
 benefit. This is a liveness guard against an orphaned registration, not a
 theft mitigation — every Pre-PegIn HTLC spend path requires the depositor's
 own BTC key regardless.
+
+***
+
+### RECLAIM\_MAX\_FEE\_RATE\_SATS\_VB
+
+```ts
+const RECLAIM_MAX_FEE_RATE_SATS_VB: 2000 = 2000;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+Hard upper bound on the per-vbyte fee rate the SDK will sign a reclaim at.
+Same reasoning and value as the refund's cap: a compromised mempool endpoint
+can legally return up to 10,000 sat/vB, and 2000 leaves margin over the
+worst historical `halfHourFee` while still blocking that by 5×.
+
+***
+
+### RECLAIM\_MAX\_FEE\_FRACTION\_NUMERATOR
+
+```ts
+const RECLAIM_MAX_FEE_FRACTION_NUMERATOR: 25n = 25n;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+Hard upper bound on the absolute fee as a fraction of the **swept total**.
+
+> ⚠️ The basis here is deliberately different from the refund's. There the
+> basis is `vault.amount` and the swept amount is far larger, so a 10% cap is
+> generous. Here the basis *is* the swept amount — roughly 33k sats — and a
+> 10% cap would block every reclaim above about 25 sat/vB. Do not "fix" this
+> into symmetry with `REFUND_MAX_FEE_FRACTION_*`; it would silently change
+> behaviour. The matching comment lives at the UI cap site.
+
+***
+
+### RECLAIM\_MAX\_FEE\_FRACTION\_DENOMINATOR
+
+```ts
+const RECLAIM_MAX_FEE_FRACTION_DENOMINATOR: 100n = 100n;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+***
+
+### RECLAIM\_WARN\_FEE\_FRACTION\_NUMERATOR
+
+```ts
+const RECLAIM_WARN_FEE_FRACTION_NUMERATOR: 10n = 10n;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts)
+
+Fraction of the swept total above which the UI warns but still allows the
+reclaim. Exported so the review screen derives its threshold from the same
+constant the SDK enforces against, rather than restating it.
 
 ***
 
--- a/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/reclaim.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/reclaim.ts
@@ -1,0 +1,312 @@
+/**
+ * Reclaim PSBT Builder Primitive
+ *
+ * Builds an unsigned PSBT that sweeps the depositor-claim reserve — PegIn
+ * vout 1 — back to the depositor's own BIP-86 address.
+ *
+ * The reserve funds the depositor's `claim_tx`, their recourse if the vault
+ * provider fails during peg-out. When the vault provider claims instead and
+ * the depositor withdraws normally, nothing ever spends it. This primitive is
+ * the only thing in the repository that can.
+ *
+ * Spend path: the single `<depositor> OP_CHECKSIG` tapleaf (see
+ * ./depositorClaim). No timelock, no counterparty signature — the depositor's
+ * Schnorr signature alone. That is also why the *caller* must gate: this
+ * builder will happily sweep a reserve whose vault is still live, which would
+ * permanently void the depositor's pre-signed recourse graph. Eligibility is
+ * enforced upstream, in the service.
+ *
+ * @module primitives/psbt/reclaim
+ */
+
+import { Buffer } from "buffer";
+import { Psbt, Transaction } from "bitcoinjs-lib";
+
+import { DUST_THRESHOLD } from "../../utils/fee/constants";
+import {
+  TAPSCRIPT_LEAF_VERSION,
+  deriveBip86ScriptPubKeyHex,
+  hexToUint8Array,
+  stripHexPrefix,
+  uint8ArrayToHex,
+} from "../utils/bitcoin";
+import {
+  PEGIN_DEPOSITOR_CLAIM_VOUT,
+  deriveDepositorClaimDescriptor,
+} from "./depositorClaim";
+
+/**
+ * nVersion for the reclaim transaction. Plain v2, not TRUC: the reserve is only
+ * offered once its PegIn is deeply confirmed, so the v2/v3 PegIn's nVersion-3
+ * topology limits no longer constrain its children.
+ */
+const RECLAIM_TX_VERSION = 2;
+
+/**
+ * nSequence for every reclaim input: RBF-enabled, no relative timelock. The
+ * claim leaf carries no CSV, so this is a fee-bumping signal only.
+ */
+const RECLAIM_INPUT_SEQUENCE = 0xfffffffd;
+
+/**
+ * Weight units of the reclaim transaction's skeleton — everything except the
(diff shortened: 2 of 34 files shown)


services/vault/src/hooks/deposit/useReclaimState.ts
```diff
@@ -0,0 +1,156 @@
...
+        if (!btcWalletProvider || !connectedBtcAddress) {
+          setError("BTC wallet not connected");
+          return;
+        }
+        if (!vaultId) {
+          setError("Missing BTC Vault ID");
+          return;
+        }
+        if (!Number.isFinite(feeRate) || feeRate <= 0) {
+          setError("Fee rate must be a positive number");
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in 8149059.

Valid — this broke the explicit rule in `CLAUDE.md` that all user-facing strings live in `copy.ts`. The four inline messages ("BTC wallet not connected", "Missing BTC Vault ID", "Fee rate must be a positive number", and the generic failure fallback) now live under `COPY.reclaim.errors` and are referenced through `COPY`.

Worth noting for context: `useRefundState` has the same inline pattern today. I left it alone since it is a separate surface, but it is the same fix if someone wants to follow up.

services/vault/src/components/deposit/ReclaimModal/ReclaimReviewContent.tsx
```diff
@@ -0,0 +1,284 @@
...
+      return;
+    }
+    setFeeRate(FALLBACK_FEE_RATE_SATS_VB);
+    setUsingFallback(true);
+  }, [defaultFeeRateSatsVb, feeRate]);
+
+  const handleFeeRateChange = (next: number) => {
+    setFeeRate(next);
+    setUsingFallback(false);
+  };
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in 8149059.

Confirmed as a real bug rather than a theoretical one. The preview query has no `staleTime` and refetches on window focus, so the failed-then-recovered path is reachable: the fee fetch fails, the hard-coded floor is seeded with `usingFallback = true`, the user tabs away and back, the refetch succeeds — and the effect early-returned on `feeRate !== null`, leaving the warning on screen with Confirm disabled until the user typed a rate by hand.

Applied essentially your suggested shape: adopt a newly-available rate while the field still holds the fallback seed, and never overwrite a rate the depositor typed themselves.

One thing this surfaces beyond this PR: `RefundReviewContent` has the identical logic and therefore the identical latent bug — this code was modelled on it. Not touched here since it is a different surface, but it should get the same fix.

services/vault/src/hooks/deposit/useReclaimModal.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const handleSuccess = useCallback(() => {
+    setReclaimingActivity((current) => {
+      if (current) {
+        const vaultId = current.id.toLowerCase();
+        setInFlightVaultIds((prev) => new Set(prev).add(vaultId));
+      }
+      return null;
+    });
+    onSuccess();
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in 8149059.

You were right that the semantics were wrong. `ReclaimModal` reaches a terminal screen on two different paths — a sweep this session broadcast, and a reserve that turned out to be *already* spent (another device, or long ago) — and both were routed through `onSuccess`, which added the vault to `inFlightVaultIds`. That made the row claim credit for work it had not done, showing "Reclaiming" for someone else’s sweep.

Split into a separate `onBroadcast(vaultId)` callback fired only on the broadcast path; `handleClose`/`handleSuccess` still handle refresh-and-close for every completion. Two tests pin it: one asserting the callback fires after a broadcast, one asserting it does not fire on the already-settled path.

services/vault/src/hooks/useReclaimVaultChainData.ts
```diff
@@ -0,0 +1,78 @@
...
+export function useReclaimVaultChainData(
+  vaultIds: ReadonlyArray<string>,
+): Map<string, ReclaimVaultChainData> {
+  const stable = useMemo(
+    () => Array.from(new Set(vaultIds.map((id) => id.toLowerCase()))).sort(),
+    [vaultIds],
+  );
+
+  const query = useQuery({
+    queryKey: [RECLAIM_CHAIN_DATA_QUERY_KEY, stable.join(",")] as const,
```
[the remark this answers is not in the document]
gbarkhatov: Partially applied in 8149059.

Took the two that carry real risk:

- `data-testid="vault-reclaim-button"` was duplicated across the enabled and disabled renders of the same control. That is a load-bearing E2E hook, so a rename could silently update one and not the other. Now a single `RECLAIM_BUTTON_TEST_ID` constant with a comment pointing at the consuming action file.
- The `w-[82px]` reserve-column width is a genuine magic number. Now `RECLAIM_METRIC_COLUMN_CLASS`, with a comment explaining why the cell sits outside the fixed action slot and what the width has to fit.

Respectfully skipping the rest, with reasons rather than silently:

- **Query-key join separator.** `stable.join(",")` is building a cache key, not user-facing or protocol-relevant data. Naming the comma would add indirection at the call site without making anything safer or clearer.
- **Reclaim-state discriminant values.** These are the discriminants of a TypeScript discriminated union (`{ type: "available" } | { type: "reclaiming" } | ...`). String literals *are* the idiom there, and the compiler already checks them exhaustively — replacing them with constants would weaken narrowing and make the union harder to read.
- **Containing-block confirmation count.** The `+ 1` in `payoutConfirmations` is "the containing block counts as one confirmation", which is already stated in a comment directly above it. A named constant would restate the comment, not replace it.
- **Hex-prefix pattern and connector chain id.** `/^0x/i` and `"BTC"` follow the conventions already used across this codebase (`utils/vaultWarnings.ts`, every `useChainConnector` call site). Deviating only in the new files would make the reclaim flow the odd one out.

Happy to revisit any of these if the team would rather be uniform about it.

services/vault/src/hooks/useReclaimVaultChainData.ts
```diff
@@ -0,0 +1,78 @@
...
+              {
+                peginTxid: derivePeginTxid(vault.depositorSignedPeginTx),
+                onChainStatus: vault.status,
+              },
+            ];
+          } catch {
+            // Fails closed: without the contract read the gate withholds the
+            // action rather than trusting the indexer's status.
+            return null;
+          }
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in 8149059 — and this was the sharpest catch of the review, thank you.

The analysis is exactly right. `getVaultFromChain` failures were caught per-vault and turned into `null`, so the query resolved **successfully** with that vault simply missing from the map. Combined with `staleTime: 5min` and `refetchOnWindowFocus: false`, there was no trigger to ever retry — a single transient RPC blip would hide Reclaim for five minutes on an open page, with no way for the user to recover short of a reload.

The fail-closed intent was right (no contract read means no action offered, rather than trusting the indexer’s status) but the recovery path was broken.

Now:
- the read is no longer swallowed, so a failure rejects the query and React Query retries with backoff (`retry: 3`);
- `refetchOnWindowFocus` is enabled — an errored query is always stale, so returning to the tab recovers it;
- the gate stays fail-closed throughout: while the query is failing there is no data, so no action is offered.

The reasoning is written into the comment at the `return` so the per-vault catch does not get reintroduced as a "resilience" improvement later.

packages/babylon-ts-sdk/docs/api/services.md
```diff
@@ -3886,6 +4131,49 @@ thresholds) are a consumer-side concern.
...
+
+[ReclaimUneconomicalError](#reclaimuneconomicalerror) if the fee breaches either cap
+
+#### Throws
+
+`Error` if any validation or script/value bind fails
+
+#### Throws
+
+anything `readVaults`, `signPsbt`, or `broadcastTx` throws
```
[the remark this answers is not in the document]
gbarkhatov: Skipping this one, with reasoning rather than silently.

**The file is generated.** `packages/babylon-ts-sdk/docs/api/services.md` is typedoc output — `typedoc.json` sets `"out": "docs/api"`, and it is rebuilt by `pnpm docs:generate`. Hand-editing the markdown to merge the headings would be overwritten the next time anyone regenerates, so the fix could only live at the source.

**At the source, the current form is the correct idiom.** The three headings come from three `@throws` tags on `buildAndBroadcastReclaim`, which is how TSDoc expresses distinct failure modes — here: the typed `ReclaimUneconomicalError`, validation/bind failures, and pass-through errors from the injected callbacks. Collapsing them into one tag with bullets would lose the per-error semantics that tooling reads, and would make the reclaim service inconsistent with `buildAndBroadcastRefund` sitting directly beside it, which uses three `@throws` the same way.

**MD024 is not enforced in this repository.** There is no markdownlint config in the tree and no markdownlint step in CI; the warning comes from the bundled `markdownlint-cli2` in this review tool rather than from the project. The generated file already contains 25 `#### Throws` headings across the existing API surface, so this PR is following the established pattern rather than introducing it.

If the team does want the generated docs MD024-clean, the right fix is repo-wide and belongs outside a feature PR — either a typedoc option that renders multiple `@throws` under a single heading, or a markdownlint ignore for `docs/api/**`. Happy to open an issue for that if it is wanted.

jonybur (approved):
The reclaim primitive holds together. Every input goes through a three-way bind before it reaches the PSBT — contract `depositorSignedPeginTx` bytes × chain observation × JS re-derivation from the *live* wallet key (`processPublicKeyToXOnly` + `deriveDepositorClaimDescriptor`) — on both script and value, and any disagreement throws with a message that names the input and the two disagreeing values. Duplicate outpoints are rejected explicitly. Fee is capped twice at the service layer (rate ≤ 2000 sat/vB, absolute ≤ 25% of swept), then again inside the primitive against dust and against `feeSats >= totalInputValue`. Vsize is derived from exact witness sizing (64-byte BIP-340 sig, 34-byte leaf, 33-byte control block, no merkle path), not a conservative estimate.

Signature verification loops over every input (`assertScriptPathSchnorrSignature` per index), which is the CLAUDE.md #8 rule for `useTweakedSigner: false` / `autoFinalized: false` script-path signing. `assertPsbtUnsignedTxMatches` gates the returned PSBT against what we asked for. `extractPayoutSignature` is reused rather than a parallel extractor — worth noting since it's now doing double duty, but the shape is the same (single sig per script-path input under SIGHASH_DEFAULT).

Eligibility is deliberately on-chain — PegIn vout 0 spent and deeply confirmed — rather than driven by the Ethereum contract state. That's the right axis: sweeping while the vault is still live would permanently void the depositor's pre-signed recourse graph (`claim_tx` hard-codes `(peginTxid, 1)` as its input and downstream Assert/Payout/NoPayout signatures commit to that txid), and the on-chain settlement gate cannot get confused about that in the way an indexer or a contract read could. The tooltip and the deposit-time reserve copy in `copy.ts` now match what the code actually does, closing the loop on the false promise the PR opens with.

Ledger is blocked explicitly in the review screen with a tooltip that names the firmware limitation. All new user-facing strings live in `copy.ts`; the reclaim-specific errors have their own `ReclaimAlreadySettledError` so the UI can distinguish concurrent-sweep from other failures.

The RBF sequence `0xfffffffd` and `nVersion=2` (plain v2, not TRUC — the PegIn is deeply confirmed by the time this runs) are documented at their constants. `readVaults.length !== vaultIds.length` is caught before signing rather than at broadcast.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/reclaim.ts
```diff
@@ -0,0 +1,312 @@
...
+    if (contractScriptPubKey !== expectedScriptPubKey) {
+      throw new Error(
+        `Input ${i}: PegIn ${peginTxid} vout ${PEGIN_DEPOSITOR_CLAIM_VOUT} pays ` +
+          `${contractScriptPubKey}, but the connected wallet's depositor-claim ` +
+          `script is ${expectedScriptPubKey}. Reclaim refused — this reserve ` +
+          `belongs to a different wallet.`,
+      );
+    }
+
+    // Bind 2: the independent chain observation must agree with both.
```
jrwbabylonlab, line 228: All three binds are depositor-wide or params-wide, never vault-specific: the script is a pure function of the depositor key (identical across every vault they own), the value a function of protocol params, and `observed` carries no outpoint. So this asserts "some UTXO has this script and value", not "the outpoint added at :271 is the one observed". The only identity check is `vaults.length !== vaultIds.length` in buildAndBroadcastReclaim, whose message claims more than it does — `ReclaimVaultData` has no vault id.

If the ETH read returns vault B's `depositorSignedPeginTx` for requested vault A (same depositor/params), every bind passes and B's reserve is swept; if B is live its claim_tx is permanently unfundable. Single-vault today makes the RPC the only guard, but the batching follow-up makes an ordering/subset mismatch directly exploitable.

Suggest adding `txid`/`vout` to `observed` and asserting them, plus `vaultId` on `ReclaimVaultData` with `vaultIds[i] === keccak256(abi.encode(peginTxHash, depositor))` per input (derivation documented at btc-vault-registry/query.ts:69).

services/vault/src/services/vault/vaultReclaimService.ts
```diff
@@ -0,0 +1,305 @@
...
+export async function buildAndBroadcastReclaimTransaction(
+  params: BroadcastReclaimParams,
+): Promise<string> {
+  const { vaultId, depositorBtcPubkey, feeRate, signPsbt, signal } = params;
+
+  // Re-probe immediately before signing: the modal may have been open a while,
+  // and sweeping an already-spent reserve is a guaranteed broadcast failure
+  // after a pointless wallet prompt.
+  const chainState = await readReclaimChainState(vaultId);
+  if (chainState.reserveSpend.spent) {
```
jrwbabylonlab, line 254: `readReclaimChainState` returns `payoutSpend` and `tipHeight` and both are discarded — only `reserveSpend.spent` is read. So the Payout gate is enforced only in the render path (`getReclaimEligibility` has exactly one caller, useReclaimRowAction.ts:107), off a poller with 60s interval + 55s staleTime.

Payout confirms 6 deep, row enables, user opens the modal, a reorg drops the Payout and restores PegIn:0. The re-probe here reads `payoutSpend.spent === false` and ignores it; the sweep signs and the recovery set is destroyed while the vault is live. After such a reorg the VP graph is dead too (its Payout needs Assert:0), so the depositor graph was the remaining path.

The SDK header states this is the caller's job as a safety control. Fix is ~5 lines with no new I/O: feed this `chainState` into `getReclaimEligibility` and refuse anything not `available`.

services/vault/src/models/reclaimEligibility.ts
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * A spend confirmed but missing its height is treated as one confirmation —
+ * enough to be real, not enough to clear the deep-confirmation bar — so a
+ * partial esplora response delays the offer instead of waving it through.
+ */
+function payoutConfirmations(
+  payoutSpend: OutpointSpend,
+  tipHeight: number,
+): number {
+  if (payoutSpend.blockHeight === undefined) return 1;
```
jrwbabylonlab, line 204: `null === undefined` is false, so a `null` height skips this guard and `tipHeight - null + 1` yields ~tip+1 confirmations — the opposite of what the doc above says. `blockHeight` is copied verbatim from esplora JSON with no validation (useReclaimStatus.ts:75, vaultReclaimService.ts:110). At tip 900000: `null` → 900001 confs (passes), `"899000"` → 1001 (passes), `-1` → 900002 (passes); only `undefined` → 1.

Note `getTipHeight` validates its input (mempoolApi.ts:237, `/^\d+$/`) while `getOutspend` validates nothing. Suggest `if (!Number.isInteger(blockHeight) || blockHeight < 0) return 1;`

services/vault/src/hooks/useReclaimStatus.ts
```diff
@@ -0,0 +1,166 @@
...
+            ]);
+            return [
+              o.depositId,
+              {
+                payoutSpend: toOutpointSpend(payout),
+                reserveSpend: toOutpointSpend(reserve),
+                reserveValueSats: BigInt(reserveUtxo.value),
+              },
+            ];
+          } catch {
```
jrwbabylonlab, line 141: The fail-closed claim holds only for the no-prior branch. With a prior it is fail-open, because `tipHeight` is fetched independently and succeeds on the same tick (:121) while the carried-forward `blockHeight` is frozen — so `tipHeight - blockHeight + 1` grows purely from tip movement on data no longer being observed.

Payout mined at H, observed at 1 conf, no button. Block is reorged out, spend returns to mempool. Per-vault requests start 429ing (plausible — MAX_CONCURRENT_VAULTS is 2 precisely because of this rate limit) while the single getTipHeight keeps succeeding. Six blocks later the frozen status reads 6 confs and the button enables on a Payout that no longer exists.

Suggest stamping each status with the tip height it was observed at and computing confirmations against that, or dropping the entry. Worth noting useReclaimVaultChainData deliberately refuses to swallow errors for the mirror-image reason.

packages/babylon-ts-sdk/src/tbv/core/services/reclaim/buildAndBroadcastReclaim.ts
```diff
@@ -0,0 +1,257 @@
...
+    );
+    assertScriptPathSchnorrSignature({
+      requestedPsbtHex: psbtHex,
+      signatureHex: signature,
+      signerXOnlyPubkeyHex: xOnlyDepositorPubkey,
+      inputIndex,
+    });
+  });
+
+  const signedTxHex = finalizeAndExtract(signedPsbtHex);
```
jrwbabylonlab, line 253: Verification above correctly recomputes the sighash from the locally built PSBT, but finalize runs over the wallet-returned one, and `assertPsbtUnsignedTxMatches` deliberately skips per-input metadata — so `tapLeafScript`, `controlBlock` and `tapKeySig` are wallet-controlled here. Per bitcoinjs-lib 6.1.7 psbt.js:344, a returned `tapKeySig` makes it finalize a key-path witness and ignore `tapScriptSig` entirely; the internal key is NUMS, so the broadcast is rejected.

No fund loss (outputs and fee are pinned) — a wasted prompt and a confusing failure. Cheap to close: after `finalizeAndExtract`, assert each input's witness is exactly `[signature, descriptor.leafScript, descriptor.controlBlock]`; the verified signature is already in hand at :247.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/reclaim.ts
jrwbabylonlab: Tracked in #2324 — flagging the batching interaction there, since the length check cannot catch an ordering or subset mismatch once `readVaults` returns more than one reserve.

services/vault/src/services/vault/vaultReclaimService.ts
jrwbabylonlab: Tracked in #2324, together with the `blockHeight: null` and stale-carry-forward issues that feed the same gate.

PR #2321, opened 2026-08-25
fix(ledger-vault): stop expecting a signature on payout input 1

jrwbabylonlab:
The payout presign failed with "missing 1: 1:<leafhash>", after the device had
already signed input 0 and burnt its one-shot slot.

The expectation table arms on tap-leaf presence alone, and #2281 added a leaf to
input 1 so the device can display the terms. Nothing signs it: firmware,
btc-vault and the VP all want input 0 only.

Threads the caller's existing signInputs through; structural gates still run on
every input. Verified on a real Flex.

### DIFF
--- a/packages/babylon-ledger-vault-signer/src/__tests__/expectedSignatures.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/expectedSignatures.test.ts
@@ -98,6 +98,8 @@ function fixtureLeafScript(vector: SignPsbtVector, inputIndex: number): Buffer {
 }
 
 describe("expected-signature table classification (G2, 22 fixtures)", () => {
+  // Pin-dependent: at fixtures commit 8f99b8b these PSBTs' input 1 carries no
+  // leaf, so no requested set is needed. A post-#2281 Payout must pass one.
   it.each(TAPSCRIPT_VECTOR_NAMES.map((name) => [name]))("%s: input 0 tapscript, other inputs absent", (name) => {
     const vector = loadVector(name);
     const { table } = prepareSignPsbt({ psbtHex: vector.psbt_hex, depositorXOnlyHex: TEST_DEPOSITOR_KEY_HEX });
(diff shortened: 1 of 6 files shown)


jeremy-babylonlabs (approved):

jonybur (approved):
The expectation table's tapscript branch now gates on the caller's requested set while every structural check (control-block leaf commitment, ownership scan, requested-but-missing-metadata check) still runs on every input. Key-path signing is deliberately untouched — the base app picks the set under a wallet policy, so narrowing would under-expect. Provider forwards `signInputs` indices unchanged. The added tests pin the payout shape (input 0 requested → one expected yield, input 1's leaf still validates but is not expected), the host-bug case (requested index 2 with no metadata), and the key-path inertness against the deposit-flow fixture.

packages/babylon-ledger-vault-signer/src/__tests__/expectedSignatures.test.ts
```diff
@@ -98,6 +98,8 @@ function fixtureLeafScript(vector: SignPsbtVector, inputIndex: number): Buffer {
 }
 
 describe("expected-signature table classification (G2, 22 fixtures)", () => {
+  // Pin-dependent: at fixtures commit 8f99b8b these PSBTs' input 1 carries no
```
gbarkhatov, line 101: **1 — The fixture corpus no longer describes what the SDK emits, and the replacement coverage cannot fail when the builder changes again.**

All 14 `TAPSCRIPT_VECTOR_NAMES` fixtures are pinned at `8f99b8b`, which predates #2281 — the test immediately below this comment asserts `[...table.byInput.keys()]).toEqual([0])` with no requested set, so none of them carries a leaf on input 1.

`payoutShapedPsbtHex` is a hand-built PSBT encoding our *belief* about what `buildPayoutPsbt` emits. A builder change like #2281 will not break it, which is precisely how this bug shipped.

This package cannot close the gap itself: its `package.json` has no `@babylonlabs-io/ts-sdk` dependency, so the two halves of the contract never meet in a test here. `services/vault` depends on both. One contract test there — real `buildPayoutPsbt` output fed to `prepareSignPsbt` with the options `createTaprootScriptPathSignOptions(pk, 1)` actually produces — would have caught #2281 before a device did. Worth landing here or filing before merge.

Separately: this pin-dependence belongs in the vendor README's "Vector provenance" list (`src/vendor/ledger-bitcoin/README.md:199-214`), so the next fixture-pin bump does not silently invalidate the comment.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -818,9 +818,12 @@ export class LedgerVaultProvider implements IBTCProvider {
         wallet: WALLET_PROVIDER_NAME,
       });
     }
+    // Carrying a leaf is not the same as being signed: since #2281 Payout input 1
+    // carries the Assert payout leaf only so the device can display the terms.
+    const signInputIndexes = options?.signInputs?.map((input) => input.index);
     let prepared: PreparedSignPsbt;
     try {
-      prepared = prepareSignPsbt({ psbtHex, depositorXOnlyHex });
+      prepared = prepareSignPsbt({ psbtHex, depositorXOnlyHex, signInputIndexes });
```
gbarkhatov, line 826: **2 — The intent-replay fingerprint is now caller-controlled.**

`signingRequestKey` (`provider.ts:108-121`) derives its key from `prepared.table.byInput`, which this call now narrows. The same Payout keys as `<txid>|0:leafA` when the caller passes `signInputs: [{index: 0}]`, and as `<txid>|0:leafA|1:leafB` when it passes nothing — so `stagedKeys` (:874) and `signedFingerprints` (:881) miss a resubmission that differs only by options.

The comment at :868-871 calls this host guard "the only defense" for NoPayout, which has no device dedup mask.

Bounded today — Payout has the device mask, and NoPayout's table is unchanged by narrowing since its inputs 1-2 carry no leaf. But request identity should be a property of the PSBT, not of what the caller asked for. Keying on the unsigned txid plus the *unnarrowed* classification would restore that.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -818,9 +818,12 @@ export class LedgerVaultProvider implements IBTCProvider {
         wallet: WALLET_PROVIDER_NAME,
       });
     }
+    // Carrying a leaf is not the same as being signed: since #2281 Payout input 1
+    // carries the Assert payout leaf only so the device can display the terms.
+    const signInputIndexes = options?.signInputs?.map((input) => input.index);
```
gbarkhatov, line 823: **3 — This makes a caller-supplied *count* authoritative over a device behaviour.**

"The firmware signs input 0 alone" is a property of the device — I confirmed it in `sign_custom_inputs.c`, where both the NoPayout and the Payout branch pass a literal `0` to `sign_sighash_schnorr_and_yield`, with no loop over inputs. But the value that now drives the host's expectation comes from `createTaprootScriptPathSignOptions(publicKey, inputCount)`, which takes a *count* and emits indices `0..inputCount-1`.

Every in-tree tapscript caller passes `1` today (`signDepositorGraph.ts:240,266`, `PayoutManager.ts:268,372`, `PeginManager.ts:1090`, `buildAndBroadcastRefund.ts:546`), so this is latent. But a caller passing `2` re-arms input 1, the device still yields input 0 only, and `assertComplete` throws the same `missing 1:<leafhash>` this PR fixes — after the one-shot payout slot is spent. Asserting that the tapscript requested set is a subset of `{0}` would pin the device fact where it belongs.

Minor, same line: `publicKey` and `useTweakedSigner` are dropped. Both are genuinely inert here — the table pins on the device-read key, and the device picks tweaking by spend type — but half-honouring the struct on a critical path deserves a line saying why.

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
```diff
@@ -271,11 +286,15 @@ export function buildExpectedSignatureTable(params: BuildExpectedSignatureTableP
...
       // BIP-371 keyData of a TAP_LEAF_SCRIPT entry IS the control block.
       assertControlBlockCommitsToLeaf(inputIndex, leafEntries[0].keyData, leafHash, leafWitnessUtxo.scriptPubKey);
-      byInput.set(inputIndex, {
-        kind: "tapscript",
-        expectedLeafHashHexes: new Set([leafHash.toString("hex")]),
-        expectedSignerXOnlyHex: depositorXOnlyHex,
-      });
+      // Gated AFTER every structural gate above: a leaf we never sign is still
+      // a leaf the spent output must have committed to.
+      if (isRequested(inputIndex)) {
```
gbarkhatov, line 291: **4 — Narrowing can blind the mixed-input gate in the provider.**

`provider.ts:830` builds `kinds` from the already-narrowed `table.byInput`, and this gate only ever removes *tapscript* entries. A PSBT with a key-path input 0 and a tapscript input 1, signed with `signInputs: [{index: 0}]`, therefore yields `kinds = {taproot-keypath}` of size 1 — so the "mixed inputs are not a vault flow" throw at :832 never fires.

The flow then enters the policy branch, re-prepares at :864 *without* `signInputIndexes` so input 1's expectation comes back, and dies in `assertComplete` after the user has approved on-device. Before this PR that was a zero-I/O `INVALID_PARAMS`.

No in-tree builder emits a mixed PSBT, so this is a defense-in-depth regression rather than a live bug. Computing `kinds` from an unnarrowed classification — and narrowing only what is handed to the interpreter — keeps the gate honest.

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
```diff
@@ -126,6 +126,13 @@ export interface BuildExpectedSignatureTableParams {
   readonly psbt: ExpectedSignaturePsbt;
   /** Connected depositor x-only key (64 lowercase hex) — pins the table. */
   readonly depositorXOnlyHex: string;
+  /**
+   * Input indices the caller asked to sign. Narrows TAPSCRIPT expectations
+   * only, and is inert for key-path/policy signing — there the base app picks
+   * the set. Structural gates still run on EVERY input. Omit to expect every
```
gbarkhatov, line 132: **5 — "Structural gates still run on EVERY input" is no longer accurate.**

The gates inside this function do run on every input. But two of the prepare-level gates are table-driven and now skip non-requested inputs: `assertDefaultSighashTypes` (`signPsbtPrepare.ts:135`) and `assertNoConflictingSignatures` (`signPsbtPrepare.ts:153`) both iterate `table.byInput`.

Their own JSDoc scopes them to "an input this round signs", so the new behaviour is arguably more correct — it is only this sentence that overstates. Worth narrowing the wording to the gates in this file.

Same paragraph, and repeated at `signPsbtPrepare.ts:194-198`: "inert under `walletPolicy`" also overstates slightly. Key-path *expectations* are never narrowed, but the requested-but-missing loop at :356 still applies to key-path indices.

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
```diff
@@ -330,6 +351,16 @@ export function buildExpectedSignatureTable(params: BuildExpectedSignatureTableP
     }
   }
 
+  // A requested index with no expectation means the caller and the builder
+  // disagree about the PSBT — a host bug, caught at zero device I/O.
+  for (const inputIndex of requestedInputs ?? []) {
```
gbarkhatov, line 356: **6 — An out-of-range index gets a message describing a different failure.**

`requestedInputs` is never checked against `inputCount`, so `signInputIndexes: [0, 5]` on a 2-input PSBT reports `input 5 was requested for signing but carries no signing metadata`. Input 5 does not exist — the caller is out of range, not the builder short of metadata.

`inputCount` is in scope on line 242; a bounds check ahead of this loop would say so.

packages/babylon-ledger-vault-signer/src/expectedSignatures.ts
```diff
@@ -330,6 +351,16 @@ export function buildExpectedSignatureTable(params: BuildExpectedSignatureTableP
...
+    if (!byInput.has(inputIndex)) {
+      throw new LedgerSignPsbtProtocolError(
+        `input ${inputIndex} was requested for signing but carries no signing metadata`,
+      );
+    }
+  }
+
   if (byInput.size === 0) {
     // A signPsbt call with nothing to sign is a host bug — thrown before ANY APDU.
     throw new LedgerSignPsbtProtocolError("PSBT contains no depositor-signable input");
```
gbarkhatov, line 366: **7 — An empty requested set produces a false diagnosis, and behaves asymmetrically.**

`options?.signInputs?.map(...)` at `provider.ts:823` maps `[]` to `[]`, not `undefined`, so `requestedInputs` becomes an empty Set: every tapscript expectation is narrowed away, the loop above iterates nothing, and this throw fires — blaming the PSBT ("contains no depositor-signable input") for what is a caller bug. On a key-path PSBT the same empty set is silently ignored, since key-path expectations are never narrowed.

Latent: nothing in tree produces `[]`, because `createTaprootScriptPathSignOptions` rejects `inputCount < 1`. But this is the only BTC provider that does not guard the empty case — `unisat/provider.ts:325`, `okx/provider.ts:184` and `onekey/provider.ts:235` all check `signInputs.length > 0`. Rejecting an empty set explicitly, with a caller-directed message, closes both halves.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -857,6 +860,7 @@ export class LedgerVaultProvider implements IBTCProvider {
       try {
         // Pass the AUGMENTED hex: the signer's merge target is whatever hex it
         // prepared, so the SDK gets the derivation fields back with the tapKeySig.
+        // No signInputIndexes: it narrows tapscript only, and this path is key-path.
```
gbarkhatov, line 863: **8 — This invariant is guaranteed by comment only, and it is the one protecting the key-path path from under-expecting.**

The harness already has everything needed: the key-path tests at `__tests__/provider.test.ts:563` and `:601` read `const [prepareArgs] = signMock.prepareSignPsbt.mock.calls.at(-1)!` and assert on it. One line — `expect(prepareArgs.signInputIndexes).toBeUndefined()` — closes it against a future edit that threads the narrowing through both prepares.

packages/babylon-ledger-vault-signer/src/__tests__/expectedSignatures.test.ts
```diff
@@ -446,6 +448,161 @@ describe("tapscript control block must commit to its leaf, driven by PegIn fixtu
...
+        prepareSignPsbt({
+          psbtHex: payoutShapedPsbtHex(true),
+          depositorXOnlyHex: TEST_DEPOSITOR_KEY_HEX,
+          signInputIndexes: [0],
+        }),
+      /input 1 control block does not commit to its TAP_LEAF_SCRIPT/,
+    );
+  });
+
+  it("completes the ceremony when the device yields input 0 alone", () => {
```
gbarkhatov, line 535: **9 — The security-relevant half of the narrowing is not covered.**

This test asserts the *permissive* direction: with input 1 narrowed out, a yield for input 0 alone now completes. The fail-closed direction is what protects you if the firmware reading is ever wrong — a YIELD for input 1 must hit `unexpected-input` at `expectedSignatures.ts:436-438`, because narrowing removed its entry from the table.

`signPsbtLoop.test.ts:572` (T9) covers that for the *pre*-#2281 shape, where input 1 is `witnessUtxo`-only and absent for a different reason. Nothing covers it for the shape this PR introduces, where input 1 carries a valid leaf and is absent *only* because the caller did not request it.

A few lines in this describe — feed the collector a yield for input 1, expect the `unexpected-input` mismatch — would pin it.

packages/babylon-ledger-vault-signer/src/__tests__/expectedSignatures.test.ts
```diff
@@ -446,6 +448,161 @@ describe("tapscript control block must commit to its leaf, driven by PegIn fixtu
...
+    if (expectation?.kind !== "tapscript") throw new Error("input 0 must classify tapscript");
+    // varint(input 0) ‖ augmLen(0x40) ‖ signer key(32) ‖ leaf hash(32) ‖ sig(64).
+    collector.assertAndRecord(
+      Buffer.from("0040" + TEST_DEPOSITOR_KEY_HEX + [...expectation.expectedLeafHashHexes][0] + "ab".repeat(64), "hex"),
+    );
+
+    expect(() => collector.assertComplete()).not.toThrow();
+  });
+
+  it("rejects a requested input that carries no signing metadata", () => {
```
gbarkhatov, line 554: **10 — The name describes a case this does not exercise.**

`payoutShapedPsbtHex()` builds two inputs, so index `2` is out of range: the loop at `expectedSignatures.ts:242` never visits it, and it trips the requested-but-missing throw for that reason rather than for carrying no signing metadata. (See finding 6 on that error message.)

The case the name promises — an input that exists, has a `witnessUtxo`, and carries no taproot metadata — is real and reachable with a committed fixture: `deposit-flow__claimer_payout__*` input 1 is exactly that shape, so `signInputIndexes: [0, 1]` against one of those vectors covers it against real bytes. Worth having both.

packages/babylon-ledger-vault-signer/src/__tests__/expectedSignatures.test.ts
```diff
@@ -446,6 +448,161 @@ describe("tapscript control block must commit to its leaf, driven by PegIn fixtu
...
+  /** Single-leaf taptree, so the control block genuinely commits to its witnessUtxo. */
+  function leafInput(script: Buffer): { scriptPubKey: Buffer; controlBlock: Buffer } {
+    const leaf = { output: script, version: 0xc0 };
+    const p2tr = payments.p2tr({ internalPubkey: Buffer.from(OTHER_KEY_HEX, "hex"), scriptTree: leaf, redeem: leaf });
+    if (!p2tr.output || !p2tr.witness) throw new Error("p2tr produced no script-path spend");
+    return { scriptPubKey: p2tr.output, controlBlock: p2tr.witness[p2tr.witness.length - 1] };
+  }
+
+  /** Two tapscript inputs, exactly the shape `buildPayoutPsbt` emits since #2281. */
+  function payoutShapedPsbtHex(mutateInput1ControlBlock = false): string {
```
gbarkhatov, line 467: **11 — Nits in this block, taste only.**

- `payoutShapedPsbtHex(mutateInput1ControlBlock = false)` is the boolean-flag parameterised helper CLAUDE.md's test philosophy discourages; two small builders read better than one flag.
- The first `forEach` on :469 binds `script` and never uses it — `addInput` only needs the index.
- `PAYOUT_OUTPUT_SCRIPT` (:454) is byte-identical to `P2TR_RANDOM_SCRIPT` (:133) and is likewise just a random P2TR, so the name claims more than it is.
- The comment on :466, "exactly the shape `buildPayoutPsbt` emits", overstates: the real builder also sets `tapInternalKey` on both inputs and per-input CSV sequences, and its input-1 control block folds a multi-leaf Assert taptree rather than the zero-hop one here. Not a coverage gap — classification is leaf-first, and multi-hop folding is covered by the real PegIn fixtures — but worth softening.
- Assertion granularity is inconsistent: :493 and :503 split one behaviour across two identical setups, while :513 asserts both together, which is the file's own convention at :107-108.

gbarkhatov (approved):
Reviewed at `ec7ea908c`. The fix is correct and lands on the right seam — approving.

I verified the core premise against firmware source rather than taking the comment on faith: in `app-babylon-vault`'s `sign_custom_inputs.c`, both the NoPayout branch and the Payout branch pass a literal `0` as the input index to `sign_sighash_schnorr_and_yield`, with no loop over inputs. So "the device signs input 0 alone" holds for both flows, and narrowing the *expectation* to match is the right fix.

Also verified:

- `isRequested` is gated after every structural gate — control-block commitment, leaf version, `witnessUtxo`, ambiguous-leaf set — so a leaf we never sign is still a leaf the spent output must have committed to.
- Key-path expectations are never narrowed, which is required: under a wallet policy the base app signs every internal input.
- Every in-tree tapscript caller requests exactly `[0]` and extracts at index 0, so no flow under-expects today.
- Both touched suites pass locally: signer 575 passed / 10 skipped, provider 106 passed.

Nothing below blocks the merge. The one I would most want addressed before or shortly after is finding 1: the fixture corpus no longer describes what the SDK emits, and the replacement coverage is structurally incapable of catching the next builder change.

Findings inline, numbered 1-11.

PR #2322, opened 2026-08-25
fix(vault): apply the updated product copy from the v3 design review

jeremy-babylonlabs:
Applies the product-copy updates from the v3 design review. Every change was verified against the current Figma frame, not just the comment. Strings live in `copy.ts`. Two changes are not copy-only and are called out under **Beyond copy** below.

## Applied

| Surface | Change |
|---|---|
| Terminology | "BTC Vault" → "BTCVault" on the flagged labels (activation modal, split column headers, liquidation event rows, reorder modal); removed "sacrificial"/"protected" wording everywhere (badges deleted from `LiquidationEventCard`, banner copy reworded, reorder info text) |
| Overview | Health-factor tooltip now names the 1.0 liquidation threshold, "If it falls below 1.0…" (shared key, propagates to Liquidations and Borrow); removed the Total Collateral Value tooltips (`OverviewSection`, `VaultsSummaryCard`, Liquidations `PositionOverview`); activity empty state → "…after your first transaction"; section titles unchanged (see **Second review pass**) |
| Liquidations | new debt-repayment fairness tooltip; seizure-order subheading updated to the frame text; chart bands are single-line ("contain vaults" sublabel removed) |
| Borrow / markets | "Available Liquidity" label casing (loan cards, asset-selection column, markets table column); removed the markets-table Available Liquidity header tooltip; interest-rate model description updated |
| Repay | Removed the Debt and Health Factor tooltips per the frame |
| Deposit flow | Fee/terms rows relabeled: "LTV / Collateral Factor" → "Collateral Factor", "Liquidation threshold (THF)" → "Target Health Factor" (its tooltip always described THF), "Liquidation Bonus (LB)" → "Max Liquidation Penalty"; split-minimum tooltip reworded; collateral-factor Hint added to the Max-to-Borrow row; signing-progress explainer extended with the borrowing note; do-not-spend warning shortened |
| Position banners | Urgent detail → "BTC price drop to $X triggers liquidation."; cliff body → "With your current BTCVaults, a single liquidation event liquidates all your BTC in collateral."; suggestion → "Adding a new BTCVault of X BTC enables partial liquidation."; cliff CTA → "Add X BTC"; dust title → "Small position - simplified view" |

## Needs a product call (applied as noted, flag before merge)

- **Transaction-reserve tooltip:** the design frame still carries pre-reclaim wording ("returned to you if unused"), but #2317 deliberately changed this string to describe the shipped reclaim flow. Kept the #2317 wording; the frame needs updating instead.
- **Cliff CTA label deviates from the frame.** The frame gives both the urgent and the cliff CTA the label "Add Collateral", but they do different things (the cliff one pre-fills an amount) and `NotificationAction` has no accessible-name field separate from `label` — so the two were indistinguishable to keyboard and screen-reader users, and the test had to select by DOM position. The cliff CTA now reads "Add 0.72 BTC", which also restores the link to the suggestion text above it. Same precedent as the transaction-reserve tooltip: the frame needs updating.
- **Reorder guidance:** `REORDER_INFO_TEXT` told users to place their smallest BTC Vault first. That is not what the calculator does, and neither is "largest first": a vault only isolates the first liquidation event if it alone covers the target seizure, so the best order depends on the vault sizes (which is what `computeOptimalOrder`'s DP is for). The text now explains the front-to-back seizure mechanism and prescribes no ordering rule. This string does not come from the design review, so product should confirm the new wording.

## Not applied (no code counterpart or out of copy scope)

- Simulator explainer tooltips (13251:41849): no tooltip slot exists on the shipped chart header; adding one is a UI change.
- Landing-page copy (10206:56371): the landing page is not in this repo.
- Wallet-connect items (12450:7033 unsupported-address error state, 2195:59090 wrong-wallet notice): unbuilt UI.
- Design/UX comments (chart size/colors, info button, Aave in the wallet list, support-link placement) — tracked separately.
- Nodes deleted from Figma since the comment (8082:92162, 10051:26752, 10233:51372).

## Beyond copy

Two items in this PR are not string-only and deserve a design/QA look:

- **Fairness tooltip on the debt-repaid branch.** `liquidationChartData` now sets a tooltip on both fairness branches, so a new info icon renders in the "Fairness Debt Repaid" row of every non-full-liquidation event card. The field is now required on the type rather than optional.
- **Chart band sublabel removed.** The "(contain vault 1, vault 2)" line is gone from every band, which re-centres each band's text block. The vault names were also part of the focusable rect's accessible name, so a new optional `accessibleDetail` field on `LiquidationBand` keeps them there for screen readers while the visual line stays removed. Whether `sublabel` should remain public core-ui API now that the vault app passes none is an open question.

## Verification

- `tsc --noEmit` clean on `services/vault` and `packages/babylon-core-ui`; eslint 0 errors; prettier clean on every touched file under each package's own config.
- `services/vault` vitest: 3604/3610 passing (6 pre-existing skips). `packages/babylon-core-ui` vitest: 113/113. All copy-asserting tests updated.
- Rebased on `main`; merges clean.
- Unrelated flake seen under parallel load in `lazyWithRetry.test.ts` and `VaultWalletConnectionProvider.disabledWallets.test.tsx` — both pass in isolation, neither is in this diff, and both mutate global state. Worth a separate look.

## Review pass

Applied after review:

- Reorder-modal strings moved out of `ReorderVaults/constants.ts` (now deleted) into `COPY.reorder`, per the centralized-copy rule.
- `CollateralFactorRow`'s new Hint uses the `attachToChildren` + `InfoIcon` pattern, so it no longer nests a `div` inside the row's `span`.
- Terminology: the three stray `BTCvault`/`BTCvaults` spellings normalised to `BTCVault`/`BTCVaults`, matching the glossary landed in #2309.
- Grammar: "a safe place", "Aave's interest rate model", "splitting a deposit", and the liquidation-events subheading split into separate sentences.
- Dead code: the `badge` field on `LiquidationEventCard` lost its only reader when the sacrificial/protected pill was removed — field, computation, stale doc comments and the tests that asserted it are gone.
- `collateralLiquidated` reverted from `"BTCVault"` to `"Collateral liquidated"`: the row labels a BTC amount, so the bare product noun misdescribed the value.

## Second review pass

Reverts, where the new copy was demonstrably wrong against the code:

- **`"% to Liquidation"` → `"Distance"`.** `distancePct = ((liqPrice - btcPrice) / btcPrice) * 100`, so an event that has *not* fired renders `-18.1%`. Under the new heading that reads backwards. The band popover also still called the same value "Distance".
- **`"Liquidation Fee"` → `"Max Liquidation Penalty"`** (key renamed too). `DynamicReserveConfig` has a distinct `liquidationFee` field this app never reads; the row renders `(maxLiquidationBonus - 1) * 100`, which is a ceiling that scales with health factor, not a fee. The old tooltip promised a "range" where a single scalar renders.
- **`"Expired Deposits"` → `"Inactive Vaults"`.** The section is fed `[...expiredActivities, ...reclaimableCandidates]`, and the latter are settled `DEPOSITOR_WITHDRAWN` rows carrying a Reclaim action.
- **`"Total Collateral"` → `"Active Vaults"`.** The heading takes a `(count)` suffix of `vaults.length`, and the same page already shows a "Total Collateral Value" stat.
- **Dust title → "Small position - simplified view".** The gate is `totalDebtUsd < $1,000 || collateralValue < $1,000` and only forces a single display group; it is not a protocol rule, partial liquidation is not disabled, and the unchanged body directly below already said "shown as one liquidation event".
- **Button label → "Close & continue later".** The declarative sentence rendered as a primary button's accessible name.
- **Artifact body keeps the privacy claim:** "Your files are stored locally and never uploaded. Keep them somewhere safe."
- **Signing-progress explainer keeps the not-yet-locked reassurance** alongside the borrowing note.
- **Repay health-factor tooltip restored** — it was removed on Repay but kept on Borrow and four other surfaces, and it is the only place in the app stating the 1.0 threshold.

Fixes:

- Cliff suggestion regained its BTC unit and terminal period: "Adding a new BTCVault of 0.72 BTC enables partial liquidation."
- Reorder `infoText` corrected — an event seizes until it covers *that event's target seizure amount* (a fraction of remaining collateral), not "the amount owed".
- `CollateralFactorRow`'s hinted row now carries `inline-flex items-center gap-1` on the parent, matching the `DepositForm` precedent; without it the icon butted flush against `(CF=72%)` and sat off-baseline.
- Its test mock now surfaces the tooltip text and asserts it, instead of stubbing `Hint` to `null` and covering the new feature with nothing.
- Dead structure removed: the `justify-between` wrapper the badge left behind, and the two `flex gap-1` label wrappers in `RepayDetailsCard`.
- `LiquidationEventCard`'s "no fairness tooltip" test replaced — it constructed a shape production can no longer emit.
- Same-surface terminology made consistent (cliff siblings, reorder success strings, deposit-flow readiness messages, dust body). Repo-wide convergence on `BTCVault` is deliberately left to the glossary bot that produced #2309; ~143 occurrences remain, including `utils/errors/errorMessages.ts`, which is outside this diff.
- Sentence-case `loans.availableLiquidityLabel`; Title Case kept on the stats bar and table column headers, which the style header exempts.
- "over seizure" hyphenated to "over-seizure" throughout.
- The core-ui story is now formatted at the package's own `printWidth: 120`. The earlier "prettier clean" claim was checked with the wrong config.

### DIFF
--- a/packages/babylon-core-ui/src/components/Notification/Notification.stories.tsx
+++ b/packages/babylon-core-ui/src/components/Notification/Notification.stories.tsx
@@ -109,13 +109,13 @@ export const WarningNoIconWithLabel: Story = {
     icon: null,
     title: "First liquidation takes everything",
     children:
-      "With your current vaults, a single liquidation event seizes all your BTC — nothing remains protected behind it.",
+      "With your current BTCvaults, a single liquidation event liquidates all your BTC in collateral.",
     suggestion: (
       <>
         <SuggestionLabel>Suggestion</SuggestionLabel>
         <Text variant="body2" as="div" className="text-accent-secondary">
           To enable partial liquidation, withdraw your 0.70 BTC and re-deposit as
-          two smaller vaults: 0.40 BTC sacrificial + 0.30 BTC protected.
+          two smaller vaults: 0.40 BTC + 0.30 BTC.
         </Text>
       </>
     ),
--- a/packages/babylon-core-ui/src/components/Notification/Notification.stories.tsx
+++ b/packages/babylon-core-ui/src/components/Notification/Notification.stories.tsx
@@ -109,13 +109,13 @@ export const WarningNoIconWithLabel: Story = {
     icon: null,
     title: "First liquidation takes everything",
     children:
-      "With your current BTCvaults, a single liquidation event liquidates all your BTC in collateral.",
+      "With your current BTCVaults, a single liquidation event liquidates all your BTC in collateral.",
     suggestion: (
       <>
         <SuggestionLabel>Suggestion</SuggestionLabel>
         <Text variant="body2" as="div" className="text-accent-secondary">
-          To enable partial liquidation, withdraw your 0.70 BTC and re-deposit as
-          two smaller vaults: 0.40 BTC + 0.30 BTC.
+          To enable partial liquidation, withdraw your 0.70 BTC and re-deposit
+          as two smaller vaults: 0.40 BTC + 0.30 BTC.
         </Text>
       </>
     ),
--- a/packages/babylon-core-ui/src/components/Notification/Notification.stories.tsx
+++ b/packages/babylon-core-ui/src/components/Notification/Notification.stories.tsx
@@ -108,14 +108,13 @@ export const WarningNoIconWithLabel: Story = {
     variant: "warning",
     icon: null,
     title: "First liquidation takes everything",
-    children:
-      "With your current BTCVaults, a single liquidation event liquidates all your BTC in collateral.",
+    children: "With your current BTCVaults, a single liquidation event liquidates all your BTC in collateral.",
     suggestion: (
       <>
         <SuggestionLabel>Suggestion</SuggestionLabel>
         <Text variant="body2" as="div" className="text-accent-secondary">
-          To enable partial liquidation, withdraw your 0.70 BTC and re-deposit
-          as two smaller vaults: 0.40 BTC + 0.30 BTC.
+          To enable partial liquidation, withdraw your 0.70 BTC and re-deposit as two smaller BTCVaults: 0.40 BTC + 0.30
+          BTC.
         </Text>
       </>
     ),
--- a/services/vault/src/applications/aave/components/LoanCard/Repay/RepayDetailsCard/RepayDetailsCard.tsx
+++ b/services/vault/src/applications/aave/components/LoanCard/Repay/RepayDetailsCard/RepayDetailsCard.tsx
@@ -1,4 +1,4 @@
-import { Hint, SubSection } from "@babylonlabs-io/core-ui";
+import { SubSection } from "@babylonlabs-io/core-ui";
 
 import {
   getHealthFactorColor,
@@ -48,7 +48,6 @@ export function RepayDetailsCard({
       <div className={ROW_CLASS}>
         <div className="flex items-center gap-1 text-accent-secondary">
           {COPY.loans.debtLabel}
-          <Hint tooltip={COPY.loans.debtTooltip} />
         </div>
         <span className="text-accent-primary">
           {debtOriginal ? (
@@ -68,7 +67,6 @@ export function RepayDetailsCard({
       <div className={ROW_CLASS}>
         <div className="flex items-center gap-1 text-accent-secondary">
           {COPY.loans.healthFactorLabel}
-          <Hint tooltip={COPY.tooltips.healthFactor} />
         </div>
         <span className="flex items-center gap-2 text-accent-primary">
           {healthFactorOriginal && originalColor ? (
--- a/services/vault/src/applications/aave/components/LoanCard/Repay/RepayDetailsCard/RepayDetailsCard.tsx
+++ b/services/vault/src/applications/aave/components/LoanCard/Repay/RepayDetailsCard/RepayDetailsCard.tsx
@@ -1,4 +1,4 @@
-import { SubSection } from "@babylonlabs-io/core-ui";
+import { Hint, SubSection } from "@babylonlabs-io/core-ui";
 
 import {
   getHealthFactorColor,
@@ -46,9 +46,7 @@ export function RepayDetailsCard({
   return (
     <SubSection className="w-full flex-col gap-4 !bg-secondary-highlight">
       <div className={ROW_CLASS}>
-        <div className="flex items-center gap-1 text-accent-secondary">
-          {COPY.loans.debtLabel}
-        </div>
+        <span className="text-accent-secondary">{COPY.loans.debtLabel}</span>
         <span className="text-accent-primary">
           {debtOriginal ? (
             <span className="flex items-center gap-2">
@@ -67,6 +65,7 @@ export function RepayDetailsCard({
       <div className={ROW_CLASS}>
         <div className="flex items-center gap-1 text-accent-secondary">
           {COPY.loans.healthFactorLabel}
+          <Hint tooltip={COPY.tooltips.healthFactor} />
         </div>
         <span className="flex items-center gap-2 text-accent-primary">
           {healthFactorOriginal && originalColor ? (
--- a/services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
+++ b/services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
@@ -10,8 +10,8 @@ import type { LiquidationEventCard as EventCardData } from "./liquidationChartDa
  * Per-event breakdown card (Figma node 10251:64809). Purely presentational —
  * every value arrives pre-formatted from `buildLiquidationChartData`.
  *
- * The top rule and the badge both carry the event's identity: the rule matches
- * the band's colour lane in the chart, the badge its role in the cascade.
+ * The top rule carries the event's identity, matching the band's colour lane
+ * in the chart.
  */
 
 const TONE_BORDER: Record<EventCardData["tone"], string> = {
@@ -117,12 +117,6 @@ function SeizureRow({
 }
 
 export function LiquidationEventCard({ card }: { card: EventCardData }) {
-  const sacrificial = card.badge === "sacrificial";
-  // Colour tracks whether the event has FIRED, not its position in the
-  // cascade. Keying off `sacrificial` left every protected event green even
-  // once the price had fallen through its trigger — with `formatSignedPct`
-  // unsigned at the time, the minus sign was the only cue left that anything
-  // had happened.
   const { triggered } = card;
 
   return (
@@ -136,16 +130,6 @@ export function LiquidationEventCard({ card }: { card: EventCardData }) {
         <h3 className="text-xl leading-[1.6] tracking-[0.15px] text-accent-primary">
           {card.title}
         </h3>
-        <span
-          className={twJoin(
-            "rounded-full px-2.5 py-1 text-xs leading-[1.4] tracking-[0.4px] text-primary-main",
-            sacrificial ? "bg-risk-amber" : "bg-risk-green",
-          )}
-        >
-          {sacrificial
-            ? COPY.liquidations.events.badgeSacrificial
-            : COPY.liquidations.events.badgeProtected}
-        </span>
       </div>
 
       <div className="flex gap-16">
--- a/services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
+++ b/services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
@@ -126,11 +126,9 @@ export function LiquidationEventCard({ card }: { card: EventCardData }) {
         TONE_BORDER[card.tone],
       )}
     >
-      <div className="flex items-center justify-between">
-        <h3 className="text-xl leading-[1.6] tracking-[0.15px] text-accent-primary">
-          {card.title}
-        </h3>
-      </div>
+      <h3 className="text-xl leading-[1.6] tracking-[0.15px] text-accent-primary">
+        {card.title}
+      </h3>
 
       <div className="flex gap-16">
         <Stat
--- a/services/vault/src/components/pages/Liquidations/liquidationChartData.ts
+++ b/services/vault/src/components/pages/Liquidations/liquidationChartData.ts
@@ -376,6 +376,7 @@ function toCard(
     : {
         label: COPY.liquidations.events.fairnessDebtRepaid,
         value: formatUsd(group.fairnessDebtRepay),
+        tooltip: COPY.liquidations.events.fairnessDebtRepaidTooltip,
       };
 
   return {
@@ -438,9 +439,6 @@ export function buildLiquidationChartData(
     return {
       key: String(i),
       label: COPY.liquidations.eventTitle(i + 1),
-      sublabel: COPY.liquidations.containVaults(
-        group.vaults.map((v) => v.name.toLowerCase()).join(", "),
-      ),
       amountLabel: formatBtcAmount(group.combinedBtc),
       priceTop: group.liquidationPrice,
       priceBottom,
--- a/services/vault/src/components/pages/Liquidations/liquidationChartData.ts
+++ b/services/vault/src/components/pages/Liquidations/liquidationChartData.ts
@@ -62,7 +62,7 @@ export interface LiquidationEventCard {
   debtRepaidLabel: string;
   liquidatorProfitLabel: string;
   /** Full group shows a wBTC fairness payment; safe groups show fairness debt repaid. */
-  fairness: { label: string; value: string; tooltip?: string };
+  fairness: { label: string; value: string; tooltip: string };
   btcRemainingLabel: string;
   debtRemainingLabel: string;
   hfAfterLabel: string;
@@ -426,6 +426,9 @@ export function buildLiquidationChartData(
     return {
       key: String(i),
       label: COPY.liquidations.eventTitle(i + 1),
+      accessibleDetail: COPY.liquidations.containVaults(
+        group.vaults.map((v) => v.name.toLowerCase()).join(", "),
+      ),
       amountLabel: formatBtcAmount(group.combinedBtc),
       priceTop: group.liquidationPrice,
       priceBottom,
--- a/services/vault/src/components/pages/Liquidations/liquidationChartData.ts
+++ b/services/vault/src/components/pages/Liquidations/liquidationChartData.ts
@@ -31,11 +31,6 @@ import {
  *    — the event's own trigger price, not the simulated price, since the
  *    fairness payment is a property of that liquidation, not of wherever the
  *    simulator currently sits.
- *  - Sacrificial vs Protected badge: no backing field exists. `calculate.ts`
- *    emits groups in seizure order but with a 1-based `index` field, so
- *    "first group" is the group's POSITION in the `groups` array (position 0),
- *    never `group.index` itself. The first-position group is the sacrificial
- *    one; the rest are protected.
  */
 
 const TONES: LiquidationBandTone[] = ["1", "2", "3"];
@@ -50,21 +45,15 @@ export interface AmountParts {
 export interface LiquidationEventCard {
   key: string;
   title: string;
-  badge: "sacrificial" | "protected";
   /** Matches the band's colour lane — drives the card's top rule. */
   tone: LiquidationBandTone;
-  /**
-   * The price has fallen to or through this event's trigger. Distinct from
-   * `badge`, which is the event's fixed role in the cascade (position 0 is
-   * sacrificial) and does not move with the simulator.
-   */
+  /** The price has fallen to or through this event's trigger. */
   triggered: boolean;
   collateralLabel: string;
   liqPriceLabel: string;
   distanceLabel: string;
   /** Sign of the distance, so a consumer can tint it without re-parsing
-   *  `distanceLabel`. The overview cards tint negative distances; the
-   *  dashboard-page cards tint by `badge` instead. */
+   *  `distanceLabel`. The overview cards tint negative distances. */
   distanceNegative: boolean;
   seizedVaults: ({ name: string } & AmountParts)[];
   targetSeizure: AmountParts;
@@ -356,8 +345,8 @@ function hfAfter(group: LiquidationGroup, cf: number): string {
 }
 
 // `position` is the group's array index: `calculate()` emits 1-based
-// `group.index`, so array position is the only safe identity for titles,
-// badges, and keys. Aftermath figures are priced at the event's own trigger —
+// `group.index`, so array position is the only safe identity for titles
+// and keys. Aftermath figures are priced at the event's own trigger —
 // the price at which the liquidation actually executes — never the ambient
 // (possibly simulated) price.
 function toCard(
@@ -366,7 +355,6 @@ function toCard(
   cf: number,
   btcPrice: number,
 ): LiquidationEventCard {
-  const sacrificial = position === 0;
   const fairness = group.isFullLiquidation
     ? {
         label: COPY.liquidations.events.fairnessPaymentWbtc,
@@ -382,7 +370,6 @@ function toCard(
   return {
     key: String(position),
     title: COPY.liquidations.eventTitle(position + 1),
-    badge: sacrificial ? "sacrificial" : "protected",
     // Same rule as a band's `state`, so a card and its band can never
     // disagree about whether the event has fired.
     triggered: btcPrice <= group.liquidationPrice,
--- a/services/vault/src/components/simple/CollateralFactorRow.tsx
+++ b/services/vault/src/components/simple/CollateralFactorRow.tsx
@@ -32,18 +32,15 @@ export function CollateralFactorRow({
   return (
     <div className="flex items-center justify-between text-sm">
       <span className="text-accent-primary">{FORM_COPY.maxToBorrowLabel}</span>
-      <span>
+      <span className="inline-flex items-center gap-1">
         <span className="text-accent-primary">
           {maxBorrowUsd !== null
             ? `${formatCompactUsd(maxBorrowUsd)} USD`
             : "--"}
         </span>
         <span className="text-accent-secondary">
-          {" "}
           {FORM_COPY.cfParenthetical(percent)}
         </span>
-        {/* attach-to-children keeps the markup inline-valid inside this
-            span (a bare Hint would nest a div inside it). */}
         <Hint tooltip={COPY.tooltips.collateralFactor} attachToChildren>
           <InfoIcon size={16} className="text-accent-secondary" />
         </Hint>
--- a/services/vault/src/components/simple/CollateralFactorRow.tsx
+++ b/services/vault/src/components/simple/CollateralFactorRow.tsx
@@ -1,4 +1,4 @@
-import { Hint } from "@babylonlabs-io/core-ui";
+import { Hint, InfoIcon } from "@babylonlabs-io/core-ui";
 
 import { COPY } from "@/copy";
 import { computeMaxBorrowUsd } from "@/utils/collateral";
@@ -42,7 +42,11 @@ export function CollateralFactorRow({
           {" "}
           {FORM_COPY.cfParenthetical(percent)}
         </span>
-        <Hint tooltip={COPY.tooltips.collateralFactor} />
+        {/* attach-to-children keeps the markup inline-valid inside this
+            span (a bare Hint would nest a div inside it). */}
+        <Hint tooltip={COPY.tooltips.collateralFactor} attachToChildren>
+          <InfoIcon size={16} className="text-accent-secondary" />
+        </Hint>
       </span>
     </div>
   );
--- a/services/vault/src/components/simple/CollateralFactorRow.tsx
+++ b/services/vault/src/components/simple/CollateralFactorRow.tsx
@@ -1,3 +1,5 @@
+import { Hint } from "@babylonlabs-io/core-ui";
+
 import { COPY } from "@/copy";
 import { computeMaxBorrowUsd } from "@/utils/collateral";
 import { formatCompactUsd } from "@/utils/formatting";
@@ -40,6 +42,7 @@ export function CollateralFactorRow({
           {" "}
           {FORM_COPY.cfParenthetical(percent)}
         </span>
+        <Hint tooltip={COPY.tooltips.collateralFactor} />
       </span>
     </div>
   );
--- a/services/vault/src/components/simple/PositionNotificationBanner/BannerActions.tsx
+++ b/services/vault/src/components/simple/PositionNotificationBanner/BannerActions.tsx
@@ -32,9 +32,9 @@ interface BuildBannerActionsArgs {
  * `Notification` `actions` slot:
  * - urgent: "Add Collateral" (primary, filled) + "Repay Debt" (secondary,
  *   outlined) — the core safety actions, matching the Figma callout.
- * - cliff with an affordable sacrificial size: "Add sacrificial vault" (generic
- *   label per Figma; the amount lives in the suggestion text) — opens the
- *   deposit flow with that amount pre-filled.
+ * - cliff with an affordable actionable size: "Add Collateral" (same label as
+ *   the urgent CTA per Figma; the amount lives in the suggestion text) —
+ *   opens the deposit flow with that amount pre-filled.
  * - optimal reorder available: "Apply Optimal Order" — filled (primary) on the
  *   standalone reorder card, secondary when it accompanies the urgent callout.
  *
@@ -79,17 +79,18 @@ export function buildBannerActions({
     );
   }
 
-  // Add-sacrificial-vault CTA on a cliff when the calculator produced an
-  // actionable size — "add a sacrificial vault of this exact size". When an
-  // urgent warning is primary it rides along as a secondary action so the
-  // safety actions lead. The calculator guarantees the amount is positive and
-  // no larger than the position. Generic "Add sacrificial vault" label (the
-  // amount lives in the suggestion text per Figma).
+  // Add-vault CTA on a cliff when the calculator produced an actionable size
+  // — "add a vault of this exact size". When an urgent warning is primary it
+  // rides along as a secondary action so the safety actions lead. The
+  // calculator guarantees the amount is positive and no larger than the
+  // position. Shares the "Add Collateral" label with the urgent CTA per
+  // Figma (the amount lives in the suggestion text); the two can render
+  // together as duplicate-labeled buttons when both warnings are active.
   const hasCliffWarning = result.warnings.some((w) => w.type === "cliff");
   if (hasCliffWarning && result.suggestedNewVaultBtc !== null) {
     const amountBtc = result.suggestedNewVaultBtc.toFixed(2);
     actions.push({
-      label: COPY.banner.addSacrificialVault,
+      label: COPY.banner.addCollateral,
       onClick: () => onDeposit(amountBtc),
       emphasis: isUrgent ? "secondary" : "primary",
       // Adding a vault is a deposit — blocked under a protocol Freeze/Pause.
--- a/services/vault/src/components/simple/PositionNotificationBanner/BannerActions.tsx
+++ b/services/vault/src/components/simple/PositionNotificationBanner/BannerActions.tsx
@@ -32,9 +32,9 @@ interface BuildBannerActionsArgs {
  * `Notification` `actions` slot:
  * - urgent: "Add Collateral" (primary, filled) + "Repay Debt" (secondary,
  *   outlined) — the core safety actions, matching the Figma callout.
- * - cliff with an affordable actionable size: "Add Collateral" (same label as
- *   the urgent CTA per Figma; the amount lives in the suggestion text) —
- *   opens the deposit flow with that amount pre-filled.
+ * - cliff with an affordable actionable size: "Add <amount> BTC" — opens the
+ *   deposit flow with that amount pre-filled. The amount sits in the label so
+ *   the button stays distinguishable from the urgent "Add Collateral" CTA.
  * - optimal reorder available: "Apply Optimal Order" — filled (primary) on the
  *   standalone reorder card, secondary when it accompanies the urgent callout.
  *
@@ -83,14 +83,13 @@ export function buildBannerActions({
   // — "add a vault of this exact size". When an urgent warning is primary it
   // rides along as a secondary action so the safety actions lead. The
   // calculator guarantees the amount is positive and no larger than the
-  // position. Shares the "Add Collateral" label with the urgent CTA per
-  // Figma (the amount lives in the suggestion text); the two can render
-  // together as duplicate-labeled buttons when both warnings are active.
+  // position. The amount sits in the label so this button never collides
+  // with the urgent "Add Collateral" CTA when both warnings are active.
   const hasCliffWarning = result.warnings.some((w) => w.type === "cliff");
   if (hasCliffWarning && result.suggestedNewVaultBtc !== null) {
     const amountBtc = result.suggestedNewVaultBtc.toFixed(2);
     actions.push({
-      label: COPY.banner.addCollateral,
+      label: COPY.banner.addVaultOfSize(amountBtc),
       onClick: () => onDeposit(amountBtc),
       emphasis: isUrgent ? "secondary" : "primary",
       // Adding a vault is a deposit — blocked under a protocol Freeze/Pause.
--- a/services/vault/src/components/simple/PositionNotificationBanner/__tests__/PositionNotificationBanner.test.tsx
+++ b/services/vault/src/components/simple/PositionNotificationBanner/__tests__/PositionNotificationBanner.test.tsx
@@ -566,7 +566,7 @@ describe("PositionNotificationBanner", () => {
     expect(screen.queryByTestId("position-notification-banner")).toBeNull();
   });
 
-  it("keeps the cliff Add Collateral CTA (secondary) when a single-vault cliff is also urgent", () => {
+  it("keeps the cliff add-vault CTA (secondary) when a single-vault cliff is also urgent", () => {
     const result = makeBaseResult({
       warnings: [
         {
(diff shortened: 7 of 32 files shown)


services/vault/src/copy.ts
```diff
@@ -2011,24 +2007,24 @@ export const COPY = {
...
     },
     liquidationThreshold: {
-      label: "Liquidation threshold (THF)",
+      label: "Target Health Factor",
       tooltip: "The ideal health factor to restore during liquidation",
     },
     liquidationBonus: {
-      label: "Liquidation Bonus (LB)",
-      tooltip: "Bonus percentage awarded to liquidators on seized collateral.",
+      label: "Liquidation Fee",
```
gbarkhatov: **`"Liquidation Bonus (LB)"` -> `"Liquidation Fee"` names a different on-chain value.**

`DynamicReserveConfig` has a distinct field literally called `liquidationFee` (`AaveSpoke.abi.json:66`, typed at `spoke.ts:375`) which this app never reads. `useProtocolFeeRows.ts:74` renders `(LB - 1) * 100` from `maxLiquidationBonus`. So the row will print "Liquidation Fee: 5%" next to a number that is not the liquidation fee.

Second problem: it's the *ceiling*, not a fixed value — `getLiquidationConfig` returns `healthFactorForMaxBonus` and `liquidationBonusFactor`, i.e. the bonus scales with how far HF has fallen; `maxLiquidationBonus` is the upper bound.

And the tooltip on the next line ("The penalty **range** applied during the liquidation") promises a range where a single scalar renders — plus "the liquidation" is a definite article with no antecedent on a protocol-parameters panel.

Suggest `"Max Liquidation Penalty"` with something like *"The maximum penalty applied to seized collateral during liquidation; the actual penalty scales with how far the health factor has fallen."* Please don't reuse the string "Liquidation Fee" for this value — that name is taken by a real field we may want to surface later.

(Related nit: the key is still `protocolFees.liquidationBonus`, and `protocolFees.liquidationThreshold` no longer names its label either — worth renaming both so the keys don't re-seed the old vocabulary.)

services/vault/src/copy.ts
```diff
@@ -1752,12 +1745,10 @@ export const COPY = {
...
       heading: "Liquidation Events",
       subheading:
-        "Vaults are seized in order — each vault group is one liquidation event.",
-      badgeSacrificial: "Sacrificial",
-      badgeProtected: "Protected",
+        "BTCVaults are seized in order. Each BTCVault group is one liquidation event. To change the order, open the Vaults page.",
       collateral: "Collateral",
       liqPrice: "Liq Price",
-      distance: "Distance",
+      distance: "% to Liquidation",
```
gbarkhatov: **`"Distance"` -> `"% to Liquidation"` inverts how the sign reads.**

`distancePct = ((pLiq - btcPrice) / btcPrice) * 100` (`calculate.ts:173`), so an event that has **not** fired renders `-18.1%` and one that already fired renders `+18.1%`. `formatSignedPct`'s own comment documents that convention as *distance*, which is exactly why the old neutral label worked.

Under the new heading a positive number reads as "18.1% of room left" — backwards — and `-18.1%` reads as nonsense. Either revert to `"Distance"`, or negate the rendered value for this label so a not-yet-triggered event shows `18.1%`.

Also note `copy.ts:1739` still labels the identical value `"Distance"` in the band popover, so the same metric now has two names on one page.

services/vault/src/copy.ts
```diff
@@ -2097,22 +2093,22 @@ export const COPY = {
...
-      body: "With your current BTC Vaults, a single liquidation event seizes all your BTC — nothing remains protected behind it.",
+      body: "With your current BTCVaults, a single liquidation event liquidates all your BTC in collateral.",
       // Header shown above the suggestion text when there is no actionable CTA
       // (the withdraw/re-deposit and multi-vault cases). Rendered uppercase.
       suggestionLabel: "Suggestion",
       // Variant A (#1948): an affordable sacrificial vault buffers the existing
       // position. The amount lives here; the CTA label stays generic.
       addSacrificialSuggestion: (sacrificialBtc: string) =>
-        `Adding a sacrificial ${sacrificialBtc} BTC Vault creates a buffer — it gets liquidated first, your existing BTC survives.`,
+        `Adding a ${sacrificialBtc} BTCVault enables partial liquidation`,
```
gbarkhatov: **The BTC unit got lost in this rename.**

`btc2()` returns a bare decimal (`calculate.ts:45`, called at `:441`), so this renders:

> Adding a **0.72 BTCVault** enables partial liquidation

In the old string ("Adding a sacrificial 0.72 BTC Vault") the "BTC" did double duty as the unit. Collapsing the product name to one word deletes it, and now `0.72` quantifies the vault rather than the bitcoin.

Also a sentence fragment with no terminal period, where every sibling in `cliff` (`body`, `withdrawResplitSuggestion`, `noSplitSuggestion`, `twoVault.suggestion`, `multiVault.suggestion`) is a full stopped sentence and renders in the same slot.

Suggest: `` `Adding a new BTCVault of ${sacrificialBtc} BTC enables partial liquidation.` ``

The computation behind it is correct — `existingBtc * liqFactor / (1 - liqFactor)`, ceil to cents, capped at position size. This is purely the rendering.

services/vault/src/copy.ts
```diff
@@ -395,7 +394,7 @@ export const COPY = {
       summary: {
         estimate: (duration: string) => `~${duration}`,
         description:
-          "Each step is divided into several wallet signature confirmations. The progress counter shows how many are completed. Your Bitcoin will only be locked once the BTC Vault is activated.",
+          "Each step is divided into several wallet signature confirmations. The progress counter shows how many are completed. You will be able to borrow assets once BTCVault is activated.",
```
gbarkhatov: **This drops the one line telling the user their BTC isn't committed yet.**

The old text ended *"Your Bitcoin will only be locked once the BTC Vault is activated."* — the only on-screen note, during a flow that asks for N wallet signatures, that funds are not yet locked. The replacement swaps that reassurance for a borrowing benefit, which answers a question the user isn't asking at this point in the flow.

If the frame mandates the borrowing sentence, consider keeping both rather than substituting.

Separately: *"once BTCVault is activated"* is missing its article. `BTCVault` is a common noun here, not a brand used attributively — compare `copy.ts:597`, which gets it right with *"Your BTCVault is now active"*. Suggest "once your BTCVault is activated".

services/vault/src/copy.ts
```diff
@@ -2153,7 +2149,7 @@ export const COPY = {
         `This position already has the maximum number of BTC Vaults (${cap}).`,
     },
     dust: {
-      title: "Position too small for BTC Vault analysis",
+      title: "Position too small to enable partial liquidation",
```
gbarkhatov: **The new title asserts a protocol constraint that doesn't exist, and contradicts its own body.**

The gate is `totalDebtUsd < $1,000 || collateralValue < $1,000` (`calculate.ts:96`) — it forces a single *display* group and skips the cascade analysis. It is not a protocol rule, and partial liquidation is not disabled (that's the separate `cliff` warning).

The unchanged `detail` directly below still says *"the cascade simplifies — all BTC Vaults are shown as one liquidation event"*, so the title now says "cannot" while the body says "we display it that way".

Because the condition is `||`, a position with $200k collateral and $500 debt — maximally safe — gets told it is "too small to enable partial liquidation".

Suggest a display-framed title, e.g. `"Small position - simplified view"` or `"Position too small for a detailed cascade"`.

services/vault/src/copy.ts
```diff
@@ -1996,13 +1987,18 @@ export const COPY = {
...
   },
   geoBlock: {
     title: "Service unavailable in your region",
     body: "We're unable to provide access from your current region due to regulatory restrictions.",
   },
   reorder: {
+    modalTitle: "Reorder Vaults",
+    modalSubtitle:
+      "This is the liquidation order of your BTCVaults. Drag to change the order.",
+    infoText:
```
gbarkhatov, line 2006: **Right about the direction, wrong about the stopping condition.**

Agreed the old text was unsupported — "place your smallest first" is not what the calculator does, and the `> MAX_DP_N` fallback actually sorts *largest*-first (`optimalOrder.ts:74`), so neither rule holds in general. Good catch, and front-to-back seizure is correct (`calculate.ts:178-186` walks from index 0 and takes `slice(0, i)`).

But *"until it covers the amount owed"* isn't right. The loop exits on `prefixSum >= curTargetSeizure * (1 - SEIZURE_TOL)` where `curTargetSeizure = remainingBtc * seizedFraction` (`calculate.ts:174`) — a fraction of *remaining collateral* derived from CF/LB/THF, not the debt. A single event deliberately does not repay the whole debt; that's the entire reason there's a cascade of events.

Suggest: "...seizes BTCVaults from the front until it covers that event's target seizure amount..."

**Also undocumented in the PR body:** the move into `copy.ts` changed the modal title and subtitle too, not just `infoText`. `REORDER_MODAL_TITLE` was `"Reorder BTC Vaults"` and is now `"Reorder Vaults"`; the subtitle was rewritten as well. Worth calling out explicitly since the description only mentions the `infoText` rewrite.

services/vault/src/copy.ts
```diff
@@ -1904,8 +1895,8 @@ export const COPY = {
       // Heading over the vaults list before any vault is active — the count
       // is dropped while the section only holds the empty state.
       vaultsTitle: "Vaults",
-      activeVaultsTitle: "Active Vaults",
-      inactiveVaultsTitle: "Inactive Vaults",
+      activeVaultsTitle: "Total Collateral",
+      inactiveVaultsTitle: "Expired Deposits",
```
gbarkhatov: **"Expired Deposits" is false for half the rows in this section.**

The section is fed `[...expiredActivities, ...reclaimableCandidates]` (`VaultsLifecycleSections.tsx:569-572`), and `reclaimableCandidates` are `ContractStatus.DEPOSITOR_WITHDRAWN` — settled, successfully withdrawn deposits with an unspent depositor-claim reserve. The comment right above that line says so explicitly: *"Settled vaults join the expired ones in the Inactive section rather than getting a section of their own."*

So a user who completed a withdrawal sees it filed under "Expired Deposits", and those rows carry a **Reclaim** action.

You flagged this as needing a product call — confirming from the code that it's a real mislabel, not just a wording preference. "Inactive Vaults" covered both states correctly; `"Closed Deposits"` would too if a new word is required.

services/vault/src/copy.ts
```diff
@@ -1904,8 +1895,8 @@ export const COPY = {
       // Heading over the vaults list before any vault is active — the count
       // is dropped while the section only holds the empty state.
       vaultsTitle: "Vaults",
-      activeVaultsTitle: "Active Vaults",
-      inactiveVaultsTitle: "Inactive Vaults",
+      activeVaultsTitle: "Total Collateral",
```
gbarkhatov: **A value noun over a count, and it collides with two things on the same page.**

This renders as `Total Collateral (3)` — the parenthetical is `vaults.length` (`VaultsActiveSection.tsx:180-183`), a count of vaults. "Total Collateral" names an amount, so suffixing a cardinal produces a heading that reads as "3 collateral".

On the same page, `VaultsSummaryCard.tsx:54` already renders a stat labeled **"Total Collateral Value"** (a BTC amount) and `:67` renders **"Active Vaults"** (the same count this heading now shows). After this change the page shows "Total Collateral Value", "Active Vaults", and "Total Collateral (3)" simultaneously — the two "collateral"-named things mean different concepts and the two vault-count labels have diverged.

Either revert, or drop the `count(...)` suffix at the call site and rename `vaults.summary.activeVaultsLabel` in the same sweep so one vocabulary wins.

services/vault/src/copy.ts
```diff
@@ -2097,22 +2093,22 @@ export const COPY = {
     // what action is feasible.
     cliff: {
       title: "First liquidation takes everything",
-      body: "With your current BTC Vaults, a single liquidation event seizes all your BTC — nothing remains protected behind it.",
+      body: "With your current BTCVaults, a single liquidation event liquidates all your BTC in collateral.",
```
gbarkhatov, line 2104: **The terminology sweep contradicts itself inside this one object.**

`body` here says `BTCVaults`, while its own siblings a few lines down still say the old form: `withdrawResplitSuggestion` (2111) says bare *"two smaller vaults"*, `noSplitSuggestion` (2113), `twoVault.suggestion` (2122) and `multiVault.suggestion` (2130-2132) all say `BTC Vault(s)`. These render in the same banner.

Same pattern elsewhere:
- **Reorder flow:** `"Reorder Vaults"` (1996) -> `"your BTCVaults"` (1998) -> `"BTC Vault order updated"` (2004) across modal and success toast.
- **Deposit flow:** `"BTCVault 1"` column headers (408) vs `"Vault 1: ... this BTC Vault cannot continue"` (798/800) for the same vault in the same flow.
- **Dust:** the title was rewritten (2152) but `detail` directly below still says "all BTC Vaults".

Verified counts: `copy.ts` now holds **19 `BTCVault`** against **79 `BTC Vault`**; repo-wide under `services/vault/src` + `packages/babylon-core-ui/src` there are **143** `"BTC Vault"` occurrences left, including 22 lines in `utils/errors/errorMessages.ts` — which CLAUDE.md designates as part of the copy surface. `BTCVaultNotFound: "BTC Vault not found..."` there has the ABI key already re-termed and the message not.

I'd either finish the sweep or defer the rename wholesale. The current middle state is worse for the reader than either endpoint, and it's what caused the unit loss at 2103.

services/vault/src/copy.ts
```diff
@@ -516,7 +514,7 @@ export const COPY = {
         { text: "Before activating, ", emphasis: false },
         { text: "download the recovery artifacts", emphasis: true },
         {
-          text: " of your BTC Vault. These files will make sure your BTC Vault is fully functional even if your vault provider becomes unavailable.",
+          text: " of your vault. These files will make sure your BTCVault is fully functional even if your vault provider becomes unavailable.",
```
gbarkhatov: **Bare "vault" and "BTCVault" naming the same object in adjacent sentences.**

This reads *"...download the recovery artifacts of your **vault**. These files will make sure your **BTCVault** is fully functional..."*. The first occurrence was `"your BTC Vault"` before this change and got downgraded to bare `vault`, which the style header at the top of this file explicitly forbids, while the second was upgraded to the new one-word form.

Suggest `" of your BTCVault. These files will make sure your BTCVault is fully functional..."`.

services/vault/src/copy.ts
```diff
@@ -93,8 +93,7 @@ const SIGN_TRANSACTION_LABEL = "Sign Transaction";
 // standalone download dialog and the activate-confirmation dialog so the
 // two can't drift.
 const ARTIFACTS_DOWNLOADED_TITLE = "Artifacts downloaded";
-const ARTIFACTS_DOWNLOADED_BODY =
-  "Your files are stored locally and never uploaded.";
+const ARTIFACTS_DOWNLOADED_BODY = "Keep your artifact files in a safe place!";
```
gbarkhatov: **This trades a factual reassurance for an instruction, and it's the only `!` in the file.**

The old body was *"Your files are stored locally and never uploaded."* — which answers the question a depositor actually has at that moment: did my recovery material just leave my machine? The new string only issues an instruction and drops the privacy claim entirely.

`grep -n '!"' copy.ts` returns exactly one hit — this line — so the exclamation mark also breaks tone with every other string in the file.

Suggest keeping both: `"Your files are stored locally and never uploaded. Keep them somewhere safe."`

services/vault/src/copy.ts
```diff
@@ -405,11 +404,10 @@ export const COPY = {
...
       defaultSuccessMessage: PRE_PEGIN_BROADCAST_CONFIRMATION_MESSAGE,
       doNotSpendWarning:
-        "To ensure a seamless deposit, do not spend the BTC allocated for this process until the transaction is confirmed.",
-      splitVaultColumnLabel: (vaultNumber: number) =>
-        `BTC Vault ${vaultNumber}`,
+        "Do not spend the BTC used for this deposit until the transactions are confirmed.",
+      splitVaultColumnLabel: (vaultNumber: number) => `BTCVault ${vaultNumber}`,
       buttons: {
-        closeContinueLater: "Close & continue later",
+        closeContinueLater: "You can close and come back later",
```
gbarkhatov: **A declarative sentence in a button slot.**

This renders as the primary fluid `<Button>` label at `DepositProgressView.tsx:512`, in the same slot as its siblings `"Sign Transaction"`, `"Retry"`, `"Close"`, `"Done"`, `"Cancel signing"` — and `DepositProgressView.test.tsx:510` confirms it becomes the button's accessible name: `getByRole("button", { name: "You can close and come back later" })`.

A button whose label is a statement about what the user *may* do gives no action to perform, and a screen-reader user hears prose where an action verb belongs.

Suggest keeping an imperative label (`"Close"` / `"Close & continue later"`) and moving the sentence into the helper-text slot beside it. Worth confirming with design rather than taking the frame literally here.

services/vault/src/copy.ts
```diff
@@ -1432,7 +1430,7 @@ export const COPY = {
     // token units (before → after the repayment).
     debtLabel: "Debt",
     healthFactorLabel: "Health factor",
-    availableLiquidityLabel: "Available liquidity",
+    availableLiquidityLabel: "Available Liquidity",
```
gbarkhatov: **Title Case among all sentence-case siblings.**

Immediate siblings are `debtLabel: "Debt"`, `healthFactorLabel: "Health factor"`, `utilizationLabel: "Utilization"`, `availableLabel: "Available"`, `atRiskOfLiquidation: "At risk of liquidation"`. This renders on the same metric row as `borrowRateLabel` and `utilizationLabel` (`ActiveLoansList.tsx:65`), so the capital L is visibly the odd one out.

Worth noting `marketInfo.stats.availableLiquidity` (1606) *is* legitimately Title Case — its block carries an explicit `// Stats bar above the fold. Title Case per the Figma` comment. `loans.*` has no such exemption, and neither do the four protocol-parameter labels that switched to Title Case in this PR (`"Min Deposit"`, `"Effective Minimum for Split"`, `"Target Health Factor"`, `"Liquidation Fee"` at 2010/2014/2022/2026). Either add a header clause covering these, or use sentence case.

Separately: this PR deletes `columns.availableTooltip` and strips the `<Hint>` from the markets table header, while `marketInfo.stats.availableLiquidityTooltip` (1607) survives and still renders on the stat card on the same page. The table now has no explanation while the card does — worth pointing the header Hint at the surviving key instead of deleting it, per this file's stated "one key per concept" pattern.

services/vault/src/copy.ts
```diff
@@ -1770,9 +1761,11 @@ export const COPY = {
       debtRepaid: "Debt Repaid",
       liquidatorProfit: "Liquidator profit",
       fairnessDebtRepaid: "Fairness Debt Repaid",
+      // Additional debt the liquidator repays because indivisible BTC Vaults
+      // forced an over-seizure; pairs with fairnessPaymentTooltip below.
+      fairnessDebtRepaidTooltip:
+        "Additional repayment of debt by the liquidator due to over seizure of collateral.",
```
gbarkhatov: Nit: "over seizure" should be hyphenated as an attributive compound — **over-seizure** — here and in `fairnessPaymentTooltip` (1772) and `overSeizure: "Over seizure"` (1756).

The comment added two lines above this in the same change already spells it `over-seizure`, so the file disagrees with itself.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -376,12 +364,12 @@ function toCard(
     : {
         label: COPY.liquidations.events.fairnessDebtRepaid,
         value: formatUsd(group.fairnessDebtRepay),
+        tooltip: COPY.liquidations.events.fairnessDebtRepaidTooltip,
```
gbarkhatov, line 367: **This adds a tooltip, so a new InfoIcon now renders — it isn't a copy change.**

The debt-repaid fairness branch was `undefined` by design. `LiquidationEventCard.tsx:214` passes `card.fairness.tooltip` into `DetailRow`, whose `{tooltip ? <Hint tooltip={tooltip} /> : null}` means an InfoIcon now appears in the "Fairness Debt Repaid" row of **every non-full-liquidation event card**.

The test comment deleted alongside it ("the debt-repaid row must not claim it") shows this reverses a deliberate invariant. This is the largest non-copy item in a PR that states "no layout or behaviour changes" — worth naming in the description, and worth a design confirm, because the tooltip makes a new claim about protocol mechanics rather than rewording an existing one.

Knock-on: with both branches of `toCard` now setting it, `fairness: { ...; tooltip?: string }` at line 65 is unreachable in its optional form — make it required. And `LiquidationEventCard.test.tsx:60` ("shows no fairness tooltip when the card supplies none") now constructs a shape production can no longer emit, so it passes while verifying nothing in `src/`.

services/vault/src/components/pages/Liquidations/liquidationChartData.ts
```diff
@@ -438,9 +426,6 @@ export function buildLiquidationChartData(
     return {
       key: String(i),
       label: COPY.liquidations.eventTitle(i + 1),
-      sublabel: COPY.liquidations.containVaults(
-        group.vaults.map((v) => v.name.toLowerCase()).join(", "),
-      ),
       amountLabel: formatBtcAmount(group.combinedBtc),
```
gbarkhatov, line 432: **Removing `sublabel` from every band is a layout and accessibility change, not a string edit.**

Two knock-ons in core-ui:

1. `BandLayer.tsx:77-79` pushes a third text line when `band.sublabel` is present, and `:127-139` vertically centers the whole stack (`nextTop = rect.y + (rect.height - stackHeight) / 2`). Dropping a line re-centers **every band's text block** on both surfaces that render bands (`pages/Liquidations/index.tsx:167` and `simple/LiquidationAnalysisSection.tsx:99`).

2. `BandLayer.tsx:163-165` builds the focusable rect's accessible name as `[band.label, band.sublabel, band.amountLabel].filter(Boolean).join(" ")`. Screen-reader users lose the vault names from it: *"Liq Event 1 contain vault 1, vault 2 0.6 BTC"* becomes *"Liq Event 1 0.6 BTC"*. The names do survive in `popoverMetrics`, but the popover isn't wired to the rect via `aria-describedby`, so it isn't part of the accessible name.

If the sublabel is going, consider keeping the vault names in the `aria-label` explicitly. Either way this belongs in the description.

Note this also leaves the vault app as the only consumer passing no `sublabel`, while core-ui still renders one and its `DROP_SUBLABEL_MAX_PX` line-drop path is now exercised only by core-ui's own fixtures — worth a deliberate call on whether that prop stays public API.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -136,16 +130,6 @@ export function LiquidationEventCard({ card }: { card: EventCardData }) {
         <h3 className="text-xl leading-[1.6] tracking-[0.15px] text-accent-primary">
```
gbarkhatov: Dead structure left by the badge removal: the wrapping `<div className="flex items-center justify-between">` on the line above now contains only this `<h3>`. `justify-between` existed to push the badge to the right edge and is inert with one child.

Drop the wrapper and render the `<h3>` directly as the article's first child.

services/vault/src/components/simple/CollateralFactorRow.tsx
```diff
@@ -40,6 +42,11 @@ export function CollateralFactorRow({
           {" "}
           {FORM_COPY.cfParenthetical(percent)}
         </span>
+        {/* attach-to-children keeps the markup inline-valid inside this
+            span (a bare Hint would nest a div inside it). */}
+        <Hint tooltip={COPY.tooltips.collateralFactor} attachToChildren>
```
gbarkhatov, line 44: **This doesn't match the pattern the comment cites, and will misalign.**

The precedent is `DepositForm.tsx:321-331`, where the **parent** carries `inline-flex items-center gap-1` and the `Hint attachToChildren` + `InfoIcon` sits inside it. Here the parent is a bare `<span>` (line 35) with no flex and no gap:

- **No spacing.** `Hint`'s own `gap-1` applies between *its children*, and under `attachToChildren` it has exactly one child — so the gap is a no-op and the icon butts flush against the `)` of `(CF=72%)`. The two existing siblings on 41-44 use an explicit `{" "}`; this one got none.
- **Baseline.** The wrapper is an `inline-flex` box in a plain inline flow, so it baselines on its first flex item's text baseline and the 16px SVG sits visibly high against the row. Every other hinted label in the codebase (`PositionStatCards.tsx:64`, `HINTED_LABEL_CLASS` in `LiquidationEventCard.tsx:29`, `RepayDetailsCard`) puts `items-center` on a flex parent containing both label and icon.

Fix is one class on line 35: `<span className="inline-flex items-center gap-1">`.

Minor: the DOM-nesting justification in the comment is only half right — react-tooltip renders in place rather than via a portal, so the tooltip body is a `<div>` inside this `<span>` while open either way. What `attachToChildren` actually avoids is the *persistent* wrapper element. Worth correcting so the next reader doesn't over-trust it.

services/vault/src/components/simple/__tests__/CollateralFactorRow.test.tsx
```diff
@@ -1,5 +1,11 @@
 import { render, screen } from "@testing-library/react";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
+
+// Component tests mock core-ui (its dist isn't built in the test run).
+vi.mock("@babylonlabs-io/core-ui", () => ({
```
gbarkhatov, line 7: This mock stubs `Hint` and `InfoIcon` to `() => null`, which means the only production change this PR makes to `CollateralFactorRow` — the new collateral-factor tooltip — is asserted nowhere. `grep` for `tooltips.collateralFactor` finds production sites only.

The repo already has the pattern for this, in a file this same PR edits — `Liquidations/__tests__/LiquidationEventCard.test.tsx:7-10` uses a *surfacing* mock precisely so the tooltip can be asserted:

```ts
vi.mock("@babylonlabs-io/core-ui", () => ({
  Hint: ({ tooltip }: { tooltip?: string }) => <span>{tooltip}</span>,
}));
```

Suggest mirroring that and adding `expect(screen.getByText(COPY.tooltips.collateralFactor)).toBeInTheDocument()`.

Also the stated reason ("its dist isn't built in the test run") isn't accurate — `packages/babylon-core-ui/dist/index.es.js` exists and resolves; sibling tests import core-ui unmocked and pass. Worth fixing the comment or dropping it.

services/vault/src/applications/aave/components/LoanCard/Repay/RepayDetailsCard/RepayDetailsCard.tsx
```diff
@@ -48,7 +48,6 @@ export function RepayDetailsCard({
       <div className={ROW_CLASS}>
         <div className="flex items-center gap-1 text-accent-secondary">
```
gbarkhatov: Both `<div className="flex items-center gap-1 text-accent-secondary">` wrappers here (this one and line 68) now hold a single text node — those classes existed only to lay out label + `<Hint>`.

The equivalent cleanup *was* done in this PR at `BorrowMarketsTable.tsx:145`, where the `flex items-center gap-1` was stripped when the Hint went, so this is internally inconsistent.

Collapse to `<span className="text-accent-secondary">{COPY.loans.debtLabel}</span>` (and likewise for the health-factor label), or restore the Hint per the comment below.

services/vault/src/applications/aave/components/LoanCard/Repay/RepayDetailsCard/RepayDetailsCard.tsx
```diff
@@ -68,7 +67,6 @@ export function RepayDetailsCard({
...
```
gbarkhatov, line 66: **The health-factor Hint is removed here but kept on Borrow.**

`COPY.tooltips.healthFactor` was *not* deleted from `copy.ts` and still renders at `BorrowMetricsCard.tsx:118`, `LoansSummary.tsx:74`, `RiskSection.tsx:130`, `VaultsSummaryCard.tsx:76` and `PositionOverview.tsx:75`. So after this change the two halves of the same LoanCard disagree.

That reads as over-deletion rather than intent — contrast `loans.debtTooltip`, which was properly removed from `copy.ts` in the same commit.

It also matters more here than anywhere else: repay is the one screen where the user is acting specifically to move that number, and that tooltip is the only place in the app that states the 1.0 liquidation threshold. `copy.ts:109-110` says the `tooltips` block exists so *"the screens can't drift apart"* — this is exactly that drift.

Either restore it, or remove it from Borrow too. Also `copy.ts:115` still lists the consumers as "borrow / repay detail cards", which is now stale either way.

services/vault/src/components/simple/PositionNotificationBanner/BannerActions.tsx
```diff
@@ -79,17 +79,18 @@ export function buildBannerActions({
...
+  // calculator guarantees the amount is positive and no larger than the
+  // position. Shares the "Add Collateral" label with the urgent CTA per
+  // Figma (the amount lives in the suggestion text); the two can render
+  // together as duplicate-labeled buttons when both warnings are active.
   const hasCliffWarning = result.warnings.some((w) => w.type === "cliff");
   if (hasCliffWarning && result.suggestedNewVaultBtc !== null) {
     const amountBtc = result.suggestedNewVaultBtc.toFixed(2);
     actions.push({
-      label: COPY.banner.addSacrificialVault,
+      label: COPY.banner.addCollateral,
```
gbarkhatov: **Two buttons with the same accessible name doing different things.**

When a position is both urgent and cliff-warned with an actionable `suggestedNewVaultBtc`, `actions` holds two entries whose `label` is `COPY.banner.addCollateral`: line 65 calls `onDeposit()` and this one calls `onDeposit(result.suggestedNewVaultBtc.toFixed(2))` to pre-fill the sacrificial size. With `Repay Debt` between them, the duplicates aren't even adjacent.

Confirming there's no mitigation available today: `NotificationAction` has no `id`, `ariaLabel`, or `testId`, and `NotificationActionButton` renders only `type` / `onClick` / `disabled` / `className`. So keyboard and screen-reader users have no way to tell them apart, and there's no analytics or test hook either.

The whole point of the second CTA is the specific amount, so putting it back in the label is the cheapest fix and restores the link to the suggestion text above it — e.g. `` (btc) => `Add ${btc} BTC` ``. Alternatively add `ariaLabel?: string` to `NotificationAction` (small core-ui change) and keep the visible Figma label; that also gives the test a stable `getByRole("button", { name })` selector instead of the positional one below.

services/vault/src/components/simple/PositionNotificationBanner/__tests__/PositionNotificationBanner.test.tsx
```diff
@@ -589,10 +588,13 @@ describe("PositionNotificationBanner", () => {
 
     const banner = screen.getByTestId("position-notification-banner");
     expect(banner.dataset.severity).toBe("red");
-    // Safety actions still lead, and the pre-filled cliff CTA is still offered.
-    expect(screen.getByText("Add Collateral")).toBeTruthy();
+    // Safety actions still lead, and the pre-filled cliff CTA is still offered
+    // — it shares the "Add Collateral" label with the urgent CTA per Figma,
+    // so both buttons render with identical text.
+    const addCollateralButtons = screen.getAllByText("Add Collateral");
```
gbarkhatov: This had to switch to DOM-order indexing because of the duplicate label — it was `getByText("Add sacrificial BTCVault")`, an unambiguous selector.

It does still discriminate (the urgent CTA calls `onDeposit()` with no arg), so it isn't tautological, but it now silently depends on push order in `buildBannerActions`: reordering the two `actions.push` calls flips which button this drives, with no compile error.

Related: every `.find((a) => a.label === COPY.banner.addCollateral)` in `BannerActions.test.ts` (lines 110, 122, 138, 166, 172) is now keyed on a non-unique label, and only avoids matching the wrong action because `makeResult()` defaults to `warnings: []` / `suggestedNewVaultBtc: null`. There's no test for the both-active case at all — worth adding one that asserts the two actions' `onClick` payloads.

packages/babylon-core-ui/src/components/Notification/Notification.stories.tsx
```diff
@@ -109,13 +109,13 @@ export const WarningNoIconWithLabel: Story = {
     icon: null,
     title: "First liquidation takes everything",
     children:
```
gbarkhatov: Small correction to the PR body's "prettier clean on every touched file" — this file isn't. `cd packages/babylon-core-ui && pnpm exec prettier --check src/components/Notification/Notification.stories.tsx` reports style issues.

The package has its own `.prettierrc` with `"printWidth": 120`; these lines are wrapped at 80, so they were formatted with the root/vault config rather than the package's. Prettier at 120 would re-join them:

```
-    children:
-      "With your current BTCVaults, a single liquidation event liquidates all your BTC in collateral.",
+    children: "With your current BTCVaults, a single liquidation event liquidates all your BTC in collateral.",
```

The file was already dirty on `main`, so this isn't a regression — but running `prettier --write` from inside the package would settle it, or the claim could be narrowed to the vault files (which all do pass).

gbarkhatov (dismissed):
Reviewed the full diff with attention to the renamed financial labels, the "no behaviour changes" claim, dead code, and the terminology sweep. Approving — nothing here is unsafe to ship and none of it touches the critical paths in CLAUDE.md — but a handful of the relabels below are factually wrong about the value they sit next to, and I'd want those fixed before merge rather than after.

**Verification (all re-run locally at `8a4364b`, matching the description):** `tsc --noEmit` clean on both packages; eslint 0 errors (124 vault / 213 core-ui pre-existing warnings); vault vitest **3602 passed / 6 skipped (3608)** with both skipped files outside the diff; core-ui **113/113**. Every removed COPY key (`badgeSacrificial`, `badgeProtected`, `containVaults`, `addSacrificialVault`, `totalCollateralValueTooltip`, `debtTooltip`, `availableTooltip`) has zero remaining references anywhere in `services/` or `packages/`, the deleted `ReorderVaults/constants.ts` has no importers, and no `data-testid` was moved or dropped — I checked the real-wallet E2E CLI for text selectors matching the changed strings and found none. One correction to the description: prettier is not clean on `Notification.stories.tsx` (inline comment).

**Also worth confirming:** the description says the health-factor tooltip "now names the 1.0 liquidation threshold". That string is byte-identical to `origin/main` — it already said that. 1.0 *is* the right threshold (`healthFactor.ts:56-59`) and it's correct on all five surfaces it propagates to, so no action needed; the changelog entry is just misattributed.

**Confirmed correct, so it doesn't get re-litigated:** `"Liquidation threshold (THF)"` -> `"Target Health Factor"` is a genuine **fix** — `useProtocolFeeRows.ts:70` renders `targetHealthFactor`, so the old label was mislabeling it. Dropping "LTV" is safe (CF and LTV are the same number in this protocol). All three "Available Liquidity" surfaces do resolve to the same `useAaveReserveLiquidity().availableLiquidity`. `useDepositFlow.ts` and `ResumeDepositContent.tsx` are comment-only, and moving the reorder strings into `copy.ts` is a compliance improvement.

**The three items flagged as "needs a product call" — two are confirmed real by the code, not just by the frame:** "Expired Deposits" genuinely mislabels settled rows, and the duplicate CTA genuinely has no accessible-name escape hatch today. Details inline.

---

A few findings land on lines outside the diff hunks, so they can't be inline comments:

**1 — `copy.ts:22-31` still mandates the spelling this PR abandons.** The style header requires *"BTC Vault (capitalized, two words) ... never bare 'vault'"*, and CLAUDE.md makes that header binding. All ten new `BTCVault` strings violate the file's own contract, and so do the two new bare-"vault" strings. This is the root cause of the terminology split I flagged at 2096 and of the unit loss at 2103 — worth updating the header (and its `"Add BTC Vault"` example) in this PR, citing the glossary from #2309.

**2 — three stale dust fixtures in a file this PR did update.** `PositionNotificationBanner.test.tsx:500`, `:552` and `:686` still hardcode `"Position too small for BTC Vault analysis"`, the title renamed at `copy.ts:2152`. The cliff fixtures in the same file *were* updated, so it's a half-sweep.

**3 — `PostDepositContinuationView.test.tsx:149` now passes against a stale fixture.** It mocks `@/copy` with `heading: "Vault activated"` / `body: "Your vault is now active..."` and asserts `getByText("Vault activated")` at `:520` and `:546`. Production is now `"BTCVault activated"` / `"Your BTCVault is now active..."`. The file isn't in the diff so nothing failed — the test has silently stopped tracking the real copy. Worth updating the mock, or asserting against `COPY.deposit.vaultActivatedSuccess.heading` so it can't drift again.

**4 — comments and dev strings left describing the old section names.** `VaultsLifecycleSections.tsx:6-8` ("inactive is the v3 name for expired" — now inverted) and `:660`; `shared/ListRow.tsx:2`; `e2e/real/actions/stepMachine.ts:63` (cites `"BTC Vault activated"`); `copy.ts:112`/`:115` (tooltip consumer lists, now wrong in both directions). Note `dev/demoDeposit.ts:798` and `:804` are user-visible dev-panel strings rather than comments, and `:804` now has the v2/v3 names exactly backwards.

**5 — inlined user-facing strings that the same rule would have moved.** This PR correctly migrated `ReorderVaults/constants.ts` into `COPY.reorder`, but left the structurally identical `PositionNotificationBanner/constants.ts:3` (a banner body, and it carries the old term). Also `AddressScreeningBanner.tsx:22` (inline JSX text, old term), `useRefundState.ts:94`, `useVaultActions.ts:884` — the last of which is internally split, since two other strings in that same file were re-termed by #2309.

**6 — one redundant test.** `liquidationChartData.test.ts:402-415` now asserts the same title array on the same input as the test at `:386`, and its tone half duplicates `:417-428`. The two were distinct before the badge assertions came out; per the no-test-bloat rule one can go.

jonybur (dismissed):
CI is now green across every check. The direction is right and the earlier review from gbarkhatov covered the substantive items.

jonybur (approved):
New commits since the last review are focused: split the cliff CTA off the urgent "Add Collateral" label with an "Add <amount> BTC" variant, add a health-factor tooltip to the repay card, fold vault names into the chart band's accessible name via a new `accessibleDetail` field, rename the protocol-parameters row to Max Liquidation Penalty. Verified `useProtocolFeeRows.ts` now reads `COPY.protocolFees.maxLiquidationPenalty` on both label and tooltip so the rename doesn't leave a dangling reference. Both fairness builders in `liquidationChartData.ts` supply a tooltip so tightening `LiquidationEventCard.fairness.tooltip` to required is consistent with every caller. The stale header-style items already flagged earlier still stand; nothing new.

PR #2325, opened 2026-08-26
feat(vault, wallet-connector): tick the signing counters per Ledger c…

jrwbabylonlab:
Ledger signs each PSBT in a `signPsbts` batch as its own device ceremony,
so "(x of n)" jumped 0→N only when the whole batch returned. The provider
now exposes `subscribeSigningProgress` (optional, mirrors `cancelSigning`)
and emits `{completed,total}` after each ceremony commits; the payout,
depositor-graph and split-deposit Pre-PegIn wrappers subscribe for the
duration of the call. Extension wallets don't implement it and keep the
single jump.

### DIFF
--- a/packages/babylon-wallet-connector/src/core/types.ts
+++ b/packages/babylon-wallet-connector/src/core/types.ts
@@ -423,10 +423,11 @@ export interface IBTCProvider extends IProvider {
   /**
    * Subscribes to per-PSBT progress inside `signPsbts`: fires after each
    * device ceremony commits, with `completed` = signatures the call will
-   * return so far and `total` = the batch length. Never fires for `signPsbt`,
-   * never before the first ceremony, never on rejection — the promise settle
-   * is the terminal signal. Listener throws are swallowed. Returns the
-   * unsubscribe.
+   * return so far and `total` = the batch length. Never fires for `signPsbt`
+   * or before the first ceremony. A ceremony that fails emits no tick, while
+   * ticks for ceremonies that committed before a later rejection stand — the
+   * promise settle is the terminal signal. Listener throws are swallowed.
+   * Returns the unsubscribe.
    *
    * Implemented only by hardware providers that run one device ceremony per
    * PSBT (currently the Ledger vault provider). Optional — callers MUST
--- a/packages/babylon-wallet-connector/src/core/types.ts
+++ b/packages/babylon-wallet-connector/src/core/types.ts
@@ -422,8 +422,8 @@ export interface IBTCProvider extends IProvider {
 
   /**
    * Subscribes to per-PSBT progress inside `signPsbts`: fires after each
-   * device ceremony commits, with `completed` = signatures the call will
-   * return so far and `total` = the batch length. Never fires for `signPsbt`
+   * device ceremony commits, with `completed` = PSBT ceremonies committed
+   * so far and `total` = the batch length. Never fires for `signPsbt`
    * or before the first ceremony. A ceremony that fails emits no tick, while
    * ticks for ceremonies that committed before a later rejection stand — the
    * promise settle is the terminal signal. Listener throws are swallowed.
--- a/packages/babylon-wallet-connector/src/core/types.ts
+++ b/packages/babylon-wallet-connector/src/core/types.ts
@@ -361,6 +361,12 @@ export interface SignPsbtOptions {
   displayMessage?: string;
 }
 
+/** One committed device ceremony inside a `signPsbts` batch. Counts only — never payload. */
+export interface SigningProgress {
+  readonly completed: number;
+  readonly total: number;
+}
+
 export interface IBTCProvider extends IProvider {
   /**
    * Signs the given PSBT in hex format.
@@ -413,6 +419,21 @@ export interface IBTCProvider extends IProvider {
    * hide the cancel affordance when the method is missing.
    */
   cancelSigning?(): void;
+
+  /**
+   * Subscribes to per-PSBT progress inside `signPsbts`: fires after each
+   * device ceremony commits, with `completed` = signatures the call will
+   * return so far and `total` = the batch length. Never fires for `signPsbt`,
+   * never before the first ceremony, never on rejection — the promise settle
+   * is the terminal signal. Listener throws are swallowed. Returns the
+   * unsubscribe.
+   *
+   * Implemented only by hardware providers that run one device ceremony per
+   * PSBT (currently the Ledger vault provider). Optional — callers MUST
+   * feature-detect (`typeof provider.subscribeSigningProgress === "function"`)
+   * and fall back to the batch settle when the method is missing.
+   */
+  subscribeSigningProgress?(listener: (progress: SigningProgress) => void): () => void;
 
   /**
    * Signs a message using either BIP322-Simple or ECDSA signing method.
--- a/packages/babylon-wallet-connector/src/core/types.ts
+++ b/packages/babylon-wallet-connector/src/core/types.ts
@@ -426,8 +426,13 @@ export interface IBTCProvider extends IProvider {
    * so far and `total` = the batch length. Never fires for `signPsbt`
    * or before the first ceremony. A ceremony that fails emits no tick, while
    * ticks for ceremonies that committed before a later rejection stand — the
-   * promise settle is the terminal signal. Listener throws are swallowed.
-   * Returns the unsubscribe.
+   * promise settle is the terminal signal. Listeners run synchronously inside
+   * the signing loop: a throw is swallowed, a returned promise is neither
+   * awaited nor observed. Ticks carry no batch identity, so implementers MUST
+   * reject an overlapping `signPsbts` rather than queue it. Returns the
+   * unsubscribe. The subscription outlives a batch but not the session —
+   * teardown (disconnect or dead-session cleanup) drops every listener, so
+   * subscribe per sign call.
    *
    * Implemented only by hardware providers that run one device ceremony per
    * PSBT (currently the Ledger vault provider). Optional — callers MUST
--- a/packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
+++ b/packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
@@ -1069,6 +1069,20 @@ describe("LedgerVaultProvider", () => {
         expect(ticks).toEqual([]);
       });
 
+      it("a subscriber from before disconnect receives nothing from the next connection's batch", async () => {
+        const provider = await approved();
+        const ticks: SigningProgress[] = [];
+        provider.subscribeSigningProgress((progress) => ticks.push(progress));
+
+        await provider.disconnect();
+        await provider.connectWallet();
+        await provider.deriveContextHash("app", "aa".repeat(32));
+        await provider.approveDepositTerms(TERMS);
+        await provider.signPsbts([PSBT_B]);
+
+        expect(ticks).toEqual([]);
+      });
+
       it("signPsbt never emits signing progress", async () => {
         const provider = await approved();
         const ticks: SigningProgress[] = [];
@@ -1089,6 +1103,28 @@ describe("LedgerVaultProvider", () => {
         expect(signMock.signPreparedVaultPsbt).toHaveBeenCalledTimes(2);
       });
 
+      it("every registered listener receives every tick, even when one between them throws", async () => {
+        const provider = await approved();
+        const before: SigningProgress[] = [];
+        const after: SigningProgress[] = [];
+        provider.subscribeSigningProgress((progress) => before.push(progress));
+        provider.subscribeSigningProgress(() => {
+          throw new Error("display bug");
+        });
+        provider.subscribeSigningProgress((progress) => after.push(progress));
+
+        await provider.signPsbts([PSBT_A, PSBT_B]);
+
+        expect(before).toEqual([
+          { completed: 1, total: 2 },
+          { completed: 2, total: 2 },
+        ]);
+        expect(after).toEqual([
+          { completed: 1, total: 2 },
+          { completed: 2, total: 2 },
+        ]);
+      });
+
       it("the returned unsubscribe stops further ticks and is idempotent", async () => {
         const provider = await approved();
         const ticks: SigningProgress[] = [];
@@ -1100,6 +1136,18 @@ describe("LedgerVaultProvider", () => {
         await provider.signPsbts([PSBT_B]);
 
         expect(ticks).toEqual([{ completed: 1, total: 1 }]);
+      });
+
+      it("a listener subscribed inside a tick first hears the next tick", async () => {
+        const provider = await approved();
+        const lateTicks: SigningProgress[] = [];
+        provider.subscribeSigningProgress(({ completed }) => {
+          if (completed === 1) provider.subscribeSigningProgress((progress) => lateTicks.push(progress));
+        });
+
+        await provider.signPsbts([PSBT_A, PSBT_B]);
+
+        expect(lateTicks).toEqual([{ completed: 2, total: 2 }]);
       });
 
       it("is an own property that survives the dApp's object spread into a wrapper", async () => {
--- a/packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
+++ b/packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
@@ -23,6 +23,7 @@ import * as ecc from "@bitcoin-js/tiny-secp256k1-asmjs";
 import { initEccLib, payments, Psbt } from "bitcoinjs-lib";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
+import type { SigningProgress } from "@/core/types";
 import { Network } from "@/core/types";
 import { getTaprootAddress, toNetwork } from "@/core/utils/wallet";
 import { ERROR_CODES, WalletError } from "@/error";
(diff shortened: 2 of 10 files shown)


jonybur (dismissed):
The provider-side subscribe surface is tight: ticks fire only after the ceremony commit inside the generation-gated loop, so the stale-batch-after-reconnect and cancel-mid-element cases correctly emit only the committed ticks. Listener errors are swallowed on purpose, matching the per-YIELD onProgress contract. The dApp-side wrapper snapshots the phase before any await, and stopObserving fires in the finally on both resolve and reject paths in both usePayoutSigningState and useDepositFlow. observeSigningProgress feature-detects the affordance and preserves the receiver via method-style call, which the prototype-provider test pins. Test coverage is thorough — the 'ticks do not flip claimers-done ref' test in particular closes a real footgun. Extension-wallet path unchanged.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -319,39 +320,45 @@ export function usePayoutSigningState({
...
                     );
-                  } finally {
-                    if (claimersDoneRef.current) {
+                    if (phase === "graph") {
                       setProgress({
-                        phase: "graph",
+                        phase,
                         completed: hexes.length,
                         total: hexes.length,
                       });
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — `"claimers"`/`"graph"` are the `PayoutSigningPhase` union members already used as literals in both files; 0/1 are a start count and a single-PSBT total, not magic

packages/babylon-wallet-connector/src/core/types.ts
```diff
@@ -414,6 +420,21 @@ export interface IBTCProvider extends IProvider {
    */
   cancelSigning?(): void;
 
+  /**
+   * Subscribes to per-PSBT progress inside `signPsbts`: fires after each
+   * device ceremony commits, with `completed` = signatures the call will
+   * return so far and `total` = the batch length. Never fires for `signPsbt`,
+   * never before the first ceremony, never on rejection — the promise settle
+   * is the terminal signal. Listener throws are swallowed. Returns the
+   * unsubscribe.
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/types.ts
```diff
@@ -414,6 +420,22 @@ export interface IBTCProvider extends IProvider {
    */
   cancelSigning?(): void;
 
+  /**
+   * Subscribes to per-PSBT progress inside `signPsbts`: fires after each
+   * device ceremony commits, with `completed` = signatures the call will
+   * return so far and `total` = the batch length. Never fires for `signPsbt`
+   * or before the first ceremony. A ceremony that fails emits no tick, while
+   * ticks for ceremonies that committed before a later rejection stand — the
+   * promise settle is the terminal signal. Listener throws are swallowed.
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

jonybur (dismissed):
The provider-side subscribe surface is tight: ticks fire only after the ceremony commit inside the generation-gated loop, so the stale-batch-after-reconnect and cancel-mid-element cases correctly emit only committed ticks. Listener errors are swallowed on purpose, matching the per-YIELD onProgress contract. subscribeSigningProgress is an own arrow-property, not a prototype method, so it survives the dApp's object spread — pinned by the test. The dApp-side wrapper snapshots the phase before any await, and stopObserving fires in the finally on both resolve and reject paths in both usePayoutSigningState and useDepositFlow. observeSigningProgress feature-detects the affordance and preserves the receiver via method-style call, which the prototype-provider test also pins. The 'ticks do not flip claimers-done ref' test closes a real footgun. Extension-wallet path unchanged.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -182,6 +182,7 @@ export class LedgerVaultProvider implements IBTCProvider {
   private activeOperation: symbol | undefined;
   /** Abort handle into the in-flight signing loop; fired by teardown (B3's only abort source). */
   private signAbortController: AbortController | undefined;
+  private readonly signingProgressListeners = new Set<(progress: SigningProgress) => void>();
```
gbarkhatov, line 196: **The listener set outlives the connection — `teardownSession` never clears it.**

`teardownSession` (`:328-346`) invalidates every other piece of connection-scoped state — `deviceState`, `signedFingerprints`, `pubkeyHexPromise`, `policyContextPromise`, `activeOperation`, `signAbortController` — but not this set.

That matters here specifically because `activeOperation` is released *synchronously* by design (`:176-181`) so a new connection can operate while a stale call is still settling, and the stale caller's `finally { stopObserving() }` cannot run until the device settles — which per the `cancelSigning` doc at `:738-741` may be "only after the user acts on the device". So: batch A awaiting `signPreparedVaultPsbt` → `disconnect()` / `connectWallet()` → batch B → `emitSigningProgress` walks a set that still holds A's listener. Reproduced against the real provider: after disconnect/reconnect, a fresh unrelated `signPsbts([PSBT_B])` delivered `[{completed:1,total:1}]` to the abandoned flow's subscriber — and that listener is wired straight into `setPeginSigningProgress` / `setProgress` (`useDepositFlow.ts:572`, `usePayoutSigningState.ts:346`).

Both current consumers do unsubscribe on every settle path, so the window is narrow — but nothing in the contract says the subscription is connection-scoped, so a consumer can't know to compensate.

Note the existing test *"a stale batch settling after reconnect emits nothing"* pins only the converse: the stale **batch** doesn't emit. It doesn't cover the stale **subscriber**.

Suggest `this.signingProgressListeners.clear()` in `teardownSession`, plus a line in the `IBTCProvider` JSDoc and the guide row stating that the subscription is provider-global, outlives the batch, and is dropped on disconnect. (The guide row would also read better naming the exported `SigningProgress` type — that's what consumers actually import.)

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -740,6 +744,26 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+    this.signingProgressListeners.add(listener);
+    return () => {
+      this.signingProgressListeners.delete(listener);
+    };
+  };
+
+  // Display-only: a listener bug must not abort a non-idempotent ceremony
+  // (same contract as the signer's per-YIELD onProgress).
+  private emitSigningProgress(progress: SigningProgress): void {
+    for (const listener of this.signingProgressListeners) {
```
gbarkhatov: **This iterates the live `Set`.** JS Set iteration visits entries added *during* iteration, which costs two properties:

- A listener that subscribes another listener inside a tick delivers *that same tick* to a subscriber that did not exist when the ceremony committed — contradicting "fires after each device ceremony commits". Confirmed: a listener subscribed during tick 1 of a 2-PSBT batch received `[{1,2},{2,2}]`, not `[{2,2}]`.
- A listener that registers a fresh closure per tick never lets the loop terminate. That hangs `signPsbts` *between elements* of a non-idempotent hardware ceremony — the exact failure mode the comment two lines up exists to prevent, and one the per-listener `try/catch` does not cover.

Both need a pathological listener, but the fix is free: `for (const listener of [...this.signingProgressListeners])`.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -725,6 +726,9 @@ export class LedgerVaultProvider implements IBTCProvider {
           }
           this.assertSameConnection(ctx.generation);
           signed.push(await this.signStaged(one, ctx, controller));
+          // After the commit (generation checked, fingerprint recorded), so a
```
gbarkhatov: **This comment asserts an invariant that isn't true, and that the interface JSDoc contradicts.**

`signPsbts` is fail-fast: if element k+1 rejects, the promise rejects and the caller receives *none* of the signatures — including the k already ticked. So a tick routinely counts a signature the call will not return. `types.ts:427-429` states the actual contract ("ticks for ceremonies that committed before a later rejection stand — the promise settle is the terminal signal"), and the new test *"a failing element emits no tick; the ticks already committed stand"* demonstrates exactly the case this comment denies.

What the placement actually buys is narrower and worth stating instead: the tick fires only after `signStaged`'s `assertSameConnection` and `signedFingerprints.add`, so a *stale or failed* ceremony never ticks. Suggest rewording to that guarantee and dropping the "will not return" clause.

packages/babylon-wallet-connector/src/core/types.ts
```diff
@@ -414,6 +420,22 @@ export interface IBTCProvider extends IProvider {
    */
   cancelSigning?(): void;
 
+  /**
+   * Subscribes to per-PSBT progress inside `signPsbts`: fires after each
+   * device ceremony commits, with `completed` = PSBT ceremonies committed
+   * so far and `total` = the batch length. Never fires for `signPsbt`
+   * or before the first ceremony. A ceremony that fails emits no tick, while
+   * ticks for ceremonies that committed before a later rejection stand — the
+   * promise settle is the terminal signal. Listener throws are swallowed.
```
gbarkhatov: Two precision gaps in this contract, worth closing now that it's a public optional method third parties are invited to implement.

**1. "Listener throws are swallowed" over-promises for async listeners.** The listener type is `(progress: SigningProgress) => void`, and TypeScript's void-return special case accepts an `async` function there without complaint. The implementation's `try/catch` catches only synchronous throws, so an async listener that rejects produces an unhandled rejection reaching `window.onunhandledrejection`. Either state here that the listener must be synchronous (ticks are delivered synchronously inside the signing loop), or handle the async case.

**2. Nothing says a provider must serialize `signPsbts`.** Ticks carry no batch identity, so every consumer attributes a tick to whichever batch it believes is in flight. That is only sound because `LedgerVaultProvider.withDeviceOperation` (`provider.ts:193-208`) rejects a second concurrent ceremony. A provider honouring the documented contract but allowing concurrent batches would silently mislabel ticks. Worth one sentence.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
```diff
@@ -885,6 +886,162 @@ describe("LedgerVaultProvider", () => {
...
+        const provider = await approved();
+        const ticks: SigningProgress[] = [];
+        provider.subscribeSigningProgress((progress) => ticks.push(progress));
+
+        await provider.signPsbt(PSBT_A);
+
+        expect(ticks).toEqual([]);
+      });
+
+      it("a listener that throws neither aborts the batch nor changes what it returns", async () => {
```
gbarkhatov, line 1096: Every one of the eleven new tests registers exactly **one** listener, which leaves two properties unpinned:

- A single-slot implementation (`this.signingProgressListener = listener`) would pass all eleven. The `Set` — the thing that makes this a multi-subscriber affordance — is untested.
- This test can't distinguish the per-listener `try/catch` *inside* the loop from a `try` wrapped around the whole loop. Moving it outward would still pass here, while silently starving every listener registered after the throwing one — which is precisely the guarantee the swallow exists for.

One test closes both: two subscribers, the first throws, assert the second still receives every tick.

services/vault/src/utils/signingProgress.ts
```diff
@@ -0,0 +1,30 @@
...
+export function observeSigningProgress(
+  provider: unknown,
+  listener: SigningProgressListener,
+): () => void {
+  const candidate = provider as Pick<
+    IBTCProvider,
+    "subscribeSigningProgress"
+  > | null;
+  return typeof candidate?.subscribeSigningProgress === "function"
+    ? candidate.subscribeSigningProgress(listener)
```
gbarkhatov: **The provider's return value isn't validated, and the failure mode discards the real signing error.**

This feature-detects the *presence* of the method but returns whatever it hands back, unchecked. All three call sites invoke `stopObserving()` from a `finally` (`useDepositFlow.ts:584`, `:1189`, `usePayoutSigningState.ts:361`). A provider that implements `subscribeSigningProgress` but returns `undefined` — or anything non-callable — makes that `finally` throw `TypeError: stopObserving is not a function`, replacing the in-flight rejection or failing a batch that actually signed.

Downstream that isn't cosmetic: `isUserCancellation(err)` then returns false, so a routine user cancel routes into `captureFunnelFailure` telemetry with the wrong error copy.

The interface is public and optional, so third-party / injectable wallets can implement it — feature-detecting the method's presence without validating its return is half a guard. Per "validate all data received from wallet APIs before use", validate at subscribe time, before any device I/O, so it can never mask a signing error:

```ts
const stop = candidate.subscribeSigningProgress(listener);
if (typeof stop !== "function") {
  throw new Error(
    `provider.subscribeSigningProgress must return an unsubscribe function; got ${typeof stop}`,
  );
}
return stop;
```

(If cosmetic progress must never be able to fail signing, `typeof stop === "function" ? stop : () => {}` also works — but that's a silent fallback, so the throw fits this repo better.)

Related, on the subscribe side: the call sits *outside* the `try` at all three sites. At `useDepositFlow.ts:1169` it runs after `advanceStep(SIGN_DEPOSITOR_GRAPH)` and before `setIsWaiting(false)`, so a throwing `subscribeSigningProgress` would leave the step advanced with no sign attempt. The emit side is deliberately defensive about exactly this class of bug; the subscribe side should be too.

services/vault/src/utils/signingProgress.ts
```diff
@@ -0,0 +1,30 @@
...
+  IBTCProvider,
+  SigningProgress,
+} from "@babylonlabs-io/wallet-connector";
+
+type SigningProgressListener = (progress: SigningProgress) => void;
+
+/**
+ * Subscribes when the provider implements the affordance; otherwise returns a
+ * no-op unsubscribe so callers never branch. Takes `unknown` because callers
+ * hold the signing provider behind an untyped ref.
```
gbarkhatov: **This justification doesn't hold for any call site.** "Takes `unknown` because callers hold the signing provider behind an untyped ref" is true of the sibling `cancelSigning.ts` — `usePayoutSigningState.ts:167` is a literal `useRef<unknown>(null)` — but neither `observeSigningProgress` call site goes through a ref: `useDepositFlow.ts:571` and `:1170` pass `confirmedBtcWallet`, `usePayoutSigningState.ts:345` passes `wallet`, all statically typed. Reads like the doc was carried over from the sibling along with the signature.

`unknown` also switches off argument checking at this seam: any object at all type-checks, and a mismatch degrades silently to the no-op branch rather than erroring. Suggest tightening the parameter to `BitcoinWallet | null` (keeping the `Pick<IBTCProvider, …>` cast internal), or exporting a `supportsSigningProgress` type guard mirroring the sibling — and either way correcting or deleting the doc line.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -558,21 +559,30 @@ export function useDepositFlow(
...
           psbtHexes,
           opts,
         ) => {
           advanceStep(DepositFlowStep.SIGN_PEGIN_BTC);
           setPeginSigningProgress({ completed: 0, total: psbtHexes.length });
-          const signed = await runCancellableSign(confirmedBtcWallet, () =>
-            confirmedBtcWallet.signPsbts!(psbtHexes, opts),
+          const stopObserving = observeSigningProgress(
+            confirmedBtcWallet,
+            (tick) => setPeginSigningProgress(tick),
```
gbarkhatov: Two small things at this listener, both of which apply to the two siblings as well (`:1171` and `usePayoutSigningState.ts:346`).

**1. `tick.total` becomes the rendered denominator, unchecked.** All three listeners write the provider's `total` verbatim, overwriting the denominator the same wrapper computed from `psbtHexes.length` three lines up. The provider's `total` is the length of *that* `signPsbts` call (`provider.ts:731`); the SDK announces the *round* total separately (`runDepositorPresignFlow.ts:372`, `:399`).

They agree today — checked on all three paths: claimers is exactly one `signPsbts` (`runDepositorPresignFlow.ts:376-380` → `PayoutManager.ts:385-388`, a direct `signPsbts!`), graph is one call (`signDepositorGraph.ts:561` → `signPsbtsWithFallback.ts:36-44`), peg-in is one call with one PSBT per vault (`PeginManager.ts:1109`). So **no live display bug**. But that invariant lives in three other packages, is unasserted at the render seam, and fails silently: if a claimers round is ever chunked, or a non-Babylon `IBTCProvider` adopts a different `total` convention, the user sees `(8 of 12)` → `(1 of 4)` — a counter going backwards with a changed denominator. It renders ungated at `components/simple/DepositProgressView/steps.ts:14-40`.

One line per site:

```ts
(tick) => {
  if (tick.total !== psbtHexes.length) return; // not our batch / foreign convention
  setPeginSigningProgress({ ...tick });
}
```

**2. This site stores the provider's own object in React state.** Both payout sites copy it (`{ phase, ...tick }`); this one passes it straight through. `SigningProgress`'s `readonly` fields are a compile-time hint, not a freeze — a provider reusing one mutable object per batch (a legal implementation of the published interface) would mutate committed React state and defeat the referential-equality re-render. `setPeginSigningProgress({ ...tick })` either way.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -319,39 +320,45 @@ export function usePayoutSigningState({
             }
           },
           signPsbt: async (hex, opts) => {
-            if (claimersDoneRef.current) {
+            // Snapshot: the ref flips only from the SDK's onProgress, after this call returns.
+            const isGraph = claimersDoneRef.current;
+            if (isGraph) {
               setProgress({ phase: "graph", completed: 0, total: 1 });
```
gbarkhatov, line 326: **Pre-existing, not introduced here — flagging only because it's the same class of dishonest counter this PR exists to fix.**

`signPsbtsWithFallback` routes a lone PSBT to `signPsbt`, but it *also* falls back to a sequential `signPsbt` loop for any wallet that doesn't implement `signPsbts` (`signPsbtsWithFallback.ts:46-50`). On such a wallet an N-PSBT depositor graph enters this branch N times, so the step renders "0 of 1 → 1 of 1" N times instead of "k of N". Same shape at `useDepositFlow.ts:1126-1148`.

Out of scope for this PR; worth an issue.

services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
```diff
@@ -1410,4 +1410,326 @@ describe("usePayoutSigningState", () => {
...
+
+      expect(result.current.progress).toEqual({
+        phase: "graph",
+        completed: 0,
+        total: 1,
+      });
+    });
+
+    it("progress ticks do not flip the claimers-done ref before the SDK reports the round complete", async () => {
+      // A full-count tick during Phase 3 must still be labelled "claimers";
```
gbarkhatov: Two issues with this test.

**The scenario it calls the discriminator is unreachable in the real SDK.** The mock drives two `btcWallet.signPsbts` calls inside Phase 3 before the SDK's own `(N,N)`. `signPayoutTransactions` (`runDepositorPresignFlow.ts:371-390`) makes exactly one `signPayoutTransactionsBatch` call between `onProgress(0,N)` and `onProgress(N,N)`, and that helper calls `signPsbts` once (`PayoutManager.ts:385`). So the comment presenting the double entry as *the* discriminator overstates it, and the test can outlive the invariant it claims to protect.

**It bakes in a count that would be wrong if the scenario were real.** It asserts `{completed: 2, total: 2}` after handing 4 PSBTs over across the two batches — encoding "2 of 2" as correct for a 4-PSBT round, which is exactly the denominator problem flagged on `useDepositFlow.ts:572`.

The invariant being pinned (a tick must not flip `claimersDoneRef`) is real and worth a test. Either correct the comment to say the second entry is a synthetic probe the SDK does not emit, or restructure it to discriminate on the single-batch shape.

services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
```diff
@@ -1410,4 +1410,326 @@ describe("usePayoutSigningState", () => {
...
+        total: 2,
+      });
+
+      await act(async () => {
+        settle.resolve(["c", "d"]);
+        await done;
+      });
+    });
+
+    it("does not subscribe on a wallet without the affordance", async () => {
```
gbarkhatov: **This doesn't test its name, and it passes on reverted production code.** The mock provider a few lines down has no `subscribeSigningProgress` at all, so there is nothing that *could* subscribe; the only assertions are `signPsbts` called once and `isComplete === true`.

Checked empirically: with `usePayoutSigningState.ts` and `useDepositFlow.ts` restored from `main`, 13 of the 15 new tests fail — and this is one of the two survivors. Its stated behaviour is already covered by `utils/__tests__/signingProgress.test.ts:35-38`.

Either delete it, or make it test its name: give the wallet a non-function `subscribeSigningProgress` and assert the batch still completes, which would also cover the unvalidated-handle point on `signingProgress.ts:28`.

(The other revert-survivor, `useDepositFlow.test.tsx:2704`, **is** justified — it guards the extension-wallet 0→N behaviour on both wrappers.)

services/vault/src/hooks/deposit/__tests__/useDepositFlow.test.tsx
```diff
@@ -2306,6 +2366,385 @@ describe("useDepositFlow", () => {
...
+        total: 1,
+      });
+
+      await act(async () => {
+        release();
+        await flow;
+      });
+    });
+
+    it("the signing-progress subscription is torn down on every settle path", async () => {
```
gbarkhatov: Three scenarios in one `it`: resolve (`:2619-2648`), reject (`:2650-2676`) and cancel (`:2678-2700`), each rendering its own hook and asserting its own `unsubscribe`. A failure in the first aborts the other two, and the test name can't say which settle path leaked. Repo rule is one behaviour per test — suggest three `it(...)` blocks.

Separate and **unconfirmed**, but this block is the deepest-reaching in the file: one cold-transform run had all six tests in `per-ceremony payout progress` fail together, and I could not reproduce it across ~12 subsequent runs (including a cold `node_modules/.vite` and the full 304-file suite). The shape matches the default 1s `waitFor` budget being exceeded on a slow first run. An explicit `{ timeout }` on the `waitFor` calls here would be cheap insurance.

gbarkhatov (dismissed):
Reviewed with focused sub-agents across the provider, the two dApp wrappers, the new util and the test suite.

The design is sound and the core behaviour change is right: moving the counter assignment out of `finally` stops the UI reporting `N/N` after a ceremony that failed, and every consumer feature-detects and unsubscribes on all settle paths. Verified independently along the way — the `phase` / `isGraph` snapshot comments are accurate (both refs are written only from the SDK's `onProgress`, which fires between wallet calls, never during an awaited sign); all three `observeSigningProgress` call sites pass the real `LedgerVaultProvider`, not a spread copy; the other nine `IBTCProvider` implementors are unaffected by the optional method; and nothing sensitive crosses the progress path.

**Verification run:** vault `signingProgress` + `useDepositFlow` + `usePayoutSigningState` 133/133; wallet-connector ledger-vault provider 119/119; `pnpm nx run-many -t lint` exit 0 (89 pre-existing warnings, none in changed files); `tsc --noEmit -p services/vault/tsconfig.lib.json` exit 0 after rebuilding wallet-connector.

Approving — the inline comments are hardening and precision, none of them blocking. Three items have no line in the diff to hang on:

**a. `services/vault/src/hooks/deposit/useDepositFlow.ts:416` — `payoutSigningProgress` isn't reset at flow start.** `executeDeposit` resets `setPeginSigningProgress(null)` but not the payout counter, which is nulled only at `:1510`, after the per-vault loop. On a failed or aborted payout the stale value survives into the next `executeDeposit`. Latent today only because `DepositSignContent` passes no `onRetry` (`DepositProgressView.tsx:496-497` falls back to `onClose`), but the moment an in-place Retry is wired the retry's early steps render a stale `(1 of 3)` on "Sign recovery transactions". One line next to 416.

**b. `services/vault/src/services/vault/vaultTransactionService.ts:120-128` — stale contract doc.** The `PeginSigningProgress` JSDoc still says a multi-vault batch "signs in one native popup for `signPsbts`-capable wallets (`completed` jumps 0 -> total around the one call)" — exactly what this PR changes for tick-capable providers. The twin comment at `useDepositFlow.ts:542-546` was updated; the doc on the type itself wasn't, so the type's contract now contradicts what it emits.

**c. PR description note (not a defect).** The vault resolves `@babylonlabs-io/wallet-connector` through `package.json` `exports` → `dist/index.d.ts`, not source. Against a stale `dist/`, `tsc -p tsconfig.lib.json` fails with `TS2305: Module … has no exported member 'SigningProgress'` plus a `TS2344` on the `Pick`. CI survives via nx `dependsOn: ["^build"]`, and the vault's own `typecheck` target is disabled, so nothing else catches it — but anyone building or just opening the file without rebuilding wallet-connector first sees two hard errors that look like the PR is broken. Worth a line in the description.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
```diff
@@ -885,6 +886,162 @@ describe("LedgerVaultProvider", () => {
...
+
+      it("a stale batch settling after reconnect emits nothing", async () => {
+        const provider = await approved();
+        let resolveStale: (value: { signedPsbtHex: string; yields: never[] }) => void = () => {};
+        signMock.signPreparedVaultPsbt.mockImplementationOnce(
+          () =>
+            new Promise<{ signedPsbtHex: string; yields: never[] }>((resolve) => {
+              resolveStale = resolve;
+            }),
+        );
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — one declaration per test; file compiles

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/types.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
jrwbabylonlab: fixed

services/vault/src/utils/signingProgress.ts
jrwbabylonlab: fixed

services/vault/src/utils/signingProgress.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: not planning — total is this call's own length by contract; fixed

services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
jrwbabylonlab: fixed

services/vault/src/components/deposit/PayoutSignModal/__tests__/usePayoutSigningState.test.tsx
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/__tests__/useDepositFlow.test.tsx
jrwbabylonlab: fixed; not planning — waitFor already runs on a 4s budget

jonybur (dismissed):
Prior round fully addressed. Teardown clears listeners inside teardownSession, and the tick fan-out iterates a snapshot copy so a listener that throws mid-fan-out cannot skip its successors — the multi-subscriber test with a throwing middle listener pins that. The stale contract wording in types.ts is gone; the JSDoc now spells out that ticks are synchronous and that overlapping calls are rejected. observeSigningProgress narrows from unknown to BitcoinWallet, and validates the unsubscribe return before any device I/O. The tick.total contract is now stated by the interface (= this call's own batch length) and only Ledger implements it, so an unchecked-denominator guard on the dApp side would be defensive polish, not a correctness gap. CI green.

jonybur (approved):
The subscribe/emit pair does what it advertises: emitSigningProgress fires only after signStaged commits, so a failed ceremony emits no tick and the committed ticks stand; teardownSession clears signingProgressListeners so a stale subscriber never sees the next connection's ticks; the snapshot iteration keeps subscribes-inside-a-tick from extending the loop and swallowing a listener throw. The arrow-property form on the provider survives the wrapper spread the dApp tests probe.

The consumer side is careful too: observeSigningProgress feature-detects, throws at subscribe time if the returned handle is not a function (so a bad implementation surfaces before device I/O rather than through a masked finally), and returns a no-op so callers never branch. The pegin and payout wrappers snapshot the phase before await, so a mid-call ref flip cannot re-tag ticks, and both call stopObserving in finally. Extension wallets keep the 0-to-N behaviour, covered by the affordance-missing test.

Tests cover the invariants that would actually break the counter: batch order, no ticks before the first ceremony, per-ceremony commit only, cancel/reject paths, prototype-vs-own-property, and disconnect. No blocking issues.

PR #2326, opened 2026-08-26
fix(vault, ledger-vault): retry in place after the device locks mid-d…

jrwbabylonlab:
Post-registration device errors (locked / wrong app / approval dropped) and
self-cancels now offer Retry, which switches the modal to the continuation
view — the same path the dashboard resume uses — instead of Close.


jonybur (dismissed):
The midCeremony split is the load-bearing change and it lands correctly. In the loop, only a lock whose terminal APDU is INS_CONTINUE rewrites to midCeremony=true — an initial-APDU lock keeps the constructor default of false. In the provider, the classifier flips the mirror reset off only when the error is a locked device AND !midCeremony, so a mid-loop lock still drops the intent like any other sign failure. The 'a lock mid-ceremony drops the intent like any other sign failure' provider test pins that path against a regression, and the earlier 'keeps the approved intent' pair pins the pre-dispatch case for both signPsbt and signPsbts. useDepositFlow only sets resumableVaultIds when registeredVaultIds !== null AND (device-recoverable OR the settled-cancel branch), and the initial reset inside executeDeposit clears it on retry — the 'resets resumableVaultIds when a new run starts' test pins this. handleResume calls onRefetchActivities before setContinuationVaultIds, matching the dashboard resume handoff. isDeviceRecoverableError walks the cause chain via chainCarriesCode, so a wrapped broadcast failure still classifies.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -919,7 +919,10 @@ export class LedgerVaultProvider implements IBTCProvider {
       const walletError = this.classifySignFailure(error, generation, label);
       // Mirror reset is signPsbt-only (PoP failures don't invalidate the vault
       // context); stale rejections and cancels were already handled in classify.
-      if (generation === this.connectionGeneration && !isLedgerSignPsbtAbortedError(error)) {
+      // Initial-APDU lock: the OS refused pre-dispatch (os_io_legacy.c:396-405 @ SDK v26.6.0); the classifier
+      // set {0x5515,0x6982,0x5303} is never emitted by the vault app/base pin e400d8d8. Mid-ceremony: caps may be committed.
+      const lockedBeforeDispatch = isLedgerDeviceLockedError(error) && !error.midCeremony;
```
gbarkhatov: **Blocker — `!error.midCeremony` fails open, silently re-enabling the exact unsafe retry this PR guards against.**

`isLedgerDeviceLockedError` matches on `.name` only — deliberately, per the `errors.ts` docblock, "so the classification survives duplicated module instances". An error minted by a **pre-PR signer build** therefore passes the predicate with `midCeremony === undefined`, and `!undefined === true` ⇒ `lockedBeforeDispatch === true` ⇒ the mirror is kept for a genuinely mid-ceremony lock. TypeScript cannot catch it: `dist/errors.d.ts` declares `midCeremony: boolean`, non-optional.

This is reproducible, not theoretical. Patching `dist/index.js` to set `this.midCeremony = void 0` (simulating an older build) fails exactly one test — the "a lock mid-ceremony drops the intent" assertion below. It is also how I found it: the wallet-connector suite fails against a stale `dist/` and passes once the signer is rebuilt, and CLAUDE.md itself calls stale `dist/` "a common issue" in this package.

Failure scenario: device auto-locks on a CONTINUE after 3 of 5 YIELDs → host keeps `intent-loaded` and `signedFingerprints` → user unlocks → retry sends a fresh SIGN_PSBT → device answers `0xB00A SW_CAP_EXCEEDED` and nullifies the intent. That is precisely the outcome the adjacent "never assume survival" comment and the abort-path test's "keep-intent retries risk SW_CAP_EXCEEDED" comment exist to prevent.

**Fix — invert the polarity so absence is the safe value:** `readonly preDispatch: boolean`, set `true` only at the one proven site, and check `error.preDispatch === true` here. An old build then yields `undefined` → `false` → pessimistic reset, i.e. exactly the pre-PR behaviour. Fail-closed instead of fail-open.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -919,7 +919,10 @@ export class LedgerVaultProvider implements IBTCProvider {
       const walletError = this.classifySignFailure(error, generation, label);
       // Mirror reset is signPsbt-only (PoP failures don't invalidate the vault
       // context); stale rejections and cancels were already handled in classify.
-      if (generation === this.connectionGeneration && !isLedgerSignPsbtAbortedError(error)) {
+      // Initial-APDU lock: the OS refused pre-dispatch (os_io_legacy.c:396-405 @ SDK v26.6.0); the classifier
```
gbarkhatov: **The load-bearing firmware claim is half-unverifiable, and this is a §7 critical path.**

I checked the sibling app checkout, and the verifiable half holds at the stated pins: the base-app submodule really is pinned at `e400d8d8…`; `SW_SECURITY_STATUS_NOT_SATISFIED 0x6982` appears only as a `#define` plus a `_Static_assert` and is never passed to `io_send_sw` in base or vault sources; `0x5515` / `0x5303` appear nowhere in app code. The vault app maps its own BIP32 failures to `0x6F00`, not `0x6982`.

What is **not** verifiable is the load-bearing half — that the *OS*, not the app, emits the word. `os_io_legacy.c:396-405 @ SDK v26.6.0` exists in neither repo, and neither repo pins an SDK version. Worse, the base's own `_Static_assert` ties `0x6982` to the SDK symbol `SWO_SECURITY_CONDITION_NOT_SATISFIED`, so the SDK is free to return it from paths other than the pre-dispatch lock check, and nothing pins that it doesn't.

Also a form issue: the package convention for a load-bearing pin is a module docblock legend (see `rawApdu.ts:11-13`, `signPsbtLoop.ts:12-16`). `provider.ts` has no such legend, and this inline comment introduces a third citation style. Please move it to a docblock and state the base-app half in terms a reviewer can actually re-check.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -919,7 +919,10 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+      // set {0x5515,0x6982,0x5303} is never emitted by the vault app/base pin e400d8d8. Mid-ceremony: caps may be committed.
+      const lockedBeforeDispatch = isLedgerDeviceLockedError(error) && !error.midCeremony;
+      if (generation === this.connectionGeneration && !isLedgerSignPsbtAbortedError(error) && !lockedBeforeDispatch) {
```
gbarkhatov, line 949: **The batch keep-intent path is unreachable for every production caller, and dead-ends worse than before.**

Preserving `signedFingerprints` is right in principle — device dedup masks persist. But no SDK caller does per-PSBT resume: `signPsbtsWithFallback` ← `signDepositorGraph` rebuilds and resubmits the **full** PSBT list on every attempt.

Scenario: `signPsbts([payout, nopayout_A, nopayout_B])`; payout signs; device auto-locks; `nopayout_A`'s initial SIGN_PSBT returns `0x5515` → mirror kept. User unlocks, flow retries → all three resubmitted → `stagePsbt` throws `INVALID_PARAMS: this PSBT was already signed under the loaded intent`. I confirmed that code is `ERROR_CODES.INVALID_PARAMS`, **not** a `DEVICE_*` code — so `isDeviceRecoverableError` returns false and this PR's own new in-modal Retry is not offered. Pre-PR the same retry produced `DEVICE_CEREMONY_INVALID`, which *is* recoverable-classified.

Net effect on the batch path: an actionable dead-end becomes an unactionable one. Either teach the batch caller to resume the unsigned remainder, or restrict the keep-intent optimisation to the single-PSBT path where it actually pays off.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
```diff
@@ -792,6 +792,48 @@ describe("LedgerVaultProvider", () => {
...
+        .mockImplementationOnce(async (_send, prepared: { originalPsbtHex: string }) => ({
+          signedPsbtHex: `signed:${prepared.originalPsbtHex}`,
+          yields: [],
+        }))
+        .mockRejectedValueOnce(new LedgerDeviceLockedError(0x5515));
+
+      await expect(provider.signPsbts([PSBT_A, PSBT_B])).rejects.toMatchObject({ code: ERROR_CODES.DEVICE_LOCKED });
+
+      // After unlock: the unsigned element signs under the same intent...
+      await expect(provider.signPsbts([PSBT_B])).resolves.toEqual([`signed:${PSBT_B}`]);
```
gbarkhatov: This assertion is what makes the batch case look covered when it isn't: `signPsbts([PSBT_B])` is a **subset call no production caller ever emits** — `signDepositorGraph` always resubmits the full list. So the test green-lights a recovery path that does not exist, and hides the `INVALID_PARAMS` dead-end described on `provider.ts`.

Per CLAUDE.md's test philosophy ("never write tests for functions that don't exist in production code"), this is a correctness problem with the test, not just a coverage gap. Asserting the real retry shape — `signPsbts([PSBT_A, PSBT_B])` again after the lock — would have surfaced it.

(Separately, +1 to the bot's note that this test asserts two independent contracts and should be split.)

packages/babylon-ledger-vault-signer/src/errors.ts
```diff
@@ -36,14 +36,20 @@ export class LedgerUserRefusedError extends Error {
...
 
-/** The device is locked; the request never reached the app. */
+/** The device is locked; the refused APDU never reached the app. */
 export class LedgerDeviceLockedError extends Error {
   readonly statusWord: number;
+  /**
+   * false: the initial SIGN_PSBT was refused before dispatch (app state untouched).
+   * true: a CONTINUE was refused after earlier rounds ran (caps may be committed).
+   */
+  readonly midCeremony: boolean;
```
gbarkhatov: Two things about the field itself, both folding into the polarity inversion suggested on `provider.ts`:

1. **The default is asserted far more widely than it is proven.** `?? false` means "app state untouched" — the claim that keeps the mirror. But `classifyStatusWord` constructs this error with no options for *every* exchange, including the DERIVE / APPROVE / pubkey ceremony APDUs, so all of those silently assert the guarantee too. Latent today (only `signStaged`'s catch consumes it), but the conservative value belongs on the other side of the boolean.

2. **The class docline now contradicts the field.** "The device is locked; the refused APDU never reached the app" reads as an unconditional guarantee — which is exactly what `midCeremony: true` exists to deny. Make the class line neutral and let the field carry the distinction.

packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
```diff
@@ -150,6 +152,11 @@ export async function runSignPsbtLoop(
         toCollectedYieldRefs(collector.yields),
       );
     }
+    // A lock on a CONTINUE lands after the app ran earlier rounds; the adapter
+    // must not treat it like the pre-dispatch refusal of the initial APDU.
+    if (isLedgerDeviceLockedError(terminal) && lastSentApdu.ins === INS_CONTINUE) {
```
gbarkhatov: Two notes on this block:

**`lastSentApdu.ins === INS_CONTINUE` compares INS across two different CLAs.** `INS_CONTINUE = 0x01` on `CLA_FRAMEWORK = 0xf8`; the initial APDU is `INS_SIGN_PSBT = 0x04` on `CLA_APP = 0xe1`. No collision today, so the code is correct as written — but the predicate you actually mean is "did we ever leave the initial APDU". A future renumber of `INS_CONTINUE` flips a mid-ceremony lock to `midCeremony: false` — the unsafe direction — with no test catching it. `lastSentApdu !== signPsbtApdu`, or a `sentContinue` boolean, says the intended thing and can't drift.

**Rebuilding the error rather than flagging it is a maintenance trap.** I checked and nothing is lost today: `terminal` is freshly constructed by `classifyStatusWord` one frame away, carries no `cause`, its only own field is `statusWord`, the message regenerates byte-identically, and nothing downstream depends on instance identity. But any field added to `LedgerDeviceLockedError` later is silently dropped here. `classifyStatusWord` already receives a context object carrying `ins` — threading the flag through it removes the duplicate construction entirely.

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts
```diff
@@ -289,17 +290,36 @@ describe("terminal status words mid-loop (T6)", () => {
     expect(sent()).toBe(2);
   });
 
-  it("0x5515 on a CONTINUE surfaces as a locked device", async () => {
+  it("a locked status word on a CONTINUE throws with midCeremony true", async () => {
```
gbarkhatov: `SW_DEVICE_LOCKED` is `{0x5515, 0x6982, 0x5303}`, and the claim asserted in the provider comment is about the whole set — but all five new tests across this PR exercise `0x5515` only. An `it.each` over the set costs nothing and pins that the `midCeremony` flag does not depend on which locked word arrived.

services/vault/src/utils/errors/deviceErrors.ts
```diff
@@ -44,3 +44,12 @@ export function isDeviceLockedError(error: unknown): boolean {
 export function isDeviceWrongAppError(error: unknown): boolean {
   return chainCarriesCode(error, DEVICE_WRONG_APP_CODE);
 }
+
+/** Device states the user can fix in place, then retry: unlock, open the app, re-approve. */
+export function isDeviceRecoverableError(error: unknown): boolean {
```
gbarkhatov: **This predicate is both too broad and hand-duplicated.**

**Too broad.** `DEVICE_CEREMONY_INVALID` is not only "the user must re-approve". `toSignFailureWalletError` also maps `isLedgerYieldMismatchError`, `isLedgerSignPsbtIncompleteError` and `isLedgerSignPsbtProtocolError` to that same code. Those are host/device protocol disagreements — not one of "unlock, open the app, re-approve" from this docstring, and not fixable by the user at all. A yield mismatch will now render an in-modal Retry that re-runs the same disagreement.

**Hand-duplicated.** This ORs exactly the three members of the `DEVICE_CODES` array declared at the top of this file, as three separate depth-10 cause walks. A fourth `DEVICE_*` code would join `DEVICE_CODES` and `deviceErrorCodeOfFrame` but silently *not* join this predicate. One chain walk testing membership in `DEVICE_CODES` is cheaper and drift-proof — and if the fix for "too broad" is to exclude the protocol subset, that exclusion wants to be explicit and named, not implied by an OR chain.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1524,9 +1537,19 @@ export function useDepositFlow(
           // your wallet. Click Retry" — misattributed, and naming a button
           // this surface doesn't render. Post-registration cancels get the
           // variant pointing at the resume path — vaults are already on-chain.
+          // Post-registration device trouble (locked / wrong app / approval
+          // dropped) or a self-cancel: the deposit is on-chain, so offer the
+          // in-modal resume.
+          if (
```
gbarkhatov, line 1557: **This resumability predicate is simultaneously too narrow and too broad, and it is a second classifier disagreeing with `mapDepositError`.**

**Too narrow — the most common post-registration failure is excluded, and its copy names the missing button.** `toSignerWalletError` maps `LedgerUserRefusedError` → `CONNECTION_REJECTED`, not a `DEVICE_*` code, and the second disjunct requires `deviceCancelSettledRef.current`, which is set only by the *in-app* Cancel button. So a **Reject pressed on the device** at the Pre-PegIn signature — ETH registration already mined, gas already spent — misses both disjuncts, falls through to `mapDepositError` → *"You rejected the request in your wallet. Click Retry to approve it and continue"* with `onRetry === undefined` → the view renders **Close**. The in-app-cancel variant of the identical underlying state does get Retry.

**Too broad — it overrides buckets `mapDepositError` deliberately ranks higher.** `isDeviceRecoverableError` is an unordered cause walk, while the copy comes from `mapDepositError`'s carefully ordered buckets: `participantKeyDrift` (3c), `depositTermsRejected` (3d, commented "the copy points at support instead of a retry") and `wrongDepositorWallet` (3f) are all resolved *before* any device-code walk (3g'/3h). An error claimed by one of those still gets a Retry button here if any `DEVICE_*` code sits anywhere in its cause chain — a button whose own copy says retrying cannot help.

**Suggested shape:** map once, then key resumability off the *mapped* error identity rather than re-deriving it from a second, differently-ordered predicate. That fixes both directions at once and removes the possibility of the two classifiers drifting apart.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -1524,9 +1537,19 @@ export function useDepositFlow(
           // your wallet. Click Retry" — misattributed, and naming a button
           // this surface doesn't render. Post-registration cancels get the
           // variant pointing at the resume path — vaults are already on-chain.
+          // Post-registration device trouble (locked / wrong app / approval
```
gbarkhatov: Nit — the appended lines leave this comment block misattached. Sentences 1-2 describe the `setError` call two statements below; only the sentence added here describes the `if` it now sits on. It also asserts "naming a button this surface doesn't render" immediately above the code that makes the button render — while, per the comment on line 1543, that claim is still true for the on-device-reject path. So the block is stale and correct in different sentences. Split it: the resume paragraph above the `if`, the cancel-copy paragraph above `setError`.

Also minor: `deviceCancelSettledRef.current && isUserCancellation(err)` and `registeredVaultIds !== null` are each evaluated twice across these two adjacent statements. Two locals (`const selfCanceled = …`, `const registered = …`) remove the duplication and make the shared precondition visible ahead of the triple-nested ternary below.

services/vault/src/copy.ts
```diff
@@ -65,7 +65,7 @@ const DEVICE_CEREMONY_INVALID_BODY =
   "Your signing device no longer holds this deposit's approval. Try again to restart from the device approval screens.";
 const DEVICE_LOCKED_TITLE = "Signing device locked";
 const DEVICE_LOCKED_BODY =
-  "Your signing device is locked. Unlock it with your PIN and try again.";
+  "Your signing device is locked. Unlock it with your PIN and retry.";
```
gbarkhatov: This body is shown for **every** device-locked error, but the Retry it now names only exists post-registration. When the device locks during the PoP signature or the Pre-PegIn build, `resumableVaultIds` is `null` → `onRetry` is `undefined` → the view renders **Close**. The user reads "unlock it with your PIN and retry" and has no Retry.

This is the exact defect class the pre-existing comment a few lines into `useDepositFlow`'s catch documents avoiding ("naming a button this surface doesn't render"), and this PR's own new test — "leaves resumableVaultIds null for a device-locked error before registration" — pins that state without noticing the copy mismatch.

The shared use at `payoutSignatureErrors.deviceLocked` is fine (`ResumeSignContent` does wire `onRetry`), which is presumably what motivated the change — but the deposit surface regressed. Either keep this string button-neutral, or split it so only the surface that renders Retry names it.

services/vault/src/copy.ts
```diff
@@ -901,11 +901,11 @@ export const COPY = {
...
       },
       // A cancel settling after the Ethereum registration is mined: gas is
-      // spent and the vaults await the Pre-PegIn, so point at the resume path
-      // instead of implying nothing happened. No Bitcoin has moved.
+      // spent and the vaults await the Pre-PegIn, so point at the in-modal
+      // Retry instead of implying nothing happened. No Bitcoin has moved.
       signingCanceledAfterRegistration: {
         title: "Signing canceled",
-        body: "You canceled the signature request. No Bitcoin was spent, but your deposit is already registered on Ethereum. Resume it from your dashboard to broadcast, or the registration will expire on its own.",
+        body: "You canceled the signature request. No Bitcoin was spent, but your deposit is already registered on Ethereum. Retry to continue signing, or the registration will expire on its own.",
```
gbarkhatov: This drops the only mention that the deposit is recoverable at all: "Resume it from your dashboard to broadcast" → "Retry to continue signing".

The modal is still freely dismissible (X, backdrop, Escape). A user who closes instead of clicking Retry now has no instruction anywhere that the deposit can still be completed — only that "the registration will expire on its own", which reads as "it's gone". Keep both affordances in the sentence: retry here, or resume from the dashboard later.

services/vault/src/components/simple/DepositSignContent.tsx
```diff
@@ -99,6 +100,14 @@ export function DepositSignContent({
     }
   }, [executeDeposit, onRefetchActivities]);
 
+  // Post-registration device error or cancel: hand off to the polling
+  // continuation (what the dashboard resume does) instead of re-running the flow.
+  const handleResume = useCallback(() => {
```
gbarkhatov, line 106: The handoff doesn't reset any flow state, which leaves two loose ends:

**1. Two notification observers go live at once.** After the switch, `processing` is `false`, so `signingNotifier.setActiveFlow(false)` re-arms the pending-deposit observer — while `useDepositSigningNotification` is still mounted with `started = true` and `currentStep` frozen at `SIGN_PEGIN_BTC`/`BROADCAST_PRE_PEGIN`, a step that *is* in `STEP_NOTIFICATION` and whose effect re-fires on `documentHidden`. Click Retry, tab away, get two "sign and broadcast" desktop notifications. The success path avoids this only because it lands on a step with no notification entry. `started && !continuationVaultIds` closes it.

**2. A narrow case parks on a false "Awaiting Bitcoin confirmation".** Normally fine — the local `addPendingPegin` record makes this handoff synchronous, so the continuation lands on `SIGN_AND_BROADCAST_TO_BITCOIN` → `ResumeBroadcastContent`, which auto-runs the broadcast. But when that write threw (quota / private browsing — caught, warned, flow continues), `getPollingResult` returns `undefined` until the indexer catches up, and the user sits on a "confirming Bitcoin" spinner, error text gone, no action, for a transaction that was never broadcast. It self-heals and no funds are at risk, but the old Close-then-dashboard behaviour at least didn't state something false. Worth withholding Retry when the persistence warning fired.

services/vault/src/components/simple/DepositSignContent.tsx
```diff
@@ -209,6 +218,7 @@ export function DepositSignContent({
         currentVaultIndex={currentVaultIndex}
         perVaultSteps={perVaultSteps}
         onClose={handleClose}
+        onRetry={resumableVaultIds ? handleResume : undefined}
```
gbarkhatov: `resumableVaultIds ? handleResume : undefined` treats `[]` as resumable, and `handleResume`'s own `if (!resumableVaultIds) return;` doesn't cover it either — while the continuation branch requires `continuationVaultIds.length > 0`. Result: a Retry button that does nothing on click, with no feedback. Unreachable today (`registerPeginBatchOnChain` throws on an empty request list), but it fails silently rather than loudly, which CLAUDE.md's no-silent-fallbacks rule is aimed at.

The same dead-button outcome is reachable for real if `flowParams.depositorEthAddress` is `undefined` at click time — the ETH wallet disconnecting between the failure and the Retry. Gating on `resumableVaultIds.length > 0 && depositorEthAddress` here means the button simply isn't offered instead of being offered and inert.

services/vault/src/components/simple/__tests__/DepositSignContent.test.tsx
```diff
@@ -59,12 +66,27 @@ vi.mock("../PostDepositContinuationContent", () => ({
 // DepositProgressView now owns both the pre-sign entry state (started=false,
 // renders the "Sign Transaction" CTA) and the in-flight stepper (started=true).
 // The mock captures `onSign` so the double-click race can be exercised, and
-// renders a stable marker so tests can tell the two states apart.
+// renders a stable marker so tests can tell the two states apart. On an error
+// it mirrors the real view's button choice: Retry when `onRetry` is handed in,
```
gbarkhatov, line 77: This mock re-implements the production decision the PR is adding, so the two new tests below assert the mock's behaviour rather than the view's.

I checked and the mock is *currently* faithful — `canContinueInBackground` sits ahead of `error` in the real ladder but `computeDepositDerivedState` forces it `false` whenever `hasError`, and `showUnlockCta`/`showCancelSigning` are unreachable with `started && error`. What it does **not** mirror is the real `disabled` computation: the file-level mock of `computeDepositDerivedState` hardcodes `canClose: true`, so these tests would pass even if the real button were disabled.

The gap that matters: `DepositProgressView`'s own test file has 58 `it(`s and zero mentioning retry, so the Retry-vs-Close branch has no coverage against the real component anywhere. In the real view the label and the `onClick` are two independent ternaries — they can drift into a button labelled "Close" that runs `handleResume`, and every test in this PR stays green. One test against the real `DepositProgressView` asserting both label and handler would close it.

gbarkhatov (dismissed):
## Review — retry in place after the device locks mid-deposit

Reviewed across four layers (device/protocol, deposit-flow state machine, UI handoff, copy & conventions). The core design is sound and the central risk I went looking for **is not present**: Retry after a Pre-PegIn sign failure does not hang. The continuation lands on `SIGN_AND_BROADCAST_TO_BITCOIN` → `ResumeBroadcastContent`, whose `useRunOnce(handleBroadcast)` re-drives sign + broadcast on mount rather than polling. Approving on that basis — but **finding 1 below is a fail-open on a §7 critical path and should land before merge.**

### Verification performed

| Suite | Result |
| --- | --- |
| `@babylonlabs-io/ledger-vault-signer` | 592 passed, 19 skipped (Speculos e2e) |
| `wallet-connector` | 176 passed — **only after rebuilding the signer `dist/`** |
| vault (touched suites) | 85 passed |
| `pnpm --filter vault lint` | 0 errors, 125 warnings (pre-existing) |

The wallet-connector suite failing against a stale `dist/` is not incidental friction — it is the symptom of finding 1, and it is how that finding was found.

### Inline findings

1 blocker (fail-open `midCeremony` polarity), several should-fixes on the batch keep-intent path, the resumability predicate, and copy naming buttons that aren't rendered; the rest are nits. All are inline.

### Not inline — files outside the diff

**A. `depositFlowSteps/payoutSigning.ts` now carries a false premise.** It passes `requireFreshDeviceCeremony: true` under the comment *"A signing failure idles the device mirror; the retry must re-derive or approval wallets strand at 'no approved intent' forever."* This PR makes that premise false for a pre-dispatch lock — the mirror is now deliberately kept. The comment is stale, and the flag forces exactly the re-derivation this PR exists to avoid on that path. Worth reconciling in this PR since it changes the invariant the comment depends on.

**B. Question — does Retry auto-firing collide with the post-abort window?** `ResumeBroadcastContent` runs the broadcast on mount, and this provider documents (twice) that a concurrent APDU in that window is eaten with `0x6A80` and desyncs the interrupt loop. A user clicking Retry promptly after a device-side failure is a plausible way into it. I could not settle this from the source alone — it likely wants a hardware check.

**C. Question — how does the release job resolve `workspace:*` for the signer?** `@babylonlabs-io/ledger-vault-signer` is a runtime dependency of the published `wallet-connector`. npm carries signer `0.6.3` while the repo says `0.5.0`, and this PR bumps nothing (no `.changeset/` in the repo). If the pipeline can publish wallet-connector without republishing the signer, **finding 1's fail-open reaches external consumers**, not just local stale builds.

### Checked and clean — worth recording

- `resumableVaultIds` can never be set after the BTC broadcast: every post-broadcast device call sits in a per-vault catch that converts to a warning and never reaches the outer catch. The "spinner for an already-broadcast tx" hazard doesn't exist.
- No window found where a stale intent is *trusted into a signature*. A real intent loss (power cycle, app exit, passphrase re-unlock) returns `0xB007` → `DEVICE_CEREMONY_INVALID` + mirror reset; a session-killing reconnect is caught earlier by `probeSessionAlive` → generation bump. Fails closed.
- Only two `LedgerDeviceLockedError` construction sites exist, and derive/approve locks never reach `signStaged`'s catch — so no ceremony lock is mis-flagged as pre-dispatch.
- Per-PSBT `runSignPsbtLoop` isolation means `midCeremony: false` on the 2nd..Nth element's initial APDU is genuinely correct *at the classification level*; the problem is caller-side.
- `vaultId` is non-optional `Hex` computed locally as `keccak256`, not read from receipt logs — the `.map` cannot yield `undefined`; no runtime assert warranted.
- State lifecycle is sound: abort/unmount cannot set `resumableVaultIds`, and a successful run leaves it at the run-start `null`, so no stale Retry survives into a later run.
- Vault-id identity is consistent across `resumableVaultIds`, the success path and the localStorage record — all from `batchRegistration.vaults[].vaultId`, so no lookup skew.
- `useRunOnce` on the resume broadcast is wallet-gated, so Retry cannot double-broadcast.
- `onRetry` already existed on the props; the button label comes from `COPY`; no `data-testid` was dropped (the E2E CLI matches by role+text); the deep `deviceErrors` import is required and matches the neighbouring `userCancellation` one.
- No key material or payload bytes added to any log line or error field.

Already covered by the bots and not repeated here: the `wallet as any` cast in the flow test, and splitting the batch provider test.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: not planning — payout retry re-derives first, fingerprints reset

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts

packages/babylon-ledger-vault-signer/src/errors.ts
jrwbabylonlab: fixed; fixed

packages/babylon-ledger-vault-signer/src/signPsbtLoop.ts
jrwbabylonlab: fixed; fixed

packages/babylon-ledger-vault-signer/src/__tests__/signPsbtLoop.test.ts

services/vault/src/utils/errors/deviceErrors.ts
jrwbabylonlab: not planning — dashboard resume already retries these; fixed

services/vault/src/hooks/deposit/useDepositFlow.ts
jrwbabylonlab: fixed

services/vault/src/hooks/deposit/useDepositFlow.ts

services/vault/src/copy.ts

services/vault/src/copy.ts

services/vault/src/components/simple/DepositSignContent.tsx
jrwbabylonlab: fixed; not planning — narrow, self-heals once indexer catches up

services/vault/src/components/simple/DepositSignContent.tsx
jrwbabylonlab: fixed

services/vault/src/components/simple/__tests__/DepositSignContent.test.tsx
jrwbabylonlab: fixed

jonybur (approved):
The blockers from the prior round are resolved. Polarity is now preDispatch (fail-closed; default false), so an older signer build that omits the options falls into the pessimistic mirror reset. isResumableDepositError keys off mapped-error identity via the COPY.deposit.errors.* refs rather than string matching, which the useDepositFlow catch reads. The batch keep-intent test uses the production shape signPsbts([PSBT_A, PSBT_B]), with a companion 're-derive, re-approve, resubmit the full batch' test that pins the payout-retry contract. canResume gates on both resumableVaultIds.length > 0 and depositorEthAddress !== undefined, so no inert-button case survives. Copy names Retry only where it renders and keeps the dashboard resume mention. Notification-observer double-fire fixed with the started && !continuationVaultIds gate. CI green.

PR #2328, opened 2026-08-26
fix(ts-sdk, vault): break the ts-sdk ↔ ledger-vault-signer devDepende…

jrwbabylonlab:
#2316 made the signer test-depend on ts-sdk and #2323 the reverse, so nx
refuses every task graph on main. Drop ts-sdk's edge and move the Ledger
payout contract test into the vault app, which may depend on both.

### DIFF
--- a/services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
+++ b/services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
@@ -1,0 +1,252 @@
+// @vitest-environment node
+// The Ledger payout contract has two halves — the SDK's Payout PSBT plus the
+// sign options PayoutManager passes, and the signer's expected-signature table.
+// No package may depend on both without an nx cycle (ts-sdk <-> signer), so the
+// vault, which already depends on both, is where the halves meet.
+
+import {
+  createPayoutConnector,
+  getAssertPayoutScriptInfo,
+  initWasm,
+} from "@babylonlabs-io/babylon-tbv-rust-wasm";
+import {
+  prepareSignPsbt,
+  type ExpectedSignatureTable,
+} from "@babylonlabs-io/ledger-vault-signer";
+import type { BitcoinWallet } from "@babylonlabs-io/ts-sdk/shared";
+import {
+  PayoutManager,
+  type SignPayoutParams,
+} from "@babylonlabs-io/ts-sdk/tbv/core/managers";
+import { deriveBip86ScriptPubKeyHex } from "@babylonlabs-io/ts-sdk/tbv/core/primitives";
+import * as ecc from "@bitcoin-js/tiny-secp256k1-asmjs";
+import { initEccLib, payments, Transaction } from "bitcoinjs-lib";
+import {
+  rootHashFromPath,
+  tapleafHash,
+} from "bitcoinjs-lib/src/payments/bip341";
+import { Buffer } from "buffer";
+import { beforeAll, describe, expect, it } from "vitest";
+
+// Deterministic x-only keys (scalar * G), the same set the SDK's PSBT tests use.
+const DEPOSITOR =
+  "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798";
+const VAULT_PROVIDER =
+  "c6047f9441ed7d6d3045406e95c07cd85c778e4b8cef3ca7abac09b95c709ee5";
+const VAULT_KEEPER =
+  "f9308a019258c31049344f85f89d5229b531c845836f99b08601f113bce036f9";
+const UNIVERSAL_CHALLENGER =
+  "2f8bde4d1a07209355b4a7250a5c5128e88b84bddc619ab7cba8d569b240efe4";
+
+const VAULT_CORE_VERSION = 1;
+const TIMELOCK_PEGIN = 100;
+const TIMELOCK_ASSERT = 144;
+const COUNCIL_QUORUM = 2;
+const COMMISSION_BPS = 500;
+const PROTOCOL_FEE_RATE = 10n;
+
+const PEGIN_VALUE_SATS = 100_000;
+const ASSERT_VALUE_SATS = 50_000;
+const VP_COMMISSION_SATS = 1_000;
+const CPFP_ANCHOR_SATS = 546;
+const PAYOUT_FEE_SATS = 5_000;
+const DEPOSITOR_PAYOUT_SATS =
+  PEGIN_VALUE_SATS +
+  ASSERT_VALUE_SATS -
+  PAYOUT_FEE_SATS -
+  VP_COMMISSION_SATS -
+  CPFP_ANCHOR_SATS;
+
+const PAYOUT_TX_VERSION = 2;
+const SEQUENCE_MAX = 0xffffffff;
+const TAPSCRIPT_LEAF_VERSION = 0xc0;
+
+// P2WPKH: OP_0 PUSH_20 <hash>; the fill byte only makes the outputs distinct.
+function dummyP2wpkh(fillNibble: string): Buffer {
+  return Buffer.from(`0014${fillNibble.repeat(40)}`, "hex");
+}
+
+/** Sorted x-only keys for scalars offset..offset+count-1 — a valid council. */
+function councilKeys(count: number, offset: number): string[] {
+  const keys: string[] = [];
+  for (let i = 0; i < count; i++) {
+    const scalar = Buffer.alloc(32);
+    scalar.writeUInt32BE(offset + i, 28);
+    const point = ecc.pointFromScalar(scalar, true);
+    if (!point) throw new Error(`invalid scalar ${offset + i}`);
+    keys.push(Buffer.from(point.subarray(1, 33)).toString("hex"));
+  }
+  return keys.sort();
+}
+
+/** BIP-341 fold of a (leaf, control block) pair to the P2TR output it spends. */
+function taprootOutputFor(scriptHex: string, controlBlockHex: string): Buffer {
+  const controlBlock = Buffer.from(controlBlockHex, "hex");
+  const leafHash = tapleafHash({
+    output: Buffer.from(scriptHex, "hex"),
+    version: TAPSCRIPT_LEAF_VERSION,
+  });
+  const output = payments.p2tr({
+    internalPubkey: controlBlock.subarray(1, 33),
+    hash: rootHashFromPath(controlBlock, leafHash),
+  }).output;
+  if (!output) throw new Error("p2tr produced no output script");
+  return output;
+}
+
+const STOP_AFTER_PREPARE = new Error("stop: table captured");
+
+/**
+ * A wallet that does what LedgerVaultProvider does with the SDK's options —
+ * hand the requested input indices to prepareSignPsbt — then bails out, so the
+ * manager never reaches signature extraction.
+ */
+function ledgerTableCapturingWallet(
+  tables: ExpectedSignatureTable[],
+): BitcoinWallet {
+  const notExercised = () => {
+    throw new Error("not exercised by this test");
+  };
+  return {
+    getPublicKeyHex: async () => DEPOSITOR,
+    signPsbt: async (psbtHex, options) => {
+      // Mirrors LedgerVaultProvider.stagePsbt: only the indices are honoured.
+      const signInputIndexes = options?.signInputs?.map((input) => input.index);
+      const { table } = prepareSignPsbt({
+        psbtHex,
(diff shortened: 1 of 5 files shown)


services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
```diff
@@ -0,0 +1,252 @@
+// @vitest-environment node
+// The Ledger payout contract has two halves — the SDK's Payout PSBT plus the
+// sign options PayoutManager passes, and the signer's expected-signature table.
+// No package may depend on both without an nx cycle (ts-sdk <-> signer), so the
+// vault, which already depends on both, is where the halves meet.
```
gbarkhatov, line 5: **"No package may depend on both" is not quite true.**

`packages/babylon-wallet-connector/package.json:60` already depends on the signer and does not depend on ts-sdk, so adding ts-sdk there as a devDependency is cycle-free — and wallet-connector is where `LedgerVaultProvider.stagePsbt`, the thing `ledgerTableCapturingWallet` mirrors, actually lives.

The vault is a defensible home; it is the real composition root and already had both halves in its runtime graph. But the header asserts an impossibility that is not one. Worth rewording to say why the vault was chosen, and naming the cost that came with it: as an app-level test this contract stops running for anyone consuming the packages standalone, and a vault-only CI run now goes red for an SDK-side bug.

Related and minor: every other file in `src/services/vault/__tests__/` is `<module>.test.ts` for a sibling in `src/services/vault/` — `vaultActivationService.test.ts`, `vaultTransactionService.test.ts`, and so on. This one has no counterpart there. Somewhere like `src/__tests__/contracts/` would make the cross-package purpose legible from the path alone.

services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
```diff
@@ -0,0 +1,252 @@
...
+import {
+  prepareSignPsbt,
+  type ExpectedSignatureTable,
+} from "@babylonlabs-io/ledger-vault-signer";
+import type { BitcoinWallet } from "@babylonlabs-io/ts-sdk/shared";
+import {
+  PayoutManager,
+  type SignPayoutParams,
+} from "@babylonlabs-io/ts-sdk/tbv/core/managers";
+import { deriveBip86ScriptPubKeyHex } from "@babylonlabs-io/ts-sdk/tbv/core/primitives";
```
gbarkhatov, line 21: **The relocated guard resolves through `dist/`, so it cannot fail on a source regression.**

These imports go by package name. `vitest.config.ts` hands `vite-tsconfig-paths` only the vault's and wallet-connector's `tsconfig.lib.json`, and neither maps `@babylonlabs-io/ts-sdk` or `@babylonlabs-io/ledger-vault-signer` — `services/vault/tsconfig.lib.json:23-25` maps `@/*` alone — so resolution goes symlink → `exports` → `./dist`. The deleted block imported `../payout` and `../constants` and could not go stale.

Concretely: set `DEPOSITOR_SIGNED_INPUT_COUNT` back to 2 in ts-sdk source, run the vault suite without rebuilding, and both cases still pass. That is the exact regression class this file exists to catch.

CI is safe, but only incidentally — `verify.yml` runs `pnpm build` before `nx affected -t test`, and `nx.json` gives `test` a `dependsOn: ["^build"]`. Nothing in the header says the guard depends on that. Either fix works:

- add `resolve.alias` entries pointing the two packages at their `src/`, which restores the property outright; or
- state the dependency in the header and name the invocation. Worth noting that `services/vault/package.json:19` is `"test": "vitest"` (watch mode), unlike ts-sdk's and the signer's `"vitest run"` — so `pnpm --filter vault test` never runs to completion locally. `test:run` is the one to document.

services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
```diff
@@ -0,0 +1,252 @@
...
+const ASSERT_VALUE_SATS = 50_000;
+const VP_COMMISSION_SATS = 1_000;
+const CPFP_ANCHOR_SATS = 546;
+const PAYOUT_FEE_SATS = 5_000;
+const DEPOSITOR_PAYOUT_SATS =
+  PEGIN_VALUE_SATS +
+  ASSERT_VALUE_SATS -
+  PAYOUT_FEE_SATS -
+  VP_COMMISSION_SATS -
+  CPFP_ANCHOR_SATS;
```
gbarkhatov, line 58: nit: the fee band these constants have to sit inside is not stated anywhere.

They imply a 5,000-sat fee, which `assertPayoutFeeInBand` bounds at `protocolFeeRate × (500 + 55 × numParticipants)` = `10 × 610` = 6,100 (`assertPayoutFeeBand.ts:46-47`, `:141-148`), over a WASM-computed floor of 4,080 for this shape. Comfortably in band — but nothing here says so, so the next change to the payout vsize model or to `PROTOCOL_FEE_RATE` will surface as a confusing fee failure in a test that is nominally about signature expectations. One line naming floor / fee / ceiling keeps that legible.

services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    output: Buffer.from(scriptHex, "hex"),
+    version: TAPSCRIPT_LEAF_VERSION,
+  });
+  const output = payments.p2tr({
+    internalPubkey: controlBlock.subarray(1, 33),
+    hash: rootHashFromPath(controlBlock, leafHash),
+  }).output;
+  if (!output) throw new Error("p2tr produced no output script");
+  return output;
+}
```
gbarkhatov, line 95: **`taprootOutputFor` reimplements something the SDK already owns.**

`computeTaprootScriptPubKey` (`packages/babylon-ts-sdk/src/tbv/core/primitives/utils/taproot.ts:103`) does exactly this fold, and it is what `buildPayoutPsbt` itself uses to bind the Assert leaf (`primitives/psbt/payout.ts:412`). It is not reachable from here — `primitives/index.ts:77` exports only `computeNumLocalChallengers` from that area — so this file hand-rolls it.

Exporting it from the `./tbv/core/primitives` barrel would delete this helper and the `bitcoinjs-lib/src/payments/bip341` import at lines 24-27 together. That import is worth losing on its own: it is extension-less into library internals and resolves only because Vite appends `.js` — plain Node ESM throws `ERR_MODULE_NOT_FOUND` on it. Nothing gates it (bitcoinjs-lib 6.1.7 ships no `exports` map) and there is precedent at `packages/babylon-ledger-vault-signer/src/__tests__/tapLeafHash.test.ts:8`, so it is not a blocker.

If the second implementation is deliberate — an independent oracle, rather than the SDK checking its own work in a critical-path test — that is a good reason, but say so here. As written a reader cannot tell whether it is a choice or a workaround.

services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    });
+
+    await expect(manager.signPayoutTransaction(params)).rejects.toBe(
+      STOP_AFTER_PREPARE,
+    );
+
+    expect(tables).toHaveLength(1);
+    expect([...tables[0].byInput.keys()]).toEqual([0]);
+    expect(tables[0].expectedYieldCount).toBe(1);
+  });
```
gbarkhatov, line 233: **The test drives the branch a Ledger never takes.**

Production Ledger presign goes through the batch branch: `runDepositorPresignFlow.ts:376` picks `signPayoutTransactionsBatch` whenever `supportsBatchSigning()` is true, and it is true here, because `LedgerVaultProvider` implements `signPsbts` (`packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts:684`). This test calls `signPayoutTransaction` and stubs `signPsbts: notExercised` — so the mock advertises batch support while the test exercises the sequential branch a Ledger never reaches.

Two reasons that matters:

- The sign options are constructed at two separate call sites, `PayoutManager.ts:269-272` and `:377-380`. A divergence at the batch site is exactly the #2321 shape, and this test cannot see it.
- The batch path has a failure mode the single path structurally cannot. `provider.ts:700` reads `options?.[index]`, so any positional mismatch between `psbtsToSign` and `signOptions` silently yields `signInputIndexes === undefined` and drops the narrowing #2321 added. `signPayoutTransactionsBatch` builds the two arrays 1:1 in a single loop today, so there is no live bug — but nothing pins that.

Suggest driving `signPayoutTransactionsBatch([params, params])` with a `signPsbts` stub that mirrors `provider.ts:698-703` (per-element `stagePsbt`, options taken as `options?.[index]`). Two elements is what actually exercises the positional contract; one would not.

gbarkhatov (approved):
Approving — the fix is correct, minimal, and points the right way.

Verified independently before approving: the cycle is real on `67b75dc41` (ts-sdk `package.json:123` ⇄ signer `package.json:48`); dropping ts-sdk's edge is the right direction, since it backed one import in one test file while the signer's edge backs the Speculos e2e fixture that drives `buildPayoutPsbt`/`buildNoPayoutPsbt` against real firmware — degrading that to pinned vectors to save a unit test would be a bad trade; `@services/vault` is a graph sink, so no new cycle; `payout.test.ts` has no orphaned imports, helpers or constants left after the deletion (checked every symbol the removed block referenced); ts-sdk retains no reference to the signer outside two prose comments, and no tsconfig/vitest/eslint/nx config referenced it; the lockfile diff is exactly the two `package.json` edits with no drift; and devDependency is the correct section for both additions — neither package is imported from production `src/`. Both assertions survive the move, and routing through `PayoutManager` covers production wiring the old test stubbed out.

Five inline comments below. Two of them (the batch entry point, and `dist/` resolution) are about the relocated guard being weaker than the one it replaces, which seems worth settling before this becomes the only real-builder coverage. The rest are follow-ups.

One finding that has no line in this diff, plus three notes:

**`packages/babylon-ledger-vault-signer/src/vendor/ledger-bitcoin/README.md:217` is now a dangling reference.** It reads "the post-#2281 contract is covered against the real builder by `babylon-ts-sdk` `payout.test.ts` (\"Ledger expected-signature contract\")" — that describe block no longer exists. This bullet is the provenance note a future fixture bumper reads before touching the pin, right after the sentence explaining that `expectedSignatures.test.ts`'s "input 0 tapscript, other inputs absent" case reads on the pre-#2281 pin. A stale pointer there has real cost under the section 7 rules. Repoint it at `services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts`.

**PR body wording.** "nx refuses every task graph" is a touch strong — `build` and `test` carry `dependsOn: ["^build"]` (`nx.json:24-27`) and do cycle, but `lint` has no `^` dependency and still resolves. Cosmetic, though the body is the durable record of why the devDep was dropped.

**The Greptile and CodeRabbit comments on `workspace:*` are false positives.** That is the repo-wide convention for internal deps — the very devDependency this PR removes from `packages/babylon-ts-sdk/package.json` was `workspace:*`, as is every other internal edge. No action needed.

**On preventing a recurrence.** `verify.yml` already builds everything on every PR, so it would have caught this had either #2316 or #2323 carried both edges — neither did, and each was locally green. No extra CI step fixes that shape; "require branches up to date before merging" or a merge queue does.

services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
jrwbabylonlab: fixed in #2329

services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
jrwbabylonlab: fixed in #2329

services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
jrwbabylonlab: fixed in #2329

services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
jrwbabylonlab: fixed in #2329

services/vault/src/services/vault/__tests__/ledgerPayoutSigningContract.test.ts
jrwbabylonlab: fixed in #2329

PR #2329, opened 2026-08-26
test(vault): resolve Ledger payout contract test from src and cover t…

jrwbabylonlab:
fix PR 2328 review comments

### DIFF
--- a/services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
+++ b/services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
@@ -1,4 +1,3 @@
-// @vitest-environment node
 // Ledger payout signing contract: the SDK's Payout PSBT + sign options
 // (ts-sdk) meet the signer's expected-signature table. wallet-connector
 // (signer, no ts-sdk) could host this cycle-free too; the vault is chosen as
@@ -6,9 +5,10 @@
 // to the LedgerVaultProvider the stubs below mirror. Cost: standalone package
 // consumers never run it, and a vault-only CI run goes red for an SDK bug.
 //
-// The `contracts` project in services/vault/vitest.config.ts aliases ts-sdk
-// and the signer to their src/, so a source regression fails this file without
-// a rebuild; only the WASM package is consumed built. Run it with
+// The `contracts` project in services/vault/vitest.config.ts resolves ts-sdk
+// and signer runtime modules from src/, so a source regression fails this file
+// without a rebuild; types still come from dist d.ts (tsc does not see Vite
+// aliases), and the WASM package is consumed built. Run it with
 // `pnpm --filter vault test:run` (`test` alone is vitest watch mode).
 
 import {
@@ -59,8 +59,13 @@ const PEGIN_VALUE_SATS = 100_000;
 const ASSERT_VALUE_SATS = 50_000;
 const VP_COMMISSION_SATS = 1_000;
 const CPFP_ANCHOR_SATS = 546;
-// Fee band for this shape (assertPayoutFeeBand.ts:141-148,160): floor 4,080
-// (WASM computePayoutFeeFloor, out1Len 22) <= 5,000 <= 10 x (500 + 55 x 2) = 6,100.
+// Both shapes below imply a 5,000-sat fee inside assertPayoutFeeInBand's band.
+// Floor = computePayoutFeeFloor(vaultCoreVersion 1, numVaultKeepers 1,
+// numUniversalChallengers 1, numLocalChallengers 1, councilSize 3, out0Len,
+// out1Len, protocolFeeRate 10n): VP-claimer out0Len 22 / out1Len 22 = 4,080;
+// VK-claimer out0Len 34 (P2TR) / out1Len undefined = 3,890. Ceiling = 10 x
+// (MAX_PAYOUT_VSIZE_BASE 500 + 55 x 2 participants + scriptExcess 0, since
+// neither out0Len exceeds PAYOUT_BOUND_ASSUMED_SCRIPT_LEN 34) = 6,100.
 const PAYOUT_FEE_SATS = 5_000;
 const DEPOSITOR_PAYOUT_SATS =
   PEGIN_VALUE_SATS +
@@ -68,10 +73,14 @@ const DEPOSITOR_PAYOUT_SATS =
   PAYOUT_FEE_SATS -
   VP_COMMISSION_SATS -
   CPFP_ANCHOR_SATS;
+// VK-claimer layout is [payout, anchor]: no commission output.
+const VK_CLAIMER_PAYOUT_SATS =
+  PEGIN_VALUE_SATS + ASSERT_VALUE_SATS - PAYOUT_FEE_SATS - CPFP_ANCHOR_SATS;
 
 const PAYOUT_TX_VERSION = 2;
 const SEQUENCE_MAX = 0xffffffff;
 const TAPSCRIPT_LEAF_VERSION = 0xc0;
+const HEX_PREFIX = "0x";
 
 // P2WPKH: OP_0 PUSH_20 <hash>; the fill byte only makes the outputs distinct.
 function dummyP2wpkh(fillNibble: string): Buffer {
@@ -149,12 +158,16 @@ function ledgerSequentialWallet(
   };
 }
 
-/** signPsbts mirrors only LedgerVaultProvider.signPsbts' index -> options[k] mapping (not its intra-batch duplicate guard), then bails. */
-function ledgerBatchWallet(tables: ExpectedSignatureTable[]): BitcoinWallet {
+/** signPsbts mirrors only the index -> options[k] pairing; every other provider gate is out of scope. */
+function ledgerBatchWallet(
+  tables: ExpectedSignatureTable[],
+  psbtHexes: string[],
+): BitcoinWallet {
   return {
     getPublicKeyHex: async () => DEPOSITOR,
     signPsbt: notExercised,
     signPsbts: async (psbtsHexes, options) => {
+      psbtHexes.push(...psbtsHexes);
       for (const [index, hex] of psbtsHexes.entries()) {
         captureTable(hex, options?.[index], tables);
       }
@@ -169,12 +182,15 @@ function ledgerBatchWallet(tables: ExpectedSignatureTable[]): BitcoinWallet {
 
 describe("Ledger payout signing contract", () => {
   let params: SignPayoutParams;
+  let vkClaimerParams: SignPayoutParams;
 
   beforeAll(async () => {
     initEccLib(ecc);
     await initWasm();
 
     const councilMembers = councilKeys(3, 9_000);
+    // `0x`-prefixed by contract (bitcoin.ts JSDoc); the raw script strips it.
+    const vkPayoutScriptHex = deriveBip86ScriptPubKeyHex(VAULT_KEEPER);
 
     // PegIn:0 is the real vault output for these participants.
     const vaultOutput = await createPayoutConnector(
@@ -248,10 +264,52 @@ describe("Ledger payout signing contract", () => {
       protocolFeeRate: PROTOCOL_FEE_RATE,
       councilMembers,
       councilQuorum: COUNCIL_QUORUM,
-      vkClaimerPayoutScriptPubKeys: {
-        [VAULT_KEEPER]: deriveBip86ScriptPubKeyHex(VAULT_KEEPER),
-      },
+      vkClaimerPayoutScriptPubKeys: { [VAULT_KEEPER]: vkPayoutScriptHex },
       vpCommissionScriptPubKey: dummyP2wpkh("e").toString("hex"),
+    };
+
+    // Batch element 1: a VK claimer. Its Assert:0 leaf (local challengers
+    // ({VP} ∪ VKs) − {VK}) and its 2-output payout layout both differ.
+    const vkAssertLeaf = await getAssertPayoutScriptInfo({
+      txGraphVersion: VAULT_CORE_VERSION,
+      claimer: VAULT_KEEPER,
+      localChallengers: [VAULT_PROVIDER],
+      universalChallengers: [UNIVERSAL_CHALLENGER],
+      timelockAssert: TIMELOCK_ASSERT,
+      councilMembers,
+      councilQuorum: COUNCIL_QUORUM,
+    });
+    const vkAssertTx = new Transaction();
+    vkAssertTx.addInput(Buffer.alloc(32, 3), 0xffffffff, SEQUENCE_MAX);
+    vkAssertTx.addOutput(
+      taprootOutputFor(
+        vkAssertLeaf.payoutScript,
+        vkAssertLeaf.payoutControlBlock,
+      ),
+      ASSERT_VALUE_SATS,
+    );
+    const vkPayoutTx = new Transaction();
+    vkPayoutTx.version = PAYOUT_TX_VERSION;
+    vkPayoutTx.addInput(
+      Buffer.from(peginTx.getId(), "hex").reverse(),
+      0,
+      TIMELOCK_PEGIN,
+    );
+    vkPayoutTx.addInput(
+      Buffer.from(vkAssertTx.getId(), "hex").reverse(),
+      0,
+      TIMELOCK_ASSERT,
+    );
+    vkPayoutTx.addOutput(
+      Buffer.from(vkPayoutScriptHex.slice(HEX_PREFIX.length), "hex"),
+      VK_CLAIMER_PAYOUT_SATS,
+    );
+    vkPayoutTx.addOutput(dummyP2wpkh("c"), CPFP_ANCHOR_SATS);
+    vkClaimerParams = {
+      ...params,
+      payoutTxHex: vkPayoutTx.toHex(),
+      assertTxHex: vkAssertTx.toHex(),
+      claimerBtcPubkey: VAULT_KEEPER,
     };
   });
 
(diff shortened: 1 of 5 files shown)


jonybur (dismissed):
Contract test move looks right. The signPsbts variant closes the batch-path coverage gap for LedgerVaultProvider — element k must be staged with options[k] is the invariant that would have caught a per-index misalignment in runDepositorPresignFlow. The BitcoinWallet stubs now implement every method (with notExercised guards) rather than casting, so a future BitcoinWallet method addition trips a compile error here too. The vitest contracts project aliases ts-sdk and the signer to src, which matches the file's docblock claim that a source regression fails it without a rebuild. Fee-band comment on PAYOUT_FEE_SATS names the exact source expressions in assertPayoutFeeBand.ts, so the number stays traceable.

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
```diff
@@ -249,4 +288,29 @@ describe("Ledger payout signing contract", () => {
...
+  // The production presign path (runDepositorPresignFlow) takes this branch
+  // whenever the wallet has signPsbts, which the Ledger provider does.
+  it("PayoutManager's batch path asks the Ledger signer for input 0 alone on every real Payout PSBT", async () => {
+    const tables: ExpectedSignatureTable[] = [];
+    const manager = new PayoutManager({
+      network: "signet",
+      btcWallet: ledgerBatchWallet(tables),
+    });
+
+    // Two elements on purpose: element k must be staged with options[k].
```
gbarkhatov: This fixture does not pin the invariant the comment claims, and it is a batch the mirrored provider rejects.

The positional claim is unobservable through this seam. `PayoutManager.ts:376-381` pushes `createTaprootScriptPathSignOptions(walletPubkeyRaw, DEPOSITOR_SIGNED_INPUT_COUNT)` per element, and both arguments are loop-invariant — `walletPubkeyRaw` is fetched once at `:337`, and the count is a module constant. So every `signOptions[k]` is content-identical for *any* batch: passing `options[0]` to every PSBT, or reversing the array, passes every assertion below. With `[params, params]` the two PSBTs are byte-identical as well, so neither half can vary. That pairing is in fact already pinned where it *can* be — `provider.test.ts:863-868` drives `signPsbts([PSBT_A, PSBT_B], [{}, { autoFinalized: true }])` and asserts the rejection names `signPsbts[1]`.

Separately, `[params, params]` is a batch `LedgerVaultProvider.signPsbts` refuses outright: identical PSBTs give the same `signingRequestKey` (`provider.ts:112-125`), and `stagePsbt` throws "this PSBT is duplicated within the batch" (`provider.ts:882-888`, pinned at `provider.test.ts:881`). Production cannot emit it either — `runDepositorPresignFlow.ts:346` sets `claimerBtcPubkey: tx.claimerPubkeyXOnly`, one PSBT per distinct claimer. (An existing inline bot comment flags this half.)

What the test *does* pin is real and worth keeping: options **absent**, or an options array **shorter** than the PSBT array. An undefined entry leaves `requestedInputs` undefined, so both leaf-carrying inputs land in `byInput` (`expectedSignatures.ts:260`, `:329`) and `toEqual([0])` fails for that element.

Suggestion: make element 1 a distinct claimer — `claimerBtcPubkey: VAULT_KEEPER`, whose payout scriptPubKey is already in `params.vkClaimerPayoutScriptPubKeys` (lines 251-253). That survives the duplicate guard, matches what production actually batches, and makes the two captured tables genuinely distinguishable. Then reword line 301 to what is pinned.

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
```diff
@@ -249,4 +288,29 @@ describe("Ledger payout signing contract", () => {
       expect(expectation.kind).toBe("tapscript");
     }
   });
+
+  // The production presign path (runDepositorPresignFlow) takes this branch
+  // whenever the wallet has signPsbts, which the Ledger provider does.
+  it("PayoutManager's batch path asks the Ledger signer for input 0 alone on every real Payout PSBT", async () => {
```
gbarkhatov: Nit: this bundles the three behaviours the sequential pair deliberately splits — index narrowing (`byInput`), `expectedYieldCount`, and input-1 tapscript classification — while the name claims only the first. `CLAUDE.md` asks for one behaviour per test, with the name describing it. Trimming this to the batch-specific delta, or splitting it the way lines 258-290 are split, would keep the pair symmetric.

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
```diff
@@ -96,32 +111,56 @@ function taprootOutputFor(scriptHex: string, controlBlockHex: string): Buffer {
...
-    getAddress: notExercised,
     signPsbts: notExercised,
+    getAddress: notExercised,
+    signMessage: notExercised,
+    getNetwork: notExercised,
+    deriveContextHash: notExercised,
+  };
+}
+
+/** signPsbts mirrors only LedgerVaultProvider.signPsbts' index -> options[k] mapping (not its intra-batch duplicate guard), then bails. */
```
gbarkhatov: The doc names the duplicate guard as *the* divergence, which reads as if everything else is mirrored. It also omits the empty-batch rejection (`provider.ts:688-694`), the `autoFinalized` rejection (`:808-816`), the hex-shape gate (`:818-824`), the mixed key-path/tapscript rejection (`:846-853`), and the entire key-path policy-augmentation branch (`:854-877`). The positive form is shorter and does not have to stay in sync: mirrors only the index to `options[k]` pairing, every other provider gate out of scope.

Also worth knowing: `ledgerSequentialWallet` sets `signPsbts: notExercised` (line 144), so `supportsBatchSigning()` returns `true` for it. Harmless here, since both sequential tests call `signPayoutTransaction` directly — but it means this fixture cannot be reused to cover the sequential fallback branch of `runDepositorPresignFlow`.

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
```diff
@@ -49,6 +59,8 @@ const PEGIN_VALUE_SATS = 100_000;
 const ASSERT_VALUE_SATS = 50_000;
 const VP_COMMISSION_SATS = 1_000;
 const CPFP_ANCHOR_SATS = 546;
+// Fee band for this shape (assertPayoutFeeBand.ts:141-148,160): floor 4,080
+// (WASM computePayoutFeeFloor, out1Len 22) <= 5,000 <= 10 x (500 + 55 x 2) = 6,100.
```
gbarkhatov: The band checks out — I ran the WASM, and `computePayoutFeeFloor(1, 1, 1, 1, 3, 22, 22, 10n)` is exactly `4080n`; the ceiling is `10 x (500 + 55x2) = 6100`, with `scriptExcess = 0` since `out0Len 22 < PAYOUT_BOUND_ASSUMED_SCRIPT_LEN 34`.

But the floor is not reproducible from what is written. The parenthetical names only `out1Len 22`, while the value also depends on `out0Len`, 1 keeper, 1 universal challenger, 3 council members and rate 10 (`assertPayoutFeeBand.ts:160-169`) — drop `out1Len` alone and you get 3,770, so the reader cannot arrive at 4,080. The two lines also mix `out1Len` (a floor input) with `out0Len` (what drives the ceiling's `scriptExcess`) without saying so, and `:141-148` is off by one at the start — the ceiling block's explanatory comment opens at `:140`. Naming the full argument tuple would make the number checkable, which is presumably the point of the comment.

services/vault/vitest.config.ts
```diff
@@ -49,6 +53,42 @@ export default defineConfig({
         inline: ["@babylonlabs-io/wallet-connector", "@noble/hashes"],
       },
     },
+    projects: [
+      {
+        extends: true,
+        test: {
+          name: "unit",
+          exclude: [...SHARED_EXCLUDE, `${CONTRACTS_TEST_DIR}/**`],
```
gbarkhatov: Nit: the `...SHARED_EXCLUDE` spread is redundant. With `extends: true`, vitest initializes the project from the root config file *plus* this object as inline config, and Vite's `mergeConfig` concatenates arrays — so the root `exclude` (line 37) is already inherited. Dumping the resolved projects shows `unit` ending up with the four shared patterns twice, then the contracts glob.

Harmless, since glob matching is idempotent. But it encodes replace-semantics that do not hold, and the `SHARED_EXCLUDE` const exists only to serve it. ``exclude: [`${CONTRACTS_TEST_DIR}/**`]`` here is equivalent and lets line 37 stay the single source.

services/vault/vitest.config.ts
```diff
@@ -49,6 +53,42 @@ export default defineConfig({
...
+        },
+      },
+      {
+        extends: true,
+        test: {
+          name: "contracts",
```
gbarkhatov, line 135: The `contracts` project inherits `environment: "jsdom"` and `setupFiles: ["./src/test/setup.ts"]` from the root — I dumped the resolved project to confirm. The one file here runs under node only because it carries its own `// @vitest-environment node` docblock; the next contracts file without that line silently gets jsdom.

`src/test/setup.ts` also still executes (~200ms of the run) and applies four vault-app `vi.mock`s, one of which is `vi.mock("@/utils/btc/wasm")` — it stubs `initWasm`. This test dodges that only because it imports `initWasm` from `@babylonlabs-io/babylon-tbv-rust-wasm` directly. A cross-package contract test that touches no vault module should not depend on the vault's React test harness staying node-safe. Suggest `environment: "node"` and `setupFiles: []` on this project, so the guarantee is structural rather than per-file.

services/vault/vitest.config.ts
```diff
@@ -49,6 +53,42 @@ export default defineConfig({
...
+          include: [`${CONTRACTS_TEST_DIR}/**/*.test.ts`],
```
gbarkhatov: The include and exclude globs are asymmetric, which can orphan a future test silently. `unit` excludes the directory wholesale (`src/__tests__/contracts/**`), but `contracts` includes only `**/*.test.ts`. Checked with picomatch against the resolved configs:

```
src/__tests__/contracts/foo.test.ts    unitExcluded: true   contractsIncluded: true
src/__tests__/contracts/foo.test.tsx   unitExcluded: true   contractsIncluded: false
src/__tests__/contracts/foo.spec.ts    unitExcluded: true   contractsIncluded: false
```

So a `ledgerDepositSigning.spec.ts`, or a `.test.tsx` for a component-level contract, runs in neither project, emits no output, and `contracts` still reports "1 passed". Suggest ``${CONTRACTS_TEST_DIR}/**/*.test.{ts,tsx}``, or mirroring the exclude pattern exactly.

services/vault/vitest.config.ts
```diff
@@ -49,6 +53,42 @@ export default defineConfig({
...
+        },
+        // Cross-package contract tests read ts-sdk and the signer from src/ so a
+        // source regression fails them without a rebuild; WASM stays on dist (Rust-built).
```
gbarkhatov: Two claims here are stronger than what the alias actually does.

**"read ts-sdk and the signer from src/"** — the alias never consults the package `exports` map: `@babylonlabs-io/ts-sdk/nope` resolves happily to `src/nope/index.ts`, which is proof it is not being checked. So a contract test can import a subpath that is not exported — `.../tbv/core/managers/pegin`, say, which exists in `src` — and pass, while every real consumer gets `ERR_PACKAGE_PATH_NOT_EXPORTED`. The contracts project is the one place you would most want the public surface enforced.

**"a source regression fails them without a rebuild"** — true at runtime only. This is a Vite resolve alias; `tsc` does not see it, and `tsconfig.lib.json` carries only `"@/*"` in `paths`. So `import type { BitcoinWallet, SignPsbtOptions } from "@babylonlabs-io/ts-sdk/shared"` is still checked against `dist/shared/index.d.ts`, and CI builds before it tests — a type-level SDK regression remains gated on a rebuild. Worth narrowing the sentence to runtime; the same claim is repeated in the test header at lines 9-11.

services/vault/vitest.config.ts
```diff
@@ -49,6 +53,42 @@ export default defineConfig({
...
+        resolve: {
+          alias: [
+            {
+              find: /^@babylonlabs-io\/ts-sdk\/(.*)$/,
```
gbarkhatov: Two gaps in this alias array.

The regex requires a `/` plus a subpath, so the bare specifier still resolves to `dist` — resolved through the actual plugin container:

```
@babylonlabs-io/ts-sdk                   => packages/babylon-ts-sdk/dist/index.js
@babylonlabs-io/ts-sdk/shared            => packages/babylon-ts-sdk/src/shared/index.ts
@babylonlabs-io/ts-sdk/tbv/core/managers => packages/babylon-ts-sdk/src/tbv/core/managers/index.ts
```

Latent today — the contracts run's module graph is 111 ts-sdk `src` modules and zero from `dist` — but the barrel is the import form vault code uses everywhere, and the SDK's own `src` avoids it only by luck (its 8 self-references are all inside JSDoc `@example` blocks). The first contract test that reaches the barrel loads two copies of the SDK in one run: `instanceof` fails across the boundary, module-level singletons and caches exist twice, and the `dist` half is whatever was last built. It would present as a bizarre logic bug, not a resolution error. An exact-match entry closes it — a separate `` /^@babylonlabs-io\/ts-sdk$/ `` -> `src/index.ts`, rather than widening this regex (an optional group would yield `src//index.ts` for the bare case).

Conversely, the signer entry on line 82 is a string `find`, which Vite treats as *prefix* matching (exact, or `find + "/"`), and its replacement is a file rather than a directory — so `@babylonlabs-io/ledger-vault-signer/<sub>` would rewrite to `.../src/index.ts/<sub>`. Unreachable today, since that package exports only `"."`, and it would fail loudly rather than silently, so this half is just worth knowing.

gbarkhatov (dismissed):
Reviewed the three-file diff plus the surrounding production paths; ran the `contracts` project locally (3 tests, ~0.8s) and dumped the resolved vitest projects and the post-run module graph to check the config claims rather than reason about them.

Approving — the direction is right and the #2328 follow-ups are addressed. What's below is one substantive test-design issue (the batch fixture), three config robustness items, and some comment-accuracy nits. None of it affects shipped code.

Verified correct, so nobody re-checks it: the README repoint is right and fixes staleness that #2328 introduced — the describe block left ts-sdk in that PR, and at the merge base `git grep "Ledger expected-signature contract"` hits only the README; the sentence's subject and verb sit on the unchanged line, so only the object moved. The split loses no tests (303 files collected before and after: 302 `unit` + 1 `contracts`). `dist` really is bypassed at runtime — 111 ts-sdk and 29 signer modules from `src`, zero from either `dist`, a single WASM instance from `dist`. The `byInput === [0]` assertions are a real signal rather than an artefact: with `signInputIndexes` undefined both inputs would land there, since both classify as tapscript (`expectedSignatures.ts:260`, `:329`). The alias regex is exhaustive against all 13 `exports` subpaths and the vite `build.lib.entry` list. Project `resolve.alias` merges rather than replaces, and the unit exclude glob is not a no-op — so there is no shadow double-run. CI runs both projects (`nx affected --target=test` infers the target from `"test": "vitest"`, with no `--project` filter). `tsc` reports only two pre-existing errors in an untouched file; eslint is clean; `tsconfig`/eslint still cover the moved path; zero repo-wide references to the old filename or helper survive.

On the critical-path gate: it fired on a path prefix, but the substantive section-7 surface — the generated vectors, `scripts/gen_*.py`, `scripts/requirements.txt`, the oracle pins — is untouched. The only change under that directory is a single provenance pointer in the README, checkable by inspection.

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
```diff
@@ -249,4 +288,29 @@ describe("Ledger payout signing contract", () => {
...
+    const tables: ExpectedSignatureTable[] = [];
+    const manager = new PayoutManager({
+      network: "signet",
+      btcWallet: ledgerBatchWallet(tables),
+    });
+
+    // Two elements on purpose: element k must be staged with options[k].
+    await expect(
+      manager.signPayoutTransactionsBatch([params, params]),
+    ).rejects.toBe(STOP_AFTER_PREPARE);
```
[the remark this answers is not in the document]
jrwbabylonlab: fixed

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
jrwbabylonlab: fixed

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
jrwbabylonlab: fixed

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
jrwbabylonlab: fixed

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
jrwbabylonlab: fixed

services/vault/vitest.config.ts
jrwbabylonlab: fixed

services/vault/vitest.config.ts
jrwbabylonlab: fixed

services/vault/vitest.config.ts
jrwbabylonlab: fixed

services/vault/vitest.config.ts
jrwbabylonlab: fixed

services/vault/vitest.config.ts
jrwbabylonlab: fixed

services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts
```diff
@@ -249,4 +288,29 @@ describe("Ledger payout signing contract", () => {
...
+
+    expect(tables).toHaveLength(2);
+    for (const table of tables) {
+      expect([...table.byInput.keys()]).toEqual([0]);
+      expect(table.expectedYieldCount).toBe(1);
+      expect([...table.classifiedByInput.keys()]).toEqual([0, 1]);
+      for (const expectation of table.classifiedByInput.values()) {
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — tests keep explicit literals by design

gbarkhatov (dismissed):

jonybur (approved):
The contracts project stays standalone with no extends: true, so it does not inherit jsdom or the src/test/setup.ts mocks — matches the header comment. The batch signPsbts test uses distinct params and vkClaimerParams to cover the VK-claimer 2-output layout. srcAliasesFromExports builds one exact-match regex per exports key, including the bare '.' → package name, so unexported subpaths are refused. dmkSession.ts memoizes DMK/webhid imports per specifier via memoizedImport, closing the vitest concurrent-import race the added test pins. Nothing touches on-chain-binding or value-computation. No dead code, no silent fallbacks; memoizedImport clears its pending slot on rejection so an open retry stays possible. CI green.

PR #2332, opened 2026-08-26
feat(packages): compute the peg-in identifiers without the Bitcoin stack

jonybur:
## Outcome

Compute the peg-in transaction ID, vault ID, and Bitcoin key validity without loading the Bitcoin engine or Bitcoin libraries.

**Status:** Open
**Risk:** Critical path
**Size:** 47 files
**Review:** Two qualified approvals required
**Issue:** #2231

## Change

- Parse and hash a Bitcoin transaction without witness data.
- Derive the vault ID from the transaction hash and Ethereum address.
- Check whether a Bitcoin public key is on the secp256k1 curve.

## Safety

- Each new calculation is compared with the implementation that it replaces.
- Tests include published vectors, boundary cases, and seeded inputs.
- The curve check is already used by current production callers.
- The transaction and vault helpers become live when later registration work lands.
- Invalid mixed-case Ethereum addresses are rejected by design.
- The existing exact `@noble/hashes` dependency stays allowed on the Ethereum-only path. #2346 tracks the remaining package contract.

## Before merge

- [x] Record the mixed-case address decision.
- [x] Resolve the two open review threads.
- [ ] Get two qualified approvals.
- [x] Keep build, lint, SDK tests, Vault tests, and differential tests green.

Tracks #2231. #2231 stays open.


.github/workflows/claude-md-drift.yml
```diff
@@ -44,6 +44,7 @@ jobs:
             "packages/babylon-ledger-vault-signer/src/"
             "packages/babylon-ledger-vault-signer/scripts/"
             "packages/babylon-ts-sdk/src/tbv/core/clients/eth/onChainBtcPubkey.ts"
+            "packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts"
```
gbarkhatov, line 47: The `pending` -> `paths` move is correct and complete — without it the next Monday run would open a "pending path that has landed" report against this very file — and I confirmed both remaining `pending` entries are still absent from the tree. `critical-path-check.yml` and CODEOWNERS already listed the path, so they need no change.

But SECURITY.md:247-248 says the inventory is hand-maintained in five places and "Update all five together when a path moves or is added". This PR updated one of the five. The prose in the other four is now false as of this commit:

- `CLAUDE.md:109` — "Files (all but the last arrive with the optional-BTC work)"
- `CLAUDE.md:115` — onChainBtcPubkey "…the optional-BTC work **replaces** its `ecc.isXOnlyPoint` curve-membership check". It is replaced, in this PR.
- `.github/CODEOWNERS:50-51` — "Registered ahead of the code arriving …, except onChainBtcPubkey.ts, which already exists". Now two exceptions.
- `SECURITY.md:251` and `SECURITY.md:768` — "section 9 is registered ahead of the optional-BTC work (#2228)" / "pre-registered ahead of #2228". Now partially landed.

Worth noting that SECURITY.md's own risk table (`:768`) names this exact failure mode: "A critical path moves but one hand-maintained inventory keeps the stale path."

packages/babylon-ts-sdk/src/tbv/core/clients/eth/onChainBtcPubkey.ts
```diff
@@ -7,12 +7,57 @@
...
+  0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2fn;
+
+/** secp256k1's curve coefficient `b`, from `y^2 = x^3 + b`. */
+const SECP256K1_CURVE_B = 7n;
+
+/**
+ * Modular exponentiation without a crypto-library dependency.
+ *
+ * secp256k1's field prime is congruent to 3 mod 4, so an x-coordinate is on
+ * the curve iff `(x^3 + 7)^((p + 1) / 4)` squares back to `x^3 + 7` modulo
```
gbarkhatov: Two small things on this block, both because section 9 obliges a reviewer to line-walk the algebra without an assistant open.

**The "iff" has an unstated precondition.** `isSecp256k1XCoordinate` returns `true` when `curveValue === 0`, since `modPow(0, e, p) === 0` and `0 * 0 ≡ 0`. That is *provably unreachable* — I checked `(-7)^((p-1)/3) mod p != 1`, so `-7` is not a cubic residue and `x^3 + 7 ≡ 0` has no solution mod p; equivalently secp256k1's group order is odd, so no point has `y = 0`. The condition really is an iff, but only given that fact, and the fact isn't written down. One sentence here closes it.

**This JSDoc is attached to the wrong function.** It explains the *curve-membership condition*, but sits on `modPow`, a generic arithmetic helper that knows nothing about secp256k1. `isSecp256k1XCoordinate` (:45) — the function it actually describes, and the one on the critical path — has no doc comment at all. Suggest moving the curve paragraph down to :45 and leaving `modPow` with a one-liner.

For the record, the math itself is clean: I ran 224,015 x-coordinates against `ecc.isXOnlyPoint` (targeted values, a ±2^32 band around p, and 200k random) with zero disagreements, and the `x = 0` claim at :71 holds — 7 is a quadratic non-residue mod p, so `bytes32(0)` is rejected by both implementations.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/onChainBtcPubkey.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    const vectors = [
+      "00".repeat(32),
+      "00".repeat(31) + "01",
+      "ff".repeat(32),
+      "fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2e",
+      "fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f",
+      "fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc30",
+      "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
+      "c6047f9441ed7d6d3045406e95c07cd85c778e4b8cef3ca7abac09b95c709ee5",
+      ...Array.from({ length: 32 }, (_, index) =>
```
gbarkhatov: `index = 0` here produces `"0".padStart(64, "0")` — a byte-for-byte duplicate of `vectors[0]` at :41. 40 vectors, 39 distinct. Harmless, but it slightly obscures that the zero-hash case is covered, and that case is the one the module JSDoc calls out as load-bearing ("an unregistered operator … surfaces as an error rather than a silent all-zero key").

The generated set itself is *not* degenerate, which is worth saying since it looks like it might be: the pad alphabet is `0`-`9` only, so these are 32 distinct repdigit-ish values all below p, splitting 15 accept / 17 reject. And the seeded LCG is used correctly — Hull-Dobell holds for these constants, and `>>> 24` takes the high byte, which is the standard mitigation for an LCG's weak low bits. Over the exact 8,192 bytes this test consumes: 256/256 distinct values, chi-square 252.9 on df=255. No concern.

The gap is elsewhere: this suite only exercises the curve branch. For the sole minter of a branded type it is worth two more cases — the length/prefix rejection at `onChainBtcPubkey.ts:79-81`, and an uppercase input pinning both the `toLowerCase()` normalisation at :77 and the "returns 64-char lowercase hex without the 0x prefix" contract at :89. Every vector here is already lowercase, so neither is currently asserted.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts
```diff
@@ -0,0 +1,185 @@
+import { sha256 } from "@noble/hashes/sha2.js";
```
gbarkhatov, line 8: The module has no top-level JSDoc, and both of its exports are reachable only from `__tests__/pegin-transaction.test.ts:9`. Neither is re-exported from `clients/eth/index.ts`, so they are absent from `dist` and from the public API too.

I read the staging rationale in the description and I am not blocking on it, but nothing mechanically holds it accountable: `knip.json` lists `src/**/__tests__/**/*.ts` as an entry, so the test import alone keeps `pnpm knip` green. If the registration PR slips, nothing will flag this file.

Could the module JSDoc record (a) the follow-up that will consume these helpers, and (b) that the two input-contract findings below go live at swap-over rather than now? That also fixes the docstring gate, which failed at 45% — this file's three helpers (`decodeHex`, `encodeHex`, `concatBytes`) plus the module are most of the gap, and `onChainBtcPubkey.ts` documents everything, so the asymmetry is real.

Minor, while the cost is still zero: `calculatePeginTxHash` is PegIn-specific, but a good half of the `calculateBtcTxHash` sites it will eventually replace hash the **Pre-PegIn** transaction — `PeginManager.ts:1052`, `rebuildDepositTermsCore.ts:101`, `useVaultActions.ts:286`. The name will read wrong at most of ~15 call sites.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts
```diff
@@ -0,0 +1,185 @@
...
+    decodeHex(transactionHex, "depositorSignedPeginTx"),
+  );
+  const digest = sha256(sha256(serialized));
+  digest.reverse();
+  return `0x${encodeHex(digest)}`;
+}
+
+/** Derive the Solidity vault ID without loading the Rust/WASM package. */
+export function derivePeginVaultId(peginTxHash: Hex, depositor: Address): Hex {
+  if (!/^0x[0-9a-fA-F]{64}$/.test(peginTxHash)) {
```
gbarkhatov: `derivePeginVaultId` requires the `0x` prefix on both arguments, and its only caller passes neither. `PeginManager.ts:1416-1420`:

```ts
const peginTxHash = calculateBtcTxHash(depositorSignedPeginTxHex);
const derivedVaultIdHex = await deriveVaultId(
  stripHexPrefix(peginTxHash),
  stripHexPrefix(depositorEthAddress),
);
const vaultId = ensureHexPrefix(derivedVaultIdHex) as Hex;
```

Three inversions against the WASM function being replaced, all of which a mechanical swap in the follow-up would hit:

1. **Input prefix.** The engine's `hexToBytes` (`babylon-tbv-rust-wasm/src/index.ts:440`) strips an optional `0x` and accepts either form, so the call site strips. This regex demands it. A literal swap throws on every registration.
2. **Output prefix.** The engine returns *unprefixed* lowercase hex — see `primitives/__tests__/deriveVaultId.test.ts:24` and the `\`0x${engineVaultId}\`` the differential itself has to build. This returns prefixed, so `ensureHexPrefix(...)` must also go.
3. **Sync vs async.** `core/wasm/index.ts:391` returns `Promise<string>`; the `await` must go.

After those three edits the values are byte-identical, and downstream `vaultId` comparisons already `.toLowerCase()`, so casing is safe either way.

Also worth reconciling within this file: `calculatePeginTxHash` accepts both prefixed and unprefixed hex (`decodeHex` at :11 strips an optional prefix, and :343 pins it), while `derivePeginVaultId` requires the prefix. Two helpers, one file, opposite input contracts. Not wrong, but a caller has to know.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts
```diff
@@ -0,0 +1,185 @@
...
+    throw new Error("peginTxHash must be a 32-byte hex value");
+  }
+  if (!isAddress(depositor)) {
```
gbarkhatov: This is the "mixed-case address decision" the description asks reviewers to record, so — recording it: **keep strict.**

viem's `isAddress` defaults to `strict: true`, so it accepts all-lowercase unconditionally and otherwise requires an exact EIP-55 checksum. The WASM `deriveVaultId` it replaces validates *length only* (`hexToBytes` -> 20-byte assert) and returns the correct id for every casing. Verified side by side:

| depositor | `derivePeginVaultId` | WASM `deriveVaultId` |
|---|---|---|
| all-lowercase | `0xf8d22e64…` | `0xf8d22e64…` |
| EIP-55 checksummed | `0xf8d22e64…` | `0xf8d22e64…` |
| all-UPPERCASE | **throws** | `0xf8d22e64…` |
| mixed case, wrong checksum | **throws** | `0xf8d22e64…` |

Strict is right. `depositor` is fed from `this.config.ethWallet.account.address` (`PeginManager.ts:1382`) — a viem `WalletClient`, and I traced it back through `ethereumSubmit.ts:55` -> `ETHWalletProvider.tsx:304` -> `appkit/provider.ts:258` to wagmi, which `getAddress()`-checksums every account it hands out. The only rejected class is a mixed-case address with a wrong checksum, which is corruption rather than a legitimate input.

Two asks, both because this file is now a registered critical path:

1. **Put the decision in the JSDoc, not only in the PR body** — including the non-obvious consequence that all-uppercase hex (`0xABC…`) is rejected too, which is the case someone hand-typing or `.toUpperCase()`-ing an address will hit. A PR description is not where the next reader looks.
2. **Add a negative test.** Neither guard has one. Note that `encodeAbiParameters` lowercases the address before padding, so this check cannot change the output — it is pure input gating, which makes it exactly the kind of silent contract narrowing section 9 exists to catch. The two txid divergences each got a pinned test *and* a justifying paragraph; this one deserves the same. `vaultIdFixtures()` at :215-219 runs everything through `getAddress()` and `GOLDEN_DEPOSITOR` at :12 is all-lowercase, so no fixture can currently reach the rejected class.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/pegin-transaction.test.ts
```diff
@@ -0,0 +1,406 @@
...
+    expect(() => calculatePeginTxHash("deadbeef")).toThrow(/Truncated/);
+  });
+
+  it("rejects trailing bytes after a complete transaction", () => {
+    expect(() => calculatePeginTxHash(`${testTransaction(false)}00`)).toThrow(
+      /trailing byte/,
+    );
+  });
+
+  // The two cases below are the only input classes found where this parser and
```
gbarkhatov: "The two cases below are the only input classes found where this parser and bitcoinjs-lib do not agree" understates the surface. The canonicality guards fire at **five** compact-size positions, not one — `pegin-transaction.ts:83/86/89` is reached from input count (:105), input script length (:111), output count (:117), output script length (:122), and witness item count/length (:130/:133). Only input count is pinned, by the test at :325.

The reasoning at :319-324 is correct and is exactly why the other four matter: the parser re-emits every compact size verbatim (`transaction.slice(start, offset)` at :114/:124) where bitcoinjs re-encodes canonically. Every fixture in this file comes from `Transaction.toHex()`, which only ever emits canonical varints, so **deleting the guard at :86 or :89 leaves the entire suite green** while the parser produces a txid the oracle does not — precisely the mutation :323-324 says cannot hide.

Confirmed by hand; all three canonicalise to the same transaction, so bitcoinjs returns `0xd4c130510bd0fa3e32aa72491dbcc32fb9935a95f2445e0b1a8ab67e2dd11be8` for each while this parser throws:

```
# non-canonical OUTPUT count (1 as fd0100)
0200000001abababababababababababababababababababababababababababababababababab0300000000fffffffffd010050c300000000000002510000000000

# non-canonical INPUT script length (0 as fd0000)
0200000001abababababababababababababababababababababababababababababababababab03000000fd0000ffffffff0150c300000000000002510000000000

# non-canonical OUTPUT script length (2 as fd0200)
0200000001abababababababababababababababababababababababababababababababababab0300000000ffffffff0150c3000000000000fd0200510000000000
```

A non-canonical *witness* item count diverges too, and that one is a slightly different flavour worth a sentence of its own if you keep it: the witness section is excluded from the txid, so rejecting it is over-strictness with no byte-equality justification behind it, unlike the other four.

Either extend `NON_CANONICAL_INPUT_COUNT_ENCODINGS` across the five positions, or reword :308 and :319 so they don't claim coverage the suite doesn't have. I'd prefer the former — the comment is right about why these guards are load-bearing.

Two cheap additions while you're in here, both currently unasserted and both agreeing between the implementations: a 0-input/0-output transaction (`020000000000` + `00000000`) and a 1-input/0-output one. They'd pin the fact that the 0-input SegWit-marker ambiguity is inherited from the consensus encoding rather than chosen by either parser.

gbarkhatov (dismissed):
Reviewed with independent differential fuzzing on top of the committed suite. What I verified rather than read:

- **secp256k1 rewrite: clean.** 224,015 x-coordinates against `ecc.isXOnlyPoint` — targeted values, a ±2^32 band around p, and 200k random — **zero disagreements**. The `x = 0` claim in the module JSDoc holds: 7 is a quadratic non-residue mod p, so `bytes32(0)` is rejected by both implementations and an unregistered operator still surfaces as an error rather than an all-zero key.
- **txid parser: clean.** 260k differentially-fuzzed transactions (200k with randomised non-canonical compact sizes, 60k canonical-only) — **zero cases where both implementations parse successfully and produce different txids**. SegWit marker/flag detection and the superfluous-witness rule are semantically identical to bitcoinjs, including the 0-input encoding ambiguity, which both inherit from the consensus format rather than choose. No unbounded loop or memory blowup on hostile input: a `ff ffffffffffff1f00` input count throws in 0.2 ms.
- 76/76 `clients/eth` tests pass (19 new), `lint` clean, `tsc -b --noEmit` exit 0.
- CLAUDE.md section 9's differential rule is satisfied for all three reimplementations.

Approving. Nothing inline blocks — the two input-contract findings become live at swap-over, not now.

### Performance, since it cuts the opposite way from what you'd expect

The hand-rolled curve check is **~17x faster** than the asm.js oracle, not slower: 97 µs vs 1630 µs per call, plus ~20 ms of module init the ETH path no longer pays. That matters because `assertOnChainBtcPubkey` *is* called in a loop — `resolveParticipantKeys.ts:44,48` inside `resolveOne`, mapped over keepers (:119) and universal challengers (:122). A 100-participant roster: ~19 ms now vs ~326 ms before.

The parser is ~4x slower and ~2x heavier than bitcoinjs on an adversarial 8 MB transaction (421 ms / 292 MB vs 110 ms / 155 MB), because `take` uses `slice` (copy) where bitcoinjs uses `subarray` (view), and `concatBytes` then copies a third time. Bounded and linear; irrelevant at real peg-in sizes. Noting it only so it isn't a surprise later.

### #2346 is not blocked by this PR

`@noble/hashes` is already a pinned exact hard dependency (`2.0.1`, `dependencies` not peer) imported by 18 SDK `src/` files, two of them already under `clients/` — and this PR touches no `package.json`. The question can be closed on that basis. If you do want it off the ETH path anyway, `viem` is already imported in this file and exports `sha256`; I checked that `sha256(bytes, "bytes")` returns a 32-byte `Uint8Array`, so it's a zero-cost swap rather than a new dependency decision.

### One architectural note, non-blocking

The subtree genuinely is Bitcoin-free now: the transitive graph from `clients/eth/index.ts` is 14 modules with `viem` as the only external specifier, where before it pulled in `tiny-secp256k1-asmjs` (via `onChainBtcPubkey.ts`) and `bitcoinjs-lib` (via `primitives/utils/bitcoin.ts`, where `hexToUint8Array` lives). That's the real payoff and it is delivered.

But there's no `./tbv/core/clients/eth` subpath export. The nearest exported entry, `./tbv/core/clients`, is 42 modules and still pulls in `bitcoinjs-lib`, `tiny-secp256k1-asmjs`, the WASM package and `buffer` via `vault-provider/auth/{serverIdentity,bip322Verify,verifyDepositorCwt}.ts`. `"sideEffects": false` lets an ESM bundler shake it out, but the CJS conditions load the whole graph. Worth adding the subpath export in the follow-up so the win is observable from outside the package.

### On the two open automated threads, I'd decline both

- `expect(oracleOutcomes).toEqual(new Set([false, true]))` is a real guard, not corpus-testing: it proves the seeded corpus straddles the accept/reject boundary, so a future seed change that silently collapsed it to all-rejects would fail loudly. Keep it.
- Naming `0n`/`1n`/`2n` inside a square-and-multiply loop costs more readability than it buys. The one fair extraction is the square-root exponent — `(SECP256K1_FIELD_PRIME + 1n) >> 2n` is the single line where an arithmetic slip would still typecheck and still pass a weak test.

Also checked and clean: `tiny-secp256k1-asmjs` correctly stays a peer dependency (12 non-test `src/` files still import it); dropping `hexToUint8Array` from `onChainBtcPubkey.ts` orphaned nothing (~30 remaining call sites); the ESLint critical-path overrides do apply to both new source files and the test file is legally exempt from the lazy-WASM-boundary import rule, both verified with `--print-config`; and the genesis-coinbase anchor and `GOLDEN_VAULT_ID` both match their published / already-pinned values.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/onChainBtcPubkey.test.ts
```diff
@@ -0,0 +1,105 @@
...
+        assertOnChainBtcPubkey(`0x${value}` as Hex, "test key");
+      const message = `seed 0x${FIELD_DIFFERENTIAL_SEED.toString(16)}, fixture ${fixture}`;
+      if (expected) {
+        expect(validate(), message).toBe(value);
+      } else {
+        expect(validate, message).toThrow(/not on the secp256k1 curve/);
+      }
+    }
+
+    expect(oracleOutcomes).toEqual(new Set([false, true]));
```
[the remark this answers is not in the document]
jonybur: Declined after review. The assertions protect the differential corpus and oracle assumptions. They make a degraded test corpus visible without changing production behavior.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/onChainBtcPubkey.ts
```diff
@@ -7,12 +7,57 @@
...
+  const xCubed =
+    (((x * x) % SECP256K1_FIELD_PRIME) * x) % SECP256K1_FIELD_PRIME;
+  const curveValue = (xCubed + SECP256K1_CURVE_B) % SECP256K1_FIELD_PRIME;
+  const y = modPow(
+    curveValue,
+    (SECP256K1_FIELD_PRIME + 1n) >> 2n,
+    SECP256K1_FIELD_PRIME,
+  );
+
+  return (y * y) % SECP256K1_FIELD_PRIME === curveValue;
```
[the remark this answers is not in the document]
jonybur: Declined after review. These small protocol literals stay next to the arithmetic and parser operations. Extracting all of them would add indirection without reducing risk.

.github/workflows/claude-md-drift.yml
jonybur: Fixed in 81897e89. `CLAUDE.md`, `CODEOWNERS`, `SECURITY.md`, and the drift workflow now use the current inventory. The remaining pending paths are still absent.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/onChainBtcPubkey.ts
jonybur: Fixed in 81897e89. `modPow` has a focused comment. The curve explanation now documents `isSecp256k1XCoordinate`, including why the zero curve value is impossible.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/onChainBtcPubkey.test.ts
jonybur: Fixed in 81897e89. The duplicate zero vector is gone. Tests now cover invalid prefixes, invalid lengths, and uppercase normalization.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts
jonybur: Fixed in 81897e89. The module now documents the follow-up and input contracts. The helper is named `calculateBtcTxHash`. PR #2336 now consumes and exports it.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts
jonybur: Fixed in 81897e89. The helper accepts optional prefixes, runs synchronously, and returns lowercase hex without a prefix. PR #2336 adds `0x` only at the Ethereum contract boundary.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts
jonybur: Fixed in 81897e89. The JSDoc records strict EIP-55 validation and all-uppercase rejection. Tests cover invalid mixed-case and uppercase addresses.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/pegin-transaction.test.ts
jonybur: Fixed in 81897e89. Tests cover every compact-size position, witness counts and lengths, and the zero-input and zero-output transaction shapes.

jonybur:
@greptileai

gbarkhatov (approved):

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/pegin-transaction.test.ts
```diff
@@ -0,0 +1,492 @@
...
+      const txHex = `${legacy.slice(0, VERSION_HEX_LENGTH)}${encoded}${legacy.slice(VERSION_HEX_LENGTH + CANONICAL_INPUT_COUNT_HEX_LENGTH)}`;
+      expect(() => calculateBtcTxHash(txHex), prefix).toThrow(
+        /Non-canonical input count/,
+      );
+      expect(calculateBitcoinJsTxHash(txHex), prefix).toBe(canonicalTxid);
+    }
+  });
+
+  // Witness compact sizes do not enter the txid. Rejecting their non-canonical
+  // form is a strict parser contract, not a byte-equality requirement.
```
gbarkhatov: **This comment is wrong for three of the five fixtures it heads.**

It says witness compact sizes do not enter the txid and that rejecting their non-canonical form is "not a byte-equality requirement" — but only the last two entries of `NON_CANONICAL_COMPACT_SIZE_POSITIONS` are witness. For input script length, output count and output script length, rejection *is* a byte-equality requirement.

`pegin-transaction.ts:108` re-emits every compact size verbatim (`transaction.slice(start, offset)`), and those bytes reach the hash preimage through `inputs` (`:127`), `outputCount.encoded` (`:168`) and `outputs` (`:137`). Witness bytes are the only ones excluded from the concat at `:164-171`, which is precisely why the sentence is true for two rows and false for three.

The failure it enables: someone reads this as the rationale for the whole group, concludes there is no oracle behind it, and relaxes or deletes a guard at `:96`/`:99`/`:102` for a non-witness position. Nothing fails — every fixture here comes from `Transaction.toHex()`, which only ever emits canonical varints — and the parser then hashes a non-canonical length verbatim, returning a vault-binding txid that no confirmable transaction produces.

That is the same class of misdocumentation the earlier version of this comment had, moved one layer down.

Fix: scope the sentence to the two witness rows — move it into those entries, or split the array into a txid-bearing group and a witness group — and give the three txid-bearing rows the rationale already written at `:369-373`.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts
```diff
@@ -0,0 +1,215 @@
+/**
+ * Dependency-free Bitcoin txid and vault-ID helpers for the registration
+ * client in #2336.
+ *
+ * This module stays test-only until that follow-up replaces the PeginManager
+ * consumers. At that swap, transaction hex and vault-ID inputs accept an
+ * optional `0x` prefix. Depositor addresses use strict EIP-55 validation.
+ * Vault IDs return synchronously as lowercase hex without a prefix.
```
gbarkhatov: The module doc records one output convention, but the file's two exports now have opposite ones: `calculateBtcTxHash` returns `0x`-prefixed `Hex` (`:181`), `derivePeginVaultId` returns unprefixed lowercase `string` (`:214`).

The asymmetry is correct and should stay — the txid half matches the incumbent `utils/transaction/btcTxHash.ts:23` it replaces, the vault-ID half matches the WASM helper, and the differential at `:487-489` depends on exactly that. It is only that this became a two-convention file in this commit and the doc still describes one of them.

Worth a clause here saying txids come back prefixed and vault IDs do not, so a #2336 consumer does not assume a single convention and double-prefix the txid or drop the prefix step on the vault ID.

CLAUDE.md
```diff
@@ -104,15 +104,15 @@ These paths handle irreversible value movement. An AI-generated mistake here is
...
   - `packages/babylon-ts-sdk/src/tbv/core/wasm/` — the lazy boundary every WASM-computed value now crosses
   - `packages/babylon-tbv-rust-wasm/src/wasm-loader.ts`, `wasm-loader-node.ts`, `raw.ts`, `raw-node.ts` — the restructured engine entry surface (the `@stability frozen` rules in section 4 still apply)
-  - `services/vault/src/utils/btc/scriptPubKeyAddress.ts` — hand-written bech32, bech32m and base58check encoding
-  - `packages/babylon-ts-sdk/src/tbv/core/clients/eth/onChainBtcPubkey.ts` — already in the tree, and guarded from now rather than on arrival: it is the sole validator minting `OnChainBtcPubkey`, and the optional-BTC work replaces its `ecc.isXOnlyPoint` curve-membership check with hand-rolled field arithmetic
+  - `services/vault/src/utils/btc/scriptPubKeyAddress.ts` - hand-written bech32, bech32m and base58check encoding; registered ahead of arrival
+  - `packages/babylon-ts-sdk/src/tbv/core/clients/eth/onChainBtcPubkey.ts` - the sole validator minting `OnChainBtcPubkey`; the optional-BTC work replaced its `ecc.isXOnlyPoint` curve-membership check with hand-rolled field arithmetic
```
gbarkhatov, line 116: Trivial: these two rewritten bullets separate path from gloss with an ASCII hyphen, where every other bullet in this list uses an em dash.

gbarkhatov (dismissed):
Follow-up to my review of `5d206664`, which was dismissed by the fix commit. This pass covers only what changed since — `5d206664...81897e89`, one commit, 7 files, +194/−57. The approval that already sits on this PR was posted without a body and did not audit that commit, so this is the pass that does.

**All seven prior findings are fixed.** I checked each against the code at HEAD rather than against the replies. Their existing threads stand as written — nothing restated here.

| # | Claim | Verdict | HEAD anchor |
|---|---|---|---|
| 1 | Drift-workflow inventory vs. hand-maintained SECURITY.md | fixed | `CLAUDE.md:109` matches `pending=()` at `claude-md-drift.yml:62-65` |
| 2 | Curve "iff" missing the `curveValue === 0` precondition | fixed | `onChainBtcPubkey.ts:43-44` |
| 3 | Duplicate zero vector — 40 vectors, 39 distinct | fixed | `onChainBtcPubkey.test.ts:51-53`, zero hash still explicit at `:43` |
| 4 | No module JSDoc; staging rationale unaccountable | fixed | `pegin-transaction.ts:1-9`, renamed at `:175` |
| 5 | `derivePeginVaultId` prefix contract inverted vs. its caller | fixed | `pegin-transaction.ts:197-214` |
| 6 | Mixed-case decision unrecorded, no negative test | fixed | `pegin-transaction.ts:187-191` |
| 7 | Comment understates the surface — five guard positions | fixed on coverage; the new group comment is wrong, see inline | `pegin-transaction.test.ts:48-84` |

The two that were section-9 load-bearing, checked in detail rather than taken on the reply:

**5** — a literal substitution for the incumbent now works with zero caller edits. `PeginManager.ts:1416-1421` strips the prefix off both arguments (re-added internally at `:197`/`:203`) and re-prefixes the result (returned unprefixed via `.slice(2)` at `:214`). Strict EIP-55 does not break that path, because `stripHexPrefix` of a checksummed address round-trips to the same string.

**7** — all five positions now carry a non-canonical fixture with a distinct error regex, plus input count at `:374-387`. The `.replace()` fixtures are not vacuous: `expect(txHex).not.toBe(canonical)` at `:398` turns any drift in `testTransaction()` into a loud failure rather than a silent pass.

Section 9 conformance holds after the delta — the differentials import the implementations actually being replaced (`utils/transaction/btcTxHash.ts` aliased as `calculateBitcoinJsTxHash`, and the WASM `deriveVaultId`) and run golden vectors plus seeded randomised inputs against them.

Three new items from the fix commit are inline below. Only the first is worth acting on before merge: it is comment-only and changes no behaviour, but it misdocuments the exact thing the compact-size finding existed to remove, in a path section 9 now covers. The other two are a doc clause and a typography nit.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts
jonybur: Fixed in ebb9171a0. The module contract now states that txids use lowercase hex with a 0x prefix and vault IDs use lowercase hex without a prefix.

CLAUDE.md
jonybur: No code change. The repository rules require regular hyphens and prohibit em dashes, so the current punctuation is intentional.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/pegin-transaction.test.ts
jonybur: Fixed in ebb9171a0. The rationale now separates txid byte-equality checks from the strict witness parser contract.

gbarkhatov (dismissed):
@jonybur branch should be updated

jonybur:
Updated the branch for final review.

- Replaced the GitHub-generated squash commit with an equivalent FIDO2-signed commit.
- Merged the current main branch with a FIDO2-signed commit.
- Preserved the PR diff and stable patch ID.
- Affected build, test, and lint targets pass.
- Proto generation passes.

The code diff is unchanged. This is ready for fresh code-owner review.

jonybur:

gbarkhatov (approved):

PR #2336, opened 2026-08-26
feat(packages): extract the Ethereum registration client and make the engine optional

### LINKED ISSUE #2231: Complete Bitcoin engine safeguards and release checks

jonybur:
## Outcome

- Add the Ethereum-only peg-in registration client.
- Verify each payout script against the proof-of-possession Bitcoin key before any RPC read or wallet call.
- Keep Bitcoin libraries and the WASM engine optional for Ethereum-only consumers.
- Add clean package and release checks.

## Change

Payout safety:

- Derive BIP-86 P2TR scripts from the x-only proof-of-possession key.
- Derive P2WPKH scripts from the compressed key in the proof witness.
- Reject foreign, malformed, unsupported, and over-length scripts.
- Validate the complete batch before the first contract read.
- Do not export a marker or branded type that can bypass the check.
- Pin single and batch calldata, selectors, ETH value, empty batches, and duplicate batches.
- Compare both script derivations with bitcoinjs-lib over fixed vectors and 64 fixed-seed randomized keys.

This changes a critical value-moving path. Two qualified reviewers must approve it.

Package contract:

- `viem ^2.38.2` is the only required peer.
- `bitcoinjs-lib@6.1.7` and `@bitcoin-js/tiny-secp256k1-asmjs@2.2.3` are optional peers.
- `@babylonlabs-io/babylon-tbv-rust-wasm` is an optional peer. The release driver replaces `workspace:*` with the exact release version.
- The WASM package exports only `.` and `./raw`. The temporary `./types` entry was never released and stays unsupported.
- The clean consumer declares only the SDK and viem. It checks types, ESM, CommonJS, exports, exact peer policy, and the absence of Bitcoin and WASM peers.

The new direct crypto dependency is pinned at `@noble/curves@1.9.7`. It was already present in the lockfile through viem. It has no install hook and no reported advisory.

## Safety

Two qualified reviewers must approve the final value-moving changes. The second distinct review record remains open in #2357.

The original release rehearsal used:

- SDK: `0.72.0`
- WASM engine: `0.17.1`
- Published SDK WASM peer: exact `0.17.1`
- ETH export: types, ESM, and CommonJS files are present in the packed tarball.

## Verification

- SDK build and clean packed-consumer check pass.
- 1,840 SDK tests pass.
- 69 release-driver tests pass.
- SDK, WASM, and release lint targets pass.
- Release typecheck passes.
- API docs validate with 0 errors.
- Syncpack reports no mismatch.
- The focused release dry run passes.

Completion targets:

- #2346: the package contract, clean packed install, release rehearsal, and published exports are complete. [Final CI passed](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/33628092288). [SDK 0.73.0](https://github.com/babylonlabs-io/babylon-toolkit/releases/tag/%40babylonlabs-io/ts-sdk/0.73.0) shipped with exact optional engine 0.18.0.
- #2357: payout validation and calldata tests are complete. Remaining: link or obtain the required second qualified final review. No further implementation is identified.

## Links

Merged into #2332, then delivered to main through #2332. Completes #2346. Tracks #2357. #2357 stays open. Tracks #2231. #2231 stays open.


jonybur:
@greptileai

jonybur:
@greptileai

jonybur:
@greptileai

packages/babylon-ts-sdk/src/tbv/core/clients/eth/payout-script.ts
```diff
@@ -0,0 +1,114 @@
+import { schnorr, secp256k1 } from "@noble/curves/secp256k1.js";
```
gbarkhatov, line 1: **MERGE BLOCKER — this file is a CLAUDE.md §9 reimplementation registered on none of the six critical-path surfaces.**

It hand-rolls BIP-341 TapTweak and hash160 P2WPKH, replacing bitcoinjs `payments.p2tr`, `payments.p2wpkh` and `address.toOutputScript` — exactly what §9 governs — and it mints the scriptPubKey committed on-chain irreversibly.

Verified by grep: `payout-script` returns **zero hits** across `CLAUDE.md`, `.github/CODEOWNERS`, `claude-md-drift.yml`, `critical-path-check.yml`, `SECURITY.md` and `eslint.config.mjs`. The same grep for its sibling `pegin-registration-client.ts` hits **all six** — this PR promotes that file from "pending" to "in tree" everywhere and adds this one nowhere. The cause looks like commit ordering: `payout-script.ts` arrived in `9d88e1c08`, after the §9 bookkeeping landed in `9bb914c18`.

Consequence: no two-reviewer requirement on the file that derives the payout script, no weekly drift check, and it is exempt from the `no-explicit-any` / `no-non-null-assertion` / `ban-ts-comment` hardening this same PR *expanded* to cover its sibling.

Fix: add the path to all six lists, alongside `pegin-registration-client.ts`.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/payout-script.ts
```diff
@@ -0,0 +1,114 @@
...
+    schnorr.utils.pointToBytes(outputKey),
+  )}`;
+}
+
+function deriveP2wpkhScript(btcPopSignature: Hex, xOnlyPubkey: string): Hex {
+  const witness = decodeWitnessStack(
+    hexToBytes(btcPopSignature.slice(2)),
+    "proof of possession witness",
+  );
+  if (witness.length !== 2) {
```
gbarkhatov: **Behaviour narrowing: a compressed-key wallet that signs a one-item PoP can no longer register a P2WPKH payout.**

`witness.length !== 2` is a hard requirement here. The code this replaced — `isAddressFromPublicKey` (`primitives/utils/bitcoin.ts`) — accepted a P2WPKH payout for any caller supplying a 66-hex compressed key, entirely independent of PoP witness shape.

Concrete failure: a wallet that exposes a compressed pubkey but signs the PoP through its Taproot address (one-item Schnorr witness) previously registered fine and now throws `"A P2WPKH payout requires a two-item proof of possession witness"`.

Not reachable from the vault app (the connected address is always passed, so the payout and PoP shapes agree), so this is an SDK-consumer regression — untested, and unmentioned in the PR body. Same root cause and same fix as the parity comment below: take the compressed key as an explicit parameter instead of deriving it from witness arity.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/payout-script.ts
```diff
@@ -0,0 +1,114 @@
...
+    );
+  }
+  try {
+    secp256k1.Point.fromBytes(pubkey);
+  } catch {
+    throw new Error(
+      "The proof of possession witness contains an invalid BTC pubkey",
+    );
+  }
+  if (bytesToHex(pubkey.subarray(1)) !== xOnlyPubkey) {
```
gbarkhatov: **MERGE BLOCKER — the P2WPKH payout binds to a y-parity byte taken from a PoP witness that is never verified.**

`deriveP2wpkhScript` decodes `btcPopSignature`, takes `witness[1]` as the compressed pubkey, and accepts it after checking only length, SEC1 prefix, on-curve-ness, and — here — that its **x** equals `depositorBtcPubkey`. Nothing in this module verifies the BIP-322 signature, so the `02`/`03` parity byte is constrained by nothing trusted.

The sharpest evidence is the code this replaced. `resolvePayoutScriptPubKey` sourced parity from a **live wallet read** (`assertPopMatchesBtcWallet` → `btcWallet.getPublicKeyHex()`) and deliberately **failed closed** on precisely this case, with the comment:

> Diagnostic carve-out: x-only key + P2WPKH address always fails (y-parity is unknowable from x-only).

This PR converts that intentional fail-closed into a recovery from an unauthenticated blob.

**Containment:** `PeginManager.ts:1354-1362` still calls `assertPopMatchesBtcWallet` and derives the script from the wallet key before delegating, so on the app path a tampered witness causes a *rejection*, not a wrong script. The gap is in the newly exported standalone `ViemPeginRegistrationClient` — whose entire audience is integrators with no BTC stack, i.e. exactly the callers who cannot verify the witness themselves. The README advertises it as validating the payout script against the PoP key, which reads stronger than what this does.

**Severity, stated precisely:** the only freedom is the parity byte, so the substituted point is −P, whose private key is `n − d` — known to the depositor. Funds are not cryptographically lost; they land at an address no consumer wallet will derive or scan, recoverable only by manually negating the key. Stranded, not stolen.

Fix: take the parity-bearing compressed key as an explicit `RegisterPeginOnChainParams` field — `PeginManager` already holds it as `verifiedBtcPubkeyRaw` — rather than sniffing it out of an unauthenticated signature.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/payout-script.ts
```diff
@@ -0,0 +1,114 @@
...
+  const xOnlyPubkey = assertOnChainBtcPubkey(
+    `0x${keyBody}` as Hex,
+    "Proof of possession BTC pubkey",
+  );
+
+  if (payoutScript === deriveP2trScript(xOnlyPubkey)) {
+    return payoutScript;
+  }
+  if (
+    /^0x0014[0-9a-f]{40}$/.test(payoutScript) &&
```
gbarkhatov: `/^0x0014[0-9a-f]{40}$/` re-hardcodes `P2WPKH_SCRIPT_PREFIX`, declared at line 12, plus the hash160 length. If the constant ever changes, this guard silently desyncs from the derivation two lines below that uses it. Build the pattern from the constant — minor, but it is the No Magic Numbers rule on a critical-path file.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts
```diff
@@ -0,0 +1,445 @@
...
+  COMMISSION_BPS_HEADROOM,
+} from "../../deposit-terms/commissionCeiling";
+import { waitForTransactionReceiptSmartAware } from "../../utils/eth";
+import { assertOnChainBtcPubkey } from "./onChainBtcPubkey";
+import { assertPayoutScriptMatchesPopKey } from "./payout-script";
+import { calculateBtcTxHash, derivePeginVaultId } from "./pegin-transaction";
+import { ViemVaultRegistryReader } from "./vault-registry-reader";
+
+const NO_REFERRAL_CODE = 0;
+const RECEIPT_TIMEOUT_MS = 120_000;
```
gbarkhatov, line 28: Three small constant issues introduced by the move:

- `RECEIPT_TIMEOUT_MS` arrives bare. The original carried its rationale — *"Matches the prior vault-service polling timeout so users see a clear error instead of an indefinite hang when a transaction is dropped from the mempool"* — which was dropped in the extraction. Carry the comment across.
- `assertHtlcVout` (line 46) inlines `255` rather than naming it as the uint8 ceiling the ABI declares.
- `NO_REFERRAL_CODE` (line 24) is now defined here **and** at `PeginManager.ts:99`. Import it from one place.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts
```diff
@@ -0,0 +1,445 @@
...
+    : value.startsWith("0X")
+      ? `0x${value.slice(2)}`
+      : `0x${value}`;
+  if (!/^0x(?:[0-9a-fA-F]{2})+$/.test(prefixed)) {
+    throw new Error(`${label} must be non-empty, even-length hex`);
+  }
+  return prefixed as Hex;
+}
+
+function assertBytes32(value: Hex, label: string): void {
```
gbarkhatov, line 42: **The new input-sanitisation layer ships with no negative-path tests.**

`assertBytes32` and `assertHtlcVout` are new in this PR (the base had only `ensureHexPrefix`, which prefixes without validating), and they are the whole guard between caller input and on-chain calldata. Verified by grepping their throw strings across `src`: `"must be a 32-byte hex value"` and `"must be a uint8"` appear **only at the throw sites** — no test anywhere exercises either.

The same gap exists in `payout-script.ts`: the off-curve `Point.fromBytes` catch (:63-69) and the x-mismatch throw (:70-74) are unexercised. The arity guard *is* covered, at `payout-script.test.ts:125`. These are the branches that reject attacker-shaped input, so they are the ones worth pinning.

A regression making `assertHtlcVout` accept `256` or `-1` would be invisible to the suite. Four short cases on the existing `setup()` helper in `pegin-registration-client.test.ts` would close it: a non-bytes32 hashlock, `htlcVout: 256`, an off-curve witness key, and an x-mismatched witness key.

packages/babylon-ts-sdk/scripts/check-packed-consumer.js
```diff
@@ -0,0 +1,201 @@
...
+      typeFixture,
+    ],
+    consumerRoot,
+  );
+
+  const runtimeCheck =
+    `const hash = "0x" + "11".repeat(32);\n` +
+    `const depositor = "0x1111111111111111111111111111111111111111";\n` +
+    `if (typeof sdk.ViemPeginRegistrationClient !== "function") throw new Error("Missing client");\n` +
+    `if ("markPayoutScriptVerifiedAgainstPopKey" in sdk) throw new Error("Unsafe payout marker is public");\n` +
```
gbarkhatov: **This guard asserts the absence of a symbol that has never existed anywhere in the repo.**

A repo-wide grep for `markPayoutScriptVerifiedAgainstPopKey` returns exactly one hit: this line. It is not exported, never was, and no source or test references it — so the check can never fire, while reading like real payout-safety coverage in a script that gates every SDK build. That false assurance is the concrete cost: a reviewer sees a fenced-off "unsafe payout marker" that was never a risk.

Delete it, or point it at a symbol that actually exists — presumably asserting `assertPayoutScriptMatchesPopKey` is *not* reachable from the public barrel, which looks like the original intent.

packages/babylon-ts-sdk/package.json
```diff
@@ -86,7 +101,8 @@
     }
   },
   "scripts": {
-    "build": "tsc -b --noEmit tsconfig.lib.json && vite build",
+    "build": "tsc -b --noEmit tsconfig.lib.json && vite build && node scripts/check-eth-build.js && pnpm check:packed-consumer",
```
gbarkhatov: **`build` now packs a tarball and runs a networked `pnpm install`.**

`check:packed-consumer` packs the package, writes a throwaway consumer, and runs `pnpm install --prefer-offline --no-lockfile` — which still needs registry metadata for `viem@2.38.2` — then two `node` runs and a `tsc`. `CLAUDE.md` requires rebuilding `ts-sdk` before every vault build, so this lands on the ordinary developer loop and on every CI job.

Concrete failure: an offline or registry-restricted machine that compiles perfectly well can no longer run `pnpm run build` at all, and every build pays the pack-and-install cost.

The check itself is valuable — keep `check-eth-build.js` (pure fs, no network) in `build` and move `check:packed-consumer` to `test` or a dedicated CI target.

packages/babylon-ts-sdk/src/tbv/core/index.ts
```diff
@@ -27,3 +27,4 @@ export * from "./services";
 export * from "./vault-secrets";
 export * from "./deposit-terms";
 export * from "./recovery";
+export { calculateBtcTxHash } from "./clients/eth";
```
gbarkhatov: **This line silently switches which `calculateBtcTxHash` the public barrel exports.**

`core/index.ts` already star-exports both `./utils` (the bitcoinjs `Transaction.getId` implementation, `utils/transaction/btcTxHash.ts`) and `./clients` (the hand-rolled parser, `clients/eth/pegin-transaction.ts`). Two star exports of *different* declarations under one name is ambiguous under ES semantics and drops the name from the barrel — so this explicit re-export is load-bearing, not cosmetic.

Its effect: `@babylonlabs-io/ts-sdk/tbv/core` now hands out the strict parser, while `/tbv/core/utils` still hands out the lenient bitcoinjs one — and `/tbv/core/utils` is the path every `services/vault` import uses. The two are not interchangeable at the edges: the new parser throws on trailing bytes and on a superfluous witness marker where `Transaction.fromHex` tolerates both, so an external consumer's previously-working call can now throw.

No consumer breaks today and #2332 supplies the differential, so this is a trap rather than a bug. Either keep one implementation package-wide (delete `utils/transaction/btcTxHash.ts` and repoint its consumers at the pure one), or add a comment here explaining why the collision is resolved this way.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/eth-import-boundary.test.ts
```diff
@@ -0,0 +1,159 @@
...
+import { describe, expect, it } from "vitest";
+
+const BANNED_RUNTIME_IMPORTS = new Set([
+  "bitcoinjs-lib",
+  "@bitcoin-js/tiny-secp256k1-asmjs",
+  "@babylonlabs-io/babylon-tbv-rust-wasm",
+]);
+
+const STATIC_SPECIFIER =
+  /(?:import\s+(?!type\b)[\s\S]*?\sfrom\s*|export\s+(?!type\b)[\s\S]*?\sfrom\s*)["']([^"']+)["']/g;
```
gbarkhatov: This regex is weaker than the one in the build gate it mirrors, and can pass vacuously.

`STATIC_SPECIFIER` requires whitespace after `import` and before `from` — precisely the bug `check-eth-build.js:6-9` documents avoiding ("a pattern that required whitespace after the keyword or before `from` would walk straight past an eager edge and report success on an empty closure"). It also excludes the type imports the build script includes deliberately.

Compounding it: unlike the build script (`MINIMUM_DEPENDENCY_CLEAN_MODULES = 52`), this test has no closure-size floor, and its only negative fixture exercises `SIDE_EFFECT_SPECIFIER` alone. So a `STATIC_SPECIFIER` regression would leave the five source-entry assertions comparing `[]` against `[]` over one-file closures, and still pass.

The emitted-output gate still catches the real thing, so this is test robustness rather than a hole in the boundary. Reuse `check-eth-build.js`'s patterns instead of re-deriving weaker ones.

gbarkhatov (approved):
## Combo review — three independent passes, deduped and verified

Reviewed against the base branch (`split/b-pegin-identifiers`, #2332), so nothing below belongs to #2332's diff. Three independent reviews ran, findings were merged, and everything posted here was re-verified directly against the source rather than accepted on agreement between reviewers.

**3 merge-blockers**, 5 medium, several nits — all inline.

### Two blockers that can't be anchored inline (the files aren't in the diff — that is the defect)

**A. `SECURITY.md:247-253` is left factually wrong, breaking its own rule.** Line 247 states the critical-path inventory "is hand-maintained in five places: this file, CLAUDE.md, CODEOWNERS, `critical-path-check.yml`, and `claude-md-drift.yml`. Update all five together when a path moves or is added." Lines 251-253 then say §9 registers `pegin-registration-client.ts` and `scriptPubKeyAddress.ts` *"ahead of the remaining optional-BTC work"* and that "those paths stay in the drift workflow's `pending` list until the files land." Both statements are false once this merges — the file landed and this PR moved it out of `pending`. Fix: name only `scriptPubKeyAddress.ts`, mirroring the wording already applied to `CLAUDE.md` and `CODEOWNERS` in this same PR.

**B. `pegin-transaction.ts:1-9` header is now false.** It reads *"This module stays test-only until that follow-up replaces the PeginManager consumers."* This PR is that follow-up — `pegin-registration-client.ts` now consumes both exports on the production registration path.

### Also worth a line, too minor to anchor

`pegin-registration-client.ts:291-298` — `normalizeBtcPubkey` checks only `startsWith("0x")`, while both its siblings (`ensureHex` at :28-32 and `normalizeHex` in `payout-script.ts:19`) handle the uppercase `0X` prefix. It fails closed inside `assertOnChainBtcPubkey`, so this is a confusing-error inconsistency, not a correctness bug. Reuse `ensureHex`.

### One framing correction, because it changes the severity

All three reviews independently described the parity substitution in `payout-script.ts` as producing an address for "a key the depositor does not control" — a framing inherited from the deleted code comment. That is not right. The only degree of freedom is the parity byte, so the substituted point is exactly −P; if the depositor's private key is `d`, the negated key's is `n − d`, which they also know. So the failure is that funds land at an address no consumer wallet will derive, scan, or spend — **stranded in practice, recoverable only by manual key negation, not key loss.** Still a defect on an irreversible path; the fix is unchanged. Flagging it so the record isn't overstated.

### Verified clean — not worth re-litigating

- **Calldata is correct.** All 11 args of `submitPeginRequest` and the 9-field batch tuple match the unchanged in-tree `BTCVaultRegistry.abi.ts` and the old `PeginManager` byte-for-byte; the 12-arg `referralCode` overload is correctly not selected; the tests pin both selectors (`0x67df3144`, `0x68d177ac`) independently rather than through mocks.
- **§9's differential requirement for `payout-script.ts` is genuinely met** — 4 fixed scalars plus 64 fixed-seed randomised keys against bitcoinjs `p2tr`/`p2wpkh`, with both parities asserted present. Two reviewers independently re-derived over 500 random keys: 0 mismatches.
- **`@noble/curves@1.9.7` is used correctly** — every symbol used exists at that exact version, and the `t >= n` guard matches BIP-341.
- **No assertions were dropped in the extraction**, and several were added (`assertBytes32`, `assertHtlcVout`, batch duplicate-`vaultId` rejection, payout verification before any RPC read).
- **The dependency-clean boundary holds at source level** — the import closures of the new ETH entry points reach no bitcoinjs / tiny-secp256k1 / WASM edge.
- **The `CLAUDE.md` / `CODEOWNERS` / eslint edits are not self-serving** — the path lists were only expanded, never weakened. Blocker #1 is a gap in that same edit, not a weakening of it.

### Checked and dropped

One reviewer claimed vaultd re-verifies the PoP on ingestion (which would contain the parity finding); another cited `crates/vaultd/src/workers/claimer/pegin.rs` for the opposite. Not resolved at acceptable cost, so the inline finding is stated without relying on either claim. A further claim that the import-boundary test's `workspace:*` assertion breaks after release pinning is unproven — `tools/release/src/manifest.ts` is a pure transform with no filesystem access — and was dropped.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/payout-script.ts
jonybur: Fixed in `2cc7275d4`. `payout-script.ts` is now on all six critical-path surfaces.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/payout-script.ts
jonybur: Fixed in `2cc7275d4`. P2WPKH validation now uses the explicit compressed wallet key. It does not depend on PoP witness arity.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/payout-script.ts
jonybur: Fixed in `2cc7275d4`. The API now requires the raw wallet key, checks its x-coordinate against the PoP key, and uses its explicit parity for P2WPKH.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/payout-script.ts
jonybur: Fixed in `2cc7275d4`. The P2WPKH shape check now uses `P2WPKH_SCRIPT_PREFIX` and `HASH160_BYTES`.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts
jonybur: Fixed in `2cc7275d4`. The timeout rationale and uint8 constant are restored. The manager now reuses `NO_REFERRAL_CODE`.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts
jonybur: Fixed in `076875766` and `2cc7275d4`. Tests cover malformed bytes32 and uint8 values, off-curve keys, and different-x keys before RPC calls.

packages/babylon-ts-sdk/scripts/check-packed-consumer.js
jonybur: Fixed in `2cc7275d4`. The packed ESM and CommonJS checks now keep the internal payout assertion out of the public API.

jonybur: Fixed in `2cc7275d4`. The packed ESM and CommonJS checks now keep the internal payout assertion out of the public API.

packages/babylon-ts-sdk/package.json
jonybur: Fixed in `2cc7275d4`. The packed-consumer install now runs as a separate PR and release CI step after build.

packages/babylon-ts-sdk/src/tbv/core/index.ts
jonybur: Fixed in `2cc7275d4`. The core barrel keeps the existing lenient utility. The ETH-only subpath keeps the strict parser.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/__tests__/eth-import-boundary.test.ts
jonybur: Fixed in `2cc7275d4`. The duplicate source scanner is removed. `check-eth-build.js` remains the single boundary gate for emitted modules.

jonybur:
Addressed the remaining review-body items in `2cc7275d4`.

- `SECURITY.md` now lists only the pending script path.
- The transaction helper header now describes the current file.
- PoP key normalization now accepts uppercase `0X`.

gbarkhatov (commented):
@jonybur merge conflicts

PR #2349, opened 2026-08-27
ci(release): stop the release test run from failing at random

jonybur:
## What

Makes the automated test run that guards every release stop failing for no reason.

## Why

Four releases in the last two days stopped part-way because a test was reported as
failed when nothing was wrong with it. Each time the work had already been approved
and merged, so someone had to notice the red mark and start the release again by
hand. Today that happened to the Bitcoin engine release and it had to be re-run.

The cause is that the release machine is asked to run every part of the project's
test suite at the same time on a small machine. Tests then wait for a free slot
rather than for the thing they are testing, and the clock they are measured against
runs out. It never shows up on a pull request, because a pull request only runs the
tests for the parts that changed.

Nothing was ever wrong with the code being tested.

## How

- Tell the release machine to work through the project one part at a time, so each
  part gets the whole machine. Building and checking still run in parallel.
- Give the two test suites that ran out of time a longer allowance, so a slow
  machine reports a slow test rather than a broken one.

Verified: both suites pass locally at the same counts CI reports, and the three
workflow files parse.

### DIFF
--- a/.github/workflows/package-release.yml
+++ b/.github/workflows/package-release.yml
@@ -72,6 +72,12 @@ jobs:
 
       - name: Test all
         run: pnpm exec nx run-many --target=test --all
+        env:
+          # One project at a time. nx defaults to 3, and each project's vitest
+          # then spawns a worker per core, which oversubscribes the 4-vCPU runner
+          # and trips vitest's 5s per-test timeout. PRs escape it by testing only
+          # affected projects; `--all` on main does not.
+          NX_PARALLEL: "1"
 
       # Must stay last for its event: the release resolves `workspace:*` deps to
       # concrete versions on disk, which makes the specifiers recorded in
--- a/.github/workflows/service-release-vault.yml
+++ b/.github/workflows/service-release-vault.yml
@@ -50,6 +50,10 @@ jobs:
 
       - name: Test
         run: pnpm run test
+        env:
+          # See package-release.yml: `--all` on a 4-vCPU runner oversubscribes
+          # at nx's default of 3 projects and trips vitest's per-test timeout.
+          NX_PARALLEL: "1"
 
       - name: Build
         run: pnpm run build
--- a/packages/babylon-ts-sdk/vitest.config.ts
+++ b/packages/babylon-ts-sdk/vitest.config.ts
@@ -1,15 +1,14 @@
 import { defineConfig } from "vitest/config";
 
+import { TEST_TIMEOUT_MS } from "../../vitest.shared";
+
 export default defineConfig({
   test: {
     globals: true,
     environment: "node",
     setupFiles: ["./src/test/setup.ts"],
     exclude: ["**/node_modules/**", "**/dist/**"],
-    // 5s (vitest's default) is too tight for a suite whose tests dynamically
-    // import heavy mocked packages on a shared CI runner. Raised so a slow
-    // runner reports a slow test rather than a failed one.
-    testTimeout: 20_000,
+    testTimeout: TEST_TIMEOUT_MS,
     // Use forked processes for test execution to handle WASM initialization properly.
     // WASM loading is stateful - once initialized in a process, it remains loaded.
     // The babylon-tbv-rust-wasm package uses a singleton initWasm() pattern that
--- a/packages/babylon-ts-sdk/vitest.config.ts
+++ b/packages/babylon-ts-sdk/vitest.config.ts
@@ -6,6 +6,10 @@ export default defineConfig({
     environment: "node",
     setupFiles: ["./src/test/setup.ts"],
     exclude: ["**/node_modules/**", "**/dist/**"],
+    // 5s (vitest's default) is too tight for a suite whose tests dynamically
+    // import heavy mocked packages on a shared CI runner. Raised so a slow
+    // runner reports a slow test rather than a failed one.
+    testTimeout: 20_000,
     // Use forked processes for test execution to handle WASM initialization properly.
     // WASM loading is stateful - once initialized in a process, it remains loaded.
     // The babylon-tbv-rust-wasm package uses a singleton initWasm() pattern that
--- a/vitest.shared.ts
+++ b/vitest.shared.ts
@@ -1,0 +1,9 @@
+/**
+ * Per-test timeout shared by the workspace's node-side vitest suites.
+ *
+ * 5s (vitest's default) is too tight for a suite whose tests dynamically import
+ * heavy mocked packages on a shared CI runner. Raised so a slow runner reports
+ * a slow test rather than a failed one. Kept in one place so the suites that
+ * need it cannot drift apart.
+ */
+export const TEST_TIMEOUT_MS = 20_000;
--- a/.github/workflows/service-release-simple-staking.yml
+++ b/.github/workflows/service-release-simple-staking.yml
@@ -34,6 +34,10 @@ jobs:
 
       - name: Test
         run: pnpm run test
+        env:
+          # See package-release.yml: `--all` on a 4-vCPU runner oversubscribes
+          # at nx's default of 3 projects and trips vitest's per-test timeout.
+          NX_PARALLEL: "1"
 
       - name: Build
         run: pnpm run build
--- a/packages/babylon-ledger-vault-signer/vitest.config.ts
+++ b/packages/babylon-ledger-vault-signer/vitest.config.ts
@@ -5,6 +5,10 @@ export default defineConfig({
     globals: true,
     environment: "node",
     exclude: ["**/node_modules/**", "**/dist/**"],
+    // 5s (vitest's default) is too tight for a suite whose tests dynamically
+    // import heavy mocked packages on a shared CI runner. Raised so a slow
+    // runner reports a slow test rather than a failed one.
+    testTimeout: 20_000,
     // Only with an emulator attached: the e2e files share ONE Speculos device,
     // and an APDU arriving mid-UX-flow answers 0x6901 (sdk `status_words.h:56`).
     // Predicate matches the suites' own skipIf, which treats "" as unset.
--- a/packages/babylon-ledger-vault-signer/vitest.config.ts
+++ b/packages/babylon-ledger-vault-signer/vitest.config.ts
@@ -1,14 +1,13 @@
 import { defineConfig } from "vitest/config";
 
+import { TEST_TIMEOUT_MS } from "../../vitest.shared";
+
 export default defineConfig({
   test: {
     globals: true,
     environment: "node",
     exclude: ["**/node_modules/**", "**/dist/**"],
-    // 5s (vitest's default) is too tight for a suite whose tests dynamically
-    // import heavy mocked packages on a shared CI runner. Raised so a slow
-    // runner reports a slow test rather than a failed one.
-    testTimeout: 20_000,
+    testTimeout: TEST_TIMEOUT_MS,
     // Only with an emulator attached: the e2e files share ONE Speculos device,
     // and an APDU arriving mid-UX-flow answers 0x6901 (sdk `status_words.h:56`).
     // Predicate matches the suites' own skipIf, which treats "" as unset.


jonybur:
@greptile-apps

vitest.shared.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Per-test timeout shared by the workspace's node-side vitest suites.
+ *
+ * 5s (vitest's default) is too tight for a suite whose tests dynamically import
+ * heavy mocked packages on a shared CI runner. Raised so a slow runner reports
+ * a slow test rather than a failed one. Kept in one place so the suites that
+ * need it cannot drift apart.
+ */
+export const TEST_TIMEOUT_MS = 20_000;
```
gbarkhatov, line 9: **Not an nx input — editing this file re-runs nothing and invalidates no cache.**

`nx.json:4` defines `default: ["{projectRoot}/**/*", "sharedGlobals"]` with `sharedGlobals: []`, and neither package's `test` target declares its own `inputs`. A workspace-root file matches neither term, so it is outside every project's hash.

```
$ nx show projects --affected --files=vitest.shared.ts
(empty)

$ nx show projects --affected --files=packages/babylon-ts-sdk/vitest.config.ts
@babylonlabs-io/ts-sdk
@babylonlabs-io/ledger-vault-signer
@babylonlabs-io/wallet-connector
@services/simple-staking
@services/vault
```

Two consequences: a future PR that tunes only `TEST_TIMEOUT_MS` runs **zero** tests on `verify.yml:94`'s `nx affected --target=test`, and subsequent runs replay a cached pass computed under the old value. The change ships unverified.

This also defeats the docstring directly above it — "kept in one place so the suites that need it cannot drift apart" is the one place nx cannot see.

Compounding it, the file sits outside every other quality gate too:
- **Not linted** — `eslint vitest.shared.ts` → *"File ignored because no matching configuration was supplied"*; root `eslint.config.mjs:15` matches only `**/package.json`.
- **Not typechecked** — root tsconfig is `files: []` + references; `tsc --listFiles` never includes it.
- **Not format-checked** — package `format` scripts are `prettier --check "src/**/*.ts"`; the root has none.

Minimum fix, `nx.json:10`:

```json
"sharedGlobals": ["{workspaceRoot}/vitest.shared.ts"]
```

Better, and it closes the gate problem in the same move: make it a private workspace package the way `tools/eslint-config` → `@internal/eslint-config` already works (both of these packages already depend on it). nx then tracks it through the project graph instead of a hand-maintained glob, and it lands inside a project's lint/typecheck/format scope. Worth noting `../../vitest.shared` is currently the only import in the repo that escapes a package root.

vitest.shared.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Per-test timeout shared by the workspace's node-side vitest suites.
```
gbarkhatov, line 2: **Scope claim vs. wiring: `services/vault` reproduces this exact failure and wasn't included.**

The docstring says "the workspace's node-side vitest suites," but only 2 of the 6 vitest projects are wired. `services/vault`, `core-ui` and `wallet-connector` remain on the 5s default (`tools/release` too, and it *is* node-side — so the wording overstates either way).

`services/vault` is the one that matters, and it isn't hypothetical. `VaultWalletConnectionProvider.disabledWallets.test.tsx` uses `vi.resetModules()` + `await import(...)` — precisely the "tests dynamically import heavy mocked packages" mechanism this docstring describes — with no explicit per-test timeout. Measured on the same 10-core machine:

| condition | `disables onekey at the connector…` | result |
|---|---|---|
| idle (2 runs) | 1872 ms / 2222 ms | pass |
| ~3.4× oversubscribed | 5005 ms | **`Error: Test timed out in 5000ms`** |

That is the PR's failure mode verbatim, in a project on the same release path — `service-release-vault.yml:52` runs the root `--all` script, so this suite executes in every vault deploy.

Fair caveat: the four observed CI failures were `dmkSession.test.ts` (ledger-vault-signer) and `bip322Verify.test.ts` (ts-sdk), both of which this PR *does* cover. So vault is a predicted next failure rather than an observed one — but it has the least headroom of anything left at 5s, and `NX_PARALLEL=1` reduces contention without eliminating it (see the ts-sdk comment). Adding `testTimeout: TEST_TIMEOUT_MS` to `services/vault/vitest.config.ts` is a one-line extension of what's already here.

Minor, while you're in this file: `hookTimeout` stays at its 10s default while `testTimeout` goes to 20s, so a setup hook now gets less budget than the test it sets up. No hook timeout appears in any CI log, so it isn't urgent — but ts-sdk has ~20 `beforeAll(async () => await initializeWasmForTests())` hooks, which is where it would surface first. Cheap to add a `HOOK_TIMEOUT_MS` alongside.

Also worth one measured sentence in the comment: `20_000` isn't tied to any observation. The numbers above (~2.2s idle, ~5s at 3.4× oversubscription) justify it well — they just aren't written down anywhere.

packages/babylon-ts-sdk/vitest.config.ts
```diff
@@ -1,11 +1,14 @@
...
 
+import { TEST_TIMEOUT_MS } from "../../vitest.shared";
+
 export default defineConfig({
   test: {
     globals: true,
     environment: "node",
     setupFiles: ["./src/test/setup.ts"],
     exclude: ["**/node_modules/**", "**/dist/**"],
+    testTimeout: TEST_TIMEOUT_MS,
```
gbarkhatov, line 11: **Adjacent, and directly load-bearing for this PR: the `poolOptions` block below is dead config.**

Lines 17-22 set `pool: "forks"` + `poolOptions: { forks: { singleFork: true } }`, and the comment on lines 12-16 claims this "ensures WASM is initialized once and shared consistently across all test files."

vitest 4 removed `poolOptions`. Neither identifier exists anywhere in the installed runtime:

```
$ grep -ro 'poolOptions' node_modules/.pnpm/vitest@4.0.9*/node_modules/vitest | wc -l
0
$ grep -ro 'singleFork'  node_modules/.pnpm/vitest@4.0.9*/node_modules/vitest | wc -l
0
```

Same zero across `@vitest/runner` and the bundled `.d.ts`. `pool: "forks"` alone is still valid; the nested object is silently ignored. v4 replaced it with top-level `maxWorkers` / `fileParallelism`. And `git log -S singleFork` puts it in `b742eff7f`, when the package already pinned `vitest ^4.0.8` — so it has never taken effect.

Why it matters *here* rather than as a drive-by: ts-sdk is 107 test files with `maxWorkers` unbounded, which makes it the largest worker fan-out in the release run — the opposite of what the comment asserts. So `NX_PARALLEL=1` takes ~3 projects × N workers down to 1 project × N workers: it removes **cross-project** oversubscription only, while the intra-project lever named in `package-release.yml:77` ("each project's vitest then spawns a worker per core") stays unset. No config in the workspace sets `maxWorkers`, `fileParallelism` (outside ledger's Speculos case) or `VITEST_MAX_WORKERS`.

Replacing it with the v4 spelling — `fileParallelism: false` or `maxWorkers: 1` — would restore the documented WASM-singleton intent *and* bound the fan-out at its source. At minimum it shouldn't be left in place, under its current comment, by the one PR specifically about vitest worker oversubscription.

Nothing catches this today: `vitest.config.ts` is outside `tsconfig.lib.json`'s `include: ["src/**/*.ts"]` and ESLint ignores it, so `defineConfig`'s excess-property check never fires. That's the same blind spot as the `vitest.shared.ts` comment.

.github/workflows/package-release.yml
```diff
@@ -72,6 +72,12 @@ jobs:
 
       - name: Test all
         run: pnpm exec nx run-many --target=test --all
+        env:
+          # One project at a time. nx defaults to 3, and each project's vitest
+          # then spawns a worker per core, which oversubscribes the 4-vCPU runner
+          # and trips vitest's 5s per-test timeout. PRs escape it by testing only
+          # affected projects; `--all` on main does not.
```
gbarkhatov, line 79: **The "PRs escape it" premise doesn't hold, and `verify.yml` was left uncovered.**

nx caps concurrency at 3 regardless of how many projects are selected, so "only affected projects" doesn't imply less contention — it only implies fewer *total* tasks. PR runs routinely select nearly the whole graph:

```
33100579901 -> Running target test for 6 projects and 7 tasks they depend on
33098026600 -> Running target test for 5 projects and 7 tasks they depend on
33077201687 -> Running target test for 5 projects and 7 tasks they depend on
```

versus 7-of-7 for `--all` on main. 6-at-parallel-3 is the same scheduler pressure as 7-at-parallel-3 — a single ts-sdk change alone affects 5 projects. `verify.yml:89` ("Test affected") runs on the same `ubuntu-latest` runner class with no `NX_PARALLEL`, so the flake this PR fixes stays reachable on the PR path.

Either give `verify.yml:94` the same treatment, or reword this justification — as written it states something the run logs contradict, which is worse than saying nothing.

Smaller thing in the same block: line 78 says "trips vitest's 5s per-test timeout," but this PR raises that to 20s for two of the six suites. It's stale the moment it lands. Phrase it as the default the *remaining* suites still sit at.

.github/workflows/service-release-vault.yml
```diff
@@ -50,6 +50,10 @@ jobs:
 
       - name: Test
         run: pnpm run test
+        env:
+          # See package-release.yml: `--all` on a 4-vCPU runner oversubscribes
+          # at nx's default of 3 projects and trips vitest's per-test timeout.
+          NX_PARALLEL: "1"
```
gbarkhatov, line 56: **This also serializes the build graph here — a cost the PR body doesn't mention.**

`test` has `dependsOn: ["^build"]` (`nx.json`), and in both service workflows the `Test` step (line 51) runs *before* the `Build` step (line 58). So the 7 build tasks execute inside this step from a cold cache, and `NX_PARALLEL=1` serializes those too — not just the tests. From run `33071432939`:

```
12:21:31  NX  Running target test for 7 projects and 7 tasks they depend on:
12:21:34  ✅ nx run @babylonlabs-io/babylon-tbv-rust-wasm:build
12:21:59  ✅ nx run @babylonlabs-io/ts-sdk:build
12:23:41  ✅ nx run @babylonlabs-io/wallet-connector:build
```

That step was 293s, roughly half of it builds. Serialized, expect materially longer here and ~+45-60% on `package-release` (221s → ~330s, from the per-project durations in run `33071433038`). Since `lint_test` gates `s3_publish`, deploys slip by the same amount. The "Verified" line in the body carries no timing — worth measuring on the first main push.

A narrower knob exists if it does hurt: `targetDefaults.test.parallelism: false` in `nx.json` (`nx-schema.json:565`, enforced in `tasks-schedule.js` — a task with `parallelism: false` runs alone). That gets the same test isolation from **one** place, leaves `build` and `lint` 3-wide, and removes all three copy-pasted env blocks. Worth considering as the primary mechanism rather than the per-workflow env — the three comment blocks have no mechanism keeping them in sync, and two of them already point at a `package-release.yml` comment that this PR makes partly stale.

gbarkhatov (approved):
Approving — the mechanism is sound and the diagnosis holds up. Findings below are follow-ups, not blockers.

**Verified correct, for the record:**

- `NX_PARALLEL` is genuinely read by nx 21.3.7 — `readParallelFromArgsAndEnv` (`nx/src/command-line/yargs-utils/shared-options.js:306-323`) falls through to `concurrency(process.env.NX_PARALLEL || '3')`, and since no step passes `--parallel`, `args['parallel']` is `undefined` and the env branch fires. `"1"` → `1`. `nx.json` sets no `parallel`, and env would win anyway (`run-command.js:825` is `??=`).
- Both service workflows have no `working-directory`/`defaults`, so `pnpm run test` resolves to the root `package.json:13` → `nx run-many --target=test --all --verbose`. The env var applies in all three places.
- All three edits are step-level `env` on `run:` steps (not `uses:`), correctly indented; no job- or workflow-level `env` shadows them.
- The relative `../../vitest.shared` import resolves through vite's esbuild config loader; no typecheck or lint fallout, and nothing ships (`files: ["dist", ...]`, `npm pack --dry-run` clean).
- The ledger Speculos e2e tests pass explicit per-test timeouts (30s–120s), so the new 20s global does not shorten them.
- **The "four releases" claim checks out exactly.** Two failures are hidden behind re-runs and don't appear in `gh run list`: `33068011140` and `33068011153` are both `run_attempt=2` on `84697bb5a` ("load the Bitcoin engine only when it is needed"), matching the body's "today that happened to the Bitcoin engine release". Four distinct pushes, all failing on the **Test** step: `32956607345`, `32964120792`, `33068011140`/`33068011153` (attempt 1), `33080930611`. Worth stating plainly because the six *other* red release runs on 8/26 (`32958086381`, `32956684253`, …) failed on **Build all** with `NX Could not execute command because the task graph has a circular dependency` — a real build-graph bug fixed separately by #2328. Anyone skimming `gh run list` will see nine red runs in three days and mis-attribute them; a line in the body ruling those out would help.

**Findings:**

1. `vitest.shared.ts` is invisible to Nx — editing it marks nothing affected and busts no cache. *(inline, `vitest.shared.ts:9`)*
2. Coverage gap: `services/vault` reproduces this exact failure under load and was not wired up. *(inline, `vitest.shared.ts:2`)*
3. ts-sdk's `poolOptions.forks.singleFork` is dead config in vitest 4 — and it's the intra-project lever this PR is about. *(inline, `packages/babylon-ts-sdk/vitest.config.ts:11`)*
4. The comment's "PRs escape it" premise is contradicted by the run logs, and `verify.yml` is left uncovered. *(inline, `.github/workflows/package-release.yml:79`)*
5. `NX_PARALLEL=1` also serializes the build graph in the two service workflows — an undisclosed few minutes each. *(inline, `.github/workflows/service-release-vault.yml:56`)*

**One caveat on landing:** all three workflows trigger only on push-to-`main`/tags, so this PR's own CI never exercises the modified path, and `workflow_dispatch` isn't free here (it publishes). The first push to `main` is the real test — worth watching that run rather than assuming green.

PR #2362, opened 2026-08-27
fix(vault): tell the user when part of their balance is unconfirmed

jonybur:
Screenshot 2026-08-27 at 16 30 21

The deposit form shows a balance that can be far lower than the one in the user's wallet, with nothing on screen to explain why.

## What

Show the "pending confirmation" notice whenever the connected address holds unconfirmed funds. It used to appear only when the confirmed balance was exactly zero.

## Why

The form counts confirmed funds only. The notice exists to explain that gap, but its condition required a zero confirmed balance, so anyone holding a mix of confirmed and unconfirmed funds saw no explanation.

Found by hand on signet. An address with 0.00329262 confirmed and 0.025 unconfirmed reads 0.00315006 in the deposit form and 0.02829262 in the wallet - about ten times lower. Nothing on the screen accounted for the difference.

## How

`hasUnconfirmedBalance` is true whenever unconfirmed funds exist, and drives the notice. The old `hasUnconfirmedBalanceOnly` stays, and keeps its narrower meaning: the confirmed balance is zero, so the depositable maximum is zero and the Max tooltip is suppressed because it has no fee buffer to describe.

No change to any balance, fee or amount that the form calculates.

## Verification

- `npx vitest run` in services/vault - 3606 passed, 302 files
- `npx tsc -p tsconfig.lib.json --noEmit` - clean
- `npx eslint` on every changed file - clean
- The new test fails against the previous condition and passes against this one
- Checked by hand in the running app against the signet address above


services/vault/src/components/simple/DepositForm.tsx
```diff
@@ -313,12 +319,12 @@ export function DepositForm({
...
-  // would sit — empty at a zero balance). The InfoIcon is wrapped with an
-  // attach-to-children Hint so the markup stays inline-valid inside the slider's
-  // right cell (a bare Hint would nest a div inside a span).
-  const pendingConfirmationField = hasUnconfirmedBalanceOnly ? (
+  // When unconfirmed funds exist, show a "pending confirmation" note in the
+  // slider's left slot, where the USD value would sit — it is only rendered
+  // while no amount is entered, so the two never compete. The InfoIcon is
+  // wrapped with an attach-to-children Hint so the markup stays inline-valid
+  // inside the slider's cell (a bare Hint would nest a div inside a span).
+  const pendingConfirmationField = hasUnconfirmedBalance ? (
```
gbarkhatov: The widened flag lands in a slot that hides it as soon as the user acts. `leftField.value` at `:427-431` is `!hasAmount ? (pendingConfirmationField ?? zeroUsdValue) : usdValue`, with `hasAmount = !!amount && amount !== "0"` at `:356`.

Under the old condition that gate was inert: a zero confirmed balance means `maxDepositSats` is `0n`, so `applyMaxAmount` (`useDepositPageForm.ts:668`) writes `"0"`, `hasAmount` stays false, and the notice never left the screen. The new audience has a spendable balance, so the first keystroke — or the Max click, which is how a depositor usually discovers the shortfall — replaces the notice with the USD value while `Balance: 0.00315006 BTC` stays on the right. The ~10x gap this PR sets out to explain is back on screen unexplained for the whole of amount entry.

What ships is therefore narrower than the description's "whenever the connected address holds unconfirmed funds" — it is "whenever unconfirmed funds exist and the amount field is empty". Either give the notice a slot that doesn't compete with the USD value, or keep this one and say so explicitly; in the second case the two new doc comments (here at `:40-43`, and `useDepositPageForm.ts:101-103`, "drives" / "shows the notice") want the qualifier, since the flag is necessary but not sufficient.

services/vault/src/components/simple/DepositForm.tsx
```diff
@@ -313,12 +319,12 @@ export function DepositForm({
...
     <span className="inline-flex items-center gap-1 text-accent-secondary">
       {COPY.deposit.form.pendingConfirmationNotice(
         `${Number(depositService.formatSatoshisToBtc(unconfirmedBalance))} ${btcConfig.coinSymbol}`,
```
gbarkhatov: `formatSatoshisToBtc` (`babylon-ts-sdk/src/tbv/core/primitives/utils/bitcoin.ts:278-287`) already returns a trimmed decimal string. The `Number()` round-trip adds nothing and re-introduces exponential notation below 1e-6: `Number("0.00000099")` stringifies to `9.9e-7`, so any unconfirmed total under 100 sats renders as `9.9e-7 BTC pending confirmation`.

That was unreachable before, since it required a zero confirmed balance, and is reachable now for any wallet holding dust in the mempool. core-ui documents this exact hazard for the same reason and formats with `toLocaleString` instead (`AmountSlider.tsx:12-20`). Dropping the `Number()` wrapper is enough here.

Related, and a judgement call rather than a defect: there is no dust threshold, so a single 1-sat unconfirmed output now puts a coloured notice on a fully funded form.

services/vault/src/hooks/deposit/useDepositPageForm.ts
```diff
@@ -353,12 +358,23 @@ export function useDepositPageForm(): UseDepositPageFormResult {
...
-  // UTXOs. Keyed on the raw confirmed balance (not the spendable `btcBalance`)
-  // so the notice never fires when confirmed funds exist but are hidden as
-  // inscriptions — that is a different reason for a zero spendable balance.
+  // True whenever the address holds unconfirmed (in-mempool) funds. The deposit
+  // form uses this to explain why the wallet shows a balance the app does not —
+  // the app only counts confirmed UTXOs. Confirmed funds elsewhere on the
+  // address do not make that gap any smaller, so this must NOT require a zero
+  // confirmed balance: a wallet holding 0.003 confirmed and 0.025 unconfirmed
+  // reads ~10x lower in the form than in the wallet, with nothing to explain it.
+  const hasUnconfirmedBalance = useMemo(
```
gbarkhatov: Suggestion, non-blocking: this flag may not need to exist. `unconfirmedBalance` is already on `DepositAmountState` and destructured in `DepositForm` at `:213`, two lines above the flag, and its only consumer is `unconfirmedBalance > 0n` at `DepositForm.tsx:327`. Deriving it at that call site removes this memo, both interface fields, the `SimpleDeposit` pass-through and the fixture line — the change collapses to one expression in one file.

The counter-argument is symmetry: keeping both conditions side by side here documents them together. Worth a deliberate choice either way. It does not apply to `hasUnconfirmedBalanceOnly`, which genuinely needs `confirmedBalance`, and the component never receives that.

If you do inline it, the two new hook tests go away with it — which leaves the render test suggested on the fixture line as the only thing pinning the behaviour.

services/vault/src/hooks/deposit/__tests__/useDepositPageForm.test.tsx
```diff
@@ -587,6 +587,42 @@ describe("useDepositPageForm", () => {
...
+      } as unknown as ReturnType<typeof useUTXOs>);
+
+      const { result } = renderHook(() => useDepositPageForm(), { wrapper });
+
+      expect(result.current.hasUnconfirmedBalance).toBe(false);
+    });
+
     it("does not flag hasUnconfirmedBalanceOnly when confirmed funds exist but are all inscriptions", () => {
       // Spendable balance is zero because the only confirmed UTXO is an
       // inscription (excluded from availableUTXOs), yet confirmed funds exist.
```
gbarkhatov, line 592: This comment now says the opposite of what ships. The fixture below is `confirmedBalance: 300000n, unconfirmedBalance: 50000n`, which makes `hasUnconfirmedBalance` true under the new condition, so the notice renders in exactly the scenario the comment says it must not. Nothing fails, because the assertion only covers `hasUnconfirmedBalanceOnly` — so the comment just rots, and the next reader takes it as the rule and narrows the condition back.

There is a behavioural question underneath it, not only stale wording: in this state the balance reads 0 because 0.003 is locked in inscriptions, and the form now attributes that zero to 0.0005 of pending funds. The Max tooltip ("excludes inscription UTXOs") is the only thing saying otherwise. Whichever way you decide, please record it in this comment.

services/vault/src/components/simple/__tests__/DepositForm.suggestedAmount.test.tsx
```diff
@@ -51,6 +51,7 @@ function amountState(
     amountSats: 0n,
     btcBalance: 100_000_000n,
     unconfirmedBalance: 0n,
+    hasUnconfirmedBalance: false,
```
gbarkhatov: This fixture line is the whole of the component-side test change, and the behaviour that actually changed is which flag gates the notice — which lives in the JSX at `DepositForm.tsx:327`, not in the hook. Swapping that line back to `hasUnconfirmedBalanceOnly` keeps every test in the repo green.

This file already renders the full form with an overridable `amountState`, so two assertions are cheap: the notice renders when confirmed and unconfirmed funds are both present, and it is hidden once an amount is entered. Writing the second one surfaces the `!hasAmount` gate immediately.

Two smaller things while here. The description's "the new test fails against the previous condition" holds only trivially — the flag is new, so there is no previous condition for those tests to fail against; they assert `2_500_000n > 0n` and `0n > 0n` through a hook. And the existing positive test at `:557` was not extended with `expect(result.current.hasUnconfirmedBalance).toBe(true)`, which is where that assertion naturally sits; the added negative at `:610` is a near-clone of `:573` minus one assertion, and writes `500000` where its own sibling three lines up uses `2_500_000n`.

gbarkhatov (dismissed):
Approving — the premise holds, nothing here touches a CLAUDE.md critical path, and no balance, fee or amount the form calculates changes. I checked that `spendableMempoolUTXOs` (`useUTXOs.ts:172-180`) is confirmed-only despite the name, and that `unconfirmedBalance` and `btcBalance` are disjoint by construction — everything spendable descends from `confirmedUTXOs` (`useUTXOs.ts:65-91`) — so the widened flag cannot invent a gap that isn't there. `hasUnconfirmedBalanceOnly` is not left dangling: `DepositForm.tsx:341` still reads it for the Max tooltip. The `right slot` -> `left slot` edit corrects a comment that was already wrong on main. Both touched test files pass locally, 43 tests.

Five comments inline. The first is the one I would resolve before merge: the notice is gated behind an empty amount field, so for the mixed-balance case this PR exists to cover, it survives only until the first keystroke. The third is a suggestion rather than a request.

Two asides not worth their own inline comments. The notice has no `role="status"` / `aria-live`, so it is never announced when the UTXO query resolves — pre-existing, but this PR turns a rare state into a common one, and the sibling notice at `DepositForm.tsx:461-464` already does it. And row 3 of the slider now carries the long left string alongside a non-zero balance, the Max tooltip and the Max pill at once, a combination the old condition never produced; the row has no wrap or truncate, so it is worth an eyeball at 375px — I did not render it, so treat that as a check rather than a claim.

services/vault/src/components/simple/DepositForm.tsx
jonybur: Fixed in 9f7dc600. The pending-funds notice now has a persistent status slot below the amount slider. It stays visible after input and Max.

services/vault/src/components/simple/DepositForm.tsx
jonybur: Fixed in 9f7dc600. The notice now uses the exact formatter string. One satoshi renders as `0.00000001 BTC`.

services/vault/src/hooks/deposit/useDepositPageForm.ts
jonybur: Applied in 9f7dc600. The pass-through flag is gone. `DepositForm` derives the condition from `unconfirmedBalance`, and component coverage pins the behavior.

services/vault/src/hooks/deposit/__tests__/useDepositPageForm.test.tsx
jonybur: Fixed in 9f7dc600. The comment now records the inscription case. The unconfirmed notice renders independently, while the Max tooltip still explains the inscription adjustment.

jonybur: Fixed in 9f7dc600. The comment now records the inscription case. The unconfirmed notice renders independently, while the Max tooltip still explains the inscription adjustment.

services/vault/src/components/simple/__tests__/DepositForm.suggestedAmount.test.tsx
jonybur: Fixed in 9f7dc600. The component test now combines confirmed funds, one unconfirmed satoshi, and a non-empty amount. It asserts the persistent status notice.

jonybur:
@greptileai

jonybur:
@greptileai

services/vault/src/components/simple/DepositForm.tsx
```diff
@@ -432,6 +425,15 @@ export function DepositForm({
           onMaxClick={onMaxClick}
           inputClassName="h-10 w-auto rounded-lg bg-primary-contrast px-4 [field-sizing:content]"
         />
+        {pendingConfirmationField && (
+          <p
+            className="flex flex-wrap items-center gap-1 text-sm"
+            role="status"
```
gbarkhatov: The live region and its content are mounted in the same commit, so the announcement this exists for will not fire. The `<p role="status">` renders only when `pendingConfirmationField` is non-null, so the node and its text enter the DOM together — and a live region has to be in the DOM *before* its content changes for the change to be announced reliably. The scenario this PR was written for is exactly that case: the user sits on the deposit form, a mempool transaction lands, `useUTXOs` repolls, and the notice appears visually while nothing is announced.

This repo already solves it and writes down why — `UtxoSplitSelectorV3.tsx:69-83` renders its `role="status"` unconditionally and empties the text instead of unmounting, with the rationale in the comment above it. Either follow that pattern with an always-mounted `sr-only` announcer alongside the visual line, or render this `<p>` unconditionally and move the condition onto its children.

While here, and only if you touch it: the wrapper duplicates the layout of its single child — `items-center` and `gap-1` are set on both this `<p>` and the `<span>` at `:317`, and `flex-wrap` can never fire with one child. `aria-live="polite"` is also implied by `role="status"`; harmless, and the sibling at `:458-459` does the same, so it is consistent either way.

services/vault/src/components/simple/__tests__/DepositForm.suggestedAmount.test.tsx
```diff
@@ -160,4 +167,17 @@ describe("DepositForm suggested amount", () => {
...
+
+  it("keeps the exact pending amount visible after amount entry", () => {
+    renderForm({
+      amount: "0.001",
+      amountSats: 100_000n,
+      btcBalance: 329_262n,
+      unconfirmedBalance: 1n,
+    });
+
+    expect(screen.getByRole("status")).toHaveTextContent(
```
gbarkhatov: This is a real behavioural assertion and it does pin the fix — with the `Number()` wrapper back it would read `1e-8` and fail. What is missing is the negative half. Every other test in the file leaves the default `unconfirmedBalance: 0n` and none of them asserts the notice is absent, so mutating `:316` to `unconfirmedBalance >= 0n` — which would put a permanent `0 BTC pending confirmation` line under the slider on every fully confirmed wallet — leaves the whole suite green. One `expect(screen.queryByRole("status")).not.toBeInTheDocument()` in the default state closes it.

Second, smaller: `getByRole("status")` is unqualified and `DepositForm` now has two `role="status"` nodes — this one and the max-vaults notice at `DepositForm.tsx:456-473`. It passes today only because `renderForm` never sets `twoVaultSplit`/`vaultCapUsage`. The day a fixture does, this fails with a "found multiple elements" error that points nowhere near the cause. `getByRole("status", { name: ... })` or a `queryAllByRole` filter would survive it.

services/vault/src/hooks/deposit/useDepositPageForm.ts
```diff
@@ -353,12 +353,12 @@ export function useDepositPageForm(): UseDepositPageFormResult {
     return BigInt(calculateBalance(availableUTXOs || []));
   }, [availableUTXOs]);
 
-  // True when the address has no confirmed funds at all but does have
-  // unconfirmed (in-mempool) funds. The deposit form uses this to explain why
-  // the wallet shows a balance the app does not — the app only counts confirmed
-  // UTXOs. Keyed on the raw confirmed balance (not the spendable `btcBalance`)
-  // so the notice never fires when confirmed funds exist but are hidden as
-  // inscriptions — that is a different reason for a zero spendable balance.
+  // The narrower case: no confirmed funds at all, so the depositable maximum is
```
gbarkhatov: "The narrower case:" was narrower than the `hasUnconfirmedBalance` memo that used to sit directly above it, and that memo was deleted in `9f7dc600a`. Nothing above defines a wider case any more — the preceding statement is the `btcBalance` memo — so a reader hits a comparative with no referent and goes looking for the other case.

Dropping the first three words fixes it: start at "No confirmed funds at all, so the depositable maximum is zero…". The rest of the comment is still accurate.

services/vault/src/components/simple/DepositForm.tsx
```diff
@@ -313,24 +312,20 @@ export function DepositForm({
...
-  ) : null;
+  const pendingConfirmationField =
+    unconfirmedBalance > 0n ? (
+      <span className="inline-flex items-center gap-1 text-accent-secondary">
+        {COPY.deposit.form.pendingConfirmationNotice(
+          `${depositService.formatSatoshisToBtc(unconfirmedBalance)} ${btcConfig.coinSymbol}`,
+        )}
+        <Hint
+          tooltip={COPY.deposit.form.pendingConfirmationTooltip}
+          attachToChildren
```
gbarkhatov, line 324: The comment that explained this prop went out with the slot it described. It said a bare `Hint` renders a `div`, which would be invalid in the inline context this sat in. The constraint did not go away — the notice now lives inside a `<p>` at `:429`, where a nested `div` is equally invalid — so the prop is now load-bearing and unmotivated, and the next person to touch this has no reason not to drop it. One line above the `Hint` saying `attachToChildren` is required because the container is a `<p>` is enough.

services/vault/src/components/simple/DepositForm.tsx
```diff
@@ -313,24 +312,20 @@ export function DepositForm({
...
-      <Hint
-        tooltip={COPY.deposit.form.pendingConfirmationTooltip}
-        attachToChildren
-      >
-        <InfoIcon size={16} className="text-accent-secondary" />
-      </Hint>
-    </span>
-  ) : null;
+  const pendingConfirmationField =
+    unconfirmedBalance > 0n ? (
```
gbarkhatov, line 316: Two small things about this condition, neither a defect.

The dust point I raised as an aside last time costs more now than it did. Under the old transient slot a 1-satoshi unconfirmed output flashed a coloured line only while the amount box was empty; with the notice made persistent it is a coloured status line under the slider for the whole session, on a fully funded form. The new test at `DepositForm.suggestedAmount.test.tsx:171-182` pins exactly that case as the intended behaviour, which is a fine decision to make — worth making deliberately rather than inheriting it from the move. If you do want a floor, gate on a threshold the form already knows rather than `> 0n`.

And `pendingConfirmationField` is now a misnomer: it is no longer an `AmountSlider` field, it is the notice body. `pendingConfirmationNotice` matches what it became.

services/vault/src/components/simple/__tests__/DepositForm.suggestedAmount.test.tsx
```diff
@@ -24,7 +24,14 @@ vi.mock("@/config", () => ({
 // and short-circuit the CTA state machine so the form renders deterministically.
 vi.mock("@/services/deposit", () => ({
   depositService: {
-    formatSatoshisToBtc: (sats: bigint) => (Number(sats) / 1e8).toString(),
+    formatSatoshisToBtc: (sats: bigint) => {
```
gbarkhatov: This mock is now a line-for-line reimplementation of `formatSatoshisToBtc` (`babylon-ts-sdk/src/tbv/core/primitives/utils/bitcoin.ts:277-286`) — same whole/fraction split, same `padStart(8, "0")`, same trailing-zero strip, minus the negative branch. Not a false green today: the component's own coercion is what the assertion at `:179` catches, so the test does its job. But `0.00000001` is now a property of this copy rather than of the formatter, and the copy drifts silently if the SDK ever changes its padding or rounding.

The module is only mocked wholesale to short-circuit `getDepositCtaState`, so an async factory that spreads `await vi.importActual("@/services/deposit")` and overrides just that one export keeps the real formatter in the path and deletes the duplicate.

gbarkhatov (dismissed):
Follow-up to my review at `6ffbdb1b7`, not a second opinion on the same code. Two commits since (`9f7dc600a`, `fe8331436`) moved the notice out of the slider and changed how it formats, so this pass settles a verdict for each of the five prior comments and then reviews only what changed.

Every verdict below was checked against the code at HEAD, not against the replies.

| # | Prior comment | Verdict | At HEAD |
|---|---|---|---|
| 1 | Notice killed by the `!hasAmount` gate on the first keystroke or Max click | **Fixed** | `DepositForm.tsx:428-436` — a persistent block below the slider, gated only on `unconfirmedBalance > 0n`; neither `hasAmount` nor `amountSats` is in that path, and `leftField` is back to the plain USD value at `:417` |
| 2 | `Number()` round-trip prints `9.9e-7 BTC` below 100 sats | **Fixed** | `DepositForm.tsx:319` — no coercion left anywhere on that string path |
| 3 | The flag need not exist | **Applied** | Repo-wide grep for `hasUnconfirmedBalance` returns only `hasUnconfirmedBalanceOnly`. The memo, both interface members, the `SimpleDeposit` destructure and pass-through and both fixtures are gone, and the two hook tests were deleted rather than left orphaned |
| 4 | Test comment says the opposite of what ships | **Fixed** | `useDepositPageForm.test.tsx:591-594` |
| 5 | Component test pinned a fixture, asserted no behaviour | **Fixed** | `DepositForm.suggestedAmount.test.tsx:171-182` |

The existing threads carry the original reasoning; nothing there is restated below.

One half-point on 4, not worth reopening the thread: the false claim is gone, but the replacement records only why `hasUnconfirmedBalanceOnly` stays false, not the other half of the decision — that in the inscription case the pending notice *does* fire, so the form now attributes a zero spendable balance to pending funds. The causality also reads backwards: the flag is false because `confirmedBalance` is non-zero; the Max tooltip surviving is the consequence, not the reason.

Both asides from the last review resolved themselves here. The notice now carries `role="status"` / `aria-live="polite"`, matching the sibling at `:458-459`; and the long string left slider row 3 entirely, so the 375px overflow concern in that row is moot. Whether the new full-width line wraps well at 375px I did not render — `visual` passed, which is weak evidence the snapshot set covers it.

**Six comments inline, none blocking.** The one I would actually act on is the first: the live region is mounted in the same commit as its content, which is the one arrangement that stops it being announced — and this repo already documents the fix in `UtxoSplitSelectorV3.tsx:69-83`. The rest are a missing negative test, a comment left dangling by the deleted memo, a deleted rationale, a naming drift, and a mock that now duplicates the SDK formatter.

Verified clean, so as not to re-litigate: no CLAUDE.md critical path is touched and no balance, fee or amount the form calculates changed; the user-facing strings are still in `copy.ts:679-682` and untouched; no `data-testid` was dropped; no animation, so the motion rules do not apply; and the removal in `9f7dc600a` left no dead field, mock, type member or orphaned test. All 13 checks pass, including `verify` and `visual`.

services/vault/src/components/simple/DepositForm.tsx
jonybur: Fixed in eac71fa48. The status region stays mounted. Only its content changes.

services/vault/src/components/simple/DepositForm.tsx
jonybur: Fixed in eac71fa48. The field is now named pendingConfirmationNotice. Every non-zero pending balance is useful to the user, including one satoshi, so this intentionally has no dust floor.

services/vault/src/hooks/deposit/useDepositPageForm.ts
jonybur: Fixed in eac71fa48. The stale comparative wording is removed.

services/vault/src/components/simple/__tests__/DepositForm.suggestedAmount.test.tsx
jonybur: Fixed in eac71fa48. The test imports the real deposit service and overrides only getDepositCtaState, so it covers the real satoshi formatter.

services/vault/src/components/simple/DepositForm.tsx
jonybur: Fixed in eac71fa48. The comment now explains that attachToChildren prevents the Hint div wrapper from becoming a child of the paragraph.

services/vault/src/components/simple/__tests__/DepositForm.suggestedAmount.test.tsx
jonybur: Fixed in eac71fa48. The positive test anchors the exact notice. A separate test verifies that the mounted status region stays empty when the pending balance is zero.

gbarkhatov (dismissed):
@jonybur lint error

gbarkhatov (approved):

jonybur:

Issue #2351, opened 2026-08-27
[eth-first] Require new approval after a wallet identity changes


jonybur:
**Epic:** #2228
**Parent:** #2230
**Status:** Blocks #2230 and #2232
**Origin:** Mixed. The live-session gap predates #2267. #2267 exposed it through identity receipts and skipped the default non-persistent mode.

## Problem

A confirmed session can stay approved after the account, wallet, key, network, or required-chain set changes.

## Fix

Keep the approved identity in memory for every session. Compare it with the live identity after each change. Persistence must affect reload only.

## Complete when

- [ ] Identity changes require approval again in both session modes.
- [ ] Required-chain changes use the same rule.
- [ ] Optional-wallet changes do not end a valid session.
- [ ] Tests use live account and network events.

jonybur:
Delivered by PR #2382, which reached main via #2363 (4f2681948).

PR #2363, opened 2026-08-27
fix(wallet-connector): bind confirmation to one active identity

### LINKED ISSUE #2350: [eth-first] Do not approve a wallet that changes during confirmation
**Epic:** #2228
**Parent:** #2230
**Status:** Blocks #2230 and #2232
**Origin:** Expanded by #2267. A narrower Bitcoin consent race existed before it.

## Problem

The app can approve wallet A, wait for a host callback, and then confirm wallet B. The same flow also allows double submit and close-while-confirming races.

## Fix

Use one wallet snapshot for the callbacks, receipt, and final confirmation. Stop if the wallet, account, key, or network changes.

## Complete when

- [ ] Every identity change during confirmation stops the flow.
- [ ] Double submit runs the callbacks once.
- [ ] Closing the dialog cannot confirm later.
- [ ] Tests fail against the merged #2267 behavior.

**Review:** https://github.com/babylonlabs-io/babylon-toolkit/pull/2267#discussion_r3872654862

jonybur:
## Outcome

Wallet confirmation approves one unchanged active identity.

## Change

- Capture the active wallet, provider identity, and configured network.
- Read the provider identity before each hook and stop if it changes.
- Release the attempt on close, hide, or unmount.
- Prevent duplicate submits and stale completions.

## Safety

- Fail closed before storage or host confirmation.
- Live identity checks use provider-returned addresses and public keys.
- Live network validation covers ETH and BTC.
- Fixed-network providers report their configured operational network because they have no independent network switch.
- BBN has no network getter. Its live account identity is still validated.
- Keep live-session revocation in #2351.
- Keep the callback snapshot immutable.

## Verification

- 217 wallet connector tests pass.
- The wallet connector build passes.
- Lint passes with no error.
- Independent review found no merge blocker.

## Links

Closes #2350


packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -28,14 +27,104 @@ function findPrimaryConnection(
...
+  const connections = collectConnections(connectors);
+  const authority = collectAuthority(connections);
+  const receipt = createConfirmationReceipt(connections, connectors);
+  const networks: Partial<Record<ChainId, string | number>> = {};
+
+  await Promise.all(
+    connections.map(async ({ chain, wallet }) => {
+      if (!wallet.provider) {
+        throw new Error(`Connected ${chain} wallet has no provider`);
+      }
```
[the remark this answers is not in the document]
jonybur: Fixed in 20ced7eb. Provider checks now apply only to ETH and BTC. A BBN wallet with no provider can continue. Tests cover this path.

jonybur:
@greptileai

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -28,14 +30,129 @@ function findPrimaryConnection(
...
 
-function collectConnections(
-  selectedWallets: Record<string, IWallet | undefined>,
-): WalletLifecycleConnection[] {
-  return Object.entries(selectedWallets).flatMap<WalletLifecycleConnection>(([chain, wallet]) =>
-    wallet?.account ? [{ chain: chain as ChainId, wallet, account: wallet.account }] : [],
+function collectConnections(connectors: Connectors): WalletLifecycleConnection[] {
+  return Object.entries(connectors).flatMap<WalletLifecycleConnection>(([chain, connector]) => {
+    const wallet = connector?.connectedWallet;
+    return wallet?.account ? [{ chain: chain as ChainId, wallet, account: { ...wallet.account } }] : [];
```
gbarkhatov, line 44: **Merge-blocking, I think:** the account/public-key arm of the identity gate can't fire in production.

This reads `connector.connectedWallet.account`, and so does every later comparison. But `Wallet.account` has exactly one writer in the whole package — `Wallet.connect()` at `core/Wallet.ts:58`, reachable only from `WalletConnector.connect()` (`WalletConnector.ts:47`). An in-extension account switch never goes through it: `ETHWalletProvider.tsx:214-236` and `BTCWalletProvider.tsx:230-274` both handle `accountsChanged` by writing React state and calling `onAddressChange`, never `connector.connect()`.

`createConfirmationSnapshot` reads only the *network* live (lines 100-103) — address and publicKeyHex stay cached. So when the user switches accounts in MetaMask while `acceptTermsOfService` is pending, `checkIdentity` compares the stale value against itself, mints a byte-identical receipt, and confirms account A while the wallet is on B. That's the race #2350's first criterion is meant to close.

The vault masks it host-side (`context/wallet/VaultWalletConnectionProvider.tsx:189,211` wires `onAddressChange: runWalletReset`), but that's a host courtesy, not this package's guarantee.

The fix looks local: resolve `getAddress()`/`getPublicKeyHex()` live here the same way the network already is, feed them through an override parameter mirroring the new `networks` one, and compare with the existing `sameIdentity` helper in `confirmationReceipt.ts:49`.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -28,14 +30,129 @@ function findPrimaryConnection(
...
+        const network =
+          chain === "ETH"
+            ? await (provider as IETHProvider).getChainId()
+            : await (provider as IBTCProvider).getNetwork();
+        if (network === undefined || network === null || !String(network)) {
+          throw new Error(WALLET_NETWORK_MISSING_ERROR_MESSAGE);
+        }
+
+        const liveReceipt = createConfirmationReceipt([candidate.connection], connectors, { [chain]: network });
+        if (liveReceipt !== candidate.receipt) {
```
gbarkhatov: This comparison can reduce to `config === config`, which makes it fail *open* in a gate whose whole point is failing closed.

`AppKitETHProvider.getChainId()` (`eth/appkit/provider.ts:351-364`) ends with `return this.config.chainId` when its cache is empty and wagmi reports no account chain. `ledger-vault`'s is `getNetwork = async () => this.network` (`btc/ledger-vault/provider.ts:1115`) — the value it was constructed with. In both cases `liveReceipt` is built from the same configured number this is compared against, so the branch can never be taken. Only UniSat/OKX/AppKit-BTC actually ask the wallet.

So "stop the attempt if the configured network changes" ships for some providers and is a no-op for others. Either require a genuinely live read on this path and assert that contract, or drop the live comparison and say plainly that only the configured network is guarded until #2351.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -58,93 +175,161 @@ export function WalletDialog({
...
 
         const primary = findPrimaryConnection(connections, requiredChainIds);
+        const checkIdentity = async () => {
+          if (!isCurrentAttempt()) return false;
+
+          const currentSnapshot = await createConfirmationSnapshot(() => connectorsRef.current, requiredChainIds);
+          if (!isCurrentAttempt()) return false;
+
+          if (
+            currentSnapshot.receipt !== confirmationSnapshot.receipt ||
```
gbarkhatov: Whole-receipt equality means a *transient* optional-chain failure kills a required-chain confirmation.

Within a single snapshot an optional chain that throws is dropped quietly. But this compares receipt strings *between* snapshots. If optional BTC is connected and succeeds in the first snapshot, then its `getNetwork()` flakes during the hooks (extension locked, UniSat's `withTimeout` firing), the second snapshot drops it, the receipts differ, and this throws `WALLET_CHANGED_ERROR_MESSAGE` — killing an ETH-only-required approval with "Wallet changed while confirming". Connecting an *extra* optional wallet mid-flight aborts the same way.

It fails closed, so not a correctness hazard — but it contradicts the package's own line that optional chains never block confirm, and the tests only cover an optional chain failing on the *first* snapshot, where the drop is consistent across all three and the receipts still match.

Comparing required entries plus the optional entries present in the *original* snapshot, and treating a later optional drop as a no-op, would keep the guard without the denial-of-service.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -58,93 +175,161 @@ export function WalletDialog({
...
-        // while they run. Re-read the live selection rather than confirming
-        // against the snapshot taken before the await, which would recreate an
-        // approval for a wallet that is no longer there.
-        const settled = collectConnections(selectedWalletsRef.current);
-        const lost = requiredChainIds.filter((chainId) => !settled.some(({ chain }) => chain === chainId));
-        if (lost.length > 0) {
-          throw new Error(`Wallet disconnected while confirming: ${lost.join(", ")}`);
-        }
+        await onConfirm?.(copyConnections(connections));
+        if (!isCurrentAttempt()) return;
```
gbarkhatov: Closing the dialog here abandons the attempt *after* the host callback already committed.

`isCurrentAttempt()` requires `visibleRef.current`, and `handleClose` calls `invalidateAttempt()`. So if the user closes during `await onConfirm?.()` — a hook a host would use to mint a server-side session — the host completes its side while this writes no receipt and never calls `confirm?.()`. Host and connector then disagree permanently, and the next reload re-prompts against a session the host thinks is live.

Latent rather than live today: nothing in `services/vault` wires the lifecycle hooks at all, so this lands on the first external host that adopts `onConfirm`. Worth deciding deliberately rather than by accident — once `onConfirm` has resolved and `checkIdentity()` still passes, arguably the approval should complete regardless of visibility, since a close after the hooks succeeded isn't a withdrawal of consent. Otherwise the host needs an explicit abandon signal.

(The close-during-`acceptTermsOfService` case is fine and tested — nothing irreversible has happened at that point.)

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -28,14 +30,129 @@ function findPrimaryConnection(
...
+        }
+
+        const liveReceipt = createConfirmationReceipt([candidate.connection], connectors, { [chain]: network });
+        if (liveReceipt !== candidate.receipt) {
+          throw new Error("Wallet network does not match the configured network");
+        }
+
+        return candidate;
+      } catch (error) {
+        if (requiredChainIds.includes(chain)) throw error;
```
gbarkhatov, line 155: The error object is discarded entirely for a non-required chain — no `onError`, no `displayError`, nothing.

Concretely: a user with BTC connected on mainnet against a signet-configured app, or a BTC wallet whose provider went missing, sees BTC listed as connected in the dialog, clicks Confirm, and BTC silently drops out of `connections`, out of what the hooks receive, and out of the stored receipt. The host proceeds believing BTC was never connected, and the user finds out only when a later route requires that chain and re-prompts.

CLAUDE.md is explicit on both counts here ("never swallow errors silently", "no silent fallbacks on critical paths"), and this is a consent boundary. Routing the caught error through the existing `onError?.()` before returning `null` would cost one line and keep the fail-closed behavior.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -58,93 +175,161 @@ export function WalletDialog({
...
+            currentSnapshot.receipt !== confirmationSnapshot.receipt ||
+            !hasSameAuthority(currentSnapshot.authority, confirmationSnapshot.authority)
+          ) {
+            throw new Error(WALLET_CHANGED_ERROR_MESSAGE);
+          }
+
+          return true;
+        };
+
+        if (!(await checkIdentity())) return;
```
gbarkhatov: Minor: this `checkIdentity()` re-runs a full snapshot — live provider round-trips included — against one taken microseconds earlier, and `createConfirmationSnapshot` already did its own settled re-read at lines 124-139 before returning. It closes no window the snapshot didn't just close itself.

Relatedly, the `isCurrentAttempt()` calls at 287, 291, 295 and 300 each sit with no `await` between them and the check beside them (`checkIdentity` opens with the same test), so they can never disagree. They read as load-bearing safety on a consent path, which makes the three that genuinely matter — 247, 266, 302 — harder to pick out.

Net effect of the extra snapshot is four live network reads per confirm; on a hardware wallet some of those are real device round trips, and the test asserting `getChainId` was called three times now pins that count as behavior.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -28,14 +30,129 @@ function findPrimaryConnection(
...
+    connection,
+    provider: connection.wallet.provider,
+    receipt: createConfirmationReceipt([connection], connectors),
+  }));
+
+  const results = await Promise.all(
+    candidates.map(async (candidate) => {
+      const { chain } = candidate.connection;
+      const { provider } = candidate;
+      if (chain !== "ETH" && chain !== "BTC") return candidate;
```
gbarkhatov: Nit: worth a comment saying *why* BBN is skipped. `IBBNProvider` exposes no network getter, so this is a necessity rather than an oversight — but as written it reads like an omission, and it means "fail closed before storage or host confirmation" holds for two of the three chains. Since simple-staking requires BBN on every route, the limit is worth stating in the PR body too.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
```diff
@@ -28,14 +30,129 @@ function findPrimaryConnection(
...
+
+      try {
+        if (!provider) throw new Error(`Connected ${chain} wallet has no provider`);
```
gbarkhatov: Nit: this and `"Wallet network does not match the configured network"` on line 110 are inline literals, while two of their siblings are module constants at lines 14-15. All four reach the user as `displayError` descriptions, and the line-110 one is probably the most likely new failure this change introduces — worth extracting alongside the other two.

packages/babylon-wallet-connector/src/components/WalletProvider/components/__tests__/WalletDialog.test.tsx
```diff
@@ -197,10 +248,528 @@ describe("confirming the dialog", () => {
...
+      await waitFor(() => {
+        expect(acceptTermsOfService).toHaveBeenCalledTimes(1);
+      });
+
+      switch (change) {
+        case "wallet":
+          selectedWallet.id = "rabby";
+          break;
+        case "address":
+          selectedWallet.account = { ...ETH_ACCOUNT, address: "0xchanged" };
```
gbarkhatov: These cases pass by mutating state production never mutates, which is why the gap in the production guard didn't show up here.

`connector.connectedWallet` is `selectedWallet` (line 437), so assigning `.account` in place does change what the code under test reads — the test is internally consistent. But production never mutates `Wallet.account`; the only writer is `Wallet.connect()`, and an in-extension account switch doesn't call it (see my comment on `WalletDialog.tsx:36`). It swaps the whole `Wallet` on the connector instead, which is what the test at line 251 already does.

So the two cases standing in for #2350's account/key criterion are green while the real path is uncovered. Driving them the way the app does — replacing `connector.connectedWallet`, or changing what `getAddress()` returns once the production fix is in — would make them fail against current `main`, which is what criterion 4 asks for.

Separately, the `"disconnect"` case looks like a tautology: the pre-PR code already threw when a required chain had no account, so it passes against the old behavior too.

packages/babylon-wallet-connector/src/components/WalletProvider/components/__tests__/WalletDialog.test.tsx
```diff
@@ -54,6 +83,16 @@ let close: ReturnType<typeof vi.fn>;
...
+function deferred<T = void>() {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((done) => {
+    resolve = done;
+  });
+
+  return { promise, resolve };
+}
 
 function setup({
```
gbarkhatov: `selectedWallets` is now dead fixture. `WalletDialog.tsx` no longer destructures it from `useWidgetState()` — zero occurrences left in the file — and everything reads `connector.connectedWallet` instead. But `setup()` still defaults it and injects it into `harness.widgetState`, and this PR *adds* nine more call sites passing hand-built wallets that nothing under test can observe.

Beyond the dead-code rule, it actively misleads: a reader reasonably concludes the user's dialog selection is the approved identity when it no longer is, and a regression that reintroduced a `selectedWallets` read would be invisible. It also leaves the test at line 251 named for a contrast that no longer exists ("while the selected wallet stays stale").

Suggest deleting the parameter, the harness field, and all call sites, and renaming that test.

packages/babylon-wallet-connector/src/components/WalletProvider/components/__tests__/WalletDialog.test.tsx
```diff
@@ -73,7 +112,7 @@ function setup({
     displayError: vi.fn(),
```
gbarkhatov: `displayError` is a bare `vi.fn()` with no handle held anywhere, so no test can assert on it — and every stop-the-flow test asserts only negatives (`confirm` not called, no receipt written).

That leaves the catch path at `WalletDialog.tsx:304-316` entirely unpinned: a regression turning the `throw` at line 272 into a silent `return` would leave the user staring at a confirm button that does nothing, and the whole suite would still pass. One assertion that `displayError` fired with the wallet-changed message would close it.

packages/babylon-wallet-connector/src/components/WalletProvider/components/__tests__/WalletDialog.test.tsx
```diff
@@ -197,10 +248,528 @@ describe("confirming the dialog", () => {
...
+        disconnect: vi.fn(),
+      },
+    };
+    setup({ requiredChainIds: [chain], selectedWallets: { [chain]: connectedWallet } });
+
+    await act(async () => {
+      screen.getByText("confirm").click();
+    });
+
+    const expectedConfirmations = chain === "BBN" ? 1 : 0;
```
gbarkhatov: This table asserts two opposite outcomes from one body — `chain === "BBN" ? 1 : 0` means the BBN case verifies confirmation *succeeds* while ETH/BTC verify it's *blocked* — with wallet id, account and config shape all chosen by stacked ternaries above. CLAUDE.md's test section is direct about this: "Three similar test functions are better than one parameterized abstraction", and each test should verify one behavior. Splitting into a BBN test and an ETH/BTC test would read top-to-bottom.

Same theme in the identity table at 426-483 (six unrelated mutations through a `switch`).

One more worth fixing while you're in here: line 395 puts `expect` calls inside `acceptTermsOfService.mockImplementation`, where the component's own `try/catch` catches the thrown assertion and converts it into a `displayError` — so a real failure surfaces as a confirmation error rather than as the assertion that failed.

gbarkhatov (dismissed):
Reviewed with three independent passes, deduped, and each finding below verified against the code at HEAD rather than accepted on agreement between passes. CI is green (`verify`, 8m49s), so nothing here is a build or test-run concern.

Approving so this isn't gated on my opinion — the findings stand as findings and you decide what lands here vs. a follow-up. One of them I do think should be settled before merge, though, because it goes to whether #2350 is actually closed:

**The account/public-key arm of the identity gate is inert in production.** Both snapshots read `connector.connectedWallet.account`, and `Wallet.account` has exactly one writer in the entire package — `Wallet.connect()` (`core/Wallet.ts:58`), reachable only from `WalletConnector.connect()`. An in-extension account switch never reaches it: both `ETHWalletProvider` and `BTCWalletProvider` handle `accountsChanged` by writing React state and calling `onAddressChange`. `createConfirmationSnapshot` reads only the *network* live; address and publicKeyHex stay cached. So `checkIdentity` compares a stale value against itself and confirms. The vault app happens to survive via a host-side `onAddressChange: runWalletReset`, but that's a host courtesy rather than the connector's guarantee, and this is a published package. The wallet-*swap* half of the original race is genuinely closed — it's the account/key half that isn't.

Worth noting the test that covers this mutates `selectedWallet.account` in place, which production never does — it swaps the whole `Wallet` on the connector. That's why the case is green while the path isn't covered.

Other themes: the live-network guard can reduce to `config === config` on two providers (fail-open in a gate meant to fail closed); a mid-flight optional-chain hiccup aborts the whole required-chain confirmation via whole-receipt equality; and `selectedWallets` is now dead fixture in the test file after production stopped reading it.

Things I checked and want to explicitly *not* leave hanging over the PR: the receipt shape and `RECEIPT_VERSION = 2` both predate this change; there's no hex/decimal hazard in the network comparison (`getChainId(): Promise<number>` against `ETHConfig.chainId: number`, and the explicit `undefined`/`null` checks are needed because `String(undefined)` is truthy); redetection can't swap a chain mid-confirmation; the close→reopen hook overlap is intended and pinned by tests; and the double-submit guard plus the generation-matched `finally` are correct — an old attempt cannot release a newer one.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in a7d51c5f2. Required baseline identities must remain present and unchanged. An original optional identity is checked only while it remains present, and later optional additions do not block confirmation. Tests cover optional loss, failure, and addition.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in a7d51c5f2. AppKit no longer falls back to the configured chain when wagmi has no live chain. Fixed-network providers still report their configured operational network because they have no independent network switch. The PR Safety section states this limit.

packages/babylon-wallet-connector/src/components/WalletProvider/components/__tests__/WalletDialog.test.tsx
jonybur: Fixed in a7d51c5f2. Provider, network, and identity cases are split into focused tests. Hook mocks mutate inputs only, and assertions run after the click.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in a7d51c5f2. Optional provider and network errors now go through onError before the optional identity is omitted. They do not block required-chain confirmation.

packages/babylon-wallet-connector/src/components/WalletProvider/components/__tests__/WalletDialog.test.tsx
jonybur: Fixed in a7d51c5f2. The dead selectedWallets fixture and all stale uses are removed. The connector replacement test now describes the active connector behavior.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in a7d51c5f2. Every snapshot now reads the provider address and public key, compares them case-insensitively with the captured account, and uses the verified live values for both hooks and the stored receipt.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in a7d51c5f2. The immediate duplicate snapshot and adjacent attempt checks are removed. Checks remain at async boundaries where the attempt can change.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in a7d51c5f2. The source explains that BBN has no network getter and still validates its live account identity. The PR Safety section now records that limit.

packages/babylon-wallet-connector/src/components/WalletProvider/components/__tests__/WalletDialog.test.tsx
jonybur: Fixed in a7d51c5f2. Address and public-key race tests now change provider return values. The provider replacement case remains separate.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: No code change. Close cancellation remains intentional and matches #2350. No current host commits side effects in onConfirm. A future side-effecting host will need an AbortSignal or transaction contract before it uses that hook.

packages/babylon-wallet-connector/src/components/WalletProvider/components/WalletDialog.tsx
jonybur: Fixed in a7d51c5f2. Provider-missing and network-mismatch messages now use named constants.

packages/babylon-wallet-connector/src/components/WalletProvider/components/__tests__/WalletDialog.test.tsx
jonybur: Fixed in a7d51c5f2. displayError is now a shared test spy, and the live identity race asserts the wallet-changed error description.

gbarkhatov (dismissed):

jonybur:
@greptileai

gbarkhatov (approved):

Issue #2353, opened 2026-08-27
[eth-first] Make AppKit setup consistent for Bitcoin and Ethereum


jonybur:
**Epic:** #2228
**Parent:** #2230
**Status:** Blocks #2230 and #2232
**Origin:** An older singleton weakness was expanded by the new entry points in #2267.

## Problem

The AppKit entry points keep different adapter state. The unified initializer can return no Wagmi configuration after another entry point created the shared modal. Capability checks also work in only one direction.

## Fix

Store one shared capability and configuration record. Validate every later initializer against it. Report setup failures to the host.

## Complete when

- [ ] Every initializer order returns the same valid configuration.
- [ ] An incompatible order fails with a clear error.
- [ ] The host receives initialization errors.
- [ ] Tests cover Ethereum-only, Bitcoin-only, and combined setup.

jonybur:
Delivered by PR #2364 (d96f4c88e), merged to main.

PR #2364, opened 2026-08-27
fix(wallet-connector): make AppKit initialization consistent

jonybur:
## Outcome

Every AppKit entry point uses one consistent setup.

## Change

- Store the modal, Wagmi configuration, Bitcoin configuration, and capabilities together.
- Publish the shared state only after setup succeeds.
- Validate every later initializer against the stored setup.
- Reject a Reown modal that was created outside this package.
- Reject later initializer calls that omit projectId after AppKit state exists.
- Cover Ethereum-only, Bitcoin-only, and combined initialization orders.

## Safety

- Failed setup does not publish partial state.
- A failed first attempt remains retryable.
- Manual and canonical setup modes stay separate for the page lifetime.
- Missing chain capabilities use typed initialization errors.

## Verification

- 216 wallet connector tests pass.
- The wallet connector build passes.
- Focused lint finds no error.
- Formatting and diff checks pass.
- Independent review found no merge blocker.

## Links

Tracks #2353. #2353 stays open.


packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
```diff
@@ -1,19 +1,195 @@
...
+}
+
+export interface AppKitState extends AppKitCapabilities {
+  readonly modal: AppKitModal;
+  readonly wagmiConfig?: Config;
+  readonly btcConfig?: SharedBtcAppKitConfig;
+}
+
+const configFunctionIds = new WeakMap<object, number>();
+let nextConfigFunctionId = 1;
```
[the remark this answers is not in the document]
jonybur: Fixed in 83196a05. `INITIAL_CONFIG_FUNCTION_ID` now names the seed. The shared state type is also generic, so the ETH entry does not import the BTC configuration.

jonybur:
@greptileai

packages/babylon-wallet-connector/src/core/wallets/appkit/__tests__/initialization.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    const combined = await import("../appKitModal");
+    const eth = await import("../../eth/appkit/modal");
+
+    combined.initializeAppKitModal(btcConfig());
+
+    expect(() => eth.initializeAppKitModal(ethConfig())).toThrow("already initialized without Ethereum support");
+    expect(mocks.createAppKit).toHaveBeenCalledTimes(1);
+  });
+
+  it("keeps the combined modal and adapter configs authoritative for each entry point", async () => {
```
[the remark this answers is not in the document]
jonybur: Fixed in f54e0ecb. Each test now verifies one production behavior.

packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
```diff
@@ -1,19 +1,194 @@
...
+
+function createConfigFingerprint(value: object): string {
+  const fingerprint = JSON.stringify(value, (_key, nestedValue) => {
+    if (typeof nestedValue === "bigint") {
+      return { type: "bigint", value: nestedValue.toString() };
+    }
+
+    if (typeof nestedValue === "function") {
+      return { type: "function", value: getConfigFunctionId(nestedValue) };
+    }
```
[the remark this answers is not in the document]
jonybur: Fixed in f54e0ecb. The runtime markers, capability names, network value, and messages now use named constants.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/sharedConfig.ts
```diff
@@ -1,11 +1,18 @@
...
 /**
  * Shared wagmi config singleton
- * 
+ *
  * This allows the AppKitProvider (class-based) to access the wagmi config
  * that's provided by the application-level WagmiProvider.
- * 
+ *
  * Usage:
  * 1. Application sets the config: setSharedWagmiConfig(wagmiConfig)
```
gbarkhatov: **Stale usage doc on a public export whose contract just became breaking.** This block still tells hosts "1. Application sets the config: `setSharedWagmiConfig(wagmiConfig)`", but as of line 24 that call goes through `registerManualAppKitConfig`, which throws once `appKitState` is set — and `validateAppKitInitialization` (`state.ts:137`) throws if the manual config was registered first. Both orderings are now fatal where the second call used to be a no-op or a `console.warn`. `setSharedWagmiConfig` is exported from `src/index.tsx:26` and `src/eth.ts:27` (and `setSharedBtcAppKitConfig` from `src/index.tsx:40`), so an external host that runs its own `WagmiProvider` *and* mounts `WalletProvider` with an `appKitConfig` breaks at startup with no doc anywhere describing the rule. Worth adding the exclusivity note to this block and to the twin at `btc/appkit/sharedConfig.ts:19-22` — or, if the manual path has no remaining users, deleting both setters and their barrel exports outright.

packages/babylon-wallet-connector/src/core/wallets/appkit/appKitModal.ts
```diff
@@ -135,20 +132,25 @@ export function initializeAppKitModal(config: AppKitModalConfig) {
     projectId,
     metadata,
   });
```
gbarkhatov, line 137: **`createAppKit` is memoizing at 1.8.12, so a non-null return isn't proof the modal is ours.** The pinned implementation (`node_modules/@reown/appkit/dist/esm/exports/react.js`) is `export let modal = undefined; export function createAppKit(options) { if (!modal) { modal = new AppKit({...}); getAppKit(modal); } return modal; }` — idempotent, and it **discards `options` on every call after the first**. Every new guard in this PR keys off our own `appKitState`, which is still `null` if a host (or any other library on the page) called `createAppKit` first. In that case we build a `BitcoinAdapter`, get the *pre-existing* modal back with our adapter and networks dropped, and then publish `btcConfig: { modal, adapter }` with `hasSharedBtcAppKitConfig() === true`. `useAppKitBtcBridge` (`src/hooks/appkit/btc/useAppKitBtcBridge.ts:48`) subscribes to an adapter the modal never registered, so Bitcoin connect never fires an account event and nothing errors. Reown exports the live binding for exactly this (`export declare let modal: AppKit | undefined`, `dist/types/exports/react.d.ts:10`) — reading it just before this call and routing a foreign modal through `failInitialization` would close it. If the intended contract is instead "hosts must never call `createAppKit` themselves", that belongs in the JSDoc on both `initializeAppKitModal` exports, since the package is published.

packages/babylon-wallet-connector/src/core/wallets/appkit/__tests__/initialization.test.ts
```diff
@@ -0,0 +1,341 @@
...
+});
+const combinedConfig = () => ({ ...ethConfig(), btc: { network: "signet" as const } });
+
+describe("shared AppKit initialization", () => {
+  beforeEach(() => {
+    vi.resetModules();
+    mocks.bitcoinError = null;
+    mocks.wagmiError = null;
+    mocks.createAppKit.mockReset();
+    mocks.createAppKit.mockImplementation(() => ({ id: Symbol("appkit-modal") }));
```
gbarkhatov: **The mock models `createAppKit` as a constructor, but the real one is a memoizing singleton.** `mockImplementation(() => ({ id: Symbol("appkit-modal") }))` returns a fresh object per call, so every `toHaveBeenCalledTimes(1)` and "reuses the modal" assertion in this suite passes purely on our own `appKitState` short-circuit — the suite is structurally unable to observe the reown-level collision described on `appKitModal.ts`. A memoizing stub (`let m; () => (m ??= { id: Symbol("appkit-modal") })`) would reflect the pinned contract, and would also make the `"AppKit modal"` retry case at lines 308-340 assert something the real dependency actually honours.

packages/babylon-wallet-connector/src/core/wallets/appkit/__tests__/initialization.test.ts
```diff
@@ -0,0 +1,341 @@
...
+  });
+
+  it("reuses the canonical Bitcoin connection event bus", async () => {
+    const combined = await import("../appKitModal");
+    const btcShared = await import("../../btc/appkit/sharedConfig");
+
+    combined.initializeAppKitModal(combinedConfig());
+    const btcConfig = btcShared.getSharedBtcAppKitConfig();
+
+    expect(btcShared.getSharedBtcAppKitConfig().connectionEvents).toBe(btcConfig.connectionEvents);
```
gbarkhatov: **This assertion compares the getter against itself and can never fail.** Line 137 binds `btcConfig = getSharedBtcAppKitConfig()`, and this line asserts `getSharedBtcAppKitConfig().connectionEvents` `toBe` `btcConfig.connectionEvents`. Both calls return `initializedState.btcConfig` — the single frozen object stored by `setAppKitState` — so it reduces to `x.connectionEvents === x.connectionEvents`. The behaviour the title names, the `config.connectionEvents ?? sharedBtcAppKitConfig?.connectionEvents ?? new EventTarget()` reuse at `btc/appkit/sharedConfig.ts:75`, is never exercised, and that bus carries the anti-spoofing property documented at `sharedConfig.ts:46-53` — so a green test here is false confidence. Either assert across two separate initialization attempts, or delete it (`tests/unit/appKitBtcConnectionBus.test.ts:88-98` already covers the setter path). Minor: the local `btcConfig` also shadows the module-level `btcConfig()` helper at line 66.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/sharedConfig.ts
```diff
@@ -14,23 +21,34 @@ import type { Config } from "wagmi";
...
+    if (!initializedState.wagmiConfig) {
+      throw new Error(MISSING_ETHEREUM_SUPPORT_ERROR);
+    }
+
+    return initializedState.wagmiConfig;
+  }
+
   if (!sharedWagmiConfig) {
     throw new Error(
       "Shared wagmi config not initialized. " +
```
gbarkhatov, line 44: **This error now recommends an API that the same PR makes fatal.** It tells the caller to "call `setSharedWagmiConfig()` in your app before using AppKit" — but that call throws if AppKit was already initialized, and poisons `manualAppKitCapabilities` so every later `initializeAppKitModal` throws if it wasn't. It's a live path: `services/vault/src/config/wagmi.ts` swallows an init failure into `wagmiInitError` and keeps rendering on the fallback config, so the first deposit reaching `depositFlowSteps/ethereumSubmit.ts:34` surfaces exactly this message. CLAUDE.md asks for actionable errors; this one misdirects. Point it at `initializeAppKitModal` and state that the two modes are exclusive. Same wording at `btc/appkit/sharedConfig.ts:93-96`.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/sharedConfig.ts
```diff
@@ -14,23 +21,34 @@ import type { Config } from "wagmi";
...
+    console.warn(SHARED_WAGMI_REPLACEMENT_WARNING);
   }
   sharedWagmiConfig = config;
 }
 
 export function getSharedWagmiConfig(): Config {
+  const initializedState = getAppKitState();
+  if (initializedState) {
+    if (!initializedState.wagmiConfig) {
+      throw new Error(MISSING_ETHEREUM_SUPPORT_ERROR);
```
gbarkhatov: **Bare `Error` where every sibling failure in this PR is a typed `WalletError`.** `state.ts` routes all its new failures through `failInitialization`, producing a `WalletError` with `ERROR_CODES.WALLET_INITIALIZATION_FAILED` and a `chainId` of `"ETH"`/`"BTC"`, and its messages carry a remedy ("A page that needs Ethereum must initialize AppKit with Ethereum first"). This throw — and its twin at `btc/appkit/sharedConfig.ts:86` — is an untyped `Error` with no code, no `chainId`, and no remedy, for the same class of condition. The `onError` handlers in `WalletProvider/index.tsx` and `ETHWalletConnectorProvider.tsx` both inspect error shape, so these two land as unknown failures and lose the chain attribution. Exporting `failInitialization` (or a thin `failMissingCapability` wrapper) and using it in both getters would make this uniform.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/modal.ts
```diff
@@ -61,17 +63,21 @@ export function createETHWagmiAdapter(chain: Chain, projectId: string): WagmiAda
...
+  const capabilities = createAppKitCapabilities({
+    projectId: config.projectId,
+    metadata: config.metadata,
+    ethChain: chain,
+  });
+  const existingState = getAppKitState();
+
+  if (existingState) {
+    assertAppKitCapabilities(existingState, capabilities);
+    return { modal: existingState.modal, wagmiConfig: existingState.wagmiConfig! };
```
gbarkhatov: **`wagmiConfig!` encodes a cross-module invariant that nothing enforces.** This is sound today: `assertAppKitCapabilities` runs first and throws when `requested.ethFingerprint` is set but `existing.ethFingerprint` isn't, and both `ethFingerprint` and `wagmiConfig` derive from `config.eth?.chain` in `appKitModal.ts:102-108,139`. But they're two independent optional fields on `AppKitState` (`state.ts:19,25`) with no type-level link, and the invariant lives in a different module from the assertion. If it ever slips, `undefined` ships typed as `Config` and the vault drops to `createFallbackConfig()` — a wagmi config with no AppKit connectors, i.e. wallet connection quietly stops working, which is the silent-fallback case CLAUDE.md calls out. Worth either an explicit `failInitialization` guard or making `AppKitState` a discriminated union so ETH-capable state carries a required `wagmiConfig`. Same at line 91 (that one is provably safe — `setAppKitState` is called with `wagmiConfig` two lines up).

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/sharedConfig.ts
```diff
@@ -90,4 +111,5 @@ export function hasSharedBtcAppKitConfig(): boolean {
  */
 export function __resetSharedBtcAppKitConfigForTests(): void {
   sharedBtcAppKitConfig = null;
+  __resetManualAppKitConfigForTests(BITCOIN_CAPABILITY);
```
gbarkhatov, line 115: **This helper's doc says it "wipes the singleton", but it can no longer clear the thing that matters.** After any initializer runs, `hasSharedBtcAppKitConfig()` reads `appKitState.btcConfig`, and `state.ts` exports no way to clear `appKitState` — `setAppKitState` throws on any second call. So a test that initializes canonically, calls this reset, and initializes again gets a confusing "already initialized" throw. The new suite works around it with `vi.resetModules()` rather than using this helper. There's also no Ethereum counterpart, so `__resetManualAppKitConfigForTests("Ethereum")` has no caller at all and that parameter's ETH arm is unreachable. A single `__resetAppKitStateForTests()` in `state.ts` that clears `appKitState` and the whole capability set, with this helper delegating to it, would fix both and let the doc be true.

packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
```diff
@@ -1,19 +1,200 @@
...
  */
-let appKitModal: AppKitModal | null = null;
+let appKitState: AppKitState<unknown> | null = null;
+const manualAppKitCapabilities = new Set<"Ethereum" | "Bitcoin">();
+
+function failInitialization(message: string, chainId?: string): never {
+  throw new WalletError({ code: ERROR_CODES.WALLET_INITIALIZATION_FAILED, message, chainId });
+}
+
+export function getAppKitState<BtcConfig = never>(): AppKitState<BtcConfig> | null {
```
gbarkhatov: **Unchecked `as` cast at the boundary the package's own docs call security-relevant.** The stored value is `AppKitState<unknown>` and this asserts it to whatever the caller names, with no validation at compile or run time. `btc/appkit/sharedConfig.ts:83` asserts `SharedBtcAppKitConfig` and hands the result straight to `useAppKitBtcBridge`, which destructures `{ adapter, network, connectionEvents }` — and `connectionEvents` is the anti-spoofing bus described at `sharedConfig.ts:46-53`. TypeScript can't catch a divergence between the shape `appKitModal.ts:142-147` writes and the shape BTC consumers read. `SharedBtcAppKitConfig` reaches `BitcoinAdapter`/`createAppKit` through `import type` only, so `state.ts` can `import type` it at zero runtime cost, type `btcConfig` concretely, and drop both the generic and the cast. Related: the `BtcConfig = never` default (line 23) means callers that omit the type argument — `eth/appkit/sharedConfig.ts:33`, `eth/appkit/modal.ts:74` — see `btcConfig: never`, so any `if (state.btcConfig)` added later narrows to `never` and typechecks against anything.

packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
```diff
@@ -1,19 +1,200 @@
...
+const BIGINT_FINGERPRINT_TYPE = "bigint";
+const FUNCTION_FINGERPRINT_TYPE = "function";
+const CONFIG_FINGERPRINT_SERIALIZATION_ERROR = "AppKit configuration cannot be serialized.";
+const INITIAL_CONFIG_FUNCTION_ID = 1;
+const configFunctionIds = new WeakMap<object, number>();
+let nextConfigFunctionId = INITIAL_CONFIG_FUNCTION_ID;
+
+function getConfigFunctionId(value: object): number {
+  const existingId = configFunctionIds.get(value);
+  if (existingId) return existingId;
```
gbarkhatov, line 40: **Function-valued chain properties are fingerprinted by reference identity, so two structurally identical chains can compare unequal.** `getConfigFunctionId` keys a `WeakMap` on the function object, so a viem `Chain` whose `formatters`/`serializers`/`fees` are constructed per call — an inline `defineChain({ fees: { estimateFeesPerGas() {...} } })`, an OP-stack/zkSync/Celo chain, or a module reloaded under Fast Refresh while `state.ts` survives — produces a different `ethFingerprint` and makes `assertAppKitCapabilities` throw "different Ethereum chain configuration" for what is the same chain. Not reachable in this repo: `services/vault/src/config/network/eth.ts` spreads viem's module-level `mainnet`/`sepolia`, which preserves function identity, and the vault initializes once. It's a trap for external consumers. Narrowing the ETH fingerprint to what `createAppKit`/`WagmiAdapter` actually consume (`id`, `name`, `nativeCurrency`, `rpcUrls.default.http[0]`) would also stop a viem version bump changing `contracts`/`blockExplorers` between two call sites. Worth noting the fixture chain at `initialization.test.ts:58-63` has no function properties, so this branch — including the `INITIAL_CONFIG_FUNCTION_ID` fix from the earlier round — never executes in the suite.

packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
```diff
@@ -1,19 +1,200 @@
...
-  appKitModal = modal;
+/** @internal */
+export function __resetManualAppKitConfigForTests(capability: "Ethereum" | "Bitcoin"): void {
+  manualAppKitCapabilities.delete(capability);
+}
+
+export function validateAppKitInitialization(projectId?: string): projectId is string {
+  if (!projectId) {
+    if (appKitState) {
+      failInitialization(
```
gbarkhatov, line 131: **Behaviour change worth calling out in the PR description.** Previously both entry points checked for an existing modal *before* validating `projectId` and simply returned it; now a re-entry with a falsy `projectId` throws instead. A host whose `projectId` comes from an env var that's present at first init but absent in a later render path (SSR/client split, a lazily-mounted subtree) surfaces an `onError` where nothing was previously wrong. The test at lines 302-305 pins this as intended, so I'm reading it as deliberate — just flagging that it's a semantic change to a published entry point rather than a refactor, and the PR body currently describes the change as consistency-only.

packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
```diff
@@ -1,19 +1,200 @@
...
+    if (nestedValue && typeof nestedValue === "object" && !Array.isArray(nestedValue)) {
+      return Object.fromEntries(
+        Object.entries(nestedValue).sort(([left], [right]) => (left < right ? -1 : left > right ? 1 : 0)),
+      );
+    }
+
+    return nestedValue;
+  });
+
+  if (fingerprint === undefined) {
```
gbarkhatov, line 66: **Unreachable guard, and it throws outside the module's error discipline.** `createConfigFingerprint` takes `value: object` and the replacer returns a serializable value for every branch, so `JSON.stringify` cannot return `undefined` here — `CONFIG_FINGERPRINT_SERIALIZATION_ERROR` can't fire. It also throws a bare `TypeError` where every other failure in this file goes through `failInitialization` → `WalletError`. Same for the `BIGINT_FINGERPRINT_TYPE` branch at line 47: the only two inputs are `AppKitMetadata` (all strings) and a viem `Chain` (numeric `id`, `decimals`, `blockCreated`), neither of which can carry a `bigint`. Zero-dead-code would say drop both; if they're deliberate hardening against a future `Chain` shape, a test that reaches them would justify keeping them.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/__tests__/provider.test.ts
```diff
@@ -39,7 +36,9 @@ vi.mock("wagmi/connectors", () => ({
 }));
 
 vi.mock("@/core/wallets/appkit/state", () => ({
+  getAppKitState: vi.fn(() => null),
   getAppKitModal: vi.fn(() => null),
+  registerManualAppKitConfig: vi.fn(),
```
gbarkhatov, line 42: **This stub makes the PR's new guard a no-op for the whole file.** `registerManualAppKitConfig: vi.fn()` means the `setSharedWagmiConfig(sharedConfig)` call at line 99 skips the exclusivity check entirely, so the ETH provider suite exercises none of the behaviour this PR adds. That's a defensible choice for a test about the provider rather than the guard, and it's what keeps the file green — but combined with the missing ETH reset helper it means `registerManualAppKitConfig`'s `"Ethereum"` path is only ever covered by `initialization.test.ts`. Worth a one-line comment here saying the stub is deliberate, so the next person doesn't read it as the guard being unnecessary.

gbarkhatov (dismissed):
Approving. The core of this PR is right and is a real fix: publishing `appKitState` only after `createAppKit` returns closes a genuine partial-state bug, and making the manual and canonical modes mutually exclusive in both directions removes the ambiguity about which config the providers read. Findings below are suggestions — you know the constraints better than I do, and a follow-up is fine for most of them.

Two I'd genuinely weigh before merge:

1. **The two public setters now throw in every ordering, and their own JSDoc still documents them as the way to configure AppKit.** `setSharedWagmiConfig` / `setSharedBtcAppKitConfig` have zero production callers left in this repo, but both are still exported from `src/index.tsx:26,40` and `src/eth.ts:27`. An external host that provides its own wagmi config and also calls `initializeAppKitModal` goes from working to a hard throw at startup, and the only description of the new rule is the runtime error.
2. **The capability guard is anchored on our `appKitState`, but reown's own `createAppKit` is the thing that actually enforces one-modal-per-page — and it's memoizing.** Verified in the pinned dependency: `node_modules/@reown/appkit/dist/esm/exports/react.js` is `export let modal = undefined; export function createAppKit(options) { if (!modal) { modal = new AppKit({...}); getAppKit(modal); } return modal; }`. It returns the existing modal and **discards the options** on any later call. Details inline on `appKitModal.ts`.

### Refuted — findings I checked and dropped, so they don't get re-raised

- **"Missing changeset for a breaking change."** There is no `.changeset/` directory in this repo; releases run through `@internal/release` / `pnpm run ci:release`, not changesets. Nothing missing here.
- **"Empty-string `projectId` silently returns `null`."** Pre-existing, not introduced here — the removed code had `if (!config.projectId) return null;` at both entry points. Only the *already-initialized* case changed (noted inline on `state.ts`).
- **"The setters are dead code under the zero-dead-code policy."** They're public exports of a published package, not dead code. The issue is the stale docs, not the existence of the functions.
- **"`btcConfig`'s three-way guard can silently degrade a BTC-capable modal."** All three terms are assigned together in the same `if (config.btc?.network)` block, so the divergence isn't reachable. Redundant narrowing, not a defect — left it out.
- **"The vault re-initializes / has no error handling around init."** `services/vault/src/config/wagmi.ts` runs `initializeVaultWagmi()` exactly once at module load, guards `!result.wagmiConfig`, and wraps the call in try/catch with a fallback config.

### Verified clean

- **No `@reown/appkit` 1.8.12 API misuse.** `networks` is a non-empty tuple in `TypesUtil.d.ts`; the `allNetworks.length === 0` check at `appKitModal.ts:124` precedes the cast, so the runtime guard is real. `WagmiAdapter.wagmiConfig` matches `AppKitState.wagmiConfig`.
- **Failure atomicity holds.** Nothing is published before `createAppKit` returns, so an adapter or modal throw leaves `appKitState`, `sharedWagmiConfig` and `sharedBtcAppKitConfig` untouched and a retry succeeds. `setSharedBtcAppKitConfig` also resolves the config fully before registering, so a throwing getter can't half-register.
- **`setAppKitModal` is fully removed** with no surviving references anywhere in the monorepo, and all three `getAppKitModal()` call sites null-guard.
- **Manual and canonical modes are genuinely exclusive in both directions**, which is what makes the getters' "canonical first, manual second" precedence unambiguous.
- **Entry-point isolation is preserved** — `appKitModal.ts` imports `SharedBtcAppKitConfig` as `import type` only, so the `./eth` graph still pulls in no Bitcoin adapter.
- **CI is green at head**: `verify` ran `pnpm build` (incl. per-package `tsc --noEmit`), `pnpm run lint` and affected tests.

---

*Reviewed with three independent passes over the full diff, deduped, and each posted finding re-verified against the code at HEAD and the pinned `@reown/appkit@1.8.12` sources.*

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/sharedConfig.ts
jonybur: Fixed in bfd9acada. Both public setter docs now state that canonical initialization and manual shared configuration are exclusive modes.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/sharedConfig.ts
jonybur: Fixed in bfd9acada. Missing-config errors now describe canonical and manual modes as exclusive alternatives.

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/sharedConfig.ts
jonybur: Fixed in bfd9acada. The helper now states that it resets only manual Bitcoin configuration and its registry entry. Reown exposes no reset API, so canonical tests continue to use module isolation.

packages/babylon-wallet-connector/src/core/wallets/appkit/appKitModal.ts
jonybur: Fixed in bfd9acada. Both combined and Ethereum-only initializers read the live Reown modal binding and throw a typed initialization error before adapter creation when another owner created it.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/sharedConfig.ts
jonybur: Fixed in bfd9acada. Missing Ethereum and Bitcoin capabilities now use the shared typed WalletError path with chain attribution and an initialization remedy.

packages/babylon-wallet-connector/src/core/wallets/appkit/__tests__/initialization.test.ts
jonybur: Fixed in bfd9acada. The self-comparison test is removed, and the remaining local is named sharedBtcConfig.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/__tests__/provider.test.ts
jonybur: Fixed in bfd9acada. The test mock now states that it deliberately bypasses mode exclusivity and that initialization tests cover the guard.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/modal.ts
jonybur: Fixed in bfd9acada. Existing Ethereum state uses an explicit capability getter, and new initialization returns the adapter configuration directly. No non-null assertions remain.

packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
jonybur: No code change. The serialization guard remains boundary hardening for external runtime objects. A root toJSON method can return undefined, and custom chain extensions can contain bigint values.

packages/babylon-wallet-connector/src/core/wallets/appkit/__tests__/initialization.test.ts
jonybur: Fixed in bfd9acada. The Reown mock now memoizes one live modal, resets between tests, and covers foreign-modal rejection for both initializers.

packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
jonybur: Fixed in the PR description. It now states that later initializer calls without projectId are rejected after AppKit state exists.

packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
jonybur: Fixed in bfd9acada. AppKitState now stores the concrete SharedBtcAppKitConfig type. The generic defaults and unchecked casts are removed. The type-only import is excluded from the runtime entry graph.

packages/babylon-wallet-connector/src/core/wallets/appkit/state.ts
jonybur: No code change. Strict function identity is intentional because formatters, serializers, and fee functions can change chain behavior even when basic fields match. Callers must reuse the same chain configuration object for one page lifetime.

gbarkhatov (approved):

jonybur:

Issue #2359, opened 2026-08-27
Require safety checks for critical pull requests


jonybur:
## Outcome

Critical pull requests require two qualified approvals and all safety checks before merge.

Status: parked. This repository-wide work is outside #2228 and is not an epic dependency. #2370 is partial implementation. This issue stays open for future work.

## Problem

The written policy requires two reviewers for critical code. The live [repository ruleset](https://github.com/babylonlabs-io/babylon-toolkit/rules/21788914) requires `verify`, `detect`, code-owner review, thread resolution, and fresh approval after changes. It still sets a one-approval minimum.

The gate proposed in #2370 runs code from the pull request. A pull request can change or remove that gate. Ownership policy changes also lack the two-reviewer gate. Earlier green runs skipped the new approval logic. The [later live run](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34167538394/job/101881474802) executed it and blocked with zero of two approvals. Live passing proof still remains.

This gap came from existing repository configuration. The epic did not introduce it.

## Scope

- Host the approval gate outside this repository. Require its trusted version through repository rules. Do not execute code from the pull request to evaluate approvals.
- Keep the approval count, critical file list, and qualified owners outside the pull request's control. Require two qualified approvals for critical changes, including ownership and gate-policy changes.
- Check owner coverage for every touched critical file. Exclude the author, non-owners, stale approvals, and withdrawn approvals.
- Recheck approvals after pushes, review changes, and changes to the target branch. Cover `main` and protected release branches.
- Keep safety documentation and critical file inventories consistent with enforcement.
- Define the exception process, its owner, and the audit record for each exception.

## Complete when

- [ ] A failed `verify` check blocks merge.
- [ ] A trusted external gate requires two qualified approvals for critical changes. A pull request cannot weaken or replace it.
- [ ] Unresolved review threads block merge.
- [ ] Material changes require fresh approval.
- [ ] The exception process has a named owner and an audit record.
- [ ] Linked live runs show failure with zero or one qualified approval and success with two. They show that stale, withdrawn, author, and non-owner approvals do not satisfy the rule.
- [ ] Linked live checks cover policy edits, target-branch changes, and protected release branches. Checks cannot pass by skipping the required approval logic.
- [ ] All remaining #2370 review findings have fixes or agreed dispositions. Safety documentation matches the final behavior.

## Dependencies

- Partial implementation: #2370. It does not close this issue on its current scope.
[...]

Issue #10, opened 2025-04-30
Staking SDK


jrwbabylonlab:
A new FE focused lib which abstract staking flow logic into a single place where minimal inputs/configs are needed in order to adopt and use the methods for perform staking operations.The goal of this SDK are:Simple to set up and use by combining all necessary tools, dependencies. (e.g by providing api, rpc, mempool urls)Isolate staking logic from react to reduce change risks and need of auditing

This is the parent ticket which serve as the investigation ticket and require tasks to be breakdown once investigation is done

PR #2370, opened 2026-08-29
ci(workflows): gate critical-path PRs on two code-owner approvals (#2359)

filippos47:
## Outcome

Critical changes require current code-owner approvals for each touched path. Tracks #2359. #2359 stays open.

Status: parked outside #2228. Signed review fixes are pushed. The trusted external approval gate and live approval proof remain incomplete.

## Change

- Read qualified owners from the target branch. Evaluate each critical file. Exclude the author, stale approvals, and withdrawn approvals. Block standing requests for changes.
- Detect both paths when a critical file moves. Preserve the existing owner list, including @kirugan.
- Protect review policy and refund validation. Keep the critical path inventories consistent.
- Check approvals before optional comments and labels. Give `verify` a read-only token.

## Safety

The live ruleset still requires one approval. This PR adds a check for two qualified approvals on critical changes.

A PR can still edit this workflow. #2359 requires a trusted external check before it closes. This PR does not complete that requirement or prove that every approval dismissal and target-branch change invalidates a prior success.

## Verification

- Actionlint 1.7.12 passes all three changed workflows. Embedded JavaScript and Bash parse. The staged diff passes whitespace checks.
- Actual review records reject zero and one approval, accept two qualified approvals, and exclude a dismissed approval.
- Replaying the critical file move from #2236 retains both paths and detects the original critical path.
- All 36 critical path entries agree across the inventories. They match 148 files. The 35 existing ownership rows match main.
- Before merge: require passing CI, two current qualified approvals, and live proof for approval dismissal, target-branch changes, and protected release branches.

## Links

Tracks #2359. #2359 stays open. Completed merge dependency: #2398.

[Repository safety rules](https://github.com/babylonlabs-io/babylon-toolkit/rules/21788914). [Live approval evidence request](https://github.com/babylonlabs-io/babylon-toolkit/pull/2370#discussion_r3893636350).


.github/workflows/critical-path-check.yml
```diff
@@ -125,3 +144,93 @@ jobs:
...
+            const required = Number(process.env.REQUIRED_CRITICAL_APPROVALS);
+            const pr = context.payload.pull_request;
+            const matched = fs.readFileSync('matched.txt', 'utf8').trim().split('\n');
+
+            // Qualified reviewers are the CODEOWNERS owners of the matched
+            // files. Same resolution rule as GitHub: for each file the last
+            // matching pattern wins. Patterns in this file are `*`, `/dir/`
+            // or `/path/to/file`; anything else is a CODEOWNERS change that
+            // has to extend this matcher.
+            const rules = fs.readFileSync('.github/CODEOWNERS', 'utf8')
```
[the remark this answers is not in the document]
filippos47: Fixed in bef64a7b: the gate now reads `CODEOWNERS` from the PR base branch (`git show origin/$BASE_REF:.github/CODEOWNERS`) instead of the merge checkout, and the required count comes from the repo Actions variable `CRITICAL_PATH_REQUIRED_APPROVALS`. Residual weakness (a PR can still edit the workflow file itself) is documented in the workflow header and the PR description.

.github/workflows/critical-path-check.yml
```diff
@@ -125,3 +144,93 @@ jobs:
...
+              .filter(line => line && !line.startsWith('#'))
+              .map(line => {
+                const [pattern, ...owners] = line.split(/\s+/);
+                return { pattern, owners };
+              });
+            const patternMatches = (pattern, file) => {
+              if (pattern === '*') return true;
+              if (pattern.includes('*')) {
+                throw new Error(`Unsupported CODEOWNERS glob "${pattern}"; extend the matcher in critical-path-check.yml`);
+              }
```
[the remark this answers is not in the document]
filippos47: Fixed in bef64a7b: CODEOWNERS patterns are now matched with gitignore-style glob semantics (`*` stays within a segment, `**` crosses, leading `/` anchors, no `/` matches any depth, trailing `/` is a directory) instead of throwing on anything other than `*`, so unrelated entries cannot break the check.

filippos47:
@greptileai please re-review the latest commit.

.github/workflows/critical-path-check.yml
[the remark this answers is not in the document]
filippos47: Already fixed in bef64a7b (your review covered 0ce18474 only): the detect step now runs `git show origin/$BASE_REF:.github/CODEOWNERS > base-codeowners.txt` and the gate reads that file, so the PR merge ref never supplies the owner list. The remaining self-modification weakness (a PR can edit the workflow file itself) is documented in the workflow header and PR description as a follow-up.

jonybur (dismissed):
The trust model holds: CODEOWNERS is read from origin/$BASE_REF (base-codeowners.txt), the approval count comes from vars.CRITICAL_PATH_REQUIRED_APPROVALS (admins only), the author is excluded from standing (login === author skip), only reviews whose commit_id equals pr.head.sha count, and DISMISSED overrides an earlier APPROVED because it is not COMMENTED-and-therefore-skipped. Re-run on push works because pull_request default types include synchronize, and GitHub's required-check resolution uses the latest run per check name on the head commit, so a stale approval-time run cannot pass HEAD. The pull_request_review runs correctly pin refs/pull/N/merge — GITHUB_BASE_REF is genuinely unset on that event, so BASE_REF: pull_request.base.ref is the right source. Known residual (a PR can edit the workflow itself) is called out in the header with the correct fix (out-of-repo hosting via a ruleset workflows rule). verify.yml swap from origin/main to origin/$GITHUB_BASE_REF is the same base-branch generalisation and matches the release/** trigger addition.

.github/workflows/critical-path-check.yml
```diff
@@ -52,8 +77,11 @@ jobs:
             "services/vault/src/utils/btc/scriptPubKeyAddress.ts"
           )
```
gbarkhatov, line 79: **Merge blocker — none of this gate has ever executed.**

`paths=` contains no `.github/` entry, so this PR (which touches only `.github/`) sets `matched=false`. I pulled the live job for `bef64a7` (`actions/jobs/99118573118`): `Detect critical-path files` logged `No critical-path files touched.`, and both `Comment on PR` and `Require code-owner approvals` report **`skipped`**.

So the green `detect` on this PR is not evidence that any of the ~118 new lines work — and because the logic is inline YAML, it cannot be run locally either. The 18 scenarios described in the PR body were run against an extracted copy that isn't in the diff.

Before the ruleset is relied on, land a throwaway PR that touches a real critical path and confirm `detect` goes red, then green on the Nth approval.

.github/workflows/critical-path-check.yml
```diff
@@ -52,8 +77,11 @@ jobs:
...
-          git fetch origin main || true
-          changed=$(git diff --name-only "origin/${GITHUB_BASE_REF}...HEAD")
+          git fetch origin "$BASE_REF"
+          changed=$(git diff --name-only "origin/${BASE_REF}...HEAD")
+          # Ownership is resolved from the base branch, not the PR: a PR
+          # must not be able to appoint the reviewers that approve it.
+          git show "origin/${BASE_REF}:.github/CODEOWNERS" > base-codeowners.txt
 
           matched=()
           for f in $changed; do
```
gbarkhatov: Three shell nits on this loop, all pre-existing lines but now load-bearing since `detect` blocks merge:

- `$changed` is unquoted, so a tracked path containing a space splits into fragments that match no prefix and is **silently skipped** — a critical file could slip the gate.
- `case "$f" in "$p"|"$p"*)` has a dead first alternative: `"$p"*` already matches the empty suffix, so `"$p"` can never be the branch that fires.
- `set -e` at :48 without `-o pipefail` lets a truncated `matched.txt` reach the gate, where it surfaces as the misleading `Only 0 code owner(s) cover the matched paths`.

.github/workflows/critical-path-check.yml
```diff
@@ -1,8 +1,26 @@
...
+# Trust boundary. The approval count comes from the repository Actions
+# variable CRITICAL_PATH_REQUIRED_APPROVALS (repo admins only) and the
+# qualified reviewers come from CODEOWNERS on the PR's BASE branch, so a PR
+# cannot lower the bar or appoint its own reviewers by editing those. What a
+# PR *can* still do is edit this file: pull_request workflows run the PR's
+# copy, so a PR that deletes the gate step passes `detect`. Reviewers must
+# treat any change to this workflow as a critical-path change. Closing that
+# hole needs the gate hosted outside the repo and required via a ruleset
+# `workflows` rule (as enforce-pinned-actions-hashes does).
 on:
```
gbarkhatov, line 7: No `concurrency:` block here, though [verify.yml:8-10](../blob/main/.github/workflows/verify.yml#L8-L10) has exactly the pattern to copy.

Two owners approving within the same window is the *normal* way a 2-approval gate is satisfied, and it runs two jobs in parallel. The marker lookup at :130-147 is an unlocked read-modify-write — both runs call `listComments`, neither sees the marker, both `createComment`.

Compounding it, that `listComments` call is unpaginated (30-item default) while :233 correctly uses `github.paginate`. On a busy PR the marker scrolls out of the first page and a duplicate comment is posted on **every** review.

```yaml
concurrency:
  group: critical-path-check-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

.github/workflows/critical-path-check.yml
```diff
@@ -1,8 +1,26 @@
...
   pull_request:
```
gbarkhatov, line 8: No `types:` here, and the docs are explicit: *"By default, a workflow only runs when a `pull_request` event's activity type is `opened`, `synchronize`, or `reopened`."*

Retargeting a PR's base branch changes no head SHA and is none of those three, so no re-run happens and the existing check result stands on the very commit the ruleset gates.

Concrete: open against `release/vault-v1.x-rc` where the critical change is already present in that branch → diff shows nothing critical → `detect` green. Retarget to `main` → no re-run → merges into `main` with zero code-owner approvals on a critical path. `verify.yml:94` has the same exposure via a stale `--base`.

**Partly unverified:** I confirmed the default-types sentence verbatim, but could not get docs.github.com to state that a retarget fires `edited` specifically — please confirm before adopting `types: [opened, synchronize, reopened, edited]`.

.github/workflows/critical-path-check.yml
```diff
@@ -1,8 +1,26 @@
...
-    branches: [main]
+    branches: [main, 'release/**']
+  pull_request_review:
```
gbarkhatov, line 11: `branches: [main, 'release/**']` applies only to `pull_request` — it is not a valid filter under `pull_request_review`.

So a review on *any* PR in the repo (a stacked PR based on a feature branch, a draft, a PR into a branch with no `.github/CODEOWNERS`) triggers this workflow, posts the "Critical path PR" comment, applies the label, and can fail a check on a PR the `critical-pr-safety-checks` ruleset never governs. In the no-CODEOWNERS case the `git show` at :84 fails outright.

Fix with a job-level guard, since `branches:` is unavailable on this event:

```yaml
if: github.event.pull_request.base.ref == 'main' || startsWith(github.event.pull_request.base.ref, 'release/')
```

.github/workflows/critical-path-check.yml
```diff
@@ -77,6 +105,8 @@ jobs:
       - name: Comment on PR
```
gbarkhatov: This step 403s on fork and Dependabot PRs — `GITHUB_TOKEN` is read-only there — and unlike the `addLabels` call at :149-157 the `createComment` at :142 is **not** wrapped, so the job dies before `Require code-owner approvals` ever runs. `detect` then sits red and no number of approvals clears it.

Low severity for this repo specifically: there's no `.github/dependabot.yml` and zero fork PRs in the last 60. But it's a one-line fix — mirror the existing `core.warning` guard — and the consequence changed from "noisy failure" to "unmergeable PR" the moment `detect` became required.

.github/workflows/critical-path-check.yml
```diff
@@ -92,7 +122,7 @@ jobs:
               ...matched.map(m => `- \`${m}\``),
               '',
               '**Required before merge:**',
-              '- Two code-owner approvals',
+              `- ${process.env.REQUIRED_APPROVALS} code-owner approvals on the current head commit (enforced by the \`detect\` check; pushing a new commit resets them)`,
```
gbarkhatov: `REQUIRED_APPROVALS` is interpolated raw here, and this step runs *before* the validation at :169-172. With the variable unset or renamed, the reviewer-facing comment renders:

> `-  code-owner approvals on the current head commit`

This comment is the policy statement reviewers act on, so it shouldn't be able to print an empty count (or `undefined`). Hoist the validation so both consumers share one checked value.

.github/workflows/critical-path-check.yml
```diff
@@ -125,3 +155,118 @@ jobs:
...
+        if: steps.detect.outputs.matched == 'true'
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0 https://github.com/actions/github-script/releases/tag/v7.1.0
+        env:
+          # Repository Actions variable, settable by repo admins only. Policy
+          # source: CLAUDE.md > "CRITICAL PATHS — HUMAN REVIEW REQUIRED".
+          REQUIRED_APPROVALS: ${{ vars.CRITICAL_PATH_REQUIRED_APPROVALS }}
+        with:
+          script: |
+            const fs = require('fs');
+            const required = Number.parseInt(process.env.REQUIRED_APPROVALS, 10);
```
gbarkhatov: `Number.parseInt` accepts trailing garbage, so the guard doesn't enforce what its error message promises. I ran these:

```
"1e3"          -> 1    passes guard: true
"2.9"          -> 2    passes guard: true
"2 approvals"  -> 2    passes guard: true
"2abc"         -> 2    passes guard: true
```

An admin typing `1e3` intending 1000 silently drops the gate to **one** approval, with a green check and no warning. That's a silent fallback on the enforcement path, which CLAUDE.md forbids.

```js
const raw = process.env.REQUIRED_APPROVALS;
if (!/^[1-9]\d*$/.test(raw ?? '')) throw new Error(`... got "${raw}"`);
const required = Number(raw);
```

.github/workflows/critical-path-check.yml
```diff
@@ -125,3 +155,118 @@ jobs:
...
+            // files, read from the base branch (base-codeowners.txt). Same
+            // resolution rule as GitHub: for each file the last matching
+            // pattern wins. Pattern syntax follows the CODEOWNERS docs:
+            // gitignore-style, where `*` does not cross `/`, `**` does, a
+            // leading `/` anchors to the repo root, a pattern without `/`
+            // matches at any depth, and a trailing `/` matches a directory.
+            const rules = fs.readFileSync('base-codeowners.txt', 'utf8')
+              .split('\n')
+              .map(line => line.trim())
+              .filter(line => line && !line.startsWith('#'))
```
gbarkhatov: This filters only lines that *start* with `#`, then :188 splits the remainder on whitespace and treats every token as an owner. Inline comments are legal CODEOWNERS syntax — GitHub's canonical example file contains:

```
*.js    @js-owner #This is an inline comment.
```

So annotating a critical entry — `/…/vaultActivationService.ts @a @b @c @d # section 5` — yields owners `['@a','@b','@c','@d','#','section','5']` and :224 throws ``Only @user owners are supported for critical paths, got "#"``. Every PR touching that path is then blocked, with an error pointing at teams rather than at the comment. Self-inflicted DoS triggered by valid syntax.

Fix: `line.replace(/#.*$/, '').trim()` before splitting. (Docs confirm there's no `\`-escape for `#`, so stripping is safe.)

.github/workflows/critical-path-check.yml
```diff
@@ -125,3 +155,118 @@ jobs:
...
+                  source += '[^/]*';
+                } else if (ch === '?') {
+                  source += '[^/]';
+                } else {
+                  source += ch.replace(/[.+^${}()|[\]\\]/g, '\\$&');
+                }
+              }
+              const prefix = anyDepth ? '(?:^|.*/)' : '^';
+              // A file or directory pattern also owns everything beneath it.
+              return new RegExp(`${prefix}${source}(?:/.*)?$`);
```
gbarkhatov: The `(?:/.*)?$` suffix is appended to **every** pattern, including wildcard ones. Executed:

```
globToRegex('docs/*')  ->  /^docs\/[^/]*(?:\/.*)?$/
  .test('docs/build-app/troubleshooting.md')  ->  true
```

The docs say the opposite, verbatim: *"The `docs/*` pattern will match files like `docs/getting-started.md` but **not** further nested files like `docs/build-app/troubleshooting.md`."*

The error direction is **permissive**: with last-match-wins at :219-221, a wildcard rule sitting after a specific critical-path rule silently *widens* that path's owner set, so approvals from people GitHub does not consider owners would satisfy the gate. Latent today only because every current entry is a literal path.

Apply the suffix only when `directory === true` (the variable is computed at :195 and then never used). Separately, `**/logs` compiles to `^.*\/logs…` and misses a root-level `logs/` — gitignore says it should match, though GitHub's docs never state the root case, so lower confidence on that half.

.github/workflows/critical-path-check.yml
```diff
@@ -125,3 +155,118 @@ jobs:
...
+            };
+            const patternMatches = (pattern, file) => pattern === '*' || globToRegex(pattern).test(file);
```
gbarkhatov: The `pattern === '*'` short-circuit is unreachable-in-effect — `globToRegex('*')` already matches every path:

```
globToRegex('*')  ->  /(?:^|.*\/)[^/]*(?:\/.*)?$/
  .test('a/b/c.ts')  ->  true
```

Dead code per the ZERO DEAD CODE POLICY; delete it.

.github/workflows/critical-path-check.yml
```diff
@@ -125,3 +155,118 @@ jobs:
...
+            const qualified = new Set();
+            for (const file of matched) {
```
gbarkhatov: `qualified` is a single `Set` unioned across **all** matched files, and :252 accepts any approver in that union — so N approvals can come entirely from the owners of one file.

With owners `{alice,bob}` on file A and `{carol,dave}` on file B, a PR touching both is satisfied by alice + bob, and nobody who owns B has looked at it. That's strictly weaker than GitHub's native code-owner rule, where each owning party must approve.

Latent today because all 26 critical entries list the same four users — but per-path ownership is exactly what a nine-group inventory exists to enable, so this opens silently the first time an owner set diverges. Build `Map<file, Set<owner>>` and require `required` approvals from each matched file's own owner set, reporting the worst-covered path.

.github/workflows/critical-path-check.yml
```diff
@@ -125,3 +155,118 @@ jobs:
...
+                if (patternMatches(rule.pattern, file)) owners = rule.owners;
+              }
+              for (const owner of owners) {
+                if (!owner.startsWith('@') || owner.includes('/')) {
+                  throw new Error(`Only @user owners are supported for critical paths, got "${owner}" for ${file}`);
+                }
+                qualified.add(owner.slice(1).toLowerCase());
+              }
+            }
+            if (qualified.size < required) {
```
gbarkhatov: This reachability guard counts the PR author, who is then excluded from the approvals at :245 — so it misses the exact case it exists to catch.

`author` isn't defined until :240. With a two-owner path, `required=2`, and one of those owners authoring, `qualified.size === 2` sails past this check; the job then fails with the generic `1/2 code-owner approvals` instead of the accurate "2 approvals can never be reached".

Compute `author` before this line and size `qualified` minus the author.

.github/workflows/critical-path-check.yml
```diff
@@ -125,3 +155,118 @@ jobs:
...
+              if (!review.user || review.state === 'COMMENTED') continue;
+              const login = review.user.login.toLowerCase();
+              if (login === author) continue;
+              standing.set(login, review);
+            }
+
+            const approvers = [];
+            const stale = [];
+            for (const [login, review] of standing) {
+              if (review.state !== 'APPROVED' || !qualified.has(login)) continue;
```
gbarkhatov: This only ever tests for `APPROVED`. A standing `CHANGES_REQUESTED` from a qualified owner is kept in `standing` at :246 and then silently skipped here, so two *other* owners turn the check green over an unresolved objection.

Low in practice — the ruleset blocks merge on requested changes independently — but worth noting that issue #2359's "unresolved review threads block merge" is satisfied by admin config, not by this code. A one-line `if (review.state === 'CHANGES_REQUESTED') core.setFailed(...)` would make the check self-contained.

.github/CODEOWNERS
```diff
@@ -5,12 +5,16 @@
...
-# code-owner approval. To enforce two approvals specifically on these paths,
-# configure a GitHub ruleset scoped to these path globs with
-# `required_approving_review_count: 2`. Do not weaken ownership here without
-# review.
+# approval count - it only assigns owners and (with "Require review from
+# Code Owners" on in the ruleset) requires at least one code-owner approval.
+# Branch rulesets cannot be scoped by path, so the two-approval requirement
+# is enforced by .github/workflows/critical-path-check.yml: its `detect` job
+# is a required status check (repository ruleset "critical-pr-safety-checks")
+# and reads the owners of the touched critical paths from THIS file as it is
```
gbarkhatov: **Merge blocker — this file is not itself gated, so one approval can appoint the reviewers for every future critical-path PR.**

Reading owners from the base branch is the right trust boundary, and it holds for the PR *being* gated. But `.github/CODEOWNERS` appears neither in `critical-path-check.yml`'s `paths=` array nor as a non-default entry below, so a PR whose only change is this file never sets `matched=true` and merges on the ruleset's one-approval floor.

Concrete: PR A adds `@attacker` to all nine critical entries and merges with one approval. PR B then touches `payout.ts` and reaches 2/2 with the attacker plus one real owner — the two-qualified-reviewer property is now protected by a one-reviewer gate.

Unlike the self-edit hole the PR body accepts as residual, **this half closes completely**: because the file is read from base, a PR editing it would be gated by the base copy. Add `.github/CODEOWNERS` (and `.github/workflows/critical-path-check.yml`) to `paths=` and to `claude-md-drift.yml`.

.github/CODEOWNERS
```diff
@@ -5,12 +5,16 @@
...
 # irreversible value movement. CODEOWNERS alone does NOT enforce a minimum
-# approval count - it only assigns owners and (when "Require review from
-# Code Owners" is enabled in branch protection) requires at least one
-# code-owner approval. To enforce two approvals specifically on these paths,
-# configure a GitHub ruleset scoped to these path globs with
-# `required_approving_review_count: 2`. Do not weaken ownership here without
-# review.
+# approval count - it only assigns owners and (with "Require review from
+# Code Owners" on in the ruleset) requires at least one code-owner approval.
+# Branch rulesets cannot be scoped by path, so the two-approval requirement
```
gbarkhatov: **Merge blocker — `SECURITY.md` still documents the opposite mechanism, and the repo's own rule forbids shipping that drift.**

`SECURITY.md:690-693` currently reads:

> `critical-path-check.yml` comments and labels PRs touching critical paths but **does not block**. Enforcement of the two-approval rule requires a GitHub ruleset scoped to those path globs with `required_approving_review_count: 2` — CODEOWNERS alone cannot express it, as the comment in `.github/CODEOWNERS` notes. Verify that ruleset exists; the file cannot.

Every clause is now wrong, and it cross-references the very comment this PR rewrote. `SECURITY.md:906-909` names this file, CLAUDE.md, `critical-path-check.yml` and `claude-md-drift.yml` as five hand-maintained inventories and says **"Change them together. They have drifted before."**

An on-call or auditor following SECURITY.md would conclude the gate is advisory and merge a critical-path change on one approval — or go build the path-scoped ruleset this PR proves is impossible. Update `SECURITY.md:690-693` here.

(For the record, the attack-matrix row at `SECURITY.md:754` is *fine* — its "the existence check does not gate merges" refers to `claude-md-drift.yml`, not to this gate.)

.github/CODEOWNERS
```diff
@@ -5,12 +5,16 @@
...
+# is enforced by .github/workflows/critical-path-check.yml: its `detect` job
+# is a required status check (repository ruleset "critical-pr-safety-checks")
+# and reads the owners of the touched critical paths from THIS file as it is
+# on the PR's base branch, with the approval count taken from the repository
+# Actions variable CRITICAL_PATH_REQUIRED_APPROVALS. Only `@user` owners are
+# supported on the critical entries below (no teams). Do not weaken
+# ownership here without review.
 # -----------------------------------------------------------------------------
 
 # 1. WASM boundary (value computation)
```
gbarkhatov, line 16: Coverage gap worth closing while this inventory is being turned into merge enforcement: `packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/refund.ts` is named in CLAUDE.md §1 as the sole consumer of the unguarded `@babylonlabs-io/babylon-tbv-rust-wasm/raw` crossing, with its cross-checks (scriptPubKey/HTLC-value comparison, 1-in/1-out re-parse) spelled out as binding.

It is in none of the inventories — not here, not in `paths=`, not in `claude-md-drift.yml`. A PR stripping those assertions draws no comment, no label, and one approval.

Pre-existing and arguably a CLAUDE.md bug (it's named in prose rather than in a `Files:` bullet), but this is the PR that makes the inventory load-bearing.

.github/workflows/verify.yml
```diff
@@ -2,7 +2,7 @@ name: Verify PR
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, 'release/**']
```
gbarkhatov, line 5: Same missing-`edited` exposure as `critical-path-check.yml:20`: with the default activity types, retargeting a PR's base branch fires no re-run, so `--base="origin/${GITHUB_BASE_REF}"` at :94 stays pinned to the *old* base and `nx affected` reports against the wrong diff.

To be clear the `--base` change itself is a genuine fix and the right one — `GITHUB_BASE_REF` is always populated for `pull_request`, and `fetch-depth: 0` makes checkout fetch `+refs/heads/*:refs/remotes/origin/*`, so `origin/release/x` resolves. The hardcoded `origin/main` it replaces was flatly wrong for any `release/**` PR.

gbarkhatov (dismissed):
## Combo review — 3 independent passes, deduped and verified

Approving as requested so this isn't left blocking, but **flagging three merge-blockers explicitly** — please treat this approval as "reviewed", not "ship it".

### Merge-blockers

1. **None of the gate logic has ever executed.** `paths=` has no `.github/` entry, so this PR matched nothing. Live job `99118573118` shows `No critical-path files touched.` with both github-script steps **`skipped`**. The green `detect` here is not evidence the new code works, and inline YAML can't be run locally either.
2. **`.github/CODEOWNERS` is not itself gated**, so a PR that only edits it merges on the one-approval floor and can appoint the reviewers for every later critical-path PR. This half closes completely, unlike the self-edit residual the description accepts.
3. **`SECURITY.md:690-693` still says the opposite of what ships** ("does not block"; "requires a ruleset scoped to those path globs"), and `SECURITY.md:906-909` requires these five inventories to change together.

### Also worth fixing

The glob compiler over-matches in the **permissive** direction (`docs/*` matches nested files — executed, and contradicts the docs verbatim); owners are unioned across files rather than checked per-file; a legal inline `#` comment in CODEOWNERS hard-throws the gate repo-wide; `parseInt("1e3") → 1` silently drops the requirement to one approval; no `concurrency` plus an unpaginated `listComments`; and `pull_request_review` carries no branch filter. Details inline.

### Checked and explicitly *not* an issue

- **The trust-boundary paragraph at :10-18 is correct.** The docs confirm it: *"The `pull_request` event (along with `pull_request_review` …) is unusual: it runs the workflow file from the merge commit of the pull request."* The self-edit hole is real and honestly described.
- **The three-dot diff at :81 is sound** against the merge ref — `merge-base(origin/BASE, M)` is the base parent even when `origin/BASE` has advanced, so the diff is exactly this PR's changes. (The `ref:` pin at :37-39 is redundant, though — the docs give `pull_request_review` an identical `GITHUB_SHA`/`GITHUB_REF` row — and its comment misleads.)
- **Review aggregation matches the REST contract.** Docs say reviews return in chronological order, so last-write-wins on `standing` is genuinely the latest verdict; the `COMMENTED` skip, author exclusion, null-`user` guard and `commit_id === head.sha` staleness demotion are all right, and `github.paginate` handles >100 reviews.
- Dropping `|| true` from `git fetch` makes it fail closed; `BASE_REF` reaches the shell via `env:` so there's no injection; both `uses:` are full-SHA pinned; inventory coverage of all nine CLAUDE.md groups is complete and all 26 paths resolve to the four expected `@user` owners.

Not delivered by this PR from #2359's "Complete when": unresolved threads blocking merge, and the exception process with an owner and audit record — confirm those are covered by the live ruleset or split them out.

.github/workflows/critical-path-check.yml
```diff
@@ -53,76 +78,120 @@ jobs:
             "services/vault/src/utils/btc/scriptPubKeyAddress.ts"
           )
 
-          git fetch origin main || true
-          changed=$(git diff --name-only "origin/${GITHUB_BASE_REF}...HEAD")
+          git fetch origin "$BASE_REF"
+          git diff --name-only -z "origin/${BASE_REF}...HEAD" > changed.txt
```
jrwbabylonlab: `git diff` folds a rename into just the new path, so renaming a critical file (e.g. `refund.ts` → `refund2.ts`) never lands the old path in `changed.txt` and the gate stays green. `--no-renames` (or matching both sides) closes it.

.github/CODEOWNERS
```diff
@@ -3,22 +3,29 @@
...
-# `required_approving_review_count: 2`. Do not weaken ownership here without
-# review.
+# The required `detect` check reads this file from the base branch.
+# CRITICAL_PATH_REQUIRED_APPROVALS sets the required count per critical path.
+# Use individual @user owners on these paths. Branch rulesets cannot select
+# files. #2359 requires a trusted external gate before the issue can close.
 # -----------------------------------------------------------------------------
 
+# Review policy changes as critical changes.
+/.github/CODEOWNERS @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov
```
jrwbabylonlab: Heads-up for the conflict with main: #2436 added @kirugan to every rule after this branch forked — keeping this side as-is would drop them everywhere.

jonybur (commented):
Reviewed at a63db0ca. The workflow half has had a lot of attention already and I have nothing to add to those threads. The half nobody has looked at is the one I want to raise: since ac599d48 merged fix/2386-wasm-pegin-output-layout into this branch, this PR also changes the SDK critical paths, and every inline comment on the thread so far sits on a .github file.

**1. The SDK critical-path changes are undeclared and unverified.** Against main this PR now modifies assertWasmPeginSizing.ts, psbt/constants.ts, pegin.ts, peginInput.ts and refund.ts — five of the seven files CLAUDE.md lists under CRITICAL PATHS section 1 — plus utils/fee/constants.ts and fundPeginTransaction.ts from section 2. The description says only "Protect review policy and refund validation", and the Verification section lists actionlint, approval-record replay and inventory counts: no SDK test run, no fee-model cross-check for the section 2 edit, and no statement that an author can walk every changed line. Section 1 requires two reviewers on exactly this material. A PR that builds the two-approval gate is the last one that should carry an unannounced payload through it — and note the gate could not have flagged this itself, since detect reads the base-branch inventory and this branch predates its own enforcement.

**2. deriveExpectedPrePeginHtlc is a section 9 reimplementation shipping without the differential test that section requires.** The new function in assertWasmPeginSizing.ts rebuilds the Pre-PegIn HTLC — both tapleaf scripts, the taptree, the output key and the control block — in TypeScript with bitcoinjs-lib and noble-curves, replacing getPrePeginHtlcConnectorInfo as the source of the bytes that get signed. peginInput.ts now puts the derived hashlock script, control block and merkle root straight into the PSBT and keeps the facade call only for its side-effect checks; refund.ts does the same. That is squarely section 9, whose rule is explicit: a differential test asserting byte-for-byte equality against the implementation it replaces, over the existing golden vectors **plus randomised inputs**, with a single hardcoded vector called out as insufficient.

What is here does not meet it:

- assertWasmPeginSizing.test.ts mocks createPrePeginTransaction and builds the stand-in WASM htlcScriptPubKeys by calling deriveExpectedPrePeginHtlc itself, so the layout assertions compare the derivation against the derivation. A wrong script would pass every one of them.
- The new peginInput.test.ts case mocks getPrePeginHtlcConnectorInfo with deliberately forged values. That proves independence from the facade, which is worth having, but it is the opposite of proving agreement with it.
- The only real cross-check left is the happy-path fixture, where the derived scriptPubKey must equal the funded HTLC output for the build to succeed. That is one parameter set — exactly the "pins one input, not the function" case the rule names.

The gap matters most where the derivation encodes an assumption the diff does not cite. Every constant added to psbt/constants.ts carries a btc-vault crates/vault reference; the script construction in deriveExpectedPrePeginHtlc carries none — not the opcode sequence, not the OP_CHECKSIGADD n-of-n shape, not the OP_NUMEQUALVERIFY versus OP_NUMEQUAL split between keepers and challengers, not the hashlock-then-refund leaf order, and not the sort inside normalizeKeyGroup. On the sort specifically: RefundPrePeginContext already documents that keeper and challenger arrays "must be pre-sorted the same way the Rust protocol sorts them", so sorting here is plausibly right, and lowercased 64-character hex does sort byte-identically — but the fixture cannot tell you that, and the contract has just moved from the caller into this function without a note. A differential over varied keeper and challenger counts, orderings and timelock values against the real facade would settle all of it at once.

To be clear about direction: deriving the signing data independently instead of trusting the facade is the right change and is what section 1 asks for. It is the evidence that is missing, not the intent.

**3. CI is not green on this head.** Only the CodeQL and GitGuardian checks have reported on a63db0ca; verify, validate, detect and visual have not run. With the SDK now in scope, verify also has to clear its generated-docs staleness gate — docs/api/utils.md is in the diff, so the regeneration looks done, but it is unconfirmed.

The cleanest resolution is to split the SDK commits back out into their own PR against main, where they get two reviewers and the section 9 differential on their own merits, and let this one stay the workflow change it says it is.

.github/workflows/critical-path-check.yml
```diff
@@ -53,76 +78,120 @@ jobs:
...
+              let owners = [];
+              for (const [pattern, ...users] of rules) {
+                if (globToRegex(pattern).test(file)) owners = users;
+              }
+              if (owners.some(owner => !/^@[a-z\d-]+$/i.test(owner))) {
+                throw new Error(`Only @user owners are supported for ${file}: ${owners.join(', ')}`);
+              }
+              const qualified = [...new Set(owners.map(owner => owner.slice(1).toLowerCase()))]
+                .filter(login => login !== author);
+              const approvers = qualified.filter(login => {
```
jeremy-babylonlabs, line 165: This check runs the PR's own version of the workflow (standard `pull_request` trigger), so a PR that edits this file can neutralise its own gate. That is the residual risk #2359 tracks (trusted external gate, no execution of PR-controlled code).

Worth a note here or in the PR description linking that as the known-open item, so reviewers do not read `detect` passing as a completed control.

jeremy-babylonlabs (commented):
Approval math looks right and fails closed: stale approvals excluded via `review.commit_id === pr.head.sha`, self-approval filtered, `CHANGES_REQUESTED` from a qualified owner blocks, an unset `CRITICAL_PATH_REQUIRED_APPROVALS` throws, CODEOWNERS resolved from `origin/$BASE_REF`, actions pinned by SHA, `permissions:` least-privilege. No `pull_request_target` / `workflow_run` escalation. One thing worth recording so the check is not over-read.

.github/workflows/critical-path-check.yml
```diff
@@ -53,76 +78,120 @@ jobs:
...
             for p in "${paths[@]}"; do
-              case "$f" in
-                "$p"|"$p"*) matched+=("$f"); break ;;
-              esac
+              if [[ "$f" == "$p" || ( "$p" == */ && "$f" == "$p"* ) ]]; then
+                matched+=("$f"); break
+              fi
             done
-          done
+          done < changed.txt
```
gbarkhatov, line 94: **Merge-blocker.** The changed-file list is written into the PR's own checkout: `> changed.txt` at line 82, read back at line 94. A PR that commits a root-level symlink `changed.txt -> /dev/null` gets through with zero code-owner approvals. Git stores the link and `actions/checkout` recreates it on Linux, so the `>` writes into `/dev/null` and the `while read` loop hits EOF straight away. Then `matched` stays empty, line 102 sets `matched=false`, the approval step is skipped, and `detect` goes green, even if the same PR rewrites `refund.ts`.

The PR already accepts that a PR can edit this workflow, and that gives the same power. The difference is that this route touches no gated file: neither the workflow nor CODEOWNERS changes, so the diff shows only an innocuous-looking `changed.txt`. It would also survive the trusted external gate #2359 asks for, if that gate reuses this script on a PR checkout. This came in with `0250d49ea`; at `bef64a7b` the list lived in a shell variable. The other two scratch files fail closed, but only by luck: a `/dev/null` `matched.txt` hits the throw at line 122, and an empty `base-codeowners.txt` fails every file at line 170.

Fix: write all three files under `"$RUNNER_TEMP"` and read them via `process.env.RUNNER_TEMP` in the script step, or feed the loop directly with `done < <(git diff …)`. `set -o noclobber` doesn't help, because it only protects regular files. (I reasoned this from redirection semantics and did not run it.)

.github/workflows/critical-path-check.yml
```diff
@@ -53,76 +78,120 @@ jobs:
...
+            for (const file of matched) {
+              let owners = [];
+              for (const [pattern, ...users] of rules) {
+                if (globToRegex(pattern).test(file)) owners = users;
+              }
+              if (owners.some(owner => !/^@[a-z\d-]+$/i.test(owner))) {
+                throw new Error(`Only @user owners are supported for ${file}: ${owners.join(', ')}`);
+              }
+              const qualified = [...new Set(owners.map(owner => owner.slice(1).toLowerCase()))]
+                .filter(login => login !== author);
```
gbarkhatov, line 164: "Author excluded" only excludes the account that opened the PR (`pr.user.login`, line 120). A qualified owner who wrote commits in the PR still counts as an independent approver. This PR shows the shape: every commit since my earlier review was authored by an account that is itself a qualified owner of both gated paths. That account plus one other owner reaches 2/2 at line 171, which leaves one reviewer independent of the code's authors, against CLAUDE.md's two-reviewer rule.

The ruleset's `require_last_push_approval` doesn't close this. It only bars the last pusher from the ruleset's own one-approval floor, and it never feeds this count. The opener-only logic dates from `0ce184743`, so the delta made it concrete rather than introducing it.

Fix: also drop from `qualified` every `author.login` and `committer.login` returned by `github.paginate(github.rest.pulls.listCommits, …)`.

.github/workflows/critical-path-check.yml
```diff
@@ -53,76 +78,120 @@ jobs:
...
+              for (let i = 0; i < body.length; i += 1) {
+                const ch = body[i];
+                if (ch === '*' && body[i + 1] === '*') {
+                  source += body[i + 2] === '/' ? '(?:.*/)?' : '.*';
+                  i += body[i + 2] === '/' ? 2 : 1;
+                } else if (ch === '*') source += '[^/]*';
+                else if (ch === '?') source += '[^/]';
+                else source += ch.replace(/[.+^${}()|[\]\\]/g, '\\$&');
+              }
+              const suffix = pattern.endsWith('/') || !/[*?]/.test(body) ? '(?:/.*)?' : '';
```
gbarkhatov, line 139: Low, latent. Following up on #10: the fix now under-matches. Any pattern containing `*` or `?` loses the `(?:/.*)?` suffix, so `**/logs` compiles to `^(?:.*/)?logs$` and owns nothing inside a `logs/` directory. `build/logs/app.log` and `logs/app.log` both test false, and `/a/*/logs` misses `a/b/logs/x.log`. GitHub's CODEOWNERS docs say `**/logs` owns "any file in a `/logs` directory such as `/build/logs`". Combined with last-match-wins at line 158, a later `**/<dir>` rule that reassigns a critical directory would be silently ignored. Base CODEOWNERS has no wildcard rows besides `*` today, so nothing is wrong yet. My earlier suggestion had the same gap.

Fix: withhold the suffix only when the last path segment contains a wildcard: `/[*?]/.test(body.split('/').pop())`.

gbarkhatov (approved):
Follow-up to my review of `bef64a7b`. Since then the branch took the unsquashed #2398 lineage (`993677fb3`, a merge of main, `a9f9bcab6`) and four CI commits (`0250d49ea`, `95f645554`, `a63db0ca2`, and the governance half of `a9f9bcab6`). None of my 18 threads has a reply, so every verdict below is judged from the code at HEAD.

**Headline:** 16 of the 18 are fixed and 2 are partly fixed. The rewrite is a real improvement: the gate now checks each file separately, reads owners from the base branch, drops stale and dismissed approvals, and blocks on a standing change request. The new code adds two gate defects, though. One lets `detect` pass with zero approvals without touching any gated file, and I'd treat that as a **merge-blocker** (inline).

**The ts-sdk / §1 content is not this PR's change.** `git diff 4b0861989 a63db0ca -- <all 20 ts-sdk and vitest paths, pnpm-lock.yaml>` is empty: it is byte-identical to the #2398 squash that's already on main. That's also why the PR conflicts. `git merge-tree` shows conflicts in `verify.yml`, `CLAUDE.md`, `SECURITY.md` and `refund.test.ts`, because main edited #2398's hunks afterwards (#2464 and #2463). **Resolve all four to main's side.** Taking the PR's side would delete main's refund tests (the "wrong input transaction" block, a CLAUDE.md §1 file's tests), drop #2464's raw-deprecation text from CLAUDE.md and SECURITY.md, and drop main's browser suites from `verify.yml`. The cleanest route is to rebase onto main and drop the #2398 commits, which leaves this PR governance-only.

|---|---|---|---|
| 1 | The gate never ran (no `.github/` entry in `paths=`) | **Partly fixed.** Both entries were added, and at `95f645554` `detect` failed with per-path "0/2" annotations, so the red path is proven. Going green on the Nth approval and flipping back on a dismissal are still unproven. There's no run at HEAD because the PR conflicts; the PR body already lists this | `critical-path-check.yml:43-44` |
| 2 | Unquoted `$changed`, a dead `case` alternative, no `pipefail` | Fixed (`-z` / `read -d ''`, `--no-renames`) | `:38, :82, :88-94` |
| 3 | No `concurrency`; `listComments` unpaginated | Fixed; the marker lookup is also limited to `github-actions[bot]` | `:18-20, :184-187` |
| 4 | No `edited` type, so a base retarget doesn't re-run | Fixed | `:10` |
| 5 | `pull_request_review` has no branch filter | Fixed (job-level `if:`) | `:24` |
| 6 | A `createComment` 403 kills the job before the gate runs | Fixed. Comment and label now run after the evaluation, inside try/catch | `:181-197` |
| 7 | `REQUIRED_APPROVALS` was rendered before validation | Fixed | `:114-118` |
| 8 | `parseInt` accepted `1e3` and `2abc` | Fixed | `:114-118` |
| 9 | An inline `#` comment in CODEOWNERS threw | Fixed | `:124` |
| 10 | The glob suffix over-matched `docs/*`, and `**/logs` missed the root | **Partly fixed.** `docs/*` is right now, but the fix swings the other way (inline) | `:139` |
| 11 | The `pattern === '*'` short-circuit was dead code | Fixed. `*` still compiles to match-all | `:126-141` |
| 12 | Owners were unioned across all matched files | Fixed (per-file loop and verdict) | `:155-175` |
| 13 | The reachability check counted the author | Fixed | `:163-164, :170` |
| 14 | A standing `CHANGES_REQUESTED` was ignored | Fixed | `:172-174` |
| 15 | `.github/CODEOWNERS` wasn't itself gated (blocker) | Fixed, and proven live at `95f645554` | `CODEOWNERS:13-14`, `:43-44` |
| 16 | SECURITY.md described the opposite mechanism (blocker) | Fixed | `SECURITY.md:719-724` |
| 17 | `refund.ts` was in no inventory | Fixed, via #2398, which is already on main | `CODEOWNERS:24`, `:52` |
| 18 | `verify.yml` was missing `edited` | Fixed | `verify.yml:6` |

**Cost note, not a defect:** with `edited` in `verify.yml`, every title or body edit cancels and restarts the full `verify` run; one run at `95f645554` was cancelled this way. Please don't fix it with a job-level `if:`, since a skipped required job reports Success.

**Refuted:**
- *"The PR changes §1 code beyond #2398."* The diff against `4b0861989` is empty for every ts-sdk path it touches.
- *"`persist-credentials: false` breaks `git fetch origin $BASE_REF`."* The `95f645554` runs got past the fetch and produced per-path annotations.
- *"`contents: read` breaks verify's release rehearsal."* `verify` succeeded at `95f645554`.
- *"The inventory drops a critical path main has since added."* `paths=` differs from main by exactly the two `.github/` rows, and CODEOWNERS rows match main's.

**Verified clean:**
- Owners come from `origin/$BASE_REF`.
- Approvals must be on `pr.head.sha`, and a DISMISSED review overwrites an earlier approval.
- `--no-renames` puts both paths of a move into the diff.
- `BASE_REF` and `REQUIRED_APPROVALS` arrive through `env:`, so there's no expression injection.
- Both actions are pinned by SHA.
- The "36 entries agree across the inventories" claim holds at HEAD.

gbarkhatov (approved):

gbarkhatov (approved):

.github/CODEOWNERS
```diff
@@ -1,61 +1,68 @@
...
-/services/vault/src/utils/btc/scriptPubKeyAddress.ts @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov
+/packages/babylon-ts-sdk/src/tbv/core/clients/eth/onChainBtcPubkey.ts @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov @kirugan
+/packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-transaction.ts @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov @kirugan
+/packages/babylon-ts-sdk/src/tbv/core/clients/eth/pegin-registration-client.ts @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov @kirugan
+/packages/babylon-ts-sdk/src/tbv/core/clients/eth/payout-script.ts @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov @kirugan
+/packages/babylon-ts-sdk/src/tbv/core/wasm/ @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov @kirugan
+/packages/babylon-tbv-rust-wasm/src/wasm-loader.ts @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov @kirugan
+/packages/babylon-tbv-rust-wasm/src/wasm-loader-node.ts @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov @kirugan
+/packages/babylon-tbv-rust-wasm/src/raw.ts @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov @kirugan
```
jrwbabylonlab: This conflicts with main: `raw.ts`/`raw-node.ts` were deleted there, and main added owner rows for `services/delegated-claim/` and `rust-wasm/src/delegatedClaim.ts`. Taking this side drops owners from those critical paths. Resolve to main's rows (and ideally rebase so the #2398 SDK commits drop out).

PR #2378, opened 2026-08-31
feat(wallet-connector, ledger-vault): make Refund signable on device

jrwbabylonlab:
Refund was unreachable for a Ledger depositor; both blockers were host-side — the device signs a standalone refund from any vault state (fw `ff1e1ce17`).

- New `refundPsbt.ts` in ledger-vault-signer: refund-leaf parser mirroring the firmware grammar byte-for-byte, a 1-in/1-out classifier, and `augmentPsbtForRefund` adding the two required TAP_BIP32_DERIVATION entries — input 0 keyed by the **untweaked** depositor key, output 0 by the **BIP-86 tweaked** witness program. Enforces the device's CSV floor (72) before any I/O.
- Provider: `signPsbt` classifies the refund from its own parse (never a caller flag) and waives only the intent gate; a wrong-vault guard rejects a refund not matching a loaded intent's terms with a typed error; refunds are exempt from the replay guard (the standalone path consumes no device mask); `signPsbts` refuses a batched refund.

No SDK changes — augmentation is provider-side, like `augmentPsbtForWalletPolicy`.

### DIFF
--- a/packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
@@ -18,7 +18,12 @@ import { Buffer } from "buffer";
 import { describe, expect, it } from "vitest";
 
 import { HARDENED } from "../bip86Path";
-import { augmentPsbtForRefund, classifyRefundPsbt, parseRefundLeafScript } from "../refundPsbt";
+import {
+  assertRefundPsbtSignable,
+  augmentPsbtForRefund,
+  classifyRefundPsbt,
+  parseRefundLeafScript,
+} from "../refundPsbt";
 
 initEccLib(ecc);
 
@@ -64,20 +69,25 @@ function buildRefundShapedPsbt(
     leaves?: readonly Buffer[];
     leafVersion?: number;
     outScript?: Buffer;
+    outValue?: number;
     extraOutput?: boolean;
     extraInput?: boolean;
     dropLeaf?: boolean;
+    dropWitnessUtxo?: boolean;
+    version?: number;
+    locktime?: number;
+    sequence?: number;
   } = {},
 ): string {
   const leaf = overrides.leaf ?? refundLeaf(DEPOSITOR_XONLY, PUSH_2016);
   const psbt = new Psbt();
-  psbt.setVersion(2);
-  psbt.setLocktime(0);
+  psbt.setVersion(overrides.version ?? 2);
+  psbt.setLocktime(overrides.locktime ?? 0);
   psbt.addInput({
     hash: PREV_HASH,
     index: 0,
-    sequence: CSV_2016,
-    witnessUtxo: { script: htlcSpk, value: HTLC_VALUE_SATS },
+    sequence: overrides.sequence ?? CSV_2016,
+    ...(overrides.dropWitnessUtxo ? {} : { witnessUtxo: { script: htlcSpk, value: HTLC_VALUE_SATS } }),
     ...(overrides.dropLeaf
       ? {}
       : {
@@ -107,7 +117,7 @@ function buildRefundShapedPsbt(
       tapInternalKey: Buffer.from(NUMS_XONLY, "hex"),
     });
   }
-  psbt.addOutput({ script: overrides.outScript ?? depositorSpk, value: REFUND_VALUE_SATS });
+  psbt.addOutput({ script: overrides.outScript ?? depositorSpk, value: overrides.outValue ?? REFUND_VALUE_SATS });
   if (overrides.extraOutput) {
     psbt.addOutput({ script: depositorSpk, value: 1_000 });
   }
@@ -210,7 +220,7 @@ describe("parseRefundLeafScript (firmware-grammar mirror)", () => {
 describe("classifyRefundPsbt", () => {
   it("classifies a 1-in/1-out PSBT carrying one refund leaf, returning the leaf terms and prevout txid", () => {
     const classified = classifyRefundPsbt(buildRefundShapedPsbt());
-    expect(classified).toEqual({
+    expect(classified).toMatchObject({
       leafKeyHex: DEPOSITOR_XONLY,
       csv: 2016,
       inputTxidInternalHex: PREV_HASH.toString("hex"),
@@ -224,14 +234,78 @@ describe("classifyRefundPsbt", () => {
     ["two TAP_LEAF_SCRIPT entries", { leaves: [refundLeaf(DEPOSITOR_XONLY, PUSH_2016), refundLeaf(FOREIGN_XONLY, PUSH_2016)] }],
     ["a non-refund leaf script", { leaf: Buffer.from([OP_PUSHBYTES_32, ...Buffer.alloc(32, 5), 0xac]) }],
     ["a non-tapscript leaf version", { leafVersion: 0xc2 }],
+    ["a version-1 transaction (the device requires version >= 2)", { version: 1 }],
+    ["a non-zero locktime", { locktime: 1 }],
   ] as const)("returns undefined for %s", (_label, overrides) => {
     expect(classifyRefundPsbt(buildRefundShapedPsbt(overrides))).toBeUndefined();
+  });
+
+  it("carries the sequence, witnessUtxo and output terms the validator pins", () => {
+    const classified = classifyRefundPsbt(buildRefundShapedPsbt());
+    expect(classified?.sequence).toBe(CSV_2016);
+    expect(classified?.witnessUtxo).toEqual({ scriptLength: 34, value: HTLC_VALUE_SATS });
+    expect(classified?.outputScriptHex).toBe(depositorSpk.toString("hex"));
+    expect(classified?.outputValue).toBe(REFUND_VALUE_SATS);
   });
 
   it("returns undefined for malformed hex instead of throwing (error precedence stays with the caller)", () => {
     expect(classifyRefundPsbt("zz")).toBeUndefined();
     expect(classifyRefundPsbt("abcd")).toBeUndefined();
     expect(classifyRefundPsbt("")).toBeUndefined();
+  });
+});
+
+describe("assertRefundPsbtSignable", () => {
+  function classified(overrides: Parameters<typeof buildRefundShapedPsbt>[0] = {}) {
+    const c = classifyRefundPsbt(buildRefundShapedPsbt(overrides));
+    if (c === undefined) throw new Error("fixture must classify as a refund");
+    return c;
+  }
+
+  it("accepts the canonical refund terms", () => {
+    expect(() => assertRefundPsbtSignable(classified())).not.toThrow();
+  });
+
+  it("accepts the device floor exactly (CSV 72)", () => {
+    expect(() =>
+      assertRefundPsbtSignable(classified({ leaf: refundLeaf(DEPOSITOR_XONLY, Buffer.from([0x01, 0x48])), sequence: 72 })),
+    ).not.toThrow();
+  });
+
+  it("rejects a CSV below the device floor of 72", () => {
+    expect(() =>
+      assertRefundPsbtSignable(classified({ leaf: refundLeaf(DEPOSITOR_XONLY, Buffer.from([0x01, 0x47])), sequence: 71 })),
+    ).toThrow(/72/);
+  });
+
+  it("rejects a sequence with the BIP-68 disable flag set", () => {
+    expect(() => assertRefundPsbtSignable(classified({ sequence: 0x80000000 + CSV_2016 }))).toThrow(/sequence/);
+  });
+
+  it("rejects a sequence with the BIP-68 time-based flag set", () => {
+    expect(() => assertRefundPsbtSignable(classified({ sequence: 0x00400000 + CSV_2016 }))).toThrow(/sequence/);
+  });
+
+  it("rejects a sequence that does not encode exactly the leaf CSV", () => {
+    expect(() => assertRefundPsbtSignable(classified({ sequence: 144 }))).toThrow(/sequence/);
+  });
+
+  it("rejects an input without a witnessUtxo", () => {
+    expect(() => assertRefundPsbtSignable(classified({ dropWitnessUtxo: true }))).toThrow(/witnessUtxo/);
+  });
+
+  it("rejects an output value above the HTLC value", () => {
+    expect(() => assertRefundPsbtSignable(classified({ outValue: HTLC_VALUE_SATS + 1 }))).toThrow(/exceed/);
+  });
+
+  it("rejects an output paying a FOREIGN P2TR address — ownership, not just shape", () => {
+    const foreignSpk = payments.p2tr({ internalPubkey: Buffer.from(FOREIGN_XONLY, "hex") }).output!;
+    expect(() => assertRefundPsbtSignable(classified({ outScript: foreignSpk }))).toThrow(/BIP-86/);
+  });
+
+  it("rejects a non-P2TR output", () => {
+    const p2wpkh = Buffer.concat([Buffer.from([0x00, 0x14]), Buffer.alloc(20, 7)]);
+    expect(() => assertRefundPsbtSignable(classified({ outScript: p2wpkh }))).toThrow(/BIP-86/);
   });
 });
 
@@ -273,8 +347,14 @@ describe("augmentPsbtForRefund", () => {
   });
 
   it("throws when the refund leaf key is not the depositor key", () => {
+    // A coherent foreign refund: leaf AND destination belong to the foreign key,
+    // so only the depositor-equality check can be what rejects it.
+    const foreignSpk = payments.p2tr({ internalPubkey: Buffer.from(FOREIGN_XONLY, "hex") }).output!;
     expect(() =>
-      augmentPsbtForRefund({ ...params, psbtHex: buildRefundShapedPsbt({ leaf: refundLeaf(FOREIGN_XONLY, PUSH_2016) }) }),
+      augmentPsbtForRefund({
+        ...params,
+        psbtHex: buildRefundShapedPsbt({ leaf: refundLeaf(FOREIGN_XONLY, PUSH_2016), outScript: foreignSpk }),
+      }),
     ).toThrow(/depositor/);
   });
 
@@ -284,10 +364,10 @@ describe("augmentPsbtForRefund", () => {
     );
   });
 
-  it("throws when output 0 is not P2TR (the device compares against the witness program)", () => {
+  it("runs the signability validator — an unowned output is rejected here too", () => {
     const p2wpkh = Buffer.concat([Buffer.from([0x00, 0x14]), Buffer.alloc(20, 7)]);
     expect(() => augmentPsbtForRefund({ ...params, psbtHex: buildRefundShapedPsbt({ outScript: p2wpkh }) })).toThrow(
-      /P2TR/,
+      /BIP-86/,
     );
   });
 
@@ -305,7 +385,7 @@ describe("augmentPsbtForRefund", () => {
     expect(() =>
       augmentPsbtForRefund({
         ...params,
-        psbtHex: buildRefundShapedPsbt({ leaf: refundLeaf(DEPOSITOR_XONLY, Buffer.from([0x01, 0x47])) }),
+        psbtHex: buildRefundShapedPsbt({ leaf: refundLeaf(DEPOSITOR_XONLY, Buffer.from([0x01, 0x47])), sequence: 71 }),
       }),
     ).toThrow(/72/);
   });
@@ -314,7 +394,7 @@ describe("augmentPsbtForRefund", () => {
     expect(() =>
       augmentPsbtForRefund({
         ...params,
-        psbtHex: buildRefundShapedPsbt({ leaf: refundLeaf(DEPOSITOR_XONLY, Buffer.from([0x01, 0x48])) }),
+        psbtHex: buildRefundShapedPsbt({ leaf: refundLeaf(DEPOSITOR_XONLY, Buffer.from([0x01, 0x48])), sequence: 72 }),
       }),
     ).not.toThrow();
   });
--- a/packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
@@ -1,0 +1,321 @@
+/**
+ * Refund-PSBT classification and augmentation (#2371).
+ *
+ * The leaf grammar mirrors the firmware's own parser byte-for-byte —
+ * `fw:sign_psbt_validate_helpers.c:77-148` (`parse_refund_leaf_script`) @
+ * `ff1e1ce17` — so every accept/reject vector here is a firmware-grammar pin,
+ * including the shapes the firmware is deliberately loose about (OP_PUSHDATA1,
+ * a non-minimal positive CScriptNum).
+ */
+
+// @vitest-environment node
+// Same rationale as provider.test.ts: the asmjs ECC backend fails bitcoinjs's
+// verifyEcc fixtures under jsdom but passes under node.
+
+import * as ecc from "@bitcoin-js/tiny-secp256k1-asmjs";
+import { initEccLib, payments, Psbt } from "bitcoinjs-lib";
+import { Buffer } from "buffer";
+import { describe, expect, it } from "vitest";
+
+import { HARDENED } from "../bip86Path";
+import { augmentPsbtForRefund, classifyRefundPsbt, parseRefundLeafScript } from "../refundPsbt";
+
+initEccLib(ecc);
+
+/** BIP-86 test-vector depositor key (same as the provider/e2e fixtures). */
+const DEPOSITOR_XONLY = "dc8d2f9eff0c4f4dbde070a48e330efc908b62a766568d91e658f284b324b878";
+const FOREIGN_XONLY = "cc8a4bc64d897bddc5fbc2f670f7a8ba0b386779106cf1223c6fc5d7cd6fc115";
+/** BIP-341 NUMS point — the HTLC's unspendable internal key. */
+const NUMS_XONLY = "50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0";
+const MASTER_FINGERPRINT = "73c5da0a";
+const DEPOSITOR_PATH: readonly number[] = [86 + HARDENED, 1 + HARDENED, 0 + HARDENED, 0, 0];
+
+const OP_PUSHBYTES_32 = 0x20;
+const OP_CHECKSIGVERIFY = 0xad;
+const OP_CSV = 0xb2;
+const TAPSCRIPT_LEAF_VERSION = 0xc0;
+
+/** Minimal CScriptNum push of 2016: `02 e0 07`. */
+const PUSH_2016 = Buffer.from([0x02, 0xe0, 0x07]);
+const CSV_2016 = 2016;
+
+const HTLC_VALUE_SATS = 1_000_000;
+const REFUND_VALUE_SATS = 990_000;
+const PREV_HASH = Buffer.alloc(32, 0x11);
+
+function refundLeaf(keyHex: string, csvPush: Buffer): Buffer {
+  return Buffer.concat([
+    Buffer.from([OP_PUSHBYTES_32]),
+    Buffer.from(keyHex, "hex"),
+    Buffer.from([OP_CHECKSIGVERIFY]),
+    csvPush,
+    Buffer.from([OP_CSV]),
+  ]);
+}
+
+const depositorSpk = payments.p2tr({ internalPubkey: Buffer.from(DEPOSITOR_XONLY, "hex") }).output!;
+/** Any P2TR-shaped previous output — classification never re-derives it. */
+const htlcSpk = Buffer.concat([Buffer.from([0x51, 0x20]), Buffer.alloc(32, 0x22)]);
+const controlBlock = Buffer.concat([Buffer.from([TAPSCRIPT_LEAF_VERSION]), Buffer.from(NUMS_XONLY, "hex")]);
+
+function buildRefundShapedPsbt(
+  overrides: {
+    leaf?: Buffer;
+    leaves?: readonly Buffer[];
+    leafVersion?: number;
+    outScript?: Buffer;
+    extraOutput?: boolean;
+    extraInput?: boolean;
+    dropLeaf?: boolean;
+  } = {},
+): string {
+  const leaf = overrides.leaf ?? refundLeaf(DEPOSITOR_XONLY, PUSH_2016);
+  const psbt = new Psbt();
+  psbt.setVersion(2);
+  psbt.setLocktime(0);
+  psbt.addInput({
+    hash: PREV_HASH,
+    index: 0,
+    sequence: CSV_2016,
+    witnessUtxo: { script: htlcSpk, value: HTLC_VALUE_SATS },
+    ...(overrides.dropLeaf
+      ? {}
+      : {
+          tapLeafScript: (overrides.leaves ?? [leaf]).map((script, i) => {
+            // bip174 requires controlBlock[0]'s version bits to equal leafVersion.
+            const versionedControlBlock = Buffer.concat([
+              Buffer.from([overrides.leafVersion ?? TAPSCRIPT_LEAF_VERSION]),
+              controlBlock.subarray(1),
+            ]);
+            return {
+              leafVersion: overrides.leafVersion ?? TAPSCRIPT_LEAF_VERSION,
+              script,
+              // Distinct per entry — bip174 keys TAP_LEAF_SCRIPT by control block.
+              controlBlock: i === 0 ? versionedControlBlock : Buffer.concat([versionedControlBlock, Buffer.alloc(32, 0x44)]),
+            };
+          }),
+        }),
+    tapInternalKey: Buffer.from(NUMS_XONLY, "hex"),
+  });
+  if (overrides.extraInput) {
+    psbt.addInput({
+      hash: Buffer.alloc(32, 0x33),
+      index: 1,
+      sequence: CSV_2016,
+      witnessUtxo: { script: htlcSpk, value: HTLC_VALUE_SATS },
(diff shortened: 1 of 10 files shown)


packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
```diff
@@ -1173,6 +1173,164 @@ describe("LedgerVaultProvider", () => {
...
+      ]);
+      const nums = Buffer.from(NUMS_XONLY_HEX, "hex");
+      const psbt = new Psbt();
+      psbt.setVersion(2);
+      psbt.setLocktime(0);
+      psbt.addInput({
+        hash: overrides.prevHashInternal ?? INTENT_TXID_INTERNAL,
+        index: 0,
+        sequence: csv,
+        witnessUtxo: { script: Buffer.concat([Buffer.from([0x51, 0x20]), Buffer.alloc(32, 0x22)]), value: 1_000_000 },
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — test philosophy prefers explicit inline values

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -0,0 +1,224 @@
...
+  if (classified.leafKeyHex !== depositorXOnlyHex) {
+    throw new Error("refund leaf key does not equal the depositor key — this device cannot sign this refund");
+  }
+  // The device never signs a refund below this floor in any state
+  // (`fw:sign_psbt_validate.c:898-903`, IDLE/HASH_DERIVED branch;
+  // `vault_constants.h:103`) — pre-empt its opaque SW_INCORRECT_DATA.
+  if (classified.csv < DEVICE_TIMELOCK_MIN_BLOCKS) {
+    throw new Error(
+      `refund leaf CSV ${classified.csv} is below the device's timelock floor of ${DEVICE_TIMELOCK_MIN_BLOCKS} blocks`,
+    );
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — matches sibling helpers; provider wraps it typed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -668,22 +689,91 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+   * itself. Every rejection before the device loop starts leaves the mirror
+   * and the loaded intent untouched.
    */
   signPsbt = async (psbtHex: string, options?: SignPsbtOptions): Promise<string> =>
     this.withDeviceOperation("signPsbt", () =>
       this.withSignAbort(async (controller) => {
-        const ctx = await this.gateSignContext();
-        const staged = await this.stagePsbt(psbtHex, options, "signPsbt", new Set(), ctx.depositorXOnlyHex);
+        const refund = classifyRefundPsbt(psbtHex);
+        const ctx = await this.gateSignContext(refund === undefined);
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — silent cached reads only; no ceremony I/O

jonybur (dismissed):
The classification path is a strict provider-side parse: exactly 1-in/1-out with one tapscript leaf whose script matches the firmware's refund grammar byte-for-byte (`sign_psbt_validate_helpers.c:77-148`), and the `undefined`-on-malformed contract preserves the caller's ordinary error precedence. The grammar mirror is deliberately as loose as the firmware — OP_PUSHDATA1 and non-minimal positive CScriptNums pass, and the sign-bit rejection caps `csv` at `0x7fffffff` so the shift-and-OR cannot overflow. Multi-input / multi-output / non-tapscript-leaf / wrong-leaf-version / no-tapleaf and hex-that-does-not-parse cases are all covered.

`augmentPsbtForRefund` re-classifies, verifies the leaf key equals the connected depositor key, checks the CSV floor of 72 (`DEVICE_TIMELOCK_MIN_BLOCKS`), and asserts output 0 is P2TR (`0x51 0x20` + 32-byte witness program) before writing the two `tapBip32Derivation` entries. Input 0 is keyed by the untweaked depositor key, output 0 by the witness-program bytes verbatim (BIP-371 permits the output key as the map key), matching the firmware's asymmetric derive-and-compare in `sign_psbt_validate.c:905-950` (input) and `:1005-1057` (output). Empty `leafHashes` is compatible with the firmware's value parser (`…helpers.c:58-63`). The unsigned tx is never touched (test pinned).

Provider-side, `assertRefundSignable` is the intent-loaded gate: when `phase === "intent-loaded"`, it rejects a refund whose prevout txid or CSV does not match `state.prepeginTxidInternalHex` / `state.htlcRefundTimelock`, pre-empting the device's opaque SW_INCORRECT_DATA from the on-device pins (`:893-897`, `:1076-1081`) with a typed error and preserving the loaded intent. `gateSignContext(false)` is scoped to the refund path via `refund === undefined`; the replay guard is threaded through `StagedPsbt.guardReplay` and correctly disabled only for the no-intent standalone refund, matching the firmware behaviour that the standalone path consumes no dedup mask. `signPsbts` rejects any batched refund at zero staging I/O.

E2E coverage under Speculos exercises the four discriminating shapes: augmentation asymmetry, sign-with-loaded-intent (same vault), post-refund intent-still-loaded proof (CAP_EXCEEDED on Pre-PegIn resubmission), and a foreign-prevout refund that only signs under the IDLE / HASH_DERIVED branch. The Schnorr yield is verified against the recomputed BIP-341 script-path sighash with `SIGHASH_DEFAULT`.

Approving.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -964,8 +1066,9 @@ export class LedgerVaultProvider implements IBTCProvider {
       });
       this.assertSameConnection(generation);
       // Success never consumes the intent: further (different) PSBTs sign
-      // under the same approval; this one never again.
-      this.signedFingerprints.add(fingerprintKey);
+      // under the same approval; this one never again. No-intent standalone
+      // refunds are exempt — the device holds no masks to consult there.
+      if (guardReplay) this.signedFingerprints.add(fingerprintKey);
```
gbarkhatov: **A declined or failed refund tears down a still-valid loaded intent.**

`guardReplay` is consulted only here, on the success path. The `catch` below at `:1080-1085` ignores it, so any device-side failure — user taps Reject, device locks mid-CONTINUE, transport hiccup — takes the pessimistic `this.deviceState = { phase: "idle" }; this.signedFingerprints = new Set()`. A `SW_DENY` surfaces as `LedgerUserRefusedError`, which is not `isLedgerSignPsbtAbortedError`, so the reset does fire.

For refunds that reset is provably unnecessary. I grepped the refund validator (`sign_psbt_validate.c:797-1104`) and the standalone signing section (`sign_custom_inputs.c:563-700`) at `ff1e1ce17`: zero `vault_context_invalidate` and zero `vault_context_transition` calls between them. `VAULT_STATE_INTENT_LOADED` is terminal and the only exit is an explicit invalidate (`vault_context.c:44-46`).

Concretely: a depositor mid-deposit with an approved intent opens the refund modal for that same vault, rejects on the device — the device stays `INTENT_LOADED`, the host drops to `idle`, and the deposit now demands a full re-derive plus re-approve. That directly contradicts `assertRefundSignable`'s own docstring at `:754-755` ("tearing down a loaded ceremony is never a silent side effect of a refund attempt"), which holds only for the two host-side throws.

Suggest carrying the standalone-ness into the `catch` and skipping the reset for a refund on a device-side rejection — the same justification that already motivates `guardReplay: false`. Keep the reset for the abort branch and for unclassifiable transport failures. There is currently no test covering a failing refund either way.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -112,6 +126,8 @@ interface StagedPsbt {
   readonly prepared: PreparedSignPsbt;
   readonly fingerprintKey: string;
   readonly label: string;
+  /** Whether success arms the replay guard — false only for a no-intent standalone refund. */
```
gbarkhatov: **This comment describes a narrower exemption than the code implements.**

`:716` is `const guardReplay = refund === undefined` — every refund is exempt, intent loaded or not. The test "a refund can be re-signed under the loaded intent too" pins the broader behaviour deliberately, so the code is right and the comment is wrong. Same wording at `:1069-1070` ("No-intent standalone refunds are exempt — the device holds no masks to consult there").

The justification is state-independent, which makes the fix easy: the standalone section consumes no dedup mask and no cap in *any* vault state (`sign_custom_inputs.c:563-700`; counters exist only at `:184`, `:307`, `:401` and `sign_psbt_validate.c:736`). Suggest "false for any standalone refund — the device's standalone path consumes no dedup mask or cap in any vault state."

On a file where reviewers are required to walk every line unaided, a comment that understates the scope of a replay-guard waiver is the wrong kind of wrong.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -672,22 +693,91 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+        let stagingHex = psbtHex;
+        // A refund routes to the device's standalone sign path in EVERY vault
+        // state (`sign_psbt_validate.c:3552-3618` fall-through dispatch), and
+        // that path consumes no dedup mask or cap (`sign_custom_inputs.c`,
+        // standalone section — contrast PegIn `:183` and Payout `:400`), so
+        // re-signing one is always a fresh user-approved ceremony.
+        const guardReplay = refund === undefined;
+        if (refund !== undefined) {
+          this.assertRefundSignable(refund, ctx.depositorXOnlyHex);
+          const { masterFingerprintHex } = await this.getPolicyContext();
```
gbarkhatov, line 728: **Narrow point on wording, not on the gate design** — the thread on `:709` already settled that this starts no ceremony, and I agree.

But `getPolicyContext()` is `getMasterFingerprintHex` + `getExtendedPublicKey`, cached per connection, so on the first refund of a session it does exchange APDUs. It runs before `augmentPsbtForRefund` can reject. That makes three statements inaccurate for the CSV-floor, non-P2TR and not-refund-shaped rejections: the error string `"signPsbt rejected before device I/O"` at `:729`, the framing in `refundPsbt.ts`, and the test name "...below the device floor of 72, before any device I/O" — which asserts only that `prepareSignPsbt`/`signPreparedVaultPsbt` were not called, and would still pass with the fingerprint read having happened.

Cheapest fix that makes the claim true rather than weakening it: move the CSV-floor check into `assertRefundSignable`, which is genuinely zero-I/O. `classifyRefundPsbt` already returns `csv`, so it needs no extra parse. Otherwise reword to "before any signing ceremony" and rename the test.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -709,6 +799,16 @@ export class LedgerVaultProvider implements IBTCProvider {
         const stagedKeys = new Set<string>();
         const staged: StagedPsbt[] = [];
         for (const [index, hex] of psbtsHexes.entries()) {
+          // A refund is a standalone single-PSBT ceremony — signPsbt classifies
+          // and augments it. Staged unaugmented here it would start a ceremony
+          // the device rejects (missing derivation entries). Refuse at zero I/O.
+          if (classifyRefundPsbt(hex) !== undefined) {
```
gbarkhatov, line 820: **This refusal sits behind the intent gate, so the caller gets the wrong error.**

`gateSignContext()` at `:798` runs with `requireIntent = true` before this loop. So `signPsbts([refundHex])` with no intent loaded returns `DEVICE_CEREMONY_INVALID` — "holds no approved intent ... restart the flow from derivation" — instead of this block's actionable "a refund signs standalone via signPsbt". That sends the caller off to run a derive-and-approve ceremony they do not need, which is exactly backwards for the mistake this guard exists to catch. Provider messages surface verbatim to the user (`useRefundState.ts:186-192`), so the wording matters.

The scan is pure and zero-I/O, so it can move: hoist the whole-array scan above `gateSignContext()`, right after the empty-array check. The existing test only covers the `approved()` path, so it would not catch this.

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+  const psbt = Psbt.fromHex(psbtHex);
+  const outScript = Buffer.from(psbt.txOutputs[0].script);
+  if (
+    outScript.length !== P2TR_SCRIPT_BYTES ||
```
gbarkhatov: **The output entry asserts ownership of bytes this function never checked.**

The only output check is structural — 34 bytes, `0x51`, `0x20` — which accepts *any* P2TR. `:217` then writes `pubkey: outScript.subarray(2)` with the device's fingerprint and the depositor path, i.e. tells the device "output 0 is mine, derived at `m/86'/…/0/0`". Input 0 gets a real equality check at `:176`; output 0 gets none. That asymmetry is undocumented, and it is the one value destination this file makes a positive ownership claim about.

The helper is already in the package and this is exactly how the sibling augmenter handles it — `bip86OutputScript` at `expectedSignatures.ts:150`, used at `policyPsbt.ts:89` to derive and compare rather than copy:

```ts
if (!outScript.equals(bip86OutputScript(depositorXOnlyHex))) {
  throw new Error("refund output does not pay the depositor's BIP-86 address");
}
```

That subsumes the shape check and lets `P2TR_SCRIPT_BYTES`, `P2TR_OP_1`, `P2TR_PUSH_32` and `P2TR_WITNESS_PROGRAM_OFFSET` all go away — which also clears a smaller nit: `P2TR_OP_1 = 0x51` (`:55`) duplicates `OP_1` (`:42`) and `P2TR_PUSH_32 = 0x20` (`:56`) duplicates `OP_PUSHBYTES_32` (`:35`), two names for one byte in one file.

Not exploitable end to end — the SDK pins the payee and value at `refund.ts:243-266`, and the device re-derives — so this is defence in depth at a trust boundary. But the whole design of this seam is host-side typed rejection in place of opaque device rejects, and this is the one gate that skips it. Worth noting the existing negative test uses a p2wpkh output, so it passes on length; a *foreign P2TR* output is untested. No relation between `witnessUtxo.value` and the output value is asserted either.

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -0,0 +1,224 @@
...
+ * caller's ordinary gates keep their error precedence.
+ */
+export function classifyRefundPsbt(psbtHex: string): RefundPsbtClassification | undefined {
+  let psbt: Psbt;
+  try {
+    psbt = Psbt.fromHex(psbtHex);
+  } catch {
+    return undefined;
+  }
+  if (psbt.data.inputs.length !== 1 || psbt.data.outputs.length !== 1) return undefined;
```
gbarkhatov, line 174: **The classifier is looser than the device's own refund preconditions.**

`_validate_display_refund` additionally requires `tx_version >= 2 && locktime == 0` (`sign_psbt_validate.c:806-809`), and `tx_version == 0` is dispatched to PoP at `:3552` — before refund dispatch is ever reached. So a version-0 or non-zero-locktime PSBT carrying a refund-grammar leaf gets the intent gate and the host replay guard waived and augmentation applied, for a PSBT the device does not treat as a refund at all.

Unreachable through `buildRefundPsbt` today, since it copies WASM's version and locktime — so this is hardening rather than a live bug. But mirroring the device exactly is this module's stated contract, and this condition is the one that decides whether two security gates are waived. Adding `psbt.version >= 2 && psbt.locktime === 0` here is two clauses.

While you are here: the grammar, the input entry and the output entry all carry firmware citations, but this dispatch condition — the one that actually triggers the waivers — has none. Worth citing it the way the rest of the file does.

packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
```diff
@@ -0,0 +1,321 @@
...
+
+/** Minimal CScriptNum push of 2016: `02 e0 07`. */
+const PUSH_2016 = Buffer.from([0x02, 0xe0, 0x07]);
+const CSV_2016 = 2016;
+
+const HTLC_VALUE_SATS = 1_000_000;
+const REFUND_VALUE_SATS = 990_000;
+const PREV_HASH = Buffer.alloc(32, 0x11);
+
+function refundLeaf(keyHex: string, csvPush: Buffer): Buffer {
```
gbarkhatov, line 51: **Nothing in the repo binds this grammar to a real WASM-built refund leaf.**

Every accept-vector in this suite, and `htlcLeaf1` in the e2e fixture, is built by a hand-rolled leaf constructor. The whole feature rests on the WASM-emitted leaf matching that shape — and if it ever stops matching, `classifyRefundPsbt` returns `undefined`, `gateSignContext(true)` runs, and refund fails with `DEVICE_CEREMONY_INVALID` for every Ledger depositor, with no failing test anywhere in the monorepo.

That drift is not hypothetical. `packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/refund.ts:60` — the file that builds this very PSBT — documents the refund script as `` `<timelockRefund> CSV DROP <depositorPubkey> CHECKSIG` ``, which is a different shape from the `<key32> CHECKSIGVERIFY <csv> CSV` this parser requires. Someone has already written the grammar down wrong next to the builder; only an untested coincidence keeps host and builder in sync. Worth fixing that comment either way.

There is a direct precedent for closing this: `services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts` exists to tie a real SDK-built Payout PSBT to the signer's expectation table. Cheapest equivalent here is an in-package test asserting `parseRefundLeafScript(getPrePeginHtlcConnectorInfo(...).refundScript)` succeeds — `@babylonlabs-io/babylon-tbv-rust-wasm` and the SDK are already devDependencies of this package, and the e2e already lazily imports the SDK.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts
```diff
@@ -1378,6 +1378,164 @@ describe("LedgerVaultProvider", () => {
...
+    });
+
+    it("refuses a refund for a DIFFERENT vault while an intent is loaded — a typed error instead of the device's opaque reject", async () => {
+      const provider = await approved();
+
+      // Wrong timelock and wrong prevout each pin to the loaded intent on-device.
+      await expect(provider.signPsbt(refundPsbtHex({ csv: 144 }))).rejects.toMatchObject({
+        code: ERROR_CODES.DEVICE_CEREMONY_INVALID,
+      });
+      await expect(provider.signPsbt(refundPsbtHex({ prevHashInternal: Buffer.alloc(32, 0x77) }))).rejects.toMatchObject({
```
gbarkhatov: **Two distinct guard branches in one test.**

The wrong-CSV case at `:1479` and the wrong-prevout case here are the two sides of the `||` at `provider.ts:766-767`, and the mock-call assertions run once at the end. Delete the txid comparison from the guard and the first `await` still passes and both counts still hold — the test goes green on a half-removed guard.

Split into two `it`s, or `it.each` over the two overrides. CLAUDE.md: "Each test verifies ONE specific behavior of production code."

Same shape one test down at `:1527`: the batch-refusal test asserts only `code: INVALID_PARAMS`, which `stagePsbt` also produces on other paths. Asserting the message would pin that the refusal fired at index 1 for the intended reason.

gbarkhatov (dismissed):
Approving. I checked out the cited pin `ff1e1ce17` and line-walked the new code against the firmware, then ran both suites locally. The design is sound and the central security property holds — augmentation cannot redirect funds, because the device re-derives both entries and compares (`sign_psbt_validate.c:905-950` untweaked input key; `:1005-1057` derive → `crypto_tr_tweak_pubkey` → compare). A lie in either entry is rejected on-device.

## Verified against `ff1e1ce17`

| Claim | Evidence | Verdict |
|---|---|---|
| `parseRefundLeafScript` mirrors the device grammar | `sign_psbt_validate_helpers.c:77-148` | Line-walked, equivalent. The TS collapses the firmware's separate `OP_0`/`OP_1NEGATE` branches into one — behaviour-identical, `0x4f` is in no other range. JS `\|=` vs C `uint32_t` is equivalent because the sign-bit rejection caps `csv` at `0x7fffffff`. |
| CSV floor 72, floor-only in IDLE/HASH_DERIVED | `vault_constants.h:103`, `sign_psbt_validate.c:888-903` | Correct — and there is deliberately **no** upper bound on that branch, so the host is right not to add one. |
| INTENT_LOADED pins CSV and input-0 prevout | `sign_psbt_validate.c:893-897`, `:1076-1081` | Correct. Only the **txid** is pinned, not the vout — `assertRefundSignable` matches exactly. |
| Replay-guard exemption is safe in every state | `sign_custom_inputs.c:563-700` ticks no counter; counters exist only at `:184` pegin, `:307` nopayout, `:401` payout, `sign_psbt_validate.c:736` pre-pegin | **Confirmed.** |
| Leaf dispatch is disjoint | `sign_psbt_validate.c:3644-3695`, `leaf_has_assert_shape` | Confirmed. Claim needs len 34 (below the host's 36 min), WC needs `leaf[34]==OP_SIZE`, Assert needs len > 68 + terminal `OP_TRUE`, Vault-UTXO is far longer than exact consumption admits. No vault PSBT can be misclassified as a refund. |
| Amount and fee are user-visible | `display.c:122-150` — txid, reclaimed amount, timelock, fee, reclaim address | Confirmed. |
| No SDK change needed | `buildAndBroadcastRefund.ts:531-565` | Confirmed. `assertPsbtUnsignedTxMatches` compares the unsigned tx only (untouched by augmentation), and `finalizeScriptPathWithSignatures` finalizes the **original** PSBT, so the injected fields never reach the broadcast tx. |

Local runs on Node 24: signer 635 passed / 23 skipped, provider suite 136 passed, eslint clean, `tsc -p tsconfig.json --noEmit` clean. Worth flagging for anyone reproducing: the connector suite fails 72 tests with `No "classifyRefundPsbt" export` until `babylon-ledger-vault-signer` is rebuilt — the stale-`dist/` trap CLAUDE.md warns about.

## Two asks outside the diff

**Provenance.** There is no Speculos job anywhere in this repo's `.github/`, so stage (13) never runs in CI and the device-behaviour claims rest on a manual run. Per the critical-path rules, please state in the PR description which firmware build stage (13) was executed against and that all four tests passed.

**Registry gap.** The operative gates — `assertRefundSignable`, the wrong-vault guard, the replay-guard exemption — live in `packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts`, which is not in the CLAUDE.md section 7 path list or `critical-path-check.yml`. This PR is flagged only because of the signer-package files. Worth registering that directory separately.

## Considered and dismissed (so they don't get re-raised)

- **No upper bound on the refund CSV.** The device has none either on that branch (`:898-903` checks only `< VAULT_TIMELOCK_MIN`). Adding one would make the host stricter than the device, which this module explicitly forbids.
- **`vault_constants.h:103` is the wrong citation for the floor.** It is correct at `ff1e1ce17`, where `VAULT_TIMELOCK_MIN` sits on line 103. (`deviceCaps.ts` cites `:100` because it is pinned to `8f99b8b`.)
- **A refund from `phase: "derived"` might consume HASH_DERIVED and leave the mirror stale.** It does not — the refund validator and the standalone signer contain zero `vault_context_transition` / `vault_context_invalidate` calls, so the mirror staying at `derived` is right.
- **Reclaim needs the same carve-out.** Already handled deliberately at `reclaimEligibility.ts:241`, whose comment reasons about exactly this distinction.
- **`updateInput` appends rather than replaces, and `leafHashes: []` is BIP-371-incorrect.** The device looks entries up by exact key, so a stray extra entry is inert, and the `[]` tradeoff is already documented in the code.
- **Fee-burn on a malformed refund.** `buildRefundPsbt` pins the output scriptPubKey to the depositor's BIP-86 address and the value to `htlcValue - refundFee`, and the device displays both.
- **The batched-refund test passes for the wrong reason.** It does not — `prepareSignPsbt` is mocked to succeed for any hex, so element 0 stages and the refusal genuinely fires at index 1.

Nothing below blocks merge. The mirror-reset one is the only behavioural item.

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -0,0 +1,224 @@
...
+  const leaves = psbt.data.inputs[0].tapLeafScript ?? [];
+  if (leaves.length !== 1) return undefined;
+  if (leaves[0].leafVersion !== TAPSCRIPT_LEAF_VERSION) return undefined;
+  const terms = parseRefundLeafScript(leaves[0].script);
+  if (terms === undefined) return undefined;
+  return { ...terms, inputTxidInternalHex: Buffer.from(psbt.txInputs[0].hash).toString("hex") };
```
gbarkhatov: **`PSBT_IN_SEQUENCE` is never validated — the device's likeliest rejection on this path.**

The classification returns the leaf key, the CSV and input 0's prevout, but drops `psbt.txInputs[0].sequence`, and `augmentPsbtForRefund` never reads it either. The firmware does, at `sign_psbt_validate.c:955-978` (@ `ff1e1ce17`), with three conditions:

- present and exactly 4 bytes, else `SW_INCORRECT_DATA`
- `BIP68_DISABLE_FLAG` (bit 31) and `BIP68_TIME_BASED_FLAG` (bit 22) both clear
- canonical equality: `(nsequence & BIP68_SEQUENCE_MASK) == (csv_value & BIP68_SEQUENCE_MASK)` — the comment there is explicit that "sequence must encode exactly the CSV timelock, not just satisfy it"

So this file gates the CSV floor, which the device applies only on the IDLE/HASH_DERIVED branch, and omits the check it applies unconditionally. Nothing upstream covers it either: the SDK copies the value straight out of the WASM-built transaction at `primitives/psbt/refund.ts:227` (`sequence: refundInput.sequence`) with no assertion.

The failure mode is worse than an opaque reject. A refund with a correct leaf but a time-flagged or short `nSequence` gets a valid signature and then fails at broadcast as `non-BIP68-final`, which `useRefundState.ts:186-189` surfaces to the depositor as "the timelock has not expired yet" — a permanently unbroadcastable refund presented as "wait longer".

Suggest returning `sequence` from `classifyRefundPsbt` (it already has the input in hand) and asserting the three conditions in `augmentPsbtForRefund`, with the masks as named constants citing `vault_constants.h`.

Worth noting the e2e cannot catch this today: `peginFixture.ts:523` builds the input with `sequence: HTLC_REFUND_TIMELOCK`, equal to the CSV by construction.

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -0,0 +1,224 @@
...
+  // The device never signs a refund below this floor in any state
+  // (`fw:sign_psbt_validate.c:898-903`, IDLE/HASH_DERIVED branch;
+  // `vault_constants.h:103`) — pre-empt its opaque SW_INCORRECT_DATA.
+  if (classified.csv < DEVICE_TIMELOCK_MIN_BLOCKS) {
+    throw new Error(
+      `refund leaf CSV ${classified.csv} is below the device's timelock floor of ${DEVICE_TIMELOCK_MIN_BLOCKS} blocks`,
+    );
+  }
+  const psbt = Psbt.fromHex(psbtHex);
+  const outScript = Buffer.from(psbt.txOutputs[0].script);
```
gbarkhatov, line 272: **Nothing bounds the refund's output value against its input.**

`witnessUtxo.value` and `txOutputs[0].value` are both in the PSBT at this point, so `fee = input - output` is free to compute, and neither is checked. The firmware checks both: `sign_psbt_validate.c:1059-1062` rejects `out_value > htlc_value` before computing the fee, and `:836-854` requires `PSBT_IN_WITNESS_UTXO` present with `spk_len == 34` — this function never verifies the input carries a `witnessUtxo` at all.

The SDK does cap the fee (`REFUND_MAX_FEE_RATE_SATS_VB = 2000` plus the fraction cap at `buildAndBroadcastRefund.ts:425-438`), so this is not a live hole through the app path. But `signPsbt` is a public `IBTCProvider` seam, and CLAUDE.md section 7 puts the envelope gate here precisely so the host does not depend on a caller's discipline — the same reasoning that motivates the CSV floor check a few lines below.

This is also the one flow that waives the intent requirement, so on a no-intent sign the on-device display is the remaining control — and stage (13) collects the review screens at `speculos.e2e.test.ts:802` and `console.log`s them without asserting the amount or the fee. Bounding the implied fee here, and asserting the expected amount appears in `refundScreens`, would close both halves.

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -0,0 +1,224 @@
...
+    csv = pushOp - OP_RESERVED;
+  } else {
+    return undefined;
+  }
+  // The sign-bit rejections above cap csv at 0x7fffffff — no uint32 wrap.
+  if (csv === 0) return undefined;
+
+  if (pos >= script.length) return undefined;
+  if (script[pos++] !== OP_CHECKSEQUENCEVERIFY) return undefined;
+  if (pos !== script.length) return undefined; // exact consumption
```
gbarkhatov, line 128: **This makes the parser stricter than the firmware, which the docstring above says it must never be.**

`:67-71` describes this function as a "byte-for-byte mirror ... deliberately EXACTLY as loose", on the stated grounds that "a host grammar stricter than the device's would strand a refund the device would sign". Exact consumption is a place where it is tighter: the firmware's dispatch fallback at `sign_psbt_validate.c:3167-3168` routes a leaf to the refund validator on the last byte being `OP_CSV`, with no equivalent requirement that the script be fully consumed.

The divergence direction is fail-safe for the guards — a host under-classification just falls back to intent-required, so it cannot waive the intent gate or the replay guard wrongly. This is not a defect to fix by loosening the parser. But it does mean the docstring's claim is not accurate as written, and under section 7 these comments are the audit trail.

Suggest keeping the check and amending the docstring to describe the host as a strict superset of the firmware's grammar, naming this line as the deliberate divergence and why it is the safe direction.

gbarkhatov (commented):
Three additions to the earlier review (`pullrequestreview-5065054744`) — all in `refundPsbt.ts`, none on lines already commented.

The first is the one I would want resolved before merge: the host gates the CSV floor, which the device applies only on the IDLE/HASH_DERIVED branch, and omits `PSBT_IN_SEQUENCE`, which it applies unconditionally (`sign_psbt_validate.c:955-978` @ `ff1e1ce17`). A wrong `nSequence` does not produce an opaque device reject — it produces a valid signature on an unbroadcastable transaction, surfaced to the depositor as "wait longer".

All three verified against the firmware checkout at `ff1e1ce17`, the pin this file already cites. Filed as a comment rather than a second approval; the existing approval stands.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/__tests__/provider.test.ts

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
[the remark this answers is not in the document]
jrwbabylonlab: fixed

gbarkhatov (approved):

jonybur (approved):
Approving. The refund PSBT classifier is a strict subset of the firmware dispatcher and the leaf-grammar mirror matches sign_psbt_validate_helpers.c:77-148 byte-for-byte, so under-classification falls back to the intent gate rather than the other way around. assertRefundPsbtSignable pins every device-checked term before any I/O: witnessUtxo present with a 34-byte P2TR scriptPubKey, output value not above the HTLC value, output pays the leaf key's own BIP-86 address, CSV at or above the device timelock floor, and the BIP-68 sequence with flags clear and the low bits matching the CSV. augmentPsbtForRefund writes the two TAP_BIP32_DERIVATION entries the device requires with the asymmetric keying the firmware compares against (input 0 by the untweaked depositor key, output 0 by the tweaked witness program). Provider-side, the intent-loaded gate rejects a refund whose prevout txid or CSV does not match the loaded intent, refund batches are refused, and the replay guard is waived only on the standalone refund path where the device consumes no dedup mask. CI green at head, and prior review threads are all resolved.

PR #2381, opened 2026-09-01
Use new API methods

kirugan:
Corresponding changes in btc-vault repo - https://github.com/babylonlabs-io/btc-vault/pull/2648

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/api.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/api.ts
@@ -19,11 +19,11 @@ import {
   type JsonRpcClientConfig,
 } from "./json-rpc-client";
 import type {
-  BatchGetPeginStatusParams,
+  BatchGetPeginStatusByVaultIdParams,
   BatchGetPeginStatusResponse,
-  BatchGetPegoutStatusParams,
+  BatchGetPegoutStatusByVaultIdParams,
   BatchGetPegoutStatusResponse,
-  GetPeginStatusParams,
+  GetPeginStatusByVaultIdParams,
   GetPeginStatusResponse,
   RequestDepositorClaimerArtifactsParams,
   RequestDepositorClaimerArtifactsResponse,
@@ -49,8 +49,8 @@ export interface VaultProviderRpcClientOptions {
   retryDelay?: number;
   /**
    * Custom retry predicate. Default retries only the idempotent read
-   * methods: `getPeginStatus`, `batchGetPeginStatus`, `batchGetPegoutStatus`,
-   * `requestDepositorPresignTransactions`.
+   * methods: `getPeginStatusByVaultId`, `batchGetPeginStatusByVaultId`,
+   * `batchGetPegoutStatusByVaultId`, `requestDepositorPresignTransactions`.
    */
   retryableFor?: (method: string) => boolean;
   /** Custom headers. */
@@ -73,7 +73,7 @@ const DEFAULT_TIMEOUT_MS = 60_000;
  * Usage:
  * ```ts
  * const client = new VaultProviderRpcClient("https://vp.example.com/rpc");
- * const status = await client.getPeginStatus({ pegin_txid: "abc..." });
+ * const status = await client.getPeginStatusByVaultId({ vault_id: "0xabc..." });
  * ```
  */
 export class VaultProviderRpcClient
@@ -158,49 +158,51 @@ export class VaultProviderRpcClient
     return response;
   }
 
-  /** Get the current pegin status from the vault provider daemon. */
-  async getPeginStatus(
-    params: GetPeginStatusParams,
+  /**
+   * Get the current pegin status from the vault provider daemon,
+   * addressed by the depositor-bound `vault_id`.
+   */
+  async getPeginStatusByVaultId(
+    params: GetPeginStatusByVaultIdParams,
     signal?: AbortSignal,
   ): Promise<GetPeginStatusResponse> {
-    const response = await this.client.call<GetPeginStatusParams, unknown>(
-      "vaultProvider_getPeginStatus",
-      params,
-      signal,
-    );
+    const response = await this.client.call<
+      GetPeginStatusByVaultIdParams,
+      unknown
+    >("vaultProvider_getPeginStatusByVaultId", params, signal);
     validateGetPeginStatusResponse(response);
     return response;
   }
 
   /**
-   * Get pegin status for many txids in one round trip. Per-result envelope
-   * isolates per-pegin failures from the overall RPC. Caller must chunk
-   * inputs at `VP_BATCH_MAX_SIZE`.
+   * Get pegin status for many vault ids in one round trip. Per-result
+   * envelope isolates per-vault failures from the overall RPC. Caller must
+   * chunk inputs at `VP_BATCH_MAX_SIZE`.
    */
-  async batchGetPeginStatus(
-    params: BatchGetPeginStatusParams,
+  async batchGetPeginStatusByVaultId(
+    params: BatchGetPeginStatusByVaultIdParams,
     signal?: AbortSignal,
   ): Promise<BatchGetPeginStatusResponse> {
     const response = await this.client.call<
-      BatchGetPeginStatusParams,
+      BatchGetPeginStatusByVaultIdParams,
       unknown
-    >("vaultProvider_batchGetPeginStatus", params, signal);
+    >("vaultProvider_batchGetPeginStatusByVaultId", params, signal);
     validateBatchGetPeginStatusResponse(response);
     return response;
   }
 
   /**
-   * Get pegout status for many txids in one round trip. Same per-result
-   * envelope semantics as `batchGetPeginStatus`.
+   * Get pegout status for many vault ids in one round trip. Same per-result
+   * envelope semantics as `batchGetPeginStatusByVaultId`.
    */
-  async batchGetPegoutStatus(
-    params: BatchGetPegoutStatusParams,
+  async batchGetPegoutStatusByVaultId(
+    params: BatchGetPegoutStatusByVaultIdParams,
     signal?: AbortSignal,
   ): Promise<BatchGetPegoutStatusResponse> {
     const response = await this.client.call<
-      BatchGetPegoutStatusParams,
+      BatchGetPegoutStatusByVaultIdParams,
       unknown
-    >("vaultProvider_batchGetPegoutStatus", params, signal);
+    >("vaultProvider_batchGetPegoutStatusByVaultId", params, signal);
     validateBatchGetPegoutStatusResponse(response);
     return response;
   }
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchAttribution.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchAttribution.ts
@@ -1,56 +1,62 @@
 /**
  * Defensive helper for attributing per-item results in a VP batch RPC
- * response back to the requested txids. The server promises 1:1 ordered
+ * response back to the requested vault ids. The server promises 1:1 ordered
  * results, but we don't trust that promise — a server bug could duplicate,
  * skip, or scramble items, and silent attribution-by-array-index would
  * mask the bug.
  *
- * Lowercases all txids on both sides to absorb case mismatch (the FE
- * strips `0x` but doesn't otherwise normalize).
+ * Normalizes vault ids on both sides (strip `0x`, lowercase) to absorb
+ * prefix and case mismatch — the server echoes the request string verbatim.
  *
  * @module tbv/core/clients/vault-provider/batchAttribution
  */
 
 /** Per-item entry in a VP batch response. */
 export interface BatchResultEntry<T> {
-  pegin_txid: string;
+  vault_id: string;
   result: T | null;
   error: string | null;
 }
 
 /** Output of {@link attributeBatchResults}. */
 export interface BatchAttributionResult<T> {
-  /** Lowercase requested txid -> per-item envelope. */
-  byTxid: Map<string, { result: T | null; error: string | null }>;
-  /** Requested txids that did not appear in the response. */
+  /** Normalized requested vault id -> per-item envelope. */
+  byVaultId: Map<string, { result: T | null; error: string | null }>;
+  /** Requested vault ids that did not appear in the response. */
   missing: string[];
-  /** Echoed txids that were not in the request — logged + dropped. */
+  /** Echoed vault ids that were not in the request — logged + dropped. */
   unexpected: string[];
-  /** Echoed txids that appeared more than once — first kept, rest dropped. */
+  /** Echoed vault ids that appeared more than once — first kept, rest dropped. */
   duplicate: string[];
 }
 
+/** Normalize a vault id for map keys: strip an optional `0x`, lowercase. */
+export function normalizeVaultId(vaultId: string): string {
+  const unprefixed = vaultId.startsWith("0x") ? vaultId.slice(2) : vaultId;
+  return unprefixed.toLowerCase();
+}
+
 /**
- * Attribute batch results to requested txids defensively.
+ * Attribute batch results to requested vault ids defensively.
  *
- * Both `requestedTxids` and the echoed `pegin_txid` field on each result
- * are lowercased before lookup. Duplicates and unexpected echoes are
+ * Both `requestedVaultIds` and the echoed `vault_id` field on each result
+ * are normalized before lookup. Duplicates and unexpected echoes are
  * surfaced so callers can flag the affected items as errored rather than
  * silently overwriting state.
  *
- * `requestedTxids` may contain duplicates; they are de-duplicated for the
- * purposes of map keys (each unique txid becomes a single map entry).
+ * `requestedVaultIds` may contain duplicates; they are de-duplicated for the
+ * purposes of map keys (each unique vault id becomes a single map entry).
  */
 export function attributeBatchResults<T>(
-  requestedTxids: string[],
+  requestedVaultIds: string[],
   results: ReadonlyArray<BatchResultEntry<T>>,
 ): BatchAttributionResult<T> {
   const requestedSet = new Set<string>();
-  for (const txid of requestedTxids) {
-    requestedSet.add(txid.toLowerCase());
+  for (const vaultId of requestedVaultIds) {
+    requestedSet.add(normalizeVaultId(vaultId));
   }
 
-  const byTxid = new Map<
+  const byVaultId = new Map<
     string,
     { result: T | null; error: string | null }
   >();
@@ -59,23 +65,23 @@ export function attributeBatchResults<T>(
   const unexpected: string[] = [];
 
   for (const entry of results) {
-    const lower = entry.pegin_txid.toLowerCase();
-    if (!requestedSet.has(lower)) {
-      unexpected.push(lower);
+    const normalized = normalizeVaultId(entry.vault_id);
+    if (!requestedSet.has(normalized)) {
+      unexpected.push(normalized);
       continue;
     }
-    if (seen.has(lower)) {
-      duplicate.push(lower);
+    if (seen.has(normalized)) {
+      duplicate.push(normalized);
       continue;
     }
-    seen.add(lower);
-    byTxid.set(lower, { result: entry.result, error: entry.error });
+    seen.add(normalized);
+    byVaultId.set(normalized, { result: entry.result, error: entry.error });
   }
 
   const missing: string[] = [];
-  for (const txid of requestedSet) {
-    if (!seen.has(txid)) missing.push(txid);
+  for (const vaultId of requestedSet) {
+    if (!seen.has(vaultId)) missing.push(vaultId);
   }
 
-  return { byTxid, missing, unexpected, duplicate };
+  return { byVaultId, missing, unexpected, duplicate };
 }
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchPoll.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchPoll.ts
@@ -17,6 +17,7 @@ import {
   type BatchResultEntry,
 } from "./batchAttribution";
 import { VP_BATCH_MAX_SIZE } from "./types";
+import { isVaultIdHex } from "./validators";
 
 export interface BatchPollByProviderOptions<TItem, TResult> {
   /** Items to poll for this provider, e.g. `DepositToPoll[]`. */
@@ -38,6 +39,11 @@ export interface BatchPollByProviderOptions<TItem, TResult> {
    *
    * Note: `envelope.vault_id` is the normalized vault id the helper
    * sent in the request, not whatever case/encoding the server echoed.
+   *
+   * Also dispatched with a locally-produced `error` for an item whose
+   * `getVaultId` is not a well-formed vault id. Such an item is never
+   * sent, because an unattributable id comes back as `missing` and would
+   * blame the provider for a caller-side defect.
    */
   onItem: (item: TItem, envelope: BatchResultEntry<TResult>) => void;
   /** Server omitted this item from the response. */
@@ -98,11 +104,25 @@ export async function batchPollByProvider<TItem, TResult>(
     const chunk = items.slice(i, i + batchSize);
     const vaultIdToItem = new Map<string, TItem>();
     const vaultIds: string[] = [];
+    // Only the items actually sent — an item rejected below already got its
+    // own `onItem` error and must not be re-reported by `onWholeBatchError`.
+    const polled: TItem[] = [];
     for (const item of chunk) {
-      const normalized = normalizeVaultId(getVaultId(item));
+      const rawVaultId = getVaultId(item);
+      if (!isVaultIdHex(rawVaultId)) {
+        onItem(item, {
+          vault_id: String(rawVaultId),
+          result: null,
+          error: `Invalid vault id "${String(rawVaultId)}" — not a 64-char hex string`,
+        });
+        continue;
+      }
+      const normalized = normalizeVaultId(rawVaultId);
       vaultIdToItem.set(normalized, item);
       vaultIds.push(normalized);
+      polled.push(item);
     }
+    if (vaultIds.length === 0) continue;
 
     // Both the RPC call and attribution sit inside the same try/catch
     // so a malformed-batch validator throw is routed through
@@ -112,7 +132,7 @@ export async function batchPollByProvider<TItem, TResult>(
       const response = await batchCall(vaultIds);
       attribution = attributeBatchResults<TResult>(vaultIds, response.results);
     } catch (error) {
-      onWholeBatchError(chunk, error);
+      onWholeBatchError(polled, error);
       continue;
     }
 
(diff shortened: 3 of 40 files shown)


packages/babylon-ts-sdk/src/tbv/core/services/deposit/waitForPeginStatus.ts
```diff
@@ -56,26 +57,26 @@ export async function waitForPeginStatus(
...
-        { pegin_txid: peginTxid },
+      const response = await statusReader.getPeginStatusByVaultId(
+        { vault_id: vaultId },
         signal,
       );
 
-      // Reject responses echoing a different pegin txid.
-      if (response.pegin_txid.toLowerCase() !== peginTxid.toLowerCase()) {
+      // Reject responses echoing a different vault id.
+      if (normalizeVaultId(response.vault_id) !== normalizeVaultId(vaultId)) {
```
gbarkhatov, line 86: This swaps an attested binding for a self-referential one, so the guard can no longer fail for the reason the old one could.

The server builds this field as `vault_id: vault_id.to_string()` on the id it parsed out of *our own request* — it round-trips what we sent and attests nothing about the row that answered. The field it replaced, `pegin_txid`, is a DB lookup (`get_pegin_txid_by_vault_id`), and is the only server-attested identifier in the payload.

That matters because `peginTxid` was dropped from `WaitForPeginStatusParams` entirely, yet `runDepositorPresignFlow` still receives both identifiers and still addresses the irreversible writes by txid — `requestDepositorPresignTransactions` / `submitDepositorPresignatures`, and `submitDepositorWotsKey`. The readiness gate is now keyed on one identifier and the signature on another, with nothing asserting they name the same vault.

Suggestion: keep `peginTxid` on the params and assert `response.pegin_txid.toLowerCase() === peginTxid.toLowerCase()` alongside this check. Keeping both is strictly better than either alone — the vault-id echo catches a mangled response, the txid catches a mispaired one.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchAttribution.ts
```diff
@@ -59,23 +65,23 @@ export function attributeBatchResults<T>(
...
-    if (seen.has(lower)) {
-      duplicate.push(lower);
+    if (seen.has(normalized)) {
+      duplicate.push(normalized);
       continue;
     }
-    seen.add(lower);
-    byTxid.set(lower, { result: entry.result, error: entry.error });
+    seen.add(normalized);
+    byVaultId.set(normalized, { result: entry.result, error: entry.error });
```
gbarkhatov, line 78: `entry.result` is stored without ever being corroborated against the key it arrived under.

Attribution keys entirely on the envelope's `vault_id`, which the server copies verbatim out of the request (`vault_id: Some(vault_id_hex)`). So a VP that pairs envelope key A with vault B's payload passes here unchallenged, and `usePeginPollingQuery` then applies B's status to deposit A — which gates `pendingDepositorSignatures` and the Sign Payouts affordance. The module docstring above claims defence against a server that could "duplicate, skip, or scramble items"; a verbatim echo can detect the first two but structurally cannot detect the third.

One caveat on the obvious fix: comparing the envelope `vault_id` to the nested `result.vault_id` is weaker than it looks, because both are echoes of the same request string and agree by construction. The DB-attested field is `result.pegin_txid`, and callers already hold the expected `peginTxHash` per vault (`peginTxHashByVaultId` in `fetchActivities.ts`). Asserting that is what would actually bite.

Also worth noting this gap is structurally pre-existing — the old envelope/nested `pegin_txid` pair was equally unchecked. What's new is the asymmetry with `waitForPeginStatus`, which does perform this check on the single-status path.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/validators.ts
```diff
@@ -58,6 +58,29 @@ export class VpResponseValidationError extends Error {
...
 
+/** Expected length (in hex chars) of a vault id (keccak256, 32 bytes). */
+const VAULT_ID_HEX_LEN = 64;
+
+/**
+ * Assert a vault id: 64 hex chars, `0x` prefix optional. The server emits
+ * `0x`-prefixed ids in status payloads and echoes request ids verbatim in
+ * batch envelopes, so both encodings are accepted.
+ */
+function assertVaultId(value: unknown, field: string): void {
```
gbarkhatov: Applying this to the batch envelope's `vault_id` turns one bad id into a whole-chunk failure.

When `parse_vault_id_hex` fails server-side, the server still returns a per-item entry echoing the malformed id with `error` populated — the per-result envelope exists precisely so one bad id degrades one item. But `assertVaultId(e.vault_id, …)` in the envelope validator throws for the *entire* response, `batchPoll` catches it and routes to `onWholeBatchError`, and `usePeginPollingQuery` marks all up-to-50 deposits in the chunk errored.

Reachable because `usePegoutPolling` and `claimTxResolver` feed ids straight from subgraph fields typed plain `string` / `string | null`.

Worth reconsidering: the envelope's `vault_id` is our own request string coming back, so a shape assert on it buys nothing that `attributeBatchResults` doesn't already handle — it classifies non-matching echoes as `unexpected`. Making it a per-item failure would preserve the envelope's whole point. The assert on the *nested* `result.vault_id` is fine to keep.

(Flagging as a design question rather than a bug — `validators.test.ts` pins the current whole-envelope throw, so this looks deliberate.)

services/vault/src/hooks/usePegoutPolling.ts
```diff
@@ -105,8 +105,9 @@ async function fetchPegoutStatusesFromProvider(
...
   await batchPollByProvider<VaultToPoll, GetPegoutStatusResponse>({
     items: vaults,
-    getTxid: (entry) => stripHexPrefix(entry.vault.peginTxHash),
-    batchCall: (pegin_txids) => rpcClient.batchGetPegoutStatus({ pegin_txids }),
+    getVaultId: (entry) => entry.vault.id,
+    batchCall: (vault_ids) =>
+      rpcClient.batchGetPegoutStatusByVaultId({ vault_ids }),
     onItem: (entry, envelope) => {
       const vaultId = entry.vault.id;
       if (envelope.error !== null) {
```
gbarkhatov, line 114: This path lost its benign "not found" handling in the migration.

Under the deprecated `batchGetPegoutStatus`, an unknown pegin came back `Ok` with `found: false` — the old resolver returned the txid without a DB lookup. Under `batchGetPegoutStatusByVaultId`, `build_pegout_status_response` falls through to `empty_pegout_status`, which raises `PegInNotFound` when the vault has no txid, and that arrives here as `envelope.error`.

This `onItem` has no `"PegIn not found"` filter — unlike its siblings in `usePeginPollingQuery` and `batchReadiness`, both of which treat it as still-pending. So it logs a warning and calls `applyPegoutFailure`, tripping `TIMED_OUT_STATE` with reason `consecutive_failures` after 10 polls instead of the previous `unknown_status` path at 20.

Suggestion: mirror the pegin path and treat `error.includes("PegIn not found")` as pending rather than incrementing `failureCounts`.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchAttribution.ts
```diff
@@ -1,56 +1,62 @@
...
-  /** Echoed txids that were not in the request — logged + dropped. */
+  /** Echoed vault ids that were not in the request — logged + dropped. */
   unexpected: string[];
-  /** Echoed txids that appeared more than once — first kept, rest dropped. */
+  /** Echoed vault ids that appeared more than once — first kept, rest dropped. */
   duplicate: string[];
 }
 
+/** Normalize a vault id for map keys: strip an optional `0x`, lowercase. */
+export function normalizeVaultId(vaultId: string): string {
```
gbarkhatov, line 34: This normalizes but validates nothing — no length check, no hex check.

Every caller-supplied id runs through here and the result goes straight onto the wire in `batchPollByProvider`, so an empty string, a truncated id, or a stringified `undefined` is transmitted as a `vault_id`. On the way back the malformed id lands in `missing` / `unexpected` and the deposit is quietly left un-updated rather than surfaced — a silent fallback on a value that keys deposit state.

`validators.ts` already has exactly the predicate needed in `assertVaultId` (64 hex chars, `0x` optional), so this is reuse rather than new logic: throw unless the unprefixed remainder matches.

Minor related point: both this and `assertVaultId` strip only a lowercase `0x`, so `0X…` would be rejected. viem never emits that, so it's unreachable today — mentioning it only so the assumption is on record.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/api.ts
```diff
@@ -158,49 +158,51 @@ export class VaultProviderRpcClient
...
   }
 
-  /** Get the current pegin status from the vault provider daemon. */
-  async getPeginStatus(
-    params: GetPeginStatusParams,
+  /**
+   * Get the current pegin status from the vault provider daemon,
+   * addressed by the depositor-bound `vault_id`.
+   */
+  async getPeginStatusByVaultId(
```
gbarkhatov, line 165: The old `getPeginStatus`, `batchGetPeginStatus` and `batchGetPegoutStatus` are deleted outright here, along with `GetPeginStatusParams`, `BatchGetPeginStatusParams` and `BatchGetPegoutStatusParams` in `types.ts`.

babylonlabs-io/btc-vault#2648 deliberately *retains* the txid methods — "the existing methods keep working and keep their exact request and response shapes, so nothing has to change on your timeline" — and schedules their deletion for after mainnet launch. So the SDK is hard-cutting ahead of the server's own migration window, and any external `@babylonlabs-io/ts-sdk` consumer breaks with no deprecation period.

Probably fine if the package is effectively internal — but worth either confirming that explicitly, or keeping thin deprecated wrappers for one release. Your call; flagging because the removal is invisible in the diff unless you're looking for it.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchPoll.ts
```diff
@@ -94,21 +96,21 @@ export async function batchPollByProvider<TItem, TResult>(
...
     const chunk = items.slice(i, i + batchSize);
-    const txidToItem = new Map<string, TItem>();
-    const txids: string[] = [];
+    const vaultIdToItem = new Map<string, TItem>();
+    const vaultIds: string[] = [];
     for (const item of chunk) {
-      const lowerTxid = getTxid(item).toLowerCase();
-      txidToItem.set(lowerTxid, item);
-      txids.push(lowerTxid);
+      const normalized = normalizeVaultId(getVaultId(item));
```
gbarkhatov: Nit, and pre-existing rather than introduced here: this `set` overwrites, so if two items in a chunk share a vault id the shadowed one gets no `onItem`, no `onMissing` and no `onDuplicate`, and keeps its stale state indefinitely — `attributeBatchResults` de-dupes `requestedSet`, so `missing` never flags it.

Meanwhile `vaultIds.push` keeps the duplicate, so the request carries `[V, V]`, the server echoes twice, and the *surviving* item gets labelled "Provider returned duplicate status entry" — blaming the VP for a client-side collision.

Since `vault_id = keccak256(peginTxHash, depositor)` the collision domain for a single depositor is unchanged by this PR, so this is genuinely low priority. Noting it because the txid-keyed version had the same shape and it survived the rewrite.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/__tests__/waitForPeginStatus.test.ts
```diff
@@ -240,45 +242,45 @@ describe("waitForPeginStatus", () => {
     expect(result).toBe(DaemonStatus.ACTIVATED);
   });
 
-  it("rejects post-WOTS status echoed for the wrong pegin txid (audit #310)", async () => {
-    const OTHER_TXID = "b".repeat(64);
+  it("rejects post-WOTS status echoed for the wrong vault id (audit #310)", async () => {
```
gbarkhatov: Nit: these two titles are being rewritten anyway, so it's a free moment to drop the `(audit #310)` suffix (here and on the sibling test below). The convention is to name the behaviour rather than the ticket, and both titles already describe the behaviour fully without it.

gbarkhatov (dismissed):
Reviewed the txid → `vault_id` migration against the merged server contract in babylonlabs-io/btc-vault#2648 (I pulled the Rust patches for `vault_provider.rs`, `pegin_status.rs` and `pegout_status.rs` rather than relying on the PR prose). **The wire contract is correct**: all four `*ByVaultId` method names and the `vault_id` / `vault_ids` param shapes match the handler character for character, `VP_BATCH_MAX_SIZE = 50` matches `MAX_BATCH_SIZE`, and dropping `pegin_txid` from the batch-envelope type is right — the server sets it to `None` with `skip_serializing_if`, so it is absent on this path.

**Merge blocker (CI):** the `verify` job's only failing step is *Check generated API docs are current* — `Build all`, `Lint all`, `Test affected`, `Typecheck the release driver` and `Rehearse the release` all passed. `packages/babylon-ts-sdk/docs/api/clients.md` and `services.md` still document `getPeginStatus` / `batchGetPeginStatus` / `pegin_txids`. Run `pnpm --filter @babylonlabs-io/ts-sdk run docs:clean` and commit.

The substantive theme in the inline comments is that **`vault_id` is an echo and `pegin_txid` is attested**. The server builds the response `vault_id` as `vault_id.to_string()` on the id it parsed out of *our own request*, while `pegin_txid` comes from a DB lookup (`get_pegin_txid_by_vault_id`). Several checks in this PR compare our request against itself, which is why they can't fail. That is the root of comments 1 and 2.

### Refuted while reviewing — noting so they don't get re-raised

- **"The VP proxy allowlist is missing the four new methods."** Not true: `vault-provider-proxy` `main` allowlists all four and classifies them as unauthenticated / not secret-bearing — [vault-provider-proxy#202 *Add new vp methods to allowed list*](https://github.com/babylonlabs-io/vault-provider-proxy/pull/202), commit [`67fdd5f8`](https://github.com/babylonlabs-io/vault-provider-proxy/commit/67fdd5f810fbe1c42c808ea0261df0f16961dd02). A local checkout at `182616d` was stale and produced a false alarm. Worth confirming **deploy ordering** though — that proxy commit landed roughly an hour after this PR's head, so the proxy needs to ship first or alongside.
- **"CLAUDE.md §3 invariants are violated."** Not on substance — `payoutSigning.ts` threads one already-present `vaultId` into status polling; payout re-derivation and the challenger-set equality assertion are untouched. §3's *process* requirement (two reviewers) still applies since the file is registered there.

### Verified clean

- The identifiers passed as vault ids really are `keccak256(abi.encode(peginTxHash, depositor))` at every call site, and this removes the old `peginTxHash!` non-null assertion.
- Keying `batchPollByProvider` on vault id fixes a real latent collision — two deposits sharing a `peginTxHash` previously overwrote each other in the map and one received neither `onItem` nor `onMissing`.
- Retry allowlist in `json-rpc-client.ts` migrated in lockstep; no dead code; no half-migrated call site outside the stale generated docs.
- `assertVaultId` itself is sound: `HEX_RE` is non-global, the non-string branch narrows before slicing, and `VAULT_ID_HEX_LEN` is kept distinct from `TXID_HEX_LEN`.

### Test gaps worth a follow-up

No test pairs an envelope `vault_id: A` with a nested `result.vault_id: B`. `batchPoll.test.ts`'s request-normalization assertion is vacuous — its fixtures are already bare lowercase, so it passes with or without `normalizeVaultId`; `claimTxResolver.test.ts` catches that only incidentally. `payoutReadiness.test.ts` and `wotsSubmission.test.ts` mock `batchPollByProvider` so completely that `getVaultId` and `batchCall` never run, so the migration is type-checked only there.

Approving — everything below is a suggestion for you to weigh, apart from the stale docs which CI already gates.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/waitForPeginStatus.ts
kirugan: Agreed — fixed. `peginTxid` is back on `WaitForPeginStatusParams`, and the loop now asserts `response.pegin_txid.toLowerCase() === peginTxid.toLowerCase()` next to the `vault_id` echo check. Both callers (`runDepositorPresignFlow`, `submitWotsPublicKey`) already held the txid, so no signature change reached the app. Kept both checks for the reason you give: the echo catches a mangled response, the txid catches a mispaired one. New test: "rejects a status whose attested pegin txid names another pegin".

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchAttribution.ts
kirugan: Agreed, and thanks for the caveat — it changed the fix. I did not compare the envelope id to the nested `result.vault_id`; as you say both are echoes of the same request string. The check is now on `result.pegin_txid`, made at the call sites that hold the expected txid:

- `usePeginPollingQuery`: compares against `deposit.activity.peginTxHash` (optional in the indexer type, so the check is skipped when it is absent) and sets a per-deposit error on mismatch.
- `usePegoutPolling`: compares against `entry.vault.peginTxHash` (required) and routes a mismatch to `applyPegoutFailure`.

This keeps `batchPollByProvider` generic — it has no way to know the expected txid for `TResult`.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/validators.ts
kirugan: Agreed — changed. The envelope `vault_id` is now `assertNonEmptyString`, not `assertVaultId`. A malformed echo therefore falls to `attributeBatchResults`, which classifies it as `unexpected`, and the requested item degrades alone through `onMissing`. The nested `result.vault_id` keeps its full format check.

`validators.test.ts` is updated: "rejects entry with wrong-length vault_id" becomes "accepts a per-item error entry whose echoed vault_id is malformed", plus a new case pinning that a non-string `vault_id` still throws.

services/vault/src/hooks/usePegoutPolling.ts
kirugan: Agreed — fixed. `onItem` now filters `error.includes("PegIn not found")` and routes it to a new `applyPegoutNotIngested`: the failure counter resets (the RPC itself succeeded), the unknown counter advances, so the vault still times out at `PEGOUT_MAX_UNKNOWN_STATUS_POLLS` with reason `unknown_status` — the path the deprecated txid RPC took when it answered with `found: false`. New test file `src/hooks/__tests__/usePegoutPolling.test.ts` covers both the pending tick and the give-up tick.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchAttribution.ts
kirugan: Agreed on the gap, with one change of placement. `normalizeVaultId` stays pure: it also runs on server echoes inside `attributeBatchResults`, and a throw there would turn one bad echo into a whole-chunk failure — the opposite of the envelope change in `validators.ts`.

Instead `assertVaultId` is refactored to an exported predicate `isVaultIdHex`, and `batchPollByProvider` applies it to every `getVaultId` result. A malformed id is never sent; the item gets `onItem` with a locally produced error naming the bad value, so the message no longer blames the provider. `onWholeBatchError` now receives only the items actually sent.

On `0X`: left as is, and recorded here — viem never emits it.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/batchPoll.ts
kirugan: Acknowledged, not fixed. Fixing it means turning `vaultIdToItem` into `Map<string, TItem[]>` and fanning every dispatch out over the list — more moving parts than the case deserves while `vault_id = keccak256(peginTxHash, depositor)` makes a same-depositor collision impossible. Leaving it for a follow-up rather than growing the dispatch loop here.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/__tests__/waitForPeginStatus.test.ts
kirugan: Done — both suffixes dropped.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/api.ts
kirugan: Fair flag, and the answer is not as comfortable as I assumed: `@babylonlabs-io/ts-sdk` is published publicly (`publishConfig.access: public`, latest on npm is 0.72.0), so this is a breaking change for anyone outside the monorepo, with no deprecation period — while btc-vault#2648 keeps the txid methods until after mainnet.

I have left the hard cut in place for now. Do you want thin deprecated wrappers (`getPeginStatus`, `batchGetPeginStatus`, `batchGetPegoutStatus` plus their param types) delegating to the vault-id methods for one release? Happy to add them in this PR if you or the SDK owner prefer that.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/validators.ts
```diff
@@ -507,33 +534,33 @@ function assertNullableString(value: unknown, field: string): void {
...
  * The inner result (when non-null) is validated via the single-item validator.
  */
 export function validateBatchGetPeginStatusResponse(
   response: unknown,
 ): asserts response is BatchGetPeginStatusResponse {
-  validateBatchEnvelope(response, "batchGetPeginStatus", (entry) => {
+  validateBatchEnvelope(response, "batchGetPeginStatusByVaultId", (entry) => {
     if (entry.result !== null) {
       validateGetPeginStatusResponse(entry.result);
     }
```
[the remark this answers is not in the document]
kirugan: Not taking this one as written, but the underlying gap is fixed a different way.

Comparing the nested `result.vault_id` to the envelope `vault_id` cannot detect a scrambled response: the server sets both from the id it parsed out of our own request (`vault_id: Some(vault_id_hex)` on the envelope, `vault_id: vault_id.to_string()` on the result), so the two agree by construction. A VP that paired envelope key A with vault B's payload would still pass that check.

The server-attested identifier is `result.pegin_txid`, which comes from a DB lookup. Both callers already hold the expected txid, so the corroboration now happens there:

- `usePeginPollingQuery` compares `result.pegin_txid` against `deposit.activity.peginTxHash`.
- `usePegoutPolling` compares it against `entry.vault.peginTxHash`.

A mismatch errors that one deposit instead of dispatching the payload. `batchPollByProvider` stays generic — it cannot know the expected txid for an arbitrary `TResult`.

Separately, the envelope `vault_id` assert was relaxed to a shape check, so one malformed id degrades one item instead of failing the whole chunk.

packages/babylon-ts-sdk/src/tbv/core/services/deposit/__tests__/waitForPeginStatus.test.ts
```diff
@@ -6,21 +6,23 @@ import type { PeginStatusReader } from "../interfaces";
 import { waitForPeginStatus } from "../waitForPeginStatus";
 
 const VALID_TXID = "a".repeat(64);
+const VALID_VAULT_ID = `0x${"1".repeat(64)}`;
 const TEST_TIMEOUT_MS = 60_000;
 const TEST_POLL_INTERVAL_MS = 100;
 /** Enough mock responses to outlast the timeout in timeout tests */
 const MOCK_RESPONSES_COUNT = 100;
```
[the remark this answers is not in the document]
kirugan: Not valid — no change made. The timeout test does not use `TEST_TIMEOUT_MS`; it declares `const shortTimeoutMs = 500` and passes that as `timeoutMs`. At `TEST_POLL_INTERVAL_MS` of 100 that is about 5 polls, well inside the 100 scripted responses. The other test that fills the array (`throws terminal when VP reports IngestionRejected`) throws on the second poll.

packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/validators.ts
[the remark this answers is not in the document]
kirugan: The checks are present on the current head (`f3ccf5d5`). Concrete lines:

- `services/vault/src/hooks/deposit/usePeginPollingQuery.ts:132-140` — `canonicalizeTxid(envelope.result!.pegin_txid) !== canonicalizeTxid(deposit.activity.peginTxHash)` sets a per-deposit error instead of dispatching.
- `services/vault/src/hooks/usePegoutPolling.ts:130-137` — same comparison against `entry.vault.peginTxHash`, routed to `applyPegoutFailure`.
- `packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/validators.ts:603` — the batch envelope `vault_id` is `assertNonEmptyString`, not `assertVaultId`. The nested `result.vault_id` keeps `assertVaultId` at lines 193 and 448.

Neither hook passes `envelope.result!` on without that comparison. Please re-check against `f3ccf5d5`.

gbarkhatov (approved):

jonybur (approved):
Reviewed the delta since @gbarkhatov's first pass. All seven inline threads from that review are addressed at head f3ccf5d5:

- The self-referential envelope `vault_id` guard is out; `waitForPeginStatus` now cross-checks `result.pegin_txid` against the caller-supplied txid, and both call sites (`runDepositorPresignFlow`, `submitWotsPublicKey`) thread it in.
- Batch attribution likewise cross-checks `result.pegin_txid` at the two dispatch sites (`usePeginPollingQuery` against `deposit.activity.peginTxHash`, `usePegoutPolling` against `entry.vault.peginTxHash`), gated on the txid being present.
- Envelope `vault_id` is validated with `assertNonEmptyString` instead of `assertVaultId`, so one malformed echo degrades a single item via `onMissing`/`unexpected` rather than failing the whole chunk. Nested `result.vault_id` keeps its full format check.
- Request-site `vault_id` hex validation lives at `batchPollByProvider` via the exported `isVaultIdHex`, so `normalizeVaultId` stays pure for use on server echoes inside `attributeBatchResults`.
- `usePegoutPolling` regains a benign path for "PegIn not found" via `applyPegoutNotIngested` — the RPC-success counter resets, the unknown counter advances, and the pegout still times out at `PEGOUT_MAX_UNKNOWN_STATUS_POLLS` with `unknown_status`, matching the deprecated txid RPC's `found: false` behavior.

Section 3 (payoutSigning) is a one-line `vaultId` thread-through only; payout re-derivation and the challenger-set equality assertion are untouched, so §3's substantive invariants are unaffected. §3's two-reviewer requirement still applies (this is the second).

kirugan:
Waiting for deploy of the feature on devnet next Monday. Don't merge before that please

services/vault/src/services/activity/claimTxResolver.ts
```diff
@@ -93,9 +90,9 @@ export async function resolveRedeemClaimTxids(
         });
         await batchPollByProvider<PerVaultEntry, GetPegoutStatusResponse>({
           items: entries,
-          getTxid: (e) => stripHexPrefix(e.peginTxHash),
-          batchCall: (pegin_txids) =>
-            rpcClient.batchGetPegoutStatus({ pegin_txids }),
+          getVaultId: (e) => e.vaultId,
```
jrwbabylonlab, line 93: The two polling hooks gained the `result.pegin_txid` cross-check, but this third consumer of the same RPC still trusts the echoed id. Same check here?

gbarkhatov (approved):

gbarkhatov (approved):

jeremytsngtsng (dismissed):

jeremy-babylonlabs (approved):

PR #2384, opened 2026-09-01
fix(wallet-connector): resolve Storybook buffer polyfill

jonybur:
## Summary

- Limit Vite Node polyfill aliases to `crypto`.
- Avoid bare shim aliases for `buffer`, `global`, and `process` under pnpm.
- Keep the plugin global injections.

This fixes the wallet connector Storybook resolution failure for `vite-plugin-node-polyfills/shims/buffer`. It uses the same configuration as the Vault and Simple Staking builds. It also matches the failure described in [vite-plugin-node-polyfills#81](https://github.com/davidmyersdev/vite-plugin-node-polyfills/issues/81).

## Verification

- Wallet connector Storybook static build passed.
- Wallet connector package build passed.
- Storybook Playwright tests passed: 214.
- Wallet connector Vitest tests passed: 194.
- Wallet connector lint passed with no errors.

### DIFF
--- a/packages/babylon-wallet-connector/vite.config.ts
+++ b/packages/babylon-wallet-connector/vite.config.ts
@@ -13,7 +13,7 @@ export default defineConfig({
       include: ["src"],
       exclude: ["src/**/*.stories.tsx", "src/**/*.test.ts", "src/**/*.test.tsx", "src/__fixtures__/**"],
     }),
-    nodePolyfills(),
+    nodePolyfills({ include: ["crypto"] }),
   ],
   build: {
     outDir: "dist",


packages/babylon-wallet-connector/vite.config.ts
```diff
@@ -13,7 +13,7 @@ export default defineConfig({
       include: ["src"],
       exclude: ["src/**/*.stories.tsx", "src/**/*.test.ts", "src/**/*.test.tsx", "src/__fixtures__/**"],
     }),
-    nodePolyfills(),
+    nodePolyfills({ exclude: ["buffer"] }),
```
gbarkhatov: Suggestion: this is now the third idiom in the repo for the same problem. `services/simple-staking/vite.config.ts:62-66` already uses `nodePolyfills({ include: ["crypto"] })` under a four-line comment describing precisely this failure — "aliases every bare `buffer` import to a bare shim specifier, which cannot resolve from a workspace dependency's dist under pnpm… vault does the same" — and `services/vault/vite.config.ts:138-144` matches it.

Two things that sibling form also closes, which `exclude: ["buffer"]` leaves open:

- **`global` and `process` keep the identical alias shape that caused this bug.** `dist/index.js:36-41` gives all three shims bare-specifier alias targets, and line 44 keeps every key `exclude` doesn't name — so a bundled dep doing `require('process')` still gets rewritten to a specifier it cannot resolve. Latent today; the next dependency reintroduces it.
- **`exclude: ["buffer"]` silently drops `node:buffer` too.** The comparator strips `node:` from both sides (`dist/index.js:6,9,35`) and node-stdlib-browser exports both keys, so one string removes two entries. A bare `node:buffer` would then be externalized with a warning rather than failing the build. Nothing in the browser graph imports it today.

One caveat if you do switch: `include: ["crypto"]` is strictly broader, not a drop-in — with `include` non-empty the plugin drops *every* alias except `crypto`. That needs a build check, not a find-and-replace.

Either way, a rationale comment here would earn its keep. This file already does exactly that ten lines down for `esmExternals`, and without one the next reader rediscovers this a fourth time.

gbarkhatov: Worth knowing before this ships: the published `dist` goes from one bundled `buffer` implementation to three.

The alias used to funnel every bare `buffer` import onto the single self-contained shim (`shims/buffer/dist/index.js` is 2371 lines with zero `import`/`require` — it inlines buffer + base64-js + ieee754). Now the bundle carries that shim via inject, **plus** `buffer@6.0.3` for the six `import { Buffer } from "buffer"` providers, **plus** `buffer@5.7.1` reached through `@keystonehq/keystone-sdk@0.12.3` → `@ngraveio/bc-ur@1.1.13` → `crc@3.8.0`, whose `lib/es6/create_buffer.js:7` does a bare `require('buffer')` and whose pnpm-local symlink is 5.7.1. keystone-sdk is deliberately not external, so it is in the graph.

Not a correctness bug — I checked the cross-realm case and it holds — but on the order of 120KB unminified of avoidable duplicate in a published package. `include: ["crypto"]` would collapse it back to the injected global only; short of that, it's worth a line in the description so it isn't discovered later as a mystery size bump.

gbarkhatov (dismissed):
Reviewed with three independent passes, then deduped and verified every claim against the installed `vite-plugin-node-polyfills@0.23.0` (`dist/index.js`, `shims/buffer/`), `node-stdlib-browser@1.3.1`'s exported map, and the pnpm symlink graph — not inferred from the option names. Approving; everything below is a suggestion, not a gate.

The change does what it claims. `exclude` feeds only the `resolve.alias` reduce, and `globals.Buffer` is never filtered by it, so global Buffer injection genuinely survives. Two inline notes: one on the option choice, one on what the published `dist` now carries.

Two things that don't fit on a diff line:

**Nothing in CI builds the thing this PR fixes.** `.github/scripts/visual-capture.sh:32` builds core-ui's Storybook only, and no workflow invokes `build-storybook`, so wallet-connector's Storybook build (`packages/babylon-wallet-connector/package.json:35`) never runs in CI. The verification in the description was local-only, which means this regression can come back silently with nothing to catch it. Worth a follow-up issue rather than anything in this PR.

**The description's mechanism is slightly off.** "Keeps global Buffer injection" is correct, but build-time inject still emits `{ Buffer: "vite-plugin-node-polyfills/shims/buffer" }` (`dist/index.js:60`) — the exact specifier the description calls unresolvable. What actually got fixed is narrower: the *alias* was rewriting third-party `import "buffer"` into a bare shim specifier that packages with no resolution path to the plugin could not follow. Also, the title scopes this to Storybook, but `.storybook/main.ts` has no `viteFinal` and inherits this config wholesale — this is the library build config, so the published `dist` changes too.

**Checked and refuted**, so they don't get re-raised later: there is no `buffer` version swap on the alias itself (`v("buffer")` returns the shim, overriding node-stdlib-browser's path, so 5.7.1 was never on that alias); there is no cross-realm `Buffer` correctness break (no `instanceof Buffer` anywhere in `src`, all copies duck-type `isBuffer` on `_isBuffer`, the package boundary is hex strings, and `bitcoinjs-lib` is external); nothing here reaches a CLAUDE.md critical path; and the `^0.23.0` range predates this PR in all five declaring packages.

**Verified clean:** `buffer` is exact-pinned at `6.0.3` in `dependencies`, and both the package-local and hoisted symlinks resolve to that same copy. All six first-party importers take only `{ Buffer }`, so no shim-only export (`Blob`, `File`, `isUtf8`, `transcode`) is lost in the swap. No `SECURITY_MODEL.md` covers this path and no critical path is touched. The red `validate` check is the "Validate commit signatures" step, not code — `verify` and `visual` both pass.

packages/babylon-wallet-connector/vite.config.ts
jonybur: Changed this to `include: ["crypto"]`. This matches Vault and Simple Staking and removes the `buffer`, `global`, and `process` aliases. The Storybook build, package build, Playwright, Vitest, and lint checks pass. I kept the rationale in the PR description so the configuration stays consistent with the sibling builds.

jonybur: Changed this to `include: ["crypto"]`. This removes the Buffer alias paths that caused the extra bundled copies. The published package build and Storybook build both pass.

jonybur:
Review updates are complete.

- Switched the polyfill aliases to include only crypto.
- Resolved both review threads.
- Replaced the rejected GitHub signature with a FIDO2 signature.
- Merged the current main branch with a FIDO2-signed commit.
- Storybook, package build, 214 Playwright tests, 206 Vitest tests, and lint pass on the merged head.

This is ready for fresh approval after CI completes.

jonybur:

gbarkhatov (approved):

PR #2391, opened 2026-09-03
feat(vault): polish the disconnected entry screen

### LINKED ISSUE #2337: Entry screen: logo colour, hero copy, feature list and testnet badge
## Context

The disconnected entry screen at `/` — the first thing a visitor sees before connecting a wallet. Six design fixes, all light theme unless stated. Rendered by `services/vault/src/components/simple/DisconnectedOverview.tsx`.

### 1. Aave logo colour

**Now:** the Aave wordmark in the header lockup renders purple, next to the orange Babylon wordmark.
**Expected:** black.
**Where:** `services/vault/src/components/shared/BrandLockup.tsx`

### 2. Hero description copy

**Now:** "Powered by Babylon Trustless Bitcoin Vaults protocol, collateralize native Bitcoin and borrow **stablecoins or WBTC** directly from Aave V4."
**Expected:** "Powered by Babylon Trustless Bitcoin Vaults protocol, collateralize native Bitcoin and borrow **supported assets** directly from Aave V4."
**Where:** `services/vault/src/copy.ts:1743`

### 3. Rate suffix

**Now:** the three borrowing-rate figures (USDT, USDC, WBTC) are suffixed `p.a.`
**Expected:** `APR`.
**Where:** `services/vault/src/copy.ts:1746` — `aprSuffix: "p.a."`

### 4. Feature list — remove the stroke

**Now:** the right-hand panel has an outer border and a divider between every row.
**Expected:** no stroke, no dividers.
**Where:** `services/vault/src/components/simple/DisconnectedFeatureCards/FeatureCard.tsx`

### 5. Feature list — cut from 5 rows to 3

**Now, in order:** 1 "Competitive borrowing rates", 2 "Native, trustless, and self-custodial", 3 "Fast access to liquidity", 4 "Partial liquidation supported", 5 "Verifiable, permissionless execution".
**Expected:** keep 1, 2 and 4. Remove 3 and 5, and update the copy on the three that stay.
**Where:** `services/vault/src/components/simple/DisconnectedOverview.tsx`; the removed rows' icons are `FastAccessIcon.tsx` and (row 5) its counterpart in `DisconnectedFeatureCards/` — delete them if nothing else uses them.

### 6. Testnet badge colour

**Now:** the `Testnet` chip is orange in both themes.
**Expected:** yellow in dark theme, dark blue in light theme.
**Where:** `services/vault/src/components/shared/NetworkBadge.tsx` (Figma node 10084:22952)

## Open question

Replacement copy for the three surviving feature rows is not specified. Needs the strings from design before item 5 can be finished.

<details>

> *Entry screen*
> • Aave logo → black colour
> • Change text in description to: "Powered by Babylon Trustless Bitcoin Vaults protocol, collateralize native Bitcoin and borrow supported assets directly from Aave V4."
> • p.a. change to → APR
> • Right block: no stroke and 3 features (remove 3rd + 5th and change text)
> • Testnet label change to yellow in dark theme and dark blue in the light one

</details>

jeremy-babylonlabs:
## Summary

Polish pass on the disconnected entry screen (`/`). Six design corrections: the Aave wordmark now inherits the lockup's text colour, the hero and rate copy are updated, the feature panel loses its stroke and dividers, and the list drops from five rows to three. The testnet badge moves off the amber warning token.

Closes #2337

## Changes

**Aave wordmark is a `currentColor` glyph.** `public/images/aave-wordmark.svg` hardcoded `fill="#9391F7"` on every path, so the mark rendered purple in both themes. Replaced the `` with an inline `AaveWordmark` component whose paths carry no fill, rendered as `text-accent-primary` — black on light, white on dark. The static asset is deleted; nothing else referenced it. Explicit `role="img"` and `aria-label="Aave"` preserve the accessible name the `alt` attribute used to supply.

**Hero body.** "borrow stablecoins or WBTC" → "borrow supported assets", so the sentence does not go stale as markets are added.

**Rate suffix.** `p.a.` → `APR`, matching the `Borrow APR` label the rest of the app already uses for the same figure.

**Feature panel.** Dropped the outer border and the hairline row dividers; the panel keeps its rounded corners and secondary background.

**Feature rows: five → three.** Kept competitive rates, self-custodial, and partial liquidation, in that order. Removed the fast-access and permissionless-execution rows along with their icon components and copy entries. The self-custodial and partial-liquidation bodies take the replacement copy from the design frame. The single-open accordion behaviour is unchanged — self-custodial is still the expandable row.

**Panel width and alignment.** The feature panel is capped at `max-w-[420px]` and, from `md` up, sits centred both horizontally in its grid column and vertically against the left column — matching the 420 x 304 frame. The grid still aligns every other child to the start.

**WBTC logo.** The app shipped a monochrome dark WBTC mark, which disappeared against a dark background. Replaced with the standard white and orange logo. The Figma node is an image fill with no vector export, so the asset is the exact 96px raster taken from it, saved as `public/images/wbtc.png`; `wbtc.svg` is deleted and `TOKEN_ICONS.WBTC` now points at the PNG. That entry feeds every WBTC surface — the landing rate row, markets and loans — so all three pick up the new mark.

**Testnet badge colour.** `text-warning-main` → `text-primary-main` on light and the app's `--risk-amber` value on dark, so the chip reads as a network indicator rather than a warning.

## Open items

- The design frame shows faint separators between the feature rows. These are deliberately not implemented: the issue text asks for no dividers, and the issue is the newer instruction. Flag it if the frame is meant to win.
- The badge colours are a token choice, not a Figma pull — the current Figma node still shows the previous amber. Worth a design confirmation before merge.

## Testing

- `pnpm exec eslint` on every changed file — clean.
- `pnpm exec tsc --noEmit -p tsconfig.lib.json` — 0 errors.
- `pnpm exec vitest run` over `DisconnectedOverview`, `NetworkBadge` and `RootLayout` — 24 tests pass.
- Updated the feature-order test to the new three-row list and the single expandable row, refreshed the `p.a.` assertions, and moved the `RootLayout` Aave lookups from `getByAltText` to `getByRole("img", { name: "Aave" })`.


services/vault/src/copy.ts
```diff
@@ -1882,21 +1883,13 @@ export const COPY = {
...
+          body: "Splitting your deposit between two Bitcoin vaults in accordance with AAVE's parameters will allow you to retain part of your position during the first liquidation.",
         },
         selfCustodial: {
           title: "Native, trustless, and self-custodial",
-          body: "No bridging. No wrapping. No pooled custody. Your native Bitcoin stays in a self-custodial vault — with no third party or signing quorum able to move or rehypothecate it.",
-        },
-        trustless: {
-          title: "Verifiable, permissionless execution",
-          body: "Collateral rules are enforced by code and cryptographic proofs — not by discretionary gatekeepers, committees, or off-chain liquidation decisions.",
+          body: "No bridging. No wrapping. No pooled custody. Your native Bitcoin stays in a self-custodial vault, without third party able to move it.",
```
gbarkhatov: **Merge-blocker.** "without third party able to move it" is missing its article — the clause isn't grammatical.

The rewrite replaced "with no third party or signing quorum able to move or rehypothecate it", and dropping "no" for "without" needs either an article or a plural: "without a third party able to move it", or simplest, keep the original construction — **"with no third party able to move it."**

Raising this as blocking rather than a nit because of where it lands: it renders verbatim at `DisconnectedOverview.tsx:293` on the disconnected entry screen, which is the first thing every new depositor reads, and nothing in the pipeline can catch it — lint, tsc, vitest and the visual snapshot all pass on a grammatically broken string. `copy.ts`'s own header names "one place to audit for English correctness and tone" as the file's first purpose, so this is the audit.

services/vault/src/copy.ts
```diff
@@ -1882,21 +1883,13 @@ export const COPY = {
...
           body: "Access to Aave V4 liquidity & its transparent, market-based variable rates.",
         },
-        fastAccess: {
-          title: "Fast access to liquidity",
-          body: "Vault setup and borrowing complete in about 3 hours.",
-        },
         partialLiquidation: {
           title: "Partial liquidation supported",
-          body: "for any loan position backed by multiple trustless Bitcoin vaults.",
+          body: "Splitting your deposit between two Bitcoin vaults in accordance with AAVE's parameters will allow you to retain part of your position during the first liquidation.",
```
gbarkhatov: **Merge-blocker.** `"AAVE's"` is the only uppercase-AAVE spelling in the entire copy file, and it ships on the landing page.

I grepped `AAVE` across all ~2000 lines of `copy.ts`: exactly one hit, this line. Every other reference is "Aave" — including `competitiveRates.body` four lines above at `:1884` ("Access to Aave V4 liquidity") and the `common.aaveWordmark: "Aave"` constant this same PR adds.

So two brand spellings render inside the same `max-w-[420px]` panel on the same screen, four lines apart in the source. That's precisely the drift `copy.ts` exists to eliminate.

`Aave's`, or "Aave V4's parameters" to match the neighbour.

services/vault/src/components/simple/DisconnectedOverview.tsx
```diff
@@ -234,22 +231,11 @@ export function DisconnectedOverview({
...
       },
-      {
-        icon: <FastAccessIcon />,
-        title: features.fastAccess.title,
-        body: features.fastAccess.body,
-      },
       {
         icon: <PartialLiquidationIcon />,
         title: features.partialLiquidation.title,
         body: features.partialLiquidation.body,
```
gbarkhatov, line 237: The disclosure affordance has ended up on the wrong row: the longest body is the one that can never collapse, while a shorter one is clamped to a single line.

`FeatureCard.tsx:21` computes `showFull = !expandable || expanded`, so a row without `expandable` always renders its body in full. This PR grew `partialLiquidation.body` from a short fragment to a 161-character sentence but left this entry non-expandable — while `selfCustodial`, now the *shorter* string at 133 characters, is the sole `expandable: true` row and is `line-clamp-1` when collapsed.

Inside the new `max-w-[420px]` cap that means row 3 renders across three-plus permanent lines while row 2 is truncated to one with a chevron beside it. Visually it reads as though the tallest block is the one that can't be collapsed, which is backwards.

Adding `expandable: true` here should be all it takes — the prop and the `expandedTitle` single-open state already support it, so no new machinery. Shortening the body would work equally well if the panel is meant to stay compact.

services/vault/src/components/simple/DisconnectedOverview.tsx
```diff
@@ -264,7 +250,7 @@ export function DisconnectedOverview({
 
         <h1
           aria-label={HERO_TITLE_TEXT}
-          className="mt-6 text-[clamp(2rem,5vw,3rem)] font-normal leading-[1.167] text-accent-primary"
+          className="mt-[26.67px] text-[clamp(2rem,5vw,3rem)] font-normal leading-[1.167] text-accent-primary"
```
gbarkhatov, line 254: `mt-[26.67px]` is a fractional-pixel magic number replacing the scale value `mt-6` (1.5rem = 24px), and it isn't one of issue #2337's six items.

Two costs for the 2.67px gain. It's a px value, so unlike `mt-6` it stops scaling with root font-size — while the heading it sits above is `clamp(2rem, 5vw, 3rem)` and does scale, so the gap drifts relative to the text at large type sizes. And it leaves the spacing scale every sibling in this file still uses (`mt-3`, `mt-4`, `mt-6` at :271, :275, :278, :282).

26.67 reads as a Figma auto-layout artifact, and CLAUDE.md's No Magic Numbers rule asks that a number in code make it obvious why that value was chosen. Reverting to `mt-6` is the simple option; if the exact value matters, `mt-[1.667rem]` with a one-line Figma citation would keep it scaling and explain itself — this file already does exactly that for `leading-[1.167]` and `tracking-[0.15px]`.

services/vault/src/components/shared/NetworkBadge.tsx
```diff
@@ -14,7 +14,11 @@ export function NetworkBadge() {
 
   return (
     <span className="flex items-center rounded-full bg-secondary-highlight px-2.5 py-1">
-      <Text as="span" variant="caption" className="text-warning-main">
+      <Text
+        as="span"
+        variant="caption"
+        className="text-primary-main dark:text-[rgb(var(--risk-amber))]"
```
gbarkhatov: Two things here, the second more substantive than the first.

`dark:text-[rgb(var(--risk-amber))]` hand-rolls a utility that already exists: `services/vault/tailwind.config.ts:23` registers `risk-amber` as `rgb(var(--risk-amber) / <alpha-value>)`. The arbitrary form resolves to the same colour but loses alpha-modifier support, so `dark:text-risk-amber` is strictly better.

More importantly, `globals.css:24-32` documents the `--risk-*` scale as v3 risk-card semantics and singles amber out as "caution". Borrowing it for a network chip means a future retune of the liquidation-risk palette silently recolours the header — the two have no reason to move together. A dedicated network-badge token would decouple them.

One latent thing while you're in here: the class pairs a tw-colors *theme* token for light with a Tailwind *variant* for dark. `primary-main` already swaps itself to `#111111` in dark (about 1.14:1 on the badge's `#252525`), so the badge is legible only because `darkMode: "class"` fires — while `globals.css:87-88` deliberately also supports a `[data-theme="dark"]` hook that the `dark:` variant doesn't see. Not reachable today; it would be if anything ever sets the attribute without the class.

services/vault/src/components/simple/__tests__/DisconnectedOverview.test.tsx
```diff
@@ -139,9 +139,7 @@ describe("DisconnectedOverview", () => {
     const titles = [
       features.competitiveRates.title,
       features.selfCustodial.title,
```
gbarkhatov, line 141: Cutting five rows to three is item 5 and this PR's central behavioural change, and I don't think either updated assertion can fail if a fourth row comes back.

`getByText` on the three titles passes regardless of extra siblings, and `getAllByRole("button", { expanded: false })).toHaveLength(1)` counts only expandable rows — a non-expandable fourth row carries no `aria-expanded` at all, so it's invisible to that query. Asserting the rendered row count directly would pin the actual change.

Separately, at lines 122-123 the APR-suffix assertion interpolates `COPY_OVERVIEW.aprSuffix` on **both** sides of the expectation, so it's a tautology — reverting `"APR"` to `"p.a."` leaves the suite green and item 3 has no real coverage either. `RootLayout.test.tsx` gets this right by pinning the literal `"Aave"`; the same treatment here would work.

services/vault/src/services/token/tokenService.ts
```diff
@@ -25,7 +25,7 @@ const btcConfig = getNetworkConfigBTC();
 const TOKEN_ICONS: Record<string, string> = {
   BTC: btcConfig.icon,
   SBTC: btcConfig.icon,
-  WBTC: "/images/wbtc.svg",
+  WBTC: "/images/wbtc.png",
```
gbarkhatov, line 28: This swap makes WBTC the only raster in the map — `btc.svg`, `usdc.svg`, `usdt.svg` and `dai.svg` all stay vector — and its reach is much wider than the entry screen.

I checked the asset: 96×96 8-bit RGBA, renders correctly, nothing is broken. But `Avatar size="large"` is 40px and `xlarge` is 64px, so a 3× display asks 120–192 device px of a 96px source. And `TOKEN_ICONS` is read by borrow markets, asset selection, loans and repay — none of which the `visual` job can reach, so any softness ships unobserved.

It's also outside issue #2337's six items, which is what makes it worth asking about rather than just noting: if the vector had to be dropped for a reason (a bad path, a licensing artifact), that's worth a line in the PR body. Otherwise re-exporting the mark as SVG keeps the set consistent.

services/vault/src/components/shared/AaveWordmark.tsx
```diff
@@ -0,0 +1,26 @@
+import { COPY } from "@/copy";
+
+/**
+ * Aave wordmark, drawn as a `currentColor` glyph so it inherits the lockup's
+ * text colour instead of shipping its own brand purple.
+ */
+export function AaveWordmark({ className }: { className?: string }) {
+  return (
+    <svg
+      viewBox="0 0 109 18"
```
gbarkhatov, line 12: The SVG carries a `viewBox` but no `width`/`height` and no default class, so all of its dimensions live in the caller's *optional* `className`. `<AaveWordmark />` with no props renders at the browser's default replaced-element size rather than the 109×18 lockup mark.

Nothing is broken today — the single call site at `BrandLockup.tsx:16` passes `h-[18px] w-[109px] text-accent-primary` — so this is a trap for the next caller rather than a live bug. Folding a default in with `twMerge("h-[18px] w-[109px]", className)`, or just adding explicit `width`/`height` attributes, makes the component safe to use bare.

Worth raising while this file is new: `packages/babylon-core-ui/src/components/Sidebar/SidebarBrandLockup.tsx` already draws the Aave wordmark as `currentColor` paths for the connected chrome. The app now carries the same vector at two scales in two places, and they'll drift the next time the brand mark changes. Might be worth core-ui exporting the glyph and both lockups composing it — though that's plausibly a follow-up rather than this PR.

gbarkhatov (dismissed):
Reviewed all six issue items, resolving every colour token through `services/vault/src/globals.css` before asserting anything about it, and re-running the zero-dead-code sweep myself across `services/` and `packages/`.

The deletion work here is clean in a way worth calling out, because it's the exact defect class that turned out to be a merge-blocker on two sibling PRs in this batch: both removed icons and both removed copy keys go out with their only reader in the same diff, and `ICON_SIZE` correctly stays behind, still read by all three surviving icons. Grep for `fastAccess`, `FastAccessIcon`, `TrustlessIcon`, `wbtc.svg`, `aave-wordmark`, "Fast access to liquidity" and "permissionless execution" returns zero hits.

The two blockers are both new landing-page copy, flagged inline. They're worth taking seriously precisely because nothing can catch them: lint, tsc, vitest and the snapshot all pass on a grammatically broken string, and this is the first screen a new depositor reads.

I also want to record four claims I chased and **disproved**, so nobody re-raises them:

- "`text-accent-primary` renders the Aave mark teal `#12495E`, not black." No — `globals.css:79` re-points `--twc-accent-primary` to `0 0% 0%` in light. Reading core-ui's `tailwind.config.js:212` alone gives the wrong answer.
- "`text-primary-main` isn't the dark blue item 6 asks for." It is — core-ui light `primary.main` is `#042F40` with no vault override.
- "The dark amber isn't the yellow the issue names." `globals.css:90` sets `--risk-amber: 255 179 0` (`#FFB300`) in dark, and you've already flagged badge colour as an open design question anyway.
- "`wbtc.png` has a transparency or clipping problem." Decoded the file: full-bleed opaque disc, transparent corners, clips cleanly under `Avatar`'s `rounded-full` + `object-cover`. Only the raster-vs-vector point survives.

**One consequence to put in front of design rather than a defect.** Item 4 asks for "no stroke, no dividers", and `DisconnectedOverview.tsx:287` delivers exactly that. The result in the default light theme is that `bg-background-secondary` `#F9F9F9` sits on the page's `surface` `#ffffff` at roughly 1.04:1 — so the container has effectively no boundary and the three rows read as loose text beside the hero. Dark is fine (`#202020` on `#121212`). This is per contract, so I'm not calling it a finding; it's just the concrete place where "no stroke" and "keep the panel" collide, and it belongs next to the divider question your PR body already opens.

**Open question worth adding to the body:** item 5 asks for updated copy on all three surviving rows, but `competitiveRates.body` (`copy.ts:1884`) is unchanged. The issue's own open question says the replacement strings were never specified, so this may be deliberate and complete — it's the one item I can't settle from the readable contract.

**On CI — and unlike this PR's siblings, the coverage here is real.** `VISUAL_TARGETS` includes `{ name: "overview", path: "/" }` at both desktop 1280×800 and mobile 390×844 with `fullPage` capture, so the layout findings are genuinely in frame. Two caveats: the job reports into a sticky comment rather than gating the merge, and it only ever photographs light theme — so the dark half of the badge and wordmark changes has no automated detection at all.

Also verified clean: the accessible name survives the `` to `role="img" + aria-label` migration and the reworked `getByRole("img", { name: "Aave" })` assertions aren't weaker (core-ui's `SidebarBrandLockup` names itself "Babylon Aave" and the matcher is exact, so the mainnet-absence case still discriminates); the inline SVG is a faithful transplant of the deleted asset's paths in the same order, and the dropped `clipPath` was a no-op against the viewBox; no critical path is touched; and the motion rules don't bind, since `FeatureCard`'s chevron already reads `--motion-duration-icon` / `--motion-ease-icon` and is unchanged.

gbarkhatov (approved):

jonybur (approved):
The delta since gbarkhatov's earlier pass addresses the two landing-page copy blockers, drops the two removed feature icons cleanly (grep shows no remaining refs to FastAccessIcon / TrustlessIcon / their copy keys), and moves the badge to the registered risk-amber utility. The AaveWordmark component correctly inherits currentColor and the RootLayout test migration to getByRole({ name: "Aave" }) preserves the accessible-name assertion — the exact-name matcher still discriminates on the mainnet-absence path.

data-testid="feature-row" is scoped to the DisconnectedOverview unit test (getAllByTestId toHaveLength(3)); it is not consumed by real-wallet E2E, so the CLAUDE.md pointer-comment rule does not bind here.

No critical-path files touched. Copy migrated to COPY.common.aaveWordmark. LGTM.

PR #2392, opened 2026-09-03
fix(vault, core-ui): polish the Overview page and sidebar

### LINKED ISSUE #2340: Overview page: logo, position block, disabled buttons, sidebar
## Context

The Overview page at `/` once a wallet is connected. Screenshot is light theme with an empty position ($0 across the board). Rendered by `services/vault/src/components/simple/DashboardPage.tsx` and `OverviewSection.tsx`.

### 1. Sidebar Babylon logo colour

**Now:** the Babylon wordmark in the sidebar renders black in light theme, while the same wordmark on the entry screen is orange.
**Expected:** orange, matching the entry screen.
**Where:** `services/vault/src/components/shared/BrandLockup.tsx`

### 2. Remove percent lines from the position block

**Now:** "Available to borrow" has a progress bar with `0% remaining` beneath it, and "Total borrowed" has one with `0% borrowed`.
**Expected:** both progress bars and their percent labels removed.
**Where:** `services/vault/src/components/simple/OverviewSection.tsx`

### 3. Total collateral value on one line

**Now:** `$0 USD` on the first line with `0 sBTC` on a second line beneath it.
**Expected:** the USD and BTC amounts on a single line.
**Where:** `services/vault/src/components/simple/OverviewSection.tsx`

### 4. Disabled button label colour

**Now:** the disabled `Borrow` and `Repay` buttons use `disabled:text-accent-disabled`, which reads as very low contrast against the grey fill.
**Expected:** a different label colour.
**Where:** `services/vault/src/components/shared/buttonClasses.ts`

### 5. Remove the total collateral value tooltip

**Now:** an info icon sits next to the "Total collateral value" label.
**Expected:** removed. The "Collateral Factor" tooltip lower on the page is not in scope.
**Where:** `services/vault/src/components/simple/OverviewSection.tsx`

### 6. Liquidation Analysis text

**Now:** the block carries its current explanatory text.
**Expected:** replacement copy — see the open question.
**Where:** `services/vault/src/components/simple/LiquidationAnalysisSection.tsx`, strings in `services/vault/src/copy.ts`

### 7. Remove the GitHub link from the sidebar

**Now:** the sidebar footer social row is GitHub, Telegram, LinkedIn, email, Discord, X.
**Expected:** GitHub removed. The entry-screen footer already drops it — see the comment at `EntryFooter.tsx:21` — so this aligns the two.
**Where:** `services/vault/src/components/shared/SidebarFooter.tsx:25`

### 8. Sidebar nav hover state

**Now:** no orange hover on the nav items.
**Expected:** Overview, Vaults, Loans, Activity and Liquidations all go orange on hover.
**Where:** `services/vault/src/components/shared/AppSidebar.tsx`

## Open question

Replacement copy for the Liquidation Analysis block (item 6) is not specified. Needs the string from design before that item can be finished.

<details>

> *Overview Page*
> • babylon logo -> orange colour in light theme
> • remove line with percent in position block, for the TCV put $ and BTC in one line
> • change label colour on disabled buttons
> • remove tooltip on "total collateral value"
> • change text in liq analysis block
> • remove github link from the sidebar
> • hover state on pages (overview, vaults and etc) should be orange

</details>

jeremy-babylonlabs:
Design polish pass on the connected Overview page, plus the three sidebar fixes from the same review.

## What changed

**Sidebar brand lockup.** The Babylon wordmark is now orange (`secondary-main`) in light mode and `accent-primary` in dark. The Aave mark keeps its black/white treatment. Done from the vault with a Tailwind arbitrary variant scoped to the SVG's `#Logo` group, so `SidebarBrandLockup` in `core-ui` is untouched and every other consumer is unaffected.

**Sidebar nav hover.** An inactive nav row now hovers to `secondary-main` (orange) instead of `accent-primary`. Active rows still render `accent-primary`. This is a `core-ui` change because `SidebarItem` composes the class with `twJoin`, so a consumer-side `className` cannot override it on stylesheet order.

**Position meters removed.** The progress bars and their "% remaining" / "% borrowed" labels are gone from the position summary. The deletion goes all the way down: the `meter` field on `PositionStatCard`, the `StatMeter` component, `formatMeterLabel`, the six `overview.*Meter*Label` strings, and the percentage computations in `DashboardPage`, `Loans` and `Liquidations`. The Liquidations position card shared the same primitive, so its borrowed meter goes with it. With no consumer left, `maxTotalDebtUsd` also comes off the `useDashboardState` and `useLiquidationsPageState` return shapes; `calculateBorrowCapacityUsd` still computes it and keeps its own tests.

**Total collateral value on one line.** The USD value and the BTC amount now render on a single line, with the BTC amount as secondary text, instead of stacking as a value plus a caption row.

**Disabled neutral button label.** Disabled Borrow / Repay labels now use `accent-secondary` instead of `accent-disabled`. Applied to the inline class in `PositionStatCards` and to `NEUTRAL_BUTTON_CLASS` / `NEUTRAL_ROW_BUTTON_CLASS` so every neutral button agrees.

**GitHub removed from the sidebar footer.** The link, its icon import, and the now-unreferenced `SOCIAL_LINK_URLS.github` constant are deleted. Comments in `SidebarFooter` and `EntryFooter` that described the old GitHub difference between the two surfaces are corrected, along with a stale `SidebarFooter` claim that the other footer draws its icons from `react-icons`.

## Not in this PR

- The total collateral value tooltip is already absent on `main` — no change needed.
- The Liquidation Analysis copy replacement is not specified by design, so `LiquidationAnalysisSection` is untouched and that item stays open.

## Testing

- `pnpm --filter @babylonlabs-io/core-ui build`
- `pnpm exec eslint` on every changed file, and on `core-ui` `SidebarItem.tsx`
- `pnpm exec tsc --noEmit -p tsconfig.lib.json` in `services/vault`
- `pnpm exec vitest run` on the Overview, PositionStatCards, DashboardPage, AppSidebar, Loans, Liquidations and formatting suites
- `pnpm exec vite build` in `services/vault`, with the generated CSS checked for the `#Logo` selector

Closes #2340


jonybur (dismissed):
Sweep is clean. Meter-related props, StatMeter, formatMeterLabel, the six copy strings, borrowedRatio and maxTotalDebtUsd all drop together with no dangling refs. calculateBorrowCapacityUsd still returns maxTotalDebtUsd as noted. The scoped [&_#Logo] arbitrary variant is a reasonable way to keep the core-ui component untouched. LGTM.

services/vault/src/components/pages/Liquidations/useLiquidationsPageState.ts
```diff
@@ -52,7 +46,6 @@ export function useLiquidationsPageState(
     collateralBtc,
```
gbarkhatov, line 46: **Merge-blocker (zero dead code).** This destructure was the last production reader of `maxTotalDebtUsd`, so the field is now computed, exported and pinned only by its own tests.

I grepped the whole tree. The two surviving consumers of `calculateBorrowCapacityUsd` take only `availableToBorrowUsd` (`hooks/useDashboardState.ts:57`, `applications/aave/components/LoanCard/Borrow/hooks/calculateMaxBorrowTokens.ts:49`), and the `maxTotalDebtUsd` at `e2e/real/borrowParams.ts:309` is an unrelated local. That leaves the `BorrowCapacityUsd` field at `applications/aave/utils/borrowCapacity.ts:14` with exactly seven readers, all assertions in `__tests__/borrowCapacity.test.ts` (lines 21, 35, 43, 55, 67, 78, 92).

Nothing fails to compile, which is precisely the shape the rule targets — an interface field alive only because a test asserts it.

The fix looks small: keep `maxTotalDebtUsd` as the local at `borrowCapacity.ts:26` (line 33 still needs it), drop it from the `BorrowCapacityUsd` interface, and rewrite those assertions against `availableToBorrowUsd` with `currentDebtUsd: 0`, which yields the identical expected number. `borrowCapacity.ts` isn't in this PR's changed-file list, but this PR is the sole cause.

services/vault/src/components/pages/Liquidations/__tests__/Liquidations.test.tsx
```diff
@@ -793,10 +783,10 @@ describe("Liquidation Dashboard position override", () => {
       ),
     ).toBeInTheDocument();
     expect(
-      screen.getByText(
-        COPY.overview.borrowedMeterLabel(expectedBorrowedPercent),
-      ),
-    ).toBeInTheDocument();
+      screen.queryByRole("progressbar", {
+        name: COPY.liquidations.position.totalBorrowed,
```
gbarkhatov: This assertion can never fail — it queries a role nothing in this tree can emit.

Swapping the positive check on `COPY.overview.borrowedMeterLabel` for a negative `queryByRole("progressbar", ...)` looks like a faithful translation of "the meter is gone", but `StatMeter` was the only `role="progressbar"` reachable from `PositionOverview`, and this PR deletes it. A tree-wide grep leaves exactly one surviving `role="progressbar"` in `services/vault/src` — `simple/DepositProgressView/ProgressBar.tsx` — which this render never mounts. So the assertion is permanently green and verifies no behaviour of code in `src/`.

It's also bolted onto a test named "derives the USD caption from the override and the live BTC price", so it's testing something the name doesn't claim.

Suggest just deleting lines 785-789 and leaving the test to its one stated behaviour — the absence of the meter is already pinned by the component no longer existing.

services/vault/src/components/shared/AppSidebar.tsx
```diff
@@ -74,7 +74,9 @@ export function AppSidebar() {
   return (
     <Sidebar
       className="top-[var(--tbv-top-banner-height,0px)] h-[calc(100svh_-_var(--tbv-top-banner-height,0px))]"
-      brand={<SidebarBrandLockup className="text-black dark:text-white" />}
+      brand={
+        <SidebarBrandLockup className="text-black dark:text-white [&_#Logo]:text-secondary-main dark:[&_#Logo]:text-accent-primary" />
```
gbarkhatov: Two things about this one line, both worth a look before merge.

**It hangs off a Figma-export `id` inside another package.** `[&_#Logo]:text-secondary-main` compiles to a descendant selector reaching into `packages/babylon-core-ui/src/components/Sidebar/SidebarBrandLockup.tsx:18`. It works today — `<g id="Logo">` exists and every path under it is `fill="currentColor"` — but that id is a Figma auto-export artifact whose siblings are `Vector`, `Vector_2`, `Line 13` and `Logo-purple 1`. Nothing marks it as a public contract: no type, no test asserting the wordmark colour, and `Sidebar.stories.tsx:51` renders `<SidebarBrandLockup />` with no className. A re-export that renames the group, or any SVGO pass that strips ids, silently reverts the wordmark to black with no compile error and no failing test. A `wordmarkClassName` prop on `SidebarBrandLockup` applied to that group would make it a real interface — and that file sits one directory from `SidebarItem.tsx`, which this PR already edits.

**The result may be the pairing Figma already rejected.** Item 1 asks for orange "matching the entry screen". This scopes orange to `<g id="Logo">`, but the Aave mark is the sibling `<g id="Logo-purple 1">` (`SidebarBrandLockup.tsx:34-42`), whose seven paths are all `fill="currentColor"` and so inherit `text-black` from the wrapper. The entry screen deliberately does the opposite — `BrandLockup.tsx:4-10` records that "Aave is an image rather than a `currentColor` glyph so it keeps its own brand colours instead of inheriting Babylon's — per Figma comments #113 / #121, which rejected the all-black lockup." Before this PR the sidebar was a uniform all-black lockup; after it, it half-matches the entry screen while still carrying the rejected black Aave mark. Either reuse `BrandLockup.tsx:16-20`'s treatment for the Aave half, or get design to confirm the sidebar is exempt.

services/vault/src/components/shared/buttonClasses.ts
```diff
@@ -17,7 +17,7 @@ export const ROW_BUTTON_MIN_WIDTH_PX = 120;
 /** 36px-tall neutral action button. `ml-auto` keeps it flush right when the
  *  row wraps. */
 export const NEUTRAL_ROW_BUTTON_CLASS =
-  "ml-auto flex h-9 min-w-[120px] shrink-0 items-center justify-center whitespace-nowrap rounded-lg bg-secondary-strokeLight px-4 text-sm leading-[1.43] tracking-[0.17px] text-accent-primary transition-[filter] enabled:hover:brightness-110 disabled:cursor-not-allowed disabled:text-accent-disabled";
+  "ml-auto flex h-9 min-w-[120px] shrink-0 items-center justify-center whitespace-nowrap rounded-lg bg-secondary-strokeLight px-4 text-sm leading-[1.43] tracking-[0.17px] text-accent-primary transition-[filter] enabled:hover:brightness-110 disabled:cursor-not-allowed disabled:text-accent-secondary";
```
gbarkhatov, line 20: The change is right, but "so every neutral button agrees" doesn't quite hold — two neutral disabled controls keep the old token.

`SimulationToolbar.tsx:66` and `:150` hand-roll the same neutral control and still carry `disabled:text-accent-disabled`. Resolved in the vault's light theme (`defaultTheme: "light"`, `providers.tsx:35`) that's `#9ab7c2` on `secondary-highlight` `#F9F9F9` — about 1.9:1 — while the controls fixed here now read `#666666` on `#dddddd`, about 4.2:1.

The low contrast in `SimulationToolbar` is pre-existing and not this PR's doing. What's new is the *divergence between visually identical controls*, which is the thing centralising the constant was meant to prevent. Having `SimulationToolbar` import `NEUTRAL_BUTTON_CLASS` instead of repeating the string would close it and make the PR body's claim true.

packages/babylon-core-ui/src/components/Sidebar/SidebarItem.tsx
```diff
@@ -23,7 +23,7 @@ export const SidebarItem = ({
       "flex h-10 items-center gap-2 rounded py-2 text-base tracking-[0.15px] transition-colors",
```
gbarkhatov, line 24: The comment above this line states the state model exhaustively, and it no longer covers the state being added.

Lines 11-14 cite Figma node 10084:23112 and assert "Active state is a text-color change only (text/primary vs text/secondary) — Figma defines no background or border treatment for the active row." The hover colour isn't in that node at all; it comes from issue #2340 item 8. Since the comment presents itself as the complete model, the next person reading it will believe the hover behaviour isn't specified anywhere — worth recording the colour and citing item 8 as its source rather than the Figma node.

Separately: `hover:text-secondary-main` sits only on the inactive branch, so the row you're currently on never goes orange on hover. Item 8 enumerates all five entries with no active-state carve-out, and hovering your current section is the common case — so this may be intended, but it's worth confirming with design rather than leaving it implied by the branch structure.

services/vault/src/components/shared/SidebarFooter.tsx
```diff
@@ -17,12 +16,11 @@ interface SidebarSocialLink {
...
 
-// Matches Figma's Social Links block exactly (icon order + Terms/Privacy
-// text). Name+URL pairs are shared with the page footer (`config/socialLinks`)
-// so a handle change can't drift between the two; only the icons differ
-// (Figma-exact here vs react-icons in the v2 footer).
+// Figma's Social Links block (icon order + Terms/Privacy text), minus the
+// GitHub link. Name+URL pairs are shared with the entry footer
+// (`config/socialLinks`) so a handle change can't drift between the two; only
+// the per-surface presentation (icon size, hover color) differs.
 const SIDEBAR_SOCIAL_LINKS: SidebarSocialLink[] = [
```
gbarkhatov: With GitHub gone, `SIDEBAR_SOCIAL_LINKS` and `ENTRY_SOCIAL_LINKS` (`EntryFooter.tsx:22-28`) are byte-identical: the same five name/url pairs, in the same order, with the same `Icon` components imported from the same package.

GitHub was the only substantive difference between them — which the comments here correctly acknowledge and update. What's left is two hand-maintained copies of one list, and the real difference (per-surface icon size and hover class) already lives at the render sites and is already parameterised.

Two copies of an ordered list will drift in order or membership on the next change, and nothing would catch it. Collapsing to a single ordered list in `config/socialLinks.ts` — which already owns the URLs — and keeping only the presentation local would remove the class of bug entirely. Not blocking, and reasonable to defer if you'd rather keep this PR to the issue's scope.

services/vault/src/components/simple/PositionStatCards.tsx
```diff
@@ -12,10 +11,6 @@ export interface PositionStatCard {
    *  `value` when set; `value` is still used for accessibility fallbacks. */
```
gbarkhatov: This JSDoc is now false, and this PR adds a second card that relies on the claim.

It says `valueNode` "overrides `value` when set; `value` is still used for accessibility fallbacks" — but line 41 is a plain `card.valueNode ?? card.value` and no `aria-label` consumes `value` anywhere in the file. `OverviewSection.tsx:59-67` now passes both `value: totalCollateralValue` and a `valueNode` rendering the same string, so the figure is supplied twice with one copy never reaching the DOM.

`value` is a required field so it can't simply be dropped. Two ways out: correct the comment, or make it true with `aria-label={card.value}` on the span at line 40 — the latter would also give the collateral figure a stable accessible name, which is worth having now that its visible form is a composed node rather than a plain string.

gbarkhatov (dismissed):
Reviewed all eight issue items across both packages and traced the deletions tree-wide. The removal work is unusually clean: I grepped for `StatMeter`, `formatMeterLabel`, every `*MeterLabel`, `meterBaseUsd`, `borrowedRatio`, `availableMeterPercent`, `borrowedMeterPercent` and `SOCIAL_LINK_URLS.github` across `services/` and `packages/` and got **zero** hits — source, tests, mocks, stories and docs all moved in lockstep. Every load-bearing `data-testid` survives with its consuming-file comment.

Two things I'd want changed before merge. One is inline; the other is in the PR description itself:

**`Closes #2340` will auto-close an issue this PR deliberately leaves unfinished.** The body's own "Not in this PR" section says item 6 (Liquidation Analysis copy) is unspecified by design and "that item stays open" — but `Closes` is a GitHub closing keyword, so merging closes #2340 and the outstanding design question goes with it. The issue is the only record of that item. `Part of #2340`, or splitting item 6 into its own issue first, keeps it alive. One word, but the loss it prevents is real.

A note on the core-ui scoping question, because I went looking for a reason to push back and didn't find one: putting the hover recolour in `SidebarItem` rather than `AppSidebar` is correct. The pre-PR inactive class was already `hover:text-accent-primary`, and `SidebarItem.tsx:2` uses `twJoin` rather than `twMerge`, so a consumer-side `className` would emit both hover utilities and let stylesheet order — not class-attribute order — pick the winner. Blast radius is nil: `SidebarItem`'s only consumers repo-wide are `AppSidebar.tsx:58` and the story.

Two smaller ones that aren't inline because their lines fall outside a diff hunk:

- **`Sidebar.stories.tsx:10,36`** still renders `GithubIcon` in `exampleFooter`, which mirrors the app's social row icon-for-icon down to the "Terms of Use - Privacy Policy" line. After item 7 the canonical rendered example of `Sidebar` disagrees with the shipped product. The icon itself correctly stays exported at `Icons/index.ts:53` — that's public core-ui API, not dead — it's just the story that wants updating.
- **`PositionStatCards.tsx:40` / `OverviewSection.tsx:62`** — the nowrap holding the USD and BTC figures together is `xl:whitespace-nowrap`, and below `xl` the literal `{" "}` is a legal break point, so a long USD figure plus the BTC amount can still wrap to two lines: the exact result item 3 asks to remove. It only bites at mobile with large values and is no worse than the caption it replaces, so it may well be fine — but if one line is meant unconditionally, the `xl:` prefix wants dropping plus truncation.

**On CI, in both directions.** I was initially going to say the connected Overview has no visual coverage — that's wrong. `e2e/visual/depositFlow.visual.spec.ts` connects an injected wallet and captures `DEPOSIT_FLOW_STOPS.connected` at both viewports, so items 2-4 *are* photographed; the "no connected wallet" rule in `targets.ts` binds only the six route targets. The opposite error is equally wrong though: the `visual` job reports into a sticky comment and its only `exit 1` sites concern failing to publish images, so green doesn't mean the pixels are right. Hover state is captured by nothing, so item 8 ships with no automated verification at all.

Also checked and dismissed: item 5 really was already done on `main` (`OverviewSection.tsx:57-67` carries no `tooltip`, and `PositionStatCards.tsx:28` renders `Hint` only when one is present); the meter deletion causes no card-alignment regression, since `StatSection` is `items-center` inside an `items-stretch` row and the 3-row/2-row asymmetry predates this PR (card 3 has never had a caption); no critical path is touched; and the motion rules don't bind — `transition-colors` on `SidebarItem.tsx:23` is pre-existing and untouched.

Item 4 is a genuine contrast fix, resolved through the vault's own token overrides rather than the core-ui config: roughly 1.5:1 to 4.2:1 in the default light theme.

gbarkhatov (approved):

jonybur (approved):
Address-review delta correctly resolves the core-ui scoping via a new wordmarkClassName prop instead of the [&_#Logo] arbitrary variant, drops the now-orphaned maxTotalDebtUsd from the borrow-capacity interface, and aligns SimulationToolbar's disabled label with NEUTRAL_BUTTON_CLASS's accent-secondary. Social links now share one source (config/socialLinks.ts) and the SidebarFooter / EntryFooter comment drift is corrected.

Sweep is clean: no residual references to StatMeter, formatMeterLabel, borrowedRatio, meterBaseUsd, availableMeterPercent, borrowedMeterPercent, or SOCIAL_LINK_URLS.github anywhere under services/ or packages/. Load-bearing data-testids survive.

No critical-path files. LGTM.

PR #2393, opened 2026-09-03
fix(vault): rebuild the Vaults pending-deposit row

### LINKED ISSUE #2341: Vaults page: rebuild the pending deposit card, drop two tooltips, pad the empty state
## Context

The Vaults page at `/vaults`. Screenshot shows the summary bar, one pending deposit mid-flow, and the empty vaults list. Rendered by `services/vault/src/components/pages/VaultsPage.tsx`.

Figma is the source of truth for the pending deposit card: [node 10247-49814](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10247-49814). Read it before changing the layout.

### 1. Remove the total collateral value tooltip

**Now:** an info icon sits next to "Total Collateral Value" in the summary bar.
**Expected:** removed.
**Where:** `services/vault/src/components/vaults/VaultsSummaryCard.tsx`

### 2. Remove the health factor description

**Now:** the summary bar prints "When the ratio falls below 1.0, liquidation may occur." under the Health Factor value.
**Expected:** removed. The Health Factor value and its own info icon stay.
**Where:** `services/vault/src/components/vaults/VaultsSummaryCard.tsx`, string in `services/vault/src/copy.ts`

### 3. Pending deposit card — button on one line

**Now:** the `Broadcast Pre-Pegin` button wraps onto a second line below the amount.
**Expected:** on the same line as the rest of the row, per Figma.
**Where:** `services/vault/src/components/vaults/VaultsLifecycleSections.tsx`, `services/vault/src/components/shared/ListRow.tsx`

### 4. Pending deposit card — remove the pending tooltip

**Now:** an info icon sits next to the orange `Pending` status chip.
**Expected:** removed.

### 5. Pending deposit card — activation time replaces the step counter

**Now:** `Step 5 of 15` prints under the `0.01016299 sBTC` amount.
**Expected:** the step counter is removed and the activation time shows in its place.

### 6. Pending deposit card — sBTC icon

**Now:** an orange circular placeholder avatar sits to the left of the amount.
**Expected:** the sBTC icon.

### 7. Pad the empty state

**Now:** the "Your BTC Vaults will appear here" block sits flush against its container.
**Expected:** padding around the block.
**Where:** `services/vault/src/components/vaults/VaultsEmptyState.tsx`

<details>

> *Vaults Page*
> • remove tooltip on total collateral value
> • remove description of health factor
> • change pending deposit design same as in figma (put button into one line, remove info near pending, remove steps under the amount and put activation time, put icon of sBTC instead of "V"): https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10247-49814
> • add padding around block "Your BTC Vaults will appear here"

</details>

jeremy-babylonlabs:
## Summary

Rebuilds the `/vaults` pending-deposit row to match Figma (node `10247-49814`) and clears the surrounding layout defects on the page. The pending row is now a single line at `xl` and up: coin avatar, amount over an activation estimate, status dot with its progress bar, provider, hash, and the action button flush right.

## Changes

**Pending row is one line.** The lifecycle rows wrapped because their container was `flex-wrap` while the shared `LIST_ROW_*` bases (384 + 144×3 + 168 = 984px, plus four 16px gaps = 1048px) exceeded the card's content box. The pending, active and inactive rows are now `flex-wrap xl:flex-nowrap`; the data columns swap `min-w-[120px]` for `min-w-0` so they can shrink, and the leading column's basis drops from 384px to 240px. `LIST_ROW_ACTION_SLOT_CLASS` stays the fixed, non-shrinking slot, so the columns still line up across Pending / Active / Inactive.

**Activation estimate replaces the step counter.** The sub-line under the amount was `inlineSubtext ?? "Step N of 15"`. It is now `inlineSubtext` when a state sets one (the Figma "Activate within 25h 17m" case), otherwise a machine-paced estimate derived from the polling result: `pendingActivationEstimateMinutes(prePeginConfirmations, requiredPrePeginDepth)` in `btcConfirmationProgress.ts`, rendered as "Activates in ~1.5 h". The estimate is withheld while `requiredPrePeginDepth` is undefined, so a loading protocol-params read never produces a guessed number, and withheld on any non-`pending` display variant — a "Failed" or "Expired" row reaches this section too, and a countdown under those labels would contradict the status beside it. `COPY.deposit.progress.stepPrefix` had no other caller and is deleted.

**sBTC coin icon in the leading cell.** The leading cell showed the vault provider's logo — the same logo the provider column already carries one cell over, so a row rendered it twice. It is now the network's Bitcoin icon (`getNetworkConfigBTC().icon`) on core-ui's 32px `Avatar`. Applied to the pending, inactive and active-vault rows for consistency with Figma; the provider column keeps its `ApplicationLogo`.

**Pending status info icon removed.** The `Hint` + `InfoIcon` beside the pending status label is gone per the design. `peginState.message` is untouched — the inactive row and other surfaces still use it.

**Health-factor caption removed** from the `/vaults` summary card, and `COPY.vaults.summary.healthFactorCaption` with it. The tooltip, value and heart stay. The Loans summary keeps its own `COPY.loans.healthFactorCaption`.

**Empty state padding.** The section placement wrapped `EmptyState` in a bordered card with no padding of its own, so the illustration sat flush against the border. `SECTION_CARD_CLASS` now carries `p-6`; `EmptyState` renders unpadded with `withCard={false}`, so there is no double padding.

The Total Collateral Value card already sets no tooltip on `main` — nothing to remove there.

## Consequences

- Long leading sub-lines now truncate instead of forcing a wrap. The longest is `COPY.pegin.messages.refundMaturing` on an inactive row; the `truncate` class was already on those spans. The comments in `ListRow.tsx` and `copy.ts` that documented the old 384px / no-truncation trade-off are rewritten to describe this.
- The row's bases now total 904px (240 + 144×3 + 168 + four 16px gaps), down from 1048px. With the 242px sidebar, the `mx-10` page gutter and the card's 16px padding, a 1280px viewport gives the card 926px of content — so from `xl` the row holds one line without shrinking any column. Below that the rows wrap exactly as they do today. The basis reduction alone already unwraps the row at the widths where the bug was reported: at a 1364px viewport the card has 1010px of content against the old 1048px requirement.
- A `pending`-variant row that is waiting on the depositor rather than on Bitcoin (WOTS submission, payout signing, "Ready to activate") still shows the estimate. It is a machine-paced lower bound on the remaining wait, not a promise of unattended completion.
- The Activity and Active Loans rows keep `flex-wrap` and are unaffected by the basis changes — they use `LIST_ROW_ACTION_SLOT_CLASS` and `ListRowMetric`, neither of which changed.

## Testing

- `pnpm exec eslint` on every changed file — clean.
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm exec tsc --noEmit -p tsconfig.lib.json` — clean.
- `pnpm exec vitest run --maxWorkers=2 src/components` — 91 files, 862 tests passing. That covers `src/components/vaults`, `btcConfirmationProgress.test.ts`, `steps.test.ts`, `ActivityRowV3.test.tsx`, `ActivityList.test.tsx`, `ActiveLoansList.test.tsx`, `VaultsPage.test.tsx` and `src/components/shared`.
- New unit test covers `pendingActivationEstimateMinutes` for null confirmations, a partial depth, and at/over depth where it floors at the pipeline overhead.

Closes #2341

### DIFF
--- a/services/vault/src/components/shared/ListRow.tsx
+++ b/services/vault/src/components/shared/ListRow.tsx
@@ -19,17 +19,17 @@ import { CARD_SHELL_CLASS } from "@/components/shared/layoutClasses";
  * cell and opening a gap before the trailing action button.
  *
  * The basis is the widest cell's intrinsic content — the status cell's 12px
- * dot and 108px label with their 4px gap. `min-w-0` is what lets the cell
- * shrink below that basis: the lifecycle rows are `xl:flex-nowrap`, so from
- * `xl` up the columns absorb any shortfall rather than breaking the action
- * button onto a second line.
+ * dot and 108px label with their 4px gap. The bases are the lever that fits a
+ * lifecycle row on one line: they sum (with the gaps and the action slot) to
+ * less than the row's width from `xl` up, where the rows are
+ * `xl:flex-nowrap`. Below `xl` the row wraps instead.
  */
 export const LIST_ROW_COLUMN_CLASS = "min-w-0 grow basis-[144px]";
 
 /**
  * The status cell. Same basis as a plain data column, but it takes a double
  * share of the leftover slack, matching the leading cell — the longest status
- * (`COPY.pegin.statusLabel.AWAITING_ACTIVATION_WINDOW`) wraps to two lines
+ * (`COPY.pegin.labels.AWAITING_ACTIVATION_WINDOW`) wraps to two lines
  * otherwise. Grow factors divide only the surplus, so they cost nothing once
  * the row is narrow enough that every column is shrinking instead.
  */
@@ -39,9 +39,9 @@ export const LIST_ROW_STATUS_COLUMN_CLASS = "min-w-0 grow-[2] basis-[144px]";
  * The leading cell — a 32px coin avatar beside a two-line amount / sub-line
  * block. The basis fits the amount and the sub-lines the rows produce in the
  * common case; longer interpolated sub-lines (notably
- * `COPY.vaults.statusHints.refundMaturing`) truncate rather than widen the
- * cell, since from `xl` up the row may not wrap. It takes a double share of
- * any remaining slack.
+ * `COPY.pegin.messages.refundMaturing`) truncate rather than widen the cell,
+ * because widening it would push the bases past the row's width and break the
+ * one-line fit from `xl` up. It takes a double share of any remaining slack.
  */
 export const LIST_ROW_LEADING_COLUMN_CLASS = "min-w-0 grow-[2] basis-[240px]";
 
--- a/services/vault/src/components/shared/ListRow.tsx
+++ b/services/vault/src/components/shared/ListRow.tsx
@@ -19,45 +19,31 @@ import { CARD_SHELL_CLASS } from "@/components/shared/layoutClasses";
  * cell and opening a gap before the trailing action button.
  *
  * The basis is the widest cell's intrinsic content — the status cell's 12px
- * dot, 108px label and 16px info icon with their 4px gaps. Keeping it that
- * tight matters: `flex-wrap` breaks lines on the basis and never shrinks to
- * avoid a break, so every extra pixel here raises the viewport width at which
- * the action button drops onto a second line. 120px is the floor below which
- * wrapping is preferable to squashing.
+ * dot and 108px label with their 4px gap. `min-w-0` is what lets the cell
+ * shrink below that basis: the lifecycle rows are `xl:flex-nowrap`, so from
+ * `xl` up the columns absorb any shortfall rather than breaking the action
+ * button onto a second line.
  */
-export const LIST_ROW_COLUMN_CLASS = "min-w-[120px] grow basis-[144px]";
+export const LIST_ROW_COLUMN_CLASS = "min-w-0 grow basis-[144px]";
 
 /**
- * The status cell. Same basis as a plain data column — so the viewport width at
- * which the action button wraps is unchanged — but it takes a double share of
- * the leftover slack, matching the leading cell.
- *
- * Slack was previously split 2:1:1:1 between the leading, status, provider and
- * hash cells, which left the leading cell roomy while the status label wrapped:
- * the basis above is sized for a 108px label, and the longest status is now
- * `COPY.pegin.statusLabel.AWAITING_ACTIVATION_WINDOW`. Widening the basis would
- * have fixed the wrap at every width but raised the action button's wrap
- * threshold, which the comment above deliberately keeps as low as possible.
- * Grow factors divide only the surplus, so they cost nothing at narrow widths.
+ * The status cell. Same basis as a plain data column, but it takes a double
+ * share of the leftover slack, matching the leading cell — the longest status
+ * (`COPY.pegin.statusLabel.AWAITING_ACTIVATION_WINDOW`) wraps to two lines
+ * otherwise. Grow factors divide only the surplus, so they cost nothing once
+ * the row is narrow enough that every column is shrinking instead.
  */
-export const LIST_ROW_STATUS_COLUMN_CLASS =
-  "min-w-[120px] grow-[2] basis-[144px]";
+export const LIST_ROW_STATUS_COLUMN_CLASS = "min-w-0 grow-[2] basis-[144px]";
 
 /**
- * The leading cell — a logo beside a two-line amount / sub-line block. Its
- * basis fits the longest sub-line the rows produce (the refund-maturity
- * notice from `COPY.vaults.statusHints.refundMaturing`, measured at 343px in
- * the app's 12px face) plus the 32px logo and its 8px gap, so that copy reads
- * in full rather than ellipsing. It takes a double share of any remaining
- * slack.
- *
- * That 343px is one measurement of an interpolated string — the block count,
- * the hours and the block/blocks plural all vary — so a longer combination
- * ellipses again. `copy.ts` points back here for that reason; re-measure both
- * together.
+ * The leading cell — a 32px coin avatar beside a two-line amount / sub-line
+ * block. The basis fits the amount and the sub-lines the rows produce in the
+ * common case; longer interpolated sub-lines (notably
+ * `COPY.vaults.statusHints.refundMaturing`) truncate rather than widen the
+ * cell, since from `xl` up the row may not wrap. It takes a double share of
+ * any remaining slack.
  */
-export const LIST_ROW_LEADING_COLUMN_CLASS =
-  "min-w-[120px] grow-[2] basis-[384px]";
+export const LIST_ROW_LEADING_COLUMN_CLASS = "min-w-0 grow-[2] basis-[240px]";
 
 /**
  * Trailing action cell. Every lifecycle and activity row routes its action
@@ -96,7 +82,8 @@ export const LIST_ROW_MIN_HEIGHT_CLASS = "min-h-[78px]";
 /**
  * Bordered row card shared by the vault, activity and loan lists. `className`
  * carries per-surface layout — the vault lifecycle rows pass
- * `LIST_ROW_MIN_HEIGHT_CLASS`; the single-line lists pass nothing.
+ * `LIST_ROW_MIN_HEIGHT_CLASS` and `xl:flex-nowrap`; the single-line lists pass
+ * nothing and keep the wrapping default.
  */
 export function ListRowCard({
   children,
--- a/services/vault/src/components/simple/DepositProgressView/btcConfirmationProgress.ts
+++ b/services/vault/src/components/simple/DepositProgressView/btcConfirmationProgress.ts
@@ -57,15 +57,16 @@ export function computeTotalEstimateMinutes(requiredDepth: number): number {
  * Estimated minutes until the deposit activates, for the /vaults pending row.
  * Same model as {@link computeTotalEstimateMinutes}, less the depth already
  * confirmed: it floors at the pipeline overhead once the required depth is
- * met, and never goes negative. `null` confirmations (no poll result yet) is
- * treated as zero.
+ * met, and never goes negative. `null` confirmations (no poll result yet)
+ * returns `null` — there is no estimate until the first poll lands.
  */
 export function pendingActivationEstimateMinutes(
   confirmations: number | null,
   requiredDepth: number,
-): number {
+): number | null {
+  if (confirmations === null) return null;
   return (
-    (computeRemainingEstimateMinutes(confirmations ?? 0, requiredDepth) ?? 0) +
+    (computeRemainingEstimateMinutes(confirmations, requiredDepth) ?? 0) +
     DEPOSIT_PIPELINE_OVERHEAD_MINS
   );
 }
--- a/services/vault/src/components/simple/DepositProgressView/btcConfirmationProgress.ts
+++ b/services/vault/src/components/simple/DepositProgressView/btcConfirmationProgress.ts
@@ -52,3 +52,20 @@ export function computeRemainingEstimateMinutes(
 export function computeTotalEstimateMinutes(requiredDepth: number): number {
   return requiredDepth * BTC_BLOCK_TIME_MINS + DEPOSIT_PIPELINE_OVERHEAD_MINS;
 }
+
+/**
+ * Estimated minutes until the deposit activates, for the /vaults pending row.
+ * Same model as {@link computeTotalEstimateMinutes}, less the depth already
+ * confirmed: it floors at the pipeline overhead once the required depth is
+ * met, and never goes negative. `null` confirmations (no poll result yet) is
+ * treated as zero.
+ */
+export function pendingActivationEstimateMinutes(
+  confirmations: number | null,
+  requiredDepth: number,
+): number {
+  return (
+    (computeRemainingEstimateMinutes(confirmations ?? 0, requiredDepth) ?? 0) +
+    DEPOSIT_PIPELINE_OVERHEAD_MINS
+  );
+}
--- a/services/vault/src/components/vaults/VaultsLifecycleSections.tsx
+++ b/services/vault/src/components/vaults/VaultsLifecycleSections.tsx
@@ -59,6 +59,7 @@ import { useReclaimVaultChainData } from "@/hooks/useReclaimVaultChainData";
 import {
   canPerformAction,
   getPeginDisplayStep,
+  hasActionableStep,
   PeginAction,
   type PeginState,
 } from "@/models/peginStateMachine";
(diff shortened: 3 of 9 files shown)


jonybur (dismissed):
Sweep is complete. stepPrefix and vaults.summary.healthFactorCaption drop with no remaining refs (loans.healthFactorCaption is untouched, as described). pendingActivationEstimateMinutes handles the null-confirmations, partial, and at/over-depth cases with an explicit floor at the pipeline overhead, and it is withheld when requiredPrePeginDepth is undefined or the row is not pending. Layout basis changes and xl:flex-nowrap gate look right. LGTM.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -223,12 +233,6 @@ function PendingRow({
             <span className="text-sm leading-[1.43] tracking-[0.17px] text-accent-primary">
```
gbarkhatov, line 246: **Merge-blocker.** Issue #2341 item 4 asks to remove the info icon beside the orange `Pending` chip, but this removes the `Hint` for *every* state the row can render — and two of them have no other surface for their text.

The reasoning recorded in `copy.ts:180-181` and `:196-198` is sound for the states it covers: `activationIncomplete` and `activationWindowTooltip` both keep a visible `*Subtext` form, so losing the tooltip there costs nothing. The gap is the states that set `message` and no `inlineSubtext`:

- `vpTerminalError` -> **FAILED** (`peginStateMachine.ts:593-595`, `displayVariant: "warning"`, `message` set, no subtext)
- `activationDeadlinePassed` -> **EXPIRED** (`:685-687`, same shape)

Both are reachable on this row: `usePendingDeposits.ts:53-58` puts contract status PENDING *and* VERIFIED into `pendingActivities`. And because `activationEstimate` is `null` for any non-`"pending"` variant, the sub-line at line 221 evaluates to `""`.

So a depositor whose vault provider rejected their deposit sees a red dot, the bare word "Failed", and an empty line — with the provider's actual error rendered nowhere in the app.

`InactiveRow` already has the shape of the fix at line 405: keeping the `Hint` gated on `displayVariant !== "pending"` satisfies item 4 (the icon leaves the orange chip) while preserving the reason on the two states that need it.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -152,6 +156,19 @@ function PendingRow({
       : null;
   const fillPercent = step !== null ? getStepFillPercent(step) : null;
 
+  const activationEstimate =
+    peginState?.displayVariant !== "pending" ||
+    result?.requiredPrePeginDepth === undefined
+      ? null
+      : COPY.vaults.pendingActivationEstimate(
+          formatDurationShort(
+            pendingActivationEstimateMinutes(
```
gbarkhatov: **Merge-blocker.** The gate is `displayVariant !== "pending"`, but `"pending"` also covers the states that are blocked on the *user*, so the estimate renders a machine-paced clock on rows where no clock is running.

All four set `displayVariant: "pending"` with no `inlineSubtext`, so the estimate wins the `??` chain: `SUBMIT_WOTS_KEY` (`peginStateMachine.ts:605-610`), `SIGN_AND_BROADCAST_TO_BITCOIN` (`:612-617`), `SIGN_PAYOUT_TRANSACTIONS` (`:619-624`), `READY_TO_ACTIVATE` (`:740-743`).

The broadcast case is the sharp one, and it's the row this PR exists to fix: the Pre-PegIn isn't on the network at all and confirmations are `null`, so the row reads **"Activates in ~70 min"** immediately beside its own `Broadcast Pre-Pegin` button — a confirmation clock that hasn't started and can't start until the user clicks it.

The helper's own docstring states the opposite intent — "User-paced signing steps are excluded" (`btcConfirmationProgress.ts:50`).

Combined with the comment on line 233, on a FAILED row the false countdown can end up as the only text. `actionStatus.type !== "available"` is already computed two lines below at line 172, so gating on that (or on `isActionablePeginAction`) looks like the cheapest fix.

services/vault/src/components/simple/DepositProgressView/btcConfirmationProgress.ts
```diff
@@ -52,3 +52,20 @@ export function computeRemainingEstimateMinutes(
...
+ * confirmed: it floors at the pipeline overhead once the required depth is
+ * met, and never goes negative. `null` confirmations (no poll result yet) is
+ * treated as zero.
+ */
+export function pendingActivationEstimateMinutes(
+  confirmations: number | null,
+  requiredDepth: number,
+): number {
+  return (
+    (computeRemainingEstimateMinutes(confirmations ?? 0, requiredDepth) ?? 0) +
```
gbarkhatov: The docstring discloses this ("`null` confirmations (no poll result yet) is treated as zero"), so it's a deliberate choice rather than an oversight — but it's the choice that makes the comment on `VaultsLifecycleSections.tsx:165` produce a confident wrong number instead of nothing.

`types/peginPolling.ts:31-35` defines `prePeginConfirmations: null` as "the first poll hasn't returned yet (or the deposit isn't in the poll set)", and the sibling field's doc four lines later says consumers "must withhold any depth conclusion rather than assume a default". The call site honours that for `requiredPrePeginDepth` (bailing when `undefined`) and coerces here, which is a bit asymmetric for two fields carrying the same kind of uncertainty. There's also a second unknown-to-default hop stacked on it: `?? 0` on `computeRemainingEstimateMinutes`'s own `null` return.

Returning `null` for unknown confirmations and letting the row skip the slot would compose better — `BtcConfirmationDetailContainer.tsx:42` already branches on the null case rather than assuming zero.

Worth knowing that the new test `it("estimates the full total when no poll result has landed yet")` asserts `pendingActivationEstimateMinutes(null, 6) === 70`, which writes this coercion down as intended behaviour — so it'd need to move with the code.

services/vault/src/components/shared/ListRow.tsx
```diff
@@ -19,45 +19,31 @@ import { CARD_SHELL_CLASS } from "@/components/shared/layoutClasses";
...
- * the leftover slack, matching the leading cell.
- *
- * Slack was previously split 2:1:1:1 between the leading, status, provider and
- * hash cells, which left the leading cell roomy while the status label wrapped:
- * the basis above is sized for a 108px label, and the longest status is now
- * `COPY.pegin.statusLabel.AWAITING_ACTIVATION_WINDOW`. Widening the basis would
- * have fixed the wrap at every width but raised the action button's wrap
- * threshold, which the comment above deliberately keeps as low as possible.
- * Grow factors divide only the surplus, so they cost nothing at narrow widths.
+ * The status cell. Same basis as a plain data column, but it takes a double
```
gbarkhatov, line 30: I don't think `xl:flex-nowrap` and `min-w-0` ever engage — the basis cut alone is what unwraps the row, and the wrap still happens below about 1261px.

Arithmetic from the constants: `PendingRow`'s bases sum to 240+144+144+144+168 plus 4×16px gaps = **904px**. /vaults renders under `PAGE_CONTENT_CLASS` (`!px-0 mx-10`, `layoutClasses.ts:29`) beside core-ui's 242px sidebar, so the row's content box is roughly `viewport − 355`. That means it stops overflowing at ~1259-1261px — *below* the `xl` (1280px) breakpoint where `flex-nowrap` first applies — and at 1280 there's ~925px available against 904px needed, so nothing ever has to shrink.

The practical consequence is that on any window narrower than ~1261px — a 1152 or 1024 laptop, or just a non-maximized window — the `Broadcast Pre-Pegin` button still drops to a second line, so item 3 is unfixed there. The rewritten comments at lines 22-25 and 41-44 describe a shrink mechanism that doesn't operate.

Confidence medium: this is arithmetic from the class constants, not a browser measurement, so worth a quick check at 1200px before acting on it. If it holds, keying the nowrap to the row's own width (a container query, or a `min-[1261px]:` variant accounting for the ~355px of chrome) would fix the real case; otherwise dropping the inert utilities and documenting the basis as the actual lever would at least stop the comment misleading.

services/vault/src/components/shared/ListRow.tsx
```diff
@@ -19,45 +19,31 @@ import { CARD_SHELL_CLASS } from "@/components/shared/layoutClasses";
...
- * That 343px is one measurement of an interpolated string — the block count,
- * the hours and the block/blocks plural all vary — so a longer combination
- * ellipses again. `copy.ts` points back here for that reason; re-measure both
- * together.
+ * The leading cell — a 32px coin avatar beside a two-line amount / sub-line
+ * block. The basis fits the amount and the sub-lines the rows produce in the
+ * common case; longer interpolated sub-lines (notably
+ * `COPY.vaults.statusHints.refundMaturing`) truncate rather than widen the
+ * cell, since from `xl` up the row may not wrap. It takes a double share of
+ * any remaining slack.
```
gbarkhatov: Two things about this comment block, since it's the instruction telling the next maintainer which strings to re-measure before touching these bases.

**The copy keys cited don't exist.** Line 32 cites `COPY.pegin.statusLabel.AWAITING_ACTIVATION_WINDOW`; the real path is `COPY.pegin.labels.AWAITING_ACTIVATION_WINDOW` (`copy.ts:127`). Lines 41-42 cite `COPY.vaults.statusHints.refundMaturing`; the real path is `COPY.pegin.messages.refundMaturing` (`copy.ts:228`). Neither `statusHints` nor `statusLabel` appears anywhere in `copy.ts`. The `statusHints` one was carried forward from the pre-PR comment, but this hunk rewrote the sentence around it, so it's a good moment to fix both.

**The 384 -> 240 cut lands on `InactiveRow` too.** The old comment recorded 384px as sized to fit `refundMaturing` at a measured 343px. At 240px that sub-line — "Your refund will be claimable in ~N Bitcoin blocks (~Nh)", the copy telling a user when they can get their BTC back — truncates at typical widths, and the span is a bare `truncate` with no `title`. `InactiveRow`'s surviving `Hint` carries `message` (the expiry notice), not `inlineSubtext`, so the full text isn't recoverable by any interaction.

The PR body does disclose the truncation trade-off, so this is a judgment call rather than an oversight — but `InactiveRow` didn't need the narrower cell to fix item 3. A second exported constant, or just `title={peginState.inlineSubtext}` on the span, keeps it readable.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -181,34 +198,27 @@ function PendingRow({
...
-              : step !== null
-                ? COPY.deposit.progress.stepPrefix(
-                    getVisualStep(step),
-                    TOTAL_VISUAL_STEPS,
-                  )
-                : ""}
+            {/* Subtext wins when a state sets one: it is state-specific (e.g.
+                an activation-window countdown) and therefore sharper than the
+                block-depth estimate, which only models the common wait. */}
+            {peginState?.inlineSubtext ?? activationEstimate ?? ""}
```
gbarkhatov: Minor: the pre-PR code used a truthiness ternary here and the rebuild switched to `??`, which falls through only on `null`/`undefined` — so an empty-string `inlineSubtext` would be accepted as a real subtext and suppress the estimate.

No state sets `inlineSubtext: ""` today, so this is latent rather than live. But the whole point of the chain is that a state without a usable subtext falls back to the estimate, and `||` preserves that at no cost.

(The trailing `?? ""` is doing real work though — that's what renders the empty line described in the comment on line 233.)

gbarkhatov (dismissed):
Reviewed all four issue items and traced the rebuilt row through `peginStateMachine`, `usePendingDeposits` and the column constants. Items 1, 2 and the avatar/empty-state work are clean, every load-bearing `data-testid` survives with its pointer comment, and the pixel bases are properly named constants with measurement rationale rather than magic numbers.

Two things I'd want resolved before merge, both inline and both about which rows the new sub-line reaches. They compound: the row that gets the wrong countdown is often the same row that lost its explanation.

**On the `Hint` removal specifically — the comments in this diff show the reasoning, and I think it's right for the cases it considered.** The new notes at `copy.ts:180-181` and `:196-198` correctly reason that states like `activationIncomplete` and `activationWindowTooltip` keep a visible `*Subtext` form, so losing the tooltip costs nothing there. The gap is the states that set `message` and *no* `inlineSubtext` — `vpTerminalError` (FAILED) and `activationDeadlinePassed` (EXPIRED). Those aren't covered by that reasoning, and they're the two where the text matters most.

Three more that aren't inline because their lines fall outside a diff hunk:

- **`peginStateMachine.ts:591-745` — orphaned producers.** I grepped every production reader of `peginState.message` under `components/` and `hooks/`: after this PR there is exactly one, `VaultsLifecycleSections.tsx:405-407` in `InactiveRow`, fed only by `expiredActivities` and `reclaimableCandidates` (contract status EXPIRED / DEPOSITOR_WITHDRAWN). So `vpTerminalError`, `awaitingWotsKey`, `broadcastMayHaveFailed`, `payoutsReadyForSigning`, `prePeginIngesting`, `prePeginBroadcast`, `waitingForDetection`, `readyToActivate`, `activationWindowTooltip` and the rest are all still computed and displayed nowhere. Fixing the first inline comment resolves this too.
- **No regression gate on either row change.** The only test added covers the pure helper; there is no `VaultsLifecycleSections` test file at all, and `VaultsPage.test.tsx` asserts nothing about the sub-line, avatar or status cell. One render test for a depositor-blocked pending state would pin the second inline comment directly.
- **The green `visual` check is not evidence here.** Every target renders without a connected wallet, so no pending, inactive or active row appears in any snapshot — and the `p-6` at `VaultsEmptyState.tsx:38` sits on the `sectionPlacement` branch, which only renders beneath a pending deposit, so it's outside the capture too. That workflow also reports into a sticky comment rather than gating; its only `exit 1` sites concern failing to publish images, not a pixel diff.

Checked and dismissed, so nobody re-raises them: `min-w-0` does not risk overflow on Activity or Active Loans — the three changed constants are imported only by the two vault files, and `ListRowMetric` keeps its own `min-w-[120px]`; dropping the `stepPrefix` import orphans nothing, since `getVisualStep` and `TOTAL_VISUAL_STEPS` keep six other consumers and only the copy key is gone, with zero surviving references; item 1 really was a no-op on `main` (`VaultsSummaryCard.tsx:56` sets only `caption`, and the icon renders only when `tooltip` is present); the surviving `healthFactorCaption` hit is the distinct `COPY.loans.*` key still read by `LoansSummary.tsx:84`, while `tooltip: COPY.tooltips.healthFactor` correctly survives at line 76; and no test was deleted in this PR.

Figma was unreachable this session, so the `xl` breakpoint choice, the avatar size and the exact sub-line wording are unverified against node 10247-49814. Nothing above depends on it.

gbarkhatov (approved):

jonybur (approved):
Address-review delta reinstates the pending-row message inline and keeps the status tooltip on non-pending rows, so the states that previously carried peginState.message (vpTerminalError, activationDeadlinePassed and the depositor-blocked pending steps) get their text back. The activation estimate is now suppressed on user-paced steps and while no poll has landed, so a 'ready to activate' or WOTS-resubmission row no longer displays a machine-paced countdown. Truncated refund-message rows expose the full string via title.

Copy-key citations are corrected. Rebuilt row layout, avatar swap, empty-state padding and the pendingActivationEstimateMinutes helper still line up as reviewed.

No critical-path files. LGTM.

PR #2394, opened 2026-09-03
fix(vault): restyle the Activity search input

### LINKED ISSUE #2342: Activity page: restyle the search input
## Context

The search field at the top of the Activity page (`/activity`), placeholder "Search tx hash or type of transaction". Two screenshots are attached: the full page and a close-up of the input.

## Change

**Now:** the input uses `border-secondary-strokeLight` (`#dddddd` in light theme), fill `bg-primary-contrast`, and `placeholder:text-accent-disabled`. Against the page background the border and placeholder both read as washed out.
**Expected:** border, fill and placeholder move to the values in the design file.
**Where:** `services/vault/src/components/Activity/ActivitySearchInput.tsx:24` (container) and `:36` (input and placeholder)

Note the component deliberately avoids the shared `Input` component because that one bakes its own radius, fill and border into a `.bbn-input` rule — see the comment at the top of the file. Keep the styles local.

## Open question

Target colours are not specified. Pull them from Figma or ask design before implementing.

<details>

> *Activity page*
> • change colours of "search" input

</details>

jeremy-babylonlabs:
Restyles the Activity toolbar search box so its border, fill, and placeholder read as a filled field rather than a hairline outline on the page surface.

## Changes

`services/vault/src/components/Activity/ActivitySearchInput.tsx`, two class strings:

| Surface | Before | After | Light | Dark |
| --- | --- | --- | --- | --- |
| Container border | `border-secondary-strokeLight` | `border-secondary-strokeDark` | `#cccccc` | `#5A5A5A` |
| Container fill | `bg-primary-contrast` + `dark:bg-transparent` | `bg-secondary-highlight` | `#F9F9F9` | `#252525` |
| Placeholder | `placeholder:text-accent-disabled` | `placeholder:text-accent-secondary` | `#666666` | `#B0B0B0` |

The `dark:bg-transparent` escape hatch is gone: `secondary-highlight` already resolves per theme, so the dark fill is a token rather than the page surface showing through.

Geometry, spacing, and the search icon are unchanged. The component stays a plain `<label>`/`<input>` — the docblock explaining why it does not use core-ui's `Input` still holds.

## Token values

The vault re-points `--twc-secondary-strokeDark` in `services/vault/src/globals.css` to `0 0% 80%`, annotated there as Figma `stroke/primary #cccccc`; the dark value `#5A5A5A` comes from core-ui's palette. `secondary-highlight` and `accent-secondary` are core-ui palette tokens, unmodified by the vault.

These tokens are a best fit from the existing vault theme, not read off the design file — the Figma page for this issue was not reachable. Flagging for design confirmation before merge; if the intended values differ, the change is two class swaps.

## Accessibility

Placeholder contrast against the new fill improves from 1.96:1 to 5.45:1 in light and from 4.24:1 to 7.07:1 in dark. The container border contrast also improves (1.36:1 to 1.61:1 light, 1.40:1 to 2.72:1 dark) but stays under 3:1 in both themes — a property of the chosen design tokens, not of this change.

The input carries `outline-none` with no focus-visible replacement, so keyboard focus is invisible. That predates this change and is left alone here to keep the diff scoped; worth a follow-up.

## Verification

- `pnpm exec eslint src/components/Activity/ActivitySearchInput.tsx` — clean
- `pnpm exec tsc --noEmit -p tsconfig.lib.json` — clean
- `pnpm exec vitest run src/components/Activity` — 6 files, 60 tests passed

No test asserted on the replaced class names, and the change carries no logic, so no test was added or updated.

Closes #2342

### DIFF
--- a/services/vault/src/components/Activity/ActivitySearchInput.tsx
+++ b/services/vault/src/components/Activity/ActivitySearchInput.tsx
@@ -21,7 +21,7 @@ export function ActivitySearchInput({
   onChange,
 }: ActivitySearchInputProps) {
   return (
-    <label className="flex w-full max-w-[360px] items-center gap-2 rounded-lg border border-secondary-strokeLight bg-primary-contrast py-2 pl-4 pr-3 dark:bg-transparent">
+    <label className="flex w-full max-w-[360px] items-center gap-2 rounded-lg border border-secondary-strokeDark bg-secondary-highlight py-2 pl-4 pr-3">
       <RiSearchLine
         size={SEARCH_ICON_SIZE}
         aria-hidden
@@ -33,7 +33,7 @@ export function ActivitySearchInput({
         aria-label={COPY.activity.searchLabel}
         placeholder={COPY.activity.searchPlaceholder}
         onChange={(event) => onChange(event.target.value)}
-        className="w-full bg-transparent text-sm leading-[1.43] tracking-[0.17px] text-accent-primary outline-none placeholder:text-accent-disabled"
+        className="w-full bg-transparent text-sm leading-[1.43] tracking-[0.17px] text-accent-primary outline-none placeholder:text-accent-secondary"
       />
     </label>
   );
--- a/services/vault/src/components/Activity/ActivitySearchInput.tsx
+++ b/services/vault/src/components/Activity/ActivitySearchInput.tsx
@@ -21,7 +21,7 @@ export function ActivitySearchInput({
   onChange,
 }: ActivitySearchInputProps) {
   return (
-    <label className="flex w-full max-w-[360px] items-center gap-2 rounded-lg border border-secondary-strokeDark bg-secondary-highlight py-2 pl-4 pr-3">
+    <label className="flex w-full max-w-[360px] items-center gap-2 rounded-lg border border-secondary-strokeDark bg-secondary-highlight py-2 pl-4 pr-3 focus-within:border-accent-primary">
       <RiSearchLine
         size={SEARCH_ICON_SIZE}
         aria-hidden
--- a/services/vault/src/components/Activity/ActivitySearchInput.tsx
+++ b/services/vault/src/components/Activity/ActivitySearchInput.tsx
@@ -21,7 +21,7 @@ export function ActivitySearchInput({
   onChange,
 }: ActivitySearchInputProps) {
   return (
-    <label className="flex w-full max-w-[360px] items-center gap-2 rounded-lg border border-secondary-strokeDark bg-secondary-highlight py-2 pl-4 pr-3 focus-within:border-accent-primary">
+    <label className="flex w-full max-w-[360px] items-center gap-2 rounded-lg border border-secondary-strokeDark bg-neutral-200 py-2 pl-4 pr-3 focus-within:border-accent-primary">
       <RiSearchLine
         size={SEARCH_ICON_SIZE}
         aria-hidden


jonybur (dismissed):
Style-only change. New tokens resolve per theme and remove the dark-mode escape hatch as described. Placeholder contrast improvement is real. Clean.

services/vault/src/components/Activity/ActivitySearchInput.tsx
```diff
@@ -21,7 +21,7 @@ export function ActivitySearchInput({
   onChange,
 }: ActivitySearchInputProps) {
   return (
-    <label className="flex w-full max-w-[360px] items-center gap-2 rounded-lg border border-secondary-strokeLight bg-primary-contrast py-2 pl-4 pr-3 dark:bg-transparent">
+    <label className="flex w-full max-w-[360px] items-center gap-2 rounded-lg border border-secondary-strokeDark bg-secondary-highlight py-2 pl-4 pr-3">
```
gbarkhatov: The dark half of this works well; the light half may not reach the goal, and the two themes now diverge in different directions.

**Light:** the aim is that the box "read as a filled field rather than a hairline outline on the page surface," but `secondary-highlight` is `#F9F9F9` against the `bg-surface` page of `#ffffff` (`RootLayout.tsx:188`) — 1.05:1, no perceptible fill. The `primary-contrast` `#F5F7F2` it replaces was 1.08:1, so the fill is fractionally flatter than before and the boundary still rests entirely on the border. That border does improve, `#dddddd` -> `#cccccc`, 1.36:1 -> 1.61:1, though it stays under WCAG 1.4.11's 3:1 for a control boundary.

**Dark:** here the fill genuinely lands (`#252525` on `#121212`), but it no longer matches the row cards immediately below it. Those go through `ListRowCard` (`ListRow.tsx:117`) to `CARD_SHELL_CLASS` (`layoutClasses.ts:50`), which deliberately overrides this token's dark value to `dark:bg-[#202020]` — and their border is `strokeLight` `#2F2F2F` against this input's `strokeDark` `#5A5A5A`. In light the two agree almost exactly; in dark they split on both fill and border, so the toolbar-to-list relationship becomes theme-dependent.

If the lift off the list is intended, it'd be worth having it in both themes; if not, `dark:bg-[#202020]` matches the shell. Either way this looks like one to settle in the same design pass you've already flagged.

services/vault/src/components/Activity/ActivitySearchInput.tsx
```diff
@@ -33,7 +33,7 @@ export function ActivitySearchInput({
         aria-label={COPY.activity.searchLabel}
         placeholder={COPY.activity.searchPlaceholder}
         onChange={(event) => onChange(event.target.value)}
-        className="w-full bg-transparent text-sm leading-[1.43] tracking-[0.17px] text-accent-primary outline-none placeholder:text-accent-disabled"
+        className="w-full bg-transparent text-sm leading-[1.43] tracking-[0.17px] text-accent-primary outline-none placeholder:text-accent-secondary"
```
gbarkhatov, line 36: Not introduced here — `outline-none` is byte-identical before and after, and every contrast axis improves — but this PR is the restyle of the exact border that would carry a focus state, so it may be the cheapest moment to fix it.

As it stands the input has `outline-none` with no replacement, neither the `<label>` nor the `<input>` defines `focus-within:`/`focus:`, and there's no global ring in `globals.css` or core-ui's `index.css`. Keyboard focus on this control is invisible (WCAG 2.4.7).

core-ui's own filled input already solves this with a pairing that would drop straight in: `Form/Input.css:2` sits at `border-secondary-strokeDark/50` at rest and `:4-6` promotes to full `border-secondary-strokeDark` on `:focus-within`. That's a token/opacity alignment rather than adopting `.bbn-input`, so it stays inside #2342's "keep the styles local" — and the `/50` resting border would be the one thing here that softens the light-mode boundary rather than sharpening it, so worth weighing against the point on line 24.

gbarkhatov, line 36: Minor, and quite possibly what design wants: the placeholder now renders in exactly the same colour as the decorative search icon on line 28, which is already `text-accent-secondary`. In core-ui's rule that token is the *entered value* colour (`Form/Input.css:2`), with the placeholder four steps lighter, so this inverts the house placeholder/value distinction — in dark the placeholder lands at 7.07:1, heavier than a lot of real text, and an empty field reads a little like a filled one.

Worth noting the obvious remedy is a trap: `placeholder-accent-disabled/50` would follow core-ui's idiom but `accent-disabled` is `#9ab7c2`, which at half alpha lands near 1.4:1 and re-breaks the contrast failure this PR is fixing. If the flatness matters, a mid-step like `accent-secondary/70` keeps the fix. Flagging for the design pass rather than asserting it's wrong.

gbarkhatov (dismissed):
Reviewed the restyle in both themes, resolving every token through core-ui's palette and the vault's `globals.css` re-points, and recomputing the contrast ratios rather than taking the PR body's on trust — they match.

The placeholder change is a real accessibility fix, not just a swap: `accent-disabled` -> `accent-secondary` takes light from 1.96:1 to 5.45:1 and dark from 4.24:1 to 7.07:1, so light goes from failing even 3:1 to clearing 4.5:1. Dropping `dark:bg-transparent` is right too — dark previously had no fill at all behind a near-invisible `#2F2F2F` border, and it now reads as a proper filled control.

**One thing worth knowing before the design confirmation you asked for: CI cannot discharge it, and the green `visual` job is not evidence here.** `services/vault/e2e/visual/targets.ts` states in its header that every target must render without a connected wallet (page-side provider construction is #1592), while `ActivityList.tsx:139` gates `<ActivitySearchInput>` behind `isConnected &&`. The `activity` target therefore photographs a toolbar that never contains this control — zero pixels of this diff are under snapshot. The design sign-off is the only check that covers it.

Three suggestions inline, all presentational and none blocking. Every colour-vs-design claim is unverified, since the Figma file was unreachable.

Also checked and found clean: no critical path touched, copy still sourced from `COPY`, no dead code or magic numbers added, and the existing tests select by `role="searchbox"` + accessible name so nothing here disturbs them. A few things I looked at and decided against raising: the mismatch with the filter button's fill is pre-existing and this diff slightly improves it in dark; `placeholder-accent-disabled/50` would follow core-ui's placeholder idiom but re-break the exact contrast failure this PR fixes; and the light-mode token override is unconditional in practice, so the core-ui teal never surfaces.

gbarkhatov (dismissed):

jonybur (dismissed):
Address-review delta adds focus-within:border-accent-primary, which resolves the outline-none focus-invisibility flagged on the earlier pass without adding a new focus ring. Contrast improvements from the token swap hold; no logic added. LGTM.

gbarkhatov (approved):
Follow-up pass over my earlier review (`5111113735`, at `4d648dd4`), not a second opinion on the same code. Delta since then: one commit, one file, `+1/-1` — *"give the Activity search input a perceptible fill"*.

|---|---|---|---|
| 1 | Light fill imperceptible (1.05:1); dark diverges from the row cards below | **partially fixed** | `ActivitySearchInput.tsx:24` |
| 2 | `outline-none` with no replacement — keyboard focus invisible (WCAG 2.4.7) | **fixed** | `ActivitySearchInput.tsx:24` |
| 3 | Placeholder is the same colour as the decorative search icon | **unaddressed** | `ActivitySearchInput.tsx:36` |

The existing threads stand as written. No new defect in the code — the changed line is clean, and the one thing worth raising is in the description rather than the diff.

### What I checked, because the new token looked like a problem and isn't

`bg-neutral-200` reads like a raw Tailwind palette class, which would have meant one fixed value in both themes and a near-white box on the dark page. It isn't. `neutral` is defined inside core-ui's `createThemes` call — the same mechanism behind `secondary-highlight` and `accent-primary` — at `tailwind.config.js:222-225` (light `#F2F2F2`) and `:274-277` (dark `#2C2C2C`), and the vault's `globals.css` does not re-point it. It is also the established idiom here, including `FilterDropdown.tsx:62`, the sibling control in this same toolbar. So this is a semantic token, and there is no dark-mode regression.

**Finding 1, partially fixed:** light moves `#F9F9F9` to `#F2F2F2` (1.05:1 to 1.12:1 against the page) and dark `#252525` to `#2C2C2C` (1.22:1 to 1.34:1). The divergence half of that comment is resolved — the input now lifts off `CARD_SHELL_CLASS` in both themes rather than only one, and in dark it sits three points from the filter trigger's `#2F2F2F`, so the toolbar mismatch is measurably smaller than before. What remains is that the light fill is still around 1.1:1, so the boundary still rests on the border.

**Finding 2, fixed:** `focus-within:border-accent-primary` gives the border `#cccccc` to `#000000` in light (13.08:1 between states) and `#5A5A5A` to `#ffffff` in dark (6.90:1), both far above the 3:1 state-change bar, with no width change so nothing reflows.

**Finding 3, unaddressed** and untouched by the delta. Against the new fill the placeholder measures 5.13:1 light and 6.44:1 dark, still comfortably AA, so its "flagging for the design pass" framing stands unchanged.

### One thing worth fixing before merge

**The description no longer matches its own diff, on four points.** The Changes table lists the container fill as `bg-secondary-highlight` `#F9F9F9`/`#252525`; the shipped class is `bg-neutral-200` `#F2F2F2`/`#2C2C2C`. The sentence crediting `secondary-highlight` with "already resolving per theme" names a token that is no longer in the diff — the claim stays true of `neutral-200`, only the name is wrong. The Accessibility figures were computed against the old fill: placeholder is now **5.13:1** light (not 5.45) and **6.44:1** dark (not 7.07), and the dark border figure is now **2.02:1** against the fill, where 2.72 was border-against-page from when the fill was transparent. And the paragraph stating that focus is *"left alone here to keep the diff scoped"* is contradicted by the `focus-within:` class now on line 24.

Every stated *conclusion* survives — the placeholder clears AA in both themes, the border stays under 3:1 — so this is bookkeeping, not a correctness problem. It matters mainly because this PR's own description is the record for a change that no automated check covers, per the note below.

### CI

`verify` passed (9m28s), so the type-check and unit suite are genuinely green on this head. `validate` fails at "Validate commit signatures" — the branch needs a signed commit before merge, which is not a code defect.

Carrying forward from the prior review, because it applies more now that the delta is a pure colour change: the green `visual` job is not evidence for this diff. `e2e/visual/targets.ts` requires every target to render without a connected wallet, while `ActivityList.tsx` gates `<ActivitySearchInput>` behind `isConnected &&`, so the `activity` target photographs a toolbar that never contains this control. Zero pixels of this change are under snapshot, and the design confirmation you flagged is still the only check that covers it.

PR #2395, opened 2026-09-03
fix(vault): polish the Deposit modal

### LINKED ISSUE #2343: Deposit modal: centre the split checkmark, drop the balance info icon, review fee naming
## Context

The Deposit modal, opened from the `Deposit` button on Overview or Vaults. Two screenshots: the amount and split-selection section, and the fee breakdown plus Protocol Parameters section below the CTA. Both dark theme. Rendered by `services/vault/src/components/simple/SimpleDeposit.tsx`.

### 1. Centre the selected-option checkmark

**Now:** the split options render with `items-start`, so the checkmark on the selected "Do not split" row pins to the top of the block while the row's text runs two lines below it.
**Expected:** the checkmark is vertically centred in the block.
**Where:** `services/vault/src/components/simple/UtxoSplitSelectorV3.tsx:125` and `:177` — `items-start` → `items-center`

### 2. Remove the balance info icon

**Now:** an info icon sits between `Balance: 0.02496049 sBTC` and the `Max` button.
**Expected:** removed.
**Where:** `services/vault/src/components/simple/DepositForm.tsx`

### 3. Review fee and parameter naming and tooltips

**Now:** the breakdown below the CTA reads `Reserve Claimer UTXO`, `Deposit Fee`, `VP commission (0.1%)`, `Net payout`; Protocol Parameters reads `Min deposit`, `Effective minimum for split`, `LTV / Collateral Factor`, `Liquidation threshold (THF)`, `Liquidation Bonus (LB)`. Every one carries an info tooltip.
**Expected:** every label and tooltip reviewed for accuracy and plain language.
**Where:** `services/vault/src/components/simple/DepositFeesBreakdown.tsx`, `FeesSection.tsx`; strings in `services/vault/src/copy.ts`

## Note

Item 3 overlaps two earlier reports in the same channel: one on 2026-07-30 asking what "transaction reserve", "deposit fee" and "net payout" mean, and one on 2026-08-07 leaving tooltip copy against the baby glossary in the main Figma file. Reconcile against that glossary rather than writing new strings.

<details>

> *Deposit modal*
> • align the checkmark in the middle of the block, not at the top
> • remove info near the "max" button

> *Deposit modal*
> • check all naming and tooltips here

</details>

jeremy-babylonlabs:
## Summary

Three polish items on the Deposit modal: centre the split-option checkmark, drop the info icon next to the balance, and reconcile the fee-breakdown and Protocol Parameters labels and tooltips against the ratified `baby-glossary` terminology.

### 1. Centre the selected-option checkmark

Both split-option rows in `UtxoSplitSelectorV3.tsx` used `items-start`, so the checkmark pinned to the top of a row whose text wraps to two lines. Now `items-center`, so it sits vertically centred in the block.

### 2. Remove the balance info icon

The icon between `Balance: … sBTC` and `Max` was the `AmountSlider` `rightField.tooltip`. Removed the `tooltip` entry, the `maxTooltip` const in `DepositForm.tsx`, and the now-orphaned `COPY.deposit.form.maxTooltip` builder. The separate `pendingConfirmationTooltip` icon in the left field is untouched.

Tradeoff worth a second look: the value beside `Balance` is the depositable maximum, not the raw wallet balance — it already nets off the fee buffer, inscription UTXOs and (when set) the supply cap. That tooltip was the only place the deduction was explained, so it is now unexplained. Removing it is the requested change; if the deduction should still be surfaced, relabelling the field is the better fix than putting the icon back.

### 3. Fee and parameter naming against the glossary

Reconciled against `babylonlabs-io/baby-glossary` `glossary/tbv.yaml` rather than writing new copy. Only ratified terms were applied; nothing was renamed on judgement alone.

**Changed**

| Label / tooltip | Was | Now | Ratified entry |
| --- | --- | --- | --- |
| `transactionReserveLabel` | Reserve Claimer UTXO | Depositor Claim Output | `docs.depositor_claim_output` — "Depositor claim output". The old label also borrowed "claimer", which the glossary reserves for the Vault Keeper executing a claim, not the depositor |
| `vpCommissionLabel` | VP commission | Vault Provider commission | `docs.vault_provider_commission` — "Vault Provider commission" |
| `networkFeeRateTooltip`, `btcNetworkFeeTooltip` | pre-pegin funding transaction | Pre-Pegin funding transaction | `operation.pegin` — "Pre-Pegin" is the protected `ui_copy` proper-noun form; lowercase is casing drift |
| `protocolFees.minForSplit.tooltip` | partial liquidation | partial-position liquidation | `docs.partial_position_liquidation` |
| `protocolFees.liquidationThreshold.tooltip` | The ideal health factor to restore during liquidation | The health factor the protocol aims to restore after partial-position liquidation. | `docs.target_health_factor` — the ratified definition says *after* partial-position liquidation, not *during* liquidation |

**Left unchanged**

| Label | Reason |
| --- | --- |
| `depositFeeLabel` / `depositFeeTooltip` | No ratified glossary entry for a protocol deposit fee |
| `netPayoutLabel` / `netPayoutTooltip` | No ratified glossary entry for net payout |
| `networkFeeRateLabel`, `btcNetworkFeeLabel` | No ratified glossary entry; only their tooltips carried drift |
| `protocolFees.minDeposit` | No ratified glossary entry |
| `protocolFees.minForSplit.label` | No ratified glossary entry; only the tooltip carried drift |
| `protocolFees.ltv.label` (Collateral Factor) | Term already canonical. Two sibling sites outside this scope also read "Collateral Factor", so re-casing only this one would introduce drift |
| `protocolFees.liquidationThreshold.label` (Target Health Factor) | Term already canonical |
| `protocolFees.maxLiquidationPenalty` label and tooltip | Two authorities disagree — see "Needs a decision" below. Left as shipped |
| `transactionReserveTooltip` | The ratified definition does not contradict it, and its wording is deliberately paired with the reclaim review copy |
| `vpCommissionTooltip` | Accurate, and "vault provider" lowercase mid-sentence is an allowed `ui_copy` variant |
| `withdraw.review.vpCommissionLabel` (VP Commission) | Deliberate divergence from `deposit.form.vpCommissionLabel`. The Review Withdraw rows are pinned Title Case to Figma 10088-38704 and to the sibling Review Refund card, so re-casing this one row to the glossary headword would break that card's internal consistency |

**Casing rule applied.** Terms were changed; each label kept its existing casing convention. The five glossary entries in play (`collateral_factor`, `target_health_factor`, `liquidation_bonus`, `depositor_claim_output`, `vault_provider_commission`) declare no `surfaces` and carry a `docs_term`, so their sentence case is an imported prose headword, not a label ruling. The two entries that do declare `surfaces: [ui_copy, docs]` are Title Case. Protocol Parameters is uniformly Title Case across its five rows, so re-casing to prose form would have introduced drift rather than removed it.

## Needs a decision: Max Liquidation Penalty

Two authorities disagree on this one label, so it is **left exactly as shipped** and is not part of this change.

- **Design review #2322** (commit 96a060ca) deliberately renamed "Liquidation Bonus (LB)" to "Max Liquidation Penalty" as part of the v3 product copy pass.
- **The glossary** ratifies `docs.liquidation_bonus` — "Liquidation bonus", owned by @marketing — and `useProtocolFeeRows` computes the displayed figure as `bonusPercent = (LB - 1) × 100`.

The glossary entry predates the design review, and the two names take opposite viewpoints: a bonus to the liquidator is a penalty to the depositor. Overturning a design-review string is a product call rather than a terminology sweep, so it needs an owner's decision rather than a default either way.

One detail for whoever picks this up: the ratified definition words the parameter as a "discount on seized collateral". That is numerically loose for a `(LB - 1) × 100` display — a 1.05 bonus is 5% extra collateral but only a ~4.76% price discount — so "extra collateral" would be the more accurate phrasing if the glossary name wins.

## Follow-ups

- The glossary has no ratified `ui_copy` form for collateral factor, health factor, target health factor, liquidation bonus or depositor claim output. Adding `surfaces: [ui_copy]` with Title Case `allowed_variants` upstream would settle the label-casing question properly.
- The fee-breakdown list is mixed case ("Network Fee Rate" beside "Net payout"). Normalising it is copy authoring, so it was left alone.
- `splitOptionDescription` still reads "partial liquidation". It sits outside the reconciled keys.

## Testing

- `pnpm exec eslint` on all four changed files — clean
- `pnpm exec tsc --noEmit -p tsconfig.lib.json` — clean
- `pnpm exec vitest run` over `UtxoSplitSelectorV3`, `DepositForm.suggestedAmount`, `DepositFeesBreakdown`, `FeesSection` — 24 passed
- `grep -rn "maxTooltip" services/vault/src` — no matches

Closes #2343


jonybur (dismissed):
Removed hasUnconfirmedBalanceOnly and maxTooltip cleanly (source, tests, consumers, and copy). Terminology reconciliations look correct against the glossary entries called out in the description. The 'Balance' field losing its deduction tooltip is already raised inline and the description flags the tradeoff. LGTM.

services/vault/src/hooks/deposit/useDepositPageForm.ts
```diff
@@ -346,19 +341,12 @@ export function useDepositPageForm(): UseDepositPageFormResult {
     availableUTXOs,
     spendableMempoolUTXOs,
     ordinalsCheckPending,
-    confirmedBalance,
     unconfirmedBalance,
```
gbarkhatov, line 344: **Merge-blocker (zero dead code).** This destructure was the sole production reader of `confirmedBalance`, so dropping `hasUnconfirmedBalanceOnly` orphans the producer too.

I grepped `services/vault/src` and `packages/` at the head SHA — after this PR the only survivors of `confirmedBalance` are its own `useMemo` at `useUTXOs.ts:74`, the return field at `:188`, assertions at `useUTXOs.test.ts:235,261,286` (including the whole test at `:264` dedicated to it), and now-dead mock setup at `useDepositPageForm.test.tsx:205,479`. Nothing reads it. The hook's comment still says it exists "solely to decide whether a zero spendable balance is genuinely due to having no confirmed funds" — describing a caller that no longer exists.

Note `unconfirmedBalance` is unaffected and still has live readers; this is only about `confirmedBalance`. Worth flagging that the fix means touching `useUTXOs.ts`, which is outside this PR's changed-file list — so the change isn't self-contained as it stands, and nothing fails to compile to tell you.

services/vault/src/copy.ts
```diff
@@ -684,21 +684,16 @@ export const COPY = {
...
       // COPY.reclaim.review.description so the promise made at deposit time
       // and the action offered after settlement read as the same thing. Until
       // the reclaim flow shipped this said the reserve "is returned to you if
       // unused", which nothing in the app could actually do.
       transactionReserveTooltip:
         "A small portion of your deposit is reserved in a dedicated output to fund a future protocol claim transaction. If it goes unused, you can reclaim it from your BTCVault once the vault has settled.",
-      // The depositable maximum is labelled as the balance, with `maxTooltip`
-      // explaining the fee buffer / cap adjustments.
+      // The depositable maximum is labelled as the balance.
       balanceLabel: "Balance",
```
gbarkhatov: **Merge-blocker.** Agreeing with the tradeoff you raised in the PR body, and suggesting the fix you proposed there.

`balanceLabel` pairs with `maxDepositLabel` in the `AmountSlider` `rightField` (`DepositForm.tsx:408-411`), whose value is `maxDepositSats` — wallet balance minus the Pre-Pegin fee buffer, minus inscription UTXOs, clamped to the supply cap. The removed `maxTooltip` was the only surface that stated any of that. The surviving `pendingConfirmationTooltip` covers unconfirmed funds, a different case, and arguably reinforces the wrong model by implying confirmation is the only thing held back.

Concretely: a depositor holding 1 BTC where some UTXOs carry inscriptions sees "Balance: 0.4 BTC" with nothing on screen explaining the difference. `COPY.loans.availableLabel` (`copy.ts:1469`) is already "Available" and is already used for this kind of computed max at `Borrow/index.tsx:331`, so relabelling is a one-line change in a file you already have open, and it keeps the icon removed.

Small thing on the same line: the rewritten comment uses "labelled", which `style_rules.spelling.american` lists in `forbidden_examples`. It's a code comment and the wording is inherited from the line it replaced, so the rule arguably doesn't bind — but this line was rewritten as part of a terminology conformance pass.

services/vault/src/copy.ts
```diff
@@ -684,21 +684,16 @@ export const COPY = {
     },
     form: {
       computingAllocation: "Computing allocation...",
-      transactionReserveLabel: "Reserve Claimer UTXO",
+      transactionReserveLabel: "Depositor Claim Output",
```
gbarkhatov: This one doesn't match its ratified canonical, and its own sibling in this commit takes the opposite treatment.

`docs.depositor_claim_output` (tbv.yaml:297-302) is `canonical: Depositor claim output` / `docs_term: Depositor claim output` — sentence case, no `surfaces` key, no `allowed_variants`. This ships `"Depositor Claim Output"`. Meanwhile `docs.vault_provider_commission` (tbv.yaml:404-409), which has the identical shape — also no `surfaces`, also with a `docs_term`, also ratified — is rendered `"Vault Provider commission"` at line 737, matching its canonical verbatim. Two entries of the same shape, opposite treatment, one commit, rendering as adjacent rows in the same breakdown.

The old label's "claimer" really did need to go, so the substance of the rename is right — it's only the casing. Sentence case would match the canonical, its own neighbour `netPayoutLabel: "Net payout"`, and the other entry changed here. Your call given the Protocol Parameters consistency argument, but I don't think the glossary supports the Title Case reading (see the review body on `actor.depositor`).

services/vault/src/copy.ts
```diff
@@ -730,15 +725,16 @@ export const COPY = {
...
       // Fee-breakdown lines (DepositFeesBreakdown) shown before the user
-      // submits. The commission label appends the percent, e.g. "VP commission
-      // (2.50%)"; net payout is the deposit minus that commission.
+      // submits. The commission label appends the percent, e.g. "Vault
+      // Provider commission (2.50%)"; net payout is the deposit minus that
+      // commission.
       networkFeeRateLabel: "Network Fee Rate",
       networkFeeRateTooltip:
-        "Bitcoin network fee rate for your pre-pegin funding transaction. Raise it during congestion so the transaction confirms sooner.",
+        "Bitcoin network fee rate for your Pre-Pegin funding transaction. Raise it during congestion so the transaction confirms sooner.",
```
gbarkhatov, line 735: This fix is correct — `operation.pegin` lists `Pre-Pegin` as a protected `ui_copy` variant (tbv.yaml:33), so lowercase really was drift. Worth noting the sweep left behind the one spelling the glossary explicitly **forbids**, though.

`copy.ts:961` (`deposit.errors.participantKeyDrift.body`) ships "Your **Pre-PegIn** was not broadcast and no Bitcoin was spent." `Pre-PegIn` is in `forbidden_variants` for that entry, and this is user-visible error copy, not a comment. It also breaks the `copy.ts` header's own "Pre-Pegin (proper-noun form)" rule.

The other three hits (`:482`, `:875`, `:928`) are code comments, so this is the only shipped one — a one-character fix in a file already open.

services/vault/src/copy.ts
```diff
@@ -2057,14 +2053,15 @@ export const COPY = {
     minForSplit: {
       label: "Effective Minimum for Split",
       tooltip:
-        "The minimum amount of BTC required to enable partial liquidation via splitting a deposit into two BTCVaults.",
+        "The minimum amount of BTC required to enable partial-position liquidation via splitting a deposit into two BTCVaults.",
```
gbarkhatov, line 2058: This tooltip is now canonical (`docs.partial_position_liquidation` — "Partial-position liquidation", tbv.yaml:349-353), but the sweep missed four other occurrences, and one of them renders inside the very modal this PR is polishing.

`copy.ts:783` — `splitOptionDescription`, "enable partial liquidation" — renders at `UtxoSplitSelectorV3.tsx:142`, a file this PR edits. The other three are `:2145`, `:2153` and `:2162`.

The PR body defers `:783` as "outside the reconciled keys", but that boundary doesn't quite hold: `transactionReserveLabel`, `networkFeeRateTooltip` and `vpCommissionLabel` are all `deposit.form` keys this PR did reconcile. Fixing `:783` at least would stop the modal showing both spellings at once.

services/vault/src/copy.ts
```diff
@@ -730,15 +725,16 @@ export const COPY = {
...
       networkFeeRateLabel: "Network Fee Rate",
       networkFeeRateTooltip:
-        "Bitcoin network fee rate for your pre-pegin funding transaction. Raise it during congestion so the transaction confirms sooner.",
+        "Bitcoin network fee rate for your Pre-Pegin funding transaction. Raise it during congestion so the transaction confirms sooner.",
       btcNetworkFeeLabel: "BTC Network Fee",
       btcNetworkFeeTooltip:
-        "Estimated Bitcoin miner fee for the pre-pegin funding transaction at the selected fee rate.",
-      vpCommissionLabel: "VP commission",
+        "Estimated Bitcoin miner fee for the Pre-Pegin funding transaction at the selected fee rate.",
+      vpCommissionLabel: "Vault Provider commission",
```
gbarkhatov, line 739: This matches `docs.vault_provider_commission` canonical exactly. The side effect is that the withdraw screen now drifts from it: `withdraw.review.vpCommissionLabel` at `copy.ts:1379` is still `"VP Commission"`, rendered at `WithdrawReviewContent.tsx:112,117` and pinned by `WithdrawReviewContent.test.tsx:52`.

Same concept, two names, and this change is what made them diverge — which is the same failure mode the PR's own methodology avoided elsewhere (Collateral Factor was left alone precisely because two sibling sites would drift).

Possibly deliberate: `copy.ts:1370-1372` pins those row labels to a Figma node, so it may be a design string that shouldn't move. Either way it seems worth a row in the "Left unchanged" table, since right now the table doesn't mention the key at all.

services/vault/src/hooks/deposit/__tests__/useDepositPageForm.test.tsx
```diff
@@ -554,59 +554,6 @@ describe("useDepositPageForm", () => {
       expect(result.current.btcBalanceFormatted).toBe(0.008);
```
gbarkhatov, line 582: Two of the three removed cases were genuinely flag-only and deleting them is right. The third was carrying something else.

It asserted `expect(result.current.btcBalance).toBe(0n)` against `availableUTXOs: []` with `confirmedBalance: 300_000n` — pinning that `btcBalance` derives from the inscription-filtered UTXO list rather than the confirmed total, which is live behaviour at `useDepositPageForm.ts:294-296`.

The surviving `"should calculate BTC balance from UTXOs"` can't stand in for it: its fixture has `availableUTXOs` summing to 500000 + 300000 = 800000 and `confirmedBalance` also `800000n`, so the two candidate sources are indistinguishable. A regression that read the confirmed total instead would pass that test green while offering the user a max they can't fund. Re-adding it is a two-line case without the flag.

gbarkhatov (dismissed):
Reviewed all three items. I pulled `glossary/tbv.yaml` from `baby-glossary` and checked every terminology claim against the ratified entries rather than against the PR description, and grepped the worktree at the head SHA for the deletion side.

The checkmark centring is exactly right, and the terminology sweep is the good kind — ratified terms only, with an explicit "left unchanged" table and a decision deliberately escalated rather than guessed. Two things I'd want a second look at before merge, both flagged inline:

1. **Removing the tooltip leaves `Balance:` labelling the fee-and-reserve-adjusted maximum with nothing explaining the gap.** You called this out yourself in the PR body, and I think your own instinct is the right one — relabelling beats putting the icon back, and `COPY.loans.availableLabel` ("Available") already exists as precedent for exactly this computed max.
2. **`hasUnconfirmedBalanceOnly` was the last reader of `confirmedBalance`**, which is now orphaned in `useUTXOs.ts` — a file outside this PR's changed set, so the change isn't self-contained as shipped.

On the casing rule in the PR body: I don't think it holds. It reasons that entries declaring `surfaces` are Title Case, but `actor.depositor` declares `surfaces: [docs, ui_copy, code_identifier]` and is canonically lowercase `depositor` (tbv.yaml:87-94). That doesn't make the Protocol Parameters consistency argument wrong — it just means the glossary doesn't supply the rule, so it rests on the in-screen consistency argument alone.

Also checked and found clean, so no need to re-litigate: no critical path is touched; `maxTooltip` and `hasUnconfirmedBalanceOnly` removal is otherwise complete with zero stragglers across `services/` and `packages/`; `effectiveRemaining` survives and still feeds `getDepositCtaState`; no `data-testid` moved and neither `e2e/real/` nor `e2e/visual/` references any renamed string; the `items-start` -> `items-center` change is applied to both rows with handlers, `tabIndex` and `aria-disabled` untouched; and the Target Health Factor rewrite matches the ratified definition verbatim.

One CI caveat worth recording: the green `visual` job doesn't cover the checkmark fix. Its workflow reports into a sticky comment rather than gating, and at the `deposit-split-options` stop the spec fills an amount below the split minimum, so the wrapping two-line row this PR fixes is never photographed.

One more that isn't inline because the line falls outside a diff hunk: `DepositFeesBreakdown.test.tsx:99-106` collects `[data-tooltip-content]` nodes and asserts only `toHaveLength(6)` plus set-uniqueness, never a string — so both reworded tooltips ship with no assertion, and the bare `6` is a magic number. Pre-existing, but this is the change it failed to catch, and the contrast is sharp given `vpCommissionLabel` did get exact-string gates.

gbarkhatov (approved):

jonybur (approved):
Address-review delta removes the unread confirmedBalance from useUTXOs (closing the pre-existing orphan flagged earlier), pins btcBalance to the filtered UTXO list rather than the raw sum, and relabels the field 'Available' to match what the value actually represents (the depositable maximum, not the wallet balance). Glossary sweep now covers Depositor claim output, Vault Provider commission, Pre-Pegin and partial-position liquidation, and the participantKeyDrift body picks up the Pre-Pegin casing too.

hasUnconfirmedBalanceOnly and maxTooltip are gone with no residual refs across services/ and packages/. Target Health Factor tooltip matches the ratified definition verbatim.

No critical-path files. LGTM.

PR #2396, opened 2026-09-03
fix(vault): polish the Deposit Progress modal

### LINKED ISSUE #2344: Deposit Progress modal: collapse inactive steps, highlight failures, restyle the detail block
## Context

The Deposit Progress modal, shown after submitting a deposit. It tracks four steps — Register deposit, Set up claim, Sign payout, Activate BTC Vault — each with its own signature counter. Three screenshots are attached: the initial dark-theme state, a mid-flow state on step 2, and a failed state. Rendered by `services/vault/src/components/simple/DepositProgressView/`.

### 1. Step number stroke colour

**Now:** the active step's number circle uses `border-2 border-primary-light`, which renders white in dark theme.
**Expected:** `#5A5A5A`.
**Where:** `services/vault/src/components/simple/DepositProgressView/StepRow.tsx:87`

Note there is no `#5A5A5A` token in the vault theme today — `secondary-strokeDark` is `#cccccc` and `secondary-strokeLight` is `#dddddd` (`services/vault/src/globals.css:80-81`), and the only `#5A5A5A` in the repo is a local constant in `applications/aave/utils/healthFactorDisplay.ts`. Confirm the intended token with design rather than hardcoding.

### 2. Remove the footer caption

**Now:** every state of the modal prints "To ensure a seamless deposit, do not spend the BTC allocated for this process until the transaction is confirmed." under the CTA.
**Expected:** removed.
**Where:** `services/vault/src/components/simple/DepositProgressView/DepositCardShell.tsx`, string in `services/vault/src/copy.ts`

### 3. Give the detail rows their own background

**Now:** on the active step, `Est. remaining` and `Pre-Pegin Bitcoin transaction` sit directly on the step block with no separation.
**Expected:** their own background block, matching Figma. Check the copy in it at the same time.
**Where:** `services/vault/src/components/simple/DepositProgressView/BtcConfirmationDetail.tsx`, `GroupBlock.tsx`

### 4. Collapse everything but the active step

**Now:** completed steps and upcoming steps both stay expanded in the list. In the mid-flow screenshot, step 1 is completed and still occupying space while steps 3 and 4 are listed below.
**Expected:** only the active step is shown expanded.
**Where:** `services/vault/src/components/simple/DepositProgressView/GroupedProgress.tsx`, `CompletedStepsPill.tsx`

### 5. Highlight the failed step in red

**Now:** on failure only the small `✗` next to "Sign and broadcast BTC Pre-Pegin transaction" is red. The sub-step label, the parent step label and the step number all stay black, so the failure is easy to miss against the "Transaction failed" card lower down.
**Expected:** the failed step itself renders red.
**Where:** `services/vault/src/components/simple/DepositProgressView/StepRow.tsx`

## Open question

"Hide completed steps and next ones" reads as show-only-the-active-step. Confirm whether completed steps collapse to a single summary line — `CompletedStepsPill.tsx` suggests that pattern already exists — or disappear entirely.

<details>

> *Deposit Progress modal*
> • stroke colour around numbers: #5A5A5A, not white
> • remove caption below: "To ensure a seamless deposit, do not spend the BTC allocated for this process until the transaction is confirmed."

> *Deposit Progress modal*
> • add background to the block with estimated time and transaction, same as in Figma. and check text there
> • hide completed steps and next ones
> • highlight error step in red colour, so it is more visible

</details>

jeremy-babylonlabs:
## Summary

Polish pass on the Deposit Progress modal: the numbered step circles get a neutral stroke, the footer caption goes away, detail panels sit on their own surface, only the group the flow is actually in stays on screen, and a failed step now reads red end to end.

## Changes

**1. Step-number circles use the stroke token, not the text colour**

`GroupHeader`'s `GroupIndicator` drew its ring in `accent-primary`, which the vault re-points to `#ffffff` in dark mode; `StepRow`'s active 32px circle drew its ring in `primary-light`, the `#387085` teal. Neither is a stroke colour. Both now use `border-secondary-strokeDark`.

The issue notes there is no `#5A5A5A` token — there is: core-ui's dark `secondary.strokeDark` is exactly `#5A5A5A`. Only the light value is re-pointed in the vault (`services/vault/src/globals.css`, to `#cccccc`), so no new token was needed.

**2. Footer caption removed**

`DepositCardShell` loses its `footnote` prop, its rendering and the copy string behind it (`COPY.deposit.progress.doNotSpendWarning`).

The issue quotes the caption as "To ensure a seamless deposit…", but the string shipped in this modal is "Do not spend the BTC used for this deposit until the transactions are confirmed." — same element, different wording. The similarly named `COPY.deposit.refundSuccess.doNotSpendWarning(symbol)` is untouched; `RefundSuccessContent` still uses it.

**3. Detail panels get their own surface**

`BtcConfirmationDetail`, `EthConfirmationDetail`, `ProviderWaitDetail` and `FeeRateSelector` all painted `bg-secondary-highlight` — the exact fill of the `GroupBlock` card that encloses them, so every panel read flat. They now share `rounded-lg border border-secondary-strokeLight bg-primary-contrast p-3`: `#F5F7F2` on the card's `#F9F9F9` in light (separation comes mostly from the `#dddddd` border there), `#191919` on `#252525` in dark. Copy inside the panels is unchanged.

Background token chosen from the existing palette; the Figma page was not reachable, so this is pending a design check.

**4. Only the current group renders**

The stepper previously hid completed groups (they fold into the "X of N steps completed" pill, which stays) and rendered every later group as a collapsed header row. Now only the group holding the current step renders — in `GroupedProgress` and in each lane of `SplitGroupedProgress`.

Split lanes need one extra rule, because the flow parks every vault on `AWAIT_BTC_CONFIRMATION` and then advances only the vault it is processing (`useDepositFlow.ts`). A queued lane therefore sits on the trunk's last step after the trunk has already folded into the pill, and taking the rule literally would erase that vault's whole column. Once the trunk is no longer rendered, each lane's group lookup is clamped into the branch range: a queued vault shows its next group as a not-started header, a finished vault keeps its last group marked complete. Group status and sub-step state still come from the lane's real step, so nothing is misreported. While the trunk is on screen the columns stay hidden — every lane is in the state the trunk card already shows.

The rule is "the group whose inclusive step range covers the current step" (new `groupContainsStep` in `steps.ts`), not "the group whose status is active". The two differ on the pre-sign entry: with `started` false `buildStepGroups` deliberately marks the current group `upcoming`, so an active-only rule would render nothing at all on a WOTS re-offer, and a rule keyed on the Pre-PegIn step would resurrect the finished Register-deposit group. The range rule handles the fee-selector entry, the mid-flow re-offer and completion (nothing renders) with one predicate.

With a single block on screen, `StepConnector` is unreachable and is deleted.

**5. A failed step reads red**

Only the 16px close icon was red before. Now `StepRow` renders an errored step's label in `font-medium text-error-main`, and `GroupHeader` takes a `hasError` prop that turns the group title and the numbered circle red. `GroupBlock` passes `hasError && group.status === "active"`, so an un-started group never inherits a failure.

Colour alone is not an accessible failure signal, so the group indicator also swaps its `aria-label` from "In progress" to a new `COPY.deposit.a11y.groupStatus.failed`. The per-step labels are unchanged.

Worth a design decision before merge: `error-main` is `#D32F2F`, which measures 4.73:1 against the light group card `#F9F9F9` (passes) but 3.08:1 against the dark `#252525` — below the 4.5:1 WCAG AA floor for 14px text. The palette has no compliant red (`error-light` `#EF5350` reaches only 4.40:1 on dark), so this uses the token the design asked for rather than inventing one.

## Tests

- `GroupedProgress.test.tsx` — upcoming groups are no longer rendered, the current one is; pre-sign entry still opens the Pre-PegIn group with its fee selector; the un-started mid-flow re-offer renders its own group and not the finished one; error styling on the group title and sub-step label.
- `SplitGroupedProgress.test.tsx` — each lane renders only its own current group; a queued vault stays visible once the trunk folds away; a finished vault stays visible while a sibling runs; no columns while the shared trunk runs; the trunk keeps the single fee selector on the un-started entry.
- `StepRow.test.tsx` — errored label colour, pending label stays muted.
- `DepositCardShell.test.tsx` / `DepositProgressView.test.tsx` — footnote assertions removed, group-header expectations updated.

211 tests pass across `DepositProgressView`, `DepositSignContent` and `ResumeDepositContent`; lint and `tsc --noEmit` are clean.

Closes #2344

### DIFF
--- a/services/vault/src/components/simple/DepositProgressView/BtcConfirmationDetail.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/BtcConfirmationDetail.tsx
@@ -47,7 +47,7 @@ export function BtcConfirmationDetail({
     : "flex items-center justify-between gap-2";
 
   return (
-    <div className="mt-3 flex flex-col gap-2 rounded-lg bg-secondary-highlight p-3">
+    <div className="mt-3 flex flex-col gap-2 rounded-lg border border-secondary-strokeLight bg-primary-contrast p-3">
       <div className={rowClass}>
         <Text as="span" variant="body2" className="text-accent-secondary">
           {depthReached ? COPY.deposit.waitDetails.status : copy.estRemaining}:
--- a/services/vault/src/components/simple/DepositProgressView/BtcConfirmationDetail.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/BtcConfirmationDetail.tsx
@@ -1,6 +1,7 @@
 import { Loader, Text } from "@babylonlabs-io/core-ui";
 
 import { CopyableHash } from "@/components/shared/CopyableHash";
+import { DETAIL_PANEL_CLASS } from "@/components/shared/layoutClasses";
 import { COPY } from "@/copy";
 import { getBtcExplorerTxUrl } from "@/utils/explorer";
 
@@ -47,7 +48,7 @@ export function BtcConfirmationDetail({
     : "flex items-center justify-between gap-2";
 
   return (
-    <div className="mt-3 flex flex-col gap-2 rounded-lg border border-secondary-strokeLight bg-primary-contrast p-3">
+    <div className={`mt-3 ${DETAIL_PANEL_CLASS}`}>
       <div className={rowClass}>
         <Text as="span" variant="body2" className="text-accent-secondary">
           {depthReached ? COPY.deposit.waitDetails.status : copy.estRemaining}:
--- a/services/vault/src/components/simple/DepositProgressView/GroupBlock.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/GroupBlock.tsx
@@ -2,10 +2,10 @@
  * GroupBlock
  *
  * Renders one step group with the deposit-progress design: the active group
- * expands into a filled card (header → divider → sub-steps), while a collapsed
- * (upcoming) group renders as a single header row. Shared by the single-vault
- * stepper ({@link GroupedProgress}) and each lane of the split stepper
- * ({@link SplitGroupedProgress}) so both look identical.
+ * expands into a filled card (header → divider → sub-steps), while an
+ * un-started (pre-entry) group renders as a single header row. Shared by the
+ * single-vault stepper ({@link GroupedProgress}) and each lane of the split
+ * stepper ({@link SplitGroupedProgress}) so both look identical.
  */
 
 import type { StepperItem } from "@babylonlabs-io/core-ui";
@@ -65,6 +65,8 @@ export function GroupBlock({
     group.startStep <= PRE_PEGIN_VISUAL_STEP &&
     PRE_PEGIN_VISUAL_STEP <= group.endStep;
 
+  const headerHasError = hasError && group.status === "active";
+
   if (!group.expanded && !showPreSignDetail) {
     return (
       <GroupHeader
@@ -73,6 +75,7 @@ export function GroupBlock({
         status={group.status}
         completedInGroup={group.completedInGroup}
         totalInGroup={group.totalInGroup}
+        hasError={headerHasError}
       />
     );
   }
@@ -91,6 +94,7 @@ export function GroupBlock({
           status={group.status}
           completedInGroup={group.completedInGroup}
           totalInGroup={group.totalInGroup}
+          hasError={headerHasError}
         />
         <div className="border-t border-secondary-strokeLight" />
         <div className="flex flex-col gap-2 px-2">
--- a/services/vault/src/components/simple/DepositProgressView/GroupBlock.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/GroupBlock.tsx
@@ -15,7 +15,7 @@ import { DepositFlowStep } from "@/hooks/deposit/depositFlowSteps/types";
 
 import { GroupHeader } from "./GroupHeader";
 import { StepRow, type StepRowState } from "./StepRow";
-import { getVisualStep, type StepGroupView } from "./steps";
+import { getVisualStep, groupContainsStep, type StepGroupView } from "./steps";
 
 /** The step whose transaction the pre-sign fee rate pays for. */
 const PRE_PEGIN_VISUAL_STEP = getVisualStep(
@@ -65,7 +65,7 @@ export function GroupBlock({
     group.startStep <= PRE_PEGIN_VISUAL_STEP &&
     PRE_PEGIN_VISUAL_STEP <= group.endStep;
 
-  const headerHasError = hasError && group.status === "active";
+  const headerHasError = hasError && groupContainsStep(group, currentStep);
 
   if (!group.expanded && !showPreSignDetail) {
     return (
@@ -101,12 +101,11 @@ export function GroupBlock({
           {showPreSignDetail ? (
             <>
               <StepRow
-                state="active"
+                state={headerHasError ? "error" : "active"}
                 number={PRE_PEGIN_VISUAL_STEP - group.startStep + 1}
                 ariaNumber={PRE_PEGIN_VISUAL_STEP}
                 label={steps[PRE_PEGIN_VISUAL_STEP - 1]?.label ?? ""}
                 compact={compact}
-                inCard
               />
               {preSignDetail}
             </>
@@ -133,9 +132,7 @@ export function GroupBlock({
                   label={step.label}
                   description={step.description}
                   detail={activeStepDetail}
-                  hasNext={subIndex < stepNumbers.length - 1}
                   compact={compact}
-                  inCard
                 />
               );
             })
--- a/services/vault/src/components/simple/DepositProgressView/GroupHeader.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/GroupHeader.tsx
@@ -13,17 +13,23 @@ interface GroupHeaderProps {
   status: GroupStatus;
   completedInGroup: number;
   totalInGroup: number;
+  /** True when this group's current step failed — render title + circle red. */
+  hasError?: boolean;
 }
 
 function GroupIndicator({
   status,
   number,
+  hasError,
 }: {
   status: GroupStatus;
   number: number;
+  hasError: boolean;
 }) {
   const base = "flex h-8 w-8 shrink-0 items-center justify-center rounded-full";
-  const ariaLabel = COPY.deposit.a11y.groupStatus[status];
+  const ariaLabel = hasError
+    ? COPY.deposit.a11y.groupStatus.failed
+    : COPY.deposit.a11y.groupStatus[status];
 
   if (status === "completed") {
     return (
@@ -38,13 +44,21 @@ function GroupIndicator({
 
   return (
     <div
-      className={twMerge(base, "border-2 border-accent-primary")}
+      className={twMerge(
+        base,
+        hasError
+          ? "border-2 border-error-main"
+          : "border-2 border-secondary-strokeDark",
+      )}
       aria-label={ariaLabel}
     >
       <Text
         as="span"
         variant="body2"
-        className="font-medium text-accent-primary"
+        className={twMerge(
+          "font-medium",
+          hasError ? "text-error-main" : "text-accent-primary",
+        )}
       >
         {number}
       </Text>
@@ -58,14 +72,18 @@ export function GroupHeader({
   status,
   completedInGroup,
   totalInGroup,
+  hasError = false,
 }: GroupHeaderProps) {
   return (
     <div className="flex items-center gap-3">
-      <GroupIndicator status={status} number={number} />
+      <GroupIndicator status={status} number={number} hasError={hasError} />
       <Text
         as="span"
         variant="body1"
-        className="flex-1 font-medium text-accent-primary"
+        className={twMerge(
+          "flex-1 font-medium",
+          hasError ? "text-error-main" : "text-accent-primary",
+        )}
       >
         {title}
       </Text>
--- a/services/vault/src/components/simple/DepositProgressView/GroupHeader.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/GroupHeader.tsx
@@ -1,5 +1,5 @@
 import { Text } from "@babylonlabs-io/core-ui";
-import { IoCheckmarkSharp } from "react-icons/io5";
+import { IoCheckmarkSharp, IoCloseSharp } from "react-icons/io5";
 import { twMerge } from "tailwind-merge";
 
 import { COPY } from "@/copy";
@@ -42,23 +42,29 @@ function GroupIndicator({
     );
   }
 
+  if (hasError) {
+    return (
+      <div
+        className={twMerge(base, "border-2 border-error-main")}
+        aria-label={ariaLabel}
+      >
+        <IoCloseSharp size={16} className="text-error-main" />
+      </div>
+    );
+  }
+
   return (
     <div
       className={twMerge(
         base,
-        hasError
-          ? "border-2 border-error-main"
-          : "border-2 border-secondary-strokeDark",
+        "border-2 border-accent-secondary dark:border-secondary-strokeDark",
       )}
       aria-label={ariaLabel}
     >
       <Text
         as="span"
         variant="body2"
-        className={twMerge(
-          "font-medium",
-          hasError ? "text-error-main" : "text-accent-primary",
-        )}
+        className="font-medium text-accent-primary"
       >
         {number}
       </Text>
--- a/services/vault/src/components/simple/DepositProgressView/SplitGroupedProgress.tsx
+++ b/services/vault/src/components/simple/DepositProgressView/SplitGroupedProgress.tsx
@@ -6,8 +6,11 @@
  * each vault is on its own VP-paced timeline (WOTS submission, payout signing,
  * artifact download, activation) and can diverge by an hour or more. This
  * component renders the shared "Register deposit" group as a single trunk and
- * the remaining groups as one column per vault, reusing the same GroupBlock
- * (filled active-group card + collapsed header rows) as the single-vault stepper.
+ * the remaining groups as one column per vault, reusing the same GroupBlock as
+ * the single-vault stepper. Each region shows one group: the trunk while the
+ * shared step is still inside it, then one per lane — the group holding that
+ * vault's own step, clamped so a queued vault shows its next group and a
+ * finished one its last.
  */
 
 import type { StepperItem } from "@babylonlabs-io/core-ui";
@@ -18,11 +21,12 @@ import { COPY } from "@/copy";
 import { DepositFlowStep } from "@/hooks/deposit/depositFlowSteps/types";
 
 import { GroupBlock } from "./GroupBlock";
-import { StepConnector } from "./StepConnector";
 import {
   buildStepGroups,
   derivePerVaultStep,
   getVisualStep,
+  groupContainsStep,
+  TOTAL_VISUAL_STEPS,
   TRUNK_END_VISUAL_STEP,
 } from "./steps";
 
@@ -66,7 +70,7 @@ interface SplitGroupedProgressProps {
   preSignDetail?: ReactNode;
 }
 
-/** One group list per vault, rendered as the new filled-card / header blocks. */
+/** One vault's lane: its label plus the group holding that vault's own step. */
 function VaultColumn({
   vaultIndex,
   branchGroups,
@@ -85,6 +89,8 @@ function VaultColumn({
   hasError: boolean;
   activeStepDetail?: ReactNode;
 }) {
+  if (branchGroups.length === 0) return null;
+
   return (
     <div className="flex min-w-0 flex-1 flex-col">
       <Text
@@ -95,24 +101,19 @@ function VaultColumn({
         {COPY.deposit.progress.splitVaultColumnLabel(vaultIndex + 1)}
       </Text>
       <div className="flex flex-col">
-        {branchGroups.map(({ group, number }, idx) => {
-          const isLast = idx === branchGroups.length - 1;
-          return (
-            <div key={group.startStep} className="flex flex-col">
-              <GroupBlock
-                group={group}
-                number={number}
-                steps={steps}
-                currentStep={perVaultVisualStep}
-                hasError={hasError}
-                activeStepDetail={activeStepDetail}
-                // Columns are narrow → stack each row's sub-counter.
-                compact
-              />
-              {!isLast && <StepConnector />}
-            </div>
-          );
-        })}
+        {branchGroups.map(({ group, number }) => (
+          <GroupBlock
+            key={group.startStep}
+            group={group}
+            number={number}
+            steps={steps}
+            currentStep={perVaultVisualStep}
+            hasError={hasError}
+            activeStepDetail={activeStepDetail}
+            // Columns are narrow → stack each row's sub-counter.
+            compact
+          />
+        ))}
       </div>
     </div>
   );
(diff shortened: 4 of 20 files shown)


jonybur (dismissed):
Group-visibility rework is coherent: groupContainsStep is the right primitive, the trunk/branch clamp in SplitGroupedProgress accounts for AWAIT_BTC_CONFIRMATION parking and finished lanes, and StepConnector is fully removed. DepositCardShell footnote and progress.doNotSpendWarning drop with no remaining refs; refundSuccess.doNotSpendWarning is a different key. Error styling threads a hasError-active gate through GroupBlock, and the a11y label swap to failed is a real signal beyond colour. Contrast tradeoff on error-main in dark is flagged in the description for design. LGTM.

services/vault/src/components/simple/DepositProgressView/GroupHeader.tsx
```diff
@@ -38,13 +44,21 @@ function GroupIndicator({
 
   return (
     <div
-      className={twMerge(base, "border-2 border-accent-primary")}
+      className={twMerge(
+        base,
+        hasError
+          ? "border-2 border-error-main"
+          : "border-2 border-secondary-strokeDark",
```
gbarkhatov: **Merge-blocker.** This swap regresses the step-number ring to near-invisible in the app's default theme — on the one element the redesign is built around.

`accent-primary` -> `secondary-strokeDark` looks like a neutral token change, but both are re-pointed by the vault. `services/vault/src/globals.css:79-82` sets light `--twc-accent-primary` to `#000000` and light `--twc-secondary-strokeDark` to `#cccccc` (the file's own comment explains the block sits below `@tailwind` so source order makes it win). `providers.tsx:35` sets `defaultTheme="light"`. So on the `#F9F9F9` card the ring goes from about **20:1 to 1.53:1** — against WCAG 1.4.11's 3:1 floor for UI components, the circles effectively disappear.

Dark is fine and does get the intended `#5A5A5A` (2.22:1), so this is a light-theme-only fix: either give light `--twc-secondary-strokeDark` a value that works for a ring rather than a hairline divider, or use `accent-secondary` (`#666666` light / `#B0B0B0` dark) and confirm against the design.

Flagging the mechanism explicitly because reading `tailwind.config.js:237` alone shows `#387085` and makes this look harmless — it isn't the value that ships.

services/vault/src/components/simple/DepositProgressView/StepRow.tsx
```diff
@@ -84,7 +84,7 @@ function StepCircle({
   if (state === "active") {
     return (
       <div
-        className={twMerge(base, "border-2 border-primary-light")}
+        className={twMerge(base, "border-2 border-secondary-strokeDark")}
```
gbarkhatov: **Merge-blocker (zero dead code).** This line is ask (1)'s stroke edit, and production never renders it.

`StepCircle`'s `if (inCard)` block at lines 28-61 returns on every path — completed, error, active, and the unconditional pending return at line 55 — so everything from line 63 down is unreachable whenever `inCard` is true. I grepped every `StepRow` consumer: `GroupBlock` is the only production one, and **both** call sites pass `inCard` (`GroupBlock.tsx:109` and `:138`). The visible ask-(1) fix actually ships via `GroupHeader`, so nothing is broken — but this edit does nothing.

The same reachability kills `showTrailingLine` (`!inCard && ...`), the `hasNext` prop, and the timeline-column `<div>`, while `GroupBlock:136` still computes and passes `hasNext`. Suggest deleting the non-`inCard` `StepCircle` branch, `hasNext` and `showTrailingLine`, and making `inCard` non-optional so the type stops permitting a shape nothing builds.

Worth noting `StepRow.test.tsx` renders without `inCard`, so those tests currently exercise the dead shape — the label assertions still hit shared code and stay valid, but the setup is misleading.

services/vault/src/components/simple/DepositProgressView/GroupBlock.tsx
```diff
@@ -65,6 +65,8 @@ export function GroupBlock({
     group.startStep <= PRE_PEGIN_VISUAL_STEP &&
     PRE_PEGIN_VISUAL_STEP <= group.endStep;
 
+  const headerHasError = hasError && group.status === "active";
```
gbarkhatov: Gating on `"active"` is the right instinct, but it suppresses the failure highlight in exactly the state ask (5) exists for.

`buildStepGroups` (`steps.ts:220`) computes `status = started || currentStep > group.startStep ? "active" : "upcoming"`, so an un-started group holding the current step is `"upcoming"`, not `"active"`. That's the pre-sign entry and the mid-flow WOTS re-offer. In that state `headerHasError` is false, so neither the title, the circle, nor any sub-step turns red — leaving only the Callout, which is precisely the "easy to miss" symptom the ask is meant to remove.

The PR body actually relies on this quirk when it justifies choosing `groupContainsStep` over an active-only rule elsewhere, which suggests the same predicate belongs here: `groupContainsStep(group, currentStep)` is already imported in this file.

Same hole from the other side at line 104: the `showPreSignDetail` branch hardcodes `state="active"`, so that row can never render `"error"`.

services/vault/src/components/simple/DepositProgressView/SplitGroupedProgress.tsx
```diff
@@ -171,6 +173,15 @@ export function SplitGroupedProgress({
...
           const columnStarted = started || vaultRawStep !== rawStep;
+          // Once the trunk folds away every lane must still show a group: the
+          // flow parks queued vaults on the trunk's last step, and a finished
+          // vault sits past the last one. Status stays the lane's real step.
+          const laneGroupStep = trunkVisible
+            ? perVaultVisualStep
+            : Math.min(
+                Math.max(perVaultVisualStep, TRUNK_END_VISUAL_STEP + 1),
+                TOTAL_VISUAL_STEPS,
+              );
```
gbarkhatov: The clamp is applied only when `!trunkVisible`, which drops the guard in the window where it still matters.

`Math.min(..., TOTAL_VISUAL_STEPS)` exists specifically to pull the `COMPLETED` sentinel (visual 16, `steps.ts:293`) back to 15. While the trunk renders, `laneGroupStep` is left as the raw `perVaultVisualStep`, so a finished lane matches no group, `branchGroups` comes back empty, and the new `if (branchGroups.length === 0) return null` at line 92 removes the whole column including its heading.

Reachability is genuinely unclear and I couldn't settle it: the live path can't get there because `useDepositFlow.ts:1135` parks every vault on `AWAIT_BTC_CONFIRMATION`, but `deriveSplitVaultProgress` computes each sibling independently and can return `COMPLETED` via `isVaultPastActivation` with no cap against the active vault's step (`useSplitVaultProgress.ts:106`), so the resume path doesn't structurally forbid it. Treat the reachability as unverified — the missing clamp isn't.

Clamping unconditionally and gating the columns region on `!trunkVisible` is the rule this file's own header comment already describes, and would make the two halves agree.

services/vault/src/components/simple/DepositProgressView/GroupHeader.tsx
```diff
@@ -38,13 +44,21 @@ function GroupIndicator({
...
       <Text
         as="span"
         variant="body2"
-        className="font-medium text-accent-primary"
+        className={twMerge(
+          "font-medium",
+          hasError ? "text-error-main" : "text-accent-primary",
+        )}
       >
         {number}
```
gbarkhatov, line 69: On `hasError` the group circle keeps its plain number glyph and changes only `border-error-main` + `text-error-main` — so at the group level, failure is signalled by hue alone.

That's WCAG 1.4.1 (use of colour) rather than just a contrast miss, and the `aria-label` swap fixes only the screen-reader path — it does nothing for a sighted low-vision or colour-deficient user. The sub-step row already does the right thing with a shape change (`IoCloseSharp`, `StepRow.tsx:36-42`); the header just doesn't follow it.

The contrast is short too: `#D32F2F` on the `#252525` dark card is the 3.08:1 you disclosed, and the title is `body1` (16px/500), so it isn't WCAG "large text" and needs 4.5:1. `error-light` `#EF5350` only reaches 4.40:1, so no existing token closes it — which is why the redundant glyph seems the better primary fix, with a `--twc-error-main` re-point inside the `.dark` block already established at `globals.css:87-92` if you want the contrast as well.

services/vault/src/components/simple/DepositProgressView/__tests__/GroupedProgress.test.tsx
```diff
@@ -103,6 +112,53 @@ describe("GroupedProgress", () => {
...
+  it("marks the group title as failed when the current step errors", () => {
+    render(<GroupedProgress steps={steps} currentStep={8} hasError />);
+
+    expect(screen.getByText(COPY.deposit.groups.signWots).className).toContain(
+      "text-error-main",
+    );
+    // The failed sub-step label turns red too.
+    expect(
+      screen.getByText(COPY.deposit.steps.awaitPayoutTransactions).className,
+    ).toContain("text-error-main");
```
gbarkhatov: The new error tests assert on Tailwind class strings, and the one thing the PR adds as a non-colour signal has no test at all.

`COPY.deposit.a11y.groupStatus.failed` is referenced only by `GroupHeader.tsx:31` — no test touches it, while its `active`, `upcoming` and `completed` siblings all have `getByLabelText` assertions in these same suites. Meanwhile this test and `StepRow.test.tsx:30-46` reach into `element.className` for `"text-error-main"`, which couples to implementation and breaks on any utility-class refactor without behaviour changing.

Both problems have the same fix: `getByLabelText(COPY.deposit.a11y.groupStatus.failed)` decouples the assertion and covers the accessibility claim in one stroke. Given the comment on line 51 about colour-alone signalling, that label is the part most worth pinning.

services/vault/src/components/simple/DepositProgressView/BtcConfirmationDetail.tsx
```diff
@@ -47,7 +47,7 @@ export function BtcConfirmationDetail({
     : "flex items-center justify-between gap-2";
 
   return (
-    <div className="mt-3 flex flex-col gap-2 rounded-lg bg-secondary-highlight p-3">
+    <div className="mt-3 flex flex-col gap-2 rounded-lg border border-secondary-strokeLight bg-primary-contrast p-3">
```
gbarkhatov: Unverified against the design, flagging so the next person doesn't "fix" it in the wrong direction: ask (3)'s fill is imperceptible in the default light theme.

The panels move to `bg-primary-contrast` (`#F5F7F2` light / `#191919` dark) inside `GroupBlock`'s `bg-secondary-highlight` card (`#F9F9F9` / `#252525`). Dark separates clearly (ΔL* about 6); light is L* 96.97 vs 97.92 and slightly *greener* rather than darker, so the separation rests entirely on the `#dddddd` hairline. Figma was unreachable, so I can't say whether that matches intent.

Two smaller things while this is open. The 68-character panel class string is now duplicated verbatim across this file, `EthConfirmationDetail.tsx:42`, `ProviderWaitDetail.tsx:43` and `FeeRateSelector.tsx:193` — whatever this settles on will need changing in four places, so it may be worth a shared constant alongside the existing layout classes.

And ask (3)'s second half — "check the copy in it" — produced no diff. The existing strings do already conform to the Pre-Pegin convention, so "reviewed, no change" may well be the honest answer; worth saying so on #2344 explicitly, or that half of the ask disappears with the issue.

gbarkhatov (dismissed):
Reviewed all five asks end to end and traced the collapse logic through `buildStepGroups`, both `StepRow` call sites, and the split-lane path. Asks (2) and (4) land cleanly, the `StepConnector` deletion is complete with zero stragglers, and the `groupContainsStep` ranges are contiguous and exhaustive so the stepper can't blank mid-flow.

Two things I'd want resolved before merge, both inline — one contrast regression in the default theme, one where the PR's own ask-(1) edit lands in a branch production never renders.

**Worth stating plainly because it caught me out too:** these colour utilities do not resolve from `packages/babylon-core-ui/tailwind.config.js` alone. core-ui builds its palette with `tw-colors` `createThemes`, so they read `--twc-*` custom properties, and `services/vault/src/globals.css:74-82` re-points several per theme — light `accent-primary` to `#000000`, light `secondary-strokeDark` to `#cccccc`. With `defaultTheme="light"` at `providers.tsx:35`, reading the config hex gives the wrong answer in both directions.

Three more that aren't inline, because their lines fall outside a diff hunk:

- **`SplitGroupedProgress.tsx:206-210`** — the lane error gate requires `vaultIndex === currentVaultIndex`, but `useDepositFlow.ts:1379-1381` calls `setCurrentVaultIndex(null)` immediately before awaiting payout readiness. With a null index no lane matches, every column gets `hasError={false}`, and the trunk isn't rendered at that step — so a throw during the payout-readiness window paints red nowhere in the stepper. Pre-existing and not in this diff; raised only because this PR is the ask-(5) work and closes the issue. A fallback marking every lane whose `perVaultVisualStep === currentStep` when the index is null would close it.
- **`StepRow.tsx:148,207`** — `hasDetail = isActive && Boolean(detail)` and `{isActive && detail}` mean an errored step renders no detail panel. `BtcConfirmationDetail` is the only place in the dialog with a BTC explorer link, so a failure at that step loses the txid. Also pre-existing, but hiding the surrounding groups makes it bite harder: the card then holds nothing but a red row and a callout. `(isActive || state === "error") && Boolean(detail)` would do it.
- **Issue #2344's Open question was resolved unilaterally.** The issue asks whether completed steps collapse to a summary or disappear, flagging that ask (4)'s "only the active step is shown expanded" contradicts the original report's "hide completed steps and next ones". This PR chose: completed to `CompletedStepsPill`, upcoming gone entirely — so a depositor at step 7 sees one card and a pill, with nothing naming "Sign payout" or "Activate BTC Vault" during a flow advertising a multi-hour estimate. That's a product decision rather than a bug; it just wants the sign-off the issue asked for, and a note on #2344 so it's traceable.

**CI caveat:** all 12 checks pass including `verify`, but `visual` proves nothing here — its connected `DEPOSIT_FLOW_STOPS` end at `deposit-split-options`, so the post-submit Deposit Progress modal is in no snapshot. Every pixel this PR changes is unphotographed.

Checked and dismissed, so nobody re-raises them: the `footnote` prop and `progress.doNotSpendWarning` are fully removed while the distinct `refundSuccess.doNotSpendWarning` and `broadcastSuccess.footnote` correctly survive with live consumers; `GroupedProgress.tsx:50`'s `return null` is reachable only at the post-completion sentinel, where the success callout and 4-of-4 pill render independently; the new `groupStatus.failed` string is correctly in `copy.ts:384` in sentence case; and the E2E walk is untouched — `DepositProgressView/` carries no production `data-testid`, and `stepMachine.ts` gates on button text, `[role="progressbar"]`, and the `/tx/` anchor at visual step 8, whose group still renders.

One claim I want to retract explicitly rather than leave standing: "a vault column vanishes whenever one sibling diverges past the trunk" is overstated — lanes rendering nothing while inside the trunk is the documented intent. The real hole is the narrower `COMPLETED`-sentinel case in the comment on `SplitGroupedProgress.tsx`.

gbarkhatov (dismissed):

jonybur (dismissed):
Address-review delta correctly promotes the ring token to a value that resolves against the vault's per-theme override (visible in light again), deletes the unreachable StepRow branch, and moves the failed-group signal off group.status onto containment so the mid-flow re-offer paints red as intended. The split-lane clamp now runs unconditionally, so a queued vault holds its column while the trunk folds away. A failing group additionally carries a glyph, so colour is not the sole error signal. Detail-panel styling is shared via DETAIL_PANEL_CLASS in layoutClasses.

Sweep from the earlier pass still holds: DepositCardShell's footnote and progress.doNotSpendWarning are fully removed, refundSuccess.doNotSpendWarning and broadcastSuccess.footnote correctly survive with live consumers, and StepConnector.tsx is gone. No critical-path files. LGTM.

services/vault/src/components/simple/DepositProgressView/GroupHeader.tsx
```diff
@@ -58,14 +78,18 @@ export function GroupHeader({
...
     <div className="flex items-center gap-3">
-      <GroupIndicator status={status} number={number} />
+      <GroupIndicator status={status} number={number} hasError={hasError} />
       <Text
         as="span"
         variant="body1"
-        className="flex-1 font-medium text-accent-primary"
+        className={twMerge(
+          "flex-1 font-medium",
+          hasError ? "text-error-main" : "text-accent-primary",
```
gbarkhatov, line 91: The 1.4.1 half of this is closed — `GroupIndicator` now renders `IoCloseSharp` in place of the number glyph on error, matching what `StepRow` already did, and `__tests__/GroupedProgress.test.tsx:175-181` pins it. Failure is no longer signalled by hue alone.

What remains is only the contrast. The title here is still `text-error-main`; `error.main` is `#D32F2F` in **both** core-ui theme blocks (`tailwind.config.js:240` and `:292`), and no `--twc-error-main` re-point was added to the `.dark` block in `globals.css`. So on the dark card surface it stays at roughly 3.1:1, against the 4.5:1 that `body1` at 16px/500 needs.

My original comment framed this as optional once the glyph landed, and that still holds — with the icon carrying the meaning, this is a judgment call rather than a residual accessibility blocker. Recording it so the state is on the record: a dark-theme `error-main` re-point would close it if you want it closed, and it would fix every other `text-error-main` in the vault at the same time.

gbarkhatov (approved):
Follow-up pass over my earlier review (`5099812357`, at `6aad40da`), not a second opinion on the same code. The material since then is `46bcf335f`, `725f15e13` ("address Deposit progress review") and `07996f35a`.

**Five of the seven are fixed; two are partially fixed.** I checked each against the code at HEAD rather than inferring from the commit messages.

|---|---|---|---|
| 1 | Step-number ring `accent-primary` to `secondary-strokeDark` regresses to ~1.53:1 in light | **fixed** | `GroupHeader.tsx:60` |
| 2 | The stroke edit sat in the unreachable non-`inCard` branch; `hasNext` / `showTrailingLine` / timeline column dead | **fixed** | `StepRow.tsx:10-54, 94` |
| 3 | `headerHasError` gated on `status === "active"`, so an `upcoming` group holding the current step showed no red; plus hardcoded `state="active"` | **fixed**, both halves | `GroupBlock.tsx:68, 104` |
| 4 | The `Math.min(…, TOTAL_VISUAL_STEPS)` clamp applied only when `!trunkVisible`, so a `COMPLETED` lane lost its whole column | **fixed** | `SplitGroupedProgress.tsx:162, 180-183` |
| 5 | Group-level failure signalled by hue alone, and `#D32F2F` on `#252525` is 3.08:1 against 4.5:1 | **partially fixed** — glyph landed, contrast half remains | `GroupHeader.tsx:45-53` fixed / `:91` remains |
| 6 | Error tests asserted Tailwind class strings; `groupStatus.failed` untested | **fixed** | `__tests__/GroupedProgress.test.tsx:150-181` |
| 7 | `bg-primary-contrast` fill imperceptible in light; the 68-char class duplicated four times | **partially fixed** — dedup landed, token improved but still design-unconfirmed | `layoutClasses.ts:52-53` |

The existing threads stand as written; nothing there needs restating. One inline note below on the half of 5 that survives.

**No new defects.** `07996f35a` is a single token swap and I verified it rather than taking the commit message's word: `background-contrast` is a real core-ui token defined in both theme blocks (`tailwind.config.js:220` light `#F2F2F2`, `:272` dark `#191919`) and is not re-pointed in the vault's `globals.css`. Dark is byte-identical to the `primary-contrast` it replaces (`#191919`); light moves `#F5F7F2` to `#F2F2F2` — a true neutral, and darker than the `#F9F9F9` card rather than greener. So the direction is now right. The magnitude stays subtle, and the "pending a design check" caveat in the description still applies.

On finding 7's other half: `DETAIL_PANEL_CLASS` at `layoutClasses.ts:52-53` is now consumed by `BtcConfirmationDetail.tsx:51`, `EthConfirmationDetail.tsx:43`, `ProviderWaitDetail.tsx:44` and `FeeRateSelector.tsx:194`, and no `bg-primary-contrast` remains anywhere under `DepositProgressView/`. That closes the four-places-to-change objection — and is what let the final fix be a one-liner.

### Not a verdict, carried over

`StepRow.tsx:83` and `:124` still gate the detail panel on `isActive` alone, so an errored step renders no detail and loses the explorer txid. That was in my review body rather than an inline thread, so no verdict is owed on it; unchanged by this delta, and noting it only so it isn't lost.

### CI

`verify` passed on this head (9m41s), so this PR's type-check and unit suite are genuinely green — unlike most of the batch, where a pre-existing `main` breakage aborts the build. `validate` fails at "Validate commit signatures": the branch needs a signed commit before merge, which is not a code defect.

PR #2397, opened 2026-09-03
feat(ts-sdk, wasm): add delegated-claim watchtower artifact assembly

kirugan:
**The goal**
Allow user to claim their btc from vault in case VP is unavailable

**Watchtower (cli tool for claiming artifacts)**

Docs - [github docs](https://github.com/babylonlabs-io/btc-vault/blob/main/docs/delegated_claim.md)
Previous test - [confluence docs](https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/458916185/Delegated+Claim+Depositor-as-Claimer+UAT+pegin+efc1edaa?atlOrigin=eyJpIjoiOWM4NzI1ZjFhNDZhNDY3YWFmNmEwMzJkZTk4MTkwYmIiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ)

**Proposed SDK changes with mock UI**
[Slack message](https://babylonlabsworkspace.slack.com/archives/C087NF7U18X/p1787058832761129)

**Required PR-s from other repos**
Vault-wasm - [here](https://github.com/babylonlabs-io/vault-wasm/pull/6)
Btc-vault - [here](https://github.com/babylonlabs-io/btc-vault/pull/2655), that depends on PR above

feat(ts-sdk, wasm): add the delegated-claim watchtower artifact surface

## What

- A depositor can now assemble, in one batched wallet interaction, the two files a delegated claim runs from: `artifacts.json` (the claimer-side signature set bundled with the vault provider's transaction graph) and `wots_keypair.json`.
- `assembleWatchtowerArtifacts` builds the Claim, Assert, both Payout and the per-challenger WronglyChallenged PSBTs, signs them in a single prompt, and maps each signature back to the challenger and garbled-circuit index it was computed for.
- `deriveClaimerWotsKeypair` re-derives the depositor's WOTS keypair from the wallet on demand and checks it against the vault's on-chain `depositorWotsPkHash` before the vault provider's graph is consulted. The keypair is never stored; the vault root and the WOTS seed are zeroed on every exit path.
- `summarizeWatchtowerArtifacts` and `assertArtifactsUsableForVault` check a file that arrives from disk: field shapes, the vault it belongs to, and every bundled signature against the file's own graph.
- `vaultIdBinding` ties a vault-provider-served graph to a vault. Neither the graph nor the artifacts file is bound to a vault id by construction, so the Claim's first input — the PegIn output the on-chain vault id is derived from — is the binding. It runs at assembly before any wallet prompt, and again when a file is read.
- Claim-time execution is exposed as well: `pinPegoutProof`, `attachFinalizedAssert`, `finalizePayout` and `finalizeWronglyChallenged`. A depositor holding the two files completes a claim from a browser with no CLI and no vault provider.
- The engine side adds `babylon-tbv-rust-wasm/src/delegatedClaim.ts`, one implementation shared by the browser and Node entries, with width guards on every integer that crosses to WASM.
- `VAULT_WASM_COMMIT` is pinned to vault-wasm `main` at `4cd516e85719ad7be30ed45c81945100bce7ab15`, whose `btc_vault_v3` is `ac4954e7` — btc-vault's merge commit on `main`. `dist/generated` is rebuilt from that pin.
- The whole surface is marked `@experimental`.

## Why

- Today a depositor whose vault provider disappears cannot recover their BTC without that provider. These two files are the recovery path: they are assembled while the provider is still reachable and the depositor is at the keyboard, and they stay valid for a claim months later.
- The signatures are Taproot script-path signatures whose sighashes do not cover witness data. The WOTS values, the Groth16 proof and the hashlock preimages are all witness-only, so one signing session authorizes a claim however it later plays out.

## Approaches

- **Leave claim-time execution to the `vaultd vp wt` CLI**: rejected. The files alone are inert for a depositor who cannot run the CLI, and the whole point is the case where nobody else will act. The four finalizers ship with the assembly surface.
- **Derive the vault-id binding in Rust, inside `build_watchtower_artifacts`**: rejected. The depositor's Ethereum address is not in the graph, so Rust would have to be handed it anyway, and the check belongs on the side that knows which vault the user chose.
- **Reject `claimableEventBlockNumber` of zero at assembly time**: rejected. The event lands when the Ethereum withdrawal starts, which is the moment the claim begins; assembly is meant to happen long before that, while the provider is still online. Forcing assembly after the event would mean assembling exactly when the provider may be gone.
- **Parse and re-serialize the artifacts file in TypeScript**: rejected. Real BaBe sessions run to hundreds of megabytes per challenger; the format stays Rust-owned and this package passes it through.

## Verification

- The `VAULT_WASM_COMMIT` bump was gated as CLAUDE.md section 4 requires. After re-pinning and rebuilding `dist/generated`, the JS golden vectors passed unchanged: `vault-secrets/__tests__/expand.test.ts`, `vault-secrets/__tests__/context.golden.test.ts`, `wots/__tests__/blockDerivation.test.ts` and the new `delegated-claim/__tests__/wotsKeypair.golden.test.ts`. The three expanders and the WOTS derivation produce identical bytes across the bump, so no existing deposit's secrets rotate.
- `wotsKeypair.golden.test.ts` also runs the claim-time WASM derivation differentially against the deposit-time TypeScript path, so the two implementations of `depositorWotsPkHash` are pinned to each other and not only to a stored vector.
- Byte-compatibility of the `artifacts.json` format itself is pinned by btc-vault's `crates/vault/src/delegated_claim/artifacts.rs` round-trip test, not here; every WASM call in the service tests is mocked.

## Not in this PR

- No caller. The browser page that drove the signet end-to-end test is deliberately excluded; it arrives later as an example.
- The Groth16 proof comes from the prover service, and nothing here watches the chain for a ChallengeAssert.
- A way to set `claimable_event_block_number` on an artifacts file that already exists (pre-review #7).
- Splitting the delegated-claim forwarders out of `core/wasm/index.ts` (pre-review #15).

## Pre-review

<details>
<summary>Pre-review: 15 findings · 0 fixed · 0 moot · 2 follow-up · 1 declined · 12 open (3 merge-blockers) · 0 undecided · lint passed, test failed (unrelated: stale core-ui dist)</summary>

|---|---|---|---|
| 1 | The wallet signs the delegated-claim Payout with no TypeScript assertion on its output script or value; everything comes from the VP-served graph. | merge-blocker | open |
| 2 | The challenger set is whatever the VP-served graph lists; nothing compares it with the on-chain keeper set, and undersigning is silent and unrecoverable. | merge-blocker | open |
| 3 | Every byte-order fixture in the new tests is a palindrome, so the reversal in `displayTxid` is untested and could be deleted with the suite still green. | merge-blocker | open |
| 4 | The new delegated-claim files appear in none of the repository's critical-path inventories, so no owner is assigned and the strict-typing lint rules do not apply to them. | normal | open |
| 5 | `vault_core_version` is parsed and returned but never compared with the version the file is verified under, and the parameter doc describes a check that does not exist. | normal | open |
| 6 | `txGraphVersion` crosses to WASM with no width guard in the very file that defines `assertU16` and documents the truncation hazard. | normal | open |
| 7 | An artifacts file assembled as the module recommends carries `claimableEventBlockNumber` `0n` and is then rejected by the module's own checker, with no in-SDK repair path. | normal | follow-up: a function that rewrites the field in an existing file, reviewed on its own |
| 8 | The four claim-time finalizers are re-exported raw from the WASM boundary into the public services entry with no guard and no caller. | normal | declined: they are the claim itself and are marked `@experimental`; the caller arrives as an example |
| 9 | The TS-versus-WASM WOTS differential runs over a single fixed seed, which pins one vector rather than the function. | normal | open |
| 10 | `parseWasmJson` blind-casts WASM JSON, and one malformed shape desynchronises the whole per-challenger signature mapping. | normal | open |
| 11 | `verifyingKeyHex` is VP-supplied, bound to nothing, and is the key the Groth16 proof is later verified against. | normal | open |
| 12 | `CLAIM_PEGIN_INPUT` is exported but used only inside its own file, and two public symbols lack the `@experimental` tag every sibling carries. | normal | open |
| 13 | A test name states a return where the test asserts a rejection, and one test asserts only truthiness. | normal | open |
| 14 | A comment states the batch is N+3 PSBTs, which predates the depositor Payout PSBT joining it. | normal | open |
| 15 | The delegated-claim forwarding block pushes the lazy WASM boundary to 644 lines, past the cohesion guideline. | normal | follow-up: split `core/wasm/index.ts` separately |

</details>


packages/babylon-tbv-rust-wasm/scripts/build-wasm.js
```diff
@@ -15,12 +15,16 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 // the binary that produces the frozen on-chain-binding secrets — re-run the
 // golden-vector gate (src/__tests__ frozen vectors) before shipping.
 const VAULT_WASM_REPO_URL = 'git@github.com:babylonlabs-io/vault-wasm.git';
-const VAULT_WASM_BRANCH = 'main';
-// main incl. PR #4 (tx graph v3 / Vault Core 3); bundles btc-vault
-// v1 @ 2c1177ec (tag v0.6.1), v2 @ 27c0062b (tag v0.8.0), v3 @ e1e50f66.
-// v3 pins the release/v0.9.x head — btc-vault has not tagged v0.9.0 yet;
-// re-pin once upstream moves it to the tag rev.
-const VAULT_WASM_COMMIT = '3accd8f614bab7b8018e40013322afe02b3ac80e';
+const VAULT_WASM_BRANCH = 'gai/delegated-claim-assembly-exports';
```
jeremy-babylonlabs: This pins to a draft branch head (`gai/delegated-claim-assembly-exports` @ `1c16496b`), whose bundled btc-vault rev is itself an open PR head. Merging to `main` like this makes the committed binary unreproducible as soon as either PR squash-merges.

Hold until vault-wasm#6 and btc-vault#2655 land, then re-pin `VAULT_WASM_BRANCH` / `VAULT_WASM_COMMIT` to main (or the v0.9.0 tag) and rebuild `dist/`.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/assembleWatchtowerArtifacts.ts
```diff
@@ -0,0 +1,236 @@
...
+
+export interface AssembleWatchtowerArtifactsParams {
+  /** Wallet holding the depositor key the graph was built with. */
+  btcWallet: BitcoinWallet;
+  /** Depositor's BTC public key (compressed or x-only hex). */
+  depositorPublicKey: string;
+  /** Graph and verifying key as the vault provider returned them. */
+  source: ClaimerArtifactsSource;
+  vault: DelegatedClaimVaultContext;
+  /**
```
jeremy-babylonlabs, line 85: Nothing checks that `source.txGraphJson` actually belongs to `vault.vaultId` — `build_watchtower_artifacts` in btc-vault#2655 only hex-normalises `vault_id_hex`. So a VP-served graph for a different vault produces a depositor-signed artifacts file stamped with this vault's id.

Before signing, take the Claim PSBT's input-0 prevout txid (the PegIn txid) and check `deriveVaultId(peginTxid, depositorEthAddress) === vault.vaultId` — or have the Rust builder derive and compare.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/assembleWatchtowerArtifacts.ts
```diff
@@ -0,0 +1,236 @@
...
+  );
+
+  let cursor = 0;
+  const claimSig = signatures[cursor++];
+  const assertClaimerSigHex = signatures[cursor++];
+  const payoutClaimerSigHex = signatures[cursor++];
+  const depositorPayoutSigHex =
+    presignedDepositorPayoutSig ?? signatures[cursor++];
+
+  const wronglyChallengedSigs: WronglyChallengedSigs = {};
```
jeremy-babylonlabs, line 197: This `catch` swallows `InvalidSignature` as well, contrary to the comment above it. Rust distinguishes `MissingSignature("... not set on the graph")` from `InvalidSignature` (bad sighash type, or fails verification).

The outcome is safe today because a fresh signature replaces it and the builder verifies, but a corrupted VP graph is masked. Only treat the `MissingSignature` case as "absent" and rethrow the rest.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/index.ts
```diff
@@ -0,0 +1,51 @@
...
+} from "./types";
+
+/**
+ * Claim-time execution. These finalizers take the signatures an
+ * `artifacts.json` already carries and return broadcastable transaction hex,
+ * so a claimer runs the whole path — Claim, Assert, Payout — without the
+ * watchtower CLI. Assert additionally needs a Groth16 proof from the prover
+ * service; check it with `verifyPegoutProof` before it reaches
+ * `finalizeAssert`.
+ */
```
jeremy-babylonlabs, line 84: `verifyPegoutProof` / `finalizeAssert` / `finalizePayout` / `finalizeWronglyChallenged` have no caller and no test, and this file's own header says claim-time execution stays with the watchtower CLI. They are also missing from `babylon-tbv-rust-wasm/src/index-node.ts`, so under the `node` export condition the `wasm/index.ts` forwarders resolve to `undefined` and throw `TypeError: ... is not a function`.

Suggest dropping them from `delegated-claim/index.ts`, `wasm/index.ts` and the browser export block until there is a caller — that also restores browser/Node export parity.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/__tests__/wotsKeypair.golden.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { describe, expect, it } from "vitest";
+
+import { wotsKeypairFromSeed } from "../../../wasm";
+
+// The fixed [0x42; 64] seed vault-wasm pins.
+const GOLDEN_SEED = new Uint8Array(64).fill(0x42);
+
+describe("wotsKeypairFromSeed frozen derivation", () => {
+  it("derives the pinned pk_hash for the golden seed", async () => {
```
jeremy-babylonlabs, line 23: Deposit-time `depositorWotsPkHash` comes from the existing TS `wots/blockDerivation.ts`; claim-time comes from wasm `wotsKeypairFromSeed`. The two golden vectors here pin different inputs ([0x42;64] vs root=0xaa) and nothing asserts the two derivations agree.

One equivalence test covers the actual on-chain binding: for the same 64-byte seed, `computeWotsBlockPublicKeysHash(await deriveWotsBlocksFromSeed(seed)) === (await wotsKeypairFromSeed(seed)).pk_hash`.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  vaultId: Hex;
+  /** Graph (vault core) version of the vault. Delegated claim requires 3. */
+  txGraphVersion: number;
+  proverCircuitVersion: number;
+  /**
+   * Block of the finalized `VaultClaimableBy` event. Pass `0n` when the
+   * withdrawal has not been initiated yet; whoever runs the claim must
+   * correct it from chain first.
+   */
+  claimableEventBlockNumber: bigint;
```
jeremy-babylonlabs, line 112: `claimableEventBlockNumber` is `bigint` here but `number` in `WatchtowerArtifactsSummary` — worth picking one (the wasm boundary takes `bigint`).

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/__tests__/assembleWatchtowerArtifacts.test.ts
```diff
@@ -0,0 +1,177 @@
...
+  buildPayoutClaimerPsbt: vi.fn(),
+  buildPayoutDepositorPsbt: vi.fn(),
+  buildWronglyChallengedPsbts: vi.fn(),
+  extractDepositorPayoutSig: vi.fn(),
+  extractTapScriptSig: vi.fn(),
+  finalizeClaimTx: vi.fn(),
+  buildWatchtowerArtifacts: vi.fn(),
+}));
+
+vi.mock("../../../wasm", async (importOriginal) => ({
```
jeremy-babylonlabs, line 36: These prove orchestration order only, since every wasm call is mocked. That is fine given byte-compat is pinned by btc-vault#2655's `artifacts.rs` round-trip test — worth noting that dependency in the file header so the next reader knows where the format guarantee actually lives.

jeremy-babylonlabs (dismissed):
I checked the artifact format against btc-vault#2655 and it matches — byte-compat is pinned Rust-side by the `artifacts.rs` round-trip test, the committed `.wasm` is built from `729fc3da` (that PR's head), and the WOTS golden `b0ad6df0...` agrees with `crates/crypto/src/wots/protocol.rs`. Good.

Two things block merge: the build pin, and a missing bind between the VP-served graph and the vault id. Also, `verify` is red on stale generated API docs — run `pnpm --filter @babylonlabs-io/ts-sdk run docs:clean` and commit.

jonybur (commented):
One small cleanup on top of Jeremy's block: the `toError` extraction into `errors.ts` is only wired up on the browser entry. `packages/babylon-tbv-rust-wasm/src/index-node.ts` still carries a private `function toError` at line 458, so after this PR the browser entry uses the shared helper while the Node entry keeps its own copy of the same normalizer. Fold `index-node.ts` onto the same `import { toError } from './errors.js';` used by `index.ts` so the two entry points stay in step and the dead copy goes away.

packages/babylon-tbv-rust-wasm/scripts/build-wasm.js
```diff
@@ -15,12 +15,16 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
...
+// vault-wasm PR #6 (delegated-claim assembly exports, graph v3 only) on top
+// of PR #4 (tx graph v3 / Vault Core 3); bundles btc-vault
+// v1 @ 2c1177ec (tag v0.6.1), v2 @ 27c0062b (tag v0.8.0), v3 @ 729fc3da.
+//
+// TEMPORARY PIN. Both upstream PRs are open: vault-wasm#6 is a draft and its
+// v3 rev is btc-vault#2655's head, which becomes unreachable once that PR
+// squash-merges. Re-pin to vault-wasm main (or the v0.9.0 tag) before this
+// package is published.
+const VAULT_WASM_COMMIT = '1c16496bd7d3a8e1eeb664816eef0a2930a6d6b4';
```
gbarkhatov: **Merge-blocker.** The committed `dist/generated/*` wasn't built from this pin, so "v3 @ 729fc3da" at line 21 doesn't describe the binary.
- The debug strings embedded in `vault_wasm_bg.wasm` show v1 and v2 compiled from cargo git checkouts (`…/checkouts/btc-vault-…/2c1177e`, `…/27c0062`). v3 was compiled from a local working tree (`…/btc-vault-claim/crates/vault/src`, `…/crates/crypto/src`).
- The generated `vault_wasm.d.ts` declares `finalizeAssert`, `finalizePayout`, `finalizeWronglyChallenged` and `verifyPegoutProof`. `git grep` for any of them, under any naming, at vault-wasm@`1c16496b` and at btc-vault@`729fc3da` finds nothing: both export only `finalizeClaimTx`.

Why this matters:
- The first honest rebuild drops those four exports, and the wrappers in `delegatedClaim.ts` stop type-checking.
- More importantly, the code that actually runs `verify_bundle`, the only gate on every §8 signature in this flow, has no reviewable source.
- CLAUDE.md §4 treats a pin bump as a hard-fork-class change precisely so the binary is reproducible from the pin. Here the only thing bounding the frozen expanders in the shipped binary is `expand.test.ts` plus the new golden vector staying green.

Fix: rebuild from `VAULT_WASM_COMMIT` on a clean checkout and delete the four wrappers and exports, or land them upstream and re-pin. Either way, a second person should regenerate `dist/generated/` from the pin and confirm it's byte-identical before merge.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/assembleWatchtowerArtifacts.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  // Challenger order is fixed here and reused when the signatures are mapped
+  // back, so a wallet that reorders nothing keeps every signature with the
+  // challenger and garbled-circuit index it was computed for.
+  const challengerPubkeys = Object.keys(wronglyChallengedPsbts);
+  for (const challengerPubkey of challengerPubkeys) {
+    for (const psbtBase64 of wronglyChallengedPsbts[challengerPubkey]) {
+      requests.push({ psbtBase64, inputIndex: WRONGLY_CHALLENGED_INPUT });
+    }
+  }
```
gbarkhatov, line 182: **Merge-blocker under the CLAUDE.md §3 rule** ("Never sign a value or accept a challenger key handed to us verbatim"). When the graph has no stored depositor Payout signature, which includes when `readPresignedDepositorPayoutSig` swallows an invalid one, line 122 has the wallet sign `buildPayoutDepositorPsbt(graph)`. That's a fresh Payout spending the PegIn vault UTXO, built entirely from the VP-served graph. Lines 130-135 then sign one WronglyChallenged PSBT per key the graph lists.

Neither gets the checks the deposit path applies:
- `buildPayoutPsbt` validates against `registeredPayoutScriptPubKey` per role.
- `deriveLocalChallengers` + `assertChallengerSetMatchesExpected` (`signDepositorGraph.ts`) pin the challenger set.

Nothing under `delegated-claim/` references either. The Rust `verify_bundle` (btc-vault@`729fc3da` `crates/vault/src/delegated_claim.rs`) doesn't cover this either. It verifies every signature, but only against keys carried by the same graph JSON (including `verify_presigned_signatures`), it never compares the Payout outputs with a registered destination, and it runs after the wallet has already signed. `docs/delegated_claim.md` rests leak-safety on "funds still go to the designated payout address", and this is the one path where that isn't enforced.

Broadcasting a redirected Payout still needs matching presigs from the real co-signers, so this defends against VP plus co-signer collusion, which is what §3 exists for. The impact of the challenger-set half is probably bounded by the same N-of-N presigs; I haven't verified that.

Fix: before signing, parse the WASM-built Payout PSBT and assert `outs[0].script` equals the vault's on-chain registered payout scriptPubKey (reuse the `buildPayoutPsbt` role validation). Export and reuse `assertChallengerSetMatchesExpected` on `Object.keys(wronglyChallengedPsbts)`. And treat an *invalid* stored signature as an error rather than falling through to fresh signing.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/readWatchtowerArtifacts.ts
```diff
@@ -0,0 +1,164 @@
...
+    throw new ArtifactsVaultMismatchError(expected, actual);
+  }
+
+  await verifyWatchtowerArtifacts(
+    params.txGraphVersion ?? DELEGATED_CLAIM_TX_GRAPH_VERSION,
+    params.artifactsJson,
+  );
+
+  return summary;
+}
```
gbarkhatov, line 209: This is the read-side twin of the graph ↔ vault-id point in the existing review, and fixing it at assembly time doesn't cover it. The doc says this "confirms it is the one for this vault", but it only string-compares `expectedVaultId` with the file's self-declared `vault_id`. The Rust `verify_watchtower_artifacts` just runs `normalize_vault_id_hex` on that field and never relates it to `graph.pegin_tx`. So a file whose graph belongs to vault B, but whose `vault_id` says A, passes. That can come from assembly, a hand-edited file, or any file the SDK didn't write.

Fix: `summarizeWatchtowerArtifacts` already parses `claim_tx`. Take input 0's prevout txid (the PegIn txid), run it through `deriveVaultId` with the depositor's ETH address, and compare that instead.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/deriveClaimerWotsKeypair.ts
```diff
@@ -0,0 +1,83 @@
...
+  await validateWotsKeypairAgainstGraph(
+    params.txGraphVersion,
+    derivation.keypair,
+    params.txGraphJson,
+  );
+
+  return {
+    wotsKeypairJson: JSON.stringify(derivation.keypair),
+    pkHash: derivation.pk_hash,
+  };
```
gbarkhatov, line 110: §1: `derivation.pk_hash` comes back from WASM and is returned unchecked, though the JSDoc at line 40 says it matches `depositorWotsPkHash`. The only gate is `validateWotsKeypairAgainstGraph`, which compares against the VP-served graph, and that graph is unbound (the existing review's point). The on-chain `depositorWotsPkHash` is the independent, VP-free anchor. A wrong `htlcVout`, the wrong wallet account, or expander drift after a re-pin (see the build-pin comment) would pass against a bad graph and only fail after the Claim is broadcast. Suggested fix: take the on-chain hash as a parameter and assert equality before returning. That's a runtime check, separate from the requested TS↔WASM test.

packages/babylon-ts-sdk/src/tbv/core/utils/signing.ts
```diff
@@ -11,10 +11,16 @@ import type { SignPsbtOptions } from "../../../shared/wallets/interfaces";
...
+ *   A wallet derives a key-path address from `publicKey` and compares it with
+ *   the input's address, so `publicKey` alone cannot sign an input that sits
+ *   at a different address — every script-path connector output. `address`
+ *   names the account instead and signs those inputs.
  */
 export function createTaprootScriptPathSignOptions(
   publicKey: string,
   inputCount: number,
+  address?: string,
 ): SignPsbtOptions {
```
gbarkhatov: The new `address?` parameter on `createTaprootScriptPathSignOptions` has no caller: all 7 production callers (PayoutManager ×2, PeginManager, signDepositorGraph ×2, refund, reclaim) pass two arguments. That's dead code under the zero-dead-code rule.

In the new `…ForInput` helper, `address ? { address } : { publicKey }` plus a non-empty `getAddress()` in `signAndExtract` means `depositorPublicKey` never reaches the wallet on the delegated-claim path. A wallet connected to the wrong account signs all N+3 PSBTs, and the mismatch only surfaces afterwards in `finalizeClaimTx` / `verify_bundle`.

Suggested fix: drop the unused parameter here, and in `assembleWatchtowerArtifacts` compare `btcWallet.getPublicKeyHex()` with `depositorPublicKey` before prompting.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/__tests__/assembleWatchtowerArtifacts.test.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/** A wallet that signs a batch and returns the PSBTs unchanged, in order. */
+function makeWallet(): BitcoinWallet {
+  return {
+    getAddress: vi.fn(() => Promise.resolve(SIGNER_ADDRESS)),
+    signPsbt: vi.fn(),
+    signPsbts: vi.fn((psbtHexes: string[]) => Promise.resolve(psbtHexes)),
+  } as unknown as BitcoinWallet;
+}
```
gbarkhatov, line 120: §8 flags are untested on this path. The mock `signPsbts` ignores its second argument, `createTaprootScriptPathSignOptionsForInput` has no unit test, and `signing.test.ts` is untouched. Dropping `useTweakedSigner: false` or `autoFinalized: false`, or signing the wrong `index`, keeps the suite green. Asserting on `wallet.signPsbts.mock.calls[0][1]`, plus a few helper tests in `signing.test.ts`, would pin them.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/readWatchtowerArtifacts.ts
```diff
@@ -0,0 +1,164 @@
...
+  );
+  // Absent on files written before the field existed; the CLI reads an
+  // absent value as 0, which means "not yet known from chain".
+  const claimableEventBlockNumber =
+    parsed.claimable_event_block_number === undefined
+      ? 0
+      : requireNumber(
+          parsed.claimable_event_block_number,
+          "claimable_event_block_number",
+        );
```
gbarkhatov, line 107: Low. An absent `claimable_event_block_number` becomes `0`, and `assertArtifactsUsableForVault` accepts it. But btc-vault@`729fc3da` `docs/delegated_claim.md` §4 says a file with a zero block number makes remote `start-claim` request a proof for the wrong block and fail before Assert. It fails safe, but a function named "usable" should reject 0, or return an explicit incomplete state. Nothing in the SDK fills that field in later.

packages/babylon-tbv-rust-wasm/src/delegatedClaim.ts
```diff
@@ -0,0 +1,386 @@
...
+          inputs.assertClaimerSigHex,
+          inputs.payoutClaimerSigHex,
+          JSON.stringify(inputs.wronglyChallengedSigs),
+          inputs.depositorPayoutSigHex,
+          inputs.verifyingKeyHex,
+          inputs.claimableEventBlockNumber,
+          inputs.proverCircuitVersion,
+          inputs.vaultIdHex,
+          inputs.babeSessionsJson,
+        );
```
gbarkhatov, line 336: Low, §1. `claimableEventBlockNumber` (`bigint` → u64) and `proverCircuitVersion` (`number` → u16) cross into WASM with no range guard, and Rust copies both into the file verbatim. Under wasm-bindgen's conversions an out-of-range value wraps or truncates rather than throwing (for example, `65543` becomes `7` for a u16), which yields a well-formed file carrying the wrong circuit version. Callers source both on-chain today, but a guard like the ones in `value-guards.ts` would make that an invariant.

packages/babylon-tbv-rust-wasm/src/delegatedClaim.ts
```diff
@@ -0,0 +1,386 @@
...
+ * Parses a JSON string a wasm export returned. A failure here is a
+ * binary/facade mismatch, not caller error, so it carries the export name.
+ */
+function parseWasmJson<T>(json: string, fnName: string): T {
+  try {
+    return JSON.parse(json) as T;
+  } catch (err) {
+    throw toError(err, `${fnName}: unparseable JSON from WASM`);
+  }
+}
```
gbarkhatov, line 76: Low. `toError` (`errors.ts`) returns any `Error` unchanged, so the promised export name never makes it onto a `JSON.parse` `SyntaxError`. And V8's JSON `SyntaxError` message quotes a snippet of its input. For `wotsKeypairFromSeed` (line 204) that input is the secret keypair JSON, so a binary/facade mismatch (made more plausible by the build-pin issue) could put WOTS secret material into an error message and from there into logs. Suggested fix: throw a fresh `Error(`${fnName}: unparseable JSON from WASM`)` with no cause text.

gbarkhatov (dismissed):
**Headline:** two new merge-blockers on top of the existing CHANGES_REQUESTED review (which I don't repeat here; its points and their status at HEAD are in the table below).

1. **The committed engine isn't built from the pin.** The v3 sources embedded in `vault_wasm_bg.wasm` resolve to a local working tree (`…/btc-vault-claim/crates/{vault,crypto}/src`), not the cargo git checkout that v1 (`2c1177e`) and v2 (`27c0062`) come from. The binary also exports `finalizeAssert`, `finalizePayout`, `finalizeWronglyChallenged` and `verifyPegoutProof`. Neither vault-wasm@`1c16496b` nor btc-vault@`729fc3da` defines any of the four: both have only `finalizeClaimTx`. Details inline on `build-wasm.js`.
2. **The claim path signs a VP-built Payout of the vault UTXO without the CLAUDE.md §3 check.** Details inline on `assembleWatchtowerArtifacts.ts`.

Also still red: `verify` (stale generated API docs).

**The existing review's points at HEAD:**

| Point | Status |
|---|---|
| Build pin is a draft-branch head | Still present, and worse in substance: the binary doesn't reproduce from the pin even today (blocker 1). Its body says "the committed `.wasm` is built from `729fc3da`"; the embedded source paths and the four extra exports say otherwise. |
| VP-served graph not bound to the vault id | Still present. It has a read-side twin in `assertArtifactsUsableForVault` (inline), and binding the graph to the vault id alone doesn't bind the Payout outputs or the challenger set (blocker 2). |
| Swallowed error in `readPresignedDepositorPayoutSig` | Still present (bare `catch { return null; }`), and worse than described: a corrupt stored signature routes into blocker 2's unchecked fresh-signing path. |
| Finalizers unused / Node export gap | Still present. The four finalizers also exist at no published upstream revision (blocker 1), and `delegatedClaim.ts:8-10` itself says claim-time execution isn't part of this surface. |
| No TS↔WASM WOTS equivalence test | Still present. |
| `claimableEventBlockNumber` bigint vs number | Still present. |
| Tests all mocked; header should cite the Rust round-trip | Still present. |

**Refuted** (considered and ruled out):
- *"The pin bump changes a frozen §4 expander output."* I found no change. The diff doesn't touch the three expander wrappers, `expand.test.ts` is untouched and green in CI, and the new golden `0xb0ad6df0…` for seed `[0x42;64]` matches btc-vault@`729fc3da` `crates/crypto/src/wots/protocol.rs`. Given blocker 1, those green vectors are the only real gate on the shipped binary.
- *"The claimer WOTS key could collide with the depositor's."* It's the same key by design. `deriveClaimerWotsKeypair` reuses the frozen `expandWotsSeed(root, htlcVout)`, and `docs/delegated_claim.md` calls `wots_keypair.json` the depositor's WOTS private key. There's no domain separation to have.
- *"`PAYOUT_CLAIMER_INPUT = 0` would pass every test."* `extractTapScriptSig` receives `requests[i].inputIndex`, so the test fails.
- *"The unchecked `as T` in `parseWasmJson` is a §1 hole."* Any drift fails in `verify_bundle` before a file is produced.

**Verified clean** (against the btc-vault@`729fc3da` source):
- §8 holds in Rust. Every signature produced with the non-standard flags is verified before a file is produced: Claim through its witness and script checks; Assert, Payout-claimer, depositor Payout (stored or fresh) and every WronglyChallenged signature in `verify_bundle`, with SIGHASH_DEFAULT enforced and full challenger coverage. Caveat: blocker 1 means the shipped binary isn't that source.
- The 7 existing `createTaprootScriptPathSignOptions` callers behave unchanged; all take the `publicKey` branch.
- The WronglyChallenged signature mapping keeps each signature with its challenger and garbled-circuit index.
- `deriveClaimerWotsKeypair` zeroes both the root and the seed in `finally`, and logs nothing.
- The §9 lazy boundary holds: the new forwarders go through `loadTbvWasm()`, and `delegatedClaim.ts` imports the bindings type-only.

Test-quality note, no inline comment: `deriveClaimerWotsKeypair.test.ts` mocks `wotsKeypairFromSeed` and uses `htlcVout: 0`, so hardcoding 0 or skipping `expandWotsSeed` would still pass. The golden test's `toBeTruthy()` checks also pass for `{}`.

packages/babylon-tbv-rust-wasm/src/delegatedClaim.ts
```diff
@@ -0,0 +1,386 @@
...
+     * BaBe sessions are multi-hundred-megabyte payloads; omit them here and
+     * join them into the file downstream rather than routing them through
+     * WASM memory.
+     */
+    async buildWatchtowerArtifacts(
+      inputs: WatchtowerArtifactsInputs,
+    ): Promise<string> {
+      const wasm = await getWasmBindings();
+      try {
+        return wasm.buildWatchtowerArtifacts(
```
jrwbabylonlab, line 319: This wraps the previous 12-argument API. vault-wasm PR #6 (head 89334b38) makes depositor_payout_sig_hex and babe_sessions_json required, adds expected_vault_core_version, and removes extractDepositorPayoutSig, and btc-vault #2655 says verifying_key_hex must never come from the vault provider (here it is taken from the VP response and passed through). Re-pinning to upstream is a redesign of this facade, not a pin bump.

packages/babylon-tbv-rust-wasm/scripts/build-wasm.js
kirugan: Correct, and thank you for digging into the debug strings — the binary was built from a local working tree, not from the pin.

Fixed by re-pinning and rebuilding through the script itself:

- `VAULT_WASM_COMMIT` is now `89334b38`, the current head of vault-wasm#6, whose `btc_vault_v3` is a pushed rev (`ff714392`). The comment at line 21 names it.
- `dist/generated/*` was regenerated with `npm run build-wasm`, so it comes from that clone rather than from anything on my disk.

The four exports you could not find now exist upstream, under different names: vault-wasm#6 took the claim-time surface in and reshaped it to be artifacts-in / artifacts-out. The binary exports `pinPegoutProof`, `attachFinalizedAssert`, `finalizePayout`, `finalizeWronglyChallenged` and `extractGraphSummary`; `finalizeAssert`, `verifyPegoutProof` and `extractDepositorPayoutSig` are gone. The facade was rewritten to match.

One caveat so this is not read as fully closed: the pin is still an open PR head, which is Jeremy's separate point above. It is reproducible now, but it must be re-pinned to main or the v0.9.0 tag before this package is published.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/assembleWatchtowerArtifacts.ts
kirugan: Right, the `catch` was wider than the comment claimed.

It is gone, along with `readPresignedDepositorPayoutSig`. vault-wasm#6 removed `extractDepositorPayoutSig` and made `depositor_payout_sig_hex` required, so the depositor Payout signature is now always signed fresh and its PSBT always joins the same batch. There is no longer an error to classify — a corrupted VP graph fails in `verify_bundle` instead of being masked.

The batch is 53 PSBTs rather than 52 as a result.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/types.ts
kirugan: Picked `bigint`, matching the WASM boundary.

`WatchtowerArtifactsSummary.claimableEventBlockNumber` is now `bigint`, so it agrees with `DelegatedClaimVaultContext`. `summarizeWatchtowerArtifacts` converts on parse and the two `readWatchtowerArtifacts` tests assert `10_985_680n` / `0n`.

packages/babylon-ts-sdk/src/tbv/core/utils/signing.ts
kirugan: Both points fixed.

- The unused `address?` parameter is off `createTaprootScriptPathSignOptions`. All 7 production callers pass two arguments, so it was dead. It stays on `…ForInput`, which the delegated-claim path does use, and the JSDoc there now says why an address rather than a public key: a wallet derives a key-path address from the key and refuses any input sitting elsewhere, which is every script-path connector output. UniSat refuses input 1 of the claimer Payout on `publicKey` and signs it on `address` — that is what this parameter exists for.
- `assembleWatchtowerArtifacts` now calls `assertWalletMatchesDepositor` before building any request: it compares `btcWallet.getPublicKeyHex()` with `depositorPublicKey` in x-only form and throws before the first prompt. A wallet on the wrong account fails immediately rather than after signing all N+3 PSBTs. Covered by a new test asserting `signPsbts` is never called.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/__tests__/assembleWatchtowerArtifacts.test.ts
kirugan: Added, in both places you named.

`assembleWatchtowerArtifacts.test.ts` now asserts on `wallet.signPsbts.mock.calls[0][1]`: every option carries `autoFinalized: false` and `useTweakedSigner: false`, each names exactly one input, and the index sequence is `[0, 0, 1, 0, 0, 0, 0]` — the claimer Payout at input 1, everything else at input 0. Dropping a flag or moving an index now fails the suite.

`signing.test.ts` gained 5 tests for `createTaprootScriptPathSignOptionsForInput`, including that an address replaces the public key rather than accompanying it.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/readWatchtowerArtifacts.ts
kirugan: Agreed — a function named "usable" should not pass that on.

`assertArtifactsUsableForVault` now throws when `claimableEventBlockNumber` is `0n`, and the message says the file was assembled before the Ethereum withdrawal was initiated and must be assembled again. `summarizeWatchtowerArtifacts` still reports `0n` without judging it, since it only describes the file.

packages/babylon-tbv-rust-wasm/src/delegatedClaim.ts
kirugan: Guarded. `assertU64` and `assertU16` in `delegatedClaim.ts` run before the call, so `claimableEventBlockNumber`, `proverCircuitVersion` and the new `expectedVaultCoreVersion` throw instead of wrapping. `gcIndex` on `finalizeWronglyChallenged` gets the same treatment.

Your `65543` → `7` example is the exact failure mode: a well-formed file carrying the wrong version, discovered much later.

packages/babylon-tbv-rust-wasm/src/delegatedClaim.ts
kirugan: Good catch — that path could put WOTS secret material into a log.

`parseWasmJson` no longer passes the cause through. It throws a fresh `Error("<fnName>: unparseable JSON from WASM")` and drops the `SyntaxError`, so V8's quoted input snippet never escapes. The comment records why, since the obvious "improvement" is to re-attach the cause.

packages/babylon-tbv-rust-wasm/src/delegatedClaim.ts
kirugan: You were right that this is a redesign rather than a pin bump. Done.

Re-pinned to vault-wasm#6 head `89334b38` and rewrote the facade to the new shape:

| was | now |
|---|---|
| `verifyPegoutProof(v, vk, proof)` | `pinPegoutProof(v, artifactsJson, proof)` |
| `finalizeAssert(v, graph, sig, kp, vk, proof)` | `attachFinalizedAssert(v, artifactsJson, kp)` |
| `finalizePayout(v, graph, sig, sig?)` | `finalizePayout(v, artifactsJson)` |
| `finalizeWronglyChallenged(6 args)` | 5 args, artifacts-based |
| `buildWatchtowerArtifacts(12 args)` | 13 — `expectedVaultCoreVersion` added |
| `extractDepositorPayoutSig` | removed |

`depositorPayoutSigHex` and `babeSessionsJson` are required upstream: the signature is always signed fresh (batch is 53 PSBTs now), and an omitted sessions argument is passed as `{}` since the sessions are joined into the file downstream.

`expectedVaultCoreVersion` comes from the finalized `PegInSubmitted` event, so `DelegatedClaimVaultContext` gained a `vaultCoreVersion` field — a public type change worth noting in review.

On `verifying_key_hex` coming from the VP response: still true on this branch, and not addressed here. Raising it as its own thread is fair.

packages/babylon-tbv-rust-wasm/scripts/build-wasm.js
kirugan: Half of this is fixed; your actual request is not, so leaving the thread open.

Fixed: `dist/generated/*` was not built from the pin at all — Govard's thread has the evidence. It is now regenerated with `npm run build-wasm` from `VAULT_WASM_COMMIT`, which moved to `89334b38` (current head of vault-wasm#6, whose `btc_vault_v3` is the pushed rev `ff714392`).

Not fixed: it is still a draft branch head, so the commit stays unreachable if that PR squash-merges. The re-pin to main or the v0.9.0 tag, plus a rebuild, still has to happen before this lands. Happy to keep this blocking until then.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/index.ts
kirugan: The Node half is fixed; on dropping them I did the opposite of what you suggested, so this is yours to judge.

Fixed: `index-node.ts` was missing them, so the `node` condition really did resolve to `undefined`. The export block now matches the browser entry, and the lazy-facade check passes for browser, node and raw.

Not dropped, and here is the argument. The four functions are the claim itself: with them a depositor holding `artifacts.json` and `wots_keypair.json` runs Claim, Assert and Payout from a browser with no CLI and no vault provider. That was tested end to end on signet against vault `0xdc8faa2e…` with the VP's `vaultd` stopped — Claim `c75d53a5…`, Assert `ea86ca34…` (16,263 bytes) and Payout `94ed0358…` all built in the tab and confirmed, and the indexer shows the vault redeemed.

The header saying claim-time execution stays with the CLI is now out of date: vault-wasm#6 took this surface in and reshaped it around the artifacts file. I have updated the header to describe what the module does.

On "no caller and no test": true in this repo. The caller is a browser page that was deliberately kept out of the PR. If you would still rather they land with a caller, the honest options are to bring that page in as an example, or to hold these four exports for a follow-up. Your call — say which and I will do it.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/assembleWatchtowerArtifacts.ts
```diff
@@ -0,0 +1,247 @@
...
+  requests: PsbtSigningRequest[],
+): Promise<string[]> {
+  // The sign options name the signer by address, because a wallet derives a
+  // key-path address from a public key and then refuses any input that sits
+  // elsewhere — every script-path connector. An address names the account, so
+  // it must be proved to be this depositor's account first: otherwise a wallet
+  // on the wrong account signs all N+3 PSBTs and the mismatch only surfaces
+  // later, in finalizeClaimTx or verify_bundle.
+  await assertWalletMatchesDepositor(btcWallet, depositorPublicKey);
+  const signerAddress = await btcWallet.getAddress();
```
jonybur: **The wallet address used to name the signer is never checked against the depositor key, although two comments say it is.**

`assertWalletMatchesDepositor` only compares `getPublicKeyHex()` with `depositorPublicKey`. The next line takes `btcWallet.getAddress()` unchecked and passes it into every sign option, and `createTaprootScriptPathSignOptionsForInput` then sends `address` instead of `publicKey` (signing.ts:68). The comment at lines 211-216 and the JSDoc at signing.ts:49-50 both say the address is proved to belong to the depositor first. Nothing does that. A wallet whose reported address and public key belong to different accounts passes the check and signs the whole batch for the wrong account. The mismatch only shows up afterwards in `finalizeClaimTx` or the artifacts verification, which is the late failure this check was added to prevent. Derive the depositor's P2TR address from `depositorPublicKey` (for example with `isAddressFromPublicKey` in primitives/utils/bitcoin.ts) and compare it with the wallet address before prompting, or correct both comments.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/types.ts
```diff
@@ -0,0 +1,58 @@
...
+  proverCircuitVersion: number;
+  /**
+   * Vault Core version from the finalized `PegInSubmitted` event. The builder
+   * refuses a graph that records a different one, which is what stops a graph
+   * built under other rules from being signed.
+   */
+  vaultCoreVersion: number;
+  /**
+   * Block of the finalized `VaultClaimableBy` event. Pass `0n` when the
+   * withdrawal has not been initiated yet; whoever runs the claim must
```
jonybur: **The context doc says to pass 0n for the block number and correct it later, but the SDK's usability check rejects 0n and nothing in the SDK can correct it.**

`DelegatedClaimVaultContext.claimableEventBlockNumber` tells callers to pass `0n` before the withdrawal starts, and says whoever runs the claim must correct it from chain first. The assembly path accepts `0n` (`assertU64` allows it) and writes it into the file. `assertArtifactsUsableForVault` (readWatchtowerArtifacts.ts:135) then rejects any such file and says to assemble it again, and its own comment notes that nothing in the SDK fills the field in later. The module header recommends assembling while the vault provider is still online, for a claim months later. So files built as documented fail the SDK's own check, and the suggested fix needs the provider's graph and a new signing session, which may no longer be available. Either reject `0n` at assembly and update the doc, or add an SDK step that sets the block number on an existing file from the finalized event.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/readWatchtowerArtifacts.ts
```diff
@@ -0,0 +1,178 @@
...
+  const proverCircuitVersion = requireNumber(
+    parsed.prover_circuit_version,
+    "prover_circuit_version",
+  );
+  // Absent on files written before the field existed; the CLI reads an
+  // absent value as 0, which means "not yet known from chain".
+  const claimableEventBlockNumber =
+    parsed.claimable_event_block_number === undefined
+      ? 0n
+      : BigInt(
```
jonybur, line 102: **summarizeWatchtowerArtifacts does not validate untrusted fields: negative or fractional block numbers, vault_core_version and babe_sessions.**

`requireNumber` accepts any finite number, and the only later check is `=== 0n`. A file with `claimable_event_block_number: -1` passes `assertArtifactsUsableForVault`'s own checks, `1.5` makes `BigInt()` throw a bare RangeError that doesn't name the field, and a value above 2^53 has already been rounded by `JSON.parse`. On line 92, `vault_core_version` is copied into a field typed `number` without a type check. On line 97, a non-object `babe_sessions` such as a string makes `Object.keys` report "0", "1", … as challenger public keys. This function exists to check files that come from disk, so require a positive safe integer for the block number, check `vault_core_version` with `requireNumber`, and check that `babe_sessions` is a plain object.

packages/babylon-tbv-rust-wasm/src/delegatedClaim.ts
```diff
@@ -0,0 +1,397 @@
...
+      try {
+        return wasm.buildPayoutClaimerPsbt(txGraphVersion, graphJson);
+      } catch (err) {
+        throw toError(err, 'buildPayoutClaimerPsbt');
+      }
+    },
+
+    /**
+     * Depositor's Payout signing PSBT (base64) — input 0, the PegIn UTXO
+     * spend. Only needed when the graph carries no presign-phase depositor
```
jonybur: **Comments reference the removed extractDepositorPayoutSig and still say claim-time execution is outside a surface that now exports the finalizers.**

The `buildPayoutDepositorPsbt` JSDoc says the PSBT is only needed when the graph carries no presigned depositor Payout signature, and points to `extractDepositorPayoutSig`. That export was removed, and this PSBT is now always signed. The file header at lines 8-10, `src/index.ts:493-495`, `src/index-node.ts:573-575` and `packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts:401-403` still say claim-time execution is not part of this surface, or not in the engine. Each of those files exports `pinPegoutProof`, `attachFinalizedAssert`, `finalizePayout` and `finalizeWronglyChallenged`. Only `services/delegated-claim/index.ts` was updated. A reader would conclude the depositor Payout signature is optional and that the finalizers don't exist. Rewrite these comments to match the current API.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/__tests__/deriveClaimerWotsKeypair.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    await derive();
+
+    expect(wasm.validateWotsKeypairAgainstGraph).toHaveBeenCalledWith(
+      3,
+      { blocks: [["00"]] },
+      "{graph}",
+    );
+  });
+
+  it("returns nothing when the keypair does not match the graph", async () => {
```
jonybur: **No test checks that the vault root and WOTS seed are zeroed, so removing the fill(0) calls keeps the suite green.**

`deriveClaimerWotsKeypair` zeroes `root` and `wotsSeed` in `finally` blocks (deriveClaimerWotsKeypair.ts:62-64 and 69-71). That is the only secret cleanup on this path. No test in this file checks it, on success or when `expandWotsSeed` or `wotsKeypairFromSeed` throws. Deleting either `fill(0)` would ship silently. Capture the buffers passed to the mocked `expandWotsSeed` and `wotsKeypairFromSeed`, and assert they are all zeros after the call, on both the success and the error path.

jonybur (commented):
Comments inline.

gbarkhatov (approved):

gbarkhatov (dismissed):

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/assembleWatchtowerArtifacts.ts
kirugan: Taking the two halves separately.

**Payout destination.** You are right that nothing here pins it, and the graph ↔ vault-id thread does not cover it: a graph that is correctly bound to this vault can still carry a Payout that pays elsewhere. What it does change is the attacker model. With the binding in place the VP can no longer swap in a graph for another vault, so a redirected Payout has to be built for *this* vault and still needs matching presigs from the real co-signers. That is VP + co-signer collusion, which is exactly what §3 is written for, so I am not arguing the check away.

I am not adding it in this PR, for one concrete reason: this path has no `registeredPayoutScriptPubKey` and no role to validate it under. `buildPayoutPsbt` gets both from the deposit context; here the caller has a graph, a vault id and a wallet. Adding the check means threading the registered script in from chain and deciding which output of which of the two Payout PSBTs it applies to, and I am not willing to guess that output layout on a signing path. Filing it as a follow-up with the layout confirmed against `crates/vault` rather than shipping a check that asserts the wrong output.

**Challenger set.** `deriveLocalChallengers` needs the on-chain VK list, which this path also does not have. Same follow-up. Note the asymmetry is milder here than at deposit: an extra WronglyChallenged signature is a signature for a key the protocol will not recognise, not a missing one, and `verify_bundle` re-checks every entry against the graph before the claim runs.

Resolving so the thread does not block, with both items carried into the follow-up issue. Reopen if you would rather hold the PR for them.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/index.ts
kirugan: Keeping the four exports, and keeping them in this PR.

The Node gap you found is fixed — `index-node.ts` exports them, browser and Node are at parity, and the lazy-facade check passes for all three entries. That was the real bug in this thread.

On dropping them: they are the claim. Without them an `artifacts.json` is a file nobody in this repo can act on, and the whole point of the feature is a depositor whose vault provider is gone. Removing them and re-adding them behind a follow-up costs a second WASM re-pin cycle for no safety gain, since the exports are inert until something calls them.

"No caller in this repo" is fair and I am not disputing it. The caller exists — the browser page that ran the signet end-to-end — and I will bring it in as an example in a follow-up rather than growing this PR further. The header no longer claims claim-time execution stays with the CLI, and the same stale claim in `delegatedClaim.ts`, `src/index.ts`, `src/index-node.ts` and `wasm/index.ts` is corrected in this round.

Resolving. Say the word if you want the exports held back and I will strip them in a separate commit.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/types.ts
kirugan: Correct contradiction, and neither of your two options is quite right, so here is the third.

Rejecting `0n` at assembly would break the feature. The `VaultClaimableBy` event lands when the Ethereum withdrawal starts, which is the moment the claim begins; assembly is meant to happen long before that, while the vault provider is still online. Forcing assembly after the event means assembling exactly when the provider may be gone, which is the case this module exists for.

So `0n` stays accepted, and the doc no longer tells a story the code does not support. `DelegatedClaimVaultContext.claimableEventBlockNumber` now says plainly: zero is accepted, a file written that way is not claimable as written, nothing in the SDK fills it in later, pass the real block whenever the event has already finalized.

Your second option is the right fix and it is cheaper than it looks: the field is file metadata, covered by no sighash, so setting it on an existing file needs no wallet and no graph — parse, set, re-verify with `verifyWatchtowerArtifacts`. I am not adding a function that rewrites a signed artifacts file as a side change to this PR; it wants its own review, with the question of whether the rewrite belongs in Rust next to `build_watchtower_artifacts` rather than in TS. Follow-up issue, filed together with the two items from the §3 thread.

Resolving on the doc fix.

packages/babylon-tbv-rust-wasm/scripts/build-wasm.js
kirugan: Done in 29501910, and this is now what you asked for rather than half of it.

vault-wasm#6 merged into main on 2026-09-17, so `VAULT_WASM_BRANCH` is `main` and `VAULT_WASM_COMMIT` is `4cd516e85719ad7be30ed45c81945100bce7ab15` — the merge commit, which is also main head. Its `btc_vault_v3` moved from the PR head `ff714392` to `ac4954e7`, which is btc-vault#2655's merge commit on main (main is 10 commits past it, so it is an ancestor, not a dangling head). v1 `2c1177ec` and v2 `27c0062b` are unchanged and tagged. No tag exists on vault-wasm yet, so the commit is the pin; every rev is reachable from a default branch and the binary is reproducible.

`dist/generated` rebuilt from that pin: `vault_wasm_bg.wasm` `e2dce7c4…` → `a677ced5…`. The generated `.js` and `.d.ts` came back byte-identical, so the exported surface did not move.

Golden-vector gate re-run, as CLAUDE.md §4 requires on any bump: `vault-secrets/__tests__/expand.test.ts`, `context.golden.test.ts`, `wots/blockDerivation.test.ts` and `wotsKeypair.golden.test.ts` all pass unchanged. The three expanders and the WOTS derivation produce the same bytes across the re-pin, so nothing rotates and no in-flight deposit is affected. Full SDK suite 2052 passing, `check-lazy-entries` clean for browser, node and raw.

The comment at the pin no longer says TEMPORARY; it records all three bundled revs and states each is reachable.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/assembleWatchtowerArtifacts.ts
kirugan: Fixed in f1042630, with your check rather than a Rust-side one.

New `delegated-claim/vaultIdBinding.ts`. Before any wallet prompt, assembly reads the Claim PSBT's input-0 prevout txid, runs it through `derivePeginVaultId(peginTxid, depositorEthAddress)` — the existing dependency-free helper in `clients/eth/pegin-transaction.ts` — and throws `VaultIdBindingError` unless it equals `vault.vaultId`. It runs on the PSBT, not on the finalized Claim, so nothing is signed first.

`DelegatedClaimVaultContext` gains a required `depositorEthAddress`, since the id is `keccak256(peginTxid, depositor)` and the txid alone does not derive it.

I did not push the derivation into `build_watchtower_artifacts`. The depositor's Ethereum address is not in the graph, so Rust would have to be handed it anyway, and the check belongs on the side that knows which vault the user chose.

Tests in `vaultIdBinding.test.ts` cover the byte order — a reversed txid would reject every correct graph — plus the depositor half of the hash, and `assembleWatchtowerArtifacts.test.ts` has a case where a Claim spending another PegIn is refused with no call to `signPsbts`. gbarkhatov's read-side twin is fixed in the same commit.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/readWatchtowerArtifacts.ts
kirugan: Fixed in f1042630, exactly as you suggested.

`summarizeWatchtowerArtifacts` now parses `claim_tx` once and reports `peginTxid` — input 0's prevout in display order — alongside `claimTxid`. `assertArtifactsUsableForVault` takes a required `depositorEthAddress`, and after the string compare against `vault_id` it re-derives the id from that PegIn txid and throws `VaultIdBindingError` on a mismatch. So a file whose graph belongs to vault B but whose `vault_id` says A now fails, whether it came from assembly, a hand edit, or anywhere else.

The check runs before `verifyWatchtowerArtifacts`, so a file for the wrong vault never reaches signature verification.

Shared with the assembly side through `vaultIdBinding.ts`, one derivation for both. The local `normalizeVaultId` moved there rather than being duplicated. The JSDoc that claimed this function "confirms it is the one for this vault" is now true, and `@throws` names the new error. Covered by a test where `vault_id` is right and the graph is not.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/deriveClaimerWotsKeypair.ts
kirugan: Fixed in f1042630 as a runtime check, the way you framed it.

`DeriveClaimerWotsKeypairParams` takes a required `expectedWotsPkHash` — the vault's on-chain `depositorWotsPkHash` — and `deriveClaimerWotsKeypair` asserts `derivation.pk_hash` against it, prefix- and case-insensitively, *before* `validateWotsKeypairAgainstGraph`. The on-chain value is the one anchor the vault provider does not supply, so it runs first and the graph check never gets to pass on its own.

That covers the three drift sources you named: a wrong `htlcVout`, the wrong wallet account, and an expander change after a re-pin. On the last one — the re-pin to vault-wasm main is in 29501910 and the golden vectors came back unchanged, so nothing rotated this time, but the runtime check now catches it if a future bump does.

The error names both hashes and says which of the three it is likely to be, rather than failing later with a graph-shaped message.

Tests: a mismatched on-chain hash rejects; the graph is not consulted when it does; a hash differing only in prefix and case is accepted. Your separate ask, the TS↔WASM equivalence test, is in jeremy's thread on `wotsKeypair.golden.test.ts` and is also done.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/assembleWatchtowerArtifacts.ts
kirugan: You were right, and it was the comments that were lying. Fixed in f1042630.

`assertWalletMatchesDepositor` now does both halves and returns the address it proved: after the `getPublicKeyHex()` compare it takes `getAddress()` and runs `isAddressFromPublicKey(signerAddress, depositorPublicKey, btcNetwork)` — the same helper `PeginManager` uses for the change address — and throws before any prompt if it does not hold. The address can no longer be read separately afterwards, since the function is the only thing that returns it.

This needs a network, which the wallet interface does not carry, so `AssembleWatchtowerArtifactsParams` gains a required `btcNetwork`.

The comment at lines 211-216 and the `signing.ts:49-50` JSDoc are now accurate rather than rewritten — the claim they made is the thing that is implemented. Note `isAddressFromPublicKey` is fail-closed for an x-only key on the P2WPKH branch, which is the behaviour we want here.

Test: a wallet reporting an address from another account is refused with `signPsbts` never called, alongside the existing wrong-public-key case. The fixtures now use the secp256k1 generator point, because the old made-up key had no derivable address.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/readWatchtowerArtifacts.ts
kirugan: Fixed in f1042630, all three.

`requireNumber` is replaced by `requireSafeInteger`, which demands a non-negative safe integer. That covers `-1`, `1.5`, and anything `JSON.parse` has already rounded above 2^53 — and it names the field, instead of `BigInt` throwing a bare RangeError. Applied to both `claimable_event_block_number` and `prover_circuit_version`.

`vault_core_version` is now `unknown` in the raw shape and goes through the same check when present, rather than being copied into a `number` field untyped.

`babe_sessions` goes through `requireRecord`, which rejects a string, an array or null before `Object.keys` sees it. Absent is still `{}`, which is what the CLI does.

Six tests, one per rejection, including `Object.keys("ab")` reporting "0" and "1" as challenger public keys.

packages/babylon-tbv-rust-wasm/src/delegatedClaim.ts
kirugan: Fixed in f1042630 — all five sites, not just the one.

`buildPayoutDepositorPsbt`'s JSDoc no longer points at `extractDepositorPayoutSig` and no longer calls the PSBT conditional. It now says the builder does not read a presigned depositor Payout signature off the graph, so the PSBT always rides in the same batch.

The "claim-time execution is NOT part of this surface" claim is corrected in `delegatedClaim.ts:8-10`, `src/index.ts`, `src/index-node.ts` and `ts-sdk/wasm/index.ts`. Each now names what the surface actually exports at claim time and keeps only the two things that are genuinely outside: the proof comes from the prover service, and nothing here watches the chain for a ChallengeAssert.

Grepped for the phrase afterwards — the only remaining hits are the corrected ones.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/__tests__/deriveClaimerWotsKeypair.test.ts
kirugan: Fixed in f1042630, on both paths you named.

`expandWotsSeed` is now mocked too, so the test can capture the exact buffers: the vault root as the argument `expandWotsSeed` received, the WOTS seed as the argument `wotsKeypairFromSeed` received. Four tests assert every byte is zero afterwards — on success for both buffers, on `expandWotsSeed` rejecting for the root, and on `wotsKeypairFromSeed` rejecting for the seed.

Deleting either `fill(0)` now fails the suite. Checked by removing each one locally before committing.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/__tests__/wotsKeypair.golden.test.ts
kirugan: Fixed in f1042630 — the equivalence test you described, verbatim.

`wotsKeypair.golden.test.ts` now asserts `computeWotsBlockPublicKeysHash(await deriveWotsBlocksFromSeed(seed)) === (await wotsKeypairFromSeed(seed)).pk_hash` for the same 64-byte seed. Both functions zero the seed they are handed, so each gets its own copy.

It runs against the real WASM, not a mock, and it passed unchanged through the re-pin to vault-wasm main in 29501910 — which is the first time it has done the job it exists for.

The two fixed vectors stay. They pin the value; this pins the agreement, which is the thing the on-chain binding actually depends on.

gbarkhatov's related point is handled separately: `deriveClaimerWotsKeypair` now asserts the derived hash against the vault's on-chain `depositorWotsPkHash` at runtime, so the two derivations agreeing on a wrong value is also caught.

packages/babylon-ts-sdk/src/tbv/core/services/delegated-claim/__tests__/assembleWatchtowerArtifacts.test.ts
kirugan: Noted in the file header in f1042630:

> Every wasm call is mocked, so nothing here says anything about the bytes the file ends up carrying. That guarantee lives in btc-vault#2655's `crates/vault/src/delegated_claim/artifacts.rs` round-trip test, which is where an `artifacts.json` format change has to be caught.

That PR has since merged, at `ac4954e7`, and the re-pin in 29501910 moves `btc_vault_v3` onto it, so the reference now points at something on main.

The suite did grow two cases that are not orchestration — a wallet address from another account, and a Claim spending another vault's PegIn — but both are refusals before signing, so the header's point about the bytes still stands.

gbarkhatov (approved):

Issue #2386, opened 2026-09-03
[Security][Critical] Reject extra outputs in WASM Pre-PegIn transactions


jonybur:
## Outcome

Reject every WASM Pre-PegIn transaction that does not have the exact expected output set. No untrusted output reaches UTXO selection, wallet signing, or broadcast.

Status: #2398 delivered the implementation and all five behavior checks below. This issue remains open for the required second qualified review record. No unfinished output-layout implementation was found in the closeout audit.

## Problem

Severity: Critical, I4 x L2 with silent fund loss.

Audit commit: `4f2681948d22257341897e58c0c79bcd5b66371b`.

The trusted check accepts more outputs than the HTLC count. It checks only the first `N` outputs ([assertWasmPeginSizing.ts:241-281](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts#L241-L281)). The auth check validates output `N`, but it does not reject later outputs ([assertAuthAnchorOpReturn.ts:36-80](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/packages/babylon-ts-sdk/src/tbv/core/managers/pegin/assertAuthAnchorOpReturn.ts#L36-L80)).

The builder sums all WASM outputs. UTXO selection funds that sum ([pegin.ts:210-228](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts#L210-L228), [PeginManager.ts:860-901](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts#L860-L901)). The funder then copies every output ([fundPeginTransaction.ts:151-154](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/packages/babylon-ts-sdk/src/tbv/core/utils/transaction/fundPeginTransaction.ts#L151-L154)).

Reproduce:

1. Use a 1,000,000-sat two-vault deposit.
[...]

PR #2398, opened 2026-09-03
fix(ts-sdk): validate WASM Pre-PegIn output layout

jonybur:
## Outcome

Rejects invalid WASM transaction data before funding or signing a deposit or refund.

The implementation satisfies all five behavior conditions in #2386. The issue stays open for the required second qualified review record.

## Change

- Validate Pre-PegIn output count, values, order, and independently derived HTLC scripts.
- Check transaction headers, signing data, auth outputs, and depositor anchor outputs.
- Check PegIn and refund data before it reaches signing.
- Register the transaction parser guard in every critical-path inventory.
- Cite the Rust source for refund constants and keep them beside the other transaction constants.
- Cover real-engine construction with unsorted keepers and multiple unsorted universal challengers.
- Include the shared build repair from #2408.

## Safety

The layout check runs before output summation and UTXO selection. Manager tests prove failures stop signing in the sizing and commit passes. Critical changes require two qualified code-owner approvals on the final changes.

## Verification

Local checks passed with Node 24:

- `pnpm nx run @babylonlabs-io/ts-sdk:test --skipNxCache`: build, 1,966 SDK tests, and 1 built facade test.
- `pnpm nx run @babylonlabs-io/ts-sdk:test-browser --skipNxCache`: 34 Chromium tests.
- `pnpm nx run @babylonlabs-io/babylon-tbv-rust-wasm:test --skipNxCache`: 7 engine tests.
- `pnpm nx run @babylonlabs-io/ts-sdk:lint --skipNxCache`: passed.
- Critical-path inventory checks, workflow YAML parsing, and `git diff --check`: passed.

The moved constants are internal. They do not add public API documentation. [Final remote verification passed](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34167533604). The records show one final approval. The required second qualified review record remains open in #2386.

Completion target: reject every invalid Pre-PegIn output layout before funding or signing. All five behavior checks in #2386 are verified. No further output-layout implementation was identified. Remaining: link or obtain the required second qualified final review.

## Links

Tracks #2386. #2386 stays open.

Tracks #2228. #2228 stays open.


packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
```diff
@@ -280,3 +291,80 @@ export function assertEncodedHtlcOutputsMatch(
...
 }
+
+/**
+ * Validate the unfunded Pre-PegIn protocol output layout.
+ *
+ * This check rejects a wrong output count, a wrong auth or CPFP output, and an
+ * HTLC output that does not match the WASM metadata. Run it before value
+ * summing, UTXO selection, or signing.
+ */
+export function assertUnfundedPrePeginOutputLayout(
```
gbarkhatov, line 488: **`nVersion` and `nLockTime` still cross the boundary unasserted and reach the signed transaction.**

This guard validates count, HTLC values/scripts, the auth output and the CPFP output — but those are four of the six things `parseUnfundedWasmTransaction` returns. `fundPeginTransaction.ts:141-142` copies the other two verbatim (`tx.version = version; tx.locktime = locktime;`) into the transaction the wallet signs.

The Rust reference pins both — `crates/vault/src/transactions/prepegin.rs:139-141` builds `version: Version::TWO, lock_time: LockTime::ZERO` — and `from_funded_transaction` rejects anything lower on the way back in (`:159-163`). A binary returning nVersion 3 makes the Pre-PegIn a TRUC parent, so the non-v3 child PegIn becomes non-relayable and the deposit strands *after* the wallet has already signed.

The precedent and the naming already exist in the `constants.ts` you edited here: `PAYOUT_TX_VERSION = 2` / `PAYOUT_TX_LOCKTIME = 0`, with a comment noting the depositor's signature commits to both. Two literal comparisons close it, and this PR registering these files under CLAUDE.md §1 is what raises the bar.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
```diff
@@ -280,3 +291,80 @@ export function assertEncodedHtlcOutputsMatch(
...
+    throw new Error(
+      `WASM Pre-PegIn output layout has ${htlcValues.length} HTLC output(s); ` +
+        `expected at least ${MIN_PRE_PEGIN_HTLC_OUTPUT_COUNT}.`,
+    );
+  }
+
+  const authOutputIndex = htlcValues.length;
+  const cpfpOutputIndex =
+    authOutputIndex + Number(authAnchorHash !== undefined);
+  const expectedCount = cpfpOutputIndex + PRE_PEGIN_CPFP_OUTPUT_COUNT;
```
gbarkhatov: **The layout formula is re-implemented rather than reused, putting the same expression in critical paths §1 and §2.**

`peginOutputCount(vaultCount, hasAuthAnchor)` at `utils/fee/constants.ts:77-91` computes `vaultCount + PEGIN_FIXED_OUTPUTS + (hasAuthAnchor ? PEGIN_AUTH_ANCHOR_OUTPUTS : 0)` — the same expression — and `prepareSizing` (`PeginManager.ts:866`) already calls it on the same numbers to size the fee. Its JSDoc says it exists "so the fee budget stays in lockstep with the output set".

So the fee model and the enforced layout now derive the same count two different ways, in two different critical-path sections. A future layout change (a second anchor, a commission output) touched in one and not the other gives a fee estimate that disagrees with what the guard accepts — the underfunding failure §2 exists to prevent.

Suggest `const expectedCount = peginOutputCount(htlcValues.length, authAnchorHash !== undefined);` and dropping `PRE_PEGIN_CPFP_OUTPUT_COUNT`.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/constants.ts
```diff
@@ -37,6 +37,15 @@ export const ASSERT_PAYOUT_OUTPUT_INDEX = 0;
  */
 export const PAYOUT_ANCHOR_DUST_SATS = 546;
 
+/**
+ * Value in sats for the Pre-PegIn CPFP anchor output. Matches `DUST_AMOUNT`
+ * in `btc-vault crates/vault/src/lib.rs`.
+ */
+export const PRE_PEGIN_CPFP_VALUE_SATS = 546;
```
gbarkhatov, line 57: **These are the guard's acceptance bounds, and this file is in none of the six critical-path inventories.**

`PRE_PEGIN_CPFP_VALUE_SATS` and `PRE_PEGIN_CPFP_OUTPUT_COUNT` decide what `assertUnfundedPrePeginOutputLayout` accepts. But `.../primitives/psbt/constants.ts` appears in neither workflow inventory, nor `.github/CODEOWNERS` (which lists `assertWasmPeginSizing.ts`, `pegin.ts` and `payout.ts` from this same directory), nor the ESLint strict-typing list.

The workflow matcher is a prefix test (`critical-path-check.yml:65-67`), so registering the two sibling files does not reach this one. Editing 546 therefore changes validation behaviour with no critical-path label and no code-owner requirement — against this PR's own SECURITY.md hunk: "**Change them together.** They have drifted before."

Adding the one path to the inventories is a small change and closes it.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
```diff
@@ -207,19 +207,20 @@ export async function buildPrePeginPsbt(
   // PegIn registration. Both the sizing and commit passes route through here.
   const minPeginFee = await assertWasmPeginSizing(result, params);
 
-  // Parse the unfunded tx to sum all output values
+  // Parse the unfunded tx before summing all output values
   // (HTLCs + optional OP_RETURN + CPFP anchor). This is the amount
   // UTXOs must cover before adding network fees.
   const parsed = parseUnfundedWasmTransaction(result.txHex);
```
gbarkhatov, line 219: **The count assertion binds the declared output varint, not the transaction bytes.**

`parseUnfundedWasmTransaction` reads the self-declared output count and parses exactly that many outputs, never checking it consumed the buffer up to the 4-byte locktime (`utils/transaction/fundPeginTransaction.ts:97-116`). A binary that declares 3 outputs but serializes 4 passes "expected exactly 3", and `buildPrePeginPsbt` returns `psbtHex: result.txHex` at `:237` — the original bytes, not a re-serialization.

To be clear about severity: **this does not reach the wallet on the in-repo path.** `fundPeginTransaction` re-parses with the same parser and rebuilds from the same 3 outputs, so the 4th is dropped. I checked this specifically because it was the one point the three reviews disagreed on.

It is still worth closing, because `psbtHex`'s own JSDoc at `:98-107` makes the caller responsible for "Funding the transaction (add inputs and change output)", and `buildPrePeginPsbt` is a public export (`primitives/index.ts:111`) described as the lower-level primitive for custom implementations. Have the parser return its consumed offset and assert it lands at `hex.length - 8`.

.github/workflows/service-release-simple-staking.yml
```diff
@@ -32,6 +32,12 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Install Chromium for browser tests
+        run: pnpm --filter @babylonlabs-io/ts-sdk exec playwright install --with-deps chromium
+
+      - name: Test SDK in Chromium
```
gbarkhatov: **A simple-staking production release is now gated on a ts-sdk Chromium suite it has no dependency on.**

`services/simple-staking/package.json` has no `@babylonlabs-io/ts-sdk` dependency, so this release now blocks on a Playwright browser download plus a Pre-PegIn test that cannot affect it. Two new failure modes (browser download flake, an unrelated SDK browser test) on a pipeline that previously ran no ts-sdk-specific target.

The same two steps are pasted unconditionally into four workflows with no composite action and no `if:` gating. `verify.yml` and `package-release.yml` are the right homes for them; suggest dropping the pair here.

.github/workflows/service-release-vault.yml
```diff
@@ -48,6 +48,12 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Install Chromium for browser tests
+        run: pnpm --filter @babylonlabs-io/ts-sdk exec playwright install --with-deps chromium
+
+      - name: Test SDK in Chromium
```
gbarkhatov, line 56: Same as the note on `service-release-simple-staking.yml`: this is the fourth copy of the two steps, unconditional and uncached, on a release pipeline that already gates on `pnpm run test`. Worth either a composite action or dropping it from the service releases and keeping it in `verify.yml` / `package-release.yml`.

SECURITY.md
```diff
@@ -765,7 +766,7 @@ only repository-local safeguards.
 | ------------------- | --------- | --------------------------------------------------------------------------------- | ----------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------- |
 | WASM boundary       | F/—       | A WASM getter returns `0n` or a wrong value and reaches a signed tx               | **User fund loss**                                                      | `assertWasmBigint` / `assertPositiveBigintArray` at the facade; call-site cross-checks on `/raw`                | `babylon-tbv-rust-wasm` value-guard tests                         |
```
gbarkhatov: **This row was not updated, and the row below it overstates what is achievable.**

The WASM-boundary row still lists mitigation as "`assertWasmBigint` / `assertPositiveBigintArray` at the facade; call-site cross-checks on `/raw`" and tests as "`babylon-tbv-rust-wasm` value-guard tests" — no mention of the Pre-PegIn layout guard this PR adds. The row immediately below it *was* edited in this diff, which makes the omission look like a miss rather than a decision. SECURITY.md's own closing rule says security-relevant PRs should reference the relevant row, and "if a row is missing, add it" — the extra-output threat has no row at all.

Separately, on line 769: "All six align" cannot hold. ESLint `files` globs resolve relative to the config file, so an override rooted at `packages/babylon-ts-sdk/` can never match the eleven non-SDK paths in the workflow inventory (`packages/babylon-tbv-rust-wasm/{src,scripts}/*`, the four `services/vault/src/...` files, `packages/babylon-ledger-vault-signer/{src,scripts}/`). The failure that follows is the next maintainer adding `services/vault/src/services/vault/vaultActivationService.ts` to that array to restore "alignment": it matches nothing, lints nothing, and reads as coverage. Suggest describing it as a package-scoped strict-typing override for the ts-sdk subset rather than a sixth parallel inventory.

CLAUDE.md
```diff
@@ -48,7 +48,10 @@ These paths handle irreversible value movement. An AI-generated mistake here is
 
 ### 1. WASM boundary (value computation)
 
-- File: `packages/babylon-tbv-rust-wasm/src/index.ts`
+- Files:
+  - `packages/babylon-tbv-rust-wasm/src/index.ts`
+  - `packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts`
+  - `packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts`
 - The Rust/WASM layer computes `htlcValue = peginAmount + depositorClaimValue + p2aAnchorValue + minPeginFee` internally (the anchor term is 0 for tx-graph v1, 240 sats for v2 and v3). JS receives outputs with no runtime validation.
```
gbarkhatov: **This line now contradicts the list directly above it.**

"JS receives outputs with no runtime validation" sits under a list whose two new entries *are* that runtime validation — per CLAUDE.md's own "After changes: check for comments/docs that reference old behavior and update them."

Worth using the rewrite to name what is still unchecked, since that is what makes this a partial fix: the HTLC `scriptPubKey`s are compared WASM-against-WASM only, so a doctored binary can still point an HTLC at an attacker output key. That is #2386's "verify every HTLC value and script" bullet. It is stated honestly in the code (`assertWasmPeginSizing.ts:249-250` — "It does not derive the HTLC scripts"), but a reader of the PR body or SECURITY.md would not find it, and "partial fix" alone does not say which bullet is unmet.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
```diff
@@ -24,9 +24,18 @@ import {
...
 
+import {
+  PRE_PEGIN_CPFP_OUTPUT_COUNT,
+  PRE_PEGIN_CPFP_VALUE_SATS,
+} from "./constants";
 import type { PrePeginParams } from "./pegin";
 
+const MIN_PRE_PEGIN_HTLC_OUTPUT_COUNT = 1;
+const PRE_PEGIN_AUTH_OUTPUT_VALUE_SATS = 0;
+const PRE_PEGIN_AUTH_SCRIPT_PREFIX = "6a20";
```
gbarkhatov: `"6a20"` is the one new constant asserting a wire format, and it is the only one with nothing pointing at its source. It encodes `OP_RETURN PUSH32`, matching `ScriptBuf::new_op_return(hash)` in `prepegin.rs:132` and `parse_auth_anchor_hash_from_script`.

`PRE_PEGIN_CPFP_VALUE_SATS` got a `btc-vault crates/vault/src/lib.rs` citation over in `constants.ts`; this one and the two above it stayed module-local and bare. Worth the same citation, and worth co-locating all five so the layout constants are in one place.

packages/babylon-ts-sdk/vitest.browser.config.ts
```diff
@@ -0,0 +1,20 @@
+import { playwright } from "@vitest/browser-playwright";
+import { defineConfig } from "vitest/config";
+
+import { TEST_TIMEOUT_MS } from "../../vitest.shared";
+
+export default defineConfig({
+  test: {
+    include: [
```
gbarkhatov, line 14: **A hardcoded one-file include, and the file it names mocks WASM wholesale.**

`assertWasmPeginSizing.test.ts` mocks `../../../wasm` entirely, so the Chromium run exercises no real WASM — the only genuinely browser-specific coverage is bitcoinjs `payments.p2tr` under the `Buffer` polyfill. Four workflows now pay an apt-installing Chromium install for that, and any future browser-sensitive test silently does not run until someone remembers to edit this list.

Also worth mirroring `globals: true` from `vitest.config.ts:6`. Harmless today because this file imports every helper explicitly, but the first added file that relies on globals fails with a confusing `describe is not defined`.

A glob (`**/*.browser.test.ts`) would fix both; failing that, a comment stating the inclusion criterion keeps the list from rotting.

packages/babylon-ts-sdk/project.json
```diff
@@ -9,6 +9,9 @@
       "outputs": ["{projectRoot}/dist"],
       "cache": true
     },
+    "test-browser": {
```
gbarkhatov, line 12: `test-browser` declares only `dependsOn` and resolves through inferred `nx:run-script`, where its sibling `test` two lines below spells out `executor`, `options.command`, `cwd` and `cache`. Two targets in one file now depend on two different resolution mechanisms — and disabling script inference or setting `includedScripts` on this project would make the CI step vanish rather than fail.

It also has no `cache`, so it re-runs on every invocation across all four workflows that call it. Suggest declaring it the way `test` is declared.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/__tests__/assertWasmPeginSizing.test.ts
```diff
@@ -272,63 +290,195 @@ describe("assertWasmPeginSizing", () => {
...
+    ).resolves.toMatchObject({ authAnchorVout: 2 });
+  });
+
+  it("rejects a layout with no HTLC outputs", async () => {
+    await expect(
+      buildPrePeginPsbt(makeParams({ hashlocks: [], pegInAmounts: [] })),
+    ).rejects.toThrow(/expected at least 1/);
+  });
+
+  it.each([1, 2, 3])(
```
gbarkhatov: **This runs the same assertion three times and reads as cross-version coverage it does not provide.**

`makeWasmResult` varies only the HTLC *value* by version (anchor 0 vs 240); the output layout it builds is byte-identical for 1, 2 and 3, the count check never reads the value, and all three cases expect the identical error string. The Pre-PegIn layout has no version dependence in the Rust reference either — the P2A anchor is a PegIn output, not a Pre-PegIn one. Suggest collapsing to one case.

Separately, there is no negative case for the layout *without* an auth anchor: every negative test routes through `makeTwoVaultBuildParams()`, which always sets `authAnchorHash` at `:365`. So the `expectedCount = N + 1` arm is only ever exercised by the happy path, and a mistake in it — treating an empty-string hash as present, say — would ship green. One extra-output case with `authAnchorHash` omitted, asserting `has 3 output(s); expected exactly 2`, covers it.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
```diff
@@ -9,7 +9,7 @@
  * P2A anchor, 240 sats for v2/v3) and JS receives the outputs with no runtime
  * validation. A doctored or buggy binary that returns a different
  * `peginAmount`, an out-of-formula `htlcValue`, or a wrong
- * `depositorClaimValue` would otherwise be committed verbatim — taxing the
+ * `depositorClaimValue` would otherwise be committed verbatim - taxing the
```
gbarkhatov, line 12: **Punctuation-only churn in a file that requires a two-reviewer line-walk.**

Roughly ten comment lines here change only `—`→`-` and `…`→`...` with no semantic content, and the sweep is inconsistent — line 198 still carries `×` inside a runtime template literal, and `pegin.ts` keeps its em-dashes. I checked `eslint.config.mjs`; there is no rule requiring ASCII.

The same pass did cost something in the test file: the module docstring ("Tests for assertWasmPeginSizing — the cross-check that guards every value-bearing field…") became "Tests for WASM Pre-PegIn result validation.", and the two comments deriving `ANCHOR_VALUE = 240n` and `RESERVE_PLAUSIBILITY_CAP = 1_000_000n` (`10 × MAX_REASONABLE_PEGIN_VBYTES`) were deleted while both constants remain — so they are now unexplained numbers.

CLAUDE.md requires the author to line-walk every changed line in this file with two reviewers. Reverting the punctuation-only hunks and restoring those two derivations would keep that review focused on the guard.

packages/babylon-ts-sdk/src/tbv/core/managers/__tests__/PeginManager.test.ts
```diff
@@ -263,6 +279,23 @@ const BASE_PREPARE_PEGIN_PARAMS = {
...
+  result: PrePeginResult,
+): PrePeginResult {
+  const parsed = parseUnfundedWasmTransaction(result.txHex);
+  const tx = new bitcoin.Transaction();
+  tx.version = parsed.version;
+  tx.locktime = parsed.locktime;
+  for (const output of parsed.outputs) {
+    tx.addOutput(output.script, output.value);
+  }
+  tx.addOutput(bitcoin.script.compile([bitcoin.opcodes.OP_RETURN]), 0);
```
gbarkhatov: **This does not reproduce the scenario the issue is closed against.**

#2386's repro is a 1,000,000-sat two-vault deposit with an appended **100,000-sat P2TR** output; this appends a bare `OP_RETURN` at **value 0** on a single-vault deposit (`amounts: [TEST_AMOUNTS.PEGIN]` at `:391`).

The count check is value-blind, so it genuinely catches both — the test is not wrong. But nothing here demonstrates the actual harm: that a value-bearing extra output would otherwise inflate `totalOutputValue` and the UTXO-selection target, and get funded and signed. Giving this output a value and using two vaults is close to a one-line change and makes the test match the criterion the issue closes on.

packages/babylon-ts-sdk/src/tbv/core/managers/__tests__/PeginManager.test.ts
```diff
@@ -317,6 +350,64 @@ describe("PeginManager", () => {
...
+            : result;
+        };
+
+        try {
+          await expect(
+            manager.preparePegin({
+              amounts: [TEST_AMOUNTS.PEGIN],
+              ...BASE_PREPARE_PEGIN_PARAMS,
+              availableUTXOs: {
+                [Symbol.iterator]: utxoIterator,
```
gbarkhatov: `{ [Symbol.iterator]: utxoIterator } as unknown as UTXO[]` plus `expect(utxoIterator).toHaveBeenCalledTimes(...)` at `:405` asserts a mock internal rather than a behaviour — it holds only while `selectUtxos` iterates the array exactly once per pass, so an unrelated refactor of the selector breaks it.

The behaviour actually worth pinning is already pinned elsewhere in this test (that `signPsbt` / `signPsbts` are never reached after a layout failure). The double cast also runs against the Type Safety rule. Minor, but the assertion could go without losing coverage.

packages/babylon-ts-sdk/src/test/setup.browser.ts
```diff
@@ -0,0 +1,14 @@
+import { Buffer } from "buffer";
```
gbarkhatov, line 1: This restates `setup.ts:1-16` — same ECC library, same `beforeEach(vi.clearAllMocks)` — with `globalThis.Buffer = Buffer` before the imports as the real difference.

Worth a comment on that dynamic `await Promise.all([import(...)])`, because it is load-bearing rather than stylistic: static imports hoist above the `globalThis.Buffer` assignment and would break bitcoinjs at module-evaluation time. As written, the next tidy-up that converts it to static imports breaks the browser run for a reason that is not obvious from the file. Importing the shared setup instead of restating it would also keep the two from drifting.

gbarkhatov (dismissed):
Reviewed this three times independently, then deduped and verified everything below against the code at `329cd19c2` and against the Rust reference in `btc-vault crates/vault/src/transactions/prepegin.rs`.

**The fix is real and correctly placed.** `outputs.length !== expectedCount` is exact, it runs from `pegin.ts:218` *before* the `totalOutputValue` sum at `:226` and before UTXO selection, and both the sizing and commit passes route through `buildPrePeginPsbt` (`PeginManager.ts:843`, `:984`). There is no N-vs-N+1 ambiguity: `authOutputIndex` and `cpfpOutputIndex` derive from the *caller's* `authAnchorHash`, never from a WASM value, so an unrequested OP_RETURN yields N+2 and fails the count. The CPFP check is genuinely independent — `deriveBip86ScriptPubKeyHex` is not mocked. **No merge-blockers.** Comments are suggestions; your call on what lands here versus a follow-up.

The strongest one is inline on `assertWasmPeginSizing.ts`: `nVersion` and `nLockTime` are the two fields `parseUnfundedWasmTransaction` returns that the new guard does *not* check, and `fundPeginTransaction.ts:141-142` copies both verbatim into the transaction the wallet signs. The Rust reference pins them and the payout path in the same `constants.ts` you edited already has the precedent.

**Refuted during verification** — recording these so they don't get re-raised:

- *"Undeclared trailing outputs ride out on `psbtHex` and reach the signed transaction."* They do not. `fundPeginTransaction` re-parses `unfundedTxHex` with the same `parseUnfundedWasmTransaction` and rebuilds from that result (`:135-154`), so an output the guard did not inspect cannot reach the wallet. A narrower residual survives and is inline on `pegin.ts:213`.
- *"SECURITY.md's 'the guard is not defence in depth — it is the only check' is now false."* That sentence scopes to the facade crossing and its value guards. `assertWasmPeginSizing` predates this PR, so this change does not falsify it.
- *"`PRE_PEGIN_CPFP_VALUE_SATS` duplicating `PAYOUT_ANCHOR_DUST_SATS` at 546 is a magic-number violation."* `constants.ts` documents that convention for numerically-equal but distinct concepts (the `BPS_DENOMINATOR` note at `:82-88`), and both carry the same citation. Correct as written.

**Verified clean:** index arithmetic covers every parsed output with no gap; keeping `assertEncodedHtlcOutputsMatch` loose was right, since its other consumer inspects an already-broadcast *funded* tx where a trailing change output is legitimate, and `:470-483` pins that deliberately; the new deps are pinned exact and match the lockfile; all actions pinned by SHA, no `pull_request_target`, no widened permissions.

One caveat on the constants: 546 against `DUST_AMOUNT` (`lib.rs:158`), the BIP-86 keypath anchor (`connectors/mod.rs:268-280`), the `6a20` prefix and the HTLCs → OP_RETURN → CPFP ordering (`prepegin.rs:66-68, 110-137`) all match — but confirmed against btc-vault `main`, **not** against `VAULT_WASM_COMMIT = 3accd8f6`, which is not checked out locally. Byte-level agreement at the pin is unverified here.

Two adjacent notes, not this PR: `services/vault/src/utils/wasm.ts:7` still names `assertEncodedHtlcOutputsMatch` as the signing-path verifier when the primary guard is now `assertUnfundedPrePeginOutputLayout`; and `pnpm-lock.yaml` now carries a second `pixelmatch` (7.1.0 alongside 7.2.0).

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
jonybur: Fixed in 34afe51d. The guard now requires Pre-PegIn version 2 and locktime 0 before funding. Tests cover versions 1 and 3 plus non-zero locktime.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
jonybur: Fixed in 34afe51d. The layout guard now uses peginOutputCount, and the duplicate CPFP output-count constant is removed.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/constants.ts
jonybur: Fixed in 34afe51d. Both layout constants files are now in each applicable critical-path inventory and the SDK strict ESLint subset.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
jonybur: Fixed in 34afe51d. The parser now requires the declared outputs to end exactly at the locktime boundary. A negative test covers undeclared output bytes.

.github/workflows/service-release-simple-staking.yml
jonybur: Fixed in 34afe51d. The unrelated SDK Chromium steps are removed from the simple-staking release workflow.

.github/workflows/service-release-vault.yml
jonybur: Kept by design in 34afe51d. Vault bundles the SDK, and this release workflow runs independently. The workflow now records why its SDK browser gate must remain.

SECURITY.md
jonybur: Fixed in 34afe51d. SECURITY.md now covers the Pre-PegIn guard and states that ESLint mirrors only the SDK subset of the five full inventories.

CLAUDE.md
jonybur: Fixed in 34afe51d. CLAUDE.md now describes the independent TypeScript output and signing-field checks. The current implementation also derives each HTLC script independently.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
jonybur: Fixed in 34afe51d. The layout constants are co-located in psbt/constants.ts, and the auth prefix cites the pinned Rust construction.

packages/babylon-ts-sdk/vitest.browser.config.ts
jonybur: Fixed in 34afe51d. Browser globals are enabled, and the explicit include now documents that this suite protects the Vault SDK boundary in both Node and Chromium.

packages/babylon-ts-sdk/project.json
jonybur: Fixed in 34afe51d. test-browser is now an explicit cached nx:run-commands target with its command and project working directory.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/__tests__/assertWasmPeginSizing.test.ts
jonybur: Fixed in 34afe51d. The redundant version loop is collapsed, and negative tests now cover extra outputs with and without an auth output.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
jonybur: Addressed in 34afe51d. I restored the removed test explanations. I kept regular hyphens because the repository style requires them.

packages/babylon-ts-sdk/src/tbv/core/managers/__tests__/PeginManager.test.ts
jonybur: Fixed in 34afe51d. The regression now uses two 500,000-sat vaults and appends a 100,000-sat foreign P2TR output.

packages/babylon-ts-sdk/src/tbv/core/managers/__tests__/PeginManager.test.ts
jonybur: Fixed in 34afe51d. The iterator mock, exact call-count assertion, and double cast are removed. The test still proves that signing does not run.

packages/babylon-ts-sdk/src/test/setup.browser.ts
jonybur: Fixed in 34afe51d. Browser setup now sets Buffer first and then dynamically imports the shared setup. The comment records why import order is required.

jonybur:

packages/babylon-ts-sdk/src/tbv/core/utils/transaction/fundPeginTransaction.ts
```diff
@@ -113,6 +113,13 @@ export function parseUnfundedWasmTransaction(
     outputs.push({ value, script });
   }
 
+  const locktimeOffset = unfundedTxHex.length - 8;
+  if (pos !== locktimeOffset) {
+    throw new Error(
```
gbarkhatov, line 118: **This file now enforces a critical-path invariant but is in none of the six inventories.**

The delta turned `parseUnfundedWasmTransaction` from a pure parser into a guard: it rejects a WASM transaction whose declared outputs do not end at the locktime boundary, and two suites assert this exact message. The same function still produces the `version` and `locktime` that `assertUnfundedPrePeginOutputLayout` pins and that `fundPeginTransaction:148-149` copies into the transaction the wallet signs.

I grepped all six inventories for both `fundPeginTransaction` and `utils/transaction` and got zero hits — no `CODEOWNERS` entry, neither workflow inventory, neither markdown inventory, nor the SDK strict-typing ESLint list. So this guard can be weakened or deleted with no code-owner review and no critical-path label, which is the precise situation the registration in finding 3 was meant to prevent.

That is the same argument you accepted for `psbt/constants.ts`, and the fix is the same shape: one line alongside each entry already added at `CODEOWNERS:21`, `critical-path-check.yml:32`, `claude-md-drift.yml:34`, `eslint.config.mjs:28` and `CLAUDE.md:54`.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/refund.ts
```diff
@@ -11,24 +11,23 @@
...
   TAPSCRIPT_LEAF_VERSION,
   deriveBip86ScriptPubKeyHex,
-  hexToUint8Array,
   stripHexPrefix,
   uint8ArrayToHex,
 } from "../utils/bitcoin";
+import { deriveExpectedPrePeginHtlc } from "./assertWasmPeginSizing";
 import { normalizeAuthAnchorHash, type PrePeginParams } from "./pegin";
 
+const REFUND_TX_VERSION = 2;
```
gbarkhatov: **These reintroduce module-local, uncited transaction literals that this same PR just finished co-locating.**

Every sibling literal the PR added — `PRE_PEGIN_TX_VERSION`, `PRE_PEGIN_TX_LOCKTIME`, `PEGIN_TX_LOCKTIME`, `PAYOUT_TX_VERSION`, `PAYOUT_TX_LOCKTIME` — now lives in `psbt/constants.ts:59-91` with a `btc-vault crates/vault/src/transactions/...` citation naming the Rust construction that fixes it. These two sit here with neither.

Two consequences. A Rust-side change to the refund transaction's version or locktime will be found in five places and missed in this one. And a reader cannot tell from this file whether the `2` is the same protocol constant as `PRE_PEGIN_TX_VERSION` or a coincidence that happens to match today.

`psbt/constants.ts` is the file this PR just registered as the single home for these values, and it is now a registered critical path with code owners — which is exactly the review surface a transaction-version constant should sit behind. Moving both there with the refund Rust citation would give them the treatment `PRE_PEGIN_AUTH_SCRIPT_PREFIX` already got.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/__tests__/assertWasmPeginSizing.test.ts
```diff
@@ -272,63 +316,246 @@ describe("assertWasmPeginSizing", () => {
...
+  it("accepts exact layouts with and without an auth output", async () => {
+    await expect(buildPrePeginPsbt(makeParams())).resolves.toMatchObject({
+      authAnchorVout: null,
+    });
+    await expect(
+      buildPrePeginPsbt(makeTwoVaultBuildParams()),
+    ).resolves.toMatchObject({ authAnchorVout: 2 });
+  });
+
+  it.each([1, 2, 3])(
```
gbarkhatov: **This `it.each([1, 2, 3])` runs one version-independent assertion three times.**

The case sets `htlcScriptOverride = SCRIPT_A` so the metadata and the encoded bytes agree, then expects the independent derivation to catch the forgery. But `deriveExpectedPrePeginHtlc` takes no version argument — its `PrePeginHtlcParams` is a `Pick` that excludes `vaultCoreVersion` (`assertWasmPeginSizing.ts:56-64`) — and the script comparison at `:533-544` never reads one either.

So all three cases execute an identical code path and expect an identical error regex. It reads as cross-version coverage of the forgery check while supplying none, which is the same objection that got the count loop collapsed one comment earlier in this file.

Collapsing this to a single case would say what is actually being tested. If cross-version coverage of the forgery path is wanted, it needs a parameter that reaches the derivation.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
```diff
@@ -9,24 +9,215 @@
...
+
+  return key;
+}
+
+function normalizeKeyGroup(values: readonly string[], label: string): string[] {
+  if (values.length === 0) {
+    throw new Error(`${label} must not be empty.`);
+  }
+
+  const keys = values
```
gbarkhatov, line 100: **Coverage gap, not a divergence — the new TS derivation is differentially checked against real WASM for only one universal challenger.**

`deriveExpectedPrePeginHtlc` is new in this delta and is now the sole source of the `tapLeafScript`, control block and `tapMerkleRoot` the depositor signs (`peginInput.ts:168-176`, `refund.ts:248-256`); WASM's own connector scripts are discarded. `normalizeKeyGroup` sorts and dedups, and `nOfNChunks` emits the `OP_CHECKSIGADD` chain.

Every real-WASM test uses one keeper and one universal challenger — where the sort is a no-op and no `OP_CHECKSIGADD` is emitted. The one multi-key real-WASM exercise (`PeginManager.test.ts:640-668`) covers the *keeper* group, which is what makes me confident the ordering is right in general; the universal-challenger group has no equivalent. `assertWasmPeginSizing.test.ts` cannot close this on its own, because `makeWasmResult` builds its "WASM" scriptPubKeys by calling `deriveExpectedPrePeginHtlc` itself — self-referential by construction.

One real-WASM case with two or more universal challengers supplied unsorted would close it. Marking this **UNVERIFIED against btc-vault** — the Rust n-of-n key ordering is not checked out locally, so this is a gap in the differential, not a claimed divergence. Worth noting a real divergence here would fail loudly at build time rather than silently mis-sign, which is why this is not a blocker.

gbarkhatov (dismissed):
Follow-up pass over my earlier review (`5112287788`, at `329cd19c`), not a second opinion on the same code. Every one of the 28 files changed in the delta, so I read the current code for all sixteen verdicts rather than inferring any from the replies.

**14 fixed, 1 partially fixed, 1 declined. No merge-blockers in the delta.**

|---|---|---|---|
| 1 | `nVersion` / `nLockTime` cross the boundary unasserted | **fixed** | `assertWasmPeginSizing.ts:497-508` |
| 2 | Layout formula re-implemented instead of reusing `peginOutputCount` | **fixed** | `assertWasmPeginSizing.ts:520-523` |
| 3 | `psbt/constants.ts` in none of the six inventories | **fixed** | `CODEOWNERS:21`, `critical-path-check.yml:32`, `claude-md-drift.yml:34`, `eslint.config.mjs:28`, `CLAUDE.md:54`, `SECURITY.md:215` |
| 4 | Count binds the declared varint, not the transaction bytes | **fixed** | `fundPeginTransaction.ts:116-121` |
| 5 | simple-staking release gated on an SDK Chromium suite | **fixed** | steps removed; file is not among the PR's 28 |
| 6 | Fourth copy of the steps on the vault release pipeline | **declined by the author** | `service-release-vault.yml:51-52` |
| 7 | Stale WASM row; "All six align" cannot hold | **fixed** | `SECURITY.md:781`, `:783` |
| 8 | "no runtime validation" contradicts the list above it | **fixed** | `CLAUDE.md:57` |
| 9 | `"6a20"` uncited; layout constants scattered | **fixed** | `psbt/constants.ts:46-51` |
| 10 | Hardcoded one-file include; `globals` not mirrored | **fixed**, per the stated fallback | `vitest.browser.config.ts:11-13` |
| 11 | `test-browser` resolves through an inferred script, uncached | **fixed** | `project.json:12-20` |
| 12 | Redundant version loop; no negative case without an auth anchor | **fixed**, both halves | `assertWasmPeginSizing.test.ts:458-464` |
| 13 | Punctuation churn; deleted constant derivations | **partially fixed** — derivations restored, punctuation declined | `assertWasmPeginSizing.test.ts:1-5, 48-53` |
| 14 | Repro does not match the issue's scenario | **fixed** | `PeginManager.test.ts:362-408` |
| 15 | Iterator mock, call-count assertion, double cast | **fixed** | `PeginManager.test.ts:405-407` |
| 16 | `setup.browser.ts` restates `setup.ts` | **fixed** | `setup.browser.ts:1-6` |

The existing threads stand as written; nothing there needs restating. Four new points from the delta are inline below.

On **13**, since its substance moved: the deleted derivations are genuinely back — the module docstring at `assertWasmPeginSizing.test.ts:1-5`, `ANCHOR_VALUE` at `:48-49`, `RESERVE_PLAUSIBILITY_CAP` at `:50-53` — so only the punctuation half remains, and that was declined as repository style. Recording for completeness that the inconsistency the original comment described does survive: `assertWasmPeginSizing.ts:387` still emits `×` inside a runtime template literal while `psbt/constants.ts` keeps em-dashes. That line falls between two diff hunks so it cannot be commented on inline, and it is cosmetic either way.

### Refuted

- **"The committed `docs/api/utils.md` is stale and the gate never ran to catch it."** Not stale. The only documented public symbol whose JSDoc this PR touches is `peginOutputCount`, and `docs/api/utils.md:1001-1010` carries the identical new text — the file's only hunk is `@@ -1001,10 +1001,10 @@`. The new psbt symbols (`deriveExpectedPrePeginHtlc`, `assertUnfundedPrePeginOutputLayout`, `PRE_PEGIN_*`) are not re-exported from `primitives/index.ts`, so they generate no entry. The gate would have passed had it run.
- **"The new `test-browser` target is a guard no CI job runs."** It is invoked at `verify.yml:96`, `package-release.yml:80` and `service-release-vault.yml:57`.
- **"`refund.ts`'s section 1 assertions were weakened when the CLAUDE.md wording was made vaguer."** The code is strictly stronger at HEAD: `assertPositiveBigintArray` on `pegInAmounts` (`:104-106`), WASM-vs-TS script equality (`:126-136`), template-vs-funded script (`:161-171`) and value (`:180-188`), exactly 1 input with txid/index/sequence binds (`:205-234`), exactly 1 output to the depositor's BIP-86 script (`:266-284`) at `htlcValue - refundFee` (`:291-299`), plus new version and locktime pins (`:192-203`).
- **"The key sort in `normalizeKeyGroup` may not match the Rust n-of-n ordering, so multi-keeper deposits would be rejected."** `PeginManager.test.ts:640-668` runs `preparePegin` end-to-end against real WASM with two keepers supplied out of order and compares the derived script against the WASM-encoded output. A sort mismatch fails that test. The narrower residual is the last inline comment below.
- **"`getPrePeginHtlcConnectorInfo` in `peginInput.ts:101-110` discards its result — dead code."** Load-bearing: `deriveExpectedPrePeginHtlc` takes no `vaultCoreVersion`, so this call is the only remaining graph-version validation on that path.
- **"The red `verify` check is a defect of this PR."** Pre-existing `main` breakage — `DepositForm.splitUnavailable.test.tsx:64` sets `hasUnconfirmedBalanceOnly`, absent from `DepositAmountState`; the file is in no PR's changed-file list and it reproduces on `main` at `8133b2b9e`.

### CI standing

`verify` aborted at that pre-existing error before reaching the SDK's suites, so none of the verification in the description — 1,825 Node tests, 29 Chromium tests, lint, the WASM package's 7 — is proven by CI on this head. Those are local figures. The `docs:clean` staleness gate likewise never ran; the docs check above is mine, by hand.

**Adjacent, not this PR:** `services/vault/src/utils/wasm.ts:7` still names `assertEncodedHtlcOutputsMatch` as the signing-path verifier when the primary guard is now `assertUnfundedPrePeginOutputLayout`. Carried over from the prior review, still open, outside the 28 files.

gbarkhatov (approved):

Issue #2383, opened 2026-09-01
Peg-in submit: adopt the contract config fingerprint, and make protocol-state reads atomic


gbarkhatov:
## Summary

The vault contracts are adding a **config fingerprint** argument to the peg-in submit functions ([vault-contracts-aave-v4#555](https://github.com/babylonlabs-io/vault-contracts-aave-v4/pull/555), open at time of writing). It closes a front-running / TOCTOU hole where protocol state can move between the moment we build the Bitcoin transaction and the moment our Ethereum registration is included in a block.

This is a **breaking ABI change to `submitPeginRequestBatch`**, which is the only peg-in submit path the dApp uses. We cannot deploy across the contract upgrade without adopting it.

While investigating, we found a **related bug that exists today and is independent of that PR**: the protocol state we build a deposit against is assembled from up to four separate, unpinned `latest`-block reads, one of which can be minutes stale. That is worth fixing on its own, and it is also a prerequisite for computing the fingerprint correctly.

Nothing is required from the Rust/WASM package, the Aave integration, or any backend service. Reasoning in [Scope](#scope-what-is-not-affected).

---

## Status

**Workstreams A, B and D are merged.** Everything that does not touch the contract call has shipped:

- **A** — [#2385](https://github.com/babylonlabs-io/babylon-toolkit/pull/2385), `c673adf48`. Deposit protocol state is read at one pinned block, the build no longer takes its configuration from the UI cache, and a guard stops the build when the cached and pinned snapshots disagree.
- **B** — [#2400](https://github.com/babylonlabs-io/babylon-toolkit/pull/2400), `b16d1fc5b`. The fingerprint encoder, pinned byte-for-byte against the vectors committed upstream.
- **D** — [#2403](https://github.com/babylonlabs-io/babylon-toolkit/pull/2403), `65e04515d`. The unversioned axis the drift guard structurally cannot see: the chosen amounts and the BTCVault count are re-validated against the pinned read, and the failure now says nothing was spent and leaves a restart that works.

[...]

PR #2400, opened 2026-09-03
feat(ts-sdk): peg-in config fingerprint encoder

gbarkhatov:
# Peg-in config fingerprint: the encoder

Part of https://github.com/babylonlabs-io/babylon-toolkit/issues/2383 (Workstream B). Follows https://github.com/babylonlabs-io/babylon-toolkit/pull/2385 (Workstream A), which is merged.

## Why we need this

When you deposit, the app builds a Bitcoin transaction that locks your BTC. The shape of that lock depends on protocol state that lives on Ethereum: which vault provider key is used, which keepers and challengers are in the list, and which parameter versions are active.

That state can move. If a key rotates, or a list changes, or a version is bumped in the seconds between the app building your Bitcoin lock and your registration landing on Ethereum, the contract would bond your vault to one set of values while your Bitcoin lock commits to a different set. The two no longer match, and the deposit is stuck.

The contracts are closing this by having the depositor send a **fingerprint** — a single hash over all the state the lock was built from. The contract recomputes that hash when the request is included and rejects it if the two differ. Nothing is signed or paid before the check, so a mismatch is a clean, early failure instead of a stranded vault.

For that to work, our TypeScript has to compute the exact same hash as the Solidity contract, byte for byte. **This PR is that computation, and nothing else.**

## What is in this PR

A new SDK module, `packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/`:

- `computePeginFingerprint(input)` — the hash to send with a registration.
- `encodePeginFingerprintPreimage(input)` — the bytes that get hashed, exposed separately so a mismatch can be traced to a specific field instead of being two hashes that just look different.
- The contract team's golden-vector file, copied in unchanged, plus tests that check our output against every entry.

## What is deliberately NOT in this PR

No call site. Nothing calls this yet, and the contract call is untouched, so **this PR cannot change the behaviour of the running app.** It is safe to merge and deploy at any time, and it does not wait on the contracts PR.

Also left out: the two new contract reads (`appKeeperKeyEpochCurrent`, `ucKeyEpochCurrent`) and reading the application entry point from the registry. Those exist on the deployed contract today, so we could have added them here — but nothing would call them until the next PR, and the repo's zero-dead-code rule is there for a reason. They go in the PR that actually uses them.

### Where the test vectors come from

**Option A — generate our own.** Write our own fixture, compute the expected hashes ourselves, commit them.

**Option B — use the contract team's committed vectors.** ← chosen

Option A means our test proves only that our code agrees with itself. If we misread the field order, we would generate vectors matching our misreading and the test would pass. Option B pins us to a file the contract team owns and tests on their side, so if the encoding ever moves, their CI notices and our copy stops matching. It also means the web app and the Rust `depositor-cli` are checked against one shared artifact rather than two separate transcriptions of it.

We copied their file **byte for byte** rather than retyping the values into TypeScript. That keeps re-checking it a one-line `diff` instead of a careful read, and it removes any chance of a typo in a 64-character hex string. The test header records the exact upstream commit and the command to verify the copy.

### How strict to be about field widths

The encoding is Solidity's `abi.encode`. A thing worth knowing before reviewing: under `abi.encode`, every field here is padded out to the same 32 bytes, so **the declared width of a number never reaches the output**. We confirmed this — encoding all ten fields as `uint256` produces an identical hash to the real widths.

So the only things that can actually break the hash are the **order** of the fields, the **number** of fields, and whether a type is fixed-size or variable-size. Review attention belongs on the field order, not on the widths. The widths are still declared, because they document what the contract says and they bound what we accept, but they are not what makes the bytes correct.

### How to test it

A golden-vector test that passes tells you very little on its own — a test that cannot fail also passes. So we did two things.

First, the vectors give **every field a different non-zero value**. This matters more than it sounds: if a fixture used `1, 1, 1, 1` for the four version fields, swapping two of them would produce the identical hash and a wrong field order would be undetectable. The contract team enforces this rule on their side with a test, which is another reason to use their file rather than ours.

Second, we deliberately broke the encoder and checked the tests caught it. The golden file runs 12 assertions. Swapping two field values fails 8 of them, dropping a field fails the same 8, and quietly using one field twice fails exactly one test, by name. The code was restored byte-identical after each.

Third, the vectors alone turn out not to be enough, and review caught this. Because every field is padded to the same 32 bytes, **changing a field's declared type upstream can leave all four vectors byte-identical** — every value in them fits the narrower type — so our golden test would stay green while our own range checks started rejecting values the contract accepts. We verified it: widening one field from `uint16` to `uint32` passes all 8 byte assertions and is caught by nothing else. So the test now also compares our declared field list against the type list the vector file itself publishes.

That check turns out to be genuinely complementary rather than belt-and-braces, and we measured both halves: a swap in the *values* fails the 8 byte assertions and not the type-list check, while a swap in the *declared list* leaves the bytes identical and fails only the type-list check. Neither catches the other's case.

We also copied the vector file's own rule — every numeric field distinct and non-zero — into a test on our side. It was enforced only in the contract repo, so a future re-copy with weaker vectors would have silently downgraded this whole file to "the encoder agrees with itself".

One honest note on what the *second* test file does and does not prove: its cases change one field at a time and assert the hash moves, which shows no field is silently ignored or aliased. **They do not pin the field order** — swapping two inputs still gives two different hashes even if the encoder has those same two fields the wrong way round. Order is pinned by the golden vectors and by the type-list check above, not by that file. It says so, so nobody takes it on trust.

## Three problems found along the way

All three would have compiled and looked correct. The first two we hit by running the code; the third came out of review.

**1. viem refuses one of the contract's own test addresses.** The `maximum_field_values` vector uses `0xfFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF`. That is a valid 20-byte address, but its upper/lower-case pattern is not a valid EIP-55 checksum. Solidity and Foundry do not care. viem does, and throws instead of encoding. The mixed case is only a display checksum over the same 20 bytes, so the encoder lowercases the address before encoding — which produces identical bytes. An address that is not 20 bytes of hex is still rejected loudly; this only relaxes the checksum, not the format. There is a test guarding it, because without one this breaks that single vector and nothing else.

The vault provider key is lowercased for a related but separate reason, found in review. viem copies the caller's hex casing straight into its output, so an upper-case key would still hash to the correct fingerprint — keccak hashes bytes, not text — but the *preimage* string would come back mixed-case and no longer compare equal to the contract's. Since the preimage exists precisely so a mismatch can be compared word by word, that would quietly defeat it.

**2. Reading the fixture the obvious way corrupts it.** Two epochs in the vectors are near the maximum a 64-bit number can hold, which is larger than JavaScript can hold in a normal number. A plain `import vectors from "./file.json"` silently rounds `18446744073709551615` up to `18446744073709551616` — and the encoder then correctly rejects it as out of range. The test reads the file as text and parses it keeping the original digits, with a comment explaining why, so nobody "simplifies" it back into an import later.

**3. The range checks did not check the type, so some bad values encoded silently.** Found in review. A check like `value < 0n || value > max` looks complete, but every comparison against `undefined`, `null` or `NaN` is `false`, so all three passed straight through — and `"7"` and `true` passed through *and then encoded*, as 7 and as 1. That is a well-formed fingerprint built from a value nobody supplied. TypeScript does not prevent it: a JS consumer, an object built from a partial snapshot, or an unawaited read all reach that code. The guards now check the type first, and there are tests for each case.

While fixing that we also tightened three fields that can only be zero if a read failed: `chainId`, the registry address, and `vaultCoreVersion` (the contract's own setter rejects 0, and the rest of the SDK already fails closed on it). Encoding any of those produces a fingerprint that can never match, so the deposit would fail on-chain with two opaque hashes instead of failing here with a named error. The three roster and params versions are deliberately left alone — 0 is a legitimate value on those axes.

## How to check this yourself

```bash
nvm use 24
pnpm --filter @babylonlabs-io/ts-sdk run test
```

1859 tests pass, including the build and the WASM-facade check that CI runs.

Confirm our copy of the vectors matches upstream:

```bash
gh api "repos/babylonlabs-io/vault-contracts-aave-v4/contents/test/data/pegin_fingerprint_vectors.json?ref=ec62ac62da7a590408ccdcd0a6339a2fa36126b0" \
  --jq '.content' | base64 -d \
  | diff - packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/__tests__/vectors/pegin_fingerprint_vectors.json
```

Confirm a vector independently, with Foundry rather than our code:

```bash
cast abi-encode 'f(bytes32,uint256,address,bytes32,uint64,uint64,uint16,uint16,uint16,uint16)' \
  0x7a76c7af21338f0474e4755a879f9ae8b6351b37d1a2e79e88e1ae95f20ea24b \
  1 0x5FbDB2315678afecb367f032d93F642f64180aa3 \
  0x79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798 \
  3 5 7 11 13 17 | cast keccak
# 0xe18d2cbf355cd69b44df7797eac6ad8c2dc522d328b63d20faa4f0be432cd73c
```

All four vectors were checked this way, and the built package (not just the source) was checked to reproduce them.

## Notes for review

- The API docs under `docs/api/` are generated. They are regenerated and included here because the SDK's public surface grew; CI fails if they are stale.
- The upstream contracts PR is still open at the commit we pinned. If it changes before this merges, the vector file needs re-copying. **Nothing automatic checks this** — no test or CI job re-fetches upstream, so our tests stay green against a stale copy. The pinned commit and the `diff` command in the test header make a re-copy *checkable by a person*, not automatic. Wiring that into CI belongs with the PR that actually depends on the value.
- The vault app is untouched.

## What comes next

The following PR adds the two contract reads, reads the application entry point from the registry instead of env config, computes the fingerprint from the same block-pinned snapshot the Bitcoin lock is built from, and passes it to `submitPeginRequestBatch`. That one changes the contract call, so it is held until the contract upgrade is deployed.

### DIFF
--- a/packages/babylon-ts-sdk/docs/api/README.md
+++ b/packages/babylon-ts-sdk/docs/api/README.md
@@ -33,6 +33,7 @@
 - [clients](clients.md)
 - [deposit-terms](deposit-terms.md)
 - [managers](managers.md)
+- [pegin-fingerprint](pegin-fingerprint.md)
 - [primitives](primitives.md)
 - [services](services.md)
 - [utils](utils.md)
--- a/packages/babylon-ts-sdk/docs/api/pegin-fingerprint.md
+++ b/packages/babylon-ts-sdk/docs/api/pegin-fingerprint.md
@@ -1,0 +1,273 @@
+[@babylonlabs-io/ts-sdk](README.md) / pegin-fingerprint
+
+# pegin-fingerprint
+
+Peg-in configuration fingerprint: the `keccak256(abi.encode(...))` commitment
+over the protocol state a Pre-PegIn transaction was built against, sent with
+the peg-in registration so the registry can reject a request whose protocol
+state moved between the build and inclusion.
+
+Reproduces `PeginLogic._peginFingerprint` from the vault contracts exactly;
+see [computePeginFingerprint](#computepeginfingerprint) for the encoding rules that matter.
+
+## Classes
+
+### PeginFingerprintInputError
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts)
+
+An input field is missing, malformed, or outside the width the contract declares for it.
+
+#### Extends
+
+- `Error`
+
+#### Constructors
+
+##### Constructor
+
+```ts
+new PeginFingerprintInputError(message): PeginFingerprintInputError;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts)
+
+###### Parameters
+
+###### message
+
+`string`
+
+###### Returns
+
+[`PeginFingerprintInputError`](#peginfingerprintinputerror)
+
+###### Overrides
+
+```ts
+Error.constructor
+```
+
+## Interfaces
+
+### PeginFingerprintInput
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts)
+
+Protocol state a Pre-PegIn transaction commits to, one field per encoded
+word.
+
+Declared in encoding order so the interface reads as the specification does.
+Reordering these fields does not change the output — the encoder names each
+one explicitly — but keeping them aligned makes a divergence from the
+contract visible on sight.
+
+#### Properties
+
+##### chainId
+
+```ts
+chainId: bigint;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts)
+
+`block.chainid` of the chain the registry is deployed on.
+
+##### registryAddress
+
+```ts
+registryAddress: `0x${string}`;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts)
+
+The `BTCVaultRegistry` address, which the contract encodes as
+`address(this)`. EIP-55 checksum casing is irrelevant to the encoded word
+and is not required here; see [encodePeginFingerprintPreimage](#encodepeginfingerprintpreimage).
+
+##### vaultProviderBtcKey
+
+```ts
+vaultProviderBtcKey: `0x${string}`;
+```
+
+Defined in: [packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-ts-sdk/src/tbv/core/pegin-fingerprint/peginFingerprint.ts)
+
+The vault provider's RESOLVED operation-BTC key, 32 bytes, `0x`-prefixed.
+
+The resolved key, not an epoch counter: the VP epoch is one registry-wide
+counter that ANY provider's key or payout append bumps, so committing to it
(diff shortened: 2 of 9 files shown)


jonybur (approved):
Encoder mirrors the Solidity contract's field order and count, with the four uint16 versions, two uint64 epochs, uint256 chainId, address, and bytes32 vault-provider key all validated to their declared widths. Domain constant is derived from the string so the pair can't drift; PEGIN_FINGERPRINT_ABI_PARAMETERS is intentionally not re-exported from the barrel and is used by the golden test to reconstruct the type list and compare against the vendored fixture's own declaration, catching a silent width change upstream that abi.encode would hide. Type-then-range check in assertUnsignedRange closes the JS-caller escape hatch for undefined/null/NaN/strings/booleans, and the parallel Number.isInteger check in assertUint16 covers the same shape for the number fields. Registry address is lowercased for EIP-55 rejection and zero-address is refused after normalisation; vault-provider key is normalised for the same preimage-comparability reason. VAULT_CORE_VERSION_MIN, chainId != 0, and non-zero registry gates are the right subset to tighten - the three roster/params versions are correctly left alone. No call sites yet so the change is inert at runtime, and the CLAUDE.md critical paths are untouched. Golden vectors are byte-vendored with the pinned commit and diff command in the header, and the field-coverage test is compile-checked against the input interface via the keyed Record. CI green.

jeremy-babylonlabs (approved):

Issue #2345, opened 2026-08-26
[SDK] Finish the remaining #2307 review cleanup


jonybur:
## Outcome

Finish the remaining #2307 review cleanup. This is optional SDK maintenance outside the Ethereum-first epic.

Status: all listed cleanup is delivered. #2466 merged the frozen comments, #2467 registered the public entry, and #2468 corrected the loader-test wording. #2469 and #2475 corrected the module and loader comments. #2464 updated the security-path text for the deprecated raw loader and its remaining bypass risk. #2446 delivered useful lazy-build error messages. Raw API safety remains in #2361.

## Problem

Stale comments, unused exports, and unclear test names make the SDK harder to maintain.

## Scope

- Remove duplicate stability comments.
- Remove or register unused public exports.
- Rename the loader test so it states what it proves.
- Correct stale code comments and module names.
- Update the security-path register text.

Group related cleanup in small PRs. Keep value-moving changes separate.

## Complete when

- [x] Lazy-build failures show useful messages. #2446.
- [x] Each remaining item is fixed or closed with a reason.
- [x] No PR mixes unrelated cleanup with value-moving code.

## Dependencies

None. Browser engine retry is complete in #2356. Raw API safety remains in #2361. Presign performance remains in #2358. This cleanup does not block their delivery.

PR #2401, opened 2026-09-03
test(wasm, ts-sdk): pin the payout connector wrappers

### LINKED ISSUE #2333: [eth-first] Test the browser Bitcoin engine boundary
**Epic:** #2228
**Parent:** #2231
**Status:** Blocks #2231
**Origin:** One browser-test gap predates #2307. #2307 added the other untested facade paths.

## Outcome

Tests fail if the browser Bitcoin engine or its production facade returns the wrong value.

## Missing tests

- Run the real browser engine entry.
- Check the `deriveVaultId` value through the facade used by production.
- Test the two presign connector wrappers directly.
- Pin safe retry behavior in the engine package.

## Complete when

- [ ] A broken browser entry fails a committed test.
- [ ] A wrong vault ID fails at the production facade.
- [ ] Both wrapper mutations fail tests.
- [ ] The engine package has a small test harness.

The known retry defect is #2356.

jonybur:
## Outcome

Closes #2333. The payout wrapper tests cover the browser entry and the built package facade.

## Change

- Pin assert-payout and payout-connector outputs against the shipped engine.
- Check the payout connector's mainnet address.
- Assert equal payout results for graph versions 1 and 3 through the built facade.
- Include the shared build repair from #2408.

## Safety

The change only adds and refines tests. All existing payout golden values stay the same. The final review fix adds no new test data.

## Verification

Local checks passed with Node 24:

- `pnpm nx run @babylonlabs-io/ts-sdk:test --skipNxCache`: build, 1,933 SDK tests, and 7 built facade tests.
- `pnpm nx run-many -t test lint -p @babylonlabs-io/babylon-tbv-rust-wasm --parallel=1 --skipNxCache`: build, 10 engine tests, and lint.
- `pnpm nx run @babylonlabs-io/ts-sdk:lint --skipNxCache`: passed.
- `git diff --check`: passed.

The updated branch still needs a green remote verification run and current code-owner approval.

## Links

Tracks #2228. #2228 stays open.

#2366 and #2367 cover the earlier boundary requirements. The unused public payout export remains in #2345.


packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
```diff
@@ -5,6 +5,7 @@ import { test } from "node:test";
 import {
   deriveVaultId,
   getAssertNoPayoutScriptInfo,
+  getAssertPayoutScriptInfo,
```
gbarkhatov, line 9: `createPayoutConnector` gets pinned through the browser entry but not here, and this is the layer that matters more of the two.

`src/index-node.ts:325` is not a re-export of `payoutConnector.ts:27` — it is a hand-duplicated second implementation that repeats the same six positional arguments into `WasmPeginPayoutConnector` and the same five getter reads. The package `exports` map routes the `node` condition to `dist/index-node.js`, and the SDK's `loadTbvWasm()` bare-imports the package, so Node consumers reach that copy while `wasm-loader.test.mjs` explicitly loads `dist/index.js`. The new browser pin therefore cannot see a drift in the node copy.

Nothing else pins it either. `primitives/scripts/__tests__/payout.test.ts:146` is a self-comparison — both sides route through the same node function, so transposing `depositor`/`vaultProvider` at `index-node.ts:333-334` changes both sides identically and still passes. `services/vault/src/__tests__/contracts/ledgerPayoutSigning.test.ts:196` uses it as a fixture builder, not as a pinned value.

That matters more here than for its sibling: `getAssertPayoutScriptInfo`'s output is re-bound to the real Assert prevout at the call site, whereas `createPayoutConnector`'s output is attached to payout input 0 with no such binding, so a silent argument-order drift reaches a signed PSBT. This PR pinned `getAssertPayoutScriptInfo` at both layers in the same change, so the asymmetry looks unintended rather than scoped.

Suggestion: add a `createPayoutConnector` block alongside the `getAssertPayoutScriptInfo` one, reusing the *same* params and goldens as the browser test — that turns browser/node divergence itself into a failure, rather than leaving two independent pins that cannot see each other.

I verified nothing is currently broken: the node entry reproduces the browser test's pinned values exactly today. This is unprotected, not wrong.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
```diff
@@ -267,3 +275,63 @@ test('pins getChallengeAssertScriptInfo through the browser entry', async () =>
...
+test('pins createPayoutConnector through the browser entry', async () => {
+  await withBrowserFacade(
+    async () =>
+      new Response(wasmBytes, {
+        headers: { 'Content-Type': 'application/wasm' },
+      }),
+    async (facade) => {
+      const result = await facade.createPayoutConnector(
+        payoutConnectorParams,
+        'signet',
```
gbarkhatov, line 312: `'signet'` does not actually pin the network argument, because signet and testnet are indistinguishable in the output.

I ran the wrapper across all four `Network` values against the shipped WASM. `scriptPubKey` and `taprootScriptHash` are network-invariant by construction, so `address` is the only assertion here that `network` can move at all — and signet and testnet both give `tb1p795tj5cunt8g6cuzghsqfcj4qattn93sqwgnxlskn3lmts656cwsk4p9uy`, since they share the `tb` hrp. A wrapper that mapped signet to testnet, or dropped the argument entirely, passes this assertion — which is precisely the forwarding the assertion is here to cover.

One extra assertion on `address` with a network whose hrp is unique closes it:
- `'regtest'` → `bcrt1p795tj5cunt8g6cuzghsqfcj4qattn93sqwgnxlskn3lmts656cwsmvtrf7`
- `'bitcoin'` → `bc1p795tj5cunt8g6cuzghsqfcj4qattn93sqwgnxlskn3lmts656cwspah2xt`

Both verified against the shipped WASM at the pinned commit. `'bitcoin'` has the mild edge of covering the network that carries real value.

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
```diff
@@ -47,16 +48,18 @@ test("pins the built public WASM facade", async () => {
     "f8d22e64c72a84a3dacdedb7d8b42e285bf06bd25850da911398c51d5a6c2dba",
   );
 
+  const assertConnectorParams = {
+    txGraphVersion: 3,
```
gbarkhatov: This file pins v3 while the browser file pins v1 (it reuses `connectorParams`, which is `txGraphVersion: 1`). That reads as deliberate differential coverage, but `txGraphVersion` is inert for these wrappers, so it supplies none.

Running the shipped WASM, `getAssertPayoutScriptInfo`, `getAssertNoPayoutScriptInfo` and `createPayoutConnector` all return byte-identical script *and* control block for v1, v2 and v3 under both files' parameter sets; only an out-of-range value throws (`unsupported tx graph version: 4 (supported: 1, 2, 3)`). So replacing `params.txGraphVersion` with a hardcoded `1` in either wrapper passes every assertion in both files.

Not a defect in what you wrote — the pins are correct either way. But given that `txGraphVersion` is the P2A-anchor discriminator elsewhere (0 sats for v1, 240 for v2/v3), the next reader will assume the v1/v3 split is load-bearing.

Suggestion: assert the property instead of implying it — one test calling `getAssertPayoutScriptInfo` at v1 and v3 and asserting the outputs are equal. That pins version-invariance as a real invariant, so if a future `VAULT_WASM_COMMIT` bump ever makes the assert-payout leaf version-dependent, the two files' goldens stop describing the same function and something fails loudly, instead of both pins quietly continuing to pass for the versions they happen to cover.

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
```diff
@@ -73,6 +76,21 @@ test("pins the built public WASM facade", async () => {
     },
   );
 
+  const payout = await getAssertPayoutScriptInfo(assertConnectorParams);
```
gbarkhatov: Minor, and pre-existing file shape rather than something this change introduces — but this is now the fourth unrelated pin inside the single `test("pins the built public WASM facade")`, and `node:test` aborts a case at the first failed assertion.

So this payout pin sits behind the `getAssertNoPayoutScriptInfo` `deepEqual` above it: a simultaneous regression in both surfaces one at a time, across repeated fix-and-rerun cycles. For a WASM value boundary that is the failure mode worth designing out, and `wasm-loader.test.mjs` already models the right shape — one `test()` per wrapper, each named for the wrapper it pins.

Splitting this file the same way would also remove the need for the hoisted `assertConnectorParams`, which is the one spot in the diff that trades self-contained, explicit-value test setup for reuse. Happy for this to be a follow-up; it is the least urgent thing on this PR.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
```diff
@@ -267,3 +275,63 @@ test('pins getChallengeAssertScriptInfo through the browser entry', async () =>
...
+      );
+      assert.deepEqual(
+        {
+          script: sha256Text(result.payoutScript),
+          controlBlock: sha256Text(result.payoutControlBlock),
```
gbarkhatov: Optional, and it cuts the other way from the rest: hashing *this* control block costs more than it saves.

`payoutControlBlock` here is 66 hex characters — single leaf, so it is just the leaf-version/parity byte plus the 32-byte internal key — which is *longer* than the 64-character sha256 that replaces it. I checked the length at runtime. Pinning it raw costs two characters, makes the internal key directly reviewable (you can eyeball that it is the NUMS point), and turns a failure from two opaque digests into a readable diff.

Hashing `payoutScript` at line 316 stays clearly justified — that one is 288 characters. And this does not apply to the assert connector's control block in the same file, which is 194 characters.

One related observation while I was measuring, not a request: control blocks collide across unrelated connectors here — the `createPayoutConnector` control block under a depositor/vaultProvider swap hashes to the same value already pinned as the `getChallengeAssertScriptInfo` control block in both test files. The control-block assertions carry less discriminating power than they look like they do; the script assertions are doing the real work.

gbarkhatov (dismissed):
Approving. Test-only, strictly increases coverage, and the pins are real.

I reproduced all three new golden sets independently by driving the shipped WASM at the pinned `VAULT_WASM_COMMIT`, and separately checked the `createPayoutConnector` values are arithmetically correct rather than merely self-consistent — NUMS internal key, single-leaf tree so merkle root == tapleaf hash == the pinned `taprootScriptHash`, and the taproot tweak reproduces the pinned `scriptPubKey` program with matching parity. The pinned address is a correct bech32m encoding of that witness program. The "generated, not hand-written" claim in the description holds.

The fixtures also discriminate: distinct keys per slot mean any positional-argument swap in either wrapper changes the pinned script. I confirmed sensitivity to `claimer`, `localChallengers`, `universalChallengers`, `timelockAssert` and `councilQuorum` for the assert connector, and to a depositor/vaultProvider swap, a keepers/universalChallengers swap and `timelockPegin` for the payout connector.

Five follow-ups inline, none blocking. The first is the one I would actually act on — the rest are precision notes.

**Measured, so nobody re-derives it:** running the shipped engine across all four networks, `signet` and `testnet` return byte-identical output (`scriptPubKey` and `taprootScriptHash` are network-invariant; only the address hrp moves — `bc1p…` / `tb1p…` / `tb1p…` / `bcrt1p…`). And `txGraphVersion` is inert for both newly pinned wrappers: v1, v2 and v3 return identical script *and* control block, with only an out-of-range value throwing.

**Checked and dismissed** — recording these so they don't come back:
- `assert.deepEqual` being loose: both files import `node:assert/strict`, so it is already strict.
- `sha256Text` silently hashing a non-string: the connector getters return hex strings, and `createHash().update()` throws on `undefined`/object — no `[object Object]` pin is possible.
- "v2/v3 are unpinned coverage gaps": measured above, the output is version-invariant, so there is nothing extra to pin.
- Sharing `assertConnectorParams` between the payout and noPayout pins as risky coupling: both getters funnel through the same `createConnector(params)`, so they are two reads of one connector and sharing the object is correct.

Note on `getPeginPayoutScriptInfo`: agreed it belongs in the #2345 cleanup. Worth recording that it returns byte-identical `payoutScript`/`payoutControlBlock` to `createPayoutConnector` for the same params, so the new pin already covers its bytes — it is a duplicate, not an untested surface.

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
jonybur: Added `pins createPayoutConnector through the built WASM facade` with the same parameters and expected values as the browser test: the four keys are the first four of that file's sorted list, named by role here. Confirmed the facade resolves to `dist/index-node.js`, and that the pin moves under a depositor/vaultProvider swap, a keepers/universalChallengers swap and a timelock change. Also added the mainnet address assertion on this side, for the same reason as the browser one.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
jonybur: Added `createPayoutConnector forwards the network to the address through the browser entry`, asserting the `bitcoin` address `bc1p795tj5cunt8g6cuzghsqfcj4qattn93sqwgnxlskn3lmts656cwspah2xt`. Recomputed against the shipped WASM at the pinned commit; it matches your value.

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
jonybur: Added `getAssertPayoutScriptInfo ignores the tx-graph version`: v2 and v3 outputs must deep-equal v1 for the same parameters. Checked against the shipped WASM: identical script and control block for 1, 2 and 3; 4 throws.

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
jonybur: Split the file into one test per wrapper with the parameters inlined in each, so the hoisted `assertConnectorParams` is gone.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
jonybur: Pinned `payoutControlBlock` raw: `c050929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0`, the `c0` byte plus the NUMS internal key. The assert-connector block stays hashed. One correction for the record: at 66 characters the raw block is two longer than the digest, not shorter; the readability argument stands on its own.

jonybur:

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
```diff
@@ -100,3 +155,57 @@ test("pins the built public WASM facade", async () => {
     },
   );
 });
+
+test("pins createPayoutConnector through the built WASM facade", async () => {
+  const result = await createPayoutConnector(
+    {
+      txGraphVersion: 1,
```
gbarkhatov: Small and optional — a consequence of what this round added rather than a re-raise of the version-invariance thread.

That thread asked for `getAssertPayoutScriptInfo`, and it got exactly that at `:113-128`. But adding it establishes version-invariance as a *documented, asserted* property of this surface, and against that backdrop `createPayoutConnector` is now the one wrapper where the version choice is stated and unasserted: this pin uses `txGraphVersion: 1` while every assert-wrapper pin in the same file uses `3`.

That is the wrapper where it matters slightly more — its output attaches to payout input 0 with no call-site rebinding, whereas the assert wrappers' output is re-bound to the real Assert prevout. As I measured last round, `createPayoutConnector` is version-inert too, so hardcoding `1` in the wrapper would still pass every assertion in both files.

One `assert.deepEqual` of the v3 result against the v1 result inside this test would close it symmetrically. Entirely your call — the pins are correct either way.

gbarkhatov (dismissed):
Follow-up pass over my earlier review (`5112093614`, at `1ce8a073`), not a second opinion on the same code. Delta is 8 commits; the two files in this PR both changed.

**All five prior findings are fixed.** I verified each against the code at HEAD rather than against the replies.

|---|---|---|---|
| 1 | `createPayoutConnector` pinned in the browser entry but not through the node facade — the hand-duplicated copy consumers actually reach | **fixed** | `tests/wasm-facade.node.mjs:159-192`, mainnet at `:194-211` |
| 2 | `'signet'` does not pin the network argument — signet and testnet share the `tb` hrp | **fixed** | `tests/wasm-loader.test.mjs:340-360` |
| 3 | The v1/v3 split reads as load-bearing differential coverage, but `txGraphVersion` is inert — assert the invariance instead | **fixed** | `tests/wasm-facade.node.mjs:113-128` |
| 4 | Four unrelated pins inside one `test()`; `node:test` aborts at the first failed assertion | **fixed** | `tests/wasm-facade.node.mjs` — seven named tests, params inlined |
| 5 | Hashing the 66-char single-leaf `payoutControlBlock` costs more than it saves — pin it raw | **fixed** | `tests/wasm-loader.test.mjs:320`, `wasm-facade.node.mjs:184` |

The existing threads stand as written; nothing there needs restating. Two points worth recording because they were the substance of the asks:

- **Finding 1 got what it asked for, not merely a second pin.** The node params resolve byte-identically to the browser file's — `payoutConnectorParams` reads `xOnlyKeys.sort()` positions 0-3, which are `2f8bde4d…`, `79be667e…`, `c6047f94…`, `e493dbf1…`, the same four the node file names by role — and all five goldens match string-for-string across the two files. So browser/node divergence is itself now a failure, which was the point.
- **Finding 5 landed with the right carve-out**: only the payout connector's control block is raw; the assert connector's stays hashed at `wasm-loader.test.mjs:230`, `:266` and `:290`.

One new low-severity note inline, arising from what the delta added rather than from the original review.

### CI standing

Not a finding, but worth saying plainly for a test-only PR whose entire value is that the tests run: `verify` is red for a reason unrelated to this change — `DepositForm.splitUnavailable.test.tsx:64` sets `hasUnconfirmedBalanceOnly`, which is not a field of `DepositAmountState`. That file is in no PR's changed-file list and the error reproduces on `main` at `8133b2b9e`; `main` is broken independently of this branch.

The consequence is specific here: the SDK's `test` script is `build && vitest run && node --test tests/wasm-facade.node.mjs`, so `wasm-facade.node.mjs` sits downstream of a build that never completed. Neither of the two files this PR adds tests to is proven to execute, let alone pass, on this head. Not a merge-blocker, and not something this branch can fix.

gbarkhatov (approved):

PR #2403, opened 2026-09-04
fix(vault): re-check deposit limits against the block the lock is built from

gbarkhatov:
# Deposit limits: check them against the chain, not against a five-minute-old copy

Part of https://github.com/babylonlabs-io/babylon-toolkit/issues/2383 (Workstream D). Follows https://github.com/babylonlabs-io/babylon-toolkit/pull/2385 (Workstream A) and https://github.com/babylonlabs-io/babylon-toolkit/pull/2400 (Workstream B), both merged.

## The problem

The deposit form checks your amount against the minimum and maximum the protocol allows. Those numbers come from a cached copy that is up to five minutes old, and nothing refreshes it while the form is open.

The actual Bitcoin transaction is built later, from parameters read fresh off the chain. If the minimum or maximum moved in between, nobody notices. The deposit is built anyway, you complete the whole signing ceremony, and then the registration fails on-chain because your amount is no longer allowed.

The guard added in [#2385](https://github.com/babylonlabs-io/babylon-toolkit/pull/2385) does not catch this, and cannot. It compares two version numbers, and the deposit limits do not have a version number — they come from a different part of the contract read, one that carries no version label at all. So the limits can move without either version number changing, and the guard sees nothing.

To be clear about scope: **this is not something [#2385](https://github.com/babylonlabs-io/babylon-toolkit/pull/2385) broke.** The same on-chain failure happened before it. That PR closed a gap between two snapshots; this one closes a gap between a snapshot and the chain. It was raised during that PR's review, deliberately deferred with the reasoning written down, and tracked here.

## What this PR does

Four things, all in the same story: catch the problem early, and make the recovery actually work.

**1. Re-check the amounts against the fresh parameters.** Right before the Bitcoin transaction is built, the amounts are validated again — this time against the numbers just read from the chain. The deposit stops there, where nothing has been signed and nothing has been paid.

**2. Check the number of BTCVaults too.** The contract limits how many BTCVaults one transaction can create. Nothing in the app read that limit at all — the only checks were hardcoded "maximum 2" literals, one of which is live on this very flow and still does not read the contract. It lives in the same unversioned part of the config as the amount limits, so it can drift the same way. The form now stops offering the split when the contract allows fewer outputs than a split needs, and the build re-checks it against the pinned read. Note the build-time branch is only reachable when the on-chain cap is exactly 1, since the cap is validated into `[1, 255]` on read.

**3. Make "start again" actually work.** The failure message says to restart. But restarting means closing and reopening the form, and the form reads that same five-minute cache — the one that just caused the failure. So you could be handed the identical error. The fresh parameters are now written into the cache when this fires, so the reopened form shows the limits that actually apply. For the BTCVault-count case the cache alone is not enough — nothing in the form read that limit — which is why the split-availability check now reads it too. Without that, "start again without splitting" was an instruction the form immediately undid by re-enabling the split.

**4. Say that nothing was spent.** This failure happened before anything was signed, broadcast or paid, but it was showing the same message as a much more expensive failure that happens after registration, where an Ethereum fee is already gone and a vault is stranded until it times out. Same words for both told nobody anything useful. The early failures now say plainly that nothing was signed and no funds were spent.

### Compare the limits, or re-check the amount?

**Option A — compare the cached limits against the fresh ones**, and stop if they differ. **Option B — re-check the chosen amount against the fresh limits.** ← chosen

Option A is the obvious one and it is wrong. If the maximum moves *up*, the two copies differ but your amount is still perfectly fine — and the deposit would be aborted for nothing. Option B catches a limit that tightened and ignores one that loosened, which is the only behaviour that matches what actually breaks. This was settled during the review of [#2385](https://github.com/babylonlabs-io/babylon-toolkit/pull/2385) and is implemented as agreed there.

### Write the check, or reuse the existing one?

The app already has a function that checks amounts against limits — the same one the deposit flow runs earlier against the cached numbers. We call that function with the fresh numbers instead of writing a second copy of the comparison.

This is not just tidiness. The two checks have to agree on what "within the limits" means, and the limits are inclusive at both ends — easy to get subtly wrong on a second writing, and a difference would reject deposits for a reason that has nothing to do with drift. It is called once per BTCVault rather than over the whole list, which also gives us the index, so the error can name the offending BTCVault without quoting the shared function's wording.

### One message, or two?

We started with one message for both new failures and it was wrong. The amount case and the BTCVault-count case need **opposite instructions**: one says change your amount, the other says do not split across two BTCVaults. A single message would give the wrong advice for whichever case it was not written for. The error now carries a tag saying which limit was hit, and there are two messages.

The same mistake turned up a second time, in the form. Making the split unavailable for a new reason meant the existing "N of M BTCVaults used" hint started explaining it — so a depositor holding 0 of 10 vaults could be told the per-position cap was why, which is simply untrue, and when that cap was unknown the split vanished with no explanation at all. Split availability now returns *why* it is unavailable rather than a bare boolean, and each reason has its own hint.

### What counts as a chain change, and what is our own bug

The shared amount checker also rejects an empty list and a zero amount. Those are bugs in our code, not governance changes, so they throw a distinct error type that maps to the generic failure message. Two things follow: the depositor is not told the chain moved when it did not, and the internal detail stays on the error for the bug report instead of becoming the message on screen.

### Refresh the cache, or clear it?

Clearing it means the next read goes back to the network, and the form can render the old numbers for a moment before the new ones arrive. We already hold the fresh numbers at the point of failure, so we write those straight into the cache. The reopened form is correct immediately with no refetch.

### Where the check lives

It could have gone into the existing guard file. It did not, because that guard answers a different question — it compares two snapshots to each other, while this one compares your choices against one snapshot. They fail for different reasons and produce different advice. That file's comment said this work "belongs with the recovery work rather than here"; it now names this module, and its claim that the gap is still open has been corrected, since it no longer is.

## What is deliberately not here

No changes to the deposit progress screen. It already shows a "Close" button for this kind of failure, which is the right thing, and [#2396](https://github.com/babylonlabs-io/babylon-toolkit/pull/2396) is currently rewriting that whole directory — editing it would mean a conflict for no gain. [#2290](https://github.com/babylonlabs-io/babylon-toolkit/pull/2290) (draft) also touches two of the files here, but those overlaps are small and additive.

## How to check this

```bash
nvm use 24
pnpm --filter @services/vault run test
pnpm --filter @services/vault run lint
```

3747 tests pass, up from 3708. Lint is clean (0 errors), and `tsc --noEmit` passes.

We also deliberately broke each new check and confirmed the tests caught it, rather than trusting a green run:

- Point the check at the cached parameters instead of the fresh ones → **81 tests fail**
- Remove the amount check → 6 fail
- Remove the BTCVault-count check → 2 fail
- Remove the cache refresh → exactly 1 fails, the test written for it
- Swap the two "which limit was hit" tags → 6 fail
- Drop the BTCVault cap from the split-availability rule → 2 fail
- Pick the form's hint from whether usage data is present instead of from the reason → 1 fails
- Untype the two "our own bug" errors so they render raw → 2 fail
- Refresh the cache on any error instead of only on drift → 1 fails
- Tag a caller-bug input as drift → 1 fails
- Put the deposit amount back into the error message → 1 fails

Every file was restored byte-identical afterwards.

That first result is the one worth understanding. The test fixture holds two sets of parameters — the fresh ones and the cached ones — and they deliberately **disagree**, with the cached copy set to values that would reject the test deposit. So the normal tests only pass while the check is reading the fresh parameters. Point it at the cache and 81 tests fall over. Without that disagreement the tests would pass either way and prove nothing, which is the exact trap that got caught during review of [#2385](https://github.com/babylonlabs-io/babylon-toolkit/pull/2385).

## Privacy note

Error messages from this guard reach Sentry as written — the logger scrubs only the browser console, not the reported exception. So the messages name the BTCVault index and the protocol limits, and never the depositor's amount, which the repo treats as identifying. There is a test asserting the amount does not appear.

## Notes for review

- No file in this PR is on the critical-path list, so the two-reviewer rule does not apply. Worth re-checking if the file list grows.
- The form checks your **total** against the limits, while the contract enforces them **per BTCVault**, and the split threshold comes from Aave rather than from the protocol. The flow already checked per BTCVault against the *cached* limits; what this PR adds on that axis is the *pinned* limits, not the per-BTCVault check itself. Moving the per-BTCVault check into the form is a separate change, noted on the issue.
- `validateVaultAmounts` in `services/deposit/validations.ts` is exported with zero callers — a pre-existing dead-code violation, in a file this PR does not touch. Separately, the **live** hardcoded `"Maximum 2 BTCVaults supported"` in that same file still runs on this flow and still never reads `maxHtlcOutputCount`. Both left alone deliberately; worth clearing together.
- The one place with no direct test is the JSX wiring in `SimpleDeposit.tsx` — that component has no test file, which predates this change. What it passes into the split-availability rule is compile-checked (the parameter is required), and the rule itself and the hint it drives are both unit-tested; but deleting the availability check from the form would compile and stay green.

### DIFF
--- a/services/vault/src/components/simple/DepositForm.tsx
+++ b/services/vault/src/components/simple/DepositForm.tsx
@@ -6,6 +6,7 @@ import { DepositButton } from "@/components/shared";
 import { getNetworkConfigBTC } from "@/config";
 import { COPY } from "@/copy";
 import { depositService } from "@/services/deposit";
+import type { SplitUnavailableReason } from "@/services/deposit/vaultCap";
 import type { VaultProviderListItem } from "@/types/vaultProvider";
 
 import { CollateralFactorRow } from "./CollateralFactorRow";
@@ -168,10 +169,11 @@ export interface DepositGatingState {
   /**
    * True when a single vault still fits but a 2-vault split would exceed the
    * cap — the deposit proceeds as a single vault and we surface the inline
-   * "vaults used / split unavailable" hint.
+   * "split unavailable" hint. The reason picks which hint: only the
+   * per-position cap can quote usage figures.
    */
-  vaultCapSplitUnavailable?: boolean;
-  /** Vault usage (used / cap) for the split-unavailable hint copy. */
+  splitUnavailableReason?: SplitUnavailableReason | null;
+  /** Vault usage (used / cap), set only for the per-position reason. */
   vaultCapUsage?: { used: number; cap: number };
 }
 
@@ -249,7 +251,7 @@ export function DepositForm({
     ordinalsCheckPending = false,
     isVaultCapReached = false,
     vaultCountCapUnavailable = false,
-    vaultCapSplitUnavailable = false,
+    splitUnavailableReason = null,
     vaultCapUsage,
   } = gatingState;
   const [openPanel, setOpenPanel] = useState<"split" | "provider" | null>(null);
@@ -450,9 +452,11 @@ export function DepositForm({
             }
           />
         )}
-        {/* Near the per-position vault cap: a split would overflow, so the
-            deposit proceeds as a single vault. Surface usage + why split is off. */}
-        {vaultCapSplitUnavailable && vaultCapUsage && (
+        {/* A split is not on offer, so the deposit proceeds as a single
+            BTCVault. Two unrelated caps can cause this and they need different
+            explanations — the per-position one can quote usage, the protocol
+            one applies even to an empty position. */}
+        {splitUnavailableReason !== null && (
           <div
             role="status"
             aria-live="polite"
@@ -463,10 +467,12 @@ export function DepositForm({
               className="mt-px shrink-0 text-accent-primary"
             />
             <span className="min-w-0 text-sm text-accent-secondary">
-              {COPY.deposit.maxVaultsReached.splitUnavailable(
-                vaultCapUsage.used,
-                vaultCapUsage.cap,
-              )}
+              {splitUnavailableReason === "per-position" && vaultCapUsage
+                ? COPY.deposit.maxVaultsReached.splitUnavailable(
+                    vaultCapUsage.used,
+                    vaultCapUsage.cap,
+                  )
+                : COPY.deposit.maxVaultsReached.splitUnavailableProtocolLimit}
             </span>
           </div>
         )}
--- a/services/vault/src/components/simple/SimpleDeposit.tsx
+++ b/services/vault/src/components/simple/SimpleDeposit.tsx
@@ -9,7 +9,10 @@ import { V3ModalShell } from "@/components/shared/V3ModalShell";
 import { FeatureFlags } from "@/config";
 import { useAddressScreening } from "@/context/addressScreening";
 import { useGeoFencing } from "@/context/geofencing";
-import { ProtocolParamsProvider } from "@/context/ProtocolParamsContext";
+import {
+  ProtocolParamsProvider,
+  useProtocolParamsContext,
+} from "@/context/ProtocolParamsContext";
 import { useBTCWallet, useETHWallet } from "@/context/wallet";
 import { COPY } from "@/copy";
 import { useBtcWalletState } from "@/hooks/deposit/useBtcWalletState";
@@ -177,6 +180,8 @@ function SimpleDepositContent({
     setFeeRate,
   } = useDepositPageFlow();
 
+  const { config } = useProtocolParamsContext();
+
   // Per-position BTC Vault cap (on-chain). Always-on value-protection guard:
   // block the deposit when even a single vault won't fit (`isAtCap`), force a
   // single vault when a split would overflow (`isSplitUnavailable`), and fail
@@ -189,11 +194,12 @@ function SimpleDepositContent({
     currentCount: collateralizableVaultCount,
     capUnavailable: vaultCountCapUnavailable,
   } = useVaultCountCap(connectedEthAddress);
-  const { isAtCap: isVaultCapReached, isSplitUnavailable: isSplitCapReached } =
+  const { isAtCap: isVaultCapReached, splitUnavailableReason } =
     resolveVaultCapState({
       existingVaultCount: collateralizableVaultCount,
       maxVaultsPerPosition: maxVaults,
       enabled: true,
+      maxHtlcOutputCount: config.maxHtlcOutputCount,
     });
 
   const isSupplementalDeposit = !!initialAmountBtc;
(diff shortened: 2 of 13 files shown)


jeremy-babylonlabs (approved):

jonybur (approved):
The gap this closes is real: the deposit bounds and the HTLC-output cap sit in the unversioned half of the config read, so `assertBuildConfigMatchesForm` cannot see them move. Re-validating the depositor's choices against the pinned read at the same call site catches a tightened minimum, a lowered maximum and a lowered vault cap before UTXO selection or signing.

Re-validating the choices rather than comparing the two bound sets is the right call — a maximum that moves up would trip a comparison for no reason at all, and would reject a perfectly valid deposit. Delegating to the SDK's `validateVaultAmounts` keeps the inclusive-bound semantics in one place; the caller-bug preconditions (empty array, non-positive leg) are correctly tagged as a distinct `BuildPreconditionError` so they map to the generic callout with the internal message stashed in diagnostics.

The cache seed on the drift path is well-argued and correctly narrow. It is gated on `isBuildConfigDriftError || isBuildLimitsDriftError` so an unrelated throw does not overwrite the key the polling hook also reads, the key is taken from `pegInConfigQueryOptions().queryKey` rather than a hand-written literal, and `setQueryData` avoids the refetch race that `invalidateQueries` would introduce. Four tests pin exactly this: drift + seed, non-drift + no seed, happy path + no seed, and rename-safety via the real factory.

The two new pre-signing callouts and the discriminated `splitUnavailableReason` are correctly separated: the amount case and the BTCVault-count case need opposite instructions, and `resolveVaultCapState` reports `per-position` first (the one the depositor can act on) with a clean fallthrough to `htlc-output-cap`. The DepositForm hint keys off the reason rather than `vaultCapUsage`, so the earlier "0 of 10 used" mis-attribution cannot recur; the invariant that `vaultCapUsage` is defined whenever the reason is `per-position` holds via the SimpleDeposit gating (`maxVaults != null` matches the early-null return in `resolveVaultCapState`).

The nothing-was-spent phrasing is factored to `NOTHING_SIGNED_OR_SPENT` and shared across the two new pre-signing errors, so a future edit cannot drift them out of the reassurance the post-registration callout deliberately does not carry. And the `BuildLimitsDriftError` mapping fallback to `depositLimitsChanged` when `.reason` is missing is a safe default with a comment explaining why.

Test set is thorough — the description's break-each-check-then-restore matrix is real evidence and the fixture disagreement between cached and pinned bounds is what makes the negative tests bite. The "protocol currently allows one BTCVault per transaction" copy is technically correct at present (`HTLC_OUTPUTS_PER_SPLIT_DEPOSIT === 2` and `maxHtlcOutputCount` validated to `[1, 255]` collapses the trigger to the single value 1); worth revisiting if that constant ever rises.

No critical-path files touched, so the two-reviewer rule does not bind. Note that the new test file uses `hasUnconfirmedBalanceOnly` on `DepositAmountState`, which #2395 removes — whichever PR merges second will need a one-line adjustment.

Issue #2402, opened 2026-09-04
Enforce the contract's `maxFundingInputCount` bound on the Pre-PegIn funding path


gbarkhatov:
## Summary

A new protocol parameter `maxFundingInputCount` bounds the number of inputs a Pre-PegIn transaction may declare. It is enforced on-chain at peg-in submission. The frontend chooses that input count and currently applies no bound, so once the parameter is set a depositor with a fragmented wallet will complete the full BTC signing ceremony — PoP, N PegIn-input device approvals — and only then hit a revert at Ethereum registration, with an error we cannot decode.

This issue covers reading the bound, enforcing it before the ceremony starts, keeping the Max button honest, and decoding the revert.

## Background

Origin: [baby-auditor-findings#747](https://github.com/babylonlabs-io/baby-auditor-findings/issues/747) (MEDIUM) — unbounded Claim and Pre-PegIn inputs amplify peer-supplied transactions into arbitrary Bitcoin-RPC work. A depositor's Pre-PegIn bytes travel in the `PegInSubmitted` event, and every VP, VK and UC daemon then issues one `getrawtransaction` per declared input, so a single Ethereum submission fans an RPC storm across the whole cluster with no admission control on the event-poller path.

Both upstream halves have merged:

- Contracts — [vault-contracts-aave-v4#569](https://github.com/babylonlabs-io/vault-contracts-aave-v4/pull/569)
- Daemons — [btc-vault#2725](https://github.com/babylonlabs-io/btc-vault/pull/2725)

## Verified upstream behaviour

| Property | Value |
| --- | --- |
| Field | `uint8 maxFundingInputCount`, appended **last** to `IProtocolParams.TBVProtocolParams` (index 7 of 8) |
| Enforcement | `PeginLogic.submitRequest`, ahead of `hashBtcTx` and `getOutputs` |
| Error | `TooManyFundingInputs(uint256 inputCount, uint256 maxAllowed)`, selector `0xf14dcc9f` |
| Comparison | Exclusive — `count > max` reverts, so exactly `max` is accepted (same convention as `maxHtlcOutputCount`) |
| What is counted | All **declared** inputs, read straight from the CompactSize counter of `unsignedPrePeginTx` via a new `BTCUtils.getInputCount`. No funding-vs-protocol filtering, and the counter is never checked against the transaction body |
[...]

PR #2404, opened 2026-09-04
feat(ts-sdk): send the peg-in config fingerprint with the registration

### LINKED ISSUE #2383: Peg-in submit: adopt the contract config fingerprint, and make protocol-state reads atomic

gbarkhatov:
# Send the peg-in config fingerprint with the registration

Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2383

## The problem

A Pre-Pegin HTLC output is built for a specific set of participants. Their public keys are baked into the Bitcoin script, along with `tRefund` from the offchain params and a PegIn output shape chosen by the vault core version.

We read all of that from Ethereum, build and sign the Bitcoin transactions, then call `submitPeginRequestBatch` to register the deposit. **The contract does not trust the values we used** — it resolves them again, live, at inclusion. If a key rotated, a roster changed, or a version was bumped in between, the contract bonds the vault to one set of values while our Bitcoin transaction commits to another. The deposit can then never complete: every daemon rejects it, because the funded HTLC does not match what the vault is bonded to.

Same class of bug as the vault-provider commission front-running we already fixed with `maxAcceptableCommissionBps`. The depositor commits to terms, and the terms move underneath them.

The contracts are closing this by adding a fingerprint argument — a hash over the protocol state the deposit was built against, which the registry recomputes at inclusion and compares. This PR sends it.

## What changed

**We now compute the fingerprint and put it on the wire.** [#2400](https://github.com/babylonlabs-io/babylon-toolkit/pull/2400) built the encoder and pinned it against the contract's own test vectors; nothing called it until now. The value is computed in `useDepositFlow` right after the participant keys resolve, from the same block-pinned reads that shape the Bitcoin scripts, and threaded down to `encodeFunctionData`.

**Three contract reads are new.** The fingerprint commits to two operation-key epochs we never read before, so `appKeeperKeyEpochCurrent` and `ucKeyEpochCurrent` are now read. The third, `getVaultProviderApplication`, is *not* a fingerprint field — it resolves which application's keeper roster, roster version and keeper epoch the deposit is bonded to, and it is what the equality guard below checks. All three already exist on the deployed contract, so the read half works against today's testnet.

**The application entry point now comes from the registry.** The dApp carries one from its own configuration; the contract uses `getVaultProviderApplication(vaultProvider)`. The two agree today, but they are different sources of truth, and that value picks which keeper roster, roster version and keeper key epoch a deposit is bonded to. We now read the registry's value, use it, and refuse to build if the configured one disagrees.

**`PeginFingerprintChanged` gets its own error and its own message.** It is classified when the revert carries data, which is the gas-estimate case — there the registration transaction is never sent. The depositor has already approved the peg-in signatures by then, so the copy cannot claim "nothing was signed". What it can say, and what holds on *both* revert paths, is that the Pre-Pegin has not been broadcast and no Bitcoin was spent. A rotation landing between the estimate and inclusion mines as a revert whose data `sendAndWait` discards; that case falls through to the generic bucket — see [#2498](https://github.com/babylonlabs-io/babylon-toolkit/issues/2498).

**The orphaned batch gas estimator is deleted.** More on that below.

### Where the three new reads live

The fingerprint is only correct if every field describes the *same* observation of the chain that shaped the Bitcoin scripts. A fingerprint computed from fresher state would be **accepted by the contract while our scripts were stale** — silently reintroducing the exact bug the fingerprint exists to prevent. So "where do the reads live" is really "how do we make that structural rather than conventional".

| Option | Why not |
|---|---|
| Add them to `getPegInConfiguration`'s multicall | That multicall backs the display path too, behind `ProtocolParamsContext`. It runs `allowFailure: false`, so one getter missing from a split deployment fails every protocol-param read and blanks the app. `ProtocolParams.abi.ts` already records that reasoning for `getPeginActivationDelay`. It also only solves one of the three reads — the other two live on different contracts. |
| A new SDK service that reads and computes | It would resolve the application entry point independently of the one `validateOnChainParticipantKeys` uses, creating a second source of truth for the value that keys the keeper roster. That divergence is the failure this PR exists to prevent. |
| Read them in the app | No reader exposes these getters, so the app would hand-roll contract calls with no compile-time link to the pinned block — the exact mistake being closed. |

**Chosen: fold them into `validateOnChainParticipantKeys`.** It is the only function whose `blockNumber` is already *required* rather than optional, it already resolves four of the nine fingerprint inputs, and the two epochs belong next to the keys they label — `appKeeperKeyEpoch` is the epoch the keeper operation keys were read at. Putting them anywhere else creates a place where an epoch and the keys it names could come from different blocks.

### Replacing the application entry point, or checking it

Silently substituting the registry's value would be simpler. It is also the wrong shape: the configured entry point drives provider selection, the pending-vault record, and every post-deposit Aave call. If the registry disagreed and we quietly used its value, the deposit would **succeed** — creating a vault under an application the dApp cannot operate, with nothing failing loudly. That is a silent fallback on a critical path.

So it is checked, not replaced, and the check runs before any roster is read. It gets no user-facing copy of its own: a depositor cannot act on a deployment or configuration fault, and inventing an instruction for them would be worse than the generic callout. But it does get a **typed** error, because an untyped one is not neutral — the mapper's last-resort bucket renders an unrecognised message verbatim as the callout body, so a plain `Error` would have put two raw addresses and three protocol values in front of the depositor. The typed error maps to the generic copy with the detail kept in diagnostics, the same shape `BuildPreconditionError` already uses one bucket above. The encoder's own input errors are mapped alongside it for the same reason.

### Giving the revert an actionable message

The SDK already maps 18 contract reverts to message strings, so adding a 19th was the smaller change. It was rejected because the app has to *branch* on this one — it is the only revert on the peg-in path with a specific recovery — and finding it by matching an English message across a package boundary produces a string nobody can safely reword.

A typed `PeginFingerprintChangedError` instead, recognised by a predicate the way `isParticipantKeyDriftError` already is. It carries the two hashes from the revert payload, which are the difference between "the chain moved" and "our encoder is wrong". Those go into diagnostics for a bug report, never in front of a depositor.

### The batch gas estimator

`estimateSubmitPeginRequestBatchGas` synthesised dummy calldata before the depositor signed anything. A dummy fingerprint reverts the simulation, so it could not be left alone.

Passing a real fingerprint would have kept it, but it destroys the function's whole premise — "estimate before the depositor has signed" — because the fingerprint needs the pinned build snapshot. That leaves a shape nobody can use, in a file that requires two reviewers.

It was added in `da9ab83e4` (2026-05-20) for an ETH network-fee row on the deposit form. Its only caller was deleted three days later in `f109057bf`. It has had zero callers and zero tests for three and a half months. **Deleted** — about 130 lines out of `PeginManager.ts`, plus its dummy-request builders and constants. It is a breaking change to a public export that no hand-written guide teaches.

## Notes for review

**The argument lands in the middle of the batch call, not at the end.** `expectedFingerprint` is argument 4, before `requests`; on the two singular overloads it is appended last. Order is the signature and the signature is the selector, so both are pinned by tests against literals taken from upstream, not re-derived from our own ABI — a re-derivation would only prove the ABI agrees with itself.

| Signature | Before | After |
|---|---|---|
| `submitPeginRequestBatch` | `0x68d177ac` | `0x3e62a2ba` |
| `submitPeginRequest` (no referral) | `0x67df3144` | `0x564feec1` |
| `submitPeginRequest` (referral) | `0x1fb3c964` | `0x1c3c4606` |
| `PeginFingerprintChanged(bytes32,bytes32)` | — | `0x846c25bb` |

**These are confirmed against upstream's own generated artifact, not against a transcription.** The contracts repo regenerates `snapshots/selectors.md` from its own contracts. On contracts `main` — [vault-contracts-aave-v4#555](https://github.com/babylonlabs-io/vault-contracts-aave-v4/pull/555) is now merged — that file lists every value above, plus the three new getters this PR reads: `appKeeperKeyEpochCurrent(address)` `0x7b1efb8e`, `ucKeyEpochCurrent()` `0x9a839841`, `getVaultProviderApplication(address)` `0x6f6262b2`. All seven match our ABI, including the batch tuple's component list. They were also re-derived independently with viem, and a test asserts our ABI reproduces them.

The upstream head moved twice while this branch was open (`ec62ac62` → `86577e40` → merged as `da430fa45`). The fingerprint half is **byte-identical** across both moves — the batch signature, both singular overloads, `_peginFingerprint`'s encode block, the error definition and the golden-vector blob all unchanged. Those moves were `main` merges, not spec changes.

**One existing test could not tell two sources apart, and this change exposed it.** `useDepositFlow.test.tsx` asserted that `preparePeginTransaction` receives `MOCK_PARAMS.vaultProviderBtcPubkey` — the indexer *hint*. Production correctly passes the *resolved* operation key. The assertion only ever passed because both were the same literal. They are now deliberately different, the assertion names the resolved key, and a fingerprint built from the wrong source fails.

**An overload hazard worth knowing about.** The two singular overloads had 11 and 12 inputs, so viem disambiguated on arity; they now have 12 and 13, which is still unambiguous. But a partial edit or a bad rebase that updates one and drops the other makes both 12, and viem falls back to resolving by argument type — silently, and possibly correctly. That is why the singular selector pin is load-bearing in a way it was not before.

**Tests that would stay green if this were wired wrong**, so nobody mistakes them for coverage: `PeginManager.test.ts:1963-2007` uses substring `toContain` on raw calldata; `PeginManager.test.ts:1811` indexes `args[6]` of the singular call, which the appended argument does not move; and every `registerPeginBatchAndWait` mock in `useDepositFlow.test.tsx` accepts anything. The two differential fingerprint tests are what actually constrain it — and for that to hold, all four `uint16` version slots in the fixture carry different values, because two slots sharing a value encode identically and a swap between them would pass.

**Pre-existing gap, not fixed here.** If the fingerprint changes between the gas estimate and mining, `sendAndWait`'s reverted-receipt branch throws a synthetic error carrying no revert data, so no selector maps and the depositor sees "check a block explorer" with the ETH fee spent. This affects all 18 mapped selectors and deserves its own issue.

## What a local review round changed

A multi-agent review of this branch raised ten findings, two of them merge-blocking. All ten are fixed here; the two that mattered are worth naming because both were **wrong prose about on-chain binding**, which is the failure mode this whole series exists to avoid.

**A JSDoc claimed the fingerprint commits to the application entry point. It does not** — the encoder takes nine fields and none of them is an entry point. `validateOnChainParticipantKeys` was returning the registry-resolved value with a comment saying the fingerprint consumed it, and nothing read it. A future reader would have believed that value was committed on-chain and skipped adding a check that is genuinely absent. The field is now gone; the getter is still read, but only for the equality guard and the keeper roster, which is what it is actually for.

**The entry-point mismatch would have shown the depositor a developer sentence.** The comment said it lands on "the generic callout with diagnostics". It did not: the mapper's last-resort bucket renders an unrecognised message verbatim as the callout body, so two raw addresses and three protocol values would have gone in front of a user, and a user-facing string would have lived outside `copy.ts`. It is now a typed error mapped to the generic copy, alongside the encoder's own input errors, which had the same problem.

A second pass then caught that the first fix was incomplete: deleting the interface field left a test fixture still writing it, which broke the vault build. Neither `test` nor `lint` sees that — `vitest` does not typecheck and eslint does not do excess-property analysis — so **`build` is now in the verification list above**, and it is the target to run before trusting a green branch.

The rest: a fixture where two `uint16` slots shared a value, so a swap between them encoded identically and the differential test could not see it; the two new epoch getters shipping with no ABI signature pin despite feeding fingerprint words 5 and 6; both quickstart snippets passing an identifier they never define; a case-sensitive selector comparison; `isAddressEqual` parsing a value the app only casts, so a malformed entry point threw viem's error instead of ours; a test that duplicated its neighbour; an asymmetric validation order; and two nits.

## Testing

Every new guard was mutation-tested — broken deliberately, confirmed to fail the right test, then restored:

| Mutation | Caught by |
|---|---|
| Fingerprint moved to the end of the batch args | batch selector pin |
| Entry-point equality check disabled | two `validateOnChainParticipantKeys` cases |
| Pinned block dropped from the keeper epoch read | the same-block assertion |
| `0x` prefix dropped from the vault provider key | `peginFingerprintInput` mapping tests |
| Typed predicate replaced with a substring match | the same-worded-different-name case |
| All-zero fingerprint guard removed | two client rejection cases |
| Fingerprint replaced with a constant in the hook | both differential deposit-flow tests |
| Entry-point mismatch reverted to an untyped `Error` | the typed-recognition test |
| The generic-copy bucket removed from the mapper | both message-hiding tests |

```
pnpm --filter @babylonlabs-io/ts-sdk run test   # 118 files, 2001 tests
pnpm --filter @services/vault run test          # 315 files, 3831 tests
pnpm run lint
pnpm --filter @babylonlabs-io/ts-sdk run build      # typecheck: vitest does not
pnpm --filter @services/vault run build             # typecheck: nor does eslint
pnpm --filter @babylonlabs-io/ts-sdk run docs:clean
```

Two reviewers required: this touches `PeginManager.ts` and `pegin-registration-client.ts`, both on the critical-path list.

## Deployment sequencing

Adding the argument changes the function selector, and there is no backwards-compatible path — it breaks in **both** directions. The new call does not exist until the contract upgrades, and the old call stops existing the moment it does: `submitPeginRequestBatch(address,address,uint16,(...)[])` `0x68d177ac` is gone rather than superseded, and so is the 11-argument singular overload. Whichever side moves first stops being able to submit, and every submission fails at `estimateGas`. That is what makes this a lockstep flip rather than a rollout.

Upstream [vault-contracts-aave-v4#555](https://github.com/babylonlabs-io/vault-contracts-aave-v4/pull/555) **merged on 2026-09-04** (`da430fa45`), so `expectedFingerprint` is on contracts `main`.

**Where each environment stands.** Probed directly — EIP-1967 implementation slot, then `eth_getCode`, then a selector search. `tbv-networks` is not a reliable signal here: these are UUPS proxies, so an implementation upgrade does not move the proxy address and the recorded `vaultContractsRef` can lag reality.

| Environment | Registry implementation | `0x3e62a2ba` (new) | `0x68d177ac` (old) |
|---|---|---|---|
| devnet | `0x60b790ee…` | present | gone |
| staging | `0x1eee661b…` | absent | present |
| testnet | `0x387b4bb1…` | absent | present |

**Devnet has flipped, and this branch is the only build that can submit a deposit there.** The devnet site, serving `main`, currently cannot submit at all — it calls a selector the contract no longer has. Unblocking that is what this PR is for, and the run below proves it against the live contract rather than against pinned vectors.

Staging and testnet still carry the old selector, and promotion to those environments is a manual step — merging does not move them. They flip on the agreed procedure, **pause contracts → upgrade web → upgrade contracts → resume**, once their contracts are ready. In-flight peg-ins are unaffected either way; only new deposits use this path.

The acceptance criterion — exercising a build end to end against a contract carrying the fingerprint — **is now met on devnet**; see [What the devnet run showed](#what-the-devnet-run-showed). Everything else in the issue was already covered.

## What the devnet run showed

The acceptance criterion is met: a real-wallet peg-in ran end to end from this branch against devnet on 2026-09-10, and the registration was accepted by a contract that carries the fingerprint.

```
node services/vault/scripts/sync-env.mjs devnet
pnpm --filter vault run e2e:cli --target=localhost --network=devnet \
  --btc=unisat --eth=metamask --action=pegin --vp=vault-provider-0 --yes
```

0.01 sBTC — the protocol minimum — through `vault-provider-0`. The registration transaction is `0x677781fe3d3d460b8c93ae5b2f224898c2ea43bca55e6f3df610a4215e582a33`, to the devnet registry `0xC394d336cA31267e979F772845d8edF94b8BD941`:

| | |
|---|---|
| selector | `0x3e62a2ba` — the new batch signature |
| status | success — 1,437,379 gas, 3 logs |
| `expectedFingerprint` | `0x4176f979de7411a6a67b6fca48e8b508e058e212db84359206d4e1391160998b` |

Read back off-chain and decoded against this branch's own ABI, the call is `submitPeginRequestBatch(address,address,uint16,bytes32,(bytes32,bytes,bytes,bytes,bytes32,uint8,uint32,bytes,bytes32)[])` with the fingerprint as argument 4. It went out as a real keccak rather than a zero default — worth checking explicitly, because a zero would also be accepted by a contract that did not enforce, and the point of the run is to distinguish those. The registry recomputed the value at inclusion and it matched.

The flow then advanced past `SUBMIT_PEGIN` into the Bitcoin broadcast and the vault-provider steps, which is only reachable once the registration succeeds — so the selector flip works against a live deployment, not only against the pinned vectors.

**The UTXO-consolidation warning was over-cautious.** `selectUtxosForPegin` sorts candidates by value descending and returns as soon as the accumulated total covers the amount plus fee, so it never walks into the tail. The run's wallet held 102 UTXOs of which 98 were 546-sat dust, and the funded Pre-Pegin took **two** inputs. Fragmentation only approaches [#2402](https://github.com/babylonlabs-io/babylon-toolkit/issues/2402)'s 20-input bound when the large UTXOs do not cover the deposit on their own. The issue is real; it is much less likely to derail a test run than stated above.

**Still outstanding: the negative run.** A green deposit proves the fingerprint *matched*. The mismatch path — the reason this exists — needs a deliberate run that perturbs a byte of the computed fingerprint before it reaches `registerPeginBatchAndWait`, confirming the depositor gets the mapped "Protocol configuration changed" callout with both hashes in diagnostics rather than raw revert hex. Note the copy is only reachable on the gas-estimate path; a mismatch that mines lands in the generic bucket, which is [#2498](https://github.com/babylonlabs-io/babylon-toolkit/issues/2498).

## Review round two

One inline finding, and it was correct. The bucket comment claimed this revert is "raised from the gas estimate … so nothing reached either chain". That covers one of the two paths and presented it as the only one: if the fingerprint rotates between the estimate and inclusion, the transaction mines as a revert, and `sendAndWait` throws a fresh `Error` built from a template string — revert data gone, so `extractErrorData` finds no selector and the typed bucket never fires. The depositor gets the generic callout having spent gas.

The copy was unreachable on that path, so nothing false was ever shown; the defect was a comment asserting a path cannot happen when it can. Fixed by correcting the comment and module docstring, and by rewording the copy so every claim holds on **both** paths — "before it was submitted" and "no fee was paid" are gone, since they are true only of the estimate path. What remains — the Pre-Pegin has not been broadcast and no Bitcoin was spent — holds either way, because registration precedes broadcast.

Recovering the revert data itself is filed as [#2498](https://github.com/babylonlabs-io/babylon-toolkit/issues/2498) rather than done here: it is pre-existing, it affects all 18 mapped selectors rather than this one, and the fix belongs in a critical-path file. The precedent already exists in `transactionFactory.ts`'s `throwRevertError`, which replays via `eth_call` at the mined block.

Also riding along in that deploy, and out of scope here: [#570](https://github.com/babylonlabs-io/vault-contracts-aave-v4/pull/570) makes `activateVaultWithSecret` unwind-and-succeed instead of reverting when the application callback fails, so our activation step will report success for a vault the contract redeemed back. Not lockstep — it needs its own issue and can land on `main` any time.


packages/babylon-ts-sdk/src/tbv/core/contracts/abis/BTCVaultRegistry.abi.ts
```diff
@@ -166,6 +178,14 @@ export const BTCVaultRegistryABI = [
...
         internalType: "uint16",
       },
+      // The fingerprint is argument 4, BEFORE `requests` — not appended, as it
+      // is on the two singular overloads. It takes no per-request input and the
+      // batch fixes one vault provider, so one value covers every entry.
+      {
+        name: "expectedFingerprint",
+        type: "bytes32",
+        internalType: "bytes32",
+      },
```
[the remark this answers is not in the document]
gbarkhatov: Correct, and intended — this is the PR's premise rather than a defect in it.

Backwards compatibility was ruled out upstream on contract-size grounds: keeping both selectors alive would need a large contract refactor. The agreed procedure is instead pause contracts → upgrade web → upgrade contracts → resume. A dual-ABI runtime branch was considered and rejected, because it puts a conditional on the one code path where taking the wrong branch means a malformed registration.

The constraint is governed rather than latent: the description opens with **DO NOT MERGE YET**, carries a "Why this cannot merge yet" section, and the PR is blocked on review with no auto-merge.

One status change since this was flagged. The contract PR [merged on 2026-09-04](https://github.com/babylonlabs-io/vault-contracts-aave-v4/pull/555) (`da430fa45`), so the new selectors now exist on contracts `main`. But no environment has them yet — `tbv-networks` still pins devnet to `2e87a85a` (2026-08-18) — so the gate is the deployment window, not the upstream merge. Both directions break, which is what makes it a lockstep flip: the old selector is gone rather than superseded, so whichever side moves first stops being able to submit a deposit.

No code change. Resolving.

jonybur (dismissed):

jeremy-babylonlabs (dismissed):

services/vault/src/utils/errors/depositErrors.ts
```diff
@@ -218,6 +247,25 @@ export function mapDepositError(err: unknown): DepositErrorContent {
     return ERRORS.participantKeyDrift;
   }
 
+  // 3b'. The registry's own fingerprint check, raised from the gas estimate
```
jrwbabylonlab: A fingerprint rotation after a successful gas estimate mines as a revert, and `sendAndWait` throws a generic error with no revert data — so this typed mapping never fires for that case and the user gets generic copy. The “nothing reached either chain” comment overstates it.

gbarkhatov: You're right, and thank you — I traced it and the mechanism is worse than "no revert data" suggests.

When the transaction mines as reverted, `sendAndWait` throws `new Error("Batch transaction reverted. Hash: 0x… Check a block explorer for details.")` — a fresh error from a template string, with no `cause`, no `.data`, and no receipt attached. `extractErrorData` then has nothing to find, and its last-resort message scan cannot help either: the regex is `/\b(0x[a-fA-F0-9]{8})\b/`, and the only `0x` in that message is a 64-character tx hash, where `\b` fails after the first 8 hex characters. So the typed bucket never runs and the depositor lands in the generic fallback having spent gas.

One nuance worth recording, because it changes what needed fixing: **the copy is unreachable on that path**, so nothing false was ever shown to a depositor. The defect was the comment asserting a path cannot happen when it can — which is worse than a wrong string, because the next person to touch this would trust it.

What changed:

- The bucket comment and the module docstring now say this classifies only when the revert carries data — the estimate case — and that a rotation landing between the estimate and inclusion falls through to the generic bucket.
- The copy is reworded so every claim holds on **both** paths, rather than only the one that reaches it today. "declined the registration before it was submitted" and "no fee was paid" are gone; both are true only of the estimate path. What remains — the Pre-Pegin has not been broadcast and no Bitcoin was spent — holds either way, because registration precedes broadcast. The point is that it stays correct if the mined path is ever routed into this bucket.

Recovering the revert data itself is filed as https://github.com/babylonlabs-io/babylon-toolkit/issues/2498 rather than done here. It is pre-existing and affects all 18 selectors the SDK maps, not this one, and the fix belongs in a two-reviewer critical-path file. The precedent is already in the repo: `transactionFactory.ts`'s `throwRevertError` replays via `eth_call` at the mined block, with an out-of-gas guard so an OOG transaction is not misread as a success. `sendAndWait` has every input that needs in scope.

jrwbabylonlab (approved):

jonybur (approved):
Traced the fingerprint from the pinned reads through to the calldata. The registry address and chain id come from the same config the registration is sent to, the two epochs stay `bigint` end to end, and `computeBuildPeginFingerprint` is handed the same `validatedKeys` / `buildConfig` objects that shape the Bitcoin scripts — so the hash cannot describe a fresher state than the lock it labels.

The selector test is the right shape: it pins all three signatures and the error against upstream's generated `snapshots/selectors.md` rather than re-deriving them from our own ABI, and it checks the two epoch getters' declared output width, which is the failure that would otherwise surface only as an opaque on-chain revert. Entry-point resolution now comes from the registry and fails closed on a mismatch before any roster is read.

The removed `estimateSubmitPeginRequestBatchGas` has no remaining consumers in the tree, and the regenerated API docs are committed.

PR #2405, opened 2026-09-04
feat(ts-sdk, vault): add the Risky health-factor band

jonybur:
## Outcome

Vault users can distinguish positions close to liquidation from moderate risk.

## Change

- Below 1.0: Liquidatable. From 1.0 through 1.1: Risky. Above 1.1 through 2.0: Moderate. Above 2.0: Safe.
- The 2.0 display threshold applies across the vault app, including loan summaries and borrow, repay, and withdrawal reviews.
- The red borrow warning applies to Risky and Liquidatable projections.
- Collateral with no debt has a green health indicator. The Overview card shows Very Safe.
- A position with debt must have a health factor. Missing data cannot show a safe state.

## Safety

The 1.1 display threshold marks positions close to liquidation. The enforced borrowing floor stays at 1.05. Transaction checks and signing rules stay unchanged.

## Verification

Local SDK checks passed: build, 1,968 tests, and the WASM facade checks. All 3,758 vault unit tests passed. SDK and vault lint checks passed. API documentation was regenerated.

## Links

[Review discussions](https://github.com/babylonlabs-io/babylon-toolkit/pull/2405/files)


services/vault/src/applications/aave/components/LoanCard/Borrow/index.tsx
```diff
@@ -225,7 +225,9 @@ export function Borrow() {
   );
   const showAtRiskCallout =
     hasProjection &&
-    (projectedHealthStatus === "warning" || projectedHealthStatus === "danger");
+    (projectedHealthStatus === "risky" ||
```
gbarkhatov: **The "At risk of liquidation" callout widened from HF < 1.5 to HF <= 2.0, not just to the Risky band.**

This condition gains `risky`, but the `warning` arm stayed while `HEALTH_FACTOR_WARNING_THRESHOLD` moved 1.5 to 2.0 — so `warning` now spans (1.1, 2.0] and the callout fires across all of it.

Concretely: a borrow landing the user at a projected HF of 1.9 now renders `COPY.loans.atRiskOfLiquidation`, where before it was silent. The same PR paints that position amber "Moderate" on the risk card, so the two surfaces now disagree about whether 1.9 is worth warning about.

The description says the callout "now also covers the Risky band"; it in fact picked up the whole 1.5-2.0 range as well. That works against this PR's own goal of making the red warning mean something.

Gating on `"risky" || "danger"` — the two red bands — would match the stated intent. Alternatively give the callout its own constant rather than borrowing the display threshold, since the two are now answering different questions.

packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts
```diff
@@ -54,10 +54,17 @@ export const AAVE_BASE_CURRENCY_RAY_DECIMALS = 53;
...
  */
-export const HEALTH_FACTOR_WARNING_THRESHOLD = 1.5;
+export const HEALTH_FACTOR_RISKY_THRESHOLD = 1.1;
+
+/**
+ * Health factor warning threshold.
+ * A position above HEALTH_FACTOR_RISKY_THRESHOLD and up to and including this
+ * value is "warning". A position above this value is "safe".
+ */
+export const HEALTH_FACTOR_WARNING_THRESHOLD = 2.0;
```
gbarkhatov, line 67: **Moving this 1.5 to 2.0 recolors six surfaces beyond the risk card, and the description names only the card.**

This constant is the sole input to the `warning` band, and `getHealthFactorColor` (warning to amber) is consumed by `LoansSummary.tsx:47`, `VaultsSummaryCard.tsx:49`, `Liquidations/PositionOverview.tsx:40`, `BorrowMetricsCard.tsx:57`, `RepayDetailsCard.tsx:37` and `WithdrawFlow/HealthFactorDelta.tsx:28`.

A depositor sitting at HF 1.8 who has changed nothing sees green flip to amber on all six next load. The withdraw review is the sharpest case: at a projected HF of 1.5 the number now renders amber while `withdrawEligibility.ts` still shows no at-risk text, because that gates on 1.1 — so the user gets a warning colour with no accompanying explanation.

None of this is necessarily wrong; it may be exactly what's wanted. But if 2.0 is meant only as the Overview card's Moderate/Safe seam, the card should read a card-local constant. If it is meant globally, worth saying so in the description so the other five surfaces get looked at deliberately.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/healthFactor.ts
```diff
@@ -29,7 +40,8 @@ export function getHealthFactorStatus(
   if (!hasDebt) return "no_debt";
   if (healthFactor === null) return "safe";
```
gbarkhatov: **A debt-bearing position whose health factor is null renders the most reassuring state on the screen.**

This line is pre-existing, but it sits in this PR's hunk and the band set is being reworked around it, so it's the moment to look at it. `null` with `hasDebt` true returns `"safe"`, and `getRiskDisplayState` then maps `safe` + null to `"verySafe"` — green, with the infinity glyph.

This is not an unreachable defensive branch. The only production caller, `useAaveUserPosition.ts`, derives `healthFactor` from `accountData.borrowCount > 0n` but `hasDebt` from `debtValueUsd > 0` — two different on-chain fields. Whenever they disagree, this branch is what runs, and the user with debt is told they are Very Safe. `healthFactor.test.ts` currently pins that reassuring answer as expected behaviour.

The no-silent-fallbacks rule in `CLAUDE.md` is aimed squarely at this: absence of a required value is an error, not a reason to default. An `unknown` member in the status union, or surfacing the absence, would say what is actually known.

packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts
```diff
@@ -54,10 +54,17 @@ export const AAVE_BASE_CURRENCY_RAY_DECIMALS = 53;
 export const WAD_DECIMALS = 18;
 
 /**
- * Health factor warning threshold
- * Positions below this are considered at risk of liquidation
+ * Health factor risky threshold.
+ * A position from 1.0 up to and including this value is "risky".
  */
-export const HEALTH_FACTOR_WARNING_THRESHOLD = 1.5;
+export const HEALTH_FACTOR_RISKY_THRESHOLD = 1.1;
```
gbarkhatov, line 60: **The Risky band overlaps the range the app still lets you borrow into, so the stated rationale doesn't quite hold.**

The description justifies 1.1 on the grounds that such a position "cannot borrow more", but `MIN_HEALTH_FACTOR_FOR_BORROW` is still 1.05 and is the only enforced floor. A position at HF 1.06-1.10 is therefore labelled red "Risky" and is still allowed to borrow — and `calculateMaxBorrowTokens` will size "Max" down into that band by design, so the primary borrow affordance produces a position the UI immediately flags red.

Two ways out, and they differ in intent rather than correctness: pin the risky threshold to the borrow floor so label and gate agree, or keep 1.1 and record the real reason for it in the JSDoc, which currently gives none. The second is fine — it just shouldn't be the "cannot borrow more" reason, because that one is checkable and doesn't hold.

services/vault/src/components/simple/RiskSection/railLayout.ts
```diff
@@ -19,7 +20,7 @@ export function getRiskDisplayState(
   if (!hasPosition) return "noPosition";
   switch (status) {
     case "no_debt":
```
gbarkhatov, line 22: **Remapping `no_debt` to `verySafe` makes the risk card disagree with every other surface for the same position.**

`getHealthFactorColor("no_debt")` still returns `HEALTH_FACTOR_COLORS.GRAY` in `healthFactorDisplay.ts`, and that file is untouched here. Before this change both sides read as muted and agreed; this line creates the divergence.

So collateral with no loan now renders a green infinity card on the Overview and a gray heart in `LoansSummary` and `VaultsSummaryCard` — same position, same instant, two different readings of how safe it is.

Returning `GREEN` for `no_debt` in `getHealthFactorColor` would bring them back into line. Leaving the risk card muted would too. Either is fine; the split is the problem.

services/vault/src/components/simple/RiskSection/RiskPriceRail.tsx
```diff
@@ -104,7 +104,7 @@ export function RiskPriceRail({
     Math.abs(currentPct - liquidationPct) >= GAP_MIN_SEPARATION_PCT;
 
   const currentRingClass =
-    state === "liquidatable"
+    state === "liquidatable" || state === "risky"
```
gbarkhatov: **Non-exhaustive ternary whose fallback is the safest colour.**

Adding `risky` required a hand edit here precisely because the chain ends in a catch-all `"border-risk-green"`. Any future `RiskDisplayState` gets a green ring on a possibly-liquidatable position, with no compile error to catch it.

`RiskSection.tsx` already does this the safe way in the same PR: `Record<RiskDisplayState, ...>` turns a missed state into a build failure. The same shape here — a `Record<RiskDisplayState, string>` ring map — would make the next band addition impossible to get wrong.

It already disagrees with itself, which is the tell: `noPosition` falls through to a green ring while `STATE_STYLE.noPosition` paints the label muted.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/healthFactor.ts
```diff
@@ -54,8 +66,9 @@ export function getHealthFactorStatusFromValue(
  * **Formula:** `HF = (Collateral × Liquidation Threshold) / Total Debt`
  *
  * Health factor determines liquidation risk:
- * - `>= 1.5` - Safe (green)
- * - `1.0 - 1.5` - Warning (amber)
+ * - `> 2.0` - Safe (green)
+ * - `1.1 - 2.0` - Warning (amber)
```
gbarkhatov: **The published band table is ambiguous at exactly the two boundaries this PR exists to create.**

`1.1 - 2.0` Warning and `1.0 - 1.1` Risky both claim 1.1, and `> 2.0` against `1.1 - 2.0` leaves 2.0 unclear. The code is unambiguous — 1.1 is Risky, 2.0 is Warning — but the prose isn't.

It matters more than a normal comment nit because this block is what typedoc emits, so the ambiguity is reproduced verbatim in `docs/api/integrations/aave.md` and in the hand-maintained table in `docs/integrations/aave/README.md`.

The module header at the top of this same file already gets it right, with "up to and including". Reusing that wording here and re-running `docs:clean` would carry the fix into both generated docs.

services/vault/src/applications/aave/utils/healthFactorDisplay.ts
```diff
@@ -20,6 +20,7 @@ export function getHealthFactorColor(
       return HEALTH_FACTOR_COLORS.GREEN;
     case "warning":
       return HEALTH_FACTOR_COLORS.AMBER;
+    case "risky":
```
gbarkhatov, line 24: **The new `risky` colour branch has no test.**

`__tests__/healthFactorDisplay.test.ts` contains no reference to `getHealthFactorColor` at all — it covers only `formatHealthFactor`. The switch is TS-exhaustive, so the case can't be silently dropped, which caps how bad this can get; but it's a new colour branch on a liquidation-warning surface with no behavioural coverage.

One line closes it: `expect(getHealthFactorColor("risky")).toBe(HEALTH_FACTOR_COLORS.RED)`.

services/vault/src/components/simple/RiskSection/railLayout.ts
```diff
@@ -48,8 +51,8 @@ const MAX_TICKS = 8;
 // the liquidation price and green at the price clearing the safe threshold.
 // Far from liquidation that ramp collapses to a hairline, hence the minimum —
 // which trades truthfulness for legibility: once it engages the stops no longer
-// mark the HF bands. At HF 9 ($63,488 / $6,962) the honest green stop is 5.3%
-// but lands at 45.3%, so ~40pp of rail reads amber for comfortably safe prices.
+// mark the HF bands. At HF 9 ($63,488 / $6,962) the honest green stop is 10.3%
+// but lands at 45.3%, so ~35pp of rail reads amber for comfortably safe prices.
```
gbarkhatov: Minor: the gradient's dependence on the warning threshold is pinned by nothing.

This comment's worked example is rewritten 5.3% to 10.3% because the green stop is `liquidationPriceUsd * HEALTH_FACTOR_WARNING_THRESHOLD`, but the only covering test asserts `greenStop - redStop >= 45`, which `GRADIENT_MIN_RAMP_PCT` guarantees for any threshold value at all.

So the stop could be wired to 1.5, 2.0 or 5.0 and every test would still pass — the comment is currently the only record of the coupling, and comments drift. One case where the min-ramp doesn't engage, asserting the green stop lands on `liquidation * HEALTH_FACTOR_WARNING_THRESHOLD`, would pin it.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/healthFactor.ts
```diff
@@ -29,7 +40,8 @@ export function getHealthFactorStatus(
   if (!hasDebt) return "no_debt";
   if (healthFactor === null) return "safe";
   if (healthFactor < 1.0) return "danger";
```
gbarkhatov, line 44: Nit, and only because this is the PR that made its neighbours constants: the two lines below now read `HEALTH_FACTOR_RISKY_THRESHOLD` and `HEALTH_FACTOR_WARNING_THRESHOLD`, while this one hardcodes `1.0`.

Unlike those two, `1.0` isn't a product choice — it's Aave's protocol liquidation rule, which is arguably the one of the three most worth naming. A `HEALTH_FACTOR_LIQUIDATION_THRESHOLD` constant would document where it comes from and why it, unlike the others, is not ours to tune.

gbarkhatov (approved):
Combo review — three independent passes, deduped and verified against the PR head.

**No merge-blockers, and the core of the change is correct.** The band operators are right at all four boundaries — `< 1.0` danger, `<= 1.1` risky, `<= 2.0` warning, else safe — which matches the description exactly, and the tests pin all four seams for both entry points. That was the likeliest place for this PR to be wrong and it isn't.

What the findings are about is blast radius rather than banding. Two of them are worth a decision before merge: the `HEALTH_FACTOR_WARNING_THRESHOLD` move from 1.5 to 2.0 is described as a risk-card change, but it is the sole input to the `warning` band and so recolors six other surfaces and widens the borrow screen's "At risk of liquidation" callout to a range the same PR paints amber "Moderate". Neither is a defect exactly — they may be intended — but the description names only the risk card, so they would land unreviewed.

### Refuted

Raised by one of the passes and killed on verification:

- **"Off-by-one at a band boundary."** The code places HF 1.0 and 1.1 in Risky, 2.0 in Warning, 2.01 in Safe — the description's five boundaries exactly. `getHealthFactorStatusFromValue` delegates to `getHealthFactorStatus`, so the two entry points cannot drift.
- **"The `docs:clean` regeneration is partial."** Reading the committed `docs/api/integrations/aave.md`: it carries the new union member, a full `HEALTH_FACTOR_RISKY_THRESHOLD` section, `HEALTH_FACTOR_WARNING_THRESHOLD: 2 = 2.0`, and JSDoc matching `constants.ts` character-for-character. No stale `1.5` survives anywhere in the docs.
- **"`COPY.risk.status.noPosition` is now orphaned dead copy."** The `!hasPosition` branch in `getRiskDisplayState` still returns `"noPosition"`, and `RiskSection.test.tsx` still exercises it.
- **Anchor error, substance kept.** Two passes placed `getRiskDisplayState` in `healthFactorDisplay.ts`. It lives in `railLayout.ts`; the finding below is re-anchored there.

### Verified clean

- The threshold move is **display-only**. `validateBorrowAction`, `validateBorrowPreSign`, `withdrawEligibility` and `calculateMaxBorrowTokens` all compare numerically against `MIN_HEALTH_FACTOR_FOR_BORROW` / `WITHDRAW_HF_BLOCK_THRESHOLD`, never against a status string. No money-moving path changes.
- Every status-to-color consumer routes through the single TS-exhaustive `getHealthFactorColor` switch. There is no second, unupdated status map in `services/vault/src`.
- No band became unreachable: `HEALTH_FACTOR_HEALTHY_THRESHOLD` is 50, so `safe` still covers (2.0, 50] and `verySafe` covers > 50.
- Copy is compliant: `risky: "Risky"` sits in `copy.ts` alongside its siblings, sentence-case, with no new inlined string in a component.
- No registered critical path is touched, and `services/vault/SECURITY_MODEL.md` carries no invariant bearing on display banding.

**CI caveat worth stating:** `verify` is red for a reason unrelated to this PR — `DepositForm.splitUnavailable.test.tsx:64` sets `hasUnconfirmedBalanceOnly`, which is not a field of `DepositAmountState`; that file is in no PR's changed-file list and the error reproduces on `main` at `8133b2b9e`. The consequence here is specific: the build aborted before this PR's type-check, unit tests, **or the `docs:clean` staleness diff** ever ran. The docs check above is mine, by hand, not CI's.

gbarkhatov (dismissed):

services/vault/src/applications/aave/components/LoanCard/Borrow/index.tsx
jonybur: The callout now applies only to Risky and Liquidatable projections. The Moderate band keeps its amber display without the red warning.

packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts
jonybur: The 2.0 threshold applies to all health-factor displays. The description now names the loan summaries and borrow, repay, and withdrawal reviews. The numeric transaction limits stay unchanged.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/healthFactor.ts
jonybur: A debt-bearing position with a missing health factor now raises an error. The position hook uses totalDebtValueRay for both debt detection and health-factor availability. The existing missing-value test now checks the error.

packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts
jonybur: The 1.1 threshold marks positions close to liquidation. It is independent of the 1.05 borrowing floor. The JSDoc and description now state that distinction.

services/vault/src/components/simple/RiskSection/railLayout.ts
jonybur: The no_debt status now uses green on the shared health-factor displays. This matches the Overview card.

services/vault/src/components/simple/RiskSection/RiskPriceRail.tsx
jonybur: The marker colors now use an exhaustive Record. The no-position marker and rail use the neutral display.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/healthFactor.ts
jonybur: The tables now say > 1.1 and <= 2.0 for Moderate, and >= 1.0 and <= 1.1 for Risky. The API docs were regenerated.

services/vault/src/applications/aave/utils/healthFactorDisplay.ts
jonybur: The color test now checks Risky in red and no debt in green.

services/vault/src/components/simple/RiskSection/railLayout.ts
jonybur: The rail test now checks both exact stops where the minimum ramp does not apply. With liquidation at 60,000, the green stop must represent 120,000.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/healthFactor.ts
jonybur: The fixed 1.0 protocol boundary stays inline to keep this change small. Its meaning is already stated in the module documentation and pinned by the boundary tests. The product display thresholds remain separate constants.

gbarkhatov (approved):

PR #2406, opened 2026-09-04
fix(vault): stack the split-deposit progress lanes

jonybur:
## Outcome

Each vault in a split deposit uses a full-width progress lane. The lanes appear one above the other.

## Change

- Stack the vault lanes and remove the old narrow-column options.
- Let confirmation and provider detail rows wrap. Keep each value close to its label.
- Check the shared trunk and the split lanes at 360, 390, and 1280 pixels.
- Check the scenario readout and use the recorded wallet chain settings.

## Safety

The change affects progress display and its tests. Deposit steps, signing rules, and transaction checks stay unchanged.

## Verification

Six Playwright layout checks passed locally. They cover the shared trunk and split lanes at all three widths. CI does not run this Playwright project. Vault lint passed. Local vault unit checks passed: 3,754 tests, with 6 skipped.

## Links

[Review discussions](https://github.com/babylonlabs-io/babylon-toolkit/pull/2406/files)


services/vault/src/components/simple/DepositProgressView/__tests__/SplitGroupedProgress.test.tsx
```diff
@@ -10,7 +10,7 @@ import { buildStepItems, getVisualStep } from "../steps";
 const steps = buildStepItems(null);
 
 describe("SplitGroupedProgress", () => {
-  it("renders a labelled column for each vault in a split deposit", () => {
+  it("renders a labelled lane for each vault in a split deposit", () => {
```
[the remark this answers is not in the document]
jonybur: Added browser-level coverage in `e2e/deposit-progress-layout.spec.ts`. It drives the dev-only god-mode panel to inject two batched demo deposits, opens the real progress card at the desktop and mobile (390px) widths, and asserts that the second lane starts below the first with the same left edge, that each lane spans the card, and that no child extends past the card's right edge or grows a page scrollbar. The spec fails when the lanes are put back side by side (checked by mutation). It runs in the behavioural suite under a `chromium-god-mode` project with its own dev server, because the panel is compile-time gated.

jonybur:

services/vault/e2e/deposit-progress-layout.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+const PANEL_VIEWPORT = { width: 1280, height: 900 };
+
+/**
+ * Slider position of the "Awaiting payout transactions" scenario in the
+ * panel's Normal segment (one position per DepositFlowStep, in enum order).
+ * The first per-vault wait: both lanes open a group and render a detail
+ * panel, and unlike the activation scenarios nothing auto-advances under the
+ * assertions.
+ */
+const AWAITING_PAYOUT_SCENARIO_INDEX = 7;
```
gbarkhatov: **The spec seeds a state in which the restructured trunk region never renders.**

Position 7 is `AWAIT_PAYOUT_TRANSACTIONS`, i.e. visual step 8. At `currentStep = 8`, `buildStepGroups` marks the only trunk group (`registerDeposit`, steps 1-6) `"completed"` — `currentStep > group.endStep` — and `SplitGroupedProgress` filters completed groups out, so `trunkGroups` is empty and the whole `trunkGroups.length > 0 && ...` region is skipped, along with the `gap-6` that now separates trunk from lane 1.

The module docstring says it measures "one lane per vault **under the shared trunk**", which is not what this position measures. It also never renders `BtcConfirmationDetail` — the widest row, the one carrying the hash, and one of the three panels this PR added `flex-wrap` to.

A trunk-phase position (slider 5, "Step 6: Awaiting BTC confirmation") renders trunk, both lanes and the hash row in one shot. The index itself is correct as written — this is about which screen it lands on, not an off-by-one.

services/vault/e2e/deposit-progress-layout.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/** The progress card shell: the nearest clipping ancestor of the lanes. */
+function progressCard(page: Page): Locator {
+  return lane(page, 1).locator(
+    "xpath=ancestor::div[contains(@class, 'overflow-hidden')][1]",
+  );
+}
+
+for (const viewport of VISUAL_VIEWPORTS) {
```
gbarkhatov: **The spec never measures 360px — the one width the central claim is about.**

`VISUAL_VIEWPORTS` is exactly `{desktop 1280x800, mobile 390x844}`. The description says the layout was verified "at 360, 390 and 1280px" and that "nothing spills past the card at 360px"; the docstring above claims "every supported width". Neither holds as committed.

360 is the narrowest common phone width and the one where the `0/3` counter overflow was reported. Neither `GroupHeader` nor `StepRow`'s label row was hardened, so the fix rests entirely on lanes being full-width — which makes 360 the measurement that actually carries the argument.

Worth giving this spec its own viewport list that adds 360 rather than borrowing `VISUAL_VIEWPORTS`: that constant is the *screenshot* target set, so trimming it later to save capture minutes would silently shrink this layout guard too.

services/vault/playwright.config.ts
```diff
@@ -81,11 +114,20 @@ export default defineConfig({
...
     {
       name: "chromium",
+      testIgnore: BEHAVIOURAL_TEST_IGNORE,
       use: {
         ...devices["Desktop Chrome"],
         baseURL: `http://localhost:${PORT_FULL_ENV}`,
       },
     },
+    {
+      name: "chromium-god-mode",
```
gbarkhatov, line 124: **Nothing in CI runs this project, so the PR's only automated regression guard is never executed.**

Playwright is invoked by exactly one workflow, `visual-regression.yml`, and it drives `playwright.visual.config.ts` — `testDir: e2e/visual`, `testMatch: **/*.visual.spec.ts`. `e2e/deposit-progress-layout.spec.ts` matches neither. That workflow copies `playwright.config.ts` only because the visual config imports `MOCK_ENV_VARS` from it (its own comment says so); it never runs it. `test:e2e` exists in `package.json` but no workflow calls it.

So `chromium-god-mode` is local-only. Compounding it, `verify` aborted at the pre-existing `main` type error before reaching this PR's own tsc and unit run.

Either wire the project into a job, or say in the description that the measurement was local — as written, "verified with tsc, eslint, the full vault unit suite" reads as CI evidence.

services/vault/e2e/deposit-progress-layout.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+  const visible = { visible: true };
+  await page.getByLabel("Inject demo").filter(visible).check();
+  await page.getByLabel("Hide real items").filter(visible).check();
+  // The panel starts with one deposit mock; a second one makes the pair.
+  await page.getByRole("button", { name: "+ Add mock" }).click();
+
+  for (const position of [1, 2]) {
+    await page
+      .getByLabel(`Mock ${position} step`)
+      .fill(String(AWAITING_PAYOUT_SCENARIO_INDEX));
```
gbarkhatov: **Nothing pins that the slider landed on the intended scenario, so a reordered gallery degrades this guard silently.**

The fill happens and the spec proceeds. Because the slider is `offset + value` within the current mode segment rather than an absolute step address, inserting or reordering one entry in `PENDING_FLOW_SCENARIOS` seeds a different but equally stable state — two lanes still render, nothing overflows, and the spec stays green while no longer covering the screen it was written for.

`e2e/visual/depositProgress.visual.spec.ts` solves exactly this one commit earlier, and its comment gives the same reasoning: it asserts `getByLabel(PANEL.state(position))` matches the expected `Step N: ...` readout immediately after the fill. Reusing that readout assertion here would close the gap.

services/vault/e2e/deposit-progress-layout.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+  injectPageWallets,
+} from "./fixtures/pageWallets";
+import { installRecordedBackend } from "./fixtures/replay";
+import { RECORDED_DEPOSITOR } from "./fixtures/replay/contracts";
+import { VISUAL_VIEWPORTS } from "./visual/targets";
+
+import { MOCK_ENV_VARS } from "../playwright.config";
+
+/** Sepolia, as a hex quantity - the chain the recording was made on. */
+const SEPOLIA_CHAIN_ID_HEX = "0xaa36a7";
```
gbarkhatov: **`SEPOLIA_CHAIN_ID_HEX` hand-duplicates a value that already has a helper deriving it from the recording.**

`e2e/visual/capture.ts` exports `recordedPageWallets()`, which returns precisely this object and computes the chain id as `` `0x${Number(RECORDED_DEPLOYMENT.ETH_CHAIN_ID).toString(16)}` `` instead of hardcoding it. The spec already imports from `./visual/targets`, so `./visual/capture` is equally reachable.

Concrete failure: re-record on another chain and `RECORDED_DEPLOYMENT.ETH_CHAIN_ID` updates while this constant keeps announcing `0xaa36a7`. The app renders a wrong-network state and the walk dies at an unrelated locator with no hint of the cause.

`await injectPageWallets(page, recordedPageWallets())` would let the constant go.

services/vault/e2e/deposit-progress-layout.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+ * it again. Controls are addressed the way a person finds them - by label -
+ * because the panel carries no testids; only the visible tab's controls are
+ * targeted, since every tab stays mounted behind `hidden`.
+ */
+async function injectTwoVaultDemoDeposit(page: Page): Promise<void> {
+  await page.getByRole("button", { name: "God mode", exact: true }).click();
+  await page.getByRole("button", { name: "Deposit & Vaults" }).click();
+
+  const visible = { visible: true };
+  await page.getByLabel("Inject demo").filter(visible).check();
```
gbarkhatov, line 53: **The panel's accessible names are re-typed here after `depositProgress.visual.spec.ts` already collected them in one place.**

That spec keeps the launcher, deposits tab, "Inject demo", "Batched (...)", "+ Add mock", the step slider and the state readout in a single `PANEL` object, with a comment pointing at the two dev-panel sources. This function re-types the same six strings.

A rename in the panel then fixes one spec and leaves the other failing on a stale string, with no compile error to connect the two. Lifting `PANEL` into a shared `e2e/` module and importing it in both would keep them together.

(The lane label at line 53 is a different case and is fine as-is — hardcoding the string with a comment citing its `copy.ts` key is the established convention in `e2e/`.)

services/vault/src/components/simple/DepositProgressView/SplitGroupedProgress.tsx
```diff
@@ -139,72 +138,71 @@ export function SplitGroupedProgress({
...
+              <GroupBlock
+                group={group}
+                number={number}
+                steps={steps}
+                currentStep={currentStep}
+                hasError={hasError}
+                activeStepDetail={renderStepDetail?.(rawStep)}
+                preSignDetail={preSignDetail}
+              />
+              {idx < trunkGroups.length - 1 && <StepConnector />}
```
gbarkhatov: **This connector branch is unreachable by construction.**

`trunkGroups` filters `buildStepGroups(...)` — a 1:1 map over the four static `STEP_GROUPS` — down to `endStep <= TRUNK_END_VISUAL_STEP`, and only `{registerDeposit, 1, 6}` qualifies (the next group starts at 7). So `trunkGroups.length` is always 0 or 1 and `idx < trunkGroups.length - 1` can never be true.

The `.map`, its `idx` parameter and this guard are dead, which the zero-dead-code policy forbids, and it misleads a reader into thinking the trunk can be multi-group.

Rendering the single trunk group directly would say what actually happens — keeping the `trunkGroups.length > 0` guard, which does matter now that the container carries `gap-6`.

services/vault/src/components/simple/DepositProgressView/BtcConfirmationDetail.tsx
```diff
@@ -36,19 +31,13 @@ export function BtcConfirmationDetail({
...
-  // Stacked: label on its own line above the value (narrow split columns).
-  // Inline: label left / value right (full-width single-column flow).
-  const rowClass = stacked
-    ? "flex flex-col gap-0.5"
-    : "flex items-center justify-between gap-2";
 
   return (
     <div className="mt-3 flex flex-col gap-2 rounded-lg bg-secondary-highlight p-3">
-      <div className={rowClass}>
+      <div className="flex flex-wrap items-center justify-between gap-2">
```
gbarkhatov: **`flex-wrap` with a symmetric `gap-2` collapses the label/value pairing on the rows that actually wrap.**

`gap-2` sets row-gap as well as column-gap, so a wrapped value sits 8px below its label — exactly the spacing the panel's own `flex flex-col gap-2` puts between two *different* rows. Two label/value pairs then read as four undifferentiated lines, which is the opposite of what wrapping is meant to preserve. The removed `stacked` variant used a tighter intra-row gap for this reason.

`gap-x-2 gap-y-0.5` keeps the horizontal spacing and tightens the wrap. Same pattern at line 53 here, and in `EthConfirmationDetail.tsx` and `ProviderWaitDetail.tsx`.

Flagging the CSS structure, not a measurement — I confirmed the classes and the long `bitcoinTx` label but did not verify in a browser that these rows wrap at 390px.

services/vault/src/components/simple/DepositProgressView/SplitGroupedProgress.tsx
```diff
@@ -86,13 +87,13 @@ function VaultColumn({
...
-    <div className="flex min-w-0 flex-1 flex-col">
+    <div className="flex flex-col">
       <Text
         as="span"
-        variant="body2"
+        variant="body1"
         className="mb-2 font-medium text-accent-primary"
       >
-        {COPY.deposit.progress.splitVaultColumnLabel(vaultIndex + 1)}
+        {COPY.deposit.progress.splitVaultLabel(vaultIndex + 1)}
```
gbarkhatov, line 101: **Minor, pre-existing, but the restructure makes it visible: a completed lane renders a full-width header with nothing under it.**

`VaultLane` always renders this label, but `branchGroups` is empty for a vault whose groups are all `"completed"` — which `useSplitVaultProgress` produces deliberately for a finished sibling, while the `status !== "completed"` filter below removes every one of them.

Resume a two-vault batch where vault 1 has activated and lane 1 is a bare "BTCVault 1" followed by 24px of gap. Side-by-side that was a narrow blank column; stacked it is a full-width dangling heading.

Not introduced here and not blocking — but if a lane has no groups to show, it probably should not render its header either.

gbarkhatov (dismissed):
Combo review — three independent passes, deduped and verified against the PR head.

**No merge-blockers.** The production change is correct: the `compact` and `stacked` removals are genuinely complete (no remaining reference anywhere in `services/` or `packages/`), no `data-testid` moved, and no critical path or `SECURITY_MODEL.md` invariant is touched. Most of what follows is in the new Playwright harness.

The one thing worth weighing before merge: that harness is this PR's only automated regression guard, and as committed no CI job runs it — see the comment on `playwright.config.ts`.

### Refuted

Raised by one of the three passes and killed on verification — recorded so they do not get re-raised:

- **"The comment on `AWAITING_PAYOUT_SCENARIO_INDEX` is wrong, because `AWAIT_PAYOUT_TRANSACTIONS = 8, not 7`."** The comment is accurate. `PENDING_FLOW_CONFIG` lists all fifteen steps in enum order and the slider is 0-based within its mode segment (`dev/panels/DepositVaults.tsx`, `offset + Number(e.target.value)`), so position 7 is the eighth entry — `AWAIT_PAYOUT_TRANSACTIONS`. The index is right and lands where intended.
- **"Import `COPY` into the spec rather than hardcoding the lane label."** The established `e2e/` convention is the opposite: hardcode the string and cite its `copy.ts` key in a comment, exactly as `e2e/real/actions/pegin.ts` does. The only `e2e/` to `src/` import in the tree is a single `import type`. The spec already follows the convention.
- **"The trunk-to-lane connector was silently dropped."** Removed deliberately and documented in the code. `StepConnector` is `h-6` and the new container gap is `gap-6`, so the vertical rhythm is preserved and only the drawn line goes.
- **"The txid row can spill past the card, so `flex-wrap` is insufficient."** `CopyableHash` renders `truncateHash(...)`, roughly 13 monospace characters, never the raw 64-char hex.
- **"Deleting `__tests__/StepRow.test.tsx` loses coverage."** Both tests asserted only on `className` substrings for the removed `compact` prop — implementation, not behaviour. `StepRow` stays covered behaviourally through `GroupBlock` in the two surviving suites.
- **"The red `verify` check is a defect of this PR."** Pre-existing breakage on `main`: `DepositForm.splitUnavailable.test.tsx:64` sets `hasUnconfirmedBalanceOnly`, which is not a field of `DepositAmountState`. That file is in no PR's changed-file list and the error reproduces on `main` at `8133b2b9e`.

### Verified clean

- Zero remaining references to `compact`, the `stacked` prop, `VaultColumn`, `columnStarted` or `splitVaultColumnLabel` anywhere in `services/` or `packages/`. The removal claim in the description is accurate.
- No `data-testid` moved, renamed or dropped. The `stepMachine.ts` and `pegin.ts` changes are comment-only, and the aria-label hooks the real-wallet machine drives are intact.
- `renderStepDetail`'s optional-`opts` signature change: three call sites, all updated; no behaviour moved.
- `copy.ts` is a key rename only — string values unchanged, so no user-visible text changed and no copy convention is affected.
- No motion tokens, magic protocol numbers, or registered critical paths touched. This is vault UI throughout.

One CI note that is not a finding: because `verify` aborts at that pre-existing `main` type error, it never reaches this PR's own tsc and unit run, so "verified with tsc, eslint, the full vault unit suite" is unproven by CI on this head. `visual` did pass.

gbarkhatov (dismissed):

services/vault/e2e/deposit-progress-layout.spec.ts
jonybur: The current view shows the shared trunk or the vault lanes. Step 6 no longer shows both, and it no longer renders the BTC hash detail. The layout spec now checks the real trunk at step 6 and the lanes at step 8, at all three widths.

services/vault/e2e/deposit-progress-layout.spec.ts
jonybur: The layout spec now owns a 360, 390, and 1280 pixel viewport list. Changes to the screenshot targets cannot reduce this coverage.

services/vault/playwright.config.ts
jonybur: The description now states that the six layout checks passed locally and that CI does not run this Playwright project. The normal CI checks are running on the new head.

services/vault/e2e/deposit-progress-layout.spec.ts
jonybur: The spec now checks the scenario readout immediately after each slider change. A reordered scenario list cannot silently test another step.

services/vault/e2e/deposit-progress-layout.spec.ts
jonybur: The spec now uses recordedPageWallets(). The wallet chain ID comes from the recording.

services/vault/e2e/deposit-progress-layout.spec.ts
jonybur: The existing accessible-name lookups stay local to each spec. This keeps the patch small and avoids a new shared module for the test strings. The tests fail directly when a panel label changes.

services/vault/src/components/simple/DepositProgressView/BtcConfirmationDetail.tsx
jonybur: The wrapped rows now use gap-x-2 and gap-y-0.5. Each value stays closer to its own label. The current demo does not render the BTC hash detail, so the browser check verifies the trunk and lane cards without claiming to measure that row.

gbarkhatov (approved):

PR #2409, opened 2026-09-07
fix(vault): drop a removed field from the deposit-form test

jonybur:
## What
Fixes the `verify` check, which fails on main and on every open pull request.

## Why
Two recent pull requests each passed their own checks, but together they left a test on main referring to a field that no longer exists. Nothing runs the checks on main itself after a merge, so the break only showed up on the next pull requests.

## How
Removes the stale field from the test. The test still passes.


gbarkhatov:
#2408 is merged, lets close this

PR #2436, opened 2026-09-08
chore(repo): add @kirugan as a code owner

jeremy-babylonlabs:
Adds `@kirugan` to the default owners and to every critical-path entry in `.github/CODEOWNERS`, so the owner list stays identical across all rules.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,5 @@
 # Default owners
-* @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov
+* @jonybur @jeremy-babylonlabs @jrwbabylonlab @gbarkhatov @kirugan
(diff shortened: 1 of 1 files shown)


jonybur (dismissed):
Reviewed. Uniformly adds @kirugan to every owner list — default owners plus all nine critical-path blocks. All lists are otherwise unchanged, ordering is preserved, and no path is silently retired or narrowed. Non-behavioural change to policy metadata; nothing else in the tree references CODEOWNERS. CI is green.

gbarkhatov (dismissed):

jonybur (approved):
Adds @kirugan to the default owner list and to every critical-path CODEOWNERS entry, keeping all lists identical. Nothing else changes and no path is narrowed or dropped. CI is green.

Issue #2438, opened 2026-09-08
[eth-first] Disconnect Ethereum without disconnecting Bitcoin


jeremy-babylonlabs:
## Outcome

Both disconnect directions are tracked in #2352.

Status: consolidated into #2352. This closure removes duplicate tracking. It does not mark the implementation delivered.

## Problem

The Ethereum and Bitcoin directions now share the same approved rule and the same open implementation PR.

## Scope

The approved Ethereum refusal rule, independent-Bitcoin preservation, unchanged state and events on refusal, connected-row guard, and provider/UI test matrix moved to #2352. jonybur selected this rule on 2026-09-09.

## Complete when

This ticket closes as consolidated after those requirements are recorded in #2352. Delivery is complete only when #2352 meets its checks and #2439 merges into main.

## Dependencies

#2480 merged into the #2439 feature branch, not main. Remaining implementation and review are tracked in #2352 and #2439. Parent: #2230. Epic: #2228.

Issue #2355, opened 2026-08-27
[eth-first] Verify the agreed wallet session flows


jonybur:
## Outcome

The connector and its two consumer apps pass the agreed wallet session checks.

Status: jonybur owns this task. #2483 merged into main at [05f60bf9](https://github.com/babylonlabs-io/babylon-toolkit/commit/05f60bf90129aa2415bc3ab1a8c95efbdc2220ec), and its two real OKX and Keplr cases now run in every CI verify job. The real Vault journey passed on the consent branch, which merged as `36d72d60`; it has not been repeated on main. Rows 4, 7, and 8 still need real-wallet evidence. This issue stays open for that evidence.

## Problem

Individual wallet fixes do not prove the agreed session flows in both products.

## Scope

The eight rows below are the complete acceptance list. Record the expected result, actual result, tested revision, and evidence link for each row. Reuse existing evidence for unchanged code. Repeat affected checks after integration.

Cover the listed account, wallet, key, network, and required-chain changes, plus confirmation, cancel, close, reopen, reload, restore, and per-chain disconnect. Do not require every provider, route, and event combination. Reference #2352's provider matrix here. Add a row only for a demonstrated gap in the original acceptance criteria or a separately approved requirement.

Baseline: `42d9f55`, merged through #2445. Its wallet source and scripts match main `8f62bcfa`. All 295 wallet tests, 23 package checks, and both clean package consumers passed. [Passing CI](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34242967596). These are earlier runs, not new results.

| Scenario | Expected result | Actual result and evidence |
|---|---|---|
| Required and optional wallets; entry modes | Required wallets cannot be omitted. Ethereum can confirm without optional Bitcoin. Ethereum-only, Bitcoin-only, and combined AppKit setup is consistent; incompatible setup fails. | Main passes: 13 optional-chain tests, 6 required-chain tests, and 22 AppKit setup tests. #2578 adds real-wallet proof: the connect screen marks Bitcoin optional, and the session confirms with Ethereum alone. |
| Account, wallet, key, network, and requirement changes | Required identity changes revoke approval. Optional changes preserve valid approval. | Main passes: 31 requirement-change tests and 14 confirmation-receipt tests. |
[...]

Issue #2354, opened 2026-08-27
[eth-first] Apply the selected consent rule


jonybur:
## Outcome

Vault and simple staking use the selected consent and saved-session rule.

Status: #2491 merged into main at `36d72d60` on 2026-09-15, with adoption in both products. Its decision record and release note were dropped under review and are not in the repository. The selected rule below is the record. This issue stays open for that decision.

On 2026-09-09, jonybur selected this rule and removed the requirement to wait for named Product and Legal approval. jonybur will contact Legal later. Legal approval is not claimed.

Selected rule: final Connect records consent for the live first required wallet identity and the current required wallet set. Valid saved consent can restore after reload. Changed or invalid identities require confirmation again. Connecting one wallet or closing the dialog does not record consent.

## Problem

#2267 changed consent timing without a recorded policy decision. Both products must apply the selected rule and test it.

## Scope

- #2491 applies the confirmed-session gate in Vault and verifies the existing rule in both products. Required wallets and session reset behavior stay unchanged.
- The selected rule above is the record. The merged change adds no decision document and no release note. Neither app adds a callback or service endpoint.
- Local checks passed: 3,788 Vault tests, 7 staking consumer tests, 75 shared consent tests, and 23 package checks. Six existing Vault tests were skipped. Scoped lint and the Vault production build passed.
- Real OKX and MetaMask connection, close, reopen, and reload passed on the change published as #2491. Both identities and saved consent survived reload. The wider matrix remains in #2355.

## Complete when

- [x] jonybur selected the rule and authorized implementation before the later Legal follow-up.
- [ ] The recorded rule and support guidance merge with adoption in both products.
- [x] Required reviews and CI pass, and #2491 merges.
- [x] Tests prove consent timing, live identity, and saved-session behavior for both products.

Use the selected rule above for any needed communication in #2233. A separate communication task does not block verified wallet adoption or #2232.

## Dependencies

Parent: #2230. Epic: #2228. Implementation is authorized. Adoption and its required checks must pass before completion of #2230 and dependent behavior changes in #2232. Final session evidence remains in #2355.

PR #2439, opened 2026-09-08
fix(wallet-connector): preserve the other wallet on partial disconnect

### LINKED ISSUE #2352: [eth-first] Disconnect one wallet without disconnecting the other
## Outcome

Disconnecting one wallet keeps the other wallet connected. A request that would end a shared session is refused. Disconnect All still ends both sessions.

Status: both directions are implemented in open #2439 at `c16ea2186`. #2480 merged into that feature branch, not main. Reviewer clearance and the main merge remain. The Bitcoin rule was approved on 2026-09-07. jonybur selected the matching Ethereum rule on 2026-09-09. #2438 is consolidated here.

## Problem

An AppKit disconnect can end both chains. Provider errors can also clear local state or report success. The connected wallet row must not bypass the connector through Reown's own Disconnect control.

## Scope

- Use an isolated namespace disconnect where AppKit supports it.
- For either direction, refuse a chain-only disconnect when WalletConnect or Auth would also disconnect the other chain. Include mixed sessions with an independent injected wallet.
- On refusal, keep both sessions and local state unchanged. Emit no disconnect event. Use the existing refusal code and the affected chain's message.
- Apply the same rule to both connected wallet rows. Do not expose an unguarded Reown disconnect for a protected shared session. Keep existing account controls for independent Ethereum sessions.
- Preserve explicit Disconnect All, including when Bitcoin is first, and local cleanup. A failed disconnect must not report success or clear the connected state.
- Keep rejected wallets out after reopen. An old disconnect must not clear a later connection. The reconnect guard covers local state and events, not remote teardown.
- Reuse provider and UI tests for injected, WalletConnect, Auth, mixed, refused, failed, and reconnect cases. Include canonical and manual AppKit setup.

## Complete when

- [ ] Bitcoin disconnect preserves independent Ethereum, and Ethereum disconnect preserves independent Bitcoin.
- [ ] Both shared-session directions follow the approved refusal rule. Refusal changes no session or local state and emits no disconnect event.
- [ ] Both connected wallet rows use the guarded path. A protected shared session cannot reach Reown's unguarded Disconnect control.
- [ ] Provider and UI tests cover injected, WalletConnect, Auth, mixed sessions, refusal, failed disconnects, rejected-wallet restore, and reconnect races.
- [ ] Disconnect All still disconnects all wallets and completes local cleanup, including when Bitcoin is first.
- [ ] The UI reports what disconnected or explains the refusal. #2439 has passing checks and merges into main.

## Dependencies

Parent: #2230. Epic: #2228. The rules are approved; no new product decision is needed. This ticket includes the requirements moved from #2438 and the implementation in #2480.

Vault state after a successful Bitcoin disconnect belongs to #2232. Final consumer session evidence belongs to #2355. This ticket remains open until the combined connector change is delivered.

jeremy-babylonlabs:
## Outcome

A Bitcoin-only or Ethereum-only disconnect preserves the other wallet. A shared WalletConnect request is refused when it would end both sessions. The dialog explains the refusal. Explicit Disconnect All still ends both sessions.

The review fixes are pushed in `cbf3c6715d8da08a2fce22b5e1e0d059669d494a`. All 12 CI checks pass. Greptile gives this commit 5/5. Reviewer clearance remains.

## Change

- Add chain, all, and local disconnect scopes. Preserve connected state on a refused or failed chain disconnect.
- Use local cleanup when Ethereum reports lost accounts or a failed connection check. Keep the public Ethereum disconnect on the guarded path.
- Allow an independent Ethereum Auth session to disconnect. Keep the shared WalletConnect guard and the defensive Bitcoin Auth guard.
- Clear connector state before its disconnect event. An old disconnect cannot clear a later connection to the same wallet.
- Complete local cleanup for Disconnect All. Prevent a rejected wallet from returning when the dialog reopens.

## Safety

This includes the reverse disconnect change from #2480, under the decision in #2352. Tests cover independent wallets, shared sessions, refusals, failures, lost accounts, reconnect races, and Disconnect All.

The Ethereum-only entry keeps its dependency checks. The reconnect guard protects local connector state and events. It does not control remote wallet teardown. Full consumer session evidence remains in #2355.

## Verification

Checks for the code pushed in `cbf3c6715d8da08a2fce22b5e1e0d059669d494a`:

| Check | Result |
|---|---|
| Wallet unit tests | 358 pass |
| Package boundary tests | 23 pass |
| Nx build and four dependency tasks | Pass |
| Full package lint | No errors; 89 existing warnings |
| Commit authentication | Pass in CI |
| [Fresh Verify CI](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34472259031) | Pass |
| [Visual CI](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34472259138) | Pass; all 387 captures match; no missing or incomplete captures |
| Greptile review of this commit | 5/5 |

Three regression checks failed before the review fixes and pass now. The [current Greptile review](https://github.com/babylonlabs-io/babylon-toolkit/pull/2439#issuecomment-5582490820) has no new findings. All 15 inline review threads are resolved. The changes-requested review still needs reviewer clearance.

## Links

Closes #2352. #2438 is consolidated into #2352. Parent: #2230 and epic #2228.

- #2480: reverse disconnect implementation, included here.
- #2499: depends on this PR.
- #2355: full consumer session evidence.
- #2354: consent adoption.
- #2232: Vault session behavior.

### DIFF
--- a/packages/babylon-wallet-connector/src/components/Chains/container.tsx
+++ b/packages/babylon-wallet-connector/src/components/Chains/container.tsx
@@ -6,6 +6,7 @@ import type { IChain } from "@/core/types";
 // adapter into the `./eth` graph, since this screen is part of the dialog.
 import { useChainProviders } from "@/context/Chain.context";
 import { APPKIT_BTC_CONNECTOR_ID, APPKIT_ETH_CONNECTOR_ID, APPKIT_OPEN_EVENT } from "@/core/wallets/appkit/constants";
+import { ethDisconnectWouldDropBitcoin } from "@/core/wallets/eth/appkit/sharedConfig";
 import { isSharedSessionRefusal } from "@/error";
 import { useWalletConnect } from "@/hooks/useWalletConnect";
 import { useWidgetState } from "@/hooks/useWidgetState";
@@ -36,59 +37,27 @@ export function ChainsContainer(props: ContainerProps) {
 
   const handleSelectChain = useCallback(
     async (chain: IChain) => {
-      // Special handling for ETH chain with only AppKit wallet
-      if (chain.id === "ETH") {
-        const ethConnector = connectors.ETH;
-        const appkitWallet = ethConnector?.wallets.find((w) => w.id === APPKIT_ETH_CONNECTOR_ID);
-
-        if (appkitWallet && ethConnector?.wallets.length === 1) {
-          // Already connected: reopen the AppKit modal so the user can switch/disconnect.
-          if (ethConnector.connectedWallet) {
-            openAppKitModal();
-            return;
-          }
-
-          // Only AppKit available, connect it directly
-          // This will trigger the AppKitProvider.connectWallet() which dispatches the event
+      if (chain.id === "ETH" || chain.id === "BTC") {
+        const connector = connectors[chain.id];
+        const appkitId = chain.id === "ETH" ? APPKIT_ETH_CONNECTOR_ID : APPKIT_BTC_CONNECTOR_ID;
+        const wallet = connector?.wallets.find((candidate) => candidate.id === appkitId);
+        if (wallet && connector?.wallets.length === 1) {
+          const connected = Boolean(connector.connectedWallet);
           try {
-            await ethConnector.connect(appkitWallet);
-          } catch (error) {
-            console.error("Failed to connect AppKit:", error);
-          }
-          return;
-        }
-      }
-
-      // Special handling for BTC chain with only AppKit wallet
-      if (chain.id === "BTC") {
-        const btcConnector = connectors.BTC;
-        const appkitBtcWallet = btcConnector?.wallets.find((w) => w.id === APPKIT_BTC_CONNECTOR_ID);
-
-        if (appkitBtcWallet && btcConnector?.wallets.length === 1) {
-          // Already connected: disconnect through the connector rather than
-          // reopening Reown, whose own Disconnect control drops a shared
-          // session for every chain. A shared session is refused and the
-          // dialog explains why; the row can be clicked again to connect.
-          if (btcConnector.connectedWallet) {
-            try {
-              await disconnect("BTC");
-            } catch (error) {
-              if (!isSharedSessionRefusal(error)) {
-                console.error(
-                  "Failed to disconnect AppKit BTC:",
-                  error instanceof Error ? error.message : "Unknown error",
-                );
-              }
+            if (!connected) {
+              await connector.connect(wallet.id);
+            } else if (chain.id === "ETH" && !ethDisconnectWouldDropBitcoin()) {
+              openAppKitModal();
+            } else {
+              await disconnect(chain.id);
             }
-            return;
-          }
-
-          // Only AppKit available, connect it directly
-          // This will trigger the AppKitBTCProvider.connectWallet() which dispatches the event
-          try {
-            await btcConnector.connect(appkitBtcWallet);
           } catch (error) {
-            console.error("Failed to connect AppKit BTC:", error);
+            if (!isSharedSessionRefusal(error)) {
+              console.error(
+                `Failed to ${connected ? "disconnect" : "connect"} AppKit ${chain.id}:`,
+                error instanceof Error ? error.message : "Unknown error",
+              );
+            }
           }
           return;
         }
--- a/packages/babylon-wallet-connector/src/components/Chains/container.tsx
+++ b/packages/babylon-wallet-connector/src/components/Chains/container.tsx
@@ -4,14 +4,11 @@ import type { IChain } from "@/core/types";
 // Connector ids come from the shared constants module, not from each chain's
 // wallet metadata — importing the metadata here would put every Bitcoin wallet
 // adapter into the `./eth` graph, since this screen is part of the dialog.
-import {
-  APPKIT_BTC_CONNECTOR_ID,
-  APPKIT_ETH_CONNECTOR_ID,
-  APPKIT_OPEN_EVENT,
-} from "@/core/wallets/appkit/constants";
+import { useChainProviders } from "@/context/Chain.context";
+import { APPKIT_BTC_CONNECTOR_ID, APPKIT_ETH_CONNECTOR_ID, APPKIT_OPEN_EVENT } from "@/core/wallets/appkit/constants";
+import { isSharedSessionRefusal } from "@/error";
 import { useWalletConnect } from "@/hooks/useWalletConnect";
 import { useWidgetState } from "@/hooks/useWidgetState";
-import { useChainProviders } from "@/context/Chain.context";
 
 import { Chains } from "./index";
 
@@ -32,7 +29,7 @@ interface ContainerProps {
 
 export function ChainsContainer(props: ContainerProps) {
   const { chains, requiredChainIds, selectedWallets, displayWallets } = useWidgetState();
-  const { selected } = useWalletConnect();
+  const { selected, disconnect } = useWalletConnect();
   const connectors = useChainProviders();
 
   const chainArr = useMemo(() => Object.values(chains), [chains]);
@@ -42,7 +39,7 @@ export function ChainsContainer(props: ContainerProps) {
       // Special handling for ETH chain with only AppKit wallet
       if (chain.id === "ETH") {
         const ethConnector = connectors.ETH;
-        const appkitWallet = ethConnector?.wallets.find(w => w.id === APPKIT_ETH_CONNECTOR_ID);
+        const appkitWallet = ethConnector?.wallets.find((w) => w.id === APPKIT_ETH_CONNECTOR_ID);
 
         if (appkitWallet && ethConnector?.wallets.length === 1) {
           // Already connected: reopen the AppKit modal so the user can switch/disconnect.
@@ -65,18 +62,24 @@ export function ChainsContainer(props: ContainerProps) {
       // Special handling for BTC chain with only AppKit wallet
       if (chain.id === "BTC") {
         const btcConnector = connectors.BTC;
-        const appkitBtcWallet = btcConnector?.wallets.find(w => w.id === APPKIT_BTC_CONNECTOR_ID);
+        const appkitBtcWallet = btcConnector?.wallets.find((w) => w.id === APPKIT_BTC_CONNECTOR_ID);
 
         if (appkitBtcWallet && btcConnector?.wallets.length === 1) {
+          // Already connected: disconnect through the connector rather than
+          // reopening Reown, whose own Disconnect control drops a shared
+          // session for every chain. A shared session is refused and the
+          // dialog explains why; the row can be clicked again to connect.
           if (btcConnector.connectedWallet) {
-            const { btcDisconnectWouldDropEthereum } = await import("@/core/wallets/btc/appkit/sharedConfig");
-            if (btcDisconnectWouldDropEthereum()) {
-              await btcConnector.disconnect().catch(() => undefined);
-              return;
+            try {
+              await disconnect("BTC");
+            } catch (error) {
+              if (!isSharedSessionRefusal(error)) {
+                console.error(
+                  "Failed to disconnect AppKit BTC:",
+                  error instanceof Error ? error.message : "Unknown error",
+                );
+              }
             }
-
-            // Already connected: reopen the AppKit modal so the user can switch/disconnect.
-            openAppKitModal();
             return;
           }
 
@@ -94,7 +97,7 @@ export function ChainsContainer(props: ContainerProps) {
       // Normal flow for other chains or if chain has multiple wallets
       displayWallets?.(chain.id);
     },
-    [displayWallets, connectors],
+    [displayWallets, connectors, disconnect],
   );
 
   return (
--- a/packages/babylon-wallet-connector/src/components/Chains/container.tsx
+++ b/packages/babylon-wallet-connector/src/components/Chains/container.tsx
@@ -68,8 +68,14 @@ export function ChainsContainer(props: ContainerProps) {
         const appkitBtcWallet = btcConnector?.wallets.find(w => w.id === APPKIT_BTC_CONNECTOR_ID);
 
         if (appkitBtcWallet && btcConnector?.wallets.length === 1) {
-          // Already connected: reopen the AppKit modal so the user can switch/disconnect.
           if (btcConnector.connectedWallet) {
+            const { btcDisconnectWouldDropEthereum } = await import("@/core/wallets/btc/appkit/sharedConfig");
+            if (btcDisconnectWouldDropEthereum()) {
+              await btcConnector.disconnect().catch(() => undefined);
+              return;
+            }
+
+            // Already connected: reopen the AppKit modal so the user can switch/disconnect.
             openAppKitModal();
             return;
           }
--- a/packages/babylon-wallet-connector/src/core/WalletConnector.ts
+++ b/packages/babylon-wallet-connector/src/core/WalletConnector.ts
@@ -2,7 +2,7 @@ import { createNanoEvents } from "nanoevents";
 
 import { Wallet } from "@/core/Wallet";
 import type { DisconnectScope, IConnector, IProvider } from "@/core/types";
-import { ERROR_CODES, WalletError } from "@/error";
+import { ERROR_CODES, WalletError, isSharedSessionRefusal } from "@/error";
 
 type DisconnectableProvider = IProvider & { disconnect?: (scope: DisconnectScope) => Promise<void> };
 
@@ -55,19 +55,22 @@ export class WalletConnector<N extends string, P extends IProvider, C> implement
   }
 
   async disconnect(scope: DisconnectScope = "chain") {
-    if (!this._connectedWallet) return;
+    const wallet = this._connectedWallet;
+    if (!wallet) return;
 
-    const provider = this._connectedWallet.provider as DisconnectableProvider | null;
+    const provider = wallet.provider as DisconnectableProvider | null;
     if (provider?.disconnect) {
       try {
         await provider.disconnect(scope);
       } catch (error) {
+        if (scope === "chain" && isSharedSessionRefusal(error)) throw error;
         this._ee.emit("error", error instanceof Error ? error : new Error(String(error)));
-        if (scope === "chain" && this._connectedWallet?.account !== null) throw error;
+        if (scope === "chain") throw error;
       }
     }
-    this._ee.emit("disconnect", this._connectedWallet);
+    if (this._connectedWallet !== wallet) return;
     this._connectedWallet = null;
+    this._ee.emit("disconnect", wallet);
   }
 
   clone() {
--- a/packages/babylon-wallet-connector/src/core/WalletConnector.ts
+++ b/packages/babylon-wallet-connector/src/core/WalletConnector.ts
@@ -15,6 +15,7 @@ export interface ConnectorEvents<P extends IProvider> {
 
 export class WalletConnector<N extends string, P extends IProvider, C> implements IConnector<N, P, C> {
   private _connectedWallet: Wallet<P> | null = null;
+  private _connectionGeneration = 0;
   private _ee = createNanoEvents<ConnectorEvents<P>>();
 
   constructor(
@@ -45,6 +46,7 @@ export class WalletConnector<N extends string, P extends IProvider, C> implement
         this._ee.emit("connecting", message, description);
       await selectedWallet.connect(reportProgress);
       this._connectedWallet = selectedWallet;
+      this._connectionGeneration += 1;
       this._ee.emit("connect", this._connectedWallet);
 
       return this.connectedWallet;
@@ -56,6 +58,7 @@ export class WalletConnector<N extends string, P extends IProvider, C> implement
 
   async disconnect(scope: DisconnectScope = "chain") {
     const wallet = this._connectedWallet;
+    const generation = this._connectionGeneration;
     if (!wallet) return;
 
     const provider = wallet.provider as DisconnectableProvider | null;
@@ -68,7 +71,7 @@ export class WalletConnector<N extends string, P extends IProvider, C> implement
         if (scope === "chain") throw error;
       }
     }
-    if (this._connectedWallet !== wallet) return;
+    if (this._connectedWallet !== wallet || this._connectionGeneration !== generation) return;
     this._connectedWallet = null;
     this._ee.emit("disconnect", wallet);
   }
--- a/packages/babylon-wallet-connector/src/core/WalletConnector.ts
+++ b/packages/babylon-wallet-connector/src/core/WalletConnector.ts
@@ -63,7 +63,7 @@ export class WalletConnector<N extends string, P extends IProvider, C> implement
         await provider.disconnect(scope);
       } catch (error) {
         this._ee.emit("error", error instanceof Error ? error : new Error(String(error)));
-        if (scope === "chain") throw error;
+        if (scope === "chain" && this._connectedWallet?.account !== null) throw error;
       }
     }
     this._ee.emit("disconnect", this._connectedWallet);
--- a/packages/babylon-wallet-connector/src/core/WalletConnector.ts
+++ b/packages/babylon-wallet-connector/src/core/WalletConnector.ts
@@ -1,10 +1,10 @@
 import { createNanoEvents } from "nanoevents";
 
 import { Wallet } from "@/core/Wallet";
-import type { IConnector, IProvider } from "@/core/types";
+import type { DisconnectScope, IConnector, IProvider } from "@/core/types";
 import { ERROR_CODES, WalletError } from "@/error";
 
-type DisconnectableProvider = IProvider & { disconnect?: () => Promise<void> };
+type DisconnectableProvider = IProvider & { disconnect?: (scope: DisconnectScope) => Promise<void> };
 
 export interface ConnectorEvents<P extends IProvider> {
   connecting: (message?: string, description?: string) => void;
@@ -54,19 +54,20 @@ export class WalletConnector<N extends string, P extends IProvider, C> implement
     }
   }
 
-  async disconnect() {
-    if (this._connectedWallet) {
-      const provider = this._connectedWallet.provider as DisconnectableProvider | null;
-      if (provider?.disconnect && typeof provider.disconnect === "function") {
-        try {
-          await provider.disconnect();
-        } catch {
-          // ignore provider disconnect errors
-        }
+  async disconnect(scope: DisconnectScope = "chain") {
+    if (!this._connectedWallet) return;
+
+    const provider = this._connectedWallet.provider as DisconnectableProvider | null;
+    if (provider?.disconnect) {
+      try {
+        await provider.disconnect(scope);
+      } catch (error) {
+        this._ee.emit("error", error instanceof Error ? error : new Error(String(error)));
+        if (scope === "chain") throw error;
       }
-      this._ee.emit("disconnect", this._connectedWallet);
-      this._connectedWallet = null;
     }
+    this._ee.emit("disconnect", this._connectedWallet);
+    this._connectedWallet = null;
   }
 
   clone() {
--- a/packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
+++ b/packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
@@ -1,0 +1,66 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { __resetSharedBtcAppKitConfigForTests, btcDisconnectWouldDropEthereum, setSharedBtcAppKitConfig } from "../sharedConfig";
+
+afterEach(() => {
+  __resetSharedBtcAppKitConfigForTests();
+});
+
+function setModal(getProviderType: () => string, getAccount: () => { isConnected: boolean } | undefined) {
+  setSharedBtcAppKitConfig({
+    modal: { getProviderType: vi.fn(getProviderType), getAccount: vi.fn(getAccount) } as never,
+    adapter: {} as never,
+    network: "signet",
+  });
+}
+
+describe("btcDisconnectWouldDropEthereum", () => {
+  it("reports true when bitcoin runs over walletconnect and ethereum is connected", () => {
+    setModal(
+      () => "WALLET_CONNECT",
+      () => ({ isConnected: true }),
+    );
+
+    expect(btcDisconnectWouldDropEthereum()).toBe(true);
+  });
+
+  it("reports true when bitcoin runs over auth and ethereum is connected", () => {
+    setModal(
+      () => "AUTH",
+      () => ({ isConnected: true }),
+    );
+
+    expect(btcDisconnectWouldDropEthereum()).toBe(true);
+  });
+
+  it("reports false for an injected bitcoin wallet even while ethereum is connected", () => {
+    setModal(
+      () => "INJECTED",
+      () => ({ isConnected: true }),
+    );
+
+    expect(btcDisconnectWouldDropEthereum()).toBe(false);
+  });
+
+  it("reports false over walletconnect when there is no ethereum account", () => {
+    setModal(
+      () => "WALLET_CONNECT",
+      () => undefined,
+    );
+
+    expect(btcDisconnectWouldDropEthereum()).toBe(false);
+  });
+
+  it("reports false over walletconnect when the ethereum account is disconnected", () => {
+    setModal(
+      () => "WALLET_CONNECT",
+      () => ({ isConnected: false }),
+    );
+
+    expect(btcDisconnectWouldDropEthereum()).toBe(false);
+  });
+
+  it("reports false when no shared bitcoin config is registered", () => {
+    expect(btcDisconnectWouldDropEthereum()).toBe(false);
+  });
+});
--- a/packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
+++ b/packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
@@ -1,5 +1,7 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 
+import { ethDisconnectWouldDropBitcoin } from "@/core/wallets/eth/appkit/sharedConfig";
+
 import {
   __resetSharedBtcAppKitConfigForTests,
   btcDisconnectWouldDropEthereum,
@@ -26,55 +28,58 @@ function setModal(session: {
   return modal;
 }
 
-describe("btcDisconnectWouldDropEthereum", () => {
-  it("reports true when bitcoin runs over walletconnect and ethereum is connected", () => {
+describe.each([
+  ["Bitcoin", "bip122", "eip155", btcDisconnectWouldDropEthereum],
+  ["Ethereum", "eip155", "bip122", ethDisconnectWouldDropBitcoin],
+] as const)("%s shared disconnect", (_chain, namespace, otherNamespace, wouldDropOther) => {
+  it("refuses WalletConnect while the other chain is connected", () => {
     const modal = setModal({
-      providerTypes: { bip122: "WALLET_CONNECT", eip155: "ANNOUNCED" },
-      accounts: { bip122: { isConnected: false }, eip155: { isConnected: true } },
+      providerTypes: { [namespace]: "WALLET_CONNECT", [otherNamespace]: "ANNOUNCED" },
+      accounts: { [namespace]: { isConnected: false }, [otherNamespace]: { isConnected: true } },
     });
 
-    expect(btcDisconnectWouldDropEthereum()).toBe(true);
-    expect(modal.getProviderType).toHaveBeenCalledWith("bip122");
-    expect(modal.getAccount).toHaveBeenCalledWith("eip155");
+    expect(wouldDropOther()).toBe(true);
+    expect(modal.getProviderType).toHaveBeenCalledWith(namespace);
+    expect(modal.getAccount).toHaveBeenCalledWith(otherNamespace);
   });
 
-  it("reports true when bitcoin runs over auth and ethereum is connected", () => {
+  it("refuses Auth while the other chain is connected", () => {
     setModal({
-      providerTypes: { bip122: "AUTH", eip155: "ANNOUNCED" },
-      accounts: { bip122: { isConnected: false }, eip155: { isConnected: true } },
+      providerTypes: { [namespace]: "AUTH", [otherNamespace]: "ANNOUNCED" },
+      accounts: { [namespace]: { isConnected: false }, [otherNamespace]: { isConnected: true } },
     });
 
-    expect(btcDisconnectWouldDropEthereum()).toBe(true);
+    expect(wouldDropOther()).toBe(true);
   });
 
-  it("reports false for an announced bitcoin extension even while ethereum is connected over walletconnect", () => {
+  it("allows an announced extension while the other chain uses WalletConnect", () => {
     setModal({
-      providerTypes: { bip122: "ANNOUNCED", eip155: "WALLET_CONNECT" },
+      providerTypes: { [namespace]: "ANNOUNCED", [otherNamespace]: "WALLET_CONNECT" },
       accounts: { bip122: { isConnected: true }, eip155: { isConnected: true } },
     });
 
-    expect(btcDisconnectWouldDropEthereum()).toBe(false);
+    expect(wouldDropOther()).toBe(false);
   });
 
-  it("reports false over walletconnect when there is no ethereum account", () => {
+  it("allows WalletConnect when the other account is absent", () => {
     setModal({
-      providerTypes: { bip122: "WALLET_CONNECT" },
-      accounts: { bip122: { isConnected: true } },
+      providerTypes: { [namespace]: "WALLET_CONNECT" },
+      accounts: { [namespace]: { isConnected: true } },
     });
 
-    expect(btcDisconnectWouldDropEthereum()).toBe(false);
+    expect(wouldDropOther()).toBe(false);
   });
 
-  it("reports false over walletconnect when the ethereum account is disconnected", () => {
+  it("allows WalletConnect when the other account is disconnected", () => {
     setModal({
       providerTypes: { bip122: "WALLET_CONNECT", eip155: "WALLET_CONNECT" },
-      accounts: { bip122: { isConnected: true }, eip155: { isConnected: false } },
+      accounts: { [namespace]: { isConnected: true }, [otherNamespace]: { isConnected: false } },
     });
 
-    expect(btcDisconnectWouldDropEthereum()).toBe(false);
+    expect(wouldDropOther()).toBe(false);
   });
 
-  it("reports false when no shared bitcoin config is registered", () => {
-    expect(btcDisconnectWouldDropEthereum()).toBe(false);
+  it("reports no shared session before initialization", () => {
+    expect(wouldDropOther()).toBe(false);
   });
 });
--- a/packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
+++ b/packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
@@ -1,61 +1,75 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 
-import { __resetSharedBtcAppKitConfigForTests, btcDisconnectWouldDropEthereum, setSharedBtcAppKitConfig } from "../sharedConfig";
+import {
+  __resetSharedBtcAppKitConfigForTests,
+  btcDisconnectWouldDropEthereum,
+  setSharedBtcAppKitConfig,
+} from "../sharedConfig";
 
 afterEach(() => {
   __resetSharedBtcAppKitConfigForTests();
 });
 
-function setModal(getProviderType: () => string, getAccount: () => { isConnected: boolean } | undefined) {
-  setSharedBtcAppKitConfig({
-    modal: { getProviderType: vi.fn(getProviderType), getAccount: vi.fn(getAccount) } as never,
-    adapter: {} as never,
-    network: "signet",
-  });
+type Namespace = "bip122" | "eip155";
+
+// Both AppKit accessors are per-namespace lookups, so every fixture gives the
+// two namespaces different values: a predicate that read the wrong one fails.
+function setModal(session: {
+  providerTypes: Partial<Record<Namespace, string>>;
+  accounts: Partial<Record<Namespace, { isConnected: boolean }>>;
+}) {
+  const modal = {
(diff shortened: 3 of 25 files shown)


packages/babylon-wallet-connector/src/hooks/useWalletConnectors.tsx
```diff
@@ -86,6 +86,17 @@ export function useWalletConnectors({ persistent, accountStorage, onError }: Pro
     (chainId) => connectors[chainId as ChainId]?.connectedWallet,
   );
 
+  const dropRejectedWallet = (connector: Pick<IConnector, "id" | "disconnect">) => {
```
jonybur: [P2] Keep rejected wallets out of the reopen path

This cleanup removes the wallet from widget state and storage, but a refused disconnect keeps connector.connectedWallet. If Bitcoin fails host verification, or the user cancels the key-mismatch warning, close and reopen the dialog. The effect at lines 215-235 selects that retained wallet without running either validation again. Receipt checks return when no confirmation receipt exists.

Clear the rejected wallet from local connector state, or prevent restore from selecting it. Keep the remote shared session unchanged. Add a test with a non-null connected wallet that changes visible after the refusal. The current test fixes connectedWallet at null and does not cover reopen.

jonybur (changes requested):
Review of 5ae352c128f21177e51bff9b752a4c716ee856f0 found one blocker.

A rejected Bitcoin wallet can return when the dialog opens again after shared-session cleanup is refused. The inline comment gives the path and required regression test. This finding comes from code tracing; it is not a browser reproduction.

The other #2352 behaviors have provider, connector, and UI tests. [Current verification passes](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34221768124/job/102046314358).

Before merge: fix the rejected-wallet restore path, test close and reopen with a retained connected wallet, pass verification on the new revision, and resolve the remaining review threads. Full consumer session proof stays in #2355.

packages/babylon-wallet-connector/src/core/WalletConnector.ts
```diff
@@ -54,19 +54,20 @@ export class WalletConnector<N extends string, P extends IProvider, C> implement
...
+  async disconnect(scope: DisconnectScope = "chain") {
+    if (!this._connectedWallet) return;
+
+    const provider = this._connectedWallet.provider as DisconnectableProvider | null;
+    if (provider?.disconnect) {
+      try {
+        await provider.disconnect(scope);
+      } catch (error) {
+        this._ee.emit("error", error instanceof Error ? error : new Error(String(error)));
+        if (scope === "chain" && this._connectedWallet?.account !== null) throw error;
```
gbarkhatov: **Merge-blocker — this preserves the wallet on every chain-scope failure, not just the refusal.**

The branch keys on `scope` and on `account`, never on the error, so it cannot tell the designed refusal apart from a genuine failure. AppKit funnels all of its internal failures through `catch (error) { throw new Error(`Failed to disconnect chains: ${error.message}`) }` at `appkit-base-client.js:629`, and `getAppKitConfig()` throws a plain `Error` too. Both land here as "keep the wallet, stay connected" — so a relay timeout or a wallet RPC failure leaves the connector claiming a session that is gone, with nothing to re-fire and heal it. #2352 asks that a *refused* request preserve state; it does not ask that of a *failed* one, and preserving there is what produces the Ethereum wedge on the other side.

The discriminator is already in hand and both symbols are already imported in this file:

```ts
if (error instanceof WalletError && error.code === ERROR_CODES.SHARED_SESSION_DISCONNECT_REFUSED) throw error;
```

That also lets the `account !== null` condition go away entirely (see the note on `dropRejectedWallet`).

Smaller thing on the same path: the error is both `emit("error")`'d and rethrown, so a caller that awaits and a subscriber both handle it. Probably worth picking one.

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/sharedConfig.ts
```diff
@@ -104,6 +104,17 @@ export function hasSharedBtcAppKitConfig(): boolean {
   return initializedState ? initializedState.btcConfig !== undefined : sharedBtcAppKitConfig !== null;
 }
 
+export function btcDisconnectWouldDropEthereum(): boolean {
+  if (!hasSharedBtcAppKitConfig()) {
+    return false;
+  }
+
+  const { modal } = getSharedBtcAppKitConfig();
+  const btcProviderType = modal.getProviderType("bip122");
```
gbarkhatov, line 119: **Merge-blocker — this reads a different AppKit state slice than the one AppKit branches on.**

`modal.getProviderType(ns)` is `ProviderController.getProviderId(ns)` (`appkit-base-client.js:54`), backed by `state.providerIds`. AppKit's widening decision uses `ConnectorController.getConnectorId(namespace)` (`appkit-base-client.js:579`), backed by `state.activeConnectorIds`. They agree in steady state, but they are filled on different clocks:

- `ConnectorController.initialize()` walks the namespaces and calls `setConnectorId(StorageUtil.getConnectedConnectorId(ns))` — **synchronous, from storage, at AppKit init**.
- `providerIds.bip122` is only written by `ProviderController.setProviderId(chainNamespace, CONNECTOR_TYPE_WALLET_CONNECT)` inside `syncWalletConnectAccount` (`appkit-base-client.js:1235`), which is awaited and runs only once `sessionAddress` resolves.

So on any load with a persisted WalletConnect Bitcoin session there is a window where `activeConnectorIds.bip122 === 'walletConnect'` — AppKit *will* widen — while `providerIds.bip122` is still `undefined`, this returns `false`, and a Bitcoin disconnect takes Ethereum down with it. That is the exact failure the PR exists to prevent, and it is most reachable in the mixed case this PR cares about: an injected Ethereum wallet rehydrates fast while the Bitcoin WalletConnect session is still coming up over the relay.

AppKit exposes no public getter for `activeConnectorIds`, so the cheap fix is to treat unresolved as unsafe rather than safe — refuse when `getProviderType("bip122")` is `undefined` while `getAccount("bip122")?.isConnected` is true — with a comment citing `appkit-base-client.js:591` so the divergence is not silently "cleaned up" later.

While here: `"WALLET_CONNECT"` and `"AUTH"` mirror AppKit's `CONNECTOR_TYPE_*` constants with no compile-time tie, so an upstream rename degrades this to a permanent `false` with no test failure.

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/provider.ts
```diff
@@ -248,12 +248,9 @@ export class AppKitProvider implements IETHProvider {
 
   async disconnect(): Promise<void> {
     const config = this.getWagmiConfig();
-    try {
-      await wagmiDisconnect(config);
-    } finally {
-      this.address = undefined;
-      this.chainId = undefined;
-    }
+    await wagmiDisconnect(config);
```
gbarkhatov: **Merge-blocker — dropping the `finally` means a rejected disconnect leaves a stale cached address.**

The removed `try/finally` guaranteed `address`/`chainId` were cleared whichever way `wagmiDisconnect` went. Now a rejection skips both lines below and the cache keeps answering `getAddress()`. That matters more than it used to, because this PR also moved React state clearing in `ETHWalletProvider` out of `disconnect()` and into the connector's `disconnect` event — and `WalletConnector.disconnect` does not emit that event when it rethrows. So one rejection now leaves the provider cache, the connector's `_connectedWallet`, and the React `address` all claiming connected at once, with no path that re-syncs them.

This is reachable: `@reown/appkit-adapter-wagmi@1.8.12`'s `WalletConnectConnector.disconnect()` rethrows anything that does not match `/No matching key/i`, so a relay or session failure — the case the visibility check exists to catch — gets you here.

Restoring the `finally` is the smallest fix. It composes correctly with the `WalletConnector` change too: local teardown always completes, and the refusal (which is thrown before any AppKit call, so it is genuinely atomic) is what preserves state.

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/provider.ts
```diff
@@ -229,17 +233,37 @@ export class AppKitBTCProvider implements IBTCProvider {
...
+        throw new WalletError({
+          code: ERROR_CODES.SHARED_SESSION_DISCONNECT_REFUSED,
+          message:
+            "Bitcoin and Ethereum share one wallet session. Disconnecting Bitcoin alone would also disconnect Ethereum. Disconnect all wallets instead.",
+          wallet: APPKIT_PROVIDER_NAME,
+          chainId: "BTC",
+        });
+      }
+      await modal.disconnect("bip122");
+      this.clearSession();
```
gbarkhatov, line 256: `clearSession()` only runs on the success path here, unlike the `"all"` branch just below which correctly puts it in a `finally`. The asymmetry assumes a rejection from `modal.disconnect("bip122")` means bip122 survived, but AppKit does not work that way — at `appkit-base-client.js:609-620` the per-namespace disconnects settle under `Promise.allSettled`, then `SendController.resetSend()`, `ConnectionController.resetWcConnection()` and `ConnectionController.syncStorageConnections()` all run, and only *then* does it throw if any promise rejected.

So by the time this rejects, AppKit has already torn the session down. Combined with the `WalletConnector` finding, the result is a provider holding a live `address`/`publicKey` for a session that no longer exists.

Worth catching, re-checking `modal.getAccount("bip122")?.isConnected`, and clearing when AppKit actually disconnected. The pre-flight refusal above is the one case that genuinely is atomic — it throws before any AppKit call — so it can keep preserving state.

packages/babylon-wallet-connector/src/hooks/useWalletConnectors.tsx
```diff
@@ -86,6 +86,16 @@ export function useWalletConnectors({ persistent, accountStorage, onError }: Pro
     (chainId) => connectors[chainId as ChainId]?.connectedWallet,
   );
 
+  const dropRejectedWallet = (connector: Pick<IConnector, "id" | "disconnect">, wallet: IWallet) => {
+    wallet.account = null;
```
gbarkhatov: This mutation is load-bearing in a way nothing at either end says out loud: it exists so that `WalletConnector.disconnect`'s `this._connectedWallet?.account !== null` check will skip the rethrow and complete local teardown. So `account` is being used as a control signal between two layers, spelled as a data field, with the reader in a different file and no comment linking them.

It works, and it is what closes the reopen path. The risk is that it is invisible from the other side: anything that legitimately nulls `account` on a live wallet — an `accountsChanged` handler, a session-expiry path — silently converts a real user-initiated disconnect into a local-only clear while the remote session stays up, which is the desync #2352 exists to prevent. A reader of `WalletConnector.ts` alone cannot tell the check means "my caller pre-marked this as rejected".

CLAUDE.md prefers a discriminated union over implicit state here. A third `DisconnectScope` member — `"local"`, which skips the provider call outright — would let this pass its intent explicitly, would let the `account` check be deleted, and would also give `useAppKitBtcBridge.ts:127` the non-refusable teardown it actually wants.

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { __resetSharedBtcAppKitConfigForTests, btcDisconnectWouldDropEthereum, setSharedBtcAppKitConfig } from "../sharedConfig";
+
+afterEach(() => {
+  __resetSharedBtcAppKitConfigForTests();
+});
+
+function setModal(getProviderType: () => string, getAccount: () => { isConnected: boolean } | undefined) {
+  setSharedBtcAppKitConfig({
+    modal: { getProviderType: vi.fn(getProviderType), getAccount: vi.fn(getAccount) } as never,
```
gbarkhatov: Both mocks are zero-arg, so they return the same value no matter which namespace they are asked about. The entire point of `btcDisconnectWouldDropEthereum` is *which* namespace it interrogates — and with these fixtures, swapping `getProviderType("bip122")` and `getAccount("eip155")` in the predicate would ship a guard that refuses on the wrong chain with all six tests still green.

An `expect(modal.getProviderType).toHaveBeenCalledWith("bip122")` and its `"eip155"` counterpart on at least the refusal and injected cases would close it. Making the mocks dispatch on the namespace argument would be better still, and would let the "mixed session" case in `provider.disconnect.test.ts` actually set up a mixed session — right now it cannot express *different* types for bip122 and eip155, so that test is a strict subset of the plain WalletConnect one despite its name.

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      () => "AUTH",
+      () => ({ isConnected: true }),
+    );
+
+    expect(btcDisconnectWouldDropEthereum()).toBe(true);
+  });
+
+  it("reports false for an injected bitcoin wallet even while ethereum is connected", () => {
+    setModal(
+      () => "INJECTED",
```
gbarkhatov: Minor, but this arm of the acceptance matrix pins a value bip122 never produces. All four Bitcoin connectors in `@reown/appkit-adapter-bitcoin@1.8.12` set `this.type = 'ANNOUNCED'` (`WalletStandardConnector.js:12`, `SatsConnectConnector.js:13`, `OKXConnector.js:20`, `UnisatConnector/index.js:12`) — `INJECTED` does not occur, so the real path an injected Bitcoin wallet takes is untested. Production behaviour is unaffected since both fall through to `false`; this is test fidelity, not a live bug.

Same kind of thing two tests up: `AUTH_CONNECTOR_SUPPORTED_CHAINS` is `['eip155', 'solana']` in `@reown/appkit-common@1.8.12`, so bip122 can never get an Auth connector and the `"AUTH"` arm of the predicate is currently unreachable. There is a real tension worth resolving rather than quietly deleting, since #2352 explicitly asks for Auth coverage — either keep the branch with a comment marking it defensive against future AppKit support, or drop branch and test together.

`ANNOUNCED` here (plus an `EXTERNAL` case) would make this test mean what its name says.

packages/babylon-wallet-connector/src/components/Chains/container.tsx
```diff
@@ -68,8 +68,14 @@ export function ChainsContainer(props: ContainerProps) {
         const appkitBtcWallet = btcConnector?.wallets.find(w => w.id === APPKIT_BTC_CONNECTOR_ID);
 
         if (appkitBtcWallet && btcConnector?.wallets.length === 1) {
-          // Already connected: reopen the AppKit modal so the user can switch/disconnect.
           if (btcConnector.connectedWallet) {
+            const { btcDisconnectWouldDropEthereum } = await import("@/core/wallets/btc/appkit/sharedConfig");
+            if (btcDisconnectWouldDropEthereum()) {
+              await btcConnector.disconnect().catch(() => undefined);
```
gbarkhatov: `.catch(() => undefined)` is an empty catch by another spelling, which CLAUDE.md's Error Handling rule rules out. The refusal does still reach the user, but only as a side effect of `WalletConnector` emitting `error` into a subscription in a different file — invisible from this call site. A non-refusal failure here (relay, wallet RPC) produces no log at all and the row simply does nothing.

The discriminating handler you want already exists a few files over at `useWalletConnectors.tsx:91-94` — return on `SHARED_SESSION_DISCONNECT_REFUSED`, `console.error` the rest. Since two sites now want identical behaviour, extracting it as a shared helper would be reasonable.

Design question while here: this block has already evaluated `btcDisconnectWouldDropEthereum()` and knows the call will be refused, then makes it anyway so the provider can re-evaluate the same predicate and throw. Calling `displayError` directly would express the intent more plainly than round-tripping through a call known to fail.

packages/babylon-wallet-connector/src/hooks/useWalletConnectors.tsx
```diff
@@ -287,6 +294,23 @@ export function useWalletConnectors({ persistent, accountStorage, onError }: Pro
           return;
         }
 
+        if (
+          error instanceof WalletError &&
+          error.code === ERROR_CODES.SHARED_SESSION_DISCONNECT_REFUSED &&
+          displayError
+        ) {
+          displayError({
```
gbarkhatov: On the `dropRejectedWallet` path this dialog is set and then immediately replaced. Because that helper nulls `account` first, `WalletConnector.disconnect` skips the rethrow and emits **both** `error` and `disconnect`; this handler sets the refusal screen, then the `disconnect` handler at line 252 calls `displayChains?.()` and the last `setState` wins. So a user who cancels the key-mismatch warning gets bounced to the chain list, and the explanation they were meant to see flashes and disappears.

Worth saying that this is narrower than it first looks, and #2352's "the UI explains the refusal" criterion is genuinely met on the primary path: from the connected Bitcoin row, `account` is still set, so the connector rethrows before `emit("disconnect")` and the dialog persists. It is only the rejected-wallet cleanup that loses it — where arguably no refusal dialog should appear at all, since the user never asked for a disconnect.

Either skip `displayChains?.()` when the current screen is the error screen, or give `dropRejectedWallet` the non-refusable `"local"` teardown discussed above, which never contacts AppKit and so never emits this.

packages/babylon-wallet-connector/src/hooks/__tests__/useWalletConnect.optionalChains.test.tsx
```diff
@@ -154,11 +154,38 @@ describe("disconnect", () => {
...
+  it("passes the chain scope when disconnecting a single chain", async () => {
+    const { result } = setup({
+      requiredChainIds: ["ETH"],
+      selectedWallets: { BTC: btcWallet, ETH: ethWallet },
+      confirmed: true,
+    });
+
+    await result.current.disconnect("BTC");
+
+    expect(disconnectBtc).toHaveBeenCalledWith();
```
gbarkhatov: The assertion says the opposite of the test name: `toHaveBeenCalledWith()` asserts that *no* scope argument is passed and the parameter default applies. As written the test would keep passing if the chain path stopped meaning `"chain"` altogether.

The cleaner fix is at the call site rather than in the test — `useWalletConnect.ts:55` currently relies on the default, and passing `"chain"` explicitly there would make the intent visible next to the `"all"` on line 62 and make this title true. (`container.test.tsx:89-90` asserts the same empty-args shape.)

Also worth a look nearby: line 157's `invocationCallOrder` check pins BTC-before-ETH ordering that I do not think is a stated requirement, and depends on connector object key order; and this file and `container.test.tsx` simulate the refusal with a plain `Error("refused")` rather than a `WalletError` carrying the code — so neither would catch the mis-gating in `WalletConnector.disconnect`.

packages/babylon-wallet-connector/src/providers/ETHWalletProvider.tsx
```diff
@@ -189,12 +185,17 @@ export const ETHWalletProvider = ({ children, callbacks }: ETHWalletProviderProp
...
     if (!ethConnector) return;
 
-    const unsubscribe = ethConnector.on("disconnect", () => {
-      disconnect();
+    return ethConnector.on("disconnect", async () => {
+      setAddress(undefined);
+      setProvider(null);
+      try {
+        // Let the connector clear its wallet before the callback runs.
+        await Promise.resolve().then(() => callbacks?.onDisconnect?.());
```
gbarkhatov: Worth recording *why* this hop is needed, because the reason lives in another file and this looks like a redundant `Promise.resolve()` that a future cleanup will delete. `WalletConnector.disconnect` emits `"disconnect"` on line 69 and only nulls `_connectedWallet` on line 70, so a synchronous handler here would observe a connector that still reports a connected wallet — the comment says what the hop does but not what breaks without it.

The root fix is one line away in a file this PR already touches: capture the wallet into a local, null `_connectedWallet`, then emit. That removes the need for the hop and the comment both, and makes the ordering correct for every subscriber rather than just this one.

gbarkhatov (dismissed):
Approving so this is not gated on my opinion — the findings below are suggestions and the calls are yours. Three of them I do think should be settled before merge, and I have marked those.

I checked the load-bearing external claim against the pinned AppKit rather than taking it on trust, and it holds: `appkit-base-client.js:586-593` widens a namespaced disconnect to every namespace exactly when the namespace's connector is WalletConnect or Auth, and leaves it scoped otherwise. The `INJECTED` exemption and the `"all"` bypass are right, `modal.disconnect("bip122")` really does scope (`ConnectionController.disconnect` maps `namespace` to `chainNamespace`), and disconnect-all can never abort part-way because `"all"` never rethrows. The design is sound; what follows is about the edges.

**The three I would settle first**

1. `WalletConnector.disconnect` preserves the wallet on *every* chain-scope failure, not only on the refusal. AppKit wraps all of its internal failures as a plain ``Error(`Failed to disconnect chains: …`)`` at `appkit-base-client.js:629`, so a relay drop and `SHARED_SESSION_DISCONNECT_REFUSED` are indistinguishable at that branch. #2352 asks that a *refused* request preserve state; nothing asks a *failed* one to.

2. The refusal predicate reads `ProviderController.providerIds`, but AppKit decides whether to widen from `ConnectorController.activeConnectorIds`. Those two fill on different clocks — `ConnectorController.initialize()` rehydrates `activeConnectorIds` synchronously from storage at init, while `providerIds` is only written inside the awaited `syncWalletConnectAccount` and only once a session address resolves. That leaves a window on every load with a persisted WalletConnect Bitcoin session where AppKit will widen but the predicate returns `false`.

3. The Ethereum provider's `try/finally` is removed in this diff, so a rejected `wagmiDisconnect` now leaves the cached address behind — and with finding 1 suppressing the `disconnect` event, the React state that used to be cleared imperatively is not cleared either.

**Not in the diff, so noted here rather than inline**

`useAppKitBtcBridge.ts:123-129` calls `btcConnector.disconnect()` with no scope, so it now defaults to `"chain"` and can be refused. That call site fires when AppKit reports bip122 already disconnected — it wants a purely local sync, not a remote attempt. Because AppKit's per-namespace disconnects run under `Promise.allSettled`, bip122 can clear while eip155 is still connected, which is precisely when the predicate refuses. The optional `scope?` parameter means there is no compile error and CI stays green. A third `DisconnectScope` member (`"local"`) would fix this and finding 5 together, and would let the `account === null` check go away.

Separately, `onError?.(error)` at `useWalletConnectors.tsx:269` runs before any narrowing, so a designed, user-facing refusal reaches host error telemetry as a fault on every shared-session row click. Pre-existing, and outside the diff, but this PR is what starts routing refusals through it.

**Explicitly not raised**

The Ethereum provider has no reciprocal shared-session guard, so an Ethereum-only disconnect can still drop Bitcoin. That is the mirror of the bug being fixed and it looks alarming, but the PR body defers the reverse direction to #2438 and #2352 is Bitcoin-side only — so I am treating it as scope, not a defect. Finding 3 is the part of the Ethereum path that *is* in scope, and is a different problem.

**One process note:** both blockers 1 and 5 live in `5ae352c1` and `39f93e23`, the two commits added in response to review. Those are the least-reviewed commits on the branch — they were written to close review threads and have not themselves been through a round. Worth a second pair of eyes on those two specifically.

Also verified clean: all three standing review comments are genuinely addressed at HEAD (the reopen path is closed at the `connectedWallet?.account` guard, `removeWallet`, the storage delete, and the auto-reconnect key); no `data-testid` moved or dropped; `DisconnectScope` is properly re-exported and the optional parameter keeps `IConnector` backward-compatible.

packages/babylon-wallet-connector/src/core/WalletConnector.ts
jonybur: Changed the discriminator to the error type, and the account check is gone. `dropRejectedWallet` now calls `disconnect("chain")` and falls back to the new `disconnect("local")` on a refusal or a failure, so no wallet mutation is needed. What did not change is keeping the wallet when a chain disconnect fails for another reason, and that is deliberate rather than an accident of the account check. In @reown/appkit 1.8.12 the Bitcoin adapter throws before it deletes the connection and emits `disconnect` (appkit-adapter-bitcoin adapter.js:297-301), so after a rejected `modal.disconnect("bip122")` AppKit and the Reown modal still show bip122 connected. Dropping locally would show Bitcoin disconnected while it is not, and the next row click would re-adopt the same account through the bridge with no prompt. #2352's Problem section names 'a provider error can also clear local Bitcoin state and report a successful disconnect' as a defect, so main's behaviour is the one the issue asks to remove. The failure is no longer silent: it is emitted on `error` (host onError) and the caller's promise rejects. The widened path, where bip122 really did fall while another namespace failed, is healed by the bridge, which now calls `disconnect("local")` when AppKit reports bip122 gone (new useAppKitBtcBridge test). On the double handling: the `error` event carries faults to the host and bounces the widget, the rejection carries control flow to the caller; the refusal now uses only the rejection because it is not a fault. Minor: the catch is at appkit-base-client.js:630-631 (:629 is the try's closing brace); same substance. Contract is written on `IConnector.disconnect` in types.ts.

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/sharedConfig.ts
jonybur: Checked line by line against the pinned 1.8.12 source. The two slices are distinct and filled on different clocks as you say (ConnectorController.initialize from storage at appkit-base-client.js:396, setProviderId at :1235). The predicate cannot run inside that window though. It is only evaluated from AppKitBTCProvider.disconnect("chain"), and every caller requires btcConnector.connectedWallet, which exists only after the bridge saw bip122 isConnected, i.e. Boolean(caipAddress). caipAddress is written by syncAccountInfo (:1342) inside `await this.syncAccount` at :1250, strictly after setProviderId at :1235; the adapter paths run syncProvider before syncAccount too (:1194-1195, and :1033 before :1051/:1055), and ChainController.initialize restores no caipAddress from storage (defaultAccountState has none). So whenever a Bitcoin disconnect is possible, getProviderType("bip122") is already resolved; an 'undefined is unsafe' arm would refuse a state AppKit does not produce and add a false-positive refusal the user cannot get past. On the literals: getProviderType is typed `ConnectorType | undefined` (appkit-base-client.d.ts:147), so a renamed literal fails `tsc -b tsconfig.lib.json` with TS2367 (verified by renaming it); the build script runs that tsc first, so it cannot degrade to a silent false. Agreed there is no public getter for activeConnectorIds. No code change here; the namespaces the predicate reads are now pinned by tests that dispatch on namespace (thread 6).

packages/babylon-wallet-connector/src/core/wallets/eth/appkit/provider.ts
jonybur: Left as is, with the reasoning corrected. @wagmi/core 2.22.1 disconnect.js runs `await connector.disconnect()` before `connections.delete` and `setState`, so when the WalletConnect connector rethrows, wagmi itself still reports the connection. `getAddress()` (provider.ts:256) reads `getAccount(config)` first and refills the cache from it, so the removed `finally` is not what keeps answering; wagmi is, and restoring it changes nothing a caller can observe. The three local layers staying connected after a rejection is agreement with wagmi, not staleness; the pre-PR code produced the opposite split (UI disconnected, wagmi and Reown still connected). The refusal never reaches this method: it is thrown by the Bitcoin provider, and the Ethereum `disconnect()` takes no scope. provider.test.ts:133 pins the current behaviour, and it matches the connector rule from thread 1 (a failed chain disconnect keeps the wallet; the failure is reported on `error`). What the rejection does leave is a user-visible silent failure (console.error in ETHWalletProvider); that is the same class as thread 1 and now reaches the host through the connector's `error` event.

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/provider.ts
jonybur: Kept the asymmetry; it mirrors the connector. On `chain` the connector keeps the wallet when AppKit rejects, so the provider keeps its cache; on `all` the connector finishes local teardown, so the provider clears in `finally`. Two corrections to the reasoning: resetSend, resetWcConnection and syncStorageConnections (:611-617) do not touch bip122's account state (getAccount().isConnected reads ChainController accountState.caipAddress, which only onDisconnectNamespace resets at :1146-1148), and in the non-widened path a rejection is the Bitcoin adapter's own `connector.provider.disconnect()` throwing at adapter.js:297, before deleteConnection and the `disconnect` emit, so bip122 survived. Your 'cannot assume survived' does hold for the widened multi-namespace path, and there the bridge sees isConnected false and now calls `disconnect("local")`, which clears the provider cache and the connector (new bridge test). A `finally` here would instead drop a cache the connector still relies on after a plain failure. The why is now in a JSDoc on the method.

packages/babylon-wallet-connector/src/hooks/useWalletConnectors.tsx
jonybur: Removed. `dropRejectedWallet` no longer touches `wallet.account`: it removes the wallet from widget state and storage, calls `disconnect("chain")`, and on a refusal (or any failure) calls the new `disconnect("local")`, which the connector forwards to the provider so AppKitBTCProvider runs clearSession only. The re-select guard in the mount effect is back to `connector.connectedWallet`. Two corrections. The bridge could not be refused on 1.8.12: onDisconnectNamespace resets bip122's provider id (ProviderController.resetChain, :1155) in the same synchronous block that clears its address (:1148), so the predicate reads `undefined` and returns false there; what `local` buys the bridge is dropping a remote echo call that, on failure, left connectedWallet set. And 'skip the provider outright' would leak the connectionEvents listener and the network subscription (clearSession stops both), which is why `local` is forwarded. The accountsChanged nulling risk was hypothetical: only Wallet.ts:58 and this site ever wrote wallet.account.

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
jonybur: Confirmed and fixed. With the zero-arg mocks, swapping the predicate to getProviderType("eip155") / getAccount("bip122") left all fifteen tests across sharedConfig.test.ts and provider.disconnect.test.ts green. sharedConfig.test.ts now builds the modal from a per-namespace map (providerTypes, accounts), every fixture gives bip122 and eip155 different values, and the first case asserts the namespaces called; the same swap now fails five tests. The 'mixed session' provider test dispatches on namespace (bip122 ANNOUNCED, eip155 WALLET_CONNECT) and asserts getProviderType("bip122"), so it is no longer a subset of the first test.

packages/babylon-wallet-connector/src/core/wallets/btc/appkit/__tests__/sharedConfig.test.ts
jonybur: Citations check out: every bip122 connector in appkit-adapter-bitcoin 1.8.12 sets type 'ANNOUNCED' (Leather inherits it from SatsConnectConnector; the sixth is the WalletConnect one), and AUTH_CONNECTOR_SUPPORTED_CHAINS is ['eip155','solana']. The fixtures now say ANNOUNCED (sharedConfig, provider.disconnect and provider.network tests). Not adding EXTERNAL: bip122 never produces that either, so it would pin another value AppKit never returns. Keeping the AUTH arm and its test: #2352's scope says to refuse when 'WalletConnect or Auth' would disconnect Ethereum and asks for Auth provider tests; a JSDoc on btcDisconnectWouldDropEthereum records that the arm is defensive until AppKit offers Auth for bip122.

packages/babylon-wallet-connector/src/components/Chains/container.tsx
jonybur: Fixed, and it became mandatory: after the #2445 merge eth.entry.test.ts walks `import()` like a static import, so the dynamic import of sharedConfig.ts failed 'resolves no Bitcoin or Cosmos wallet implementations'. The container no longer knows about the predicate. A connected AppKit Bitcoin row calls useWalletConnect().disconnect("BTC"); that hook shows the 'Wallets share one session' dialog on a refusal and rethrows, the container ignores that one error (already explained) and logs anything else. No blanket catch is left. Behaviour change to note: an announced (non-shared) AppKit Bitcoin row now disconnects bip122 instead of reopening Reown; clicking the row again opens Reown for a fresh choice. Not extracting a catch-site helper: the shared piece is `isSharedSessionRefusal` in @/error, used by the connector, both hooks and the container; the two catch sites do different things after it. One correction: a non-refusal failure was never silent to the host, WalletConnector emits it on `error`; only the call-site log was missing. eth.entry.test.ts 12/12, container.test.tsx rewritten.

packages/babylon-wallet-connector/src/hooks/useWalletConnectors.tsx
jonybur: Confirmed by tracing the real connector and hook: on cancel the order was displayChains, then the refusal's error event (dialog), then the disconnect event (displayChains); React 18 batched the last two, so the refusal screen never painted, and the host's onError still received a fault. Fixed at the root rather than by ordering: a refused chain disconnect is now a rejection only, never an `error` event (WalletConnector rethrows before the emit), and the dialog is shown by useWalletConnect().disconnect(chain), the one user-facing caller. dropRejectedWallet therefore sees the refusal as a return value, falls back to `disconnect("local")`, and no dialog or error report happens on that path. Neither proposed fix was used: skipping displayChains on ERROR also changes the 'Connection Failed'/'Update wallet' bounce, and a local-only teardown would leave a rejected announced AppKit wallet connected in AppKit (the bridge re-adopts it on the next connect), which is why the chain disconnect is still attempted first. Side effect worth knowing: after #2445 the hook test passed no btcValidation and walked 'Connection Failed'; it now passes a validator and walks the mismatch path it was written for.

packages/babylon-wallet-connector/src/hooks/__tests__/useWalletConnect.optionalChains.test.tsx
jonybur: Fixed all three. useWalletConnect passes 'chain' explicitly and the test asserts toHaveBeenCalledWith('chain'); the invocationCallOrder line is gone (the Bitcoin 'all' path calls modal.disconnect() with no namespace and the Ethereum provider ignores the scope, so either order ends the same). The plain Error case is kept and renamed to the failed-disconnect case it actually is (no dialog, rejection propagates), and a separate refusal case with the real WalletError asserts the dialog and the rejection. container.test.tsx rejects with the WalletError for the refusal case and with a plain Error for the logged-failure case. Minor: the empty-args assertion was at container.test.tsx:84, not :89-90.

packages/babylon-wallet-connector/src/providers/ETHWalletProvider.tsx
jonybur: Applied. WalletConnector captures the wallet, nulls `_connectedWallet`, then emits `disconnect`, so every subscriber sees a cleared connector; the Promise.resolve() hop and its comment in ETHWalletProvider are gone. Every `on("disconnect")` subscriber in the package, the vault and simple-staking reads the emitted payload or nothing, so none depended on the old order. What the hop was actually guarding is a synchronous re-entry into disconnect() (e.g. a host onDisconnect that calls disconnectAll) running wagmiDisconnect twice and emitting a second event with a null payload; the reorder closes that structurally, and a new connector test pins that connectedWallet is null during the event.

jonybur:
What changed since your review (one commit on top of 39f93e23, plus the origin/main merge with #2445):

- A refused Bitcoin disconnect is now a rejection, not an `error` event. WalletConnector rethrows SHARED_SESSION_DISCONNECT_REFUSED before the emit; `useWalletConnect().disconnect(chain)` shows the 'Wallets share one session' dialog and rethrows. The refusal can no longer reach a host's onError (NOTE-B), no dialog can flash on the rejected-wallet path (thread 9), and the `wallet.account = null` control signal is gone (thread 5).
- Third DisconnectScope `"local"`: forget the wallet with no remote call; never refused. The bridge uses it once AppKit reports bip122 gone (NOTE-A); the rejected-wallet path tries `"chain"` first so an AppKit wallet is still released, then falls back to `"local"`. AppKitBTCProvider handles it with clearSession only.
- The chains container no longer imports the Bitcoin predicate (the #2445 entry test walks `import()` now). A connected AppKit Bitcoin row disconnects through the connector; a shared session is refused and explained, an announced session now disconnects bip122 instead of reopening Reown (thread 8).
- Connector nulls its wallet before emitting `disconnect`; the microtask hop in ETHWalletProvider is gone (thread 11). `useWalletConnect` passes `"chain"` explicitly (thread 10).
- Tests: namespace-dispatching AppKit mocks (thread 6), ANNOUNCED fixtures with the AUTH arm kept and documented (thread 7), new cases for the refusal dialog, the local scope, the failed-chain-keeps-wallet rule, the null-before-emit order, and the bridge.
- Not changed, with the AppKit and wagmi evidence in-thread: a failed (not refused) chain disconnect still keeps the wallet (threads 1, 3, 4). AppKit keeps bip122 when its adapter throws (adapter.js:297-301) and wagmi keeps the connection when its connector throws, so dropping locally would misreport; #2352 lists that misreport as a defect. The failure is reported on `error` and the caller's promise rejects. Thread 2's rehydration window is not reachable through this predicate; the literal tie is enforced by tsc (TS2367).

Verification: tsc clean, eslint 0 errors, vitest 22 files / 328 tests, `pnpm run build` green (boundary tests 23/23), eth.entry.test.ts 12/12 and the two #2445 connect tests 4/4 on the merged tree.

Public API: `DisconnectScope` (root entry only, not `./eth`) gains the member `"local"`. Additive for callers; breaking only for code that switches exhaustively over the scope or implements `IConnector` itself — no consumer in this repo does either. `isSharedSessionRefusal` is now exported alongside the error code so a host can discriminate the rejection.

Before asking for re-review this change went through a five-lens adversarial pass of its own, since it is the third commit written in response to review. That found no blocking defect and three important ones, all fixed here: the rejected-wallet path could re-select the wallet while its remote disconnect was still in flight (a `visible` toggle re-ran the connect effect); the failure half of the chain→local fallback was unpinned; and a deleted test had pinned the error-event subscriber, which this change does not touch. Each fix is pinned by a test that was checked to fail without it.

packages/babylon-wallet-connector/src/core/WalletConnector.ts
```diff
@@ -54,19 +54,23 @@ export class WalletConnector<N extends string, P extends IProvider, C> implement
...
+        await provider.disconnect(scope);
+      } catch (error) {
+        if (scope === "chain" && isSharedSessionRefusal(error)) throw error;
+        this._ee.emit("error", error instanceof Error ? error : new Error(String(error)));
+        if (scope === "chain") throw error;
       }
-      this._ee.emit("disconnect", this._connectedWallet);
-      this._connectedWallet = null;
     }
+    if (this._connectedWallet !== wallet) return;
```
jrwbabylonlab: This guard is identity-only and wallet objects are reused, so reconnecting the same wallet while a disconnect is in flight passes it — the resolving disconnect then clears the fresh connection and its persisted session. A generation counter would close it.

gbarkhatov (dismissed):

gbarkhatov (approved):
@jonybur checks failing

jonybur (commented):
Re-reviewed at c16ea218. The scope model is much better than the previous round: keying the connector branch on the error type rather than on `account`, the generation counter, and the new `local` scope all land where they should, and the `dropRejectedWallet` chain-then-local fallback is the right shape. CI is green now.

One thing the `local` scope was introduced for, and did not get applied to.

**`ETHWalletProvider` still syncs an already-gone wallet through the refusable `chain` scope.** `providers/ETHWalletProvider.tsx:55` calls `ethConnector?.disconnect()` with no argument, so it takes the `chain` default. Three call sites reach it, and all three mean "the wallet is already gone, catch local state up":

- `onAccountsChanged` when the wallet reports zero accounts and there was a previous address (`:233`)
- `checkETHConnection` when `provider.getAddress()` comes back empty (`:265`)
- `checkETHConnection` catch, whose own comment says "wallet likely disconnected" (`:275`)

With a shared WalletConnect or Auth session and Bitcoin connected, `AppKitProvider.disconnect("chain")` throws `SHARED_SESSION_DISCONNECT_REFUSED` before it touches wagmi. `WalletConnector` rethrows it for chain scope and deliberately does not emit `disconnect`, which is correct for a user-initiated request. But here nothing was requested — the wallet is already gone. The refusal is swallowed by the catch at `:56` as a `console.error`, so the `on("disconnect")` subscriber at `:187` never runs: `setAddress(undefined)`, `setProvider(null)` and `callbacks.onDisconnect` are all skipped. `useETHWallet` keeps reporting a connected ETH address that the wallet dropped, and because `useVisibilityCheck` is gated on `Boolean(provider && address)`, the check re-fires and is refused again on every tab focus.

This is the same defect as the Bitcoin bridge one, which this PR fixed by passing `local` at `useAppKitBtcBridge.ts:133`. The ETH side has the identical pattern and kept the default. Passing `"local"` from these three sync paths fixes it, and matches the scope doc in `core/types.ts`: "one the provider backend already reports disconnected".

Worth deciding separately what the context-exposed `disconnect` at `:293` should do, since it shares this function. A host calling it on a shared session currently gets a silent no-op plus a console line. If it is meant to stay user-initiated, it needs to be a distinct callback from the three sync paths so the refusal can still surface.

Two smaller notes, neither blocking:

- `ChainsContainer` now disconnects a connected Bitcoin row outright, while a connected Ethereum row still opens the AppKit modal whenever the session is not shared. The stated intent is to keep the dialog on the Ethereum-only entry, but the effect is that a user with a non-shared Bitcoin session loses the only way to switch Bitcoin wallets from that row and instead gets disconnected with no confirmation. Worth confirming that asymmetry is what you want.
- The stale greptile P1 on `ETHWalletProvider` describes the pre-reorder code and no longer applies — the React state clearing moved into the `on("disconnect")` subscriber. Worth resolving so it does not read as open.

Everything else I checked holds: the disconnect-all loop cannot abort part-way because non-chain scopes never rethrow, the generation counter closes the reconnect race for the identity check, `isSharedSessionRefusal` keeps designed refusals off the `error` channel, and `clearSession` runs in a `finally` on the Bitcoin `all` path.

jonybur (commented):
One thing on the new Ethereum-side predicate, `ethDisconnectWouldDropBitcoin` in `packages/babylon-wallet-connector/src/core/wallets/eth/appkit/sharedConfig.ts:64-69`. The `AUTH` arm does not mirror cleanly from the Bitcoin twin, and I think it refuses a disconnect that is safe.

AppKit's auth connector serves only `eip155` and `solana` - `AUTH_CONNECTOR_SUPPORTED_CHAINS` is `['eip155','solana']` in `@reown/appkit-common@1.8.12`, the same pin `btcDisconnectWouldDropEthereum`'s JSDoc cites when it calls its own `AUTH` arm defensive. On the Bitcoin side that arm costs nothing precisely because bip122 never gets an auth connector, so the branch is unreachable. Here the direction is reversed: eip155 does get one, so the branch is reachable, and whatever serves bip122 is a different connector in every case.

Concretely: a user signs into Ethereum with email or a social login and connects a Bitcoin wallet separately. `getProviderType("eip155")` is `AUTH`, `getAccount("bip122")?.isConnected` is true, so line 67 sets `sharesSession` and the function returns true. `AppKitProvider.disconnect("chain")` then throws `SHARED_SESSION_DISCONNECT_REFUSED`, and `useWalletConnect().disconnect("ETH")` shows "Wallets share one session - Disconnecting Ethereum alone would also disconnect Bitcoin." Neither statement holds, and the only way past that dialog is disconnect-all, which drops the Bitcoin wallet the guard exists to protect. `wagmiDisconnect` only tears down the wagmi connection, so bip122 was never at risk on this path.

This is reachable in the shipped apps rather than theoretical: `createAppKit` in `core/wallets/appkit/appKitModal.ts:132` passes no `features`, so AppKit's email and social logins are on by their defaults.

The matrix in `eth/appkit/__tests__/provider.test.ts` pins the behaviour, so the arm looks deliberate rather than accidental. If it is deliberate, this function could use the same JSDoc treatment as its Bitcoin twin, naming the AppKit behaviour it defends against for eip155. If it is symmetry carried over from that twin, dropping `AUTH` from `sharesSession` leaves `WALLET_CONNECT`, which is the connector type that genuinely widens.

jonybur:
@greptileai

Please review the full PR at current commit `d1e1507fa2a7534a52738b7c8e3c92e563700cf3` against `main`. The existing 4/5 score covers `46b93452bf7fb4c7e96f81a622b825ebe022750d`. All 15 review threads are resolved, and current CI passes.

Check the failed-disconnect state, rejected-wallet restore, shared-session refusal, and reconnect race fixes. Report any remaining findings and the confidence score for this commit. This PR is a merge dependency of #2499.

gbarkhatov (dismissed):

jonybur:
The two review findings are fixed in [`cbf3c6715d8d`](https://github.com/babylonlabs-io/babylon-toolkit/commit/cbf3c6715d8da08a2fce22b5e1e0d059669d494a).

- Ethereum backend-loss events now use local cleanup. Empty accounts, an empty visibility check, and a failed visibility check clear Ethereum state without disconnecting Bitcoin. Public Ethereum disconnect still shows the refusal dialog and rejects a shared WalletConnect request.
- Independent Ethereum Auth sessions can disconnect. The WalletConnect guard and the defensive Bitcoin Auth guard remain. The pinned AppKit 1.8.12 Auth connector disconnects its frame provider. Wagmi disconnects only the active connector.

Three regression checks failed before the fixes and pass now. All 358 connector tests and 23 package boundary tests pass. The Nx build and four dependency tasks pass. Full package lint has no errors and 89 existing warnings. Commit authentication passes. Fresh CI and reviewer clearance remain.

@greptileai review the full PR at `cbf3c6715d8da08a2fce22b5e1e0d059669d494a` against `main`. Check these fixes and report any remaining findings and the confidence score for this commit. The earlier 5/5 covered `d1e1507fa2a7534a52738b7c8e3c92e563700cf3`.

jonybur (approved):
The `AUTH` arm from the last round is gone. `ethDisconnectWouldDropBitcoin` now refuses only on `WALLET_CONNECT`, so an Ethereum session opened through email or a social login, with a Bitcoin wallet connected separately, disconnects normally instead of hitting a shared-session dialog that did not apply to it — and instead of pushing the user toward disconnect-all, which would have dropped the very wallet the guard exists to protect.

The Bitcoin twin keeps its `AUTH` arm and now carries the JSDoc recording why: AppKit limits Auth to `eip155` and `solana` at the pinned version, so that arm is unreachable on bip122 and costs nothing there. That is the right asymmetry, and it is now written down rather than inferred.

The scope model reads consistently across the layers. `local` is what the sync paths use where the backend already reports the wallet gone, `chain` is the refusable one and rejects rather than emitting an `error` event, and `all` finishes local teardown in a `finally` so it cannot abort part-way. The generation counter keeps a resolving disconnect from clearing a later reconnection to the same wallet object, which the identity-only guard could not do.

Collapsing the two AppKit branches in `ChainsContainer` into one keeps both connect paths intact and routes a connected row through `disconnect(chain.id)`, with `isSharedSessionRefusal` keeping the designed refusal off the error log while a genuine failure still gets reported.

All 12 checks pass at this head.

gbarkhatov (approved):

jrwbabylonlab (approved):

PR #2440, opened 2026-09-08
fix(vault): say the peg-in TVL cap is reached in deposit errors

### LINKED ISSUE #2433: Deposit error copy: say the peg-in TVL cap is reached, not "Remaining capacity"
## Context

A user hit a deposit error worded around "Remaining capacity". The wording does not say what happened — the peg-in TVL cap has been reached.

## Changes

**Now:** `services/vault/src/services/deposit/validations.ts:225-229` (`capBelowMinimumLabel`):

```ts
return `Remaining capacity (${formatSatoshisToBtc(effectiveRemaining)} BTC) is below the minimum deposit (${formatSatoshisToBtc(minDeposit)} BTC)`;
```

Produced from the submit gate at `validations.ts:349` and from inline validation at `services/vault/src/hooks/deposit/useDepositValidation.ts:78-83`, in the case where the cap has a small positive remainder below the protocol minimum.

A second string at `validations.ts:374` reads `BTCVault size exceeds remaining capacity (… BTC)` and fires when the entered amount exceeds a non-zero remaining cap.

**Expected:** say the peg-in TVL cap has been reached. Rejected wording: "Remaining peg in capacity for the network" — it implies TBV cannot take any more peg-ins at all, which is not what happened.
**Where:** `validations.ts:225-229,374`

Not this one: the fully-exhausted case already reads correctly — "Supply cap reached — deposits temporarily paused" (`validations.ts:334-339`, fires when `effectiveRemaining === 0n`).

## Not proven

`validations.ts:198` carries a comment that this mirrors `validateRemainingCapacity` in `@babylonlabs-io/ts-sdk` and that "message strings must match exactly". Check the SDK before renaming, and either change both or drop that constraint deliberately.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788179005910819)

> We may need to improve on this error message. Instead of "Remaining capacity", perhaps we can say "Remaining peg in capacity for the network".
>
> — reply: I'd rather say that the pegin TVL cap has been reached. "Remaining peg in capacity for the network" might imply that TBV can't handle any more pegins

</details>

jeremy-babylonlabs:
## Outcome

Deposit errors caused by the peg-in TVL cap say so. "Peg-in TVL cap reached — only 0.005 BTC remains" replaces "BTCVault size exceeds remaining capacity", and the below-minimum case reads "Peg-in TVL cap reached — only 0.003 BTC remains, below the minimum deposit of 0.005 BTC".

## Change

- Both strings live in `COPY.deposit.errors` as `exceedsCap` and `capBelowMinimum`.
- `validateRemainingCapacity` in the vault wraps the SDK validator and replaces its message for the positive-remainder case. The fully exhausted case still returns the SDK's "Supply cap reached — deposits temporarily paused".
- The submit-gate CTA and the inline validation use the same copy.

## Safety

Copy only. Validation outcomes, thresholds, and the SDK are unchanged. The old comment claiming the strings must match the SDK exactly is replaced, since the vault now owns the wording deliberately.

## Verification

| Gate | Result |
|---|---|
| `vitest` for `services/deposit` and `hooks/deposit` | 27 files, 450 tests pass |

Two new tests pin the wrapper: positive remainder gets the new copy, zero remainder passes the SDK result through.

## Links

Closes #2433.

### DIFF
--- a/services/vault/src/copy.ts
+++ b/services/vault/src/copy.ts
@@ -851,6 +851,10 @@ export const COPY = {
       // broadcast, so no funds are locked and the user can retry once it resumes.
       protocolPaused:
         "New deposits are temporarily disabled while the protocol is frozen or paused. No Bitcoin was sent — please try again once it resumes.",
+      capBelowMinimum: (remainingBtc: string, minBtc: string) =>
+        `Peg-in TVL cap reached — only ${remainingBtc} BTC remains, below the minimum deposit of ${minBtc} BTC`,
+      exceedsCap: (remainingBtc: string) =>
+        `Peg-in TVL cap reached — only ${remainingBtc} BTC remains`,
       cannotActivateInState: (state: string) =>
         `Cannot activate: BTCVault is in ${state} state. Activation is only valid when VERIFIED.`,
       // Deliberately worded without the token "broadcast". These are state
--- a/services/vault/src/hooks/deposit/__tests__/useDepositValidation.test.tsx
+++ b/services/vault/src/hooks/deposit/__tests__/useDepositValidation.test.tsx
@@ -146,7 +146,7 @@ describe("useDepositValidation", () => {
 
       expect(validationResult.valid).toBe(false);
       expect(validationResult.error).toBe(
-        "Remaining capacity (0.00005 BTC) is below the minimum deposit (0.0001 BTC)",
+        "Peg-in TVL cap reached — only 0.00005 BTC remains, below the minimum deposit of 0.0001 BTC",
       );
     });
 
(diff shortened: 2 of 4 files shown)


jonybur (approved):
Checked the wrapper against the SDK validator it wraps. `validateRemainingCapacity` in the SDK has exactly three outcomes: valid, `effectiveRemaining === 0n` giving "Supply cap reached", and `amount > effectiveRemaining` giving the exceeds-capacity string. The guard here passes through the first two and rewords only the third, so no other failure reason can be swallowed or relabelled — that was the thing worth checking, and it holds.

Both strings live in `copy.ts` and follow the conventions at the top of that file: "Peg-in" hyphenated and only capitalised because it leads the sentence, and the value interpolation is a function in `copy.ts` rather than assembled in the caller. `capBelowMinimumLabel` and the CTA branch now build their text from the same two entries, so the inline error and the submit-gate label cannot drift apart.

The re-export swap is clean: `validateRemainingCapacity` is dropped from the pass-through block and redefined locally, and `useDepositValidation` is the only consumer, reaching it through `depositService`. The two new tests pin both halves of the branch, including that the fully-exhausted case still returns the SDK's own wording. Replacing the stale comment about matching the SDK exactly is right, since the vault now owns this wording on purpose.

CI is green.

PR #2441, opened 2026-09-08
fix(vault): truncate a long collateral total inside its stat card

### LINKED ISSUE #2428: Vaults summary: a long collateral total overflows its stat card
## Context

A large deposit amount breaks the layout of the vaults summary strip.

## Changes

**Now:** the stat column is `flex-[1_0_0]` — `flex-shrink: 0`, so it refuses to shrink (`services/vault/src/components/simple/PositionStatCards.tsx:25`) — and the value span is `flex items-center gap-2 text-xl … xl:whitespace-nowrap` with no `truncate`, no `overflow-hidden` and no `min-w-0` (`:40`). A long formatted BTC or USD string has nowhere to go but out of the card. Consumed for `totalCollateralBtc` by `services/vault/src/components/vaults/VaultsSummaryCard.tsx:56-61`.
**Expected:** the value truncates or wraps inside its card.
**Where:** `PositionStatCards.tsx:25,40`

This is the one value in the summary strip with the missing-truncate and non-shrinking combination. Per-row amounts already handle it — `VaultsActiveSection.tsx:72` uses `min-w-0 truncate`, `VaultsLifecycleSections.tsx:222-223` uses `truncate`.

## Not proven

Confirm from the screenshot whether the oversized value was the summary total or a row amount; the summary card is the stronger candidate.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788269208122279)

> big deposit number breaks css

</details>

jeremy-babylonlabs:
## Outcome

A long collateral total truncates inside its stat card instead of pushing out of the summary strip.

## Change

In `PositionStatCards`, `min-w-0` on the section and on the inner column removes the `min-width:auto` content floor so a long value can be squeezed. Both are required: the column is a row-flex item of the section, so dropping either one restores the floor. `flex-[1_0_0]` → `flex-1` is not what lets it shrink: both are grow 1 / basis 0, and with a zero basis `flex-shrink` never engages.

The value truncates at `xl` only (`xl:truncate`), keeping the sub-`xl` wrapping this component gates behind `xl:`, and carries `title` so the full figure stays reachable.

The caption row gets the same treatment. It carries the strings the issue is about — the USD total and the liquidation order — and without clipping it escapes over the divider at `xl`.

The clip sits on the shared value span, so a custom `valueNode` is bounded by the column edge too, not only the plain `value` branch.

## Verification

| Gate | Result |
|---|---|
| `vitest` for `PositionStatCards.test.tsx` | 7 tests pass |

Tests render a long value and a long caption and assert the `xl:truncate` treatment, the `title` attributes, and `min-w-0` on both the section and the column.

## Links

Closes #2428.


jonybur (dismissed):
The fix is the right shape. `flex-[1_0_0]` pinned `flex-shrink: 0`, which is what let a long value push the column past the card; `flex-1` keeps the same grow and zero basis, so the columns still divide the row evenly, and only the shrink behaviour changes. The two `min-w-0` additions are both needed — the outer column and the inner label/value stack are each flex items whose automatic minimum size would otherwise be their content width.

`truncate` on the inner span works here because `overflow: hidden` sets that span's own automatic minimum size to zero, so it can shrink inside the `flex` value row. `valueNode` is left untouched, so the Overview collateral card and the Loans health-factor card keep rendering their custom content unchanged.

The test-file edit to `.parentElement` is the correct follow-through — the plain value now sits one level deeper, and `xl:whitespace-nowrap` still belongs to the wrapper. CI is green.

services/vault/src/components/simple/PositionStatCards.tsx
```diff
@@ -37,8 +37,8 @@ function StatSection({ card }: { card: PositionStatCard }) {
           )}
         </div>
 
-        <span className="flex items-center gap-2 text-xl leading-[1.6] tracking-[0.15px] text-accent-primary xl:whitespace-nowrap">
-          {card.valueNode ?? card.value}
+        <span className="flex min-w-0 items-center gap-2 text-xl leading-[1.6] tracking-[0.15px] text-accent-primary xl:whitespace-nowrap">
+          {card.valueNode ?? <span className="truncate">{card.value}</span>}
         </span>
 
         {card.caption ? (
```
gbarkhatov, line 50: **Merge-blocking, I think:** the caption span just below keeps `xl:whitespace-nowrap` with no `truncate` and no `overflow-hidden`, and it holds exactly the strings #2428 is about — `VaultsSummaryCard.tsx:56` passes `caption: totalCollateralUsd` on the very card the issue names, `PositionOverview.tsx:51-53` passes `` `${formatUsd(...)} USD` ``, and `VaultsSummaryCard.tsx:69` passes `liquidationOrder`, which is unbounded (`copy.ts:1949` joins one amount per active vault with `" → "`).

Before this change, `flex-[1_0_0]`'s `min-width:auto` floored the column at the widest of its three rows, so a long caption widened the card. Now `min-w-0` at `:25`/`:26` removes that floor and nothing in the chain clips, so at `xl`+ the nowrap caption escapes sideways over the divider at `:73` and onto the next column's text. That's a worse failure than the one being fixed.

The sibling the issue points at truncates *every* row inside its `min-w-0` column, not one — `VaultsLifecycleSections.tsx:224` and `:229`. Suggest `truncate` on the caption span. The label at `:27` has the same gap, though `COPY`-driven labels are bounded so it's much lower risk.

services/vault/src/components/simple/PositionStatCards.tsx
```diff
@@ -37,8 +37,8 @@ function StatSection({ card }: { card: PositionStatCard }) {
           )}
         </div>
 
-        <span className="flex items-center gap-2 text-xl leading-[1.6] tracking-[0.15px] text-accent-primary xl:whitespace-nowrap">
-          {card.valueNode ?? card.value}
+        <span className="flex min-w-0 items-center gap-2 text-xl leading-[1.6] tracking-[0.15px] text-accent-primary xl:whitespace-nowrap">
+          {card.valueNode ?? <span className="truncate">{card.value}</span>}
```
gbarkhatov: **Merge-blocking, I think:** the clip lands on the `value` branch only. The description frames "custom `valueNode` content untouched" as a scope line, but `OverviewSection.tsx:58-67` builds a card labelled `totalCollateralValueLabel` whose `valueNode` is `{totalCollateralValue} {collateralBtc}` — the same money-plus-BTC string, on `/`. It inherits `xl:whitespace-nowrap` from `:40`, has no clipping ancestor, and now sits in a column with no min-width floor, so it escapes onto the neighbour just as the caption does.

The other three `valueNode` uses (`VaultsSummaryCard.tsx:78`, `PositionOverview.tsx:68`, `LoansSummary.tsx:70`) are short health-factor content and are genuinely safe — this is the one that isn't.

Suggest `overflow-hidden` on the shared span at `:40` so both branches clip at the column edge, keeping `truncate` here for the ellipsis.

gbarkhatov: Separate, non-blocking: `truncate` includes `white-space:nowrap` at *every* breakpoint, while this component deliberately gates nowrap behind `xl:` on all three rows — PR #2296 added that gating, and the guard test at `:40-47` in the test file records the intent. So on mobile a long total that previously wrapped to a second line, fully readable, now renders clipped with an ellipsis and no way to see the rest.

#2428 accepts "truncates **or** wraps", so this is compliant rather than wrong — flagging it because it's an undiscussed change on the narrowest viewport, for a financial figure. `VaultsLifecycleSections.tsx:228` pairs `truncate` with `title={subLine}` for exactly this reason.

If you want the old sub-`xl` behaviour back: `xl:truncate` plus `title={card.value}`.

services/vault/src/components/simple/PositionStatCards.tsx
```diff
@@ -37,8 +37,8 @@ function StatSection({ card }: { card: PositionStatCard }) {
           )}
         </div>
 
-        <span className="flex items-center gap-2 text-xl leading-[1.6] tracking-[0.15px] text-accent-primary xl:whitespace-nowrap">
-          {card.valueNode ?? card.value}
+        <span className="flex min-w-0 items-center gap-2 text-xl leading-[1.6] tracking-[0.15px] text-accent-primary xl:whitespace-nowrap">
```
gbarkhatov: Nit: this `min-w-0` is inert. The span is a flex item of the `flex-col` container at `:26`, so width is its *cross* axis — CSS automatic minimum size only applies on the main axis, and `min-width:auto` already resolves to `0` here.

Truncation works for a different reason: the inner span's `overflow:hidden` zeroes *its* automatic minimum size inside this row-direction span. Harmless, but it reads as load-bearing on the one line someone will study to understand why the fix works. Suggest dropping it and keeping the genuinely necessary `min-w-0` at `:25` and `:26`. (Unrelated to adding `overflow-hidden` here, which I'd still suggest for the `valueNode` branch.)

services/vault/src/components/simple/PositionStatCards.tsx
```diff
@@ -22,8 +22,8 @@ export interface PositionStatCard {
 function StatSection({ card }: { card: PositionStatCard }) {
   const hasAction = card.actionLabel != null && card.onAction != null;
   return (
-    <div className="flex flex-[1_0_0] items-center justify-between gap-4 xl:max-[1439px]:gap-2">
-      <div className="flex flex-col gap-3">
+    <div className="flex min-w-0 flex-1 items-center justify-between gap-4 xl:max-[1439px]:gap-2">
```
gbarkhatov, line 28: Worth a comment on the line itself, since the description attributes the fix to the wrong half: `flex-[1_0_0]` → `flex-1` isn't what lets this shrink. Both are grow 1 / basis 0, and with a zero basis the sum of flex bases never exceeds the container, so `flex-shrink` never engages either way.

The load-bearing change is `min-w-0` — here and at `:26` — which removes the `min-width:auto` content floor. Both are required: `:26` is a row-flex item of `:25`, so dropping either one restores the floor and the value stops truncating. A short comment saying so would stop a future cleanup "simplifying" one of them away.

services/vault/src/components/simple/__tests__/PositionStatCards.test.tsx
```diff
@@ -59,4 +59,20 @@ describe("PositionStatCards", () => {
...
+    render(
+      <PositionStatCards
+        cards={[{ label: "Total collateral value", value: longValue }]}
+      />,
+    );
+
+    const value = screen.getByText(longValue);
+    expect(value).toHaveClass("truncate");
+
+    const column = value.closest("div.justify-between");
```
gbarkhatov: `closest("div.justify-between")` resolves to the `StatSection` div at `PositionStatCards.tsx:25` — `justify-between` is on `:25`, not on the column at `:26` — so the variable named `column` is actually the section, and these two assertions pass for a different element than the name suggests.

The practical gap: `:26`'s `min-w-0` is the link that actually enables truncation (that div is a row-flex item of `:25`, so without it the column keeps its min-content floor and the value can never be squeezed), and nothing here asserts it. Delete it and truncation silently breaks with all four tests still green.

Suggest asserting `min-w-0` on the inner column too. I'd also consider dropping the `flex-1` assertion — per the comment on `:25` it isn't load-bearing, so it pins a Tailwind class name rather than a behaviour.

services/vault/src/components/simple/__tests__/PositionStatCards.test.tsx
```diff
@@ -59,4 +59,20 @@ describe("PositionStatCards", () => {
     expect(depositButton).toHaveClass("xl:max-[1439px]:w-[100px]");
     expect(depositButton).toHaveClass("w-[120px]");
   });
+
+  it("truncates a long collateral total instead of overflowing the card (#2428)", () => {
```
gbarkhatov: Nit: this file's own convention keeps the `it()` string a plain behaviour statement and puts tracking references in a comment — see `:41`, "Regression guard for PR #2296". `(#2428)` in the description shows up in runner output and reads as noise once the issue is closed. Suggest moving it to a comment above the test.

gbarkhatov (dismissed):
The `value` path is genuinely fixed — `totalCollateralBtc` arrives via `value` (`VaultsSummaryCard.tsx:55`), and the `min-w-0` chain at `:25`/`:26` plus the truncate span's own `overflow:hidden` makes truncation actually engage. Approving so this isn't gated on my opinion; the findings below are suggestions and you're better placed to judge which belong here versus a follow-up.

Two I'd genuinely weigh before merging, because they're the same bug on the other two rows and this change arguably makes their failure mode worse:

1. **The `caption` row is still unclipped.** It carries `totalCollateralUsd` on the very card #2428 names, and `liquidationOrder`, which grows one amount per active vault. Removing the `min-width:auto` floor without clipping that row swaps "the card widens" for "nowrap text paints over the divider and the next column" — there's no `overflow-hidden` anywhere in the chain (`:68` is `rounded-lg … p-6`, and `rounded-lg` doesn't clip).
2. **The `valueNode` branch is uncovered**, and `OverviewSection.tsx:58-67` uses it for the Overview page's collateral total — the same money-plus-BTC string.

One correction to the description, since it'll mislead whoever touches this next: `flex-[1_0_0]` → `flex-1` is **not** what lets the column shrink. Both are grow 1 / basis 0, and with a zero basis the sum of bases never exceeds the container, so `flex-shrink` never engages at all. `min-w-0` is doing all the work.

Worth knowing that the green `visual` job isn't evidence for any of this — `e2e/visual/targets.ts:80-91` captures these routes against recorded fixtures with short values, so no long total or multi-vault `liquidationOrder` has ever been photographed. The overflow path is exactly what that suite can't see.

I checked and found clean: no E2E hook disturbed (`actionTestId` at `:56` and the carry-over comment at `VaultsSummaryCard.tsx:60-64` are untouched by the new wrapper), no `copy.ts` obligation, no critical-path or security-model surface, no dead code. The `.parentElement` edit at test `:33` is correct — it re-points at the span that still carries `xl:whitespace-nowrap`.

services/vault/src/components/simple/PositionStatCards.tsx
```diff
@@ -22,8 +22,8 @@ export interface PositionStatCard {
...
+      <div className="flex min-w-0 flex-col gap-3">
```
jrwbabylonlab, line 29: Now that this column can shrink, the caption below (still `xl:whitespace-nowrap`, no truncate) can bleed under the adjacent button — e.g. a long USD total or the liquidation order. Same truncate treatment for the caption?

gbarkhatov (approved):

jonybur (commented):
Reviewed at ed885cf2. The revision closes the caption and valueNode gaps from the earlier round. Two points on the new code are worth settling before merge.

**1. PositionStatCards.tsx:42 — below xl this clips a money figure to one line with no way to read the rest.**

Ellipsis engages whenever content overflows the line box, and that happens at every breakpoint, not only at xl. A formatted total is one unbreakable run: UAX #14 forbids a break on either side of a comma or a full stop sitting between digits, so the 27-character fixture the new tests use has no break opportunity at all. min-w-0 now removes the column's automatic min-width floor, so the column shrinks to whatever the section leaves it — on a 390px viewport roughly 170px once the 120px action button and the gap come out — and the value renders clipped with an ellipsis and no title attribute.

So dropping truncate in favour of overflow-hidden plus text-ellipsis does not buy the sub-xl wrapping it was made for. Production values hit this too, not just the test fixture: formatUsdValue (utils/formatting.ts:174-177) returns an amount, a space, then USD, so its only break opportunity is before the suffix and the amount itself still clips. That string is the caption on the very card #2428 names (VaultsSummaryCard.tsx:56, fed by useVaultsPageData.ts:75).

The valueNode branch is worse. That span keeps overflow visible, so its automatic minimum size stays at min-content and it cannot shrink — the parent's new overflow-hidden cuts it hard, with no ellipsis at all. OverviewSection.tsx:60-67 renders the collateral total that way, so a clipped figure there reads as a complete, smaller number, with nothing signalling that anything was cut.

A title attribute carrying the full value on this span, and the same on the caption, is the pattern the sibling already uses for exactly this — VaultsLifecycleSections.tsx:228 and :407. It keeps the full figure reachable whichever branch renders.

**2. PositionStatCards.test.tsx:117 — the behaviour this test names does not hold for its own fixture, and its assertions cannot detect that.**

The fixture has no line-break opportunity, so it does not wrap below xl. It clips to one line with an ellipsis, which is the opposite of what the test name and the comment above it assert. The two assertions only guard against someone re-adding those exact class names: xl:whitespace-nowrap lives on the parent row at PositionStatCards.tsx:40 and is a different class string, and nothing in the component ever puts truncate or whitespace-nowrap on this span. They therefore pass for any implementation that keeps clipping, including one where the value is unreadable on mobile. Suggest dropping the test, or retargeting it at something falsifiable — that no unconditional nowrap exists anywhere in the value chain and the only nowrap is the xl-gated one on the parent.

**Checked and clean.** The column query now resolves to the real column at :26 and asserts the load-bearing min-w-0, which was the gap in the previous round. In the valueNode test, getByText matches on direct text children only, so it returns the inner span and its parentElement is genuinely the value row at :40 — that test asserts what it says it does. No data-testid moved or dropped, actionTestId still reaches the button at :56. No new user-facing string, no critical-path file, no security-model surface.

CI is not green at this head: validate fails on unsigned commits.

gbarkhatov (dismissed):

jonybur (dismissed):
This addresses the earlier round. The caption and the shared value row both clip now, so the unbounded strings that reach this card - the USD total and the liquidation order - stay inside the column instead of escaping over the divider once `min-w-0` removes the content floor.

Keeping the ellipsis on an inner span rather than reaching for `truncate` preserves the sub-`xl` wrap that PR #2296 set up, and the added test pins that. The test also asserts `min-w-0` on the inner column, which is the link that actually makes the truncation work.

jonybur (dismissed):
`value` and `caption` are both declared `string` on `PositionStatCard`, so the `title` attributes are safe at every call site, and both `min-w-0` lifts are genuinely needed — the section and the column each carry the `min-width: auto` floor that was keeping the value from giving way.

Gating the clip behind `xl:` keeps the sub-xl wrapping behaviour from #2296 intact, and wrapping `valueNode` content in the `overflow-hidden` row means a custom node is bounded too rather than only plain values.

gbarkhatov (dismissed):

jonybur (commented):
One point on the delta commit.

The commit's stated purpose is "keep the value tooltip on stat cards that render a custom node," and the `title={card.value}` lift to the outer span at `services/vault/src/components/simple/PositionStatCards.tsx` does exactly that. But the "clips custom valueNode content" test in `services/vault/src/components/simple/__tests__/PositionStatCards.test.tsx` only asserts `overflow-hidden` on the value row — it does not assert `toHaveAttribute("title", <full value>)` on that row. The regression this commit fixes therefore has no direct test guard: a future refactor that puts `title` back on the inner span (behind the `??`) would pass all tests and reintroduce the bug. A one-line `expect(valueRow).toHaveAttribute("title", value)` on the valueNode test closes it.

services/vault/src/components/simple/PositionStatCards.tsx
```diff
@@ -37,12 +39,18 @@ function StatSection({ card }: { card: PositionStatCard }) {
           )}
         </div>
 
-        <span className="flex items-center gap-2 text-xl leading-[1.6] tracking-[0.15px] text-accent-primary xl:whitespace-nowrap">
-          {card.valueNode ?? card.value}
+        <span
+          title={card.value}
```
gbarkhatov, line 44: Minor. `title={card.value}` now puts a tooltip on `valueNode` cards too, but it shows `card.value`, not what the node draws. On the Overview collateral card `value` is only `totalCollateralValue` (`OverviewSection.tsx:59`), while its `valueNode` also renders `collateralBtc` (`:60-66`). The BTC figure sits at the end of the row, so it is what gets clipped first, and the tooltip leaves it out.

Fix: pass the full text as `value` there, e.g. `` `${totalCollateralValue} ${collateralBtc}` ``. The three health-factor cards already pass the same text they render, so they're fine.

gbarkhatov (dismissed):
Follow-up to my review at `ab2d280a8a`. The rebase carried the three reviewed commits over unchanged, so the only delta is `097ab4e34` (keep the value tooltip on cards that render a custom node).

| # | Prior finding | Verdict at `097ab4e34` |
|---|---|---|
| T1 | The caption row is nowrap and nothing clips it | **Fixed.** `title={card.caption}` plus `xl:truncate` (`PositionStatCards.tsx:50-53`) |
| T2 | `valueNode` content is not clipped | **Fixed.** `overflow-hidden` on the shared value row (`:44`) |
| T3 | `truncate` forces nowrap at every width | **Fixed.** Now `xl:truncate` (`:46`) |
| T4 | `min-w-0` on the value span does nothing | **Fixed.** Removed |
| T5 | Say on the line why both `min-w-0` are needed | **Fixed.** Comment at `:25-26` |
| T6 | The `column` query finds the section, and the inner `min-w-0` has no test | **Fixed.** Queries are separated (test `:76-80`) |
| T7 | `(#2428)` belongs in a comment, not the `it()` name | **Fixed** |

One new minor point, inline: the new tooltip on the Overview collateral card leaves out the BTC figure.

It needs a matching doc change outside the diff. The JSDoc on `valueNode` at `PositionStatCards.tsx:10-11` still says that when `valueNode` is set, "`value` is not rendered at all". Since `097ab4e34`, `value` is the tooltip for those cards, so the comment should say that.

Test gap: reverting `097ab4e34` fails the plain-value test at `:74`, but no test checks `title` on a `valueNode` card (`:97-113`). The behaviour this commit is named for is therefore untested.

Not a code issue: `validate` (Authenticate Commits) fails on this head. All four commits are unsigned (`git log %G?` = `N`).

gbarkhatov (approved):
Follow-up review. Changes since my last review (`097ab4e3` → `d99dc033`).

**Headline:** all 8 prior threads are fixed. After the squash and rebase, three things changed:
- the T8 fix in `OverviewSection.tsx:59` (`` value: `${totalCollateralValue} ${collateralBtc}` ``);
- a clearer `valueNode` JSDoc: `value` still supplies the row `title`, so it must carry the same text the node draws;
- a new Overview test pinning the collateral title.

The commit is signed and `validate` passes. CI is green, including `verify`. No merge-blockers. No new defects.

### Status of prior threads

| # | Finding | Status at `d99dc033` |
|---|---|---|
| T1 | Caption row nowrap with nothing clipping it | Fixed (`title={card.caption}` + `xl:truncate`) |
| T2 | `valueNode` content not clipped | Fixed (`overflow-hidden` on the shared value row) |
| T3 | `truncate` forced nowrap at every width | Fixed (`xl:truncate`, row `title`) |
| T4 | `min-w-0` on the value span did nothing | Fixed (removed) |
| T5 | Say on the line why both `min-w-0` are needed | Fixed |
| T6 | `column` query found the section, so the inner `min-w-0` was untested | Fixed |
| T7 | `(#2428)` in the `it()` name | Fixed |
| T8 | Overview collateral tooltip left out the BTC figure | Fixed in `d99dc033`, covered by `OverviewSection.test.tsx:81-86` |

### Verified clean

- The other three `valueNode` callers already meet the new JSDoc contract: `LoansSummary`, `PositionOverview` and `VaultsSummaryCard` each pass `value: healthFactorText` and draw that same text plus an icon.
- The new title test fails if the Overview value change is reverted, and also if the row `title` is dropped.
- These are native `title` attributes: no popper, no animation, and no new copy literals (it joins two already-formatted values).
- Another reviewer's open point still stands: the "below xl, a long value wraps" test would pass even with the change reverted. I'm not repeating it here.

PR #2444, opened 2026-09-08
fix(vault): keep pending deposits under the Deposit activity filter

### LINKED ISSUE #2426: Activity: filtering by Deposit hides pending deposits that are labelled Deposit
## Context

On the Activity page, filtering by Deposit hides rows that are labelled "Deposit".

## Changes

**Now:** a pending deposit row carries `type: "Pending Deposit"` (`services/vault/src/types/activityLog.ts:10`, `PENDING_DEPOSIT_TYPE`), which is a distinct `ActivityType` from `"Deposit"` — but it is *rendered* with the label "Deposit", because `COPY.activity.typeLabels` maps `"Pending Deposit": "Deposit"` (`services/vault/src/copy.ts:2008-2016`).

The filter is an exact type match:

```ts
visible = activities.filter(row => (filter === null || row.type === filter) && ...)
```

(`services/vault/src/components/Activity/ActivityList.tsx:109-115`)

The filter menu offers only `Deposit | Withdraw | Repay | Borrow | Partially/Fully Liquidated` (`copy.ts:1997-2004`); `"Pending Deposit"` is excluded deliberately per the comment at `ActivityList.tsx:56-58`. So selecting "Deposit" compares against `"Deposit"`, which is false for a pending row, and the row disappears even though it reads "Deposit" on screen.

**Expected:** filtering by Deposit keeps pending deposits, since that is what the row says it is.
**Where:** `ActivityList.tsx:109-115`

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788268916708069)

> Activity filter doesnt work?
> When I am on this screen (pic 1) showing deposits, if I filter by deposits I should still see them

</details>

jeremy-babylonlabs:
## Outcome

Filtering the Activity list by Deposit keeps pending deposits, which already render with the label "Deposit".

## Change

The filter in `ActivityList` treats a `"Pending Deposit"` row as a match for the `"Deposit"` filter. The filter menu options are unchanged.

## Safety

One condition added to the existing filter. Other filters and the search behave as before.

## Verification

| Gate | Result |
|---|---|
| `vitest` for `components/Activity` | 6 files, 61 tests pass |

New test filters a Deposit, a Pending Deposit, and a Borrow row by Deposit and expects both deposit rows.

## Links

Closes #2426.

### DIFF
--- a/services/vault/src/components/Activity/ActivityList.tsx
+++ b/services/vault/src/components/Activity/ActivityList.tsx
@@ -5,7 +5,11 @@ import { twJoin } from "tailwind-merge";
 import { ListRowCard } from "@/components/shared/ListRow";
 import { COPY } from "@/copy";
 import { useMidnightTick } from "@/hooks/useMidnightTick";
-import type { ActivityRow, ActivityType } from "@/types/activityLog";
+import {
+  PENDING_DEPOSIT_TYPE,
+  type ActivityRow,
+  type ActivityType,
+} from "@/types/activityLog";
 import { formatActivityDateGroup } from "@/utils/formatting";
 import { useDebounce } from "@/utils/hooks";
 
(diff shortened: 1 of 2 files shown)


jonybur (approved):
Correct and narrowly scoped. `PENDING_DEPOSIT_TYPE` is the sentinel declared in `types/activityLog.ts` and is deliberately absent from `COPY.activity.filterTypes`, so the filter menu is unchanged and the new clause only widens the existing `Deposit` option. That matches `typeLabels`, which already renders a pending row as "Deposit" — the list and the filter now agree.

Checked that no other filter is affected: the added condition is guarded on `filter === "Deposit"`, and search matching is untouched. The new test exercises all three row kinds and asserts the borrow row is excluded, so it would fail if the clause were widened further. Comment update on `FILTER_OPTIONS` is accurate. CI is green.

Issue #2233, opened 2026-08-12
[eth-first] Prove the Ethereum-first flow before release


jonybur:
## Outcome

Release the Ethereum-first Vault after its browser checks pass, with a tested rollback path.

Status: baseline CI is complete. Preparation can start. jonybur coordinates. The release owner must accept responsibility before enablement.

## Problem

Merged code and local checks do not prove the complete browser journey or a safe rollback.

## Scope

#2232 owns implementation and the disabled control. #2453 owns baseline CI. #2355 owns the agreed consumer session checks. #2354 owns consent adoption. Link their evidence and repeat only checks affected by integration.

The deferral recorded in #2228 applies to #2506 and #2507. These two follow-ups do not block release. They are accepted exceptions to the related completion conditions below.

Use the smallest release plan that proves the changed behavior and permits rollback. The release owner selects the release method and owns rollback, or names the person who will perform it. Separate authority roles, a separate formal rehearsal, fixed internal and limited rollout stages, and control removal are optional. None is a universal completion gate. This replaces the former staged-rollout checklist and #2454-2458.

## Complete when

- [x] Baseline package, browser, and visual checks pass in merged #2463 at `ea328745`. Main merge: `1d8560d17`. Evidence is in #2453.
- [ ] After integration, the real-browser Vault journey passes: Ethereum-only access, Bitcoin prompting, cancel, explicit retry, reload, restore, and existing recovery. Connection starts no pending action. Existing two-wallet flows pass in both apps. Link #2232, #2355, and #2354 evidence instead of repeating their matrices.
- [ ] The release owner approves the revision, initial exposure, checks, stop conditions, and rollback steps before enabling the change. No consent release note exists, so state the consent rule in the approval.
- [ ] The disabled control restores the existing two-wallet behavior on the release revision. Required transaction and recovery checks pass. Existing automated or browser evidence can prove this; a separate rehearsal is not required.
- [ ] All security defects and required security reviews that apply to the release are resolved, including applicable findings in #2231 and #2361.
- [ ] The approved release completes with its checks passing. Record the revision, control state, release result, approval, and evidence links. Apply the stop and rollback rules if a check fails.

A failed required check blocks release advancement. Download-size improvements and later control removal do not block this issue or #2228.

## Dependencies

Epic: #2228. Required behavior evidence: #2232, #2355, and #2354. Reuse completed baseline and package evidence from #2453, #2346, and #2360 where the code is unchanged.

[...]

PR #2445, opened 2026-09-08
feat(wallet-connector)!: separate optional wallet dependencies

### LINKED ISSUE #2360: [eth-first] Check optional wallet dependencies before publish
## Outcome

The published wallet connector supports a clean Ethereum-only install without Bitcoin dependencies.

Decision approved on 2026-09-07. Implementation is in open PR #2445. All five package conditions have branch and [CI evidence](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34230013317/job/102073500519). Required owner approval, review-thread disposition, and merge remain.

## Problem

A package can pass workspace tests and still fail after publish if an optional wallet dependency is missing or declared incorrectly.

This check was planned in #2230 but was not delivered by #2267.

## Scope

- Derive the dependency check from the real build externals and package manifest.
- Test the packed wallet connector outside the workspace.
- Install and use the Ethereum-only entry without Bitcoin dependencies.
- Check static and dynamic imports for Bitcoin dependency leaks.
- Test a clean Bitcoin consumer with its required dependencies.
- Run the checks in normal package verification.

## Complete when

- [ ] Missing and stale optional peers fail the check.
- [ ] Static and dynamic import leaks fail test fixtures.
- [ ] The packed Ethereum-only consumer installs and runs without Bitcoin dependencies.
- [ ] The packed Bitcoin consumer passes a clean install.
- [ ] The check runs in normal package verification.

[Closeout review](https://github.com/babylonlabs-io/babylon-toolkit/pull/2445#pullrequestreview-5142815475) maps the conditions to evidence. Keep this issue open until the reviewed implementation merges with passing checks. Full session and rollout proof remain in #2355 and #2233.

## Dependencies

- Epic: #2228. Parent: #2230.
- This issue blocks #2230 and #2232.
- #2346 covers the SDK package contract. This issue covers the wallet connector.
- The package contract is approved. No further product decision is needed to start.

jonybur:
## Outcome

The packed `./eth` entry installs and runs without Bitcoin wallet dependencies. Both native module formats are checked. Local checks pass.

## Change

- Check emitted imports, declarations, and bundled dependencies against the build settings and package manifest.
- Reject missing or stale optional peers and static, dynamic, or bundled Bitcoin leaks into the Ethereum entry.
- Keep Bitcoin deep imports bundled. Load both packed entries as native ESM and CommonJS before browser checks.
- Emit CommonJS files as `.cjs` in wallet connector and Core UI. Check both public runtime exports in each format.
- Require Bitcoin validation before selecting or storing a wallet, including restored connections.
- Run isolated ETH and BTC consumers in normal CI. Run the boundary self-tests during build and release.
- Declare all six optional wallet peers in both existing apps. Remove the unused direct QR dependency.

## Safety

**Breaking install change:** Apps that import the package root must install all six optional peers. These packages are no longer installed automatically. Both existing apps now declare them.

The consumers use explicit compatible versions. They pin `@reown/appkit-adapter-wagmi>@wagmi/connectors` to `6.0.1` and use `zod@3.22.4`. They use Starknet 7 because `@keplr-wallet/types@0.12.272` requires it. `@cosmjs/proto-signing` remains an optional peer because the published Cosmos configuration exposes its `Registry` type.

Consumer type checks use `skipLibCheck=true`. A clean packed ETH check with this option disabled reports 272 upstream diagnostics, including 209 in Wagmi and 22 in Starknet. None point to the connector, Core UI, or the consumer fixture. The package check separately validates declaration imports. This does not prove strict compatibility with all upstream declarations.

This task checks browser consumers and native ESM and CommonJS module loading. The packed checks use fresh wallet connector and Core UI tarballs. Both packages must be released together. The release driver pins wallet connector to the exact new Core UI version, and Nx publishes Core UI first. A wallet-only release does not supply this fix. SSR rendering is not tested. Full session and rollout proof remains in #2355 and #2233.

## Verification

- 295 wallet connector tests, 113 Core UI tests, and 23 package tests pass through Nx.
- Both package builds pass. Both clean packed consumers install, load native ESM and CommonJS exports, and render their providers in Chromium.
- Lint passes with 0 errors. Wallet connector reports 89 existing warnings. Core UI reports 213 existing warnings.
- A frozen offline install and dependency version checks pass.
- The emitted graph rejects injected bundled Bitcoin code. Existing fixtures reject missing or stale peers and static or dynamic imports.

The remaining review decisions concern upstream declaration limits, an optional emitted-file count check, and the required Cosmos declaration peer. Remaining before merge: passing final CI, review-thread disposition, and required owner approval.

## Links

Closes #2360.

Parent: #2230. Epic: #2228. Related: #2232. Follow-up CI task: #2453.


jonybur (commented):
Review of 6b516ba9dc4a14800b4d3e6f2d8fcacee910253d found no confirmed code blocker against the current #2360 package contract.

- The package checks reject missing or stale optional peers and static or dynamic Bitcoin import leaks.
- Isolated packed Ethereum and Bitcoin consumers pass. Ethereum runs without Bitcoin peers. [CI evidence](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34230013317/job/102073500519).
- The app-manifest concern in [the existing thread](https://github.com/babylonlabs-io/babylon-toolkit/pull/2445#discussion_r3958571366) does not establish a regression in the supported deployment. [Vault](https://github.com/babylonlabs-io/babylon-toolkit/blob/6b516ba9dc4a14800b4d3e6f2d8fcacee910253d/.github/workflows/service-release-vault.yml#L161) and [simple staking](https://github.com/babylonlabs-io/babylon-toolkit/blob/6b516ba9dc4a14800b4d3e6f2d8fcacee910253d/.github/workflows/service-release-simple-staking.yml#L105) install the full workspace, bundle the apps, and publish static output. The connector development dependencies supply these peers during the build. Published root-entry consumers must install the declared peers, as the breaking contract and packed Bitcoin test require.

Completion target: all five conditions in #2360. Remaining: record the disposition of open review threads, obtain required owner approval, retain passing checks on the final revision, and merge. #2360 stays open until then. Broader session and rollout proof remains in #2355 and #2233.

This comment records the review and evidence. It is not an approval.

packages/babylon-wallet-connector/vite.config.ts
```diff
@@ -37,29 +39,9 @@ export default defineConfig({
...
-        // @reown packages that use viem internally
-        "@reown/appkit",
-        "@reown/appkit-adapter-wagmi",
-        "@reown/appkit-adapter-bitcoin",
-        /^@reown\//, // Match all @reown/* packages
-        // React Query must be external to share context with consuming app
-        "@tanstack/react-query",
-      ],
+      external: (id) =>
+        [...requiredExternals, ...walletExternals].some((name) => id === name || id.startsWith(`${name}/`)) ||
```
gbarkhatov: **This is the one I'd fix before merge.** The `external` list changed from a literal array to a predicate, and that silently changed the matching semantics from **exact** to **prefix**.

The old array matched ids exactly — which is precisely why `react/jsx-runtime` had to be listed separately alongside `react`. The new predicate matches `id === name || id.startsWith(`${name}/`)`, so every deep subpath of all fourteen names is now externalised too.

That matters because three source sites import deep paths into `bitcoinjs-lib`:
- `src/core/utils/wallet.ts:3` — `bitcoinjs-lib/src/psbt/bip371`
- `src/core/wallets/btc/keystone/provider.ts:5-7` — `bip341`, `bip371`, `psbtutils`

Those were previously **bundled** (exact match failed). I confirmed against the pre-PR `dist/index.es.js`: the only emitted specifier is the bare `"bitcoinjs-lib"`, no `bitcoinjs-lib/src/*` anywhere. After this change they become external, so the published ESM emits extensionless deep paths into `bitcoinjs-lib@6.1.7` — a package with **no `exports` map and no `"type"` field**. Node's ESM resolver rejects those with `ERR_MODULE_NOT_FOUND`. Bundlers resolve them happily, which is exactly why every in-repo app and all of CI stays green.

Fix: keep exact-match for `bitcoinjs-lib` (or list the deep specifiers explicitly). The Node load check in my comment on `check-packed-consumer.js:154` is what would have caught this automatically.

Separately, and while you're in here: the removed array carried load-bearing rationale that went with it — *"React Query must be external to share context with consuming app"*, *"@reown packages that use viem internally"*, and *"Issues linking with Next.js"* on the commented-out `@keystonehq/keystone-sdk`. `requiredExternals` now reads as a flat, fungible list, so someone trimming `@tanstack/react-query` to shrink the bundle would break context sharing with the host app with no compile error and no failing test. Worth carrying those comments over to `package-boundary.js`.

packages/babylon-wallet-connector/package.json
```diff
@@ -52,57 +52,66 @@
...
+  "nx": {
+    "targets": {
+      "check:packed-consumer": {
+        "dependsOn": [
+          "build"
+        ]
+      }
+    }
+  },
   "peerDependencies": {
```
gbarkhatov, line 79: The breaking install contract isn't satisfied by this repo's own root-entry consumers.

pnpm does not auto-install *optional* peers, and I checked both service manifests:

- **`services/vault`** — has `bitcoinjs-lib`, and `@babylonlabs-io/ledger-vault-signer` only as a **devDependency**. Missing `@cosmjs/proto-signing`, `@cosmjs/stargate`, `@keystonehq/sdk`, `@reown/appkit-adapter-bitcoin`.
- **`services/simple-staking`** — has both `@cosmjs/*` and `bitcoinjs-lib`. Missing `@babylonlabs-io/ledger-vault-signer`, `@keystonehq/sdk`, `@reown/appkit-adapter-bitcoin`.

Both import the package **root**, so all six apply. They resolve today only because resolution walks up into `packages/babylon-wallet-connector/node_modules` — where this PR just moved those names to `devDependencies`. That makes wallet-connector's *devDependencies* load-bearing for vault's runtime, and nothing pins them in the consumers' own lockfile importers. It's a phantom dependency: it survives the hoisted monorepo layout and dies on a registry install or a `--prod` build.

Not breaking anything today (there's no such install path in-repo), which is why CI is green. But this PR is the change that creates the obligation, so adding the six to both service manifests here would make the repo's own apps prove the contract the package now declares.

packages/babylon-wallet-connector/scripts/check-packed-consumer.js
```diff
@@ -0,0 +1,268 @@
...
+`,
+    );
+    run(
+      process.execPath,
+      [
+        resolve(consumerRoot, "node_modules/typescript/bin/tsc"),
+        "--noEmit",
+        "--strict",
+        "--skipLibCheck",
+        "true",
```
gbarkhatov: `--skipLibCheck true` here is the direct opposite of the sibling this script mirrors. `packages/babylon-ts-sdk/scripts/check-packed-consumer.js:165-166` passes `"--skipLibCheck", "false"` — deliberately, so that the packed `.d.ts` closure is proven to type-resolve *without* the optional Bitcoin peers installed.

With `true`, `tsc` never opens `dist/eth.d.ts` or anything it references. A declaration-level leak reaching an ETH-only consumer — an `import("bitcoinjs-lib").Psbt` surfacing in an exported type, say — is then caught only by the hand-rolled `declarations.eth` scanner in `check-package.js`, never by the compiler that consumers will actually run.

The PR description lists the relaxation as a fact but doesn't say what forces it. If a third-party `.d.ts` genuinely can't pass under `false`, could you name that package in a comment and scope the relaxation to `btc` only? The ETH mode is the one where this check earns its keep.

packages/babylon-wallet-connector/scripts/package-boundary.js
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+  const declared = {
+    ...manifest.dependencies,
+    ...manifest.optionalDependencies,
+    ...manifest.peerDependencies,
+    ...manifest.devDependencies,
+  };
+  const buildOnlyDependencies = [...(runtime.root.bundledPackages ?? [])].filter(
+    (name) => !runtime.eth.bundledPackages?.has(name) && !required.has(name) && declared[name],
```
gbarkhatov: `bundledPackages` is computed for both graphs at `check-package.js:14`, but the only thing that ever reads `runtime.eth.bundledPackages` is this filter — where it *subtracts* names from the reported list. Nothing anywhere bans a Bitcoin package from being bundled **into** the ETH chunk.

That gap is load-bearing now: `@bitcoin-js/tiny-secp256k1-asmjs`, `@scure/bip32`, `@scure/btc-signer` and `bip174` are no longer in the external list, so they're inlined rather than imported — which means the specifier scan never sees them at all. The gate can print "Ethereum imports none" over a bundle that contains Bitcoin cryptography. The install boundary still holds; "no Bitcoin code in the ETH entry" is the part that's unproven.

Worth adding an explicit ban that iterates `runtime.eth.bundledPackages` and throws on any Bitcoin package, mirroring the sibling's `dependencyCleanBanned`.

Minor, same line: `[...(runtime.root.bundledPackages ?? [])]` means that if the field is ever absent, `buildOnlyDependencies` comes out empty and the loop below never fires — a silent pass rather than an error (CLAUDE.md "No Silent Fallbacks on Critical Paths"). `check-package.js:14` always populates it today, so this is defensive code hiding a future regression rather than a live bug. Requiring the field and throwing when it's missing would be stricter, with the fixtures passing `new Set()` explicitly.

packages/babylon-wallet-connector/src/hooks/useWalletConnectors.tsx
```diff
@@ -118,12 +123,13 @@ export function useWalletConnectors({ persistent, accountStorage, onError }: Pro
 
           if (!visible) return;
 
-          validateAddress(connector.config.network, connectedWallet.account.address);
+          if (!btcValidation) throw new Error("Bitcoin address validation is unavailable");
```
gbarkhatov: This guard fires after the wallet has already been selected and persisted. Just above, `selectWallet?.("BTC", connectedWallet)` and `accountStorage.set(connector.id, connectedWallet.id)` have both already run.

The `catch` further down calls `connector.disconnect()` and `removeWallet?.(connector.id)` but never `accountStorage.delete(connector.id)` — so with `persistent`, a host that forgot the validator leaves a BTC wallet in storage and auto-reconnects it on the next load. And unlike a bad address, a missing `btcValidation` is a permanent configuration fact, knowable before any wallet event ever arrives, so this repeats every single time.

Compounding it, `if (!visible) return` on line 124 short-circuits *above* this guard — so the silent reconnect path selects and persists a BTC wallet with no address validation at all. (That ordering predates this PR, but it sits in the block you're editing and it undercuts "keep Bitcoin address validation in the root provider".)

Suggest hoisting the guard above the `selectWallet` call — a host that cannot validate should never select or persist a BTC wallet — or better, not registering the `BTC` handler at all when `btcValidation` is absent. Also worth raising it as a `WalletError` with an `ERROR_CODES` entry: this bare message reaches the user verbatim as the "Connection Failed" body, where "Bitcoin address validation is unavailable" names neither the missing peer nor the remedy.

packages/babylon-wallet-connector/scripts/check-packed-consumer.js
```diff
@@ -0,0 +1,268 @@
...
+      if (mode === "eth") {
+        assert(
+          !installedPackages.some((entry) => entry.startsWith(`${name.replaceAll("/", "+")}@`)),
+          `ETH installed ${name}`,
+        );
+      }
+    }
+    const specifier = `${manifest.name}${mode === "eth" ? "/eth" : ""}`;
+    const exports = manifest.exports[mode === "eth" ? "./eth" : "."];
+    for (const condition of ["types", "import", "require"]) {
```
gbarkhatov: This only `existsSync`-checks the `types` / `import` / `require` targets — the package is never actually loaded in Node. The only Node process this script runs is the `check-optional-peers.cjs` fixture, which resolves the *peers*, not the package.

The sibling does load it, both ways: `packages/babylon-ts-sdk/scripts/check-packed-consumer.js:184-196` runs `check-esm.mjs` (`import * as sdk from …`) and `check-cjs.cjs` (`require(…)`) against the real entry. So here `dist/eth.cjs.js` is verified to exist and never to run, which is why the packed-CJS half of the issue's "ETH-only consumer installs and runs" criterion isn't actually proven — the Vite consumer app is ESM-only.

This is also the check that would have caught the `external` widening I flagged on `vite.config.ts:43`: a `node -e 'import("<pkg>")'` on the root entry is precisely what surfaces the `ERR_MODULE_NOT_FOUND` from the newly-externalised deep `bitcoinjs-lib` subpaths. Adding an ESM import and a CJS `require` smoke for both entries would close both holes at once.

packages/babylon-wallet-connector/scripts/package-boundary.js
```diff
@@ -0,0 +1,143 @@
...
+    for (const source of map.sources) {
+      const parts = source.split("node_modules/");
+      if (parts.length > 1) packages.add(packageName(parts.at(-1)));
+    }
+  }
+  return packages;
+}
+
+export function checkPackage(manifest, runtime, declarations) {
+  const required = new Set([...runtime.eth.packages, ...declarations.eth.packages]);
```
gbarkhatov, line 123: Context first, because an earlier review pass got this wrong and I don't want it costing you time: the claim that a collapsed `./eth` closure passes this gate **vacuously** is false, and I verified that by executing the real `checkPackage` against the real `dist` graphs. Both collapse scenarios throw. Line 105 sweeps any `used` package that isn't in `required` into `optionalPeers`, and line 110 then demands each be a declared optional peer — and `declarations.eth.packages` really does contain `@keplr-wallet/types` and `nanoevents`, both plain `dependencies`, so a collapse trips that rule. Zero inputs does **not** equal pass.

The real point is much smaller: the collapse is caught only *incidentally*, and the error names the wrong thing. My probe failed with `Missing optional wallet peer: ws` — which tells a future maintainer nothing about the actual cause, that the emitted walk stopped reaching code.

The sibling floors this explicitly at `check-eth-build.js:336-347`, with the reasoning stated in a comment: *"Finding no violation is only meaningful if the walk actually reached the code."* A named floor on `runtime.eth.files.size` would make the failure say what actually broke. Low priority — the property is already defended, just not legibly.

(`src/__tests__/eth.entry.test.ts:166` does floor the **source** graph, but that walks `src/`, not the emitted `dist/` artifact this gate ships.)

packages/babylon-wallet-connector/package.json
```diff
@@ -52,57 +52,66 @@
...
     "@changesets/cli": "^2.29.8",
     "@chromatic-com/storybook": "^3.2.2",
     "@commitlint/cli": "^19.8.0",
     "@commitlint/config-conventional": "^19.8.0",
+    "@cosmjs/proto-signing": "0.36.0",
+    "@cosmjs/stargate": "0.36.0",
     "@eslint/js": "^9.31.0",
     "@internal/eslint-config": "workspace:*",
+    "@keplr-wallet/provider-extension": "0.12.272",
+    "@keystonehq/animated-qr": "0.10.0",
```
gbarkhatov: `@keystonehq/animated-qr` has zero importers anywhere under `src/`. It was in the pre-PR `external` list, and the only mention left in the tree is the `BITCOIN_WALLET_PACKAGES` literal in `src/__tests__/eth.entry.test.ts:17` — i.e. it's named only by the test that asserts it *isn't* reachable.

It was carried over into `devDependencies` rather than dropped. ZERO DEAD CODE POLICY says delete it; this PR is already the one auditing exactly this surface, so it's a natural place to do it.

packages/babylon-wallet-connector/scripts/check-packed-consumer.js
```diff
@@ -0,0 +1,268 @@
...
+      join(consumerRoot, "package.json"),
+      JSON.stringify({
+        name: `wallet-connector-packed-${mode}-consumer`,
+        private: true,
+        type: "module",
+        packageManager,
+        dependencies: {
+          [sourceManifest.name]: tarballs[sourceManifest.name],
+          // Keplr types and the Core UI Tailwind helper require these application peers.
+          starknet: "7.6.4",
```
gbarkhatov, line 97: `starknet: "7.6.4"` pins a version nothing in this workspace uses — `pnpm-lock.yaml` resolves `starknet@6.24.1` throughout. So the gate certifies a dependency graph no shipped app has, and because neither side references the other it will drift silently.

(To be clear, the `zod: "3.22.4"` pin two lines down is **not** an instance of this — I checked, and the lock genuinely contains 3.22.4 alongside 4.1.12. That one's fine.)

Deriving these from the workspace-resolved versions the way `peersFor` already does for the rest would keep the fixture honest. Failing that, a comment saying why the fixture must diverge would at least make the drift intentional. Same line of thinking for `15_000` and the other inline version literals here — CLAUDE.md No Magic Numbers asks for named constants with the reason each value was chosen, and here the naming would also make the divergence visible.

packages/babylon-wallet-connector/package.json
```diff
@@ -52,57 +52,66 @@
...
+      "check:packed-consumer": {
+        "dependsOn": [
+          "build"
+        ]
+      }
+    }
+  },
   "peerDependencies": {
+    "@babylonlabs-io/ledger-vault-signer": "workspace:*",
+    "@cosmjs/proto-signing": "0.36.0",
```
gbarkhatov, line 81: `@cosmjs/proto-signing` is a **type-only** import promoted to a runtime optional peer. `src/providers/CosmosWalletProvider.tsx` uses `OfflineSigner` and `Registry` in type positions only (`registry?: Registry`).

It isn't in `walletExternals` — it reaches `optionalPeers` via the generic sweep at `package-boundary.js:105`, because it appears in the declaration graph but not the ETH graph. Net effect: every root consumer is now instructed to install a runtime package it needs only for types.

Switching both usages to `import type` would drop it out of the emitted graph entirely and take it off the consumer's install list.

packages/babylon-wallet-connector/package.json
```diff
@@ -52,57 +52,66 @@
     "test:unit:ui": "playwright test --ui --project=unit",
     "test:unit:debug": "playwright test --debug --project=unit",
     "test:all": "npm run test:unit && npm run test:e2e",
-    "test": "vitest run",
-    "visualizer": "pnpm exec vite-bundle-visualizer"
+    "test": "node --test scripts/package-boundary.test.js && vitest run",
```
gbarkhatov, line 55: The boundary self-tests live in `test`, but `release` is `npm run build && changeset publish` — and `build` runs only `check-package.js`, not these. So an `importSpecifiers` that regressed (stopped seeing `require()` or dynamic `import()`, say) would produce a passing build and a published package whose boundary was never meaningfully checked.

The sibling avoids this by embedding its specifier self-tests *inside* `check-eth-build.js`, so `build` itself fails when the parser rots.

CI `verify` does run `test`, so this only bites a release performed outside the normal CI path — low severity. Running `node --test scripts/package-boundary.test.js` from `build`, or adding it to `release`, would close it.

packages/babylon-wallet-connector/scripts/package-boundary.test.js
```diff
@@ -0,0 +1,209 @@
...
+  };
+  const declarations = {
+    root: emittedDependencies([join(root, "root.d.ts")]),
+    eth: emittedDependencies([join(root, "eth.d.ts")]),
+  };
+  return (candidate = manifest) => checkPackage(candidate, runtime, declarations);
+}
+
+test("accepts optional peers derived from the build externals and declarations", (t) => {
+  assert.deepEqual(fixture(t)().optionalPeers, [...walletExternals, "@cosmjs/proto-signing"].sort());
```
gbarkhatov: This asserts the result against `walletExternals` — the same production constant `checkPackage` reads to compute it. A wrong or missing entry in that constant moves both sides of the assertion together, so this test keeps passing.

Given it's the acceptance test for the exact set of peers the package now demands consumers install, spelling the six names out as literals would make it actually pin something. CLAUDE.md test philosophy points the same way: *"Prefer explicit values over magic constants imported from elsewhere."*

gbarkhatov (dismissed):
Reviewed with three independent passes, then deduped and verified each finding against the code myself — including executing the real `checkPackage` over the real `dist` graphs. Approving; the findings below are suggestions and yours to triage.

The gate itself is good work, and in two respects it is **stronger** than the ts-sdk sibling it mirrors: the leak scanner is a TypeScript AST walk rather than the sibling's regex + `stripComments` (it catches template-literal specifiers, `export *`, side-effect-only imports and `import type`, and *throws* on computed specifiers instead of ignoring them), and the ETH-mode `node_modules/.pnpm` scan proves the optional peers aren't pulled in transitively, not merely that they're unresolvable. CI reachability is real: `verify.yml:95-96` is unconditional, uses `nx run` not `nx affected`, and `check:packed-consumer` is absent from `nx.json` `targetDefaults` so it is uncached.

**The one I'd genuinely like you to look at before merge is the first comment** — the `external` list changing from exact-match to prefix-match. It is the only finding here I'd call a shipping bug, it isn't mentioned in the PR description, and no test covers it. Everything else is hardening or tidy-up.

**One correction to save you time.** An earlier pass flagged, as a merge-blocker, that a collapsed emitted `./eth` entry passes the gate vacuously. **That is wrong and I've refuted it** — I ran the real `checkPackage` against the real `dist` graphs and simulated both collapse scenarios; both throw. `package-boundary.js:105` sweeps any `used` package that isn't in `required` into `optionalPeers`, and line 110 then demands it be a declared optional peer — and `declarations.eth.packages` genuinely contains `@keplr-wallet/types` and `nanoevents`, both plain `dependencies`, so a collapse trips that rule. The real (much smaller) point is that the failure is incidental and reports a misleading name; that's the comment on `package-boundary.js:101`. Also refuted: `workspace:*` in `peerDependencies` is fine (it matches the ts-sdk house pattern and `nx.json`'s `preserveLocalDependencyProtocols: false` rewrites it at release), and `zod@3.22.4` is *not* workspace drift — the lock contains 3.22.4 alongside 4.1.12.

On the issue's acceptance criteria: 4 of 5 are proven by real fixtures. The packed-**CJS** half of "the ETH-only consumer installs and runs" is not — that entry is only checked to exist, never loaded.

CLAUDE.md section 9 does **not** bind: `validateAddress` / `validateAddressWithPK` in `src/core/utils/wallet.ts` are unchanged and still backed by `bitcoinjs-lib`; only the import site moved. Nothing was reimplemented, so no differential test is owed.

One minor note with no diff line to hang it on: `src/__tests__/eth.entry.test.ts:166` uses a bare `expect(ethGraph.files.size).toBeGreaterThan(30)` — the sibling names its equivalent floors as constants and documents how they were measured.

packages/babylon-wallet-connector/scripts/check-packed-consumer.js
jonybur: The clean packed ETH check with skipLibCheck disabled reports 272 upstream diagnostics, including 209 in Wagmi and 22 in Starknet. None point to the connector, Core UI, or the consumer fixture. The relaxation therefore also applies to ETH. The PR records this limit, and the script now names upstream declaration failures. The independent declaration import scan still runs.

packages/babylon-wallet-connector/scripts/check-packed-consumer.js
jonybur: Both packed consumers now load ESM and CommonJS and check the WalletProvider and createWalletConfig exports. The check exposed a filename defect in the connector and Core UI: .cjs.js files were treated as ES modules. Both packages now emit .cjs files, and their manifest targets match. The packed check uses fresh tarballs for both packages. The connector release must include the fixed Core UI version. SSR rendering is not tested.

packages/babylon-wallet-connector/scripts/package-boundary.js
jonybur: Agreed that a direct error would be clearer. The current emitted ETH graph has four files. As your corrected probe shows, a collapsed graph already fails the package check. I propose keeping an explicit graph floor and its diagnostic as a separate improvement. This PR retains the existing collapse protection.

packages/babylon-wallet-connector/package.json
jonybur: CosmosWalletConfig exposes Registry in its public registry property. An import type change still emits the @cosmjs/proto-signing reference in the .d.ts file. The package check intentionally includes declaration dependencies, so that change does not remove the peer obligation. The emitted declaration confirms this reference. Removing the peer would leave the published root types unresolved.

gbarkhatov (approved):

PR #2307, opened 2026-08-24
feat(packages): load the Bitcoin engine only when it is needed

jonybur:
## Result

Merged as part 1 of #2231. #2231 stays open.

This release shipped as Bitcoin engine 0.17.0 and SDK 0.71.0.

## Change

- The Bitcoin engine loads on first use.
- One SDK facade owns normal engine access.
- Raw engine classes moved to an explicit subpath.
- Existing derivation values and generated engine bytes stayed unchanged.
- Build checks protect the lazy import boundary.

## Post-merge audit

- #2356: a failed browser engine download cannot retry without reload.
- #2361: the published raw loader lets callers bypass SDK value checks.
- #2333: the browser entry and some facade values lack direct tests.
- #2345: small review cleanup remains.
- #2358: the safer per-call connector has a low-priority performance cost.

## Verification at merge

Workspace build, lint, and unit checks passed. Those checks did not run the real browser engine entry.

Tracks #2231. #2231 stays open.

PR #2446, opened 2026-09-08
fix(wasm): explain lazy build check failures

jonybur:
## Outcome

The lazy-build check explains missing declarations and eager import failures. It rejects a build output directory that differs from the published package layout. Local checks pass.

## Change

Read JSONC and inherited settings with TypeScript. Require the resolved output directory to match `dist`. Give missing emitted files, removed generated imports, and unresolved declarations separate messages. Keep the original file-read errors as causes.

## Safety

An eager facade import still fails the check. Only the first WASM call may report missing generated code. A relocated build fails before a temporary copy can hide missing or stale published files. Declaration errors include the import path and explain the required sibling layout.

## Verification

- Nx build, lint, and all 10 existing tests pass.
- Eleven temporary checks against real build files pass. They cover normal builds, JSONC, inherited output, absent output, relocated output with stale or missing files, malformed or missing configuration, missing emitted declarations, removed generated imports, and unresolved declaration paths.

## Links

Tracks #2345. #2345 stays open.

This covers the diagnostic error item accepted in [the #2307 review](https://github.com/babylonlabs-io/babylon-toolkit/pull/2307#discussion_r3865457140). Epic: #2228.

### DIFF
--- a/packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
+++ b/packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
@@ -2,15 +2,31 @@ import { cpSync, readFileSync, rmSync, mkdtempSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { dirname, extname, join, resolve } from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
+import ts from 'typescript';
 
 const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..');
 const packageManifest = JSON.parse(
   readFileSync(resolve(packageRoot, 'package.json'), 'utf8'),
 );
-const { compilerOptions } = JSON.parse(
-  readFileSync(resolve(packageRoot, 'tsconfig.lib.json'), 'utf8'),
+function failConfig(diagnostic) {
+  throw new Error(
+    `Cannot read tsconfig.lib.json or its extends chain: ${ts.flattenDiagnosticMessageText(diagnostic.messageText, '\n')}`,
+  );
+}
+const config = ts.getParsedCommandLineOfConfigFile(
+  resolve(packageRoot, 'tsconfig.lib.json'),
+  undefined,
+  { ...ts.sys, onUnRecoverableConfigFileDiagnostic: failConfig },
 );
-const outputDirectory = resolve(packageRoot, compilerOptions.outDir);
+const configError =
+  config.options.configFile.parseDiagnostics[0] ?? config.errors[0];
+if (configError) failConfig(configError);
+const outputDirectory = config.options.outDir;
+if (outputDirectory !== resolve(packageRoot, 'dist')) {
+  throw new Error(
+    'tsconfig.lib.json and its extends chain must resolve compilerOptions.outDir to dist to match package exports',
+  );
+}
 if (
   JSON.stringify(Object.keys(packageManifest.exports).sort()) !==
   JSON.stringify(['.', './raw'])
@@ -128,20 +144,31 @@ for (const loaderName of ['wasm-loader.ts', 'wasm-loader-node.ts']) {
   }
 }
 
-// Each emitted loader must resolve its generated type import.
-// Use the configured output directory to check changes to the build layout.
+// Check generated imports that TypeScript and skipLibCheck consumers can miss.
 for (const loaderName of ['wasm-loader.d.ts', 'wasm-loader-node.d.ts']) {
   const emitted = resolve(outputDirectory, loaderName);
+  let declaration;
   try {
-    const match = readFileSync(emitted, 'utf8').match(
-      /from ['"]([^'"]*generated\/vault_wasm\.js)['"]/,
+    declaration = readFileSync(emitted, 'utf8');
+  } catch (cause) {
+    throw new Error(
+      `Missing dist/${loaderName}. Check compilerOptions.outDir and rebuild the package.`,
+      { cause },
     );
-    if (!match) throw new Error(`${loaderName} must import generated types`);
+  }
+  const match = declaration.match(
+    /from ['"]([^'"]*generated\/vault_wasm\.js)['"]/,
+  );
+  if (!match) {
+    throw new Error(
+      `${loaderName} no longer pins its bindings to the generated declarations`,
+    );
+  }
+  try {
     readFileSync(resolve(dirname(emitted), match[1].replace(/\.js$/, '.d.ts')));
   } catch (cause) {
     throw new Error(
-      `${loaderName} must resolve its generated declarations from ${outputDirectory}. ` +
-        `Check compilerOptions.outDir and the generated import path.`,
+      `${loaderName} emits '${match[1]}', which resolves to no declaration from dist. The emit directory and src must stay siblings one level under the package root.`,
       { cause },
     );
   }
@@ -164,8 +191,7 @@ for (const [rawName, loaderName] of [
   }
 }
 
-// Remove generated code from a copy of the build. Both facade entries must
-// import before the first call fails. Raw entries must fail during import.
+// Without generated code, facade entries must import and raw entries must fail.
 const isolatedPackage = mkdtempSync(join(tmpdir(), 'tbv-wasm-lazy-'));
 try {
   cpSync(
--- a/packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
+++ b/packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
@@ -7,6 +7,10 @@ const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..');
 const packageManifest = JSON.parse(
   readFileSync(resolve(packageRoot, 'package.json'), 'utf8'),
 );
+const { compilerOptions } = JSON.parse(
+  readFileSync(resolve(packageRoot, 'tsconfig.lib.json'), 'utf8'),
+);
+const outputDirectory = resolve(packageRoot, compilerOptions.outDir);
 if (
   JSON.stringify(Object.keys(packageManifest.exports).sort()) !==
   JSON.stringify(['.', './raw'])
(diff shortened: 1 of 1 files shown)


packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
```diff
@@ -124,31 +128,21 @@ for (const loaderName of ['wasm-loader.ts', 'wasm-loader-node.ts']) {
...
-  const match = readFileSync(emitted, 'utf8').match(
-    /from ['"]([^'"]*generated\/vault_wasm\.js)['"]/,
-  );
-  if (!match) {
-    throw new Error(
-      `${loaderName} no longer pins its bindings to the generated declarations`,
-    );
-  }
-  const [, specifier] = match;
+  const emitted = resolve(outputDirectory, loaderName);
```
gbarkhatov, line 149: **1 — the `outDir` rename still produces no crafted message, and now fails later and less legibly than before.**

The accepted item names one scenario: rename `outDir`, and instead of the crafted message you get a raw `ENOENT` on `dist/wasm-loader.d.ts`. Following `outDir` here doesn't fix that, because three literals govern the real layout and none of them follow it: `src/wasm-loader.ts:8` hardcodes `'../dist/generated/vault_wasm.js'`, `scripts/build-wasm.js:27` writes `dist/generated` unconditionally, and `package.json` pins `./dist/*`.

Trace `outDir: "build"`. tsc emits `build/wasm-loader.d.ts`, and per the comment at `src/wasm-loader.ts:5-7` it copies the specifier in verbatim — so the emitted file still says `../dist/generated/vault_wasm.js`. `resolve(dirname(emitted), …)` therefore lands on `<pkg>/dist/generated/vault_wasm.d.ts`, which is committed (`.gitignore` ignores `dist/*.d.ts` but un-ignores `!dist/generated/`). The read at `:140` succeeds and **this check passes**, on an emitted declaration now pointing outside the emit directory at a stale sibling.

The run then dies at `:259-261` reading `…/dist/generated/vault_wasm_bg.wasm` out of the temp copy made at `:256` from a `build/` that has no `generated/` — a bare `ENOENT` on an ephemeral `/var/folders/…/tbv-wasm-race-XXXX/` path, inside a block about connector concurrency. So the build still fails, which is the important half; the diagnostic just moved from a stable meaningful path to a temp one 120 lines away, on a PR whose title is about explaining these failures.

One line fixes it: `packageManifest` is already parsed four lines above `:13`, so assert `outputDirectory === resolve(packageRoot, 'dist')` (or cross-check the `./dist/` prefix in `packageManifest.exports`) and throw a crafted message naming both sides. That covers the rename with one comparison and no new reads.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
```diff
@@ -7,6 +7,10 @@ const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..');
 const packageManifest = JSON.parse(
   readFileSync(resolve(packageRoot, 'package.json'), 'utf8'),
 );
+const { compilerOptions } = JSON.parse(
+  readFileSync(resolve(packageRoot, 'tsconfig.lib.json'), 'utf8'),
+);
+const outputDirectory = resolve(packageRoot, compilerOptions.outDir);
```
gbarkhatov: **2 — this read reintroduces the unguarded-crash class the PR set out to remove, above every check in the file.**

`JSON.parse` then `resolve(packageRoot, compilerOptions.outDir)` with no guard, at module top level. Two realistic edits kill `build` with an error naming nothing about this gate:

- `tsconfig.base.json` declares no `outDir` today, so consolidating it up one level is a plausible cleanup — and leaves `outDir` undefined, so `path.resolve` throws `TypeError [ERR_INVALID_ARG_TYPE]`.
- tsconfig is JSONC. A `//` comment is legal and already house style here — `packages/babylon-core-ui/tsconfig.node.json` carries several, explaining exactly the kind of non-obvious `include` decision this file also has — and plain `JSON.parse` throws a bare `SyntaxError` on it.

Because it's top-level, either one takes down all ten downstream checks rather than one. It's also reading a single leaf of a format defined by an `extends` chain and trusting the result, so it can silently disagree with what tsc resolves.

Per the repo's "No Silent Fallbacks" rule the fix isn't `?? 'dist'` — it's an explicit `if (typeof compilerOptions?.outDir !== 'string') throw new Error(…)` naming the file and the `extends` chain. (Moot if you take comment 1's suggestion and drop the read entirely.)

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
```diff
@@ -124,31 +128,21 @@ for (const loaderName of ['wasm-loader.ts', 'wasm-loader-node.ts']) {
...
-  const [, specifier] = match;
+  const emitted = resolve(outputDirectory, loaderName);
   try {
-    readFileSync(
-      resolve(dirname(emitted), specifier.replace(/\.js$/, '.d.ts')),
+    const match = readFileSync(emitted, 'utf8').match(
+      /from ['"]([^'"]*generated\/vault_wasm\.js)['"]/,
     );
-  } catch {
+    if (!match) throw new Error(`${loaderName} must import generated types`);
```
gbarkhatov: **3 — this throw is caught by the `try` it sits in, so a source regression gets reported as a resolution failure.**

The `try` now spans three distinct exits — the emitted-file read, this `!match` test, and the declaration read — and all three surface as "must resolve its generated declarations from `<outDir>`. Check `compilerOptions.outDir` and the generated import path."

But `!match` means the loader stopped pinning its bindings to the generated declarations. That's a regression in `src/`, and `outDir` has nothing to do with it, and the import path isn't wrong — it's absent. The message written right here is immediately demoted to a `cause`, so the first line a developer reads points at the wrong file and the wrong knob. On a PR about explaining these failures that's a net loss, even though the underlying detail survives in the `[cause]` chain.

Guard only the two `readFileSync` calls and hoist this throw out between them, so each of the three causes — missing emit, pin removed, specifier unresolvable — keeps its own first-line diagnosis. The old text ("no longer pins its bindings to the generated declarations") was also more specific than "must import generated types"; worth keeping.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
```diff
@@ -124,31 +128,21 @@ for (const loaderName of ['wasm-loader.ts', 'wasm-loader-node.ts']) {
...
+    readFileSync(resolve(dirname(emitted), match[1].replace(/\.js$/, '.d.ts')));
+  } catch (cause) {
     throw new Error(
-      `${loaderName} emits '${specifier}', which resolves to no declaration ` +
-        `from dist. The emitted specifier reaches the generated surface only ` +
-        `while the emit directory and src stay siblings one level under the ` +
-        `package root.`,
+      `${loaderName} must resolve its generated declarations from ${outputDirectory}. ` +
```
gbarkhatov: **4 — the message drops the one non-obvious fact it existed to carry, and `src/` still promises this script asserts it.**

The item that prompted this change called out "the message a few lines down that names the sibling-layout precondition" and closed with "the crafted messages are the reason the gate exists rather than a bare assertion." This makes that message reachable and then removes its payload: the offending specifier (`match[1]` is now discarded rather than printed) and the sentence explaining that the emitted specifier reaches the generated surface only while the emit directory and `src` stay siblings one level under the package root.

That sentence was the mechanism — it's why `../dist/generated/…` resolves identically from `src/` and from `dist/`. And `src/wasm-loader.ts:5-7` still tells readers "check-lazy-entries.js asserts it" about exactly that precondition, which the script now states nowhere. "Check `compilerOptions.outDir`" points at a field that, per comment 1, doesn't govern this resolution at all.

The replaced comment at `:131-132` has the same problem one level up: the deleted version recorded *why* a bespoke gate is needed here — tsc is silent, and every consumer inherits `skipLibCheck`, which drops the error inside the emitted declaration. "Each emitted loader must resolve its generated type import" restates the loop. A future reader now has no argument against deleting it. (`:167-168` similarly drops "because their generated import is eager," though that one's milder.)

Suggest restoring the specifier value and the sibling-layout sentence in the message, and the `skipLibCheck` rationale in the comment.

Minor, same line: `${outputDirectory}` is absolute, so this prints a machine-local path into CI logs. Every other message in the file names package-relative artifacts.

gbarkhatov (dismissed):
Approving. The `.catch` half of the accepted item lands cleanly, and one deviation from the literal request is a genuine improvement worth calling out (see below). The `outDir` half is the problem: I traced the rename scenario the item names and it still produces a raw `ENOENT`, just a later and less legible one. Four comments inline, of which the first is the one I'd want fixed before this merges.

**The deviation that is correct, so nobody "fixes" it later:** the item asked for the facade import to move *inside* the `try`, as the raw-entry loop does. Doing that literally would have been a real bug — an eager facade's `ERR_MODULE_NOT_FOUND` names `…/dist/generated/vault_wasm.js`, which the filter at `:197` matches, so it would have been silently swallowed and the gate would have passed on the exact violation it exists to catch. The `.catch` at `:185` is the right shape. Worth a comment saying so.

**Verified clean:** `new Error(msg, { cause })` is valid on Node 24 and `[cause]` chains print for uncaught exceptions, so nothing wrapped is lost — only some top lines mislead. Copying `outputDirectory` into a temp directory literally named `dist` is deliberate and correct, since the copied manifest resolves `exports` against `./dist/*`. No published surface changes and the script still runs only under `build`. Nothing registered under CLAUDE.md §4 or §9 is touched, there is no `VAULT_WASM_COMMIT` bump, and SECURITY.md's build-gate rule (every path throws, none downgraded to a warning) is honoured. The removed `specifier` binding went with its last use, so no dead code.

**Optional, pre-existing, not asking for it here:** the emitted-declaration scan at `:136` matches against raw text while the source walk at `:95` runs `stripComments` first, so JSDoc prose containing the specifier would satisfy the pin check vacuously. Contrived, and it predates this PR — noting it only because the helper is free and those are `+` lines now.

**One question:** the description cites 16 temporary checks including "output directory changes." Comment 1 predicts a flat `outDir` rename passes the declaration gate outright. Which of the 16 covered that, and what error did it actually produce? If one of them caught this, I've got the trace wrong and I'd like to know where.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
```diff
@@ -170,17 +164,15 @@ for (const [rawName, loaderName] of [
...
+// Remove generated code from a copy of the build. Both facade entries must
+// import before the first call fails. Raw entries must fail during import.
 const isolatedPackage = mkdtempSync(join(tmpdir(), 'tbv-wasm-lazy-'));
 try {
   cpSync(
     resolve(packageRoot, 'package.json'),
     join(isolatedPackage, 'package.json'),
   );
-  cpSync(resolve(packageRoot, 'dist'), join(isolatedPackage, 'dist'), {
+  cpSync(outputDirectory, join(isolatedPackage, 'dist'), {
```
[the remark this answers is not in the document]
jonybur: The check now requires the resolved output directory to match the published dist layout. It rejects relocated output before any temporary copy. Regression checks cover stale and missing dist files.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
jonybur: The output layout check now runs before the declaration and runtime checks. A renamed output directory fails with a message that names the TypeScript configuration and the required dist export layout.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
jonybur: The check now uses TypeScript to read JSONC and resolve the extends chain. Configuration errors name tsconfig.lib.json and retain the diagnostic. An absent or different output directory fails the explicit published-layout check. Regression checks cover comments, inherited output, missing output, malformed configuration, and a missing configuration file.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
jonybur: The missing-pin guard now runs outside both file-read handlers. Missing emitted files, removed generated imports, and unresolved declarations each have a separate first-line message. File-read errors keep their original cause.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
jonybur: The unresolved-declaration message again includes the actual specifier and the sibling-layout rule. It names dist rather than a machine-local path. The comment also explains why consumers with skipLibCheck need this separate check.

gbarkhatov (approved):

PR #2459, opened 2026-09-08
feat(ts-sdk, vault): finish the health-factor band review

jonybur:
## Outcome

Health-factor bands use a named liquidation boundary. The split-deposit layout suite runs in CI. Local checks pass. CI and renewed review remain.

## Change

Export the unchanged 1.0 boundary. Cover all five status colors and derive the gradient expectation from the warning threshold. Distinguish the dimensionless health-factor boundary from the per-asset threshold in basis points.

Run the six layout cases through Nx with Vault's own Chromium version. Reject focused tests, disable retries, and retain failure traces and reports.

## Safety

Threshold values and comparison operators do not change. No wallet or transaction behavior changes.

## Verification

- 1,968 SDK unit tests and 7 built-package tests pass.
- 3,759 Vault tests pass. Six existing tests are skipped.
- Six layout cases pass without retries. Dependency builds pass.
- A focused-test probe passes the old gate and fails the revised gate.
- A forced layout failure has no trace with the old zero-retry settings. The revised settings retain screenshots and network data in a trace.
- Full SDK and Vault lint, formatting, YAML parsing, and diff checks pass. API documentation is regenerated. Executable output is unchanged.

## Links

Follow-up to the reviews on #2405 and #2406.

The workflow merges cleanly with #2463. The layout gate saves failure files before later browser tests can replace them.


.github/workflows/verify.yml
```diff
@@ -98,6 +98,21 @@ jobs:
       - name: Test SDK in Chromium
         run: pnpm exec nx run @babylonlabs-io/ts-sdk:test-browser
 
+      # The vault's first browser test on the PR gate. Scoped to the one
+      # project on purpose: the sibling `chromium` project holds the
+      # error-scenario specs, which depend on a developer's local .env and have
+      # never been runnable on a clean runner. Chromium is already installed
+      # above and both packages pin @playwright/test 1.56.1, so this reuses
```
gbarkhatov: **This comment's load-bearing sentence is not true, and it is the entire justification for reusing the browser.**

It says "both packages pin @playwright/test 1.56.1". They don't pin the same package:

- `services/vault/package.json:69` → `"@playwright/test": "1.56.1"`
- `packages/babylon-ts-sdk/package.json:170` → `"playwright": "1.56.1"`

Two different dependency names, which syncpack's single-version policy does not tie together. The install at line 92-93 is scoped `--filter @babylonlabs-io/ts-sdk`, so this new step is running the *vault's* Playwright against a browser fetched for a different package's. It works today only because both resolve the same `playwright-core` and `playwright install` writes to the shared browser cache — bump either one alone and every PR fails on `Executable doesn't exist`, with nothing in the workflow explaining why.

`visual-regression.yml` already does this correctly by installing with `--filter @services/vault`. Mirroring that removes the coupling entirely; failing that, correcting the comment to name the real invariant (a shared `playwright-core` resolution, not a shared pin) at least leaves the next person something true to reason from.

services/vault/package.json
```diff
@@ -21,6 +21,7 @@
     "test:coverage": "vitest --coverage",
     "test:run": "vitest run",
     "test:e2e": "playwright test",
+    "test:layout": "playwright test --project=chromium-god-mode",
```
gbarkhatov: This script now backs a **required** gate, so it inherits two settings from a config that until now was only ever run locally. I checked both at `services/vault/playwright.config.ts:103-104`:

- `forbidOnly: false` — a stray committed `test.only` silently shrinks the gate to one test and still reports green.
- `retries: 2` — a genuinely flaky layout assertion passes on the third attempt.

Both are reasonable defaults for a developer loop and wrong for a blocking gate. Cheapest fix is right here on the script the PR is already adding, leaving the shared config untouched for local use:

```json
"test:layout": "playwright test --project=chromium-god-mode --forbid-only --retries=0"
```

Worth noting sibling #2463 passes `--retries=0` on its vault e2e step, so this would be the house pattern rather than a new one.

services/vault/project.json
```diff
@@ -15,9 +15,20 @@
...
             "command": "./scripts/watch-deps.sh {projectName}"
           },
-          { "command": "pnpm exec vite" }
+          {
+            "command": "pnpm exec vite"
+          }
         ]
       }
+    },
+    "test-layout": {
```
gbarkhatov, line 24: The gate captures no failure evidence.

`cache: true` is declared with neither `inputs` nor `outputs`. Playwright's html reporter and traces land in `playwright-report/` and `test-results/`, so neither is registered as an Nx output, and there is no `if: failure()` upload step in the workflow either.

The practical consequence lands on someone else: when this six-test bounding-box suite goes red on an unrelated PR, the author gets terminal output and nothing more — no report, no trace, no screenshot of the layout that failed. For a suite whose entire value is that it can see layout where jsdom cannot, the failure artifact *is* the finding.

Two small additions: declare `"outputs": ["{projectRoot}/playwright-report", "{projectRoot}/test-results"]` here, and add an `if: failure()` upload step in `verify.yml`. (The missing `inputs` is fine on its own — Nx's `DEFAULT_INPUTS` already covers vault `src/`, and caching is inert on CI anyway with no Nx Cloud token.)

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/healthFactor.ts
```diff
@@ -6,7 +6,7 @@
  *
  * Status thresholds:
  * - no_debt: No active debt (null health factor)
- * - danger: < 1.0 (can be liquidated)
+ * - danger: below HEALTH_FACTOR_LIQUIDATION_THRESHOLD (can be liquidated)
  * - risky: from 1.0 up to and including HEALTH_FACTOR_RISKY_THRESHOLD
```
gbarkhatov: The JSDoc block this PR set out to de-literalise is left half-converted, and the two halves are adjacent lines.

Line 9 now reads `danger: below HEALTH_FACTOR_LIQUIDATION_THRESHOLD`, while this line still reads `risky: from 1.0 up to and including HEALTH_FACTOR_RISKY_THRESHOLD` — the same boundary, named two different ways, two lines apart. Line 5 ("A health factor below 1.0 means the position can be liquidated") is the same literal a third time.

Since `constants.ts:70` *was* converted in this same PR, the two files now describe that boundary inconsistently, which is the opposite of the change's stated goal. One-word edit.

packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts
```diff
@@ -53,9 +53,22 @@ export const AAVE_BASE_CURRENCY_RAY_DECIMALS = 53;
...
+ * Health factor at or above which Aave will not liquidate a position.
+ *
+ * Unlike HEALTH_FACTOR_RISKY_THRESHOLD and HEALTH_FACTOR_WARNING_THRESHOLD,
+ * this one is not ours to tune. Aave expresses the health factor in WAD so that
+ * exactly 1.0 is the liquidation boundary (ISpoke.sol - "healthFactor expressed
+ * in WAD. 1e18 represents a health factor of 1.00"). It is definitional to the
+ * encoding rather than a stored parameter, so there is nothing to read from the
+ * contract.
+ */
+export const HEALTH_FACTOR_LIQUIDATION_THRESHOLD = 1.0;
```
gbarkhatov, line 63: Worth one JSDoc sentence rather than a rename: this namespace now has two things called a "liquidation threshold", and they are different in kind.

- `HEALTH_FACTOR_LIQUIDATION_THRESHOLD = 1.0` — a dimensionless health-factor boundary, definitional to Aave's WAD encoding.
- `liquidationThresholdBps` — Aave's governance-settable, per-asset LT in basis points, genuinely fetched from chain.

Both are `number`, so a swap typechecks and nothing catches it. The name is upstream-faithful and I would not change it — but the JSDoc you already wrote here is the natural place to say "not to be confused with the per-asset `liquidationThresholdBps`, which is fetched", so the distinction survives the next person reading only one of the two.

gbarkhatov (dismissed):
Reviewed at `461e2774` with three independent passes. **No merge blockers.** The four cleanups all do what they claim, and I checked the two claims that would have mattered most if they were wrong.

**"No numeric change" is literally true, operator included.** `constants.ts:66` is `export const HEALTH_FACTOR_LIQUIDATION_THRESHOLD = 1.0` — exactly the literal it replaced — and `healthFactor.ts:45` is still `healthFactor < HEALTH_FACTOR_LIQUIDATION_THRESHOLD`, `<` and not `<=`. HF of exactly 1.0 still classifies `risky`, not `danger`. One call site, no others. I checked the operator specifically because `<` vs `<=` at a liquidation boundary is the one way a rename like this turns into a behavioural change, and it is invisible in a diff that looks like pure naming. Worth adding that the pre-existing band tests already pin `0.99 → danger` and `1.0 → risky` with explicit literals, so a slip would have gone red on its own.

**The protocol-parameter exemption holds, and it is worth recording so nobody re-raises it.** The 1.0 boundary is a property of Aave's WAD encoding (1e18 ≡ HF 1.00), not a stored value, so there is no getter to call. The governance-settable input to the formula — the per-asset liquidation threshold in bps — *is* fetched, and is threaded through as `liquidationThresholdBps`. The JSDoc at `constants.ts:56-65` makes precisely this argument in code and cites `ISpoke.sol`, which is the right place for it.

### The greptile "Test Bypasses Nx" thread is resolved in code — I verified the whole chain

`verify.yml:114` is `pnpm exec nx run @services/vault:test-layout`; `project.json:3` declares the project name `@services/vault`; `project.json:24-32` defines `test-layout`; it runs `pnpm test:layout` → `package.json:24` → `playwright test --project=chromium-god-mode`, whose `testMatch` is `**/deposit-progress-layout.spec.ts`. It is a plain `run:` step with no `continue-on-error`, no `|| true` and no `if:` guard, and `nx:run-commands` propagates the exit code — so it can genuinely fail the job.

Neither failure mode I was watching for is present: the target is not invoked with the old direct `playwright` command, and Nx is not pointed at a target that does not exist. The thread can be closed. What survives are the operational gaps below.

### Checked and dismissed

- **Generated docs stale?** No — `docs/api/integrations/aave.md:2319` carries `const HEALTH_FACTOR_LIQUIDATION_THRESHOLD: 1 = 1.0` and `:2344` the reworded RISKY prose; `index.ts:54` exports it. The staleness gate passed.
- **Unpinned dependency?** No — `services/vault/package.json` gains only the `test:layout` script. Nothing added or loosened; `@playwright/test` is exact at `1.56.1`.
- **Is the new export dead code?** No — consumed at `healthFactor.ts:45`, exported once, no wildcard re-export.
- **Do the three new colour tests just mirror the implementation?** No. `getHealthFactorColor` takes a `HealthFactorStatus`, so no threshold is in play at that seam; the direction (danger→red, warning→amber, safe→green) is an independently stated design decision documented before this PR, and a silent amber/red swap on a liquidation warning is exactly what these now catch. This is the strongest of the four changes.
- **Is Nx caching hiding the gate on CI?** No — no Nx Cloud token and no `.nx/cache` restore, so every run is a cold miss and the gate actually executes.

### On the derived green stop — a real trade-off, not a defect

My passes split on this one, so I looked myself. Production at `railLayout.ts:124-127` is `Math.max(stopPct(liquidationPriceUsd * HEALTH_FACTOR_WARNING_THRESHOLD), min(100, red + GRADIENT_MIN_RAMP_PCT))`, and the test now computes the same first argument. So retuning the threshold 2.0 → 2.5 moves test and production together and stays green — the *value* is no longer pinned. But the test deliberately does **not** mirror the `Math.max` clamp, so it still pins that the truthful branch wins for these inputs, still pins the base (liquidation price, not current price), and would fail outright at a threshold large enough to hit the clamp.

That is a trade the comment at 253-254 says you made knowingly, and it is defensible: stating the coupling was the point, and the literal it replaced was itself unexplained. If you want both properties, keep `120000` and add one line asserting the coupling.

### Two smaller notes

**`healthFactorDisplay.test.ts`**: the three new tests each assert one behaviour, but the pre-existing `"shows Risky in red and no debt in green"` beside them still asserts two — so the one-behaviour-per-test rule is now satisfied for half the file. Coverage across all five `HealthFactorStatus` members is complete and correct for the first time, which is the substantive win.

**`VaultsLifecycleSections.tsx:206-210`** (outside the diff, so no inline comment): promoting the layout spec to a blocking gate gives `pending-deposit-row` a second consumer, while its pointer comment names only the real-wallet E2E CLI. The convention exists so a rename does not break something invisible — worth extending that comment when convenient.

**Merge order:** #2463 also edits `.github/workflows/verify.yml`, inserts at the same position after `Test SDK in Chromium`, rewrites the very Chromium-install step this PR's comment depends on, and its `verify` is currently failing. These will conflict — resolve deliberately rather than letting a merge driver do it.

One house-style note: the body credits a reviewer by @handle; the convention here is to attribute to the artifact — "the review on #2405/#2406".

.github/workflows/verify.yml
```diff
@@ -98,6 +98,21 @@ jobs:
...
+      # that browser rather than fetching a second one.
+      #
+      # Cost: about 35s locally for 6 tests, most of it booting the three vite
+      # dev servers the config declares (Playwright starts all three whatever
+      # project is selected). It buys the only automated check that the
+      # split-deposit progress lanes actually stack and stay inside the card at
+      # 360px - jsdom cannot see layout, so nothing else in CI catches that.
+      - name: Test vault layout in Chromium
+        run: pnpm --filter @services/vault exec playwright test --project=chromium-god-mode
```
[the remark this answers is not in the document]
jonybur: Reviewed at d2a06a9d0f6529f2254118344385fee0967ba82e.

The layout suite runs through the @services/vault:test-layout Nx target. Commit 461e2774 introduced that target and CI invocation. The revised gate still uses that path, and all six layout cases pass.

.github/workflows/verify.yml
jonybur: Reviewed at d2a06a9d0f6529f2254118344385fee0967ba82e.

CI now installs Chromium through @services/vault before the layout gate. It no longer relies on the SDK and Vault resolving the same Playwright version. The comment states the actual test scope.

services/vault/package.json
jonybur: Reviewed at d2a06a9d0f6529f2254118344385fee0967ba82e.

The layout script now sets --forbid-only and --retries=0. A temporary test.only probe passes the old discovery command and fails the revised gate. All six restored layout cases pass without retries.

services/vault/project.json
jonybur: Reviewed at d2a06a9d0f6529f2254118344385fee0967ba82e.

The Nx target now declares playwright-report and test-results as outputs. CI uploads both directories immediately after a failed layout gate, before later browser tests can replace them. The layout script uses --trace=retain-on-failure because on-first-retry records nothing when retries are disabled. A forced layout assertion failure confirms that the revised settings retain a trace with screenshots and network data on the first attempt.

packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/healthFactor.ts
jonybur: Reviewed at d2a06a9d0f6529f2254118344385fee0967ba82e.

Both remaining literal references are corrected. The liquidation sentence and risky-band range now use HEALTH_FACTOR_LIQUIDATION_THRESHOLD. The comparison operators and numeric values remain unchanged.

packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts
jonybur: Reviewed at d2a06a9d0f6529f2254118344385fee0967ba82e.

The JSDoc now states that this health-factor boundary is dimensionless and not tunable. It distinguishes the per-asset liquidationThresholdBps value fetched from chain in basis points. The generated API page is updated.

jonybur:
@greptileai Please review commit d2a06a9d0f6529f2254118344385fee0967ba82e.

Check the package-owned Chromium install, strict layout gate, first-attempt failure evidence, and distinction between health-factor and per-asset thresholds.

1,968 SDK unit tests, 7 built-package tests, 3,759 Vault tests (six existing skips), and all six layout cases pass. Full SDK and Vault lint, formatting, YAML parsing, generated docs, and executable-output checks pass. Focused-test and forced-layout-failure probes verify gate failures and trace retention. The workflow merges cleanly with #2463 and preserves both gate groups. The review-fix diff is net zero lines.

gbarkhatov (approved):

PR #2460, opened 2026-09-08
fix(vault): centralize deposit error messages

jonybur:
## Outcome

Deposit and refund error messages now share one source. The affected messages use the documented Pre-Pegin spelling.

## Change

- Move repeated hash, missing-vault, integrity, signing, and broadcast messages into `COPY.deposit.errors`.
- Keep the refund-specific hash-mismatch message.
- Document the broadcast substring match. Use the real formatter in the existing error classification tests.

## Safety

Transaction checks, cancellation codes, and error causes stay the same. Wording changes are limited to Pre-Pegin spelling and a regular hyphen. All existing test assertions remain.

## Verification

- Full Vault suite: 3,756 passed, 6 existing skips.
- Focused deposit, resume, refund, and error tests: 273 passed.
- Vault build and lint passed. Lint has 136 existing warnings.
- Formatting and diff checks passed. The review fix adds 30 lines and removes 35.

## Links

[Original copy finding on #2348](https://github.com/babylonlabs-io/babylon-toolkit/pull/2348#discussion_r3871047445).

### DIFF
--- a/services/vault/src/copy.ts
+++ b/services/vault/src/copy.ts
@@ -962,13 +962,19 @@ export const COPY = {
         body: "We couldn't broadcast your Bitcoin transaction to the network. Please try again.",
       },
       hashMismatch: (computedHash: string, chainHash: string) =>
-        `Pre-PegIn transaction hash mismatch: computed ${computedHash} from indexer tx, but on-chain contract has ${chainHash}. Aborting to prevent potential attack.`,
+        `Pre-Pegin transaction hash mismatch: computed ${computedHash} from indexer tx, but on-chain contract has ${chainHash}. Aborting to prevent potential attack.`,
+      refundHashMismatch: (computedHash: string, chainHash: string) =>
+        `Pre-Pegin transaction hash mismatch: computed ${computedHash} from indexer tx, but on-chain contract has ${chainHash}. Aborting refund to prevent potential attack.`,
+      vaultNotFound: "BTCVault not found. Please try again.",
+      prePeginIndexerTxMismatch:
+        "Transaction mismatch: the indexer returned a transaction that differs from the locally stored copy. Aborting to prevent a potential attack.",
       prePeginIntegrityMismatch:
-        "Transaction integrity check failed: the Pre-PegIn transaction does not match the hash stored on-chain. Aborting to prevent a potential attack.",
+        "Transaction integrity check failed: the Pre-Pegin transaction does not match the hash stored on-chain. Aborting to prevent a potential attack.",
       prePeginSigningCanceled:
-        "Signing canceled — the signed Pre-PegIn was not broadcast",
+        "Signing canceled - the signed Pre-Pegin was not broadcast",
+      // depositErrors.ts matches "broadcast" to select the failure callout.
       prePeginBroadcastFailed: (error: unknown) =>
-        `Failed to broadcast batch Pre-PegIn transaction: ${error instanceof Error ? error.message : String(error)}`,
+        `Failed to broadcast batch Pre-Pegin transaction: ${error instanceof Error ? error.message : String(error)}`,
       providerNotFound: {
         title: "Vault provider not found",
         body: "The selected vault provider could not be found. Please refresh and try again.",
--- a/services/vault/src/copy.ts
+++ b/services/vault/src/copy.ts
@@ -961,6 +961,14 @@ export const COPY = {
         title: "Broadcast failed",
         body: "We couldn't broadcast your Bitcoin transaction to the network. Please try again.",
       },
+      hashMismatch: (computedHash: string, chainHash: string) =>
+        `Pre-PegIn transaction hash mismatch: computed ${computedHash} from indexer tx, but on-chain contract has ${chainHash}. Aborting to prevent potential attack.`,
+      prePeginIntegrityMismatch:
+        "Transaction integrity check failed: the Pre-PegIn transaction does not match the hash stored on-chain. Aborting to prevent a potential attack.",
+      prePeginSigningCanceled:
+        "Signing canceled — the signed Pre-PegIn was not broadcast",
+      prePeginBroadcastFailed: (error: unknown) =>
+        `Failed to broadcast batch Pre-PegIn transaction: ${error instanceof Error ? error.message : String(error)}`,
       providerNotFound: {
         title: "Vault provider not found",
         body: "The selected vault provider could not be found. Please refresh and try again.",
--- a/services/vault/src/components/simple/ResumeDepositContent.tsx
+++ b/services/vault/src/components/simple/ResumeDepositContent.tsx
@@ -445,9 +445,10 @@ export function ResumeWotsContent({
         computedTxHash.toLowerCase() !== onChainPrePeginTxHash.toLowerCase()
       ) {
         throw new Error(
-          `Pre-PegIn transaction hash mismatch: computed ${computedTxHash} from indexer tx, ` +
-            `but on-chain contract has ${onChainPrePeginTxHash}. ` +
-            `Aborting to prevent potential attack.`,
+          COPY.deposit.errors.hashMismatch(
+            computedTxHash,
+            onChainPrePeginTxHash,
+          ),
         );
       }
 
--- a/services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
+++ b/services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
@@ -333,7 +333,7 @@ describe("ResumeWotsContent — Pre-PegIn tx hash trust boundary", () => {
 
     await waitFor(() => {
       expect(getByTestId("progress-view").textContent).toContain(
-        "Pre-PegIn transaction hash mismatch",
+        "Pre-Pegin transaction hash mismatch",
       );
     });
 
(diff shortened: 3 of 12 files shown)


services/vault/src/copy.ts
```diff
@@ -961,6 +961,14 @@ export const COPY = {
         title: "Broadcast failed",
         body: "We couldn't broadcast your Bitcoin transaction to the network. Please try again.",
       },
+      hashMismatch: (computedHash: string, chainHash: string) =>
```
jrwbabylonlab, line 964: This same message is still inlined in `htlcSecretDerivation.ts` (and a near-identical one in `vaultRefundService.ts`), so edits here will drift from those sites. Worth migrating them too.

services/vault/src/copy.ts
```diff
@@ -961,6 +961,14 @@ export const COPY = {
...
+        `Pre-PegIn transaction hash mismatch: computed ${computedHash} from indexer tx, but on-chain contract has ${chainHash}. Aborting to prevent potential attack.`,
+      prePeginIntegrityMismatch:
+        "Transaction integrity check failed: the Pre-PegIn transaction does not match the hash stored on-chain. Aborting to prevent a potential attack.",
```
jrwbabylonlab: These four new strings spell it “Pre-PegIn”, but this file’s header pins “Pre-Pegin” for visible strings. Note a few tests match the exact text.

services/vault/src/copy.ts
```diff
@@ -961,6 +961,14 @@ export const COPY = {
...
+      prePeginSigningCanceled:
+        "Signing canceled — the signed Pre-PegIn was not broadcast",
+      prePeginBroadcastFailed: (error: unknown) =>
```
gbarkhatov, line 976: Low-severity, and I went back and forth on it — flagging the coupling rather than asking for a change.

`depositErrors.ts:369` classifies deposit failures with `msg.includes("broadcast")`, and its comment at 364-368 documents that coupling explicitly, naming this wrapper text. So this string is now load-bearing for routing, not only for display.

The practical risk is low: any plausible rewording of a broadcast-failure message keeps the word "broadcast". But the header at `copy.ts:9` advertises this file as "one place to audit for English correctness and tone" — which is an open invitation to reword, extended to someone with no reason to know a classifier reads it. And the classifier tests construct the literal themselves rather than calling the COPY function, so nothing here would catch a rewrite that dropped the word.

A one-line comment on this key noting that `depositErrors.ts` substring-matches `"broadcast"` would be enough to close the gap.

gbarkhatov (dismissed):
Reviewed at `9a97d41a` with three independent passes. **No merge blockers — this is a faithful migration**, and I checked that claim by hand rather than inferring it from green CI.

**All four strings are byte-identical pre- to post-image.** I reassembled each one from the diff's pre-image fragments: the trailing spaces supply the joins with no doubling; `hashMismatch` reassembles to exactly `copy.ts:965` and `prePeginIntegrityMismatch` to `:967`; `prePeginSigningCanceled` is verbatim including the U+2014 em dash; and `prePeginBroadcastFailed` reproduces the `error instanceof Error ? error.message : String(error)` narrowing from the deleted `errorMsg` local. Even the deliberate asymmetry survives — `hashMismatch` says "prevent potential attack", `prePeginIntegrityMismatch` says "prevent **a** potential attack".

`cause` and the cancellation code are both preserved: `useDepositFlow.ts:1100-1102` keeps `{ cause: error }` with its explanatory comment, and `:1074-1077` keeps `Object.assign(…, { code: WALLET_CONNECTION_REJECTED_CODE })`.

---

### A correction to the "Pre-PegIn" thread — the substance stands, but the attribution changes

The header at `copy.ts:18` does pin `"Pre-Pegin" (proper-noun form) for the broadcast phase / transaction`, and these four are the only *visible-string* occurrences of the camel-I form in the file (the other hits at 489, 888 and 941 are code comments, which the rule does not bind). So the finding is real.

But **every pre-image in the diff already read "Pre-PegIn"** — lines 9, 54, 70, 86 and 104 of `pr.diff`. This is inherited shipped text migrated faithfully, not a regression introduced here. That matters twice over: it means the spelling is a pre-existing style debt rather than a defect in this PR, **and** it means it is not evidence against the byte-fidelity claim above. If you fix it, the cost is three-to-five test assertions that pin the prefix case-sensitively (`ResumeDepositContent.test.tsx:336,638`, `ensureAuthenticatedVpClient.test.ts:112`).

Worth knowing which of the four actually reach a depositor raw: `hashMismatch` and `prePeginIntegrityMismatch` do; `prePeginBroadcastFailed` is absorbed by `depositErrors.ts`' `msg.includes("broadcast")` bucket and `prePeginSigningCanceled` by the cancellation bucket, so those two are replaced with generic copy before display.

### The other existing thread also stands, with a small path correction

The hash-mismatch message is still inlined at **`services/vault/src/services/vault/htlcSecretDerivation.ts:66-70`** — a three-fragment concatenation that renders byte-identical to the new `COPY.deposit.errors.hashMismatch`, differing only in local variable names, so it is a zero-risk drop-in. And at **`vaultRefundService.ts:242-246`**, which differs by exactly one word (`Aborting refund to prevent potential attack.`) and so needs either a second key or a parameterized verb phrase. Both live under `services/vault/src/services/vault/`, not `hooks/deposit/`.

### The one genuinely new item

**Two user-facing strings are still inlined inside `useVaultActions.ts` — the PR's own file, one of them 22 lines from the line it did migrate.**

- `:246` — `"BTCVault not found. Please try again."`
- `:267-269` — `"Transaction mismatch: the indexer returned a transaction that differs from the locally stored copy. Aborting to prevent a potential attack."`

Both are thrown down the same resume path as the string migrated at line 290, and both fall through `mapDepositError` to the raw-message callout, so both are user-visible. This is a stronger case than the two files above, because no new file needs opening — the author is already editing this exact function, and CLAUDE.md's rule is that a string found still inlined gets migrated as part of the change. Suggested keys: `prePeginIndexerTxMismatch` and a vault-not-found entry, alongside the four new ones.

I could not anchor this inline — those lines fall outside the diff hunk, which covers roughly 287-296.

### Checked and dismissed

- **"Green CI proves all four strings are unchanged."** Weaker than it looks, and worth recording. The tests assert only *prefixes*, so the interpolated tails are uncovered, and `prePeginSigningCanceled` has **no** exact-text assertion anywhere. The cross-check holds for three of four; for that one it is the byte comparison above, not CI, that establishes preservation.
- **Dropped `cause` or cancellation code.** Neither — verified in the diff.
- **Dead code from the removed `errorMsg` local.** None; the two surviving `errorMsg` bindings are independent declarations in unrelated catch scopes.
- **"Mixed shapes in `COPY.deposit.errors` break the namespace."** No — `cannotBroadcastInState` (a builder) alongside `defaultTitle` and `genericBody` (bare strings) already establishes the precedent, and nothing iterates the object generically.

### Verified clean

- Net line count genuinely neutral at +19/−19, with no logic riding along.
- `COPY` is imported and used in every consuming file; the diff correctly adds the import only to `ensureAuthenticatedVpClient.ts:32` — the other three already had it.
- **Interpolation argument order preserved on both call sites** — computed hash first, chain hash second. A swap would have silently inverted a security diagnostic; it did not happen.
- Using functions for the two interpolated messages follows the "prefer a function in `copy.ts`" rule rather than working around it.
- No §1-§9 critical path, no `data-testid`, no motion, no SDK public surface — `docs:clean` not implicated.

services/vault/src/copy.ts
jonybur: Reviewed at 420dc76d2add3e59b87a79c764c203400b8278c7.

The HTLC path now uses the shared hash-mismatch formatter. The refund path uses a COPY formatter that keeps the refund-specific wording. Both keep the computed hash before the on-chain hash. Existing refund checks pass.

services/vault/src/copy.ts
jonybur: Reviewed at 420dc76d2add3e59b87a79c764c203400b8278c7.

The affected messages now use Pre-Pegin, as required by the copy header. Existing mismatch assertions use the same spelling. The cancellation message also uses a regular hyphen. Transaction checks, cancellation codes, and error causes stay the same.

services/vault/src/copy.ts
jonybur: Reviewed at 420dc76d2add3e59b87a79c764c203400b8278c7.

The formatter now has a comment that records the broadcast substring match. Existing classification tests use the real COPY formatter, so a wording change that drops broadcast fails those tests. The cause and cancellation assertions remain in place.

jonybur:
Reviewed at 420dc76d2add3e59b87a79c764c203400b8278c7.

Follow-up to [this review](https://github.com/babylonlabs-io/babylon-toolkit/pull/2460#pullrequestreview-5152859930).

The two remaining resume messages from this review now use COPY: the missing-vault message and the indexer transaction-mismatch message. The duplicate HTLC and refund messages also moved to COPY. The affected messages now use Pre-Pegin. Existing transaction checks, cancellation codes, and error causes stay the same.

jonybur:
@greptileai Please review commit 420dc76d2add3e59b87a79c764c203400b8278c7.

Finish the remaining deposit and refund error-copy findings. Keep error behavior and refund-specific meaning.

Full Vault suite: 3,756 passed, 6 existing skips. Focused checks: 273 passed. Vault build, lint, formatting, and diff checks passed. Lint has 136 existing warnings.

gbarkhatov (approved):

PR #2461, opened 2026-09-08
fix(vault): wait for offscreen visual changes

jonybur:
## Outcome

Vault visual checks wait for pixel changes below the viewport. Scan failures now retain the useful error and report the measured scan cost.

## Change

Compare overlapping viewport captures across the full document, including one-pixel overflow. Restore the original scroll position after each scan. Keep the original capture path when the document fits the viewport.

Report the last scan's capture count, duration, and document dimensions on timeout. Preserve the scan error if restoration also fails, and log the restore error. A standalone restore failure still rejects the capture.

## Safety

The viewport dimensions stay fixed. The existing layout and error checks run before the final image is written. This change is limited to the visual test helper.

## Verification

- Final visual suite: 18 tests passed and all 58 expected images produced.
- Five checks on existing recorded screens passed: both restore-error paths, measured timeout details, one-pixel overflow with exact scroll restoration, and the fitted-page fast path without scrolling or resizing.
- Full Vault tests: 3,756 passed, 6 existing skips. Build, lint, helper type checking, formatting, and diff checks passed. Lint reports 136 existing warnings.
- Earlier offscreen-change check: a three-second change at fixed document dimensions made the old check return after 667ms; the fixed check waited 3,901ms. Earlier 6x and 20x CPU slowdown checks kept the viewport and desktop layout intact.
- Earlier repeat-image checks found a pre-existing variation: one deposit image differed at three pixels by one color level. The unchanged helper reproduced it. Final and control runs produced the same two image hashes. This did not prevent the complete visual runs from passing.

## Links

Follow-up to [the Greptile finding on #2308](https://github.com/babylonlabs-io/babylon-toolkit/pull/2308#discussion_r3847745871).

### DIFF
--- a/services/vault/e2e/visual/stabilize.ts
+++ b/services/vault/e2e/visual/stabilize.ts
@@ -71,6 +71,9 @@ const HIDE_GOD_MODE_LAUNCHER_CSS = `
   }
 `;
 
+/** Overlap captures so fixed headers do not hide the same document pixels. */
+const STABILITY_SCROLL_STEP_RATIO = 0.5;
+
 /** How long the frame must stay byte-identical before we trust it. */
 const STABILITY_QUIET_MS = 300;
 /** Consecutive identical frames required. */
@@ -120,10 +123,11 @@ async function readFrameSignature(page: Page): Promise<FrameSignature> {
   if (maxX === 0 && maxY === 0) {
     return { pixels: await page.screenshot(), documentWidth, documentHeight };
   }
+  const stepX = viewportWidth * STABILITY_SCROLL_STEP_RATIO;
+  const stepY = viewportHeight * STABILITY_SCROLL_STEP_RATIO;
   try {
-    // Overlap captures so fixed headers do not hide the same document pixels.
-    for (let top = 0; ; top = Math.min(top + viewportHeight / 2, maxY)) {
-      for (let left = 0; ; left = Math.min(left + viewportWidth / 2, maxX)) {
+    for (let top = 0; ; top = Math.min(top + stepY, maxY)) {
+      for (let left = 0; ; left = Math.min(left + stepX, maxX)) {
         await page.evaluate(
           ([left, top]) => window.scrollTo({ left, top, behavior: "instant" }),
           [left, top],
--- a/services/vault/e2e/visual/stabilize.ts
+++ b/services/vault/e2e/visual/stabilize.ts
@@ -71,7 +71,10 @@ const HIDE_GOD_MODE_LAUNCHER_CSS = `
   }
 `;
 
-/** Overlap captures so fixed headers do not hide the same document pixels. */
+/**
+ * The half-viewport overlap (400px desktop, 422px mobile) must exceed the fixed
+ * header height so each document pixel appears in at least one capture.
+ */
 const STABILITY_SCROLL_STEP_RATIO = 0.5;
 
 /** How long the frame must stay byte-identical before we trust it. */
@@ -103,11 +106,14 @@ interface FrameSignature {
   readonly pixels: Buffer;
   readonly documentWidth: number;
   readonly documentHeight: number;
+  readonly frameCount: number;
+  readonly durationMs: number;
 }
 
 /** Compare overlapping captures. Restore the scroll position after each poll. */
 async function readFrameSignature(page: Page): Promise<FrameSignature> {
-  // Read the document size first to settle layout before capture.
+  const startedAt = performance.now();
+  // Measuring the document first settles layout and prevents unstable deposit card images.
   const { documentWidth, documentHeight, viewportWidth, viewportHeight, x, y } =
     await page.evaluate(() => ({
       documentWidth: document.documentElement.scrollWidth,
@@ -121,10 +127,17 @@ async function readFrameSignature(page: Page): Promise<FrameSignature> {
   const maxX = Math.max(0, documentWidth - viewportWidth);
   const maxY = Math.max(0, documentHeight - viewportHeight);
   if (maxX === 0 && maxY === 0) {
-    return { pixels: await page.screenshot(), documentWidth, documentHeight };
+    return {
+      pixels: await page.screenshot(),
+      documentWidth,
+      documentHeight,
+      frameCount: 1,
+      durationMs: Math.round(performance.now() - startedAt),
+    };
   }
   const stepX = viewportWidth * STABILITY_SCROLL_STEP_RATIO;
   const stepY = viewportHeight * STABILITY_SCROLL_STEP_RATIO;
+  let scanFailed = false;
   try {
     for (let top = 0; ; top = Math.min(top + stepY, maxY)) {
       for (let left = 0; ; left = Math.min(left + stepX, maxX)) {
@@ -137,16 +150,34 @@ async function readFrameSignature(page: Page): Promise<FrameSignature> {
       }
       if (top === maxY) break;
     }
+  } catch (error) {
+    scanFailed = true;
+    throw error;
   } finally {
-    await page.evaluate(
-      ([left, top]) => window.scrollTo({ left, top, behavior: "instant" }),
-      [x, y],
-    );
+    await page
+      .evaluate(
+        ([left, top]) => window.scrollTo({ left, top, behavior: "instant" }),
+        [x, y],
+      )
+      .catch((error) => {
+        if (!scanFailed) throw error;
+        // eslint-disable-next-line no-console -- Keep the secondary browser failure in the test log.
+        console.error(
+          "Scroll restoration also failed after the visual scan:",
+          error,
+        );
+      });
   }
-  return { pixels: Buffer.concat(frames), documentWidth, documentHeight };
+  return {
+    pixels: Buffer.concat(frames),
+    documentWidth,
+    documentHeight,
+    frameCount: frames.length,
+    durationMs: Math.round(performance.now() - startedAt),
+  };
 }
 
-/** True only when both halves of the signal are unchanged. */
+/** Dimensions detect document growth; pixels detect paint changes at the same size. */
 function isSameFrame(a: FrameSignature, b: FrameSignature): boolean {
   return (
     a.documentWidth === b.documentWidth &&
@@ -207,7 +238,8 @@ export async function waitForVisualStability(page: Page): Promise<void> {
 
   throw new Error(
     `Page did not reach a stable frame within ${STABILITY_TIMEOUT_MS}ms. ` +
-      `Something on this screen animates or refetches indefinitely - freeze it ` +
-      `in stabilize.ts rather than accepting a flaky baseline.`,
+      `Last scan: ${previous.frameCount} captures in ${previous.durationMs}ms ` +
+      `for a ${previous.documentWidth}x${previous.documentHeight}px document. ` +
+      `Check scan cost, animations, and refetches before changing the timeout.`,
   );
 }
--- a/services/vault/e2e/visual/stabilize.ts
+++ b/services/vault/e2e/visual/stabilize.ts
@@ -25,24 +25,10 @@
  *   covered by `waitForVisualStability`, which polls until the rendered
  *   frame stops changing rather than guessing a fixed delay.
  *
- * The stability poll photographs the VIEWPORT, never the full page. A
- * full-page screenshot of a page taller than the viewport makes Playwright
- * take Chromium's `captureBeyondViewport` path, and Chromium emulates a
- * 1x1 viewport for a moment while it does. The page gets a real `resize`
- * event with `window.innerWidth === 1`, then a restore ~10-250ms later.
- * Any throttled or debounced resize listener in app code takes the 1x1
- * edge and keeps it: `setFixedTime` above freezes `Date.now()` for the
- * life of the page, lodash.throttle derives its window from `Date.now()`,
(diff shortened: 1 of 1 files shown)


jonybur:
@greptileai

services/vault/e2e/visual/stabilize.ts
```diff
@@ -110,45 +99,51 @@ export async function installVisualDeterminism(page: Page): Promise<void> {
...
+  const frames: Buffer[] = [];
+  const maxX = Math.max(0, documentWidth - viewportWidth);
+  const maxY = Math.max(0, documentHeight - viewportHeight);
+  if (maxX === 0 && maxY === 0) {
+    return { pixels: await page.screenshot(), documentWidth, documentHeight };
+  }
+  const stepX = viewportWidth * STABILITY_SCROLL_STEP_RATIO;
+  const stepY = viewportHeight * STABILITY_SCROLL_STEP_RATIO;
+  try {
+    for (let top = 0; ; top = Math.min(top + stepY, maxY)) {
```
gbarkhatov, line 142: **Per-poll cost now scales with document height, against a `STABILITY_TIMEOUT_MS` that did not move.** Non-blocking — the `visual` job passed at 16m10s on this head — but the headroom is worth knowing about.

`readFrameSignature` used to be one screenshot. It is now `(⌈maxY/stepY⌉+1) × (⌈maxX/stepX⌉+1)` evaluate+screenshot round trips, uncapped, and the loops are nested so the two multiply.

The arithmetic is tighter than it looks. `deadline` is set at `:191` *before* the first read at `:192`, and success needs three reads (`previous`, then two loop iterations to reach `STABILITY_CONSECUTIVE_MATCHES = 2`). The `while` guard before the third read is evaluated at `2R + 300 ms`, so any per-read cost above **~7.35 s** makes success arithmetically impossible no matter how still the page is. At 1280×800 / 390×844 a tall page is roughly 7-15 frames per read.

The half of the fix I'd actually push for is the cheap one — **the error message at `:208-211` misdirects.** It says "Something on this screen animates or refetches indefinitely," which sends the next engineer hunting an animation that does not exist. Putting the frame count and the measured read duration into that message turns a confusing failure into an obvious one.

Capping frames per scan, or hashing each frame instead of concatenating raw PNGs into `Buffer.concat`, is optional on top.

services/vault/e2e/visual/stabilize.ts
```diff
@@ -110,45 +99,51 @@ export async function installVisualDeterminism(page: Page): Promise<void> {
...
+        await page.evaluate(
+          ([left, top]) => window.scrollTo({ left, top, behavior: "instant" }),
+          [left, top],
+        );
+        frames.push(await page.screenshot());
+        if (left === maxX) break;
+      }
+      if (top === maxY) break;
+    }
+  } finally {
```
gbarkhatov, line 156: Putting the restore in a `finally` is the right call — it is exactly what makes the wrong-baseline failure mode impossible, and I checked that it works.

One refinement: a `finally` that *throws* discards the in-flight exception by JS abrupt-completion rules. If the scan aborts because the page navigated or closed, that is the actionable error — but this `page.evaluate` will then throw `Execution context was destroyed`, and that is what surfaces instead.

One line keeps the real error:

```ts
} finally {
  try {
    await page.evaluate(([left, top]) => window.scrollTo({ left, top, behavior: "instant" }), [x, y]);
  } catch {
    // A destroyed context has no scroll position to restore, and must not
    // displace the error that destroyed it.
  }
}
```

services/vault/e2e/visual/stabilize.ts
```diff
@@ -110,45 +99,51 @@ export async function installVisualDeterminism(page: Page): Promise<void> {
...
+    await page.evaluate(() => ({
+      documentWidth: document.documentElement.scrollWidth,
+      documentHeight: document.documentElement.scrollHeight,
+      viewportWidth: window.innerWidth,
+      viewportHeight: window.innerHeight,
+      x: window.scrollX,
+      y: window.scrollY,
+    }));
+  const frames: Buffer[] = [];
+  const maxX = Math.max(0, documentWidth - viewportWidth);
```
gbarkhatov, line 127: One pixel of horizontal overflow doubles the screenshot count for no added signal.

Columns come out as `⌈maxX/stepX⌉ + 1`, so any `maxX` in `1..stepX` yields two columns — at `maxX === 1` the second capture is one pixel from the first. Because the horizontal loop is nested inside the vertical one, that doubles the **total** frame count, which feeds straight into the timeout headroom I noted on `:129`.

At the 390 px mobile viewport a single stray pixel of overflow is easy to introduce by accident, and this would silently halve the timeout budget when it happens.

Clamping columns to 1 when `maxX` is below a named minimum (a `STABILITY_MIN_SCAN_OVERFLOW_PX`, say) costs one constant and removes the cliff.

services/vault/e2e/visual/stabilize.ts
```diff
@@ -85,6 +71,9 @@ const HIDE_GOD_MODE_LAUNCHER_CSS = `
   }
 `;
 
+/** Overlap captures so fixed headers do not hide the same document pixels. */
```
gbarkhatov: Residue on the greptile thread, not a re-raise — the extraction itself is done properly and both loops share this constant.

What is still missing against "it must be obvious why that value was chosen" is the *bound*: the comment gives the purpose of overlapping (fixed headers must not hide the same document pixels) but not what fixes the ratio at one half rather than, say, 0.75.

The reason is checkable — the overlap has to exceed the tallest fixed/sticky overlay, and 0.5 clears it at both viewports (400 px desktop, 422 px mobile). One clause recording that turns the number from a plausible default into a justified one, and tells the next person what would invalidate it.

services/vault/e2e/visual/stabilize.ts
```diff
@@ -110,45 +99,51 @@ export async function installVisualDeterminism(page: Page): Promise<void> {
...
- *   nothing below it, and it is the half that must not be full-page (see
- *   the header comment).
- * - `documentWidth` / `documentHeight` cover what the crop hides. A list
- *   that grows below the fold, a lazy chunk that lands off-screen and a
- *   collapsing skeleton all move the document box, so growth the pixels
- *   cannot show still reads as "not settled".
- *
- * A change to EITHER part means the screen is still moving.
- */
 interface FrameSignature {
```
gbarkhatov, line 105: The rewrite dropped the rationale for two design decisions while keeping the decisions themselves.

Gone: the block explaining why `FrameSignature` has two halves ("neither alone is enough"), and the one explaining why the document box is measured *before* the pixels — which recorded a real observed flake, the deposit dialog's amount card coming out two ways across 12 runs.

The code still does both things, and the surviving comments still *reference* them — `isSameFrame` says "both halves of the signal" and `:110` says "settle layout before capture" — but nothing now records why either matters. So the next rewrite has nothing to stop it collapsing the signature to pixels alone, or reordering the reads, and rediscovering that flake the expensive way.

One sentence each is cheaper than finding it twice.

gbarkhatov (dismissed):
Reviewed at `5d21a16a` with three independent passes. **No merge blockers**, and the two hazards I went in most worried about are both genuinely clean.

**The greptile thread is fully addressed and can be resolved.** Both halves of it: the inline `2` is gone from *both* loops, and a single `STABILITY_SCROLL_STEP_RATIO = 0.5` feeds `stepX` **and** `stepY`, so horizontal and vertical traversal are now structurally incapable of diverging. That was the part of the finding that was easy to half-fix, and it wasn't half-fixed.

### What I checked hardest, and what it came back as

- **Is the scroll restore exception-safe?** Yes — `:128-145` is a real `try`/`finally` restoring the exact pre-scan `scrollX`/`scrollY` (fractional offsets included) on every path, including a mid-scan throw. This was my main concern going in, because the failure mode is silent *wrong baseline images* rather than an error, and it does not exist here. The fast path at `:123-125` returns before the `try` and never scrolls, so it correctly owes no restore.
- **Does the overlap leave an unscanned band?** No. `Math.min(top + stepY, maxY)` clamps the final sample to land exactly on the bound, so the last row and column are always taken and consecutive samples overlap by half a viewport. At the real viewports the steps are integers (640/400 desktop, 195/422 mobile), so the `left === maxX` / `top === maxY` float equality is exact.
- **Can the scan loop hang or exit early?** No, for any realistic input. Height is read once per read, so a document growing mid-scan under-scans *that one poll* — but the next poll reads the new height, the frame count changes, `Buffer.concat` lengths differ, and the poll correctly keeps waiting. Self-correcting by construction.
- **Is the "fits the viewport" fast path equivalent to the old behaviour?** Yes — `maxX === 0 && maxY === 0` returns a bare `page.screenshot()` with the same `documentWidth`/`documentHeight`. Byte-for-byte the pre-PR path.

Your verification write-up is unusually thorough and it held up. The one thing none of it exercises is the exception path, which is why I went at it directly.

### Checked and dismissed

- **`stepY === 0` infinite loop.** Raised by two of the three passes as a defensive finding. It needs `window.innerHeight === 0` with a non-zero document; Playwright always sets a non-zero viewport, and the Chromium 1×1 emulation this file's own header documents fires only on `fullPage`, which the poll no longer takes. Hardening with `page.viewportSize()` is reasonable, but it is a suggestion, not a defect.
- **`documentWidth`/`documentHeight` are now redundant.** No — they are read in `isSameFrame` at `:152-153` and still carry signal the pixels cannot: in the fast path they are the *only* growth detector, and in the scan path a document growing into uniform whitespace can shift `maxY` while the clamped last frame stays pixel-identical.
- **The deleted antialiasing hazard is reintroduced because `scrollTo` flushes layout between rasters.** No — each scroll is a completed, awaited CDP round trip before the next screenshot is requested, so the flush/raster ordering is explicit. The nondeterminism the old comment described required an *unsynchronized* flush.
- **Arithmetic based on a 1280×720 viewport.** That is the `devices["Desktop Chrome"]` default at `playwright.visual.config.ts:126`; the operative viewports are `VISUAL_VIEWPORTS` at `targets.ts:54-55` — 1280×800 and 390×844. The inline note below uses those.

### Verified clean

- No dead code — every constant, `FrameSignature` and `isSameFrame` has a live use; the rewrite orphaned nothing.
- The motion assumption holds: `contextOptions: { reducedMotion: "reduce" }` at `playwright.visual.config.ts:113`, so the reliance on core-ui's global `*` reset plus `FREEZE_SPINNER_CSS` is intact and untouched by this PR.
- No swallowed errors and no silent fallbacks — the `Math.max(0, …)` bounds are legitimate clamps, not masking defaults.
- Sibling #2463 touches `playwright.config.ts`, but the visual suite reads `playwright.visual.config.ts` and takes viewports from `targets.ts`, so there is no concrete interaction with this helper.

services/vault/e2e/visual/stabilize.ts
jonybur: Reviewed at 7d8b539859442b0e2e3f7aae2364d79bff7b6bd4.

The timeout now reports the actual capture count, elapsed time, and document dimensions from the last scan. It asks the reader to check scan cost, animations, and refetches. It no longer claims indefinite animation. A probe that changed the existing page pixels reached the timeout and verified the reported count and measured duration.

services/vault/e2e/visual/stabilize.ts
jonybur: Reviewed at 7d8b539859442b0e2e3f7aae2364d79bff7b6bd4.

The scan error now remains the primary error when scroll restoration also fails. The secondary restore error is written to the test log. A restore failure after a successful scan still rejects the capture. Real-page probes verified both paths by closing the browser page during capture.

services/vault/e2e/visual/stabilize.ts
jonybur: Reviewed at 7d8b539859442b0e2e3f7aae2364d79bff7b6bd4.

I kept the one-pixel scan. At maxX = 1, the rightmost document column is outside the first viewport capture. Clamping to one column would omit those pixels and could accept a changing edge. The existing-page probe captured x = 0 and x = 1, then restored the original y = 137 offset. The new timeout details expose the extra scan cost. No overflow threshold was added.

services/vault/e2e/visual/stabilize.ts
jonybur: Reviewed at 7d8b539859442b0e2e3f7aae2364d79bff7b6bd4.

The comment now records the bound: the half-viewport overlap is 400px on desktop and 422px on mobile, and it must exceed the fixed header height. This explains the ratio and states when it must change.

services/vault/e2e/visual/stabilize.ts
jonybur: Reviewed at 7d8b539859442b0e2e3f7aae2364d79bff7b6bd4.

The comments now explain both decisions. Dimensions detect document growth; pixels detect paint changes at the same size. Measuring the document before capture settles layout and prevents unstable deposit card images.

jonybur:
@greptileai Please review commit 7d8b539859442b0e2e3f7aae2364d79bff7b6bd4.

Report measured visual scan cost and preserve both scan and restore failures. Keep complete pixel coverage and explain the capture rules.

Final visual suite: 18 passed, 58 images. Five real-page probes passed. Full Vault tests: 3,756 passed, 6 existing skips. Vault build, lint, helper type checking, formatting, and diff checks passed. Lint has 136 existing warnings.

gbarkhatov (approved):

PR #2462, opened 2026-09-08
fix(vault): keep Aave debt visible without indexer data

### LINKED ISSUE #2389: [Aave][Critical] Use on-chain state when the indexer returns no positions
## Outcome

Show live Aave debt and keep Repay available when the indexer omits an active position.

Status: the fix is pushed on [fix/2389-aave-onchain-position](https://github.com/babylonlabs-io/babylon-toolkit/tree/fix/2389-aave-onchain-position) at [7c076c8](https://github.com/babylonlabs-io/babylon-toolkit/commit/7c076c8a6abc8c90e3b2fe1a827675b8763702db). GitHub verifies the signature. The commit matches the reviewed nine-file patch. All 3,765 tests, build, and lint passed before the push. PR #2462 is open. PR checks and required reviews remain.

## Problem

Severity: Critical, I4 x L2 with silent liquidation risk.

Audit commit: `4f2681948d22257341897e58c0c79bcd5b66371b`.

`getUserPositionsWithLiveData` returns an empty array when GraphQL returns no positions. It does not read the on-chain position first ([positionService.ts:117-129](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/services/vault/src/applications/aave/services/positionService.ts#L117-L129)).

The hook converts this result to zero collateral and zero debt. The Loans page then shows an empty state and removes Repay ([useAaveUserPosition.ts:162-210](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/services/vault/src/applications/aave/hooks/useAaveUserPosition.ts#L162-L210), [Loans.tsx:108-145](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/services/vault/src/components/pages/Loans.tsx#L108-L145)).

Reproduce:

1. Create an on-chain Aave position with active debt.
2. Return `items: []` from the active-position GraphQL request.
3. Open the Loans page.
4. Observe zero debt and no Repay action.
5. Keep the empty response active while the market moves through the liquidation threshold.

The later proxy guard cannot help because it runs only after Repay starts ([positionTransactions.ts:199-227](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/services/vault/src/applications/aave/services/positionTransactions.ts#L199-L227)).

## Scope

- Read position existence and proxy data on-chain before accepting an empty indexer result.
- If the position exists, query the Spoke for all configured debt reserve IDs.
- Use indexer data only as enrichment.
- Keep emergency Repay available from authoritative chain state.
- Add empty-indexer regression tests.

The SDK already has an on-chain position helper ([query.ts:57-110](https://github.com/babylonlabs-io/babylon-toolkit/blob/4f2681948d22257341897e58c0c79bcd5b66371b/packages/babylon-ts-sdk/src/tbv/integrations/aave/clients/query.ts#L57-L110)).

## Complete when

- Active on-chain debt stays visible when the indexer returns no positions.
- Repay stays available for that debt.
- No on-chain position still gives the empty state.
- RPC failures produce an error instead of false zero values.
- Tests cover both empty-indexer cases.

## Dependencies

None. Related history: #1500 and #1586.

jonybur:
## Outcome

Aave debt and Repay remain available when indexer data is missing or a background read fails. The current commit is ready for merge: code-owner approval, passing CI, and Greptile 5/5.

## Change

Read position existence, collateral totals, and vault IDs from the chain. Show a warning when indexed collateral is incomplete. Keep loaded debt and urgent liquidation warnings visible after read failures. Count each activation once.

## Safety

Repay requires loaded debt for the selected asset. Borrow and missing debt retain the error block. Incomplete collateral cannot enable cascade or reorder calculations. Retry handles failed reads. Repayment keeps its wallet, balance, identity, and transaction checks.

## Verification

At `73bf96aab1652d21ffbea3fa000e2d72a9474f33`:

- All CI checks pass. Local Vault tests pass: 3,809 tests, with 6 existing skips. Build and lint pass.
- All 10 review threads are resolved. Code-owner approval covers this commit.
- [Visual review](https://github.com/babylonlabs-io/babylon-toolkit/pull/2462#issuecomment-5605905374) confirms all 58 Vault images and 329 Storybook images are present. All 42 deposit images are unchanged. One unchanged Storybook case has an external icon capture difference.
- [Greptile gives 5/5](https://github.com/babylonlabs-io/babylon-toolkit/pull/2462#issuecomment-5593120618) for this commit, with no remaining findings.

The separate full E2E typecheck has an existing recovery-command error. It is outside this PR.

## Links

Closes #2389.


jonybur:
Deposit verification is incomplete at `7c076c8a6abc8c90e3b2fe1a827675b8763702db`. The deposit screen was not deleted.

- The [visual run](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34289193096/job/102271677736) has four failed capture tests and 14 passed tests. The baseline captured 58 Vault images; this branch captured 14.
- The saved session has no response for the new `getPosition(address)` chain read (`0x16c19739`). Deposit-flow tests show the load-error state without the Deposit button. Deposit-progress tests reject the unanswered read. These failures leave 44 images unwritten.
- The report labels missing candidate images as “removed by this PR - before only”. The workflow also allows the partial capture failure to finish green. This result does not prove that the deposit flow passes.
- A valid rerun needs a real recorded chain response. The empty indexer result cannot establish that no chain position exists. The report also needs to distinguish incomplete captures from removed screens and fail incomplete runs.

The earlier claim that all online checks passed missed these capture failures. No code changes were made during this investigation.

services/vault/src/applications/aave/services/positionService.ts
```diff
@@ -121,19 +74,24 @@ export async function getUserPositionsWithLiveData(
...
-  // User can only have one position in Babylon vault integration
-  const position = positions[0];
-  const proxyAddress = position.proxyContract as Address;
+  const [position, indexedPositions] = await Promise.all([
+    getPosition(
+      ethClient.getPublicClient(),
+      getAaveAdapterAddress(),
+      depositor as Address,
+    ),
+    fetchAaveActivePositionsWithCollaterals(depositor).catch(() => []),
```
jrwbabylonlab: When the indexer call fails this swallows the error — no log, no UI signal — so the page shows the chain total above an empty vault list with no explanation. A log plus an “incomplete data” hint would help.

services/vault/src/components/pages/Loans.tsx
```diff
@@ -110,15 +113,19 @@ export default function Loans() {
   // section, whose Borrow button was enabled whenever collateral was present.
   const hasPosition = hasCollateral || hasLoans || godModeAffectsPage;
 
-  // Position is still loading for a connected wallet: hold on a spinner rather
-  // than flashing the full-page "deposit" empty state before the summary lands
-  // (hasCollateral/hasLoans are false until the position resolves).
-  if (isConnected && isLoading && !godModeAffectsPage) {
+  // Unknown position data must not show the empty state.
+  if (isConnected && (positionError || isLoading) && !godModeAffectsPage) {
```
jrwbabylonlab: This gate also fires when a background poll fails while React Query still holds the last good data (v5 sets `error` but keeps `data`), so a blip swaps the visible debt and Repay button for the error state. Prefer the held data when it exists.

services/vault/src/applications/aave/components/Detail/index.tsx
```diff
@@ -130,6 +133,17 @@ export function LoanFlowOverlay({
       );
     }
     const mode = picker ?? tab;
+    if (mode === LOAN_TAB.REPAY && isConnected && positionError) {
```
gbarkhatov: **Merge blocker — same defect as the `Loans.tsx:117` thread, but at a second site that thread does not cover, and here the consequence is worse.**

`if (mode === LOAN_TAB.REPAY && isConnected && positionError)` returns `PositionGate` with `{null}` children, discarding `repayAssets` **unconditionally**. `repayAssets` (:71-80) derives from `borrowedAssets` → the same retained `position` that React Query keeps on a refetch failure. So a transient poll blip empties a picker the user has just opened, mid-flow.

Worse than the Loans page, because `AssetSelectionPanel.tsx:248` carries the `asset-select-row-<symbol>` selector that the real-wallet E2E CLI drives (`e2e/real/actions/repay.ts:231`). The repay flow becomes unreachable, not merely degraded — and repay is the action a user most needs when their position is deteriorating.

```ts
if (mode === LOAN_TAB.REPAY && isConnected && positionError && repayAssets.length === 0) {
```

The fall-through already passes `assetsLoading={isPositionLoading}`, so the genuinely-unknown case stays covered.

services/vault/src/components/pages/Loans.tsx
```diff
@@ -110,15 +113,19 @@ export default function Loans() {
...
     return (
       <Container className={`${PAGE_CONTENT_CLASS} pb-6`}>
-        <div className="flex items-center justify-center py-12">
-          <Loader />
-        </div>
+        <PositionGate
+          positionError={positionError}
+          ancillaryError={null}
```
gbarkhatov, line 129: `ancillaryError={null}` is hardcoded here and at `Detail/index.tsx:140` — which leaves the seam that would fix the swallowed-indexer thread dead exactly where it is most needed.

`PositionGate.tsx:43` already renders a soft-warn banner whenever `ancillaryError` is non-null, using `COPY.loans.detail.ancillaryLoadWarning` — copy that already reads *"Some data couldn't be loaded… repay still works from your loaded debt"*. That is a precise description of the state `positionService.ts:83`'s `.catch(() => [])` produces. `ReserveDetailPanel.tsx:186` passes a real one from `useAaveReserveDetail.ts:279`, so the pattern is established.

So the missing "incomplete data" signal asked for on that thread does not need new UI or new copy — it needs the indexer rejection carried out of `positionService` as a distinct `indexerError` and threaded to these two literals.

A zero-RPC detector already exists if you want one without touching the service: `position.vaultIds` is now the chain-authoritative set, so `vaultIds.length !== rawCollateralVaults.length` means the indexer is behind.

services/vault/src/components/pages/__tests__/Loans.test.tsx
```diff
@@ -145,43 +148,54 @@ describe("Loans page — loading gate", () => {
...
-      "data-connected",
-      "false",
-    );
-    // Disconnected prompts for a wallet — it must not claim the depositor has
-    // no loans when we haven't looked at an address yet.
-    expect(screen.getByText(COPY.loans.emptyDisconnected)).toBeInTheDocument();
-    expect(container.querySelector("svg")).not.toBeInTheDocument();
-    expect(screen.queryByTestId("loans-summary")).not.toBeInTheDocument();
-  });
+  it.each<[string, boolean, boolean, boolean, boolean, Error | null]>([
```
gbarkhatov: Two things about this table, and the first explains why green CI says nothing about the merge blockers above.

**The error case is only ever tested with no data.** The single error row is `["error", true, false, false, false, new Error("RPC unavailable")]` — collateral false, debt false. The regression on `Loans.tsx:117` is *error **with** retained data*, which no row exercises. Adding one row reproduces it as a failing test:

```ts
["summary", true, false, false, true, new Error("RPC unavailable")],
```

Relatedly, `CONNECTED_LOADED` omits `refetchPosition` entirely, so the error row renders `PositionGate` with `refetchPosition={undefined}` and nothing ever clicks Retry — so the throwing-refetch wiring is untested too.

**Second: this is the parameterized abstraction the test philosophy explicitly rejects**, and it replaced four named tests. The assertions are formulas restating the production branch conditions — `expect(Boolean(queryByTestId("loans-empty-state"))).toBe(view === "empty" || (view === "summary" && !debt))` — so they cannot fail when that logic is *wrong*, only when it is inconsistent with itself. That is the mechanism by which the blocker above slipped through.

Also: the title template on :159 has five `%s` for six parameters, so the error parameter never appears in the reporter, and the deleted `data-connected="false"` assertion leaves the stub's `data-connected` attribute as dead scaffolding.

Worth noting sibling #2465 hit the identical `it.each` finding and de-parameterized (7 tables → 19 explicit tests, 64/64 assertions intact).

services/vault/src/applications/aave/services/__tests__/getUserPositionsWithLiveData.test.ts
```diff
@@ -29,228 +43,191 @@ import { getUserPositionsWithLiveData } from "../positionService";
...
-    // empty for downstream tests — vi.clearAllMocks does not drain queued
-    // onces.
-    mockGetUserPositionsBatch.mockImplementation(async () => [null, null]);
-
-    const result = await getUserPositionsWithLiveData(DEPOSITOR, SPOKE, {
-      borrowableReserveIds: [USDC_RESERVE_ID, DAI_RESERVE_ID],
-      vbtcReserveId: VBTC_RESERVE_ID,
-    });
-
+  it("does not read debt reserves when the on-chain borrowCount is zero", async () => {
```
gbarkhatov, line 141: Dead setup left by a deleted test: this line sets `mockGetUserPositionsBatch.mockResolvedValue([null, null])` inside a test that then asserts the batch is **never called**. It is residue from the removed *"treats per-reserve probe failures (null in the batch) as 'no debt'"* case.

That removal matters on its own: the `position &&` null guard at `positionService.ts:163` is still live production code on the debt path and is now covered by nothing.

Separately, `it.each([mockGetPosition, mockGetUserPositionWithAccountData])` at :211 has no `%s` in its title, so both cases report under an identical name — a failure does not say which one broke.

services/vault/src/applications/aave/services/positionService.ts
```diff
@@ -20,98 +19,52 @@ import {
...
   drawnShares: bigint;
   premiumShares: bigint;
   totalDebt: bigint;
 }
 
-/**
- * Position with live on-chain data
- */
-export interface AavePositionWithLiveData extends AavePosition {
+/** Chain position with optional indexer timestamps and collateral details. */
```
gbarkhatov: `Omit<AavePosition, "createdAt" | "updatedAt">` plus re-declaring both as optional exists only so `{...indexedPosition}` typechecks when the indexer is absent — they are spread residue, not fields anyone consumes.

I grepped `applications/aave`, `hooks` and `components`: the only hits are `useAaveVaults.ts` (a different type) and `fetchPositions.ts`' own declaration. Nothing reads either field off this interface.

Zero-dead-code says drop them from the interface; the spread still typechecks because the explicit reassignments below cover the real fields.

services/vault/src/hooks/useDashboardState.ts
```diff
@@ -152,16 +153,19 @@ export function useDashboardState(connectedAddress: string | undefined) {
     }
   }, [rawCollateralVaults, activatingVaults, clearActivatingVault]);
 
-  // Display-only BTC total: indexer collateral plus optimistic activating
-  // amounts. The financial `collateralBtc` (health factor / withdraw math)
-  // stays indexer/oracle-pure and is returned unchanged below.
+  // Add only activations that the current chain snapshot does not include.
   const displayCollateralBtc =
```
gbarkhatov: **A new divergence: the headline BTC total and the rows beneath it now come from different sources with no reconciliation.**

`displayCollateralBtc` is chain `totalCollateralBTC` plus activating entries, but `collateralVaults` is built purely from indexer `collaterals` (`:77-83`). A vault that is on-chain but not yet indexed — and no longer in the in-memory `activatingVaults` map, e.g. after a page reload — inflates the headline with no row beneath it.

Before this PR both halves came from one indexer payload and *could not* diverge. That is a real consequence of the source swap, and it is the kind of thing a balanced +349/−349 diff hides well.

The detector is already in hand and costs no RPC: `position.vaultIds` is the chain-authoritative set, so `vaultIds.length !== rawCollateralVaults.length` means the indexer is behind — which is also the cheapest way to produce the `ancillaryError` signal I mentioned on `Loans.tsx:122`.

Minor, same lines: this is the one derivation in the hook left outside `useMemo`. Every neighbour (`rawCollateralVaults`, `activatingEntries`, `collateralVaults`) is memoized, while this one runs a `filter` scanning `vaultIds` per entry on every render. Small n, but it is the rule's own file.

gbarkhatov (dismissed):
Reviewed at `7c076c8a` with three independent passes. This is the first PR in the current batch with real production logic on a money-adjacent path, so I verified every load-bearing claim against the worktree myself rather than relaying.

**Your two Safety invariants: the first holds, the second holds only for debt.**

*"Chain errors remain errors and cannot become an empty position"* — **confirmed**. `ts-sdk/tbv/integrations/aave/clients/query.ts:82-95` returns `null` only on a decoded `InvalidProxyContract` revert walked through the cause chain and rethrows everything else; `Promise.all` propagates; `positionService.ts:85` returns `[]` only for a genuine `null`. Clean.

*"Missing indexer details cannot hide debt"* — **holds for debt, fails for collateral and liquidation warnings.** That is the substance of the two existing threads, and both are worse than they were described.

---

### The two live threads are still standing, and both understate the problem

**`Loans.tsx:117` — the transient-poll-failure gate.** Confirmed at HEAD. `if (isConnected && (positionError || isLoading) && !godModeAffectsPage)` sits above the `hasPosition` branch at :135, and `useAaveUserPosition.ts:204` forwards `useQuery`'s `error` raw. The query runs on a 30 s `refetchInterval` with `networkMode: "always"` and no `throwOnError`/`placeholderData`, and React Query v5 retains `data` on a refetch failure while setting `error` — I had that checked against the reducer in the pinned `@tanstack/query-core@5.90.20` build (`query.js:355-368`: the `"error"` action spreads `...state` and never clears `data`), not asserted from memory. So one exhausted retry cycle swaps a working page for "Couldn't load your position" while `hasLoans`, `debtValueUsd` and `borrowedAssets` are all still correct in memory. That inverts this PR's stated goal and strands the user exactly as liquidation risk rises.

The cleanest fix is to gate on a *load* error rather than any error. `useDashboardState` does not currently export `position` (`:175-195`); exporting it and gating on `positionError && !position` is the precise form, and `hasCollateral || hasLoans` works as an in-scope proxy today.

**`positionService.ts:83` — the swallowed indexer error.** Confirmed, and materially worse than "no log, no UI signal". `.catch(() => [])` swallows everything — transport, GraphQL errors, the `BigInt()` casts in `fetchPositions.ts:171-172` — and `:129` then does `collaterals: indexedPosition?.collaterals ?? []`. I traced the consequence: `useDashboardState.ts:77-83` → empty `collateralVaults` → `usePositionNotifications.ts:124-133` filters to non-activating rows, hits `indexedVaults.length === 0`, and early-returns `status: "no-vaults"` — **above** the live-HF urgent guardrail at `:158-172`.

So for a position whose chain-derived debt and health factor the app knows perfectly well, the on-chain `healthFactor <= LIVE_HF_URGENT_THRESHOLD` warning never fires. The guardrail's own comment says it exists to "force one based on the on-chain oracle's health factor" when the calculator misses it — and this path makes it unreachable. Carrying the rejection out as a distinct `indexerError` instead of laundering it into `[]` fixes both halves.

I have not duplicated either thread inline; they carry the discussion.

---

### A third site the threads do not cover

`Detail/index.tsx:136` has the same defect on the Repay picker, with no existing thread. Inline below.

### Retry is now the only escape, and it can reject unhandled

`useAaveUserPosition.ts:205-211` wraps `refetch` to throw on `result.isError`, and `PositionGate.tsx:34` calls it as `onClick={() => void refetchPosition()}`. Against a persistent outage with `networkMode: "always"`, the button produces an unhandled rejection and no UI change — it looks dead. Pre-existing in `PositionGate`, but this PR newly routes the whole Loans page and the repay picker through it, so the blast radius is new.

---

### Checked and dismissed

- **"An empty or partial reserve list could read as no debt."** Disproved by two independent guards: `positionService.ts:105-119` throws when `borrowCount > 0n` with an empty list, and throws again when `debtPositions.size < borrowCount`; `AaveConfigContext.tsx:49-80` blocks render on config load/error, its comment citing exactly this scenario. Fails closed both ways.
- **"De-duplication by array position or render order."** Disproved — `useDashboardState.ts:160-165` matches `entry.vaultId` against `position.vaultIds` with case-insensitive string comparison, and `activatingEntries` dedupes the same way. Stable identity throughout.
- **"The reserve list is contract-backed."** Two of my three passes asserted this; they were wrong and I checked. `fetchConfig.ts:206` is `graphqlClient.request(GET_AAVE_APP_CONFIG)` — indexer-sourced GraphQL, not chain. The fail-closed conclusion still holds, but not for the reason two agents gave.
- **"+349/−349 hides a load-bearing removal."** Narrowed, not confirmed. No production logic was silently swapped — the fail-closed reserve guards are byte-identical and only their comment prose changed. What the balanced diff absorbed is **test coverage**; see the inline notes.
- **`data-testid` and copy.** Clean. No element carrying a testid was moved or re-implemented; `loans-borrow-button` / `loans-repay-button` live in the untouched `LoansSummary.tsx:63-64`. Both new strings are already in `copy.ts` (`positionLoadError` :1660, `retry` :1657).

### Verified clean

- Debt discovery is fully chain-sourced and fail-closed — `borrowCount` from Spoke, two throws guarding an incomplete probe. No indexer field reaches the debt or Repay path.
- Units match across the source swap: `totalCollateralBTC` is satoshis (`integrations/aave/types.ts:16-21`), the same 1e8 scale the indexer's `totalCollateral` fed through `satoshiToBtcNumber`. No silent rescaling.
- `position?.vaultIds.some(...)` cannot throw — `vaultIds` is required and always assigned by its sole producer.
- `{...indexedPosition}` is type-honest: the three real fields are explicitly reassigned after the spread, and spreading `undefined` lands nothing.

**Adjacent, not this PR:** `AaveConfigContext.tsx:71-77` inlines "Something went wrong" / "Retry" outside `copy.ts`; `getPosition`'s JSDoc (`ts-sdk/…/query.ts:62-64`) still says "prefer the indexer for position data", now contradicted by this PR's direction.

**Batch note:** #2465 also edits `LoanFlowOverlay.test.tsx` and `Loans.test.tsx`. Both branch from `main`, neither contains the other, and the `it.each` rewrite noted inline means that conflict will not resolve mechanically.

services/vault/src/applications/aave/services/positionService.ts
jonybur: Addressed in 38696a8846ce1c156ec8504487c7776d71b73279.

The service now returns a separate indexer error and logs it. Loans and the Repay picker show the existing warning. The Vaults page also warns when its rows are incomplete. Chain reads still determine debt and position existence.

services/vault/src/components/pages/Loans.tsx
jonybur: Addressed in 38696a8846ce1c156ec8504487c7776d71b73279.

The load-error gate now requires an absent position. A failed background read keeps the loaded debt and Repay action visible, with a warning. The new regression fails against the previous gate.

services/vault/src/applications/aave/components/Detail/index.tsx
jonybur: Addressed in 38696a8846ce1c156ec8504487c7776d71b73279.

The Repay picker now blocks only when the position is absent. A loaded debt row stays selectable after a background read fails. The regression checks that selection still opens the Repay form.

services/vault/src/components/pages/Loans.tsx
jonybur: Addressed in 38696a8846ce1c156ec8504487c7776d71b73279.

Loaded pages now pass the position or indexer error to the existing warning banner. The hard load-error state remains separate. No new warning component or loan copy was added.

services/vault/src/components/pages/__tests__/Loans.test.tsx
jonybur: Addressed in 38696a8846ce1c156ec8504487c7776d71b73279.

Replaced the table with named tests. Added failed-background-read and failed-Retry cases. Restored the disconnected-state assertion. Retry now handles rejection and disables the button during the read.

services/vault/src/applications/aave/services/__tests__/getUserPositionsWithLiveData.test.ts
jonybur: Addressed in 38696a8846ce1c156ec8504487c7776d71b73279.

Removed the dead batch setup. Restored a null reserve-probe case that must fail the incomplete-debt guard. Adapter and Spoke failures now have separate named tests.

services/vault/src/applications/aave/services/positionService.ts
jonybur: Addressed in 38696a8846ce1c156ec8504487c7776d71b73279.

Removed both unused timestamp fields and the indexer object spread. The returned position keeps only the fields its consumers use.

services/vault/src/hooks/useDashboardState.ts
jonybur: Addressed in 38696a8846ce1c156ec8504487c7776d71b73279.

The service compares active indexed vault IDs and amounts with the chain. It excludes removed history with the existing active-collateral filter. Mismatches produce a warning and disable incomplete cascade calculations. The displayed collateral total is memoized. Tests cover wrong IDs, stale amounts, and removed history.

jonybur:
@greptileai Please review commit 38696a8846ce1c156ec8504487c7776d71b73279.

Keep loaded debt and Repay after a read failure. Report incomplete indexed collateral, preserve live risk warnings, and restore deposit screenshots with a real archived response.

3,782 Vault tests pass, with 6 existing skips. Build and lint pass. All 18 visual cases pass and all 58 expected screenshots exist. Nine regression tests fail before the fixes. The separate full E2E typecheck still has the unchanged recovery-CLI blockNumber error noted in the PR body.

services/vault/src/applications/aave/components/Detail/index.tsx
```diff
@@ -131,23 +134,35 @@ export function LoanFlowOverlay({
...
-              pathname,
-              search: getReserveDetailSearch(selectedReserveId, mode),
-            },
-            { replace: true },
-          )
+      <PositionGate
+        positionError={
+          mode === LOAN_TAB.REPAY && isConnected && !position
+            ? positionError
+            : null
```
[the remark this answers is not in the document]
jonybur: Addressed in 399bad1db925add535ba426cde14bde312d0cb20.

The form now uses loaded debt to distinguish a failed background refresh from a failed initial load. Repay remains visible with a warning after asset selection. Borrow and empty positions retain the hard error block. The regression test fails with the previous gate and passes with the fix. All 174 loan-form and repayment tests pass. Max repayment still requires fresh debt and balance reads; typed repayment keeps its existing checks before signing.

jonybur:
@greptileai Please review commit 399bad1db925add535ba426cde14bde312d0cb20.

Keep the Repay form available when loaded debt remains after a background position read fails. Show the warning. Keep the hard error block for Borrow, missing debt, and empty positions. Leave the existing checks before signing unchanged.

All 174 loan-form and repayment tests pass. Build, lint, formatting, and diff checks pass. The Repay regression fails with the previous gate and passes with this fix. The change adds no net lines.

gbarkhatov (dismissed):

services/vault/src/applications/aave/components/Detail/ReserveDetailPanel.tsx
```diff
@@ -182,8 +179,10 @@ export function ReserveDetailPanel({
   return (
     <LoanProvider value={loanContextValue}>
       <PositionGate
-        positionError={positionError}
-        ancillaryError={ancillaryError}
+        positionError={
+          tab === LOAN_TAB.REPAY && totalDebtValueUsd > 0 ? null : positionError
```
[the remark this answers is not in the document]
jonybur: Reviewed at 2069f890127e00b821c297c53ec77ea87b75785e.

The exception now requires proven debt for the selected reserve. The existing aggregate-debt check also remains, so an empty position cannot expose the Borrow fallback. The regression covers selected debt of zero while another reserve has debt. It fails with the previous gate and passes with this fix. The positive selected-debt case still passes. All 174 related loan tests, build, focused lint, formatting, and diff checks pass.

jonybur:
@greptileai Please review commit 2069f890127e00b821c297c53ec77ea87b75785e.

Require loaded debt for the selected asset before keeping Repay open after a background position error. Keep the existing aggregate-debt check so an empty position cannot expose the Borrow fallback. Preserve the positive-debt regression.

All 3,782 Vault tests pass, with 6 existing skips. All 174 related loan tests also pass. Build, full lint, formatting, and diff checks pass. The selected-debt regression fails with the previous gate and passes with this fix.

gbarkhatov (dismissed):

jonybur:
@greptileai Please review commit 73bf96aab1652d21ffbea3fa000e2d72a9474f33.

Review the merge from main at 619ced7e. Both parent contributions are preserved. Production files match their contributing parent, so the merge adds no new production change. Shared tests retain the position/error coverage and the wallet audit assertions from #2465.

Merged Vault build and all 3,809 unit tests pass, with 6 existing skips. All 47 shared-file cases pass. Removing the three account-query gates fails exactly five audit cases; the full suite passes after byte-for-byte restoration. Lint passes with 0 errors and 134 existing warnings. Formatting and diff checks pass.

jonybur:
@greptileai

Please review the current commit `73bf96aab1652d21ffbea3fa000e2d72a9474f33`. The earlier request has not produced a review for this commit.

jonybur:
Visual report reviewed for `73bf96aab1652d21ffbea3fa000e2d72a9474f33`: [passing run](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34380043349).

Both sides passed all 329 Storybook cases and 18 Vault cases. All 329 Storybook images and 58 Vault images are present. No captures are missing or incomplete. All 42 deposit images compare unchanged, including the form, amount, and final progress screens.

The one reported difference is a broken external placeholder icon in the unchanged SmallBalance Storybook case. The source is unchanged in this PR. The capture evidence is consistent with an image-load failure, but it does not show the exact request failure. No Vault product defect was found in this review.

gbarkhatov (approved):

jonybur:
@greptileai

Please review the full PR at current commit `73bf96aab1652d21ffbea3fa000e2d72a9474f33`. Check the diff against its target branch. Report any remaining issues and the confidence score for this commit.

gbarkhatov (approved):

PR #2463, opened 2026-09-08
ci(workflows): run wallet and app browser checks

### LINKED ISSUE #2453: [eth-first] Run the pre-change browser and package checks in CI
## Outcome

Pull requests run the browser and package checks needed before Vault connection behavior changes.

Status: all gates pass at [ea328745](https://github.com/babylonlabs-io/babylon-toolkit/commit/ea3287458ecf08c2ba108989d3a11fb1286b2065). PR #2463 merged into main at [1d8560d17](https://github.com/babylonlabs-io/babylon-toolkit/commit/1d8560d17ca59f13e75fd71f04381d41e27e4807) on 9 September 2026. [Greptile gives 5/5](https://github.com/babylonlabs-io/babylon-toolkit/pull/2463#issuecomment-5593131191) for this commit. All review conversations are resolved. This issue is complete and closed.

| Gate | Evidence at ea328745 |
|---|---|
| Clean SDK and wallet package consumers | [Verify PR passes](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34387375632) |
| Browser journeys under normal CI load | Same run passes 2 wallet, 23 Vault, and 16 staking cases, plus layout and SDK browser checks |
| Unit tests and generated API documentation | Same run passes both gates |
| Reviewed visual comparisons | [Visual regression passes](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34387375577). Both sides have all 58 Vault images, including 42 deposit images, and all 329 Storybook images. No missing images or reported visual changes. Small raw pixel differences remain below the existing comparison threshold. |

The deposit screen remains present. The earlier missing desktop images came from a temporary viewport resize during capture. The repair preserves the full page and reports missing images as incomplete evidence. The upstream browser download failure is cleared. No workflow change was needed.

## Problem

Local test results do not show that browser checks pass under CI load. Earlier test repairs do not prove that all release gates run.

## Scope

- Run wallet connector, Vault, and simple staking browser tests on pull requests.
- Run the current full browser journeys under the normal CI load.
- Run clean production-install checks using the package checks owned by #2346 and #2360.
- Verify that Vault visual comparisons run and cover the changed screens. Use the screenshot repair in #2407.
- Link each required check and its passing run. Use the CI load repair in #2349, but verify the journeys themselves.

## Complete when

- [x] A passing PR run includes connector, Vault, and simple staking browser tests at `ea328745`.
- [x] The current full browser journeys pass under normal CI load at `ea328745`.
- [x] Clean production-install checks pass for the SDK and wallet connector at `ea328745`. See the verification run above.
- [x] Vault visual comparisons run and their results are reviewed at `ea328745`.
- [x] The issue records the tested revision, each gate, its workflow, and its passing CI link.

## Dependencies

Parent: #2233. Epic: #2228. Reuse package checks from completed #2346 and #2360. Test repairs: #2349 and #2407. Preparation does not wait for #2232, consent approval, or release owner assignment. This ticket closes only after its package and browser gates pass. Final Ethereum-first proof is in #2233.

jonybur:
## Outcome

PR verification runs wallet connector, Vault, and staking browser checks. A browser failure no longer skips unit tests or API documentation checks.

## Change

Build the wallet test page before CI opens it. Complete staking responses with real recordings and existing chain responses. Validate signer inputs and the Bitcoin balance. Restore query-error, Retry, and error-reporting coverage. Mark fulfilled reward requests as handled. Name the withdrawal fixture input and fee rate without changing its data.

Supply the existing Cosmos signer before OKX detection. Report required OKX setup failures with safe password errors. Wait for Keplr's completed setup screen, then close its tab.

During full-page screenshots, filter Chromium's temporary 1x1 resize until the viewport is restored. Keep the fixed clock and all content and layout checks. Check each capture against its own screen list so missing images are reported as incomplete.

## Safety

Wallet tests use fresh, unfunded wallets. Submissions are intercepted and checked. No real transactions are sent. Wallet traces, screenshots, and video are disabled. Seed and password input errors use safe messages without the original cause. App failure traces remain available.

The visual repair changes test and report code only. Screenshots retain all content below the viewport. Other resize events still reach the app. Failed screenshots clear the filter and preserve the original error. Pixel differences remain advisory.

## Verification

- [CI passes at ea328745](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34387375632). Both clean package consumers, all browser gates, affected unit tests, and generated API documentation pass.
- [Visual regression passes](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34387375577). Both sides have all 58 Vault images, including 42 deposit images, and all 329 Storybook images. No missing images or reported visual changes. Small raw pixel differences remain below the existing comparison threshold.
- [Greptile gives 5/5](https://github.com/babylonlabs-io/babylon-toolkit/pull/2463#issuecomment-5593131191) for this commit. All review conversations are resolved. The PR is approved.
- The fixture clarification preserves all exported data and transaction bytes. Four withdrawal browser cases and 371 unit tests pass locally, with one existing skip. It adds no net lines.

These checks cover the current flows. Final Ethereum-first release evidence remains in #2233.

## Links

Closes #2453.

Parent: #2233. Epic: #2228.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -89,14 +89,52 @@ jobs:
       - name: Lint all
         run: pnpm run lint
 
-      - name: Install Chromium for browser tests
-        run: pnpm --filter @babylonlabs-io/ts-sdk exec playwright install --with-deps chromium
+      - name: Install browsers for browser tests
+        id: browsers
+        run: pnpm --filter @babylonlabs-io/ts-sdk exec playwright install --with-deps chromium firefox
 
       - name: Check packed wallet consumer
         run: pnpm exec nx run @babylonlabs-io/wallet-connector:check:packed-consumer
 
       - name: Test SDK in Chromium
         run: pnpm exec nx run @babylonlabs-io/ts-sdk:test-browser
+
+      - name: Test wallet connections in Chromium
+        if: ${{ !cancelled() && steps.browsers.outcome == 'success' }}
+        env:
+          PLAYWRIGHT_HTML_OPEN: never
+        run: xvfb-run -a pnpm exec nx run @babylonlabs-io/wallet-connector:test:e2e
+
+      - name: Test Vault browsers
+        if: ${{ !cancelled() && steps.browsers.outcome == 'success' }}
+        env:
+          PLAYWRIGHT_HTML_OPEN: never
+        run: pnpm exec nx run @services/vault:test:e2e
+
+      - name: Build staking for browser tests
+        id: staking-browser-build
+        if: ${{ !cancelled() && steps.browsers.outcome == 'success' }}
+        env:
+          NEXT_BUILD_E2E: 'true'
+          DISABLE_SENTRY: 'true'
+        run: pnpm exec nx run @services/simple-staking:build:e2e
+
+      - name: Test staking browsers
+        if: ${{ !cancelled() && steps.staking-browser-build.outcome == 'success' }}
+        env:
+          PLAYWRIGHT_HTML_OPEN: never
+        run: pnpm exec nx run @services/simple-staking:test:e2e
+
+      - name: Upload app browser results
+        if: ${{ !cancelled() && steps.browsers.outcome == 'success' }}
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2 https://github.com/actions/upload-artifact/releases/tag/v4.6.2
+        with:
+          name: app-browser-results
+          path: |
+            services/vault/playwright-report
+            services/vault/test-results
+            services/simple-staking/playwright-report
+            services/simple-staking/test-results
 
       - name: Test affected
         id: tests
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -109,7 +109,7 @@ jobs:
         if: ${{ !cancelled() && steps.browsers.outcome == 'success' }}
         env:
           PLAYWRIGHT_HTML_OPEN: never
-        run: pnpm exec nx run @services/vault:test:e2e
+        run: pnpm exec nx run @services/vault:test:e2e --retries=0
 
       - name: Build staking for browser tests
         id: staking-browser-build
@@ -123,7 +123,7 @@ jobs:
         if: ${{ !cancelled() && steps.staking-browser-build.outcome == 'success' }}
         env:
           PLAYWRIGHT_HTML_OPEN: never
-        run: pnpm exec nx run @services/simple-staking:test:e2e
+        run: pnpm exec nx run @services/simple-staking:test:e2e --retries=0
 
       - name: Upload app browser results
         if: ${{ !cancelled() && steps.browsers.outcome == 'success' }}
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -12,6 +12,7 @@ concurrency:
 jobs:
   verify:
     runs-on: ubuntu-latest
+    timeout-minutes: 45
 
     steps:
       - name: Checkout code
@@ -137,6 +138,7 @@ jobs:
             services/simple-staking/test-results
 
       - name: Test affected
+        if: ${{ !cancelled() }}
         id: tests
         shell: bash
         run: |
@@ -166,6 +168,7 @@ jobs:
       # typedoc-plugin.mjs. MUST stay after "Build all": docs output depends on
       # the gitignored babylon-tbv-rust-wasm/dist.
       - name: Check generated API docs are current
+        if: ${{ !cancelled() }}
         shell: bash
         run: |
           set -o pipefail
--- a/packages/babylon-wallet-connector/playwright.config.ts
+++ b/packages/babylon-wallet-connector/playwright.config.ts
@@ -35,8 +35,7 @@ export default defineConfig({
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
     baseURL,
-    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: "on-first-retry",
+    trace: "off",
     locale: "en-US",
     launchOptions: {
       args: ["--lang=en-US", "--force-lang=en-US", "--accept-lang=en-US"],
@@ -76,9 +75,11 @@ export default defineConfig({
   ],
   /* Run your local dev server before starting the tests */
   webServer: {
-    command: "npm run dev",
-    url: baseURL,
-    timeout: 120 * 1000,
+    command: process.env.CI
+      ? "pnpm run build-storybook && pnpm exec vite preview --outDir storybook-static --port 6006 --strictPort"
+      : "npm run dev",
+    url: `${baseURL}/iframe.html`,
+    timeout: 300_000,
     reuseExistingServer: !process.env.CI,
   },
 });
--- a/packages/babylon-wallet-connector/playwright.config.ts
+++ b/packages/babylon-wallet-connector/playwright.config.ts
@@ -28,9 +28,9 @@ export default defineConfig({
   /* Retry on CI only */
   retries: 0,
   /* Opt out of parallel tests on CI. */
-  workers: 2,
+  workers: process.env.CI ? 1 : 2,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: "html",
+  reporter: "line",
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
@@ -60,6 +60,9 @@ export default defineConfig({
       testIgnore: "**/specs/wallets/**",
       use: {
         ...devices["Desktop Chrome"],
+        trace: "off",
+        screenshot: "off",
+        video: "off",
         locale: "en-US",
         launchOptions: {
           args: ["--lang=en-US", "--force-lang=en-US", "--accept-lang=en-US"],
@@ -76,6 +79,6 @@ export default defineConfig({
     command: "npm run dev",
     url: baseURL,
     timeout: 120 * 1000,
-    reuseExistingServer: true,
+    reuseExistingServer: !process.env.CI,
   },
 });
--- a/packages/babylon-wallet-connector/tests/e2e/fixtures/setupExtensions.ts
+++ b/packages/babylon-wallet-connector/tests/e2e/fixtures/setupExtensions.ts
@@ -1,84 +1,44 @@
-import { test as base, type BrowserContext, chromium } from "@playwright/test";
+import { test as base, type BrowserContext } from "@playwright/test";
+import { english, generateMnemonic, generatePrivateKey } from "viem/accounts";
 
-import { EXTENSION_CHROME_STORE_IDS, getExtensionPath } from "../setup/downloadExtensions";
-
+import { launchWalletContext, type SupportedWallet } from "./launch";
 import { setupKeplrWallet } from "./wallets/keplr";
 import { setupOKXWallet } from "./wallets/okx";
 
-type SupportedWallets = keyof typeof EXTENSION_CHROME_STORE_IDS;
-
 type ExtensionSetup = {
-  context: BrowserContext;
-  setupExtensions: (extensions: SupportedWallets[]) => Promise<{
+  setupExtensions: (extensions: SupportedWallet[]) => Promise<{
     context: BrowserContext;
   }>;
 };
 
 export const test = base.extend<ExtensionSetup>({
-  context: async ({}, use) => {
-    const context = await chromium.launchPersistentContext("", {
-      headless: false,
-      channel: "chromium",
-      locale: "en-US",
-      args: ["--lang=en-US", "--force-lang=en-US", "--accept-lang=en-US"],
-      permissions: ["clipboard-read", "clipboard-write"],
-    });
-    await use(context);
-    await context.close();
-  },
-
   setupExtensions: async ({}, use) => {
-    const setup = async (extensions: SupportedWallets[]) => {
-      // Create new context with all extensions loaded
-      const extensionPaths = extensions.map((ext) => getExtensionPath(EXTENSION_CHROME_STORE_IDS[ext]));
-      const newContext = await chromium.launchPersistentContext("", {
-        headless: false,
-        channel: "chromium",
-        locale: "en-US",
-        args: [
-          `--disable-extensions-except=${extensionPaths.join(",")}`,
-          `--load-extension=${extensionPaths.join(",")}`,
-          "--lang=en-US",
-          "--force-lang=en-US",
-          "--accept-lang=en-US",
-        ],
-        permissions: ["clipboard-read", "clipboard-write"],
-      });
-
-      // Open a blank page
-      const blankPage = await newContext.newPage();
-      await blankPage.goto("about:blank");
-
-      // Wait 5 seconds
-      await new Promise((resolve) => setTimeout(resolve, 5000));
-
-      // Close all other pages
-      for (const page of newContext.pages()) {
-        if (page !== blankPage) {
-          await page.close();
-        }
-      }
-
-      // Setup wallets one at a time
-      for (const ext of extensions) {
-        try {
-          if (ext === "OKX") {
-            await setupOKXWallet(newContext, process.env.E2E_WALLET_MNEMONIC!, process.env.E2E_WALLET_PASSWORD!);
-            console.log("OKX wallet setup completed");
-          } else if (ext === "KEPLR") {
-            await setupKeplrWallet(newContext, process.env.E2E_WALLET_MNEMONIC!, process.env.E2E_WALLET_PASSWORD!);
-            console.log("Keplr wallet setup completed");
+    const contexts: BrowserContext[] = [];
+    try {
+      await use(async (extensions) => {
+        const context = await launchWalletContext(extensions);
+        contexts.push(context);
+        const mnemonic = generateMnemonic(english);
+        const password = generatePrivateKey().slice(2, 34);
+        for (const ext of extensions) {
+          try {
+            if (ext === "OKX") {
+              await setupOKXWallet(context, mnemonic, password);
+            } else if (ext === "KEPLR") {
+              await setupKeplrWallet(context, mnemonic, password);
+            }
+          } catch (error) {
+            const reason = [mnemonic, password, ...mnemonic.split(" ")].reduce(
+              (message, secret) => message.replaceAll(secret, "[redacted]"),
+              error instanceof Error ? error.message : "Unknown setup error",
+            );
+            throw new Error(`${ext} wallet setup failed: ${reason}`);
           }
-        } catch (error) {
-          console.error(`Failed to setup ${ext} wallet:`, error);
-          await newContext.close();
-          throw error;
         }
-      }
-
-      return { context: newContext };
-    };
-
-    await use(setup);
+        return { context };
+      });
+    } finally {
+      await Promise.all(contexts.map((context) => context.close()));
+    }
   },
 });
--- a/packages/babylon-wallet-connector/tests/e2e/fixtures/setupExtensions.ts
+++ b/packages/babylon-wallet-connector/tests/e2e/fixtures/setupExtensions.ts
@@ -28,7 +28,7 @@ export const test = base.extend<ExtensionSetup>({
               await setupKeplrWallet(context, mnemonic, password);
             }
           } catch (error) {
-            const reason = [mnemonic, password, ...mnemonic.split(" ")].reduce(
+            const reason = [mnemonic, password].reduce(
               (message, secret) => message.replaceAll(secret, "[redacted]"),
               error instanceof Error ? error.message : "Unknown setup error",
             );
--- a/services/simple-staking/e2e/constants/staking.ts
+++ b/services/simple-staking/e2e/constants/staking.ts
@@ -2,5 +2,3 @@ import { satoshiToBtc } from "@/ui/common/utils/btc";
 
 export const STAKING_AMOUNT_SAT = 50000;
 export const STAKING_AMOUNT_BTC = satoshiToBtc(STAKING_AMOUNT_SAT);
-export const STAKING_TX_HASH =
-  "47af61d63bcc6c513561d9a1198d082052cc07a81f50c6f130653f0a6ecc0fc1";
--- a/services/simple-staking/e2e/constants/staking.ts
+++ b/services/simple-staking/e2e/constants/staking.ts
@@ -1,4 +1,4 @@
-import { satoshiToBtc } from "@/ui/legacy/utils/btc";
+import { satoshiToBtc } from "@/ui/common/utils/btc";
 
 export const STAKING_AMOUNT_SAT = 50000;
 export const STAKING_AMOUNT_BTC = satoshiToBtc(STAKING_AMOUNT_SAT);
--- a/services/simple-staking/e2e/fixtures/wallet_connect.ts
+++ b/services/simple-staking/e2e/fixtures/wallet_connect.ts
@@ -1,8 +1,13 @@
 import type { Page } from "@playwright/test";
 
-import { injectBBNWallet, injectBTCWallet } from "../mocks/blockchain";
+import {
+  injectBBNWallet,
+  injectBTCWallet,
+  type TestWalletOptions,
+} from "../mocks/blockchain";
 import { mockVerifyBTCAddress } from "../mocks/handlers";
 
+import { PageNavigationActions } from "./page_navigation";
 import {
   BUTTON_SELECTORS,
   CHECKBOX_SELECTOR,
@@ -210,10 +215,11 @@ export class WalletConnectActions {
     }
   }
 
-  async setupWalletConnection() {
+  async setupWalletConnection(options?: TestWalletOptions) {
+    await new PageNavigationActions(this.page).waitForPageLoad();
     await this.setupMocks();
-    await injectBTCWallet(this.page);
-    await injectBBNWallet(this.page);
+    await injectBTCWallet(this.page, "OKX", options);
+    await injectBBNWallet(this.page, "Leap", options);
     await this.clickConnectButton();
     await this.acceptTermsAndConditions();
     await this.clickInjectableWalletButton();
--- a/services/simple-staking/e2e/fixtures/wallet_connect.ts
+++ b/services/simple-staking/e2e/fixtures/wallet_connect.ts
@@ -10,7 +10,6 @@ import { mockVerifyBTCAddress } from "../mocks/handlers";
 import { PageNavigationActions } from "./page_navigation";
 import {
   BUTTON_SELECTORS,
-  CHECKBOX_SELECTOR,
   CONNECT_BUTTON_SELECTOR,
   DIALOG_SELECTORS,
   WALLET_SELECTORS,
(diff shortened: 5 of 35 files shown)


services/simple-staking/e2e/mocks/blockchain/bbnWallet.ts
```diff
@@ -7,16 +9,53 @@ import { verifyBBNWalletInjected } from "./verification";
...
+      ...data,
+      bbnWallet: {
+        ...data.bbnWallet,
+        walletAddress: account.address,
+        pubkeyArray: Array.from(account.pubkey),
+      },
+    };
+    await page.exposeFunction(
+      "e2eSignDirect",
+      async (address: string, doc: any) => {
```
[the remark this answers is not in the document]
jonybur: Fixed in 6c458d73f53f4a1f061148b64e86fedcdbf5fb17. The signing bridge uses explicit serialized document, response, and browser types. Strict type checks pass, including invalid-input type checks. All 16 staking browser cases pass without retries.

.github/workflows/verify.yml
```diff
@@ -89,15 +89,53 @@ jobs:
...
+        env:
+          NEXT_BUILD_E2E: 'true'
+          DISABLE_SENTRY: 'true'
+        run: pnpm exec nx run @services/simple-staking:build:e2e
+
+      - name: Test staking browsers
+        if: ${{ !cancelled() && steps.staking-browser-build.outcome == 'success' }}
+        env:
+          PLAYWRIGHT_HTML_OPEN: never
+        run: pnpm exec nx run @services/simple-staking:test:e2e
```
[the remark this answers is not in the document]
jonybur: Fixed in 6c458d73f53f4a1f061148b64e86fedcdbf5fb17. Both CI browser commands now pass --retries=0. Workflow validation passes. All 16 staking browser cases pass with CI settings and no retries.

jonybur:
@greptileai Please review commit 6c458d73f53f4a1f061148b64e86fedcdbf5fb17.

The browser signer bridge now uses explicit document and response types. Vault and staking CI commands disable retries.

Verification: all 16 staking browser tests pass across Chromium and Firefox without retries. Strict signer type checks, lint, workflow validation, formatting, and diff checks pass.

services/simple-staking/e2e/constants/staking.ts
```diff
@@ -1,4 +1,4 @@
-import { satoshiToBtc } from "@/ui/legacy/utils/btc";
+import { satoshiToBtc } from "@/ui/common/utils/btc";
```
jrwbabylonlab, line 1: `STAKING_TX_HASH` lost its last consumer in this PR — dead code now, can it be deleted?

.github/workflows/verify.yml
```diff
@@ -89,15 +89,53 @@ jobs:
...
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2 https://github.com/actions/upload-artifact/releases/tag/v4.6.2
+        with:
+          name: app-browser-results
+          path: |
+            services/vault/playwright-report
+            services/vault/test-results
+            services/simple-staking/playwright-report
+            services/simple-staking/test-results
+
       - name: Test affected
```
gbarkhatov, line 140: **Merge blocker, and the most consequential thing I found — it is not about the browser tests at all.**

Every new browser step (102, 108, 116, 122, 128) correctly carries `if: ${{ !cancelled() && … }}`. But `Test affected` on this line has **no `if:`**, and neither does `Check generated API docs are current` at line 168 — so both inherit the default `success()`.

The browser gates were inserted *above* them. So any browser failure makes the job unsuccessful and **silently skips the entire unit-test run and the docs-staleness gate.** That is exactly what happened in this run: the log jumps from the staking failure straight to `Post job cleanup`. `nx affected --target=test` and `docs:clean` never executed.

That inverts the intent of the change — adding browser coverage should not be able to remove unit coverage. And CLAUDE.md singles out the docs-staleness gate as the one that is easy to miss locally precisely because it is not part of `build`, `lint` or `test`; this makes it easy to miss in CI too, so a stale-docs PR could go green-ish while the real gate never ran.

Either add `if: ${{ !cancelled() }}` to both steps, or move the three browser gates below them.

packages/babylon-wallet-connector/playwright.config.ts
```diff
@@ -76,6 +79,6 @@ export default defineConfig({
     command: "npm run dev",
     url: baseURL,
     timeout: 120 * 1000,
```
gbarkhatov: **Merge blocker — this line is the root cause of Failure A.**

`url: baseURL` is the Storybook *manager* root. Playwright treats the server as ready the moment :6006 binds, but `command: "npm run dev"` is Storybook's Vite dev builder, which compiles the **preview** bundle on demand. `connectOKXKeplr.spec.ts` navigates straight to `/iframe.html?id=…`, so the very first spec waits on a cold compile inside its own 120 s test budget.

The CI timeline matches exactly: test starts 23:41:38, `[WebServer]` still emitting Babel output at 23:43:37 ("core-ui/dist/index.es.js exceeds 500KB"), test dies 23:43:39 — start + 120,000 ms to the second. `mnemonic.spec.ts` then passes in 58 s against the warm bundle, and `workers: 1` on CI serialises them so only the first spec ever pays.

This also explains why you cannot reproduce locally: `reuseExistingServer: !process.env.CI` is true off-CI, so a developer with Storybook already running never pays the compile at all. The flip on the line below is correct hygiene and is not itself the problem.

Two fixes, either works:

```ts
url: `${baseURL}/iframe.html`,   // readiness probe hits the preview
```

or build `build-storybook` output once and serve it statically in CI, which also removes the 120 s cliff entirely.

services/simple-staking/package.json
```diff
@@ -19,7 +19,7 @@
     "sort-imports": "eslint --fix .",
     "test": "jest",
     "test:watch": "jest --watch",
-    "test:e2e": "playwright test e2e/specs/",
+    "test:e2e": "playwright test",
```
gbarkhatov, line 22: **Merge blocker (shared with `wallet_connect.ts:219`) — this widening is what pulled Failure B into the gate.**

`"playwright test e2e/specs/"` → `"playwright test"` adds `unbonding.spec.ts` — **three of the four staking failures** — to CI for the first time.

The underlying defect is pre-existing, not yours: `injectBBNQueries`' catch-all `page.route("**")` calls `route.continue()` on unmocked GETs (`/v2/apr`, `/v2/delegation`, `v1/fees/recommended`), `vite preview` 404s them, react-query retries ~59 s and raises a blocking "System Error" dialog. `acceptTermsAndConditions` (`wallet_connect.ts:51`) then does a **non-waiting** `.isVisible()`, falls into `handleAlternativeDialog`, and blind-clicks the dialog's first button — which on an error modal is "Copy error details", a no-op. Ten seconds later `clickInjectableWalletButton` times out at line 113.

All four `error-context.md` snapshots in the run artifact show that same modal, and all 8 Firefox runs of these specs passed — so this is a timing race, not a broken flow.

Two things to fix before this glob widens: make `unbonding.spec.ts` and `balanceAddress.spec.ts` hermetic the way `staking.spec.ts` already is (abort-and-record unrouted requests instead of `continue()`), and make the terms check actually wait rather than using an unwaited `.isVisible()`.

services/simple-staking/e2e/fixtures/wallet_connect.ts
```diff
@@ -210,10 +215,11 @@ export class WalletConnectActions {
     }
   }
 
-  async setupWalletConnection() {
+  async setupWalletConnection(options?: TestWalletOptions) {
+    await new PageNavigationActions(this.page).waitForPageLoad();
```
gbarkhatov, line 125: This added `waitForPageLoad()` is the second half of Failure B, and it is what turned an intermittent race into a near-certain one.

It is a `networkidle` wait, and networkidle cannot settle while react-query is retrying 404s for ~59 s. So instead of racing the error modal, the flow now reliably waits *through* the retry storm and arrives after the modal has already rendered.

The wait itself is reasonable — the problem is what it is waiting on. Once the unrouted endpoints are mocked (see my note on `package.json:22`), this becomes harmless and useful. Worth landing in that order rather than removing it.

services/simple-staking/e2e/mocks/blockchain/btcWallet.ts
```diff
@@ -138,7 +198,7 @@ export const injectBTCWallet = async (
                   case "btc_getBalance":
                     return btcData.balance;
                   case "btc_signPsbt":
-                    return methodParams?.psbt || btcData.simplifiedPsbt;
+                    return btcWallet.signPsbt(methodParams?.psbt);
```
gbarkhatov: **The `any` fix from the earlier thread is partial — this added line routes an untyped value straight into the signer the PR just typed.**

The three sites named in that thread genuinely are fixed: `bbnWallet.ts:40` (`SerializedSignDoc`), `:80-83` (`DirectSignResponse`), and `btcWallet.ts:45`, with shapes in `e2e/types.d.ts`. Good.

But line 185 still declares `request: async (params: { method: string; params?: any })`, and this line — an added line in this PR — does `btcWallet.signPsbt(methodParams?.psbt)`. So `methodParams` is `any`, and `?.psbt` is unchecked, feeding the newly-typed `signPsbt` an unvalidated value. CLAUDE.md's rule is `unknown` plus narrowing, not `any`.

Narrow it at the case, which also documents the wire shape:

```ts
case "btc_signPsbt": {
  const { psbt } = methodParams as { psbt: string };
  return btcWallet.signPsbt(psbt);
}
```

Flagging it here rather than on the original thread since that thread is marked fixed and this is a different line.

services/vault/e2e/high-severity-errors.spec.ts
```diff
@@ -1,229 +1,29 @@
...
-    });
-  });
+import { installRecordedBackend } from "./fixtures/replay";
+
+test("app loads without error UI or unhandled errors with the recorded backend", async ({
+  page,
+}) => {
+  const pageErrors: Error[] = [];
+  page.on("pageerror", (error) => pageErrors.push(error));
+  const backend = await installRecordedBackend(page);
```
gbarkhatov, line 11: This file goes 229 → 29 lines, and two behaviours leave with no replacement anywhere in the repo:

- *"should show error UI when applications query fails"*
- *"should send Sentry event when applications query fails"*

Grepping `Failed to load applications` across `services/vault` returns nothing, so the GraphQL-failure error UI and its Sentry reporting are now untested repo-wide.

Much of what else went was genuinely dead or superseded, and I am not asking for all 229 lines back. But deleting behavioural coverage inside a PR titled "run wallet and app browser checks" is the kind of thing that is invisible in review — the diff reads as test cleanup while the net effect is less coverage than before. Either keep these two cases or call the removal out in the description so it is a decision rather than a side effect.

packages/babylon-wallet-connector/tests/e2e/fixtures/setupExtensions.ts
```diff
@@ -1,84 +1,44 @@
...
+        const password = generatePrivateKey().slice(2, 34);
+        for (const ext of extensions) {
+          try {
+            if (ext === "OKX") {
+              await setupOKXWallet(context, mnemonic, password);
+            } else if (ext === "KEPLR") {
+              await setupKeplrWallet(context, mnemonic, password);
+            }
+          } catch (error) {
+            const reason = [mnemonic, password, ...mnemonic.split(" ")].reduce(
```
gbarkhatov: Redacting every individual mnemonic word corrupts the diagnostics this is meant to sanitise.

`[mnemonic, password, ...mnemonic.split(" ")].reduce((m, s) => m.replaceAll(s, "[redacted]"), …)` replaces each word **in isolation**, and the BIP-39 English list contains ordinary words: `seed`, `grid`, `select`, `close`, `target`, `input`, `problem`, `report`.

So `okx.ts`'s carefully written *"OKX: seed grid shows 12 inputs but the phrase has 24 words…"* can come back as *"OKX: [redacted] [redacted] shows 12 [redacted]s…"* — and it degrades *more* the more helpful the message is.

One word out of a twelve-word phrase is not recoverable secret material. Redacting the full mnemonic and the password is the real protection:

```ts
[mnemonic, password].reduce((m, s) => m.replaceAll(s, "[redacted]"), message)
```

The intent is right — this matters more than usual given how little diagnostic output the wallet-connector suite produces (see the trace note in my review body).

services/simple-staking/e2e/specs/balanceAddress.spec.ts
```diff
@@ -27,8 +27,8 @@ test.describe("Balance and address checks after connection", () => {
     const stakableBalance = await balanceActions.getStakableBalance();
     const babylonBalance = await balanceActions.getBabylonBalance();
 
-    expect(stakedBalanceText).toContain("0.09876543 sBTC");
-    expect(stakableBalance).toContain("0.00074175 sBTC");
-    expect(babylonBalance).toContain("1 tBABY");
+    expect(stakedBalanceText).toContain("0.09876543 BTC");
+    expect(stakableBalance).toContain("<0.01 BTC");
```
gbarkhatov: This assertion got weaker in a way that is easy to miss: `"0.00074175 sBTC"` → `"<0.01 BTC"`.

`<0.01 BTC` is the UI's below-threshold placeholder, so **any** balance under the threshold now satisfies it — including a wrong one, and including zero if the mock ever stops applying. The adjacent line still asserts a full 8-decimal value, so the file now checks the same class of thing at two very different strengths.

Since this is a mock, the cleanest fix is to raise the mocked balance above the display threshold and keep asserting the exact rendered figure. If the placeholder path is itself what you want to pin, that is fine — but it deserves its own named test rather than replacing the precise assertion in this one.

gbarkhatov (dismissed):
Reviewed at `6c458d73` with three independent passes, focused on the red `verify`. Your description already says passing browser gates are outstanding, so this is "here is why", not "you missed something".

## The red CI is two different problems, not one

**Failure A (wallet-connector) is caused by this PR.** `webServer.url` points at the Storybook manager root, which binds as soon as :6006 is up — but the spec navigates to `/iframe.html?id=…`, which needs the preview bundle Vite compiles *on demand*. The log timeline is exact: test starts 23:41:38, `[WebServer]` is still emitting Babel output at 23:43:37, test dies at 23:43:39 = start + 120,000 ms. `mnemonic.spec.ts` then passes in 58 s against the now-warm bundle, because `workers: 1` serialises them. **This is also the local-vs-CI divergence in your body**: locally `reuseExistingServer` is true, so a developer with Storybook already running never pays the cold compile.

**Failure B (staking) is pre-existing non-hermeticity that this PR made blocking.** All four `error-context.md` snapshots in the run artifact show the same blocking dialog — `heading "System Error"`, *"Error fetching data from the API"* — over a page where the chain picker never opened. `injectBBNQueries`' catch-all `page.route("**")` calls `route.continue()` on unmocked GETs, `vite preview` 404s them, and react-query retries ~59 s before raising the modal. Decisive evidence that it is not structural: **all 8 Firefox runs of the same specs passed in the same job** (two projects × 8 specs = the 16 in the log). A broken selector cannot pass under one engine.

### Both hypotheses I went in with were wrong, and I want to record that

- **No `data-testid` was dropped.** `select-bitcoin-wallet-button` is still emitted at `ChainButton/index.tsx:18`, and its selector constant is unchanged. The only edit in `wallet_connect.selectors.ts` is `DONE`, and it is a *repair*: `chains-connect-button` is what `Chains/index.tsx:81` renders; `chains-done-button` exists nowhere.
- **The extension setup did not die.** All five extensions downloaded and unpacked (`✓ Downloaded MetaMask` at 23:41:37.5), and `mnemonic.spec.ts` passed immediately after on the same fixture. `"Target page, context or browser has been closed"` is the fixture's `finally { context.close() }` firing *after* the 120 s timeout — the symptom, not the cause.
- **Disabling retries did not surface Failure A.** `packages/babylon-wallet-connector/playwright.config.ts:29` was **already** `retries: 0` before this PR.

## Live threads

**Thread "Browser failures are retried" — genuinely fixed, and completely.** `--retries=0` is on both app commands, and the wallet-connector config was already `retries: 0`, which is why your reply mentions only two. **No gate is left retry-masked.** One residue, inline below: with retries forced to 0, `trace: "on-first-retry"` never fires.

**Thread "Signer bridge bypasses typing" — the fix is PARTIAL.** Two of my three passes called it complete; they were wrong and I checked myself. Details inline on `btcWallet.ts:201`.

**Thread `STAKING_TX_HASH` — confirmed dead.** Repo-wide grep returns only the declaration at `services/simple-staking/e2e/constants/staking.ts:5`; its last consumer went with the `staking.spec.ts` rewrite. `STAKING_AMOUNT_SAT`/`_BTC` are still used, so the file stays — just the one constant. It is outside the diff hunk so I have not posted inline; answering here since that thread is still open.

## Two more that could not be anchored inline

- **`BUTTON_SELECTORS.NEXT` targets a testid no component renders.** `wallet_connect.selectors.ts:13` declares `terms-next-button`; repo-wide grep returns exactly one hit — that declaration. So the *happy* path of `acceptTermsAndConditions` clicks a locator that never resolves; only the `handleAlternativeDialog` fallback has ever worked, which is what makes Failure B's blind-click possible. The sibling `DONE` selector was repaired in this PR and `NEXT` was missed.
- **`services/vault/vite.config.ts:96`** — `cacheDir: process.env.PLAYWRIGHT_VITE_CACHE_DIR` is ungated by `mode`, putting a test-harness knob in the production build config.
- **`staking.spec.ts:118-129`** puts ~15 `expect()` calls inside a `page.route` handler, so a failure hangs the request and surfaces ~90 s later as a poll timeout with the real assertion buried.
- The `verify` job has no `timeout-minutes`; it already runs 22m14s.

## Verified clean

- **Your "fresh, unfunded" claim holds.** `setupExtensions.ts:21-22` generates `generateMnemonic(english)` and a random password per run; the staking key is `"4".padStart(64,"0")`, a well-known throwaway. No `console.log` of key or payload bytes in any changed fixture.
- Gates can genuinely go red — no `continue-on-error`, no `|| true`; the log ends `Process completed with exit code 1`.
- Browser install is correct and ordered: `playwright install --with-deps chromium firefox` before all three suites, each gated on `steps.browsers`.
- The `@/ui/legacy` → `@/ui/common` import repairs were **necessary** — those specs did not resolve on `main`. They were dead code, which is exactly why nobody noticed they had never run in CI.
- No new or unpinned dependencies; `services/simple-staking/package.json` changes only the `test:e2e` script.
- Removed mock fields are fully cleaned up — zero dangling references in `e2e/`.

**Cross-PR:** #2459 also edits `.github/workflows/verify.yml` from `main`. These will conflict, and whichever merges second must re-add its steps *and* re-check the step-ordering issue below.

services/simple-staking/e2e/constants/staking.ts
jonybur: Addressed in 2b492fb22bac03ed4b7b21954340ec36fa187b11.

Removed the unused transaction-hash constant. No consumers remain.

.github/workflows/verify.yml
jonybur: Addressed in 2b492fb22bac03ed4b7b21954340ec36fa187b11.

Unit tests and API documentation checks now use the same cancellation guard as the browser steps. A browser failure cannot skip them. Added a 45-minute job limit. Workflow validation passes.

packages/babylon-wallet-connector/playwright.config.ts
jonybur: Addressed in 2b492fb22bac03ed4b7b21954340ec36fa187b11.

CI now builds and serves static Storybook before checking iframe readiness. This also removes the first-navigation compilation race. Both wallet cases pass from a fresh build without retries.

services/simple-staking/package.json
jonybur: Addressed in 2b492fb22bac03ed4b7b21954340ec36fa187b11.

Kept all staking cases in the gate. Shared setup now supplies the existing fees and Tip responses plus three real APR recordings for the exact requested inputs. Unknown API requests fail the test. All 16 cases pass without retries.

services/simple-staking/e2e/fixtures/wallet_connect.ts
jonybur: Addressed in 2b492fb22bac03ed4b7b21954340ec36fa187b11.

The missing API responses caused the wait to stall. Shared routing now covers them and rejects unknown requests. Removed obsolete terms, Save, and error-dialog handling. Kept the page-load wait after confirming all 16 staking cases pass.

services/simple-staking/e2e/mocks/blockchain/btcWallet.ts
jonybur: Addressed in 2b492fb22bac03ed4b7b21954340ec36fa187b11.

Request parameters now use unknown. The bridge checks for an object with a string psbt before calling the signer. Strict signer type checks and all staking browser cases pass.

services/vault/e2e/high-severity-errors.spec.ts
jonybur: Addressed in 2b492fb22bac03ed4b7b21954340ec36fa187b11.

The old applications-query caller no longer renders that error UI. Added a failure case for the current GetAaveAppConfig path instead. It verifies the error message, Retry control, and Sentry event. All 23 Vault browser cases pass.

packages/babylon-wallet-connector/tests/e2e/fixtures/setupExtensions.ts
jonybur: Addressed in 2b492fb22bac03ed4b7b21954340ec36fa187b11.

Removed individual-word replacement. Full mnemonic and password redaction remains. Both wallet importers already replace seed-input errors with safe messages. Both onboarding tests pass.

services/simple-staking/e2e/specs/balanceAddress.spec.ts
jonybur: Addressed in 2b492fb22bac03ed4b7b21954340ec36fa187b11.

The assertion now uses the existing confirmed balance and checks the exact display, 0.12345678 BTC. It can detect decimal or balance-source errors. Both browser projects pass.

jonybur:
@greptileai Please review commit 2b492fb22bac03ed4b7b21954340ec36fa187b11.

Keep core checks running after browser failures. Use a built wallet test page, complete recorded request coverage, typed signer input, safe diagnostics, and exact balance checks.

All 41 browser cases pass with no retries. Connector 295, Vault 3,756, and staking 371 unit tests pass; 23 package checks pass. Build, lints, signer/Vite types, workflow validation, and formatting pass.

gbarkhatov (dismissed):

services/simple-staking/e2e/mocks/blockchain/bbnQueries.ts
```diff
@@ -341,7 +442,7 @@ export const injectBBNQueries = async (
     }
 
     try {
-      await route.continue();
+      await route.fallback();
     } catch (error) {}
```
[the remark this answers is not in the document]
jonybur: Reviewed at f63e8556a5f95da46396f1d523c82d49d4931a8a.

Removed the empty catch. The browser run then exposed a request handled twice: RewardGauges fulfilled its request but returned undefined, so the dispatcher also called fallback. It now awaits fulfillment and returns true, like the existing bank handler. All 16 staking browser cases pass with no retries. All 371 unit tests pass, with one existing skip. Lint, strict type, formatting, and diff checks pass.

jonybur:
@greptileai Please review commit f63e8556a5f95da46396f1d523c82d49d4931a8a.

Mark fulfilled reward requests as handled and let routing errors fail the test. Supply the existing Cosmos signer before OKX wallet detection to fix the four CI wallet setup failures.

All 16 staking browser cases pass with no retries. All 371 unit tests pass, with one existing skip. Lint, strict signer type checks, formatting, and diff checks pass. The final patch adds no net lines.

gbarkhatov (dismissed):

jonybur:
@greptileai Please review commit ea5d4464171c974a7b68880dc6bdd53c91ea2bf1.

Follow-up to [the OKX setup finding](https://github.com/babylonlabs-io/babylon-toolkit/pull/2463#issuecomment-5593131191) at `f63e8556a5f95da46396f1d523c82d49d4931a8a`.

The OKX setup now reports failures in required navigation, seed confirmation, password selection and input, primary buttons, welcome checkboxes, and the final home wait. Missing password frames fail explicitly. Password input uses a fixed error without the original cause. A real disabled-field check confirms that the reported error contains no password or recovery phrase. Optional tab and popup cleanup stays separate.

The final patch also fixes the Keplr failure from CI. Finish only closes the completed setup page. The fixture now waits for that screen and closes it through Playwright, so tab closure cannot interrupt the button click. Both real connector cases pass locally without retries or secret-bearing artifacts. The PR body records the CI result and final checks.

The final combined patch passes both real wallet browser cases without retries, connector lint, strict helper types, changed-line formatting, and diff checks. The unchanged OKX patch also passed connector build, 23 package checks, 295 unit tests, and a real password-error redaction probe. CI at f63e8556 passed all other verification steps but hit the Keplr Finish close race. A new CI run must confirm the repair.

jonybur:
@greptileai

Please review the current commit `ea5d4464171c974a7b68880dc6bdd53c91ea2bf1`. The earlier request has not produced a review for this commit.

jonybur:
@greptileai Please review commit ad0f0d0f92b55a8f2aa5103a0bc5ea2e45b19ebc.

Repair the visual capture failure. Filter only Chromium's temporary 1x1 resize during a full-page screenshot and viewport restoration. Keep the fixed clock, complete page pixels, and all layout and content guards. Clear the filter on failure and preserve the error. Compare each side with its own screen list so missing images are reported as incomplete, not removed UI.

The controlled resize failed before the fix and passes afterward. All 17 resulting desktop progress images are byte-identical to the complete original images. All 18 visual cases pass with 58 images and no retries. Three resize and error-cleanup controls, 50 report tests, nine real-artifact checks, Vault lint, TypeScript checks, changed-line formatting, and diff checks pass. The repair removes four net lines. New CI and visual review remain required.

services/simple-staking/e2e/mock/tx/withdrawing.ts
```diff
@@ -1,32 +1,72 @@
...
+
+import { DelegationState } from "@/ui/common/types/delegations";
+
+import { mockNetworkInfo } from "../../mocks/handlers/constants";
+import { activeTX } from "./unbonding";
+
+initBTCCurve();
+export const withdrawalPrivateKeyHex = "4".padStart(64, "0");
+const publicKey = Buffer.from(
+  ecc.pointFromScalar(Buffer.from(withdrawalPrivateKeyHex, "hex"), true)!,
```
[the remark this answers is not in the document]
jonybur: Reviewed at ea3287458ecf08c2ba108989d3a11fb1286b2065.

Named the funding outpoint and the fee rate. The fee declaration states its unit. The fixed key already has the name `withdrawalPrivateKeyHex`. `output_index: 0` selects the builder's sole unbonding output. `recordedAPRs` contains the recorded API responses with named fields, a source URL, a capture date, and request keys. Those response values stay together so the recording remains clear.

All exported fixture data and transaction bytes match the previous version. The four withdrawal browser cases pass in Chromium and Firefox with no retries. All 371 unit tests pass. The formatted change adds no net lines.

jonybur:
@greptileai Please review commit ea3287458ecf08c2ba108989d3a11fb1286b2065.

Name the withdrawal funding outpoint and fee rate. Keep the existing named private key and recorded APR payloads. All fixture data and transaction bytes are unchanged.

Four withdrawal browser cases pass in Chromium and Firefox with no retries. All 371 unit tests pass, with one existing skip. Lint has no errors and 75 existing warnings. Formatting, diff checks, and a before/after digest comparison pass. The follow-up adds no net lines.

gbarkhatov (approved):

Issue #2361, opened 2026-08-27
Make public raw Bitcoin engine access safe


jonybur:
## Outcome

The raw engine path added in August is rolled back. The engine exports its classes from the main import again, as it did before August.

Status: decision by jonybur on 2026-09-15 to roll back the raw path to the engine's earlier exports. #2516 rolled back the raw entry and replaced the guard PRs #2481 and #2493. It merged and released as engine and SDK 1.0.0. A second qualified final review remains. This issue does not block #2228 or its release.

## Problem

#2307 moved the engine's unchecked classes from the main import to a new `./raw` entry and added the SDK's `loadRawTbvWasm`. #2464 then deprecated that entry. #2228 needs neither.

## Scope

- Export the four engine classes from the engine's main import again.
- Remove the `./raw` entry and the SDK's `loadRawTbvWasm`. The refund builder loads the classes through the SDK's on-demand engine loader and keeps all of its checks.
- Remove the deprecation notes and the migration guide.
- Keep the SDK's on-demand engine loading and the refund safety tests.

## Complete when

- [x] The engine's main import exports the four classes, and no `./raw` entry exists.
- [x] The SDK still loads the engine only on demand.
- [x] The refund builder keeps its checks and tests.
- [ ] Two qualified final reviews and CI pass, and the rollback merges. CI passed and #2516 merged with one qualified approval, from gbarkhatov.
- [x] The engine and SDK release together. Both released as 1.0.0.

## Dependencies

- Parent: #2231. Related epic: #2228, which this issue does not block.
- #2481 and #2493 are closed in favor of #2516.
- The engine classes stay public without value checks, as before August. The SDK builders keep their checks.
- The rollback touches critical paths and needs two qualified human reviews.

PR #2464, opened 2026-09-08
feat(wasm): deprecate raw engine access and record migration paths

jonybur:
## Outcome

The raw Bitcoin engine API is deprecated. The migration guidance states which operations still lack guarded replacements.

## Change

Add deprecation metadata through typed constant exports. Keep class identities and signatures. Check that each raw class remains callable and that published declaration imports resolve. Use explicit refund guard tests.

## Safety

Raw access still bypasses value checks. Saved-transaction restoration still lacks trusted inputs for those checks. Runtime guards and the compatibility release decision remain unresolved. Two qualified reviews and the required line review apply before merge.

## Verification

At `896a1a1c81d5bb0e1469ddf263b672c90bfc6433`, 1,971 SDK tests, 9 built-package tests, and 14 engine tests pass. Builds, lint, formatting, and declaration-path checks pass. API documentation generation produces no diff.

Checks with a removed export and an inline documentation link confirm that the revised assertions detect the review defects. Final PR checks and review remain.

## Links

Tracks #2361. #2361 stays open.

Parent: #2231. Epic: #2228.


packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts
```diff
@@ -53,8 +53,12 @@ export function loadTbvWasm(): Promise<TbvWasmModule> {
...
 /**
- * Load the explicit raw-class entry for the one SDK primitive that must
- * reconstruct a stateful WASM transaction object (refund construction).
+ * Load and initialize the raw engine classes. This bypasses SDK value checks.
+ * The internal refund builder applies equivalent checks at its call site.
+ *
+ * @deprecated Use buildPrePeginPsbt, buildPeginTxFromFundedPrePegin, or
+ * buildRefundPsbt from @babylonlabs-io/ts-sdk/tbv/core/primitives.
+ * Raw callers must independently check transaction values and signing data.
  */
```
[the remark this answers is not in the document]
jonybur: The documentation check passes. I ran the SDK docs:clean target at this branch revision. It reports zero errors and produces no change under packages/babylon-ts-sdk/docs/api/.

loadRawTbvWasm is outside the entry points in packages/babylon-ts-sdk/typedoc.json. It is not re-exported by those entry points. The generated documentation therefore has no page to update for this JSDoc change. The migration guide in this PR records the deprecated API and its replacement paths.

jonybur:
@greptileai Please review commit 99bf455e11ebb2264cfa2f38524bd193c50ac5d7.

The declaration test host now exposes case sensitivity as a callback. The original tests reproduced both Linux failures. Documentation generation produces no API diff because loadRawTbvWasm is outside the configured TypeDoc entry points; see the reply on the documentation thread.

Verification: all 14 engine tests pass with Linux case sensitivity and on the native host. Engine lint and SDK docs:clean pass.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/__tests__/refund.test.ts
```diff
@@ -185,119 +185,78 @@ describe("buildRefundPsbt", () => {
   });
 
   describe("raw WASM refund transaction checks", () => {
-    it("rejects a refund transaction with the wrong version", async () => {
-      const { txHex, params } = await buildFundedPrePegin();
-
-      await withMutatedRawRefundTx(
+    it.each<
```
jrwbabylonlab: This folds six standalone tests into one `it.each`; the repo’s test philosophy prefers separate similar tests over a parameterized abstraction.

packages/babylon-tbv-rust-wasm/src/raw.ts
```diff
@@ -1,11 +1,24 @@
...
- * classes directly and therefore necessarily opt into eager glue loading.
- */
-// prettier-ignore
+import type * as Bindings from '../dist/generated/vault_wasm.js';
 // @ts-expect-error - generated artifacts live in dist/generated
-export { WasmPeginTx, WasmPeginPayoutConnector, WasmPrePeginTx, WasmPrePeginHtlcConnector } from './generated/vault_wasm.js';
+import * as raw from './generated/vault_wasm.js';
+
+/** @deprecated Raw access skips SDK value checks. Use buildPeginTxFromFundedPrePegin from @babylonlabs-io/ts-sdk/tbv/core/primitives. */
+export const WasmPeginTx: typeof Bindings.WasmPeginTx = raw.WasmPeginTx;
```
gbarkhatov, line 9: **This restructure converts a load-time link error into a silent `undefined`, and the new compatibility test cannot detect it.** Same applies to `raw-node.ts`.

Previously `export { WasmPeginTx } from './generated/vault_wasm.js'` — a missing export raised an ESM link error the instant `/raw` was imported. Now it is `export const WasmPeginTx = raw.WasmPeginTx` where `raw` is a namespace import that is `@ts-expect-error`'d on line 2-3. If the generated module ever stops exporting that name, the binding is `undefined`, the module loads clean, and the consumer fails later with `TypeError: WasmPeginTx is not a constructor` — far from the cause.

The new guard in `tests/wasm-loader.test.mjs` cannot see it, for two reasons:

- `Object.keys(raw)` still lists the name, because an exported `const` is enumerable whatever its value — so the `deepEqual` at :35-38 passes.
- `assert.equal(raw[name], generated[name])` at :39 then compares `undefined === undefined` and passes too.

The language-service test at :43-102 resolves only `.d.ts`, so a `.js`/`.d.ts` divergence slips through both.

Reachability is low — the generated pair is committed together — and the fix is one line inside the existing loop at `tests/wasm-loader.test.mjs:39`:

```js
assert.equal(typeof raw[name], 'function', name);
```

packages/babylon-tbv-rust-wasm/src/raw.ts
```diff
@@ -1,11 +1,24 @@
-/**
- * Explicit eager access to wasm-bindgen's generated browser module.
- *
- * Most consumers should use the package root, whose facade loads this module
- * lazily. This subpath exists for low-level callers that construct raw WASM
- * classes directly and therefore necessarily opt into eager glue loading.
- */
-// prettier-ignore
+import type * as Bindings from '../dist/generated/vault_wasm.js';
```
gbarkhatov, line 4: **These two files adopt `wasm-loader.ts`'s sibling-dependent emit trick without its warning comment, and the build-time guard was not extended to cover them.**

`raw.ts` and `raw-node.ts` now both carry `import type * as Bindings from '../dist/generated/vault_wasm.js'`. `wasm-loader.ts:5-7` documents exactly why that specifier is fragile:

> tsc copies this specifier into the emitted declaration unchanged, so it has to resolve from dist as well as from src… `check-lazy-entries.js` asserts it.

But `scripts/check-lazy-entries.js:148` hardcodes its loop:

```js
for (const loaderName of ['wasm-loader.d.ts', 'wasm-loader-node.d.ts']) {
```

So `raw.d.ts` and `raw-node.d.ts` now share the dependency with none of the protection. An `outDir` change would fail loudly for the loaders and ship silently broken for raw — handing consumers `any`-typed classes, which is the exact typing this PR exists to add.

Two-part fix: add `'raw.d.ts', 'raw-node.d.ts'` to that loop, and copy the three-line rationale comment onto both raw entries so the next person does not remove the import.

packages/babylon-tbv-rust-wasm/src/raw.ts
```diff
@@ -1,11 +1,24 @@
...
 // @ts-expect-error - generated artifacts live in dist/generated
-export { WasmPeginTx, WasmPeginPayoutConnector, WasmPrePeginTx, WasmPrePeginHtlcConnector } from './generated/vault_wasm.js';
+import * as raw from './generated/vault_wasm.js';
+
+/** @deprecated Raw access skips SDK value checks. Use buildPeginTxFromFundedPrePegin from @babylonlabs-io/ts-sdk/tbv/core/primitives. */
```
gbarkhatov: **The class-level `@deprecated` tags route per-class while the migration guide routes per-method, and they disagree in two places.** The tooltip is the only migration text most raw callers will ever read, so the disagreement matters more than it looks.

- **This line** tells every `WasmPeginTx` caller to use `buildPeginTxFromFundedPrePegin`. But guide row 48 says of `WasmPeginTx.fromJson` / `toJson`: *"No guarded round-trip replacement. See the compatibility blocker below."* A caller whose IDE strikes through `fromJson` is sent to a builder that cannot serve them — and away from the section explaining why.
- **Line 14** tells `WasmPrePeginTx` callers to use `buildPrePeginPsbt` or `buildRefundPsbt`, omitting `buildPeginTxFromFundedPrePegin` — yet guide row 45 lists `fromFundedTransaction` and `buildPeginTx` (both on that class), and guide line 16 says that builder "replaces funded reconstruction and `buildPeginTx`".

Fix: name all applicable builders on each tag, and point `fromJson`/`toJson` at the blocker section rather than at a builder.

packages/babylon-ts-sdk/docs/guides/raw-engine-migration.md
```diff
@@ -0,0 +1,94 @@
...
+Values can wrap during that conversion. A check inside the constructor cannot
+recover the original `bigint` values. Raw methods are synchronous; the existing
+SDK builders are asynchronous. A replacement must account for both constraints.
+
+The approved deprecation does not resolve these limits. The separate release
+decision must define how callers supply trusted inputs and how existing
+`fromJson` calls remain supported. Do not close #2361 or #2231 on this change.
+The guarded product-path work in #2386 remains separate.
+
+This change calls for a minor deprecation release of the SDK and engine.
```
gbarkhatov, line 90: This paragraph prescribes "a `feat` release commit without a breaking-change marker. It adds deprecation metadata and guidance, with **no runtime or signature change**."

The signature half is true — the language-service test pins the types. The runtime half is not quite: `raw.ts` and `raw-node.ts` moved from live re-export bindings to snapshotted `const` aliases. That is a real change in export form, in two files registered under CLAUDE.md §1 and §9, and it is the whole mechanism behind the `undefined`-binding note I left on `raw.ts:6`.

Combined with the `docs(wasm):` title, the framing invites a reviewer to treat those two files as untouched prose and skip them — which is the opposite of what §1 and §9 ask for. Worth stating the export-form change explicitly here instead of asserting no runtime change.

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
```diff
@@ -10,6 +13,46 @@ import {
...
+      true,
+    )
+    .statements.find(
+      (node) =>
+        ts.isFunctionDeclaration(node) && node.name?.text === "loadRawTbvWasm",
+    );
+  assert.ok(declaration, "The published raw loader declaration must remain");
+  const deprecation = ts.getJSDocDeprecatedTag(declaration);
+  assert.ok(deprecation, "The published raw loader must be deprecated");
+  assert.match(deprecation.comment, /buildRefundPsbt/);
```
gbarkhatov: `assert.match` requires a string, but `ts.getJSDocDeprecatedTag(...).comment` is typed `string | NodeArray<JSDocComment>`.

It is a plain string today only because the `@deprecated` text at `wasm/index.ts:59-61` happens to contain no inline tag. Add `{@link buildRefundPsbt}` — the natural improvement to a deprecation notice, and a plausible review request on this very PR — and it flips to a `NodeArray`. The pin then dies with `ERR_INVALID_ARG_TYPE` instead of a clean assertion failure, for a change that should have *satisfied* it.

```js
assert.match(ts.getTextOfJSDocComment(deprecation.comment) ?? "", /buildRefundPsbt/);
```

packages/babylon-tbv-rust-wasm/README.md
```diff
@@ -6,10 +6,9 @@ The normal package entry is a lazy facade: importing it does not load the
 wasm-bindgen glue or instantiate/download the `.wasm` binary. The generated
 module is fetched on the first facade call that needs it.
 
-Low-level consumers that construct wasm-bindgen classes directly must opt in
-to the eager `@babylonlabs-io/babylon-tbv-rust-wasm/raw` subpath and call its
-`initWasm` export before constructing a class. Raw classes are intentionally
-not exported from the lazy root entry.
+The eager `@babylonlabs-io/babylon-tbv-rust-wasm/raw` classes are deprecated.
```
gbarkhatov, line 9: The removed sentence — "Raw classes are intentionally not exported from the lazy root entry" — reads as editorial, but it documents a machine-enforced invariant.

`scripts/check-lazy-entries.js:130` throws "`${entryName}` statically reaches generated WASM glue, so importing the facade loads the engine eagerly" if anyone re-exports them from the root.

The replacement text explains only that `/raw` is deprecated — which invites precisely the wrong remedy. A contributor reading "these are deprecated, stop using the subpath" will reasonably try to hoist the classes onto the lazy root to retire `/raw`, and hit a build failure whose rationale is no longer written anywhere in the README.

Restoring one sentence under "Raw WASM Types" covers it.

(The separately deleted `await initWasm()` worked example I'd leave alone — the section still states the requirement at :473.)

gbarkhatov (dismissed):
Reviewed at `99bf455e` with three independent passes. **No merge blockers.** Because this PR edits `CLAUDE.md` and `SECURITY.md` — the documents governing every other PR in the repo — I read the policy diff more carefully than the code diff, and answered three questions before anything else.

**Does the policy text still say everything it said?** Yes. Both edits are strictly additive: `CLAUDE.md @@ -59,6 +59,7 @@` adds one bullet, `SECURITY.md @@ -221,6 +221,10 @@` adds four lines. Zero deletions, no "must" softened. "The rule above binds at the call site, not at the export" survives verbatim, as does its `SECURITY.md` counterpart, and §9's registration of `raw.ts`/`raw-node.ts` is untouched.

**Is a deprecation notice the right instrument?** It is not an enforcement instrument, and nothing in the repo backs it — there is no `@typescript-eslint/no-deprecated` rule in any of the three eslint configs (root, ts-sdk, engine); I grepped, zero hits. The tags produce an editor hover and two declaration-shape tests, nothing in CI. **To the PR's credit the docs never claim otherwise** — "The retained path still permits a bypass" appears in all three documents, alongside "No removal date is set" and "Do not close #2361 or #2231 on this change." So there is no overstatement. There is also no new protection: nothing stops `services/vault` importing `/raw` tomorrow. Worth stating plainly since that is the gap the epic still has to close.

**Do the new tests exercise the guards, or only the deprecation?** They exercise the guards, genuinely. `withMutatedRawRefundTx` spies on `buildRefundTx`, calls the original, then mutates real engine output. Three rows are new — wrong input txid, extra input (`/must have exactly 1 input, got 2/`), extra output (`/must have exactly 1 output, got 2/`) — which is precisely §1's re-parse guard. Every row asserts a specific message regex rather than a bare `toThrow()`.

### The §4 frozen-file check passes

`packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts` — the single hunk is entirely inside the JSDoc at lines 55-62. `export function loadRawTbvWasm(): Promise<RawTbvWasmModule>` at :63 and its whole body through :88 are byte-identical. No signature, argument order, label, export ordering, or frozen expander touched. Zero executable bytes changed.

### Checked and dismissed

- **greptile's "Generated API docs are stale" — the reply is correct and the thread can retire.** I confirmed the mechanism positively rather than by absence: `typedoc.json` lists nine `entryPoints`, none of which is `src/tbv/core/wasm/index.ts`. Only `primitives/index.ts` reaches `../wasm`, re-exporting six named functions and five types — `loadRawTbvWasm` is not among them. Those six re-exported functions *do* have pages under `docs/api/`; `loadRawTbvWasm` has none. So there is no page for this JSDoc to update, `docs:clean` produces no diff, and `verify` passing is consistent. The reasoning is sound and reusable on future PRs.
- **"The new `SECURITY.md` paragraph makes 'The mitigation is…' read as covering `/raw`."** Refuted on the pre-image: that sentence already directly followed the `/raw` paragraph before this PR, so the adjacency is pre-existing, and line 240 explicitly re-separates them ("…**or a new `/raw` consumer without call-site cross-checks**…").
- **Class identity concerns from the re-export change.** `export const X = raw.X` preserves `instanceof`, statics (`WasmPeginTx.fromJson`) and `refund.ts`'s `new WasmPrePeginTx(...)`; the paired `export type X` restores the dual value/type meaning the class re-export gave for free.
- **The migration guide's retained-operation audit** was checked member by member against `dist/generated/vault_wasm.d.ts` — every member maps to a table row, no invented method names. All cross-document links resolve, and all five builders named in the docs really are exported from `tbv/core/primitives`.

### Two notes that are not this PR's doing

**`CLAUDE.md` §1's guard enumeration is already condensed at this PR's base**, and I want to flag it because it is easy to assume this PR did it. It did not — the line is unchanged context here. The specifics were condensed in `4b0861989` (#2398): §1 used to name `htlcVout`, "exactly 1 input", and the `htlcValue - refundFee` output value, and now reads only "compares the WASM template's script and value with the funded output, and checks the final refund input and output". Since the new `refund.test.ts` rows assert exactly those dropped specifics, restoring them to §1 would be a natural companion change — separate PR, not this one.

**The `it.each` thread from a prior reviewer is still standing** at `refund.test.ts:188-271`. It folds **nine** cases into one table (not six as the thread says — I counted the rows). It stands under the test-philosophy rule. Noting the batch inconsistency rather than re-arguing it: sibling #2465 had the identical finding and was de-parameterized there (7 tables → 19 explicit tests, 64/64 assertions intact, which I verified). That existing thread carries this; I have not duplicated it.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/__tests__/refund.test.ts
jonybur: Addressed in 896a1a1c81d5bb0e1469ddf263b672c90bfc6433.

Replaced the nine-row refund table with nine explicit tests. Each retains its transaction setup, builder call, and exact expected error. The SDK suite passes.

packages/babylon-tbv-rust-wasm/src/raw.ts
jonybur: Addressed in 896a1a1c81d5bb0e1469ddf263b672c90bfc6433.

Each exported raw class must now be callable as well as equal to the generated class. Removing an export from a copy of the real module passes the old equality check and fails the new check.

packages/babylon-tbv-rust-wasm/src/raw.ts
jonybur: Addressed in 896a1a1c81d5bb0e1469ddf263b672c90bfc6433.

The existing emitted-import guard now includes raw.d.ts and raw-node.d.ts. Both source files state why the import must resolve from both source and output directories.

packages/babylon-tbv-rust-wasm/src/raw.ts
jonybur: Addressed in 896a1a1c81d5bb0e1469ddf263b672c90bfc6433.

Both raw entry points now list all applicable builders for WasmPrePeginTx. The WasmPeginTx tag sends fromJson/toJson callers to the compatibility blocker instead of implying a drop-in builder replacement.

packages/babylon-ts-sdk/docs/guides/raw-engine-migration.md
jonybur: Addressed in 896a1a1c81d5bb0e1469ddf263b672c90bfc6433.

The guide now states that live re-exports become typed constant aliases. It claims preserved class identities and signatures. The PR title and description now describe the export-form change.

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
jonybur: Addressed in 896a1a1c81d5bb0e1469ddf263b672c90bfc6433.

The assertion now uses TypeScript's JSDoc text normalizer. A probe with an inline link confirms that valid documentation no longer fails the check.

packages/babylon-tbv-rust-wasm/README.md
jonybur: Addressed in 896a1a1c81d5bb0e1469ddf263b672c90bfc6433.

Restored the explicit rule that raw classes stay out of the lazy root entry. Raw imports remain an eager path.

jonybur:
@greptileai Please review commit 896a1a1c81d5bb0e1469ddf263b672c90bfc6433.

Check callable raw exports and declaration imports. Use explicit refund tests and accurate migration guidance.

1,971 SDK tests, 9 built-package tests, and 14 engine tests pass. Builds, lint, and documentation checks pass. Export and inline-JSDoc probes confirm the old checks fail to detect the reviewed cases.

gbarkhatov (approved):

PR #2465, opened 2026-09-08
test(vault): record wallet requirements for current actions

### LINKED ISSUE #2447: [eth-first] Record the wallet requirement for every Vault action
## Outcome

Every Vault connection check has a recorded wallet requirement and a test for current behavior.

Status: complete. PR #2465 merged at [8a480159](https://github.com/babylonlabs-io/babylon-toolkit/commit/8a4801591d97cbe49a1690362801849342c06422). The audit and current-behavior tests are delivered. Wallet behavior changes remain in #2232.

## Problem

The existing connection signal can require both wallets for an Ethereum action. The implementation needs a complete action list before behavior changes.

## Scope

- Audit every use of the old connection signal. Classify each action as Ethereum, Bitcoin, or both.
- Record prompt, restore, disconnect, and pending-action dependencies.
- Add tests for current behavior. Keep behavior unchanged.

## Complete when

- [x] The audit lists every old connection check and its action requirement.
- [x] Tests record the current result for each audited route or action.
- [x] A linked PR contains the audit and passing tests without behavior changes.

## Dependencies

Parent: #2232. Epic: #2228. Later behavior changes remain blocked by #2230 and #2231.

jonybur:
## Outcome

The Vault wallet audit records the current connection requirements. Tests now detect removal of the account-query gates.

## Change

Test disconnected states while an Ethereum address remains available. Record the Bitcoin balance and deposit-entry gates. Keep one test for the connected sidebar. State exactly what the pending-deposit tests check.

## Safety

Only tests and the audit change. Wallet behavior changes remain in #2232.

## Verification

At `ae5bc3afdf81d562727112c3ad0e46cce8d5685c`, 3,783 Vault tests pass with 6 existing skips. Lint and formatting pass. Lint reports 134 existing warnings. Removing the three query gates makes all five affected cases fail.

Real-wallet acceptance evidence remains in #2355. Required PR checks pass. Human review remains.

This PR is ready for human review. Merge #2462 first. Then update this PR to keep both sets of tests in the three conflicting files. Rerun the Vault tests and required CI checks before merging this PR.

## Links

Closes #2447.

Tracks #2232. #2232 stays open. Epic: #2228.


services/vault/src/hooks/__tests__/usePendingDeposits.test.tsx
```diff
@@ -130,24 +133,31 @@ describe("usePendingDeposits", () => {
...
+        "a1",
+      ]);
+      expect(result.current.expiredActivities.map((a: any) => a.id)).toEqual([
+        "a2",
+      ]);
+      expect(result.current.hasPendingDeposits).toBe(connected);
+      expect(result.current.hasExpiredDeposits).toBe(connected);
+      expect(useVaultDeposits).toHaveBeenCalledWith("0xethtest");
+    },
+  );
```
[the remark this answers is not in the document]
jonybur: Fixed in 86990d785d411e76014969bedb91d785d1bcad99. The seven tables are now 19 explicit tests. All 64 assertions remain. Full Vault tests pass: 3,782 passed and 6 skipped. Build, typecheck, lint, formatting, and diff checks pass. The audit references match the new test names.

jonybur:
@greptileai Please review commit 86990d785d411e76014969bedb91d785d1bcad99.

The seven grouped tests are now 19 explicit cases. All 64 assertions remain. The audit references and verification counts match the final tests.

Verification: 3,782 Vault tests pass, with 6 skipped. Build, typecheck, lint, formatting, and diff checks pass.

services/vault/src/components/pages/__tests__/Loans.test.tsx
```diff
@@ -179,6 +179,7 @@ describe("Loans page — loading gate", () => {
     // Disconnected prompts for a wallet — it must not claim the depositor has
     // no loans when we haven't looked at an address yet.
     expect(screen.getByText(COPY.loans.emptyDisconnected)).toBeInTheDocument();
+    expect(useDashboardStateMock).toHaveBeenCalledWith(undefined);
```
gbarkhatov, line 182: **This assertion passes with or without the gate it exists to pin.**

Line 166 sets `useETHWalletMock.mockReturnValue({ address: undefined })` alongside `isConnected: false` on 165. The guard is `useDashboardState(isConnected ? address : undefined)` at `Loans.tsx:50`. With `address` already `undefined`, both branches of the ternary evaluate to `undefined` — so deleting `isConnected ?` entirely leaves this test green.

That is precisely the edit #2232 will make, which makes this the one assertion in the file that most needs to be discriminating and currently is not.

One-word fix, and the pattern is already in the repo: `VaultsPage.test.tsx:56` sets a real `address: "0x1111111111111111111111111111111111111111"` in its disconnected case, so its otherwise-identical `toHaveBeenCalledWith(undefined)` genuinely proves the guard fired. Giving the ETH mock here a real address does the same.

services/vault/src/components/pages/Liquidations/__tests__/Liquidations.test.tsx
```diff
@@ -252,6 +252,7 @@ describe("Liquidation Dashboard — connection and position gates", () => {
 
     const emptyState = screen.getByTestId("liquidations-empty-state");
     expect(emptyState).toHaveAttribute("data-connected", "false");
+    expect(useDashboardStateMock).toHaveBeenCalledWith(undefined);
```
gbarkhatov, line 255: Same issue as `Loans.test.tsx:182`, same fix.

Line 238 sets `address: undefined` with `isConnected: false`, and the guard is `const account = isConnected ? address : undefined` (`Liquidations/index.tsx:79`). The address is already `undefined`, so removing `isConnected ?` keeps this assertion green.

Give the ETH mock a real address, as `VaultsPage.test.tsx:56` does.

services/vault/src/components/simple/__tests__/DashboardPage.test.tsx
```diff
@@ -158,6 +174,39 @@ beforeEach(() => {
...
 describe("DashboardPage composition", () => {
+  it("shows the landing page when both wallets are missing", () => {
+    walletMock.btcConnected = false;
+    walletMock.ethConnected = false;
+
+    render(<DashboardPage />);
+
+    expect(screen.getByTestId("disconnected-overview")).toBeInTheDocument();
+    expect(screen.queryByTestId("overview-section")).not.toBeInTheDocument();
+    expect(useDashboardState).toHaveBeenCalledWith(undefined);
```
gbarkhatov, line 185: Two of the three new landing-page assertions here are vacuous for the same reason as `Loans.test.tsx:182` — this one and line 196.

The shared mock at line 41 is `address: walletMock.ethConnected ? "0xabc" : undefined`, so in the "both missing" and "only Bitcoin" cases the address is already `undefined` and the assertion cannot reach the guard at `DashboardPage.tsx:95`.

Lower severity than the Loans one, because the third case — "only Ethereum", where the ETH address is present but `isConnected` is false — *does* exercise the guard, so the gate is pinned once in this file. The cost is that the three read as symmetric matrix coverage while only one third carries weight, which is misleading to whoever edits them next.

docs/audits/2447-vault-wallet-requirements.md
```diff
@@ -0,0 +1,135 @@
...
+| [VaultWalletConnectionProvider](../../services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx)                                                             | Configures BTC and ETH. Sets `requiredChains={["BTC", "ETH"]}`. Restores a persistent wallet session.                                                            |
+| [RootLayout](../../services/vault/src/components/pages/RootLayout.tsx)                                                                                                 | Computes `btcConnected && ethConnected`. Uses it for the home entry screen, sidebar, and connected banners.                                                      |
+| [Connect](../../services/vault/src/components/Wallet/Connect.tsx)                                                                                                      | Requires dialog confirmation and both wallets for the wallet menu. The menu also checks geofencing. A locked BTC wallet replaces the menu with an unlock button. |
```
gbarkhatov: **This row omits what is probably the most consequential ETH-gates-BTC coupling in the app**, and it is invisible unless you trace the variable past its name.

The row records Connect as requiring "dialog confirmation and both wallets for the wallet menu." But `Connect.tsx:68-69` computes `isWalletConnected = walletSessionConfirmed && btcConnected && ethConnected`, feeds it into `canShowWalletMenu` on line 72, and then uses that as the `enabled` flag for a **Bitcoin** query:

```ts
const utxoOptions = useMemo(() => ({ enabled: canShowWalletMenu }), [canShowWalletMenu]);
const { inscriptionUTXOs } = useUTXOs(btcAddress, utxoOptions);
```

So a Bitcoin-address UTXO read is currently suppressed unless **Ethereum** is connected. The source comment on line 71 says so outright — "Single source for both the UTXO query gate and the menu render branch below" — but the doc records only the menu half and drops the query half.

For an audit whose whole purpose is to make Bitcoin optional (#2232 / epic #2228), that is the exact coupling someone needs to find. One clause fixes it.

docs/audits/2447-vault-wallet-requirements.md
```diff
@@ -0,0 +1,135 @@
...
+| Reorder vaults for liquidation priority                 | ETH wallet client and account. No BTC signer.                                                                             | [useReorderVaults](../../services/vault/src/applications/aave/hooks/useReorderVaults.ts).                                                                                                                                                     |
+| Create and register a deposit                           | BTC wallet for the deposit secret and Bitcoin signatures. ETH wallet for registration.                                    | [useDepositFlow](../../services/vault/src/hooks/deposit/useDepositFlow.ts) and [ethereumSubmit](../../services/vault/src/hooks/deposit/depositFlowSteps/ethereumSubmit.ts).                                                                   |
```
gbarkhatov: The deposit row cites only `useDepositFlow` and `ethereumSubmit`, and `SimpleDeposit.tsx` appears in neither table — but it carries its own click-time Bitcoin gate.

`SimpleDeposit.tsx:377` calls `verifyBtcWalletLiveness(btcWalletProvider, connectedBtcAddress, …)` on the deposit click with an explicit fail-loud branch, and the CTA state gates on `isWalletConnected` at lines 110 and 499.

This matters because of how the document reads as a whole: the first table asserts completeness ("These are all nine callers" — which I checked, and it is accurate), so the second table reads as complete too. Someone implementing #2232 from this doc would miss the click-time check.

Either add a row for it, or state the second table's inclusion criterion so its scope is explicit.

services/vault/src/hooks/__tests__/usePendingDeposits.test.tsx
```diff
@@ -130,23 +133,42 @@ describe("usePendingDeposits", () => {
     expect(result.current.ethAddress).toBe("0xethtest");
   });
 
-  it("places EXPIRED activities with unsignedPrePeginTx into expiredActivities, not pendingActivities", () => {
+  it("keeps deposit records and enables their flags when Bitcoin is connected", () => {
```
gbarkhatov: This rename quietly absorbed a routing rule that the test still asserts, under a name that no longer mentions it.

`pr.diff:1031` removes `it("places EXPIRED activities with unsignedPrePeginTx into expiredActivities, not pendingActivities")`. The replacement is named for the Bitcoin connection — but the `a3` fixture on line 140 (EXPIRED *without* `unsignedPrePeginTx`) and the assertions on lines 149 and 168 still verify that EXPIRED routing rule.

The practical cost: a future change to `usePendingDeposits.ts`'s expiry routing fails a test whose name points at the wallet-connection subsystem, so whoever hits it looks in the wrong place first. Either split the routing assertions back into their own named test, or widen this name to cover both.

services/vault/src/components/pages/__tests__/RootLayout.test.tsx
```diff
@@ -155,13 +161,15 @@ describe("RootLayout — header wiring", () => {
     expect(screen.getByText(COPY.header.networkBadge)).toBeInTheDocument();
   });
 
-  it("disconnected: drops the sidebar and page title for the entry chrome", () => {
-    walletMock.connected = true;
+  it("shows the entry layout when both wallets are missing", () => {
```
gbarkhatov: Minor, and a direct artifact of the de-parameterization — worth fixing while the rewrite is fresh.

All three entry-layout tests (lines 165-169, 187-191, 206-210) open by setting both wallets connected, rendering, asserting `aside` is present, and unmounting — before testing the thing their name describes. That connected-shell behaviour is asserted three times and named zero times.

The pre-fix `it.each` wrote the preamble once, so this is new duplication rather than pre-existing. Per "each test verifies ONE specific behavior", lifting it into a single test of its own would leave each of the three testing only what it claims to.

gbarkhatov (dismissed):
Reviewed at `86990d78` with three independent passes. **No hard merge-blockers** — no production source changes and no §1–§9 critical path touched.

**First, the greptile thread is genuinely closed, and I verified it rather than taking the reply on trust.** Zero `it.each` / `test.each` / `describe.each` survivors across all 12 changed test files. The 7 tables really did become **19** explicit tests with **64 → 64** assertions intact, and all three passes independently reconstructed the same per-file breakdown (LoanFlowOverlay 2/6, ReserveDetailPanel 3/6, Connect 4/16, RootLayout 3/15, DashboardPage 3/9, VaultWalletConnectionProvider 2/2, usePendingDeposits 2/10). No assertion was dropped. The 19 names describe behaviour plus precondition rather than input tuples, so there is no naming finding either.

**The substantive issue is different, and it defeats this PR's own purpose.** The point of the change is to pin today's gates *before* #2232 rewrites them. Four of the new assertions cannot do that, because the test's own mock already satisfies them — deleting the very guard they exist to protect keeps them green. Details inline; the fix is one word per test, and `VaultsPage.test.tsx:56` already shows the right pattern by using a real `0x1111…` address.

**The audit document is broadly trustworthy** — `useConnection = btcConnected && ethConnected` with exactly nine application callers, the `requiredChains` config, the 3000 ms BTC debounce, dialog-scoped ETH-disconnect suppression, immediate address-change resets, and ~30 relative links all check out at head. Two omissions are worth fixing before it becomes #2232's reference, both inline.

### Checked and dismissed

- **Audit/issue-number references in test names or comments?** None in the new work. The single hit, `(audit F7)` at `ReserveDetailPanel.test.tsx:5`, is pre-existing and absent from this diff's added lines.
- **`getByRole("button", { name: "Connect" })` hardcodes user-facing copy?** No. That string is core-ui's `ConnectButton` default `text` prop (`ConnectButton.tsx:19`), not vault `@/copy` text, and `Connect.tsx` passes no `text`. The copy.ts rule does not bite.
- **Dead code left by the de-parameterization?** None. The removed tables held inline case literals, not extracted constants or row-mappers, so nothing was orphaned; `mockUseConnection` is fully removed with no dangling references.
- **`data-testid` moved, or a `copy.ts` string inlined?** Neither.

### Verified clean

- `reports a combined connection before the menu accepts confirmation` (`Connect.test.tsx`) renders the real `useConnection` against the mocked connector and pins the genuine hook-vs-menu disagreement — the most valuable single test in this PR for #2232.
- The new `VaultWalletConnectionProvider` cases are real coverage, notably that `scheduleBtcReset` never consults `connectModalVisibleRef` while `handleEthDisconnect` does — an asymmetry the doc records honestly rather than wishing away.
- The `ReserveDetailPanel` note that the reserve query receives the address *before* the display gate is subtle and correct.

### Not verified

One agent reported that `useVaultActions.test.ts:383` pins `broadcastError` still reading "BTC wallet is not connected" *after* reconnect, alongside the behaviour the test names — which would fail as a false retry regression if a UX pass ever clears that error on reconnect. I did not confirm it myself, so I am flagging it as unverified rather than asserting it.

### Sibling conflict

#2462 also edits `LoanFlowOverlay.test.tsx` and `Loans.test.tsx` and changes production Aave debt derivation. Both sit on `main` and neither contains the other — expect a textual conflict in those two files, and re-check this PR's `useAaveUserPosition` argument assertions after whichever merges second.

services/vault/src/components/pages/__tests__/Loans.test.tsx
jonybur: Addressed in ae5bc3afdf81d562727112c3ad0e46cce8d5685c.

The disconnected test keeps an Ethereum address available. Removing the connection gate now makes the query assertion fail.

services/vault/src/components/pages/Liquidations/__tests__/Liquidations.test.tsx
jonybur: Addressed in ae5bc3afdf81d562727112c3ad0e46cce8d5685c.

The disconnected test keeps an Ethereum address available. Removing the connection gate now makes the query assertion fail.

services/vault/src/components/simple/__tests__/DashboardPage.test.tsx
jonybur: Addressed in ae5bc3afdf81d562727112c3ad0e46cce8d5685c.

All disconnected wallet combinations retain the existing Ethereum test address. All three cases fail when the production query gate is removed.

docs/audits/2447-vault-wallet-requirements.md
jonybur: Addressed in ae5bc3afdf81d562727112c3ad0e46cce8d5685c.

The audit now records that the Bitcoin UTXO query requires the wallet connection, confirmation, and location checks.

docs/audits/2447-vault-wallet-requirements.md
jonybur: Addressed in ae5bc3afdf81d562727112c3ad0e46cce8d5685c.

The audit now covers both deposit entry points, the combined-wallet button gate, and the Bitcoin check at click time.

services/vault/src/hooks/__tests__/usePendingDeposits.test.tsx
jonybur: Addressed in ae5bc3afdf81d562727112c3ad0e46cce8d5685c.

Both test names now describe expired-record routing and the Bitcoin connection flag. They no longer claim broader action behavior.

services/vault/src/components/pages/__tests__/RootLayout.test.tsx
jonybur: Addressed in ae5bc3afdf81d562727112c3ad0e46cce8d5685c.

The connected sidebar behavior now has one named test. The three entry-layout tests no longer repeat that setup and assertion.

jonybur:
@greptileai Please review commit ae5bc3afdf81d562727112c3ad0e46cce8d5685c.

Make disconnected query tests retain the test address. Record the current deposit gates and remove unrelated repeated test setup.

3,783 Vault tests pass, with 6 existing skips. Lint and formatting pass. Removing the three production query gates makes all five affected cases fail.

gbarkhatov (approved):

gbarkhatov (approved):

PR #2466, opened 2026-09-08
docs(ts-sdk): consolidate frozen expander comments

jonybur:
## Outcome

Each frozen expander has one attached stability comment. The comments retain the permanent effect of changing derived bytes.

## Change

Remove duplicate comments. Keep Rust ownership, the effect of byte changes, and the reference to CLAUDE.md section 4. Use the required regular hyphens.

## Safety

Emitted code is unchanged. The required owner reviews apply to this registered critical path.

## Verification

At `ce1f46848670c1bbea71e00ae5aed607d5be46fb`, 1,966 SDK unit tests and 7 built-package tests pass. Lint and API documentation checks pass. Generated documentation is unchanged. Emitted code matches the previous commit.

## Links

Covers [the duplicate stability comment finding](https://github.com/babylonlabs-io/babylon-toolkit/pull/2307#discussion_r3865457104).

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts
@@ -348,7 +348,7 @@ export async function expandAuthAnchor(root: Uint8Array): Promise<Uint8Array> {
 /**
  * Derive 32-byte `hashlockSecret` for HTLC `htlcVout` (preimage → `activateVaultWithSecret`).
  * @stability frozen - btc-vault Rust owns this API.
- * Changing the derived bytes prevents affected vaults from activating. See CLAUDE.md §4.
+ * Changing the derived bytes means affected vaults can never activate. See CLAUDE.md §4.
  */
 export async function expandHashlockSecret(
   root: Uint8Array,
@@ -360,7 +360,7 @@ export async function expandHashlockSecret(
 /**
  * Derive 64-byte `wotsSeed` for HTLC `htlcVout` (→ WOTS keys, hashed as `depositorWotsPkHash`).
  * @stability frozen - btc-vault Rust owns this API.
- * Changing the derived bytes breaks existing `depositorWotsPkHash` values and prevents claims. See CLAUDE.md §4.
+ * Changing the derived bytes breaks existing `depositorWotsPkHash` values. No claim path remains. See CLAUDE.md §4.
  */
 export async function expandWotsSeed(
   root: Uint8Array,
--- a/packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts
@@ -338,11 +338,8 @@ export async function getChallengeAssertScriptInfo(
 
 /**
  * Derive 32-byte `authAnchor` (OP_RETURN preimage → VP bearer token).
- * @stability frozen — forwards the frozen expander in `@babylonlabs-io/babylon-tbv-rust-wasm`; see CLAUDE.md §4.
- */
-/**
- * Derive 32-byte `authAnchor` (OP_RETURN preimage → VP bearer token).
- * @stability frozen — owned by btc-vault Rust via the vault-wasm pin (`VAULT_WASM_COMMIT`); rotation breaks VP auth for existing deposits.
+ * @stability frozen - btc-vault Rust owns this API through the vault-wasm pin (`VAULT_WASM_COMMIT`).
+ * Changing the derived bytes breaks VP auth for existing deposits. See CLAUDE.md §4.
  */
 export async function expandAuthAnchor(root: Uint8Array): Promise<Uint8Array> {
   return (await loadTbvWasm()).expandAuthAnchor(root);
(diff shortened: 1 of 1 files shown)


packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts
```diff
@@ -338,23 +338,17 @@ export async function getChallengeAssertScriptInfo(
 
 /**
  * Derive 32-byte `authAnchor` (OP_RETURN preimage → VP bearer token).
- * @stability frozen — forwards the frozen expander in `@babylonlabs-io/babylon-tbv-rust-wasm`; see CLAUDE.md §4.
- */
-/**
- * Derive 32-byte `authAnchor` (OP_RETURN preimage → VP bearer token).
- * @stability frozen — owned by btc-vault Rust via the vault-wasm pin (`VAULT_WASM_COMMIT`); rotation breaks VP auth for existing deposits.
+ * @stability frozen - btc-vault Rust owns this API through the vault-wasm pin (`VAULT_WASM_COMMIT`).
```
gbarkhatov, line 341: Non-blocking, but worth a moment because it is a registered §4 surface. This applies to all three markers — lines 341, 350 and 362.

**The separator changed from an em dash to an ASCII hyphen.** I grepped every `@stability frozen` marker in `packages/`: there are 17, and after this PR these three are the **only** hyphenated ones. The other 14 — including `vault-secrets/context.ts`, `deriveVaultRoot.ts`, `wots/blockDerivation.ts` and the six engine wrappers — all use ` — `.

That matters more than typography usually would: someone auditing the frozen registry with `grep "@stability frozen —"` now misses exactly the three expanders. Nothing parses the tag today, so there is no functional consequence — it is a maintenance hazard, not a bug.

**Second, smaller: the wording drifted from the wrappers this hop delegates to.** `packages/babylon-tbv-rust-wasm/src/index.ts:370,383,400` and `src/index-node.ts:466,479,496` still carry the pre-image sentences verbatim, with stronger permanence framing:

- engine: `rotation means affected vaults can never activate` → here: `prevents affected vaults from activating`
- engine: `rotation breaks existing depositorWotsPkHash → no claim path` → here: `breaks existing depositorWotsPkHash values and prevents claims`

CLAUDE.md §4 registers the SDK hop and both engine wrappers as one frozen surface, and these were previously word-for-word matched pairs.

Either direction is fine — restore the em dash and the original phrasing on these three, or land the same rewording on the six engine wrappers in this PR so the mirror stays exact. I'd lean to the first, since the engine wording is the stronger of the two.

gbarkhatov (dismissed):
Reviewed at `f137dc14` with three independent passes, on a CLAUDE.md §4 registered frozen path. **No merge blockers.** One non-blocking consistency nit inline.

**The diff is genuinely comment-only** — I read all 46 lines and the post-image myself. Every `+`/`−` sits inside a `/** … */` block; the three `export async function` lines, their parameter lists and order, and the `loadTbvWasm()` delegation bodies are unchanged context. No signature, label string, argument reorder or export-ordering change, so §4's hard-fork rule is not engaged and the golden-vector / coordinated-doc requirement does not fire. `VAULT_WASM_COMMIT`, `derive-vault-secrets.md` and `expand.test.ts` are untouched.

**The consolidation strengthened the docs rather than weakening them, and that is worth stating** because it is the opposite of what a net `−9` on a safety comment usually means. Pre-PR each expander carried *two stacked* JSDoc blocks. TypeScript binds only the **lower** one — and the lower block was the one missing `see CLAUDE.md §4`. The §4 pointer lived in the upper block, attached to nothing. After this PR every expander's *binding* block carries it. That is a real improvement to a registered critical path.

I also checked the specific failure mode I was most worried about — one expander ending up with a thinner warning than the other two — and it does not occur. Each of the three keeps Rust ownership, a distinct consequence, and the §4 pointer. The `VAULT_WASM_COMMIT` pin appearing only on `expandAuthAnchor` is **pre-existing**, not introduced here, and matches the engine wrappers exactly.

### Checked and dismissed

- **Textual conflict with #2466's sibling #2464, which edits this same file?** Does not materialize. #2464's only hunk here is `@@ -53,8 +53,12 @@` (the `loadTbvWasm`/`loadRawTbvWasm` JSDoc), ~285 lines above these hunks at `-338`/`-350`/`-365`, off the same base blob `34b84c17`. Merges cleanly in either order. Its `CLAUDE.md` edit appends a bullet under §1 and adds no section, so the `See CLAUDE.md §4` references written here stay correct after both land.
- **Dangling cross-reference from the deletion?** None. `grep "forwards the frozen expander"` over the worktree returns nothing, and `vault-secrets/index.ts:15` still resolves — each surviving block states a distinct failure mode.
- **Could the reworded JSDoc make `docs/api/` stale?** No — the generated tree does not cover this module, which corroborates the green `verify` gate rather than merely trusting it.

### Verified clean

- No stacked-duplicate JSDoc remains; the consolidation is complete, not partial.
- The surviving text is factually correct — `CLAUDE.md:82` is still `### 4. Vault-secret derivation (frozen on-chain-binding API)`.
- The `export type { … }` tail block is untouched.

One adjacent note, not this PR and not a request: the boundary asserts nothing about the expanders' 32/32/64-byte return lengths. Pre-existing; this PR neither adds nor removes that gate.

packages/babylon-ts-sdk/src/tbv/core/wasm/index.ts
jonybur: Addressed in ce1f46848670c1bbea71e00ae5aed607d5be46fb.

Restored the wording that derived-byte changes permanently change the vault secrets. Kept regular hyphens because the project instructions require them. The emitted code remains identical.

jonybur:
@greptileai Please review commit ce1f46848670c1bbea71e00ae5aed607d5be46fb.

Restore the permanent effect of changing derived bytes. Keep the required regular hyphens.

1,966 SDK unit tests and 7 built-package tests pass. Lint and API documentation checks pass. Emitted code is unchanged.

gbarkhatov (approved):

PR #2467, opened 2026-09-08
chore(ts-sdk): register the public WASM entry

jonybur:
## Outcome

The unused-code check registers the published SDK WASM entry through an explicit path.

## Change

List the vault-secret and WASM entries on separate lines. Both paths resolve to the intended files.

## Safety

TAP_INTERNAL_KEY remains part of the published SDK API. External callers can use it even when repository code does not. No source code, exports, or values change. The separate engine package is outside this check.

## Verification

At `617b47825fb392f85133ed6937edc7822bfe85b2`, the two literal paths resolve to the same files as the previous pattern. The earlier Knip 5.88.1 comparison found no unused value exports and 16 existing type findings. Formatting and diff checks pass.

## Links

Covers [the public entry finding](https://github.com/babylonlabs-io/babylon-toolkit/pull/2307#discussion_r3865457130).


knip.json
```diff
@@ -13,7 +13,7 @@
         "src/tbv/core/services/index.ts",
         "src/tbv/core/managers/index.ts",
         "src/tbv/core/contracts/index.ts",
-        "src/tbv/core/vault-secrets/index.ts",
+        "src/tbv/core/{vault-secrets,wasm}/index.ts",
```
gbarkhatov: Low severity, and the change is correct as written — but two literal lines would be strictly better than the brace here:

```json
"src/tbv/core/vault-secrets/index.ts",
"src/tbv/core/wasm/index.ts",
```

Three reasons, in increasing order of how much I'd weight them:

1. **Greppability.** `grep "wasm/index.ts" knip.json` currently returns nothing, which is unfortunate for a file whose whole job is to record what is deliberately registered.
2. **Convention.** Every other line in this `entry` array is one literal path.
3. **The evidence gap.** A literal path that stops resolving is still a path knip acts on; a *pattern* that matches nothing resolves to zero entries with no error — and knip exits 1 both before and after, so CI would not tell you. Your "before and after reports are identical" check can't discriminate "matched both" from "matched neither", because `core/index.ts:29` does `export * from "./vault-secrets"`, so under a total brace failure `vault-secrets/index.ts` stays barrel-reachable and the report looks unchanged either way.

To be clear, I don't think it *is* failing — this file already relies on brace expansion at lines 32, 33 and 37 (`*.{ts,tsx}`), and those demonstrably work or every vault test file would be reported unused. It's the ambiguity I'd rather not have in a guard file, for one extra line.

**Separately, a cross-PR coupling worth knowing about:** this registration is path-literal, so if #2466 or #2464 renames or moves `src/tbv/core/wasm/index.ts` — both modify that exact file — this entry silently un-registers with no CI signal. Not a defect here; just an ordering risk across the stack.

gbarkhatov (dismissed):
Reviewed at `ad09c9a9` with three independent passes. **No merge blockers.** The central question — does the new glob match exactly what you claim, and does it silently match more? — I settled empirically against the worktree rather than by reasoning about glob semantics: it expands to exactly `vault-secrets/index.ts` and `wasm/index.ts`, nothing else.

The registration is **necessary, not cosmetic**, and that is worth stating because it is not obvious from a one-line diff. `src/tbv/core/index.ts` re-exports eleven sibling directories — `primitives`, `utils`, `managers`, `clients`, `contracts`, `wots`, `services`, `vault-secrets`, `pegin-fingerprint`, `deposit-terms`, `recovery` — and conspicuously **not** `./wasm`. So `./tbv/core/wasm`, a published subpath at `package.json:62`, was the one export no registered barrel reached. That is exactly why it alone needed registering.

One low-severity suggestion inline; the rest is context.

### One note on the description's rationale

Half the safety argument cites `getPeginPayoutScriptInfo` needing to stay registered as published API. That symbol lives only in `packages/babylon-tbv-rust-wasm`, and the root knip script is `pnpm dlx knip@5.88.1 --workspace packages/babylon-ts-sdk --workspace services/vault` — that package is never passed, and `knip.json` declares no block for it. The claim is true, but it is not load-bearing for this change. The `TAP_INTERNAL_KEY` half does hold: it is defined at `wasm/constants.ts:9`, re-exported at `wasm/index.ts:29`, has no production consumer outside its own directory, and is reachable only via the published subpath — precisely the symbol at risk of being proposed for deletion.

### Checked and dismissed

- **Are three other published subpaths (`./tbv/core/utils/eth`, `./tbv/core/clients/eth`, `./tbv/core/clients/mempool`) missing from `entry` in the same way?** One agent raised this; it is wrong. All three are re-exported from barrels that *are* registered entries — `utils/index.ts:14` and `clients/index.ts:10,12`, registered at `knip.json:10-11`. Their exports are already reachable. There is no follow-up gap to file.
- **Could the widened glob swallow internals and silently weaken the dead-code guard?** No. The pattern terminates at `/index.ts`, so `wasm/constants.ts` and `wasm/value-guards.ts` stay fully under knip's reporting, as do the other `index.ts` files under `src/tbv/core/`.
- **Is the knip version pinned?** Yes, exact at `5.88.1`, and untouched by this diff.

### Verified clean

- Coverage of the published `exports` map is now complete — every subpath is registered directly or reachable via a registered barrel.
- Direction is right under CLAUDE.md §4/§9: registering an entry stops knip proposing deletion of frozen on-chain-binding surface.
- No runtime code, values, exports, copy, motion or tests changed. Zero behavioural surface.

knip.json
jonybur: Addressed in 617b47825fb392f85133ed6937edc7822bfe85b2.

Replaced the pattern with two literal paths on separate lines. Both resolve to the same existing files. The PR description now limits its rationale to the SDK entry covered by this check.

jonybur:
@greptileai Please review commit 617b47825fb392f85133ed6937edc7822bfe85b2.

Use two literal entry paths and limit the description to the SDK public API.

Both literal paths resolve to the same existing files as the old pattern. JSON and diff checks pass.

gbarkhatov (approved):

PR #2468, opened 2026-09-08
test(ts-sdk): state the loader cache reset behavior

jonybur:
## Outcome

The loader test states that it proves SDK cache reset. Its comment explains why a rejected promise must leave the cache.

## Change

Correct the test name and comments. Keep all assertions.

## Safety

Generated engine initialization failures remain latched. The test does not claim that those failures recover. #2367 covers browser download and compile retries.

## Verification

At `b61802c78f4c61bc51acfbdb2c7125c1103ac937`, 1,966 SDK unit tests and 7 built-package tests pass. Dependent builds, lint, and diff checks pass. Only wording changes; emitted code is identical.

## Links

Covers [the loader test wording finding](https://github.com/babylonlabs-io/babylon-toolkit/pull/2307#discussion_r3865457120).

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/wasm/__tests__/loader.test.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/wasm/__tests__/loader.test.ts
@@ -105,7 +105,7 @@ describe("loadRawTbvWasm", () => {
   });
 
   it("clears the SDK cache after an initWasm rejection", async () => {
-    // This mock proves that the SDK clears its cache after a rejection.
+    // Without the SDK cache reset, later calls would repeat this rejection.
     // The real engine keeps generated initialization failures latched.
     let initCalls = 0;
     vi.doMock(RAW, () => ({
--- a/packages/babylon-ts-sdk/src/tbv/core/wasm/__tests__/loader.test.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/wasm/__tests__/loader.test.ts
@@ -104,9 +104,9 @@ describe("loadRawTbvWasm", () => {
     expect(causeMessages(error)).toContain("wasm instantiation failed");
   });
 
-  it("clears the cache after an initWasm failure so a retry re-initializes", async () => {
-    // initWasm runs inside the cached promise, so a failure there has to clear
-    // the cache too or the engine is unusable for the rest of the session.
+  it("clears the SDK cache after an initWasm rejection", async () => {
+    // This mock proves that the SDK clears its cache after a rejection.
+    // The real engine keeps generated initialization failures latched.
     let initCalls = 0;
     vi.doMock(RAW, () => ({
       initWasm: async () => {


packages/babylon-ts-sdk/src/tbv/core/wasm/__tests__/loader.test.ts
```diff
@@ -104,9 +104,9 @@ describe("loadRawTbvWasm", () => {
     expect(causeMessages(error)).toContain("wasm instantiation failed");
   });
 
-  it("clears the cache after an initWasm failure so a retry re-initializes", async () => {
-    // initWasm runs inside the cached promise, so a failure there has to clear
-    // the cache too or the engine is unusable for the rest of the session.
+  it("clears the SDK cache after an initWasm rejection", async () => {
+    // This mock proves that the SDK clears its cache after a rejection.
```
gbarkhatov: Optional, take it or leave it — no failure mode either way.

This line restates the `it(...)` name directly above it, and it credits the mock, which only supplies a fail-then-succeed `initWasm`. What actually proves the behaviour is lines 120-124.

The line it replaced did carry production rationale — that a failure there has to clear the cache too, or the engine is unusable for the rest of the session — which is what motivates `wasm/index.ts:74`, and that reasoning is now absent from the file. If you want to keep the information, line 109 is fine as-is and only this line needs to change, e.g.:

```
// A latched SDK cache would replay this rejection for the rest of the session.
// The real engine keeps generated initialization failures latched.
```

Secondary, and below the bar on its own: read alone, line 109 invites "then why does the SDK clear at all?" The answer is that the clear is load-bearing for the *other* failure class — a failed resolve/read nulls `bindingsPromise` at `wasm-loader-node.ts:46`, so a retry genuinely recovers there.

gbarkhatov (dismissed):
Reviewed at `f3875d67` with three independent passes. **No merge blockers** — the rename is a genuine correction and I verified the central claim myself rather than taking it on trust.

The assertions at `loader.test.ts:120-124` (reject → resolve → `initCalls === 2`) can only pass if `rawWasmModulePromise` was set to `undefined` at `wasm/index.ts:74`; a retained cache would replay the rejection and never re-enter `.then`. The new name claims exactly that and stops.

Worth stating plainly, because it is the reverse of what a wording-only diff usually looks like: **the old name was the over-claim.** Its "so a retry re-initializes" asserted a production recovery that does not exist — `bindings.initSync(...)` at `wasm-loader-node.ts:53` sits *outside* the try/catch spanning 34-48, so a throw there leaves `bindingsPromise` non-null and latched for the rest of the session. Only the resolve/read path (nulled at line 46) actually recovers. So scoping the new comment to "generated initialization failures" is precisely right, not accidentally broad.

One optional nit inline, below.

### Checked and dismissed

- **Does the rename introduce an over-claim about engine restart?** No. Line 107 says "the SDK cache", and the resolving second call at 121 is the mock's, not attributed to the real engine.
- **Is the SDK's cache clear at `index.ts:74` pointless, given the engine latches?** No. `wasm-loader-node.ts:43-47` nulls `bindingsPromise` on a failed resolve/read, so the clear is load-bearing for that class. It is only inert on the post-`initSync` path.
- **Any stale reference to the old test name?** None. Grepped `packages/`, `services/` and `docs/` for the old phrasings — zero hits.

### Verified clean

- Name matches the assertions exactly; satisfies the test-philosophy rule that the name describes the exact behaviour tested.
- `wasm/index.ts` is **not** touched by this PR, so the §4/§9 frozen boundary is intact and the WASM-facade pin check is unaffected.
- No name collision among the three `it(...)` in `describe("loadRawTbvWasm")` — the line-50 test covers `loadTbvWasm`, a different function.
- `describe` labels, the `causeMessages` helper, and all 64 assertions are untouched.

packages/babylon-ts-sdk/src/tbv/core/wasm/__tests__/loader.test.ts
jonybur: Addressed in b61802c78f4c61bc51acfbdb2c7125c1103ac937.

The comment now explains that a retained SDK promise would replay the rejection for the rest of the session. The next line still states that generated initialization failures remain latched.

jonybur:
@greptileai Please review commit b61802c78f4c61bc51acfbdb2c7125c1103ac937.

Explain why a rejected SDK loader promise must leave the cache.

1,966 SDK unit tests and 7 built-package tests pass. Lint passes. Emitted code is unchanged.

gbarkhatov (approved):

PR #2469, opened 2026-09-08
docs(packages): correct loader and module comments

jonybur:
## Outcome

The code comments match the current loading behavior and published module path. Local checks pass.

## Change

- Describe the Node loader's asynchronous read on first use.
- Keep the import-type style rule and remove its inaccurate explanation.
- Use the full published path in the status module tag.

## Safety

The emitted code is unchanged. The engine remains an optional peer, so that accurate comment stays. The required owner reviews still apply to the Node entry.

## Verification

- Nx SDK and engine lint checks passed.
- Nx engine build and lazy-entry check passed.
- Nx SDK API documentation check passed. Generated documentation is unchanged.
- The patch adds 5 lines and removes 8. Diff checks pass.

## Links

Covers the accepted stale comment and module name findings from #2307. The optional-peer contract changed in #2336.

### DIFF
--- a/packages/babylon-tbv-rust-wasm/src/index-node.ts
+++ b/packages/babylon-tbv-rust-wasm/src/index-node.ts
@@ -1,9 +1,8 @@
 // Node.js entry point for the WASM bindings.
 //
-// Loads the committed web WASM binary synchronously from disk using
-// readFileSync and initializes it via initSync. This avoids fetch()-based
-// loading, which does not work in Node.js environments, and does not require
-// a separate wasm-pack --target nodejs build step.
+// The loader reads the committed web WASM binary asynchronously on first use.
+// It initializes the binary with initSync. No fetch() or separate
+// wasm-pack --target nodejs build is needed.
 
 import {
   getWasmBindings,
--- a/packages/babylon-ts-sdk/eslint.config.mjs
+++ b/packages/babylon-ts-sdk/eslint.config.mjs
@@ -79,9 +79,7 @@ export default defineConfig([
       ],
     },
   },
-  // `import { type X } from "pkg"` leaves a side-effect import behind under
-  // verbatimModuleSyntax, which would defeat allowTypeImports above. Keep
-  // every type-only import in the top-level `import type` form.
+  // Keep every type-only import in the top-level `import type` form.
   {
     files: ["src/**/*.ts"],
     rules: {
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/status/index.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/vault-provider/status/index.ts
@@ -6,7 +6,7 @@
  * for unauthenticated status reads and polling that must remain usable in an
  * Ethereum-only application session.
  *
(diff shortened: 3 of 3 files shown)


gbarkhatov (approved):
Verified all three new comments against the code they sit next to, in a worktree pinned at `41abbbcc`. All three are true, and the one they replace was genuinely false. One non-blocking completeness note.

### 1 — The stale claim this PR corrects survives verbatim in a sibling file. `packages/babylon-tbv-rust-wasm/scripts/build-wasm.js:246-248`. Non-blocking.

That file still says:

```js
// Copy generated files to dist/generated. The node entrypoint
// (src/index-node.ts) loads this same web artifact via readFileSync +
// initSync, so no separate nodejs-target build is needed.
```

This is the same sentence `index-node.ts` is being corrected for, so after this PR the repo states both the corrected and the uncorrected version of the same fact. It is wrong on both halves: the read is `await readFile` from `node:fs/promises` (`wasm-loader-node.ts:1,36`), not `readFileSync`; and `index-node.ts` does not load anything itself — it imports `getWasmBindings` from the loader.

Worth fixing here because `build-wasm.js` hosts the `VAULT_WASM_COMMIT` pin, which makes it a CLAUDE.md §4 registered file and exactly where someone checking the frozen boundary lands. Three words: drop `readFileSync +`, or say "reads the artifact and calls `initSync`". The load-bearing half — "no separate nodejs-target build is needed" — stays true either way.

No inline comment for this one: the file is not in the diff, so there is no hunk to anchor to.

### Checked and dismissed

- **Does deleting the eslint rationale lose a real constraint?** No. `verbatimModuleSyntax` has zero hits across `packages/babylon-ts-sdk` and `tsconfig.base.json`, so the deleted text's premise did not hold for this package and `allowTypeImports: true` was never at risk. The shorter line is a correction, not a loss, and it is accurate for the rule it precedes (`no-import-type-side-effects`).
- **Is the browser-side twin left half-fixed?** No. `src/index.ts` opens directly on imports and carries no file header, so there is no paired claim there. `build-wasm.js` is the only survivor.
- **Could the `@module` rename make the generated API docs stale?** No. `status/index.ts` is not a typedoc `entryPoints` member, and `verify` passed on this head.

### Verified clean

- The new `index-node.ts` header is true in every clause: `wasm-loader-node.ts:36` does `await readFile(...)`, memoized behind `bindingsPromise ??=` at line 31 so nothing runs at module load, and initializes via `initSync({ module: binary })` at line 53. No `fetch()` on the Node path.
- `@module tbv/core/clients/vault-provider/status` matches `package.json:57` verbatim. The old short form matched no export at all.
- The §4-registered `index-node.ts` really is comment-only — one hunk over header lines 1-9, no wrapper signature, argument, `@stability frozen` label or ordering moved. The §9 loader is untouched.
- No dangling reference from the removal: `allowTypeImports` occurs once repo-wide, in the block above, which still stands.

jonybur:
Follow-up: https://github.com/babylonlabs-io/babylon-toolkit/pull/2475

This corrects the stale Node loader comment. Executable code is unchanged.

PR #2474, opened 2026-09-09
feat(vault, ledger-vault): env-gated Speculos transport for the dApp E2E

jrwbabylonlab:


services/vault/src/main.tsx
```diff
@@ -31,14 +32,27 @@ window.addEventListener("vite:preloadError", () => {
...
+  );
+}
+
+// E2E-only (#2110): with NEXT_PUBLIC_TBV_E2E_SPECULOS_URL set, arm the Ledger
+// DMK's Speculos transport BEFORE first render so a driven connect can never
+// race the override. Returns undefined in production — the render call below
+// is then the same synchronous call it always was.
+const speculosReady = initSpeculosTransportForE2E();
+if (speculosReady !== undefined) {
+  void speculosReady.then(renderApp);
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — E2E-only path; production never enters it

services/vault/src/config/featureFlags.ts
```diff
@@ -267,4 +267,26 @@ export default {
...
+   * main/react/babylon), so a production visitor never downloads it.
+   * Why needed: the full-dApp emulator E2E (#2110) — WebHID needs a user
+   * gesture and real hardware; the emulator speaks HTTP.
+   * Default: unset (never set outside E2E runs).
+   *
+   * Non-boolean gating config, so it uses the NEXT_PUBLIC_ prefix without
+   * _FF_ (per rule 5).
+   */
+  get e2eSpeculosUrl(): string | undefined {
+    const url = process.env.NEXT_PUBLIC_TBV_E2E_SPECULOS_URL;
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — rule 5 in this file: non-boolean gating config drops _FF_ (same as TBV_DISABLED_BTC_WALLETS)

services/vault/src/config/featureFlags.ts
```diff
@@ -267,4 +267,26 @@ export default {
...
+    return url === undefined || url === "" ? undefined : url;
+  },
```
gbarkhatov, line 292: **Worth an answer before merge — the Speculos URL becomes the Ledger APDU transport with no validation and no compile-time gate.**

This is a plain runtime read: any non-empty string passes straight to `speculosTransportFactory` and becomes the endpoint every APDU crosses — DERIVE_CONTEXT_HASH, APPROVE_VAULT_INTENT, Pre-PegIn and payout signing. There's no `new URL()` parse, no scheme check, no loopback check, and no `.trim()` (so `"  "` reads as set, unlike the neighbouring `noticeBannerMessage:219-222` and `disabledBtcWallets:263-269`, which both normalise).

What makes it stand out is the in-file comparison: `isGodModePanelEnabled:238-243` hard-gates a *debug panel* on `import.meta.env.DEV`, with JSDoc explaining that a production build "forces it false at compile time (and lets the bundler drop the god-mode code)". A debug panel gets a stronger gate than the switch that redirects hardware signing. Rule 4 of this file's own header says flags are DevOps-configurable in mainnet environments, and `SECURITY_MODEL.md:101-102` names precisely this adversary — "Operator. Honest but fallible … a configuration mistake ships as a release".

Suggestion: reject any host that isn't `localhost`/`127.0.0.1`/`[::1]` and throw with the offending value, plus a build-time gate — `import.meta.env.DEV` if the E2E drives a dev build, otherwise a dedicated vite mode or the `NEXT_BUILD_E2E` that `vite.config.ts:85` already reads. The gate also makes the branch statically foldable, which fixes the bundle-claim comment separately.

services/vault/src/e2e/__tests__/speculosTransportBootstrap.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    // Explicit: the invoking shell may export the flag (a local Speculos
+    // session does exactly that) — the getter maps "" to undefined.
+    vi.stubEnv("NEXT_PUBLIC_TBV_E2E_SPECULOS_URL", "");
+
+    expect(initSpeculosTransportForE2E()).toBeUndefined();
+    expect(setDmkTransportOverride).not.toHaveBeenCalled();
+  });
+
+  it("arms the seam with the Speculos factory for the configured URL when the flag is set", async () => {
+    vi.stubEnv("NEXT_PUBLIC_TBV_E2E_SPECULOS_URL", "http://127.0.0.1:5055");
```
gbarkhatov: **Worth an answer before merge — the page CSP looks like it blocks this URL, so the transport can't connect.**

The Speculos kit reaches the emulator with `fetch()` (`HttpSpeculosDatasource`), so `connect-src` governs it. The vault's meta CSP at `services/vault/index.html:55` is:

```
connect-src 'self' https: wss: http://localhost:* ws://localhost:*
```

A plain-http loopback by **IP** matches none of those — CSP host-parts are literal strings, so `127.0.0.1` is not `localhost`, `'self'` is the app origin (different port), and `https:` is scheme-matched. Every documented Speculos URL in this repo is `http://127.0.0.1:5055` (`dmkTransport.e2e.test.ts:12`, `speculos.e2e.test.ts:33`, `speculosClient.test.ts:50`, and this line), and the meta tag is served in dev too, since the Playwright `webServer` runs `vite` rather than a preview build.

So the first real run would get a silent CSP block rather than a connection — and this unit test can't catch it, because it mocks the transport away. Two ways out: standardise on `http://localhost:5055` and say so in the bootstrap JSDoc, or add `http://127.0.0.1:*` beside the dev carve-out that `vite.config.ts:21-25` already documents for the Playwright mock servers.

Flagging as a question rather than a defect since I can't run the #2110 harness — if it's already pinned to `localhost`, this is moot and only the JSDoc needs to say so.

services/vault/src/main.tsx
```diff
@@ -31,14 +32,27 @@ window.addEventListener("vite:preloadError", () => {
...
+} else {
+  renderApp();
+}
```
gbarkhatov, line 68: A rejected bootstrap white-screens the app with no diagnostic.

`void speculosReady.then(renderApp)` has no rejection handler. `setDmkTransportOverride` throws by design (`dmkSession.ts:102-104`), and either dynamic import can fail — a stale-chunk 404 is exactly what this file's own `vite:preloadError` handler at line 31 exists for. On rejection `renderApp` never runs, so `#root` stays empty, and `ErrorBoundary`/`GlobalError` live *inside* the tree that never mounts, so nothing reports it. The E2E then hangs to timeout reading "app never rendered" instead of "Speculos bootstrap failed".

```js
void speculosReady.then(renderApp, (error) => {
  console.error("Speculos transport bootstrap failed", error);
  renderApp();
});
```

A failed E2E seam should degrade to a working app rather than a blank page.

services/vault/src/e2e/speculosTransportBootstrap.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * E2E-ONLY Speculos transport bootstrap (#2110 full-dApp emulator E2E).
+ *
+ * When `NEXT_PUBLIC_TBV_E2E_SPECULOS_URL` is set, arms the ledger-vault-signer
+ * transport seam ({@link setDmkTransportOverride}) with the Speculos HTTP
+ * transport BEFORE the app renders, so the wallet's first connect — which
+ * builds the DMK singleton and freezes its transport — can never race the
+ * override. With the flag unset this module does nothing and imports nothing:
+ * the two `@ledgerhq` test kits are reached only through the flag-guarded
+ * dynamic imports below, keeping production bundles speculos-free.
```
gbarkhatov: Both halves of this claim are stronger than what the build delivers.

"With the flag unset this module does nothing and imports nothing" — it statically imports `setDmkTransportOverride` from the signer barrel at line 15 regardless of the flag. Harmless in practice, but it isn't "imports nothing".

"keeping production bundles speculos-free" — `vite-plugin-environment@1.1.3` (`dist/index.js:32`) computes `keys = Object.keys(env)` when passed `"all"`, so it emits a `define` **only for variables present at build time**. With the var unset there's no define at all, the read resolves at runtime against the `process` shim, and Rollup can't prove the branch dead (the `featureFlags` getter indirection would block folding independently anyway). Both `@ledgerhq` kits are therefore emitted as chunks into every production `dist`.

The narrower claim over in `featureFlags.ts:277-280` — "a production visitor never downloads it" — does hold, since the chunk is lazy and unpreloaded. That one's fine. It's "bundles stay speculos-free" that isn't. Either reword to "not downloaded in production", or make it literally true with the build-time gate suggested on `featureFlags.ts`.

packages/babylon-ledger-vault-signer/src/dmkSession.ts
```diff
@@ -83,9 +83,13 @@ let transportOverride: DmkTransportOverride | undefined;
  * The injector passes the factory in; this module never imports that package,
  * so production bundles stay speculos-free.
```
gbarkhatov, line 84: Two doc points on this file, both cheap, and it's a CLAUDE.md §7 critical-path file where a line-walking reviewer reads the header first.

1. The premise here still holds — `dmkSession.ts` genuinely never imports the speculos package — but the conclusion drawn from it no longer follows. The injector is now `services/vault/src/e2e/speculosTransportBootstrap.ts`, which ships inside the vault's production module graph and does import the package, so "production bundles stay speculos-free" is no longer true downstream even though this module is still clean. Worth scoping the sentence to this package.

2. Line 5 of the module header still reads "swappable pre-build through the **test-only** `setDmkTransportOverride` seam", while this same PR rewrites the function's JSDoc just below to declare it public surface whose intended caller ships in a production bundle. A header contradicting the exported contract is the first thing the next reviewer hits. One-word fix: "test/E2E-only".

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -51,7 +51,10 @@ export const LEDGER_VAULT_WALLET_ID = "ledger_btc_vault";
 // WebHID (`navigator.hid` — desktop Chromium only, secure context); without it
 // the entry still renders clickable and only fails on connect, so hide it up
 // front. `in` check because TS's DOM lib does not declare `Navigator.hid`.
-const isWebHidAvailable = "hid" in navigator;
+// E2E exception: with the Speculos transport armed (#2110) the DMK never
+// touches WebHID, so the emulator flag stands in for device availability.
+const isWebHidAvailable =
+  "hid" in navigator || featureFlags.e2eSpeculosUrl !== undefined;
```
gbarkhatov: This now gates on configuration rather than capability, and the constant's name stops being true.

`isWebHidAvailable` becomes true whenever the env string is non-empty, with no check that the override actually armed. `speculosTransportFactory` never contacts the server — it only builds a factory closure (confirmed against the pinned 1.2.1 source) — so a wrong port or a dead emulator still opens the row, and the failure surfaces deep inside DMK discovery on connect. That's precisely the "still renders clickable and only fails on connect" outcome the original two-term gate was written to prevent, per the comment three lines up. It also now reports "WebHID available" in a browser where WebHID is absent.

Suggestion: export an `isSpeculosTransportArmed()` from the bootstrap, set only after `setDmkTransportOverride` returns, gate on that, and rename the constant to something honest like `canReachLedgerDevice`. That closes the ordering gap too — this constant is evaluated at module load, which happens before `main.tsx`'s body runs `initSpeculosTransportForE2E()`.

services/vault/src/e2e/__tests__/speculosTransportBootstrap.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  it("returns undefined and arms nothing when the env flag is unset — the production path", () => {
+    // Explicit: the invoking shell may export the flag (a local Speculos
+    // session does exactly that) — the getter maps "" to undefined.
+    vi.stubEnv("NEXT_PUBLIC_TBV_E2E_SPECULOS_URL", "");
+
+    expect(initSpeculosTransportForE2E()).toBeUndefined();
+    expect(setDmkTransportOverride).not.toHaveBeenCalled();
+  });
```
gbarkhatov: The test named for the unset flag never exercises the unset branch.

`vi.stubEnv(key, "")` exercises the `url === ""` arm; the `url === undefined` arm — the actual production path the name claims — is never hit. That branch is the interesting one, since with the var genuinely absent `vite-plugin-environment` emits no define at all and the value comes back from the `process` shim.

Two other gaps in the same file, both cheap:

- Both `@ledgerhq` kits are `vi.mock`ed at module scope, so the headline claim in the module JSDoc ("with the flag unset this module … imports nothing") would pass identically if the module imported them statically. vitest calls a `vi.mock` factory lazily on first import, so asserting the factories were never invoked in the unset case actually pins that claim.
- No coverage for a rejected dynamic import — which is the gap that lets the `main.tsx` blank-page case through.

Per the repo's test philosophy, three separate `it`s beat parameterising this.

packages/babylon-ledger-vault-signer/src/index.ts
```diff
@@ -13,7 +13,15 @@
 
 export { getExtendedPublicKey, getMasterFingerprintHex, getXOnlyPublicKeyHex } from "./derivation";
 export { createDmkApduSender, createDmkRawApduSender } from "./dmkApduSender";
-export { closeDmk, connectDmkSession, disconnectDmkSession, isSessionAlive, type DmkSessionHandle } from "./dmkSession";
+export {
+  closeDmk,
+  connectDmkSession,
+  disconnectDmkSession,
+  isSessionAlive,
+  setDmkTransportOverride,
```
gbarkhatov: Promoting this seam documents the missing barrier rather than adding one.

The package's `exports` map exposes only `"."`, so before this PR the setter was genuinely unreachable from outside despite being a module export — that was the barrier. It's now callable by any consumer of a `publishConfig.access: public` package, guarded solely by call ordering (`dmkSession.ts:101-105`), and the new JSDoc openly concedes "A pre-build call from production code would take effect".

A `./testing` subpath in the `exports` map would let the dApp bootstrap import it while keeping it off the main production surface, and it composes with the build-time gate suggested on `featureFlags.ts`. The package already has an `@internal` convention for this kind of thing (`signPsbtPrepare.ts:247,257`), and neither is used here.

Noting for process rather than as a finding: §7 wants two reviewers and an author line-walk on this file, and the PR body is empty, so there's no line-walk declaration recorded.

packages/babylon-ledger-vault-signer/src/__tests__/dmkSession.test.ts
```diff
@@ -248,6 +248,11 @@ describe("setDmkTransportOverride", () => {
   };
   const FAKE_IDENTIFIER = "FAKE-SPECULOS";
 
+  it("is part of the package's public surface — the dApp E2E bootstrap imports it from the index", async () => {
+    const pkg = await import("../index");
+    expect(typeof pkg.setDmkTransportOverride).toBe("function");
+  });
```
gbarkhatov: This asserts a wiring pin rather than a production behaviour.

`expect(typeof pkg.setDmkTransportOverride).toBe("function")` restates what `tsc` already enforces on the export line, and it pulls the whole package barrel in under this file's DMK mocks to do it. The repo's test philosophy asks what production behaviour a test verifies; this one verifies that a re-export exists.

It does fail if the export is dropped, so it isn't worthless — but a test asserting the *now-public* setter still refuses a post-build swap would cover the export and the invariant that actually matters, which is the one the JSDoc says is the only thing the code enforces.

gbarkhatov (dismissed):
Reviewed with three independent passes, then deduped and verified against the source at `e05a4b19`. Approving — none of this blocks you unilaterally, you own the calls. Two findings I'd want an answer on before merge are inline and marked; the rest are suggestions.

The two marked ones: the Speculos URL becomes the Ledger APDU transport with **no validation and no compile-time gate** (a debug panel twenty lines above it in the same file gets a harder gate than hardware signing does), and the page CSP appears to **block the very URL this PR's own test uses**.

### Three findings that have no diff line to anchor to

**A — An exported `NEXT_PUBLIC_TBV_E2E_SPECULOS_URL` will contaminate every other vault Playwright suite.** `services/vault/playwright.config.ts:32,140-143`

The `webServer` blocks inherit the parent shell. The config already pins `NEXT_PUBLIC_SENTRY_DSN: ""` for exactly this reason, with a comment saying "so a developer's exported NEXT_PUBLIC_SENTRY_DSN can't enable Sentry here". The new test's own comment concedes "the invoking shell may export the flag (a local Speculos session does exactly that)" but defends only the empty-string case. A developer with the var exported runs the behavioural, god-mode and visual suites with the WebHID gate relaxed and the Speculos transport armed. Adding `NEXT_PUBLIC_TBV_E2E_SPECULOS_URL: ""` to `MOCK_ENV_VARS` mirrors the Sentry precedent.

**B — The variable is documented nowhere an operator reads, and it makes an existing `.env.example` statement false.** `services/vault/.env.example:96-108`

Grepping for `TBV_E2E_SPECULOS` returns only the four files this PR touches — no `.env.example`, no README, no CI workflow, and notably not `services/vault/e2e/real/`, so the #2110 harness that would consume this doesn't exist yet. Meanwhile the Ledger flag's own entry at line 107 ends "the entry is hidden in browsers without navigator.hid regardless of this flag", which this PR makes untrue. `scripts/sync-env.mjs --check` only diffs the network `FIELD_MAP`, so no CI job catches the omission. Worth a commented entry naming the E2E-only constraint, the loopback requirement, and the companion `NEXT_PUBLIC_FF_ENABLE_LEDGER_VAULT_WALLET=true` — plus amending that Ledger comment.

**C — Minor: `src/e2e/` is production source under a name vitest excludes by pattern.** `services/vault/src/e2e/`

`main.tsx` imports it statically, so it ships, yet it sits in a directory named like the Playwright tree at `services/vault/e2e/`. `vitest.config.ts:102-103` excludes `**/e2e/**/*.spec.ts` and `**/e2e/helpers/**`; the new test survives only because it is `.test.ts` and not under `helpers/`. A future tightening to `**/e2e/**` would delete this module's only coverage with a fully green run. `src/bootstrap/` would leave `e2e` meaning test infrastructure.

### Raised during review and disproved — so nobody re-raises them

- **Feature-flag naming violation** (`NEXT_PUBLIC_` without `_FF_`) — rule 5 of the file header permits non-boolean gating config to use other prefixes, and `NEXT_PUBLIC_NOTICE_BANNER_MESSAGE` / `NEXT_PUBLIC_TBV_DISABLED_BTC_WALLETS` are established precedents. The JSDoc cites the rule correctly.
- **devDependencies break a production build** — vite and typescript are themselves devDeps, so `pnpm install --prod` can't build the vault at all, and no workflow uses it.
- **Unset variable causes a `ReferenceError` on `process`** — `public/process-polyfill.js` sets `window.process.env = {}` and `nodePolyfills` leaves `globals.process` at its default.
- **Eager-bundle regression from the static signer import** — the signer sets `sideEffects: false`, only `import type`s the DMK, and was already eager via wallet-connector's static import chain.
- **Runtime `window.process` injection vector** — `@rollup/plugin-inject` binds `process` module-locally.

### Verified clean

- `speculosTransportFactory(url, isE2E, DeviceModelId.NANO_SP)` and `speculosIdentifier` match the pinned v1.2.1 declaration in arity, order and semantics, and match the sibling `dmkTransport.e2e.test.ts:47` exactly. `isE2E` genuinely skips the 2 s `listenForDisconnect` interval; the third argument genuinely overrides the `STAX` default.
- One `dmkSession` singleton in a built app: the signer is external in wallet-connector's lib build (`scripts/package-boundary.js:38`), is a real vault dependency, and both resolve the identical peer-suffixed DMK `1.7.1(bufferutil@4.0.9)(rxjs@7.8.2)(utf-8-validate@5.0.10)`.
- Arming timing is sound — static imports fully evaluate before `main.tsx`'s body, nothing builds a DMK at import time, and `dmkSession.ts:101-106` still throws loudly on a post-build swap, so a driven connect cannot race it.
- The two-term Ledger gate survives: `isLedgerVaultWalletEnabled && isWebHidAvailable` still stands; the flag substitutes only for the WebHID term.
- Dependency hygiene: both kits pinned exact (`1.7.1`, `1.2.1`), no new lockfile `packages:`/`snapshots:` entries, no duplicate DMK. The `babylon-ts-sdk` importer hunk is an alphabetical re-sort.
- `optimizeDeps.include` is correct and dev-only in Vite 6 — with wallet-connector force-optimized, an unlisted linked signer would be pre-bundled as a second `dmkSession` singleton. The comment correctly scopes its claim to dev.
- No CLAUDE.md §7 encoding surface touched: no APDU/TLV changes, no golden vectors, no payload-byte logging.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

jonybur (dismissed):
Reviewed at e05a4b19. Approving — I found nothing substantive beyond what is already on the thread, and CI is green.

Where I went looking, and what it came back as:

- **The section 7 surface is untouched.** No APDU layout, TLV encoding, envelope gate or golden vector changes; dmkSession.ts gains only JSDoc, and index.ts only re-exports two existing symbols. No payload bytes or key material reach a log — the deliberate absence of addLogger on both build paths survives.
- **The transport-override lifecycle is sound.** getDmk reads transportOverride exactly once per build and pairs the identifier with the DMK that registered it, so connectDmkSession can never discover against a transport the instance does not hold. The setter's guard keys on dmkPromise, which closeDmk clears, so the throw covers both the built and the still-building states.
- **The DMK singleton stays single.** The two new devDependencies resolve to the same peer-suffixed 1.7.1 the signer already depends on, so no second copy lands, and the optimizeDeps entry is the dev-mode counterpart of that.
- **The wallet-row gate keeps both terms.** isLedgerVaultWalletEnabled is still required; the emulator flag substitutes only for the WebHID term.
- No new user-facing string outside copy.ts, and no data-testid moved, renamed or dropped.

The open threads on the URL validation and the build-time gate, the CSP question, and the rejected-bootstrap blank page are the ones I would want answered before merge, but they are already raised and are yours to judge — not gating this approval.

services/vault/src/config/featureFlags.ts
jrwbabylonlab: fixed

services/vault/src/e2e/__tests__/speculosTransportBootstrap.test.ts
jrwbabylonlab: fixed

services/vault/src/main.tsx
jrwbabylonlab: fixed

services/vault/src/e2e/speculosTransportBootstrap.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/dmkSession.ts
jrwbabylonlab: fixed

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jrwbabylonlab: fixed

services/vault/src/e2e/__tests__/speculosTransportBootstrap.test.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/index.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/dmkSession.test.ts
jrwbabylonlab: fixed

services/vault/src/main.tsx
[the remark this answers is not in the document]
jrwbabylonlab: fixed

jeremy-babylonlabs (approved):

PR #2481, opened 2026-09-09
feat(wasm)!: guard HTLC and payout connector outputs

jonybur:
## Outcome

HTLC and payout connectors reject WASM output fields that differ from their
independently derived values. Both raw entries and async factories use the
checks before returning connector data.

## Change

- Derive expected scripts, control blocks, addresses, hashes, and versions.
- Check SDK payout scripts against the trusted request.
- Reject malformed inputs and unsupported versions before use.
- Preserve validation and cleanup errors. Release each object once.
- Keep tests for changed outputs, supported versions, networks, and input limits.

## Safety

This is a breaking engine change. Guarded wrappers change raw class identity
and reject unchecked inputs. Release the engine and SDK together.

Callers must still verify constructor inputs against on-chain data. The
independent TypeScript derivations share a package with the WASM binary.
The accepted package-provenance limits remain documented. Raw transaction
guards remain in #2493.

Two qualified final reviews and the required author line review remain.

## Verification

- All CI checks pass on `fe24681d0fac7b0a883405f007b83df682cf7c66`.
- Greptile reports 5/5 for `fe24681d0fac7b0a883405f007b83df682cf7c66`. All 16 review threads are resolved.
- Affected builds, lint, and tests pass across 10 projects and their prerequisite.
- Engine and SDK builds and lint pass.
- All 35 engine tests, 2,008 SDK tests, and nine built-package checks pass.
- Engine lazy-entry checks and SDK import-boundary checks pass.
- All 34 SDK browser tests and the packed SDK consumer check pass.
- All 23 Vault browser tests pass. The separate layout run passes all six tests.
- Packed wallet consumers and the release-driver type check pass.
- All three wallet browser tests pass on an unchanged second run. The first run had an OKX setup timeout.
- All 16 staking browser tests pass in Chromium and Firefox.
- Dependency version checks pass. Generated API docs have no changes.
- The engine tarball contains all export targets, guard modules, types, and WASM. Packed files match the tested build.
- The release dry run completes. Its publish simulation stops at the unchanged, already published engine version. Separate engine, SDK, and wallet package checks pass.

## Links

Tracks #2361. #2361 stays open. Parent: #2231. Related epic: #2228.

Includes the payout work from #2492 and #2509. Follow-up: #2493.


jonybur:

Please review the full PR at current commit `3c50224669b895b628affac6e52dfa95e34b2f27`. Check the diff against its target branch. Report any remaining issues and the confidence score for this commit.

packages/babylon-tbv-rust-wasm/src/raw.ts
```diff
@@ -19,9 +20,9 @@ export const WasmPrePeginTx: typeof Bindings.WasmPrePeginTx =
   raw.WasmPrePeginTx;
 export type WasmPrePeginTx = Bindings.WasmPrePeginTx;
 
-/** @deprecated Raw access skips SDK value checks. Use buildPeginInputPsbt or buildRefundPsbt from @babylonlabs-io/ts-sdk/tbv/core/primitives. */
+/** @deprecated Use buildPeginInputPsbt or buildRefundPsbt from @babylonlabs-io/ts-sdk/tbv/core/primitives. This class checks its HTLC fields against the constructor inputs. */
 export const WasmPrePeginHtlcConnector: typeof Bindings.WasmPrePeginHtlcConnector =
-  raw.WasmPrePeginHtlcConnector;
+  GuardedPrePeginHtlcConnector;
```
gbarkhatov, line 26: **MERGE BLOCKER** — the guard is wired only into `/raw`, which has no in-repo connector consumer, so it currently protects zero call sites.

`raw.ts` and `raw-node.ts` substitute `GuardedPrePeginHtlcConnector`, but the facade `getPrePeginHtlcConnectorInfo` (`src/index.ts:161-188`, unchanged by this PR) still does `const { WasmPrePeginHtlcConnector } = await getWasmBindings()` and returns hashlock/refund scripts, both control blocks, the address and the scriptPubKey **unchecked** — and that is the path with real callers. I confirmed two production ones: `deposit-terms/rebuildDepositTermsCore.ts:193` and `primitives/psbt/peginInput.ts:101`, both routing through `core/wasm/index.ts:148-151`. Meanwhile the only SDK `/raw` consumer, `refund.ts:75`, destructures `WasmPrePeginTx` and never touches the connector at all.

Not a signing bypass, since the signing builders re-derive independently — the narrower failure is that a wrong engine result silently mis-drives candidate discrimination and the deposit-terms rebuild, where `rebuildDepositTermsCore` treats `connector.scriptPubKey` as the authority it compares against. The fix looks like one identifier in two files: construct the guarded class inside `getPrePeginHtlcConnectorInfo`.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
```diff
@@ -63,159 +49,12 @@ type PrePeginHtlcParams = Pick<
...
-  return {
-    hashlockScript,
-    hashlockControlBlock,
-    refundScript,
-    refundControlBlock,
-    scriptPubKey: output,
-    tapMerkleRoot: hash,
-  };
+) {
+  return (await loadTbvWasm()).deriveExpectedPrePeginHtlc(params, hashlock);
```
gbarkhatov, line 57: **MERGE BLOCKER (wants an explicit §1 decision)** — the independent check now ships from, and loads through, the same npm package as the binary it checks.

To be precise about what did *not* regress: `prePeginHtlc.ts` is pure TypeScript with no wasm or generated-binding reference, so the derivation algorithm remains independent of the Rust binary and a doctored `.wasm` is still caught. What regressed is provenance. The SDK now gets its second opinion via `loadTbvWasm()`, so the value and its expected value ship in one artifact. Before this PR, defeating the cross-check meant compromising two separately published packages; now a single compromised — or merely stale or version-skewed — publish of the engine controls both sides, and `refund.ts:127`, `peginInput.ts:111` and the loop at `assertWasmPeginSizing.ts:372-382` lose their guard together.

The docstring at line 64 still says "Pure-JS, binary-INDEPENDENT", which is true of the algorithm but no longer of where it comes from. Either keep `prePeginHtlc.ts` in the SDK (or a leaf package that doesn't depend on the engine) and have `rawHtlcConnector.ts` import *it*, inverting the direction — or amend §1 and SECURITY.md to state that the guard is now co-located with the guarded binary, and why that is accepted.

Smaller, same lines: this wrapper also lost the explicit `ExpectedPrePeginHtlc` return annotation the pre-PR version had, so no consumer can name the shape of a value that feeds tapLeafScript entries and scriptPubKey comparisons. And since the engine is an *optional* peer, a version-skewed consumer gets `…deriveExpectedPrePeginHtlc is not a function` mid-signing rather than a clear error.

packages/babylon-tbv-rust-wasm/src/prePeginHtlc.ts
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/**
+ * Derive the canonical Pre-PegIn HTLC without using vault WASM output.
+ */
+export function deriveExpectedPrePeginHtlc(
+  params: PrePeginHtlcParams,
+  hashlock: string,
+): ExpectedPrePeginHtlc {
+  initEccLib(ecc);
```
gbarkhatov: **MERGE BLOCKER** — `initEccLib(ecc)` silently replaced `assertEccInitialized()`, turning a pure derivation into a per-call mutation of bitcoinjs-lib global state.

I diffed this against the pre-PR SDK copy: it previously *asserted* that the host had installed a backend (`assertEccInitialized`, still live at `primitives/utils/bitcoin.ts:302`). bitcoinjs-lib keeps a single module-global ECC cache, so the engine now overrides whatever the host installed on **every call** — once per HTLC inside the loop at `assertWasmPeginSizing.ts:372-382`. Engine and SDK both pin `bitcoinjs-lib@6.1.7`, so under dedupe they share one instance, and a consumer who installed the native `tiny-secp256k1` has it swapped for asm.js process-wide for every subsequent taproot sign and verify.

It also drops the actionable error in favour of bitcoinjs's generic "No ECC Library provided". This is benign inside this monorepo only because everything pins the same asm.js build, which is exactly why CI is green on it. Suggest hoisting to a single module-scope `initEccLib(ecc)` at import time, or keeping the assert-don't-mutate contract.

packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts
```diff
@@ -0,0 +1,145 @@
...
+      depositor,
+      provider,
+      keepers,
+      challengers,
+      hashlock,
+      timelock,
+    ] = args;
+    // The independent derivation covers the three pinned graph versions.
+    if (version !== 1 && version !== 2 && version !== 3) {
+      throw new Error(`Unsupported HTLC graph version: ${version}.`);
```
gbarkhatov, line 36: Supported graph versions are hardcoded as `1 && 2 && 3` in a file this PR just registered under §1 — an unlinked second copy of the Rust authority (`SUPPORTED_TX_GRAPH_VERSIONS` / `parse_version` at the pin), whose own doc comment notes that adding a version there is a compile error at every Rust dispatch site. This JS copy is not.

The generated module this file already imports exports `supportedTxGraphVersions(): Uint16Array` as a sync free function, and raw callers must have run `initWasm()` first, so it is callable right here. When v4 lands, the raw connector rejects it while `getPrePeginHtlcConnectorInfo` accepts it. It fails closed, so this is availability rather than correctness — but it is also a no-magic-numbers violation on a §1 path.

packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts
```diff
@@ -0,0 +1,145 @@
...
+    this.#inner = new RawConnector(...args);
+    try {
+      this.getTxGraphVersion();
+      this.getHashlockScript();
+      this.getHashlockControlBlock();
+      this.getRefundScript();
+      this.getRefundControlBlock();
+      this.getScriptPubKey('bitcoin');
+    } catch (error) {
+      this.#inner.free();
```
gbarkhatov: This cleanup path can replace the guard-mismatch error with an unrelated memory error. If `#inner.free()` throws here — an already-freed pointer, or detached memory after a second wasm init, a race this package's own tests exercise at `tests/wasm-loader.test.mjs:286` — that throw propagates and the original `WASM HTLC <field> does not match the constructor inputs.` is lost.

That is the one diagnostic saying a doctored binary was caught, so losing it means an operator sees a loader bug instead of a value-mismatch alarm. Wrapping the free in its own try/catch would let the guard error always win.

packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts
```diff
@@ -0,0 +1,145 @@
...
+      throw new Error(
+        'WASM HTLC version does not match the constructor input.',
+      );
+    }
+    return version;
+  }
+
+  free(): void {
+    this.#inner.free();
+  }
```
gbarkhatov, line 162: `free()` is not idempotent, so `free()` followed by `[Symbol.dispose]()` double-frees a wasm-bindgen pointer. `free()` calls `this.#inner.free()` and `[Symbol.dispose]()` calls `this.free()`, with no guard flag between them; wasm-bindgen's `free()` zeroes `__wbg_ptr` via `__destroy_into_raw()`, so the second call passes a null pointer.

This is reachable with the pattern the codebase already uses on this exact class — `src/index.ts:176-187` frees it in a `finally`, so `using c = new WasmPrePeginHtlcConnector(...)` alongside that would free twice. Since this PR makes the wrapper the thing exported under the raw name *and* adds the dispose alias, the guard belongs here: a `#freed` boolean with an early return from both.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
```diff
@@ -457,3 +472,210 @@ test('createPayoutConnector forwards the network to the address through the brow
...
+              pinned.getRefundScript(),
+              pinned.getRefundControlBlock(),
+            ].join('|'),
+          ),
+          '625a8b363190adc2e1f6b3edea52cbed792b96967dbb2bb5be5e2635ad1598cf',
+        );
+      } finally {
+        pinned.free();
+      }
+      for (let sample = 0; sample < 12; sample += 1) {
```
gbarkhatov, line 872: The randomised differential varies the key *bytes* but never the input *shape*, so §9's "plus randomised inputs" is met in form more than in substance.

Every one of the 12 samples uses exactly 2 keepers and exactly 2 challengers, and the timelock is `1 + sample * 31`, topping out at 342 of a declared 1..65535 range; the deterministic test above adds only 1- and 2-keeper cases. So the `OP_CHECKSIG`/`OP_CHECKSIGADD` chain in `nOfNChunks` is never differentially checked at 3 or more keys per group, and `bscript.number.encode`'s switch to a 3-byte sign-extended push at ≥32768 is never compared against Rust at all.

Concretely: a vault with 3+ keepers, or a timelock ≥32768, could diverge between the TS derivation and Rust with this whole suite green. Drawing group sizes in 1..4 and timelocks spanning the encoder boundaries (1, 127, 128, 32767, 32768, 65535) would close it.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
```diff
@@ -198,6 +204,10 @@ for (const [rawName, loaderName] of [
 
 // Without generated code, facade entries must import and raw entries must fail.
 const isolatedPackage = mkdtempSync(join(tmpdir(), 'tbv-wasm-lazy-'));
+symlinkSync(
+  resolve(packageRoot, 'node_modules'),
```
gbarkhatov: Symlinking the real `node_modules` into the isolated package disarms the half of this harness that proved the Ethereum-only root imports no Bitcoin stack.

The facade check only asserts the module *imports successfully*. Previously the isolated dir had no `node_modules`, so any bare static import from `dist/index.js` failed resolution and got reported — incidental, but load-bearing. The symlink is genuinely needed here, because `raw.js` now statically reaches `rawHtlcConnector.js` and therefore `bitcoinjs-lib`, which would otherwise fail the raw check's filter.

But granting `node_modules` to the whole isolated package switches off the facade half: turning the runtime `await import('./prePeginHtlc.js')` at `src/index.ts:501` into a static import would drop bitcoinjs-lib onto the Ethereum-only path with nothing failing — the exact property the migration guide asserts. Two temp dirs, keeping the facade check `node_modules`-free, would preserve both.

packages/babylon-tbv-rust-wasm/package.json
```diff
@@ -52,5 +52,11 @@
     "@internal/eslint-config": "workspace:*",
     "shelljs": "^0.8.5",
     "typescript": "^5.8.3"
+  },
+  "dependencies": {
+    "@bitcoin-js/tiny-secp256k1-asmjs": "2.2.3",
+    "@noble/curves": "1.9.7",
```
gbarkhatov, line 58: The engine declares the Bitcoin stack as hard `dependencies`, while the SDK declares the same two packages as *optional peers* (`babylon-ts-sdk/package.json`, `bitcoinjs-lib` and `@bitcoin-js/tiny-secp256k1-asmjs` with `optional: true`) — and `raw-engine-migration.md` says they are the same: "The engine now declares the same pinned Bitcoin and curve dependencies used by the SDK." The versions do match (6.1.7, 2.2.3); the declaration kind does not, so that sentence is inaccurate as written.

The practical consequence is that a consumer pinning a different bitcoinjs-lib gets a second nested copy — extra asm.js in the bundle, and two independent ECC globals of which the app can only ever initialise one. That compounds the `initEccLib` finding on `prePeginHtlc.ts`. Mirroring the SDK (those two as optional peers, `@noble/curves` and `buffer` as real deps) would keep one instance.

CLAUDE.md
```diff
@@ -58,8 +60,8 @@ These paths handle irreversible value movement. An AI-generated mistake here is
...
+- The package exports a second crossing at `@babylonlabs-io/babylon-tbv-rust-wasm/raw` (`src/raw.ts`, `src/raw-node.ts`, registered in section 9). The HTLC connector checks its scripts, control blocks, address, and graph version against its constructor inputs. The other three classes remain unguarded, so their callers must apply the rule above. The only SDK consumer is `packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/refund.ts`. It guards `pegInAmounts` with `assertPositiveBigintArray`, derives the canonical HTLC and signing data in TypeScript, compares the WASM template's script and value with the funded output, and checks the final refund input and output before it emits the PSBT. Every new `/raw` consumer must do equivalent cross-checks.
+- The raw classes and SDK raw loader are deprecated. Only the raw HTLC connector has complete output checks. The other raw paths still permit a bypass. See the [migration guide](packages/babylon-ts-sdk/docs/guides/raw-engine-migration.md) for guarded alternatives and the compatibility blocker in #2361.
```
gbarkhatov: "Only the raw HTLC connector has complete output checks" overclaims, and a future reader will quote it as permission to skip the call-site cross-check.

Line 63 is precise — the guard checks "against its constructor inputs". But the guard re-derives from the *caller's own arguments*; it never verifies those arguments are the canonical on-chain keeper/challenger sets, hashlock or timelock, which is what §1's rule actually requires. So "complete output checks" reads as stronger than it is: substitute a keeper set and the connector produces self-consistent, fully "checked" scripts.

Suggest something like "checks internal consistency against its own constructor inputs; callers must still verify those inputs against on-chain data", and keeping §1 binding on all four raw classes. Same issue in `SECURITY.md:218-224`, where every sentence is individually true but pairing "rawHtlcConnector.ts checks the HTLC connector" with "the only SDK consumer is refund.ts" implies the SDK's raw crossing is covered — `refund.ts:75` uses `WasmPrePeginTx`, still unguarded.

gbarkhatov (dismissed):
Reviewed at `3c50224` via three independent passes (7 agents total), deduped, with every finding below re-checked against the source.

**The guard itself is good work** — it is fail-closed, it covers all seven value-bearing members of the generated class, `#inner` is unreachable, and the §9 differential is real: versions 1-3 across five networks against the *actual* generated WASM class, with a pinned vector plus randomised keys. The problems are about *where it was wired* and *what shipped alongside it*, not the guard's design.

Three findings I'd want answered before this merges (marked MERGE BLOCKER inline). None of them is a signing bypass — `buildRefundPsbt` and `buildPeginInputPsbt` still re-derive independently — so I'm approving rather than holding the branch, per how we treat review comments here. But the first one in particular means the guard currently protects zero in-repo call sites, which is probably not the intent.

- *"The moved derivation silently drifted."* No. Normalising whitespace and quotes, the helpers and the whole script-construction body hash identical between the old SDK copy and the new engine copy. `initEccLib` is the sole behavioural delta in the entire move.
- *"The independent check now checks itself."* No — I read `prePeginHtlc.ts`, and it is pure TypeScript with zero wasm or generated-binding references. The algorithm stays independent of the Rust binary; a doctored `.wasm` is still caught. The real regression is packaging, not self-reference — see the comment on `assertWasmPeginSizing.ts`.
- *"§9's differential-test requirement is unmet."* No, it is genuinely met. My comment on the test is about depth of the randomisation, not its existence.
- *"The class-identity swap is a type lie."* No. The generated class has zero static members, no generated function accepts or returns one, and instance signatures match.
- *"`getScriptPubKey` discarding its network lookup is a bug."* No — a P2TR scriptPubKey is network-independent; the lookup exists to reject unknown network strings.

**Verified clean:** no enforcement guard weakened (both new files added additively to `critical-path-check.yml`, the real `paths` array in `claude-md-drift.yml`, CODEOWNERS, and the *strict* eslint block); §4's frozen surface untouched — the `index.ts` / `index-node.ts` diffs are pure EOF appends and `VAULT_WASM_COMMIT` is unperturbed; the sync-to-async conversion is complete with every call site awaited and no coverage deleted; no dead code from the -182 deletion (`assertEccInitialized` and `TAPSCRIPT_LEAF_VERSION` both still used elsewhere); the `#network` map is faithful to rust-bitcoin `Network::from_str` at the pin.

*Provenance: I confirmed the three blockers, the `free()`/catch/version-constant findings, and both hash comparisons directly against the worktree. The test-depth, lazy-entry-harness, dependency-declaration and doc findings are reported from the review lanes with their own verification, not re-derived by me.*

jonybur:

I checked each of the ten open threads against a prepared fix set. I have not pushed anything. A push would dismiss the standing approval (`dismiss_stale_reviews_on_push` is on) while most threads stay open, so the fixes wait until they are complete.

The main ruleset also sets `required_review_thread_resolution`. All ten threads must close before this can merge.

### Done (3)

- **`raw.ts:25` — guard reaches the facade.** Both `index.ts` and `index-node.ts` now build the guarded class inside `getPrePeginHtlcConnectorInfo`, so the production callers get checked scripts, control blocks, address and scriptPubKey. Facade-level rejection tests cover both entries.
- **`rawHtlcConnector.ts` — double free.** A `#freed` flag makes release idempotent, so `free()` then `[Symbol.dispose]()` cannot pass a null pointer twice.
- **`wasm-loader.test.mjs` — differential depth.** The randomised run now varies group sizes 1..4 and uses timelocks 1, 127, 128, 32767, 32768 and 65535, so the OP_CHECKSIGADD chain and the 3-byte sign-extended push are both compared against Rust.

### Partly done (2)

- **`check-lazy-entries.js` — facade half of the harness.** The facade import now runs before `node_modules` is exposed, and it does catch a bare static import. It still does not use the two-temp-dir split you asked for, so the raw check and the facade check share one directory.
- **`package.json` — dependency kind.** The migration guide sentence is corrected. The engine still declares `bitcoinjs-lib` and `@bitcoin-js/tiny-secp256k1-asmjs` as hard dependencies, not optional peers, so the second nested copy you describe is still possible.

### Still open (5)

- **`assertWasmPeginSizing.ts` — package provenance.** The SDK derivation is restored to its own package, but the facade now reads its expected values from the engine-co-located guard. That moves the co-location onto the path that has real callers instead of removing it. Neither branch of your either/or is satisfied yet.
- **`prePeginHtlc.ts` — ECC backend.** The per-call mutation is gone. The engine still installs the process-wide backend by itself, on call paths that never touched it before. The assert-do-not-mutate contract is not restored.
- **`rawHtlcConnector.ts` — graph versions.** Still a hardcoded list. It is now named `SUPPORTED_HTLC_GRAPH_VERSIONS` with a comment, but it does not read `supportedTxGraphVersions()` from the generated module.
- **`rawHtlcConnector.ts` — cleanup masks the guard error.** Two of the three release paths are wrapped. The third, added by this PR, is not, so a release failure can still replace the value-mismatch alarm.
- **`CLAUDE.md:64` — overclaim.** The CLAUDE.md line is reworded. The same claim survives in the SECURITY.md counterpart, and the rewrite adds a new one of the same kind.

packages/babylon-tbv-rust-wasm/src/raw.ts
jonybur: Fixed at `97b4385b`. You are right that the guard protected nothing that ships.

`getPrePeginHtlcConnectorInfo` in `src/index.ts` and `src/index-node.ts` now builds `GuardedPrePeginHtlcConnector`. That puts the check on the two production callers: `rebuildDepositTermsCore.ts:193`, which was feeding `connector.scriptPubKey` into `assertEncodedHtlcOutputsMatch` as its own expected value, and `peginInput.ts:101`.

The import has to be dynamic. `rawHtlcConnector.ts:4` imports the generated glue statically, so a static edge from `index.ts` would put it on the Ethereum-only eager graph. `check-lazy-entries.js` proves that: with a static import it fails.

Three tests added: each facade rejects an unsupported version with the guard's own message (`Unsupported HTLC graph version`, which the generated class does not produce), and the node facade still returns byte-identical scripts to the generated class.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
jonybur: Recorded rather than inverted, at `97b4385b`.

Inverting is not available: the SDK peer-depends on the engine, so an engine that imports the SDK is a cycle. CLAUDE.md section 1 and SECURITY.md now state plainly that the derivation ships in the same package as the binary it checks, that one compromised engine publish therefore controls both the value and the expected value, and why that is accepted: no shared code, no shared data, no shared build step, and the differential test stands.

Three sub-claims did not survive checking, so I left the code alone:
- `assertWasmPeginSizing.ts:64`'s "Pure-JS, binary-INDEPENDENT" describes that function's own arithmetic, which is still entirely in the SDK. It is accurate.
- The return annotation is not a regression. At base, `interface ExpectedPrePeginHtlc` had no `export`, so no consumer could name it; `prePeginHtlc.ts` exports it now.
- Version skew is already blocked. The SDK peers the engine at `workspace:*`, which publishes as an exact version, and a missing engine gives `loadTbvWasm`'s actionable message, not a bare TypeError.

packages/babylon-tbv-rust-wasm/src/prePeginHtlc.ts
jonybur: Fixed at `97b4385b`. `initEccLib(ecc)` moved to module scope in `prePeginHtlc.ts`, so the backend is installed once when this lazily-imported module is first evaluated, not inside a derivation.

One correction for the record: because the same `ecc` object is passed every time, the swap was already effectively once per process, not once per HTLC. The reachable cost was a host that installed native `tiny-secp256k1` having it replaced process-wide. That is now a single module-evaluation side effect, which is what the engine needs while it ships its own `bitcoinjs-lib` copy.

Dropping `initEccLib` and re-asserting instead only becomes correct once there is one shared instance. That is the peer-dependency thread, and I would rather do it as its own change than fold a lockfile move into this one.

packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts
jonybur: Fixed at `97b4385b`, taking the first half. The literals are now `SUPPORTED_HTLC_GRAPH_VERSIONS = [1, 2, 3] as const`, with a comment saying the pin is the set the derivation is differentially tested against.

I declined reading `supportedTxGraphVersions()` at runtime. That would silently extend the guard's claimed coverage to a version with no differential test, which is what section 9 forbids. Instead there is a new test asserting the constant equals `supportedTxGraphVersions()`, so an engine bump that adds v4 fails this package's suite rather than failing closed in a depositor's resume.

That matters more now: with the facade rewired in the sibling thread, this pin gates `rebuildDepositTermsCore` and `buildPeginInputPsbt` in production.

packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts
jonybur: Fixed at `97b4385b`. The release is now wrapped so the guard error always wins, with a comment inside the empty catch explaining why the usual no-silent-catch rule is waived there.

One correction, so the team does not record a hazard that cannot happen: the reachability you cite, a second wasm init detaching memory, is impossible here. Everything from `new RawConnector(...)` through the guard calls is synchronous with no `await`, and `initWasm` is async, so nothing can interleave. The `browserRace` test at line 286 frees an object held across an `await`, which is a different shape.

The residual path is narrower but real, and is the one the guard exists for: a doctored or panicking binary can trap inside `getScriptPubKey` and leave the allocator in a state where `free()` traps too.

packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts
jonybur: Fixed at `97b4385b`. `free()` now early-returns on a `#freed` flag, so `free()` followed by `[Symbol.dispose]()` cannot free the null pointer. Verified against the generated glue, which zeroes `__wbg_ptr` with no `ptr === 0` guard.

Two details of the framing are off, though the fix stands. This PR does not add the dispose alias: `dist/generated/vault_wasm.js:1060` already aliases `[Symbol.dispose]` to `free` on the generated class, so the wrapper mirrored existing behaviour. And the reachability you cite conflicts with your own `raw.ts:25` finding: at that head, `index.ts` still freed the generated class, not the wrapper. The wrapper is now strictly safer than the class it replaces, which is worth the six lines.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
jonybur: Fixed at `97b4385b`. The randomised differential now draws keeper and challenger group sizes in 1..4 independently, and walks the script-number encoder boundaries: 1, 127, 128, 32767, 32768 and 65535.

Both gaps were as you describe. Group size never exceeded 2 anywhere in the suite, so the `OP_CHECKSIGADD` chain and its count push were never compared past the second element, and no sample reached 32768, where `bscript.number.encode` switches to a 3-byte sign-extended push.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
jonybur: Fixed at `97b4385b`, by deleting the symlink rather than splitting the harness.

Your premise that the raw half needs `node_modules` does not hold. In both `raw.ts` and `raw-node.ts` the first value import is `./generated/vault_wasm.js`, so Node's depth-1 `ERR_MODULE_NOT_FOUND` for the missing glue wins before `bitcoinjs-lib` is ever resolved. The raw half stays green without it.

I checked the fix does what you asked, rather than trusting the green run: with a static `import { networks } from 'bitcoinjs-lib'` added to `src/index.ts`, `check-lazy-entries.js` now fails with `ERR_MODULE_NOT_FOUND`; with it reverted, it passes. That property was unenforced before this change.

The second symlink, for `browserRacePackage`, stays. That directory keeps `dist/generated` and actually runs `raw.js`.

packages/babylon-tbv-rust-wasm/package.json
jonybur: Half taken at `97b4385b`. The guide sentence was inaccurate as written, so it now says the engine declares those packages directly while the SDK declares `bitcoinjs-lib` and `@bitcoin-js/tiny-secp256k1-asmjs` as optional peers, and that an application installing both can resolve two copies of `bitcoinjs-lib`.

I did not move the engine to peer dependencies in this PR. It changes install semantics for `/raw` consumers, who would then have to install the Bitcoin peer themselves, and it touches the lockfile, so it wants its own change and its own release note rather than riding along here. It is also the precondition for dropping `initEccLib` entirely, which is the other thread. Happy to open it as a follow-up if you want it before the release.

CLAUDE.md
jonybur: Fixed at `97b4385b`, in both files, with your phrasing.

CLAUDE.md now says the connector checks its scripts, control blocks, address and graph version for internal consistency against its own constructor inputs, and that the caller must still check those inputs against on-chain data. The "Only the raw HTLC connector has complete output checks" line is gone.

SECURITY.md is split as you asked: the connector paragraph no longer sits adjacent to the `refund.ts` sentence, and that sentence now names `WasmPrePeginTx` explicitly as one of the unguarded classes, since that is what the SDK actually pulls from `/raw`.

jonybur:
Review fixes are pushed as `97b4385b`.

All ten threads are answered. Seven produced changes, three took part of the ask with the reasoning in the thread. The two that change behaviour: the facade entries now build the guarded connector, so `rebuildDepositTermsCore` stops using a WASM value as its own expected value, and `initEccLib` no longer runs inside a derivation.

The lazy-entry harness is armed again. I proved it rather than trusting the green run: a static `bitcoinjs-lib` import on the facade root now fails `check-lazy-entries.js`, and passes once reverted.

Local checks at this head: 25 engine tests, 1,971 SDK tests in 117 files plus 9 facade checks, 3,809 Vault tests, 34 Chromium tests, build, lint, syncpack, and `docs:clean` with no diff.

jeremy-babylonlabs (commented):
Guard logic reads correct — derivation runs before engine construction, failed construction frees, the double-free latch is in place, and the version is pinned to `supportedTxGraphVersions()`. One gap between the PR body and the tests.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
```diff
@@ -457,3 +472,287 @@ test('createPayoutConnector forwards the network to the address through the brow
...
+        } finally {
+          prototype[getter] = original;
+          prototype.free = free;
+          checked.free();
+        }
+      }
+    });
+  });
+
+  test(`${entry} keeps trusted HTLC inputs after caller mutation and supports dispose`, async () => {
```
jeremy-babylonlabs, line 984: The description says malformed keys and hashlocks, and empty or duplicate key groups, fail before use — but this test only covers version, `timelockRefund: 0`, and the network string. Those paths are unexercised at the raw entry, and this PR can merge on its own.

Add the same negative matrix #2492 has for the payout class (0x-prefixed key, `'ff'.repeat(32)` off-curve, `[]`, case-variant duplicate, 63-char hashlock) against `raw.WasmPrePeginHtlcConnector`, so the HTLC guard is pinned here rather than in the stacked PR.

packages/babylon-ts-sdk/docs/guides/raw-engine-migration.md
```diff
@@ -46,7 +70,7 @@ Both browser and Node.js `/raw` entries export the same four classes and
...
 | `WasmPeginTx.fromJson`; `toJson`                                                                                                            | No guarded round-trip replacement. See the compatibility blocker below.                                                                                                                                  |
-| `WasmPrePeginHtlcConnector` constructor; `getAddress`; `getScriptPubKey`; hashlock and refund script/control-block getters                  | Use the signing builders above. `getPrePeginHtlcConnectorInfo` at the engine root manages object lifetime; it does not independently check signing data.                                                 |
+| `WasmPrePeginHtlcConnector` constructor; `getAddress`; `getScriptPubKey`; hashlock and refund script/control-block getters                  | The raw connector now checks these fields against its inputs. Use the signing builders above for transactions. `getPrePeginHtlcConnectorInfo` at the engine root still has no independent signing-data check.                                                 |
```
gbarkhatov: This row now contradicts itself. Its last sentence ("`getPrePeginHtlcConnectorInfo` at the engine root still has no independent signing-data check") was true at `3c50224`, but `97b4385` made it false: the root facade now builds the same `GuardedPrePeginHtlcConnector` (`index.ts:169-181`), which re-derives every field it returns. CLAUDE.md §1 sends readers to this guide for the guarded alternatives, so it should be accurate. Suggested replacement: say that the root facade uses the same guard, that the guard checks internal consistency against its inputs, and that the caller still has to verify those inputs on-chain.

The CLAUDE.md §1 file label for `rawHtlcConnector.ts` ("raw HTLC output checks") is now incomplete for the same reason. One word fixes it.

packages/babylon-tbv-rust-wasm/src/index.ts
```diff
@@ -161,9 +161,16 @@ export async function buildPeginTxFromPrePegin(
...
+  await getWasmBindings();
+  const { GuardedPrePeginHtlcConnector } = await import(
+    './rawHtlcConnector.js'
+  );
 
-  const connector = new WasmPrePeginHtlcConnector(
+  const connector = new GuardedPrePeginHtlcConnector(
     params.txGraphVersion,
     params.depositorPubkey,
     params.vaultProviderPubkey,
```
gbarkhatov: Low severity, diagnostics only. The constructor hardening from the earlier cleanup-masking comment doesn't reach this path. The constructor checks `getScriptPubKey('bitcoin')`. The network-specific `getAddress(params.network)` / `getScriptPubKey(params.network)` checks run later, inside the `try` below, and that `try` is followed by an unwrapped `finally { connector.free(); }`. If the engine traps, so that the check throws and `free()` then throws too, the `free()` error replaces "WASM HTLC address does not match…", and that message is the one diagnostic that says the engine disagreed with its inputs. It fails closed either way.

Before this commit the facade had no guard errors that could be masked, so this is new. The fix is to wrap the facade's `free()` the way `rawHtlcConnector.ts:58-63` does, or to make `GuardedPrePeginHtlcConnector.free()` swallow a release failure once a check has already thrown. The same applies to `index-node.ts:152-160`.

packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  free(): void {
+    // wasm-bindgen zeroes its pointer without guarding a second call, so a
+    // free() then [Symbol.dispose]() would free the null pointer.
+    if (this.#freed) {
+      return;
+    }
+    this.#freed = true;
+    this.#inner.free();
+  }
```
gbarkhatov, line 162: Neither lifecycle fix has a test that fails if it's reverted:
- Deleting `#freed` leaves the suite green. The only dispose test (`wasm-loader.test.mjs:684`) disposes once, and a second `free()` on the raw object would throw "null pointer passed to rust".
- Removing the `try/catch` around the constructor's `#inner.free()` (lines 57-65) also leaves it green. The tamper test's patched `free` (`wasm-loader.test.mjs:643-646`) counts releases but never throws.

Two small additions would pin them:
1. `checked.free(); assert.doesNotThrow(() => checked[Symbol.dispose]());`
2. In the tamper loop, make `prototype.free` throw, and assert the construction error still matches `/does not match/`.

gbarkhatov (dismissed):
Follow-up to my review of `3c50224`. The only change since then is one commit, `97b4385` (9 files, +153/−25), so this pass covers that commit plus a verdict on each of the 10 earlier comments.

**Headline:** all three earlier merge-blockers are resolved, either fixed or answered with a documented position. The new commit introduces nothing that blocks merge. I've left three small inline comments. The larger risk sits in the stack above this PR (see the end).

|---|---|---|---|
| 1 | Guard wired only into `/raw`; the root facade was unguarded (blocker) | Fixed. Both facades now build `GuardedPrePeginHtlcConnector`, symmetric across browser and node, and each has a test that fails if the guard is reverted | `index.ts:161-195`, `index-node.ts:145-179` |
| 2 | The independent check ships from the same package as the binary it checks (blocker) | Settled. Recorded instead of inverting the dependency, since an engine that imports the SDK would be a cycle. The provenance paragraph is accurate at HEAD | `CLAUDE.md:65`, `SECURITY.md:230-235` |
| 3 | `initEccLib(ecc)` ran on every call (blocker) | Fixed. Moved to module scope. Collapsing onto one shared instance is deferred, with a reason | `prePeginHtlc.ts:13-16` |
| 4 | Graph versions hardcoded | Fixed. Now the named `SUPPORTED_HTLC_GRAPH_VERSIONS`, plus a pin test against `supportedTxGraphVersions()` | `rawHtlcConnector.ts:7-12` |
| 5 | Cleanup `free()` could mask the guard error | Fixed in the constructor. The facade path has the same hazard (inline) | `rawHtlcConnector.ts:57-65` |
| 6 | `free()` then `[Symbol.dispose]()` double-freed | Fixed with `#freed`. There's no test that fails if it's reverted (inline) | `rawHtlcConnector.ts:152-160` |
| 7 | The randomised differential never varied input shape | Fixed. Keeper and challenger counts and timelocks now cycle | `wasm-loader.test.mjs:591-613` |
| 8 | A symlinked `node_modules` disarmed the facade half of the lazy-entry harness | Fixed by removing the symlink. This breaks after #2493 rebases (below) | `check-lazy-entries.js:205-221` |
| 9 | Hard deps vs optional peers; one guide sentence was wrong | Partly fixed. The sentence is corrected. Moving to peers is deferred as its own change | `raw-engine-migration.md:23-27` |
| 10 | "Only the raw HTLC connector has complete output checks" overclaimed | Fixed in both CLAUDE.md and SECURITY.md | `CLAUDE.md:63-64`, `SECURITY.md:218-228` |

**Stack note for #2492 / #2493.** Both children fork at `3c50224` and don't contain `97b4385`, so their rebase can quietly undo fixes here. Nothing is wrong in this PR itself; this is so the rebase is done deliberately.
- `prePeginHtlc.ts`: #2492 deletes the `// BIP-341` block right after where the module-scope `initEccLib(ecc)` now sits, so the rebase should conflict there. The in-function call's removal applies cleanly. Resolving the conflict with #2492's side leaves no `initEccLib` at all.
- CLAUDE.md §1 and SECURITY.md: #2492 reintroduces "The raw HTLC and payout connectors have complete output checks", the overclaim #10 retired. #2493 rewrites the same bullets without the caveat about on-chain inputs. Neither child carries the provenance paragraph that settled #2.
- The lazy-entry harness: at `bf4fad83`, #2493's `raw.ts` no longer imports the generated glue first, and its first value import pulls in `bitcoinjs-lib`. With the symlink gone, the raw half of `check-lazy-entries.js` should hit a bare-package `ERR_MODULE_NOT_FOUND` and fail `build`. Re-adding the symlink would silently disarm the facade half again. The durable fix is the two-temp-dir split from the #8 thread.
- The sibling guards in the children still have the pre-fix patterns: per-call `initEccLib` (#2492 `peginPayout.ts`), an unwrapped constructor-catch `free()`, and a non-idempotent `free()`/dispose (#2492 `rawPayoutConnector.ts`; #2493 `rawPrePeginTx.ts`, `rawPeginTx.ts`).

**Refuted:**
- *"The facade guard runs before the WASM instance is initialized."* The facade awaits `getWasmBindings()` before importing `./rawHtlcConnector.js`, and both loaders import the same generated specifier.
- *"The facade now rejects production inputs it used to accept."* The removed SDK derivation had identical empty, duplicate and timelock rules, and every deposit already passes through `assertWasmPeginSizing`.
- *"The new facade tests would pass with the guard reverted."* The engine's own error text doesn't match `/Unsupported HTLC graph version: 4/`, so both tests go red on revert.

**Verified clean:**
- In `index.ts` and `index-node.ts`, the delta touches only `getPrePeginHtlcConnectorInfo`. The three §4 expander wrappers are byte-unchanged.
- Static edges from the facade to the guard are caught by `eagerGeneratedEdges`.
- No dead code.
- All 11 checks pass at `97b4385`.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
jonybur: Added at `79612752`. `raw rejects malformed HTLC keys, groups, and hashlocks` (and its `raw-node` twin) runs the same five cases against `raw.WasmPrePeginHtlcConnector`, each with its own matcher: a `0x`-prefixed key, an off-curve key, an empty keeper group, a case-variant duplicate, and a 63-character hashlock.

One note on the first case. On this branch the prefixed key is rejected by the engine constructor, so the matcher is `/malformed public key/`. #2492 tightens `normalizeXOnlyKey` to reject the prefix in the derivation itself, which moves that case to `/must be a 32-byte x-only public key/`. I will update the matcher in that merge rather than write a matcher here that matches two different mechanisms.

packages/babylon-ts-sdk/docs/guides/raw-engine-migration.md
jonybur: Fixed at `79612752`. You are right that my own change made that sentence false.

The guide row now says the root facade builds the same guarded connector, that the check is internal consistency against the inputs it was given, and that the caller still has to verify those inputs on-chain.

The CLAUDE.md section 1 label is now "raw and facade HTLC output checks".

packages/babylon-tbv-rust-wasm/src/index.ts
jonybur: Fixed at `79612752`, in both entries. The facade's `finally { connector.free(); }` wraps the release, with the same comment `rawHtlcConnector.ts` carries, so a trapping engine cannot replace `WASM HTLC address does not match the constructor inputs`.

You are also right that this is new: before this commit the facade produced no guard errors that could be masked.

packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts
jonybur: Fixed at `79612752`. Both are pinned now, and I checked each one by reverting the fix rather than trusting the green run.

1. The dispose test does `checked.free()` and then asserts both a second `free()` and a `[Symbol.dispose]()` stay quiet. Deleting the `#freed` latch fails 2 tests.
2. The tamper loop now sets `prototype.free` to throw and asserts the construction error still matches `/does not match/`. Unwrapping the constructor's release fails 2 tests.

With both fixes in place the suite is 27 pass, 0 fail.

jonybur:
Second review round is pushed as `79612752`.

All four threads produced changes. Two of them were defects my own previous commit introduced: the migration-guide row that said the root facade has no independent check, and the facade's unwrapped release that could replace a guard error.

Both lifecycle fixes now have tests that fail when the fix is reverted. Deleting the `#freed` latch fails 2 tests; unwrapping the constructor's release fails 2 tests; with both in place the engine suite is 27 pass, 0 fail.

Local checks at this head: 27 engine tests, 1,971 SDK tests in 117 files plus 9 facade checks, 3,809 Vault tests, 34 Chromium tests, build, lint, syncpack, and `docs:clean` with no diff.

jonybur:
@greptileai review the full PR at current commit `796127526a04218e181d75a0931b2647dcc1f4b2`. Check the diff against its target branch, including the changes since your last review. Verify that the review findings are addressed in the code. Report any remaining issues and the confidence score for this commit.

jonybur:
@greptileai review the full PR at current commit c1a41725cc396f0ab544b77451c6cb920a64ac9d. The review fixes are published and local checks pass. Please check the current code, report any remaining findings, and give the confidence score for this commit.

packages/babylon-tbv-rust-wasm/src/index.ts
```diff
@@ -183,7 +190,12 @@ export async function getPrePeginHtlcConnectorInfo(
       scriptPubKey: connector.getScriptPubKey(params.network),
     };
   } finally {
-    connector.free();
+    try {
+      connector.free();
+    } catch {
+      // A release failure must not replace a guard error: that message is the
+      // one signal that the engine disagreed with its own inputs.
+    }
```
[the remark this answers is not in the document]
jonybur: Fixed in c1a41725cc396f0ab544b77451c6cb920a64ac9d. The constructor and both facades preserve the primary guard message and include both errors in AggregateError. A release-only failure now propagates. Cleanup still runs once. The 27 engine tests, 1,971 SDK tests, nine built-package checks, builds, and lint pass.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/assertWasmPeginSizing.ts
```diff
@@ -63,159 +49,12 @@ type PrePeginHtlcParams = Pick<
...
 }
 
 /**
```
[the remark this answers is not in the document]
jonybur: These two helpers are internal module exports. No SDK index.ts re-exports deriveExpectedPrePeginHtlc or assertUnfundedPrePeginOutputLayout, and package.json exposes no deep PSBT path. TypeDoc reads the explicit public entry points in typedoc.json; those entry points do not export either helper. The current CI job ran docs:clean and reported that generated API docs are up to date: https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34778874314/job/103782081043 . No generated output is missing from this update. Please reassess this finding against the public exports and current CI evidence.

jonybur:
@greptileai please re-review current commit c1a41725cc396f0ab544b77451c6cb920a64ac9d after the response in https://github.com/babylonlabs-io/babylon-toolkit/pull/2481#discussion_r4000767736 . The reported async helpers are not exported by the SDK public entry points. The current CI docs:clean check passes with no generated API changes. Please report any remaining actionable findings and the confidence score for this commit.

gbarkhatov (dismissed):
Follow-up to my review at `97b4385bd8`. This is a fast-forward of two commits: `796127526` pins the guard lifecycle and adds input-rejection tests, and `c1a41725c` reports guard and cleanup failures together.

All 13 prior threads are settled, and nothing new blocks merge.

| # | Prior finding | Verdict at `c1a41725` |
|---|---|---|
| T1 | Guard wired only into `/raw`, not the root facade | **Fixed.** Both facades build `GuardedPrePeginHtlcConnector` (`index.ts:167-179`, `index-node.ts:151-163`) |
| T2 | The check ships in the same package as the binary it checks | **Declined, with reason** (an engine that imports the SDK would be a cycle). Recorded at `CLAUDE.md:65` |
| T3 | `initEccLib(ecc)` ran on every call | **Fixed.** Now runs once at module scope (`prePeginHtlc.ts:13-16`) |
| T4 | Graph versions hardcoded | **Fixed** as an explicit list with a pin test (`wasm-loader.test.mjs:761-772`). Reading them at runtime was **declined, with reason** |
| T5 | Cleanup `free()` masked the guard error (constructor) | **Fixed.** `AggregateError([guard, cleanup])` keeps the guard message (`rawHtlcConnector.ts:57-67`) |
| T6 | `free()` then `[Symbol.dispose]()` double-freed | **Fixed.** A `#freed` flag, pinned at test `:718-720` |
| T7 | Randomised differential never varied input shape | **Fixed.** Group sizes 1..4 and boundary timelocks (`:591-613`) |
| T8 | Symlinked `node_modules` disarmed the facade harness | **Fixed** (`check-lazy-entries.js:205-221`) |
| T9 | Hard deps where the SDK uses optional peers | **Partially fixed.** The guide wording is corrected; the move to peers is **deferred** to its own change |
| T10 | "Only the raw HTLC connector has complete output checks" overclaimed | **Fixed** (`CLAUDE.md:63-64`, `SECURITY.md:219-226`) |
| T11 | Guide row contradicted itself; CLAUDE.md label incomplete | **Fixed** |
| T12 | Facade `finally { free() }` could mask the guard error | **Fixed** with the same `AggregateError` shape (`index.ts:191-203`, `index-node.ts:175-187`) |
| T13 | Neither lifecycle fix had a test that fails if reverted | **Fixed** (`:667-682`, `:718-720`) |

**Correcting my own earlier suggestion (T5/T12).** I suggested letting the guard error always win by swallowing the release failure. At `796127526` that became an empty catch, which CLAUDE.md forbids. `c1a41725c` does it properly: on a guard failure both errors are reported under the guard's message, and on success a release failure is thrown on its own (`index.ts:202`). No SDK or vault code matches on the error type or the guard's message text, so callers are unaffected.

**Checked, and fine:**
- Each engine object is released exactly once on every path: constructor failure, facade failure and facade success.
- Both facades are identical apart from JSDoc, and both run all six getters on the requested network.

**Optional nit.** The tamper-loop test at `wasm-loader.test.mjs:635-690` now also asserts facade rejection and the `AggregateError` shape from both entries. Its name mentions neither, so a split or rename would match the one-behaviour-per-test rule.

jonybur:
@greptileai review

Please review the full diff against `main` at commit `fe24681d0fac7b0a883405f007b83df682cf7c66`. It includes the HTLC and payout connector guards. Report any remaining findings and the confidence score for this commit.

jrwbabylonlab (changes requested):
Need discussion as this is a big and risky change

packages/babylon-tbv-rust-wasm/src/peginPayout.ts
```diff
@@ -0,0 +1,81 @@
...
+  normalizeXOnlyKey,
+  normalizeKeyGroup,
+  nOfNChunks,
+} from './connectorScripts.js';
+import type { PayoutConnectorParams } from './types.js';
+
+// Set the global ECC backend once when this module loads.
+initEccLib(ecc);
+
+const SUPPORTED_PAYOUT_GRAPH_VERSIONS: readonly number[] = [1, 2, 3];
```
gbarkhatov, line 22: The T4 fix wasn't carried over to this sibling.

`SUPPORTED_HTLC_GRAPH_VERSIONS` is an exported `as const` that `wasm-loader.test.mjs:1039-1047` pins to `supportedTxGraphVersions()`. This list is an unexported `readonly number[]`, and nothing pins it. The payout tests loop over a literal `[1, 2, 3]` (`:540`, `:809`), so they still pass after an engine bump that adds v4.

On such a bump, the HTLC pin test forces someone to look at `rawHtlcConnector.ts`, but this list is easy to miss. `createPayoutConnector`, `buildPayoutPsbt` and `buildPeginTxFromFundedPrePegin` would then fail closed for every v4 vault. That is an availability problem, not a correctness one.

Suggest exporting it `as const` and adding it to the existing pin test next to the HTLC list.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
```diff
@@ -457,3 +475,575 @@ test('createPayoutConnector forwards the network to the address through the brow
...
+      const timelocks = [1, 16, 17, 127, 128, 255, 256, 32767, 32768, 65535];
+      for (let sample = 0; sample < timelocks.length; sample += 1) {
+        const keys = Array.from({ length: 36 }, () =>
+          Buffer.from(
+            secp256k1
+              .getPublicKey(secp256k1.utils.randomPrivateKey(), true)
+              .subarray(1),
+          ).toString('hex'),
+        );
+        const count = [1, 2, 16, 17][sample % 4];
```
gbarkhatov, line 611: The T7 fix wasn't carried over here. A single `count` sizes both `vaultKeepers` (`:616`) and `universalChallengers` (`:617`), so every sample has equal group sizes, and the golden fixture is also 1 and 1.

A later edit that computes one group's count push from the *other* group's length would produce identical bytes in every sample and pass. The HTLC differential avoids this by drawing `keeperCount` and `challengerCount` independently (`:873-874`). That matters more now that CLAUDE.md §1 places the §9 differential duty on `peginPayout.ts`.

Suggest drawing the two counts independently from `[1, 2, 16, 17]`, and keeping the 16/17 boundary, which is a good addition.

gbarkhatov (approved):
Follow-up review. Changes since my last review (`c1a41725` → `fe24681d`).

**Headline:** all 13 prior threads are settled. Since the squash and rebase, the new material is mostly the payout connector guard: `connectorScripts.ts`, `peginPayout.ts`, `rawPayoutConnector.ts`, `payoutConnector.ts`, the vault-output bind in `pegin.ts`, and tests. CI is green on all checks, including `verify`. No merge-blockers from me.

Two inline comments cover low-severity test gaps in the payout guard. Both repeat patterns that T4 and T7 already fixed on the HTLC side.

The open changes-request on this PR asks for a discussion and makes no specific technical claim, so there is nothing in it for me to confirm or refute.

### Status of prior threads

| # | Finding | Status at `fe24681d` |
|---|---|---|
| T1 | Guard wired only into `/raw`, not the root facade | Fixed |
| T2 | The independent check ships in the same package as the binary it checks | Declined: the SDK peer-depends on the engine, so inverting would be a cycle. The delta applies the same accepted exception to `peginPayout.ts`. |
| T3 | `initEccLib` ran on every call | Fixed. `peginPayout.ts` follows the same module-scope pattern. |
| T4 | HTLC graph versions hardcoded | Fixed with a pin test; the runtime-read half was declined. The payout sibling has no pin; see the inline comment. |
| T5 | Constructor cleanup `free()` could mask the guard error | Fixed (`AggregateError`) |
| T6 | `free()` then `[Symbol.dispose]()` double-freed | Fixed (`#freed` latch) |
| T7 | The randomized differential never varied input shape | Fixed for the HTLC connector. Not carried over to the payout differential; see the inline comment. |
| T8 | Symlinked `node_modules` disarmed the facade harness | Fixed |
| T9 | Hard deps where the SDK uses optional peers | Partly fixed: the guide sentence is corrected, and the peer move was deferred to its own change |
| T10 | "Only the raw HTLC connector has complete output checks" overclaimed | Fixed |
| T11 | Guide row contradicted itself; CLAUDE.md label incomplete | Fixed |
| T12 | Facade `finally { free() }` could mask the guard error | Fixed, and applied to both new payout factories |
| T13 | Neither lifecycle fix had a test that fails on revert | Fixed, with payout equivalents |

### Where the checks removed from `assertWasmPeginSizing.ts` went

- **Moved into the engine package:**
  - The key-shape, key-group, multisig-chunk, hashlock, timelock, script, taptree and control-block checks now live in `connectorScripts.ts` and `prePeginHtlc.ts`.
  - That move was already in the version I reviewed. It is the T2 exception, which you declined.
- **Compared on the SDK side:**
  - the `assertWasmPeginSizing` loop;
  - `peginInput.ts`;
  - `refund.ts`;
  - the new vault-output bind in `pegin.ts`, which compares against the derivation rather than WASM metadata against WASM bytes. That is stronger than before.

### Follow-up, outside this diff

This is §3. `assemblePayoutPsbt` in `payout.ts` puts the WASM payout leaf script and control block into input 0's `tapLeafScript`. In `payout.ts` I found no check binding them to `peginPrevOut.script`; that check exists only inside the engine-side guard, against the requested keys. Input 1 (Assert) does get an SDK-side bind through `computeTaprootScriptPubKey` (`payout.ts:412-417`). Applying the same check to input 0 would make it SDK-side too. This predates the PR, and the delta improves it, so it fits a follow-up rather than this PR.

### Stack note

#2493 is stacked on this branch but has not been rebased onto `fe24681d`: it is 13 commits behind. When it is rebased, check that this round's fixes survive.

jonybur:
Closing in favor of #2516. It rolls back the raw engine entry instead of guarding it. The decision is recorded in #2361.

PR #2482, opened 2026-09-09
feat(vault): prompt for Bitcoin before wallet actions

### LINKED ISSUE #2505: P2: Require a payout action after Bitcoin reconnects during a wait
## Outcome

Require a separate payout action when Bitcoin reconnects before payout readiness arrives.

Status: the fix is published in #2482 at `8efd5f6e` and included in #2501 at `52de1111`. #2482 passes CI and has a fresh Greptile 5/5. Required human reviews and merge remain.

The regression tests use the real continuation parent, gate, and resume component. They cover delayed polling, reconnect, restored consent, delayed keys, and prior cancellation. Old production fails six checks; the fixed full Vault suite passes.

## Problem

In #2501 at `3c4fcb76`, payout readiness can arrive after Bitcoin reconnects. The first gate mount then starts signing without a separate payout action. Wallet approval is still required. See [gate initialization](https://github.com/babylonlabs-io/babylon-toolkit/blob/3c4fcb7685d4f726a8b81d0d1151bb8123898717/services/vault/src/components/Wallet/BtcActionGate.tsx#L13).

## Scope

Keep the user's action choice while the continuation waits for payout readiness. Cover this delayed first-mount case.

## Complete when

- [ ] Connection alone does not start payout signing in the delayed first-mount case.
- [ ] A separate explicit payout action starts signing after the wallet is ready.
- [ ] Tests cover delayed readiness, delayed key loading, and prior cancellation through the parent continuation.

## Dependencies

Related epic: #2228. Implementation follow-up to #2482 and #2501. The other consent and cancellation requirements remain.

jonybur:
## Outcome

Vault asks for Bitcoin before a Bitcoin action. An already connected user can resume after the first key read. Wallet loss stops active payout work. Reconnect requires a separate Retry.

Status: The branch includes the latest main changes. Commit `312622fd6` uses the approved security key. Fresh CI and Greptile review are pending for this commit. All review threads are resolved. Jeremy's earlier change request is dismissed. A fresh code-owner approval is still required.

## Change

- Preserve the opening action while the first key read and status polling finish.
- Revoke that action after wallet, consent, or key loss. Delayed payout, WOTS, and broadcast status cannot start work after reconnect.
- Stop active payout polling and signing when the required session changes. Cancel the original Ledger provider and reject late results.
- Keep Cancel visible until the active call settles. Then show a recoverable error with Retry when the current key is available. If the parent closes the payout view, cancel the original device request.
- Show a loader and Cancel while the connected wallet key is unavailable. Use the shared deposit content width.
- Remove duplicate test properties introduced by the main branch update.

## Safety

Cancel leaves a pending action unstarted. Connection alone does not start it. The payout-address and signing checks remain in place. This PR does not enable Ethereum-first mode.

The cached-token rejection case remains deferred in #2507 under #2228. Required human reviews and applicable Bitcoin safety checks remain.

## Verification

- Before the main update: 3,814 Vault tests pass; six existing tests are skipped.
- After the main update and follow-up fixes: 138 targeted tests pass. These cover payout cancellation, wallet prompts, continuation, and page layout.
- Production build and full lint pass. The changed files have no lint errors or warnings.
- Regression coverage includes the real first key read, delayed status, reconnect, consent loss, key loss, early asynchronous checks, Ledger cancellation, late signing results, and Retry after immediate cancellation.
- Real-wallet release evidence remains in #2355 and #2233.

## Links

Closes #2505. Tracks #2232. #2232 stays open.

Epic: #2228. Wallet dependency: #2230. Final integration: #2501.

### DIFF
--- a/services/vault/src/components/Wallet/BtcActionGate.tsx
+++ b/services/vault/src/components/Wallet/BtcActionGate.tsx
@@ -1,4 +1,4 @@
-import { Button, Heading } from "@babylonlabs-io/core-ui";
+import { Button, Heading, Loader } from "@babylonlabs-io/core-ui";
 import { useState, type PropsWithChildren } from "react";
 
 import { COPY } from "@/copy";
@@ -8,17 +8,45 @@ export function BtcActionGate({
   children,
   onClose,
 }: PropsWithChildren<{ onClose: () => void }>) {
-  const { connected, requireBtcWallet } = useBtcAction();
+  const {
+    connected,
+    btcConnected,
+    sessionConfirmed,
+    loading,
+    requireBtcWallet,
+  } = useBtcAction();
   const [started, setStarted] = useState(connected);
 
   // Keep an active flow mounted during a temporary wallet disconnect.
   if (started) return <>{children}</>;
 
+  // A saved session is still restoring, so neither prompt is right yet. The
+  // children stay unmounted: every resume branch auto-runs on mount.
+  if (loading) {
+    return (
+      <div className="mx-auto flex max-w-[564px] flex-col items-center gap-4">
+        <Loader />
+        <p className="text-accent-secondary">
+          {COPY.wallet.btcAction.resolving}
+        </p>
+      </div>
+    );
+  }
+
   return (
     <div className="mx-auto flex max-w-[564px] flex-col gap-4">
       <Heading variant="h5">{COPY.wallet.btcAction.heading}</Heading>
-      <p className="text-accent-secondary">{COPY.wallet.btcAction.body}</p>
+      <p className="text-accent-secondary">
+        {btcConnected && !sessionConfirmed
+          ? COPY.wallet.btcAction.confirmBody
+          : COPY.wallet.btcAction.body}
+      </p>
+      {/* These data-testids exist so the real-wallet E2E
+          (e2e/real/actions/stepMachine.ts) can disambiguate these controls
+          from the progress view's Retry, which shares this gate's "Retry"
+          label. Carry them over if you move or rename the controls. */}
       <Button
+        data-testid={connected ? "btc-action-retry" : "btc-action-connect"}
         onClick={() => {
           if (requireBtcWallet()) setStarted(true);
         }}
@@ -27,7 +55,11 @@ export function BtcActionGate({
           ? COPY.wallet.btcAction.retry
           : COPY.wallet.btcAction.connect}
       </Button>
-      <Button variant="outlined" onClick={onClose}>
+      <Button
+        data-testid="btc-action-cancel"
+        variant="outlined"
+        onClick={onClose}
+      >
         {COPY.wallet.btcAction.cancel}
       </Button>
     </div>
--- a/services/vault/src/components/Wallet/BtcActionGate.tsx
+++ b/services/vault/src/components/Wallet/BtcActionGate.tsx
@@ -1,0 +1,35 @@
+import { Button, Heading } from "@babylonlabs-io/core-ui";
+import { useState, type PropsWithChildren } from "react";
+
+import { COPY } from "@/copy";
+import { useBtcAction } from "@/hooks/useBtcAction";
+
+export function BtcActionGate({
+  children,
+  onClose,
+}: PropsWithChildren<{ onClose: () => void }>) {
+  const { connected, requireBtcWallet } = useBtcAction();
+  const [started, setStarted] = useState(connected);
+
+  // Keep an active flow mounted during a temporary wallet disconnect.
+  if (started) return <>{children}</>;
+
+  return (
+    <div className="mx-auto flex max-w-[564px] flex-col gap-4">
+      <Heading variant="h5">{COPY.wallet.btcAction.heading}</Heading>
+      <p className="text-accent-secondary">{COPY.wallet.btcAction.body}</p>
+      <Button
+        onClick={() => {
+          if (requireBtcWallet()) setStarted(true);
+        }}
+      >
+        {connected
+          ? COPY.wallet.btcAction.retry
+          : COPY.wallet.btcAction.connect}
+      </Button>
+      <Button variant="outlined" onClick={onClose}>
+        {COPY.wallet.btcAction.cancel}
+      </Button>
+    </div>
+  );
+}
--- a/services/vault/src/components/deposit/EmergencyWithdrawModal/index.tsx
+++ b/services/vault/src/components/deposit/EmergencyWithdrawModal/index.tsx
@@ -26,6 +26,7 @@ import { V3ModalShell } from "@/components/shared/V3ModalShell";
 import { useETHWallet } from "@/context/wallet";
 import { COPY } from "@/copy";
 import { useActivationState } from "@/hooks/deposit/useActivationState";
+import { useBtcAction } from "@/hooks/useBtcAction";
 import { useEnsureVaultApplicationActive } from "@/hooks/useVaultApplicationActive";
 import {
   captureFunnelFailure,
@@ -53,6 +54,7 @@ export function EmergencyWithdrawModal({
   onClose,
   onSuccess,
 }: EmergencyWithdrawModalProps) {
+  const { requireBtcWallet } = useBtcAction();
   const btcConnector = useChainConnector("BTC");
   const btcWalletProvider =
     (btcConnector?.connectedWallet?.provider as BitcoinWallet | undefined) ??
@@ -92,6 +94,10 @@ export function EmergencyWithdrawModal({
 
   const handleConfirm = useCallback(async () => {
     if (withdrawing) return;
+    if (!requireBtcWallet()) {
+      setLocalError(COPY.wallet.btcAction.body);
+      return;
+    }
     if (!btcWalletProvider || !connectedBtcAddress) {
       setLocalError(
         COPY.deposit.emergencyWithdraw.errors.btcWalletNotConnected,
@@ -149,6 +155,7 @@ export function EmergencyWithdrawModal({
       if (mountedRef.current) setDeriving(false);
     }
   }, [
+    requireBtcWallet,
     withdrawing,
     activity,
     btcWalletProvider,
--- a/services/vault/src/components/deposit/EmergencyWithdrawModal/index.tsx
+++ b/services/vault/src/components/deposit/EmergencyWithdrawModal/index.tsx
@@ -94,16 +94,6 @@ export function EmergencyWithdrawModal({
 
   const handleConfirm = useCallback(async () => {
     if (withdrawing) return;
-    if (!requireBtcWallet()) {
-      setLocalError(COPY.wallet.btcAction.body);
-      return;
-    }
-    if (!btcWalletProvider || !connectedBtcAddress) {
-      setLocalError(
-        COPY.deposit.emergencyWithdraw.errors.btcWalletNotConnected,
-      );
-      return;
-    }
     if (!depositorEthAddress) {
       setLocalError(
         COPY.deposit.emergencyWithdraw.errors.ethWalletNotConnected,
@@ -126,6 +116,16 @@ export function EmergencyWithdrawModal({
       // confirm screen reads, so that gate re-renders with the explanation and
       // the button disabled. Setting `localError` would print it twice.
       if ((await ensureApplicationActive(activity.id)) === false) return;
+      if (!requireBtcWallet()) {
+        setLocalError(COPY.wallet.btcAction.body);
+        return;
+      }
+      if (!btcWalletProvider || !connectedBtcAddress) {
+        setLocalError(
+          COPY.deposit.emergencyWithdraw.errors.btcWalletNotConnected,
+        );
+        return;
+      }
 
       const secretHex = await deriveHtlcSecretHex({
         activity,
--- a/services/vault/src/components/deposit/EmergencyWithdrawModal/index.tsx
+++ b/services/vault/src/components/deposit/EmergencyWithdrawModal/index.tsx
@@ -117,7 +117,7 @@ export function EmergencyWithdrawModal({
       // the button disabled. Setting `localError` would print it twice.
       if ((await ensureApplicationActive(activity.id)) === false) return;
       if (!requireBtcWallet()) {
-        setLocalError(COPY.wallet.btcAction.body);
+        setLocalError(COPY.wallet.btcAction.error);
         return;
       }
       if (!btcWalletProvider || !connectedBtcAddress) {
--- a/services/vault/src/components/pages/__tests__/RootLayout.test.tsx
+++ b/services/vault/src/components/pages/__tests__/RootLayout.test.tsx
@@ -83,6 +83,11 @@ vi.mock("@/components/Wallet", () => ({
 // `Network` is the only export RootLayout's real tree needs here because
 // NetworkBadge imports it directly.
 vi.mock("@babylonlabs-io/wallet-connector", () => ({
+  useBTCWallet: () => ({ connected: walletMock.btcConnected }),
+  useWalletConnect: () => ({
+    connected: walletMock.btcConnected && walletMock.ethConnected,
+    open: vi.fn(),
+  }),
   Network: { MAINNET: "mainnet", SIGNET: "signet" },
 }));
 
--- a/services/vault/src/components/simple/PostDepositContinuationView.tsx
+++ b/services/vault/src/components/simple/PostDepositContinuationView.tsx
@@ -151,9 +151,15 @@ export function PostDepositContinuationView({
     onClose();
   }, [navigate, onClose]);
 
+  // Actionability keys on the vault's on-chain depositor key, not the wallet
+  // key, so the payout branch mounts while Bitcoin is disconnected and its
+  // BtcActionGate can prompt for the wallet instead of a wait state hiding it.
   const isActionable = (id: string): boolean => {
-    const state = getPollingResult(id)?.peginState;
-    return isCandidateVault(state) && hasActionableStep(state, btcPublicKey);
+    const result = getPollingResult(id);
+    return (
+      isCandidateVault(result?.peginState) &&
+      hasActionableStep(result?.peginState, result?.depositorBtcPubkey)
+    );
   };
 
   // Which vault drives the rendered action branch. Two rules:
@@ -406,9 +412,11 @@ export function PostDepositContinuationView({
     );
   }
 
+  // Same on-chain key as `isActionable`: a payout-ready vault with an unknown
+  // depositor key is not selected, and must not mount from the wait fallback.
   if (
     activity &&
-    btcPublicKey &&
+    pollingResult?.depositorBtcPubkey &&
     actions.includes(PeginAction.SIGN_PAYOUT_TRANSACTIONS)
   ) {
     return (
--- a/services/vault/src/components/simple/ResumeDepositContent.tsx
+++ b/services/vault/src/components/simple/ResumeDepositContent.tsx
@@ -87,7 +87,12 @@ interface CaughtError {
 
 export interface ResumeSignContentProps {
   activity: VaultActivity;
-  btcPublicKey: string;
+  /**
+   * The connected wallet's key. `undefined` while Bitcoin is disconnected or
+   * the key is still being read after a connect; the branch mounts either
+   * way so BtcActionGate can prompt, and signing waits for the key.
+   */
+  btcPublicKey: string | undefined;
   depositorEthAddress: Hex;
   /**
    * Every vault ID sharing this deposit's Pre-PegIn (the split-pegin
@@ -103,8 +108,53 @@ export interface ResumeSignContentProps {
 export function ResumeSignContent(props: ResumeSignContentProps) {
   return (
     <BtcActionGate onClose={props.onClose}>
-      <ResumeSignContentConnected {...props} />
+      {props.btcPublicKey === undefined ? (
+        <ResumeSignContentAwaitingKey
+          activity={props.activity}
+          siblingVaultIds={props.siblingVaultIds}
+          onClose={props.onClose}
+        />
+      ) : (
+        <ResumeSignContentConnected
+          {...props}
+          btcPublicKey={props.btcPublicKey}
+        />
+      )}
     </BtcActionGate>
+  );
+}
+
+// The wait state the continuation view used to render for a payout-ready
+// vault without a wallet key: the deposit rests on the auth-anchor step (see
+// getPeginDisplayStep) until the key read settles and the ceremony can start.
+function ResumeSignContentAwaitingKey({
+  activity,
+  siblingVaultIds,
+  onClose,
+}: Pick<ResumeSignContentProps, "activity" | "siblingVaultIds" | "onClose">) {
+  const { vaultCount, currentVaultIndex, perVaultSteps } =
+    useSplitVaultProgress(
+      siblingVaultIds,
+      activity.id,
+      DepositFlowStep.SIGN_AUTH_ANCHOR,
+    );
+
+  return (
+    <DepositProgressView
+      currentStep={DepositFlowStep.SIGN_AUTH_ANCHOR}
+      offchainParamsVersion={activity.offchainParamsVersion}
+      error={null}
+      isComplete={false}
+      isProcessing
+      canClose
+      canContinueInBackground
+      payoutSigningProgress={null}
+      peginSigningProgress={null}
+      vaultCount={vaultCount}
+      currentVaultIndex={currentVaultIndex}
+      perVaultSteps={perVaultSteps}
+      onClose={onClose}
+    />
   );
 }
 
@@ -115,7 +165,7 @@ function ResumeSignContentConnected({
   siblingVaultIds,
   onClose,
   onSuccess,
-}: ResumeSignContentProps) {
+}: ResumeSignContentProps & { btcPublicKey: string }) {
   const {
     signing,
     progress,
--- a/services/vault/src/components/simple/ResumeDepositContent.tsx
+++ b/services/vault/src/components/simple/ResumeDepositContent.tsx
@@ -28,6 +28,7 @@ import { useCallback, useEffect, useRef, useState } from "react";
 import type { Address, Hex } from "viem";
 
 import { getVaultRegistryReader } from "@/clients/eth-contract/sdk-readers";
+import { BtcActionGate } from "@/components/Wallet/BtcActionGate";
 import { computeDepositDerivedState } from "@/components/deposit/DepositSignModal/depositStepHelpers";
 import { usePayoutSigningState } from "@/components/deposit/PayoutSignModal/usePayoutSigningState";
 import { useDepositPollingResult } from "@/context/deposit/PeginPollingContext";
@@ -47,6 +48,7 @@ import { useBroadcastState } from "@/hooks/deposit/useBroadcastState";
 import { useReleaseVpTokenOnUnmount } from "@/hooks/deposit/useReleaseVpTokenOnUnmount";
 import { useRequiredPrePeginDepth } from "@/hooks/deposit/useRequiredPrePeginDepth";
 import { useSplitVaultProgress } from "@/hooks/deposit/useSplitVaultProgress";
+import { useBtcAction } from "@/hooks/useBtcAction";
 import { useRunOnce } from "@/hooks/useRunOnce";
 import { logger } from "@/infrastructure";
 import {
@@ -98,7 +100,15 @@ export interface ResumeSignContentProps {
   onSuccess: () => void;
 }
 
-export function ResumeSignContent({
+export function ResumeSignContent(props: ResumeSignContentProps) {
+  return (
+    <BtcActionGate onClose={props.onClose}>
+      <ResumeSignContentConnected {...props} />
+    </BtcActionGate>
+  );
+}
+
+function ResumeSignContentConnected({
   activity,
   btcPublicKey,
   depositorEthAddress,
@@ -252,7 +262,15 @@ export interface ResumeBroadcastContentProps {
   onSuccess: () => void;
 }
 
-export function ResumeBroadcastContent({
+export function ResumeBroadcastContent(props: ResumeBroadcastContentProps) {
+  return (
+    <BtcActionGate onClose={props.onClose}>
+      <ResumeBroadcastContentConnected {...props} />
+    </BtcActionGate>
+  );
+}
+
+function ResumeBroadcastContentConnected({
   activity,
   batchVaultIds,
   depositorEthAddress,
@@ -335,12 +353,21 @@ export interface ResumeWotsContentProps {
   onSuccess: () => void;
 }
 
-export function ResumeWotsContent({
+export function ResumeWotsContent(props: ResumeWotsContentProps) {
+  return (
+    <BtcActionGate onClose={props.onClose}>
+      <ResumeWotsContentConnected {...props} />
+    </BtcActionGate>
+  );
+}
+
+function ResumeWotsContentConnected({
   activity,
   siblingVaultIds,
   onClose,
   onSuccess,
 }: ResumeWotsContentProps) {
+  const { requireBtcWallet } = useBtcAction();
   const btcConnector = useChainConnector("BTC");
   const btcWalletProvider =
     (btcConnector?.connectedWallet?.provider as BitcoinWallet | undefined) ??
@@ -394,7 +421,7 @@ export function ResumeWotsContent({
   const trackPrimedTxid = useReleaseVpTokenOnUnmount();
 
   const handleSubmit = useCallback(async () => {
-    if (!btcWalletProvider || !connectedBtcAddress) {
+    if (!requireBtcWallet() || !btcWalletProvider || !connectedBtcAddress) {
       setError({ raw: COPY.deposit.resume.walletNotConnected });
       setLoading(false);
       return;
@@ -553,6 +580,7 @@ export function ResumeWotsContent({
     }
   }, [
     activity,
+    requireBtcWallet,
     btcWalletProvider,
     connectedBtcAddress,
     btcConnector?.connectedWallet?.id,
@@ -678,13 +706,22 @@ export interface ResumeActivationContentProps {
   onGoToDashboard: () => void;
 }
 
-export function ResumeActivationContent({
+export function ResumeActivationContent(props: ResumeActivationContentProps) {
+  return (
+    <BtcActionGate onClose={props.onClose}>
+      <ResumeActivationContentConnected {...props} />
+    </BtcActionGate>
+  );
+}
+
+function ResumeActivationContentConnected({
   activity,
   depositorEthAddress,
   siblingVaultIds,
   onClose,
   onGoToDashboard,
 }: ResumeActivationContentProps) {
+  const { requireBtcWallet } = useBtcAction();
   const btcConnector = useChainConnector("BTC");
   const btcWalletProvider =
     (btcConnector?.connectedWallet?.provider as BitcoinWallet | undefined) ??
@@ -720,7 +757,7 @@ export function ResumeActivationContent({
   });
 
   const handleSubmit = useCallback(async () => {
-    if (!btcWalletProvider || !connectedBtcAddress) {
+    if (!requireBtcWallet() || !btcWalletProvider || !connectedBtcAddress) {
       setLocalError({
         raw: COPY.deposit.resume.walletNotConnected,
       });
@@ -763,6 +800,7 @@ export function ResumeActivationContent({
     }
   }, [
     activity,
+    requireBtcWallet,
     btcWalletProvider,
     connectedBtcAddress,
     btcConnector?.connectedWallet?.id,
--- a/services/vault/src/components/simple/__tests__/DepositSignContent.test.tsx
+++ b/services/vault/src/components/simple/__tests__/DepositSignContent.test.tsx
@@ -9,6 +9,17 @@ import type { DepositErrorContent } from "@/utils/errors";
 import type { DepositProgressViewProps } from "../DepositProgressView";
 import { DepositSignContent } from "../DepositSignContent";
 
+const btcActionWallet = vi.hoisted(() => ({ connected: true, open: vi.fn() }));
+vi.mock("@babylonlabs-io/wallet-connector", () => ({
+  useBTCWallet: () => ({ connected: btcActionWallet.connected }),
+  useWalletConnect: () => ({ connected: true, open: btcActionWallet.open }),
+}));
+
+beforeEach(() => {
+  btcActionWallet.connected = true;
+  btcActionWallet.open.mockClear();
+});
+
 const mockExecuteDeposit = vi.hoisted(() => vi.fn());
 const mockCancelDeviceSign = vi.hoisted(() => vi.fn());
 // Mutable so tests can drive the flow's cancel seam through the static mock.
@@ -146,6 +157,22 @@ describe("DepositSignContent", () => {
     deviceCancelState.requested = false;
     flowErrorState.error = null;
     flowErrorState.resumableVaultIds = null;
+  });
+
+  it("keeps a deposit unstarted until the user retries after connection", async () => {
+    btcActionWallet.connected = false;
+    mockExecuteDeposit.mockResolvedValue(null);
+    const first = renderContent();
+    fireEvent.click(first.getByTestId("summary-sign"));
+    expect(btcActionWallet.open).toHaveBeenCalledWith("BTC");
+    expect(mockExecuteDeposit).not.toHaveBeenCalled();
+    expect(first.queryByTestId("progress")).toBeNull();
+    first.unmount();
+    btcActionWallet.connected = true;
+    const next = renderContent();
+    expect(mockExecuteDeposit).not.toHaveBeenCalled();
+    fireEvent.click(next.getByTestId("summary-sign"));
+    expect(mockExecuteDeposit).toHaveBeenCalledOnce();
   });
 
   it("renders Retry and switches to the continuation view after a post-registration device error", async () => {
--- a/services/vault/src/components/simple/__tests__/DepositSignContent.test.tsx
+++ b/services/vault/src/components/simple/__tests__/DepositSignContent.test.tsx
@@ -1,5 +1,6 @@
 import type { BitcoinWallet } from "@babylonlabs-io/ts-sdk/shared";
 import { act, fireEvent, render } from "@testing-library/react";
+import { cloneElement } from "react";
 import type { Address, Hex } from "viem";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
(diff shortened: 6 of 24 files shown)


jonybur:

Please review the full PR at current commit `5bda6cda66cd1a3a95be90584952510285979dbb`. Check the diff against its target branch. Report any remaining issues and the confidence score for this commit.

services/vault/src/components/deposit/EmergencyWithdrawModal/index.tsx
```diff
@@ -92,6 +94,10 @@ export function EmergencyWithdrawModal({
 
   const handleConfirm = useCallback(async () => {
     if (withdrawing) return;
+    if (!requireBtcWallet()) {
+      setLocalError(COPY.wallet.btcAction.body);
+      return;
+    }
```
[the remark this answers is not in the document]
jonybur: Fixed in `468fa99cca875c2e53afa26ddddc023041021e1d`.

The application status check now runs before the Bitcoin connection guard. An inactive recovery action stops without opening the wallet dialog, deriving a secret, or submitting a transaction.

Two named tests cover connected and disconnected wallets. The disconnected case failed before the fix. All six emergency recovery tests now pass. All 215 tests across the affected files pass, and scoped lint and diff checks pass.

jonybur:

Please review the full PR at current commit `468fa99cca875c2e53afa26ddddc023041021e1d` against `main`. The emergency recovery status check now runs before the Bitcoin prompt. Payout Retry is also restored after a recoverable connection error. Both regression checks failed before the fixes and passed after them. All 215 tests across the affected files pass.

Report any remaining findings and the confidence score for this commit.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -98,7 +100,15 @@ export interface ResumeSignContentProps {
   onSuccess: () => void;
 }
 
-export function ResumeSignContent({
+export function ResumeSignContent(props: ResumeSignContentProps) {
+  return (
+    <BtcActionGate onClose={props.onClose}>
```
gbarkhatov: **Merge-blocker — for payout this gate cannot render when it is needed, and connecting Bitcoin auto-starts the ceremony.**

The only mount site for `ResumeSignContent` is `PostDepositContinuationView.tsx:409-413`, which renders it only when `btcPublicKey` is truthy — and `btcPublicKey` is `useBtcPublicKey(btcConnected)`, so it is `undefined` whenever BTC is disconnected. The vault also drops out of `hasActionableStep` at `:156` in that state. Two consequences: the "Connect your Bitcoin wallet" screen is unreachable for payout, and the instant BTC connects the branch mounts with `connected === true`, so `BtcActionGate` seeds `started` true on its first render, `ResumeSignContentConnected` renders, and `useRunOnce(handleSign, !wasCanceled)` at `:142` fires with no user gesture.

That is what #2232 rules out — "Connection never runs it; require a separate explicit retry" — and audit L70. It also means the PR's "keep active resume flows mounted during a temporary disconnect" cannot hold here: on disconnect `btcPublicKey` goes `undefined` and the whole subtree unmounts, taking the latch with it.

Fix: make `started` a real user-action latch initialised `false`, and give actionability a connection-independent depositor key — the resume path already reads `basic.depositorBtcPubKey` at `:451` — or move the `btcPublicKey &&` condition inside `ResumeSignContentConnected` so the gate is what handles the missing key.

services/vault/src/hooks/useBtcAction.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  useBTCWallet,
+  useWalletConnect,
+} from "@babylonlabs-io/wallet-connector";
+import { useCallback } from "react";
+
+export function useBtcAction() {
+  const { connected: btcConnected } = useBTCWallet();
+  const { connected: sessionConfirmed, open } = useWalletConnect();
+  const connected = btcConnected && sessionConfirmed;
```
gbarkhatov, line 10: **Merge-blocker — this is a whole-session gate wearing a Bitcoin label, and it settles an open decision in passing.**

`sessionConfirmed` is `useWalletConnect().connected`, which `useWalletConnect.ts:70-77` computes as `selected && confirmed`, where `selected` requires *every* required chain — and the vault sets `requiredChains={["BTC", "ETH"]}` at `VaultWalletConnectionProvider.tsx:261`. So a missing or unconfirmed **Ethereum** wallet fails a Bitcoin gate, while `open("BTC")` drops the user on the Bitcoin list and `COPY.wallet.btcAction.body` promises only Bitcoin is needed. The user can loop on Retry without ever being shown the thing that is actually missing.

`RootLayout` now carries both definitions at once: `isWalletConnected = btcConnected && ethConnected` at `:90` still paints the fully connected shell while `openDeposit` at `:151-156` silently re-opens the wallet dialog.

The audit lists this as unresolved at L131 — "Define when ETH-first access requires explicit dialog confirmation. The current menu and page checks differ. Do not copy one rule to the other without this decision." Benign while the vault mandates both chains, but this gate exists for the ETH-first work, where it becomes a live trap.

Fix: return `btcConnected` and `sessionConfirmed` separately, call `open()` with no chain when BTC is already present so it falls through to `displayChains()`, and add a confirmation-pending copy variant — or land the decision first.

services/vault/src/hooks/deposit/useReclaimState.ts
```diff
@@ -72,6 +74,10 @@ export function useReclaimState({
   const handleReclaim = useCallback(
     async (feeRate: number) => {
       if (inFlightRef.current || reclaiming) return;
+      if (!requireBtcWallet()) {
```
gbarkhatov, line 77: This gate is unreachable in exactly the state it exists for. `useReclaimRowAction` decides whether to offer the action via `isOwnedByConnectedWallet(depositorBtcPubkey, publicKeyNoCoord)`, which **fails closed** on an empty connected key — deliberately, per its own comment at `useReclaimRowAction.ts:63-66`. With BTC disconnected `publicKeyNoCoord` is empty, the row action is never offered, and `handleReclaim` is never called, so `requireBtcWallet()` here is dead in the disconnected case.

Refund is unaffected — `useRefundRowAction` carries no BTC-key dependency — which is what makes this a per-flow gap rather than a blanket one. Same fix shape as the payout comment: derive ownership from the on-chain depositor key and let the gate handle the missing wallet.

services/vault/src/components/Wallet/BtcActionGate.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import { COPY } from "@/copy";
+import { useBtcAction } from "@/hooks/useBtcAction";
+
+export function BtcActionGate({
+  children,
+  onClose,
+}: PropsWithChildren<{ onClose: () => void }>) {
+  const { connected, requireBtcWallet } = useBtcAction();
+  const [started, setStarted] = useState(connected);
```
gbarkhatov: `useState(connected)` snapshots once at mount and nothing except a click ever advances it — but `connected` is legitimately `false` during a normal restore. `BTCWalletProvider` starts `loading = true` and only sets provider state after two awaited provider calls (`getAddress`, `getPublicKeyHex`), so a resume view mounted inside that window latches `started` false and shows "Connect your Bitcoin wallet" to a depositor whose wallet is in fact connected.

Recoverable rather than fatal — the button reads "Retry" once `connected` flips and works — so this is a spurious interstitial, not a lock. It bites independently of the payout comment on the broadcast, WOTS and activation flows, where the gate *is* reachable.

`useBTCWallet()` already returns `loading` alongside `connected`: holding the latch open while `loading` is true is enough.

services/vault/src/components/Wallet/BtcActionGate.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <div className="mx-auto flex max-w-[564px] flex-col gap-4">
+      <Heading variant="h5">{COPY.wallet.btcAction.heading}</Heading>
+      <p className="text-accent-secondary">{COPY.wallet.btcAction.body}</p>
+      <Button
+        onClick={() => {
+          if (requireBtcWallet()) setStarted(true);
+        }}
+      >
+        {connected
+          ? COPY.wallet.btcAction.retry
```
gbarkhatov: This label collides with the real-wallet E2E driver. `services/vault/e2e/real/actions/stepMachine.ts:58` defines `RETRY_BUTTON_RX = /^retry$/i` and resolves it with `.first()`; `COPY.wallet.btcAction.retry` is exactly `"Retry"`, so this interstitial now offers a second matching button that the step machine may auto-click, and none of the three new controls carries a `data-testid` to disambiguate.

Per CLAUDE.md's E2E hooks section these selectors are load-bearing test infrastructure and a breakage here surfaces with no compile error. Adding distinct testids to the connect/retry and cancel controls is enough.

services/vault/src/copy.ts
```diff
@@ -1284,6 +1284,13 @@ export const COPY = {
     },
   },
   wallet: {
+    btcAction: {
+      heading: "Connect your Bitcoin wallet",
+      body: "This action needs your Bitcoin wallet. Connect it, then try the action again.",
```
gbarkhatov, line 1331: This string is authored as prompt-panel prose ("…then try the action again") and then reused verbatim as an inline **error** at five call sites: `EmergencyWithdrawModal/index.tsx:120`, `useReclaimState.ts:78`, `useRefundState.ts:90`, and `useArtifactDownload.ts:151` and `:463`. At those sites it renders into a red callout at the same moment `open("BTC")` puts the wallet dialog on screen, so the user is told to retry by an error standing next to the dialog that is already the next step. Meanwhile the same condition surfaces `COPY.deposit.emergencyWithdraw.errors.btcWalletNotConnected` and the structured `COPY.deposit.payoutSigningGuards.walletNotConnected` elsewhere — three messages for one state, which is the drift `copy.ts` exists to prevent.

A separate `COPY.wallet.btcAction.error` (or reusing the existing per-flow strings) would settle it.

While you are in `useRefundState.ts`: `:97`, `:101` and `:109` still inline `"BTC wallet not connected"`, `"Missing BTC Vault ID"` and `"ETH wallet not connected"` a few lines below the new gate. The PR now imports `COPY` into that exact file, and CLAUDE.md asks for inlined strings to be migrated as part of the change that touches them.

services/vault/src/components/pages/__tests__/RootLayout.test.tsx
```diff
@@ -83,6 +83,11 @@ vi.mock("@/components/Wallet", () => ({
 // `Network` is the only export RootLayout's real tree needs here because
 // NetworkBadge imports it directly.
 vi.mock("@babylonlabs-io/wallet-connector", () => ({
+  useBTCWallet: () => ({ connected: walletMock.btcConnected }),
+  useWalletConnect: () => ({
+    connected: walletMock.btcConnected && walletMock.ethConnected,
```
gbarkhatov: This mock encodes a different semantic than production: `useWalletConnect().connected` is `selected && confirmed`, not `btcConnected && ethConnected`. The confirmation dimension is mocked out of existence, so this suite structurally cannot catch the `useBtcAction` conflation — and "both wallets connected, dialog unconfirmed" is the *only* state in which `openDeposit`'s new guard changes behaviour, since the deposit CTA is already disabled while disconnected.

The same collapse appears in `EmergencyWithdrawModal.test.tsx`, `usePayoutSigningState.test.tsx`, `DepositSignContent.test.tsx` and `useVaultActions.test.ts`, all hardcoding `connected: true`. `ResumeDepositContent.test.tsx` already models the two knobs separately — worth copying that shape here, along with a test for `openDeposit` itself, which currently has none.

services/vault/src/components/simple/__tests__/DepositSignContent.test.tsx
```diff
@@ -148,6 +159,22 @@ describe("DepositSignContent", () => {
...
 
+  it("keeps a deposit unstarted until the user retries after connection", async () => {
+    btcActionWallet.connected = false;
+    mockExecuteDeposit.mockResolvedValue(null);
+    const first = renderContent();
+    fireEvent.click(first.getByTestId("summary-sign"));
+    expect(btcActionWallet.open).toHaveBeenCalledWith("BTC");
+    expect(mockExecuteDeposit).not.toHaveBeenCalled();
+    expect(first.queryByTestId("progress")).toBeNull();
+    first.unmount();
```
gbarkhatov: The first half of this test is the real thing — it asserts `open` was called with `"BTC"` and that the deposit did not run. The tail is where it stops biting: after `first.unmount()` the re-render at `:172-173` is a fresh mount, so `expect(mockExecuteDeposit).not.toHaveBeenCalled()` proves only that a newly mounted component does not auto-start, which is already true for an imperative handler with no self-start path. The interesting assertion — that flipping `connected` to true on a *live* component does not start the action — is the one the unmount removes.

`useArtifactDownload.test.tsx:166-186` and `useVaultActions.test.ts:371-390` have the same shape, and `ResumeDepositContent.test.tsx:1136-1138` asserts after `view.unmount()`. Worth contrasting with `ResumeDepositContent.test.tsx:1094-1116`, `:1141-1160` and `:1162-1181`, which get it right: rerender live with `connected` flipped, assert nothing ran, then click Retry.

services/vault/src/hooks/deposit/useArtifactDownload.ts
```diff
@@ -141,6 +143,14 @@ export function useArtifactDownload(options?: {
       // god-mode panel's "Mock artifact download" toggle; off in production
       // builds, where the god-mode gate is compile-time false.
       const demoDownload = getArtifactDownloadOverride();
+      if (
+        !demoDownload &&
+        !vpTokenRegistry.peek(stripHexPrefix(peginTxid)) &&
```
gbarkhatov: Two small things, both cosmetic — the cached-token bypass itself is correct and I verified it (`peek()` truthiness means the auth anchor was already derived, and token re-acquire needs no wallet signature).

`stripHexPrefix(peginTxid)` here duplicates the `normalizedPeginTxid` constant derived at `:183` for the same registry key; hoisting that above the gate stops the two drifting apart later.

Also, this early return sits *above* the `abortController` assignment at `:170`, so unlike every other bail in this function it cannot mark an in-flight flow stale — it sets `loading: false` while an earlier flow parked in the retry sleep keeps running and can later overwrite the error with a success. The window is narrow (it needs a cold registry plus a disconnect), but moving the guard below the controller assignment costs nothing.

services/vault/src/hooks/deposit/useArtifactDownload.ts
```diff
@@ -452,6 +458,11 @@ export function useArtifactDownload(options?: {
           // retry the stream.
           if (!primeAttempted && isAuthRejectedError(err)) {
             primeAttempted = true;
+            vpTokenRegistry.peek(normalizedPeginTxid)?.invalidate();
```
gbarkhatov, line 465: Worth a note that this `invalidate()` changed scope: it used to live inside `tryPrimeAndRetry` *after* its `if (!primeContext) return false` bail, and now fires unconditionally on the first `isAuthRejectedError` — including on the collateral re-download path that `:322-324` documents as deliberately lacking a prime context, and when the new BTC gate below refuses.

I traced it as benign: `invalidate()` nulls `cachedJsonRpc`/`cachedGrpc` only, the registry entry survives (`tokenRegistry.ts` evicts solely via `release()`), so the pre-flight `peek()` bypass still holds, and re-acquire is a plain JSON-RPC call carrying `authAnchorHex` off the provider instance. So no defect — but it is an unremarked behaviour change in a gating PR, and the `primeContext === null` path has no test covering it.

gbarkhatov (dismissed):
Reviewed at `468fa99c` with three independent passes, then deduplicated and verified every surviving finding against `docs/audits/2447-vault-wallet-requirements.md` and the wallet-connector source.

Approving. Comments are suggestions and the calls are yours — two are flagged as merge-blockers because they defeat the PR's own stated contract, not because I want to hold the merge.

**The gate placement is right.** Every Bitcoin-requiring action in the audit has a gate, no Ethereum-only action was over-gated, `handleActivation` is untouched, emergency withdraw resolves application status before asking for Bitcoin, and the guard ordering leaves Retry available everywhere (gates sit before the re-entrancy and loading flags with no intervening `await`). The problems are in what the gate *does*, not where it was put.

**Two blockers.** For payout the gate can never render in the state it exists for: `PostDepositContinuationView.tsx:409-413` renders `ResumeSignContent` only when `btcPublicKey` is truthy, and that comes from `useBtcPublicKey(btcConnected)`. So the prompt is unreachable while BTC is disconnected, and the moment BTC connects the branch mounts with `connected === true`, `started` is true on the first render, and `useRunOnce(handleSign)` fires — which is the one thing #2232 rules out: "Connection never runs it; require a separate explicit retry." Separately, `useBtcAction` gates on `useWalletConnect().connected`, which is `selected && confirmed` across *both* required chains, so a missing or unconfirmed Ethereum wallet fails a Bitcoin gate while the copy and `open("BTC")` both promise that only Bitcoin is needed. The audit records that as an open decision (L131) rather than something to settle in passing.

**Three claims I refuted, recorded so they don't get re-raised.** The `invalidate()` moved to the call site does *not* strand the collateral re-download path: it nulls `cachedJsonRpc`/`cachedGrpc` only, the registry evicts solely via `release()`, so `peek()` stays truthy and the pre-flight bypass keeps working — and re-acquire is a plain JSON-RPC call carrying `authAnchorHex` held on the provider instance, with no wallet interaction. For the same reason the `peek()` bypass is sound rather than a cached-token confusion, and matches "Allow recovery downloads with cached authentication". Third, `useReclaimRowAction`'s key read is not display-only — it fails closed, which is precisely what makes the reclaim comment below real.

The remainder are smaller: an E2E button-name collision, copy reuse across one condition, and a few new tests that cannot fail.

jrwbabylonlab (dismissed):

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -98,7 +100,15 @@ export interface ResumeSignContentProps {
...
+      <ResumeSignContentConnected {...props} />
+    </BtcActionGate>
+  );
+}
```
jeremy-babylonlabs, line 122: `BtcActionGate` cannot actually gate payout resume here. `PostDepositContinuationView` only renders `ResumeSignContent` once `btcPublicKey` is truthy, and that requires BTC to already be connected (`useBtcPublicKey(btcConnected)`). So the gate's `useState(connected)` initialises `started = true` on first mount, the prompt is skipped entirely, and `useRunOnce(handleSign, !wasCanceled)` auto-fires the payout sign — the "connection alone must never start it" case this PR is meant to prevent.

Could you add a regression test that renders `ResumeSignContent` fresh (not via a `cloneElement` rerender) with BTC already connected, and asserts `handleSign` does not fire until an explicit click?

services/vault/src/hooks/useBtcAction.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  useBTCWallet,
+  useWalletConnect,
+} from "@babylonlabs-io/wallet-connector";
+import { useCallback } from "react";
+
+export function useBtcAction() {
+  const { connected: btcConnected } = useBTCWallet();
+  const { connected: sessionConfirmed, open } = useWalletConnect();
```
jeremy-babylonlabs, line 9: Following on from @gbarkhatov's review: `useWalletConnect().connected` reflects `selected && confirmed` across both required chains, not BTC alone — so a BTC-only action can be blocked on Ethereum wallet state while the copy and the `open("BTC")` call both promise only Bitcoin is needed.

Since this hook is now the shared gate for every BTC action site, worth settling explicitly here rather than leaving it as an open audit item.

jeremy-babylonlabs (dismissed):
Gating is applied broadly and the cancel path is covered, but the payout resume path bypasses the guarantee this PR exists to provide.

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: Fixed in 585f7b8f. The payout branch now mounts on the action alone. PostDepositContinuationView keys actionability and the branch guard on the vault's on-chain depositor key (`result.depositorBtcPubkey`), the same input VaultsLifecycleSections passes to `hasActionableStep`, so the branch renders while Bitcoin is disconnected and BtcActionGate prompts instead of a wait state hiding it. ResumeSignContent takes `btcPublicKey: string | undefined`; after Retry it renders the previous wait state until the wallet key read settles, then mounts the signing hook, which keeps its wallet-derived key so the wrong-wallet guards still bite. The `btcPublicKey !== undefined` case in peginStateMachine stays, and the sibling-selection test at PostDepositContinuationView.test.tsx:845 still passes with the on-chain key as its input. One limit: the SIGN_PAYOUT_TRANSACTIONS action comes from the VP poll, and usePeginPollingQuery is disabled without the wallet key. The gate therefore covers a mid-session disconnect, where keepPreviousData keeps the action visible. It does not cover a modal opened on a fresh load with Bitcoin disconnected: there is no action yet, the view waits, and if Bitcoin connects while that wait state is open the branch still mounts with the gate already open and auto-runs. Closing that needs the payout action without the wallet key. Draft #2501 changes the poll to key on the on-chain depositor key, which is where it closes; it is recorded on #2232.

services/vault/src/hooks/useBtcAction.ts
jonybur: Fixed in 585f7b8f. `requireBtcWallet` now calls `open(btcConnected ? undefined : "BTC")`. A missing Bitcoin wallet lands on the Bitcoin wallet list. A connected but unconfirmed session lands on the chain list, which carries the confirm step. The hook returns `btcConnected` and `sessionConfirmed` as separate fields. BtcActionGate shows a confirmation-pending body (`COPY.wallet.btcAction.confirmBody`) in that state. Which chains a Bitcoin action truly requires is left to #2230 and #2232.

services/vault/src/hooks/deposit/useReclaimState.ts
jonybur: No code change here. The reclaim gate is live in the connected-but-unconfirmed state. The row action passes there and `requireBtcWallet` opens the confirm step, so it is not dead code. Making the reclaim prompt reachable while disconnected means deriving ownership from the on-chain depositor key instead of the connected wallet's key. That changes the fail-closed contract of `useReclaimRowAction` and belongs with the payout ownership work on #2232, where it is now recorded.

services/vault/src/components/Wallet/BtcActionGate.tsx
jonybur: Fixed in 585f7b8f. The gate now reads the wallet's `loading` flag through `useBtcAction`. While a saved session restores, it renders a loader with `COPY.wallet.btcAction.resolving` instead of the connect prompt. The children stay unmounted during that window. Every resume branch auto-runs on mount, so seeding the latch from `connected || loading` would start wallet work on a half-restored wallet with no user gesture. Once the restore settles, the prompt shows Retry and one click mounts the flow. A gate-level test now pins the branch (`src/components/Wallet/__tests__/BtcActionGate.test.tsx`): with the wallet `loading`, the gate renders `COPY.wallet.btcAction.resolving`, the children are not mounted and no prompt controls render. Removing the loading branch fails that test.

services/vault/src/components/Wallet/BtcActionGate.tsx
jonybur: Fixed in 585f7b8f. The controls carry `data-testid`: `btc-action-connect` or `btc-action-retry` on the primary button (it follows the label), and `btc-action-cancel` on Cancel. The inline comment says the ids exist so `e2e/real/actions/stepMachine.ts` can disambiguate these controls from the progress view's Retry. stepMachine.ts does not reference them yet. I left the `retryButton` matcher in stepMachine.ts as is. Scoping it to the progress view is a separate E2E change.

services/vault/src/copy.ts
jonybur: Fixed in 585f7b8f. Added `COPY.wallet.btcAction.error` ("Bitcoin wallet not connected.") and used it at the five inline error sites: EmergencyWithdrawModal, useArtifactDownload (two), useReclaimState and useRefundState. `body` is now only the prompt panel's text. The four inline strings in useRefundState moved to `COPY.deposit.refundReview.errors`, next to the other refund review copy, since there is no top-level `refund` key. That also normalises "Missing BTC Vault ID" to "Missing BTCVault ID", matching `reclaim.errors`.

services/vault/src/components/pages/__tests__/RootLayout.test.tsx
jonybur: Fixed in 585f7b8f with a direct unit test instead of a mock rework. `src/hooks/__tests__/useBtcAction.test.tsx` covers the case this suite could not: Bitcoin connected, session unconfirmed, `requireBtcWallet()` returns false and `open` is called with `undefined`. It also pins the disconnected case (`open("BTC")`), the pass-through case and the loading flag. I left RootLayout's mock as is. An `openDeposit` click test also needs the geofencing mock off `isLoading: true` and a replacement for the `SimpleDeposit` stub, which is more than this PR should grow.

services/vault/src/components/simple/__tests__/DepositSignContent.test.tsx
jonybur: Fixed in 585f7b8f. The DepositSignContent test no longer unmounts. It hoists the element (`contentElement()`), flips the wallet flag, rerenders the same tree with `cloneElement`, asserts the deposit is still unstarted, then clicks Sign and asserts it ran once. The rerender is required. The blocked path changes no state, so `requireBtcWallet` stays closed over the stale value without it. The WOTS cancel test in ResumeDepositContent.test.tsx got the same treatment: after cancel it flips the flag, rerenders live, and asserts the prompt shows Retry with no work started.

services/vault/src/hooks/deposit/useArtifactDownload.ts
jonybur: Fixed in 585f7b8f. `normalizedPeginTxid` is hoisted above the guard and the `peek()` call uses it. The guard now calls `abortControllerRef.current?.abort()` before it sets the error, the same as `cancel()`. An older in-flight flow then reads `isStale()` true and settles silently. The guard stays where it is, before `openArtifactSaveTarget`, so the save picker keeps its user activation.

services/vault/src/hooks/deposit/useArtifactDownload.ts
jonybur: Trace confirmed, no behaviour change. `invalidate()` clears only the two cached tokens. The registry entry is evicted solely by `release()`, so `peek()` stays truthy and the pre-flight bypass is unaffected. On the `primeContext === null` path the worst case is one extra token acquire, which needs no wallet. The new contract: the first auth rejection always drops the cached token, then re-primes only when a prime context exists and the Bitcoin gate passes. The explanatory comment that used to sit on this call is restored at the new call site in 585f7b8f.

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: Fixed in 585f7b8f together with the thread at line 105, with one limit. The branch no longer waits for the wallet key: PostDepositContinuationView mounts ResumeSignContent on the vault's on-chain depositor key, so when the payout action is already known and Bitcoin drops mid-session the gate renders closed. After Connect and Retry the branch holds the wait state until the wallet key read settles, then mounts usePayoutSigningState and its auto-run. Still open: the action itself comes from the VP poll, and usePeginPollingQuery is disabled without the wallet key. On a fresh load with Bitcoin disconnected there is no action, the view waits, and if Bitcoin connects while that wait state is open the branch mounts with `started=true` and auto-runs, as you describe. Closing that needs the poll to run without the wallet key. Draft #2501 changes the poll to key on the on-chain depositor key, which is where it closes; it is recorded on #2232. On the test: a fresh render with Bitcoin already connected latches the gate open by design (opening the modal is the gesture), so `handleSign` does fire there and the literal assertion cannot hold. The regression test I added instead is 'holds the payout wait state after retry until the wallet key resolves' in ResumeDepositContent.test.tsx: fresh render while disconnected, connect, no auto-run, Retry, wait state, key resolves, then exactly one `handleSign`.

services/vault/src/hooks/useBtcAction.ts
jonybur: Fixed in 585f7b8f. Contract as implemented: `requireBtcWallet()` passes only when a Bitcoin wallet is connected and the session dialog is confirmed; otherwise it returns false and opens the Bitcoin wallet list when no Bitcoin wallet is connected, or the chain list with its confirm step when Bitcoin is connected but the session is unconfirmed. The hook exposes `btcConnected` and `sessionConfirmed` separately, so callers and the gate copy can tell the two states apart. Whether a Bitcoin action should ever require an Ethereum wallet is the #2230 / #2232 decision and is not settled here.

jonybur:
Review fixes are pushed as `585f7b8f`. This answers the changes requested.

Ten threads produced changes, two are answered with the reasoning in the thread. The two that matter: the payout resume branch now mounts on the vault's on-chain depositor key, so a mid-session Bitcoin drop shows the prompt instead of auto-running, and `requireBtcWallet` now checks Bitcoin separately from session consent, opening the confirm step rather than the wallet list when only consent is missing.

One residual stays open and is now recorded on #2232: on a fresh load with Bitcoin disconnected the poll is off, so the branch can still auto-run when Bitcoin connects during the wait state. Draft #2501 changes the poll to the on-chain key, which is where it closes.

Two tests were checked by reverting their fix: the gate's loading state test, and the payout wait-state test.

Local checks at this head: 3,806 Vault tests in 316 files, eslint clean on every edited file.

jonybur:
@greptileai review the full PR at current commit `585f7b8f3ce7a1b241335eea55a7184b04316a18`. Check the diff against its target branch, including the changes since your last review. Verify that the review findings are addressed in the code. Report any remaining issues and the confidence score for this commit.

Known follow-ups #2505 and #2507 are deferred under #2228.

jonybur:
@greptileai review the full PR at current commit 8efd5f6ea3305cda40d17a41d76b0797e112b454. The review fixes are published and local checks pass. Please check the current code, report any remaining findings, and give the confidence score for this commit. The delayed first-mount payout issue in #2505 is now fixed and tested through the real continuation parent. #2507 remains an accepted follow-up in #2228.

services/vault/src/components/simple/ResumeDepositContent.tsx
```diff
@@ -98,14 +101,40 @@ export interface ResumeSignContentProps {
...
 
-export function ResumeSignContent({
+export function ResumeSignContent(props: ResumeSignContentProps) {
+  const [lastWalletKey, setLastWalletKey] = useState(props.btcPublicKey);
+  useEffect(() => {
+    if (props.btcPublicKey) setLastWalletKey(props.btcPublicKey);
+  }, [props.btcPublicKey]);
+  // Keep the active flow and its cancel control mounted during a key read.
+  // A new signing attempt still needs the current wallet key.
+  const btcPublicKey = props.btcPublicKey ?? lastWalletKey;
```
gbarkhatov, line 106: Worth a decision. Falling back to `lastWalletKey` keeps `ResumeSignContentConnected` mounted through key loss. At `468fa99c` that loss unmounted the component, and the unmount was what aborted an in-flight signing (`usePayoutSigningState.ts:148-159`, the only abort path). Now an attempt already running keeps its start-time wallet and signal: VP polling and signature submission continue after Bitcoin disconnects. `ResumeDepositContent.test.tsx:957-986` pins `signing: true` surviving the disconnect.

#2232 says active payout attempts stop when their required wallet session changes, and reconnect requires an explicit retry. There is no value risk, because the payout-address guard and the SDK checks ran at start. It only matters where the provider object stays usable after disconnect (an extension, a Ledger transport).

One shape: keep the view and its cancel control mounted, but abort `abortRef` (and `handleCancel` for Ledger) when `walletKeyReady` goes false during `signing`. The existing re-offer path then requires an explicit re-sign.

services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
```diff
@@ -880,46 +902,87 @@ describe("ResumeSignContent — post-cancel re-offer", () => {
...
-
-  it("re-runs the ceremony only on the explicit CTA click after a recorded cancel", () => {
-    markPayoutSignCanceled(baseActivity.id);
-
-    const { getByTestId } = renderSign();
-    fireEvent.click(getByTestId("sign"));
-
-    expect(handleSign).toHaveBeenCalledTimes(1);
-    expect(getByTestId("started").textContent).toBe("true");
+  it.each(["absent", "connected", "confirmed", "key", "canceled"] as const)(
```
gbarkhatov: Low. This is one `it.each` over five cases with four `if (loss === …)` branches, and the case label doubles as a mock field name (`btcActionWallet[loss] = false`). CLAUDE.md prefers "three similar test functions … over one parameterized abstraction", read top to bottom. Coverage is fine as is; splitting it into five named tests would make each regression name itself.

services/vault/src/components/Wallet/BtcActionGate.tsx
```diff
@@ -0,0 +1,74 @@
...
+    );
+  }
+
+  return (
+    <div className="mx-auto flex max-w-[564px] flex-col gap-4">
+      <Heading variant="h5">{COPY.wallet.btcAction.heading}</Heading>
+      <p className="text-accent-secondary">
+        {btcConnected && !sessionConfirmed
+          ? COPY.wallet.btcAction.confirmBody
+          : COPY.wallet.btcAction.body}
```
gbarkhatov: Low. When the wallet is connected but its key hasn't resolved (`connected && !ready`), this renders `btcAction.body`, the "connect your Bitcoin wallet" copy, next to a disabled Retry (`:55`). If `getPublicKeyHex` failed, `useBtcPublicKey` returns an undefined key and nothing wires its `refetch`, so the user is told to connect a wallet that is already connected, with no enabled action. Before this delta that state was an endless wait screen, so it's no worse a dead end, but the copy now points the wrong way. Rendering the `resolving` loader, or a key-read error, for that state would fix it.

gbarkhatov (dismissed):
Follow-up to my review at `468fa99cca`. This is a fast-forward of `585f7b8f` and `8efd5f6e`. The payout-gate blocker is fixed, and all CI checks pass.

| # | Prior finding | Verdict at `8efd5f6e` |
|---|---|---|
| T1 | The payout gate can't render, and connecting Bitcoin auto-starts signing | **Fixed.** The admission latch only goes true→false (`PostDepositContinuationView.tsx:127-130`). Auto-run needs readiness held since mount (`ResumeDepositContent.tsx:161`), and the `absent` case pins that connecting does not start it |
| T2 | The Bitcoin gate is really a whole-session gate | **Partially fixed** (dialog target, fields, copy). The Ethereum requirement is **deferred** to #2230/#2232 |
| T3 | Reclaim gate unreachable while disconnected | **Declined, deferred** to #2232 |
| T4 | Gate shows "Connect" during a saved-session restore | **Fixed** (`BtcActionGate.tsx:31-40`) |
| T5 | Gate "Retry" collides with the E2E Retry matcher | **Fixed.** The gate renders the prompt or its children, never both |
| T6 | Prompt text reused as inline error; inline strings in `useRefundState` | **Fixed** (`COPY.btcAction.error`) |
| T7 | Test mocks hide the unconfirmed-session case; `openDeposit` untested | **Partially fixed**, rest declined |
| T8 | Tests unmount before the assertion | **Partially fixed.** `DepositSignContent` and WOTS cancel now rerender live; the `useArtifactDownload` and `useVaultActions` cases are unchanged |
| T9 | Duplicated `stripHexPrefix`; early return can't mark a download stale | **Fixed** (abort in place, `useArtifactDownload.ts:146-154`). My suggested placement was wrong: it would have opened the save picker before refusing |
| T10 | `invalidate()` changed scope silently | **Fixed** (comment restored) |

**Jeremy's change request (5175915632):**
- **Payout resume bypass:** addressed, by the same evidence as T1. The test he asked for (render fresh with Bitcoin already connected, expect no sign until a click) doesn't match #2232, in my view. #2232 forbids *connecting* from starting the action. Opening the continuation with the wallet already connected is the user's gesture, and `ResumeDepositContent.test.tsx:918-924` pins that deliberately. The regression that matters is the `absent` case at `:925-956`.
- **Ethereum state in `useBtcAction`:** partially addressed and explicitly deferred rather than settled here. Only he can clear the review.

Checked, and fine:
- Every attempt re-runs `requireBtcWallet()` inside `handleSign`.
- Retry and Sign are withheld while the wallet key is undefined (`:259-263`).
- A wallet switch to a different key fails closed at the payout scriptPubKey guard.

Two inline points, the first worth a decision.

services/vault/src/components/simple/PostDepositContinuationView.tsx
```diff
@@ -123,6 +124,10 @@ export function PostDepositContinuationView({
   onClose,
   onAdvancedWithdraw,
 }: PostDepositContinuationViewProps) {
+  const { connected } = useBtcAction();
+  const payoutReady = connected && !!btcPublicKey;
+  const [payoutAdmitted, setPayoutAdmitted] = useState(payoutReady);
```
jrwbabylonlab: This is seeded before `useBtcPublicKey` has resolved (its first render is always `undefined`) and it never transitions false→true, so `autoStart` is always false in production: every payout resume shows the connect prompt with Retry to a user whose wallet is already connected, and the opening auto-run never happens. The tests pass the key on the first render, which hides it.

jonybur:
@greptileai review the full PR at commit 599ae31002011fec4007641124e375f3a67bf77d against main.

The review fixes cover the first asynchronous wallet-key read, cancellation when the required wallet session changes, late signing results, and explicit Retry after reconnect. Please check the current code and tests, report any remaining findings, and give the confidence score for this commit.

All 3,814 Vault tests pass locally. The production build and lint pass. #2507 remains an accepted follow-up under #2228.

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: Fixed in 599ae31002011fec4007641124e375f3a67bf77d. The active payout attempt now stops when its wallet, key, account, or required session changes. Cancellation uses the original provider. Checks around wallet calls reject late results and prevent further signing or submission. The view keeps Cancel until the active call settles. Then the wallet error permits an explicit Retry. Tests cover early checks, device cancellation, late signatures, and immediate cancellation. All 3,814 Vault tests and all 12 CI checks pass.

services/vault/src/components/simple/__tests__/ResumeDepositContent.test.tsx
jonybur: Fixed in 599ae31002011fec4007641124e375f3a67bf77d. Separate named tests replace the conditional it.each. Each case shows its wallet change and expected Retry behavior directly. The reconnect and recorded-cancel coverage remains. All 3,814 Vault tests pass.

services/vault/src/components/Wallet/BtcActionGate.tsx
jonybur: Fixed in 599ae31002011fec4007641124e375f3a67bf77d. A connected wallet with an unresolved key now shows the resolving loader. Cancel stays available. It no longer shows the connect prompt beside a disabled Retry. The gate test covers this state.

services/vault/src/components/simple/PostDepositContinuationView.tsx
jonybur: Fixed in 599ae31002011fec4007641124e375f3a67bf77d. A continuation opened with a confirmed wallet retains permission while its first key read completes. Session loss or loss of a previously available key removes that permission. Reconnect requires Retry. The regression uses the real useBtcPublicKey hook and its initial undefined key. All 3,814 Vault tests and the production build pass.

services/vault/src/components/simple/ResumeDepositContent.tsx
jonybur: The reconnect bypass is fixed in 599ae31002011fec4007641124e375f3a67bf77d. The continuation retains the opening action while the first key read finishes. A disconnect, consent loss, or later key loss removes that permission. Reconnect and delayed status cannot start a pending payout. Active payout work also stops when the required session changes, and late results cannot start more work or report success.

The retained behavior treats opening the continuation with an already confirmed wallet as the explicit action. It does not require a second Retry on that path. The tests therefore cover reconnect after opening, and separately cover the initial asynchronous key read through the real useBtcPublicKey hook. This is the behavior distinction behind the earlier test request.

All 3,814 Vault tests pass. All 12 CI checks pass, and Greptile gives this commit 5/5. Please review the updated behavior and clear the change request if it meets your concern.

services/vault/src/hooks/useBtcAction.ts
jonybur: The dialog and copy now distinguish a missing Bitcoin wallet from a connected wallet that still needs session confirmation. The latter opens the chain list and shows the confirmation message.

The Ethereum requirement remains in this PR. Vault still requires both wallets, and useBtcAction uses the confirmed required-wallet session. The consent decision is recorded in #2354 and adopted by the merged #2491. Ethereum-first behavior remains in #2501. This PR does not change that required-wallet set.

Please confirm whether that PR boundary addresses this concern. Current commit: 599ae31002011fec4007641124e375f3a67bf77d. CI passes and Greptile gives 5/5.

jonybur:
@greptileai review the full PR at commit 3a1c1cb121d200cf0d9f8ebcb29b26a554ff0f18 against main. This commit includes the latest main changes and preserves the payout session fixes from 599ae31002011fec4007641124e375f3a67bf77d. Check wallet consent, cancellation, reconnect, the first key read, and delayed continuation status. Report findings and a confidence score for this commit.

services/vault/src/components/Wallet/BtcActionGate.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const [admitted, setAdmitted] = useState(connected && autoStart);
+  const [started, setStarted] = useState(connected && ready && autoStart);
+  if (admitted && (!connected || !autoStart)) setAdmitted(false);
+  if (!started && admitted && connected && autoStart && ready) setStarted(true);
+
+  // Keep the active flow and its Cancel control mounted after wallet loss.
+  if (started) return <>{children}</>;
+
+  return (
+    <div className="mx-auto flex max-w-[564px] flex-col gap-4">
```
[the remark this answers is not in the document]
jonybur: Addressed in 312622fd60a91901d06e023426cfba927d721d95. The prompt now uses DEPOSIT_CONTENT_MAX_WIDTH_CLASS from the shared deposit layout constants. The width stays at 564px. The prompt test, production build, and lint pass.

jonybur:
@greptileai review the full PR at commit 312622fd60a91901d06e023426cfba927d721d95 against main. The inline width finding is fixed with a shared layout constant. The update also cancels the original device request when the payout view unmounts and removes duplicate test properties from the main merge. All 138 targeted tests, production build, and lint pass locally. Report any remaining findings and the confidence score for this exact commit.

gbarkhatov (approved):

PR #2483, opened 2026-09-09
test(wallet-connector): verify real wallet session restore

jonybur:
## Outcome

Real OKX and Keplr tests cover dialog close, reopen, and page reload. This is partial coverage for #2355.

## Change

- Keep both wallet identities when the dialog closes and reopens.
- Require final Connect before confirmation.
- Restore the same identities and confirmation when persistence is enabled.
- Save no session and clear the selection on reload when persistence is disabled.

## Safety

Reuse the existing wallet setup with fresh generated identities. Add no mock data. Keep screenshots, video, and traces off.

## Verification

- Real OKX 4.15.19 and Keplr 0.13.46: 2 tests passed with no retries in the final run.
- Unit tests: 295 passed. Package checks: 23 passed.
- Lint passed with 89 existing warnings. Formatting and diff checks passed.
- After integrating main: wallet build, 295 unit tests, 23 package checks, story lint, formatting, and two-case test discovery passed. The real-wallet test code is unchanged; no new live run is claimed.

An earlier run missed the initial OKX popup. The isolated rerun and final full run passed.

## Links

#2463 is merged. This PR targets `main`. The integration merge preserves exactly the original two-file session change.

Tracks #2355. #2355 stays open. Parent: #2230. Epic: #2228.

Consumer routes and the wider wallet matrix still need coverage.


jonybur (commented):
Reviewed the wallet-connector half at 35989acf — the last two commits on the branch. The first four are #2463, which squash-merged into main three seconds before this head was pushed, so the diff currently replays them; a rebase will drop them.

The session-restore spec itself is the right shape. Driving both cases from one parameterised body means the persistent and non-persistent paths cannot drift, `data-confirmed` on the story button gives the reload assertion something real to read rather than inferring from a rendered address, and comparing the reopened sections against the identities captured before the close is what makes the test about restore rather than about connecting twice. Asserting the empty `baby-connected-wallet-accounts` in the non-persistent case pins the storage side too, not just the UI. Generating a fresh mnemonic and password per context, redacting both out of any setup error, and keeping trace, screenshot and video off are all correct for a suite that drives a real seed phrase through a real extension.

Two things I would fix before merge.

**1. The Keplr seed grid has no shape guard, unlike the OKX one.** `keplr.ts:26-30` now fills the i-th match of a page-wide `input[type=text], input[type=password]` selector, anchored to nothing. The selector it replaced was at least scoped to the grid through its label text. `okx.ts` handles the same problem properly: it counts the boxes, compares the count against `words.length`, and throws an error naming `selectOkxSeedWordCount` as the thing to re-derive when OKX changes its import UI. Keplr gets none of that, so if a future Keplr build renders one extra text input above the grid, every word shifts by one and the run fails somewhere downstream on an unrelated assertion instead of saying the seed grid changed. Count the matched inputs against `words.length` before the loop and throw the same kind of message.

**2. The mnemonic and password guards name an environment variable that this path no longer reads.** `keplr.ts:10-11` still throws "Missing E2E_WALLET_MNEMONIC in environment variables" and "Missing E2E_WALLET_PASSWORD in environment variables", and `okx.ts` carries the same pair. Both values are now generated inside `setupExtensions`, so the branch is unreachable through the fixture, and anyone who does reach it is sent to configure a variable nothing consults. Either drop the guards or reword them to name the generator.

Not blocking, and noted rather than raised: removing the swallow from `checked.count()` and `uncheck()` in the OKX welcome-screen frame loop makes a frame that detaches mid-iteration a hard setup failure, and CI runs this with retries disabled. Failing loud is the right instinct, but a detaching frame is not a product failure — worth watching for flakes over the first few CI runs.

The main workflows have not run on this head yet, so there is no CI evidence for it beyond the two checks that reported.

jrwbabylonlab (approved):

PR #2484, opened 2026-09-09
fix(vault): keep the deposit step counter inside its card on mobile

### LINKED ISSUE #2419: Deposit progress: the step counter overflows its card on mobile
## Context

The step-group counter in the Deposit Progress modal, at narrow viewport widths.

## Changes

**Now:** `GroupHeader` renders `<div className="flex items-center gap-3">` with the group title as `flex-1` and no `min-w-0`/`truncate` (`services/vault/src/components/simple/DepositProgressView/GroupHeader.tsx:99-108`), and the `x/y` counter as a trailing span with no `shrink-0` (`:109-117`). The counter string comes from `COPY.groups.stepCounter` (`services/vault/src/copy.ts:387-388`). The card wrapper is `rounded-2xl bg-secondary-highlight p-4` with no `overflow-hidden` (`GroupBlock.tsx`).
**Expected:** the counter stays inside the card at mobile widths.
**Where:** `GroupHeader.tsx:99-117`

Flex items default to `min-width: auto`, so a long title such as "Activate BTCVault" refuses to shrink below its intrinsic width and pushes the counter past the card edge — nothing above it clips overflow.

## Not proven

This is the classic flexbox `min-width: auto` overflow and matches the symptom, but it was not reproduced. Check at ~360px width whether adding `min-w-0`/`truncate` to the title at `GroupHeader.tsx:99` fixes it, and rule out `SplitGroupedProgress.tsx`'s `flex gap-6` column layout as the actual container.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1787799557932569)

> In Mobile mode small issue with 0/3 getting out of the box.

</details>

jeremy-babylonlabs:
Closes #2419

## What

The `x/y` step-group counter in the Deposit Progress modal overflowed its card at mobile widths.

The group title was `flex-1` with no `min-width: 0`, so it refused to shrink below its content width and pushed the counter past the card edge. The counter had no `shrink-0`, so it could also be compressed.

- Title gets `min-w-0` so it can shrink, plus `overflow-hidden text-ellipsis` so it clips.
- Counter gets `shrink-0` so it always renders whole.

## Why not `truncate`

`truncate` sets `white-space: nowrap` at every breakpoint. This component gates nowrap behind `xl:` (added in #2296) precisely so narrow viewports wrap instead of overflowing. Using the shorthand would clip text that should wrap on mobile — a new bug in place of the old one. `overflow-hidden text-ellipsis` leaves the existing breakpoint-gated `whitespace-nowrap` in charge.

## Testing

One test in `GroupedProgress.test.tsx` asserts the mechanism: the counter and the title are siblings in one flex row, the title carries `min-w-0 overflow-hidden`, and the counter carries `shrink-0`. The group header and its step counter are photographed at 390×844 by the existing visual harness (`deposit-progress-step-NN--mobile.png` and `deposit-progress-split--mobile.png`), so a reintroduced overflow surfaces as a diff in the visual-regression report.

`pnpm --filter vault run lint` — 0 errors. `pnpm --filter vault run test` — full suite green.


jonybur (dismissed):
Verified at f36d0f86. The diagnosis holds: `flex-1` alone leaves `min-width: auto` on the flex item, so the title floors at its content width and the counter is pushed out. `min-w-0` on the title plus `shrink-0` on the counter is the minimal pair that fixes it, and it is the right way round — the counter is short and fixed, the title is the variable-length one.

`overflow-hidden text-ellipsis` rather than `truncate` is correct for the same reason as #2419: `truncate` would force `white-space: nowrap` at every breakpoint and clip text that is meant to wrap on narrow viewports, trading one overflow bug for another.

Switching the counter to `twMerge` keeps the conditional colour classes merging properly alongside the new `shrink-0`. Both assertions fail on revert, and the ellipsis test pinning the absence of `truncate` is what stops the regression coming back. CI green.

jrwbabylonlab (dismissed):

jonybur (dismissed):
`min-w-0` on the title plus `shrink-0` on the counter is what keeps the counter inside the card — the title was previously holding the row open at its content width. Nothing else in the header depends on the title being able to grow.

services/vault/src/components/simple/DepositProgressView/GroupHeader.tsx
```diff
@@ -87,7 +87,7 @@ export function GroupHeader({
         as="span"
         variant="body1"
         className={twMerge(
-          "flex-1 font-medium",
+          "min-w-0 flex-1 overflow-hidden text-ellipsis font-medium",
```
gbarkhatov: `text-ellipsis` is inert here, and the rest of the fix may not bind either.

`text-overflow: ellipsis` only renders on a single-line, non-wrapping overflow context. Nothing gives this span `white-space: nowrap`: core-ui's `Text` renders exactly one element, joining only the variant class with the caller's `className`, and across all of `Text.css` no variant sets `white-space`, truncation or line-clamp — `body1` is `@apply text-base tracking-0.15`. The only `whitespace-nowrap` in this directory is in `FeeRateSelector`, a different component. So the class ships with no effect, which the ZERO DEAD CODE policy rules out.

Separately, `min-w-0` probably doesn't bind: `min-width: auto` on a flex item floors at **min-content** — the longest word — not max-content. The longest word across the four titles (`copy.ts:383-386`) is "BTCVault"/"Register", roughly 70px, well under the width available to the title at 390px. `flex-1` was already present before this diff.

I can't disprove the overflow without a browser, so treat this as **unproven rather than wrong**. A 390px before/after screenshot would settle it; if it can't be reproduced, #2419 might be better closed as not-reproducible than as fixed. Worth noting too that `overflow-hidden` on a wrapping block clips nothing — the box grows in height instead. It only engages on an unbreakable token, which it then cuts mid-glyph with no ellipsis, per the above.

services/vault/src/components/simple/DepositProgressView/__tests__/GroupedProgress.test.tsx
```diff
@@ -180,6 +180,20 @@ describe("GroupedProgress", () => {
...
+
+    const title = screen.getByText(COPY.deposit.groups.registerDeposit);
+    const counter = screen.getByText(COPY.deposit.groups.stepCounter(0, 6));
+    const row = title.parentElement as HTMLElement;
+
+    expect(counter.parentElement).toBe(row);
+    expect(row.className).toContain("flex");
+    expect(title.className).toContain("min-w-0");
+    expect(title.className).toContain("overflow-hidden");
+    expect(counter.className).toContain("shrink-0");
```
gbarkhatov: Four of these five assertions pin Tailwind class strings — and the commit message claims the opposite. `1103061` reads *"test(vault): assert the step counter row structure instead of its classes"*.

jsdom loads no Tailwind and computes no layout, so nothing here can observe the overflow this test is named after. `expect(counter.parentElement).toBe(row)` at line 190 is the only structural assertion, and it was already green before the fix — `GroupHeader.tsx:84` (`flex items-center gap-3`) is untouched by this diff, and the counter was already its direct child. `toContain("flex")` at line 191 is also loose: it matches `flex-col` and `flex-1` too.

Suggest keeping line 190 and the `min-w-0` pin — that one does catch a future `twMerge` collapse — dropping the `flex` and `overflow-hidden` pins, and renaming the test to what it actually asserts. The layout claim belongs in `e2e/visual/`, which already photographs this at `{ name: "mobile", width: 390, height: 844 }`.

gbarkhatov (dismissed):
Reviewed via three independent passes, deduped and verified against the source at `1103061`.

`shrink-0` on the counter is correct and load-bearing — because the title's flex basis is `0%`, shrinkage distributes by basis x factor, so the counter previously absorbed all of any compression. Keep that.

The concern is that the rest of the fix may not do anything, and nothing in the PR demonstrates otherwise. Two comments inline. Approving — none of this is a reason to hold the branch, but I'd rather #2419 not be closed on an unproven fix.

**Verified clean:** copy rule respected (both strings via `COPY`, nothing inlined); no `data-testid` touched, so no E2E hook risk; no transition or motion touched; the fixture is right — `steps.ts:101` gives `startStep: 1, endStep: 6`, so `stepCounter(0, 6)` at `currentStep={1}` is correct.

**One caveat on CI:** the green `visual` check is not pixel evidence. `.github/workflows/visual-regression.yml:68` states it "reports and comments, it does not fail the build", so it passing says nothing about whether the mobile screenshot changed.

gbarkhatov (approved):
Follow-up to my review at `110306176b`. The rebase carried both reviewed commits over unchanged, so the only delta is `f538cdc11` (use `truncate` on the group title).

**Merge blocker (not code):** all 3 commits are unsigned (`git log %G?` = `N`), and `main`'s ruleset has `required_signatures`. `validate` fails for that reason.

| # | Prior finding | Verdict at `f538cdc1` |
|---|---|---|
| T1 | `text-ellipsis` does nothing without `nowrap`; `min-w-0` likely no effect; fix unproven | **Partially fixed.** `truncate` adds `nowrap`, so the ellipsis now renders (`GroupHeader.tsx:90`). I withdraw the `min-w-0` sub-point: with `nowrap` it is required. Still unproven: the visual bot reports no change at 390px against the merge-base, and the layout spec's right-edge check predates this PR, so nothing reproduces #2419 |
| T2 | The test pinned class strings | **Fixed.** It now checks sibling structure plus `min-w-0` |

**The PR description now argues against the code.** "Why not `truncate`" says the shorthand would break mobile wrapping because "this component gates nowrap behind `xl:` (added in #2296)". That `xl:whitespace-nowrap` lives in `PositionStatCards.tsx`; `GroupHeader` has none at base or head. The Testing section also still describes the old `overflow-hidden` / `shrink-0` assertions. Worth rewriting both before a squash merge puts that reasoning in history.

Checked, and fine: no `data-testid` is touched, copy is unchanged, and the accessible name still carries the full title.

PR #2485, opened 2026-09-09
fix(vault): wire the market header Borrow button to the borrow form

### LINKED ISSUE #2431: Market Info: the Borrow button is hardcoded disabled
## Context

A user with both wallets connected found the Borrow button in the Market Info window greyed out.

## Changes

**Now:** the button is unconditionally disabled and has no handler (`services/vault/src/components/pages/BorrowingMarketsData/index.tsx:393-395`):

```tsx
<button type="button" className={NEUTRAL_BUTTON_CLASS} disabled>
  {COPY.marketData.borrowAction}
</button>
```

No wallet, position or market state is consulted. It is a placeholder, not a gated call to action.

**Expected:** either wire it to the borrow flow, or stop rendering it until it works — a permanently grey button reads as a bug to everyone who tries it.
**Where:** `BorrowingMarketsData/index.tsx:393-395`

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788433818105909)

> I am trying to borrow directly from "Market Info" window: For some reason the button is greyed out - my wallets are connected. Maybe the button is not configured or I am doing something wrong?

</details>

jeremy-babylonlabs:
Closes #2431

The Market Info header's Borrow button rendered permanently disabled. It had no wiring to any flow.

The button now opens the loan overlay's borrow form for the market's own reserve. It navigates search-only — `?reserve=<id>&tab=borrow` on the current pathname — so the overlay renders over the market page. The markets route already sits under `AaveOverlayLayout`, which documents a market page as a valid entry point. This mirrors `useLoanActions.goToReserve`, used by the Overview and Loans entry points.

The button is disabled only when there is no real reserve to borrow from: demo mode, or no reserve resolved for the route. The button does no wallet gating. The overlay owns the wallet prompt, as it does for every other entry point.

Tests: clicking Borrow in the header navigates to `/markets/1?reserve=1&tab=borrow`; the button is disabled for a demo fixture.


jonybur (dismissed):
Verified the cleanup is complete at 344dadc4. `COPY.marketData.borrowAction` had exactly one consumer, the button being removed; `COPY.overview.borrowAction` is a distinct key still read by `PositionStatCards.tsx:122` and asserted in `OverviewSection.test.tsx`, and it is untouched. `NEUTRAL_BUTTON_CLASS` keeps its remaining consumer in `LiquidationAnalysisSection.tsx:131`, so the import is the only thing that became orphaned and the constant correctly stays. Nothing under `services/vault/e2e/` referenced the removed control, and no `data-testid` was dropped.

Collapsing the outer wrapper is right — it existed only to sit the button beside the identity block, and the inner `min-w-0` that did the actual shrink work is carried over onto the surviving div.

Removing rather than wiring is the correct scope call for a control that consulted no wallet, position or market state. CI green.

jrwbabylonlab (dismissed):

jonybur (dismissed):
Checked that the search-only navigation actually lands somewhere: `location` here is the router's, not `window.location`, and `/markets/:market` sits under `AaveOverlayLayout`, which mounts `LoanFlowOverlay` whenever the `reserve` param is present. So the button opens the form over the market page rather than painting another route behind the dialog.

Routing by reserve id rather than symbol keeps the indexer out of the decision, consistent with the rest of `routes.ts`. Disabling on the demo fixture is correct — there is no on-chain reserve behind it to borrow from.

services/vault/src/components/pages/BorrowingMarketsData/__tests__/BorrowingMarketsData.test.tsx
```diff
@@ -412,6 +418,48 @@ describe("BorrowingMarketsData", () => {
...
+
+    renderPage("1");
+
+    const header = within(screen.getByTestId("market-section-identity"));
+    fireEvent.click(
+      header.getByRole("button", { name: COPY.marketData.borrowAction }),
+    );
+
+    expect(screen.getByTestId("location-probe").textContent).toBe(
+      "/markets/1?reserve=1&tab=borrow",
```
gbarkhatov: This assertion can't distinguish the reserve id from the URL slug, so it doesn't actually pin the id-not-symbol rule it exists to protect.

`USDC_RESERVE.reserveId` is `1n` and its underlying `0x1111…` is unknown to the token registry, so `getMarketSlug` takes its fallback branch — `return symbol ? symbol.toLowerCase() : reserveId.toString()` (`routes.ts:105-110`) — and yields `"1"`. The route param and the reserve id are therefore the same literal string, and `/markets/1?reserve=1&tab=borrow` passes identically whether the component emits `selectedReserve.reserveId` (correct) or `params[MARKET_PARAM]` (the indexer-steerable path this file's own header comment forbids). A refactor that rebuilds the search from the URL segment would keep this test green while reintroducing symbol-slug routing.

No new fixture needed — `REGISTERED_USDC_RESERVE` already exists at line 139, same `reserveId: 1n` via spread with the underlying overridden to the mainnet USDC address the registry knows, so its slug is `usdc`. Rendering `renderPage("usdc")` and asserting `/markets/usdc?reserve=1&tab=borrow` would make the two identifiers distinguishable. The sibling test at line 463 already uses the fixture exactly that way.

services/vault/src/components/pages/BorrowingMarketsData/__tests__/BorrowingMarketsData.test.tsx
```diff
@@ -412,6 +418,48 @@ describe("BorrowingMarketsData", () => {
     ).toBeInTheDocument();
   });
 
+  it("opens the borrow form for the routed reserve without leaving the market page", () => {
```
gbarkhatov: The name promises a borrow form this test never mounts. `renderPage` builds a bare `MemoryRouter` with only the `BorrowingMarketsData` route plus the location probe — `AaveOverlayLayout` and `LoanFlowOverlay` aren't in the tree, so no form can exist in this render.

The "without leaving the market page" half is genuinely verified by the pathname assertion; "opens the borrow form" is true only transitively, via the overlay's behaviour which is covered separately in `Detail/__tests__/LoanFlowOverlay.test.tsx`. CLAUDE.md's test philosophy asks the name to describe the exact behaviour verified — something like *"navigates search-only to the routed reserve's borrow params, keeping the pathname"* would match what this actually asserts.

gbarkhatov (dismissed):
Reviewed at `9e2fa9e5` via three independent passes, deduped and verified against the source.

**Small and correct — no merge blockers.** I traced the wiring end to end and it needs nothing outside the diff: `/markets/:market` sits inside the `AaveOverlayLayout` route, the layout mounts `LoanFlowOverlay` on any truthy `?reserve=` with no pathname condition, and `ReserveDetailPanel` renders `LoanCard` on the borrow tab. The reserve always resolves, since `borrowableReserves` is a strict subset of the `allBorrowReserves` the overlay resolves from — no dead end. The identifier round-trip also holds: `getReserveDetailSearch` emits `reserveId.toString()` and `parseReserveId` accepts `/^\d+$/`, so no symbol slug survives the trip.

Everything below is test strength, not behaviour. Approving.

- *"The `onClick` early return is dead code."* No. It is unreachable (when `!isDemo && selectedReserve === null` the page already early-returns `reserveNotFound` before the button renders) *and* required anyway, since TypeScript can't narrow `AaveReserveConfig | null` from the `disabled` prop.
- *"`disabled={isDemo || selectedReserve === null}` under-disables non-borrowable reserves."* No — `borrowableReserves` is already filtered to `borrowable && !paused && !frozen`, so a paused or frozen reserve never reaches the button. Zero-liquidity stays enabled, which is parity with `AssetSelectionPanel`, and the borrow form owns amount validation.
- *"An E2E `data-testid` was dropped by the div-to-section change."* No. The removed lines carry no testid, and `services/vault/e2e/` has zero hits for `market-section` — the real-wallet borrow flow enters via `loans-borrow-button`.
- *"A user-facing string was inlined instead of living in `copy.ts`."* No — the button renders the pre-existing `COPY.marketData.borrowAction` (copy.ts:1685) and the diff adds no literal.

**Verified clean:** fails loudly at every stop rather than silently (disconnected → connect prompt; unresolvable id → `reserveNotFound` / `reserveLinkOutdated`; identity failure → `ReserveIdentityBlock`), and closing drops only the search, returning the depositor to the market page rather than `/loans`. `LOAN_TAB` was already imported; `getReserveDetailSearch` is the only added import and is used. No magic numbers, no motion touched, and none of CLAUDE.md's critical paths cover these files.

One deliberate non-issue worth recording: the inline `onClick` duplicates `useLoanActions.goToReserve` verbatim, but reusing that hook would drag its required `borrowedAssets` prop — and `useAaveBorrowedAssets` / `useAaveUserPosition` — onto a page that deliberately renders without a wallet. Four lines of duplication is the right call.

gbarkhatov (approved):
Follow-up to my review at `9e2fa9e522`. The rebase carried the three reviewed commits over unchanged, so the only delta is `9fe257f52`, which touches tests only.

| # | Prior finding | Verdict at `9fe257f52` |
|---|---|---|
| T1 | The borrow test's market slug was `"1"`, the same as the reserve id, so a slug/reserve-id mix-up could not fail it | **Fixed.** It now uses `REGISTERED_USDC_RESERVE` and expects `/markets/usdc?reserve=1&tab=borrow` (`BorrowingMarketsData.test.tsx:421-434`) |
| T2 | The test name promised a borrow form the render never mounts | **Fixed.** The name now describes the navigation it asserts (`:421`) |

The test now catches both mistakes. Building `?reserve=` from the slug, or the path from the reserve id, fails the `toBe` at `:431-433`.

No new defects in the delta. It changes no testids or copy and leaves no dead code.

Not a code issue: `validate` (Authenticate Commits) fails on this head. All four commits are unsigned (`git log %G?` = `N`), so they need re-signing before merge.

PR #2486, opened 2026-09-09
fix(vault): align the liquidation event card stats at narrow widths

### LINKED ISSUE #2434: Liquidation event card: Collateral / Liq Price / Distance misalign at narrow widths
## Context

Reported by a community manager: the Distance field on the Liquidation Analysis event card misaligns under unusual viewport widths, for example with developer tools docked.

## Changes

**Now:** the Collateral / Liq Price / Distance triplet is plain flexbox with no wrapping and no column sizing (`services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx:139-155`):

```tsx
<div className="flex gap-16">
  <Stat label={COPY.liquidations.events.collateral}>…</Stat>
  <Stat label={COPY.liquidations.events.liqPrice}>…</Stat>
  <Stat label={COPY.liquidations.events.distance}>…</Stat>
</div>
```

`Stat` (`LiquidationEventCard.tsx:59-76`) puts no `min-w-0` or truncation on its label or value. With no `flex-wrap`, no fixed or minimum column width and a fixed `gap-16`, the three items compress unevenly once the container narrows.

**Expected:** the columns keep their alignment, or wrap, at narrow widths.
**Where:** `LiquidationEventCard.tsx:59-76,139-155`

## Not proven

"Extreme conditions … opening developer tools" implies a specific width. Reproduce with devtools docked and confirm it is this card rather than the chart hover popover, which uses the same word from a different key (`COPY.liquidations.popover.distance`, `copy.ts:1811`, rendered in `liquidationChartData.ts:446`).

Related: #2418 renames this field to "% to Liquidation".

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788534237626269)

> Our community manager reported that under extreme conditions (for example, when opening developer tools), the `distance` field is displayed with some misalignment.

</details>

jeremy-babylonlabs:
Closes #2434

## What

The Collateral / Liq Price / Distance triplet on the Liquidation Analysis event card misaligned at narrow widths (reported with devtools docked). Three flex children with different natural widths and a fixed `gap-16` had no room to give, so the columns drifted.

One change inside the shared `Stat`, which covers all three columns:

- `flex-1 min-w-0` on the wrapper — equal share, and no intrinsic minimum, so a wide value can no longer push its neighbours.
- `block overflow-hidden text-ellipsis` on the value. `block` is required: `text-overflow` does not apply to an inline `span`.
- `title` on the value with the full figure, so an ellipsized `$104,231` can still be read on hover. `Stat` takes a `value: string`, so no call site can drop it.

`gap-16` is untouched; at the narrowest supported width the 128px of total gap still fits once the columns can shrink.

## Why not `truncate`

It forces `white-space: nowrap` at every breakpoint. Values here are allowed to wrap at their space on narrow screens; only unbreakable overflow should ellipsis. Same reasoning as #2419.

## Testing

Tests assert the `title` on all six stat values and that the three top stats are siblings in one row container.

`pnpm --filter vault run lint` — 0 errors. `pnpm --filter vault run test` — full suite green.

## Width bands

Between 1024px and 1279px the stats row is about 291px wide (viewport minus the 242px sidebar, page padding and card padding, halved). The 32px gap needs 394px, so the row wraps to two lines in that band. That wrap is accepted. The 64px gap applies from 1536px (`2xl`), where the row is 547px wide and the three stats fit on one line.


jonybur (dismissed):
Verified at 6fc07a0b. `Stat` has exactly two call sites in this file and both are three-column `flex gap-16` rows — the Collateral / Liq Price / Distance triplet and the BTC Remaining / Debt Remaining / HF After row — so `flex-1 min-w-0` is correct at every use and does not change layout anywhere it was not meant to.

`block` on the value span is genuinely required: `text-overflow` does not apply to a non-replaced inline box, so without it the ellipsis would never render. Leaving `whitespace-nowrap` off is the right trade — an unbreakable token still overflows its line box and gets ellipsized, while ordinary values keep wrapping on narrow screens. Same reasoning as #2419 and consistent with it.

The second test asserting the absence of `truncate` and of a bare `whitespace-nowrap` is what makes this stick, since either would silently reintroduce the mobile-wrap problem. CI green.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -66,11 +66,16 @@ function Stat({
...
   return (
-    <div className="flex flex-col">
+    <div className="flex min-w-0 flex-1 flex-col">
       <span className="text-sm leading-[1.43] tracking-[0.17px] text-accent-secondary">
         {label}
       </span>
-      <span className={twJoin("leading-[1.6] tracking-[0.15px]", className)}>
+      <span
+        className={twJoin(
+          "block overflow-hidden text-ellipsis leading-[1.6] tracking-[0.15px]",
```
jrwbabylonlab: These values are money, and each is a single unbreakable token (`$104,231`, `0.12345678 BTC`). At 375px the three columns split what's left after the row's two 64px gaps, so roughly 55px each, and they now render as `$…` with no `title` and no tooltip. Before this change they overflowed but stayed readable. A truncated liquidation price is a misreadable figure, not a cosmetic clip — worth a `title` at minimum.

jonybur (dismissed):
The title attribute now carries the full value on the Stat span, addressing the concern about unreadable clipped money figures at narrow widths. The Stat call sites in this file are all three-column flex-16 rows, so the min-w-0 addition is safe at every use. Block on the value span is genuinely required since text-overflow does not apply to a non-replaced inline box. CI green.

jonybur (dismissed):
Every label the `Stat` prop now takes is already declared `string` on `EventCardData`, so narrowing from `ReactNode` loses nothing at the call sites. `min-w-0 flex-1` is what gives the three columns an equal share instead of letting the widest value push the row, and the full figure stays reachable through `title`.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -58,20 +57,26 @@ function DetailRow({
...
   className,
 }: {
   label: string;
-  children: ReactNode;
+  value: string;
   className?: string;
 }) {
   return (
-    <div className="flex flex-col">
+    <div className="flex min-w-0 flex-1 flex-col">
```
gbarkhatov: **MERGE BLOCKER — this truncates the liquidation price at mobile width, where merge-base rendered it in full.**

The wrapper went from `flex flex-col` (`flex: 0 1 auto`, content-sized) to `flex min-w-0 flex-1 flex-col` (`flex: 1 1 0%`), which forces three *exactly equal* columns of (card − 128px of `gap-16`) ÷ 3 at every width — roughly 66px each at 390px. Because `$59,050` and `$8,851.62` are unbreakable tokens with no spaces, `text-ellipsis` genuinely engages on them, and this PR's own mobile artifact shows the result: `$59,050` → `$59,…`, `$8,851.62` → `$8,851.…`, against a merge-base panel where both are complete and aligned.

`min-w-0` alone is the right half of the fix — it removes the min-content floor issue #2434 identified. **`flex-1` is what starves the columns; dropping it should resolve this.** If roughly-equal columns are actually wanted, the repo already has the idiom for exactly this problem: `min-w-0 grow-[N] basis-[Npx]` with a content-sized basis, used in core-ui's `ListRow.tsx:36,46` precisely because grow factors divide only the surplus and cost nothing once every column is already shrinking.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -58,20 +57,26 @@ function DetailRow({
...
       <span className="text-sm leading-[1.43] tracking-[0.17px] text-accent-secondary">
         {label}
       </span>
-      <span className={twJoin("leading-[1.6] tracking-[0.15px]", className)}>
-        {children}
+      <span
+        title={value}
```
gbarkhatov: **MERGE BLOCKER — the `title` recovery path doesn't exist on the devices where the truncation happens.**

The two halves land differently, so to be precise: **screen readers are fine.** `overflow-hidden` clips visually but the complete text node stays in the accessibility tree, so AT still announces `$59,050` — `aria-label` would add nothing here and risks a double announcement, so it is *not* the better remedy. **Touch is the problem.** `title` renders as a hover tooltip, this span is non-focusable, and narrow widths are overwhelmingly phones. So the remedy is structurally absent in exactly the case it was added for, and the head commit's claim that every stat now has "a title with the full figure" doesn't hold for the affected user.

The fix probably isn't a different attribute but not hiding anything: once `flex-1` is gone, `[overflow-wrap:anywhere]` would let an unbreakable figure wrap rather than clip, and `title` could go entirely — on desktop it currently pops a native tooltip repeating text that is already fully visible. If a hover affordance is still wanted, this file already imports the tap-capable `Hint`.

services/vault/src/components/pages/Liquidations/__tests__/LiquidationEventCard.test.tsx
```diff
@@ -75,4 +75,42 @@ describe("LiquidationEventCard", () => {
...
+      render(<LiquidationEventCard card={STAT_CARD} />);
+
+      const [collateral, liqPrice, distance] = [
+        "1.23456789 tBTC",
+        "$104,231",
+        "-10%",
+      ].map((value) => screen.getByTitle(value).parentElement);
+
+      expect(collateral?.parentElement).toBe(liqPrice?.parentElement);
+      expect(liqPrice?.parentElement).toBe(distance?.parentElement);
```
gbarkhatov: Neither added test can fail if the alignment change is reverted.

The diff *does* contain the alignment change — `min-w-0 flex-1` at `LiquidationEventCard.tsx:68`; the head commit subject under-describes it — but nothing gates it. This test asserts that the three value spans share a grandparent, which is true at merge-base and would still be true if `min-w-0 flex-1` were deleted tomorrow. That's DOM shape rather than behaviour, which the test philosophy rules out, and jsdom computes no layout so nothing in this file can observe alignment or truncation at any width.

Its companion above is named "…so truncation cannot hide it" but asserts only that a static attribute is present — and per my comment on line 73 the figure *is* still hidden on touch, so the name promises a guarantee the code doesn't deliver. Suggest deleting the structural assertion and treating the mobile screenshot as the acceptance evidence, since the advisory capture is the only thing that can actually see this change — and it did catch it.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -133,28 +138,25 @@ export function LiquidationEventCard({ card }: { card: EventCardData }) {
       <div className="flex gap-16">
```
gbarkhatov: The `gap-16` that actually consumes the width is untouched. Two flat 64px gaps is 128px at every breakpoint — roughly half the card at 390px — spent before any column gets to shrink.

Issue #2434 asks that the columns "keep their alignment, **or wrap**". Nothing here wraps, and `min-w-0` delivers only the first half, by squeezing all three columns equally. A responsive gap (`gap-4 sm:gap-16`) or `flex-wrap` is what would deliver the issue's literal expectation, and may well be the actual fix if the original report was taken at ~300px with devtools docked.

Related, and worth settling before this lands: #2434 explicitly flagged the original report as **Not proven** — no width named, possible confusion with the chart hover popover — and asked whoever fixed it to reproduce first. The merge-base artifact shows these columns fitting, aligned and fully legible at 390px, so it would be worth stating the viewport at which the misalignment actually reproduces.

gbarkhatov (dismissed):
Reviewed at `83ea84b` via three independent passes, deduped and verified — including by opening this PR's own published visual artifacts rather than reasoning about the CSS alone.

**Please read the two blocking comments before merging: as it stands this PR regresses the exact case it is named for.** I'm approving rather than blocking, per how we treat review comments here — the call is yours — but I'd hold this one.

From `visual-diffs/pr2486/.../vault/liquidations-chart--mobile.png`, top-to-bottom (merge-base, then this PR):

| stat | merge-base | this PR |
|---|---|---|
| Liq Price | `$59,050` | **`$59,…`** |
| Debt remaining | `$8,851.62` | **`$8,851.…`** |

At 390px — the only narrow viewport this repo captures — merge-base renders both figures complete, columns aligned, nothing overflowing. This PR clips the single most important number on a liquidation card.

**A factual correction to the PR body:** it states that `block` is required because "`text-overflow` does not apply to an inline `span`". That span is a flex item of `div.flex.flex-col`, and flex children are blockified by their container per CSS Flexbox §4 — so `block` is a no-op and the stated rationale for it doesn't hold.

**Refuted — worth not re-raising:**
- *"Equal-width columns make the labels wrap by different amounts, so values misalign vertically."* The artifact is direct counter-evidence: all three labels wrap to the same number of lines and the values sit at a common offset.
- *"`flex-1` restretches the desktop columns edge-to-edge."* Real but tiny — about 3px, 0.288% of pixels, nothing truncated. A nit, not a merge concern, though a pure narrow-width fix would ideally have been a 0% desktop diff.

**Verified clean:** the `copy.ts` rule is not violated — `title={value}` passes through an already-formatted figure and the labels still come from `COPY.liquidations.events.*`, so no new entry is owed. No `data-testid` exists anywhere under `components/pages/Liquidations/`, so nothing load-bearing was dropped in the `Stat` rewrite. ZERO DEAD CODE is clean — `ReactNode` was the only orphan and it's removed. And the `children: ReactNode` → `value: string` narrowing is a genuine improvement: it makes the truncation contract expressible, and all six call sites already passed plain strings.

services/vault/src/components/pages/Liquidations/LiquidationEventCard.tsx
```diff
@@ -130,31 +134,28 @@ export function LiquidationEventCard({ card }: { card: EventCardData }) {
         {card.title}
       </h3>
 
-      <div className="flex gap-16">
+      <div className="flex flex-wrap gap-4 sm:gap-8 xl:gap-16">
```
gbarkhatov: The `xl:gap-16` step brings back a two-line row on desktop. `xl:` turns on at a 1280px viewport, and `/liquidations` also shows the 242px sidebar from 768px up (`Sidebar.tsx:29`). That leaves roughly 412–419px for this row at 1280, while three `basis-[110px]` stats plus two 64px gaps need 458px, so Distance (or HF) wraps onto its own line. The head desktop visual capture shows this: it grew from 1280×1570 to 1280×1776 against the merge-base.

The effect is that widening past 1280 turns one row back into two lines. Moving the wide gap to `2xl:gap-16`, or dropping it (the 32px gap fits at 1280), avoids that. Around 1024–1229px the row also wraps with the 32px gap; say in the PR whether that band is accepted.

gbarkhatov (dismissed):
Follow-up to my review at `83ea84b16d`. The rebase carried the three reviewed commits over unchanged, so the delta is `148f7cae5` (wrap instead of truncating) and `1a4aa2056` (responsive gaps).

**Merge blocker (not code):** all 5 commits are unsigned (`git log %G?` = `N`), and `main`'s ruleset requires the commit-authentication workflow, so `validate` fails.

| # | Prior finding | Verdict at `1a4aa205` |
|---|---|---|
| T1 | `flex-1` squeezed the columns and clipped Liq Price on mobile | **Fixed.** `min-w-0 grow basis-[110px]`, with no ellipsis (`LiquidationEventCard.tsx:68`) |
| T2 | A `title` tooltip can't be opened on touch | **Fixed.** `title` and the truncation classes are removed; `[overflow-wrap:anywhere]` added (`:72-79`) |
| T3 | Neither added test could fail if the layout were reverted | **Fixed.** Both deleted; jsdom can't see layout |
| T4 | `gap-16` left flat; no reproduction width given | **Partially fixed.** Gaps are responsive now (`:137`, `:225`), but the `xl:` step causes the inline issue, and the PR body still names no reproduction width |

No dead code is left behind: the truncation classes, `title` and the two tests all went in the same commit.

The PR description is out of date. It still describes `flex-1 min-w-0`, `block overflow-hidden text-ellipsis`, the `title` tooltip and "`gap-16` is untouched", none of which exists at head.

gbarkhatov (approved):
Follow-up review. Changes since my last review (`1a4aa205` → `b741bd90`).

**Headline:** after the squash and rebase, the code differs from what I reviewed in one place only: `xl:gap-16` became `2xl:gap-16` on both stat rows (`LiquidationEventCard.tsx:137`, `:225`). The squashed commit is signed and `validate` passes, which clears the blocker from my last review. No merge-blockers. No new defects.

### Status of prior threads

| # | Finding | Status at `b741bd90` |
|---|---|---|
| T1 | `flex-1` squeezed the columns and clipped Liq Price on mobile | Fixed (`min-w-0 grow basis-[110px]`, `:68`) |
| T2 | `title` tooltip unreachable on touch | Fixed (no `title`, no truncation, `[overflow-wrap:anywhere]`, `:72-79`) |
| T3 | Neither added test could fail if the layout were reverted | Moot: the tests are gone, so the PR touches one file |
| T4 | `gap-16` left flat; no width given for where the bug reproduces | Partly fixed. The gaps are now responsive with no desktop regression. The PR body still doesn't say at what viewport the original misalignment reproduces. |
| T5 | `xl:gap-16` brings back a two-line row at 1280px | Fixed. With the step at `2xl` (1536px), a row of about 419–546px fits three 110px stats plus two 32px gaps (394px) between 1280 and 1535px. |

### Notes on the PR description (not code)

- The "What" and "Testing" sections still describe the earlier approach (`flex-1 min-w-0`, the ellipsis and `title` tooltip, "`gap-16` is untouched") and tests that no longer exist. I raised this last time; it is carried over, not new.
- "Width bands" says the row is about 291px and wraps between 1024px and 1279px. 291px is the width at 1024px only. By the same formula (viewport − 242px sidebar − 80px page margin − 24px grid gap, halved, minus 48px card padding), the row reaches the 394px the three stats need at a 1230px viewport. So the wrap band is 1024–1229px, and 1230–1279px fits on one line.

### Verified clean

- No `data-testid`, `aria-*` or `role` was removed. The only removed lines are the `ReactNode` import, the old wrapper and span classes, and the `children` props.
- Labels still come from `COPY.liquidations.events.*`; no inline copy was added.
- No transition or animation classes and nothing positioned by a popper or tooltip were touched, so the motion rules don't apply.
- CI is green at head, including `verify` and `visual`.

jonybur (approved):
Looks good to me.

PR #2487, opened 2026-09-09
fix(core-ui): stop an open accordion clipping tooltips inside it

### LINKED ISSUE #2424: Tooltips in the Protocol Parameters accordion are clipped by the panel
## Context

The tooltip on the top fee row inside the Protocol Parameters accordion is clipped by the panel.

## Changes

**Now:** the accordion panel both clips and creates a containing block.

- `packages/babylon-core-ui/src/components/Accordion/Accordion.css:14` — `.bbn-accordion-details { overflow: hidden }`
- `packages/babylon-core-ui/src/components/Accordion/Accordion.css:21-25` — `.bbn-accordion-content[data-expanded="true"]` runs `bbn-accordion-content-in` with `animation-fill-mode: forwards`, whose final keyframe is `transform: translateY(0)`. Because of `forwards`, that non-`none` transform stays on the panel after it opens.
- `packages/babylon-core-ui/src/components/Hint/Hint.tsx:44,104,121,137,154` — the trigger sits inside that panel, the tooltip uses `positionStrategy="fixed"`, and `Hint` defaults to `placement="top"`, so it pops above its row.
- `services/vault/src/components/shared/DetailRow.tsx:41-46` (`FeeDetailRow`) wires a `Hint` on every protocol-fee row, consumed by `FeesSection.tsx`.

A `transform` on an ancestor — even `translateY(0)` — makes that element the containing block for `position: fixed` descendants. The tooltip's viewport-relative positioning is then resolved inside the transformed, `overflow: hidden` box, so the topmost row's tooltip is clipped.

`docs/motion-system.md:22-24` documents exactly this: never animate `transform` on a popper-positioned element; animate opacity on the positioned element and put translate on an inner wrapper. Here the ancestor content wrapper carries the lingering transform, so the fix belongs in the accordion — clear the transform once the animation ends, or move it to an inner wrapper.

**Expected:** the tooltip renders above the panel.
**Where:** `Accordion.css:14,21-25`

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788197440076399)

> tool tip still not visible
> See the top one hidding
> under protocol params

</details>

jeremy-babylonlabs:
Closes #2424

## What

The tooltip on the fee row inside the Protocol Parameters accordion was clipped by the panel.

`Hint` renders `react-tooltip` with `positionStrategy="fixed"`. A `position: fixed` element resolves against the viewport — unless an ancestor carries a non-`none` `transform`, which makes that ancestor the containing block instead. `.bbn-accordion-content[data-expanded="true"]` ran its enter animation with `forwards`, and the final keyframe is `transform: translateY(0)`. The fill mode held that transform on the panel permanently, so the open panel became the tooltip's containing block and `.bbn-accordion-details { overflow: hidden }` clipped it.

The fix drops `forwards`. Nothing else in the cascade sets `opacity` or `transform` on `.bbn-accordion-content`, so once the animation ends the computed style reverts to the unset base — `opacity: 1` and a literal `transform: none`. Same final appearance, no containing block.

One line of CSS. No DOM or JSX change, so no consumer of `Accordion` is structurally affected.

## Notes

- This affected **every** consumer, not just apps that opted into the motion tokens: the shift token falls back to `0px`, and `translateY(0)` is still a non-`none` transform.
- `to { transform: none }` on the keyframe was considered and rejected. CSS resolves `none` against the other side's identity before interpolating, so the frozen `forwards` value at p=1 is spec-wise `translateY(0px)` rather than the keyword — engine-dependent, not safe to rely on. Removing the fill mode is deterministic.
- `overflow: hidden` on `.bbn-accordion-details` stays. It is needed for the height animation, and once no ancestor is transformed it no longer clips a fixed-position descendant.

## Testing

jsdom neither runs keyframe animations nor computes `animation-fill-mode`, so the guard asserts the source invariant that causes the bug: the open-panel rule declares no fill mode. It rejects `both` as well as `forwards` — `both` would hold the final transform just the same.

`nx build`, `nx test`, `nx lint` on core-ui, plus vault lint and tests — all green.


jonybur (dismissed):
Verified the fix at c5134d41. Dropping the fill mode is the right call: `.bbn-accordion-content` declares only `flex w-full`, and nothing else in the file or the cascade sets `opacity` or `transform` on it, so the post-animation computed style really is `opacity: 1` / `transform: none`. That matches the keyframe's `to` block exactly, so there is no flash at the end and no residual containing block for the fixed-position tooltip. The `to { transform: none }` alternative you rejected would indeed have frozen as `translateY(0px)` under `forwards`; removing the fill mode is the deterministic form.

Keeping `overflow: hidden` on `.bbn-accordion-details` is correct — it is load-bearing for the height transition, and it only ever clipped the tooltip because the panel had been promoted to the containing block.

The CSS-source guard is the right shape given jsdom computes neither keyframes nor fill mode, and rejecting `both` alongside `forwards` closes the obvious regression. No other consumer of `Accordion` is affected, since the change is CSS-only. CI green.

jrwbabylonlab (dismissed):

jonybur (dismissed):
Right fix. `bbn-accordion-content-in` ends at `opacity: 1` and `translateY(0)`, which is exactly the panel's base style, so dropping `forwards` costs nothing visually and removes the lingering transform that was establishing a containing block for fixed-position children inside the panel. Nothing else in the file depends on that fill.

packages/babylon-core-ui/src/components/Accordion/__tests__/Accordion.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    );
+
+    fireEvent.click(screen.getByText("Vault details"));
+
+    const panel = screen.getByText("tooltip host").parentElement!;
+
+    fireEvent.animationEnd(panel);
+
+    expect(panel.matches(OPEN_PANEL_SELECTOR)).toBe(true);
+    expect(panel.style.getPropertyValue("transform")).toBe("");
```
gbarkhatov: This assertion is a tautology — it passes identically with `forwards` restored, so it cannot catch the regression it exists for.

`panel.style.getPropertyValue("transform")` reads the **inline** style map. I checked `AccordionDetails.tsx`: the only element that ever receives an inline `style` is the details *wrapper* (`height`, `visibility`, `transitionDuration`); the content div gets `ref`, `data-expanded` and `className` and nothing else. jsdom additionally loads no stylesheet, runs no keyframes, and implements no `animation-fill-mode` — so this compares `""` to `""` unconditionally.

Put `forwards` back in the CSS and lines 30–37 still go green. The only assertion that actually fails is the CSS-source regex at line 42. Suggest dropping this line and the scaffolding that exists only to feed it, or keeping the render and splitting out `it("clicking the summary marks the content panel expanded")` — which line 36 genuinely does verify.

packages/babylon-core-ui/src/components/Accordion/__tests__/Accordion.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <span>tooltip host</span>
+        </AccordionDetails>
+      </Accordion>,
+    );
+
+    fireEvent.click(screen.getByText("Vault details"));
+
+    const panel = screen.getByText("tooltip host").parentElement!;
+
+    fireEvent.animationEnd(panel);
```
gbarkhatov: `fireEvent.animationEnd(panel)` dispatches into the void — no component in the Accordion tree binds `onAnimationEnd`. I grepped `animationend|onAnimationEnd` across the Accordion directory and `packages/babylon-core-ui/src/hooks`; the only hit is `hooks/useDrawerAnimation.ts`, a different component.

The line implies a lifecycle the test never exercises, which the zero-dead-code policy rules out. It reads as meaningful partly because the idiom does exist here — `FullScreenDialog.tsx:46` really does wire `onAnimationEnd={unmount}` — which makes it easy to assume this one is load-bearing.

packages/babylon-core-ui/src/components/Accordion/__tests__/Accordion.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+const OPEN_PANEL_SELECTOR = '.bbn-accordion-content[data-expanded="true"]';
+const OPEN_PANEL_RULE = /\.bbn-accordion-content\[data-expanded="true"\]\s*\{([^}]*)\}/;
+
+const COMPILED_ACCORDION_CSS = readFileSync(path.resolve(__dirname, "../Accordion.css"), "utf-8").replace(
+  /&-/g,
+  ".bbn-accordion-",
+);
+
+describe("Accordion", () => {
+  it("an open panel keeps no transform after its enter animation, so fixed-position children are not clipped", () => {
```
gbarkhatov: The name promises a rendered-behaviour check the body doesn't perform. Nothing here observes a computed transform, an animation, or a clip — the one load-bearing assertion is a regex over CSS source text.

Per the test philosophy's "test name describes the exact behavior being tested", the honest name is closer to *"the open-panel rule declares no animation fill mode"*. This is the same defect as my comment on line 37 seen from the naming side; fixing that one fixes this.

packages/babylon-core-ui/src/components/Accordion/__tests__/Accordion.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { describe, expect, it } from "vitest";
+
+import { Accordion } from "../Accordion";
+import { AccordionDetails } from "../components/AccordionDetails";
+import { AccordionSummary } from "../components/AccordionSummary";
+
+const OPEN_PANEL_SELECTOR = '.bbn-accordion-content[data-expanded="true"]';
+const OPEN_PANEL_RULE = /\.bbn-accordion-content\[data-expanded="true"\]\s*\{([^}]*)\}/;
+
+const COMPILED_ACCORDION_CSS = readFileSync(path.resolve(__dirname, "../Accordion.css"), "utf-8").replace(
```
gbarkhatov: `COMPILED_ACCORDION_CSS` isn't compiled, and the rule-body regex is nesting-fragile.

It's raw source with a blind global `&-` → `.bbn-accordion-` substitution — no PostCSS, no Tailwind nesting resolution. That's correct today only because every `&-` in `Accordion.css` sits exactly one level under `.bbn-accordion`; a second top-level block with nested `&-` selectors would silently yield wrong selector text. Separately, `\{([^}]*)\}` stops at the first `}`, so a nested block inside that rule — say a `@media (prefers-reduced-motion)` — truncates the capture and a trailing `animation-fill-mode: forwards` slips through green.

A selector respelling does fail loudly via `not.toBeNull()`, so that path is covered. Suggest renaming to `ACCORDION_CSS_SOURCE` and asserting over the whole stylesheet that no rule matching `.bbn-accordion-content` declares `forwards`/`both`, rather than over one rule's captured body — that also closes the gap where a sibling `&-content { animation-fill-mode: forwards }` re-creates the exact bug undetected.

gbarkhatov (dismissed):
Reviewed at `fbfb5b7` via three independent passes, deduped and verified against the source.

**The one-line fix is correct, correctly diagnosed, and safe for every consumer.** A non-`none` transform establishes a containing block for `position: fixed` descendants, and `Hint` renders react-tooltip **inline** rather than portaled, with `positionStrategy="fixed"` — so the panel really is a DOM ancestor of the tooltip. `overflow: hidden` is the clipping *mechanism* but not the removable cause: without a transformed ancestor a fixed box resolves against the viewport and no intermediate `overflow` touches it. Keeping `overflow: hidden` on `.bbn-accordion-details`, which is load-bearing for the height animation, is right.

**Two things the PR body undersells.** First, this silently repairs the reduced-motion path, which was previously worst-hit: the global `*` reset in `core-ui/src/index.css` clamps `animation-duration` and `iteration-count` but never `animation-fill-mode`, so the old rule latched a permanent transform on reduced-motion users *instantly*. Second, the "every consumer was affected" claim is correct — a zero-duration animation still occurs, still fires `animationstart`/`animationend`, and with `forwards` still displays its last frame, so consumers on the `0ms` fallback were left holding `translateY(0)` too.

**A merge-blocker was raised in review and I'm refuting it here so it doesn't get re-raised.** The claim was that the deposit modal's own `.bbn-dialog-fullscreen` holds a frozen transform, so the tooltip stays clipped and #2424 isn't closed. Every repo fact behind it checks out — `tailwind.config.js` defines `modal-in` with `forwards`, `FullScreenDialog.tsx` applies it while the dialog is open, and the FeesSection → … → `FullScreenDialog` chain is real. But the conclusion doesn't follow. Per CSS 2.1 §11.1.1, a clipping region established by `overflow` does not apply to a positioned descendant whose containing block is an *ancestor* of the clipping element. Pre-fix, the tooltip's containing block is `.bbn-accordion-content`, a **child** of the `overflow: hidden` details — clipper above containing block, so it clips, which is the reported bug. Post-fix, the nearest transformed ancestor is at worst `.bbn-dialog-fullscreen`, an **ancestor** of the details — so the details no longer clip the tooltip at all. The worst case is the positioning origin shifting from the viewport to the modal's padding box, and since `.bbn-dialog-fullscreen` is `fixed inset-0`, those coincide. The fix closes #2424 either way.

**Verified clean:** no snap or flash at animation end — the `to` keyframe is pixel-identical to the base cascade, and `.bbn-accordion-content` is styled by exactly one rule repo-wide, so no app-level override can reintroduce a mismatch. No cross-consumer regression across all nine users of the component (vault `ExpandablePanel`, `FeesSection`, `UtxoSplitSelectorV3`, `VaultProviderSelectorV3`; simple-staking `StakeExpansionSection`, `FAQ/Section`, `SignModal/Step`, plus two E2E specs) — no DOM or JSX change, and no consumer passes a transform utility. Losing the panel's stacking context is benign: `Popover` renders through a `Portal`, and the Hint tooltip's `zIndex: 99999` wants to paint on top anyway.

**Worth a follow-up, not a change here:** the transform is still live for the full 180ms enter animation, so the clipping window is narrowed rather than closed — a tooltip already open when the panel expands is clipped until the animation ends. The durable form is the one `Popover.css` already documents for this hazard: animate `opacity` on the panel and move the translate to an inner wrapper. Also worth noting `docs/motion-system.md` principle 6 covers only *popper-positioned* elements; the bug class here is "don't freeze a transform on an **ancestor** of fixed-position content", which no rule currently covers. Adding that would have caught this at authoring time — and the same `forwards` + `translateY` shape still exists in `Popover.css` and `LiquidationChart.css`.

gbarkhatov (approved):
Follow-up to my review at `fbfb5b77d7`. The rebase carried both reviewed commits over unchanged, so the only delta is `f51485250`, which touches tests only.

| # | Prior finding | Verdict at `f51485250` |
|---|---|---|
| T1 | `panel.style.getPropertyValue("transform")` compared `""` with `""` | **Fixed.** The render check is gone |
| T2 | `fireEvent.animationEnd(panel)` fired an event nothing listens to | **Fixed.** Removed |
| T3 | The test name promised a render check the body never does | **Fixed.** The name now describes the stylesheet property it checks (`Accordion.test.tsx:9`) |
| T4 | `COMPILED_ACCORDION_CSS` was not compiled, and its `&-` / `[^}]*` capture was fragile | **Fixed.** It reads the real `../Accordion.css` source and scans every `animation` / `animation-fill-mode` declaration (`:6-16`) |

The check can't pass trivially: it reads the file with `fs`, not jsdom styles, and `values.length > 0` (`:13`) fails if the `animation:` declaration disappears.

Optional nit, not a blocker: the scan only sees literal `animation` declarations. A fill mode added through a Tailwind utility would get past it. For example, `@apply … animate-modal-in` in `Accordion.css` would bring back the frozen transform, because every `animate-*` in `tailwind.config.js:196-202` ends in `forwards`. A one-line `expect(ACCORDION_CSS_SOURCE).not.toMatch(/@apply[^;]*\banimate-/)` would close that.

Not a code issue: `validate` (Authenticate Commits) fails on this head. It is the commit-signature gate, so the commits need signing before merge.

Issue #2478, opened 2026-09-09
Deposit row shows progress and a contradictory signing prompt during vault-provider lag


jeremy-babylonlabs:
## Context

Follow-up to #2423, which is fixed and should ship as it is. This issue covers the residue that fix deliberately does not address.

#2423's cause: payout signing persists `PAYOUT_SIGNED` to localStorage, but the bridge that hides vault-provider (VP) lag lives in a **session-only** optimistic store that dies on reload. After a refresh, `applyTrackingOverrides` (`services/vault/src/models/peginStateMachine.ts:497-508`) treats the persisted marker as stale-or-tampered while the VP still reports `transactionsReady`, restores the full action set, and the displayed step drops from `AWAIT_VP_VERIFICATION` (11/15 = 73.3%) to `SIGN_AUTH_ANCHOR` (8/15 = 53.3%).

The fix floors the **displayed progress** only. Actions still derive entirely from the VP poll, because letting persisted local state suppress a signing request is the unsafe direction and must stay forbidden.

## The residue

During the VP lag window the deposit row now reads, simultaneously:

- progress **73%**
- status **Signing required**
- an enabled **Sign Payouts** button

The bar is right and the action is safe, but the two contradict each other on screen. `services/vault/src/components/vaults/VaultsLifecycleSections.tsx:153` and `:239`.

There is also a narrow set of states where the floor shows progress the VP does not agree with: a tampered or stale localStorage marker, a reorg, or a VP response whose write was lost. In each the action stays correct — the user is still asked to sign — so nothing unsafe happens; only the bar is optimistic.

## What is explicitly NOT wanted

Do not resolve this by trusting the persisted marker to suppress the signing request. That is the unsafe direction #2423 was careful to avoid.

## Suggested direction

Present the lag window honestly instead of hiding it: keep the action available, but frame it as a submission the vault provider has not confirmed yet, with a retry affordance, rather than as a fresh "Signing required" demand. Needs a design decision and a new string in `services/vault/src/copy.ts`.

## Acceptance criteria

- [ ] Design signs off on how the unconfirmed-submission window reads.
- [ ] A row in that window no longer shows a progress figure and a contradictory status at the same time.
- [ ] New copy lives in `services/vault/src/copy.ts` and follows that file's style rules.
- [ ] Actions still derive from the VP poll — persisted state never suppresses a signing request; a test pins this.
[...]

PR #2488, opened 2026-09-09
fix(vault): stop deposit progress rolling back after a reload

### LINKED ISSUE #2423: Deposit progress rolls back after a reload and asks for the same signature again
## Context

Reproduced from a user report: after completing a signature the progress bar reached 73%; after a page refresh it fell back to 53% and asked for a signature again. A 38-second screen recording is attached to the Slack thread.

## Changes

**Now:** progress is `getStepFillPercent(step) = max(0, visualStep - 1) / TOTAL_VISUAL_STEPS` with `TOTAL_VISUAL_STEPS = 15` (`services/vault/src/components/simple/DepositProgressView/steps.ts:1335-1339`). 73.3% is visual step 12, `AWAIT_VP_VERIFICATION`; 53.3% is visual step 9, `SIGN_AUTH_ANCHOR`. The two reported numbers land exactly on those steps, so this is the mechanism.

The live flow advances through `SIGN_AUTH_ANCHOR` (9) → `SIGN_PAYOUTS` (10) → `SIGN_DEPOSITOR_GRAPH` (11) → `AWAIT_VP_VERIFICATION` (12) as each signing step completes. None of those intermediate positions is persisted. `signAndSubmitPayouts` reports progress only through an in-memory `onProgress` callback, and writes `LocalStorageStatus.PAYOUT_SIGNED` **after the whole batch resolves** (`services/vault/src/hooks/deposit/usePayoutSigningState.ts:411-422`).

On reload the resume path calls `getPeginDisplayStep(polledPeginState)` (`services/vault/src/components/simple/ResumeDepositContent.tsx:598`), which is explicit about the fallback (`services/vault/src/models/peginStateMachine.ts:955-963`):

```ts
if (availableActions.includes(PeginAction.SIGN_PAYOUT_TRANSACTIONS)) {
  // ...next step the deposit is positioned on is SIGN_AUTH_ANCHOR, not SIGN_PAYOUTS
  return DepositFlowStep.SIGN_AUTH_ANCHOR;
}
```

So a reload landing mid-batch — before `signAndSubmitPayouts` resolves and writes `PAYOUT_SIGNED` — returns `SIGN_AUTH_ANCHOR` regardless of how many payout signatures were already collected. That is the rollback and the repeated signature request.

The branch reads both sources: `contractStatus`/`availableActions` come from a chain/VP poll, but which branch is taken depends on `localStatus`, which is `localStorage`-only and is the piece a reload can catch mid-flight.

**Expected:** progress survives a reload, and a signature already given is not requested again.
**Where:** `usePayoutSigningState.ts:411-422`, `peginStateMachine.ts:955-963`, `steps.ts:1335-1339`

## Not proven

Reproduce by reloading immediately after the auth-anchor signature but before the payout batch resolves, and confirm `getPeginDisplayStep` returns `SIGN_AUTH_ANCHOR` while payout signatures had already been submitted.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788534677955579) — screen recording attached there

> I have reproduced an issue previously reported by a user. After I completed the signature, the progress bar showed 73%. However, after refreshing the page, the progress rolled back to 53% and another signature was required.

</details>

jeremy-babylonlabs:
Closes #2423

## What

After completing a payout signature the progress bar showed 73%; after a refresh it fell back to 53% and the app re-offered the same signature.

## Cause

Payout signing persists `PAYOUT_SIGNED` to localStorage (`payoutSigning.ts:123-127`) and mirrors it into a **session-only** optimistic store (`usePayoutSigningState.ts:423`). That session store is what forces `transactionsReady` to `false` while the vault provider catches up (`computeDepositPollingResult.ts:187-191`) — and it dies on reload.

After a refresh, `applyTrackingOverrides` (`peginStateMachine.ts:497-508`) sees the persisted marker while the VP still reports `transactionsReady`, treats it as stale or tampered, and restores the full action set. `getPeginDisplayStep` then yields `SIGN_AUTH_ANCHOR` — visual step 9, 8/15 = 53.3% — instead of `AWAIT_VP_VERIFICATION`, step 12, 11/15 = 73.3%. Both reported percentages land exactly on those steps.

The lag window is real and bounded: the daemon sits in `PendingDepositorSignatures` until it has verified the submitted signatures, then moves to `PendingACKs`. A genuinely rejected signature lands in a terminal daemon error, never back in `PendingDepositorSignatures`, so this is lag rather than a legitimate re-ask.

## Fix

`getPeginProgressStep` — `getPeginDisplayStep` plus one display-only floor: with `localStatus === PAYOUT_SIGNED` and a derived step of exactly `SIGN_AUTH_ANCHOR`, report `AWAIT_VP_VERIFICATION`. Applied at the polled display sites only. The floor holds for 20 minutes after the persisted `payoutSignedAt` stamp; past that window, or with no timestamp, the progress step follows the VP's requested action, so rejected, rotated, or lost signatures cannot read as progress for the life of the localStorage entry.

Deliberately **not** changed: `applyTrackingOverrides`, and anything that decides which signatures are requested. Persisted local state must never stand in for a VP check — showing too low a step is annoying, skipping a required signature is not. The Sign Payouts CTA still appears and re-signing still works.

The floor engages on one state only. Where the VP affirmatively contradicts the marker — a WOTS demand, or never having ingested the deposit — the lower step still shows, and a test pins that.

Inference consumers keep the raw function: `pastWots` and `isDepositWarningResolved` still read `getPeginDisplayStep`, so no signing nudge or WOTS-advance inference moves.

## Known residue

During the lag window the row reads 73% with a "Signing required" status and a Sign button. The bar is right and the action is safe, but they contradict each other on screen. Resolving that properly means re-presenting the action as an unconfirmed submission, which needs design input — filed as #2478.

## Testing

Two tests, both verified by mutation rather than assertion. Replacing the helper body with a plain delegate produces exactly one failure (`expected 9 to be 12`). The WOTS-contradiction test passes under a plain delegate but fails an earlier `step < AWAIT_VP_VERIFICATION` variant that wrongly lifted the WOTS and never-ingested states too, so it pins a real boundary. A third covers the `PostDepositContinuationView` fallback path where `btcPublicKey` is unavailable. Three more pin the window: inside it, at its boundary, and with no timestamp.

`pnpm --filter vault run lint` — 0 errors. `pnpm --filter vault run test` — full suite green twice consecutively.


jonybur (dismissed):
Traced the floor and its call sites at a7d61336; the containment holds.

`getPeginProgressStep` can only fire on `contractStatus === PENDING` with `localStatus === PAYOUT_SIGNED` and `SIGN_PAYOUT_TRANSACTIONS` as the highest-priority action, since that is the only path in `getPeginDisplayStep` that yields `SIGN_AUTH_ANCHOR`. Broadcast and WOTS are both checked ahead of it, so a VP that contradicts the marker still wins — the second test pins exactly that. On `VERIFIED` the same `localStatus` yields `RETRIEVE_SECRET`, which falls through unchanged.

Checked every converted site for a non-display consumer and found none: in `PostDepositContinuationView` the value only reaches `showBtcDepthPanel` and `StatusView currentStep`, and the action branches read `peginState.availableActions` directly, so the Sign Payouts CTA is untouched; in `PendingRow` it only feeds `getStepFillPercent`; in `deriveSplitVaultProgress` it only fills the sibling lanes. Since the floor maps only `SIGN_AUTH_ANCHOR`, it cannot newly produce or suppress `AWAIT_PAYOUT_TRANSACTIONS`, so the depth panel's gate is unaffected.

Leaving the two inference consumers on the raw function is the right split, and both are genuinely inference: `ResumeDepositContent`'s `pastWots` needs the unfloored step or it would read a locally-marked deposit as past WOTS, and `depositWarnings` is resolution logic. Not touching `applyTrackingOverrides` is the correct call for the same reason.

The known residue — 73% next to a live Sign button — is honestly stated and filed as #2478. CI green.

services/vault/src/models/peginStateMachine.ts
```diff
@@ -988,6 +988,18 @@ export function getPeginDisplayStep(state: PeginState): DepositFlowStep | null {
   return null;
 }
 
+export function getPeginProgressStep(
```
jrwbabylonlab, line 998: This floor has no TTL. Both sibling suppressions in this codebase are time-bounded and say why — `REFUND_BROADCAST_SUPPRESSION_MS` (6h, "a sticky marker would otherwise hide the action") and `WOTS_SUBMISSION_SUPPRESSION_MS` (20m). A deposit whose payout signatures were rejected, rotated, or lost behind a 200 will read the floored step forever, on every reload, for the life of the localStorage entry. Gating on a fresh timestamp the same way would bound it.

jonybur (dismissed):
The TTL closes the point from the last round. `PAYOUT_SIGNED_SUPPRESSION_MS` bounds the floor the same way the refund suppression does, `payoutSignedAt` is stamped on the status write and validated on read, and the ahead-of-clock and missing-timestamp cases both fall back to the raw step rather than flooring.

The floor condition is tight: `SIGN_AUTH_ANCHOR` is only reachable from the `SIGN_PAYOUT_TRANSACTIONS` branch of `getPeginDisplayStep`, so pairing it with `localStatus === PAYOUT_SIGNED` picks out exactly the reload-lag state and nothing else. Actions and the inference consumers still read the raw step, so nothing about which signatures get requested moves.

jonybur (dismissed):
The floor only lifts `SIGN_AUTH_ANCHOR` to `AWAIT_VP_VERIFICATION`, which is forward in the flow and matches what `getWarningPeginDisplayStep` already maps `PAYOUT_SIGNED` to — so progress holds rather than jumping ahead of the deposit's real position. Anchoring it on a persisted stamp with a bounded TTL is what makes it survive a reload without becoming sticky for the life of the entry, and both the missing-stamp and ahead-of-clock cases fall through to the unfloored step.

A contradicting VP demand still wins: a `SUBMIT_WOTS_KEY` or re-broadcast action keeps its own lower step, which is the right precedence when the VP's view disagrees with local status. Storage validation for the new field rejects non-numeric and negative values, and every renamed call site moved over.

services/vault/src/storage/peginStorage.ts
```diff
@@ -198,6 +199,16 @@ function hasValidSecurityFields(entry: unknown): entry is PendingPeginRequest {
     }
   }
 
+  if (pegin.payoutSignedAt !== undefined) {
```
gbarkhatov: **MERGE BLOCKER — a malformed cosmetic timestamp discards the deposit's entire resume record, permanently.**

This makes `hasValidSecurityFields` return `false` when `payoutSignedAt` is non-numeric, non-finite or negative. I traced what that costs: `getPendingPegins` filters on exactly this validator (line 357-358), so the *whole* entry is dropped — taking `unsignedTxHex` and `selectedUTXOs` with it, which `services/vault/SECURITY_MODEL.md` invariant 8 names as "the only copy of some unrecoverable in-flight state".

And the loss is **persisted**, not transient: `updatePendingPeginStatus` (line 510) reads through `getPendingPegins`, maps, and writes the filtered array back via `savePendingPegins` — so the next status write on any *sibling* deposit erases the rejected entry from storage for good.

The `refundBroadcastAt` precedent this copies (lines 192-200) earns entry-fatal treatment because it gates an **action**. `payoutSignedAt` gates a progress-bar percentage. Stripping the offending field and keeping the entry would be free — `isPayoutSignedWithinTtl` already treats `undefined` as "no floor", so degrading to absent costs nothing but the bar.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -153,7 +153,7 @@ function PendingRow({
   const peginState = result?.peginState;
   const step =
     result && !result.loading
-      ? (result.displayStepOverride ?? getPeginDisplayStep(result.peginState))
+      ? (result.displayStepOverride ?? getPeginProgressStep(result.peginState))
```
gbarkhatov, line 156: The floored bar contradicts the "Signing required" badge and the live Sign CTA in the same row.

In `PendingRow`, `step` and `fillPercent` now come from the floored `getPeginProgressStep`, while `actionStatus` (160-162) and `subLine` (179+) still come from the unfloored `peginState`. The floor is active *precisely when* `applyTrackingOverrides` has decided the local marker is stale and kept `SIGN_PAYOUT_TRANSACTIONS` in `availableActions` — so for up to 20 minutes the row shows a 73.3% "verifying" bar directly above a live "Sign payouts" button and a "payouts ready for signing" message.

The action being offered is the safe direction, so this isn't a correctness bug. But a depositor can reasonably read the bar as authoritative and not click. Putting the decision in one place would fix it — floor the label and sub-line alongside the percent, or leave the bar unfloored while the action is live.

services/vault/src/models/peginStateMachine.ts
```diff
@@ -552,6 +564,12 @@ function isRefundBroadcastWithinTtl(
   return elapsedMs >= 0 && elapsedMs < REFUND_BROADCAST_SUPPRESSION_MS;
 }
 
+function isPayoutSignedWithinTtl(payoutSignedAt: number | undefined): boolean {
```
gbarkhatov: This TTL bypasses the injected clock and makes `getPeginProgressStep` impure.

`isPayoutSignedWithinTtl` calls `Date.now()` unconditionally and takes no `now`, unlike its immediate sibling `isRefundBroadcastWithinTtl(refundBroadcastAt, now)` at 551-565. That falsifies a contract the calling module states in prose: `computeDepositPollingResult.ts:134-143` documents `now` as overriding "*every* suppression TTL this compute touches… One injected clock must drive both, or a single call would judge the two TTLs at different times."

A third TTL now opts out. The state is computed at poll time on the injected clock while the floor is judged at render on the wall clock, so the same `PeginState` can yield different steps on two renders — which is also unsafe inside a `useMemo`. It's why the new tests need `vi.useFakeTimers` where the refund tests don't. `getPeginState` already receives both `options.payoutSignedAt` and `options.now`, so computing the boolean there and hanging it on the state — the way `activationFloorBlocksRemaining` is at 119-125 — would keep one clock.

services/vault/src/models/__tests__/peginStateMachine.test.ts
```diff
@@ -1020,6 +1021,76 @@ describe("peginStateMachine", () => {
...
+        transactionsReady: true,
+        payoutSignedAt: now + 60_000,
+      });
+      expect(getPeginProgressStep(state)).toBe(
+        DepositFlowStep.SIGN_AUTH_ANCHOR,
+      );
+      vi.useRealTimers();
+    });
+
+    it("keeps the lower step when the VP contradicts PAYOUT_SIGNED with a WOTS demand", () => {
```
gbarkhatov, line 1086: This test can't fail: delete the guard it's named for and the whole new suite stays green.

"Keeps the lower step when the VP contradicts PAYOUT_SIGNED with a WOTS demand" builds its state with **no** `payoutSignedAt`, so `isPayoutSignedWithinTtl` short-circuits to `false` and the `step === DepositFlowStep.SIGN_AUTH_ANCHOR &&` guard at line 1016 is never reached. It's a duplicate of the "carries no payout-signed timestamp" test two blocks above.

The surviving mutation isn't cosmetic: without that guard, a deposit stamped `PAYOUT_SIGNED` whose VP then demands the WOTS key inside the 20-minute window renders `AWAIT_VP_VERIFICATION` instead of `SUBMIT_WOTS_KEYS`. Adding `payoutSignedAt: now - 60_000` under fake timers would make the step-equality guard the only thing preventing the floor, which is what the test claims to check.

services/vault/src/components/simple/__tests__/PostDepositContinuationView.test.tsx
```diff
@@ -421,6 +422,22 @@ describe("PostDepositContinuationView", () => {
     expect(getByTestId("progress-view")).toBeTruthy();
   });
 
+  it("shows the floored progress step on the payout wait when btcPublicKey is unavailable", () => {
```
gbarkhatov: This test asserts the return value of its own mock. `@/models/peginStateMachine` is module-mocked at line 127 with `getPeginProgressStep: vi.fn(() => "AWAIT_BTC_CONFIRMATION")`; the test overrides that stub to return `AWAIT_VP_VERIFICATION`, then asserts the rendered step is `AWAIT_VP_VERIFICATION`.

No production floor logic executes at any point, and the `localStatus: "payout_signed"` fixture is inert because the mock ignores its argument. Deleting the entire floor body leaves this green. It also duplicates "waits (no payout) when the BTC public key is unavailable" eleven lines above.

Either delete it and cover the floor in the state-machine suite where the real function runs, or unmock the helper for this one case.

services/vault/src/storage/peginStorage.ts
```diff
@@ -500,7 +511,15 @@ export function updatePendingPeginStatus(
   const normalizedId = normalizeTransactionId(vaultId);
 
   const updatedPegins = existingPegins.map((pegin) =>
-    pegin.id === normalizedId ? { ...pegin, status } : pegin,
+    pegin.id === normalizedId
+      ? {
+          ...pegin,
+          status,
+          ...(status === LocalStorageStatus.PAYOUT_SIGNED
```
gbarkhatov: `payoutSignedAt` is never cleared when the status leaves `PAYOUT_SIGNED` — and the test that looks like it covers this passes against the unfixed code.

The spread only *adds* the stamp on the `PAYOUT_SIGNED` branch, so any other status carries a previously written stamp forward indefinitely. The test "leaves payoutSignedAt unset for a status that is not PAYOUT_SIGNED" starts from `validPegin`, which never carries a stamp — so reverting this hunk to `{ ...pegin, status }` leaves it green. It asserts state that was already true.

The stale value is inert today, since the floor also requires `localStatus === PAYOUT_SIGNED`, so this is a latent trap rather than a live bug. Writing `payoutSignedAt: undefined` on the other branch and seeding the fixture with a stamp would make the test pin the clearing.

services/vault/src/models/__tests__/peginStateMachine.test.ts
```diff
@@ -1020,6 +1021,76 @@ describe("peginStateMachine", () => {
...
+        PeginAction.SIGN_PAYOUT_TRANSACTIONS,
+      );
+      expect(getPeginDisplayStep(state)).toBe(DepositFlowStep.SIGN_AUTH_ANCHOR);
+      expect(getPeginProgressStep(state)).toBe(
+        DepositFlowStep.AWAIT_VP_VERIFICATION,
+      );
+      vi.useRealTimers();
+    });
+
+    it("releases the floor to SIGN_AUTH_ANCHOR once the payout-signed window has passed", () => {
```
gbarkhatov, line 1053: The 20-minute window is bracketed from one side only, so the constant can shrink 10× with a green suite.

The "holds" test uses 60s elapsed and this "releases" test uses *exactly* `20 * 60_000`. Set `PAYOUT_SIGNED_SUPPRESSION_MS` to two minutes and both still pass — 60s is still inside, 20 min still outside. Growing the window *is* caught; only shrinking slips through, and the window is the thing the head commit exists to document.

A case at 19 minutes asserting the floor still holds would close it.

services/vault/src/models/peginStateMachine.ts
```diff
@@ -988,6 +1006,19 @@ export function getPeginDisplayStep(state: PeginState): DepositFlowStep | null {
   return null;
 }
 
+export function getPeginProgressStep(
```
gbarkhatov, line 998: Two things about this function, one behavioural and one documentary.

**The fix converts an immediate rollback into a delayed one.** The floor is anchored on `payoutSignedAt`, not on the reload, so it expires 20 minutes after signing regardless of whether the VP has verified. A deposit whose verification outlasts that window — and nothing in the daemon semantics bounds it — drops back to step 9 on the next poll tick, un-ticking "Sign payouts" and "Sign recovery transactions" in front of the user, now with no page reload to explain it. That's #2423's symptom deferred rather than removed. The doc justifies the bound by the "rejected, rotated, or lost behind a 200" cases, but those are indistinguishable from "still verifying" — the window trades one wrong display for the other. Worth saying in the PR whether a 20-minute VP verification is considered out of spec; if it isn't, the release condition should be a VP signal rather than elapsed time.

**And it ships with no JSDoc** while `getPeginDisplayStep` directly above carries a nine-line contract. The head commit is titled "document the payout-signed floor window", but the documentation landed on the constant rather than on the function a future reader will call — and this is the function whose whole purpose is to deliberately disagree with the action set derived from the same state. The display-only invariant currently exists only in the PR description.

services/vault/src/context/deposit/computeDepositPollingResult.ts
```diff
@@ -177,6 +184,7 @@ export function computeDepositPollingResult(
     optimisticRefundBroadcastAt,
```
gbarkhatov, line 184: In-session and post-reload suppression of the same deposit disagree, and the persisted one is the shorter.

`resolvePayoutSignedAt` feeds a floor that expires at 20 minutes and only affects display, while `justSignedPayoutsThisSession` a few lines below is untimed and forces `transactionsReady = false`, removing the CTA for the life of the tab. Same deposit, same clock: a tab left open three hours shows 73.3% with no CTA; a reload at 21 minutes shows 53.3% with the CTA back.

Anchoring both on the persisted stamp would let the two paths decay together.

gbarkhatov (dismissed):
Reviewed at `d95668a` via three independent passes, deduped, with every finding re-checked against the source at HEAD.

**The floor itself is sound, and I want to lead with that because it's the thing worth getting right.** It is display-only and a genuine one-step clamp: the only transition it can produce is `SIGN_AUTH_ANCHOR → AWAIT_VP_VERIFICATION`, and only while `localStatus === PAYOUT_SIGNED`. Storage cannot name a step. The WOTS branch, the broadcast branch and the warning/danger null all win ahead of it, and it cannot reach `RETRIEVE_SECRET`/`ACTIVATE_VAULT`, which sit behind `contractStatus === VERIFIED`. Forging `status: payout_signed` plus a fresh stamp buys an attacker a progress bar and nothing signable.

**No §3 presigning concern — I verified this rather than assuming it.** `getPeginProgressStep` has exactly four production call sites (`PostDepositContinuationView.tsx:221` and `:252`, `VaultsLifecycleSections.tsx:156`, `useSplitVaultProgress.ts:90`), all rendering. `applyTrackingOverrides` is byte-identical and never sees `payoutSignedAt`; the CTA derives from `availableActions`. The two gating consumers of the *unfloored* function were correctly left alone — `depositWarnings.ts:66`, where flooring would auto-dismiss a live payout warning since 12 > 11, and `ResumeDepositContent.tsx:599`'s `pastWots` discriminator. That's the right call; my comment on the helper is that nothing in the code records why.

One merge-blocking comment below, and it isn't about the floor — it's about the validation added alongside it.

**Worth stating plainly: the PR body's "verified by mutation" claim doesn't hold for three of the added tests.** Reverting the production change leaves them green — see my comments on `peginStateMachine.test.ts:1082`, `PostDepositContinuationView.test.tsx:425`, and `peginStorage.ts:518`. The genuinely revert-catching ones are `peginStorage.test.ts:558-601` and `603-622`, `computeDepositPollingResult.test.ts:72-88`, and four of the new state-machine cases.

**Refuted:**
- *"The floor reaches presigning / can cause a payout signature to be skipped or repeated."* Killed by the call-site enumeration above. All three lanes reached this independently.
- *"Dead code from the rename."* `getPeginDisplayStep` retains two live production callers. Deliberate, not dead.
- *"Magic number."* `PAYOUT_SIGNED_SUPPRESSION_MS` is named and documented.
- *"copy.ts violation" / "dropped E2E testid."* Neither — no new user-facing string, and `pending-deposit-row` is untouched with its comment intact.

**Verified clean:** the floor introduces no silent fallback (`undefined` short-circuits to "no floor"); both `PAYOUT_SIGNED` writers reach `updatePendingPeginStatus`, so stamping inside the storage function covers every path including `PayoutSignModal`; cross-tab and multi-deposit keying is correct, with split-vault siblings each getting their own stamp and no lane able to borrow another's; and adding `payoutSignedAt` to `PeginState` perturbs no memo or identity comparison.

**Adjacent, not this PR (unpostable inline):** `depositWarnings.ts:44` now reads falsely — it says `getPeginDisplayStep` is "the same mapping the progress bar uses", and after this change the progress bar no longer uses it.

services/vault/src/models/peginStateMachine.ts
```diff
@@ -988,6 +980,35 @@ export function getPeginDisplayStep(state: PeginState): DepositFlowStep | null {
...
+/**
+ * Display-only variant of `getPeginDisplayStep` for progress bars.
+ *
+ * Right after the user signs payouts the VP can briefly keep asking for
+ * signatures again (a stale poll, or a verification still in flight). The
+ * action set must follow the VP, so the "Sign payouts" button stays live, but
+ * the progress bar deliberately disagrees: while the persisted
+ * `payoutSignedAt` stamp is within `PAYOUT_SIGNED_SUPPRESSION_MS` it holds at
+ * AWAIT_VP_VERIFICATION instead of dropping back to SIGN_AUTH_ANCHOR. The
+ * floor lifts on the next render that re-derives the step — a reload, or an
```
gbarkhatov: Low. "Not on a timer" is only true when every deposit is awaiting signatures. `usePeginPollingQuery.ts:399-406` stops `refetchInterval` only if all deposits are resolved. With a split deposit, or a second deposit still progressing, polling continues every `POLLING_INTERVAL_MS`, and the floor lifts on an ordinary poll tick. `staleTime: 0` with the default `refetchOnWindowFocus` also refetches on focus. Suggest: "lifts on the next re-derivation: a poll tick while another deposit is still polling, a focus refetch, or a reload."

gbarkhatov (dismissed):
Follow-up to my review at `d95668a388`. The rebase carried the three reviewed commits over unchanged, so the delta is `12ca3ac48` (keep a record whose payout-signed stamp is malformed) and `76f7f4b95` (shared TTL predicate). The earlier blocker is fixed.

**Merge blocker (not code):** all 5 commits are unsigned (`git log %G?` = `N`), and `main`'s ruleset has `required_signatures`. `validate` fails for that reason.

| # | Prior finding | Verdict at `76f7f4b9` |
|---|---|---|
| T1 | A malformed `payoutSignedAt` threw away the whole record | **Fixed.** `sanitizePayoutSignedAt` drops just the stamp and logs a warning (`peginStorage.ts:273-292`) |
| T2 | The floored bar contradicts "Signing required" and the live Sign button in the same row | **Unaddressed**, deferred to #2478 in the PR body. The existing thread stands |
| T3 | The TTL ignored the injected clock | **Partially fixed.** `now` is injectable, but none of the four production callers passes it. It is now documented as deliberate, so this is a design preference, not a defect |
| T4 | The WOTS-contradiction test could not fail | **Fixed** |
| T5 | `PostDepositContinuationView` test only asserted its own mock | **Fixed** (deleted) |
| T6 | `payoutSignedAt` never cleared | **Fixed** (`peginStorage.ts:525-536`), with a test that starts with a stamp |
| T7 | 20-minute window tested from one side | **Fixed** (19 and 20 minutes) |
| T8 | (a) The rollback is delayed, not removed; (b) no JSDoc | (b) **fixed**, (a) **unaddressed** |
| T9 | Suppression differs between an open tab and a reload | **Unaddressed** (`computeDepositPollingResult.ts` source not in the delta). The existing thread stands |

Two of my T3 sub-points were wrong: no caller memoizes the result, and the `computeDepositPollingResult` TTL contract is not broken.

Checked, and fine:
- A malformed stamp cannot re-prompt signing. The only production reader, `resolvePayoutSignedAt`, feeds display-step derivation alone, and a bad stamp turns the floor off, which is the safe direction.
- `isWithinTtl` reproduces all three previous predicates exactly: `elapsed >= 0 && elapsed < TTL`, in ms, with the same clock defaults and constants.
- The stamp is written only after `submitDepositorPresignatures` resolves.

The PR body is out of date. Its Testing section still describes the deleted `PostDepositContinuationView` test and doesn't mention the stamp-dropping behaviour from `12ca3ac48`.

jonybur (commented):
Notes:

- `services/vault/src/storage/peginStorage.ts:567` - **markRefundBroadcast moves the status off PAYOUT_SIGNED but carries a stale payoutSignedAt forward, unlike updatePendingPeginStatus.**
  
  This PR makes `updatePendingPeginStatus` clear `payoutSignedAt` whenever the status leaves `PAYOUT_SIGNED`, so the stamp should only ever exist next to that status. `markRefundBroadcast` also rewrites the status, to `REFUND_BROADCAST`, but spreads `...pegin` and keeps any existing stamp. A refunded deposit can therefore still hold a payout-signed timestamp in storage. It is harmless today only because `getPeginProgressStep` also checks `localStatus === PAYOUT_SIGNED`, so any future reader that trusts the stamp alone would misread the deposit. Setting `payoutSignedAt: undefined` here too, or clearing it in one shared status-write path, would keep the rule true for every writer.

gbarkhatov (approved):
Follow-up review. Changes since my last review (`76f7f4b9` → `4b4212d4`).

**Headline:** after the squash and rebase, the only substantive change is the `getPeginProgressStep` JSDoc sentence T10 asked for. It now says the floor lifts on "a reload, an unrelated re-render, or a poll tick while another deposit is still progressing — not on a timer of its own". The commit is signed and `validate` passes, which clears the blocker from my last review. CI is green, including `verify`. No merge-blockers. No new defects.

### Status of prior threads

| # | Finding | Status at `4b4212d4` |
|---|---|---|
| T1 | A malformed `payoutSignedAt` discarded the whole record | Fixed (`peginStorage.ts:273-292`) |
| T2 | The floored bar contradicts "Signing required" and the live Sign button | Deferred by the author to #2478 |
| T3 | The TTL ignores the injected clock | Partly fixed, unchanged: `now` can be injected, but no production caller passes it, and that is documented as deliberate |
| T4 | The WOTS-contradiction test could not fail | Fixed |
| T5 | The `PostDepositContinuationView` test only asserted its own mock | Fixed (deleted) |
| T6 | `payoutSignedAt` was never cleared | Fixed (`peginStorage.ts:525-536`) |
| T7 | The 20-minute window was tested from one side only | Fixed (tests at 19 and 20 minutes) |
| T8 | (a) The rollback is delayed, not removed; (b) no JSDoc | (a) Unaddressed, and the existing thread stands. (b) Fixed. |
| T9 | Suppression after a reload differs from suppression in an open tab | Unaddressed; the existing thread stands |
| T10 | "Not on a timer" is only true when every deposit is awaiting signatures | Fixed |

### Withdrawn

- **Part of my own T10 suggestion was wrong.** I proposed listing "a focus refetch" as something that lifts the floor. The app's `QueryClient` sets `refetchOnWindowFocus: false` (`config/queryClient.ts:164`), and the polling hook doesn't override it. You were right to leave that clause out.

### Notes

- **When the floor applies.** It only ever rewrites `SIGN_AUTH_ANCHOR → AWAIT_VP_VERIFICATION`, and only while all of the following hold:
  - the local status is `PAYOUT_SIGNED`;
  - the contract is `PENDING`;
  - the stamp is under 20 minutes old.
- **What it never touches.** Expired or invalid vaults return before the floor runs. Actions are never floored.
- **After the window.** A tab with a single deposit can keep the old bar until something re-derives the step, because polling stops and focus refetch is off. That is the T8(a) behaviour, now documented accurately.
- **PR description.** The Testing section still describes the deleted `PostDepositContinuationView` test. Carried over from last time.

jonybur (approved):
Looks good to me.

jonybur (commented):
Comments inline.

services/vault/src/components/simple/PostDepositContinuationView.tsx
```diff
@@ -218,7 +218,7 @@ export function PostDepositContinuationView({
   );
   const peginState = pollingResult?.peginState;
   const waitStep = peginState
```
jonybur, line 220: **Without a connected Bitcoin key, the continuation view shows a "verifying deposit" status while the provider still asks for payout signatures**

`waitStep` now comes from `getPeginProgressStep`, but this site is not only a progress bar. When `SIGN_PAYOUT_TRANSACTIONS` is in the action set and `btcPublicKey` is set, the view renders `ResumeSignContent` and never shows `waitStep`. So the floor only reaches the screen when `btcPublicKey` is undefined, for example after a reload with the Bitcoin wallet not yet connected. The view then falls through to `StatusView` with `currentStep={waitStep}`, `isProcessing` and `canContinueInBackground`. With the floor active that step is `AWAIT_VP_VERIFICATION`, so `DepositProgressView` renders `ProviderWaitDetail`, which shows `waitDetails.verifyingDeposit` with a spinner. Before this change the same case rendered `SIGN_AUTH_ANCHOR` with no provider-wait panel. For up to 20 minutes, a depositor who still owes signatures is told the provider is verifying and that they can continue in the background. This is a new status message, not only a higher bar, and #2478 does not cover it. Fix: keep `getPeginDisplayStep` for `waitStep` here (the split-lane `perVaultSteps` can stay floored), or skip the floor when the sign branch cannot render.

PR #2489, opened 2026-09-09
feat(vault): let a browser-local pending deposit be dismissed

### LINKED ISSUE #2422: Pending deposits with no on-chain match are never pruned and cannot be dismissed
## Context

A user had two pending deposits more than a month old that could not be finalised or removed. The workaround given was to clear cookies and cache — which is, per the code, the only way today.

## Changes

**Now:** pending deposits live in `localStorage` under `vault-pending-pegins-<ethAddress>` (`services/vault/src/constants.ts:10`, `STORAGE_KEY_PREFIX`), holding an array of `PendingPeginRequest` — id, timestamp, unsigned tx hex, selected UTXOs, build-version pins and a status (`services/vault/src/storage/peginStorage.ts:29-88`). Written by `addPendingPegin` (`:451-486`) and `updatePendingPeginStatus`; read by `getPendingPegins` (`:302-353`) on every render via `usePeginStorage.ts`, then merged with confirmed on-chain activity in `useVaultDeposits.ts` and `pendingActivities.ts`.

Pruning happens in `filterPendingPegins` (end of `peginStorage.ts`) and only when the entry matches a **confirmed on-chain vault**:

```ts
const confirmedPegin = normalizedConfirmedPegins.find(p => p.id === normalizedPeginId);
if (!confirmedPegin) return true;   // keep — no chain match
return !shouldRemoveFromLocalStorage(confirmedPegin.status, pegin.status, pegin.refundBroadcastAt);
```

`shouldRemoveFromLocalStorage` (`services/vault/src/models/peginStateMachine.ts:1080-1112`) drops an entry only once the contract status reaches `ACTIVE`/`REDEEMED`/`LIQUIDATED`/`INVALID`/`DEPOSITOR_WITHDRAWN`/`EXPIRED`, or when `localStatus === PENDING && contractStatus === VERIFIED`.

**Two gaps:**

1. Nothing checks `pegin.timestamp`. If a deposit never reaches a matching on-chain vault — the Pre-PegIn never confirmed, or the user abandoned it before it indexed — `confirmedPegin` stays `undefined` and the entry is kept forever. No TTL exists anywhere in `peginStorage.ts` or `peginStateMachine.ts`.
2. There is no user-facing dismiss. `removePendingPegin` exists (`peginStorage.ts:509-517`) and is wired through `useDepositFlow.ts:1018` and `useVaultActions.ts:449`, but only internal flow logic calls it. `PendingDepositCard` and `VaultsLifecycleSections.tsx` render pending cards with no delete action.

**Expected:** a stale entry can be removed without clearing browser storage — an age-based prune, a dismiss action, or both.
**Where:** `peginStorage.ts` (`filterPendingPegins`), `peginStateMachine.ts:1080-1112`, and the pending card in `VaultsLifecycleSections.tsx`

## Not proven

Confirm at runtime that the specific stale entries have no matching on-chain vault — open devtools and read `localStorage['vault-pending-pegins-<addr>']` — to rule out a data-corruption variant. The missing TTL and the missing dismiss are both visible in source.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788250659166559)

> I also have some pending deposits:
> 1st one is the current one
> 2nd and 3rd are more than a month old - how can I get rid of them? I am not able to finalize them.

</details>

jeremy-babylonlabs:
Closes #2422

## What

A user had two pending deposits over a month old that could neither be finalised nor removed. The only workaround anyone could offer was "clear cookies and cache" — and per the code that was genuinely the only way.

Adds a user-initiated dismiss with a confirmation, gated so it can only ever remove a record this browser holds alone.

## Why no automatic pruning

The stored record holds the unsigned Pre-Pegin transaction and the selected UTXOs — the user's resume material. Deleting it on a heuristic destroys recoverable work, so automatic pruning was only ever acceptable on a provably dead condition.

There isn't one reachable without new per-row network calls. `assertUtxosAvailable` looks like a dead signal but is not: a **successfully broadcast** Pre-Pegin spends exactly those inputs, so their absence is also the success signature. Pruning on it would delete live deposits. The two genuinely provable signals — `getOutspend` per selected UTXO, or `depositor === zeroAddress` past ETH finality — both need per-row requests that do not exist today.

So: dismiss only. No age-based sweep.

## The gate

A pending deposit is offered a Remove control only on positive evidence that this browser holds the only record of it. `fetchVaultsByDepositor` returns `{ vaults, droppedCount }`: it drops any indexed row `transformVaultItem` rejects, so a resolved query is not by itself a complete list. The gate therefore reads `indexedVaultIds`, which is `null` — no control offered — unless the query succeeded **and** `droppedCount === 0`. A failing, still-loading, or row-dropping indexer offers nothing; absence is only evidence when the answer was complete.

The gate is re-checked against live data immediately before the write, so polling learning the deposit is indexed while the dialog sits open aborts the dismiss. Ids are compared case-insensitively on both sides.

This matters because an indexed deposit's record carries the participant-key stamp that blocks an unsafe later broadcast.

## Split deposits

The records of a split deposit share one funded Pre-PegIn transaction. The control is offered only when the record **and** every `getBatchSiblings` sibling are absent from that complete set, and confirming removes the whole batch in a single write. The confirmation says how many records it will remove. Removing one alone would leave a sibling able to broadcast a transaction funding a vault whose record and tx hex are gone.

## Other changes

- `removePendingPegin` now works off the raw stored array, so legacy or malformed sibling records the read filter hides are no longer silently dropped when one entry is removed. Malformed JSON causes no write at all; an empty result deletes the key rather than writing `[]`.
- `removePendingPegins` returns `false` when localStorage cannot be read or the write fails, and the entries are still stored. On `false` the confirmation stays open with a `role="alert"` message instead of closing, so a blocked browser reports what happened rather than silently repeating the same no-op on every retry.
- The action cell is only occupied when the control renders, so untouched Pending rows keep their alignment.

## Copy

The confirmation says plainly that the funds are unaffected — removing a local record does not move or spend any Bitcoin. What is lost is the unsigned transaction this browser holds, so it can no longer be broadcast from here.

## Known residue

If the indexer later starts returning a vault whose local record was dismissed, the row comes back from indexer data without the build-version and participant-key stamps, and resume skips those two checks. That is the cross-device resume path the code already documents; the gate makes it reachable only when the indexer positively did not have the vault at dismiss time.

## Testing

12 tests. Every one verified by mutation: dropping the confirm-time recheck, treating a null indexer as answered, or making either id comparison case-sensitive each fails a specific named test. Reverting the feature fails 11.

`pnpm --filter vault run lint` — 0 errors. `pnpm --filter vault run test` — full suite green. `tsc --noEmit` clean.


jonybur (dismissed):
Reviewed at 575b6257. The gate and the storage rewrite both hold up under tracing.

The safety argument is the load-bearing part and it is right. `indexedVaultIds` is `null` unless `status === "success"`, and TanStack Query v5 moves a query with existing data to `status: "error"` when a background refetch fails — so an indexer outage hides the control rather than making every indexed deposit look browser-local. The confirm-time recheck through the same `canDismiss` closes the dialog-open window, and both id comparisons lowercase on both sides. `realActivityIds` coming from `allActivities` also keeps demo rows out.

Your reasoning for dismiss-only rather than automatic pruning is correct on the code: a successfully broadcast Pre-PegIn spends exactly the selected UTXOs, so `assertUtxosAvailable` failing is the success signature as much as the dead one. Pruning on it would delete live deposits.

The storage rewrite is a real improvement beyond this feature. Operating on the raw array means legacy records without the build-version stamps — which `getPendingPegins` hides but which are still the user's only copy — survive a sibling removal instead of being written away with it, and that also fixes the existing broadcast-path caller. `readStoredEntryId` gates on the same `BYTES32_HEX_RE` the read path validates with, so an entry that renders can always be matched; nothing can render as dismissible and then silently fail to match. Returning `true` on a missing key or an absent target is the right idempotent answer, and returning `false` only on a genuine write failure is what lets the dialog stay open with the `role="alert"` message instead of closing on a no-op.

Splitting `persistStoredEntries` out is clean: `savePendingPegins` keeps its best-effort swallow for the cosmetic callers, `addPendingPegin` still throws, and `normalizeTransactionId` cannot throw, so moving it above the try changes nothing. Empty result deleting the key rather than writing `[]` is right.

All new strings are in `copy.ts`, the new testid is new rather than a moved E2E hook, and the known residue about a re-appearing indexed row is honestly scoped. CI green.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -670,6 +693,52 @@ export function VaultsLifecycleSections({
...
+
+  const handleDismiss = useCallback((depositId: string) => {
+    setDismissFailed(false);
+    setDismissingId(depositId);
+  }, []);
+  const handleDismissCancel = useCallback(() => {
+    setDismissFailed(false);
+    setDismissingId(null);
+  }, []);
+  const handleDismissConfirm = useCallback(() => {
```
jrwbabylonlab, line 770: Split deposits write one record per vault, all sharing the same `fundedPrePeginTxHex`. Dismiss row A, then hit Broadcast on row B: the shared tx funds both vaults on Bitcoin, but A no longer has a record, a txHex, or a resume path. The details and broadcast paths expand through `getBatchSiblings` (~line 601); this removes `dismissingId` alone. Either expand to siblings here, or block dismiss while any sibling record remains.

services/vault/src/storage/peginStorage.ts
```diff
@@ -507,13 +549,38 @@ export function updatePendingPeginStatus(
...
-export function removePendingPegin(ethAddress: string, vaultId: Hex): void {
-  const existingPegins = getPendingPegins(ethAddress);
-  const normalizedId = normalizeTransactionId(vaultId);
-  const filtered = existingPegins.filter((p) => p.id !== normalizedId);
-  savePendingPegins(ethAddress, filtered);
+export function removePendingPegin(ethAddress: string, vaultId: Hex): boolean {
+  if (!ethAddress) return false;
+
+  const stored = readStoredEntries(ethAddress);
+  if (stored === null) return true;
```
jrwbabylonlab: This returns `true` — and so does the not-found path on line 574 — without dispatching a storage-update event. If localStorage is unreadable after the row was loaded (private mode, blocked site data), the caller reports success, the dialog closes, and the row is still there because nothing re-read. Every retry repeats it silently. That's the loop this PR is meant to end. A `false` here, or a discriminated result, would let the dialog say what actually happened.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -670,6 +693,52 @@ export function VaultsLifecycleSections({
...
+   * vault. A failing, empty-because-erroring, or still-loading indexer leaves
+   * `indexedVaultIds` null and offers nothing — an indexed deposit's record
+   * carries the participant-key stamp that blocks an unsafe later broadcast,
+   * and must never be discarded on the mere absence of a row.
+   */
+  const canDismiss = useCallback(
+    (activity: VaultActivity) => {
+      if (activity.isPending !== true || indexedVaultIds === null) return false;
+      if (!realActivityIds.has(activity.id)) return false;
+      return !indexedVaultIds.has(activity.id.toLowerCase());
```
jrwbabylonlab: The positive-evidence premise doesn't quite hold: `fetchVaultsByDepositor` catches `transformVaultItem` failures per item, logs, and returns the rest (`fetchVaults.ts:458-469`), so the query still resolves `success` with that vault missing. An indexed vault dropped that way is absent from `indexedVaultIds`, and this returns true for it — offering to delete a record that carries the participant-key stamp.

jonybur (dismissed):
Reviewed the incremental changes at head. The dismiss gate still requires the indexer to have a successful answer with the target absent. The id-reader gates on the same hex regex as the read path so a rendered dismissible entry can always match. Empty-result deletion clears the key rather than persisting an empty array. The batch-sibling read at confirm time closes the dialog-open window. All new copy strings sit in the central copy file, and no test id was dropped. CI green.

jonybur (dismissed):
The dismiss gate is fail-closed in the right places. `indexedVaultIds` stays null unless the query genuinely succeeded, and threading `droppedCount` out of `fetchVaultsByDepositor` closes the case that mattered most — a row the transform rejected is otherwise indistinguishable from a vault the indexer never held, and reading that absence as evidence is exactly what would discard a record that still carries the participant-key stamp.

Batch handling is all-or-nothing, and `removePendingPegins` operating on the raw stored array means siblings the read filter hides (legacy records without build-version stamps) are written back rather than dropped along with the targets. The boolean return is threaded all the way to the dialog, so a blocked localStorage write is reported instead of reading as a removal.

Every `useVaults` consumer is updated for the new result shape.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -670,6 +693,70 @@ export function VaultsLifecycleSections({
...
+
+  /**
+   * A deposit may be discarded only on positive evidence that it is the
+   * browser's alone: the indexer answered completely and did not return this
+   * vault. A failing, still-loading, or row-dropping indexer leaves
+   * `indexedVaultIds` null and offers nothing — an indexed deposit's record
+   * carries the participant-key stamp that blocks an unsafe later broadcast,
+   * and must never be discarded on the mere absence of a row.
+   */
+  const canDismissRecord = useCallback(
```
gbarkhatov, line 732: **MERGE BLOCKER — the gate never reads the record's own status, so an already-broadcast Pre-PegIn is dismissible.**

`canDismissRecord` tests exactly three things: `activity.isPending !== true`, membership in `realActivityIds`, and `!indexedVaultIds.has(id)`. It never reads `pending.status` — and `isPending` is *not* an independent signal, because `usePeginStorage.ts:200-232` hardcodes `contractStatus: PENDING` and `isPending: true` for every local record the indexer didn't match, mapping `offchainParamsVersion` across but not `status`. So the gate counts one indexer answer twice.

I traced the write ordering in `useDepositFlow.ts`: `registerPeginBatchAndWait` (:834) → `addPendingPegin` (:953) → `broadcastPrePeginTransaction` (:1060) → `LocalStorageStatus.CONFIRMING` (:1110). A record at `CONFIRMING` — funding UTXOs already spent into the HTLC — is byte-identical to a never-broadcast one as far as this gate can see. Concrete failure: user broadcasts, indexer lags 30s, user clicks ×, and `unsignedTxHex` plus `selectedUTXOs` for a transaction with already-spent inputs are erased.

`PendingRow` already holds the ground truth via `useDepositPollingResult(activity.id).prePeginBroadcastSeen`, which `getPeginState` uses to strip the broadcast action. Requiring `prePeginBroadcastSeen === false` **and** `status === LocalStorageStatus.PENDING` would make the gate say what it means.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -670,6 +693,70 @@ export function VaultsLifecycleSections({
...
+
+  const handleDismiss = useCallback((depositId: string) => {
+    setDismissFailed(false);
+    setDismissingId(depositId);
+  }, []);
+  const handleDismissCancel = useCallback(() => {
+    setDismissFailed(false);
+    setDismissingId(null);
+  }, []);
+  const handleDismissConfirm = useCallback(() => {
```
gbarkhatov, line 770: **MERGE BLOCKER — dismissal destroys the RFC-006 participant-key stamp, converting a fail-closed resume into a fail-open one.**

This is the sharpest version of the "absence is not non-existence" problem, because the codebase already documents this exact failure. I read `useVaultActions.ts:450-470`, and its comment describes it verbatim: dropping the record *"would discard the stamp, and the next attempt would find no local copy, fall back to the indexer's transaction, pass the `prePeginTxHash` check — it is the registered transaction — and broadcast the very Pre-PegIn this just refused."*

The guard is `if (pendingPegin?.buildParticipantOperationKeys)`, and the four build-version stamps are guarded the same way — all of them skip silently when `pendingPegin` is absent. Dismissal performs precisely that discard, authorised by an indexer signal. The stamps exist nowhere but localStorage.

Either refuse dismissal for any record carrying `buildParticipantOperationKeys`, or require a positive chain read first. `useStuckVaultChainConfirm.ts` is the reusable pattern: `getVaultBasicInfo(id)`, treat `depositor === zeroAddress` as genuinely unregistered, and fail closed on RPC error.

services/vault/src/copy.ts
```diff
@@ -1982,6 +1982,22 @@ export const COPY = {
       withdraw: "Withdraw",
       viewDetails: "View Details",
     },
+    dismissPending: {
+      rowLabel: "Remove this pending deposit",
+      title: "Remove this pending deposit?",
+      body: "This deposit's record is saved in this browser only. Removing it does not move or spend any Bitcoin, and your funds stay in your wallet.",
```
gbarkhatov, line 2072: **MERGE BLOCKER — this asserts fund safety the gate does not establish.**

Verbatim: *"This deposit's record is saved in this browser only. Removing it does not move or spend any Bitcoin, and your funds stay in your wallet."* Given the gate problem on `VaultsLifecycleSections.tsx:709`, that is outright false for a `CONFIRMING` record — the Bitcoin has moved, and the dialog tells the user it hasn't.

Even with that gate fixed, it remains a claim about *chain state* derived from an untrusted indexer's silence. The `warning` string is honest about irreversibility ("This cannot be undone"), but neither string mentions the selected UTXO set or the participant-key stamp — the two things that actually matter here.

Worth noting `copy.ts` already hoists this reassurance into one constant, `NOTHING_SIGNED_OR_SPENT`, precisely so the wording can't drift — and this reinvents a *stronger* claim inline. Either soften it to what is actually known, or gate on status so the claim becomes true.

services/vault/src/hooks/useVaultDeposits.ts
```diff
@@ -52,9 +55,23 @@ export function useVaultDeposits(connectedAddress: Address | undefined) {
...
     if (!data) return [];
 
-    return data.map(transformVaultToActivity);
+    return data.vaults.map(transformVaultToActivity);
   }, [data]);
 
+  /**
+   * Lowercased ids of every vault the indexer returned, or null unless that
+   * set is known to be complete — a failed or in-flight query is never
+   * evidence that a vault is absent, and neither is a successful one that
```
gbarkhatov, line 64: `indexedVaultIds` is non-null on stale cache from a failing indexer, so absence isn't the evidence the doc comment claims.

The gate is `status === "success" && data?.droppedCount === 0`. In TanStack Query, `status` stays `"success"` while cached data exists — including when every background refetch since has thrown, which surfaces on `error`, destructured at :38 and unused here. With `useVaults` on a 60s `refetchInterval` and 5min `gcTime`, an indexer down for minutes presents a stale vault list as a complete successful answer, and the × appears on every vault registered since it broke.

The doc comment says "a failed or in-flight query is never evidence that a vault is absent" — but an in-flight *background* refetch over stale cache is exactly that, and isn't excluded. That's invariant 10: uncertainty resolving to a default rather than blocking the action. Adding `error === null` plus a freshness term (`isFetching === false`, or a `dataUpdatedAt` bound), and calling `refetch()` on dialog open rather than reading cache, would close it.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -753,6 +842,14 @@ export function VaultsLifecycleSections({
         </section>
       )}
 
+      <DismissPendingDepositDialog
```
gbarkhatov, line 905: Two related problems from deriving the open dialog's content from a live batch.

**A confirm the gate has since invalidated closes silently and removes nothing.** `dismissBatch` recomputes on every render, so when a poll lands mid-dialog and the vault reappears it collapses to `[]`; `handleDismissConfirm` then hits the early return, clearing state and closing with no message while the row stays. That's precisely the abort this design is proud of, but the user's only available reading is "it worked". The storage layer repeats it one level down — `removePendingPegins` returns `true` for both the `empty` and nothing-matched branches — and the new test asserts only that `removePendingPegins` was *not* called, pinning the silent path as correct.

**And the copy mutates under the user.** `count={dismissBatch.length}` with `isBatch = count > 1`, so if the gate flips while they're reading, the dialog changes in place from "Remove these 2 pending deposits?" to the singular title, body and warning — describing a different, smaller action than the one on screen a moment ago, and one that will in fact do nothing.

Freezing the batch into state when `handleDismiss` opens the dialog, rendering from that snapshot, and using live `canDismiss` only to permit the confirm would fix both. The existing `role="alert"` slot in `DismissPendingDepositDialog.tsx` is the natural place for an aborted-state message.

services/vault/src/copy.ts
```diff
@@ -1982,6 +1982,22 @@ export const COPY = {
...
+      warning:
+        "You will lose the unsigned Pre-Pegin transaction this browser holds, so you can no longer broadcast it from here. This cannot be undone.",
+      batchTitle: (count: number) => `Remove these ${count} pending deposits?`,
+      batchBody: (count: number) =>
+        `These ${count} deposits share one Pre-Pegin transaction, so they are removed together. Their records are saved in this browser only. Removing them does not move or spend any Bitcoin, and your funds stay in your wallet.`,
+      batchWarning: (count: number) =>
+        `You will lose the unsigned Pre-Pegin transaction this browser holds for all ${count} deposits, so you can no longer broadcast it from here. This cannot be undone.`,
+      failed:
```
gbarkhatov: This misattributes a reachable failure and gives advice that cannot work. *"Your browser blocked the change… Check your privacy settings or free up local storage, then try again."*

But `removePendingPegins` also returns `false` on the `readStoredEntries` → `"unreadable"` branch (`peginStorage.ts:600`) — corrupt or foreign JSON under the key, not a browser block. In that branch the diagnosis is wrong *and* retrying can never succeed, so the user loops on Remove indefinitely with no path forward.

Distinguishing the two `false` causes at the seam and adding a second string would let each one say something true.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -670,6 +693,70 @@ export function VaultsLifecycleSections({
...
+    [indexedVaultIds, realActivityIds],
+  );
+
+  /**
+   * The records of a split deposit share one funded Pre-PegIn transaction, so
+   * discarding one alone would leave a sibling able to broadcast a transaction
+   * that funds a vault whose record — and its Pre-PegIn hex — is gone. The
+   * whole batch is therefore discardable together or not at all.
+   */
+  const canDismiss = useCallback(
```
gbarkhatov, line 750: The batch invariant is silently unenforced for exactly the records it was written about.

`batchKey` maps `unsignedPrePeginTx === ""` — the explicit cross-device "no local data" marker — to `standalone:<id>`, so `getBatchSiblings` returns a single element and each record of a split deposit resumed on a second device becomes dismissible alone. The stated "removed together or not at all" invariant doesn't hold there, and a mixed batch drops the hex while leaving a sibling row behind.

Grouping by the stored `batchId`/`batchIndex`/`batchTotal` rather than by tx-hex identity would hold in both cases.

services/vault/src/hooks/__tests__/useVaultDeposits.test.ts
```diff
@@ -78,6 +79,34 @@ describe("useVaultDeposits", () => {
     expect(setIntervalSpy).not.toHaveBeenCalled();
   });
 
+  it("exposes the indexed vault ids when the indexer returned every row", () => {
```
gbarkhatov, line 85: The id-lowercasing the gate depends on is untested. Both new tests pass `vaults: []`, so `new Set(confirmedActivities.map((a) => a.id.toLowerCase()))` never sees a mixed-case id — deleting `.toLowerCase()` keeps the suite green.

That undercuts the PR body's "mutation-verified" claim on the one line where a case mismatch would offer dismissal of a *live* deposit. One test with an id like `"0xABCDEF…"` would pin it.

services/vault/src/storage/__tests__/peginStorage.test.ts
```diff
@@ -591,6 +592,136 @@ describe("removePendingPegin", () => {
...
+      expect(setItemSpy).toHaveBeenCalledTimes(1);
+      expect(JSON.parse(localStorage.getItem(storageKey) as string)).toEqual([
+        survivor,
+      ]);
+    } finally {
+      setItemSpy.mockRestore();
+    }
+  });
+
+  it("notifies listeners after a real removal so the row re-reads", () => {
```
gbarkhatov: The head-commit test proves the same-tab custom event, not the cross-tab storage event it's named for.

It listens on `STORAGE_UPDATE_EVENT`, an in-page `CustomEvent` that by definition never crosses tabs. The actual cross-tab path is the native `storage` listener at `usePeginStorage.ts:144-158` — pre-existing and outside this diff — which nothing in the suite exercises.

The production behaviour is correct (I traced it, and `removeItem` propagates fine because the listener keys on `e.key` alone); it's the commit subject that overstates what's pinned. Either rename to "same-tab", or add a `usePeginStorage` test dispatching a synthetic `StorageEvent`.

services/vault/src/hooks/useVaultCountCap.ts
```diff
@@ -80,7 +80,8 @@ export function useVaultCountCap(
     refetchOnWindowFocus: false,
   });
 
-  const { data: vaults, isError: vaultsError } = useVaults(address);
+  const { data: vaultsResult, isError: vaultsError } = useVaults(address);
+  const vaults = vaultsResult?.vaults;
```
gbarkhatov, line 84: This consumer reads `.vaults` and discards `droppedCount`, which the PR just introduced as a completeness signal.

Dropped rows under-count `currentCount`, so an at-cap user passes the gate and reverts at activation. The hook already fails closed on `vaultsError`, so folding `droppedCount > 0` into that same unavailable state would be consistent with how it already handles uncertainty.

gbarkhatov (dismissed):
Reviewed at `7f0ed85` via three independent passes (7 agents), deduped, with every anchor re-checked against the code.

**First, the reassuring part, because all three lanes traced it independently and agreed: dismissal does not strand the user on funds.** The indexer serves `unsignedPrePeginTx` and `depositorBtcPubKey` in `VAULT_FIELDS`, `handleBroadcast` takes `pendingPegin` as optional and falls back to it, integrity is anchored on-chain by comparing `calculateBtcTxHash(unsignedTxHex)` against `prePeginTxHash`, UTXOs resolve from the mempool, and vault secrets stay deterministically derivable from the funding outpoints of that same indexed transaction. What is irrecoverably lost is the **safety layer**, not the recovery material. The user isn't stranded; they're un-guarded. That distinction is what the three blocking comments below are about, and it's the finding it would have been worst to get wrong in either direction.

**The structural problem is that the gate is proof of *indexer silence*, and cannot be more.** The record is written only after a mined Ethereum registration, so every dismissible record corresponds to a vault that **is** registered. Absence from the indexer is therefore always ingest lag, staleness, or a reorg window — never non-existence. This inverts the precedent the reclaim work established in this repo: gate on Bitcoin state, never on an Ethereum/indexer signal, for an irreversible action.

**Cross-tab behaviour is clean** and I'd rather say so explicitly than leave it implied. Same-tab goes through `dispatchStorageUpdateEvent`; cross-tab, the write goes through `setItem`/`removeItem` and the native listener keys only on `e.key` rather than `newValue`, so the `removeItem` case bumps `storageVersion` in tab B identically. `getStorageKey` builds the key byte-identically to the listener's comparison, with no casing skew. Tab B's row disappears rather than going stale, and can't be acted on. The only defect there is test coverage — see my comment on `peginStorage.test.ts`.

- *"`persistStoredEntries` swallows the write error."* It genuinely rethrows at `peginStorage.ts:432-438`, so `removePendingPegins`' catch correctly returns `false`.
- *"Cross-tab propagation is broken / tab B keeps an actionable stale row."* No — the native listener keys on `e.key` alone, so `removeItem` propagates fine.
- *"Dismissal destroys unrecoverable recovery material / strands the user."* No, per the first paragraph. The loss is the guard.
- *"A wrong record could be deleted, or a target could survive."* No — `readStoredEntryId` shares `BYTES32_HEX_RE` with `hasValidSecurityFields`, so any entry that can render a row can be targeted.

**Verified clean:** `removePendingPegins` operating on the raw stored array is right — legacy and mangled siblings survive, `unreadable` never authorises a write, and an empty result deletes the key rather than writing `[]`. Case handling is consistent end to end. The `{ vaults, droppedCount }` plumbing is a faithful mechanical unwrap at all four `useVaults` call sites. The failed-write path is honest throughout: throws, returns `false`, dialog stays open with `role="alert"`, no "removed" claim on a failed removal. `copy.ts` policy is satisfied — `batchTitle`/`batchBody`/`batchWarning` are functions in `copy.ts` with nothing user-facing inlined.

**Adjacent, not this PR (unpostable inline):** `fetchVaults.ts:446`'s pagination guard `while (hasNextPage && endCursor != null)` exits silently when `hasNextPage` is true but `endCursor` is null, returning a page-1 prefix with `droppedCount === 0` — the `MAX_VAULT_PAGES` fail-closed throw is *inside* the loop and doesn't cover it. Pre-existing, but this PR newly depends on it. The sibling `fetchVaultIdsByDepositor` already solves it by projecting `totalCount` and throwing on mismatch.

jonybur (commented):
One finding on the second delta commit's copy branching.

`VaultsLifecycleSections.tsx` around lines 760-774 selects `noLongerRemovable` — "its Pre-Pegin transaction has since been broadcast or found on-chain" — whenever a batch member's `localRecordStatuses.get(id)` is not `PENDING`. When another tab dismisses only a subset of the batch, the dismissed sibling is no longer in `realActivityIds` and its local status is `undefined`, so `undefined !== PENDING` fires this branch and shows the "broadcast or found on-chain" copy. But the sibling was dismissed elsewhere, not broadcast. The user is told the wrong story.

The first "all vanished" branch above does not cover this because it requires *every* batch member gone. A partial-vanish should either close silently (with the remaining siblings intact) or report "changed elsewhere". Not a correctness bug — no unsafe write occurs — but the copy is factually wrong on this narrow path, and the delta commit's own scope ("close or explain a dismiss whose record changed in another tab") is exactly that path.

services/vault/src/storage/peginStorage.ts
```diff
@@ -507,13 +565,78 @@ export function updatePendingPeginStatus(
...
+  if (read.status === "unreadable") return "unreadable";
+  if (read.status === "empty") {
+    dispatchStorageUpdateEvent(ethAddress);
+    return "removed";
+  }
+
+  const targets = new Set(
+    vaultIds.map((vaultId) => normalizeTransactionId(vaultId).toLowerCase()),
+  );
+  const remaining = read.entries.filter((entry) => {
```
gbarkhatov, line 691: Should fix. The removal filters on id alone, so the status re-check in `handleDismissConfirm` (`VaultsLifecycleSections.tsx:766-775`) is a check-then-delete. That check reads `localRecordStatuses`, which is React state. When another tab writes CONFIRMING, it takes a `storage` event, a `storageVersion` bump and an effect before that state changes. A Confirm click in that gap deletes a record that has just been broadcast, and the other tab's later `updatePendingPeginStatus` then finds nothing to update.

Fix: pass the expected status into `removePendingPegins`, and inside this same raw-array read return e.g. `"changed"` (mapped to `noLongerRemovable`) without writing if any target entry's `status !== PENDING`. The check and the delete then share one synchronous read-modify-write.

services/vault/src/copy.ts
```diff
@@ -2004,6 +2004,28 @@ export const COPY = {
...
+        "You will lose the unsigned Pre-Pegin transaction this browser holds, so you can no longer broadcast it from here. This cannot be undone.",
+      batchTitle: (count: number) => `Remove these ${count} pending deposits?`,
+      batchBody: (count: number) =>
+        `These ${count} deposits share one Pre-Pegin transaction, so they are removed together. Their records are saved in this browser only. Removing them does not move or spend any Bitcoin, and your funds stay in your wallet.`,
+      batchWarning: (count: number) =>
+        `You will lose the unsigned Pre-Pegin transaction this browser holds for all ${count} deposits, so you can no longer broadcast it from here. This cannot be undone.`,
+      writeFailed:
+        "Your browser blocked the change, so the deposit is still saved here. Check your privacy settings or free up local storage, then try again.",
+      unreadable:
+        "This browser's saved deposit records could not be read, so nothing was changed. Clearing this site's stored data removes them.",
```
gbarkhatov: "Clearing this site's stored data removes them" tells the user to wipe every pending-deposit record for every address, including readable ones that still carry participant-key stamps. That is the loss this PR's gate exists to prevent, done in bulk. It also doesn't help in the case the test uses: when `getItem` throws `SecurityError`, clearing data doesn't restore access.

Suggest dropping the advice and keeping only what's true, e.g. "…could not be read, so nothing was changed. Reload the page and try again."

services/vault/src/hooks/useVaultDeposits.ts
```diff
@@ -52,9 +55,25 @@ export function useVaultDeposits(connectedAddress: Address | undefined) {
...
+   * dropped rows it could not transform. React Query keeps `status` on
+   * "success" while serving stale cache through a failing background refetch,
+   * so a non-null `error` withholds the set too.
+   */
+  const indexedVaultIds: ReadonlySet<string> | null = useMemo(
+    () =>
+      status === "success" && error === null && data?.droppedCount === 0
+        ? new Set(confirmedActivities.map((a) => a.id.toLowerCase()))
```
gbarkhatov, line 70: My earlier T4 premise was wrong, and this term is the fallout. In TanStack Query v5 the `error` action sets `status: 'error'` (`query-core/src/query.ts`, the `case 'error'` reducer), and `success` resets it. So `status === "success" && error !== null` never occurs and `error === null` can't change the result. The doc comment above ("React Query keeps `status` on 'success' while serving stale cache through a failing background refetch") is false, and the test at `useVaultDeposits.test.ts:116-131` pins a state the library can't produce.

Suggest removing the term, that comment sentence and the test. The real gap is freshness (see the review body).

gbarkhatov (dismissed):
Follow-up to my review at `7f0ed85977`. The rebase carried the three reviewed commits over unchanged, so the delta is `5b7f98933` (gate on the record's own status, settle the dialog) and `3c3816f2d` (close or explain a dismiss whose record changed in another tab).

**Merge blocker (not code):** all 5 commits are unsigned (`git log %G?` = `N`), and `main`'s ruleset has `required_signatures`. `validate` fails for that reason.

| # | Prior finding | Verdict at `3c3816f2` |
|---|---|---|
| T1 | The gate never read the record's own status, so a broadcast Pre-PegIn was dismissible | **Partially fixed**, see below |
| T2 | Dismissal destroys the participant-key stamp (fail-closed resume becomes fail-open) | **Unaddressed.** No stamp check or on-chain read was added to `canDismissRecord` (`VaultsLifecycleSections.tsx:720-730`). The existing thread stands |
| T3 | Dialog body claims fund safety the gate doesn't establish | **Partially fixed.** True for records that correctly read PENDING, still not for the T1 residue |
| T4 | A stale, failing indexer cache still yields a non-null id set | **Partially fixed, and my premise was wrong**, see below |
| T5 | Confirm closes silently, and copy changes under the user | **Fixed** (snapshot at `:745-753`, explicit outcomes at `:773-789`) |
| T6 | "Browser blocked" copy shown for the unreadable branch too | **Fixed.** Now a separate `unreadable` outcome, but see the inline on its copy |
| T7 | A cross-device `""` tx hex breaks "removed together or not at all" | **Unaddressed** (`utils/batchedPegin.ts` is not in the delta). The existing thread stands |
| T8 | Id lowercasing untested | **Fixed** |
| T9 | The "cross-tab" test only proved the same-tab event | **Fixed** (renamed) |
| T10 | `useVaultCountCap` ignored `droppedCount` | **Fixed** |

**T1, changed.** The new status gate is only as strong as the CONFIRMING write, and every broadcast path writes it after the push resolves. For example, `useDepositFlow.ts` broadcasts at `:1078` and writes CONFIRMING at `:1123-1130`, and `useBroadcastState.ts:95-99` does the same.
- While another tab is signing and pushing, the record still reads PENDING everywhere.
- If that best-effort write fails, the record reads PENDING permanently.
- `prePeginBroadcastSeen` from the polling result is the Bitcoin-side signal that would catch the second case once the mempool sees the transaction.

**T4, changed, and correcting myself.** My earlier claim that TanStack keeps `status: "success"` through failing refetches is wrong for v5. The `error` action sets `status: "error"`, and retries only touch `fetchFailureCount`. So the `error === null` term added for T4 cannot change the result (inline). The real gap is freshness: data up to `staleTime` old, or served during the retry window, still reads as a complete id set. A freshness bound, or a `refetch()` when the dialog opens with confirm allowed only against newer data, would close it.

Verified clean: dismissal cannot change the vault-cap count, because `useVaultCountCap` counts only indexer rows, and the `droppedCount` term makes it stricter.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -670,6 +702,109 @@ export function VaultsLifecycleSections({
...
+  const handleDismissConfirm = useCallback(() => {
+    if (dismissBatch === null) return;
+    if (dismissBatch.every((a) => !realActivityIds.has(a.id))) {
+      setDismissError(null);
+      setDismissBatch(null);
+      return;
+    }
+    // Broadcast from another tab, already observed here: no longer this
+    // browser's to discard. A broadcast this tab has not observed yet is
+    // caught by the storage write below, which re-reads the stored status.
```
gbarkhatov, line 779: This comment claims more than the storage check does. The re-read in `removePendingPegins` only catches a broadcast once the other tab has *written* CONFIRMING, and two cases slip past that:

- **Broadcast still in flight.** `useDepositFlow.ts` pushes the Pre-PegIn at `:1078` and only calls `updatePendingPeginStatus(..., CONFIRMING)` at `:1123-1130`, after the push resolves. A dismissal confirmed in between still sees PENDING in storage and removes a record whose transaction is already on the network.
- **A CONFIRMING write that failed silently.** `updatePendingPeginStatus` goes through `savePendingPegins`, which swallows write errors, so that record reads PENDING indefinitely.

This is what remains of T1, narrower than before now that the React-state lag is closed. Two options:

- **Reword** to match what the check does: "A status another tab has already written is caught by the storage write below."
- **Close the gap** by adding a Bitcoin-side input to `canDismissRecord`, e.g. requiring `prePeginBroadcastSeen === false` from the polling result, as the original T1 thread suggested.

gbarkhatov (dismissed):
Follow-up review. Changes since my last review (`3c3816f2` → `1478d1cf`).

**Headline:** three things changed since my last review:
- `removePendingPegins` now takes an `expectedStatus` and checks it against the same raw read it writes back, returning `"changed"` on a mismatch. This closes the check-then-delete gap (T11).
- The unreadable-storage copy no longer advises clearing site data (T12).
- The dead `error === null` term is gone (T13).

The commit is signed, and CI is green, including `verify`. No merge-blockers. One inline comment is below: what remains of T1, and a new comment that overstates what the storage check catches.

### Status of prior threads

| # | Finding | Status at `1478d1cf` |
|---|---|---|
| T1 | The gate ignores the record's own status, so a broadcast Pre-PegIn is dismissible | Partly fixed, and the gap is narrower. See the inline comment. |
| T2 | Dismissal throws away the participant-key stamp, so the resume check no longer fails closed | Unaddressed; the existing thread stands |
| T3 | The dialog body promises funds are safe, which the gate doesn't prove | Partly fixed, unchanged since last time |
| T4 | A stale indexer cache still counts as a complete answer | Partly fixed: there is still no freshness bound or `refetch()` when the dialog opens |
| T5 | Confirm closes silently and the copy changes under the user | Fixed |
| T6 | "Browser blocked" copy also shown when storage is unreadable | Fixed |
| T7 | A cross-device `""` tx hex splits a batch | Unaddressed; the existing thread stands |
| T8 | Id lowercasing untested | Fixed |
| T9 | The "cross-tab" test only proved the same-tab event | Fixed (renamed) |
| T10 | `useVaultCountCap` ignored `droppedCount` | Fixed |
| T11 | Removal filtered on id alone, so the status check before it was a check-then-delete | Fixed (`peginStorage.ts:628-657`, tests at `peginStorage.test.ts:689` and `:718`) |
| T12 | "Clear this site's stored data" advice would wipe stamped records | Fixed |
| T13 | The `error === null` term could not change the result | Fixed |

### Verified clean

- **The new storage status check fails closed.** A missing status reads as PENDING, the same rule the read path uses. An unknown status compares unequal, so removal returns `"changed"` without writing or dispatching an event.
- **The dismissal gate still fails closed when the indexer drops rows.** `indexedVaultIds` stays null unless `droppedCount === 0`, and `useVaultCountCap` does the same.
- **Cross-tab changes between opening and confirming the dialog are refused.** They are caught either at the React-state check, at the live gate re-check, or inside the removal. None of them removes a record in its new state.

jonybur (commented):
Comments inline.

services/vault/src/storage/peginStorage.ts
```diff
@@ -422,9 +437,64 @@ function persistPendingPegins(
...
+/**
+ * The tracked status of a raw stored entry. A missing status reads as
+ * `PENDING`, matching the backward-compatible default `getPendingPegins`
+ * applies; anything else is returned verbatim so a status this build does not
+ * know never passes for one it does.
+ */
+function readStoredEntryStatus(entry: unknown): unknown {
+  if (!entry || typeof entry !== "object") return undefined;
+  const status = (entry as { status?: unknown }).status;
+  return status || LocalStorageStatus.PENDING;
```
jonybur, line 520: **readStoredEntryStatus uses `status || PENDING`, so a falsy stored status ("", 0, false, null) passes the expectedStatus PENDING check and the entry is removed.**

`readStoredEntryStatus` returns `status || LocalStorageStatus.PENDING`, so any falsy stored status (`""`, `0`, `false`, `null`) reads as PENDING. The read path only defaults a missing status: `hasValidSecurityFields` rejects anything that is not `undefined` or a known enum string. So this check is looser than the doc comment claims ("matching the backward-compatible default", "a status this build does not know never passes for one it does"). An entry with the target id and `status: ""`, for example a hidden duplicate next to a visible PENDING record, is deleted, while `status: "x"` correctly returns `"changed"`. Use `status === undefined ? LocalStorageStatus.PENDING : status` so the fail-closed check matches the read path.

services/vault/src/hooks/__tests__/useVaultDeposits.test.ts
```diff
@@ -78,6 +82,51 @@ describe("useVaultDeposits", () => {
...
+      error: null,
+      refetch: vi.fn(),
+    });
+
+    const { result } = renderHook(() => useVaultDeposits(ADDRESS));
+
+    expect(result.current.indexedVaultIds).toEqual(new Set(["0xabcdef"]));
+  });
+
+  it("withholds the indexed vault ids when the fetch dropped a row", () => {
```
jonybur, line 116: **No test covers indexedVaultIds staying null when the query status is not "success" but data is present; removing the status term keeps the suite green.**

Now that the `error === null` term is gone, `status === "success"` in `useVaultDeposits.ts:69` is the only thing that hides the Remove control when a background refetch fails over cached data. Every new hook test mocks `status: "success"`. The default mock has `data: undefined` and no status, so it returns null because `data` is missing, not because of the status. The component tests pass `indexedVaultIds` in directly. You could delete the status term and every test would still pass. Add a case with `data: { vaults: [], droppedCount: 0 }` and `status: "error"` that expects `indexedVaultIds` to be null.

services/vault/src/hooks/usePendingDeposits.ts
```diff
@@ -128,6 +134,22 @@ export function usePendingDeposits() {
...
     refetchActivities,
+    /**
+     * Discards browser-local pending deposits in one storage write — the whole
+     * batch of a split deposit at once. Returns the failure reason when the
+     * records are still stored.
+     */
+    removePendingPegins,
+    /**
+     * Lowercased ids of every vault the indexer returned, or null unless that
+     * set is known to be complete (answered successfully, no error, no dropped
```
jonybur, line 145: **Doc comment on indexedVaultIds still says the set requires "no error", a gate term that was removed; the real gate is only status success and droppedCount 0.**

The JSDoc on `indexedVaultIds` says the set is non-null only when the query "answered successfully, no error, no dropped rows". The `error === null` term was removed from `useVaultDeposits.ts`, and the gate there is now just `status === "success" && data?.droppedCount === 0`. This set is the only input the dismiss gate trusts, so the comment now describes a safety check that the code no longer performs, which will mislead the next person who reasons about it. Drop "no error" so it matches the hook's own comment.

gbarkhatov (dismissed):

gbarkhatov (approved):

services/vault/src/hooks/useVaultCountCap.ts
```diff
@@ -99,9 +100,13 @@ export function useVaultCountCap(
...
-    // activation. Mirrors useApplicationCap (capsQuery.error ?? usageQuery.error).
-    capUnavailable: capError || (address !== undefined && vaultsError),
+    // vaults list — and on a list that dropped rows it could not transform. A
+    // list-fetch error or a dropped row would otherwise under-count and let an
+    // at-cap user lock BTC, then revert at activation. Mirrors
+    // useApplicationCap (capsQuery.error ?? usageQuery.error).
+    capUnavailable:
+      capError ||
+      (address !== undefined &&
+        (vaultsError || (vaultsResult?.droppedCount ?? 0) > 0)),
```
jonybur, line 110: **capUnavailable is true whenever any indexer row fails to transform, so one malformed or unknown-status vault permanently blocks deposits and retrying never clears it.**

capUnavailable now includes `droppedCount > 0`, and `fetchVaultsByDepositor` increments that for every row `transformVaultItem` rejects: a null required field, malformed hex, or a status the frontend doesn't map yet, regardless of the vault's app or whether it is REDEEMED or EXPIRED. A depositor with a single such row gets the deposit CTA disabled with a "try again" message, but every refetch rebuilds the same list and drops the same row, so the block lasts until the indexer data or the frontend changes; a new indexer status would do this to every depositor holding one. The concern the count cap guards against is under-counting, which can be handled without a hard block: add `droppedCount` to `currentCount` so the check over-counts conservatively (as the comment above COLLATERALIZABLE_STATUSES already intends), and only at-cap users are stopped. The new test in useVaultCountCap.test.ts that asserts the blocking behaviour would need to follow.

jonybur (commented):
Comments inline.

gbarkhatov (approved):

PR #2490, opened 2026-09-09
fix(vault): show withdrawing vaults and revalidate after activation

### LINKED ISSUE #2429: Withdrawn vaults disappear from the UI instead of showing a withdrawing state
## Context

After initiating a withdrawal the vault vanishes from the UI rather than showing a withdrawing or settled state.

## Changes

**Now:** the vaults list is built by `toCollateralVaultEntries` (`services/vault/src/hooks/useDashboardState.ts:82-89`), the only place `collateralVaults` — which feeds `/vaults` — is assembled. It filters through `isActiveCollateral` (`services/vault/src/utils/collateral.ts:23-30`), which drops any row whose indexer `removedAt` is non-null, and through `EXCLUDED_VAULT_STATUSES = new Set(["liquidated", "depositor_withdrawn"])` (`collateral.ts:15`).

There is no intermediate state in the type or the UI — a vault is either shown or gone.

**Expected:** a withdrawing vault stays visible with a status, and disappears only once settled.
**Where:** `collateral.ts:15,23-30`, `useDashboardState.ts:82-89`

## Not proven

When the indexer sets `removedAt` relative to withdraw initiation versus confirmation is decided in `babylon-vault-indexer`, not this repo. Check there before adding an interim status — the fix may be to key off a status the indexer already exposes.

Related: #2430 covers the same page not revalidating after a withdrawal.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788367731470809)

> I withdrew a vault from devnet and it just disappeared from the UI

</details>

### LINKED ISSUE #2430: Vaults page does not revalidate after an activation or a withdrawal
## Context

`/vaults` does not refresh itself after an activation or a withdrawal. The reporter had to reload the page to see a second activated vault, and again to make a withdrawn vault leave the list.

## Changes

**Now:** the page's data comes from `VaultsPage.tsx:38,60-61` → `useVaultsPageData` (`services/vault/src/hooks/useVaultsPageData.ts:28`) → `useDashboardState` (`services/vault/src/hooks/useDashboardState.ts:32`) → `useAaveUserPosition` (`services/vault/src/applications/aave/hooks/useAaveUserPosition.ts:107-127`):

- key `["aaveUserPosition", connectedAddress, spokeAddress, vbtcReserveId, borrowableReserveIdsKey]`
- `refetchOnMount: true`, `refetchInterval: POSITION_REFETCH_INTERVAL_MS` = 30s (`applications/aave/constants.ts:78`), no explicit `staleTime`

A second hook, `useVaults` (`services/vault/src/hooks/useVaults.ts:29-47`), owns `["vaults", depositorAddress]` with `staleTime: 60_000`, `refetchOnWindowFocus: false`, and polls at 5s only when `options.poll` is set.

**Withdraw:** `useWithdrawCollateralTransaction.ts:97-101` calls `markVaultsAsPending(...)` then `invalidateVaultQueries(queryClient, address)`, which invalidates both `["vaults", address]` and `["aaveUserPosition", address]` (`services/vault/src/utils/queryKeys.ts:31-45`). That fires once, immediately after the tx is mined — before the Aave indexer has necessarily processed the event. The refetch can return stale data, which marks the query fresh again, and nothing retries until the 30s poll.

**Activate:** no invalidation at all. `useActivationState.ts:94-97` has `onRefetchActivities: () => { /* No-op … the polling interval refreshes activities on its own */ }`. It only calls `addActivatingVault(...)` (`applications/aave/context/ActivatingVaultsContext.tsx:104-119`), an in-memory optimistic placeholder that is neither React Query state nor persisted.

**The mechanism that fixes this is wired to the wrong page.** `useSyncPendingVaults` (`applications/aave/context/PendingVaultsContext.tsx:151-206`) watches the fast-polling vaults result and, once the confirmed change is visible there, calls `queryClient.refetchQueries({ queryKey: ["aaveUserPosition", address] })` and clears the pending marker. It is called only from `services/vault/src/components/simple/DashboardPage.tsx:119`. `VaultsPage.tsx` calls neither `useAaveVaults` nor `useSyncPendingVaults`.

**Expected:** `/vaults` reconciles after both operations without a manual reload.
**Where:** `VaultsPage.tsx`, `PendingVaultsContext.tsx:151-206`, `useActivationState.ts:94-97`

The smallest fix is to use `useAaveVaults` + `useSyncPendingVaults` on `VaultsPage` as `DashboardPage` already does, and to invalidate on activation instead of relying on the optimistic placeholder.

## Not proven

Measure the actual indexer lag on devnet at the moment `invalidateVaultQueries` fires, to size a retry rather than a fixed delay. Also check whether the activation guard at `useActivationState.ts:111-118` (`Number.isFinite(amountBtc) && amountBtc > 0`) ever skips adding the placeholder for a real activity shape.

Related: #2429 covers withdrawn vaults disappearing outright.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788446153825619)

> This page has some refreshing problem, demo.vault-devnet.babylonlabs.io/vaults. After both vaults are activated, I could only see one. I had to refresh the page manually to see both.
>
> After I initiated a withdrawal of a vault, and returned to this page, both vaults were still shown to be withdrawable. Similarly, I had to manually refresh the page to have one disappear.

</details>

jeremy-babylonlabs:
Closes #2429
Closes #2430

Two causes of one failure on `/vaults`, so they share a branch. Separable in the diff.

## #2429 — a withdrawn vault vanished

The list only knew "shown" or "gone": `isActiveCollateral` dropped any row with a non-null `removedAt`, and `EXCLUDED_VAULT_STATUSES` excluded `depositor_withdrawn`. The moment a withdrawal was recorded on-chain the row disappeared, even though the Bitcoin payout takes hours.

Adds a real third state, driven by the indexer's own status rather than local state. Verified against `babylon-vault-indexer`:

- `core/pegin.ts:403` — the `VaultClaimableBy` handler sets `REDEEMED` when the depositor's withdrawal is recorded on-chain.
- `applications/aave/position.ts:253` — `CollateralWithdrawn` sets `removedAt` in the same transaction.
- `btcMonitorSync.ts:15` — `depositor_withdrawn` lands later, only once btc-monitor sees the vault UTXO spent. That is settlement.

So `redeemed` = peg-out in flight, `depositor_withdrawn` / `liquidated` = settled and still excluded. This is the same signal `useAaveVaults.redeemedVaults` already uses, not a new source of truth.

`CollateralVaultLifecycle` (`"active" | "activating" | "withdrawing"`) replaces the optional `isActivating` flag — a discriminant rather than a growing set of booleans. Withdrawing rows show status and no ordinal, and are excluded from reorder, withdraw, and the liquidation cascade's `totalBtc`.

## #2430 — the page did not revalidate

Activation's `onRefetchActivities` was a no-op, and nothing invalidated the queries a withdrawal changes, so the page served its cache until the next 30s poll.

One helper, `invalidateVaultQueries(queryClient)`, invalidating by root key `["vaults"]` and `["aaveUserPosition"]`, called from the withdraw transaction hook, activation success, and reorder success. It deliberately takes no `address`: the withdraw hook passes wagmi's checksummed address while the position key is built from `useETHWallet`, so exact-segment matching could silently miss.

Fixed by invalidation, not by shortening the poll interval — a shorter interval is a workaround that costs every user RPC calls forever.

## Display consistency

Three numbers describe a position, and they now agree on what a withdrawing vault is:

- **BTC total.** `displayCollateralBtc` in `useDashboardState` starts from the chain's `collateralBtc`, adds activating vaults the chain has not counted yet, and subtracts withdrawing vaults the chain still counts. Both adjustments key off the same `vaultIds` read as `collateralBtc`, so nothing is applied twice.
- **Summary card count.** `activeVaultsCount` reads `countActiveVaults(displayVaults)`, which counts `active` and `activating` rows only, matching the BTC total.
- **Section heading.** "Active Vaults (n)" counts every row rendered beneath it, withdrawing rows included, so the heading never claims fewer rows than the reader can see.

Action gates stay on chain state: `hasCollateral` is `collateralBtc > 0`, so an optimistic row never enables borrow, withdraw, or the /overview liquidation card before the collateral exists on-chain. A withdrawal-only position reads "0 sBTC" with the row visible showing its status.

## Notes

- Activation success invalidates the activities query as well as the vault queries, so the pending row leaves the list without waiting for the next poll. The pending-vaults refetch after a withdrawal is limited to `type: "active"` queries.
- The withdraw review pins its selected vaults, collateral, and health factor at confirm time, so the amounts no longer flash to zero while the transaction is in flight and the progress view mounts.

- The optimistic withdrawal overlay (`applyPendingWithdrawals`) runs in `useDashboardState`, so every position view — /overview, /loans, /vaults and the notification banner — sees a mined-but-unindexed vault as `withdrawing`. `usePendingVaults()` throws outside `PendingVaultsProvider`, so the router now mounts that provider for `/activity` too (`ActivityWithProviders`), which reaches `useDashboardState` through the god-mode section.
- Deleted `vaultOrderQueryKey`. Nothing has ever registered a query under `["vaultOrder"]`; it has been invalidate-only since #2095.
- Copy: reuses the existing `REDEEM_IN_PROGRESS` label rather than minting a second string for the same protocol state. The row's button says "Withdraw" while the status says "Redeem" — that is app-wide terminology, not this row.
- No `data-testid` moved or renamed. `e2e/real/actions/withdraw.ts` keys on the Withdraw button's disabled state, so a visible withdrawing row is skipped, not mistaken for a candidate.

## Testing

New tests cover: a peg-out in flight stays as a withdrawing row; a redeemed vault whose `removedAt` is not yet indexed stays; settled and liquidated vaults stay excluded; withdrawing rows order after active ones and stay out of the liquidation sequence; root-key invalidation fires on withdrawal and activation and leaves unrelated queries alone; the withdrawal-only case asserts one visible row **and** a zero count together, so the pair cannot drift again; and the display total nets a withdrawing vault out while the chain still counts it.

`pnpm --filter vault run lint` — 0 errors. `pnpm --filter vault run test` — 318 files, 3888 tests green.


jonybur (commented):
Traced both halves at 2a4a5b05 and the logic holds throughout. One concrete defect below; nothing functional blocks.

**The two doc blocks in `services/vault/src/utils/collateral.ts` are swapped.** Two JSDoc blocks are now stacked on `countActiveVaults` (around lines 77-83). The first — "Re-tags entries whose withdrawal transaction has been mined but which the indexer has not reported yet" — describes `applyPendingWithdrawals`, not the counter it sits on: `countActiveVaults` counts rows and re-tags nothing. `applyPendingWithdrawals` is then left with no doc at all, and it is the function that most needs one. It is the seam between the local pending marker and the indexer, and the case-insensitive id match is a non-obvious detail worth recording where a reader will look for it. Move the re-tagging block down onto `applyPendingWithdrawals` and leave the one-line count description on `countActiveVaults`.

**Checked and correct:**

- `classifyCollateral` ordering is right. Letting status lead means `depositor_withdrawn` and `liquidated` drop whatever `removedAt` says, `redeemed` becomes `withdrawing` whether or not `removedAt` has landed, and everything else falls through to the old `removedAt` rule. The redeemed-with-null-removedAt case the second new test pins is exactly the window the two indexer handlers open.
- The `withdrawing` tag reaches every action surface. `rawCollateralVaults` returns the post-overlay list, so `WithdrawFlow` and `VaultsPage`'s `reorderableVaults` both see it, and the per-row Withdraw button is disabled on a non-active lifecycle. A vault cannot be offered for a second withdrawal during the indexer gap.
- Narrowing `isReorderOverrideReconciled` to `activeCollateralVaults` is the correct half to change. On the raw list a withdrawal arriving after a reorder would clear the override and snap the ordinals back; the fourth new test pins that.
- The `usePendingVaults` placement is right and the hazard you describe is real. `/vaults` sits inside the Aave layout provider and `useVaultsPageData` has no other consumer, so the throw is unreachable; `/activity` reaching `useDashboardState` through the god-mode panel is genuinely outside it.
- `useSyncPendingVaults` awaiting the position refetch before `clearPendingVaults` is what keeps the handoff seamless, and moving that refetch to the root key closes a real hole: the old address-scoped form could miss on a checksum mismatch, silently no-op, and clear the marker against stale data. The same reasoning carries the address out of `invalidateVaultQueries`.
- `activatingEntries` still dedupes against `rawCollateralVaults`, which now includes withdrawing rows, so a vault cannot appear as both.
- `countActiveVaults` backs both the summary count and the section heading, so the two cannot drift, and the withdrawal-only test asserting one visible row against a zero count is the right pairing.
- No `data-testid` moved or dropped, and `REDEEM_IN_PROGRESS` is an existing copy key rather than a new string.

services/vault/src/utils/queryKeys.ts
```diff
@@ -5,42 +5,28 @@
...
-    // Invalidate vaults query to refresh vault list
-    queryClient.invalidateQueries({
-      queryKey: [VAULTS_QUERY_KEY, address],
-    }),
-
-    // Invalidate Aave user position to refresh collateral amount
+    queryClient.invalidateQueries({ queryKey: [VAULTS_QUERY_KEY] }),
     queryClient.invalidateQueries({
-      queryKey: ["aaveUserPosition", address],
+      queryKey: [AAVE_USER_POSITION_QUERY_KEY],
```
jrwbabylonlab, line 30: Borrow and repay weren't migrated: `useBorrowTransaction.ts:133-135` and `useRepayTransaction.ts:231-233` still invalidate the literal `["aaveUserPosition", address]` with wagmi's checksummed address, while the query is keyed from `useETHWallet().address` (`useAaveUserPosition.ts:106-112`). Neither side normalizes case, so React Query's structural match misses and the position serves stale cache until the next poll — the hazard this change exists for.

services/vault/src/applications/aave/hooks/usePositionNotifications.ts
```diff
@@ -119,10 +119,11 @@ export function usePositionNotifications(
...
     // yet, and a sentinel `liquidationIndex`. Including them would inflate
     // `totalBtc`, pushing every liquidation price DOWN — understating the
-    // risk — and label a band "Vault 9007199254740992". The cascade models
-    // what the protocol would seize, so it sees indexed vaults only.
+    // risk — and label a band "Vault 9007199254740992". A withdrawing vault
+    // has left the position and would inflate it the same way. The cascade
+    // models what the protocol would seize, so it sees active vaults only.
     const indexedVaults = collateralVaults.filter(
-      (entry) => !entry.isActivating,
+      (entry) => entry.lifecycle === "active",
```
jrwbabylonlab, line 149: This reads `useDashboardState` directly, but `applyPendingWithdrawals` is only applied inside `useVaultsPageData` — so during the mined-to-indexed gap the withdrawn vault is still tagged `active` here. Its BTC stays in `totalBtc`, pushing every liquidation price down and understating risk, which is what the comment just above says this filter is for. /vaults already shows the row as withdrawing while /overview and /loans don't.

services/vault/src/hooks/useDashboardState.ts
```diff
@@ -164,9 +182,13 @@ export function useDashboardState(connectedAddress: string | undefined) {
...
   // Display gate — drives the Collateral section's summary-vs-empty rendering,
-  // so the just-activated vault shows during the indexer gap.
+  // so the just-activated vault shows during the indexer gap and a withdrawing
+  // vault, whose `collateralBtc` the indexer has already zeroed, still shows
+  // until its payout settles.
   const hasDisplayCollateral =
-    collateralBtc > 0 || activatingEntries.length > 0;
+    collateralBtc > 0 ||
+    activatingEntries.length > 0 ||
+    withdrawingCollateralVaults.length > 0;
```
jrwbabylonlab, line 222: Withdraw your only vault and this goes true while `collateralBtc` is 0 and `hasLoans` is false. `DashboardPage` passes it through as `LiquidationAnalysisSection.hasCollateral`, whose `hasCollateral && !hasLoans` branch then renders the "you have collateral, borrow" CTA — on /overview, which shows no withdrawing row to explain it. Previously that position got the correct no-deposit empty state.

jonybur (dismissed):
Re-reviewed at head. The doc-block swap I flagged earlier is now correctly placed. The invalidation callers that previously took an address argument have all been converted to the root-key form, so the checksummed-vs-lowercased mismatch is no longer a hazard. The lifecycle discriminated union replaces the ad-hoc isActivating flag consistently across the whole callchain, including the demo scenarios and the tests. The new withdrawing-branch test in VaultsActiveSection pins the disabled Withdraw button and the peg-out label. CI green.

services/vault/src/hooks/useDashboardState.ts
```diff
@@ -159,14 +195,19 @@ export function useDashboardState(connectedAddress: string | undefined) {
...
 
-  // Financial gate — drives action-enabling (e.g. Borrow). Indexer-pure: an
-  // optimistic activating row must NOT unlock financial actions before the
-  // collateral actually exists on-chain.
-  const hasCollateral = collateralBtc > 0;
+  // Financial gate — drives action-enabling (e.g. Borrow). Counts the indexer
+  // rows, which carry no optimistic activating row, so collateral that does not
+  // exist on-chain yet cannot unlock a financial action, and a vault whose
+  // withdrawal is already mined stops backing one before the indexer catches up.
+  const hasCollateral = countActiveVaults(rawCollateralVaults) > 0;
```
gbarkhatov: This moves a financial gate from an indexer scalar onto a list that a 24h localStorage marker can zero — and two routes that consume it never mount the hook that clears the marker.

`hasCollateral` was `collateralBtc > 0`. It's now `countActiveVaults(rawCollateralVaults) > 0`, and `rawCollateralVaults` has already had `applyPendingWithdrawals` applied at lines 91-100 of this file. So a `"withdraw"` marker read out of `localStorage` can flip this false while the indexer still reports the collateral.

`Liquidations/index.tsx:209` is `if (!isGodMode && !hasCollateral)` and returns the empty state **before** it ever looks at `hasLoans` — so a user who withdraws and navigates to `/liquidations` gets a blanked page while still holding debt. It can't self-clear there: I grepped all of `services/vault/src`, and `useSyncPendingVaults` is mounted in exactly two production places, `DashboardPage.tsx:118` and `VaultsPage.tsx:59`. The outer bound is the storage TTL. `SECURITY_MODEL.md` invariant 8 names browser storage an untrusted boundary, and it's now load-bearing for a gate.

I want to be fair to the intent — the comment directly above says the mined-withdrawal behaviour is deliberate, and as a *Borrow* gate that reasoning is sound. It's the collateral damage on routes that don't clear the marker that looks unintended. Keeping the gate indexer-pure (`collateralBtc > 0`, minus only the pending rows' `amountBtc`) and letting the marker drive display would preserve the intent without it.

services/vault/src/hooks/useVaultsPageData.ts
```diff
@@ -55,13 +56,15 @@ export function useVaultsPageData(connectedAddress: string | undefined) {
...
   // is already liquidation-ordered by useDashboardState; demo rows keep their
-  // merged position. Optimistic activating rows stay out of the sequence —
-  // their liquidation position is a placeholder until the indexer ingests the
-  // vault (their row hides the ordinal for the same reason) — while
-  // `activeVaultsCount` still includes them so the summary matches the Active
-  // Vaults section header count.
+  // merged position. Only `active` rows enter the sequence — an activating
+  // vault's liquidation position is a placeholder and a withdrawing vault has
+  // left the position (their rows hide the ordinal for the same reason).
+  // `activeVaultsCount` counts the same rows the collateral total does, so an
```
gbarkhatov: This comment asserts an invariant that the marker window breaks — and all three review lanes found it independently.

`activeVaultsCount` is `countActiveVaults(displayVaults)`, which excludes `withdrawing`, while `totalCollateralBtc` resolves to `displayCollateralBtc` = the indexer's `collateralBtc` plus activating amounts. For an *indexer-driven* `redeemed` row the two do agree, because the indexer's `CollateralWithdrawn` handler decrements `totalCollateral` in the same transaction that sets `removedAt` — confirmed in the local checkout.

But for a row re-tagged by the **local marker**, the indexer hasn't decremented anything, so the card renders something like "2.0 BTC" above "1 vault". That's precisely the case this PR adds, so the comment is false exactly where it's most load-bearing.

services/vault/src/components/vaults/VaultsActiveSection.tsx
```diff
@@ -182,7 +191,7 @@ export function VaultsActiveSection({
       <Heading variant="h6" as="h2" className="font-normal text-accent-primary">
         {COPY.vaults.sections.activeVaultsTitle}{" "}
         <span className="text-accent-secondary">
-          {COPY.vaults.sections.count(vaults.length)}
+          {COPY.vaults.sections.count(countActiveVaults(vaults))}
```
gbarkhatov: "Active Vaults (0)" can render directly above N visible withdrawing rows.

The heading count switched to `countActiveVaults(vaults)`, which excludes `withdrawing`, but line 198 still maps over the full `vaults` array. So a position whose only vault has a peg-out in flight shows the count as zero with the row plainly rendered beneath it.

Either count the rendered rows, or give withdrawing rows their own subheading so the count and the list describe the same set.

services/vault/src/utils/collateral.ts
```diff
@@ -61,6 +74,31 @@ export function toCollateralVaultEntries(
   });
 }
 
+/** Rows that count towards the position's vault count — every row but a peg-out in flight. */
```
gbarkhatov: This helper serves a display count and a financial gate with contradictory contracts.

It counts `active` **and** `activating` — right for a display count, wrong for a Borrow gate. It stays correct at `useDashboardState.ts:202` only because `rawCollateralVaults` happens never to contain an activating row (those are built separately and joined later). That invariant is stated in a comment and enforced nowhere — while the same helper is *deliberately* fed activating rows at `useVaultsPageData.ts:79` and `VaultsActiveSection.tsx:194`.

So the safety of the gate rests on a caller-side coincidence that the helper's own other callers already violate by design. Anyone who later passes `collateralVaults` here unlocks Borrow on collateral that doesn't exist on-chain. Splitting it, or filtering `lifecycle === "active"` inline at the gate, would make that impossible rather than merely unlikely.

services/vault/src/hooks/deposit/useActivationState.ts
```diff
@@ -92,9 +95,7 @@ export function useActivationState({
           pendingPegin,
           updatePendingPeginStatus,
           onRefetchActivities: () => {
-            // No-op: the optimistic CONFIRMED status set below drives the
-            // activated terminal, and the polling interval refreshes
-            // activities on its own.
+            void invalidateVaultQueries(queryClient);
```
gbarkhatov, line 99: `onRefetchActivities` no longer refetches activities. This now calls `invalidateVaultQueries`, which invalidates only `["vaults"]` and `["aaveUserPosition"]` — but the activities key is `[ACTIVITIES_QUERY_KEY, userAddress]` = `["user-activities", addr]`, and neither prefix matches.

It's strictly better than the previous no-op, so this isn't a regression. But the name now misdescribes what it does, and `/activity` can keep showing the pre-activation row. Either rename it, or add the activities key to the invalidation set — this is the one key the set is missing.

services/vault/src/components/simple/WithdrawFlow/index.tsx
```diff
@@ -64,9 +64,12 @@ function WithdrawFlowContent({
   // fake vaultId) and must never be selectable for a real withdraw, even if a
   // caller mistakenly passes the demo-merged list. Mirrors CollateralSection's
   // actionableVaults filter. Always a no-op in production (the flag is never
-  // set there).
+  // set there). Only `active` vaults back the position.
   const withdrawableVaults = useMemo(
-    () => collateralVaults.filter((v) => !v.displayOnly),
+    () =>
```
gbarkhatov, line 59: Cosmetic, but user-visible: the Review pane's amounts flash to 0 BTC / $0 after Confirm, before it navigates.

Adding `v.lifecycle === "active"` to `withdrawableVaults` means the selection empties the moment the marker is written. In `useWithdrawCollateralTransaction.ts:97`, `markVaultsAsPending` fires *before* the `await invalidateVaultQueries(queryClient)` on line 101, and `goToProgress()` only runs after `executeWithdraw` resolves — so React commits a render in between with `renderedStep` still `REVIEW`, and `selectedBtc`/`selectedUsd`/`projectedHealthFactor` all recompute to zero for the length of an RPC round-trip.

Navigating to PROGRESS before awaiting would close the window.

services/vault/src/applications/aave/context/PendingVaultsContext.tsx
```diff
@@ -185,16 +185,14 @@ export function useSyncPendingVaults(vaults: VaultData[]): void {
       // This ensures we show the loading UI until fresh RPC data is available,
       // preventing the brief flash of stale cached data.
```
gbarkhatov, line 186: Dropping the address segment was right, but it widened this call more than intended. `invalidateQueries` defaults to `refetchType: "active"`; `refetchQueries` refetches **all** matching entries. With the `address` segment gone, every cached `aaveUserPosition` variant is now refetched, and the `await` gates `clearPendingVaults`.

Adding `type: "active"` costs nothing and restores the previous scope. Worth saying in the same breath that removing the address was a genuine fix, not tidying — the old form matched a wagmi-checksummed address against a key built elsewhere, and could silently match nothing.

services/vault/e2e/real/actions/withdraw.ts
```diff
@@ -157,7 +157,7 @@ async function openWithdrawForRow(
         "No active vault rows on /vaults — this position has nothing to withdraw.",
       );
     throw new Error(
-      `No withdrawable vault on /vaults after ${Math.round(WITHDRAW_CTA_ENABLE_TIMEOUT_MS / MS_PER_SECOND)}s (${found.rowCount} row(s) shown — every Withdraw button is disabled: the vault is not in use, still activating, or withdrawals are paused by the protocol). Repay outstanding debt first so collateral can be released.`,
+      `No withdrawable vault on /vaults after ${Math.round(WITHDRAW_CTA_ENABLE_TIMEOUT_MS / MS_PER_SECOND)}s (${found.rowCount} row(s) shown — every Withdraw button is disabled: the vault is not in use, still activating, already withdrawing, or withdrawals are paused by the protocol). Repay outstanding debt first so collateral can be released.`,
```
gbarkhatov, line 160: Low, and only partly actionable here: this hunk updates the failure message for the new lifecycle, but lines **29** and **380** of this same file still describe rows as `isActivating` — a field this PR deletes.

Those two lines fall outside every hunk in the diff, so I can't anchor a comment to them directly; flagging here since this is the one place in the file the diff touches. Worth a sweep while the file is open.

gbarkhatov (dismissed):
Reviewed at `7f0f3c9` via three independent passes (7 agents), deduped, with every finding re-checked against the source at HEAD.

**No hard merge blockers — the central design is sound, and both security-flavoured questions come out clean.**

**The "withdrawing" derivation does not repeat the reclaim mistake.** It has two sources — the indexer's Ethereum-side `status === "redeemed"`, and the local mined-but-unindexed marker — and both are used *only* to display a peg-out and to **remove** actionable surface, never to enable one. Every gate keyed on it is restrictive: Withdraw is disabled for `lifecycle !== "active"`, `withdrawableVaults` and `reorderableVaults` filter to active, and the liquidation cascade excludes it, which *overstates* risk — the safe direction. The row only disappears on **Bitcoin settlement** (`depositor_withdrawn`), which is exactly the reclaim rule. So each status is used for a purpose it can bear: the earlier Ethereum signal shows an in-flight peg-out, and Bitcoin settlement is what retires the row.

**Cache revalidation is under-inclusive by exactly one key, and not over-broad.** `invalidateVaultQueries` covers `["vaults"]` and `["aaveUserPosition"]`, both of which prefix-match the real keys, across six call sites all placed after the receipt. The missing one is `["user-activities", addr]` — see my comment on `useActivationState.ts`. The "still showing a withdrawn vault" direction is closed, because `useSyncPendingVaults` awaits the position refetch before clearing the marker. The "showing a withdrawal that never landed" direction is my first comment. Also worth saying: **dropping the address segment from the key was a real fix, not tidying** — the old form used a wagmi-checksummed address against a key built elsewhere and could silently match nothing.

**Per-row hooks and duplicated polling: clean.** `ActiveVaultRow` instantiates no hooks, no `useQuery`/`useEffect` was added inside any list body across the 35 files, and nothing duplicates `PeginPollingContext`. One thing to confirm is intended rather than incidental: `VaultsPage.tsx:56-59` now mounts `useAaveVaults` for a single destructured field, adding a polling observer to the shared `["vaults", addr]` key — and `hasRedeemedVaults` keeps that true for the whole multi-hour peg-out window, so `/vaults` now polls at 5s where it previously did not. CPU and RPC only, but it's a behaviour change that isn't in the description.

- *"`redeemed` now overrides `removedAt`, so a row the indexer removed can never leave."* Not a regression — `classifyCollateral` drops on the settled set first, and the only exit being `depositor_withdrawn` is the *intended* fix for #2429. Keeping the row until Bitcoin settlement is the point.
- *"A `removedAt` arriving before the status write silently drops the row."* That path returns `null`, which is exactly what the pre-PR `isActiveCollateral` did for any non-null `removedAt`. Strictly more permissive now. Worth a sentence on emission order, nothing more.
- *"The permissive `return "active"` fallback classifies `expired`/`invalid` as actionable collateral."* Pre-existing — the old code used a deny-list too.

**Verified clean:** zero dead code — `vaultOrderQueryKey` has no survivors across `services/` and `packages/`, and both call sites are converted. E2E hooks intact: `vault-row-${vaultId}` and `vault-withdraw-button` both retained with their pointer comments, and the harness keys on the *enabled* Withdraw button, so a visible withdrawing row is skipped rather than mistaken for a candidate. No new inlined copy — the withdrawing badge reuses `COPY.pegin.labels.REDEEM_IN_PROGRESS`. The discriminated union is applied properly: `applyPendingWithdrawals` is idempotent and returns the same array reference when the marker set is empty, so memo stability is preserved. And the stale-selection race is closed — `getEffectiveVaultSelection` intersects rather than falling back to "all", so a vault flipping to `withdrawing` mid-dialog yields an empty selection rather than a full-position withdraw.

jonybur (commented):
One finding on the second commit — the "Active Vaults (N)" heading now mislabels the rendered count.

`VaultsActiveSection.tsx` renders the heading with `vaults.length` after `91eb7b8e`, which resolves the earlier "Active Vaults (0) above N withdrawing rows" contradiction. Because `vaults` still contains `withdrawing` rows, the heading reads "Active Vaults (3)" when 2 rows are active and 1 is withdrawing. `services/vault/src/utils/collateral.ts:78-83`'s `countActiveVaults` still feeds the summary card, which continues to exclude withdrawing. The vaults page therefore renders "Vaults: 2" in the summary card next to "Active Vaults (3)" over 3 rows — two contradictory counts on the same screen.

The fix closed the local rows-vs-count contradiction but opens a summary-vs-heading one. Two options: drop `withdrawing` from the list under the "Active Vaults" heading (matching `countActiveVaults`), or rename the heading to a lifecycle-neutral term ("Vaults") that matches the row set. Either closes both.

services/vault/src/hooks/useDashboardState.ts
```diff
@@ -153,27 +188,38 @@ export function useDashboardState(connectedAddress: string | undefined) {
...
+    withdrawingCollateralVaults,
+    position?.vaultIds,
+  ]);
 
   // Optimistic rows must not enable actions before collateral exists on-chain.
   const hasCollateral = collateralBtc > 0;
   // Display gate — drives the Collateral section's summary-vs-empty rendering,
-  // so the just-activated vault shows during the indexer gap.
+  // so the just-activated vault shows during the indexer gap and a withdrawing
+  // vault, whose `collateralBtc` the indexer has already zeroed, still shows
```
gbarkhatov: Stale after the rebase onto #2462. `collateralBtc` is now the chain adapter's `totalCollateralBTC` (`positionService.ts:157` → `useAaveUserPosition.ts:178-181`), so the indexer no longer zeroes it. The test name at `useDashboardState.test.ts:236` ("…after the indexer zeroes the collateral") has the same problem, and it contradicts that test's own comment at `:244` ("The chain snapshot already dropped the vault"). Suggest "chain" in both places.

gbarkhatov (dismissed):
Follow-up to my review at `7f0f3c9953`. The commits were rewritten in the rebase, so I read the range-diff. All 8 prior threads are fixed.

**Merge blocker (not code):** the rebase dropped the commit signatures. `968e99baf`, `6a038be39` and `91eb7b8e8` are all unsigned (`git log %G?` = `N`), while the commits I reviewed were signed. `main`'s ruleset has `required_signatures`, and `validate` fails on its "Validate commit signatures" step. Re-signing the three commits, e.g. `git rebase --exec 'git commit --amend --no-edit -S' <base>`, clears it.

| # | Prior finding | Verdict at `91eb7b8e` |
|---|---|---|
| T1 | `hasCollateral` moved onto a list the localStorage marker can zero | **Fixed.** Now `collateralBtc > 0`, the chain total (`useDashboardState.ts:213-214`) |
| T2 | The "count matches total" comment was false in the marker window | **Fixed.** Withdrawing rows still in the chain `vaultIds` are netted out (`:195-211`), with a test at `useDashboardState.test.ts:261-274` |
| T3 | "Active Vaults (0)" above visible withdrawing rows | **Fixed as written.** See below |
| T4 | `countActiveVaults` served both display and a financial gate | **Fixed.** Its only caller is display (`useVaultsPageData.ts:82`) |
| T5 | `onRefetchActivities` did not refetch activities | **Fixed** (`useActivationState.ts:98-103`), with a test |
| T6 | Review amounts flash to 0 after Confirm | **Fixed.** Review inputs are pinned at confirm (`WithdrawFlow/index.tsx:81-98`, `:145-166`) |
| T7 | Root-key `refetchQueries` widened scope | **Fixed** with `type: "active"` (`PendingVaultsContext.tsx:188-191`) |
| T8 | `isActivating` still mentioned in the e2e doc | **Fixed** |

**T3, and correcting my own suggestion.** The author took my first option, counting the rendered rows. The heading now includes withdrawing rows (`VaultsActiveSection.tsx:193`), while the summary card still excludes them (`collateral.ts:83-87` via `useVaultsPageData.ts:82`), so the page can show "Vaults: 2" next to "Active Vaults (3)". My second option, a separate subheading for withdrawing rows, avoids that. The same point is already open in another review at this head, so I'm not opening a new thread.

**Optional:** the Review pin in `WithdrawFlow/index.tsx:74-98` has no test. It deliberately uses a ternary rather than `??` so that a pinned `null` health factor is kept, and it exists to prevent a false `withdraw-hf-block-warning`, which the e2e harness reads. A render test that changes the collateral props mid-submit would pin it.

Checked, and fine:
- The "withdrawing" state never feeds a financial value. `collateralBtc` is chain-derived, and HF/USD come from the Spoke.
- Excluding `redeemed` rows from the chain comparison is stricter, not looser. A vault still on chain but marked `redeemed` now raises `indexerError`.
- All five testids used by `e2e/real/actions/withdraw.ts` are present at head.
- Liquidated vaults can't show as withdrawing (`VaultClaimableBy` skips them).

gbarkhatov (approved):
Follow-up review. Changes since my last review (`91eb7b8e` → `0742d9df`).

**Headline:** all 9 prior threads are fixed. After the squash and rebase, the net patch differs from what I reviewed only in the two wording lines T9 asked for:
- the test name in `useDashboardState.test.ts:236`, now "after the chain zeroes the collateral";
- the comment in `useDashboardState.ts:215-218`, now "the chain snapshot has already zeroed".

The commit is signed and `validate` passes, which clears the blocker from my last review. CI is green on all checks, including `verify` and `visual`. No merge-blockers. No new defects.

### Status of prior threads

| # | Finding | Status at `0742d9df` |
|---|---|---|
| T1 | `hasCollateral` read a list the localStorage marker can zero | Fixed: `collateralBtc > 0` (`useDashboardState.ts:213-214`) |
| T2 | "Count matches total" comment was false during the marker window | Fixed: withdrawing rows still present in chain `vaultIds` are subtracted (`useDashboardState.ts:191-211`, `useVaultsPageData.ts:62-66`) |
| T3 | "Active Vaults (0)" rendered above visible withdrawing rows | Fixed as written (`VaultsActiveSection.tsx:193`). The heading-vs-summary mismatch this creates is another reviewer's open point, so I'm not repeating it. |
| T4 | `countActiveVaults` served both a display count and a financial gate | Fixed: its only production caller is `useVaultsPageData.ts:82`, and its JSDoc says "Not a gate" |
| T5 | `onRefetchActivities` did not refetch activities | Fixed (`useActivationState.ts:98-103`), and the new test covers it |
| T6 | Review amounts flashed to 0 after Confirm | Fixed: the review inputs are pinned at confirm and released on failure (`WithdrawFlow/index.tsx`) |
| T7 | Root-key `refetchQueries` widened the scope | Fixed: `type: "active"` (`PendingVaultsContext.tsx:188-191`) |
| T8 | `e2e/real/actions/withdraw.ts` still mentioned `isActivating` | Fixed (`lifecycle === "activating"`) |
| T9 | Comment and test name still said "indexer zeroes" after the rebase | Fixed in `0742d9df` |

### Notes

- **Where "withdrawing" comes from.** The indexer's `redeemed` status plus the localStorage marker, never the chain. Both ways it can be wrong fail safe:
  - A stale `redeemed` on a vault still on chain disables Withdraw and makes the liquidation total understate collateral, which overstates risk. It also makes the count disagree with `position.vaultIds`, which raises `indexerError` on `/vaults` and `/loans`.
  - Indexer lag that hides a real withdrawal leaves Withdraw enabled, and a repeat withdraw reverts on-chain. That is the same as before this PR.
  - Every financial gate stays chain-only.
- **Revalidation after activation.** `invalidateVaultQueries` together with `["user-activities"]` covers every per-user key the activation changes.
- **E2E hooks.** All five testids used by `e2e/real/actions/withdraw.ts` exist on live controls at head.

services/vault/src/components/pages/VaultsPage.tsx
```diff
@@ -52,6 +53,10 @@ export default function VaultsPage() {
   const deposits = usePendingDeposits();
   const { isLoading, isEmpty, hasError, hasPartialError } =
     useVaultsPageEmptiness(deposits);
+  const { vaults: aaveVaults } = useAaveVaults(
+    isConnected ? address : undefined,
+  );
+  useSyncPendingVaults(aaveVaults);
```
jonybur, line 59: **useSyncPendingVaults on /vaults can clear a withdraw marker while the vault list is still loading or errored, reverting the row to active and stopping polling.**

useSyncPendingVaults treats a withdraw as confirmed as soon as the vault id is missing from useAaveVaults().vaults (PendingVaultsContext.tsx:177), but that list is an empty array while useVaults is loading or has errored. allVaults also gets a new empty array reference when usePrice("BTC") goes from 0 to the loaded price, so if prices resolve before vaults (or the vaults query fails), the effect fires with the marker already loaded from localStorage and clears every pending withdraw. With this PR the marker drives the withdrawing row via applyPendingWithdrawals, so reloading /vaults in the mined-to-indexed gap flips the row back to active with Withdraw enabled, and because poll depends on hasPendingOperations the vaults query also stops polling, so the row can stay wrong until something else refetches. Skip the sync until the underlying vaults query has data and no error, for example by returning isLoading/error from useAaveVaults and bailing in the effect.

jonybur (commented):
Comments inline.

PR #2491, opened 2026-09-09
fix(wallet): require confirmed consent for Vault and staking

jonybur:
## Outcome

Vault pages and BTC staking actions require confirmed wallet consent. Saved Retry actions stop working after consent loss or unmount.

Commit `e2ffb811` contains the full change in one signed commit on current main. All local checks pass. Fresh GitHub checks and Greptile are pending. All 15 earlier review threads are resolved.

## Change

- Require confirmed consent for Vault access and the BTC staking manager.
- Revoke saved actions when consent ends. Later confirmation permits fresh actions only.
- Keep wallet warnings visible before consent. Consent controls access; live wallet connections control read-only warnings.
- Restore staking TSX test discovery. Repair stale tests. Remove tests for the deleted multistaking flag and registration subscription.
- Remove unused mocks for the deleted registration subscription.

## Safety

A required-wallet disconnect clears consent. The user must select Connect again. Required wallet sets and session reset behavior stay in place.

The saved Retry recovery work remains in #2506. The selected consent rule is recorded in #2354. This PR adds no consent decision document or release note. Legal approval is not claimed.

## Verification

- Local builds pass for Vault, staking, the wallet package, and their dependencies.
- Unit tests pass: 3,897 Vault tests, 515 staking tests, and 371 wallet tests. Three existing skipped tests are unchanged. Wallet package checks also pass.
- All 42 browser tests pass: 23 Vault, 16 staking, and three wallet tests.
- Lint passes in all three affected projects. One local staking lint run hit a dependency build-output race; it passed after the build output was stable.
- The full PR patch on current main matches Git's clean merge result. The only further change removes four unused mock lines.
- The previous `ddc1be3e` commit passed CI and received Greptile 5/5. New results must cover `e2ffb811`.

## Links

Tracks #2354. #2354 stays open for the remaining completion conditions.

- Wider live wallet checks: #2355
- Saved Retry recovery: #2506
- Parent: #2230 and #2228


services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -77,16 +77,14 @@ const DEPOSIT_DISABLED_BANNER_Z_CLASS = "z-30";
 export default function RootLayout() {
   const gate = useProtocolGateState();
   const { theme, setTheme } = useTheme();
-  const { connected: btcConnected } = useBTCWallet();
-  const { connected: ethConnected } = useETHWallet();
+  const { isConnected: isWalletConnected } = useConnection();
```
jrwbabylonlab: This signal also gates the compliance banners. `AddressScreeningBanner`, `AddressTypeBanner` and `DepositDisabledBanner` were previously gated on raw `btcConnected && ethConnected`; they now additionally require a confirmed receipt. So a blocked or unsupported BTC address with both wallets connected but no valid receipt (fresh install, cleared storage) sees none of the three. No test pins banner visibility against `confirmed`.

services/vault/src/context/wallet/useConnection.ts
```diff
@@ -1,31 +1,26 @@
...
- */
+// Consent contract: docs/decisions/2354-wallet-consent.md.
 export function useConnection(): ConnectionState {
+  const { connected: confirmed } = useWalletConnect();
   const { connected: btcConnected } = useBTCWallet();
   const { connected: ethConnected } = useETHWallet();
 
   return {
-    isConnected: btcConnected && ethConnected,
+    isConnected: confirmed && btcConnected && ethConnected,
```
jrwbabylonlab, line 22: A transient BTC disconnect now costs the whole shell, not just the header menu. The connector's required-chain handler deletes `WALLET_CONFIRMATION_RECEIPT_KEY` immediately (`useWalletConnectors.tsx:229-246`), outside the 3s debounce in `VaultWalletConnectionProvider.tsx:71-91` — that debounce only delays `disconnectAll`. BTC reconnects, but the receipt is gone and nothing re-confirms, so RootLayout flips to the entry layout mid-session until the user presses Connect again.

services/simple-staking/tests/context/wallet/WalletConnectionProvider.test.ts
```diff
@@ -0,0 +1,115 @@
+/** @jest-environment jsdom */
```
jrwbabylonlab: The `.ts` rename works around a broken config rather than fixing it. `services/simple-staking/jest.config.js` declares `projects` twice (lines 117 and 164); JS keeps the second, node-only definition, whose `testMatch` excludes every `.test.tsx` — 13 suites that currently never run. This file also lands in that node project, where `setupFiles` is empty and only the file-level jsdom directive rescues the environment. Deleting the duplicate key would fix all 13.

services/vault/src/context/wallet/useConnection.ts
jonybur: The report is correct. The three-second delay protects wallet teardown. It does not preserve consent after a required wallet disconnect.

The shared connector deliberately clears that approval. Existing tests require confirmation to end after a required provider disconnects. They also prevent approval from returning when a disconnect interrupts validation. See [the consent rule](https://github.com/babylonlabs-io/babylon-toolkit/blob/91c47c4d662e3170755165c63326ba1f2f8dc957/docs/decisions/2354-wallet-consent.md) and [the disconnect tests](https://github.com/babylonlabs-io/babylon-toolkit/blob/91c47c4d662e3170755165c63326ba1f2f8dc957/packages/babylon-wallet-connector/src/hooks/__tests__/useWalletConnectors.requirementChurn.test.tsx).

Automatic approval after reconnect would change that rule. This PR keeps actions blocked until the user selects Connect again. This thread stays open for reviewer clearance.

services/simple-staking/tests/context/wallet/WalletConnectionProvider.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  it("keeps action gates closed until confirmation or verified restore", () => {
+    const { result, rerender } = renderHook(() => ({
+      auth: useAuthGuard().connected,
+      loggedIn: useIsLoggedIn(),
+    }));
+    expect(result.current).toEqual({ auth: false, loggedIn: false });
+    (useWalletConnect as jest.Mock).mockReturnValue({ connected: true });
+    rerender();
+    expect(result.current).toEqual({ auth: true, loggedIn: true });
```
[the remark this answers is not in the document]
jonybur: Fixed in `8615fc7786d2f3a2b7fd48fe3dfe1757c2bd6e0d`. The BTC staking manager now requires the confirmed wallet session before it can create a staking manager. The EOI regression uses the real transaction and staking-manager hooks. It checks that live wallets cannot sign before consent, can sign after consent, and cannot start another action after consent is lost. The full staking suite passes: 515 tests, with one existing skip.

services/vault/src/components/pages/RootLayout.tsx
jonybur: Fixed in `8615fc7786d2f3a2b7fd48fe3dfe1757c2bd6e0d`. The three wallet warnings use the live BTC and ETH connections. The page layout still requires confirmed consent. The RootLayout tests check address screening, address type, and disabled-deposit warnings with consent both false and true. All 39 scoped Vault tests pass.

services/simple-staking/tests/context/wallet/WalletConnectionProvider.test.ts
jonybur: Fixed in `8615fc7786d2f3a2b7fd48fe3dfe1757c2bd6e0d`. Jest now has one projects setting with both node and jsdom suites. The provider test is back in a .tsx file and uses the shared setup. The restored suites exposed stale imports and assertions; those are fixed. Tests for the deleted multistaking flag and removed registration subscription were removed. All 57 current suites pass: 515 tests, with one existing skip.

jonybur:
@greptileai

Please review the full PR at current commit `8615fc7786d2f3a2b7fd48fe3dfe1757c2bd6e0d` against `main`. The existing 4/5 score covers the prior commit.

The update guards the real BTC staking manager with confirmed consent. Its EOI regression covers consent absent, accepted, and lost. Vault warnings stay visible before consent. Jest now runs the restored UI suites; all 57 current staking suites pass locally.

Please check the current code and report any remaining findings with a confidence score for this commit. The required-wallet disconnect rule stays in place, as explained in the open review thread.

services/simple-staking/src/ui/common/hooks/services/useStakingManagerService.ts
```diff
@@ -45,6 +47,7 @@ export const useStakingManagerService = () => {
   const { chainId } = getNetworkConfigBBN();
 
   const isLoading =
+    !confirmed ||
```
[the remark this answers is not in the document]
jonybur: Fixed in `10e82d53ecd8d81e4f33552c2390c2fe3c6f6346`. Each confirmed consent period now has its own revocable state. Consent loss and unmount permanently revoke old callbacks. Confirmation creates fresh callbacks and does not restore an old retry. The real EOI service regression retains the callback and checks consent loss, later confirmation, and unmount under React StrictMode. Fresh callbacks work after confirmation. All 57 staking suites pass: 515 tests, with one existing skip.

jonybur:
@greptileai

Please review the full PR at current commit `10e82d53ecd8d81e4f33552c2390c2fe3c6f6346` against `main`. The previous 3/5 review covers `8615fc7786d2`.

The Retry fix revokes saved actions when consent ends or the component unmounts. Later confirmation permits fresh actions and does not restore an old callback. The real staking service regression checks these cases under React StrictMode. All 57 staking suites pass: 515 tests, with one existing skip.

Please report any remaining findings and a confidence score for this commit. The required-wallet disconnect rule stays in place, as explained in the open review thread.

services/simple-staking/src/ui/common/hooks/services/useStakingManagerService.ts
```diff
@@ -12,6 +13,14 @@ import { useAppState } from "@/ui/common/state";
 import { getTxHex } from "@/ui/common/utils/mempool_api";
 
 export const useStakingManagerService = () => {
+  const { connected: confirmed } = useWalletConnect();
+  const consent = useMemo(() => ({ confirmed }), [confirmed]);
```
gbarkhatov, line 17: **MERGE BLOCKER — this latch permanently poisons any retained callback; a stored Retry action is bricked forever even after consent returns.**

`consent` is a `useMemo` object recreated per `confirmed` value, and the `useLayoutEffect` cleanup writes `false` into the **previous** object on every transition. So a callback captured before a consent flicker reads `false` forever — re-consent creates a *new* object rather than healing the old one.

This isn't theoretical. `useStakingService.ts:184` stores `retryAction: () => stakeDelegation(delegation)` into error state, and `ErrorModal.tsx:37` invokes that stored closure later. Concrete failure: stake fails → Retry modal appears → any `confirmed` blip → every press of Retry throws `"BTC Staking Manager not initialized"` permanently, with a healthy confirmed wallet, recoverable only by remount.

The failure *direction* is safe — a signer never survives consent loss — but it's still a user-facing brick. A ref reads live in both directions, fails closed on loss, and self-heals on return: `const consentRef = useRef(confirmed); consentRef.current = confirmed;` read at call time. Also worth noting line 19 (`consent.confirmed = confirmed` in the effect body) is a no-op — the memo initializer already set it.

One interaction to be aware of: `useTransactionService.test.tsx:275-283` currently asserts the stale action rejects *even after* `confirmed` returns to `true` — so it pins this behaviour as the intended contract, and fixing this means changing that assertion.

services/simple-staking/jest.config.js
```diff
@@ -173,18 +160,18 @@ const config = {
...
-    //   testMatch: ["<rootDir>/tests/**/*.test.tsx"],
-    //   setupFiles: ["./jest.jsdom.setup.js"],
-    //   moduleNameMapper: {
-    //     "^@/(.*)$": "<rootDir>/src/$1",
-    //     "\\.(png|jpg|jpeg|gif|webp|avif|bmp)$":
-    //       "<rootDir>/tests/__mocks__/svgMock.js",
-    //     "\\.svg$": "<rootDir>/tests/__mocks__/svgMock.js",
-    //   },
-    // },
+    {
```
gbarkhatov, line 163: **MERGE BLOCKER — and this hunk contains the most consequential fact in the PR, currently invisible in the description.**

I checked the base commit: `jest.config.js` had **two** `projects:` keys, at lines 117 and 164. The later one wins in a JS object literal, and in it the jsdom project was commented out — so **no `.test.tsx` in simple-staking was running at all.** That's the real reason nine test files needed mock-path repair. This PR correctly deletes the shadowed key and re-enables jsdom, which is a genuinely valuable fix that deserves to be stated.

But it changes `setupFiles` from `./jest.jsdom.setup.js` — which *both* base blocks named — to `./jest.setup.js`, and the two are not equivalent. The retired file installs an MSW `setupServer` with blanket `get/post/put/delete/patch("*")` handlers plus lifecycle hooks, exposes `global.mswServer`, and shims `BroadcastChannel`. `jest.setup.js` has none of that. I grepped the worktree at HEAD: **zero references to `jest.jsdom.setup` remain, and the 1825-byte file is still present** — a flat ZERO DEAD CODE violation, which is what makes this blocker-eligible on rule grounds rather than severity, since CI is green.

Either point the jsdom project at `jest.jsdom.setup.js`, or fold its MSW and `BroadcastChannel` setup into `jest.setup.js` and delete the loser.

**Separately, and I can't anchor it inline because it's a deletion-only hunk:** `tests/utils/featureFlag/MultiStaking.test.tsx` is deleted outright — 74 lines, 0 insertions — in a commit titled "restore consumer test coverage". It held the only assertion that `FeatureFlagService.IsMultiStakingEnabled` makes Home render the multistaking form rather than `StakingForm`. `MultistakingFormWrapper` is still live production code (`src/ui/common/page.tsx:12,37`), and `grep -rln "IsMultiStakingEnabled" services/simple-staking/tests/` now returns nothing. The likely cause is this same config fix: the suite was dormant, re-enabling it surfaced a failure, and it was deleted rather than repaired. That's invisible in a passing CI run, which is why it's worth calling out — repair and re-land it, or say in the PR body why the coverage is being dropped.

services/vault/src/context/wallet/useConnection.ts
gbarkhatov, line 22: Adding `confirmed` here turns a transient BTC-extension blip into a permanent, app-wide ejection.

I traced the whole chain: `useWalletConnectors.tsx:247-250` handles a required-chain `disconnect` by synchronously deleting `WALLET_CONFIRMATION_RECEIPT_KEY` and calling `removeWallet`, which sets `confirmed: false` (`State.context.tsx:181`). The vault's own `BTC_DISCONNECT_DEBOUNCE_MS = 3000` only defers `disconnectAll()` — it cannot cover the connector-level receipt deletion.

**Before this PR the blip self-healed**: `btcConnected` returned within the window and `isConnected` recovered. Now `confirmed` stays false and only a manual Connect restores it, dropping the user to the entry layout on every reload where the extension is slow to wake. The repo's own 17-line comment on that debounce, and the 1500→3000 ms bump, are evidence this happens in practice.

The root cause sits in wallet-connector, outside this diff, so the fix is a follow-up — but this PR is what extends the blast radius from the navbar to every page, so the decision belongs here.

Related, same hook: `ConnectionState` returns only `isConnected`/`btcConnected`/`ethConnected`, so no consumer can distinguish "no wallet" from "wallets live, consent unconfirmed" — three mutually exclusive states collapsed into one boolean, where CLAUDE.md asks for a discriminated union. Your own decision doc depends on that distinction: it says an expired approval should tell the user to press **Connect** again, but all nine consumers currently render the identical generic state. It also forces `RootLayout.tsx:81` to re-derive presence inline as `isWalletConnected = btcConnected && ethConnected` — while `useDepositPageForm.ts:217` binds that *same identifier name* to the consent-inclusive value. Both are correct today, but one spelling now means two different things in one app with no type or lint signal.

services/simple-staking/src/ui/common/hooks/services/useStakingManagerService.ts
```diff
@@ -55,16 +65,8 @@ export const useStakingManagerService = () => {
     versionedParams.length === 0;
 
   const createBtcStakingManager = useCallback(() => {
```
gbarkhatov, line 67: The diagnostic log was gutted from seven fields to one, on a path with eight non-consent failure causes. All three review lanes flagged this independently.

The old log emitted `cosmosConnected`, `btcConnected`, `btcNetwork`, `signPsbt`, `signMessage`, `signBbnTx` and `versionedParams`; it now emits only `{ confirmed }`. But `isLoading` is still a nine-term disjunction (lines 56-65), so the common failures — `versionedParams` empty, `signPsbt` absent — now log `confirmed: true` and nothing else, which actively reads as "consent was fine". And a stale-closure rejection logs `confirmed: false`, indistinguishable from "user never consented". Every call site throws the same fixed `"BTC Staking Manager not initialized"`.

That's a clear Error Handling regression against CLAUDE.md's "messages must be actionable", and it's unrelated to this PR's purpose — keeping the seven fields and adding `confirmed` as an eighth would cost nothing.

services/simple-staking/src/ui/common/hooks/services/useStakingManagerService.ts
```diff
@@ -45,6 +54,7 @@ export const useStakingManagerService = () => {
   const { chainId } = getNetworkConfigBBN();
 
   const isLoading =
+    !confirmed ||
```
gbarkhatov, line 57: Route-driven `requiredChains` churn now disables every delegation action with no recovery affordance.

Staking's `requiredChains` is route-derived (`/baby` → `["BBN"]`, else `["BTC","BBN"]`), and `State.context.tsx:98,118` sets `confirmed: requirementsExpanded ? false : state.confirmed`. So confirming on BTC staking, visiting `/baby`, then navigating back re-widens the set and unconfirms the session. That was previously invisible to the staking manager; adding `!confirmed ||` here makes `isStakingManagerReady` false and disables delegation actions on a plain back-navigation.

Comparing against the confirmed receipt's chain set rather than the last-seen requirement array would avoid it — or surfacing an explicit "confirm again" prompt instead of an indefinite disabled state.

services/vault/src/components/pages/RootLayout.tsx
```diff
@@ -77,21 +77,20 @@ const DEPOSIT_DISABLED_BANNER_Z_CLASS = "z-30";
 export default function RootLayout() {
   const gate = useProtocolGateState();
   const { theme, setTheme } = useTheme();
-  const { connected: btcConnected } = useBTCWallet();
-  const { connected: ethConnected } = useETHWallet();
+  const { isConnected, btcConnected, ethConnected } = useConnection();
```
gbarkhatov, line 80: The entry frame can now render wallet-specific warning banners. `isEntryLayout` uses the confirmed gate while the address-type, screening and deposit-disabled banners still use presence — before this PR they were one variable, so the entry frame could never carry a wallet banner.

Now "both wallets live, consent unconfirmed, pathname `/`" renders the disconnected chrome *and* the taproot warning together: "you are not connected" beside a warning about the connected wallet's address type.

One lane raised this and another refuted it as fail-safe, and both are right — it errs toward showing more warnings, and your decision doc sanctions presence-driven read data. The new test at `RootLayout.test.tsx:272` pins it deliberately, so I'm raising it as a decision to confirm rather than a bug to fix. If it's intended, it may be worth a note in the doc, since the copy reads incoherently.

docs/decisions/2354-wallet-consent.md
```diff
@@ -0,0 +1,26 @@
+# Wallet consent rule
+
+Status: jonybur authorized this rule on 9 September 2026 for [#2354](https://github.com/babylonlabs-io/babylon-toolkit/issues/2354). The user overrode the issue's prior Product and Legal approval requirement and will contact Legal separately.
```
gbarkhatov: This names an individual and records a legal-approval override in permanent repo history. Verbatim: *"Status: jonybur authorized this rule on 9 September 2026 … The user overrode the issue's prior Product and Legal approval requirement and will contact Legal separately."*

Two problems. Repo markdown should attribute to the artifact rather than a person, and "The user" is conversational framing that doesn't belong in a checked-in record. More substantively, the override claim belongs on the issue thread where Legal can see and respond to it — recording it here silently blesses the bypass in a document Legal is unlikely to read.

Something like *"Authorized on 9 September 2026 for #2354"* would keep the fact without either problem.

services/vault/src/components/pages/__tests__/RootLayout.test.tsx
```diff
@@ -31,28 +31,25 @@ vi.mock("@/config", () => ({
...
-// hosts the `useGeoFencing` export consumed here) imports `@/config/wagmi`
-// at module scope, which imports `@babylonlabs-io/wallet-connector` — and
-// that package's build cannot be transformed by Vitest in this workspace
-// (see the wallet-connector mock below), so the real module can't even be
-// loaded, not just "unsafe to render". Mocked here to return the exact same
-// default the real context has (`isLoading: true`, so RootLayout stays on
-// its Loader branch and the content-branch providers/components below it —
-// AaveConfigProvider, ActivatingVaultsProvider, SimpleDeposit, GeoBlockState,
-// ProtocolStatusBanner — never mount, matching the real unmocked behavior).
 vi.mock("@/context/geofencing", () => ({
```
gbarkhatov, line 42: A 13-line rationale comment was deleted here while the mock it explains was kept. The still-valid half explains why `@/context/geofencing` must be mocked — it imports `@/config/wagmi` → `wallet-connector`, which Vitest can't transform in this context — so without it the next contributor will unmock it and hit an opaque failure.

Smaller, same file: this PR collapses four entry-layout cases into one `it.each`, where CLAUDE.md says "three similar test functions are better than one parameterized abstraction". I checked and **no assertion was lost**, so this is style rather than coverage. The sharper instance is `useTransactionService.test.tsx:227-297` — a single 62-line `it` covering five behaviours with a shared mutable mock and a trailing `toHaveBeenCalledTimes(2)` coupling them all, so a failure anywhere gives no signal about which broke.

gbarkhatov (dismissed):
Reviewed at `10e82d5` via three independent passes, deduped, with every finding re-checked against the source and against the base commit.

**The consent gate itself is well-shaped, and no bypass exists.** Routing all nine vault consumers through `useConnection` is the right move — one edit propagated the gate with no per-site changes, and no vault production file still hand-rolls `btcConnected && ethConnected` as a stand-in. In staking, every `createBtcStakingManager()` call site funnels a `null` manager into `validateStakingManagerInputs`, which throws `ClientError(INITIALIZATION_ERROR)` — genuinely fail-closed, never a silent no-op.

**Identity binding is intact.** I verified `State.context.tsx:181` sets `confirmed: required ? false : state.confirmed` and `useWalletConnectors.tsx:247-250` deletes `WALLET_CONFIRMATION_RECEIPT_KEY` on a required-chain disconnect. BTC is required for vault, so a wallet swap re-establishes the binding rather than inheriting it — the property #2499 relies on holds, and this PR doesn't touch it. That same verified mechanism is what makes my comment on `useConnection.ts` real, though: the binding is correct, and this PR extends its blast radius.

**Refactor vs behaviour, since you asked for it separated.** Runtime behaviour changes are confined to four places: `useConnection.ts:23`, `RootLayout.tsx:80-93`, `useStakingManagerService.ts` (latch, gate, log), and — disguised as config — `jest.config.js`. Everything else in the +459/−458 is movement: the `@/ui/legacy/*` → `@/ui/common/*` mock-path repair across nine staking test files, Prettier reflows, and two provider files that are comment-only.

- *"A consent receipt can survive a wallet/identity change."* Refuted with code, as above.
- *"`createBtcStakingManager()` returning `null` is a silent no-op."* Refuted — all call sites throw via `validateStakingManagerInputs`.
- *"A stale memoized staking manager could be reused after consent loss."* Refuted — no call site stores an instance; every operation re-creates one.
- *"The deposit dialog stays open across consent loss, so signing is reachable."* Refuted — the CTA gate reads the consent-inclusive value via `useDepositPageForm.ts:217`. A UX wrinkle, not a bypass.
- *"The parameterized `WalletDialog` suite lost the 'first required chain, not first connected wallet' property."* Refuted — `harness.connectors` is seeded BTC/ETH/BBN, so the `["ETH"]` and `["BBN"]` cases still fail if insertion order were used. Net stronger than what it replaced.
- *"This PR touches a CLAUDE.md critical path."* It doesn't.

**Verified clean:** the gate works end to end at this SHA with no dependency on sibling PR #2499 — `confirm()` at `WalletDialog.tsx:377` and receipt-restore at `useWalletConnectors.tsx:320` are both present. `cosmosConnected`/`btcConnected` were correctly dropped from the `useCallback` deps, since only the removed log referenced them. The `subscribeToSigningSteps` test-block deletion is correct — that API no longer exists in `src/`. No user-facing string changed, no motion, no SDK public surface.

**Adjacent, not this PR:** `Connect.tsx:68-70` still hand-rolls `confirmed && btcConnected && ethConnected` instead of consuming the new hook — drift risk worth a follow-up. And `useV1TransactionService.test.tsx`, `useDelegationStorage.test.tsx` and `BalanceState.test.tsx` are newly live under the fixed jest config without having been reviewed as part of this PR.

services/vault/src/context/wallet/useConnection.ts
```diff
@@ -1,31 +1,26 @@
...
   ethConnected: boolean;
 }
 
-/**
- * Hook to get combined wallet connection state
- * @returns Connection state for both wallets
- */
+// Consent contract: docs/decisions/2354-wallet-consent.md.
 export function useConnection(): ConnectionState {
+  const { connected: confirmed } = useWalletConnect();
```
jeremy-babylonlabs, line 17: Staking gates the signer (`createBtcStakingManager` returns null when `!consent.confirmed`), but vault only gates pages and CTAs.

`unconfirm()` fires on `accountsChanged` / `chainChanged` for a required chain while `btcConnected` / `ethConnected` stay true, so a reclaim/refund/payout modal already mounted on Activity or Vaults can still reach `signPsbt` with the new, un-consented account. Today that is caught only by the per-deposit depositor-address checks, not by consent.

Can we read `useWalletConnect().connected` at the vault signing seam (the hooks that own `signPsbt` / `sendTransaction`) so both apps enforce the same rule?

services/simple-staking/jest.config.js
```diff
@@ -173,18 +160,18 @@ const config = {
...
+      displayName: "jsdom",
+      testEnvironment: "jsdom",
+      testMatch: ["<rootDir>/tests/**/*.test.tsx"],
+      setupFiles: ["./jest.setup.js"],
```
jeremy-babylonlabs, line 167: `jest.jsdom.setup.js` is orphaned after this switch. Delete it — or, if the MSW server / `BroadcastChannel` shim it provided is needed by any of the newly-live `.tsx` suites, point `setupFiles` back at it.

jeremy-babylonlabs (commented):
Consent gate itself is fail-closed and correctly scoped: default `confirmed: false`, receipt keyed by wallet/address/pubkey/network per chain, deleted on required-chain disconnect or identity change. No fresh-state bypass found, and the un-consented path is asserted to block in both apps.

Good catch on the `jest.config.js` duplicate `projects` key — that had been silently disabling every `.test.tsx` suite in simple-staking.

One asymmetry between the two apps is worth resolving before this lands.

services/simple-staking/jest.config.js
jonybur: Deleted services/simple-staking/jest.jsdom.setup.js in 067127e7. Nothing referenced it once the jsdom project moved to jest.setup.js. The suite passes without the MSW server and the BroadcastChannel shim. msw stays in devDependencies: the Playwright e2e mocks import it (e2e/mocks/handlers.ts imports setupServer from msw/node, and the six files under e2e/mocks/handlers/ import http and HttpResponse from msw). On MultiStaking.test.tsx: the test asserted the IsMultiStakingEnabled flag, which FeatureFlagService no longer has, and two module paths that no longer exist under src/ (components/Staking/StakingForm and components/Delegations/Activity). CLAUDE.md says not to keep tests for code that does not exist in production. The file stays deleted. I will say so in the PR body.

services/simple-staking/jest.config.js
jonybur: Same fix as the thread at jest.config.js:163. jest.jsdom.setup.js is deleted in 067127e7. The jsdom project keeps jest.setup.js.

services/simple-staking/src/ui/common/hooks/services/useStakingManagerService.ts
jonybur: Restored the seven fields in 067127e7 and added confirmed as the first field. A failure now logs which input was missing, not only the consent flag. cosmosConnected and btcConnected are back in the useCallback dependency list for that reason.

services/vault/src/components/pages/__tests__/RootLayout.test.tsx
jonybur: Restored the still-valid part of the comment above the geofencing mock in 067127e7. It keeps the two facts the next contributor needs: the real module imports @/config/wagmi, which imports @babylonlabs-io/wallet-connector, which Vitest cannot transform here; and the mock returns isLoading: true so RootLayout stays on its Loader branch. I dropped the sentence about addressScreening and addressType. Both are mocked in this file now, so that sentence was false. On splitting the consent test in useTransactionService.test.tsx: I am leaving it as one test. Each of the five behaviours needs the same mount, confirm, and capture-the-retained-closure sequence. Five independent tests add about 40 lines of repeated setup and pin the same call count in five places. One test keeps one call-count assertion for the whole sequence, and its name says what it verifies.

docs/decisions/2354-wallet-consent.md
jonybur: Replaced the Status line in 067127e7. It now reads: Status: Authorized on 9 September 2026 for #2354. The override statement is out of the repo. I will post it as a comment on #2354 where Legal can see it.

services/vault/src/context/wallet/useConnection.ts
jonybur: Renamed the RootLayout local to walletsPresent in 067127e7: the declaration at :81 and the three banner use sites at :95, :107 and :179. Presence and consent no longer share a name. On the discriminated union: ConnectionState does not collapse the three states. useConnection.ts:22-26 returns isConnected, btcConnected and ethConnected as separate fields. RootLayout.tsx:81 derives presence from the last two, and the entry-frame gate at :93 reads isConnected. A consumer can tell 'no wallet' (btcConnected or ethConnected false) from 'wallets live, consent unconfirmed' (both true, isConnected false), so no union is needed. The blip half of this comment is the same point as the earlier thread on this line; my answer is there.

jonybur: No code change in this PR. The regression is real and accepted for now. This PR widens the effect of a transient BTC disconnect from the wallet menu to the whole shell. The cause is in wallet-connector, not in the vault: all four receipt-wipe paths treat a transient required-chain disconnect as a real one. They are the connector disconnect handler at useWalletConnectors.tsx:266, removeWallet at State.context.tsx:171-181, the live validation at useWalletConnectors.tsx:326, and the identity watch at useWalletConnectors.tsx:380-389, which subscribes to the BTC provider's own disconnect event (confirmationReceipt.ts:29). None can change here without changing shared connector semantics and the tests that pin them: requirementChurn.test.tsx:424-434 'ends the confirmation after a required provider disconnects' pins the identity-watch path, and :486-509 'does not restore approval after a required disconnect interrupts validation' pins the connector handler. #2354 stays open for the follow-up. It must give a required-chain disconnect a bounded reconnect window before any of the four paths clears the receipt, then add requirementChurn cases for disconnect-then-reconnect with the same identity (re-confirms) and with a different identity (receipt deleted). Removing one of the paths alone changes no user-visible behaviour, so I am not doing that as a partial measure.

services/simple-staking/src/ui/common/hooks/services/useStakingManagerService.ts
jonybur: No code change. Line 19 is not a no-op. Under StrictMode the effect runs setup, cleanup, setup on the first commit. The cleanup sets consent.confirmed to false and the second setup restores it. Without line 19 the retained object would stay false after the remount. The test at useTransactionService.test.tsx:256-263 wraps the hook in StrictMode for that reason. I am not moving to a live ref. The assertions at :277-283 pin that a retained action rejects after a confirmed false-to-true cycle, and a render-time ref is not safer: a discarded render leaves the stamp in place. The UX point behind the comment stands. After a consent flicker a stored Retry throws 'BTC Staking Manager not initialized' with a healthy wallet. The right fix is to clear retryAction from the error state on consent loss, so a dead button is never offered. That is a change in ErrorProvider, not in this hook. The thread at useStakingManagerService.ts:57 shows a plain back-navigation produces exactly that cycle, so the follow-up is worth filing.

services/vault/src/components/pages/RootLayout.tsx
jonybur: The split is deliberate. Consent gates access. Wallet presence supplies read-only warnings. Hiding a taproot or screening warning until consent lands would hide a real problem from a user whose wallet is connected. I added one sentence under Decision in docs/decisions/2354-wallet-consent.md in 067127e7 to make the entry-frame case explicit. If design review finds the pairing of entry chrome and a wallet warning incoherent, that is a copy.ts change to the entry-frame text, not a change to this gate.

services/simple-staking/src/ui/common/hooks/services/useStakingManagerService.ts
jonybur: No code change. The session re-confirms itself from the surviving receipt. When the required set widens and the stored receipt still covers it, useWalletConnectors.tsx:407 calls unconfirm(true), which keeps the receipt. The next validation pass then calls confirm(confirmationCandidate) at :332. requirementChurn.test.tsx:167-180 'restores an unchanged approval after the required set widens' pins that. So confirm on BTC staking, visit /baby, and return gives a short confirmed false-to-true cycle, not an indefinite unconfirmed state, and the actions come back. Comparing against the confirmed receipt's chain set instead of the last-seen requirement array would weaken State.context.tsx:94-100 and :107-121, the guard that stops a widened requirement set from inheriting a narrower approval. I am not adopting it. The cycle does matter for the thread at useStakingManagerService.ts:17: it is the transition that permanently revokes a retained callback there, which is the case for clearing retryAction on consent loss.

services/vault/src/context/wallet/useConnection.ts
jonybur: I traced this. The mechanism does not hold for the vault: a new account on a required chain does not leave btcConnected and ethConnected true.

1. On a real BTC account switch, BTCWalletProvider.tsx:255-273 sees a new address and awaits callbacks.onAddressChange. The vault wires that to runWalletReset for BTC and ETH (VaultWalletConnectionProvider.tsx:198 and :220). runWalletReset calls disconnectAll (:135). That calls connector.disconnect("all") on every connector (useWalletConnect.ts:80-84), which sets _connectedWallet to null (WalletConnector.ts:75) and emits disconnect. A new ETH account takes the same path (ETHWalletProvider.tsx:209).

2. The signing hooks read the wallet from the connector, not from state captured at mount: useReclaimState.ts:40-42 and useRefundState.ts:36-38 read it on every render, and usePayoutSigningState.ts reads the connector at :132 and the wallet fields inside the handler at :207-208 and :243. After the teardown btcWalletProvider is undefined, and each handler refuses before any wallet call (useReclaimState.ts:78-81, useRefundState.ts:89-92, usePayoutSigningState.ts:244-247).

3. The async window between the raw accountsChanged event and the teardown is covered. connectedWallet.account is set once, inside Wallet.connect (Wallet.ts:58), and nothing else writes it, so connectedBtcAddress in the hooks stays the consented address. Every path calls verifyBtcWalletLiveness with it before signing (useReclaimState.ts:98-106, useRefundState.ts:153-161, usePayoutSigningState.ts:272-276). It throws addressMismatch when the wallet's live getAddress differs (verifyBtcWalletLiveness.ts:101-103). Payout also compares that address to the vault's registered payout scriptPubKey (usePayoutSigningState.ts:203-216).

What is true in the comment: nothing closes a mounted modal on unconfirm alone, and no vault signing hook reads consent. Consent goes false while both wallets stay live in three cases: an ETH chainChanged, a BTC networkChanged (confirmationReceipt.ts:29 lists it for BTC and BTCWalletProvider.tsx has no handler for it), and a BTC accountsChanged that repeats the same address (the same-identity short circuit at confirmationReceipt.ts:251-261 is ETH-only). In all three, the BTC key that signs is the identity the user confirmed. No un-consented account can reach signPsbt. I am not adding a consent check inside the signing hooks in this PR. The vault action gate lands in #2482 (useBtcAction). A signing-time consent check belongs with it under #2232, so the gate and the hooks follow one rule.

jonybur:
Review fixes are pushed as `067127e7`, on top of a merge of main.

Seven threads produced changes (two of them the same deletion), four are answered with the reasoning in the thread. The one deletion is the orphaned `jest.jsdom.setup.js`; `msw` stays because the Playwright e2e mocks import it.

On the newest thread, the un-consented signing claim: traced and refuted for the vault, with the narrow residual named in the thread and left for #2232.

Local checks at this head: 3,861 Vault tests in 315 files, 515 simple-staking tests in 57 suites, eslint clean on every edited file.

jonybur:
@greptileai review the full PR at current commit `067127e718b0772a2c201f0660788cb7c624e3cf`. Check the diff against its target branch, including the changes since your last review. Verify that the review findings are addressed in the code. Report any remaining issues and the confidence score for this commit.

Retry usability is deferred in #2506. Consent revocation must remain.

gbarkhatov (dismissed):
Follow-up to my review at `10e82d53ec`. Main has been merged in since then, so I compared the PR's own net patch at both commits. Nothing new in the delta.

| # | Prior finding | Verdict at `067127e7` |
|---|---|---|
| T1 | The consent latch permanently breaks a saved Retry after any consent flicker | **Declined, deferred to #2506.** Part of my point was wrong: line 19 is not a no-op. Under StrictMode the first commit runs setup, cleanup, setup, and only line 19 turns `confirmed` back on. The regression test mounts under `<StrictMode>` |
| T2 | `jest.jsdom.setup.js` orphaned; `MultiStaking.test.tsx` deleted | Setup file **fixed** (deleted; nothing references it). The MultiStaking half was **wrong**: `IsMultiStakingEnabled` no longer exists in `src/`, so the test pinned a branch that is gone |
| T3 | A Bitcoin blip ejects the user app-wide; one name meaning two things | Name clash **fixed** (`walletsPresent` for banners, `isConnected` for access, `RootLayout.tsx:80-95`). The blip regression and the union are **declined**; #2354 stays open |
| T4 | Diagnostic log cut from seven fields to one | **Fixed** (`useStakingManagerService.ts:68-78`) |
| T5 | Route-driven required-chain churn disables delegation actions | **Declined** (the session re-confirms from the surviving receipt) |
| T6 | Entry frame can show wallet warnings with "not connected" chrome | **Declined as deliberate**; now recorded in `docs/decisions/2354-wallet-consent.md:14`, and the code matches |
| T7 | Decision doc names a person | **Fixed** (`:3`) |
| T8 | Geofencing mock rationale deleted | **Fixed** (`RootLayout.test.tsx:34-41`). The split test is **declined** |

Checked, and fine:
- Every BTC staking signing path goes through `createBtcStakingManager`, which returns `null` unless consent is confirmed.
- A Retry captured before consent loss rejects after re-confirmation and after unmount, and the test proves both.
- The consent test fails if either the `!confirmed` gate or the latch is reverted.

Adjacent, not introduced here: `RewardsPreviewModal` renders outside `<AuthGuard>` in `simple-staking/src/ui/common/rewards/index.tsx` (the guard closes at `:490`, the modal is at `:493`). A preview opened while confirmed can still Proceed into `useCombinedRewardsService` → `signBbnTx` (`:75`) after consent drops. The signer is still the live BBN wallet, so the impact is small, but it is the same "page gated, signer not" shape as the vault hooks you deferred to #2482. Worth folding into that follow-up.

gbarkhatov (dismissed):
Follow-up review. Changes since my last review (`067127e7` → `ddc1be3e`).

**Headline:** the new commit only deletes `docs/decisions/2354-wallet-consent.md`, `docs/releases/2354-wallet-consent.md`, and the three one-line comments that pointed at them. It changes no logic and no tests. Consent gating is unchanged at `useConnection.ts:22` (`confirmed && btcConnected && ethConnected`) and at `useStakingManagerService.ts:57` and `:68`. No merge-blockers.

### Status of prior threads

| # | Finding | Status at `ddc1be3e` |
|---|---|---|
| T1 | Consent latch breaks a saved Retry after a consent flicker | Declined, deferred to #2506. Not touched by this commit. |
| T2 | Orphaned `jest.jsdom.setup.js`; deleted MultiStaking test | Setup file still fixed (`jest.config.js` uses `./jest.setup.js` only). The MultiStaking half was my mistake and stays withdrawn. |
| T3 | BTC blip ejects the user app-wide; one name meaning two things | Rename still fixed (`walletsPresent`). The blip regression stays accepted as declined, with #2354 tracking it. |
| T4 | Diagnostic log reduced to one field | Still fixed. |
| T5 | Route-driven required-chain churn disables delegation actions | Declined. Unchanged. |
| T6 | Entry frame can show wallet warnings under "not connected" chrome | Declined as deliberate, and the code split is unchanged (`RootLayout.tsx` gates on `isConnected` and warns on `walletsPresent`). See note 2. |
| T7 | Decision doc names a person | Moot: the file is deleted. |
| T8 | Geofencing mock rationale deleted | Still fixed. The split-test suggestion stays declined. |

### New in this delta

1. **The PR body now describes and links documents that no longer exist.** Under Change, it still lists "Record the selected consent rule, support guidance, and release note". Under Links, "Consent rule" and "Release note" point at the two paths this commit deleted, so both 404 from the branch. Verification also still cites "Current commit `067127e7`". Please drop that bullet and the two links, update the commit reference, and say in one line where the consent rule is now recorded (for example on #2354), or that it is no longer recorded in the repo.

2. **T6's decline is now recorded only in the review thread.** I accepted that decline because the deleted decision doc stated the split explicitly: "consent decides access, and wallet presence still drives read-only wallet warnings". With the doc and the `// Consent contract:` pointer gone, the only in-code description is the JSDoc at `useConnection.ts:8`, and it doesn't mention the warning split. Nothing binding requires a decision doc: neither `SECURITY.md` nor `services/vault/SECURITY_MODEL.md` mentions consent. So this doesn't block merge, but a sentence on #2354 would keep the rationale findable.

### Refuted

- *"Deleting the docs leaves dangling references."* A grep for `2354-wallet-consent` across the worktree (excluding `node_modules` and `.git`) returns nothing. The remaining `#2354` mentions in tests cite the issue, not the doc.
- *"A security model cites the consent doc."* Neither `SECURITY.md` nor `services/vault/SECURITY_MODEL.md` mentions consent or the decision doc.

### Verified clean

- No test files in the delta. The consent test in `useTransactionService.test.tsx` (StrictMode wrap, the rejection cases, the call count) is intact.
- No path newly reaches signing or a connected-wallet action without confirmed consent.
- CI is green at `ddc1be3e`, including `verify`.

jonybur:
@greptileai review the full PR at current commit `ddc1be3e5ff777360cbbe11aa7e48b6f3cf1252e`. Check the changes against the target branch and report any remaining issues and the confidence score for this commit.

The PR description now reflects the removed documents and current verification evidence. The selected consent rule remains in #2354. The accepted Retry recovery deferral remains in #2506.

gbarkhatov (approved):

jonybur:
@greptileai review the full PR at current commit `e2ffb811db583b7dcae9fa2cb4af6e546b71b596`. Check the changes against the target branch and report any remaining issues and the confidence score for this commit.

The branch is now one signed commit on current main. Its full patch preserves the previous PR behavior. The only additional change removes four unused mock lines. All local builds, lint checks, unit tests, and 42 browser tests pass. The selected consent rule remains in #2354. The accepted Retry recovery deferral remains in #2506.

gbarkhatov (approved):

PR #2492, opened 2026-09-09
feat(wasm)!: guard raw payout connector outputs

jonybur:
## Outcome

Payout connector fields match the trusted keys and timelock before use. Validation errors now remain visible when cleanup also fails.

Status: [Greptile gives 5/5](https://github.com/babylonlabs-io/babylon-toolkit/pull/2492#issuecomment-5617196710) for `f51a6fcd`. All review threads are resolved. Required human reviews remain. This PR is stacked on #2481; full GitHub verification follows the parent merge.

## Change

- Keep the independent payout script and PegIn vault-output checks.
- Preserve the primary guard message and report both errors after failed cleanup.
- Report cleanup failure after successful validation.
- Reuse the existing payout factory for script-only reads. Return the same two fields.

## Safety

The engine and SDK must release together because this changes the raw class identity. Private engine objects, input guards, free/dispose support, and lazy Ethereum imports remain.

Two qualified human reviews remain. The transaction classes are covered separately in #2493.

## Verification

- Fixed source: 35 engine tests pass. Old source fails six cleanup regression checks.
- Engine build and lint pass.
- The full descendant tree in #2493 passes 90 engine tests, 1,978 SDK tests, nine built-package checks, both builds, and both lints.
- Full GitHub verification runs only for PRs to main. It remains required after the parent merges and this PR targets main.

## Links

Tracks #2361. #2361 stays open. Parent: #2231. Related epic: #2228.

Depends on #2481. Transaction guards: #2493.


jonybur:

Please review the full PR at current commit `66af6b70f05a79830b0f5d00622dee2d09806937`. Check the diff against its target branch. Report any remaining issues and the confidence score for this commit.

.github/workflows/critical-path-check.yml
```diff
@@ -60,6 +60,10 @@ jobs:
             "packages/babylon-tbv-rust-wasm/src/raw-node.ts"
             "packages/babylon-tbv-rust-wasm/src/prePeginHtlc.ts"
             "packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts"
+            "packages/babylon-tbv-rust-wasm/src/connectorScripts.ts"
```
gbarkhatov, line 63: **MERGE BLOCKER (process, not code).** This hunk adds the four new payout paths to a workflow that cannot run on this PR. The trigger at the top of this file is `on: pull_request: branches: [main]`, and this PR's base is `feat/2361-guard-raw-connectors` — a stacked branch. I confirmed the consequence with `gh pr checks 2492`: **only GitGuardian and Greptile ran.** No build, no test, no lint, no critical-path check, no CLAUDE.md drift check.

So everything this PR rests on is asserted by code that was never executed: the §9 differential in `wasm-loader.test.mjs`, the new SDK guard at `pegin.ts:465-489`, both edited SDK test files, and the TypeScript assignability of `GuardedPeginPayoutConnector` to `typeof Bindings.WasmPeginPayoutConnector` at `raw.ts:15-16` — where a dropped or renamed method is a compile error that nothing here would catch. The CODEOWNERS additions are inert for the same reason, which means §1's two-reviewer requirement for these exact files is being enforced by branch protection that never evaluated them.

Not asking for a rebase necessarily — linking a green run at this head (a main-targeted run, or a local `pnpm --filter @babylonlabs-io/babylon-tbv-rust-wasm test` plus typecheck) would discharge it. Worth noting #2493 sits on the same stack and will have the same gap.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
```diff
@@ -461,14 +462,29 @@ async function assertPeginTxShape(
...
-  ).toLowerCase();
+  const { deriveExpectedPeginPayout } = await loadTbvWasm();
+  const expectedVaultScript = (
+    await deriveExpectedPeginPayout({
+      txGraphVersion: version,
+      depositor: params.prePeginParams.depositorPubkey,
+      vaultProvider: params.prePeginParams.vaultProviderPubkey,
+      vaultKeepers: params.prePeginParams.vaultKeeperPubkeys,
+      universalChallengers: params.prePeginParams.universalChallengerPubkeys,
+      timelockPegin: params.timelockPegin,
```
gbarkhatov, line 494: `timelockPegin` reaches the engine unvalidated on this path and is silently truncated to `u16`, and the only thing that catches it is a side effect of the new derivation — after the transaction is already built.

`params.timelockPegin` is handed to `buildPeginTxFromPrePegin` at line 318 with no range check, and `assertPeginTxShape` runs at line 323, afterwards. The only 1..65535 check anywhere in this path lives inside the derivation at `peginPayout.ts:40-44`. One review lane probed the pinned binary directly: `timelock_pegin: 65537` is **accepted**, producing a payout script ending `51 b2` — `OP_1 OP_CSV`, a **one-block timelock**. Only `0` and `65536` are rejected, as `NonZeroU16`.

Today the PegIn is still rejected, but incidentally, via a mismatch whose error never mentions truncation — so a caller passing `65537` learns that something disagreed, not that their timelock silently became 1. Asserting the integer range on `params.timelockPegin` at the top of `buildPeginTxFromFundedPrePegin`, naming the truncation, would make it explicit; keep the derivation's check as the backstop.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
```diff
@@ -461,14 +462,29 @@ async function assertPeginTxShape(
     );
   }
   const encodedVaultScript = encodedVaultOut.script.toString("hex");
-  const expectedVaultScript = stripHexPrefix(
-    result.vaultScriptPubKey,
-  ).toLowerCase();
+  const { deriveExpectedPeginPayout } = await loadTbvWasm();
```
gbarkhatov: This destructures the new derivation straight off the loader, bypassing the registered lazy WASM boundary.

CLAUDE.md §4 and §9 register `packages/babylon-ts-sdk/src/tbv/core/wasm/` as "the lazy boundary every WASM-computed value now crosses", and that file already wraps this function's own sibling — `core/wasm/index.ts` has a named `createPayoutConnector` wrapper doing exactly this hop. `core/wasm/index.ts` isn't touched by this PR, and grep confirms lines 465 and 467 here are the only SDK references to `deriveExpectedPeginPayout`.

The effect is that the single derivation the entire vault-output bind depends on now sits outside the registered boundary — and therefore outside that file's CODEOWNERS. Adding a wrapper in `core/wasm/index.ts` and importing it like the rest would keep the boundary honest.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
```diff
@@ -461,14 +462,29 @@ async function assertPeginTxShape(
...
+      universalChallengers: params.prePeginParams.universalChallengerPubkeys,
+      timelockPegin: params.timelockPegin,
+    })
+  ).scriptPubKey.toString("hex");
+  if (
+    stripHexPrefix(result.vaultScriptPubKey).toLowerCase() !==
+    expectedVaultScript
+  ) {
+    throw new Error(
+      "WASM PegIn vault scriptPubKey does not match the requested payout connector.",
```
gbarkhatov: This error names neither value, unlike its sibling six lines below. The throw at 485-488 interpolates both hex strings, as does every other throw in `assertPeginTxShape` — including the one at 459-462 immediately above.

This is the branch that fires when the engine binary disagrees with the independent TypeScript derivation. That makes it the highest-value diagnostic in the file, and it ships blind: a depositor on a stale or doctored engine generates a support ticket with nothing to diff. CLAUDE.md requires error messages to state what failed and what the expected state was. Interpolating `result.vaultScriptPubKey` and `expectedVaultScript` costs nothing and both are already in scope.

CLAUDE.md
```diff
@@ -52,6 +52,10 @@ These paths handle irreversible value movement. An AI-generated mistake here is
   - `packages/babylon-tbv-rust-wasm/src/index.ts`
   - `packages/babylon-tbv-rust-wasm/src/prePeginHtlc.ts` - shared independent HTLC derivation
   - `packages/babylon-tbv-rust-wasm/src/rawHtlcConnector.ts` - raw HTLC output checks
+  - `packages/babylon-tbv-rust-wasm/src/connectorScripts.ts` - shared key and multisig script rules
```
gbarkhatov, line 55: These four files are registered under §1 but not §9, which detaches them from the rule that actually mandates their differential test.

`connectorScripts.ts` and `peginPayout.ts` are textbook §9 artifacts: hand-assembled tapscript, and in `normalizeXOnlyKey` a hand-rolled secp256k1 curve-membership check replacing the engine's `XOnlyPublicKey::from_str` — the same category as `onChainBtcPubkey.ts`, which §9 lists for precisely that reason. §1's rule is "assert WASM output against expected bounds"; only §9 demands "a differential test … over golden vectors **plus** randomised inputs."

Nothing is broken today — that test exists and is thorough. The gap is forward-looking: a future edit to `normalizeXOnlyKey` or `nOfNChunks` will consult §1, which does not require one. Worth noting `claude-md-drift.yml:72` already treats these paths as section-9 registered, so the two are currently inconsistent with each other.

packages/babylon-tbv-rust-wasm/src/connectorScripts.ts
```diff
@@ -0,0 +1,67 @@
...
+export const MAX_U16 = 0xffff;
+const SEC1_EVEN_Y_PREFIX = 0x02;
+type ScriptChunk = number | Buffer;
+
+function stripHexPrefix(hex: string): string {
+  return hex.startsWith('0x') || hex.startsWith('0X') ? hex.slice(2) : hex;
+}
+
+export function normalizeXOnlyKey(value: string, label: string): string {
+  const key = stripHexPrefix(value).toLowerCase();
```
gbarkhatov: The "independent" normalizer accepts `0x`-prefixed keys that the engine rejects, so it can return a script the engine could never produce.

`normalizeXOnlyKey` strips `0x`/`0X` before validating. A review lane confirmed against the pinned binary that `new WasmPeginPayoutConnector(1, …, ['0x…'], …)` fails with `malformed public key`. Inside the guard this is harmless, since the inner constructor throws immediately — but `deriveExpectedPeginPayout` is a *public* engine export, consumed bare as the trusted expectation at `pegin.ts:467`. An input shape that only one side of a comparison accepts is exactly the asymmetry an independent cross-check must not have.

No live failure today, and the behaviour originates in code moved from `prePeginHtlc.ts` rather than being introduced here — but this PR extends it to a second caller. Rejecting the prefix outright would mirror `XOnlyPublicKey::from_str`.

packages/babylon-tbv-rust-wasm/src/prePeginHtlc.ts
```diff
@@ -10,18 +9,19 @@ import { Buffer } from 'buffer';
...
-const TAPSCRIPT_LEAF_VERSION = 0xc0;
+import {
+  MAX_U16,
+  TAPSCRIPT_LEAF_VERSION,
+  normalizeXOnlyKey,
+  normalizeKeyGroup,
+  nOfNChunks,
+} from './connectorScripts.js';
 
 function stripHexPrefix(hex: string): string {
```
gbarkhatov: The extraction that was meant to remove duplication duplicated `stripHexPrefix`. All three review lanes independently flagged this.

This PR created `connectorScripts.ts` to hoist the shared helpers — it moved `normalizeXOnlyKey`, `normalizeKeyGroup`, `nOfNChunks`, `MAX_U16` and `TAPSCRIPT_LEAF_VERSION` — but *copied* `stripHexPrefix` rather than exporting it, leaving a byte-identical twin at `connectorScripts.ts:10`. This file already imports from `./connectorScripts.js` two lines above, so the import is free.

Both files are §1 critical paths, and the failure mode is divergence: tightening prefix handling in one copy (see my comment on `connectorScripts.ts:15`) would leave HTLC and payout key normalization silently disagreeing on the same input. Export it once, delete the copy.

packages/babylon-tbv-rust-wasm/src/connectorScripts.ts
```diff
@@ -0,0 +1,67 @@
+import { secp256k1 } from '@noble/curves/secp256k1.js';
+import { script as bscript, opcodes } from 'bitcoinjs-lib';
+import { Buffer } from 'buffer';
+
+export const TAPSCRIPT_LEAF_VERSION = 0xc0;
```
gbarkhatov, line 6: The move dropped the only provenance comment on a Taproot magic number. `TAPSCRIPT_LEAF_VERSION = 0xc0` carried `// BIP-341: the leaf version for a Taproot script.` in `prePeginHtlc.ts` and re-lands here bare; `MAX_U16` and `SEC1_EVEN_Y_PREFIX = 0x02` are bare too.

Separately, `nOfNChunks` is now the shared encoder behind *both* the HTLC hashlock leaf and the payout leaf, yet cites no reference — while `peginPayout.ts:41-42` correctly cites `pegin_payout.rs` at all three pins. CLAUDE.md: "If a number appears in code, it must be obvious why that value was chosen." Carrying the BIP-341 line over and citing `script_utils.rs::build_m_of_n_multisig_script` on `nOfNChunks` would restore it.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
```diff
@@ -505,6 +516,256 @@ function htlcArgs(version = 1, keepers = [xOnlyKeys[2]]) {
...
+          );
+          assert.throws(
+            () => checked.getScriptPubKey('invalid'),
+            /Unsupported Bitcoin network/,
+          );
+        } finally {
+          checked.free();
+          original.free();
+        }
+        assert.throws(() => checked.getPayoutScript());
```
gbarkhatov: This matcher-less `assert.throws` pins nothing. It sits outside the `try/finally`, after `checked.free()`, with no error pattern — so it passes on *any* throw, including a `TypeError` from a future typo or from a guard that broke in an unrelated way, rather than the intended use-after-free.

It also tests wasm-bindgen's own lifetime behaviour rather than a behaviour of code in `src/`, which the test philosophy asks us to avoid. Either add a matcher, or drop it — disposal is already covered via `[Symbol.dispose]()` later in the suite.

packages/babylon-tbv-rust-wasm/src/payoutConnector.ts
```diff
@@ -64,15 +70,17 @@ export async function createPayoutConnector(
...
 ): Promise<{ payoutScript: string; payoutControlBlock: string }> {
-  const { WasmPeginPayoutConnector } = await getWasmBindings();
+  await getWasmBindings();
+  const { GuardedPeginPayoutConnector: WasmPeginPayoutConnector } =
+    await import('./rawPayoutConnector.js');
 
   const connector = new WasmPeginPayoutConnector(
     params.txGraphVersion,
     params.depositor,
     params.vaultProvider,
```
gbarkhatov: Informational rather than a defect, but reviewers signing off under §1 should know it: half the surface this PR hardens has no in-repo production consumer. Grepping all of `packages/` and `services/` finds `getPeginPayoutScriptInfo` only in `tests/wasm-loader.test.mjs:683`, `README.md:397`, and the two facade re-exports.

It is published API, so external consumers may well exist — this is explicitly *not* a suggestion to delete it. The point is only that its new guard can currently fire nowhere but in tests, which is worth knowing when weighing how much assurance the guard actually buys today.

gbarkhatov (approved):
Reviewed at `66af6b7` via three independent passes (7 agents), deduped, with everything below re-checked against the source.

**The headline is good news: the gap I flagged on #2481 does not repeat here.** I read `payoutConnector.ts`, `index-node.ts` and `raw.ts` myself — `createPayoutConnector` and `getPeginPayoutScriptInfo` construct `GuardedPeginPayoutConnector` on *both* the browser and Node entries, `/raw` uses the same class, and a repo-wide grep finds no unguarded `new WasmPeginPayoutConnector` outside `rawPayoutConnector.ts` and the tests. The SDK cross-checks again at the call site. That was the single most important thing this review was looking for, and it is absent.

One lane went further than static reading: it loaded the pinned engine binary in Node and ran the TypeScript derivation against it over **780 comparisons** (3 graph versions × key counts 1–18 × 16 timelocks × 5 networks) with **zero mismatches**, and independently recomputed all four pinned vectors. The payout script layout matches the Rust at all three pins.

One merge blocker, and it is about process rather than code — see the comment on `critical-path-check.yml`. The rest are non-blocking. Approving.

- *"Divergence from the Rust payout connector."* None at any pin. Component order, the `OP_CHECKSIG`/`OP_CHECKSIGADD` shape, lexicographic key sort, the single `0xc0` leaf at depth 0 and the NUMS internal key all match `generate_payout_script` — plus 780 live binary comparisons.
- *"TS rejects empty/duplicate key groups but the engine does not."* It does — both groups route through `VaultKeepers::new` / `UniversalChallengers::new`, which sort, dedup-check and reject empty. Confirmed by probing the binary.
- *"The guard wrapper drops a method."* No — the generated `.d.ts` lists exactly the eight members the wrapper forwards.
- *"Missing breaking-release marker for `feat(wasm)!`."* No — the repo uses semantic-release and has no `.changeset/`; the major is derived from the commit subject.
- *"`index-node.ts` sits outside the critical-path gate / eslint strict tier"*, *"the eslint change weakens a rule"*, *"`pegin.ts` relaxes an assertion"*, *"browser/Node `/raw` parity hole"*, *"§9's differential test is missing"* — all five false; the eslint hunk only *adds* files to the stricter block, and `raw.ts`/`raw-node.ts` differ by one import line.

**Verified clean:** §9's differential requirement is genuinely met, not gestured at — golden hex pins for scriptPubKey, leaf hash and control block, plus randomised samples over key counts 1/2/16/17 and timelocks hitting every CScriptNum boundary (1, 16, 17, 127, 128, 255, 256, 32767, 32768, 65535), each compared field-by-field on both `/raw` entries, with a negative path that patches the getters and asserts the mismatch. `assertPeginTxShape` is strictly stronger than what it replaced: both the WASM-reported `vaultScriptPubKey` and the encoded output bytes are now bound to a TypeScript expectation derived from the caller's own requested keys and timelock, so a consistent engine-wide redirect is caught. Object lifetime is handled — the constructor frees the inner engine object on guard failure, both facades free in `finally`, and the tests count releases rather than only asserting the throw.

Worth saying: the PR body undersells a real hardening. The new input validation closes the engine's silent `u16` truncation of `timelock_pegin` at every *guarded* entry point. The comment on `pegin.ts:473` is that the SDK's own builder path is not one of them.

jonybur (commented):
The payout guard is well shaped, and its differential is stronger than the HTLC one it follows: group sizes of 1, 2, 16 and 17 against timelocks at 1, 16, 17, 127, 128, 255, 256, 32767, 32768 and 65535 exercise both the `OP_CHECKSIG`/`OP_CHECKSIGADD` chain and the script-number push boundaries against the real generated class, across all three graph versions and five networks. The reversed-roster assertion pins the sort as well. Two things on `rawPayoutConnector.ts`.

**1. `free()` is not idempotent, so `free()` followed by `[Symbol.dispose]()` reaches the generated free twice.**

`free()` forwards unconditionally to `#inner.free()`, and `[Symbol.dispose]()` calls `this.free()`, with no flag between them. wasm-bindgen's generated `free()` zeroes `__wbg_ptr` through `__destroy_into_raw()` before calling the free import, so the second pass hands that import a null pointer and it is released as if it were an object — heap corruption rather than a clean error.

It is reachable from what this PR ships rather than only in theory: `createPayoutConnector` and `getPeginPayoutScriptInfo` both free in a `finally`, so a caller who writes `using connector = new WasmPeginPayoutConnector(...)` alongside either gets two frees. A `#freed` boolean with an early return from both methods closes it. The existing test covers dispose-then-getter but not free-then-dispose, so nothing currently pins this either way.

**2. The constructor's cleanup can replace the guard error with an unrelated one.**

In the catch, `this.#inner.free()` runs before the rethrow. If that free throws — an already-released pointer, or detached memory after a second wasm init — the memory error propagates and `WASM payout <field> does not match the constructor inputs.` is lost. That message is the one diagnostic saying a doctored or version-skewed binary was caught, so losing it turns a value-mismatch alarm into what reads like a loader fault. Wrapping the free in its own try/catch would let the guard error always win.

**Checked and clean.**

- The guard covers every member of the generated class, and the `typeof Bindings.WasmPeginPayoutConnector` annotation on the `/raw` export is what enforces that — a getter added to the generated class later cannot slip through unguarded without failing the build.
- Expected values are captured at construction from the caller's arguments and `normalizeKeyGroup` maps before it sorts, so neither the guard's expectations nor the caller's arrays can move afterwards. The mutation test pins that.
- Failed async construction releases the engine object on all three factory paths, and the release counter in the test proves it rather than assuming it.
- `getScriptPubKey` discarding its network lookup is right: a P2TR scriptPubKey is network-independent and the lookup is there to reject an unknown network name.
- The new SDK check in `assertPeginTxShape` is strictly stronger than what it replaced. It derives the vault script independently and rejects both the WASM-reported and the encoded value against that derivation, where before it only compared the two WASM answers to each other — which is exactly the redirect the new `pegin.test.ts` case now exercises for all three versions.

jonybur:

I checked each of the ten open threads against a prepared fix set. I have not pushed anything. A push would dismiss the standing approval (`dismiss_stale_reviews_on_push` is on) while the threads stay open.

Nine threads need work. One needs only this reply.

### The CI thread is correct, and it decides the merge order

`verify` and `detect` are the two required checks. Both workflows trigger on `pull_request: branches: [main]`, so neither runs while this PR targets `feat/2361-guard-raw-connectors`. No amount of review closes that. This PR must retarget to `main` after #2481 merges, and #2493 must then retarget after this one. I will not merge it as a stacked PR.

### Still open (8)

- **`pegin.ts` — `timelockPegin` range.** No explicit 1..65535 check at the top of `buildPeginTxFromFundedPrePegin`. The value still reaches the engine and truncates to u16, caught only as a side effect.
- **`pegin.ts` — lazy WASM boundary.** The `loadTbvWasm()` destructure is gone, but it was replaced by a local TypeScript derivation rather than a named wrapper in `core/wasm/index.ts`. The boundary file is untouched, so the path is still outside its CODEOWNERS entry, and nothing pins the new derivation to the binary.
- **`pegin.ts` — error names neither value.** The scriptPubKey mismatch throw still interpolates nothing. Its sibling six lines below prints both hex strings.
- **`CLAUDE.md` — section 9 registration.** The new connector and derivation files are still registered only under section 1, so the section-9 differential-test rule does not bind them. `claude-md-drift.yml` already claims they are section-9 registered.
- **`connectorScripts.ts` — `0x` prefix.** `normalizeXOnlyKey` still strips `0x` and `0X` and accepts the key. Rust `XOnlyPublicKey::from_str` rejects it, so the independent expectation can still return a script the engine could never produce.
- **`prePeginHtlc.ts` — duplicate `stripHexPrefix`.** Both copies are still present, one in each file.
- **`connectorScripts.ts` — provenance comments.** The BIP-341 comment on `TAPSCRIPT_LEAF_VERSION` is still dropped. `MAX_U16` and `SEC1_EVEN_Y_PREFIX` are still bare, and `nOfNChunks` does not cite `script_utils.rs::build_m_of_n_multisig_script`.
- **`wasm-loader.test.mjs` — matcher-less `assert.throws`.** Unchanged. It still passes on any throw and only exercises wasm-bindgen's own use-after-free.

### Noted, no change (1)

- **`payoutConnector.ts` — no in-repo consumer.** Understood, and I agree it should be on the record for anyone signing off under section 1. `getPeginPayoutScriptInfo` is reached today only by tests, the README and the two facade re-exports. The guard is written for the consumer that arrives with the optional-Bitcoin work. I am keeping it and leaving this on the PR rather than resolving it silently.

.github/workflows/critical-path-check.yml
jonybur: Confirmed, and discharged with a local run.

The trigger is `pull_request: branches: [main]` in `verify.yml`, `critical-path-check.yml` and `visual-regression.yml`, so none of them ran for this stacked base. One correction: `claude-md-drift.yml` runs on `schedule` and `workflow_dispatch` only, so it runs for no pull request, main-targeted or not.

Results at `0e2b5521`, in a clean worktree with a fresh `pnpm install --frozen-lockfile`:

- engine `tsc -b tsconfig.lib.json`: pass. This is the check that covers the `GuardedPeginPayoutConnector` assignability at `raw.ts:15-16`.
- `node scripts/check-lazy-entries.js`: pass.
- engine tests: 30 pass, 0 fail.
- SDK: 1,978 tests in 117 files, plus the 9 `wasm-facade.node.mjs` checks.
- Chromium: 34 pass.
- `pnpm build`, `pnpm run lint`, `syncpack list-mismatches`: pass.
- `docs:clean`: no diff in `docs/api`.

Not covered locally: the visual-regression job, which needs a main-targeted run, and the release rehearsal, which needs a CI token.

#2493 sits on the same stack and will carry the same evidence.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
jonybur: Fixed at `0e2b5521`. `buildPeginTxFromFundedPrePegin` now rejects the timelock before the engine call, and the message names the truncation. Four cases are pinned: 0, 65536, 65537 and 1.5.

Three notes from a direct probe of the pinned binary:

- On this branch the failure is not an unnamed mismatch. `deriveExpectedPeginPayout` checks the range first, so 65537 already produces `timelockPegin must be an integer from 1 to 65535.` The defect is that the check ran after the transaction was built. That is what moved.
- Your description does hold for the code before this PR. At `HEAD~1` the vault-script bind compared `result.vaultScriptPubKey` with itself, so 65537 returned a one-block vault with a valid txid and the right value. The rejection today rides on a call this PR added for another purpose, which is a good reason to make the check explicit.
- 0 and 65536 did not reach that check at all. The engine threw `timelock_pegin must be non-zero`, which misdescribes 65536: the value is not zero, it truncated to zero. Both now fail with the range message before the engine call.
- Rejection is not limited to 0 and 65536. Every multiple of 65536 is rejected as `NonZeroU16`; 4294967296 throws the same way.

The gap stays open for callers that do not use the SDK: the facade `buildPeginTxFromPrePegin` passes both `timelockPegin` and `timelockRefund` into `WasmPrePeginTx` unchecked, and both truncate. That is one of the two raw transaction classes this PR leaves unguarded on purpose, so I will carry it into #2493.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
jonybur: Fixed at `0e2b5521`. `core/wasm/index.ts` now exports a `deriveExpectedPeginPayout` wrapper, and `pegin.ts` imports it instead of destructuring the loader.

One correction: CODEOWNERS already covered both files. Line 22 covers `pegin.ts` and line 63 covers `core/wasm/`, with the same five owners, and this PR added `peginPayout.ts`, where the derivation lives. So there was no ownership gap, only a boundary that did not name the function.

The sibling has the same shape and this PR does not touch it: `assertWasmPeginSizing.ts:53-58` wraps `deriveExpectedPrePeginHtlc` locally instead of in `core/wasm/`. I will move it in a follow-up if you want the boundary complete.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/pegin.ts
jonybur: Fixed at `0e2b5521`. The throw now interpolates `result.vaultScriptPubKey` and `expectedVaultScript`, like its sibling six lines below. The redirect test asserts the reported script appears in the message.

CLAUDE.md
jonybur: Partly accepted. Fixed at `0e2b5521`, by a different route.

The rule you want is now attached where the files are registered. Section 1's Rule states that `connectorScripts.ts`, `prePeginHtlc.ts` and `peginPayout.ts` carry the same differential-test duty as section 9, over the golden vectors plus randomised inputs.

I did not move them into section 9. That section covers primitives rebuilt without `bitcoinjs-lib`, `tiny-secp256k1` or the engine. These files import `bitcoinjs-lib` and `@noble/curves`, so the heading does not describe them.

I also left CODEOWNERS alone, after weighing it. The six engine derivation paths do sit under its `# 9.` header while CLAUDE.md files them under section 1. But every entry in that file carries the same five owners, so regrouping moves no ownership, and the same mismatch already exists for `index-node.ts` and `build-wasm.js`, which sit under `# 1.` while CLAUDE.md files them under section 4. Regrouping the headers is worth doing for the whole file in its own PR, not as churn on a control file inside this one.

One correction: `claude-md-drift.yml:72` is a comment line in the block that documents `pending=(...)`, and that array holds only `scriptPubKeyAddress.ts`. The `paths=(...)` array carries no section attribution, so drift.yml and CLAUDE.md did not disagree.

If you would rather have them registered in section 9, say so and I will move them.

packages/babylon-tbv-rust-wasm/src/connectorScripts.ts
jonybur: Fixed at `0e2b5521`. `normalizeXOnlyKey` no longer strips the prefix, so it accepts exactly what `XOnlyPublicKey::from_str` accepts. A new assertion pins both sides against the same input: the guarded class throws `depositor must be a 32-byte x-only public key`, the generated class throws `malformed public key`.

For the record, a probe found the lenient branch unreachable with a wrong value: every consumer hands the same keys to the engine first, and no accepted input produced different key bytes on the two sides. So this removes the asymmetry, not a live defect, and the visible effect is which error text surfaces at the two guarded raw classes. Say so if you would rather keep the strip and document the seam instead.

packages/babylon-tbv-rust-wasm/src/prePeginHtlc.ts
jonybur: Fixed at `0e2b5521`. `connectorScripts.ts` exports `stripHexPrefix` and `prePeginHtlc.ts` imports it. The copy is deleted. The hashlock at `prePeginHtlc.ts:69` still uses it, which is the one place that keeps prefix handling.

packages/babylon-tbv-rust-wasm/src/connectorScripts.ts
jonybur: Fixed at `0e2b5521`. `TAPSCRIPT_LEAF_VERSION` carries the BIP-341 line again, and `MAX_U16` and `SEC1_EVEN_Y_PREFIX` now state why they hold those values. `nOfNChunks` cites btc-vault `crates/vault/src/connectors/script_utils.rs` `build_m_of_n_multisig_script` at 2c1177ec, 27c0062b and e1e50f66.

One detail: the path is `connectors/script_utils.rs`, not `utils/`. I checked the function at line 29 in all three pins.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
jonybur: Fixed at `0e2b5521`, and in five more places. Every matcher-less `assert.throws` in the file now carries one: `/null pointer passed to rust/` at the four use-after-free sites, and `/Unsupported Bitcoin network/` at the two network sites. Both strings come from a probe of the pinned binary.

I kept the assertion rather than deleting it. After the guard's own `free()`, it is the only check that the guard released the engine object it owns, which is behavior of `rawPayoutConnector.ts`, not of wasm-bindgen alone.

packages/babylon-tbv-rust-wasm/src/payoutConnector.ts
jonybur: Acknowledged. No code change.

The grep agrees, with two details: `index-node.ts:354` is a second hand-written export with the same body rather than a re-export, and the definition is `payoutConnector.ts:70`. So the guard on that surface fires only in tests today. Recorded for the section 1 sign-off.

jonybur:
Review fixes are pushed as `0e2b5521`.

Seven of the ten comments produced a change. Three are answered without one, with the reasoning in the thread.

Behavior changes: the independent key normalizer now rejects a `0x` prefix, so it accepts exactly what the engine accepts, and `buildPeginTxFromFundedPrePegin` rejects a PegIn timelock outside 1..65535 before the engine can truncate it.

CI at this head, run locally because the stacked base runs no workflow: engine `tsc -b` and 30 engine tests, 1,978 SDK tests in 117 files plus 9 facade checks, 34 Chromium tests, `pnpm build`, `pnpm run lint`, syncpack, and `docs:clean` with no diff.

gbarkhatov (approved):

packages/babylon-tbv-rust-wasm/src/rawPayoutConnector.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (version !== this.#version) {
+      throw new Error(
+        'WASM payout version does not match the constructor input.',
+      );
+    }
+    return version;
+  }
+
+  free(): void {
+    this.#inner.free();
```
jeremy-babylonlabs, line 130: `free()` has no `#freed` latch, so a `using` declaration plus an explicit `free()` in a `finally` calls `__wbg_*_free(0, 0)` and throws `null pointer passed to rust`. `GuardedPrePeginHtlcConnector.free()` in #2481 already has the latch and documents this exact case — worth mirroring it here.

packages/babylon-tbv-rust-wasm/src/rawPayoutConnector.ts
```diff
@@ -0,0 +1,122 @@
...
+    this.#inner = new RawConnector(...args);
+    try {
+      this.getTxGraphVersion();
+      this.getPayoutScript();
+      this.getPayoutControlBlock();
+      this.getTaprootScriptHash();
+      this.getScriptPubKey('bitcoin');
+    } catch (error) {
+      this.#inner.free();
+      throw error;
```
jeremy-babylonlabs, line 43: Wrap this `free()` in try/catch like `rawHtlcConnector.ts` does, so a release failure cannot replace the guard's mismatch error.

jeremy-babylonlabs (commented):
Payout derivation matches the pinned versions and the cross-check against the vault output is good. The guard class drops one protection #2481 established.

Also note only one check is passing here because the verification workflows run against `main` — worth confirming CI after the retarget.

packages/babylon-tbv-rust-wasm/src/rawPayoutConnector.ts
jonybur: Fixed at `a7e9a1b2`. `free()` now early-returns on a `#freed` latch, mirroring `GuardedPrePeginHtlcConnector`, and a test pins that a second free() and a dispose() after it both stay quiet.

packages/babylon-tbv-rust-wasm/src/rawPayoutConnector.ts
jonybur: Fixed at `a7e9a1b2`. The release in the constructor's catch is wrapped, with the same comment `rawHtlcConnector.ts` carries, so a release failure cannot replace the guard's mismatch error.

jonybur:
Merged the updated base (#2481 at `79612752`) into this branch as `a7e9a1b2`.

Four files conflicted and were resolved by hand:

- `prePeginHtlc.ts` keeps this PR's shared imports from `connectorScripts.ts` and takes #2481's module-scope `initEccLib`, so the ECC backend is installed once per module evaluation instead of per derivation.
- `CLAUDE.md` and `SECURITY.md` now carry #2481's correction across both connectors: they check their own constructor inputs for internal consistency, and the caller must still check those inputs against on-chain data. The "complete output checks" wording this PR introduced is gone, and the co-location note now names both derivations.
- `wasm-loader.test.mjs` keeps both test sets.

Re-verified at this head: 33 engine tests, 1,978 SDK tests in 117 files plus 9 facade checks, 3,809 Vault tests, 34 Chromium tests, build, lint, and `docs:clean` with no diff.

jonybur:
@greptileai review the full PR at current commit `a7e9a1b215008298572c1e01f2c585339c6f0d0d`. Check the diff against its target branch, including the changes since your last review. Verify that the review findings are addressed in the code. Report any remaining issues and the confidence score for this commit.

jonybur:
@greptileai review the full PR at current commit f51a6fcdcc3e8ea113db46b0e26cb7831b711929. The review fixes are published and local checks pass. Please check the current code, report any remaining findings, and give the confidence score for this commit.

packages/babylon-tbv-rust-wasm/src/peginPayout.ts
```diff
@@ -0,0 +1,81 @@
...
+    );
+  }
+  if (
+    !Number.isInteger(params.timelockPegin) ||
+    params.timelockPegin < 1 ||
+    params.timelockPegin > MAX_U16
+  ) {
+    throw new Error('timelockPegin must be an integer from 1 to 65535.');
+  }
+  initEccLib(ecc);
```
gbarkhatov, line 46: Non-blocking. `initEccLib(ecc)` still runs on every guarded construction here. Its sibling `prePeginHtlc.ts` now calls it once at module scope, via #2481's fix merged into this branch, with the comment that bitcoinjs-lib keeps one global ECC backend a derivation shouldn't swap. The two derivations now disagree, and whichever file is edited next will take the other as precedent. Suggest moving it to module scope the same way.

Optional while here: `:30-38` spells out `!== 1 && !== 2 && !== 3`, where the HTLC side uses `SUPPORTED_HTLC_GRAPH_VERSIONS`.

gbarkhatov (approved):
Follow-up to my approval at `0e2b5521aa`. Since then there are three commits: `a7e9a1b2` (release flag and constructor cleanup), `f51a6fcd` (facade cleanup; the script-only read now goes through `createPayoutConnector`), plus #2481 merged in. All 10 threads from the first review are settled, and nothing new is a correctness defect.

**Merge condition:** no repo CI has run on this head. This PR's base is #2481's branch, and the workflows trigger on `branches: [main]`, so `pr-2492.checks.txt` shows only GitGuardian and Greptile. The author's local run was at `0e2b552`, before these three commits. Until this PR is retargeted to `main` and `verify` passes, none of these is proven:
- the rewritten cleanup test and the release-flag test
- `check-lazy-entries.js`
- the TS check that the guarded class can stand in for the generated one (`raw.ts:15-16`)
- `docs:clean`
- the critical-path/CODEOWNERS enforcement

The PR body already commits to that. I traced the cleanup test by hand for each tampered getter, with and without a failing release, and it looks right, but that is reading, not a run.

| # | Prior finding | Verdict at `f51a6fcd` |
|---|---|---|
| T1 | Repo CI can't run on a stacked base | **Still true**, see above |
| T2 | `timelockPegin` reached the engine unchecked and was truncated to u16 | **Fixed** (`pegin.ts:298-312`) |
| T3 | Payout derivation bypassed the `core/wasm` lazy boundary | **Fixed** (`pegin.ts:22-26`, `core/wasm/index.ts:302-305`) |
| T4 | Mismatch error named neither value | **Fixed** (`pegin.ts:501-504`) |
| T5 | Derivation files lacked a differential-test duty | **Fixed differently.** §1 now assigns it (`CLAUDE.md:66`) |
| T6 | `normalizeXOnlyKey` accepted `0x` keys the engine rejects | **Fixed** (`connectorScripts.ts:19-27`), with a test |
| T7 | `stripHexPrefix` copied, not shared | **Fixed** |
| T8 | Provenance comments dropped; `nOfNChunks` uncited | **Fixed** |
| T9 | `assert.throws` calls without a matcher | **Fixed** |
| T10 | `getPeginPayoutScriptInfo` has no in-repo production caller | **Acknowledged** |

**Checked, and fine:**
- Expected values are derived from constructor inputs only, before any engine object exists.
- Every getter, and the SDK-side vault script and output bytes, is compared against that derivation.
- The script-only read now runs every guard plus the address check.
- The cleanup copies #2481's settled shape: `AggregateError` on validation plus release failure, and a single release guarded by `#freed`.
- #2481's head `c1a41725` is an ancestor of this head, so there is no stack drift.

**Corrections to my own T1/T3/T5 wording:**
- `claude-md-drift.yml` only runs on schedule or manual dispatch, so it doesn't belong in the "did not run" list.
- Both paths in T3 were already under the same CODEOWNERS.
- The `claude-md-drift.yml:72` citation was a comment line.

**Optional:** a comment on the hard-coded `'bitcoin'` in `getPeginPayoutScriptInfo`, saying the network doesn't affect the two fields it returns. As written it reads like a mainnet assumption.

PR #2493, opened 2026-09-09
feat(wasm)!: guard raw transaction inputs and restoration

jonybur:
## Outcome

Raw Pre-PegIn and PegIn transactions reject outputs that differ from the trusted request. Cleanup failures remain visible without hiding a guard error.

Status: [Greptile gives 5/5](https://github.com/babylonlabs-io/babylon-toolkit/pull/2493#issuecomment-5617182224) for `aa3a9a2e`. All review threads are resolved. Required human reviews remain. This PR is stacked on #2492; full GitHub verification follows the parent merge.

## Change

- Keep original bigint amounts, independent protocol outputs, refund checks, and trusted restoration inputs.
- Preserve both errors when constructor validation and object cleanup fail.
- Include the connector cleanup fixes from #2481 and #2492.
- Correct the migration guide: saved transactions require complete prevouts. Remove the false missing/null compatibility claim.

## Safety

This remains a breaking engine API. Existing callers must migrate, and the engine and SDK must release together. Expected values do not come from saved engine JSON.

Two qualified human reviews and the required line review remain. The optional-Bitcoin Vault work does not depend on adopting this API.

## Verification

- Final engine tree: 90 tests pass, including real engine vectors and cleanup error cases.
- SDK: 1,978 tests and nine built-package checks pass.
- Engine and SDK builds, lint, lazy-boundary checks, and diff checks pass.
- Independent source review found no further issue in this update.
- Full GitHub verification runs only for PRs to main. It remains required after the parent merges and this PR targets main.

## Links

Tracks #2361. #2361 stays open. Depends on #2492.

Parent: #2231. Related epic: #2228. Vault implementation: #2232.


jonybur:

Please review the full PR at current commit `bf4fad83f3433d24516c13bb14f57fe2f73a06de`. Check the diff against its target branch. Report any remaining issues and the confidence score for this commit.

jonybur (commented):
The fee differential is the strongest part of this and worth saying first: 100 random rosters per graph version against the real engine, the CompactSize and ScriptNum boundaries picked out individually, the u64 saturation edge asserted on both sides, and the above-cap rosters checked against the raw constructor's own reserve rather than the public estimator that rejects them. That is what section 9 asks for, and it is the part that most needed it.

**1. `peginFees.ts` ships its sizing constants without provenance.**

Every number that decides how much BTC ends up locked lives in this file: `11456n`, the `34n` per key, `264n`, `132n * (21n + 2n)`, the per-signature `65n`, the `8n + 1n + ... + 40n + ... + 43n * p2trOutputs` body of `vsize`, `wronglyVsize`'s `65n + 33n + 74n + 162n`, `claimVsize`'s `67n`, `councilVsize`'s `94n * 4n + ... + 32n * depth`, and the `330n` / `546n` / `240n` thresholds. Three comments explain a grouping; none of them names the Rust function the number came from. The file header gives three commit hashes and no path.

Compare the convention this repo already holds elsewhere. Every protocol layout constant in `psbt/constants.ts` carries a `btc-vault crates/vault` reference, and `peginPayout.ts` on the parent branch cites `crates/vault/src/connectors/pegin_payout.rs` at all three pins for a derivation that is far simpler than this one.

This is the file where that convention earns its keep. The differential is the only thing standing between a wrong constant and an underfunded transaction, and a differential can only catch a constant that is wrong today. The next person to touch a byte count has nothing to check it against, and a `VAULT_WASM_COMMIT` bump that moves a Rust size leaves no trace here — the tests would simply start failing with no indication of which of forty numbers moved or why. A `crates/vault/...` path and function name beside each group would make each one checkable on its own.

**2. `free()` is not idempotent on either new class.**

`GuardedPrePeginTx.free()` and `GuardedPeginTx.free()` both forward unconditionally to `#inner.free()`, and both `[Symbol.dispose]()` implementations call `this.free()`, with no flag between them. wasm-bindgen's generated `free()` zeroes `__wbg_ptr` through `__destroy_into_raw()` before calling the free import, so a second pass hands that import a null pointer and it is released as if it were an object. `fromFundedTransaction` already frees an inner on the success path and calls `copy.free()` in its catch, so a caller holding the same wrapper under `using` gets two frees. A `#freed` flag with an early return from both methods closes it — and the same flag is wanted on all four wrappers, since the two connector classes have the identical shape.

**3. Both constructors can lose the guard error to their own cleanup.**

`GuardedPrePeginTx`'s catch runs `this.#inner.free()` and `GuardedPeginTx`'s runs `inner.free()`, each before the rethrow. A throw from either replaces `WASM Pre-PegIn <field> does not match the original request.` or the PegIn JSON mismatch with a memory error. Those messages are the only signal that a doctored or version-skewed binary was caught, so the failure that most needs to be legible is the one most likely to be overwritten. Wrapping each free in its own try/catch keeps the guard error.

**Checked and clean.**

- Moving amount validation out of the two facade entries opens no window. `deriveExpectedPrePegin` runs before `new RawTransaction(...)` in the guard constructor, `assertPositiveBigintArray` runs on a copy of `pegInAmounts`, and the `BigUint64Array` conversion now happens inside the guard from that validated snapshot rather than from caller-supplied values.
- `decodeTransaction` rejecting a non-canonical re-encode and a superfluous witness marker closes the malleability room a plain `fromHex` would leave, and `transactionId` hashing the `RawOldTx` encoding is the right txid for a segwit transaction.
- `assertFundedPrePegin` requiring empty `scriptSig` is correct for this repo rather than a wallet-compatibility hazard: `fundedPrePeginTxHex` is the unsigned funded template everywhere in the vault, not a broadcast parent.
- `buildRefundTx` rebuilding the whole expected serialization instead of comparing fields is the right shape, and the `htlcValues[index] - refundFee` floor is checked before the engine call rather than after.
- `#index` bounding on both u8 and the actual HTLC count, and `fromJson` refusing when the version disagrees with the trusted request, both fail closed.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
```diff
@@ -0,0 +1,413 @@
...
+        htlc_connector: this.#inputConnector,
+        depositor_sig: spender.depositor_sig,
+        vault_provider_sig: spender.vault_provider_sig,
+        vault_keeper_sigs: spender.vault_keeper_sigs,
+        universal_challenger_sigs: spender.universal_challenger_sigs,
+      },
+      ...(Object.hasOwn(data, 'prepegin_htlc_prevout')
+        ? {
+            prepegin_htlc_prevout:
+              data.prepegin_htlc_prevout === null ? null : this.#prevout,
```
gbarkhatov: **MERGE BLOCKER — the restoration guard fails open here, and the untrusted input is what decides whether it applies.**

`exact()` at lines 50-52 compares *sorted key sets* before recursing. This expectation is spread in conditionally on `Object.hasOwn(data, 'prepegin_htlc_prevout')`, where `data` is the saved JSON being validated. So: if the saved JSON omits the key, the expected object omits it too, the key sets match, and nothing is checked. If the value is `null`, the expectation is literally `null` and matches. Only when the field is present *and* non-null does it get compared against `this.#prevout`. Either way the spent HTLC value and scriptPubKey go unbound, and `toJson()` hands that object straight back out.

This is the only `if (present) check` shape in the file, and `GuardedPeginTx` is only ever constructed from a trusted Pre-PegIn, so `Some(prevout)` is the sole legitimate state. Mitigating: `btc-vault crates/vault/src/transactions/pegin.rs` returns `Err("HTLC prevout not available…")` on `None` downstream — but a guard shouldn't lean on a downstream error for its own invariant. Dropping the conditional and always expecting `this.#prevout` would make non-null-and-equal mandatory.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
```diff
@@ -0,0 +1,413 @@
...
+        .reverse()
+        .concat([...keepers].reverse(), provider, depositor);
+      const transaction = Transaction.fromRaw(
+        Buffer.from(this.#unsignedHex, 'hex'),
+        { allowUnknownInputs: true, allowUnknownOutputs: true },
+      );
+      this.#sighashes = new Map(
+        [0, 1].map((hashType) => [
+          hashType,
+          transaction.preimageWitnessV1(
```
gbarkhatov, line 222: **MERGE BLOCKER — the signature digest is the one derivation with no external check, and the test signs with the same function it verifies with.**

`#sighashes` is built from this `preimageWitnessV1` call, and every stored signature and witness item is then accepted or rejected against it in `#verifySignature`. But `tests/pegin-restore.test.mjs:111` *produces* its fixtures with the same `preimageWitnessV1` call, and `computePeginInputSighash` appears nowhere in that file. So if leaf-version, code-separator or annex handling diverges from BIP-342, the suite stays green while every genuinely signed PegIn fails `fromJson` — restoration dies for in-flight deposits and nothing here would tell you.

The oracle already exists at the pin: `vault-wasm@3accd8f src/dispatch/pegin_tx.rs:180` exports `js_name = computePeginInputSighash`, and the fixture already carries all three JSON blobs it needs. Asserting the map's entries against it, for both hash types across v1/v2/v3, would close the last unverified derivation in this file.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
```diff
@@ -65,10 +58,16 @@ for (const entry of ['raw', 'raw-node']) {
       'type Same<A, B> = (<T>() => T extends A ? 1 : 2) extends (<T>() => T extends B ? 1 : 2) ? true : false;',
       'type Check<T extends true> = T;',
       ...rawClassNames.flatMap((name) => [
-        `type CheckType${name} = Check<Same<${name}, generated.${name}>>;`,
-        `type CheckConstructor${name} = Check<Same<typeof ${name}, typeof generated.${name}>>;`,
+        `type CheckType${name} = Check<Same<${name}, ${name.includes('Connector') ? `generated.${name}` : name === 'WasmPeginTx' ? `ReturnType<typeof ${name}.fromJson>` : `InstanceType<typeof ${name}>`}>>;`,
```
gbarkhatov: **MERGE BLOCKER — the type-parity gate for the two rewritten classes was replaced with tautologies.**

The emitted checks are `Same<WasmPrePeginTx, InstanceType<typeof WasmPrePeginTx>>` and `Same<WasmPeginTx, ReturnType<typeof WasmPeginTx.fromJson>>`. But `raw.ts:12-13` and `:21-22` alias both names to the guard class (`export type WasmPeginTx = GuardedPeginTx`), so each reduces to `Same<T, T>` — true for any `T`, and it would still compile if the guard lost half its methods. `CheckConstructor` is now emitted only for the two connectors, which keep the real `Same<typeof X, typeof generated.X>`.

Dropping the old annotation is *correct* — the diverging signature is precisely the declared breaking change — but nothing replaced it. The consequence is that a `VAULT_WASM_COMMIT` bump adding or renaming an engine method leaves the guard silently missing it, with no compile error, on a §1 path. Keep the two genuine new assertions (`OriginalAmounts`, `TrustedRestore`) and add a structural check over the unchanged surface, e.g. `GuardedPeginTx extends generated.WasmPeginTx`.

packages/babylon-tbv-rust-wasm/src/prePeginTransaction.ts
```diff
@@ -0,0 +1,229 @@
...
+      params.vaultKeeperPubkeys.length,
+      params.universalChallengerPubkeys.length,
+      params.minPeginFeeRate,
+    ),
+  );
+  const outputs: BitcoinOutput[] = params.hashlocks.map((hashlock, index) => ({
+    amount: htlcValues[index],
+    script: deriveExpectedPrePeginHtlc(params, hashlock).scriptPubKey,
+  }));
+  if (outputs.some((output) => output.amount < 330n)) {
```
gbarkhatov: This dust guard is on a value that cannot be dust, while the value that *can* be is only caught after the money is committed.

At this point `outputs` holds only HTLC values, and `computeMinHtlcValue = amount + claimValue + anchor + fee`. `peginFees.ts:116-125` shows `computeMinClaimValue` summing three independent `546n` terms, so it floors at 1638 — this branch is unreachable. The value that can actually be dust is `pegInAmounts[i]`, which is checked only `> 0` here via `assertPositiveBigintArray`; its 330-sat floor lives over in `rawPeginTx.ts:156`, for one selected index, at PegIn-build time.

Concrete failure: `pegInAmounts: [90000n, 100n]` constructs, funds and broadcasts a Pre-PegIn cleanly, then dies at `buildPeginTx(_, 1)` — after the BTC is on-chain, leaving only the refund path. Applying the floor to every `pegInAmounts[i]` inside `deriveExpectedPrePegin` (and deleting the unreachable branch) would fail it before broadcast.

packages/babylon-tbv-rust-wasm/src/prePeginTransaction.ts
```diff
@@ -0,0 +1,229 @@
...
+    computeMinHtlcValue(
+      params.txGraphVersion,
+      amount,
+      depositorClaimValue,
+      params.vaultKeeperPubkeys.length,
+      params.universalChallengerPubkeys.length,
+      params.minPeginFeeRate,
+    ),
+  );
+  const outputs: BitcoinOutput[] = params.hashlocks.map((hashlock, index) => ({
```
gbarkhatov, line 152: The create path now rejects an empty universal-challenger roster that the pinned Rust supports — and that this PR's own fee model already handles.

`deriveExpectedPrePegin` builds each HTLC via `deriveExpectedPrePeginHtlc`, which runs `normalizeKeyGroup` over the UC set, and `connectorScripts.ts:39` throws `must not be empty`. Against the pin, `btc-vault crates/vault/src/transactions/prepegin.rs:40` documents the field as "Universal Challengers (M-of-M, **may be empty**)".

The internal contradiction is the strongest evidence: `peginFees.ts:146` branches explicitly on `universal === 0n ? 0n : …`, while `:137` throws only for empty *keepers*. So this PR models zero-UC in its fee math and forbids it in its output derivation. Before this change the restriction bound only `refund.ts`; now it binds `createPrePeginTransaction` — the deposit path. Either lift the empty-group rejection for the UC set specifically and add a zero-UC case to `prepegin-transactions.test.mjs`, or state in the migration guide that zero-UC is deliberately out of scope.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
```diff
@@ -0,0 +1,413 @@
...
+      );
+    }
+    expected.forEach((value, index) => exact(actual[index], value));
+  } else if (expected !== null && typeof expected === 'object') {
+    const object = record(actual);
+    exact(Object.keys(object).sort(), Object.keys(expected).sort());
+    for (const [key, value] of Object.entries(expected))
+      exact(object[key], value);
+  } else if (actual !== expected) {
+    throw new Error('PegIn JSON does not match the trusted transaction.');
```
gbarkhatov: `exact()` reports roughly thirty comparison sites through three fixed strings that name no field, no actual and no expected value.

Transaction version, locktime, outpoint, sequence, every output value and script, payout roles and timelock, HTLC hashlock and timelock, and the prevout all surface as `PegIn JSON does not match the trusted transaction.` This is the same guard-message issue I raised on #2492, and it's worse here because `exact` fans out recursively across a whole object graph rather than checking one field.

Concrete failure: a depositor resuming a saved deposit produces an error support cannot use to separate benign engine-version drift from actual tampering — on the exact path this guard exists to protect. Threading a `path` string through the recursion and interpolating it plus both values would match the style `refund.ts` already uses.

packages/babylon-tbv-rust-wasm/src/peginFees.ts
```diff
@@ -0,0 +1,179 @@
...
+    94n * 4n + quorum * 64n + council * 34n + 2n + 33n + 32n * depth,
+  );
+  const wronglyVsize = vsize(0n, 4n, 65n + 33n + 74n + 162n, 1n);
+  // The pinned Claim estimator counts script + control block as scriptSig bytes.
+  const claimVsize = vsize(67n, 1n, 65n, 2n);
+  const challengerOutput = u64(
+    wronglyVsize * rate + 330n + 546n,
+    'challenger output',
+  );
+  return u64(
```
gbarkhatov, line 143: The new fee and transaction model is written almost entirely in unnamed literals, on a path this PR registers as §1 critical.

Across this file: `11456n`, `34n`, `132n * (21n + 2n)`, `264n`, `162n`, `74n`, `67n`, `43n`, `40n`, `56n`, `13n`, `107n`, `66n`, and `546n` four times — plus `330n` at `prePeginTransaction.ts:152` and `rawPeginTx.ts:156`, `546n` at `prePeginTransaction.ts:166`, `'6a20'` at `:163`, and `240n` / `'51024e73'` at `rawPeginTx.ts:175`. Only the WOTS term carries a derivation.

`330n` and `546n` are `btc-vault`'s `P2TR_DUST_THRESHOLD` and `DUST_AMOUNT` — six inlined copies of two protocol constants, so a pin that moved either floor needs six independent edits with nothing to catch a miss. Beyond the no-magic-numbers rule, §1 obliges the author to line-walk every changed line unaided, which this form makes hard to meet. The package already has `src/constants.ts`; hoisting and naming each term after the Rust function it mirrors would fix both.

packages/babylon-tbv-rust-wasm/src/prePeginTransaction.ts
```diff
@@ -0,0 +1,229 @@
...
+/** Bind funded protocol outputs to the original request. Preserve wallet change. */
+export function assertFundedPrePegin(
+  txHex: string,
+  expected: ExpectedPrePegin,
+): {
+  transaction: BitcoinTransaction;
+  txHex: string;
+  txid: string;
+} {
+  const transaction = decodeTransaction(txHex);
```
gbarkhatov, line 195: `assertFundedPrePegin` leaves `lockTime` and input `sequence` unconstrained. It checks version ≥ 2, non-empty inputs, empty scriptSigs, and pins every protocol output positionally — but not these two, even though the derived template at line 169 fixes both to 0.

The funded parent's txid is what `GuardedPeginTx` binds the PegIn to, so an RBF-signalling or time-locked parent is a parent whose txid can still change after the depositor pre-signs. Trailing wallet change is genuinely fine here, since outputs `0..n-1` are pinned by script *and* value, and total-fee checking is impossible because a raw transaction carries no prevout amounts — so this is the one gap in this function worth closing.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
```diff
@@ -0,0 +1,413 @@
...
+      throw new Error(
+        'PegIn signature does not match the trusted transaction and signer.',
+      );
+    }
+  }
+
+  #checkJson(json: string, transaction: BitcoinTransaction): void {
+    const data = record(parse(json, undefined, parseNumberAndBigInt));
+    const spender = record(data.pegin_input_spender);
+    const depositor = this.#payout.depositor as string;
```
gbarkhatov: Four unchecked `as` casts re-narrow fields that were typed one function earlier. `#payout` is declared `Record<string, unknown>` and then recovered with `as string` here and at line 278, and `as string[]` at 289 and 296 — even though every one of those fields is written from typed locals at lines 141-150.

Rename a `#payout` key and this still compiles, `depositor` silently becomes `undefined`, and the first symptom is `schnorr.verify` failing with a generic signature error rather than a compile error pointing at the rename. Declaring a `PeginPayoutFields` interface would let all four casts go.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
```diff
@@ -0,0 +1,413 @@
...
+          this.#inputConnector.hashlock
+      ) {
+        throw new Error(
+          'PegIn witness preimage does not match the trusted hashlock.',
+        );
+      }
+    }
+  }
+
+  #checkAll() {
```
gbarkhatov, line 405: Every getter re-runs the full verification, Schnorr checks included. `getTxid`, `getTxGraphVersion`, `getVaultScriptPubKey` and `getVaultValue` each call `#checkAll()`, which re-decodes the transaction, re-encodes it, re-parses the JSON and re-verifies every stored and witness signature — despite needing none of that.

`index.ts:141-144` calls four of them in sequence, so reading four fields on a fully-signed PegIn costs 4 × (2N+4) Schnorr verifications. The same shape appears at `rawPrePeginTx.ts:170-186`, where `fromFundedTransaction` constructs a second `GuardedPrePeginTx` and discards the inner one — I checked the pin, and `WasmPrePeginTx::from_funded_transaction` takes `&self` and returns a new object, so that copy isn't needed for correctness. Net effect on the real path is three full `deriveExpectedPrePegin` runs, each with a bitcoinjs taproot tweak per HTLC, per PegIn build. Verifying once in the constructor and caching would keep the guarantee at a fraction of the cost.

gbarkhatov (approved):
Reviewed at `bf4fad8` via three independent passes (7 agents), deduped, with every finding below re-checked against the source and against the pinned Rust.

**The substance here is good.** The guards sit on the real product path, the trusted inputs genuinely originate outside WASM, and the byte-level derivations check out against `btc-vault` at all three pins. Restoration is properly trust-independent — every expectation derives from `trusted.{prePeginParams, fundedPrePeginTxHex, htlcVout, timelockPegin}` rather than from the saved engine JSON, with exactly one exception, which is my first comment. Most of what follows is about *proof and failure modes* rather than wrong transaction bytes.

**CI, since this is the third PR in the stack:** `gh pr checks 2493` returns exactly two jobs — GitGuardian and Greptile. No build, no test, no lint, no `critical-path-check`, no `claude-md-drift`. The same stacked-base gap I raised on #2492 recurs verbatim: `critical-path-check.yml` still triggers `on: pull_request: branches: [main]`, untouched by this PR even though this PR *edits that very file* to add four new paths. You acknowledge it in the description and substitute local run counts (79 engine / 1,974 SDK), which I have no way to verify from here. Worth linking a green run at this head — this is the first change to touch all four paths it registers.

- *"The u8 index cap is justified by a comment that is false against the pin."* One lane argued the comment was wrong because `vault-wasm@3accd8f` declares `htlc_vout: u32`. True, but the wrong layer — `btc-vault crates/vault/src/wasm/api.rs` does `u8::try_from(htlc_vout)` at lines 610, 619, 628, 651, 700 and 753. The engine does enforce u8; the comment is accurate. Two lanes reached opposite conclusions and the source settled it.
- *"Fee-math divergence from the pinned Rust."* None. Every constant traced to `tx_graph/config.rs`, `fee.rs` and `assert_marker.rs`, and the goldens hand-reproduced (v1 → 23694, v2 → 23862; fee v1 → 2086, v2 → 2177).
- *"Witness ordering / signature layout wrong."* Refuted independently by two lanes against `crates/vault/src/sign.rs` — the engine pushes `[preimage, D, VP, VK…, UC…]` then reverses, matching the preimage at `witness[len-3]`.
- *"`assertFundedPrePegin`'s empty-scriptSig rule excludes a real funding type."* It doesn't — the SDK supports only P2TR, P2WPKH and P2WSH, all empty-scriptSig.

**Verified clean:** §9's differential requirement is genuinely met — `pegin-fees.test.mjs` and `prepegin-transactions.test.mjs` compare the TS reimplementations against real WASM over golden values, CompactSize/ScriptNum boundaries, u64 overflow edges, rosters to 256, and ~100 randomised draws per graph version. Well past "a single hardcoded vector". The `bigint[]` breaking change is real and correctly motivated: `assertPositiveBigintArray` now runs *before* `new BigUint64Array(...)`, closing the mod-2⁶⁴ wrap the old ordering allowed. PegIn and Pre-PegIn shapes match the pin, and the zero-input SegWit marker splice is correct rather than a hand-rolled bug. And no test passes vacuously — all three new `.mjs` files `await getWasmBindings()` at module top level with no try/catch or skip, which matters more than usual given no CI job runs here.

jonybur:

I checked each of the ten open threads against a prepared fix set. None of the ten is closed by it, so I have not pushed anything. A push would dismiss the standing approval (`dismiss_stale_reviews_on_push` is on) and buy nothing.

### Merge order

`verify` and `detect` are the two required checks, and both trigger only on `pull_request: branches: [main]`. Neither has ever run on this PR. It must retarget to `main` after #2492 merges, which itself waits on #2481.

### Still open (10)

- **`rawPeginTx.ts:336` — restoration guard fails open.** The conditional spread on `Object.hasOwn(data, 'prepegin_htlc_prevout')` is unchanged, so a saved PegIn JSON that omits the field or sets it to `null` still skips the HTLC value and scriptPubKey binding. The untrusted input still decides whether the check applies.
- **`rawPeginTx.ts:190` — signature digest unchecked.** `#sighashes` is still built with `preimageWitnessV1` and verified against fixtures produced by the same call. No cross-check against `computePeginInputSighash`.
- **`wasm-loader.test.mjs:61` — type parity.** The two checks are still tautologies. A `VAULT_WASM_COMMIT` bump that renames an engine method still compiles.
- **`prePeginTransaction.ts:152` — dust floor on the wrong value.** The unreachable branch on HTLC outputs is still there, and `pegInAmounts[i]` is still unchecked before funding and broadcast.
- **`prePeginTransaction.ts:148` — empty universal-challenger set.** Still rejected on the deposit path, though the pinned Rust allows it and this PR's own fee math models it. Neither the code nor the migration guide has moved.
- **`rawPeginTx.ts:56` — `exact()` error messages.** Still three fixed strings across about thirty comparison sites. No field path, no actual, no expected. A restore failure still cannot tell engine drift from tampering.
- **`peginFees.ts:116` — unnamed literals.** This file is untouched. `11456n`, `34n`, `264n`, `162n`, `74n`, `67n`, `43n`, `40n`, `56n`, `13n`, `107n`, `66n` and four copies of `546n` are all still inline, on a path this PR registers as section 1.
- **`prePeginTransaction.ts:196` — `lockTime` and `sequence`.** `assertFundedPrePegin` still leaves both unconstrained, so an RBF-signalling or time-locked parent keeps a txid that can change after the depositor pre-signs.
- **`rawPeginTx.ts:277` — unchecked casts.** The four `as` casts on `#payout` are unchanged. A key rename still yields `undefined` instead of a compile error.
- **`rawPeginTx.ts:369` — repeated verification.** Every getter still re-runs `#checkAll()`, Schnorr checks included, and `rawPrePeginTx.ts` still builds the second `GuardedPrePeginTx` copy.

I will work through these before asking for another look.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
```diff
@@ -0,0 +1,413 @@
...
+  getVaultScriptPubKey(): string {
+    this.#checkAll();
+    return hex(this.#transaction.outputs[0].script);
+  }
+  getVaultValue(): bigint {
+    this.#checkAll();
+    return this.#transaction.outputs[0].amount;
+  }
+  free(): void {
+    this.#inner.free();
```
jeremy-babylonlabs, line 455: Same missing `#freed` latch as #2492 — a second `free()` or dispose throws `null pointer passed to rust`. Mirror `GuardedPrePeginHtlcConnector.free()`.

packages/babylon-tbv-rust-wasm/src/rawPrePeginTx.ts
```diff
@@ -0,0 +1,310 @@
...
+  getHtlcAddress(htlcVout: number): string {
+    const index = this.#index(htlcVout);
+    return this.#check(
+      this.#inner.getHtlcAddress(index),
+      htlcAddress(this.#expected, index),
+      'HTLC address',
+    );
+  }
+  free(): void {
+    this.#inner.free();
```
jeremy-babylonlabs, line 319: Needs the `#freed` latch too, and it is more reachable here: `fromFundedTransaction` already calls `copy.free()` on its error path, so a caller's own `finally { free() }` after a failed copy is a realistic double-free.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
```diff
@@ -0,0 +1,413 @@
...
+      throw new Error('PegIn witness has an unexpected number of items.');
+    }
+    exact(witness.slice(-2).map(hex), [
+      hex(this.#htlc.hashlockScript),
+      hex(this.#htlc.hashlockControlBlock),
+    ]);
+    const keys = partial ? [depositor] : this.#signatureKeys;
+    keys.forEach((pubkey, index) => {
+      this.#verifySignature(witness[index], pubkey);
+      const saved = partial ? stored.at(-1) : stored[index];
```
jeremy-babylonlabs, line 388: Every getter re-runs `#checkAll`, which re-parses the engine JSON and re-verifies every Schnorr signature. Worth caching the result after construction — unless re-detecting engine state drift on each read is deliberate, in which case a comment saying so would help.

jeremy-babylonlabs (commented):
Thorough — original bigint amounts captured before u64 conversion, independent fee math with overflow checks, canonical byte re-encode, funded-output binding with change preserved, per-key Schnorr verification, strict JSON key sets, and negative tests for mutations and lossy amounts. Two classes repeat the missing free latch from #2492.

Only two checks are passing because the verification workflows run against `main` — worth confirming CI after the retarget.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
jonybur: Fixed at `90ebb21d`. `pegin-restore.test.mjs` now pins the digest against the Rust oracle, once per graph version: `preimageWitnessV1` against `computePeginInputSighash`. All three match, so this closes a proof gap rather than a wrong derivation.

Only the SIGHASH_DEFAULT entry is pinned. The export takes no sighash-type argument, so there is nothing to compare hash type 1 against.

packages/babylon-tbv-rust-wasm/src/peginFees.ts
jonybur: Fixed at `90ebb21d`. `DUST_AMOUNT`, `P2TR_DUST_THRESHOLD`, `WOTS_SCRIPT_BYTES`, `PER_KEY_PUSH_BYTES`, `P2TR_OUTPUT_BYTES`, `OUTPOINT_AND_SEQUENCE_BYTES` and `P2A_ANCHOR_VALUE` are named once in `peginFees.ts` and used at every site, including the three outside that file.

The two dust values cite `btc-vault crates/vault/src/lib.rs` at 2c1177ec, 27c0062b and e1e50f66, where they are `DUST_AMOUNT = 546` and `P2TR_DUST_THRESHOLD = 330`. This also settles point 3 of my own review.

I kept them in `peginFees.ts` rather than `src/constants.ts`, for the reason you would have hit next: `constants.ts` is not in `critical-path-check.yml`, CODEOWNERS or the drift check, so moving section 1 values there would take them out from under the gate this PR extends.

Terms I did not name: the witness-size arithmetic that has no Rust counterpart to name it after (`65n`, `132n * (21n + 2n)`, `94n * 4n`, `264n`). Each still sits under the comment that explains the estimator it reproduces. Say the word if you want those named too.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
jonybur: Fixed at `90ebb21d`. Both tautologies are replaced with a structural comparison of the public instance surface:

- `Check<Same<Omit<WasmPeginTx, never>, Omit<generated.WasmPeginTx, never>>>` - `Omit` strips the `#private` brand that made a direct `Same<>` fail.
- The Pre-PegIn check omits `fromFundedTransaction` and `buildPeginTx`, the two methods whose return types diverge by design.

I checked it catches the case you describe rather than trusting the green run: deleting `getVaultValue()` from `GuardedPeginTx` makes the gate fail with `Type 'false' does not satisfy the constraint 'true'`, and restoring it passes. The old form compiled fine with the method gone.

The deprecation probe moved with it, since it looks for the class name at the surface-check site.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
jonybur: Fixed at `90ebb21d`, taking the strict branch. `prepegin_htlc_prevout: this.#prevout` is unconditional now, so the saved object cannot opt out of binding the spent HTLC value and scriptPubKey.

The test that pinned the tolerant behaviour is flipped: `v1/v2/v3 rejects a saved object with no prevout metadata` asserts both the omitted and the null form throw, and the error names the field.

`raw-engine-migration.md` records the consequence: a saved transaction without the field is rejected and must be rebuilt from its request and funded parent. I found no evidence of a real older format that omitted it - the fixtures the engine produces all carry it - so I did not add a compatibility path for one.

packages/babylon-tbv-rust-wasm/src/prePeginTransaction.ts
jonybur: Half taken at `90ebb21d`.

Removed, because both are unreachable: the `outputs.some((output) => output.amount < 330n)` branch in `deriveExpectedPrePegin`, whose values are HTLC totals that always clear dust, and the `expected.depositorClaimValue < 330n` clause in `rawPeginTx.ts`.

Not added: a `pegInAmounts[i] < 330n` floor inside `deriveExpectedPrePegin`. The pinned engine accepts amounts under 330, and `prepegin-transactions.test.mjs` draws `pegInAmounts` from `randomInt(1, 1_000_000)` across three versions, so a blanket floor would both narrow the guard below the engine and fail the differential that test asserts. The per-index check stays where it is, on the PegIn build path, and now names the amount and the threshold in its error.

One correction to the framing: that check does not run "after the money is committed". It runs while the PegIn transaction is being built from the funded parent, before anything is signed.

packages/babylon-tbv-rust-wasm/src/prePeginTransaction.ts
jonybur: Checked against the pinned binary at `3accd8f`; no change.

The engine rejects an empty universal-challenger roster itself, at both the `WasmPrePeginTx` and `WasmPrePeginHtlcConnector` constructors, with `Empty key list for UniversalChallengers`. The guard only moves the same rejection earlier and changes the message, so nothing that the pinned Rust supports is being narrowed.

The zero-UC branches in `peginFees.ts` are not evidence to the contrary: they model the pinned Rust estimators, which do accept zero, and `pegin-fees.test.mjs` pins them against the engine at the `[1, 0]` roster. If a deployment ever needs zero universal challengers, that is an engine change, not a guard change.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
jonybur: Fixed at `90ebb21d`. `exact()` and `record()` thread a field path, so the three fixed strings now read as `pegin.tx.output[0].script_pubkey`, `pegin.pegin_payout_connector.timelock_pegin`, and so on.

I did not interpolate the actual and expected values, and I would rather not: `exact` recurses into `tx.input[0].witness`, and on a final witness the 32-byte HTLC preimage sits three items from the end. A mismatch there would put the vault secret into an error string. If you want a value alongside the path, I can emit a length or a hash instead.

The same change is worth making in #2492, where you raised it as well.

packages/babylon-tbv-rust-wasm/src/prePeginTransaction.ts
jonybur: No change, and I want to be explicit about why rather than leave it implied.

A `sequence < 0xfffffffe` rejection is not available: `prepegin-transactions.test.mjs` builds its funded fixture with `sequence: 0xfffffffd`, which is BIP-125 opt-in RBF, and feeds it to `fromFundedTransaction` in nine places. Those are the differential tests against the pinned engine, which accepts it.

A `lockTime !== 0` rejection would pass the tests but narrow the guard below the engine and reject a standard wallet's anti-fee-sniping locktime on the `/raw` path. That is the same wallet-compatibility hazard we already ruled out for the scriptSig rule.

The vault path itself emits neither an RBF sequence nor a non-zero locktime. The txid is bound by the guard through the funded parent it is given, so a caller who hands us a replaceable parent gets a guard that matches the parent they chose.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
jonybur: Fixed at `90ebb21d`. `#payout` is typed as a `PeginPayoutFields` interface (`depositor`, `vault_provider`, `vault_keepers`, `universal_challengers`, `timelock_pegin`), so all four `as` casts are gone and a renamed key is a compile error.

`#inputConnector` keeps `Record<string, unknown>`. Its only narrowing use is a `!==` against `this.#inputConnector.hashlock`, which is legal against `unknown`, so typing it would be surface for its own sake. Happy to do it if you want the symmetry.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
jonybur: Not taken, and this one is a judgement call rather than a refutation.

Caching the verification result would trade a re-check on every read for speed, on the class whose entire purpose is that re-check. The argument that `#inner` cannot change is sound today, but it makes the guard's safety depend on a fact about the surrounding code rather than on what the getter does. On a section 1 path I would rather pay three derivations.

The second half is a real waste and I agree with it: `fromFundedTransaction` builds a `GuardedPrePeginTx` whose expectation it already holds. The fix needs a private constructor path that takes an already-derived `ExpectedPrePegin`, which is new surface on a section 1 class, so I would rather do it deliberately than fold it into a review pass. I will open it as a follow-up unless you want it here.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
jonybur: Fixed at `90ebb21d`. `GuardedPeginTx.free()` now early-returns on a `#freed` latch, and `v1/v2/v3 releases the engine object only once` pins that a second free() and a dispose() after it both stay quiet.

packages/babylon-tbv-rust-wasm/src/rawPrePeginTx.ts
jonybur: Fixed at `90ebb21d`, and you are right that this one is the reachable case. `GuardedPrePeginTx.free()` has the latch too, and the test covers exactly the path you name: after `fromFundedTransaction` frees its copy, a caller's own `free()` on that copy is quiet.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
jonybur: Deliberate, and now said out loud at `90ebb21d`.

The comment above the getters states it: the engine object behind `#inner` is the thing being checked, so a cached verdict would report on the state at construction rather than the state being read. The cost is one JSON parse and the signature checks per read, which is what the guarantee costs.

The related waste is real though, and I have not taken it here: `fromFundedTransaction` builds a `GuardedPrePeginTx` whose expectation it already holds. That needs a private constructor path taking an already-derived `ExpectedPrePegin`, which is new surface on a section 1 class, so I would rather do it deliberately than fold it into a review pass.

jonybur:
Review fixes are pushed as `90ebb21d`.

Seven of the ten threads produced changes; three are answered with the reasoning in the thread. Your three top-level points are covered too: `free()` is idempotent on both new classes, both constructors now release inside a `try` so the guard error cannot be masked, and the fee-model constants are named and cited.

Two behaviour changes worth naming: `WasmPeginTx.fromJson` now requires `prepegin_htlc_prevout` and rejects a saved object that omits it or stores null, which `raw-engine-migration.md` records; and the JSON mismatch errors name the field path, without ever interpolating a value, because that walk reaches the witness that carries the HTLC preimage.

Two checks were proven rather than assumed. The new type-parity gate fails when `getVaultValue()` is deleted from the guard, where the old tautology compiled. The input digest now matches `computePeginInputSighash` at v1, v2 and v3, so the suite no longer signs with the same derivation it verifies.

jeremy-babylonlabs's three later threads are covered by the same commit: both new classes have the release latch, with a test each, and the deliberate re-check on every getter is now stated in the code.

Local checks at this head: 85 engine tests across four files, 1,974 SDK tests in 117 files plus 9 facade checks, 3,809 Vault tests, 34 Chromium tests, build, lint, and `docs:clean` with no diff. This PR is still stacked on #2492 and does not run a gating job; it needs a rebase onto #2492's merged head before it can go green on its own.

jonybur:
Merged the updated base (#2492 at `a7e9a1b2`) into this branch as `4f333114`.

The merge found a real regression that neither branch shows on its own, and the harness #2481 re-armed is what caught it.

This PR makes `raw.ts`'s generated-glue import type-only, which is right: all four classes are guarded now, so the re-exports come from the guard modules. That makes `rawHtlcConnector.js` the first value import. #2481 removed the `node_modules` symlink from the lazy-entry harness, so on the merged tree the raw entries started failing on `bitcoinjs-lib` rather than on the missing WASM glue, and `check-lazy-entries.js` failed.

The fix keeps both guarantees rather than weakening either. The facade loop keeps its `node_modules`-free directory, so a static Bitcoin edge on the Ethereum-only root still fails it; the raw loop gets its own copy with `node_modules`, so it fails for the reason it is meant to. I planted a static `bitcoinjs-lib` import on `index.ts` to confirm the facade half is still armed, and removed it again.

This is the import-reorder risk gbarkhatov raised on #2481; it arrived one PR later.

Documentation conflicts were resolved so that CLAUDE.md and SECURITY.md state that all four classes are guarded, while keeping #2481's correction that the check is internal consistency against the inputs given and the caller must still verify those inputs on-chain. The migration guide keeps this PR's deliberate removal of the retained-operation audit; the facade-guard fact moved into the connector section so it is not lost.

Re-verified at this head: 90 engine tests across four files, 1,978 SDK tests in 117 files plus 9 facade checks, 3,809 Vault tests, 34 Chromium tests, build, lint, and `docs:clean` with no diff.

jonybur:
@greptileai review the full PR at current commit aa3a9a2e08e126dca0d98d74448abed6f22fce83. The review fixes are published and local checks pass. Please check the current code, report any remaining findings, and give the confidence score for this commit.

packages/babylon-tbv-rust-wasm/src/rawPrePeginTx.ts
```diff
@@ -0,0 +1,324 @@
...
+      const unfunded = copy.#inner;
+      copy.#inner = unfunded.fromFundedTransaction(funded.txHex);
+      unfunded.free();
+      copy.#txHex = funded.txHex;
+      copy.#txid = funded.txid;
+      copy.#funded = true;
+      copy.#checkAll();
+      return copy;
+    } catch (error) {
+      copy.free();
```
gbarkhatov, line 192: Minor. `aa3a9a2e0` moved both constructors to the `AggregateError` shape (`:101-111` here, `rawPeginTx.ts:235-245`), but this third cleanup site still does `copy.free(); throw error;`. If the release throws, it replaces the mismatch error, e.g. "WASM Pre-PegIn transaction does not match the original request", on the funded-reconstruction path. That is the case the merge set out to fix. It's low-reachability (only a failing engine release can throw here), the same as #2481 T5/T12, which was fixed anyway. The throwing-`free` harness in `prepegin-transactions.test.mjs:80-97` only runs the constructor, so one output mutation at `:180-218` under that harness would pin this site too.

packages/babylon-tbv-rust-wasm/src/peginFees.ts
```diff
@@ -0,0 +1,206 @@
...
+export const DUST_AMOUNT = 546n;
+export const P2TR_DUST_THRESHOLD = 330n;
+
+// Transaction sizing terms, in bytes.
+// An outpoint (36) plus its sequence (4).
+const OUTPOINT_AND_SEQUENCE_BYTES = 40n;
+// A P2TR output: value (8), script length (1) and the 34-byte scriptPubKey.
+const P2TR_OUTPUT_BYTES = 43n;
+// One 32-byte x-only key push plus its length byte, per signer.
+const PER_KEY_PUSH_BYTES = 34n;
```
gbarkhatov, line 16: Low. The comment gives the wrong breakdown: "one 32-byte x-only key push plus its length byte" is 33. The pinned Rust (`btc-vault crates/vault/src/tx_graph/config.rs` @`e1e50f66`) counts the 34 as `push32 + key(32) + OP_CHECKSIG/ADD`.

The constant is also reused for a different term. The trailing `+ PER_KEY_PUSH_BYTES` in the Assert witness (`:121`) is the depth-0 control block plus its length byte (`claim_assert.rs:232` → `sizes::control_block_size(0)`, `fee.rs:26`). Both happen to be 34, so the differential stays green, and it would stay green if the name led someone to edit the wrong term. That undoes what T7's naming was for on a §1/§2 file.

Suggest fixing the comment and giving `:121` its own constant, e.g. `ASSERT_CONTROL_BLOCK_WITNESS_BYTES = 1n + 33n`, citing `control_block_size(0)`.

packages/babylon-tbv-rust-wasm/src/rawPeginTx.ts
```diff
@@ -0,0 +1,460 @@
...
+      }
+    }
+  }
+
+  #checkAll() {
+    const txHex = this.#inner.toHex();
+    const transaction = decodeTransaction(txHex);
+    exact(hex(RawOldTx.encode(transaction)), this.#unsignedHex);
+    exact(this.#inner.getTxGraphVersion(), this.#version);
+    exact(this.#inner.getTxid(), this.#txid);
```
gbarkhatov, line 410: T6 follow-up. The path is now threaded through the recursive JSON walk (`:346-375`), but direct calls still fall back to `path = 'pegin'` (`exact` at `:49`, `record` at `:35`). That covers `#checkAll`'s bytes, version, txid, vault script and value checks here (`:408-415`), the signature field-name check (`:276`), the witness script/control block (`:381-384`), the witness-vs-stored signature (`:389`), and the `record()` calls at `:275`, `:314`, `:315` and `:326`.

So a `getTxid()` disagreement or a wrong witness control block now reads "PegIn JSON pegin does not match…", which points support at the saved JSON's root for failures that aren't in the JSON. That's arguably worse than the old fixed strings. Explicit paths like `'engine.getTxid'` or `'pegin.tx.input[0].witness[-2]'` carry no value, so the HTLC-preimage concern still holds.

gbarkhatov (approved):
Follow-up to my approval at `bf4fad83f3`. Since then there are three commits: `90ebb21de` (fixes), `4f3331149` (#2492's fixes merged in, plus hand edits to CLAUDE.md, SECURITY.md, the guide and `check-lazy-entries.js`) and `aa3a9a2e0` (#2492's head merged in, with `AggregateError` added to both constructors). Nothing blocks merge. The no-repo-CI point for this stack is already on #2492.

| # | Prior finding | Verdict at `aa3a9a2e` |
|---|---|---|
| T1 | A saved JSON could skip the `prepegin_htlc_prevout` check | **Fixed.** Missing and null are both rejected (`rawPeginTx.ts:372-374`, test `pegin-restore.test.mjs:346-357`) |
| T2 | No independent oracle for the signature digest | **Fixed.** Compared with `wasm.computePeginInputSighash` for v1/v2/v3 (`:315-344`) |
| T3 | Type-parity checks were tautologies | **Fixed.** Compares the public instance surface with `Omit<>` |
| T4 | Dust guard on a value that can't be dust; no floor on `pegInAmounts` | **Declined, with reason.** The unreachable branch is removed |
| T5 | Create path rejects an empty universal-challenger list | **Declined, and my premise was wrong.** The engine rejects that itself (`Empty key list for {0}`, `btc-vault error.rs` @`e1e50f66`) |
| T6 | `exact()` errors name no field | **Partially fixed**, see the inline on `rawPeginTx.ts:410` |
| T7 | Unnamed numbers in fee and transaction code | **Partially fixed.** Dust, anchor and sizing values are named; the witness arithmetic is declined. One new name is wrong, see the inline on `peginFees.ts:16` |
| T8 | Funded parent's `lockTime`/`sequence` unchecked | **Declined, with reason** (the pinned engine accepts them) |
| T9 | Four `as` casts on `#payout` | **Partially fixed.** Both `as string` casts are gone. `(keys as string[])` remains at `:328` and `:335`; it is redundant now that `#payout` is typed, so the rename risk is closed |
| T10 | Getters re-run full verification; redundant copy in `fromFundedTransaction` | **Declined, with reason** (§1 path, so pay for the derivations) |

Checked, and fine:
- **Pre-PegIn values:** every HTLC value is independently derived with `computeMinHtlcValue`, including the v2/v3 `P2A_ANCHOR_VALUE`, and compared byte for byte.
- **PegIn values:** the expected tx is built from `pegInAmounts[index]`, the claim value and the anchor, then compared on bytes, txid, vault script and value.
- **Bigint:** amounts stay bigint end to end, with lossless-json parsing for saved JSON.
- **Restore:** signature digests use the trusted `#prevout`, so a forged prevout fails `schnorr.verify`.
- **`refund.ts`:** unchanged here, and it still does every §1 `/raw` cross-check.
- **`peginFees.ts`:** every literal swap keeps its value. The deleted HTLC dust branch was unreachable (`computeMinClaimValue` ≥ 3 × 546).
- **`check-lazy-entries.js`:** the edit keeps #2481 T8. The facade loop still runs without `node_modules`.
- **Other:** no new dependencies in the delta, the §4 expander wrappers are untouched, and `f51a6fcd` is an ancestor of head.

Nit, unanchorable: the comment explaining the digest oracle (`pegin-restore.test.mjs:302-304`) sits above the release-latch test, not above the digest test at `:315`.

jonybur:
Closing in favor of #2516. It rolls back the raw engine entry instead of guarding it. The decision is recorded in #2361.

PR #2495, opened 2026-09-10
fix(ledger-vault): match the device's unmasked refund sequence check

jrwbabylonlab:
- Refund gate now mirrors firmware `b0c0ac4d` (V-005): `nSequence` must equal the leaf CSV
exactly, and the CSV must fit the BIP-68 block-count field. The old masked compare let
through PSBTs the device rejects. Bound matches btc-vault's `NonZeroU16` refund timelock.
- `deviceCaps.ts` re-pinned to `b0c0ac4d`; fee-rate cap `0xffffffff` → `10_000n` with its
provenance (vaultd config default; contract caps at 1000, so unreachable). Firmware
line citations refreshed.

### DIFF
--- a/packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
@@ -2,8 +2,8 @@
  * Refund-PSBT classification and augmentation (#2371).
  *
  * The leaf grammar mirrors the firmware's own parser byte-for-byte —
- * `fw:sign_psbt_validate_helpers.c:77-148` (`parse_refund_leaf_script`) @
- * `ff1e1ce17` — so every accept/reject vector here is a firmware-grammar pin,
+ * `fw:sign_psbt_validate_helpers.c:77-156` (`parse_refund_leaf_script`) @
+ * `b0c0ac4d` — so every accept/reject vector here is a firmware-grammar pin,
  * including the shapes the firmware is deliberately loose about (OP_PUSHDATA1,
  * a non-minimal positive CScriptNum).
  */
@@ -158,10 +158,17 @@ describe("parseRefundLeafScript (firmware-grammar mirror)", () => {
     });
   });
 
-  it("parses the maximum 4-byte CScriptNum the firmware admits", () => {
-    expect(parseRefundLeafScript(refundLeaf(DEPOSITOR_XONLY, Buffer.from([0x04, 0xff, 0xff, 0xff, 0x7f])))).toEqual({
+  it("parses the maximum CSV the firmware admits — the BIP-68 block-count field (0xffff)", () => {
+    expect(parseRefundLeafScript(refundLeaf(DEPOSITOR_XONLY, Buffer.from([0x03, 0xff, 0xff, 0x00])))).toEqual({
       leafKeyHex: DEPOSITOR_XONLY,
-      csv: 0x7fffffff,
+      csv: 0xffff,
+    });
+  });
+
+  it("parses a 4-byte CScriptNum push — the firmware's length cap — when its value fits the field", () => {
+    expect(parseRefundLeafScript(refundLeaf(DEPOSITOR_XONLY, Buffer.from([0x04, 0xff, 0xff, 0x00, 0x00])))).toEqual({
+      leafKeyHex: DEPOSITOR_XONLY,
+      csv: 0xffff,
     });
   });
 
@@ -170,6 +177,8 @@ describe("parseRefundLeafScript (firmware-grammar mirror)", () => {
     ["OP_1NEGATE push", Buffer.from([0x4f])],
     ["negative CScriptNum (sign bit set)", Buffer.from([0x01, 0x80])],
     ["zero CScriptNum", Buffer.from([0x01, 0x00])],
+    ["CSV one above the BIP-68 block-count field (0x10000)", Buffer.from([0x03, 0x00, 0x00, 0x01])],
+    ["CSV 0x7fffffff — the 4-byte maximum the firmware accepted before V-005", Buffer.from([0x04, 0xff, 0xff, 0xff, 0x7f])],
     ["5-byte CScriptNum", Buffer.from([0x05, 0x01, 0x00, 0x00, 0x00, 0x00])],
     ["OP_PUSHDATA1 with a 5-byte length", Buffer.from([0x4c, 0x05, 0x01, 0x00, 0x00, 0x00, 0x00])],
     ["push extending past the script end", Buffer.from([0x04, 0x01])],
@@ -290,6 +299,11 @@ describe("assertRefundPsbtSignable", () => {
     expect(() => assertRefundPsbtSignable(classified({ sequence: 144 }))).toThrow(/sequence/);
   });
 
+  it("rejects a sequence with a bit above the block-count field even though the low 16 bits match the CSV", () => {
+    // A masked compare accepted this; the device compares unmasked (V-005).
+    expect(() => assertRefundPsbtSignable(classified({ sequence: 0x00010000 + CSV_2016 }))).toThrow(/sequence/);
+  });
+
   it("rejects an input without a witnessUtxo", () => {
     expect(() => assertRefundPsbtSignable(classified({ dropWitnessUtxo: true }))).toThrow(/witnessUtxo/);
   });
@@ -327,7 +341,7 @@ describe("augmentPsbtForRefund", () => {
     expect(inputDeriv![0].path).toBe("m/86'/1'/0'/0/0");
     expect(inputDeriv![0].leafHashes).toEqual([]);
 
-    // The asymmetry the device demands (`fw:sign_psbt_validate.c:1005-1057`):
+    // The asymmetry the device demands (`fw:sign_psbt_validate.c:1025-1069`):
     // the output entry is keyed by the scriptPubKey's witness program, which
     // is the BIP-86 TWEAK of the depositor key — never the depositor key itself.
     const outputDeriv = augmented.data.outputs[0].tapBip32Derivation;
--- a/packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
+++ b/packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
@@ -77,6 +77,7 @@ function buildRefundShapedPsbt(
     version?: number;
     locktime?: number;
     sequence?: number;
+    sighashType?: number;
   } = {},
 ): string {
   const leaf = overrides.leaf ?? refundLeaf(DEPOSITOR_XONLY, PUSH_2016);
@@ -107,6 +108,9 @@ function buildRefundShapedPsbt(
         }),
     tapInternalKey: Buffer.from(NUMS_XONLY, "hex"),
   });
+  // Assigned directly: bip174's addInput `canAdd` treats a sighashType of 0 as
+  // falsy and refuses it, while `fromHex` decodes an explicit 0 like any other.
+  if (overrides.sighashType !== undefined) psbt.data.inputs[0].sighashType = overrides.sighashType;
   if (overrides.extraInput) {
     psbt.addInput({
       hash: Buffer.alloc(32, 0x33),
@@ -176,9 +180,10 @@ describe("parseRefundLeafScript (firmware-grammar mirror)", () => {
     ["OP_0 push", Buffer.from([0x00])],
     ["OP_1NEGATE push", Buffer.from([0x4f])],
     ["negative CScriptNum (sign bit set)", Buffer.from([0x01, 0x80])],
+    ["negative 4-byte CScriptNum (sign bit in the top byte)", Buffer.from([0x04, 0x00, 0x00, 0x00, 0x80])],
     ["zero CScriptNum", Buffer.from([0x01, 0x00])],
     ["CSV one above the BIP-68 block-count field (0x10000)", Buffer.from([0x03, 0x00, 0x00, 0x01])],
-    ["CSV 0x7fffffff — the 4-byte maximum the firmware accepted before V-005", Buffer.from([0x04, 0xff, 0xff, 0xff, 0x7f])],
+    ["CSV 0x7fffffff — the 4-byte maximum the firmware accepted before the refund-sequence fix", Buffer.from([0x04, 0xff, 0xff, 0xff, 0x7f])],
     ["5-byte CScriptNum", Buffer.from([0x05, 0x01, 0x00, 0x00, 0x00, 0x00])],
     ["OP_PUSHDATA1 with a 5-byte length", Buffer.from([0x4c, 0x05, 0x01, 0x00, 0x00, 0x00, 0x00])],
     ["push extending past the script end", Buffer.from([0x04, 0x01])],
@@ -288,11 +293,15 @@ describe("assertRefundPsbtSignable", () => {
   });
 
   it("rejects a sequence with the BIP-68 disable flag set", () => {
-    expect(() => assertRefundPsbtSignable(classified({ sequence: 0x80000000 + CSV_2016 }))).toThrow(/sequence/);
+    expect(() => assertRefundPsbtSignable(classified({ sequence: 0x80000000 + CSV_2016 }))).toThrow(
+      /disable\/time flags clear/,
+    );
   });
 
   it("rejects a sequence with the BIP-68 time-based flag set", () => {
-    expect(() => assertRefundPsbtSignable(classified({ sequence: 0x00400000 + CSV_2016 }))).toThrow(/sequence/);
+    expect(() => assertRefundPsbtSignable(classified({ sequence: 0x00400000 + CSV_2016 }))).toThrow(
+      /disable\/time flags clear/,
+    );
   });
 
   it("rejects a sequence that does not encode exactly the leaf CSV", () => {
@@ -300,8 +309,16 @@ describe("assertRefundPsbtSignable", () => {
   });
 
   it("rejects a sequence with a bit above the block-count field even though the low 16 bits match the CSV", () => {
-    // A masked compare accepted this; the device compares unmasked (V-005).
+    // A masked compare accepted this; the device compares unmasked.
     expect(() => assertRefundPsbtSignable(classified({ sequence: 0x00010000 + CSV_2016 }))).toThrow(/sequence/);
+  });
+
+  it("accepts an explicit SIGHASH_DEFAULT on the refund input", () => {
+    expect(() => assertRefundPsbtSignable(classified({ sighashType: 0 }))).not.toThrow();
+  });
+
+  it("rejects an explicit SIGHASH_ALL — the device admits only SIGHASH_DEFAULT on this path", () => {
+    expect(() => assertRefundPsbtSignable(classified({ sighashType: 1 }))).toThrow(/sighash/);
   });
 
   it("rejects an input without a witnessUtxo", () => {
--- a/packages/babylon-ledger-vault-signer/src/deviceCaps.ts
+++ b/packages/babylon-ledger-vault-signer/src/deviceCaps.ts
@@ -41,6 +41,13 @@ export const DEVICE_MAX_VAULTS_PER_INTENT = 10;
  */
 export const DEVICE_MAX_BASE_FEE_RATE_SAT_PER_VB = 10_000n;
 
+/**
+ * Upper bound on `prepegin_max_fee` (`vault_tlv.c:170`; `PREPEGIN_MAX_FEE_LIMIT`
+ * at `vault_constants.h:98`). Firmware-only — no btc-vault counterpart, and
+ * the SDK's sized fee cannot approach it at the contract's 1000 sat/vB cap.
+ */
+export const DEVICE_MAX_PREPEGIN_FEE_SATS = 100_000_000n;
+
 /** Shared lower bound for the CSV/refund timelocks (`vault_constants.h:103`). */
 export const DEVICE_TIMELOCK_MIN_BLOCKS = 72;
 
--- a/packages/babylon-ledger-vault-signer/src/deviceCaps.ts
+++ b/packages/babylon-ledger-vault-signer/src/deviceCaps.ts
@@ -1,6 +1,6 @@
 /**
  * Ledger vault-app intent caps, mirrored from the firmware
- * (`app-babylon-vault` @ `8f99b8b`: `vault_tlv.c`, `vault_intent.h`,
+ * (`app-babylon-vault` @ `b0c0ac4d`: `vault_tlv.c`, `vault_intent.h`,
  * `vault_constants.h`). The device rejects violations with SW_INCORRECT_DATA
  * and a dead session, so we fail comprehensibly BEFORE the ceremony.
  *
@@ -20,7 +20,7 @@ export const DEVICE_MAX_VAULT_CORE_VERSION = 3;
 
 /**
  * Keeper and universal-challenger counts, each `[1, 32]` (`vault_intent.h:8-9`,
- * checked `vault_tlv.c:130,137`). Far below the protocol: the contract allows
+ * checked `vault_tlv.c:148,155`). Far below the protocol: the contract allows
  * 1500 universal challengers and sets no keeper cap at all.
  */
 export const DEVICE_MAX_PARTICIPANTS_PER_ROLE = 32;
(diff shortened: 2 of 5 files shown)


packages/babylon-ledger-vault-signer/src/deviceCaps.ts
```diff
@@ -34,23 +34,25 @@ export const DEVICE_MAX_VAULTS_PER_INTENT = 10;
...
  * Fee-rate bounds (sat/vB): the intent parser rejects rate == 0 AND
- * rate > UINT32_MAX (`vault_tlv.c:73`), so the `>= 1` floor is both a device
- * check and the contract invariant.
+ * rate > 10000 (`vault_tlv.c:75`). The ceiling is vaultd's critical-tx config
+ * default (`btc-vault crates/vaultd/src/config/btc.rs:19` @ d0d139305), not a
+ * protocol constant; the contract caps `feeRate` at 1000 (`ProtocolParams.sol:44`),
+ * so a contract-sourced rate can never reach it.
  */
-export const DEVICE_MAX_BASE_FEE_RATE_SAT_PER_VB = 0xffffffffn;
+export const DEVICE_MAX_BASE_FEE_RATE_SAT_PER_VB = 10_000n;
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — no Ledger parity claim exists; SDK is vendor-neutral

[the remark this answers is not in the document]
jrwbabylonlab: not planning — that sentence is the vsize model, still 500/55

jeremy-babylonlabs (dismissed):

jonybur (dismissed):
Sequence check now matches the device's unmasked comparison, and the fee-rate ceiling update is documented with its provenance. Tests cover the new validation boundaries, firmware citations are consistent throughout, and no dead code or silent fallbacks were introduced. CI green.

packages/babylon-ledger-vault-signer/src/envelope.ts
```diff
@@ -85,7 +85,7 @@ export function assertDepositTermsDeviceCompatible(terms: DepositTerms): void {
 
   // Raw u64 validity first, then the semantic floor — same ordering as the
   // per-vault loop. The intent parser rejects prepegin_max_fee == 0
-  // (vault_tlv.c:152).
+  // (vault_tlv.c:170).
   requireU64("prepeginMaxFee", terms.prepeginMaxFee);
```
gbarkhatov, line 90: **Merge-blocker under CLAUDE.md §7** ("the envelope gate must reject out-of-range terms with the seam's typed error BEFORE any device I/O"). Pre-existing, but this PR refreshes the very comment that names the line — so the citation now asserts a conformance the code does not have.

The comment cites `vault_tlv.c:170`. That firmware line is:

```c
if (out->prepegin_max_fee == 0 || out->prepegin_max_fee > PREPEGIN_MAX_FEE_LIMIT)
    return VAULT_TLV_ERR_VALIDATION;
```

with `PREPEGIN_MAX_FEE_LIMIT ((uint64_t) 100000000u)` at `vault_constants.h:98`. The host mirrors only the `== 0` half — `requireU64` plus `< 1n`. Any `prepeginMaxFee > 1e8` sails through this gate and dies on-device as an opaque `SW_INCORRECT_DATA` with a dead session, which is precisely the failure mode this PR eliminates for `protocolFeeRate` one screen up.

Two lines, on a line you are already editing — add `DEVICE_MAX_PREPEGIN_FEE_SATS = 100_000_000n` to `deviceCaps.ts` beside the fee-rate cap and extend the check here.

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -194,12 +195,13 @@ export function classifyRefundPsbt(psbtHex: string): RefundPsbtClassification |
...
+ * (`:1075`), destination ownership (the device derives, BIP-86-tweaks
+ * and compares — `:1025-1069`; host-side the leaf key is the derivation
  * anchor, so the output must pay ITS BIP-86 address), the no-intent CSV floor
- * (`:898-903`, `vault_constants.h:103`), and the BIP-68 sequence pins
- * (`:956-979` — present, flags clear, low bits encoding exactly the CSV).
+ * (`:912`, `vault_constants.h:103`), and the BIP-68 sequence pins
+ * (`:972-991` — present, flags clear, equal to the CSV; compared unmasked
+ * because the leaf parser already bounds the CSV to the block-count field).
  */
 export function assertRefundPsbtSignable(refund: RefundPsbtClassification): void {
```
gbarkhatov, line 214: The rewritten doc says this mirrors "every term the device validates that needs NO device data", then enumerates five. The device validates a sixth that needs no device data: `PSBT_IN_SIGHASH_TYPE`, at `fw:sign_psbt_validate.c:833-846` — present-and-not-4-bytes rejects, and `res == 4 && sighash_type != SIGHASH_DEFAULT` rejects. Note the refund path is *stricter* than the PoP path at `:584`, which also permits `SIGHASH_ALL`; here explicit `SIGHASH_ALL` is deliberately rejected.

Concrete failure: a refund PSBT from any non-SDK source carrying `sighashType = 1` passes every host gate, reaches the device, and dies with an opaque `SW_INCORRECT_DATA` plus `signStaged`'s pessimistic mirror reset — the exact outcome this function exists to pre-empt. Nothing breaks today because `buildRefundPsbt` never sets the field.

Either add `sighashType` to `RefundPsbtClassification` (bitcoinjs exposes `input.sighashType` on the same `input` object already read at line 186) and reject anything but `undefined`/`SIGHASH_DEFAULT`, or narrow the doc claim to the five terms actually enforced.

packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
```diff
@@ -170,6 +177,8 @@ describe("parseRefundLeafScript (firmware-grammar mirror)", () => {
     ["OP_1NEGATE push", Buffer.from([0x4f])],
     ["negative CScriptNum (sign bit set)", Buffer.from([0x01, 0x80])],
     ["zero CScriptNum", Buffer.from([0x01, 0x00])],
+    ["CSV one above the BIP-68 block-count field (0x10000)", Buffer.from([0x03, 0x00, 0x00, 0x01])],
+    ["CSV 0x7fffffff — the 4-byte maximum the firmware accepted before V-005", Buffer.from([0x04, 0xff, 0xff, 0xff, 0x7f])],
```
gbarkhatov: Audit-finding identifiers do not belong in code, comments, or test descriptions — they are meaningless to anyone without the audit document, and the behaviour is fully describable without them. Two sites here:

- this line: "the 4-byte maximum the firmware accepted before V-005"
- line 303: "the device compares unmasked (V-005)"

Firmware commit `b0c0ac4d` is already cited throughout and is the durable provenance; "…the firmware accepted before the refund-sequence fix" and "…the device compares unmasked" say the same thing and survive the audit document.

packages/babylon-ledger-vault-signer/src/deviceCaps.ts
```diff
@@ -34,23 +34,25 @@ export const DEVICE_MAX_VAULTS_PER_INTENT = 10;
 
 /**
  * Fee-rate bounds (sat/vB): the intent parser rejects rate == 0 AND
- * rate > UINT32_MAX (`vault_tlv.c:73`), so the `>= 1` floor is both a device
- * check and the contract invariant.
+ * rate > 10000 (`vault_tlv.c:75`). The ceiling is vaultd's critical-tx config
+ * default (`btc-vault crates/vaultd/src/config/btc.rs:19` @ d0d139305), not a
```
gbarkhatov, line 38: **UNVERIFIED — I could not resolve this pin, and I am not asserting it is wrong.** The JSDoc cites `btc-vault crates/vaultd/src/config/btc.rs:19 @ d0d139305`. In the local btc-vault clone (`22622fb5`) `d0d139305` is not a valid object and there is no `crates/vaultd/src/config/` directory; the constant with this value lives at `crates/vaultd/src/config.rs:2152` (`DEFAULT_CRITICAL_TX_MAX_FEE_RATE_SAT_VB: u64 = 10_000`). That commit is plausibly newer than my checkout and may well have split `config.rs` into a module — could you confirm the path resolves at `d0d139305`?

The value itself is independently verified and not in question: `vault_tlv.c:75` is `if (rate == 0 || rate > 10000)`, and the firmware's own comment there gives the same provenance ("btc-vault daemon caps ingest at 10,000 sat/vB; Babylon contract caps at 1,000"). The `ProtocolParams.sol:44` half is exact. This is provenance auditability only.

Worth also reconsidering the framing: the binding source of truth for a *device* cap is the firmware line, with vaultd merely being where the number originally came from.

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -8,10 +8,10 @@
  * directly), output 0 by the BIP-86 TWEAKED key — the scriptPubKey's witness
  * program (derived, tweaked, then compared).
  *
- * Firmware citations: `fw:` = LedgerHQ/app-babylon-vault @ `ff1e1ce17`
- * (`fix/feedback`, v0.10.0). Leaf grammar: `fw:sign_psbt_validate_helpers.c:77-148`
- * (`parse_refund_leaf_script`); input entry: `fw:sign_psbt_validate.c:905-950`;
- * output entry: `fw:sign_psbt_validate.c:1005-1057`; both entries are REQUIRED —
+ * Firmware citations: `fw:` = LedgerHQ/app-babylon-vault @ `b0c0ac4d`
+ * (`develop`, 2026-09-04). Leaf grammar: `fw:sign_psbt_validate_helpers.c:77-156`
+ * (`parse_refund_leaf_script`); input entry: `fw:sign_psbt_validate.c:920-948`;
```
gbarkhatov: The refreshed input-entry citation `fw:sign_psbt_validate.c:920-948` stops before the check that gives the entry its meaning. The range runs from `deriv_key[0] = PSBT_IN_TAP_BIP32_DERIVATION` (`:920`) to `check_bip86_path`, but the derive-and-compare — `memcmp(xpub.compressed_pubkey + 1, leaf_key, ...)` at `:961` — is what makes the entry a *proof* that the leaf key is this device's.

The sibling output citation on the next line (`1025-1069`) does reach its `memcmp`, so the two are inconsistent about where the cited block ends. Under §7 these citations are the audit trail, and a reviewer who follows this one never sees the enforcement. `918-965` covers the whole block.

packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
```diff
@@ -290,6 +299,11 @@ describe("assertRefundPsbtSignable", () => {
     expect(() => assertRefundPsbtSignable(classified({ sequence: 144 }))).toThrow(/sequence/);
   });
 
+  it("rejects a sequence with a bit above the block-count field even though the low 16 bits match the CSV", () => {
+    // A masked compare accepted this; the device compares unmasked (V-005).
+    expect(() => assertRefundPsbtSignable(classified({ sequence: 0x00010000 + CSV_2016 }))).toThrow(/sequence/);
```
gbarkhatov, line 313: Anchoring here because this is the assertion that causes it: the two BIP-68 flag tests just above (lines 290-296) can no longer distinguish the branch they exist to pin. Now that `parseRefundLeafScript` bounds `csv <= 0xffff`, any sequence with `0x80000000` or `0x00400000` set is already `!== csv`, so the equality check at `refundPsbt.ts:230` subsumes the flag branch at `:225`. Both tests assert only `.toThrow(/sequence/)` — and the new equality message also contains "sequence", so **deleting `refundPsbt.ts:225-229` entirely would leave both of these green**.

Keep the branch (it gives the better error, and it mirrors the device's own ordering at `fw:sign_psbt_validate.c:983`), but assert the distinct message text — e.g. `.toThrow(/disable\/time flags clear/)` — so the branch is actually pinned.

While here: `04 ff ff ff 7f` at line 181 is now rejected by the `> 0xffff` bound rather than by the sign bit, so no vector pins sign-bit rejection on a *4-byte* push (`04 00 00 00 80`) — the one place JS `<<24` signedness could bite. The branch is still covered by the 1-byte vector at line 178, so this is coverage tidiness, not a hole.

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -120,8 +120,9 @@ export function parseRefundLeafScript(script: Uint8Array): RefundLeafTerms | und
   } else {
     return undefined;
   }
-  // The sign-bit rejections above cap csv at 0x7fffffff — no uint32 wrap.
-  if (csv === 0) return undefined;
+  // Consensus reads only the low 16 bits of a CSV operand, so the firmware
```
gbarkhatov: Minor, but this rationale will mislead whoever next considers relaxing the bound. "Consensus reads only the low 16 bits of a CSV operand" is not quite the rule: BIP-68/112 compares `nSequence` against the operand under `SEQUENCE_LOCKTIME_TYPE_FLAG | SEQUENCE_LOCKTIME_MASK` = `0x0040ffff`, so bit 22 also participates — it selects block-count vs 512-second units. As written the sentence implies a time-based operand is interchangeable with a block-count one, which is exactly the case this bound exists to exclude.

The accurate version is roughly: *the firmware admits only block-count refunds, so the operand must fit the 16-bit block-count field — which also makes the disable and time-based flags unsettable, so callers can compare to nSequence exactly.* (That is close to the firmware's own comment at `helpers.c:136-144`, which carries the same looseness — worth fixing on both sides if you touch the C.)

gbarkhatov (dismissed):
## Combo review — three independent passes, deduped and verified against the firmware

I line-walked `app-babylon-vault` at the pinned `b0c0ac4d8145ed7f243489f65bad28b455c2519f` myself rather than taking any reviewer's word for it.

### The question this PR turns on: does the host gate now accept exactly what the device accepts?

**Yes, in both directions.** Verified statement-for-statement:

- **Not too permissive** (the stated bug): `sign_psbt_validate.c:991` is `if (nsequence != csv_value)` — unmasked, with a firmware comment that literally reads *"Compared unmasked"*. `refundPsbt.ts:230` now mirrors it. The old masked compare admitted `sequence = 0x00010000 | csv`; the new vector at `refundPsbt.test.ts:304` pins exactly that rejection.
- **Not too strict** (the mirror-image risk a fix like this introduces): the new `csv > 0xffff` bound is *the device's own bound*, not a host addition — `sign_psbt_validate_helpers.c:145` is `if (csv > BIP68_SEQUENCE_MASK) return false;`, using the same `vault_constants.h:128` constant. Edges agree on both sides: `csv == 0` rejected (`helpers.c:136`), `0xffff` accepted, `0x10000` rejected, 4-byte non-minimal pushes whose value fits accepted. Nothing can be stranded end to end: the SDK builder already asserts `refundInput.sequence === prePeginParams.timelockRefund` (`ts-sdk .../psbt/refund.ts:227`), and `timelockRefund` is confined to `[72, 4320]` by `envelope.ts:70-75` — far inside the new bound.
- **The PR body's `NonZeroU16` claim is verified, not unverified:** `btc-vault crates/vaultd/src/pegin_broadcast.rs:88` declares `timelock_refund: NonZeroU16`, i.e. exactly `[1, 0xffff]` — the same set as the new host bound.
- **Fee-rate cap:** `10_000n` is exactly `vault_tlv.c:75` (`if (rate == 0 || rate > 10000)`), and the firmware's own comment there gives the same provenance story the JSDoc does. Genuinely unreachable via the contract, and `assertDepositTermsDeviceCompatible` is pure and throws the seam's `DepositTermsRejectedError` before any device I/O — CLAUDE.md §7's ordering rule holds.

Good change, and the citation sweep is unusually careful. Findings below are ordered by severity; only the first is something I would not want to lose track of.

### Findings

1. **(merge-blocker under §7)** `envelope.ts:89-92` mirrors only half of the firmware line it now cites — `prepeginMaxFee` has no upper bound.
2. `refundPsbt.ts:195-207` — the refreshed doc claims it covers *every* zero-device-data term; `PSBT_IN_SIGHASH_TYPE` is not covered.
3. `refundPsbt.test.ts:181,303` — "V-005" audit-finding references in a test description and a comment.
4. `deviceCaps.ts:38` — the btc-vault provenance pin does not resolve (UNVERIFIED, not refuted).
5. `refundPsbt.ts:13` — the input-entry citation stops before the check that gives the entry its meaning.
6. `refundPsbt.test.ts:290-296` — the BIP-68 flag branch is no longer pinned by any test that can distinguish it.
7. `refundPsbt.ts:123-124` — the rationale comment misstates the consensus rule it invokes.

**Body-only (outside a diff hunk, so no inline comment):** `refundPsbt.ts:185` — `sequence: psbt.txInputs[0].sequence ?? 0` turns an *absent* `PSBT_IN_SEQUENCE` into the value `0`, where the device rejects absence outright (`sign_psbt_validate.c:977-978`, `seq_res != 4`). The `0` then trips the CSV floor and reports a misleading timelock error instead of the missing field. Fail-safe and pre-existing — noting it against CLAUDE.md § No Silent Fallbacks, not asking for it here.

### Refuted — findings raised in review that the source killed

Recording these so they are not re-raised:

- **"The conformance claim is unverified and needs a firmware line-walk before merge."** Refuted — I did the line-walk: `helpers.c:136,145` and `sign_psbt_validate.c:983,991`. The host is a statement-for-statement mirror. A Speculos vector would be nice-to-have, not a gate.
- **"Reusing `BIP68_SEQUENCE_MASK` as a numeric upper bound is a footgun; introduce `BIP68_MAX_BLOCK_COUNT`."** Refuted — the firmware bounds it with that exact constant from `vault_constants.h:128`. Parity with the C beats a host-only rename.
- **"Putting the CSV ceiling in the classifier (rather than beside the floor) is an undocumented divergence."** Refuted twice over: the firmware splits it identically (parser rejects `csv > mask`; validator holds the floor at `:912`), and `refundPsbt.ts:85-90` already documents under-classification as the deliberate fail-safe direction.
- **"`protocolFeeRate` is missing its `requireU64` width guard, so an over-width value reaches the TLV encoder as a raw `Error`."** Refuted — `[1n, 10_000n]` is strictly tighter than u64, so no value passes the range check and fails a width check. Unreachable today.
- **"The new envelope test asserts two behaviours in one `it`."** Refuted — it matches the neighbouring convention at `envelope.test.ts:244-249` (`timelockRefund` accepts both bounds and rejects one past each). Consistent with how this repo reads "one behavior per test".
- **"The btc-vault path in `deviceCaps.ts` does not exist."** Overreach, downgraded to UNVERIFIED — see finding 4. The local clone simply does not contain that commit.

### Verified clean

- All ~25 refreshed firmware citations spot-checked accurate: `vault_tlv.c:75,94,148,155,170`, `vault_constants.h:103,106,109,126-128`, `sign_psbt_validate.c:819,850-863,912,972-991,1025-1069`.
- `parseRefundLeafScript` is branch-for-branch equivalent to `parse_refund_leaf_script`, including the `OP_PUSHDATA1` path and non-minimal-push tolerance.
- §7 pre-I/O ordering intact — a rejection still costs zero APDUs.
- No payload bytes logged, no `any`, no dead code; every constant in `refundPsbt.ts` still has a live consumer.
- No masked compare survives anywhere in the signer or the connector.

### Adjacent, not this PR

`babylon-wallet-connector/.../ledger-vault/provider.ts:704,750,765-767` still cites pre-`b0c0ac4d` line numbers for the very functions this PR re-pinned — worth a follow-up so the repo's citations for one firmware file agree on one pin.

packages/babylon-ledger-vault-signer/src/envelope.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/deviceCaps.ts
jrwbabylonlab: not planning — resolves at d0d139305; config.rs split in #2737

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/refundPsbt.test.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/refundPsbt.ts
```diff
@@ -193,15 +201,22 @@ export function classifyRefundPsbt(psbtHex: string): RefundPsbtClassification |
...
+ * device I/O (not even a liveness probe). Mirrors, in order: the sighash pin
+ * (`fw:sign_psbt_validate.c:829-847` — absent or SIGHASH_DEFAULT), the
+ * witnessUtxo requirement (`:850-863`), the output-value cap (`:1075`),
+ * destination ownership (the device derives, BIP-86-tweaks and compares —
+ * `:1025-1069`; host-side the leaf key is the derivation anchor, so the output
+ * must pay ITS BIP-86 address), the no-intent CSV floor (`:912`,
+ * `vault_constants.h:103`), and the BIP-68 sequence pins (`:972-991` —
+ * present, flags clear, equal to the CSV; compared unmasked because the leaf
+ * parser already bounds the CSV to the block-count field).
  */
```
gbarkhatov, line 213: Partly fixed, following up on the earlier sighash comment. The sighash pin is now mirrored, thanks. The doc still says this covers "every term the device validates that needs NO device data", though, and one of those terms is still missing on the host: the NUMS internal-key pin.

Firmware `_refund_verify_taproot_commitment` (`sign_psbt_validate.c`, called just after the input-entry block) needs only the control block. It rejects any internal key that isn't `VAULT_NUMS_XONLY` (`memcmp(internal_key, VAULT_NUMS_XONLY, …)`, constant at `vault_script.c:90`), then checks the merkle root, parity and scriptPubKey.

On the host:
- `RefundPsbtClassification` has no control-block or internal-key field, and `assertRefundPsbtSignable` never looks at either.
- The commitment and parity checks do run, but only later in staging (`assertControlBlockCommitsToLeaf` in `expectedSignatures.ts`), after `gateSignContext`'s liveness probe. So "not even a liveness probe" isn't accurate for those two.
- The NUMS pin is enforced nowhere in production code. `assertControlBlockCommitsToLeaf` only checks that the internal key is an x-only point, and NUMS appears only in test fixtures.

The failure case: a refund PSBT whose commitment is valid but whose internal key isn't NUMS passes every host gate. It then dies on-device with an opaque `SW_INCORRECT_DATA` and `signStaged`'s pessimistic mirror reset, which is exactly what this function exists to prevent. `buildRefundPsbt` always uses NUMS, so only a PSBT from another source can reach this. Low severity.

Either of these would fix it:
- Carry the control-block internal key in the classification, and compare it here against a named `NUMS_XONLY` constant that cites `vault_script.c:90`.
- Or narrow the doc: name the taproot commitment as the one device-data-free term deferred to staging, and say the NUMS pin isn't mirrored.

The sibling JSDoc on `RefundPsbtClassification` ("every transaction field the device validates on the refund path") overclaims the same way. It's outside the diff, so I'm noting it here instead.

gbarkhatov (approved):
Follow-up to my review of `c798888c`. The rebase onto main was clean: the old commit re-applies unchanged as `5df6ffc62`. So this pass covers only the new commit, `0e01224bc`, plus a verdict on each of the 7 earlier comments.

**Headline:** 5 of the 7 are fixed, 1 was declined with a reason, and 1 is partly fixed. The new commit adds no defects. Nothing here blocks merge.

|---|---|---|---|
| 1 | `prepeginMaxFee` had no upper bound; `vault_tlv.c:170` rejects `> PREPEGIN_MAX_FEE_LIMIT` | Fixed. `DEVICE_MAX_PREPEGIN_FEE_SATS = 100_000_000n` matches `vault_constants.h:98`, the band check throws `DepositTermsRejectedError`, and `envelope.test.ts` fails without the ceiling | `envelope.ts:87-95` |
| 2 | The doc said it mirrors every term needing no device data, but the sighash pin was missing | **Partly fixed.** The sighash pin is now mirrored. The NUMS internal-key pin is still missing (inline comment) | `refundPsbt.ts:201-213` |
| 3 | An audit identifier in a test name and a comment | Fixed. None left in the package `src/` | `refundPsbt.test.ts` |
| 4 | The btc-vault pin `config/btc.rs:19 @ d0d139305` could not be resolved | Declined with a reason ("resolves at d0d139305; config.rs split in #2737"). Settled | `deviceCaps.ts:36-41` |
| 5 | The input-entry citation stopped before the `memcmp` at `:961` | Fixed. Now cites `918-965` | `refundPsbt.ts:13` |
| 6 | The flag-branch tests couldn't tell the flag branch from the equality branch, and there was no 4-byte sign-bit vector | Fixed. The tests match `/disable\/time flags clear/`, and the `04 00 00 00 80` vector fails without the sign check | `refundPsbt.test.ts` |
| 7 | The CSV rationale misstated the consensus rule | Fixed | `refundPsbt.ts:127-128` |

**Refuted** (I considered these and ruled them out, so they don't need re-raising):
- *"The host doesn't mirror the device's rejection of a present-but-not-4-byte sighash (`sign_psbt_validate.c:836`)."* That branch is unreachable. `call_get_merkleized_map_value_u32_le` returns -1 for any length other than 4, so the device treats a malformed value as absent. On the host, bip174 throws on a short value (so `classifyRefundPsbt` returns `undefined`, the fail-safe direction), and `augmentPsbtForRefund`'s round-trip re-encodes a long value to exactly 4 bytes.
- *"The explicit `SIGHASH_DEFAULT` test only ever sees an absent field."* bip174 serializes a `0` value, so the test really does exercise an explicit `SIGHASH_DEFAULT`.
- *"The 1e8 fee ceiling could strand a valid SDK deposit."* `prepeginMaxFee` is `sizing.fee`. At the contract's 1000 sat/vB cap, 1e8 sats is 100k vB, which is already the standardness limit.
- *"`requireU64` on `prepeginMaxFee` is now dead code."* It still produces its own u64 error message, and `envelope.test.ts` pins that message.

**Verified clean:**
- The unmasked `sequence !== csv` compare matches the firmware at `:991`, the flag check matches `:983`, and the CSV bound matches `helpers.c:145`.
- Neither new check is stricter than the device: sighash accepts only absent or `0`, and the fee accepts `[1, 1e8]`.
- Both new rejections fire before any device I/O. `assertRefundPsbtSignable` runs before `gateSignContext`, and each has a test that fails if the check is removed.
- The new constants are named and cite the firmware, and no payload bytes are logged.

Checked against `app-babylon-vault` @ `b0c0ac4d` and its `bitcoin_app_base` submodule @ `e400d8d8`.

jonybur (approved):
Re-reviewed at head. The rebase applied cleanly and the follow-up commit `5df6ffc` adds only the `prepeginMaxFee` upper bound, the explicit `SIGHASH_DEFAULT` pin, and the refreshed firmware citations. The three additions all mirror `b0c0ac4d` statement-for-statement: `DEVICE_MAX_PREPEGIN_FEE_SATS = 100_000_000n` matches `vault_constants.h:98`, the unmasked `sequence !== csv` compare matches `sign_psbt_validate.c:991`, and the `csv > BIP68_SEQUENCE_MASK` bound matches `helpers.c:145`. Every earlier round-one comment now has either a fix at head or a documented reason for the deferral, and the two new checks fire before any device I/O with tests that fail on removal. CI green. Nothing new to raise.

Issue #2375, opened 2026-08-31
Ledger: record the reclaim decision (D10) and fix the reserve copy


jrwbabylonlab:
Reclaim of the depositor-claim reserve is blocked for Ledger wallets (`reclaimEligibility.ts:112,228-243`, copy at `copy.ts:1316-1317`, gap noted at `featureFlags.ts:135-137`). That block is **correct and should stay** — but the decision isn't recorded, and the copy it shows points users somewhere unsafe.

## Why the block is right

The device has no validator for the sweep shape: `_validate_display_claim` (`sign_psbt_validate.c:2346` @`ff1e1ce17`) opens with `n_inputs != 1 || n_outputs != 2`, and the dispatcher (`:3649`) routes any 34-byte `<D> OP_CHECKSIG` leaf there — so the N-in/1-out sweep dies on output count before the sequence check is even reached. Not a bug to work around; a missing validator.

## The copy problem

The reserve is spendable **only** through the `<D> OP_CHECKSIG` tapleaf (`reclaim.ts:322-329`). The current wording suggests connecting "the same wallet through another BTC wallet" — which in practice means **restoring the seed into software**. That is not a route to suggest to a hardware-wallet user.

## Tasks

- [x] Record decision **D10**: ship the flip with reclaim blocked on Ledger; the firmware ask (audit §B9, a reserve-sweep validator) is a post-flip item
- [x] Rewrite the blocked-reclaim copy so it does not imply restoring the seed elsewhere
- [x] Make `COPY.deposit.form.transactionReserveTooltip` (`copy.ts:665-672`) a function of an `isLedgerVaultWallet` flag passed from `DepositFeesBreakdown.tsx:105` — one clause, no new component, **no deposit gating**

All user-facing strings live in `services/vault/src/copy.ts` (see CLAUDE.md).

Audit §A6. Related: #2317 (merged).

---
**Status 2026-09-10 — done locally, PR pending.**

- **D10 recorded** in the Ledger knowledge base (`2110-2111-implementation-plan.md` §D10, `03-open-questions.md` product decisions, remaining-work status note). At the live firmware tip `b0c0ac4d` the sweep still has no validator (`_validate_display_claim` `:2369-2370`, dispatcher `:3679`). The B9 validator ask is effectively open with Ledger: the reclaim thread was raised in #ext-ledger-babylon on 2026-09-09.
[...]

Issue #2373, opened 2026-08-31
Ledger: bump Speculos ELF to b0c0ac4d (app 0.10.1) and flip the F1 NoPayout negative pin


jrwbabylonlab:
Firmware moved to **`ff1e1ce17`** (v0.10.0, tip of `fix/feedback`) — 24 commits past the `eacb873b6` pin our tests and notes were written against. Two of our test assertions now pin behaviour Ledger has already fixed.

Corroborated by 6 of the 8 audit lenses.

## What changed on their side

- **F1 fixed** (their PR #8, `b71bcfe`): NoPayout was routed by a PegIn txid the device cannot recompute. Our Speculos **stage 11 still asserts the divergence is present**, so it now pins the wrong behaviour and will fail against the new ELF.
- 0.10.0 also made Assert intent-bound (a BREAKING change on their side) and added a leaf signer-prefix binding.

## What we do

- [x] Bump the Speculos ELF to `ff1e1ce17` (or later) — bumped straight to `develop` @ `b0c0ac4d` (app 0.10.1)
- [x] Flip the F1 negative pin in stage 11 to assert the *fixed* behaviour
- [x] Re-run the presign stages against the new build — full suite 39/39

**Note:** at `base_fee_rate = 1` the fixture's connector value is 509 < 546, so the fixture does **not** exercise the F2 cap (#2372) — refuters explicitly rejected adding a production-valued divergence stage for it. Don't add one while F2 is still open upstream. *(Moot since 0.10.1: the connector cap is gone, see status below.)*

## No production code change

0.10.0 introduces **no** APDU, TLV or status-word change — verified across the whole `eacb873b6..ff1e1ce17` diff. This is test infrastructure plus stale comments only.

Audit §C1. Firmware delta: `todo/ledger/audit-2026-08-30/data/fw-delta-eacb873b6-to-ff1e1ce17.json`.

---
**Status 2026-09-10 — done locally, PR pending.** Target moved past `ff1e1ce17` to the live `develop` tip `b0c0ac4d` (app 0.10.1; tags `*_0.10.1_sdk_v26.6.1`; ELF from CI run 33860782380, artifact `compiled_app_binaries_testnet`, `nanos2/bin/app.elf`).

[...]

PR #2496, opened 2026-09-10
fix(vault, ledger-vault): pin firmware 0.10.1 e2e and fix Ledger recl…

jrwbabylonlab:
Speculos e2e (#2373): pins the suite to firmware `develop` @ `b0c0ac4d` (app 0.10.1, asserted at stage 1), flips stage 11 from "production NoPayout rejected" to "signs and far-side-verifies per challenger", deletes the firmware-shaped fixture, and re-pins every firmware citation. Live run: 40/41 unchanged (only the old pin failed), 39/39 after.

Reclaim copy (#2375): records decision D10 (ship with reclaim blocked on Ledger; validator is a post-flip firmware ask). The blocked-reclaim copy no longer points at "another wallet", and the reserve tooltip takes an `isLedgerVaultWallet` flag threaded from `SimpleDeposit` with the reclaim row's predicate. Display-only, no gating.


packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
```diff
@@ -182,7 +190,7 @@ describe.skipIf(SPECULOS_URL === "")("Speculos end-to-end vault signing", () =>
     async () => {
       const app = await getAppAndVersion(sendRaw);
       expect(app.name).toContain("Babylon Vault");
-      expect(app.version).toMatch(/^\d+\.\d+\.\d+$/);
+      expect(app.version).toBe(PINNED_APP_VERSION);
```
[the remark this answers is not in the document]
jrwbabylonlab: not planning — device and Speculos expose only name/version

jeremy-babylonlabs (dismissed):

jonybur (dismissed):
Firmware pin bump, NoPayout production shape signing, and the reclaim-blocking UX all trace cleanly. The Ledger reserve tooltip threads the new flag through SimpleDeposit → DepositForm → DepositFeesBreakdown, and every call site of the newly parameterised copy function passes the boolean. The dropped dual-shape mechanism removes the workaround the firmware fix made unnecessary. CI green.

services/vault/src/components/simple/__tests__/DepositFeesBreakdown.test.tsx
```diff
@@ -141,3 +144,48 @@ describe("DepositFeesBreakdown commission disclosure", () => {
     expect(screen.getByText(/0\.00000006/)).toBeInTheDocument();
   });
 });
+
+describe("DepositFeesBreakdown reserve tooltip", () => {
+  const reserveTooltipCopy = COPY.deposit.form.transactionReserveTooltip;
```
gbarkhatov: **MERGE BLOCKER — these two tests compare production copy against itself, so no copy regression can fail them.**

`reserveTooltipCopy` is bound to `COPY.deposit.form.transactionReserveTooltip` — the same function the component renders. Lines 169–170 and 188–189 then assert the rendered `data-tooltip-content` against that same function. Invert the ternary at `copy.ts:702-705` and a Ledger depositor is shown *"you can reclaim it once your BTCVault has settled"* while both tests stay green, because the expectation and the subject move together.

That is precisely the bug this PR exists to fix, and these are its only automated guard on the user-visible half. The `.not.toBe` guard at line 187 proves the two branches *differ*, never which is which. Per the test philosophy's "prefer explicit values over magic constants imported from elsewhere", pinning literal substrings would close it — `expect.stringContaining("not supported yet")` for the Ledger render, `"you can reclaim it"` for the other.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/depositorGraphFixture.ts
```diff
@@ -81,20 +74,21 @@ const COUNCIL_QUORUM = 1;
...
 
-/** ChallengeAssert connector value: device requires ≤ DUST (`sign_psbt_validate.c:2149-2167`). */
+/** Arbitrary — since 0.10.1 the device leaves inputs 1-2 unbounded (`sign_psbt_validate.c:2247-2274`). */
 const CHALLENGE_ASSERT_CONNECTOR_VALUE_SATS = 546;
 
 /** Fixture `timelock_challenge_assert` (nopayout.rs:157-176 CAX/CAY sequences); device reads no NoPayout sequence. */
 const CHALLENGE_ASSERT_TIMELOCK = 72;
 
-/** Fixture NoPayout fee; the device validates the output's spk, never its value (`sign_psbt_validate.c:2169-2188`). */
+/** ≤ base_fee_rate × MAX_NOPAYOUT_VSIZE (450) since 0.10.1 (`sign_psbt_validate.c:90`, `:2317-2322`);
```
gbarkhatov: **MERGE BLOCKER — the version attribution is off by a release, in the direction that matters.**

This comment and the one at line 84 both say "since 0.10.1" for the connector-cap removal and the `MAX_NOPAYOUT_VSIZE (450)` fee bound. I checked the firmware repo: both landed in `7c2f184` ("D1 bound NoPayout fee instead of capping connector inputs"), and the `Makefile` at `7c2f184` reads `APPVERSION_P = 0` — **version 0.10.0**. The bump to 0.10.1 came later at `2827377` ("fix fuzz assembly"), whose parent is still 0.10.0.

So a future reader deciding whether an older ELF is safe will wrongly reject 0.10.0 builds that already have the behaviour. §7 makes citation accuracy binding on these files, which is why I'm flagging a comment rather than letting it ride. `since 7c2f184 (0.10.0)` would be exact and wouldn't rot.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/depositorGraphFixture.ts
```diff
@@ -81,20 +74,21 @@ const COUNCIL_QUORUM = 1;
...
+ * (`sign_psbt_validate.c:1725-1755` payout, `:2214-2225` nopayout).
  */
 const ASSERT0_VALUE_SATS = 546 + 386;
 
-/** Non-VP payout Out1 must be exactly DUST (`sign_psbt_validate.c:1898-1904`). */
+/** Non-VP payout Out1 must be exactly DUST (`sign_psbt_validate.c:1997-1999`). */
 const PAYOUT_ANCHOR_VALUE_SATS = 546;
 
-/** ChallengeAssert connector value: device requires ≤ DUST (`sign_psbt_validate.c:2149-2167`). */
+/** Arbitrary — since 0.10.1 the device leaves inputs 1-2 unbounded (`sign_psbt_validate.c:2247-2274`). */
```
gbarkhatov: Separate from the version wording on line 90: keeping `CHALLENGE_ASSERT_CONNECTOR_VALUE_SATS` at `546` means stage 11 doesn't pin the change the ELF was bumped for.

The firmware's own rationale at the cited `:2247-2274` — which I read at the pin — states btc-vault funds each ChallengeAssert:0 at **509 sat at 1 sat/vB** (688 at 2), and that the formula "never yields exactly 546, so the cap admitted only a value the protocol cannot produce." 546 is therefore the single value that signed under the old `≤ DUST` cap *and* signs now — so a firmware regression reinstating the cap would leave this stage green.

Setting it to 509 keeps every other bound satisfied (fee 400 ≤ 1×450; out0 = 932+509+509−400 = 1550 ≥ 546) and makes the suite fail loudly if the cap ever returns. All three review lanes converged on this independently.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/depositorGraphFixture.ts
```diff
@@ -223,7 +215,7 @@ export async function buildDepositorGraphFixture(peginTxHex: string): Promise<De
   payoutTx.addInput(assertTx.getHash(), 0, PAYOUT_TIMELOCK);
   // Depositor claimer: Out0 = V + assert0 − fee − DUST to BIP-86(D), Out1 =
   // DUST CPFP anchor to BIP-86(D) — both script-verified on-device
```
gbarkhatov, line 218: The re-pinned citation stops just before the check it claims. The comment says Out0 and Out1 are "both script-verified on-device (`sign_psbt_validate.c:1910-1999`)", but at the pin Out0's BIP-86 reconstruction is at 1916–1930 while **Out1's** — `_bip86_p2tr_spk` + `memcmp` → `SW_BAD_CPFP_ANCHOR` — is at **2003–2014**, outside the range. The range ends at 1999, inside Out1's *value* check ("Non-VP Out1 must be exactly DUST").

The prior citation was two ranges and covered both halves; collapsing to one dropped the half that supports the word "both". `:1910-2014`, or `:1910-1953, 2003-2014`, would restore it.

Worth noting one review lane cleared this exact range as correct while sweeping ~30 citations — I read 1985–2020 myself and it doesn't cover the claim.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
```diff
@@ -97,11 +97,15 @@ async function loadSdkVerifier() {
 }
 type SdkVerifier = Awaited<ReturnType<typeof loadSdkVerifier>>;
 
+/** App version of the pinned ELF (`b0c0ac4d`); stage 1 asserts it so a run
```
gbarkhatov: `PINNED_APP_VERSION` can't do what the comment claims. It says stage 1 asserts the version "so a run against another build cannot pass as this pin" — but `git log 2827377..b0c0ac4d` is `94ea0eb`, `9227ffb`, `143c081`, `b0c0ac4`, so `2827377` plus four descendants all report `0.10.1`. Five commits satisfy this assertion.

And `sign_psbt_validate.c` moved across that window — this PR's own `:539-543` → `:741-744` re-pin is the evidence. Concrete failure: an operator builds from `2827377`, passes stage 1, then hits stage 11 or 13 failing on line-shifted behaviour and reads it as a protocol regression rather than a wrong-ELF error.

`GET_APP_AND_VERSION` exposes nothing commit-specific, so the honest fix is to soften the comment to "a differently-versioned build cannot pass" and let the CI-artifact provenance carry the real pin.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
```diff
@@ -22,13 +22,13 @@
  *       LAST — it re-derives, which resets the vault session.
  *
  * Stage 5 MUST precede stage 6: a successful Pre-PegIn sign arms the one-shot
- * cap and the next one answers SW_CAP_EXCEEDED (`sign_psbt_validate.c:539-543`).
+ * cap and the next one answers SW_CAP_EXCEEDED (`sign_psbt_validate.c:741-744`).
  *
  * Requires a running container with the vault app (nanosp, testnet build,
```
gbarkhatov, line 27: The ELF's provenance is a bare CI run id that expires. "CI run 33860782380, `compiled_app_binaries_testnet` → `nanos2/bin/app.elf`" is the sole record of which binary the claimed 39/39 run used; the binary isn't committed, and this carries no repo qualifier or URL.

GitHub Actions artifacts expire — 90 days by default — so within a quarter nobody can re-obtain the artifact that every line citation in this file is pinned to. Adding the full run URL and workflow name, or making the reproducible build command the primary provenance, would fix it.

This matters more given my comment on line 100: with five commits reporting `0.10.1`, the run id is currently the *only* real pin.

services/vault/src/copy.ts
```diff
@@ -692,13 +696,14 @@ export const COPY = {
...
       transactionReserveLabel: "Depositor claim output",
-      // Describes the real mechanism, and deliberately shares wording with
-      // COPY.reclaim.review.description so the promise made at deposit time
-      // and the action offered after settlement read as the same thing. Until
-      // the reclaim flow shipped this said the reserve "is returned to you if
-      // unused", which nothing in the app could actually do.
-      transactionReserveTooltip:
-        "A small portion of your deposit is reserved in a dedicated output to fund a future protocol claim transaction. If it goes unused, you can reclaim it from your BTCVault once the vault has settled.",
+      // Promise only what the app can do: the Ledger vault app cannot sign the
+      // reclaim sweep (decision D10, models/reclaimEligibility.ts).
```
gbarkhatov: "decision D10" is cited in seven new comments and defined nowhere in the repo. I grepped `services/vault/src`, `packages/babylon-ledger-vault-signer/src`, `docs/` and `markdown/`: `D10` appears in exactly those seven comments and in **zero** markdown files. Four of them send the reader to `models/reclaimEligibility.ts` for the definition, and that file says only "(decision D10)" — the reference is circular.

It also collides. The pinned firmware uses its own D-numbering in the same subject area: `7c2f184` is D1, and the block cited at `:2272` says "See D1/D2/D13 in tmp/hld-discrepancies-2026-09-01.md". So the identifier now means two different things across the two halves of one PR. The description says this PR "records decision D10", but nothing in the diff records it — stating it in one sentence at `reclaimEligibility.ts` and pointing the other six there, or just using `#2375`, would.

services/vault/src/components/simple/__tests__/DepositFeesBreakdown.test.tsx
```diff
@@ -141,3 +144,48 @@ describe("DepositFeesBreakdown commission disclosure", () => {
...
+
+  function renderedTooltips(): string[] {
+    return Array.from(document.querySelectorAll("[data-tooltip-content]")).map(
```
gbarkhatov: The tooltip lookup is page-wide rather than scoped to the reserve row. `renderedTooltips()` collects `[data-tooltip-content]` across the whole breakdown — the protocol fee, commission and network fee rows all emit one — and both tests assert array membership.

So a refactor that hands the reserve copy to the protocol-fee row and vice versa leaves both tests green while the depositor reads the Ledger caveat next to the wrong number. Scoping the query to the row carrying `FORM_COPY.transactionReserveLabel` would pin placement as well as presence. Independent of, and additive to, the copy-against-itself problem on line 149.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -39,9 +39,8 @@ const ALWAYS_DISABLED_WALLETS: string[] = [
 ];
 
 /**
```
gbarkhatov, line 44: The rewritten JSDoc names two files, and neither imports the symbol. It claims "the reclaim gate (`models/reclaimEligibility`) and the deposit fee breakdown both key on it" — but the only importers of `LEDGER_VAULT_WALLET_ID` are `hooks/deposit/useReclaimRowAction.ts:29` and `components/simple/SimpleDeposit.tsx:17`.

`reclaimEligibility.ts` receives `isLedgerWallet` as a plain boolean and never sees the id; `DepositFeesBreakdown.tsx` likewise. The previous comment was vaguer but true — this rewrite trades vagueness for two pointers that don't resolve. Naming the two real call sites would keep the improvement without the inaccuracy.

services/vault/src/components/simple/SimpleDeposit.tsx
```diff
@@ -497,6 +498,11 @@ function SimpleDepositContent({
                 }}
                 walletState={{
                   isWalletConnected,
+                  // Same predicate as the reclaim row (useReclaimRowAction); the
```
gbarkhatov: This reintroduces the drift class the PR is closing. `btcConnector?.connectedWallet?.id === LEDGER_VAULT_WALLET_ID` is now written out both here and at `hooks/deposit/useReclaimRowAction.ts:87`, and the added comment ("Same predicate as the reclaim row") acknowledges the coupling rather than removing it.

They're character-identical today, so this isn't a live defect. But if the reclaim gate later widens to another connector id and only the hook is updated, the deposit-time tooltip keeps promising a reclaim the row refuses — the exact contradiction this PR exists to fix. Exporting one `isLedgerVaultConnector(btcConnector)` next to `LEDGER_VAULT_WALLET_ID` and calling it from both sites would make that impossible.

services/vault/src/copy.ts
```diff
@@ -1389,10 +1394,10 @@ export const COPY = {
     blocked: {
       protocolPaused:
         "Reclaim is paused while the protocol is under maintenance. Your reserve is safe and will remain reclaimable.",
-      // The Ledger vault app's firmware cannot sign this transaction shape.
-      // See models/reclaimEligibility.ts for the firmware reference.
+      // Ledger cannot sign this shape (D10, models/reclaimEligibility.ts). Never
+      // point at "another wallet": for a hardware user that means restoring the seed.
       ledgerUnsupported:
```
gbarkhatov, line 1421: Copy nit rather than a blocker, but worth a look: this pairs a temporariness claim with a permanence claim and offers no next step. *"Reclaiming with a Ledger device is not supported yet. Your reserve is safe and stays locked to the wallet that created the deposit."*

One entry above, `protocolPaused` solves the same problem with "…is safe and **will remain reclaimable**." Dropping the "another wallet" pointer was right — `ledger_btc` and `ledger_btc_v2` are in `ALWAYS_DISABLED_WALLETS`, so it really did mean restoring the seed, and the comment above says so deliberately. But nothing replaced it, so "stays locked" reads as indefinite lock-up to a depositor who has just been told support is coming. Mirroring the sibling — "will remain reclaimable once support ships" — costs nothing.

Smaller, same area: "once your BTCVault has settled" at `:705` names a state no user-visible label uses; the status label is "Redeemed".

gbarkhatov (dismissed):
Reviewed at `22a7e5c` via three independent passes (7 agents), deduped, with everything below re-checked against the source and against the firmware repo at the pin.

**First, the §7 question, because it governs how this PR must be reviewed: the "re-derivation replaces the author line-walk" clause does not attach here.** This PR touches no file under `src/vendor/ledger-bitcoin/__tests__/vectors/`, and no generator, `scripts/requirements.txt`, or fixtures pin. The oracle stays `ledger-bitcoin==0.4.0` and the vectors' fixtures pin stays `app-babylon-vault @ 8f99b8b`; `git diff --stat 8f99b8b b0c0ac4d -- tests/vectors` is 11 files, 144 insertions, **0 deletions**, all under a new `generated-speculos/` subtree. No regenerated byte moves, so **no reviewer regeneration is owed**. The ELF pin bumped here governs only the live Speculos run.

Also confirmed, since it's the other thing §7-adjacent changes tend to get wrong: **the reclaim change is display-only and the gate is unchanged.** `reclaimEligibility.ts` and its test change are comment-only in this diff — the `isLedgerWallet` branch, its ordering against `isReclaimInFlight`/`payoutSettled`, and the Bitcoin-state gate (`peginTxid:0` spent, ≥6 confs via `isPayoutSettled`) are byte-identical. The gate remains Bitcoin state, never the Ethereum status.

Two blocking comments below, both about *proof* rather than shipped behaviour. Approving and leaving the call to you.

- *"`sign_psbt_validate.c:1910-1999` is a correct citation."* One lane cleared it in a sweep of ~30 citations. I read firmware lines 1985–2020 myself: Out1's script check is at 2003–2014, outside the range. See my comment on the fixture. This is the correlated-agreement failure worth naming — a sweep confirming a range *exists* is not a check that it *covers the claim*.
- *"The stage-11 flip lost coverage."* It didn't — the firmware now routes NoPayout by leaf challenger key rather than a PegIn-txid prevout, so the production shape is the one that signs, and the SDK verification now runs over a broadcastable PSBT instead of a synthetic one. Deleting `firmwareShapedPsbtHex` was required by ZERO DEAD CODE.
- *"Removing 'connect through another BTC wallet' strips a real remedy."* It doesn't — both other Ledger adapters are in `ALWAYS_DISABLED_WALLETS`, so the app cannot offer that path. Removing it was correct.
- *"`sign_psbt_validate.c` lives under `src/handler/`."* At the pin it is `src/sign_psbt_validate.c`.

**Verified clean:** ZERO DEAD CODE holds — no surviving reference anywhere to `firmwareShapedPsbtHex`, `signAndVerifyFirmwareShapedNoPayout` or "firmware-shaped", and every symbol the deletions could have orphaned still has live uses. The remaining ~28 firmware citations are exact at `b0c0ac4d`, checked individually rather than swept — including `:90` (`MAX_NOPAYOUT_VSIZE 450u`), `:741-744`, `:2247-2274`, `:2313-2322` and `display.c:364-366`. Fixture arithmetic satisfies every device bound at the pin (Assert:0 932 ∈ [546, 1046]; NoPayout fee 400 ≤ 1×450; out0 1624 ≥ 546). No `data-testid` was moved, renamed or dropped. Prop wiring is complete — `DepositWalletState` has three construction sites and all three are updated.

One caveat on CI: the Speculos suite is env-gated and skipped, so the green `verify` run does **not** corroborate the "39/39" claim in the description. That rests entirely on the author's live run — which is what my comments on the version pin and the artifact provenance are about making auditable.

services/vault/src/components/simple/__tests__/DepositFeesBreakdown.test.tsx
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/depositorGraphFixture.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/depositorGraphFixture.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/depositorGraphFixture.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/speculos.e2e.test.ts
jrwbabylonlab: fixed

services/vault/src/copy.ts

services/vault/src/components/simple/__tests__/DepositFeesBreakdown.test.tsx
jrwbabylonlab: fixed

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jrwbabylonlab: fixed

services/vault/src/components/simple/SimpleDeposit.tsx
jrwbabylonlab: fixed

services/vault/src/copy.ts
jrwbabylonlab: fixed; not planning — settled is the payout-settlement gate, not Redeemed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/depositorGraphFixture.ts
```diff
@@ -81,20 +74,22 @@ const COUNCIL_QUORUM = 1;
...
 
-/** Non-VP payout Out1 must be exactly DUST (`sign_psbt_validate.c:1898-1904`). */
+/** Non-VP payout Out1 must be exactly DUST (`sign_psbt_validate.c:1997-1999`). */
 const PAYOUT_ANCHOR_VALUE_SATS = 546;
 
-/** ChallengeAssert connector value: device requires ≤ DUST (`sign_psbt_validate.c:2149-2167`). */
-const CHALLENGE_ASSERT_CONNECTOR_VALUE_SATS = 546;
+/** btc-vault's ChallengeAssert:0 at rate 1 (`sign_psbt_validate.c:2254`). Since 7c2f184 (0.10.0) inputs 1-2
+ * are unbounded (`:2247-2274`); the old ≤ DUST cap admitted only 546, so 509 pins that cap's removal. */
+const CHALLENGE_ASSERT_CONNECTOR_VALUE_SATS = 509;
```
gbarkhatov: Correcting my own earlier comment. You applied what I suggested, but 509 doesn't pin the cap removal, because the old cap was an upper bound, not an equality. Before `7c2f184`, the only connector check was `if (connector_value > VAULT_DUST_LIMIT) → SW_INCORRECT_DATA` (`7c2f184^:src/sign_psbt_validate.c`, just after the `PSBT_IN_WITNESS_UTXO` read), with `VAULT_DUST_LIMIT = 546`. So it admitted every value in `[0, 546]`, including 509. The comment's premise ("the old ≤ DUST cap admitted only 546") is therefore false. I think the firmware's own prose ("the cap admitted only a value the protocol cannot produce") misled me, but the code is what counts.

Concretely, under the old firmware this fixture still passes every NoPayout check: inputs total 932 + 2×509 = 1950, out0 is 1550, and the fee is 400. A firmware change that reinstated the cap would leave stage 11 green.

Fix: use a value above 546. btc-vault's own value at 2 sat/vB is 688, and the comment at `b0c0ac4d:2254` gives it. With 688, out0 = 932 + 1376 − 400 = 1908 (≥ DUST), and the fee stays 400 ≤ 450. Reword to "the old cap rejected anything above DUST (`7c2f184^`), so 688 pins its removal."

Since the Speculos suite is env-gated and skipped in CI, and this value changed after the recorded 39/39 live run, it would be worth a live re-run against the pinned ELF.

services/vault/src/copy.ts
```diff
@@ -692,13 +696,14 @@ export const COPY = {
...
-      // unused", which nothing in the app could actually do.
-      transactionReserveTooltip:
-        "A small portion of your deposit is reserved in a dedicated output to fund a future protocol claim transaction. If it goes unused, you can reclaim it from your BTCVault once the vault has settled.",
+      // Promise only what the app can do: the Ledger vault app cannot sign the
+      // reclaim sweep (#2375, models/reclaimEligibility.ts).
+      transactionReserveTooltip: (isLedgerVaultWallet: boolean) =>
+        `${RESERVE_PURPOSE} ${
+          isLedgerVaultWallet
+            ? "If it goes unused, it stays locked to the wallet that created the deposit; reclaiming it with a Ledger device is not supported yet."
+            : "If it goes unused, you can reclaim it once your BTCVault has settled."
```
gbarkhatov, line 705: Low-severity copy nit, and the other half of #11. At deposit time, a Ledger user reads "…it stays locked to the wallet that created the deposit; reclaiming it with a Ledger device is not supported yet." On the reclaim row (`:1422`), the same user now reads "…will remain reclaimable once support ships." So the tooltip still pairs a permanent-sounding "stays locked" with "not supported yet", and now contradicts the fixed string. Mirroring `:1422` here would fix it. The test at `DepositFeesBreakdown.test.tsx:198` asserts "not supported yet", so keep that phrase.

gbarkhatov (dismissed):
Follow-up to my review of `22a7e5cf`. The rebase onto main re-applied both earlier commits unchanged (`17e4267d0` → `afc4a2e43`, `22a7e5cfd` → `91d091fae`), so this pass covers only the new commit `3d6d3c45e`, plus a verdict on each of the 11 earlier comments. The delta touches no generated vector, generator or oracle pin, so no §7 regeneration is owed.

**Headline:** 9 of the 11 are fixed. The other two changed in substance, and one of those is my own error. The connector value my comment #3 prescribed (509) doesn't pin the cap removal (inline). Nothing blocks merge.

|---|---|---|---|
| 1 | The reserve-tooltip tests compared the copy against itself (blocker) | Fixed. They now assert literal substrings unique to each branch, so inverting the ternary fails both | `DepositFeesBreakdown.test.tsx:180-181, 198-199` |
| 2 | "since 0.10.1" misattributed the cap removal and the 450 fee bound (blocker) | Fixed. Now "since 7c2f184 (0.10.0)" | `depositorGraphFixture.ts:84, 91` |
| 3 | A connector value of 546 doesn't pin the cap removal | **Applied as I asked, but my suggested value was wrong** (inline) | `depositorGraphFixture.ts:84-86` |
| 4 | `:1910-1999` missed Out1's script check | Fixed. Now `:1910-2014` | `depositorGraphFixture.ts:219` |
| 5 | The `PINNED_APP_VERSION` comment overclaimed | Fixed | `speculos.e2e.test.ts:103-105` |
| 6 | The ELF provenance was a bare CI run id | Fixed. Full URL, workflow name, expiry date (2026-12-03), and a rebuild fallback | `speculos.e2e.test.ts:28-33` |
| 7 | "Decision D10" was defined nowhere | Fixed in code: stated once against #2375. The PR description still says "records decision D10" | `reclaimEligibility.ts:239-242` |
| 8 | The tooltip lookup was page-wide | Fixed. It now walks up from the reserve label to its own row | `DepositFeesBreakdown.test.tsx:151-166` |
| 9 | The JSDoc named files that don't import the symbol | Fixed | `VaultWalletConnectionProvider.tsx:44-48` |
| 10 | The Ledger predicate was duplicated | Fixed. `isLedgerVaultConnector` is used at both sites, with the expression moved verbatim | `VaultWalletConnectionProvider.tsx:49-53` |
| 11 | `ledgerUnsupported` paired "not supported yet" with "stays locked"; "settled" matches no status label | Partly fixed. `:1422` is fixed; the "settled" half was declined with a reason. The same pairing survives in the sibling tooltip (inline) | `copy.ts:1422`, `copy.ts:704` |

**Refuted:**
- *"The shared predicate changes behavior for non-Ledger wallets."* The expression moved verbatim from both removed inline copies.
- *"Ad-hoc Ledger checks remain."* The other `connectedWallet?.id` reads pass the id to liveness and signing helpers; none compares it to the Ledger id.
- *"Un-exporting `LEDGER_VAULT_WALLET_ID` breaks a test."* No test imports it. The `disabledWallets` tests assert the literal `"ledger_btc_vault"`, which also guards the constant against a typo.

**Verified clean:**
- The delta leaves no dead code.
- Every firmware citation it adds is exact at `b0c0ac4d`: `:90`, `:1910-2014`, `:2247-2274`, `:2254`, `:2313-2322`.
- The `reclaimEligibility.ts` Ledger gate is byte-identical (comment-only change).
- All changed strings are in `copy.ts`, and no `data-testid` moved.

jonybur (dismissed):
Re-reviewed at head. The delta since the prior round is the single follow-up commit `3d6d3c45`, and every point it touches maps to a review thread already open on this PR: the challenge-assert connector value adjustment, the widened firmware citation range, the ELF provenance addendum on the Speculos test, the shared `isLedgerVaultConnector` predicate replacing the two inline checks in `VaultWalletConnectionProvider`, the scoped `reserveRowTooltip` walker with literal-substring assertions in `DepositFeesBreakdown.test.tsx`, and the copy tweaks that swap the audit identifier for `#2375` and drop the "another wallet" phrasing. Reclaim gate logic in `reclaimEligibility.ts` is unchanged - the diff there is comment-only, so the Bitcoin-state gate ordering still holds. Nothing new to raise.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/depositorGraphFixture.ts
jrwbabylonlab: fixed

services/vault/src/copy.ts
jrwbabylonlab: fixed

gbarkhatov (approved):
Follow-up to my approval at `3d6d3c45e1`. The rebase onto `main`, which now includes #2439, #2503 and #2495, left commits 1–2 patch-identical. Only commit 3 was rewritten (`841416d0c`), and within this PR it changes just two things: the ChallengeAssert connector fixture value 509 → 688, with its comment, and the Ledger reserve tooltip copy. Nothing new to raise.

| # | Prior finding | Verdict at `841416d0` |
|---|---|---|
| T1 | Tooltip tests compared copy with itself | **Fixed** at `3d6d3c45`, and it still holds against the new Ledger string |
| T2 | "since 0.10.1" named the wrong release | **Fixed.** Now "since 7c2f184" |
| T3 / T12 | The connector value didn't prove the cap was removed. My 509 suggestion in T3 was wrong, as T12 said: the old cap admitted [0, 546] | **Fixed with 688** (`depositorGraphFixture.ts:84-86`). `7c2f184^:sign_psbt_validate.c:2256` is `connector_value > VAULT_DUST_LIMIT`, so firmware with the cap rejects 688, and the `b0c0ac4d` citation (`:2254-2255`, "688 at 2") is exact |
| T4 | Out1 script check missed | **Fixed** |
| T5 | `PINNED_APP_VERSION` comment overclaimed | **Fixed** |
| T6 | ELF provenance was a bare run id | **Fixed.** Full URL, workflow and expiry |
| T7 | "Decision D10" undefined | **Fixed in code.** The PR description still says "records decision D10" |
| T8 | Tooltip lookup searched the whole page | **Fixed** |
| T9 | JSDoc named files that don't import the symbol | **Fixed** |
| T10 | Ledger predicate duplicated | **Fixed.** One `isLedgerVaultConnector`, an exact match on `"ledger_btc_vault"`, used by both the reclaim row and the deposit tooltip |
| T11 | "not supported yet" vs "stays locked"; "settled" | **"Stays locked" half fixed. The "settled" half was declined, with reason** |
| T13 | Deposit tooltip still said "stays locked" | **Fixed** (`copy.ts:704`) |

Checked, and fine:
- **#2495 compatibility.** The rebase onto #2495's stricter Ledger refund checks doesn't break the Speculos fixtures. Refund sequence equals the leaf CSV (144), no `sighashType` is set, and `protocolFeeRate`/`prepeginMaxFee` are within the new bounds.
- **Firmware pin.** 0.10.1 / `b0c0ac4d` is consistent across `speculos.e2e.test.ts`, `PINNED_APP_VERSION` and the fixture citations. #2495 moved its own pins onto the same commit.
- **Ledger check.** Only a wallet whose id is exactly `ledger_btc_vault` matches. The other Ledger adapters are always-disabled, and signing branches feature-probe instead of keying on the id.
- **CRITICAL PATH §7.** The delta touches no generated vector, generator or `requirements.txt`, so no regeneration is owed.

**What's still unproven:** stage 11 is the only check that fails if the cap comes back, and the Speculos suite runs only with `SPECULOS_URL` set. There is no CI job for it, and the "39/39" in the description predates the 688 change. A live run at this head would close that. `verify` and `visual` were still pending when I posted.

PR #2499, opened 2026-09-10
fix(vault): preserve Ethereum after Bitcoin disconnect

jonybur:
## Outcome

A sustained Bitcoin disconnect clears only Bitcoin state. Ethereum stays connected. This implements the Bitcoin-loss part of #2232 step 3.

Status: all 12 GitHub checks pass for `cd093327`. The fresh Greptile review gives 5/5. Required human reviews remain. The PR targets main; #2439 has merged.

## Change

Use the connector's local cleanup after the existing three-second reconnect delay. Clear the connection guard before cleanup so its disconnect event cannot start another timer. Update the existing Vault lifecycle tests.

## Safety

Both wallets remain required. Bitcoin loss still blocks an incomplete session and clears its required-wallet consent. Account changes and the existing Ethereum-disconnect rules still reset the full session. This change does not start wallet actions or change transaction code.

## Verification

Local integration check: #2499 head `1b9aaa8c` with parent #2439 at `cbf3c6715` passed 19 Vault wallet checks, 358 connector tests, and 23 package boundary checks. These checks used the combined source in a separate worktree. No integration commit was created.

Current commit `cd093327`: [all 12 GitHub checks pass](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34716634111). [Greptile gives 5/5](https://github.com/babylonlabs-io/babylon-toolkit/pull/2499#issuecomment-5617205330), reviewed on 2026-09-13. All seven earlier review threads are resolved. Required human reviews remain.

The local results below were recorded on the earlier revision. Current CI supplies the updated automated evidence.

- Vault: 3,799 tests pass; six existing tests are skipped.
- Wallet connector: 358 tests pass, including local cleanup, shared-session handling, consent, and restore checks.
- Vault production build passes.
- Vault lint passes with no errors and 135 existing warnings.
- The focused checks pass with the fix. Three checks fail with the old implementation.
- Real-wallet browser validation remains with #2355 and #2233. The Vault unit tests verify callback routing; they do not prove a full browser journey.

## Links

Tracks #2232. #2232 stays open. Epic: #2228. Connector dependency: #2439.


jonybur:

Please review the full PR at current commit `1b9aaa8c39a898b4e16c141bc6e90eb1150db800`. Check the diff against its target branch. Report any remaining issues and the confidence score for this commit.

jonybur (commented):
One thing on the new cleanup call, and it has two halves.

**`void btcConnector?.disconnect("local")` fires with the guard already lowered and with nothing attached to the rejection.**

`hasBtcConnectedRef.current = false` runs before the call rather than after it resolves. If `disconnect("local")` rejects, there is no second chance: `scheduleBtcReset` returns early on `!hasBtcConnectedRef.current`, so every later Bitcoin `disconnect` event for the rest of the session is dropped. The connector keeps its Bitcoin wallet, the app keeps rendering Bitcoin as connected, and nothing retries. The failure is silent from the user's side and terminal.

The rejection itself is also unhandled. `void` discards the promise rather than handling it, so a failed local cleanup surfaces only as an unhandled rejection. Setting the ref after the call settles, and logging a failure, would make this recoverable instead — the next disconnect event would still find the guard raised and re-arm the timer.

The optional chain is the same shape one level up. If `useChainConnector("BTC")` returns null when the timer fires, the callback does nothing at all, and the ref has still been cleared. The path it replaces always ran `disconnectAll()`, so this converts a guaranteed teardown into a silent no-op. An explicit branch would keep a missing connector from being indistinguishable from a successful cleanup.

**Checked and clean otherwise.** `runWalletReset` keeps its account-change and Ethereum-disconnect callers, so nothing is orphaned by dropping it from this path, and the `isDisconnectingRef` re-entrancy guard still covers those. The new re-entrancy test pins the right mechanism: the nested `onDisconnect` is dropped by the lowered `hasBtcConnectedRef`, not by `isDisconnectingRef`, which is what the replacement comment describes. And the one-liner left on `BTC_DISCONNECT_DEBOUNCE_MS` still records both why the delay exists and why 1500 ms was not enough, so the constant keeps its justification even though the surrounding rationale block went with it.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -138,31 +121,20 @@ function WalletProviders({ children }: PropsWithChildren) {
...
-      // arrived — treat it as a real disconnect. We deliberately do NOT consult
-      // the connector's `connectedWallet` as a liveness signal: an
-      // extension-initiated disconnect tears down BTCWalletProvider without
-      // calling connector.disconnect(), so `connectedWallet` stays stale-set
-      // and would wrongly suppress a genuine disconnect.
-      void runWalletReset();
+      // Prevent the cleanup event from starting another timer. The local scope
+      // clears Bitcoin selection and storage without disconnecting Ethereum.
+      hasBtcConnectedRef.current = false;
+      void btcConnector?.disconnect("local");
```
gbarkhatov, line 144: The comment compression deleted the rationale this new call actually depends on. All three review lanes flagged it independently, and I verified the mechanism rather than relaying it.

`WalletConnector.disconnect` early-returns at `packages/babylon-wallet-connector/src/core/WalletConnector.ts:62` — `const wallet = this._connectedWallet; … if (!wallet) return;`. So `disconnect("local")` only clears anything *because* an extension-initiated disconnect tears down `BTCWalletProvider` without calling `connector.disconnect()`, leaving `_connectedWallet` stale-set. That is exactly what the deleted paragraph explained.

The irony is that the old note justified why the previous code *ignored* `connectedWallet`; this PR promotes it to the precondition for the new call working at all, and then removes it. Concrete risk: a later reader adds a plausible-looking `if (!btcConnector.connectedWallet) return;` fast-path here, the fix becomes a silent no-op that leaves the persisted Bitcoin session on disk, and no test fails. One sentence at the call site would prevent that.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -138,31 +121,20 @@ function WalletProviders({ children }: PropsWithChildren) {
...
   const scheduleBtcReset = useCallback(() => {
     if (isDisconnectingRef.current) return;
-    // A disconnect before BTC ever finished connecting this session is a
-    // startup/reconnect blip, not a real disconnect — there is no live session
-    // to tear down. Escalating it would wipe the persisted session for BOTH
-    // wallets (including ETH) over a wallet that simply hasn't woken up yet.
     if (!hasBtcConnectedRef.current) return;
     if (pendingBtcResetRef.current !== undefined)
       clearTimeout(pendingBtcResetRef.current);
     pendingBtcResetRef.current = setTimeout(() => {
```
gbarkhatov, line 130: The deferred action no longer re-checks `isDisconnectingRef` at fire time, unlike the `runWalletReset()` it replaced.

`scheduleBtcReset` tests `isDisconnectingRef.current` at *schedule* time on line 126, and the old timer body called `runWalletReset`, whose first statement re-tested it at *fire* time. The new body has no such re-check, so a timer armed at T and a user-initiated disconnect-all at T+2.9s let `disconnect("local")` run concurrently with the in-flight `disconnect("all")`.

Impact today is nil — `useWalletConnect.disconnect()` iterates BTC first, so `_connectedWallet` is already null and the early return at `WalletConnector.ts:62` bails. It becomes live only if a Bitcoin provider that implements `disconnect` is ever enabled, and the only one is the AppKit BTC provider this same file permanently disables. `if (isDisconnectingRef.current) return;` inside the timer body would close it.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -138,31 +121,20 @@ function WalletProviders({ children }: PropsWithChildren) {
...
-      // (fired from the provider's onConnect). If we get here, no reconnect
-      // arrived — treat it as a real disconnect. We deliberately do NOT consult
-      // the connector's `connectedWallet` as a liveness signal: an
-      // extension-initiated disconnect tears down BTCWalletProvider without
-      // calling connector.disconnect(), so `connectedWallet` stays stale-set
-      // and would wrongly suppress a genuine disconnect.
-      void runWalletReset();
+      // Prevent the cleanup event from starting another timer. The local scope
+      // clears Bitcoin selection and storage without disconnecting Ethereum.
+      hasBtcConnectedRef.current = false;
```
gbarkhatov, line 136: Two smaller things on this line and the next.

**The destructive branch is the only one with no log.** `cancelBtcReset` logs "Suppressed transient BTC disconnect…" when a blip is *absorbed*, but the branch that actually tears the Bitcoin session down — clearing `selectedWallets.BTC`, the confirmation receipt and `accountStorage.BTC` — emits nothing. A "my Bitcoin wallet dropped by itself" ticket therefore shows only the suppressions in telemetry and never the teardown, which is the event you'd want to correlate against the debounce window. The `logger.info` with `category: "Wallet connection"` is already imported.

**`hasBtcConnectedRef` is flipped unconditionally, before a call the optional chain can silently skip.** If `btcConnector` were null, `disconnect` vanishes, no event fires, `accountStorage.delete("BTC")` never runs, a reload silently auto-restores a wallet the user lost — and further disconnect events are now ignored until the next `onConnect`. I confirmed this is **not reachable today** (`useChainConnector.ts:8` can return null, but the timer requires a prior successful connect and connectors are never set back to null), so it's a nit rather than a bug — but it's the shape CLAUDE.md's no-silent-fallbacks rule is about. An explicit `if (!btcConnector) { logger.error(...); return; }` would match this file's own style.

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.test.tsx
```diff
@@ -179,7 +179,22 @@ describe("WalletConnectionProvider wallet resets", () => {
     expect(h.disconnectAll).not.toHaveBeenCalled();
     await vi.advanceTimersByTimeAsync(PAST_DEBOUNCE_MS);
 
-    expect(h.disconnectAll).toHaveBeenCalledTimes(1);
+    expect(h.btcConnector.disconnect).toHaveBeenCalledExactlyOnceWith("local");
```
gbarkhatov, line 183: This test no longer tests the dialog, and now duplicates the test 24 lines above it.

I read both tests and the production file. `connectModalVisibleRef` is read at exactly one site — `handleEthDisconnect` — and `scheduleBtcReset` never consults it. This test drives no Ethereum event, so setting `h.visible = true` reaches no code it exercises. Its sequence and both closing assertions are now equivalent to the test at `:149-159`.

It was already a near-duplicate before this PR; the rename removed the last distinguishing assertion, since `disconnectAll` used to be expected *called* here. Per "if you can't name what production behavior a test verifies, the test shouldn't exist", either delete it, or make it earn its name by firing an Ethereum event alongside the Bitcoin one so it actually proves Ethereum survives while the modal is open.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
```diff
@@ -82,29 +83,16 @@ const WALLET_DIALOG_LEFT_INSET_CLASS =
...
-// this bounded amount — never dropped.
-//
-// The window has to outlast a slow Unisat wake-up + auto-reconnect handshake
-// (which is fire-and-forget on reload and can take up to the provider's RPC
-// timeout). 1500ms was shorter than that handshake, so a slow extension wake
-// was being treated as a real disconnect and wiped both wallets. The
-// `hasBtcConnectedRef` gate (below) is the primary guard against startup blips;
-// this debounce + cancelBtcReset cover post-connect wake-up blips.
+// Allow a slow extension to reconnect before clearing its Bitcoin session.
+// A 1500 ms delay was shorter than the UniSat restore handshake.
```
gbarkhatov, line 87: Nit: the surviving comment justifies why *not* 1500, but not why 3000. "A 1500 ms delay was shorter than the UniSat restore handshake" gives only the lower bound.

The deleted text carried the upper bound ("up to the provider's RPC timeout") and the invariant that matters most here — a real disconnect is honoured, just delayed, never dropped. The no-magic-numbers rule asks that the choice of value be obvious, and right now a reader can tell 3000 is more than enough but not why it isn't 10000.

gbarkhatov (dismissed):
Reviewed at `1b9aaa8` via three independent passes, deduped and verified against the source.

**No merge blockers — the change is small and correct.** Switching the debounced Bitcoin teardown from `disconnectAll()` to `btcConnector.disconnect("local")` matches the in-tree connector contract (`useAppKitBtcBridge.ts:133` and `useWalletConnectors.tsx:108` both use `"local"`), the Bitcoin teardown it triggers is complete, and the Ethereum session is genuinely untouched. Everything below is low or nit.

**The question that could have made this a blocker — can a stale Bitcoin account or pubkey survive the disconnect and reach an address derivation, payout script, or signing request? — is clean.** Two independent clears. `BTCWalletProvider.disconnect` wipes `btcWalletProvider`, `address`, `publicKeyNoCoord`, `network`, `locked` and the lock-probe ref at *event* time, three seconds before this timer fires. And every real consumer reads live connector state rather than a captured copy: `usePayoutSigningState.ts:207-212` re-reads the address at click time, hard-fails when absent, matches it against the registered payout scriptPubKey, and runs `verifyBtcWalletLiveness` before signing. Signing is fail-loud rather than stale-valued — `signPsbt`/`getAddress`/`getPublicKeyHex`/`deriveContextHash` all throw "BTC Wallet not connected".

One security-relevant corollary worth recording, since `"local"` scope newly makes a Bitcoin-identity swap possible under a live Ethereum session: the swap cannot bypass re-confirmation. Bitcoin is in `requiredChains`, so the receipt is deleted and `confirmed` goes false. The session-to-chain-identity binding in `SECURITY_MODEL.md:115` is re-established, not inherited.

**Ethereum genuinely survives.** The Bitcoin `disconnect` event does exactly four things — bumps the validation generation, deletes `WALLET_CONFIRMATION_RECEIPT_KEY`, `removeWallet("BTC")`, `accountStorage.delete("BTC")`. It never touches the Ethereum connector, `selectedWallets.ETH`, `accountStorage.ETH`, or the wagmi/AppKit session, and critically does **not** call `reset?.()`, which lives only on the disconnect-all path. No vault code has ever cleared React Query caches on disconnect, so nothing was relying on the old `disconnectAll()` to flush them.

The reverse direction is untouched and still cascades: `handleEthDisconnect` and both `onAddressChange` wirings still route to `runWalletReset` → `disconnectAll()`. That asymmetry looks deliberate and matches the one-direction scope, but worth confirming the Ethereum→Bitcoin case is a later step in the stack rather than an oversight.

**CI:** only GitGuardian and Greptile ran — no build, test, lint or critical-path job. All three real workflows declare `on: pull_request: branches: [main]` and this PR targets `fix/2352-btc-disconnect-keeps-eth`. Same gap as #2492 and #2493, so "3,799 vault tests pass" rests entirely on your local run.

**Refuted:**
- *"The `void`-ed `disconnect("local")` swallows a rejection."* I read `WalletConnector.ts:68-72` — it rethrows only when `scope === "chain"`; otherwise it emits `error`, which reaches this file's `onError` → `logger.error`. `void` discards a promise that cannot reject.
- *"`"chain"` would have been the right scope."* No — `"chain"` can be refused with `SHARED_SESSION_DISCONNECT_REFUSED`, leaving a genuine disconnect unhandled. `"local"` is never refused.
- *"The `[btcConnector]` dep makes `scheduleBtcReset` unstable."* No — `useChainConnector` returns the connector *instance*, and `Chain.context.tsx` preserves instance identity across map rebuilds. The new dep is strictly more stable than the old `[runWalletReset]`.
- *"The new re-entrancy test only exercises its own mock."* No — deleting `hasBtcConnectedRef.current = false` makes it fail, because the mock body has no `await`, so the inner `onDisconnect` runs synchronously and arms a second timer inside the window.

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.test.tsx
```diff
@@ -179,7 +179,22 @@ describe("WalletConnectionProvider wallet resets", () => {
...
+      btc().onDisconnect();
+    });
+
+    act(() => btc().onConnect());
+    act(() => btc().onDisconnect());
+    await vi.advanceTimersByTimeAsync(PAST_DEBOUNCE_MS * 2);
+
+    expect(h.btcConnector.disconnect).toHaveBeenCalledExactlyOnceWith("local");
+    expect(h.disconnectAll).not.toHaveBeenCalled();
   });
```
jeremy-babylonlabs, line 199: The one new failure mode of this design — the guard being lowered at cleanup — is untested. Nothing proves a BTC reconnect after local cleanup re-raises `hasBtcConnectedRef`, so a second sustained disconnect could be dropped forever and no test would notice.

One case covers it: connect -> disconnect -> advance past debounce (expect `disconnect("local")` x1) -> `onConnect()` -> `onDisconnect()` -> advance again -> expect `disconnect("local")` x2.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jeremy-babylonlabs, line 144: For the record, since it came up in review: `WalletConnector.disconnect` only rethrows for scope `"chain"` — the `"local"` path catches, emits `error`, and still nulls and emits `disconnect`. So the `void` here cannot strand the guard. No change needed.

jeremy-babylonlabs (commented):
Approach is right — lowering the guard before `disconnect("local")` so the resulting event cannot re-arm the timer is the correct ordering. Note the PR is currently `CONFLICTING`; it needs a rebase onto `main` before CI can run.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jonybur: Restored the rationale in cd093327. Lines 137-140 now say that an extension-initiated disconnect tears down BTCWalletProvider without calling connector.disconnect(), so connectedWallet stays stale-set. That is why the call still clears the persisted session, and why it is not gated on connectedWallet. I re-checked BTCWalletProvider.disconnect on main: it clears React state and fires onDisconnect, and never calls the connector.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jonybur: Added the fire-time check in cd093327. The timer callback now returns when isDisconnectingRef is set, so a full reset that starts inside the window runs alone. The check sits right after the pending-timer ref is cleared, so a later reconnect does not log a suppression for a timer that already fired. A test for this guard is in cd093327: it holds disconnectAll pending across the timer and asserts no disconnect("local"). I did not add a timer clear to runWalletReset: a full reset leaves _connectedWallet null, so a timer that fires after it hits the early return in WalletConnector.disconnect and clears nothing. It would still emit the new "Clearing the Bitcoin session" log line once. Say so if you want the timer cleared in runWalletReset to drop that line.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jonybur: Added the log line in cd093327. The callback now logs "Clearing the Bitcoin session after a sustained disconnect" under the Wallet connection category before the call. hasBtcConnectedRef stays lowered before the call on purpose: for every BTC wallet this file enables the provider has no disconnect method, so WalletConnector.disconnect emits "disconnect" synchronously and re-enters scheduleBtcReset; the existing cleanup-event test pins that. Lowering it first also holds if a provider with disconnect were ever enabled, since the emit then follows the await. The null-connector branch cannot occur. Chain.context never writes null over a built connector, and the timer only arms after a connect, which needs the connector. An error branch for it would add code the fix does not need.

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.test.tsx
jonybur: The test now drives an Ethereum disconnect while the dialog is open in cd093327. It calls eth.onDisconnect right after the Bitcoin disconnect and keeps the same assertions. That exercises the modal suppression in handleEthDisconnect, which is the one branch h.visible controls, and proves Ethereum survives while the modal is open.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jonybur: No change. The two surviving lines give the reason the delay exists and the lower bound that moved it from 1500 ms to 3000 ms. The code now shows the rest: cancelBtcReset clears the timer on reconnect, and a real disconnect only waits for the timer. It is never dropped. If you want the upper bound recorded, one clause on line 87 covers it: the window must outlast the auto-reconnect handshake, which can run to the provider's RPC timeout. Say so and I will add it.

services/vault/src/context/wallet/__tests__/VaultWalletConnectionProvider.test.tsx
jonybur: Added the test in cd093327, as described: connect, disconnect, advance past the debounce, expect disconnect("local") once, then onConnect, onDisconnect, advance again, expect disconnect("local") twice. It fails if a reconnect after a local cleanup leaves hasBtcConnectedRef lowered.

services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx
jonybur: Agreed. WalletConnector.disconnect rethrows only for scope "chain". For "local" it emits "error" and still finishes the local teardown, and the BTC AppKit provider's "local" branch only calls clearSession. So the void call cannot strand the guard. No change. Resolving.

jonybur:
Review fixes are pushed as `cd093327`, on top of a merge of main now that #2439 has landed; the branch no longer conflicts.

Five threads produced changes, two are answered. The deferred Bitcoin cleanup now re-checks the full-reset flag when it fires, logs what it does, and carries the rationale it depends on. Two tests were added: the reconnect-after-cleanup case jeremy asked for, and one for the fire-time guard; both fail with their fix reverted.

Local checks at this head: 3,859 Vault tests in 315 files, eslint clean on every edited file.

jonybur:
@greptileai review the full PR at current commit `cd093327f0cf37847640632dac6267712f811090`. Check the diff against its target branch, including the changes since your last review. Verify that the review findings are addressed in the code. Report any remaining issues and the confidence score for this commit.

gbarkhatov (approved):
Follow-up to my review at `1b9aaa8c39`. Main has been merged in since then, and #2439 has landed, so I compared the PR's own net patch at both commits. Nothing new to raise.

| # | Prior finding | Verdict at `cd093327` |
|---|---|---|
| T1 | The comment explaining why `connectedWallet` stays stale-set after an extension disconnect was deleted | **Fixed** (`VaultWalletConnectionProvider.tsx:137-140`) |
| T2 | The timer did not re-check `isDisconnectingRef` when it fires | **Fixed** (`:132-133`). Pinned by the test at `VaultWalletConnectionProvider.test.tsx:201-212`, which fails if `:133` is removed |
| T3 | The branch that clears Bitcoin logged nothing, and the guard is lowered before a call that may do nothing | **Log added** (`:141-143`). The null-connector branch is **declined, with reason**: the timer only arms after a connect, which needs the connector |
| T4 | The dialog test no longer exercised the dialog | **Fixed.** It fires `eth.onDisconnect` (`:179`), so removing the modal check fails it |
| T5 | The comment gives only the lower bound for 3000 ms | **Declined** |

My earlier "no build/test job ran" point no longer applies: `verify` (build, lint, affected tests) passes on this head.

Checked, and fine:
- There is no JS turn between lowering the guard and the local cleanup, so a full reset cannot interleave with it.
- The Bitcoin `disconnect` event clears the stored consent receipt, so an Ethereum-only session is never treated as complete.
- A reconnect that lands in the gap between the connector's `connect` event and `cancelBtcReset` can still be cleared by the timer. The result is fail-safe (the user reconnects), and the window predates this PR.

PR #2500, opened 2026-09-10
test(vault): cover payout addresses on both networks

jonybur:
## Outcome

Status: all 12 GitHub checks pass for `c578deb0`. The fresh Greptile review gives 5/5. Required human reviews remain.

Payout-address checks cover mainnet and signet for Ethereum-only Vault access. This supplies address-regression evidence for #2232 step 4.

## Change

- Check P2PKH, P2SH, P2WPKH, P2WSH, and Taproot with fixed expected addresses and the real network configuration.
- Check uppercase hex, odd-length hex, unsupported scripts, and an invalid Taproot output key. Keep the existing prefix and round-trip checks.
- Load the built wallet entry through Node in tests. Vite's test transform otherwise changes a valid JavaScript label into invalid syntax.

## Safety

The existing converter supplies every result. No production code changes. The expected Segwit values include [BIP350 vectors](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki#test-vectors-for-v0-v16-native-segregated-witness-addresses). The other values are derived from bitcoinjs-lib 6.1.7.

Scope decision in #2228: download-size improvements are optional. Keep the current converter and Taproot validation. This PR does not require a new encoder or a loading refactor. The loading observations below are background evidence, not completion gates.

## Verification

Current commit `c578deb0`: [all 12 GitHub checks pass](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34716737151). [Greptile gives 5/5](https://github.com/babylonlabs-io/babylon-toolkit/pull/2500#issuecomment-5617646646), reviewed on 2026-09-13. All seven earlier review threads are resolved. The current focused address suite passes all 18 tests.

Earlier evidence at `1a272544177934ef84be24b3412b9c80eb6ff5ab`: [Verify PR](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34468982808/job/102844176213) and [Visual regression](https://github.com/babylonlabs-io/babylon-toolkit/actions/runs/34468982859/job/102844176542) pass. Both visual versions contain all 329 Storybook screens and 58 Vault screens. No images are missing or changed. [Greptile gives 5/5](https://github.com/babylonlabs-io/babylon-toolkit/pull/2500#issuecomment-5617646646).

- Full Vault suite: 3,824 tests pass; two tests are skipped.
- Vault lint and production build pass.
- A deliberate mainnet-to-testnet conversion error fails the new mainnet check. The original converter is restored.
- Parsed the built JavaScript import edges and source maps at main `05f60bf9`. The app entry, address decoder, withdrawal address helper, and redeemed-vault hook each have zero Rust-engine source modules in their static import closures. Each still includes 30 Bitcoin library modules and five curve modules. The two HTML modulepreloads also contain no Rust-engine sources.
- The 3,118,058-byte Rust WASM asset remains behind dynamic loading. This is build evidence. It does not prove a browser journey or exclude imports called during rendering.

## Links

Tracks #2232. #2232 stays open. Epic: #2228. Final browser journey: #2233.

### DIFF
--- a/services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
+++ b/services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
@@ -8,8 +8,15 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 vi.unmock("@/config");
 vi.unmock("@/config/network");
 
-// Segwit vectors: https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki
-// The remaining addresses retain the existing bitcoinjs-lib 6.1.7 results.
+// The four segwit v0 addresses are the BIP-173 examples:
+// https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#examples
+// BIP-350's valid table also attests bc1qw508...kv8f3t4 and tb1qrp33...q0sl5k7
+// with their scriptPubKeys, and adds bc1p0xlxvlh...vqzk5jj0:
+// https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki#test-vectors-for-v0-v16-native-segregated-witness-addresses
+// The rest are derived, not BIP vectors. tb1p0xlxvlh...vq47zagq is the same
+// Taproot key under the tb HRP. The four base58 values are base58check of
+// hash160 751e76e8199196d454941c45d1b3a323f1433bd6 under version bytes 0x00,
+// 0x6f, 0x05 and 0xc4. The derived values were recorded from bitcoinjs-lib 6.1.7.
 const P2PKH_SCRIPT = "0x76a914751e76e8199196d454941c45d1b3a323f1433bd688ac";
 const P2SH_SCRIPT = "0xa914751e76e8199196d454941c45d1b3a323f1433bd687";
 const SCRIPT_HEX_PREFIXED = "0x0014751e76e8199196d454941c45d1b3a323f1433bd6";
@@ -30,7 +37,7 @@ beforeEach(async () => {
   configureBabylonConfig({
     btcNetwork: "signet",
     ethChainId: 11155111,
-    ethRpcUrl: process.env.NEXT_PUBLIC_ETH_RPC_URL!,
+    ethRpcUrl: "https://test.example/eth",
   });
   ({ btcAddressToScriptPubKeyHex, scriptPubKeyHexToBtcAddress } = await import(
     "../btc"
@@ -42,51 +49,77 @@ beforeEach(async () => {
 afterEach(() => bitcoin.initEccLib(undefined));
 
 describe("scriptPubKeyHexToBtcAddress", () => {
-  it("decodes supported payout scripts on the configured signet network", () => {
+  it("decodes a P2PKH scriptPubKey to its signet address", () => {
     expect(scriptPubKeyHexToBtcAddress(P2PKH_SCRIPT)).toBe(
       "mrCDrCybB6J1vRfbwM5hemdJz73FwDBC8r",
     );
+  });
+
+  it("decodes a P2SH scriptPubKey to its signet address", () => {
     expect(scriptPubKeyHexToBtcAddress(P2SH_SCRIPT)).toBe(
       "2N3vVYSK5XRgVSGWy21PnsRmBUywSQNdCsf",
     );
+  });
+
+  it("decodes a P2WPKH scriptPubKey to its signet address", () => {
     expect(scriptPubKeyHexToBtcAddress(SCRIPT_HEX_PREFIXED)).toBe(
       EXPECTED_TESTNET_ADDRESS,
     );
+  });
+
+  it("decodes a P2WSH scriptPubKey to its signet address", () => {
     expect(scriptPubKeyHexToBtcAddress(P2WSH_SCRIPT)).toBe(
       "tb1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3q0sl5k7",
     );
+  });
+
+  it("decodes a P2TR scriptPubKey to its signet address", () => {
     expect(scriptPubKeyHexToBtcAddress(P2TR_SCRIPT)).toBe(
       "tb1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vq47zagq",
     );
   });
 
-  it("decodes supported payout scripts on the configured mainnet network", async () => {
-    vi.resetModules();
-    const { configureBabylonConfig } = await import(
-      "../../config/network/runtime"
-    );
-    configureBabylonConfig({
-      btcNetwork: "mainnet",
-      ethChainId: 1,
-      ethRpcUrl: process.env.NEXT_PUBLIC_ETH_RPC_URL!,
-    });
-    const { scriptPubKeyHexToBtcAddress } = await import("../btc");
+  describe("on the configured mainnet network", () => {
+    let decode: typeof import("../btc").scriptPubKeyHexToBtcAddress;
 
-    expect(scriptPubKeyHexToBtcAddress(P2PKH_SCRIPT)).toBe(
-      "1BgGZ9tcN4rm9KBzDn7KprQz87SZ26SAMH",
-    );
-    expect(scriptPubKeyHexToBtcAddress(P2SH_SCRIPT)).toBe(
-      "3CNHUhP3uyB9EUtRLsmvFUmvGdjGdkTxJw",
-    );
-    expect(scriptPubKeyHexToBtcAddress(SCRIPT_HEX_PREFIXED)).toBe(
-      "bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4",
-    );
-    expect(scriptPubKeyHexToBtcAddress(P2WSH_SCRIPT)).toBe(
-      "bc1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3qccfmv3",
-    );
-    expect(scriptPubKeyHexToBtcAddress(P2TR_SCRIPT)).toBe(
-      "bc1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vqzk5jj0",
-    );
+    beforeEach(async () => {
+      vi.resetModules();
+      const { configureBabylonConfig } = await import(
+        "../../config/network/runtime"
+      );
+      configureBabylonConfig({
+        btcNetwork: "mainnet",
+        ethChainId: 1,
+        ethRpcUrl: "https://test.example/eth",
+      });
+      ({ scriptPubKeyHexToBtcAddress: decode } = await import("../btc"));
+    });
+
+    it("decodes a P2PKH scriptPubKey to its mainnet address", () => {
+      expect(decode(P2PKH_SCRIPT)).toBe("1BgGZ9tcN4rm9KBzDn7KprQz87SZ26SAMH");
+    });
+
+    it("decodes a P2SH scriptPubKey to its mainnet address", () => {
+      expect(decode(P2SH_SCRIPT)).toBe("3CNHUhP3uyB9EUtRLsmvFUmvGdjGdkTxJw");
+    });
+
+    it("decodes a P2WPKH scriptPubKey to its mainnet address", () => {
+      expect(decode(SCRIPT_HEX_PREFIXED)).toBe(
+        "bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4",
+      );
+    });
+
+    it("decodes a P2WSH scriptPubKey to its mainnet address", () => {
+      expect(decode(P2WSH_SCRIPT)).toBe(
+        "bc1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3qccfmv3",
+      );
+    });
+
+    it("decodes a P2TR scriptPubKey to its mainnet address", () => {
+      expect(decode(P2TR_SCRIPT)).toBe(
+        "bc1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vqzk5jj0",
+      );
+    });
   });
 
   it("accepts unprefixed hex (no leading 0x)", () => {
@@ -113,11 +146,12 @@ describe("scriptPubKeyHexToBtcAddress", () => {
     );
   });
 
-  it("rejects a Taproot script whose output key is not on the curve", () => {
+  it("rejects a Taproot output key that is not a valid x-only point", () => {
+    expect(scriptPubKeyHexToBtcAddress(P2TR_SCRIPT)).toBe(
+      "tb1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vq47zagq",
+    );
     expect(() =>
-      scriptPubKeyHexToBtcAddress(
-        "0x5120ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
-      ),
+      scriptPubKeyHexToBtcAddress(`0x5120${"ff".repeat(32)}`),
     ).toThrow("has no matching Address");
   });
 
--- a/services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
+++ b/services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
@@ -1,38 +1,91 @@
+// @vitest-environment node
+// The curve library needs Node's typed arrays for these non-DOM checks.
+
+import * as ecc from "@bitcoin-js/tiny-secp256k1-asmjs";
 import * as bitcoin from "bitcoinjs-lib";
-import { describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
-import {
-  btcAddressToScriptPubKeyHex,
-  scriptPubKeyHexToBtcAddress,
-} from "../btc";
+vi.unmock("@/config");
+vi.unmock("@/config/network");
 
-/**
- * Build the test fixture via bitcoinjs-lib's own payment helper so the script
- * bytes are guaranteed valid.
- *
- * Test config mocks the BTC network as signet (treated as testnet by btcUtils),
- * so the expected address is bech32-encoded with the `tb1` HRP.
- */
-const PUBKEY_HASH_BYTES = new Uint8Array([
-  0x75, 0x1e, 0x76, 0xe8, 0x19, 0x91, 0x96, 0xd4, 0x54, 0x94, 0x1c, 0x45, 0xd1,
-  0xb3, 0xa3, 0x23, 0xf1, 0x43, 0x3b, 0xd6,
-]);
+// Segwit vectors: https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki
+// The remaining addresses retain the existing bitcoinjs-lib 6.1.7 results.
+const P2PKH_SCRIPT = "0x76a914751e76e8199196d454941c45d1b3a323f1433bd688ac";
+const P2SH_SCRIPT = "0xa914751e76e8199196d454941c45d1b3a323f1433bd687";
+const SCRIPT_HEX_PREFIXED = "0x0014751e76e8199196d454941c45d1b3a323f1433bd6";
+const P2WSH_SCRIPT =
+  "0x00201863143c14c5166804bd19203356da136c985678cd4d27a1b8c6329604903262";
+const P2TR_SCRIPT =
+  "0x512079be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798";
+const EXPECTED_TESTNET_ADDRESS = "tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx";
 
-const { output, address: EXPECTED_TESTNET_ADDRESS } = bitcoin.payments.p2wpkh({
-  hash: Buffer.from(PUBKEY_HASH_BYTES),
-  network: bitcoin.networks.testnet,
+let btcAddressToScriptPubKeyHex: typeof import("../btc").btcAddressToScriptPubKeyHex;
+let scriptPubKeyHexToBtcAddress: typeof import("../btc").scriptPubKeyHexToBtcAddress;
+
+beforeEach(async () => {
+  vi.resetModules();
+  const { configureBabylonConfig } = await import(
+    "../../config/network/runtime"
+  );
+  configureBabylonConfig({
+    btcNetwork: "signet",
+    ethChainId: 11155111,
+    ethRpcUrl: process.env.NEXT_PUBLIC_ETH_RPC_URL!,
+  });
+  ({ btcAddressToScriptPubKeyHex, scriptPubKeyHexToBtcAddress } = await import(
+    "../btc"
+  ));
+  // Match the curve setup in the application's main.tsx.
+  bitcoin.initEccLib(ecc);
 });
 
-if (!output || !EXPECTED_TESTNET_ADDRESS) {
-  throw new Error("Test fixture setup failed: could not derive p2wpkh");
-}
-
-const SCRIPT_HEX_PREFIXED = `0x${Buffer.from(output).toString("hex")}`;
+afterEach(() => bitcoin.initEccLib(undefined));
 
 describe("scriptPubKeyHexToBtcAddress", () => {
-  it("decodes a P2WPKH scriptPubKey hex back to its testnet address", () => {
+  it("decodes supported payout scripts on the configured signet network", () => {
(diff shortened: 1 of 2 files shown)


jonybur (commented):
One finding on the new rejection tests.

**`rejects a Taproot script whose output key is not on the curve` cannot fail for the reason it names.** `address.fromOutputScript` tries each payment type inside its own `try {} catch {}` and, when every one of them fails, throws the single generic `<asm> has no matching Address`. So three different outcomes are indistinguishable at that assertion: the P2TR branch rejecting the key because `isXOnlyPoint` said no, the P2TR branch throwing because no ECC backend was installed, and the script never matching P2TR at all. The test one block above (`0x6a`, no destination) asserts the same string, which is the tell.

That matters here specifically because this file tears the curve backend down between tests. `afterEach` calls `initEccLib(undefined)`, and every test re-imports `../btc` after `vi.resetModules()`. The positive P2TR assertions would catch a lost backend, but this test would keep passing while the guarantee it claims to pin — that an off-curve output key is rejected rather than encoded into an address — went unpinned.

Asserting the distinguishing behaviour would close it: pin that the same 32 bytes with a valid x-coordinate do decode to a `tb1p…` address and that the off-curve variant does not, or reach past `fromOutputScript` and assert on `payments.p2tr({ output })` directly, where the failure is the specific `Invalid pubkey for p2tr`.

**Checked and correct.** Every expected address matches its scriptPubKey. `bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4` / `tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx`, both P2WSH values and both P2TR values are the BIP-173 and BIP-350 vectors verbatim. `1BgGZ9tcN4rm9KBzDn7KprQz87SZ26SAMH` and `mrCDrCybB6J1vRfbwM5hemdJz73FwDBC8r` are the mainnet and testnet base58check encodings of hash160 `751e76e8199196d454941c45d1b3a323f1433bd6`, and the two P2SH values carry the same 20 bytes under versions `0x05` and `0xc4`.

The mainnet block cannot leak into the tests below it: it re-imports after `vi.resetModules()` and binds a local `scriptPubKeyHexToBtcAddress`, while the module-level binding the later tests read is reassigned by `beforeEach` against a fresh signet configuration each time. Signet and testnet share the `tb` HRP, so the `tb1…` expectations are right for the configured network.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
```diff
@@ -1,39 +1,92 @@
+// @vitest-environment node
```
gbarkhatov, line 1: Moving the whole file to the node environment deletes the only coverage of the jsdom Buffer hazard the production code exists to defend against.

`btcUtils.ts:51-63` carries an eight-line comment and a hand-rolled byte loop precisely because, under jsdom, the npm `buffer` polyfill's `Buffer.from(hex, "hex")` *silently returns zero bytes* — so it deliberately reaches for `globalThis.Buffer.alloc`. I confirmed `environment: "jsdom"` is the vault-wide default (`vitest.config.ts:90`) and that `buffer` 6.0.3 is a direct dependency of `services/vault`, so this file was that workaround's only guard.

Under `node`, the imported `Buffer` and `globalThis.Buffer` are the same object. So replacing that loop with a plain `Buffer.from(cleanHex, "hex")` now keeps all eleven tests green while the browser bundle decodes a **wrong payout destination** for a depositor. Only the two P2TR cases actually need node — splitting those into a node-env sibling and leaving the rest on jsdom would keep both properties. (Raised as a merge-blocker by one lane; I'd call it lower than that, since this PR ships no production change — it removes a guard rather than shipping a defect.)

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
```diff
@@ -1,39 +1,92 @@
...
-});
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.unmock("@/config");
+vi.unmock("@/config/network");
 
-if (!output || !EXPECTED_TESTNET_ADDRESS) {
-  throw new Error("Test fixture setup failed: could not derive p2wpkh");
-}
+// Segwit vectors: https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki
```
gbarkhatov: The provenance comment overclaims: only 3 of the 10 expectations are BIP vectors, and the other 7 are recorded outputs of the library under test.

I fetched BIP350 and grepped its tables. Exactly three asserted pairs are verbatim valid vectors: `BC1QW508D6…KV8F3T4`, `tb1qrp33g0q…q0sl5k7`, and `bc1p0xlxvlh…vqzk5jj0`. The other three segwit values are cross-network transpositions appearing in no BIP valid list — and `tb1p0xlxvlh…vq47zagq` occurs in BIP350 **only in the invalid-vector list**, as mixed-case `…vq47Zagq`. BIP350 contains no base58 vectors at all, so the four base58 literals are unattested by the cited source.

The second line is also wrong on its face: the pre-PR file contained one address, and it was *derived* — the four base58 literals are new bitcoinjs output pasted in, not retained baselines.

The values are all correct; only the citation needs fixing. It matters because #2232 will diff a hand-written encoder against exactly these fixtures, and its reviewer will read this comment as "BIP-anchored". Marking the three genuine entries and saying plainly that the rest are derived would set the right expectation.

services/vault/vitest.config.ts
```diff
@@ -117,6 +117,8 @@ export default defineConfig({
     },
     server: {
       deps: {
+        // Use Node for the built wallet entry. Vite corrupts a label in this bundle.
+        external: [/\/babylon-wallet-connector\/dist\/index\.es\.js$/],
```
gbarkhatov, line 130: This works, but only because the `inline` entry beside it is inert under pnpm's symlinked layout — which makes it fragile in a non-obvious way.

I read `_shouldExternalize` in vitest 4.0.9: `inline` is tested at line 8909 and `external` at 8911, so inline wins whenever both match. This regex survives only because the adjacent string entry `"@babylonlabs-io/wallet-connector"` never matches — `matchPattern` requires `id.includes("/node_modules/@babylonlabs-io/wallet-connector")`, and that path is a symlink to `packages/babylon-wallet-connector`, which Vite realpaths away. Convert that string to a regex, or set `preserveSymlinks`, and this fix silently reverts with no failing test.

Two smaller things: it sits above `projects`, so it flips module loading for *every* vault unit test in order to satisfy one file; and the comment names no label, no error and no issue, so nobody can later tell whether it's still needed. A scoped `vi.mock("@babylonlabs-io/wallet-connector", …)` in the test file would avoid the global change — the adapter reads only `Network.MAINNET`/`SIGNET`.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
```diff
@@ -42,6 +95,32 @@ describe("scriptPubKeyHexToBtcAddress", () => {
...
+    );
+  });
+
+  it("rejects a script without a destination address", () => {
+    expect(() => scriptPubKeyHexToBtcAddress("0x6a")).toThrow(
+      "has no matching Address",
+    );
+  });
+
+  it("rejects a Taproot script whose output key is not on the curve", () => {
```
gbarkhatov: This test cannot observe the curve check it's named for. I read `fromOutputScript` in bitcoinjs-lib 6.1.7: it wraps all six payment attempts in bare `catch (e) {}` and throws one generic `has no matching Address`.

So the assertion is green if the curve check runs, if `getEccLib()` throws because ecc was never initialised, **and** if the check were deleted from `p2tr` altogether — the `_toFutureSegwitAddress` fallthrough rejects witness v1 regardless. It asserts the identical string as the test directly above it, making the two mutually indistinguishable, and that string belongs to a third-party library rather than to `src/`.

Either rename it to what it actually verifies, or assert the specific cause by calling `bitcoin.payments.p2tr({ output, network })` directly and expecting `Invalid pubkey for p2tr`.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
```diff
@@ -1,39 +1,92 @@
...
+  ));
+  // Match the curve setup in the application's main.tsx.
+  bitcoin.initEccLib(ecc);
+});
+
+afterEach(() => bitcoin.initEccLib(undefined));
 
 describe("scriptPubKeyHexToBtcAddress", () => {
-  it("decodes a P2WPKH scriptPubKey hex back to its testnet address", () => {
+  it("decodes supported payout scripts on the configured signet network", () => {
```
gbarkhatov: Each network test bundles five distinct behaviours into one `it`, so the first failure hides the other four.

This test and its mainnet sibling each assert P2PKH, P2SH, P2WPKH, P2WSH and P2TR in one chain. `expect` throws on first mismatch, so a P2SH version-byte regression aborts before the segwit assertions run, and red CI reports only "decodes supported payout scripts on the configured signet network" — telling the reader nothing about whether taproot also broke, on a path whose output is a withdraw destination.

CLAUDE.md is explicit here: one behaviour per test, the name describes the exact behaviour, and three similar test functions beat one bundle. The setup already lives in `beforeEach`, so each split test is two self-contained lines. Worth noting the two blocks are also asymmetric — signet comes from `beforeEach` while mainnet is reconfigured inline, silently discarding that setup for one test.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
```diff
@@ -1,39 +1,92 @@
...
+    vi.resetModules();
+    const { configureBabylonConfig } = await import(
+      "../../config/network/runtime"
+    );
+    configureBabylonConfig({
+      btcNetwork: "mainnet",
+      ethChainId: 1,
+      ethRpcUrl: process.env.NEXT_PUBLIC_ETH_RPC_URL!,
+    });
+    const { scriptPubKeyHexToBtcAddress } = await import("../btc");
```
gbarkhatov: The mainnet P2TR assertion is coupled to bitcoinjs-lib escaping `vi.resetModules()`. Line 64 resets modules and this line re-imports `../btc`, but `bitcoin.initEccLib(ecc)` is not re-called.

`p2tr` reaches `getEccLib().isXOnlyPoint()` on the *valid* path too, and its cache is module-scoped — so the P2TR expectation passes only because bitcoinjs-lib is externalized and therefore survives the reset. Add `bitcoinjs-lib` to the inline list at any future point and that line turns red with a message pointing nowhere near the cause. One line — `bitcoin.initEccLib(ecc)` after this import — removes the coupling.

Smaller, same file: `ethRpcUrl: process.env.NEXT_PUBLIC_ETH_RPC_URL!` at lines 33 and 71 makes an address-decoding test depend on the `env` block in `vitest.config.ts`, and the `!` is the type-safety escape CLAUDE.md asks us to avoid. `configureBabylonConfig` only requires the field be non-empty, so a literal would keep the test self-contained.

gbarkhatov (dismissed):
Reviewed at `1a27254` via three independent passes, deduped and verified against the source.

**No merge blockers, and all ten asserted addresses are correct.** Two lanes re-derived every one from BIP173/BIP350/base58check using their own encoders rather than bitcoinjs, and I confirmed three directly against BIP350's table. Nothing here can ship a wrong payout address.

**The best thing about this PR is something the description doesn't claim.** Across this review set I've now hit the same trap three times — a test that compares production code against itself (#2496 aliased the `COPY` function the component renders; #2484 and #2486 asserted class strings already true pre-fix). This PR *removes* an instance of it: the old file derived `EXPECTED_TESTNET_ADDRESS` via `bitcoin.payments.p2wpkh(...)` and compared it against `bitcoin.address.fromOutputScript(...)` — bitcoinjs agreeing with itself. Every new expectation is a frozen literal and no `payments.` call remains. That's a real improvement and worth stating.

**"Both networks" is genuine coverage, not cosmetic:** `bc1`/`tb1` HRPs, base58 `1`/`m` and `3`/`2` version bytes, and bech32m on both P2TR. Flipping the ternary in `getBitcoinJsNetwork()` fails both tests, which is the property that matters.

**Refuted — worth not re-raising:**
- *"The new `external` entry conflicts with the `inline` entry above it, and which wins is a resolution-order detail."* Half right, wrong conclusion. Inline *is* checked first in vitest 4.0.9 (`_shouldExternalize` tests inline at 8909, external at 8911) — but the adjacent string entry never matches the realpathed pnpm id, so there's no shadowing and the regex is genuinely required. See my comment on `vitest.config.ts` for the real hazard.
- *"The stated node-env reason is unfalsifiable."* No — `e2e/fixtures/__tests__/replay.test.ts:1-7` documents the same constraint in full: the secp256k1 asm.js build validates with `instanceof Uint8Array`, and under jsdom a Node `Buffer` is cross-realm, so every point is rejected and `initEccLib` fails. The reason is real and precedented. This also rules out "just revert to jsdom" as a fix for my second comment — the split is required.
- *"The tests compare production code against itself"* in its strong form. Refuted, as above.
- *"Order-dependence / config leak between tests."* No — `resetModules` gives `runtime.ts` a fresh null state each time, so `configureBabylonConfig`'s throw-on-re-entry never fires.

**Verified clean:** `vi.unmock("@/config")` and `"@/config/network"` are load-bearing rather than dead, since `src/test/setup.ts` mocks both with a hardcoded signet. "Accepts uppercase hex and its prefix" exercises a real branch — `stripHexPrefix` handles `0X`, without which the `X` would trip the non-hex guard. No dead code in the diff.

**Forward-looking, not blocking:** ten fixtures resting on one hash160, one WSH hash and one x-only key won't satisfy §9's "golden vectors **plus** randomised inputs" when the hand-written encoder in `utils/btc/scriptPubKeyAddress.ts` lands. The blind spots worth noting on the epic: no base58 payload with more than one leading zero byte, and no witness-program length other than 20 and 32.

jrwbabylonlab (dismissed):

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
```diff
@@ -42,6 +95,32 @@ describe("scriptPubKeyHexToBtcAddress", () => {
...
+
+  it("rejects a script without a destination address", () => {
+    expect(() => scriptPubKeyHexToBtcAddress("0x6a")).toThrow(
+      "has no matching Address",
+    );
+  });
+
+  it("rejects a Taproot script whose output key is not on the curve", () => {
+    expect(() =>
+      scriptPubKeyHexToBtcAddress(
```
jeremy-babylonlabs: This throws the same generic string as the "no destination" test above it, so it cannot distinguish an off-curve rejection from a missing ECC backend or a non-matching script — and this file tears the curve backend down between tests (`afterEach` -> `initEccLib(undefined)`), so a regression that silently drops ECC would leave this passing.

Assert the specific failure instead (e.g. call `bitcoin.payments.p2tr({ output })` directly and check for `Invalid pubkey for p2tr`), or additionally assert that a valid x-only key at the same length does decode to a `tb1p...` address.

jeremy-babylonlabs (commented):
Coverage is a good addition. One assertion cannot fail for the reason it names.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
jonybur: No change. The jsdom guard for the byte loop is not this file. `services/vault/src/applications/aave/utils/__tests__/payoutAddresses.test.ts` runs under jsdom, leaves `@/utils/btc` unmocked, and calls the real `scriptPubKeyHexToBtcAddress` through `getUniquePayoutAddresses` with valid P2WPKH scripts. I checked the regression you describe. With the loop replaced by the polyfill `Buffer.from(cleanHex, "hex")`, that file fails 4 of 5 tests under jsdom: `expected [] to have a length of 1 but got +0` at lines 63 and 83, and `expected [] to have a length of 2 but got +0` at lines 72 and 96. So the suite cannot stay green through that change. The narrower point is right. The direct unit test of the converter no longer runs under jsdom, so that guard is now incidental rather than deliberate. A one-line pointer in the `btcUtils.ts` comment naming `payoutAddresses.test.ts` as the jsdom guard would make it deliberate. That is a production-file edit, so it stays out of this test-only PR. Splitting this file for it does not pay for itself.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
jonybur: Fixed in c578deb0. The comment now names each source. The four segwit v0 addresses are the BIP-173 examples. BIP-350's valid table also attests `bc1qw508...kv8f3t4` and `tb1qrp33...q0sl5k7` with their scriptPubKeys, and adds `bc1p0xlxvlh...vqzk5jj0`. The comment states that `tb1p0xlxvlh...vq47zagq` and the four base58 values are derived, not BIP vectors. They are base58check of hash160 `751e76e8199196d454941c45d1b3a323f1433bd6` under version bytes 0x00, 0x6f, 0x05 and 0xc4, recorded from bitcoinjs-lib 6.1.7. The "retain the existing" line is gone. The pre-PR file held one derived address, so there was no baseline to retain.

services/vault/vitest.config.ts
jonybur: Comment fixed in c578deb0. The config is unchanged. The comment now describes the construct and quotes the error. It names the mechanism rather than the minified binding, because a rebuild can rename that letter. In the current bundle the `@keystonehq/sdk` default import is bound to `Y`, and `Y:` is also used as a statement label in nested labelled blocks. Vite 6.4.1 rewrites that label to `__vite_ssr_import_N__.default:`. `scriptPubKeyHexToBtcAddress.test.ts` loads the bundle through the real `@/config` and fails with `SyntaxError: Unexpected token ':'`. The upstream fix is vitejs/vite#22451, merged 2026-05-14 and not in the Vite 6 line this package uses (6.4.2 and 6.4.3 carry only security backports). The comment also states that the neighbouring `inline` string entry never matches the pnpm-realpathed id, so the regex is load-bearing. On the smaller points. The revert is not silent. I removed the regex and the test file fails all 18 tests with that SyntaxError. A regex anchored on `@babylonlabs-io/` would not match the realpathed id either. The entry sits above `projects` because the bundle is reached through `@/config`. The regex matches one module id, `dist/index.es.js`, so it changes how that one module loads in every test that reaches it, and the full vault suite passes with it. A scoped `vi.mock("@babylonlabs-io/wallet-connector", ...)` would replace the real `Network` enum with a fake. That deletes the only assertion chain that pins `Network.MAINNET === "mainnet"` against the literal compare in `getBitcoinJsNetwork()`.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
jonybur: Fixed in c578deb0. The test is now `rejects a Taproot output key that is not a valid x-only point`. It pins both directions in one `it`. The BIP-350 key decodes to `tb1p0xlxvlh...vq47zagq`. `0x5120` followed by 32 bytes of `ff` throws `has no matching Address`. The pair is discriminating. With no ECC backend, `p2tr` throws from `getEccLib()`, the future-segwit fallback rejects witness v1, and the valid key throws instead of decoding, so the first assertion fails. I confirmed this against bitcoinjs-lib 6.1.7. Without `initEccLib` both keys throw `has no matching Address`. With it the valid key decodes and the off-curve key throws. If the curve check were skipped, bech32m would encode the off-curve bytes as `tb1pllll...s5k48qr`, so the second assertion fails. Only the curve check running makes both pass. I did not assert `payments.p2tr({ output }).address` throws `Invalid pubkey for p2tr`. That pins a third-party message and bypasses `src/`.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
jonybur: Fixed in c578deb0. Each network now has five one-assertion tests, `decodes a P2PKH scriptPubKey to its signet address` and so on for P2SH, P2WPKH, P2WSH and P2TR. The signet tests use the file-level `beforeEach` directly. The mainnet tests sit in `describe("on the configured mainnet network")` with their own `beforeEach`. It resets modules, configures mainnet, and binds a local `decode` from a fresh `../btc` import. The inline reconfiguration inside the old mainnet `it` is gone. The file-level `beforeEach` still runs first for every mainnet test, so the ECC backend it installs is in place. A P2SH version-byte regression now reports as one named failure and leaves the segwit results visible.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
jonybur: The `ethRpcUrl` point is fixed in c578deb0. Both `configureBabylonConfig` calls pass the literal `"https://test.example/eth"`. The `!` escape and the dependency on the `env` block in `vitest.config.ts` are gone. I did not add `bitcoin.initEccLib(ecc)` after the mainnet re-import. That line does not remove the coupling. If bitcoinjs-lib were inlined, `vi.resetModules()` would give `../btc` a fresh bitcoinjs-lib instance. `bitcoin.initEccLib(ecc)` through this file's top-level import would then install the backend on the old instance, not the one the converter uses. The file-level `beforeEach` has the same shape, so the exposure is file-wide, not specific to the mainnet block. Under the current config both blocks share one externalized instance, and the file-level `beforeEach` installs the backend before every test, including the mainnet ones. A lost backend is not silent. The valid P2TR decodes throw and their tests fail. If hardening is still wanted, the working form is `const b = await import("bitcoinjs-lib"); b.initEccLib(ecc);` inside `beforeEach`. That binds to the instance the converter gets whether or not the library is inlined. I left it out to keep this PR small.

services/vault/src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts
jonybur: Fixed in c578deb0, together with the thread on line 116. The test now pins both directions in one `it`. The valid BIP-350 key decodes to `tb1p0xlxvlh...vq47zagq`. The same length with 32 bytes of `ff` throws `has no matching Address`. That separates the three causes. A missing ECC backend makes the valid key throw too, because `p2tr` fails in `getEccLib()` and the future-segwit fallback rejects witness v1, so the first assertion fails. A skipped curve check would let bech32m encode the off-curve bytes, so the second assertion fails. A non-matching script cannot produce the first assertion's address. The `afterEach` teardown does not weaken this. `beforeEach` reinstalls the backend before every test, and if it did not, the first assertion goes red. I did not use `payments.p2tr({ output })` with `Invalid pubkey for p2tr`. That pins a third-party message and does not go through `src/`.

jonybur:
Review fixes are pushed as `c578deb0`.

Six threads produced changes, one is answered. The Taproot rejection test now pins both directions in one case, so a lost curve backend fails it instead of passing by accident; the bundled network tests are split so one failure no longer hides four; the provenance comment names which values are BIP vectors and which are derived from bitcoinjs-lib 6.1.7; the config comment states the mechanism it works around.

Local checks at this head: 3,824 Vault tests, eslint clean on both edited files.

jonybur:
@greptileai review the full PR at current commit `c578deb0e0094c8cf9dfda689d614b9c60bd20a6`. Check the diff against its target branch, including the changes since your last review. Verify that the review findings are addressed in the code. Report any remaining issues and the confidence score for this commit.

gbarkhatov (approved):
Follow-up to my review at `1a27254417`. This is a fast-forward of one commit. Nothing new to raise.

| # | Prior finding | Verdict at `c578deb0` |
|---|---|---|
| T1 | Moving the file to the node environment dropped the only test of the jsdom Buffer workaround | **Declined, with reason.** My "only guard" wording was too strong: `payoutAddresses.test.ts` still runs under jsdom against the real converter |
| T2 | The provenance comment overclaimed BIP sourcing | **Fixed** (`:11-19`). The BIP-173 v0 addresses and BIP-350 rows 1, 2 and 8 match; the remaining values are now labelled as derived |
| T3 | The `external` regex was fragile; a scoped mock would be better | **Comment fixed, config change declined, and my point was partly wrong.** Removing the regex fails the suite loudly with the `Unexpected token ':'` error rather than silently reverting. The scoped `vi.mock` I suggested would also have replaced the real `Network` enum |
| T4 | The off-curve test couldn't see the curve check | **Fixed** (`:149-156`). It pairs a valid-key decode with the off-curve rejection, so both a missing ECC backend and a skipped curve check now fail |
| T5 | One `it` per network hid failures | **Fixed.** Now single-assertion tests per script type and network |
| T6 | (i) `initEccLib` not re-called after the mainnet re-import; (ii) `ethRpcUrl` via `process.env…!` | **(ii) Fixed**, now a literal. **(i) Declined, and my suggested fix was wrong:** the file's top-level `bitcoin` binding predates `vi.resetModules()`, so calling `initEccLib` on it would not reach a re-imported instance |

Every expected address is still a frozen literal, and none is derived from the converter under test. The `vitest.config` change affects how one module loads, not which file loads. Vault already consumed wallet-connector's built `dist/` through `workspace:*` before this PR.

PR #2501, opened 2026-09-10
feat(vault): allow confirmed Ethereum-only access

jonybur:
## Outcome

A confirmed Ethereum session can use Vault without Bitcoin when NEXT_PUBLIC_FF_ENABLE_ETH_FIRST=true. The control starts disabled.

Status: 253c5d4a8 includes current main, where the four prerequisite PRs have merged. Every review thread is answered at that commit. Required human reviews remain.

## Change

- Include current main and the latest wallet consent, Bitcoin prompt, disconnect, and payout-address changes.
- Require Bitcoin for signing actions. Reconnect and delayed payout readiness require a separate Retry.
- Preserve active payout controls and reject late results after wallet or consent loss.
- Bind resumed deposits to the Ethereum owner and Bitcoin key registered on chain before signing.
- Poll pending deposits without a Bitcoin key only when no Bitcoin wallet is attached; a connected wallet whose key is still loading polls nothing, as before.
- Offer an unlock path for a locked Bitcoin wallet: in the navbar menu, in the Bitcoin action prompt, and before any Bitcoin action runs.
- Derive the "Connect Bitcoin wallet" reclaim row from the reclaim eligibility model, and keep the deposit button clickable so it opens the Bitcoin prompt when Bitcoin drops mid-dialog.

## Safety

The control requires the exact value true. The disabled state requires both wallets. Account changes keep the full-session reset.

Consent, screening, geofencing, ownership, protocol, and transaction checks remain required. For payout signing, an existing device prompt may stay open after cancellation; its late result is rejected.

WOTS and activation keep the existing behavior: an admitted action can continue after later Bitcoin or consent loss. Their chain checks and wallet approval remain. Payout cancellation for active attempts is main's behavior from #2298 and #2482; this PR keeps it and adds a test that an address, provider or consent change during the liveness read stops the attempt before signing. [The review records this scope](https://github.com/babylonlabs-io/babylon-toolkit/pull/2501#discussion_r4000774047).

The two accepted follow-ups in #2228 remain: #2506 and #2507. No release or enablement is authorized by this PR. Critical signing changes require two qualified human reviews.

## Verification

- Full Vault test suite, lint and TypeScript pass on 253c5d4a8.
- Default-off, consent, reconnect, cancel, delayed readiness, address, owner/key, locked-wallet and polling-ownership regression tests pass.
- Tests use controlled wallet and chain fixtures. Real-wallet and final browser evidence remain in #2355 and #2233.

## Links

Tracks #2232. #2232 stays open. Epic: #2228.

Prerequisites #2482, #2491, #2499 and #2500 are merged. The independent disconnect foundation #2439 is merged. This update carries the deposit-owner fix in #2508 at its current review state; its required reviews remain.

Final browser journey and release: #2233. Consumer session evidence: #2355. Engine work remains separate in #2231 and #2361.


jonybur:
@greptileai review the full PR at current commit 52de111157e444410b158a64401125fd56244981. The review fixes are published and local checks pass. Please check the current code, report any remaining findings, and give the confidence score for this commit. This now includes the current wallet prerequisites, the delayed payout-start fix, and the owner/key fix in #2508. #2506 and #2507 remain accepted follow-ups in #2228.

services/vault/src/components/Wallet/BtcActionGate.tsx
```diff
@@ -0,0 +1,74 @@
...
+    connected,
+    btcConnected,
+    sessionConfirmed,
+    loading,
+    requireBtcWallet,
+  } = useBtcAction();
+  const [started, setStarted] = useState(connected && ready && autoStart);
+
+  // Keep an active flow mounted during a temporary wallet disconnect.
+  if (started) return <>{children}</>;
```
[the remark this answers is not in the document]
jonybur: The late continuation path is real. It is also present on main at 751069c6 before this PR: WOTS and activation continue after component unmount, and their mountedRef checks only control UI updates. The async submission bodies, useActivationState, htlcSecretDerivation, wotsSubmission, and vaultActivationService are unchanged here. Unmounting the gate does not cancel those existing async calls.

The existing checks still bind the operation to the deposit. WOTS reads the key, expected WOTS hash, provider, and transaction hash from the contract. It rejects transaction or derived-key hash mismatches before submission. Activation rejects an invalid secret against the contract hashlock and requests the selected depositor's Ethereum account. It still requires wallet approval and contract simulation. No wrong-key acceptance or wallet-approval bypass was found in the bounded trace.

The approved scope in #2232 requires cancellation of active payout attempts on session change. This PR adds that cancellation. #2354 defines admission and reconfirmation; it does not change every admitted transaction into a cancellable operation. New WOTS and activation clicks still use the current requireBtcWallet check.

The residual is a valid WOTS submission or an Ethereum activation prompt after an already-admitted attempt loses Bitcoin or its receipt. That is existing behavior, not a new P1 bypass introduced by retaining the payout controls. Please reassess the severity and release impact against the base code and approved scope. Removing the retained-child rule would not cancel the old async calls and would remove active payout controls.

jonybur:
@greptileai please re-review current commit 52de111157e444410b158a64401125fd56244981 after the scope and base-code review in https://github.com/babylonlabs-io/babylon-toolkit/pull/2501#discussion_r4000774047 . That reply downgraded the existing WOTS/activation continuation behavior to informational and removed the release blocker. The PR description now states that continuation behavior explicitly. CI passes. Please report any remaining actionable finding and update the confidence score for this commit.

services/vault/src/utils/peginPolling.ts
```diff
@@ -75,7 +77,7 @@ export function getDepositsNeedingPolling(
       // Check if this deposit should be polled
       const shouldPoll =
         contractStatus === ContractStatus.PENDING &&
-        !!btcPublicKey &&
+        (featureFlags.isEthFirstEnabled || !!btcPublicKey) &&
```
gbarkhatov: With the flag on, polling runs without a key, and `isVaultOwnedByWallet(depositKey, undefined)` returns `true` ("assume owned", `vaultWarnings.ts:50-53`). Two consequences:

- **No Bitcoin wallet:** every pending deposit under the Ethereum address polls, so WOTS/activation can show as actionable and `useSigningRequiredNotifications` can nudge for them. Before this PR, polling never ran without a key.
- **Bitcoin connected but key not yet loaded, or failed:** `AppPeginPollingProvider` passes `useBtcPublicKey(btcConnected).publicKey`, which is `undefined` in that window. Deposits registered to a *different* Bitcoin key then read as owned and actionable, where flag-off filters them out. That is a silent fallback on ownership.

The effect is display-only, because signing re-checks via `requireBtcWallet()`, the payout script and the on-chain key. Suggest skipping the key filter only when Bitcoin is actually absent (`btcConnected` is already in `AppPeginPollingProvider`), not whenever the key is undefined, and deciding separately whether the no-wallet case should notify.

services/vault/src/components/Wallet/Connect.tsx
```diff
@@ -104,7 +99,12 @@ export const Connect: React.FC<ConnectProps> = ({ loading = false, text }) => {
   // A silently locked BTC wallet keeps `connected` true (cached session), so it
   // would otherwise render the connected wallet menu. Surface an unlock button
   // in the navbar instead so the user can re-authorize in one click.
-  if (btcLocked && !isGeoBlocked && !isGeoLoading) {
+  if (
+    btcLocked &&
+    !featureFlags.isEthFirstEnabled &&
```
gbarkhatov, line 104: With the flag on, a locked Bitcoin wallet loses the navbar Unlock button. `btcLocked` keeps `btcConnected` true, so the connected menu renders with the cached address and key, while this same PR stops the UTXO query and inscriptions toggle for that state (`:69-81`). `useBtcAction` computes `connected = btcConnected && sessionConfirmed` and ignores `locked`, so reclaim/refund/payout entry points let the locked wallet through, and it only fails later at the liveness probe. Nothing else in the app unlocks it.

Either keep an unlock entry in the menu wired to `handleUnlock`, or treat `locked` as not connected in `useBtcAction` so the prompt appears. If hiding it is intended, the PR body should say why.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
```diff
@@ -161,6 +168,26 @@ export function usePayoutSigningState({
...
+  const identityRef = useRef({
+    connected,
+    btcAddress,
+    btcProvider,
+    btcPublicKey,
+  });
+  useEffect(() => {
+    const previous = identityRef.current;
+    identityRef.current = { connected, btcAddress, btcProvider, btcPublicKey };
+    if (
```
gbarkhatov: The identity abort stops the run but not the wallet/Ledger prompt, so Retry does nothing until that prompt settles. `abortRef.current?.abort()` plus `setError(walletNotConnected)` leaves `inFlightRef.current` true until the pending `signPsbt` resolves in `finally`. Meanwhile Retry goes through `handleSign`, which returns immediately on `inFlightRef.current || signing` (`:199`). If an extension prompt never settles, the view is stuck until remount. The PR body says the prompt stays open, but not that Retry does nothing in the meantime.

Suggest also cancelling the in-flight ceremony here: reuse `handleCancel`'s body against `signingProviderRef.current` when it supports cancellation. Whether an already-disconnected Ledger transport accepts a cancel is unverified.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
```diff
@@ -374,6 +378,15 @@ function InactiveRow({
     depositorBtcPubkey: activity.depositorBtcPubkey,
     isReclaimInFlight,
   });
+  const { connected: btcConnected, requireBtcWallet } = useBtcAction();
+  const showReclaimConnection =
+    FeatureFlags.isEthFirstEnabled &&
```
gbarkhatov: `showReclaimConnection` re-implements reclaim eligibility inline: REDEEMED, reserve unspent, not in flight, payout settled. `useReclaimRowAction` names `models/reclaimEligibility` as the single decision point, and anything that model gates on but this predicate doesn't (e.g. a withdraw pause) offers Connect for a row that then renders a disabled Reclaim after connecting. Deriving it from `getReclaimEligibility({ ...inputs, isOwnedByWallet: true })` (Connect when `available`/`blocked`), or adding a "needs wallet" outcome to the model, keeps one source of truth.

services/vault/src/hooks/deposit/useDepositPageForm.ts
```diff
@@ -214,7 +214,8 @@ export interface UseDepositPageFormResult {
 
 export function useDepositPageForm(): UseDepositPageFormResult {
   const { address: btcAddress, connected: btcConnected } = useBTCWallet();
-  const { isConnected: isWalletConnected } = useConnection();
+  const { isConnected: sessionConnected } = useConnection();
```
gbarkhatov, line 224: Low, flag on only. `isWalletConnected = sessionConnected && btcConnected`, so if Bitcoin drops while the dialog is open, `validations.ts` returns a disabled "Connect your wallet" CTA. The `requireBtcWallet()` prompt in `handleDeposit` can then never be reached, and the user has to close and reopen the dialog. Blocking fee/UTXO work without Bitcoin is right; the CTA could stay enabled so the click opens the prompt.

docs/decisions/2354-wallet-consent.md
```diff
@@ -0,0 +1,26 @@
...
+Status: Authorized on 9 September 2026 for [#2354](https://github.com/babylonlabs-io/babylon-toolkit/issues/2354).
+
+## Decision
+
+- A wallet connection does not confirm consent. The user must select the final **Connect** button.
+- The shared wallet dialog checks the live wallet identities before it calls `acceptTermsOfService`, if supplied. It confirms the session only after that callback succeeds.
+- The callback receives the live address and public key of the first required chain. It also receives the connected wallets. It does not use the order in which wallets connected.
+- Vault requires BTC then ETH. Bitcoin staking requires BTC then BBN. BABY staking requires BBN. Thus the callback identity is BTC, BTC, and BBN, respectively.
+- Neither app supplies a consent callback or has a consent service endpoint. Both use the shared dialog and its saved approval record.
+- A current saved approval restores the session only after live checks match the required wallet, address, public key, and network. A valid restore does not repeat the callback.
```
gbarkhatov: This consent contract is now wrong for flag-on mode. It says "Vault requires BTC then ETH", with BTC as the callback identity, and "It does not enable Ethereum-first mode." With the flag on, `requiredChains` is `["ETH"]` (`VaultWalletConnectionProvider.tsx:256`), and a Bitcoin wallet attached later is not re-confirmed: `useWalletConnect.open` keeps a confirmed session when an optional chain attaches, and `handleConfirm` only runs while unconfirmed. That wallet is still address-screened, and a Bitcoin address change still fully resets. Worth recording the flag-on rules here, since this doc is the reference for what consent covers.

gbarkhatov (dismissed):
No merge blockers. I scoped this to the files only this PR touches and to the seams where it combines #2491, #2482, #2499, #2500 and #2508, all reviewed separately today. Their findings stand on their own PRs and aren't repeated here. `52de111` is an octopus merge of exactly those five heads plus this PR's earlier state, and each head is an ancestor, so nothing drifted. CI is fully green.

Most of what's inline applies only with `NEXT_PUBLIC_FF_ENABLE_ETH_FIRST=true`. The one with a trust flavour is the polling key filter (`peginPolling.ts:80`). Even that affects only what is displayed, because signing still re-checks the key.

**Verified clean:**
- **Flag parsing:** it requires exactly `"true"` (`featureFlags.ts:17-20`).
- **Flag off:** no Ethereum-only path is reachable. `requiredChains` stays `["BTC","ETH"]`, the unlock branch is unchanged, and `getDepositsNeedingPolling` still returns `[]` without a key, so the query stays disabled.
- **Flag on, signing:** nothing signs or broadcasts without Bitcoin. Every such path goes through `requireBtcWallet()`, the deposit form needs `btcConnected`, and payout signing still checks the registered payout script against the wallet's.
- **Polling centralisation:** there is still one `usePeginPollingQuery` in `PeginPollingProvider`. `useBtcAction()` in `InactiveRow` is only context reads.
- **Address screening:** a Bitcoin wallet attached after an Ethereum-only confirmation is still screened (`AddressScreeningProvider` screens whenever `btcAddress` is present).
- **E2E hooks:** `wallet-menu-trigger` renders in both menu variants, and the pending-deposit and reclaim testids are unchanged.

**Relation to my #2482 thread on `ResumeDepositContent.tsx:111`:** the new identity abort in `usePayoutSigningState` covers the "abort a running payout on key loss" half here. The Ledger/extension cancel half is still missing (inline).

**Not wired for release:** `NEXT_PUBLIC_FF_ENABLE_ETH_FIRST` isn't in the vault release workflow's build env or the vault README. Leaving it out of the workflow is consistent with "no enablement authorized", but #2233 will need it, and the README should list it now.

docs/decisions/2354-wallet-consent.md
```diff
@@ -0,0 +1,26 @@
+# Wallet consent rule
+
+Status: Authorized on 9 September 2026 for [#2354](https://github.com/babylonlabs-io/babylon-toolkit/issues/2354).
+
+## Decision
+
+- A wallet connection does not confirm consent. The user must select the final **Connect** button.
+- The shared wallet dialog checks the live wallet identities before it calls `acceptTermsOfService`, if supplied. It confirms the session only after that callback succeeds.
+- The callback receives the live address and public key of the first required chain. It also receives the connected wallets. It does not use the order in which wallets connected.
+- Vault requires BTC then ETH. Bitcoin staking requires BTC then BBN. BABY staking requires BBN. Thus the callback identity is BTC, BTC, and BBN, respectively.
```
jrwbabylonlab: This says Vault always requires BTC then ETH, but VaultWalletConnectionProvider.tsx:256 in the same PR sets requiredChains to ["ETH"] when the flag is on. Note also that #2491's current head deletes this document and the release note, so this PR would resurrect both.

services/vault/src/components/Wallet/BtcActionGate.tsx
```diff
@@ -0,0 +1,74 @@
...
+        {btcConnected && !sessionConfirmed
+          ? COPY.wallet.btcAction.confirmBody
+          : COPY.wallet.btcAction.body}
+      </p>
+      {/* These data-testids exist so the real-wallet E2E
+          (e2e/real/actions/stepMachine.ts) can disambiguate these controls
+          from the progress view's Retry, which shares this gate's "Retry"
+          label. Carry them over if you move or rename the controls. */}
+      <Button
+        disabled={connected && !ready}
```
jrwbabylonlab: With the flag on and the BTC extension locked, `connected` stays true while the key read fails, so this Retry is permanently disabled, `refetch` from useBtcPublicKey is never wired in, and Connect.tsx hides the unlock button under the flag. The user has no way out short of closing the modal and unlocking manually.

services/vault/src/utils/peginPolling.ts
jonybur: Fixed in 253c5d4a8. Polling now takes one explicit input, `btcWalletAbsent`, computed only in AppPeginPollingProvider as `isEthFirstEnabled && !btcConnected`. The ownership rule lives in a single helper, `shouldPollForWallet`: with a key, a deposit polls only if the key matches; with no key, it polls only when Bitcoin is truly absent. A connected wallet whose key is still loading or failed therefore polls nothing, exactly as flag-off did, and the query guard, `getDepositsNeedingPolling` and the two mempool/HTLC filters all use that one rule. Display ownership keeps its assume-owned default so no wrong-wallet hint flashes during the key-loading window. On notifications: in the loading window nothing is actionable, and with no Bitcoin wallet the continuation view offers the action behind the Connect-Bitcoin gate, so a nudge is not a dead end; they stay as they are.

services/vault/src/components/Wallet/Connect.tsx
jonybur: Both halves are in 253c5d4a8. The navbar keeps the Ethereum menu under the flag, since an Ethereum-only session must stay usable while Bitcoin is locked, and that menu now carries an "Unlock wallet" entry through the existing `connectAction` prop, wired to the same unlock the flag-off button uses (`wallet-menu-unlock`). It appears only while Bitcoin is locked; with the flag off the navbar unlock button is unchanged. On the other side, `useBtcAction` now treats a locked wallet as not connected and `requireBtcWallet()` triggers the unlock prompt, so reclaim, refund and payout entry points prompt instead of failing later at the liveness probe. Tests cover the menu entry, its absence when Bitcoin is not locked, and the flag-off prompt.

services/vault/src/components/deposit/PayoutSignModal/usePayoutSigningState.ts
jonybur: Addressed at 253c5d4a8 by taking main's version of this hook. A wallet identity change during an attempt now goes through `handleCancel`: it aborts the poll and calls `cancelSigning` on the provider that started the ceremony when that provider supports it, and `cancelRequested` shows the cancelling state until the call settles. Retry becomes available as soon as the cancelled call returns. A wallet that cannot cancel still has to settle its own prompt first; the description says so. The tests "stops a software attempt on disconnect and waits for an explicit retry", "cancels the provider that started the sign, not a wallet swapped in mid-prompt" and "resets to idle after the provider confirms cancellation" cover the three cases.

services/vault/src/components/vaults/VaultsLifecycleSections.tsx
jonybur: Agreed, the inline predicate was a second copy of the model. In 253c5d4a8 the decision moves into `useReclaimRowAction`: when Ethereum-first is on, no Bitcoin wallet is connected and the depositor key is known, the hook evaluates `getReclaimEligibility` with `isOwnedByWallet: true` and returns the owner's view minus the ability to act, exposed as a new `needsWallet` outcome. Connect is offered only when the model says available; a paused protocol renders the disabled Reclaim with its reason, a spent or in-flight reserve renders the Reclaiming state, and absent stays absent. The row component just reads `needsWallet` and no longer imports the status enum, the flag or `isPayoutSettled`. Tests cover the pause case at the row level plus each `needsWallet` condition in the hook.

services/vault/src/hooks/deposit/useDepositPageForm.ts
jonybur: Fixed in 253c5d4a8. The CTA takes a second input, `canConnectBtcWallet`, true when the session is confirmed, Bitcoin is absent and the flag is on. In that state the not-connected branch returns an enabled "Connect Bitcoin wallet" button instead of the disabled "Connect your wallet". `isWalletConnected` is unchanged, so fee and UTXO work still waits for Bitcoin. The click lands in `handleDeposit`. The form's own reconnect and unlock branch now applies only while a Bitcoin wallet is attached, so with Bitcoin absent the click reaches `requireBtcWallet()` and opens the Bitcoin prompt without closing the dialog, even when an earlier wallet check had failed. With the flag off nothing changes.

docs/decisions/2354-wallet-consent.md
jonybur: This document never reached main: #2491 removed it and the release note before merging, and the team's rule is that tests carry the contract. 253c5d4a8 drops both files from this PR too. The flag-on rules you list are pinned by tests instead: `VaultWalletConnectionProvider.test.tsx` checks the required chains (`["ETH"]` with the control on, `["BTC", "ETH"]` off), the chain order, and that a Bitcoin or Ethereum address change still resets both wallets under either setting; the wallet-connector's `useWalletConnect.optionalChains.test.tsx` checks that a confirmed session survives attaching the optional chain without a second confirmation and that a single required chain is enough to report connected. Address screening of a later Bitcoin wallet is unchanged and covered by the existing screening tests.

docs/decisions/2354-wallet-consent.md
jonybur: Correct on both counts. 253c5d4a8 merges current main, where #2491 landed without this document or the release note, and removes both from this PR so nothing is resurrected. The flag-on requirement (`requiredChains` of `["ETH"]`) is asserted in `VaultWalletConnectionProvider.test.tsx` for the control on, off and unset.

services/vault/src/components/Wallet/BtcActionGate.tsx
jonybur: Fixed in 253c5d4a8. `useBtcAction` now treats a locked extension as not connected, and `requireBtcWallet()` triggers the wallet's unlock prompt through the shared unlock hook. The gate no longer falls into the loader for a locked wallet: it shows the "Bitcoin wallet locked" heading with an "Unlock wallet" button (`btc-action-unlock`), and once the provider clears `locked` the gated flow resumes on its own. `useBtcPublicKey` re-reads the key whenever `locked` flips, so the stuck key error clears after the unlock without wiring `refetch` by hand. The navbar menu also keeps an unlock entry under the flag, and the deposit form runs its own unlock path ahead of the prompt so its "Unlocking wallet..." state still shows. Tests cover the locked state in the hook, the gate and the key hook.

jonybur:
@greptileai review

253c5d4a8 merges current main (the four prerequisite PRs landed today), carries the #2508 follow-up at its current review state, and answers the eight open review threads: polling without a Bitcoin key only when no Bitcoin wallet is attached, an unlock path for a locked Bitcoin wallet in the navbar menu and the Bitcoin action prompt, the reclaim "Connect Bitcoin wallet" row derived from the eligibility model, a clickable deposit button that opens the Bitcoin prompt when Bitcoin drops mid-dialog, and the two consent documents that never reached main removed from this branch.

jrwbabylonlab (approved):

gbarkhatov (approved):

PR #2502, opened 2026-09-10
feat(ts-sdk, vault): enforce the on-chain maxFundingInputCount bound on the Pre-Pegin funding path

jeremy-babylonlabs:
## Summary

Enforces the new on-chain `maxFundingInputCount` bound on the Pre-Pegin funding path so a depositor with a fragmented wallet is stopped before the signing ceremony instead of at an undecodable revert.

Decision from the #tbv-protocol and #web threads: no consolidation transaction, the frontend caps UTXO selection at the contract's bound (20 on devnet), tells the user passively once amount and vault provider are filled, and blocks the deposit when it cannot fit.

## SDK (`@babylonlabs-io/ts-sdk`)

- `ProtocolParamsReader.getMaxFundingInputCount(blockNumber?)` reads the 8-component `getTBVProtocolParams` through its own ABI const and its own call, never through the 6-component tuple that feeds the blocking `ProtocolParamsContext`. It classifies the raw return by word count: 6 or 7 words is a deployment that predates the field (`unsupported`), 8 or more words decodes the first 8, so a field appended later by governance does not halt deposits. Fewer than 6 words or a length that is not whole 32-byte words throws. A value of `0` is `unpublished` and never becomes "no bound". Published values are validated into `[1, 255]`.
- `selectUtxosForPegin` takes a required `maxInputCount: number | null`. The cap check runs only once a prefix of the sorted valid set could fund `peginAmount + fee`; a wallet whose whole valid set falls short reaches the insufficient-funds throw, so the depositor is told to add funds rather than to consolidate. At the cap it throws the typed `FundingInputCountExceededError` before pushing input `max + 1`. Exactly `max` inputs is accepted, matching the registry's exclusive `>` check.
- `FundingInputCountExceededError` lives in its own dependency-free module, `utils/utxo/fundingInputCountExceeded.ts`, and is re-exported from `selectUtxos.ts`. The registry ABI adds `TooManyFundingInputs(uint256,uint256)`, `CONTRACT_ERRORS` carries its selector, and `handleContractError` throws the typed error when the revert payload decodes, so the registration path maps it to the same copy as the selection path. The split keeps `bitcoinjs-lib` out of the ETH-only `contracts` subpath.
- `computeFundingBudget` returns the top-N spendable UTXOs with the selector's own filter and ordering, so the Max the form offers is always fundable under the bound.
- `PeginManager.preparePegin` takes `maxFundingInputCount`, enforces it at selection (before `deriveVaultRoot` and any signing or approval prompt), and re-asserts the funded transaction's declared input count against both the selection and the bound, throwing the typed error in both cases.

## Vault app

- New `useFundingInputBound` query on its own key. Blocked states (`unpublished`, read error) poll every 60 s so the CTA recovers without a reload. A failed background refetch sets `status: "error"` in query-core, so polling continues and the form fails closed.
- `useEstimatedBtcFee` sizes Max from the capped budget and passes the bound to selection. The SDK error is mapped to copy by type guard; the raw message never reaches the CTA label.
- CTA: `unpublished` blocks with "Peg-ins temporarily unavailable", a failed read blocks with "Unable to verify UTXO limit — please try again". A deployment without the field behaves exactly as today.
- Passive notice once amount and vault provider are filled, shown only when the capped budget's balance is below the uncapped budget's balance and the Max is not clamped by the supply cap: "You can deposit up to X BTC with your current Bitcoin UTXOs". Invalid-script UTXOs never trigger it because both budgets use the selector's filter.
- `useDepositFlow` reads the bound at the pinned block, refreshes the form cache from that read, and aborts with a "Peg-ins temporarily unavailable" callout if it is `unpublished`. The abort sits outside the drift `try/catch`, so it never overwrites the peg-in config cache; the `unpublished`/`unsupported`/`published` mapping into the build is an exhaustive switch.
- Contract error messages for `TooManyFundingInputs` and `TooManyHtlcOutputs`.

## Critical path review

`selectUtxos.ts`, `peginFeeMath` tests and `PeginManager.ts` are on the fee-calculation-consistency critical path (CLAUDE.md section 2). This needs two reviewers and an author line-walk. The estimator-versus-selection agreement test in `peginFeeMath.test.ts` pins that the capped Max is fundable with exactly `max` inputs and that one satoshi more throws the typed error.

The issue's scoped plan put the pinned-block re-check in `pinnedBuildLimits.ts`. It lives in `PeginManager.preparePegin` instead (selection throw plus the funded-transaction assert, both fed by the bound read at the pinned block), so there is one enforcement site rather than two that must agree.

## Out of scope

- UTXO consolidation UX.
- `SPLIT_TX_FEE_SAFETY_MULTIPLIER` deletion (dead constant, separate change).
- Excluding dust past the top N from the cap notice. The two-budget comparison still counts dust as consolidatable; a `computeMaxDeposit` comparison would be needed to drop it.

## Testing

- SDK: selection at bound / one over / null bound / 25 small UTXOs below the amount → insufficient funds; budget helper; estimator-vs-selection agreement; reader `published` / `unpublished` / `unsupported` (6 and 7 words) / 9 words decodes / non-whole-word length throws / RPC error propagates / `blockNumber` forwarded; manager rejects before derive and sign, and throws the typed error when the funder returns an extra input; `handleContractError` on both revert branches; ABI selector agreement.
- App: estimator loading while unresolved, copy label not SDK message, Max from capped budget; the bound is read at the pinned block; drift path seeds the bound cache, aborts, and leaves the peg-in config cache untouched; `resolveMaxInputCount` four cases; CTA disabled in `unpublished`; cap notice on/off for the supply-cap and invalid-script cases; error mapping for both new callouts.
- `pnpm --filter @babylonlabs-io/ts-sdk run test` — 118 files, 2029 tests. `pnpm --filter vault run test` — 317 files, 3874 tests. Lint and `tsc --noEmit` clean on both.

Refs #2402

### DIFF
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/eth/protocol-params-reader.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/eth/protocol-params-reader.ts
@@ -6,7 +6,7 @@
  */
 
 import type { Abi, Address, Hex, PublicClient } from "viem";
-import { decodeFunctionResult, encodeFunctionData, size } from "viem";
+import { decodeFunctionResult, encodeFunctionData, size, slice } from "viem";
 
 import {
   ProtocolParamsABI,
@@ -241,6 +241,11 @@ export class ViemProtocolParamsReader implements ProtocolParamsReader {
       throw new Error("getTBVProtocolParams returned no data");
     }
 
+    if (size(data) % ABI_WORD_BYTES !== 0) {
+      throw new Error(
+        `Invalid getTBVProtocolParams return: expected whole 32-byte words, got ${size(data)} bytes`,
+      );
+    }
     const words = size(data) / ABI_WORD_BYTES;
     if (
       (TBV_PROTOCOL_PARAMS_LEGACY_WORD_COUNTS as readonly number[]).includes(
@@ -249,16 +254,21 @@ export class ViemProtocolParamsReader implements ProtocolParamsReader {
     ) {
       return { status: "unsupported" };
     }
-    if (words !== TBV_PROTOCOL_PARAMS_WORD_COUNT) {
+    if (words < TBV_PROTOCOL_PARAMS_WORD_COUNT) {
       throw new Error(
-        `Invalid getTBVProtocolParams return: expected ${TBV_PROTOCOL_PARAMS_WORD_COUNT} words (or a legacy 6/7-word tuple), got ${size(data)} bytes`,
+        `Invalid getTBVProtocolParams return: expected at least ${TBV_PROTOCOL_PARAMS_WORD_COUNT} words (or a legacy 6/7-word tuple), got ${size(data)} bytes`,
       );
     }
 
+    // `TBVProtocolParams` is append-only — it has grown twice, and
+    // `IProtocolParams.sol` documents each new field as appended last. A 9th
+    // word is therefore a field this SDK does not know about, not a corrupt
+    // payload, so decode the prefix this SDK does know and ignore the rest
+    // rather than halting every deposit until the next SDK release.
     const result = decodeFunctionResult({
       abi: ProtocolParamsFundingInputBoundABI,
       functionName: "getTBVProtocolParams",
-      data,
+      data: slice(data, 0, TBV_PROTOCOL_PARAMS_WORD_COUNT * ABI_WORD_BYTES),
     });
 
     const raw = result.maxFundingInputCount;
--- a/packages/babylon-ts-sdk/src/tbv/core/clients/eth/protocol-params-reader.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/clients/eth/protocol-params-reader.ts
@@ -6,16 +6,22 @@
  */
 
 import type { Abi, Address, Hex, PublicClient } from "viem";
+import { decodeFunctionResult, encodeFunctionData, size } from "viem";
 
-import { ProtocolParamsABI } from "../../contracts/abis/ProtocolParams.abi";
+import {
+  ProtocolParamsABI,
+  ProtocolParamsFundingInputBoundABI,
+} from "../../contracts/abis/ProtocolParams.abi";
 import {
   assertValidOffchainParamsVersion,
+  validateMaxFundingInputCount,
   validateOffchainParams,
   validatePegInConfiguration,
   validateTBVProtocolParams,
 } from "./protocol-params-validation";
 import type {
   AllOffchainParamsData,
+  FundingInputBound,
   OnSkippedOffchainParamsVersion,
   PegInConfiguration,
   ProtocolParamsReader,
@@ -29,6 +35,18 @@ import type {
  */
 const UINT16_MAX = 65535;
 
+/** Width of one ABI-encoded word, in bytes. */
+const ABI_WORD_BYTES = 32;
+
+/**
+ * Return lengths of the two `TBVProtocolParams` struct generations that
+ * predate `maxFundingInputCount`: the original 6-component struct, and the
+ * 7-component one that appended `peginActivationDelay`.
+ */
+const TBV_PROTOCOL_PARAMS_LEGACY_WORD_COUNTS = [6, 7] as const;
+
+/** Word count of the current 8-component `TBVProtocolParams` struct. */
+const TBV_PROTOCOL_PARAMS_WORD_COUNT = 8;
 
 /**
  * Raw shape viem returns for VersionedOffchainParams struct.
@@ -206,6 +224,49 @@ export class ViemProtocolParamsReader implements ProtocolParamsReader {
       );
     }
     return raw;
+  }
+
+  async getMaxFundingInputCount(
+    blockNumber?: bigint,
+  ): Promise<FundingInputBound> {
+    const { data } = await this.publicClient.call({
+      to: this.contractAddress,
+      data: encodeFunctionData({
+        abi: ProtocolParamsFundingInputBoundABI,
+        functionName: "getTBVProtocolParams",
+      }),
+      blockNumber,
+    });
+    if (data === undefined) {
+      throw new Error("getTBVProtocolParams returned no data");
+    }
+
+    const words = size(data) / ABI_WORD_BYTES;
+    if (
+      (TBV_PROTOCOL_PARAMS_LEGACY_WORD_COUNTS as readonly number[]).includes(
+        words,
+      )
+    ) {
+      return { status: "unsupported" };
+    }
+    if (words !== TBV_PROTOCOL_PARAMS_WORD_COUNT) {
+      throw new Error(
+        `Invalid getTBVProtocolParams return: expected ${TBV_PROTOCOL_PARAMS_WORD_COUNT} words (or a legacy 6/7-word tuple), got ${size(data)} bytes`,
+      );
+    }
+
+    const result = decodeFunctionResult({
+      abi: ProtocolParamsFundingInputBoundABI,
+      functionName: "getTBVProtocolParams",
+      data,
+    });
+
+    const raw = result.maxFundingInputCount;
+    if (raw === 0) {
+      return { status: "unpublished" };
+    }
+    validateMaxFundingInputCount(raw);
+    return { status: "published", maxInputs: raw };
   }
 
   /**
--- a/packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
@@ -79,6 +79,7 @@ import {
 } from "../primitives/utils/bitcoin";
 import {
   calculateBtcTxHash,
+  FundingInputCountExceededError,
   fundPeginTransaction,
   getNetwork,
   getPsbtInputFields,
@@ -284,7 +285,7 @@ export interface PreparePeginParams {
    * pinned block — the most inputs one Pre-PegIn may spend. `null` only
    * when the deployment predates the field.
    *
-   * Enforced at UTXO selection, before any wallet or device I/O, and
+   * Enforced at UTXO selection, before any signing or approval prompt, and
    * re-asserted against the funded transaction.
    */
   maxFundingInputCount: number | null;
@@ -1055,15 +1056,14 @@ export class PeginManager {
           `spend exactly the selected inputs.`,
       );
     }
+    // Typed, not a plain Error: the consuming app routes this class to the
+    // "Too many UTXOs" copy, and a generic failure here would look like a
+    // bug in the app rather than a bound the depositor can act on.
     if (
       params.maxFundingInputCount !== null &&
       fundedInputCount > params.maxFundingInputCount
     ) {
-      throw new Error(
-        `Pre-PegIn declares ${fundedInputCount} inputs, above the on-chain ` +
-          `maxFundingInputCount of ${params.maxFundingInputCount}; refusing to ` +
-          `submit a transaction the registry will reject.`,
-      );
+      throw new FundingInputCountExceededError(params.maxFundingInputCount);
     }
 
     const prePeginTxid = stripHexPrefix(
--- a/packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
+++ b/packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
@@ -280,6 +280,16 @@ export interface PreparePeginParams {
   availableUTXOs: readonly UTXO[];
 
   /**
+   * On-chain `ProtocolParams.maxFundingInputCount` read at this build's
+   * pinned block — the most inputs one Pre-PegIn may spend. `null` only
+   * when the deployment predates the field.
+   *
+   * Enforced at UTXO selection, before any wallet or device I/O, and
+   * re-asserted against the funded transaction.
+   */
+  maxFundingInputCount: number | null;
+
+  /**
    * Bitcoin address for receiving change from the Pre-PegIn transaction.
    */
   changeAddress: string;
@@ -874,6 +884,7 @@ export class PeginManager {
       prePegin.totalOutputValue,
       params.mempoolFeeRate,
       peginOutputCount(prePegin.htlcValues.length, true),
+      params.maxFundingInputCount,
     );
 
     return {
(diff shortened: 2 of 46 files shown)


packages/babylon-ts-sdk/src/tbv/core/utils/utxo/selectUtxos.ts
```diff
@@ -131,6 +218,12 @@ export function selectUtxosForPegin(
   // selector charged for — silent depositor overpayment at the dust
   // boundary.
   for (const utxo of sortedUTXOs) {
+    // Reject before the push, so a selection that would need input
+    // `maxInputCount + 1` never reaches the fee math or the caller — the
+    // registry rejects `inputCount > maxFundingInputCount`.
+    if (maxInputCount !== null && selectedUTXOs.length >= maxInputCount) {
+      throw new FundingInputCountExceededError(maxInputCount);
+    }
```
gbarkhatov: If the wallet has more than `maxInputCount` valid UTXOs but even all of them together can't cover `peginAmount + fee`, this reports "too many UTXOs" instead of "insufficient funds". The loop takes the `continue` at line 237 for `maxInputCount` rounds, then on the next round throws `FundingInputCountExceededError` here, so the insufficient-funds throw at the bottom is never reached. In the dApp that maps to `COPY.deposit.errors.tooManyFundingInputs` ("consolidate your UTXOs"), which can't fix a shortfall, and SDK integrators get the wrong typed error.

It's reachable through `FeeRateSelector` (a higher fee rate), or at build time when UTXOs have shrunk since the form loaded. Otherwise the form's balance gate usually fires first.

Suggested fix: in the cap branch, throw the cap error only when the full sorted valid set could cover the target, and otherwise fall through to insufficient funds. A test with 25 small UTXOs and an amount above their total pins it.

services/vault/src/utils/errors/errorMessages.ts
```diff
@@ -65,6 +65,10 @@ export const CONTRACT_ERROR_MESSAGES: Record<string, string> = {
     "This Pre-Pegin output has already been used to activate another BTCVault.",
   PeginTransactionAlreadyUsed:
     "This peg-in transaction has already been used to activate another BTCVault.",
+  TooManyFundingInputs:
+    "This deposit spends more Bitcoin UTXOs than the protocol allows in one Pre-Pegin transaction. Consolidate your UTXOs and try again.",
+  TooManyHtlcOutputs:
+    "This deposit creates more BTCVaults than the protocol allows in one Pre-Pegin transaction. Try again without splitting.",
```
gbarkhatov, line 71: As far as I can trace, these two messages can't be shown on peg-in registration, the one path that raises these reverts. Registration goes through the SDK: `pegin-registration-client.ts` calls `handleContractError`, which only consults the selector table `CONTRACT_ERRORS` in `contracts/errors.ts`. That table has no entry for `TooManyFundingInputs` or `TooManyHtlcOutputs`, so the raw error is rethrown. `CONTRACT_ERROR_MESSAGES` is read only by `mapViemErrorToContractError`, and its callers (`transactionFactory`, the Aave hooks) don't submit a peg-in.

This follows what #2402 prescribed, so it's a gap in that approach rather than a deviation from it. The window is narrow (the limit lowered between the pinned read and inclusion) and no BTC moves. Still, the depositor would see a raw revert. The SDK-side fix, outside this hunk, is to add the `TooManyFundingInputs` selector to `CONTRACT_ERRORS`, or better, throw a typed error in `handleContractError` the way `PeginFingerprintChangedError` is handled, and map it to `ERRORS.tooManyFundingInputs`.

packages/babylon-ts-sdk/src/tbv/core/clients/eth/protocol-params-reader.ts
```diff
@@ -208,6 +226,49 @@ export class ViemProtocolParamsReader implements ProtocolParamsReader {
...
+        words,
+      )
+    ) {
+      return { status: "unsupported" };
+    }
+    if (words !== TBV_PROTOCOL_PARAMS_WORD_COUNT) {
+      throw new Error(
+        `Invalid getTBVProtocolParams return: expected ${TBV_PROTOCOL_PARAMS_WORD_COUNT} words (or a legacy 6/7-word tuple), got ${size(data)} bytes`,
+      );
+    }
```
gbarkhatov, line 261: A design question rather than a bug. Any return other than exactly 6, 7 or 8 words throws, so if governance appends a 9th field to `TBVProtocolParams`, every deposit halts until an SDK release. The form sits in "unable to verify", and the pinned build read in `useDepositFlow` throws. It fails closed, which is the right direction, but this struct is append-only: it has grown twice, `IProtocolParams.sol` documents the new field as "declared last … appended", and the existing 6-field read already tolerates trailing words.

One option: accept `words >= 8` and decode only the first 8 words, still throwing below 6 words and on lengths that aren't whole words. If halting on an unknown layout is deliberate, a line in the doc comment saying so would stop the next reader wondering.

services/vault/src/hooks/deposit/useDepositPageForm.ts
```diff
@@ -751,6 +787,12 @@ export function useDepositPageForm(): UseDepositPageFormResult {
     appVersionUnsupported,
     p2aAnchorValueSats: p2aAnchorValueSats ?? null,
     ordinalsCheckPending,
+    maxInputCount,
+    fundingInputBoundUnpublished: fundingInputBound?.status === "unpublished",
+    fundingInputBoundUnavailable,
+    fundingInputCapBites:
+      fundingInputBound?.status === "published" &&
+      (spendableMempoolUTXOs?.length ?? 0) > fundingInputBound.maxInputs,
```
gbarkhatov: This adds a case to the open thread here. `fundingInputCapBites` is just a count comparison, but `DepositForm` documents the prop as "the bound, not the balance, caps the depositable maximum", and shows the "consolidate your UTXOs" notice on that basis. Besides the invalid-script UTXOs already raised, the notice also misleads when:
- the supply-cap clamp in `adjustedMaxDepositSats` (`effectiveRemaining`) is the real ceiling, or
- the UTXOs past the top N are dust, so consolidating adds nothing.

A tighter predicate: set it only when `computeFundingBudget(utxos, max).totalBalance < computeFundingBudget(utxos, null).totalBalance`, and the Max isn't clamped by `effectiveRemaining`.

packages/babylon-ts-sdk/src/tbv/core/managers/PeginManager.ts
```diff
@@ -1032,6 +1043,29 @@ export class PeginManager {
...
+    if (
+      params.maxFundingInputCount !== null &&
+      fundedInputCount > params.maxFundingInputCount
+    ) {
+      throw new Error(
+        `Pre-PegIn declares ${fundedInputCount} inputs, above the on-chain ` +
+          `maxFundingInputCount of ${params.maxFundingInputCount}; refusing to ` +
+          `submit a transaction the registry will reject.`,
+      );
+    }
```
gbarkhatov, line 1067: This is the broadcast-site cross-check §2 asks for, so it's the right place. Two small gaps:
- It throws a plain `Error`, so if it ever fired it would bypass `isFundingInputCountExceededError` and the user would get a generic failure instead of the "Too many UTXOs" copy. `FundingInputCountExceededError` fits here.
- Neither this branch nor the `fundedInputCount !== sizing.selectedUTXOs.length` branch above has a test. Grepping for either message finds only this file, and the #2402 test plan asked for "the post-funding assertion fires". One test that hands the funder an extra input would pin both.

services/vault/src/hooks/deposit/useFundingInputBound.ts
```diff
@@ -0,0 +1,69 @@
...
+ * the estimator `null` would fail open by selecting without any cap.
+ */
+export function resolveMaxInputCount(
+  bound: FundingInputBound | undefined,
+): number | null | undefined {
+  if (bound === undefined) return undefined;
+  if (bound.status === "unsupported") return null;
+  if (bound.status === "published") return bound.maxInputs;
+  return undefined;
+}
```
gbarkhatov, line 69: This mapping is what keeps the form from running uncapped (`unpublished` → `undefined`, never `null`), and nothing tests it. There's no test file for this hook, and both consumer tests mock `resolveMaxInputCount` to return `20`. The `fundingInputBoundUnpublished` / `fundingInputBoundUnavailable` CTA gates and the 60-second blocked-state polling are untested too.

A regression wouldn't produce an over-limit transaction, because `useDepositFlow` still aborts on `unpublished`. It would let a depositor start a deposit that can't finish. A four-case unit test for `resolveMaxInputCount`, plus one disabled-CTA render, would cover it.

services/vault/src/hooks/deposit/__tests__/useDepositFlow.test.tsx
```diff
@@ -1170,6 +1180,63 @@ describe("useDepositFlow", () => {
...
+      );
+
+      // Governance lowered the bound below the 20 the form sized its Max
+      // against. Nothing aborts here — the SDK would reject the selection
+      // outside the drift guard — so only the unconditional refresh stops a
+      // restarted form offering the same unfundable Max again.
+      const lowered: FundingInputBound = { status: "published", maxInputs: 10 };
+      chainMocks.getMaxFundingInputCount.mockResolvedValueOnce(lowered);
+
+      const { result } = renderHook(() => useDepositFlow(MOCK_PARAMS));
```
gbarkhatov, line 1232: The production call reads the bound at the pinned block (`getMaxFundingInputCount(pinnedBlock)`), but no test asserts that. The existing pinned-block assertions cover only `getPegInConfiguration` and the participant keys, so dropping `pinnedBlock` from the new call leaves the suite green. One line fixes it: `expect(chainMocks.getMaxFundingInputCount).toHaveBeenCalledWith(chainMocks.pinnedBlock)`.

services/vault/src/hooks/deposit/useDepositFlow.ts
```diff
@@ -685,6 +698,17 @@ export function useDepositFlow(
...
+          // A chain that publishes no funding-input bound rejects every
+          // Pre-PegIn built against it, so there is nothing to build. Raised as
+          // drift because the form gated on a bound that has since gone away,
+          // and the recovery is the same: stop before anything is signed.
+          if (fundingInputBound.status === "unpublished") {
+            throw new BuildLimitsDriftError(
+              "Deposit limits changed while preparing this deposit: the chain no longer publishes a funding-input bound, so peg-ins are paused.",
+              "funding-inputs",
+            );
+          }
```
gbarkhatov: Nit: because this abort sits inside the drift `try`, the `catch` below also overwrites the peg-in config cache with the `head − 2` build config. The comment there justifies that overwrite only "on the axis that failed", and the config didn't fail here. Throwing after the `try/catch` keeps the overwrite scoped. Relatedly, the later `unpublished` → `null` ("no cap") mapping relies on this throw having happened; an exhaustive switch there would make that dependency explicit.

services/vault/src/services/vault/vaultTransactionService.ts
```diff
@@ -63,6 +63,13 @@ export interface PreparePeginParams {
   councilQuorum: number;
   /** N in M-of-N council multisig */
   councilSize: number;
+  /**
+   * On-chain `ProtocolParams.maxFundingInputCount` read at the build's pinned
+   * block; `null` only when the deployment predates the field. Enforced at UTXO
+   * selection before any wallet or device I/O, and re-asserted against the
+   * funded transaction.
+   */
```
gbarkhatov, line 71: Nit: "before any wallet or device I/O" is slightly too strong. `PeginManager` calls `btcWallet.getPublicKeyHex()` and `requireChangeAddress` before the sizing pass that selects UTXOs. What's actually guaranteed is "before any signing or approval prompt" (selection runs before `deriveVaultRoot`).

gbarkhatov (dismissed):
**Headline:** the bound is enforced where it matters. Selection throws before `deriveVaultRoot` and before any signing prompt. The dApp estimator reuses the SDK selector rather than re-implementing it, so the two can't drift. The read has its own 8-field ABI, is pinned to the build block, and has no default. I found no defect that should block merging on the code itself. The inline comments are all low to medium, mostly error-path accuracy and missing tests.

**Merge-blocker (outside the code):** the PR's only commit, `4489ac8c6`, is unsigned (`verification.reason: "unsigned"`). That's why the required `validate` check (Authenticate Commits → Validate commit signatures) fails. It needs re-signing and a force-push.

**What I checked, per question:**
- **Enforced before any Bitcoin signing:** yes. `selectUtxosForPegin` throws `FundingInputCountExceededError` inside the sizing pass, and that runs before `deriveVaultRoot`. Resumed deposits reuse the stored funded Pre-PegIn and never re-select, and the registry already checked the input count at registration.
- **§2 estimator vs. selector:** they agree by construction. `useEstimatedBtcFee` and the Max computation call the SDK's own `selectUtxosForPegin` and `computeFundingBudget` with the same cap, over the shared `selectableUtxosLargestFirst`. `PeginManager` re-checks the funded transaction's input count at the broadcast site.
- **The read:** a separate `ProtocolParamsFundingInputBoundABI`, a length check before decoding, `[1, 255]` validation, read at the same `pinnedBlock` as `getPegInConfiguration`, and `0` mapped to `unpublished`, which blocks. No `?? 20` anywhere.
- **Multi-vault:** one selection funds the whole batch, so the limit is N inputs in total, not N per vault. That matches the registry's per-request check.
- **Contract parity:** `uint8 maxFundingInputCount` is the last struct field, the registry check is `>` (matching the selector's pre-push `length >= max`), and `TooManyFundingInputs(uint256,uint256)` matches.

**Refuted** (considered and ruled out):
- *"Gas estimation's `0xab` placeholder decodes as a 171-input CompactSize and reverts `TooManyFundingInputs(171, 20)`."* That path is gone. `estimateSubmitPeginRequestBatchGas` no longer exists at this head, and registration gas is estimated on the real calldata. The deferred item in #2402 about this can be closed as obsolete.
- *"The limit isn't enforced when a deposit is resumed."* A resume never re-selects inputs or re-registers.
- *"The dApp estimator and SDK selector could disagree."* The estimator doesn't re-implement selection; it calls the SDK with the same cap.
- *"The read can hit viem's same-name first-match or too-wide-tuple hazard."* The new read uses its own ABI const, and the 6-field `ProtocolParamsABI` is untouched.

**Verified clean:**
- The §4 frozen vault-secret primitives are untouched. `PeginManager` only gains a parameter and checks, and the shared filter and stable sort leave the funding outpoints, and therefore `buildVaultContext`, byte-identical.
- Every new user-visible string is in `copy.ts` or `errorMessages.ts`.
- `docs/api` is regenerated, and `verify` passes.
- Two-reviewer rule: `selectUtxos.ts` and `useEstimatedBtcFee.ts` are both CLAUDE.md §2 files.

PR #2504, opened 2026-09-12
feat(vault): cap one Pre-Pegin at the 20 largest wallet UTXOs

### LINKED ISSUE #2402: Enforce the contract's `maxFundingInputCount` bound on the Pre-PegIn funding path

jeremy-babylonlabs:
Closes #2402. Supersedes #2502, which enforced the bound inside the SDK; the team settled on a frontend-only guard rail in [#tbv-protocol](https://babylonlabsworkspace.slack.com/archives/C087XA9MTMX/p1788253574630889) and [#web](https://babylonlabsworkspace.slack.com/archives/C087NF7U18X/p1788995676250559).

## Summary

The contract bounds a Pre-PegIn at `maxFundingInputCount` inputs, 20 on devnet, and reverts a registration that exceeds it only after the whole BTC signing ceremony. The dApp now caps UTXO selection at the 20 largest and blocks a deposit whose amount those 20 cannot fund, before any wallet or device prompt. The contract stays the enforcement; this only keeps the dApp from building a transaction the contract is guaranteed to reject.

No SDK changes. `selectUtxosForPegin`, `PeginManager` and the protocol-params reader are untouched, so `docs/api` is unchanged.

## How it works

- `MAX_PRE_PEGIN_FUNDING_INPUTS = 20` and `capFundingUtxos` in `services/deposit/fundingInputCap.ts`. The helper sorts value-descending and takes the top 20, the same order the SDK selector uses, so the estimate and the build always agree on the set in play. The constant mirrors the contract deploy default `INITIAL_MAX_FUNDING_INPUT_COUNT`; if governance moves the bound with `setTBVParams`, the constant moves with it.
- `useEstimatedBtcFee` runs `computeMaxDeposit` and the selection on the capped set, so Max is what the 20 largest UTXOs can fund and the Max button never lands on a blocked CTA. The balance label still shows the wallet's full spendable balance.
- `useDepositFlow` hands the capped set to `preparePegin`, and `usePendingVaultOverlapCheck` selects from it, as a safety net that agrees with the estimate.
- `fundingInputCapExceeded` in `useDepositPageForm`: an amount is entered, the wallet holds more than 20 spendable UTXOs, the amount exceeds the UTXO-capped Max before the supply-cap clamp, and the full balance could still cover it. A supply-cap hit never reads as a UTXO problem, and a wallet within the cap never sees this.
- `DepositForm` renders an inline notice below the "Max to Borrow" row, "You **can use up to 20 UTXOs** per deposit. Please combine your UTXOs to proceed.", and the CTA reads "Consolidate your UTXOs to proceed". The validator branch sits ahead of `amountExceedsMax`, so the user is not told "Insufficient balance" when the wallet does hold enough. Design: [Figma 13622-55993](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=13622-55993), notice node 13673-21621: 18px info icon, 16/8 padding, 8px gap, secondary text with the UTXO count in primary.
- God mode, Deposit & Vaults tab, "Fragment wallet UTXOs into N": splits the connected wallet's spendable balance into N synthetic UTXOs for the form's estimate and cap gate only. `useDepositFlow` keeps building from the real set, so no synthetic outpoint can reach a PSBT or a wallet prompt. Lets anyone reproduce the capped Max and the notice without holding 21+ confirmed signet UTXOs.
- `errorMessages.ts` gains copy for `TooManyFundingInputs` and `TooManyHtlcOutputs`. With the cap ahead of the ceremony this only fires if governance lowers the bound between the estimate and inclusion.

## Not done, on purpose

- No contract read of the bound and no `unpublished` (0) handling. A zero bound is a governance outage every submission hits; the dApp cannot tell it apart from a lowered bound and does not try.
- UTXO consolidation UX.
- `SPLIT_TX_FEE_SAFETY_MULTIPLIER` deletion and the gas-estimator placeholder, both still latent.

## Testing

- `capFundingUtxos`: 25 UTXOs give the 20 largest in descending order, 20 or fewer pass through, input not mutated.
- Estimator versus selection over the capped set, in the contracts vitest project against the real SDK: the capped Max is fundable from at most 20 inputs at the same fee rate, and one satoshi more throws. It lives there because the unmocked selector hits a Buffer dual-package mismatch under jsdom.
- `validations`: an amount above the capped Max and within the balance disables the CTA with the consolidate label, including with no provider selected; an amount within the Max is unaffected.
- `DepositForm`: the callout renders with the flag and not without.
- `pnpm --filter vault run test` — 319 files, 3871 tests. Lint and `tsc --noEmit` clean.
- `fragmentUtxos`: N fragments, total preserved, unique outpoints, script carried over.
- Verified on devnet with the god-mode knob at N = 30: Max drops to what 20 of 30 fragments fund, an amount between that Max and the balance shows the notice and the consolidate CTA, an amount above the balance shows "Insufficient balance", N ≤ 20 restores normal behaviour.
- Not verified: a real wallet with more than 20 confirmed UTXOs end to end through signing.

Deposit form with the funding-input-cap notice and the consolidate CTA


jonybur (commented):
Three points from tracing the funding-input cap end to end. Greptile's two P1s (hardcoded 20; cap before script validation) are already on the thread — these are three separate substantive findings.

**1. Inline amount error contradicts the CTA label when the cap fires.** `useDepositPageForm.ts:621-627` gates `fundingInputCapExceeded` on `amountSats > utxoCappedMaxSats && amountSats <= btcBalance`, and `getDepositCtaState` surfaces the "Consolidate your UTXOs to proceed" label. But `useDepositValidation.validateAmount` in `services/vault/src/hooks/deposit/useDepositValidation.ts:62-115` never checks this branch — the entered amount is validated against `minDeposit/maxDeposit` and `effectiveRemaining` only. On blur or submit the inline field error therefore reports the generic "Maximum X BTC" from `validateDepositAmount` while the CTA reads "Consolidate your UTXOs to proceed", giving the user two contradictory pieces of guidance for the same input. The PR description's own contract — "the user is not told 'Insufficient balance' when the wallet does hold enough" — is met at the CTA but broken at the inline error.

**2. Broadcast site has no cap assertion.** `services/vault/src/hooks/deposit/useDepositFlow.ts` passes `capFundingUtxos(spendableUTXOs)` to `preparePegin`, but no runtime `assert(selectedUTXOs.length <= MAX_PRE_PEGIN_FUNDING_INPUTS)` sits before the batch registration. Critical path §2 asks for "cross-check assertions at the broadcast site, not only at the estimator." A future refactor that drops the cap at any one site (the estimator, the overlap check, or the flow) ships silently — `fundingInputCapAgreement.test.ts` pins the estimator/selector pair statically, not the flow. A single length assertion on the returned selection before `registerPeginBatchAndWait` closes the drift-detection gap.

**3. Test bloat.** `services/vault/src/components/simple/__tests__/DepositForm.fundingInputCap.test.tsx` has both "does not render the callout when the cap is not exceeded" and "does not render the callout when the prop is omitted". Both traverse the same `fundingInputCapExceeded = false` branch after default destructuring, so they verify the same production behaviour. The "prop omitted" variant is TypeScript default-parameter behaviour rather than a distinct code path — one test suffices.

services/vault/src/hooks/deposit/useDepositPageForm.ts
```diff
@@ -582,24 +596,38 @@ export function useDepositPageForm(): UseDepositPageFormResult {
...
+
+  // Compared against the pre-supply-cap max: a supply-cap hit must not read
+  // as a UTXO problem, and a wallet within the cap never sees this.
+  const fundingInputCapExceeded = useMemo(
+    () =>
+      amountSats > 0n &&
+      estimateUtxos.length > MAX_PRE_PEGIN_FUNDING_INPUTS &&
+      utxoCappedMaxSats !== null &&
+      amountSats > utxoCappedMaxSats &&
+      amountSats <= btcBalance,
```
gbarkhatov: Should fix. `btcBalance` is the plain sum of `availableUTXOs`. `utxoCappedMaxSats` above subtracts the claim value, pegin fee, P2A anchor and safety buffer (and the network fee via `maxDepositSats`), but this upper bound takes off none of them. So an amount just under the full balance, in a wallet with more than 20 UTXOs, now shows "Consolidate your UTXOs to proceed" instead of "Insufficient balance", and that check runs first (`validations.ts:406`). The user pays to consolidate and then gets "Insufficient balance" anyway.

There's a second mismatch: `btcBalance` excludes inscription UTXOs, while `estimateUtxos` includes them when the user turns exclusion off.

One fix: have `useEstimatedBtcFee` also return an uncapped max over all `estimateUtxos`, apply the same reserves, and compare against that.

services/vault/src/utils/errors/errorMessages.ts
```diff
@@ -65,6 +65,10 @@ export const CONTRACT_ERROR_MESSAGES: Record<string, string> = {
     "This Pre-Pegin output has already been used to activate another BTCVault.",
   PeginTransactionAlreadyUsed:
     "This peg-in transaction has already been used to activate another BTCVault.",
+  TooManyFundingInputs:
```
gbarkhatov: Should fix (dead code). Neither new entry can be shown on the deposit path. A registration revert surfaces in the SDK's `estimateGas`, and `pegin-registration-client.ts:499` hands it to the SDK's `handleContractError`. That looks up the SDK's own `CONTRACT_ERRORS`, which has no `TooMany*` names (`TooManyFundingInputs` isn't in `BTCVaultRegistry.abi.ts` either). `CONTRACT_ERROR_MESSAGES` is only read by `mapViemErrorToContractError` (`contract.ts:192`), whose callers are the Aave hooks and `transactionFactory`. `mapDepositError` never consults it.

So the case this copy exists for, governance lowering the limit, shows generic revert text. Either add both selectors to the SDK's `CONTRACT_ERRORS` (an SDK change, so `docs:clean` applies) or drop these two entries until the contract read lands.

services/vault/src/services/deposit/validations.ts
```diff
@@ -399,6 +401,15 @@ export function getDepositCtaState(params: DepositCtaParams): DepositCtaState {
     };
   }
 
+  // Ahead of `amountExceedsMax`: the wallet holds enough, so "Insufficient
+  // balance" would send the user the wrong way.
+  if (params.fundingInputCapExceeded) {
```
gbarkhatov, line 394: Should fix. This branch sits below `maxBelowMinimum(params.maxDepositSats, …)` at `:385`, and `maxDepositSats` is now the capped max. Take a wallet whose 20 largest UTXOs, after reserves, fall below the protocol minimum, while the whole wallet is enough: the button reads "Minimum deposit is X BTC" for an amount that already meets the minimum, and it never offers the one fix that works. Move `fundingInputCapExceeded` above the `maxBelowMinimum` check, or give that check the uncapped max from the `useDepositPageForm.ts:627` comment.

services/vault/src/__tests__/contracts/fundingInputCapAgreement.test.ts
```diff
@@ -0,0 +1,88 @@
...
+// therefore imports no vault module, matching the project's convention.
+
+import {
+  computeMaxDeposit,
+  selectUtxosForPegin,
+  type UTXO,
+} from "@babylonlabs-io/ts-sdk/tbv/core";
+import { describe, expect, it } from "vitest";
+
+const MAX_PRE_PEGIN_FUNDING_INPUTS = 20;
```
gbarkhatov: This re-declares the constant and `capFundingUtxos`, so changing the value or the sort in `fundingInputCap.ts` fails nothing, and every assertion here exercises only the SDK. The stated reason (import no vault module) doesn't apply: `fundingInputCap.ts` imports nothing, so `import { capFundingUtxos, MAX_PRE_PEGIN_FUNDING_INPUTS } from "../../services/deposit/fundingInputCap"` pulls in no jsdom or app setup. It has to be relative because the `contracts` project has no `@/` alias.

Two smaller points:
- `selectedUTXOs.length <= 20` at `:75-77` can't fail, since the input already has exactly 20 entries.
- No test drives `fundingInputCapExceeded` through `useDepositPageForm`.

services/vault/src/copy.ts
```diff
@@ -305,6 +305,12 @@ export const COPY = {
       splitUnavailableProtocolLimit:
         "The protocol currently allows one BTCVault per transaction. BTCVault split unavailable.",
     },
+    fundingInputCap: {
+      noticeBefore: "You ",
+      noticeEmphasis: "can use up to 20 UTXOs",
```
gbarkhatov: Nit. "20" duplicates `MAX_PRE_PEGIN_FUNDING_INPUTS`, which the module doc says moves with governance. An interpolating function fed the constant, as the copy.ts rule suggests, keeps them in step. Also minor: the notice says "combine" while the CTA says "Consolidate".

services/vault/src/services/deposit/fundingInputCap.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Client-side guard rail on the number of wallet UTXOs one Pre-PegIn may
+ * spend. Mirrors the contract deploy default `INITIAL_MAX_FUNDING_INPUT_COUNT`
+ * (`vault-contracts-aave-v4`) — governance can move that bound with
+ * `setTBVParams`, and this constant would need to move with it. The contract
+ * remains the enforcement; this only keeps the dApp from building a
+ * transaction the contract is guaranteed to reject.
+ */
+export const MAX_PRE_PEGIN_FUNDING_INPUTS = 20;
```
gbarkhatov, line 15: Nit. CLAUDE.md "Protocol Parameters" asks for a `TODO` with context when a contract value isn't read yet. The docblock explains the mirror but has none. A `TODO(#2402)` pointing at the `pinnedBuildLimits.ts` pattern would record that reading `maxFundingInputCount` from the params tuple is the intended end state.

gbarkhatov (dismissed):
The cap holds on every path I could find: fee estimate, Max, overlap check and the single build path. Nothing signs more than 20 inputs. No code merge blockers. Six points are inline, all about what the form tells the user, the error copy and the tests.

**Merge blocker (not code):** `1e11fcf71` is unsigned (`git log %G?` = `N`), and `main`'s ruleset requires the commit-authentication workflow, so `validate` fails. The two later commits are signed.

**Verified clean:**
- `capFundingUtxos` sorts largest-first with the same comparator as `selectUtxos.ts`, then takes 20. When a funding set of 20 or fewer inputs exists, greedy selection over that slice finds it.
- `useEstimatedBtcFee` uses the capped set for input count, total and `selectUtxosForPegin`, and every memo dependency was switched over.
- `preparePeginTransaction` has one caller, and it passes the capped set.
- Resume re-signs the registered transaction without selecting UTXOs again.
- Splitting `utxoCappedMaxSats` from `adjustedMaxDepositSats` keeps the supply-cap clamp, and a supply-cap hit never reads as a UTXO problem.
- God-mode UTXO fragments feed only estimation, behind `isGodModePanelEnabled`. The build takes `spendableUTXOs`.
- Inscription filtering runs before the cap on both paths.
- The contract rejects only inputs above the limit (`PeginLogic.sol` at `e66637c2`), so exactly 20 passes. It counts all declared inputs, and the SDK adds none beyond the selected UTXOs.
- A split deposit is one Pre-PegIn with N HTLC outputs, capped once, which matches the contract counting inputs per transaction.

**On a lower on-chain limit (governance moves it below 20):** the user signs the PegIn inputs and the PoP, and only then does registration's gas estimate revert. No gas is spent and no BTC is committed. Separately, the fee rate can still differ between estimate and build (`estimatedFeeRate` vs `mempoolFeeRate`); that predates this PR.

**Refuted:**
- *`TooManyHtlcOutputs` decodes fine because it is in the ABI.* True only on the vault's viem decoder path, which registration doesn't use. See the `errorMessages.ts` comment.
- *Capping before script validation can sign an invalid input.* The ordering is real, but every candidate UTXO comes from the wallet's one address, so they share a script. The worst case is blocking a valid deposit, never over-signing.
- *A zero or fractional fragment count.* Unreachable: the only setter clamps to an integer from 1 to 200, behind the god-mode gate.

services/vault/src/services/deposit/validations.ts
```diff
@@ -371,6 +373,31 @@ export function getDepositCtaState(params: DepositCtaParams): DepositCtaState {
       label: capBelowMinimumLabel(params.effectiveRemaining, params.minDeposit),
     };
   }
+  if (
```
gbarkhatov, line 376: Moving `exceedsCap` up here, above `maxBelowMinimum` (`:408`), breaks the mirror this function's comments promise (`:366-369`, `:407`). `useDepositValidation.validateAmount` still checks `maxBelowMinimum` first (`useDepositValidation.ts:91-99`) and remaining capacity last (`:108-111`).

Concrete case: a wallet whose fee-adjusted max is below the minimum, with an entered amount above `effectiveRemaining`. The button reads "exceeds cap" while the inline error reads "Minimum deposit is X". It doesn't need more than 20 UTXOs to happen. Both paths stay disabled, so nothing unsafe is reachable, but the two messages disagree.

Two options:

- Apply the same order in `validateAmount`, running `validateRemainingCapacity` before the `maxBelowMinimum` branch.
- Or keep `exceedsCap` where it was and move only the `fundingInputCapExceeded` branch between the supply-cap and balance branches.

services/vault/src/hooks/deposit/useEstimatedBtcFee.ts
```diff
@@ -55,17 +79,28 @@ export function useEstimatedBtcFee(
...
-      feeRate,
-    });
-  }, [utxos, feeRate, numOutputs, isLoading]);
+    if (isLoading || feeRate === 0) return null;
+    return maxDepositOver(cappedUtxos, numOutputs, feeRate);
+  }, [cappedUtxos, feeRate, numOutputs, isLoading]);
+
+  const uncappedMaxDeposit = useMemo(() => {
+    if (isLoading || feeRate === 0) return null;
+    return maxDepositOver(spendableUtxos, numOutputs, feeRate);
```
gbarkhatov, line 102: Low severity, and it fails safe. `maxDepositOver(spendableUtxos, …)` is `computeMaxDeposit` over *every* spendable input: total balance minus the fee for `utxos.length` inputs (`peginFeeMath.ts:170-178`).

Any input worth less than its own input fee pulls this bound down. At 50 sat/vB a 546-sat output costs about 2,900 sats of fee, and wallets with more than 20 UTXOs are the dust-heavy ones. So an amount the largest k > 20 UTXOs could fund can still sit above `uncappedMaxSats`. The user then sees "Insufficient balance" when "Consolidate your UTXOs" is the right advice. Nothing is over-signed; the notice is just suppressed.

If you want the notice to be accurate, take the best value over the sorted prefixes: the max over k of (sum of the k largest UTXOs − the base fee for k inputs). That is exactly what greedy selection could reach without the cap.

services/vault/src/hooks/deposit/__tests__/useDepositPageForm.test.tsx
```diff
@@ -1171,4 +1175,109 @@ describe("useDepositPageForm", () => {
...
+      mockOverCapWallet();
+      const { result } = renderHook(() => useDepositPageForm(), { wrapper });
+
+      await enterAmountWithSettledMax(result, "0.02");
+
+      expect(result.current.amountSats).toBe(2_000_000n);
+      expect(result.current.fundingInputCapExceeded).toBe(false);
+    });
+
+    it("does not flag 21 entries whose malformed scripts leave 19 spendable, so both maxes agree", async () => {
```
gbarkhatov, line 1257: Nit, per the test-philosophy rules. The name promises "21 entries whose malformed scripts leave 19 spendable", but `mockOverCapWallet` builds 21 identical UTXOs with `scriptPubKey: "0xabc"`, and `useEstimatedBtcFee` is mocked to return equal `maxDeposit` and `uncappedMaxDeposit`. What this test actually checks is "equal maxes → no flag"; script filtering is covered for real in `useEstimatedBtcFee.test.tsx`. Suggest renaming it to the behavior it asserts.

gbarkhatov (approved):
Follow-up review. Changes since my last review (`e685cb0e` → `293f5294`).

**Headline:** 5 of 6 prior threads are fixed and T5 is partly fixed. The squashed commit is signed, which clears the blocker from my last review. At `a264698f`, `verify` failed: the new `DepositForm.fundingInputCap.test.tsx` fixture was missing the `isLedgerVaultWallet` field that main now requires. The follow-up `293f5294` adds exactly that line, and CI is now green, including `verify`. No merge-blockers. There are three inline comments on the new material: a CTA/inline-error ordering mismatch, a low-severity max that understates dust-heavy wallets, and a test-name nit.

### Status of prior threads

| # | Finding | Status at `293f5294` |
|---|---|---|
| T1 | Upper bound was `btcBalance` (no reserves, different inscription population) | Fixed: both maxes subtract reserves and start from the same `withSpendableScripts` set |
| T2 | `TooMany*` entries in `errorMessages.ts` were unreachable | Fixed (dropped) |
| T3 | `fundingInputCapExceeded` sat below `maxBelowMinimum` | Fixed, now above it. The reorder creates a new mismatch; see the inline comment on `validations.ts`. |
| T4 | Agreement test re-declared the helper; `<= 20` couldn't fail; no hook-level test | Fixed: it imports the production helper, the value-floor assertion can fail, and hook tests were added |
| T5 | "20" hardcoded in copy; "combine" vs "Consolidate" | Partly fixed. The count is now interpolated from `MAX_PRE_PEGIN_FUNDING_INPUTS`, but `copy.ts:315` still says "Please combine your UTXOs" next to `:316` "Consolidate your UTXOs". The existing thread stands on the wording. |
| T6 | No `TODO` for reading the parameter from the contract | Fixed (`fundingInputCap.ts:12-13`, `TODO(#2402)`) |

### Notes

- **Where the cap comes from.** It is still the constant `MAX_PRE_PEGIN_FUNDING_INPUTS = 20` with the TODO, and no test pins it to on-chain `maxFundingInputCount`. The agreement test pins the helper against the SDK selector, and they do agree: `capFundingUtxos` applies the same decompile filter and value-descending sort as `selectUtxosForPegin`, and the capped set is what reaches `preparePegin`.
- **Broadcast-site fee check.** Another reviewer's point under CLAUDE.md §2 still stands: nothing at the broadcast site compares the selected input count or the fee against the estimate. Not repeated here.
- **Dev override.** `overrides/utxos.ts` can't be reached in production: the store returns null unless the God-mode flag is on, and that flag requires `import.meta.env.DEV`.

### Refuted

- *"The new script filter crashes in the browser."* `hasValidScript` uses `globalThis.Buffer`, the same pattern as `btcUtils.ts`, which the node polyfills provide in production.
- *"Capping changes what the build signs."* `capFundingUtxos` only drops UTXOs the SDK filter would drop anyway, in the same order, so the build receives the same prefix.

services/vault/src/services/deposit/validations.ts
jonybur, line 376: **exceedsCap was moved ahead of maxBelowMinimum in the CTA, but useDepositValidation.validateAmount still checks maxBelowMinimum first, so the button and inline error disagree.**

Besides adding the `fundingInputCapExceeded` branch, this change moves the `amountSats > effectiveRemaining` check from after `maxBelowMinimum` to before it. `useDepositValidation.validateAmount` was not reordered and still returns the max-below-minimum error before `validateRemainingCapacity`. Take a wallet whose fee-adjusted max is below the minimum while the remaining cap is above it, and type an amount above the remaining cap. The CTA now shows the exceeds-cap label while blur/submit shows "Minimum deposit is X", and lowering the amount flips the label again. The "Mirrored in useDepositValidation.validateAmount" comment at line 407 is now false. Keep only the consolidate branch above `maxBelowMinimum` and put `exceedsCap` back where it was, or apply the same order in `validateAmount`.

services/vault/src/hooks/deposit/__tests__/useEstimatedBtcFee.test.tsx
```diff
@@ -117,3 +117,40 @@ describe("useEstimatedBtcFee fee-rate override", () => {
...
+    value: (i + 1) * 100_000,
+    scriptPubKey: "00",
+  })) as unknown as MempoolUTXO[];
+
+  it("reports the whole wallet's max alongside the funding-input-capped one", () => {
+    const { result } = renderHook(() =>
+      useEstimatedBtcFee(AMOUNT, OVER_CAP_UTXOS, NUM_OUTPUTS),
+    );
+
+    expect(result.current.maxDeposit).toBe(22_999_000n);
```
jonybur, line 134: **The new capped/uncapped max tests use a computeMaxDeposit mock that ignores numInputs, so a wrong input count on the fee critical path would pass.**

The file mocks `computeMaxDeposit` as `totalBalance - feeRate*100`, ignoring `numInputs`. The expected values in the new tests (22,999,000 vs 23,099,000, and 20,899,000) differ only by the balance of the dropped UTXOs. If the capped path passed the uncapped input count, or the other way round, every assertion would still pass. This hook is on the fee-calculation critical path. The only real-SDK check covers the capped max, and nothing checks `uncappedMaxDeposit`, which decides whether the consolidate gate fires. Make the mock depend on `numInputs` (e.g. subtract a per-input amount) so the expected values pin 20 versus 21 inputs.

services/vault/src/__tests__/contracts/fundingInputCapAgreement.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    expect(maxDeposit).not.toBeNull();
+    expect(maxDeposit).toBeGreaterThan(0n);
+
+    const selection = selectUtxosForPegin(
+      cappedUtxos,
+      maxDeposit as bigint,
+      FEE_RATE,
+      NUM_OUTPUTS,
+    );
+    expect(selection.selectedUTXOs.every((u) => u.value >= 60_000)).toBe(true);
```
jonybur, line 72: **The only assertion on the real selection result checks every selected UTXO is >= 60,000 sats, which the capped fixture guarantees by construction.**

`cappedUtxos` is the 20 largest of the fixture, all between 60,000 and 250,000 sats. So `selection.selectedUTXOs.every((u) => u.value >= 60_000)` cannot fail whatever the selector does. Only the "max + 1 throws" check tests agreement between the estimator and the selector, and nothing compares a fee for the fixture as the fee-consistency rule asks. Assert that `selection.fee` equals `totalBalance - maxDeposit` (the selection should spend all 20 with no change), or at least that exactly 20 inputs are selected. A malformed-script entry in this real-SDK fixture would also pin the filter ordering.

services/vault/src/hooks/deposit/__tests__/useDepositPageForm.test.tsx
jonybur, line 1257: **Test named for malformed scripts leaving 19 spendable uses 21 identical valid-looking scripts and a mocked estimator, so no script filtering is exercised.**

This test claims 21 entries whose malformed scripts leave 19 spendable. But `mockOverCapWallet` gives all 21 UTXOs the same `scriptPubKey: "0xabc"`, and `useEstimatedBtcFee` is mocked to return equal `maxDeposit` and `uncappedMaxDeposit`. What it actually checks is that equal capped and uncapped maxes don't set the flag. A reader would assume the hook's script-filtering path is covered here when it isn't. Rename it to describe the equal-max behaviour, or drop it, since the filtering is covered in `useEstimatedBtcFee.test.tsx`.

services/vault/src/services/deposit/fundingInputCap.ts
```diff
@@ -0,0 +1,52 @@
...
+ * validity filter and the sort order match `selectUtxosForPegin`'s internal
+ * `validUTXOs`/`sortedUTXOs` exactly, so the estimator, the selector, and the
+ * build all agree on which UTXOs are in play — and a malformed high-value
+ * entry, which the selector would discard anyway, cannot occupy a capped slot.
+ * Does not mutate the input.
+ */
+export function capFundingUtxos<
+  T extends { value: number; scriptPubKey: string },
+>(utxos: readonly T[]): T[] {
+  return withSpendableScripts(utxos)
```
jonybur, line 49: **capFundingUtxos slices to 20 before selectUtxosForPegin's duplicate-outpoint assertion, so duplicates outside the top 20 are silently dropped.**

`selectUtxosForPegin` calls `assertNoDuplicateUtxos` on its input and throws. Its own comment calls duplicates a sign of a buggy or compromised UTXO source. `capFundingUtxos` now sorts and slices before that call on the build, estimate and overlap paths. A duplicate outpoint below the 20 largest, or a pair split across positions 20 and 21, never reaches the check. Nothing extra gets signed, but a corrupted UTXO feed that used to stop the flow now goes unnoticed. Check the full list for duplicate outpoints before capping, e.g. in `withSpendableScripts`.

jonybur (commented):
Comments inline, plus:

- `services/vault/src/hooks/deposit/useDepositPageForm.ts:701` - **The clamp effect lowers any amount above the UTXO-capped max whenever the max changes, which silently clears fundingInputCapExceeded and the consolidate notice.**
  
  This effect runs whenever `adjustedMaxDepositSats` changes and lowers any amount above it to the max. With this PR that value is the 20-UTXO-capped max, and `fundingInputCapExceeded` is only true when the amount is above it. So every time the max moves, the flag clears and the notice and "Consolidate your UTXOs to proceed" label disappear. The max moves when fee rates refetch (every 60s in `useNetworkFees`), when the claim value loads after a provider is picked, or when split toggles. The user's typed amount is replaced with a smaller one they never entered. The new hook tests wait for the max to settle before typing, so they miss this. Skip the clamp when the amount is at or below `uncappedMaxSats` (or only clamp when Max is pinned), and add a test that types first and then lets the max change.

PR #2508, opened 2026-09-13
fix(vault): verify depositor wallets before resume signing

jonybur:
## Why

Resume trusted the connected wallets. A wallet that was not the registered depositor still reached the signing request, and a mismatch often showed a generic or "broadcast failed" error.

## What

- **Broadcast resume** checks the live Ethereum account and the Bitcoin wallet key against the registry contract record. It checks again before the broadcast service and at signing. The PSBT uses the contract key.
- **Payout signing** checks the live Bitcoin wallet key against the contract key before VP authentication, any device approval, and any signature. It signs no Ethereum transaction, so it does not check the Ethereum account for software wallets.
- A wrong wallet now shows "Wrong wallet connected" or "Wrong Bitcoin wallet connected" and asks the user to connect the depositor's wallet.

## How

- New typed `DepositorBtcKeyMismatchError` next to `DepositorWalletMismatchError`. Both error mappers branch on it, and the broadcast service passes both through unwrapped.
- The resume hook runs the identity check after the wallet liveness probe.
- `getVaultFromChain` returns the registered Bitcoin key. `prepareSigningContext`, `signAndSubmitPayouts` and the terms rebuild compare against it.

Tested on the branch merged with main: lint 0 errors, type check passes, full vault suite 3,952 passed and 2 skipped. No live wallet or hardware run.

Tracks #2228.


services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -484,55 +491,31 @@ export function useVaultActions(): UseVaultActionsReturn {
...
+      if (signal.aborted) return;
+      await assertDepositorWallet();
+      await broadcastPrePeginTransaction({
+        unsignedTxHex,
+        btcWalletProvider: {
+          ...forwardDeriveContextHash(btcWalletProvider),
+          ...forwardDepositApproval(btcWalletProvider),
+          signPsbt: async (psbtHex: string) => {
+            // Input resolution can wait on the network. Check again at signing.
+            await assertDepositorWallet();
```
gbarkhatov, line 533: Should fix. A mismatch caught here reaches the user as "broadcast failed". This check runs inside `broadcastPrePeginTransaction`, whose catch rethrows everything except `DepositTermsRejected` as `Error("Failed to broadcast Pre-PegIn transaction: …", { cause })` (`vaultPeginBroadcastService.ts:369-378`). `isDepositorWalletMismatchError` is a plain `instanceof` (`depositorWalletMismatch.ts:35`) and never walks `cause`, so the mapper's typed branch (`depositErrors.ts:306`) misses. The error then lands on `msg.includes("broadcast")` (`:430`) → `broadcastFailed`. Nothing was broadcast, and retrying won't help: the user needs to reconnect the depositor wallet. A Bitcoin key mismatch caught here goes the same way.

The tests can't see it because the broadcast service is mocked, so nothing gets wrapped, and they only assert `broadcastError` is null or non-null.

Fix: either let the mismatch errors through unwrapped (like `isDepositTermsRejectedError`), or make the classifier walk `cause`. Then assert the specific copy.

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -367,6 +369,35 @@ export function useVaultActions(): UseVaultActionsReturn {
...
+            expectedDepositor: finalBasicInfo.depositor,
+            connectedDepositor,
+          });
+        }
+        return validateWalletPubkey(
+          walletPubkey,
+          stripHexPrefix(finalBasicInfo.depositorBtcPubKey),
+        ).depositorPubkey;
+      };
+      const depositorBtcPubkey = await assertDepositorWallet();
```
gbarkhatov: Two things at this call:

1. **Ordering.** `assertDepositorWallet()` calls `btcWalletProvider.getPublicKeyHex()` before `verifyBtcWalletLiveness` at `:405`, and the base hook never called `getPublicKeyHex` here. The probe exists to turn a locked or unresponsive wallet into `COPY.wallet.liveness.*`. A wallet that locked during the finality wait may now fail in `getPublicKeyHex` first and get the generic fallback. Nothing between `:399` and `:405` uses the key, so moving this below the probe is free. How each wallet's `getPublicKeyHex` behaves while locked is unverified.

2. **Bitcoin mismatch copy.** `validateWalletPubkey` (`:394-397`) throws a plain `Error`. That is either "Wallet public key does not match vault depositor. Expected: …, Got: …" or, for an empty contract key, the developer message "requires expectedDepositorPubkey … avoid a self-comparison". Neither matches a mapper bucket, so the user gets the default title with raw SDK text. The Ethereum case has `wrongDepositorWallet`, but that copy says "Ethereum account". The Bitcoin case needs its own typed error and copy entry.

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -484,55 +491,31 @@ export function useVaultActions(): UseVaultActionsReturn {
...
-          unsignedTxHex,
-          btcWalletProvider: {
-            signPsbt: (psbtHex: string) => btcWalletProvider.signPsbt(psbtHex),
-          },
-          depositorBtcPubkey,
-          expectedUtxos,
-        });
       }
+      if (signal.aborted) return;
+      await assertDepositorWallet();
```
gbarkhatov, line 525: Untested. The "rejects %s mismatch" cases set the bad state before render, so they fail at `:399`. The "checks %s state at signing" cases change state inside the mocked broadcast, so only `:511` catches them. Deleting this line therefore leaves the suite green.

It matters most on Ledger: `ensurePrePeginTermsApproval` runs the device ceremony inside the broadcast service, before `signPsbt`, and `deriveContextHash`/`approveDepositTerms` are forwarded unchecked. That makes this the last identity check before the user approves on the device. A Ledger case that switches wallets inside a mocked `rebuildDepositTerms` and asserts `approveDepositTerms` is never called would pin it.

services/vault/src/hooks/deposit/useVaultActions.ts
```diff
@@ -367,6 +369,35 @@ export function useVaultActions(): UseVaultActionsReturn {
...
 
+      // Bind both wallets to the contract record. Indexer identity is untrusted.
+      const assertDepositorWallet = async () => {
+        signal.throwIfAborted();
+        const walletPubkey = await btcWalletProvider.getPublicKeyHex();
+        signal.throwIfAborted();
+        const connectedDepositor = getAccount(getSharedWagmiConfig()).address;
+        if (!connectedDepositor) {
+          throw new Error(
+            COPY.deposit.emergencyWithdraw.errors.ethWalletNotConnected,
```
gbarkhatov: Nit. Resume throws `COPY.deposit.emergencyWithdraw.errors.ethWalletNotConnected`. It maps correctly today, but rewording that string for emergency withdraw would silently change this flow too. A `deposit.errors` entry of its own avoids the coupling.

services/vault/src/hooks/deposit/__tests__/useVaultActions.test.ts
```diff
@@ -1374,6 +1319,78 @@ describe("useVaultActions — handleBroadcast intent (Ledger) resume branch", ()
     mockFetchVaultById.mockResolvedValue(baseVault as never);
   });
 
+  describe.each(["software", "intent"])("%s wallet identity", (wallet) => {
```
gbarkhatov: Nit, test philosophy. The nested `describe.each` × `it.each` expands to 18 cases with conditional spreads keyed on case names, `state === "matching" ? 1 : 0` expectations and `if (state !== "cancelled")`. It also mixes the success case into a failure table, and software-wallet cases sit inside the "intent (Ledger) resume branch" describe. CLAUDE.md asks for one behaviour per test, readable top to bottom ("three similar test functions are better than one parameterized abstraction"). The PR also folds existing EXPIRED/VERIFIED and version-drift tests into `it.each` the same way, and the renamed test at `:628-661` no longer says it asserts the version check is skipped.

gbarkhatov (dismissed):
No merge blockers. The new wallet checks fail closed on every path I traced, both identities come from the contract record, and a signature from any other key is rejected before broadcast. The inline points are about how refusals reach the user, one ordering issue, and a test gap. CI is fully green.

**Verified clean:**
- **Contract source.** Both identities come from `finalBasicInfo`, the `getVaultBasicInfo` read after the finality depth. A zero depositor or `createdAt` polls, then throws `PeginRegistrationMissingError`, with no silent skip.
- **Ethereum binding.** Lowercased compare against both the contract depositor and the action's address, which handles EIP-55. Ledger resume already required this (`rebuildDepositTerms.ts:216`).
- **Bitcoin binding.** The wallet key is reduced to x-only and compared to the contract key, case-insensitive. The PSBT and `rebuildDepositTerms` now use the contract key instead of the indexer's, and the test proves it because the indexer fixture differs from `DEPOSITOR_BTC_KEY`.
- **Shared broadcast call.** It keeps both PENDING gates, the UTXO check, the version and participant-key checks and the abort return. Software wallets are unaffected by the forwarded approval methods: `ensurePrePeginTermsApproval` returns early without an approve function or terms.
- **Returned signatures.** They are verified against the requested PSBT's prevouts before `pushTx`, so a switch inside the wallet's own signing popup cannot broadcast a foreign-key signature.
- **No conflict with #2482,** which only adds an early `requireBtcWallet()` return in `handleBroadcast`.

**Refuted:**
- *§9 is broken because `assertOnChainBtcPubkey` isn't used.* No `OnChainBtcPubkey` is minted and nothing is reimplemented; the contract key is accepted only if it equals the wallet's x-only key.
- *Aborting after the Ledger device approves is a defect.* It is intentional and fails closed; the "cancelled" case pins it, and a retry can reuse the held approval.
- *The added `throwIfAborted` calls leave a stale error after dismissal.* The abort fires only on real unmount, and the catch skips `setBroadcastError` once unmounted.

services/vault/src/hooks/deposit/useVaultActions.ts
jrwbabylonlab: This only covers the broadcast resume. The payout-signing resume (usePayoutSigningState.ts:403 → prepareSigningContext) still signs with whatever BTC key is connected and never compares it to the on-chain depositor key on the FE; it relies on the VP rejecting the auth anchor. Either apply the same check there or narrow the title/description to the Pre-PegIn broadcast path.

jrwbabylonlab (dismissed):

services/vault/src/hooks/deposit/useVaultActions.ts
jonybur: Fixed. The broadcast service now passes both wallet-mismatch errors through unwrapped, the same way it passes `DepositTermsRejected`. Nothing walks `cause` for this.

- An Ethereum mismatch maps to `wrongDepositorWallet`.
- A Bitcoin mismatch maps to the new `wrongDepositorBtcWallet`.
- Two service tests throw each error from `signPsbt` and check that the same error object comes out and `pushTx` is not called.
- The hook tests now assert the exact copy, not only a non-null error.

services/vault/src/hooks/deposit/useVaultActions.ts
jonybur: Both fixed.

1. **Ordering.** The identity check now runs after `verifyBtcWalletLiveness`. A new test uses a wallet the probe round-trips, makes the probe fail, and checks that the user gets the liveness copy and `getPublicKeyHex` is never called. On how wallets behave while locked: most providers return a key cached at connect, so the old order failed only when that cache was gone, for example Ledger Vault after its session ended.
2. **Bitcoin copy.** A wrong key now throws `DepositorBtcKeyMismatchError`, which both mappers send to new `wrongDepositorBtcWallet` copy. An empty contract key now fails before any wallet call with its own copy, `depositorBtcKeyMissing`, instead of the SDK developer message. The PSBT still uses the contract key.

services/vault/src/hooks/deposit/useVaultActions.ts
jonybur: Pinned. The Ledger describe has a new test: "does not ask the device to approve terms when the ETH account switches during the terms rebuild".

- The mocked `rebuildDepositTerms` switches the live Ethereum account.
- The mocked broadcast calls the forwarded `approveDepositTerms` and then `signPsbt`, in the real service's order.
- The test asserts that `approveDepositTerms` and the broadcast are never called, and that the error is `DepositorWalletMismatchError` with `wrongDepositorWallet` copy.

I deleted the pre-broadcast `assertDepositorWallet()` locally, and this test failed. I also mutation-checked the call inside `signPsbt`, the Bitcoin comparison, and the probe order. Each change made at least one test fail.

services/vault/src/hooks/deposit/useVaultActions.ts
jonybur: Fixed. Resume now throws its own `COPY.deposit.errors.ethWalletNotConnected`. A mapper test pins that it still maps to `walletNotConnected`. The emergency-withdraw string is unchanged and still used there.

services/vault/src/hooks/deposit/__tests__/useVaultActions.test.ts
jonybur: Fixed. The file has no `it.each` or `describe.each` left.

- The EXPIRED and VERIFIED tests and the three version-drift tests are back exactly as they are on main.
- The test at `:628` has its original name again, which says the version check is skipped. The two indexer-key cases are now two separate tests.
- The software-wallet identity cases moved to a new "handleBroadcast depositor wallet binding" describe, one behaviour per test, with the success case on its own.
- The Ledger describe keeps one intent-specific refusal test and the new device-approval test. I dropped intent copies of cases that fail at the same wallet-agnostic check, so 18 table cases became 11 tests.

services/vault/src/hooks/deposit/useVaultActions.ts
jonybur: Applied the check on the payout path rather than narrowing the PR.

- `prepareSigningContext` now reads the registered Bitcoin key with the vault record. It refuses a caller key that differs, before any other read. Both payout callers go through it: the resume modal and the inline deposit flow.
- `signAndSubmitPayouts` then reads the live wallet key and compares it as well. This happens before VP authentication (`deriveContextHash`), `approveDepositTerms`, and any signature, so a wallet switch after the modal read the key is also refused.
- On resume with an approval wallet, the terms rebuild runs first. It now refuses a wrong Bitcoin key with the same typed error before any sibling read, so the user sees "Wrong Bitcoin wallet connected" instead of a generic error.
- The error is not terminal, so Retry works once the user connects the right wallet.

For software wallets this path checks the Bitcoin key only, because payout signing needs no Ethereum signature. The PR description now says so.

jonybur:
Addressed all review threads. Each thread has a reply with the change.

- Merged current main, including #2482. There were no conflicts.
- On the merged tree: lint 0 errors, type check passes, full vault suite 3,952 passed and 2 skipped.
- The push dismisses the earlier approvals, so this needs a fresh review.

gbarkhatov (approved):

PR #2511, opened 2026-09-14
fix(core-ui): prevent tooltip hover in visual captures

jonybur:
## Outcome

Prevent false visual changes when the pointer opens a Hint tooltip during capture. The fix is open for review.

## Change

Use a named coordinate to move the pointer outside the page before each Storybook story loads. A pointer over the default Hint text opens a tooltip above the viewport, where only its lower edge and shadow remain visible.

## Safety

The change only affects the capture test. It preserves component styles and image comparison thresholds. CI uses the same capture test for both sides.

## Verification

- Reproduced the clipped tooltip by placing the pointer over the Hint text before navigation.
- All 10 repeated captures with the pointer outside the page matched exactly.
- All 329 Storybook captures passed locally.
- After the constant change: 12 Hint captures and 113 unit tests passed. All six Hint images match the earlier captures.
- Core UI lint passed with 213 existing warnings and no errors.
- `git diff --check` passed.

## Links

- No linked issue.


jonybur:
@greptile-apps

gbarkhatov (approved):

Issue #2437, opened 2026-09-08
Multi-hub: the same token from two Hubs must be two markets in the UI


gbarkhatov:
## Outcome

A depositor can tell the two USDC markets apart, pick the one they mean, and see the rate, the liquidity and the borrow limit that actually apply to it.

The Confluence page "Multi-Hub Configuration" adds a second Aave v4 Hub, and the Babylon Core Lending Spoke lists reserves from both. The same underlying token then appears twice on one spoke as two unrelated reserves, with independent APR, liquidity, caps and user debt.

Design: Figma "Concept B — Multi-hub asset borrowing", https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=13093-11274 (the "Borrow flow" and "Repay Flow" rows).

## Delivery plan

Three PRs, in order, plus one unrelated small fix.

| PR | Scope | Why it is its own PR |
|---|---|---|
| **1. Prove the hub on-chain** — https://github.com/babylonlabs-io/babylon-toolkit/pull/2512 (merged) | Scope 1 | Changes the trust model and needs two reviewers. The hub name the design asks users to choose between is derived from this address. It touches no UI, so it can merge while PR 2 is built. |
| **2. The Figma flow** — https://github.com/babylonlabs-io/babylon-toolkit/pull/2515 (merged) | Scopes 2, 3, 4, 8, 9, 10; landing APR card | This has to be one PR: inserting the Select hub step breaks the real-wallet E2E driver, so the UI and the driver must land together. It came to 88 files (+4,061 / −1,113), about half of it tests, including the E2E `multi-hub` and `repay-all` actions. The replay fixture re-record moved to a follow-up (see Follow-ups). |
| **3. Hub correctness not visible in Figma** — https://github.com/babylonlabs-io/babylon-toolkit/pull/2551 (in review) | Scopes 5, 6, 7, 11 | Logic the design doesn't show: caps, hub state, revert decoding, activity edge cases. Needed before mainnet, not for the devnet flow. It follows PR 2 because both touch the borrow form. Scope 12 moved to https://github.com/babylonlabs-io/babylon-toolkit/issues/2513. |
| Separate | `TOKEN_REGISTRY` non-devnet entries are Optimism addresses | Unrelated to multi-hub, but it blocks mainnet. |

No feature flag is needed. The Select hub step is skipped when a token has a single visible reserve, so single-hub networks look unchanged.

[...]

PR #2512, opened 2026-09-14
fix(vault): prove Aave reserve hub and asset identity on-chain at config load

gbarkhatov:
## What

`fetchAaveAppConfig` now proves every reserve the indexer returns against the Core Spoke before the app uses it. One hard-fail multicall reads `getReserve` for every indexed reserve id from the Core Spoke that is already resolved on-chain from the env-pinned adapter, and the config load throws if the indexer lists a reserve id more than once or disagrees with the chain on any field fixed at reserve listing:

- `underlying`, and `token.address` (must equal the underlying)
- `hub`
- `assetId`
- `reserve.decimals` and `token.decimals`

Addresses compare case-insensitively. Once proven equal, the indexer's values are what callers receive, so every consumer of `AaveReserveConfig` gets an on-chain-proven hub and asset id without having to remember to check.

In the diff:

- `services/vault/src/applications/aave/services/fetchConfig.ts`: the proof, run before any reserve is derived into `vbtcReserve`, `borrowableReserves` or `allBorrowReserves`.
- `services/vault/src/applications/aave/clients/spoke.ts`: `getReservesBatch`, a hard-fail `getReserve` multicall using the SDK's exported `AaveSpokeABI`.
- `services/vault/src/applications/aave/services/__tests__/fetchConfig.test.ts`: tests for the batch read, a mismatch on each proven field (including on the vBTC reserve), a reserve listed twice, case-only address differences, a failed read and a short result.
- `services/vault/e2e/fixtures/replay/supplements.ts` and `chain.ts`: the replay recording only held `getReserve(3)`, so the new read would have left the visual capture on the error screen. Supplements now rebuild each `getReserve(id)` answer from the recorded GetAaveAppConfig row, on the spoke taken from the recording's own `BTC_VAULT_CORE_SPOKE()` answer. Recorded answers still take precedence, and `services/vault/e2e/fixtures/__tests__/replay.test.ts` checks a rebuilt answer against its recorded row.
- `SECURITY.md`: a threat-table row and a paragraph in the Indexer section.

## Why

`reserve.hub` and `reserve.assetId` came from the indexer unchecked. The hub is the contract every rate and liquidity read in `clients/aaveHub.ts` targets, and those reads cap the borrow Max. An indexer that kept `reserveId -> underlying` truthful (so the existing per-transaction `assertReserveMatchesOnChain` passes) but swapped the hub could show false rates and liquidity in front of an irreversible borrow. Under multi-hub, the hub also becomes the thing a user chooses between. This is Scope 1 of https://github.com/babylonlabs-io/babylon-toolkit/issues/2437 and the base for the multi-hub UI.

## Approaches

- **Verify once at config load, not per row at render.** One batched read covers every reserve, and every consumer receives proven values. Per-row checks would multiply RPC calls and rely on each surface remembering to verify.
- **Fail closed on any mismatch**, matching the existing adapter-address and vBTC-reserve-id checks in the same function. The compared fields are fixed when a reserve is listed, so a mismatch is never indexer lag.
- **Compare only immutable fields.** Paused, frozen, borrowable and the risk parameters change legitimately, and the indexer trails them by its refresh interval; fail-closed comparison would break the app on ordinary governance actions.
- **Batch read in the app, not the SDK.** Adding it to the SDK would change the SDK's public surface and require regenerating its API docs; the app already runs its own multicalls in `clients/aaveHub.ts`.
- **Replay supplements rather than a re-recording.** Re-recording needs a live run; the supplement values are recovered from the recording itself, and the app's own check would reject any other value.

## Not in this PR

- Hub labels, the Select hub step and any multi-hub display (PR 2 in https://github.com/babylonlabs-io/babylon-toolkit/issues/2437).
- Draw caps, hub active/halted handling and hub revert decoding (PR 3).
- Proving mutable reserve fields (flags, collateral risk, dynamic config key, collateral factor) or token symbol and name.
- Detecting reserves that exist on-chain but are missing from the indexer.
- Changes to the per-transaction `assertReserveMatchesOnChain`.

## Pre-review

<details>
<summary>Pre-review: 11 findings · 8 fixed · 0 moot · 0 follow-up · 3 declined · 0 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | The on-chain proof accepts a reserve id the indexer lists twice | normal | fixed |
| 2 | SECURITY.md threat row lists a 'hub name' impact the app does not show yet | normal | fixed |
| 3 | SECURITY.md Indexer paragraph names only some of the fields still indexer-sourced | normal | fixed |
| 4 | Every mismatch test targets the USDC reserve; none covers the first (vBTC) reserve | normal | fixed |
| 5 | The new getReserve replay supplements have no unit test | normal | fixed |
| 6 | Supplement row guard rejects undefined but not null fields | normal | fixed |
| 7 | RecordedAnswer is exported but used only inside supplements.ts | normal | fixed |
| 8 | A negative indexer reserve id surfaces as a Core Spoke read failure | normal | declined: fails closed with the viem error kept as cause; ids come from an unsigned GraphQL field |
| 9 | Indexer numeric fields are not type-checked at runtime before strict comparison | normal | declined: the indexer and recording send JSON numbers, and a string still fails closed |
| 10 | The 'returned no reserve' guard cannot fire against real viem | normal | declined: kept as an explicit invariant so a short result gives an actionable error |
| 11 | The JSDoc for assertReservesMatchOnChain does not mention the duplicate-id rejection | normal | fixed |

</details>


services/vault/src/applications/aave/services/fetchConfig.ts
```diff
@@ -191,6 +192,123 @@ function mapReserveConfig(raw: GraphQLReserveItem): AaveReserveConfig | null {
...
 }
 
+function assertAddressMatches(
+  reserveId: bigint,
+  field: string,
+  indexed: Address,
+  onChain: Address,
+): void {
+  if (indexed.toLowerCase() !== onChain.toLowerCase()) {
+    throw new Error(
```
jonybur: **New reserve identity mismatches throw plain Error instead of ReserveMismatchError, so a detected spoof is retried and shown as a transient 'try again' failure.**

`assertAddressMatches`, `assertNumberMatches` and the duplicate-id check throw a plain `Error` when the indexer disagrees with the Core Spoke. The vault already has `ReserveMismatchError`, and `isIntegrityFailure` in `verifyReserveIdentity.ts` exists so the app does not treat such a result as a transient fault. As written, the default `shouldRetry` in `config/queryClient.ts` retries the config query up to three more times, repeating the GraphQL request, the adapter reads and the `getReserve` multicall. `AaveConfigContext` then shows "Something went wrong / Please try again in a moment" with a Retry button. Please throw `ReserveMismatchError` for the field and duplicate mismatches, keep the wrapped read failure as a plain `Error` so transport faults stay retryable, and give the config query a `retry` predicate that stops on `isIntegrityFailure`.

jonybur (commented):
Comments inline.

services/vault/src/applications/aave/clients/spoke.ts
```diff
@@ -106,6 +107,31 @@ export async function getReserve(spokeAddress: Address, reserveId: bigint) {
...
+): Promise<AaveSpokeReserve[]> {
+  const publicClient = ethClient.getPublicClient();
+  const results = await publicClient.multicall({
+    contracts: reserveIds.map((reserveId) => ({
+      address: spokeAddress,
+      abi: AaveSpokeABI,
+      functionName: "getReserve",
+      args: [reserveId],
+    })),
+    allowFailure: false,
```
jrwbabylonlab: just checking, they don't have batch method out of the box?

jrwbabylonlab (dismissed):

services/vault/src/applications/aave/services/fetchConfig.ts
gbarkhatov: Fixed in 2160e8637. Field, duplicate and vBTC reserve-id mismatches in `fetchAaveAppConfig` now throw `ReserveMismatchError`; the vBTC check had the same issue before this PR. Read failures stay a plain `Error`. `AaveConfigProvider` skips retry on `isIntegrityFailure`, and every other error goes through the default `shouldRetry` (now exported from `config/queryClient.ts`), so a 451 is still not retried. A mismatch cannot be RPC lag: `getReserve` reverts `ReserveNotListed` for an unlisted id, which surfaces as a read failure and is retried. The panel still says "try again"; integrity-specific copy for a config-level mismatch is left for a follow-up.

services/vault/src/applications/aave/clients/spoke.ts
gbarkhatov: Right, the SDK only had a single-reserve `getReserve`; batch reads existed for positions and debts only. Added `getReserves` to the SDK next to `getUserTotalDebts` in 2160e8637, the same hard-fail multicall, and the vault `getReservesBatch` is now a thin wrapper like the rest of `clients/spoke.ts`. API docs regenerated.

jrwbabylonlab (approved):

Issue #2514, opened 2026-09-15
fix(vault): back-to-back wallet transactions fail with "nonce too low" because the wallet's node lags the receipt


gbarkhatov:
## Summary

When the vault app sends two Ethereum transactions in a row from the same account, the second one intermittently fails before it is broadcast with `nonce too low`. The wallet (MetaMask) assigns the second transaction the nonce the first one just used. The app waits for the first transaction's receipt on its own RPC, but the wallet's node has not caught up yet, so the wallet hands out a stale nonce.

Nothing is broadcast and no gas is spent, but the user sees a raw error ("Repay failed: ... Nonce provided for the transaction is lower than the current nonce of the account") in the middle of a multi-step flow, after an approval they already paid for.

## Evidence

**Manual, devnet (Sepolia), MetaMask.** Repay all on USDC (Core Hub, reserve 4). The ERC-20 approve mined with nonce 766 and allowance 1.005004 was confirmed on-chain. The repay that followed was signed with nonce 766 and rejected:

```
repay to Aave Core position failed: Nonce provided for the transaction is lower than the current nonce of the account.
Details: RPC 0xaa36a7 Infura eth_sendRawTransaction: nonce too low: next nonce 767, tx nonce 766
```

Retrying from the UI a moment later succeeded.

**E2E, devnet, MetaMask** (`e2e/real` multi-hub run with `--all-reserves --borrow-usd=1`). Six borrows and four repays passed; then:

- `12:27:38.9` repay on USDT (Core Hub, reserve 5) succeeds, nonce 787
- `12:27:41.1` next leg starts: repay on WBTC (Core Hub, reserve 6)
- `12:27:46.2` the approve for that repay is confirmed in MetaMask, signed with nonce 787
- `12:27:48.5` it is rejected:

```
Repay failed: approve ERC20 failed: ... RPC 0xaa36a7 Infura eth_sendRawTransaction: nonce too low: next nonce 788, tx nonce 787
Contract Call: address: 0x504579d0424b7b7cb4b17e16626f6a2f67bca054 function: approve(address spender, uint256 amount)
```

The same approve simulates fine, so this is not a token quirk. Frequency so far: 1 in about 10 back-to-back transactions in the E2E run, plus 1 in manual testing. Within one repay, the approve and the repay were usually about 15 s apart and passed. The failures had shorter gaps: the repay right after its approve in the manual case, and the approve about 7 s after the previous leg's repay in the E2E case.

## Root cause

[...]

Issue #2513, opened 2026-09-15
feat(vault): surface Aave's one-borrowable-asset-per-position limit


jrwbabylonlab:
## Summary

Aave has decided that at launch a position may hold debt in **one borrowable asset at a time** (USDG *or* USDC *or* USDT), while the Spoke keeps listing all three. Borrowing more of the same asset stays allowed; switching asset requires repaying the current loan to zero first. Decided in Slack on 2026-09-11; leadership asked for light, web-only changes for the phase-1 launch that make the limitation clear to users, with proper multi-asset support in a later phase.

## How it is enforced

The rule already exists in Aave v4: `Spoke.MAX_USER_RESERVES_LIMIT` (constructor immutable). `Spoke.borrow` reverts `MaximumUserReservesExceeded()` when the reserve is not already borrowed and `borrowCount >= limit` (`lib/aave-v4/src/spoke/Spoke.sol:284-294`). Repaying to zero clears the borrowing bit.

- No Babylon adapter change needed; `AaveAdapter.borrowFromCorePosition` and the position proxy add no debt-asset check and don't need one.
- The launch Spoke must be deployed with `MAX_USER_RESERVES_LIMIT=1` (`script/deployment/applications/aave/02_SpokeInfrastructure.s.sol:48-55`). It is immutable, there is no setter. Today's Sepolia Spokes are at 4 and 65535, so nothing enforces it yet.
- btc-vault, vault-provider-proxy and the indexer carry no debt-asset information. No change there.

## Frontend scope

The app already has everything it needs: one position per user, `position.debtPositions` (per-reserve debt map) and `accountData.borrowCount` from `useAaveUserPosition`.

- [ ] SDK: read `MAX_USER_RESERVES_LIMIT()` from the Core Spoke (`packages/babylon-ts-sdk/src/tbv/integrations/aave/clients/spoke.ts`). Do not hard-code 1.
- [ ] Rule: reserve R is borrow-blocked iff `debtPositions.size >= limit && !debtPositions.has(R)`. Treat a loading position as blocked, not as debt-free.
- [ ] Asset picker (`AssetSelectionPanel`) and in-form switcher (`AssetPill` / `AssetListItem`): disabled state + reason for blocked reserves. Rows have no disabled prop today.
[...]

PR #2515, opened 2026-09-15
feat(vault): borrow and repay multi hub

gbarkhatov:
## What

One Aave v4 Core Spoke now lists reserves from more than one hub, so the same token (for example USDC) is several markets, each with its own rate, liquidity and debt. This change lets the user see and choose the hub everywhere a loan is picked, opened or shown, following the multi-hub Figma flow.

- **Borrow:** Select asset shows one card per token. Select hub then lists that token's reserves with Borrow APR, compact USD liquidity and a Market Info link. The borrow form gains a Hub row. Select hub is skipped when a token has exactly one visible (not frozen, not paused) borrowable reserve.
- **Back arrow:** Select hub goes back to Select asset, and the borrow form shows a back arrow when it was reached through the pickers (the URL carries the token's underlying address and it matches the open reserve); every other entry keeps the close button. Picker steps navigate with `replace`.
- **Repay:** the repay picker lists each debt by reserve id with its hub and amount, and the repay form shows Hub as its first row.
- **Loans:** rows gain a Hub column and stay grouped by token; success and validation messages name the hub.
- **Hub labels** come from a local registry (`services/vault/src/services/aave/hubRegistry.ts`) keyed by the hub address that config load already proves on-chain, never from the indexer. An unregistered hub is shown by its short address with a warning hint and never blocks an action, so existing debt on it stays repayable.
- **Markets:** the market page moves to `/markets/:reserveId`; a legacy symbol link shows the outdated-link state. The markets table gains a Hub column, and the page header and chart labels name the hub.
- **Elsewhere:** the token dropdown lists token and hub entries; the landing card shows the lowest APR across a token's hubs; USDG joins the token registry; activity rows show the hub in their caption, and a past borrow or repay on a reserve that has since been frozen or paused keeps its token, amount and hub.
- **Every flow resolves to a reserve id before a form opens.** The `asset` query parameter is validated as an address and only filters the hub list and decides whether the back arrow shows.
- **E2E real-wallet driver** (`services/vault/e2e/real`): reserve resolution is hub-aware (`--borrow-hub`, `--repay-hub`), rows are selected by testids keyed on reserve id or underlying, and two actions are added: `multi-hub` (borrow from each reserve, check a Loans row per reserve names its hub, then repay each in full, with `--all-reserves` and `--borrow-usd`) and `repay-all` (clear every debt on every hub). Each leg checks on-chain that the debt moved on that reserve, and a `multi-hub` repay leg only counts once the reserve is back at its pre-run debt. The CLI refuses to guess on these real-value runs: a non-interactive run needs a token whenever more than one reserve or loan could match, a flag the chosen action would ignore or a value flag passed without a value stops the run, and a failed `multi-hub` run names the reserves still owed and points at `--action=repay-all`. Until https://github.com/babylonlabs-io/babylon-toolkit/issues/2514 is fixed, a borrow or repay submit the wallet rejects with "nonce too low" is resubmitted once (nothing was broadcast); that retry is removed with the fix.
- **Devnet E2E run:** `multi-hub --all-reserves --borrow-usd=1` borrowed about $1 from all six borrowable reserves across both hubs, confirmed each debt landed on its own reserve and that each Loans row names its hub, and repaid the first reserve. The next repay was rejected by the wallet with "nonce too low" (#2514), and `repay-all` then repaid the remaining five debts, ending with no outstanding debt on any hub.

## Why

Before this change devnet showed two identical USDC rows with no hub, `/markets/usdc` resolved to nothing, and the E2E driver clicked the first USDC it found. A user could not tell which market they were borrowing from or repaying. This is PR 2 of https://github.com/babylonlabs-io/babylon-toolkit/issues/2437; PR 1 (https://github.com/babylonlabs-io/babylon-toolkit/pull/2512) proved each reserve's hub and asset id on-chain at config load, which this change builds on.

## Approaches

- **Route by reserve id, not by token and hub.** Every picker resolves to a reserve id before the form opens, so the form, the market page and the E2E driver never guess between two reserves of one token.
- **Hub labels from a compile-time registry, not the indexer.** A hub contract has no `name()`, and an indexer-supplied label could make two different markets read the same. Labels must be unique within a deployment.
- **Unknown hub: warn, never block.** Blocking would strand existing debt on a hub the build doesn't know.
- **Back arrow derived from the URL, not from history.** Picker steps use `replace`, so browser Back and deep links behave as before, and a form opened from a Loans row, the market page or the token dropdown keeps its close button.
- **App copy conventions over the Figma text** (sentence case, no label colons, the hub named in success and validation messages).
- **No feature flag:** a single-hub deployment skips Select hub automatically.

## Not in this PR

- The Liquidation preview accordion.
- Draw caps, halted or inactive hubs, coupling between hubs, hub revert decoding and liquidation activity showing every repay (PR 3 in https://github.com/babylonlabs-io/babylon-toolkit/issues/2437).
- Re-recording the replay fixture from the two-hub devnet.
- A USDG icon, and the token registry's Optimism-address entries.
- The wallet nonce race seen when transactions run back to back (https://github.com/babylonlabs-io/babylon-toolkit/issues/2514).

## Pre-review

<details>
<summary>Pre-review: 26 findings · 23 fixed · 0 moot · 0 follow-up · 3 declined · 0 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | E2E per-reserve debt check scales with prior debt, so a landed borrow can fail and strand debt | normal | fixed |
| 2 | E2E CLI silently ignores --borrow-hub for multi-hub and --repay-hub when there is one loan | normal | fixed |
| 3 | repay-all with no debt is refused in the pre-flight with a --borrow-first hint it doesn't accept | normal | fixed |
| 4 | Hub registry lists only devnet hubs, so the testnet build shows every hub as an address with a warning | normal | fixed |
| 5 | BorrowedAsset.underlying is written but never read | normal | fixed |
| 6 | New data-hub and data-token-symbol attributes have no reader | normal | fixed |
| 7 | Borrow-form back arrow compares the reserve id as a string while the form parses it | normal | fixed |
| 8 | RepaySelectionPanel E2E hook comment names selectRepayDebt; the consumer is selectDebt | normal | fixed |
| 9 | Landing APR comment says reserves match on the registry label, but unregistered tokens match on the indexer symbol | normal | fixed |
| 10 | copy.ts: tokenOnHub comment overclaims, and the Available Liquidity header is duplicated | normal | fixed |
| 11 | Unknown-hub warning is missing on text-only surfaces (activity caption, success and validation copy) | normal | declined: sentences can't carry a tooltip; the hint lives on the structured Hub rows, the unknown hub is logged, and #4 removes the case on known deployments |
| 12 | e2e/real/actions/borrow.ts grows to about 600 lines | normal | declined: no repo line-limit rule; the hub steps sit with the borrow flow they drive |
| 13 | confirmBorrowSuccess doc still says it fails fast on any Transaction failed callout | normal | fixed |
| 14 | Stale-nonce retry helper discards the original waitFor error | normal | fixed |
| 15 | A --borrow-first repay may not target the reserve it just borrowed from, so the borrow can land and the repay refuse | normal | fixed |
| 16 | multi-hub accepts --borrow-amount=max, so leg 1 uses all borrowing capacity and the run aborts before any repay | normal | fixed |
| 17 | multi-hub's Loans-row check can still throw and skip every repay leg | normal | fixed |
| 18 | The stranded-debt report can miss a small leg on a heavily borrowed reserve | normal | fixed |
| 19 | The Select hub condition is written twice and can drift | normal | fixed |
| 20 | The useActivities test never covers the vBTC reserve | normal | fixed |
| 21 | The two new hubRegistry tests cannot fail independently because the registry is validated at import | normal | declined: the tests guard the case the PR reviewer raised: validation moving out of import time would otherwise go unnoticed |
| 22 | A stray --borrow-hub picks the repay hub on a run that does not borrow | normal | fixed |
| 23 | The --borrow-first repay comment says an overriding --repay-token is left alone, but the same token now pins the reserve | normal | fixed |
| 24 | The multi-hub failure message says repay-all 'confirms' the debt, but it repays every debt with Max | normal | fixed |
| 25 | repayBorrowedReserve's doc says the CLI pins the reserve whenever its read succeeded | normal | fixed |
| 26 | BorrowLeg is exported but only used inside borrow.ts | normal | fixed |

</details>


services/vault/e2e/real/actions/multiHub.ts
```diff
@@ -0,0 +1,217 @@
...
+    return [...reserves].sort((a, b) => (a.reserveId < b.reserveId ? -1 : 1));
+  }
+
+  const tokens = findMultiHubTokens(reserves);
+  if (tokens.length === 0)
+    throw new Error(
+      `multi-hub: no token is borrowable from more than one hub on ${network} — nothing to exercise.`,
+    );
+  const wanted = ctx.config.borrowToken?.trim();
+  if (!wanted) return tokens[0].reserves;
```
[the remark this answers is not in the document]
gbarkhatov: Agreed that a real-value run shouldn't pick a token on its own, but the implicit choice happened one level up: a non-interactive CLI run filled `--borrow-token` with the first multi-hub token before the action ran, so a throw in the action alone wouldn't have caught `--yes` runs. Fixed in both places: `cli.ts` now refuses a non-interactive `multi-hub` run without `--borrow-token` or `--all-reserves` and lists the candidate tokens (interactive runs still get the menu), and the action's first-token fallback is now an error.

gbarkhatov:
On the activity metadata finding in `useActivities.ts`: confirmed. It isn't new in this PR (main builds the same map from `borrowableReserves`) and was tracked for the next PR under https://github.com/babylonlabs-io/babylon-toolkit/issues/2437 Scope 11, but this PR already touches that map, so it now builds it from `allBorrowReserves` plus vBTC. A past borrow or repay on a frozen or paused reserve keeps its token, scaled amount and hub. `hooks/__tests__/useActivities.test.tsx` covers a frozen reserve.

services/vault/e2e/real/cli.ts
```diff
@@ -499,37 +569,97 @@ async function resolveConfig(
...
+        });
+        borrowToken = reserve.symbol;
+        borrowReserveId = reserve.reserveId.toString();
+      }
+    }
+
+    // Multi-hub extras: every reserve (`--all-reserves`) or one token on every hub that lists it, and a
+    // per-leg USD size (`--borrow-usd`). The token is validated up front against the live reserve list so
+    // a token on only one hub fails before the browser opens.
+    const allReserves =
```
jonybur, line 640: **--borrow-usd/--all-reserves are silently ignored outside multi-hub, and repay-all silently overrides --repay-hub/--repay-token/--repay-amount.**

`allReserves` and `borrowUsd` are only read when `action === "multi-hub"`, and nothing rejects them for other actions. So `borrow --borrow-usd=5` quietly borrows the default share of the max instead of $5, and `borrow --all-reserves` becomes a single-reserve borrow. The same happens for repay-all: `--repay-hub`, `--repay-token` and `--repay-amount` are parsed and validated here, but `repayAll.ts` overwrites them with `repayHub: undefined` and `MAX_AMOUNT_KEYWORD` on every leg, so `repay-all --repay-hub=core` repays every debt on every hub. These runs move real value, and this PR already refuses the matching mistake for multi-hub ("drop --borrow-hub"). Throw when these flags are passed to an action that doesn't use them.

services/vault/e2e/real/cli.ts
```diff
@@ -160,6 +179,54 @@ function optionalChoice<T extends string>(
...
+        label: describe(reserve),
+      })),
+    );
+    return candidates[Number(index)];
+  };
+
+  if (token === undefined) {
+    if (hub !== undefined)
+      throw new Error(`--${flag}-hub needs --${flag}-token as well.`);
+    return interactive ? choose(reserves) : reserves[0];
```
jonybur: **Non-interactive borrow with no --borrow-token still takes reserves[0], silently picking a hub now that one token spans several hubs.**

When `--borrow-token` is omitted, `pickReserve` returns `reserves[0]` in a non-interactive run, and `resolveBorrowReserve` in `e2e/real/actions/borrow.ts` (line 218) does the same when the CLI's reserve read failed. With two hubs listing the same tokens, `borrow --yes` now borrows real value from whichever reserve has the lowest id, without saying which hub. This is the same implicit selection that was fixed for `multi-hub` in this PR, and CLAUDE.md's no-silent-fallbacks rule points the same way. When there is more than one borrowable reserve, throw in non-interactive runs and list the candidates with `describeReserve`, as the multi-hub guard does.

services/vault/src/applications/aave/components/Detail/index.tsx
```diff
@@ -101,15 +117,59 @@ export function LoanFlowOverlay({
...
+  // through the pickers, and only for the token it names. A form opened from a
+  // Loans row, a market page or the token dropdown carries no `asset` and
+  // keeps the close button.
+  // Parsed the way the form parses it, so `?reserve=05` names reserve 5 here too.
+  const openReserveId = parseReserveId(reserveId);
+  const openReserve =
+    openReserveId === null
+      ? undefined
+      : allBorrowReserves.find((r) => r.reserveId === openReserveId);
+  const backSearch =
```
jonybur: **Select hub step has no back arrow to Select asset, although the form after it has one; every step uses replace, so browser Back exits the flow.**

`backSearch` is only computed when `showForm` is true, so at `?picker=borrow&asset=<underlying>` the shell renders only the close button. From the borrow form, back returns to Select hub, but from Select hub there is no way back to Select asset. Because every step navigates with `replace`, browser Back leaves the whole flow. A user who picked the wrong token has to close and start again. Set `backSearch` to `getAssetPickerSearch(LOAN_TAB.BORROW)` when the hub picker is showing, and add a case to the overlay tests.

services/vault/src/services/aave/__tests__/hubRegistry.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(
+      vi
+        .mocked(logger.warn)
+        .mock.calls.filter(([message]) => message.includes(unknownHub)),
+    ).toHaveLength(1);
+  });
+});
+
+describe("buildHubRegistry", () => {
+  it("builds the deployed registry without a duplicate label or address", () => {
```
jonybur: **Test named 'builds the deployed registry without a duplicate label or address' only checks two devnet labels differ; it never validates the deployed table.**

The test's name says it validates the deployed registry for duplicate labels and addresses, but its body only asserts that the two devnet hubs have different labels. It never runs `buildHubRegistry` on the real table, never checks addresses, and ignores the testnet entries. The real guard is the import-time `buildHubRegistry(HUBS_BY_DEPLOYMENT)` in `hubRegistry.ts`, which would crash the whole file rather than fail this test, and this test would keep passing if that validation became lazy. Export the table and assert `buildHubRegistry` doesn't throw on it, or rename the test to what it checks.

jonybur (commented):
Comments inline.

services/vault/e2e/real/cli.ts
gbarkhatov: Fixed. `cli.ts` now rejects `--all-reserves` and `--borrow-usd` on any action other than `multi-hub`, and rejects `--repay-token`, `--repay-hub` and `--repay-amount` on `repay-all`, so none of them is silently ignored on a run that moves value.

services/vault/e2e/real/cli.ts
gbarkhatov: Fixed, for repay as well as borrow. With no token, a non-interactive run now takes a reserve only when there is a single candidate and otherwise throws with the candidates listed (`pickReserve` in `cli.ts`). The action-side fallbacks match: `resolveBorrowReserve` in `borrow.ts` and `resolveRepayDebt` in `repay.ts` no longer take the first entry. Interactive runs still get the menu.

services/vault/src/applications/aave/components/Detail/index.tsx
gbarkhatov: Fixed. While Select hub is showing, the overlay now offers back to Select asset, so a wrong token pick no longer means closing the flow. Added an overlay test, "goes back to Select asset from Select hub".

services/vault/src/services/aave/__tests__/hubRegistry.test.ts
gbarkhatov: Fixed. `HUBS_BY_DEPLOYMENT` is exported, and the tests now assert that `buildHubRegistry(HUBS_BY_DEPLOYMENT)` doesn't throw and that every hub in the deployed table (devnet and testnet) resolves from the registry rather than by address.

jonybur (commented):
Comments inline.

services/vault/e2e/real/cli.ts
```diff
@@ -499,37 +575,105 @@ async function resolveConfig(
...
+        borrowReserveId = reserve.reserveId.toString();
+      }
+    }
+
+    // Multi-hub extras: every reserve (`--all-reserves`) or one token on every hub that lists it, and a
+    // per-leg USD size (`--borrow-usd`). The token is validated up front against the live reserve list so
+    // a token on only one hub fails before the browser opens.
+    const allReserves =
+      action === "multi-hub" && flagBool(flags["all-reserves"]);
+    const borrowUsd =
```
jonybur, line 642: **A bare --borrow-usd on multi-hub (no value) is silently ignored, so every leg borrows the default fraction instead of failing.**

`parseFlags` stores a flag with no `=` as `true`, and `borrowUsd` is only set when `typeof flags["borrow-usd"] === "string"`, so `--action=multi-hub --borrow-usd` leaves `borrowUsd` undefined. The positive-amount check below never runs, and the multi-hub-only guard at line 626 only fires for other actions. The run then borrows the conservative default on every reserve instead of stopping on a forgotten value, which is exactly the silent fallback the rest of this CLI now refuses on real-value runs. Throw when `flags["borrow-usd"] === true` (the same applies to `--borrow-amount`).

services/vault/e2e/real/cli.ts
```diff
@@ -579,40 +736,39 @@ async function resolveConfig(
...
+              });
+        repayToken = debt.symbol;
+        repayReserveId = debt.reserveId.toString();
       }
     }
-    // For a --borrow-first run (`repay` or `withdraw`), repay the token we just borrowed unless the user
-    // overrode --repay-token.
+    // For a --borrow-first run (`repay` or `withdraw`), repay the reserve we just borrowed from unless
+    // the user overrode --repay-token.
     if (
```
jonybur, line 823: **With --borrow-first, --repay-hub is ignored, and an explicit --repay-token falls back to --borrow-hub for a different token.**

When `--borrow-first` is set and `--repay-token` is omitted, this block pins the repay to `borrowReserveId` and never reads `repayHub`, so `--repay-hub=babylon` is silently dropped. When `--repay-token` is given, the CLI skips resolution, and `resolveRepayDebt` in `e2e/real/actions/repay.ts` (line 142) uses `repayHub ?? borrowHub`. So `repay --borrow-first --borrow-token=USDC --borrow-hub=core --repay-token=WBTC` repays the Core Hub WBTC debt without asking, even if WBTC is also owed on another hub. Reject `--repay-hub` without `--repay-token` under `--borrow-first`, and only fall back to `borrowHub` when the repay token equals the borrow token.

services/vault/e2e/real/cli.ts
```diff
@@ -499,37 +575,105 @@ async function resolveConfig(
...
+      (flags["all-reserves"] !== undefined || flags["borrow-usd"] !== undefined)
+    )
+      throw new Error(
+        "--all-reserves and --borrow-usd only apply to --action=multi-hub.",
+      );
+    if (allReserves && borrowToken !== undefined)
+      throw new Error(
+        "--all-reserves borrows from every reserve — drop --borrow-token.",
+      );
+    if (action === "multi-hub" && borrowHub !== undefined)
```
jonybur, line 669: **multi-hub silently ignores --repay-token, --repay-hub and --pegin-first instead of rejecting them.**

The CLI refuses `--borrow-hub` on multi-hub and refuses the repay flags on repay-all, but nothing refuses `--repay-token`, `--repay-hub` or `--pegin-first` on multi-hub. `legContext` in `multiHub.ts` overwrites `repayToken`/`repayHub` with each reserve's values, and `peginFirst` is never read because `willBorrow` excludes multi-hub. So `multi-hub --borrow-token=USDC --repay-hub=core` repays every hub, and `--pegin-first` borrows against existing collateral without pegging in. Add these flags to the multi-hub rejection block next to the `--borrow-hub` check.

services/vault/e2e/real/actions/multiHub.ts
```diff
@@ -0,0 +1,220 @@
...
+          ctx,
+          reserve,
+          before,
+          "rise",
+          borrowed.mode === "amount" ? borrowed.value : undefined,
+        );
+      }
+
+      currentStep = "loans-rows";
+      await assertLoanRowsNameHubs(ctx, reserves);
```
jonybur: **multi-hub runs a display-only Loans-row check after borrowing on every reserve; if it throws, all that debt is left open.**

All borrow legs run first (lines 165-191), then `assertLoanRowsNameHubs` throws if a row doesn't appear in time or its text lacks the hub label. The `finally` block only stops the recorder and removes the popup handler, so a label mismatch aborts the run with open debt on every reserve it just borrowed from, and the error doesn't mention `--action=repay-all`. A cosmetic check shouldn't strand real debt. Collect label mismatches and throw after the repay legs, or at least point the error at `repay-all` for cleanup.

services/vault/e2e/real/config.ts
```diff
@@ -157,6 +170,13 @@ export interface RunConfig {
   peginFirst?: boolean;
   /** Borrow only: token symbol to borrow (`--borrow-token`); defaults to the first borrowable reserve. */
```
jonybur: **RunConfig doc comments still say borrowToken/repayToken default to the first reserve or loan, which this PR removed.**

`borrowToken` is documented as defaulting to the first borrowable reserve, and `repayToken` (lines 190 and 207) as the sole/first outstanding loan. This PR changed `pickReserve`, `resolveBorrowReserve` and `resolveRepayDebt` so non-interactive runs only take a sole candidate and otherwise throw. On a CLI that moves real value, a reader trusting these comments will expect an implicit pick that no longer happens. CLAUDE.md asks for comments describing old behavior to be updated with the change. Reword them to describe the single-candidate or menu-or-error behavior.

services/vault/e2e/real/cli.ts
gbarkhatov: Fixed. `cli.ts` now refuses any value-taking flag passed bare — `--borrow-usd` and `--borrow-amount`, the token and hub flags, plus `--delay`, `--amount`, `--vp`, `--fixtures` and `--txid` — so a forgotten value stops the run instead of falling back to a default. The menu flags (`--target`, `--network`, `--btc`, `--eth`, `--action`, `--data`) already rejected it through `optionalChoice`.

services/vault/e2e/real/cli.ts
gbarkhatov: Fixed in both places. A `--borrow-first` run without `--repay-token` now rejects `--repay-hub` instead of dropping it, and `resolveRepayDebt` in `repay.ts` inherits `--borrow-hub` only when the repay token came from the borrow leg — an explicit `--repay-token` for another token no longer takes the borrow's hub.

services/vault/e2e/real/cli.ts
gbarkhatov: Fixed. `multi-hub` now rejects `--repay-token`, `--repay-hub` and `--pegin-first` next to the `--borrow-hub` check. `--repay-amount` stays accepted, since every repay leg uses it.

services/vault/e2e/real/actions/multiHub.ts
gbarkhatov: Fixed. The Loans-row check now collects what it finds wrong and reports it after the repay legs, so a display mismatch can no longer leave the borrowed debt open. A repay leg that fails also names `--action=repay-all` for clearing what the run leaves behind.

services/vault/e2e/real/config.ts
gbarkhatov: Fixed. `borrowToken`, `repayToken` and the `--repay-first` note now say that an interactive run offers a menu and a non-interactive one requires the flag unless there is a single candidate.

services/vault/e2e/real/actions/multiHub.ts
```diff
@@ -0,0 +1,247 @@
...
+              reserves.map((r) => r.reserveId),
+            );
+
+      for (const reserve of reserves) {
+        const borrowAmount = legBorrowAmount(ctx, reserve, pricesUsd);
+        const before = await readReserveDebt(ctx, reserve.reserveId);
+        log(
+          `── Borrow leg: ${describeReserve(reserve)}${borrowAmount ? ` — ${borrowAmount} ${reserve.symbol}` : ""}`,
+        );
+        const borrowed = await runBorrowWithOptionalPegin(
```
jonybur: **A failing borrow leg (or a repay leg's debt check) aborts multi-hub with earlier legs' real debt open and no pointer to repay-all.**

The borrow loop calls `runBorrowWithOptionalPegin` and `waitForReserveDebt` without any try/catch, so if leg 3 of 6 times out or its debt check fails, the run throws and legs 1-2 stay borrowed on Sepolia. The same is true of `waitForReserveDebt(ctx, reserve, before, "fall")` at line 234, which sits outside the try that adds the cleanup hint. The earlier fix only wrapped `runRepayFlow`, so these paths still end with an error that doesn't say debt was left behind or how to clear it. Wrap each borrow leg (once an earlier leg has borrowed) and the repay-side debt check in the same catch that appends `clear it with --action=repay-all`.

services/vault/src/services/activity/projection.ts
```diff
@@ -184,10 +184,16 @@ export function toNumericAmount(amount: string, decimals: number): number {
  * activity data.
  */
 export interface FetchUserActivitiesDeps {
-  /** Map of debtReserveId -> { symbol, decimals, icon }, e.g. from `useAaveConfig().borrowableReserves`. */
+  /** Map of debtReserveId -> { symbol, decimals, icon, hubLabel }, e.g. from `useAaveConfig().borrowableReserves`. */
```
jonybur: **Doc comment this PR edited still says the activity reserve map comes from borrowableReserves; it is now built from allBorrowReserves plus vBTC.**

This line was touched to add `hubLabel`, but it still says the map comes "e.g. from `useAaveConfig().borrowableReserves`". `useActivities.ts` now builds it from `allBorrowReserves` plus the vBTC reserve, and that is exactly what keeps frozen or paused reserves labelled in activity. A reader following the comment would reintroduce the bug that was just fixed. CLAUDE.md asks that comments describing old behavior be updated with the change, so point it at `allBorrowReserves`.

jonybur (commented):
Comments inline.

services/vault/e2e/real/actions/multiHub.ts
gbarkhatov: Fixed properly this time. One try/catch now wraps the whole borrow-and-repay sequence and counts the reserves that have borrowed but not yet been repaid, so a mid-way borrow failure, the repay-side debt check and the Loans-row report all end with how many reserves are still owed and the `--action=repay-all` pointer. A failure before the first borrow is unchanged.

services/vault/src/services/activity/projection.ts
gbarkhatov: Fixed. The comment now points at `useAaveConfig().allBorrowReserves` plus the vBTC reserve, and says why: every loan reserve, not just the borrowable ones, so history on a reserve that has since been frozen or paused still renders.

services/vault/e2e/real/actions/multiHub.ts
```diff
@@ -0,0 +1,253 @@
...
+              borrowHub: undefined,
+              borrowAmount,
+              peginFirst: false,
+            }),
+            (step) => {
+              currentStep = `borrow:${reserve.reserveId}:${step}`;
+            },
+          );
+          // The borrow transaction confirmed, so this reserve owes from here until its repay leg.
+          openLegs += 1;
```
jonybur: **multi-hub counts a borrow leg as owed only after runBorrowWithOptionalPegin returns, so a borrow that was submitted but then fails a post-check is not reported as open debt.**

`openLegs += 1` runs only after `runBorrowWithOptionalPegin` returns, but that call can throw after the borrow was already submitted: `confirmBorrowSuccess` times out waiting for the success screen, or `assertBorrowDebtIncreased` fails after the success screen (borrow.ts:429 and :468). If that happens on the first leg, `openLegs` is still 0 and the catch at line 241 rethrows the plain error with no `--action=repay-all` hint, even though real debt may be open. On a later leg the reported count is one too low. The comment on line 200 also claims more than the code has checked at that point. Count the leg as open right before the submit click (for example, pass a callback into the borrow flow), or re-read the current reserve's debt against `before` in the catch and count it if it rose.

services/vault/e2e/real/actions/multiHub.ts
```diff
@@ -0,0 +1,253 @@
...
+              repayHub: undefined,
+              borrowFirst: false,
+            }),
+            (step) => {
+              currentStep = `repay:${reserve.reserveId}:${step}`;
+            },
+          );
+          currentStep = `repay:${reserve.reserveId}:reserve-debt`;
+          await waitForReserveDebt(ctx, reserve, before, "fall");
+          openLegs -= 1;
```
jonybur: **Without --repay-amount each multi-hub repay leg pays only ~25% of the debt, yet the leg is counted as cleared and the run reports success.**

The repay legs pin the reserve but leave `repayAmount` as passed, and the CLI only defaults it to `max` for `withdraw --repay-first` (cli.ts:755). With no `--repay-amount`, `resolveRepayAmount` pays `debtTokens * CONSERVATIVE_REPAY_FRACTION` (repay.ts:179-183), and `waitForReserveDebt(..., "fall")` only requires `after < before`, so `openLegs -= 1` runs while about 75% of the debt is still owed. A clean run logs "Multi-hub complete" with debt left on every reserve, and if a later leg fails, the "N reserves still owed" count is too low. Either default multi-hub repay legs to `max` (as withdraw --repay-first does), or only decrement `openLegs` once the reserve's debt reaches zero.

jonybur (commented):
Comments inline, plus:

- `services/vault/src/applications/aave/types.ts:7` - **Asset.priceUsd is left with no reader or writer after this PR removed its only use in AssetSelectionPanel.**
  
  The rewrite of `AssetSelectionPanel` removed `pricesByReserveId[reserveKey] ?? asset.priceUsd`, which was the only read of this optional field, and `AssetListItem` dropped its own `priceUsd` prop. Nothing in `services/vault/src` sets or reads `Asset.priceUsd` any more; the remaining `priceUsd` hits are `useAaveReservePrice` results and local variables. The repo rules ask that code this change orphans is removed with it, so drop the field from the `Asset` interface.

services/vault/e2e/real/actions/multiHub.ts
gbarkhatov: Fixed. Each leg is now recorded as attempted before its borrow flow starts, so a borrow that was submitted and then failed its success-screen or debt check is covered. On failure, the run reads the debt again and names the reserves that still owe more than before the run. It also names attempted legs that don't read as owed, since a small leg on a heavily borrowed reserve can look like interest. The message ends by saying to check the debt and clear it with `--action=repay-all` if needed. If the debt can't be read, it lists every reserve that may still be owed.

services/vault/e2e/real/actions/multiHub.ts
gbarkhatov: Fixed. Every multi-hub repay leg now uses the form's Max, and a leg only counts as cleared once its reserve's debt is back to its pre-run level, so a clean run can't finish with debt left. On multi-hub, `--repay-amount` now accepts only `max`, which replaces my earlier reply saying it stayed accepted.

services/vault/e2e/real/actions/multiHub.ts
```diff
@@ -0,0 +1,302 @@
...
+            );
+
+      // Each reserve's debt before this run borrowed from it, the legs whose borrow was started, and the
+      // legs confirmed back at that debt. A leg counts as started before its borrow flow runs: the flow can
+      // submit and then still throw on a post-check, and that debt is real.
+      const baselines = new Map<bigint, number>();
+      const attempted: BorrowReserve[] = [];
+      const cleared = new Set<bigint>();
+      try {
+        for (const reserve of reserves) {
```
jrwbabylonlab, line 198: This loop completes every borrow before the first repay, so a debt-free run needs N simultaneous borrowed reserves. Aave's one-borrowable-asset decision is enforced by `Spoke.MAX_USER_RESERVES_LIMIT`, which counts borrowed reserve ids across the position-wide bitmap with no hub dimension (`Spoke.borrow` → `positionStatus.borrowCount`). If a spoke ships with that limit at 1, the second distinct reserve reverts `MaximumUserReservesExceeded` in the pre-flight simulation, before signing.

Nothing to change now. But this action and `--all-reserves` would then need to interleave borrow and repay, and drop the simultaneous multi-row assertion. Tracked in #2513.

jrwbabylonlab (approved):

services/vault/e2e/real/actions/multiHub.ts
gbarkhatov: Agreed. The current devnet spoke allows it: the devnet run borrowed from all six reserves at once. If a spoke ships with the limit at 1, `multi-hub` and `--all-reserves` will need to alternate borrow and repay per reserve and check each Loans row on its own. That change belongs with https://github.com/babylonlabs-io/babylon-toolkit/issues/2513.

PR #2516, opened 2026-09-15
feat(wasm, ts-sdk)!: restore engine class exports and drop the raw entry

jonybur:
## Why

#2307 moved the engine's wasm-bindgen classes from its main import to a new `./raw` entry, so the main import could skip loading the generated script. That entry then needed a deprecation (#2464) and a large guard stack (#2481, #2493). #2228 does not need it: the SDK and Vault load the engine only on demand either way.

## What

- **Breaking:** the engine's `./raw` entry and the SDK's `loadRawTbvWasm()` are removed. Import the classes from the engine root, or call `loadTbvWasm()`. Both READMEs give the migration path.
- The engine's main import exports `WasmPeginTx`, `WasmPeginPayoutConnector`, `WasmPrePeginTx` and `WasmPrePeginHtlcConnector` again, as before #2307.
- Importing the engine's main entry loads its 85 KB generated script. The 3.1 MB binary still loads only on `initWasm()` or the first facade call.
- The raw-engine migration guide is removed.

## How

- `index.ts` and `index-node.ts` re-export the classes with their generated types. The facade functions and loaders are unchanged, so the classes and the facade share one initialization.
- `loadTbvWasm()` now initializes the engine and names the package when that fails. `buildRefundPsbt` gets `WasmPrePeginTx` from it. All of its checks are unchanged.
- A test pins each entry's glue re-exports to exactly the four classes. An SDK lint rule allows the classes only in `refund.ts`.
- `check-lazy-entries.js` allows only a namespace import of the generated script in each entry. It fails if importing either entry loads the binary, in the browser or in Node. The SDK's `check-eth-build.js` is unchanged.
- CLAUDE.md, SECURITY.md, CODEOWNERS and both critical-path workflows no longer name the raw files.

Tested: engine 15 tests, SDK 2,012 tests plus 8 facade checks, Vault 3,952 passed and 2 skipped, Ledger signer 658 passed; lint, type check and API docs clean.

Tracks #2361. Replaces #2481 and #2493.


packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
```diff
@@ -11,49 +11,37 @@ const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..');
...
-    );
-    for (const name of rawClassNames) {
-      assert.equal(typeof raw[name], 'function', name);
-      assert.equal(raw[name], generated[name]);
+    for (const name of classNames) {
+      assert.equal(typeof main[name], 'function', name);
+      assert.equal(main[name], generated[name], name);
     }
-    assert.equal(raw.initWasm, loader.initWasm);
   });
```
gbarkhatov: **Fix before merge — the exact-export-set assertion is gone, so a fifth unguarded class can land on a §1 critical-path entry with nothing failing.**

Base asserted `deepEqual(Object.keys(raw).sort(), [...rawClassNames, 'initWasm'].sort())` — an exclusivity pin on the unguarded surface. This loop only checks the four named classes are *present* and `=== generated[name]`. Presence, never exclusivity.

That matters because the glue exports more than four. `grep -n "^export class" dist/generated/vault_wasm.d.ts` returns eight: `PeginP2aAnchorOutput`, `WasmAssertChallengeAssertConnector`, `WasmAssertPayoutNoPayoutConnector`, **`WasmPayoutTx`**, plus the four re-exported here. So one line in `index.ts` — `export const WasmPayoutTx = generated.WasmPayoutTx` — would hand out the §3 payout builder with no facade value guards and pass everything: the entry walk in `check-lazy-entries.js` filters that edge by specifier string (identical), and this test does not enumerate.

I acknowledge an exact `deepEqual` on `Object.keys(main)` is no longer possible — base pinned a dedicated module that held only those five exports, whereas `index.js` carries the whole facade. So pin the intersection instead, which restores the base guarantee without trying to enumerate the facade:

```js
assert.deepEqual(
  Object.keys(main)
    .filter((k) => Object.hasOwn(generated, k) && main[k] === generated[k])
    .sort(),
  [...classNames].sort(),
);
```

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
```diff
@@ -300,21 +234,20 @@ try {
...
-  const facadeBrowserUrl = pathToFileURL(
+  const browserUrl = pathToFileURL(
     join(browserRacePackage, 'dist', 'index.js'),
   ).href;
-  const rawBrowser = await import(`${rawBrowserUrl}?shared-browser-init=raw`);
-  const facadeBrowser = await import(
-    `${facadeBrowserUrl}?shared-browser-init=facade`
-  );
+  const browserEntry = await import(`${browserUrl}?shared-browser-init`);
+  if (fetchCalls !== 0)
```
gbarkhatov, line 251: **Fix before merge — this lost all runtime coverage of the node entry, and the "browser and node" claim at :330 is now false.**

All three review lanes converged here and I confirmed it directly. `grep -n "index-node"` on this file at head returns only line 126 (static source walk) and line 153 (`.d.ts` pin) — **no runtime import of `index-node.js` anywhere**. The deleted isolated-package block was the sole runtime import of both entries, looping `['index.js', 'index-node.js']` after wiping `dist/generated`.

Its replacement is one browser-only line: `if (fetchCalls !== 0) throw ...`. That counter could never cover node regardless — `src/wasm-loader-node.ts:1,36` reads the binary with `readFile` from `node:fs/promises`, not `fetch`. Yet line 330 still prints `'Lazy WASM facade boundary verified (browser and node).'`

To be clear that this was not gratuitous: **deleting the old block was necessary.** It wiped `dist/generated` before importing the entries, which cannot survive entries that now statically import the glue. The gap is that the replacement only covers one side.

Concrete regression that passes head CI: add a top-level `await initializeWasm()` to `src/index-node.ts` and every Node/SSR consumer eagerly pulls 3.1 MB at import time, with nothing failing.

Either add a node arm — stub `fs/promises.readFile` with a counter mirroring the `fetchCalls` pattern at :227 — or narrow the message at :330 to "(browser)" so the script does not overstate its coverage.

packages/babylon-tbv-rust-wasm/src/index.ts
```diff
@@ -492,3 +498,20 @@ export {
...
+  generated.WasmPeginPayoutConnector;
+export type WasmPeginPayoutConnector = Bindings.WasmPeginPayoutConnector;
+
+export const WasmPrePeginTx: typeof Bindings.WasmPrePeginTx =
+  generated.WasmPrePeginTx;
+export type WasmPrePeginTx = Bindings.WasmPrePeginTx;
+
+export const WasmPrePeginHtlcConnector: typeof Bindings.WasmPrePeginHtlcConnector =
+  generated.WasmPrePeginHtlcConnector;
+export type WasmPrePeginHtlcConnector = Bindings.WasmPrePeginHtlcConnector;
```
gbarkhatov, line 520: **Suggestion — §1's "only consumer is `refund.ts`" is now enforced by nothing mechanical, and both signals that did enforce it are deleted in this same commit.**

This is the highest-value hardening I found, and it is also the substantive version of greptile's open P2: the README gap it names is real but cosmetic; the missing piece is mechanical, not documentary.

At base, a consumer had to call `loadRawTbvWasm()` — a `@deprecated` function whose deprecation was itself asserted by two tests (`wasm-facade.node.mjs` "publishes the raw loader deprecation in its declaration", `wasm-loader.test.mjs` "marks consumer imports deprecated"). Both tests and both tags are deleted here, and the classes now ride the ordinary `loadTbvWasm()` return. So `const { WasmPeginTx } = await loadTbvWasm();` type-checks anywhere in `src/` with no strikethrough and no lint error.

I checked the one rule that could catch it: `packages/babylon-ts-sdk/eslint.config.mjs:60-81` bans the package specifier outside `src/tbv/core/wasm/**`, but the new path is a destructure off `../../wasm`, which a specifier rule cannot see.

Un-deprecating is the point of this PR, so this is not a request to revert — it is a request to replace the lost signal. Cheapest replacement, in the eslint block already there: a `no-restricted-syntax` rule matching `ObjectPattern > Property[key.name=/^Wasm(PeginTx|PrePeginTx|PeginPayoutConnector|PrePeginHtlcConnector)$/]`, with `ignores` limited to `refund.ts` and tests. That turns §1's sentence into an enforced invariant rather than a convention.

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
```diff
@@ -32,30 +27,6 @@ test("retains the raw loader and guarded builders in ESM and CommonJS", async ()
   }
```
gbarkhatov: **Suggestion — the ESM/CJS parity loop no longer loads `tbv/core/wasm`, leaving that §9-registered subpath's CommonJS condition unchecked.**

Base ran `const wasm = await load("@babylonlabs-io/ts-sdk/tbv/core/wasm")` inside this `[import, require]` loop; head loads only `.../tbv/core/primitives`. The file's top-level import of `tbv/core/wasm` at line 12 is ESM-only, so nothing now exercises the `require()` export condition for the SDK's lazy WASM boundary — a CJS resolution break in the exports map would ship green.

Removing the `loadRawTbvWasm` assertion was required. Dropping the module load along with it was not. Keeping the load in the loop and asserting `typeof wasm.loadTbvWasm === "function"` restores the coverage.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/refund.ts
```diff
@@ -72,7 +72,8 @@ export interface BuildRefundPsbtResult {
 export async function buildRefundPsbt(
   params: BuildRefundPsbtParams,
 ): Promise<BuildRefundPsbtResult> {
-  const { WasmPrePeginTx } = await loadRawTbvWasm();
+  const { initWasm, WasmPrePeginTx } = await loadTbvWasm();
+  await initWasm();
```
gbarkhatov: **Suggestion — an engine init failure now surfaces unwrapped, where the deleted loader added actionable context.**

The deleted `loadRawTbvWasm` awaited `initWasm()` *inside* the loader and wrapped init failure as "resolved but its WebAssembly binary failed to initialize", with a `cause`. Head splits that: `loadTbvWasm()` then `await initWasm()` at the call site. I confirmed `loadTbvWasm` (`wasm/index.ts:36-50`) `.catch`es **only the dynamic import** — an init failure passes straight through.

So a stale or missing `vault_wasm_bg.wasm` now surfaces as a bare `ENOENT` from `wasm-loader-node.ts:36`, naming neither the engine package nor the rebuild. CLAUDE.md asks that error messages be actionable.

Two things moved at once here: the init *guarantee* left the boundary module for a call site (no test covers its absence — `refund.test.ts` pre-inits in `beforeAll`), and the init *diagnostics* were dropped. Wrapping `await initWasm()` inside `loadTbvWasm` restores both, and every future caller inherits it.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
```diff
@@ -122,15 +120,18 @@ function eagerGeneratedEdges(entry) {
...
-      `${entryName} statically reaches generated WASM glue, so importing the ` +
-        `facade loads the engine eagerly. Reach the glue only through ` +
-        `import('./generated/vault_wasm.js') inside a loader, or through the ` +
-        `explicit raw entry:\n${edges.join('\n')}`,
+      `${entryName} statically reaches generated WASM glue through a module ` +
+        `other than the entry. Reach the glue only through ` +
+        `import('./generated/vault_wasm.js') inside a loader:\n${edges.join('\n')}`,
     );
   }
 }
```
gbarkhatov, line 145: **Suggestion (low, pairs with the export-set item) — the new allowance is specifier-shaped, so it would also whitelist a blanket re-export of the whole generated surface.**

`classEdge` is the string `` `${entry} -> ./generated/vault_wasm.js` `` and the filter drops any edge equal to it. `export * from './generated/vault_wasm.js'` produces exactly that string and would be silently allowed — re-exporting all eight generated classes, not the four intended.

Head uses explicit `export const X = generated.X` (verified at `index.ts:504-517`), so this is hypothetical today. But base forbade any glue edge from the entry outright, so the door is new. Fixing the export-set assertion in `wasm-loader.test.mjs` closes this off too, which is why I'd treat that one as the real fix and this as context for it.

gbarkhatov (dismissed):
Reviewed across three independent lanes, then verified every finding against the code at head myself before posting. Refuted claims are listed below with the evidence that killed them.

**The value-safety story holds.** `buildRefundPsbt`'s checks really are unchanged — the entire hunk is two lines (loader swap plus `await initWasm()`), and `assertPositiveBigintArray`, the `deriveExpectedPrePeginHtlc` cross-check, `expectedHtlcValue` captured before `fromFundedTransaction`, funded-output script and value, version/locktime, input count/txid/index/sequence, output count/BIP-86 destination and the final `htlc − fee` check are all untouched. It is the same class object either way: both `raw.ts` and `index.ts:511` export `generated.WasmPrePeginTx`. The PR also *adds* a guard test that base lacked.

**What the change actually costs is mechanical enforcement, not correctness.** Two automated guards that pinned the unguarded surface are deleted, and what replaces them is prose. Nothing here ships a wrong value today, and both items below are small to restore. Neither is a request to revert the re-export — un-deprecating is the PR's point.

**Two items I'd fix before merge**, both inline: the deleted exact-export-set assertion no longer prevents a fifth unguarded wasm-bindgen class from appearing on a §1 critical-path entry (the glue exports eight, including `WasmPayoutTx`, the §3 payout builder), and `check-lazy-entries.js` lost all runtime coverage of the node entry while still printing "verified (browser and node)".

### Body-only notes

- **Both guard documents drop the pointer to #2361 in the very change that widens the crossing.** Base §1 ended "…the compatibility blocker in #2361" and SECURITY.md carried the same reference; both bullets are removed here. The PR body says it tracks #2361 and does not close it, so a reader of the guard documents now has no route from "this export is unguarded" to the open issue. One restored sentence in each.
- **Two published export surfaces are removed under a `feat` title with no breaking marker.** `nx.json:70` sets `conventionalCommits: true`, and both packages are 0.x (`babylon-tbv-rust-wasm` 0.15.0, `ts-sdk` 0.63.0), so a minor bump is a legitimate home for a break — not a strict semver violation. But an installed consumer calling `loadRawTbvWasm()` or importing `.../babylon-tbv-rust-wasm/raw` gets `ERR_PACKAGE_PATH_NOT_EXPORTED` on a routine upgrade, and `docs/guides/raw-engine-migration.md` is deleted in the same commit, so the URL the old `@deprecated` tags pointed at 404s on `main`. A `BREAKING CHANGE:` footer, or two lines in the SDK README pointing removed callers at `loadTbvWasm()`.
- **The rewritten README section drops the names of the guarded alternatives.** The removed text named `buildPrePeginPsbt`, `buildPeginTxFromFundedPrePegin` and `buildRefundPsbt`; the replacement's only sample constructs `new WasmPrePeginTx(...)` directly. A one-line pointer under "WASM Classes" restores it. (Flagging only this concrete half — the docs-gap framing is already greptile's open P2, and I don't want to double-bill it.)

### Merge-order note — #2517 conflicts with this PR on one file

Not a defect in either PR, but worth knowing before merge. Both PRs branch from the same `origin/main` commit `2a77c384`. I ran `git merge-tree` on the two heads: exactly one conflict, `modify/delete` on `packages/babylon-ts-sdk/docs/guides/raw-engine-migration.md` — deleted here, modified in #2517. The other five shared files (`CLAUDE.md`, `SECURITY.md`, `babylon-tbv-rust-wasm/src/index.ts`, `refund.ts`, `tbv/core/wasm/index.ts`) auto-merge cleanly, and I checked the merged result: no stale `/raw`, `raw.ts`, `loadRawTbvWasm` or migration-guide references survive in either guard document.

#2517's entire change to the conflicting file is one line — `The guarded product-path work in #2386 remains separate.` becomes a note that #2386 was closed as not planned. So the correct resolution is simply to take this PR's deletion; that note dies with the file. #2517 is CLEAN and approved while this one is BLOCKED, so #2517 lands first and this PR's author takes the conflict.

### Refuted — please don't re-raise these

- **"The race block is structurally unfalsifiable and `: 200` is dead code."** Read `check-lazy-entries.js:222-271`. The block starts `browserEntry.initWasm()` and `browserEntry.createPayoutConnector(...)` concurrently — the loader's explicit init and the facade's internal `getWasmBindings()` are distinct code paths, so `fetchCalls === 1` is a real property, not a tautology, and the `: 200` arm is the failure-detection delay that fires exactly when the property breaks. It is the assertion's failure path, not dead code. The test also now exercises a re-exported class at :250, which base did not.
- **"The guard story was deleted rather than replaced."** This is the framing the diff most invites, and it is wrong. CLAUDE.md §1 swaps the `/raw` bullet for "The engine main entry … also exports the wasm-bindgen classes … the rule above binds at each call site, not at the export … Every new class consumer must do equivalent cross-checks", with SECURITY.md:217-221 and :232 mirrored. The scope is *widened*, not dropped. My two inline items are about enforcement, not about the text.
- **§4 frozen-expander concerns.** Raised from a `git diff` against a stale local `main` and belonging to other commits. Re-derived from the PR diff: this PR's CLAUDE.md change is exactly the §1 and §9 hunks (+2/−3), SECURITY.md exactly the second-crossing paragraph plus the reviewer-rule line (+4/−9). **§4 is not engaged by this PR** — the re-export blocks touch none of the three expanders, and `index.ts:502-517` and `index-node.ts:582-597` are byte-identical to each other.

### Verified clean

- **Zero dead `/raw` references.** My own sweep — `git grep -nE "loadRawTbvWasm|raw-engine-migration|tbv-rust-wasm/raw|RawTbvWasm|rawWasmModulePromise|raw-node|dist/raw"` excluding `dist/` — returns nothing across the tree, including the lockfile, `.github/`, `docs/api/` and all markdown.
- **The guard-registry edit does not outpace the code.** All six removed lines in CODEOWNERS, `critical-path-check.yml` and `claude-md-drift.yml` are fully-anchored literal paths to `src/raw.ts` / `src/raw-node.ts` with no glob metacharacters, and both files are deleted here. `index.ts`, `index-node.ts`, `wasm-loader.ts`, `wasm-loader-node.ts`, `src/tbv/core/wasm/` and `refund.ts` all remain registered in all three. `claude-md-drift.yml` gates on `[ -e "$p" ]`, so leaving the paths in would have produced a false failure every run — removing them was required.
- **A guard was also added:** `test('exports only the root entry')` pins `Object.keys(packageExports)` to exactly `['.']`, so a future subpath cannot reappear unnoticed.
- **`check-eth-build.js` is untouched and still bites**, and `loadTbvWasm` remains a dynamic `import(...)` at `wasm/index.ts:37` — the SDK still reaches the engine only via dynamic import.

Everything below the two fix-before-merge items is a suggestion; your call as author.

packages/babylon-tbv-rust-wasm/tests/wasm-loader.test.mjs
jonybur: Fixed in a03ce11f3e. The pin now matches by value, not by name, for `index` and `index-node`. It collects every entry export whose value is a glue export or the glue namespace, and requires exactly the four classes.

Checked by mutation: `export const WasmPayoutTx = generated.WasmPayoutTx` in either entry, a renamed glue export, and `export *` all fail the test.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
jonybur: Fixed in a03ce11f3e. `check-lazy-entries.js` has a Node arm again. It copies the package, counts reads of `vault_wasm_bg.wasm` through `fs/promises.readFile` and `fs.readFileSync`, imports `dist/index-node.js`, waits one macrotask, and requires zero reads. It then calls `initWasm()` and requires exactly one read, which proves the counter sees the loader. The browser arm also waits one macrotask now, so a fire-and-forget init fails in both arms.

Checked by mutation: a top-level `await initWasm()` and a `void initWasm()` in `index-node.ts` both fail the build. The "browser and node" message is now true.

packages/babylon-tbv-rust-wasm/src/index.ts
jonybur: Added in a03ce11f3e, in the same eslint config. A `no-restricted-syntax` block for `src/**/*.ts` flags the four class names in a destructure (renamed and string keys too) and in member access (`wasm.WasmPeginTx`, `wasm["WasmPeginTx"]`). It ignores only `refund.ts` and tests. The message points to the guarded builders.

Checked by mutation in `pegin.ts`: a destructure and member access both fail lint, and `refund.ts` passes. The rule matches literal names, so a key held in a variable still passes. It is a tripwire, not a proof.

packages/babylon-ts-sdk/tests/wasm-facade.node.mjs
jonybur: Fixed in a03ce11f3e. The `[import, require]` loop loads `@babylonlabs-io/ts-sdk/tbv/core/wasm` again. It asserts that `loadTbvWasm()` resolves to the engine module in both module systems.

Checked by mutation: removing the CommonJS export from the built file fails the test.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/refund.ts
jonybur: Fixed in a03ce11f3e, as suggested. `loadTbvWasm()` awaits `initWasm()` after the import. On failure it clears its cache and throws an error that names the engine package, with the original error as its cause. `refund.ts` no longer calls `initWasm()`, so every caller gets an initialized module.

Three loader tests cover the message, the cause chain and a retry. Two engine mocks gained `initWasm`. Every facade caller already awaited the same engine initialization, so this adds no load.

packages/babylon-tbv-rust-wasm/scripts/check-lazy-entries.js
jonybur: Fixed in a03ce11f3e. The entry's glue edge is allowed only as a namespace import (`import * as generated from './generated/vault_wasm.js'`). `export * from` and `export * as x from` fail the build for either entry. The value-based pin from the export-set thread also catches a re-export of the namespace object.

packages/babylon-tbv-rust-wasm/README.md
```diff
@@ -466,30 +462,24 @@ Value: `"50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0"`
...
-is set. Deprecation alone does not close
-[#2361](https://github.com/babylonlabs-io/babylon-toolkit/issues/2361).
+```ts
+import {
+  initWasm,
+  WasmPrePeginTx,
+} from "@babylonlabs-io/babylon-tbv-rust-wasm";
+
+await initWasm();
+const transaction = new WasmPrePeginTx(/* ... */);
```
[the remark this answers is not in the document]
jonybur: Fixed in a03ce11f3e. The "WASM Classes" section now:

- names the guarded builders: `buildPrePeginPsbt`, `buildPeginTxFromFundedPrePegin`, `buildRefundPsbt`, `buildPeginInputPsbt` and `buildPayoutPsbt`
- says a caller that signs must derive the expected outputs and signing data independently, and compare them with the class values before signing
- says `initWasm()` and the facade's amount bounds do not prove that the bytes match the request
- tells callers to pass amounts through `assertPositiveBigintArray` first, because `new BigUint64Array()` wraps bad values without an error
- says `WasmPeginTx.fromJson` takes no trusted inputs
- tells callers to `free()` each instance

Each class export also has a JSDoc pointer to this section.

jonybur:
@gbarkhatov Thanks for the review. a03ce11f3e also covers the body notes:

- **#2361 pointer:** CLAUDE.md section 1 and SECURITY.md both say "#2361 tracks this unguarded export." Both also say that the SDK's public `loadTbvWasm()` returns the classes.
- **Breaking marker:** the title is now `feat(wasm, ts-sdk)!:`, so the release bumps the major version. The SDK README has a "`loadRawTbvWasm()` is removed" section. The engine README says that the `/raw` entry is removed and where the classes are now.
- **Guarded alternatives:** the engine README names the five builders again.
- **#2517:** noted. If it lands first, this PR keeps its deletion of the migration guide.

The push dismissed your approval. Please take another look.

jonybur:
@greptileai review the full PR at current commit a03ce11f3e80c6604e1a2b787da18e78d146d1cf.

The P1 compatibility break is now explicit. The title has a breaking-change marker, so the release bumps the major version. The SDK README and the engine README give the migration path for `loadRawTbvWasm()` and the `/raw` entry. The P2 README thread is fixed and resolved.

Verification: the engine build and 15 tests pass. The SDK passes 2,012 tests plus 8 built-package checks. Vault passes 3,952 tests with 2 skipped. Lint and API docs are clean.

gbarkhatov (approved):

Issue #2371, opened 2026-08-31
Ledger: make Refund reachable on device (intent gate + two TAP_BIP32_DERIVATION fields)


jrwbabylonlab:
> **STATUS 2026-08-31: implemented on the working tree** (signer `refundPsbt.ts` + provider gates + Speculos stage 13), dual-reviewed. Review corrections folded in:
> - Refunds consume **no device dedup mask in any state** (standalone dispatch fall-through, `sign_psbt_validate.c:3552-3618`; standalone sign section touches no `G_vault_context`) — refunds are fully exempt from the host replay guard, and a batched refund in `signPsbts` is refused with a typed error.
> - The device CSV floor (`csv >= 72` with no intent, `:898-903`) is pre-empted host-side in `augmentPsbtForRefund`.
> - The wrong-vault pins fire pre-approval on-device (see correction inline below).
>
> **Speculos e2e PASSED 2026-08-31**: full 13-stage suite 41/41 against a nanosp testnet ELF rebuilt from `29beb88d5` — refund signed under the loaded intent, intent/caps survived, and a foreign-prevout refund signed under IDLE (proves the no-intent branch). Remaining: real-device check in #2110's manual run.

Refund is the only escape from a deposit that stalls, and it is unreachable for a Ledger depositor. **Both blockers are ours — the device is willing to sign a refund.** Child of #2111 (T6).

```
depositor clicks Refund
        │
        ▼
  SDK builds refund PSBT ──► pays BIP-86 P2TR(depositor)   OK already
        │                     witnessUtxo, tapLeafScript, tapInternalKey
        │                     ✗ no tapBip32Derivation anywhere        ← BLOCKER 2
        ▼
  LedgerVaultProvider.signPsbt
        │
        └─► gateSignContext() → requireIntent defaults true
                              → throws DEVICE_CEREMONY_INVALID        ← BLOCKER 1
                                (never reaches the device)
```

Pins for every citation below: firmware `ff1e1ce17` (v0.10.0, tip of `fix/feedback`), btc-vault `2ba93bbaa`, toolkit `bf9fa6f99`.

## Blocker 1 — our own intent gate

`provider.ts:804` — `gateSignContext(requireIntent = true)`; `:816` throws when `deviceState.phase !== "intent-loaded"`. `signPsbt` (`:682`) takes the default, so every standalone sign dies before device I/O. The only `false` today is PoP (`:1068`).

[...]

PR #2517, opened 2026-09-16
chore(ledger-vault, ts-sdk): re-pin firmware citations to 0.10.1 and …

jrwbabylonlab:
chore(ledger-vault, ts-sdk): re-pin firmware citations to 0.10.1 and correct SDK docs

## What

Re-pins every firmware and SDK citation in the Ledger signer package and the wallet-connector Ledger provider to `app-babylon-vault` @ `b0c0ac4d` (app 0.10.1) and `ledger-secure-sdk` `v26.6.1`. Maps the SDK's new `0x6901` busy status word to a readable message; the refused and locked classification sets are untouched.

Corrects SDK and repo docs that were wrong against source, `btc-vault` or the firmware. The notable ones: the device *does* constrain the Pre-PegIn input sequence, ChallengeAssert is not claimer-signed, NoPayout is the challenger-wins leg of a dispute, Payout is the happy path, and `buildNoPayoutPsbt` does not cross-check prevouts.

## Why

Citations in the signer package are load-bearing under CLAUDE.md §7, and the app moved to 0.10.1. Separately, every runnable SDK quickstart example failed to type-check and several protocol descriptions contradicted btc-vault.

## Not in this PR

- `0x6901` preserving the loaded intent on the initial SIGN_PSBT (https://github.com/babylonlabs-io/babylon-toolkit/issues/2530).
- A host-side Pre-PegIn sequence assert in `fundPeginTransaction.ts`, §1 critical path (https://github.com/babylonlabs-io/babylon-toolkit/issues/2531).
- Whether the unused public `buildChallengeAssertPsbt` ships or goes (https://github.com/babylonlabs-io/babylon-toolkit/issues/2532).
- `DaemonStatus` missing btc-vault's `BabeSetupFailed`; deposit fee rate never validated as positive.
- Typedoc entry points for the vault-secrets, wasm and wots modules, whose frozen functions have no generated API page.
- add symbol names beside line numbers across the signer package's cross-package cites (pre-review #28)

## Pre-review

<details>
<summary>Pre-review: 28 findings · 26 fixed · 0 moot · 1 follow-up · 1 declined · 0 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | rawApdu.ts provenance comment misattributes 0xB009 to the base app and line 57 still cites a fw:sw.h the new block says does not exist | normal | fixed |
| 2 | primitives.md NoPayout note says the builder uses the VP's prevouts and that signDepositorGraph cross-checks them; the VP sends no prevouts and the service derives them from the parents | normal | fixed |
| 3 | managers.md end-to-end example no longer type-checks: seven parameters added to the snippets are undeclared in the anchoring declare-const block | normal | fixed |
| 4 | rawApdu.ts 0x6901 comment says the heartbeat path is present since v26.6.0; it already exists at v26.5.0 | normal | fixed |
| 5 | primitives.md keeps a trailing sentence that all these builders use extractPayoutSignature after the ChallengeAssert example stopped signing | normal | fixed |
| 6 | JSDoc in payout.ts and PayoutManager.ts still calls Payout the challenge path, and the generated API docs in this diff reproduce it, contradicting the corrected quickstart | normal | fixed |
| 7 | managers.md phase-returns table still omits depositTerms and authAnchorHex from the preparePegin result | normal | fixed |
| 8 | managers.md approval-wallet note omits that runDepositorPresignFlow also requires depositTerms, and the phase-5 call does not pass it | normal | fixed |
| 9 | peginFixture.ts provider.ts citations are off by one: the buildIntentFromTerms range excludes the txid flip it names, and the HARDENED-levels cite is stale | normal | fixed |
| 10 | peginFixture.ts leaf-builder range :174-259 overshoots; the builders end at :253 | normal | fixed |
| 11 | managers-advanced.md calls what Assert reveals the burn proof; btc-vault calls it the peg-out proof | normal | fixed |
| 12 | primitives.md says primitives are async; extractPayoutSignature is synchronous | normal | fixed |
| 13 | primitives.md intro still frames ChallengeAssert as a depositor-as-claimer builder, contradicting the corrected section | normal | fixed |
| 14 | primitives.md Pre-PegIn step omits the optional auth-anchor OP_RETURN that the glossary now names | normal | fixed |
| 15 | primitives.md NoPayout example's 688 prevout value reads as a protocol constant; it is a computed example value | normal | fixed |
| 16 | New Payout wording says only NoPayout blocks Payout; CouncilNoPayout does too | normal | fixed |
| 17 | The long Payout peg-out-paths paragraph is duplicated verbatim across two interface JSDocs | normal | fixed |
| 18 | rawApdu.ts cites sdk: paths but the signer defines no sdk: legend | normal | fixed |
| 19 | peginFixture.ts says a single-key shortcut leaf is 4 bytes short; it is 2 | normal | fixed |
| 20 | provider.ts cites init_global_state.c unprefixed, which its own legend reads as the vault app | normal | fixed |
| 21 | wasm/index.ts says each challenger has 3 ChallengeAssert transactions; btc-vault says 2 (X and Y) | normal | fixed |
| 22 | noPayout.ts still says prevouts come from the VP; the VP payload carries only tx_hex | normal | fixed |
| 23 | walletPolicy.ts leaves init_global_state.c unprefixed and unpinned on the line below a re-pinned cite | normal | fixed |
| 24 | payout.ts says four peg-out paths while the cited spec heading says three | normal | fixed |
| 25 | PayoutManager.ts uses a bare backtick for buildPayoutPsbt where the file links it everywhere else | normal | fixed |
| 26 | peginFixture.ts names only OP_NUMEQUAL; the encoder emits OP_NUMEQUALVERIFY for non-final groups | normal | fixed |
| 27 | 0x6f00 maps to a generic internal-error message while the vault app uses it as SW_BIP32_FAIL | normal | declined: the mapping is correct; a BIP32-specific message would misreport the generic use |
| 28 | peginFixture.ts cites bare provider.ts line numbers that have now drifted twice | normal | follow-up: add symbol names beside line numbers across the signer package's cross-package cites |

</details>


jeremy-babylonlabs (dismissed):

packages/babylon-ledger-vault-signer/src/rawApdu.ts
```diff
@@ -72,6 +74,12 @@ const STATUS_WORDS: Record<number, string> = {
   0x6a86: "The device rejected the instruction parameters",
   0x6a87: "The device rejected the payload length",
   0x6d00: "The running app does not support this instruction",
+  // SWO_COMMAND_NOT_ACCEPTED (`sdk:include/status_words.h:56`) — sent by the SDK
+  // IO layer before dispatch, so the app never ran the command: from the UX
+  // heartbeat when an APDU lands mid-approval (`sdk:io_legacy/src/os_io_legacy.c:132`,
+  // already in v26.6.0) and, new in v26.6.1, when a command arrives while a
+  // reply is still pending (`:400-412`).
+  0x6901: "The device was still busy with the previous request — try again",
```
jonybur: **The new 0x6901 message tells the user to retry, but a 0x6901 triggers the provider's pessimistic reset, so the retry fails with DEVICE_CEREMONY_INVALID.**

The new STATUS_WORDS entry maps 0x6901 to a message ending "try again", and the comment above it argues the word is emitted by the SDK IO layer before dispatch, so the app never ran the command. But classifyStatusWord leaves 0x6901 out of both SW_USER_REFUSED and SW_DEVICE_LOCKED, so it returns a plain LedgerDeviceError, and in the wallet-connector provider the keep-the-intent carve-out in signStaged only covers a locked-device error with preDispatch === true. A 0x6901 therefore falls through to the pessimistic reset that sets deviceState to idle and clears signedFingerprints. A user who follows the advice and retries then hits the DEVICE_CEREMONY_INVALID throw, "has no freshly derived context root ... Restart the flow from derivation", so the promised retry cannot succeed without redoing the derive and both approval screens. Either drop the "try again" from the message, or extend the carve-out to a pre-dispatch 0x6901 using the preDispatch flag the sign loop already computes.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
```diff
@@ -343,17 +343,20 @@ export function termsToIntent(terms: DepositTerms): IntentFixture {
...
 
 const PREPEGIN_INPUT_COUNT = 2;
 /** Distinct but otherwise arbitrary prevout hashes — the device never fetches them. */
 const PREPEGIN_PREVOUT_HASH_BASE = 0x40;
 /**
- * `_validate_prepegin` (fw `sign_psbt_validate.c:334-545` @ e2d0c45b) does not
- * constrain the sequence — it only hashes it into the txid it compares against
- * the intent (`:283-292`). So this must match what the toolkit's funding path
- * emits: bitcoinjs' DEFAULT_SEQUENCE (fundPeginTransaction.ts:148).
+ * The device REQUIRES `PSBT_IN_SEQUENCE` present and equal to SEQUENCE_FINAL on
```
jonybur, line 353: **Comment documents a fail-closed device requirement on Pre-PegIn input sequence that no host-side check enforces before device I/O.**

This comment now states that the device requires PSBT_IN_SEQUENCE present and equal to SEQUENCE_FINAL on every Pre-PegIn input and rejects anything else fail-closed, while conceding that the toolkit emits the value only as bitcoinjs' default. The host half of that is accurate: fundPeginTransaction.ts builds inputs with a bare tx.addInput(txHash, utxo.vout), and neither policyPsbt.ts nor signPsbtPrepare.ts mentions sequence at all, so nothing asserts the value before any device I/O. If the firmware behaves as documented, a funding path that signals RBF or otherwise deviates from the default dies on-device after the depositor has already stepped through the approval screens, which is exactly the failure the package's zero-I/O gates exist to prevent. Add an explicit sequence assertion alongside the other pre-I/O checks in augmentPsbtForWalletPolicy, or set the sequence explicitly when funding, and replace the "host-gap note in the Ledger KB" pointer with a citation a reader of this repository can follow.

packages/babylon-ts-sdk/docs/quickstart/managers.md
```diff
@@ -48,16 +48,21 @@ You do **not** generate or persist HTLC secrets. `preparePegin()` derives them d
...
     wotsPkHashes: Hex[],              // for `registerPeginOnChain.depositorWotsPkHash`
     htlcSecretHexes: string[],        // 64-char hex, no 0x; SHA256 → on-chain hashlock
+    authAnchorHex: string,
   },
 }
 ```
 
 The `secret` you pass to `activateVault()` is `0x${derivedSecrets.htlcSecretHexes[i]}`. The `hashlock` you pass to `registerPeginOnChain()` is `computeHashlock(secret)`.
 
+> **Approval-capable wallets** (e.g. Ledger) additionally require `depositTerms` on `signAndBroadcast()` and `runDepositorPresignFlow()`, and `quotedCommissionBps` on `registerPeginOnChain()`. Omitting them throws.
```
jonybur, line 64: **New callout says approval wallets require depositTerms and quotedCommissionBps, but the walkthrough in the same file omits both.**

The new callout states that approval-capable wallets additionally require depositTerms on signAndBroadcast() and quotedCommissionBps on registerPeginOnChain(), and that omitting them throws. Both throws are real: ensurePrePeginTermsApproval throws "This wallet requires approved deposit terms before signing the Pre-PegIn transaction" when the wallet exposes approveDepositTerms and no terms are passed, and assertQuotedCommissionPresentWhenRequired throws when quotedCommissionBps is undefined and the wallet supports deposit approval. The walkthrough further down the same page contradicts the callout: the registerPeginOnChain call passes no quotedCommissionBps and the signAndBroadcast call passes no depositTerms, while the runDepositorPresignFlow call does pass depositTerms with the comment "required for approval-capable wallets". An integrator copying the walkthrough for a Ledger wallet therefore throws at step 3 and again at step 4. Add both arguments to the example with the same comment, or say plainly that the walkthrough targets a wallet without deposit approval.

packages/babylon-ts-sdk/docs/get-started/README.md
```diff
@@ -251,11 +251,11 @@ Terms used throughout the docs.
...
 | **Vault UTXO** | The on-chain Bitcoin output produced by the PegIn tx that represents the live vault — the target of later payout transactions |
-| **HTLC** | Hash Time-Lock Contract — the Bitcoin script on the Pre-PegIn output. Locks the vault amount + depositor claim value + minimum peg-in fee. Spendable via the secret (claim path) or after a CSV timelock (refund path) |
-| **HTLC secret** | 32-byte preimage the depositor reveals on Ethereum to activate the vault. Persist it between registration and activation — losing it strands the vault until refund |
+| **HTLC** | Hash Time-Lock Contract — the Bitcoin script on the Pre-PegIn output. Locks the vault amount + depositor claim value + P2A anchor (240 sats on vault core 2/3, none on core 1) + minimum peg-in fee. Leaf 0 needs the secret **and** all-party signatures; leaf 1 is the refund after a CSV timelock |
+| **HTLC secret** | 32-byte preimage the depositor reveals on Ethereum to activate the vault. Re-derivable from the wallet at any time via `expandHashlockSecret`; nothing to persist |
```
jonybur: **Glossary says the HTLC secret needs nothing persisted, but derivation also requires the Pre-PegIn funding outpoints.**

The glossary entry now says the HTLC secret is "Re-derivable from the wallet at any time via expandHashlockSecret; nothing to persist". The secret is not derivable from the wallet alone: the root comes from deriveVaultRoot over a VaultContextInput whose fundingOutpoints field is documented as the funding outpoints of the Pre-PegIn transaction and must be non-empty, so the exact funding input set is part of the derivation. The recovery path in the vault service makes this concrete, throwing "Missing Pre-Pegin transaction; cannot recover HTLC secret" when the Pre-PegIn tx is absent and then parsing the outpoints out of it. Line 23 of this same file already words it correctly, "from the same wallet and on-chain context", so the glossary should match; a depositor who reads this line literally and discards the Pre-PegIn hex before registration has no way back to the secret.

jonybur (commented):
Comments inline.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts
```diff
@@ -210,11 +210,11 @@ export interface PayoutPsbtResult {
...
- * Payout is used in the **challenge path** when the claimer proves validity:
- * 1. Vault provider submits Claim transaction
- * 2. Challenge is raised during challenge period
- * 3. Claimer submits Assert transaction to prove validity
- * 4. Payout can be executed (references Assert tx)
+ * Payout is the happy path: PegIn → Claim → Assert → Payout.
+ * 1. The claimer submits Claim
+ * 2. The claimer submits Assert, revealing the peg-out proof
+ * 3. Payout can be executed once the Assert timelock matures (references Assert tx)
+ * A raised challenge diverts to ChallengeAssert and away from Payout.
```
gbarkhatov: **The one I'd most want fixed before merge — this corrected sentence is itself incomplete, on a registered §3 file.**

"A raised challenge diverts to ChallengeAssert and away from Payout." I checked this against the authoritative spec at `btc-vault/crates/vault/docs/btc-transactions-spec.md:29-33`, which enumerates **three** peg-out paths plus an emergency path, not two:

```
Happy path:                      Claim -> Assert -> Payout
Challenge path (claimer wins):   Claim -> Assert -> ChallengeAssert -> WronglyChallenged -> Payout
Challenge path (challenger wins): Claim -> Assert -> ChallengeAssert -> NoPayout
```

A raised challenge does not divert *away from* Payout — it opens a race that Payout can still win. The claimer-wins branch terminates **in Payout**, via WronglyChallenged. ChallengeAssert alone blocks nothing; only NoPayout does.

This matters more than a usual doc nit because the whole purpose of this PR is replacing wrong protocol statements with right ones, and this is the one place it replaces a wrong statement with a partly-wrong one — which then gets laundered through an approval and read as authoritative. It is also mirrored into `docs/api/primitives.md` and `docs/quickstart/primitives.md`, so the fix needs a `docs:clean` regeneration.

Suggested shape: name all three branches — a challenge opens the ChallengeAssert path; Payout still executes if the claimer wins via WronglyChallenged, and is blocked only by NoPayout.

packages/babylon-ledger-vault-signer/src/rawApdu.ts
gbarkhatov: **Suggestion (follow-up, not a regression) — 0x6901 now gets a message but still no classification, so a command the comment proves never ran still destroys the loaded intent.**

The new comment states 0x6901 is "sent by the SDK IO layer before dispatch, so the app never ran the command" — i.e. the on-device intent is *provably* intact. But `classifyStatusWord` at `rawApdu.ts:140-155` puts 0x6901 in neither `SW_USER_REFUSED` nor `SW_DEVICE_LOCKED`, so it falls through to a plain `LedgerDeviceError`. And `signStaged` at `provider.ts:1078-1119` excludes only `isLedgerSignPsbtAbortedError`, `lockedBeforeDispatch` and `standaloneRefund` from its reset guard — so a plain device error takes `this.deviceState = { phase: "idle" }` and clears `signedFingerprints`.

Net effect: the host discards an intent it has just documented as intact, costing the user a full re-derive plus re-approval on a hardware signer.

To be fair to the change: **this is not a regression.** 0x6901 was unmapped before and took the identical path, so nothing got worse — the PR establishes the premise without acting on it. The mechanism to fix it already exists and is used for 0x5515 (`preDispatch` → keep the mirror). Worth a follow-up issue rather than a change here.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -1065,8 +1065,8 @@ export class LedgerVaultProvider implements IBTCProvider {
    *
    * Locked-device words on the INITIAL SIGN_PSBT keep the intent because the app
    * never ran it: 0x5515 is sent only by the SDK IO layer before dispatch
-   * (`sdk:io_legacy/src/os_io_legacy.c:396-406`, inside the `io_exchange` receive
-   * loop `:243-245` that the base app reads from, `base:src/boilerplate/dispatcher.c:74`);
+   * (`sdk:io_legacy/src/os_io_legacy.c:416`, inside the `io_exchange` receive
+   * loop `:245-247` that the base app reads from, `base:src/boilerplate/dispatcher.c:74`);
```
gbarkhatov, line 1070: **Suggestion — the ledger-secure-sdk citations are the only claims in this PR that no reviewer can currently check, and I want to be straight that I could not verify them.**

Cited across here and `rawApdu.ts:77-81`: `sdk:include/status_words.h:56`, `sdk:io_legacy/src/os_io_legacy.c:132`, `:400-412`, `:416`, `:245-247`. What I found:

- `/Users/gbarkhatov/Documents/projects/ledger` contains only `app-babylon-vault`.
- `git show b0c0ac4d:.gitmodules` lists exactly two submodules — `bitcoin_app_base` and `ledger-app-ai-instructions`. Neither is the secure SDK.
- A `find` for `status_words.h` and `os_io_legacy.c` anywhere under the firmware checkout returns nothing.

So I'm relaying these as **unverified**, not as confirmed. One of my three review lanes did report them as verified "against LedgerHQ/ledger-secure-sdk at the pinned tags" without stating how, and I'm not passing that along as fact when no local source supports it.

Two things in your favour, though. The *version pairing* is independently solid — I checked the release tags, and all five 0.10.1 builds carry the `sdk_v26.6.1` suffix (`stax_1.10.1_0.10.1_sdk_v26.6.1`, `flex_1.6.1_0.10.1_sdk_v26.6.1`, and so on), so app 0.10.1 genuinely does build against secure-SDK v26.6.1. And this is still better provenance than the base text it replaces, which pointed at a floating image tag in `lite/Dockerfile`.

§7 asks that APDU citations name a source a reviewer can open. Naming the exact ledger-secure-sdk tag you read in the PR body would close that, cheaply.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
```diff
@@ -343,17 +343,20 @@ export function termsToIntent(terms: DepositTerms): IntentFixture {
...
+ * every Pre-PegIn input; absent or any other value is rejected fail-closed
+ * (fw `_compute_prepegin_txid`, `sign_psbt_validate.c:470-483` @ b0c0ac4d;
+ * `SEQUENCE_FINAL` = `vault_constants.h:136`). The toolkit's funding path emits
+ * it only as bitcoinjs' DEFAULT_SEQUENCE (`fundPeginTransaction.ts:155`,
+ * `bitcoinjs-lib@6.1.7 transaction.js:123,541`) — a library default, not an
+ * assertion. See the host-gap note in the Ledger KB.
```
gbarkhatov: **Suggestion — the newly documented fail-closed rule defers to an out-of-repo note, and nothing host-side enforces it.**

The comment is correct on every checkable part — I verified the device behaviour it describes at `b0c0ac4d:src/sign_psbt_validate.c` (`if (seq_rc != 4) return false; if (seq != SEQUENCE_FINAL) return false;`, with `SEQUENCE_FINAL 0xFFFFFFFFu` at `vault_constants.h:136`).

The issue is line 359: `See the host-gap note in the Ledger KB.` Every other reference in this file resolves to a file and line a reviewer can open; this one resolves to nothing, so the gap it defers to can't be tracked or closed. The file already cites issues by number (#2371, #2219, #2221) — one of those would work.

Separately, and the reason the deferral matters: the toolkit satisfies the device requirement only by inheriting bitcoinjs' `DEFAULT_SEQUENCE` at `fundPeginTransaction.ts:155` — a library default, not an assertion. So it is correct today by accident rather than by construction. A `sequence === SEQUENCE_FINAL` assert at the funding site would turn an opaque mid-ceremony `SW_INCORRECT_DATA` into a host-side failure at zero device I/O.

packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
```diff
@@ -76,30 +76,30 @@ const CPFP_ANCHOR_VALUE_SATS = 546;
...
- * use `+` (unsigned, provider.ts:205-212): `|` yields negative int32s that
+ * m/86'/1'/0'/0/0 — depositor_path(coin_type=1) (vault_client.py:307). Levels
+ * use `+` (unsigned, provider.ts:245-247): `|` yields negative int32s that
  * `assertBip86Path` rejects; the wire bytes are identical (`>>> 0` serializers).
  */
 export const DEPOSITOR_PATH: readonly number[] = [86 + HARDENED, 1 + HARDENED, 0 + HARDENED, 0, 0];
 const TESTNET_COIN_TYPE = 1;
 
-/** P2A anchor scriptPubKey `51 02 4e 73` (test_sign_psbt_validate.py:349). */
+/** P2A anchor scriptPubKey `51 02 4e 73` (test_sign_psbt_validate.py:2222). */
```
gbarkhatov: **Nit — the P2A citation is true but points at a txid-recompute helper rather than the builder this fixture says it replicates.**

The citation `test_sign_psbt_validate.py:2222` does check out — at that line the bytes appear as `buf += struct.pack('<Q', _PEGIN_ANCHOR_VALUE) + b'\x04\x51\x02\x4e\x73'  # P2A anchor`. But there they're embedded in a length-prefixed serialization, where the leading `\x04` is the script length rather than part of the script.

The canonical standalone constant is at `:544` — `p2a_spk = bytes([0x51, 0x02, 0x4e, 0x73])` — inside `_build_pegin_psbt`, which is the function this fixture's own header says it replicates, and the exact form the TS constant mirrors.

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/challengeAssert.ts
```diff
@@ -1,12 +1,15 @@
...
+ * is split across two single-input transactions — ChallengeAssertX (spends the
+ * challenger's ConnectorX Assert output) and ChallengeAssertY (spends
+ * ConnectorY). This builder handles one such transaction.
+ *
+ * The protocol collects no claimer signature on ChallengeAssert — the
+ * challenger signs it at broadcast (btc-vault `tx_graph/challenger.rs`). This
+ * builder exists for tooling that reconstructs the graph; it is not on the
+ * depositor presign path (`signDepositorGraph` signs Payout + NoPayout only).
  *
  * @module primitives/psbt/challengeAssert
```
gbarkhatov, line 14: **Suggestion (follow-up) — accurate new text, but it converts a silent gap into a documented one.**

The correction is right and an improvement on the base text. Two lanes independently grepped all of `packages/` and `services/` for `buildChallengeAssertPsbt` and found exactly three hits: this definition, the public re-export in `primitives/index.ts`, and the doc line. No caller, no test.

So after this PR the tree contains a published primitive that emits a signing-ready PSBT for a dispute-path transaction the depositor must never sign, now explicitly documented as never claimer-signed. It's public surface, so it isn't dead code under the zero-dead-code policy, and the new rationale ("exists for tooling that reconstructs the graph") is a legitimate reason to keep it.

Worth a follow-up issue deciding whether it ships or goes, rather than a change in this PR.

gbarkhatov (approved):
Reviewed across three independent lanes, then verified every claim against the actual sources myself before posting — for a PR whose entire risk surface is citation truth, a review that doesn't open the firmware is worth nothing.

**No §7 merge-blockers.** The load-bearing claim — that the device *does* constrain the Pre-PegIn input sequence, reversing the old comment — is true, and I confirmed it at the pinned commit rather than taking it on trust:

- `b0c0ac4d` exists; its Makefile is `APPVERSION_M=0 / N=10 / P=1`, so the 0.10.1 pin is correct.
- `git show b0c0ac4d:src/sign_psbt_validate.c` contains `if (seq_rc != 4) return false; if (seq != SEQUENCE_FINAL) return false;`. The old comment was wrong; the correction is right.

Six findings below, all suggestions. If you fix one before merge, make it the first.

### Verified clean — the things this PR could have broken and didn't

- **No status word was reclassified.** I checked question-by-question: `SW_USER_REFUSED` does not appear in the diff *at all*, and `SW_DEVICE_LOCKED` appears exactly once — as a hunk-header context line (`@@ -54,15 +54,17 @@ const SW_DEVICE_LOCKED = ...`), not a changed line. 0x6901 goes from "unmapped, raw-hex error" to "mapped message, same error class". Only the string changed. The PR's claim is literally true.
- **§7's regeneration requirement is not triggered.** No `scripts/gen_*.py`, no `scripts/requirements.txt`, and no bytes under `vendor/ledger-bitcoin/__tests__/vectors/` are in the diff — only the vendor `README.md`. No reviewer regeneration is owed here, and no TLV encoding or APDU layout moved.
- **§4's frozen files are comment-only.** `babylon-tbv-rust-wasm/src/index.ts:229` and `ts-sdk/src/tbv/core/wasm/index.ts:184` change only `*`-prefixed JSDoc adding the `p2aAnchorValue` term. No expander touched, no `VAULT_WASM_COMMIT` bump — the hard-fork rule does not apply.
- **`docs/api/` is genuinely generated, not hand-edited.** Every hunk is verbatim identical to a JSDoc edit elsewhere in the same PR — which is why `verify` passed, and the inverse check finds no orphaned prose that the next regeneration would clobber.
- **CLAUDE.md and SECURITY.md weaken nothing.** The CLAUDE.md delta is confined to the Key Packages bullet list; §§1-9 are untouched. SECURITY.md changes only "five packages" to "six", adding `ledger-vault-signer`. This was the most severe thing this PR could plausibly have done, and it did not.
- **The PR fixes an off-by-two it inherited:** `SW_BAD_CPFP_ANCHOR` at `:50` and `SW_CAP_EXCEEDED` at `:52`, where the base text cited `:50` for both.

### Refuted

- **"A doc correction is wrong about `buildNoPayoutPsbt`."** Refuted — `noPayout.ts` takes `prevouts` verbatim with only a presence check and never sees the Assert tx, so the new warning is accurate. The *old* text ("the builders cross-check") was false, on a path where a wrong `witnessUtxo.value` signs the wrong sighash. That correction is the most valuable line in the PR.

### A note on the existing approval

The standing approval carries an empty body, so for a PR whose whole risk surface is whether citations match their sources, it records no verification at all. That's the gap this review is trying to fill rather than duplicate.

Everything below is a suggestion — your call as author.

gbarkhatov (dismissed):

packages/babylon-ledger-vault-signer/src/rawApdu.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
jrwbabylonlab: fixed; not planning — tracked in #2531

packages/babylon-ts-sdk/docs/quickstart/managers.md
jrwbabylonlab: fixed

packages/babylon-ts-sdk/docs/get-started/README.md
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/payout.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/rawApdu.ts
jrwbabylonlab: not planning — tracked in #2530

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
jrwbabylonlab: fixed

packages/babylon-ledger-vault-signer/src/__tests__/e2e/peginFixture.ts
jrwbabylonlab: fixed

packages/babylon-ts-sdk/src/tbv/core/primitives/psbt/challengeAssert.ts
jrwbabylonlab: not planning — tracked in #2532

gbarkhatov (dismissed):

jonybur (commented):
Comments inline, plus:

- `packages/babylon-ts-sdk/docs/quickstart/managers-advanced.md:183` - **Multi-vault example still loops registerPeginOnChain and omits quotedCommissionBps/depositTerms, contradicting the new batch advice and throwing on approval-capable wallets.**
  
  The rules list now says to prefer `registerPeginBatchOnChain()`, but the example right below still calls `registerPeginOnChain()` once per vault. It also passes no `quotedCommissionBps` to registration and no `depositTerms` to `signAndBroadcast()`, and the destructuring at line 169 drops `result.depositTerms` entirely. For a wallet that supports deposit approval (e.g. Ledger), `assertQuotedCommissionPresentWhenRequired` in `pegin-registration-client.ts` and the Pre-PegIn approval check in `prePeginApproval.ts` both throw, so an integrator copying this example fails at registration and again at broadcast. The same gap was fixed in `managers.md` in this PR but not here. Switch the example to `registerPeginBatchOnChain()` and pass `quotedCommissionBps` and `depositTerms: result.depositTerms`, with the same approval-wallet comment used in `managers.md`.

packages/babylon-ts-sdk/docs/quickstart/primitives.md
```diff
@@ -125,17 +141,20 @@ const signature = extractPayoutSignature(signedPsbtHex, depositorBtcPubkey);
...
+1. **Payout** (1 per vault) — `buildPayoutPsbt` with `claimerBtcPubkey` set to the depositor's key.
+2. **NoPayout** (1 per challenger) — the challenger-wins leg of a dispute. 3 inputs (Assert:0, ChallengeAssertX:0, ChallengeAssertY:0), 1 output.
 
-1. **Payout** (1 per vault) — sign with `buildPayoutPsbt` (section 2 above) with `claimerBtcPubkey` set to the depositor's own key; it validates the output layout and the implicit-fee band before returning the PSBT
-2. **NoPayout** (1 per challenger) — covers the case where the vault expires without a successful claim
-3. **ChallengeAssert** (1 per challenger, with 3 inputs) — covers the challenge-assert spending paths
+> **ChallengeAssert is not signed by the claimer.** NoPayout references specific ChallengeAssert txids, so a challenger who broadcasts a different one cannot execute NoPayout (btc-vault `tx_graph/challenger.rs`).
 
 The vault provider supplies the unsigned transaction hexes. The depositor must
 also supply the parent transactions (peg-in tx for Payout, Assert tx for
```
jonybur: **Rewritten trust note lists only the peg-in tx as Payout's trusted parent, but buildPayoutPsbt also validates input 1 against the Assert tx.**

The rewritten paragraph says the depositor must supply the parent transactions from a trusted source: the peg-in tx for Payout and the Assert tx for ChallengeAssert. But `buildPayoutPsbt` also resolves Payout input 1 against `assertTxHex` (`payout.ts:304-310`), and the example on the same page passes it. A reader following this note could take the Assert tx for Payout straight from the vault provider without treating it as a trust input, which weakens the prevout check that the note itself says protects against a malicious VP. Say "peg-in and Assert tx for Payout".

jrwbabylonlab: fixed

gbarkhatov (dismissed):

gbarkhatov (dismissed):

gbarkhatov (approved):

packages/babylon-ts-sdk/docs/quickstart/managers-advanced.md
```diff
@@ -90,15 +90,22 @@ const payoutManager = new PayoutManager({
...
   peginTxHex: "...",
-  assertTxHex: "...",                        // Assert transaction (challenge path)
+  assertTxHex: "...",                        // Assert transaction the payout spends
   depositorBtcPubkey: "...",
   vaultProviderBtcPubkey: "...",
   vaultKeeperBtcPubkeys: [/* … */],
   universalChallengerBtcPubkeys: [/* … */],
   timelockPegin: 100,                        // from vault-pinned offchain params
+  timelockAssert: 144,
+  vaultCoreVersion,
```
jonybur, line 100: **The rewritten signPayoutTransaction example uses six identifiers the page never declares, so the snippet does not type-check.**

This snippet gains `vaultCoreVersion`, `commissionBps`, `councilMembers`, `councilQuorum`, `vkClaimerPayoutScriptPubKeys` and `vpCommissionScriptPubKey` as bare identifiers, but the only declarations on this page are at lines 20, 21, 77 and 147-159, none of which cover them (`protocolFeeRate` was already bare before). A reader who copies the block gets six unresolved names, which is the same defect this change repaired on the sibling page: `managers.md` declares every one of them at lines 145-151. Nothing catches it either, since the only docs tasks are typedoc generation and validation over source, and no markdown snippet is type-checked. The field names themselves match `SignPayoutParams`; add a `declare const` block above the snippet mirroring `managers.md`.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -3,9 +3,10 @@
  *
  * Citation legend — `base:` = LedgerHQ/app-bitcoin `baseapp` @ `e400d8d8` (the
  * vault app's submodule pin, paths under `src/`); `sdk:` = LedgerHQ/ledger-secure-sdk
- * @ `v26.6.0` — the SDK the app's CI image compiles in (`ledger-app-builder-lite:latest`,
- * `lite/Dockerfile:3-7` @ 786151a7f; run 32839601193 for eacb873b6); unprefixed `.c`
- * paths are LedgerHQ/app-babylon-vault @ `eacb873b6`.
+ * @ tag `v26.6.1` of https://github.com/LedgerHQ/ledger-secure-sdk — the SDK the app
+ * builds against (all five 0.10.1 release tags record `SDK version: v26.6.1`);
+ * unprefixed `.c` paths are
```
jonybur, line 8: **Bumping the legend's unprefixed-.c clause to b0c0ac4d re-attributes citations in this file that were never re-read at that revision.**

The legend's "unprefixed `.c` paths" clause moves from `eacb873b6` to `b0c0ac4d`, and that re-anchors every unprefixed citation in the file, including three this change never revisits: `sign_custom_inputs.c` `:184` and `:401` at line 753, and `vault_context.c:44-47` at line 1113. The same diff shows how far the firmware moved between those revisions — `vault_script.c:711-713` became `:766-767` and `sign_psbt_validate.c:3205-3213` became `:3573-3578` — so untouched line numbers are likely stale while now reading as verified at a revision where nobody opened them. That is the failure the citation rule exists to prevent: a reviewer follows the legend, opens the line, and lands somewhere unrelated. Either re-read those three cites at `b0c0ac4d` and correct them, or note in the PR which unprefixed cites were re-checked and which were only inherited.

packages/babylon-ledger-vault-signer/src/walletPolicy.ts
```diff
@@ -4,8 +4,9 @@
  * The base app signs internal (key-path) inputs only when a wallet policy is
  * present (`bitcoin_app_base/src/handler/sign_psbt.c:142-148` @ e400d8d8);
  * without one a PoP returns SW_OK with no signature
- * (`app-babylon-vault/src/sign_custom_inputs.c:101-107` @ 4decf822). A default
- * policy needs no registration: empty name (`init_global_state.c:238-242`),
+ * (`app-babylon-vault/src/sign_custom_inputs.c:101-115` @ b0c0ac4d). A default
+ * policy needs no registration: empty name
+ * (`bitcoin_app_base/src/handler/sign_psbt/init_global_state.c:238-242` @ e400d8d8),
  * template `tr(@0/**)`, one key info `[fpr/86'/coin'/account']xpub`
  * (`tests/test_screen7_pop.py:135-142`), hmac 32×00. The id is
```
jonybur, line 11: **The two test_screen7_pop.py citations carry no repository and no revision, and the provider legend binds only .c paths.**

Lines 5, 7 and 9 of this JSDoc each name a repository and a commit, with 7 and 9 re-pinned here, but the key-info citation on line 11 is still a bare `tests/test_screen7_pop.py:135-142` with neither. The identical unpinned range sits at `provider.ts:254`, where the legend rewritten in this change binds only unprefixed `.c` paths to `app-babylon-vault`, so an unprefixed `.py` path resolves to no revision at all. Python line numbers demonstrably move between these revisions — this change shifts `test_sign_psbt_validate.py:63` to `:85` and `:591` to `:823` — yet these two are the only firmware references left at their old numbers, so a reader has no way to check them. Give them the same `app-babylon-vault/tests/...` @ `b0c0ac4d` form as their neighbours, or extend the provider legend to cover unprefixed `.py` paths.

jonybur (commented):
Comments inline.

PR #2525, opened 2026-09-16
chore(ci): require a pre-review record on PRs

gbarkhatov:
chore(ci): require a pre-review record on PRs

## What

- New CI job `pre-review-check`: fails a PR to `main` whose description has no `/pre-review` record taken on the PR's branch. It skips PRs with the `skip-pre-review` label, drafts and bots, and keeps one comment on the PR explaining what to do. It runs on `pull_request_target` from `main`'s copy and never executes PR code, so a PR cannot edit the check it is judged by; it therefore starts checking PRs once merged. A separate read-only workflow tests the check's scripts on the PR that changes them.
- `scripts/pre-review/snapshot.mjs`: `record` (used by the skill to snapshot the change) and `check` (used by CI), with tests.
- Local nudges: a husky pre-push warning (never blocks) and an opt-in Claude Code hook that stops `gh pr create` on a branch without a record.
- Docs: a new Enforcement section in `docs/pre-review.md`; `SECURITY.md` gains an entry for the workflow (its `pull_request_target` trust boundary and known bypasses); `CLAUDE.md`, `SKILL.md` and `formats.md` updated.

## Why

Nothing forced anyone to run `/pre-review`. A local hook can always be skipped, so the only real enforcement is a CI check made required.

## Approaches

- **Running `/pre-review` in a git hook:** rejected. The skill is interactive, a full run takes minutes and hundreds of thousands of tokens, and `--no-verify` bypasses it.
- **Requiring the code to match the reviewed snapshot exactly:** rejected for now. It only enforced "commit before you fix": review, commit, fix passed while review, fix, commit failed. CI checks that the review ran on the branch; changes after the review are the author's to own.

## Not in this PR

- Creating the `skip-pre-review` label and adding `pre-review-check` as a required check in the `main` ruleset (manual, after merge). A PR could still add its own workflow with a job of the same name; watch for that in review.
- A stricter check that compares the record with the code and requires a re-run past 150 changed lines or a critical path (TODO on `checkSnapshot`).

## Pre-review

<details>
<summary>Pre-review: 33 findings · 24 fixed · 0 moot · 0 follow-up · 8 declined · 1 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | A quoted or malformed snapshot line crashes check: red job, no comment, stale comment left | normal | fixed |
| 2 | Claude hook ignores tool_input.command, relies on the settings if filter, and denies gh pr create --draft | normal | fixed |
| 3 | Sticky comment is found by marker regardless of author | normal | fixed |
| 4 | A skipped result rewrites a failure comment as Resolved | normal | fixed |
| 5 | The pre-push hook can block a push despite "never blocks" | normal | fixed |
| 6 | SKILL.md copy instruction does not map the script's JSON keys to the snapshot fields | normal | fixed |
| 7 | Script tests never delete their temporary repositories | normal | fixed |
| 8 | Status and label strings are duplicated between the workflow and the script | normal | declined: cross-language boundary; an unknown status already fails closed |
| 9 | record follows symlinks and aborts on a directory path | normal | declined: no symlinks or gitlinks tracked; digest not compared in CI yet |
| 10 | Local hooks check the local branch name while CI checks the PR head ref | normal | declined: rare push form; local hooks are advisory |
| 11 | Test gaps: CLI exit codes, pre-push skip rules, detached HEAD | normal | declined: check CLI runs for real in CI; fixes add the key tests |
| 12 | A PR can disable its own check: the workflow and scripts run from the PR head | normal | fixed |
| 13 | Distinct branches feat/x and feat__x share one local record file | normal | declined: existing skill key format; branch= comparison reports the collision |
| 14 | Changes to the check's scripts get no CI test run before merge | normal | fixed |
| 15 | A PR body over 128 KiB in bytes fails the job before any script runs | normal | fixed |
| 16 | Docs overclaim that a PR cannot change its own check | normal | fixed |
| 17 | A lone carriage return in a snapshot line can break out of the malformed comment's code fence | normal | fixed |
| 18 | Hook's --draft=true and single & separator have no tests | normal | fixed |
| 19 | Docs name only one of the hook's false allows | normal | fixed |
| 20 | Hook docs omit --draft=true from the flags that read as a draft | normal | fixed |
| 21 | The other-branch comment echoed the author's branch= value inside an inline code span | normal | fixed |
| 22 | Malformed-line test assertions matched the echoed input, passing whichever validation threw | normal | fixed |
| 23 | A tilde-only branch= value closes the other-branch comment's fence | normal | declined: cosmetic; no author text can render, and git names cannot contain ~ |
| 24 | First pull_request_target workflow and a bypassable merge gate landed without the SECURITY.md update | normal | fixed |
| 25 | Claude hook denied stacked PRs and PRs into other repositories, saying CI would fail | normal | fixed |
| 26 | recordBlobs test passed even if content was no longer written to the object database | normal | fixed |
| 27 | Hook's base/repo scan allows PRs into main it should stop, and the docs don't say so | normal | fixed |
| 28 | The hook's deny path for an explicit main base is only tested with --base=main | normal | fixed |
| 29 | SECURITY.md entry omits the bot-author skip and calls the gate advisory | normal | fixed |
| 30 | Hook header comment and JSDoc still say only PRs into another repository are allowed | normal | open |
| 31 | Docs wording gaps: -R naming this repository, short flag forms, SECURITY.md reflow | normal | fixed |
| 32 | An empty --base value reads as a stacked PR | normal | declined: unverified gh behaviour; advisory hook, CI still checks the PR |
| 33 | --repo=owner/name is not recognised, though the docs say any --repo is allowed | normal | declined: fails safe (a false deny); the spaced form works; advisory hook |

</details>


.github/workflows/pre-review-check.yml
```diff
@@ -0,0 +1,167 @@
...
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0 https://github.com/actions/setup-node/releases/tag/v4.4.0
+        with:
+          node-version: '24.2.0'
+
+      # A wrong parsing rule would pass unreviewed PRs or fail every PR, and
+      # neither is visible from a green or red check alone.
+      - name: Test the snapshot scripts
+        run: node --test "scripts/pre-review/__tests__/*.test.mjs"
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in 412576635. Running only the scripts from the base revision would not have been enough: on `pull_request` the workflow YAML also comes from the PR, so a PR could just as well edit the job to `exit 0`. The check now runs on `pull_request_target`, so GitHub uses the workflow file and `scripts/pre-review/` from `main`, and no step checks out or executes PR code (everything is read from the event payload, the body from `$GITHUB_EVENT_PATH`). The scripts are tested on the PR that changes them by a separate read-only `pull_request` workflow, `pre-review-scripts.yml`. Remaining limit, documented in `docs/pre-review.md`: a PR could add its own workflow with a same-named job, which shows in its diff.

.github/workflows/pre-review-check.yml
```diff
@@ -0,0 +1,167 @@
...
+          else
+            printf '%s' "$PR_BODY" > "$RUNNER_TEMP/pr-body.md"
+            node scripts/pre-review/snapshot.mjs check \
+              --body-file "$RUNNER_TEMP/pr-body.md" \
+              --branch "$HEAD_REF" > "$result_file"
+          fi
+
+          cat "$result_file"
+          status=$(node -e 'console.log(JSON.parse(require("fs").readFileSync(process.argv[1], "utf8")).status)' "$result_file")
+          echo "status=$status" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
gbarkhatov: Fixed in 412576635. `checkSnapshot` now returns a `malformed` status (with the offending line) instead of throwing, so the check step completes, the PR gets a "Pre-review record is malformed" comment, and the final step fails the job. The docs table has a `malformed` row, and a test covers a quoted template line. The snapshot line pattern also excludes `\r`, so the quoted line cannot break out of the comment's code fence.

scripts/pre-review/snapshot.mjs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+/**
+ * The local hooks' check: the description `/pre-review` keeps for a branch
+ * (`/` in the name becomes `__`). `missing` when it never ran there.
+ */
+export function checkBranchRecord({ branch, cwd }) {
+  const recordPath = path.join(cwd, STATE_DIRECTORY_PREFIX, `${branch.replaceAll("/", "__")}.md`);
+  if (!fs.existsSync(recordPath)) return { status: CHECK_STATUS.MISSING };
+  return checkSnapshot({ text: fs.readFileSync(recordPath, "utf8"), branch });
```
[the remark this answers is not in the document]
gbarkhatov: Not changing this one. The `/` to `__` key is the existing `/pre-review` state-file format (`.pre-review/<key>.json` and `.md`), and changing it would orphan every existing local record. A collision cannot pass silently: the local hooks compare the record's `branch=` with the actual branch and report `other-branch`, and the skill itself asks before touching state whose branch differs. The hooks are advisory; CI compares `branch=` with the PR head ref, which is unaffected.

.github/workflows/pre-review-check.yml
```diff
@@ -0,0 +1,190 @@
...
+            const failures = {
+              missing: [
+                '### Pre-review record missing',
+                '',
+                `This description has no \`/pre-review\` record. Run \`/pre-review\` locally, decide its findings, and use \`PR.md\` as the description. ${skipHint} See ${doc}.`,
+              ],
+              'other-branch': [
+                '### Pre-review record is from another branch',
+                '',
+                `The record in this description was taken on \`${result.recorded?.branch}\`, not on this PR's branch. ${rerun} See ${doc}.`,
```
jonybur: **The other-branch comment puts the author-controlled branch= value into inline backticks without escaping, so the bot's comment can render arbitrary Markdown.**

For an `other-branch` result the comment builds `` `${result.recorded?.branch}` ``, and `parseSnapshots` only requires that `branch=` is non-empty and has no space or line break. A description line such as ``branch=x`@​org/team`` or ``branch=x``` closes the code span, and the rest renders as live Markdown (mentions, images, links) in a comment posted by github-actions[bot]. The malformed-record message avoids this by quoting inside a `~~~` fence, but this path got no such guard. Either restrict `branch=` in `parseSnapshots` to git-safe characters (throwing `MalformedSnapshotError` otherwise), or escape backticks / print the value inside the same fence.

scripts/pre-review/__tests__/snapshot.test.mjs
```diff
@@ -0,0 +1,275 @@
...
+});
+
+test("parseSnapshots counts the same line pasted twice as one record", () => {
+  const line = snapshotLine();
+
+  assert.equal(parseSnapshots(`${line}\n\n${line}`).length, 1);
+});
+
+test("parseSnapshots rejects a snapshot line with a truncated digest", () => {
+  assert.throws(() => parseSnapshots(snapshotLine({ sha256: "abc123" })), /files-sha256/);
```
jonybur: **The malformed-line tests match patterns that also appear in the echoed input line, so they pass no matter which validation rejected it.**

Every `MalformedSnapshotError` message ends with the full snapshot line (`: ${match[0]}`). The assertion `/files-sha256/` here always matches because the input line itself contains `files-sha256=abc123`, and the same goes for `/files=<count>/` at line 128, where the quoted template literally contains `files=<count>`. If the digest or count validation broke and a different check threw, both tests would still pass. Assert on the specific message text instead, e.g. `/no valid files-sha256=/` and `/no valid files=<count>/`.

jonybur (commented):
Comments inline.

.github/workflows/pre-review-check.yml
gbarkhatov: Fixed in c762097b4. The `other-branch` comment no longer puts the recorded branch in a code span: it says the record was taken on another branch and shows the value inside a `~~~` fence, the same way the malformed comment does. I did not restrict `branch=` to a character set, because git itself accepts both characters in a branch name (`git check-ref-format --branch 'feat/a`b@c'` succeeds), so a whitelist would mark a real branch as malformed and leave its author unable to pass without renaming. The fence cannot be closed by author text: the snapshot line pattern excludes `\r` and `\n`, and a `~~~` in the middle of a line is not a fence. The one remaining edge, a value made only of tildes, closes the fence early but leaves nothing after it to render.

scripts/pre-review/__tests__/snapshot.test.mjs
gbarkhatov: Fixed in c762097b4, as suggested. The three malformed-line assertions now match text that only the error message contains: `/no valid files-sha256=<hex>/`, `/no branch=<name>/` and `/no valid files=<count>/`. To check it, I ran the parser with the count validation broken on a line whose digest is also truncated: the old `/files-sha256/` still matched (through the quoted line), while the new assertion does not.

.github/workflows/pre-review-check.yml
```diff
@@ -0,0 +1,197 @@
...
+# its diff). That event carries a write token, which is safe only because no
+# step checks out or runs the PR's code: everything is read from the event
+# payload. Keep it so. The scripts are tested on the PR that changes them by
+# pre-review-scripts.yml.
+#
+# `edited` re-runs the check when the author pastes an updated PR.md, which
+# changes only the body. `labeled`/`unlabeled` re-run it for skip-pre-review.
+# `synchronize` is still needed: a required check must report on every head.
+on:
+  pull_request_target:
```
jonybur, line 21: **First pull_request_target workflow and a new bypassable merge gate land without the SECURITY.md update the security model requires.**

This is the only workflow in the repo that runs on `pull_request_target`, which gives it a write token on fork PRs. It is also meant to become a required merge gate that can be bypassed on purpose (the skip label, a hand-typed snapshot line, or a same-named job added by a PR). SECURITY.md says a change that alters trust boundaries MUST update that file in the same PR. For a new gate, it also asks whether the gate is enforced somewhere the user cannot edit, and says to note it there if not. The rule 'never check out or run PR code on this event' currently lives only in the workflow comment and docs/pre-review.md. Please add an entry for this workflow under SECURITY.md's Workflows section covering the no-checkout rule and the known bypasses.

jonybur (commented):
Comments inline.

scripts/pre-review/claude-pr-create-hook.mjs
```diff
@@ -0,0 +1,94 @@
...
+ * Whether a Bash command opens a non-draft PR. Best effort, not a shell
+ * parser: the draft flag is looked for anywhere after `gh pr create`, so it is
+ * still found after a multi-line `--body`, and a flag quoted inside a title
+ * reads as a draft. CI is the enforcement; this only saves a round trip.
+ */
+function opensReadyPullRequest(command) {
+  const segments = command.split(COMMAND_SEPARATOR_PATTERN);
+  const createIndex = segments.findIndex((segment) => PR_CREATE_PATTERN.test(segment.trim()));
+  if (createIndex === -1) return false;
+  const tokens = segments.slice(createIndex).join(" ").split(/\s+/);
```
jonybur, line 75: **Hook denies gh pr create for PRs not targeting main (e.g. stacked PRs) and wrongly says CI would fail.**

`opensReadyPullRequest` only looks for draft flags. It ignores `--base`/`-B` and `--repo`/`-R`. The CI job only runs for PRs into `main` (`branches: [main]`), but a stacked PR opened with `gh pr create --base feat/parent` on a branch without a record is still denied. The deny reason says the PR's pre-review-check would fail, which is not true for that PR. Scan the same tokens for a `--base`/`-B` value other than `main`, or a `--repo` flag, and allow the command in those cases.

scripts/pre-review/__tests__/snapshot.test.mjs
```diff
@@ -0,0 +1,280 @@
...
+  repo.write("src/added.ts", "added\n");
+  fs.rmSync(path.join(repo.dir, "removed.ts"));
+  repo.write("PR.md", "description\n");
+  repo.write(".pre-review/feat__x.md", "description\n");
+
+  const blobs = recordBlobs(changedPathsInWorktree("main", repo.dir), repo.dir);
+
+  assert.deepEqual([...blobs.keys()].sort(), ["kept.ts", "removed.ts", "src/added.ts"]);
+  assert.equal(blobs.get("removed.ts"), DELETED);
+  assert.equal(blobs.get("src/added.ts"), repo.git("hash-object", "src/added.ts"));
```
jonybur, line 82: **recordBlobs test compares with hash-object without -w, so it passes even if content is no longer written to the object database.**

The test compares the recorded id with `git hash-object src/added.ts`, which runs without `-w`. The id is the same whether or not `recordBlobs` writes the object, so removing `-w` from snapshot.mjs line 96 would not fail this test. Later runs rely on the object existing: the skill runs `git cat-file -e <blob>` and treats a missing object as pruned, so a missing object quietly turns a small blob diff into a full re-review. Add an assertion such as `repo.git("cat-file", "-e", blobs.get("src/added.ts"))` to pin that behaviour.

.github/workflows/pre-review-check.yml
gbarkhatov: Added in 03cb50cf0: a `pre-review-check.yml` entry under `SECURITY.md` > `### Workflows`. It records that this is the only `pull_request_target` workflow, so it runs with a write token (fork PRs included) from `main`'s copy of the workflow and `scripts/pre-review/`, and it states the rule: never check out or run the PR's code there (inputs come from the event payload, the description from `$GITHUB_EVENT_PATH`, and author text reaches the bot comment only inside a fenced block). For the gate question it says the check only gates merges once a ruleset makes it required, that the ruleset lives outside the repo and must be verified, and that the gate is bypassable by design: the `skip-pre-review` label, a hand-written snapshot line (proves a record is present, not that a review ran), a PR opened by a bot account (skipped by the job), and a PR adding its own workflow with a same-named job (visible in its diff). It also notes `pre-review-scripts.yml` is the only pre-merge test of the scripts.

scripts/pre-review/claude-pr-create-hook.mjs
gbarkhatov: Fixed in 03cb50cf0. The hook now only stops PRs the CI job would check: a `--base`, `-B` or `--base=` value other than `main` is let through (surrounding quotes are stripped first, so `-B "main"` is still denied), and so is any `--repo`/`-R`. Tests cover a stacked PR (`--base feat/parent`, allowed), `--base=main` and `-B "main"` (denied), and `-R other/repo` (allowed). Two limits, documented in `docs/pre-review.md` and the hook comment: any `-R` passes, even one naming this repository (whose PR CI still checks), since comparing it with the remote is more than an advisory hook needs; and `--repo=owner/name` (the `=` form) is not recognised, which is a false deny, not a bypass. Flag detection stays a best-effort token scan; CI remains the enforcement.

scripts/pre-review/__tests__/snapshot.test.mjs
gbarkhatov: Fixed in 03cb50cf0, as suggested: the test now also asserts `git cat-file -e` on the recorded blob, with a comment on why (a later run diffs against that object and treats a missing one as pruned). To confirm it pins the behaviour, I ran the test against a copy of `snapshot.mjs` with `-w` removed from `recordBlobs`: it fails with `Command failed: git cat-file -e d5f7fc3…`, while it passes on the real module.

jrwbabylonlab (approved):

PR #2528, opened 2026-09-16
fix(vault): show a year of candles and expose the chart reset control

### LINKED ISSUE #2417: Liquidation Analysis preview: one-year timeline and a reachable reset control
## Context

The Liquidation Analysis preview embedded in the borrow flow, rendered by `services/vault/src/components/simple/RiskSection/LiquidationAnalysisSection.tsx`.

### 1. Show a year of candles

**Now:** the visible window is 60 daily candles — roughly two months. `TIMELINE_VISIBLE_CANDLES = 60` is defined at `LiquidationAnalysisSection.tsx:49` and again, independently, at `services/vault/src/components/pages/Liquidations/index.tsx:60`. The fetch itself caps at `CANDLE_LIMIT = 180` with `CANDLE_INTERVAL = "day_1"` (`services/vault/src/hooks/useBtcPriceCandles.ts:25,31`), both module constants, not props.
**Expected:** one year.
**Where:** raising the visible window past 180 also requires raising `CANDLE_LIMIT` — 365 daily candles are not fetched today. Fold the two duplicated constants into one while you are there.

### 2. The reset control never renders

**Now:** the zoom controls, including the `⟲` button with `aria-label="Reset view"`, only render when `zoomEnabled = Boolean(interactions?.zoom) && candles.length > MIN_ZOOM_CANDLES` (`services/vault/src/components/simple/Timeline/Timeline.tsx:152,363-382`). No call site passes `zoom`: the preview passes no `interactions` prop at all (`LiquidationAnalysisSection.tsx:137-148`), and the full page passes `interactions={{ crosshair: true, pan: true }}` (`Liquidations/index.tsx:349`).
**Expected:** the control is reachable. Note this is not preview-specific — "Reset view" renders nowhere in the app today.
**Where:** `Timeline.tsx:152`, `LiquidationAnalysisSection.tsx:137-148`, `Liquidations/index.tsx:349`

Not this: the simulator's own "Reset" (`COPY.liquidations.reset`, `SimulationToolbar.tsx:152`) is deliberately absent from the preview, which is read-only by design — see `LiquidationAnalysisSection.test.tsx:150-156`.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1787811491261369)

> *Liquidation Analysis preview*
> • here candles look strange, let's show a timeline of 1 year.
> • "reset" button is not visible

</details>

jeremy-babylonlabs:
Closes #2417

## What changed

- The Timeline on the Liquidation Analysis page now shows one year of daily candles at rest (365) and fetches two years (730) so pan still has room. The visible count lives in one exported constant, `TIMELINE_VISIBLE_CANDLES`, next to the fetch limit that derives from it, instead of two private copies that had to agree.
- Zoom is enabled on the Liquidation Analysis page, so core-ui's Reset view control renders there. Before this the page allowed pan only, and the reset control never mounted, which is why it was unreachable.
- The date axis labels every tick as `Sep 12` instead of a bare day number with a month name on the first week. A year-long window with bare day numbers said nothing about where in the year you were.

## What did not change

- The Overview preview keeps its own 60-candle window (`PREVIEW_VISIBLE_CANDLES` in `LiquidationAnalysisSection.tsx`). The card never pans or zooms, so its window is everything it will ever draw; a year of candles in the card's width renders as sub-pixel hairlines. Only the full page rests at 365.
- Wheel zoom on the full page is intended: core-ui's `zoom` switch enables wheel, the +/- buttons and Reset together, and the reset control is the point of this change. Scrolling the page with the cursor over the chart zooms it first; a wheel-less zoom option in core-ui is a follow-up.
The Overview preview stays static. Its wheel listener would trap page scroll, and a preview that neither pans nor zooms has nothing to reset. Explore still opens the full page.

## Testing

- Liquidations page test asserts zoom is on and the reset control is reachable.
- Chart-data tests updated for the new axis label format.
- Replay fixture topped up so the recorded candle series covers the larger fetch.

Timeline at rest: month-day axis labels and the Reset view control

Timeline after zooming in

Timeline after Reset view

## Not in this PR

- Width-aware tick thinning in core-ui's `Timeline`: it draws seven labels at any width, and `Sep 12` labels can touch on a narrow card. A core-ui follow-up.
- A `zoom` option without wheel capture in core-ui's `Timeline`, so a page can offer the buttons and Reset without trapping page scroll.


jeremy-babylonlabs:
Local check on devnet with the god-mode simulated cascade. The devnet indexer holds 28 daily candles, so the chart shows every candle it has. The 365-candle window is covered by the tests.

Screenshot 1, Liquidation Analysis page: the zoom controls and the reset control render on the chart.
Liquidation Analysis chart with reset control

Screenshot 2, zoomed in three steps.
Zoomed in 3x

Screenshot 3, after clicking reset: the view is back to the full window.
After reset view

services/vault/src/components/pages/Liquidations/index.tsx
```diff
@@ -345,7 +345,7 @@ export default function Liquidations() {
             priceAxis={chart.timelinePriceAxis}
             safeZone={chart.safeZone}
             visibleCandles={TIMELINE_VISIBLE_CANDLES}
-            interactions={{ crosshair: true, pan: true }}
+            interactions={{ crosshair: true, pan: true, zoom: true }}
```
gbarkhatov: With `zoom: true`, core-ui's `Timeline` attaches a non-passive wheel listener and calls `preventDefault` whenever it can still zoom that way (`Timeline.tsx:205-218`). From the 365 rest window, scrolling the page with the cursor over the chart zooms out until 730 before the page moves, and scrolling up zooms in down to 10 candles; one trackpad flick does most of that.

That's the same scroll trap the PR body gives as the reason to keep the Overview preview static. If it's intended on the full page, worth a line in the body. If not, core-ui has one `zoom` switch for wheel, +/− and Reset, so a follow-up option to keep the buttons without wheel zoom would be the fix.

services/vault/src/components/pages/Liquidations/__tests__/Liquidations.test.tsx
```diff
@@ -53,6 +53,7 @@ const useBtcPriceCandlesMock = vi.fn();
 
 vi.mock("@/applications/aave/hooks/useBtcPriceCandles", () => ({
   useBtcPriceCandles: () => useBtcPriceCandlesMock(),
+  TIMELINE_VISIBLE_CANDLES: 365,
```
gbarkhatov, line 71: Nit: this mock (and the one in `LiquidationAnalysisSection.test.tsx:24`) hard-codes `TIMELINE_VISIBLE_CANDLES: 365` and renders 30 candles, so setting the real constant in `useBtcPriceCandles.ts` back to 60 would still pass every unit test. The "365-candle window is covered by the tests" comment is really covered by the zoom/reset test and the date-format tests. Optional: render >365 candles and assert exactly 365 candles render, or just note it's pinned by the replay fixture.

gbarkhatov (approved):
Looks good. Nothing blocking, two small points inline.

**Checked and clean**
- Reset view restores the 365-candle window, not all 730: `resetView` clears the override and `defaultWindow` falls back to `visibleCandles` (core-ui `Timeline.tsx:153,194-197`).
- The page and the Overview preview read the same module constant through the same hook, so they share one cached fetch.
- 730 is derived from the named constant; the two old private copies and `MONTH_LABEL_MAX_DAY` are gone with no references left. No new user-facing strings ("Reset view" is core-ui's label), no testid or motion changes.
- Replay fixture: the new `limit: 730` exchange matches the script's `CANDLE_LIMIT` and the replay lookup key.

**Considered and dropped**
- "The old `limit: 180` line in `recorded-run.jsonl` is dead": it isn't. `recording.ts:320-324` keeps it on purpose because the visual-regression job runs the merge-base app, which still sends `limit: 180`.

jonybur (dismissed):
Looks good to me.

gbarkhatov (dismissed):

services/vault/e2e/fixtures/replay/recorded-run.jsonl
```diff
@@ -41,3 +41,4 @@
...
...026\",\"high\":\"8041455618445\",\"low\":\"7877308000000\",\"close\":\"8041455618445\",\"decimals\":8},{\"bucketStart\":\"1787702400\",\"open\":\"7839310000000\",\"high\":\"7909539206986\",\"low\":\"4585539150262\",\"close\":\"7909022000000\",\"decimals\":8},{\"bucketStart\":\"1787616000\",\"open\":\"7895704424844\",\"high\":\"8116230484034\",\"low\":\"7835118000000\",\"close\":\"7847510000000\",\"decimals\":8},{\"bucketStart\":\"1787529600\",\"open\":\"7755801285648\",\"high\":\"7944420000000\",\"low\":\"7686278394156\",\"close\":\"7873841683364\",\"decimals\":8},{\"bucketStart\":\"1787443200\",\"open\":\"7724604070297\",\"high\":\"7787158117742\",\"low\":\"7597362075453\",\"close\":\"7756093036507\",\"decimals\":8},{\"bucketStart\":\"1787356800\",\"open\":\"7781108800000\",\"high\":\"7870182500000\",\"low\":\"7687362700000\",\"close\":\"7700564192470\",\"decimals\":8},{\"bucketStart\":\"1787270400\",\"open\":\"7355972213388\",\"high\":\"7850491865415\",\"low\":\"7355972213388\",\"close\":\"7850491865415\",\"decimals\":8},{\"bucketStart\":\"1787184000\",\"open\":\"7245643077671\",\"high\":\"7313845184618\",\"low\":\"7245643077671\",\"close\":\"7313845184618\",\"decimals\":8}]}}}"}
```
gbarkhatov, line 44: we need to think whether we need to exclude the recorded run from the repo itself

jonybur (commented):
Notes:

- `services/vault/src/components/simple/LiquidationAnalysisSection.tsx:144` - **Collapsing the two constants also widens the Overview preview's window from 60 to 365 candles, which the static card cannot zoom out of.**
  
  The preview used to carry its own `TIMELINE_VISIBLE_CANDLES = 60` with the rationale that it never pans, so Explore opens the full page; this PR deletes it and reuses the page's constant, which is now 365. The two copies were not really redundant: the page's number is the resting window of a pannable, zoomable chart, while the preview's is everything the card will ever show. In core-ui's `layoutCandles`, `slot = regionWidth / candles.length` and `bodyWidth = slot * 0.6` with no minimum width (only the body height is clamped to 1px), so 365 candles in the Overview card — and worse on mobile — render as sub-pixel hairlines with no interaction to recover from. Nothing in CI catches it: the replay fixture holds 22 candles and devnet 28, so `defaultWindow` clamps to the series length either way and the visual-regression run only flagged the liquidations-chart screens. Keep a separate `PREVIEW_VISIBLE_CANDLES = 60` for the card, or say in the description that the preview window tripled on purpose and show it at a realistic candle count.

gbarkhatov (dismissed):

jrwbabylonlab (dismissed):

gbarkhatov (approved):

gbarkhatov:
@jeremy-babylonlabs checks failing

gbarkhatov (approved):

jonybur (commented):
Notes:

- `services/vault/src/components/simple/LiquidationAnalysisSection.tsx:111` - **The Timeline price-axis top is still reduced over the whole fetched series, now 730 candles, while only 60 (preview) or 365 (page) are drawn, so the visible candles get squashed toward the floor.**
  
  Both Timeline call sites derive the price-axis ceiling by reducing `candle.high` over the whole fetched series, which this PR grows from 180 to 730 candles. `Timeline` takes `priceAxis` as a fixed prop and reads `priceMax` from `priceAxis[0]`, so the vertical domain never re-fits to the visible window, on pan or on zoom. The Overview preview is the sharp case: it draws only the last 60 candles but scales them against the highest high of the last two years, and it has no pan or zoom to recover, so a peak from 18 months ago pushes the visible candles and the safe zone into the bottom of the plot with dead space above it. The full page has the same mismatch at 365 of 730 in `services/vault/src/components/pages/Liquidations/index.tsx:163`, and zooming in there does not restore the vertical scale either. Nothing in CI catches this, since the replay fixture holds 22 candles and devnet 28, so the window clamps to the series length. Reduce `topPrice` over the trailing `visibleCandles` slice at each site rather than over the full series.

Issue #2522, opened 2026-09-16
chore(vault): Ledger in-flow UX polish (counters, cancel, wait copy, lock banner)


jrwbabylonlab:
## Summary

Smaller in-flow polish items found in the same audit. Independent of each other; none blocks the others.

## Items

- [ ] **Signing counter is suppressed for single-vault deposits.** `DepositProgressView/steps.ts:21-30` renders `(x of n)` only when `peginTotal > 1`, even though the Ledger provider emits per-ceremony ticks that are subscribed (`useDepositFlow.ts:607-609`) and discarded. Single-vault is the common case.
- [ ] **No prose about the long waits.** During the ~10 min Bitcoin confirmation and ~15–20 min vault provider setup, only the footer button relabelling to "Close & continue later" hints that leaving is safe. Nothing says so, and nothing says whether the Ledger can be unplugged or must be reconnected and the vault app reopened on return.
- [ ] **The "Bitcoin wallet locked" banner is Ledger-blind and misworded.** `DepositProgressView.tsx:277` reads `useBTCWallet().locked`, set only by the `getAccounts` poll, which `LedgerVaultProvider` does not implement — so it can never fire for Ledger. Its copy also says "extension", which is wrong for a hardware device. Fix the copy, or hide the banner for Ledger.

Issue #2521, opened 2026-09-16
fix(vault): make a Ledger session lost to an app switch recoverable mid-deposit


jrwbabylonlab:
## Summary

The most likely real-world device failure for a one-Ledger depositor — the DMK session dying because the user switched the device to the Ethereum app — is not in the device error taxonomy, is not resumable, and offers no way back. The user gets "Something went wrong during your deposit" with a Close button.

## Cause

Switching the device app re-enumerates USB. DMK's WebHID reconnect window is 6 seconds (`RECONNECT_DEVICE_TIMEOUT` in `@ledgerhq/device-transport-kit-web-hid`); a real Ethereum signature takes longer, so the session is removed.

The next BTC call then:

1. `holdsApprovedDepositTerms` is a pure host-side mirror read (`ledger-vault/provider.ts:543-559`), so it may return a stale `true` and the SDK skips the re-approval.
2. `signPsbt` → `gateSignContext` probes liveness, finds the session dead, tears down, and throws `WALLET_NOT_CONNECTED` (`provider.ts:928-935`).
3. A transport-level death instead maps to `CONNECTION_FAILED` (`provider.ts:1370-1379`).

Neither code is one of the three typed device codes, and neither is in `RESUMABLE_AFTER_REGISTRATION` (`services/vault/src/utils/errors/depositErrors.ts:124-129`), so both land in the generic fallback (`depositErrors.ts:486-491`, `copy.ts:897-898`).

## Two further problems on the same path

**No reconnect affordance in the deposit modal.** Reopening the device needs `navigator.hid.requestDevice()`, which requires a user gesture. The only reconnect control in the app is on the pre-deposit form (`SimpleDeposit.tsx:317-339`). This was already identified as missing in `todo/ledger/2110-2111-investigation-2026-08-23/2110-surface.md` §6(c) and never built.

**Returning to the tab can drop the wallet entirely.** `checkBTCConnection` (`BTCWalletProvider.tsx:332`) skips its silent pre-check because `LedgerVaultProvider` has no `getAccounts`, so it falls through to `connectWallet()` at `:382`. `doConnect` is idempotent *while the session lives* (`provider.ts:301-305`) — but after an app switch the session is dead, so it proceeds to `requestDevice()` outside a gesture, rejects, and the catch at `:404-408` calls `disconnect()`. The guard covers every case except this one. Related: #1836.

## Scope

[...]

Issue #2520, opened 2026-09-16
feat(vault): tell Ledger users when to switch device apps during a deposit


jrwbabylonlab:
## Summary

A depositor using one Ledger for both chains must physically switch the device between the Babylon Vault app and the Ethereum app three times during a single deposit. Ethereum wallets (MetaMask, OKX) tell them to switch. Our Bitcoin side says nothing, at any point.

## Evidence

An exhaustive grep across `services/vault`, `babylon-wallet-connector`, `babylon-core-ui` and `babylon-ledger-vault-signer` finds exactly two production strings naming the vault app, and both are reactive error text fired after the device has already refused an APDU:

- `services/vault/src/copy.ts:80-81` — "A different app is open on your signing device…"
- `packages/babylon-ledger-vault-signer/src/rawApdu.ts:78` — the `0x6E00` status-word message

Nothing proactive exists.

## The three moments that need a hint

| Step | Device app needed | Today |
|---|---|---|
| 4 — ETH registration (`submitPeginRequestBatch`) | switch **to** Ethereum | silent |
| 5 — Pre-PegIn sign and broadcast | switch **back** to Babylon Vault | silent, and costs two extra approvals |
| 13 → 14 — retrieve secret, then activate | vault app, then Ethereum, back to back | silent |

Step 5 is the sharp one. The app switch wipes the intent (Ledger HLD v22: "any disconnection (app switch, unplug, power-off) wipes them"), so `ensurePrePeginTermsApproval` re-runs the full derive plus approve ceremony — two extra physical approvals — while the stepper sits frozen on "Sign and broadcast BTC Pre-Pegin transaction" with a spinner and no step change.

A third forced ceremony hides under step 9's label "Authenticate session with vault provider" (`payoutSigning.ts:119-121`), which reads like a network call.

## The predicate already exists and is unused in the flow

`isLedgerVaultConnector` (`services/vault/src/context/wallet/VaultWalletConnectionProvider.tsx:50-54`) has exactly two consumers today: a fee tooltip and the reclaim row. No deposit step, modal or progress component reads it.

## Scope

- [ ] Copy before step 4: you will need to open the Ethereum app on your device.
- [ ] Copy before step 5: switch back to the Babylon Vault app, and expect to approve again.
- [ ] Copy between steps 13 and 14 for the activation pair.
[...]

Issue #2519, opened 2026-09-16
feat(wallet): pre-connect device instructions for the Ledger vault wallet


jrwbabylonlab:
## Summary

There is no guidance anywhere telling a Ledger user what to do before they click connect: plug in, unlock, and open the Babylon Vault app for this network. The wallet row shows only `label: "Hardware wallet"`.

## Precedent in our own repo

Keystone already does this. `packages/babylon-wallet-connector/src/core/wallets/btc/keystone/provider.ts:76-86` passes a five-line instruction panel, including a minimum firmware version, shown before the user does anything. Its comment states the intent:

> we cannot detect or gate it programmatically the way software wallets gate their injected version — it is surfaced as a hint in the connect dialog instead

Ledger is in a strictly worse position: it *can* detect the state programmatically and still shows nothing.

## The seam already exists and Ledger doesn't use it

`ProgressReporter` shipped in #1792. UniSat uses it (`unisat/provider.ts:114`) to render a title and description on the connect loader. `LedgerVaultProvider.connectWallet` takes no `onProgress` argument at all (`ledger-vault/provider.ts:286`), so it shows the generic "Connecting Ledger Vault" from `WalletConnector.ts:43`.

This overlaps #1793 ("Extend connect-progress messages to OKX, OneKey, Ledger, Keystone") — that issue is the natural home for the plumbing; this one is the Ledger copy and content.

## Scope

- [ ] Accept `onProgress` in `LedgerVaultProvider.connectWallet` and report a title plus description.
- [ ] Pre-connect instruction content, modelled on Keystone's panel and MetaMask's `ledgerConnectionInstruction*` strings: plug in, unlock with PIN, open the Babylon Vault app for this network.
- [ ] Include the Ledger Live discovery instruction. The app is published at catalog provider 4, so stock Ledger Live does not show it; testers must set Settings → Experimental features → My Ledger provider = 4. This currently lives only in #2110's closing note, and without it a user cannot even install the app.
- [ ] Note for Firefox/Safari users: the Ledger row is hidden entirely by the WebHID gate (`VaultWalletConnectionProvider.tsx:66-79`) with no explanation. Decide whether to show a disabled row with a "Chrome required" reason instead.

## Reference

MetaMask's equivalent, for tone:

```
"Prior to clicking confirm:"
"Be sure your Ledger is plugged in and to select the Ethereum app."
```

PR #2533, opened 2026-09-17
fix(wallet): show Ledger connect failures instead of bouncing to chai…

jrwbabylonlab:
fix(wallet): show Ledger connect failures instead of bouncing to chain selection

## What

The connect dialog rendered copy for one error code the Ledger provider never emitted, so a device sitting on the dashboard looked identical to nothing happening. `doConnect` now refuses a wrong or outdated app using the preflight it already ran, and the dialog renders the device codes it already had titles for.

- **Connect gate.** `LedgerVaultProvider.doConnect` reads the `GET_APP_AND_VERSION` preflight that `connectDmkSession` already runs. The app name must be the network's build ("Babylon Vault" on mainnet, "Babylon Vault Testnet" on test networks) and the version at least 0.10.1. A refused session is disconnected before the error is thrown and never installed. A version that is not plain `M.N.P` fails closed without echoing the device's string. A failed preflight is let through, so the first vault APDU reports its own typed error.
- **Re-gate on retry.** A session installed after a failed preflight was never gated, and every later connect reused it. A retried connect now reads the app on that session while no ceremony is in flight, refuses and tears down a wrong or outdated app, and otherwise rebuilds the APDU senders with the app identity. The signer package gains `refreshSessionApp` for this.
- **Dialog.** `useWalletConnectors` keeps the dialog open for `DEVICE_WRONG_APP` and `DEVICE_LOCKED` as well as `INCOMPATIBLE_WALLET_VERSION`. Every other code still returns to chain selection.
- **Status words.** 0x6D00, the base app's answer for a known class with an unknown instruction, maps to `DEVICE_WRONG_APP` alongside 0x6E00.

Tests: the provider pins the dashboard, both wrong-network apps, the version floor, a non-canonical version, a failed preflight, and refuse-on-retry and keep-on-retry; the dialog pins all three codes, the fall-through, and a code named after an inherited object property.

## Why

https://github.com/babylonlabs-io/babylon-toolkit/issues/2518: every connect-time failure of the Ledger Vault wallet was silent. The dialog's in-dialog copy existed only for `INCOMPATIBLE_WALLET_VERSION`, which the Ledger provider never emits, so the most common case, a device plugged in but on the dashboard, bounced the user back to chain selection with nothing shown.

## Approaches

- Gate on the preflight the session already runs: chosen. No extra device round trip on the normal path, and the refusal happens before any vault APDU.
- Map every connect error into the dialog: rejected. Most codes have no actionable copy, and a transport failure is better served by chain selection.
- Map 0x6D00 differently when the connect-time app is already the vault app: rejected. A gated app can only answer 0x6D00 if the host added an instruction without raising the version floor.

## Not in this PR

- Pre-connect device instructions (https://github.com/babylonlabs-io/babylon-toolkit/issues/2519), app-switch guidance and mid-deposit dead-session recovery (https://github.com/babylonlabs-io/babylon-toolkit/issues/2520, https://github.com/babylonlabs-io/babylon-toolkit/issues/2521).
- Route typed wallet errors from the visibility check to the UI instead of a silent disconnect, alongside #2521. (pre-review #6)
- Fold into #2522's Ledger polish items. (pre-review #7)

## Pre-review

<details>
<summary>Pre-review: 21 findings · 13 fixed · 0 moot · 2 follow-up · 6 declined · 0 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | The re-gate's idle-only guard does not exclude an in-flight device ceremony | merge-blocker | fixed |
| 2 | New signer export refreshSessionApp has no signer-level test and its doc overstates | normal | fixed |
| 3 | 0x6D00's status-word row is keyed by a literal and no test pins its message or app hint | normal | fixed |
| 4 | The stock-Bitcoin-app claim in the 0x6D00 comment is uncited | normal | fixed |
| 5 | New connect-dialog titles use sentence case in a file that uses Title Case | normal | fixed |
| 6 | A tab return turns a re-gate refusal into a silent disconnect | normal | follow-up: Route typed wallet errors from the visibility check to the UI instead of a silent disconnect, alongside #2521. |
| 7 | Connect-dialog copy shows raw device diagnostics and titles a stale app 'Update Ledger Vault' | normal | follow-up: Fold into #2522's Ledger polish items. |
| 8 | 0x6D00 maps to 'wrong app' for an old vault app on an ungated session | normal | declined: Re-litigates the scoped 0x6D00 decision; the re-gate reads the app on the next connect. |
| 9 | A disconnect racing the liveness probe passes undefined into the re-gate, which dereferences it | normal | fixed |
| 10 | The ceremony-lock docs now misdescribe the lock: the re-gate takes it, so the busy throw is not only a caller bug | normal | fixed |
| 11 | Three doc sites still scope the app hint to 0x6E00 after the branch widened it to 0x6D00 | normal | fixed |
| 12 | The 0x6D00 citation names the repo's former name while the package records the current one twice | normal | fixed |
| 13 | A re-read that learns nothing still swaps the session handle and rebuilds both senders | normal | fixed |
| 14 | Dialog titles use Title Case while the vault app's copy for the same conditions uses sentence case | normal | declined: each surface follows its own convention; finding 5 asked for Title Case here and finding 7 owns dialog copy |
| 15 | The in-flight test does not assert the re-gate resumes after the ceremony releases | normal | declined: release is an unconditional token-guarded finally and a neighbouring test pins gate-once on a free session |
| 16 | The signer re-read failure test covers the throw path, not the error-result path | normal | declined: both exits return the same value, so the second path adds no observable behaviour |
| 17 | A disconnect during the liveness probe still gets one app read on the torn-down session before the gate checks the token | normal | fixed |
| 18 | The lock comment records an unverified guess about DMK waiting for an unlock, and runs eight lines | normal | fixed |
| 19 | A ceremony started while the re-gate holds the lock gets the caller-bug busy error; proposal to await instead of throw | normal | declined: re-litigates the fail-fast decision recorded under finding 10; a retry design is out of scope |
| 20 | The gate's `deviceState.phase !== "idle"` branch is untested | normal | fixed |
| 21 | The name-present / version-absent let-through in refuseUnexpectedApp is untested | normal | declined: the signer returns both fields or neither, so the branch is unreachable through the real seam |

</details>


packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -338,8 +342,61 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+  private refuseUnexpectedApp(session: DmkSessionHandle): WalletError | undefined {
+    const expected = APP_NAME_BY_NETWORK[this.network];
+    if (session.appName !== undefined && session.appName !== expected) {
+      return new WalletError({
+        code: ERROR_CODES.DEVICE_WRONG_APP,
+        message: `Open the ${expected} app on your Ledger and try again.`,
+        wallet: WALLET_PROVIDER_NAME,
+      });
+    }
+    if (session.appVersion === undefined) return undefined;
```
gbarkhatov, line 419: When the preflight fails (`readAppAndVersion` returns `{}`), this lets the session through and `doConnect` installs it. If the next step of `selectedWallet.connect` then fails, `WalletConnector.connect` only emits `error` and never disconnects the provider, so the session stays installed. A locked device is the likely case, and it's exactly what the new dialog is for.

On retry, `doConnect` returns early at line 319 because `probeSessionAlive` → `isSessionAlive` only asks DMK whether the session exists. The name and version are never checked for that session, so a user who unlocks and opens an outdated or wrong app goes through ungated.

Not a regression, since main had no gate. Two possible fixes: re-run the preflight when the installed session has no `appName`, or don't install a session whose preflight failed. I haven't confirmed that a locked device fails GET_APP_AND_VERSION; that's Ledger OS behavior, not in the app repo.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -1348,7 +1405,12 @@ function toSignerWalletError(error: unknown): WalletError | undefined {
       { cause: error },
     );
   }
-  if (isLedgerDeviceError(error) && error.statusWord === SW_CLA_NOT_SUPPORTED) {
+  // Both mean the running app is not the vault app: an unknown class, or a known
+  // class (the shared Bitcoin base) without the vault instructions.
+  if (
+    isLedgerDeviceError(error) &&
+    (error.statusWord === SW_CLA_NOT_SUPPORTED || error.statusWord === SW_INS_NOT_SUPPORTED)
```
gbarkhatov, line 1453: Minor: this mapping is shared by the signing and intent paths, not just connect. The vault app itself answers 0x6D00 for any instruction it doesn't implement (`APP_SPECIFICATION.md:411`), so a host newer than the firmware would show "Wrong app on device" when the right action is updating the app. The message passed through here is also the raw `(ins 0x.. p1 0x.., sw 0x6d00)` text, and unlike the 0x6E00 path it doesn't name the app to open.

If you want to tighten it: when the connect-time `appName` is already the expected vault name, map 0x6D00 to `UNKNOWN_ERROR` (or an update prompt) rather than `DEVICE_WRONG_APP`.

packages/babylon-ledger-vault-signer/src/rawApdu.ts
```diff
@@ -53,6 +53,13 @@ const SW_DEVICE_LOCKED = new Set([0x5515, 0x6982, 0x5303]);
 
 /** CLA not supported — what the dashboard or a wrong app returns. */
 export const SW_CLA_NOT_SUPPORTED = 0x6e00;
+/**
+ * INS not supported. The vault app is built on `bitcoin_app_base`, so the stock
+ * Bitcoin app shares CLA 0xE1 and answers this class; its dispatcher returns
+ * this for a vault instruction it does not implement (`base:dispatcher.c:170`
```
gbarkhatov: §7 citation: `b0c0ac4d` is the app-repo commit, but `dispatcher.c` lives in the `bitcoin_app_base` submodule, pinned at `e400d8d8` (`git ls-tree origin/develop`), under `src/boilerplate/`. The claim itself is right: line 170 is `} else if (!ins_found) {` and line 171 sends `SW_INS_NOT_SUPPORTED`. To match the `SW_BAD_STATE` citation just below: `base:src/boilerplate/dispatcher.c:170-171` @ e400d8d8.

gbarkhatov (dismissed):
Nothing blocking. The gate does what the body says, and the status-word mapping matches the firmware. Three inline points, one of them a §7 citation fix.

**Checked against the firmware** (`app-babylon-vault@b0c0ac4`, `bitcoin_app_base@e400d8d8`)
- App names: "Babylon Vault" for mainnet, "Babylon Vault Testnet" for testnet and signet (`Makefile:44,53`). Compared with strict equality, so the other network's app and the dashboard ("BOLOS") are both refused.
- `MIN_APP_VERSION = "0.10.1"` matches `Makefile:23-26`. A version below it and an unparseable version both fail closed, and an unparseable string isn't echoed back to the user.
- 0x6E00 for an unknown CLA and 0x6D00 for a known CLA with an unknown INS (`src/boilerplate/dispatcher.c:167-171`).

**Checked and clean**
- A refused session is disconnected before the error is thrown, and `this.session` is never set on refusal. The disconnect-race token check still runs first.
- Only `INCOMPATIBLE_WALLET_VERSION`, `DEVICE_WRONG_APP` and `DEVICE_LOCKED` keep the dialog open. Everything else still returns to chain selection, and `WalletConnector` emits `error` only from `connect`/`disconnect`, so a mid-session signing error can't open this dialog.
- The `SW_INS_NOT_SUPPORTED` re-export in `index.ts` is used, and no payload bytes are logged.

**Tests, optional**
- Only the mainnet-app-on-signet mismatch is tested. The reverse, a testnet app on a mainnet provider, isn't.
- The 0x6D00 app-name hint in `rawApdu.ts` has no test (`rawApdu.test.ts` isn't in the diff).
- The `toString` test can't fail on either `Set.has` or `Map.get`, so it doesn't pin anything.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: not planning — MIN_APP_VERSION bumps cover new instructions

packages/babylon-ledger-vault-signer/src/rawApdu.ts
jrwbabylonlab: fixed

gbarkhatov (approved):

gbarkhatov (approved):
@jrwbabylonlab needs rebasing

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -338,8 +346,84 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+  /**
+   * A session installed after a failed preflight was never gated, and a
+   * retry (unlock, open an app, connect again) reuses it while it lives. Read
+   * the app now and gate it. The phase alone is not "no ceremony in flight"
+   * (derive, approve, PoP and refund all send at idle), so the read skips a
+   * running ceremony and holds the ceremony lock itself.
+   */
+  private async gateUngatedSession(session: DmkSessionHandle, token: number): Promise<void> {
+    if (session.appName !== undefined || this.activeOperation || this.deviceState.phase !== "idle") return;
+    const refreshed = await this.withDeviceOperation("connectWallet", () => refreshSessionApp(session));
```
jonybur, line 389: **The retry re-gate holds the ceremony lock during its device read, so a user ceremony started meanwhile fails with the 'caller bug' INVALID_PARAMS error.**

`gateUngatedSession` sends the `GET_APP_AND_VERSION` read inside `withDeviceOperation("connectWallet", ...)`, which takes the same lock that signing, approval and derive ceremonies use. `withDeviceOperation` throws `INVALID_PARAMS` ("already running a device ceremony; wait for it to finish before calling ...") whenever the lock is held, and its comment says that only happens on a caller bug. Now a tab-return visibility check, which calls `connectWallet()` without any user gesture, can start this read on an ungated session. If the user then clicks sign, approve or sign-message, the action is rejected with a confusing error. That is most likely on a locked device, where the read may be slow. The new guard covers a ceremony that is already running, but not one that starts during the read. Either have ceremonies wait for a pending gate read, or map this collision to a retryable error and add a test for it.

packages/babylon-ledger-vault-signer/src/__tests__/dmkSession.test.ts
```diff
@@ -189,6 +190,27 @@ describe("connectDmkSession", () => {
...
+  it("fills the app name and version from a successful re-read on the same session", async () => {
+    const handle = { dmk: dmkStub as never, sessionId: "session-1" as never };
+
+    const refreshed = await refreshSessionApp(handle);
+
+    expect(refreshed).toMatchObject({ sessionId: "session-1", appName: "Babylon Vault Testnet", appVersion: "0.9.4" });
+    expect(dmkStub.sendCommand).toHaveBeenCalledWith(expect.objectContaining({ sessionId: "session-1" }));
+  });
+
+  it("keeps the handle's fields when the re-read fails", async () => {
```
jonybur, line 203: **The 'keeps the handle's fields when the re-read fails' test can't fail if refreshSessionApp drops or clears those fields.**

The handle passed to `refreshSessionApp` has no `appName` or `appVersion`, so on a failed read there is nothing to keep. `toEqual` also ignores properties that are `undefined`. An implementation that returns `{ ...handle, appName: undefined }` or drops the fields would still pass, so the test doesn't pin the "same fields when it fails" promise in `dmkSession.ts`. Seed the handle with `appName`/`appVersion` values and assert with `toStrictEqual`. The command-level error branch (`status: "ERROR"`) of `refreshSessionApp` could get the same test.

jonybur (commented):
Comments inline.

gbarkhatov (approved):

gbarkhatov (dismissed):

gbarkhatov (approved):

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
```diff
@@ -339,8 +356,88 @@ export class LedgerVaultProvider implements IBTCProvider {
...
+   * A session installed after a failed preflight was never gated, and a
+   * retry (unlock, open an app, connect again) reuses it while it lives. Read
+   * the app now and gate it. The phase alone is not "no ceremony in flight"
+   * (derive, approve, PoP and refund all send at idle), so the read skips a
+   * running ceremony and holds the ceremony lock itself.
+   */
+  private async gateUngatedSession(session: DmkSessionHandle, token: number): Promise<void> {
+    // A disconnect that landed during the probe already owns the session.
+    if (token !== this.disconnectToken || this.session !== session) return;
+    if (session.appName !== undefined || this.activeOperation || this.deviceState.phase !== "idle") return;
```
jonybur, line 388: **The re-gate's in-flight guard misses the pubkey and policy reads, which send APDUs without taking the ceremony lock.**

gateUngatedSession decides it is safe to put GET_APP_AND_VERSION on the wire when the ceremony lock is free and deviceState.phase is idle, then takes the lock itself. But getDevicePubkeyHex (485-496) and getPolicyContext (503-547) deliberately send APDUs without withDeviceOperation and never move the phase, and both are reachable from the public getAddress and getPublicKeyHex. A second connectWallet() that lands while one of those reads is outstanding passes this guard and issues a second command on the same session, which is exactly the concurrent APDU the lock doc at 223-231 says gets eaten with 0x6A80 and desyncs the interrupt loop. Every caller inside this repo awaits the two sequentially, so I could not trigger it here; a host that calls getAddress() while a visibility check fires connectWallet() would. The cheap fix is a pending flag set alongside pubkeyHexPromise/policyContextPromise that this guard also checks, rather than wrapping those getters in withDeviceOperation, which would make the Promise.all in Wallet.connect collide with itself.

jonybur (commented):
Comments inline, plus:

- `packages/babylon-ledger-vault-signer/src/dmkApduSender.ts:96` - **createDmkApduSender's doc still scopes the connect-time app hint to 0x6E00 after this change widened it to 0x6D00.**
  
  The doc on createDmkApduSender says it carries the connect-time app identity "so 0x6E00 names the app that answered", but classifyStatusWord now emits that hint for SW_INS_NOT_SUPPORTED too (rawApdu.ts:161-164). The sibling docs at rawApdu.ts:128 and :177 were updated to say "0x6E00/0x6D00"; this fourth site was missed, so the only doc a reader of the sender sees understates when the hint appears. It is a one-line fix: name both status words here as well.

packages/babylon-ledger-vault-signer/src/dmkSession.ts
```diff
@@ -200,7 +201,7 @@ export async function connectDmkSession(): Promise<DmkSessionHandle> {
  * `GET_APP_AND_VERSION` preflight, run only at connect — the most useful fact
```
jonybur, line 201: **readAppAndVersion's docblock still says the preflight is run only at connect, which refreshSessionApp in the same file makes false.**

This docblock opens with "`GET_APP_AND_VERSION` preflight, run only at connect". The PR rewrote its last sentence but left that clause, and refreshSessionApp (line 247), added by this same change, calls readAppAndVersion again on a retry, so the statement no longer describes the code. In this package the comments are the spec a reader works from, and the repo asks that comments referencing old behaviour be updated after a change. Reword it to say the preflight runs at connect and again when the host re-gates a session whose first read failed.

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: not planning — one instant exchange, even when locked

packages/babylon-ledger-vault-signer/src/__tests__/dmkSession.test.ts
jrwbabylonlab: fixed

packages/babylon-wallet-connector/src/core/wallets/btc/ledger-vault/provider.ts
jrwbabylonlab: not planning — DMK serialises sends; single-shot reads keep no state

packages/babylon-ledger-vault-signer/src/dmkSession.ts
jrwbabylonlab: fixed

PR #2551, opened 2026-09-17
feat(vault): respect hub borrow limits, hub state and hub errors

### LINKED ISSUE #2437: Multi-hub: the same token from two Hubs must be two markets in the UI

gbarkhatov:
Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2437

## What

One Aave v4 Core Spoke lists reserves from two Hubs, and each Hub keeps its own limits and state for our spoke. This change makes the app respect them, closing the last part of https://github.com/babylonlabs-io/babylon-toolkit/issues/2437.

- **Borrow limit per hub.** Each Hub caps what our spoke may draw on a reserve (its draw cap), separately from the Hub's liquidity, which every spoke shares. A live Hub read (`getSpokeConfig`, `getSpokeTotalOwed`, `getSpokeDeficitRay`) gives what is left under that cap, computed as `Hub._validateDraw` counts it. The borrow form's Available and Max are now the smallest of collateral, liquidity and that borrow limit, with its own "Amount exceeds borrow limit" and "borrow limit reached" messages, and Select hub shows the smaller of liquidity and the borrow limit. A read that fails leaves the borrow uncapped, as the liquidity read already does.
- **Hub active and halted.** Our spoke's config on each reserve's Hub is read at config load, and read again live by the borrow, repay and withdraw forms. A reserve on a Hub where our spoke is inactive or halted leaves the borrowable list, so the pickers and Borrow more stop offering it. The borrow form is blocked with a reason naming the hub when its own Hub is inactive or halted, or when a Hub where the user has debt is inactive, since every borrow refreshes the user's risk premium on each of those Hubs. The repay form is blocked only by its own Hub, since a repay touches no other; the debt stays listed. The withdraw review is blocked when the BTCVault collateral's Hub is inactive or halted, or a Hub where the user has debt is inactive. A config that could not be read never blocks, and every failed read is logged.
- **Activity.** A liquidation card lists a "Debt repaid" row for every reserve its transaction repaid, in log order; when one transaction liquidates several BTCVaults, the repays appear on its first card only.
- **Readable Hub errors.** The Hub's errors (`DrawCapExceeded`, `AddCapExceeded`, `InsufficientLiquidity`, `SpokeNotActive`, `SpokeHalted`, `InvalidPremiumChange`) are decoded on borrow, repay, withdraw and vault activation, and a decoded error keeps its arguments. Borrow and repay show a draw cap in whole tokens and liquidity scaled by the asset's decimals, name the hub where the revert pins it, and fall back to fixed text; the common Spoke reserve errors get fixed text too.
- **Critical path.** `services/vault/src/services/vault/vaultActivationService.ts` (CLAUDE.md §5) changes only the error ABIs passed to the activation write, adding the Hub errors from the import-free `clients/hubErrors.ts`; the secret and hashlock path is untouched.
- **Devnet E2E run.** `multi-hub --borrow-token=USDC --borrow-usd=1` on localhost borrowed 1 USDC from Babylon Hub and 1 USDC from Core Hub, confirmed each debt landed on its own reserve and each Loans row names its hub, then repaid both with Max, ending with no debt on either reserve. Devnet has no halted or inactive Hub and no reachable borrow limit, so the blocked and capped paths are covered by unit tests only.

## Why

The multi-hub flow (https://github.com/babylonlabs-io/babylon-toolkit/pull/2515) let users pick a hub but ignored what the Hub enforces. A Core Hub credit line can hold far less than the Hub's liquidity, a halted or inactive Hub rejects a borrow, repay or withdraw, and those reverts reached the user as camel-case error names. Devnet has no halted Hub and no reachable cap, so none of this is visible there, but mainnet Hubs can change their config for our spoke.

## Approaches

- **Hub state filtered at config load, checked live in the forms.** The borrowable list is filtered in one place, so every surface that offers a reserve follows it. The config is not refreshed while the app stays open, so the forms that block an action read the Hub state again every minute: a cached "halted" must not keep a user from repaying after the Hub lifts it.
- **The borrow limit read live, not at config load.** What our spoke owes moves with every borrow on that Hub, so a cached figure would drift quickly; a borrow or repay also refreshes it, and the Hub's liquidity, straight away.
- **A conservative coupled block.** The Spoke skips the risk-premium refresh when the user's premium is zero before and after, and then an inactive debt Hub would not reject a borrow or withdraw. Predicting that needs the premium the Spoke would compute, so any inactive debt Hub blocks both.
- **Attribution by checking before signing.** `SpokeNotActive`, `SpokeHalted` and `InvalidPremiumChange` carry no arguments and the adapter passes them through unchanged, so a revert cannot say which Hub raised it. The form names the Hub from the state it read instead. `SpokeNotActive` on a borrow may come from another Hub, so it keeps the fixed text.
- **A failed read never blocks.** The transaction simulation and the decoded revert remain the backstop, and a failed display read must not lock a user out of repaying.
- **No SDK change.** The Hub reads and error fragments live app-side, next to the existing Hub liquidity and rate reads.

## Not in this PR

- The one-borrowable-asset-per-position limit (`MAX_USER_RESERVES_LIMIT`), tracked in https://github.com/babylonlabs-io/babylon-toolkit/issues/2513.
- Indexing the debt a liquidation repays, per reserve: today's contracts and indexer record none, so the liquidation card change is not reachable on chain until https://github.com/babylonlabs-io/babylon-vault-indexer/issues/172.
- The markets table, market page and Loans rows keep showing the Hub's shared liquidity.
- A cap re-read in `validateBorrowPreSign`.
- Predicting `InvalidPremiumChange`, which depends on every user of our spoke.

## Pre-review

<details>
<summary>Pre-review: 25 findings · 19 fixed · 0 moot · 0 follow-up · 6 declined · 0 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | A cached hub halted/inactive flag keeps repay, borrow and withdraw blocked after the hub recovers | normal | fixed |
| 2 | A transaction that liquidates several vaults repeats every repay on every liquidation card | normal | fixed |
| 3 | The hub-config read swallows its failures, and the draw-usage read returns an error field nothing reads | normal | fixed |
| 4 | Borrow form messages conflict when a hub is blocked, and an exhausted borrow limit says nothing until an amount is typed | normal | fixed |
| 5 | Select hub shows an empty value instead of the borrow limit when liquidity failed to load | normal | fixed |
| 6 | Borrow form comments and the margin constant still describe liquidity only, and the one-reserve array is rebuilt every render | normal | fixed |
| 7 | The real-wallet E2E borrow action does not fail fast on 'Amount exceeds borrow limit' | normal | fixed |
| 8 | The inactive-debt-hub block is stricter than the contract when the user's risk premium is 0 | normal | fixed |
| 9 | The borrow-limit tie-break and the draw-headroom hook have no tests | normal | fixed |
| 10 | Markets on a halted or inactive hub disappear from the markets table, landing card and pickers | normal | declined: same behavior paused and frozen reserves already have |
| 11 | aaveHub.ts grows past about 500 lines | normal | declined: no repo line-limit rule; the hub reads sit with the other hub reads |
| 12 | A liquidation card's summary names only the first repay | normal | declined: no production code renders a liquidation group's summary |
| 13 | ONE_MINUTE_MS is declared again in the new headroom hook | normal | declined: matches the five sibling hooks; consolidating them is unrelated cleanup |
| 14 | The live hub-state hook falls back to the config-load copy whenever its reserve set changes | normal | fixed |
| 15 | The repay form's callout ranks amount and transaction errors above a protocol or hub block | normal | fixed |
| 16 | The multi-vault liquidation test does not pin which card gets the repays | normal | fixed |
| 17 | One getBorrowLimit test asserts four behaviors | normal | fixed |
| 18 | Nothing tests the borrow form's callout order or its borrow-limit-reached warning | normal | declined: the Borrow form has no component test harness; the order is plain branching over tested inputs |
| 19 | The E2E waitForBorrowCta comment still lists the old instant-fail labels | normal | fixed |
| 20 | The borrow margin's comment claims it absorbs interest against the borrow limit, which it cannot | normal | fixed |
| 21 | The borrow-limit headroom is not refreshed after a borrow or repay | normal | fixed |
| 22 | While a new reserve set loads, a reserve the config-load copy blocks reads as usable | normal | fixed |
| 23 | A hub-read test resets its rejecting mock only after its assertions | normal | fixed |
| 24 | invalidateHubQueries' comments name two reads but it also refreshes the hub-config query | normal | fixed |
| 25 | Nothing tests the repay form's callout order | normal | declined: the loan forms have no component test harness; same as finding 18 |

</details>


services/vault/src/applications/aave/hooks/useAaveReserveDrawHeadroom.ts
```diff
@@ -0,0 +1,76 @@
+/**
+ * What our Core Spoke may still draw on each reserve's hub before it hits the
+ * hub's draw cap, in whole tokens. A hub's liquidity is shared by every spoke on
+ * it, but our spoke may only draw up to its own cap, so the two limits differ.
+ *
+ * `null` means no cap, or that the read is loading or failed: callers leave the
+ * borrow uncapped, the same fallback as a missing liquidity read.
```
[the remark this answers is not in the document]
gbarkhatov: The fallback is deliberate. These reads decide what the form offers; they don't feed the transaction. Borrow, repay and withdraw all simulate with `eth_call` on the same RPC client before the wallet is opened (`executeTx` in `clients/transaction.ts`). A borrow above our spoke's draw cap, or against an inactive or halted Hub, fails that simulation, nothing is signed, and this PR's Hub error ABIs turn the revert into a readable message (the cap in whole tokens, or the hub state). So a failed read doesn't lead to a rejected on-chain transaction. The one leftover cost is a repay that may send its ERC-20 approve first; that moves no funds.

Blocking on an unknown state would be worse. If the RPC is down, the simulation fails anyway. If a read keeps failing, for example after a Hub upgrade changes `getSpokeConfig`, every borrow, repay and withdraw would be blocked until a frontend release, even though the contract would accept them. Blocking repay and withdraw is the dangerous part, because a user near liquidation must be able to repay. The CLAUDE.md fallback rule covers values used in amounts, calldata or protocol parameters. The amount here is user input checked by the contract, and the Hub state never reaches the calldata. The existing Hub liquidity read already falls back the same way. Every failed read is logged, and each distinct failure sends one Sentry event, so an outage shows up in Sentry.

services/vault/e2e/fixtures/replay/recorded-run.jsonl
```diff
@@ -40,3 +40,4 @@
...
...00000000000000000000000000000000000ffffff000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000ffffffffff00000000000000000000000000000000000000000000000000000000017d78400000000000000000000000000000000000000000000000000000000000ffffff000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000ffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007d000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000\"}\n"}
```
jrwbabylonlab, line 43: oh god. this file is huge

jrwbabylonlab (approved):

services/vault/e2e/fixtures/replay/recorded-run.jsonl
gbarkhatov: It's the recorded devnet run that the browser tests replay instead of calling a live backend: the visual check, the Ethereum-only access spec, the high-severity-errors spec, one network-failures test and the deposit-progress layout spec. It is 70 KB over 43 lines, but each line holds one full request and response with the multicall hex inline, which is why one line looks enormous in the diff. This PR adds one line: an archive read of `getSpokeConfig` at the recording's block. The config load now makes that read, and the replay fails on any read the recording doesn't hold. Without the fixture, those tests would go back to screenshotting the error boundary, which is why #2265 added it. The real issue is that the recording predates the multi-hub deployment, so each new read gets patched in by hand. Re-recording is tracked in https://github.com/babylonlabs-io/babylon-toolkit/issues/2552.

PR #2554, opened 2026-09-17
fix(vault): wait for the wallet to count each transaction before the next one

### LINKED ISSUE #2514: fix(vault): back-to-back wallet transactions fail with "nonce too low" because the wallet's node lags the receipt

gbarkhatov:
image

## What

Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2514.

- **Wait for the wallet after each mined transaction.** Both send helpers, `executeWrite` (`clients/eth-contract/transactionFactory.ts`) and the Aave `executeTx` (`applications/aave/clients/transaction.ts`), now wait after the receipt until the wallet's own provider counts the transaction, reading `eth_getTransactionCount` at `latest` through the wallet client, before returning. Before each read the app asks the wallet for the receipt, which makes MetaMask poll its latest block at once instead of on its 20 s timer. The wait is skipped when the connected account did not send the transaction (a Safe's transaction is sent by its executor), and it never fails a transaction that already mined: every read is bounded by one 45 s deadline, and when the wallet does not catch up in time or cannot answer, it logs a warning and returns.
- **Retry a stale nonce once.** If the node rejects a send with "nonce too low", the helper sends once more only when the rejected transaction cannot have mined: the account's count on the app's RPC is the same before the send and after the rejection, and the wallet has caught up to exactly that count. It then runs the pre-flight simulation again and sends; in every other case, and on a second rejection, the error is shown to the user. "Already known", "already imported" and "replacement transaction underpriced" are never retried.
- **Readable error.** A stale nonce is its own error kind, split from "already submitted", with copy in `copy.ts` instead of viem's "Nonce provided for the transaction is lower than the current nonce of the account". It covers borrow, repay, withdraw, reorder, ERC-20 approve and vault activation.
- **E2E harness.** The temporary stale-nonce resubmit in `e2e/real` (`staleNonceRetry.ts`, marked for removal with #2514) is deleted, so a stale nonce now fails the run, and a failed borrow or repay logs the text of the form's "Transaction failed" callout.
- **Vault activation.** `vaultActivationService.ts` (CLAUDE.md §5) is not changed, but it sends through `executeWrite`, so it gets the wait and the one retry. The retry re-sends the same calldata after the SDK's hashlock check, and only when the account's count shows the rejected send did not mine.
- **Devnet E2E run.** `multi-hub --all-reserves --borrow-usd=1` (borrow $1 on each of the 6 reserves, then repay each with Max) on localhost with MetaMask. Before the wait and retry, run 2 failed on a repay sent 4 s after its approve, and the form showed the readable stale-nonce message. With them, 5 runs in a row plus a `repay-all` passed: 99 wallet transactions, no stale nonce, and 18 wallet confirms per run, so the retry never fired. One more run on the final code, after the pre-review hardening (read deadlines, the retry's count checks), also passed: 18 wallet transactions, 18 confirms, no stale nonce, and no debt left.

## Why

A repay is an approve followed straight away by the repay, and a user often starts the next action seconds after the last one. The app decides a transaction is mined from its own RPC, but MetaMask picks the next nonce from its own node, reading the count at the latest block it polled about every 20 s, so it can hand out the nonce the previous transaction just used. The node rejects the send before broadcasting it, and the user saw raw node text mid-flow, labelled as "already submitted".

## Approaches

- **Wait for the wallet's count at `latest`.** That is the value MetaMask's nonce tracker reads; `pending` skips MetaMask's block pin and reports caught up too early. The receipt request only speeds the wait up; the count decides.
- **The app does not set the nonce.** The wallet also counts other dApps' pending transactions, a user can edit the nonce, and some wallets ignore a dApp-supplied one.
- **Not batched with `wallet_sendCalls` (EIP-5792), not `permit`, not a fixed delay.** Batching needs an account upgrade prompt on MetaMask and does not cover consecutive actions; the adapter has no permit path; a fixed delay is either unreliable or slows every transaction.
- **No SDK change.** The Safe skip compares the mined transaction's sender with the connected account instead of exporting the SDK's smart-account detection.

## Not in this PR

- The SDK's peg-in registration send (`sendAndWait`), which sends one transaction and is followed by activation much later.

## Pre-review

<details>
<summary>Pre-review: 12 findings · 7 fixed · 0 moot · 0 follow-up · 5 declined · 0 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | The stale-nonce retry can send a second copy when the rejected transaction itself was already mined | normal | fixed |
| 2 | The post-mine wallet wait can hang forever because wallet reads have no timeout | normal | fixed |
| 3 | The stale-nonce retry still prompts again when the wallet never caught up | normal | fixed |
| 4 | A single not-found read of the just-mined transaction skips the wallet wait | normal | fixed |
| 5 | walletNonce.ts declares its own sleep instead of the existing abortableSleep | normal | fixed |
| 6 | Several walletNonce branches have no behavioural test | normal | fixed |
| 7 | The new waits ignore cancellation | normal | declined: no caller passes a signal to executeWrite/executeTx, and the existing repay retry is not cancellable either |
| 8 | The E2E repay timeout does not cover two worst-case 45 s wallet waits | normal | declined: only non-MetaMask wallets could reach it; MetaMask leg times did not change across 5 devnet runs |
| 9 | On WalletConnect wallets the count read may come from a relay RPC, not the wallet's node | normal | declined: unverified and harmless: the wait returns early and the one retry still backs it |
| 10 | Every send now reads the app RPC's transaction count before the wallet prompt | normal | declined: one read bounded by the HTTP transport timeout; a failure only disables the retry |
| 11 | The wallet-equals-app check refuses a legitimate retry when another transaction from the account mines during the wait | normal | declined: fails closed: no duplicate, and the readable error lets the user resend |
| 12 | The sendWithStaleNonceRetry doc comment leaves "Already known" alone on a line after the rewrap | normal | fixed |

</details>


jonybur (approved):
Looks good to me.

jrwbabylonlab (approved):

Issue #2067, opened 2026-07-16
[v3] Withdraw & refund flows restyle


jeremy-babylonlabs:
**Area:** `services/vault` (`WithdrawFlow/`, `WithdrawVaults/`, `ExpiredDepositSection`, `components/deposit/RefundModal/`)

**Design:** [Review Withdraw](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10088-38704) · [Review Refund](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=6288-85814)

## Context

The existing withdraw review already presents amount, fee rate, commission, Health Factor delta, ETA, and payout address. Refund supports review, success, and not-broadcast states. V3 restyles these surfaces without changing their transaction flows.

## Scope

- [ ] Restyle Review Withdraw and its fee/address rows
- [ ] Restyle Refund review, success, and not-broadcast states
- [ ] Align refund and withdrawal status treatment with the v3 row system
- [ ] Preserve signing, submission, polling, completion, and retry behavior
- [ ] Keep the implementation behind `ENABLE_V3_UI`

## Open question

Confirm whether withdraw progress remains the existing single terminal screen or becomes a multi-step presentation. Do not add a second transaction state machine if the difference is visual only.

## Acceptance criteria

- [ ] Withdraw and refund review states match Figma in light and dark
- [ ] Displayed amounts, fees, health-factor change, ETA, and payout address remain unchanged
- [ ] Existing withdraw and refund transactions complete through the same code paths

## Depends on

- #2041

PR #2560, opened 2026-09-18
fix(vault): show the withdraw health-factor block warning as a banner

### LINKED ISSUE #2472: Withdraw review: show the health-factor warnings as banners, not bare text
**Design:** [Withdraw banner, health factor below 1.0](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=13586-75758)

## Context

`services/vault/src/components/simple/WithdrawFlow/WithdrawReviewContent.tsx:178-198` renders the two health-factor warnings as bare `<Text>` lines — no container, no title, no icon.

The design shows a full `Callout`: bordered card, red warning glyph, title **"Withdraw unavailable"**, and a body identical to today's `hfBlockWarning` string (`services/vault/src/copy.ts:1423-1424`). Confirm is already disabled on `wouldBreachHF` (`WithdrawReviewContent.tsx:206`), matching the design.

This state was not covered by #2067, which restyled Review Withdraw at node `10088-38704`.

## Load-bearing test hooks

`data-testid="withdraw-hf-block-warning"` and `data-testid="withdraw-hf-at-risk-warning"` are both consumed by E2E:

- `services/vault/e2e/pages/WithdrawModal.ts:21,25` — both
- `services/vault/e2e/real/actions/withdraw.ts:72` — the block one

Carry both onto the new markup. Dropping one breaks the E2E run with no compile error.

## Acceptance criteria

- [ ] The block warning renders as `Callout variant="error"` with title "Withdraw unavailable" and a body unchanged from `REVIEW_COPY.hfBlockWarning`.
- [ ] The new title string is added to `copy.ts` under the withdraw-review section, not inlined in the component.
- [ ] Both testids are queryable after the change, asserted in a test.
- [ ] Confirm stays disabled whenever `isProcessing || wouldBreachHF` (`WithdrawReviewContent.tsx:206`) — unchanged.
- [ ] Thresholds still come from `getWithdrawHfWarningState`; no literal `1.0` in the component.

## Out of scope

The at-risk banner (`isAtRisk`, `WithdrawReviewContent.tsx:190-198`) has no design and no agreed title. Leave it as-is and file separately once design covers it.

## Key files

- `services/vault/src/components/simple/WithdrawFlow/WithdrawReviewContent.tsx`
- `services/vault/e2e/pages/WithdrawModal.ts`

Verified against `8f62bcfa`.

jeremy-babylonlabs:
Closes #2472

## What

On the Review Withdraw screen, the warning that a withdrawal would drop the health factor below the on-chain floor now renders as the shared `Callout` error banner instead of a bare red text line: a bordered card with a warning triangle, the title **Withdraw unavailable**, and the same body copy as before.

The diff is three files:

- `services/vault/src/components/simple/WithdrawFlow/WithdrawReviewContent.tsx` — the `wouldBreachHF` branch swaps `<Text variant="body2" className="text-error-main">` for `<Callout variant="error" title={REVIEW_COPY.hfBlockTitle} icon={<WarningIcon size={14} color="text-accent-contrast" />}>`. `WarningIcon` joins the existing `@babylonlabs-io/core-ui` import. An inline comment above the banner flags its `data-testid` as a real-wallet E2E hook and names the consuming action file, matching the comment the confirm button below already carries.
- `services/vault/src/copy.ts` — one key, `hfBlockTitle: "Withdraw unavailable"`, inside `COPY.withdraw.review` next to the existing `hfBlockWarning`.
- the component's test file — two tests.

Behaviour deliberately unchanged: the body string, the threshold source (`getWithdrawHfWarningState` and `WITHDRAW_HF_BLOCK_THRESHOLD`, so no numeric literal enters the component), and the confirm button staying disabled on `isProcessing || wouldBreachHF || hubBlockMessage !== null`. Both `data-testid` hooks (`withdraw-hf-block-warning`, `withdraw-hf-at-risk-warning`) are preserved because the real-wallet E2E CLI and the Playwright page object drive the flow through them; `e2e/` is untouched.

`role="alert"` is not set at the call site. `Callout` derives it from `variant === "error"`, so the banner is announced by assistive technology without the component needing to say so.

## Why

[Issue #2472](https://github.com/babylonlabs-io/babylon-toolkit/issues/2472): the block warning was a bare `<Text>` line with no container, title or icon, while the design ([Figma node 13586-75758](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=13586-75758)) shows a full banner. The state was missed by #2067, which restyled the rest of Review Withdraw at a different node. Confirm was already disabled on this condition, which matches the design.

## Approaches

**Chosen: replace only the `wouldBreachHF` line with a `Callout`.** The component already imported `Callout` and used it for the transaction-failed message below, so this reuses a shared primitive with no new styling and no new component.

**Rejected: convert all three review warnings to Callouts.** The at-risk warning has no design and no agreed title, and the hub-block message is a different surface. Converting them would invent design.

**Rejected: a wrapper component for "review warnings".** One call site and no second concrete case, so an abstraction would be speculative.

**Rejected: `variant="warning"` instead of `variant="error"` plus an icon override.** `Callout`'s `warning` variant ships this exact `WarningIcon` by default, which makes it look like the tidier choice, but it paints the tile `bg-warning-main` #EF6C00 (amber). The design's tile is red, so `variant="error"` with the icon passed explicitly is the faithful pairing.

## Not in this PR

- The at-risk warning (`isAtRisk`) keeps its bare-text treatment: no design, no agreed title. Worth its own issue once design covers it.
- The hub-block message (`hubBlockMessage`) keeps its bare-text treatment; a different surface from this issue.
- The banner's tile colour. The design specifies `error/dark` **#C62828**; `Callout`'s error variant paints `bg-error-main` **#D32F2F**. Both are existing core-ui tokens, so the design's shade is not a new value, but retoning `VARIANT_BG.error` would change every error `Callout` in the app. That is a shared-primitive decision rather than part of this fix, and it is the only known remaining deviation from the design node.

## Testing

`pnpm nx affected -t lint` over the changed files passes: 0 errors, 134 pre-existing warnings, none of them in these files.

`pnpm nx affected -t test` reported 3 failed files / 336 passed / 1 skipped. Every failure is `Test timed out in 5000ms` with no assertion failure, in `src/dev/__tests__/demoArtifactDownload.test.ts`, `src/utils/__tests__/lazyWithRetry.test.ts` and `src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts` — three files this change does not touch. nx flagged the task flaky. Re-running exactly those three files on an idle machine gives 29 passed / 29. The vault suite passed in full (4205 passed / 2 skipped) on an earlier run of the same tree. `WithdrawReviewContent.test.tsx` passed 6/6 in every run.

Two tests were added, each proved to fail when the line it guards is broken:

- *announces the block warning as a titled alert banner* — asserts `role="alert"`, the title text, and the warning glyph. Breaking `variant="error"` to `"warning"` turns the role into `status`; deleting the `title` prop drops the title; deleting the `icon` prop falls back to the default close glyph.
- *warns without blocking when the projection is above the floor but at risk* — covers the `withdraw-hf-at-risk-warning` hook, which no unit test previously touched, and pins that the at-risk state does not disable Confirm.

The glyph assertion is `banner.querySelector('svg[viewBox="0 0 22 20"]')` rather than the plainer `querySelector("svg")` used elsewhere in the repo. The icons carry no testid, role or distinguishing class and the tile is `aria-hidden`, so a bare `svg` query would pass with the default close glyph still in place and could never fail. `viewBox` is the only attribute separating `WarningIcon` (`0 0 22 20`) from `CloseIcon` (`0 0 14 14`).

Confirm staying disabled is pinned by the pre-existing test at `WithdrawReviewContent.test.tsx:85`.

Not verified in a browser: no dev server was run, and the banner has not been visually compared against the design node.

## Pre-review

<details>
<summary>Pre-review: 1 finding · 1 fixed · 0 moot · 0 follow-up · 0 declined · 0 open (0 merge-blockers) · 0 undecided · lint passed, test failed on unrelated timeouts</summary>

|---|---|---|---|
| 1 | The withdraw-hf-block-warning testid carries no inline E2E-hook comment, unlike its sibling testid in the same file | normal | fixed |

One reviewer claim was refuted: that `variant="warning"` should replace `variant="error"` plus the icon override. `VARIANT_BG.warning` is amber #EF6C00 while the design's tile is red, so the substitution would move the banner further from the design, not closer.

</details>


jonybur (approved):
Looks good to me.

gbarkhatov (approved):

gbarkhatov (approved):

PR #2561, opened 2026-09-18
fix(vault): seed the fee-rate selector from the fastest tier first

### LINKED ISSUE #2479: Fee-rate selector highlights Slow when every tier returns the same rate
## Context

On the Sign and broadcast Pre-Pegin step, the fee-rate selector highlights **Slow** even though the app chose **Fast**, whenever the three mempool tiers return the same number. Every tier reads `1 sat/vB` and the pill under `Slow` is the one outlined, captioned `~1 hour`.

## Cause

The app's default really is Fast:

- `services/vault/src/hooks/useNetworkFees.ts:40` maps `defaultFeeRate` to mempool's `fastestFee`.
- `services/vault/src/hooks/deposit/useEstimatedBtcFee.ts:51-56` uses `defaultFeeRate` when no override is supplied.
- `services/vault/src/components/simple/SimpleDeposit.tsx:425` commits that value as the deposit's fee rate.
- `services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx:79` seeds the selector's own state as `useState<TierKey>("fast")`.

The selector then reconciles that committed rate back to a tier, and the reconciliation is order-dependent (`FeeRateSelector.tsx:99-107`):

```ts
if (feeRate === slowFee) {
  key = "slow";
} else if (feeRate === avgFee) {
  key = "avg";
} else if (feeRate === fastestFee) {
  key = "fast";
}
```

When `hourFee === halfHourFee === fastestFee` — a quiet mempool, testnet, or regtest — all three comparisons are true and the first branch wins. The selection flips to `slow`.

## Impact

The committed fee rate is correct and unchanged; only the reported tier is wrong. That still matters:

- The outlined pill claims the user picked the cheapest, slowest option when the app picked the fastest.
- The caption is actively misleading. `slowHint` is `"~1 hour"` (`services/vault/src/copy.ts:529`) while the identical rate is also sold as `~10 min` under Fast. Nothing about the transaction differs.
- It makes the deposit look mis-configured during a step where the user is about to sign.

## Expected

When several tiers share a rate, the selector reports the tier the app actually selected — Fast for the seeded default — rather than the first tier that happens to compare equal.

## Acceptance criteria

- [ ] With `hourFee === halfHourFee === fastestFee`, the selector highlights **Fast** on first render, not Slow, and shows the Fast hint.
- [ ] With distinct tier values, seeding is unchanged: a committed rate equal to a tier selects that tier.
- [ ] A committed rate matching no tier still selects **Custom** and populates the custom field.
- [ ] The live-follow behavior after seeding is unchanged — the selected tier keeps tracking its own mempool value, and a custom rate stays frozen until the user edits it.
- [ ] The committed fee rate is byte-identical before and after this change for every case above; this is presentation only.
- [ ] A test covers the all-tiers-equal case and fails if the branch order is restored.

## Key files

- `services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx`
- `services/vault/src/components/simple/DepositProgressView/__tests__/` (selector tests)

## Not in scope

Whether the default tier should be Fast or Avg at all. That changes the fee actually paid and touches critical path 2 (fee calculation consistency), and the activation window is measured in Ethereum blocks from peg-in registration (`block.number > createdAt + pegInActivationTimeout`, see `services/vault/src/utils/errors/contract.ts:381`) — the Bitcoin confirmation wait sits inside that window, so a slower rate eats the margin left for the remaining steps. That needs its own issue with real mainnet numbers.

Verified on `main` at `068a1dd8`.

jeremy-babylonlabs:
Closes #2479

## What

The Pre-PegIn fee-rate selector seeds its own tier selection once, from the committed `feeRate` prop, using a chain of equality comparisons. This reorders that chain from `slowFee → avgFee → fastestFee → custom` to `fastestFee → avgFee → slowFee → custom`.

The diff is four lines of production code in `services/vault/src/components/simple/DepositProgressView/FeeRateSelector.tsx` plus three tests. The seed stays a single pass inside the existing `seededRef` guard: no new state, refs, helpers or props, and nothing after the seed block is touched.

The three tests pin one seed leg each. The first seeds with every tier at `1 sat/vB` and asserts Fast is pressed — it fails under the old slow-first order. The second seeds with `feeRate={2}` against distinct tiers and asserts Slow is pressed, the third with `feeRate={5}` and asserts Avg is pressed — each fails if its own branch is dropped from the chain, since the committed rate then falls through to Custom. All three were confirmed by mutation, and each fails alone.

The seed rule the code now expresses is a fixed priority — first tier whose current value equals the committed rate, in the order fast, avg, slow, else custom. It does not read the current selection. It coincides with "keep what is already selected" only because `selectedKey` is initialised to `"fast"` and the seed runs before any click can land.

## Why

On the "Sign and broadcast Pre-Pegin" step the selector highlighted **Slow** while the app had committed **Fast**, whenever the three mempool tiers returned the same number — a quiet mempool, testnet, or regtest. Every tier read `1 sat/vB` and the outlined pill was Slow, captioned `~1 hour`.

The app's default is Fast: `useNetworkFees.ts` maps `defaultFeeRate` to mempool's `fastestFee`, `useEstimatedBtcFee.ts` uses it when no override is supplied, `SimpleDeposit.tsx` commits it as the deposit's fee rate, and the selector initialises its own state to `"fast"`. The seed then reconciled that committed rate back to a tier by testing `slowFee` first, so with all three tiers equal every comparison was true and the first branch won.

The committed fee rate was always correct — this is a presentation defect. It still mattered: the outlined pill claimed the depositor had picked the cheapest, slowest option, and the caption was actively misleading, selling an identical rate as `~1 hour` under Slow and `~10 min` under Fast, on the screen where the depositor is about to sign.

Fixes https://github.com/babylonlabs-io/babylon-toolkit/issues/2479

## Approaches

**Chosen: reorder the seed comparisons.** The selector already knows which tier the app picked — it is the initial state. Making the comparison order agree with it is the smallest change that expresses the rule, and it touches nothing outside the seed block.

**Rejected: track the tier key in the parent's committed deposit state.** Reconciling a tier from a rate is only ambiguous because the key is not persisted. Persisting it would remove the ambiguity outright, but it changes `SimpleDeposit` and the persisted deposit shape — a schema change carrying resume and migration cost — for a defect that never moves a satoshi.

**Rejected: deduplicate equal tiers, hiding the redundant pills.** This removes the ambiguity by removing the duplicates, but it changes the UI the design specifies, and it makes the control's shape depend on mempool conditions.

## Behaviour that changes, and what does not

The committed rate is byte-identical **at seed time** in every case. A branch only wins when `feeRate` already equals that tier's value, so the `derivedRate` computed below the seed equals `feeRate` whichever branch fires, and the `derivedRate !== feeRate` guard suppresses the callback exactly as before. The fix cannot call `onFeeRateChange` where the old code did not, nor with a different value, on first render.

**After** seeding, live-follow tracks the seeded tier by design. Where the old code seeded `slow` on a tie and then followed `slowFee` once the tiers diverged on a later refetch, it now follows `fastestFee`. That is the intended correction — the app committed `fastestFee`, so following the Fast tier is what the depositor asked for — and under mempool's `fastestFee >= halfHourFee >= hourFee` ordering the drift is toward overpaying, never toward underfunding a transaction.

One narrower consequence: when `avgFee` equals `slowFee` but `fastestFee` differs, a committed rate equal to both now seeds `avg` where it seeded `slow`. The number is unchanged, since the tied tiers hold the same value; only the highlighted pill differs, and it prefers the faster of the two.

## Not in this PR

- Whether the default tier should be Fast or Avg at all. That changes the fee actually paid and touches critical path 2 (fee calculation consistency); the activation window is measured in Ethereum blocks from peg-in registration, so a slower rate eats the margin left for the remaining steps. It needs its own issue with real mainnet numbers.

## Pre-review

<details>
<summary>Pre-review: 5 findings · 4 fixed · 0 moot · 0 follow-up · 1 declined · 0 open (0 merge-blockers) · 0 undecided · lint passed, test failed</summary>

|---|---|---|---|
| 1 | Two of the new test's five assertions cannot fail under the pre-fix branch order, and the test name claims a behaviour they do not pin | normal | fixed |
| 2 | Acceptance criterion 5's "byte-identical for every case" holds at seed time only; live-follow after seeding intentionally commits a different value than the old code would | normal | fixed |
| 3 | The seed rule as briefed ("a committed rate that matches the currently selected tier keeps that tier") does not describe the code, which never reads `selectedKey` | normal | fixed |
| 4 | No seed test pins the reordered `avg` or `slow` legs; deleting the `slowFee` branch outright would fail no test | normal | fixed |
| 5 | With `avgFee === slowFee` and a distinct `fastestFee`, a committed rate equal to both now seeds `avg` where it previously seeded `slow` | normal | declined: strictly less wrong than the bug being fixed, committed rate unaffected, and the only fix is the parent-state design this PR rejected |

Two runs. Run 1 was the full reviewer set over the whole change; run 2 re-checked the one file that moved since then, the test file, with a verdict and new-defect lane, and raised no new defect. Three claims were raised in run 1 and refuted against the code: that seeding can latch against the placeholder zero tiers when the network-fee query errors (the CTA is disabled at `estimatedFeeRate <= 0`, so the sign step is unreachable); that a remount after a rejected signature re-seeds over the depositor's earlier choice (`started` is never reset, and re-entry re-commits the rate); and that a click during `isLoading` is silently discarded by the seed (the shared fee query already holds data before the selector can mount).

Checks, run 2: `nx affected -t lint` passed with 0 errors and rewrote no file. `nx affected -t test` failed — 334 test files passed, 5 failed, 1 skipped, of 340. None of the five references the changed component, and the scoped `FeeRateSelector` suite passes 10/10. Three of them (`GodModePanel`, `VaultWalletConnectionProvider.disabledWallets`, `Liquidations`) pass in isolation, 78/78, so they are parallel-load timeouts; the other two (`demoArtifactDownload`, `scriptPubKeyHexToBtcAddress`) are the same pre-existing failures run 1 recorded, and reproduce in isolation on their own.

</details>


jonybur (approved):
Looks good to me.

gbarkhatov (approved):

gbarkhatov (approved):

PR #2562, opened 2026-09-18
fix(vault): use the designed padlock and title on the wallet-locked banner

### LINKED ISSUE #2471: Wallet locked banner: use the designed padlock icon and title
**Design:** [Wallet locked banner](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=13586-63856)

## Context

The banner ships already and its body copy matches the design. Two gaps remain: the title string and the icon.

- The design's title is **"Bitcoin wallet is locked"**. `services/vault/src/copy.ts:1287` reads `"Bitcoin wallet locked"`.
- The design's glyph is a padlock. `variant="error"` supplies `CloseIcon` (a ✕) — `packages/babylon-core-ui/src/components/Callout/Callout.tsx:44-50`. `Callout` already accepts an `icon` prop (`Callout.tsx:30`), so no new component is needed, but core-ui has no padlock: `packages/babylon-core-ui/src/components/Icons/common/` holds 26 icons and none is a lock.

`COPY.wallet.locked.title` is shared by three call sites, so the title edit changes all of them:

- `services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx:425`
- `services/vault/src/components/deposit/ReclaimModal/ReclaimReviewContent.tsx:248`
- `services/vault/src/components/deposit/RefundModal/RefundReviewContent.tsx:212`

## Open questions

- Does the padlock apply to all three lock banners, or only the deposit-progress one?
- The design still shows the footer caption "Do not spend the BTC used for this deposit until the transactions are confirmed." #2344 asked for that caption to be removed, and it is still in the tree at `copy.ts:483`. Confirm with design whether the caption stays or the mock predates #2344.

## Acceptance criteria

- [ ] `LockIcon` added under `packages/babylon-core-ui/src/components/Icons/common/`, exported from `Icons/index.ts`, and present in `Icons.stories.tsx`.
- [ ] `COPY.wallet.locked.title` is `"Bitcoin wallet is locked"`, asserted as a string literal in a test. The existing tests read the constant (`DepositProgressView.test.tsx:848,863,878,893`), so they cannot catch a wrong value.
- [ ] `RefundModal.test.tsx:148` asserts the old title as a literal — updated.
- [ ] The banner renders the padlock and its icon tile matches node `13586-63856`.
- [ ] The banner keeps `role="alert"` (`Callout.tsx:66`) — a locked wallet blocks signing and must stay announced.
- [ ] The render condition at `DepositProgressView.tsx:424` and the unlock CTA (computed at `:328`, rendered at `:484`) are unchanged.

## Key files

- `services/vault/src/components/simple/DepositProgressView/DepositProgressView.tsx`
- `services/vault/src/components/deposit/ReclaimModal/ReclaimReviewContent.tsx`
- `services/vault/src/components/deposit/RefundModal/RefundReviewContent.tsx`
- `packages/babylon-core-ui/src/components/Callout/Callout.tsx`
- `packages/babylon-core-ui/src/components/Icons/index.ts`
- New: `packages/babylon-core-ui/src/components/Icons/common/LockIcon.tsx`

Verified against `8f62bcfa`.

jeremy-babylonlabs:
Closes #2471

## What

Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2471.

The wallet-locked banner now matches its design. Three things changed.

`COPY.wallet.locked.title` is now `"Bitcoin wallet is locked"`.

core-ui gains a `LockIcon` (Material "lock" path, 24×24 viewBox, `currentColor`, same prop shape as its sibling icons), exported from `Icons/index.ts` and listed in the icon gallery story.

`Callout` gains a variant `accent`, whose icon tile is `bg-secondary-main` — the `#CE6533` the design calls `accent/primary`. Its default glyph is `InfoIcon`, the same as `info`: a generic accent variant should not default to a padlock. The three wallet-locked call sites pass the padlock explicitly through the existing `icon` prop, at the design's 24px inside the unchanged 40px tile. Every other variant keeps its 14px glyph and its background.

All three call sites also pass `role="alert"` explicitly. They used to inherit it from `variant="error"`; `accent` resolves to `status` on its own, and a locked wallet blocks signing, so the announcement has to stay. There is a test for that at each call site and one for the variant's role resolution in core-ui.

Three wallet-connector e2e fixture files quote the banner title verbatim in doc comments explaining why each wallet's auto-lock is disabled. Those quotes were updated to the new string. No fixture behaviour changed.

## Why

The banner already shipped and its body copy already matched the design. Two gaps remained: the title read `"Bitcoin wallet locked"` where the design reads `"Bitcoin wallet is locked"`, and `variant="error"` supplied a `CloseIcon` (a ✕) where the design shows a padlock on an orange tile. core-ui had 26 icons and no lock.

## Approaches

**A named `accent` variant (chosen).** The design's tile is `accent/primary` `#CE6533`. `secondary.main` is that exact hex in both the light and dark palettes, so the tile is correct in either theme without a per-theme rule.

**Reuse `variant="warning"` (rejected).** Its tile is `warning/main` `#EF6C00`, a different colour from the design's `accent/primary` `#CE6533`. Both hexes were read from the design file; matching the node is the rule.

**Pass a `className` to recolour the tile (rejected).** The tile has no className hook. Adding one to a shared component so a single call site can repaint it is worse than naming the variant the design system already describes.

**Make the padlock the `accent` default icon (rejected).** `accent` is a general-purpose tile colour, not a lock notice. The lock belongs at the call sites that mean a lock.

## Not in this PR

- The footer caption "Do not spend the BTC used for this deposit until the transactions are confirmed" is still in the design mock and still in the tree. Removing it is https://github.com/babylonlabs-io/babylon-toolkit/issues/2344.
- The `DepositProgressView` render condition and the unlock CTA are untouched.
- `Callout.stories.tsx` has no `accent` story; it lists stories by hand rather than iterating a variants array.
- `BtcActionGate` also renders `COPY.wallet.locked.title` and picks up the new string, but it is a full-page gate with its own unlock button, not a `Callout` banner. It gets no padlock, by design.
- A shared test hook for the core-ui icon set, so tests stop asserting glyph geometry (pre-review #2).

## Pre-review

<details>
<summary>Pre-review: 5 findings · 1 fixed · 0 moot · 1 follow-up · 3 declined · 0 open (0 merge-blockers) · 0 undecided · lint passed, test passed</summary>

|---|---|---|---|
| 1 | The comment above the wallet-lock hook still calls the banner an "error-style Callout" after the variant became `accent`. | normal | fixed |
| 2 | Three vault tests assert the padlock by the rendered SVG path's `d` prefix, coupling app tests to core-ui's glyph geometry. | normal | follow-up: add a shared test hook to the core-ui icon set, then re-point the three assertions at it |
| 3 | Three `wallet.locked.title` assertions in `DepositProgressView.test.tsx` read the COPY constant rather than the literal string. | normal | declined: they are negative assertions, where a literal would go vacuous on copy drift; the literal assertion the criteria ask for is at line 848 |
| 4 | The variant name `accent` points at the `accent-*` text-token family while its tile uses `bg-secondary-main`. | normal | declined: `accent` names the Figma token the design specifies; the tailwind alias is an implementation detail |
| 5 | The three banner tests each assert two behaviours under one name. | normal | declined: one mount, one banner behaviour, both halves independently mutation-falsifiable |

</details>


jonybur (approved):
Looks good to me.

gbarkhatov (approved):

gbarkhatov (approved):

PR #2564, opened 2026-09-18
fix(vault): reserve Very Safe for no borrow and give Liquidatable its own colour

### LINKED ISSUE #2410: Health factor: adopt the five-band status scale (Liquidatable / Risky / Moderate / Safe / Very Safe)
## Context

The health factor status label shown on the Risk section, the Overview position block and the Borrow/Repay cards. Design wants five bands; the app has four, and the boundaries differ.

The label is produced by two layers:

- `packages/babylon-ts-sdk/src/tbv/integrations/aave/utils/healthFactor.ts:25` — `getHealthFactorStatus()` returns `no_debt | danger | warning | safe`
- `services/vault/src/components/simple/RiskSection/railLayout.ts:14` — `getRiskDisplayState()` maps that to the display state `noPosition | verySafe | safe | moderate | liquidatable`, and `services/vault/src/copy.ts:1974-1980` holds the strings.

## Current vs expected bands

| HF | Today | Expected |
|---|---|---|
| < 1.0 | Liquidatable | Liquidatable |
| 1.0 – 1.1 | Moderate | **Risky** (new) |
| 1.1 – 1.5 | Moderate | Moderate |
| 1.5 – 2.0 | Safe | **Moderate** |
| 2.0 – 50 | Safe | **Safe** |
| > 50 | Very Safe | Safe |
| no borrow | **No Position** | **Very Safe** |

The reported case is HF **1.05** rendering as "Moderate". It falls in `warning` (`>= 1.0`, `< HEALTH_FACTOR_WARNING_THRESHOLD` = 1.5) → `moderate`. It is also exactly `MIN_HEALTH_FACTOR_FOR_BORROW` (`packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts:66`), which is why borrowing more is blocked — the label and the borrow gate disagree with each other today.

### 1. Add a `risky` band

**Now:** `RiskDisplayState` has no `risky` member and `COPY.risk.status` has no `risky` string.
**Expected:** new state for 1.0 – 1.1, with its own label and colour.
**Where:** `railLayout.ts:7-12` (type), `railLayout.ts:14-33` (`getRiskDisplayState`), `services/vault/src/copy.ts:1974`

### 2. Move the band boundaries

**Now:** the only numeric boundaries are 1.0 (`danger`), `HEALTH_FACTOR_WARNING_THRESHOLD` = 1.5 and `HEALTH_FACTOR_HEALTHY_THRESHOLD` = 50.
**Expected:** 1.0 / 1.1 / 2.0.
**Where:** `packages/babylon-ts-sdk/src/tbv/integrations/aave/constants.ts:60`, `services/vault/src/applications/aave/utils/healthFactorDisplay.ts:33`

`HEALTH_FACTOR_WARNING_THRESHOLD` is exported from the published SDK (`@babylonlabs-io/ts-sdk/tbv/integrations/aave`) and is also the green stop of the risk rail gradient (`railLayout.ts`, `GRADIENT_MIN_RAMP_PCT` block). Changing 1.5 → 2.0 moves that gradient too. Decide whether the four SDK statuses gain a fifth member or the new bands live entirely in the vault display layer — the second keeps the SDK contract stable and is the smaller change.

### 3. "No borrow" reads Very Safe, not No Position

**Now:** `no_debt` and "no position" both map to `noPosition` → "No Position"; `verySafe` is reserved for HF > 50.
**Expected:** a position with collateral and no borrow reads **Very Safe**. `HEALTH_FACTOR_HEALTHY_THRESHOLD` = 50 then has no consumer for this label.
**Where:** `railLayout.ts:19-27`

Pairs with the Overview ask in the same review — "Health factor → Very Safe and an infinity sign next to the heart when we have collateral".

### 4. Colour for the new band

**Now:** `getHealthFactorColor()` covers four statuses — green / amber / red / grey.
**Expected:** five display states need five colours. Confirm the `risky` colour with design.
**Where:** `services/vault/src/applications/aave/utils/healthFactorDisplay.ts:5-27`

## Open question

Does the borrow gate move with the labels? `MIN_HEALTH_FACTOR_FOR_BORROW` is 1.05, which now sits inside the new "Risky" band. Leaving it there means a user can be labelled "Risky" and still borrow down to 1.05.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788253486068059)

> let's have the following **[HF statuses](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=10162-15543)**:
> • **< 1.0 — Liquidatable**
> • **1.0–1.1 — Risky**
> • **1.1–2.0 — Moderate**
> • **> 2.0 — Safe**
> • **No borrow — Very Safe**
>
> because right now my HF is **1.05**, but it's shown as **Moderate**, while this is already quite risky I can't even borrow more.

Reported state — Overview, HF 1.05 labelled Moderate

</details>

jeremy-babylonlabs:
Closes #2410

## What

Completes the five-band health-factor scale in the vault display layer, after https://github.com/babylonlabs-io/babylon-toolkit/pull/2405 landed the SDK side.

- `getRiskDisplayState` now returns `safe` for every SDK `safe` status. `verySafe` is reached only through `no_debt`. `HEALTH_FACTOR_HEALTHY_THRESHOLD` (50) loses its last consumer and is deleted along with its re-export.
- The function no longer takes a `healthFactor` argument; its single caller passes the status and `hasPosition`.
- `HEALTH_FACTOR_COLORS` gains `DARK_RED: "#C62828"`, which `getHealthFactorColor` returns for `danger`. `risky` keeps `#FF1744`, so the two red bands are now distinguishable.
- `GREEN` and `AMBER` move to the design hexes `#15B768` and `#F7931A`, matching the `--risk-green` and `--risk-amber` values the risk rail already used in light mode.
- `globals.css` gains `--risk-red-dark`, `tailwind.config.ts` the matching `risk-red-dark` colour token, and both per-state colour maps in the Risk section point at them: `STATE_STYLE.liquidatable` for the heart, dot and label, and `RiskPriceRail`'s `currentRingClass.liquidatable` for the current-price marker ring.

## Why

A position at health factor 1.05 was labelled "Moderate" while sitting exactly on the borrow gate, which read as safer than it was. https://github.com/babylonlabs-io/babylon-toolkit/issues/2410 asks for five bands: Liquidatable below 1.0, Risky to 1.1, Moderate to 2.0, Safe above that, and Very Safe when there is no borrow at all.

Two display-layer gaps remained after the SDK change. Very Safe was still a numeric label for health factor above 50 rather than the "no borrow" state the design defines, so a borrower at 60 read as Very Safe. And `danger` and `risky` shared one red, so the band the SDK had just learned to distinguish still looked identical on screen.

The colour constants and the rail's CSS variables had also drifted apart: the heart icon painted `#00E676` and `#FFC400` while the rail painted `#15B768` and `#F7931A` from the same design tokens. Aligning the constants removes that split in light mode.

## Approaches

**Very Safe.** Keeping a numeric cutoff and raising it above the borrow range was considered and rejected: the design defines Very Safe as the absence of a borrow, not as a large number, and any cutoff would reintroduce the same "which number is very safe" question. Deleting the threshold rather than leaving it unused follows the zero-dead-code rule.

**The fifth colour.** Distinguishing the two red bands by weight or opacity instead of a second hex was rejected — the design gives Liquidatable its own token (`error/dark`), and deriving it in CSS would have put the value in a second place. The chosen approach keeps one source per colour: a hex constant for the heart icon, an RGB-channel custom property for the rail, both carrying the design value.

**Where the new bands live.** The issue offered a choice between extending the SDK's status union and keeping the new bands in the vault display layer. The SDK already gained the fifth status in #2405, so this change touches no SDK source and only maps what is already there.

## Not in this PR

- Moving `MIN_HEALTH_FACTOR_FOR_BORROW` (1.05). It now sits inside the Risky band, so a position can read "Risky" and still borrow. This is the issue's stated open question and needs a product decision, not a display change.
- The Overview "infinity sign next to the heart when we have collateral" ask from the same review.
- Dark-mode `--risk-amber` and `--risk-green`. No dark design values were read, so the existing dark values stand.
- Making `getHealthFactorColor` theme-aware so the heart icon and the rail share one source in both themes (pre-review #4).
- Removing the unused `HEALTH_FACTOR_COLORS.GRAY` (pre-review #5).

## Pre-review

<details>
<summary>Pre-review: 8 findings · 5 fixed · 0 moot · 2 follow-up · 1 declined · 0 open (0 merge-blockers) · 0 undecided · lint and test passed</summary>

|---|---|---|---|
| 1 | WithdrawReviewContent.test.tsx still asserts the old RED for a danger status, so the test suite is red. | merge-blocker | fixed |
| 2 | RiskPriceRail's per-state ring map still paints liquidatable bright red, so one panel shows two different reds for the same state. | normal | fixed |
| 3 | Re-declaring --risk-red-dark in the .dark block with the light value is a no-op that reads as a dark-mode override. | normal | fixed |
| 7 | The risk-var comment enumerates a Figma source for every var except the new --risk-red-dark. | normal | fixed |
| 8 | The comment added for --risk-red-dark spells it "colour" while the same comment block says "colors". | normal | fixed |
| 4 | Aligning HEALTH_FACTOR_COLORS.GREEN to the light hex regresses dark mode, where the heart icon and the rail used to agree. | normal | follow-up: make getHealthFactorColor return rgb(var(--risk-*)) so both themes share one source |
| 5 | HEALTH_FACTOR_COLORS.GRAY is never returned by getHealthFactorColor and has no consumer. | normal | follow-up: delete GRAY and narrow the exported HealthFactorColor union |
| 6 | The colour tests assert raw hex literals instead of importing HEALTH_FACTOR_COLORS. | normal | declined: the repo's test philosophy prefers explicit values, and asserting the constant the function returns is tautological |

</details>

Loans summary card, health factor 0.95 (Liquidatable)

Loans summary card, health factor 1.05 (Risky)

Loans summary card, health factor 1.25 (Moderate)

Loans summary card, health factor 2.4 (Safe)


jonybur (commented):
Notes:

- `services/vault/src/components/simple/RiskSection/RiskSection.tsx:112` - **A dust-debt position with HF above 1000 now renders the label "Safe" next to "-" instead of the infinity glyph, which reads as missing data.**
  
  With the `safe` to `verySafe` promotion removed from `getRiskDisplayState`, every SDK `safe` status now reaches `formatHealthFactor(healthFactor)` here. That function returns "-" for Infinity or anything above `HEALTH_FACTOR_DISPLAY_CAP` (1000). So a borrower with only dust debt, for example leftover interest after a repay, used to see the infinity glyph and now sees "Safe" next to a green "-". That is the same placeholder the no-position state uses, so it looks like missing data. Keeping the "Safe" label is the intended change. The problem is the value shown beside it. Consider rendering `COPY.risk.healthFactorInfinity` or a capped string such as ">1000" when the state is `safe` and the value is above the cap. Also add a render test for that case, since the tests that covered `safe` with null or Infinity were deleted.

gbarkhatov (approved):

gbarkhatov (approved):

PR #2566, opened 2026-09-18
fix(vault): match the Pre-Pegin broadcast modal to its design

### LINKED ISSUE #2473: Decide whether the Pre-Pegin broadcast modal stays, then match its design
**Design:** [Pre-Pegin broadcast modal](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=13603-77420)

## Open question — blocks this issue

Raised by design: *"do we need to show this pre-pegin broadcast modal, or could we skip it?"*

## The modal is not dead code

`services/vault/src/components/deposit/BroadcastSuccessModal/index.tsx` renders from `services/vault/src/components/simple/PendingDepositModals.tsx:119` on `broadcastModal.successOpen`. `PendingDepositModals` is mounted by `services/vault/src/components/Activity/ActivityListWithRefund.tsx:90` and `services/vault/src/components/vaults/VaultsLifecycleSections.tsx:756` (handler at `:621-624`, condition at `:677`). `services/vault/src/hooks/deposit/useBroadcastState.ts:8-9` documents it taking over after a successful broadcast.

Removing it is a product change, not a cleanup.

## Design deltas if it stays

Heading, body and the `Done` label already match the design (`services/vault/src/copy.ts:469-476`). Two visual gaps:

- The BTC mark is a contained circle in the design; the code renders `max-w-[160px] w-full` (`BroadcastSuccessModal/index.tsx:44-48`), roughly 1.7× that.
- The design's second paragraph is one sentence. `COPY.deposit.broadcastSuccess.footnote` (`copy.ts:473-474`) is about three times longer and adds the HTLC secret plus a "Check back here" instruction.

The mock spells "authorisations". `copy.ts` is American English by its own style rules, so ship "authorizations".

## Acceptance criteria

- [ ] The keep-or-remove decision is recorded on this issue by design plus one engineer before any code lands.
- [ ] If removed: delete `BroadcastSuccessModal/`, `COPY.deposit.broadcastSuccess`, the success state in `services/vault/src/hooks/deposit/useBroadcastModal.ts:27` and its test `useBroadcastModal.test.ts`, the `successOpen` / `successAmount` / `handleSuccessClose` props (`PendingDepositModals.tsx:27,119-123`), the `broadcastModal.successOpen` term at `VaultsLifecycleSections.tsx:677`, and the duplicated mocks in `VaultsLifecycleSections.test.tsx` and `usePendingDeposits.test.tsx`. No dangling references, per the zero-dead-code policy. Verify the post-broadcast path still reaches the deposit-progress view with no gap.
- [ ] If kept: the BTC icon renders at the diameter in node `13603-77420` rather than `max-w-[160px]`.
- [ ] If kept: the footnote is shortened to the design's single sentence in American English, and a new test asserts the rendered footnote — none exists today.
- [ ] If kept: heading, body and the `Done` label stay byte-identical to `copy.ts:470-475`.

## Key files

- `services/vault/src/components/deposit/BroadcastSuccessModal/index.tsx`
- `services/vault/src/components/simple/PendingDepositModals.tsx`
- `services/vault/src/hooks/deposit/useBroadcastModal.ts`
- `services/vault/src/components/vaults/VaultsLifecycleSections.tsx`
- `services/vault/src/components/Activity/ActivityListWithRefund.tsx`

Verified against `8f62bcfa`.

jeremy-babylonlabs:
Closes #2473

## What

The Pre-Pegin broadcast modal now renders the design in [node 13603-77420](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=13603-77420).

- The BTC mark renders at its 100px diameter instead of filling the dialog up to a 160px maximum width.
- The footnote is the design's single sentence: "Once confirmed, you'll be asked to submit a WOTS key, sign payout authorizations, and activate your BTCVault." It replaces three sentences that restated the deposit stepper and told the depositor to check back.
- The dialog is 564px wide on desktop with a 24px radius, 40px of top padding, 24px at the sides and bottom, and a 40px gap above the action. The action keeps its 40px height and takes the design's 8px radius.
- A new test asserts the rendered heading and footnote. The component had no test before.

Heading, body and the `Done` label are byte-identical to what shipped.

## Why

The modal is reached from `PendingDepositModals` after a successful broadcast, and it had drifted from its design in two visible ways: the Bitcoin mark rendered roughly 1.7 times its intended diameter, and the second paragraph ran about three times the length of the designed sentence, duplicating the stepper that sits behind the modal.

The issue also asked whether the modal should exist at all. It is not dead code: `PendingDepositModals` renders it on `broadcastModal.successOpen`, and `useBroadcastState` documents it taking over after a successful broadcast. This change implements the issue's "if kept" branch.

## Approaches

**Delete the modal.** The issue's other branch, raised by design as an open question. Rejected: the modal is rendered from `PendingDepositModals` on `broadcastModal.successOpen`, mounted from both `ActivityListWithRefund` and `VaultsLifecycleSections`, so removing it changes the depositor's post-broadcast experience. That is a product decision rather than a cleanup, and it needs the keep-or-remove decision recorded on the issue first.

**Keep the long footnote and only resize the icon.** Rejected: the footnote is half of what the design changed, and its "check back here" instruction repeats the stepper the depositor is already looking at.

**Restyle the shared `ResponsiveDialog` in core-ui.** Rejected: every dialog in the app consumes it, so a width, radius or padding change there reaches screens this issue never looked at. The geometry is set from the call site instead, on the props the component already exposes.

**Split the body string to colour the emphasised span.** The design paints part of the second sentence in `text/primary`. Rejected: the acceptance criteria require the body string to stay byte-identical, and segmenting it into coloured spans breaks that. It needs its own change and a copy decision.

**Reach the inner dialog box through `dialogClassName` rather than a descendant selector.** `Dialog` exposes that prop and `ActivateConfirmationModal` already uses it. Rejected: `ResponsiveDialog` forwards its rest props to `MobileDialog`, which spreads them onto a plain `div`, so the prop becomes an unrecognised DOM attribute and React warns on every mobile render. The scoped `[&_.bbn-dialog]` variants compile to two-class selectors that win on specificity alone and match nothing in the mobile tree.

## Not in this PR

- Confirm the mobile rendering against design, and add a mobile frame to the node if the scale should differ below 640px (pre-review #5).

## Pre-review

<details>
<summary>Pre-review: 7 findings · 4 fixed · 0 moot · 1 follow-up · 2 declined · 0 open (0 merge-blockers) · 0 undecided · lint and test passed</summary>

|---|---|---|---|
| 1 | The unscoped `p-0` in the `ResponsiveDialog` className zeroes the mobile bottom sheet's padding, so content collides with the close button below 640px | normal | fixed |
| 2 | The unscoped `w-[564px]` leaves the mobile sheet left-anchored at 564px on viewports between 564px and 640px | normal | fixed |
| 3 | The doc comment still says the copy makes the HTLC-secret step explicit, which the shortened footnote no longer does | normal | fixed |
| 4 | The explicit `text-base` on both `Text` elements is redundant because the `body1` variant already applies it | normal | fixed |
| 5 | Dropping the `sm:` variants gives the mobile modal the desktop type scale, and no mobile design frame was checked | normal | follow-up: confirm the mobile rendering with design |
| 6 | The `[&_.bbn-dialog]` variants reimplement the `dialogClassName` prop `Dialog` already exposes | normal | declined: the prop leaks onto the mobile DOM node through rest props and warns; the scoped variants are specificity-guaranteed and breakpoint-safe |
| 7 | The footnote test asserts the sentence as a literal rather than against the copy constant | normal | declined: a literal is the point; asserting against the same constant the component renders is a mirror test |

Two runs. The first raised the seven findings above and the four fix-now items were applied; the second re-checked the changed file, confirmed all four fixes and found no new defect.

The first run's `test` target failed on two files this change does not touch, `src/dev/__tests__/demoArtifactDownload.test.ts` and `src/utils/__tests__/scriptPubKeyHexToBtcAddress.test.ts`, both on a 5s test timeout and a 10s hook timeout against 424s of environment time. Both passed in isolation, and the second run's full affected target passed with 4205 tests, so they were load-induced.

</details>

Closes https://github.com/babylonlabs-io/babylon-toolkit/issues/2473


jonybur (approved):
Looks good to me.

gbarkhatov (approved):

PR #2567, opened 2026-09-18
fix(vault): stop deleting pending deposits when their stored blob cannot be read

### LINKED ISSUE #2476: A single unparseable pending-deposit blob wipes every pending deposit for that address
## Context

`getPendingPegins` reads the browser's only record of in-flight deposits. Each `PendingPeginRequest` holds the unsigned Pre-Pegin transaction hex and the selected UTXOs — the material the resume and broadcast paths need. There is no server-side copy for a deposit the indexer has not yet returned.

Its outer `catch` deletes the entire storage key for that address:

```ts
} catch (error) {
  logger.error(…, { context: "[peginStorage] Failed to parse pending pegins - Clearing corrupted data" });
  try {
    localStorage.removeItem(getStorageKey(ethAddress));
  } catch (clearError) { … }
  return [];
}
```

`services/vault/src/storage/peginStorage.ts:330` (function), `:373-377` (the wipe).

## Why this is wrong

Per-entry corruption is already handled correctly and gracefully — `hasValidSecurityFields` filters a bad entry, logs it, and keeps every sibling. So the `catch` is not the per-entry guard; it fires only on a failure that reaches the whole blob, chiefly `JSON.parse` throwing. Realistic causes: a write truncated by a storage quota, a partially-flushed write, or any other single-byte corruption of the serialized array.

The response to "I could not read this blob" is to destroy it. That converts a possibly recoverable string — one that could be inspected, salvaged, or exported by hand — into permanent loss of every pending deposit for that address, silently, on a plain page render.

It is also a silent fallback on a value-bearing path, which `CLAUDE.md` forbids: a missing value here indicates a real problem and must be surfaced, not papered over with `[]`.

## Expected

- Never delete the key on a read failure. Reading must not be destructive.
- Return empty and surface the failure to the user, so a stuck-looking Vaults page is distinguishable from a genuinely empty one.
- Preserve the unparseable string so it can be recovered — for example move it aside under a quarantine key rather than removing it.

## Acceptance criteria

- [ ] A `JSON.parse` failure in `getPendingPegins` leaves the original stored string retrievable, asserted by a test that seeds malformed JSON and reads the raw value back afterwards.
- [ ] That failure surfaces to the user rather than rendering as an empty list.
- [ ] Existing per-entry filtering behavior is unchanged: one malformed entry still drops only itself and keeps its siblings (regression test).
- [ ] No path in `getPendingPegins` calls `localStorage.removeItem`.

## Key files

- `services/vault/src/storage/peginStorage.ts`
- `services/vault/src/storage/__tests__/peginStorage.test.ts`

Found while working on #2422. Verified on `main` at `068a1dd8`.

jeremy-babylonlabs:
Closes #2476

## What

`getPendingPegins` no longer deletes anything. When the stored JSON for an address cannot be read, it throws a typed `PendingPeginStorageReadError` carrying the address, the unparseable string, an `errorCode` and the cause's name, and the stored value stays exactly where it is.

- `services/vault/src/storage/peginStorage.ts` — the `catch` that called `localStorage.removeItem` is gone. One `try` now covers `getItem`, `JSON.parse` and the array check, so blocked storage is a read failure too (`raw` is `null`, `errorCode` is `PENDING_PEGIN_STORAGE_BLOCKED`) rather than an uncaught `DOMException`. A module-private `readPeginsForUpdate` makes `updatePendingPeginStatus`, `removePendingPegin` and `markRefundBroadcast` return without writing on an unreadable blob, so a status flip after a broadcast cannot report a successful on-chain action as failed and cannot overwrite the record. `addPendingPegin` keeps propagating the error. Per-entry validation (`hasValidSecurityFields`) is unchanged.
- `services/vault/src/storage/usePeginStorage.ts` — the typed error becomes `storageReadError` hook state with `pendingPegins: []`, exposed on the return. One `logger.error` per address reports it, deduplicated through a module-level map of the failure last reported for each address, so the several mounted instances of this hook do not each bill a Sentry event and a corrupt blob that later becomes blocked storage is still reported; the read site itself emits no telemetry, because it runs on every polling tick.
- `services/vault/src/hooks/useVaultDeposits.ts`, `services/vault/src/hooks/usePendingDeposits.ts`, `services/vault/src/hooks/useVaultsPageEmptiness.ts`, `services/vault/src/components/pages/VaultsPage.tsx`, `services/vault/src/copy.ts` — the error reaches the existing partial-load `Notification`, which now also renders over the empty state, with a new string next to `partialLoadError`.
- `services/vault/src/hooks/deposit/usePendingVaultOverlapCheck.ts`, `services/vault/src/services/activity/pendingActivities.ts` — each catches only `PendingPeginStorageReadError` at its own call site, so an unreadable blob does not replace the Activity page or abort a new deposit through the app-level error boundary.
- `updatePendingPeginStatus` and `markRefundBroadcast` now read the raw stored array, as `removePendingPegins` does, so a status flip writes hidden siblings back untouched and never deletes the key because every entry was hidden. `readPeginsForUpdate` is gone with its last callers. An empty stored string is nothing stored for both readers; `getPendingPegins` documents its throw; the skip-log no longer stringifies a stored id, so a crafted id cannot escape the read guard as an untyped error.
- `services/vault/src/hooks/deposit/useDepositFlow.ts` — the stored records are read once more immediately before the on-chain registration. An unreadable blob stops the deposit there, with copy that says what to do, instead of registering a vault whose resume record can never be saved. `mapDepositError` maps the typed error to that copy.
- `services/vault/src/hooks/useVaultsPageEmptiness.ts` — an unreadable record counts as a failed source: with nothing else to show the page reports it instead of claiming an empty account under a warning that the records could not be read; with rows to show, or alongside a failed remote read, it drives the warning.
- `usePendingVaultOverlapCheck` returns `"unreadable"` instead of an empty result when the records cannot be read, and the deposit form says the reserved-input check did not run. The three `getPendingPegins` call sites that absorb the typed error now have a test each, with the mock factories exporting `PendingPeginStorageReadError`.

## Why

`getPendingPegins` holds the browser's only record of in-flight deposits: the unsigned Pre-PegIn transaction hex and the selected UTXOs, with no server-side copy for a deposit the indexer has not yet returned. Its outer `catch` deleted the whole storage key for the address. A single byte of corruption — a write truncated by a storage quota, a partially flushed write — therefore turned "I could not read this" into permanent loss of every pending deposit for that address, silently, on a plain page render, while the page rendered as if the depositor simply had none.

Per-entry corruption was already handled correctly: one malformed entry is filtered and its siblings are kept. The outer `catch` was never that guard.

Fixes https://github.com/babylonlabs-io/babylon-toolkit/issues/2476.

## Approaches

**Chosen: a typed, non-destructive read error that the existing partial-load surface reports.** Reading stays a pure read; the depositor is told the records could not be read and that nothing was deleted; the raw string is carried on the error so a support path can recover it.

**Rejected — quarantine the blob under a second key.** The issue's "Expected" suggests it, but it adds a write to a read path and a second storage key to maintain, and it still returns `[]` silently to the caller.

**Rejected — return `{ pegins, error }` from `getPendingPegins`.** Every caller's shape changes for the same outcome the typed error already gives.

**Rejected — return `[]` for blocked storage, as the sibling `addressScreeningStorage` does.** A silent `[]` on a value-bearing path is what the repository rules forbid; blocked storage and a corrupt blob now reach the same surface and are separated in telemetry by `errorCode` instead.

**Rejected — keep the mutators propagating the error.** Three of them run after an irreversible action, where a throw reported a successful broadcast, presign or refund as a failure. They return without writing instead, which preserves the actual invariant: an unreadable blob is never overwritten.

## Not in this PR

- A quarantine or recovery path for a blob that cannot be read. The copy can only promise that nothing was deleted and that a reload may clear a transient failure; a persistent one blocks new deposits from that browser until the records read again.
- Giving the Activity feed its own read-failure surface; its rows still vanish with no on-page indication, though the failure is reported in telemetry (pre-review #4)
- Collapsing the four duplicated `instanceof PendingPeginStorageReadError` guards into one helper (pre-review #5)
- Telling the depositor about a remote gap that coincides with an unreadable blob; the notification body is an exclusive ternary (pre-review #9)
- Distinguishing a read failure from a write failure in the deposit-creation and refund catches, which currently show copy about a failed save (pre-review #19)
- Keying the telemetry dedupe guard's clear on the address that was actually read, so switching wallets away from an unreadable address and back does not re-report the same blob (pre-review #20)
- Age-based pruning and the stale-deposit dismiss, which belong to https://github.com/babylonlabs-io/babylon-toolkit/issues/2422

## Pre-review

<details>
<summary>Pre-review: 26 findings · 19 fixed · 0 moot · 5 follow-up · 2 declined · 0 open (0 merge-blockers) · 0 undecided · lint passed · test passed</summary>

|---|---|---|---|
| 1 | The typed read error escapes the three mutators at call sites that run after an irreversible action, so a successful broadcast, presign or refund is reported as failed | merge-blocker | fixed |
| 2 | `localStorage.getItem` sat outside the `try`, so blocked storage reached the app-level error boundary as a full-page crash | merge-blocker | fixed |
| 3 | The new throw path emitted no telemetry at all | normal | fixed |
| 8 | The read failure billed one Sentry event per read, and the first dedupe guard was per hook instance while the page mounts two | normal | fixed |
| 12 | `String(error)` at the throw site put a fragment of the stored blob into a breadcrumb | normal | fixed |
| 14 | The per-read breadcrumb saturates Sentry's 100-slot breadcrumb ring | normal | fixed |
| 15 | Blocked storage and a corrupt blob arrived in Sentry as one indistinguishable issue | normal | fixed |
| 21 | The dedupe guard keyed on the address alone, so an escalation from a corrupt blob to blocked storage was never reported | normal | fixed |
| 22 | `causeName` resolved to `"unknown"` for blocked storage under the repository's own jsdom test environment, and no test pinned it | normal | fixed |
| 6 | A test named for reporting the unreadable record pinned only the error message | normal | fixed |
| 10 | A test comment claimed the full-page and partial error flags can never both fire | normal | fixed |
| 11 | A substring assertion passed whether one notification body rendered or both | normal | fixed |
| 13 | Four comments described behaviour the code no longer has | normal | fixed |
| 16 | The dedupe guard's reset and per-address keying survived the suite untested | normal | fixed |
| 17 | Three mutator tests sat in the describe of a function they do not exercise | normal | fixed |
| 18 | `addPendingPegin`'s `@throws` omitted the typed read error | normal | fixed |
| 23 | Two fixtures paired a raw string with a `SyntaxError` message `JSON.parse` never produces for it | normal | fixed |
| 24 | A test name claimed a post-broadcast scenario the body does not set up | normal | fixed |
| 25 | The two `errorCode` values were inline string literals | normal | fixed |
| 4 | The Activity feed drops pending rows on a read failure with no error channel | normal | follow-up: give it its own read-failure surface |
| 5 | The same four-line `instanceof` recovery block is duplicated in four files | normal | follow-up: collapse into one helper that cannot become a silent accessor |
| 9 | With rows, a failed remote source and an unreadable blob, the remote gap is never mentioned | normal | follow-up: render both notification bodies |
| 19 | The deposit-creation and refund catches show copy about a failed save for a read failure | normal | follow-up: separate the two cases at both sites |
| 20 | Switching wallets away from an unreadable address and back re-reports the same blob | normal | follow-up: key the guard's clear on the address that was read |
| 7 | `PendingPeginStorageReadError.raw` has no production reader | normal | declined: deliberate support-path field, carried by the error and asserted in tests |
| 26 | The two `errorCode` constants are exported with no importer outside their own file | normal | declined: forward-looking surface for comparing `errorCode`; knip does not flag it |

</details>


services/vault/src/storage/peginStorage.ts
```diff
@@ -345,69 +346,97 @@ function backfillBuildVaultCoreVersion(entry: unknown): unknown {
...
+  const migrated = parsed.map(backfillBuildVaultCoreVersion);
+
+  // Filter out entries whose security-critical fields (unsignedTxHex,
+  // selectedUTXOs) fail a strict format check. A tampered entry would
+  // otherwise feed malformed hex into downstream consumers.
+  const validated = migrated.filter((entry): entry is PendingPeginRequest => {
+    if (hasValidSecurityFields(entry)) return true;
+    const maybeId =
+      entry && typeof entry === "object" && "id" in entry
+        ? String((entry as { id: unknown }).id)
```
jonybur, line 420: **Per-entry validation now runs outside the read guard, and String(entry.id) throws an untyped TypeError on a crafted id, so it crashes the app on every load.**

This PR narrowed the try in getPendingPegins to getItem, JSON.parse and the array check, so the migrate/filter/normalize steps now run unguarded. The skip-log line calls String((entry as { id: unknown }).id) on an entry that failed validation; a stored entry like {"id":{"toString":0}} makes String() throw "Cannot convert object to primitive value". That is not a PendingPeginStorageReadError, so readPendingPegins in usePeginStorage rethrows it from the useState initializer and the root error boundary takes the app down on every load for that address, where previously the same data was caught. The updated hasValidSecurityFields JSDoc says exactly this must not happen. Use typeof id === "string" ? id : "unknown" for the log value and add a test with this fixture asserting the entry is skipped and its siblings kept.

jonybur (commented):
Comments inline, plus:

- `services/vault/src/storage/peginStorage.ts:520` - **An unreadable blob makes every later addPendingPegin for the address throw after on-chain registration, so new deposits lose their stamped resume records.**
  
  Since nothing ever rewrites or quarantines an unreadable blob, addPendingPegin calls getPendingPegins, which throws PendingPeginStorageReadError for that address forever. In useDepositFlow this happens after the vault is registered on-chain, so every subsequent deposit from that browser ends with the terminal "record not saved" warning and no local record carrying the build-version and participant-key stamps, which the resume path then skips checking. Before this PR only the existing records were lost; now every future deposit is degraded as well. The new storageReadError copy ("…could not be read. Nothing was deleted.") also gives the user no next step. Consider stopping the deposit before registration when the blob is unreadable, or giving the user an actionable recovery path in the copy.

gbarkhatov (approved):

services/vault/src/storage/peginStorage.ts
```diff
@@ -345,69 +346,97 @@ function backfillBuildVaultCoreVersion(entry: unknown): unknown {
...
   if (!ethAddress) return [];
 
+  let stored: string | null = null;
+  let parsed: unknown[];
   try {
-    const key = getStorageKey(ethAddress);
-    const stored = localStorage.getItem(key);
-    if (!stored) return [];
+    stored = localStorage.getItem(getStorageKey(ethAddress));
+    if (stored === null) return [];
```
jonybur, line 401: **An empty stored string is unreadable to getPendingPegins but empty-and-removed to readStoredEntries, so one module holds two definitions of "nothing stored".**

The guard changed from `if (!stored)` to `if (stored === null)`, so a key holding `""` now reaches `JSON.parse("")`, throws, and becomes a `PendingPeginStorageReadError` that is never cleared — the new test "reports an empty stored string as unreadable" pins that. `readStoredEntries` at line 510 was left on the old `!stored` test, so for that same value `removePendingPegins` takes the `empty` branch at lines 716-719, dispatches a storage-update event and returns `"removed"` without writing anything. A truncated write therefore leaves an address permanently warned while the one path that could clear the value reports a removal that never happened, and `usePeginStorage` re-reads on that event and throws again. Pick one definition: either restore `!stored` in `getPendingPegins`, or move `readStoredEntries` to `stored === null` and let its parse branch classify `""` as unreadable.

services/vault/src/storage/peginStorage.ts
```diff
@@ -345,69 +346,97 @@ function backfillBuildVaultCoreVersion(entry: unknown): unknown {
...
+      "name" in cause &&
+      typeof cause.name === "string"
+        ? cause.name
+        : "unknown";
+  }
+}
+
 /**
  * Get all pending peg-ins from localStorage for an address
  * Pure read function - no side effects
```
jonybur, line 392: **getPendingPegins' JSDoc still documents a non-throwing pure read with no @​throws, after this PR made it throw.**

The doc block above the function is untouched: "Get all pending peg-ins from localStorage for an address / Pure read function - no side effects", with a plain `PendingPeginRequest[]` return and no `@​throws`. This is the one function whose contract the PR inverts — every one of its five callers now has to decide whether to guard — and `addPendingPegin` did get `@​throws PendingPeginStorageReadError` at line 579 for the same error, so the omission is inconsistent inside the file. A future caller reading this signature has nothing telling them that an unguarded call reaches the app-level error boundary. Add the `@​throws` clause naming `PendingPeginStorageReadError`, and point at `readPeginsForUpdate` and `usePeginStorage`'s `readPendingPegins` as the guarded helpers to reuse; CLAUDE.md's CODE STYLE rule on updating docs that describe old behaviour applies here.

services/vault/src/hooks/useVaultsPageEmptiness.ts
```diff
@@ -78,7 +82,9 @@ export function useVaultsPageEmptiness(deposits: VaultsPageDeposits): {
   const hasError =
     isConnected && !isLoading && !hasAnythingToShow && anySourceFailed;
   const hasPartialError =
-    isConnected && !isLoading && hasAnythingToShow && anySourceFailed;
+    isConnected &&
+    !isLoading &&
+    ((hasAnythingToShow && anySourceFailed) || Boolean(storageReadError));
   const isEmpty =
```
jonybur, line 88: **storageReadError is left out of anySourceFailed, so the page claims an empty account on the back of a failed local read.**

`storageReadError` feeds only `hasPartialError` at line 87; `anySourceFailed` at lines 79-81 still covers just the two queries, so `hasError` is false, `isEmpty` is true, and `VaultsPage.tsx:167` falls through to `renderEmptyState(false)`. A depositor whose stored records could not be read therefore sees "Deposit BTC to create your first BTCVault" directly under a warning saying those records could not be read — the page asserts they have no deposits using the one source that just failed. The header of this file states the opposite rule for the remote case at lines 15-19 ("an empty account must never be claimed on the back of a failed read"), and line 26 was extended for the warning without carving out this exception. Either fold `storageReadError` into the emptiness gate so the empty-account card is suppressed, or narrow that invariant in the prose to say the local read is deliberately exempt.

jonybur (commented):
Comments inline, plus:

- `services/vault/src/storage/peginStorage.ts:655` - **updatePendingPeginStatus and markRefundBroadcast write back the validated view, so hidden-but-stored entries are dropped and the key is deleted when every entry is hidden.**
  
  Both mutators now read through `readPeginsForUpdate`, which returns `getPendingPegins`' output — the entries that passed `hasValidSecurityFields` — and then hand that list straight to `savePendingPegins` (here and at line 780). Entries the read hides are not tampered data: a legacy record written before the build-version stamps existed fails the check at lines 226-247, and `removePendingPegins` documents at lines 691-694 that exactly those siblings must be written back untouched, which is why it goes through `readStoredEntries` instead. So a status flip on one vault silently drops every hidden sibling, and when every stored entry is hidden the mapped list is empty, `persistStoredEntries` takes the `entries.length === 0` branch at line 479 and calls `localStorage.removeItem` — the same destructive delete this PR removed from the read path, reached through a presign, broadcast or refund instead. The behaviour predates this change, but these are the lines the PR rewrote, and the fix is to give both mutators the raw-array treatment `removePendingPegins` already has.

- `services/vault/src/hooks/__tests__/useVaultDeposits.test.ts:30` - **The peginStorage mock factories omit PendingPeginStorageReadError, so the three new catch branches are untested and cannot be tested as mocked.**
  
  This factory returns only `getPendingPegins`, and `pendingActivities.test.ts:9-11` does the same, while `useVaultDeposits.ts:91`, `pendingActivities.ts:71` and `usePendingVaultOverlapCheck.ts:71` all use `PendingPeginStorageReadError` as a value in an `instanceof` check. A test that made the mocked read throw would not exercise the branch: accessing an export the factory never returned fails on the mock itself. As it stands no test in the suite makes `getPendingPegins` throw at any of these three call sites, so the error handling this PR exists to add ships unexercised — including the rethrow that is supposed to let a non-typed error (the `String(entry.id)` TypeError, for instance) propagate. Add `PendingPeginStorageReadError` to both factories and one test per call site: the typed error is absorbed, any other error still escapes.

services/vault/src/hooks/deposit/usePendingVaultOverlapCheck.ts
```diff
@@ -60,14 +64,20 @@ export function usePendingVaultOverlapCheck({
         // Let the real signing path surface insufficient-funds errors.
         return null;
       }
+      let pendingPegins: PendingPeginRequest[] = [];
+      try {
+        pendingPegins = getPendingPegins(ethAddress ?? "");
+      } catch (error) {
+        if (!(error instanceof PendingPeginStorageReadError)) throw error;
```
jrwbabylonlab, line 71: The description says this returns `"unreadable"` and that `useDepositFlow` stops the deposit before registration, and that `readPeginsForUpdate` is gone. None of that is in 80203d2c (`useDepositFlow.ts` isn't touched; `readPeginsForUpdate` is still at `peginStorage.ts:620`). As pushed, an unreadable blob is swallowed here, the vault registers, then `addPendingPegin` throws, so every new deposit from that browser loses its resume record behind the "free up storage" warning. On main the blob was deleted, so this is a regression. Missing push?

gbarkhatov (approved):

PR #2568, opened 2026-09-18
feat(vault): add a vault-selection step to the withdraw flow

### LINKED ISSUE #2432: Batch withdrawal: let the Vaults page withdraw more than one vault
## Context

Batch withdrawal is designed but not built. Design: Figma `13604-77935`.

## Current baseline

The plumbing is already multi-vault; only the UI is single-vault.

- One Withdraw button per row, calling `onWithdraw(vault.vaultId)` with a single id — no checkbox or multi-select markup (`services/vault/src/components/vaults/VaultsActiveSection.tsx`)
- `handleWithdrawRow` always sets a one-element array: `setWithdrawVaultIds([vaultId])` (`services/vault/src/components/pages/VaultsPage.tsx:83-85`)
- The flow already takes an array: `preSelectedVaultIds: string[]` → `getEffectiveVaultSelection` (`services/vault/src/components/simple/WithdrawFlow/index.tsx:34,44,74-78`)
- The contract call already takes an array: `withdrawSelectedCollateral(walletClient, chain, vaultIds: Hex[])` (`services/vault/src/applications/aave/hooks/useWithdrawCollateralTransaction.ts:95`)

## Changes

**Expected:** row selection on the Vaults page and a "Withdraw selected" trigger that passes more than one id into the existing flow.
**Where:** `VaultsActiveSection.tsx`, `VaultsPage.tsx:83-85`

No service or contract change is needed.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1788446004964119)

> the batch withdrawal design is not in yet?
>
> — reply: batch withdrawal design is ready, [Figma link](https://www.figma.com/design/NgDfn7fVnkQZP0XH9Uki63/TBV-v.3--Premium-Design-?node-id=13604-77935)

</details>

jeremy-babylonlabs:
Closes #2432

## What

The Withdraw modal opens on a new selection step before Review, so one withdrawal can release several BTCVaults in a single transaction.

- `useWithdrawFlow.ts` gains a `SELECT` step ahead of `REVIEW`: the flow is now `SELECT → REVIEW → PROGRESS`, `reset` returns to `SELECT`, and `goToReview` advances out of it.
- `WithdrawSelectContent.tsx` (new) is the step itself: one row per selectable BTCVault in liquidation order, each with its amount, its liquidation ordinal and a checkbox; a risk callout above the list when the projected health factor lands in the at-risk band or below the on-chain floor; and a full-width `Withdraw <sum> BTC` action.
- `WithdrawFlow/index.tsx` holds the live selection (seeded from the row the user clicked) and the at-risk acknowledgement. The acknowledgement is keyed to the health factor as displayed at two decimals: a position refetch that moves the raw float without changing the shown number keeps the tick, a change to the shown number drops it in the same render, and moving away and back to the same shown number keeps it. Review shows the same acknowledgement row when the projection is at risk and blocks Confirm until it is ticked, so a projection that moves after Select cannot be confirmed unacknowledged. Review carries a back arrow to Select while no submit is in flight. `index.tsx` also and feeds the live selection into the existing amount and health-factor memo through the unchanged `getEffectiveVaultSelection`. The acknowledgement covers exactly the health factor the risk card named, so any move in that projection — a changed selection, a price move, or the projection leaving the at-risk band and returning — asks again. The list also now requires `inUse`, matching the per-row Withdraw button's own gate. Review's confirm-time pinning is untouched.
- `copy.ts` gains `COPY.withdraw.select.*`.
- The real-wallet E2E action and the withdraw page object gained a step that passes the new screen.

The Vaults page is unchanged: one Withdraw button per row, no checkboxes. No service or contract change — `withdrawSelectedCollateral(walletClient, chain, vaultIds: Hex[])` already took an array.

## Why

Batch withdrawal was designed but not built ([#2432](https://github.com/babylonlabs-io/babylon-toolkit/issues/2432)). Everything below the UI was already multi-vault: `getEffectiveVaultSelection` takes an id array, and the contract call takes `bytes32[]`. Only the entry point was single-vault — `handleWithdrawRow` always built a one-element array — so releasing three BTCVaults meant three transactions and three gas payments.

## Approaches

**Selection inside the modal (chosen).** The design (Figma section `13604-77935`) keeps the Vaults page rows as plain per-row buttons and puts the multi-select in a new first step of the modal. Clicking a row's Withdraw pre-checks that vault and the user ticks more there. This keeps the page's row affordance unchanged and gives the selection its own surface, with the health-factor projection updating as the selection changes.

**Checkboxes on the Vaults page plus a "Withdraw selected" bar — rejected.** This is what the issue text proposes, but not what the design shows: frame `13604-79890` keeps plain per-row buttons and no checkboxes. It would also put a selection affordance on a list that is otherwise read-only.

**Merging the selection into the Review step — rejected.** Review is the confirm surface and pins every input at confirm time, because the withdraw marks its vaults pending and awaits a refetch before the Progress step runs. A mutable selection there re-opens the flash-to-zero class of bug that pinning exists to prevent.

## Not in this PR

- Drag-to-reorder inside the modal. The Figma subtitle's second sentence ("Drag to change the liquidation priority.") is therefore omitted; only the first sentence ships.
- The Reorder button and `ReorderVaultsModal` on the Vaults page.
- Mobile layout beyond what `V3ModalShell` already does.
- Confirm on-chain that `withdrawCollaterals(bytes32[])` is order-independent; this is the first production call that passes it more than one element (pre-review #5).

The four new members of `e2e/pages/WithdrawModal.ts` have no callers, like the health-factor getters already in that file: the page object is provisioned ahead of the per-flow withdraw ticket that will use it. The real-wallet CLI action drives the same screen through its own locators and is a separate harness, not a duplicate.

## Pre-review

<details>
<summary>Pre-review: 11 findings · 7 fixed · 0 moot · 1 follow-up · 3 declined · 0 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | The real-wallet E2E withdraw action cannot pass the new at-risk acknowledgement gate, so an at-risk withdrawal times out with a wrong diagnosis. | merge-blocker | fixed |
| 2 | The per-row checkbox data-testid is a load-bearing real-wallet E2E hook with no inline comment pointing at its consuming action file. | normal | fixed |
| 3 | COPY.withdraw.select.continueAnyway is Title Case against copy.ts's own sentence-case rule for in-flow controls. | normal | fixed |
| 4 | `acknowledged` is not reset when the projected health factor changes without a selection change, so a stale acknowledgement can carry a number the user never saw. | normal | fixed |
| 5 | Whether `withdrawCollaterals(bytes32[])` is order-independent is unverified, and this change is the first to pass it more than one element. | normal | follow-up: confirm on-chain before merging; sort the array on the client if it is not. |
| 6 | The new `inUse` filter and the acknowledgement reset are production lines with no covering test. | normal | fixed |
| 7 | The four new WithdrawModal page-object members have no callers anywhere in the repo. | normal | declined: the file is deliberately pre-provisioned ahead of the per-flow withdraw ticket and its existing getters have no callers either; this change's brief required the step. |
| 8 | passSelectStep's timeout message and its docstring name only the health-factor floor, but an empty selection can also hold the submit disabled. | normal | fixed |
| 9 | The acknowledgement tick uses Playwright's check() with no force fallback and no catch. | normal | declined: the input is the topmost hit target, so a plain check() lands; force would skip the actionability assertion that should fail loudly if core-ui's overlay changes. |
| 10 | The acknowledgement-reset test would also pass under the mechanism it replaced, so it did not pin the price move that motivated #4. | normal | fixed |
| 11 | rerenderWith merges against the first render's props, so chained rerenders in one test would not accumulate. | normal | declined: merging against the initial render is a coherent contract for a one-call helper; the accumulating version needs mutable state for a second caller that does not exist. |

</details>


services/vault/src/components/simple/WithdrawFlow/index.tsx
```diff
@@ -144,6 +159,14 @@ function WithdrawFlowContent({
     reviewCurrentHealthFactor,
   ]);
 
+  // The risk card names one health factor, so an acknowledgement only covers
+  // that number. Drop it whenever the projection moves — a changed selection, a
+  // price move, or the projection leaving the at-risk band and returning — so
+  // the user accepts what is on screen rather than what used to be.
+  useEffect(() => {
```
jonybur: **The at-risk acknowledgement resets on every sub-display change of the raw projected health factor, so the 30s position poll can untick it while the shown value is unchanged.**

This effect clears `acknowledged` whenever the float `projectedHealthFactor` changes. That value comes from `currentHealthFactor` (`wadToNumber(accountData.healthFactor)`), and `useAaveUserPosition` refetches it every `POSITION_REFETCH_INTERVAL_MS` (30s). With debt outstanding, interest accrual moves it slightly on almost every refetch, but the risk card shows only `formatHealthFactor(...)` at two decimals. So in the at-risk band a user's "Continue anyway" tick gets silently cleared and Continue greys out again, while the number they accepted is still on screen. Because the reset runs in an effect, there is also one render where a genuinely new number is paired with the old acknowledgement. Consider storing the acknowledged display value, e.g. `acknowledgedFor = formatHealthFactor(projectedHealthFactor)` on tick. Then derive `acknowledged` during render by comparing it with the current formatted value, and remove the effect.

services/vault/src/components/simple/WithdrawFlow/index.tsx
```diff
@@ -170,6 +193,20 @@ function WithdrawFlowContent({
   return (
     <V3ModalShell open={open} onClose={onClose}>
```
jonybur: **The flow is now Select then Review, but Review has no way back to Select; V3ModalShell's onBack is never passed.**

`useWithdrawFlow` only moves forward (`goToReview`, `goToProgress`), and `V3ModalShell` is rendered without `onBack`, although the shell already supports it and renders a back chevron. A user who notices a wrong selection on Review, or whose submit fails and leaves them on Review, can only close the modal. Closing unmounts the flow and throws the selection away, so they must start again from a single pre-checked row. Unless the design leaves this out on purpose, add a `goToSelect` step and pass `onBack` while on Review and not processing.

services/vault/e2e/real/actions/withdraw.ts
```diff
@@ -20,14 +20,14 @@
...
+ *   Review ("Confirm") → one MetaMask tx → "Withdrawal initiated" → "Done".
  *
  * The row's Withdraw button IS the eligibility gate: the app disables it for a paused protocol, a vault
  * that is not in use, a demo (`displayOnly`) row and an optimistic (`lifecycle === "activating"`) one — see
- * VaultsActiveSection. Health-factor gating surfaces one step later, on the Review screen.
+ * VaultsActiveSection. Health-factor gating surfaces one step later, on the selection screen, and again
+ * on Review.
  *
  * Default withdraws ONE vault (the first withdrawable), keeping the position alive for reuse.
  * `--withdraw-all` repeats the whole row → Review → Done cycle per withdrawable vault: v3 opens the
```
jonybur: **Header and runWithdrawFlow docs still say releasing several vaults needs several transactions because there is no multi-select, which this PR makes untrue.**

The file header (lines 32-36) still says v3 opens the flow with exactly one preselected vault, "so releasing several means several transactions, not one multi-select". The `runWithdrawFlow` docstring (lines 370-374) makes the same claim. This PR adds exactly that multi-select, and it edited the surrounding header lines without updating these. CLAUDE.md asks that comments describing old behaviour be updated after a change. Reword them to say that `--withdraw-all` deliberately runs one transaction per vault even though the modal now supports batch selection.

jonybur (commented):
Comments inline.

gbarkhatov (approved):

gbarkhatov (dismissed):

jrwbabylonlab (dismissed):

gbarkhatov (approved):

gbarkhatov:
@jeremy-babylonlabs verify pr failing

gbarkhatov (approved):

services/vault/src/components/simple/WithdrawFlow/WithdrawReviewContent.tsx
```diff
@@ -205,23 +212,45 @@ export function WithdrawReviewContent({
...
+            </div>
           )}
 
           {/* This control's data-testid is a real-wallet E2E hook (e2e/real/actions/withdraw.ts) — carry it over if you move or rename the element. */}
           <Button
             variant="contained"
             color="secondary"
             className="w-full"
-            disabled={isProcessing || wouldBreachHF || hubBlockMessage !== null}
+            disabled={
```
jonybur: **Confirm has no empty-selection gate, so a selection that empties while Review is open can submit withdrawCollaterals([]).**

The new Select step refuses to continue when nothing is checked, but Confirm's disabled expression still only covers isProcessing, wouldBreachHF, hubBlockMessage and the acknowledgement. effectiveSelectedVaultIds is re-filtered against the in-use vault list on every poll, so if the selected vaults leave that set while Review is on screen the selection collapses to an empty array: selectedBtc becomes 0 and the projected health factor becomes exactly the current one, so neither the breach nor the at-risk branch fires and Confirm stays enabled. handleConfirm then calls executeWithdraw([]), and neither the hook nor withdrawSelectedCollateral guards the length, so a real transaction goes out that releases nothing and the flow advances to "Withdrawal initiated". Review is also the one screen that never enumerates which vaults it is releasing — it shows only a sum — so there is nothing on screen to contradict it. Add effectiveSelectedVaultIds.length === 0 to the disabled expression, mirroring the gate WithdrawSelectContent already applies.

services/vault/src/components/simple/WithdrawFlow/index.tsx
```diff
@@ -167,9 +196,32 @@ function WithdrawFlowContent({
     goToProgress,
   ]);
 
+  // Review reached from Select can return to it; not while a submit is in flight.
   return (
-    <V3ModalShell open={open} onClose={onClose}>
+    <V3ModalShell
+      open={open}
+      onClose={onClose}
+      onBack={
```
jonybur: **A failed submit's error banner survives the new Back trip and is re-shown against a different selection.**

error lives in useWithdrawCollateralTransaction and is cleared only at the start of the next executeWithdraw. With the back arrow this PR adds, a failed confirm now leaves Review reachable again: the submit fails, setConfirmed(null) runs, isProcessing drops, the chevron appears, and the user can go back to Select, change which vaults are checked, and return. Review then renders the previous attempt's "Transaction failed" callout underneath an enabled Confirm for a selection that failure never described. The hook already exports clearError and nothing in this flow calls it, while the sibling Borrow and Repay flows do; wire it into goToSelect so the back transition drops the stale failure.

services/vault/src/components/simple/WithdrawFlow/WithdrawSelectContent.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <div className="flex flex-col gap-2">
+        <Heading variant="h5" className="font-normal text-accent-primary">
+          {SELECT_COPY.heading}
+        </Heading>
+        <Text variant="body1" className="text-accent-secondary">
+          {SELECT_COPY.subtitle}
+        </Text>
+      </div>
+
+      {(wouldBreachHF || isAtRisk) && (
```
jonybur: **The risk callout fires on an empty selection, announcing a health-factor drop for a withdrawal of nothing.**

The callout is gated on wouldBreachHF || isAtRisk alone, with no check that anything is selected. computeProjectedHealthFactor returns the current health factor unchanged when the withdrawal amount is zero, so a user already sitting in the at-risk band who unchecks every row sees "Liquidation risk is high — Your health factor will drop to 1.05" plus a "Continue anyway" tick for a withdrawal that would move nothing, and a user below the floor sees the on-chain rejection message for the same non-event. Unchecking the last row is a first-class state here — the new test suite exercises it — so this is easy to hit, not theoretical. Gate the callout on a non-empty selection as well.

jonybur (commented):
Comments inline, plus:

- `services/vault/src/components/simple/WithdrawFlow/__tests__/WithdrawReviewContent.test.tsx:103` - **A pre-existing Review test now passes only because baseProps hardcodes acknowledged: true, and its name states behaviour the code no longer has.**
  
  The test is named "warns without blocking when the projection is above the floor but at risk" and asserts that Confirm is enabled at a projected health factor of 1.05. This PR added (isAtRisk && !acknowledged) to the Confirm disabled expression, so at 1.05 the button does block; the assertion survives only because baseProps quietly gained acknowledged: true in the same diff. A reader now takes the opposite of the real rule from the test name, which is exactly what the test-philosophy rule about names describing the exact behaviour is there to prevent. Rename it to say it passes once acknowledged, or delete it as a duplicate of the new "enables Confirm once acknowledged" case a few lines below.

Issue #2413, opened 2026-09-07
Copy: finish the "vault" to "BTCVault" rename in the remaining strings


jeremy-babylonlabs:
## Context

`services/vault/src/copy.ts` states the naming rule at lines 20-23 — "BTCVault … when naming the product … never bare 'vault'" — and 108 strings already follow it. A handful do not.

## Changes

**Now:** bare "vault"/"Vault" as the product name remains in roughly a dozen strings:

- `copy.ts:559` — "Activate vault"
- `copy.ts:837,839,843` — `Vault ${vaultNumber}: …` templates
- `copy.ts:994-995` — "Vault operator keys changed" and its body
- `copy.ts:1916` — marketing copy
- `copy.ts:1945` — the "1 Vault" / "N Vaults" counter
- `copy.ts:1739,1812,1957` — bare "Vaults" in nav and page titles
- `copy.ts:1861` — `activeVaultsLabel: "Active Vaults"`

**Expected:** "BTCVault", per the rule already written at the top of the file.
**Where:** `services/vault/src/copy.ts`

Out of scope, leave as is: the role terms "vault provider" (71 occurrences) and "vault keeper" (1), and the app name "Babylon Vault" (3).

## Open question

Do the nav item and page title become "BTCVaults"? That reads long in the sidebar. Confirm with design before changing `copy.ts:1739,1812,1957`.

<details>
<summary>Original report</summary>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1787811285553199)

> • change "vault" to "BTCVault"

</details>

Issue #2412, opened 2026-09-07
Downloading BTCVault artifacts: give the download its own modal and progress bar


jeremy-babylonlabs:
## Context

The artifact download shown while activating a BTCVault. The design asks for a separate modal; the app has none.

### 1. There is no download modal

**Now:** `RecoveryArtifactsCard` renders its progress UI *inside* `ActivateConfirmationModal`. The icon, title and description only switch on `downloaded`, never on `loading`/`isDownloading`, so the download state reuses the activation modal's chrome. A stale comment at `services/vault/src/components/deposit/RecoveryArtifactsCard.tsx:202-203` refers to "ArtifactDownloadModal" — no such component exists in the tree.
**Expected:** its own modal with its own title, description and icon, per Figma `10060-45705`.
**Where:** `RecoveryArtifactsCard.tsx:207-244`, `services/vault/src/components/simple/ActivateConfirmationModal.tsx`

### 2. Progress bar

**Now:** `ProgressBar` from `services/vault/src/components/simple/DepositProgressView/ProgressBar` with byte counts and a percentage, fill hardcoded to `PROGRESS_BAR_FILL_COLOR = "#CE6533"` (`RecoveryArtifactsCard.tsx:29`).
**Expected:** the Figma progress bar.
**Where:** `RecoveryArtifactsCard.tsx:29,210-234`

### 3. Only a cancel button

**Now:** the footer at `ActivateConfirmationModal.tsx:202-224` renders both Cancel and Activate; Activate is `disabled` while `isDownloading` (`:98-101`), so only cancel is usable.
**Expected:** render only the cancel action during download, so the dead button is not shown.

<details>
<summary>Original report</summary>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1787811265200999)

> *Downloading BTCvault artifacts modal*
> • here should be another modal with another title, description, and icon
> • change progress bar UI
> • we have only a cancel button

</details>

PR #2569, opened 2026-09-18
fix(vault): size the Activate BTCVault modal to the design and drop the withdraw link

### LINKED ISSUE #2411: Activate BTCVault modal: cap the modal height, decide the fate of the withdraw-without-activating link
## Context

The Activate BTCVault modal, opened from a pending deposit on the Vaults page. Rendered by `services/vault/src/components/simple/ActivateConfirmationModal.tsx`.

### 1. Modal height

**Now:** the modal is a `ResponsiveDialog` at `ActivateConfirmationModal.tsx:104-108`, `w-[564px] max-w-full` with no height cap. On desktop `.bbn-dialog` (`packages/babylon-core-ui/src/components/Dialog/index.css:1-8`) sizes to content; below `WINDOW_BREAKPOINT` it swaps to `MobileDialog`, whose `.bbn-dialog-mobile` (`index.css:36-38`) is a bottom sheet at `max-h-full`.
**Expected:** a shorter modal, per Figma `10060-45730`.
**Where:** `ActivateConfirmationModal.tsx:104-108`

There is no branch that is unconditionally full-height. Either the review was done at the mobile breakpoint, or the content itself (icon + `RecoveryArtifactsCard` + checkbox + link) is what pushes the sheet to `max-h-full`. Confirm which before changing the dialog.

### 2. "Unable to activate?" link

**Now:** `ActivateConfirmationModal.tsx:188-197` renders `COPY.deposit.advancedWithdrawLink` — "Unable to activate? Withdraw without activating" (`services/vault/src/copy.ts:566`).
**Expected:** decide whether to keep it. Removing it takes away the only in-modal route to withdraw a deposit that cannot be activated.

## Open question

Which of the two causes above makes the modal tall? Answer decides whether this is a dialog change or a content change.

<details>

[Slack thread](https://babylonlabsworkspace.slack.com/archives/C07EG11U45A/p1787811237574319)

> *Activate your BTCVault modal*
> • let's make the height of the modal smaller, not the full height of the screen as now
> • Do we need the link "Unable to activate? withdraw without activation"? Can we remove it?

</details>

jeremy-babylonlabs:
Closes #2411

## What

Brings the Activate BTCVault modal to its Figma node and removes the "Unable to activate? Withdraw without activating" link.

Layout, in `ActivateConfirmationModal.tsx`:

- Width 564px → 600px; dialog background set to `background/contrast` (#F2F2F2) over the primitive's default white.
- Above the breakpoint, `DialogBody` and `DialogFooter` no longer carry their own `px-6` / `pt-10` / `pb-6`: `.bbn-dialog`'s own `p-6` is the only inset, so the horizontal padding is the design's 24px rather than the 48px the two layers stacked, and the top inset is 24px rather than 64px. Below the breakpoint the previous padding is kept: the mobile sheet's own inset is 16px, and dropping the component padding there would have narrowed the sheet's sides from 40px to 16px.
- Section gap 40px → 32px; header gap 40px → 24px and the title/body gap 16px → 24px.
- Both footer buttons pick up an 8px radius over the button primitive's 6px.
- The design's close control is rendered top-right as a labelled button (`aria-label` from `COPY.common.close`, the shape `V3ModalShell` uses), wired to the same guarded close handler as the footer's Cancel, so dismissing mid-download still cancels the in-flight artifact request. core-ui's `DialogHeader` was not reused: it renders an empty `h5` and an unlabelled icon button. The control is rendered above the breakpoint only, gated on `useIsMobile(WINDOW_BREAKPOINT)` so the decision matches `ResponsiveDialog`'s own switch: the mobile sheet already draws its own close button.
- The host's `V3ModalShell` draws its own close X, but it is not a second live control while this dialog is open: the dialog mounts in a later portal root, and each portal root is a stacking context, so the whole dialog including its backdrop paints above the shell. A click on the shell's X lands on the backdrop and dismisses through the guarded handler.

No `max-h` was added. The modal is content-sized on desktop, which is what the design shows; the reported full-screen height came from the content stack, not from a height rule.

Link removal: the link and its `onAdvancedWithdraw` prop are gone, along with the prop's chain through `ActivationGate`, `PostDepositContinuationView`, `PostDepositContinuationContent` and `VaultsLifecycleSections`, and the `advancedWithdrawLink` string in `copy.ts`. Removing the last caller of the advanced route also orphaned its plumbing, so this change collapses it: `handleWithdrawClick` loses its single-valued `source` argument, `stuckStateDetected` is gone from the hook and from both `EmergencyWithdrawModal` components, and the unreachable `bodyAdvanced` copy is deleted. The two `EmergencyWithdrawModal` tests drop the prop they no longer need.

Two tests are added to `RecoveryArtifactsCard.test.tsx` covering the idle card's labels and its download wiring, and two to `ActivateConfirmationModal.test.tsx`: one asserting the close control calls `onClose`, one asserting it is absent below the breakpoint.

## Why

https://github.com/babylonlabs-io/babylon-toolkit/issues/2411

The modal was reported as rendering at the full height of the screen, and the withdraw link it carried is not in the design. The link was a second entry point into the activate-and-redeem escape hatch; the Vaults pending row reaches the same `EmergencyWithdrawModal` through `PendingRow.routeAction` → `handleEmergencyWithdraw`, which is what surfaces automatically for a deposit that genuinely cannot be activated.

## Approaches

Capping the height with `max-h-[Npx] overflow-auto` was rejected: it hides content behind a scrollbar the design does not have, and it treats the symptom rather than the content stack that produced the height.

Moving the withdraw link onto the Vaults row was rejected: the row already carries that action.

Leaving the orphaned `"advanced"` branch in place was rejected. It is dead the moment this change lands, and the repository's zero-dead-code rule covers references a change orphans, not only the lines it deletes.

## Not in this PR

- The 4px progress rule the design places under the header. The modal has no step source to drive it; a hardcoded percent would be a magic number, and reading live progress state belongs to https://github.com/babylonlabs-io/babylon-toolkit/issues/2412, which owns the progress UI. (pre-review #5)
- The mobile sheet keeps the primitive's white background. `ResponsiveDialog` forwards `dialogClassName` only to the desktop dialog, and `MobileDialog` does not accept it, so the design background reaches desktop only. The mobile-sheet primitives are out of scope here. (pre-review #4)
- Confirmation that the rendered height matches the design. The reduction is derived from the token values, not measured in a browser, and the modal is not covered by a layout test. (pre-review #6)
- Product confirmation that dropping the link is the intended behaviour for a Verified vault whose peg-in has not been swept. The row's withdraw action appears only once the peg-in is swept, so that state loses a voluntary early opt-out; the swept-state action and the post-expiry refund both remain. (pre-review #3)
- The activate button label, owned by https://github.com/babylonlabs-io/babylon-toolkit/issues/2413.

## Pre-review

<details>
<summary>Pre-review: 11 findings · 3 fixed · 0 moot · 5 follow-up · 3 declined · 0 open (0 merge-blockers) · 0 undecided · lint/test passed</summary>

|---|---|---|---|
| 1 | Removing the advanced-withdraw link orphaned its plumbing: the `"advanced"` source, `stuckStateDetected`, and `bodyAdvanced` are unreachable | merge-blocker | fixed |
| 2 | Three comments still describe the advanced-withdraw link as a live UI path | normal | fixed |
| 9 | The mobile sheet renders two close controls once the header is added | normal | fixed |
| 7 | The design's close button is absent from the desktop dialog | normal | follow-up: give core-ui's `DialogHeader` an accessible name and a title-less layout |
| 3 | The row offers the withdraw action only once the peg-in is swept, so a Verified-but-unswept vault loses a voluntary early opt-out | normal | follow-up: product to confirm dropping the early opt-out |
| 4 | The new background token never reaches the mobile bottom sheet | normal | follow-up: forward `dialogClassName` to `MobileDialog` |
| 5 | The 4px progress rule the design specifies was not implemented | normal | follow-up: add it once #2412 establishes the step source |
| 6 | The height reduction is derived from tokens, never rendered | normal | follow-up: confirm against the design node in a browser pass |
| 8 | The layout values this change introduces are not covered by any test | normal | declined: layout assertions would be mirror tests |
| 10 | The modal test file mocks core-ui's `Text`, which the component does not render | normal | declined: pre-existing on main, not orphaned by this change |
| 11 | The close-control tests pin the conditional but not the breakpoint it reads | normal | declined: reviewer raised it as a note; the gate is mutation-verified |

Two reviewer claims were disproved and are recorded as refuted: that the buttons' `rounded-lg` is a no-op (the vault's own utility is emitted after the core-ui bundle, so 8px wins), and that the dialog border stays `#d7e1e7` (the app overrides the token to `#DDDDDD`).

</details>

Activate your BTCVault modal with the close control, default state

Activate your BTCVault modal, risk acknowledged


services/vault/src/components/simple/ActivateConfirmationModal.tsx
```diff
@@ -104,15 +105,16 @@ export function ActivateConfirmationModal({
...
-          handleClose via ResponsiveDialog). The top padding stands in for
-          the removed header row. */}
-      <DialogBody className="flex flex-col items-stretch gap-10 px-6 pb-2 pt-10 text-accent-primary">
-        <div className="flex flex-col items-center gap-10">
+      {/* Header carries only the design's close control: the title is not a
+          header row here, it sits centred in the body under the icon. The
+          empty title is load-bearing — the header row is justify-between, so
+          it is what holds the button to the right. */}
+      {!isMobile && <DialogHeader title="" onClose={handleClose} />}
+      <DialogBody className="flex flex-col items-stretch gap-8 text-accent-primary">
```
jonybur: **Removing DialogBody/DialogFooter padding also strips the mobile sheet's inset (40px→16px sides, 88px→48px top), which the description does not account for.**

The description justifies dropping `px-6 pt-10 pb-2` from `DialogBody` and `px-6 pb-6` from `DialogFooter` because `.bbn-dialog`'s `p-6` becomes the only inset, but that only holds on desktop. Below `WINDOW_BREAKPOINT`, `ResponsiveDialog` renders `MobileDialog`, whose `.bbn-dialog-mobile` only has `px-4 pb-4 pt-12` (packages/babylon-core-ui/src/components/Dialog/index.css line 39). So on phones the side inset drops from 40px to 16px, the top from 88px to 48px and the bottom from 40px to 16px, which lines up with the mobile screen change in the visual-regression report. Could you confirm this tighter sheet matches the mobile design, or keep the previous padding under the mobile breakpoint (e.g. via the existing `isMobile` flag)?

jonybur (commented):
Comments inline.

gbarkhatov (dismissed):

gbarkhatov (dismissed):

jonybur (commented):
Comments inline.

services/vault/src/components/simple/ActivateConfirmationModal.tsx
```diff
@@ -104,15 +105,16 @@ export function ActivateConfirmationModal({
...
-          through the footer actions (Escape/backdrop still route through
-          handleClose via ResponsiveDialog). The top padding stands in for
-          the removed header row. */}
-      <DialogBody className="flex flex-col items-stretch gap-10 px-6 pb-2 pt-10 text-accent-primary">
-        <div className="flex flex-col items-center gap-10">
+      {/* Header carries only the design's close control: the title is not a
+          header row here, it sits centred in the body under the icon. The
+          empty title is load-bearing — the header row is justify-between, so
+          it is what holds the button to the right. */}
+      {!isMobile && <DialogHeader title="" onClose={handleClose} />}
```
jonybur: **The new close control has no accessible name and adds an empty h5 heading to the dialog.**

The header renders `<DialogHeader title="" onClose={handleClose} />`. `DialogHeader` renders `<Heading variant="h5">{title}</Heading>` unconditionally and passes no label to its `IconButton` (packages/babylon-core-ui/src/components/Dialog/components/DialogHeader.tsx:15-18, packages/babylon-core-ui/src/components/Button/IconButton.tsx:16-25), so this dialog now ships an empty heading plus a close button whose accessible name is the empty string. A screen-reader user tabbing into the dialog hears "button" with no indication that it dismisses, where the mobile sheet's equivalent carries `aria-label="Close"`. The description defers this on the grounds that both fixes live in core-ui, but that only follows from reusing `DialogHeader`: V3ModalShell.tsx:77-85 already hand-rolls a labelled close button in this repo, and rendering that shape here — with the label added to `copy.ts` per the user-facing-copy rule — names the control and drops the empty heading without touching a shared primitive.

jonybur: **The surrounding V3ModalShell already draws a close X, so desktop now shows two live close controls that behave differently mid-download.**

The host reaches this modal through `V3ModalShell` (VaultsLifecycleSections.tsx:913 → PostDepositContinuationContent → PostDepositContinuationView → ActivationGate), and the shell draws its own close X at the top-left of an opaque full-screen dialog (V3ModalShell.tsx:77-85). That shell is `z-50` while this dialog's backdrop is `z-40`, so the shell is never dimmed and its X stays visible and clickable while the confirmation dialog is open — adding the header control puts two live close affordances on screen above the breakpoint. The comment at line 90 reasons only about the mobile sheet's built-in button, not about the shell. The two are also not equivalent: the shell's X calls its own `onClose` directly, so clicking it mid-download tears the modal down without `cardRef.current?.cancel()`, which is precisely the guarantee the new control was wired to `handleClose` for. Worth confirming which control the design intends, and routing whichever survives through the guarded handler.

services/vault/src/components/simple/__tests__/ActivateConfirmationModal.test.tsx
```diff
@@ -190,6 +203,35 @@ describe("ActivateConfirmationModal", () => {
     expect(onClose).toHaveBeenCalledTimes(1);
   });
 
+  it("calls onClose when the header close control is clicked", () => {
```
jonybur: **The new header-close test asserts only onClose, leaving the download cancellation it was wired for unpinned.**

This test clicks the header control and asserts only that `onClose` fired. The stated reason for routing that control through `handleClose` rather than handing `onClose` straight to `DialogHeader` is that dismissing mid-download cancels the in-flight artifact request (ActivateConfirmationModal.tsx:76-79), and that half goes unasserted — rewiring line 115 to `onClose={onClose}` keeps this test green while abandoning a streaming ~1 GB fetch. The sibling test at line 123 already guards the footer's cancel with `cardCancelSpy`, and the same setup applies here: click `card-download-start` first, then assert `cardCancelSpy` fired alongside `onClose`.

gbarkhatov (dismissed):
checks failing @jeremy-babylonlabs

services/vault/src/components/deposit/ArtifactDownloadContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+  /** The hook's status line, shown while the total is still unknown. */
+  status: string;
+}
+
+/**
+ * Body of the activate dialog while recovery artifacts stream: its own icon,
+ * title and description, the transfer's progress bar and the byte/percent
+ * card. Replaces the activation body for the duration of the download.
+ */
+export function ArtifactDownloadContent({
```
jonybur: **Merge blocker:** **The PR description covers none of the download redesign and its pre-review record predates this head.**

The description scopes this PR to modal sizing and the withdraw-link removal. It never mentions this new 140-line component, the card rendering nothing while bytes stream (RecoveryArtifactsCard.tsx:190), the body swap at ActivateConfirmationModal.tsx:126, the `onLoadingChange` to `onStateChange` change, or the Activate button being removed from the footer rather than disabled during a download (ActivateConfirmationModal.tsx:225) - all user-visible behaviour changes. It also asserts things the head does not contain: that the close control is a labelled button and that the previous padding is kept below the breakpoint, neither of which is in ActivateConfirmationModal.tsx. The embedded pre-review snapshot is pinned to `base=dbd74ebf… files=14` while this head sits on merge-base `fe7713c3` with 17 changed files, so the record does not cover this work at all. CLAUDE.md makes the description the review record the PR is opened from, and an approval is already on this PR; please regenerate the record and the description against this head before merging.

services/vault/src/components/simple/ActivateConfirmationModal.tsx
```diff
@@ -104,59 +114,68 @@ export function ActivateConfirmationModal({
...
-                    ),
-                  )}
-            </p>
+      {/* Header carries only the design's close control: the title is not a
+          header row here, it sits centred in the body under the icon. The
+          empty title is load-bearing — the header row is justify-between, so
+          it is what holds the button to the right. */}
+      {!isMobile && <DialogHeader title="" onClose={handleClose} />}
+      <DialogBody className="flex flex-col items-stretch gap-8 text-accent-primary">
+        {isDownloading ? (
```
jonybur: **The whole dialog swaps to the downloading screen while the OS save picker is still open, before any bytes move.**

`isDownloading` is `downloadState.loading`, and the hook sets `loading: true` with `progress = choosingSaveLocation` at useArtifactDownload.ts:179-185, before it awaits `openArtifactSaveTarget` at line 205 - that await blocks on `showSaveFilePicker`. So the instant Download is pressed, the icon, title, body and Activate button are replaced by "Downloading BTCVault artifacts" while the user has not yet chosen a file, and if they dismiss the picker (line 208 resets to the initial state) the entire dialog snaps back. The old code entered the same state, but it only swapped a small chip inside the card, so the dialog kept its identity; promoting it to the whole body makes the premature transition a visible, jarring one that also claims a download is in progress when none is. Gate the body swap on the transfer having actually started - for example require `totalBytes > 0`, or a status past `choosingSaveLocation` - and keep the card's chip for the pre-transfer statuses.

services/vault/src/components/simple/ActivateConfirmationModal.tsx
```diff
@@ -85,16 +92,19 @@ export function ActivateConfirmationModal({
...
+  // exactly one of the two renders at every width.
+  const isMobile = useIsMobile(WINDOW_BREAKPOINT);
+
   const canRenderCard = Boolean(providerAddress && peginTxid && depositorPk);
   const gate = useProtocolGateState();
-  // Blocked while a download streams: confirming unmounts this modal and
-  // would abandon the in-flight transfer uncancelled.
   const canActivate =
     (downloaded || acknowledged) &&
     !isDownloading &&
```
jonybur, line 102: **The `!isDownloading` term in `canActivate` is unreachable now that the Activate button is unmounted during a download.**

`canActivate` is computed at lines 108-111 with `!isDownloading` as one term, but its only reader is the `disabled` prop at line 232, which sits inside the `{!isDownloading && (…)}` block introduced at line 225. Both read the same `isDownloading` value in the same render, so wherever the term is evaluated it is always true - it can never change the result. The diff also deleted the comment that explained why it was there ("confirming unmounts this modal and would abandon the in-flight transfer uncancelled"), leaving an unexplained no-op guard, which is the orphaned-reference case the zero-dead-code policy covers and the same rule this PR cites to justify collapsing the `"advanced"` plumbing. Either drop `!isDownloading` from `canActivate`, or keep the button rendered-and-disabled during the download and restore the comment.

services/vault/src/components/deposit/RecoveryArtifactsCard.tsx
```diff
@@ -192,57 +176,18 @@ export const RecoveryArtifactsCard = forwardRef<
   }, [delivered, onDelivered]);
 
   useEffect(() => {
-    onLoadingChange?.(loading);
-  }, [loading, onLoadingChange]);
+    onStateChange?.({ loading, receivedBytes, totalBytes, status: progress });
```
jonybur: **Lifting per-chunk progress into the modal re-renders the whole dialog tree on every read of a ~1 GB stream.**

`artifactDownloadService.ts:247` calls `onProgress` once per `reader.read()` chunk with no throttling, and `useArtifactDownload.ts:364-366` turns each into a `setState`. The effect here now forwards every one of those ticks to the parent through `onStateChange`, where it lands in `ActivateConfirmationModal`'s own state (line 189), so `ResponsiveDialog`, `DialogBody`, `ArtifactDownloadContent` and the SVG ring all re-render at chunk rate for the duration of a roughly 1 GB transfer - on a fast link that is on the order of hundreds of renders per second. Previously the ticks only re-rendered the card's own subtree and the parent saw about two updates per download. Coalescing the lift (throttle, or only forward when the rounded percent changes) keeps the new presentation without paying for it on every chunk.

jonybur (commented):
Comments inline.

jrwbabylonlab (dismissed):

gbarkhatov (approved):
checks failing

gbarkhatov (approved):

services/vault/src/components/simple/__tests__/ActivateConfirmationModal.test.tsx
```diff
@@ -34,12 +35,19 @@ vi.mock("@babylonlabs-io/core-ui", () => ({
...
+  // className is forwarded so the sheet-inset assertions can read it.
   DialogBody: (props: Record<string, unknown>) => (
-    <div>{props.children as ReactNode}</div>
+    <div data-testid="dialog-body" className={props.className as string}>
+      {props.children as ReactNode}
+    </div>
   ),
   DialogFooter: (props: Record<string, unknown>) => (
-    <div>{props.children as ReactNode}</div>
+    <div data-testid="dialog-footer" className={props.className as string}>
```
jonybur: **The DialogFooter mock gains a data-testid and className forwarding that no test reads, and the mobile footer inset it was added for stays untested.**

The core-ui mock in this file now renders DialogFooter with `data-testid="dialog-footer"` and forwards its className, under a comment saying the className is forwarded so the sheet-inset assertions can read it. Nothing reads it: that line is the only occurrence of `dialog-footer` anywhere under `services/vault/src`, and the new "restores the sheet inset below the breakpoint" test asserts only against `dialog-body`. So the footer half of this change, `isMobile && "px-6 pb-6"` at ActivateConfirmationModal.tsx:214 — the part that keeps the mobile sheet's 40px sides and bottom instead of the 16px `.bbn-dialog-mobile` supplies — has no coverage at all, and deleting that conditional would reintroduce the tighter sheet with the suite still green. It also leaves unused scaffolding in a test, which the zero-dead-code rule covers explicitly. Assert the footer's class alongside the body's in that test, which closes the gap and makes the mock's testid live.

jonybur (commented):
Comments inline.

